PR #67, opened 2016-02-11
chore(): change all components to onPush detection

jelbourn:
R: @hansl @kara 
CC: @robertmesserle 

This is something I'd forgot to mention earlier. The general idea here is that the `onPush` strategy tells Angular to only check the component and its view children when either one of its `@Input` properties change or when an event is handled. For components that only bind either primitives or immutable objects, this should "just work" and speed up change detection. As far as I can tell, all of the components so far just bind primitives (or nothing at all, for cards).


hansl:
Interesting. LGTM.

hansl:
Question: is it only for this component or also for its content components?

jelbourn:
@hansl It applies to the the component's view but not its content.

PR #68, opened 2016-02-11
feat(md-whiteframe): Create Whiteframe

bakgat:
The MdWhiteframe component based on angular 1.x version.

### DIFF
--- a/src/components/whiteframe/README.md
+++ b/src/components/whiteframe/README.md
@@ -1,0 +1,23 @@
+# MdWhiteframe
+
+MdWhiteframe is a component that applies an elevation shadow to an element.
+
+## Screenshot
+<img src="http://www.klimtoren.be/content/material2/mdwhiteframe-example.png">
+
+### Notes
+* The values needs be a number between 1 and 24.
+* If there is no value specified it defaults to 4dp.
+* If the value is not valid it defaults to 4 dp.
+
+### References
+<a href="https://www.google.com/design/spec/what-is-material/elevation-shadows.html">
+https://www.google.com/design/spec/what-is-material/elevation-shadows.html
+</a>
+
+### Usage
+```html
+<div md-whiteframe="3">
+    Elevated 3dp.
+</div>
+```
--- a/src/components/whiteframe/whiteframe.html
+++ b/src/components/whiteframe/whiteframe.html
@@ -1,0 +1,1 @@
+<ng-content></ng-content>
--- a/src/components/whiteframe/whiteframe.scss
+++ b/src/components/whiteframe/whiteframe.scss
@@ -1,0 +1,80 @@
+@import "variables";
+
+.md-whiteframe-1dp, .md-whiteframe-z1 {
+  box-shadow: $whiteframe-shadow-1dp;
+}
+.md-whiteframe-2dp {
+  box-shadow: $whiteframe-shadow-2dp;
+}
+.md-whiteframe-3dp {
+  box-shadow: $whiteframe-shadow-3dp;
+}
+.md-whiteframe-4dp, .md-whiteframe-z2{
+  box-shadow: $whiteframe-shadow-4dp;
+}
+.md-whiteframe-5dp {
+  box-shadow: $whiteframe-shadow-5dp;
+}
+.md-whiteframe-6dp {
+  box-shadow: $whiteframe-shadow-6dp;
+}
+.md-whiteframe-7dp, .md-whiteframe-z3 {
+  box-shadow: $whiteframe-shadow-7dp;
+}
+.md-whiteframe-8dp {
+  box-shadow: $whiteframe-shadow-8dp;
+}
+.md-whiteframe-9dp {
+  box-shadow: $whiteframe-shadow-9dp;
+}
+.md-whiteframe-10dp, .md-whiteframe-z4 {
+  box-shadow: $whiteframe-shadow-10dp;
+}
+.md-whiteframe-11dp {
+  box-shadow: $whiteframe-shadow-11dp;
+}
+.md-whiteframe-12dp {
+  box-shadow: $whiteframe-shadow-12dp;
+}
+.md-whiteframe-13dp, .md-whiteframe-z5{
+  box-shadow: $whiteframe-shadow-13dp;
+}
+.md-whiteframe-14dp {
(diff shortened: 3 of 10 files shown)


src/components/whiteframe/whiteframe.ts
```diff
@@ -0,0 +1,39 @@
...
+    ngOnInit() {
+        this.elevation_ = parseInt(this.mdWhiteframe, 10) || MdWhiteframe.DEFAULT_DP;
+
+        if (this.elevation_ > MdWhiteframe.MAX_DP || this.elevation_ < MdWhiteframe.MIN_DP) {
+            this.elevation_ = MdWhiteframe.DEFAULT_DP;
+        }
+
+        this.dpCssClass = 'md-whiteframe-' + this.elevation_ + 'dp';
+        this.el.nativeElement.className +=
+            (this.el.nativeElement.className !== '' ? ' ' : '') +
```
DevVersion: Why you append the class to `className`?. `classList` will do the same in a native element. ([here](https://developer.mozilla.org/de/docs/Web/API/Element/classList)

Better use:
```
this.el.nativeElement.className.add(this.dpCssClass)
```

src/components/whiteframe/whiteframe.ts
```diff
@@ -0,0 +1,39 @@
+import {Component, Input, ElementRef,ViewEncapsulation, OnInit} from 'angular2/core';
```
DevVersion: Little style thing - You should space the `ViewEncpsulation` from the comma.

src/components/whiteframe/whiteframe.ts
```diff
@@ -0,0 +1,39 @@
...
+@Component({
+    selector: '[md-whiteframe]',
+    templateUrl: './components/whiteframe/whiteframe.html',
+    styleUrls: ['./components/whiteframe/whiteframe.css'],
+    encapsulation: ViewEncapsulation.None
+})
+export class MdWhiteframe implements OnInit {
+    @CONST() static MIN_DP:number = 1;
+    @CONST() static MAX_DP:number = 24;
+    @CONST() static DEFAULT_DP:number = 4;
```
DevVersion: I also suggest spacing the type definition from the constant (to have the same code-style as in other components). 
Like so: `@CONST() static DEFAULT_DP: number = 4;`

src/components/whiteframe/whiteframe.html
```diff
@@ -0,0 +1 @@
+<ng-content></ng-content>
```
DevVersion: I'm not sure about using a component for the `whiteframe`. Wrapping the content in a component template makes not really a sense. Thoughts?

src/components/whiteframe/whiteframe.scss
```diff
@@ -0,0 +1,80 @@
...
+.md-whiteframe-7dp, .md-whiteframe-z3 {
+  box-shadow: $whiteframe-shadow-7dp;
+}
+.md-whiteframe-8dp {
+  box-shadow: $whiteframe-shadow-8dp;
+}
+.md-whiteframe-9dp {
+  box-shadow: $whiteframe-shadow-9dp;
+}
+.md-whiteframe-10dp, .md-whiteframe-z4 {
```
DevVersion: I would not start introducing the `zN` whiteframe classes. These are deprecated and will be soon removed from `material1`

src/components/whiteframe/README.md
```diff
@@ -0,0 +1,23 @@
+# MdWhiteframe
+
+MdWhiteframe is a component that applies an elevation shadow to an element.
+
+## Screenshot
+<img src="http://www.klimtoren.be/content/material2/mdwhiteframe-example.png">
```
DevVersion: Also hosting the screenshot on your private server is not a good idea I think.

src/components/whiteframe/whiteframe.ts
bakgat: Done. Changed to this.el.nativeElement.classList.add(this.dpCssClass)

src/components/whiteframe/whiteframe.ts
bakgat: Thank you, done.

src/components/whiteframe/whiteframe.ts
bakgat: How about `private elevation_:number;` ?

(I'm glad to get all the feedback. I've never contributed any project, so it's very new for me :smile: )

src/components/whiteframe/whiteframe.scss
bakgat: Removed zN classes.

src/components/whiteframe/README.md
bakgat: Where do I host screenhost better?

src/components/whiteframe/whiteframe.html
bakgat: I believe Directive whould be better. But where should the css classes be loaded then?

src/components/whiteframe/README.md
DevVersion: The Material team itself host their images [there](https://github.com/angular/code.material.angularjs.org). So I think they will host the screenshot their self.

src/components/whiteframe/whiteframe.html
DevVersion: Yes that will be a problem. We need a feedback from one of the team.

src/components/whiteframe/README.md
bakgat: Do I remove the screenshot then?

DevVersion: No keep it temporary

jelbourn:
@bakgat Please read our [contributing guidelines](https://github.com/angular/material2/blob/master/CONTRIBUTING.md).

Before beginning work on any new features, please open an issue so we can discuss before anyone starts writing code. At this point, we're not sure `whiteframe` is one of the components we want to include in material2.

bakgat:
@jelbourn Well, I've learned a lot about contributing today. :smile: 
Is any contributing accepted right now? Because I've coded some basic components already to get my own apps up and running.

jelbourn:
@bakgat  Yes, we'd love to have contributions! It's important to go through the documented process, though, so that everyone is on the same page. The first step is opening an issue with the feature you're proposing and we can go from there.

Issue #74, opened 2016-02-12
Error: Cannot find module 'request'


mcwebdev:
```
$ ng build --watch
version: 1.13.13
Cannot find module 'request'
Error: Cannot find module 'request'
    at Function.Module._resolveFilename (module.js:339:15)
    at Function.Module._load (module.js:290:25)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Leek._enqueue (C:\Users\mattchar\AppData\Roaming\npm\node_modules\angular-cli\node_modules\leek\lib\leek.js:60:30)
    at Leek.track (C:\Users\mattchar\AppData\Roaming\npm\node_modules\angular-cli\node_modules\leek\lib\leek.js:87:15)
    at Class.Command.validateAndRun (C:\projects\testit\node_modules\ember-cli\lib\models\command.js:117:18)
    at C:\projects\testit\node_modules\ember-cli\lib\cli\cli.js:86:22
    at lib$rsvp$$internal$$tryCatch (C:\projects\testit\node_modules\rsvp\dist\rsvp.js:1036:16)
    at lib$rsvp$$internal$$invokeCallback (C:\projects\testit\node_modules\rsvp\dist\rsvp.js:1048:17)

```

gbataille:
+1

If I use npm run build (which supposedly calls ng build), I get `Cannot find module './source-map/source-map-generator'` coming from some ts2dart dependency...

I saw in package.json that engines is `node < 5`. I have node 5.1.0. Could it be the issue?

gbataille:
that's it, at least for me. Downgrading to node 4.3.0 works. Don't know why though.

sothychan:
I have not run into this issue at all.  I'm running Node 5.5.

gbataille:
agreed, works in 5.6 for me too...
so I think I used to be in 5.1. Maybe there is a problem there.
@mcwebdev which version of node are you using?

jelbourn:
@hansl do you know anything about the CLI and required node version?

hansl:
We test the CLI on Node 4 and Node 5 for every commits: https://github.com/angular/angular-cli/blob/master/.travis.yml. This might be ts2dart specific.

jelbourn:
Looks like it might just be with node 5.1, so updating node should be the solution for this.

PR #77, opened 2016-02-15
fix(#74): fix node version with nvm

gbataille:


hansl:
Let's drop this PR after thje discussion in #74.

PR #86, opened 2016-02-16
fix(): OnPush does not work for the SidenavLayout.

hansl:


hansl:
@jelbourn: For Layout it's probably fine (performance-wise). I understand what's going on here and it's possible to fix it with `OnPush` strategy, but just not as elegant as the default strategy.

jelbourn:
What's the TL;DR why push state doesn't work?

hansl:
I've added a comment for this. PTAL.

jelbourn:
LGTM

PR #89, opened 2016-02-19
Fix #75: md-button attaches class "md-undefined" to host element when on color provided

### LINKED ISSUE #75: md-button attaches class "md-undefined" to host element when on color provided
Given an html snippet with an md-button directive like so:

```html
```

This produces the following ouput:

```html
<button md-button ... class="md-undefined">...</button>
```

What it should produce is:

```html
<button md-button ...>...</button>
```

The `md-undefined` class is getting attached due to [this code](https://github.com/angular/material2/blob/master/src/components/button/button.ts#L21-L40) since `setClassList()` returns `"md-undefined"` when the `color` attribute is unset. 

I tried working around the by returning an empty string when `color` is undefined, but that resulted in this html:

```html
<button md-button ... class>...</button>
```

Which is also invalid.

bakgat:


src/components/button/button.ts
```diff
@@ -1,11 +1,14 @@
...
-  HostListener,
-  ChangeDetectionStrategy,
+    Component,
+    ElementRef,
+    ViewEncapsulation,
+    Input,
+    HostBinding,
+    HostListener,
+    ChangeDetectionStrategy,
+    Renderer,
```
DevVersion: As mentioned in the style conventions, you should not indent with a tab, better indent with two spaces.

bakgat: done

src/components/button/button.ts
```diff
@@ -37,15 +55,15 @@ export class MdButton {
...
-  setClassList() {return `md-${this.color}`;}
-
   setMousedown() {
     // We only *show* the focus style when focus has come to the button via the keyboard.
     // The Material Design spec is silent on this topic, and without doing this, the
     // button continues to look :active after clicking.
     // @see http://marcysutton.com/button-focus-hell/
     this.isMouseDown = true;
-    TimerWrapper.setTimeout(() => { this.isMouseDown = false; }, 100);
+    TimerWrapper.setTimeout(() => {
```
DevVersion: Why aren't you using the lambda expression instead?
```js
TimeWrapper.setTimeout(() => this.isMouseDown = false, 100);
```

hansl: You need brackets for ts2dart to work as expected.

DevVersion: @hansl Ah well, thanks for the information :)

hansl:
Two things; you need a proper constructor in MdAnchor to fix the Dartanalyzer errors. And you should squash your commits so this PR includes only one commit. I'm going to review the code later when I have some time.

src/components/button/button.ts
hansl: This is because `TimerWrapper.setTimeout()` expects the function to be void. Dart is stricter than TypeScript in that it strictly enforce return values of callbacks. In TypeScript the return value is simply ignored. Please note that there are proposals to enforce stricter rules in TypeScript as well.

DevVersion: Thanks great!, indeed - just saw, lambda will always return the expression.

bakgat:
Thanks @hansl , still learning the first steps in contributing. Fixed the issues.

src/components/button/button.ts
```diff
@@ -37,15 +55,15 @@ export class MdButton {
   /** Whether a mousedown has occurred on this element in the last 100ms. */
   isMouseDown: boolean = false;
 
-  setClassList() {return `md-${this.color}`;}
```
jelbourn: Why can this fix not be something like:
```ts
setClassList() {
  return color != null ? `md-${this.color}` : '';
}
```
?

bakgat: 1. This results in an empty class attribute which is not valid XHTML
2. It overrides any other specified class in the component

`<button md-button class="myClass" color="warn">Submit</button>`
results in 
`<button md-button class="md-warn">Submit</button>`

It should be 
`<button md-button class="myClass md-warn">Submit</button>`

src/components/button/button.ts
```diff
@@ -29,7 +30,24 @@ import {TimerWrapper} from 'angular2/src/facade/async';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class MdButton {
-  color: string;
+  private color_: string;
+
+  // Setting the class when a color property is provided
+  // cannot be done through [ngClass] on host or [class] = 'getClassList()'
+  // @see http://stackoverflow.com/questions/35506395/looking-for-nativeelement-classlist-add-alternative/35514006
```
jelbourn: You can shorten this link:
```
http://stackoverflow.com/questions/35506395
```

src/components/button/button.ts
```diff
@@ -29,7 +30,24 @@ import {TimerWrapper} from 'angular2/src/facade/async';
...
+  @Input()
+  set color(value: string) {
+    this.color_ = value;
+
+    this.renderer.setElementClass(this.elementRef, 'md-' + this.color_, true);
```
jelbourn: If the color changes, I don't think this removes the previous class.

src/components/button/button.ts
jelbourn: I chatted with Misko and he agreed that this is an issue in Angular: https://github.com/angular/angular/issues/7289

In the meantime I would be okay with the renderer approach.

jelbourn:
This change needs to add unit tests that cover the bug.

bakgat:
I confirm I did author these commits.

bakgat:
Two things:
1. Why do the unit tests fail? There's a timeout for two tests. When I test locally, they succeed.
2. I've changed the fix. When the color attribute change, the previous 'md-color_' is replaced by the new one. 
But: when a color is provided and later on it is removed (color=null) (ánd there are no other css-classes)
the class attribute is left empty, which is invalid xhtml.
This is a bug in angular.

PR #96, opened 2016-02-25
feat(portal): add initial portal foundation

jelbourn:
R: @hansl @kara 

See the [design doc](https://docs.google.com/document/d/1jSMjkKSdzbQa2cuCeSFpzFAwDHr7xY4eqsXkfF6MeqA/preview)

### DIFF
--- a/src/core/portal/README.md
+++ b/src/core/portal/README.md
@@ -1,0 +1,72 @@
+# Portals
+
+### Overview
+
+A `Portal `is a piece of UI that can be dynamically rendered to an open slot on the page. 
+
+The "piece of UI" can be either a `Component` or a `TemplateRef`.
+
+The "open slot" is a `PortalHost`.
+
+Portals and PortalHosts are low-level building blocks that other concepts, such as overlays, can
+be built upon.
+ 
+##### `Portal<T>`
+| Method | Description |
+| --- | --- |
+| `attach(PortalHost): Promise<T>` | Attaches the portal to a host. |
+| `detach(): Promise<void>` | Detaches the portal from its host. |
+| `isAttached: boolean` | Whether the portal is attached. |
+
+##### `PortalHost`
+| Method | Description |
+| --- | --- |
+| `attach(Portal): Promise<void>` | Attaches a portal to the host. |
+| `detach(): Promise<void>` | Detaches the portal from the host. |
+| `dispose(): Promise<void>` | Permanently dispose the host. |
+| `hasAttached: boolean` | Whether a portal is attached to the host. |
+
+
+
+
+### Using portals
+
+
+
+##### `TemplatePortalDirective`
+Used to get a portal from a `<template>`. `TemplatePortalDirectives` *is* a `Portal`.
+
+Usage:
+```html
+<template portal>
+  <p>The content of this template is captured by the portal.</p>
+</template>
+
+<!-- OR -->
+
+<!-- This result here is identical to the syntax above -->
+<p *portal>
+  The content of this template is captured by the portal.
+</p>
+```
+
+A component can use `@ViewChild` or `@ViewChildren` to get a reference to a 
+`TemplatePortalDiective`.
+
+##### `ComponentPortal`
+Used to create a portal from a component type.
+
+Usage:
+```ts
+this.userSettingsPortal = new ComponentPortal(UserSettingsComponent);
+```
+
+
+##### `PortalHostDirective`
+Used to add a portal host to a template. `PortalHostDirective` *is* a `PortalHost`.
+
+Usage:
+```html
+<!-- Attaches the `userSettingsPortal` from the previous example. -->
+<template [portalHost]="userSettingsPortal"></template>
+```
--- a/src/core/portal/portal-directives.ts
+++ b/src/core/portal/portal-directives.ts
@@ -1,0 +1,104 @@
+import {Portal} from './portal';
+import {TemplatePortal} from './portal';
+import {ComponentRef} from 'angular2/core';
+import {ComponentPortal} from './portal';
+import {Directive} from 'angular2/core';
+import {TemplateRef} from 'angular2/core';
+import {BasePortalHost} from './portal';
+import {DynamicComponentLoader} from 'angular2/core';
+import {ElementRef} from 'angular2/core';
+import {ViewContainerRef} from 'angular2/core';
+
+
+/**
+ * Directive version of a `TemplatePortal`. Because the directive *is* a TemplatePortal,
+ * the directive instance itself can be attached to a host, enabling declarative use of portals.
+ *
+ * Usage:
+ * <template portal #greeting>
+ *   <p> Hello {{name}} </p>
+ * </template>
+ */
+@Directive({
+  selector: '[portal]',
+  exportAs: 'portal',
+})
+export class TemplatePortalDirective extends TemplatePortal {
+  constructor(templateRef: TemplateRef) {
+    super(templateRef);
+  }
+}
+
+
+/**
+ * Directive version of a PortalHost. Because the directive *is* a PortalHost, portals can be
+ * directly attached to it, enabling declarative use.
+ *
+ * Usage:
+ * <template [portalHost]="greeting"></template>
+ */
+@Directive({
+  selector: '[portalHost]',
+  inputs: ['portal: portalHost']
+})
+export class PortalHostDirective extends BasePortalHost {
+  /** The attached portal. */
+  private portal_: Portal<any>;
+
+  constructor(
+      private dynamicComponentLoader_: DynamicComponentLoader,
+      private elementRef_: ElementRef,
+      private viewContainerRef_: ViewContainerRef) {
+    super();
+  }
+
+  get portal(): Portal<any> {
+    return this.portal_;
+  }
+
+  set portal(p: Portal<any>) {
+    this.replaceAttachedPortal_(p);
+  }
+
+  /** Attach the given ComponentPortal to this PortlHost using the DynamicComponentLoader. */
+  attachComponentPortal(portal: ComponentPortal): Promise<ComponentRef> {
+    portal.setAttachedHost(this);
+
+    // If the portal specifies an origin, use that as the logical location of the component
+    // in the application tree. Otherwise use the location of this PortalHost.
+    let elementRef = portal.origin != null ? portal.origin : this.elementRef_;
+
+    // Typecast is necessary for Dart transpilation.
+    return <Promise<ComponentRef>>
+        this.dynamicComponentLoader_.loadNextToLocation(portal.component, elementRef)
+        .then(ref => {
+          this.setDisposeFn(() => ref.dispose());
+          return ref;
+        });
+  }
+
+  /** Attach the given TemplatePortal to this PortlHost as an embedded View. */
+  attachTemplatePortal(portal: TemplatePortal): Promise<Map<string, any>> {
+    portal.setAttachedHost(this);
+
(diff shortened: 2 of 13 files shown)


src/core/portal/portal-directives.ts
```diff
@@ -0,0 +1,110 @@
+import {Portal} from './portal';
```
hansl: Why not put this in `src/directives/portal`? Is this not something we want to expose publicly?

src/core/portal/portal-directives.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Usage:
+ * <template [portalHost]="greeting"></template>
+ */
+@Directive({
+  selector: '[portalHost]',
+  inputs: ['portal: portalHost']
+})
+export class PortalHostDirective extends BasePortalHost {
+  /** The attached portal. */
+  private portal_: Portal<any>;
```
hansl: `any` might be a bit too generic? `Portal<Type>` might be better and convey your meaning.

src/core/portal/portal-directives.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  constructor(
+      private dynamicComponentLoader: DynamicComponentLoader,
```
hansl: `dynamicComponentLoader_`?

src/core/portal/portal-directives.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** Attach the given ComponentPortal to this PortlHost using the DynamicComponentLoader. */
+  attachComponentPortal(portal: ComponentPortal): Promise<ComponentRef> {
+    portal.setAttachedHost(this);
+
+    // If the portal specifies an origin, use that as the logical location of the component
+    // in the application tree. Otherwise use the location of this PortalHost.
+    let elementRef = portal.origin != null ? portal.origin : this.elementRef_;
+
+    // Typecast is necessary for Dart transpilation.
+    return <Promise<ComponentRef>>
```
hansl: It seems like this cast is unnecessary, as `loadNextToLocation` already returns this.

src/core/portal/portal-directives.ts
```diff
@@ -0,0 +1,110 @@
...
+    let viewRef = this.viewContainerRef_.createEmbeddedView(portal.templateRef);
+    portal.locals.forEach((k, v) => viewRef.setLocal(k, v));
+    this.setDisposeFn(() => this.viewContainerRef_.clear());
+
+    // TODO(jelbourn): return locals from view
+    // Typecast is necessary for Dart transpilation.
+    return <Promise<Map<string, any>>> PromiseWrapper.resolve(new Map<string, any>());
+  }
+
+  /** Detatches the currently attached Portal (if there is one) and attaches the given Portal. */
```
hansl: typo: `detaches`.

src/core/portal/portal-directives.ts
jelbourn: The general best practice is to organize code by feature / area than by type of thing it is (directive, service, etc.). As such, all the portal code should go together. This would be part of the general component-building toolkit.

src/core/portal/portal-directives.ts
jelbourn: For a `TemplatePortal` the generic is `Map<string, any>` rather than `Type`

src/core/portal/portal-directives.ts
jelbourn: Done

src/core/portal/portal-directives.ts
jelbourn: And yet DartAnalyzer fails completely without it.

src/core/portal/portal-directives.ts
jelbourn: Done.

hansl:
README.md could be really nice, even if it's simple.

jelbourn:
@hansl Added a readme. 

cc @pkozlowski-opensource since you've been looking at the design doc.

src/core/portal/README.md
```diff
@@ -0,0 +1,72 @@
...
+<p *portal>
+  The content of this template is captured by the portal.
+</p>
+```
+
+A component can use `@ViewChild` or `@ViewChildren` to get a reference to a 
+`TemplatePortalDiective`.
+
+##### `ComponentPortal`
+Used to create a portal form a component type.
```
kara: Typo?  `form` -> `from`

src/core/portal/portal-directives.ts
```diff
@@ -0,0 +1,110 @@
...
+  private replaceAttachedPortal_(p: Portal<any>): void {
+    if (this.hasAttached()) {
+      this.detach().then(_ => {
+        if (p != null) {
+          this.attach(p);
+          this.portal_ = p;
+        }
+      });
```
kara: Nitpicky, but these four lines (98-102) are repeated in 104-107. Perhaps they should be abstracted into a little method?

src/core/portal/README.md
jelbourn: Done.

src/core/portal/portal.ts
```diff
@@ -0,0 +1,194 @@
...
+      this.attachedPortal_ = portal;
+      return this.attachComponentPortal(portal);
+    } else if (portal instanceof TemplatePortal) {
+      this.attachedPortal_ = portal;
+      return this.attachTemplatePortal(portal);
+    }
+
+    throw new BaseException(
+        'Attempting to attach an unknown Portal type. ' +
+        'BasePortalHost accepts either a ComponentPortal or a TemplatePortal.');
```
kara: Maybe we could use a template string here for multi-line rather than concatenation?

kara:
Made some nitpicky comments, but LGTM

src/core/portal/portal.ts
jelbourn: I originally had a template string, but the line breaks carried into the error message, which made it not show the whole message without expansion in the dev tools.

src/core/portal/portal-directives.ts
jelbourn: Excellent point; I simplified it.

src/core/portal/portal.ts
kara: Ah, that makes sense then.

jelbourn:
Fixed a new emo bug that popped up (angular2-polyfills was loaded in the wrong place) and added another test for _changing_ the portal.

src/core/portal/portal-directives.ts
```diff
@@ -0,0 +1,105 @@
...
+import {TemplatePortal} from './portal';
+import {ComponentRef} from 'angular2/core';
+import {ComponentPortal} from './portal';
+import {Directive} from 'angular2/core';
+import {TemplateRef} from 'angular2/core';
+import {BasePortalHost} from './portal';
+import {DynamicComponentLoader} from 'angular2/core';
+import {ElementRef} from 'angular2/core';
+import {ViewContainerRef} from 'angular2/core';
+import {PromiseWrapper} from 'angular2/src/facade/promise';
```
hansl: You shouldn't need this.

src/core/portal/portal.ts
```diff
@@ -0,0 +1,194 @@
+import {TemplateRef, Type} from 'angular2/core';
+import {ElementRef} from 'angular2/core';
+import {ComponentRef} from 'angular2/core';
+
+import {BaseException} from 'angular2/src/facade/exceptions';
+import {PromiseWrapper} from 'angular2/src/facade/promise';
```
hansl: You should not need this.

src/core/portal/portal.ts
```diff
@@ -0,0 +1,194 @@
...
+    super();
+    this.templateRef = template;
+  }
+
+  get origin(): ElementRef {
+    return this.templateRef.elementRef;
+  }
+
+  attach(host: PortalHost, locals?: Map<string, any>): Promise<Map<string, any>> {
+    this.locals = locals == null ? new Map<string, any>() : locals;
```
hansl: `isPresent(locals)`?

src/core/portal/portal.ts
```diff
@@ -0,0 +1,194 @@
...
+
+
+/**
+ * A `TemplatePortal` is a portal that represents some embedded template (TemplateRef).
+ */
+export class TemplatePortal extends Portal<Map<string, any>> {
+  /** The embedded template that will be used to instantiate an embedded View in the host. */
+  templateRef: TemplateRef;
+
+  /** Additional locals for the instantiated embedded view. */
```
hansl: What are `locals`? Just scoped values to share between the portal and its host?

src/core/portal/portal.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      testAppComponent.selectedPortal = new ComponentPortal(PizzaMsg);
+      appFixture.detectChanges();
+
+      flushMicrotasks();
+
+      expect(hostContainer.textContent).toContain('Pizza');
+    }));
+
+  });
```
hansl: Regression of testing for the behaviour expected of `locals`?

src/core/portal/README.md
```diff
@@ -0,0 +1,72 @@
...
+| `detach(): Promise<void>` | Detaches the portal from its host. |
+| `isAttached: boolean` | Whether the portal is attached. |
+
+##### `PortalHost`
+| Method | Description |
+| --- | --- |
+| `attach(Portal): Promise<void>` | Attaches a portal to the host. |
+| `detach(): Promise<void>` | Detaches the portal from the host. |
+| `dispose(): Promise<void>` | Permanently dispose the host. |
+| `hasattached: boolean` | Whether a portal is attached to the host. |
```
hansl: `hasAttached()`

src/core/portal/portal.ts
jelbourn: They're extra locals passed to the [`EmbeddedView`](https://angular.io/docs/ts/latest/api/core/EmbeddedViewRef-class.html) via `setLocals`. I added a unit test where you can see how they're used.

src/core/portal/portal.ts
jelbourn: Why change it? Using `== null` is idomatic TS / JS that has the wonderfully rare benefit of also working in Dart.

src/core/portal/portal-directives.ts
jelbourn: Done.

src/core/portal/portal.ts
jelbourn: Done.

src/core/portal/portal.spec.ts
jelbourn: Well, this sent me down a long, long journey of trying to get the test working for this. There _was_ a bug, which was that I had `k` and `v` swapped in my `forEach`.

However, there is triumph, and there's now  a unit test for the `locals` behavior as well as a test for a `TemplatePortal` that contains a binding.

src/core/portal/README.md
jelbourn: Done.

jelbourn:
@hansl all comments should be addressed now. Good call on the `locals` thing, it had slipped my mind.

hansl:
Could you add a link to https://angular.io/docs/ts/latest/api/core/EmbeddedViewRef-class.html for the locals? I still find "local" ridiculously unfamiliar. Finally, if you can change `Portal<any>` to maybe `Portal<ComponentRef>` (or something close). `Portal<number>` doesn't make sense ;)

Otherwise, LGTM :+1:

jelbourn:
Done on the description for `locals`. As discussed, the `any` on `Portal` make sense because it really can be anything (e.g., `Portal<number>` would be very simple, but still make sense).

PR #125, opened 2016-02-29
feat(radio): Radio component.

rfru:
@jelbourn @hansl @kara 

Looking to get an initial pass first and then if it's looking close I'll update with tests.
Thanks!

### DIFF
--- a/src/components/radio/radio-button.scss
+++ b/src/components/radio/radio-button.scss
@@ -1,0 +1,129 @@
+@import "default-theme";
+
+$md-radio-width: 20px !default;
+
+// Top-level host container.
+md-radio-button {
+  display: inline-block;
+}
+
+// Inner label container, wrapping entire element.
+// Enables focus by click.
+.md-radio-label {
+  cursor: pointer;
+  display: block;
+  padding: 8px;
+  white-space: nowrap;
+}
+
+// Container for radio circles and ripple.
+.md-radio-container {
+  box-sizing: border-box;
+  display: inline-block;
+  height: $md-radio-width;
+  position: relative;
+  top: 2px;
+  width: $md-radio-width;
+}
+
+// TODO(mtlin): Replace when ink ripple component is implemented.
+// A placeholder ink ripple, shown when keyboard focused.
+.md-ink-ripple {
+  background-color: md-color($md-accent);
+  border-radius: 50%;
+  height: 48px;
+  left: 10px;
+  opacity: 0;
+  pointer-events: none;
+  position: absolute;
+  top: 10px;
+  transform: translate(-50%,-50%);
+  transition: opacity ease 0.28s, background-color ease 0.28s;
+  width: 48px;
+  overflow: hidden;
+
+  // Fade in when radio focused.
+  .md-radio-focused & {
+    opacity: 0.1;
+  }
+
+  // TODO(mtlin): This corresponds to disabled focus state, but it's unclear how to enter into
+  // this state.
+  .md-radio-disabled & {
+    background: #000;
+  }
+}
+
+// The outer circle for the radio, always present.
+.md-radio-outer-circle {
+  border-color: md-color($md-foreground, secondary-text);
+  border: solid 2px;
+  border-radius: 50%;
+  box-sizing: border-box;
+  height: $md-radio-width;
+  left: 0;
+  position: absolute;
+  top: 0;
+  transition: border-color ease 0.28s;
+  width: $md-radio-width;
+
+  .md-radio-checked & {
+    border-color: md-color($md-accent);
+  }
+
+  .md-radio-disabled & {
+    border-color: md-color($md-foreground, disabled);
+  }
+}
+
+// The inner circle for the radio, shown when checked.
+.md-radio-inner-circle {
+  background-color: md-color($md-accent);
+  border-radius: 50%;
+  box-sizing: border-box;
+  height: $md-radio-width;
+  left: 0;
+  position: absolute;
+  top: 0;
+  transition: transform ease 0.28s, background-color ease 0.28s;
+  transform: scale(0);
+  width: $md-radio-width;
+
+  .md-radio-checked & {
+    transform: scale(0.5);
+  }
+
+  .md-radio-disabled & {
+    background-color: md-color($md-foreground, disabled);
+  }
+}
+
+// Text label next to radio.
+.md-radio-label-content {
+  display: inline-block;
+  float: right;
+  line-height: 24px;
+  // Equal padding on both sides, for RTL.
+  padding-left: 8px;
+  padding-right: 8px;
+  position: relative;
+  vertical-align: top;
+}
+
+// Underlying native input element.
+// Visually hidden but still able to respond to focus.
+.md-radio-input {
+  position: absolute;
+  width: 0;
+  height: 0;
+  margin: 0;
+  padding: 0;
+  opacity: 0;
+  appearance: none;
+  border: none;
+}
+
+// Basic disabled state.
+.md-radio-disabled, .md-radio-disabled .md-radio-label {
+  cursor: default;
+}
--- a/src/components/radio/radio_button.html
+++ b/src/components/radio/radio_button.html
@@ -1,0 +1,24 @@
+<!-- TODO(jelbourn): render the radio on either side of the content -->
+<!-- TODO(mtlin): Evaluate trade-offs of using native radio vs. cost of additional bindings. -->
+<label [attr.for]="inputId" class="md-radio-label">
+  <!-- The actual `radio` part of the control. -->
+  <div class="md-radio-container">
+    <div class="md-radio-outer-circle"></div>
+    <div class="md-radio-inner-circle"></div>
+    <div class="md-ink-ripple"></div>
+  </div>
+
+  <input #input class="md-radio-input" type="radio"
+          [id]="inputId"
+          [checked]="checked"
+          [disabled]="disabled"
+          [name]="name"
+          (change)="onInputChange()"
+          (focus)="onInputFocus()"
+          (blur)="onInputBlur()" />
+
+  <!-- The label content for radio control. -->
+  <div class="md-radio-label-content">
+    <ng-content></ng-content>
+  </div>
+</label>
--- a/src/components/radio/radio_button.spec.ts
+++ b/src/components/radio/radio_button.spec.ts
@@ -1,0 +1,292 @@
+import {
+  fakeAsync,
+  inject,
+  tick,
+  TestComponentBuilder
+} from 'angular2/testing';
+import {
+  it,
+  describe,
+  expect,
+  beforeEach,
+} from '../../core/facade/testing';
+import {Component, DebugElement} from 'angular2/core';
+import {By} from 'angular2/platform/browser';
+
+import {MdRadioButton, MdRadioGroup, MdRadioChange} from './radio_button';
+import {MdRadioDispatcher} from './radio_dispatcher';
+
+export function main() {
+  describe('MdRadioButton', () => {
+    let builder: TestComponentBuilder;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+    }));
+
+    it('should have same name as radio group', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, `
+            <md-radio-group name="my_group">
+              <md-radio-button></md-radio-button>
+            </md-radio-group>`)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let button = fixture.debugElement.query(By.css('md-radio-button'));
+
+          fixture.detectChanges();
+          expect(button.componentInstance.name).toBe('my_group');
+        }).then(done);
+    });
+
+    it('should not allow click selection if disabled', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, '<md-radio-button disabled></md-radio-button>')
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let button = fixture.debugElement.query(By.css('md-radio-button'));
+
+          fixture.detectChanges();
+          expect(button.componentInstance.checked).toBe(false);
+
+          button.nativeElement.click();
+          expect(button.componentInstance.checked).toBe(false);
+        }).then(done);
+    });
+
+    it('should be disabled if radio group disabled', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, `
+            <md-radio-group disabled>
+              <md-radio-button></md-radio-button>
+            </md-radio-group>`)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let button = fixture.debugElement.query(By.css('md-radio-button'));
+
+          fixture.detectChanges();
+          expect(button.componentInstance.disabled).toBe(true);
+        }).then(done);
+    });
+
+    it('updates parent group value when selected and value changed', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, `
+            <md-radio-group>
+              <md-radio-button value="1"></md-radio-button>
+            </md-radio-group>`)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let button = fixture.debugElement.query(By.css('md-radio-button'));
+          let group = fixture.debugElement.query(By.css('md-radio-group'));
+
+          group.componentInstance.selected = button.componentInstance;
+          fixture.detectChanges();
+          expect(group.componentInstance.value).toBe('1');
+
+          button.componentInstance.value = '2';
+          fixture.detectChanges();
+          expect(group.componentInstance.value).toBe('2');
+        }).then(done);
+    });
+
+    it('should be checked after input change event', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, '<md-radio-button></md-radio-button>')
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let button = fixture.debugElement.query(By.css('md-radio-button'));
+          let input = button.query(By.css('input'));
+
+          fixture.detectChanges();
+          expect(button.componentInstance.checked).toBe(false);
+
+          let event = createEvent('change');
+          input.nativeElement.dispatchEvent(event);
+          expect(button.componentInstance.checked).toBe(true);
+        }).then(done);
+    });
+
+    it('should emit event when checked', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, '<md-radio-button></md-radio-button>')
+        .createAsync(TestApp)
+        .then((fixture) => {
+          fakeAsync(function() {
+            let button = fixture.debugElement.query(By.css('md-radio-button'));
+            let changeEvent: MdRadioChange = null;
+            button.componentInstance.change.subscribe((evt: MdRadioChange) => {
+              changeEvent = evt;
+            });
+            button.componentInstance.checked = true;
+            fixture.detectChanges();
+            tick();
+
+            expect(changeEvent).not.toBe(null);
+            expect(changeEvent.source).toBe(button.componentInstance);
+          });
+        }).then(done);
+    });
+
+    it('should be focusable', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, '<md-radio-button></md-radio-button>')
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let button = fixture.debugElement.query(By.css('md-radio-button'));
+          let input = button.query(By.css('input'));
+
+          fixture.detectChanges();
+          expect(button.nativeElement.classList.contains('md-radio-focused')).toBe(false);
+
+          let event = createEvent('focus');
+          input.nativeElement.dispatchEvent(event);
+          fixture.detectChanges();
+          expect(button.nativeElement.classList.contains('md-radio-focused')).toBe(true);
+
+          event = createEvent('blur');
+          input.nativeElement.dispatchEvent(event);
+          fixture.detectChanges();
+          expect(button.nativeElement.classList.contains('md-radio-focused')).toBe(false);
+        }).then(done);
+    });
+  });
+
+  describe('MdRadioDispatcher', () => {
+    let builder: TestComponentBuilder;
+    let dispatcher: MdRadioDispatcher;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+      dispatcher = new MdRadioDispatcher();
+    }));
+
+    it('notifies listeners', () => {
+      let notificationCount = 0;
+      const numListeners = 2;
+
+      for (let i = 0; i < numListeners; i++) {
+        dispatcher.listen(() => {
+          notificationCount++;
+        });
+      }
+
+      dispatcher.notify('hello');
+
+      expect(notificationCount).toBe(numListeners);
+    });
+  });
+
+  describe('MdRadioGroup', () => {
+    let builder: TestComponentBuilder;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+    }));
+
+    it('can select by value', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, `
+            <md-radio-group>
+              <md-radio-button value="1"></md-radio-button>
+              <md-radio-button value="2"></md-radio-button>
+            </md-radio-group>`)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let buttons = fixture.debugElement.queryAll(By.css('md-radio-button'));
+          let group = fixture.debugElement.query(By.css('md-radio-group'));
+
+          fixture.detectChanges();
+          expect(group.componentInstance.selected).toBe(null);
+
+          group.componentInstance.value = '2';
+
+          fixture.detectChanges();
+          expect(group.componentInstance.selected).toBe(buttons[1].componentInstance);
+        }).then(done);
+    });
+
+    it('should select uniquely', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, `
+            <md-radio-group>
+              <md-radio-button></md-radio-button>
+              <md-radio-button></md-radio-button>
+            </md-radio-group>`)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let buttons = fixture.debugElement.queryAll(By.css('md-radio-button'));
+          let group = fixture.debugElement.query(By.css('md-radio-group'));
+
+          fixture.detectChanges();
+          expect(group.componentInstance.selected).toBe(null);
+
+          group.componentInstance.selected = buttons[0].componentInstance;
+          fixture.detectChanges();
+          expect(isSinglySelected(buttons[0], buttons)).toBe(true);
+
+          group.componentInstance.selected = buttons[1].componentInstance;
+          fixture.detectChanges();
+          expect(isSinglySelected(buttons[1], buttons)).toBe(true);
+        }).then(done);
+    });
+
+    it('should emit event when value changes', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, `
+            <md-radio-group>
+              <md-radio-button></md-radio-button>
+              <md-radio-button></md-radio-button>
+            </md-radio-group>`)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          fakeAsync(function() {
+            let buttons = fixture.debugElement.queryAll(By.css('md-radio-button'));
+            let group = fixture.debugElement.query(By.css('md-radio-group'));
+
+            let changeEvent: MdRadioChange = null;
+            group.componentInstance.change.subscribe((evt: MdRadioChange) => {
+              changeEvent = evt;
+            });
+
+            group.componentInstance.selected = buttons[1].componentInstance;
+            fixture.detectChanges();
+            tick();
+
+            expect(changeEvent).not.toBe(null);
+            expect(changeEvent.source).toBe(buttons[1].componentInstance);
+          });
+        }).then(done);
+    });
+  });
+}
+
+/** Checks whether a given button is uniquely selected from a group of buttons. */
+function isSinglySelected(button: DebugElement, buttons: DebugElement[]): boolean {
+  let component = button.componentInstance;
+  let otherSelectedButtons =
+      buttons.filter((e: DebugElement) =>
+          e.componentInstance != component && e.componentInstance.checked);
+  return component.checked && otherSelectedButtons.length == 0;
+}
+
+/** Browser-agnostic function for creating an event. */
+function createEvent(name: string): Event {
+  let ev: Event;
+  try {
+    ev = createEvent(name);
+  } catch (e) {
+    ev = document.createEvent('Event');
+    ev.initEvent(name, true, true);
+  }
+  return ev;
+}
+
+
+/** Test component. */
+@Component({
+  directives: [MdRadioButton, MdRadioGroup],
+  providers: [MdRadioDispatcher],
+  template: ''
+})
+class TestApp {}
--- a/src/components/radio/radio_button.ts
+++ b/src/components/radio/radio_button.ts
@@ -1,0 +1,306 @@
+import {
+  AfterContentInit,
+  Component,
+  ContentChildren,
+  Directive,
+  EventEmitter,
+  HostBinding,
+  HostListener,
+  Input,
+  OnInit,
+  Optional,
+  Output,
+  QueryList,
+  ViewEncapsulation,
+  forwardRef
+} from 'angular2/core';
+
+import {Event} from 'angular2/src/facade/browser';
+
+import {MdRadioDispatcher} from './radio_dispatcher';
+
+// TODO(mtlin):
+// Ink ripple is currently placeholder.
+// Determine motion spec for button transitions.
+// Design review.
+// RTL
+// Support forms API.
+// Use ChangeDetectionStrategy.OnPush
+
+var _uniqueIdCounter = 0;
+
+/** A simple change event emitted by either MdRadioButton or MdRadioGroup. */
+export class MdRadioChange {
+  source: MdRadioButton;
+  value: any;
+}
+
+@Directive({
+  selector: 'md-radio-group',
+  host: {
+    'role': 'radiogroup',
+  },
+})
+export class MdRadioGroup implements AfterContentInit {
+  /** The value for the radio group. Should match currently selected button. */
+  private _value: any = null;
+
+  /** The HTML name attribute applied to radio buttons in this group. */
+  private _name: string = null;
+
+  /** Disables all individual radio buttons assigned to this group. */
+  private _disabled: boolean = false;
+
+  /** The currently selected radio button. Should match value. */
+  private _selected: MdRadioButton = null;
+
+  /** Event emitted when the group value changes. */
+  @Output()
+  change: EventEmitter<MdRadioChange> = new EventEmitter();
+
+  /** Child radio buttons. */
+  @ContentChildren(forwardRef(() => MdRadioButton))
+  private _radios: QueryList<MdRadioButton> = null;
+
+  /**
+   * Initialize properties once content children are available.
+   * This allows us to propagate relevant attributes to associated buttons.
+   */
+  ngAfterContentInit() {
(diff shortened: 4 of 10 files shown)


src/components/radio/radio-button.scss
```diff
@@ -0,0 +1,133 @@
...
+md-radio-button {
+  display: inline-block;
+  cursor: pointer;
+  padding: 8px;
+  position: relative;
+  white-space: nowrap;
+}
+
+md-radio-group {
+}
```
jelbourn: Empty definition?

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+  radios_: MdRadioButton[];
+
+  /** Disables all individual radio buttons assigned to this group. */
+  disabled_: boolean;
+
+  /** The currently selected radio button. Should match value. */
+  selected_: MdRadioButton;
+
+  /** Event emitted when the group value changes. */
+  @Output() change: EventEmitter<any>;
```
jelbourn: Instead of `any`, this should be a more specific type.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  constructor(@Attribute('name') name: string, @Attribute('disabled') disabled: boolean) {
+    // TODO(mtlin): Should we be able to dynamically change the group name?
+    if (isPresent(name)) {
```
jelbourn: I don't think you need need inject the name attribute here. Instead, you should be able to simple set the default name in `ngOnInit` if it's not set (the bound will be present at that point).

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+    } else {
+      this.name = `md-radio-group-${_uniqueIdCounter++}`;
+    }
+    this.radios_ = [];
+    this.change = new EventEmitter();
+    this.selected_ = null;
+    this.value_ = null;
+
+    // The simple presence of the `disabled` attribute dictates disabled state.
+    this.disabled_ = isPresent(disabled);
```
jelbourn: Rather than doing what you're currently doing for `disabled`, you can have a setter that looks like this:

```ts
    // The presence of *any* disabled value makes the component disabled, *except* for false.
    this.disabled_ = (value != null && value != false) ? true : null;
  }
```

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+
+
+@Component({
+  selector: 'md-radio-button',
+  templateUrl: './components/radio/radio_button.html',
+  styleUrls: ['./components/radio/radio-button.css'],
+  encapsulation: ViewEncapsulation.None
+})
+export class MdRadioButton {
+  /** Input field wrapped reference. */
```
jelbourn: Expand this comment to explain what the internal input is used for.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+      this.name = radioGroup.name;
+      this.radioGroup.register(this);
+    }
+  }
+
+  ngAfterViewInit() {
+    this.inputElement_ = this.inputRef_.nativeElement;
+    this.inputElement_.addEventListener('change', this.onChange.bind(this));
+    this.inputElement_.addEventListener('focus', this.onFocus.bind(this));
+    this.inputElement_.addEventListener('blur', this.onBlur.bind(this));
```
jelbourn: Instead of getting the raw input element and adding event listeners directly, it should just add events to the input in the template:
```html
<input type="radio" (change)="doSomething()" (focus)="doSomethingElse()">
```

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    this.radioGroup = radioGroup;
+    this.radioDispatcher = radioDispatcher;
+    this.value_ = null;
+    this.checked_ = false;
+    this.change = new EventEmitter();
+
+    this.id = isPresent(id) ? id : `md-radio-${_uniqueIdCounter++}`;
+
+    radioDispatcher.listen((name: string) => {
```
jelbourn: If you're driving the component with a real radio button, why do you still need the `radioDispatcher`?

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+  Input,
+  OnInit,
+  Optional,
+  Output,
+  SkipSelf,
+  ViewEncapsulation,
+  ViewChild
+} from 'angular2/core';
+
+import {isPresent, StringWrapper, NumberWrapper} from 'angular2/src/facade/lang';
```
jelbourn: -StringWrapper, -NumberWrapper

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+@Component({
+  selector: 'md-radio-group',
+  host: {
+    'role': 'radiogroup',
+  },
+  templateUrl: './components/radio/radio_group.html',
+  styleUrls: ['./components/radio/radio-group.css'],
+  encapsulation: ViewEncapsulation.None
+})
+export class MdRadioGroup implements OnInit {
```
jelbourn: Not strictly necessary for the first PR, but eventually it would need to support the forms api the same way the checkbox does it.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,269 @@
...
+  /** The parent radio group. May or may not be present. */
+  radioGroup: MdRadioGroup;
+
+  /** Dispatcher for coordinating radio unique-selection by name. */
+  radioDispatcher: MdRadioDispatcher;
+
+  /** Event emitted when the group value changes. */
+  @Output() change: EventEmitter<any>;
+
+  constructor(@Optional() @SkipSelf() @Host() radioGroup: MdRadioGroup, @Attribute('id') id: string,
```
jelbourn: You should be able to omit `@SkipSelf` and `@Host` now (they used to be necessary but shouldn't be any more).

src/components/radio/radio_button.ts
jelbourn: Anywhere you use `isPresent`, you should be able to just use `!= null` (not triple-equals). That should capture `null` for Dart and `null` / `undefined` for JS.

src/components/radio/radio_group.html
```diff
@@ -0,0 +1 @@
+<ng-content></ng-content>
```
hansl: This should be inlined in the component definition, using `template:` instead of `templateUrl:`.

jelbourn: Actually, if the template consists of nothing but `<ng-content>`, it should just be a `@Directive` instead of `@Component`.

src/demo-app/radio/radio-demo.scss
```diff
@@ -0,0 +1,10 @@
+button {
```
jelbourn: Can you prefix these styles with `demo-` ?

hansl:
Could you rebase? I have a feeling it will be green(er).

src/components/radio/radio_group.html
zoechi: @jelbourn sorry for the noise but I'm really curious how you could add `<ng-content>` with a `@Directive()` any example pointer? This then would be the same here https://github.com/angular/material2/blob/4009fb965dc5ce157c5bccaa6a879e29ef218712/src/components/sidenav/sidenav.ts#L49 ?

jelbourn: Yeah, that sidenav should also be a directive AFAICT (unless there's some css reason it's like that, @hansl?).

You use `@Directive` when there _is_ no template, so the content is just where it naturally resides in the tree.

src/components/radio/radio-button.scss
```diff
@@ -0,0 +1,133 @@
...
+  width: $radio-width;
+
+  .md-radio-checked & {
+    transform: scale(0.5);
+  }
+}
+
+.md-radio-label {
+  display: inline-block;
+  float: right;
```
rfru: I was planning to hold off on RTL until a later PR, since there might need to be more thought put into it.

src/components/radio/radio_button.ts
rfru: For now, how about we use `string` since that's the basic value for a native input?

src/components/radio/radio_button.ts
rfru: Ah, much better, thanks!

src/components/radio/radio_button.ts
rfru: Added a note in the radio dispatcher class. Unfortunately the change event only triggers on the to-be-selected element, so we'd have to programmatically uncheck the previous selection.

src/components/radio/radio_button.ts
rfru: Added TODO

rfru:
Ready for another round. Will get started with tests soon.

src/components/radio/radio-button.scss
```diff
@@ -0,0 +1,139 @@
...
+  height: $radio-width;
+  position: relative;
+  top: 2px;
+  width: $radio-width;
+}
+
+// TODO(mtlin): Replace when ink ripple component is implemented.
+// A placeholder ink ripple, shown when keyboard focused.
+.md-ink-ripple {
+  background: #000;
```
jelbourn: Should this be one of the theme background colors?

src/components/radio/radio-button.scss
```diff
@@ -0,0 +1,139 @@
...
+
+// The inner circle for the radio, shown when checked.
+.md-radio-inner-circle {
+  border-radius: 50%;
+  box-sizing: border-box;
+  height: $radio-width;
+  left: 0;
+  position: absolute;
+  top: 0;
+  transition: transform ease 0.28s, background-color ease 0.28s;
```
jelbourn: Should this be one of the common easing curves?

src/components/radio/radio-button.scss
```diff
@@ -0,0 +1,139 @@
+@import "default-theme";
+
+$radio-width: 20px !default;
```
jelbourn: Can you prefix the variables with `md-` ?

src/components/radio/radio-button.scss
```diff
@@ -0,0 +1,139 @@
...
+$radio-height: $radio-width !default;
+
+// Top-level container for entire button.
+md-radio-button {
+  display: inline-block;
+  cursor: pointer;
+  padding: 8px;
```
jelbourn: How possible is it to move things like this padding into variables that could be overridden by users in the name of density (which I know the designers have been thinking about but not yet really addressing in the spec)?

src/components/radio/radio_button.html
```diff
@@ -0,0 +1,22 @@
+<!-- TODO(jelbourn): render the radio on either side of the content -->
+
+<!-- The actual `radio` part of the control. -->
+<div class="md-radio-container">
+  <div class="md-radio-outer-circle"></div>
+  <div class="md-radio-inner-circle"></div>
+  <div class="md-ink-ripple"></div>
+</div>
+
+<input #input class="md-radio-input" type="radio"
```
jelbourn: Can you add a TODO that's something like
"Evaluate trade-offs of using native radio vs. the cost of additional bindings"

src/components/radio/radio_button.ts
jelbourn: I'd say just make an empty `MdRadioChange` class for now and we can figure out what to do with it later.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+
+@Directive({
+  selector: 'md-radio-group',
+  host: {
+    'role': 'radiogroup',
+  },
+})
+export class MdRadioGroup implements AfterContentInit {
+  /** The value for the radio group. Should match currently selected button. */
+  value_: string;
```
jelbourn: Missing `private` keyword for the private members

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  /** Event emitted when the group value changes. */
+  @Output() change: EventEmitter<any>;
+
+  /** Child radio buttons. */
+  @ContentChildren(forwardRef(() => MdRadioButton))
+  private radios_: QueryList<MdRadioButton>;
+
+  constructor() {
+    this.name_ = null;
```
jelbourn: FYI: you can also initialize these in their declarations. E.g.,
```ts
private name_: string = null;
```

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+    this.radios_ = null;
+    this.change = new EventEmitter();
+    this.selected_ = null;
+    this.value_ = null;
+    this.disabled_ = false;
+  }
+
+  ngAfterContentInit() {
```
jelbourn: Add a comment explaining what you need to do specifically after content init.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+    }
+    // Trigger attribute propagation once child buttons are ready.
+    this.name = this.name;
+
+    if (this.disabled_) {
+      this.disabled = true;
+    }
+  }
+
+  @Input('name')
```
jelbourn: You can put this on the `name_` field at the top:

```ts
@Input('name') private name: string = null;
```

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+    this.isFocused_ = true;
+  }
+
+  private onBlur() {
+    this.isFocused_ = false;
+  }
+
+  /** Input change handler, called only on keyboard selection. */
+  private onChange(ev: Event) {
+    this.checked = (<HTMLInputElement>ev.target).checked;
```
jelbourn: Since radio buttons only call `change` when they're selected, could you not just say
```ts
this.checked = true;
```
?

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Input field wrapped reference.
+   * We use a hidden native input field to handle changes to focus state via keyboard navigation,
+   * with visual rendering done separately. The native element is kept in sync with the overall
+   * state of the component.
+   */
+  @ViewChild('input') private inputRef_: ElementRef;
+
+  /** Stores the raw input element. */
+  private inputElement_: HTMLInputElement;
```
jelbourn: I don't think you need inputRef or inputElement

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  /** Dispatcher for coordinating radio unique-selection by name. */
+  radioDispatcher: MdRadioDispatcher;
+
+  /** Event emitted when the group value changes. */
+  @Output() change: EventEmitter<any>;
+
+  constructor(@Optional() radioGroup: MdRadioGroup, @Attribute('id') id: string,
+              radioDispatcher: MdRadioDispatcher) {
+    // Assertions. Ideally these should be stripped out by the compiler.
```
jelbourn: I don't think there are assertions here any more.

src/components/radio/radio_dispatcher.ts
```diff
@@ -0,0 +1,27 @@
...
+    this.listeners_ = [];
+  }
+
+  /** Notify other radio buttons that selection for the given name has been set. */
+  notify(name: string) {
+    this.listeners_.forEach(listener => listener(name));
+  }
+
+  /** Listen for future changes to radio button selection. */
+  listen(listener: Function) {
```
jelbourn: I _think_ this signature can be made more specific:
```ts
listen(listener: (name: string) => void) { ... }
```
(though I'm not 100% sure if that will transpile to dart)

src/demo-app/radio/radio-demo.ts
```diff
@@ -0,0 +1,14 @@
+import {Component} from 'angular2/core';
+import {MdRadioButton, MdRadioGroup} from '../../components/radio/radio_button';
+import {MdRadioDispatcher} from '../../components/radio/radio_dispatcher';
+
+@Component({
+    selector: 'radio-demo',
+    templateUrl: 'demo-app/radio/radio-demo.html',
+    styleUrls: ['demo-app/radio/radio-demo.css'],
+    providers: [MdRadioDispatcher],
+    directives: [MdRadioButton, MdRadioGroup]
```
jelbourn: nit: too much indent

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,270 @@
...
+// RTL
+// Support forms API.
+
+var _uniqueIdCounter: number = 0;
+
+@Directive({
+  selector: 'md-radio-group',
+  host: {
+    'role': 'radiogroup',
+  },
```
jelbourn: I forgot to mention this to you and @traviskaufman earlier, and it can be a TODO for a follow-up PR, but we should try to use the configuration
 ```ts
```

This tells Angular to only perform change detection when one of the `@Input` properties is changed or when there is a local event. For checkbox and radio button it _should_ "just work" since all of the bound values are primitives.

File: src/components/radio/radio_dispatcher.ts [hunk 29]
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Class for radio buttons to coordinate unique selection based on name.
+ * Indended to be consumed as an Angular service.
+ * This service is needed because native radio change events are only fired on the item currently
+ * being selected, and we still need to uncheck the previous selection.
+ */
+@Injectable()
+export class MdRadioDispatcher {
+  // TODO(jelbourn): Change this to TypeScript syntax when supported.
+  listeners_: Function[];
```
jelbourn: Add `private`

jelbourn:
Are you planning on doing the keyboard interaction for the radio-group in a follow-up PR?

hansl:
You need to `git reset --hard` your master to the latest `upstream/master`. You still show up the merge commit from 

If you need help doing that just ask on slack and we'll be happy to assist you :)

src/components/radio/radio-button.scss
rfru: Done.

src/components/radio/radio-button.scss
rfru: Possible, though since we don't have any systematic guidance of that kind across components, my imagination is that users should just override the styles using their own cascaded styles.

src/components/radio/radio_button.ts
rfru: Added TODO

src/components/radio/radio_button.ts
rfru: Good catch.

src/components/radio/radio_button.ts
rfru: Yep!

src/components/radio/radio_button.ts
rfru: Is there a preference for single line vs multiple line declarations?

src/components/radio/radio-button.scss
rfru: Potentially, but I am holding off on any design/motion tweaks because I need to get those reviewed by the team first.

rfru:
@jelbourn re: keyboard interaction - It should be working already (benefit of native inputs), try it out and let me know if it's not.

@hansl It's up-to-date now. Let me know if you see anything wrong.

rfru:
Sorry for delay, have been tied up with other stuff. Comments addressed. Will work on tests ASAP.

src/components/radio/radio-button.scss
jelbourn: I don't see any change here.

rfru: This one's below, in the theme section.

src/components/radio/radio-button.scss
```diff
@@ -0,0 +1,138 @@
...
+  height: 0;
+  margin: 0;
+  padding: 0;
+  opacity: 0;
+  appearance: none;
+  border: none;
+}
+
+
+// THEME
```
jelbourn: The theme stuff doesn't have to be separated from the rest of the css; this was an artifact left over from when I copied and pasted the styles from the ng1 version (where they are separated).

src/components/radio/radio_button.html
```diff
@@ -0,0 +1,22 @@
...
+        [checked]="checked"
+        [disabled]="disabled"
+        [name]="name"
+        (change)="onInputChange($event)"
+        (focus)="onInputFocus()"
+        (blur)="onInputBlur()" />
+
+<!-- The label for radio control. -->
+<div class="md-radio-label">
```
jelbourn: The text should label the input, either using a `<label>` element or with `aria-labelledby`

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,254 @@
...
+})
+export class MdRadioGroup implements AfterContentInit {
+  /** The value for the radio group. Should match currently selected button. */
+  @Input('value')
+  private _value: MdRadioChange = null;
```
jelbourn: Rather than put the `@Input('name')` here, you can instead put just `@Input()` on the getter.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Initialize properties once content children are available.
+   * This allows us to propagate relevant attributes to associated buttons.
+   */
+  ngAfterContentInit() {
+    if (this._name == null) {
+      this.name = `md-radio-group-${_uniqueIdCounter++}`;
+    }
+    // Trigger `name` attribute propagation.
+    this.name = this.name;
```
jelbourn: I don't follow why this is necessary. If you're relying on this forcing the setter to be called, there should instead be a separate function that's called both here and in the setter.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,254 @@
...
+  get name(): string {
+    return this._name;
+  }
+
+  set name(value: string) {
+    this._name = value;
+
+    // Propagate name attribute to radio buttons.
+    if (this._radios != null) {
+      iterateListLike(this._radios, (radio: MdRadioButton) => {
```
jelbourn: `QueryList` _should_ now have a `forEach` method.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,254 @@
...
+    this._disabled = (value != null && value !== false) ? true : null;
+  }
+
+  get value(): MdRadioChange {
+    return this._value;
+  }
+
+  set value(value: MdRadioChange) {
+    if (this._value != value) {
+      ObservableWrapper.callEmit(this.change, null);
```
jelbourn: You should be able to just do `this.change.emit(null)` now.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,254 @@
...
+  get disabled(): boolean {
+    return this._disabled;
+  }
+
+  set disabled(value) {
+    // The presence of *any* disabled value makes the component disabled, *except* for false.
+    this._disabled = (value != null && value !== false) ? true : null;
+  }
+
+  get value(): MdRadioChange {
```
jelbourn: I don't think this is quite right, conceptually. The type of `value` should be `any`, and then the `MdRadioChage` is used as an event when the `change` event is fired, where the `MdRadioChange` has additional information about the change in it (the value, the radio button instance, etc.).

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,254 @@
...
+
+
+@Component({
+  selector: 'md-radio-button',
+  templateUrl: './components/radio/radio_button.html',
+  styleUrls: ['./components/radio/radio-button.css'],
+  encapsulation: ViewEncapsulation.None
+})
+export class MdRadioButton {
+  @HostBinding('class.md-radio-focused') private _isFocused: boolean;
```
jelbourn: It's inconsistent whether the decorator is on the next / same line.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,254 @@
...
+  onInputFocus() {
+    this._isFocused = true;
+  }
+
+  onInputBlur() {
+    this._isFocused = false;
+  }
+
+  /** Input change handler, called only on keyboard selection. */
+  onInputChange(ev: Event) {
```
jelbourn: Why does this need the event as an argument?

src/components/radio/radio_button.ts
rfru: Yeah, you're right, sorry about that. Fixed this up.

rfru:
Resolved comments and added tests. Ready for another look.

src/components/radio/radio_button.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+        .overrideTemplate(TestApp, '<md-radio-button></md-radio-button>')
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let button = getChildDebugElement(fixture.debugElement, 'md-radio-button');
+          let input = getChildDebugElement(button, 'input');
+
+          fixture.detectChanges();
+          expect(button.componentInstance.checked).toBe(false);
+
+          let event = new Event('change');
```
jelbourn: Doesn't `new Event` not work in IE11?

src/components/radio/radio_button.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+          input.nativeElement.dispatchEvent(event);
+          expect(button.componentInstance.checked).toBe(true);
+        }).then(done);
+    });
+
+    it('should be keyboard focusable', (done: () => void) => {
```
jelbourn: I don't follow how this test checks that its _keyboard_ focusable

src/components/radio/radio_button.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+  describe('MdRadioDispatcher', () => {
+    let builder: TestComponentBuilder;
+    let dispatcher: MdRadioDispatcher;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+      dispatcher = new MdRadioDispatcher();
+    }));
+
+    it('notifies listeners', (done: () => void) => {
```
jelbourn: If the test isn't async, you shouldn't need to pass in / call `done`

src/components/radio/radio_button.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+  return component.checked && otherSelectedButtons.length == 0;
+}
+
+/** Gets a child DebugElement by tag name. */
+function getChildDebugElement(parent: DebugElement, selector: string): DebugElement {
+  return parent.query(By.css(selector));
+}
+
+/** Gets an array of child DebugElements by tag name. */
+function getChildDebugElements(parent: DebugElement, selector: string): DebugElement[] {
```
jelbourn: You don't need this function any more; sayng `parent.queryAll(By.css(...))` is shorter than calling this function. It was left-over from a time when the `By.css` API didn't exist and you had to give a full predicate function.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,298 @@
...
+   * This allows us to propagate relevant attributes to associated buttons.
+   */
+  ngAfterContentInit() {
+    if (this._name == null) {
+      this.name = `md-radio-group-${_uniqueIdCounter++}`;
+    } else {
+      this._updateChildRadioNames();
+    }
+
+    if (this._disabled) {
```
jelbourn: I'm confused why this is necessary

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,298 @@
...
+  set name(value: string) {
+    this._name = value;
+
+    this._updateChildRadioNames();
+  }
+
+  /** Propagate name attribute to radio buttons. */
+  private _updateChildRadioNames() {
+    if (this._radios != null) {
+      this._radios.forEach((radio: MdRadioButton) => {
```
jelbourn: TypeScript _should_ know the type of `radio` implicitly since `QueryList` is generic; does it work if you just do
```ts
this._radios.forEach(radio => {
  radio.name = this._name;
});
```
?

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,298 @@
...
+    // The presence of *any* disabled value makes the component disabled, *except* for false.
+    this._disabled = (value != null && value !== false) ? true : null;
+  }
+
+  @Input()
+  get value(): any {
+    return this._value;
+  }
+
+  set value(value: any) {
```
jelbourn: Rename the parameter `newValue` to disambiguate?

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        // Change the selection immediately.
+        this.selected = matched[0];
+      }
+
+      let event = new MdRadioChange();
+      event.source = this._selected;
+      event.value = value;
+      this.change.emit(event);
+    } else {
```
jelbourn: Isn't this `else` a no-op?

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,298 @@
...
+    this._disabled = (value != null && value !== false) ? true : null;
+  }
+
+  @Input()
+  get value(): any {
+    return this._value;
+  }
+
+  set value(value: any) {
+    if (this._value != value) {
```
jelbourn: This setter is pretty complicated; could you break it up into a couple of pieces? I'm thinking something like:
```ts
set value(newValue: any) {
  if (this._value != newValue) {
    this._value = newValue;
    this._updateSelectedRadioFromValue();
    this._emitChangeEvent();
  }
}
```

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  constructor(@Optional() radioGroup: MdRadioGroup, @Attribute('id') id: string,
+              radioDispatcher: MdRadioDispatcher) {
+    // Assertions. Ideally these should be stripped out by the compiler.
+    // TODO(jelbourn): Assert that there's no name binding AND a parent radio group.
+
+    this.radioGroup = radioGroup;
+    this.radioDispatcher = radioDispatcher;
+
+    this.id = id != null ? id : `md-radio-${_uniqueIdCounter++}`;
```
jelbourn: You could set `this.id` in `ngOnInit` and then not have to inject `@Attribute('id')` in tbe constructor.

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  /** MdRadioGroup reads this to assign its own value. */
+  @Input()
+  get value(): any {
+    return this._value;
+  }
+
+  set value(value: any) {
+    if (this._value != value) {
+      let event = new MdRadioChange();
```
jelbourn: I'm not sure the radio _button_ (vs. group) should fire an event when its `value` changes. Did you have something in mind for that case?

jelbourn: Actually, now that I think about it, the use case for this is the parent radio group listening to this event to update its own value. Can this go into it's own function like `_emitValueChangeEvent` ?

src/components/radio/radio_button.spec.ts
rfru: Added a helper function for this.

src/components/radio/radio_button.spec.ts
rfru: Removed "keyboard" word to match expectation.

src/components/radio/radio_button.ts
rfru: Ah, sorry, this was when disabled actually did something besides setting value. Removed.

src/components/radio/radio_button.ts
rfru: Yeah, removed.

src/components/radio/radio_button.ts
rfru: Thanks, good suggestion.

src/components/radio/radio_button.ts
rfru: Sorry, don't know what I was thinking here, got this confused with checked.
I moved the code into checked, which matches the native input change event behavior (emit change only when selected).

rfru:
Ready for another look.
I had to fix a focus handling bug when introducing the label element earlier, so I've re-arranged the DOM slightly.

jelbourn:
LGTM

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,310 @@
...
+  Input,
+  OnInit,
+  Optional,
+  Output,
+  QueryList,
+  ViewEncapsulation,
+  forwardRef
+} from 'angular2/core';
+
+import {EventEmitter} from 'angular2/src/facade/async';
```
kara: I think you should be able to get this from 'angular2/core' as well

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,310 @@
...
+/** A simple change event emitted by either MdRadioButton or MdRadioGroup. */
+export class MdRadioChange {
+  source: MdRadioButton;
+  value: any;
+}
+
+@Directive({
+  selector: 'md-radio-group',
+  host: {
+    'role': 'radiogroup',
```
kara: Convert this to @HostBinding to be consistent?

jelbourn: This one you don't want as a `HostBinding` because it's static

src/components/radio/radio_dispatcher.ts
```diff
@@ -0,0 +1,27 @@
+import {Injectable} from 'angular2/core';
+
+/**
+ * Class for radio buttons to coordinate unique selection based on name.
+ * Indended to be consumed as an Angular service.
```
kara: Typo: "Intended"

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,310 @@
...
+    this._isFocused = false;
+  }
+
+  /** Input change handler, called only on keyboard selection. */
+  onInputChange() {
+    this.checked = true;
+  }
+
+  get inputId(): string {
+    return this.id + '-input';
```
kara: Use string interpolation here?

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+
+  @Input()
+  get disabled(): boolean {
+    return this._disabled;
+  }
+
+  set disabled(value) {
+    // The presence of *any* disabled value makes the component disabled, *except* for false.
```
hansl: @jelbourn: Just a FYI that I'm adding BooleanFieldValue as a decorator to do exactly this: https://github.com/angular/material2/pull/161/files#diff-6ef16658b3712f44bd8299cc8b9c9584R20

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  set name(value: string) {
+    this._name = value;
+
+    this._updateChildRadioNames();
+  }
+
+  /** Propagate name attribute to radio buttons. */
+  private _updateChildRadioNames() {
```
kara: Add the return type - `void`, it looks like

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,310 @@
...
+  /** Dispatch change event with current selection and group value. */
+  private _emitChangeEvent() {
+    let event = new MdRadioChange();
+    event.source = this._selected;
+    event.value = this._value;
+    this.change.emit(event);
+  }
+
+  @Input()
+  get selected() {
```
kara: Return type

src/components/radio/radio_button.ts
```diff
@@ -0,0 +1,310 @@
...
+  /** Event emitted when the group value changes. */
+  @Output()
+  change: EventEmitter<MdRadioChange> = new EventEmitter();
+
+  constructor(@Optional() radioGroup: MdRadioGroup, radioDispatcher: MdRadioDispatcher) {
+    // Assertions. Ideally these should be stripped out by the compiler.
+    // TODO(jelbourn): Assert that there's no name binding AND a parent radio group.
+
+    this.radioGroup = radioGroup;
+    this.radioDispatcher = radioDispatcher;
```
kara: Can remove this line and line 194 if you add "public" to the constructor, i.e

```javascript
constructor(@Optional() radioGroup: MdRadioGroup, public radioDispatcher: MdRadioDispatcher)
```

src/components/radio/radio_button.ts
kara: Gotcha

rfru:
Updated with changes from @kara 's comments.
I seem to be getting a timeout error in the build from one of checkbox's tests?

jelbourn:
Those are just flakes on Edge

Issue #40, opened 2016-01-26
mdProgressBar


jelbourn:
- [ ] Determinate mode
- [ ] Indeterminate mode
- [ ] Buffer mode
- [ ] Query mode
- [ ] Unit tests
- [ ] E2E tests
- [ ] Documentation
- [ ] Demo page

PR #130, opened 2016-03-04
feat(md-progress-bar): Create initial ProgressBar implementation.

josephperrott:
For #40

### DIFF
--- a/src/components/progress-bar/progress-bar.html
+++ b/src/components/progress-bar/progress-bar.html
@@ -1,0 +1,5 @@
+<!-- The background div is named as such because it appears below the other divs and is not sized based on values. -->
+<div class="md-progress-bar-background"></div>
+<div class="md-progress-bar-buffer" [style.transform]="bufferTransform()"></div>
+<div class="md-progress-bar-primary md-progress-bar-fill" [style.transform]="primaryTransform()"></div>
+<div class="md-progress-bar-secondary md-progress-bar-fill"></div>
--- a/src/components/progress-bar/progress-bar.scss
+++ b/src/components/progress-bar/progress-bar.scss
@@ -1,0 +1,297 @@
+@import "variables";
+@import "default-theme";
+
+$md-progress-bar-height: 5px !default;
+$md-progress-bar-full-animation-duration: 2s !default;
+$md-progress-bar-piece-animation-duration: 250ms !default;
+
+// TODO(josephperrott): Find better way to inline svgs.
+/** In buffer mode a repeated SVG object is used as a background.  Each of the following defines the SVG object for each
+    of the class defined colors.
+
+    Each string is a URL encoded version of:
+
+    <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
+         version="1.1" x="0px" y="0px" enable-background="new 0 0 5 2"
+         xml:space="preserve" viewBox="0 0 5 2" preserveAspectRatio="none slice">
+      <circle cx="1" cy="1" r="1" fill="{INJECTED_COLOR}"/>
+    </svg>
+
+    */
+$md-buffer-bubbles-primary: (
+  "data:image/svg+xml;charset=UTF-8,%3Csvg%20version%3D%271.1%27%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27" +
+  "%20xmlns%3Axlink%3D%27http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%27%20x%3D%270px%27%20y%3D%270px%27%20enable-backgroun" +
+  "d%3D%27new%200%200%205%202%27%20xml%3Aspace%3D%27preserve%27%20viewBox%3D%270%200%205%202%27%20preserveAspectRatio" +
+  "%3D%27none%20slice%27%3E%3Ccircle%20cx%3D%271%27%20cy%3D%271%27%20r%3D%271%27%20fill%3D%27" +
+  md-color($md-primary, 100) + "%27%2F%3E%3C%2Fsvg%3E") !default;
+$md-buffer-bubbles-accent: (
+  "data:image/svg+xml;charset=UTF-8,%3Csvg%20version%3D%271.1%27%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27" +
+  "%20xmlns%3Axlink%3D%27http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%27%20x%3D%270px%27%20y%3D%270px%27%20enable-backgroun" +
+  "d%3D%27new%200%200%205%202%27%20xml%3Aspace%3D%27preserve%27%20viewBox%3D%270%200%205%202%27%20preserveAspectRatio" +
+  "%3D%27none%20slice%27%3E%3Ccircle%20cx%3D%271%27%20cy%3D%271%27%20r%3D%271%27%20fill%3D%27" +
+  md-color($md-accent, 100) + "%27%2F%3E%3C%2Fsvg%3E") !default;
+$md-buffer-bubbles-warn: (
+  "data:image/svg+xml;charset=UTF-8,%3Csvg%20version%3D%271.1%27%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27" +
+  "%20xmlns%3Axlink%3D%27http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%27%20x%3D%270px%27%20y%3D%270px%27%20enable-backgroun" +
+  "d%3D%27new%200%200%205%202%27%20xml%3Aspace%3D%27preserve%27%20viewBox%3D%270%200%205%202%27%20preserveAspectRatio" +
+  "%3D%27none%20slice%27%3E%3Ccircle%20cx%3D%271%27%20cy%3D%271%27%20r%3D%271%27%20fill%3D%27" +
+  md-color($md-warn, 100) + "%27%2F%3E%3C%2Fsvg%3E") !default;
+
+:host {
+  display: block;
+  // Height is provided for md-progress-bar to act as a default.
+  height: $md-progress-bar-height;
+  overflow: hidden;
+  position: relative;
+  // translateZ is added to force the md-progress-bar into its own GPU layer.
+  transform: translateZ(0);
+  transition: opacity $md-progress-bar-piece-animation-duration linear;
+  width: 100%;
+
+  // The progress bar background is used to show the bubble animation scrolling behind a buffering progress bar.
+  .md-progress-bar-background {
+    background: url($md-buffer-bubbles-primary);
+    background-repeat: repeat-x;
+    background-size: 10px 4px;
+    height: 100%;
+    position: absolute;
+    visibility: hidden;
+    width: 100%;
+  }
+
+  /**
+   * The progress bar buffer is the bar indicator showing the buffer value and is only visible beyond the current value
+   * of the primary progress bar.
+   */
+  .md-progress-bar-buffer {
+    background-color: md-color($md-primary, 100);
+    height: 100%;
+    position: absolute;
+    transform-origin: top left;
+    transition: transform $md-progress-bar-piece-animation-duration ease;
+    width: 100%;
+  }
+
+  /**
+   * The secondary progress bar is only used in the indeterminate animation, because of this it is hidden in other uses.
+   */
+  .md-progress-bar-secondary {
+    visibility: hidden;
+  }
+
+  /**
+  * The progress bar fill fills the progress bar with the indicator color.
+  */
+  .md-progress-bar-fill {
+    animation: none;
+    height: 100%;
+    position: absolute;
+    transform-origin: top left;
+    transition: transform $md-progress-bar-piece-animation-duration ease;
+    width: 100%;
+  }
+
+  /**
+  * A pseudo element is created for each progress bar bar that fills with the indicator color.
+  */
+  .md-progress-bar-fill::after {
+    animation: none;
+    background-color: md-color($md-primary, 600);
+    content: '';
+    display: inline-block;
+    height: 100%;
+    position: absolute;
+    width: 100%;
+  }
+
+  &[color="accent"] {
+    .md-progress-bar-background {
+      background: url($md-buffer-bubbles-accent);
+      background-repeat: repeat-x;
+      background-size: 10px 4px;
+    }
+    .md-progress-bar-buffer {
+      background-color: md-color($md-accent, 100);
+    }
+    .md-progress-bar-fill::after {
+      background-color: md-color($md-accent, 600);
+    }
+  }
+
+  &[color="warn"] {
+    .md-progress-bar-background {
+      background: url($md-buffer-bubbles-warn);
+      background-repeat: repeat-x;
+      background-size: 10px 4px;
+    }
+    .md-progress-bar-buffer {
+      background-color: md-color($md-warn, 100);
+    }
+    .md-progress-bar-fill::after {
+      background-color: md-color($md-warn, 600);
+    }
+  }
+
+  &[mode="query"] {
+    transform: rotateZ(180deg);
+  }
+
+  &[mode="indeterminate"],
+  &[mode="query"] {
+    .md-progress-bar-fill {
+      transition: none;
+    }
+    .md-progress-bar-primary {
+      animation: md-progress-bar-primary-indeterminate-translate $md-progress-bar-full-animation-duration infinite linear;
+      left: -145.166611%;
+    }
+    .md-progress-bar-primary.md-progress-bar-fill::after {
+      animation: md-progress-bar-primary-indeterminate-scale $md-progress-bar-full-animation-duration infinite linear;
+    }
+    .md-progress-bar-secondary {
+      animation: md-progress-bar-secondary-indeterminate-translate $md-progress-bar-full-animation-duration infinite linear;
+      left: -54.888891%;
+      visibility: visible;
+    }
+    .md-progress-bar-secondary.md-progress-bar-fill::after {
+      animation: md-progress-bar-secondary-indeterminate-scale $md-progress-bar-full-animation-duration infinite linear;
+    }
+  }
+
+  &[mode="buffer"] {
+    .md-progress-bar-background {
+      animation: md-progress-bar-background-scroll $md-progress-bar-piece-animation-duration infinite linear;
+      visibility: visible;
+    }
+  }
+}
+
+
+// Reverse the apparent directionality of progress vars for rtl.
+:host-context([dir="rtl"]) {
+  transform: rotateY(180deg);
+}
+
+
+/** The values used for animations in md-progress-bar, both timing and transformation, can be considered magic values.
+    They are sourced from the Material Design example spec and duplicate the values of the original designers
+    definitions.
+
+
+    The indeterminate state is essentially made up of two progress bars, one primary (the one that is shown in both the
+    determinate and indeterminate states) and one secondary, which essentially mirrors the primary progress bar in
+    appearance but is only shown to assist with the indeterminate animations.
+
+
+    KEYFRAME BLOCK	                    DESCRIPTION
+    primary-indeterminate-translate     Translation of the primary progressbar across the screen
+    primary-indeterminate-scale         Scaling of the primary progressbar as it's being translated across the screen
+    secondary-indeterminate-translate   Translation of the secondary progressbar across the screen
+    secondary-indeterminate-scale       Scaling of the secondary progressbar as it's being translated across the screen
+
+    Because two different transform animations need to be applied at once, the translation is applied to the outer
+    element and the scaling is applied to the inner element, which provides the illusion necessary to make the animation
+    work.
+*/
+
+// Progress Bar Timing functions:
+// $md-progress-bar-primary-indeterminate-translate-step-1 has no timing function.
+$md-progress-bar-primary-indeterminate-translate-step-2: cubic-bezier(.5, 0, .701732, .495819) !default;
+$md-progress-bar-primary-indeterminate-translate-step-3: cubic-bezier(.302435, .381352, .55, .956352) !default;
+// $md-progress-bar-primary-indeterminate-translate-step-4 has no timing function.
+
+// $md-progress-bar-primary-indeterminate-scale-step-1 has no timing function
+$md-progress-bar-primary-indeterminate-scale-step-2: cubic-bezier(.334731, .124820, .785844, 1) !default;
+$md-progress-bar-primary-indeterminate-scale-step-3: cubic-bezier(.06, .11, .6, 1) !default;
+// $md-progress-bar-primary-indeterminate-scale-step-4 has no timing function
+
+$md-progress-bar-secondary-indeterminate-translate-step-1: cubic-bezier(.15, 0, .515058, .409685) !default;
+$md-progress-bar-secondary-indeterminate-translate-step-2: cubic-bezier(.310330, .284058, .8, .733712) !default;
+$md-progress-bar-secondary-indeterminate-translate-step-3: cubic-bezier(.4, .627035, .6, .902026) !default;
+// $md-progress-bar-secondary-indeterminate-translate-step-4 has no timing function
+
+$md-progress-bar-secondary-indeterminate-scale-step-1: cubic-bezier(.15, 0, .515058, .409685) !default;
+$md-progress-bar-secondary-indeterminate-scale-step-2: cubic-bezier(.310330, .284058, .8, .733712) !default;
+$md-progress-bar-secondary-indeterminate-scale-step-3: cubic-bezier(.4, .627035, .6, .902026) !default;
+// $md-progress-bar-secondary-indeterminate-scale-step-4 has no timing function
+
+/** Animations for indeterminate and query mode. */
+// Primary indicator.
+@keyframes md-progress-bar-primary-indeterminate-translate {
+  0% {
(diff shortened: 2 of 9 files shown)


josephperrott:
I signed it!

docs/dart/writing-compatible-typescript.md
```diff
@@ -43,3 +43,6 @@ Here's a list of gotchas to keep in mind when writing TypeScript code that will
 * **Lambdas need to abide to the type required.** Meaning that if a function requires a function that takes one argument, the lambda cannot be `() => {}`. Use `_` for temporary parameters. This is notable in Promises.
 * **Dynamic return values fat arrows can't return non-primitive types without casting.** For example, a fat arrow that return a promise (or Future) generates a warning if using directly, like so: `promiseA.then(() => promiseB)`. In this case simply using a block works as expected; `promiseA.then(() => { return promiseB; })`.
 * **Dart does not have [generic methods](http://www.typescriptlang.org/Handbook#generics)** ([issue](https://github.com/dart-lang/sdk/issues/254)).
+* **Dart does not have destructuring** but it does have a [proposal](https://github.com/dart-lang/dart_enhancement_proposals/issues/24) for it.
+* **Dart does not support type aliases.** Code like `type Nop = () => void` will not compile.
+* **Dart does not support [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals)**.
```
jelbourn: Looks like you need to rebase again, since these changes came from an earlier commit.

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,205 @@
+@import "variables";
```
jelbourn: Prefix classes and variables with `md-`.

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,205 @@
...
+      <circle cx="1" cy="1" r="1" fill="{INJECTED_COLOR}"/>
+    </svg>
+
+    */
+$buffer-bubbles-primary: (
+  "data:image/svg+xml;charset=UTF-8,%3Csvg%20version%3D%271.1%27%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27" +
+  "%20xmlns%3Axlink%3D%27http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%27%20x%3D%270px%27%20y%3D%270px%27%20enable-backgroun" +
+  "d%3D%27new%200%200%205%202%27%20xml%3Aspace%3D%27preserve%27%20viewBox%3D%270%200%205%202%27%20preserveAspectRatio" +
+  "%3D%27none%20slice%27%3E%3Ccircle%20cx%3D%271%27%20cy%3D%271%27%20r%3D%271%27%20fill%3D%27" +
+  md-color($md-primary, 100) + "%27%2F%3E%3C%2Fsvg%3E") !default;
```
jelbourn: You shouldn't need to encode SVGs in order for them to act as a background image; in fact, it can [make them rather larger](https://css-tricks.com/probably-dont-base64-svg/). You could also avoid repeating the entire string by having a scss function that takes the fill color and returns the result.

However, I wonder if there's another way of doing this, especially because the bulk of the SVG is repeated. Would it be possible to put the SVG in the template to accomplish the same thing (which would let you style the `fill` with css)?
cc @traviskaufman @hansl @robertmesserle @kara for ideas

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,205 @@
...
+/** Animations for indeterminate and query mode. */
+// Primary indicator.
+@keyframes primary-indeterminate-translate {
+  0%      { transform: translateX(0px); }
+  20%     { animation-timing-function: $progress-bar-primary-indeterminate-translate-step-2;
+            transform: translateX(0px); }
+  59.15%  { animation-timing-function: $progress-bar-primary-indeterminate-translate-step-3;
+            transform: translateX(83.67142%); }
+  100%    { transform: translateX(200.611057%); }
+}
```
jelbourn: If all of the steps fit on one line this style would be good, but since some of them carry-over, I'd do it like:
```scss
0% {
  transform: ...;
}
20% {
  transform: ...;
  ...;
}
```

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,205 @@
...
+
+/** Animations for indeterminate and query mode. */
+// Primary indicator.
+@keyframes primary-indeterminate-translate {
+  0%      { transform: translateX(0px); }
+  20%     { animation-timing-function: $progress-bar-primary-indeterminate-translate-step-2;
+            transform: translateX(0px); }
+  59.15%  { animation-timing-function: $progress-bar-primary-indeterminate-translate-step-3;
+            transform: translateX(83.67142%); }
+  100%    { transform: translateX(200.611057%); }
```
jelbourn: Add comments explaining where these specific transform values come from?

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,205 @@
...
+  /** Height is provided for md-progress-bar to act as a default. */
+  height: $progress-bar-height;
+  overflow: hidden;
+  position: relative;
+  /** translateZ is added to force the md-progress-bar into its own GPU layer. */
+  transform: translateZ(0);
+  transition: opacity $progress-bar-piece-animation-duration linear;
+  width: 100%;
+
+  .background {
```
jelbourn: Can these styles not be nested under `:host`?

src/components/progress-bar/progress-bar.spec.ts
```diff
@@ -0,0 +1,143 @@
+import {beforeEach, ddescribe, expect, inject, it, TestComponentBuilder} from 'angular2/testing';
+import {Component, DebugElement} from 'angular2/core';
+import {By} from 'angular2/platform/browser'
+import {MdProgressBar} from './progress-bar';
+
+
+export function main() {
+  describe('MdProgressCircular', () => {
+    let builder:TestComponentBuilder;
```
jelbourn: +1 space after colon (before type)

src/components/progress-bar/progress-bar.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+            expect(progressComponent.bufferTransform()).toBe('scaleX(0.6)');
+            done();
+          });
+    });
+  });
+}
+
+
+/** Gets a child DebugElement by tag name. */
+function getChildDebugElement(parent: DebugElement, selector: string): DebugElement {
```
jelbourn: You shouldn't need this function any more since doing `parent.query(By.css(selector))` is shorter than calling this function.

src/components/progress-bar/progress-bar.ts
```diff
@@ -0,0 +1,135 @@
...
+  styleUrls: ['./components/progress-bar/progress-bar.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MdProgressBar {
+  /**
+   * Value of the progress bar.
+   *
+   * Input:number, defaults to 0.
+   * value_ is bound to the host as the attribute aria-valuenow.
+   */
```
jelbourn: Could be more concise:
```
Value of the progressbar. Defaults to zero. Mirrored to aria-valuenow..
```

src/components/progress-bar/progress-bar.ts
```diff
@@ -0,0 +1,135 @@
...
+  bufferTransform() {
+    if (this.mode == ProgressMode.BUFFER) {
+      let scale = this.bufferValue / 100;
+      return `scaleX(${scale})`;
+    }
+  }
+
+
+  /** Clamps a value to be between two numbers, by default 0 and 100. */
+  static clamp(v: number, min: number = 0, max: number=100) {
```
jelbourn: Spaces around `=` for `max`

src/demo-app/progress-bar/progress-bar-demo.scss
```diff
@@ -0,0 +1,12 @@
+.demo-progress-bar-container {
+  width: 100%;
+}
+
+md-progress-bar {
```
jelbourn: Can you do this with a class instead and also prefix all the classes with `demo-` ?

src/components/progress-bar/progress-bar.ts
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * Input must be one of the values from ProgressMode, defaults to 'determinate'.
+   * mode is bound to the host as the attribute host.
+   */
+  @Input()
+  @OneOf([ProgressMode.DETERMINATE, ProgressMode.INDETERMINATE, ProgressMode.BUFFER, ProgressMode.QUERY])
+  mode: string;
+  @HostBinding('attr.mode')
+  get _mode() {
+      return this.mode;
```
jelbourn: -2 indent

src/components/progress-bar/progress-bar.ts
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * Input:number, defaults to 0.
+   * value_ is bound to the host as the attribute aria-valuenow.
+   */
+  @Input('value')
+  value_: number = 0;
+  @HostBinding('attr.aria-valuenow')
+  get _value() {
+    return this.value_;
+  }
```
jelbourn: I would format this like:
```ts
value_: number = 0;

@Input()
@HostBinding('attr.aria-valuenow')
get value() {
  ...
}
```

src/core/style/_variables.scss
```diff
@@ -31,3 +31,24 @@ $swift-ease-in: all $swift-ease-in-duration $swift-ease-in-timing-function !defa
 $swift-ease-in-out-duration: 0.5s !default;
 $swift-ease-in-out-timing-function: $ease-in-out-curve-function !default;
 $swift-ease-in-out: all $swift-ease-in-out-duration $swift-ease-in-out-timing-function !default;
+
+// Progress Bar Timing functions:
+// $progress-bar-primary-indeterminate-translate-step-1 has no timing function.
+$progress-bar-primary-indeterminate-translate-step-2: cubic-bezier(.5, 0, .701732, .495819) !default;
```
jelbourn: Seeing these, I think they're actually specific enough to go in the progressbar scss.
(I think we talked about this in person, I didn't know how specific they were).

src/core/style/_variables.scss
```diff
@@ -31,3 +31,24 @@ $swift-ease-in: all $swift-ease-in-duration $swift-ease-in-timing-function !defa
...
+$progress-bar-primary-indeterminate-translate-step-3: cubic-bezier(.302435, .381352, .55, .956352) !default;
```
jelbourn: Is there any info on where all the very specific values come from?

src/demo-app/progress-bar/progress-bar-demo.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  stepBufferProgressVal(val: number) {
+    this.bufferProgressValue += val;
+  }
+
+  stepBufferBufferVal(val: number) {
+    this.bufferBufferValue += val;
+  }
+
```
jelbourn: Not in this PR, but in a follow-up could you add a demo that fills up / resets on a timer?

jelbourn:
Finished first pass of comments.

src/components/progress-bar/progress-bar.scss
hansl: That article is the reference for this. Used to reference it everytime this topic came.

@jelbourn do you know if it's possible to use something like LESS Hat (http://lesshat.madebysource.com/) but for SASS? Maybe we have to build our own functions... Basically it's a library that allows you to use and declare complex functions and end up with stuff like:

```
md-css-svg-optimize('
    <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
         version="1.1" x="0px" y="0px" enable-background="new 0 0 5 2"
         xml:space="preserve" viewBox="0 0 5 2" preserveAspectRatio="none slice">
      <circle cx="1" cy="1" r="1" fill="${md-color($md-primary, 100)}"/>
    </svg>');
```

directly in the code and then perform all the string operations necessary to inject and optimize the proper code. I'm pretty sure SCSS is that powerful, and it would help streamline our optimizations AS WELL AS declare the SVG natively inside the css. Would be a big PLUS!

Such an optimization is to remove, for example, the xmlns, since browsers don't need it for `image/svg+xml` mimetypes.

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,205 @@
...
+}
+
+
+// Reverse the apparent directionality of progress vars for rtl.
+:host-context([dir="rtl"]) {
+  transform: rotateY(180deg);
+}
+
+
+/** Animations for indeterminate and query mode. */
```
hansl: Could you point to the resource you used to get those percentage for animations? They seem like magic numbers without context.

src/components/progress-bar/progress-bar.spec.ts
hansl: As well as everywhere. If you're using the default TypeScript formatting in WebStorm it will do that, you can change it in the preferences (Editor > Code Style > TypeScript > Spaces (tab) > [x] After type reference colon)

src/demo-app/progress-bar/progress-bar-demo.ts
josephperrott: TODO created

src/core/style/_variables.scss
josephperrott: Ack

src/core/style/_variables.scss
josephperrott: Moved to progressbar scss

src/components/progress-bar/progress-bar.ts
josephperrott: Since we want to bind `value` input to `value_`, we need to keep `@Input` as a decorator on the `value_` definition.

src/components/progress-bar/progress-bar.ts
josephperrott: Ack

src/demo-app/progress-bar/progress-bar-demo.scss
josephperrott: Ack

src/components/progress-bar/progress-bar.ts
josephperrott: Ack

src/components/progress-bar/progress-bar.ts
josephperrott: Ack

src/components/progress-bar/progress-bar.spec.ts
josephperrott: Ack

src/components/progress-bar/progress-bar.spec.ts
josephperrott: Ack

src/components/progress-bar/progress-bar.scss
josephperrott: Discussed offline.

Because the styles are structural, defining how the animations works we will leave them nested for this component.

src/components/progress-bar/progress-bar.scss
josephperrott: Ack

src/components/progress-bar/progress-bar.scss
josephperrott: Ack

src/components/progress-bar/progress-bar.scss
josephperrott: Ack

src/components/progress-bar/progress-bar.scss
josephperrott: Ack

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,265 @@
...
+// $md-progress-bar-secondary-indeterminate-translate-step-4 has no timing function
+
+$md-progress-bar-secondary-indeterminate-scale-step-1: cubic-bezier(.15, 0, .515058, .409685) !default;
+$md-progress-bar-secondary-indeterminate-scale-step-2: cubic-bezier(.310330, .284058, .8, .733712) !default;
+$md-progress-bar-secondary-indeterminate-scale-step-3: cubic-bezier(.4, .627035, .6, .902026) !default;
+// $md-progress-bar-secondary-indeterminate-scale-step-4 has no timing function
+
+/** Animations for indeterminate and query mode. */
+// Primary indicator.
+@keyframes md-progress-bar-primary-indeterminate-translate {
```
jelbourn: Is it possible at all to go into more detail about what each keyframe animation is actually doing?
Perhaps @traviskaufman could add more info

src/components/progress-bar/progress-bar.ts
jelbourn: It does the same:
http://plnkr.co/edit/foHQWXVgRgJHtNHeS6Zk?p=preview

src/components/progress-bar/progress-bar.scss
traviskaufman: Sure can :smiley: 

The indeterminate state is essentially made up of two progress bars, one **primary** (the one that is shown in both the determinate and indeterminate states) and one **auxiliary**, which essentially mirrors the primary progress bar in appearance but is only shown to assist with the indeterminate animations.

Given the [mocks](https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B0NGgBg38lWWYmNmallST001a1k/components-progressactivity-typesofindicators-061101_Linear_Sheet_xhdpi_003.webm) for the indeterminate APIs, you essentially have 4 animations being performed in tandem:

Keyframe Block | Description
--------------------- | ---------------
`primary-indeterminate-translate` | Translation of the primary progressbar across the screen
`primary-indeterminate-scale` | Scaling of the primary progressbar as it's being translated across the screen
`auxiliary-indeterminate-translate` | Translation of the auxiliary progressbar across the screen
`auxiliary-indeterminate-scale` | Scaling of the auxiliary progressbar as it's being translated across the screen

Because two different transform animations need to be applied at once, the translation is applied to the outer element and the scaling is applied to the inner element, which provides the illusion necessary to make the animation work.

josephperrott: added.  I just copied the description you gave into the comment.

src/components/progress-bar/progress-bar.scss
josephperrott: Are we okay with a TODO here while we determine how we want to go about this optimization?

jelbourn: I'm okay with addressing it in a follow-up PR if you tackle that next.

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,282 @@
...
+$md-buffer-bubbles-warn: (
+  "data:image/svg+xml;charset=UTF-8,%3Csvg%20version%3D%271.1%27%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27" +
+  "%20xmlns%3Axlink%3D%27http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%27%20x%3D%270px%27%20y%3D%270px%27%20enable-backgroun" +
+  "d%3D%27new%200%200%205%202%27%20xml%3Aspace%3D%27preserve%27%20viewBox%3D%270%200%205%202%27%20preserveAspectRatio" +
+  "%3D%27none%20slice%27%3E%3Ccircle%20cx%3D%271%27%20cy%3D%271%27%20r%3D%271%27%20fill%3D%27" +
+  md-color($md-warn, 100) + "%27%2F%3E%3C%2Fsvg%3E") !default;
+
+:host {
+  display: block;
+  /** Height is provided for md-progress-bar to act as a default. */
```
jelbourn: Use `//` style for "inline" comments that explaina some piece of code / css (vs. block style for describing a class, method, mixin, etc, )

src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,282 @@
...
+  height: $md-progress-bar-height;
+  overflow: hidden;
+  position: relative;
+  /** translateZ is added to force the md-progress-bar into its own GPU layer. */
+  transform: translateZ(0);
+  transition: opacity $md-progress-bar-piece-animation-duration linear;
+  width: 100%;
+
+  .background {
```
jelbourn: Even if nesting is okay, I still want to prefix these classes so that it's clear in the dev tools where they come from.

jelbourn: Can you also add a brief description to each class explaining its responsibility / what it's used for?

src/components/progress-bar/progress-bar.ts
```diff
@@ -0,0 +1,132 @@
...
+    'role': 'progressbar',
+    'aria-valuemin': '0',
+    'aria-valuemax': '100',
+  },
+  templateUrl: './components/progress-bar/progress-bar.html',
+  styleUrls: ['./components/progress-bar/progress-bar.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MdProgressBar {
+  /**
```
jelbourn: Nit: could condense to one line:
```ts
/** Value of the progressbar. Defaults to zero. Mirrored to aria-valuenow. */
```

src/components/progress-bar/progress-bar.ts
```diff
@@ -0,0 +1,132 @@
...
+   * Value of the progressbar.
+   *
+   * Defaults to zero. Mirrored to aria-valuenow.
+   */
+  @Input('value')
```
jelbourn: The `@Input()` decorator should go on the getter, otherwise angular is going to directly set the private property (and not use the setter).

src/demo-app/progress-bar/progress-bar-demo.ts
```diff
@@ -0,0 +1,29 @@
+import {Component} from 'angular2/core';
+import {MdButton} from '../../components/button/button';
+import {MdProgressBar} from '../../components/progress-bar/progress-bar';
+
+// TODO(josephperrott): Add an automatically filling example progress bar.
+
+@Component({
+    selector: 'progress-bar-demo',
```
jelbourn: -2 indent

src/demo-app/progress-bar/progress-bar-demo.ts
```diff
@@ -0,0 +1,29 @@
...
+    templateUrl: 'demo-app/progress-bar/progress-bar-demo.html',
+    styleUrls: ['demo-app/progress-bar/progress-bar-demo.css'],
+    directives: [MdProgressBar, MdButton]
+})
+export class ProgressBarDemo {
+  determinateProgressValue: number = 30;
+  bufferProgressValue: number = 30;
+  bufferBufferValue: number = 40;
+
+  stepDeterminateProgressVal(val: number, ref: string) {
```
jelbourn: `ref`?

src/components/progress-bar/progress-bar.html
```diff
@@ -0,0 +1,4 @@
+<div class="background"></div>
```
jelbourn: Could you add a comment explaining why `background` doesn't _contain_ the other elements?

src/components/progress-bar/progress-bar.scss
josephperrott: Done

src/components/progress-bar/progress-bar.scss
josephperrott: Done

src/components/progress-bar/progress-bar.ts
josephperrott: The thought here was the separation of what description of the variable from the "requirements" of the variable.

Personally I prefer the separation, but if you would like it on one line I can do that.  But as a wise man once told me, "vertical space is free"

src/demo-app/progress-bar/progress-bar-demo.ts
josephperrott: Done

src/demo-app/progress-bar/progress-bar-demo.ts
josephperrott: My bad, originally was going to make a single stepping method.

src/components/progress-bar/progress-bar.html
josephperrott: Done

src/components/progress-bar/progress-bar.ts
josephperrott: Not sure I understand what you are saying here.

I dont think it makes sense to put `@Input` on a getter.

It currently is:

```
@Input('value') _value: number = 0;
@HostBinding('attr.aria-valuenow') get value_() { return this._value; }
```

jelbourn:
Needs a fresh rebase

src/components/progress-bar/progress-bar.ts
jelbourn: That's not not really a distinction I make when reading comments. I think that single-lining description blocks like this when possible does make the code more readable, since the comment tokens themselves (and the IDE color) cue your brain what's happening. (vs `{ }` on the same line where your brain has to do more work). This is just bikeshedding, though- not super important.

src/components/progress-bar/progress-bar.ts
jelbourn: It should be 

```ts
private _value: number = 0;
...
@Input()
get value(): number {
  ...
}

set value(v: number) {
  ...
}
```

When the `@Input` is on the private member, the `setter` is not called when Angular updates `value`, which isn't what you want. Angular is smart enough to know that putting `@Input` on a getter or a setter means to simply treat that as the bound property.

josephperrott:
Rebased, PTAL

jelbourn:
LGTM

PR #131, opened 2016-03-07
feat(md-checkbox): Implement md-checkbox.

### LINKED ISSUE #35: mdCheckbox
- [x] Determine if component can be built on top of a native `<input type="checkbox">`
- [x] Unit tests
- [x] E2E tests
- [x] Documentation
- [x] Demo page

traviskaufman:
This commit adds the `md-checkbox` component to the material2 repo. It
includes unit tests and demos for the component. Note that while there
is a demo which exhibits all functionality of the component, the "Nested
Checklist" demo is still a WIP.

Fixes #35

### DIFF
--- a/src/components/checkbox/checkbox.html
+++ b/src/components/checkbox/checkbox.html
@@ -1,0 +1,22 @@
+<div class="md-checkbox-layout">
+  <div class="md-checkbox-inner-container">
+    <div class="md-checkbox-frame"></div>
+    <div class="md-checkbox-background">
+      <svg version="1.1"
+           class="md-checkbox-checkmark"
+           xmlns="http://www.w3.org/2000/svg"
+           viewBox="0 0 24 24"
+           xml:space="preserve">
+        <path class="md-checkbox-checkmark-path"
+              fill="none"
+              stroke="white"
+              d="M4.1,12.7 9,17.6 20.3,6.3"/>
+      </svg>
+      <!-- Element for rendering the indeterminate state checkbox. -->
+      <div class="md-checkbox-mixedmark"></div>
+    </div>
+  </div>
+  <label [id]="labelId">
+    <ng-content></ng-content>
+  </label>
+</div>
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -1,0 +1,460 @@
+@import "default-theme";
+@import "theme-functions";
+@import "variables";
+
+/** The width/height of the checkbox element. */
+$md-checkbox-size: 18px !default;
+/** The width of the line used to draw the checkmark / mixedmark. */
+$md-checkbox-mark-stroke-size: 2/15 * $md-checkbox-size !default;
+/** The width of the checkbox border shown when the checkbox is unchecked. */
+$md-checkbox-border-width: 2px;
+/** The color of the checkbox border. */
+$md-checkbox-border-color: if($md-is-dark-theme, rgba(white, 0.7), rgba(black, 0.54)) !default;
+/** The color of the checkbox's checkmark / mixedmark. */
+$md-checkbox-mark-color: md-color($md-background, background);
+/** The color that is used as the checkbox background when it is checked. */
+$md-checkbox-background-color: md-color($md-primary, 500);
+/** The base duration used for the majority of transitions for the checkbox. */
+$md-checkbox-transition-duration: 140ms;
+/** The amount of spacing between the checkbox and its label. */
+$md-checkbox-item-spacing: 4px;
+
+// Manual calculation done on SVG
+$_md-checkbox-mark-path-length: 22.910259;
+$_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1);
+
+/**
+ * Fades in the background of the checkbox when it goes from unchecked -> {checked,indeterminate}.
+ */
+@keyframes md-checkbox-fade-in-background {
+  0% {
+    opacity: 0;
+  }
+
+  50% {
+    opacity: 1;
+  }
+}
+
+/**
+ * Fades out the background of the checkbox when it goes from {checked,indeterminate} -> unchecked.
+ */
+@keyframes md-checkbox-fade-out-background {
+  0%, 50% {
+    opacity: 1;
+  }
+
+  100% {
+    opacity: 0;
+  }
+}
+
+/**
+ * "Draws" in the checkmark when the checkbox goes from unchecked -> checked.
+ */
+@keyframes md-checkbox-unchecked-checked-checkmark-path {
+  0%, 50% {
+    stroke-dashoffset: $_md-checkbox-mark-path-length;
+  }
+
+  50% {
+    animation-timing-function: $md-linear-out-slow-in-timing-function;
+  }
+
+  100% {
+    stroke-dashoffset: 0;
+  }
+}
+
+/**
+ * Horizontally expands the mixedmark when the checkbox goes from unchecked -> indeterminate.
+ */
+@keyframes md-checkbox-unchecked-indeterminate-mixedmark {
+  0%, 68.2% {
+    transform: scaleX(0);
+  }
+
+  68.2% {
+    animation-timing-function: cubic-bezier(0, 0, 0, 1);
+  }
+
+  100% {
+    transform: scaleX(1);
+  }
+}
+
+/**
+ * "Erases" the checkmark when the checkbox goes from checked -> unchecked.
+ */
+@keyframes md-checkbox-checked-unchecked-checkmark-path {
+  from {
+    animation-timing-function: $md-fast-out-linear-in-timing-function;
+    stroke-dashoffset: 0;
+  }
+
+  to {
+    stroke-dashoffset: $_md-checkbox-mark-path-length * -1;
+  }
+}
+
+/**
+ * Rotates and fades out the checkmark when the checkbox goes from checked -> indeterminate. This
+ * animation helps provide the illusion of the checkmark "morphing" into the mixedmark.
+ */
+@keyframes md-checkbox-checked-indeterminate-checkmark {
+  from {
+    animation-timing-function: $md-linear-out-slow-in-timing-function;
+    opacity: 1;
+    transform: rotate(0deg);
+  }
+
+  to {
+    opacity: 0;
+    transform: rotate(45deg);
+  }
+}
+
+/**
+ * Rotates and fades the checkmark back into position when the checkbox goes from indeterminate ->
+ * checked. This animation helps provide the illusion that the mixedmark is "morphing" into the
+ * checkmark.
+ */
+@keyframes md-checkbox-indeterminate-checked-checkmark {
+  from {
+    animation-timing-function: $_md-checkbox-indeterminate-checked-easing-function;
+    opacity: 0;
+    transform: rotate(45deg);
+  }
+
+  to {
+    opacity: 1;
+    transform: rotate(360deg);
+  }
+}
+
+/**
+ * Rotates and fades in the mixedmark when the checkbox goes from checked -> indeterminate. This
+ * animation, similar to md-checkbox-checked-indeterminate-checkmark, helps provide an illusion
+ * of "morphing" from checkmark -> mixedmark.
+ */
+@keyframes md-checkbox-checked-indeterminate-mixedmark {
+  from {
+    animation-timing-function: $md-linear-out-slow-in-timing-function;
+    opacity: 0;
+    transform: rotate(-45deg);
+  }
+
+  to {
+    opacity: 1;
+    transform: rotate(0deg);
+  }
+}
+
+/**
+ * Rotates and fades out the mixedmark when the checkbox goes from indeterminate -> checked. This
+ * animation, similar to md-checkbox-indeterminate-checked-checkmark, helps provide an illusion
+ * of "morphing" from mixedmark -> checkmark.
+ */
+@keyframes md-checkbox-indeterminate-checked-mixedmark {
+  from {
+    animation-timing-function: $_md-checkbox-indeterminate-checked-easing-function;
+    opacity: 1;
+    transform: rotate(0deg);
+  }
+
+  to {
+    opacity: 0;
+    transform: rotate(315deg);
+  }
+}
+
+/**
+ * Horizontally collapses and fades out the mixedmark when the checkbox goes from indeterminate ->
+ * unchecked.
+ */
+@keyframes md-checkbox-indeterminate-unchecked-mixedmark {
+  0% {
+    animation-timing-function: linear;
+    opacity: 1;
+    transform: scaleX(1);
+  }
+
+  32.8%, 100% {
+    opacity: 0;
+    transform: scaleX(0);
+  }
+}
+
+/**
+ * Applied to elements that cover the checkbox's entire inner container.
+ */
+%md-checkbox-cover-element {
+  bottom: 0;
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+}
+
+/**
+ * Applied to elements that are considered "marks" within the checkbox, e.g. the checkmark and
+ * the mixedmark.
+ */
+%md-checkbox-mark {
+  $width-padding-inset: 2 * $md-checkbox-border-width;
+
+  width: calc(100% - #{$width-padding-inset});
+  will-change: opacity, transform;
+}
+
+/**
+ * Applied to elements that appear to make up the outer box of the checkmark, such as the frame
+ * that contains the border and the actual background element that contains the marks.
+ */
+%md-checkbox-outer-box {
+  @extend %md-checkbox-cover-element;
+  border-radius: 2px;
+  box-sizing: border-box;
+  pointer-events: none;
+}
+
+.md-checkbox {
+  cursor: pointer;
+
+  &:focus {
+    // TODO(traviskaufman): Add ink ripple on focus state, once ripple is implemented.
+    outline: none;
+  }
+}
+
+.md-checkbox-layout {
+  align-items: baseline;
+  display: inline-flex;
+}
+
+.md-checkbox-inner-container {
+  display: inline-block;
+  height: $md-checkbox-size;
+  line-height: 0;
+  margin: auto;
+  margin-right: $md-checkbox-item-spacing;
+  order: 0;
+  position: relative;
+  vertical-align: middle;
+  white-space: nowrap;
+  width: $md-checkbox-size;
+
+  [dir="rtl"] & {
+    margin:{
+      left: $md-checkbox-item-spacing;
+      right: auto;
+    }
+  }
+}
+
+.md-checkbox-frame {
+  @extend %md-checkbox-outer-box;
+
+  background-color: transparent;
+  border: $md-checkbox-border-width solid $md-checkbox-border-color;
+  transition: border-color $md-checkbox-transition-duration $md-linear-out-slow-in-timing-function;
+  will-change: border-color;
+}
+
+.md-checkbox-background {
+  @extend %md-checkbox-outer-box;
+
+  align-items: center;
+  background-color: $md-checkbox-background-color;
+  opacity: 0;
+  display: inline-flex;
+  justify-content: center;
+  transition: background-color $md-checkbox-transition-duration
+                  $md-linear-out-slow-in-timing-function,
+              opacity $md-checkbox-transition-duration $md-linear-out-slow-in-timing-function;
+  will-change: background-color, opacity;
+}
+
+.md-checkbox-checkmark {
+  @extend %md-checkbox-cover-element;
+  @extend %md-checkbox-mark;
+
+  fill: $md-checkbox-mark-color;
+  width: 100%;
+}
+
+.md-checkbox-checkmark-path {
+  // !important is needed here because a stroke must be set as an attribute on the SVG in order
+  // for line animation to work properly.
+  stroke: $md-checkbox-mark-color !important;
+  stroke: {
+    dashoffset: $_md-checkbox-mark-path-length;
+    dasharray: $_md-checkbox-mark-path-length;
+    width: $md-checkbox-mark-stroke-size;
+  }
+}
+
+.md-checkbox-mixedmark {
+  @extend %md-checkbox-mark;
+
+  background-color: $md-checkbox-mark-color;
+  height: $md-checkbox-mark-stroke-size;
+  opacity: 0;
+  transform: scaleX(0) rotate(0deg);
+}
+
+.md-checkbox-align-end {
+  .md-checkbox-inner-container {
+    order: 1;
+    margin: {
+      left: $md-checkbox-item-spacing;
+      right: auto;
+    }
+
+    [dir="rtl"] & {
+      margin: {
+        left: auto;
+        right: $md-checkbox-item-spacing;
+      }
+    }
+  }
+}
+
+.md-checkbox-checked {
+  .md-checkbox-checkmark {
+    opacity: 1;
+  }
+
+  .md-checkbox-checkmark-path {
+    stroke-dashoffset: 0;
+  }
+
+  .md-checkbox-mixedmark {
+    transform: scaleX(1) rotate(-45deg);
+  }
+
+  .md-checkbox-background {
+    opacity: 1;
+  }
+}
+
+.md-checkbox-indeterminate {
+  .md-checkbox-background {
+    opacity: 1;
+  }
+
+  .md-checkbox-checkmark {
+    opacity: 0;
+    transform: rotate(45deg);
+  }
+
+  .md-checkbox-checkmark-path {
+    stroke-dashoffset: 0;
+  }
+
+  .md-checkbox-mixedmark {
+    opacity: 1;
+    transform: scaleX(1) rotate(0deg);
+  }
+}
+
+.md-checkbox-disabled {
+  // NOTE(traviskaufman): While the spec calls for translucent blacks/whites for disabled colors,
+  // this does not work well with elements layered on top of one another. To get around this we
+  // blend the colors together based on the base color and the theme background.
+  $white-30pct-opacity-on-dark-theme: #686868;
+  $black-26pct-opacity-on-light-theme: #b0b0b0;
+  $disabled-color: if(
+      $md-is-dark-theme, $white-30pct-opacity-on-dark-theme, $black-26pct-opacity-on-light-theme);
+
+  cursor: default;
+
+  &.md-checkbox-checked,
+  &.md-checkbox-indeterminate {
+    .md-checkbox-background {
+      background-color: $disabled-color;
+    }
+  }
+
+  &:not(.md-checkbox-checked) {
+    .md-checkbox-frame {
+      border-color: $disabled-color;
+    }
+  }
+}
+
+.md-checkbox-anim {
+  $indeterminate-change-duration: 500ms;
+
+  &-unchecked-checked {
+    .md-checkbox-background {
+      animation: $md-checkbox-transition-duration * 2 linear 0s md-checkbox-fade-in-background;
+    }
+
+    .md-checkbox-checkmark-path {
+      // Instead of delaying the animation, we simply multiply its length by 2 and begin the
+      // animation at 50% in order to prevent a flash of styles applied to a checked checkmark
+      // as the background is fading in before the animation begins.
+      animation:
+        $md-checkbox-transition-duration * 2 linear 0s md-checkbox-unchecked-checked-checkmark-path;
+    }
+  }
+
+  &-unchecked-indeterminate {
+    .md-checkbox-background {
+      animation: $md-checkbox-transition-duration * 2 linear 0s md-checkbox-fade-in-background;
+    }
+
+    .md-checkbox-mixedmark {
+      animation:
+        $md-checkbox-transition-duration linear 0s md-checkbox-unchecked-indeterminate-mixedmark;
+    }
+  }
+
+  &-checked-unchecked {
+    .md-checkbox-background {
+      animation: $md-checkbox-transition-duration * 2 linear 0s md-checkbox-fade-out-background;
+    }
+
+    .md-checkbox-checkmark-path {
+      animation:
+        $md-checkbox-transition-duration linear 0s md-checkbox-checked-unchecked-checkmark-path;
+    }
+  }
+
+  &-checked-indeterminate {
+    .md-checkbox-checkmark {
+      animation:
+        $md-checkbox-transition-duration linear 0s md-checkbox-checked-indeterminate-checkmark;
+    }
+
+    .md-checkbox-mixedmark {
+      animation:
+        $md-checkbox-transition-duration linear 0s md-checkbox-checked-indeterminate-mixedmark;
+    }
+  }
+
+  &-indeterminate-checked {
+    .md-checkbox-checkmark {
+      animation:
+        $indeterminate-change-duration linear 0s md-checkbox-indeterminate-checked-checkmark;
+    }
+
+    .md-checkbox-mixedmark {
+      animation:
+        $indeterminate-change-duration linear 0s md-checkbox-indeterminate-checked-mixedmark;
+    }
+  }
+
+  &-indeterminate-unchecked {
+    .md-checkbox-background {
+      animation: $md-checkbox-transition-duration * 2 linear 0s md-checkbox-fade-out-background;
+    }
+
+    .md-checkbox-mixedmark {
+      animation:
+        $indeterminate-change-duration * 0.6 linear 0s
+        md-checkbox-indeterminate-unchecked-mixedmark;
+    }
+  }
+}
--- a/src/components/checkbox/checkbox.spec.ts
+++ b/src/components/checkbox/checkbox.spec.ts
@@ -1,0 +1,733 @@
+import {
+  fakeAsync,
+  inject,
+  tick,
+  ComponentFixture,
+  TestComponentBuilder,
+} from 'angular2/testing';
+
+import {
+  it,
+  iit,
+  describe,
+  ddescribe,
+  expect,
+  beforeEach,
+} from '../../core/facade/testing';
+
+import {Component, DebugElement, EventEmitter} from 'angular2/core';
+import {By} from 'angular2/platform/browser';
+
+import {MdCheckbox} from './checkbox';
+
+// IE11 does not support event constructors, so we need to perform this check.
+var BROWSER_SUPPORTS_EVENT_CONSTRUCTORS: boolean = (function() {
+  // See: https://github.com/rauschma/event_constructors_check/blob/gh-pages/index.html#L39
+  try {
+    return new Event('submit', { bubbles: false }).bubbles === false &&
+           new Event('submit', { bubbles: true }).bubbles === true;
+  } catch (e) {
+    return false;
+  }
+})();
+
+export function main() {
+  describe('MdCheckbox', function() {
+    var builder: TestComponentBuilder;
+
+    beforeEach(inject([TestComponentBuilder], function(tcb: TestComponentBuilder) {
+      builder = tcb;
+    }));
+
+    it('attaches a class "md-checkbox" to the host element', function(done: () => void) {
+      builder.createAsync(CheckboxController).then(function(fixture) {
+        fixture.detectChanges();
+        let el = fixture.debugElement.query(By.css('.md-checkbox'));
+        expect(el).not.toBeNull();
+      }).then(done).catch(done);
+    });
+    it('attaches a unique id to the host element', function(done: () => void) {
+      builder.createAsync(CheckboxMultiController).then(function(fixture) {
+        fixture.detectChanges();
+        let first = fixture.debugElement.query(By.css('.md-checkbox:first-of-type'));
+        let second = fixture.debugElement.query(By.css('.md-checkbox:nth-of-type(2)'));
+        expect(first.nativeElement.id).toMatch(/^md\-checkbox\-\d$/g);
+        expect(second.nativeElement.id).toMatch(/^md\-checkbox\-\d$/g);
+        expect(first.nativeElement.id).not.toEqual(second.nativeElement.id);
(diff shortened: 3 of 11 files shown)


src/components/checkbox/checkbox.html
```diff
@@ -0,0 +1,22 @@
+<div class="md-checkbox-layout">
+  <div class="md-checkbox-inner-container">
+    <div class="md-checkbox-frame"></div>
+    <div class="md-checkbox-background">
+      <svg version="1.1"
+           class="md-checkbox-checkmark"
+           xmlns="http://www.w3.org/2000/svg"
+           xmlns:xlink="http://www.w3.org/1999/xlink"
+           viewBox="0 0 24 24"
+           xml:space="preserve">
```
jelbourn: Are `xml:space` and `xmlns:xlink` necessary?

src/components/checkbox/checkbox.html
```diff
@@ -0,0 +1,22 @@
...
+        <path class="md-checkbox-checkmark-path"
+              fill="none"
+              stroke="white"
+              d="M4.1,12.7 9,17.6 20.3,6.3"/>
+      </svg>
+      <div class="md-checkbox-mixedmark"></div>
```
jelbourn: Add a comment for this element?
```html
<!-- Element for rendering the indeterminate state checkbox. -->
```

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,401 @@
...
+@import "variables";
+
+/** The width/height of the checkbox element. */
+$md-checkbox-size: 18px !default;
+/** The width of the line used to draw the checkmark / mixedmark. */
+$md-checkbox-mark-stroke-size: 2/15 * $md-checkbox-size !default;
+/** The width of the checkbox border shown when the checkbox is unchecked. */
+$md-checkbox-border-width: 2px;
+/** The color of the checkbox border. */
+$md-checkbox-border-color: if($md-is-dark-theme, rgba(white, 0.7), rgba(black, 0.54)) !default;
```
jelbourn: Could this border color be defined in terms of the theme background / foreground color, or is it unique to the checkbox?

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,401 @@
...
+/** The color of the checkbox's checkmark / mixedmark. */
+$md-checkbox-mark-color: map-get($md-background, background);
```
jelbourn: Rather than using `map-get` directly, use `md-color` (just for consistency)

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,401 @@
...
+/** The color that is used as the checkbox background when it is checked. */
+$md-checkbox-background-color: md-color($md-primary, 500);
+/** The base duration used for the majority of transitions for the checkbox. */
+$md-checkbox-transition-duration: 140ms;
```
jelbourn: Is this duration something that could be applied to multiple components / effects or is it specific to the checkbox?

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,401 @@
...
+/** The amount of spacing between the checkbox and its label. */
+$md-checkbox-item-spacing: 4px;
+
+// Manual calculation done on SVG
+$_md-checkbox-mark-path-length: 22.910259;
+$_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1);
+
+@keyframes fade-in-background {
```
jelbourn: Should the keyframe animation names be prefixed with `md-` in order to avoid collisions?

jelbourn: It would be good, for each keyframe, to have a brief explanation of what it's used for in the checkbox.

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,401 @@
...
+
+%md-checkbox-background {
+  @extend %md-checkbox-cover-element;
+  border-radius: 2px;
+  box-sizing: border-box;
+  pointer-events: none;
+}
+
+.md-checkbox {
+  &:not(.md-checkbox-disabled) {
```
jelbourn: Why not

```css
.md-checkbox {
  cursor: pointer;
}

.md-checkbox-disabled {
  pointer: default;
}
```
?

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,401 @@
...
+    transform: scaleX(1);
+  }
+
+  32.8%, 100% {
+    opacity: 0;
+    transform: scaleX(0);
+  }
+}
+
+%md-checkbox-cover-element {
```
jelbourn: Could you add a brief comment on the role of each placeholder / class where it's not obvious? E.g., what the "cover-element" or "checkbox-frame" are used for.

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,401 @@
...
+    transform: scaleX(1) rotate(0deg);
+  }
+}
+
+.md-checkbox-disabled {
+  // NOTE(traviskaufman): While the spec calls for translucent blacks/whites for disabled colors,
+  // this does not work well with elements layered on top of one another. To get around this we
+  // blend the colors together based on the base color and the theme background.
+  $white-30pct-opacity-on-dark-theme: #686868;
+  $black-26pct-opacity-on-light-theme: #b0b0b0;
```
jelbourn: I actually wrote an scss alpha-blend function a while ago if that would make this better: https://gist.github.com/jelbourn/f890a79a48cb21a5246a

src/components/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,749 @@
...
+    var builder: TestComponentBuilder;
+
+    beforeEach(inject([TestComponentBuilder], function(tcb: TestComponentBuilder) {
+      builder = tcb;
+    }));
+
+    it('attaches a class "md-checkbox" to the host element', function(done: () => void) {
+      builder.createAsync(CheckboxController).then(function(fixture) {
+        fixture.detectChanges();
+        const el = fixture.debugElement.query(By.css('.md-checkbox'));
```
jelbourn: Using `const` like this won't work when we eventually transpile our unit tests to dart.

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+  templateUrl: './components/checkbox/checkbox.html',
+  styleUrls: ['./components/checkbox/checkbox.css'],
+  host: {
+    '[id]': 'id',
+    '[class.md-checkbox]': 'true',
+    '[class.md-checkbox-indeterminate]': 'indeterminate',
+    '[class.md-checkbox-checked]': 'checked',
+    '[class.md-checkbox-disabled]': 'disabled',
+    '[class.md-checkbox-align-end]': 'align == "end"',
+    '[attr.role]': '"checkbox"',
```
jelbourn: You can just do:
```ts
'role': 'checkbox' 
```
without the `[ ]` binding

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+    '[attr.tabindex]': 'disabled ? -1 : tabindex',
```
jelbourn: You should be able to do `tabIndex` without `attr.` since it's a real property on `HTMLElement`

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+    '[attr.aria-label]': 'ariaLabel',
+    '[attr.aria-labelledby]': 'labelId',
+    '[attr.aria-checked]': 'getAriaChecked()',
+    '[attr.aria-disabled]': 'disabled',
+    '(click)': 'toggle($event)',
+    '(keyup.space)': 'toggle($event)',
+    '(blur)': 'onTouched()'
+  },
```
jelbourn: FYI, in case you didn't know: you can use the `@HostBinding` and `@HostListener` decorators to declare these same things inline in the component. For example:
```ts
@HostBinding('class.md-checkbox-disabled')
public disabled: boolean;

@HostListener('blur')
onTouched() {
}
```

In _general_ the decorators are preferred, but the checkbox has a lot of expressions and classes such that it seems cleaner (to me) to leave them where they are.

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+   * will immediately set `indeterminate` to false.
+   */
+  @Input() get checked() {
+    return this._checked;
+  }
+
+  set checked(checked: boolean) {
+    // When checked is set, it immediately overwrites an indeterminate state, regardless of
+    // whether it's true or false.
+    this._indeterminate = false;
```
jelbourn: What made you go with this over the native behavior? (where `indeterminate = true` always trumps `checked = true`)

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+  private _checked: boolean = false;
+
+  private _indeterminate: boolean = false;
+
+  private _changeSubscription: {unsubscribe: () => any} = null;
+
+  constructor(private _renderer: Renderer, private _elementRef: ElementRef) {}
+
+  /**
+   * True when the checkbox is checked, false otherwise. Note that setting `checked` while
```
jelbourn: nit: Change `True when..., false otherwise` wording to `Whether the checkbox is checked.`

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+const MD_CHECKBOX_CONTROL_VALUE_ACCESSOR = CONST_EXPR(new Provider(
+    NG_VALUE_ACCESSOR, {
+      useExisting: forwardRef(() => MdCheckbox),
+      multi: true
+    }));
+
+/**
+ * Represents the different states that require custom transitions between them.
+ */
+enum UiTransitionState {
```
jelbourn: How about `CheckTransitionState` ?

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+   * be interacted with. The correct ARIA attributes are applied to denote this to assistive
+   * technology.
+   */
+  @Input() disabled: boolean = false;
+
+  /**
+   * The tabindex attribute for the checkbox. Note that when the checkbox is disabled, the attribute
+   * on the host element will be set to -1, regardless of the actual tabindex value.
+   */
+  @Input() tabindex: number = 0;
```
jelbourn: If you make this `tabIndex` (camelCase), it should capture both the component's state and the underlying native element property.

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+
+  /** Event emitted when the checkbox's `checked` value changes. */
+  @Output() change: EventEmitter<boolean> = new EventEmitter();
```
jelbourn: I think `boolean` isn't exactly the right type, here, but as I mentioned to Michael, I'm not sure yet whether we should be making our own events for everything.

src/components/checkbox/checkbox.html
traviskaufman: I don't think `xmlns:xlink` is but I'd argue `xml:space` is because it explicitly instructs browsers how to parse the SVG XML. From the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xml:space)

> Note that this attribute influences the way a browser should parse the xml content and therefore will change the way the DOM is built.

src/components/checkbox/checkbox.html
traviskaufman: Are you sure you want that showing up every time the checkbox is rendered?

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+  /** Returns the proper aria-checked attribute value based on the checkbox's state. */
+  getAriaChecked() {
+    if (this.indeterminate) {
+      return 'mixed';
+    }
+    return this.checked ? 'true' : 'false';
+  }
+
+  /** Toggles the checked state of the checkbox. If the checkbox is disabled, this does nothing. */
+  toggle(event: Event) {
```
jelbourn: Might want to use an internal method for the event-handling toggle so that `toggle()` as a method can be part of the public API.

src/components/checkbox/checkbox.scss
traviskaufman: The spec [explicitly specifies](https://www.google.com/design/spec/components/selection-controls.html#selection-controls-checkbox) the colors that the checkbox border should be: 

> Light
>
> On: Swatch 500, Opacity  100%
> Off: #000000, Opacity  54%
> Disabled: #000000, Opacity  26%

> Dark
>
> On: Swatch 500, Opacity  100%
> Off: #FFFFFF, Opacity  70%
> Disabled: #FFFFFF, Opacity  30%

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+    if (this.disabled) {
+      event.stopPropagation();
+      return;
+    }
+    this.checked = !this.checked;
+  }
+
+  /** Implemented as part of ControlValueAccessor. */
+  writeValue(value: any) {
+    this.checked = !!value;
```
jelbourn: What Dart codes does this actually transpile to?

src/components/checkbox/checkbox.scss
traviskaufman: AFAIK it's specific to the checkbox. My thinking behind variable-izing it was that if you had unique checkboxes where you wanted to change the overall durations of things (e.g. slow down the duration on a very large checkbox), this would allow you to do so. I'm happy to make it "private" if you'd prefer!

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+    this._changeSubscription = <{unsubscribe: () => any}>this.change.subscribe(fn);
+  }
+
+  /** Implemented as part of ControlValueAccessor. */
+  registerOnTouched(fn: any) {
+    this.onTouched = fn;
+  }
+
+  private _transitionUiState(newState: UiTransitionState) {
```
jelbourn: Maybe `_transitionCheckState` ?

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+    const oldState = this._currentUiState;
```
jelbourn: These `const` won't work in Dart (have to use `let`)

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+    const renderer = this._renderer;
+    const elementRef = this._elementRef;
+
+    if (oldState === newState){
+      return;
+    }
+    if (this._currentAnimationClass) {
```
jelbourn: Truthy checks on strings won't work in Dart; conditions have to be exactly `boolean`

src/components/checkbox/checkbox.scss
traviskaufman: :+1: Good point, I will prefix with `md-checkbox-`

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    this._currentAnimationClass = this._getAnimationClassForStateTransition(oldState, newState);
+    this._currentUiState = newState;
+
+    if (this._currentAnimationClass) {
+      renderer.setElementClass(elementRef.nativeElement, this._currentAnimationClass, true);
+    }
+  }
+
+  private _getAnimationClassForStateTransition(
```
jelbourn: Perhaps `_getAnimationClassForCheckStateTransition`?

src/demo-app/checkbox/checkbox-demo.ts
```diff
@@ -0,0 +1,97 @@
...
+
+@Component({
+  selector: 'md-checkbox-demo-nested-checklist',
+  styles: [`
+    li {
+      margin-bottom: 4px;
+    }
+  `],
+  template: `
+    <h2>Tasks</h2>
```
jelbourn: Why not `templateUrl` w/ separate file?

src/components/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,749 @@
...
+        expect(el).not.toBeNull();
+      }).then(done).catch(done);
+    });
+
+    it('creates a checkbox frame element', function(done: () => void) {
```
jelbourn: You don't need unit tests for the simple presence of elements in the template; they're really more implementation details than behaviors that need to be tested.

src/components/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,749 @@
...
+
+    it('adds a checkbox role attribute to the host element', function(done: () => void) {
+      builder.createAsync(CheckboxController).then(function(fixture) {
+        fixture.detectChanges();
+        const el = fixture.debugElement.query(By.css('.md-checkbox'));
+        expect(el.nativeElement.getAttribute('role')).toEqual('checkbox');
+      }).then(done).catch(done);
+    });
+
+    it('sets "aria-checked" to "false" on the host element', function(done: () => void) {
```
jelbourn: `sets` > `defaults` ?

src/components/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,749 @@
...
+        }).then(done).catch(done);
+      });
+
+      it('sets an "md-checkbox-align-end" class on the host element', function() {
+        const el = fixture.debugElement.query(By.css('.md-checkbox'));
+        expect(el.nativeElement.className).toContain('md-checkbox-align-end');
+      });
+    });
+
+    describe('when the checkbox\'s checked value is set', function() {
```
jelbourn: Feel free to use backticks to avoid escaping the `'`

src/components/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,749 @@
...
+      var fixture: ComponentFixture;
+      var controller: CheckboxController;
+      var changePromise: Promise<boolean>;
+      var waitingForChange: boolean;
+
+      function waitForChange(): Promise<boolean> {
```
jelbourn: Would it work if, instead of doing this, you instead injected the `NgZone` and added a handler for `onEventDone`?

src/components/checkbox/checkbox.scss
traviskaufman: :+1:

src/components/checkbox/checkbox.html
jelbourn: We'll strip html comments eventually (when we have a prod build).

src/components/checkbox/checkbox.scss
traviskaufman: :+1:

src/components/checkbox/checkbox.scss
traviskaufman: :+1: I think you meant `cursor: default;` though

src/components/checkbox/checkbox.scss
traviskaufman: I'm not sure that would work since it looks like it produces a non-opaque color, however it's a really cool function :smile: (I starred it!). Since the disabled colors are theme-independent, and the variable names seem semantic enough, I don't think there's a major need to include it.

src/components/checkbox/checkbox.spec.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.spec.ts
traviskaufman: not sure I 100% agree with that, as I've definitely been bitten in the past by refactoring code and accidentally deleting a char from a className resulting in time spent tracking down where I made that mistake. Just my unsolicited $0.02 though; I'll go ahead and remove them.

src/components/checkbox/checkbox.spec.ts
traviskaufman: is there an example of this somewhere which I can look at?

src/components/checkbox/checkbox.spec.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: OOC what's the reasoning behind `boolean` not being the right type? Would you rather have an event as an object/interface for extensibility? Do you want me to change anything about this implementation for this PR?

src/components/checkbox/checkbox.ts
traviskaufman: :+1: duly noted; based on your comments above I will leave as is.

src/components/checkbox/checkbox.ts
traviskaufman: I wasn't necessarily able to get that working properly. Also, foregoing a binding would mean making the code that handles setting the native element's `tabindex` to `-1` when disabled more complicated, so I'm more partial to the current implementation. WDYT?

src/components/checkbox/checkbox.ts
traviskaufman: That case is only true in the situation where `indeterminate` is set _after_ `checked`. If checked is then set again, it overrides any indeterminate state.

src/components/checkbox/checkbox.ts
jelbourn: Not for this PR, mainly just thinking at you in comment form. We _probably_ want to event have something like `MdCheckboxChangeEvent` that contains the new state as well as a reference to the component instance and maybe some other stuff.

src/demo-app/checkbox/checkbox-demo.ts
traviskaufman: Seemed like overkill. I'm from the camp that likes everything in one file (scripts, styles, templates, etc.) barring a _major_ file size overhaul. Keeping everything in one place reduces developer fatigue of jumping around files to edit tightly coupled dependencies, and creates one singular source of truth for component information. Furthermore, barring some exception (e.g. the plethora of animations and style changes material design input controls have to acommodate, or just using preprocessors in general), if a component's DOM / styles get too large, it's probably a good indication that it's time to refactor that component into smaller ones.

WDYT? I'm happy to move it into its own file, IDC that much either way :smiley:

src/components/checkbox/checkbox.ts
traviskaufman: :+1: makes sense!

src/components/checkbox/checkbox.ts
traviskaufman:
```dart
/** Implemented as part of ControlValueAccessor. */
writeValue(dynamic value) {
  this.checked = !!value;
}
```
\> [dartpad](https://dartpad.dartlang.org/1883da33198e6b78662d)

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: :+1: my bad I'll look more closely at the Dart Gotchas doc next time.

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.spec.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.scss
traviskaufman: :+1:

src/components/checkbox/checkbox.html
traviskaufman: :+1: gotcha!

src/components/checkbox/checkbox.ts
traviskaufman: Gotcha. That sounds reasonable, and it future-proofs events in case additional information needs to be added.

src/components/checkbox/checkbox.ts
jelbourn: https://jsbin.com/hozuse/edit?html,js,output

If I set `indeterminate = true` and then `checked = true`, it stays in indeterminate mode. Am I missing something?

src/demo-app/checkbox/checkbox-demo.ts
jelbourn: I'm much more on the side of separate files for everything; I generally find it easier to get overwhelmed when everything is all in one. 

I would say to go for separate files here, though, mainly because it's more toolable (i.e., able to glob all html files).

src/components/checkbox/checkbox.spec.ts
jelbourn: You inject `NgZone` the same was as `TestComponentBuilder`:
```ts
inject([TestComponentBuilder, NgZone], (tcb, ngZone) => {
  zone = ngZone;
});
```

Then, in the test, you could try
```
zone.onEventDone(() => {
  // make assertions
});
```

I'm not sure if it will actually work, but it may reduce the code you have here without going to `fakeAsync` and its problems with `templateUrl`.

src/components/checkbox/checkbox.ts
jelbourn: Huh, I remember this working at some point, but it seems to not be the case now. Nevermind,

src/components/checkbox/checkbox.ts
zoechi: Dartpad doesn't use checked mode.

https://dartpad.dartlang.org/8f79d250451143ec113c shows my results when run locally in checked mode.

src/components/checkbox/checkbox.ts
traviskaufman: Ah I see, so that's my mistake. It only becomes indeterminate when you _interact_ with the checkbox, not change it programmatically. This makes sense for the web platform, but since `[checked]` is an input in this component, I can see use cases where if the `checked` value changed programmatically while the checkbox was indeterminate, it would update. 

Do you think that it would be more idiomatic to angular to have the checked value take precedence   when a change is detected, or should we stick to what the web platform is doing? 

As a third option, we could provide an input to opt into this behavior, e.g. `overrideIndeterminate` or something of that sort, but personally I think going with one of the above options is better since this just bloats the API more.

jelbourn: It's tricky because the imperative API and the declarative API are at odds. The problem with having one of the properties un-set the other is that now the state of the component can depend on the order in which Angular sets the properties when you're using `[checked]` and `[indetermindate]`. 

My initial thought was to follow the native element completely, which makes it easy to defend the choice; "Of course we do it that way, it's what the native checkbox does." The more I think about it, though, the more I think the current approach is actually better since it more closely captures how people will actually want to use it.

traviskaufman: I feel the same way. That's a good point about properties clobbering one another when they're updated within the same cycle, but I feel like there aren't that many situations where you'd be setting both of those at the same time anyway. I'll update the comments to reflect this discussion.

src/demo-app/checkbox/checkbox-demo.ts
traviskaufman: :+1: duly noted! I'll move the html into its own file.

src/core/style/_variables.scss
```diff
@@ -18,6 +18,11 @@ $pi: 3.14159265;
 // Easing Curves
 // TODO(jelbourn): all of these need to be revisited
 
+// The default animation curves used by material design.
+$linear-out-slow-in-timing-function: cubic-bezier(0.0, 0.0, 0.2, 0.1) !default;
```
jelbourn: Can you prefix these variables with `md-`? (it's a bug that the existing ones don't have a prefix)

src/components/checkbox/checkbox.spec.ts
traviskaufman: I tried this out and couldn't get it to work. Are you okay with leaving it the way it is?

src/core/style/_variables.scss
traviskaufman: :+1:

traviskaufman:
@jelbourn all changes made!

src/components/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,713 @@
...
+        let label = el.nativeElement.querySelector('label');
+        expect(el.nativeElement.getAttribute('aria-labelledby')).toEqual(label.id);
+      }).then(done).catch(done);
+    });
+
+    describe('when given an "align" input with a value of "end"', function() {
+      var fixture: ComponentFixture;
+
+      beforeEach(function(done: () => void) {
+        builder.createAsync(CheckboxEndAlignedController).then(function(_fixture_) {
```
jelbourn: Rather than `_fixture_`, I'd be fine with just saying `f` (since the first thing you do is set it)

src/components/checkbox/checkbox.spec.ts
jelbourn: It's good for now, but could you leave a TODO pointed to your github issue?

src/components/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,713 @@
...
+    value: ' ',
+    enumerable: false,
+    writable: false,
+    configurable: true
+  });
+  spyOn(kbdEvent, 'stopPropagation').and.callThrough();
+  el.nativeElement.dispatchEvent(kbdEvent);
+  fixture.detectChanges();
+  return kbdEvent;
+}
```
jelbourn: When we start running unit tests in Dart these definitely aren't going to work, but we can address that in a follow-up. Could you leave a TODO?

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,255 @@
...
+    '[class.md-checkbox-checked]': 'checked',
+    '[class.md-checkbox-disabled]': 'disabled',
+    '[class.md-checkbox-align-end]': 'align == "end"',
+    '[tabindex]': 'disabled ? -1 : tabindex',
+    '[attr.aria-label]': 'ariaLabel',
+    '[attr.aria-labelledby]': 'labelId',
+    '[attr.aria-checked]': 'getAriaChecked()',
+    '[attr.aria-disabled]': 'disabled',
+    '(click)': 'onInteractionEvent($event)',
+    '(keyup.space)': 'onInteractionEvent($event)',
```
jelbourn: Why `keyup` instead of `keydown` ?

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,255 @@
...
+    '(blur)': 'onTouched()'
+  },
+  providers: [MD_CHECKBOX_CONTROL_VALUE_ACCESSOR],
+  encapsulation: ViewEncapsulation.None
```
jelbourn: You should be able to add 
```
```
with everything still working since all of the inputs are primitives

src/components/checkbox/checkbox.ts
traviskaufman: Check out http://jsbin.com/reyefucite/edit?html,output (which is literally just a checkbox), if you focus on the checkbox, and then press the spacebar down, notice how the checkbox does not become checked until it's released. Would you prefer md-checkbox just becomes checked on keydown?

src/components/checkbox/checkbox.spec.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.spec.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
jelbourn: I was just curious- following that sounds good to me.

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

traviskaufman:
@jelbourn all changes made!

jelbourn:
LGTM

@kara can you do a pass? I think it's generally good to have a second set of eyes on any larger change.

jelbourn:
@traviskaufman Looks like CI is failing:
```
IE 11.0.0 (Windows 8.1 0.0.0) MdCheckbox when the checkbox is clicked toggles the checked value FAILED
	TypeError: Object doesn't support this action
```
Is this something we can change in the code? Looks like there's a missing polyfill.

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,460 @@
...
+  @extend %md-checkbox-background;
+
+  background-color: transparent;
+  border: $md-checkbox-border-width solid $md-checkbox-border-color;
+  transition: border-color $md-checkbox-transition-duration $md-linear-out-slow-in-timing-function;
+  will-change: border-color;
+}
+
+.md-checkbox-background {
+  @extend %md-checkbox-background;
```
kara: Nitpick: do you think it's confusing that the class and the placeholder share the same name?  Since the comment above cites the "outer box", that might be a more descriptive name for the placeholder.

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  private _transitionCheckState(newState: TransitionCheckState) {
+    let oldState = this._currentCheckState;
+    let renderer = this._renderer;
+    let elementRef = this._elementRef;
+
+    if (oldState === newState){
+      return;
+    }
+    if (this._currentAnimationClass.length > 0) {
```
kara: Nitpick: you could shorten this to `if (this._currentAnimationClass.length)`

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  /**
+   * Whether the checkbox is indeterminate. This is also known as "mixed" mode and can be used to
+   * represent a checkbox with three states, e.g. a checkbox that represents a nested list of
+   * checkable items. Note that whenever `checked` is set, indeterminate is immediately set to
+   * false. This differs from the web platform in that indeterminate state on native
+   * checkboxes is only remove when the user manually checks the checkbox (rather than setting the
+   * `checked` property programmatically). However, we feel that this behavior is more accommodating
+   * to the way consumers would invision using this component.
```
kara: Typo - I think you meant "envision"?

src/components/checkbox/checkbox.ts
jelbourn: Not in Dart-land, unfortunately

kara: I had a sneaking suspicion that might be the case...

src/components/checkbox/checkbox.ts
```diff
@@ -0,0 +1,257 @@
...
+    '[class.md-checkbox-disabled]': 'disabled',
+    '[class.md-checkbox-align-end]': 'align == "end"',
+    '[tabindex]': 'disabled ? -1 : tabindex',
+    '[attr.aria-label]': 'ariaLabel',
+    '[attr.aria-labelledby]': 'labelId',
+    '[attr.aria-checked]': 'getAriaChecked()',
+    '[attr.aria-disabled]': 'disabled',
+    '(click)': 'onInteractionEvent($event)',
+    '(keyup.space)': 'onInteractionEvent($event)',
+    '(blur)': 'onTouched()'
```
kara: I'm probably missing something, but why is the "onTouched" function called on blur?  At first glance, seems like it should be onBlur

traviskaufman: Yep! That's exactly it. I stole this strategy (and therefore method name) from the native [CheckboxControlValueAccessor](https://github.com/angular/angular/blob/8bb66a5eb382f454af80523894efba9c61fc7e7b/modules/angular2/src/common/forms/directives/checkbox_value_accessor.ts#L25).

src/components/checkbox/checkbox.html
```diff
@@ -0,0 +1,22 @@
...
+        <path class="md-checkbox-checkmark-path"
+              fill="none"
+              stroke="white"
+              d="M4.1,12.7 9,17.6 20.3,6.3"/>
+      </svg>
+      <!-- Element for rendering the indeterminate state checkbox. -->
+      <div class="md-checkbox-mixedmark"></div>
+    </div>
+  </div>
+  <label [id]="labelId" [attr.for]="id">
```
jelbourn: @marcysutton just tried this out and find that it doesn't quite work. Instead we should use `aria-labelledby` on the host element.

marcysutton: Yep. `<label for="">` only works with native inputs.

src/components/checkbox/checkbox.scss
traviskaufman: :+1:

src/components/checkbox/checkbox.ts
traviskaufman: :+1:

src/components/checkbox/checkbox.html
traviskaufman: :+1: I'll remove `for`; `aria-labelledby` is already on there.

traviskaufman:
@jelbourn @kara @marcysutton all changes made!

jelbourn:
@traviskaufman looks like there a few failures left on CI?

traviskaufman:
@jelbourn looking into it now

PR #134, opened 2016-03-08
build(linting): Enforce tslint in travis-ci

### LINKED ISSUE #85: Travis CI enforces TSLint

d3viant0ne:
- Executes tslint via npm script to be replaced by ng lint when it matures
- allows local execution with "npm run tslint"
- Executes in the before_script to fail fast ( approx. 50 seconds for feedback ): [Failed Build Example](https://travis-ci.org/d3viant0ne/material2/builds/114421734)
- tslint runs against <pre> "src/***/**.ts" </pre>

- Fixes numerous linting errors, specifics per component are listed in the commit messages

> Requires tslint version update to 3.5.0 to take advantage of ignore patterns

resolves #85


src/core/facade/test_injector.ts
```diff
@@ -1,14 +1,16 @@
 import {Injector, Provider, PLATFORM_INITIALIZER} from 'angular2/core';
-import {BaseException, ExceptionHandler} from 'angular2/src/facade/exceptions';
+import {BaseException} from 'angular2/src/facade/exceptions';
 import {ListWrapper} from 'angular2/src/facade/collection';
 import {FunctionWrapper, isPresent, Type} from 'angular2/src/facade/lang';
 
 export class TestInjector {
+  /* tslint:disable:variable-name */
```
hansl: I think we should fix the names instead of disabling the tslint.  This is just a matter of moving the `_` at the end, and removing it from the var.

src/components/sidenav/sidenav.ts
```diff
@@ -325,6 +325,7 @@ export class MdSidenavLayout implements AfterContentInit {
     }
   }
 
+  /* tslint:disable:no-unused-variable */
```
hansl: Why is this exception necessary?

src/components/sidenav/sidenav.spec.ts
```diff
@@ -178,12 +180,13 @@ export function main() {
...
         let fixture = createFixture(builder, template, style);
-        let sidenav: MdSidenav = fixture.debugElement.query(By.directive(MdSidenav)).componentInstance;
+        let sidenav: MdSidenav = fixture.debugElement
+          .query(By.directive(MdSidenav)).componentInstance;
 
         let closePromise: Promise<void>;
-        let openCalled: boolean = false;
-        let openCancelled: boolean = false;
-        let closeCalled: boolean = false;
+        let openCalled = false;
```
hansl: Have you tried compiling this to Dart?

src/components/sidenav/sidenav.spec.ts
```diff
@@ -111,7 +110,8 @@ export function main() {
         let sidenavElement = fixture.debugElement.query(By.css('md-sidenav'));
         let sidenavBackdropElement = fixture.debugElement.query(By.css('.md-sidenav-backdrop'));
         expect(getComputedStyle(sidenavElement.nativeElement).visibility).toEqual('visible');
-        expect(getComputedStyle(sidenavBackdropElement.nativeElement).visibility).toEqual('visible');
+        expect(getComputedStyle(sidenavBackdropElement.nativeElement)
```
hansl: If this is only a matter of line length, could you split after `.visibility`? it looks better like this:

```
expect(getComputedStyle(sidenavBackdropElement.nativeElement).visibility)
    .toEqual('visible');
```

src/components/button/button.spec.ts
```diff
@@ -125,15 +123,18 @@ export function main() {
 }
 
 /** Shortcut function to use instead of `injectAsync` for less boilerplate on each `it`. */
+/* tslint:disable:no-unused-variable */
```
hansl: What's the unused variable?

d3viant0ne: tslint picks up function testAsync() as unused

src/components/sidenav/sidenav.spec.ts
d3viant0ne: screen shot 2016-03-08 at 12 51 53 pm

src/components/button/button.spec.ts
hansl: We should probably remove it unless it's actually used somewhere else.

src/components/sidenav/sidenav.ts
d3viant0ne: tslint doesn't exclude private class members in the no-unused-varaible rule

screen shot 2016-03-08 at 1 14 56 pm

d3viant0ne: This has the same effect without wrapping the entire block if you prefer.
screen shot 2016-03-08 at 1 23 41 pm

src/components/sidenav/sidenav.spec.ts
d3viant0ne: resolved with 158bea1

src/components/sidenav/sidenav.spec.ts
hansl: No it shouldn't cause issues.

src/components/button/button.ts
```diff
@@ -37,7 +37,7 @@ export class MdButton {
   /** Whether a mousedown has occurred on this element in the last 100ms. */
   isMouseDown: boolean = false;
 
-  setClassList() {return `md-${this.color}`;}
+  setClassList() {return `md-${this.color}`; }
```
hansl: There should be a space before the `return` as well.

tslint.json
```diff
@@ -20,7 +20,7 @@
     "no-bitwise": true,
     "no-shadowed-variable": true,
     "no-unused-expression": true,
-    "no-unused-variable": true,
+    "no-unused-variable": [true, {"ignore-pattern": "ddescribe|iit|.*_$"}],
```
hansl: I don't think we need it ignore `ddescribe` and `iit`. They shouldn't be imported anyway. @jelbourn any opinion?

src/components/button/button.ts
d3viant0ne: This can not currently be enforced, there is an enhancement tslint issue open that would eventually cover this. I fixed the style issue

tslint.json
jelbourn: This was actually my suggestion. From a dev experience perspective, you frequently find yourself wanting to swap stuff between `describe` and `ddescribe` / `it` and `iit` (I even have a [WebStorm plugin](https://plugins.jetbrains.com/plugin/7233?pr=idea) that does it). If you have to update the imports every time to swap these, it gets annoying. 

This doesn't have any cost for the library itself since these are only in unit tests.

src/core/annotations/one-of.ts
```diff
@@ -27,8 +27,8 @@ function OneOfFactory(values: any[]) {
         }
       }
     });
-  }
+  };
 }
 
 
-export const OneOf = OneOfFactory;
+export const oneOf = oneOfFactory;
```
jelbourn: Hmm.... I'm not a fan of the camelCase decorator. @hansl what do you think?

d3viant0ne: @jelbourn Already discussing other options with Hans. Nobody is a fan of it, to include me.

hansl: Following Angular2, it should be `OneOf`.

d3viant0ne: @hansl - resolved with 4b830ec

hansl:
Hey! Glad to see this is really close to being done!

The only thing remaining is to squash your commits together into a single commit. You can do this by running:

```
git rebase -i master
```
And then picking the first commit you want to keep (might be the first) and squashing all the others.

If you need help I'll be away on gitter but available.

d3viant0ne:
I can squash with the best of them, be done in a sec.

jelbourn:
LGTM

Issue #128, opened 2016-03-03
Collect code coverage


jelbourn:
We should collect and surface our unit test code coverage. While a high coverage does not necessarily tell you much about the quality of your tests, a _low_ coverage does tell you that something is amiss.

d3viant0ne:
@jelbourn - I'll pick this one up.

I assume I should work out the particulars with @hansl again?

jelbourn:
@d3viant0ne sounds good! If @hansl is willing to chat about it on gitter that works for me, but we can also do the discussion here. 

As a starting point, I'm okay with just printing out the coverage as part of the Travis jobs. Then, as a follow-up, we can send that data off somewhere (probably some firebase instance) and incorporate it into a dashboard. 

As far as actual implementation goes, I'm not sure what the best avenue will be. The `karma-coverage` plugin uses the instanbul tool, but by itself doesn't support TypeScript. So there's exploration to happen in determining which tools are viable.

hansl:
@d3viant0ne 

Supporting TypeScript in karma with coverage is the problematic part of this plan, actually. Otherwise it should be straightforward. There is a few results for searching "code coverage karma typescript" on Google, but nothing out of the box. It is likely that we will need a custom made solution. For now, making a personal repo (or even just adding files to Material2) and using that should be great. That would be super useful for other TypeScript developers!

I'm not hanging often on gitter so let me know what would be a good time and I'll try to show up.

d3viant0ne:
@hansl - I just did this for Minko's seed project so it's definitely possible in some form. It really just depends if the less than perfect solution is acceptable or if we would rather wait for a home grown version. 
I'll take a run at it, get it executing in travis and we can discuss.

And don't worry about Gitter, I imagine you get spammed with question if your status goes green.

d3viant0ne:
@hansl - I've done TypeScript coverage using [remap-istanbul](https://www.npmjs.com/package/remap-istanbul) on a few projects but the implementation has always been very project specific.

My question is .. Do we want to do this in two parts?

**Phase 1:** Simple implementation of karma-coverage outputting to the command line when running "npm test"

This I could PR right now, karma doesn't exactly make coverage reporting hard :) 
There is also a feature in remap-istanbul#master related to relative paths I would need to make it work.

**Phase 2:** Tailored, more complex implementation using remap-istanbul for dev TS coverage which also reports coverage in CI to a 3rd Party service && comments on the pull request.

[...]

PR #137, opened 2016-03-09
build(coverage): enable dev js code coverage

d3viant0ne:
Phase 1 of #128 - JavaScript coverage for local development
  
Adds the following npm devDependencies
    - karma-coverage
    - rimraf
  Adds the following npm scripts both convienence items
    - postinstall: installs typings & runs npm prune
    - reinstall: cleans cache, generated content, node_modules & runs reinstall
  Enables JavaScript coverage reporting with multiple outputs
    - Reports code coverage at the commandline
    - Generates an HTML coverage report
    - Generates a coverage.json for future coverage history

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "build": "ng build",
     "dartanalyzer": "cd dist/dart && pub install && cd ../.. && ts-node scripts/ci/dart_analyzer",
     "demo-app": "cd src && ng serve",
+    "reinstall": "git clean -dfx -e '.*'",
     "test": "karma start test/karma.conf.js",
     "tslint" : "tslint -c tslint.json 'src/**/*.ts'",
     "typings": "typings install --ambient"
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "karma": "^0.13.15",
     "karma-browserstack-launcher": "^0.1.7",
     "karma-chrome-launcher": "^0.2.1",
+    "karma-coverage": "^0.5.5",
     "karma-dart": "^0.3.0",
     "karma-firefox-launcher": "^0.1.7",
     "karma-jasmine": "^0.3.6",
--- a/test/karma.config.ts
+++ b/test/karma.config.ts
@@ -13,6 +13,7 @@ export function config(config) {
     frameworks: ['jasmine'],
     plugins: [
       require('karma-jasmine'),
+      require('karma-coverage'),
       require('karma-browserstack-launcher'),
       require('karma-sauce-launcher'),
       require('karma-chrome-launcher'),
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -11,6 +11,8 @@
     "build": "ng build",
     "dartanalyzer": "cd dist/dart && pub install && cd ../.. && ts-node scripts/ci/dart_analyzer",
     "demo-app": "cd src && ng serve",
+    "postinstall": "typings install  --ambient && npm prune",
+    "reinstall": "npm cache clean && rimraf coverage dist typings && rimraf node_modules && npm install",
```
jelbourn: @hansl I think you had some opinions on `postinstall` scripts?

test/karma.config.ts
```diff
@@ -89,7 +101,7 @@ export function config(config) {
     browserNoActivityTimeout: 100000,
     browsers: ['Chrome'],
 
-    singleRun: false
+    singleRun: true
```
jelbourn: Why change to single-run?

package.json
hansl: We cannot have any `postinstall`/`reinstall` scripts as we still don't have a publishing story. So I would remove `postinstall` and `reinstall` until we know if we can use them. Or @jelbourn maybe it's okay for now since nobody can `npm install material2` (yet).

test/karma.config.ts
d3viant0ne: Process never gets past the tests and thus won't report coverage.

IMO it should really be split into test & test-watch

package.json
hansl: And instead of running `rimraf`, just do a `git clean -dfx`.

hansl: After discussing, keep the `postinstall`/`reinstall` scripts. When we do the publishing process we'll revisit this decision.

Replace `rimraf` (and remove the dependency) by `git clean -dfx`. This will remove everything that's not part of the git repository, even cached files for everything. This is okay for a `reinstall`, which is the same as rebuilding everything anyway. Alternatively, we could do a `git clean -dfx -e '.*'` (single quotes important) to keep dotfiles around (such as `.idea` and `.pub`).

test/karma.config.ts
```diff
@@ -57,13 +58,24 @@ export function config(config) {
     customLaunchers: customLaunchers,
 
     exclude: [],
-    preprocessors: {},
-    reporters: ['dots'],
+    preprocessors: {
+      'dist/**/!(*spec).js': ['coverage']
+    }
+    reporters: ['dots', 'coverage'],
```
hansl: I _think_ we force `--reporters=dots` on the command line in `build-and-test.sh`.

d3viant0ne: build-and-test.sh is only run in CI. The coverage currently is not.

hansl: Just something to keep in mind.

d3viant0ne: It was something I was going to work out with you before I PR the typescript coverage, one of the two will have to change.

package.json
d3viant0ne: Resolved in c7c2799

PR #146, opened 2016-03-10
feat(): add toolbar component

DevVersion:
--- WIP --- 

- Restricted Flexbox Usage
- Multi Line Support
- Custom CSS Support
- Input for Theme Color

The design doc isn't finished yet, but as discussed with @jelbourn, I'll submit that PR, so I can retrieve some feedback.

Open Points:
- [x] Item Alignment in the Toolbar 

PS: Issue intentionally not referenced yet.

image

### DIFF
--- a/src/components/toolbar/README.md
+++ b/src/components/toolbar/README.md
@@ -1,0 +1,68 @@
+# MdToolbar
+`MdToolbar` is a vertical aligned bar, which can hold the application title or actions.
+
+### Screenshots
+![Preview](https://cloud.githubusercontent.com/assets/4987015/13727769/6d952c78-e900-11e5-890a-ccfd46996812.png)
+
+## `<md-toolbar>`
+### Bound Properties
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `color` | `"primary" | "accent" | "warn"` | The color palette for the toolbar |
+
+### Notes
+The `md-toolbar` component will use by default the background palette.
+
+### Examples
+A basic toolbar would have the following markup.
+```html
+<md-toolbar [color]="myColor">
+  <span>My Application Title</span>
+</md-toolbar>
+```
+
+Toolbars can also have multiple rows.<br/>
+Multiple rows inside of a `md-toolbar` can be added by appending as many as needed `<md-toolbar-row>` elements.
+
+```html
+<md-toolbar [color]="myColor">
+  <span>First Row</span>
+  
+  <md-toolbar-row>
+    <span>Second Row</span>
+  </md-toolbar-row>
+  
+  <md-toolbar-row>
+    <span>Third Row</span>
+  </md-toolbar-row>
+</md-toolbar>
+```
+
+### Alignment
+The alingment inside of a toolbar row can be easily done by using the flexbox layout.<br/>
+For example, the following markup aligns the items on the `right`.
+
+Custom HTML
+```html
+<md-toolbar color="primary">
+  <span>Application Title</span>
+  
+  <!-- This fills the remaining space of the current row -->
+  <span class="example-fill-remaining-space"></span>
+  
+  <span>Right Aligned Text</span>
+</md-toolbar>
+```
+
+Custom SCSS
+```scss
+.example-fill-remaining-space {
+  // This fills the remaining space, by using flexbox. 
+  // Every toolbar row uses a flexbox row layout.
+  flex: 1 1 auto;
+}
+```
+
+Output
+![image](https://cloud.githubusercontent.com/assets/4987015/13730760/0864894e-e959-11e5-9312-7f3cb990d80a.png)
--- a/src/components/toolbar/toolbar.html
+++ b/src/components/toolbar/toolbar.html
@@ -1,0 +1,6 @@
+<div class="md-toolbar-layout">
+  <md-toolbar-row>
+    <ng-content></ng-content>
+  </md-toolbar-row>
+  <ng-content select="md-toolbar-row"></ng-content>
+</div>
--- a/src/components/toolbar/toolbar.scss
+++ b/src/components/toolbar/toolbar.scss
@@ -1,0 +1,57 @@
+@import "variables";
+@import "mixins";
+@import "default-theme"; //TODO: remove that soon.
+
+$md-toolbar-min-height: 64px !default;
+$md-toolbar-font-size: 20px !default;
+$md-toolbar-padding: 16px !default;
+
+@mixin toolbar-theme($palette) {
+  background: md-color($palette);
+  color: md-color($palette, default-contrast);
+}
+
+:host {
+  display: flex;
+  box-sizing: border-box;
+
+  width: 100%;
+  min-height: $md-toolbar-min-height;
+
+  // Font Styling
+  font-size: $md-toolbar-font-size;
+  font-weight: 400;
+  font-family: $md-font-family;
+
+  padding: 0 $md-toolbar-padding;
+
+  flex-direction: column;
+
+  background: md-color($md-background, app-bar);
+  color: md-color($md-foreground, text);
+
+  &.md-primary {
+    @include toolbar-theme($md-primary);
+  }
+
+  &.md-accent {
+    @include toolbar-theme($md-accent);
+  }
+
+  &.md-warn {
+    @include toolbar-theme($md-warn);
+  }
+
+  md-toolbar-row {
+    display: flex;
+    box-sizing: border-box;
+
+    width: 100%;
+    height: $md-toolbar-min-height;
+
+    // Flexbox Vertical Alignment
+    flex-direction: row;
+    align-items: center;
+  }
+
+}
--- a/src/components/toolbar/toolbar.ts
+++ b/src/components/toolbar/toolbar.ts
@@ -1,0 +1,42 @@
+import {
+  Component,
+  ChangeDetectionStrategy,
+  Input
+} from 'angular2/core';
+import {Renderer} from 'angular2/core';
+import {ElementRef} from 'angular2/core';
+
+@Component({
+  selector: 'md-toolbar',
+  templateUrl: './components/toolbar/toolbar.html',
+  styleUrls: ['./components/toolbar/toolbar.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MdToolbar {
+
+  private _color: string;
+
+  constructor(private elementRef: ElementRef, private renderer: Renderer) { }
+
+  @Input()
+  get color(): string {
+    return this._color;
+  }
+
+  set color(value: string) {
+    this._updateColor(value);
+  }
+
+  _updateColor(newColor: string) {
+    if (this._color != null && this._color != '') {
+      this.renderer.setElementClass(this.elementRef.nativeElement, `md-${newColor}`, false);
+    }
+
+    if (newColor != null && newColor != '') {
+      this.renderer.setElementClass(this.elementRef.nativeElement, `md-${newColor}`, true);
+    }
+
+    this._color = newColor;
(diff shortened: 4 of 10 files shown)


src/components/toolbar/toolbar.html
```diff
@@ -0,0 +1 @@
+<ng-content></ng-content>
```
jelbourn: When the template of a component is just `<ng-content>`, it can be a `@Directive` (no template) instead of a `@Component`

src/components/toolbar/toolbar.scss
```diff
@@ -0,0 +1,66 @@
...
+
+  &.md-accent {
+    background: md-color($md-accent);
+    color: md-color($md-accent, default-contrast);
+  }
+
+  &.md-warn {
+    background: md-color($md-warn);
+    color: md-color($md-warn, default-contrast);
+  }
```
kara: This primary/accent/warn section looks like it could possibly be refactored into a function that accepts the palette as an arg

src/components/toolbar/toolbar.scss
```diff
@@ -0,0 +1,66 @@
+@import "variables";
+@import "mixins";
+@import "default-theme"; //TODO: remove that soon.
+
+$toolbar-min-height: 64px !default;
+$toolbar-font-size: 20px !default;
+$toolbar-padding: 16px !default;
```
jelbourn: Prefix variables with `md-`

src/components/toolbar/toolbar.html
DevVersion: :+1: Thanks good idea.

src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,38 @@
...
+})
+export class MdToolbar {
+
+  private color_: string;
+
+  constructor(private elementRef: ElementRef, private renderer: Renderer) {}
+
+  set color(value: string) {
+    if (this.color_) {
+      this.renderer.setElementClass(this.elementRef.nativeElement, 'md-' + this.color_, false);
```
jelbourn: You should be able to pass the `elementRef` and not the `nativeElement`, no?

src/demo-app/toolbar/toolbar-demo.scss
```diff
@@ -0,0 +1,8 @@
+:host {
```
jelbourn: Prefix and demo css with `demo-`

src/components/toolbar/toolbar.ts
DevVersion: This caused problems with the BrowserDOMAdapater for me. But I'll recheck.

src/demo-app/toolbar/toolbar-demo.html
```diff
@@ -0,0 +1,63 @@
...
+    <span>Default Toolbar</span>
+
+    <span class="md-fill"></span>
+
+    <i class="material-icons">code</i>
+  </md-toolbar>
+</p>
+
+<p>
+  <md-toolbar [color]="'primary'">
```
jelbourn: Doesn't need to be a binding; it can just be ` color="primary" `

src/components/toolbar/toolbar.ts
zoechi: This changed in beta.1 https://github.com/angular/angular/blob/master/CHANGELOG.md#200-beta1-catamorphic-involution-2016-01-08 (last line)

> renderer.setElementProperty, ..setElementStyle, ..setElementAttribute now take a native element instead of an ElementRef
Renderer interface now operates on plain native nodes, instead of RenderElementRefs or RenderViewRefs

DevVersion: Thanks @zoechi for the point to the changelog :+1:

src/components/toolbar/toolbar.scss
```diff
@@ -0,0 +1,66 @@
...
+    color: md-color($md-accent, default-contrast);
+  }
+
+  &.md-warn {
+    background: md-color($md-warn);
+    color: md-color($md-warn, default-contrast);
+  }
+
+  // This allows the user to dynamically apply different rows for the toolbar.
+  &.md-toolbar-rows {
```
jelbourn: I think this could be a little more intelligent. Something like...

```html
<!-- basic toolbar -->
<md-toolbar>
  <!-- buttons and labels -->
</md-toolbar>

<!-- multiple rows -->
<md-toolbar>
  <!-- stuff on the first row -->
  <md-toolbar-row>
    <!-- stuff on the second row -->
  </md-toolbar-row>
</md-toolbar>
```

@hansl @robertmesserle @kara what do you think?

src/demo-app/toolbar/toolbar-demo.html
DevVersion: Cool thanks :+1:

src/components/toolbar/toolbar.scss
DevVersion: Just to ping you - the desired behavior is now implemented.

src/components/toolbar/toolbar.scss
```diff
@@ -0,0 +1,59 @@
...
+
+    // Flexbox Vertical Alignment
+    flex-direction: row;
+    align-items: center;
+  }
+
+  // Fills the remaining space of the toolbar row.
+  .md-fill {
+    flex: 1 1 auto;
+  }
```
jelbourn: I'm tempted to say that we eliminate this class and just people apply the styles they want to apply. I don't want to end up with a giant bag of classes that can be applied to different components in different contexts.

src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,37 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MdToolbar {
+
+  private _color: string;
+
+  constructor(private elementRef: ElementRef, private renderer: Renderer) {}
+
+  set color(value: string) {
+    if (this._color) {
```
jelbourn: Extract this code out into a function called something like `_updateColor` and make the setter just a call to it.

jelbourn: Using the truithiness of values doesn't work in Dart, so you have to explicitly check for `== null` and `== ''`

src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,37 @@
...
+      this.renderer.setElementClass(this.elementRef.nativeElement, 'md-' + this._color, false);
+    }
+
+    if (value) {
+      this.renderer.setElementClass(this.elementRef.nativeElement, 'md-' + value, true);
```
jelbourn: Use template string expressions:
```ts
`md-${newColor}`
```

src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    this._color = value;
+  }
+
+  get color(): string {
```
jelbourn: Put the getter above the setter

src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  Component,
+  ChangeDetectionStrategy
+} from 'angular2/core';
+import {Renderer} from 'angular2/core';
+import {ElementRef} from 'angular2/core';
+
+@Component({
+  selector: 'md-toolbar',
+  inputs: ['color'],
```
jelbourn: Put an `@Input()` decorator on the getter instead of this config here,.

src/demo-app/toolbar/toolbar-demo.scss
```diff
@@ -0,0 +1,8 @@
+.demo-toolbar {
+  padding: 15px;
+
+  .toolbar-icon {
```
jelbourn: Add `demo-` prefix.

jelbourn:
Let a couple of comments and also added a screenshot to your PR description. Looking good! I like the API. 

Aside from the comments, it just needs ` README.md` file similar to the other components.

@robertmesserle can you also give this a look?

src/components/toolbar/toolbar.scss
DevVersion: Well, I'll remove that, but IMO there should be a class / element which fills the remaining space in the toolbar row. Otherwise the user has to apply the flex fill himself.

DevVersion:
@jelbourn Completed your suggestions and added a `README.md`.

src/components/toolbar/README.md
```diff
@@ -0,0 +1,59 @@
...
+| `color` | `"primary" | "accent" | "warn"` | The color palette for the toolbar |
+
+### Notes
+By the default the `md-toolbar` will use the background palette.
+
+Multiple rows inside of a `md-toolbar` can be added by appending as many as needed `<md-toolbar-row>` elements.
+
+### Example
+```html
+<md-toolbar [color]="myColor">
```
jelbourn: Start off with a simple example of the toolbar with just a single row (most common case), and then say something like "The toolbar also supports multiple rows" with the second example.

src/components/toolbar/README.md
```diff
@@ -0,0 +1,59 @@
...
+  </md-toolbar-row>
+  
+  <md-toolbar-row>
+    <span>Third Toolbar Row</span>
+  </md-toolbar-row>
+</md-toolbar>
+```
+
+### Alignment
+The `md-toolbar` component doesn't provide any API to align the items probably in the row.
```
jelbourn: I'd omit this line, and just go straight into saying that the toolbar content can be directly styles with flex, and then show some examples.

src/components/toolbar/README.md
```diff
@@ -0,0 +1,59 @@
...
+The alingment inside of a toolbar row can be easily done by using flexbox properties.<br/>
+For example, the following markup aligns the items on the `right`.
+
+Custom HTML
+```html
+<md-toolbar>
+  <span>Application Title</span>
+  
+  <!-- This fills the remaining space of the current row -->
+  <span class="demo-fill"></span>
```
jelbourn: Use `example-fill-remaining-space` here

src/components/toolbar/README.md
```diff
@@ -0,0 +1,59 @@
...
+  
+  <span>Right Alingned Text</span>
```
jelbourn: `Aligned`

src/components/toolbar/README.md
```diff
@@ -0,0 +1,59 @@
...
+</md-toolbar>
+```
```
jelbourn: It would be good to add a screenshot immediately after this example to show the result.

DevVersion: Yes, that would be a good idea.

src/components/toolbar/README.md
DevVersion: Thanks.

src/components/toolbar/README.md
DevVersion: Oh, feels a bit long, but it's okay.

src/components/toolbar/README.md
DevVersion: Just wanted to clarify that there's no public API by the component itself.

src/demo-app/toolbar/toolbar-demo.scss
```diff
@@ -0,0 +1,12 @@
+.demo-toolbar {
+  padding: 15px;
+
+  .demo-toolbar-icon {
+    padding: 0 14px 0 14px;
+  }
+
+  .demo-flex {
```
jelbourn: I'd called this `demo-fill-remaining`

DevVersion: Everything changed as desired.

jelbourn:
LGTM

PR #148, opened 2016-03-10
initial setup for screenshot tests

robertmesserle:
@jelbourn @hansl @kara

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Files ending with `.screenshot.png` will be stored using `git-lfs` rather than stored in the git repository
+*.screenshot.png filter=lfs diff=lfs merge=lfs -text
--- a/e2e/_screenshot.js
+++ b/e2e/_screenshot.js
@@ -1,0 +1,81 @@
+"use strict";
+
+const path = require('path');
+const child_process  = require('child_process');
+const mapnik = require('mapnik');
+
+const SHA = process.env.TRAVIS_COMMIT;
+const LFS_BASE_URL = 'https://media.githubusercontent.com/media/angular/material2';
+
+/**
+ * Generates a screenshot from the current state of the Protractor test and compares it to the
+ * previous stored screenshot.  If the screenshots do not match or if no existing screenshot is
+ * found, an error will be thrown.  In both cases, the new screenshot will be stored so that it can
+ * be added to git.
+ */
+class Screenshot {
+  /**
+   * @param {string} id A unique identifier used for the screenshot
+   */
+  constructor(id) {
+    this.id   = id;
+    this.path = path.resolve(__dirname, '..', 'screenshots', id + '.screenshot.png');
+    this.url  = `${LFS_BASE_URL}/${SHA}/screenshots/${encodeURIComponent(id)}.screenshot.png`;
+    browser.takeScreenshot().then(png => this.storeScreenshot(png));
+  }
+
+  /**
+   * Stores a local copy of the screenshot for future comparison
+   * @param {string} png The base64-encoded screenshot generated from the current browser state
+   */
+  storeScreenshot(png) {
+    console.info(`[STATUS] Generated new screenshot for "${this.id}"`);
+    this.png = mapnik.Image.fromBytes(new Buffer(png, 'base64'));
+    if (SHA) {
+      this.downloadFromGithub();
+    } else {
+      this.compareScreenshots();
+    }
+  }
+
+  /**
+   * Since we are using `git-lfs`, screenshots are not necessarily available within our local
+   * directory.  To get around this, we download the latest screenshot from Github.
+   */
+  downloadFromGithub() {
+    console.info(`[STATUS] Downloading screenshot from Github: ${this.url} => ${this.path}`);
+    child_process.execSync(`curl ${this.url} > "${this.path}"`);
+    this.compareScreenshots();
+  }
+
+  /**
+   * Compares the generated screenshot to the existing screenshot.  If it does not match, an error
+   * will be thrown.
+   */
+  compareScreenshots() {
+    console.info(`[STATUS] Comparing screenshots`);
+    try {
+      let referenceScreenshot = mapnik.Image.open(this.path);
+      this.overwriteExistingScreenshot();
+      if (referenceScreenshot.compare(this.png)) {
+        throw new Error(`screenshot "${this.id}" has changed.`);
+      } else {
+        console.info('[STATUS] Screenshot has not changed');
+      }
+    } catch (e) {
+      console.info(`[STATUS] No reference screenshot found`);
+      this.overwriteExistingScreenshot();
+      throw new Error(`screenshot "${this.id}" was not found.`);
+    }
+  }
+
+  /**
+   * Replaces the existing screenshot with the newly generated one.
+   */
+  overwriteExistingScreenshot() {
+    console.info(`[STATUS] Saving new screenshot`);
+    this.png.save(this.path);
+  }
+}
+
+module.exports = (id) => new Screenshot(id);
--- a/scripts/sauce/sauce_connect_setup.sh
+++ b/scripts/sauce/sauce_connect_setup.sh
@@ -12,7 +12,6 @@ set -e -o pipefail
 # before_script:
 #   - curl https://gist.github.com/santiycr/5139565/raw/sauce_connect_setup.sh | bash
 
-CONNECT_URL="https://saucelabs.com/downloads/sc-4.3.11-linux.tar.gz"
 CONNECT_DIR="/tmp/sauce-connect-$RANDOM"
 CONNECT_DOWNLOAD="sc-latest-linux.tar.gz"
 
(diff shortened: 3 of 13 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.screenshot.png filter=lfs diff=lfs merge=lfs -text
```
jelbourn: Add a comment explaining what this does / how the attributes are used.

robertmesserle: Hm, this is a generated file.  Any idea how comments work in here? =p

hansl: Probably same as all `git` files; `#`.

robertmesserle: Ah, looks like just standard `#comment` syntax

e2e/_screenshot.js
```diff
@@ -0,0 +1,53 @@
...
+var SHA = process.env.TRAVIS_COMMIT;
+
+function screenshot(id) {
+  var screenshotPath = path.resolve(__dirname, '..', 'screenshots', id + '.screenshot.png');
+  var screenshotUrl = `https://media.githubusercontent.com/media/angular/visualdiff-dev/${ SHA }/screenshots/${ encodeURIComponent(id) }.screenshot.png`;
+  var newScreenshot;
+
+  browser.takeScreenshot().then(handleNewScreenshot);
+
+  function handleNewScreenshot(png) {
```
jelbourn: Do these functions have to be nested inside of `screenshot` ?

e2e/_screenshot.js
```diff
@@ -0,0 +1,53 @@
+module.exports = screenshot;
+
+var path = require('path');
+var child_process  = require('child_process');
+var mapnik = require('mapnik');
+
+var SHA = process.env.TRAVIS_COMMIT;
```
jelbourn: Add comment saying what you use the `SHA` for.

e2e/_screenshot.js
robertmesserle: Nope.  This was shorthand while developing, but I should probably organize this into a class.

e2e/_screenshot.js
```diff
@@ -0,0 +1,53 @@
...
+
+function screenshot(id) {
+  var screenshotPath = path.resolve(__dirname, '..', 'screenshots', id + '.screenshot.png');
+  var screenshotUrl = `https://media.githubusercontent.com/media/angular/visualdiff-dev/${ SHA }/screenshots/${ encodeURIComponent(id) }.screenshot.png`;
```
jelbourn: Could you shorten this line by breaking the `https://media.githubusercontent.com/media/angular/visualdiff-dev/` into a separate const near the top of the file?

e2e/_screenshot.js
jelbourn: That would be even better

e2e/_screenshot.js
```diff
@@ -0,0 +1,53 @@
...
+      overwriteExistingScreenshot();
+      if (goldScreenshot.compare(newScreenshot)) {
+        throw new Error('screenshot "' + id + '" has changed.');
+      } else {
+        status('Screenshot has not changed');
+      }
+    } catch (e) {
+      status('Gold screenshot not found');
+      overwriteExistingScreenshot();
+      throw new Error('screenshot "' + id + '" was not found.');
```
jelbourn: Turn into a template string

e2e/_screenshot.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+}
+
+function status(msg) {
+  console.info('STATUS', msg);
+}
```
jelbourn: Get rid of this function and just use `console.log` directly?

scripts/sauce/sauce_connect_setup.sh
```diff
@@ -21,6 +20,11 @@ CONNECT_STDOUT="$LOGS_DIR/sauce-connect.stdout"
 CONNECT_STDERR="$LOGS_DIR/sauce-connect.stderr"
 
 # Get Connect and start it
+if [ `uname -s` = "Darwin" ]; then
```
jelbourn: wat

e2e/_screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+ * be added to git.
+ */
+class Screenshot {
+  /**
+   * @param {string} id A unique identifier used for the screenshot
+   */
+  constructor(id) {
+    this.id   = id;
+    this.path = path.resolve(__dirname, '..', 'screenshots', id + '.screenshot.png');
+    this.url  = `${ LFS_BASE_URL }/${ SHA }/screenshots/${ encodeURIComponent(id) }.screenshot.png`;
```
jelbourn: As discussed, annihilate the whitespace

e2e/_screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Stores the screenshot
+   * @param {image} png The screenshot generated from the current browser state
+   */
+  handleScreenshot(png) {
+    console.info(`[STATUS] Generated new screenshot for "${ this.id }"`);
+    this.png = mapnik.Image.fromBytes(new Buffer(png, 'base64'));
+    if (SHA) this.downloadFromGithub();
```
jelbourn: Add braces

e2e/_screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+  constructor(id) {
+    this.id   = id;
+    this.path = path.resolve(__dirname, '..', 'screenshots', id + '.screenshot.png');
+    this.url  = `${ LFS_BASE_URL }/${ SHA }/screenshots/${ encodeURIComponent(id) }.screenshot.png`;
+    browser.takeScreenshot().then(png => this.handleScreenshot(png));
+  }
+
+  /**
+   * Stores the screenshot
+   * @param {image} png The screenshot generated from the current browser state
```
jelbourn: Is the type actually a `mapnik.Image` ?

e2e/_screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+   */
+  handleScreenshot(png) {
```
jelbourn: Does it make sense to just call this function `storeScreenshot` if that's all it's doing?

scripts/sauce/sauce_connect_setup.sh
```diff
@@ -21,6 +20,11 @@ CONNECT_STDOUT="$LOGS_DIR/sauce-connect.stdout"
 CONNECT_STDERR="$LOGS_DIR/sauce-connect.stderr"
 
 # Get Connect and start it
+if [ `uname -s` = "Darwin" ]; then # If user is running Mac, download Mac version
```
jelbourn: I had never seen "Darwin" in this context before. Can we include a link?
https://en.wikipedia.org/wiki/Darwin_(operating_system)

jelbourn:
LGTM aside from remaining comments

e2e/_screenshot.js
robertmesserle: Pretty aggressive phrasing there.  You must mean business!

e2e/_screenshot.js
robertmesserle: Yea, I'd like something to distinguish between storing a local copy of the screenshot (what this is doing) and saving the screenshot (what `overwriteExistingScreenshot` is doing).

scripts/sauce/sauce_connect_setup.sh
```diff
@@ -20,7 +19,15 @@ CONNECT_LOG="$LOGS_DIR/sauce-connect"
 CONNECT_STDOUT="$LOGS_DIR/sauce-connect.stdout"
 CONNECT_STDERR="$LOGS_DIR/sauce-connect.stderr"
 
-# Get Connect and start it
+# Get the appropriate URL for downloading Sauce Connect
+if [ `uname -s` = "Darwin" ]; then
```
hansl: Wanna do Windows too for external contributors, or just ignore it until the problem comes?

e2e/_screenshot.js
```diff
@@ -0,0 +1,81 @@
+"use strict";
```
hansl: Any reason this is ES6 rather than TS? Just curious.

scripts/sauce/sauce_connect_setup.sh
jelbourn: We do need to support Windows contributors, but I'm okay with omitting it in this PR if we file an issue for it. I tried doing `ng serve` on my home Windows machine today and that didn't work already, so there's higher-priority Windows stuff.

scripts/sauce/sauce_connect_setup.sh
```diff
@@ -28,6 +35,7 @@ mkdir sauce-connect
 tar --extract --file=$CONNECT_DOWNLOAD --strip-components=1 --directory=sauce-connect > /dev/null
 rm $CONNECT_DOWNLOAD
 
+
```
hansl: ?

scripts/sauce/sauce_connect_setup.sh
robertmesserle: Yea, the issue with Windows support is that you have to handle each scenario separately.  For example, we would have to check for `cigwin` as well as any other terminal being used.  I say we open an issue for it and deal with it later.

e2e/_screenshot.js
robertmesserle: We decided to switch over to TS later rather than further delaying this feature.  Need to address the node typing issues with TS first.

scripts/sauce/sauce_connect_setup.sh
robertmesserle: Yea, pretty sure that was not intentional. =p

PR #153, opened 2016-03-12
build(): add gulpfile to inline resources.

jelbourn:
R: @hansl 

I had to change the `` ` `` in `sidenav.scss` to `'` because the html inliner is not super smart.

It's not the ideal solution, but at least it's pretty tiny and only needs to be used when we do a release.


hansl:
Wait, does this thing simply copy-paste the content without anything else?

jelbourn:
@hansl pretty much, yeah. Did you expect something else?

PR #154, opened 2016-03-12
chore(): move `dir` directive to core

jelbourn:
R: @hansl 

Move the `dir` directive under `core/` so that our releases are only concerned with `core/` and `components/`.


hansl:
LGTM.

PR #155, opened 2016-03-12
build(): make FF beta optional until SL fixes it.

jelbourn:
R: @hansl


hansl:
Lgtm

hansl:
On a tangential note; if we're just going to throw Optional tests like these, we should disable optional tests entirely. It's just noise.

jelbourn:
That's a good point; we should have periodic check of the optional to see whether we should drop them or fix them. Maybe it's something we can look into during milestone planning.

PR #157, opened 2016-03-13
feat(portal): add DomPortalHost

jelbourn:
R: @hansl @kara 

All of the stuff in `platform/` is copied from `angular/angular` and is, for our purposes, the facades necessary for DOM manipulation. (I also had to get rid of implicit `any` and make it pass tslint... ugh). I talked with Tobias and Misko about how much this sucks, and their answer is https://github.com/angular/angular/issues/7561. For now copying this stuff unblocks progress.

`DomPortalHost` is the last part of the `portal/` code needed for overlays. It's will be used when opening an overlay in the attaching of the overlay element to the document body.

Setting `locals` for a `TemplatePortal` attached to a `DomPortalHost` does not work; I spent quite a while trying to figure out why and came up short. I'm going to follow up with Tobias on that, but in the meantime it's not a blocking issue.

### DIFF
--- a/src/core/portal/dom-portal-host.ts
+++ b/src/core/portal/dom-portal-host.ts
@@ -1,0 +1,54 @@
+import {DynamicComponentLoader, AppViewManager, ComponentRef} from 'angular2/core';
+import {BasePortalHost, ComponentPortal, TemplatePortal} from './portal';
+import {MdComponentPortalAttachedToDomWithoutOriginException} from './portal-exceptions';
+
+
+/**
+ * A PortalHost for attaching portals to an arbitrary DOM element outside of the Angular
+ * application context.
+ *
+ * This is the only part of the portal core that directly touches the DOM.
+ */
+export class DomPortalHost extends BasePortalHost {
+  constructor(
+      private _hostDomElement: Element,
+      private _componentLoader: DynamicComponentLoader,
+      private _viewManager: AppViewManager) {
+    super();
+  }
+
+  /** Attach the given ComponentPortal to DOM element using the DynamicComponentLoader. */
+  attachComponentPortal(portal: ComponentPortal): Promise<ComponentRef> {
+    if (portal.origin == null) {
+      throw new MdComponentPortalAttachedToDomWithoutOriginException();
+    }
+
+    return this._componentLoader.loadNextToLocation(portal.component, portal.origin).then(ref => {
+      this._hostDomElement.appendChild(ref.hostView.rootNodes[0]);
+      this.setDisposeFn(() => ref.dispose());
+      return ref;
+    });
+  }
+
+  attachTemplatePortal(portal: TemplatePortal): Promise<Map<string, any>> {
+    let viewContainer = this._viewManager.getViewContainer(portal.templateRef.elementRef);
+    let viewRef = viewContainer.createEmbeddedView(portal.templateRef);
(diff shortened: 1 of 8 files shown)


src/core/portal/dom-portal-host.ts
```diff
@@ -0,0 +1,56 @@
...
+      private _hostDomElement: Element,
+      private _componentLoader: DynamicComponentLoader,
+      private _viewManager: AppViewManager) {
+    super();
+  }
+
+  /** Attach the given ComponentPortal to DOM element using the DynamicComponentLoader. */
+  attachComponentPortal(portal: ComponentPortal): Promise<ComponentRef> {
+    if (portal.origin == null) {
+      throw new BaseException(
```
hansl: Nit; I don't think we ever had that discussion; I'd prefer having an `MdNullPortalOriginException`.

src/core/portal/dom-portal-host.ts
```diff
@@ -0,0 +1,56 @@
...
+      this.setDisposeFn(() => ref.dispose());
+      return ref;
+    });
+  }
+
+  attachTemplatePortal(portal: TemplatePortal): Promise<Map<string, any>> {
+    let viewContainer = this._viewManager.getViewContainer(portal.templateRef.elementRef);
+    let viewRef = viewContainer.createEmbeddedView(portal.templateRef);
+    // TODO(jelbourn): locals don't currently work with DomPortalHost; investigate whether there
+    // is a bug in Angular.
```
hansl: Have you tried to see if all locals actually get set properly? This is peculiar; looks like it should work.

src/core/portal/dom-portal-host.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * A PortalHost for attaching portals to an arbitrary DOM element outside of the Angular
+ * application context.
+ *
+ * This is the only part of the portal core that directly touches the DOM.
+ */
+export class DomPortalHost extends BasePortalHost {
+  constructor(
+      private _hostDomElement: Element,
```
hansl: Would it be cleaner to get an `ElementRef` here? Alternatively, I'm wondering if you could get the `Renderer` and use `projectNodes()`.

src/core/portal/dom-portal-host.ts
jelbourn: The local gets set properly, but Angular doesn't seem to _apply_ the local. I'm going to dig more into it once I have Victor or Tobias around.

src/core/portal/dom-portal-host.ts
jelbourn: I went down that road with Tobias, and the advice from him is to just do the DOM stuff where you need it. Eventually Angular should provide a common way to do marshalling when web-workers are a thing. As it is now, an instance of `Renderer` is always tied to a directive and is meant for manipulating its view.

src/core/portal/dom-portal-host.ts
jelbourn: We can chat about it on Monday. I don't yet see the end-of-the-day value in having different error types, but I'm open to convincing.

hansl:
Okay then LGTM with the nit that we should have a proper discussion about using Exception classes rather than always throwing BaseException.

jelbourn:
@hansl added exceptions.

PR #158, opened 2016-03-13
style(demo): prefix css w/ demo- & use roboto

jelbourn:
R: @kara @hansl 

The motivation behind the `demo-` prefix is so that people can look at the demos and `Inspect element` and immediately see which styles are from the component itself and which are part of the demo. This was a big problem in material1 where people would have a hard time discerning between the two.


kara:
LGTM

PR #160, opened 2016-03-14
test(e2e): adds readyfile to travis config

robertmesserle:
Currently, tests relying on `sauce-connect` are flaky, and it seems to be caused by the script continuing before SauceLabs is ready.


jelbourn:
LGTM

PR #161, opened 2016-03-14
component: MdInput

hansl:


src/core/annotations/field-value.ts
```diff
@@ -0,0 +1,37 @@
...
+ *   <component myField>
+ * or:
+ *   <component myField="">
+ */
+function booleanFieldValueFactory() {
+  return function booleanFieldValueMetadata(target: any, key: string): void {
+    const defaultValue = target[key];
+
+    // Use a fallback if Symbol isn't available.
+    const localKey = isPresent(Symbol) ? Symbol(key) : `@@$_${key}`;
```
mhevery: adding special characters such as `@@` makes the object into slow object. Better to use `__private_symbol_` or something like that.

hansl: Done. Firefox used to support `@@` as a prefix for symbols, thus my usage of this.

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+  selector: 'md-input',
+  templateUrl: 'components/input/input.html',
+  styleUrls: ['components/input/input.css'],
+  providers: [MdAria, MD_INPUT_CONTROL_VALUE_ACCESSOR],
+  directives: [MdAriaTarget],
+})
+export class MdInput implements ControlValueAccessor {
+  private _empty: boolean = true;
+  private _focused: boolean = false;
+  private _onTouched: () => void = () => {};
```
ThomasBurleson: Does ng2 provide a noop function like ng1 ?

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+  private _onChangeFn: (_: any) => void = (_: any) => {};
```
ThomasBurleson: ! WAT ?

src/core/utilities/aria-directives.ts
```diff
@@ -0,0 +1,43 @@
...
+import {BooleanFieldValue} from '../annotations/field-value';
+
+
+/**
+ * aria directive, handling aria parts of the standard.
+ */
+@Directive({
+  selector: '[aria-label], [aria-labelledby], [aria-disabled], [aria-required], [aria-invalid]'
+})
+export class MdAria implements AttributeTargetDirectiveInterface {
```
ThomasBurleson: Does this consider (or is it based upon) the functionality of the ng1 aria service ?

src/components/input/input.ts
hansl: What?

src/core/utilities/aria-directives.ts
hansl: No. This is not a service.

src/components/input/input.ts
ThomasBurleson: again... this is a noop... Perhaps a `noop` Symbol can be created and easily used ?

m0t0r: very sorry to interfere but angular2 has internal [noop function] (https://github.com/angular/angular/blob/master/modules/angular2/src/facade/lang.ts#L157)

src/components/input/input.ts
hansl: Done.

src/components/input/input.ts
hansl: Okay so after looking deeper, it _is_ invalid Dart to assign a `function(){}` to one that expects an argument. `ts2dart` though emits `dynamic` typing in this case, which remove type checking. I've updated the code to use `noop`, but note this might become invalid in the future.

src/components/input/README.md
```diff
@@ -0,0 +1,12 @@
+# mdInput
+
+Inputs are the basic input component of Material 2. The spec can be found [here](https://www.google.com/design/spec/components/text-fields.html).
+
+
+
+## Type
+
+At the time of writing this README, the `[type]` attribute is copied to the actual `<input>` element in the `<md-input>`. Therefore, the valid types and behavior of those are the ones supported by your browser.
+
```
zoechi: except `file`?

src/components/input/input.html
```diff
@@ -0,0 +1,29 @@
+<div class="md-wrapper" (click)="input.focus()">
```
jelbourn: `md-wrapper` doesn't really tell you anything

src/components/input/input.html
```diff
@@ -0,0 +1,29 @@
...
+  <div class="md-flex">
+    <prefix><ng-content select="[md-prefix]"></ng-content></prefix>
```
jelbourn: Prefix is not an HTML element; use a css class.
(same for suffix)

src/components/input/input.html
```diff
@@ -0,0 +1,29 @@
...
+
+    <label [attr.for]="'md-input-' + _inputId"
```
jelbourn: Store the entire ID on the class instead of doing the string concat twice in the template?

src/components/input/input.html
```diff
@@ -0,0 +1,29 @@
...
+           [class.md-empty]="empty" [class.md-focused]="focused"
+           [class.md-float]="floatingPlaceholder" [class.md-accent]="dividerColor == 'accent'">
+      {{placeholder}}
+      <span *ngIf="required">*</span>
+    </label>
+
+    <input #input aria-target
+           [class.md-end]="align == 'end'"
+           [attr.id]="'md-input-' + _inputId"
+           [disabled]="disabled" [required]="required"
```
jelbourn: One attr per line for scanability?

src/components/input/input.html
```diff
@@ -0,0 +1,29 @@
...
+           [attr.maxlength]="maxLength" [attr.type]="type"
```
jelbourn: `maxLength` is a real property on HTMLInputElement, so you don't need the `attr.`

src/components/input/input.html
```diff
@@ -0,0 +1,29 @@
...
+
+    <suffix><ng-content select="[md-suffix]"></ng-content></suffix>
+
+  </div>
+
+  <div class="md-underline" [class.md-disabled]="disabled">
+    <span [class.md-focused]="focused" [class.md-accent]="dividerColor == 'accent'"></span>
+  </div>
+  <div *ngIf="hintLabel != ''" class="md-hint">{{hintLabel}}</div>
+  <div *ngIf="_characterCounterLabel() != ''" class="md-hint md-right">{{_characterCounterLabel()}}</div>
```
jelbourn: Private expressions in templates won't work for dart.

src/components/input/input.scss
```diff
@@ -0,0 +1,153 @@
...
+    margin: 16px 0;
+  }
+  .md-flex {
+    display: inline-table;
+    flex-flow: column;
+    vertical-align: bottom;
+    width: 100%;
+  }
+  .md-flex > * {
+    display: table-cell;
```
jelbourn: wat

src/components/input/input.scss
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  input {
```
jelbourn: Can we do these with classes instead of styling the elements directly?

src/components/input/input.scss
```diff
@@ -0,0 +1,153 @@
...
+    transition: transform $swift-ease-out-duration $swift-ease-out-timing-function,
+                scale $swift-ease-out-duration $swift-ease-out-timing-function,
+                color $swift-ease-out-duration $swift-ease-out-timing-function;
+
+    &.md-empty {
+      visibility: visible;
+      cursor: text;
+    }
+
+    &.md-float:not(.md-empty), &.md-float.md-focused {
```
jelbourn: Add comment to explain what this state means

src/components/input/input.scss
```diff
@@ -0,0 +1,153 @@
...
+$md-input-underline-disabled-color: md-color($md-foreground, hint-text);
+$md-input-underline-focused-color: md-color($md-primary);
+
+
+:host {
+  display: inline-block;
+  position: relative;
+  font-family: $md-font-family;
+
+  .md-wrapper {
```
jelbourn: Could you add a brief description of what each class is responsible?

src/components/input/input.scss
```diff
@@ -0,0 +1,153 @@
...
+  .md-underline {
+    position: absolute;
+    height: 1px;
+    width: 100%;
+    margin-top: 4px;
+    border-top: 1px solid $md-input-underline-color;
+
+    &.md-disabled {
+      border-top: 0;
+      background-image: linear-gradient(to right, rgba(0,0,0,0.26) 0%, rgba(0,0,0,0.26) 33%, transparent 0%);
```
jelbourn: All of these background definitions seems like they could benefit from some explanation as to what's happening

src/components/input/input.scss
```diff
@@ -0,0 +1,153 @@
...
+      background-position: 0;
+      background-size: 4px 1px;
+      background-repeat: repeat-x;
+    }
+
+    > span {
```
jelbourn: Definitely a class here; encountering just `> span` is very hard to map back to what's actually being styled in the template.

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+
+
+const MD_INPUT_CONTROL_VALUE_ACCESSOR = CONST_EXPR(new Provider(
+    NG_VALUE_ACCESSOR, {
+      useExisting: forwardRef(() => MdInput),
+      multi: true
+    }));
+
+
+let uniqueId = 0;
```
jelbourn: `nextUniqueId` ?

src/components/input/input.html
hansl: Done.

src/components/input/input.html
hansl: done.

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+  @Input() hintLabel: string = '';
+  @Input() maxLength: number = -1;
+  @Input() placeholder: string;
+  @Input() type: string = 'text';
+
+  get value(): any { return this._value; };
+  @Input() set value(v: any) {
+    if (this._value !== v) {
+      this._value = v;
+      this._empty = v === undefined || v === null || (''+v == '');
```
jelbourn: You should be able to do
```ts
v == null || v == '';
```

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+export class MdInput implements ControlValueAccessor {
+  private _empty: boolean = true;
+  private _focused: boolean = false;
+  private _onTouched: () => void = noop;
+  private _onChangeFn: (_: any) => void = noop;
+  private _value: any = '';
+  private _inputId: number = uniqueId++;
+
+  get focused() { return this._focused; }
+  get empty() { return this._empty; }
```
jelbourn: If these don't have setters you don't need getters, just normal properties.

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+  providers: [MdAria, MD_INPUT_CONTROL_VALUE_ACCESSOR],
+  directives: [MdAriaTarget],
+})
+export class MdInput implements ControlValueAccessor {
+  private _empty: boolean = true;
+  private _focused: boolean = false;
+  private _onTouched: () => void = noop;
+  private _onChangeFn: (_: any) => void = noop;
+  private _value: any = '';
+  private _inputId: number = uniqueId++;
```
jelbourn: `md-input-${nextUniqueId++}`

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+@Component({
+  selector: 'md-input',
+  templateUrl: 'components/input/input.html',
+  styleUrls: ['components/input/input.css'],
+  providers: [MdAria, MD_INPUT_CONTROL_VALUE_ACCESSOR],
+  directives: [MdAriaTarget],
+})
+export class MdInput implements ControlValueAccessor {
+  private _empty: boolean = true;
+  private _focused: boolean = false;
```
jelbourn: `empty` and `focused` to `isEmpty` and `isFocused`?

jelbourn: Actually, why isn't `isEmpty()` a method? I don't see why you need a property for it since it can be computed from `value`.

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+  get value(): any { return this._value; };
+  @Input() set value(v: any) {
+    if (this._value !== v) {
+      this._value = v;
+      this._empty = v === undefined || v === null || (''+v == '');
+      this._onChangeFn(v);
+    }
+  }
+
+  @Output('change') change = new EventEmitter(false);
```
jelbourn: Doesn't `EventEmitter` need a generic?

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  @HostBinding('attr.align') private get _align(): any { return null; }
```
jelbourn: Personally I'm not a fan of singe-lining getters, which I find makes the code more difficult to scan. Especially in this case where the single line has the decorator, its parameter, access modifier, name, type, and the return value.

Also this always returns null- TODO?

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  /** @override */
```
jelbourn: Instead of the `override` comment, say 
```
Implemented as part of ControlValueAccessor.
```

src/components/input/input.ts
jelbourn: I don't see this event actually being emitted anywhere.

src/components/input/input.ts
jelbourn: Why use an `attr` instead of a class?

src/components/input/input.ts
```diff
@@ -0,0 +1,114 @@
...
+    this._focused = true;
+  }
+  onBlur() {
+    this._focused = false;
+    this._onTouched();
+  }
+
+  private _characterCounterLabel(): string {
+    if (this.characterCounter) {
+      return String(this.value || '').length + (this.maxLength > 0 ? ' / ' + this.maxLength : '');
```
jelbourn: `` `${value}`.length ``?

jelbourn: Does `X / Y` need to be different for different locales?

src/components/input/input.html
hansl: Done.

src/components/input/input.html
hansl: Done.

src/components/input/input.html
hansl: Actually maxLength is touchy. Done the rest.

src/components/input/input.html
hansl: Done.

src/components/input/input.scss
hansl: wat

src/core/utilities/aria-directives.ts
```diff
@@ -0,0 +1,43 @@
...
+  Directive,
+  ElementRef,
+  Input,
+  Renderer,
+} from 'angular2/core';
+import {AttributeTargetDirectiveInterface, MdAttributeTarget} from './attribute-target';
+import {BooleanFieldValue} from '../annotations/field-value';
+
+
+/**
```
jelbourn: I'm concerned about the performance implications of instantiating a directive for every single one of these ARIA attributes.

src/components/input/input.scss
hansl: Done.

src/core/utilities/aria-directives.ts
jelbourn: If we keep this (wrt concerns about performance), the name is too nonspecific. It would be something like `MdAriaAttributeForwarder`

src/core/utilities/attribute-target.ts
```diff
@@ -0,0 +1,26 @@
...
+  attributeMap: { [name: string]: any };
+}
+
+
+export class MdAttributeTarget<T extends AttributeTargetDirectiveInterface> {
+  public directive: T;
+  public element: ElementRef;
+  public renderer: Renderer;
+
+  ngAfterViewInit() {
```
jelbourn: I don't believe this will keep the attributes up to date when the bindings change.

src/demo-app/input/input-demo.scss
```diff
@@ -0,0 +1,23 @@
+@import 'default-theme';
+@import 'variables';
+
+
+.material-icons {
```
jelbourn: Prefix demo styles with `-demo`. This lets people inspect the elements in the demos and immediately nkow which styles are part of the component itself and which are just from the demo.

src/tsconfig.json.backup
```diff
@@ -0,0 +1,17 @@
+{
```
jelbourn: ?

src/tsconfig.jsonbackup
```diff
@@ -0,0 +1,17 @@
+{
```
jelbourn: ?

jelbourn:
@hansl finished my first pass, let me know when you're ready for another look

src/components/input/input.scss
hansl: Done.

src/components/input/input.scss
hansl: Done.

src/components/input/input.scss
hansl: All of them 1? ;) done.

src/components/input/input.scss
hansl: Done.

src/components/input/input.ts
hansl: Done.

src/components/input/input.ts
hansl: But truthiness is EVIL :(

src/components/input/input.ts
hansl: What do you mean by normal properties? Just `empty: string`? I don't want them to be publicly changeable, they're really readonly.

src/components/input/input.ts
hansl: It's not. Removed.

hansl: To answer your question; no.

src/components/input/input.ts
hansl: Added a comment to explain why this is needed.

src/components/input/input.ts
hansl: Done. Should we do that for every interface?

src/components/input/input.ts
hansl: Fixed, after offline discussion.

src/components/input/input.ts
jelbourn: The tao of `===` is overrated.

src/components/input/input.ts
hansl: done.

src/components/input/input.ts
hansl: Properties are the same but cleaner IMO. Why should everything be a method, particularly when there's no side effect?

Anyway, done.

src/core/utilities/aria-directives.ts
hansl: After offline discussion; removed these classes and inserted the bindings directly into the input component.

src/core/utilities/attribute-target.ts
hansl: Code has been removed.

src/tsconfig.json.backup
hansl: Removed.

src/tsconfig.jsonbackup
hansl: Not sure what happened there.

hansl:
@jelbourn ready for round #2. What's coming for me; adding unittests as we speak, adding README documentation, adding comments to MdInput.

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+})
+export class MdInput implements ControlValueAccessor, AfterContentInit, OnChanges {
+  private _focused: boolean = false;
+  private _onTouched: () => void = noop;
+  private _onChangeFn: (_: any) => void = noop;
+  private _value: any = '';
+
+  /**
+   * Aria related inputs.
+   */
```
jelbourn: Optional: I'm cool with condensing short block comments to 
```
/** Aria related inputs. */
```

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+    return this.placeholder != null || this._placeholderChild != null;
+  }
+
+  private _setValue(value: any): boolean {
+    if (this._value !== value) {
+      this._value = value;
+      return true;
+    } else {
+      return false;
+    }
```
jelbourn: It's odd that a `set` function returns a value. Why not just emit the event from here?
```ts
_setValue(newValue: any): void {
  if (newValue !== value) {
    this._value = newValue();
    emitChange();
  }
}
```

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+ */
+@Directive({
+  selector: 'md-hint',
+  host: {
+    '[class.md-right]': 'align == "end"',
+    '[class.md-hint]': 'true'
+  }
+})
+export class MdHint {
+  @Input() @OneOf(['start', 'end']) align: string;
```
jelbourn: Add comment for `align`

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+}
+
+
+/**
+ * <md-input> component.
```
jelbourn: Expand description to explain what behaviors it exhibits over a native `<input>`

src/components/input/input.ts
jelbourn: Also, though, I think the `Aria related inputs`, `Content directives`, and `Bindings` section comments don't actually tell you anything. The fields themselves are self-explanatory as it is.

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+  @Input('aria-labelledby') ariaLabelledBy: string;
+  @Input('aria-disabled') @BooleanFieldValue() ariaDisabled: boolean;
+  @Input('aria-required') @BooleanFieldValue() ariaRequired: boolean;
+  @Input('aria-invalid') @BooleanFieldValue() ariaInvalid: boolean;
+
+  /**
+   * Content directives.
+   */
+  @ContentChild(MdPlaceholder) private _placeholderChild: MdPlaceholder;
+  @ContentChildren(MdHint) private _hintChildren: QueryList<MdHint>;
```
jelbourn: You have have multiple hints?

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+  @Input() set value(v: any) {
+    if (this._setValue(v)) {
+      this._onChangeFn(v);
+    }
+  }
+
+  // This is to remove the `align` property of the `md-input` itself. Otherwise HTML5
+  // might place it as RTL when we don't want to. We still want to use `align` as an
+  // Input though, so we use HostBinding.
+  @HostBinding('attr.align') private get _align(): any { return null; }
```
jelbourn: Is this necessary, since we're just using any of the (deprecated) native `align` values?

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+  selector: 'md-input',
+  templateUrl: 'components/input/input.html',
+  styleUrls: ['components/input/input.css'],
+  providers: [MdAria, MD_INPUT_CONTROL_VALUE_ACCESSOR],
+  directives: [MdAriaTarget],
+})
+export class MdInput implements ControlValueAccessor, AfterContentInit, OnChanges {
+  private _focused: boolean = false;
+  private _onTouched: () => void = noop;
+  private _onChangeFn: (_: any) => void = noop;
```
jelbourn: Add comment for `_onChangeFn` and _onTouched` like
```
/** Callback registered via registerOnChange (ControlValueAccessor). */
```
Maybe also rename them to `_onChangeCallback` / `_onTouchedCallback` ?

src/components/input/input.ts
hansl: Unfortunately I cannot do that. There are cases where we want to change the `_value` but not trigger the event; if we do that on a call to `writeValue()` for example, it results in an infinite loop.

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+  registerOnChange(fn: any) {
+    this._onChangeFn = fn;
+  }
+
+  /** Implemented as part of ControlValueAccessor. */
+  registerOnTouched(fn: any) {
+    this._onTouched = fn;
+  }
+
+  /** Implemented as part of AfterContentInit. */
```
jelbourn: I think this comment is unnecessary for lifecycle events.

(I think it's only useful when it's not obvious why the class has the method to begin with)

src/components/input/input.ts
hansl: Left aligned and right aligned.

src/components/input/input.ts
hansl: Yes. I need to remove it from the `<md-input>` tag.

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+  ngAfterContentInit() {
+    if (this.placeholder != '' && this.placeholder != null && this._placeholderChild != null) {
```
jelbourn: Need a high-level comment describing what you need to do _in_ `AfterContentInit`. Part of my wants to say that lifecycle events should only ever just call other functions, so you cleanly separate the *when* and the *what*.

src/components/input/input.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  ngOnChanges(changes: {[key: string]: SimpleChange}) {
+    if (!this._hintChildren) {
+      return;
+    }
+
+    // Initialize the hintChildren.
+    let startHint: MdHint = null;
+    let endHint: MdHint = null;
+    iterateListLike(this._hintChildren, (hint: MdHint) => {
```
jelbourn: `QueryList` has a `forEach` method.

src/components/input/input.ts
hansl: Done.

src/components/input/input.ts
hansl: Honestly I don't understand why the change of API conventions; `writeValue` should really be `ngWriteValue`, etc... while `AfterContentInit` should be required to be able to call the `ngAfterContentInit` function. :shrug:

src/core/annotations/field-value.ts
```diff
@@ -0,0 +1,38 @@
+import {isPresent} from 'angular2/src/facade/lang';
+
+
+declare var Symbol: any;
```
jelbourn: Is this legal? TS already supports ES6 `Symbol` for built-in symbols, so wouldn't there be a conflict? Why use Symbol at all? I'm not sure I see the benefit.

src/demo-app/input/input-demo.scss
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.demo-transform {
+  transition: color $swift-ease-out-duration $swift-ease-out-timing-function;
+}
+.demo-primary {
+  color: md-color($md-primary);
+}
+
+md-card {
```
jelbourn: moar `.demo-` prefix

src/components/input/input.ts
hansl: mmmh. Apparently it's the `map` function that doesn't work. cool.

src/components/input/input.ts
hansl: done

src/components/input/input.scss
```diff
@@ -0,0 +1,182 @@
...
+
+  // Global wrapper. We need to apply margin to the element for spacing, but
+  // cannot apply it to the host element directly.
+  .md-input-wrapper {
+    margin: 16px 0;
+  }
+
+  // We use a table layout to baseline align the prefix and suffix classes.
+  // The underline is outside of it so it can cover all of the elements under
+  // this table.
```
jelbourn: Can you talk about why `display: inline-table` is more suitable for this than other approaches (especially flex)?

jelbourn:
Done w/ second pass

src/core/annotations/field-value.ts
hansl: No. We simply declare the ambient variable. Otherwise it's not defined and TypeScript will scream. Apparently `es6-shim` doesn't have that.

src/components/input/input.ts
hansl: Done. Moved all this.

src/demo-app/input/input-demo.scss
hansl: done.

src/components/input/input.scss
hansl: Took a shot, but my brain is a jello today.

hansl:
PTAL.

src/components/input/input.html
```diff
@@ -0,0 +1,47 @@
...
+  </div>
+
+  <div class="md-input-underline"
+       [class.md-disabled]="disabled">
+    <span class="md-input-ripple"
+          [class.md-focused]="focused"
+          [class.md-accent]="dividerColor == 'accent'"></span>
+  </div>
+
+  <div *ngIf="hintLabel != ''" class="md-hint">{{hintLabel}}</div>
```
jelbourn: No actions now, but my gut is worried about each input having ~25 bindings. Something to explore later.

src/components/input/input.scss
```diff
@@ -0,0 +1,184 @@
...
+    outline: none;
+    padding: 0;
+    width: 100%;
+
+    &.md-end {
+      text-align: right;
+    }
+  }
+
+  // The placeholder label. This is invisible unless it is. The logic to show it is
```
jelbourn: `"This is invisible unless it is.`"

Deep.

src/components/input/input.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    beforeEach(inject([TestComponentBuilder], function (tcb: TestComponentBuilder) {
+      builder = tcb;
+    }));
+
+    it('creates a native <input> element', injectAsync([], () => {
+      return builder.createAsync(MdInputBaseTestController)
+        .then((fixture) => {
+          fixture.detectChanges();
+          expect(fixture.debugElement.query(By.css('input'))).not.toBeNull();
```
jelbourn: Would this pass?
```
expect(undefined).not.toBeNull()
```

src/components/input/input.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+          expect(el.getAttribute('aria-label')).toEqual('label');
+          instance.ariaLabel = 'label 2';
+          fixture.detectChanges();
+          expect(el.getAttribute('aria-label')).toEqual('label 2');
+
+          expect(el.getAttribute('aria-disabled')).not.toBeNull();
+        });
+    }));
+
+    it('validates there\'s only one hint label per side', injectAsync([], () => {
```
jelbourn: Use backticks or double-quotes to avoid escaping?

src/components/input/input.ts
jelbourn: We may want to consider a different attribute name across the library if we're regularly going to have to set up an extra binding just to address this.

src/components/input/input.ts
jelbourn: Can you add that explanation to a comment here?

src/components/input/input.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+          let instance = fixture.componentInstance;
+          fixture.detectChanges();
+          let inputInstance = fixture.debugElement.query(By.directive(MdInput)).componentInstance;
+          expect(inputInstance.characterCount).toEqual(0);
+
+          instance.model = 'hello';
+          fixture.detectChanges();
+          expect(inputInstance.characterCount).toEqual(5);
+        });
+    }));
```
jelbourn: Should there also be a test for _showing_ the character count?

src/components/input/input.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    it('creates a native <input> element', injectAsync([], () => {
+      return builder.createAsync(MdInputBaseTestController)
+        .then((fixture) => {
+          fixture.detectChanges();
+          expect(fixture.debugElement.query(By.css('input'))).not.toBeNull();
+        });
+    }));
+
+    it('implements ControlValueAccessor', injectAsync([], () => {
```
jelbourn: Should the test be `it('should support ngModel' ...`, since implementing `ControlValueAccessor` is just an implementation detail?

src/components/input/input.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+    it('validates the type', injectAsync([], () => {
+      return builder.createAsync(MdInputInvalidTypeTestController)
+        .then((fixture: ComponentFixture) => {
+          fakeAsync(() => {
+            expect(() => fixture.detectChanges()).toThrow(MdInputUnsupportedTypeException);
+          });
+        });
+    }));
+  });
+}
```
jelbourn: Needs tests for various hint and placeholder configurations.

src/components/input/input.spec.ts
hansl: There's no standard for character count, so really it's two tests; one that check the property is what's expected, and the other that checks the hint label is what's expected. The character count is simply a usage of both.

src/components/input/input.scss
hansl: Knew you'd like :)

src/components/input/input.html
hansl: Noted. There should be a greater task of performance checks AND visual checking before full release. In beta maybe.

Also, https://www.youtube.com/watch?v=fpaQpyU_QiM

src/components/input/input.spec.ts
hansl: I thought the linting was smart enough to know if you don't need backtick because you don't use `${}`. And double quotes are forbidden by linting too. Will check tomorrow.

src/components/input/input.ts
hansl: Actually after looking at it, now that we don't set a bunch of side variables I can just inline that function.

src/components/input/input.ts
hansl: We may. Let's chat when you come back.

src/components/input/input.spec.ts
hansl: `getAttribute()` returns `null` or  `""`, never `undefined` according to spec.

But yes, that would pass since `toBe` is strict equality. Replaced with `toBeTruthy()`.

src/components/input/input.spec.ts
hansl: Done.

src/components/input/input.spec.ts
hansl: Done. Added a bunch of tests.

jelbourn:
LGTM

PR #163, opened 2016-03-15
chore(): add release staging script 

jelbourn:
R: @hansl @kara @IgorMinar @robwormald 

* Creates a script to stage the release, `stage-release.sh`
* Adds a `package.json` for each component and for `core/`
* Rename progress circle files to match directory.

I tested installing the package tarballs and got them to work with our demo site. It simply requires the following configuration for each component (and core) in the SystemJS packages:
```js
'@angular2-material/sidenav': {
  format: 'cjs',
  defaultExtension: 'js',
  main: 'sidenav.js'
},
'@angular2-material/core': {
  format: 'cjs',
  defaultExtension: 'js',
  main: 'core.js'
},
```


kara:
Lgtm.  Question: did we want to run tests as a sanity check after building?

jelbourn:
@kara not sure what you mean.

scripts/release/stage-release.sh
```diff
@@ -0,0 +1,46 @@
...
+# Inline the css and html into the component ts files.
+./node_modules/gulp/bin/gulp.js inline-resources
+
+# deploy/ serves as a working directory to stage the release.
+mkdir deploy
+
+# Copy all components/ to deploy/ and replace "../core" with just "core" since each
+# component directory will now live as a sibling to core/
+cp -R ./dist/components/* ./deploy/
+find ./deploy -type f \( -name "*.js" -o -name "*.ts" \) -exec sed -i '' 's|\.\./core|core|g' {} \;
```
hansl: This is just to output in the logs, I presume?

jelbourn: Nope, this is actually changing the imports. I'm doing this so that we can keep our source code organized into `components/` and `core/`, but actually release with the components as siblings to core. Since the components include core via relative path, we have "un up" the path to core by one. 

I know it seems weird, but it will be safe if we stick to our current directory structure.

scripts/release/stage-release.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Clear dist/ and deploy/ so that we guarantee there are no stale artifacts.
+rm -rf ./dist
+rm -rf ./deploy
+
+# For packaging for npm only, replace the TypeScript module format to commonjs
+TS_CONFIG_ORIGINAL=$(cat ./src/tsconfig.json)
+SYSTEM_FORMAT="\"module\": \"system\","
+COMMONJS_FORMAT="\"module\": \"commonjs\","
+TS_CONFIG_MODIFIED=${TS_CONFIG_ORIGINAL/${SYSTEM_FORMAT}/${COMMONJS_FORMAT}}
```
hansl: Can you use `sed` here instead of some local variable?

```bash
sed -i .backup  "s/\(.*\"module\": \).*/\\1\"commonjs\",/" src/tsconfig.json
ng build
mv -f ${TSCONFIG_PATH}.backup ${TSCONFIG_PATH}
```

scripts/release/stage-release.sh
hansl: Yeah I saw the `-i` after my comment, and removed my original comment :)

src/components/button/package.json
```diff
@@ -0,0 +1,29 @@
+{
```
hansl: Could we generate those files during build, and have a generic `package.json`?  Do you plan on having differences other than the component name in these?

This could be a phase 2.

kara:
@jelbourn As in, we likely don't want to release if unit tests are failing. Theoretically, we shouldn't be merging code in with failing tests, but it happens.  Maybe this isn't the right step to check, but thought I'd mention it.

src/components/button/package.json
jelbourn: As discussed, they'll have different dependencies and, versions, descriptions, and keywords. For now we'll stick with separate files and explore different approaches if managing them becomes an issue.

scripts/release/stage-release.sh
jelbourn: Done.

jelbourn:
@kara we can handle that with process for the time being.

src/core/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "material design",
+    "components"
+  ],
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/angular/material2/issues"
+  },
+  "homepage": "https://github.com/angular/material2#readme",
+  "peerDependencies": {
+    "angular2": "2.0.0-beta.8"
```
IgorMinar: can you use ^2.0.0-beta.8 instead?

I think it's better right now to error on the side of claiming compatibility with future Angular versions containing breaking changes rather than being too strict. If you are too strict people will not be able to use angular2@2.0.0-beta.9 until you cut a release. If you use ^ and we break material, then people can stick to beta.8 via package.json or shrinkwrap.

IgorMinar: you'll need to make this change in all package.json files.

scripts/release/stage-release.sh
```diff
@@ -0,0 +1,42 @@
...
+# material2 repo.
+
+# Make sure you are not running `ng serve` or `ng build --watch` when running this.
+
+
+# Clear dist/ and deploy/ so that we guarantee there are no stale artifacts.
+rm -rf ./dist
+rm -rf ./deploy
+
+# For packaging for npm only, replace the TypeScript module format to commonjs
```
IgorMinar: it's not obvious for the casual reader that this will create tsconfig.json.backup, maybe state that explicitly in the comment

scripts/release/stage-release.sh
```diff
@@ -0,0 +1,42 @@
...
+# component directory will now live as a sibling to core/
+cp -R ./dist/components/* ./deploy/
+find ./deploy -type f \( -name "*.js" -o -name "*.ts" \) -exec sed -i '' 's|\.\./core|core|g' {} \;
+
+# Copy the core/ directory directly into ./deploy
+cp -R ./dist/core/ ./deploy/core/
+
+# Create tarball packages for local testing (not for publishing / distribution)
+# (if installing these packages doesn't work, ensure your npm version is 3.5.0 or greater)
+find ./deploy/* -type d -maxdepth 0 | xargs npm pack
```
IgorMinar: I'd use `npm link` for testing instead.

src/core/package.json
jelbourn: Done.

scripts/release/stage-release.sh
jelbourn: Done.

scripts/release/stage-release.sh
```diff
@@ -0,0 +1,42 @@
...
+mkdir ./deploy/packages/
+mv *.tgz ./deploy/packages/
```
IgorMinar: can you also document instructions for how to test the build locally? it will be useful for people not familiar with npm link (or npm pack if you stick to that)

scripts/release/stage-release.sh
jelbourn: Removed tarball creation; will use `npm link` locally for testing.

scripts/release/stage-release.sh
```diff
@@ -0,0 +1,37 @@
...
+# deploy/ serves as a working directory to stage the release.
+mkdir deploy
+
+# Copy all components/ to deploy/ and replace "../core" with just "core" since each
+# component directory will now live as a sibling to core/
+cp -R ./dist/components/* ./deploy/
+find ./deploy -type f \( -name "*.js" -o -name "*.ts" \) -exec sed -i '' 's|\.\./core|core|g' {} \;
+
+# Copy the core/ directory directly into ./deploy
+cp -R ./dist/core/ ./deploy/core/
```
IgorMinar: please document instructions on how to test the build via `npm link`

IgorMinar:
per https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit (if you fully adopted that) the commit message should look like:

```
build: add release staging script
```

jelbourn:
@IgorMinar made suggested changes (I used chore() since I didn't see it as affecting the _build_, but I don't really care either way).

IgorMinar:
just fyi the empty `()` is not required. most tools and people treat `build:`, `build():` and `build(*):` in the same way.

hansl:
@robertmesserle See the Travis build which failed on `apt-get install libstdc++6`

hansl:
LGTM

PR #173, opened 2016-03-16
build(): make npm install also install typings

jhades:
This simplifies the installation process, we get one less step. This would help avoid erroneous bug reports due to missing typings


jhades:
I signed it!

jelbourn:
@hansl has opinions about `postinstall` scripts.

hansl:
Now that we use separate packages for every components, this is okay.

LGTM.

hansl:
Committed with https://github.com/angular/material2/commit/ead66f70d2e2d9ea625916c91d3314ea19d68687

Issue #167, opened 2016-03-15
Packaged version is using System.register instead of commonjs


UnwrittenFun:
How do you propose this to be used without SystemJS?

I am currently using webpack for my angular 2 app and have been following this repo closely, however with alpha.0 it seems it can only be consumed if you are using SystemJS due to the packages using System.register.

It would be nice to see this usable in a commonjs environment.

jelbourn:
@UnwrittenFun Hmm, this looks like a bug in the release process. It was supposed to be commonjs modules in the release.

aberenyi:
Is it not just the simple case of replacing `system` with `commonjs` here https://github.com/angular/material2/blob/master/src/tsconfig.json#L7?

jelbourn:
It is. The release script was supposed to do that automatically.

jelbourn:
This should be fixed now with the `alpha.1` release.

PR #176, opened 2016-03-16
build(): change tsconfig to always use commonjs modules

jelbourn:
R: @hansl 

This reduces the complexity of the release staging script and will prevent things like #167 from happening again. Demo site works exactly the same with the new config.


hansl:
LGTM.

PR #180, opened 2016-03-16
chore: analyze travis commit and check for blocked statements.

### LINKED ISSUE #175: CI should fail when encountering ddescribe, iit, debugger, etc.
Travis CI should fail whenever one of the following is accidentally left in a commit.
* `ddescribe`
* `fdescribe`
* `iit`
* `fit`
* `xdescribe`
* `xit`
* `debugger;`

We want this to not apply to `import` statements for the jasmine statements.

DevVersion:
Attention: The CI is failing, because of the changes itself. 

image

Fixes #175.

### DIFF
--- a/scripts/ci/analyze-commits.js
+++ b/scripts/ci/analyze-commits.js
@@ -1,0 +1,30 @@
+'use strict';
+/*
+ * This script analyzes the current commits of the CI.
+ * It will search for blocked statements, which have been added in the commits and fail if present.
+ */
+
+var child_process = require('child_process');
+var blocked_statements = [
+  'ddescribe',
+  'fdescribe',
+  'iit',
+  'fit',
+  'xdescribe',
+  'xit',
+  'debugger;'
+];
+
+var blockedRegex = new RegExp('\\+.*(' + blocked_statements.join('|') + ').*$', 'mg');
+
+var diff = child_process.execSync('git diff --unified=0 HEAD~1 ./src ./e2e').toString();
+var isInvalid = blockedRegex.test(diff);
+
+if (isInvalid) {
+  console.warn('Warning: You are using a statement in your commit, which is not allowed.\n' +
+    'Blocked Statements are: ' + blocked_statements.join(', ') + '\n' +
+    'Please remove them, and the CI will continue.');
+  process.exit(1);
+} else {
+  console.log('Info: The commits have been analyzed and are valid!');
+}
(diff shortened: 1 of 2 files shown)


scripts/ci/analyze-commits.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
jelbourn: I wonder if this would be faster with a bash script that directly greps the output of the `git diff`.

@hansl what do you think?

DevVersion: @jelbourn In bash, this would require more work with the multiple lines. Since the bash output strips the line delimiters out. I don't say this won't work, but it would require more work and wouldn't be that elegant as in node.

scripts/ci/analyze-commits.js
```diff
@@ -0,0 +1,30 @@
...
+  'debugger;'
+];
+
+var blockedRegex = new RegExp('\\+.*(' + blocked_statements.join('|') + ').*$', 'mg');
+
+var diff = child_process.execSync('git diff --unified=0 HEAD~1').toString();
+var isInvalid = blockedRegex.test(diff);
+
+if (isInvalid) {
+  console.warn('Warning: You are using a statement in your commit, which is not allowed.\n' +
```
jelbourn: Use `console.error` since this actually fails?

scripts/ci/analyze-commits.js
```diff
@@ -0,0 +1,30 @@
...
+  'iit',
+  'fit',
+  'xdescribe',
+  'xit',
+  'debugger;'
+];
+
+var blockedRegex = new RegExp('\\+.*(' + blocked_statements.join('|') + ').*$', 'mg');
+
+var diff = child_process.execSync('git diff --unified=0 HEAD~1').toString();
```
jelbourn: Restrict the diff to the `src/` directory?

scripts/ci/analyze-commits.js
DevVersion: `console.warn` is an alias for `console.error`

See the node reference: https://nodejs.org/api/console.html#console_console_warn_data

But if you want me to change it, let me know.

jelbourn: Ah, I didn't know. Thanks!

scripts/ci/analyze-commits.js
hansl: and `e2e/`

jelbourn:
LGTM

PR #181, opened 2016-03-16
fix(MdCheckbox): Cancel spacebar keydown events when component focused

### LINKED ISSUE #162: Pressing space on MdCheckbox scrolls the page
We need to prevent the event from bubbling in this case.

traviskaufman:
This commit ensures that when a checkbox element is focused, spacebar
`keydown` events will not bubble up to window, causing side effects like
page scrolling.

This commit also fixes an issue as part of this change where the
component could still be focused even when it was disabled.

Finally, some of the test code has been cleaned up and corrected.

Fixes #162

### DIFF
--- a/src/components/checkbox/checkbox.ts
+++ b/src/components/checkbox/checkbox.ts
@@ -68,12 +68,13 @@ enum TransitionCheckState {
     '[class.md-checkbox-checked]': 'checked',
     '[class.md-checkbox-disabled]': 'disabled',
     '[class.md-checkbox-align-end]': 'align == "end"',
-    '[tabindex]': 'disabled ? -1 : tabindex',
+    '[attr.tabindex]': 'disabled ? null : tabindex',
     '[attr.aria-label]': 'ariaLabel',
     '[attr.aria-labelledby]': 'labelId',
     '[attr.aria-checked]': 'getAriaChecked()',
     '[attr.aria-disabled]': 'disabled',
     '(click)': 'onInteractionEvent($event)',
+    '(keydown.space)': 'onSpaceDown($event)',
     '(keyup.space)': 'onInteractionEvent($event)',
     '(blur)': 'onTouched()'
   },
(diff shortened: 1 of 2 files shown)


src/components/checkbox/checkbox.ts
```diff
@@ -192,6 +199,15 @@ export class MdCheckbox implements ControlValueAccessor {
     this.toggle();
   }
 
+  onFocus(evt: Event) {
+    window.addEventListener('keydown', this._focusWindowKeydownHandler);
```
jelbourn: There's actually a special syntax for doing this without directly touching window:
```ts
  host: {
    '(window:domEvent)': 'onWindowEvent($event.type)',
    '(document:domEvent)': 'onDocumentEvent($event.type)',
    '(body:domEvent)': 'onBodyEvent($event.type)'
  }
```

traviskaufman: ah that's way better. will change.

traviskaufman:
@jelbourn changes made!

src/components/checkbox/checkbox.ts
```diff
@@ -192,6 +198,33 @@ export class MdCheckbox implements ControlValueAccessor {
...
+    this._isFocused = false;
+    this.onTouched();
+  }
+
+  /**
+   * Event handler used for (keydown.space) events on the window. Used to prevent spacebar events
+   * from bubbling when the component is focused, which prevents side effects like page scrolling
+   * from happening.
+   */
+  onWindowKeyDown(evt: KeyboardEvent) {
```
jelbourn: As discussed in chat: just use `keydown` on the host element since it should only fire when it has focus anyway?

traviskaufman: :+1: yeah I think I overthought it. Changed.

jelbourn:
LGTM

PR #187, opened 2016-03-16
chore(): add package.json for radio

jelbourn:
R: @hansl 

Also rename `radio_button` to just `radio` to be more consistent with other components.


hansl:
Lgtm

PR #193, opened 2016-03-17
fix(radio): export the dispatcher in radio.ts

jelbourn:
R: @kara 

This will let consumers of the npm package simply do 
```ts
import {MdRadioDispatcher} from '@angular2-material/radio';
```
instead of 
```ts
import {MdRadioDispatcher} from '@angular2-material/radio/radio_dispatcher';
```


kara:
LGTM

PR #194, opened 2016-03-17
fix(e2e): screenshot changes will display proper error message

### LINKED ISSUE #191: bug(e2e): screenshot tests output inaccurate error message
Currently, all failing screenshot tests throw an error saying that the screenshots were not found.

### LINKED ISSUE #192: Screenshots in master need to be updated

robertmesserle:
closes #191
closes #192

### DIFF
--- a/e2e/_screenshot.js
+++ b/e2e/_screenshot.js
@@ -58,14 +58,14 @@ class Screenshot {
       let referenceScreenshot = mapnik.Image.open(this.path);
       this.overwriteExistingScreenshot();
       if (referenceScreenshot.compare(this.png)) {
-        throw new Error(`screenshot "${this.id}" has changed.`);
+        throw `screenshot "${this.id}" has changed.`;
       } else {
         console.info('[STATUS] Screenshot has not changed');
       }
     } catch (e) {
-      console.info(`[STATUS] No reference screenshot found`);
+      console.error(`[ERROR] `, e);
       this.overwriteExistingScreenshot();
-      throw new Error(`screenshot "${this.id}" was not found.`);
+      throw e;
     }
   }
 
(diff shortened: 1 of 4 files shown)


robertmesserle:
@jelbourn @hansl

e2e/_screenshot.js
```diff
@@ -58,14 +58,14 @@ class Screenshot {
       let referenceScreenshot = mapnik.Image.open(this.path);
       this.overwriteExistingScreenshot();
       if (referenceScreenshot.compare(this.png)) {
-        throw new Error(`screenshot "${this.id}" has changed.`);
+        throw `screenshot "${this.id}" has changed.`;
```
jelbourn: Why throw a string instead of an error? Add a comment?

jelbourn:
Can address commit in a follow-up if necessary. LGTM for getting the build green.

robertmesserle:
@jelbourn No real reason, to be honest, other than when I was looking at what others were doing in Node, it seemed more common to just throw a string.

e2e/index.e2e.js
```diff
@@ -5,7 +5,7 @@ describe('hello, protractor', function () {
     browser.get('/');
     it('should have a title', function () {
       expect(browser.getTitle()).toBe('Material2');
-      screenshot('initial state');
+      screenshot('demo page: index');
```
hansl: Is that still going in the file name? If so Windows is allergic to `:`s

robertmesserle: Ah, good point.  I plan to add filtering to that so it will strip special
characters.  Ideally, I think it would be nice to use slashes... like
"component_name/screenshot_name".

hansl: Then do that in this PR if you're going to change the name :)

robertmesserle: Yea, I'll update that on Monday.

PR #195, opened 2016-03-17
fix(button): apply color classes correctly.

### LINKED ISSUE #75: md-button attaches class "md-undefined" to host element when on color provided

DevVersion:
Fixes #75. Closes #89

### DIFF
--- a/src/components/button/button.ts
+++ b/src/components/button/button.ts
@@ -5,6 +5,8 @@ import {
   HostBinding,
   HostListener,
   ChangeDetectionStrategy,
+  ElementRef,
+  Renderer,
 } from 'angular2/core';
 import {TimerWrapper} from 'angular2/src/facade/async';
 
@@ -18,7 +20,6 @@ import {TimerWrapper} from 'angular2/src/facade/async';
   inputs: ['color'],
   host: {
     '[class.md-button-focus]': 'isKeyboardFocused',
-    '[class]': 'setClassList()',
     '(mousedown)': 'setMousedown()',
     '(focus)': 'setKeyboardFocus()',
     '(blur)': 'removeKeyboardFocus()'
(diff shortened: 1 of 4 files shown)


src/components/button/button.ts
```diff
@@ -48,6 +57,18 @@ export class MdButton {
     TimerWrapper.setTimeout(() => { this.isMouseDown = false; }, 100);
   }
 
+  _updateColor(newColor: string) {
+    if (this._color != null && this._color != '') {
+      this.renderer.setElementClass(this.elementRef.nativeElement, `md-${newColor}`, false);
```
jelbourn: Should be `this._color` in the template expression

DevVersion: Indeed -  fixed it, thanks!

src/components/button/button.ts
```diff
@@ -48,6 +57,18 @@ export class MdButton {
...
   }
 
+  _updateColor(newColor: string) {
+    if (this._color != null && this._color != '') {
+      this.renderer.setElementClass(this.elementRef.nativeElement, `md-${this._color}`, false);
+    }
+
+    if (newColor != null && newColor != '') {
+      this.renderer.setElementClass(this.elementRef.nativeElement, `md-${newColor}`, true);
+    }
```
kara: There's some repetition between lines 61-63 and lines 65-67. Can you refactor to share more code, like:

```javascript
set color(value: string) {
   this._updateColor(this._color, false);
   this._updateColor(value, true);
}

_updateColor(color: string, bool: boolean): void {
   if (color != null && color != '') {
      this.renderer.setElementClass(this.elementRef.nativeElement, `md-${color}`, bool);
   }
}
```

DevVersion: @kara Changed it as discussed. Thanks for the suggestion.

jelbourn:
LGTM

PR #197, opened 2016-03-17
chore: travis should only build pushes on master.

DevVersion:
Travis should not build twice, when pushing changes to a Pull Request, which has its branch on the upstream repository.

**Caution**: Seems like the CI is not running on this PR here:
  Maybe Build PR's is disabled in Travis? (caused by new changes - require option)
  CI has problems?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ addons:
     packages:
       - libstdc++6
 
+branches:
+  only:
+  - master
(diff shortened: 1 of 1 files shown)


jelbourn:
@DevVersion I don't think we want to do this. AFAICT it will prevent travis from running on any PR that's coming from a branch.

PR #202, opened 2016-03-18
chore: travis should only build pushes on master.

DevVersion:
As discussed with @jelbourn, I will submit a new PR, because we're not able to reopen the old PR.

We are trying to find the issue with CI branch restriction.


jelbourn:
LGTM - this is what material1 does and it works fine.

PR #206, opened 2016-03-19
feat(button): add md-icon-button styling.

### LINKED ISSUE #188: Add md-icon-button variant for MdButton
Should follow the rest of the button variants:
```html
<button md-icon-button>
  <!-- md-icon coming soon! -->
</button>
```

DevVersion:
* Added md-icon-button component / styling
* Showcased the md-icon-button in the button demos.

I noticed that the icon-button is like a clone of a min-fab, maybe we can make the fab mixin more generic, to let it use by the icon-button too. 

Fixes #188

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -31,6 +31,28 @@
   @include md-fab($md-mini-fab-size, $md-mini-fab-padding)
 }
 
+[md-icon-button] {
+  @extend %md-button-base;
+
+  min-width: 0;
+  padding: 0;
+
+  width: $md-icon-button-size;
+  height: $md-icon-button-size;
+
+  line-height: $md-icon-button-line-height;
+  border-radius: $md-icon-button-border-radius;
+
+  &.md-button-focus {
+    background-color: md-color($md-foreground, base, 0.12);
+    @include md-button-theme('background-color', 0.12);
+  }
+
+  .md-button-wrapper > * {
+    vertical-align: middle;
+  }
+}
+
 // Applies a clearer border for high-contrast mode (a11y)
 @media screen and (-ms-high-contrast: active) {
   .md-raised-button, .md-fab, .md-mini-fab {
--- a/src/demo-app/button/button-demo.html
+++ b/src/demo-app/button/button-demo.html
@@ -46,6 +46,20 @@
   </section>
 
   <section>
+    <button md-icon-button color="primary">
+      <i class="material-icons md-24">menu</i>
+    </button>
+
+    <button md-icon-button color="accent">
+      <i class="material-icons md-24">favorite</i>
+    </button>
+
+    <button md-icon-button>
+      <i class="material-icons md-24">more_vert</i>
+    </button>
+  </section>
+
+  <section>
     <div>
       <span>isDisabled: {{isDisabled}}</span>
       <button md-raised-button (click)="isDisabled=!isDisabled">Disable buttons</button>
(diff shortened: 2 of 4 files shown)


src/components/button/button.scss
```diff
@@ -31,6 +31,28 @@
...
+  @extend %md-button-base;
+
+  min-width: 0;
+  padding: 0;
+
+  width: 40px;
+  height: 40px;
+
+  line-height: 24px;
+  border-radius: 50%;
```
jelbourn: I see that the material1 icon-button is the same, but I'm curious: why does it use `border-radius: 50%`?

src/components/button/button.scss
```diff
@@ -31,6 +31,28 @@
...
+
+  &.md-button-focus {
+    background-color: md-color($md-foreground, base, 0.12);
+    @include md-button-theme('background-color', 0.12);
+  }
+
+  .md-button-wrapper > * {
```
jelbourn: Why does this need a `*` selector?

jelbourn:
cc @kara

src/components/button/button.scss
DevVersion: This is needed to align the first level child items of the wrapper in the middle (correctly to the line-height)

src/components/button/button.scss
DevVersion: Because if you have a 24x24 square, then 50% of all sides will be used to archive a responsive perfect circle.

So it's needed to make the element round, like in polymer-icon-button and Material 1

src/components/button/button.scss
```diff
@@ -31,6 +31,28 @@
   @include md-fab($md-mini-fab-size, $md-mini-fab-padding)
 }
 
+[md-icon-button] {
+  @extend %md-button-base;
+
+  min-width: 0;
+  padding: 0;
+
+  width: 40px;
```
kara: Would you mind putting this in a Sass variable?  All other button sizes are variables in _button_base.scss, so dense layouts can adjust as necessary.

src/components/button/button.scss
```diff
@@ -31,6 +31,28 @@
   @include md-fab($md-mini-fab-size, $md-mini-fab-padding)
 }
 
+[md-icon-button] {
+  @extend %md-button-base;
```
kara: Given that this extends the base button rather than a fab, it will likely need styling for its disabled state. Is this covered somewhere?

src/demo-app/button/button-demo.html
```diff
@@ -46,6 +46,20 @@
...
+      <i class="material-icons md-24">menu</i>
+    </button>
+
+    <button md-icon-button color="accent">
+      <i class="material-icons md-24">favorite</i>
+    </button>
+
+    <button md-icon-button>
+      <i class="material-icons md-24">more_vert</i>
+    </button>
```
kara: Can you add an icon-button to the disabled section at the bottom as well?  It will ensure disabled styles are applied properly to this type of button over time.

DevVersion: Yes, great idea.

src/components/button/button.scss
DevVersion: Yes, it is already covered by the default button base.

image

src/components/button/button.scss
DevVersion: Yes, done :+1:

jelbourn:
LGTM

matheo:
I wonder why this isn't included in the documentation examples :(

willshowell:
@matheo it will be soon. Here is the example https://github.com/angular/material2/pull/5628

matheo:
@willshowell awesome! thanks

Issue #208, opened 2016-03-19
Not able to bind md-radio-group to an ngModel


doctorrustynelson:
* **Do you want to request a *feature* or report a *bug*?** Bug

* **What is the current behavior?** md-radio-group does not support ngModel binding

* **If the current behavior is a bug,
please provide steps to reproduce and if possible a minimal demo of the problem**
https://plnkr.co/edit/NaRV51Ylr6oj1sGR7PmQ?p=preview

* **What is the expected behavior?** ngModel should be able to bind to an md-radio-group

* **What is the motivation / use case for changing the behavior?** Being able to get user input into 

* **Which version of Angular and Material, and which browser and OS does this issue affect?** All versions of Angular and Material up to now (master commit 111)
Did this work in previous versions of Angular / Material? No

* **Other information**
(e.g. detailed explanation, stacktraces, related issues, suggestions how to fix)

PR #209, opened 2016-03-19
feat(radio): support ngModel on md-radio-group

doctorrustynelson:
* Added support for ngModel on md-radio-group
* Updated Documentation
* Added Demo App Example

### DIFF
--- a/src/components/radio/radio.ts
+++ b/src/components/radio/radio.ts
@@ -10,13 +10,30 @@ import {
   OnInit,
   Optional,
   Output,
+  Provider,
   QueryList,
   ViewEncapsulation,
   forwardRef
 } from 'angular2/core';
 
+import {
+  NG_VALUE_ACCESSOR,
+  ControlValueAccessor
+} from 'angular2/src/common/forms/directives/control_value_accessor';
+import {CONST_EXPR} from 'angular2/src/facade/lang';
+
 import {MdRadioDispatcher} from './radio_dispatcher';
 export {MdRadioDispatcher} from './radio_dispatcher';
+
+/**
+ * Provider Expression that allows md-radio-group to register as a ControlValueAccessor. This
+ * allows it to support [(ngModel)] and ngControl.
+ */
+const MD_RADIO_GROUP_CONTROL_VALUE_ACCESSOR = CONST_EXPR(new Provider(
+    NG_VALUE_ACCESSOR, {
+      useExisting: forwardRef(() => MdRadioGroup),
+      multi: true
+    }));
 
 // TODO(mtlin):
 // Ink ripple is currently placeholder.
@@ -36,11 +53,12 @@ export class MdRadioChange {
 
 @Directive({
   selector: 'md-radio-group',
+  providers: [MD_RADIO_GROUP_CONTROL_VALUE_ACCESSOR],
   host: {
     'role': 'radiogroup',
   },
 })
-export class MdRadioGroup implements AfterContentInit {
+export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
   /** The value for the radio group. Should match currently selected button. */
   private _value: any = null;
 
@@ -52,6 +70,11 @@ export class MdRadioGroup implements AfterContentInit {
 
   /** The currently selected radio button. Should match value. */
   private _selected: MdRadioButton = null;
+
+  /** Change event subscription set up by registerOnChange (ControlValueAccessor). */
+  private _changeSubscription: {unsubscribe: () => any} = null;
+
+  onTouched: () => any = () => {};
 
   /** Event emitted when the group value changes. */
   @Output()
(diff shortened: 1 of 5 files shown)


doctorrustynelson:
This should fix #208 that I submitted earlier.

src/components/radio/radio.ts
```diff
@@ -53,9 +71,13 @@ export class MdRadioGroup implements AfterContentInit {
...
   private _selected: MdRadioButton = null;
 
+  private _changeSubscription: {unsubscribe: () => any} = null;
+
+  onTouched: () => any = () => {};
+
   /** Event emitted when the group value changes. */
   @Output()
-  change: EventEmitter<MdRadioChange> = new EventEmitter();
+  change: EventEmitter<any> = new EventEmitter();
```
jelbourn: Why are you removing the event object?

doctorrustynelson: I changed the type of the event from the MdRadioChange to just the value of the radio button because if I found that if I didn't the ngModel I would bind to would get an MdRadioChange event as it's value instead of the intended value.  The changes related to this line up with similar code that I saw in the MdCheckbox implementation which I used as a reference for most of my changes.

jelbourn: We should keep the event object (checkbox should also have one eventually), and make it so that you just call the `fn` argument with the value extracted from the event object:
```ts
registerOnChange(fn: any) {
  if (this._changeSubscription) {
    this._changeSubscription.unsubscribe();
  }
  this._changeSubscription = <{unsubscribe: () => any}>this.change.subscribe(changeEvent => {
    fn(changeEvent.value);
  });
}```

Comment (doctorrustynelson) [inline src/components/radio/radio.ts]: I have re-included the MdRadioChange event as requested.

File: src/components/radio/radio.ts [hunk 2]
```diff
@@ -10,14 +10,31 @@ import {
...
+  NG_VALUE_ACCESSOR,
+  ControlValueAccessor
+} from 'angular2/src/common/forms/directives/control_value_accessor';
+import {CONST_EXPR} from 'angular2/src/facade/lang';
+
 import {MdRadioDispatcher} from './radio_dispatcher';
 export {MdRadioDispatcher} from './radio_dispatcher';
 
+/**
+ * Provider Expression that allows md-checkbox to register as a ControlValueAccessor. This allows it
```
jelbourn: "md-checkbox" > "md-radio-group"

src/components/radio/radio.ts
```diff
@@ -53,6 +71,10 @@ export class MdRadioGroup implements AfterContentInit {
   /** The currently selected radio button. Should match value. */
   private _selected: MdRadioButton = null;
 
+  private _changeSubscription: {unsubscribe: () => any} = null;
```
jelbourn: Add comment like
```ts
/** Change event subscription set up by registerOnChange (ControlValueAccessor). */ 
```

src/components/radio/radio.ts
```diff
@@ -155,6 +177,25 @@ export class MdRadioGroup implements AfterContentInit {
 
     selected.checked = true;
   }
+
+   /** Implemented as part of ControlValueAccessor. */
+  writeValue(value: any) {
+    this._value = value;
```
jelbourn: This should use `this.value` so that the setter runs.

src/components/radio/radio.ts
```diff
@@ -210,6 +251,10 @@ export class MdRadioButton implements OnInit {
     if (this.id == null) {
       this.id = `md-radio-${_uniqueIdCounter++}`;
     }
+
+    if ( this.radioGroup && this._value == this.radioGroup.value ) {
```
jelbourn: Remove leading / trailing spaces inside the parens.

src/demo-app/radio/radio-demo.ts
```diff
@@ -11,4 +11,10 @@ import {MdRadioDispatcher} from '../../components/radio/radio_dispatcher';
 })
 export class RadioDemo {
   isDisabled: boolean = false;
+  chosenOption: number = 3;
+  options = [
+    {label: 'Option 1', value: 1},
+    {label: 'Option 2', value: 2},
+    {label: 'Option 3', value: 3},
```
jelbourn: optional: come up with something a little more fun than `Option 1, Option 2...` :)

jelbourn:
Changes seem good, just needs unit tests now.

doctorrustynelson:
I have cleaned up the code as suggested and added a unit test.

src/components/radio/radio.ts
```diff
@@ -210,6 +252,10 @@ export class MdRadioButton implements OnInit {
     if (this.id == null) {
       this.id = `md-radio-${_uniqueIdCounter++}`;
     }
+
+    if (this.radioGroup && this._value == this.radioGroup.value) {
```
jelbourn: Looks like this added behavior also needs a test.

jelbourn:
LGTM aside from one more unit test.

cc @rfru

jelbourn:
@doctorrustynelson forgot one last thing: can you squash your commits into one? The commit message should be something like
```
feat(radio): support ngModel
```
Thanks!

doctorrustynelson:
@jelbourn Thank you for all the suggestions; I have done as requested and squashed the commits.

jelbourn:
LGTM

jelbourn:
Thanks for your contribution!

PR #210, opened 2016-03-20
fix(ci): fix flakiness by inlining html & css

jelbourn:
I was eating empanadas and working on the overlay, which prompted me to look at the portal spec. A part of my brain said:
"You know, I've never seen this test flake."
And another part responded:
"Of course it doesn't flake, it doesn't have a tem-".
"Oh."

So I've fixed the flakiness by inlining all the templates and CSS. After I did this, one of the checkbox tests was actually failing, so I fixed that too. 

This also has the pleasant side effect of making the tests in all browsers significantly faster.

Also, I removed firefox beta because all it does is make the SauceLabs optional task take forever to finish failing.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,6 @@ var inlineNg2Template = require('gulp-inline-ng2-template');
 
 gulp.task('inline-resources', function(){
   gulp.src('./dist/components/**/*.js')
-      .pipe(inlineNg2Template({base: './dist'}))
+      .pipe(inlineNg2Template({base: './dist', target: 'es5'}))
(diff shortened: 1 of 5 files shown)


jelbourn:
R: @hansl @robertmesserle @kara

hansl:
I don't see any change for karma to inline the changes. Did I miss a commit?

hansl:
Never mind I just saw you ran it as part of the Travis script. Lgtm.

robertmesserle:
First of all, **NICE!**

Second, LGTM!

PR #211, opened 2016-03-20
feat(overlay): add basic core of overlay

jelbourn:
R: @hansl @kara 

This is the first step on the journey to overlays. I'll be doing it in as small of pieces as possible for ease of reviewing. The PR covers adding an overlay element to the DOM and attaching a portal to it. Future PRs will cover styling, sizing, positioning, and helper directives.

### DIFF
--- a/src/core/overlay/overlay.ts
+++ b/src/core/overlay/overlay.ts
@@ -1,0 +1,98 @@
+import {
+  DynamicComponentLoader,
+  AppViewManager,
+  OpaqueToken,
+  Inject,
+  Injectable} from 'angular2/core';
+import {CONST_EXPR} from 'angular2/src/facade/lang';
+import {OverlayState} from './overlay-state';
+import {DomPortalHost} from '../portal/dom-portal-host';
+import {OverlayRef} from './overlay-ref';
+import {DOM} from '../platform/dom/dom_adapter';
+
+// Re-export OverlayState and OverlayRef so they can be imported directly from here.
+export {OverlayState} from './overlay-state';
+export {OverlayRef} from './overlay-ref';
+
+/** Token used to inject the DOM element that serves as the overlay container. */
+export const OVERLAY_CONTAINER_TOKEN = CONST_EXPR(new OpaqueToken('overlayContainer'));
+
+/** Next overlay unique ID. */
+let nextUniqueId = 0;
+
+/** The default state for newly created overlays. */
+let defaultState = new OverlayState();
+
+
+/**
+ * Service to create Overlays. Overlays are dynamically added pieces of floating UI, meant to be
+ * used as a low-level building building block for other components. Dialogs, tooltips, menus,
+ * selects, etc. can all be built using overlays. The service should primarily be used by authors
+ * of re-usable components rather than developers building end-user applications.
+ *
+ * An overlay *is* a PortalHost, so any kind of Portal can be loaded into one.
+ */
+ @Injectable()
+export class Overlay {
+  constructor(
+      @Inject(OVERLAY_CONTAINER_TOKEN) private _overlayContainerElement: Element,
+      private _dynamicComponentLoader: DynamicComponentLoader,
+      private _appViewManager: AppViewManager) {
+  }
+
+  /**
+   * Creates an overlay.
+   * @param state State to apply to the overlay.
+   * @returns A reference to the created overlay.
+   */
+  create(state: OverlayState = defaultState): Promise<OverlayRef> {
+    return this._createPaneElement(state).then(pane => this._createOverlayRef(pane));
+  }
+
+  /**
+   * Creates the DOM element for an overlay.
+   * @param state State to apply to the created element.
+   * @returns Promise resolving to the created element.
+   */
+  private _createPaneElement(state: OverlayState): Promise<Element> {
+    var pane = DOM.createElement('div');
+    pane.id  = `md-overlay-${nextUniqueId++}`;
+    DOM.addClass(pane, 'md-overlay-pane');
+
+    this.applyState(pane, state);
+    this._overlayContainerElement.appendChild(pane);
+
+    return Promise.resolve(pane);
+  }
+
+  /**
+   * Applies a given state to the given pane element.
+   * @param pane The pane to modify.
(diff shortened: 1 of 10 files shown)


src/core/overlay/overlay.ts
```diff
@@ -0,0 +1,98 @@
...
+/** Next overlay unique ID. */
+let nextUniqueId = 0;
+
+/** The default state for newly created overlays. */
+let defaultState = new OverlayState();
+
+
+/**
+ * Service to create Overlays. Overlay are dynamically added pieces of floating UI, meant to be
+ * used as low-level a building building block for other components. Dialogs, tooltips, menus,
```
kara: A few typos: 
 "Overlay are" --> "Overlay**s** are"
"used as low-level a building building block" --> "used as a low-level building block"

src/core/overlay/overlay.ts
```diff
@@ -0,0 +1,98 @@
...
+      private _dynamicComponentLoader: DynamicComponentLoader,
+      private _appViewManager: AppViewManager) {
+  }
+
+  /**
+   * Creates an overlay.
+   * @param state State to apply to the overlay.
+   * @returns A reference to the created overlay.
+   */
+  create(state: OverlayState = defaultState): Promise<OverlayRef> {
```
kara: I'm not sure default arguments work with TS2Dart yet. You may have to write this by checking if `isPresent`

src/demo-app/overlay/overlay-demo.ts
```diff
@@ -0,0 +1,45 @@
...
+  templateUrl: 'demo-app/overlay/overlay-demo.html',
+  styleUrls: ['demo-app/overlay/overlay-demo.css'],
+  directives: [TemplatePortalDirective],
+  providers: [
+    Overlay,
+    provide(OVERLAY_CONTAINER_TOKEN, {useValue: document.body})
+  ]
+})
+export class OverlayDemo {
+  @ViewChildren(TemplatePortalDirective) templatePortals: QueryList<Portal<any>>;
```
kara: Looks like there is an extra `>` at the end here?

src/core/overlay/overlay.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Creates the DOM element for an overlay.
+   * @param state State to apply to the created element.
+   * @returns Promise resolving to the created element.
+   */
+  private _createPaneElement(state: OverlayState): Promise<Element> {
+    var pane = DOM.createElement('div');
+    pane.id  = `md-overlay-${nextUniqueId++}`;
+    pane.classList.add('md-overlay-pane');
```
kara: Could we maybe use `DOM.addClass` here instead and pass in the element?

src/demo-app/overlay/overlay-demo.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  constructor(public overlay: Overlay, public elementRef: ElementRef) {
+    BrowserDomAdapter.makeCurrent();
```
DevVersion: Just a question: Why are you initializing the BrowserDomAdapater yourself?

Here is my post on Stackoverflow - [Click](http://stackoverflow.com/questions/34686856/angular-2-0-beta-access-dom-adapter-browserdomadapter)
And here's the Angular 2 Issue, which has been created from the stack: [Click](https://github.com/angular/angular/issues/6904)

As you can read in my stackoverflow post, the browser should initialize itself.

jelbourn: This isn't actually the `BrowserAdapter` from ng2, it's here in `src/core/platform` copied and pasted from `angular/angular`. It's only necessary for transpiling to Dart right now and will eventually be dropped when https://github.com/angular/angular/issues/7561 is done.

src/demo-app/overlay/overlay-demo.ts
jelbourn: Not that I can tell.

src/core/overlay/overlay.ts
jelbourn: They should work- I've seen them used throughout `angular/angular`.

src/demo-app/overlay/overlay-demo.ts
DevVersion: Oh well, didn't notice the `import`. Thanks.

src/demo-app/overlay/overlay-demo.ts
kara: You're right; missed the first one before Portal somehow :)

src/core/overlay/overlay.ts
```diff
@@ -0,0 +1,98 @@
...
+/** Next overlay unique ID. */
+let nextUniqueId = 0;
+
+/** The default state for newly created overlays. */
+let defaultState = new OverlayState();
+
+
+/**
+ * Service to create Overlays. Overlays are dynamically added pieces of floating UI, meant to be
+ * used as a low-level building building block for other components. Dialogs, tooltips, menus,
```
kara: Still has "building" twice?

kara:
LGTM, though still has that one typo if you want to fix

hansl:
@robertmesserle: the e2e is failing on this; https://travis-ci.org/angular/material2/builds/117214613

robertmesserle:
That means it’s working. He didn’t update the screenshot.

jelbourn:
I did update it. Seems like it fails to get the image.

robertmesserle:
@jelbourn Ah, my bad - I was looking on my phone earlier and didn't see the screenshot in the changed files.  Must've missed it.  Anyway, hard to say what is happening, but this is one of the first things that made me want to push for going back to the server idea for screenshots: our method of manually handling `git-lfs` files is definitely hacky and I wouldn't be surprised if that was the cause.

robertmesserle:
Also, after looking at the screenshot URL, it's likely the issue I brought up earlier about PR's that live in other people's repos.  The URL for your screenshot doesn't match the pattern of other screenshots we've dealt with so far, so manually handling the `git-lfs` files might not even really be a valid option.

jelbourn:
@robertmesserle another bullet point for your doc.

PR #216, opened 2016-03-21
fix(ci): analyze commits should scan complete commit range.

DevVersion:
This fixes the CI, to use the complete commit range.

- Needs to be tested.

### DIFF
--- a/scripts/ci/forbidden-identifiers.js
+++ b/scripts/ci/forbidden-identifiers.js
@@ -16,15 +16,18 @@ var blocked_statements = [
 ];
 
 var blockedRegex = new RegExp('\\+.*(' + blocked_statements.join('|') + ').*$', 'mg');
+var commit_range = process.env.TRAVIS_COMMIT_RANGE || 'HEAD~1';
 
-var diff = child_process.execSync('git diff --unified=0 HEAD~1 ./src ./e2e').toString();
-var isInvalid = blockedRegex.test(diff);
+var diff = child_process.execSync(`git diff --unified=0 ${commit_range} ./src ./e2e`).toString();
 
-if (isInvalid) {
+if (blockedRegex.test(diff)) {
+
   console.warn('Warning: You are using a statement in your commit, which is not allowed.\n' +
     'Blocked Statements are: ' + blocked_statements.join(', ') + '\n' +
     'Please remove them, and the CI will continue.');
+
   process.exit(1);
+
 } else {
   console.log('Info: The commits have been analyzed and are valid!');
 }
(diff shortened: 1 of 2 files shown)


scripts/ci/analyze-commits.js
```diff
@@ -16,15 +16,19 @@ var blocked_statements = [
 ];
```
jelbourn: While you're at it, can you rename this file to `forbidden-identifiers.js` ?

DevVersion: Sure, will change that.
By the way, now it works perfectly, just tested it:

image

cc. @hansl

scripts/ci/analyze-commits.js
```diff
@@ -16,15 +16,19 @@ var blocked_statements = [
...
 
 var blockedRegex = new RegExp('\\+.*(' + blocked_statements.join('|') + ').*$', 'mg');
+var commit_range = process.env.TRAVIS_COMMIT_RANGE || 'HEAD~1';
```
jelbourn: Why doesn't just `master` work? If not, maybe you can `git remote add` first?

DevVersion: Huh, sorry I didn't got that. 

`TRAVIS_COMMIT_RANGE` is the most perfect solution, because it checks the complete PR and nothing else. And `HEAD~1` is just a fallback.

scripts/ci/forbidden-identifiers.js
```diff
@@ -16,15 +16,19 @@ var blocked_statements = [
 ];
 
 var blockedRegex = new RegExp('\\+.*(' + blocked_statements.join('|') + ').*$', 'mg');
+var commit_range = process.env.TRAVIS_COMMIT_RANGE || 'HEAD~1';
 
-var diff = child_process.execSync('git diff --unified=0 HEAD~1 ./src ./e2e').toString();
-var isInvalid = blockedRegex.test(diff);
+var diff = child_process.execSync(
+  'git diff --unified=0 ' + commit_range + ' ./src ./e2e'
```
jelbourn: Use a template string?

jelbourn:
Good point on the branch feature thing. 

cc @hansl to take a look.

scripts/ci/forbidden-identifiers.js
DevVersion: I can do that, but I don't think it's really necessary, because a plain string concatenation is doing the same. Plus additionally it's compatible with older node versions.

hansl: (just want to add that template strings are valid Node since Node 4.x)

hansl: IMO it's fine. The CLI targets >= 4.5, I don't think Material2 should be compatible with anything older.

jelbourn:
LGTM, giving @hansl a change to chime in

hansl:
LGTM.

PR #220, opened 2016-03-22
fix (card) : card-header templateUrl

boulix3:
Fix the templateUrl path for card-header.
Error occurs when the website is not located in root folder (ex : http://localhost/TestApp/).


boulix3:
I signed it!

jelbourn:
LGTM

Issue #48, opened 2016-01-28
Make sure easing curves and shadows are on-spec


jelbourn:
Current easing curves and shadows are pulled from Angular 1 version. We need to double check that these reflect the current state of the spec.

bakgat:
In _shadows.scss this link is provided as documentation for shadows 
http://google.com/design/spec/what-is-material/objects-in-3d-space.html

In the new spec the link should be http://www.google.com/design/spec/what-is-material/elevation-shadows.html

I think there are differences in shadows. 
eg For raised buttons it states 2dp elevation, 8dp on active.

![elevation diagram]
(http://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0Bzhp5Z4wHba3VG9SaVpNbkpHb2s/whatismaterial_3d_elevation2.png)

traviskaufman:
mdElevation Design Doc: https://docs.google.com/document/d/1W3NGSLqDZzjbBBLW2C6y_6NUxtvdZAVaJvg58LY3Q0E/edit?usp=sharing

jelbourn:
Done for elevation.
@traviskaufman did you have any feedback on easing curves, or should we be able to grab what we need from the spec?

jelbourn:
@mmalerba the task here is to make sure all the timings and curves we're using are on-spec and shared between components. This is a lower-priority item for beta.

mmalerba:
Sorry, didn't mean to close this. Interesting part of the spec related to animations is dynamic durations: https://material.googleplex.com/motion/duration-easing.html#duration-easing-dynamic-durations (the idea that the duration changes based on the size of your device). Is this something we want to include in our implementation?

jelbourn:
I hadn't seen that before. I'd consider that a "nice-to-have" for now, something we can refine later.

mmalerba:
pr for shadows: #1857 

still need to do animations

thw0rted:
Hello from the future! I don't know if anybody is still subscribed, but I came here to find out how material2 handles changing its animation timing based on device type (as the spec requires) -- looks like maybe the answer at this time is that you don't?

Should I open a new ticket to bring timing in line with the spec or would it be better to reopen this one?  I am still trying to figure out how to actually decide what timing to use in my own animations.  Best I have right now is:

* if touch isn't supported, use desktop timing (150-200ms)
[...]

PR #222, opened 2016-03-22
feat(elevation): Add elevation core styles

traviskaufman:
This commit replaces `_shadows.scss` within `src/core/styles` with
`_elevation.scss`, according to the proposed design outlined
[here](https://goo.gl/Kq0k9Z). All components that were dependent on the
old shadows have been modified to use elevation. Note however that none
of the values in those old components have been changed; it's simply a
to the new API.

Part of #48.

### DIFF
--- a/src/core/style/_elevation.scss
+++ b/src/core/style/_elevation.scss
@@ -1,0 +1,179 @@
+/**
+ * A collection of mixins and CSS classes that can be used to apply elevation to a material
+ * element.
+ * See: https://www.google.com/design/spec/what-is-material/elevation-shadows.html
+ * Examples:
+ *
+ *
+ * .md-foo {
+ *   @include $md-elevation(2);
+ *
+ *   &:active {
+ *     @include $md-elevation(8);
+ *   }
+ * }
+ *
+ * <div id="external-card" class="md-elevation-z2"><p>Some content</p></div>
+ *
+ * For an explanation of the design behind how elevation is implemented, see the design doc at
+ * https://goo.gl/Kq0k9Z.
+ */
+
+// Colors for umbra, penumbra, and ambient shadows. As described in the design doc, each elevation
+// level is created using a set of 3 shadow values, one for umbra (the shadow representing the
+// space completely obscured by an object relative to its light source), one for penumbra (the
+// space partially obscured by an object), and one for ambient (the space which contains the object
+// itself). For a further explanation of these terms and their meanings, see
+// https://en.wikipedia.org/wiki/Umbra,_penumbra_and_antumbra.
+
+$_umbra-color: rgba(black, 0.2);
+$_penumbra-color: rgba(black, 0.14);
+$_ambient-color: rgba(black, 0.12);
+
+// Maps for the different shadow sets and their values within each z-space. These values were
+// created by taking a few reference shadow sets created by Google's Designers and interpolating
+// all of the values between them.
+
+$_umbra-elevation-map: (
+  0: '0px 0px 0px 0px #{$_umbra-color}',
+  1: '0px 2px 1px -1px #{$_umbra-color}',
+  2: '0px 3px 1px -2px #{$_umbra-color}',
+  3: '0px 3px 3px -2px #{$_umbra-color}',
+  4: '0px 2px 4px -1px #{$_umbra-color}',
+  5: '0px 3px 5px -1px #{$_umbra-color}',
+  6: '0px 3px 5px -1px #{$_umbra-color}',
+  7: '0px 4px 5px -2px #{$_umbra-color}',
+  8: '0px 5px 5px -3px #{$_umbra-color}',
+  9: '0px 5px 6px -3px #{$_umbra-color}',
+  10: '0px 6px 6px -3px #{$_umbra-color}',
+  11: '0px 6px 7px -4px #{$_umbra-color}',
+  12: '0px 7px 8px -4px #{$_umbra-color}',
+  13: '0px 7px 8px -4px #{$_umbra-color}',
+  14: '0px 7px 9px -4px #{$_umbra-color}',
+  15: '0px 8px 9px -5px #{$_umbra-color}',
+  16: '0px 8px 10px -5px #{$_umbra-color}',
+  17: '0px 8px 11px -5px #{$_umbra-color}',
+  18: '0px 9px 11px -5px #{$_umbra-color}',
+  19: '0px 9px 12px -6px #{$_umbra-color}',
+  20: '0px 10px 13px -6px #{$_umbra-color}',
+  21: '0px 10px 13px -6px #{$_umbra-color}',
+  22: '0px 10px 14px -6px #{$_umbra-color}',
+  23: '0px 11px 14px -7px #{$_umbra-color}',
+  24: '0px 11px 15px -7px #{$_umbra-color}'
+);
+
+$_penumbra-elevation-map: (
+  0: '0px 0px 0px 0px #{$_penumbra-color}',
+  1: '0px 1px 1px 0px #{$_penumbra-color}',
+  2: '0px 2px 2px 0px #{$_penumbra-color}',
+  3: '0px 3px 4px 0px #{$_penumbra-color}',
+  4: '0px 4px 5px 0px #{$_penumbra-color}',
+  5: '0px 5px 8px 0px #{$_penumbra-color}',
+  6: '0px 6px 10px 0px #{$_penumbra-color}',
+  7: '0px 7px 10px 1px #{$_penumbra-color}',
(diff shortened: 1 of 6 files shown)


src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+    $easing: $md-elevation-transition-timing-function) {
+  transition: md-elevation-transition-rule($duration, $easing);
+  will-change: $md-elevation-property;
+}
+
+// Provides external CSS classes for each elevation value. Each CSS class is formatted as
+// `md-elevation-z$zValue` where `$zValue` correspond to the z-space in which the element is
+// elevated to.
+@for $zValue from 0 through 24 {
+  .md-elevation-z#{$zValue} {
```
DevVersion: In Material 1, we deprecated the `z` suffix, and decided to use `dp`. Is here another background?

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+ *   @include $md-elevation(2);
+ *
+ *   &:active {
+ *     @include $md-elevation(8);
+ *   }
+ * }
+ * ```
+ *
+ * ```html
+ * <div id="external-card" class="md-elevation-z2"><p>Some content</p></div>
```
DevVersion: What will happen, when `ViewEncapsulation` is disabled and you import the `elevation` styles in those components. Won't that cause some duplicates? 

Or in general, how can the user access these classes? Will there be a standalone stylesheet?

src/core/style/_elevation.scss
traviskaufman: I talk about this in the [design doc](https://docs.google.com/document/d/1W3NGSLqDZzjbBBLW2C6y_6NUxtvdZAVaJvg58LY3Q0E/edit#heading=h.djieovs0b5kz). Essentially, z-space makes more sense to speak about elevation on the web platform. `dp` works great for a design spec but possibly isn't as semantic for web developers.

DevVersion: I agree with you. `dp` always remembers me on native app development. I was just curious, because we deprecated those in Material 1.

src/core/style/_elevation.scss
traviskaufman: If I understand your question correctly (which I _think_ I do but I'm not 100%), I imagine that if the user imports a vendor style such as those from `ngMaterial` once in a top-level bootstrap file or a `vendor.ts`-esque file, then these styles will only be loaded once and there won't be any `ViewEncapsulation` issues. 

I'm not as familiar with how angular caches loaded styles but would the framework be smart enough not to load styles more than once if the `ViewEncapsulation` is disabled? If not, let me know what makes the most sense in terms of exporting these styles.

DevVersion: Yes, I think you understand my question correctly. Let's take the standalone / bootstrap stylesheet for a moment out of the way.

I don't think Angular 2 scans the whole stylesheet for duplicate identifiers, that won't be even possible, since some selectors are just overriding a previous defined selector.

The point is, for example: we have the button component, which has view encapsulation disabled. ([Source](https://github.com/angular/material2/blob/master/src/components/button/button.ts#L30))

Then the buttons stylesheet will import the `_elevation.scss` file, this causes the `md-elevation-z*` classes to be generated.

So now imagine, if you have more than one component, which includes the elevation and has `ViewEncapsulation` disabled. This will cause several generations of the same classes.

In my opinion we should separate between a bootstrap / standalone stylesheet and the styles which are dynamically inserted into the DOM.

Removing those class generation would be also possible, since the mixins will only generate the specific elevation (thats awesome!)

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+// be used in the same way by clients.
+@mixin md-elevation-transition(
+    $duration: $md-elevation-transition-duration,
+    $easing: $md-elevation-transition-timing-function) {
+  transition: md-elevation-transition-rule($duration, $easing);
+  will-change: $md-elevation-property;
+}
+
+// Provides external CSS classes for each elevation value. Each CSS class is formatted as
+// `md-elevation-z$zValue` where `$zValue` correspond to the z-space in which the element is
```
kara: Typo: "corresponds" ?

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * ```html
+ * <div id="external-card" class="md-elevation-z2"><p>Some content</p></div>
+ * ```
+ *
+ * For an explanation of the design behind how elevation is implemented, see the design doc at
+ * https://goo.gl/Kq0k9Z.
+ */
+
+$_umbra-color: rgba(black, 0.2);
```
jelbourn: Are you able to comment on what each of `umbra`, `penumbra` and `ambient` mean?

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+$_penumbra-color: rgba(black, 0.14);
+$_ambient-color: rgba(black, 0.12);
+
+$_umbra-elevation-map: (
```
jelbourn: Add comment that these values have been hand-tuned by Google's designers and can't quite be generated programmatically?

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+);
+
+/**
+ * The css property used for elevation. In most cases this should not be changed. It is exposed
+ * as a variable for abstraction / easy use when needing to reference the property directly, for
+ * example in a `will-change` rule.
+ */
+$md-elevation-property: box-shadow !default;
+
+/**
```
jelbourn: optional: can scrunch short comments to one line:
```scss
/** The default duration value for elevation transitions. */
```

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+ * more than one property.
+ *
+ * ```scss
+ * .foo {
+ *   transition: md-elevation-transition-rule(), opacity 100ms ease;
+ *   will-change: $md-elevation-property, opacity;
+ * }
+ * ```
+ */
+@function md-elevation-transition-rule(
```
jelbourn: Just `md-elevation-transition` ?

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+    $duration: $md-elevation-transition-duration,
+    $easing: $md-elevation-transition-timing-function) {
+  @return #{$md-elevation-property} #{$duration} #{$easing};
+}
```
jelbourn: It would be nice if we could think of a name that more clearly distinguishes this from the below `md-elevation-transition`...

`md-elevation-transition-definition`? `md-get-elevation-transition-content`?

src/core/style/_elevation.scss
```diff
@@ -0,0 +1,186 @@
...
+    $duration: $md-elevation-transition-duration,
+    $easing: $md-elevation-transition-timing-function) {
+  transition: md-elevation-transition-rule($duration, $easing);
+  will-change: $md-elevation-property;
+}
+
+// Provides external CSS classes for each elevation value. Each CSS class is formatted as
+// `md-elevation-z$zValue` where `$zValue` correspond to the z-space in which the element is
+// elevated to.
+@for $zValue from 0 through 24 {
```
jelbourn: Could you pop these concrete classes out into a separate (non-partial) file?

jelbourn:
@traviskaufman The CI is failing due to the `` ` `` characters in elevation.scss, which the `inline-resources` utility can't cope with.

src/core/style/_elevation.scss
traviskaufman: The [CSSOM](https://www.w3.org/TR/cssom-1/#css-declarations) spec refers to this as a _value_, and refers to the receiver for the value as a _property_. How about `md-elevation-transition-property-value`?

jelbourn: That sounds good

traviskaufman:
@jelbourn all changes made and CI issue fixed

src/core/style/core.scss
```diff
@@ -0,0 +1,12 @@
+// Core styles that can be used to apply material design treatments to any element.
```
jelbourn: This is good for now, but FYI I'm going to move it to `src/main.scss` once my outstanding PR lands.

jelbourn:
LGTM

Would be good if one of @kara / @hansl / @robertmesserle also took a look

hansl:
:+1: this LGTM.

ross-nordstrom:
At risk of getting the famous "we haven't done theming yet" response, is there a way to use `md-elevation` in apps using Material2?

PR #225, opened 2016-03-23
feat(overlay): add overlay container & styles

jelbourn:
R: @hansl @kara

### DIFF
--- a/src/core/overlay/overlay-container.ts
+++ b/src/core/overlay/overlay-container.ts
@@ -1,0 +1,14 @@
+import {DOM} from '../platform/dom/dom_adapter';
+
+
+/**
+ * Create the overlay container element, which is simply a div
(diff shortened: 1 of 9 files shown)


src/core/overlay/overlay-container.ts
```diff
@@ -0,0 +1,14 @@
+import {DOM} from '../platform/dom/dom_adapter';
+
+
+/**
+ * Create the overlay container element, which is simply a div
+ * with the 'md-overlay-container' class on the document body.
+ */
+export function getOverlayContainer(): Element {
```
DevVersion: Wouldn't it make more sense to call that function `createOverlayContainer()`?

jelbourn: That is better, done.

PR #227, opened 2016-03-24
fix(checkbox): Handle transition when unset checkbox is interacted with

### LINKED ISSUE #183: mdCheckbox: Styles not applied correctly when not explicitly given a `[checked]` value
When a checkbox is not explicitly given a `[checked]` input value, the first time it's clicked it simply fades in instead of drawing itself in. This is likely an issue of the svg `stroke-dashoffset` not being applied correctly within the unchecked state.

traviskaufman:
This commit fixes a bug where the initial `unchecked -> checked`
animation wasn't firing correctly when the checkbox was interacted with
from its initial state (e.g. when `[checked]` or `[(ngModel)]` isn't
bound to the component).

Fixes #183.


jelbourn:
LGTM

PR #229, opened 2016-03-25
feat(overlay): add global position strategy

jelbourn:
R: @hansl @robertmesserle @kara 

After much API discussion, this is what I went with. It's probable that this global strategy gets refactored once we actually start using it to build components and find it lacking.

### DIFF
--- a/src/core/overlay/position/global-position-strategy.ts
+++ b/src/core/overlay/position/global-position-strategy.ts
@@ -1,0 +1,112 @@
+import {PositionStrategy} from './position-strategy';
+import {DOM} from '../../platform/dom/dom_adapter';
+
+
+/**
+ * A strategy for positioning overlays. Using this strategy, an overlay is given an
+ * explicit position relative to the browser's viewport.
+ */
+export class GlobalPositionStrategy implements PositionStrategy {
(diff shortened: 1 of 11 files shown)


src/core/overlay/position/global-position-strategy.ts
```diff
@@ -0,0 +1,87 @@
...
+    this._cssPosition = 'fixed';
+    return this;
+  }
+
+  absolute() {
+    this._cssPosition = 'absolute';
+    return this;
+  }
+
+  top(value: string) {
```
hansl: Should default value also be `0px`?

src/core/overlay/position/global-position-strategy.ts
```diff
@@ -0,0 +1,87 @@
...
+    this._top = value;
+    return this;
+  }
+
+  left(value: string) {
+    this._left = value;
```
hansl: You should set `this._right = ''` and `this._translateX = []` to counter `.center(...).left(...)` and `.right().left()`. Same with others.

jelbourn:
@hansl addressed comments

hansl:
LGTM.

jelbourn:
@hansl removed `classList`

PR #238, opened 2016-03-28
feat(): add aria live announcer

### LINKED ISSUE #106: Build aria-live announcer service / component

DevVersion:
Fixes #106

### DIFF
--- a/src/core/live-announcer/live-announcer.ts
+++ b/src/core/live-announcer/live-announcer.ts
@@ -1,0 +1,44 @@
+import {Injectable} from 'angular2/core';
+
+export type AriaLivePoliteness = 'off' | 'polite' | 'assertive';
+
+@Injectable()
+export class MdLiveAnnouncer {
+
+  private _liveElement: Element;
+
+  constructor() {
+    this._liveElement = this._createLiveElement();
+  }
+
+  /**
+   * @param message Message to be announced to the screenreader
+   * @param politeness The politeness of the announcer element.
+   */
+  announce(message: string, politeness: AriaLivePoliteness = 'polite'): void {
+    this._liveElement.textContent = '';
+
+    // TODO: ensure changing the politeness works on all environments we support.
+    this._liveElement.setAttribute('aria-live', politeness);
+
+    // This 100ms timeout is necessary for some browser + screen-reader combinations:
+    // - Both JAWS and NVDA over IE11 will not announce anything without a non-zero timeout.
+    // - With Chrome and IE11 with NVDA or JAWS, a repeated (identical) message won't be read a
+    //   second time without clearing and then using a non-zero delay.
+    // (using JAWS 17 at time of this writing).
+    setTimeout(() => this._liveElement.textContent = message, 100);
+  }
+
+  private _createLiveElement(): Element {
+    let liveEl = document.createElement('div');
+
+    liveEl.classList.add('md-live-announcer');
+    liveEl.setAttribute('aria-atomic', 'true');
+    liveEl.setAttribute('aria-live', 'polite');
+
+    document.body.appendChild(liveEl);
+
+    return liveEl;
+  }
+
+}
--- a/src/core/live-announcer/README.md
+++ b/src/core/live-announcer/README.md
@@ -1,0 +1,30 @@
+# MdLiveAnnouncer
+`MdLiveAnnouncer` is a service, which announces messages to several screenreaders.
+
+### Methods
+
+| Name |  Description |
+| --- | --- |
+| `announce(message, politeness)` | This announces a text message to the supported screenreaders. <br><br>The politeness parameter sets the `aria-live` attribute on the announcer element |
+
+### Examples
+The service can be injected in a component.
+```ts
+@Component({
+  selector: 'my-component'
+  providers: [MdLiveAnnouncer]
+})
+export class MyComponent {
+
+ constructor(live: MdLiveAnnouncer) {
+   live.announce("Hey Google");
+ }
+
+}
+```
+
+### Supported Screenreaders
+- JAWS (Windows)
+- NVDA (Windows)
+- VoiceOver (OSX and iOS)
+- TalkBack (Android)
--- a/src/core/live-announcer/live-announcer.scss
+++ b/src/core/live-announcer/live-announcer.scss
@@ -1,0 +1,5 @@
+@import 'mixins';
+
+.md-live-announcer {
+  @include md-visually-hidden();
+}
--- a/src/core/live-announcer/live-announcer.spec.ts
+++ b/src/core/live-announcer/live-announcer.spec.ts
@@ -1,0 +1,121 @@
+import {
+  inject,
+  TestComponentBuilder,
+  ComponentFixture,
+  fakeAsync,
+  flushMicrotasks,
+  tick,
+  beforeEachProviders
+} from 'angular2/testing';
+import {
+  it,
+  describe,
+  expect,
+  beforeEach,
+} from '../../core/facade/testing';
+import {Component} from 'angular2/core';
+import {By} from 'angular2/platform/browser';
+import {MdLiveAnnouncer} from './live-announcer';
+
+export function main() {
+  describe('MdLiveAnnouncer', () => {
+    let live: MdLiveAnnouncer;
+    let builder: TestComponentBuilder;
+    let liveEl: Element;
+
+    beforeEachProviders(() => [MdLiveAnnouncer]);
+
+    beforeEach(inject([TestComponentBuilder, MdLiveAnnouncer],
+      (tcb: TestComponentBuilder, _live: MdLiveAnnouncer) => {
+        builder = tcb;
+        live = _live;
+        liveEl = getLiveElement();
+    }));
+
+    afterEach(() => {
+      // In our tests we always remove the current live element, because otherwise we would have
+      // multiple live elements due multiple service instantiations.
+      liveEl.parentNode.removeChild(liveEl);
+    });
+
+    it('should correctly update the announce text', fakeAsyncTest(() => {
+      let appFixture: ComponentFixture = null;
+
+      builder.createAsync(TestApp).then(fixture => {
+        appFixture = fixture;
+      });
+
+      flushMicrotasks();
+
+      let buttonElement = appFixture.debugElement
+        .query(By.css('button')).nativeElement;
+
+      buttonElement.click();
+
+      // This flushes our 100ms timeout for the screenreaders.
+      tick(100);
+
+      expect(liveEl.textContent).toBe('Test');
+    }));
+
+    it('should correctly update the politeness attribute', fakeAsyncTest(() => {
+      let appFixture: ComponentFixture = null;
+
+      builder.createAsync(TestApp).then(fixture => {
+        appFixture = fixture;
+      });
+
+      flushMicrotasks();
+
+      live.announce('Hey Google', 'assertive');
+
+      // This flushes our 100ms timeout for the screenreaders.
+      tick(100);
+
+      expect(liveEl.textContent).toBe('Hey Google');
+      expect(liveEl.getAttribute('aria-live')).toBe('assertive');
+    }));
+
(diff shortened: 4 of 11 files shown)


src/components/aria-live/aria-live.scss
```diff
@@ -0,0 +1,17 @@
+:host {
+
+  // This hides the announcer element visually.
+  // That means it's still accessible for screen-readers but not visible.
+  ._md-aria-announcer {
```
jelbourn: You can omit the underscore from the name.

src/components/aria-live/aria-live.scss
```diff
@@ -0,0 +1,17 @@
...
+    border: 0;
+    clip: rect(0 0 0 0);
+    height: 1px;
+    margin: -1px;
+    overflow: hidden;
+    padding: 0;
+    position: absolute;
+    text-transform: none;
+    width: 1px;
```
jelbourn: Can you break this styling out to a scss mixin called `md-visually-hidden` ?

src/components/aria-live/aria-live.html
```diff
@@ -0,0 +1,2 @@
+<ng-content></ng-content>
```
jelbourn: I don't think this should actually have an `ng-content`

src/components/aria-live/aria-live.ts
```diff
@@ -0,0 +1,31 @@
...
+  private announcerEl: HTMLElement;
+
+  constructor(private elementRef: ElementRef) { }
+
+  ngAfterContentInit() {
+    this.announcerEl = DOM.querySelector(this.elementRef.nativeElement, '._md-aria-announcer');
+  }
+
+  announce(message: string): void {
+    this.announcerEl.textContent = '';
```
jelbourn: Did it not work to simply put a `{{ }}` binding inside the `aria-live` element, where the bound property is only updated after a timeout?

src/components/aria-live/aria-live.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    // The most screen-readers need a delay after the clear action, to probably recognise the
+    // announcement. 100ms is a average delay, which works for the most supported readers.
```
jelbourn: Can you use my comment from my original code verbatim?
```
  // This 100ms timeout is necessary for some browser + screen-reader combinations:
  // - Both JAWS and NVDA over IE11 will not announce anything without a non-zero timeout.
  // - With Chrome and IE11 with NVDA or JAWS, a repeated (identical) message won't be read a
  //   second time without clearing and then using a non-zero delay.
  // (using JAWS 17 at time of this writing).
```

src/components/aria-live/aria-live.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  private announcerEl: HTMLElement;
+
+  constructor(private elementRef: ElementRef) { }
+
+  ngAfterContentInit() {
+    this.announcerEl = DOM.querySelector(this.elementRef.nativeElement, '._md-aria-announcer');
+  }
+
+  announce(message: string): void {
```
jelbourn: It would be good if the announce function took a second, optional argument for the politeness.

src/components/aria-live/aria-live.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  describe('MdAriaLive', () => {
+    let builder: TestComponentBuilder;
+    let timeoutMock: any;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+    }));
+
+    beforeAll(() => {
+      timeoutMock = mockTimeout();
```
jelbourn: You should be able to do this more easily with Angular 2's `fakeAsync`; check out the portal tests for an example.

src/components/aria-live/README.md
```diff
@@ -0,0 +1,37 @@
+# MdAriaLive
```
jelbourn: I kind of want to call this `md-live-announcer`. 

@hansl @kara @robertmesserle thoughts?

src/components/aria-live/aria-live.spec.ts
```diff
@@ -0,0 +1,95 @@
+import {
```
jelbourn: This may seem weird, but could you move this component into `core/` for now? I don't yet want to publish it as a first-class component.

jelbourn:
cc @marcysutton in case you're interested :)

src/components/aria-live/README.md
hansl: Yes, MdLiveAnnouncer would probably be better. Do we want this to be a component or a service (or both)?

jelbourn: I believe this does make the most sense as a component, since you can inject it both ways.

src/components/aria-live/aria-live.spec.ts
DevVersion: Can do that, but it seems like you're only compiling styles within `src/core/styles`, I can add the component to BrocolliSass as well, or we keep it temporary in the `src/components`

src/components/aria-live/aria-live.scss
DevVersion: Yes right, seems to be unnecessary, because of the view encapsulation.

src/components/aria-live/aria-live.scss
DevVersion: Done.

src/components/aria-live/aria-live.html
DevVersion: Oh, if it has no `ng-content`, then it can't be a parent component. That would mean, we are not able to inject the component. 

Otherwise we need to provide a global OpaqueToken or something else. 
In my opinion, injecting it as the parent component is very elegant.

What are your thoughts?

src/components/aria-live/aria-live.ts
DevVersion: I tested that out, but when we we use a bound text property, then we have to set the property always to undefined, to clear the text content, setting to `''` isn't working for me in this case. (Voiceover iOS)

Additionally, I think, that a change detection is here not necessary. 
If you have other thoughts let me know, and I'll change it to Angular's Property Bound.

src/components/aria-live/aria-live.spec.ts
DevVersion: Awesome, that's what I've been searching for.

src/components/aria-live/README.md
DevVersion: Changed it to `MdLiveAnnouncer`.

src/components/aria-live/aria-live.ts
DevVersion: Done

src/components/aria-live/aria-live.ts
DevVersion: Good idea, will implement that.

src/components/live-announcer/README.md
```diff
@@ -0,0 +1,37 @@
+# MdAriaLive
```
jelbourn: The readme still uses `aria-live`

src/components/live-announcer/live-announcer.ts
```diff
@@ -0,0 +1,36 @@
...
+  styleUrls: ['./components/live-announcer/live-announcer.css'],
+})
+export class MdLiveAnnouncer implements AfterContentInit {
+
+  private announcerEl: HTMLElement;
+
+  constructor(private elementRef: ElementRef) { }
+
+  ngAfterContentInit() {
+    this.announcerEl = DOM.querySelector(this.elementRef.nativeElement, '.md-live-announcer');
```
jelbourn: Why did you change this away from using the `Renderer` ? If the binding doesn't work, using the `Renderer` is preferable.

src/components/live-announcer/README.md
DevVersion: Oh, shameful. I was sure to change that.

jelbourn:
@DevVersion sorry to change course on you with this PR, but I talked with hansl@ and we realized that doing this as a service would actually make more sense. 
* It ensures that there's only one announcer globally
* It can be injected anywhere in the app without having to be a parent of anything

The service would be responsible for making its own DOM element and appending it to the document body (and also checking whether there is already an announcer element present).

DevVersion:
@jelbourn That's no problem. I will change that as soon as possible and ping you. (lot of stress atm)

jelbourn:
@DevVersion No worries! We don't have any immediate need for this.

DevVersion:
@jelbourn Updated it. Please take a look again.

src/core/live-announcer/live-announcer.ts
```diff
@@ -0,0 +1,52 @@
+import {Injectable} from 'angular2/core';
+import {DOM} from '../../core/platform/dom/dom_adapter';
+
+@Injectable()
+export class MdLiveAnnouncer {
+
+  private announcerEl: Element;
```
jelbourn: How about `_liveElement` ?

src/core/live-announcer/live-announcer.ts
```diff
@@ -0,0 +1,52 @@
...
+    // This 100ms timeout is necessary for some browser + screen-reader combinations:
+    // - Both JAWS and NVDA over IE11 will not announce anything without a non-zero timeout.
+    // - With Chrome and IE11 with NVDA or JAWS, a repeated (identical) message won't be read a
+    //   second time without clearing and then using a non-zero delay.
+    // (using JAWS 17 at time of this writing).
+    setTimeout(() => this.announcerEl.textContent = message, 100);
+  }
+
+  private _getAnnouncerElement(): Element {
+    let documentBody = DOM.getGlobalEventTarget('body');
```
jelbourn: Since we decided to drop TS support yesterday, you can simply use `document.createElement`, etc.

Instead of looking for an existing announcer element, the service should just always create one.

src/core/live-announcer/live-announcer.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  constructor() {
+    this.announcerEl = this._getAnnouncerElement();
+  }
+
+  /**
+   * @param message Message to be announced to the screenreader
+   * @param politeness The politeness of the announcer element.
+   */
+  announce(message: string, politeness = 'polite'): void {
```
jelbourn: Cool TypeScript trick: you can introduce your own type for the politeness

```
// higher up
export type AriaLivePoliteness: 'off' | 'polite' | 'assertive';

// then, here
 announce(message: string, politeness: AriaLivePoliteness = 'polite'): void
```
This enforces that people only provide a valid politeness

src/core/live-announcer/live-announcer.ts
marcysutton: Just make sure any straggling elements get cleaned up so there aren't conflicts.

src/core/live-announcer/live-announcer.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.announcerEl.textContent = '';
+
+    this.announcerEl.setAttribute('aria-live', politeness);
```
marcysutton: I think we'll need to validate that a changing politeness level is correctly utilized by Assistive Technologies.

src/core/live-announcer/live-announcer.ts
jelbourn: The service is meant to be a singleton, so it should only ever create one element for itself. It might make sense to add another API later to pass it an existing `aria-live` element to use instead of creating its own.

DevVersion: Yes we can do this later, by providing an OpaqueToken IMO.

src/core/live-announcer/live-announcer.ts
jelbourn: That's a good point; @DevVersion can you add a `TODO`:
```ts
TODO: ensure changing the politness works on all environments we support.
```

jelbourn:
LGTM

jelbourn:
@DevVersion Just notices one last thing from CI: IE11 doesn't support `Element.prototype.remove`. Once that's resolved I can merge

PR #239, opened 2016-03-28
docs(dart): fix outdated link for generic functions

DevVersion:
Just a very small PR, which fixes the wrong link (since the new typescript website) for the typescript generic functions.


jelbourn:
LGTM

PR #242, opened 2016-03-29
style(_button-base.scss): change Length Unit from px to %

RichOren:
`$md-icon-button-border-radius` change to be 50% of size instead of 50px.


hansl:
Could you clarify _why_ you did this PR? Your message only state what you did, which is evident, but not why you did it.

RichOren:
The 50px Length Unit was probably a typo by the original author who mistakenly use px instead of %. $md-icon-button-border-radius needs to be 50% of width not 50px.

jelbourn:
@hansl the `50px` was a mistake, it should always have been `50%`

LGTM

PR #246, opened 2016-03-29
fix: vertically align buttons with the surrounding text.

hansl:
Also add a section for baseline vertical alignment.

### DIFF
--- a/src/demo-app/button/button-demo.html
+++ b/src/demo-app/button/button-demo.html
@@ -1,5 +1,18 @@
 <div class="demo-button">
   <section>
+    Link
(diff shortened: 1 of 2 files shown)


hansl:
@kara @robertmesserle

src/demo-app/button/button-demo.html
```diff
@@ -1,5 +1,11 @@
 <div class="demo-button">
   <section>
+    Link
+    <a href="http://www.google.com" md-button color="primary">link</a>
+    Button
+    <button md-button>flat</button>
```
hansl: Done. Added another section.

PR #252, opened 2016-04-01
chore: Refactor the situation a bit for protractor. Can run locally now

hansl:
Also removed the screenshot test. It will live forever in the git history.


hansl:
FYI @jelbourn @robertmesserle

robertmesserle:
LGTM

PR #253, opened 2016-04-01
chore: moving e2e tests to TypeScript, including typings for protractor.

hansl:


hansl:
FYI @jelbourn  @robertmesserle

hansl:
The change to forbidden-identifiers is to not block on something like `exit` because it doesn't check if `xit` is preceded/succeeded by a non word token.

e2e/typings/es6-shim/es6-shim.d.ts
```diff
@@ -0,0 +1,670 @@
+// Compiled using typings@0.6.10
```
jelbourn: Why do we have to copy these typings into our project instead of installing them?

jelbourn:
Looks like the changes to the forbidden keywords script is broken on CI

hansl:
I'm going to move the forbidden keyword script into a separate PR. Still using this PR for testing though

jelbourn:
LGTM

PR #255, opened 2016-04-03
chore: strengthen and improve forbidden-keywords.

hansl:
* Move the lint and forbidden-keywords check to a required travis matrix dimension. When those fail, the tests will still be run.
* Only install Dart SDK on tests that require Dart.


robertmesserle:
LGTM

PR #263, opened 2016-04-05
fix(checkbox): Remove GPU layer promotion for mark elements

traviskaufman:
This commit removes the CSS that promotes checkmarks to the GPU layer,
due to an issue in Chrome which makes transitions appear incorrect when
the promotions are applied.

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -202,9 +202,7 @@ $_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1)
  */
 %md-checkbox-mark {
   $width-padding-inset: 2 * $md-checkbox-border-width;
-
   width: calc(100% - #{$width-padding-inset});
-  will-change: opacity, transform;
 }
 
 /**


traviskaufman:
More context: If you look at an earlier version of [the demo plunker](http://plnkr.co/edit/P7qce8lN9n2flS6kBhDy?p=preview) (e.g. version "79" ) and compare it to the most recent, you'll see that the "indeterminate to checked" transition does not work _consistently_ on the previous version but does on the most recent. Trying to put together a more bare-bones example of this to file a bug for, as the opt-in layer promotion from `will-change` would be nice.

hansl:
So is this a temporary fix for Chrome < 49 ?

hansl:
LGTM.

jelbourn:
Closed via 82a22a71d02d3f9d69c975e5fd467463af993c87

traviskaufman:
@hansl sorry follow-up to your question: it's actually a fix for Chrome >= 49. It wasn't happening until I upgraded to latest stable (and it was happening on Canary as well). As I mentioned I wanna get a simpler demo of the issue so I can send it to the team to reproduce.

PR #266, opened 2016-04-05
chore: remove facades and dart-specific patterns

jelbourn:
R: @hansl @kara @robertmesserle

* Replaces `@OneOf` with string literal types.
* Removes facades and uses standard TS / JS APIs
* Uses standard `angular2/testing` imports

This leaves only two things from facades: `BaseException` and `PromiseCompleter`, which don't have a complete drop-in replacement.


src/main.ts
```diff
@@ -1,13 +1,10 @@
 import {bootstrap} from 'angular2/platform/browser';
 import {DemoApp} from './demo-app/demo-app';
 import {ROUTER_PROVIDERS} from 'angular2/router';
-import {BrowserDomAdapter} from './core/platform/browser/browser_adapter';
 import {OVERLAY_CONTAINER_TOKEN} from './core/overlay/overlay';
 import {provide} from 'angular2/core';
 import {createOverlayContainer} from './core/overlay/overlay-container';
 
-BrowserDomAdapter.makeCurrent();
```
hansl: This isn't needed anymore?

hansl:
One interrogation, but LGTM!  :shipit:

hansl:
Actually, I'm looking at the errors in Travis and these seem legitimate, based on the change from `OneOf(....)` to string enumeration types. Could you take a deeper look?

jelbourn:
@hansl all better now; it was mostly that `@OneOf` sneakily also set a default value. Also changed sidenav's `toggle` to use a default param and removed a progress-circle test for a behavior it shouldn't actually exhibit.

src/main.ts
jelbourn: Nay!

jelbourn:
Also added a type alias in `Dir` to reduce repetition.

src/core/annotations/field-value.ts
```diff
@@ -22,13 +19,13 @@ function booleanFieldValueFactory() {
...
     // Use a fallback if Symbol isn't available.
-    const localKey = isPresent(Symbol) ? Symbol(key) : `__md_private_symbol_${key}`;
+    const localKey = Symbol ? Symbol(key) : `__md_private_symbol_${key}`;
     target[localKey] = defaultValue;
 
     Object.defineProperty(target, key, {
       get() { return this[localKey]; },
       set(value: boolean) {
-        this[localKey] = isPresent(value) && value !== null && String(value) != 'false';
+        this[localKey] = value != null && `${value}` !== 'false';
```
hansl: `\`${value}\`` kinda weirds me out. Could you use `toString()` or `'' + ` ? or just `value != 'false'`?

hansl:
Few nits, no blockers. LGTM.

hansl:
Also, this means that unless the template compiler verify typings (which I'm not sure it does), the replacement of `OneOf` by native string literal types may lead to undefined behaviour for typos or invalid values.

src/core/annotations/field-value.ts
jelbourn: `value != false` isn't allowed by TypeScript (comparing string to boolean).

To me `${value}` is way cleaner than using `toString()` and more readable than `'' + value` (due to the boolean operator) while being the same number of characters).

PR #270, opened 2016-04-07
fix: do not throw when there's no sidenav.

### LINKED ISSUE #269: MdSidenavlayout should not throw when no MdSidenav is present

Bug

If I have an `MdSidenavLayout` without an `MdSidenav` child, an exception will be raised. For instance, I wanted to avoid pre-compiling sidenav content in a pre-generated html document, so I threw an ngIf on the md-sidenav to hide it, which caused an error.

Definitely alpha.1, and I think this is still present in alpha.2

hansl:
Fixes #269.

### DIFF
--- a/src/components/sidenav/sidenav.ts
+++ b/src/components/sidenav/sidenav.ts
@@ -20,11 +20,6 @@ import {PromiseCompleter} from 'angular2/src/facade/promise';
 
 
 /**
- * Exception thrown when a MdSidenavLayout is missing both sidenavs.
- */
-export class MdMissingSidenavException extends BaseException {}
-
-/**
  * Exception thrown when two MdSidenav are matching the same side.
  */
 export class MdDuplicatedSidenavException extends BaseException {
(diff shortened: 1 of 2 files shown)


hansl:
@jelbourn @jeffbcross

src/components/sidenav/sidenav.ts
```diff
@@ -286,9 +281,6 @@ export class MdSidenavLayout implements AfterContentInit {
    */
   private _validateDrawers() {
     this._start = this._end = null;
-    if (this._sidenavs.length === 0) {
-      throw new MdMissingSidenavException();
-    }
```
jelbourn: Add a test? While the code isn't explicitly throwing an exception when a sidenav is missing, having a test would guard against accidentally throwing an error if the sidenav is missing (such as, for example, a null dereference).

hansl: Donerino.

jelbourn:
LGTM

PR #281, opened 2016-04-08
md-icon

dozingcat:
Preview PR for md-icon. Doc at https://docs.google.com/a/google.com/document/d/1zWTjnSLFPV4II9IxpIehyeSFtnzw_efdy7kEerMRIuo/edit?usp=sharing


src/components/icon/icon-provider.ts
```diff
@@ -0,0 +1,255 @@
...
+  *  Configuration for an icon, possibly including the cached SVG element.
+  */
+class IconConfig {
+  svgElement: SVGElement = null;
+  constructor(public url: string, public viewBoxSize: number) {
+  }
+}
+
+@Injectable()
+export class MdIconProvider {
```
jelbourn: I'm trying to come up with a better name for this class, since it would be good to move away from the Angular 1 "provider" concept. What do you think about `MdIconRegistry` ?

src/components/icon/icon-provider.ts
```diff
@@ -0,0 +1,255 @@
+import {Injectable, Renderer} from 'angular2/core';
+import {Http, Response, HTTP_PROVIDERS} from 'angular2/http';
+import {AsyncSubject, Observer, Observable} from 'rxjs/Rx';
+
+/**
+  *  Configuration for an icon, possibly including the cached SVG element.
+  */
+class IconConfig {
+  svgElement: SVGElement = null;
+  constructor(public url: string, public viewBoxSize: number) {
```
jelbourn: Do you know why we do anything with viewBox? I know material1 does the same thing, but I'm iffy on the background.

src/components/icon/icon-provider.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+}
+
+@Injectable()
+export class MdIconProvider {
+  // IconConfig objects and cached SVG elements for individual icons.
+  // First level of cache is icon set (which is the empty string for the default set).
+  // Second level is the icon name within the set.
+  private _iconConfigs = new Map<string, Map<string, IconConfig>>();
```
jelbourn: Rather than doing a Map of Map, why not just treat the key as `namespace:icon-name` ? For the default namespace, it's just the empty string.

src/components/icon/icon-provider.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  // IconConfig objects and cached SVG elements for icon sets.
+  // These are stored only by set name, but multiple URLs can be registered under the same name.
+  private _iconSetConfigs = new Map<string, [IconConfig]>();
```
jelbourn: Does the format `Map<string, IconConfig[]>` work here?

src/components/icon/icon-provider.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  private _inProgressUrlFetches = new Map<string, Observable<string>>();
+  
+  private _defaultViewBoxSize = 24;
+  private _defaultFontSetClass = 'material-icons';
+  
+  constructor(private _http: Http) {
+  }
+  
+  public addIcon(iconName: string, url: string, viewBoxSize:number=0): MdIconProvider {
```
jelbourn: Cool TypeScript trick: you can make your return type `this`
```ts
public addIcon(name: string, url: string, viewBoxSize: number = 0): this { ... }
```

src/components/icon/icon-provider.ts
```diff
@@ -0,0 +1,255 @@
...
+    return this.addIconInSet('', iconName, url, viewBoxSize);
+  }
+  
+  public addIconInSet(
```
jelbourn: I think we can separate the concept of icon namespaces and icon sets. We can treat an icon-set is a bundle of icons all in one file, which can be added to any namespace. Perhaps it makes sense to just use the term "icon bundle".

```ts
addIcon(name, url, opt_namespace)

addIconBundle(url, opt_namespace)
```

Then again, maybe it does make more sense to _always_ treat a "bundle" as a namespace. Let me run this idea by the rest of the team.

src/components/icon/icon-provider.ts
jelbourn: Oh, I see now that this is a tuple type, not an array (which isn't something I had seen used before). Why is it a tuple?

src/components/icon/icon-provider.ts
dozingcat: Yeah it seems like an obscure feature. You can use it to have a version of the icon that's zoomed in or out, but only from one corner because the top and left coordinates are always 0. I don't know of any especially compelling use cases, happy to remove it.

src/components/icon/icon-provider.ts
dozingcat: SG

src/components/icon/icon-provider.ts
dozingcat: It seemed slightly cleaner to use nested maps rather than encoding the nesting in the key. No strong opinion either way though.

src/components/icon/icon-provider.ts
dozingcat: Yes IconConfig[] works. It is just supposed to be an array.

src/components/icon/icon-provider.ts
dozingcat: Neat.

src/components/icon/icon-provider.ts
dozingcat: That's basically what it does now, I think. You can do addIconSet('animals', 'animals.svg') and addIconInSet('animals', 'cat', 'cat1.svg') and then svgIcon="animals:cat" will use the individual icon while "animals:dog" will search the set/bundle. And you can pass an empty namespace to addIconSet to add it to the default namespace.

I like your names and API, assuming we can get rid of the current optional viewBoxSize parameter.

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  constructor(public url: string) {
+  }
+}
+
+@Injectable()
+export class MdIconRegistry {
+  // IconConfig objects and cached SVG elements for individual icons.
+  // First level of cache is namespace (which is the empty string if not specified).
+  // Second level is the icon name within the namespace.
+  private _iconConfigs = new Map<string, Map<string, IconConfig>>();
```
jelbourn: My preference would be for flattening the icon namespace + name into a single key, rather than doing a map-of-maps. @hansl do you have an opinion here?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  svgElement: SVGElement = null;
+  constructor(public url: string) {
+  }
+}
+
+@Injectable()
+export class MdIconRegistry {
+  // IconConfig objects and cached SVG elements for individual icons.
+  // First level of cache is namespace (which is the empty string if not specified).
+  // Second level is the icon name within the namespace.
```
jelbourn: Use `/** */` style comments for method and member descriptions

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  *  Configuration for an icon, including the URL and possibly the cached SVG element.
+  */
+class IconConfig {
+  svgElement: SVGElement = null;
+  constructor(public url: string) {
+  }
+}
+
+@Injectable()
+export class MdIconRegistry {
```
jelbourn: Need description on `MdIconRegistry`

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  private _iconSetConfigs = new Map<string, IconConfig[]>();
+
+  // Cache for icons loaded by direct URLs.
+  private _cachedIconsByUrl = new Map<string, SVGElement>();
+
+  private _fontClassNamesByAlias = new Map<string, string>();
+
+  private _inProgressUrlFetches = new Map<string, Observable<string>>();
+
+  private _defaultFontSetClass = 'material-icons';
```
jelbourn: Need descriptions for these members

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
+import {Injectable} from 'angular2/core';
+import {Http} from 'angular2/http';
+import {Observable} from 'rxjs/Rx';
+
+/**
+  *  Configuration for an icon, including the URL and possibly the cached SVG element.
+  */
+class IconConfig {
+  svgElement: SVGElement = null;
+  constructor(public url: string) {
```
jelbourn: nit: empty functions can close on the same line
```ts
constructor(public url: string) { }
```
(here and elsewhere)

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  private _iconConfigs = new Map<string, Map<string, IconConfig>>();
+
+  // IconConfig objects and cached SVG elements for icon sets.
+  // Multiple icon sets can be registered under the same namespace.
+  private _iconSetConfigs = new Map<string, IconConfig[]>();
+
+  // Cache for icons loaded by direct URLs.
+  private _cachedIconsByUrl = new Map<string, SVGElement>();
+
+  private _fontClassNamesByAlias = new Map<string, string>();
```
jelbourn: `_fontCssClassesByAlias` ?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
+import {Injectable} from 'angular2/core';
+import {Http} from 'angular2/http';
+import {Observable} from 'rxjs/Rx';
+
+/**
+  *  Configuration for an icon, including the URL and possibly the cached SVG element.
+  */
+class IconConfig {
```
jelbourn: `SvgIconConfig` ?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  private _fontClassNamesByAlias = new Map<string, string>();
+
+  private _inProgressUrlFetches = new Map<string, Observable<string>>();
+
+  private _defaultFontSetClass = 'material-icons';
+
+  constructor(private _http: Http) {
+  }
+
```
jelbourn: Add method descriptions

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+    if (!iconSetMap) {
+      iconSetMap = new Map<string, IconConfig>();
+      this._iconConfigs.set(namespace, iconSetMap);
+    }
+    iconSetMap.set(iconName, new IconConfig(url));
+    return this;
+  }
+
+  public addIconSet(url: string): this {
+      return this.addIconSetInNamespace('', url);
```
jelbourn: Too much indent

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  private _fontClassNamesByAlias = new Map<string, string>();
+
+  private _inProgressUrlFetches = new Map<string, Observable<string>>();
+
+  private _defaultFontSetClass = 'material-icons';
+
+  constructor(private _http: Http) {
+  }
+
+  public addIcon(iconName: string, url: string): this {
```
jelbourn: I'm debating whether the svg-based methods should have "svg" in the name, e.g., `addSvgIcon`.
@hansl @kara @robertmesserle thoughts?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+    const config = new IconConfig(url);
+    if (this._iconSetConfigs.has(namespace)) {
+      this._iconSetConfigs.get(namespace).push(config);
+    } else {
+      this._iconSetConfigs.set(namespace, [config]);
+    }
+    return this;
+  }
+
+  public registerFontSet(alias: string, className?: string): this {
```
jelbourn: You could do 
```ts
registerFontSet(alias: string, className = alias): this {
```

src/components/icon/icon-registry.ts
jelbourn: We've been omitting the `public` keyword so far since it's the default.

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  public setDefaultFontSetClass(className: string) {
+    this._defaultFontSetClass = className;
+    return this;
+  }
+
+  public getDefaultFontSetClass(): string {
+    return this._defaultFontSetClass;
+  }
+
+  loadIconFromNamespaceByName(namespace: string, iconName: string): Observable<SVGElement> {
```
jelbourn: Maybe just `getSvgByName` ?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+    // Return (copy of) cached icon if possible.
+    if (this._iconConfigs.has(namespace) && this._iconConfigs.get(namespace).has(iconName)) {
+      const config = this._iconConfigs.get(namespace).get(iconName);
+      if (config.svgElement) {
+        // We already have the SVG element for this icon, return a copy.
+        return Observable.of(config.svgElement.cloneNode(true));
+      } else {
+        // Fetch the icon from the config's URL, cache it, and return a copy.
+        return this._loadIconFromConfig(config)
+          .do((svg: SVGElement) => config.svgElement = svg)
```
jelbourn: +2 more for continuation indent

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+          .map((svg: SVGElement) => svg.cloneNode(true));
+      }
+    }
+    // See if we have any icon sets registered for the set name.
```
jelbourn: This function could benefit from being broken up into a few smaller ones.

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+      const namedIcon = this._extractIconWithNameFromAnySet(iconName, iconSetConfigs);
+      if (namedIcon) {
+        // We could cache namedSvg in _iconConfigs, but since we have to make a copy every
+        // time anyway, there's probably not much advantage compared to just always extracting
+        // it from the icon set.
+        return Observable.of(namedIcon);
+      }
+      // Not found in any cached icon sets. If there are icon sets with URLs that we haven't
+      // fetched, fetch them now and look for iconName in the results.
+      const iconSetFetchRequests = <[Observable<SVGElement>]>[];
```
jelbourn:
```ts
const iconSetFetchRequests: Observable<SVGElement>[] = [];
```
?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+      iconSetConfigs.forEach((setConfig) => {
```
jelbourn: I'd do `iconSetConfig`

nit: don't need parents around `setConfig`

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+      return Observable.forkJoin(iconSetFetchRequests)
+        .map((ignoredResults: any) => {
+          const foundIcon = this._extractIconWithNameFromAnySet(iconName, iconSetConfigs);
+          if (!foundIcon) {
+            throw Error(`Failed to find icon name: ${iconName} in namespace: ${namespace}`);
+          }
+          return foundIcon;
+        });
+    }
+    return Observable.throw(Error(`Unknown icon name: ${iconName} in namespace: ${namespace}`));
```
jelbourn: We've been using custom error classes to make testing easier (see the `portal/` code for an example)

src/components/icon/icon-registry.ts
jelbourn: If this class is meant to be internal (i.e., users of the component never interactive with it), add `@internal` to the description.

src/components/icon/icon-registry.ts
jelbourn: The type `SVGElement` shouldn't be necessary here since TS will infer it from the generic, no?

src/components/icon/icon-registry.ts
jelbourn: Actually maybe just `getSvg` ?
```ts
getSvg(name: string, namespace = '') {
```

src/components/icon/icon-registry.ts
robertmesserle: If it's SVG-only, then I think that would make the most sense.  If it supports other formats (ie. PNG), then the more generic option works.

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+    if (!iconNode) {
+      return null;
+    }
+    // createElement('SVG') doesn't work as expected; the DOM ends up with
+    // the correct nodes, but the SVG content doesn't render. Instead we
+    // have to create an empty SVG node using innerHTML and append its content.
+    // http://stackoverflow.com/questions/23003278/svg-innerhtml-in-firefox-can-not-display
+    const svg = this._svgElementFromString('<svg></svg>');
+    // Clone the node so we don't remove it from the parent icon set element.
+    svg.appendChild(iconNode.cloneNode(true));
```
jelbourn: Is this making any assumptions about the element type of `iconNode`?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,239 @@
...
+    this._setSvgAttributes(svg, config);
+    return svg;
+  }
+
+  private _svgElementFromString(str: string): SVGElement {
+    // TODO: Is there a better way than innerHTML? Renderer doesn't appear to have a method for
+    // creating an element from an HTML string.
+    const div = document.createElement('DIV');
+    div.innerHTML = str;
+    const svg = <SVGElement>div.querySelector('svg');
```
jelbourn: This prompted me to file an issue on the TypeScript repo, since it _could_ statically infer what the type is.
https://github.com/Microsoft/TypeScript/issues/8114

src/components/icon/icon.scss
```diff
@@ -0,0 +1,19 @@
...
+encapsulation, the selector would need to be ":host".
+*/
+md-icon {
+    background-repeat: no-repeat;
+    display: inline-block;
+    fill: currentColor;
+    height: $md-icon-size;
+    margin: auto;
+    vertical-align: middle;
+    width: $md-icon-size;
```
jelbourn: Too much indent 
(should only be +2)

src/components/icon/icon.ts
```diff
@@ -0,0 +1,178 @@
...
+
+@Component({
+  template: '<ng-content></ng-content>',
+  selector: 'md-icon',
+  styleUrls: ['./components/icon/icon.css'],
+  host: {
+    'role': 'img',
+  },
+  directives: [NgClass],
+  encapsulation: ViewEncapsulation.None,
```
jelbourn: You should be able to add 
```ts
```
without changing any behavior (for perf boost for change detection)

src/components/icon/icon.ts
```diff
@@ -0,0 +1,178 @@
...
+   *   'social:cake' -> ['social', 'cake']
+   *   'penguin' -> ['', 'penguin']
+   *   null -> ['', '']
+   */
+  private _splitIconName(iconName: string): [string, string] {
+    if (!iconName) {
+      return ['', ''];
+    }
+    const sepIndex = this.svgIcon.indexOf(':');
+    if (sepIndex == -1) {
```
jelbourn: I'd expand to `separatorIndex` or `delimiterIndex`. "sep" make me think 7.

src/demo-app/icon/assets/core-icon-set.svg
```diff
@@ -0,0 +1,33 @@
...
... 0 .72-.05 1.03-.15.32-.1.6-.25.83-.44s.42-.43.55-.72c.13-.29.2-.61.2-.97 0-.19-.02-.38-.07-.56-.05-.18-.12-.35-.23-.51-.1-.16-.24-.3-.4-.43-.17-.13-.37-.23-.61-.31.2-.09.37-.2.52-.33.15-.13.27-.27.37-.42.1-.15.17-.3.22-.46.05-.16.07-.32.07-.48 0-.36-.06-.68-.18-.96-.12-.28-.29-.51-.51-.69-.2-.19-.47-.33-.77-.43C9.1 8.05 8.76 8 8.39 8c-.36 0-.69.05-1 .16-.3.11-.57.26-.79.45-.21.19-.38.41-.51.67-.12.26-.18.54-.18.85h1.3c0-.17.03-.32.09-.45s.14-.25.25-.34c.11-.09.23-.17.38-.22.15-.05.3-.08.48-.08.4 0 .7.1.89.31.19.2.29.49.29.86 0 .18-.03.34-.08.49-.05.15-.14.27-.25.37-.11.1-.25.18-.41.24-.16.06-.36.09-.58.09H7.5v1.03h.77c.22 0 .42.02.6.07s.33.13.45.23c.12.11.22.24.29.4.07.16.1.35.1.57 0 .41-.12.72-.35.93-.23.23-.55.33-.95.33zm8.55-5.92c-.32-.33-.7-.59-1.14-.77-.43-.18-.92-.27-1.46-.27H12v8h2.3c.55 0 1.06-.09 1.51-.27.45-.18.84-.43 1.16-.76.32-.33.57-.73.74-1.19.17-.47.26-.99.26-1.57v-.4c0-.58-.09-1.1-.26-1.57-.18-.47-.43-.87-.75-1.2zm-.39 3.16c0 .42-.05.79-.14 1.13-.1.33-.24.62-.43.85-.19.23-.43.41-.71.53-.29.12-.62.18-.99.18h-.91V9.12h.97c.72 0 1.27.23 1.64.69.38.46.57 1.12.57 1.99v.4zM12 0l-.66.03 3.81 3.81 1.33-1.33c3.27 1.55 5.61 4.72 5.96 8.48h1.5C23.44 4.84 18.29 0 12 0z"/></g>
```
jelbourn: Should we change these to be `<svg>` inside of a `<defs>`?
(via https://github.com/angular/material/issues/5378)

jelbourn:
Ran through my first pass, haven't looked at the tests yet.

src/components/icon/icon-registry.ts
dozingcat: Done, it does make the code cleaner so I'm fine with it.

src/components/icon/icon-registry.ts
dozingcat: Added "svg" to the relevant method names.

src/components/icon/icon-registry.ts
dozingcat: Actually if classNameForFontAlias returns its argument if it hasn't been registered as an alias, it seems like the second argument here doesn't need to be optional.

src/components/icon/icon-registry.ts
dozingcat: How about getNamedSvgIcon, to distinguish from getSvgIconFromUrl?

src/components/icon/icon-registry.ts
dozingcat: Done-ish.

src/components/icon/icon-registry.ts
dozingcat: Done and made more functional-y.

src/components/icon/icon-registry.ts
dozingcat: Added exceptions, will add tests for them soon.

src/components/icon/icon-registry.ts
dozingcat: No, although I added a check so that <svg> elements are used directly rather than wrapped in another <svg>. Should we require it to be <svg> or <g>?

src/components/icon/icon-registry.ts
dozingcat: That would be cool.

dozingcat:
Going to add tests for:
- Throwing exceptions for nonexistent icon names and 
- <svg> children in icon sets
- Confirming that modifying an mdIcon's SVG element manually doesn't affect other instances of the same icon (every SVG element should be cloned from a pristine instance)

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+ * - Registers icon URLs by namespace and name.
+ * - Registers icon set URLs by namespace.
+ * - Registers aliases for CSS classes, for use with icon fonts.
+ * - Loads icons from URLs and extracts individual icons from icon sets.
+ */
+@Injectable()
+export class MdIconRegistry {
+  /**
+   * URLs and cached SVG elements for individual icons. Keys are of the format "[namespace]:[icon]".
+   */
```
jelbourn: optional: I'm cool with single-lining description blocks when they fit on one line.
```ts
/** Like this */
```

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+  private _cachedIconsByUrl = new Map<string, SVGElement>();
+
+  /**
+   * In-progress icon fetches. Used to coalesce multiple requests to the same URL.
+   */
+  private _inProgressUrlFetches = new Map<string, Observable<string>>();
+
+  /**
+   * Map from font identifiers to their CSS class names. Used for icon fonts.
```
jelbourn: Accurate to say `Used for icon fonts that are not ligature based.` ?

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+   */
+  addSvgIcon(iconName: string, url: string): this {
+    return this.addSvgIconInNamespace('', iconName, url);
+  }
+
+  /**
+   * Registers an icon by URL in the specified namespace.
+   */
+  addSvgIconInNamespace(namespace: string, iconName: string, url: string): this {
+    const iconKey = namespace + ':' + iconName;
```
jelbourn: We should have one place for this, something like `iconKey(namespace, name)`. It can be a function inside the module so that it doesn't have to be qualified.

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+   * Returns the cached icon for a SvgIconConfig if available, or fetches it from its URL if not.
+   */
+  private _getSvgFromConfig(config: SvgIconConfig): Observable<SVGElement> {
+    if (config.svgElement) {
+      // We already have the SVG element for this icon, return a copy.
+      return Observable.of(config.svgElement.cloneNode(true));
+    } else {
+      // Fetch the icon from the config's URL, cache it, and return a copy.
+      return this._loadSvgIconFromConfig(config)
+          .do((svg: SVGElement) => config.svgElement = svg)
```
jelbourn: Is the explicit `SVGElement` type necessary here? I would have thought TS inferred it from the generic.

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+        .map(svgText => this._createSvgElementForSingleIcon(svgText, config));
+  }
+
+  /**
+   * Loads the content of the icon set URL specified in the SvgIconConfig and creates an SVG element
+   * from it.
+   */
+  private _loadSvgIconSetFromConfig(config: SvgIconConfig): Observable<SVGElement> {
+    return this._fetchUrl(config.url)
+        .map((svgText) => this._svgElementFromString(svgText));
```
jelbourn: My XSS sense is tingling here. I'm trying to imagine a situation where an app would use `md-icon` to load an icon based on user input, but nothing is really coming to mind. Could you add a TODO for something like "Document that icons should only be loaded from trusted sources."

src/components/icon/icon-registry.ts
jelbourn: That's good- checking for a direct `svg` element was what I was thinking.

src/components/icon/icon.ts
```diff
@@ -0,0 +1,209 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MdIcon implements OnChanges, OnInit, AfterContentChecked {
+  @Input() svgSrc: string;
+  @Input() svgIcon: string;
+  @Input() fontSet: string;
+  @Input() fontIcon: string;
+  @Input() alt: string;
+
+  @Input('aria-label') ariaLabelFromParent: string = '';
```
jelbourn: `ariaLabelFromParent` > `hostAriaLabel`

src/components/icon/icon.ts
```diff
@@ -0,0 +1,209 @@
...
+   */
+  private _splitIconName(iconName: string): [string, string] {
+    if (!iconName) {
+      return ['', ''];
+    }
+    const separatorIndex = this.svgIcon.indexOf(':');
+    if (separatorIndex == -1) {
+      return ['', iconName];
+    }
+    return [iconName.substring(0, separatorIndex), iconName.substring(separatorIndex + 1)];
```
jelbourn: `slice` is generally preferred to `substring` since the later is easily confused with `substr`, which behaves completely differently.

src/components/icon/icon.ts
```diff
@@ -0,0 +1,209 @@
...
+    // Update aria label here because it may depend on the projected text content.
+    // (e.g. <md-icon>home</md-icon> should use 'home').
+    this._updateAriaLabel();
+  }
+
+  private _updateAriaLabel() {
+      const ariaLabel = this._getAriaLabel();
+      if (ariaLabel) {
+        this._renderer.setElementAttribute(this._element.nativeElement, 'aria-label', ariaLabel);
+        this._changeDetectorRef.detectChanges();
```
jelbourn: I'm reasonably sure that calling `detectChanges` here is going to be problematic, especially since it's being done from other lifecycle events. Why is the call necessary?

src/components/icon/icon.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+    // The "content" of an SVG icon is not a useful label.
+    if (this._usingFontIcon()) {
+      const text = this._element.nativeElement.textContent;
+      if (text) {
+        return text;
+      }
+    }
+    // Warn here?
+    return null;
```
jelbourn: We're holding off on logging aria warnings now because we want to be able to only output them in dev mode.

src/components/icon/icon.ts
```diff
@@ -0,0 +1,209 @@
...
+        this.ariaLabelFromParent ||
+        this.alt ||
+        this.fontIcon ||
+        this._splitIconName(this.svgIcon)[1];
+    if (label) {
+      return label;
+    }
+    // The "content" of an SVG icon is not a useful label.
+    if (this._usingFontIcon()) {
+      const text = this._element.nativeElement.textContent;
```
jelbourn: Can you pull textContent from the `Renderer`?

src/demo-app/icon/icon-demo.scss
```diff
@@ -0,0 +1,7 @@
+md-icon {
+  color: purple;
+}
+
+md-icon.green {
+  color: green;
+}
```
jelbourn: Can you prefix all of the demo styles with `.demo-`? This makes it easy for people to inspect the elements and immediately see which styles are from the component and which are from the demo.

jelbourn:
Got to everything but the unit tests, my gbus is arriving at its stop now.

src/components/icon/icon.spec.ts
```diff
@@ -0,0 +1,527 @@
...
+import {MdIcon} from './icon';
+import {MdIconRegistry} from './icon-registry';
+
+/** Returns the CSS classes assigned to an element as a sorted array. */
+const sortedClassNames = (elem: Element) => elem.className.split(' ').sort();
+
+/**
+ * Verifies that an element contains a single <svg> child element, and returns that child.
+ */
+const verifyAndGetSingleSvgChild = (element: any): any => {
```
jelbourn: Are the types here not `Element` and `SVGElement` respectively?

src/components/icon/icon.spec.ts
```diff
@@ -0,0 +1,527 @@
...
+  const svgChild = element.childNodes[0];
+  expect(svgChild.tagName.toLowerCase()).toBe('svg');
+  return svgChild;
+};
+
+/**
+ * Verifies that an element contains a single <path> child element whose "id" attribute has
+ * the specified value.
+ */
+const verifyPathChildElement = (element: any, attributeValue: string) => {
```
jelbourn: `Element` or `HTMLElement` ?

src/components/icon/icon.spec.ts
```diff
@@ -0,0 +1,527 @@
...
+        (tcb: TestComponentBuilder, mir: MdIconRegistry, mockBackend: MockBackend) => {
+      builder = tcb;
+      mdIconRegistry = mir;
+      // Set fake responses for various SVG URLs.
+      // Keep track of requests so we can verify caching behavior.
+      httpRequestUrls = [];
+      mockBackend.connections.subscribe((connection: any) => {
+        const url = connection.request.url;
+        httpRequestUrls.push(url);
+        switch (url) {
```
jelbourn: Can we break these into a separate file like `fake-svgs.ts` or `test-svgs.ts`, where each is associated with its name?  Possibly something like `getFakeSvgHttpResponses`.

jelbourn:
LGTM

@kara Can you also do a pass on this PR?

I think you have to rebase again because @kara alphabetized the demos

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
+import {Injectable} from 'angular2/core';
+import {BaseException} from 'angular2/src/facade/exceptions';
+import {Http} from 'angular2/http';
+import {Observable} from 'rxjs/Rx';
+
+
+/** Exception thrown when attmepting to load an icon with a name that cannot be found. */
```
kara: Typo:  attmepting --> attempting

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+export class MdIconNameNotFoundException extends BaseException {
+  constructor(iconName: string) {
+      super(`Unable to find icon with the name "${name}"`);
+  }
+}
+
+/**
+ * Exception thrown when attmepting to load SVG content that does not contain the expected
```
kara: Typo: attmepting --> attempting

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+   */
+  classNameForFontAlias(alias: string): string {
+    return this._fontCssClassesByAlias.get(alias) || alias;
+  }
+
+  /**
+   * Sets the CSS class name to be used for icon fonts when an <md-icon> component does not
+   * have a fontSet input value, and is not loading an icon by name or URL.
+   */
+  setDefaultFontSetClass(className: string) {
```
kara: Missing return value type (`this`)

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  /**
+   * Returns an Observable that produces the icon (as an <svg> DOM element) with the given name
+   * and namespace. The icon must have been previously registered with addIcon or addIconSet;
+   * if not, the Observable will throw an MdIconNameNotFoundException.
+   */
+  getNamedSvgIcon(name: string, namespace = ''): Observable<SVGElement> {
+    // Return (copy of) cached icon if possible.
+    const iconKey = namespace + ':' + name;
```
kara: Nitpick: Use template string here rather than concat?   i.e. `${namespace}:${name}`

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+    if (this._svgIconConfigs.has(iconKey)) {
+      return this._getSvgFromConfig(this._svgIconConfigs.get(iconKey));
+    }
+    // See if we have any icon sets registered for the namespace.
+    const iconSetConfigs = this._iconSetConfigs.get(namespace);
+    if (iconSetConfigs) {
+      return this._getSvgFromIconSetConfigs(name, this._iconSetConfigs.get(namespace));
```
kara: Any reason we're not using the `iconSetConfigs` const we just created? e.g.

```typescript
return this._getSvgFromIconSetConfigs(name, iconSetConfigs);
```

src/components/icon/icon-registry.ts
jelbourn: I actually think the concatenation is cleaner in this particular case.

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+                .catch((err: any, source: any, caught: any): Observable<SVGElement> => {
+                  // Swallow errors fetching individual URLs so the combined Observable won't
+                  // necessarily fail.
+                  console.log(`Loading icon set URL: ${iconSetConfig.url} failed: ${err}`);
+                  return Observable.of(null);
+                })
+                .do((svg: SVGElement) => {
+                  // Cache SVG element.
+                  if (svg) {
+                    iconSetConfig.svgElement = svg;
```
kara: Optional nitpick: collapse onto one line?

```if (svg) iconSetConfig.svgElement = svg;```

Comment (jelbourn) [inline src/components/icon/icon-registry.ts]: I actually prefer *always* having braces and newlines.

File: src/components/icon/icon-registry.ts [hunk 50]
```diff
@@ -0,0 +1,370 @@
...
+    return this._setSvgAttributes(svg, config);
+  }
+
+  /**
+   * Creates a DOM element from the given SVG string.
+   */
+  private _svgElementFromString(str: string): SVGElement {
+    // TODO: Is there a better way than innerHTML? Renderer doesn't appear to have a method for
+    // creating an element from an HTML string.
+    const div = document.createElement('DIV');
```
kara: You can use `DOM.createElement` here (same with `DOM.querySelector`)

src/components/icon/icon-registry.ts
kara: Really?  Don't like the waste of vertical space :-p  But completely subjective!

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+      throw new MdIconSvgTagNotFoundException();
+    }
+    return svg;
+  }
+
+  /**
+   * Sets the default attributes for an SVG element to be used as an icon.
+   */
+  private _setSvgAttributes(svg: SVGElement, config: SvgIconConfig): SVGElement {
+    if (!svg.getAttribute('xmlns')) {
```
kara: This can go through the renderer, e.g. `this._renderer.getElementAttribute` and  `this._renderer.setElementAttribute`

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+    // Clone the node so we don't remove it from the parent icon set element.
+    svg.appendChild(iconNode.cloneNode(true));
+    return this._setSvgAttributes(svg, config);
+  }
+
+  /**
+   * Creates a DOM element from the given SVG string.
+   */
+  private _svgElementFromString(str: string): SVGElement {
+    // TODO: Is there a better way than innerHTML? Renderer doesn't appear to have a method for
```
hansl:
```javascript
p.parseFromString('<a>hello<b>world</a>', 'text/html').body
```

src/components/icon/icon.ts
kara: I'm probably missing something, but any reason we're not using `iconName.split(':')`?  (and checking the arr's length to see if ":" was present)

src/components/icon/icon.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  private _usingFontIcon(): boolean {
+    return !(this.svgIcon || this.svgSrc);
+  }
+
+  private _setSvgElement(svg: SVGElement) {
+    // Can we use Renderer here somehow?
+    const layoutElement = this._element.nativeElement;
+    layoutElement.innerHTML = '';
+    layoutElement.appendChild(svg);
```
kara: I think you can use `renderer.appendNodes(layoutElement, [svg])`

src/components/icon/icon-registry.ts
jelbourn: Huh, I had never seen this API before.

src/components/icon/icon-registry.ts
jelbourn: The `DOM` facades are actually gone now (since we're not transpiling to dart anymore)

src/components/icon/icon-registry.ts
jelbourn: This is a service, for which there isn't a renderer instance. (each `Renderer` is tied to a specific component).

kara: Whoops, forgot where I was!

src/components/icon/icon-registry.ts
kara: Gotcha

src/components/icon/icon-registry.ts
dozingcat: You could have a custom icon font that uses ligatures and still set an alias for its CSS class and use it like ```<md-icon fontSet="myfont">home</md-icon>```, right?

jelbourn: That's right, never mind me.

src/components/icon/icon-registry.ts
dozingcat: Yeah, that felt sketchy to me also. Although I don't think it's any worse than ngMaterial 1.x.

src/components/icon/icon.ts
dozingcat: Ah, yeah split is better. I think I was trying to be clever about the case where there's more than one colon, but that should probably be an error anyway.

src/components/icon/icon.ts
dozingcat: I think it's a leftover from when I didn't have the right lifecycle hooks and was trying to force things. Removing it doesn't seem to hurt the tests or demo.

src/components/icon/icon.ts
dozingcat: added TODO

src/components/icon/icon.ts
dozingcat: I only see a setText method in https://github.com/angular/angular/blob/master/modules/angular2/src/core/render/api.ts

src/components/icon/icon-registry.ts
dozingcat: No reason at all, changed.

src/components/icon/icon-registry.ts
dozingcat: I reluctantly agree with @jelbourn. (On a related note, I'm very pleased with the 100-char line length).

src/components/icon/icon-registry.ts
```diff
@@ -0,0 +1,370 @@
...
+   */
+  private _fetchUrl(url: string): Observable<string> {
+    // Store in-progress fetches to avoid sending a duplicate request for a URL when there is
+    // already a request in progress for that URL. It's necessary to call share() on the
+    // Observable returned by http.get() so that multiple subscribers don't cause multiple XHRs.
+    if (this._inProgressUrlFetches.has(url)) {
+      return this._inProgressUrlFetches.get(url);
+    }
+    const req = this._http.get(url)
+        .map((response) => response.text())
```
hansl: You probably want a `flatMap` here.

Also, http can cache IIRC.  Let me check the API.

src/components/icon/icon-registry.ts
dozingcat: Neat, and it specifically supports SVG documents. Requires IE10 per https://developer.mozilla.org/en-US/docs/Web/API/DOMParser, is that ok?

jelbourn: Yeah, we only support IE11+

hansl: Yes. We support IE11 and up. You might want to create the DOMParser once.

src/components/icon/icon.ts
dozingcat: renderer.projectNodes? And renderer.detachView seems to remove the existing children.

dozingcat: ...except detachView calls element.classList, which isn't implemented on SVGElement in IE11. (It is in Edge). I'll file an issue against angular2.

dozingcat: Known bug: https://github.com/angular/angular/issues/6327

src/components/icon/icon-registry.ts
dozingcat: I've played with this for a while, but it appears that using DOMParser to create an <svg> element has the same problem as document.createElement('svg'), where the DOM content looks correct but the icon doesn't render. (similar to the referenced http://stackoverflow.com/questions/23003278/svg-innerhtml-in-firefox-can-not-display)

src/components/icon/icon.spec.ts
dozingcat: Done?

src/components/icon/icon.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+  directives: [MdIcon],
+})
+class MdIconLigatureTestApp {
+  ariaLabel: string = null;
+  iconName = '';
+}
+
+@Component({
+  selector: 'test-app',
+  template: `<md-icon aria-label="{{ariaLabel}}" alt="{{altText}}">{{iconName}}</md-icon>`,
```
jelbourn: One last comment:
for all these tests where you're using the `{{ }}` inside the attribute value, they should instead use the `[ ]` syntax, e.g.:
```
<md-icon [attr.aria-label]="ariaLabel" [alt]="altText">{{iconName}}</md-icon>
```

src/demo-app/icon/icon-demo.scss
```diff
@@ -0,0 +1,7 @@
+md-icon-demo md-icon {
```
jelbourn: Just `demo-` rather than `md-icon-demo`

jelbourn:
LGTM aside from those last two minor comments.

src/components/icon/icon.spec.ts
dozingcat: Done. Incidentally the documentation slightly leans toward interpolation: https://angular.io/docs/ts/latest/guide/template-syntax.html#!#property-binding

> There is no technical reason to prefer one form to the other. We lean toward readability, which tends to favor interpolation.

PR #284, opened 2016-04-08
fix(demo): add missing polyfill and up ng version

jelbourn:
R: @hansl 

Fixes the demo site in IE11


jelbourn:
CC @josephperrott in case you were curious

hansl:
> "It is a strange fate we should suffer so much fear and doubt… over so small a thing. Such a little thing."
> - Boromir

PR #290, opened 2016-04-09
feat(): add progress-fab component

DevVersion:
The feature is still under discussion. just decided to push it, so others can review and suggest improvements.

https://i.gyazo.com/4cb5fe4e70c4845e58d1c382d95f1eed.mp4

----
# MdProgressFab
`MdProgressFab` is a normal FAB button surrounded with a `progress-circle`.

### Screenshots
Preview

### Bound Properties

| Name | Type | Description |
| --- | --- | --- |
| `color` | `"primary" | "accent" | "warn"` | The color palette for the FAB button |
| `value` | `number` | Value for the `progress-circle`.<br/> Necessary when using the `determinate` mode. |
| `mode` | `"determinate" | "indeterminate"` | Mode for the `progress-circle`.<br/> |
| `progressColor` | `"primary" | "accent" | "warn"` | Color for the `progress-circle`.<br/> |

### Examples
A basic progress-fab will have the markup:
```html
<button md-progress-fab color="accent">
  <i class="material-icons md-24">favorite</i>
  </button>
```
It will use by default a `indeterminate` progress circle.

### DIFF
--- a/src/components/progress-fab/progress-fab.html
+++ b/src/components/progress-fab/progress-fab.html
@@ -1,0 +1,13 @@
+<span class="md-button-wrapper">
+    <ng-content></ng-content>
+
+    <div class="md-progress-wrap">
+        <md-progress-circle
+          [mode]="progressMode"
+          [value]="progressValue"
+          [attr.color]="progressColor">
+        </md-progress-circle>
+    </div>
+</span>
+
+
--- a/src/components/progress-fab/progress-fab.spec.ts
+++ b/src/components/progress-fab/progress-fab.spec.ts
@@ -1,0 +1,123 @@
+import {
+  it,
+  describe,
+  expect,
+  beforeEach,
+  inject,
+} from '@angular/core/testing';
+import {TestComponentBuilder} from '@angular/compiler/testing';
+import {Component} from '@angular/core';
+import {By} from '@angular/platform-browser';
+import {MdProgressFab} from './progress-fab';
+import {MdProgressCircle} from '../progress-circle/progress-circle';
+
+export function main() {
+  describe('MdProgressFab', () => {
+    let builder: TestComponentBuilder;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+    }));
+
+    it('should correctly apply the color attribute on the progress circle', (done: () => void) => {
+      return builder
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let testComponent = fixture.debugElement.componentInstance;
+          let progressDebugElement = fixture.debugElement.query(By.css('md-progress-circle'));
+
+          testComponent.progressColor = 'primary';
+          fixture.detectChanges();
+
+          expect(progressDebugElement.nativeElement.getAttribute('color')).toBe('primary');
+
+          testComponent.progressColor = 'accent';
+          fixture.detectChanges();
+
+          expect(progressDebugElement.nativeElement.getAttribute('color')).toBe('accent');
+
+          done();
+      });
+    });
+
+    it('should correctly apply the mode on the progress circle', (done: () => void) => {
+      return builder
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let testComponent = fixture.debugElement.componentInstance;
+          let progressComponent: MdProgressCircle = fixture.debugElement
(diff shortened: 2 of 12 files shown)


DevVersion:
Notes: This needs to be rebased with the latest changes of d24a5b3e3bf9bf4bb63f0f8e22b0047318dc6ea0 , which
- cause alignment issues with the relative positioning

jelbourn:
@DevVersion Now needs rebasing on my mega change #384; feel free to ping me if you run into any issues

DevVersion:
@jelbourn Done.

src/components/progress-fab/progress-fab.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export function main() {
+  describe('MdProgressFab', () => {
+    let builder: TestComponentBuilder;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+    }));
+
+    it('should correctly apply the color attribute on the progress circle', (done: () => void) => {
```
jelbourn: You can use the new `async` function instead of the `done` callback now

src/components/progress-fab/progress-fab.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+      return builder
+        .createAsync(TestApp)
+        .then((fixture) => {
```
jelbourn: not: `fixture` doesn't need parenthesis around it.

src/demo-app/progress-fab/progress-fab-demo.scss
```diff
@@ -0,0 +1,26 @@
...
+            transition-timing-function: ease-out;
+
+            transform: scale(0.8);
+            opacity: 0;
+          }
+        }
+      }
+
+    }
+  }
```
jelbourn: Could you flatten these style definitions (not nested),

jelbourn:
@DevVersion sorry for the delay on this PR, I need to set some time aside to see what the a11y story for this component should be.

DevVersion:
@jelbourn It's no problem :)
There currently other things with higher priorities. 

Just ping me on the PR, once you like to revisit it.

src/demo-app/progress-fab/progress-fab-demo.scss
nesukun: wouldn't that hurt, albeit just a bit, maintainability? In the same way single-line 'if' blocks not surrounded by {} are usually considered kinda rude to the next developer who has to add further statements and is forced to add the braces and reindent

src/components/progress-fab/progress-fab.html
```diff
@@ -0,0 +1,13 @@
+<span class="md-button-wrapper">
+    <ng-content></ng-content>
+
+    <div class="md-progress-wrap">
+        <md-progress-circle
```
marcysutton: To be announced, this progress circle needs to be a sibling to the focusable button, and it needs `role="progressbar"` along with other related ARIA attributes: https://www.w3.org/TR/wai-aria/roles#progressbar

Specifically the progressbar needs to specify `aria-valuenow`, `aria-valuemin` and `aria-valuemax` for screen readers to follow along. 

Lastly, what is the progress circle indicating (i.e. what is loading)? That binding should be indicated somehow...the ARIA roles spec recommends using `aria-describedby` on the progressbar and `aria-busy` on whatever is loading.

marcysutton:
To answer @jelbourn's question from #305, this definitely does need some a11y treatment. While the button is the focusable element, the `progressbar` component (which needs ARIA) should be a sibling so it doesn't live inside of the button. That way the progressbar can be consumed effectively by assistive technologies, instead of trying to announce it in the context of a button (which has a different purpose and function).

If the progressbar component can't bind itself to whatever is actually loading on the page with `aria-describedby`, you'll have to make that step clear in the component documentation.

src/components/progress-fab/progress-fab.html
DevVersion: Thanks @marcysutton, I'll try making the process circle a sibling. 

- The role is automatically applied from the Progress Circle Component
- Valuenow, ValueMin, ValueNow are also covered by the circle component. 

Forwarding the `aria-label` / or creating a default one, makes definitely sense. thx

DevVersion:
Going to close this for now, since this is a _nice to have_ component

PR #302, opened 2016-04-13
chore(deps): update angular2 and zones versions

kara:
r: @jelbourn 

- Angular update is necessary to enable my gestures PR
- Zones update is required by newest version of Angular
- Broccoli-merge-trees should be in package.json (or fresh npm installs will break).


jelbourn:
LGTM

PR #305, opened 2016-04-14
feat(): add interaction service for a11y

### LINKED ISSUE #289: Service for handling keyboard vs. mouse focus behavior
Currently, we always show a focus effect on the button component, when there was no mouse event the last `100ms`.

This is really hard-coded and there are bugs. For example, switching browser tabs, is restoring focus and showing the focus-effect (should not happen).

A service for interaction, which detects the last interaction type would solve that issue very elegant and would be also helpful for other components.

I also implemented that in Material 1 (still pending after re-submit) and I would like to implement it in Material 2 as well.

So let me know if you need this, and I will implement it :)

DevVersion:
Closes #289.

### DIFF
--- a/src/core/interaction/interaction.spec.ts
+++ b/src/core/interaction/interaction.spec.ts
@@ -1,0 +1,51 @@
+import {
+  it,
+  describe,
+  expect,
+  beforeEach,
+  inject,
+  beforeEachProviders,
+} from 'angular2/testing';
+import {MdInteraction, MdInteractionType} from './interaction';
+
+export function main() {
+  describe('MdInteraction', () => {
+    let interaction: MdInteraction;
+
+    beforeEachProviders(() => [MdInteraction]);
+
+    beforeEach(inject([MdInteraction], (_interaction: MdInteraction) => {
+      interaction = _interaction;
+    }));
+
+    it('should correctly detect the keyboard interaction', () => {
+
+      let event = <KeyboardEvent> document.createEvent('Event');
+
+      event.keyCode = 37;
+      event.initEvent('keydown', false, true);
+
+      document.body.dispatchEvent(event);
+
+      expect(interaction.getLastInteractionType()).toBe(MdInteractionType.KEYBOARD);
+    });
+
+    it('should correctly detect the mouse interaction', () => {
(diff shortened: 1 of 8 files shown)


kara:
We're working on a fastclick solution in core that should make some of this easier (handling touch delay, possibly new APIs for global event listeners, etc).   It would probably make sense to wait on this functionality a bit and re-visit after.

cc @jelbourn

DevVersion:
@kara I'm fine with that, this PR can wait for the fastclick solution, but it should be also able to land, because once the Fastclick Solution is landed, we just need to remove the buffer, which is really small effort.

jelbourn:
@DevVersion I chatted with @kara about this and we agreed that it would be better to wait for the fastclick to land because it will make this service simpler.

DevVersion:
@jelbourn Okay cool, just come back if fastclick is available and I will rebase on top of that ;)

jelbourn:
@marcysutton do you have any thoughts about the combination "progressbar" / "button" from any a11y standpoint? 

I would guess that you'd want the two to be siblings in the DOM rather than one inside of the other. I'd also think that you'd want the "progressbar" to be aria-hidden until the button interaction, meaning you might want it to seem like the progress bar is "inserted" after the button is clicked. Not sure if it would make any sense whatsoever to manipulate focus during this interaction.

jelbourn:
Just realized my previous comment was on the wrong issue. It should have been on #290

jelbourn:
@DevVersion, we're going to try to approach this instead by adding a feature to Zones that can tell you the type of the most recent event was, and then only show if it was a `KeyboardEvent`. Going to close this.

marcysutton:
I'll be very curious to see what that feature looks like. Differentiating between keydown and click can get tricky depending on the element's role.

DevVersion:
@jelbourn Sounds interesting. Also wouldn't that mean that the events have to be triggered inside of a zone?

jelbourn:
@DevVersion not sure, depends on how @robertmesserle implements it. 

@marcysutton I can CC you on the PR once it exists (and if we validate that the idea works in general).

PR #307, opened 2016-04-14
fix(button): remove ref to Event global (for node)

### LINKED ISSUE #306: `Event` type in button causes error during nodejs precompilation

Bug

If I use prerender libraries for angular universal to precompile a component that uses `MdButton`, it fails as soon as the compiled button.js module is loaded because of the presence of a global `Event` object that doesn't exist in node.

This line is the offender: https://github.com/angular/material2/blob/d4a3cde52ba736da64285bdfaf039129d0dd8312/src/components/button/button.ts#L122

Because of the `HostListener` decorator, the type information for that method gets preserved as follows:

```js
__decorate([
        core_1.HostListener('click', ['$event']),
        __metadata('design:type', Function),
        __metadata('design:paramtypes', [Event]), // <-- there's the problem
        __metadata('design:returntype', void 0)
    ], MdAnchor.prototype, "haltDisabledEvents", null);
```

Beta.1 and Beta.2.

* **Other information**

```
ReferenceError: Could not require 'angular-cli-build.js': Event is not defined
    at project/node_modules/@angular2-material/button/button.ts:121:37
    at Object.<anonymous> (project/node_modules/@angular2-material/button/button.ts:94:1)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (project/src/client/app/issue-cli.ts:7:1)
    at Module._compile (module.js:413:34)
```

jelbourn:
R: @kara 

Because the method `haltDisabledEvents` was decorated, the type was preserved at run-time. This caused the app to fail in non-browser environments due to the `Event` global not being present. 

This unblocks @jeffbcross for now, but we need to add a CI check to prevent this from happening in the future.

Closes #306

### DIFF
--- a/src/components/button/button.ts
+++ b/src/components/button/button.ts
@@ -3,7 +3,6 @@ import {
   ViewEncapsulation,
   Input,
   HostBinding,
-  HostListener,
   ChangeDetectionStrategy,
   ElementRef,
   Renderer,
@@ -22,7 +21,7 @@ import {
     '[class.md-button-focus]': 'isKeyboardFocused',
     '(mousedown)': 'setMousedown()',
     '(focus)': 'setKeyboardFocus()',
-    '(blur)': 'removeKeyboardFocus()'
+    '(blur)': 'removeKeyboardFocus()',
   },
   templateUrl: './components/button/button.html',
   styleUrls: ['./components/button/button.css'],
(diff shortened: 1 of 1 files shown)


src/components/button/button.ts
```diff
@@ -22,7 +22,7 @@ import {
     '[class.md-button-focus]': 'isKeyboardFocused',
     '(mousedown)': 'setMousedown()',
     '(focus)': 'setKeyboardFocus()',
-    '(blur)': 'removeKeyboardFocus()'
+    '(blur)': 'removeKeyboardFocus()',
```
kara: Extra comma?

jelbourn: Trailing commas4life

jeffbcross:
I love the quick fix turnaround!

PR #321, opened 2016-04-18
fix(input): prefix copied input id

matthewdenobrega:


src/components/input/input.ts
```diff
@@ -134,7 +136,7 @@ export class MdInput implements ControlValueAccessor, AfterContentInit, OnChange
   @Input() @BooleanFieldValue() disabled: boolean = false;
   @Input() @BooleanFieldValue() floatingPlaceholder: boolean = true;
   @Input() hintLabel: string = '';
-  @Input() id: string = `md-input-${nextUniqueId++}`;
+  @Input() id: string = '';
```
jelbourn: I think the change here should actually be:
1. Rename `id` to `inputId`
2. Remove the `@Input()` decorator, since the component itself doesn't ever _use_ the `id` from the host element.

jelbourn:
cc @hansl

matthewdenobrega:
I'm happy to make this change if I get a call either way - the design doc should be updated as well and I don't have access to those.

jelbourn:
Closed in favor of #450

PR #341, opened 2016-04-22
chore(demos): fix sidenav styles

kara:
r: @jelbourn 

Small PR to fix the sidenav styles in the demo.

### DIFF
--- a/src/demo-app/demo-app.scss
+++ b/src/demo-app/demo-app.scss
@@ -1,13 +1,31 @@
 .demo-root {
   font-family: Roboto, 'Helvetica Neue', sans-serif;
-  padding: 20px;
-  min-height: 1000px;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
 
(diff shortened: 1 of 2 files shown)


src/demo-app/demo-app.scss
```diff
@@ -36,3 +35,18 @@
 .demo-content {
   padding: 15px;
 }
+
+md-sidenav-layout {
```
jelbourn: Prefix with `demo-`

kara: Always forget!  Will do

src/demo-app/demo-app.scss
```diff
@@ -31,8 +29,25 @@
     font-size: 20px;
   }
 
-}
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
jelbourn: All the styles that apply directly `demo-root` should go right under the definition, before the nested selectors.

kara: Done.

jelbourn:
LGTM

PR #343, opened 2016-04-24
feat(input): add support for more input attributes

matthewdenobrega:
Notes
 - I've included all the attributes from [the MDN docs on input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) that I thought would potentially be useful and are reasonably widely supported
 - I haven't implemented any checking of attributes against the type of the input - for example checking that max is a date when the input type is date - my feeling is this should be up to the developer
 - I've followed the pattern of minLength (which was already implemented) and used camelCase for attributes like readOnly
 - I've been quite explicit with the tests - I can remove some if this is overkill
 - The handling of the spellCheck attribute is different from the other boolean attributes - this is because it is rendered differently (as ="true" or ="false") by Angular2


matthewdenobrega:
@jelbourn @hansl can someone review this please?

hansl:
@jelbourn: I'm fine with this. It also becomes a good benchmark of how much bindings are a bottleneck. I needed to do this anyway for data-table.

jelbourn:
@matthewdenobrega can you rebase this PR on master?

matthewdenobrega:
I've rebased on master and updated the specs so they are more inline with the other specs for input. I'm not sure what the CLA warning is - I've signed the CLA and no-one else has contributed commits.

jelbourn:
@matthewdenobrega looks like your rebase didn't go very smoothly- 124 changed files.

matthewdenobrega:
@jelbourn I don't rebase often so maybe I'm missing something, but the only differences between master and my branch are my changes:(https://github.com/angular/material2/compare/master...matthewdenobrega:input-attributes)

If something went wrong with the rebase I can branch off master again and redo the changes.

jelbourn:
@matthewdenobrega Not sure where it went awry, but the PR has 25 commits in it now (most of which are already in master). It might be easier to create a clean branch from `upstream/master` and then `git cherry-pick` your individual commits into it, then force-push back to this PR branch.

matthewdenobrega:
New PR opened: #447

jelbourn:
Closed in favor of #447

PR #345, opened 2016-04-24
Use <md-icon> instead of <i class="material-icons"> in demos.

dozingcat:
Also removed vertical-align in mdIcon styles because it causes the button icons to be misaligned.

### DIFF
--- a/src/components/button/_button-base.scss
+++ b/src/components/button/_button-base.scss
@@ -107,7 +107,7 @@ $md-mini-fab-padding: 8px !default;
   height: $size;
   padding: 0;
 
-  md-icon {
+  i, md-icon {
     padding: $padding 0;
   }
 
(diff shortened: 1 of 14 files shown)


jelbourn:
LGTM, just needs rebase

dozingcat:
Done. @kara might want to double check the merges.

src/components/button/_button-base.scss
```diff
@@ -107,7 +107,7 @@ $md-mini-fab-padding: 8px !default;
   height: $size;
   padding: 0;
 
-  i {
+  md-icon {
```
kara: This should actually be `i`, not `md-icon`.  `md-icon` is already aligned correctly; it's just regular `i` that need the padding.

dozingcat: Done

kara:
LGTM

PR #349, opened 2016-04-25
feat(list): add icon support

kara:
r: @jelbourn

Avatars and icons need slightly different styling (icons need to be smaller). This PR adds support for default styling for icons.

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -3,6 +3,7 @@
 
 $md-list-side-padding: 16px;
 $md-list-avatar-size: 40px;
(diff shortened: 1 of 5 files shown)


src/components/list/list.scss
```diff
@@ -83,6 +84,13 @@ based on whether the list is in dense mode.
     height: $md-list-avatar-size;
     border-radius: 50%;
   }
+
+  [md-list-icon] {
```
kara: Could also be `md-list-icon-avatar`, but felt too long maybe?

jelbourn:
LGTM

PR #352, opened 2016-04-25
feat(sidenav): add fullscreen mode

kara:
r: @jelbourn 

This PR introduces a `fullscreen` mode for sidenav, so sidenav spans the whole viewport.


jelbourn:
LGTM; @hansl should take a look

PR #354, opened 2016-04-25
fix(button): cleaned up button theming

kara:
r: @jelbourn


src/components/button/_button-base.scss
```diff
@@ -102,9 +102,14 @@ $md-mini-fab-padding: 8px !default;
   min-width: 0;
   border-radius: $md-fab-border-radius;
   background-color: md-color($md-accent);
+  color: md-color($md-accent, default-contrast);
   width: $size;
   height: $size;
-  padding: $padding;
+  padding: 0;
+
+  i {
```
jelbourn: I don't think this should just be `i`, which isn't necessarily an icon.

jelbourn:
LGTM

PR #360, opened 2016-04-26
fix(checkbox): fix horizontal spacing between checkboxes and radio buttons

kara:
r: @jelbourn
This PR fixes the baselines for checkboxes and radio buttons stacked vertically. A more robust and permanent solution is needed to fix baselines for checkboxes and radio buttons on the same line.

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -3,7 +3,7 @@
 @import "variables";
 
 /** The width/height of the checkbox element. */
-$md-checkbox-size: 18px !default;
+$md-checkbox-size: $md-toggle-size !default;
 /** The width of the line used to draw the checkmark / mixedmark. */
 $md-checkbox-mark-stroke-size: 2/15 * $md-checkbox-size !default;
 /** The width of the checkbox border shown when the checkbox is unchecked. */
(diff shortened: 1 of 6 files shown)


src/components/checkbox/checkbox.scss
```diff
@@ -3,7 +3,7 @@
 @import "variables";
 
 /** The width/height of the checkbox element. */
-$md-checkbox-size: 18px !default;
+$md-checkbox-size: 20px !default;
```
jelbourn: Can we refactor this out into something like `$md-toggle-size` and use it for both checkbox and radio button?

kara: Yes! Good call.

src/components/checkbox/checkbox.scss
```diff
@@ -250,6 +250,10 @@ $_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1)
   }
 }
 
+.md-checkbox-layout label {
+  line-height: 24px;
```
jelbourn: I'm not sure if we should be setting line-height. I would think that the checkbox should just use whatever line height is already applied. 

@hansl, I think you will have thoughts on this

kara: The label looks a bit high without it.  I noticed radio buttons were using line-height, so tried to make it consistent.

src/core/style/_variables.scss
```diff
@@ -15,6 +15,11 @@ $z-index-drawer: 100 !default;
 // Global constants
 $pi: 3.14159265;
 
+// Padding between input toggles and their labels
+$md-toggle-padding: 8px;
+// Width and height of input toggles
+$md-toggle-size: 20px;
```
jelbourn: `!default` ?

jelbourn:
LGTM aside from the `!default` comment

src/core/style/_variables.scss
kara: Fixed!

PR #361, opened 2016-04-26
fix(sidenav): add min width for empty navs

kara:
r: @hansl @jelbourn

### DIFF
--- a/src/components/sidenav/sidenav.scss
+++ b/src/components/sidenav/sidenav.scss
@@ -107,6 +107,7 @@ $md-sidenav-push-background-color: md-color($md-background, dialog) !default;
     top: 0;
     bottom: 0;
     z-index: 3;
+    min-width: 5%;
 
     background-color: $md-sidenav-background-color;
 


jelbourn:
LGTM

src/components/sidenav/sidenav.scss
```diff
@@ -107,6 +107,7 @@ $md-sidenav-push-background-color: md-color($md-background, dialog) !default;
     top: 0;
     bottom: 0;
     z-index: 3;
+    min-width: 5%;
```
DevVersion: Hm, just questioning -  I don't think that using a percentage is good here. `5%` sounds like a random percentage. 
- Percentage in relative to the parent, do you think that looks / works good on small screens?

hansl: That's just in case there's no content in the sidenav. Likely

hansl:
LGTM.

sunderw:
Hi, this CSS rule is causing an overflow on the X axis on my page. I don't really understand which issue it is supposed to fix. Could someone enlighten me ?

Issue #90, opened 2016-02-21
feat(): switch component


DevVersion:
Hello,
Just wanted to know, if someone is already working on the switch component?
As I recently trying things with Angular 2 I built a few components and would like to help if desired at the material2 component. I know it's still in early stages, but I think it never hurts to ask.

![Preview](https://i.gyazo.com/8b1b212257b831537fe3ba413d605253.png)

jelbourn:
We definitely plan on building one, but we're waiting for completion of both the checkbox (#35) since it will share most of the code, and gestures (#72) for having a common way of doing the sliding. Once those pieces are done I'll come back and update this issue.

DevVersion:
@jelbourn Is this now unblocked? Since HammerJS is now available.

jelbourn:
@DevVersion I suppose it is. Is this something you want to contribute? 

Couple of notes:
* A lot of the TS code should be shared between this and checkbox.
* Users should be able to drag the switch part.

jelbourn:
cc @kara

DevVersion:
@jelbourn Yes definitely.

I already implemented it myself a few months ago (#7)
- But you may have noticed, it used much things from Material 1 and even used an internal gesture service. 

I will start working on this new component the next days (Material 1 - 1.1.0 coming).

I try to extend as much as possible and also work on the drag for the switch part.

jelbourn:
@DevVersion it would be good to break the work into stages. Something like:

1. Add empty skeleton for slide-toggle
2. Refactor any checkbox code that would be used by both
3. Add the dragging to the slide-toggle

(I forgot to mention earlier, we want to change the name to `md-slide-toggle` since "switch" is already overloaded with the `switch` control construct, e.g., `ng-switch`).

DevVersion:
@jelbourn Yea this sounds good. Am I allowed to ask some questions on Slack, if needed?

jelbourn:
@DevVersion of course!

DevVersion:
Fixed with #362.

PR #362, opened 2016-04-26
feat(): add slide-toggle component.

DevVersion:
First step; Basic Skeleton for the Slide Toggle Component.

- [x] Readme
- [x] Visually Hidden Checkbox
- [x] Doesn't expose the EventEmitter
- [x] Replace Hue 400, 500 in background palette (@jelbourn) 

Adressed in other PR's
-  Drag (Other PR - as discussed in #90)
- Refactoring / Depending on Checkbox

### DIFF
--- a/src/components/slide-toggle/README.md
+++ b/src/components/slide-toggle/README.md
@@ -1,0 +1,66 @@
+# MdSlideToggle
+`MdSlideToggle` is a two-state control, which can be also called `switch`
+
+### Screenshots
+![image](https://material.angularjs.org/material2_assets/slide-toggle/toggles.png)
+
+## `<md-slide-toggle>`
+### Bound Properties
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `disabled` | boolean | Disables the `slide-toggle` |
+| `color` | `"primary" | "accent" | "warn"` | The color palette of the `slide-toggle` |
+| `checked` | boolean | Sets the value of the `slide-toggle` |
+
+### Events
+| Name | Type | Description |
+| --- | --- | --- |
+| `change` | boolean | Event will be emitted on every value change.<br/>It emits the new `checked` value. |
+
+### Examples
+A basic slide-toggle would have the following markup.
+```html
+<md-slide-toggle [(ngModel)]="slideToggleModel">
+  Default Slide Toggle
+</md-slide-toggle>
+```
+
+Slide toggle can be also disabled.
+```html
+<md-slide-toggle disabled>
+  Disabled Slide Toggle
+</md-slide-toggle>
+```
+
+The `slide-toggle` can be also set to checked without a `ngModel`
+```html
+<md-slide-toggle [checked]="isChecked">
+  Input Binding
+</md-slide-toggle>
+```
+
+You may also want to listen on changes of the `slide-toggle`<br/>
+The `slide-toggle` always emits the new value to the event binding `(change)`
+```html
+<md-slide-toggle (change)="printValue($event)">
+  Prints Value on Change
+</md-slide-toggle>
+```
+
+## Theming
+A slide-toggle is default using the `accent` palette for its styling.
+
+Modifying the color on a `slide-toggle` can be easily done, by using the following markup.
+```html
+<md-slide-toggle color="primary">
+  Primary Slide Toggle
+</md-slide-toggle>
+```
+
+The color can be also set dynamically by using a property binding.
+```html
+<md-slide-toggle [color]="myColor">
+  Dynamic Color
+</md-slide-toggle>
+```
--- a/src/components/slide-toggle/slide-toggle.html
+++ b/src/components/slide-toggle/slide-toggle.html
@@ -1,0 +1,22 @@
+<label class="md-slide-toggle-label">
+  <div class="md-slide-toggle-container">
+    <div class="md-slide-toggle-bar"></div>
+    <div class="md-slide-toggle-thumb-container">
+      <div class="md-slide-toggle-thumb"></div>
+    </div>
+
+    <input #input class="md-slide-toggle-checkbox" type="checkbox"
+           [id]="getInputId()"
+           [tabIndex]="tabIndex"
+           [checked]="checked"
+           [disabled]="disabled"
+           [attr.name]="name"
+           [attr.aria-label]="ariaLabel"
+           [attr.aria-labelledby]="ariaLabelledby"
+           (blur)="onTouched()"
+           (change)="onChangeEvent()">
+  </div>
+  <span class="md-slide-toggle-content">
+    <ng-content></ng-content>
+  </span>
+</label>
--- a/src/components/slide-toggle/slide-toggle.scss
+++ b/src/components/slide-toggle/slide-toggle.scss
@@ -1,0 +1,153 @@
+@import "../../core/style/variables";
+@import "../../core/style/mixins";
+@import "../../core/style/elevation";
+
+//TODO(): remove the default theme.
+@import "../../core/style/default-theme";
+
+$md-slide-toggle-width: 36px !default;
+$md-slide-toggle-height: 24px !default;
+$md-slide-toggle-bar-height: 14px !default;
+$md-slide-toggle-thumb-size: 20px !default;
+$md-slide-toggle-margin: 16px !default;
+
+@mixin md-switch-checked($palette) {
+  .md-slide-toggle-thumb {
+    background-color: md-color($palette);
+  }
+
+  .md-slide-toggle-bar {
+    background-color: md-color($palette, 0.5);
+  }
+}
+
+:host {
+  display: flex;
+  height: $md-slide-toggle-height;
+
+  margin: $md-slide-toggle-margin 0;
+  line-height: $md-slide-toggle-height;
+
+  white-space: nowrap;
+  user-select: none;
+
+  outline: none;
+
+  &.md-checked {
+    @include md-switch-checked($md-accent);
+
+    &.md-primary {
+      @include md-switch-checked($md-primary);
+    }
+
+    &.md-warn {
+      @include md-switch-checked($md-warn);
+    }
+
+    .md-slide-toggle-thumb-container {
+      transform: translate3d(100%, 0, 0);
+    }
+  }
+
+  &.md-disabled {
+
+    .md-slide-toggle-label, .md-slide-toggle-container {
+      cursor: default;
+    }
+
+    .md-slide-toggle-thumb {
+      // The thumb of the slide-toggle always uses the hue 400 of the grey palette in dark or light themes.
+      // Since this is very specific to the slide-toggle component, we're not providing
+      // it in the background palette.
+      background-color: md-color($md-grey, 400);
+    }
+    .md-slide-toggle-bar {
+      background-color: md-color($md-foreground, divider);
+    }
+  }
+}
+
+// The label is our root container for the slide-toggle / switch indicator and label text.
+// It has to be a label, to support accessibility for the visual hidden input.
+.md-slide-toggle-label {
+  display: flex;
+  flex: 1;
+
+  cursor: pointer;
+}
+
+// Container for the composition of the slide-toggle / switch indicator.
+.md-slide-toggle-container {
+  cursor: grab;
+  width: $md-slide-toggle-width;
+  height: $md-slide-toggle-height;
+
+  position: relative;
+  user-select: none;
+
+  margin-right: 8px;
+}
+
+// The thumb container is responsible for the dragging functionality.
+// It moves around and holds the actual circle as a thumb.
+.md-slide-toggle-thumb-container {
+  position: absolute;
+  top: $md-slide-toggle-height / 2 - $md-slide-toggle-thumb-size / 2;
+  left: 0;
+  z-index: 1;
+
+  width: $md-slide-toggle-width - $md-slide-toggle-thumb-size;
+
+  transform: translate3d(0, 0, 0);
+
+  transition: $swift-linear;
+  transition-property: transform;
+}
+
+// The thumb will be elevated from the slide-toggle bar.
+// Also the thumb is bound to its parent thumb-container, which manages the movement of the thumb.
+.md-slide-toggle-thumb {
+  position: absolute;
+  margin: 0;
+  left: 0;
+  top: 0;
+
+  height: $md-slide-toggle-thumb-size;
+  width: $md-slide-toggle-thumb-size;
+  border-radius: 50%;
+
+  background-color: md-color($md-background, background);
+  @include md-elevation(1);
+}
+
+// Horizontal bar for the slide-toggle.
+// The slide-toggle bar is shown behind the thumb container.
+.md-slide-toggle-bar {
+  position: absolute;
+  left: 1px;
+  top: $md-slide-toggle-height / 2 - $md-slide-toggle-bar-height / 2;
+
+  width: $md-slide-toggle-width - 2px;
+  height: $md-slide-toggle-bar-height;
+
+  // The bar of the slide-toggle always uses the hue 500 of the grey palette in dark or light themes.
+  // Since this is very specific to the slide-toggle component, we're not providing
+  // it in the background palette.
+  background-color: md-color($md-grey, 500);
+
+  border-radius: 8px;
+}
+
+// The slide toggle shows a visually hidden checkbox inside of the component.
+// This checkbox allows us to take advantage of the browsers support.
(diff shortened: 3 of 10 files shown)


jelbourn:
@DevVersion Now needs rebasing on my mega change #384. Feel free to ping me if you run into any issues.

DevVersion:
@jelbourn Done :)

src/components/slide-toggle/slide-toggle.html
```diff
@@ -0,0 +1,9 @@
+<div class="md-container">
+  <div class="md-bar"></div>
+  <div class="md-thumb-container">
+    <div class="md-thumb"></div>
```
jelbourn: For the slide-toggle, we want to drive the component with an underlying native checkbox, similar to what @robertmesserle is doing in #415 (since the slide-toggle is pretty much just a checkbox with a different appearance). 

This gives you the accessibility and keyboard interaction for free.

src/components/slide-toggle/slide-toggle.scss
```diff
@@ -0,0 +1,132 @@
...
+
+    .md-bar {
+      background-color: md-color($palette, 0.5);
+    }
+  }
+
+  &.md-checked {
+    @include md-switch-checked($md-accent);
+
+    &.md-primary {
```
jelbourn: We should stick with the same `color="..."` API as the other components.

src/components/slide-toggle/slide-toggle.html
```diff
@@ -0,0 +1,9 @@
+<div class="md-container">
```
jelbourn: Can you rename the CSS classes so that they're more explicitly for the slide-toggle? E.g., 
`md-slidetoggle-container`

src/components/slide-toggle/slide-toggle.scss
```diff
@@ -0,0 +1,132 @@
...
+
+  &[disabled] {
+    cursor: default;
+
+    .md-container {
+      cursor: default;
+    }
+  }
+
+  .md-container {
```
jelbourn: Could you add a brief description of what each class represents?

jelbourn: Each rule should be broken out so that they're not nested under `:host` (to reduce specificity).

src/components/slide-toggle/slide-toggle.scss
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    &.md-warn {
+      @include md-switch-checked($md-warn);
+    }
+  }
+
+  &[disabled] {
+    .md-thumb {
+      background-color: md-color($md-grey, 400);
```
jelbourn: This one shouldn't reference `$md-grey` directly, and instead use the theme's background palette.

src/components/slide-toggle/slide-toggle.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export function main() {
+  describe('MdSlideToggle', () => {
+    let builder: TestComponentBuilder;
+
+    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+      builder = tcb;
+    }));
+
+    it('should update the model correctly', (done: () => void) => {
```
jelbourn: Rather than using the `done` argument for these tests, you should be able to use the new `async()` function that will complete the test automatically when the async operations are done. You can see the checkbox tests as an example.

src/components/slide-toggle/slide-toggle.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      return builder.createAsync(TestApp).then((fixture) => {
+        let testComponent = fixture.debugElement.componentInstance;
+        let slideToggleEl = fixture.debugElement.query(By.css('md-slide-toggle')).nativeElement;
+
+        fixture.detectChanges();
+
+        expect(slideToggleEl.classList.contains('md-checked')).toBe(false);
```
jelbourn:
```ts
expect(slideToggleEl.classList).toContain('md-checked');
```

src/components/slide-toggle/slide-toggle.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+@Component({
+  selector: 'test-app',
+  template: `
+    <md-slide-toggle [(ngModel)]="slideModel" [disabled]="isDisabled">
+      <span>Test Slide Toggle</span>
+    </md-slide-toggle>
+  `,
+  directives: [MdSlideToggle]
+})
+class TestApp {
```
jelbourn: This can be more specific, e.g., `SlideToggleTestApp`

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -0,0 +1,103 @@
...
+    '(click)': 'onClick()'
+  },
+  templateUrl: './components/slide-toggle/slide-toggle.html',
+  styleUrls: ['./components/slide-toggle/slide-toggle.css']
+})
+export class MdSlideToggle implements OnInit, ControlValueAccessor {
+
+  private nativeElement: HTMLElement;
+  private switchContainer: HTMLElement;
+  private thumbContainer: HTMLElement;
```
jelbourn: I don't quite see what you need these native DOM elements for.

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    this.onTouched();
+
+    // Update the ngModel value accessor.
+    this.onChange(this._checked);
+
+    this.renderer
+      .setElementAttribute(this.nativeElement, 'aria-checked', this._checked);
+
+    this.nativeElement.classList.toggle('md-checked', this.checked);
```
robertmesserle: Any reason not to do all of this through host bindings?

src/components/slide-toggle/slide-toggle.scss
DevVersion: Definitely agree. I missed that. thx

src/components/slide-toggle/slide-toggle.scss
DevVersion: Yeah, it even has a smaller specificity and looks also much more clean :+1:

src/components/slide-toggle/slide-toggle.spec.ts
DevVersion: Very cool thanks :)

src/components/slide-toggle/slide-toggle.ts
DevVersion: I had some really weird issues with that. But I fixed those. :+1:

src/components/slide-toggle/slide-toggle.ts
DevVersion: Not needed anymore, since I fixed my issue with the Host bindings.

src/components/slide-toggle/slide-toggle.scss
DevVersion: So, I thought about adding this for hue `400` and `500` (code below) to the background palette. But I seemed for me like a duplicate and unnecessary, to add `$md-grey` to light and dark.

Please explain, what's desired?

src/components/slide-toggle/README.md
```diff
@@ -0,0 +1,44 @@
...
+```html
+<md-slide-toggle disabled>
+  Disabled Slide Toggle
+</md-slide-toggle>
+```
+
+## Theming
+A slide-toggle is default using the `accent` palette for its styling.
+
+Modifiying the color on a `slide-toggle` can be easily done, by using the following classes.
```
jelbourn: Need to update the readme for `color="..."`

src/components/slide-toggle/slide-toggle.scss
```diff
@@ -0,0 +1,133 @@
...
+    &.md-primary {
+      @include md-switch-checked($md-primary);
+    }
+
+    &.md-warn {
+      @include md-switch-checked($md-warn);
+    }
+
+    .md-slide-toggle-thumb-container {
+      transform: translate3d(100%, 0, 0);
```
jelbourn: Why `transform3d` instead of `translateX`?

DevVersion: Translate3d pushes the CSS animations into hardware acceleration. This should increase the performance.

robertmesserle: Yea, I think you typically always want to use `translate3d` if the property will be animated.

src/components/slide-toggle/slide-toggle.html
```diff
@@ -0,0 +1,22 @@
...
+    <div class="md-slide-toggle-thumb-container">
+      <div class="md-slide-toggle-thumb"></div>
+    </div>
+
+    <input #input class="md-slide-toggle-checkbox" type="checkbox"
+           [id]="getInputId()"
+           [tabIndex]="tabIndex"
+           [checked]="checked"
+           [disabled]="disabled"
+           [attr.name]="name"
```
jelbourn: Should be able to just do `[name]` without the `attr.`

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -0,0 +1,130 @@
...
+  providers: [MD_SLIDE_TOGGLE_VALUE_ACCESSOR],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MdSlideToggle implements ControlValueAccessor {
+
+  private onChange = (_: any) => {};
+  private onTouched = () => {};
+
+  // A unique id for the slide-toggle. By default the id is auto-generated.
+  private _uniqueId = `md-slide-toggle-${++nextId}`;
```
jelbourn: You can just initialize `id` to this without the `_uniqueId` field; if the user gives their own id, it will overwrite this (which is fine).

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -0,0 +1,130 @@
...
+  @Input() id: string = this._uniqueId;
+  @Input() tabIndex: number = 0;
+  @Input() ariaLabel: string = '';
+  @Input() ariaLabelledby: string = null;
+
+  @Output('change') private _change: EventEmitter<boolean> = new EventEmitter<boolean>();
+  change: Observable<boolean> = this._change.asObservable();
+
+  // Returns the unique id for the visual hidden input.
+  getInputId = () => `${this.id || this._uniqueId}-input`;
```
jelbourn: If you remove uniqueId, this should still be fine as `${this.id}-input`

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  constructor(private elementRef: ElementRef,
+              private renderer: Renderer) {
```
jelbourn: Prefix `private` fields with underscore.

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -0,0 +1,130 @@
...
+    this.checked = value;
+  }
+
+  /** @internal */
+  registerOnChange(fn: any): void {
+    this.onChange = fn;
+  }
+
+  /** @internal */
+  registerOnTouched(fn: any): void {
```
jelbourn: Can you make the comment on each of these:
```ts
  /**
   * Implemented as part of ControlValueAccessor.
   * @internal
   */
```
I know we repeat it everywhere, but I think it makes it apparent for anyone that jumps into the source for a single component.

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -0,0 +1,130 @@
...
+  @Input()
+  get checked() {
+    return !!this._checked;
+  }
+
+  set checked(value) {
+    if (this.checked !== !!value) {
+      this._checked = value;
+      this.onChange(this._checked);
+      this._change.emit(this._checked);
```
jelbourn: One of the things I ran into recently was that checkbox and radio were firing `change` events for their initial value, which is unexpected. I fixed it by adding an `_isInitialized` field and setting it to true in `ngAfterContentInit` and then gating the `emit` on `_isInitialized` being true. This also prevents `ng-pristine` from being immediately changed to `ng-dirty` on first load.

src/core/style/_palette.scss
```diff
@@ -335,7 +335,9 @@ $md-light-theme-background: (
   hover:             rgba(black, 0.04), // TODO(kara): check style with Material Design UX
   card:       white,
   dialog:     white,
-  disabled-button:   rgba(black, 0.12)
+  disabled-button:   rgba(black, 0.12),
+  slide-toggle-bar: map_get($md-grey, 500),
+  slide-toggle-thumb: map_get($md-grey, 400)
```
jelbourn: Thinking about this more, if these are super specific to the slide-toggle we should probably just move them to slide-toggle.scss and use `$md-is-dark-theme` directly there. Something like
```scss
$md-slide-toggle-bar-color: if($md-is-dark-theme, ..., ...);
...
```

src/components/slide-toggle/slide-toggle.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+
+      it('should update the model correctly', () => {
+        expect(slideToggleElement.classList).not.toContain('md-checked');
+
+        testComponent.slideModel = true;
+        fixture.detectChanges();
+
+        expect(slideToggleElement.classList).toContain('md-checked');
+      });
```
jelbourn: There should be a test checking that the `NgControl` pristine and touched states are set correctly. 
You can see where I do this for radio button in my PR: https://github.com/angular/material2/pull/443/files#diff-ebc684a852af87cf7ac222fa6411cbe2R228

jelbourn:
Fantastic PR! Just a few comments for things that are in-flight for existing components.

src/components/slide-toggle/slide-toggle.ts
DevVersion: Yea, that should be the same.

src/core/style/_palette.scss
DevVersion: Yes. I had that before. Checking for dark, is not needed, since the slide-toggle always uses the same hues for dark or light themes. Added a comment for :)

src/components/slide-toggle/slide-toggle.ts
DevVersion: @jelbourn Just tested that again, and created some tests, but it seems like we didn't notice, that sometimes the id will set to null. And then we have to fallback to the unique id.

Currently, we will have ids on input like
```
null-input
```

src/components/slide-toggle/slide-toggle.ts
DevVersion: I tested that, and it is now covered by tests. But it seems like this is a more specific problem with the radio component. I cannot reproduce this issue with the `slide-toggle` component.

But, I've noticed that as well, some commits before. But it seems like I fixed it anyway. 

eg. this check in the setter, does not trigger any change (if present) before init, because it's the same value.
```ts
    if (this.checked !== !!value) {

```

src/components/slide-toggle/slide-toggle.html
DevVersion: This is a good point. I also had that before. But this will cause issues with a falsy value.

For example, if the user, uses `[name]="myName", and the `myName` property is set to `null` will result in
```html
<input name="null">
```

src/components/slide-toggle/slide-toggle.spec.ts
DevVersion: Done :smile:

src/components/slide-toggle/slide-toggle.html
jelbourn: After trying it out, you're right, `attr.name` is what we want for now.

src/components/slide-toggle/README.md
```diff
@@ -0,0 +1,66 @@
+# MdSlideToggle
+`MdSlideToggle` is a two-state control, which can be also called `switch`
+
+### Screenshots
+![image](https://raw.githubusercontent.com/angular/code.material.angularjs.org/master/material2_assets/slide-toggle/toggles.png)
```
jelbourn: I added a hosted asset for you:
https://material.angularjs.org/material2_assets/slide-toggle/toggles.png

jelbourn:
LGTM, just needs a rebase and to update the readme asset.

src/components/slide-toggle/README.md
DevVersion: Haha :smile: I was already using that, but over the GitHub user content

DevVersion:
@jelbourn Everything done

PR #372, opened 2016-04-28
docs(icon): add README.md

dozingcat:


kara:
Can you change your commit name to "docs(icon)"? :D

src/components/icon/README.md
```diff
@@ -0,0 +1,116 @@
...
+### Ligatures
+
+Some fonts are designed to show icons by using [ligatures](https://en.wikipedia.org/wiki/Typographic_ligature), for example by rendering the text "home" as a home image. To use a ligature icon, put its text in the content of the `md-icon` component, for example:
+```html
+<md-icon>home</md-icon>
+```
+
+By default the [Material icons font](http://google.github.io/material-design-icons/#icon-font-for-the-web) is used. (You will still need to include the HTML to load the font and its CSS, as described in the link). You can specify a different font by setting the `fontSet` input to either the CSS class to apply to use the desired font, or to an alias previously registered with `MdIconRegistry.registerFontClassAlias`, for example:
+```javascript
+mdIconRegistry.registerFontClassAlias('myfont', 'my-icon-font-class');
```
jelbourn: Can we use [FontAwesome](https://fortawesome.github.io/Font-Awesome/) in the example?

dozingcat: FontAwesome doesn't use ligatures as far as I can tell so it might be confusing. Changed the CSS font icon example below.

jelbourn:
LGTM

PR #373, opened 2016-04-28
fix(sidenav): fullscreen sidenavs should not scroll

kara:
r: @jelbourn 

This PR ensures that it's not possible to scroll past the sidenav backdrop when the sidenav is open.

### DIFF
--- a/src/components/sidenav/sidenav.ts
+++ b/src/components/sidenav/sidenav.ts
@@ -13,6 +13,7 @@ import {
   Type,
   ChangeDetectionStrategy,
   EventEmitter,
+  Renderer
 } from 'angular2/core';
 import {BaseException} from 'angular2/src/facade/exceptions';
 import {Dir} from '../../core/rtl/dir';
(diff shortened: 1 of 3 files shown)


src/components/sidenav/sidenav.ts
```diff
@@ -258,9 +260,20 @@ export class MdSidenavLayout implements AfterContentInit {
   ngAfterContentInit() {
     // On changes, assert on consistency.
     this._sidenavs.changes.subscribe(() => this._validateDrawers());
+    this._sidenavs.forEach((sidenav: MdSidenav) => this._watchSidenavToggle(sidenav));
     this._validateDrawers();
   }
 
+  private _watchSidenavToggle(sidenav: MdSidenav): void {
```
jelbourn: Add functions descriptions for these two

kara: Updated!

jelbourn:
LGTM

Issue #295, opened 2016-04-12
Indeterminate progress-circle renders nothing on IE11


jelbourn:
Determinate works, but the animations don't run (it just jumps).

Indeterminate renders nothing.

[Microsoft's site says that IE11 doesn't support transitions and animations on SVG elements](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/csstransitionsanimationsforsvgelements). Is that the issue?

PR #375, opened 2016-04-28
fix(progress-circle): render determinate and indeterminate across all supported browsers

josephperrott:
For #295

### DIFF
--- a/src/components/progress-circle/progress-circle.scss
+++ b/src/components/progress-circle/progress-circle.scss
@@ -4,14 +4,11 @@
 
 /* Animation Durations */
 $md-progress-circle-duration : 5.25s !default;
-$md-progress-circle-value-change-duration : $md-progress-circle-duration * 0.25 !default;
 $md-progress-circle-constant-rotate-duration : $md-progress-circle-duration * 0.55 !default;
 $md-progress-circle-sporadic-rotate-duration : $md-progress-circle-duration !default;
 
 /** Component sizing */
 $md-progress-circle-stroke-width: 10px !default;
-$md-progress-circle-radius: 40px !default;
-$md-progress-circle-circumference: $pi * $md-progress-circle-radius * 2 !default;
 // Height and weight of the viewport for md-progress-circle.
 $md-progress-circle-viewport-size : 100px !default;
 
@@ -24,51 +21,31 @@ $md-progress-circle-viewport-size : 100px !default;
   width: $md-progress-circle-viewport-size;
 
   /** SVG's viewBox is defined as 0 0 100 100, this means that all SVG children will placed
-      based on a 100px by 100px box.
-
-      The SVG and Circle dimensions/location:
-        SVG
-          Height: 100px
-          Width: 100px
-        Circle
-          Radius: 40px
-          Circumference: 251.3274px
-          Center x: 50px
-          Center y: 50px
+      based on a 100px by 100px box.  Additionally all SVG sizes and locations are in reference to
+      this viewBox.
   */
   svg {
     height: 100%;
     width: 100%;
-    transform: rotate(-90deg);
     transform-origin: center;
   }
 
 
-  circle {
+  path {
     fill: transparent;
     stroke: md-color($md-primary, 600);
     /** Stroke width of 10px defines stroke as 10% of the viewBox */
     stroke-width: $md-progress-circle-stroke-width;
-    /** SVG circle rotations begin rotated 90deg clockwise from the circle's center top */
-    transition: stroke-dashoffset 0.225s linear;
-    /** The dash array of the circle is defined as the circumference of the circle. */
-    stroke-dasharray: $md-progress-circle-circumference;
-    /** The stroke dashoffset is used to "fill" the circle, 0px represents an full circle,
-        while the circles full circumference represents an empty circle. */
-    stroke-dashoffset: 0px;
   }
 
-
-  &[color="accent"] circle {
+  &[color="accent"] path {
     stroke: md-color($md-accent, 600);
   }
 
 
-  &[color="warn"] circle {
+  &[color="warn"] path {
     stroke: md-color($md-warn, 600);
   }
-
-
 
 
   &[mode="indeterminate"] {
@@ -80,14 +57,6 @@ $md-progress-circle-viewport-size : 100px !default;
                                linear;
     animation-iteration-count: infinite;
     transition: none;
-
-    circle {
-      animation-duration: $md-progress-circle-value-change-duration;
-      animation-name: md-progress-circle-value-change;
-      animation-timing-function: $ease-in-out-curve-function;
-      animation-iteration-count: infinite;
-      transition: none;
-    }
   }
 }
 
@@ -106,8 +75,4 @@ $md-progress-circle-viewport-size : 100px !default;
   75%      { transform: rotate( 810deg); }
   87.5%    { transform: rotate( 945deg); }
   100%     { transform: rotate(1080deg); }
-}
-@keyframes md-progress-circle-value-change {
-  0%       { stroke-dashoffset: 261.3274px; }
-  100%     { stroke-dashoffset: -241.3274px; }
 }
--- a/src/components/progress-circle/progress-circle.ts
+++ b/src/components/progress-circle/progress-circle.ts
@@ -1,12 +1,29 @@
 import {
   Component,
-  ChangeDetectionStrategy,
   HostBinding,
+  ChangeDetectorRef,
+  ChangeDetectionStrategy,
   Input
 } from 'angular2/core';
 
-
 // TODO(josephperrott): Benchpress tests.
+
+/** A single degree in radians. */
+const DEGREE_IN_RADIANS = Math.PI / 180;
+/** Duration of the indeterminate animation. */
+const DURATION_INDETERMINATE = 667;
+/** Duration of the indeterminate animation. */
+const DURATION_DETERMINATE = 225;
+/** Start animation value of the indeterminate animation */
+let startIndeterminate = 3;
+/** End animation value of the indeterminate animation */
+let endIndeterminate = 80;
+
+
+export type ProgressCircleMode = 'determinate' | 'indeterminate';
+
+type EasingFn = (currentTime: number, startValue: number,
+                 changeInValue: number, duration: number) => number
 
 
 /**
@@ -24,15 +41,43 @@ import {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class MdProgressCircle {
+  /** The id of the last requested animation. */
+  private _lastAnimationId: number = 0;
+
+  /** The id of the indeterminate interval. */
+  private _interdeterminateInterval: number;
+
+  /** The current path value, representing the progres circle. */
+  private _currentPath: string;
+  get currentPath() {
+    return this._currentPath;
+  }
+  set currentPath(path: string) {
+    this._currentPath = path;
+    // Mark for check as our ChangeDetectionStrategy is OnPush, when changes come from within the
+    // component, change detection must be called for.
+    this._changeDetectorRef.markForCheck();
+  }
+
   /**
    * Value of the progress circle.
    *
    * Input:number, defaults to 0.
    * _value is bound to the host as the attribute aria-valuenow.
    */
+  private _value: number = 0;
+  @Input()
   @HostBinding('attr.aria-valuenow')
-  @Input('value')
-  _value: number = 0;
+  get value() {
+    return this._value;
+  }
+  set value(v: number) {
+    if (v) {
+      let newValue = clamp(v);
+      this._animateCircle(this.value, newValue, linearEase, DURATION_DETERMINATE, 0);
+      this._value = newValue;
+    }
+  }
 
   /**
    * Mode of the progress circle
@@ -41,46 +86,99 @@ export class MdProgressCircle {
    * mode is bound to the host as the attribute host.
    */
   @HostBinding('attr.mode')
-  @Input() mode: 'determinate' | 'indeterminate' = 'determinate';
+  @Input()
+  get mode() {
+    return this._mode;
+  }
+  set mode(m: ProgressCircleMode) {
+    if (m == 'indeterminate') {
+      this._startIndeterminateAnimation();
+    } else {
+      this._cleanupIndeterminateAnimation();
+    }
+    this._mode = m;
+  }
+  private _mode: ProgressCircleMode = 'determinate';
+
+  constructor(private _changeDetectorRef: ChangeDetectorRef) {
+  }
 
 
   /**
-   * Gets the current stroke dash offset to represent the progress circle.
+   * Animates the circle from one percentage value to another.
    *
-   * The stroke dash offset specifies the distance between dashes in the circle's stroke.
-   * Setting the offset to a percentage of the total circumference of the circle, fills this
-   * percentage of the overall circumference of the circle.
+   * @param animateFrom The percentage of the circle filled starting the animation.
+   * @param animateTo The percentage of the circle filled ending the animation.
+   * @param ease The easing function to manage the pace of change in the animation.
+   * @param duration The length of time to show the animation, in milliseconds.
+   * @param rotation The starting angle of the circle fill, with 0° represented at the top center
+   *    of the circle.
    */
-  strokeDashOffset() {
-    // To determine how far the offset should be, we multiple the current percentage by the
-    // total circumference.
+  private _animateCircle(animateFrom: number, animateTo: number, ease: EasingFn,
+                        duration: number, rotation: number) {
+    let id = ++this._lastAnimationId;
+    let startTime = now();
+    let changeInValue = animateTo - animateFrom;
 
-    // The total circumference is calculated based on the radius we use, 45.
-    // PI * 2 * 45
-    return 251.3274 * (100 - this._value) / 100;
-  }
+    // No need to animate it if the values are the same
+    if (animateTo === animateFrom) {
+      this.currentPath = getSvgArc(animateTo, rotation);
+    } else {
+      let animation = (currentTime: number) => {
+        let elapsedTime = Math.max(
+          0, Math.min((currentTime || now()) - startTime, duration));
 
+        this.currentPath = getSvgArc(
+          ease(elapsedTime, animateFrom, changeInValue, duration),
+          rotation
+        );
 
-  /** Gets the progress value, returning the clamped value. */
-  get value() {
-    return this._value;
-  }
-
-
-  /** Sets the progress value, clamping before setting the internal value. */
-  set value(v: number) {
-    if (v != null) {
-      this._value = MdProgressCircle.clamp(v);
+        // Prevent overlapping animations by checking if a new animation has been called for and
+        // if the animation has lasted long than the animation duration.
+        if (id === this._lastAnimationId && elapsedTime < duration) {
+          requestAnimationFrame(animation);
+        }
+      };
+      requestAnimationFrame(animation);
     }
   }
 
 
-  /** Clamps a value to be between 0 and 100. */
-  static clamp(v: number) {
-    return Math.max(0, Math.min(100, v));
+  /**
+   * Starts the indeterminate animation interval, if it is not already running.
+   */
+  private _startIndeterminateAnimation() {
+    let rotationStartPoint = 0;
+    let start = startIndeterminate;
+    let end =  endIndeterminate;
+    let duration = DURATION_INDETERMINATE;
+    let animate = () => {
+      this._animateCircle(start, end, materialEase, duration, rotationStartPoint);
+      // Prevent rotation from reaching Number.MAX_SAFE_INTEGER.
+      rotationStartPoint = (rotationStartPoint + end) % 100;
+      let temp = start;
+      start = -end;
+      end = -temp;
+    };
+
+    if (!this._interdeterminateInterval) {
+      this._interdeterminateInterval = setInterval(
+        animate, duration + 50, 0, false);
+      animate();
+    }
+  }
+
+
+  /**
+   * Removes interval, ending the animation.
+   */
+  private _cleanupIndeterminateAnimation() {
+    if (this._interdeterminateInterval) {
+      clearInterval(this._interdeterminateInterval);
+      this._interdeterminateInterval = null;
+    }
   }
 }
-
 
 
 /**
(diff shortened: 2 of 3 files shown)


src/components/progress-circle/progress-circle.scss
```diff
@@ -3,15 +3,12 @@
 @import "default-theme";
 
 /* Animation Durations */
-$md-progress-circle-duration : 5.25s !default;
-$md-progress-circle-value-change-duration : $md-progress-circle-duration * 0.25 !default;
+$md-progress-circle-duration : 5332ms !default;
```
jelbourn: Can you add any info on where this value comes from?

src/components/progress-circle/progress-circle.ts
```diff
@@ -1,13 +1,19 @@
 import {
   Component,
-  ChangeDetectionStrategy,
   HostBinding,
+  ChangeDetectorRef,
+  ChangeDetectionStrategy,
   Input
 } from 'angular2/core';
-
+import {isPresent} from 'angular2/src/facade/lang';
```
jelbourn: `isPresent` is gone. Just use truthiness or falsiness now.

src/components/progress-circle/progress-circle.ts
```diff
@@ -1,13 +1,19 @@
...
 
 // TODO(josephperrott): Benchpress tests.
 
+let DEGREE_IN_RADIANS = Math.PI / 180;
+let durationIndeterminate: number = 5332 / 8;
+let startIndeterminate = 3;
+let endIndeterminate = 80;
```
jelbourn: Need some comments what what each of these are used for.

src/components/progress-circle/progress-circle.ts
```diff
@@ -25,14 +31,51 @@ import {
...
   @HostBinding('attr.aria-valuenow')
-  @Input('value')
-  _value: number = 0;
+  get value() {
+    return this._value;
+  }
+  set value(v: number) {
+    if (isPresent(v)) {
+      let newValue = MdProgressCircle.clamp(v);
+      this._renderCircle(this.value, newValue, MdProgressCircle.linearEase, 225, 0);
```
jelbourn: Where does the `225` here come from? Move it to a constant?

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
    * mode is bound to the host as the attribute host.
    */
   @HostBinding('attr.mode')
-  @Input() mode: 'determinate' | 'indeterminate' = 'determinate';
+  @Input()
+  get mode() {
+    return this._mode;
+  }
+  set mode(m: 'determinate' | 'indeterminate') {
```
jelbourn: Rather than recreating the string literal type each time, you can declare it once and reuse it:
```ts
export type ProgressCircleMode = 'determinate' | 'indeterminate';
```

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
+      largeArcFlag = endAngle <= 180 ? 0 : 1;
+    }
+
+    return `M${start}A${pathRadius},${pathRadius} 0 ${largeArcFlag},${arcSweep} ${end}`;
+  }
+
+
   /** Clamps a value to be between 0 and 100. */
-  static clamp(v: number) {
+  private static clamp(v: number) {
```
jelbourn: Rather than being a private static method, you can just have a function that lives in the module. 
```ts
function clamp() {
  ...
}
```

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
+
+
+  /**
+   * Returns the current timestamp either based on the performance global or a date object.
+   */
+  private static _animationNow() {
+    if (window.performance) {
+      return performance.now();
+    }
+    return (new Date).getTime();
```
jelbourn: `Date.now()` is available in all of the browsers that we support.

src/components/progress-circle/progress-circle.ts
```diff
@@ -25,14 +31,51 @@ import {
 })
 export class MdProgressCircle {
   /**
+   * The id of the last requested animation.
+   */
+  private _lastAnimationId: number = 0;
+
+  /**
+   * The id of the indeterminate interval.
```
jelbourn:
```ts
/** The id of the most recent indeterminate setInterval. */
```

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
-   *
-   * The stroke dash offset specifies the distance between dashes in the circle's stroke.
-   * Setting the offset to a percentage of the total circumference of the circle, fills this
-   * percentage of the overall circumference of the circle.
+   * Animates the rendered circle.
    */
-  strokeDashOffset() {
-    // To determine how far the offset should be, we multiple the current percentage by the
-    // total circumference.
+  private _renderCircle(animateFrom: number, animateTo: number, ease: Function,
```
jelbourn: Definitely need a description that explains how this functions works at a high level.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
-    return 251.3274 * (100 - this._value) / 100;
+        this.currentPath = MdProgressCircle.getSvgArc(
+          ease(elapsedTime, animateFrom, changeInValue, animationDuration),
+          rotation
+        );
+
+        // Prevent overlapping animations by checking if a new animation has been called for and
+        // if the animation has lasted long than the animation duration.
+        if (id === this._lastAnimationId && elapsedTime < animationDuration) {
+          window.requestAnimationFrame(animation);
```
jelbourn: You shouldn't need to use `window.` anywhere. Just `requestAnimationFrame` or `setInterval` etc.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
+  private static clamp(v: number) {
     return Math.max(0, Math.min(100, v));
   }
-}
 
 
+  /**
+   * Easing function for linear animation.
+   */
+  private static linearEase(t: number, b: number, c: number, d: number) {
```
jelbourn: Isn't an easing function supposed to just be a function of `t`? What are the other arguments?

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
+
+  /**
+   * Easing function to match material design indeterminate animation.
+   */
+  private static materialEase(t: number, b: number, c: number, d: number) {
+    // via http://www.timotheegroleau.com/Flash/experiments/easing_function_generator.htm
+    // with settings of [0, 0, 1, 1]
+    var ts = (t /= d) * t;
+    var tc = ts * t;
+    return b + c * (6 * tc * ts + -15 * ts * ts + 10 * tc);
```
jelbourn: Even with the link, we need some explanation of what's going on here

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
+
+    return (radius + (pathRadius * Math.cos(angleInRadians))) +
+      ',' + (radius + (pathRadius * Math.sin(angleInRadians)));
+  }
+
+
+  /**
+   * Returns the current timestamp either based on the performance global or a date object.
+   */
+  private static _animationNow() {
```
jelbourn: This can just be a module function and you can call it `now`.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
+      this._interdeterminateInterval = null;
     }
   }
 
 
+  /**
+   * Determines the path value to define the arc.
+   */
+  private static getSvgArc(current: number, rotation: number) {
+    // The angle can't be exactly 360, because the arc becomes hidden.
```
jelbourn: Also need high-level description of what's going on here.

src/components/progress-circle/progress-circle.ts
jelbourn: Rather than using the type `Function`, it should be a type with the function signature of the easing function.

src/components/progress-circle/progress-circle.ts
```diff
@@ -1,13 +1,19 @@
...
+  ChangeDetectorRef,
+  ChangeDetectionStrategy,
   Input
 } from 'angular2/core';
-
+import {isPresent} from 'angular2/src/facade/lang';
 
 // TODO(josephperrott): Benchpress tests.
 
+let DEGREE_IN_RADIANS = Math.PI / 180;
```
robertmesserle: Should this be a const?

robertmesserle: Also, should all of these maybe be constants?  I only picked this one because it's all caps.

src/components/progress-circle/progress-circle.ts
```diff
@@ -25,14 +31,51 @@ import {
...
+
+  /**
+   * The id of the indeterminate interval.
+   */
+  private _interdeterminateInterval: number;
+
+  /**
+   * A reference to the change detector.
+   */
+  private _ref: ChangeDetectorRef;
```
robertmesserle: Maybe a more descriptive name?

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
    */
-  strokeDashOffset() {
-    // To determine how far the offset should be, we multiple the current percentage by the
-    // total circumference.
+  private _renderCircle(animateFrom: number, animateTo: number, ease: Function,
+                        duration: number, rotation: number) {
+    let id = ++this._lastAnimationId;
+    let startTime = MdProgressCircle._animationNow();
+    let changeInValue = animateTo - animateFrom;
+    let animationDuration = duration || 225;
```
robertmesserle: Should this be set as a default value in the arguments?  ie. `duration: number = 225`

robertmesserle: Or do we even really need a default here on a private method?

src/components/progress-circle/progress-circle.ts
robertmesserle: This looks like it was modeled after one of the Robert Penner easing functions:
`// t: current time, b: begInnIng value, c: change In value, d: duration`

Maybe expand these names to make it more obvious what is going on.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,47 +84,172 @@ export class MdProgressCircle {
...
+    return (radius + (pathRadius * Math.cos(angleInRadians))) +
+      ',' + (radius + (pathRadius * Math.sin(angleInRadians)));
+  }
+
+
+  /**
+   * Returns the current timestamp either based on the performance global or a date object.
+   */
+  private static _animationNow() {
+    if (window.performance) {
```
robertmesserle: Would it make sense to check with `typeof performance !== 'undefined'` rather than checking `window`?

src/components/progress-circle/progress-circle.ts
josephperrott: Done.

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done, just to confirm, we do support iOS safari 8.4 as this is the only browser that may fall in scope of supported browsers that does that have performance.now

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.scss
josephperrott: Reverted back

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done, I moved to more descriptive params.

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Updated to move the appropriate one to consts

src/components/progress-circle/progress-circle.ts
josephperrott: True, we can be sure that it will always get the duration provided since its private.

jelbourn:
@joeyperson  Now needs rebasing on my mega change #384. Let me know if you need any help with that (next week, of course).

josephperrott:
Rebased.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,48 +84,100 @@ export class MdProgressCircle {
...
-   * Setting the offset to a percentage of the total circumference of the circle, fills this
-   * percentage of the overall circumference of the circle.
+   * Animates the rendered circle.
    */
-  strokeDashOffset() {
-    // To determine how far the offset should be, we multiple the current percentage by the
-    // total circumference.
+  private _renderCircle(animateFrom: number, animateTo: number,
+                        ease: (currentTime: number, startValue: number,
+                               changeInValue: number, duration: number) => number,
```
jelbourn: You should be able to alias this type so that it's a little more readable here:
```ts
type EasingFn = (currentTime: number, startValue: number, changeInValue: number, duration: number) => number
```
(see the `Dir` directive for an example)

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,48 +84,100 @@ export class MdProgressCircle {
...
-   *
-   * The stroke dash offset specifies the distance between dashes in the circle's stroke.
-   * Setting the offset to a percentage of the total circumference of the circle, fills this
-   * percentage of the overall circumference of the circle.
+   * Animates the rendered circle.
    */
-  strokeDashOffset() {
-    // To determine how far the offset should be, we multiple the current percentage by the
-    // total circumference.
+  private _renderCircle(animateFrom: number, animateTo: number,
```
jelbourn: Could you add some `@param` descriptions for this function? I think it would be helpful since it's not drop-dead obvious what the parameters are. Check out the grid-list for examples.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,48 +84,100 @@ export class MdProgressCircle {
...
+    let id = ++this._lastAnimationId;
+    let startTime = now();
+    let changeInValue = animateTo - animateFrom;
 
-    // The total circumference is calculated based on the radius we use, 45.
-    // PI * 2 * 45
-    return 251.3274 * (100 - this._value) / 100;
-  }
+    // No need to animate it if the values are the same
+    if (animateTo === animateFrom) {
```
jelbourn: I think this fn would be a little more eadable if it were formatted like
```ts
    // No need to animate it if the values are the same
    if (animateTo === animateFrom) {
      this.currentPath = getSvgArc(animateTo, rotation);
      return;
    }

    requestAnimationFrame(now => {
      // ...
    });
```

jelbourn: Oh, nevermind on the second part, I missed the re-use of the `animation` inside.

src/components/progress-circle/progress-circle.ts
```diff
@@ -96,11 +191,93 @@ export class MdProgressCircle {
...
+  return changeInValue * currentTime / duration + startValue;
+}
+
+
+/**
+ * Easing function to match material design indeterminate animation.
+ */
+function materialEase(currentTime: number, startValue: number,
+                      changeInValue: number, duration: number) {
+  var ts = (currentTime /= duration) * currentTime;
```
jelbourn: It's weird to use `/=` in an intermediate calculation.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,48 +84,100 @@ export class MdProgressCircle {
...
+          ease(elapsedTime, animateFrom, changeInValue, duration),
+          rotation
+        );
 
-  /** Gets the progress value, returning the clamped value. */
-  get value() {
-    return this._value;
+        // Prevent overlapping animations by checking if a new animation has been called for and
+        // if the animation has lasted long than the animation duration.
+        if (id === this._lastAnimationId && elapsedTime < duration && pageVisible()) {
```
jelbourn: Since you're gating on `document.hidden`, do you need to listen for `visibilitychange` events to restart later?

src/components/progress-circle/progress-circle.ts
```diff
@@ -96,11 +191,93 @@ export class MdProgressCircle {
...
+  var tc = ts * currentTime;
+  return startValue + changeInValue * (6 * tc * ts + -15 * ts * ts + 10 * tc);
+}
+
+
+/**
+ * Determines the path value to define the arc.
+ */
+function getSvgArc(current: number, rotation: number) {
+  // The angle can't be exactly 360, because the arc becomes hidden.
```
jelbourn: This function would benefit from some more explanation of what's going on at a high level.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,48 +84,100 @@ export class MdProgressCircle {
...
+        // Prevent rotation from reaching Number.MAX_SAFE_INTEGER.
+        rotationStartPoint = (rotationStartPoint + end) % 100;
+        let temp = start;
+        start = -end;
+        end = -temp;
+      }
+    };
+
+    if (!this._interdeterminateInterval) {
+      // Note that this interval isn't supposed to trigger a digest.
```
jelbourn: We don't say the word "digest" any more 😉

When a `setInterval` function runs, it _will_ trigger a round of change detection unless you explicitly inject the `NgZone` and use that to `runOutsideAngular`.

src/components/progress-circle/progress-circle.ts
```diff
@@ -24,15 +38,44 @@ import {
...
+  private _changeDetectorRef: ChangeDetectorRef;
+
+  /** The current path value, representing the progres circle. */
+  private _currentPath: string;
+  get currentPath() {
+    return this._currentPath;
+  }
+  set currentPath(path: string) {
+    this._currentPath = path;
+    this._changeDetectorRef.markForCheck();
```
jelbourn: Add explanation of why you need to do this.

src/components/progress-circle/progress-circle.ts
```diff
@@ -96,11 +191,93 @@ export class MdProgressCircle {
...
+    largeArcFlag = endAngle >= -180 ? 0 : 1;
+  } else {
+    largeArcFlag = endAngle <= 180 ? 0 : 1;
+  }
+
+  return `M${start}A${pathRadius},${pathRadius} 0 ${largeArcFlag},${arcSweep} ${end}`;
+}
+
+function pageVisible() {
+  return !document.hidden;
```
jelbourn: I'd call this function `isPageVisible`, but then you might as well just use `document.hidden` directly since it's not much shorter. Up to you.

src/components/progress-circle/progress-circle.ts
```diff
@@ -96,11 +191,93 @@ export class MdProgressCircle {
...
+  var ts = (currentTime /= duration) * currentTime;
+  var tc = ts * currentTime;
+  return startValue + changeInValue * (6 * tc * ts + -15 * ts * ts + 10 * tc);
+}
+
+
+/**
+ * Determines the path value to define the arc.
+ */
+function getSvgArc(current: number, rotation: number) {
```
jelbourn: Need more explanation about what the function is returning.

src/components/progress-circle/progress-circle.ts
josephperrott: Done.

src/components/progress-circle/progress-circle.ts
josephperrott: Sorry, didn't realize I left this in.  browsers already prevent the rendering when not visible, so this is redundant.

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Removed.

src/components/progress-circle/progress-circle.ts
josephperrott: Done.

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,48 +89,102 @@ export class MdProgressCircle {
...
+    if (m == 'indeterminate') {
+      this._startIndeterminateAnimation();
+    } else {
+      this._cleanupIndeterminateAnimation();
+    }
+    this._mode = m;
+  }
+  private _mode: ProgressCircleMode = 'determinate';
+
+  constructor(changeDetectorRef: ChangeDetectorRef) {
```
jelbourn: You can shorten this to 
```ts
constructor(private _changeDetectorRef: ChangeDetectorRef) { }
```

src/components/progress-circle/progress-circle.ts
```diff
@@ -96,11 +198,101 @@ export class MdProgressCircle {
...
+/**
+ * Easing function to match material design indeterminate animation.
+ */
+function materialEase(currentTime: number, startValue: number,
+                      changeInValue: number, duration: number) {
+  let time = currentTime / duration;
+  let timeCubed = Math.pow(time, 3);
+  let timeQuad = Math.pow(time, 4);
+  let timeQuint = Math.pow(time, 5);
+  return startValue + changeInValue * ((6 * timeQuint) +
```
jelbourn: This whole expression will actually fit on one line (our max is 100 cols)

jelbourn:
LGTM aside from a couple minor style nits.

src/components/progress-circle/progress-circle.ts
josephperrott: Done

src/components/progress-circle/progress-circle.ts
josephperrott: Done.

PR #376, opened 2016-04-28
feat(tabs): adds the `md-tab-group` component

robertmesserle:
@jelbourn @kara @hansl

### DIFF
--- a/src/components/tab-group/ink-bar.ts
+++ b/src/components/tab-group/ink-bar.ts
@@ -1,0 +1,41 @@
+import {Directive, Renderer, ElementRef} from '@angular/core';
+
+/**
+ * The ink-bar is used to display and animate the line underneath the current active tab label.
+ * @internal
+ */
+@Directive({
+  selector: 'md-ink-bar',
+})
+export class MdInkBar {
+  constructor(private _renderer: Renderer, private _elementRef: ElementRef) {}
+
+  /**
+   * Calculates the styles from the provided element in order to align the ink-bar to that element.
+   * @param element
+   */
+  alignToElement(element: HTMLElement) {
+    this._renderer.setElementStyle(this._elementRef.nativeElement, 'left',
+        this._getLeftPosition(element));
+    this._renderer.setElementStyle(this._elementRef.nativeElement, 'width',
+        this._getElementWidth(element));
+  }
+
+  /**
+   * Generates the pixel distance from the left based on the provided element in string format.
+   * @param element
+   * @returns {string}
+   */
+  private _getLeftPosition(element: HTMLElement): string {
+    return element.offsetLeft + 'px';
+  }
+
+  /**
+   * Generates the pixel width from the provided element in string format.
+   * @param element
+   * @returns {string}
+   */
+  private _getElementWidth(element: HTMLElement): string {
+    return element.offsetWidth + 'px';
+  }
+}
--- a/src/components/tab-group/tab-content.ts
+++ b/src/components/tab-group/tab-content.ts
@@ -1,0 +1,12 @@
+import {Directive, TemplateRef, ViewContainerRef} from '@angular/core';
+import {TemplatePortalDirective} from '../../core/portal/portal-directives';
+
+/** Used to flag tab contents for use with the portal directive */
+@Directive({
+  selector: '[md-tab-content]'
+})
+export class MdTabContent extends TemplatePortalDirective {
+  constructor(templateRef: TemplateRef<any>, viewContainerRef: ViewContainerRef) {
+    super(templateRef, viewContainerRef);
+  }
+}
--- a/src/components/tab-group/tab-group.html
+++ b/src/components/tab-group/tab-group.html
@@ -1,0 +1,25 @@
+<div class="md-tab-header" role="tablist"
+     (keydown.arrowRight)="focusNextTab()"
+     (keydown.arrowLeft)="focusPreviousTab()"
+     (keydown.enter)="selectedIndex = focusIndex">
+  <div class="md-tab-label" role="tab" md-tab-label-wrapper
+       *ngFor="let label of labels; let i = index"
+       [id]="getTabLabelId(i)"
+       [tabIndex]="selectedIndex == i ? 0 : -1"
+       [attr.aria-controls]="getTabContentId(i)"
+       [attr.aria-selected]="selectedIndex == i"
+       [class.md-active]="selectedIndex == i"
+       (click)="focusIndex = selectedIndex = i">
+    <template [portalHost]="label"></template>
+  </div>
+  <md-ink-bar></md-ink-bar>
+</div>
+<div class="md-tab-body-wrapper">
+  <div class="md-tab-body"
+       *ngFor="let content of contents; let i = index"
+       [id]="getTabContentId(i)"
+       [class.md-active]="selectedIndex == i"
+       [attr.aria-labelledby]="getTabLabelId(i)">
+    <template role="tabpanel" [portalHost]="content" *ngIf="selectedIndex == i"></template>
+  </div>
+</div>
--- a/src/components/tab-group/tab-group.scss
+++ b/src/components/tab-group/tab-group.scss
@@ -1,0 +1,64 @@
+@import 'variables';
+@import 'default-theme';
+
+$md-tab-bar-height: 48px !default;
+
+:host {
+  display: block;
+  font-family: $md-font-family;
+}
+
+/** The top section of the view; contains the tab labels */
+.md-tab-header {
+  overflow: hidden;
+  position: relative;
+  display: flex;
+  flex-direction: row;
+  border-bottom: 1px solid md-color($md-background, status-bar);
+}
+
+/** Wraps each tab label */
+.md-tab-label {
+  line-height: $md-tab-bar-height;
+  height: $md-tab-bar-height;
+  padding: 0 12px;
+  font-size: $md-body-font-size-base;
+  font-family: $md-font-family;
+  font-weight: 500;
+  cursor: pointer;
+  box-sizing: border-box;
+  color: currentColor;
+  opacity: 0.6;
+  min-width: 160px;
+  text-align: center;
+  &:focus {
+    outline: none;
+    opacity: 1;
+    background-color: md-color($md-primary, 100, 0.3);
+  }
+}
+
+/** The bottom section of the view; contains the tab bodies */
+.md-tab-body-wrapper {
+  position: relative;
+  height: 200px;
+  overflow: hidden;
+  padding: 12px;
+}
+
+/** Wraps each tab body */
+.md-tab-body {
+  display: none;
+  &.md-active {
+    display: block;
+  }
+}
+
+/** The colored bar that underlines the active tab */
+md-ink-bar {
+  position: absolute;
+  bottom: 0;
+  height: 2px;
+  background-color: md-color($md-primary, 500);
+  transition: 0.35s ease-out;
+}
--- a/src/components/tab-group/tab-group.spec.ts
+++ b/src/components/tab-group/tab-group.spec.ts
@@ -1,0 +1,132 @@
+import {
+    it,
+    expect,
+    beforeEach,
+    inject,
+    describe,
+    async
+} from '@angular/core/testing';
+import {TestComponentBuilder, ComponentFixture} from '@angular/compiler/testing';
+import {MD_TAB_GROUP_DIRECTIVES, MdTabGroup} from './tab-group';
+import {Component} from '@angular/core';
+import {By} from '@angular/platform-browser';
+
+describe('MdTabGroup', () => {
+  let builder: TestComponentBuilder;
+  let fixture: ComponentFixture<SimpleTabsTestApp>;
+
+  beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+    builder = tcb;
+  }));
+
+  describe('basic behavior', () => {
+    beforeEach(async(() => {
+      builder.createAsync(SimpleTabsTestApp).then(f => {
+        fixture = f;
+      });
+    }));
+
+    it('should default to the first tab', () => {
+      checkSelectedIndex(1);
+    });
+
+    it('should change selected index on click', () => {
+      let component = fixture.debugElement.componentInstance;
+      component.selectedIndex = 0;
+      checkSelectedIndex(0);
+
+      // select the second tab
+      let tabLabel = fixture.debugElement.query(By.css('.md-tab-label:nth-of-type(2)'));
+      tabLabel.nativeElement.click();
+      checkSelectedIndex(1);
+
+      // select the third tab
+      tabLabel = fixture.debugElement.query(By.css('.md-tab-label:nth-of-type(3)'));
+      tabLabel.nativeElement.click();
+      checkSelectedIndex(2);
+    });
+
+    it('should cycle through tab focus with focusNextTab/focusPreviousTab functions', () => {
+      let tabComponent = fixture.debugElement.query(By.css('md-tab-group')).componentInstance;
+      tabComponent.focusIndex = 0;
+      fixture.detectChanges();
+      expect(tabComponent.focusIndex).toBe(0);
+
+      tabComponent.focusNextTab();
+      fixture.detectChanges();
+      expect(tabComponent.focusIndex).toBe(1);
+
+      tabComponent.focusNextTab();
+      fixture.detectChanges();
+      expect(tabComponent.focusIndex).toBe(2);
+
+      tabComponent.focusNextTab();
+      fixture.detectChanges();
+      expect(tabComponent.focusIndex).toBe(2); // should stop at 2
+
+      tabComponent.focusPreviousTab();
+      fixture.detectChanges();
+      expect(tabComponent.focusIndex).toBe(1);
+
+      tabComponent.focusPreviousTab();
+      fixture.detectChanges();
+      expect(tabComponent.focusIndex).toBe(0);
+
+      tabComponent.focusPreviousTab();
+      fixture.detectChanges();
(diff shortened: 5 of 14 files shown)


src/components/tab-group/tab-group.spec.ts
```diff
@@ -0,0 +1,123 @@
+import {
+    it, describe, expect, beforeEach, inject, TestComponentBuilder, ComponentFixture,
+    fakeAsync, tick, ddescribe, iit, flushMicrotasks
+} from 'angular2/testing';
+import {MD_TAB_GROUP_DIRECTIVES, MdTabGroup} from "./tab-group";
+import {Component} from "angular2/core";
+import {By} from 'angular2/platform/browser';
+
+export function main() {
+  ddescribe('MdTabGroup', () => {
```
kara: Looks like you forgot a `ddescribe` here :)

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,65 @@
...
+  constructor() {
+    this._groupId = nextId++;
+  }
+
+  /**
+   * Returns a unique id for each tab label element
+   * @param i
+   * @returns {string}
+   */
+  getTabLabelId(i: number) {
```
kara: Put the return type on the function itself?

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,65 @@
...
+    return 'md-tab-label-' + this._groupId + '-' + i;
```
kara: It seems like this would be more readable with a template string, what do you think?

```typescript
return `md-tab-label-${this._groupId}-${i}`;
```

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,65 @@
...
+   * @returns {string}
+   */
+  getTabContentId(i: number) {
+    return 'md-tab-content-' + this._groupId + '-' + i;
+  }
+
+  /**
+   * Increment the selected index by 1; prevent going over the number of tabs
+   */
+  nextTab() {
```
kara: These `nextTab` and `previousTab` functions look similar to the methods in `TabBar`.  Any way we can re-use?

src/components/tab-group/tab-group.scss
```diff
@@ -0,0 +1,61 @@
+@import 'variables';
+@import 'default-theme';
+
+:host {
+  display: block;
+  font-family: sans-serif;
```
kara: Seems like maybe this should use the common `$font-family` variable? Or should this font be different?

src/components/tab-group/tab-group.scss
```diff
@@ -0,0 +1,61 @@
...
+    .md-tab-label.md-active {
+      background-color: md-color($md-primary, 100, 0.2);
+    }
+  }
+}
+
+.md-tab-label {
+  float: left;
+  line-height: 48px;
+  height: 48px;
```
kara: Have you considered putting this into a Sass variable?  Might help make it more customizable.

src/components/tab-group/tab-group.ts
robertmesserle: These will most-likely be reused when tab-bar is implemented, but this isn't in this iteration.  I only left this file in because Jeremy said to leave it in.

robertmesserle:
@kara Addressed your comments.

@jelbourn This might actually be okay to merge this week - I got it working with focus state and modified the arrow functionality so that it cycles focus target.  Enter will actually select the tab.

src/components/tab-group/tab-group.html
```diff
@@ -0,0 +1,25 @@
+<div class="md-tab-header" role="tablist"
+     (keydown.arrowRight)="focusNextTab()"
+     (keydown.arrowLeft)="focusPreviousTab()"
+     (keydown.enter)="selectedIndex = focusIndex">
+  <div class="md-tab-label" role="tab" md-tab-label-wrapper
+       *ngFor="#label of labels; #i = index"
```
princemaple: let

src/components/tab-group/tab-group.html
```diff
@@ -0,0 +1,25 @@
...
+       [attr.tabindex]="selectedIndex == i ? 0 : -1"
+       [class.md-active]="selectedIndex == i"
+       (click)="focusIndex = selectedIndex = i">
+    <template [portalHost]="label"></template>
+  </div>
+  <md-ink-bar [alignToElement]="currentLabelWrapper"></md-ink-bar>
+</div>
+<div class="md-tab-body-wrapper">
+  <div class="md-tab-body"
+       *ngFor="#content of contents; #i = index"
```
princemaple: let - ng2 beta 17

jelbourn:
@robertmesserle Now needs rebasing on my mega change #384

src/components/tab-group/tab-group.html
```diff
@@ -0,0 +1,25 @@
...
+       (click)="focusIndex = selectedIndex = i">
+    <template [portalHost]="label"></template>
+  </div>
+  <md-ink-bar [alignToElement]="currentLabelWrapper"></md-ink-bar>
+</div>
+<div class="md-tab-body-wrapper">
+  <div class="md-tab-body"
+       *ngFor="let content of contents; let i = index"
+       [class.md-active]="selectedIndex == i"
+       [attr.id]="getTabContentId(i)"
```
jelbourn: Just `id` should work.

src/components/tab-group/ink-bar.ts
```diff
@@ -0,0 +1,23 @@
+import {Directive, Renderer, ElementRef} from '@angular/core';
+
+@Directive({
```
jelbourn: Add class description explaining what the ink bar is and how it's used.

jelbourn: You probably also want to add an `@internal` JsDoc; this makes TypeScript ignore it when generating the `.d.ts.`. For example:

```ts
/**
 * This is the description of the ...
 * @internal
 */
```

src/components/tab-group/ink-bar.ts
```diff
@@ -0,0 +1,23 @@
...
+  selector: 'md-ink-bar',
+})
+export class MdInkBar {
+  constructor(public renderer: Renderer, public elementRef: ElementRef) {}
```
jelbourn: These fields should probably be `private`

src/components/tab-group/ink-bar.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  alignToElement(element: HTMLElement) {
+    this.renderer.setElementStyle(this.elementRef.nativeElement, 'left',
+        this.getLeftPosition(element));
+    this.renderer.setElementStyle(this.elementRef.nativeElement, 'width',
+        this.getElementWidth(element));
+  }
+
+  getLeftPosition(element: HTMLElement): string {
+    return element.offsetLeft + 'px';
```
jelbourn: Does `ElementRef` actually have `offsetLeft` and `offsetWidth` properties? I think you actually need to pull them through the renderer.

src/components/tab-group/tab-content.ts
```diff
@@ -0,0 +1,11 @@
+import {Directive, TemplateRef, ViewContainerRef} from '@angular/core';
+import {TemplatePortalDirective} from '../../core/portal/portal-directives';
+
+@Directive({
```
jelbourn: Add class description (here and for other classes throughout the PR)

src/components/tab-group/tab-group.html
```diff
@@ -0,0 +1,25 @@
+<div class="md-tab-header" role="tablist"
+     (keydown.arrowRight)="focusNextTab()"
+     (keydown.arrowLeft)="focusPreviousTab()"
+     (keydown.enter)="selectedIndex = focusIndex">
+  <div class="md-tab-label" role="tab" md-tab-label-wrapper
+       *ngFor="let label of labels; let i = index"
+       [id]="getTabLabelId(i)"
+       [attr.aria-controls]="getTabContentId(i)"
+       [attr.aria-selected]="selectedIndex == i"
+       [attr.tabindex]="selectedIndex == i ? 0 : -1"
```
jelbourn: You should be able to bind `tabIndex` without `attr.`

src/components/tab-group/tab-group.scss
```diff
@@ -0,0 +1,66 @@
+@import 'variables';
+@import 'default-theme';
+
+$tab-bar-height: 48px !default;
```
jelbourn: Prefix scss vars with `md-`

src/components/tab-group/tab-group.scss
```diff
@@ -0,0 +1,66 @@
...
+
+:host {
+  display: block;
+  font-family: $md-font-family;
+  .md-tab-header {
+    border-bottom: 1px solid md-color($md-background, status-bar);
+  }
+}
+
+.md-tab-header {
```
jelbourn: For each of the css classes here, could you add a brief overview of what they're for or which element/state they correspond to?

src/components/tab-group/tab-group.scss
```diff
@@ -0,0 +1,66 @@
...
+  overflow: hidden;
+  position: relative;
+  display: flex;
+  flex-direction: row;
+  &:before, &:after {
```
jelbourn: Add comment explaining what the `:before` and `:after` are doing here?

src/components/tab-group/tab-group.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      </md-tab>
+      <md-tab>
+        <template md-tab-label>Tab Three</template>
+        <template md-tab-content>Tab three content</template>
+      </md-tab>
+    </md-tab-group>
+  `,
+  directives: [MD_TAB_GROUP_DIRECTIVES]
+})
+class TestApp {
```
jelbourn: I'd be more specific with this name, like `TabsTestApp`

src/components/tab-group/tab-group.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  it('should change selected index on click', (done: () => void) => {
+    return builder.createAsync(TestApp).then(fixture => {
+      let component = fixture.debugElement.componentInstance;
+      component.selectedIndex = 0;
+      checkSelectedIndex(fixture, 0);
+
+      // select the second tab
+      let tabLabel = fixture.debugElement.query(By.css('.md-tab-label:nth-of-type(2)'));
```
jelbourn: It will probably be much cleaner to do the `createAsync` in a beforeEach, and then capture all of the tab labels in an array and access them directly in each test. I'm doing in in my in-progress rewrite of the radio button tests, I can show you if you're curious.

src/components/tab-group/tab-group.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+import {By} from '@angular/platform-browser';
+
+describe('MdTabGroup', () => {
+  let builder: TestComponentBuilder;
+
+  beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
+    builder = tcb;
+  }));
+
+  it('should default to the first tab', (done: () => void) => {
```
jelbourn: Rather than passing in `done` and calling it manually, you can now use `async` when dealing with `createAsync`, e.g.:
```ts
it('should do something', async(() => {
  builder.createAsync(TestApp).then(fixture => {
    expect(1).toBe(1);
  });
}));
```
`async()` hooks into zones and makes it so that the test doesn't "complete" until all of the microtasks have been executed.

src/components/tab-group/tab-group.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      component.selectedIndex = 2;
+      checkSelectedIndex(fixture, 2);
+
+      done();
+    });
+  });
+});
+
+function checkSelectedIndex(fixture: ComponentFixture<TestApp>, index: number) {
```
jelbourn: Add function description

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+import {ContentChildren} from '@angular/core';
+import {PortalHostDirective} from '../../core/portal/portal-directives';
+import {MdTabLabel} from './tab-label';
+import {MdTabContent} from './tab-content';
+import {MdTabLabelWrapper} from './tab-label-wrapper';
+import {MdInkBar} from './ink-bar';
+
+/**
+ * Used to generate unique ID's for each tab component
+ * @type {number}
```
jelbourn: Don't need `@type`, it's in the language now

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+  selector: 'md-tab-group',
+  templateUrl: './components/tab-group/tab-group.html',
+  styleUrls: ['./components/tab-group/tab-group.css'],
+  directives: [PortalHostDirective, MdTabLabelWrapper, MdInkBar],
+})
+export class MdTabGroup {
+  @ContentChildren(MdTabLabel) labels: QueryList<MdTabLabel>;
+  @ContentChildren(MdTabContent) contents: QueryList<MdTabContent>;
+  @ViewChildren(MdTabLabelWrapper) labelWrappers: QueryList<MdTabLabelWrapper>;
+  @ViewChildren(MdInkBar) inkBar: QueryList<MdInkBar>;
```
jelbourn: Each of these `__Children` properties should probably be `private`

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+  constructor(private _zone: NgZone) {
+    this._groupId = nextId++;
+  }
+
+  /**
+   * Waits one frame for the view to update, then upates the ink bar
+   * Note: This must be run outside of the zone or it will create an infinite change detection loop
+   */
+  ngAfterViewChecked(): void {
+    this._zone.runOutsideAngular(() => {
```
jelbourn: Add explanation of why you're doing the `runOutsideAngular`

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+      window.requestAnimationFrame(() => {
```
jelbourn: It should work to just say `requestAnimationFrame` without explicitly dereferencing `window`.

src/components/tab-group/tab-group.ts
jelbourn: You can use `@ViewChild` instead of `@ViewChildren` to just get a single instance.
```ts
 @ViewChild(MdInkBar) inkBar: MdInkBar;
```

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+  updateInkBar(): void {
+    this.inkBar.toArray()[0].alignToElement(this.currentLabelWrapper);
+  }
+
+  /**
+   * Reference to the current label wrapper; defaults to null for initial render before the
+   * ViewChildren references are ready.
+   * @returns {HTMLElement | null}
+   */
+  get currentLabelWrapper(): HTMLElement {
```
jelbourn: I think this should just be a function rather than a getter property.

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+    return this.labelWrappers
+        ? this.labelWrappers.toArray()[this.selectedIndex].elementRef.nativeElement
+        : null;
+  }
+
+  /**
+   * Tracks which element has focus; used for keyboard navigation
+   * @returns {number}
```
jelbourn: You don't need the type in the JsDoc

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+   * @returns {HTMLElement | null}
+   */
+  get currentLabelWrapper(): HTMLElement {
+    return this.labelWrappers
+        ? this.labelWrappers.toArray()[this.selectedIndex].elementRef.nativeElement
+        : null;
+  }
+
+  /**
+   * Tracks which element has focus; used for keyboard navigation
```
jelbourn:
```ts
/** Gets the index of the tab that currently has focus. */
```
?

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+      this.labelWrappers.toArray()[value].focus();
+    }
+  }
+
+  /**
+   * Returns a unique id for each tab label element
+   * @param i
+   * @returns {string}
+   */
+  getTabLabelId(i: number): string {
```
jelbourn: Could the labels and contents generate their own IDs?

src/components/tab-group/tab-group.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (this.labelWrappers && this.focusIndex < this.labelWrappers.length - 1) {
+      this.focusIndex++;
+    }
+  }
+
+  /**
+   * Decrement the focus index by 1; prevent going below 0
+   */
+  focusPreviousTab(): void {
+    if (this.focusIndex > 0) {
```
jelbourn: optional: could do
```ts
this.focusIndex = Math.max(0, this.focusIndex - 1);
```

src/components/tab-group/tab-group.scss
```diff
@@ -0,0 +1,66 @@
+@import 'variables';
+@import 'default-theme';
+
+$md-tab-bar-height: 48px !default;
+
+:host {
+  display: block;
+  font-family: $md-font-family;
+  .md-tab-header {
+    border-bottom: 1px solid md-color($md-background, status-bar);
```
jelbourn: Why is this one rule for `md-tab-header` here instead of in the rule below?

jelbourn:
Looks good aside from that one last comment and a needed rebase.

jelbourn:
LGTM

PR #378, opened 2016-04-30
fix: update angular to beta.17

jelbourn:
R: @hansl @kara

### DIFF
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -145,7 +145,7 @@
         <button (click)="addABunch(1000)">Add 1000</button>
       </td>
     </thead>
-    <tr *ngFor="#item of items; #i = index">
+    <tr *ngFor="let item of items; let i = index">
       <td>{{i+1}}</td>
       <td>
         <md-input type="number" placeholder="value" aria-label="hello" [(ngModel)]="items[i].value"></md-input>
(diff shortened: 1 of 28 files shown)


src/components/sidenav/sidenav.spec.ts
```diff
@@ -77,6 +77,7 @@ export function main() {
         xhr.get('./components/sidenav/sidenav.css').then((css) => { style = css; })
       ]).catch((err: any) => {
         console.error(err);
+        console.error(err);
```
kara: Duplication by mistake?

src/demo-app/input/input-demo.html
```diff
@@ -145,7 +145,7 @@
         <button (click)="addABunch(1000)">Add 1000</button>
       </td>
     </thead>
-    <tr *ngFor="#item of items; #i = index">
+    <tr *ngFor="let item of items; #i = index">
```
kara: Looks like you need to replace the `#i` here with `let i`?

src/components/sidenav/sidenav.spec.ts
jelbourn: Fixed.

src/demo-app/input/input-demo.html
jelbourn: Fixed.

Issue #72, opened 2016-02-11
Core gesture handling module


jelbourn:
- [x] Research existing gesture libraries (polymer, hammer, etc.)
- [x] Propose approach
- [x] Define gesture vocabulary (swipe, pan, tap, drag, etc.)
- [x] Solidify path forward with hammer (alpha.2) (https://github.com/hammerjs/hammer.js/issues/917)
- [x] Support for devices with touch and mouse (alpha.3)
- [x] Create / modify Angular event plugin on top of module (alpha.3)
- [ ] Ensure "drag" recognizer doesn't overwrite "pan"

jelbourn:
Research is done. Moving issue to next milestone for execution.

jelbourn:
Moving to alpha.3 for completion of final parts

jelbourn:
@kara closing this as done.

PR #382, opened 2016-05-01
fix(gestures): ensure default gestures are not overwritten by custom gestures

kara:
r: @jelbourn 

This PR ensures that our custom gestures don't interfere with the default Hammer gestures.  In other words, "drag" won't overwrite "pan", so users can still use "pan" in their own code if they like.

See https://github.com/angular/material2/issues/72

### DIFF
--- a/src/core/gestures/MdGestureConfig.ts
+++ b/src/core/gestures/MdGestureConfig.ts
@@ -5,22 +5,46 @@ import {HammerGestureConfig} from '@angular/platform-browser';
 @Injectable()
 export class MdGestureConfig extends HammerGestureConfig {
   /* List of new event names to add to the gesture support list */
-  events: string[] = ['drag', 'longpress'];
+  events: string[] = [
+    'drag',
+    'dragstart',
+    'dragend',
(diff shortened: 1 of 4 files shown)


src/core/gestures/MdGestureConfig.ts
```diff
@@ -5,22 +5,46 @@ import {HammerGestureConfig} from 'angular2/src/platform/browser_common';
...
+   * https://www.google.com/design/spec/patterns/gestures.html#gestures-touch-mechanics
+   *
+   * More information on default recognizers can be found in Hammer docs:
+   * http://hammerjs.github.io/recognizer-pan/
+   * http://hammerjs.github.io/recognizer-press/
+   *
+   * TODO: Confirm threshold numbers with Material Design UX Team
+   * */
+  buildHammer(element: HTMLElement) {
+    var mc = new Hammer(element);
```
soyzamudio: is there a reason you're using `var` instead of `let`? Just wondering because I've seen `let` being used in other places

jelbourn:
LGTM

Issue #388, opened 2016-05-04
input's maxLength=-1 property not works in MS Edge


davidgabrichidze:
When attribute 'maxLength' is removed from input, behavior is corrected.

changing to `@Input() maxLength: number = null;` in input.ts removes attribute.

angular: rc-0;
material2: alpha4;
Edge  version: 25.10

sebakerckhof:
Duplicate of https://github.com/angular/material2/issues/336

Issue #342, opened 2016-04-23
Firefox md-input placeholder text doesn't appear above user input


dozingcat:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
Run the demo app in Firefox, or go to https://material2-app.firebaseapp.com/ . Click on a text field with placeholder text. The placeholder text will turn blue but remain in the same place. Start typing and the placeholder text will disappear entirely.

* **What is the expected behavior?**
Chrome behaves correctly; when the text field becomes focused, the placeholder text should animate above the entered text.

* **Which version of Angular and Material, and which browser and OS does this issue affect?**
Behavior is identical in Firefox 45 on Mac OS X 10.11, Windows 10, and Android. Running against material2 head.

* **Other information**
Commenting out the "input:-webkit-autofill + .md-input-placeholder" selector in input.scss (https://github.com/angular/material2/blob/master/src/components/input/input.scss#L140) makes Firefox work as expected. Apparently Firefox believes "input:-webkit-autofill" is invalid CSS and won't apply any selectors that contain it.

davidgabrichidze:
One option is to split webkit-autofill and other selectors from each other. This will duplicate several lines of css, but will work for firefox, without affectine chrome/safary autofill behavior.

e.g. rendered css which works as expected
`[_nghost-dxv-12] .md-input-placeholder.md-float:not(.md-empty), [_nghost-dxv-12] .md-input-placeholder.md-float.md-focused, [_nghost-dxv-12] .md-input-placeholder {
  visibility: visible;
  padding-bottom: 5px;
  -webkit-transform: translateY(-100%) scale(0.75);
          transform: translateY(-100%) scale(0.75); }
  [_nghost-dxv-12] .md-input-placeholder.md-float:not(.md-empty) .md-placeholder-required, [_nghost-dxv-12] .md-input-placeholder.md-float.md-focused .md-placeholder-required, [_nghost-dxv-12] .md-input-placeholder .md-placeholder-required {
    color: #9c27b0; }

[_nghost-dxv-12] input:-webkit-autofill + .md-input-placeholder {
  visibility: visible;
  padding-bottom: 5px;
  -webkit-transform: translateY(-100%) scale(0.75);
          transform: translateY(-100%) scale(0.75); }
  [_nghost-dxv-12] input:-webkit-autofill + .md-input-placeholder .md-placeholder-required {
    color: #9c27b0; }`

PR #393, opened 2016-05-05
Fix MS Edge and Firefox issues #342, #388

davidgabrichidze:
Fix md-input maxLength problem for MS Edge #388 
Fix floating placeholder in Firefox, without affecting autofill behavior in Chrome/Safari #342


davidgabrichidze:
I've signed it 2 days ago, where is my cla:yes :))))

DevVersion:
@davidgabrichidze This is caused by your multiple commits. Some commits are not committed properly with your GitHub Account Email.

I suggest, squashing all commits into one, then the CLA check will pass properly.

davidgabrichidze:
@DevVersion thanks for advise. So, I should close this PR and request another one, with joined commits?

DevVersion:
@davidgabrichidze That shouldn't be necessary. You can just squash the commits on this PR and force push the changes.

Here is a useful Git Alias for Squashing
```
  squash = "!f(){ git reset --soft HEAD~${1} && git commit --edit -m\"$(git log --format=%B --reverse HEAD..HEAD@{1})\"; };f"

```

And here is another link for squashing: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

davidgabrichidze:
@DevVersion Thanks, it worked :+1:

src/components/input/input.scss
```diff
@@ -19,6 +19,12 @@ $md-input-underline-focused-color: md-color($md-primary);
 $md-input-underline-disabled-background-image: linear-gradient(to right,
         rgba(0,0,0,0.26) 0%, rgba(0,0,0,0.26) 33%, transparent 0%);
 
+/**
+ * Removes firefox box-shadow on required fields...
+ */
+:-moz-ui-invalid {
+    box-shadow: none;
```
jelbourn: -2 indent

src/components/input/input.scss
```diff
@@ -19,6 +19,12 @@ $md-input-underline-focused-color: md-color($md-primary);
 $md-input-underline-disabled-background-image: linear-gradient(to right,
         rgba(0,0,0,0.26) 0%, rgba(0,0,0,0.26) 33%, transparent 0%);
 
+/**
+ * Removes firefox box-shadow on required fields...
```
jelbourn: Comment can be something like
```ts
/** 
 * Undo the red box-shadow glow added by Firefox on invalid inputs.
 * See https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-ui-invalid
 */
```

src/components/input/input.scss
```diff
@@ -33,6 +39,15 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
...
 
+/**
+ * Applies a floating placeholder above the input before chrome's and safari's auto-fill commits.
+ * Firefox has selector issues for :-webkit-autofill, thats why this mixin is duplicated and is used seperately below 
+ */
+%md-input-placeholder-floating-autofill {
+  visibility: visible;
+  padding-bottom: 5px;
+  transform: translateY(-100%) scale(0.75);
+}
```
jelbourn: I don't follow the reasoning for why you're duplicating these rules.

davidgabrichidze: @jelbourn that helped to split generated css

src/components/input/input.scss
davidgabrichidze: Done

src/components/input/input.scss
davidgabrichidze: Done

src/components/input/input.scss
davidgabrichidze: @jelbourn what about this?

jelbourn: You can just make this a mixin (rather than a placeholder) and use it in both places rather than duplicating it.

davidgabrichidze: @jelbourn I'm not sure that I clearly understand your note. If you suggest to make **one **mixin, than it means to **rollback ** change, because it **WAS** one mixin which doesn't work. **Why?** It generates joined css, something like this: `.md-input-placeholder.md-float:not(.md-empty), .md-input-placeholder.md-float.md-focused, input:-webkit-autofill + .md-input-placeholder` which is not selected by FF. Why? **input:-webkit-autofill** css is invalid for FF and it ignores all joined styles.

How to improve? Well, if we split it in two different groups of css `input:-webkit-autofill` and others, than it works.

Only way i found to create these two groups in rendered css is to make another mixin. That's why it is 'duplicated'. May be there is some other option to force scss compiler render different groups in css when we use 'extend' of mixins?

jelbourn: These definitions are scss **placeholders**, not mixins, which is the problem. See 
http://thesassway.com/intermediate/understanding-placeholder-selectors

If you change to a single _mixin_, it should be fine.

davidgabrichidze: Oh, now it's clear :) - will update soon

davidgabrichidze: Done

jelbourn:
LGTM

PR #400, opened 2016-05-07
feat(input): add ability to hide text divider

soyzamudio:
@hansl @kara @jelbourn 

Adds the ability to show or hide the underline beneath the input (just like with the floating placeholder)

screen shot 2016-05-06 at 10 40 37 pm

### DIFF
--- a/src/components/input/README.md
+++ b/src/components/input/README.md
@@ -15,6 +15,18 @@ At the time of writing this README, the `[type]` attribute is copied to the actu
 
 The valid `type` attribute values are any supported by your browser, with the exception of `file`, `checkbox` and `radio`. File inputs aren't supported for now, while check boxes and radio buttons have their own components.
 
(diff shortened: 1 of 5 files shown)


soyzamudio:
I signed it!

peterblazejewicz:
That's called text divider, not underline. 
Should it be allowed to hide per docs (just wondering)? 
https://www.google.com/design/spec/components/text-fields.html#text-fields-input

jelbourn:
@soyzamudio I chatted with @hansl about this and we agreed that this is something that would be better done with CSS than a new directive API. We want people to be able to do something like
```scss
md-input-underline {
  display: none;
}
```
To hide this part of the component. Right now the view encapsulation makes this difficult, so we're going to change the component to `ViewEncapsulation.None` and document the approach once we have our more complete documentation.

PR #405, opened 2016-05-09
feat(grid-list): add grid-list component

kara:
r: @jelbourn 

In this PR:
- [x] Ported Jeremy's existing code from core
- [x] Removed Dart facades and Dart-specific patterns
- [x] Refactored to use new constructs like @ContentChildren and removed deprecated syntax
- [x] Split out TileStyler into its own class
- [x] Various bug fixes (e.g. hostbinding issue, rowspan edge case, etc)
- [x] Added usability improvements
   - [x] Friendly error when cols property is missing
   - [x] Default rowHeight to 1:1 instead of silent fail
   - [x] Use px by default if no units instead of silent fail
   - [x] Set height automatically for fixed/ratio mode 

Still TODO (next PR):
-  Unit tests
-  Docs
-  Responsive layout
-  Grid tile header and footer
-  Sass review/refactor

### DIFF
--- a/src/components/grid-list/grid-list.scss
+++ b/src/components/grid-list/grid-list.scss
@@ -1,0 +1,67 @@
+// TODO(kara): Review this to see if MD spec updates are needed
+
+md-grid-list {
+  display: block;
+  position: relative;
+}
+
+md-grid-tile {
+  display: block;
+  position: absolute;
+
+  figure {
+    display: flex;
+    position: absolute;
+
+    align-items: center;
+    justify-content: center;
+    height: 100%;
+
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+
+    padding: 0;
+    margin: 0;
+  }
+
+  // Headers & footers
+  md-grid-tile-header,
+  md-grid-tile-footer {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    height: 48px;
+    color: #fff;
+    background: rgba(0, 0, 0, 0.18);
+    overflow: hidden;
+
+    // Positioning
+    position: absolute;
+    left: 0;
+    right: 0;
+
+    h3,
+    h4 {
+      font-weight: 400;
+      margin: 0 0 0 16px;
+    }
+
+    h3 {
+      font-size: 14px;
+    }
+
+    h4 {
+      font-size: 12px;
+    }
+  }
+
+  md-grid-tile-header {
+    top: 0;
+  }
+
+  md-grid-tile-footer {
+    bottom: 0;
+  }
+}
--- a/src/components/grid-list/grid-list.ts
+++ b/src/components/grid-list/grid-list.ts
@@ -1,8 +1,158 @@
-import {Component} from '@angular/core';
+import {
+  Component,
+  ViewEncapsulation,
+  AfterContentChecked,
+  OnInit,
+  Input,
+  ContentChildren,
+  QueryList,
+  Renderer,
+  ElementRef
+} from '@angular/core';
+import {MdGridTile} from './grid-tile';
+import {TileCoordinator} from './tile-coordinator';
+import {
+    TileStyler,
+    FitTileStyler,
+    RatioTileStyler,
+    FixedTileStyler
+} from './tile-styler';
+import {MdGridListColsError} from './grid-list-errors';
+import {Dir} from '../../core/rtl/dir';
+
+// TODO(kara): Conditional (responsive) column count / row size.
+// TODO(kara): Re-layout on window resize / media change (debounced).
+// TODO(kara): gridTileHeader and gridTileFooter.
+
+const MD_FIT_MODE = 'fit';
 
 @Component({
   selector: 'md-grid-list',
+  host: { 'role': 'list' },
   templateUrl: './components/grid-list/grid-list.html',
   styleUrls: ['./components/grid-list/grid-list.css'],
+  encapsulation: ViewEncapsulation.None,
 })
-export class MdGridList {}
+export class MdGridList implements OnInit, AfterContentChecked {
+  /** Number of columns being rendered. */
+  _cols: number;
+
+  /** Row height value passed in by user. This can be one of three types:
+   * - Number value (ex: "100px"):  sets a fixed row height to that value
+   * - Ratio value (ex: "4:3"): sets the row height based on width:height ratio
+   * - "Fit" mode (ex: "fit"): sets the row height to total height divided by number of rows
+   * */
+  _rowHeight: string;
+
+  /** The amount of space between tiles. This will be something like '5px' or '2em'. */
+  _gutter: string = '1px';
+
+  /** Sets position and size styles for a tile */
+  _tileStyler: TileStyler;
+
+  /** Query list of tiles that are being rendered. */
+  @ContentChildren(MdGridTile) _tiles: QueryList<MdGridTile>;
+
+  constructor(private _renderer: Renderer, private _element: ElementRef,
+              private _dir: Dir) {}
+
+  @Input()
+  get cols() {
+    return this._cols;
+  }
+
+  set cols(value: any) {
+    this._cols = coerceToNumber(value);
+  }
+
+  @Input('gutterSize')
+  get gutterSize() {
+    return this._gutter;
+  }
+
+  set gutterSize(value: any) {
+    this._gutter = coerceToString(value);
+  }
+
+  /** Set internal representation of row height from the user-provided value. */
+  @Input()
+  set rowHeight(value: string | number) {
+    this._rowHeight = coerceToString(value);
+    this._setTileStyler();
+  }
+
+  ngOnInit() {
+    this._checkCols();
+    this._checkRowHeight();
+  }
+
+  /** The layout calculation is fairly cheap if nothing changes, so there's little cost
+   * to run it frequently. */
+  ngAfterContentChecked() {
+    this._layoutTiles();
+  }
+
+  /** Throw a friendly error if cols property is missing */
+  private _checkCols() {
+    if (!this.cols) {
+      throw new MdGridListColsError();
+    }
+  }
+
+  /** Default to equal width:height if rowHeight property is missing */
+  private _checkRowHeight(): void {
+    if (!this._rowHeight) {
+      this._tileStyler = new RatioTileStyler('1:1');
+    }
+  }
+
+  /** Creates correct Tile Styler subtype based on rowHeight passed in by user */
+  private _setTileStyler(): void {
+    if (this._rowHeight === MD_FIT_MODE) {
+      this._tileStyler = new FitTileStyler();
+    } else if (this._rowHeight && this._rowHeight.match(/:/g)) {
+      this._tileStyler = new RatioTileStyler(this._rowHeight);
+    } else {
+      this._tileStyler = new FixedTileStyler(this._rowHeight);
+    }
+  }
+
+  /** Computes and applies the size and position for all children grid tiles. */
+  private _layoutTiles(): void {
+    let tiles = this._tiles.toArray();
+    let tracker = new TileCoordinator(this.cols, tiles);
+    this._tileStyler.init(this.gutterSize, tracker, this.cols, this._dir);
+
+    for (let i = 0; i < tiles.length; i++) {
+      let pos = tracker.positions[i];
+      let tile = tiles[i];
+      this._tileStyler.setStyle(tile, pos.row, pos.col);
+    }
+    this.setListStyle(this._tileStyler.getComputedHeight());
+  }
+
+  /** Sets style on the main grid-list element, given the style name and value.
+   *  @internal
+   */
+  setListStyle(style: [string, string]): void {
+    if (style) {
+      this._renderer.setElementStyle(this._element.nativeElement, style[0], style[1]);
+    }
+  }
+}
+
+/** Converts values into strings. Falsy values become empty strings.
+ * @internal
+ */
+export function coerceToString(value: string | number): string {
+  return `${value || ''}`;
+}
+
+/** Converts a value that might be a string into a number.
+ * @internal
+ */
+export function coerceToNumber(value: string | number): number {
+  return typeof value === 'string' ? parseInt(value, 10) : value;
+}
+
+export const MD_GRID_LIST_DIRECTIVES: any[] = [MdGridList, MdGridTile];
--- a/src/components/grid-list/grid-tile.html
+++ b/src/components/grid-list/grid-tile.html
@@ -1,0 +1,4 @@
+<!-- TODO(kara): Revisit why this is a figure.-->
+<figure>
+  <ng-content></ng-content>
+</figure>
--- a/src/components/grid-list/grid-tile.ts
+++ b/src/components/grid-list/grid-tile.ts
@@ -1,0 +1,53 @@
+import {
+  Component,
+  ViewEncapsulation,
+  Renderer,
+  ElementRef,
+  Input,
+} from '@angular/core';
+
+import {coerceToNumber} from './grid-list';
+
+@Component({
+  selector: 'md-grid-tile',
+  host: { 'role': 'listitem' },
+  templateUrl: './components/grid-list/grid-tile.html',
+  styleUrls: ['./components/grid-list/grid-list.css'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class MdGridTile {
+  _rowspan: number = 1;
+  _colspan: number = 1;
+  _element: HTMLElement;
+
+  constructor(private _renderer: Renderer, element: ElementRef) {
+    this._element = element.nativeElement;
+  }
+
+  @Input()
+  get rowspan() {
+    return this._rowspan;
+  }
+
+  @Input()
+  get colspan() {
+    return this._colspan;
+  }
+
+  set rowspan(value) {
+    this._rowspan = coerceToNumber(value);
+  }
+
+  set colspan(value) {
+    this._colspan = coerceToNumber(value);
+  }
+
+  /** Sets the style of the grid-tile element.  Needs to be set manually to avoid
+   * "Changed after checked" errors that would occur with HostBinding.
+   * @internal
+   */
+  setStyle(property: string, value: string): void {
+    this._renderer.setElementStyle(this._element, property, value);
+  }
+
+}
--- a/src/components/grid-list/tile-coordinator.ts
+++ b/src/components/grid-list/tile-coordinator.ts
@@ -1,0 +1,142 @@
+import {MdGridTile} from './grid-tile';
+import {MdGridTileTooWideError} from './grid-list-errors';
+
+/**
+ * Class for determining, from a list of tiles, the (row, col) position of each of those tiles
+ * in the grid. This is necessary (rather than just rendering the tiles in normal document flow)
+ * because the tiles can have a rowspan.
+ *
+ * The positioning algorithm greedily places each tile as soon as it encounters a gap in the grid
+ * large enough to accommodate it so that the tiles still render in the same order in which they
+ * are given.
+ *
+ * The basis of the algorithm is the use of an array to track the already placed tiles. Each
+ * element of the array corresponds to a column, and the value indicates how many cells in that
+ * column are already occupied; zero indicates an empty cell. Moving "down" to the next row
+ * decrements each value in the tracking array (indicating that the column is one cell closer to
+ * being free).
+ */
+export class TileCoordinator {
+  /** Tracking array (see class description). */
+  tracker: number[];
+
+  /** Index at which the search for the next gap will start. */
+  columnIndex: number = 0;
+
+  /** The current row index. */
+  rowIndex: number = 0;
+
+  /** Gets the total number of rows occupied by tiles */
+  get rowCount(): number { return this.rowIndex + 1; }
+
+  /** Gets the total span of rows occupied by tiles.
+   * Ex: A list with 1 row that contains a tile with rowspan 2 will have a total rowspan of 2. */
+  get rowspan() {
+    let lastRowMax = Math.max(...this.tracker);
+    // if any of the tiles has a rowspan that pushes it beyond the total row count,
+    // add the difference to the rowcount
+    return lastRowMax > 1 ? this.rowCount + lastRowMax - 1 : this.rowCount;
+  }
+
+  /** The computed (row, col) position of each tile (the output). */
+  positions: TilePosition[];
+
+  constructor(numColumns: number, tiles: MdGridTile[]) {
+    this.tracker = new Array(numColumns);
+    this.tracker.fill(0, 0, this.tracker.length);
+
+    this.positions = tiles.map(tile => this._trackTile(tile));
+  }
+
+  /** Calculates the row and col position of a tile. */
+  private _trackTile(tile: MdGridTile): TilePosition {
+    // Find a gap large enough for this tile.
+    let gapStartIndex = this._findMatchingGap(tile.colspan);
+
+    // Place tile in the resulting gap.
+    this._markTilePosition(gapStartIndex, tile);
+
+    // The next time we look for a gap, the search will start at columnIndex, which should be
+    // immediately after the tile that has just been placed.
+    this.columnIndex = gapStartIndex + tile.colspan;
+
+    return new TilePosition(this.rowIndex, gapStartIndex);
+  }
+
+  /** Finds the next available space large enough to fit the tile. */
+  private _findMatchingGap(tileCols: number): number {
+    if (tileCols > this.tracker.length) {
+      throw new MdGridTileTooWideError(tileCols, this.tracker.length);
+    }
+
+    // Start index is inclusive, end index is exclusive.
+    let gapStartIndex = -1;
+    let gapEndIndex = -1;
+
+    // Look for a gap large enough to fit the given tile. Empty spaces are marked with a zero.
+    do {
+      // If we've reached the end of the row, go to the next row.
+      if (this.columnIndex + tileCols > this.tracker.length) {
+        this._nextRow();
+        continue;
+      }
+
+      gapStartIndex = this.tracker.indexOf(0, this.columnIndex);
+
+      // If there are no more empty spaces in this row at all, move on to the next row.
+      if (gapStartIndex == -1) {
+        this._nextRow();
+        continue;
+      }
+
+      gapEndIndex = this._findGapEndIndex(gapStartIndex);
+
+      // If a gap large enough isn't found, we want to start looking immediately after the current
+      // gap on the next iteration.
+      this.columnIndex = gapStartIndex + 1;
+
+      // Continue iterating until we find a gap wide enough for this tile.
+    } while (gapEndIndex - gapStartIndex < tileCols);
+    return gapStartIndex;
+  }
+
+  /** Move "down" to the next row. */
+  private _nextRow(): void {
+    this.columnIndex = 0;
+    this.rowIndex++;
+
+    // Decrement all spaces by one to reflect moving down one row.
+    for (let i = 0; i < this.tracker.length; i++) {
+      this.tracker[i] = Math.max(0, this.tracker[i] - 1);
+    }
+  }
+
+  /**
+   * Finds the end index (exclusive) of a gap given the index from which to start looking.
+   * The gap ends when a non-zero value is found.
+   */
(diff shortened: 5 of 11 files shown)


src/components/grid-list/grid-list.scss
```diff
@@ -0,0 +1,67 @@
...
+md-grid-list {
+  display: block;
+  position: relative;
+}
+
+md-grid-tile {
+  display: block;
+  position: absolute;
+
+  figure {
```
jelbourn: Can you do this as a class rule rather than `md-grid-tile figure` ?

kara: Noted. Will update in next PR when I do a Sass pass (for the class)... ha.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+    FixedTileStyler
+} from './tile-styler';
+import {Dir} from '../../core/rtl/dir';
+
+// TODO(kara): Conditional (responsive) column count / row size.
+// TODO(kara): Re-layout on window resize / media change (debounced).
+// TODO(kara): gridTileHeader and gridTileFooter.
+
+function ensureStringInput(value: string | number): string {
+  return typeof value === 'number' ? value.toString() : value;
```
jelbourn: Why not just 
```ts
return `${value}`;
```

jelbourn: Could also call it `stringify` and move it to a shared place.

kara: Yeah, that's what I had originally, but it doesn't work for when the value is null or undefined (you get "null" or "undefined").

jelbourn:
```ts
return `${value || ''}`;
```

kara: Done

josephperrott:
```ts
let value = 0;
return `${value || ''}`;
```
returns

```ts
''
```
rather than
```ts
0
```
as desired

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
   templateUrl: './components/grid-list/grid-list.html',
   styleUrls: ['./components/grid-list/grid-list.css'],
+  encapsulation: ViewEncapsulation.None
 })
-export class MdGridList {}
+export class MdGridList implements OnInit, AfterContentChecked {
+  /** Number of columns being rendered. */
+  _cols: number;
+
+  /** Row height value passed in by user */
```
jelbourn: Can you expand the string to mention what types of values are supported?

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+
+  /* Array of grid tiles being rendered (from query list) */
+  _tiles: MdGridTile[] = [];
+
+  constructor(private _renderer: Renderer, private _element: ElementRef,
+              private _dir: Dir) {}
+
+  get cols() { return this._cols; }
+
+  @Input('cols')
```
jelbourn: You don't need the `'cols'` string if it matches the property name. 

Also, I think we've generally been putting the `@Input` on the getter rather than the setter (it should be identical AFAIK).

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+  }
+
+  setListStyle(style: string[]): void {
+    if (style) {
+      this._renderer.setElementStyle(this._element.nativeElement, style[0], style[1]);
+    }
+  }
+}
+
+export function ensureStringInput(value: string | number): string {
```
jelbourn: The name of this function should be something like `coerceToString` or `makeString` or something like that. `ensure` sounds like it is doing some assertion.

jelbourn: Also add the `/** @internal */` marker.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+      this._tileStyler = new RatioTileStyler(this._rowHeight);
+    } else {
+      this._tileStyler = new FixedTileStyler(this._rowHeight);
+    }
+  }
+
+  /** Query list of tiles that are being rendered. */
+  @ContentChildren(MdGridTile)
+  set tiles(tiles: QueryList<MdGridTile>) {
+    this._tiles = tiles.toArray();
```
jelbourn: I don't think you want to do this; the benefit of `QueryList` is that it's live, reflecting the current state of `ngIf`, `ngFor`, etc. If you grab the array only once, you'll lose that.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+  /** Sets position and size styles for a tile */
+  _tileStyler: TileStyler;
+
+  /* Array of grid tiles being rendered (from query list) */
+  _tiles: MdGridTile[] = [];
+
+  constructor(private _renderer: Renderer, private _element: ElementRef,
+              private _dir: Dir) {}
+
+  get cols() { return this._cols; }
```
jelbourn: Personally, I really don't like having function bodies on the same line, even when they're short.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+  }
+
+  ngOnInit() { this._checkDefaults(); }
+
+  ngAfterContentChecked() { this.layoutTiles(); }
+
+  _checkDefaults(): void {
+    // if cols property is missing, throw a friendly error
+    if (!this.cols) {
+      throw `md-grid-list: must pass in number of columns. Example: <md-grid-list cols="3">`;
```
jelbourn: Can you create an `Error` class for this error? Similar to what Sidenav and Portal do.
(I think having a separate errors file is nice).

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+  @Input('gutterSize')
+  set gutterSize(value: any) {
+    this._gutter = ensureStringInput(value);
+  }
+
+  /** Set internal representation of row height from the user-provided value. */
+  @Input('rowHeight')
+  set rowHeight(value: string | number) {
+    this._rowHeight = ensureStringInput(value);
+    if (value === RowHeightMode.FIT) {
```
jelbourn: Can you break the setting of `_tileStyler` into its own function that you just call here?

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+  ngOnInit() { this._checkDefaults(); }
+
+  ngAfterContentChecked() { this.layoutTiles(); }
+
+  _checkDefaults(): void {
+    // if cols property is missing, throw a friendly error
+    if (!this.cols) {
+      throw `md-grid-list: must pass in number of columns. Example: <md-grid-list cols="3">`;
+    }
+    // if rowHeight property is missing, default to equal width:height
```
jelbourn: Can you expand this with the rationale we discussed for _why_ the 1:1 ratio makes the most sense as a default?

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+
+    for (let i = 0; i < this._tiles.length; i++) {
+      let pos = tracker.positions[i];
+      let tile = this._tiles[i];
+      this._tileStyler.setStyle(tile, pos.row, pos.col);
+    }
+    this.setListStyle(this._tileStyler.getComputedHeight());
+  }
+
+  setListStyle(style: string[]): void {
```
jelbourn: Rather than an array of strings, I think you've actually got a tuple type here.
https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple

jelbourn: Add could use a function description.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+  @ContentChildren(MdGridTile)
+  set tiles(tiles: QueryList<MdGridTile>) {
+    this._tiles = tiles.toArray();
+  }
+
+  ngOnInit() { this._checkDefaults(); }
+
+  ngAfterContentChecked() { this.layoutTiles(); }
+
+  _checkDefaults(): void {
```
jelbourn: Needs a function description

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+    // if cols property is missing, throw a friendly error
+    if (!this.cols) {
+      throw `md-grid-list: must pass in number of columns. Example: <md-grid-list cols="3">`;
+    }
+    // if rowHeight property is missing, default to equal width:height
+    if (!this._rowHeight) {
+      this._tileStyler = new RatioTileStyler('1:1');
+    }
```
jelbourn: I think it might be better to make assertions and set defaults in separate functions rather than combine the two.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+
+  /** Query list of tiles that are being rendered. */
+  @ContentChildren(MdGridTile)
+  set tiles(tiles: QueryList<MdGridTile>) {
+    this._tiles = tiles.toArray();
+  }
+
+  ngOnInit() { this._checkDefaults(); }
+
+  ngAfterContentChecked() { this.layoutTiles(); }
```
jelbourn: Add a comment talking about how the relayout is very cheap if nothing actually changes, so there's little cost in running it too frequently?

src/components/grid-list/grid-tile.html
```diff
@@ -0,0 +1,3 @@
+<figure>
```
jelbourn: Do you know why this uses a `<figure>` element? I actually have no idea.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,127 @@
...
+// TODO(kara): Conditional (responsive) column count / row size.
+// TODO(kara): Re-layout on window resize / media change (debounced).
+// TODO(kara): gridTileHeader and gridTileFooter.
 
 @Component({
   selector: 'md-grid-list',
+  host: { 'role': 'list' },
   templateUrl: './components/grid-list/grid-list.html',
   styleUrls: ['./components/grid-list/grid-list.css'],
+  encapsulation: ViewEncapsulation.None
```
jelbourn: This component config block is _pretty_ nice, but it would be even nicer with a trailing comma...

src/components/grid-list/tile-coordinator.ts
```diff
@@ -0,0 +1,136 @@
...
+  rowIndex: number = 0;
+
+  // Gets the total number of rows occupied by tiles
+  get rowCount(): number { return this.rowIndex + 1; }
+
+  // Gets the total span of rows occupied by tiles
+  // Ex: a list with 1 row that contains a tile with rowspan 2 will have a total rowspan of 2
+  get rowspan() {
+    let lastRowMax = Math.max(...this.tracker);
+    return lastRowMax > 1 ? this.rowCount + lastRowMax - 1 : this.rowCount;
```
jelbourn: I think it would be good to have a comment here explaining something like "If any of the tiles has a rowspan that pushes it beyond the total number of rows..."

src/components/grid-list/tile-coordinator.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  // The computed (row, col) position of each tile (the output).
+  positions: Position[];
+
+  constructor(numColumns: number, tiles: MdGridTile[]) {
+    this.tracker = new Array(numColumns);
+    this.tracker.fill(0, 0, this.tracker.length);
```
jelbourn: Do we have a polyfill loaded for `Array.prototype.fill`? It's not supported in IE11. We probably only want to rely on that polyfill if Angular itself already does so.

src/components/grid-list/grid-list.ts
kara: Fair point; not sure why I did that. Done!

src/components/grid-list/grid-list.ts
kara: Agreed; name changed. Marker, function description, and rainbow added.

src/components/grid-list/grid-list.ts
kara: Done!

src/components/grid-list/grid-list.ts
kara: I got in the habit based on angular core's clang, but I can see either way. Separated it out!

src/components/grid-list/grid-list.ts
kara: Oh, good call.  Broke it out.

src/components/grid-list/grid-list.ts
kara: Ah, that's right.  Added!

src/components/grid-list/grid-list.ts
kara: Should have known :p

src/components/grid-list/grid-list.ts
kara: Done

src/components/grid-list/grid-list.ts
kara: It's part of the `tiles` setter, so it's called every time the `tiles` QueryList updates.  I would have kept the QueryList as it is, but it makes iteration in `layoutTiles` a pain because QueryList's  `forEach` doesn't support an index parameter.

src/components/grid-list/grid-list.ts
kara: Yeah, that's better. Done!

src/components/grid-list/tile-coordinator.ts
kara: Yeah, it's part of the ES6 shim.

src/components/grid-list/tile-coordinator.ts
kara: Done!

src/components/grid-list/grid-list.ts
kara: Actually, we do want `''` in that case.  "0" isn't a valid input, so we want to strip it out for checks later.

src/components/grid-list/grid-list.ts
kara: Done!  Agree.

src/components/grid-list/grid-list.ts
kara: Added

src/components/grid-list/grid-tile.html
kara: Not sure either. I was planning to look into it as part of the Sass rewrite.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,153 @@
...
+
+  /** Set internal representation of row height from the user-provided value. */
+  @Input()
+  set rowHeight(value: string | number) {
+    this._rowHeight = coerceToString(value);
+    this._setTileStyler();
+  }
+
+  /** Query list of tiles that are being rendered. */
+  @ContentChildren(MdGridTile) _tiles: QueryList<MdGridTile>;
```
jelbourn: Can you move this definition above the constructor?

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,153 @@
...
+  /** Sets style on the main grid-list element, given the style name and value */
+  setListStyle(style: [string, string]): void {
+    if (style) {
+      this._renderer.setElementStyle(this._element.nativeElement, style[0], style[1]);
+    }
+  }
+}
+
+/** internal */
+/**  Converts values into strings. Falsy values become empty strings. */
```
jelbourn: The JsDoc and `@internal` can go in the same comment block:
```ts
/** 
 * Converts values into strings. Falsy values become empty strings.
 * @internal
 */
```

src/components/grid-list/grid-tile.html
jelbourn: Add a TODO?

src/components/grid-list/grid-tile.ts
```diff
@@ -0,0 +1,50 @@
...
+  _element: HTMLElement;
+  _isInitialized: boolean = false;
+
+  constructor(private _renderer: Renderer, element: ElementRef) {
+    this._element = element.nativeElement;
+  }
+
+  @Input('rowspan')
+  get rowspan() {
+  return this._rowspan;
```
jelbourn: Indent, and you also don't need the `'rowspan'` string (same for colspan below).

src/components/grid-list/grid-tile.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  set rowspan(value) {
+    this._rowspan = coerceToNumber(value);
+  }
+
+  set colspan(value) {
+    this._colspan = coerceToNumber(value);
+  }
+
+  setStyle(property: string, value: string): void {
```
jelbourn: This seems like it should also be marked as `@internal`. I think it would also benefit from a comment explaining why styles need to be directly set.

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,153 @@
...
+      this._tileStyler = new FitTileStyler();
+    } else if (this._rowHeight && this._rowHeight.match(/:/g)) {
+      this._tileStyler = new RatioTileStyler(this._rowHeight);
+    } else {
+      this._tileStyler = new FixedTileStyler(this._rowHeight);
+    }
+  }
+
+  /** Computes and applies the size and position for all children grid tiles. */
+  layoutTiles(): void {
```
jelbourn: This method could be `private`

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,153 @@
...
+    for (let i = 0; i < tiles.length; i++) {
+      let pos = tracker.positions[i];
+      let tile = tiles[i];
+      this._tileStyler.setStyle(tile, pos.row, pos.col);
+    }
+    this.setListStyle(this._tileStyler.getComputedHeight());
+  }
+
+  /** Sets style on the main grid-list element, given the style name and value */
+  setListStyle(style: [string, string]): void {
```
jelbourn: This method should probably be `@internal`

src/components/grid-list/grid-tile.ts
```diff
@@ -0,0 +1,50 @@
...
+  host: { 'role': 'listitem' },
+  templateUrl: './components/grid-list/grid-tile.html',
+  styleUrls: ['./components/grid-list/grid-list.css'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class MdGridTile {
+  _rowspan: number = 1;
+  _colspan: number = 1;
+  _element: HTMLElement;
+  _isInitialized: boolean = false;
```
jelbourn: Is `_isInitialized` used anywhere?

src/components/grid-list/tile-coordinator.ts
```diff
@@ -0,0 +1,138 @@
...
+  /** Update the tile tracker to account for the given tile in the given space. */
+  _markTilePosition(start: number, tile: MdGridTile): void {
+    for (let i = 0; i < tile.colspan; i++) {
+      this.tracker[start + i] = tile.rowspan;
+    }
+  }
+}
+
+/** Simple data structure for tile position (row, col). */
+export class Position {
```
jelbourn: `Position` > `TilePosition`?

src/components/grid-list/tile-coordinator.ts
```diff
@@ -0,0 +1,138 @@
...
+  columnIndex: number = 0;
+
+  /** The current row index. */
+  rowIndex: number = 0;
+
+  /** Gets the total number of rows occupied by tiles */
+  get rowCount(): number { return this.rowIndex + 1; }
+
+  /** Gets the total span of rows occupied by tiles
+   ** Ex: a list with 1 row that contains a tile with rowspan 2 will have a total rowspan of 2 */
```
jelbourn: Typical JsDoc format is only one `*` on each comment line after the first, with the `*/` going on its own line.

src/components/grid-list/tile-coordinator.ts
```diff
@@ -0,0 +1,138 @@
...
+  get rowspan() {
+    let lastRowMax = Math.max(...this.tracker);
+    // if any of the tiles has a rowspan that pushes it beyond the total row count,
+    // add the difference to the rowcount
+    return lastRowMax > 1 ? this.rowCount + lastRowMax - 1 : this.rowCount;
+  }
+
+  // The computed (row, col) position of each tile (the output).
```
jelbourn: This one should be `/** */` style

src/components/grid-list/tile-coordinator.ts
```diff
@@ -0,0 +1,138 @@
...
+  positions: Position[];
+
+  constructor(numColumns: number, tiles: MdGridTile[]) {
+    this.tracker = new Array(numColumns);
+    this.tracker.fill(0, 0, this.tracker.length);
+
+    this.positions = tiles.map(tile => this._trackTile(tile));
+  }
+
+  _trackTile(tile: MdGridTile): Position {
```
jelbourn: Add function descriptions for those that don't have one?

src/components/grid-list/tile-coordinator.ts
```diff
@@ -0,0 +1,136 @@
...
+    // find a gap large enough for this tile
```
jelbourn: optional nit: most comments capitalize the first letter and use punctuation at the end of the sentence, might want to be consistent throughout the file.

src/components/grid-list/tile-coordinator.ts
jelbourn: Also could be `@internal`, I believe.

src/components/grid-list/grid-list.ts
kara: Done

src/components/grid-list/grid-list.ts
kara: Done

src/components/grid-list/tile-styler.ts
```diff
@@ -0,0 +1,196 @@
...
+/* Sets the style properties for an individual tile, given the position calculated by the
+* Tile Coordinator. */
+export class TileStyler {
+  _gutterSize: string;
+  _rows: number = 0;
+  _rowspan: number = 0;
+  _cols: number;
+  _dir: Dir;
+
+  init(_gutterSize: string, tracker: TileCoordinator, cols: number, dir: Dir): void {
```
jelbourn: Why `init` instead of the constructor? Add a comment explaining?

src/components/grid-list/grid-list.ts
kara: Done

src/components/grid-list/grid-tile.html
kara: Done

src/components/grid-list/grid-tile.ts
kara: Good catch. It's from an earlier iteration. Removed.

src/components/grid-list/grid-tile.ts
kara: Ack, don't know why I keep doing that. Removed.

src/components/grid-list/grid-list.ts
kara: Done

src/components/grid-list/grid-tile.ts
kara: Done

src/components/grid-list/tile-coordinator.ts
kara: Done

src/components/grid-list/tile-coordinator.ts
kara: Done

src/components/grid-list/tile-coordinator.ts
kara: Done

src/components/grid-list/tile-styler.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  // This method will be implemented by each type of TileStyler
+  getComputedHeight(): [string, string] { return null; }
+}
+
+
+/*  This type of styler is instantiated when the user passes in a fixed row height
+*   Example <md-grid-list cols="3" rowHeight="100px"> */
+export class FixedTileStyler extends TileStyler {
+  mode = RowHeightMode.FIXED;
```
jelbourn: Is the `mode` on the `Styler` classes ever used for anything?

src/components/grid-list/tile-coordinator.ts
kara: Done

src/components/grid-list/tile-coordinator.ts
kara: Done

src/components/grid-list/grid-list.ts
```diff
@@ -1,8 +1,155 @@
...
+    for (let i = 0; i < tiles.length; i++) {
+      let pos = tracker.positions[i];
+      let tile = tiles[i];
+      this._tileStyler.setStyle(tile, pos.row, pos.col);
+    }
+    this.setListStyle(this._tileStyler.getComputedHeight());
+  }
+
+  /** @internal
+   * Sets style on the main grid-list element, given the style name and value. */
```
jelbourn: Format should be 
```ts
/**
 * Sets style on the main grid-list element, given the style name and value.
 * @internal
 */
```
(here and elsewhere)

src/components/grid-list/tile-styler.ts
```diff
@@ -0,0 +1,194 @@
...
+/* Sets the style properties for an individual tile, given the position calculated by the
+* Tile Coordinator. */
+export class TileStyler {
+  _gutterSize: string;
+  _rows: number = 0;
+  _rowspan: number = 0;
+  _cols: number;
+  _dir: Dir;
+
+  /** Adds grid-list spec once it is available. Cannot be processed in the constructor
```
jelbourn: "spec" > "layout info" ?

src/components/grid-list/tile-styler.ts
```diff
@@ -0,0 +1,194 @@
+import {MdGridTile} from './grid-tile';
+import {TileCoordinator} from './tile-coordinator';
+import {MdGridListBadRatioError} from './grid-list-errors';
+import {Dir} from '../../core/rtl/dir';
+
+/* Sets the style properties for an individual tile, given the position calculated by the
+* Tile Coordinator. */
```
jelbourn: Can all of the `TileStyler` classes be marked as `@internal`?

src/components/grid-list/tile-styler.ts
```diff
@@ -0,0 +1,194 @@
...
+    // Base vertical size of a column.
+    let baseTileHeight = this.getBaseTileSize(percentHeightPerTile, gutterHeightPerTile);
+
+    tile.setStyle('top', this.getTilePosition(baseTileHeight, rowIndex));
+    tile.setStyle('height', calc(this.getTileSize(baseTileHeight, tile.rowspan)));
+  }
+}
+
+/** Wraps a CSS string in a calc function */
+function calc(exp: string): string { return `calc(${exp})`; }
```
jelbourn: Using this function is the same amount of characters as not using it. Would it make more sense to just inline it everywhere?

kara: That's what I had originally, but it was a huge pain to figure out where the template string markers went with all the parentheses. It's cleaner and easier to maintain with the function.

src/components/grid-list/tile-styler.ts
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Gets The horizontal or vertical position of a tile, e.g., the 'top' or 'left' property value.
+   * @param offset Number of tiles that have already been rendered in the row/column.
+   * @param baseSize Base size of a 1x1 tile (as computed in getBaseTileSize).
+   * @return Position of the tile as a CSS calc() expression.
+   */
+  getTilePosition(baseSize: string, offset: number): string {
+    // The position comes the size of a 1x1 tile plus gutter for each previous tile in the
+    // row/column (offset).
+    return `calc( (${baseSize} + ${this._gutterSize}) * ${offset} )`;
```
jelbourn: Use `calc` function here?

jelbourn:
LGTM

zigzag95:
Ouff

Envoyé de mon iPhone

zigzag95:
Cool this tool jelbournalyzer.

PR #407, opened 2016-05-09
feat(input): add focus function on host element

jean-merelis:
Added focus() function to the component.
Clicking anywhere on the host it will focus the input.


jelbourn:
@jean-merelis you've got some lint errors on CI

PR #415, opened 2016-05-10
feat(checkbox): adds invisible input element to leverage built-in functionality

robertmesserle:
@jelbourn @kara @hansl

### DIFF
--- a/src/components/checkbox/checkbox.html
+++ b/src/components/checkbox/checkbox.html
@@ -1,5 +1,18 @@
-<div class="md-checkbox-layout">
+<label class="md-checkbox-layout">
   <div class="md-checkbox-inner-container">
+    <input class="md-checkbox-input" type="checkbox"
+           [id]="inputId"
+           [checked]="checked"
+           [disabled]="disabled"
+           [name]="name"
+           [tabIndex]="tabindex"
+           [indeterminate]="indeterminate"
+           [attr.aria-label]="ariaLabel"
+           [attr.aria-labelledby]="ariaLabelledby"
+           (focus)="onInputFocus()"
+           (blur)="onInputBlur()"
+           (change)="onInteractionEvent($event)">
+    <div class="md-ink-ripple"></div>
     <div class="md-checkbox-frame"></div>
     <div class="md-checkbox-background">
       <svg version="1.1"
@@ -16,7 +29,7 @@
       <div class="md-checkbox-mixedmark"></div>
     </div>
   </div>
-  <label [id]="labelId">
+  <span class="md-checkbox-label">
     <ng-content></ng-content>
-  </label>
-</div>
+  </span>
+</label>
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -1,6 +1,7 @@
 @import "default-theme";
 @import "theme-functions";
 @import "variables";
+@import "mixins";
 
 /** The width/height of the checkbox element. */
 $md-checkbox-size: $md-toggle-size !default;
@@ -217,11 +218,8 @@ $_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1)
 }
 
 md-checkbox {
-  cursor: pointer;
-
-  &:focus {
-    // TODO(traviskaufman): Add ink ripple on focus state, once ripple is implemented.
-    outline: none;
+  &, label {
+    cursor: pointer;
   }
 }
 
@@ -251,7 +249,7 @@ md-checkbox {
 }
 
 // TODO(kara): Remove this style when fixing vertical baseline
-.md-checkbox-layout label {
+.md-checkbox-layout .md-checkbox-label {
   line-height: 24px;
 }
 
@@ -461,3 +459,11 @@ md-checkbox {
     }
   }
 }
+
+// Underlying native input element.
+// Visually hidden but still able to respond to focus.
+.md-checkbox-input {
+  @include md-visually-hidden;
+}
+
+@include md-temporary-ink-ripple(checkbox);
--- a/src/components/checkbox/checkbox.spec.ts
+++ b/src/components/checkbox/checkbox.spec.ts
@@ -13,7 +13,7 @@ import {By} from '@angular/platform-browser';
 import {MdCheckbox} from './checkbox';
 import {PromiseCompleter} from '../../core/async/promise-completer';
 
-
+// TODO: Implement E2E tests for spacebar/click behavior for checking/unchecking
 
 describe('MdCheckbox', () => {
   let builder: TestComponentBuilder;
@@ -28,6 +28,8 @@ describe('MdCheckbox', () => {
     let checkboxNativeElement: HTMLElement;
     let checkboxInstance: MdCheckbox;
     let testComponent: SingleCheckbox;
+    let inputElement: HTMLInputElement;
+    let labelElement: HTMLLabelElement;
 
     beforeEach(async(() => {
       builder.createAsync(SingleCheckbox).then(f => {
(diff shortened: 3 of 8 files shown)


src/components/checkbox/checkbox.html
```diff
@@ -1,5 +1,19 @@
...
+           [checked]="checked"
+           [disabled]="disabled"
+           [name]="name"
+           [attr.tabindex]="tabindex"
+           (focus)="onInputFocus()"
+           (blur)="onInputBlur()"
+           [attr.aria-label]="ariaLabel"
+           [attr.aria-labelledby]="labelId"
+           [attr.aria-checked]="getAriaChecked()"
+           [attr.aria-disabled]="disabled"
```
jelbourn: Are `aria-checked` and `aria-disabled` necessary for a native checkbox?

src/components/checkbox/checkbox.html
```diff
@@ -1,5 +1,19 @@
-<div class="md-checkbox-layout">
-  <div class="md-checkbox-inner-container">
+<label class="md-checkbox-layout">
+  <div class="md-checkbox-inner-container" [attr.for]="inputId">
+    <input #input class="md-checkbox-input" type="checkbox"
+           [id]="inputId"
+           [checked]="checked"
+           [disabled]="disabled"
+           [name]="name"
+           [attr.tabindex]="tabindex"
```
jelbourn: Should be able to do just `[tabIndex]`

src/components/checkbox/checkbox.html
```diff
@@ -1,5 +1,19 @@
...
+           (focus)="onInputFocus()"
+           (blur)="onInputBlur()"
+           [attr.aria-label]="ariaLabel"
+           [attr.aria-labelledby]="labelId"
+           [attr.aria-checked]="getAriaChecked()"
+           [attr.aria-disabled]="disabled"
+           (change)="onInteractionEvent($event)">
```
jelbourn: Does `change` always capture what you want vs. `click` ?

robertmesserle: It appears so

src/components/checkbox/checkbox.html
```diff
@@ -16,7 +30,7 @@
       <div class="md-checkbox-mixedmark"></div>
     </div>
   </div>
-  <label [id]="labelId">
+  <span class="md-checkbox-label">
```
hansl: Wouldn't that generates problems with clicks, screen readers and custom label?

robertmesserle: No - the outer wrapper is a label which automatically handles this properly since this relies on native behavior.

src/components/checkbox/checkbox.scss
```diff
@@ -461,3 +458,44 @@ md-checkbox {
...
+
+// Underlying native input element.
+// Visually hidden but still able to respond to focus.
+.md-checkbox-input {
+  position: absolute;
+  width: 0;
+  height: 0;
+  margin: 0;
+  padding: 0;
+  opacity: 0;
```
jelbourn: Setting both `height: 0` and `width: 0` will definitely cause the checkbox to be considered "hidden" in some browser + screen-reader combinations. You should use the `md-visually-hidden` mixin.

src/components/checkbox/checkbox.scss
```diff
@@ -461,3 +458,44 @@ md-checkbox {
...
+  appearance: none;
+  border: none;
+}
+
+// TODO(mtlin): Replace when ink ripple component is implemented.
+// A placeholder ink ripple, shown when keyboard focused.
+.md-ink-ripple {
```
jelbourn: Rather than copying this, can we turn it into a mixin and use it in both places?

src/components/checkbox/checkbox.spec.ts
```diff
@@ -294,14 +251,47 @@ describe('MdCheckbox', () => {
...
+        expect(inputElement.getAttribute('aria-labelledby')).toBe('some-id');
+      });
+    }));
+
+    it('should not assign aria-labelledby if none is provided', async(() => {
+      builder.createAsync(SingleCheckbox).then(f => {
+        fixture = f;
+        checkboxDebugElement = fixture.debugElement.query(By.directive(MdCheckbox));
+        checkboxNativeElement = checkboxDebugElement.nativeElement;
+        inputElement = <HTMLInputElement>checkboxNativeElement.querySelector('input');
```
jelbourn: These two tests duplicate some behavior. It would be good to just do it once in a `beforeEach`

jelbourn: Ah, they're different components. Nevermind.

src/components/checkbox/checkbox.ts
```diff
@@ -248,6 +215,32 @@ export class MdCheckbox implements ControlValueAccessor {
...
+  /**
+   * Event handler for checkbox input element.  Toggles checked state if element is not disabled.
+   * @param event
+   */
+  onInteractionEvent(event: Event) {
+    if (this.disabled) {
+      event.stopPropagation();
+    } else {
+      this.checked = !this.checked;
+      this._elementRef.nativeElement.querySelector('input').focus();
```
jelbourn: Add a comment explaining why the `focus` is necessary

src/components/checkbox/checkbox.ts
```diff
@@ -248,6 +215,32 @@ export class MdCheckbox implements ControlValueAccessor {
...
+
+  /**
+   * Event handler for checkbox input element.  Toggles checked state if element is not disabled.
+   * @param event
+   */
+  onInteractionEvent(event: Event) {
+    if (this.disabled) {
+      event.stopPropagation();
+    } else {
+      this.checked = !this.checked;
```
jelbourn: `this.toggle()` ?

src/components/checkbox/checkbox.scss
robertmesserle: This was pulled from radio-button - going to fix in both places.

src/components/checkbox/checkbox.ts
robertmesserle: I believe I removed `this.toggle()` since it wasn't actually doing what the comment described it as doing.  Should I re-add it?

src/components/checkbox/checkbox.ts
robertmesserle: Actually, now that I implemented the label-based solution, this isn't necessary at all.

src/components/checkbox/checkbox.ts
jelbourn: Yes, it should be part of the checkbox's public API

src/components/checkbox/checkbox.html
```diff
@@ -1,5 +1,18 @@
-<div class="md-checkbox-layout">
+<label class="md-checkbox-layout">
   <div class="md-checkbox-inner-container">
+    <input #input class="md-checkbox-input" type="checkbox"
```
DevVersion: Why has the input a local variable? Can't see it used in any other files.

src/components/checkbox/checkbox.ts
```diff
@@ -248,6 +215,35 @@ export class MdCheckbox implements ControlValueAccessor {
...
+   * We use a hidden native input field to handle changes to focus state via keyboard navigation,
+   * with visual rendering done separately. The native element is kept in sync with the overall
+   * state of the component.
+   */
+  onInputFocus() {
+    this.hasFocus = true;
+  }
+
+  onInputBlur() {
+    this.hasFocus = false;
```
DevVersion: Should this set the ValueAccessor to touched as well? Excuse me if I'am wrong. Just wondering, because the `(blur)` host binding still exists on the components root.

src/components/checkbox/checkbox.html
jelbourn: Yeah, I don't think this is used any more.

src/components/checkbox/checkbox.ts
jelbourn: Good catch- that `(blur)` on the host should be removed and the `onTouched` should be called here.

src/components/checkbox/checkbox.ts
```diff
@@ -248,6 +215,35 @@ export class MdCheckbox implements ControlValueAccessor {
...
+  }
+
+  toggle() {
```
jelbourn: Add a JsDoc for `toggle`
(I know it's obvious, but we're eventually going to pull some API docs from these)

src/components/checkbox/checkbox.html
```diff
@@ -1,5 +1,18 @@
...
+           [id]="id"
+           [checked]="checked"
+           [disabled]="disabled"
+           [name]="name"
```
DevVersion: Also it looks like, the `name` property doesn't exist as an input / property.

src/components/checkbox/checkbox.ts
```diff
@@ -248,6 +215,35 @@ export class MdCheckbox implements ControlValueAccessor {
...
+    this.checked = !this.checked;
+  }
+
+  /**
+   * Event handler for checkbox input element.  Toggles checked state if element is not disabled.
+   * @param event
+   */
+  onInteractionEvent(event: Event) {
```
jelbourn: This should be `@internal`

src/components/checkbox/checkbox.ts
```diff
@@ -248,6 +215,35 @@ export class MdCheckbox implements ControlValueAccessor {
...
+  /*
+   * We use a hidden native input field to handle changes to focus state via keyboard navigation,
+   * with visual rendering done separately. The native element is kept in sync with the overall
+   * state of the component.
+   */
+  onInputFocus() {
+    this.hasFocus = true;
+  }
+
+  onInputBlur() {
```
jelbourn: Both of these methods should be `@internal`

src/components/checkbox/checkbox.html
```diff
@@ -1,5 +1,18 @@
-<div class="md-checkbox-layout">
+<label class="md-checkbox-layout">
   <div class="md-checkbox-inner-container">
+    <input #input class="md-checkbox-input" type="checkbox"
+           [id]="id"
```
jelbourn: You're using the same ID as the host element 
(pointed out by @DevVersion)

src/components/checkbox/checkbox.html
robertmesserle: Ah, good call - it isn't used anymore.

jelbourn:
LGTM

PR #431, opened 2016-05-13
chore(icons): use individual RxJS imports

robwormald:
using `rxjs/Rx` import brings in the entire RxJS library (300+ KB) - this changes to import only the required operators.


jelbourn:
@robwormald seems to have broke all the things

jelbourn:
LGTM

Issue #327, opened 2016-04-18
Radio buttons do not become unchecked when a bound ngModel property is removed


pmulac:
Consider the following radio group:

```html
<md-radio-group  [ngModel]="model.accountType" ngControl="accountType">
    <md-radio-button value="I">Individual</md-radio-button>
    <md-radio-button value="J">Joint</md-radio-button>
</md-radio-group>
```

And then the ngModel object changes like so:

Original state:  `{accountType: 'J'}`  
New state:  `{}`   

I would expect the corresponding radio button ('Joint') to be unchecked.  However, the 'Joint' radio button remains checked, which does not represent the true state of the control.

As a workaround, I'm able to bind to the `md-radio-checked` class on each button, which removes the checked state of a radio button when the corresponding binding changes to undefined.

```html
<md-radio-group  [ngModel]="model.accountType" ngControl="accountType">
    <md-radio-button value="I" [class.md-radio-checked]="model.accountType === 'I'">Individual</md-radio-button>
    <md-radio-button value="J" [class.md-radio-checked]="model.accountType === 'J'">Joint</md-radio-button>
</md-radio-group>
```

This is especially important for a redux/ngrx application, in which you would expect radio buttons to become checked/unchecked as you replay recorded actions.

lobo-tuerto:
I'm observing the same behavior with reactive forms :(

PR #436, opened 2016-05-13
fix(radio-button): Radio buttons are not tab stops in Safari

trik:


trik:
I signed it!

robertmesserle:
Thanks for your contribution!  However, this issue was fixed as part of this PR: https://github.com/angular/material2/pull/415

trik:
@robertmesserle what about this one? https://github.com/angular/material2/pull/436/commits/d56946ad6ca8d41a77dc4e6118361f7926bffd78

sorry, i had to make a different pr, it refers #327

robertmesserle:
@trik We do still need a fix for this, but when I tried applying that commit to master and running the tests, it doesn't appear to fix the bug.  Were there other changes in one of the other commits required to make this work?

trik:
@robertmesserle no, only that one.. why it doesn't fix? in then unit test i set a value for the radio group model and i get one button selected, then i set the model to null and all buttons are de-selected

robertmesserle:
The failure I get is on the test you added.  It needed the following line:

```ts
this._radios.forEach(radio => radio.checked = false);
```

Without that, it never actually unchecked any of the child radio buttons.

robertmesserle:
Ah, I see why your tests weren't failing.  Your tests were never running.

There was a bug in the radio button tests where `fakeAsync` was wrapped around a function, which returns a new function - but that new function is never called.  So your tests would have passed regardless of what they checked.  (This was an issue with all tests in radio-button and is being fixed now by @jelbourn).

There were some other changes required to make this work as well.  I'll open a PR with the working changes and mention you in it.

PR #439, opened 2016-05-13
fix(button): remove unused arg

kara:
r: @jelbourn 

This unused arg is causing the offline template compiler to complain.


jelbourn:
LGTM

PR #441, opened 2016-05-13
fix(radio): unsetting the model will deselect all radio buttons

### LINKED ISSUE #327: Radio buttons do not become unchecked when a bound ngModel property is removed

robertmesserle:
closes #327 

Thanks to @trik for help resolving this issue.

### DIFF
--- a/src/components/radio/radio.spec.ts
+++ b/src/components/radio/radio.spec.ts
@@ -22,8 +22,8 @@ export function main() {
       builder = tcb;
     }));
 
-    it('should have same name as radio group', (done: () => void) => {
-      builder
+    it('should have same name as radio group', () => {
+      return builder
         .overrideTemplate(TestApp, `
             <md-radio-group name="my_group">
               <md-radio-button></md-radio-button>
@@ -34,11 +34,11 @@ export function main() {
 
           fixture.detectChanges();
           expect(button.componentInstance.name).toBe('my_group');
-        }).then(done);
+        });
     });
 
-    it('should not allow click selection if disabled', (done: () => void) => {
-      builder
+    it('should not allow click selection if disabled', () => {
+      return builder
         .overrideTemplate(TestApp, '<md-radio-button disabled></md-radio-button>')
         .createAsync(TestApp)
         .then(fixture => {
@@ -49,11 +49,11 @@ export function main() {
 
           button.nativeElement.click();
           expect(button.componentInstance.checked).toBe(false);
-        }).then(done);
+        });
     });
 
-    it('should be disabled if radio group disabled', (done: () => void) => {
-      builder
+    it('should be disabled if radio group disabled', () => {
+      return builder
         .overrideTemplate(TestApp, `
             <md-radio-group disabled>
               <md-radio-button></md-radio-button>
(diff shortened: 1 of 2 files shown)


robertmesserle:
@jelbourn Should be ready to merge now.

src/components/radio/radio.spec.ts
```diff
@@ -324,6 +324,29 @@ export function main() {
         }).then(done);
     });
 
+    it('should deselect all buttons when model is null or undefined', (done: () => void) => {
```
jelbourn: You should be able to use `async(` instead of `done`

src/components/radio/radio.spec.ts
```diff
@@ -336,6 +359,11 @@ function isSinglySelected(button: DebugElement, buttons: DebugElement[]): boolea
   return component.checked && otherSelectedButtons.length == 0;
 }
 
+function allDeselected(buttons: DebugElement[]): boolean {
+  let selectedButtons = buttons.filter((e: DebugElement) => e.componentInstance.checked);
```
jelbourn: You shouldn't need the type on `e` here.

src/components/radio/radio.spec.ts
```diff
@@ -336,6 +359,11 @@ function isSinglySelected(button: DebugElement, buttons: DebugElement[]): boolea
...
+  return selectedButtons.length == 0;
```
jelbourn: This method can be
```ts
return buttons.every(debugEl => !debugEl.componentInstance.checked);
```

src/components/radio/radio.ts
```diff
@@ -150,12 +150,14 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
       });
 
       if (matched.length == 0) {
-        // Didn't find a button that matches this value, return early without setting.
-        return;
+        if (this.value == null) {
+          this.selected = null;
+          // Manually uncheck all radio buttons
```
jelbourn: Move this comment above the `if` and say something like
```
// When the value of the group is cleared to null, deselect all radio button in the group.
```

jelbourn:
LGTM

PR #443, opened 2016-05-13
fix(radio): refactor tests and fix ngModel

jelbourn:
R: @kara @robertmesserle 

* Completely rewrite radio-button unit tests to be much simpler (and also actually run several tests that weren't running before).
* Fix ngControl always being set to `dirty` (vs. `pristine` initially)
* Fix ngControl never being able to become `touched` 
* Fix setting `checked` directly on a radio button not updating the group's `value`.
* Fix firing a `change` event on initial setting of value at application load.
* Mark methods as `@internal`

### DIFF
--- a/src/components/radio/radio.spec.ts
+++ b/src/components/radio/radio.spec.ts
@@ -1,396 +1,390 @@
 import {
-  it,
-  describe,
-  expect,
-  beforeEach,
-  fakeAsync,
-  inject,
-  tick,
+    it,
+    describe,
+    beforeEach,
+    beforeEachProviders,
+    inject,
+    async,
+    fakeAsync,
+    tick
 } from '@angular/core/testing';
-import {TestComponentBuilder} from '@angular/compiler/testing';
-import {Component, DebugElement} from '@angular/core';
+import {FORM_DIRECTIVES, NgControl} from '@angular/common';
+import {TestComponentBuilder, ComponentFixture} from '@angular/compiler/testing';
+import {Component, DebugElement, provide} from '@angular/core';
 import {By} from '@angular/platform-browser';
-
-import {MdRadioButton, MdRadioGroup, MdRadioChange} from './radio';
+import {MD_RADIO_DIRECTIVES, MdRadioGroup, MdRadioButton} from './radio';
 import {MdRadioDispatcher} from './radio_dispatcher';
 
-export function main() {
-  describe('MdRadioButton', () => {
-    let builder: TestComponentBuilder;
 
-    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
-      builder = tcb;
-    }));
+describe('MdRadio', () => {
+  let builder: TestComponentBuilder;
+  let dispatcher: MdRadioDispatcher;
 
-    it('should have same name as radio group', () => {
-      return builder
-        .overrideTemplate(TestApp, `
-            <md-radio-group name="my_group">
-              <md-radio-button></md-radio-button>
-            </md-radio-group>`)
-        .createAsync(TestApp)
-        .then(fixture => {
-          let button = fixture.debugElement.query(By.css('md-radio-button'));
-
-          fixture.detectChanges();
-          expect(button.componentInstance.name).toBe('my_group');
-        });
-    });
-
-    it('should not allow click selection if disabled', () => {
-      return builder
-        .overrideTemplate(TestApp, '<md-radio-button disabled></md-radio-button>')
-        .createAsync(TestApp)
-        .then(fixture => {
-          let button = fixture.debugElement.query(By.css('md-radio-button'));
-
-          fixture.detectChanges();
-          expect(button.componentInstance.checked).toBe(false);
-
-          button.nativeElement.click();
-          expect(button.componentInstance.checked).toBe(false);
-        });
-    });
-
-    it('should be disabled if radio group disabled', () => {
-      return builder
-        .overrideTemplate(TestApp, `
-            <md-radio-group disabled>
-              <md-radio-button></md-radio-button>
-            </md-radio-group>`)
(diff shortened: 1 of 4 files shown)


src/components/radio/radio.spec.ts
```diff
@@ -1,370 +1,398 @@
...
+    beforeEach(async(() => {
+      builder.createAsync(RadiosInsideRadioGroup).then(f => {
+        fixture = f;
+        fixture.detectChanges();
+
+        testComponent = fixture.debugElement.componentInstance;
+
+        groupDebugElement = fixture.debugElement.query(By.directive(MdRadioGroup));
+        groupNativeElement = groupDebugElement.nativeElement;
+        groupInstance = groupDebugElement.injector.get(MdRadioGroup);
```
robertmesserle: Is this the same as `groupDebugElement.componentInstance`?

src/components/radio/radio.spec.ts
```diff
@@ -1,370 +1,398 @@
...
-        .createAsync(TestApp)
-        .then(fixture => {
-          let button = fixture.debugElement.query(By.css('md-radio-button'));
+    it('should disable click interaction when the group is disabled', () => {
+      testComponent.isGroupDisabled = true;
+      fixture.detectChanges();
 
-          fixture.detectChanges();
-          expect(button.componentInstance.checked).toBe(false);
+      radioNativeElements[0].click();
```
robertmesserle: We could actually remove the click handler altogether for Radio Buttons as we did with Checkbox; however, if that happens, this test would have to be rewritten to work in IE11.

src/components/radio/radio.spec.ts
```diff
@@ -1,370 +1,398 @@
...
+        }
+      });
+
+      radioInstances[0].checked = true;
+      fixture.detectChanges();
+
+      radioInstances[0].checked = false;
+      fixture.detectChanges();
+
+      return promiseCompleter.promise;
```
robertmesserle: Couldn't this be done more easily by using `spyOn`?  It seems needlessly confusing the way it's written.

src/components/radio/radio.spec.ts
jelbourn: Nope; radio-group is a `@Directive` rather than a `@Component`, so it doesn't have a `componentInstance`.

src/components/radio/radio.spec.ts
jelbourn: As discussed, we'll just leave this for now.

src/components/radio/radio.spec.ts
```diff
@@ -1,370 +1,398 @@
...
-    beforeEach(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
-      builder = tcb;
-      dispatcher = new MdRadioDispatcher();
-    }));
+      expect(nativeRadioInput.classList).not.toContain('md-radio-focused');
 
-    it('notifies listeners', () => {
-      let notificationCount = 0;
-      const numListeners = 2;
+      dispatchFocusChangeEvent('focus', nativeRadioInput);
```
robertmesserle: Would `nativeRadioInput.focus()` work for this?

src/components/radio/radio.ts
```diff
@@ -71,9 +77,13 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
...
   private _selected: MdRadioButton = null;
 
+  /** Whether the `value` has been set to its initial value. */
+  private _isInitialized: boolean = false;
+
   /** Change event subscription set up by registerOnChange (ControlValueAccessor). */
   private _changeSubscription: {unsubscribe: () => any} = null;
 
+  /** onTouch function registered via registerOnTouch (ControlValueAccessor). */
   onTouched: () => any = () => {};
```
robertmesserle: This is very confusing to read.  I don't have a suggestion really, but that's entirely too many `=` in a row.

src/components/radio/radio.spec.ts
jelbourn: When I originally wrote it (pre-rc0), `fakeAsync` didn't work for this, but now it seems to be good. Done.

src/components/radio/radio.spec.ts
jelbourn: I tried it and it does not seem to work. Added a TODO to write an e2e test that uses real focus.

src/components/radio/radio.ts
jelbourn: It is a bit odd, but I don't think it's that big of a deal.

jelbourn:
@robertmesserle addressed comments

src/components/radio/radio.ts
```diff
@@ -137,25 +132,66 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
...
+  touch() {
+    if (this.onTouched) {
+      this.onTouched();
     }
   }
 
+  /** Updates the `selected` radio button from the internal _value state. */
   private _updateSelectedRadioFromValue(): void {
-    // Update selected if different from current value.
+    // If the value already matches the selected radio, no dothing.
```
iveysaur: Did you mean "do nothing"?

src/components/radio/radio.ts
```diff
@@ -137,25 +132,66 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
...
+    if (selected && !selected.checked) {
+      selected.checked = true;
+    }
+  }
+
+  /**
+   * Initialize properties once content children are available.
+   * This allows us to propagate relevant attributes to associated buttons.
+   */
+  ngAfterContentInit() {
```
DevVersion: Just curious, shouldn't this method be `@internal`?

jelbourn:
Verbal lgtm from @robertmesserle

PR #444, opened 2016-05-14
fix(sidenav): correctly parse opened boolean attribute.

### LINKED ISSUE #300: md-sidenav is open even when opened="false"
i used the following code for side nav
` <md-sidenav #start mode="side" opened="false">`

but the sidenav is still open instead of closed.

DevVersion:
* The sidenav didn't parse the boolean attribute correctly.

Fixes #300.


jelbourn:
LGTM

PR #447, opened 2016-05-16
feat(input): add support for more input attributes

matthewdenobrega:
@jelbourn I've cleaned up #343 so that it is off an up-to-date master and only one commit.


src/components/input/input.html
```diff
@@ -12,11 +12,20 @@
              [attr.aria-disabled]="ariaDisabled"
              [attr.aria-required]="ariaRequired"
              [attr.aria-invalid]="ariaInvalid"
-             [id]="id"
+             [attr.autocomplete]="autoComplete"
```
jelbourn: All of the bindings you're adding in this PR are actual properties on `HTMLInputElement`, so none of them need `attr.`

src/components/input/input.spec.ts
```diff
@@ -239,6 +239,368 @@ export function main() {
           expect(typeof fixture.componentInstance.value).toBe('number');
         });
     });
+
+    it('supports the autoComplete attribute', () => {
+      var template = '<md-input [autoComplete]="autoComplete"></md-input>';
+
+      return builder.overrideTemplate(MdInputOptionalAttributeController, template)
+        .createAsync(MdInputOptionalAttributeController)
+        .then((fixture) => {
```
jelbourn: Don't need parens around  `fixture`

src/components/input/input.html
jelbourn: Though note that when using the property `tabIndex` is camelCase.

jelbourn:
Looks good side from a few minor comments.

matthewdenobrega:
@jelbourn if we use the property notation (eg. [autocomplete]="autoComplete) then the property is always present on the resulting input element. In some cases it's possible to set a default (for example 'off' for autocomplete), but this may not be the required behavior - for example with autocomplete if the property is not set it defaults to the parent form's value, where if we set it explicitly by default to false it will not. It's not clear for other properties (like min and max) what a reasonable default would be.

So I think using the attribute notation and only setting the input properties when they have been set on the md-input will give a result closer to normal expectation than using property notation and defaults. It also results in cleaner generated markup - for example not setting step="any" on input[type="text"] elements where it doesn't make sense. This is the behavior that's captured by the tests.

Point taken on the parentheses for fixture - that's a quick fix.

jelbourn:
@matthewdenobrega good point, I hadn't seen Angular tacking on the superfluous attributes yet. I also looked at `tabIndex` with Misko today and it seems that it's going to need special handling in core due to its weirdness. 

Looks good other than needing a rebase now.

matthewdenobrega:
@jelbourn I've removed the parentheses from the tests and rebased.

jelbourn:
LGTM

PR #449, opened 2016-05-16
fix(input): adds blur and focus event support

### LINKED ISSUE #337: md-input not firing blur event

Bug

Blur event is not firing on <md-input> elements, even if we provide `<md-input (blur)="logBlur()">`.

please provide steps to reproduce and if possible a minimal demo of the problem**
via https://plnkr.co or similar.
Can't find base plunker for md-input

Blur event should fire

Tested on 2.0.0.alpha-2 in Chrome

* **Other information**

robertmesserle:
closes #337

### DIFF
--- a/src/components/input/input.html
+++ b/src/components/input/input.html
@@ -18,10 +18,10 @@
              [spellcheck]="spellcheck"
              [attr.maxlength]="maxLength"
              [type]="type"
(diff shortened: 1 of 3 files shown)


jelbourn:
LGTM

PR #450, opened 2016-05-16
fix(input): input element should have a different id from outer element

### LINKED ISSUE #320: Duplicate id's on md-input

Report a bug

The id attribute of an md-input is copied onto the inner input element, which results in duplicate id's on the page.

please provide steps to reproduce and if possible a minimal demo of the problem**
via https://plnkr.co or similar.

This is as per the design document - 'The id will be transferred to the native input element.' - however because the original element also has the id, this results in duplicate id's on the page. The internal input id should be modified to avoid this.

Current behavior is not compliant with the HTML spec

Tested on 2.0.0.alpha-2 in Chrome

* **Other information**
I've sketched a possible solution in #321

robertmesserle:
closes #320

### DIFF
--- a/src/components/input/input.html
+++ b/src/components/input/input.html
@@ -12,7 +12,7 @@
              [attr.aria-disabled]="ariaDisabled"
              [attr.aria-required]="ariaRequired"
              [attr.aria-invalid]="ariaInvalid"
-             [id]="id"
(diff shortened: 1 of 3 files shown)


jelbourn:
LGTM

PR #452, opened 2016-05-16
fix(input): adds support for name property for input element

### LINKED ISSUE #446: Add support for name in MdInput
name attribute is necessary for form method="post"
it will be useful if md-input supports name

robertmesserle:
R: @jelbourn @hansl 
closes #446

### DIFF
--- a/src/components/input/input.html
+++ b/src/components/input/input.html
@@ -27,6 +27,7 @@
              [attr.step]="step"
              [attr.tabindex]="tabIndex"
              [type]="type"
+             [attr.name]="name"
              (focus)="handleFocus($event)"
              (blur)="handleBlur($event)"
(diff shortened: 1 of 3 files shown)


jelbourn:
LGTM, you can rebase and merge

src/components/input/input.html
```diff
@@ -18,6 +18,7 @@
              [spellcheck]="spellcheck"
              [attr.maxlength]="maxLength"
              [type]="type"
+             [name]="name"
```
jelbourn: For the time being, this actually needs to be `attr.name` so that Angular does add an `name="null"` attribute when it's absent.

robertmesserle:
@jelbourn rebased

PR #456, opened 2016-05-17
fix(radio): model should update before change event is fired

### LINKED ISSUE #448: Should md-radio-group (change) event update the model before calling the event handler?
When a `md-radio-group` is bound via `[(ngModel)]` to a class property, the value emitted by the `change` event and passed to the event handler does not match the class property.

Example:
`<md-radio-group name="mastery-group" [(ngModel)]="_chosen" (change)="onChange($event.value)">`

Suppose that `_chosen` is set to TRUE in `ngOnInit`.  If the user selects the radio button with value FALSE, in the `onChange` handler, the value received will be FALSE (as expected), but `_chosen` will still equal TRUE (unexpected).  `_chosen` is not set to FALSE until after the event handler is run.  It seems the event is triggered before the ngModel bound property is updated. Is this by design?

It seems inconsistent with how Angular2 manages other events. Typically, if an input is [(ngModel)] bound to a class property, you don't need to pass the event to the handler to figure out the new value.  The handler is able to read the new value directly from the class property. 

I'm a newbie at Angular so I'm not too sure about this.  Am I wrong here?

robertmesserle:
R: @jelbourn 
closes #448

### DIFF
--- a/src/components/radio/radio.spec.ts
+++ b/src/components/radio/radio.spec.ts
@@ -297,6 +297,50 @@ describe('MdRadio', () => {
     }));
   });
 
+  describe('group with ngModel and change event', () => {
+    let fixture: ComponentFixture<RadioGroupWithNgModel>;
+    let groupDebugElement: DebugElement;
+    let groupNativeElement: HTMLElement;
+    let radioDebugElements: DebugElement[];
+    let radioNativeElements: HTMLElement[];
+    let groupInstance: MdRadioGroup;
+    let radioInstances: MdRadioButton[];
+    let testComponent: RadioGroupWithNgModel;
+    let groupNgControl: NgControl;
+
+    beforeEach(async(() => {
+      builder.createAsync(RadioGroupWithNgModel).then(f => {
+        fixture = f;
+
+        testComponent = fixture.componentInstance;
+
+        groupDebugElement = fixture.debugElement.query(By.directive(MdRadioGroup));
+        groupNativeElement = groupDebugElement.nativeElement;
+        groupInstance = groupDebugElement.injector.get(MdRadioGroup);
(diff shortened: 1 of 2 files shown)


src/components/radio/radio.ts
```diff
@@ -71,8 +71,8 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
   /** The currently selected radio button. Should match value. */
   private _selected: MdRadioButton = null;
 
-  /** Change event subscription set up by registerOnChange (ControlValueAccessor). */
-  private _changeSubscription: {unsubscribe: () => any} = null;
+  /** The method to be called in order to update ngModel */
+  private _onChangeHandler: (value: number) => void = (value) => {};
```
jelbourn: `number`?

jelbourn: What do you think of renaming this to `_controlValueAccessorChangeFn`? (similar for the touch one as well)

robertmesserle: 1. I have no idea where `number` came from.
2. I'm for the rename - I was going to ask you to rename this for me since I had no idea what to call it.

robertmesserle:
@jelbourn rebased

src/components/radio/radio.spec.ts
```diff
@@ -273,6 +273,50 @@ describe('MdRadio', () => {
     }));
   });
 
+  describe('group with ngModel and change event', () => {
```
jelbourn: Why not use the existing `"group with ngModel"` block?

robertmesserle:
@jelbourn Addressed comments

robertmesserle:
@jelbourn this is once again green and good to merge!

jelbourn:
LGTM

PR #457, opened 2016-05-17
fix(input): adds support for min/max on number inputs

### LINKED ISSUE #413: md-input ignores min and max values
The code in question is:

      <md-input type="number" class="form-control"  min="1" align="end"  [(ngModel)]="numberOf" required></md-input>

vs:

      <input type="number" class="form-control"  min="1" align="end"  [(ngModel)]="numberOf" required>

I would guess this is a bug since the html5 code works?

My input is supposed to only take positive numbers (checked this in FF and Chrome):

(material is on the left, html5 is on the right):

Using arrows, I cannot make the html5 number input lower than the min value (in this case 1). It also displays as red when manually lower than the min value. 

* **Demo of issue** 

http://plnkr.co/edit/XKVHqFNNGDUDZ0S6fTHs?p=preview

To prevent negative values

Because it works in HTML 5 by default.

I am currently using the latest of Angular 2 and Material. I've tested it in Chrome and Firefox.

* **Other information**

I have no idea how to fix... sorry.

robertmesserle:
closes #413

### DIFF
--- a/src/components/input/input.html
+++ b/src/components/input/input.html
@@ -13,6 +13,8 @@
              [attr.aria-required]="ariaRequired"
              [attr.aria-invalid]="ariaInvalid"
              [id]="inputId"
+             [attr.min]="min"
+             [attr.max]="max"
              [disabled]="disabled"
              [required]="required"
              [spellcheck]="spellcheck"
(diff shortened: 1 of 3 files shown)


DevVersion:
@robertmesserle Not sure, This seems to be a duplicate of #447?

robertmesserle:
@DevVersion Ah, didn't see that one.  I'll close this one for now.

PR #458, opened 2016-05-17
fix(radio): radio group without name should no longer error

### LINKED ISSUE #403: radio does not work as mentioned in the docs
the radio button example given in the [README.md](https://github.com/angular/material2/tree/master/src/components/radio) does not work reliably and gives this error 
`Expression has changed after it was checked. Previous value: 'undefined'. Current value: `

to make it work reliably , you need to have the name attribute in the group and radio buttons. The name attribute should match.

If this is by design , the documentation is not correct

robertmesserle:
R: @jelbourn 
closes #403

### DIFF
--- a/src/components/radio/radio.ts
+++ b/src/components/radio/radio.ts
@@ -4,6 +4,7 @@ import {
   ContentChildren,
   Directive,
   EventEmitter,
+  ChangeDetectorRef,
   HostBinding,
   Input,
   OnInit,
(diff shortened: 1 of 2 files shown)


src/components/radio/radio.ts
```diff
@@ -84,16 +85,14 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
...
    * Initialize properties once content children are available.
    * This allows us to propagate relevant attributes to associated buttons.
    */
   ngAfterContentInit() {
-    if (this._name == null) {
-      this.name = `md-radio-group-${_uniqueIdCounter++}`;
-    } else {
-      this._updateChildRadioNames();
-    }
+    this._updateChildRadioNames();
```
jelbourn: As discussed, change this such that the child radio pulls its name from the parent instead of the other way around.

jelbourn:
@robertmesserle Needs rebase

robertmesserle:
This seems to have been resolved by another commit - closing.

PR #460, opened 2016-05-18
fix(sidenav): removes margin transitions

### LINKED ISSUE #404: Reconsider margin transition on sidenav
Bug

Poor performance for sidenav toggle transition

@jelbourn's [material2 demo](https://material2-app.firebaseapp.com/) and [sidenav template](https://github.com/angular/material2/blob/master/src/components/sidenav/sidenav.html#L6) show that transition is targeting `margin-left`.

I would expect animating `transform: translateX` unless there's a reason that I missed

Smoother toggle transition

Material alpha 4

* **Other information**
I would love to fix this if you agree
Thanks!

robertmesserle:
R: @jelbourn @hansl 
closes #404

### DIFF
--- a/src/components/sidenav/sidenav-transitions.scss
+++ b/src/components/sidenav/sidenav-transitions.scss
@@ -5,9 +5,6 @@
 
 
 :host {
-  transition: margin-left $swift-ease-out-duration $swift-ease-out-timing-function,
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #467, opened 2016-05-19
fix(button): focus state matches spec

### LINKED ISSUE #420: Focus indicators on md-fab is really subtle
Hard to see the difference unless you really focus on it.

jelbourn:
R: @robertmesserle @kara 

Also, @traviskaufman, could you verify that I'm interpreting [the spec](https://www.google.com/design/spec/components/buttons.html#buttons-raised-buttons) correctly here?

Closes #420

### DIFF
--- a/src/components/button/_button-base.scss
+++ b/src/components/button/_button-base.scss
@@ -85,9 +85,19 @@ $md-mini-fab-padding: 8px !default;
     @include md-elevation(8);
   }
 
-  &.md-button-focus {
-    background-color: md-color($md-foreground, base, 0.12);
-    @include md-button-theme('background-color', 600);
+  &.md-button-focus:after {
+    // The button spec calls for focus on raised buttons (and FABs) to be indicated with a
+    // black, 12% opacity shade over the normal color (for both light and dark themes).
+    // We do this by placing an :after psuedo-element with the appropriate shade over the button.
+    position: absolute;
+    top: 0;
+    left: 0;
+    bottom: 0;
+    right: 0;
+    content: '';
+    background-color: rgba(black, 0.12);
+    border-radius: inherit;
+    pointer-events: none;
   }
 
   &[disabled] {
@@ -109,9 +119,5 @@ $md-mini-fab-padding: 8px !default;
 
   i, md-icon {
     padding: $padding 0;
-  }
-
-  &.md-button-focus {
-    background-color: md-color($md-accent, 600);
   }
 }
--- a/src/components/button/_button-base.scss
+++ b/src/components/button/_button-base.scss
@@ -63,6 +63,10 @@ $md-mini-fab-padding: 8px !default;
   &[disabled] {
     cursor: default;
   }
+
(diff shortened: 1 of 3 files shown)


src/components/button/_button-base.scss
```diff
@@ -85,9 +85,17 @@ $md-mini-fab-padding: 8px !default;
     @include md-elevation(8);
   }
 
-  &.md-button-focus {
-    background-color: md-color($md-foreground, base, 0.12);
-    @include md-button-theme('background-color', 600);
+  &.md-button-focus:after {
+    // The button spec calls for focus on rasied buttons (and FABs) to be indicated with a
```
robertmesserle: typo: rasied => raised

src/components/button/_button-base.scss
```diff
@@ -85,9 +85,17 @@ $md-mini-fab-padding: 8px !default;
...
+    // black, 12% opacity shade over the normal color (for both light and dark themes).
+    // We do this by placing an :after psuedo-element with the appropriate shade over the button.
+    position: absolute;
+    top: 0;
+    left: 0;
+    bottom: 0;
+    right: 0;
+    content: '';
+    background-color: rgba(black, 0.12);
```
robertmesserle: Probably want to include `pointer-events: none`.  Not sure if it's fully necessary in our use-case, but it's possibly that the pseudo element can intercept click events without it.

jelbourn:
@robertmesserle comments addressed

src/components/button/_button-base.scss
DevVersion: This color should be able to be picked from `$md-foreground` with the hue-key `divider`?

jelbourn: According to the spec, this shade should always be black 12% opacity regardless of theme.

traviskaufman: The UX treatments LGTM. However this looks like it only applies to raised buttons. Flat buttons still seem to be [styled using background color](https://github.com/jelbourn/material2/blob/c8462f59f537b8c12573f818f12e021dd6f88411/src/components/button/button.scss#L10). Is the intent to style the two types of buttons using different strategies?

src/components/button/_button-base.scss
```diff
@@ -111,7 +120,8 @@ $md-mini-fab-padding: 8px !default;
     padding: $padding 0;
   }
 
-  &.md-button-focus {
-    background-color: md-color($md-accent, 600);
+  // The focus shading for FABs needs to match the FAB's circular shape.
+  &.md-button-focus:after {
+    border-radius: 50%;
```
traviskaufman: Would it be better to use `border-radius: inherit` in the declaration above instead? This way it will also respect the border radii for normal flat + raised buttons.

jelbourn:
@kara I followed Travis' advice and unified the focus styles to all be done the same way. Can you take another look?

kara:
LGTM

PR #469, opened 2016-05-19
fix(input): inputs now work in IE11

### LINKED ISSUE #336: md-input does not allow typing on IE
Bug

I cannot type on IE (11)

Focus any MD-INPUT on IE11 (probably below too but 11 is most relevant). And try to type with keyboard. Does not work.
https://material2-app.firebaseapp.com/ 
You can actually see it here.

Text appears!

Include IE 11 support seems a must have right?

Angular2 Beta 14, Material Alpha 2

* **Other information**

robertmesserle:
closes #336


robertmesserle:
R: @jelbourn @hansl

PR #476, opened 2016-05-19
test(e2e): add support for e2e app and running tests.

hansl:


e2e/components/button/button.e2e.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference path="../../e2e.d.ts" />
```
jelbourn: Is there not a way to make this implicit in the tsconfig?

src/demo-app/index.html
```diff
@@ -5,7 +5,11 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Material2</title>
   <base href="/">
-  {{content-for 'head'}}
+
+  {{#unless environment.production}}
+  <script src="/ember-cli-live-reload.js" type="text/javascript"></script>
+  {{/unless}}
```
jelbourn: Can you add comments to this templating so that passerby who look at our index.html know what it is?

src/e2e-app/button/button-demo.html
```diff
@@ -0,0 +1,19 @@
+<div class="demo-button">
```
jelbourn: Can we call these files something other than `button-demo`? I'd generally like the examples to be more like `simple-buttons`, `tabs-inside-dialog`, etc.

src/e2e-app/button/button-demo.scss
```diff
@@ -0,0 +1,19 @@
+
+.demo-button {
```
jelbourn: Rather than `demo-`, prefix `e2e` ?

e2e/components/button/button.e2e.ts
hansl: We're not using that tsconfig. I'm still looking into that part.

src/e2e-app/button/button-demo.scss
hansl: CSS isn't necessary here, so I'm removing all of it.

src/demo-app/index.html
hansl: Done.

e2e/components/button/button.e2e.ts
hansl: Fixed by upgrading the version of ts-node we were using.

hansl:
PTAL.

hansl:
R: @jelbourn, @robertmesserle

src/demo-app/index.html
```diff
@@ -1,11 +1,16 @@
-<!doctype html>
+{{!-- This file is an handlebar file that gets filled at build time. --}}<!doctype html>
```
jelbourn: doctype on next line?

jelbourn:
LGTM

PR #478, opened 2016-05-19
Fix/403/radio without name

robertmesserle:
R: @jelbourn @hansl

### DIFF
--- a/src/components/radio/radio.ts
+++ b/src/components/radio/radio.ts
@@ -69,7 +69,7 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
   private _value: any = null;
 
   /** The HTML name attribute applied to radio buttons in this group. */
-  private _name: string = null;
+  private _name: string = `md-radio-group-${_uniqueIdCounter++}`;
 
   /** Disables all individual radio buttons assigned to this group. */
   private _disabled: boolean = false;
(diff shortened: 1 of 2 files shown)


src/components/radio/radio.ts
```diff
@@ -69,7 +69,10 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
   private _value: any = null;
 
   /** The HTML name attribute applied to radio buttons in this group. */
-  private _name: string = null;
+  private _name: string = `md-radio-group-${_uniqueIdCounter++}`;
+
+  /** @internal */
+  nameEmitter: EventEmitter<string> = new EventEmitter<string>();
```
jelbourn: Instead of this, could each radio pull the group's name in its own `ngOnInit`, and then the group can update them all whenever its name changes (in the setter)?

src/components/radio/radio.ts
```diff
@@ -283,13 +275,30 @@ export class MdRadioButton implements OnInit {
...
+  private _listenToRadioGroup() {
+    // name from `md-radio-group` overrides button name
+    this.name = this.radioGroup.name;
+    this.radioGroup.nameEmitter.subscribe((name: string) => {
+      this.name = name;
+    });
+
+    // listen to value changes on parent
+    this.radioGroup.change.subscribe((change: MdRadioChange) => {
+      this._checked = change.value == this.value;
```
jelbourn: Why can't the group just update this as part of its setter?

src/components/radio/radio.ts
robertmesserle: That could work - that was kind of what I was suggesting when we spoke earlier when I wasn't sure if I liked the approach I took here.

src/components/radio/radio.ts
robertmesserle: Replied to the previous comment - going to look into that as an option and see if it ends up being cleaner.

robertmesserle:
@jelbourn Comments addressed

jelbourn:
LGTM

PR #499, opened 2016-05-23
feat(stepper): add stepper component

Lordnoname:
Add a stepper component, which matches the google material specs pretty good: https://www.google.com/design/spec/components/steppers.html

Types of steppers: 
- [X] Linear Stepper
- [X] Non-Linear Stepper
- [X] Horizontal Stepper
- [ ] Vertical Stepper
- [ ] Mobile Stepper

Types of steps:
- [X] Editable/Non-editable
- [X] Optional

Also it has an swipe animation and navigation buttons at the bottom. Here is a screenshot from my component and one from the specs to show the similarity:

image

image

### DIFF
--- a/src/components/stepper/step.scss
+++ b/src/components/stepper/step.scss
@@ -1,0 +1,27 @@
+.md-step-content {
+  position: absolute;
+  left: 0;
+  right: 0;
+  bottom: auto;
+  opacity: 0;
+  will-change: transform;
+  transition-duration: 0.5s;
+  transition-property: opacity, transform;
+  width: 100%;
+  padding: 24px;
+
+  &.md-active {
+    opacity: 1;
+    position: relative;
+  }
+
+  &.md-right {
+    transform: translate3d(120%, 0, 0);
+    opacity: 0;
+  }
+
+  &.md-left {
+    transform: translate3d(-120%, 0, 0);
+    opacity: 0;
+  }
+}
--- a/src/components/stepper/stepper.ts
+++ b/src/components/stepper/stepper.ts
@@ -1,0 +1,129 @@
+import {
+  Component,
+  Input,
+  ViewEncapsulation,
+  ContentChild,
+  Renderer,
+  ElementRef
+} from '@angular/core';
+import { BooleanFieldValue } from '@angular2-material/core/annotations/field-value';
+import { MdIcon, MdIconRegistry } from '@angular2-material/icon/icon';
+import { NgFormModel } from '@angular/common';
+
+@Component({
+  selector: 'md-stepper',
+  templateUrl: 'stepper.html',
+  styleUrls: ['stepper.css'],
+  providers: [MdIconRegistry],
+  directives: [MdIcon],
+  encapsulation: ViewEncapsulation.None
+})
+export class MdStepper {
+
+  steps: MdStep[] = [];
+
+  @Input() mode: 'linear' | 'nonlinear' = 'nonlinear';
+
+  addStep(step: MdStep): void {
+    if (this.steps.length === 0) step.active = true;
+    this.steps.push(step);
+  }
+
+  //To check if all steps are completed before a specific step
+  allStepsBeforeCompleted(step: MdStep): boolean {
+    for (var i = this.steps.indexOf(this.currentStep); i < this.steps.indexOf(step); i++) {
+      if (!this.steps[i].completed && !this.steps[i].optional) return false;
+    }
+    return true;
+  }
+
+  selectStep(step: MdStep): void {
+    this.currentStep.completed = this.currentStep.valid ? true : false;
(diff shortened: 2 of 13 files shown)


src/components/stepper/stepper.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (this.getNextEditableStep(step)) this.selectStep(this.getNextEditableStep(step));
+  }
+
+  selectPreviousStep(step: MdStep): void {
+    if (this.getPreviousEditableStep(step)) this.selectStep(this.getPreviousEditableStep(step));
+  }
+
+
+  getNextEditableStep(step: MdStep): MdStep {
+    if (this.steps.indexOf(this.currentStep) === this.steps.length - 1) return undefined;
```
DevVersion: Just using `return;` would work as well.

src/components/stepper/stepper.ts
```diff
@@ -0,0 +1,122 @@
...
+    for (var i = this.steps.indexOf(this.currentStep) + 1; i < this.steps.length; i++) {
+      if (this.steps[i].isTargetable) {
+        return this.steps[i];
+      }
+    }
+    return undefined;
```
DevVersion: Here as well. Just using `return;` would work as well - and will be more clear.

src/components/stepper/stepper.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  getPreviousEditableStep(step: MdStep): MdStep {
+    if (this.steps.indexOf(this.currentStep) === 0) return undefined;
+    for (var i = this.steps.indexOf(this.currentStep) - 1; i >= 0; i--) {
```
DevVersion: I think, since those functions (`previous` and `next`) are having a really close logic, we should be able to group this into a single function, which is more generic.

src/components/stepper/step.scss
```diff
@@ -0,0 +1,27 @@
...
+  width: 100%;
+  padding: 24px;
+
+  &.md-active {
+    opacity: 1;
+    position: relative;
+  }
+
+  &.md-right {
+    transform: translate3d(120%, 0, 0);
```
DevVersion: This should be revisited, because using `120%` is quite a random value (as in ng-attack)

src/components/stepper/stepper.ts
```diff
@@ -0,0 +1,122 @@
...
+  get isRight(): boolean {
+    return this.steps.steps.indexOf(this) > this.steps.steps.indexOf(this.steps.currentStep);
+  }
+
+  get isTargetable(): boolean {
+    return !this.editable && this.completed ? false : true;
+  }
+
+  private get _stepperIconLigature(): string {
+    return (!this.active && this.completed) ? 'done' : 'create';
```
DevVersion: Some users, may not have the material design icon font included. I think, (once) this component will be available, the SVG should be inlined.

src/components/stepper/stepper.ts
```diff
@@ -0,0 +1,122 @@
...
+    });
+    step.active = true;
+  }
+
+  get currentStep(): MdStep {
+    return this.steps.filter(step => step.active)[0];
+  }
+
+  selectNextStep(step: MdStep): void {
+    if (this.getNextEditableStep(step)) this.selectStep(this.getNextEditableStep(step));
```
DevVersion: This is actually a repeat of calling the same function. IMO a variable which holds the step object temporary is more appropriated. Otherwise we would run the iteration twice each time.

src/components/stepper/stepper.ts
```diff
@@ -0,0 +1,122 @@
...
+    this.steps.push(step);
+  }
+
+  selectStep(step: MdStep): void {
+    this.currentStep.completed = this.currentStep.valid ? true : false;
+    if (this.mode === 'linear') {
+      if (!this.currentStep.valid) return;
+      //Checks if all steps before are completed or optional
+      for (var i = this.steps.indexOf(this.currentStep); i < this.steps.indexOf(step); i++) {
+        if (!this.steps[i].completed && !this.steps[i].optional) return;
```
DevVersion: I think this logic, should have its own public method, because it may be useful for the public API.

> To check if all steps are completed before a specific index / step

jelbourn:
Hi @Lordnoname, we appreciate your PR, but we have a formal process for adding large features like this in our [contributing guide](https://github.com/angular/material2/blob/master/CONTRIBUTING.md#-want-a-feature).

Please open an issue for discussion, and from there we can go into a design doc phase. Any new components need to have a [design doc](https://github.com/angular/material2/wiki/Design-doc-directory) before moving on to implementation so that we're on the same page for API and implementation. 

After the design doc phase we can come back to a PR. Let me know if you have any questions!

PR #505, opened 2016-05-24
chore(): use core-js instead of es6-shim

### LINKED ISSUE #477: Switch from es6-shim to core-js for demo-app and e2e app
See https://github.com/angular/angular/issues/5755

DevVersion:
Replace es6-shim, es6-promise and reflect-metadata with core-js, because it coveres all required dependencies / shims in one single dependency / library.

@jelbourn - Not sure, with replacing the reflect-metadata dependency with core-js (https://github.com/zloirock/core-js/issues/152).

Closes #477.


jelbourn:
LGTM

PR #515, opened 2016-05-25
fix: use the right ID for the label.

### LINKED ISSUE #512: md-input does not label the input with the placeholder text when using VoiceOver
When visiting our demo-app page for md-input using VoiceOver on OSX, the placeholder text for the inputs is not used to label the inputs. When moving focus to the `<input>`, the reader simple announces "edit text blank". This occurs with Safari, Chrome, and Firefox.

hansl:
Fixes #512.


jelbourn:
LGTM

PR #517, opened 2016-05-25
Add Initial Toggle Button

iveysaur:
R: @jelbourn @robertmesserle @kara 

This is a start for the [toggle buttons](https://www.google.com/design/spec/components/buttons.html#buttons-toggle-buttons). Currently supports exclusive and multiple selection with styles and the disabled state. Exclusive selection has a value which is set to the selected toggle.

Still needs:
- A value for multiple selection
- Focus states
- Keyboard selection
- NgModel support

[Design Doc] (https://docs.google.com/document/d/1WwTKHdXZY51FzK1uGldM25OeXqY83STqURWdo0-jAKc/edit?usp=sharing)


iveysaur:
I signed it!

src/components/toggle/toggle.html
```diff
@@ -0,0 +1,13 @@
+<label [attr.for]="inputId" class="md-toggle-label">
+  <input #input class="md-toggle-input"
+         [type]="type"
+         [id]="inputId"
+         [checked]="checked"
+         [disabled]="disabled"
+         [name]="name"
+         (change)="onClick()"/>
```
jelbourn: We generally haven't been using the `/` at the end of void elements.

src/components/toggle/toggle.html
```diff
@@ -0,0 +1,13 @@
+<label [attr.for]="inputId" class="md-toggle-label">
```
jelbourn: I'm thinking we should probably rename from just `toggle` to `toggle-button`, which is more explicit. Anyone else have thoughts on this?

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,42 @@
+@import "elevation";
+@import "default-theme";
+@import "palette";
+@import "mixins";
+
+$md-padding: 0 16px !default;
+$md-line-height: 36px !default;
+$md-border-radius: 3px !default;
```
jelbourn: These variables should have `toggle` (or `toggle-button` if we change it) in their name, e.g., `$md-toggle-button-padding`.

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,42 @@
...
+  .md-toggle-input {
+    @include md-visually-hidden;
+  }
+
+  .md-toggle-label-content {
+    display: inline-block;
+    line-height: $md-line-height;
+    padding: $md-padding;
+    cursor: pointer;
+  }
```
jelbourn: Generally try to avoid nesting, as it increases the specificity of the rules.

src/components/toggle/toggle.html
iveysaur: I think it makes sense as there is also the toggle slider, I'm fine with changing it but will wait for a go-ahead.

src/components/toggle/toggle.html
iveysaur: Got it from radio, sorry about that! Should there be a space between `"` and `>` or does it not matter?

src/components/toggle/toggle.scss
iveysaur: Got it! I got wrapped up in the "do ALL the sass things!" mindset 😛

src/components/toggle/toggle.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  /** Disables all toggles in the group. */
+  private _disabled: boolean = null;
+
+  /** The currently selected toggle, should match the value. */
+  private _selected: MdToggle = null;
+
+  /** Event emitted when the group's value changes. */
+  @Output()
+  change: EventEmitter<MdToggleChange> = new EventEmitter<MdToggleChange>();
```
jelbourn: The other components still need to be updated to do this, but the `EventEmitter` should be privite while exposing an `Observable` on the public API:
```ts
private _change: EventEmitter<MdToggleChange> = new EventEmitter<MdToggleChange>();
@Output() get change(): Observable<MdToggleChange> { 
  return this._change.asObservable();
}
```

src/components/toggle/toggle.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  /**
+   * Initializes the names once content children are available.
+   * This allows us to propagate relevant attributes to associated toggles.
+   */
+  ngAfterContentInit() {
+    if (this.name == null) {
+      this.name = `md-toggle-group-${_uniqueIdCounter++}`;
+    } else {
+      this._updateChildToggleNames();
```
jelbourn: I believe @robertmesserle has a PR that changes this behavior in the radio button since it was causing problems.

src/components/toggle/toggle.ts
```diff
@@ -0,0 +1,332 @@
...
+  set value(newValue: any) {
+    if (this._value != newValue) {
+      this._value = newValue;
+
+      this._updateSelectedToggleFromValue();
+      this._emitChangeEvent();
+    }
+  }
+
+  private _updateSelectedToggleFromValue(): void {
```
jelbourn: I'd like to find a way to extract this (and similar methods) into something resuable from both this component and radio-button, since they're effectively identical. Already the `radio.ts` version of this method has changed, so they're definitely get out of sync if they duplicate logic.

src/components/toggle/toggle_dispatcher.ts
```diff
@@ -0,0 +1,27 @@
...
+
+
+/**
+ * Class for toggles to coordinate unique selection based on name.
+ * Intended to be consumed as an Angular service.
+ * This service is needed because native radio change events are only fired on the item
+ * currently being selected, and we still need to uncheck the previous selection.
+ */
+@Injectable()
+export class MdToggleDispatcher {
```
jelbourn: I think we should be able to rename the `MdRadioDispatcher` to something like `MdUniqueSelectionDispatcher`, move it to `core/` and use it in both components.

src/demo-app/toggle/toggle-demo.ts
```diff
@@ -0,0 +1,12 @@
+import {Component} from '@angular/core';
+import {MdToggle, MdToggleGroup, MdToggleGroupMultiple} from '@angular2-material/toggle/toggle';
+import {MdToggleDispatcher} from '@angular2-material/toggle/toggle_dispatcher';
+
+@Component({
+  moduleId: module.id,
+  selector: 'toggle-demo',
+  templateUrl: 'toggle-demo.html',
+  providers: [MdToggleDispatcher],
+  directives: [MdToggle, MdToggleGroup, MdToggleGroupMultiple]
```
jelbourn: Should use the `MD_TOGGLE_DIRECTIVES` in the demo

src/components/toggle/toggle.html
jelbourn: Good point- `button-toggle` aligns better with `slide-toggle`

iveysaur:
All comments addressed.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+
+var _uniqueIdCounter = 0;
+
+/** A simple change event emitted by either MdButtonToggle or MdButtonToggleGroup. */
+export class MdButtonToggleChange {
+  source: MdButtonToggle;
+  value: any;
+}
+
+/** Exclusive selection button toggle group. */
```
jelbourn: I'd expand this comment to include something like "Behaves like a radio-button group."

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  /** Event emitted when the group's value changes. */
+  private _change: EventEmitter<MdButtonToggleChange> = new EventEmitter<MdButtonToggleChange>();
+  @Output() get change(): Observable<MdButtonToggleChange> {
+    return this._change.asObservable();
+  }
+
+  /** Child button toggle buttons. */
+  @ContentChildren(forwardRef(() => MdButtonToggle))
+  private _buttonToggles: QueryList<MdButtonToggle> = null;
```
jelbourn: Looks like `_buttonToggles` is currently unused, which lint will probably complain about.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+  templateUrl: 'button-toggle.html',
+  styleUrls: ['button-toggle.css'],
+  encapsulation: ViewEncapsulation.None
+})
+export class MdButtonToggle implements OnInit {
+  /** Whether or not this button toggle is checked. */
+  private _checked: boolean = false;
+
+  /** Type of the button toggle. Either 'radio' or 'checkbox'. */
+  private _type: string = null;
```
jelbourn: TypeScript has a cool feature that would be useful here: string literal types. You can do this:
```ts
private _type: 'checkbox' | 'radio';
```
And it will enforce via the compiler that the only values allowed for the property are `checkbox` and `radio`.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  /** The unique ID for this button toggle. */
+  @HostBinding('id')
```
jelbourn: You don't need to specify `id` in the `@HostBinding()` if it exactly matches the name of the property

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+  change: EventEmitter<MdButtonToggleChange> = new EventEmitter<MdButtonToggleChange>();
+
+  constructor(@Optional() toggleGroup: MdButtonToggleGroup,
+              @Optional() toggleGroupMultiple: MdButtonToggleGroupMultiple,
+              public buttonToggleDispatcher: MdUniqueSelectionDispatcher) {
+    this.buttonToggleGroup = toggleGroup;
+
+    this.buttonToggleGroupMultiple = toggleGroupMultiple;
+
+    buttonToggleDispatcher.listen((id: string, name: string) => {
```
jelbourn: You should only need to do this if this is a radio-flavor toggle.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+      if (id != this.id && name == this.name) {
+        this.checked = false;
+      }
+    });
+  }
+
+  ngOnInit() {
```
jelbourn: Mark the `ngOnInit` method as `/** @internal /*`

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+      this.id = `md-button-toggle-${_uniqueIdCounter++}`;
+    }
+
+    if (this.buttonToggleGroup && this._value == this.buttonToggleGroup.value) {
+      this._checked = true;
+    }
+
+    // Even if there is no group at all, treat the button toggle as a checkbox so it can be
+    // toggled on or off.
+    if (this.buttonToggleGroup) {
```
jelbourn: You should be able to do this check in the constructor since you'll know at that point which type this toggle is.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  @HostBinding('class.md-button-toggle-checked')
+  @Input()
+  get checked(): boolean {
+    return this._checked;
+  }
+
+  set checked(value: boolean) {
+    if (value && (this.buttonToggleGroup)) {
```
jelbourn: I would add a getter like `isSingleSelection` and use that whenever you need to check if this is a radio-flavor toggle, since it's more semantically readable than checking the presence of the toggle group.

src/components/button-toggle/button-toggle.ts
iveysaur: This one is used, I removed the unused one in the MdButtonToggleGroupMultiple though.

iveysaur:
@jelbourn: Addressed comments.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,339 @@
...
+
+  get isSingleSelector(): boolean {
+    return this._isSingleSelector;
+  }
+
+  set isSingleSelector(value: boolean) {
+    this._isSingleSelector = value;
+  }
+
+  @HostListener('click', ['$event'])
```
jelbourn: This will actually be a problem in server-side rendering environments, since having a decorator will make TypeScript preserve the type information at run-time. On the server environment, `Event` won't be available as a global. Check out what `radio.ts` has now instead.

src/components/button-toggle/button-toggle.html
```diff
@@ -0,0 +1,13 @@
+<label [attr.for]="inputId" class="md-button-toggle-label">
+  <input #input class="md-button-toggle-input"
+         [type]="type"
+         [id]="inputId"
+         [checked]="checked"
+         [disabled]="disabled"
+         [name]="name"
+         (change)="onClick()">
```
robertmesserle: Should this method be called `onClick` if it is used to handle change events?

iveysaur: It was causing odd issues with the radio vs the checkbox but I just fixed that in another branch, I'll move it over to this PR though.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,341 @@
...
+  get isSingleSelector(): boolean {
+    return this._isSingleSelector;
+  }
+
+  set isSingleSelector(value: boolean) {
+    this._isSingleSelector = value;
+  }
+
+  /** @internal */
+  onClick(event: Event) {
```
hansl: Could we rename this as to what it does? `onClick` could be a number of things. `toggle()` would probably make more sense and could be public.

iveysaur: Changed in a follow-up PR (to onInputChange), had to make other changes to the function anyway + tests so we determined it would be better as a follow-up.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,341 @@
...
+    this._checked = value;
+  }
+
+  get type(): ToggleType {
+    return this._type;
+  }
+
+  set type(value: ToggleType) {
+    this._type = value;
+  }
```
DevVersion: Is it really necessary to have a getter and setter here?
You're just setting the value of the private property. It would be the same, to have a normal property here.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,331 @@
...
+
+  /** The parent button toggle group (exclusive selection). Optional. */
+  buttonToggleGroup: MdButtonToggleGroup;
+
+  /** The parent button toggle group (multiple selection). Optional. */
+  buttonToggleGroupMultiple: MdButtonToggleGroupMultiple;
+
+  /** Event emitted when the group value changes. */
+  @Output()
+  change: EventEmitter<MdButtonToggleChange> = new EventEmitter<MdButtonToggleChange>();
```
jelbourn: Should use the same pattern for the `@Output` here as you did on the group.

jelbourn:
LGTM aside from one minor nit on the `@Output`

iveysaur:
@jelbourn: addressed comment.

src/components/button-toggle/button-toggle.spec.ts
```diff
@@ -0,0 +1,346 @@
...
+    let groupNativeElement: HTMLElement;
+    let buttonToggleDebugElements: DebugElement[];
+    let buttonToggleNativeElements: HTMLElement[];
+    let groupInstance: MdButtonToggleGroup;
+    let buttonToggleInstances: MdButtonToggle[];
+    let testComponent: ButtonTogglesInsideButtonToggleGroup;
+
+    beforeEach(async(() => {
+      builder.createAsync(ButtonTogglesInsideButtonToggleGroup).then(f => {
+        fixture = f;
```
kara: Why not name the arg `fixture` rather than assigning it here?

iveysaur: `fixture` is declared on the outer scope.

src/components/button-toggle/button-toggle.spec.ts
```diff
@@ -0,0 +1,346 @@
...
+      fixture.detectChanges();
+
+      buttonToggleNativeElements[0].click();
+      expect(buttonToggleInstances[0].checked).toBe(false);
+    });
+
+    it('should update the group value when one of the toggles changes', () => {
+      expect(groupInstance.value).toBeFalsy();
+      let nativeCheckboxLabel =
+          <HTMLLabelElement> buttonToggleNativeElements[0].querySelector('label');
```
kara: Is it possible for you to use the `fixture.debugElement.query(By.css('label))` for this instead?  If we ever run the tests in a different context, it wouldn't work.

src/components/button-toggle/button-toggle.spec.ts
kara: Oh, I see. Didn't see this was in the beforeEach.

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,333 @@
...
+}
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-button-toggle',
+  templateUrl: 'button-toggle.html',
+  styleUrls: ['button-toggle.css'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class MdButtonToggle implements OnInit {
```
kara: Have you considered breaking this out into its own file?  Current file is getting pretty long, and it will be easier to find.

jelbourn: There are circular dependencies between the button-toggle and the group.

kara: Ah, okay.

kara:
LGTM

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,333 @@
...
+    return this._name;
+  }
+
+  set name(value: string) {
+    this._name = value;
+
+    this._updateButtonToggleNames();
+  }
+
+  @Input()
```
hansl: `@BooleanFieldValue()` ?

src/components/button-toggle/button-toggle.ts
```diff
@@ -0,0 +1,333 @@
...
+  templateUrl: 'button-toggle.html',
+  styleUrls: ['button-toggle.css'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class MdButtonToggle implements OnInit {
+  /** Whether or not this button toggle is checked. */
+  private _checked: boolean = false;
+
+  /**
+   * @internal
```
hansl: The convention inside Google is to put the annotations at the end of the comment.

src/components/button-toggle/button-toggle.ts
hansl: seems to me the toggle doesn't need a dependency to its group and that would simplify its logic. The group could `ContentChildren` and listen to change events and coordinate its own state.

It feels weird that the toggle is also responsible for changing the state of the group, and by clearing up responsibility you could definitely set `ChangeDetectionStrategy` to `OnPush`.

src/components/button-toggle/button-toggle.scss
```diff
@@ -0,0 +1,46 @@
...
+
+.md-button-toggle-label-content {
+  display: inline-block;
+  line-height: $md-button-toggle-line-height;
+  padding: $md-button-toggle-padding;
+  cursor: pointer;
+}
+
+md-button-toggle md-icon {
+  vertical-align: middle;
```
robertmesserle: Would we want to do this for all direct children?  Or just md-icon?

iveysaur: I don't have a strong opinion, I just wanted to be safe for this iteration when I wrote that.

src/components/button-toggle/button-toggle.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+    it('should emit a change event from the button toggle group', fakeAsync(() => {
+      expect(groupInstance.value).toBeFalsy();
+
+      let changeSpy = jasmine.createSpy('button-toggle-group change listener');
+      groupInstance.change.subscribe(changeSpy);
+
+      groupInstance.value = 'test1';
+      fixture.detectChanges();
+      tick();
+      expect(changeSpy).toHaveBeenCalled();
```
robertmesserle: Do we want to be more specific  here?  ie. `expect(changeSpy).toHaveBeenCalledWith('test1')`

src/components/button-toggle/button-toggle.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+      let nativeRadioInput = buttonToggleDebugElements[0].query(By.css('input')).nativeElement;
+
+      nativeRadioInput.click();
+      fixture.detectChanges();
+
+      expect(buttonToggleInstances[0].checked).toBe(true);
+      expect(groupInstance.value);
+    });
+
+    it('should emit a change even from button toggles', fakeAsync(() => {
```
robertmesserle: *event

Issue #531, opened 2016-05-26
Add stylelint to CI


jelbourn:
I'd like to try out running [stylelint](http://stylelint.io/) in our existing lint job on Travis CI. As a simple getting started point, the only rule applied should be [no id selectors](http://stylelint.io/user-guide/rules/selector-no-id/).

jelbourn:
Closed via #532

PR #532, opened 2016-05-26
Stylelint

sendilkumarn:
#531 

Added this in travis too. Let me check if it is running without any failure.


.stylelintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "selector-no-id":true,
```
jelbourn: Missing space after colon

.stylelintrc
```diff
@@ -0,0 +1,5 @@
+{
```
jelbourn: Rather than call this file the default `.stylelintrc`, I'd do `stylelint-config.json` so that IDEs can actually syntax highlight it.

scripts/ci/style-linter.js
```diff
@@ -0,0 +1,143 @@
+'use strict';
```
jelbourn: I don't this this script is really necessary; the `stylelint` cli tool is pretty fast (I just tried it), so we should be able to treat it the same we we currently treat `tslint`, which is to just have an npm script like
```json
"stylelint" : "stylelint 'src/**/*.scss' --syntax scss",
```
(plus specifying the config file)

sendilkumarn:
@jelbourn 👍

jelbourn:
@sendilkumarn looks good, just needs a rebase on master

sendilkumarn:
rebased :)

jelbourn:
LGTM

jelbourn:
Thanks!

Issue #519, opened 2016-05-25
Determinate progress-bar broken in @angular rc1 #430 [Still not working]


Pixcell:
#### Bug, feature request, or proposal:

Bug (https://github.com/angular/material2/issues/430)

I did everything like you said before closing the issue 430, but it is still not working (cf my last response)
I think you don't see comment notifications on closed issues, so I open a new one.

#### What is the expected behavior?

Progress bar to be filled with correct percentage

#### What is the current behavior?

![screen shot 2016-05-17 at 08 48 40](https://cloud.githubusercontent.com/assets/4005291/15313687/2c478302-1c0d-11e6-940e-9a7230c567af.png)

![screen shot 2016-05-17 at 08 56 05](https://cloud.githubusercontent.com/assets/4005291/15313694/3e3e94b0-1c0d-11e6-9b73-c393c79660af.png)

#### What are the steps to reproduce?

```JS
import {MdToolbar} from '@angular2-material/toolbar';
import {MdButton} from '@angular2-material/button';
import {MdCard} from '@angular2-material/card';
import {MdProgressBar} from '@angular2-material/progress-bar';

@Component({
  selector: 'home',
  templateUrl: 'components/home/home.component.html',
  styleUrls: ['components/home/home.component.css'],
  providers: [],
  directives: [MdToolbar, MdButton, MdCard, MdProgressBar]
})
export class HomeComponent implements OnInit{
```

``` HTML
<p>
     <span>PID completion {{h.pid.percentage}}%</span><span *ngIf="h.pid.warning">{{h.pid.warning}}        </span>
    <md-progress-bar mode="determinate" [value]="h.pid.percentage"></md-progress-bar>
</p>
```

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: http://plnkr.co/edit/o077B6uEiiIgkC0S06dd

#### Which versions of Angular, Material, OS, browsers are affected?

```JS
    "@angular/common": "^2.0.0-rc.1",
    "@angular/compiler": "^2.0.0-rc.1",
    "@angular/core": "^2.0.0-rc.1",
    "@angular/http": "^2.0.0-rc.1",
    "@angular/platform-browser": "^2.0.0-rc.1",
    "@angular/platform-browser-dynamic": "^2.0.0-rc.1",
    "@angular/router": "^2.0.0-rc.1",
    "@angular/router-deprecated": "^2.0.0-rc.1",
    "@angular2-material/button": "^2.0.0-alpha.4",
[...]

PR #542, opened 2016-05-27
Determinate progress-bar fix

sendilkumarn:
`style:transform` currently not working it seems [as per comments in sidenav.ts]. I have changed it to ngStyle and returned scale transform value as an object. Now determinate progress bar is working. 

#519


jelbourn:
LGTM, thanks for investigating this!

PR #545, opened 2016-05-27
fix(grid-list): account for gutter in total list height

kara:
r: @jelbourn 

Caught this while writing unit tests:  the height of the list wasn't taking into account the size of the gutters, so heights were a pixel or two off.    This corrects the height.


jelbourn:
LGTM

PR #551, opened 2016-05-27
fix(): visual hidden inputs should not bubble change event

### LINKED ISSUE #544: checkbox: change event emits [Event] instead of boolean
Sometimes, when triggering the change event on the `input`, by clicking on it, the change output will emit the actual event object, instead of the correct boolean.
- https://developer.mozilla.org/de/docs/Web/Events/change

Already working on a PR, which includes some tests and fixes it.

Sample test (without fix)
image

After searching in the Angular Repository I found the right issue
- https://github.com/angular/angular/issues/4059

DevVersion:
* Currently the change event of the visual hidden inputs will bubble up and emit its event object to the component's `change` output.

This is currently an issue of Angular 2
- See https://github.com/angular/angular/issues/4059

To prevent the events from bubbling up, we have to stop propagation on
change.

Fixes #544.

### DIFF
--- a/src/components/checkbox/checkbox.spec.ts
+++ b/src/components/checkbox/checkbox.spec.ts
@@ -286,7 +286,27 @@ describe('MdCheckbox', () => {
       fixture.detectChanges();
 
       tick();
-      expect(testComponent.handleChange).toHaveBeenCalled();
+
+      expect(testComponent.handleChange).toHaveBeenCalledTimes(1);
+      expect(testComponent.handleChange).toHaveBeenCalledWith(true);
+    }));
+
+    it('should not emit a DOM event to the change output', async(() => {
+      expect(testComponent.handleChange).not.toHaveBeenCalled();
+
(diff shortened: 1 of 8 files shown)


src/components/checkbox/checkbox.spec.ts
```diff
@@ -521,8 +541,8 @@ class CheckboxWithNameAttribute {}
 /** Simple test component with change event */
 @Component({
   directives: [MdCheckbox],
-  template: `<md-checkbox (change)="handleChange()"></md-checkbox>`
+  template: `<md-checkbox (change)="handleChange($event)"></md-checkbox>`
 })
 class CheckboxWithChangeEvent {
-  handleChange(): void {}
+  handleChange(event: any): void {}
```
jelbourn: The type should be `Event`

jelbourn:
FYI @iveysaur for button-toggle

jelbourn:
LGTM

Issue #553, opened 2016-05-27
Automatically create a live preview for every PR


hansl:
It might be a firebase server, or AppEngine.  I would consider building a docker on Travis as a Part I and then sending that docker to a server as Part 2.

jelbourn:
One possible approach:
1) Add GitHub hook to remote server when a PR is opened / updated
2) Remote server pulls the PR, builds, copies build output to a directory and serves it.
3) Bot posts to the PR with a link to the preview

DevVersion:
So, I'm working on this issue.

The following steps will be executed:
- Waiting for Github Webhooks
- Applies PR patch to a extra branch for the PR (`preview/pr-517`)
- Builds Distribution with Angular CLI
- Bundles **required** dependencies into a bundle (accelerates firebase deploy and load time significant) 
- Modifies the `index.html` with a virtual DOM, to use the bundle file.
- Removes unnecessary vendors (e.g ´@angular` - it's now in the bundle)
- Deploys new change to the Firebase Hosting
- Post's link to the PR preview to the PR.

I've already finished this and it's working very well.

I saved a lot of time, by building / enhancing the internal Angular Bot for Material 1 (in progress)

@jelbourn If you wan't to chat about the implementation or how we should continue, let me know :)

DevVersion:
Here is a live preview, how it currently looks like. 

It has a build time of ~ 1-2 minutes (*minus* wasted time for manually executing the hook).

- https://github.com/DevVersion/material2/pull/2

Notice, how the bot posts the link of the automatically built preview. (Message is just temporary)

jelbourn:
@DevVersion wow, that was fast. 

> Bundles required dependencies into a bundle (accelerates firebase deploy and load time significant)

Are you just running `npm install` on each PR to get the latest clean dependencies?

Overall sounds great. Once you make a PR, I can spin up a new firebase instance and we can simply point the code to that.

Issue #533, opened 2016-05-26
Packages don't contain scss files


VFedyk:
I have installed 2.0.0-alpha-5 and there is no SCSS files. What was the reason of doing that?

kreed-rmn:
I found the reason for this was node-sass hadn't been recompiled. 

`npm rebuild node-sass` fixed it for me.

VFedyk:
It is how package looks for me now:
![list-2 0 0-alpha 5 tgz _126](https://cloud.githubusercontent.com/assets/489576/15579029/8ae007ee-236b-11e6-8218-d6a366ce2083.png)

VFedyk:
The same list package but earlier version:
![list-2 0 0-alpha 4 1 tgz _127](https://cloud.githubusercontent.com/assets/489576/15579162/17f64026-236c-11e6-868d-9b4b0e46114b.png)

Delagen:
I am waiting for theming solution. Current implementation so ugly, in state of npm is even impossible. Sad that no any issue with plan for distribution with theme support and bundling into single css file in case that shadow css isn.t the best solution in some cases

zigzag95:
I added an issue but Mister @jelbourn closed it without any explanation.

jelbourn:
The missing scss files wasn't intentional; it was likely due to changes in the angular-cli. We'll look into the problem.

As far as a real theming API: it will come eventually, but in the short term we're focused on having components _to_ theme 😃. Generally issues asking "When will X be done?" will just be closed.

zigzag95:
I just asked to add a milestone for theming. To know if we workaround it or wait for it.

VFedyk:
@jelbourn, maybe it wasn't intentional but they were in at least two last versions and we built project relying on these files. And then, suddenly, they disappeared. Is it possible to rebuild recent packages?

jelbourn:
Just published an alpha.5-2 which contains the scss source.

TheCaptainDate:
@jelbourn 
Still broken.
There is css and scss files now with same name, so @import inside material2 scss files don't know which file should be imported, and returns an error. 
I think this can be fixed if you put extensions in material2 scss files, or remove all css files from package.

Delagen:
While no any distribution policy It is fun. Because all styles are embedded in source ((

ddjonline:
What is the recommended way to handle the scss and css in the same directories with the node-sass option (I know the angular-cli "just works" but that isn't an option for some)?

PR #566, opened 2016-05-28
fix(demo-app): resolve mapping paths in relative.

DevVersion:
* Currently, the demo app is mapping its core and components absolutely (`/`), which is working perfectly for the served application
* This commit changes the absolute mapping paths to relative paths.<br/>
  This allows developers, to serve the demo-app in sub-directories (not only in the root).

Notice, that all other mappings in the configurations are also relative, so it will also restore consistency.

This is also required for some bundlers, for example SystemJS Builder is not able to build a bundle of the demo app, with absolute paths, because the baseURL may have changed.

This is a required commit, for the Live Preview (#533), because I'm planning to bundle the output, to accelerate deploy and load time.

References #553

### DIFF
--- a/src/demo-app/system-config.ts
+++ b/src/demo-app/system-config.ts
@@ -25,9 +25,9 @@ const components = [
 
 /** Map relative paths to URLs. */
 const map: any = {
-  '@angular2-material/core': '/core',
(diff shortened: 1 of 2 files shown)


hansl:
LGTM.

Issue #575, opened 2016-05-29
 MdCheckBox is behaving different to native checkbox


mmrath:
#### Bug, feature request, or proposal: Bug
Links for both native and MdCheckBox version below

#### What is the expected behavior?
Native Checkbox version
http://plnkr.co/edit/cZbFWOlCeZuj7YdUczPm?p=preview

#### What is the current behavior?
This is with MdCheckBox
http://plnkr.co/edit/FFpqdNhQfvbUynbPE1ZD?p=preview

#### What are the steps to reproduce?

1. Click on the top select all check box for 'Write'
2. Now click (reset) Write for one of the resource

Native version has the desired behaviour that is, unselect the top most checkbox if one of the below one is unchecked. But with MdCheckBox the change to reset top most checkbox triggers on change event, which is resetting all the below checkboxes.

I don't know which one is correct behaviour or both are correct.

Any help to get the MdCheckBox version working is appreciated.

#### What is the use-case or motivation for changing an existing behavior?
Get MdCheckBox behave as close to native checkbox as possible

#### Which versions of Angular, Material, OS, browsers are affected?
alpha 5

#### Is there anything else we should know?
Bo

sendilkumarn:
Interesting also, check this out

http://plnkr.co/edit/nDWK7EmmCBgMkHIKRIg7?p=preview

the checkbox click function is called twice, why so?

zahlprish:
@sendilkumarn I have experienced the double fires since the last release (alpha 5).

sendilkumarn:
@jelbourn is this correct, I will submit a PR if so 

Add a method for click event in checkbox. we can re use the checkbox change method itself. This kinda solves problem.

jelbourn:
It seems that the issue here isn't a double-firing `change`, but is instead caused by different behavior between native inputs and our checkbox for _when_ change events should be fired. Native inputs only fire a change event when they have user focus. The `MdCheckbox` fires a change event whenever it changes for any reason. I've actually had a note written about this to discuss with the rest of the team for a couple weeks now.

sendilkumarn:
Yeah agree to that. But clicks firing two times is also an issue. Check #613

Aim is to prevent double firing of click event

sendilkumarn:
one more weird behavior with respect to checkbox. Is this even correct or am I missing out anything.

click / change one - just printing ng model value =>  returns true / false
click / change  two -  just printing ng model value =>  returns MDCheckboxChange Object

is this expected?

[...]

PR #582, opened 2016-05-30
checkbox fix for #575

sendilkumarn:
#575

### DIFF
--- a/src/components/checkbox/checkbox.html
+++ b/src/components/checkbox/checkbox.html
@@ -11,7 +11,8 @@
            [attr.aria-labelledby]="ariaLabelledby"
            (focus)="onInputFocus()"
            (blur)="onInputBlur()"
-           (change)="onInteractionEvent($event)">
+           (change)="onInteractionEvent($event)"
+           (click)="onInteractionEvent($event)">
     <div class="md-ink-ripple"></div>
     <div class="md-checkbox-frame"></div>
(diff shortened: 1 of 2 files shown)


jelbourn:
@sendilkumarn Can you title your PRs such that it's clear what they're doing, and include some description that has a high level summary of your changes and why they fix the issue? Seeing just "checkbox fix" makes it more difficult to review.

src/components/checkbox/checkbox.html
```diff
@@ -11,7 +11,8 @@
            [attr.aria-labelledby]="ariaLabelledby"
            (focus)="onInputFocus()"
            (blur)="onInputBlur()"
-           (change)="onInteractionEvent($event)">
+           (change)="onInteractionEvent($event)"
+           (click)="onInteractionEvent($event)">
```
jelbourn: This doesn't seem right. What is the underlying bug you're trying to address?

jelbourn:
This change will also need a unit test.

src/components/checkbox/checkbox.html
sendilkumarn: upon clicking, checkbox fires click function twice. Ideally it should fire it once right?

sendilkumarn:
For PR - I will :)

Unit Test, I totally forgot. I will add it by today.

sendilkumarn:
I will close this PR and add another.

Issue #557, opened 2016-05-27
MdCheckbox does not render opacity correctly on Safari


chrismcgregor:
MdCheckbox has irregular behavior for Safari browser (cannot select).

sendilkumarn:
We can select it in safari, but the opacity in transition is not setting properly.

@jelbourn  I think we have to use `transition-timing-function` here. That is also supported by all latest versions of browsers.

Tested in Firefox / Opera / Chrome and Safari.  Seems to work fine.

jelbourn:
@chrismcgregor Please follow the issue template for future issues.

jelbourn:
Should be fixed in #594

chrismcgregor:
Excellent, thank you!

PR #588, opened 2016-05-31
transition to transition-timing-function for beizer

sendilkumarn:
#557


src/components/checkbox/checkbox.scss
```diff
@@ -270,7 +270,7 @@ md-checkbox {
   opacity: 0;
   display: inline-flex;
   justify-content: center;
-  transition: background-color $md-checkbox-transition-duration
+  transition-timing-function: background-color $md-checkbox-transition-duration
```
jelbourn: I don't think this is the right syntax for `transition-timing-syntax` (including the `background-color` in the value)

PR #590, opened 2016-05-31
adds stylelint config settings

robertmesserle:
R: @jelbourn @kara @hansl @iveysaur


stylelint-config.json
```diff
@@ -1,5 +1,59 @@
 {
   "rules": {
+    "color-hex-case": "lower",
+    "color-no-invalid-hex": true,
```
jelbourn: Man, now I won't be able use to [`chucknorris`](http://stackoverflow.com/questions/8318911/why-does-html-think-chucknorris-is-a-color) as a color any more.

jelbourn:
LGTM, but @kara and @hansl should also chime in before we merge.

stylelint-config.json
```diff
@@ -1,5 +1,59 @@
 {
   "rules": {
+    "color-hex-case": "lower",
```
kara: Why lower?

stylelint-config.json
```diff
@@ -1,5 +1,59 @@
...
+    "function-name-case": "lower",
+    "function-url-quotes": "always",
+    "function-whitespace-after": "always",
+
+    "number-leading-zero": "always",
+    "number-no-trailing-zeros": true,
+    "number-zero-length-no-unit": true,
+
+    "string-no-newline": true,
+    "string-quotes": "double",
```
kara: This is a pretty big change. Reasoning?

hansl: So we're okay differing between TS and style quotes?  I think that will be confusing switching between the two.

src/demo-app/gestures/gestures-demo.scss
```diff
@@ -3,5 +3,5 @@
   width: 200px;
   background: gray;
   padding: 15px;
-  color: #FFF;
-}
\ No newline at end of file
+  color: white;
```
hansl: Are we okay with this?  `#FFF` != `white`.

jelbourn: I don't understand what you mean

stylelint-config.json
jelbourn: We already use double-quotes in html files. I don't care which one we do, Rob said it was mixed ~50/50 already.

hansl: My vote goes to consistency; we could also use single quotes in HTML.

kara: Doesnt tslint already force single quotes? Besides HTML, I'd be surprised if there are any other double quotes.

src/demo-app/gestures/gestures-demo.scss
hansl: :shrug: I meant it was smaller. I don't think it's that important after thinking about it.

stylelint-config.json
```diff
@@ -1,5 +1,59 @@
...
+
+    "unit-case": "lower",
+    "unit-no-unknown": true,
+    "unit-whitelist": ["px", "%", "deg", "s", "em"],
```
hansl: Maybe I'm alone in this, but `250ms` is easier to read than `0.25s`. I agree we need a single time unit though.

kara: Agree with Hans on this one

stylelint-config.json
jelbourn: tslint enforces single quotes... in ts. Personally I'd be quite opposed to switching html to single quotes. I have the feeling that double is much more prevalent in html, though I don't have anything to back that claim up with right now. For CSS I don't care which we use, but I think I naturally tend to write single-quotes.

hansl: We could change tslint to double-quotes, but I'd think it's a mistake. My vote goes to single quote in CSS and TS. I don't care much about HTML, since inline templates are simpler with double quotes HTML attributes.

kara: Yeah, my vote would be single quotes everywhere but in HTML.

stylelint-config.json
jelbourn: Let's switch to just `ms` for now and see if we ever have a case that breaks into seconds.

stylelint-config.json
robertmesserle: Easier to type and was more common in the existing code.  Also personal preference - much easier to type.

stylelint-config.json
robertmesserle: I'm fine with single quotes - I think I went by existing code but it was very mixed.

stylelint-config.json
robertmesserle: I don't feel too strongly one way or the other.  It's a question of precision.  Most of my animations only ever require one decimal place, so it's always shorter than ms: `0.5s` vs `500ms`.

robertmesserle:
@jelbourn @kara @hansl Switched quotes to single and switched time to ms

src/core/style/_mixins.scss
```diff
@@ -7,12 +7,12 @@
...
 
   // Use a transform to create a new stacking context.
-  transform: translate3D(0, 0, 0);
+  transform: translate3d(0, 0, 0);
 }
 
 /**
  * This mixin hides an element visually.
- * That means it's still accessible for screen-readers but not visible in view.
+ * That means it"s still accessible for screen-readers but not visible in view.
```
iveysaur: Missed a double quote 😄

robertmesserle: Ah, that was the one I mentioned - forgot to re-push after fixing it =x

hansl:
I restarted browserstack.

jelbourn:
LGTM

iveysaur:
Grid list scss file needs single quotes on the import statement.

hansl:
@robertmesserle ping.

PR #594, opened 2016-06-01
fix for checkbox opacity issues in safari

sendilkumarn:
#557 

 Variables - changed cubic-beizer for linear-out-slow-in based on [this](https://github.com/google/material-design-lite/blob/master/src/_variables.scss)

Calculation of opacity value in safari has some issues when rendering and it corrects itself upon moving focus out of window. i.e., on recalculating.

moved background and opacity from `.md-checkbox-background` and set them dynamically upon checkbox transition.

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -360,7 +360,7 @@ md-checkbox {
 
 .md-checkbox-unchecked {
   .md-checkbox-background {
-    background-color: none; 
+    background-color: transparent; 
   }
 }
 
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -241,7 +241,7 @@ md-checkbox {
   width: $md-checkbox-size;
 
   [dir="rtl"] & {
-    margin:{
+    margin: {
       left: $md-checkbox-item-spacing;
       right: auto;
     }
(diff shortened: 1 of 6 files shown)


src/core/style/_variables.scss
```diff
@@ -24,7 +24,7 @@ $md-toggle-size: 20px !default;
 // TODO(jelbourn): all of these need to be revisited
 
 // The default animation curves used by material design.
-$md-linear-out-slow-in-timing-function: cubic-bezier(0.0, 0.0, 0.2, 0.1) !default;
```
jelbourn: I'd prefer not to change this in this PR and instead do a review directly with the Material Design UX team.

src/components/checkbox/checkbox.scss
```diff
@@ -359,6 +357,13 @@ md-checkbox {
   }
 }
 
+
+.md-checkbox-unchecked {
+  .md-checkbox-background {
+    background-color: none;
```
jelbourn: I don't believe `none` is a valid value for `background-color`. Did you mean` transparent`?

src/core/style/_variables.scss
sendilkumarn: Can you guide on how?

src/components/checkbox/checkbox.scss
sendilkumarn: Yeah transperant.

src/core/style/_variables.scss
jelbourn: I mean to just revert this line here.

sendilkumarn: Okay then. Will change this and background thing

jelbourn:
LGTM

jelbourn:
Thanks!

Issue #586, opened 2016-05-31
radio: adding actionName attribute as label fallback (aria-label)


MartinMa:
#### Bug, feature request, or proposal:
feature request

#### What is the expected behavior?
If the radio button has an empty label, it should be possible (for accessibility reasons) to provide a value for the `aria-label` attribute.

#### What is the current behavior?
When the radio button is unlabeled, there is no way to set an aria-label on the input element as a fallback.

#### What are the steps to reproduce?
not applicable

#### What is the use-case or motivation for changing an existing behavior?
Think of a table with radio buttons in each row. The de facto labels are located in the `<th>`-part of the table.

This could be a solution:
```html
<md-radio-button name="group1" actionName="Option 1"></md-radio-button>
```

It results in:
```html
<md-radio-button _ngcontent-jha-3="" actionname="Option 1" name="group1" id="md-radio-0">
  <label class="md-radio-label" for="md-radio-0-input">
    <div class="md-radio-container">
      <div class="md-radio-outer-circle"></div>
      <div class="md-radio-inner-circle"></div>
      <div class="md-ink-ripple"></div>
    </div>
    <input class="md-radio-input" type="radio" id="md-radio-0-input" name="group1" aria-label="Option 1">
    <div class="md-radio-label-content">
    </div>
  </label>
</md-radio-button>
```

#### Which versions of Angular, Material, OS, browsers are affected?
Material 2.0.0-alpha.5

jelbourn:
We should just support setting `aria-label` directly on the `md-radio-button`

sendilkumarn:
😢 i finished it just now, without seeing this 👍

PR #596, opened 2016-06-01
feat(radio): support aria-label(ledby) on md-radio

MartinMa:
Adds feature #586 
**radio: adding actionName attribute as label fallback (aria-label)**

Added

- [x] tests
- [x] revised the documentation
- [x] and improved the examples/demo.

Please give me feedback, in case something needs to be changed.


src/components/radio/README.md
```diff
@@ -54,6 +54,7 @@ The `md-radio-group` component has no button initially selected.
 | `value` | `any` | The value of this radio button. |
 | `checked` | `boolean` | Whether the radio is checked. |
 | `disabled` | `boolean` | Whether the radio is disabled. |
+| `actionName` | `string` | Used to set the `aria-label` attribute of the underlying input element. |
```
DevVersion: This should be `aria-label`. We just want to forward the attribute. 

- As same as in the slide-toggle

src/components/radio/radio.html
```diff
@@ -13,6 +13,7 @@
           [checked]="checked"
           [disabled]="disabled"
           [name]="name"
+          [attr.aria-label]="actionName"
```
DevVersion: We should also support the `aria-labelled-by` attribute.

src/components/radio/radio.spec.ts
```diff
@@ -393,6 +401,20 @@ describe('MdRadio', () => {
...
+
+      fixture.detectChanges();
+      expect(nativeRadioInput.getAttribute('aria-label')).toBe('Banana');
+    });
+
+    it('should not add aria-label when actionName is not provided', () => {
+      let nativeRadioInput = <HTMLElement> fruitRadioNativeElements[1].querySelector('input');
+
+      fixture.detectChanges();
+      expect(nativeRadioInput.getAttribute('aria-label')).toBe(null);
```
DevVersion: I would also test the `aria-label` and `aria-labelled-by` attribute as a bound property, which means modifying the `@Input`'s input at runtime triggering a change detection cycle.

src/demo-app/radio/radio-demo.html
```diff
@@ -28,3 +28,30 @@
   </md-radio-group>
   <p>Your favorite season is: {{favoriteSeason}}</p>
 </section>
+<h1>ARIA Example</h1>
```
DevVersion: Not sure about adding an extra demo, just for the `aria-label` and `aria-labelled-by` attribute. It seems unnecessary to me.

src/components/radio/README.md
MartinMa: I think it is not a good idea to have a *pseudo* `aria-label` attribute on the MdRadio component and at the same time a real one on the underlying input element. It might confuse screen readers and its users. Do you know what I mean? That's why I chose a different name as seen on the [Angular 2 documentation for template syntax](https://angular.io/docs/ts/latest/guide/template-syntax.html) (see Property Binding / We must use attribute binding when there is no element property to bind).

But while taking a look a the current implementation for MdSlideToggle, I see that the property used there is named `ariaLabel`, which is different from `aria-label`. So this subtle difference is probably enough avoid potential issues with screen readers.

So if everyone agrees and doesn't see any more issues, I can change the property name (and add `aria-labelledby` as well).

src/components/radio/radio.html
MartinMa: Agreed.

src/components/radio/README.md
DevVersion: We also using the `aria-label` input in the checkbox as well. (See [here](https://github.com/angular/material2/blob/master/src/components/checkbox/checkbox.ts#L80)).

Screenreaders shouldn't be confused about that attribute, because the host element is normally not focusable.

Without any attribute binding, you're still using the dash cased attribute
```html
<md-radio-button aria-label="My Label Input">
```

But with attribute binding, you're right, it will use the camelCased attribute name.

jelbourn:
Agreed on all of @DevVersion's comments

src/components/radio/README.md
MartinMa: Thanks for the explanation. I will conform to the pattern used for the checkbox component.

MartinMa:
I updated the pull request.

- [x] Changed property to `ariaLabel` (corresponding attribute is called `aria-label`).
- [x] Added property `ariaLabelledby` (corresponding attribute is called `aria-labelledby`).
- [x] Updated tests and added new tests.
- [x] Updated README.
- [x] Removed demo/example code.

Let me know, if this is ok. I will squash all commits then, so it will be ready to be merged.

src/components/radio/radio.spec.ts
```diff
@@ -393,6 +409,40 @@ describe('MdRadio', () => {
       expect(weatherRadioInstances[1].checked).toBe(false);
       expect(weatherRadioInstances[2].checked).toBe(true);
     });
+
+    it('should add aria-label attribute to the underlying input element if defined', () => {
+      expect(fruitRadioNativeInputs[0].getAttribute('aria-label')).toBe('Banana');
+    });
+
+    it('should not add aria-label attribute if not defined', () => {
+      expect(fruitRadioNativeInputs[1].getAttribute('aria-label')).toBe(null);
```
jelbourn: I'd use `hasAttribute` here and in other tests where you're checking for the _presence_ of the attribute.

jelbourn:
Looks good aside from that one minor comment

MartinMa:
Thanks for the feedback. I updated the two tests to use `hasAttribute` and squashed all commits. :hammer:

jelbourn:
LGTM

jelbourn:
Thanks!

PR #608, opened 2016-06-01
fix: remove circular deps and add ci check

jelbourn:
Removes the three instances of circular deps and adds a CI check to prevent future regressions.


jelbourn:
R: @hansl @kara 
CC: @mprobst @vikerman

kara:
LGTM. We should eventually change our `system-config.ts` to have a main file configured at some point, so we don't have to do `@angular2-material/core/core`

PR #613, opened 2016-06-02
fix for checkbox click firing two times

sendilkumarn:
Whenever checkbox is clicked, click function is called twice. To prevent it. 

Guess it is because of bubbling in change event. This method will prevent it 
@jelbourn


sendilkumarn:
@jelbourn change retriggers click function.  instead we can use click straightaway. which works well

jelbourn:
Closed in favor of #672

PR #615, opened 2016-06-02
fix(overlay): add a z-index.

### LINKED ISSUE #584: md-overlay needs z-index.
**Bug**
Tested on latest/Chrome. 

The md-overlay-pane should have a z-index set to at least 1. 
Many of my other elements don't use z-index so I didn't notice until a label from a md-input was on-top of my overlay. So is the underline when the item was in focus. 

Adding: `z-index:1;` to the .md-overlay-pane fixes the issue. 

Images attached:
Before:
screen shot 2016-05-30 at 5 28 06 pm

After:
screen shot 2016-05-30 at 5 28 46 pm

jelbourn:
R: @kara 

Fixes #584

### DIFF
--- a/src/core/overlay/overlay.scss
+++ b/src/core/overlay/overlay.scss
@@ -19,4 +19,5 @@
   position: absolute;
   pointer-events: auto;
   box-sizing: border-box;
+  z-index: 1;
 }


hansl:
SHIP ITTTTTTTT!

PR #617, opened 2016-06-02
perf(progress-circle): clean up animation on destroy

crisbeto:
The indeterminate animation is based on an interval that keeps running, even after the element has been destroyed.
This change cleans up when the element is destroyed.

Also adds an extra null check for `performance.now`.

### DIFF
--- a/src/components/progress-circle/progress-circle.spec.ts
+++ b/src/components/progress-circle/progress-circle.spec.ts
@@ -74,6 +74,27 @@ describe('MdProgressCircular', () => {
         done();
       });
   });
+
+  it('should clean up the indeterminate animation when the element is destroyed',
+    (done: () => void) => {
+      let template = `<md-progress-circle
+                        mode="indeterminate"
+                        *ngIf="!isHidden"></md-progress-circle>`;
+
+      builder
+        .overrideTemplate(TestApp, template)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          fixture.detectChanges();
+          let progressElement = getChildDebugElement(fixture.debugElement, 'md-progress-circle');
+          expect(progressElement.componentInstance.interdeterminateInterval).toBeTruthy();
+
+          fixture.debugElement.componentInstance.isHidden = true;
+          fixture.detectChanges();
+          expect(progressElement.componentInstance.interdeterminateInterval).toBeFalsy();
+          done();
+        });
+    });
 });
 
 
--- a/src/components/progress-circle/progress-circle.ts
+++ b/src/components/progress-circle/progress-circle.ts
@@ -3,6 +3,7 @@ import {
   HostBinding,
   ChangeDetectorRef,
   ChangeDetectionStrategy,
+  OnDestroy,
   Input
 } from '@angular/core';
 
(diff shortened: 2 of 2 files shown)


jelbourn:
@crisbeto Can you add a unit test for this?

crisbeto:
Done @jelbourn, although some of the CI builds are failing because of unrelated issues. (missing `requestAnimationFrame`, Node errors)

src/components/progress-circle/progress-circle.spec.ts
```diff
@@ -74,6 +74,22 @@ describe('MdProgressCircular', () => {
...
   });
+
+  it('should clean up the indeterminate animation when the element is destroyed', (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, '<md-progress-circle mode="indeterminate" *ngIf="!isHidden"></md-progress-circle>')
+        .createAsync(TestApp)
+        .then((fixture) => {
+          fixture.detectChanges();
+          let progressElement = getChildDebugElement(fixture.debugElement, 'md-progress-circle');
+          expect(progressElement.componentInstance._interdeterminateInterval).toBeTruthy();
```
jelbourn: You can't access the private member here; The field either needs to be changed to `/** @internal */` or we need some other proxy for checking that the interval has stopped.

src/components/progress-circle/progress-circle.spec.ts
```diff
@@ -74,6 +74,22 @@ describe('MdProgressCircular', () => {
         done();
       });
   });
+
+  it('should clean up the indeterminate animation when the element is destroyed', (done: () => void) => {
```
jelbourn: Rather than using `done`, you can wrap the test function in `async()` and it will magically work. Check out the checkbox spec for an example.

jelbourn:
Only remaining CI errors are lint

src/components/progress-circle/progress-circle.spec.ts
crisbeto: Alright, I'll give it a try. I was following what the other tests were doing.

crisbeto:
It's updated to use a getter/setter on the interval and resetting it on the setter instead to avoid intervals stacking up. The CI is still throwing a Node ELIFECYCLE error, though.

jelbourn:
That node error message is just garbage that gets thrown when there's a tslint error. The actual problem is just above: 
https://travis-ci.org/angular/material2/jobs/134867948#L264

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,12 +42,19 @@ type EasingFn = (currentTime: number, startValue: number,
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class MdProgressCircle {
+export class MdProgressCircle implements OnDestroy {
   /** The id of the last requested animation. */
   private _lastAnimationId: number = 0;
 
   /** The id of the indeterminate interval. */
   private _interdeterminateInterval: number;
+  get interdeterminateInterval() {
```
jelbourn: This would be marked as `/** @internal */`

crisbeto:
There we go.

src/components/progress-circle/progress-circle.ts
```diff
@@ -41,12 +42,22 @@ type EasingFn = (currentTime: number, startValue: number,
...
 })
-export class MdProgressCircle {
+export class MdProgressCircle implements OnDestroy {
   /** The id of the last requested animation. */
   private _lastAnimationId: number = 0;
 
-  /** The id of the indeterminate interval. */
+  /**
+   * The id of the indeterminate interval.
+   * @internal
```
jelbourn: The `@internal` actually goes on the public member rather than the private one; this tells TypeScript to omit the property when outputting a `.d.ts` file so that downstream users can't depend on it.

crisbeto: Sorry, I'm still not too familiar with all the ng2 concepts. It's updated now.

jelbourn:
LGTM

PR #625, opened 2016-06-03
chore(ci): block import of entire rxjs library

jelbourn:
R: @hansl @kara  

Thanks to @DevVersion for making me realize we need a check for this.

### DIFF
--- a/scripts/ci/forbidden-identifiers.js
+++ b/scripts/ci/forbidden-identifiers.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 'use strict';
 /*
  * This script analyzes the current commits of the CI.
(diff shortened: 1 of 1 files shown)


hansl:
Can you elaborate? Is it because you want tree shaking?

jelbourn:
@hansl Effectively; all of the imports for rxjs should be for the specific parts you want. If you do `rxjs/Rx`, it brings in the entire library. See `icon.ts`, for example, which does:
```ts
import 'rxjs/add/observable/forkJoin';
import 'rxjs/add/observable/of';
import 'rxjs/add/operator/filter';
import 'rxjs/add/operator/do';
import 'rxjs/add/operator/share';
import 'rxjs/add/operator/finally';
import 'rxjs/add/operator/catch';
```

See #431

hansl:
LGTM

hansl:
Thank you for context.

PR #631, opened 2016-06-04
fix(e2e): remove static typings and use typings manager

### LINKED ISSUE #630: e2e: typings are invalid for e2e tsconfig
It seems like the e2e are always failing locally and on the CI, because of some incorrect typings.

image

See for example, CI job: https://travis-ci.org/angular/material2/jobs/135156835

Those typings are quite outdated and hard-copied into the repository. We should better install the typings from the `typings manager`

- See https://github.com/Microsoft/TypeScript/issues/8737

As a temporary fix, we can remove the node typings from the e2e typings. (those aren't used yet)
Once we upgrade typescript we can re-introduce the node typings again.

DevVersion:
* Removes static typings and uses the typings manager
* Fixes duplicate identifier issue of Typescript temporary

 - See Microsoft/TypeScript#8737

This makes the E2E tests working on the CI, so the tests will probably fail, but the test fixes should be definitely addressed in another PR.

Darn it :smile: those much file deletions look weird ;P

Fixes #630.


DevVersion:
FYI: We'll probably go with #637

PR #636, opened 2016-06-05
build: add root to the sourcemap so it won't use absolute pathing

### LINKED ISSUE #541: release: incorrect sourcemaps
In previous releases like `alpha.4`, the sourcemap paths were correctly resolved in relative to the transpiled file.

This has changed in `alpha.5`. 
```js
//# sourceMappingURL=/usr/local/google/home/jelbourn/material2/tmp/broccoli_type_script_compiler-input_base_path-IydvmmBU.tmp/0/components/slide-toggle/slide-toggle.js.map
```

Since, the sourcemap paths are not valid anymore, SystemJS is not able to create a bundle of Angular Material 2 components. (applies also to the CLI - uses SystemJS builder as well).

I temporary fixed this, by stripping the souremap paths out of the components.

hansl:
Fixes #541, I think.

Need to chat to @jelbourn to try to see if that fixes the release itself, not jsut local.

### DIFF
--- a/angular-cli-build.js
+++ b/angular-cli-build.js
@@ -101,10 +101,10 @@ function _buildAppTree(defaults) {
     vendorNpmFiles: [
       'systemjs/dist/system-polyfills.js',
       'systemjs/dist/system.src.js',
-      'zone.js/dist/*.js',
+      'zone.js/dist/*.+(js|js.map)',
       'core-js/client/core.js',
-      'rxjs/**/*.js',
-      '@angular/**/*.js',
+      'rxjs/**/*.+(js|js.map)',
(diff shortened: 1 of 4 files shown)


hansl:
This requires the latest release of the CLI to get https://github.com/angular/angular-cli/pull/1028.

hansl:
Also fix prod builds, which actually don't make sense for material, but :shrug:

jelbourn:
I'm going to merge this since it's _better_, but I think the sourceMaps still aren't exactly right.

danielrasmuson:
Thanks guys. We hit this yesterday.

PR #637, opened 2016-06-05
fix: remove duplicated typing. TS stdlib provides enough for e2e tests

hansl:


DevVersion:
@hansl As seen in #631, can we not use the typings manager here and remove the static typings?

hansl:
We'll just skip typings and go directly to `@types` when TS 2.0 is released.

hansl:
R: @jelbourn

jelbourn:
LGTM

Issue #416, opened 2016-05-10
VoiceOver does not recognize indeterminate progress status


jelbourn:
When using VoiceOver on both progress-bar and process-circle, the screen-reader does not pick up the "indeterminate" state when mode is set to indeterminate. Instead, it simply reads "0 percent, progress indicator". 

It *does* work for `md-spinner`, which sets the mode in the constructor, where it reads "indeterminate, progress indicator".

jelbourn:
This should have been fixed by #640

PR #640, opened 2016-06-06
fix(mdProgressCircle) Fix aria labels based on mode.

josephperrott:
#416

### DIFF
--- a/src/components/progress-circle/progress-circle.spec.ts
+++ b/src/components/progress-circle/progress-circle.spec.ts
@@ -42,16 +42,34 @@ describe('MdProgressCircular', () => {
       });
   });
 
-  it('should define a default value for the value attribute', (done: () => void) => {
+  it('should define a default value of undefined for the value attribute', (done: () => void) => {
     builder
       .overrideTemplate(TestApp, '<md-progress-circle></md-progress-circle>')
       .createAsync(TestApp)
       .then((fixture) => {
         fixture.detectChanges();
         let progressElement = getChildDebugElement(fixture.debugElement, 'md-progress-circle');
-        expect(progressElement.componentInstance.value).toBe(0);
+        expect(progressElement.componentInstance.value).toBeUndefined();
         done();
       });
+  });
+
+  it('should set the value to undefined when the mode is set to indeterminate',
+    (done: () => void) => {
+      builder
+        .overrideTemplate(TestApp, `<md-progress-circle value="50"
+                                                        [mode]="mode"></md-progress-circle>`)
+        .createAsync(TestApp)
+        .then((fixture) => {
+          let progressElement = getChildDebugElement(fixture.debugElement, 'md-progress-circle');
+          fixture.debugElement.componentInstance.mode = 'determinate';
+          fixture.detectChanges();
+          expect(progressElement.componentInstance.value).toBe(50);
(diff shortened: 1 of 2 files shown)


src/components/progress-circle/progress-circle.spec.ts
```diff
@@ -42,18 +42,36 @@ describe('MdProgressCircular', () => {
...
-  it('should define a default value for the value attribute', (done: () => void) => {
+  it('should define a default value of undefined for the value attribute', (done: () => void) => {
     builder
       .overrideTemplate(TestApp, '<md-progress-circle></md-progress-circle>')
       .createAsync(TestApp)
       .then((fixture) => {
         fixture.detectChanges();
         let progressElement = getChildDebugElement(fixture.debugElement, 'md-progress-circle');
-        expect(progressElement.componentInstance.value).toBe(0);
+        expect(progressElement.componentInstance.value).toBe(undefined);
```
jelbourn: jasmine actually has a function `toBeUndefined`

src/components/progress-circle/progress-circle.spec.ts
```diff
@@ -42,18 +42,36 @@ describe('MdProgressCircular', () => {
...
         done();
       });
   });
 
+  it('should set the value to undefined when the mode is set to indeterminate',
```
jelbourn: Rather than using `done`, you can do the test like
```ts
it('should do a thing', async(() => {
  // do a bunch of stuff, some of which is async
}));
```

And it will magically know when its done.

src/components/progress-circle/progress-circle.ts
```diff
@@ -50,6 +50,20 @@ export class MdProgressCircle implements OnDestroy {
   private _interdeterminateInterval: number;
 
   /** @internal */
+  get ariaValueMin() {
+    if (this.mode != 'indeterminate') {
+      return 0;
+    }
```
jelbourn: How about
```
return this.mode == 'determinate' ? 0 : null
```
?

src/components/progress-circle/progress-circle.ts
```diff
@@ -50,6 +50,20 @@ export class MdProgressCircle implements OnDestroy {
   private _interdeterminateInterval: number;
 
   /** @internal */
+  get ariaValueMin() {
+    if (this.mode != 'indeterminate') {
```
jelbourn: Add comment explaining why we want `valuemin` and `valueman` to be omited for indeterminate.

src/components/progress-circle/progress-circle.spec.ts
josephperrott: Done.

src/components/progress-circle/progress-circle.spec.ts
josephperrott: Done, unchanged as we cannot use async while setInterval is used.

src/components/progress-circle/progress-circle.ts
josephperrott: Done.

src/components/progress-circle/progress-circle.ts
josephperrott: Done.

jelbourn:
LGTM

PR #641, opened 2016-06-06
fix(tabs): removes minimum height

### LINKED ISSUE #570: md-tab-body height doesn't collapse height properly
Consider the following template:
```
<md-tab-group>
  <md-tab>
    <template md-tab-label>Label</template>
    <template md-tab-content>
      Any object shorter than 199px
    </template>
  </md-tab>
</md-tab-group>
<div>footer text</div>
```
The generated html puts the tab contents inside a `<div class="md-tab-body">`.  Problem is that this div is always at least 199px in height.  This poses problems when my content is meant to take very little height in the browser positioning logic.  In the example above, notice how much blank space there is between the content and the footer

My specific user case is that within the tab pane I have a content-div whose aspect ratio needs to remain fixed. I've been using  [the solution from this post](http://stackoverflow.com/questions/1495407/maintain-the-aspect-ratio-of-a-div-with-css). It has been working well until I attempted to put the content in the new `md-tab` directive. Now, the bottom padding of the container-div no longer respects the limit set in CSS `padding-bottom` rule and I can't figure out why.

robertmesserle:
closes #570 

R: @jelbourn

### DIFF
--- a/src/components/tabs/tab-group.scss
+++ b/src/components/tabs/tab-group.scss
@@ -2,13 +2,11 @@
 @import 'default-theme';
 
 $md-tab-bar-height: 48px !default;
-$md-tab-content-min-height: 200px !default;
 
 :host {
   display: flex;
   flex-direction: column;
   font-family: $md-font-family;
-  min-height: $md-tab-bar-height + $md-tab-content-min-height;
 }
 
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #650, opened 2016-06-08
feat(tabs): adds e2e tests for tabs

### LINKED ISSUE #549: Add e2e tests for tabs

robertmesserle:
closes #549 

R: @jelbourn

### DIFF
--- a/e2e/components/tabs/tabs.e2e.ts
+++ b/e2e/components/tabs/tabs.e2e.ts
@@ -1,0 +1,97 @@
+import ElementArrayFinder = protractor.ElementArrayFinder;
+import ElementFinder = protractor.ElementFinder;
+
+describe('tabs', () => {
+  describe('basic behavior', () => {
+    let tabGroup: ElementFinder;
+    let tabLabels: ElementArrayFinder;
+    let tabBodies: ElementArrayFinder;
+
+    beforeEach(() => {
+      browser.get('/tabs');
+      tabGroup = element(by.css('md-tab-group'));
+      tabLabels = element.all(by.css('.md-tab-label'));
+      tabBodies = element.all(by.css('.md-tab-body'));
+    });
+
+    it('should change tabs when the label is clicked', () => {
+      tabLabels.get(1).click();
+      expect(getActiveStates(tabLabels)).toEqual([false, true, false]);
+      expect(getActiveStates(tabBodies)).toEqual([false, true, false]);
+
+      tabLabels.get(0).click();
+      expect(getActiveStates(tabLabels)).toEqual([true, false, false]);
+      expect(getActiveStates(tabBodies)).toEqual([true, false, false]);
+    });
+
+    it('should change focus with keyboard interaction', () => {
+      tabLabels.get(0).click();
+      expect(getFocusStates(tabLabels)).toEqual([true, false, false]);
+
+      pressKey(protractor.Key.RIGHT);
+      expect(getFocusStates(tabLabels)).toEqual([false, true, false]);
+
+      pressKey(protractor.Key.RIGHT);
+      expect(getFocusStates(tabLabels)).toEqual([false, false, true]);
+
+      pressKey(protractor.Key.RIGHT);
+      expect(getFocusStates(tabLabels)).toEqual([false, false, true]);
+
+      pressKey(protractor.Key.LEFT);
+      expect(getFocusStates(tabLabels)).toEqual([false, true, false]);
+
+      pressKey(protractor.Key.LEFT);
+      expect(getFocusStates(tabLabels)).toEqual([true, false, false]);
+
+      pressKey(protractor.Key.LEFT);
+      expect(getFocusStates(tabLabels)).toEqual([true, false, false]);
+    });
+  });
+});
+
+/**
+ * A helper function to perform the sendKey action
+ * @param key
+ */
+function pressKey(key: string) {
+  browser.actions().sendKeys(key).perform();
+}
+
+/**
+ * Returns an array of true/false that represents the focus states of the provided elements
+ * @param elements
+ * @returns {webdriver.promise.Promise<Promise<boolean>[]>|webdriver.promise.Promise<T[]>}
+ */
+function getFocusStates(elements: ElementArrayFinder) {
+  return elements.map(element => {
+    return element.getText().then(elementText => {
+      return browser.driver.switchTo().activeElement().getText().then(activeText => {
+        return activeText === elementText;
+      });
+    });
+  });
+}
+
+/**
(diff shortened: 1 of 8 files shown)


src/e2e-app/index.html
```diff
@@ -1,6 +1,6 @@
 <!doctype html>
 {{!-- This file is an handlebar file that gets filled at build time. --}}
-<html>
+<html lang="en">
```
jelbourn: Update this for demmo-app as well?

src/e2e-app/tabs/tabs-e2e.ts
```diff
@@ -0,0 +1,10 @@
+import {Component} from '@angular/core';
+import {MD_TABS_DIRECTIVES} from '@angular2-material/tabs/tabs';
+
+@Component({
+  moduleId: module.id,
+  selector: 'tabs-e2e',
+  templateUrl: 'tabs-e2e.html',
+  directives: [MD_TABS_DIRECTIVES]
+})
+export class TabsE2E {}
```
jelbourn: Rather than calling this `TabsE2E`, I'd prefer to call it something like `BasicTabs`, since we can eventually have multiple scenarios for each component and also want to reuse these things for examples.

e2e/components/tabs/tabs.e2e.ts
```diff
@@ -0,0 +1,24 @@
+describe('tabs', function () {
```
jelbourn: Change `function() {` to `() => {` throughout this file.

src/e2e-app/index.html
DevVersion: You could probably use this PR for the a11y test fixes instead of #654 (because `lang="en"` on the demo-app will fix one)

e2e/components/tabs/tabs.e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+  describe('basic behavior', function () {
+    beforeEach(function() {
+      browser.get('/tabs');
+    });
+    it('should change tabs when the label is clicked', function () {
+      element(by.css('.md-tab-label:nth-of-type(2)')).click();
```
jelbourn: Rather than using `nth-of-type`, you can do `element.all().get(index)`, e.g.
```ts
let tabLabels = element.all(by.css('.md-tab-label'));
tabLabels.get(2).click();
...
```

jelbourn: Actually, I'll just CC @andresdominguez for input on the test

e2e/components/tabs/tabs.e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+      expect(element(by.css('.md-tab-label:nth-of-type(2)')).getAttribute('class')).not
+          .toMatch('md-active');
+      expect(element(by.css('.md-tab-label:nth-of-type(1)')).getAttribute('class'))
+          .toMatch('md-active');
+      expect(element(by.css('.md-tab-body:nth-of-type(2)')).getAttribute('class')).not
+          .toMatch('md-active');
+      expect(element(by.css('.md-tab-body:nth-of-type(1)')).getAttribute('class'))
+          .toMatch('md-active');
+    });
+  });
```
jelbourn: Testing for focus behavior? Keyboard interaction?

e2e/components/tabs/tabs.e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+    });
+    it('should change tabs when the label is clicked', function () {
+      element(by.css('.md-tab-label:nth-of-type(2)')).click();
+      expect(element(by.css('.md-tab-label:nth-of-type(2)')).getAttribute('class'))
+          .toMatch('md-active');
+      expect(element(by.css('.md-tab-body:nth-of-type(2)')).getAttribute('class'))
+          .toMatch('md-active');
+
+      element(by.css('.md-tab-label:nth-of-type(1)')).click();
+      expect(element(by.css('.md-tab-label:nth-of-type(2)')).getAttribute('class')).not
```
andresdominguez: To assert that the tabs contain a class you can do something like this:

```js
// Note: $$('.abc') === element.all(by.css('.abc'))
let tabsActive = $$('.md-tab-label').map((item, index) => {
  // Note: I am creating an array that contains {index, isActive} where isActive means that
  // it has the active class
  return {
    index: index,
    isActive: item.getAttribute('class').then(
        classAttr => !!classAttr.match('dropdown'))
  };
});

expect(tabsActive).toEqual([
  {index: 0, isActive: false}, 
  {index: 1, isActive: true},
  {index: 2, isActive: false}
]);
```

Or you can just do:

```js
let tabsActive = $$('.md-tab-label')
    .map(item => item.getAttribute('class')
        .then(classAttr => !!classAttr.match('dropdown')));

expect(tabsActive).toEqual([false, true, false]);
```

robertmesserle:
@jelbourn This should be okay to merge now.

e2e/components/tabs/tabs.e2e.ts
```diff
@@ -0,0 +1,76 @@
+import ElementArrayFinder = protractor.ElementArrayFinder;
+import ElementFinder = protractor.ElementFinder;
```
jelbourn: `import {ElementArrayFinder} from ...` ? etc.

robertmesserle: Apparently, this is how WebStorm fixed my code.

e2e/components/tabs/tabs.e2e.ts
```diff
@@ -0,0 +1,76 @@
...
+      expect(getFocusStates(tabLabels)).toEqual([true, false, false]);
+    });
+  });
+});
+
+function pressKey(key: string) {
+  browser.actions().sendKeys(key).perform();
+}
+
+function getFocusStates(elements: ElementArrayFinder) {
```
jelbourn: Add descriptions explaining what these functions are returning?

src/components/tabs/tabs.ts
```diff
@@ -18,6 +18,14 @@ import {MdInkBar} from './ink-bar';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/operator/map';
 
+// Due to a bug in the ChromeDriver, Angular 2 keyboard events do not work when using dot notation
+// such as `(keydown.rightArrow)` in E2E tests.  To get around this, we are temporarily using a
```
jelbourn: It's not that the events don't work, it's that `sendKeys` won't trigger them.

jelbourn:
@robertmesserle needs a rebase on top of the new router. LGTM aside from that (and one minor nit). 

Feel free to merge directly after rebase.

Issue #205, opened 2016-03-18
Typography


jelbourn:
https://www.google.com/design/spec/style/typography.html

Combination of documentation and CSS.

sendilkumarn:
can we add it per component or use typography.scss ?

sendilkumarn:
if this is fine I will update the remaining components

jelbourn:
To clarify: this issue is about making typography styles available for users to have in their own applications. I'm going to mark it as blocked since this will be influenced by how we approach theming as well.

michaelwarren1106:
what's the implementation plan for typography? Are you guys planning on making this be a web component, or basically a stylesheet that will need to be imported/required into app components individually?

I've been debating the "proper" approach to typography and how best to componentize re-usable styles that generally are "global" in nature but dont really need a web component to be executed.

If you could outline some thoughts on your approach and the pros/cons thereof, that would be awesome!

jelbourn:
Current thinking is to have an `@angular2-material/typography` package that includes a css file that you can drop into a `<link>` element in your index.html (or concat into other css files with your own build). We haven't finalized on this approach yet, though.

michaelwarren1106:
with shadomDOM coming, and native in angular2, will that approach work? I've done a little bit of testing and it seems that this approach wont work since styles in the parent page wont bleed into a shadowDOM component...

kylecordes:
I stumbled across this today while looking for an answer to the following: For applications starting to use Angular Material 2 today, prior to the typography implementation, what is best interim short-term solution for styling application content that is not a angular-material Component? For example, is it best to grab one of the other Material implementations, and use it alongside, for styling other content? Any advice for which one fits in with the least trouble?

krigton:
Does it makes sense using typography from material 1 (no webcomponents usage yet), untill this at least started?

krigton:
Will typography support theming? Like using different font for md-display and body copy.

thanks

michaeljota:
@krigton it's blocked cause of theming, so I would say it will.

shlomiassaf:
We've started using a webpack loader to override styles, since CSS comes bundled in we can't actually change much, only colors.

We also had hard time changing the font.

This is the loader [angular2-restyle-loader](https://github.com/shlomiassaf/angular2-restyle-loader)
It's fresh out of the oven.

We use it to re-compile SCSS so we can set the variables.
It can also do a lot more with higher maintenance risk.

For example, you can restyle the slide toggle  
[...]

PR #656, opened 2016-06-08
typography - added 

sendilkumarn:
#205 

components addressed 

- [x]  Cards
- [x] Toggle Button
- [x]  Tabs


jelbourn:
@sendilkumarn what's meant in #205 is creating a typography module that people can consume for their own applications. I genuinely appreciate your enthusiasm, but you probably want to wait for a bit of discussion before diving into a PR to make sure we're all on the same page.

PR #664, opened 2016-06-09
fix(checkbox): only fires one click event per click

### LINKED ISSUE #651: MdCheckbox fires a click event twice
I noticed that the (click) event is being called twice for MdCheckbox.

robertmesserle:
closes #651 
R: @jelbourn


jelbourn:
@robertmesserle Seems to fail on IE11, Edge, and iOS Safari.

Can you add an explanation for what's going on here both in the PR and in comments in the code?

jelbourn:
There's an other PR #613 for this; does changing the event to just `click` solve this problem?

robertmesserle:
Hm, it might - I was mostly re-adding what I had removed from checkboxes
during the refactor, but if that works, that's even better!

robertmesserle:
@jelbourn We should use #613 for this

Issue #626, opened 2016-06-03
Refactor radio and button-toggle to decouple from parent component


jelbourn:
Current the radio button (and the button-toggle, which does the same thing) depends on its parent group and sets values to it directly. This could potentially be made to be less coupled if the parent group listened to `change` and `valueChange` events on the radio.

Issue #609, opened 2016-06-01
bug(radio): unchecking a button does not update parent value


robertmesserle:
When unchecking a radio button programmatically, the parent group is never notified.  The parent group will need to know in order to update `selected` and `value`.  It will also need to emit its own `change` event.

jelbourn:
@tinayuangao we need to check whether this is still a bug; it may have been fixed since this was filed.

PR #669, opened 2016-06-11
refactor(Radio): Decouple the radio from the radio-group component.

josephperrott:
For #626, #609


jelbourn:
Going to close this since it would be rebase hell now. We will revisit this someday!

PR #672, opened 2016-06-12
fix(slide-toggle): visual hidden input should not bubble click

### LINKED ISSUE #671: (click) event generates twice on md-slide-toggle and md-chekbox
Bug

(click) to generate a single event

it generates double event

on the md-checkbox and md-slide-toggle create a (click) event binding and you will see it runs the function twice.
Plunker template: http://plnkr.co/edit/FxHKvcNJe2Zq5Mq5Imni?p=info

This issue might exist on other components but I haven't tested them.

DevVersion:
* Currently the (click) event gets called twice.
  This is caused by the bubbling of the (click) event on the input.

Actually the second click will be called by the label element.
The browsers default behavior is, to trigger a click on the input, once the label got clicked.

This is in my opinion the correct approach, because we don't want to depend on the `click` event, instead of the `change` event.

- https://github.com/angular/angular/blob/master/modules/%40angular/common/src/forms-deprecated/directives/checkbox_value_accessor.ts

Fixes #671.

### DIFF
--- a/src/components/checkbox/checkbox.ts
+++ b/src/components/checkbox/checkbox.ts
@@ -278,6 +278,18 @@ export class MdCheckbox implements AfterContentInit, ControlValueAccessor {
     }
   }
 
+  /** @internal */
+  onInputClick(event: Event) {
+    // We have to stop propagation for click events on the visual hidden input element.
+    // By default, when a user clicks on a label element, a generated click event will be
+    // dispatched on the associated input element. Since we are using a label element as our
+    // root container, the click event on the `checkbox` will be executed twice.
+    // The real click event will bubble up, and the generated click event also tries to bubble up.
+    // This will lead to multiple click events.
+    // Preventing bubbling for the second event will solve that issue.
+    event.stopPropagation();
+  }
+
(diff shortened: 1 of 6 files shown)


DevVersion:
@jelbourn Updated the PR (added a comment and moved the event into the component source)

src/components/checkbox/checkbox.ts
```diff
@@ -278,6 +278,18 @@ export class MdCheckbox implements AfterContentInit, ControlValueAccessor {
...
 
+  /** @internal */
+  onInputClick(event: Event) {
+    // We have to stop propagation for click events on the visual hidden input element.
+    // By default, when a user clicks on a label element, a generated click event will be
+    // dispatched on the associated input element. Since we are using a label element as our
+    // root container, the click event on the `checkbox` will be executed twice.
+    // The real click event will bubble up, and the generated click event also tries to bubble up.
+    // This will lead to multiple click events.
+    // Preventing bubbling for the second event will solve that issue.
```
jelbourn: Fantastic comment

jelbourn:
LGTM

PR #680, opened 2016-06-13
fix(input): placeholder text is hidden by parent visibility

### LINKED ISSUE #670: Input placeholder is not hiding if parent's visibility:hidden;
BUG

Children content should be invisible, when parent's visibility:none

.md-input-placeholder.md-empty class overrides parents visibility.

set visibility: hidden to parent div, containing input with placeholder.

angular rc1, material2 alpha.5

Can submit fix PR, If BUG is accepted

robertmesserle:
Previously, adding `visibility: hidden` to a parent element would not hide the placeholder text for `md-input`.

Closes #670 

R: @hansl @jelbourn

### DIFF
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -31,7 +31,7 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
  * Applies a floating placeholder above the input itself.
  */
 @mixin md-input-placeholder-floating {
-  visibility: visible;
+  display: block;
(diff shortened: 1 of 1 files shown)


robertmesserle:
@hansl Doesn't look like there were any transitions used on `visibility: hidden`, seems identical with `display: none`

hansl:
If you don't see any problems, then LGTM :)

PR #689, opened 2016-06-15
fix(input): Fix floating label width

### LINKED ISSUE #688: Input floating label should expand when entering floating state
Bug

Floating label resize the width to use all available space

currently it scale but not resize the width, this way the text is truncated inappropriately 

Current behavior:

out0

Expected behavior:

out

jean-merelis:
Fix floating label to resize the width to use all available space.

Fixes #688

### DIFF
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -34,7 +34,7 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
   visibility: visible;
   padding-bottom: 5px;
   transform: translateY(-100%) scale(0.75);
-  transition: width 0.1s ease-in-out;
+  transition: width 150ms ease-in-out;
   width: 133%; // 100/75 of the original space now available
 	
   .md-placeholder-required {
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -2,7 +2,6 @@
 @import 'mixins';
 @import 'variables';
 
-
 // Placeholder colors. Required is used for the `*` star shown in the placeholder.
 $md-input-placeholder-color: md-color($md-foreground, hint-text);
 $md-input-floating-placeholder-color: md-color($md-primary);
@@ -33,9 +32,9 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
 @mixin md-input-placeholder-floating {
   visibility: visible;
   padding-bottom: 5px;
-  transform: translateY(-100%) scale(0.75);
-  width: 133%; // 100/75 of the original space now available
-	
+  transform: translateY(-100%) scale($md-input-floating-placeholder-scale-factor);
+  width: calc(100% / #{$md-input-floating-placeholder-scale-factor});
+
   .md-placeholder-required {
     color: $md-input-required-placeholder-color;
   }
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -34,7 +34,6 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
   visibility: visible;
   padding-bottom: 5px;
   transform: translateY(-100%) scale(0.75);
-  transition: width 150ms ease-in-out;
   width: 133%; // 100/75 of the original space now available
 	
   .md-placeholder-required {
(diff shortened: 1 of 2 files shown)


crisbeto:
You should keep in mind that this will cause an overflow on IE and Edge. They don't consider the scale when determining whether to overflow the parent.

src/components/input/input.scss
```diff
@@ -34,7 +34,8 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
   visibility: visible;
   padding-bottom: 5px;
   transform: translateY(-100%) scale(0.75);
-
+  width: 133%; // 100/75 of the original space now available
```
hansl: Instead of magic number, please use a scale constant and make the width `100% / placeholderScaleFactor`.

Also please transition the width as well.

jean-merelis: It is not a magic number. It is just 100% / placeholderScaleFactor. 100/75=1.33

src/components/input/input.scss
```diff
@@ -34,7 +34,9 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
   visibility: visible;
   padding-bottom: 5px;
   transform: translateY(-100%) scale(0.75);
-
+  transition: width 150ms ease-in-out;
```
hansl: So this will overwrite or be overwritten by the transition already on the `.md-input-placeholder`. The transition should be added to line 125 AND use material transition functions.

hansl:
@crisbeto: I think in those case we'll see the same behaviour as before this change, which is fine by me (for now).

crisbeto:
We had a similar issue on Material 1, even though the label is scaled down to 75%, IE/Edge still consider it as if it was 100%. Then when you set it to 133% it overflows. See https://github.com/angular/material/pull/8116

src/components/input/input.scss
hansl: What I'm saying is that this is a good time to move the `0.75` out into a variable and use that variable in the width calculation. This will clean up stuff and make it clear what the intent is. The comment `100 / 75` doesn't tell us anything about where `75` came from. Nor what is the purpose of 75.

Also, if we change the `0.75` scale, we can't possibly know what else to change.

jelbourn:
@hansl are you able to see if this is an issue in IE / Edge?

src/core/style/_variables.scss
```diff
@@ -20,6 +20,9 @@ $md-toggle-padding: 8px !default;
 // Width and height of input toggles
 $md-toggle-size: 20px !default;
 
+
+$md-input-floating-placeholder-scale-factor: 0.75 !default;
```
hansl: Variables that are specific to a single component should live in that component's SCSS.

Just move this to the file where it's used.

src/components/input/input.scss
```diff
@@ -33,7 +32,8 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
 @mixin md-input-placeholder-floating {
   visibility: visible;
   padding-bottom: 5px;
-  transform: translateY(-100%) scale(0.75);
+  transform: translateY(-100%) scale($md-input-floating-placeholder-scale-factor);
+  width: calc(100% / #{$md-input-floating-placeholder-scale-factor});
```
hansl: No need for `calc()` AFAIK.

hansl:
Re IE/Edge I haven't had a chance to get my Windows Machine up last week. Will check it out tomorrow on a local VM.

hansl:
I'm installing a VM to test this on IE 11. Will come back when done.

hansl:
This seems to be working fine on IE11. Syncing up with Robert for more info.

Proof:

hansl:
@crisbeto: ^ seems to work correctly. If you have something to add before we merge this in :)

Please note our DOM is largely different than Material 1, and much simpler/less reliant on hacks due to Shadow DOM and compartmentalization.

hansl:
As soon as we clarify the situation with IE11/Edge I'll merge this in. :)

crisbeto:
I'll try cloning it to check as well.

crisbeto:
Alright, it works as it is now, but it starts overflowing if the any of the parents get `overflow: auto`. I'm not if it's an issue in this case, but here's what it looks like:

a

hansl:
Interesting. By putting `overflow: hidden` on the md-input itself we mitigate setting it to anything above, which would be okay I guess. The worst case would be if a customer sets `overflow: auto` to the `md-input` itself, which we can say "don't do that" since there's no real reason to do it. And it would be fine if he sets it on the element above (and beyond). @crisbeto: What do you think?

@jean-merelis: Can you add `overflow: hidden` to the `:host`?

crisbeto:
Yeah I'd say that setting a default overflow would be fine, since it's up to the end user to determine whether to override it in that case.

hansl:
Thanks Kristiyan. Sorry if this was such a mess, but better be right the first time :)

After discussion with @robertmesserle seems like this is the right solution. I'll merge it as soon as Travis finishes.

PR #690, opened 2016-06-15
feat(input): Add ripple effect for input underline when it receives the focus

jean-merelis:
Add ripple effect for input underline when it receives  the focus

### DIFF
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -178,13 +178,17 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
       z-index: 1;
       background-color: $md-input-underline-focused-color;
       top: -1px;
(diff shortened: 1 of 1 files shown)


DevVersion:
I know some Material Design Reference Implementations are supporting that functionality (like Polymer), but the specs aren't mentioning anything about that.

See for example this [video](https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0B7WCemMG6e0VcHFla0dwOE5pVVE/forms-supportive-content-v3_xhdpi.webm) from the [Material Design Specifications](https://material.google.com/components/text-fields.html#text-fields-search-filter)

jean-merelis:
You are right.

PR #694, opened 2016-06-15
Add ngModel support to button toggles

iveysaur:
R: @jelbourn @kara

### DIFF
--- a/src/components/button-toggle/button-toggle.spec.ts
+++ b/src/components/button-toggle/button-toggle.spec.ts
@@ -8,6 +8,7 @@ import {
   fakeAsync,
   tick,
 } from '@angular/core/testing';
+import {NgControl} from '@angular/common';
 import {TestComponentBuilder, ComponentFixture} from '@angular/compiler/testing';
 import {Component, DebugElement, provide} from '@angular/core';
 import {By} from '@angular/platform-browser';
@@ -15,7 +16,8 @@ import {
   MD_BUTTON_TOGGLE_DIRECTIVES,
   MdButtonToggleGroup,
   MdButtonToggle,
-  MdButtonToggleGroupMultiple
+  MdButtonToggleGroupMultiple,
+  MdButtonToggleChange,
 } from './button-toggle';
 import {
   MdUniqueSelectionDispatcher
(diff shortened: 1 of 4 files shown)


src/components/button-toggle/button-toggle.spec.ts
```diff
@@ -8,14 +8,16 @@ import {
   fakeAsync,
   tick,
 } from '@angular/core/testing';
+import {NgControl} from '@angular/common';
```
kara: Can you add a TODO to update this to @angular/forms when we have rc.2?

src/components/button-toggle/button-toggle.ts
```diff
@@ -1,26 +1,42 @@
...
 } from '@angular2-material/core/coordination/unique-selection-dispatcher';
 import {BooleanFieldValue} from '@angular2-material/core/annotations/field-value';
 
 export type ToggleType = 'checkbox' | 'radio';
 
 
+
+/**
+ * Provider Expression that allows md-button-toggle-group to register as a ControlValueAccessor.
+ * This allows it to support [(ngModel)] and ngControl.
```
kara: Can you remove reference to ngControl? It doesn't exist in the new forms, so the comment will be stale.

src/components/button-toggle/button-toggle.ts
```diff
@@ -1,26 +1,42 @@
...
+ */
+const MD_BUTTON_TOGGLE_GROUP_CONTROL_VALUE_ACCESSOR = new Provider(
```
kara: Nitpick: any chance you could make this a bit shorter?  Maybe lose the CONTROL?  MD_BUTTON_TOGGLE_GROUP_VALUE_ACCESSOR

src/components/button-toggle/button-toggle.ts
```diff
@@ -108,6 +131,17 @@ export class MdButtonToggleGroup {
...
   }
 
+  /**
+   * Mark this group as being "touched" (for ngModel).
+   * Meant to be called by the contained button toggles upon their blur.
+   * @internal
+   */
+  touch() {
+    if (this.onTouched) {
+      this.onTouched();
```
DevVersion: The extra method is quite unnecessary, because you are defining a No-Op function for the onTouched variabled, so the variable won't be undefined.

You should be able to use `onTouched` function / variable instead.

src/components/button-toggle/button-toggle.ts
```diff
@@ -1,26 +1,43 @@
...
 
 export type ToggleType = 'checkbox' | 'radio';
 
 
+
+/**
+ * Provider Expression that allows md-button-toggle-group to register as a ControlValueAccessor.
+ * This allows it to support [(ngModel)].
+ */
+const MD_BUTTON_TOGGLE_GROUP_VALUE_ACCESSOR = new Provider(
```
jelbourn: Add `export` here.

jelbourn:
Looks good aside from the two comments

iveysaur:
Address comments @DevVersion and @jelbourn

jelbourn:
LGTM

PR #702, opened 2016-06-16
feat(tabs): adds support for two-way bindings on selectedIndex

### LINKED ISSUE #687: Tabs does not respect setting selectedIndex
We have an md-tab-group with a number of md-tabs and we're setting an initial selected tab using the `selectedIndex` property and the user can then select to reset back to that initial tab. The problem we're finding is that if the user navigates using the labels to other tabs then chooses to reset back to the initial tab it doesn't work.

We've traced through and we can see that `selectedIndex` is never updated by the tab component when the user selects other tabs, so when we try to set `selectedIndex` back to the initial value, the angular change detection kicks in and says nothing changed.

We know this is early days but is this behaviour expected to change? Can we submit a PR to fix it? Or is there a rationale for it being this way?

robertmesserle:
closes #687 

R: @jelbourn

### DIFF
--- a/src/components/tabs/tab-group.spec.ts
+++ b/src/components/tabs/tab-group.spec.ts
@@ -39,15 +39,30 @@ describe('MdTabGroup', () => {
       checkSelectedIndex(0);
 
       // select the second tab
-      let tabLabel = fixture.debugElement.query(By.css('.md-tab-label:nth-of-type(2)'));
+      let tabLabel = fixture.debugElement.queryAll(By.css('.md-tab-label'))[1];
       tabLabel.nativeElement.click();
       checkSelectedIndex(1);
 
       // select the third tab
-      tabLabel = fixture.debugElement.query(By.css('.md-tab-label:nth-of-type(3)'));
+      tabLabel = fixture.debugElement.queryAll(By.css('.md-tab-label'))[2];
       tabLabel.nativeElement.click();
       checkSelectedIndex(2);
     });
+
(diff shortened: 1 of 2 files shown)


src/components/tabs/tab-group.spec.ts
```diff
@@ -49,6 +49,21 @@ describe('MdTabGroup', () => {
       checkSelectedIndex(2);
     });
 
+    it('should support two-way binding for selectedIndex', async(() => {
+      let component = fixture.componentInstance;
+      component.selectedIndex = 0;
+
+      fixture.detectChanges();
+
+      let tabLabel = fixture.debugElement.query(By.css('.md-tab-label:nth-of-type(2)'));
```
jelbourn: Shorter to say
```ts
let tabLabel = fixture.debugElement.query(By.css('.md-tab-label'))[2];
```

src/components/tabs/tabs.ts
```diff
@@ -68,6 +69,10 @@ export class MdTabGroup {
     return this._selectedIndex;
   }
 
+  @Output('selectedIndexChange') private get _selectedIndexChange(): Observable<number> {
```
jelbourn: Add a comment here like 
```ts
/** Output to enable support for two-way binding on `selectedIndex`. */
```

jelbourn:
LGTM

Fiontan:
We cannot get this to work on MAterial ^5.2.4 is this in a subsequent release? Using npm version 5.8.0

Issue #487, opened 2016-05-20
Allow to manage overflow within md-sidenav-layout


cortopy:
* **Do you want to request a *feature* or report a *bug*?**

feature

* **What is the current behavior?**

The following won't work if the layout of the whole application is to take 100% of the height

```SASS
md-sidenav-layout
  height: 100%

  md-content
    height: inherit
```

If there is a Component too large within md-content, the whole page will scroll, including the sidebar if it's open and any top toolbars. In this instance, changing the css of the sidenav component to:

```CSS
  :host > md-content {
    position: relative;
    height: inherit
    -webkit-transform: translate3D(0, 0, 0);
            transform: translate3D(0, 0, 0);
    display: block; }
```

allows adjusting the `overflow` of children components within the main area using percentages. Percentages will not work unless the height of `md-content` is set. The advantage of this approach is that sidenav can stay open and won't scroll. Top toolbars don't scroll either.

* **What is the expected behavior?**

Being able to style md-content to manage heights and widths of components within it.
Adding a `height: inherit` for `md-content`fixes the issue, but I'm not sure hard coding this would be the best solution.

* **What is the motivation / use case for changing the behavior?**

Taking control of how scrolling affects each part of the whole layout. Currently it's an all or nothing approach

* **Which version of Angular and Material, and which browser and OS does this issue affect?**
Did this work in previous versions of Angular / Material?

Material2: 2.0.0-alpha.4
Angular: rc01

hansl:
I'm not sure I follow. For me setting the height to `inherit` does nothing (and it's already inherited). Could you give an example on plunkr or a screenshot please?

ollwenjones:
:+1:  I'm running into a related problem. since the `md-sidenav-layout` ng-contents things into an `md-content` wrapper that I can't seem to target to set height, I'm not able to get app wrapper to consume the viewport. I could be overlooking some obvious CSS selector mistake, but I've burned some real time here on something that should be trivial.

ollwenjones:
Also @hansl the `md-content` height seems to be `initlal` not `inherit` by default. (at least in Chrome.)

ollwenjones:
Reason I was spinning my wheels: failure to understand some shadow DOM things related to the component and CSS scope and tag selectors. Still agree with @cortopy's assessment.

cortopy:
I was just testing this issue with the alpha.5 release and my css seems to be taken into account now:

```
md-sidenav-layout
  height: 100%

  md-content
    height: inherit
```

[...]

PR #703, opened 2016-06-16
fix(sidenav): md-content now fills height

### LINKED ISSUE #606: md-sidenav: fullscreen doesn't respect md-content
Possibly a bug

Given is
```
<md-sidenav-layout fullscreen>
...
</md-sidenav-layout>
```
In DOM the content is wrapped in `md-content`. I would expect `md-content` to be fullscreen as `md-sidenav-layout`, `md-sidenav` and the backdrop are.

`md-sidenav` uses the full hight because of
```
    position: fixed;
    top: 0;
    bottom: 0;
```
`md-content` only respects the width
```
    margin-left: 0px;
    margin-right: 0px;
```

http://plnkr.co/edit/NjimGzKqXw6bKJjk8Pyg?p=preview

To have control over the vertical overflow in the sidenav content. 
To realize scrollable content that consumes the full height.

alpha.5

This is possibly a duplicate of (or at least related to) https://github.com/angular/material2/issues/487
I created a new issue as the use-case and targeted feature (fullscreen attribute vs. styling the height) are different. Although i think the solution could look quiet the same.
I also only found some very "hacky" ways to get the `md-content` becoming fullscreen. Like !important on`:host md-content` to ignore proper view encapsulation or using global css (also ignoring view encapsulation).

robertmesserle:
move overflow from `md-sidenav-layout` to `md-content`
stretches `md-content` to fill available height

closes #606 

R: @jelbourn @hansl

### DIFF
--- a/src/components/sidenav/sidenav.scss
+++ b/src/components/sidenav/sidenav.scss
@@ -60,7 +60,6 @@ $md-sidenav-push-background-color: md-color($md-background, dialog) !default;
   @include md-stacking-context();
 
   box-sizing: border-box;
-  overflow-y: auto;
   -webkit-overflow-scrolling: touch;
 
   // TODO(hansl): Update this with a more robust solution.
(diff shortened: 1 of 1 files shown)


hansl:
Would this also fixes #487?

hansl:
LGTM.

PR #705, opened 2016-06-17
fix(overlay): make sure div is always removed after tests

dozingcat:
This was causing some mkInkRipple positioning tests to fail on mobile Safari (e.g. https://travis-ci.org/angular/material2/jobs/138190850), figured it would be useful to fix separately.


jelbourn:
LGTM

Good catch, thanks!

PR #712, opened 2016-06-17
chore: fix deprecated error

jesselpalmer:
When you run `npm run typings` you get the following error:

```console
typings ERR! deprecated The "ambient" flag is deprecated. Please use "global" instead
```

This PR fixes that error.


jelbourn:
Thanks!

PR #713, opened 2016-06-17
update(): prevent change event to be fired at initialization

### LINKED ISSUE #709: Slide-toggle fires change event on load

Slide-toggle triggers change event on initialization if initial value of "checked" is true.

The event should only be fired in response to an interaction.

The component fires the event before any user action has occurred.

Plunkr to reproduce here: http://plnkr.co/KCZtmelCWK6JKj0FDlh6

If the app triggers an expensive operation, e.g. an API request, in response to the change event, this results in misfired updates and slows load time.

All.

I don't think so. :)

DevVersion:
* Adds a test which confirms, that the slide-toggle isn't firing the (change) event multiple times.

Closes #709

### DIFF
--- a/src/components/slide-toggle/slide-toggle.ts
+++ b/src/components/slide-toggle/slide-toggle.ts
@@ -76,7 +76,7 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
               private _renderer: Renderer) {
   }
 
-  /** @internal */
+  /** TODO: internal */
   ngAfterContentInit() {
     // Mark this component as initialized in AfterContentInit because the initial checked value can
     // possibly be set by NgModel or the checked attribute. This would cause the change event to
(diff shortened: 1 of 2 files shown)


src/components/slide-toggle/slide-toggle.ts
```diff
@@ -74,6 +76,14 @@ export class MdSlideToggle implements ControlValueAccessor {
               private _renderer: Renderer) {
   }
 
+  /** @internal */
```
jelbourn: We figured out in alpha.5 that we can't actually mark life-cycle events as `@internal`, since they then no longer satisfy the corresponding interface when compiled downstream. So I've been doing `TODO: internal` for when TS adds a first class `internal` access modifier.

jelbourn:
LGTM aside from that `@internal` comment

src/components/slide-toggle/slide-toggle.ts
DevVersion: Ah right, I've heard about that in another issue, but I missed that. Thanks again for the clarification, that makes sense.

jelbourn:
LGTM

PR #727, opened 2016-06-20
Update the http dep to rc2

anant-singh:


anant-singh:
I signed it!

jelbourn:
This is captured in #729 which updates everything.

PR #729, opened 2016-06-20
chore: update to angular rc3 & newest router

jelbourn:
R: @hansl or @robertmesserle 

Need the new router to work with offline compiler.


jelbourn:
CI failure is due to https://github.com/angular/vladivostok/issues/95

src/demo-app/index.html
```diff
@@ -5,7 +5,7 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Material2</title>
-  <base href="/">
+  <base href=".">
```
robertmesserle: Why is this necessary?

hansl:
LGTM except for @robertmesserle's comment.

hansl:
And we probably want to wait for that router issue to be fixed since master will become red.

jelbourn:
Yeah, waiting on Victor to fix the circular dep in the router.

src/demo-app/index.html
jelbourn: Angular requires an explicit base tag, and [the docs recommend this value](https://angular.io/docs/ts/latest/guide/router.html).

zigzag95:
Please be careful because base href="." May break apps using webpack.
Why are you using v3 router?

hansl:
We have constant testing with webpack currently. Router v3 is ready for simple cases and Material2 is a good test bed for it.

jelbourn:
@zigzag95 This use of the router is only for our own demo app and doesn't affect any consumers of the components.

zigzag95:
So It must be a peer depency or dev depency

ddomingues:
Hi @zigzag95 and @jelbourn
 
Wouldn't be interesting already update it to angular rc3?

zigzag95:
I think WIP.

PR #734, opened 2016-06-21
Create windows batch file for staging releases

NCC1701M:
Windows user can build material and create a release version.

### DIFF
--- a/scripts/release/stage-release.bat
+++ b/scripts/release/stage-release.bat
@@ -1,0 +1,28 @@
+@echo off
+REM Stages a release by putting everything that should be packaged and released
+REM into the ./deploy folder. This script should be run from the root of the
+REM material2 repo.
+
+REM Make sure you are not running `ng serve` or `ng build --watch` when running this.
+
+
+REM Clear dist/ and deploy/ so that we guarantee there are no stale artifacts.
+rmdir /S /Q dist
+rmdir /S /Q deploy
+
+REM Perform a build with the modified tsconfig.json.
+call ng build
+
+REM Inline the css and html into the component ts files.
+call npm run inline-resources
+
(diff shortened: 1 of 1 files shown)


NCC1701M:
I signed it!

jelbourn:
@NCC1701M why do you need this?

NCC1701M:
For example as a workaround as long as custom themes are not implemented for everyone who want to try material2 but don't want to use the default theme. And for all who want to compile material2 on a windows machine

jelbourn:
LGTM

Perhaps we could delete this once bash support in Windows 10 makes it to the public release.

DevVersion:
@NCC1701M @jelbourn I'm just a bit wondering. 

Normally in other projects (like Material 1 etc.) there will be no BAT scripts. 

The default developer environment on Windows includes a Bash Terminal, like the Git Bash (MingW) or Cygwin. 

Once we add BAT files to projects, we will have to add BAT files for other scripts as well. (Even) if the script is actually doing the same as the Bash Script (just different syntax).

Just throwing some thoughts here.

jelbourn:
@DevVersion I just don't really care that much about having the extra file. I'm not going to take extra steps to maintain it, though.

NCC1701M:
@DevVersion @jelbourn  As I mentioned above it's a workaround as long as theming isn't working. As far as I'm concerned the file can be deleted after implementing theming.

@jelbourn That's ok if there is any problem with the file just contact me. I'll take care.

@DevVersion I am a windows developer and I've never needed a Bash Terminal ;-)

DevVersion:
I'm fine with it. I just wanted to get the initial thoughts here.

It would be probably better to just make it a node file. This would work on all environments.

PR #738, opened 2016-06-21
chore: make the library (more) ready for offline compile

jelbourn:
R: @hansl @robertmesserle 

This fixes the compile errors with the current state of the offline compiler. We're still waiting on some changes from Tobias before we can _run_ the offline compiled output.

This requires:
* Explicitly listing *all* directives and pipes, even the normally ambient ones (because each individual app makes the decision about which things are ambient).
* Export more things than we were
* Not referencing any browser-platform specific stuff (such as `HTMLElement`) outside of class methods (i.e., directly in the module). These changes are are needed to work with Angular Universal.

### DIFF
--- a/src/demo-app/radio/radio-demo.ts
+++ b/src/demo-app/radio/radio-demo.ts
@@ -1,6 +1,7 @@
 import {Component} from '@angular/core';
+import {FORM_DIRECTIVES, NgFor} from '@angular/common';
 import {MdCheckbox} from '@angular2-material/checkbox/checkbox';
-import {MdRadioButton, MdRadioGroup} from '@angular2-material/radio/radio';
+import {MD_RADIO_DIRECTIVES} from '@angular2-material/radio/radio';
 import {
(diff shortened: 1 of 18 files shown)


src/demo-app/radio/radio-demo.ts
```diff
@@ -1,6 +1,7 @@
 import {Component} from '@angular/core';
+import {FORM_DIRECTIVES, NgFor} from '@angular/common';
```
kara: We should probably import FORM_DIRECTIVES from @angular/forms, no?  Unless you want that to be added to the open PR.

robertmesserle:
LGTM

src/demo-app/radio/radio-demo.ts
jelbourn: We should do that in the PR that bumps to new forms.

kara: Haha, I was afraid you'd say that.  I'll add it to my to-dos

PR #739, opened 2016-06-21
fix(checkbox): should match native checkbox behavior

### LINKED ISSUE #575: MdCheckBox is behaving different to native checkbox

robertmesserle:
closes #575 

This appears to work, but I am trying a few different approaches before settling on one.

This approach has one small difference between native behavior: when indeterminate is set to true, the `checked` value is set to false.  This is not how native checkboxes work, but will only ever be seen if you check the `checked` value via JS while `indeterminate` is true.

### DIFF
--- a/src/components/checkbox/checkbox.ts
+++ b/src/components/checkbox/checkbox.ts
@@ -7,6 +7,7 @@ import {
     Output,
     Provider,
     Renderer,
+    ViewChild,
     ViewEncapsulation,
     forwardRef,
     AfterContentInit
@@ -113,6 +114,8 @@ export class MdCheckbox implements AfterContentInit, ControlValueAccessor {
   /** Event emitted when the checkbox's `checked` value changes. */
   @Output() change: EventEmitter<MdCheckboxChange> = new EventEmitter<MdCheckboxChange>();
 
+  @ViewChild('input') _inputElement: ElementRef;
+
   /** Called when the checkbox is blurred. Needed to properly implement ControlValueAccessor. */
   onTouched: () => any = () => {};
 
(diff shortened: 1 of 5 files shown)


robertmesserle:
R: @jelbourn should be ready for review.

src/components/checkbox/checkbox.ts
```diff
@@ -150,7 +153,9 @@ export class MdCheckbox implements AfterContentInit, ControlValueAccessor {
 
       // Only fire a change event if this isn't the first time the checked property is ever set.
       if (this._isInitialized) {
-        this._emitChangeEvent();
+        if (document.activeElement === this._inputElement.nativeElement) {
```
jelbourn: I think using `document.activeElement` _might_ cause problems with Angular Universal. Since the checkbox already keeps track of focus with `hasFocus`, just use that?

src/components/checkbox/checkbox.ts
```diff
@@ -150,7 +153,9 @@ export class MdCheckbox implements AfterContentInit, ControlValueAccessor {
...
+          this._emitChangeEvent();
+        }
```
jelbourn: Can we add a test for this?

jelbourn:
Closing in favor of #820

PR #745, opened 2016-06-22
fix(forms): update packages to use new forms module

### LINKED ISSUE #696: Support @angular/forms
It seems that material does not work with new forms. I have used formControlName directive and I got error that there is no value accessor for certain field in my case 'email':

> No value accessor for 'email'

I have changed import in input/index.js from:

`var common_1 = require('@angular/common');`

to:

`var common_1 = require('@angular/forms');`

and it seems to start working.

kara:
This PR updates components using form imports to use the new forms module.   Should fix #696.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@angular/platform-browser": "2.0.0-rc.3",
     "@angular/platform-browser-dynamic": "2.0.0-rc.3",
     "@angular/router": "v3.0.0-alpha.8",
+    "@angular/forms": "^0.1.0",
     "core-js": "^2.4.0",
     "hammerjs": "^2.0.8",
     "rxjs": "5.0.0-beta.6",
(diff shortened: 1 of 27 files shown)


jelbourn:
LGTM

Will merge once we cut a release with the old forms, and then we'll do another release with the new forms.

mmrath:
There is no harm doing a release with old forms. But since anyway everyone is going to use new forms, it is probably ok to skip the release with old forms. I am eagerly waiting for the next alpha with new forms.

kara:
Note: This will also have to go in after our RC.2 update.

TobiasKrogh:
Hey, great work. Stumbled upon this while looking whether material2 was already working with @angular/forms. Is there a specific reason not to use "@angular/forms" 0.1.1  directly? Just being curious!

PR #748, opened 2016-06-23
Update GETTING_STARTED.md

geevi:
doesnt work without adding .js


kara:
@geevi This works for me without `.js`.  What's not working for you?   The CSS is also necessary to bring in.

PR #750, opened 2016-06-23
feat(slide-toggle): add drag functionality to thumb

DevVersion:
This adds the dragging functionality to the `slide-toggle`.

It works pretty neat. Also tested on older touch devices. Seems like HammerJS is doing a pretty good job.

FYI: I intentionally didn't use the `class` / `style` bindings, because we don't want to trigger a change detection for each Drag Event.

### DIFF
--- a/src/components/slide-toggle/slide-toggle.scss
+++ b/src/components/slide-toggle/slide-toggle.scss
@@ -129,6 +129,12 @@ $md-slide-toggle-margin: 16px !default;
 
   transition: $swift-linear;
   transition-property: transform;
+
+  // Once the thumb container is being dragged around, we remove the transition duration to
+  // make the drag feeling fast and not delayed.
+  &.md-dragging {
+    transition-duration: 0ms;
+  }
 }
 
 // The thumb will be elevated from the slide-toggle bar.
--- a/src/components/slide-toggle/slide-toggle.ts
+++ b/src/components/slide-toggle/slide-toggle.ts
@@ -59,7 +59,7 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   private _hasFocus: boolean = false;
   private _isMousedown: boolean = false;
   private _isInitialized: boolean = false;
-  private _slideRenderer: MdSlideToggleRenderer = null;
+  private _slideRenderer: SlideToggleRenderer = null;
 
   // State of the current drag, which holds required variables for the drag.
   private _dragState: {
@@ -80,13 +80,12 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   // Returns the unique id for the visual hidden input.
   getInputId = () => `${this.id || this._uniqueId}-input`;
 
-  constructor(private _elementRef: ElementRef,
-              private _renderer: Renderer) {
-    this._slideRenderer = new MdSlideToggleRenderer(this._elementRef);
-  }
+  constructor(private _elementRef: ElementRef, private _renderer: Renderer) {}
 
   /** TODO: internal */
   ngAfterContentInit() {
+    this._slideRenderer = new SlideToggleRenderer(this._elementRef);
+
     // Mark this component as initialized in AfterContentInit because the initial checked value can
     // possibly be set by NgModel or the checked attribute. This would cause the change event to
     // be emitted, before the component is actually initialized.
@@ -105,7 +104,7 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
     event.stopPropagation();
 
     // Once a drag is currently in progress, we do not want to toggle the slide-toggle on a click.
-    if (!this.disabled && !this._dragState) {
+    if (!this.disabled && !this._slideRenderer.isDragging()) {
       this.toggle();
     }
   }
@@ -223,53 +222,21 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
 
   /** TODO: internal */
   _onDragStart() {
-    if (this._dragState) {
-      return;
-    }
-
-    let thumbBarRect = this._slideRenderer.getThumbBarClientRect();
-    let thumbRect = this._slideRenderer.getThumbClientRect();
-
-    this._dragState = {
-      barWidth: thumbBarRect.width - thumbRect.width
-    };
-
-    this._slideRenderer.toggleDragging(true);
+    this._slideRenderer.startThumbDrag(this.checked);
   }
 
   /** TODO: internal */
   _onDrag(event: HammerInput) {
-    if (!this._dragState) {
-      return;
-    }
-
-    let percentage = (event.deltaX / this._dragState.barWidth) * 100;
-
-    // When the slide-toggle was initially checked, then we have to start the drag at 100%
-    if (this.checked) {
-      percentage += 100;
-    }
-
-    percentage = Math.max(0, Math.min(percentage, 100));
-
-    this._slideRenderer.updateThumbPosition(percentage);
-    this._dragState.percentage = percentage;
+    this._slideRenderer.updateThumbPosition(event.deltaX);
   }
 
   /** TODO: internal */
   _onDragEnd() {
-    if (!this._dragState) {
-      return;
-    }
-
-    this.checked = this._dragState.percentage > 50;
-
-    this._slideRenderer.updateThumbPosition(null);
-    this._slideRenderer.toggleDragging(false);
-
-    // We have to clear the drag after one tick, because otherwise
-    // the click event will fire and toggle the slide-toggle again.
-    setTimeout(() => { this._dragState = null; }, 0);
+    // Notice that we have to stop outside of the current event handler,
+    // because otherwise the click event will be fired and will reset the new checked variable.
+    setTimeout(() => {
+      this.checked = this._slideRenderer.stopThumbDrag();
+    }, 0);
   }
 
 }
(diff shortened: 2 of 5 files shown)


DevVersion:
@jelbourn Not sure how HammerJS will run on the users Environment later. What were your thoughts?

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -84,6 +92,23 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
...
+    // waiting for the drag to exceed the elements width.
+    hammerManager.add(new Hammer.Pan({
+      threshold: 0,
+      pointers: 0,
+      direction: Hammer.DIRECTION_HORIZONTAL
+    }));
+
+    hammerManager.on('panstart', this._onDragStart.bind(this));
+    hammerManager.on('panmove', this._onDrag.bind(this));
+    hammerManager.on('panend', this._onDragEnd.bind(this));
```
jelbourn: Why are you adding the events directly instead of using handlers defined via the template, e.g.,
```html
<div class="md-slidetoggle-thumb" (dragstart)="..." ...>
```
?

Check out @iveysaur's PR #779 for the slider; she's using the baked-in gesture stuff

jelbourn: Also, I tried this on my phone (Galaxy S6) and it was a bit sluggish. I wonder if we should throttle these drag events (would we need a throttle function in the library, though...)

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -59,6 +61,12 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   private _isMousedown: boolean = false;
   private _isInitialized: boolean = false;
 
+  // Thumb Container Element, which is required for the dragging functionality.
+  @ViewChild('thumbContainer') private _thumbContainerRef: ElementRef;
+
+  // Pointer Object for the dragging functionality of the thumb.
+  private _dragPointer: any;
```
jelbourn: This should have a real type. I also don't quite follow what it's meant to encompass.

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -214,6 +240,54 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
...
+    }
+
+    this.checked = this._dragPointer.percentage > 50;
+
+    this._thumbContainerRef.nativeElement.style.transform = '';
+    this._thumbContainerRef.nativeElement.classList.remove('md-dragging');
+
+    // We have to clear the drag after one tick, because otherwise
+    // the click event will fire and toggle the slide-toggle again.
+    setTimeout(() => { this._dragPointer = null; }, 1);
```
jelbourn: Zero should work, no?

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -214,6 +240,54 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
...
+
+    let percentage = (event.deltaX / this._dragPointer.barWidth) * 100;
+
+    // When the `slide-toggle` is currently checked, we invert the percentage, because it
+    // starts from the right.
+    percentage = this.checked ? 100 + percentage : percentage;
+    percentage = Math.max(0, Math.min(percentage, 100));
+
+    this._dragPointer.percentage = percentage;
+    this._thumbContainerRef.nativeElement.style.transform = `translate3d(${percentage}%, 0, 0)`;
```
jelbourn: Use the `applyTransform` from core/style, which also sets `webkitTransform` for iOS 8

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -214,6 +240,54 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
     this._change.emit(event);
   }
 
+
+  private _onDragStart() {
+    if (this._dragPointer) {
+      return;
+    }
+
+    let thumbBarEl = this._elementRef.nativeElement.querySelector('.md-slide-toggle-bar');
```
jelbourn: Could you move the logic that does the direct DOM manipulation into a separate class called `SlideToggleRenderer`? The constructor should just take the `ElementRef` for the component. The component class can then just call methods on the `slideToggleRenderer` to do the stuff it needs. 

Once we support web-workers with Angular, the main component code and the stuff that touches the DOM will have to be split into separate classes with marshalling between them. Breaking up the code now makes us more ready for that later.

src/components/slide-toggle/slide-toggle.ts
jelbourn: Actually, it's fine for now- can you just add a `TODO` to throttle the events later?

src/components/slide-toggle/slide-toggle.ts
DevVersion: Agreed. The comment is actually pretty weird.

src/components/slide-toggle/slide-toggle.ts
DevVersion: > Why are you adding the events directly instead of using handlers defined via the template, e.g.,
Yeah great idea - as we discussed on Slack

> Also, I tried this on my phone (Galaxy S6) and it was a bit sluggish. I wonder if we should throttle these drag events (would we need a throttle function in the library, though...)
Yeah I fixed that by calculating the absolute difference (as done in the slider)

> Actually, it's fine for now- can you just add a TODO to throttle the events later?
Not sure, why throttling is necessary here? It seems like throttling would probably make it less smooth.

src/components/slide-toggle/slide-toggle.ts
DevVersion: Works pretty neat. Thanks

src/components/slide-toggle/slide-toggle.ts
DevVersion: Make sense. Will change it.

DevVersion:
@jelbourn Addressed your comments
@kara I've implemented the `slide` event as discussed (also the cleanup)

One thing I've noticed:
- When calculating the percentage for the `drag` absolutely, there will be a gap, when starting to drag
  It will look like the drag starts in a higher `x` position.

- Dragging / Sliding should be calculated relatively, by using the distance of the Hammer event.
  The difference is, that the distance will take care of the cursor position (whether you click at the start / end of the thumb)<br/>
  This means, that we miss those "pixels" for the calculation and the drag would not feel smooth / perfectly.

- After some investigation I finally found the perfect way for the thumb drag functionality and it seems like works like a charm.

cc. @iveysaur (Also noticed this in the Slider component)

src/components/slide-toggle/slide-toggle.scss
```diff
@@ -129,6 +129,12 @@ $md-slide-toggle-margin: 16px !default;
 
   transition: $swift-linear;
   transition-property: transform;
+
+  // Once the thumb container is being dragged around, we remove the transition duration to
+  // make the drag feeling fast and not delayed.
+  &.md-dragging {
+    transition-duration: 0ms;
```
jelbourn: Could just be `0` without the `ms` (hopefully the stylelint doesn't want this?)

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -58,6 +59,13 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   private _hasFocus: boolean = false;
   private _isMousedown: boolean = false;
   private _isInitialized: boolean = false;
+  private _domRenderer: MdSlideToggleRenderer = null;
```
jelbourn: Maybe `slideRenderer` or `dragRenderer`?

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -58,6 +59,13 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   private _hasFocus: boolean = false;
   private _isMousedown: boolean = false;
   private _isInitialized: boolean = false;
+  private _domRenderer: MdSlideToggleRenderer = null;
+
+  // Drag pointer, which holds information about the current drag.
+  private _dragPointer: {
```
jelbourn: Maybe `_dragStatus` or `_dragState`?

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -202,13 +212,101 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
...
+  /** Emits the change event to the `change` output EventEmitter */
   private _emitChangeEvent() {
     let event = new MdSlideToggleChange();
     event.source = this;
     event.checked = this.checked;
     this._change.emit(event);
   }
 
+
+  /** @internal */
```
jelbourn: As of #781 we're avoiding `@internal` everywhere now. Mark it as `TODO: internal` for now.

src/components/slide-toggle/slide-toggle.ts
DevVersion: Sounds better.

src/components/slide-toggle/slide-toggle.ts
DevVersion: I like `dragState`

src/components/slide-toggle/slide-toggle.ts
DevVersion: Ah sorry, I still miss this sometimes.

src/components/slide-toggle/slide-toggle.scss
DevVersion: My IDE (Webstorm) is not allowing to have just `zero` here, because it needs a time unit.

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -58,6 +59,13 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   private _hasFocus: boolean = false;
   private _isMousedown: boolean = false;
   private _isInitialized: boolean = false;
+  private _slideRenderer: MdSlideToggleRenderer = null;
+
+  // State of the current drag, which holds required variables for the drag.
```
jelbourn: Nit: use `/** */` style comments for member / method descriptions

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -202,13 +212,101 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
...
+
+    // We have to clear the drag after one tick, because otherwise
+    // the click event will fire and toggle the slide-toggle again.
+    setTimeout(() => { this._dragState = null; }, 0);
+  }
+
+}
+
+/**
+ * Renderer for the Slide Toggle component, which separates DOM modification in it's own class
```
jelbourn: Nit: `its` (without the apostrophe)

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -202,13 +212,101 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
...
+  getThumbBarClientRect(): ClientRect {
+    let thumbBarEl = this._elementRef.nativeElement.querySelector('.md-slide-toggle-bar');
+    return thumbBarEl.getBoundingClientRect();
+  }
+
+  /**
+   * Updates the thumb containers position by using the specified percentage.
+   * When the percentage is set to `null`, the custom thumb position will be removed.
+   */
+  updateThumbPosition(percentage: number) {
```
jelbourn: You could do `percentage: number = null` instead of explicitly passing null.

src/components/slide-toggle/slide-toggle.ts
```diff
@@ -58,6 +59,13 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   private _hasFocus: boolean = false;
   private _isMousedown: boolean = false;
   private _isInitialized: boolean = false;
+  private _slideRenderer: MdSlideToggleRenderer = null;
+
+  // State of the current drag, which holds required variables for the drag.
+  private _dragState: {
+    barWidth: number;
+    percentage?: number;
+  };
```
jelbourn: What do you think of moving the dragState values to be properties of the SliderRenderer? The `updateThumbPosition` function would change to take the `deltaX` from the event. Your event handlers could then look something like
```ts
_onSlideStart() {
  this._slideRenderer.startSliding();
}

_onSlide(slideEvent: HammerInput) {
  this._slideRenderer.updateThumbPosition(event.deltaX);
}

_onSlideEnd() {
  this._slideRenderer.stopSliding();
}
```

and the renderer could look something like
```ts
class SlideRenderer {
  private _thumb: HTMLElement;
  private _slideBar: HTMLElement;
  private _slideBarWidth: number;

  constructor(elementRef: ElementRef) {
    this._thumb = elementRef.nativeElement.querySelector('.md-slide-toggle-thumb-container');
    this._slideBar = elementRef.nativeElement.querySelector('.md-slide-toggle-bar');
  }

  isSliding(): boolean {
    return !!this._slideBarWidth;
  }

  startSliding(): void {
    if (!this.isSliding()) {
      // We can't grab the width until we actually start sliding in order to work with server-side pre-rendering.
      // Assume that the width of the slide bar doesn't change during one slide.
      this._sldieBarWidth = this._slideBar.clientWidth - this._thumb.clientWidth;
    }
  }

  stopSliding(): void {
    if (this.isSliding()) {
      this._slideBarWidth = null;
      this._clearThumbPosition();
    }
  }

  private _clearThumbPosition(): void {
    applyCssTransform(this._thumb, '');
  }

  updateThumbPosition(deltaX: number): void {
    let transformPercent = this._getThumbTransformPercent(deltaX);
    applyCssTransform(this._thumb, `translate3d(${transformPercent}%, 0, 0)`);
  }

  private _getThumbTransformPercent(deltaX: number): void {
    let percentage = (event.deltaX / this._dragState.barWidth) * 100;
    // When the slide-toggle was initially checked, then we have to start the drag at 100%
    if (this.checked) {
      percentage += 100;
    }

    return Math.max(0, Math.min(percentage, 100));    
  }
}
```
...okay, got a little carried away, but I think this is good :smiley: . (I wrote it in GitHub, so there might be some typos). 

Notes on the above:
* All of the sliding logic is encapsulated in the slide renderer.
* We only `querySelector` once per slider instance, since the elements won't ever change. 
* `clientWidth` should do the same job as `getBoundingClientRect().width` in this case
* Function names are more straightforward and the functions are more single-responsibility
* I left out the comments for brevity's sake in this comment

DevVersion: @jelbourn Updated the PR, and hopefully matched your expected result.

jelbourn:
LGTM!

Issue #113, opened 2016-02-27
Overlay service


jelbourn:
- [x] Portals and PortalHosts
- [x] DomPortalHost
- [x] Skeleton Overlay service
- [ ] Overlay sizing
- [x] Overlay global positioning
- [x] Overlay relative positioning
- [ ] Overlay scrolling
- [ ] Overlay animation

zigzag95:
How can we help to close this issue and start interesting components

ratscrew:
would it be possible to return the component from the promise after its attached:

line 14 from https://github.com/angular/material2/blob/master/src/core/overlay/overlay-ref.ts
```
  attach(portal: Portal<any>): Promise<any> {
    return this._portalHost.attach(portal).then(() => {
      this._updatePosition();
    });
  }
``` 

would there be any objection to making it:
```
  attach(portal: Portal<any>): Promise<any> {
    return this._portalHost.attach(portal).then((_comp) => {
      this._updatePosition();
      return _comp;
    });
  }
``` 

my current use case is for a right click context menu that I need to bind to after its attached.

jelbourn:
@ratscrew I might be possible to play with the generics a bit to cut out some of the `any`; it's definitely something I want to explore before I consider the portal / overlay stuff to be mature enough to release on its own.

DennisSmolek:
+1 for @ratscrew's issue.. I was also just tearing my hair out trying to figure out how to access the component ref of the resulting portal component when I used the ComponentPortal. 

I'm re-writing a set of components that use the dynamicComponentLoader and pass values from a directive to the resulting popup but I can't seem to access the ComponentRef any other way. For the time being I'm using the modded version..

jelbourn:
Closing this issue for now, going to track remaining issues in smaller issues.

PR #751, opened 2016-06-23
Edit to overlay ref

ratscrew:
have attach return the component

### DIFF
--- a/src/core/overlay/overlay-ref.ts
+++ b/src/core/overlay/overlay-ref.ts
@@ -12,8 +12,9 @@ export class OverlayRef implements PortalHost {
       private _state: OverlayState) { }
 
   attach(portal: Portal<any>): Promise<any> {
-    return this._portalHost.attach(portal).then(() => {
+    return this._portalHost.attach(portal).then((_comp) => {
       this._updatePosition();
+      return _comp;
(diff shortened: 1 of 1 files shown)


ratscrew:
I signed it!

src/core/overlay/overlay-ref.ts
```diff
@@ -12,8 +12,9 @@ export class OverlayRef implements PortalHost {
       private _state: OverlayState) { }
 
   attach(portal: Portal<any>): Promise<any> {
-    return this._portalHost.attach(portal).then(() => {
+    return this._portalHost.attach(portal).then((_comp) => {
       this._updatePosition();
+      return _comp;
```
DevVersion: I can't see the point of that change. 

The resolved value will be automatically returned, so there is no need to pass it as an attribute and return it "again".

ratscrew:
@DevVersion the component is not being returned without the line, others are having the same issue: #113

DevVersion:
@ratscrew Yeah, just took a look at the issue and now it makes sense.

jelbourn:
I also fixed this behavior in #761, closing this in favor of that one (which should be merging very soon).

Issue #753, opened 2016-06-24
the @angular2/icon used a fix version of dependency of @angular2/http


hantsy:
Whend I installed the lastest Angular 2 material icon(2.0.0-alpha.5-2) into my project, I found there is an UNMET error caused by the dependency declared in  "@angular/http": "2.0.0-rc.1".

It used a **fixed** version.

I have upgraded my  Angular to 2.0.0-RC.3 in my project.  It should be compatible with Angular to 2.0.0-RC.3.

dweitz43:
i think this was indeed fixed on github but maybe the packages have not been updated in the npm registry? the icon package tarball that is downloaded from the npm registry still specifies 2.0.0-rc.1 as the version for the angular http peer dependency

hantsy:
@dweitz43 I have updated the latest alpha.6, but icon still depends on a fixed version http 2.0.0.-rc.3, I have to try to upgrade Angular to 2.0.0-rc.4, which will cause another new unmet error.

dweitz43:
@hantsy indeed, it appears as if alpha.6 is still dependent upon the older rc, i guess we will have to wait for it to be updated to reference rc.4...

simonykq:
+1  
Same here. Not it requires ** @angular/http@2.0.0-rc.3**, it should be updated to ** @angular/http@2.0.0-rc.4**, which is the latest version by this far

valentinlup:
+1

jelbourn:
This should be resolved now.

PR #754, opened 2016-06-24
issue753:changed angular2/http dep version of icon compnent to ^2.0.0-rc.3

hantsy:
address #753


jelbourn:
This will be captured in #727 (which will be bumped to rc3) once the new version of the router with its circular dependency removed is released to npm.

PR #761, opened 2016-06-24
feat(dialog): inital framework for md-dialog

jelbourn:
R: @hansl @kara @robertmesserle 
CC: @iveysaur

* Adds the ability to pass a custom injector with `ComponentPortal`
* Adds `MdDialog` service that can open a dialog (and literally nothing else). Additional APIs will come in subsequent PRs. Dialog is also yet unstyled.

The `setTimeout` hack in the dialog tests is an unfortunate consequence of an issue with either zones or (more likely) the testing infrastructure built on top of it. I'm going to sit down with Misko when he gets back from vacation to get to the root of it.

### DIFF
--- a/src/components/dialog/dialog-container.html
+++ b/src/components/dialog/dialog-container.html
@@ -1,0 +1,1 @@
+<template portalHost></template>
--- a/src/components/dialog/dialog-container.ts
+++ b/src/components/dialog/dialog-container.ts
@@ -7,6 +7,7 @@ import {
 import {PortalHostDirective} from '@angular2-material/core/portal/portal-directives';
 import {PromiseCompleter} from '@angular2-material/core/async/promise-completer';
 import {MdDialogConfig} from './dialog-config';
+import {MdDialogContentAlreadyAttachedError} from './dialog-errors';
 
 
 /**
@@ -49,20 +50,24 @@ export class MdDialogContainer extends BasePortalHost implements AfterViewInit {
 
         this._deferredAttachPortal = null;
         this._deferredAttachCompleter = null;
-      });
+      }, () => this._deferredAttachCompleter.reject());
     }
   }
 
   /** Attach a portal as content to this dialog container. */
   attachComponentPortal<T>(portal: ComponentPortal<T>): Promise<ComponentRef<T>> {
     if (this._portalHost) {
+      if (this._portalHost.hasAttached()) {
+        throw new MdDialogContentAlreadyAttachedError();
+      }
+
       return this._portalHost.attachComponentPortal(portal);
     } else {
       // The @ViewChild query for the portalHost is not resolved until AfterViewInit, but this
       // function may be called before this lifecycle event. As such, we defer the attachment of
       // the portal until AfterViewInit.
       if (this._deferredAttachCompleter) {
-        this._deferredAttachCompleter.reject();
+        throw new MdDialogContentAlreadyAttachedError();
       }
 
       this._deferredAttachPortal = portal;
--- a/src/components/dialog/dialog-container.ts
+++ b/src/components/dialog/dialog-container.ts
@@ -50,7 +50,11 @@ export class MdDialogContainer extends BasePortalHost implements AfterViewInit {
 
         this._deferredAttachPortal = null;
         this._deferredAttachCompleter = null;
-      }, () => this._deferredAttachCompleter.reject());
+      }, () => {
+        this._deferredAttachCompleter.reject();
+        this._deferredAttachCompleter = null;
+        this._deferredAttachPortal = null;
+      });
     }
   }
 
--- a/src/components/dialog/dialog-container.ts
+++ b/src/components/dialog/dialog-container.ts
@@ -61,6 +61,10 @@ export class MdDialogContainer extends BasePortalHost implements AfterViewInit {
       // The @ViewChild query for the portalHost is not resolved until AfterViewInit, but this
       // function may be called before this lifecycle event. As such, we defer the attachment of
       // the portal until AfterViewInit.
+      if (this._deferredAttachCompleter) {
+        this._deferredAttachCompleter.reject();
+      }
+
       this._deferredAttachPortal = portal;
       this._deferredAttachCompleter = new PromiseCompleter();
       return this._deferredAttachCompleter.promise;
--- a/src/components/dialog/dialog-container.ts
+++ b/src/components/dialog/dialog-container.ts
@@ -1,0 +1,73 @@
+import {Component, ComponentRef, ViewChild, AfterViewInit} from '@angular/core';
+import {
+  BasePortalHost,
+  ComponentPortal,
+  TemplatePortal
+} from '@angular2-material/core/portal/portal';
+import {PortalHostDirective} from '@angular2-material/core/portal/portal-directives';
+import {PromiseCompleter} from '@angular2-material/core/async/promise-completer';
+import {MdDialogConfig} from './dialog-config';
+
+
+/**
+ * Internal component that wraps user-provided dialog content.
+ */
+@Component({
+  moduleId: module.id,
+  selector: 'md-dialog-container',
+  templateUrl: 'dialog-container.html',
+  styleUrls: ['dialog-container.css'],
+  directives: [PortalHostDirective],
+  host: {
+    'class': 'md-dialog-container',
+    '[attr.role]': 'dialogConfig?.role'
+  }
+})
+export class MdDialogContainer extends BasePortalHost implements AfterViewInit {
+  /** The portal host inside of this container into which the dialog content will be loaded. */
+  @ViewChild(PortalHostDirective) private _portalHost: PortalHostDirective;
+
+  /**
+   * Completer used to resolve the promise for cases when a portal is attempted to be attached,
+   * but AfterViewInit has not yet occured.
+   */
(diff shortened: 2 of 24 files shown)


src/components/dialog/dialog-container.ts
```diff
@@ -0,0 +1,70 @@
+import {Component, ComponentRef, ViewChild, AfterViewInit} from '@angular/core';
+import {BasePortalHost, ComponentPortal, TemplatePortal} from '../../core/portal/portal';
+import {PortalHostDirective} from '../../core/portal/portal-directives';
+import {PromiseCompleter} from '../../core/async/promise-completer';
```
DevVersion: Shouldn't those imports use the NPM scope package instead?

As @hansl did in 9e308a6fa2d263a117e6a147cfab29387797b46c?

jelbourn: Yep, I just forgot to correct what WebStorm auto-imported. Fixed.

src/components/dialog/dialog-container.html
```diff
@@ -0,0 +1 @@
+<template portalHost></template>
```
hansl: 1 liner should probably be inlined.

jelbourn: There will be more; this is just the first PR

src/components/dialog/dialog-container.ts
```diff
@@ -0,0 +1,70 @@
+import {Component, ComponentRef, ViewChild, AfterViewInit} from '@angular/core';
+import {BasePortalHost, ComponentPortal, TemplatePortal} from '../../core/portal/portal';
```
iveysaur: Is there a reason this one uses `../../core/portal/` rather than `@angular2-material/core/portal/` ?

robertmesserle:
LGTM aside from the imports that @iveysaur pointed out (which happens in a few files).

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  /**
+   * Attaches the user-provided component to the already-created MdDialogContainer.
+   * @param component The type of component being loaded into the dialog.
+   * @param containerRef Reference to the wrapping MdDialogContainer.
+   * @returns A promise resolving to the MdDialogRef that should be returned to the user.
+   */
+  private _attachDialogContent<T>(
+      component: {new (): T},
```
hansl: This type differs than the component declaration above.

src/core/overlay/overlay-ref.ts
```diff
@@ -12,9 +12,12 @@ export class OverlayRef implements PortalHost {
       private _state: OverlayState) { }
 
   attach(portal: Portal<any>): Promise<any> {
-    return this._portalHost.attach(portal).then(() => {
+    let attachPromise = this._portalHost.attach(portal);
```
hansl: Revert this file?

src/components/dialog/dialog-container.ts
hansl: Even further; it should be `@angular2-material/core` since it will be bundled like Angular.

DevVersion: I thought it be worth to add some checks for that. Also with auto-suggestion. See #764 
This would probably help avoiding those common mistakes.

Lavan:
I think it could be useful to split this PR into two parts; one for the generic portal implentation and one for the dialog implementation that uses it. The reason for this split is simple to enable injection into portal objects as soon as possible without waiting for the dialog implementation which could take a while.

src/core/overlay/overlay-ref.ts
jelbourn: This change is necessary to make the dialog work.

jelbourn:
@robertmesserle @hansl addressed comments

src/components/dialog/dialog-container.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  /** The dialog configuration. */
+  dialogConfig: MdDialogConfig;
+
+  /** TODO: internal */
+  ngAfterViewInit() {
+    // If there was an attempted call to `attachComponentPortal` before this lifecycle stage,
+    // we actually perform the attachment now that the `@ViewChild` is resolved.
+    if (this._deferredAttachCompleter) {
+      this.attachComponentPortal(this._deferredAttachPortal).then(componentRef => {
```
hansl: catch and reject?

src/components/dialog/dialog-container.ts
```diff
@@ -0,0 +1,73 @@
...
+
+        this._deferredAttachPortal = null;
+        this._deferredAttachCompleter = null;
+      });
+    }
+  }
+
+  /** Attach a portal as content to this dialog container. */
+  attachComponentPortal<T>(portal: ComponentPortal<T>): Promise<ComponentRef<T>> {
+    if (this._portalHost) {
```
hansl: If this is called twice before the promise resolve, it seems like it could be a bug (the first promise will never resolve). You should return the original promise if `_deferredAttachPortal` is still there. Also, a unit test for this particular use case would be nice :)

src/components/dialog/dialog-injector.ts
```diff
@@ -0,0 +1,16 @@
+import {Injector} from '@angular/core';
+import {MdDialogRef} from './dialog-ref';
+
+
+/** Custom injector type specifically for instantiating components with a dialog. */
+export class DialogInjector implements Injector {
```
hansl: Not a blocker, but having an `Injector` with a `set(token, notFoundValue)` method would be a nice addition to angular/core.

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,113 @@
...
+// TODO(jelbourn): focus trapping
+// TODO(jelbourn): potentially change API from accepting component constructor to component factory.
+
+
+
+/**
+ * Service to open Material Design modal dialogs.
+ */
+@Injectable()
+export class MdDialog {
```
hansl: This is in `components/...` but is not a component. As a general discussion should we have a naming convention for services that aren't in core? Is there going to be an `MdDialog` that's actually a component?

src/components/dialog/dialog-container.ts
```diff
@@ -0,0 +1,73 @@
...
+      return this._portalHost.attachComponentPortal(portal);
+    } else {
+      // The @ViewChild query for the portalHost is not resolved until AfterViewInit, but this
+      // function may be called before this lifecycle event. As such, we defer the attachment of
+      // the portal until AfterViewInit.
+      this._deferredAttachPortal = portal;
```
hansl: Add an assertion here that there's a portal so we can see the bug if it happens.

src/core/portal/portal.ts
```diff
@@ -60,21 +67,28 @@ export abstract class Portal<T> {
...
   /**
    * [Optional] Where the attached component should live in Angular's *logical* component tree.
    * This is different from where the component *renders*, which is determined by the PortalHost.
    * The origin necessary when the host is outside of the Angular application context.
    */
-  public viewContainerRef: ViewContainerRef;
+  viewContainerRef: ViewContainerRef;
+
+  /** [Optional] Injector used for the instantiation of the component. */
+  injector: Injector;
```
hansl: `injector?: Injector`

hansl:
Outside of the assertion and good citizenry regarding promises, LGTM.

src/components/dialog/dialog.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    });
+
+    detectChangesForDialogOpen(viewContainerFixture);
+  }));
+});
+
+
+/** Runs the necessary detectChanges for a dialog to complete its opening. */
+function detectChangesForDialogOpen(fixture: ComponentFixture<ComponentWithChildViewContainer>) {
+  // TODO(jelbourn): figure out why the test zone is "stable" when where are still pending
```
robertmesserle: where => there

src/core/portal/portal.ts
jelbourn: That syntax is only for an `interface` (saying that implementing it is optional). When a _value_ is optional, it just means that the value can be `null` or `undefined`

src/components/dialog/dialog.ts
jelbourn: As discussed, treating this as "UI components" more than `@Components`s

jelbourn:
Comments addressed. Needs LGTM label so that caretaker knows to merge it.

src/components/dialog/dialog-container.ts
```diff
@@ -0,0 +1,82 @@
...
+  ngAfterViewInit() {
+    // If there was an attempted call to `attachComponentPortal` before this lifecycle stage,
+    // we actually perform the attachment now that the `@ViewChild` is resolved.
+    if (this._deferredAttachCompleter) {
+      this.attachComponentPortal(this._deferredAttachPortal).then(componentRef => {
+        this._deferredAttachCompleter.resolve(componentRef);
+
+        this._deferredAttachPortal = null;
+        this._deferredAttachCompleter = null;
+      }, () => this._deferredAttachCompleter.reject());
```
hansl: Nit: set to null?

jelbourn: Done.

hansl:
LGTM.

jonathan-casarrubias:
wahoo!!! 🎉

PR #772, opened 2016-06-27
Update package.json

sceendy:
With rc3 out now, it'd be nice for icons to work with it.


sceendy:
nevermind, closing. Looks like #729 handles this.

PR #779, opened 2016-06-28
feat(md-slider): initial version for md-slider

iveysaur:
R: @jelbourn @kara 
CC: @robertmesserle @hansl 

Supports:
* Min
* Max
* Value
* Disabled
* Set value from click
* Set value from drag
* Transition on click, no transition on drag (expected)
* Scale thumb on click and drag

### DIFF
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -1,20 +1,22 @@
 @import 'default-theme';
+@import '_variables';
 
 // This refers to the thickness of the slider. On a horizontal slider this is the height, on a
 // vertical slider this is the width.
-$slider-thickness: 20px !default;
-$slider-min-size: 128px !default;
+$md-slider-thickness: 20px !default;
+$md-slider-min-size: 128px !default;
+$md-slider-padding: 8px !default;
 
-$slider-track-height: 2px !default;
-$slider-thumb-size: 20px !default;
+$md-slider-track-height: 2px !default;
+$md-slider-thumb-size: 20px !default;
 
-$slider-thumb-default-scale: 0.7 !default;
-$slider-thumb-focus-scale: 1 !default;
+$md-slider-thumb-default-scale: 0.7 !default;
+$md-slider-thumb-focus-scale: 1 !default;
 
 // TODO(iveysaur): Find an implementation to hide the track under a disabled thumb.
-$slider-off-color: rgba(black, 0.26);
-$slider-focused-color: rgba(black, 0.38);
-$slider-disabled-color: rgba(black, 0.26);
+$md-slider-off-color: rgba(black, 0.26);
+$md-slider-focused-color: rgba(black, 0.38);
+$md-slider-disabled-color: rgba(black, 0.26);
 
 /**
  * Uses a container height and an item height to center an item vertically within the container.
@@ -26,17 +28,17 @@ $slider-disabled-color: rgba(black, 0.26);
 /**
  * Positions the thumb based on its width and height.
  */
-@mixin slider-thumb-position($width: $slider-thumb-size, $height: $slider-thumb-size) {
+@mixin slider-thumb-position($width: $md-slider-thumb-size, $height: $md-slider-thumb-size) {
   position: absolute;
-  top: center-vertically($slider-thickness, $height);
+  top: center-vertically($md-slider-thickness, $height);
   width: $width;
   height: $height;
   border-radius: max($width, $height);
 }
 
 md-slider {
-  height: $slider-thickness;
-  min-width: $slider-min-size;
+  height: $md-slider-thickness;
+  min-width: $md-slider-min-size;
   position: relative;
   padding: 0;
   display: inline-block;
@@ -56,12 +58,12 @@ md-slider *::after {
 .md-slider-wrapper {
   width: 100%;
   height: 100%;
-  padding-left: 8px;
-  padding-right: 8px;
+  padding-left: $md-slider-padding;
+  padding-right: $md-slider-padding;
 }
 
 /**
- * Holds the isActive and isDragging classes as well as help with positioning the children.
+ * Holds the isActive and isDragging classes as well as helps with positioning the children.
  * Cannot be merged with .md-slider-wrapper.
  */
 .md-slider-container {
@@ -71,8 +73,8 @@ md-slider *::after {
 .md-slider-track-container {
   width: 100%;
   position: absolute;
-  top: center-vertically($slider-thickness, $slider-track-height);
-  height: $slider-track-height;
+  top: center-vertically($md-slider-thickness, $md-slider-track-height);
+  height: $md-slider-track-height;
 }
 
 .md-slider-track {
@@ -80,12 +82,12 @@ md-slider *::after {
   left: 0;
   right: 0;
   height: 100%;
-  background-color: $slider-off-color;
+  background-color: $md-slider-off-color;
 }
 
 .md-slider-track-fill {
-  transition-duration: 400ms;
-  transition-timing-function: cubic-bezier(0.25, 0.8, 0.25, 1);
+  transition-duration: $swift-ease-out-duration;
+  transition-timing-function: $swift-ease-out-timing-function;
   transition-property: width, height;
   background-color: md-color($md-accent);
 }
@@ -95,29 +97,29 @@ md-slider *::after {
   left: 0;
   top: 50%;
   transform: translate3d(-50%, -50%, 0);
-  transition-duration: 400ms;
-  transition-timing-function: cubic-bezier(0.25, 0.8, 0.25, 1);
+  transition-duration: $swift-ease-out-duration;
+  transition-timing-function: $swift-ease-out-timing-function;
   transition-property: left, bottom;
 }
 
 .md-slider-thumb-position {
-  transition: transform 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
+  transition: transform $swift-ease-out-duration $swift-ease-out-timing-function;
 }
 
 .md-slider-thumb {
   z-index: 1;
 
-  @include slider-thumb-position($slider-thumb-size, $slider-thumb-size);
-  transform: scale($slider-thumb-default-scale);
-  transition: transform 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
+  @include slider-thumb-position($md-slider-thumb-size, $md-slider-thumb-size);
+  transform: scale($md-slider-thumb-default-scale);
+  transition: transform $swift-ease-out-duration $swift-ease-out-timing-function;
 }
 
 .md-slider-thumb::after {
   content: '';
   position: absolute;
-  width: $slider-thumb-size;
-  height: $slider-thumb-size;
-  border-radius: max($slider-thumb-size, $slider-thumb-size);
+  width: $md-slider-thumb-size;
+  height: $md-slider-thumb-size;
+  border-radius: max($md-slider-thumb-size, $md-slider-thumb-size);
   border-width: 3px;
   border-style: solid;
   transition: inherit;
@@ -132,10 +134,10 @@ md-slider *::after {
 }
 
 .md-slider-active .md-slider-thumb {
-  transform: scale($slider-thumb-focus-scale);
+  transform: scale($md-slider-thumb-focus-scale);
 }
 
 .md-slider-disabled .md-slider-thumb::after {
-  background-color: $slider-disabled-color;
-  border-color: $slider-disabled-color;
+  background-color: $md-slider-disabled-color;
+  border-color: $md-slider-disabled-color;
 }
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -1,6 +1,6 @@
 @import 'default-theme';
 
-$slider-size: 48px !default;
+$slider-size: 20px !default;
 $slider-min-size: 128px !default;
 
 $slider-track-height: 2px !default;
@@ -23,12 +23,11 @@ md-slider {
   height: $slider-size;
   min-width: $slider-min-size;
   position: relative;
-  margin-left: 4px;
-  margin-right: 4px;
   padding: 0;
-  display: block;
+  display: inline-block;
   flex-direction: row;
   outline: none;
+  vertical-align: middle;
 }
 
 md-slider *,
@@ -39,6 +38,8 @@ md-slider *::after {
 .md-slider-wrapper {
   width: 100%;
   height: 100%;
+  padding-left: 8px;
+  padding-right: 8px;
 }
 
 .md-slider-container {
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -67,3 +67,7 @@ md-slider *, md-slider *:after {
   background-color: blue;
   border-color: blue;
 }
+
+.md-slider-dragging {
+  transition: none;
+}
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -31,7 +31,8 @@ md-slider {
   outline: none;
 }
 
-md-slider *, md-slider *:after {
+md-slider *,
+md-slider *::after {
   box-sizing: border-box;
 }
 
@@ -83,7 +84,7 @@ md-slider *, md-slider *:after {
   transition: all 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
 }
 
-.md-slider-thumb:after {
+.md-slider-thumb::after {
   content: '';
   position: absolute;
   width: $slider-thumb-width;
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -1,11 +1,12 @@
 @import 'default-theme';
 
-$slider-size: 20px !default;
+// This refers to the thickness of the slider. On a horizontal slider this is the height, on a
+// vertical slider this is the width.
+$slider-thickness: 20px !default;
 $slider-min-size: 128px !default;
 
 $slider-track-height: 2px !default;
-$slider-thumb-width: 20px !default;
-$slider-thumb-height: $slider-thumb-width !default;
+$slider-thumb-size: 20px !default;
 
 $slider-thumb-default-scale: 0.7 !default;
 $slider-thumb-focus-scale: 1 !default;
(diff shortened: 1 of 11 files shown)


src/components/slider/slider.scss
```diff
@@ -0,0 +1,120 @@
...
+  border-radius: max($width, $height);
+}
+
+md-slider {
+  height: $slider-size;
+  min-width: $slider-min-size;
+  position: relative;
+  padding: 0;
+  display: inline-block;
+  flex-direction: row;
```
robertmesserle: `flex-direction` + `inline-block`?

src/components/slider/slider.scss
```diff
@@ -0,0 +1,120 @@
...
+
+$slider-thumb-default-scale: 0.7 !default;
+$slider-thumb-focus-scale: 1 !default;
+
+// TODO(iveysaur): Find an implementation to hide the track under a disabled thumb.
+$slider-off-color: rgba(black, 0.26);
+$slider-focused-color: rgba(black, 0.38);
+$slider-disabled-color: rgba(black, 0.26);
+
+@mixin slider-thumb-position($width: $slider-thumb-width, $height: $slider-thumb-height) {
```
robertmesserle: This only seems to be used in one place (with the default values).  Is this necessary?

src/components/slider/slider.scss
```diff
@@ -0,0 +1,120 @@
...
+}
+
+.md-slider-container {
+  position: relative;
+}
+
+.md-slider-track-container {
+  width: 100%;
+  position: absolute;
+  top: ($slider-size / 2) - ($slider-track-height) / 2;
```
robertmesserle: The parentheses look wrong on the 2nd half.  This also looks like it could be abstracted into a function (along with line 20) since they appear to use the same math.

src/components/slider/slider.scss
```diff
@@ -0,0 +1,120 @@
...
+.md-slider-track {
+  position: absolute;
+  left: 0;
+  right: 0;
+  height: 100%;
+  background-color: $slider-off-color;
+}
+
+.md-slider-track-fill {
+  transition: all 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
```
robertmesserle: I think it might make sense to either remove `all` (which should work the same way) or specify `transition-duration` and `transition-timing-function` separately to make it more obvious what is going on.

src/components/slider/slider.scss
iveysaur: Currently it is only used in one place, in the future it should be used more (focus ring). I can remove it for now if that makes more sense? I based this off of material 1 😄

src/components/slider/slider.scss
iveysaur: I think I had something else in there and then remove it without removing the parentheses. I didn't even realize sass functions were a thing either. Fixed with next push.

src/components/slider/slider.scss
```diff
@@ -0,0 +1,125 @@
+@import 'default-theme';
+
+$slider-size: 20px !default;
+$slider-min-size: 128px !default;
+
+$slider-track-height: 2px !default;
+$slider-thumb-width: 20px !default;
+$slider-thumb-height: $slider-thumb-width !default;
```
jelbourn: Combine these two variables into `$slider-thumb-size` ?

src/components/slider/slider.scss
```diff
@@ -0,0 +1,125 @@
+@import 'default-theme';
+
+$slider-size: 20px !default;
```
jelbourn: What does the `$slider-size` represent? Add a comment?

src/components/slider/slider.scss
```diff
@@ -0,0 +1,125 @@
...
+
+$slider-thumb-default-scale: 0.7 !default;
+$slider-thumb-focus-scale: 1 !default;
+
+// TODO(iveysaur): Find an implementation to hide the track under a disabled thumb.
+$slider-off-color: rgba(black, 0.26);
+$slider-focused-color: rgba(black, 0.38);
+$slider-disabled-color: rgba(black, 0.26);
+
+@function center-vertically($outerHeight, $innerHeight) {
```
jelbourn: Add function description comment

src/components/slider/slider.scss
```diff
@@ -0,0 +1,125 @@
...
+  @return ($outerHeight / 2) - ($innerHeight / 2);
+}
+
+@mixin slider-thumb-position($width: $slider-thumb-width, $height: $slider-thumb-height) {
```
jelbourn: Add mixin description comment

src/components/slider/slider.scss
```diff
@@ -0,0 +1,125 @@
...
+  outline: none;
+  vertical-align: middle;
+}
+
+md-slider *,
+md-slider *::after {
+  box-sizing: border-box;
+}
+
+.md-slider-wrapper {
```
jelbourn: Add comments for `slider-wrapper` and `slider-container` since it's not immediately obvious what they're used for.

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+  /** The maximum value that the slider can have. */
+  private _max: number = 100;
+
+  /** The percentage of the slider that coincides with the value. */
+  private _percent: number = 0;
+
+  /**
+   * Whether or not the thumb is currently being dragged.
+   * Used to determine if there should be a transition for the thumb and fill track.
+   * @internal
```
jelbourn: You can remove all use of `@internal`. We discovered last week that it breaks when used with the offline compiler.

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  /**
+   * Whether or not the slider is active (clicked or is being dragged).
+   * Used to shrink and grow the thumb as according to the Material Design spec.
+   * @internal
+   */
+  isActive: boolean = false;
+
+  /** Indicator for if the value has been set or not. */
+  private _isValueInitialized: boolean = false;
```
jelbourn: Would it work to just have `_isInitialized`?

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+  private _value: number = 0;
+
+  @Input()
+  @HostBinding('class.md-slider-disabled')
+  get disabled(): boolean {
+    return this._disabled;
+  }
+
+  set disabled(value) {
+    this._disabled = (value != null && value !== false) ? true : null;
```
jelbourn: You can use the `BooleanFieldValue` for this (for now)

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+  isActive: boolean = false;
+
+  /** Indicator for if the value has been set or not. */
+  private _isValueInitialized: boolean = false;
+
+  /** Value of the slider. */
+  private _value: number = 0;
+
+  @Input()
+  @HostBinding('class.md-slider-disabled')
```
jelbourn: Should probably also have a `@HostBinding` for `attr.aria-disabled`

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+    return this._value;
+  }
+
+  set value(v: number) {
+    this._value = Number(v);
+    this._isValueInitialized = true;
+    this.updatePercentFromValue();
+  }
+
+  constructor(private _elementRef: ElementRef) {
```
jelbourn: Don't make the `ElementRef` a private class property; this will help make sure you only use it inside of the `SliderRenderer`

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+    this._renderer = new SliderRenderer(_elementRef);
+  }
+
+  /**
+   * Once the slider has rendered, grab the dimensions and update the position of the thumb and
+   * fill track.
```
jelbourn: Mark as "TODO: internal"

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  /**
+   * Return a number between two numbers.
+   */
+  clamp(value: number, min = 0, max = 1) {
+    return Math.max(min, Math.min(value, max));
+  }
+}
+
+export class SliderRenderer {
```
jelbourn: Add class descripion

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    let position = percent * width;
+    // The thumb needs to be shifted to the left by half of the width of itself so that it centers
+    // on the value.
+    let thumbPosition = position - (thumbWidth / 2);
+
+    fillTrackElement.style.width = `${position}px`;
+    thumbPositionElement.style.transform = `translateX(${thumbPosition}px)`;
+    // Mobile Safari 8.0 does not support transform without the webkit prefix.
+    thumbPositionElement.style.webkitTransform = `translateX(${thumbPosition}px)`;
```
jelbourn: Use the `applyTransform` function from core/style

src/components/slider/test-gesture-config.ts
```diff
@@ -0,0 +1,39 @@
+import {Injectable} from '@angular/core';
+import {MdGestureConfig} from '@angular2-material/core/gestures/MdGestureConfig';
+
+/**
+ * To test the dragging behavior on the slider, we need to be able to access the hammer instances
+ * to emit events for a drag.
+ */
```
jelbourn: Class description should be something like 
```
An extension of MdGestureConfig that exposes the unerlying HammerManager instances. 
Tests can use this instances to emit fake gesture events.
```

src/components/slider/test-gesture-config.ts
```diff
@@ -0,0 +1,39 @@
...
+      this.hammerInstances.set(element, [mc]);
+    } else {
+      this.hammerInstances.get(element).push(mc);
+    }
+
+    return mc;
+  }
+
+  /**
+   * Hammer creates a new instance for every listener so we need to apply our event on all instances
```
jelbourn: I think it's more "The Angular event plugin for Hammer creates a new HammerManager instance for each..."

src/components/slider/slider.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+    beforeEach(async(() => {
+      builder.createAsync(StandardSlider).then(f => {
+        fixture = f;
+        fixture.detectChanges();
+
+        testComponent = fixture.debugElement.componentInstance;
+
+        sliderDebugElement = fixture.debugElement.query(By.directive(MdSlider));
+        sliderNativeElement = sliderDebugElement.nativeElement;
+        sliderInstance = sliderDebugElement.injector.get(MdSlider);
```
jelbourn: You should be able to just say `sliderDebugElement.componentInstance` here. You only need to use `injector.get` when you're getting a `@Directive` instead of a `@Component`

src/components/slider/slider.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+    });
+
+    it('should update the track fill on click', () => {
+      let trackFillElement = sliderNativeElement.querySelector('.md-slider-track-fill');
+      let trackFillDimensions = trackFillElement.getBoundingClientRect();
+      let sliderDimensions =
+          sliderNativeElement.querySelector('.md-slider-track').getBoundingClientRect();
+
+      expect(trackFillDimensions.width).toBe(0);
+      dispatchClickEvent(sliderNativeElement, 0.5);
```
jelbourn: I noticed that you're using `0.5` quite a bit throughout the test; it would be a little bit better to use different values in different tests to make sure you don't accidentally preserve any state between tests and that `0.5` isn't a special case that works when something like `0.37` wouldn't.

src/components/slider/slider.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+
+      trackFillDimensions = trackFillElement.getBoundingClientRect();
+      expect(trackFillDimensions.width).toBe(sliderDimensions.width * 0.5);
+    });
+
+    it('should update the thumb position on click', () => {
+      let thumbElement = sliderNativeElement.querySelector('.md-slider-thumb-position');
+      let thumbDimensions = thumbElement.getBoundingClientRect();
+      let thumbWidth =
+          sliderNativeElement.querySelector('.md-slider-thumb').getBoundingClientRect().width;
```
jelbourn: You query for the track and the thumb pretty frequently; would it make sense to just get them once in the `beforeEach`?

src/components/slider/slider.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+      dispatchDragEvent(sliderNativeElement, 0, 0.5, gestureConfig);
+      expect(sliderInstance.value).toBe(50);
+    });
+  });
+});
+
+// The transition has to be removed in order to test the updated positions without setTimeout.
+@Component({
+  directives: [[MdSlider]],
+  template: `
```
jelbourn: optional: you could condense some of these string literals, e.g.
```ts
template: `<md-slider></md-slider>`,
styles: [`
  .md-slider-track-fill { 
    ... 
  }
`]

File: src/components/slider/slider.spec.ts [hunk 24]
```diff
@@ -0,0 +1,440 @@
...
+/**
+ * Dispatches a click event from an element.
+ * @param element The element from which the event will be dispatched.
+ * @param percentage The percentage of the slider where the click should occur. Used to find the
+ * physical location of the click.
+ */
+function dispatchClickEvent(element: HTMLElement, percentage: number): void {
+  let dimensions = element.getBoundingClientRect();
+  let width = dimensions.width;
+  let left = dimensions.left;
```
jelbourn: optional, could use destructuring:
```ts
let {width, left} = dimensions;
```

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+  Input,
+  ViewEncapsulation,
+  AfterContentInit,
+} from '@angular/core';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-slider',
+  host: {
+    'tabindex': '0',
```
hansl: @jelbourn Any reason we're bringing back `host` instead of adding them as `@HostBinding()` and `@HostListener`?

src/components/slider/slider.scss
robertmesserle: Maybe rename rather than just adding a comment. Would `$slider-height` be more accurate?

iveysaur: That's what I was thinking of doing, the only thing that makes me question that is when you have a vertical slider it would then be the width... but that's not in this PR so I'm not sure what's better.

hansl: Wouldn't a height/size in `em` or `rem` be more accurate?

src/components/slider/slider.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  @Input()
+  @HostBinding('attr.aria-valuenow')
+  get value() {
+    return this._value;
+  }
+
+  set value(v: number) {
+    this._value = Number(v);
+    this._isValueInitialized = true;
```
jelbourn: Rather than setting this here, you should set `_isInitialied = true` in `ngAfterViewInit` 
(pointed by by @DevVersion)

iveysaur:
@jelbourn: addressed comments.

src/components/slider/slider.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+    it('should remove the md-slider-active class on blur', () => {
+      let containerElement = sliderNativeElement.querySelector('.md-slider-container');
+
+      dispatchClickEvent(sliderNativeElement, 0.95);
+      fixture.detectChanges();
+
+      expect(containerElement.classList).toContain('md-slider-active');
+
+      // The test won't activate the onBlur function. Tried clicking/focusing the body and neither
+      // worked. Should update this if a way is ever found.
```
jelbourn: Comment can just be something like 
```ts
// Call the `onBlur` handler directly because we cannot simulate a focus event in unit tests.
```

src/components/slider/slider.ts
```diff
@@ -0,0 +1,247 @@
...
+  clamp(value: number, min = 0, max = 1) {
+    return Math.max(min, Math.min(value, max));
+  }
+}
+
+/**
+ * Renderer class in order to keep all dom manipulation in one place and outside of the main class.
+ */
+export class SliderRenderer {
+  constructor(private _elementRef: ElementRef) { }
```
jelbourn: You could store the `nativeElement` here so that you don't always have to go through the `elementRef` throughout the whole class

jelbourn:
LGTM other than last couple of comments. @kara should also do a pass

src/components/slider/slider.scss
```diff
@@ -0,0 +1,141 @@
+@import 'default-theme';
+
+// This refers to the thickness of the slider. On a horizontal slider this is the height, on a
+// vertical slider this is the width.
+$slider-thickness: 20px !default;
```
kara: Can we prefix these Sass variables? That's what we're doing elsewhere.  `$slider` -> `$md-slider`

src/components/slider/slider.scss
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Exists in order to pad the slider and keep everything positioned correctly.
+ * Cannot be merged with the .md-slider-container.
+ */
+.md-slider-wrapper {
+  width: 100%;
+  height: 100%;
+  padding-left: 8px;
+  padding-right: 8px;
```
kara: Can this be moved into a Sass variable?

src/components/slider/slider.scss
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/**
+ * Holds the isActive and isDragging classes as well as help with positioning the children.
```
kara: help -> helps

src/components/slider/slider.scss
```diff
@@ -0,0 +1,141 @@
...
+  position: absolute;
+  left: 0;
+  right: 0;
+  height: 100%;
+  background-color: $slider-off-color;
+}
+
+.md-slider-track-fill {
+  transition-duration: 400ms;
+  transition-timing-function: cubic-bezier(0.25, 0.8, 0.25, 1);
```
kara: Seems like you could use `$swift-ease-out-timing-function` from `_variables` here?

src/components/slider/slider.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+}
+
+/**
+ * Renderer class in order to keep all dom manipulation in one place and outside of the main class.
+ */
+export class SliderRenderer {
+  private _sliderElement: HTMLElement;
+
+  constructor(private _elementRef: ElementRef) {
```
kara: Is it necessary to save `_elementRef` on the class? Looks like you are only using it for `this._sliderElement`.

src/components/slider/test-gesture-config.ts
```diff
@@ -0,0 +1,39 @@
...
+   * Create a mapping of Hammer instances to element so that events can be emitted during testing.
+   */
+  buildHammer(element: HTMLElement) {
+    let mc = super.buildHammer(element);
+
+    if (!this.hammerInstances.get(element)) {
+      this.hammerInstances.set(element, [mc]);
+    } else {
+      this.hammerInstances.get(element).push(mc);
+    }
```
kara: Nitpick: any way you can flip this conditional?  Trying to scan "if not / else" is a bit harder than "if / else".

```ts
if (this.hammerInstances.get(element)) {
  this.hammerInstances.get(element).push(mc);  
} else {
  this.hammerInstances.set(element, [mc]);
}
```

src/demo-app/slider/slider-demo.ts
```diff
@@ -0,0 +1,10 @@
+import {Component} from '@angular/core';
+import {MdSlider} from '@angular2-material/slider/slider';
+
+@Component({
+  moduleId: module.id,
+  selector: 'slider-demo',
+  templateUrl: 'slider-demo.html',
+  directives: [MdSlider],
```
kara: Can we add a MD_SLIDER_DIRECTIVES array so we can put it here? So we don't forget.

src/components/slider/slider.ts
iveysaur: Nope not necessary, removed it. Thanks for catching that!

iveysaur:
@kara: addressed comments

jelbourn:
Still LGTM

PR #781, opened 2016-06-29
chore: remove all @internal to fix offline compile

jelbourn:
R: @robertmesserle @hansl @kara 
CC: @iveysaur @DevVersion 

When the code generated by the angular compiler interacts with the component, it does so through its class's public API. If anything is missing from the `.d.ts`, it will be an error when a downstream application tries to compile.

Also fixes some errors with tab's use of `ngIf`

### DIFF
--- a/src/components/button-toggle/button-toggle.ts
+++ b/src/components/button-toggle/button-toggle.ts
@@ -81,7 +81,7 @@ export class MdButtonToggleGroup implements ControlValueAccessor {
 
   /** Child button toggle buttons. */
   @ContentChildren(forwardRef(() => MdButtonToggle))
-  private _buttonToggles: QueryList<MdButtonToggle> = null;
+  _buttonToggles: QueryList<MdButtonToggle> = null;
 
   @Input()
   get name(): string {
@@ -219,11 +219,8 @@ export class MdButtonToggle implements OnInit {
   /** Whether or not this button toggle is checked. */
   private _checked: boolean = false;
 
-  /**
-   * Type of the button toggle. Either 'radio' or 'checkbox'.
-   * @internal
-   */
-  type: ToggleType;
+  /** Type of the button toggle. Either 'radio' or 'checkbox'. */
+  _type: ToggleType;
 
   /** The unique ID for this button toggle. */
   @HostBinding()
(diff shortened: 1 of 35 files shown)


hansl:
Did you just remove all of them or did you actrually only removed the ones that were harmful?

jelbourn:
To elaborate, with this PR, anything that is _not part of the public API_ is prefixed with an underscore. Some of those things may be actual `private`, and other are TypeScript "public" so that Angular itself can see them, but are not intended to be exposed to end developers.

jelbourn:
@hansl All of them, because there were very, very few that were actually not a problem (just a few in grid-list). I'm saying it's better to completely forgo using `@internal` because it's too easy to introduce an error with it.

jelbourn:
Also, anything marked with `TODO: internal` is something for us to eventually mark as "Looks public, but needs to be omitted from the API docs", such as `ControlValueAccessor` methods and lifecycle hooks.

src/components/button-toggle/button-toggle.ts
```diff
@@ -81,7 +81,7 @@ export class MdButtonToggleGroup implements ControlValueAccessor {
 
   /** Child button toggle buttons. */
   @ContentChildren(forwardRef(() => MdButtonToggle))
-  private _buttonToggles: QueryList<MdButtonToggle> = null;
+  _buttonToggles: QueryList<MdButtonToggle> = null;
```
robertmesserle: Why was this one changed?

jelbourn: `@ContentChildren` etc. also can't be private, since Angular itself needs to touch them

src/components/progress-circle/progress-circle.ts
```diff
@@ -53,30 +53,29 @@ export class MdProgressCircle implements OnDestroy {
    * Values for aria max and min are only defined as numbers when in a determinate mode.  We do this
    * because voiceover does not report the progress indicator as indeterminate if the aria min
    * and/or max value are number values.
-   *
-   * @internal
    */
   get ariaValueMin() {
```
robertmesserle: Did you miss this one?

src/components/progress-circle/progress-circle.ts
```diff
@@ -53,30 +53,29 @@ export class MdProgressCircle implements OnDestroy {
...
     return this.mode == 'determinate' ? 0 : null;
   }
 
-  /** @internal */
   get ariaValueMax() {
```
robertmesserle: And this one?

src/components/progress-circle/progress-circle.ts
robertmesserle: To clarify, the host binding up top is changed to use an underscore for this.

jelbourn: Yeah, I did. Good catch.

src/components/progress-circle/progress-circle.ts
jelbourn: Fixed.

hansl:
LGTM as well.

PR #792, opened 2016-06-30
docs(readme): add badges

jesselpalmer:
Adds dependencies and dev-dependencies badges similarly to [Angular CLI's README](https://github.com/angular/angular-cli/blob/master/README.md).


jelbourn:
I kind of feel like this is badge overload to the point where it's just adding too much noise to the page. Happy to hear a counter argument, though.

jesselpalmer:
I hear ya. I don't feel particularly strong about them. It can get a bit cluttered if you add too many.

PR #793, opened 2016-06-30
doc: Remove mode="side" since it breakes the backtick

drager:
Hello,

The `mode="side"` breaks the backtick of the start sidenav and therefore I removed it since it is not a working example, the example [here](https://github.com/angular/material2/tree/master/src/components/sidenav) works just fine though but that's because the `mode="side"` is not present in that example.


drager:
Or this may actually be a bug to take care of?

jelbourn:
@drager I don't understand the issue you're trying to fix.

drager:
@jelbourn: I thought a backdrop should always be present independently of the mode setting. But there should be no backdrop present when using `mode="side"`? If that is correct you can close this pull request.

jelbourn:
It is correct that `mode="side"` intentionally does not have a backdrop.

drager:
I see, great. Then I had misunderstood it before.

PR #808, opened 2016-07-03
chore(menu): add basic scaffold for menu

kara:
This simply adds the file scaffold for menu, so it's easier to read the implementation in the next PR.

r: @jelbourn


jelbourn:
Should also add `menu.spec.ts` while you're at it. Other than that LGTM.

Issue #791, opened 2016-06-30
Components should only fire a change event when they are focused


jelbourn:

DevVersion:
FYI: A better approach has been landed in `master`. We only emit a change event, when the native input emits one.

- This makes sure, that the given component completely matches the native component behavior.

Already fixed for: `md-slide-toggle`, `md-checkbox`

PR #820, opened 2016-07-06
fix(checkbox, slide-toggle): only emit change event if native input emitted one.

DevVersion:
* We should only emit a change event, when the native input element emitted one.<br/><br/>
  This makes sure that the given component matches the native browser behavior and doesn't emit a change event for example, when changing the native `checked` property.

---
FYI: This change is also required for the `radio` component, but this should be addressed in another PR, since this will require a bit more work, through the `radio` group.

References #791

### DIFF
--- a/src/components/checkbox/checkbox.spec.ts
+++ b/src/components/checkbox/checkbox.spec.ts
@@ -3,8 +3,7 @@ import {
     inject,
     async,
     fakeAsync,
-    flushMicrotasks,
-    tick
+    flushMicrotasks
 } from '@angular/core/testing';
 import {
     FORM_DIRECTIVES,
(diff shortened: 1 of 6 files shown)


jelbourn:
@DevVersion could you rebase?

DevVersion:
@jelbourn Done!

src/components/checkbox/checkbox.spec.ts
```diff
@@ -214,13 +213,47 @@ describe('MdCheckbox', () => {
       expect(testComponent.onCheckboxClick).toHaveBeenCalledTimes(1);
     });
 
-    it('should emit a change event when the `checked` value changes', async(() => {
+    it('should trigger the change event properly', async(() => {
```
jelbourn: "Property" isn't very specific. How about "should trigger a change event only when the native input does" ?

jelbourn:
LGTM besides the test name.


@robertmesserle I think this fix covers what you were doing in #739, any thoughts?

robertmesserle:
@jelbourn @DevVersion I think so.  Does this fix the nested checkbox demo?  I was using that as a test-case that I wanted to support.  We should make sure that this fixes that demo while we're at it.

https://github.com/angular/material2/issues/575 Has a good demo for what I was trying to fix.

src/components/checkbox/checkbox.spec.ts
DevVersion: Sounds good

DevVersion:
@robertmesserle 
> Does this fix the nested checkbox demo?

Yeah, one minor thing needs to be fixed, because the `(change)` is now emitting a `MdCheckboxChange` class, instead of just the boolean.

> #575 has a good demo for what I was trying to fix.

I tested that out, and it seems to work as well, when doing the adjustments for the `MdCheckboxChange` class.

robertmesserle:
@DevVersion @jelbourn :+1:

jelbourn:
LGTM, just needs a rebase.

DevVersion:
@jelbourn Uff, I accidentally had the Slider PR from testing in there. but I was able to fix the branch, so it's now rebased again :smile:

PR #828, opened 2016-07-07
fix(slide-toggle): fix runtime exception for incorrect mousedown binding

DevVersion:
* Commit fad4ef5 made all internal functions prefixed with an underscore.
  The commit did accidentally miss to add the underscore the `(mousedown)` host binding function, which
  causes a Runtime Exception now.

cc. @jelbourn


jelbourn:
@DevVersion thanks!

PR #843, opened 2016-07-11
chore: update to angular RC4 and other update other packages

jelbourn:
This removes the following deleted / deprecated APIs:
* Angular's exported versions of `describe`, `it`, `beforeEach`, `expect`, etc.
* `beforeEachProviders`, `provide(...)` and `new Provider(...)`
* `injectAsync`
* Remove jasmine matchers that came from `angular/angular`


hansl:
LGTM.

jelbourn:
CC @robertmesserle @DevVersion @iveysaur

package.json
```diff
@@ -28,23 +28,23 @@
...
-    "@angular/platform-browser": "2.0.0-rc.3",
-    "@angular/platform-browser-dynamic": "2.0.0-rc.3",
+    "@angular/common": "2.0.0-rc.4",
+    "@angular/compiler": "2.0.0-rc.4",
+    "@angular/core": "2.0.0-rc.4",
+    "@angular/http": "2.0.0-rc.4",
+    "@angular/platform-browser": "2.0.0-rc.4",
+    "@angular/platform-browser-dynamic": "2.0.0-rc.4",
     "@angular/router": "v3.0.0-alpha.8",
     "@angular/forms": "^0.1.0",
```
kara: Update forms to `0.2.0`?

kara:
We should probably update the package.jsons of form-related components to use `0.2.0`:
https://github.com/angular/material2/blob/master/src/components/checkbox/package.json

Otherwise looks good

package.json
jelbourn: Just tried this, and some tests start failing. We can do that update in a separate PR.

iveysaur:
LGTM, tests for gridlist are failing in this PR but I'm not sure if they are related to the change you made.

jelbourn:
Currently some grid-list tests are failing on iOS8. Seems like they were never running before and I accidentally fixed them. Debugging...

jonathan-casarrubias:
🎉

PR #846, opened 2016-07-12
fix(MdInput): Input should not be treated as empty if it is a date field. 

### LINKED ISSUE #845: bug(md-input): MdInput with type date cannot handle placeholders
Bug

Placeholders should always be small for date inputs since it defaults to: `mm/dd/yyyy`.

Placeholder is large for the default value for date inputs.

http://plnkr.co/edit/MURSTrNjlRj3AGzZkEfO?p=preview

* **Angular version:** 2.0.0-rc.4
* **Material version:** 2.0.0-alpha.6
* **Browser:** Chrome Version 51.0.2704.106 (64-bit)
* **Language:** TypeScript 2.0.0-dev
* **Built with**: Webpack

drager:
Fixes #845


src/components/input/input.html
```diff
@@ -35,7 +35,7 @@
 
       <label class="md-input-placeholder"
              [attr.for]="inputId"
-             [class.md-empty]="empty"
+             [class.md-empty]="empty && type !== 'date'"
```
DevVersion: This should be better handled inside of the "empty" getter.

drager: Alright, I decided to do that in the template since I saw you guys did that with some other things. But it's fixed now in the latest commit.

DevVersion: Any references? I guess in that cases there weren't any getters. If not the getter should probably not include that check.

drager: Well, these lines: https://github.com/drager/material2/blob/d8603948c36de9718e39284e58d12665a53ca569/src/components/input/input.html#L41-L42 as well as line 9.

DevVersion: At thx, yeah that's a difference, since this is just an inline check. And the given variable / getter is only returning the value and not the boolean. `empty` is actually a property, which returns a boolean, which means that the check with the `date` is part of the is `empty` check.

drager: Alright, fair enough. The code is updated and correct then.

jelbourn:
This is probably satisfactory to at least not be _bad_ for `type="date"`. It does need a unit test, though.

@hansl any thoughts?

drager:
I can probably get some unit tests done tomorrow.

hansl:
Do we have an issue filed describing why this is necessary?

Please add a unit test and I'll approve. Thanks!

jelbourn:
CI is failing because IE11 actually throws an error if you try to set `type="date"` on an `<input>`. In this case, I think it would be valid to skip this test for IE11. @hanslm do you have any opinions on this?

hansl:
Well look at that... http://caniuse.com/#search=date

Yeah it's fine. Unfortunate but IE11.

drager:
@hansl: Alright, is this something that I shoul do? If so, how do I disable tests for a specific browser?

jelbourn:
@drager We don't currently have a way to do it. For now I think you can use the `!(window.ActiveXObject) && "ActiveXObject" in window` trick in the test (though still in an `isInternetExplorer11` function) and we'll refactor it later.

jmcgoldrick:
This should probably apply to type=time too? I noticed it does that same thing.

drager:
@jelbourn, @hansl: I guess this is ready to get merged if you accept the changes. Regarding @jmcgoldrick's issue, do you guys think this PR also should include that? I can fix that if that is the case.

ross-nordstrom:
Any updates on this PR?

hansl:
LGTM. Rebase and we'll merge. Thanks!

hansl:
The failing tests are flaky, don't worry about them. thanks!

PR #848, opened 2016-07-13
chore: include .metadata.json files in packages for deployment

jelbourn:
R: @hansl 

Some of the things I've done with bash here are... questionable. Let me know if you have ideas on how to improve it. I'm going to do a publish under the `next` label so that Misko can install and test.


hansl:
LGTM.

PR #854, opened 2016-07-13
fix(): properly build demo-app in production.

DevVersion:
* Currently it was not possible to build the demo-app in production, due to the fact that the fileExtensions for the SystemJS Builder were not available.

> This was caused by the deprecated flag `defaultJSExtension`.

@jelbourn - This change is necessary to easily build a production build for the Material Bot.
 > Currently I'm just building normally and then running the SystemJS builder manually + modifying the entry point with jsdom.


jelbourn:
LGTM

PR #857, opened 2016-07-14
fix(progress-circle): remove performance.now to support non browser envs

kara:
This PR removes the reference to the performance global, as it doesn't work in environments without a window.

r: @jelbourn


jelbourn:
LGTM

Issue #789, opened 2016-06-30
md-input - lack of possibility to turn off autocomplete


maciejsobala:
#### Bug, feature request, or proposal:
putting autocomplete property to md-input

#### What is the expected behavior?
< md-input autocomplete="on/off"></mdinput> should be propagated to input element

#### What is the current behavior?
Lack of such mechanism

#### Which versions of Angular, Material, OS, browsers are affected?
Angular: 2.0.0-rc.3
Material: 2.0.0-alpha.5-2
Browsers: Safari, Chrome, FireFox
OS: OS X El Capitan, iOS 9.3

joshfriend:
#447 added support for the `autoComplete` and [is present in the current `2.0.0-alpha.6` release.](https://github.com/angular/material2/blob/2.0.0-alpha.6/src/components/input/input.ts)

`autocorrect`, `autocapitalize`, and `spellcheck` attributes are still missing though.

jelbourn:
It is currently `autoComplete` (camelCase). They should really match, though, so I opened #861

PR #858, opened 2016-07-14
Support autocapitalize and autocorrect attributes on md-input

joshfriend:


jelbourn:
LGTM

joshfriend:
@jelbourn per your comment https://github.com/angular/material2/issues/789#issuecomment-232823164 should I update this to be not use camel case?

kara:
@joshfriend It's better to make those changes all at once, so we'll keep the naming consistent with what we have for now.

PR #859, opened 2016-07-14
chore: add extract_metadata task to CI

jelbourn:
Also inlines resources for e2e tests which should speed them up (and hopefully make them less flaky); the build triggered by `ng serve` would have been clobbering the previous inlining.


hansl:
LGTM.

PR #865, opened 2016-07-15
Create Node script for stage-release

NCC1701M:
Created a stage-release script file for node. Hoping it can replace both stage-release.bat and stage-release.sh


jelbourn:
Closed since we're using gulp now.

PR #867, opened 2016-07-16
feat(menu): add menu trigger support

kara:
This PR adds basic menu toggling functionality (see [design doc here](https://docs.google.com/document/d/1cSUrY2WAqSN8ePldLLqrEVkk2BkYbPB2fysC6brvj4Q/edit#)).

For the below, clicking the button will automatically toggle the menu.

```html
<button [md-menu-trigger-for]="menu">More</button>
<md-menu #menu="mdMenu">
   <button md-menu-item>Item one</button>
</md-menu>
```

Still TODO (in future PRs):
- Animation for menu opening
- Position setting
- Keyboard events
- Prevent-close attribute

### DIFF
--- a/src/components/menu/menu-item.ts
+++ b/src/components/menu/menu-item.ts
@@ -1,4 +1,4 @@
-import {Directive, Input, HostBinding, HostListener} from '@angular/core';
+import {Directive, Input, HostBinding} from '@angular/core';
 
 /**
  * This directive is intended to be used inside an md-menu tag.
@@ -16,7 +16,10 @@ export class MdMenuItem {}
  */
 @Directive({
   selector: 'a[md-menu-item]',
-  host: {'role': 'menuitem'}
+  host: {
+    'role': 'menuitem',
+    '(click)': 'checkDisabled($event)'
+  }
 })
 export class MdMenuAnchor {
   _disabled: boolean;
@@ -33,7 +36,7 @@ export class MdMenuAnchor {
 
   @HostBinding('attr.aria-disabled')
   get isAriaDisabled(): string {
-    return this.disabled ? 'true' : 'false';
+    return String(this.disabled);
   }
 
   @HostBinding('tabIndex')
@@ -41,8 +44,7 @@ export class MdMenuAnchor {
     return this.disabled ? -1 : 0;
   }
 
-  @HostListener('click', ['$event'])
-  checkDisabled(event: any) {
+  checkDisabled(event: Event) {
     if (this.disabled) {
       event.preventDefault();
       event.stopPropagation();
--- a/src/components/menu/menu-item.ts
+++ b/src/components/menu/menu-item.ts
@@ -1,0 +1,51 @@
+import {Directive, Input, HostBinding, HostListener} from '@angular/core';
+
+/**
+ * This directive is intended to be used inside an md-menu tag.
+ * It exists mostly to set the role attribute.
+ */
+@Directive({
+  selector: 'button[md-menu-item]',
+  host: {'role': 'menuitem'}
+})
+export class MdMenuItem {}
+
+/**
+ * This directive is intended to be used inside an md-menu tag.
+ * It sets the role attribute and adds support for the disabled property to anchors.
+ */
+@Directive({
+  selector: 'a[md-menu-item]',
+  host: {'role': 'menuitem'}
+})
+export class MdMenuAnchor {
+  _disabled: boolean;
+
+  @HostBinding('attr.disabled')
+  @Input()
+  get disabled(): boolean {
+    return this._disabled;
+  }
+
+  set disabled(value: boolean) {
+    this._disabled = (value === false || value === undefined) ? null : true;
+  }
+
+  @HostBinding('attr.aria-disabled')
+  get isAriaDisabled(): string {
+    return this.disabled ? 'true' : 'false';
+  }
+
+  @HostBinding('tabIndex')
+  get tabIndex(): number {
+    return this.disabled ? -1 : 0;
+  }
+
+  @HostListener('click', ['$event'])
+  checkDisabled(event: any) {
+    if (this.disabled) {
+      event.preventDefault();
+      event.stopPropagation();
+    }
+  }
+}
--- a/src/components/menu/menu-trigger.ts
+++ b/src/components/menu/menu-trigger.ts
@@ -35,14 +35,14 @@ import {
 })
 export class MdMenuTrigger implements AfterViewInit, OnDestroy {
   private _portal: TemplatePortal;
-  private _overlay: OverlayRef;
+  private _overlayRef: OverlayRef;
   menuOpen: boolean = false;
 
   @Input('md-menu-trigger-for') menu: MdMenu;
   @Output() onMenuOpen = new EventEmitter();
   @Output() onMenuClose = new EventEmitter();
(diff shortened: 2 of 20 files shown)


src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,51 @@
...
+    return this._disabled;
+  }
+
+  set disabled(value: boolean) {
+    this._disabled = (value === false || value === undefined) ? null : true;
+  }
+
+  @HostBinding('attr.aria-disabled')
+  get isAriaDisabled(): string {
+    return this.disabled ? 'true' : 'false';
```
jelbourn: Just realized we could do this as `String(this.disabled)`, which is slightly shorter.

src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @HostBinding('tabIndex')
+  get tabIndex(): number {
+    return this.disabled ? -1 : 0;
+  }
+
+  @HostListener('click', ['$event'])
+  checkDisabled(event: any) {
```
jelbourn: Use the `host` property on the `@Directive` instead so we can have the real event type

src/components/menu/menu-trigger.ts
```diff
@@ -0,0 +1,109 @@
...
+export class MdMenuTrigger implements AfterViewInit, OnDestroy {
+  private _portal: TemplatePortal;
+  private _overlay: OverlayRef;
+  menuOpen: boolean = false;
+
+  @Input('md-menu-trigger-for') menu: MdMenu;
+  @Output() onMenuOpen = new EventEmitter();
+  @Output() onMenuClose = new EventEmitter();
+
+  constructor(private _overlayBuilder: Overlay, private _element: ElementRef,
```
jelbourn: I'd just call it `_overlay`

src/components/menu/menu-trigger.ts
```diff
@@ -0,0 +1,109 @@
...
+    return this._overlay.detach()
+        .then(() => this._setMenuState(false));
+  }
+
+  destroyMenu(): void {
+    this._overlay.dispose();
+  }
+
+  // set state rather than toggle to support triggers sharing a menu
+  private _setMenuState(bool: boolean): void {
```
jelbourn: `_setIsMenuOpen`

Also, `bool` ?

src/components/menu/menu-trigger.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.menuOpen ? this.onMenuOpen.emit(null) : this.onMenuClose.emit(null);
+  }
+
+  private _checkMenu() {
+    if (!this.menu || !(this.menu instanceof MdMenu)) {
+      throw new MdMenuMissingError();
+    }
+  }
+
+  private _createOverlay(): void {
```
jelbourn: Add function description for these last three functions (I think the rest are pretty obvious)

src/components/menu/menu-trigger.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  // set state rather than toggle to support triggers sharing a menu
+  private _setMenuState(bool: boolean): void {
+    this.menuOpen = bool;
+    this.menu._setClickCatcher(bool);
+    this.menuOpen ? this.onMenuOpen.emit(null) : this.onMenuClose.emit(null);
+  }
+
+  private _checkMenu() {
```
jelbourn: Add function description for these last four functions (I think the rest are pretty obvious)

src/components/menu/menu.ts
```diff
@@ -9,13 +20,23 @@ import {Component, Directive, ViewEncapsulation} from '@angular/core';
...
-  host: {'role': 'menuitem'}
-})
-export class MdMenuItem {}
+  @Output() close = new EventEmitter;
+  @ViewChild(TemplateRef) templateRef: TemplateRef<any>;
+
+  /**
+   * This function toggles the display of the menu's click catcher element.
+   * This element covers the viewport when the menu is open to detect clicks outside the menu.
+   * @internal
```
jelbourn: Don't use `@internal` any more, it messes up the offline compiler. Just add `TODO: internal`, and we'll eventually mark our docs appropriately.

src/core/style/_palette.scss
```diff
@@ -335,6 +335,7 @@ $md-light-theme-background: (
   hover:      rgba(black, 0.04), // TODO(kara): check style with Material Design UX
   card:       white,
   dialog:     white,
+  menu:       white,
```
jelbourn: I think a menu is technically a "card" for coloring purposes.

src/demo-app/menu/menu-demo.scss
```diff
@@ -0,0 +1,8 @@
+.demo-menu {
+  display: flex;
+
+  div {
```
jelbourn: `div` by itself is pretty broad (even nested)

src/components/menu/menu-trigger.ts
```diff
@@ -0,0 +1,109 @@
...
+    return this._overlay.attach(this._portal)
+      .then(() => this._setMenuState(true));
+  }
+
+  closeMenu(): Promise<void> {
+    return this._overlay.detach()
+        .then(() => this._setMenuState(false));
+  }
+
+  destroyMenu(): void {
```
jelbourn: Is this supposed to be part of the public api?

kara: Yeah, this was in the doc.  Do you think it's unnecessary?  I'm on the fence about it.

src/components/menu/menu-trigger.ts
kara: Fair :)   Changing

zigzag95:
Menu shoould be closed when clicking outside or resizing the window (body:click & window:resize)
A z-index strategy should be managed in overlay class (adding a menu on a dialog, nested menus, etc.)

src/components/menu/menu-trigger.ts
```diff
@@ -0,0 +1,126 @@
...
+  @Input('md-menu-trigger-for') menu: MdMenu;
+  @Output() onMenuOpen = new EventEmitter();
+  @Output() onMenuClose = new EventEmitter();
+
+  constructor(private _overlay: Overlay, private _element: ElementRef,
+              private _viewContainerRef: ViewContainerRef) {}
+
+  ngAfterViewInit() {
+    this._checkMenu();
+    this._createOverlay();
```
jelbourn: I just realized that it _might_ be a problem in Angular Universal if we try to create the overlay on init rather than once a user interaction occurs, since under the hood the overlay creates an element with browser APIs.

kara: Hmm, good point. I'll check it out.

src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,53 @@
...
+  selector: 'a[md-menu-item]',
+  host: {
+    'role': 'menuitem',
+    '(click)': 'checkDisabled($event)'
+  }
+})
+export class MdMenuAnchor {
+  _disabled: boolean;
+
+  @HostBinding('attr.disabled')
```
jelbourn: Also `attr.aria-disabled`

jelbourn:
LGTM

PR #874, opened 2016-07-17
fix(checkbox): MD-Checkbox Without Label has Different Vertical Align…

### LINKED ISSUE #813: MD-Checkbox Without Label has Different Vertical Alignment
#### Bug:

When I don't provide text between the `<md-checkbox></md-checkbox>` tags the vertical alignment is different then when text is provided `<md-checkbox> my-text </md-checkbox>`

I would expect that checkbox alignment does not depend on the presence of a label.

See this plunker: https://plnkr.co/edit/kbBb7a?p=preview

I want to use md-checkbox in a html `<table>`, therefore I don't need a label since the table header already takes care of that.

 Angular RC.x, Material Alpha.6, all browsers

trik:
…ment
closes #813

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -251,6 +251,9 @@ md-checkbox {
 // TODO(kara): Remove this style when fixing vertical baseline
 .md-checkbox-layout .md-checkbox-label {
   line-height: 24px;
+  :empty::after {
+    content: '\00a0';
+  }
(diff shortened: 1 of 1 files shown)


jelbourn:
R: @hansl

jelbourn:
Looks like this breaks tests on all browsers.

jelbourn:
Closed in favor of #1721

PR #888, opened 2016-07-20
feat(md-divider): initial divider implementation

Dona278:
Like in angular-material v1

### DIFF
--- a/src/components/divider/divider.scss
+++ b/src/components/divider/divider.scss
@@ -1,0 +1,15 @@
+@import 'variables';
+@import 'default-theme';
+
+md-divider {
+  display: block;  
+  border-top: 1px solid md-color($md-foreground, divider);
+  margin: 0;
+
+  &[md-inset] {
+    //16px list-item padding-left
+    //40px image width
+    //16px text padding-left
+    margin-left: 72px;
+  }
+}
--- a/src/components/divider/divider.ts
+++ b/src/components/divider/divider.ts
@@ -1,0 +1,26 @@
+import { 
+    Component,
+    Directive, 
+    ViewEncapsulation    
+} from '@angular/core';
+
+/** The inset directive, used to change divider behavior from "full-bleed" to "inset". */
+@Directive({    
+    selector: 'md-inset'    
+})
+export class MdInset {}
+
+/**
+ * Component that represents a horizontal line, along with styling it according to the Material Design.
+ */
+@Component({    
+    moduleId: module.id,
+    selector: 'md-divider',    
+    template: '',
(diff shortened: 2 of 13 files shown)


Dona278:
I signed it!

src/components/divider/divider.ts
```diff
@@ -0,0 +1,26 @@
...
+    Component,
+    Directive, 
+    ViewEncapsulation    
+} from '@angular/core';
+
+/** The inset directive, used to change divider behavior from "full-bleed" to "inset". */
+@Directive({    
+    selector: 'md-inset'    
+})
+export class MdInset {}
```
DevVersion: Why having an extra directive for that?

src/components/divider/divider.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Component that represents a horizontal line, along with styling it according to the Material Design.
+ */
+@Component({    
+    moduleId: module.id,
+    selector: 'md-divider',    
+    template: '',
+    styleUrls: ['divider.css'],
+    encapsulation: ViewEncapsulation.None
```
DevVersion: Indention should be with two spaces.

src/components/divider/divider.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+export class MdDivider {}
+
+//Export directives
```
DevVersion: Unnecessary comment IMO + there should be a space between the text and `//`, to pass the TSLint checks

src/components/divider/divider.scss
```diff
@@ -0,0 +1,15 @@
+@import 'variables';
+@import 'default-theme';
+
+md-divider {
```
DevVersion: Not sure about that PR in general, but the divider is already implemented in the list component (as it belongs to that)
- See https://github.com/angular/material2/blob/master/src/components/list/list.scss#L156

jelbourn:
@Dona278 We love your enthusiasm, but divider is not a component we plan on adding to the library. Be sure to check out our [contributing guidelines](https://github.com/angular/material2/blob/master/CONTRIBUTING.md#-want-a-feature) in the future so we can give you feedback before jumping into writing code.

Dona278:
Thank you @jelbourn and @DevVersion for reply. I wrote this "micro" component because i use it often with the first version of angular material (and it's a directive and not a simple css class).

In answer to @DevVersion in the material design spec you can see an image like email composer which are input text divided by divider.

In the future i check better your contributing guidelines, promised :)

Thanks again!

PR #893, opened 2016-07-21
feat(menu): add custom position support to menu

kara:
r: @jelbourn 

This PR adds custom position support to menu. The default menu opens "after" and "below" the trigger.  If you want to change that, you can modify the "x-position" and "y-position" attributes (see [design doc here](https://docs.google.com/document/d/1cSUrY2WAqSN8ePldLLqrEVkk2BkYbPB2fysC6brvj4Q/edit#)).

x-position: 'before' | 'after'
y-position: 'above' | 'below'

Examples: 

```html
<md-menu x-position="before">
...
</md-menu>
```

FYI, this PR needs to wait until my other menu PR goes in, so check out the last commit.

### DIFF
--- a/e2e/components/menu/menu-page.ts
+++ b/e2e/components/menu/menu-page.ts
@@ -14,11 +14,9 @@ export class MenuPage {
 
   body() { return element(by.tagName('body')); }
 
-  firstItem() { return element(by.id('one')); }
-
-  secondItem() { return element(by.id('two')); }
-
-  thirdItem() { return element(by.id('three')); }
+  items(index: number) {
+    return element.all(by.css('[md-menu-item]')).get(index);
+  }
 
   textArea() { return element(by.id('text')); }
 
--- a/e2e/components/menu/menu-page.ts
+++ b/e2e/components/menu/menu-page.ts
@@ -1,0 +1,59 @@
+import ElementFinder = protractor.ElementFinder;
+
+export class MenuPage {
+
+  constructor() {
+    browser.get('/menu');
+  }
+
+  menu() { return element(by.css('.md-menu')); }
+
+  trigger() { return element(by.id('trigger')); }
+
+  triggerTwo() { return element(by.id('trigger-two')); }
+
+  body() { return element(by.tagName('body')); }
+
+  firstItem() { return element(by.id('one')); }
+
+  secondItem() { return element(by.id('two')); }
+
+  thirdItem() { return element(by.id('three')); }
+
+  textArea() { return element(by.id('text')); }
+
+  beforeTrigger() { return element(by.id('before-t')); }
+
+  aboveTrigger() { return element(by.id('above-t')); }
+
+  combinedTrigger() { return element(by.id('combined-t')); }
+
+  beforeMenu() { return element(by.css('.md-menu.before')); }
+
+  aboveMenu() { return element(by.css('.md-menu.above')); }
+
+  combinedMenu() { return element(by.css('.md-menu.combined')); }
+
+  expectMenuPresent(expected: boolean) {
+    return browser.isElementPresent(by.css('.md-menu')).then((isPresent) => {
+      expect(isPresent).toBe(expected);
+    });
+  }
+
+  expectMenuLocation(el: ElementFinder, {x,y}: {x: number, y: number}) {
+    el.getLocation().then((loc) => {
+      expect(loc.x).toEqual(x);
+      expect(loc.y).toEqual(y);
+    });
+  }
+
+  expectMenuAlignedWith(el: ElementFinder, id: string) {
+    element(by.id(id)).getLocation().then((loc) => {
+      this.expectMenuLocation(el, {x: loc.x, y: loc.y});
+    });
+  }
+
+  getResultText() {
+    return this.textArea().getText();
+  }
+}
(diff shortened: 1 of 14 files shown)


src/components/menu/menu.ts
```diff
@@ -22,10 +24,21 @@ import {
 })
 export class MdMenu {
   private _showClickCatcher: boolean = false;
+  private _classList: Object;
+  positionX: 'before' | 'after' = 'after';
+  positionY: 'above' | 'below' = 'below';
```
jelbourn: Create a `type` for these, i.e.,
```ts
export type MenuPositionX: 'before' | 'after';
```

src/components/menu/menu.ts
```diff
@@ -35,6 +48,35 @@ export class MdMenu {
     this._showClickCatcher = bool;
   }
 
+  /**
+   * This method takes classes set on the host md-menu element and applies them on the
+   * menu template that displays in the overlay container.  Otherwise, it's difficult
+   * to style the containing menu from outside the component.
+   * @param classes: list of class names
+   */
+  private _mirrorHostClasses(classes: string): void {
```
jelbourn: `_copyHostClassesToMenuOverlay` ?

src/components/menu/menu.ts
```diff
@@ -35,6 +48,35 @@ export class MdMenu {
     this._showClickCatcher = bool;
   }
 
+  /**
+   * This method takes classes set on the host md-menu element and applies them on the
+   * menu template that displays in the overlay container.  Otherwise, it's difficult
+   * to style the containing menu from outside the component.
+   * @param classes: list of class names
```
jelbourn: Don't need a colon after the param name

e2e/components/menu/menu-page.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  triggerTwo() { return element(by.id('trigger-two')); }
+
+  body() { return element(by.tagName('body')); }
+
+  firstItem() { return element(by.id('one')); }
+
+  secondItem() { return element(by.id('two')); }
+
+  thirdItem() { return element(by.id('three')); }
```
jelbourn: By ID?

src/components/menu/menu.ts
```diff
@@ -22,10 +24,21 @@ import {
...
+  private _classList: Object;
+  positionX: 'before' | 'after' = 'after';
+  positionY: 'above' | 'below' = 'below';
 
   @Output() close = new EventEmitter;
   @ViewChild(TemplateRef) templateRef: TemplateRef<any>;
 
+  constructor(@Attribute('x-position') posX: 'before' | 'after',
+              @Attribute('y-position') posY: 'above' | 'below',
+              @Attribute('class') classes: string) {
```
jelbourn: This is only going to capture static classes; it won't get any of
* `[class]="expression"`
* `[class.magical]="isMagical"`
* `ngClass`

kara: Yeah, was deliberately trying to keep it static in case a class affects the dimensions of the menu, which would throw off the position calculation if it changes later.  But if we are switching creation of the overlay to user interaction (rather than onInit), it doesn't matter. I'll update!

e2e/components/menu/menu-page.ts
kara: Would you prefer element.all + index?  Can update.

src/components/menu/menu.ts
kara: Good call!  Way better.

src/components/menu/menu.ts
kara: Now that I'm changing this over to use @Input, it's a setter anyhow. Hopefully the comment will be sufficient?

kara:
@jelbourn comments should be addressed: 
https://github.com/angular/material2/pull/893/commits/08296313c629933c24ca1efbad11871209e41b6f

e2e/components/menu/menu.e2e.ts
```diff
@@ -0,0 +1,113 @@
+import { MenuPage } from './menu-page';
+
+describe('menu', function () {
+  let page: MenuPage;
+
+  beforeEach(function() {
+    page = new MenuPage();
+  });
+
+  it('should open menu when the trigger is clicked', function () {
```
jelbourn: `() =>`

kara: Oh, whoa.  Will update all these.

e2e/components/menu/menu-page.ts
jelbourn: I just don't see where the IDs are being set.

kara: They were being set in the menu-e2e.html. Removed the IDs in favor of this though.

src/components/menu/menu.ts
```diff
@@ -22,10 +26,34 @@ import {
 })
 export class MdMenu {
   private _showClickCatcher: boolean = false;
+  private _classList: Object;
```
jelbourn: `Object` isn't super clear; add a comment explaining that this is the input into `ngClass` with an example?

jelbourn:
LGTM aside from existing comments; did you figure out the scrollbars on SauceLabs thing?

PR #894, opened 2016-07-21
fix(overlay): lazily create container

kara:
This PR changes overlay containers to be created lazily. This allows them and components using them to be loaded in app shells or contexts outside the browser.


src/core/overlay/overlay-container.ts
```diff
@@ -1,12 +1,21 @@
...
- * Create the overlay container element, which is simply a div
- * with the 'md-overlay-container' class on the document body.
- */
-export function createOverlayContainer(): Element {
-  let container = document.createElement('div');
-  container.classList.add('md-overlay-container');
-  document.body.appendChild(container);
-  return container;
+export class OverlayContainer {
+  private _containerElement: Element;
```
jelbourn: Type should be `HTMLElement`

src/core/overlay/overlay-container.ts
```diff
@@ -1,12 +1,21 @@
...
+
+  getContainerElement(): Element {
```
jelbourn: Add function description that includes the fact that this method will lazily create the container.

src/core/overlay/overlay-container.ts
```diff
@@ -1,12 +1,21 @@
...
-/**
- * Create the overlay container element, which is simply a div
- * with the 'md-overlay-container' class on the document body.
- */
-export function createOverlayContainer(): Element {
-  let container = document.createElement('div');
-  container.classList.add('md-overlay-container');
-  document.body.appendChild(container);
-  return container;
+export class OverlayContainer {
```
jelbourn: Add class description

jelbourn:
LGTM aside from comment comments.

Issue #896, opened 2016-07-22
md-input padding does not affect underline


maciejsobala:
#### Bug, feature request, or proposal:

bug
#### What is the expected behavior?

Underline should be affected
#### What is the current behavior?

setting padding for md-input doesn't affect underline
#### What are the steps to reproduce?

http://plnkr.co/edit/Jl20dtBLH8hILAahubzu?p=preview
#### Which versions of Angular, Material, OS, browsers are affected?

"@angular2-material/core": "^2.0.0-alpha.6",
"@angular/core": "2.0.0-rc.4",
#### Is there anything else we should know?

Fix: 
change .md-input-underline position from absolute to relative

hansl:
Hi @maciejsobala, it seems like you're trying to add spacing around the input itself. You should use `margin` instead of `padding` in this case. Changing padding inside the input is not supported.

PR #897, opened 2016-07-22
bugfix #896

maciejsobala:
bugfix:
md-input padding does not affect underline #896

### DIFF
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -160,7 +160,7 @@ $md-input-underline-disabled-background-image: linear-gradient(to right,
   // The underline is what's shown under the input, its prefix and its suffix.
   // The ripple is the blue animation coming on top of it.
   .md-input-underline {
-    position: absolute;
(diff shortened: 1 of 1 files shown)


maciejsobala:
I signed it!

hansl:
Could you add a case like this to the demo so we can easily see it in different browsers? We use the demo extensively internally (at least until we have screenshot tests).

maciejsobala:
I don't got your point. You want me to put it to material2/src/demo-app/ for example?

hansl:
I want you to a seciton in the input demo with padding (and maybe large borders).

jelbourn:
@hansl is this still relevant?

hansl:
Hard to say.  I'll check the bug tomorrow.

hansl:
Hi @maciejsobala,

Changing the position of the underline will affect the `md-hint`, which now is misaligned. This PR is not the proper solution.

PR #901, opened 2016-07-22
fix(gestures): ensure drag and pan are recognized with slide

kara:


jelbourn:
LGTM aside from previously discussed comment

PR #902, opened 2016-07-22
fix(gestures): custom recognizers should not inherit twice.

DevVersion:
* HammerJS does not allow custom recognizers to use a default recognizer twice globally.
  This means that we can't create a `drag` and `slide` recognizer, which depends for all on one default recognizers.

As discussed with @kara this is another approach for fixing that gesture issues.


kara:
@DevVersion  A few notes:
1)  Swipe still appears to be broken. Can you fix?  
2)  Can you add slide to the gestures demo test page?
3)  Rebase now needed. Thanks!

DevVersion:
@kara Done!

kara:
Thanks for updating!  LGTM

PR #906, opened 2016-07-23
docs(menu): add initial docs for menu

kara:
This PR adds the docs for menu. Also slipped in a change that removes setting `menuOpen` directly, and changes it to a getter.


jelbourn:
LGTM

PR #908, opened 2016-07-23
fix(checkbox): wrong cursor when disabled

### LINKED ISSUE #907: [checkbox] Change cursor when disabled
I think it is confusing to users that the cursor is still a pointer when the checkbox is disabled. The pointer indicates that something will happen `onclick` but of course that's not true when a checkbox is disabled.

It would probably be better to use the default pointer in that case unless the current behavior is already consistent with native checkbox implementations.

crisbeto:
Fixes the checkbox using a pointer cursor, even when the element is disabled.

Fixes #907.

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -218,12 +218,13 @@ $_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1)
 }
 
 md-checkbox {
-  &, label {
-    cursor: pointer;
-  }
+  cursor: pointer;
 }
 
 .md-checkbox-layout {
+  // `cursor: inherit` ensures that the wrapper element gets the same cursor as the md-checkbox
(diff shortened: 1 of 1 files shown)


src/components/checkbox/checkbox.scss
```diff
@@ -218,8 +218,10 @@ $_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1)
 }
 
 md-checkbox {
-  &, label {
-    cursor: pointer;
+  cursor: pointer;
+
+  label {
+    cursor: inherit;
```
jelbourn: Can you un-nest the rule out from md-checkbox and just use md-checkbox-layout? This will keep the specificity down.

Add a comment explaining the intent of adding cursor: inherit.

crisbeto: Done. Hadn't noticed that the `label` uses `md-checkbox-layout`.

jelbourn:
LGTM

PR #909, opened 2016-07-23
fix(button): hover styles being applied to disabled buttons

### LINKED ISSUE #866: Hover style still shows on disabled buttons
To replicate: in demo app, hit "disable buttons" and hover over disabled buttons.    

Possible cause: https://github.com/angular/material2/commit/b24d321213d7a0e7c72c7c9de9e53b112a5a3a8d

crisbeto:
Fixes the button hover styles being applied, even if a button is disabled.

Fixes #866.

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -4,11 +4,17 @@
 @import 'button-base';
 
 // TODO(kara): Replace attribute selectors with class selectors when possible
-[md-button] {
+[md-button], [md-icon-button] {
   @extend %md-button-base;
 
+  // Only flat buttons and icon buttons (not raised or fabs) have a hover style.
+  &:hover {
+    // Use the same visual treatment for hover as for focus.
+    @include md-button-focus();
+  }
+
   &[disabled]:hover {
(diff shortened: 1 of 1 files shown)


src/components/button/button.scss
```diff
@@ -44,9 +44,11 @@
 }
 
 // Only flat buttons and icon buttons (not raised or fabs) have a hover style.
-[md-button]:hover, [md-icon-button]:hover {
-  // Use the same visual treatment for hover as for focus.
-  @include md-button-focus();
+[md-button], [md-icon-button] {
+  &:not([disabled]):hover {
```
jelbourn: I'm hesitant to get so specific with the rule. Can it be done such that the disabled style is defined after the hover style with the same specificity as to override it?

crisbeto: Done, although I think that this adds a little to the selector length, but reduces specificity. It seemed strange that the `:after` is used to add the background.

jelbourn:
@crisbeto LGTM if you can rebase

crisbeto:
Sorry @jelbourn, I was away last week. It's rebased now.

PR #911, opened 2016-07-23
fix(radio): radio-button should only emit change event if native input does.

### LINKED ISSUE #791: Components should only fire a change event when they are focused

DevVersion:
* The radio-button should only emit a change event, when the native input does.
  This ensures that the radio-button matches its behavior with the native radio buttons.

* The button-toggle should only emit a change event, when the native input does.
  This ensures that the button-toggle matches its behavior with the native visual hidden input.
  Notice that the `button-toggle` sets the input type differently based on the type of the toggle.

Breaking Changes: 
- radio-button will no longer emit change event on de-select<br/>
  (Matches native behavior)
- button-toggle will no longer emit change event on de-select when type is set to `radio`<br/>
  (Matches native behavior)

Fixes #791

### DIFF
--- a/src/components/button-toggle/button-toggle.html
+++ b/src/components/button-toggle/button-toggle.html
@@ -5,7 +5,8 @@
          [checked]="checked"
          [disabled]="disabled"
          [name]="name"
-         (change)="_onInputChange($event)">
+         (change)="_onInputChange($event)"
+         (click)="_onInputClick($event)">
 
   <div class="md-button-toggle-label-content">
     <ng-content></ng-content>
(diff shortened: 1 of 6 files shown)


jelbourn:
The changes look good, but we probably need to do the same thing to button-toggle. Do you want to do a pass on that as well?

DevVersion:
@jelbourn Done! Also added some tests for the change event on the `button-toggle`, because the behavior changes due different modes of the `button-toggle`.

jelbourn:
@DevVersion LGTM, just needs rebase

DevVersion:
@jelbourn Done.

PR #923, opened 2016-07-27
feat(tooltip): adds animations for tooltip

robertmesserle:
R: @jelbourn @kara 
CC: @josephperrott

### DIFF
--- a/src/components/tooltip/tooltip.scss
+++ b/src/components/tooltip/tooltip.scss
@@ -19,4 +19,27 @@ $md-tooltip-padding: 8px;
   margin: $md-tooltip-margin;
   height: $md-tooltip-height;
   line-height: $md-tooltip-height;
+  opacity: 0;
+  pointer-events: none;
+  transform: scale(0);
+}
+.md-tooltip-initialized {
+  transition: $swift-ease-out-duration $swift-ease-out-timing-function;
+  transition-property: opacity, transform;
+}
+.md-tooltip-below {
+  transform-origin: 50% 0;
+}
+.md-tooltip-above {
+  transform-origin: 50% 100%;
+}
+.md-tooltip-before {
+  transform-origin: 100% 50%;
+}
+.md-tooltip-after {
+  transform-origin: 0 50%;
+}
+.md-tooltip-visible {
+  opacity: 1;
+  transform: scale(1);
 }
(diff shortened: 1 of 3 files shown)


jelbourn:
Lots of CI failures.

src/components/tooltip/tooltip.scss
```diff
@@ -19,4 +19,27 @@ $md-tooltip-padding: 8px;
   margin: $md-tooltip-margin;
   height: $md-tooltip-height;
   line-height: $md-tooltip-height;
+  opacity: 0;
+  pointer-events: none;
+  transform: scale(0);
+  &.md-initialized {
```
jelbourn: Rather than creating these as nested classes, can they each be a standalone `md-tooltip-above`, etc.?

src/components/tooltip/tooltip.ts
```diff
@@ -43,6 +43,7 @@ export class MdTooltip {
   }
 
   private _overlayRef: OverlayRef;
+  private _tooltipRef: TooltipComponent;
```
jelbourn: I didn't think of this in the first PR, but you should rename `TooltipComponent` to `TooltipContainer`

(and then just call this member `_tooltipContainer`)

src/components/tooltip/tooltip.ts
```diff
@@ -142,6 +144,23 @@ export class MdTooltip {
    * Hides the tooltip and returns a promise that will resolve when the tooltip is hidden
    */
   hide(): Promise<any> {
+    if (this._tooltipRef) {
+      return new Promise(done => {
+        this._tooltipRef.remove().then(() => {
+          this.destroy().then(() => done());
+        });
+        this._tooltipRef = null;
+      });
```
jelbourn: I think this block could just be:
```ts
return this._tooltipRef.remove()
    .then(() => this.destroy())
  this._tooltipRef = null;  
```

Setting the tooltipRef to null doesn't need to be part of the chain, and you already have a promise coming back from both `remove` and `destroy`, so you shouldn't need a new one.

robertmesserle:
@jelbourn Addressed your comments and updated the unit tests to be passing.

jelbourn:
Closed in favor of #1644

PR #924, opened 2016-07-27
Add support for drawing tick marks on the slider

iveysaur:
Add tick interval to slider. A tick interval of `auto` will find a tick separation of at least 30px, otherwise the tick separation value is calculated by the number given. 

The tick interval is based off of the step, meaning a slider with a step of 3 and a tick interval of 4 will draw ticks every 4 steps/every 12 values.

A tick is always drawn at the end of a slider, with the second to last tick being removed if it is too close to the last one.

A repeating linear background is used to draw the ticks because I believe it is more performant/easier than creating multiple divs or using a canvas. This needs to be tested still.

R: @jelbourn @kara 
CC: @hansl @robertmesserle

### DIFF
--- a/src/components/slider/slider.spec.ts
+++ b/src/components/slider/slider.spec.ts
@@ -25,6 +25,8 @@ describe('MdSlider', () => {
         SliderWithMinAndMax,
         SliderWithValue,
         SliderWithStep,
+        SliderWithAutoTickInterval,
+        SliderWithSetTickInterval
       ],
     });
 
@@ -561,13 +563,11 @@ class SliderWithValue { }
 class SliderWithStep { }
 
 @Component({
-  directives: [MD_SLIDER_DIRECTIVES],
   template: `<md-slider step="5" tick-interval="auto"></md-slider>`
 })
 class SliderWithAutoTickInterval { }
 
 @Component({
-  directives: [MD_SLIDER_DIRECTIVES],
   template: `<md-slider step="3" tick-interval="6"></md-slider>`
 })
 class SliderWithSetTickInterval { }
--- a/src/components/slider/slider.spec.ts
+++ b/src/components/slider/slider.spec.ts
@@ -464,12 +464,12 @@ describe('MdSlider', () => {
       // toContain is used rather than toBe because FireFox adds 'transparent' to the beginning
       // of the background before the repeating linear gradient.
       expect(tickContainer.style.background).toContain('repeating-linear-gradient(to right, ' +
-      'rgb(0, 0, 0), rgb(0, 0, 0) 2px, transparent 2px, transparent 32.6px)');
+          'black, black 2px, transparent 2px, transparent 32.6px)');
     });
 
     it('should draw a tick mark on the end of the track', () => {
-      expect(lastTickContainer.style.background).toContain('linear-gradient(to left, rgb(0, 0, 0), '
-          + 'rgb(0, 0, 0) 2px, transparent 2px, transparent)');
+      expect(lastTickContainer.style.background).toContain('linear-gradient(to left, black, black' +
+          ' 2px, transparent 2px, transparent)');
     });
 
     it('should not draw the second to last tick when it is too close to the last tick', () => {
@@ -506,12 +506,12 @@ describe('MdSlider', () => {
       // which is at the position 20.16px and 1px is subtracted to center, giving a tick
       // separation of 19.16px.
       expect(tickContainer.style.background).toContain('repeating-linear-gradient(to right, ' +
-          'rgb(0, 0, 0), rgb(0, 0, 0) 2px, transparent 2px, transparent 19.16px)');
+          'black, black 2px, transparent 2px, transparent 19.16px)');
     });
 
     it('should draw a tick mark on the end of the track', () => {
-      expect(lastTickContainer.style.background).toContain('linear-gradient(to left, rgb(0, 0, 0), '
-          + 'rgb(0, 0, 0) 2px, transparent 2px, transparent)');
+      expect(lastTickContainer.style.background).toContain('linear-gradient(to left, black, '
+          + 'black 2px, transparent 2px, transparent)');
     });
   });
 });
--- a/src/components/slider/slider.spec.ts
+++ b/src/components/slider/slider.spec.ts
@@ -479,7 +479,7 @@ describe('MdSlider', () => {
     });
   });
 
-  fdescribe('slider with set tick interval', () => {
+  describe('slider with set tick interval', () => {
     let fixture: ComponentFixture<SliderWithSetTickInterval>;
     let sliderDebugElement: DebugElement;
     let sliderNativeElement: HTMLElement;
--- a/src/components/slider/slider.spec.ts
+++ b/src/components/slider/slider.spec.ts
@@ -461,13 +461,13 @@ describe('MdSlider', () => {
       // both sides. The value 30 will be located at the position 33.6px, and 1px is removed from
       // the tick mark location in order to center the tick. Therefore, the tick separation should
       // be 32.6px.
-      expect(tickContainer.style.background).toBe('repeating-linear-gradient(to right, ' +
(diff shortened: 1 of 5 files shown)


src/components/slider/slider.ts
```diff
@@ -96,6 +108,19 @@ export class MdSlider implements AfterContentInit {
     this._max = Number(v);
   }
 
+  @Input('tick-interval')
+  get tickInterval() {
+    return this._tickInterval;
+  }
+
+  set tickInterval(v: 'auto' | number) {
+    if (v == 'auto') {
```
hansl: You can safely ignore this `if` and straight do `this._tickInterval = v`

src/components/slider/slider.ts
```diff
@@ -215,6 +241,80 @@ export class MdSlider implements AfterContentInit {
...
 
   /**
+   * Calls the appropriate method for calculating the tick separation depending on what the tick
+   * interval is. If there is no tick interval or the interval is set to something other than a
+   * number or 'auto', nothing happens.
+   */
+  private _calculateTickSeparation() {
+    if (this.tickInterval == 'auto') {
+      this._calculateAutoTickSeparation();
+    } else if (typeof this.tickInterval == 'number') {
```
hansl: No need for an `if` here, because typescript understand union of types.

iveysaur: This is needed because this function is called regardless of if `tick-interval` was set when creating a slider and ticks should only be drawn when `tick-interval` is explicitly set.

src/components/slider/slider.ts
```diff
@@ -215,6 +228,80 @@ export class MdSlider implements AfterContentInit {
   }
 
   /**
+   * Calls the appropriate method for calculating the tick separation depending on what the tick
+   * interval is. If there is no tick interval or the interval is set to something other than a
+   * number or 'auto', nothing happens.
+   */
+  private _calculateTickSeparation() {
```
jelbourn: Perhaps call this `_updateTickSeparation`? When I see `calculate`, I think it should return something.

src/components/slider/slider.ts
```diff
@@ -215,6 +228,80 @@ export class MdSlider implements AfterContentInit {
...
+    if (this._tickInterval == 'auto') {
+      this._calculateAutoTickSeparation();
+    } else if (Number(this._tickInterval)) {
+      this._calculateTickSeparationFromInterval();
+    }
```
jelbourn: optional: you could do
```ts
this._tickInterval == 'auto' ?
    this._calculateAutoTickSeparation() :
    this._calculateTickSeparationFromInterval();
```

jelbourn: Actually nevermind, there's a second condition.

src/components/slider/slider.ts
```diff
@@ -215,6 +228,80 @@ export class MdSlider implements AfterContentInit {
   }
 
   /**
+   * Calls the appropriate method for calculating the tick separation depending on what the tick
+   * interval is. If there is no tick interval or the interval is set to something other than a
+   * number or 'auto', nothing happens.
+   */
```
jelbourn: This function description should just be something like "Calculates the separation in pixels of tick marks...". The fact that it delegates to another function is not something you need to mention in the description.

src/components/slider/slider.ts
```diff
@@ -215,6 +228,80 @@ export class MdSlider implements AfterContentInit {
...
+   * needs a tick and eliminating the number of ticks until there is a distance of at least 30px
+   * between each tick.
+   */
+  private _calculateAutoTickSeparation() {
+    // The pixel value for how far apart the ticks should be.
+    let tickSeparation = 0;
+    // Keeps track of how many steps to multiply the slider's step by.
+    let stepCounter = 1;
+
+    while (tickSeparation < MIN_AUTO_TICK_SEPARATION) {
```
jelbourn: So, you don't actually need this loop; you can calculate the appropriate multiple of `step`. This would look something like
```ts
stepCounter = Math.ceil(MIN_AUTO_TICK_SEPARATION * (max - min) / width * step)
```
(though obviously it's not a "counter" any more)

[working it out](https://goo.gl/photos/qyjjBkR5QG8YG9P79)

src/components/slider/slider.ts
```diff
@@ -215,6 +228,80 @@ export class MdSlider implements AfterContentInit {
...
+    // The percentage of the step on the slider is needed in order to calculate the pixel offset
+    // from the beginning of the slider. This offset is the tick separation.
+    let tickPercentage = this.calculatePercentage(tickValue);
+    this._renderer.drawTicks(this._sliderDimensions.width * tickPercentage);
+  }
+
+  /**
+   * Calculates the percentage of the slider that a value is.
+   * @param value The value a percentage is needed for.
+   * @returns {number} The percentage.
```
jelbourn: Don't need types in the function description

src/components/slider/slider.ts
```diff
@@ -264,6 +351,30 @@ export class SliderRenderer {
...
   }
+
+  /**
+   * Draws ticks onto the tick container.
+   * @param tickSeparation How far apart to draw the ticks.
+   */
+  drawTicks(tickSeparation: number) {
+    let tickContainer = <HTMLElement>this._sliderElement.querySelector('.md-slider-tick-container');
+    let tickContainerWidth = tickContainer.getBoundingClientRect().width;
+    let lastTickContainer =
```
jelbourn: Add comment explaining why you need an extra element for the last tick.

src/components/slider/slider.ts
```diff
@@ -264,6 +351,30 @@ export class SliderRenderer {
...
+        <HTMLElement>this._sliderElement.querySelector('.md-slider-last-tick-container');
+    // Subtract 1 from the tick separation to center the tick.
+    // TODO: Perf test this.
```
jelbourn: `TODO: Evaluate the rendering performance of using repeating background gradients`

src/components/slider/slider.ts
```diff
@@ -264,6 +351,30 @@ export class SliderRenderer {
...
+    tickContainer.style.background = `repeating-linear-gradient(to right, #000000, #000000 2px,
```
jelbourn: `#000000` -> `black`

This is okay for this PR, but we also need to consider whether this should be a different color for dark themes.

src/components/slider/slider.spec.ts
```diff
@@ -419,6 +419,84 @@ describe('MdSlider', () => {
...
+      });
+    }));
+
+    it('should set the correct tick separation', () => {
+      // The first tick mark is going to be at value 30 as it is the first step after 30px. The
+      // width of the slider is 112px because the minimum width is 128px with padding of 8px on
+      // both sides. The value 30 will be located at the position 33.6px, and 1px is removed from
+      // the tick mark location in order to center the tick. Therefore, the tick separation should
+      // be 32.6px.
+      expect(tickContainer.style.background).toContain('repeating-linear-gradient(to right, ' +
```
jelbourn: Also explain why you use `toContain` rather than `toBe`

src/components/slider/slider.ts
iveysaur: Thanks for the picture that worked it out! Happy to remove that loop 😄

iveysaur:
@jelbourn: Addressed comments.

src/components/slider/slider.ts
```diff
@@ -215,6 +228,68 @@ export class MdSlider implements AfterContentInit {
...
+  /**
+   * Calculates the optimal separation in pixels of tick marks based on the minimum auto tick
+   * separation constant.
+   */
+  private _updateAutoTickSeparation() {
+    let width = this._sliderDimensions.width;
+
+    // This calculates which step is far enough away from the beginning of the slider to draw a tick
+    // at. This value will then be used to draw ticks at every tickStep steps.
+    let tickStep =
```
jelbourn: I think this would benefit from some more explanation as to where the calculation came from.

jelbourn:
LGTM

PR #934, opened 2016-07-29
feat(tabs): adds support for `disabled` tabs

### LINKED ISSUE #645: [feature] Add disabled property to md-tab
A common UX situation requires us to disable certain areas of the view (buttons, dropdowns etc...).  My use case is that certain tabs should not be accessible to the user based on a model value.  A clean solution would be to have a `disabled` property on the tab, to be used like:
```
<md-tab-group>
  <md-tab [disabled]="isDisabled">
    <template md-tab-label>One</template>
    <template md-tab-content>
      <h1>Some tab content</h1>
      <p>...</p>
    </template>
  </md-tab>
  <md-tab>
    <template md-tab-label>Two</template>
    <template md-tab-content>
      <h1>Some more tab content</h1>
      <p>...</p>
    </template>
  </md-tab>
</md-tab-group>
```
The best alternative I could think about is very clumsy: to watch the `selectedIndex` property with a `setInterval` or `click` handler, and if that becomes the index of a tab that should not be accessible, to reset it to the previous index value.

If the tab group emitted a `selected` event with the index of the newly selected tab, I could bind to that instead but even that wouldn't be as clean as a `disabled` property that could be data-bound.

PS: I realize that `*ngIf` could be used to remove a tab from the DOM but this doesn't serve the same purpose: sometimes you need the user to be aware of the UI element that she cannot activate (just as you would disable rather than hide a form submit button)

### LINKED ISSUE #880: MD-TAB does not support [disabled] property
MD-TAB does not support [disabled] property
https://github.com/angular/material2/tree/master/src/components/tabs

for the code
```html
<md-tab-group>
    <md-tab [disabled]="true"></md-tab>
    <md-tab></md-tab>
</md-tab-group>
```
the first tab should be disabled
both tabs are enabled, operational

create a code like this:
```html
<md-tab-group>
    <md-tab [disabled]="true"></md-tab>
    <md-tab></md-tab>
</md-tab-group>
```

in Angular Material 1 it is functional:
https://material.angularjs.org/latest/demo/tabs

robertmesserle:
closes #645 

R: @jelbourn

### DIFF
--- a/src/components/tabs/tab-group.scss
+++ b/src/components/tabs/tab-group.scss
@@ -40,6 +40,11 @@ $md-tab-bar-height: 48px !default;
   }
 }
 
+.md-tab-disabled {
+  cursor: default;
+  pointer-events: none;
+}
+
 /** The bottom section of the view; contains the tab bodies */
 .md-tab-body-wrapper {
   position: relative;
--- a/src/components/tabs/tabs.ts
+++ b/src/components/tabs/tabs.ts
@@ -43,6 +43,16 @@ export class MdTabChangeEvent {
 export class MdTab {
   @ContentChild(MdTabLabel) label: MdTabLabel;
   @ContentChild(MdTabContent) content: MdTabContent;
+
+  // TODO: Replace this when BooleanFieldValue is removed.
+  private _disabled = false;
+  @Input('disabled')
+  set disabled(value: boolean) {
+    this._disabled = (value != null && `${value}` !== 'false');
+  }
+  get disabled(): boolean {
+    return this._disabled;
+  }
 }
 
 /**
@@ -67,7 +77,7 @@ export class MdTabGroup {
   private _selectedIndex: number = 0;
   @Input()
   set selectedIndex(value: number) {
-    if (value != this._selectedIndex) {
+    if (value != this._selectedIndex && this.isValidIndex(value)) {
       this._selectedIndex = value;
 
       if (this._isInitialized) {
@@ -77,6 +87,19 @@ export class MdTabGroup {
   }
   get selectedIndex(): number {
     return this._selectedIndex;
+  }
+
+  /**
+   * Determines if an index is valid.  If the tabs are not ready yet, we assume that the user is
+   * providing a valid index and return true.
+   */
+  isValidIndex(index: number): boolean {
+    if (this._tabs) {
+      const tab = this._tabs.toArray()[index];
+      return tab && !tab.disabled;
+    } else {
+      return true;
+    }
   }
 
   /** Output to enable support for two-way binding on `selectedIndex`. */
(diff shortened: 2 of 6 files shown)


src/components/tabs/tab-group.scss
```diff
@@ -38,6 +38,10 @@ $md-tab-bar-height: 48px !default;
     opacity: 1;
     background-color: md-color($md-primary, 100, 0.3);
   }
+  &.md-disabled {
```
jelbourn: Un-nest and make this something like `md-tab-disabled`?

src/components/tabs/tabs.ts
```diff
@@ -42,6 +42,15 @@ export class MdTabChangeEvent {
 export class MdTab {
   @ContentChild(MdTabLabel) label: MdTabLabel;
   @ContentChild(MdTabContent) content: MdTabContent;
+
+  private _disabled = false;
+  @Input('disabled')
+  set disabled(value: boolean) {
+    this._disabled = (value != null && `${value}` !== 'false');
```
jelbourn: Add TODO
```
Replace this when BooleanFieldValue is removed.
```

src/components/tabs/tabs.ts
```diff
@@ -79,6 +88,19 @@ export class MdTabGroup {
...
+  /**
+   * Determines if an index is valid.  If the tabs are not ready yet, we assume that the user is
+   * providing a valid index and return true.
+   */
+  isValidIndex(index: number): boolean {
+    if (this._tabs) {
+      const tab = this._tabs.toArray()[index];
+      return tab && !tab.disabled;
+    } else {
+      return true;
```
jelbourn: If you don't have any tabs, why is it `true` instead of `false`? Wouldn't any index be "invalid", then?

src/components/tabs/tabs.ts
```diff
@@ -181,17 +205,28 @@ export class MdTabGroup {
...
   focusPreviousTab(): void {
-    if (this.focusIndex > 0) {
-      this.focusIndex--;
+    if (this._labelWrappers) {
+      for (let i = this.focusIndex - 1; i >= 0; i--) {
+        if (this.isValidIndex(i)) {
+          this.focusIndex = i;
+          return;
+        }
+      }
```
jelbourn: Lots of duplicated code between `focusPreviousTab` and `focusNextTab`. You should do something like
```ts
focusNextTab() {
  moveFocus(1);
}

focusPreviousTab() {
  moveFocus(-1);
}

moveFocus() {
  // ...
}
```

jelbourn: Kara has something like this in her most recent menu PR

src/components/tabs/tabs.ts
robertmesserle: Ah, I should add a comment explaining why.  When the component is first loaded, the tabs aren't ready yet, so we allow any value to pass through.  In the case of async tabs, we don't know when, exactly, the tabs will be ready, so I handle it in a generic way that does nothing while the component is in an invalid state (ie. no tabs).

robertmesserle: Actually, it is kind of explained in the method comment

robertmesserle:
@jelbourn Addressed comments, rebased.

jelbourn:
LGTM

PR #943, opened 2016-08-02
fix(overlay): fixes blurry overlay

### LINKED ISSUE #932: bug(overlay): overlays are sometimes blurred in Chrome
I think you should change the text to "Use your eye glasses to see tooltips"

image

PS: My chrome zoom is set to 100%

robertmesserle:
- rounds pixel values used in `connected-position-strategy`
- replaces percent-based offset with calculated pixel values in
  `global-position-strategy`

closes #932 

R: @jelbourn

### DIFF
--- a/src/components/tooltip/tooltip.scss
+++ b/src/components/tooltip/tooltip.scss
@@ -15,7 +15,7 @@ $md-tooltip-padding: 8px;
   padding: 0 $md-tooltip-padding;
   border-radius: 2px;
   font-family: $md-font-family;
-  font-size: 10px;
+  font-size: 11px;
   margin: $md-tooltip-margin;
   height: $md-tooltip-height;
   line-height: $md-tooltip-height;
--- a/src/core/overlay/position/connected-position-strategy.ts
+++ b/src/core/overlay/position/connected-position-strategy.ts
@@ -191,8 +191,8 @@ export class ConnectedPositionStrategy implements PositionStrategy {
   private _setElementPosition(element: HTMLElement, overlayPoint: Point) {
     let scrollPos = this._viewportRuler.getViewportScrollPosition();
 
-    let x = overlayPoint.x + scrollPos.left;
-    let y = overlayPoint.y + scrollPos.top;
+    let x = Math.round(overlayPoint.x + scrollPos.left);
+    let y = Math.round(overlayPoint.y + scrollPos.top);
 
     // TODO(jelbourn): we don't want to always overwrite the transform property here,
     // because it will need to be used for animations.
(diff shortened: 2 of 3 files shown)


src/components/tooltip/tooltip.scss
```diff
@@ -15,7 +15,7 @@ $md-tooltip-padding: 8px;
   padding: 0 $md-tooltip-padding;
   border-radius: 2px;
   font-family: $md-font-family;
-  font-size: 10px;
+  font-size: 11px;
```
jelbourn: Add comment explaining why we're going off-spec.

src/core/overlay/position/connected-position-strategy.ts
```diff
@@ -191,8 +191,8 @@ export class ConnectedPositionStrategy implements PositionStrategy {
   private _setElementPosition(element: HTMLElement, overlayPoint: Point) {
     let scrollPos = this._viewportRuler.getViewportScrollPosition();
 
-    let x = overlayPoint.x + scrollPos.left;
-    let y = overlayPoint.y + scrollPos.top;
+    let x = Math.round(overlayPoint.x + scrollPos.left);
+    let y = Math.round(overlayPoint.y + scrollPos.top);
```
jelbourn: Add comment explaining why it rounds.

src/core/overlay/position/global-position-strategy.ts
```diff
@@ -96,12 +84,16 @@ export class GlobalPositionStrategy implements PositionStrategy {
...
-    // because it will need to be used for animations.
-    let tranlateX = this._reduceTranslateValues('translateX', this._translateX);
-    let translateY = this._reduceTranslateValues('translateY', this._translateY);
+    // Wait one frame for the content to be updated so that we can calculate the correct offset
+    requestAnimationFrame(() => {
+      // TODO(jelbourn): we don't want to always overwrite the transform property here,
+      // because it will need to be used for animations.
+
+      let translateX = Math.round(element.offsetWidth * -0.5);
+      let translateY = Math.round(element.offsetHeight * -0.5);
```
jelbourn: It looks like you're _always_ centering the element now? It should only do that when it's configured to be centered.

jelbourn:
Need unit tests for this change.

jelbourn:
@robertmesserle looks like legit test failures

robertmesserle:
Work still needed:

- [ ] The unit tests will need to be refactored using `async` rather than `fakeAsync` due to the addition of `requestAnimationFrame`
- [ ] E2E tests are failing on Travis only - I have not been able to recreate this error locally in order to debug.
- [ ] We should make the percent conversion a more generic process that supports any values rather than a hard-coded check for `-50%`.

philipooo:
I copied these changes to alpha.9 and noticed that the dialog gets not repositioned to the middle of the screen when the content changes (when open). Thanks for the change btw!

jelbourn:
Replaced by #1962

PR #948, opened 2016-08-05
build: move from angular-cli to gulp

hansl:


RemKolomna:
Please, share the reasons behind moving to gulp build. Are you willing to stay with Systemjs or cli is very unstable now?

hansl:
@RemKolomna CLI is moving to WebPack, and as much as it pains me, it's not suitable for libraries (yet). We'll come back to using it when it does.

snolflake:
@hansl glad to know they're going to use webpack

Mind to share with us why you believe webpack isn't suitable for libraries? Just curious...

hansl:
I don't think this PR is the right place for this discussion so I'll be short; Angular-CLI is not suitable for libraries (yet). Webpack is suitable for apps, not libraries, because it's not meant to output files that will be published to npm. It's meant to bundle/pack files that will be served. I might move our demo-app serving and building to webpack though as it's faster than what we have here, but for now this fills my current set of goals to not be dependent on the CLI.

gulpfile.js
```diff
@@ -0,0 +1,371 @@
...
+  '@angular/compiler-cli', 'ngc', ['-p', path.relative(__dirname, componentsDir)]
+));
+
+
+/***************************************************************************************************
+ * Vendor files Build Tasks.
+ */
+gulp.task(':build:vendor', function() {
+  const npmVendorFiles = [
+    'core-js/client', 'zone.js/dist', 'hammerjs', 'systemjs/dist', 'rxjs', '@angular', 'hammerjs'
```
danielgek: @hansl you have double hammerjs  in vendor files

hansl: It's twice as important ^_^

danielgek: ahah, can you explain why ?

hansl: too much copy-pasta from angular-cli.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+const e2eAppDir = path.join(srcDir, 'e2e-app');
+
+const outDir = 'dist';
+const outLibDir = path.join(outDir, '@angular2-material');
+
+
+/**
+ * Create a TS Build Task, based on the options.
+ */
+function createTsBuildTask(options) {
```
jelbourn: Can we move these `createXxx` functions into a separate file?

jelbourn: I think these would each be better without `create` in the name, then the individual rules would look like:
```js
gulp.task(':build:components:ts', tsBuildTask({ tsConfigPath: componentsDir }));
```

jelbourn: Is `options` of any particular type? I generally try to avoid JS functions that just accept some `options` object without any additional information about what the options are. Makes me wonder if we could have this as a `gulpfile.ts` instead of `.js`.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+  const tsConfigDir = options.tsConfigPath;
+  const tsConfigPath = path.join(tsConfigDir, 'tsconfig.json');
+  const tsConfig = JSON.parse(fs.readFileSync(tsConfigPath, 'utf-8'));
+  const dest = path.join(tsConfigDir, tsConfig.compilerOptions.outDir);
+
+  return function() {
```
jelbourn: `() =>` ?
(here and elsewhere)

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+const outLibDir = path.join(outDir, '@angular2-material');
+
+
+/**
+ * Create a TS Build Task, based on the options.
+ */
+function createTsBuildTask(options) {
+  const tsConfigDir = options.tsConfigPath;
+  const tsConfigPath = path.join(tsConfigDir, 'tsconfig.json');
+  const tsConfig = JSON.parse(fs.readFileSync(tsConfigPath, 'utf-8'));
```
jelbourn: Can we ever get into a situation where the same `tsconfig` would be parsed multiple times from one task?

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+const componentsDir = path.join(srcDir, 'lib');
+const devAppDir = path.join(srcDir, 'demo-app');
+const e2eAppDir = path.join(srcDir, 'e2e-app');
+
+const outDir = 'dist';
+const outLibDir = path.join(outDir, '@angular2-material');
+
+
+/**
+ * Create a TS Build Task, based on the options.
```
jelbourn: Expand this comment to explain what this function does beyond `gulp-typescript` by itself.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+  const tsConfig = JSON.parse(fs.readFileSync(tsConfigPath, 'utf-8'));
+  const dest = path.join(tsConfigDir, tsConfig.compilerOptions.outDir);
+
+  return function() {
+    const tsProject = gulpTs.createProject(tsConfigPath, {
+      typescript: require('typescript')
+    });
+
+    let pipe = tsProject.src()
+      .pipe(gulpSourcemaps.init())
```
jelbourn: What's going on with the source maps here that `tsc` doesn't do on its own? Comment?

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+/**
+ * Create a Gulp task that executes a process.
+ */
+function createExecTask(packageName, executable, args) {
+  if (!args) {
+    args = executable;
+    executable = undefined;
+  }
+  return function(done) {
+    resolveBin(packageName, { executable: executable }, function (err, binPath) {
```
jelbourn: Should be able to just have `{executable}`

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+      if (err) {
+        console.error(err);
+        throw err;
+      }
+
+      const spawn = child_process.spawn(binPath, args);
+
+      spawn.stdout.on('data', (data) => {
+        process.stdout.write(data);
+      });
```
jelbourn: nit: could be a single line
```js
spawn.stdout.on('data', data => process.stdout.write(data));
spawn.stderr.on('data', data => process.stderr.write(data));
```

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+    executable = undefined;
+  }
+  return function(done) {
+    resolveBin(packageName, { executable: executable }, function (err, binPath) {
+      if (err) {
+        console.error(err);
+        throw err;
+      }
+
+      const spawn = child_process.spawn(binPath, args);
```
jelbourn: `spawn` > `childProcess` ?

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+        }
+        done();
+      });
+    });
+  }
+}
+
+/***************************************************************************************************
+ * Components Build Tasks.
+ */
```
jelbourn: I'm not a huge fan of the full-bleed comment section markers; I generally think that two newlines and a regular comment is enough. E.g.,

```js
/** Component build tasks. */

...


/** Vendor file build tasks */

...
```

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+gulp.task(':build:components:ts', createTsBuildTask({ tsConfigPath: componentsDir }));
```
jelbourn: I don't quite follow your task naming convention. I'm guessing the leading colon means it's "private" (only meant to be consumed as the dep of another rule)? 

What do you think of using `_` instead of a leading colon and also using dots instead of colons for semantic delimitiers, e.g. `_build.components.assets`. I find the dots to be easier to read and to type.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+gulp.task(':build:components:assets', function() {
+  return gulp.src(path.join(componentsDir, '*/**/*.!(ts|spec.ts)'))
```
jelbourn: This will also capture `README` and `package.json` files. Is that intentional?

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+    .pipe(gulp.dest(outLibDir));
+});
+gulp.task(':build:components:scss', function() {
```
jelbourn: Is there overlap between this task and `:build:components:assets`?

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+));
+
+
+/***************************************************************************************************
+ * Vendor files Build Tasks.
+ */
+gulp.task(':build:vendor', function() {
+  const npmVendorFiles = [
+    '@angular', 'core-js/client', 'hammerjs', 'rxjs', 'systemjs/dist', 'zone.js/dist'
+  ];
```
jelbourn: This `const` could be defined outside the task.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+
+
+/***************************************************************************************************
+ * DevApp Build Tasks.
+ */
+gulp.task(':build:devapp:ts', [':build:components:ts'], createTsBuildTask({ tsConfigPath: devAppDir }));
+gulp.task(':build:devapp:scss', [':build:components:scss'], createSassBuildTask({
+  dest: outDir,
+  root: devAppDir,
+  // Change this once we have a better strategy for releasing SCSS files.
```
jelbourn: I think we should omit this for now. Currently the demo-app doesn't use any functions, mixins, etc. from the library.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+gulp.task('build:devapp', [
+  'build:components',
+  ':build:vendor',
+  ':build:devapp:ts',
+  ':build:devapp:scss',
+  ':build:devapp:assets'
+]);
+
+/***************************************************************************************************
+ * DevApp Build Tasks.
```
jelbourn: e2e app

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+
+
+/***************************************************************************************************
+ * Global tasks.
+ */
+gulp.task('default', ['help']);
+
+gulp.task('help', function() {
+  const tasks = Object.keys(gulp.tasks)
+    .filter(x => !x.startsWith(':'))
```
jelbourn: nit: `x` > `t`

gulpfile.js
hansl: * moving to separate file: I wasn't sure where those files would live. I'm going to split this actually, a lot. I think it'll be for the best.
* I don't like `material`. I'll just put these files into their own utilities and import them.
* I'll try to move to typescript. We already have `ts-node` for e2e, so no added deps.

gulpfile.js
hansl: Parsing/reading `tsconfig` is really less than 1% of the total time spent.

gulpfile.js
hansl: I tried to keep this file 100% ES5 (easier). I'll move to TypeScript.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+gulp.task(':clean:spec', function() {
+  return gulp.src('dist/**/*.spec.*', { read: false })
+    .pipe(gulpClean());
+});
+gulp.task(':clean:assets', function() {
+  return gulp.src('dist/**/*+(.html|.css)', { read: false })
+    .pipe(gulpClean());
+});
+
+gulp.task('lint', createExecTask('tslint', ['-c', 'tslint.json', 'src/**/*.ts']));
```
jelbourn: Let's call this `tslint` and then have a `gulp lint` that runs tslint, stylelint, and madge.

gulpfile.js
```diff
@@ -0,0 +1,381 @@
...
+      fallback: 'index.html',
+      port: 4200
+    }));
+
+  gulp.watch(path.join(componentsDir, '**/*.ts'), [':build:components:ts']);
+  gulp.watch(path.join(componentsDir, '**/*.scss'), [':build:components:scss']);
+  gulp.watch(path.join(componentsDir, '**/*.html'), [':build:components:assets']);
+  gulp.watch(path.join(devAppDir, '**/*.ts'), [':build:devapp:ts']);
+  gulp.watch(path.join(devAppDir, '**/*.scss'), [':build:devapp:scss']);
+  gulp.watch(path.join(devAppDir, '**/*.html'), [':build:devapp:assets']);
```
jelbourn: How do these individual `gulp.watch` calls interact with `stream`?

gulpfile.js
hansl: It's technically not `tsc`. It's its own gulp typescript plugin, and it removes the sourcemap from the tsconfig before creating its program.

gulpfile.js
hansl: You'd think so, but it's not supported by Node's ES5 and a half.

gulpfile.js
hansl: `process` would override the global process object. `childProcess` works.

gulpfile.js
hansl: Splitting those in separate files solves this neatly.

gulpfile.js
hansl: Yes, like I said you can go to those directory and directly `npm publish`. There's no `deploy` folder anymore.

gulpfile.js
hansl: I wanted to provide some guidance, so I output the list of tasks when running `gulp` or `gulp help`, which is nice. I remove those who start with `:` or `ci:`.

Now, using `:` is just gulp common practice (see Angular's files or also [here](http://drewbarontini.com/articles/building-a-better-gulpfile/)). I don't care much about using `.`, if you insist.

scripts/release/stage-release.sh
```diff
@@ -1,72 +1,4 @@
...
-# Remove css files from src/
-find ./src -iname "*.css" | xargs rm
-
-# Remove test files from deploy/
-find ./deploy -iname "*.spec.d.ts" | xargs rm
-find ./deploy -iname "*.spec.js" | xargs rm
-find ./deploy -iname "*.spec.js.map" | xargs rm
-
-# To test the packages, simply `npm install` the package directories.
+$(npm bin)/gulp build:release
```
jelbourn: You can go ahead and just delete this file.

gulpfile.js
hansl: The `stream` watches and serves the `dist` folder itself, and `watch` runs tasks if a file was changed. When a file change, `build:components:XXX` will run and update the file in `dist`. The livereload server will see the file change and trigger a reload.

So, basically, indirectly.

scripts/release/stage-release.sh
hansl: Done.

gulpfile.js
hansl: Done.

gulpfile.js
hansl: Done.

gulpfile.js
hansl: Removed.

gulpfile.js
hansl: Done.

gulpfile.js
hansl: Done.

gulpfile.js
hansl: `scss` compiles the `css` files and copy the result. `assets` copy the `scss` sources.

hansl:
@jelbourn: Ready for a second pass, and you'll love it; I added a `gulp release` task that checks if you're logged in then build then for each component go in their dir and `npm publish`. Wheee.

gulpfile.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+const path = require('path');
+
+
```
jelbourn: Add a comment here pointing people to the appropriate location for the tasks?

scripts/ci/build-and-test.sh
```diff
@@ -14,42 +14,12 @@ start_tunnel
...
 wait_for_tunnel
 if is_lint; then
-  npm run tslint  
-  npm run ci:forbidden-identifiers
-  npm run stylelint
-elif is_circular_deps_check; then
-  # Build first because madge needs the JavaScript output.
-  ng build
-  npm run check-circular-deps
+  $(npm bin)/gulp ci:lint
```
jelbourn: Not sure if this would work, but could you add at the beginning `alias gulp='$(npm bin)/gulp'` and then just use `gulp` afterwards?

tools/gulp/tasks/ci.ts
```diff
@@ -0,0 +1,10 @@
+import {task} from 'gulp';
+
+
+task('ci:lint', ['ci:forbidden-identifiers', 'lint']);
+task('ci:test', ['test:single-run'], function() { process.exit(0); });
```
jelbourn:
```ts
task('ci:test', ['test:single-run'], () => process.exit(0));
``` 
?  (and similar elsewhere)

scripts/ci/build-and-test.sh
hansl: I'd rather not even put the time. This file will be going away in the next PR.

gulpfile.js
hansl: Done.

tools/gulp/tasks/e2e.ts
```diff
@@ -0,0 +1,50 @@
...
+
+task(':build:e2eapp:vendor', vendorTask());
+task(':build:e2eapp:ts', [':build:components:ts'], tsBuildTask(appDir));
+task(':build:e2eapp:scss', [':build:components:scss'], sassBuildTask(outDir, appDir, []));
+task(':build:e2eapp:assets', copyTask(appDir, outDir));
+
+task('build:e2eapp', buildAppTask('e2eapp'));
+
+
+task(':test:protractor:setup', execNodeTask('protractor', 'webdriver-manager', ['update']));
```
jelbourn: Good call on the setup task.

tools/gulp/tasks/release.ts
```diff
@@ -0,0 +1,56 @@
...
+    .forEach(dirName => {
+      const componentPath = path.join(DIST_COMPONENTS_ROOT, dirName);
+      const stat = statSync(componentPath);
+
+      if (!stat.isDirectory()) {
+        return;
+      }
+
+      process.chdir(componentPath);
+      execSync('npm publish');
```
jelbourn: This also needs `--access public` and an option for the label.

hansl: Done

tools/gulp/tasks/ci.ts
hansl: Added a comment.

test/protractor.conf.js
```diff
@@ -13,14 +13,13 @@ const config = {
   useAllAngular2AppRoots: true,
   specs: [ path.join(__dirname, '../e2e/**/*.e2e.ts') ],
   baseUrl: E2E_BASE_URL,
-  allScriptsTimeout: 30000,
-  getPageTimeout: 30000,
+  allScriptsTimeout: 120000,
```
jelbourn: Should we have a TODO to lower this back down?

tools/gulp/task_helpers.ts
```diff
@@ -0,0 +1,190 @@
...
+          done(options.errMessage);
+        }
+        return;
+      }
+      done();
+    });
+  }
+}
+
+export function execNodeTask(packageName: string, executable: string[] | string, args?: string[]) {
```
jelbourn: Can you add comments with examples for `execTask` and `execNodeTask`?

tools/gulp/tasks/release.ts
```diff
@@ -0,0 +1,58 @@
...
+task(':publish:whoami', execTask('npm', ['whoami'], {
+  silent: true,
+  errMessage: 'You must be logged in to publish.'
+}));
+
+task(':publish:logout', execTask('npm', ['logout']));
+
+task(':publish', function() {
+  const label = process.argv.slice(2)[1];  // [0] would be ':publish'
+  const labelArg = label ? `--dist-tag ${label}` : '';
```
jelbourn: `--dist-tag`? Why not just `--tag`?

tools/gulp/tasks/spec.ts
```diff
@@ -0,0 +1,34 @@
+import gulp = require('gulp');
```
jelbourn: Rename this file `spec.ts` > `unit-test.ts`?

jelbourn:
LGTM aside from last few comments

test/protractor.conf.js
hansl: We'll only have more tests, so I don't see a future where this will go down significantly.

tools/gulp/task_helpers.ts
hansl: Done.

tools/gulp/tasks/release.ts
hansl: My bad. `dist-tag` is the command, `tag` is the argument. Not confusing at all.

tools/gulp/tasks/spec.ts
hansl: Done.

tools/gulp/tasks/e2e.ts
hansl: Acknowledge.

PR #950, opened 2016-08-05
Add NgModules for material

jelbourn:
R: @hansl @kara 

This PR includes:
* Temporarily move to head Angular until RC5 is released
* Introducing modules for each component and remove all `directives: [...]`
* Introducing an "all" module (`MaterialModule`)
* Bootstrapping via modules
* Update tests to use modules
* Workaround for https://github.com/angular/angular/issues/10503
* Bumps the version to alpha.6-3 (which I published under `@experimental` yesterday)
* Update to new `Provider` syntax

PR #956, opened 2016-08-08
chore: fix Material to support offline compiler.

gajeravijayb:


jelbourn:
@gajeravijayb I think you may be using an out-of-date version of the offline compiler. As of #950, everything in Material should work fine with the latest offline compiler (Angular master).

PR #961, opened 2016-08-08
docs: update guides and public APIs for NgModules

jelbourn:
R: @kara @hansl

PR #968, opened 2016-08-09
update: forbidden identifiers check all source files

DevVersion:
* Removes invalid relative import from other scope.
* Forbidden identifiers should run against all sources when not checking a PR.
* Cleanup Forbidden identifiers script, by using ES6 arrow functions.

### DIFF
--- a/scripts/ci/forbidden-identifiers.js
+++ b/scripts/ci/forbidden-identifiers.js
@@ -1,27 +1,21 @@
 #!/usr/bin/env node
 
 'use strict';
+
 /*
- * This script analyzes the current commits of the CI.
- * It will search for blocked statements, which have been added in the commits and fail if present.
+ * The forbidden identifiers script will check for blocked statements and also detect invalid
+ * imports of other scope packages.
+ *
+ * When running against a PR, the script will only analyze the specific amount of commits inside
+ * of the Pull Request.
+ *
+ * By default it checks all source files and fail if any errors were found.
  */
 
 const child_process = require('child_process');
 const fs = require('fs');
 const path = require('path');
-
-const exec = function(cmd) {
-  return new Promise(function(resolve, reject) {
-    child_process.exec(cmd, function(err, stdout /*, stderr */) {
-      if (err) {
-        reject(err);
-      } else {
-        resolve(stdout);
-      }
-    });
-  });
-};
-
+const glob = require('glob').sync;
 
 const blocked_statements = [
   '\\bddescribe\\(',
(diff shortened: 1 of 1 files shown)


jelbourn:
@DevVersion FYI I fixed that import in my other PR #961 (it was failing travis)

DevVersion:
@jelbourn That works too. This PR is mainly to change the behavior of the forbidden identifiers check.

I had a quick discussion with @hansl regarding running the forbidden identifiers check against all source files when not validating a Pull Request.

> I will rebase my PR once your fix is in.

scripts/ci/forbidden-identifiers.js
```diff
@@ -34,68 +28,135 @@ const blocked_statements = [
...
+        })
+        .map(line => line.split(/\s+/, 2)[1]);
     });
 }
 
 /**
+ * Resolves all files, which should run against the forbidden identifiers check.
+ * @return {Promise.<Array.<string>>} Files to be checked.
+ */
+function resolveFiles() {
```
jelbourn: The name of this function doesn't do a good job of describing what it does.

src/components/button/button.ts
```diff
@@ -11,7 +11,7 @@ import {
 } from '@angular/core';
 import {CommonModule} from '@angular/common';
 import {BooleanFieldValue} from '@angular2-material/core/annotations/field-value';
-import {MdRippleModule} from '../../core/ripple/ripple';
```
jelbourn: Should already be fixed.

scripts/ci/forbidden-identifiers.js
DevVersion: Can't think of anything better - I'll try to find something ;)

src/components/button/button.ts
DevVersion: Yeah, as I mentioned, I still have to rebase. Will do right now.

scripts/ci/forbidden-identifiers.js
```diff
@@ -34,68 +28,135 @@ const blocked_statements = [
...
-function findForkPoint() {
-  return exec('git merge-base --fork-point HEAD master')
-    .then(function(stdout) {
-      return stdout.split('\n')[0];
+
+resolveFiles()
+
+  /* Only match .js or .ts, and remove .d.ts files. */
+  .then(fileList => {
+    return fileList.filter(name => name.match(/\.(js|ts)$/) && !name.match(/\.d\.ts$/));
```
jelbourn: You can use `,search()` instead of `.match()` whenever you're just checking for the presence of a match and don't need the match itself.

scripts/ci/forbidden-identifiers.js
```diff
@@ -34,68 +28,135 @@ const blocked_statements = [
...
+  })
+
+  /* Read content of the filtered files */
+  .then(fileList => {
+    return fileList.map(fileName => {
+      return {
+        name: fileName,
+        content: fs.readFileSync(fileName, 'utf-8')
+      };
```
jelbourn: Could make this a bit more concise:
```ts
.then(files => files.map(name => ({name, content: fs.readFileSync(name, 'utf-8')})))
```

scripts/ci/forbidden-identifiers.js
DevVersion: The most of those changes weren't actually from me. Also it seems like `.search` is not super elegant, because it returns `-1`, which can't be easily tested with a falsy check.

I recommend `regex.test()` which returns a plain boolean.

scripts/ci/forbidden-identifiers.js
```diff
@@ -34,68 +28,135 @@ const blocked_statements = [
...
+
+      return errors;
+
+    }, []);
+  })
+
+  /* Print the resolved errors to the console */
+  .then(errors => {
+    if (errors.length > 0) {
+      console.error('Error: You are using a statement in your commit, which is not allowed.\n');
```
jelbourn: While you're in here, can you remove the common in this line? Or we could also change it to something like 
`"You are using one or more blocked statements:"`

scripts/ci/forbidden-identifiers.js
DevVersion: Wow, those tricks combined :smile:

Didn't know that you could create a JSON object without specifying a key :)

jelbourn:
lgtm

PR #971, opened 2016-08-09
chore(changelog): properly name tags for github compare links

DevVersion:
* The convention changelog currently auto-prefixes the previous semver tag with a `v`, which causes the github compare links to not work.
* Manually resolve the previous and current tag from the git metadata. (As in angular/material#9264)

> I know the way we resolve the latest / previous tags looks at first glance pretty odd.


jelbourn:
LGTM

PR #976, opened 2016-08-09
Add Thumb Label to Slider

iveysaur:
Adds a thumb label (https://material.google.com/components/sliders.html#sliders-discrete-slider) to the slider. The thumb label is displayed when `thumb-label` is added to the slider. It will appear whenever the slider is active (a click or a slide).

R: @jelbourn @kara 
CC: @hansl @robertmesserle

### DIFF
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -196,7 +196,7 @@ md-slider *::after {
   display: none;
 }
 
-.md-slider-container.md-slider-active.md-slider-thumb-label-showing .md-slider-thumb {
+.md-slider-active.md-slider-thumb-label-showing .md-slider-thumb {
   transform: scale(0);
 }
 
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -21,12 +21,11 @@ $md-slider-disabled-color: rgba(black, 0.26);
 $md-slider-thumb-arrow-height: 16px !default;
 $md-slider-thumb-arrow-width: 28px !default;
 
-$md-slider-thumb-label-height: 28px !default;
-$md-slider-thumb-label-width: 28px !default;
+$md-slider-thumb-label-size: 28px !default;
 // The thumb has to be moved down so that it appears right over the slider track when visible and
 // on the slider track when not.
 $md-slider-thumb-label-top: ($md-slider-thickness / 2) -
-    ($md-slider-thumb-default-scale * $md-slider-thumb-size / 2) - $md-slider-thumb-label-height -
+    ($md-slider-thumb-default-scale * $md-slider-thumb-size / 2) - $md-slider-thumb-label-size -
     $md-slider-thumb-arrow-height + 10px !default;
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/slider/slider.scss
```diff
@@ -137,6 +148,58 @@ md-slider *::after {
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  position: absolute;
+  left: -($md-slider-thumb-label-height / 2);
+  top: $md-slider-thumb-label-top;
+  width: $md-slider-thumb-label-width;
+  height: $md-slider-thumb-label-height;
+  border-radius: max($md-slider-thumb-label-height, $md-slider-thumb-label-width);
```
robertmesserle: I'm not 100% sure what this line is doing.  Will these ever not be equal?  If so, how would we want this to look?

robertmesserle: To clarify, your method would result in a sort of pill-shape when these values do not match.  If this is desired, I think what you actually want is half of the smaller dimension: `border-radius: min($md-slider-thumb-label-height, $md-slider-thumb-label-width) / 2`.  Your method will work too, but that's because the browser caps the border radius at this value.

If you wanted more of an oval shape, you can also simplify it to `border-radius: 50%`

src/components/slider/slider.scss
```diff
@@ -137,6 +148,58 @@ md-slider *::after {
...
+
+  transform: scale(0.4) translate3d(0, (-$md-slider-thumb-label-top + 10) / 0.4, 0);
+  transition: 300ms $swift-ease-in-out-timing-function;
```
robertmesserle: We should probably specify which properties we want to transition.  It may not be necessary now, but it can be really annoying to debug if a transition applies to an incompatible property.  As was my case with Tooltip a few weeks ago.

robertmesserle:
LGTM, just a few minor comments

src/components/slider/slider.ts
```diff
@@ -47,6 +47,11 @@ export class MdSlider implements AfterContentInit {
   @HostBinding('attr.aria-disabled')
   disabled: boolean = false;
 
+  /** Whether or not to show the thumb label. */
+  @Input('thumb-label')
+  @BooleanFieldValue()
+  private _thumbLabel: boolean = false;
```
jelbourn: This can't be `private` if you're referencing it in a template when doing offline compile. Just remove the `private` and leave the underscore.

src/components/slider/slider.spec.ts
```diff
@@ -26,7 +26,8 @@ describe('MdSlider', () => {
         SliderWithValue,
         SliderWithStep,
         SliderWithAutoTickInterval,
-        SliderWithSetTickInterval
+        SliderWithSetTickInterval,
+        SliderWithThumbLabel
```
jelbourn: Include a trailing comma
(trailingcommas4life)

src/components/slider/slider.spec.ts
```diff
@@ -516,6 +517,77 @@ describe('MdSlider', () => {
...
+    });
+
+    it('should update the thumb label text on click', () => {
+      expect(thumbLabelTextElement.textContent).toBe('0');
+
+      dispatchClickEvent(sliderTrackElement, 0.49);
+      fixture.detectChanges();
+
+      // The thumb label text is set to the slider's value. These should always be the same.
+      expect(thumbLabelTextElement.textContent).toBe(`${sliderInstance.value}`);
```
jelbourn: Use the hard-coded value instead of `sliderInstance.value`

src/components/slider/slider.spec.ts
```diff
@@ -572,6 +644,17 @@ class SliderWithAutoTickInterval { }
 })
 class SliderWithSetTickInterval { }
 
+@Component({
+  template: `<md-slider thumb-label></md-slider>`,
+  styles: [`
+    .md-slider-thumb-label, .md-slider-thumb-label-text {
```
jelbourn: Add comment for why this is necessaery

iveysaur:
@jelbourn: Addressed comments

jelbourn:
LGTM

Issue #974, opened 2016-08-09
bug(rc5 and alpha7)url_resolver.js:238Uncaught TypeError: uri.match is not a function


qdouble:
I'm refiling for this issue: https://github.com/angular/material2/issues/963 as it was not resolved in alpha 7. I've tested it on two fresh seed projects and the error still occurs, so it should be a common problem with Webpack at least, not sure if this shows up in a SystemJS build or not.

FoodBuster:
I second this. Having the same problem, using the new angular-cli webpack release.

qdouble:
This seems to be related to the moduleId how it's dealing with the moduleId, if you comment out the moduleId of the components, they compile without error.

qdouble:
It appears this pull request will pull this issue https://github.com/angular/angular/pull/10595/commits/f54f9efc0e70eb1c71ac8eb6a8ed690f1e7f09f9. You guys can close this if you want to as it appears to be an angular compiler issue and not specific to material

FoodBuster:
qdouble, that pull request does fix the "TypeError: uri.match is not a function" problem, but gives me a BaseException error with the following message: 

`Template parse errors:↵Can't bind to 'md-ripple-trigger' since it isn't a known property of 'div'. ("*ngIf="isRippleEnabled()" class="md-button-ripple" [class.md-button-ripple-round]="isRoundButton()" [ERROR ->][md-ripple-trigger]="getHostElement()" [md-ripple-color]="isRoundButton() ? 'rgba(255, 255, 255, 0.2)"): MdButton@0:180↵Can't bind to 'md-ripple-color' since it isn't a known property of 'div'. ("ton-ripple" [class.md-button-ripple-round]="isRoundButton()" [md-ripple-trigger]="getHostElement()" [ERROR ->][md-ripple-color]="isRoundButton() ? 'rgba(255, 255, 255, 0.2)' : ''" md-ripple-background-color="rgb"): MdButton@0:219`

If this is an angular-cli-specific issue, happy to take it up there.

qdouble:
@FoodBuster that's an unrelated error, it saying 'md-ripple-trigger' isn't defined on the MdButton directive/component.

FoodBuster:
@qdouble Thanks, got it working.

jelbourn:
I have released an `alpha.7-2` that _should_ fix this.

zigzag95:
Thanks

neclepsio:
[...]

PR #984, opened 2016-08-10
fix(inline-resources): remove module ids from component metadata.

DevVersion:
Removes the module ids of the component metadata.

Since the templates and styles are now inlined, the module id has become unnecessary and can cause unexpected issues.

This also fixes https://github.com/angular/material2/issues/974 in meanwhile.

FYI: The regex is designed to match the whole line (including line feeds) and only the compiler specific module ids.


zigzag95:
Do we have to wait for alpha.8.
This is a very blocker issue for Webpack and doing this is very expensive :

```
   {
           test: /.js$/,
           loader: 'string-replace-loader',
           query: {
             search: 'moduleId: module.id,',
             replace: '',
             flags: 'g'
           }
          }
```

jelbourn:
LGTM

PR #986, opened 2016-08-10
style: cleans up code

jesselpalmer:
Cleans up e2es style a little to conform closer to the [Angular Style Guide](https://angular.io/styleguide).

### DIFF
--- a/src/e2e-app/button/button-e2e.ts
+++ b/src/e2e-app/button/button-e2e.ts
@@ -1,4 +1,5 @@
-import {Component} from '@angular/core';
+import { Component } from '@angular/core';
+
 
 @Component({
   moduleId: module.id,
--- a/src/e2e-app/tabs/tabs-e2e.ts
+++ b/src/e2e-app/tabs/tabs-e2e.ts
@@ -1,8 +1,9 @@
-import {Component} from '@angular/core';
+import { Component } from '@angular/core';
+
 
 @Component({
   moduleId: module.id,
   selector: 'tabs-e2e',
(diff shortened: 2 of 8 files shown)


src/e2e-app/button/button-e2e.ts
```diff
@@ -1,4 +1,5 @@
-import {Component} from '@angular/core';
+import { Component } from '@angular/core';
```
pkozlowski-opensource: Why on earth this is in the style guide??? This is just a formatting choice that each project should make on its own. As long as formatting is consistent it should be none of style guide business....

jelbourn: If this is in the Angular style guide, we're not going to follow it.

src/e2e-app/tabs/tabs-e2e.ts
```diff
@@ -1,8 +1,9 @@
...
+import { Component } from '@angular/core';
+
 
 @Component({
   moduleId: module.id,
   selector: 'tabs-e2e',
   templateUrl: 'tabs-e2e.html',
 })
-export class BasicTabs {}
+export class TabsE2E {}
```
jelbourn: I'm actually planning on renaming all of these the other direction. I dislike calling them `XxxE2e` and want the same to reflect what the test is, e.g. `BasicTabs`, `TabsInDialog`, etc.

jelbourn:
Going to keep these things as-is.

jesselpalmer:
@jelbourn Works for me.

PR #987, opened 2016-08-10
chore: add tooltip package.json & bump to alpha.7-2

jelbourn:
R: @robertmesserle


clydin:
Not sure if it's intentional but dialog doesn't have a package.json

jelbourn:
@clydin dialog is not yet done enough to be published.

clydin:
That's what I assumed. Only reason i mentioned it was that I noticed it was included in the 'all' component

PR #991, opened 2016-08-10
fix(button): Avoid breakage by angular change

vikerman:
Type is becoming Type<?> in https://github.com/angular/angular/pull/10616


vikerman:
I signed it!

PR #1004, opened 2016-08-11
chore: temporarily remove ripples from buttons

jelbourn:
R: @kara 
CC: @dozingcat 

Temporarily remove ripples from buttons (will be added back before alpha.8).


kara:
LGTM

PR #1015, opened 2016-08-12
chore(toolbar): remove TestComponentBuilder

jelbourn:
R: @DevVersion

### DIFF
--- a/src/components/toolbar/toolbar.spec.ts
+++ b/src/components/toolbar/toolbar.spec.ts
@@ -18,29 +18,29 @@ describe('MdToolbar', () => {
   it('should apply class based on color attribute', () => {
     let fixture = TestBed.createComponent(TestApp);
     let testComponent = fixture.debugElement.componentInstance;
-    let toolbarDebugElement = fixture.debugElement.query(By.css('md-toolbar'));
+    let toolbarElement = fixture.debugElement.query(By.css('md-toolbar')).nativeElement;
 
     testComponent.toolbarColor = 'primary';
     fixture.detectChanges();
 
-    expect(toolbarDebugElement.nativeElement.classList.contains('md-primary')).toBe(true);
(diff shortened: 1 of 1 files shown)


src/components/toolbar/toolbar.spec.ts
```diff
@@ -20,45 +15,32 @@ describe('MdToolbar', () => {
...
 
-@Component({
-  selector: 'test-app',
-  template: `
-    <md-toolbar [color]="toolbarColor">
-      <span>Test Toolbar</span>
-    </md-toolbar>`
-})
+
+@Component({template: `<md-toolbar [color]="toolbarColor"><span>Test Toolbar</span></md-toolbar>`})
```
DevVersion: Not necessary, but since the template is now in a single line, we could probably remove the `span` element to make the template more clean.

src/components/toolbar/toolbar.spec.ts
```diff
@@ -20,45 +15,32 @@ describe('MdToolbar', () => {
...
-    return builder.createAsync(TestApp).then((fixture) => {
-      let testComponent = fixture.debugElement.componentInstance;
-      let toolbarDebugElement = fixture.debugElement.query(By.css('md-toolbar'));
-
-      testComponent.toolbarColor = 'primary';
-      fixture.detectChanges();
+  it('should apply class based on color attribute', () => {
+    let fixture = TestBed.createComponent(TestApp);
+    let testComponent = fixture.debugElement.componentInstance;
+    let toolbarDebugElement = fixture.debugElement.query(By.css('md-toolbar'));
```
DevVersion: This is old from me - what do you think about using the `nativeElement` right here, instead of calling its getter each time?

DevVersion:
LGTM.

> I've added two minor comments

PR #1027, opened 2016-08-12
chore(overlay): change ComponentResolver to ComponentFactoryResolver

jelbourn:
R: @kara


kara:
There was a target failing but it looked like it could have been a flake. LGTM.

PR #1030, opened 2016-08-13
chore(live-announcer): remove TestComponentBuilder

jelbourn:
R: @DevVersion


DevVersion:
LGTM. 

> Seems like Browserstack errored, but it seems to be unrelated and probably due to some flakiness

PR #1033, opened 2016-08-13
chore(ripple): remove TestComponentBuilder

jelbourn:
R: @kara or @dozingcat

### DIFF
--- a/src/core/ripple/ripple.spec.ts
+++ b/src/core/ripple/ripple.spec.ts
@@ -1,12 +1,7 @@
-import {
-    TestBed,
-    inject,
-    async,
-    TestComponentBuilder,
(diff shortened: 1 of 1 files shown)


src/core/ripple/ripple.spec.ts
```diff
@@ -1,13 +1,8 @@
-import {
-    TestBed,
-    inject,
-    async,
-    TestComponentBuilder,
-    ComponentFixture,
-} from '@angular/core/testing';
+import {TestBed, async, ComponentFixture} from '@angular/core/testing';
```
dozingcat: Is async used anywhere after this change?

dozingcat:
LGTM

PR #1044, opened 2016-08-14
fix(tabs): change active tab css class from md-active to md-tab-active

tlein:
Reference to the active tab was changed from md-active to
md-tab-active but the css class that set the body of the tab
back to block display was missed during this renaming effort.
This resulted in the body of a tab to always be empty. This
commit changes the css class from md-active to md-tab-active.


jelbourn:
LGTM

PR #1060, opened 2016-08-16
chore: bump version to alpha.7-3

jelbourn:
R: @hansl or @kara 

For the future: do either of you care if we skip the review process on version bumps like this?


hansl:
Only if we're using a testable and repeatable process (ie. a single script).

hansl:
LGTM.

PR #1066, opened 2016-08-17
fix(input): attributes should match native ones

### LINKED ISSUE #1065: Un-camelCase md-input properties
Many of the properties of md-input have been made camelCase white the native input uses all lowercase, e.g., `readOnly` vs. `readonly`. MdInput should be changed to match the native input.

cc @hansl

DevVersion:
* The input attribute bindings did not match the native ones, and this could cause confusion.
   Now since they are consistent it is super simple to port a previous Angular 2 native input to an `md-input`

Fixes #1065.


jelbourn:
@DevVersion LGTM, just needs rebase

DevVersion:
@jelbourn Done

PR #1041, opened 2016-08-14
feat(dialog): add backdrop

jelbourn:
R: @kara @hansl 

This adds support to the Overlay to add a backdrop. Components that consume the overlay can take the clicks on the backdrop and do whatever they want with them. 

This also cleans up the dialog tests; they no longer need a `setTimeout` with the recent changes to `fakeAsync` in core.

PR #1079, opened 2016-08-19
feat(overlay): make overlays synchronous

jelbourn:
R: @kara @hansl 

Also portals. 
This dramatically cleans up a lot of stuff and is probably faster.

If I was smart I would have merged #1041 first.

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -43,8 +43,6 @@ export class MdDialog {
     let overlayRef = this._createOverlay(config);
     let dialogContainer = this._attachDialogContainer(overlayRef, config);
 
-    // TODO: probably need to wait for dialogContainer ngOnInit before attaching.
-
     return this._attachDialogContent(component, dialogContainer, overlayRef);
   }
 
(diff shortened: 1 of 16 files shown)


jelbourn:
This might also fix our unit test flakiness...

src/lib/dialog/dialog.ts
```diff
@@ -39,21 +39,21 @@ export class MdDialog {
...
-    let overlayRef: OverlayRef;
+  open<T>(component: ComponentType<T>, config: MdDialogConfig): MdDialogRef<T> {
+    let overlayRef = this._createOverlay(config);
+    let dialogContainer = this._attachDialogContainer(overlayRef, config);
 
-    return this._createOverlay(config)
-        .then(overlay => overlayRef = overlay)
-        .then(overlay => this._attachDialogContainer(overlay, config))
-        .then(containerRef => this._attachDialogContent(component, containerRef, overlayRef));
+    // TODO: probably need to wait for dialogContainer ngOnInit before attaching.
```
hansl: This seems to imply we should still by async...?

hansl:
One question, but LGTM in general.

src/lib/dialog/dialog.ts
jelbourn: Was an in-progress thought while I was working, ended up not being the case. Removed.

PR #1099, opened 2016-08-22
build: fix publish gulp task and output more information.

hansl:
The whole process will fail on first failure, but you will not be logged out on this.


jelbourn:
LGTM

PR #1101, opened 2016-08-22
chore: re-add moduleId temporarily

jelbourn:
R: @hansl 
Needed by Google's internal build tools for the near term.


hansl:
LGTM.

PR #1103, opened 2016-08-23
fix checkbox: add fixed width to checkbox

### LINKED ISSUE #1097: Squished checkbox (md-checkbox)

Checkbox (md-checkbox) squishes if the label text is too long.

Checkbox has always the same size, independent from the label text lenght

Checkbox squishes if label text is too long

ex. 
1. Set parent element to fixed width
2. add a very long label text 

https://plnkr.co/Fw8GMnMA7DGDXW1ZdXjY 

Chrome : Version 52.0.2743.116 (64-bit)
OS X : 10.11.3 (15D21)

dimfried:
add fixed width to the checkbox so they do not squish if the parent
wrapper has a fixed width and the label text is too long.

closes #1097

### DIFF
--- a/src/lib/checkbox/checkbox.scss
+++ b/src/lib/checkbox/checkbox.scss
@@ -240,6 +240,7 @@ md-checkbox {
   vertical-align: middle;
   white-space: nowrap;
   width: $md-checkbox-size;
+  flex: 0 0 $md-checkbox-size;
 
   [dir='rtl'] & {
     margin: {


src/lib/checkbox/checkbox.scss
```diff
@@ -240,6 +240,7 @@ md-checkbox {
   vertical-align: middle;
   white-space: nowrap;
   width: $md-checkbox-size;
+  flex: 0 0 $md-checkbox-size;
```
DevVersion: Why not just applying `flex-shrink: 0`? 
Setting another flex basis should be unnecessary.

dimfried: @DevVersion yeah, would also work this way. I can change it very quick if you want. Sorry for that!

jelbourn: Agreed- `flex-shrink: 0` would be better.

tinayuangao: +1 for `flex-shrink: 0`

jelbourn:
Closed in favor of #1713

PR #1113, opened 2016-08-24
fix(ngc): _onDragStart/End are called with one param

### LINKED ISSUE #1112: Some components don't build with ngc
one example:
```
third_party/javascript/angular2_material/src/lib/slide-toggle/slide-toggle.ngfactory.ts:300:29: Supplied parameters do not match any signature of call target.
third_party/javascript/angular2_material/src/lib/slide-toggle/slide-toggle.ngfactory.ts:310:29: Supplied parameters do not match any signature of call target.
```

because `_onDragStart` does not accept parameters:
https://github.com/angular/material2/blob/d6f3e7773c0fcea008e48807617fa0e39263049e/src/lib/slide-toggle/slide-toggle.ts#L216
but is called with one parameter:
https://github.com/angular/material2/blob/d6f3e7773c0fcea008e48807617fa0e39263049e/src/lib/slide-toggle/slide-toggle.html#L7

alexeagle:
Fixes #1112

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.html
+++ b/src/lib/slide-toggle/slide-toggle.html
@@ -4,9 +4,9 @@
     <div class="md-slide-toggle-bar"></div>
 
     <div class="md-slide-toggle-thumb-container"
-         (slidestart)="_onDragStart($event)"
+         (slidestart)="_onDragStart()"
          (slide)="_onDrag($event)"
-         (slideend)="_onDragEnd($event)">
+         (slideend)="_onDragEnd()">
 
       <div class="md-slide-toggle-thumb">
         <div class="md-ink-ripple"></div>


src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -213,7 +213,7 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
 
 
   /** TODO: internal */
-  _onDragStart() {
+  _onDragStart(event: Event) {
```
DevVersion: - Why don't we remove the parameters from the HTML instead (This does not show any linter warnings / infos etc.)
- Or make it consistent with the `onDrag` function, by using `HammerInput` instead? (to get proper autocompletion)

alexeagle: good idea, removed the parameter at the callsite instead.

DevVersion: Thank you :)

jelbourn:
LGTM

PR #1114, opened 2016-08-25
fix(sidenav): turn off view encapsulation and refactor css

jelbourn:
R: @hansl 

This change is necessary for the sidenav to continue working with https://github.com/angular/angular/pull/10997 and is blocking that PR's syncing into Google. The PR tightens the css encapsulation such that rules *only* apply to elements within the view of the component in which they are defined. In this case, the styles for `md-sidenav` were defined in the styles for `md-sidenav-layout`, so they were no longer being applied to the sidenav element (which is in a different view). 

I've also refactored the css to reduce the specificity across the board.


src/demo-app/tabs/tab-group-demo.ts
```diff
@@ -8,7 +7,6 @@ import {Observable} from 'rxjs/Observable';
   selector: 'tab-group-demo',
   templateUrl: 'tab-group-demo.html',
   styleUrls: ['tab-group-demo.css'],
-  pipes: [AsyncPipe],
```
jelbourn: This change was necessary to work with the patched PR.

hansl:
LGTM.

PR #1121, opened 2016-08-25
fix(RC6): add directives for custom elements that are part of the APIs.

hansl:


hansl:
This was tested with https://github.com/angular/angular/pull/11026 and passed.

kara:
LGTM

PR #1108, opened 2016-08-23
feat: add `forRoot` to modules

jelbourn:
R: @kara @hansl 

Also remove deprecated `MD_XXX_DIRECTIVES`. 

 @robwormald or @IgorMinar can one of you take a look at this and make sure I'm doing "the right thing" WRT use of `forRoot`, especially as it pertains to the `MaterialModule`?

PR #1122, opened 2016-08-25
feat(modules): add `forRoot` to material modules

hansl:
Rebasing https://github.com/angular/material2/pull/1108 on top of https://github.com/angular/material2/pull/1121.  I thought this was committed already but it wasn't.


hansl:
Review already done in #1108, so LGTM.

PR #913, opened 2016-07-24
chore: add script to automatically update plunker bundle

DevVersion:
Adds a script to automatically update the plunker bundle by using the Angular CLI with Production Mode.

The firebase hosting will contain files:
- `HEAD_bundle.js` (Latest Bundle for each commit)
- `2.0.0-alpha.6` (Bundle for that tag)
- etc.

---
Open Points:
- When deploying to firebase, we overwrite the current hosting and loose the old bundles (like for tags)<br/>We need a way to pull the old files or find another approach for hosting

Please give some thoughts @jelbourn

PR #1138, opened 2016-08-31
chore(ci): finish different matrixes in fast finish mode

DevVersion:
- Travis CI runs the optional tasks in "waiting" mode, which means that the whole build is only ready once the optional tasks are also ready.
- The Travis Fast Finish mode will mark the build as finished, once all "required" tasks have finished (https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/)

FYI: I decided to make this now an extra PR, because #913 has still open questions, and the build right now keeps running ~6-7 minutes extra. (https://travis-ci.org/angular/material2/builds/156389515)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ env:
     - MODE=browserstack_optional
 
 matrix:
+  fast_finish: true
   allow_failures:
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM - thanks!

PR #1140, opened 2016-08-31
docs(tabs): remove invalid property in documentation

### LINKED ISSUE #1134: Setting focusIndex won't work in md-tab-group

Can't bind to 'focusIndex' since it isn't a known property of 'md-tab-group'.

md-tab-group should accept focusIndex property.

Adding focusIndex property to md-tab-group causes an unhandled promise rejection.

http://plnkr.co/edit/h7d6DDvqRFCfmChp7Qvj

focusIndex property should works as expected.

Tested with angular `2.0.0-rc.5` and Material `2.0.0-alpha.7-4`

DevVersion:
* The `focusIndex` has been removed, and was actually the same as the `selectedIndex` from the description as well.
* Also added a quick example for the `selectedIndex`

Fixes #1134

### DIFF
--- a/src/lib/tabs/README.md
+++ b/src/lib/tabs/README.md
@@ -1,6 +1,20 @@
 # MdTabGroup
 Tab groups allow the user to organize their content by labels such that only one tab is visible at any given time.
 
+## `<md-tab-group>`
+### Properties
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `selectedIndex` | `number` | The index of the currently active tab. |
+
+### Events
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `focusChange` | `Event` | Fired when focus changes from one label to another |
+| `selectChange` | `Event` | Fired when the selected tab changes |
+
 ### Examples
 A basic tab group would have the following markup.
 ```html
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

marc-ferrer:
In fact, setting selectedIndex does no get you the same result as the focus.
The selectedIndex property does not set the background in the tab selector as the focus does when you click on the tab. Which was actually my concern with #1134.

Does this issue mean that the selectedIndex will now have the same behaviour as the focus, or conversely we won't be able to set the focus on the desired tab?

DevVersion:
@marc-ferrer Sure it does not. But as said in my PR description, the documentation had the same description for the `selectedIndex` and `focusIndex`.

Also I guess, the `focusIndex` input has been removed, and is now only for internal use, to keep track of the current focused tab label wrapper.

If we make this public as an `@Input`, then it doesn't even work, because the tabs wrappers are not initialized at this time.

This functionality may come back in the future, but right now it does not make any sense (also since only tabs, which are selected are currently focusable)

marc-ferrer:
@DevVersion So I guess that it is not possible to manually set focus on a tab currently. And selectedIndex behaviour is not expected to change to include the focus efect.

> This functionality may come back in the future, but right now it does not make any sense (also since only tabs, which are selected are currently focusable)

Then how are we suposed to manage the tabs without mouse? When you use the tab key to select a tab from a list the focus moves accross the tabs and then you select the tab you want to move into.
How will this standard behaviour be implemented when "only tabs, which are selected are currently focusable"?

Don't get me wrong, I get that focusIndex got published as a public property when it shouldn't have been this way, and does not seem easy to implement an actual focusIndex. I just got a bit confused with that sentence of yours since what does not make any sense for me, is that only selected tabs are focusable.
All I wanted was to make sure how focus and select would actually work if `selectedIndex` and `focusIndex` had the same description. I Didn't want to be rude at all.

DevVersion:
@marc-ferrer I'm sorry, my sentence was probably wrong and invalid. 

I'm only talking about the things I can see right now - I'm not the author of this component.

Right now, you are able to move the focus between the different `tab-label-wrappers` per Arrow keys, but not with the <kbd>Tab</kbd> key.

A few minutes ago, I made the `focusIndex` property public again, to test how it would look and work.

But from my findings it seems like this property was `never` marked as an `@Input()` . 

---
Right now, this PR just makes sure that the documentation matches the **current** state of implementation.

Issue #1125, opened 2016-08-26
Inconsistancy of triggering change event on the very first init across components.


colinfang:
For example, in `MdRadioGroup`, it only fires a change event if this isn't the first time the value is ever set. Same applies to `TabGroup`

However in `MdButtonToggleGroup`, it fires a change event even if it is the first time. Same applies to `SlideToggle`

jelbourn:
If a component fires a change event upon initialization it is a bug.

DevVersion:
I just took a look at it for the `md-slide-toggle` - I was not able to recognize any events at initialization.

Is it possible that you use an older version of Material 2?

colinfang:
`change` is emitted before click on the `MdButtonToggle`

http://plnkr.co/edit/LiCpyMHS1Ok98ZVWvoXs?p=preview

DevVersion:
@colinfang I did not reference that issue to be closed, because I thought there are also issues with the `TabGroup`?

DevVersion:
Just tested the `selectChange` and `focusChange` output and everything seemed to work as expected.

PR #1144, opened 2016-08-31
fix(button-toggle): toggle group should not emit an initial change event.

DevVersion:
This is the addition of https://github.com/angular/material2/commit/23a61ab8687d5e4f962c39423ed095959d2da82c, which was only taking care of the `ButtonToggle` itself, and not of the group.

Since the `md-button-toggle-group` does not contain any underlaying input element, and we can't use the `emit event if native input does`  approach, we need to wait for the component view to be ready.

References #1125

### DIFF
--- a/src/lib/button-toggle/button-toggle.spec.ts
+++ b/src/lib/button-toggle/button-toggle.spec.ts
@@ -306,11 +306,11 @@ describe('MdButtonToggle', () => {
       buttonToggleNativeElements =
           buttonToggleDebugElements.map(debugEl => debugEl.nativeElement);
       buttonToggleInstances = buttonToggleDebugElements.map(debugEl => debugEl.componentInstance);
+
+      fixture.detectChanges();
     }));
 
     it('should update the model before firing change event', fakeAsync(() => {
-      fixture.detectChanges();
-
       expect(testComponent.modelValue).toBeUndefined();
       expect(testComponent.lastEvent).toBeUndefined();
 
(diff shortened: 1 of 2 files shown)


colinfang:
Just wondering, can we put `this._isInitialized = true` in `ngOnInit`? As the property binding are finished before entering `ngOnInit`.

DevVersion:
@colinfang Yeah that works too. I do not recall why we started using the `ngAfterViewInit` hook instead.

jelbourn:
@DevVersion @colinfang Using `ngOnInit` does not work because the order of initialization for directives on the same element is not defined. As such, an `ngModel` on the element may initialize *after* the `ngOnInit` of the component itself.

DevVersion:
@jelbourn Sounds reasonable. Reverted my commit about the `ngOnInit`

jelbourn:
LGTM

colinfang:
@jelbourn  Sorry I don't quite follow up what you mean, I have never made a `ngModel` for a custom component before. Could you please elaborate a little bit more?

Are you saying in 

     <my-component  [ngModel]="abc" ></my-component>

If `ngModel` is mapped to a property say  `Output() myInput`, in `MyComponent.ngOnInit` `myInput` is not defined?

jelbourn:
@colinfang when you have an `ngModel`, _it_ can be responsible for initializing the component, so you want to set `isInitialized = true` _after_ `ngModel` has run, which can be after the component's `onInit`.

PR #1149, opened 2016-09-01
chore(release): update versions for RC.6

kara:


hansl:
The system-config used in our tests probably need to be updated as well.

src/lib/button-toggle/package.json
```diff
@@ -24,6 +24,6 @@
   "homepage": "https://github.com/angular/material2#readme",
   "peerDependencies": {
     "@angular2-material/core": "2.0.0-alpha.7-12",
```
jelbourn: Need to update the version to alpha 8

jelbourn: Also need to update the change log. Those two things can happen in a separate pr, though.

kara: Yeah, I'll include in a separate PR

kara:
Got a verbal LGTM from Jeremy

PR #1154, opened 2016-09-01
chore(release): update versions for alpha.8

kara:


fxck:
Wouldn't the need of `forRoot()` on some components be worth mentioning? 

https://github.com/angular/material2/blob/9ff619669ae4e16409772b96e9f8d98f8477a3e0/src/lib/all/all.ts#L73

kara:
@fxck Modules IDs are stripped during the release build process.  We need to keep them in the repo for internal syncing reasons, but they're not part of the NPM package.

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
+<a name="2.0.0-alpha.8"></a>
+# [2.0.0-alpha.8 ectoplasm-helicopter](https://github.com/angular/material2/compare/g3_v2_0...2.0.0-alpha.8) (2016-09-01)
+
+### BREAKING CHANGES
+
+* all: we've updated our packaging to match angular/angular's packaging.  If you're using System.js in your project, you will probably want to
```
jelbourn: `System.js` > `SystemJS`

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
...
+```ts
+'@angular2-material/core': {
+    format: 'cjs',
+    main: 'core.umd.js'
+  }
+```
+
+You can see a more detailed example in our demo app's system config [here](https://github.com/angular/material2/blob/master/src/demo-app/system-config.ts).
+
+* all: material modules must be included with `forRoot()` now so that they work properly with lazy loading:
```
jelbourn: This should say "When bootstrapping". 

For lazily-loaded modules, you would not use `forRoot`

jelbourn:
LGTM aside from those two minor comments

CHANGELOG.md
kara: Oh, I guess that's confusing. I meant we need both options so lazy loading will work *without*. Probably best to leave that out.

codespicelabs:
With webpack alpha.8 (https://www.npmjs.com/~angular2-material) fails to this:


> WARNING in ./~/@angular2-material/button-toggle/button-toggle.js
> 15:24 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
> 

> WARNING in ./~/@angular2-material/button/button.js
> 17:24 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

> WARNING in ./~/@angular2-material/checkbox/checkbox.js
> 12:24 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

... and for all the others MD2 components

kara:
@codespicelabs  see https://github.com/angular/material2/pull/1168

PR #1166, opened 2016-09-02
fix(all): add forRoot() to all modules

kara:
This PR adds the `forRoot()` option to all Material modules, so users don't have to guess which modules have providers (and require the call) and which modules don't. This will allow us to add providers to components later without breaking anyone's code.


jelbourn:
LGTM

PR #1185, opened 2016-09-06
feat(slide-toggle): forward focus and blur event

DevVersion:
This is an older draft and could resolve the focus / blur event issue for components with an underlaying input element.

> It is also possible to use a normal Angular '@Output` instead.


DevVersion:
I don't think that's the correct approach, so closing this one.

PR #1188, opened 2016-09-07
fix(sidenav): fix Promise type to match

kara:
This PR fixes TS error that `Promise<{}>` can't be assigned to` Promise<void>`. 

r: @hansl


hansl:
LGTM.

PR #1201, opened 2016-09-08
refactor(all): rename the prefix from Md to Mat.

hansl:
A bit of explanation:

Because Material 1 already uses `md` as their prefix, and also because we want people to be able to use Material 1 and 2 in their apps (in an ng-upgrade kind of situation), we needed to change our prefix to something else.

After a long list of prefixes was considered (including the much loved but probably trademarked `mc`), `mat` was decided as being the best.

This is the renaming PR which moved all prefixes to use Mat. We also understand that this will break your apps. Considering this is alpha, we decided it was better to just rip the bandaid as soon as possible so people would have to change now instead of when we move to beta.

@kara @jelbourn Please prioritize this PR as it will likely conflict with everything if I need to rebase it.

### DIFF
--- a/src/demo-app/button/button-demo.html
+++ b/src/demo-app/button/button-demo.html
@@ -1,92 +1,92 @@
 <div class="demo-button">
   <section>
-    <button md-button>flat</button>
-    <button md-raised-button>raised</button>
-    <button md-fab>
-      <md-icon class="md-24">check</md-icon>
+    <button mat-button>flat</button>
+    <button mat-raised-button>raised</button>
+    <button mat-fab>
+      <mat-icon class="mat-24">check</mat-icon>
     </button>
-    <button md-mini-fab>
-      <md-icon class="md-24">check</md-icon>
+    <button mat-mini-fab>
+      <mat-icon class="mat-24">check</mat-icon>
     </button>
   </section>
 
   <section>
-    <a href="http://www.google.com" md-button color="primary">SEARCH</a>
-    <a href="http://www.google.com" md-raised-button>SEARCH</a>
-    <a href="http://www.google.com" md-fab>
-      <md-icon class="md-24">check</md-icon>
+    <a href="http://www.google.com" mat-button color="primary">SEARCH</a>
+    <a href="http://www.google.com" mat-raised-button>SEARCH</a>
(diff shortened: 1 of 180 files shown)


src/demo-app/button/button-demo.html
```diff
@@ -1,92 +1,92 @@
 <div class="demo-button">
   <section>
-    <button md-button>flat</button>
-    <button md-raised-button>raised</button>
-    <button md-fab>
-      <md-icon class="md-24">check</md-icon>
+    <button mat-button>flat</button>
+    <button mat-raised-button>raised</button>
+    <button mat-fab>
+      <mat-icon class="mat-24">check</mat-icon>
```
kara: I think this is actually part of the Material icon css file, so we shouldn't change it.

jelbourn:
LGTM

src/demo-app/card/card-demo.html
```diff
@@ -1,58 +1,58 @@
...
-        <md-card-title>Card with title</md-card-title>
-        <md-card-subtitle>Subtitle</md-card-subtitle>
-        <img md-card-md-image>
-    </md-card-title-group>
-  </md-card>
+  <mat-card>
+    <mat-card-title-group>
+        <mat-card-title>Card with title</mat-card-title>
+        <mat-card-subtitle>Subtitle</mat-card-subtitle>
+        <img mat-card-mat-image>
```
kara: This should be ``. The second `md` here stands for "medium"

src/demo-app/button/button-demo.html
hansl: Which part?  The `md-24`?

kara:
@hansl Found a few false positives that should be fixed before we merge

src/demo-app/card/card-demo.html
hansl: done.

src/demo-app/button/button-demo.html
hansl: done.

kara:
LGTM now

jelbourn:
Update for observers: we might not actually go through with this, exploring an alternative approach to prevent collisions.

zigzag95:
If not, this will be a big breaking change for me.

PR #1203, opened 2016-09-08
chore(forbidden-identifiers): check for deep imports and notify the user

DevVersion:
* No longer create the scope definitions for each line
* Resolve scope definitions and import path from "already" scoped imports to be able to check for deep imports (those are now invalid with the new distribution process)
* Also does no longer use files like `tsconfig.json` as package folders (minor issue in the glob)


jelbourn:
R: @hansl

DevVersion:
The check for the deep imports can be still valid, but other than that the scope check, which was updated here is not valid anymore.

PR #1209, opened 2016-09-09
fix(aot): make the demo app build with AoT compiler

alexeagle:


alexeagle:
Travis failures seem likely flakes?

jelbourn:
LGTM aside from those two minor things; failure is a flake (Edge started flaking recently)

alexeagle:
updated the internal bits

PR #1216, opened 2016-09-09
chore: reduce SauceLabs + Edge flakiness

jelbourn:
The changes in this PR _reduce_ the flakiness of Edge on SauceLabs but does not entirely eliminate it. 

* Remove unnecessary use of `async` and `fakeAsync`
* Inline resources for `gulp test:single-run` and `gulp e2e`
* Add `:watch:components:spec` task, and use that for `gulp test`, which fixes the watch behavior (but only if you're running only tests).

### DIFF
--- a/tools/gulp/tasks/unit-test.ts
+++ b/tools/gulp/tasks/unit-test.ts
@@ -6,7 +6,7 @@ import {PROJECT_ROOT, DIST_COMPONENTS_ROOT} from '../constants';
 import {sequenceTask} from '../task_helpers';
 
 const karma = require('karma');
-const inlineResources = require('../../../scripts/release/inline-resources');
+const runSequence = require('run-sequence');
 
 gulp.task(':build:test:vendor', function() {
   const npmVendorFiles = [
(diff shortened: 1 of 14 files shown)


tools/gulp/tasks/unit-test.ts
```diff
@@ -33,7 +37,10 @@ gulp.task('test', [':test:deps'], (done: () => void) => {
     configFile: path.join(PROJECT_ROOT, 'test/karma.conf.js')
   }, done).start();
 });
+
 gulp.task('test:single-run', [':test:deps'], (done: () => void) => {
+  inlineResources([DIST_COMPONENTS_ROOT]);
```
DevVersion: Should this be replaced with the new gulp task you created?

jelbourn:
R: @hansl

hansl:
LGTM, but please fix the nit pointed by @DevVersion

tools/gulp/tasks/unit-test.ts
jelbourn: Done.

PR #1218, opened 2016-09-09
chore: remove @param on setter, which Closure Compiler can't understand

jelbourn:
R: @kara 

@alexeagle needs this; will eventually be resolved by a fix in tsickle


alexeagle:
We may not need it, Evan is syncing tsickle fix into google3 as cl/132718669

jelbourn:
Closing since it seems to no longer be necessary

PR #1222, opened 2016-09-10
chore(): add gulp task to create a systemjs bundle

DevVersion:
- Gulp task to create a bundle, which includes all components and all Angular 2 packages
- Designed to run inside of the Plunker Template
- Task to deploy to firebase (as per #913)

Ping @hansl 

Closes #913


DevVersion:
Rebased on top of the latest package restructuring changes :)

DevVersion:
Closing because we can use the UMD bundles.

> Unfortunately RxJS does not support those yet.

PR #1225, opened 2016-09-11
fix(AOT): Unmark Input fields as private

MikeRyan52:


src/lib/button/button.ts
```diff
@@ -34,7 +33,7 @@ import {MdRippleModule} from '@angular2-material/core';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class MdButton {
-  private _color: string;
+  @Input('color') _color: string;
```
jelbourn: This `@Input` should go on the getter for `color`

jelbourn:
Thanks for the fix! Just one minor comment

MikeRyan52:
@jelbourn Same for `sidenav.ts` ?

jelbourn:
@MikeRyan52 yes, though it looks like it was already in the wrong place, so that would prbably fix another bug as well.

MikeRyan52:
@jelbourn 👍 Fixed the location of both input decorators.

jelbourn:
LGTM

jelbourn:
Looks like that sidenav change actually started making some unit tests fail. Go ahead and move the `@Input` back to the `_opened` property and we'll resolve it later.

Issue #123, opened 2016-02-27
Provide a theming solution for apps consuming material


jelbourn:
Needs design doc

zoechi:
angular/angular#2567

francisoud:
@jelbourn about your comment https://github.com/angular/material2/issues/123#issue-136879577 what do you mean by _Needs design doc_
As an end user, I would expect theming to work kind of [angular1](https://material.angularjs.org/latest/Theming/03_configuring_a_theme)  

I would expect  something like this code in a `app/main.ts` since it's a central part of the application:
```
import { bootstrap }      from '@angular/platform-browser-dynamic';
import { mdThemingProvider } from '@angular2-material/core';

import { AppComponent }         from './app.component';
import { appRouterProviders }   from './app.routes';

bootstrap(AppComponent, [
  appRouterProviders,
  mdThemingProvider
]);
```

but in this case I don't know where to put this kind of code

```
mdThemingProvider.theme('default')
    .primaryPalette('pink')
    .accentPalette('orange');
```

so may be another solution it to have 'main.ts' deleguate to another provider like this:
```
import { bootstrap }      from '@angular/platform-browser-dynamic';

import { AppComponent }         from './app.component';
import { appRouterProviders }   from './app.routes';
import { appThemingProvider }   from './app.theme';

bootstrap(AppComponent, [
  appRouterProviders,
  appThemingProvider
]);
```
and a `app.theme.ts`with:
```
import { theme } from '@angular2-material/core';

export const appThemingProvider = [
    theme('default')
        .primaryPalette('pink')
        .accentPalette('orange');
];
```

a alternative would be to reproduce what is done un the [router doc](https://angular.io/docs/ts/latest/guide/router.html)
`app.theme.ts` 
```
import { provideTheme, ThemingConfig } from '@angular2-material/core';

const themes: ThemingConfig = [
  { id: 'default', primaryPalette: 'pink',  accentPalette: 'orange'},
  { ... },
  { ... }
];

export const appRouterProviders = [
  provideTheme(themes);
];

```

jelbourn:
The primary way of theming for material2 will be static. The run-time generation of styles is not suitable for most apps because it is rather expensive. That said, we will possibly also have an optional run-time configuration as well.

davidreher:
maybe it is a stupid question, but can theming be done today at all?

jelbourn:
Not easily until we finish implementing user-facing theming.

a5hik:
[...]

PR #1239, opened 2016-09-14
feat(theming): add initial version for user-facing theming

jelbourn:
R: @hansl @kara 
#123 

Apologies for not breaking up this PR into smaller pieces, the changes snowballed on me.

### In this PR:
* Split theme-dependent styles for each component (and core) into their own file prefixed with `-theme`.
* Create an `md-light-theme` and `md-dark-theme` functions that outputs a "theme object", which is a sass map containing all of the theme's palettes.
* Move the contrast palettes into properties on the main palettes so that the end-user never has to worry about passing the palettes around.
* Add a few pre-built themes to start off.
* Reorganize core scss files into the appropriate directories and correct import paths everywhere.
* Change `core.scss` to `_core.scss`. The practical effect of this is that users either use one of the prebuilt theme css files OR, in defining their custom theme, `@include` the `md-core()` mixin (once). With either approach, the user ends up only importing one css file for material in their app.
* Change scss comments from `/** */` to `//`. I made this change so that users who consume our mixins do not end up with a bunch of random comments in their css output.
* Remove the grab-bag `_mixins.scss` and move its contents to more specific files.
* Change `md-visually-hidden` from a mixin to a single class and apply that class to use-sites.
* Change md-input to ViewEncapsulation.None. Without this, the styles were broken (have been for a while, since vicb's change to how Angular's css shimming works).
* Removed the "empty block" styleline rule in order to allow empty scss mixins (such as for grid-list and icon, which we have for consistency).

### Coming in follow-up PRs:
* Documentation on how to use theming
* A fix for applying a nested theme to the contents of an overlay.
* Add more pre-built themes.


src/lib/button/_button-base.scss
```diff
@@ -1,11 +1,25 @@
...
+@import '../core/style/elevation';
+@import '../core/style/button-common';
+
+
+// Applies a focus style to an md-button element.
+@mixin md-button-focus {
+  &::after {
+    // The button spec calls for focus on raised buttons (and FABs) to be indicated with a
+    // black, 12% opacity shade over the normal color (for both light and dark themes).
+    // We do this by placing an :after psuedo-element with the appropriate shade over the button.
```
kara: Typo: psuedo -> pseudo

src/lib/core/theming/_palette.scss
```diff
@@ -1,5 +1,13 @@
 // Color palettes from the Material Design spec.
 // See https://www.google.com/design/spec/style/color.html
+//
+// Contrast colors are hard-coded because it is too difficult (probably impossible) to
+// calculate them. These contrast colors are pulled from the public Material Design spec swatches.
+// While the contrast colors in the spec are not perscriptive, we use them for convenience.
```
kara: Typo: perscriptive -> prescriptive

PR #1240, opened 2016-09-15
chore(select): create select scaffold

kara:
Simply creates the scaffold that `md-select` will be built on.  `md-select` has no functionality yet.

r: @jelbourn.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1,0 +1,23 @@
+import {TestBed, async} from '@angular/core/testing';
+import {Component} from '@angular/core';
+import {MdSelectModule} from './index';
+
+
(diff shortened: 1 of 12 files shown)


zigzag95:
It will be great to have Readme.md committed while scaffolding.

zigzag95:
I know it's early but are you going to implement search and multi directives on md-select?

src/lib/select/select.spec.ts
```diff
@@ -0,0 +1,24 @@
+import {TestBed, async} from '@angular/core/testing';
+import {Component} from '@angular/core';
+import {MdSelectModule} from './index';
+
+
+describe('MdSelect', () => {
+
```
jelbourn: Indent is off in this file

PR #1241, opened 2016-09-15
Fix: Input should be disabled when disabling the formcontrol

### LINKED ISSUE #1171: [input @alpha.8-1] Setting bound FormControl to disabled causes valueAccessor.setDisabledState error

Bug

```
```

With

```
```

The error will go away

Sabartius:
Fixes #1171 for inputs.


Sabartius:
I signed it!

Sabartius:
I signed it!

PR #1242, opened 2016-09-15
Fix: Input/Radio should be disabled when disabling the formcontrol

### LINKED ISSUE #1171: [input @alpha.8-1] Setting bound FormControl to disabled causes valueAccessor.setDisabledState error

Sabartius:
Fixes #1171 for Input-Elements and Radio

@jelbourn @kara I've tried to repair the checkbox too, but i couldn't test it properly.
My test for the checkbox only worked, when i deleted the ChangeDetectionStrategy.OnPush from the Component, but thats not the solution i wanted to push.

### DIFF
--- a/src/lib/input/input.spec.ts
+++ b/src/lib/input/input.spec.ts
@@ -1,9 +1,16 @@
 import {
-  async,
-  TestBed,
+    async,
+    TestBed,
+    ComponentFixture,
 } from '@angular/core/testing';
 import {Component} from '@angular/core';
-import {FormsModule} from '@angular/forms';
+import {
+    FormsModule,
+    FormGroup,
+    FormBuilder,
+    ReactiveFormsModule,
+    FormControl
+} from '@angular/forms';
 import {By} from '@angular/platform-browser';
 import {MdInput, MdInputModule} from './input';
 
@@ -14,7 +21,7 @@ function isInternetExplorer11() {
 describe('MdInput', function () {
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      imports: [MdInputModule.forRoot(), FormsModule],
+      imports: [MdInputModule.forRoot(), FormsModule, ReactiveFormsModule],
       declarations: [
         MdInputNumberTypeConservedTestComponent,
         MdInputPlaceholderRequiredTestComponent,
@@ -57,6 +64,7 @@ describe('MdInput', function () {
         MdInputTextTestController,
         MdInputPasswordTestController,
         MdInputNumberTestController,
+        MdInputReactiveFormsTestController,
       ],
     });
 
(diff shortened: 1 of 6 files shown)


Sabartius:
I tried the checkbox in an demo app and it disables & enables successfully. So it's only something with the change-detection in the testenvironment that's not working as expected.

For Radio-Buttons i couldn't even get the reactiveForms to work with them. Maybe you could give me a hint how i need to configure the radio-group to set the Value of the FormControl?

src/lib/input/input.spec.ts
```diff
@@ -608,6 +616,39 @@ describe('MdInput', function () {
...
+
+      formControl.disable();
+      fixture.detectChanges();
+
+      fixture.whenStable().then(() => {
+        // Temporary workaround, see https://github.com/angular/angular/issues/10148
+        fixture.detectChanges();
+        fixture.whenStable().then(() => {
+          expect(inputElement.disabled).toBeTruthy('InputElement should be disabled');
+          expect(mdInput.disabled).toBeTruthy('Component should be disabled');
```
jelbourn: Huh, I had no idea you could pass a message like that.

src/lib/radio/radio.spec.ts
```diff
@@ -424,6 +432,48 @@ describe('MdRadio', () => {
...
+    let fixture: ComponentFixture<RadioGroupWithReactiveForms>;
+    let formControl: FormControl;
+    let radioInputs: Array<MdRadioButton> = [];
+    let nativeRadioInputs: Array<HTMLInputElement> = [];
+
+    beforeEach(async(() => {
+      fixture = TestBed.createComponent(RadioGroupWithReactiveForms);
+      fixture.detectChanges();
+
+      fixture.whenStable().then(() => {
```
jelbourn: Why do you need a `whenStable` in the `beforeEach`?

src/lib/radio/radio.spec.ts
```diff
@@ -424,6 +432,48 @@ describe('MdRadio', () => {
...
+        // Temporary workaround, see https://github.com/angular/angular/issues/10148
+        fixture.detectChanges();
+        fixture.whenStable().then(() => {
+          radioInputs.forEach(input =>
+              expect(input.disabled).toBeTruthy(input.value + ' should be disabled'));
+          nativeRadioInputs.forEach(input =>
+              expect(input.disabled).toBeTruthy('Native ' + input.id + ' should be disabled'));
```
jelbourn: You can use template strings for these messages

jelbourn:
@kara should take a look

src/lib/input/input.spec.ts
Sabartius: It helps alot when looking for errors :)

src/lib/radio/radio.spec.ts
Sabartius: Didn't need that one, got removed in the next commit

src/lib/radio/radio.spec.ts
Sabartius: Done

Sabartius:
@jelbourn @kara what's your opinion on the checkbox? I can push a solution, even a test for the checkbox, but because of the changedetection-strategy of the checkbox-component i cannot test if the underlying input-element gets disabled, only that the component gets disabled.

jelbourn:
@Sabartius can you push the change for checkbox including the failing test? We can take a look at see what needs to be changed.

Sabartius:
@jelbourn Pushed it. Hope you can find the problem :)

jelbourn:
@tinayuangao can you follow up on this?

jelbourn:
Superseded by #1750

PR #1245, opened 2016-09-15
fix(sidenav): move @Input to getter/setter instead of private field

alxhub:
This breaks AOT compile since the generated factory tries to set the private field.

I'm unsure of the ramifications of change detection setting `opened` via the setter, which will call `toggle()` where it didn't before - someone who understands the code should look at that.


jelbourn:
Closing this since the fix is already captured in #1225

PR #1246, opened 2016-09-15
docs(theming): add initial guide for theming.

jelbourn:
R: @kara 

For merging after #1239


docs/theming.md
```diff
@@ -0,0 +1,126 @@
...
+
+### What is a theme?
+A **theme** is the set of colors that will be applied to the Angular Material components. The
+library's approach to theming is based on the guidance from the [Material Design spec][1]. 
+
+In Angular Material, a theme is created by composing multiple palettes. In particular, 
+a theme consists of:
+* A primary palette: colors most widely used across all screens and components.
+* An accent palette: colors used for the floating action button and interactive elements.
+* A "warn" palette: colors used to convey error state.
```
kara: Curious - why is this the only one with quotes?

docs/theming.md
```diff
@@ -0,0 +1,126 @@
...
+// Plus includes for other components in your app.
+```
+
+You only need this single Sass file; you do not need to use Sass to style the rest of your app.
+
+If you are using the Angular CLI, support for compiling Sass to css is built-in; you only have to 
+add a new entry to the `"styles"` list in `angular-cli.json` pointing to the theme 
+file (e.g., `unicorn-app-theme.scss`).
+
+If you're not using the Angular CLI, you can use any existing sass tooling to build the file (such
```
kara: sass -> Sass (for consistency)

docs/theming.md
```diff
@@ -0,0 +1,126 @@
...
+You can use the `md-color` function to extract a specific color from a palette. For example:
+```scss
+.unicorn-carousel {
+  background-color: md-color($primary);
+  color: md-color($primary, default-contrast);
+}
+```
+
+### Future work
+* When the "all" package is release, there will be a mixin that captures all of the component's
```
kara: release -> released

docs/theming.md
```diff
@@ -0,0 +1,126 @@
...
+```css
+@import '~@angular2-material/core/theming/prebuilt/deeppurple-amber.css';
+```
+
+Alternatively, you can just reference the file directly. This would look something like
+```html
+<link href="node_modules/@angular2-material/core/theming/prebuilt/indigo-pink.css" rel="stylesheet">
+``` 
+
+You can also concatenate the file with the rest of your application's css.
```
kara: I think it might be handy to have a list of the existing pre-built themes here as bullet points.

docs/theming.md
jelbourn: "Primary palette" and "accent palette" both sounded natural, while "warn palette" seemed weird.

docs/theming.md
jelbourn: I'll add that once we have a full complement of prebuilts (only three right  now)

docs/theming.md
jelbourn: Done.

docs/theming.md
jelbourn: Done.

jimitndiaye:
Just gotta say, looking at this PR - this is awesome! Really like the use of mixins for encapsulating the theming logic. Will be poaching those ideas for other projects, @jelbourn .
Theming was my stumbling block for integrating material2 so anxiously waiting for this to be released in the next alpha. Well done!

PR #1256, opened 2016-09-17
latest cli build includes webpack

cre8:
`@webpack` isn't needed any more, the latest master build includes webpack.


jelbourn:
Fixed in #1257

Issue #1260, opened 2016-09-19
md-input-element color should be currentColor, to match a transparent background


jimitndiaye:
#### Bug, feature request, or proposal:
The default background is transparent. The default color should thus be current color in order for the input text to constrast properly. It is currently using the default value of ```initial``` but this causes problems when the background/foreground color changes. This is illustrated by @jelbourn 's https://github.com/jelbourn/material2-app (switch to dark mode and you'll see that the md-input's still have black text on a dark background)

#### What is the expected behavior?
Color should contrast correctly

#### What is the current behavior?
It retains the initial value and does not update dynamically.

#### What are the steps to reproduce?
Run https://github.com/jelbourn/material2-app and click on the **Toggle Dark Theme** button.

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: http://plnkr.co/edit/o077B6uEiiIgkC0S06dd

#### What is the use-case or motivation for changing an existing behavior?
Text should always be easily visible

jimitndiaye:
closed by #1262

PR #1262, opened 2016-09-19
fix(theming): set default color to currentColor in input theme (#1260)

jimitndiaye:
Currently input color does not update when switching between dark and light theme. Since the background of ```md-input-element``` is transparent it is currently possible to have black text on a dark background making it hard to read. This PR sets the color for ```.md-input-element``` to ```currentColor```. 

You can see the current behavior in @jelbourn 's https://github.com/jelbourn/material2-app (just toggle the dark theme).


jelbourn:
Thanks!

PR #1268, opened 2016-09-19
fix(slide-toggle): disabled theme not working and dragging works if disabled

DevVersion:
| Quick Summary |
| ------ |
| If checked and disabled theme is now invalid |
| Dragging while disabled |
| Improved dragging for UX | 

<details>
> * It seems like as per the new theming feature, view encapsulation turned off and now the checked theming overwrites the disabled theme (too high specificity)
> * If a slide-toggle is disabled, users are still able to drag the thumb (which is invalid)
> * Fix invalid `user-select` property, and now dragging works without clamps.
</details>

---
> Notice: We are two times extra checking for `isDragging`, that's because we want to keep the renderer as abstract as possible

FYI: I'm trying to get some good HammerJS tests into the slide-toggle in the future.

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.scss
+++ b/src/lib/slide-toggle/slide-toggle.scss
@@ -24,6 +24,12 @@ $md-slide-toggle-margin: 16px !default;
   line-height: $md-slide-toggle-height;
 
   white-space: nowrap;
+
+  // Disable user selection to ensure that dragging is smooth without grabbing some elements
+  // accidentally. Manually prefixing here, because the un-prefixed property is not supported yet.
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
   user-select: none;
 
   outline: none;
@@ -68,7 +74,6 @@ $md-slide-toggle-margin: 16px !default;
   height: $md-slide-toggle-height;
 
   position: relative;
-  user-select: none;
 
   margin-right: 8px;
 }
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -408,18 +408,11 @@ describe('MdSlideToggle', () => {
     let slideToggleControl: NgControl;
     let slideThumbContainer: HTMLElement;
 
-    // This initialization is async() because it needs to wait for ngModel to set the initial value.
     beforeEach(async(() => {
       fixture = TestBed.createComponent(SlideToggleTestApp);
 
       testComponent = fixture.debugElement.componentInstance;
 
-      // Enable jasmine spies on event functions, which may trigger at initialization
-      // of the slide-toggle component.
-      spyOn(fixture.debugElement.componentInstance, 'onSlideChange').and.callThrough();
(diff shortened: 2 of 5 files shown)


src/lib/slide-toggle/slide-toggle.scss
```diff
@@ -24,6 +24,12 @@ $md-slide-toggle-margin: 16px !default;
   line-height: $md-slide-toggle-height;
 
   white-space: nowrap;
+
+  // Disable user selection to ensure that dragging is smooth without grabbing some elements
+  // accidentally. Manually prefixing here, because the un-prefixed property is not supported yet.
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
```
jelbourn: Wouldn't autoprefixer do this?

DevVersion: Seems like there is no autoprefixer anymore, and `user-select` is also kind of a special property for the prefixer (e.g Material 1)

jelbourn:
Huh, autoprefixer must have been accidentally omitted from the build change. How is `user-select` special?

DevVersion:
I'm not really sure about it, If I recall I read a few things about it, e.g it does not get prefixed automatically. That's why I assumed that Material 1.x also does that. Might relate to the prefixer settings.

src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -214,16 +214,24 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
 
   /** TODO: internal */
   _onDragStart() {
-    this._slideRenderer.startThumbDrag(this.checked);
+    if (!this.disabled) {
+      this._slideRenderer.startThumbDrag(this.checked);
+    }
```
DevVersion: I actually planned to add some gesture tests in another PR (because of big changes), but I will try to add them to this PR now :)

DevVersion:
@jelbourn Added the tests for the dragging functionality.

Please notice that I currently use the Gesture Config from the slider package. I would rather move this config to a more generic place in another PR, because this will bloat the Diff.

src/lib/slide-toggle/slide-toggle.spec.ts
```diff
@@ -392,6 +398,96 @@ describe('MdSlideToggle', () => {
...
+
+    it('should drag from start to end', fakeAsync(() => {
+      expect(slideToggle.checked).toBe(false);
+
+      gestureConfig.emitEventForElement('slidestart', slideThumbContainer);
+
+      expect(slideThumbContainer.classList).toContain('md-dragging');
+
+      gestureConfig.emitEventForElement('slide', slideThumbContainer, {
+        deltaX: 200 // Use a random number which will be clamped.
```
jelbourn:
```ts
// Arbitrary, large delta that will be clamped to the end of the slide-toggle.
```

jelbourn:
LGTM aside from some comment wording

src/lib/slide-toggle/slide-toggle.spec.ts
DevVersion: Yeah, sounds better. Just changed it for all.

Issue #546, opened 2016-05-27
md-textarea


fxck:
I would like to add `md-textarea`. From what I've tried(c&p md-input code, changing selector and template), it should be pretty straightforward, as it just worked without any changes to style or the component itself.

The differences between input and textarea are

- textarea doesn't have `type`, `step`, `list` attributes 
- textarea might have auto expand/size(probably not needed in the first iteration)

Since 90% of the component is made of annotations, and those cannot be inherited, the way to go(I think) might be either to extend selector

```
selector: 'md-input,md-textarea',
```

and use `ElementRef` to check which selector matched(unless there's a better way I don't know of) and `*ngIf`the template to either `<input>` or `<textarea>`. 

or keep only one selector(that is `md-input`), and add `md-textarea` as attribute, and `*ngIf` the template depending on that.

In both cases I'd be needed add check that `type` was not added along `<md-textarea>` or `md-textarea`, because of `_convertValueForInputType`, there won't be no need to do anything with `step` or `list`, since those simply won't be on the `*ngIf`'d template.

Thoughts?

---

as for autosize, there's this tiny thing https://github.com/stevepapa/angular2-autosize 

```javascript
import { ElementRef, HostListener, Directive} from 'angular2/core';

@Directive({
    selector: 'textarea[autosize]'
})

export class Autosize {
 @HostListener('input',['$event.target'])
  onInput(textArea) {
    this.adjust();
  }
  constructor(public element: ElementRef){
  }
  ngOnInit(){
    this.adjust();
  }
  adjust(){
    this.element.nativeElement.style.overflow = 'hidden';
    this.element.nativeElement.style.height = 'auto';
    this.element.nativeElement.style.height = this.element.nativeElement.scrollHeight + "px";
  }
}
```

which utilizes `scrollHeight`, which is apparently supported by most browsers https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight

jelbourn:
@hansl did you have any existing plans WRT `textarea` implementation details?

I'm thinking it might be good to have a `MdTextField` base class that both `MdInput` and `MdTextArea` extend, mainly so they can set up different bindings. We could possibly also introduce an internal directive like `MdTextFieldWrapper` that has all the wrapper elements and styles simply content-projects either the input element or textarea element.

The autosize would be a good follow-up. It would be good to see a simple proof-of-concept on the `scrollHeight` technique, which would be pretty neat.

fxck:
[...]

PR #1282, opened 2016-09-20
chore(textarea): create textarea

dharmeshpipariya:
creates the md-textarea feature [546](https://github.com/angular/material2/issues/546)


jelbourn:
@dharmeshpipariya it will probably be a bit before we can review this; we're all tied up preparing for Angular Connect next week.

dharmeshpipariya:
CLA signed.

jelbourn:
@dharmeshpipariya It looks like this is mostly copy-and-paste from `md-input` without any refactoring to reuse common parts between them. This kind of approach has a higher cost in the long run because you're now maintaining two sets of code that do the same thing. The two components should share code, potentially even being the same component with some conditionals. I'm going to close this PR since, at a high level, it's not really the way we want to approach the implementation. If you want to try a different approach, comment on #546 with your plan and we can discuss further.

PR #1285, opened 2016-09-20
build: do not inline resources for normal builds

DevVersion:
Inlining the resources for normal builds does not refresh the view.

> For example if I change a HTML file, it will be properly copied over (per assets gulp task), but it's still inlined wrong and does not update the view.

<details>
> By default all build tasks should not inline the resources, and only the build tasks (like with the ng compiler) should inline the resources.
</details>

I can imagine that it is intentionally to **always** inline the resources, there were the following approaches
- Always re-inline the resources (not cool for a dev app IMO)
- Only inline the resources if explicitly requested (release, ngc)

I'm up for some other approaches or requested changes.

### DIFF
--- a/tools/gulp/task_helpers.ts
+++ b/tools/gulp/task_helpers.ts
@@ -147,14 +147,18 @@ export function cleanTask(glob: string) {
 
 
 /** Build an task that depends on all application build tasks. */
-export function buildAppTask(appName: string) {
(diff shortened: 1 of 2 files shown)


DevVersion:
Closing this, since this PR is no longer valid after b5c8c63a27fb0e620d4f48108a4a9d55fc88c2ca.

PR #1286, opened 2016-09-20
feat(packaging): move packaging to a single "@angular/material"

hansl:
Moving forward, we will only publish a single `@angular/material` package for the Angular2 Material project. This simplifies installation and usage for everyone, and won't increase the size of the library for people using AoT. It is also assumed that most people who use Angular Material will use most of the components so using a single UMD bundle should not increase their package size too much in JIT.

# BREAKING CHANGES

People using `npm install @angular2-material/COMPONENT_NAME` need to remove all their `@angular2-material` dependencies and use a single `@angular/material` dependency instead.


tools/gulp/tasks/components.ts
```diff
@@ -63,48 +53,38 @@ task(':build:components:rollup', [':build:components:ts'], () => {
...
-        const result = bundle.generate({
-          moduleName: `md.${camelCase(name)}`,
-          format: 'umd',
-          globals
-        });
-        const outputPath = path.join(DIST_COMPONENTS_ROOT, name, `${name}.umd.js`);
-        writeFileSync( outputPath, result.code );
-      });
-  }, Promise.resolve());
+  // Build the all package.
```
jelbourn: Comment something like "Use rollup to create the published umd bundle." ?

kyleraymorgan:
When should we start using `npm install @angular/material`?

src/demo-app/system-config.ts
```diff
@@ -78,27 +25,40 @@ const barrels: string[] = [
   'live-announcer',
   'portal',
   'overlay',
-  ...components
-  /** @cli-barrel */
 ];
 
-const _cliSystemConfig: any = angularPackages;
+
+const angularPackages: any = angularDeps.reduce((config: any, pkg: string) => {
```
hansl: System-Config has been simplified. PTAL

src/demo-app/system-config.ts
```diff
@@ -78,27 +25,40 @@ const barrels: string[] = [
...
+  config[pkg] = { main: `bundles/${pkg.replace('@angular/', '')}.umd.js` };
+  return config;
+}, {});
 barrels.forEach((barrelName: string) => {
-  _cliSystemConfig[barrelName] = { main: 'index' };
+  angularPackages[barrelName] = { main: 'index' };
 });
 
+const angularMaps: any = angularDeps.reduce((config: any, pkg: string) => {
```
hansl: System-Config has been simplified. PTAL

tools/gulp/tasks/components.ts
hansl: Done.

hansl:
@jelbourn: this is ready to be reviewed.

Sourcemaps are still messy (they give the `tsc` output), but for now this will unblock you for AC. I'll improve in a subsequent PR.

src/lib/core/theming/_all-theme.scss
```diff
@@ -1,23 +1,23 @@
...
+@import '../../menu/menu-theme';
+@import '../../progress-bar/progress-bar-theme';
+@import '../../progress-circle/progress-circle-theme';
+@import '../../radio/radio-theme';
+@import '../../sidenav/sidenav-theme';
+@import '../../slide-toggle/slide-toggle-theme';
+@import '../../slider/slider-theme';
+@import '../../tabs/tabs-theme';
+@import '../../toolbar/toolbar-theme';
+@import '../../tooltip/tooltip-theme';
```
jelbourn: Want to move this mixin into `_theming.scss` and just delete this file?

hansl:
Okay. Will do that when I'm back from dinner

src/lib/core/theming/_all-theme.scss
hansl: Done.

src/lib/core/theming/_theming.scss
```diff
@@ -82,3 +82,51 @@
...
+@import '../../sidenav/sidenav-theme';
+@import '../../slide-toggle/slide-toggle-theme';
+@import '../../slider/slider-theme';
+@import '../../tabs/tabs-theme';
+@import '../../toolbar/toolbar-theme';
+@import '../../tooltip/tooltip-theme';
+
+
+// Create a theme.
+@mixin angular-material-theme($theme) {
```
jelbourn: As discussed: internal presubmit revealed that moving this here caused an unresolvable circular dependency. Need to move it back to it's own file. I suggest we call it `_all-component-themes.scss`

hansl: Dono

PR #1288, opened 2016-09-21
feat(a11y): initial interactivity checker

jelbourn:
R: @hansl 

This is the first step towards adding a general solution focus trapping, with the first goal of supporting MdDialog. I drew a lot of knowledge from [ally.js](https://allyjs.io).

Current checks are basic compared to the robustness of ally.js, but should capture most common use cases.


jelbourn:
I'll add an export when I start using it elsewhere. Looks like there's a legit failure on IE11 and Edge, though.

jelbourn:
@hansl Fixed IE and Edge problem, any further comments?

PR #1297, opened 2016-09-21
fix(sidenav): align text at start

kara:
End sidenavs were aligning text at "end" because the `align="end"` attribute was still on the DOM. This PR removes the attribute so it won't confuse browsers.


src/lib/sidenav/sidenav.ts
```diff
@@ -38,6 +38,7 @@ export class MdDuplicatedSidenavError extends MdError {
   template: '<ng-content></ng-content>',
   host: {
     '(transitionend)': '_onTransitionEnd($event)',
+    '[attr.align]': 'null'
```
jelbourn: Needs comment explaining why this is here.

jelbourn:
LGTM

PR #1303, opened 2016-09-22
fix(slider): correct click/slide event position

### LINKED ISSUE #1234: md-slider not working inside md-sidenav
When md-slider is located inside a md-sidenav it works fine if the mode is set to side. If the sidenav mode is push the slider stops working. When touching the slider handler the slider handler jumps to 0 and is stuck there. This is strange because it works fine if sidenav mode is set to side. Maybe it has something to do with the animation or the hidden state of the sidenav when in push mode. Any idea how to fix this?

*Update it looks like the slider touch detection area is not animated along with the slider itself when the sidenav appears. If I drag the slider handler and the mouse is in the area where the sidenav would be hidden it will start to move the slider handlebar.

jmcgoldrick:
fixes #1234

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -182,6 +182,7 @@ export class MdSlider implements AfterContentInit, ControlValueAccessor {
 
     this.isActive = true;
     this.isSliding = false;
+    this._sliderDimensions = this._renderer.getSliderDimensions();
     this._renderer.addFocus();
     this.updateValueFromPosition(event.clientX);
     this.snapThumbToValue();
@@ -207,6 +208,7 @@ export class MdSlider implements AfterContentInit, ControlValueAccessor {
     event.preventDefault();
     this.isSliding = true;
     this.isActive = true;
+    this._sliderDimensions = this._renderer.getSliderDimensions();
     this._renderer.addFocus();
     this.updateValueFromPosition(event.center.x);
   }


jmcgoldrick:
The slider's dimensions were only being updated when created/resized. If the sliders position changes outside of these conditions the calculations for the thumb etc. were out.

It appears you only need one of the updates (either in `onClick` or `onSlideStart`) but I couldn't figure out why, also for consistency it seemed better to put it in both methods.

Is there a better way to account for position changes outside of the elements control?

src/lib/slider/slider.ts
```diff
@@ -182,6 +182,7 @@ export class MdSlider implements AfterContentInit, ControlValueAccessor {
 
     this.isActive = true;
     this.isSliding = false;
+    this._sliderDimensions = this._renderer.getSliderDimensions();
```
jelbourn: If this is only necessary in one handler, I would just leave it in the click handler. Also needs a comment explaining why the dimensions need to be recalculated here.

jmcgoldrick: Hi @jelbourn, two things:

- to make the unit tests pass, `_sliderDimensions` needs to be updated in both `onClick` and `onSlideStart`, though it may be because the tests generate the click/slide events explicitly.
- initially I couldn't write a **failing** test, I believe it is due to the way the click/slide events are generated in the tests.

The click/slide events in the unit tests take a "percentage", grab the **up-to-date** dimensions of the element and calculate the event coordinates based on that, which is backwards to how the browser does it. This means the values are always "correct" even if you move the sliders' parent around. See: https://github.com/angular/material2/blob/master/src/lib/slider/slider.spec.ts#L688

Adjusting these methods to take (pre-calculated) coordinates of an event, tests fail/pass as expected. Am I missing something obvious though...?

chouclee:
This PR also fixes #1389 
The problem of #1389 is because when calling ngAfterContentInit(), _sliderDimensions would be an empty rectangle area (couldn't figure out why). Resizing the window would call this._renderer.getSliderDimensions() again to update dimensions.

mmalerba:
This partially fixes the issue, but if the slider has an initial value it still won't show up properly. #1663 refactors this component to use percentage-based measurements, rather than pixel based ones for the thumb position which should catch this last edge case.

jelbourn:
Superseded #1663

PR #1311, opened 2016-09-23
feat(a11y): add basic focus-trap directive

jelbourn:
R: @kara 

Very basic approach, as described in the comments. 

The template isn't in its own file for now because apparently our build isn't set up to handle `templateUrl` for anything in `core/`.


jelbourn:
Also R: @hansl

src/lib/core/a11y/focus-trap.ts
```diff
@@ -0,0 +1,78 @@
+import {Component, ViewEncapsulation, ViewChild, ElementRef} from '@angular/core';
+import {InteractivityChecker} from './interactivity-checker';
+
+
+/**
+ * Directive for trapping focus within a region.
+ *
+ * NOTE: This directive currently uses a very simple (naive) approach to focus trapping.
+ * It assumes that the tab order is the same as DOM order, which is not necessarily true.
+ * Things like tabIndex > 0, flex `order`, and shadow roots can cause to two to misalign.
```
kara: `to two` -> `the two`

PR #1319, opened 2016-09-23
fix(sidenav): fix dark theme

kara:


src/lib/core/theming/_palette.scss
```diff
@@ -660,6 +661,7 @@ $md-dark-theme-background: (
   hover:      rgba(white, 0.04), // TODO(kara): check style with Material Design UX
   card:       map_get($md-grey, 800),
   dialog:     map_get($md-grey, 800),
+  sidenav:    #303030,
```
jelbourn: You should just use the existing `background` key. (that's where I got the #303030 from)

src/lib/sidenav/_sidenav-theme.scss
```diff
@@ -14,10 +14,17 @@
...
   $sidenav-push-background-color: md-color($background, dialog) !default;
 
+  md-sidenav-layout {
+    background-color: $sidenav-layout-background-color;
+    color: md-color($foreground, base);
+  }
+
   md-sidenav {
     background-color: $sidenav-background-color;
+    color: md-color($foreground, base);
```
jelbourn: I think `color` should just be `text`

jelbourn:
LGTM

PR #1326, opened 2016-09-25
style(card): add default outer-margin on mobile

### LINKED ISSUE #1325: Add margin to cards
Feature Request: Cards do have a margin of 8dp [according to Material Design spec](https://material.google.com/components/cards.html#cards-content-blocks)

Cards should have this margin automatically by themselves.

Cards do not have a margin.

dasois:
Previously cards did not have the outer margin, that is specified in the Material Design spec.
 Now cards have a 8px outer margin on top, the sides and in between of them.
 The outer margin does not apply to the bottom of the cards.

 Closes #1325

### DIFF
--- a/src/demo-app/card/card-demo.scss
+++ b/src/demo-app/card/card-demo.scss
@@ -2,9 +2,11 @@
   display: flex;
   flex-flow: column nowrap;
 
-  md-card {
-    margin: 0 16px 16px 0;
-    width: 350px;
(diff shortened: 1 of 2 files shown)


dasois:
I signed it now!

jelbourn:
We very intentionally do not add margins to any of our components in material2 because, in practice, this is always going to depend on the application.

Issue #1342, opened 2016-09-27
Alpha 9 AoT compiler error


spock123:
Hey guys,
- I updated to the new Alpha 9, upgraded to the single Material Module.
So far, things work great in development.

BUT: running the AoT compiler results in errors: (compiler-cli 0.6.2):
```

Error: parameters received {"filePath":"/Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/material/core/a11y/live-announcer.d.ts","name":"MdLiveAnnouncer"} which is not a StaticSymbol
    at StaticReflector.parameters (/Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler-cli/src/static_reflector.js:92:19)
    at CompileMetadataResolver.getDependenciesMetadata (/Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler/bundles/compiler.umd.js:14336:81)
    at CompileMetadataResolver.getTypeMetadata (/Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler/bundles/compiler.umd.js:14301:28)
    at /Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler/bundles/compiler.umd.js:14448:43
    at Array.forEach (native)
    at CompileMetadataResolver.getProvidersMetadata (/Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler/bundles/compiler.umd.js:14428:21)
    at /Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler/bundles/compiler.umd.js:14435:43
    at Array.forEach (native)
    at CompileMetadataResolver.getProvidersMetadata (/Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler/bundles/compiler.umd.js:14428:21)
    at /Users/lrj/projects/angular2-webpack2-typescript2/node_modules/@angular/compiler/bundles/compiler.umd.js:14120:73
Compilation failed
```

Anyone stumbled across this?

mmrath:
Same error for me. It would be nice to have AOT compile validation as a part of build if it not already there.

spock123:
@mmrath same thoughts here, I just wasn't sure I did something wrong 😎

Thanks for confirming u have the same issue

jelbourn:
[...]

PR #1368, opened 2016-09-29
fix(aot): add constant for a11y providers, use in core module

jelbourn:
With these changes I have AoT working on my material2-app: https://material2-app.firebaseapp.com/

Also bumps the version so I can publish the fix.

See #1342


jelbourn:
Wrong branch.

PR #1371, opened 2016-09-29
chore: fix CI flakiness caused by race condition

jelbourn:
The `:test:deps` task included both `:build:components:inline` and `:watch:components:spec`.  The latter use the CJS output (which we want for karma), but the former uses the ESM output (which we don't want). This led to a race condition where the tests only passed when the CJS output ended up being there when the tests started.

I also:
- Made sure that watch mode isn't used on CI
- Added a bunch of comments to the tasks


jelbourn:
R: @hansl

Issue #1382, opened 2016-09-30
Bug: MdSidenav [opened] causes unresolved event _openPromiseReject


dima-gusyatiner:
Bug when using MdSidenav open as default:
`<md-sidenav #demo opened="true">`

The browser claims of an error when trying to close it with 
`<button md-button (click)="demo.close() / toggle()">Close</button>`

```
core.umd.js:3427 EXCEPTION: Uncaught (in promise): undefinedErrorHandler.handleError @ core.umd.js:3427next @ core.umd.js:6901schedulerFn @ core.umd.js:6150SafeSubscriber.__tryOrUnsub @ Subscriber.ts:238SafeSubscriber.next @ Subscriber.ts:190Subscriber._next @ Subscriber.ts:135Subscriber.next @ Subscriber.ts:95Subject.next @ Subject.ts:61EventEmitter.emit @ core.umd.js:6142onError @ core.umd.js:6366onHandleError @ core.umd.js:6241ZoneDelegate.handleError @ zone.js:207Zone.runGuarded @ zone.js:113_loop_1 @ zone.js:379drainMicroTaskQueue @ zone.js:386ZoneTask.invoke @ zone.js:308
core.umd.js:3432 ORIGINAL STACKTRACE:ErrorHandler.handleError @ core.umd.js:3432next @ core.umd.js:6901schedulerFn @ core.umd.js:6150SafeSubscriber.__tryOrUnsub @ Subscriber.ts:238SafeSubscriber.next @ Subscriber.ts:190Subscriber._next @ Subscriber.ts:135Subscriber.next @ Subscriber.ts:95Subject.next @ Subject.ts:61EventEmitter.emit @ core.umd.js:6142onError @ core.umd.js:6366onHandleError @ core.umd.js:6241ZoneDelegate.handleError @ zone.js:207Zone.runGuarded @ zone.js:113_loop_1 @ zone.js:379drainMicroTaskQueue @ zone.js:386ZoneTask.invoke @ zone.js:308
core.umd.js:3433 Error: Uncaught (in promise): undefined
    at resolvePromise (zone.js:429)
    at MdSidenav.eval [as _openPromiseReject] (zone.js:406)
    at MdSidenav._onTransitionEnd (sidenav.js:164)
    at DebugAppView._View_AppComponent0._handle_transitionend_2_2 (AppComponent.ngfactory.js:732)
    at eval (core.umd.js:9678)
    at eval (platform-browser.umd.js:1813)
[...]

PR #1383, opened 2016-09-30
Fixes MdSidenav issue when closing a sidenav with [opened]="true"

dima-gusyatiner:
Fixes MdSidenav issue with sidenav opened as default.
Setting the attribute on the html component creates `_openPromise` via `toggle()`, which is never resolved, like it's usually done with with `_onTransitionEnd` (issue #1382).

The reason is when MdSidenav component is initialized:

```
@Input()
  get opened(): boolean { return this._opened; }
  set opened(v: boolean) {
    // TODO(jelbourn): this coercion goes away when BooleanFieldValue is removed.
    let booleanValue = v != null && `${v}` !== 'false';
    this.toggle(booleanValue);
  } 
```

`toggle()` method creates `_openPromiseReject` and `_openPromiseReject`:

```
    if (isOpen) {
      if (this._openPromise == null) {
        this._openPromise = new Promise<void>((resolve, reject) => {
          this._openPromiseResolve = resolve;
          this._openPromiseReject = reject;
        });
      }
      return this._openPromise;
    } else {
      if (this._closePromise == null) {
        this._closePromise = new Promise<void>((resolve, reject) => {
          this._closePromiseResolve = resolve;
          this._closePromiseReject = reject;
        });
      }
      return this._closePromise;
    }
```

They are usually resolved at `_onTransitionEnd`, but since no transition occurs on application start, this method never gets called. When we click the close button, it rejects the open promise.

```
  _onTransitionEnd(transitionEvent: TransitionEvent) {
    if (transitionEvent.target == this._elementRef.nativeElement
        // Simpler version to check for prefixes.
        && transitionEvent.propertyName.endsWith('transform')) {
      this._transition = false;
      if (this._opened) {
        if (this._openPromise != null) {
          this._openPromiseResolve();
        }
        if (this._closePromise != null) {
          this._closePromiseReject();
        }

        this.onOpen.emit(null);
      } else {
        if (this._closePromise != null) {
          this._closePromiseResolve();
        }
        if (this._openPromise != null) {
          this._openPromiseReject();
        }

        this.onClose.emit(null);
      }

      this._openPromise = null;
      this._closePromise = null;
    }
  }
```

My current solution is to clear the promises when setting the propery:

```
  @Input()
  get opened(): boolean { return this._opened; }
  set opened(v: boolean) {
    // TODO(jelbourn): this coercion goes away when BooleanFieldValue is removed.
    let booleanValue = v != null && `${v}` !== 'false';
    this.toggle(booleanValue);

    this._openPromise = null;
    this._closePromise = null;
  }
```

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -83,6 +83,9 @@ export class MdSidenav {
     // TODO(jelbourn): this coercion goes away when BooleanFieldValue is removed.
     let booleanValue = v != null && `${v}` !== 'false';
     this.toggle(booleanValue);
+
+    this._openPromise = null;
+    this._closePromise = null;
   }
 
 


dima-gusyatiner:
I signed in!

hansl:
This will introduce a bug and is unsafe. The bug is that when calling open twice the wrong promise will be returned, but never resolved.

I fixed this in a separate PR #1666 and added a  unit test for avoiding a future regression.

Thanks!

PR #1384, opened 2016-09-30
(doc) Update theming.md

vinagreti:
@jelbourn, I've used /src/lib/input/_input-theme.scss as base.


vinagreti:
There was some class name mistakes.

PR #1290, opened 2016-09-21
chore: remove BooleanFieldValue

jelbourn:
Based on top of #1289 

This removes the custom decorator as it prevents tree-shaking of components that use it. Replaced with a `coerceBooleanProperty` function.

PR #1402, opened 2016-10-02
fix(checkbox) should be disabled when *any* disabled value makes the component disabled, *except* for false.

dharmeshpipariya:
fix(checkbox) should be disabled when *any* disabled value makes the component disabled, *except* for false.


DevVersion:
I'm curious why this change is necessary when `@BooleanFieldValue` is already used.

The slide-toggle for example seems to work as expected
- http://plnkr.co/edit/o077B6uEiiIgkC0S06dd?p=preview

jelbourn:
`@BooleanFieldValue` is going away in #1290 (which needs rebase). Can you just leave me a comment there to add the fix to checkbox?

PR #1404, opened 2016-10-02
﻿fix(input): asterisk color when not focused

### LINKED ISSUE #1395: Misleading UX in <md-input>
The required mark (asterisk) should not take warn (red) color after entering valid input.

The required mark (asterisk) takes warn (red) color even after entering valid input, which is misleading, the users may think the entered data is invalid. 
This behaviour is observed in Chrome on Ubuntu 16.04 LTS.
See the screen shot bellow:

input

DevVersion:
* Currently the asterisk is still using the warn color when the input is not focused anymore.

* Per Material Specs the asterisk should be only red when the input is focused ([Specs Video](https://material-design.storage.googleapis.com/publish/material_v_9/0B7WCemMG6e0VcHFla0dwOE5pVVE/forms-supportive-content-v3_xhdpi.webm))

| Focused | Blurred |
| ------- | ---------- |
|  |  |
 

Fixes #1395.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -39,8 +39,7 @@
   }
 
   // See md-input-placeholder-floating mixin in input.scss
-  md-input input:-webkit-autofill + .md-input-placeholder,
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #1405, opened 2016-10-02
fix(slide-toggle): emit change event after drag end 

### LINKED ISSUE #1390: md-slide-toggle does not emit event when thumb is dragged
Bug.

When the user changes state by dragging the thumb, a change event should be emitted.

Change event is not emitted.

Please see this plunk: http://plnkr.co/edit/gMK2P5tCWyT4mszL08ih

Click on the switch to change the state. Note that we see a message indicating that a change event was received. Now drag the thumb from one side to the other. Note that the change event is not generated.

I believe we should follow the principle of "If the feature is there, it should work". If we're giving the user the capability of changing the state by dragging the thumb, then this action should take affect.
As an average user, I was surprised that it doesn't affect the state.

Angular 2.0 final.
@angular2-material/slide-toggle - v2.0.0-alpha.8-2 - Angular 2 Material Slide Toggle.
Server is running linux.
Browser is running on Windows 7, Chrome v53.0.2785.116 m

No.

DevVersion:
* Emits a change event after the drag completed.

Fixes #1390.

**FYI**: Rebased on top of #1268 (should be merged before)

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -494,7 +494,7 @@ describe('MdSlideToggle', () => {
       expect(slideThumbContainer.classList).toContain('md-dragging');
 
       gestureConfig.emitEventForElement('slide', slideThumbContainer, {
-        deltaX: 200 // Use a random number which will be clamped.
+        deltaX: 200 // Arbitrary, large delta that will be clamped to the end of the slide-toggle.
       });
 
       gestureConfig.emitEventForElement('slideend', slideThumbContainer);
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -486,6 +486,27 @@ describe('MdSlideToggle', () => {
       expect(slideThumbContainer.classList).not.toContain('md-dragging');
     }));
 
+    it('should should emit a change event after drag', fakeAsync(() => {
+      expect(slideToggle.checked).toBe(false);
+
+      gestureConfig.emitEventForElement('slidestart', slideThumbContainer);
+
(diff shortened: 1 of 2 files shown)


jelbourn:
@DevVersion can you rebase this now (to make review easier)?

DevVersion:
@jelbourn I just rebased on the #1268 changes again, but I assume you meant to rebase on `master`
>  (This would be a mess, because the testing structures are not available then)

https://github.com/angular/material2/pull/1405/commits/1e16eb2afe802e6882e8ce64d6af3b13aff4d594 actually is the only change (last commit in the PR)

jelbourn:
Whoops, I thought I had already merged #1268. Now I _really_ merged it.

DevVersion:
Just rebased now on top of `master`.

src/lib/slide-toggle/slide-toggle.spec.ts
```diff
@@ -486,6 +486,27 @@ describe('MdSlideToggle', () => {
...
 
+    it('should should emit a change event after drag', fakeAsync(() => {
+      expect(slideToggle.checked).toBe(false);
+
+      gestureConfig.emitEventForElement('slidestart', slideThumbContainer);
+
+      expect(slideThumbContainer.classList).toContain('md-dragging');
+
+      gestureConfig.emitEventForElement('slide', slideThumbContainer, {
+        deltaX: 200 // Use a random number which will be clamped.
```
jelbourn: Same comment on the comment as the other PR

jelbourn:
LGTM aside for that same comment

src/lib/slide-toggle/slide-toggle.spec.ts
DevVersion: Oh, I just rebased it, and didn't look at it again. Just made the change :)

PR #1413, opened 2016-10-03
fix(snack-bar): add explicit box-sizing

### LINKED ISSUE #1412: snackbar is unreadable when using border-box: box-sizing
I'm setting border-box to `box-sizing` globally(pretty [standard](https://www.paulirish.com/2012/box-sizing-border-box-ftw/) thing nowadays) and snackbar apparently requires `content-box`, I guess it should be explicitly defined in its styles then.

Otherwise it looks like this

image 

and is hardly readable.

crisbeto:
Fixes the snack bar not being readable in the cases where `box-sizing: border-box` is applied globally.

Fixes #1412.

**Note:** this can also be fixed by removing the explicit height, however I wasn't sure whether it wasn't there on purpose (e.g. to prevent line wraps).

### DIFF
--- a/src/lib/snack-bar/snack-bar-container.scss
+++ b/src/lib/snack-bar/snack-bar-container.scss
@@ -16,4 +16,5 @@ $md-snack-bar-max-width: 568px !default;
   min-width: $md-snack-bar-min-width;
   overflow: hidden;
   padding: $md-snack-bar-padding;
+  box-sizing: content-box;
 }


fxck:
Removing explicit height messes up alignment. But honestly, `box-sizing: border-box` should be explicitly set for all of the components and they should be styled with that in mind. Rather than having different box sizing for each component.

crisbeto:
Hopefully this wouldn't be too much of a problem. In Material 1, there's only 2-3 cases total where the box sizing needs to be reset, I don't expect it to be much more for Material 2. This can also be fixed by removing the padding, but then the vertical alignment is a little trickier since it'd have to use `line-height` which isn't very consistent across browsers.

jelbourn:
LGTM

Issue #1432, opened 2016-10-05
ability to manually control overlay's z-index 


fxck:
This would be handy for both user created overlays, as well as those used by angular components such as md-menu.

Take this for example http://plnkr.co/edit/AqjYdsScCYPlK9w44aHQ?p=preview you have a fixed element(some sort custom toolbar), inside it you have a button which triggers md-menu, it works fine, but the `md-menu-click-catcher`, which is for some reason inside the actual `md-menu` element, has is actually stacked above `md-overlay-container`, causing any attempt to click on the menu to close it.

1. `md-overlay-container` should *not* have z-index at all, as it renders z-indexes of `md-overlay-pane` useless
2. you should be able to manually control z-index of `md-overlay-pane`

zfanning:
+1

meowhunter:
Here is a good example why cdk-overlay-container should not have z-index

![z-index](https://user-images.githubusercontent.com/5663121/61463456-0d42d280-a97d-11e9-8fc6-927212626bc4.jpg)

am-awais:
Any updates??

am-awais:
I sort it out myself by using two stylessheet one `Global `and other `component's `stylesheet, In global i set `z-index` to to lower value(`1000`) so that it goes behind the `header` and in `popup` `component` styles sheet i set that to high value(`2000`) so that `header` goes behind my overlay. That's how i manage to solve it 
Thank me later

aksenovdev:
Maybe it help
https://blog.bbogdanov.net/post/angular-material-overlay-hack-overlay-container/

4javier:
Any news? There are some situations where somebody want his overlay to overlap even a fixed header. Think about tooltips. 
Inside OverlayConfig we already have BackdropClass and PanelClass. Why not ContainerClass too?

Harpush:
I am pretty sure the correct approach is something like 'bringToFront' method... That way we can do it only when needed and not absolutely. For example when we open a select it will bring itself to front.

EliezerB123:
> Maybe it help https://blog.bbogdanov.net/post/angular-material-overlay-hack-overlay-container/

I'm horrified, but this might actually work. Thank you, @aksenovdev .

This would be SO much better if we could actually just do this normally, though.

Currently, if I have a fixed header, there's no way for me to specify that

[Menus](https://material.angular.io/components/menu/examples) will go above the header, 
but
[Selects](https://material.angular.io/components/select/examples) will go below the header,
[...]

PR #1438, opened 2016-10-05
feat(overlay): add ability to set custom class on panes

fxck:
It's more of a proof-of-concept for https://github.com/angular/material2/issues/1432, since I feel like you'd rather control this by passing the config a z-index value, but I think this is more flexible(for example when you need to control this with css media queries).

What I've changed:

**Overlay** - You can pass `paneClassName`, which can be either a string or an array of string, to a config for `Overlay`s `create`, it then adds it to the created `md-overlay-pane`, I have also added a new helper class to it `md-has-index-${id}`(which can then be used along your custom class to differentiate between more panes inside one overlay)

**Dialog** - added `paneClassName` option to `MdDialogConfig`, which is then passes to `OverlayState` and create example in the dialog-demo

**Menu** - added a new `@Input` to `md-menu`, which takes either a string or array of string, and passes it down to `OverlayState` in `[md-menu-trigger-for]`

**Overlay** - removed z-index from `overlay.scss` as it was making z-indexes of panes useless

### DIFF
--- a/src/lib/core/overlay/overlay.scss
+++ b/src/lib/core/overlay/overlay.scss
@@ -20,7 +20,6 @@
     left: 0;
     height: 100%;
     width: 100%;
-    z-index: $md-z-index-overlay-container;
   }
 
   // A single overlay pane.
--- a/src/lib/core/overlay/overlay.ts
+++ b/src/lib/core/overlay/overlay.ts
@@ -37,7 +37,7 @@ export class Overlay {
    * @returns A reference to the created overlay.
    */
   create(state: OverlayState = defaultState): OverlayRef {
-    return this._createOverlayRef(this._createPaneElement(), state);
+    return this._createOverlayRef(this._createPaneElement(state), state);
   }
 
   /**
@@ -52,10 +52,21 @@ export class Overlay {
    * Creates the DOM element for an overlay and appends it to the overlay container.
    * @returns Promise resolving to the created element.
    */
-  private _createPaneElement(): HTMLElement {
+  private _createPaneElement(state: OverlayState): HTMLElement {
     var pane = document.createElement('div');
-    pane.id = `md-overlay-${nextUniqueId++}`;
+    const id = nextUniqueId++;
+    pane.id = `md-overlay-${id}`;
     pane.classList.add('md-overlay-pane');
+    pane.classList.add(`md-has-index-${id}`);
+
+    if (state.paneClassName) {
+      // check if custom class is an array of classes
+      if (state.paneClassName.constructor === Array) {
+        pane.classList.add(...<Array<string>>state.paneClassName);
+      } else {
+        pane.classList.add(<string>state.paneClassName);
+      }
+    }
 
     this._overlayContainer.getContainerElement().appendChild(pane);
 
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -14,6 +14,9 @@ export class MdDialogConfig {
   /** The ARIA role of the dialog element. */
   role: DialogRole = 'dialog';
 
+  /** Optional custom class to be added to dialog's overlay pane. */
+  paneClassName: string | Array<string>;
+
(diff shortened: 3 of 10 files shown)


fxck:
@jelbourn can I have some feedback please? I'm having a really hard time with overlays and components using them without this feature..

src/lib/core/overlay/overlay.ts
```diff
@@ -52,10 +52,21 @@ export class Overlay {
...
+    pane.classList.add(`md-has-index-${id}`);
+
+    if (state.paneClassName) {
+      // check if custom class is an array of classes
+      if (state.paneClassName.constructor === Array) {
+        pane.classList.add(...<Array<string>>state.paneClassName);
+      } else {
+        pane.classList.add(<string>state.paneClassName);
+      }
+    }
```
jelbourn: I think this is a bit cleaner:
```ts
let overlayClass = state.overlayClass;
if (overlayClass) {
  if (typeof overlayClass == 'string') {
    overlayClass = [overlayClass];
  }
  pane.classList.add(...overlayClass);
}
```

src/lib/dialog/dialog-config.ts
```diff
@@ -14,6 +14,9 @@ export class MdDialogConfig {
   /** The ARIA role of the dialog element. */
   role: DialogRole = 'dialog';
 
+  /** Optional custom class to be added to dialog's overlay pane. */
+  paneClassName: string | Array<string>;
```
jelbourn: `string | string[]`

src/lib/dialog/dialog.ts
```diff
@@ -120,6 +120,10 @@ export class MdDialog {
         .centerHorizontally()
         .centerVertically();
 
+    if (dialogConfig.paneClassName) {
```
jelbourn: What are your use-cases for exposing this through to dialog and menu? I can see the case for custom components built on top of overlay, but I'm not sure if it's the right API surface for the leaf components.

src/lib/core/overlay/overlay.scss
```diff
@@ -20,7 +20,6 @@
     left: 0;
     height: 100%;
     width: 100%;
-    z-index: $md-z-index-overlay-container;
```
fxck: I think it HAS to be removed, otherwise any inner z-index will be ignored.

See http://jsbin.com/mesujat/1/edit?html,css,output yellow is over green, despite it having thousand times higher z-index.

src/lib/core/overlay/overlay.ts
```diff
@@ -52,10 +52,21 @@ export class Overlay {
...
    * @returns Promise resolving to the created element.
    */
-  private _createPaneElement(): HTMLElement {
+  private _createPaneElement(state: OverlayState): HTMLElement {
     var pane = document.createElement('div');
-    pane.id = `md-overlay-${nextUniqueId++}`;
+    const id = nextUniqueId++;
+    pane.id = `md-overlay-${id}`;
     pane.classList.add('md-overlay-pane');
+    pane.classList.add(`md-has-index-${id}`);
```
fxck: So you can more granularly control stacking order of multiple panes inside a single overlay.. I guess you could do the same with `:nth-child`.. but you know, with BEM and everything having a class could be handy.

I won't mind removing it though, you tell me.

src/lib/dialog/dialog.ts
fxck: The whole https://github.com/angular/material2/issues/1432 is a use-case for menu, see http://plnkr.co/edit/AqjYdsScCYPlK9w44aHQ?p=preview (try opening the menu and then clicking on something).. if I removed z-index from `md-overlay-container`, it would work, but only as long as my fixed element had z-index lower than 1000.. with this option I could simple set the menu pane to always have higher z-index.

fxck: Same thing for dialog really, I could have a fixed element that will have z-index higher than 1000 and it would be overlaying the overlay.

fxck:
Closing in favour of https://github.com/angular/material2/pull/1584 ..

PR #1446, opened 2016-10-06
fix(slide-toggle): remove view encapsulation

### LINKED ISSUE #1343: md-slide-toggle cannot be styled from outside due to encapsulation
#### Bug:
md-slide-toggle has no `encapsulation: ViewEncapsulation.None` set so it cannot be easily styled from the outside.

`encapsulation: ViewEncapsulation.None` should be turn on.
It should be easy to override the component styles

Cannot easily override component style

Alpha.9

DevVersion:
* Removes the ViewEncapsulation from the slide-toggle component to allow developer overwriting the styles easly.

Fixes #1343.

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.ts
+++ b/src/lib/slide-toggle/slide-toggle.ts
@@ -10,6 +10,7 @@ import {
   AfterContentInit,
   NgModule,
   ModuleWithProviders,
+  ViewEncapsulation,
 } from '@angular/core';
 import {HAMMER_GESTURE_CONFIG} from '@angular/platform-browser';
 import {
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -61,7 +63,7 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   private _uniqueId = `md-slide-toggle-${++nextId}`;
   private _checked: boolean = false;
   private _color: string;
-  _hasFocus: boolean = false;
+  private _hasFocus: boolean = false;
```
jelbourn: This is used in a host binding, so making it private will break AoT

DevVersion: Didn't know about that. Thanks - I reverted it and also added a comment for it.

jelbourn:
LGTM

Issue #1457, opened 2016-10-09
Hammer.js overrides native drag&drop


PseudoPsycho:
#### Bug, feature request, or proposal:
Probably a bug. Might be a proposal. Not entirely sure.

#### What is the expected behavior?
You should be able to use native HTML5 drag&drop, at least on non-Material elements.

#### What is the current behavior?
The dragstart event gets overriden by Hammer.js, so you can't access required event properties like dataTransfer.

#### What are the steps to reproduce?
With Hammer.js installed, bind the dragstart event to a function and try to access dataTransfer. It will be undefined.
Alternatively uninstall Hammer.js and do the dame steps. You will get the error message: "Hammer.js is not loaded, can not bind dragstart event".
For the first case I set up a plunker right here: http://plnkr.co/edit/CMzTNi0HKVxI9xI3ztsc?p=preview Make sure to check console output once you start dragging the first div.

#### What is the use-case or motivation for changing an existing behavior?
Angular Material 2 should not prevent you from using native APIs on other tags.

#### Which versions of Angular, Material, OS, browsers are affected?
Tested with Angular 2.0.1, Material 2.0.0-alpha.9-3 and Hammerjs 2.0.8
All browsers should be affected. Tested with Firefox 49.0.1 and Chrome 53.0.2785.143

#### Is there anything else we should know?
Overriding dragstart should not be necessary during Hammerjs initilisation. The drag event contains all necessary information.
You can read the discussion on Reddit for additional information: https://www.reddit.com/r/Angular2/comments/56kh4g/html5_drag_drop/

DevVersion:
This is because the [MdGestureConfig](https://github.com/angular/material2/blob/master/src/lib/core/gestures/MdGestureConfig.ts) is registering the drag event in Angular's event manager.

- [hammer_gestures.ts#L55](https://github.com/angular/angular/blob/2.0.2/modules/%40angular/platform-browser/src/dom/events/hammer_gestures.ts#L55)

It would be probably better to just have the `slide` event for now, because the `drag` event is not used by any component yet (only by the gestures demo)

jimitndiaye:
@DevVersion I think the *drag* event should be sufficient for gestures, if needed. *dragstart* and dragend* are required for HTML5 drag and drop and should not be overridden.
https://github.com/angular/material2/blob/master/src/lib/core/gestures/MdGestureConfig.ts#L11 should probably be modified to exclude them.

jelbourn:
[...]

PR #1458, opened 2016-10-09
Compatability with drag&drop API

PseudoPsycho:
Remove dragstart and dragend from Hammer.js config so they don't override native HTML5 drag&drop API.
drag event unused right now except for one demo. dragstart and dragend not used at all, should be removed. Eventually, their features can be accessed via drag event anyway.
This should fix issue #1457


jelbourn:
LGTM

PR #1463, opened 2016-10-10
feat(checkbox): Add color attribute.

### LINKED ISSUE #947: Add support for `color` attribute to checkbox
How do I bind the color of the checkbox widget, it does not appear to work and is stuck on accent color ?

to be able to bind the color property to the something in the parent component

stuck on accent color

try and bind the color to something other other than accent

mathebox:
Fixes #947 

Do you any suggestions how improve setting the default value?
([`src/lib/checkbox/checkbox.ts` line 138](https://github.com/angular/material2/pull/1463/files#diff-e0ff0b2171d25bd1f16f8674a3acb527R138))

### DIFF
--- a/src/lib/checkbox/_checkbox-theme.scss
+++ b/src/lib/checkbox/_checkbox-theme.scss
@@ -15,11 +15,6 @@
   // The color of the checkbox's checkmark / mixedmark.
   $checkbox-mark-color: md-color($background, background);
 
-  // The color that is used as the checkbox background when it is checked.
-  $checkbox-background-color-primary: md-color($primary, 500);
-  $checkbox-background-color-accent: md-color($accent, 500);
-  $checkbox-background-color-warn: md-color($warn, 500);
-
   // NOTE(traviskaufman): While the spec calls for translucent blacks/whites for disabled colors,
   // this does not work well with elements layered on top of one another. To get around this we
   // blend the colors together based on the base color and the theme background.
@@ -46,22 +41,16 @@
   }
 
   .md-checkbox-indeterminate, .md-checkbox-checked {
-    &.md-primary {
-      .md-checkbox-background {
-        background-color: $checkbox-background-color-primary;
-      }
+    &.md-primary .md-checkbox-background {
+      background-color: md-color($primary, 500);
     }
 
-    &.md-accent {
-      .md-checkbox-background {
-        background-color: $checkbox-background-color-accent;
-      }
+    &.md-accent .md-checkbox-background {
+      background-color: md-color($accent, 500);
     }
 
-    &.md-warn {
-      .md-checkbox-background {
-        background-color: $checkbox-background-color-warn;
-      }
+    &.md-warn .md-checkbox-background {
+      background-color: md-color($warn, 500);
     }
   }
 
@@ -81,22 +70,16 @@
 
   // TODO(jelbourn): remove style for temporary ripple once the real ripple is applied.
   .md-checkbox-focused {
-    &.md-primary {
-      .md-ink-ripple {
-        background-color: md-color($primary, 0.26);
-      }
+    &.md-primary .md-ink-ripple {
+      background-color: md-color($primary, 0.26);
     }
 
-    &.md-accent {
-      .md-ink-ripple {
-        background-color: md-color($accent, 0.26);
-      }
+    &.md-accent .md-ink-ripple {
+      background-color: md-color($accent, 0.26);
     }
 
-    &.md-warn {
-      .md-ink-ripple {
-        background-color: md-color($warn, 0.26);
-      }
+    &.md-warn .md-ink-ripple {
+      background-color: md-color($warn, 0.26);
     }
   }
 }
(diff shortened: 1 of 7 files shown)


src/lib/checkbox/_checkbox-theme.scss
```diff
@@ -63,7 +80,23 @@
...
 
   // TODO(jelbourn): remove style for temporary ripple once the real ripple is applied.
-  .md-checkbox-focused .md-ink-ripple {
-    background-color: md-color($accent, 0.26);
+  .md-checkbox-focused {
+    &.md-primary {
+      .md-ink-ripple {
+        background-color: md-color($primary, 0.26);
+      }
+    }
```
jelbourn: It would also be a little more concise as
```scss
.md-checkbox-focused {
  &.primary .md-ink-ripple {
    background: md-color($primary, 0.26);
  }
  ...
}
```

src/lib/checkbox/checkbox.scss
```diff
@@ -189,6 +190,10 @@ $_md-checkbox-indeterminate-checked-easing-function: cubic-bezier(0.14, 0, 0, 1)
 
 md-checkbox {
   cursor: pointer;
+
+  // Animation
+  transition: background $swift-ease-out-duration $swift-ease-out-timing-function,
+              md-elevation-transition-property-value();
```
jelbourn: Why is `md-elevation-transition-property-value` here? This was only meant to be used with elevation

src/lib/checkbox/checkbox.ts
```diff
@@ -128,11 +128,15 @@ export class MdCheckbox implements ControlValueAccessor {
 
   private _indeterminate: boolean = false;
 
+  private _color: string;
```
jelbourn: You can just say `private _color: string = 'accent';`

mathebox: Unfortunately that alone won't do the magic. The `md-checkbox` DOM element has to get the class `md-accent`. To achieve this you have the following options:

1. Add one of these lines to the constructor (All are somehow weird calls):
  - `this.color = 'accent';`
  - `this._updateColor(this._color)`
  - `this._setElementColor(this._color, true)`
  - ```this._renderer.setElementClass(this._elementRef.nativeElement, `md-${this._color}`, true);```
1. Alternative way: Modify the theme css like this (Introduce a "base" color):
```
.md-checkbox-indeterminate, .md-checkbox-checked {
  .md-checkbox-background {
    background-color: md-color($accent, 500);
  }
  &.md-primary .md-checkbox-background {
    background-color: md-color($primary, 500);
  }
  ...
}
```

Which option do you think is the best? Have a missed an option?

src/lib/checkbox/_checkbox-theme.scss
```diff
@@ -15,7 +16,9 @@
   $checkbox-mark-color: md-color($background, background);
 
   // The color that is used as the checkbox background when it is checked.
-  $checkbox-background-color: md-color($accent, 500);
+  $checkbox-background-color-primary: md-color($primary, 500);
+  $checkbox-background-color-accent: md-color($accent, 500);
+  $checkbox-background-color-warn: md-color($warn, 500);
```
mathebox: Done.

src/lib/checkbox/_checkbox-theme.scss
mathebox: Done if you meant `&.md-primary` instead of `&.primary`.
`MdButton` also uses the CSS classes `md-primary`, `md-accent` and `md-warn`.

src/lib/checkbox/checkbox.ts
mathebox: @jelbourn Any thoughts on this?

jelbourn:
LGTM

I'm not sure we'll keep all of the `color` attributes in the long term (vs. preferring styling via theming), but this is good for now.

jaibatrik:
Now that this is merged, how do I actually change the color of the checkbox? I have tried doing this - 
``` html
<md-checkbox [checked]="(task.state == 'STATE_CLEARED')" color="red"></md-checkbox>
```

All it does is to add the `mat-red` class to `md-checkbox` element. I tried putting the following styles in my Component CSS - 

``` css
.mat-red .mat-checkbox-background {
  background-color: red;
}
```

But the selectors get converted to `.mat-red[_ngcontent-c7]   .mat-checkbox-background[_ngcontent-c7]`.

I don't want to define global styles for this.

willshowell:
@jaibatrik your color options are `primary|accent|warn`. See the [docs](https://material.angular.io/components/component/checkbox#theming)

jaibatrik:
Oh. But what if I want my checkbox to be colored red? Or any custom color? Is there any way to do it? Let's suppose none of primary, accent, and warn are red in my theme.

willshowell:
In that case, global styles or the `/deep/` selector (search for it on [this page](https://angular.io/docs/ts/latest/guide/component-styles.html)) should do it for you.

If you're still unable to get the override, make sure your selectors have a [higher specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) than the original selector.

jaibatrik:
Just so that others can know, I was able to achieve this by using the `/deep/` selector and constructing the selector like the following - 

```css
.mat-red.mat-checkbox-checked /deep/ .mat-checkbox-background {
  background-color: red;
}
```

According to the docs, `\deep\` will propagate to any depth of children components; but since the checkbox doesn't have any, I'll be safe here.

Thanks!

PR #1466, opened 2016-10-11
fix(list): ensure multi-line lists expand to fill space

kara:
The Material Design spec only recommends a maximum of three lines in list items, so this is what we formally support. However, if you decide to go off-spec and add more than three lines to a list item, all the lines are currently squeezed into a shorter height (which looks awful). 

This PR allows list items with 4+ lines to expand to match the height of the contained text so they will look a bit saner.

r: @jelbourn

### DIFF
--- a/src/lib/list/list.scss
+++ b/src/lib/list/list.scss
@@ -50,8 +50,16 @@ $md-dense-three-line-height: 76px;
     height: $two-line-height;
   }
 
+
   &.md-3-line .md-list-item {
     height: $three-line-height;
+  }
+
+  // list items with more than 3 lines should expand to match
+  // the height of its contained text
+  &.md-multi-line .md-list-item {
+    height: 100%;
+    padding: 8px $md-list-side-padding;
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/list/list.scss
```diff
@@ -54,6 +54,11 @@ $md-dense-three-line-height: 76px;
     height: $three-line-height;
   }
 
+  &.md-multi-line .md-list-item {
```
jelbourn: Add comment explaining what the `md-multi-line` class means.

src/lib/list/list.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('MdList', () => {
         ListWithItemWithCssClass,
         ListWithDynamicNumberOfLines,
         ListWithMultipleItems,
+        ListWithManyLines
```
jelbourn: Trailing comma

kara: ye with the trailing commas. done!

PR #1470, opened 2016-10-11
fix(tooltip): remove tooltip component after it's parent destroyed

### LINKED ISSUE #1111: bug(md-tooltip): tooltip is not removed when parent is
Consider a scenario where there's a button inside a container. The button has a tooltip and on click it should set a variable to another value. Based on that value the container element is added / removed with `ngIf`. Simple example:

``` html
<div *ngIf="!hidden">
  <button
    md-button
    md-tooltip="Hide this button"
    tooltip-position="after"
    (click)="hidden = true">HIDE ME!</button>
</div>
```

After clicking the button the container div is removed, however, the tooltip remains still, no way to remove it.

[Plunkr example](http://plnkr.co/edit/GX8CnKNQ8Zlzm5Zras7r?p=preview).

gucheen:
fix #1111

### DIFF
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -50,6 +50,14 @@ describe('MdTooltip', () => {
       tooltipDirective._handleMouseLeave(null);
       expect(overlayContainerElement.textContent).toBe('');
     });
+
+    it('should be removed after parent destroyed', () => {
+      tooltipDirective._handleMouseEnter(null);
+      expect(tooltipDirective.visible).toBeTruthy();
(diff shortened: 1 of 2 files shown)


gucheen:
I signed it!

jelbourn:
@gucheen could you add a unit test for this?

gucheen:
Unit test has been added, but there was something wrong with the test scripts. 'gulp' was missing in travis.

jelbourn:
@andrewseguin this may be obsolete based on your animation change, thoughts?

jelbourn:
LGTM

Issue #1445, opened 2016-10-06
Secondary router-outlet not working with md-tabs


tomwanzek:
#### Bug, feature request, or proposal:

First of, thanks for all the hard work on Material 2!

I am using alpha 9-3 and as it appears, the issue I am encountering is related to this item: #524.

I intended to use a sidenav which contains a tab group. One of the tabs was conceived to serve as a context tab which is populated by a secondary, named router outlet `<router-outlet name="context"></router-outlet>.

#### What is the expected behavior?

* Option 1:
The secondary router-outlet can contain a complete `<md-tab>...</md-tab>`, which can be switched to and from when changing the active tab in the group. (preferred)

* Option 2:
The secondary router-outlet provides the content for an already "hard-coded"
```html
<template md-tab-content>
   <router-outlet name="context"></router-outlet>
</template md-tab-content>
```
This implies:
(a) a tertiary router-outlet is required to provide the `<template md-tab-label>` name
(b) the tab shell remains in the template, even for routes that do not required the context menu.

#### What is the current behavior?

Option 1: The secondary router outlet does not appear in the DOM. No rendering. No error messages.

Option 2: The secondary router outlet is initially rendered as expected into the tab content. However, when switching away from the tab and  then back to it, the content is empty.

#### What are the steps to reproduce?

see description of steps above.

#### What is the use-case or motivation for changing an existing behavior?

The ability to use tabs based on router-state with primary and secondary outlets, e.g. as described above.

#### Which versions of Angular, Material, OS, browsers are affected?

angular material 9-3
angular-cli: 1.0.0-beta.16
node: 6.5.0
os: win32 x64

Browsers: Chrome, Firefox, Edge

#### Is there anything else we should know?

EDIT: Added issue reference to #524 which was accidentally omitted above.

jelbourn:
Using tabs (or a tabs-like component) is captured by #524 and is planned work.

tomwanzek:
@jelbourn @andrewseguin I just had a quick look over the proposed solution to #524. While viable in a number of use cases, it does not seem to address the above use case.

Any suggestions? The only other solution I can see is related to using a [hide] option as contemplated in #731.

Thanks in advance for your insight.

jelbourn:
@tomwanzek we will be adding a "hidden" option for tabs (per #731)

tomwanzek:
@jelbourn thx.

chenyingyin:
[...]

Issue #524, opened 2016-05-25
Tabs + Routing


jelbourn:
This captures using tabs (or a tab-like ui) with routing.

apals:
Any ETA on this?

naveedahmed1:
It would be really nice to have md-tabs linked with router.

zigzag95:
You mean with async routing and a nice progress circle while getting data. Yeah it will be great.

DennisSmolek:
So I was wondering why build the routing directly into the component? With so many routers out there (Core, Beta, UI, ngrx, ionic) wouldn't it be better to just have the tabs component emit events and allow the user to attach to it?

Instead of doing `<md-tabs><md-tab [router-link]="['/whatever',"1234"]"></md-tab></md-tabs>` you would do: `<md-tabs><md-tab (tabNavigate)="doRouterAction(navData)"></md-tab>` or even: `(tabNavigate)="router.location(event)"`

for progress you could do: `[nav-progress]="getNavProgress()"`

This way the tabs stay flexible for dev's to use the router of their choice..

you could even attach the event on just the main `<md-tabs>` and keep the rest clean...

jelbourn:
It's just tabs that have a routing outlet in them and the "tabs" are anchors.

DennisSmolek:
That's true, I wasn't thinking about the outlet being built in...

kyleledbetter:
Probably obvious, but if tabs are used as a router (or regardless), it would be nice to conditionally not render the `<div class="md-tab-body-wrapper">`

https://github.com/angular/material2/blob/master/src/components/tabs/tab-group.html#L17-L25

zigzag95:
@jelbourn you are ok for a PR implementing md-tab-routed or you prefer upgrade actual md-tab?

jelbourn:
@robertmesserle is already working on this.

beakid:
@jelbourn @robertmesserle Any updates? :)

vespertilian:
A bit like what @DennisSmolek discussed, I think it would be nice to be able to access the 'tab headers' separately from the tabs. So you could roll your own tabs with the excellent tab transition that angular materials gives you, using a vanilla router and outlet.

I imagine something like this.   

```html
<md-tab-group-manual>
  <!--tab headers-->
  <md-tab-headers>
    <md-tab-label>
      <a routerLink="/tabA">Tab A</a>
    </md-tab-label>
    <md-tab-label>
[...]

PR #1472, opened 2016-10-11
Ability to Hide Tabs

RicardoVaranda:
This PR was made for people to give people the ability to hide tab contents instead of the traditional *ngIf which removes it and adds it to the DOM.

This will allow us to use the tabs just like before but by passing hideTabs input of true it will hide the tab contents instead of removing and adding them to the DOM, there are some setbacks to this and it should be used with moderation, i.e you shouldn't hide the contents if you have a large number of tabs e.g 10+

### DIFF
--- a/src/demo-app/tabs/tab-group-demo.html
+++ b/src/demo-app/tabs/tab-group-demo.html
@@ -13,6 +13,21 @@
   </md-tab>
 </md-tab-group>
 
+<h1>Tab Group Hiding tabs Demo</h1>
+
+<md-tab-group [hideTabs]="true" class="demo-tab-group">
+  <md-tab *ngFor="let tab of tabs; let i = index" [disabled]="i == 1">
+    <template md-tab-label>{{tab.label}}</template>
+    <template md-tab-content>
+      {{tab.content}}
+      <br>
+      <br>
+      <br>
+      <md-input placeholder="Tab Label" [(ngModel)]="tab.label"></md-input>
+    </template>
+  </md-tab>
+</md-tab-group>
+
 <h1>Async Tabs</h1>
 
 <md-tab-group class="demo-tab-group">
--- a/src/lib/tabs/tab-group.html
+++ b/src/lib/tabs/tab-group.html
@@ -23,7 +23,8 @@
     <template [ngIf]="selectedIndex == i && !hideTabs">
       <template [portalHost]="tab.content"></template>
     </template>
-    <div [class.hidden]="selectedIndex !== i && hideTabs">
+    <div [ngIf]="selectedIndex == i && hideTabs"
+         [class.hidden]="selectedIndex !== i && hideTabs">
       <template [portalHost]="tab.content"></template>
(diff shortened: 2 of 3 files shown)


ZeBigDuck:
Please accept this PR! Without this, md-tabs are inusable.

ksom:
+1
Hi, thanks @RicardoVaranda.
I'm waiting for this fix too.
Is there something missing ? Or it's possible to merged it?

Thanks

RicardoVaranda:
@ksom I'm using this in a project for about a month now and it is working fine considering you don't have 10+ tabs opened at which point you should/will start to notice a decrease in performance.

andrewseguin:
Tab contents are now loaded all at once when the tab group is initialized due to change #1645. Switching between tabs will not re-initialize/destroy the content but rather just perform some portal logic to attach the  tab contents.  The hideTabs input should not be necessary anymore.

Ricardo, it would be great if you can verify that the current behavior in master satisfies what you needed in your app without this PR.

tomwanzek:
@andrewseguin are the changes in #1645 addressing the use case described in issue #1445?

I.e. the intent is to have sidenav which contains a tab-group where:
* one tab contains a named secondary-router outlet with a route-specific context menu
* other tabs contain components, which are persisted across routes (i.e. not router-outlets)

Thanks in advance for a brief reply, as the "hide" option was the fall-back now that #524 (Tabs + Routing) is closed.

andrewseguin:
@tomwanzek I believe this scenario will work just fine. I tested the scenario of a tab group containing one tab with a router outlet and the other tab with an md-button. The tab with the router outlet displayed the router outlet content correctly when navigating between links (even when the tab was not active). The second tab created the md button once and did not recreate each time the tab was displayed.

Does this satisfy your requirements? I know that you suggested another scenario of having the router outlet itself containing the md-tab template but the md-tab-group currently will not update itself when tabs appear/disappear dynamically.

tomwanzek:
@andrewseguin Thanks for your response. I think the main reason I initially leaned towards having the secondary router-outlet containing a tab (option 1 in issue #1445) was the idea of conditionality.

I.e. if there is a route which does not utilize the secondary router-outlet e.g. a "home" route without an additional context-related tab, then there would be no spurious tab at all.

So as you point out, if the `md-tab-group` does not currently dynamically update, if tabs appear/disapear (e.g. also `ngIf` ), it seems there is not really an way of addressing it.

But at least it is possible to mix-and-match "regular" tabs and  tabs containing a refreshing (secondary) router outlet. So thanks for the progress made 👍

andrewseguin:
For sure, and it sounds like your situation might be worked around with using CSS to hide the context-related tab. I'd like to look further into the possibility of dynamically added/removed tabs as well.

anupvarma:
Can somebody let me know if my issue #5239 is related to this issue.
plunker: http://plnkr.co/edit/DADYadzGOFI5GqkkITkA?p=preview

Issue #1104, opened 2016-08-23
bug(menu): Error Host already has a portal attached


cladera:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
When open a md-menu programatically using trigger's api it should not throw an exception

#### What is the current behavior?

When a menu is opened and closed several times it eventually throws an exception:

```
browser_adapter.js:95 Error: Uncaught (in promise): Error: Host already has a portal attached
    at resolvePromise (zone.js:538)
    at resolvePromise (zone.js:523)
    at zone.js:571
    at ZoneDelegate.invokeTask (zone.js:356)
    at Object.onInvokeTask (ng_zone_impl.js:55)
    at ZoneDelegate.invokeTask (zone.js:355)
    at Zone.runTask (zone.js:256)
    at drainMicroTaskQueue (zone.js:474)
    at HTMLInputElement.ZoneTask.invoke (zone.js:426)
    at MdInput.focus (input.js:202)
```

#### What are the steps to reproduce?

http://plnkr.co/edit/1KDM4LwHtlpJQAUbAuur?p=preview

#### What is the use-case or motivation for changing an existing behavior?

I'd want to be able to open and close a menu not only with click event but other events such as "focus".

#### Which versions of Angular, Material, OS, browsers are affected?
Angular 2.0.0-rc.5 Material 1.0.0-alpha.7-4

#### Is there anything else we should know?

Besides the error, the plunker does not behave as expected. When using "focus" handler the menu opens but immediately closes. It is necessary to click the input again to open the menu. In my application this does not happen.

PR #1478, opened 2016-10-11
fix(menu): make menu open idempotent

kara:
Calling `openMenu()` multiple times shouldn't cause issues.

Closes https://github.com/angular/material2/issues/1104.

### DIFF
--- a/src/lib/menu/menu.spec.ts
+++ b/src/lib/menu/menu.spec.ts
@@ -1,6 +1,7 @@
 import {TestBed, async} from '@angular/core/testing';
-import {Component} from '@angular/core';
-import {MdMenuModule} from './menu';
+import {Component, ViewChild} from '@angular/core';
+import {By} from '@angular/platform-browser';
+import {MdMenuModule, MdMenuTrigger} from './menu';
 
 
 describe('MdMenu', () => {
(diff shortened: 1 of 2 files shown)


src/lib/menu/menu.spec.ts
```diff
@@ -14,11 +14,24 @@ describe('MdMenu', () => {
...
 
-  it('should add and remove focus class on focus/blur', () => {
+  it('should open the menu as an idempotent operation', () => {
     let fixture = TestBed.createComponent(TestMenu);
-    expect(fixture).toBeTruthy();
+    fixture.detectChanges();
+    expect(() => {
+      fixture.componentInstance.trigger.openMenu();
+      fixture.componentInstance.trigger.openMenu();
+    }).not.toThrowError();
```
jelbourn: And also expect that the menu is still open?

src/lib/menu/menu.spec.ts
```diff
@@ -14,11 +14,24 @@ describe('MdMenu', () => {
...
-class TestMenu {}
+@Component({
+  template: `
+    <button [md-menu-trigger-for]="menu">Toggle menu</button>
+    <md-menu #menu="mdMenu">
+      Content
+    </md-menu>
+  `
+})
+class TestMenu {
```
jelbourn: How about `SimpleMenu`?

PR #1486, opened 2016-10-12
docs: Document MdInput focus() method

feloy:
Document the MdInput focus() method

### DIFF
--- a/src/lib/input/README.md
+++ b/src/lib/input/README.md
@@ -90,7 +90,7 @@ You can put the focus on an input component using the `focus()` method.
 
 ```ts
 export class MyComponent implements OnInit {
-  @ViewChild(MdInput) nameInput: MdInput;
+  @ViewChild('nameInput') nameInput: MdInput;
 
   ngOnInit() {
     this.nameInput.focus();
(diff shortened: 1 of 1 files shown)


fxck:
I think really think this belongs to the docs, it's a workaround for a missing feature..

src/lib/input/README.md
```diff
@@ -78,7 +78,25 @@ The divider (line under the `<md-input>` content) color can be changed by using
...
 You can label the `<md-input>` as you would a regular `<input>`.
 
+## Focus
 
+You can put the focus on an input component using the `focus()` method.
+
+### Example
+
+```html
+<md-input #nameInput placeholder="name"></md-input>
```
jelbourn: The `#nameInput` isn't necessary in this example because you're using `@ViewChild` with the class. I'd omit it so that people don't mistakenly get the impression that a `@ViewChild` name has to match some template variable name.

feloy: You're right, I used the ViewChild incorrectly. What I wanted to write was instead:
@ViewChild('nameInput') nameInput: MdInput;
to that this sample still works when multiple md-input are present in the template.
Do you agree?

jelbourn: Yep, that works too.

Issue #1494, opened 2016-10-13
Update Travis CI to use yarn instead of npm install


jelbourn:
Need to verify that it's faster, but yarn will likely be an improvement.

jelbourn:
Relevant https://twitter.com/kentcdodds/status/786282951210262528

pi0:
Please checkout [YarNPM](https://github.com/pi0/yarnpm) :)

d3viant0ne:
@jelbourn - Tinkered around with this on my fork, the end result is it's a wash.

We spend ~70 seconds give or take to install with npm ( no cache ) and under 10 seconds with it.
With yarn it takes ~50 seconds ( no cache ) and ~8 seconds with it. 

In either case you have to pull down the cache from Amazon s3, a large chunk of the overall performance improvements you see with yarn locally disappear when you cache isn't local. 

It's a pretty consistent theme. Yarn is great for a lot of things, container driven, hosted CI really isn't one of them. The only place I have had a great deal of success is an internal Jenkins server running ec2 slaves.

Also worth noting, there are side effects in the gulp scripting that would have to be fixed should you chose to have this implemented.

d3viant0ne:
If you still want it done, i'll clean up the work and PR in the Travis changes and fixes to the gulp scripting.

d3viant0ne:
It's also worth noting, if switching to yarn is creating side effects in the build scripting it's reasonable we are going to see it in the library proper. IMO the minimal time savings that will exist once yarn is included in the Travis containers doesn't outweigh the risk of the above mentioned side effects.

jelbourn:
@d3viant0ne thanks for investigating! Since there's not a noticeable improvement, I'm okay with sticking to `npm`.

PR #1496, opened 2016-10-13
chore: use yarn instead of npm in Travis

abdulhaq-e:
This is an attempt at https://github.com/angular/material2/issues/1494

~~The method used is slightly different than the one referenced in the issue, yarn was installed from its repository rather than using npm.~~
~~We'll see what Travis does.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,8 @@ matrix:
   - env: "MODE=saucelabs_optional"
   - env: "MODE=browserstack_optional"
 
+# i will add yarn
+
 install:
   - npm install
 


abdulhaq-e:
Hello @googlebot , I signed it!

.travis.yml
```diff
@@ -43,10 +43,15 @@ matrix:
...
 install:
-  - npm install
+  - yarn install
 
 before_script:
   - mkdir -p $LOGS_DIR
+  - sudo apt-key adv --keyserver pgp.mit.edu --recv D101F7899D41F3C3
+  - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+  - sudo apt-get update -qq
+  - sudo apt-get install yarn -y
```
DevVersion: From what I know `Yarn` can be also installed per `NPM`.

```
npm i -g yarn
```

Should not need any extra sources for APT.

abdulhaq-e: Yes I'm aware of that. I just wanted to test this, then perhaps I'll revert to `npm i -g yarn`

abdulhaq-e:
The build fails, I reverted everything just in case `yarn` was the culprit but it still fails:

> TSError: ⨯ Unable to compile TypeScript
> node_modules/@types/orchestrator/node_modules/@types/q/index.d.ts (7,21): Duplicate identifier 'Q'. (2300)
> node_modules/@types/q/index.d.ts (7,21): Duplicate identifier 'Q'. (2300)

jelbourn:
I also got this error when I tried yarn locally, but everything was still fine with `npm install`. Not sure what's up yet

jelbourn:
Closing this until we figure out why `yarn` doesn't work while `npm i` does.

Issue #1479, opened 2016-10-11
Prevent CSS collisions between Angular Material v1 and v2


jelbourn:
Presently, both versions use the same element names which can lead to collisions during migration.

andrewseguin:
Fixed by #2737

PR #1498, opened 2016-10-13
feat(core): directive to add extra classes to components for m1 compat

jelbourn:
See #1479

I don't like the name "Compatibility" but I don't have any better ideas. 

I had originally been thinking we'd have to add the class in the constructor for each individual components, but this approach is much cleaner.


jelbourn:
Good to merge if CI is green

jelbourn:
Nope, problem with md-menu (which already has an `md-menu` class).

PR #1505, opened 2016-10-14
chore: add autoprefixer to gulp

DevVersion:
* At the time of moving to gulp, the autoprefixer has been removed from the project.

* Re-introducing the autoprefixer for the "latest two versions" as in the README stated, does make sure that all CSS properties are working within the supported browsers.

### DIFF
--- a/tools/gulp/constants.ts
+++ b/tools/gulp/constants.ts
@@ -7,7 +7,11 @@ export const DIST_ROOT = join(PROJECT_ROOT, 'dist');
 export const DIST_COMPONENTS_ROOT = join(DIST_ROOT, '@angular/material');
 
 export const SASS_AUTOPREFIXER_OPTIONS = {
-  browsers: ['last 2 versions'],
+  browsers: [
+    'last 2 versions',
+    'not ie <= 10',
+    'not ie_mob <= 10',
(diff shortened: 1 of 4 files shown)


tools/gulp/constants.ts
```diff
@@ -6,6 +6,10 @@ export const SOURCE_ROOT = join(PROJECT_ROOT, 'src');
 export const DIST_ROOT = join(PROJECT_ROOT, 'dist');
 export const DIST_COMPONENTS_ROOT = join(DIST_ROOT, '@angular/material');
 
+export const SASS_AUTOPREFIXER_OPTIONS = {
+  browsers: ['last 2 versions'],
+  cascade: false,
+};
```
jelbourn: If I recall correctly, we need to explicitly exclude IE10 and mobile IE because autoprefixer considers "IE" and "Edge" to be separate browsers, thus IE10 falls within the last two versions. This is necessary to prevent it from adding the outdated ms-flex prefixed rules.

DevVersion: I see. Just took a look at https://github.com/ai/browserslist and you're right.

jelbourn:
LGTM

Issue #1506, opened 2016-10-14
Menu should destroy click catcher when trigger is destroyed


kara:
If the menu trigger is destroyed while the menu is open (e.g. through an ngIf), the menu's overlay is destroyed, but it never kills the click catcher.

jelbourn:
Closed via #1534

PR #1507, opened 2016-10-14
fix(menu): destroy click catcher when trigger is destroyed

kara:
Eventually we should switch to using the overlay's backdrop as a click-catcher. Hot fix until then.

Closes https://github.com/angular/material2/issues/1506


src/lib/menu/menu-trigger.ts
```diff
@@ -84,6 +84,7 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
     if (this._overlayRef) {
       this._overlayRef.dispose();
       this._overlayRef = null;
+      this._resetMenu();
```
jelbourn: Unit test?

PR #1508, opened 2016-10-15
chore(saucelabs): fix microsoft edge version

DevVersion:
The last days the Saucelabs mode in the CI is failing with Microsoft Edge.

> As per error message, there seems to be an invalid browser configuration.

While looking at the Saucelabs API and comparing the "available" versions for `SL_EDGE`, it looks like they have changed the versions to match the `HTML` engine

See https://gist.github.com/DevVersion/ffcd03143de80a22bc49aae52b612d16


jelbourn:
Ah, good catch. Thanks!

Issue #1475, opened 2016-10-11
MdRadioGroup is marked "not pristine" when setValue(...) is called


feloy:
#### Bug, feature request, or proposal:

Bug

#### What is the expected behavior?

In a reactive Form, when calling setValue(...) method of a formControl, the formControl should not be marked as not pristine.

#### What is the current behavior?

A MdRadioGroup in a reactive form is marked as not pristine when setValue(...) is called on the control form.

#### What are the steps to reproduce?

View github https://github.com/feloy/ng2-master-detail 
 - tag bug-dirty-ok : works as expected with input radio
 - tag bug-dirty-error : does not work as expected with md-radio-button

#### Which versions of Angular, Material, OS, browsers are affected?

Angular 2.0.2
Material 2.0.0-alpha.9-3
OS Linux
Browsers tested: Chrome 50, Firefox 46

feloy:
More details:
Here is the comment @angular/forms describing that a form control should not be marked ditry when changed programmatically:
https://github.com/angular/angular/blob/master/modules/%40angular/forms/src/model.ts#L180

I also see an error in the radio tests, where it is tested that the control should become dirty after being changed programmatically:
https://github.com/angular/material2/blob/master/src/lib/radio/radio.spec.ts#L283

PR #1521, opened 2016-10-18
fix(radio): only call change callback with user input

kara:
Currently, we're calling the ControlValueAccessor change callback whenever the value changes, whether that's externally (control.setValue) or internally (through user input). It should really only be called due to internal changes. Otherwise, the control will be marked dirty erroneously.

r: @jelbourn

Closes https://github.com/angular/material2/issues/1475


jelbourn:
LGTM

PR #1532, opened 2016-10-19
feat(overlay): add custom classes for backdrop

kara:
This PR allows components to add a custom class to the overlay backdrop. This will allow us to use the backdrop as a click catcher in components like the menu and the select.

r: @jelbourn

### DIFF
--- a/src/demo-app/overlay/overlay-demo.html
+++ b/src/demo-app/overlay/overlay-demo.html
@@ -25,3 +25,5 @@
 <template portal>
   <p class="demo-fusilli"> Fusilli </p>
 </template>
+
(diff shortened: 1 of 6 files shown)


fxck:
https://github.com/angular/material2/pull/1438

kara:
@fxck Can you elaborate? Your PR appears to have to do with overlay panes, not overlay backdrops. I assume @jelbourn will want to weigh in on the design for that one.

fxck:
Hm, thought the backdrop was a child of the pane and thus the custom class on it could be used, but it's a child of the overlay.. so yeah. Ignore my comment.

jelbourn:
LGTM

Issue #1360, opened 2016-09-28
Issue : multiple md-menu opened at the same time


Ploppy3:
#### Bug, feature request, or proposal:
It's a bug.
It's possible to open multiple md-menu at the same time due to a z-index conflict.

#### What is the expected behavior?
It should not be possible to open multiple md-menu at the same time.

#### What is the current behavior?
Clicking on multiple menu buttons open their respective menus at the same time instead of closing the previous one.

#### What are the steps to reproduce?
I did not manage to get the md-menu to work on plunker.
To reproduce, place multiple md-icon-button and their respective md-menu, and click on them.
Screenshot of the issue : http://i.imgur.com/9S0JsLe.png

#### What is the use-case or motivation for changing an existing behavior?
This bug affects the user experience.

#### Which versions of Angular, Material, OS, browsers are affected?
Angular-CLI : 1.0.0-beta.16
Material : 2.0.0-alpha.9
Browsers : Every browsers

#### Is there anything else we should know?
The issue is related to the the z-index of the material design components.
The md-button-ripple-round used for the md-icon-button has a z-index of 1 and the md-menu-click-catcher z-index is not set by default therefore it is possible to click on the ripple of any md-icon-button without clicking on the md-menu-click-catcher.

Adding a z-index of 1 to the md-menu-click-catcher seems to fix the issue.

It's my first issue report. I hope it's not a mistake on my side.

Issue #1499, opened 2016-10-14
md-menu click catcher is fixed to a container.


shlomiassaf:
#### Bug, feature request, or proposal:
bug

#### What is the expected behavior?
The click catcher for md-menu should be applied on the whole screen.

#### What is the current behavior?
```
<div class="md-menu-click-catcher"></div>
```
The click catcher element is located within the menu directive this means that in some situations the fixed viewport is not the body but some element in the DOM tree between the body and the click catcher.
This can happen when an element (ancestor of css catcher) applies CSS transform. When this happen the click catcher will be bound to the element's view port and not the body.

A great example is the `SideNav`, the sidenav's content element applies a `transform: translate3d(0px, 0px, 0px);` and when opened the `margin-left` CSS property reduce's the content size thus reduces the click catcher size, in this scenario clicks over the sidebar navigation are will not trigger a close event. (hovers will works as well... etc)

The menu is rendered in the the overlay container area but its not using the overlay backdrop, I guess since you can't control the styling of the backdrop to make it transparent...

shlomiassaf:
Another issue related to the click catcher element is when a menu item has a `routerLink` that does not exist. This will trigger an exception.

When this happens the click catcher is not removed, if the click catcher is not inside a router outlet (above it) the whole app becomes non responsive (can't click)

fxck:
related
https://github.com/angular/material2/issues/998
https://github.com/angular/material2/issues/1417

shlomiassaf:
The issue is really using a backdrop inside the menu content but creating the dom element inside the overlay container.

The right solution is to use the overlay's backdrop and being able to control the backdrop background color

jelbourn:
The current behavior is just temporary until menu uses the Overlay's backdrop.

jelbourn:
Closed via #1534

PR #1534, opened 2016-10-19
fix(menu): update to use overlay backdrop

kara:
Closes https://github.com/angular/material2/issues/1360, https://github.com/angular/material2/issues/1499, #1506 .

r: @jelbourn

### DIFF
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -23,6 +23,7 @@ import {
     HorizontalConnectionPos,
     VerticalConnectionPos
 } from '../core';
+import { Subscription } from 'rxjs/Subscription';
 
 /**
  * This directive is intended to be used in conjunction with an md-menu tag.  It is
@@ -40,6 +41,7 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
   private _portal: TemplatePortal;
   private _overlayRef: OverlayRef;
   private _menuOpen: boolean = false;
+  private _backdropSubscription: Subscription;
 
   // tracking input type is necessary so it's possible to only auto-focus
   // the first item of the list when the menu is opened via the keyboard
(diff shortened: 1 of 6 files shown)


src/lib/menu/menu-trigger.ts
```diff
@@ -70,6 +70,7 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
     if (!this._menuOpen) {
       this._createOverlay();
       this._overlayRef.attach(this._portal);
+      this._overlayRef.backdropClick().subscribe(() => this.closeMenu());
```
jelbourn: Unit test?

kara: There's an e2e that covers this already, but I can add a unit too if you think it will help coverage?

fxck: Shouldn't it be 

```
this._overlayRef.backdropClick().take(1).subscribe(() => this.closeMenu());
```

to unsubscribe the observer since the backdrop element will be removed from DOM? see https://jsfiddle.net/chy8gLjs/ vs. https://jsfiddle.net/j4p7gq1v/

jelbourn: Since the backdrop doesn't have anything to do with positioning, I'd do this as a unit tests. We generally want to prefer capturing behaviors with unit tests because they are so much faster.

@fxck I would use `.first()`, but yeah that should be the the case.

fxck: Same thing. Anyway I think there's couple more of these potential leaks across the repo. Either not using take/first when it's supposed to happen only once or not unsubscribing on destroy, when it's actually supposed to listen on changes until the component is destroyed. I'll create an issue if I get couple of minutes.

jelbourn:
LGTM

jelbourn:
Looks like you might have to `.trim()` your `textContent` for mobile safari

kara:
@jelbourn Finally passing.

PR #1535, opened 2016-10-19
chore: close sidenav on nav item click

andrewseguin:
Use ngFor to loop through nav items on the sidenav. Add click event handler for each item to close the sidenav.

Note: did not add click handler to md-nav-list so that clicking the `<hr>` element is no-op


jelbourn:
LGTM

PR #1536, opened 2016-10-19
fix(progress-circle): allow value to be set to 0

andrewseguin:
Value should be considered invalid when null, not just a falsey value. Otherwise 0 is considered an invalid value.


jelbourn:
LGTM

Issue #1541, opened 2016-10-20
Theming: Double stylesheet file in node package


NCC1701M:
#### What is the expected behavior?
Building a custom theme with sass, gulp and gulp-sass

#### What is the current behavior?
When I try to build an own theme with sass I get a transpilation error because in material/core/overlay are both files overlay.css and overlay.scss and so sass doesn't know which it should pick.

![image](https://cloud.githubusercontent.com/assets/7260053/19550770/5652bb4e-96ab-11e6-9508-b6438edbaeb9.png)

![image](https://cloud.githubusercontent.com/assets/7260053/19550991/68f75aac-96ab-11e6-9b9f-b4c367144ce7.png)

#### What are the steps to reproduce?
Follow the steps at [https://github.com/angular/material2/blob/master/docs/theming.md#defining-a-custom-theme](your theming guide)

Run a gulp task to compile the scss files to css. Use at least the following devDependencies

````json
{
   // ...
   "gulp": "^3.9.1",
   "gulp-sass": "^2.3.2",
   "sass": "^0.5.0",
   // ...
}
````

#### Which versions of Angular, Material, OS, browsers are affected?
Angular: 2.1.0
Material: 2.0.0-alpha.9 cobalt-kraken
OS: Microsoft Windows [Version 6.3.9600] and [Version 6.1.7601]
other versions see in above

jelbourn:
Yeah, `overlay.scss` should be a partial (`_overlay.scss`).

PR #1544, opened 2016-10-20
change overlay.scss to be a sass partial

fxck:
I think this should be the only change needed. Fixes https://github.com/angular/material2/issues/1541


jelbourn:
LGTM, thanks!

PR #1549, opened 2016-10-20
`unsubscribe` some subscriptions on destroy / `first` some subscriptions

fxck:
I'm not 100% sure you need to unsubscribe those `changes` subscriptions, but I don't see it being done automatically anywhere(https://github.com/angular/angular/blob/2.1.0/modules/%40angular/core/src/linker/query_list.ts#L11-L100).

Also I'm not 100% sure that dialog and icon needs that `first`, but looking at code I think both should have it.

There's also a couple of subscriptions on EventEmitter(in sidenav particularly) which I haven't touched since I supposed those *are* automatically unsubscribed.

Also I haven't touched the one mentioned in https://github.com/angular/material2/pull/1534

@jelbourn 

// edit 
looks like `changes` is event emitter as well https://github.com/angular/angular/blob/master/modules/%40angular/facade/src/async.ts so I guess either it's not needed to unsubscribed at all, or those event emitter subscriptions in sidenav needs to be unsubscribed as well

### DIFF
--- a/src/lib/core/line/line.ts
+++ b/src/lib/core/line/line.ts
@@ -6,6 +6,7 @@ import {
     QueryList
 } from '@angular/core';
 
+import { Subscription  } from 'rxjs/Subscription';
 /**
  * Shared directive to count lines inside a text area, such as a list item.
  * Line elements can be extracted with a @ContentChildren(MdLine) query, then
@@ -16,15 +17,20 @@ export class MdLine {}
 
 /* Helper that takes a query list of lines and sets the correct class on the host */
 export class MdLineSetter {
+  private _changesSubscription: Subscription;
+
   constructor(private _lines: QueryList<MdLine>, private _renderer: Renderer,
               private _element: ElementRef) {
     this._setLineClass(this._lines.length);
 
-    this._lines.changes.subscribe(() => {
+    this._changesSubscription = this._lines.changes.subscribe(() => {
       this._setLineClass(this._lines.length);
     });
   }
 
+  public destroy() {
+    this._changesSubscription.unsubscribe();
+  }
 
   private _setLineClass(count: number): void {
     this._resetClasses();
--- a/src/lib/core/line/line.ts
+++ b/src/lib/core/line/line.ts
@@ -6,7 +6,6 @@ import {
     QueryList
 } from '@angular/core';
 
-import { Subscription  } from 'rxjs/Subscription';
 /**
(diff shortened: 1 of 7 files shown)


src/lib/core/line/line.ts
```diff
@@ -6,6 +6,7 @@ import {
     QueryList
 } from '@angular/core';
 
+import { Subscription  } from 'rxjs/Subscription';
```
jelbourn: Remove spaces from around imported symbol

src/lib/core/line/line.ts
```diff
@@ -16,15 +17,20 @@ export class MdLine {}
...
 
-    this._lines.changes.subscribe(() => {
+    this._changesSubscription = this._lines.changes.subscribe(() => {
       this._setLineClass(this._lines.length);
     });
   }
 
+  public destroy() {
+    this._changesSubscription.unsubscribe();
+  }
```
jelbourn: This particular unsubscription isn't necessary because the `QueryList` will be destroyed at the same time as the component (and the subscription is contained within the component that also contains this LineSetter).

src/lib/sidenav/sidenav.ts
```diff
@@ -271,11 +274,16 @@ export class MdSidenavLayout implements AfterContentInit {
   /** TODO: internal */
   ngAfterContentInit() {
     // On changes, assert on consistency.
-    this._sidenavs.changes.subscribe(() => this._validateDrawers());
+    this._sidenavChangesSubscription = this._sidenavs.changes
```
jelbourn: This one also shouldn't be necessary because the `QueryList` is destroyed when the component is destroyed

fxck:
What about these https://github.com/angular/material2/blob/master/src/lib/sidenav/sidenav.ts#L286 @jelbourn ?

jelbourn:
@fxck `EventEmitter` should also be auto-destroyed when the component is destroyed.

fxck:
@jelbourn I know this is rather naive test http://plnkr.co/edit/vDeFQunSeXehUssR1Pzr?p=preview but both ViewChildren and Output / EventEmitter seems to still have observers on them even after the component is destroyed..

jelbourn:
@fxck they still exist in that case because you're saving references to them; if nothing references the emitter, it will be garbage collected.

fxck:
Alright then, I have removed everything but those fee `first`s.

jelbourn:
LGTM

PR #1553, opened 2016-10-20
feature(radio): Add ripple effect to radio button

tinayuangao:
Removed the ripple placeholder and added the real md-ripple

R: jelbourn

### DIFF
--- a/src/lib/radio/radio.ts
+++ b/src/lib/radio/radio.ts
@@ -3,6 +3,7 @@ import {
   Component,
   ContentChildren,
   Directive,
+  ElementRef,
   EventEmitter,
   HostBinding,
   Input,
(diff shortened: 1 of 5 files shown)


jelbourn:
CC @dozingcat if you happen to have a bit of time, could you also take a look?

src/lib/radio/radio.ts
```diff
@@ -411,10 +417,19 @@ export class MdRadioButton implements OnInit {
       this.radioGroup._touch();
     }
   }
+
+  getHostElement() {
+    return this._elementRef.nativeElement;
+  }
+
+  isRippleEnabled() {
```
tinayuangao: We can bind directly to the disableRipple property. Just referred what button did for ripple. I can change that for button too.

tinayuangao: Synced and resolved conflicts. Use coerceBooleanProperty for disableRipple.

jelbourn:
LGTM

PR #1554, opened 2016-10-21
fix rollup tips  aboat 'rxjs/add/operator/first'

mniya:
Treating 'rxjs/add/operator/first' as external dependency
No name was provided for external module 'rxjs/add/operator/first' in options.globals – guessing 'rxjs_add_operator_first'


jelbourn:
Thanks!

Issue #1135, opened 2016-08-30
[feature request] md-input focus class


sonderbazz:
#### Bug, feature request, or proposal:

Focus on md-input-element sets the md-focused class on md-input
#### What is the expected behavior?

I would like to be able to style on all sub elements within the md-input when the input has focus.
#### What is the current behavior?

Only ripple and placeholder gets a md-focused class added
#### What are the steps to reproduce?

Set focus on a md-input element check to see if its possible to style all subelements to md-input
#### What is the use-case or motivation for changing an existing behavior?

Would give more flexibility to modding the design

hansl:
There's a perfectly correct version by doing something like this:

```html
<md-input #example [class.my-custom-focus]="example.focused">
```

We do not want to export special classes only for the purpose of styling like this. If you need styling for a component you should apply your own custom classes to it.

Thanks!

sonderbazz:
Fair enough - and thanks for the simple solution. :)

PR #1337, opened 2016-09-27
feat(input): add md-focused class on wrapper

leocaseiro:
Add `.md-focused` to the `md-input-wrapper`

Fix #1135

PR #1559, opened 2016-10-21
feat(input): add `md-focused` class to `md-input` element

fxck:
Fixes https://github.com/angular/material2/issues/1135

### DIFF
--- a/src/lib/input/input.ts
+++ b/src/lib/input/input.ts
@@ -101,7 +101,10 @@ export class MdHint {
   templateUrl: 'input.html',
   styleUrls: ['input.css'],
   providers: [MD_INPUT_CONTROL_VALUE_ACCESSOR],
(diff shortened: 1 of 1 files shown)


fxck:
fails because of timeouts

jelbourn:
Not sure if I want to add this vs.:
```
<md-input #firstName [class.my-custom-focus]="firstName.focused">
```

@hansl thoughts?

fxck:
Just noticed there's a second PR as well https://github.com/angular/material2/pull/1337 with the difference that mine puts the class directly on the host element, which I think is more desirable.

hansl:
There's a completely fine solution outlined by @jelbourn above and I think the issue should be marked as Intended Behaviour. We don't want to go into making a bunch of classes that can be styled by customers, since they can set the classes they want.

PR #1562, opened 2016-10-21
feat(textarea): add md-textarea element

fxck:
I went with the having one class and multiple selector way. Also added the autosize / autoexpand directive. Modified demo page and README aswell.

I suppose some tests are needed, I'm not really good at those, so I'd appreciate assistance with that.

Should close https://github.com/angular/material2/issues/546

cc @jelbourn


fxck:
Not sure why this particular test is failing, nothing I've done is affecting the name attribute..

src/lib/input/input.ts
```diff
@@ -30,6 +32,9 @@ import {Observable} from 'rxjs/Observable';
 
 const noop = () => {};
 
+const MD_INPUT_SELECTOR = 'md-input';
+const MD_TEXTAREA_SELECTOR = 'md-textarea';
```
jelbourn: You can just use the strings inline without introducing constants; the values are obvious and aren't going to change frequently.

src/lib/input/input.html
```diff
@@ -34,6 +35,38 @@
...
+             [attr.minlength]="minlength"
+             [readonly]="readonly"
+             [required]="required"
+             [spellcheck]="spellcheck"
+             [attr.tabindex]="tabindex"
+             [attr.name]="name"
+             (focus)="_handleFocus($event)"
+             (blur)="_handleBlur($event)"
+             [(ngModel)]="value"
+             (change)="_handleChange($event)"></textarea>
```
jelbourn: You can remove min, max, and autocorrect. They don't exist on `<textarea>`

We should add `rows`, `cols`, and `wrap`. Need to think about how `rows` interacts with autosize. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

src/lib/input/input.ts
```diff
@@ -90,14 +95,46 @@ export class MdHint {
...
 
+/**
+ * Enables auto expanding behaviour for the textarea
+ */
+@Directive({
+  selector: 'textarea[mdAutosize]'
+})
+export class MdAutosize implements OnInit {
+  @HostListener('input', ['$event.target'])
+  public onChange() {
```
jelbourn: I don't see where this `onChange` method is called.

src/lib/input/input.ts
```diff
@@ -90,14 +95,46 @@ export class MdHint {
...
+
+  ngOnInit() {
+    if (this._isActive) {
+      this._adjust();
+    }
+  }
+
+  private _adjust() {
+    this._elRef.nativeElement.style.overflow = 'hidden';
+    this._elRef.nativeElement.style.height = 'auto';
```
jelbourn: These two properties don't need to be set every time.

src/lib/input/input.ts
```diff
@@ -90,14 +95,46 @@ export class MdHint {
...
+    this._elRef.nativeElement.style.height = `${this._elRef.nativeElement.scrollHeight}px`;
```
jelbourn: The logic for sizing the textarea is much more complicated in material1, which makes me think this may be missing something.

@crisbeto do you have insights on this?

src/lib/input/input.ts
fxck: Well, it was for my own convenience, but ok.

src/lib/input/input.ts
fxck: `overflow` I does not, `height = auto` does, it's what causes the textarea shrink when you remove text

src/lib/input/input.ts
fxck: It's what the `@HostListener` is calling. Or you mean it should be private? I suppose..

src/lib/input/input.ts
fxck: Seems to be working fine, seems to be supported in all browsers as well https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight

fxck: Well.. it's considerably more complex when you have to  deal with the `rows` attribute. So there are two options, not allowing autosize and rows on the same element(because honestly, who's using rows nowadays, it's only good for devices not supporting css) or go with the 10 times more complex code. I started working on the latter, but I'd be much happier with the former.

crisbeto: Correct, most of the complexity is because of the `rows`, however I think that they're a pretty important part of the component, because they're what sets an empty `textarea` apart from an `input`. In Material 1 we also do the following, although I'm not the one who wrote it so I'm not sure what the reasoning is:

```js
function getHeight() {
  var offsetHeight = node.offsetHeight;
  var line = node.scrollHeight - offsetHeight;
  return offsetHeight + Math.max(line, 0);
}
```

src/lib/input/input.ts
```diff
@@ -90,14 +94,131 @@ export class MdHint {
...
+        padding: 0px;
+        visibility: hidden;
+        opacity: 0;
+        font-family: ${computedStyles.fontFamily};
+        font-size: ${computedStyles.fontSize};
+      `
+    );
+
+    // fill in one row
+    tempEl.innerHTML = '-';
```
crisbeto: Wouldn't it be better to use the `value` property?

fxck:
@jelbourn I have modified the autosize directive to work along `rows` attribute.. it's a little complex so I'll try to explain how it works:

A row's height is calculated from `line-height` of the element, but unless the line-height is explicitly set, its initial value is `normal`, so to be able to dynamically set `rows` depending on the height of the textarea, I need to find the actual line-height.

So I'm creating a temp textarea element, reseting it's paddings and margins, visually hiding it, setting the rows attribute to `1`(default is `2`, but I'm not sure how consistent is it across browsers, so I'm setting it explicitly) and getting it's height, removing the element afterwards. When line-height is actually explicitly set(doesn't matter if it's by ems, pixels, percents, `getComputedStyles` always returns pixel values), I'm simply returning it as a number and doing none of above.

Now that I have actual `line-height`, I'm setting the `min-height` of the element to be 

`user provideded rows value` * `line-height I found`

The rest is similar to how it was before when `input` event happens, `height` is set to `auto` and `rows` to the initial value, which allows me to get the real `scrollHeight` even if user is removing instead of adding characters, afterwards new `rows` is calculated by diving the new `scrollHeight`  by `lineHeight`.

Also none of above happens unless `rows` is explicitly provided by the user.

src/lib/input/input.ts
fxck: Well since I won't ever need the actual value, it doesn't really matter.. having said that, `value` is apparently marginally faster https://jsperf.com/innerhtml-vs-value-on-textarea so yeah.

src/lib/input/input.ts
```diff
@@ -90,14 +94,131 @@ export class MdHint {
...
+  private _getLineHeight(): number {
+    const el: HTMLElement = this._elRef.nativeElement;
+
+    // get the actual computed styles for the element
+    const computedStyles = window.getComputedStyle(el);
+
+    // if line height is explicitly set(to a pixel value), use that
+    if (computedStyles.lineHeight && computedStyles.lineHeight.toLowerCase().indexOf('px') >= 0) {
+      // return stripped of the "px" and as a number
+      return +computedStyles.lineHeight.replace('px', '');
```
fxck: Yeah, that might be faster.

fxck:
I tried adding a test for the rows attribute but I can't get it to succeed.

```
  it('supports the rows attribute', () => {
    let fixture = TestBed.createComponent(MdInputTextareaWithRows);
    let input: MdInput = fixture.debugElement.query(By.directive(MdInput)).componentInstance;

    fixture.detectChanges();

    let el: HTMLInputElement = fixture.debugElement.query(By.css('textarea')).nativeElement;
    expect(el).not.toBeNull();

    fixture.detectChanges();
    expect(el.getAttribute('rows')).toEqual(null);

    input.rows = '8';

    fixture.detectChanges();
    
    expect(el.getAttribute('rows')).toEqual('8');

  });

@Component({template: `<md-textarea [rows]="rows"></md-textarea>`})
class MdInputTextareaWithRows { }

```

fails with "expected null to equal '8'" or something

jelbourn:
@fxck do you want to break the autosize into a separate PR? I want to merge the textarea, but need some more time to review/think about the autosize logic.

fxck:
Yeah I can, but I still can't figure out those tests..

jelbourn:
I don't see anything obviously wrong with the test; if you rebase the PR I can try to take a look

fxck:
@jelbourn should be rebased + autosize removed

src/lib/input/README.md
```diff
@@ -106,6 +106,17 @@ export class MyComponent implements OnInit {
 }
 ```
 
+## Textareas
+
+Supports same attributes as MdInput, except `type`, additionally it support `autosize` attribute
+which enables auto expanding functionality.
+
+```html
+<md-textarea autosize placeholder="Textarea with autosize"></md-textarea>
```
jelbourn: Should remove autosize from the readme for this PR

src/lib/input/input.scss
```diff
@@ -95,6 +95,10 @@ md-input {
   }
 }
 
+.md-input-element-textarea {
+  min-height: 60px;
+}
```
jelbourn: Where does this 60px come from? I don't see anything in the spec and it looks like material1 uses a min-height to match the line-height. Can we remove it?

src/lib/input/input.ts
```diff
@@ -229,6 +229,10 @@ export class MdInput implements ControlValueAccessor, AfterContentInit, OnChange
 
   @ViewChild('input') _inputElement: ElementRef;
 
+  public elementType: 'input' | 'textarea' = undefined;
+
+  constructor(private _elRef: ElementRef) { }
```
jelbourn: If you move the setting of `elementType` here, you don't need the `private` and can just call it `elementRef`.

src/lib/input/input.ts
```diff
@@ -282,6 +286,13 @@ export class MdInput implements ControlValueAccessor, AfterContentInit, OnChange
   ngAfterContentInit() {
     this._validateConstraints();
 
+    // Set the element type depending on normalized selector used(md-input / md-textarea)
+    if (this._elRef.nativeElement.tagName.toLowerCase() === 'md-input') {
+      this.elementType = 'input';
+    } else {
+      this.elementType = 'textarea';
+    }
```
jelbourn: You should be able to do this in the constructor. I'd also make it a bit more concise with a ternary:
```ts
this.elementType = this._elementRef.nativeElement.nodeName.toLowerCase() === 'md-input' ?
    'input' : 
    'textarea';
```

src/lib/input/input.ts
```diff
@@ -229,6 +229,10 @@ export class MdInput implements ControlValueAccessor, AfterContentInit, OnChange
 
   @ViewChild('input') _inputElement: ElementRef;
 
+  public elementType: 'input' | 'textarea' = undefined;
```
jelbourn: You can omit the `public` (as the default). Can you also prefix this with an underscore (which we're using with the absence of the `private` keyword as "internal").

src/lib/input/input.scss
fxck: I guess it should be equal to two rows, which is line-height * 2.

fxck:
I had those inputs there, must have accidentally deleted them while rebasing.. anyway addressed all comments.

@jelbourn

src/lib/input/input.ts
```diff
@@ -229,6 +234,15 @@ export class MdInput implements ControlValueAccessor, AfterContentInit, OnChange
 
   @ViewChild('input') _inputElement: ElementRef;
 
+  _elementType: 'input' | 'textarea';
+
+  constructor(public _elementRef: ElementRef) {
```
jelbourn: It can just be
```ts
 constructor(elementRef: ElementRef) {
```
Adding a `public` or `private` modifier here makes it a field on the class which isn't necessary.

fxck: Huh, can it? Changed.

fxck:
Gah, needs rebase now.. sec.

fxck:
Done, passing, all good I guess.

jelbourn:
LGTM

jelbourn:
Thanks!

Issue #1560, opened 2016-10-21
Support [md-menu-trigger-for] with anything conforming to MdMenu interface


jelbourn:
Right now, `[md-menu-trigger-for]` requires that its value be an instance of `MdMenu`:

``` html
<button md-icon-button [md-menu-trigger-for]="menu"> Open Menu </button>

<md-menu #menu="mdMenu">
  <button md-menu-item> Hello! </button>
</md-menu>
```

It would be more flexible to allow the value to be anything that conforms to the expected interface (something like `interface MdCustomMenu`).

trshafer:
I went back and forth on whether or not to make it an abstract class with many of the implementation details on the abstract class. I went with the interface because it's less invasive to the consumer than an abstract class.

PR #1564, opened 2016-10-21
Allow menu-trigger to take a menu interface.

trshafer:
This should enable menus to be more easily extended for custom menu implementations.

Addresses #1560.

### DIFF
--- a/src/lib/menu/menu-panel.ts
+++ b/src/lib/menu/menu-panel.ts
@@ -1,0 +1,10 @@
+import {EventEmitter, TemplateRef} from '@angular/core';
+import {MenuPositionX, MenuPositionY} from './menu-positions';
+
+export interface MdMenuPanel {
+  positionX: MenuPositionX;
+  positionY: MenuPositionY;
+  templateRef: TemplateRef<any>;
+  close: EventEmitter<void>;
+  focusFirstItem: () => void;
+}
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -16,6 +16,7 @@ import {MenuPositionX, MenuPositionY} from './menu-positions';
 import {MdMenuInvalidPositionX, MdMenuInvalidPositionY} from './menu-errors';
 import {MdMenuItem} from './menu-item';
 import {ListKeyManager} from '../core/a11y/list-key-manager';
+import {MdMenuPanel} from './menu-panel';
 
 @Component({
   moduleId: module.id,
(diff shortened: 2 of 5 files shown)


trshafer:
I went back and forth on whether or not to make it an abstract class with many of the implementation details on the abstract class. I went with the interface because it's less invasive to the consumer than an abstract class.

src/lib/menu/menu-interface.ts
```diff
@@ -0,0 +1,11 @@
+import {EventEmitter, TemplateRef} from '@angular/core';
+import {MenuPositionX, MenuPositionY} from './menu-positions';
+
+export interface MdMenuInterface {
+  positionX: MenuPositionX;
+  positionY: MenuPositionY;
+  templateRef: TemplateRef<any>;
+  close: EventEmitter<void>;
+  _focusFirstItem: () => void;
+  _setClickCatcher: (bool: boolean) => void;
```
jelbourn: If these two are going to be part of this interface, they'll should lose the underscore (which implies "internal api only")

trshafer:
Force pushed to address the conflict in merge of #1534.

src/lib/menu/menu-interface.ts
```diff
@@ -0,0 +1,11 @@
+import {EventEmitter, TemplateRef} from '@angular/core';
+import {MenuPositionX, MenuPositionY} from './menu-positions';
+
+export interface MdMenuInterface {
```
jelbourn: Let's call this `MdMenuPanel`
(generally trying to avoid using the word "interface" in the names of interfaces)

trshafer:
Hey @jelbourn, I have made the requested updates. Good suggestions. The test really helps demonstrate the creation of custom menus.

jelbourn:
LGTM aside from a couple last comments. @kara should also take a look

src/lib/menu/menu-panel.ts
```diff
@@ -0,0 +1,10 @@
+import {EventEmitter, TemplateRef} from '@angular/core';
+import {MenuPositionX, MenuPositionY} from './menu-positions';
+
+export interface MdMenuPanel {
+  positionX: MenuPositionX;
+  positionY: MenuPositionY;
+  templateRef: TemplateRef<any>;
+  close: EventEmitter<void>;
+  _focusFirstItem: () => void;
```
kara: Agree with @jelbourn that underscore should be removed if possible

trshafer:
Hey @kara, thanks for the review. I have removed the underscore from the interface, rebased, squashed my commits, and updated the commit message style.

Issue #1574, opened 2016-10-23
Uncaught Error: Cannot find module '@angular/http'


DAB0mB:
#### Uncaught Error: Cannot find module '@angular/http'

When installing `@angular/material` and importing it the following error is thrown:

```
Uncaught Error: Cannot find module '@angular/http'
```

Weird because `@angular/http` is registered as one of the dependencies, but I double checked the issue just to make sure and still had the same result.

fxck:
I think having it under `dependencies` in package.json doesn't actually install it for you, you still have to install it yourself. And it's not necessarily a bug either, it's basically telling you that you forgot to install it, but I guess it should be stated in GETTING_STARTED(that it's required for the icon module). I guess they assumed you wouldn't use angular2 without Http.

Urigo:
why would you need HTTP for icon?

gdi2290:
the MdIconRegistry requires it
https://github.com/angular/material2/blob/99396a47f6c4d12e55c30ab9e6e6a7d02102a3b1/src/lib/icon/icon-registry.ts#L77

https://github.com/angular/material2/blob/99396a47f6c4d12e55c30ab9e6e6a7d02102a3b1/src/lib/icon/icon.ts#L231

SamiAl90:
@Urigo because you are able to load iconfonts from URIs and you use the MdIconRegistry to register them.

DAB0mB:
OK so it just seems like a `peerDependency` field is needed

jelbourn:
Fixed in #1579, will be a part of the next release.

PR #1578, opened 2016-10-24
chore(getting started): add note about the requirement of angular/http

fxck:
related https://github.com/angular/material2/issues/1574


jelbourn:
I think this is captured now by #1579 adding http to the peer deps.

PR #1579, opened 2016-10-24
fix(#1574): '@angular/http' module not exist

DAB0mB:


DAB0mB:
I signed it!

package.json
```diff
@@ -37,6 +36,9 @@
     "systemjs": "0.19.38",
     "zone.js": "^0.6.23"
   },
+  "peerDependencies": {
```
jelbourn: This isn't the `package.json` that's used to publish. That one is in `src/lib`

jelbourn: Strike that last idea- the import paths would still be a problem.

DAB0mB:
@jelbourn whether there's a better solution or not as a temporary fix it's better than what's already published. I will make that fix you mentioned above and I suggest in the near future some of you will come up with a better solution.

jelbourn:
LGTM, thanks!

PR #1583, opened 2016-10-24
feat(overlay): set overlay size

kara:
This PR adds two properties to the overlay state to manually configure the size of the overlay.

r: @jelbourn

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -23,6 +23,7 @@ export class OverlayRef implements PortalHost {
     }
 
     let attachResult = this._portalHost.attach(portal);
+    this.updateSize();
     this.updatePosition();
 
     return attachResult;
@@ -55,6 +56,17 @@ export class OverlayRef implements PortalHost {
   updatePosition() {
     if (this._state.positionStrategy) {
       this._state.positionStrategy.apply(this._pane);
+    }
+  }
+
+  /** Updates the size of the overlay based on the overlay config. */
+  updateSize() {
+    if (this._state.width || this._state.width === 0) {
+      this._pane.style.width = formatCssUnit(this._state.width);
+    }
+
+    if (this._state.height || this._state.height === 0) {
+      this._pane.style.height = formatCssUnit(this._state.height);
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/core/overlay/overlay-state.ts
```diff
@@ -12,10 +12,16 @@ export class OverlayState {
...
   hasBackdrop: boolean = false;
 
+  /** Custom class to add to the backdrop **/
   backdropClass: string = 'md-overlay-dark-backdrop';
 
+  /** The width of the overlay panel **/
+  width: number;
+
+  /** The height of the overlay panel **/
+  height: number;
```
jelbourn: We should support percent with this as well. I'd make the type `number|string`, assume numbers are `px` and strings are whole units.

src/lib/core/overlay/overlay.spec.ts
```diff
@@ -98,6 +98,30 @@ describe('Overlay', () => {
...
+      overlay.create(state).attach(componentPortal);
+      const pane = (<HTMLElement>overlayContainerElement.children[0]);
+      expect(pane.style.width).toEqual('500px');
+    });
+
+    it('should apply the height set in the config', () => {
+      state.height = 500;
+
+      overlay.create(state).attach(componentPortal);
+      const pane = (<HTMLElement>overlayContainerElement.children[0]);
```
jelbourn: I want to start preferring the `as` cast syntax.

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -58,6 +59,17 @@ export class OverlayRef implements PortalHost {
     }
   }
 
+  /** Updates the size of the overlay based on the overlay config. */
+  updateSize() {
+    if (this._state.width) {
```
jelbourn: It should _probably_ work for zero as well.

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -98,3 +110,7 @@ export class OverlayRef implements PortalHost {
     }
   }
 }
+
+export function coerceToCSSFormat(value: number | string) {
```
jelbourn: `formatCssUnit` ?

Also it doesn't need to be exported just yet

kara:
@jelbourn Comments addressed!

PR #1584, opened 2016-10-24
feat(overlay): add ability to set custom class on panes

fxck:
I re-did https://github.com/angular/material2/pull/1438 

It includes tests for overlay(I didn't see any similar tests for say backdrop class on menu or dialog, so I assumed it's not required since it should inherently work when overlay tests passes) and comments from the previous PR with the exception of https://github.com/angular/material2/pull/1438#discussion_r84730695 and https://github.com/angular/material2/pull/1438#discussion_r84734492 which didn't make sense to me.

@jelbourn

### DIFF
--- a/src/lib/core/overlay/_overlay.scss
+++ b/src/lib/core/overlay/_overlay.scss
@@ -20,7 +20,6 @@
     left: 0;
     height: 100%;
     width: 100%;
-    z-index: $md-z-index-overlay-container;
   }
 
   // A single overlay pane.
--- a/src/lib/core/overlay/overlay-state.ts
+++ b/src/lib/core/overlay/overlay-state.ts
@@ -14,6 +14,9 @@ export class OverlayState {
 
   backdropClass: string = 'md-overlay-dark-backdrop';
 
+  /** Optional custom class to be added to dialog's overlay pane. */
+  overlayClass: string | string[];
+
   // TODO(jelbourn): configuration still to add
   // - overlay size
   // - focus trap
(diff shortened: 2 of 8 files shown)


jelbourn:
@fxck Would your specific problem be solved by simply having a higher z-index on the overlay container? (right now it's just 1 to create a new stacking context).

fxck:
No, unless I were in control of that z-index. But what might be worth doing is sticking that custom class on container rather than pane. I'm not sure what was the original reasoning for putting it on panes, it's been a while.

fxck:
I see, the reason why I added it to `pane` instead of `container` was that there is only one `container`. 

Then there's this ..

image

..when you open, menu from inside a dialog, the stacking order is not correct, dialog itself it above menu's backdrop. 

I think is issue is a bit deeper, there should be some sort of stacking order manager.

@jelbourn @kara

crisbeto:
Are there any plans to continue with this PR? Something similar (being able to add a class to the `.overlay-pane`) should help with an issue where the tooltip can sometimes become hidden immediately.

fxck:
I never got the reply about stacking manager. I think adding classes to either pane or overlay would be only a temporary fix and more issues will come up sooner or later. But I don't want to do anything without having a discussion.

jelbourn:
This has been on the backburner since I've been wanting to spend some more time thinking about how stacking should work in general.

fxck:
Stacking aside, this still could be useful, currently there's no way to properly style md-menu(stuff like `width`).

fxck:
I take it back, there is, it has to be global though.

mmalerba:
@jelbourn @fxck what's the status of this should it be rebased? closed?

fxck:
for what it was, this PR is obsolete, altho most of my points stands, there still are cases when inability to control z-index of panes and overlays still can cause problems, as well as I think there needs to be a class that handles z-index of various elements.

mmalerba:
@fxck ok will close this PR then, if you have concerns about z-index that aren't already being tracked by an issue, feel free to open a new issue to discuss them.

Issue #1545, opened 2016-10-20
gulp commands log no output


fxck:
At the current master(fresh clone), doing `npm run demo-app` prints this 

```
Treating 'rxjs/add/operator/first' as external dependency
material2 Admin$ No name was provided for external module 'rxjs/add/operator/first' in options.globals – guessing 'rxjs_add_operator_first'
```

then nothing, when I fix it(https://github.com/angular/material2/pull/1529), it just doesn't do anything at all no success message, no error message, nothing.

Using osx with these versions of gulp and node:

```
material2 Admin$ gulp -v
[18:41:05] CLI version 3.9.1
[18:41:05] Local version 3.9.1
material2 Admin$ node -v
v6.5.0
```

![image](https://cloud.githubusercontent.com/assets/1303561/19569162/0aa631b8-96f5-11e6-97bf-0ad4fe7fd8d8.png)

fxck:
Ok, now it suddenly started throwing errors, this was in the log

```
0 info it worked if it ends with ok
1 verbose cli [ '/usr/local/bin/node', '/usr/local/bin/npm', 'run', 'demo-app' ]
2 info using npm@3.8.9
3 info using node@v6.5.0
4 verbose run-script [ 'predemo-app', 'demo-app', 'postdemo-app' ]
5 info lifecycle material2@2.0.0-alpha.9-3~predemo-app: material2@2.0.0-alpha.9-3
6 silly lifecycle material2@2.0.0-alpha.9-3~predemo-app: no script for predemo-app, continuing
7 info lifecycle material2@2.0.0-alpha.9-3~demo-app: material2@2.0.0-alpha.9-3
8 verbose lifecycle material2@2.0.0-alpha.9-3~demo-app: unsafe-perm in lifecycle true
9 verbose lifecycle material2@2.0.0-alpha.9-3~demo-app: PATH: /usr/local/lib/node_modules/npm/bin/node-gyp-bin:/Users/Admin/www/material2/node_modules/.bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
10 verbose lifecycle material2@2.0.0-alpha.9-3~demo-app: CWD: /Users/Admin/www/material2
11 silly lifecycle material2@2.0.0-alpha.9-3~demo-app: Args: [ '-c', 'gulp serve:devapp' ]
12 silly lifecycle material2@2.0.0-alpha.9-3~demo-app: Returned: code: 1  signal: null
[...]

PR #1586, opened 2016-10-24
Log the port the demo app is running on

trshafer:
Hello Angular Material2,

I was running the demo app and had to dig in the source to find out what was going on. The demo app was running but it was not immediately obvious. This change logs a message to developers where the demo app is running.

Thanks!


jelbourn:
This is actually caused by #1545

PR #1589, opened 2016-10-24
feat(tab-nav-bar): add initial functionality of tab nav bar

andrewseguin:
Initial development of the tab nav bar


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -0,0 +1,40 @@
+import { Component, Input, ViewChild, ElementRef } from '@angular/core';
+import { MdInkBar } from '../ink-bar';
```
jelbourn: No spaces inside imports (this might be a WebStorm settings)

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-tab-nav-bar',
+  templateUrl: 'tab-nav-bar.html',
+  styleUrls: ['tab-nav-bar.css'],
+})
+export class MdTabNavBar {
+  @ViewChild(MdInkBar) private _inkBar: MdInkBar;
```
jelbourn: `@ViewChild` can't be applied to a private member (since Angular itself is external to this class and has to touch it).

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  updateActiveLink(element: HTMLElement) {
```
jelbourn: Add function description comment

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -0,0 +1,40 @@
+import { Component, Input, ViewChild, ElementRef } from '@angular/core';
+import { MdInkBar } from '../ink-bar';
+
+@Component({
```
jelbourn: Add class description for what this component does (since it's not one of the totally obvious ones like radio-button).

src/lib/tabs/tabs-common.scss
```diff
@@ -0,0 +1,51 @@
+@import '../core/style/variables';
+
+$md-tab-bar-height: 48px !default;
+
+
+@mixin tab-label {
```
jelbourn: Add comment explaining what the mixin is for

src/lib/tabs/tab-nav-bar/tab-nav-bar.html
```diff
@@ -0,0 +1,4 @@
+<div class="md-tab-header">
```
jelbourn: I'm thinking that this should probably be a `<nav>` element instead of a div. 

@marcysutton Is it appropriate to always use a `<nav>` element for a component that contains anchors and navigates the page, or should `<nav>` only be used for _major_ site navigation?

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -0,0 +1,40 @@
...
+  moduleId: module.id,
+  selector: '[md-tab-link]',
+  template: '<ng-content></ng-content>',
+  styleUrls: ['tab-link.css']
+})
+export class MdTabLink {
+  private _isActive: boolean = false;
+
+  @Input()
+  get active(): boolean {
```
jelbourn: I didn't know about this before, but looking through the router docs turned up this:
https://angular.io/docs/ts/latest/guide/router.html#!#router-link-active

Seems like we could rely on that to apply the appropriate class based on the route being active

src/demo-app/tabs/routes.ts
```diff
@@ -0,0 +1,10 @@
+import {Routes} from '@angular/router';
+
+import {RoutedContent1, RoutedContent2, RoutedContext3} from '../tabs/tabs-demo';
+
+export const TABS_DEMO_ROUTE: Routes = [
```
andrewseguin: Done.

src/demo-app/tabs/routes.ts
```diff
@@ -0,0 +1,10 @@
...
+  { path: '', redirectTo: 'content-1', pathMatch: 'full' },
+  { path: 'content-1', component: RoutedContent1 },
+  { path: 'content-2', component: RoutedContent2 },
+  { path: 'content-3', component: RoutedContext3 },
```
andrewseguin: Done.

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
andrewseguin: Done.

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
andrewseguin: Done.

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
andrewseguin: Done.

src/demo-app/tabs/routes.ts
```diff
@@ -0,0 +1,10 @@
+import {Routes} from '@angular/router';
+
+import {SunnyTabContent, RainyTabContent, FoggyTabContent} from '../tabs/tabs-demo';
+
+export const TABS_DEMO_ROUTES: Routes = [
+  { path: '', redirectTo: 'sunny-tab', pathMatch: 'full' },
+  { path: 'sunny-tab', component: SunnyTabContent },
+  { path: 'rainy-tab', component: RainyTabContent },
+  { path: 'foggy-tab', component: FoggyTabContent },
```
jelbourn: Remove the leading/trailing spaces in object literals

src/lib/tabs/tab-nav-bar/tab-nav-bar.scss
```diff
@@ -0,0 +1,10 @@
+@import '../tabs-common';
+
+:host {
+  display: block;
+}
+
+a[md-tab-link] {
```
jelbourn: I'd remove the `a` part of this selector just to reduce the specificity.

src/lib/tabs/tabs-common.scss
```diff
@@ -0,0 +1,51 @@
+@import '../core/style/variables';
```
jelbourn: This file should either consist of exclusively mixins or exclusively concrete styles. 

If it's the former, then the file name should start with an underscore (which tells the sass compiler that it's a "partial" file that doesn't have output).

If it's the latter, it should be included in the "styleUrls" for each of the components that use it (rather than doing a sass `@import`)

andrewseguin:
Thanks for the review, I made the tabs common to be purely mixins

src/lib/tabs/tab-nav-bar/tab-nav-bar.html
andrewseguin: Thanks for the input Marcy. Jeremy and I discussed this and agree that we can move this responsibility to the user by making this a component with an attribute selector.

Usage of this component will be written by the user as such:
`<nav md-tab-nav-bar aria-label="navigation links"> ... </nav>`

jelbourn:
LGTM- nice work!

PR #1591, opened 2016-10-25
feat(connected-overlay): support all overlay config properties

kara:
This PR passes all the overlay config properties through to the connected overlay directive.

r: @jelbourn

### DIFF
--- a/src/lib/core/overlay/overlay-directives.ts
+++ b/src/lib/core/overlay/overlay-directives.ts
@@ -2,11 +2,12 @@ import {
     NgModule,
     ModuleWithProviders,
     Directive,
+    EventEmitter,
     TemplateRef,
     ViewContainerRef,
-    OnInit,
     Input,
     OnDestroy,
+    Output,
     ElementRef
 } from '@angular/core';
 import {Overlay, OVERLAY_PROVIDERS} from './overlay';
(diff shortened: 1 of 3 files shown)


src/lib/core/overlay/overlay-directives.ts
```diff
@@ -50,15 +52,51 @@ export class OverlayOrigin {
...
+  @Input() backdropClass: string;
+
+  /** Whether or not the overlay should attach a backdrop. */
+  @Input()
+  get hasBackdrop() {
+    return this._hasBackdrop;
+  }
+
+  set hasBackdrop(value: any) {
+    this._hasBackdrop = value === '' || (!!value && value !== 'false');
```
jelbourn: Should this be the same logic as here?
https://github.com/jelbourn/material2/blob/a70e3e569d026aaf4c0ad8728d9b6091f8673901/src/lib/core/coersion/boolean-property.ts#L3

src/lib/core/overlay/overlay-directives.ts
```diff
@@ -50,15 +52,51 @@ export class OverlayOrigin {
...
+    return this._open;
+  }
+
+  set open(value: boolean) {
+    value ? this._attachOverlay() : this._detachOverlay();
+    this._open = value;
+  }
+
+  /** Event emitted when the backdrop is clicked. */
+  @Output() backdropClick = new EventEmitter();
```
jelbourn: Doesn't this require a generic type?

kara: Yeah, I think it does. It's not throwing, so maybe it's implicitly typed to `EventEmitter<void>` when set this way?

src/lib/core/overlay/overlay-directives.ts
jelbourn: Not in master yet; I'd copy the logic and add a TODO

PR #1593, opened 2016-10-25
feat(overlay): overlay rtl support

kara:
Blocked on #1591 (needs rebase). Adds RTL support to overlay ref and position strategy.  APIs need discussion.

r: @jelbourn

### DIFF
--- a/src/lib/core/overlay/overlay-directives.ts
+++ b/src/lib/core/overlay/overlay-directives.ts
@@ -5,6 +5,7 @@ import {
     EventEmitter,
     TemplateRef,
     ViewContainerRef,
+    Optional,
     Input,
     OnDestroy,
     Output,
@@ -18,6 +19,7 @@ import {ConnectionPositionPair} from './position/connected-position';
 import {PortalModule} from '../portal/portal-directives';
 import {ConnectedPositionStrategy} from './position/connected-position-strategy';
 import {Subscription} from 'rxjs/Subscription';
+import {Dir, LayoutDirection} from '../rtl/dir';
 
 /** Default set of positions for the overlay. Follows the behavior of a dropdown. */
 let defaultPositionList = [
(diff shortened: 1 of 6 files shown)


src/lib/core/overlay/overlay-directives.ts
```diff
@@ -103,14 +105,19 @@ export class ConnectedOverlayDirective implements OnDestroy {
...
+      viewContainerRef: ViewContainerRef,
+      @Optional() private _dir: Dir) {
     this._templatePortal = new TemplatePortal(templateRef, viewContainerRef);
   }
 
   get overlayRef(): OverlayRef {
     return this._overlayRef;
   }
 
+  get dir(): 'ltr' | 'rtl' {
```
jelbourn: Type is `LayoutDirection` from `dir.ts`\

Do you need this getter at all, though, instead of just saying `this._dir.value`? It defaults to `ltr` in `Dir` itself.

src/lib/core/overlay/overlay-state.ts
```diff
@@ -21,6 +21,9 @@ export class OverlayState {
   /** The height of the overlay panel. If a number is provided, pixel units are assumed. **/
   height: number | string;
 
+  /** The direction of the text in the overlay panel. */
+  direction: 'ltr' | 'rtl' = 'ltr';
```
jelbourn: Same here for `LayoutDirection`

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -59,6 +60,11 @@ export class OverlayRef implements PortalHost {
     }
   }
 
+  /** Updates the text direction of the overlay panel. **/
+  private updateDirection() {
+    this._pane.style.direction = this._state.direction;
```
jelbourn: Instead of setting the css `direction` property, it should probably use do
```ts
this._pane.setAttribute('dir', this._state.direction);
```

src/lib/core/overlay/overlay-directives.ts
kara: Ah, didn't know about that type. Handy.  I think I still need it in case someone isn't using `dir`

jelbourn:
LGTM

PR #1598, opened 2016-10-25
Fix (slider): Update slider position and ticks when the min and max

### LINKED ISSUE #1556: md-slider not updating when min or max is changed
Bug

When change the max value of the slider it should update the thumb position in the slider.

When the max value is changed the slider is not updated.

<md-slider [(ngModel)]="sliderValue" min="{{sliderValueMin}}" max="{{sliderValueMax}}" value="{{sliderValue}}"></md-slider>

When sliderValueMax or sliderValueMin are changed nothing happens.

When some conditions on a form happen and need to change the limits of a slider.

Angular 2.0.0, Material 2.0.0-alpha.8-2, All, All

mmalerba:
values change. Closes #1556

### DIFF
--- a/src/lib/slider/slider.spec.ts
+++ b/src/lib/slider/slider.spec.ts
@@ -329,6 +329,9 @@ describe('MdSlider', () => {
       expect(trackFillDimensions.width).toBe(sliderDimensions.width * 6 / 8);
       expect(tickContainerDimensions.width)
           .toBe(sliderDimensions.width - sliderDimensions.width * 6 / 8);
+      expect(tickContainerElement.style.background)
+          .toEqual(`repeating-linear-gradient(to right, black, black 2px, transparent 2px, ` +
+              `transparent ${sliderDimensions.width * 6 / 8 - 1}px)`);
     });
 
     it('should adjust thumb and ticks when max changes', () => {
@@ -344,6 +347,9 @@ describe('MdSlider', () => {
       expect(trackFillDimensions.width).toBe(sliderDimensions.width * 6 / 12);
       expect(tickContainerDimensions.width)
           .toBe(sliderDimensions.width - sliderDimensions.width * 6 / 12);
+      expect(tickContainerElement.style.background)
+          .toEqual(`repeating-linear-gradient(to right, black, black 2px, transparent 2px, ` +
+              `transparent ${sliderDimensions.width * 6 / 12 - 1}px)`);
     });
   });
 
--- a/src/lib/slider/slider.spec.ts
+++ b/src/lib/slider/slider.spec.ts
@@ -330,7 +330,7 @@ describe('MdSlider', () => {
       expect(tickContainerDimensions.width)
           .toBe(sliderDimensions.width - sliderDimensions.width * 6 / 8);
       expect(tickContainerElement.style.background)
-          .toEqual(`repeating-linear-gradient(to right, black, black 2px, transparent 2px, ` +
+          .toContain(`repeating-linear-gradient(to right, black, black 2px, transparent 2px, ` +
               `transparent ${sliderDimensions.width * 6 / 8 - 1}px)`);
     });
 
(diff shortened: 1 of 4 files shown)


src/lib/slider/slider.ts
```diff
@@ -265,14 +266,28 @@ export class MdSlider implements AfterContentInit, ControlValueAccessor {
    */
   snapThumbToValue() {
     this.updatePercentFromValue();
-    this._renderer.updateThumbAndFillPosition(this._percent, this._sliderDimensions.width);
+    if (this._sliderDimensions) {
```
jelbourn: In which cases does `this._sliderDimensions` not exist?

src/lib/slider/slider.ts
```diff
@@ -425,8 +440,10 @@ export class SliderRenderer {
    * Draws ticks onto the tick container.
    */
   drawTicks(tickSeparation: number) {
+    let sliderTrackContainer =
+        <HTMLElement>this._sliderElement.querySelector('.md-slider-track-container');
+    let tickContainerWidth = sliderTrackContainer.getBoundingClientRect().width;
```
jelbourn: Why does the size calculation for the tick container change?

src/lib/slider/slider.ts
```diff
@@ -444,6 +461,8 @@ export class SliderRenderer {
     // separation), don't show it by decreasing the width of the tick container element.
     if (tickContainerWidth % tickSeparation < (tickSeparation / 2)) {
       tickContainer.style.width = tickContainerWidth - tickSeparation + 'px';
+    } else {
+      tickContainer.style.width = '';
```
jelbourn: Add comment here like
```ts
// If there is enough space for the second-to-last tick, restore the default width of the tick container.
```

src/lib/slider/slider.ts
```diff
@@ -265,14 +266,28 @@ export class MdSlider implements AfterContentInit, ControlValueAccessor {
...
+    if (this._sliderDimensions) {
+      this._renderer.updateThumbAndFillPosition(this._percent, this._sliderDimensions.width);
+    }
+  }
+
+  /**
+   * Initializes the tick and slider positions.
+   * @private
+   */
+  private _initTicksAndSlider() {
```
jelbourn: In this case I think it would be better to call the individual functions in the setters / init because it's more clear in those locations what's happening; you won't know what `_initTicksAndSlider` does precisely without jumping to the definition.

src/lib/slider/slider.ts
mmalerba: When min or max is set before ngAfterContentInit.

src/lib/slider/slider.ts
mmalerba: It was checking the width of the same container whose width it was setting. This wasn't a problem before since it never changed after initialization, but now it can change.

src/lib/slider/slider.spec.ts
```diff
@@ -311,7 +316,33 @@ describe('MdSlider', () => {
...
+    it('should adjust thumb and ticks when min changes', () => {
+      testComponent.min = -2;
+      fixture.detectChanges();
 
+      let trackFillDimensions = trackFillElement.getBoundingClientRect();
+      let tickContainerDimensions = tickContainerElement.getBoundingClientRect();
+
+      expect(trackFillDimensions.width).toBe(sliderDimensions.width * 6 / 8);
+      expect(tickContainerDimensions.width)
+          .toBe(sliderDimensions.width - sliderDimensions.width * 6 / 8);
```
jelbourn: I believe we also need to check that the linear gradient background for the ticks has been updated (since the tick interval will have changed).

mmalerba: Done

jelbourn:
Looks like there are some test failures on CI:
https://travis-ci.org/angular/material2/jobs/170800244

jelbourn:
LGTM

PR #1600, opened 2016-10-25
chore(menu): move common styles into core

kara:
r: @jelbourn


src/lib/core/style/_menu-common.scss
```diff
@@ -0,0 +1,45 @@
+@import './variables';
+@import './elevation';
+@import './list-common';
+
```
jelbourn: Add top-level comment here saying that these styles are shared between menu and select.

PR #1604, opened 2016-10-25
chore(build): fix watch to re-build components

kara:
Currently the demo app does not update if the changes occur in the components (only happens if the demo app itself is updated).  This PR fixes it.


jelbourn:
LGTM

kara:
There is probably a better/leaner way to do this, but this will at least get it working for dev. @hansl can look when he gets back, maybe?

PR #1612, opened 2016-10-26
fix(slide-toggle): match theme with specs

### LINKED ISSUE #1568: Theme colors incorrect for md-slide-toggle
Bug

md-slide-toggle obeys the Material Design spec

image
image

md-slide-toggle uses background color for `.md-slide-toggle-thumb`, and possibly incorrect in other areas as well. This is especially noticeable in dark theme, where thumb is very hard to see.

image

Look at md-slide-toggle in both light/dark themes

Bug

"@angular/material": "^2.0.0-alpha.9-3",

No

DevVersion:
* Currently the theme stylesheet for the slide-toggle was not following the specs.

* This commit ensures that the stylesheet follows the specs with all details (https://material.google.com/components/selection-controls.html#selection-controls-switch)
 

| --- | --- |
| --- | --- |
|  |  |
| | 

@jelbourn

-  I assume the way the bar color currently is determined is not looking very good (but it's per specs and I don't wanted to add it to the foreground palette)
- Regarding passing `$thumb-on-hue` - we can't place mixins inside of mixins (so I'd rather make a follow up and discuss possible approaches)

Fixes #1568.

### DIFF
--- a/src/lib/slide-toggle/_slide-toggle-theme.scss
+++ b/src/lib/slide-toggle/_slide-toggle-theme.scss
@@ -2,22 +2,22 @@
 @import '../core/theming/theming';
 
 
-@mixin _md-slide-toggle-checked($palette, $thumb-on-hue) {
+@mixin _md-slide-toggle-checked($palette, $thumb-checked-hue) {
   // Do not apply the checked colors if the toggle is disabled, because the specificity would be to high for
   // the disabled styles.
   &.md-checked:not(.md-disabled) {
     .md-slide-toggle-thumb {
-      background-color: md-color($palette, $thumb-on-hue);
+      background-color: md-color($palette, $thumb-checked-hue);
     }
 
     .md-slide-toggle-bar {
-      background-color: md-color($palette, $thumb-on-hue, 0.5);
+      background-color: md-color($palette, $thumb-checked-hue, 0.5);
     }
   }
 }
 
 // TODO(jelbourn): remove this when the real ripple has been applied to slide-toggle.
-@mixin _md-slide-toggle-ripple($palette, $foreground, $thumb-on-hue) {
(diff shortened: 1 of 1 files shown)


src/lib/slide-toggle/_slide-toggle-theme.scss
```diff
@@ -60,20 +68,18 @@
...
     }
     .md-slide-toggle-bar {
-      background-color: md-color($foreground, divider);
+      background-color: $bar-disabled-color;
     }
   }
 
   .md-slide-toggle-thumb {
-    background-color: md-color($background, background);
+    background-color: md-color($md-grey, $thumb-off-hue);
```
jelbourn: `$thumb-off-hue` is only used in one place, so I would do this here instead:
```scss
$thumb-background: if($is-dark, md-color(...), md-color(...));
```

src/lib/slide-toggle/_slide-toggle-theme.scss
```diff
@@ -2,55 +2,63 @@
...
 
 @mixin md-slide-toggle-theme($theme) {
+  $is-dark: map_get($theme, is-dark);
   $primary: map-get($theme, primary);
   $accent: map-get($theme, accent);
   $warn: map-get($theme, warn);
   $background: map-get($theme, background);
   $foreground: map-get($theme, foreground);
 
+  $thumb-on-hue: if($is-dark, 200, 500);
```
jelbourn: How about `$thumb-checked-hue`?

jelbourn: Can you add a comment here with a link to the spec and explaining the use of different hues based on light vs. dark theme?

jelbourn:
LGTM

PR #1614, opened 2016-10-26
fix(overlay): raise z-index for overlay-container

jelbourn:
R: @kara 
This will help prevent the overlay container from being covered up by other stacking contexts that are directly on the document body.

(of course there's always a _higher_ z-index...)

### DIFF
--- a/src/lib/core/style/_variables.scss
+++ b/src/lib/core/style/_variables.scss
@@ -15,8 +15,12 @@ $z-index-fab: 20 !default;
 $z-index-drawer: 100 !default;
 
 // Overlay z indices.
+
+// We want overlays to always appear over user content, so set a baseline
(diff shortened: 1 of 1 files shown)


fxck:
I hope this is not a replacement for https://github.com/angular/material2/pull/1584#issuecomment-255947612

Otherwise good luck this that https://github.com/search?utf8=%E2%9C%93&q=%22z-index%3A+9999%22&type=Code&ref=searchresults ..

jelbourn:
Not a replacement, but should generally make things better. Though, I still don't want to remove this z-index in #1584 because it's important that the overlay container always create a new stacking context.

fxck:
Read the comment. I agree, I proposed moving the custom class from pane to container as well as changing z-indexes similar to what you just did.

PR #1618, opened 2016-10-26
Fix(slider): Add support for change event.

### LINKED ISSUE #1070: [md-slider @alpha 7-3] Change detection not working

Bug I think.

When the slider value changes, it should emit a `(change)` event the way the native input range does

Moving the `md-slider` does not emit the `(change)` event
- [Open this plunker](http://plnkr.co/edit/IOLgz9F7OiI03xUcWZ5v?p=preview)
- Move both the md-slider and the native input range. Notice that only the latter's changes trigger updates to the view

Since `md-slider` functionally replaces the native `input type='range'`, I would expect their behaviors to be compatible

`Angular 2 rc.5` with `Material 2 alpha 7-3`

mmalerba:
Also clean up test slide event test code. Closes #1070

### DIFF
--- a/src/lib/slider/slider.spec.ts
+++ b/src/lib/slider/slider.spec.ts
@@ -26,6 +26,7 @@ describe('MdSlider', () => {
         SliderWithTwoWayBinding,
         SliderWithValueSmallerThanMin,
         SliderWithValueGreaterThanMax,
+        SliderWithChangeHandler,
       ],
       providers: [
         {provide: HAMMER_GESTURE_CONFIG, useFactory: () => {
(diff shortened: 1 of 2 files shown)


src/lib/slider/slider.spec.ts
```diff
@@ -884,81 +929,99 @@ class SliderWithValueSmallerThanMin { }
...
+class SliderWithChangeHandler {
+  onChange() { }
+}
+
 /**
  * Dispatches a click event from an element.
  * Note: The mouse event truncates the position for the click.
- * @param element The element from which the event will be dispatched.
+ * @param trackElement The track element from which the event location will be calculated.
+ * @param containerElement The container element from which the event will be dispatched.
```
jelbourn: Instead of always passing in both the track element and the root slider element, why not just say the argument is `sliderElement` and `querySelector` for the trackElement?

mmalerba: Done

jelbourn:
LGTM

PR #1619, opened 2016-10-26
fix(input): floating placeholder on by default

jelbourn:
R: @tinayuangao 

I accidentally set this to false in an earlier PR.


tinayuangao:
LGTM

PR #1620, opened 2016-10-26
fix: correct EventEmitter generic type across lib

jelbourn:
R: @tinayuangao 
CC: @kara 

The default generic type of `EventEmitter` is `{}`. If TS has `strictNullChecks` enabled, emitting `null` on such an emitter will cause a build error:
```
Type 'EventEmitter<{}>' is not assignable to type 'EventEmitter<null>'
```


tinayuangao:
LGTM

PR #1623, opened 2016-10-27
fix(ripple): disable pointer events on ripple

jelbourn:
R: @tinayuangao 

I noticed on the checkbox demo that you could sometimes click outside of the checkbox and on the ripple and it would toggle the checkbox.


tinayuangao:
LGTM

andrewseguin:
It seems that this breaks our ripple demo on the devapp. Clicking on the large "Click me" div no longer registers a click because pointer events are not registering.

PR #1631, opened 2016-10-27
fix(checkbox) should be disabled when *any* disabled value makes the component disabled, *except* for false.

dharmeshpipariya:
fix(checkbox) should be disabled when *any* disabled value makes the component disabled, *except* for false.


jelbourn:
LGTM

Issue #1570, opened 2016-10-22
[spinner @alpha 9] AoT compilation causes high CPU use via requestAnimationFrame()


RoxKilly:
#### Bug, feature request, or proposal:

Bug
#### What is the expected behavior?
- `md-spinner` should not use more CPU when app is AoT compiled than when it is JiT compiled
- `md-spinner` should not use CPU resources once it is taken out of the DOM
#### What is the current behavior?

App uses AoT compilation (JiT is not affected) and has `<md-spinner>` in the template
- High CPU usage even once the spinner is removed from the DOM
- `Zonetask.invoke` calls [`window.requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) in an infinite loop, choking the CPU
- Minimizing the browser brings CPU utilization down. Restoring does the opposite
#### What are the steps to reproduce?

**app.component.html**

```
<md-card-content *ngIf="loading">
    <md-spinner></md-spinner>
</md-card-content>
<md-card-content *ngIf="!loading">
    Content goes here
</md-card-content>
```

`loading===true` when `AppComponent` loads, so the spinner is shown. In `ngOnInit`, after data fetched from the server is received, `loading=false`.  The spinner disappears but CPU utilization remains high.  This only affects AoT app (not in-browser compiled JiT app).

I replaced

```
<md-spinner></md-spinner>
```

with...

```
<md-progress-bar mode="indeterminate"></md-progress-bar>
```

and I no longer experience the problem.
#### What is the use-case or motivation for changing an existing behavior?

This makes `md-spinner` too costly to use with `AoT compilation`
#### Which versions of Angular, Material, OS, browsers are affected?

Tested on `Angular 2.0.1`, `@angular/compiler-cli 0.6.4`, `Material 9-3`, `Firefox 49`, `Chrome 54`, `Windows 7 x64`
#### Is there anything else we should know?

Get more details [from this issue](https://github.com/angular/angular/issues/12253#issuecomment-254139552)

ivoviz:
@crisbeto @RoxKilly The PR might just solve the problem I reported long ago (#1283) - just saying, since it's not referenced in the PR itself, but probably can be closed after it's merged.

crisbeto:
Yeah, let's see, I didn't mark it as "fixes" on purpose since it may not fix it completely.

jelbourn:
[...]

Issue #1511, opened 2016-10-15
md-spinner has very high CPU utilization


shanemgrey:
#### Bug, feature request, or proposal:

Bug
#### What is the expected behavior?

The spinner should not elevate CPU usage so much
#### What is the current behavior?

One spinner produces about 27% CPU utilization on Google Chrome Helper or 35% in Firefox.  Add more spinners and each adds 7 to 10%.
#### What are the steps to reproduce?

Sample code from the sample app can be used.
Can be see in chrome on the demo app https://material2-app.firebaseapp.com/ 

Remove this code and CPU use is is normal. 
`<md-spinner class="app-spinner"></md-spinner>`

Add it back and CPU spikes.
#### Which versions of Angular, Material, OS, browsers are affected?

Angular 2.0.2
AngularMaterial2 2.0.0-alpha.9-3
All browsers
#### Anything else?

Looking at the DOM in chrome while the spinner is active, the calculations behind it seem to be going at full speed, and to a high precision.   Probably just need to delay the recalculations.

fxck:
Can confirm, it eats buttload of CPU for me as well.

jelbourn:
@josephperrott I'm starting to think we should go back to doing the svg animation by default and then do a off-spec animation for IE11.

crisbeto:
Instead of doing an off-spec animation, why not use the JS-based approach as a fallback @jelbourn? It shouldn't be a big issue, apart from doing the feature detection properly.  I'm not sure whether you can detect whether an element is actually animating, but you could sniff some of the IE-specific global variables.

Also note that Edge doesn't support transitions on SVG either.

jelbourn:
Closing this form work on #1635

PR #1635, opened 2016-10-27
perf(progress-circle): improved rendering performance

crisbeto:
Currently the progress circle hits ZoneJS on every animation frame, due to `requestAnimationFrame`. This change:
- Wraps the `requestAnimationFrame` calls in `runOutsideAngular` in order to avoid hitting the change detection on every frame.
- Switches from using an attribute binding to manipulating the `path` node directly.

Referencing #1570, #1511.

**Note:** I'm open to suggestions on better ways to fetch a reference to the `path` node. I've tried doing it in the constructor and in an `OnInit` method, but certain animations fire too early for the element to be rendered.

### DIFF
--- a/src/lib/progress-circle/progress-circle.ts
+++ b/src/lib/progress-circle/progress-circle.ts
@@ -6,7 +6,9 @@ import {
   ChangeDetectorRef,
   ChangeDetectionStrategy,
   OnDestroy,
-  Input
+  Input,
+  ElementRef,
+  NgZone
 } from '@angular/core';
 
 // TODO(josephperrott): Benchpress tests.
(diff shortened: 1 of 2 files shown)


src/lib/progress-circle/progress-circle.ts
```diff
@@ -210,6 +210,23 @@ export class MdProgressCircle implements OnDestroy {
   private _cleanupIndeterminateAnimation() {
     this.interdeterminateInterval = null;
   }
+
+  /**
+   * Renders the arc onto the SVG element. Proxies `getArc` while setting the proper
+   * DOM attribute on the `<path>`.
+   * @param {number} currentValue Current value to be rendered.
+   * @param {number} rotation Rotation of the circle.
```
jelbourn: Remove the types from the JsDoc, since they're just part of the code.

crisbeto: Done.

PR #1638, opened 2016-10-28
feat(sidenav): emit event when backdrop is clicked

### LINKED ISSUE #1427: Freature Request: callback when backdrop is clicked for md-sidenav

Feature Request.

I would like to be called back when the overlay for `md-sidenav` is clicked.

There is no callback available.

N/A

I have a layout where I display a sub-pane using a sidenav. This is displayed in a router outlet within the sidenav component. Clicking a button on the main page changes the route and opens the sidenav to display an edit screen/popup. I want to be called back when the user clicks on the sidenav overlay, so that I can change the route back to the main page's route. Using the `onCloseStart` and `onClose` events on the sidenav itself is not really right for this.

N/A

I would be more than happy to put up a pull-request for this. I would like to ask the community whether they would want control via the callback whether the sidenav should close or not, or whether just an informational callback (e.g. event) to inform that sidenav will be closing is enough (the latter is fine for my use-case).

ianjoneill:
This allows clients to distinguish between close events caused by calling close() and those caused by the backdrop being clicked.

Closes #1427

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -293,6 +293,9 @@ export class MdSidenavLayout implements AfterContentInit {
   get start() { return this._start; }
   get end() { return this._end; }
 
+  /** Event emitted when the sidenav backdrop is clicked. */
+  @Output('backdrop-clicked') onBackdropClicked = new EventEmitter<void>();
+
   /** The sidenav at the start/end alignment, independent of direction. */
   private _start: MdSidenav;
   private _end: MdSidenav;
@@ -398,6 +401,8 @@ export class MdSidenavLayout implements AfterContentInit {
   }
 
   _closeModalSidenav() {
+    this.onBackdropClicked.emit(null);
+
     if (this._start != null && this._start.mode != 'side') {
       this._start.close();
     }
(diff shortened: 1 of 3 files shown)


ianjoneill:
I resigned the CLA for my additional email address.

ianjoneill:
Rebased onto current master branch to get alpha-10 goodness and resolve a merge conflict.

src/lib/sidenav/sidenav.ts
```diff
@@ -398,6 +401,8 @@ export class MdSidenavLayout implements AfterContentInit {
   }
 
   _closeModalSidenav() {
+    this.onBackdropClicked.emit(null);
```
mmalerba: You don't need to pass null, can just call emit()

src/lib/sidenav/sidenav.ts
```diff
@@ -398,6 +401,8 @@ export class MdSidenavLayout implements AfterContentInit {
   }
 
   _closeModalSidenav() {
```
mmalerba: I would create a new method _onBackdropClicked that emits and then calls this, we will probably add other bindings that call this function later (e.g. esc key).

ianjoneill: Done

src/lib/sidenav/sidenav.ts
ianjoneill: Ah yes, looks like that parameter was made optional in Angular 2 RC5.

Done

PR #1644, opened 2016-10-28
feat(tooltip): add tooltip animations

andrewseguin:
Adding animations to the tooltip

### DIFF
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -7,7 +7,7 @@ import {MdTooltipModule} from './tooltip';
 
 const initialTooltipMessage = 'initial tooltip message';
 
-describe('MdTooltip', () => {
+fdescribe('MdTooltip', () => {
   let overlayContainerElement: HTMLElement;
 
 
@@ -49,12 +49,27 @@ describe('MdTooltip', () => {
       fixture.detectChanges();
       expect(overlayContainerElement.textContent).toContain(initialTooltipMessage);
 
+      // After hide called, a timeout delay is created that will to hide the tooltip.
       tooltipDirective.hide();
       expect(tooltipDirective._isTooltipVisible()).toBe(true);
 
-      // After hidden, expect that the tooltip is not visible.
+      // After the tooltip delay elapses, expect that the tooltip is not visible.
       tick(TOOLTIP_HIDE_DELAY);
       expect(tooltipDirective._isTooltipVisible()).toBe(false);
+    }));
+
+    fit('should not follow through with hide if show is called after', fakeAsync(() => {
+      tooltipDirective.show();
+      expect(tooltipDirective._isTooltipVisible()).toBe(true);
+
+      // After hide called, a timeout delay is created that will to hide the tooltip.
+      tooltipDirective.hide();
+      expect(tooltipDirective._isTooltipVisible()).toBe(true);
+
+      // Before delay time has passed, call show which should cancel intent to hide tooltip.
+      tooltipDirective.show();
+      tick(TOOLTIP_HIDE_DELAY);
+      expect(tooltipDirective._isTooltipVisible()).toBe(true);
     }));
 
     it('should remove the tooltip when changing position', () => {
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -1,7 +1,7 @@
 import {async, ComponentFixture, TestBed, tick, fakeAsync} from '@angular/core/testing';
 import {Component, DebugElement} from '@angular/core';
 import {By} from '@angular/platform-browser';
-import {TooltipPosition, MdTooltip, MATERIAL_TOOLTIP_HIDE_DELAY} from './tooltip';
+import {TooltipPosition, MdTooltip, TOOLTIP_HIDE_DELAY} from './tooltip';
 import {OverlayContainer} from '../core';
 import {MdTooltipModule} from './tooltip';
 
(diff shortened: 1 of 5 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -51,143 +55,205 @@ export class MdTooltip {
...
-    if (this.visible && this._overlayRef && this._overlayRef.hasAttached()) {
-      this.visible = false;
-      this._overlayRef.detach();
-    }
+  /** Begins the animation to hide the tooltip after the provided delay in ms */
+  hide(delay: number): void {
+    this._hideTimeoutId = setTimeout(() => {
+      this._visibility = 'hidden';
+      this._closeOnInteraction = false;
+    }, delay);
```
jelbourn: Unit test for `show()` being called while `hide()` is in-progress?

src/lib/tooltip/tooltip.spec.ts
```diff
@@ -1,14 +1,16 @@
...
 import {By} from '@angular/platform-browser';
-import {TooltipPosition, MdTooltip} from './tooltip';
+import {TooltipPosition, MdTooltip, TOOLTIP_HIDE_DELAY} from './tooltip';
 import {OverlayContainer} from '../core';
 import {MdTooltipModule} from './tooltip';
 
+const initialTooltipMessage = 'initial tooltip message';
 
-describe('MdTooltip', () => {
+fdescribe('MdTooltip', () => {
```
jelbourn: `fdescribe` and `fit`

src/lib/tooltip/tooltip.ts
```diff
@@ -1,46 +1,50 @@
...
-    NgModule,
-    ModuleWithProviders,
-    Component,
-    Directive,
-    Input,
-    ElementRef,
-    ViewContainerRef,
-    ChangeDetectorRef
+    NgModule, ModuleWithProviders, Component, Directive, Input, ElementRef, ViewContainerRef,
+    style, trigger, state, transition, animate, AnimationTransitionEvent, NgZone
```
jelbourn: These are generally wrapped to be one-per line; WebStorm's `Optimize imports` command should do this automatically (though it always removes the trailing comma and uses 4 spaces).

andrewseguin:
Thanks for the comments, I removed the fdescribe and updated the imports

andrewseguin:
Realized I needed a detectChanges to trigger the done animation and then used fixture.whenStable() to test for the observable trigger that follows. Change is ready for review

src/lib/tooltip/tooltip.spec.ts
```diff
@@ -38,25 +39,87 @@ describe('MdTooltip', () => {
...
+      tooltipDirective.hide();
+      expect(tooltipDirective._isTooltipVisible()).toBe(true);
+
+      // After the tooltip delay elapses, expect that the tooltip is not visible.
+      tick(TOOLTIP_HIDE_DELAY);
+      fixture.detectChanges();
+      expect(tooltipDirective._isTooltipVisible()).toBe(false);
+
+      // On animation complete, should expect that the tooltip has been detached.
+      fixture.whenStable().then(() => {
```
jelbourn: You shouldn't need `whenStable` with `fakeAsync`, since you're in control of ticking the clock. 
(unless there's something special going on with ngAnimate that I don't know about)

src/lib/tooltip/tooltip.ts
```diff
@@ -51,145 +75,209 @@ export class MdTooltip {
...
-   * Updates the tooltip's position
+   * Interactions on the HTML body should close the tooltip immediately as defined in the
+   * material design spec.
+   * https://material.google.com/components/tooltips.html#tooltips-interaction
    */
-  private _updatePosition() {
-    if (this._overlayRef) {
-      this._changeDetectionRef.detectChanges();
-      this._overlayRef.updatePosition();
+  _handleBodyInteraction(): void {
```
jelbourn: I thought about this a little more, and started worry that this will be a performance problem- every instance of tooltip will add a handler that will run on every click in the application, which could add up to a slowdown. I'm thinking now that it would be better to directly add and remove the click handler when the tooltip is shown/hidden.

andrewseguin: In practice, this is happens because the parent will create an overlay w/ tooltip on show, and detach the tooltip when the tooltip finishes the hide animation. Would you like this behavior to be implemented explicitly in the tooltip component?

jelbourn:
LGTM

jelbourn:
@rolandjitsu can you open an issue requesting that?

rolandjitsu:
@jelbourn sure thing 👍

andrewseguin:
Check out #1806 to track the issue

src/lib/tooltip/tooltip.ts
```diff
@@ -51,154 +75,217 @@ export class MdTooltip {
...
+    let strategy = this._overlay.position().connectedTo(this._elementRef, origin, position);
+    let config = new OverlayState();
+    config.positionStrategy = strategy;
+
+    this._overlayRef = this._overlay.create(config);
+  }
+
+  /** Disposes the current tooltip and the overlay it is attached to */
+  private _disposeTooltip(): void {
+    this._overlayRef.dispose();
```
rolandjitsu: I am seeming some `TypeError: this._overlayRef is null` here, I think it originates from here, not 100% sure. Is this a known issue or is there need for one to be opened?

andrewseguin:
I haven't seen this issue, please open an issue with how to reproduce

rolandjitsu:
@andrewseguin I just opened a bunch more issues :)

Issue #1628, opened 2016-10-27
bug(menu): menu doesn't work in full screen mode


quanterion:
If I put md-toolbar with md-menu on a div element and go into full screen mode using

```
var elem = document.getElementById("my-div-id");
if (elem.requestFullscreen) {
  elem.requestFullscreen();
}
else if (elem.webkitRequestFullScreen) {
  elem.webkitRequestFullScreen();
}
```

then md-menu doesn't show at all

plunker: http://plnkr.co/edit/NCYYbZklZIdYra0xYsfF?p=preview
PS: you should launch it in separate window in order to process fullscreen requests

Environment: Angular 2.0.1 with Material alpha-9-3 using ng-cli
Browsers: latest Chrome & Firefox

quanterion:
Also the same problems appears with dialog, tooltip - all overlay-based components. It looks like overlay service should be tweaked to support full-screen mode (i.e. temporarily use another ElementRef to inject overlays)

willshowell:
Should be closed via #1949, right?

jelbourn:
Yep

nancy-ly:
Can this be re-opened? I am still able to reproduce this issue: https://codepen.io/anon/pen/zRXqwm

quanterion:
@nancy-ly This is a library for Angular 2 and later, not for Angular 1.x

arlowhite:
@jelbourn this should be reopened
https://angular-material2-issue-siqbmy.stackblitz.io

quanterion:
you should provide full screen overlay:

`{ provide: OverlayContainer, useClass: FullscreenOverlayContainer },`

```
import {
  OverlayContainer,
  FullscreenOverlayContainer
} from '@angular/cdk/overlay';
```

PR #1653, opened 2016-10-29
fix(overlay): make overlay work in fullscreen mode

quanterion:
This allows all overlay-based components like dialog, menu, tooltip etc work in full screen mode https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen

Implementation moves overlay container into fullscreen element and back to document.body as user navigates between fullscreen and non-fullscreen modes. Apps not using fullscreen mode are unaffected.

fixes https://github.com/angular/material2/issues/1628

PS: I added Fullscreen button near to RTL one to test fix across all components
PPS: I have no idea about tests for this fix, any advice?

### DIFF
--- a/src/demo-app/demo-app/demo-app.html
+++ b/src/demo-app/demo-app/demo-app.html
@@ -25,6 +25,9 @@
       </button>
       <div class="demo-toolbar">
         <h1>Angular Material 2 Demos</h1>
+        <button md-button (click)="toggleFullscreen()" title="Toggle fullscreen">
+          Fullscreen
+        </button>
         <button md-button (click)="root.dir = (root.dir == 'rtl' ? 'ltr' : 'rtl')" title="Toggle between RTL and LTR">
           {{root.dir.toUpperCase()}}
         </button>
--- a/src/demo-app/demo-app/demo-app.scss
+++ b/src/demo-app/demo-app/demo-app.scss
@@ -39,3 +39,9 @@ body {
     font-size: 20px;
   }
 }
+
+// stretch to screen size in fullscreen mode
+.demo-content {
+  width: 100%;
+  height: 100%;
+}
--- a/src/demo-app/demo-app/demo-app.scss
+++ b/src/demo-app/demo-app/demo-app.scss
@@ -44,4 +44,5 @@ body {
 .demo-content {
   width: 100%;
   height: 100%;
+  background-color: white;
 }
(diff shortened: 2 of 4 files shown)


quanterion:
@jelbourn can you review my PR? It would be great to have this fix in next release.

jelbourn:
@quanterion working my way here- I need to set aside a chunk of time to look at the issue.

jelbourn:
Thanks for your patience on this one. I took a look at the issue and think a cleaner approach would be to introduce a new type of `OverlayContainer` that handles the fullscreen parts. It would look something like this:
```ts
@Injectable()
export class FullscreenFriendlyOverlayContainer extends OverlayContainer {
  protected _createContainer(): void {
    super._createContainer();

    this._addFullscreenChangeListener(() => this._adjustParentForFullscreenChange());
  }

  private _adjustParentForFullscreenChange(): void {
    if (!this._containerElement) {
      return;
    }

    let fullscreenElement = this._getFullscreenElement();
    let parent = fullscreenElement || document.body;
    parent.appendChild(this._containerElement);
  }

  private _addFullscreenChangeListener(fn: () => void) {
    if (document.fullscreenEnabled) {
      document.addEventListener('fullscreenchange', fn);
    } else if (document.webkitFullscreenEnabled) {
      document.addEventListener('webkitfullscreenchange', fn);
    } else if ((document as any).mozFullScreenEnabled) {
      document.addEventListener('mozfullscreenchange', fn);
    } else if ((document as any).msFullscreenEnabled) {
      document.addEventListener('MSFullscreenChange', fn);
    }
  }

  private _getFullscreenElement(): Element {
    return document.fullscreenElement ||
        document.webkitFullscreenElement ||
        (document as any).mozFullScreenElement ||
        (document as any).msFullscreenElement ||
        null;
  }
}
```

To use this container instead of the default, you would add a new provider to your app:
```ts
providers: [
  {provide: OverlayContainer, useClass: FullscreenFriendlyOverlayContainer},
]
```

Doing it this way has a couple of advantages:
* All of the code dealing with fullscreen is encapsulated within a single class.
* If an app isn't using `FullscreenFriendlyOverlayContainer`, the code can be completely dropped by optimizers like Closure Compiler and rollup. This is good, since most apps don't need to deal with fullscreen.

Would you be willing up update your PR to use this approach and add the appropriate tests / comments / docs?

quanterion:
closed in favor of https://github.com/angular/material2/pull/1949

PR #1655, opened 2016-10-31
feat(select): integrate with Angular forms

kara:
This PR adds basic forms integration. 
Still TODO for forms:
- disabled support (see #1667)

r: @jelbourn

### DIFF
--- a/src/lib/select/option.ts
+++ b/src/lib/select/option.ts
@@ -2,6 +2,7 @@ import {
   Component,
   ElementRef,
   EventEmitter,
+  Input,
   Output,
   Renderer,
   ViewEncapsulation
@@ -16,7 +17,7 @@ import {ENTER, SPACE} from '../core/keyboard/keycodes';
     'tabindex': '0',
     '[class.md-selected]': 'selected',
     '[attr.aria-selected]': 'selected.toString()',
-    '(click)': 'select()',
+    '(click)': '_selectViaInteraction()',
     '(keydown)': '_handleKeydown($event)'
   },
   templateUrl: 'option.html',
@@ -25,6 +26,9 @@ import {ENTER, SPACE} from '../core/keyboard/keycodes';
 })
 export class MdOption {
   private _selected = false;
+
+  /** The form value of the option. */
+  @Input() value: any;
 
   /** Event emitted when the option is selected. */
   @Output() onSelect = new EventEmitter();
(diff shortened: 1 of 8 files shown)


src/lib/select/option.ts
```diff
@@ -46,9 +50,9 @@ export class MdOption {
   }
 
   /** Selects the option. */
-  select(): void {
+  select(isUserInput = false): void {
```
kara: That makes way more sense. Good call.

src/lib/select/select.spec.ts
```diff
@@ -206,6 +207,102 @@ describe('MdSelect', () => {
...
+    beforeEach(() => {
+      fixture = TestBed.createComponent(BasicSelect);
+      fixture.detectChanges();
+
+      trigger = fixture.debugElement.query(By.css('.md-select-trigger')).nativeElement;
+    });
+
+    it('should set the view value from the form', () => {
+      let value = fixture.debugElement.query(By.css('.md-select-value'));
+      expect(value).toBeNull();
```
kara: Ack, I always forget about that

src/lib/select/select.ts
```diff
@@ -110,6 +136,38 @@ export class MdSelect implements AfterContentInit, OnDestroy {
...
   }
 
+  /**
+   * Sets the select's value. Part of the ControlValueAccessor interface
+   * required to integrate with Angular's core forms API.
+   */
+  writeValue(value: any): void {
+    if (!this.options) { return; }
+
+    this.options.forEach((option: MdOption) => {
```
kara: this.options is a QueryList, so you'd have to use toArray() first.

kara:
@jelbourn Comments addressed

src/lib/select/select.spec.ts
```diff
@@ -206,6 +207,114 @@ describe('MdSelect', () => {
...
+
+    it('should set the view value from the form', () => {
+      let value = fixture.debugElement.query(By.css('.md-select-value'));
+      expect(value).toBeNull('Expected trigger to start with empty value.');
+
+      fixture.componentInstance.control.setValue('pizza-1');
+      fixture.detectChanges();
+
+      value = fixture.debugElement.query(By.css('.md-select-value'));
+      expect(value.nativeElement.textContent).toContain('Pizza',
```
jelbourn: Prefer breaking at the higher syntactic level, e.g.:
```ts
expect(value.nativeElement.textContent)
    .toContain('Pizza', `Expected trigger to be populated by the control's new value.`);
```
(general rule-of-thumb that makes code more readable in all languages)

kara: Yeah, that does look better. I'll reformat.

PR #1661, opened 2016-10-31
fix(overlay): not taking up entire viewport if body is scrollable

### LINKED ISSUE #1633: bug(dialog): opening a dialog, scrolls my page to the top.

Bug.

Whenever I open a dialog, my page scrolls to the top.

I would like my dialog to be instantiated at the center of what is currently in view.

Create a page with a long list, and open a dialog while at the bottom of the list.

I would like to have a list, where each entry can produce a dialog.

Angular 2.1.0, @angular/material": "2.0.0-alpha.9-3, Chrome 54

<3

crisbeto:
Fixes the overlay not taking up the entire viewport if the `body` is scrollable.

Fixes #1633.

### DIFF
--- a/src/lib/core/overlay/_overlay.scss
+++ b/src/lib/core/overlay/_overlay.scss
@@ -10,7 +10,7 @@
 
   // The overlay-container is an invisible element which contains all individual overlays.
   .md-overlay-container {
-    position: absolute;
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

Issue #1389, opened 2016-09-30
[Bug] Md-Slider not working within Md-Tabs


Daeluse:
#### Bug, feature request, or proposal:

Bug, related to #1234 & #1303
#### What is the current behavior?

If an md-slider is contained within an md-tab, the slider handle starts at left(0px) and cannot be dragged, resizing fixes this issue.
#### What are the steps to reproduce?

Create an Tab Group, a Tab, and place a Slider within the Tab

Plunker: http://plnkr.co/edit/dKYCUkngCeIRwIi9Q0pT?p=preview
#### Which versions of Angular, Material, OS, browsers are affected?

Angular 2.0.0
Material 2.0.0-alpha.8-2
Google Chrome Version 53.0.2785.116 m (64-bit)

Issue #1304, opened 2016-09-22
md-slider ist not working properly


socke1995:
#### Bug, feature request, or proposal:

bug
#### What is the expected behavior?

slider handler is dragable
#### What is the current behavior?

handler is just fix at zero, after resizing the window the handler ist dragable 
after this the slider still does not give values back
the slider doesn't hold on to the steps I put in as attributes
#### What are the steps to reproduce?

I have a sidenav layout with another sidenav layout in it
there I have a sidenav with there three tabs
Within these three tabs there ist some content and the slider
#### Which versions of Angular, Material, OS, browsers are affected?

Angular2, Material Alpha 8
tested this under FF and Chrome

nisdec:
I can confirm this.

jmcgoldrick:
I think this is actually related to: #1234 which also explains why it works after window resize. #1303 fixes this issue, though I'm not totally sold that this is the only/best way to handle the issue...

berc:
Hi, I have the same issue when slider is not displayed (display: none) during page load.

svcninja:
Same issue

PR #1663, opened 2016-10-31
fix(slider): refactor the slider to use percent values for the track …

mmalerba:
…fill and thumb position.

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -127,7 +127,6 @@ export class MdSlider implements ControlValueAccessor {
   private _tickIntervalPercent: number = 0;
 
   get tickIntervalPercent() { return this._tickIntervalPercent; }
-  get halfTickIntervalPercent() { return this._tickIntervalPercent / 2; }
 
   /** The percentage of the slider that coincides with the value. */
   private _percent: number = 0;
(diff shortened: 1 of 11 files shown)


mmalerba:
fixes #1389 #1304 #1234

src/lib/slider/slider.ts
```diff
@@ -213,21 +251,21 @@ export class MdSlider implements AfterContentInit, ControlValueAccessor {
...
   /** TODO: internal */
   onResize() {
+    // Consider the slider to be sliding during resize to prevent animation.
     this.isSliding = true;
-    this._sliderDimensions = this._renderer.getSliderDimensions();
-    // Skip updating the value and position as there is no new placement.
-    this._renderer.updateThumbAndFillPosition(this._percent, this._sliderDimensions.width);
+    setTimeout(() => {
+      this.isSliding = false;
+    }, 0);
```
jelbourn: You can use `requestAnimationFrame`
```ts
requestAnimationFrame(() => this.isSliding = false);
```

stylelint-config.json
```diff
@@ -28,7 +28,9 @@
 
     "property-case": "lower",
 
-    "declaration-block-no-duplicate-properties": true,
+    "declaration-block-no-duplicate-properties": [ true, {
+      "ignore": ["consecutive-duplicates"]
```
jelbourn: Let's use `consecutive-duplicates-with-different-values` in case someone just accidentally duplicates a line.

src/lib/slider/slider.ts
mmalerba: I think this is not even necessary anymore now that the position is percent based.

stylelint-config.json
mmalerba: So I actually tried that and I got: Invalid value "consecutive-duplicates-with-different-values", maybe we're not using the latest version of stylelint?

mmalerba:
Comments addressed and rebased

PR #1664, opened 2016-10-31
test(snackbar): fix tests expectations in dismissal subscription

andrewseguin:
The afterDismissal subscription in the tests were not being called. The assertions are moved to fixture.whenStable which will also check that afterDismissal completed


jelbourn:
LGTM

PR #1665, opened 2016-10-31
Disable broken optional tests.

mmalerba:


test/browser-providers.ts
```diff
@@ -314,7 +314,8 @@ const sauceAliases: AliasMap = {
   'BETA': ['SL_CHROMEBETA', 'SL_FIREFOXBETA'],
   'DEV': ['SL_CHROMEDEV', 'SL_FIREFOXDEV'],
   'REQUIRED': buildConfiguration('unitTest', 'SL', true),
-  'OPTIONAL': buildConfiguration('unitTest', 'SL', false)
```
jelbourn: Rather than commenting out this, we can just change the `configuration` object to have `null` for `target` on the failing environments.

PR #1668, opened 2016-11-01
Disable broken optional tests.

mmalerba:


test/browser-providers.ts
```diff
@@ -329,7 +329,7 @@ const browserstackAliases: AliasMap = {
   'IOS': ['BS_IOS7', 'BS_IOS8', 'BS_IOS9'],
   'SAFARI': ['BS_SAFARI7', 'BS_SAFARI8', 'BS_SAFARI9'],
   'REQUIRED': buildConfiguration('unitTest', 'BS', true),
-  'OPTIONAL': buildConfiguration('unitTest', 'BS', false)
+  'OPTIONAL': buildConfiguration('unitTest', null, false)
```
jelbourn: Not quite what I meant, sorry- 

In this configuration block:
https://github.com/angular/material2/blob/673e7ccc8e21c01d27f16507536ae1f106078697/test/browser-providers.ts#L27-L27

Any environment that we don't care about (like IE9) has `target: null`. We should change it here in this block.

mmalerba: Done

jelbourn:
LGTM

PR #1677, opened 2016-11-01
chore(ci): only run within tunnel if necessary

DevVersion:
**TEST**


DevVersion:
This would be another approach to the `$MODE` check, but both way work fine.

PR #1679, opened 2016-11-01
feat(dialog): allow for a config object to be passed on init

crisbeto:
Adds the ability to pass in a configuration object when creating a new `MdDialogConfig`. This should make working with a lot of dialog options a little easier.

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -15,7 +15,7 @@ export class MdDialogConfig {
   role?: DialogRole = 'dialog';
 
   /** Whether the user can use escape or clicking outside to close a modal. */
-  disableClose = false;
+  disableClose?: boolean = false;
 
   // TODO(jelbourn): add configuration for size, lifecycle hooks, ARIA labelling.
 }
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -40,7 +40,9 @@ export class MdDialog {
    * @param component Type of the component to load into the load.
    * @param config
    */
-  open<T>(component: ComponentType<T>, config = new MdDialogConfig()): MdDialogRef<T> {
+  open<T>(component: ComponentType<T>, config?: MdDialogConfig): MdDialogRef<T> {
+    config = this._applyConfigDefaults(config);
+
     let overlayRef = this._createOverlay(config);
     let dialogContainer = this._attachDialogContainer(overlayRef, config);
 
@@ -124,6 +126,15 @@ export class MdDialog {
         .centerVertically();
 
     return state;
+  }
+
+  /**
+   * Applies default options to the dialog config.
+   * @param dialogConfig Config to be modified.
+   * @returns The new configuration object.
+   */
+  private _applyConfigDefaults(dialogConfig: MdDialogConfig): MdDialogConfig {
+    return Object.assign(new MdDialogConfig(), dialogConfig);
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/lib/dialog/dialog-config.ts
```diff
@@ -3,17 +3,25 @@ import {ViewContainerRef} from '@angular/core';
...
-export class MdDialogConfig {
+export class MdDialogConfig implements MdDialogConfigObject {
   viewContainerRef: ViewContainerRef;
-
-  /** The ARIA role of the dialog element. */
   role: DialogRole = 'dialog';
 
+  constructor(config?: MdDialogConfigObject) {
+    Object.assign(this, config);
+  }
```
crisbeto: I've been trying this approach out, but it seems like the `role` stays undefined in this case.

jelbourn: I see what you mean. I want to keep it such the preferred API is 
```ts
dialog.open(SomeComponent, {
  someConfig: true
});
```
And all the defaults are applied. 

How about just having a method in `MdDialog` like `_applyConfigDefaults` that takes the config and applies the default values for things that are unset?

crisbeto: That sounds good. If we remove the `MdDialogConfig` class, it would be a breaking change, right?

jelbourn: We don't have to remove the class; it will still be the type of the argument, people can just _additionally_ use a matching literal.

crisbeto:
Updated with your feedback @jelbourn.

src/lib/dialog/dialog.ts
```diff
@@ -122,6 +124,20 @@ export class MdDialog {
 
     return state;
   }
+
+  /**
+   * Applies default options to the dialog config.
+   * @param {MdDialogConfig} config Config to be modified.
+   * @returns The new configuration object.
```
DevVersion: Are the JSDoc types necessary here? I think the TypeScript types should be enough.

crisbeto: Good point. Removed the types and renamed the argument to match the other methods.

src/lib/dialog/dialog.ts
```diff
@@ -125,6 +127,20 @@ export class MdDialog {
...
+
+  /**
+   * Applies default options to the dialog config.
+   * @param dialogConfig Config to be modified.
+   * @returns The new configuration object.
+   */
+  private _applyConfigDefaults(dialogConfig: MdDialogConfig): MdDialogConfig {
+    let defaults: MdDialogConfig = {
+      role: 'dialog',
+      disableClose: false
```
jelbourn: Why not leave the defaults in `MdDialogConfig` and simply say `new MdDialogConfig` here?

crisbeto: True, I'll update it.

crisbeto:
Updated again.

Nogostradamus:
I have following error:
Argument of type '{ disableClose: boolean; }' is not assignable to parameter of type 'MdDialogConfig'.
  Object literal may only specify known properties, and 'disableClose' does not exist in type 'MdDialogConfig'.

this is using example from MdDialog readme:
`this.dialogRef = this.dialog.open(PizzaDialog, {
      disableClose: false
    });`

crisbeto:
It works for me @Nogostradamus, can you retry checking your Material version?

Nogostradamus:
Sorry, my bad. Now it works. I was still on experimental pizza v9 after an update. I had to removed caret from package.json  "@angular/material": "^2.0.0-alpha.10" in order to update successfully.

PR #1680, opened 2016-11-01
fix(radio): only emit change event with user input

### LINKED ISSUE #1523: Radio group change event should only fire with user interaction
Currently, if you set the `value` input programmatically, the change event fires. This behavior doesn't match Angular forms or native forms.

tinayuangao:
Closes #1523 

R: @kara @jelbourn

### DIFF
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -152,8 +152,8 @@ describe('MdRadio', () => {
       expect(spies[1]).toHaveBeenCalledTimes(1);
     });
 
-    it('should not emit a change event from the radio group when change group value '
-         + 'programmatically', () => {
+    it(`should not emit a change event from the radio group when change group value 
+        programmatically`, () => {
       expect(groupInstance.value).toBeFalsy();
 
       let changeSpy = jasmine.createSpy('radio-group change listener');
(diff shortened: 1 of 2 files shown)


src/lib/radio/radio.spec.ts
```diff
@@ -152,21 +152,22 @@ describe('MdRadio', () => {
       expect(spies[1]).toHaveBeenCalledTimes(1);
     });
 
-    it('should emit a change event from the radio group', () => {
+    it('should not emit a change event from the radio group when change group value '
+         + 'programmatically', () => {
```
kara: Have you considered using backticks instead? That way you don't need the `+`

src/lib/radio/radio.spec.ts
```diff
@@ -242,34 +243,36 @@ describe('MdRadio', () => {
 
       fixture.detectChanges();
 
-      expect(changeSpy).toHaveBeenCalled();
+      expect(changeSpy).toHaveBeenCalledTimes(0);
```
kara: I think you can do 

```
expect(changeSpy).not.toHaveBeenCalled()
```

Might be more natural-sounding

src/lib/radio/radio.ts
```diff
@@ -171,6 +166,13 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
     }
   }
 
+  emitChangeEvent(): void {
```
kara: Did you want this to be part of the public API?  It seems to me like it should use the `_` prefix convention so people know it's internal (only used by the button)

tinayuangao:
Comment addressed. Please take another look. Thanks for review!

PR #1678, opened 2016-11-01
feat(dialog): add disableClose option

crisbeto:
Adds a config option that allows users to disable closing a dialog via a backdrop click or pressing escape.

Fixes #1419.

PR #1682, opened 2016-11-01
feat(dialog): add escapeToClose config option

crisbeto:
Adds the ability to disable the pressing escape to close a dialog.

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -14,6 +14,9 @@ export class MdDialogConfig {
   /** The ARIA role of the dialog element. */
   role: DialogRole = 'dialog';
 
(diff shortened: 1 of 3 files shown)


src/lib/dialog/dialog-config.ts
```diff
@@ -14,6 +14,9 @@ export class MdDialogConfig {
   /** The ARIA role of the dialog element. */
   role: DialogRole = 'dialog';
 
+  /** Whether the escape key should close the dialog. */
+  escapeToClose = true;
```
jelbourn: Let's just make this `disableClose` and have it apply to both escape and backdrop click, since you're unlikely to want to allow one but not the other.

crisbeto:
Closing in favor of combining it with #1678.

PR #1683, opened 2016-11-01
chore: update readme status for November

jelbourn:
R: @kara

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,14 +28,15 @@ and which pieces are blocked) and make a comment.
 Also see our [`Good for community contribution`](https://github.com/angular/material2/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+for+community+contribution%22) 
 label.
 
-High level items planned for October 2016:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,14 +65,14 @@ High level items planned for October 2016:
 | dialog           |  Started, not yet ready for release | [README][22] |   [#114][0114] |
 | snackbar / toast | Initial version, needs enhancements | [README][21] |   [#115][0115] |
 | select           |                      Design started |           -  |   [#118][0118] |
```
kara: Do we want to put `select` as started? Given that the first PR is in master.

fxck:
Could you include some info about releases? I could really use some of the stuff/fixes that are in master..

jelbourn:
@fxck we're doing an alpha.10 tomorrow unless something goes awry

jelbourn:
@fxck some overlay and menu changes collided, causing bugs with menu that need to be resolved before a release. It will probably be Monday now.

PR #1684, opened 2016-11-01
fix(ripple): allow pointer events on ripple; disable on checkbox ripple

andrewseguin:
Followup on #1623 (checkbox ripple was clickable)

### DIFF
--- a/src/lib/checkbox/checkbox.scss
+++ b/src/lib/checkbox/checkbox.scss
@@ -425,4 +425,5 @@ md-checkbox {
   bottom: -$md-checkbox-ripple-size;
   border-radius: 50%;
   z-index: 1;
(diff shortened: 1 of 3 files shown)


src/lib/checkbox/checkbox.scss
```diff
@@ -425,4 +425,5 @@ md-checkbox {
   bottom: -$md-checkbox-ripple-size;
   border-radius: 50%;
   z-index: 1;
+  pointer-events: none;
```
andrewseguin: We do, added another commit to include this

jelbourn:
LGTM

tinayuangao:
LGTM. Thanks for the fix!

PR #1690, opened 2016-11-02
fix (radio): add required attribute

krigton:
Issue: 1656. Put required property on the radio group

### DIFF
--- a/src/demo-app/radio/radio-demo.html
+++ b/src/demo-app/radio/radio-demo.html
@@ -8,18 +8,28 @@
 <section class="demo-section">
   <div>
     <span>isDisabled: {{isDisabled}}</span>
-    <button md-raised-button (click)="isDisabled=!isDisabled" class="demo-button">
+   <button md-raised-button (click)="isDisabled=!isDisabled" class="demo-button">
       Disable buttons
+    </button>
+  </div>
+  <div>
+    <span>isRequired: {{isRequired}}</span>
+    <button md-raised-button (click)="isRequired=!isRequired" class="demo-button">
+      Requrie buttons
     </button>
   </div>
   <div>
     <span><md-checkbox [(ngModel)]="isAlignEnd">Align end</md-checkbox></span>
   </div>
-  <md-radio-group name="my_options" [disabled]="isDisabled" [align]="isAlignEnd ? 'end' : 'start'">
-    <md-radio-button value="option_1">Option 1</md-radio-button>
-    <md-radio-button value="option_2">Option 2</md-radio-button>
-    <md-radio-button value="option_3">Option 3</md-radio-button>
-  </md-radio-group>
+  <form>
+    <md-radio-group name="my_options" [disabled]="isDisabled" [align]="isAlignEnd ? 'end' : 'start'" [required]="isRequired">
+      <md-radio-button value="option_1">Option 1</md-radio-button>
+      <md-radio-button value="option_2">Option 2</md-radio-button>
+      <md-radio-button value="option_3">Option 3</md-radio-button>
+    </md-radio-group>
+    <br/>
+    <button md-raised-button>Send form</button>
+  </form>
 </section>
 <h1>Favorite Season Example</h1>
 <h2>Dynamic Example with two-way data-binding</h2>
--- a/src/demo-app/radio/radio-demo.html
+++ b/src/demo-app/radio/radio-demo.html
@@ -8,14 +8,14 @@
 <section class="demo-section">
   <div>
     <span>isDisabled: {{isDisabled}}</span>
-   <button md-raised-button (click)="isDisabled=!isDisabled" class="demo-button">
+    <button md-raised-button (click)="isDisabled=!isDisabled" class="demo-button">
       Disable buttons
     </button>
   </div>
   <div>
     <span>isRequired: {{isRequired}}</span>
     <button md-raised-button (click)="isRequired=!isRequired" class="demo-button">
-      Requrie buttons
+      Require buttons
     </button>
   </div>
   <div>
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -80,13 +80,26 @@ describe('MdRadio', () => {
       }
     });
 
-    it('should set required to each individual radio when the group is requried', () => {
+    it('should set required to each individual radio when the group is required', () => {
       testComponent.isGroupRequired = true;
       fixture.detectChanges();
 
       for (let radio of radioInstances) {
         expect(radio.required).toBe(true);
       }
+    });
+
+    it('should set / unset the required attribute on the underlaying radio input', () => {
+      let nativeRadioInput = <HTMLElement> radioNativeElements[0].querySelector('input');
+      radioInstances[0].required = true;
+      fixture.detectChanges();
+
+      expect(nativeRadioInput.getAttribute('required')).toBeDefined();
+
+      radioInstances[0].required = null;
+      fixture.detectChanges();
+
+      expect(nativeRadioInput.getAttribute('required')).toBeNull();
     });
 
     it('should disable each individual radio when the group is disabled', () => {
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -89,14 +89,14 @@ describe('MdRadio', () => {
       }
     });
 
-    it('should set / unset the required attribute on the underlaying radio input', () => {
+    it('should set / unset the required attribute on the underlying radio input', () => {
       let nativeRadioInput = <HTMLElement> radioNativeElements[0].querySelector('input');
       radioInstances[0].required = true;
       fixture.detectChanges();
 
       expect(nativeRadioInput.getAttribute('required')).toBeDefined();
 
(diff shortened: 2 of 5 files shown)


src/lib/radio/radio.ts
```diff
@@ -149,6 +152,18 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
...
+  get required(){
+    return this._required;
+  }
+
+  set required(required: boolean) {
+    this._required = required;
+
+    if (this._radios) {
+      this._radios.forEach(radio => radio.required = this.required);
+    }
```
jelbourn: This is problematic if a radio button is added to the group after it has been marked as required. A better approach would be that the individual radio button uses as its `required` check:
```ts
this.required || (this.group && this.group.required)
```
([which is how `disabled` works](https://github.com/angular/material2/blob/065469ae3626bc6c99cf69100eefb30a881a7bc7/src/lib/radio/radio.ts#L350-L350))

Also, the setters need to use the `coerceBooleanProperty` function.

src/lib/radio/radio.ts
```diff
@@ -260,6 +275,9 @@ export class MdRadioButton implements OnInit {
   /** The 'aria-labelledby' attribute takes precedence as the element's text alternative. */
   @Input('aria-labelledby') ariaLabelledby: string;
 
+  /** Whether this radio is required. */
+  @Input('required') required: any = null;
```
jelbourn: If the name of the input matches the name of the field, you don't need to specify it.

src/lib/radio/radio.ts
```diff
@@ -366,6 +384,7 @@ export class MdRadioButton implements OnInit {
       this.checked = this.radioGroup.value === this._value;
       // Copy name from parent radio group
       this.name = this.radioGroup.name;
+      this.required = this.radioGroup.required ? 'true' : null;
```
jelbourn: This won't be necessary when the getter is changed to act like `disabled`

src/lib/radio/radio.spec.ts
```diff
@@ -80,6 +80,15 @@ describe('MdRadio', () => {
...
 
+    it('should set required to each individual radio when the group is requried', () => {
+      testComponent.isGroupRequired = true;
+      fixture.detectChanges();
+
+      for (let radio of radioInstances) {
+        expect(radio.required).toBe(true);
+      }
+    });
+
```
jelbourn: There should also be a test that an individual radio button with `required = true` properly sets `required` to the underlying native `<input type="radio">`

jelbourn:
R: @tinayuangao for any additional comments

src/demo-app/radio/radio-demo.html
```diff
@@ -8,18 +8,28 @@
 <section class="demo-section">
   <div>
     <span>isDisabled: {{isDisabled}}</span>
-    <button md-raised-button (click)="isDisabled=!isDisabled" class="demo-button">
+   <button md-raised-button (click)="isDisabled=!isDisabled" class="demo-button">
```
tinayuangao: very nit: add space before <button

src/demo-app/radio/radio-demo.html
```diff
@@ -8,18 +8,28 @@
...
       Disable buttons
     </button>
   </div>
   <div>
+    <span>isRequired: {{isRequired}}</span>
+    <button md-raised-button (click)="isRequired=!isRequired" class="demo-button">
+      Requrie buttons
```
tinayuangao: typo: Require

src/lib/radio/radio.ts
```diff
@@ -80,6 +80,9 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
   /** Whether the `value` has been set to its initial value. */
   private _isInitialized: boolean = false;
 
+  /** Wheter this radio is required */
```
tinayuangao: typo: Whether

src/demo-app/radio/radio-demo.html
tinayuangao: required: boolean;

tinayuangao: Use `boolean` instead of `any`.

krigton:
I'll do the changes. Thank you gusy

src/lib/radio/radio.ts
krigton: I think it will be just like the disabled attribute at the end

krigton: @jelbourn I think there is a strange situation here and I'm not good enough in angular 2 yet in order to understand. So if I add an item with `required="true"` then setting the group required property will not affect this one. It's a bit strange behaviour.
I see that for the disabled attribute works the same way and I can see the reason that for disabled you might want to update the disabled radio button directly not through the group, but for required it's enough to have one of the buttons required and then message will show up if you submit. 
Not sure what to do here. Also why is that for disabled there will be an `ng-reflect-disabled` and for required is not. Is this a native behaviour? Thanks

krigton: I found the `ng-reflect-required` thingy. I used the `required="true"` without the brackets.

krigton: I understand now how it works.

src/lib/radio/radio.ts
krigton: @jelbourn should we add some kind of required styling?

src/lib/radio/radio.ts
```diff
@@ -347,13 +344,13 @@ export class MdRadioButton implements OnInit {
   }
 
   set value(value: any) {
-    if (this._value != value) {
-      if (this.radioGroup != null && this.checked) {
-        this.radioGroup.value = value;
-      }
-      this._value = value;
+  if (this._value != value) {
```
tinayuangao: Please revert the changes in `set value`

krigton: Done. pushed again

src/lib/radio/radio.ts
tinayuangao: Use `coerceBooleanProperty` for the setters?

krigton: Not sure what you mean but I guess instead of this : `(value != null && value !== false) ? true : null;` simply use `coerceBooleanProperty`. I changed on the disabled as well. Pushed also

src/lib/radio/radio.ts
```diff
@@ -274,7 +289,7 @@ export class MdRadioButton implements OnInit {
 
   @Input()
   get disableRipple(): boolean { return this._disableRipple; }
-  set disableRipple(value) { this._disableRipple = coerceBooleanProperty(value); }
+  set disableRipple(value) { this._disableRipple = (value); }
```
tinayuangao: Keep the `coerceBooleanProperty` for disableRipple

src/lib/radio/radio.ts
```diff
@@ -117,6 +120,15 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
   }
 
   @Input()
+  get required(): boolean {
+    return this._required;
+  }
+
+  set required(value) {
+    this._required = (value != null && value !== false) ? true : null;
```
tinayuangao: Use `coerceBooleanProperty` here.

src/lib/radio/radio.ts
krigton: hm. I ran `npm test` after changes this one got away. sorry. changed and pushed again. thanks

src/lib/radio/radio.ts
tinayuangao: Could you please also update this?

krigton: I did, I'm not sure but initially broke the code and the test, then I used the [required] instead of [attr.required] and it worked. is this a correct behaviour I mean [attr.required] not having the same effect as [required]? thanks

tinayuangao: `[attr.required]` is not the same as `[required]`. `[attr.required]` set the attribute `required` in the html element, and when the value is null the attribute `required` is removed. 

The `[required]` calls the `set required` in the component.

src/lib/radio/radio.ts
```diff
@@ -117,6 +120,15 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
   }
 
   @Input()
+  get required(): boolean {
+    return this._required;
+  }
+
+  set required(value) {
+    this._required = coerceBooleanProperty(value)
```
tinayuangao: Please add `;`. This broke lint test.

src/lib/radio/radio.ts
krigton: thanks

src/lib/radio/radio.ts
krigton: corrected

tinayuangao:
LGTM

krigton:
@tinayuangao should I create an issue regarding to the value problem?

mmalerba:
@tinayuangao what needs to be done here?

krigton:
hm I didn't checked this one lately I'll do it this evening and commit. thanks

jelbourn:
Going to close this since I think it's out of date now. Feel free to open a new PR if you'd like to do an update.

Issue #50, opened 2016-01-28
Add focus trapping to sidenav.


hansl:
Depends on issue #49.

PR #1695, opened 2016-11-02
Focus capturing for sidenav.

mmalerba:
Captures focus when sidenav is open in "over" or "push" mode, but not
when opened in "side" mode.

### DIFF
--- a/src/lib/core/a11y/focus-trap.ts
+++ b/src/lib/core/a11y/focus-trap.ts
@@ -8,6 +8,7 @@ import {
   NgZone
 } from '@angular/core';
 import {InteractivityChecker} from './interactivity-checker';
+import {coerceBooleanProperty} from '../coersion/boolean-property';
 
 
 /**
@@ -28,17 +29,8 @@ export class FocusTrap implements AfterContentInit {
   @ViewChild('trappedContent') trappedContent: ElementRef;
 
   @Input()
-  get active(): boolean {
-    return this._active;
-  }
-  set active(val: boolean) {
-    this._active = val;
-    if (val && this._contentReady) {
-      this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-        this.focusFirstTabbableElement();
-      });
-    }
-  }
+  get active(): boolean { return this._active; }
+  set active(val: boolean) { this._active = coerceBooleanProperty(val); }
 
   /** Whether the DOM content is ready. */
   private _contentReady: boolean = false;
@@ -52,10 +44,28 @@ export class FocusTrap implements AfterContentInit {
     this._contentReady = true;
     // Trigger setter behavior.
     if (this.active) {
-      this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-        this.focusFirstTabbableElement();
-      });
+
     }
+  }
+
+  /**
+   * Waits for microtask queue to empty, then focuses the first tabbable element within the focus
+   * trap region.
+   */
+  focusFirstTabbableElementWhenReady() {
+    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
+      this.focusFirstTabbableElement();
+    });
+  }
+
+  /**
+   * Waits for microtask queue to empty, then focuses the last tabbable element within the focus
+   * trap region.
+   */
+  focusLastTabbableElementWhenReady() {
+    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
+      this.focusLastTabbableElement();
+    });
   }
 
   /** Focuses the first tabbable element within the focus trap region. */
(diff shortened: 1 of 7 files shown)


src/lib/dialog/dialog-container.ts
```diff
@@ -57,13 +56,8 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
...
     let attachResult = this._portalHost.attachComponentPortal(portal);
 
-    // If were to attempt to focus immediately, then the content of the dialog would not yet be
-    // ready in instances where change detection has to run first. To deal with this, we simply
-    // wait for the microtask queue to be empty.
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-      this._elementFocusedBeforeDialogWasOpened = document.activeElement;
-      this._focusTrap.focusFirstTabbableElement();
-    });
+    this._elementFocusedBeforeDialogWasOpened = document.activeElement;
```
mmalerba: Do you think its worth moving this to the FocusTrap as well? Maybe like this: focusOnDeactivate="<element>"

mmalerba:
fixes #50

jelbourn:
cc @crisbeto, relevant to dialog

jelbourn:
@marcysutton as far as you know, if there ever a time when you would want to trap focus within some subtree of the DOM and _not_ automatically move focus into that subtree as soon as it's added to the page?

src/lib/core/a11y/focus-trap.ts
```diff
@@ -1,4 +1,6 @@
-import {Component, ViewEncapsulation, ViewChild, ElementRef} from '@angular/core';
+import {
+    Component, ViewEncapsulation, ViewChild, ElementRef, Input, AfterContentInit, NgZone
```
crisbeto: I think these are supposed to all be on separate lines.

src/lib/core/a11y/focus-trap.ts
```diff
@@ -15,15 +17,44 @@ import {InteractivityChecker} from './interactivity-checker';
...
 
-  constructor(private _checker: InteractivityChecker) { }
+  @Input()
+  get active(): boolean {
+    return this._active;
+  }
+  set active(val: boolean) {
+    this._active = val;
+    if (val && this._contentReady) {
+      this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
```
crisbeto: This one is used here and repeated down inside `ngAfterContentInit`. It might a good idea to move it to a private method like `waitForTabbableElement`.

src/lib/dialog/dialog-container.ts
```diff
@@ -45,6 +41,9 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
   /** Reference to the open dialog. */
   dialogRef: MdDialogRef<any>;
 
+  /** Whether the focus trap is active. */
+  focusTrapActive: boolean = false;
```
crisbeto: Not sure what the style guide says about this, but wouldn't the `boolean` be inferred from the default value anyway @jelbourn?

marcysutton:
@jelbourn focus traps are usually enabled/disabled on command when the UI requires it, such as a modal dialog being shown. Otherwise you run the risk of trapping a keyboard user inside a menu at the wrong time.

You might want to check out HTML `inert` and the related polyfill for this purpose. https://github.com/GoogleChrome/inert-polyfill

mmalerba:
> focus traps are usually enabled/disabled on command when the UI requires it
@marcysutton That sounds like what I'm doing here, I added an `active` attribute to `focus-trap` when it changes to `true` the focus trap automatically focuses the first focusable element inside it.

mmalerba:
Changed to not automatically focus the first element when it's activated.

src/lib/sidenav/sidenav.spec.ts
```diff
@@ -251,6 +252,55 @@ describe('MdSidenav', () => {
...
+    let link1Element: HTMLElement;
+    let link2Element: HTMLElement;
+
+    beforeEach(() => {
+      fixture = TestBed.createComponent(SidenavWitFocusableElements);
+      testComponent = fixture.debugElement.componentInstance;
+      sidenav = fixture.debugElement.query(By.directive(MdSidenav)).componentInstance;
+      link1Element = fixture.debugElement.query(By.css('.link1')).nativeElement;
+      link2Element = fixture.debugElement.query(By.css('.link1')).nativeElement;
+      link2Element.focus();
```
jelbourn: Can we call these something like `firstFocusableElement` and `lastFocusableElement`?

jelbourn:
LGTM

kara:
@mmalerba Presubmit is unhappy because  

```Property 'active' does not exist on type 'FocusTrap'```

mmalerba:
@kara fixed, strange that neither build nor lint caught that

kara:
@mmalerba `Property '_focusTrap' is private and only accessible within class 'MdSidenav'`. 

I think it needs to be internal instead of private.

jelbourn:
@mmalerba can you rebase?

PR #1696, opened 2016-11-02
fix(select): fix panel close animation

kara:
Tweaks to make the menu close animation smoother.

r: @hansl

### DIFF
--- a/src/lib/select/select-animations.ts
+++ b/src/lib/select/select-animations.ts
@@ -58,10 +58,10 @@ export const transformPanel: AnimationEntryMetadata = trigger('transformPanel', 
       width: '100%',
       transform: `translate3d(0, 0, 0) scaleY(0)`
     }),
-    animate(`150ms cubic-bezier(0.55, 0, 0.55, 0.2)`)
+    animate(`150ms cubic-bezier(0.25, 0.8, 0.25, 1)`)
   ]),
   transition('* => void', [
-    animate('250ms linear', style({opacity: 0}))
+    animate('250ms 100ms linear', style({opacity: 0}))
   ])
 ]);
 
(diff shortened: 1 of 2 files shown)


src/lib/select/select-animations.ts
```diff
@@ -58,10 +58,10 @@ export const transformPanel: AnimationEntryMetadata = trigger('transformPanel',
       width: '100%',
       transform: `translate3d(0, 0, 0) scaleY(0)`
     }),
-    animate(`150ms cubic-bezier(0.55, 0, 0.55, 0.2)`)
+    animate(`150ms cubic-bezier(0.25, 0.8, 0.25, 1)`)
```
hansl: Where did you get those values?  Seems like a more aggressive ease-out.

src/lib/select/select.ts
```diff
@@ -158,7 +159,6 @@ export class MdSelect implements AfterContentInit, OnDestroy {
       this._focusCorrectOption();
       this.onOpen.emit();
     } else {
-      this._focusHost();
```
hansl: Please change the comment above to reflect the code.

src/lib/select/select.ts
```diff
@@ -108,6 +108,7 @@ export class MdSelect implements AfterContentInit, OnDestroy {
   /** Closes the overlay panel. */
   close(): void {
     this._panelOpen = false;
+    this._focusHost();
```
hansl: Please change the comment above to reflect the code.

src/lib/select/select.scss
```diff
@@ -16,6 +16,7 @@ md-select {
   height: $md-select-trigger-height;
   min-width: $md-select-trigger-min-width;
   cursor: pointer;
+  user-select: none;
```
hansl: ?

src/lib/select/select-animations.ts
kara: These are the Material 1 values for the ease out animation.

Issue #1040, opened 2016-08-14
md-sidenav no animation on close


erasethis:
#### Bug, feature request, or proposal:

Bug
#### What is the expected behavior?

Menu should be animated (slide out) when closing
#### What is the current behavior?

Menu just vanishes when closing
#### What are the steps to reproduce?

Set 'opened' on md-sidenav to true, menu slides in, set 'opened' to false, _poof_ menu is hidden instantly

Providing a Plunker (or similar) is the _best_ way to get the team to see your issue.
Plunker template: http://plnkr.co/edit/o077B6uEiiIgkC0S06dd
#### Which versions of Angular, Material, OS, browsers are affected?

Angular 2 RC5, Material 2.0.0-alpha.7-2
#### Is there anything else we should know?

If I set these styles in my CSS, the menu behaves as expected:

md-sidenav:not(.md-sidenav-opened) {
    -webkit-transition: -webkit-transform 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
    transition: -webkit-transform 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
    transition: transform 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
    transition: transform 400ms cubic-bezier(0.25, 0.8, 0.25, 1), -webkit-transform 400ms cubic-bezier(0.25, 0.8, 0.25, 1), visibility 400ms cubic-bezier(0.25, 0.8, 0.25, 1) !important;
}

.md-sidenav-backdrop {
    transition: background-color 400ms cubic-bezier(0.25, 0.8, 0.25, 1), visibility 400ms cubic-bezier(0.25, 0.8, 0.25, 1);
}

DevVersion:
I just took a look at your issue and tried to reproduce it in a Plunker, but it seems to work as expected.

> http://plnkr.co/edit/pvtTxpQbNuzauzyuCow7?p=preview

You could probably modify the Plunker to demonstrate your issue.

gxclarke:
@erasethis It looks like your missing the outer `md-sidenav-layout` container on your plunker. Although I've also noticed that if you don't set the mode on 'md-sidenav' it defaults to 'side' (rather than 'over' as the documentation says). At least for me, 'side' doesn't animate the content area when the sidenav is opened or closed.

fxck:
[...]

PR #1697, opened 2016-11-02
fix(sidenav): transition margin left & right.

### LINKED ISSUE #1215: "side" mode sidenav doesn't animate content

Bug

Content should animate when opening/closing sidenav in 'side' mode, as it does in 'push' mode.

Content position is adjusted instantly, while the draw animates. Likely CSS transitions not being applied to the contents.

Make a component with a sidenav set to 'side' mode:
http://plnkr.co/edit/H5s8r7830pbJTBGXfgWC?p=preview

Disjointed appearance & inconsistent behavior with other modes.

Angular RC 4-6, Material alphas 5-8. Windows. Chrome, FF, Edge

No.

mmalerba:
fixes #1215 #1040

### DIFF
--- a/src/lib/sidenav/sidenav-transitions.scss
+++ b/src/lib/sidenav/sidenav-transitions.scss
@@ -6,7 +6,8 @@ md-sidenav {
 }
 
 .md-sidenav-content {
-  transition: transform $swift-ease-out-duration $swift-ease-out-timing-function;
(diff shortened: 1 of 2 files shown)


jelbourn:
@mmalerba see https://github.com/angular/material2/pull/460
We had this in the past but it was rather janky on mobile devices. I'd back-burner this since it's not a major priority for beta.

PR #1700, opened 2016-11-02
feat(tabs): add ripples to the tab group and nav bar links

andrewseguin:


src/demo-app/tabs/tabs-demo.html
```diff
@@ -2,7 +2,7 @@
 
 <div class="demo-nav-bar">
   <nav md-tab-nav-bar aria-label="weather navigation links">
-    <a md-tab-link
+    <a md-tab-link md-ripple
```
kara: Does the user have to add the ripple themselves if using tab-link?

kara: Sure, sounds good. If you haven't already, you might want to check out md-button's ripple. It's also an attr directive on an anchor.

andrewseguin:
Okay, I made MdTabLink extend the MdRipple so that it automatically applies the ripple effect on the element. What do you guys think of this approach?

andrewseguin:
Tests are failing due to lacking changes from PR #1684 which fixes ripple pointer events

andrewseguin:
Tests are passing, able to be reviewed now

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -24,7 +25,7 @@ export class MdTabNavBar {
 @Directive({
   selector: '[md-tab-link]',
 })
-export class MdTabLink {
+export class MdTabLink extends MdRipple {
```
kara: Conceptually, I find it a bit confusing that the link extends a ripple.  Any reason why we wouldn't convert this to a component and use the md-ripple-trigger property?

andrewseguin: I agree, it's odd to say that the md tab link extends a ripple. I'd prefer to keep this an attribute directive to preserve the end user's HTML using all native elements (nav, a), e.g.:

```
<nav md-tab-nav-bar>
  <a md-tab-link>Link</a>
  <a md-tab-link>Link</a>
</nav>
```

So there are two more alternatives:
1. Add [md-tab-link] selector to the MdRipple
2. Create a new MdTabLinkRipple directive that will only extend MdRipple and match [md-tab-link]

In talking with Jeremy, I think (2) sounds better than MdTabLink extending MdRipple and (1) just sounds more silly than either.

jelbourn:
@andrewseguin can you rebase?

andrewseguin:
Conflicts resolved, thanks

PR #1703, opened 2016-11-02
fix(portals): dispose portalhost when destroyed

andrewseguin:


andrewseguin:
Would it make sense to also add an onDestroy to the TemplatePortalDirective as well (detach itself from the host on destroy)

jelbourn:
LGTM

@andrewseguin We don't want to have the portal detach itself because the template may be rendered in a host somewhere completely separated from where the portal originally came from. 

However, we should also make this change to `DomPortalHost`
(but wait for #1704 to go in before doing that)

PR #1705, opened 2016-11-03
fix(ripple): color flickering

### LINKED ISSUE #1686: Clicking repeatedly on radio button will flash between grey and theme color
Repeatedly clicking on the radio button will toggle the radio theme styling
`.md-radio-focused .md-radio-ripple .md-ripple-foreground`

The reason is that given a focused radio button:
Mouse down will call _onInputBlur and remove md-radio-focused
Mouse up will call _onInputFocus and add md-radio-focused

One proposed solution could be to check if the _onInputBlur event's target came from within the radio input label and if so, ignore it.

Superd22:
- fix style to generate dark ripples only on disabled radio/checkbox
(same fashion as in ng1 material)

Closes #1686

### DIFF
--- a/src/lib/checkbox/_checkbox-theme.scss
+++ b/src/lib/checkbox/_checkbox-theme.scss
@@ -68,7 +68,7 @@
     }
   }
 
-  .md-checkbox-focused {
+  .md-checkbox:not(.md-checkbox-disabled) {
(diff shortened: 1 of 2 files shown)


tinayuangao:
LGTM

PR #1711, opened 2016-11-03
chore(): remove unused config folder

DevVersion:
* It seems like the `config/` folder in the project root is not used anymore and was from the Angular-CLI previously.


jelbourn:
LGTM

PR #1712, opened 2016-11-03
update(slide-toggle): remove CRLF line endings

DevVersion:
* Previously the slide-toggle files were commited with CRLF line endings, but to provide consistency those should be only LF.


jelbourn:
LGTM

Confirmed that this removes the CR characters (github doesn't show them in the diff)

PR #1713, opened 2016-11-03
fix(checkbox): fix the checkbox squishes problem

### LINKED ISSUE #1097: Squished checkbox (md-checkbox)

tinayuangao:
Add flex-shrink to checkbox so they do not squish if the parent wrapper has a fixed width and the label text is too long.
Continued from #1103

Closes #1097 

R: @jelbourn

### DIFF
--- a/src/lib/checkbox/checkbox.scss
+++ b/src/lib/checkbox/checkbox.scss
@@ -218,6 +218,7 @@ md-checkbox {
   vertical-align: middle;
   white-space: nowrap;
   width: $md-checkbox-size;
+  flex-shrink: 0;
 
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #1716, opened 2016-11-03
fix(dialog): backdrop not being removed if it doesn't have transitions

### LINKED ISSUE #1607: Turning backdrop transition off causes backdrop to stick around
If you remove the backdrop transition (e.g. with a backdropClass), the backdrop is never removed because 'transitionend' doesn't fire. This is a problem because then nothing is clickable in the app.

crisbeto:
Fixes the dialog's backdrop not being removed if it's transition have been disabled.

Fixes #1607.

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -92,7 +92,9 @@ export class OverlayRef implements PortalHost {
 
     // Add class to fade-in the backdrop after one frame.
     requestAnimationFrame(() => {
-      this._backdropElement.classList.add('md-overlay-backdrop-showing');
+      if (this._backdropElement) {
+        this._backdropElement.classList.add('md-overlay-backdrop-showing');
+      }
     });
   }
 
@@ -101,10 +103,11 @@ export class OverlayRef implements PortalHost {
     let backdropToDetach = this._backdropElement;
 
     if (backdropToDetach) {
-      backdropToDetach.classList.remove('md-overlay-backdrop-showing');
-      backdropToDetach.classList.remove(this._state.backdropClass);
-      backdropToDetach.addEventListener('transitionend', () => {
-        backdropToDetach.parentNode.removeChild(backdropToDetach);
+      let onTransitionEnd = () => {
+        // It may not be attached to anything in certain cases (e.g. unit tests).
+        if (backdropToDetach && backdropToDetach.parentNode) {
+          backdropToDetach.parentNode.removeChild(backdropToDetach);
+        }
 
         // It is possible that a new portal has been attached to this overlay since we started
         // removing the backdrop. If that is the case, only clear the backdrop reference if it
@@ -112,7 +115,16 @@ export class OverlayRef implements PortalHost {
         if (this._backdropElement == backdropToDetach) {
           this._backdropElement = null;
         }
-      });
+      };
+
+      backdropToDetach.classList.remove('md-overlay-backdrop-showing');
+      backdropToDetach.classList.remove(this._state.backdropClass);
+      backdropToDetach.addEventListener('transitionend', onTransitionEnd);
+
+      // If the backdrop doesn't have a transition, the `transitionend` event won't fire.
+      // In this case we make it unclickable and we try to remove it after a delay.
+      backdropToDetach.style.pointerEvents = 'none';
+      setTimeout(onTransitionEnd, 500);
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/lib/core/overlay/overlay-ref.ts
```diff
@@ -101,9 +103,8 @@ export class OverlayRef implements PortalHost {
     let backdropToDetach = this._backdropElement;
 
     if (backdropToDetach) {
-      backdropToDetach.classList.remove('md-overlay-backdrop-showing');
-      backdropToDetach.classList.remove(this._state.backdropClass);
-      backdropToDetach.addEventListener('transitionend', () => {
+      let styles = getComputedStyle(backdropToDetach);
```
jelbourn: Instead of using `getComputedStyle`, let's just set `pointer-events: none` on the backdrop and dismiss it after whichever comes frist: transitionend or a short wait (say, 300ms). This will ensure that the backdrop doesn't get in the way of doing anything and that it always gets removed.

crisbeto:
Updated @jelbourn. I went with a 500ms timeout, because the default transition is 400ms.

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -101,18 +103,28 @@ export class OverlayRef implements PortalHost {
...
+      };
+
+      backdropToDetach.classList.remove('md-overlay-backdrop-showing');
+      backdropToDetach.classList.remove(this._state.backdropClass);
+      backdropToDetach.addEventListener('transitionend', onTransitionEnd);
+
+      // If the backdrop doesn't have a transition, the `transitionend` event won't fire.
+      // In this case we make it unclickable and we try to remove it after a delay.
+      backdropToDetach.style.pointerEvents = 'none';
+      setTimeout(onTransitionEnd, 500);
```
rolandjitsu: I'm not sure if timeouts are reliable. What if at some point the transition timing is changed?

crisbeto: When I initially opened the PR, it was using `getComputedStyle`, however it can cause performance issues in some slower browsers.

crisbeto: Ah, forgot to mention that it would probably make more sense to switch this over to Angular's animation API. This fix should suffice for now, though.

rolandjitsu: Just wanted to point out that there might be a better way of handling this, but if this can make it to the next alpha I guess it works 👍

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -101,18 +103,28 @@ export class OverlayRef implements PortalHost {
     let backdropToDetach = this._backdropElement;
 
     if (backdropToDetach) {
-      backdropToDetach.classList.remove('md-overlay-backdrop-showing');
-      backdropToDetach.classList.remove(this._state.backdropClass);
-      backdropToDetach.addEventListener('transitionend', () => {
-        backdropToDetach.parentNode.removeChild(backdropToDetach);
+      let onTransitionEnd = () => {
```
jelbourn: Since this can be used in two ways now, we should rename it to something like `finishDetach`

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -101,18 +103,28 @@ export class OverlayRef implements PortalHost {
...
         // removing the backdrop. If that is the case, only clear the backdrop reference if it
         // is still the same instance that we started to remove.
         if (this._backdropElement == backdropToDetach) {
           this._backdropElement = null;
         }
-      });
+      };
+
+      backdropToDetach.classList.remove('md-overlay-backdrop-showing');
+      backdropToDetach.classList.remove(this._state.backdropClass);
```
jelbourn: Do we need to remove the `backdropClass`? Wouldn't this potentially cause the backdrop to appear in cases where we use the class to hide it?

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -101,18 +103,28 @@ export class OverlayRef implements PortalHost {
...
+      backdropToDetach.addEventListener('transitionend', onTransitionEnd);
+
+      // If the backdrop doesn't have a transition, the `transitionend` event won't fire.
+      // In this case we make it unclickable and we try to remove it after a delay.
+      backdropToDetach.style.pointerEvents = 'none';
```
jelbourn: Need unit test to assert that `pointerEvents` is set.

src/lib/core/overlay/overlay-ref.ts
crisbeto: That's the custom class that can be added to the backdrop. It has been handled like that since https://github.com/angular/material2/commit/0b54668b4da909c4704135327b2d877446a7a295, I only moved it around.

crisbeto:
Updated with a unit test for the `pointer-events` and renamed callback.

jelbourn:
LGTM

PR #1717, opened 2016-11-03
fix(progress-bar): bar being thrown off by parent's text-align

### LINKED ISSUE #1165: Determinate progress-bar breaks in `text-align: right`

Bug

Progress bar is not affected by text-align right context.

Determinate progress bar display fails in `text-align: right;` context.

Place it inside any container that has text-alignment set to  `right`
#### Plunker

http://plnkr.co/edit/rJCIiDvu4ZcQhYjLOK99?p=preview

It's easy enough to work around, but a block-level element shouldn't break on text-alignment.

Windows, Chrome, Angular2 RC 4/5/6, Material alpha.6 - alpha.7.4

Nope. Thanks for the great work!

crisbeto:
Fixes the `md-progress-bar`'s bar being positioned weirdly, if it's parent has `text-align: right`.
Note that this doesn't require any special styling for RTL since we use a transform rotation to flip the styling.

Fixes #1165.

### DIFF
--- a/src/lib/progress-bar/progress-bar.scss
+++ b/src/lib/progress-bar/progress-bar.scss
@@ -62,6 +62,7 @@ $md-progress-bar-piece-animation-duration: 250ms !default;
     height: 100%;
     position: absolute;
     width: 100%;
+    left: 0;
   }
 
   &[mode='query'] {


jelbourn:
LGTM

I was about to assign the issue to you when I saw you already had a PR.

PR #1722, opened 2016-11-04
chore(forbidden-identifiers): detect incorrect line endings

DevVersion:
* Removes stale scope checking (from `@angular2-material/$COMPONENT` times)
* Adds support for error grouping (useful when having a file full of invalid line endings)
* Moves chain functions into external functions to easily 

understand procession.

> Here you can see the error grouping - Notice that we don't show any special message for the line endings (because to keep the script small and easy to maintain)

### DIFF
--- a/scripts/ci/forbidden-identifiers.js
+++ b/scripts/ci/forbidden-identifiers.js
@@ -104,7 +104,7 @@ function groupErrors(errors) {
     if (initialError && isSimilarError(error)) {
       previousLine = error.lineNumber;
 
-      // Overwrite the lineNumber with a string representing the range of lines.
+      // Overwrite the lineNumber with a string, which indicates the range of lines.
       initialError.lineNumber = `${initialLine}-${previousLine}`;
 
       return false;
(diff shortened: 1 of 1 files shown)


scripts/ci/forbidden-identifiers.js
```diff
@@ -47,69 +46,105 @@ findTestFiles()
...
+
+    return true;
   });
 
+  /** Checks whether the given error is similar to the previous one. */
+  function isSimilarError(error) {
+    return initialError.fileName === error.fileName &&
+           initialError.statement === error.statement &&
+           previousLine === (error.lineNumber - 1);
+  }
```
DevVersion: I know this is not very pretty, but we don't have any Lodash helpers here, and this seems to be an elegant approach.

jelbourn:
@hansl can you take a look?

DevVersion:
@hansl Do you have time to take a quick look at this? :)

hansl:
Hi @DevVersion,

Sorry, this fell off a bit. Looking at it now.

hansl:
LGTM for now.

@DevVersion What I'd like to see though is moving away from a `forbidden-identifier` file and using custom `tslint` rules instead. It would be better and less hacky (since reporting is done by tslint, for example). See documentation here: https://www.npmjs.com/package/tslint#custom-rules

What do you think?

DevVersion:
@hansl - Great idea. I can do a follow-up PR to this one with the custom TS-Lint rules-

hansl:
Yes, a follow up PR. This one is okay (for now).

PR #1724, opened 2016-11-04
chore(ci): ensure browserstack binaries are latest

DevVersion:
* The Browserstack versions are always fetched from the cache, and never update them self. The `TunnelWrapper` package does not update the binaries if a newer version is available (that's not implemented yet [here](https://github.com/pghalliday/node-BrowserStackTunnel))

* It might be good to get rid of that `wrapper` in the future and do something similar to the saucelabs scripts

**Note**: Right now older versions can cause issues like 
- https://travis-ci.org/angular/material2/jobs/173114233
- https://travis-ci.org/angular/material2/jobs/173114074

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@types/minimist": "^1.1.28",
     "@types/node": "^6.0.34",
     "@types/run-sequence": "0.0.27",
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

Issue #421, opened 2016-05-10
md-fab: High Contrast mode has no focus indicator


hansl:
At all.

jelbourn:
@crisbeto since you're on Windows, could you do a pass on the components in high-contrast mode and add styles to make things better?

crisbeto:
Here's what I got:

- Fab buttons are barely visible when focused and impossible in high contrast mode.
- Really hard to see whether raised buttons (accent, primary, warn) are focused.
- Button toggles don't have any focus indication.
- Checkboxes don't have focus indication.
- Radio buttons don't have focus indication.
- The tooltips don't trigger on focus.
- Sliders don't have focus indication and keyboard support.

I'll try to take care of most of them, although the slider and tooltip will need a little bit more work.

PR #1730, opened 2016-11-04
fix(checkbox): missing focus indicator

crisbeto:
Adds a ripple when a checkbox is focused.

Referencing #421.

PR #1731, opened 2016-11-04
fix(radio): radio buttons not being a tab stop and missing focus indication

crisbeto:
* Fixes the radio buttons not being tabable.
* Adds focus indication to the radio buttons.

Referencing #421.

### DIFF
--- a/src/lib/radio/radio.html
+++ b/src/lib/radio/radio.html
@@ -1,6 +1,7 @@
 <!-- TODO(jelbourn): render the radio on either side of the content -->
 <!-- TODO(mtlin): Evaluate trade-offs of using native radio vs. cost of additional bindings. -->
-<label [attr.for]="inputId" class="md-radio-label">
+<!-- TODO(crisbeto): only show focus indication if focus came from a keyboard event -->
+<label [attr.for]="inputId" class="md-radio-label" [class.md-ripple-focused]="_isFocused">
   <!-- The actual 'radio' part of the control. -->
   <div class="md-radio-container">
     <div class="md-radio-outer-circle"></div>
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -574,6 +574,24 @@ describe('MdRadio', () => {
         expect(document.activeElement).toBe(fruitRadioNativeInputs[i]);
       }
     });
+
+    it('should add a ripple when the native input element is focused', () => {
+      let hostElement = seasonRadioInstances[0].getHostElement() as HTMLElement;
+      let input = hostElement.querySelector('input') as HTMLInputElement;
+      let label = hostElement.querySelector('label') as HTMLLabelElement;
+
+      expect(label.classList).not.toContain('md-ripple-focused');
+
+      dispatchEvent('focus', input);
+      fixture.detectChanges();
+
+      expect(label.classList).toContain('md-ripple-focused');
+
+      dispatchEvent('blur', input);
+      fixture.detectChanges();
+
+      expect(label.classList).not.toContain('md-ripple-focused');
+    });
   });
 });
 


src/lib/radio/radio.ts
```diff
@@ -239,6 +239,9 @@ export class MdRadioButton implements OnInit {
   @HostBinding('class.md-radio-focused')
   _isFocused: boolean;
 
+  /** Tabindex for the input element. */
+  @Input() tabindex: number = 0;
```
jelbourn: The property should be `tabIndex`

I'm confused why this wasn't working, though- focus should have gone to the native radio without having to explicitly set a tabindex. Where were you seeing the problem?

src/lib/radio/radio.spec.ts
```diff
@@ -485,6 +485,28 @@ describe('MdRadio', () => {
 
       expect(fruitRadioNativeInputs[0].getAttribute('aria-labelledby')).toBe('uvw');
     });
+
+    it('should make the native input element a tab stop', () => {
+      expect(fruitRadioNativeInputs[0].tabIndex).toBe(0);
+    });
```
jelbourn: Being a tab stop is something better covered by an e2e test where we can actually press the tab key, but it is good to have a unit test that asserts we're forwarding the tabIndex property.

src/lib/radio/radio.html
```diff
@@ -1,6 +1,6 @@
 <!-- TODO(jelbourn): render the radio on either side of the content -->
 <!-- TODO(mtlin): Evaluate trade-offs of using native radio vs. cost of additional bindings. -->
-<label [attr.for]="inputId" class="md-radio-label">
+<label [attr.for]="inputId" class="md-radio-label" [class.md-ripple-focused]="_isFocused">
```
jelbourn: Can you add a TODO to only show this focus indicator when the focus came from a keyboard event?

src/lib/radio/radio.ts
crisbeto: I don't remember right now, but I was testing against the local demo app. I'm retrying now and it works fine without the tabindex, so I assume it was because I wasn't seeing the ripple. I've removed  the tabindex changes from the PR.

crisbeto:
Updated the PR to address the comments.

src/lib/radio/radio.html
```diff
@@ -1,6 +1,7 @@
 <!-- TODO(jelbourn): render the radio on either side of the content -->
 <!-- TODO(mtlin): Evaluate trade-offs of using native radio vs. cost of additional bindings. -->
-<label [attr.for]="inputId" class="md-radio-label">
+<!-- TODO(crisbeto): only show focus indication if focus came from a keyboard event -->
+<label [attr.for]="inputId" class="md-radio-label" [class.md-ripple-focused]="_isFocused">
```
jelbourn: I pulled down the PR and tested it out; I found that changing the selected radio within a group somewhat quickly (such as when using the keyboard) leaves a trailing grey circle on the radio.

jelbourn:
@crisbeto what's the status on this one?

crisbeto:
Sorry, @jelbourn I missed this one in between all of the notifications. The issue is that the ripple still animates for a little after you move focus away from the radio button. It can be fixed by changing [this line](https://github.com/angular/material2/pull/1731/files#diff-347bf74e8fa8f126282ea3f706f9eef6R9) to `*ngIf="!_isRippleDisabled() && _isFocused"`, however now there won't be ripples when you click multiple times on a radio button.

 I tried setting up the the `blur` event so that it checks if the focus is still within the radio button, but the `event.relatedTarget` comes up `null` occasionally (not sure why) and using `document.activeElement` requires that we use a timeout. Perhaps we should try to find some more generic solution within the ripple? 

Alternatively this could become easier to fix once we have the ability to know where the last focus event came from (whether it was a click or a keyboard event).

andrewseguin:
Please rebase

crisbeto:
Looks like much of the setup for this and #1730 has changed @andrewseguin. I'll probably end up closing them and redoing.

andrewseguin:
Sounds good, apologies for the delay in getting this merged when you sent it out

crisbeto:
Closing and will re-do based on the current APIs.

PR #1733, opened 2016-11-04
fix(button-toggle): missing hover state

crisbeto:
Adds a hover effect to the button toggles.

Referencing #421.


jelbourn:
LGTM

PR #1734, opened 2016-11-04
fix(md-input) set the line-height to normal.

### LINKED ISSUE #1082: [Chrome] md-input up/down arrows showing when input type is not set to number

Bug

In all browsers the md-input up/down buttons should be visible only when input type is set to "number".

In Chrome the md-input up/down buttons show even when type="number" is not set and when "style='line-height: 1'" is set on body such as when using this reset.css : http://meyerweb.com/eric/tools/css/reset/
Firefox shows no issue there.

If normalize.css is used instead of reset.css then the issue disappears.

http://plnkr.co/edit/suhu6SElSc2GwEEXMKq2?p=preview

Removing "line-height" on body solves the problem. 
So does removing the surrounding md-card-content

Chrome Version 52.0.2743.116 m
Chromium Version 51.0.2704.79 Ubuntu 16.04 (64-bit)
"@angular/core": "2.0.0-rc.5",
"@angular2-material/core": "^2.0.0-alpha.7-4",
"@angular2-material/input": "^2.0.0-alpha.7-4"

mmalerba:
fixes #1082

### DIFF
--- a/src/lib/input/input.scss
+++ b/src/lib/input/input.scss
@@ -19,6 +19,7 @@ md-input, md-textarea {
   display: inline-block;
   position: relative;
   font-family: $md-font-family;
+  line-height: normal;
 
   // To avoid problems with text-align.
   text-align: left;


jelbourn:
R: @hansl

hansl:
LGTM

PR #1735, opened 2016-11-05
fix(radio): Fix radio group behavior when radio button value changed

### LINKED ISSUE #1152: Setting md-radio-group value dynamically does not change 'checked' property of md-radio-button

**Bug:**
Setting md-radio-group value dynamically does not change 'checked' property of md-radio-button (md-radio-checked class is not enabled for this radio button)

When md-radio-group value is changed to one of the md-radio-button values, the corresponding radio button should get checked (md-radio-checked class should be enabled for the radio button)

Initial md-radio-group value ensures that correct radio button is checked when application starts but changing the md-radio-group value dynamically does not do the same.

1) Visit plunker: http://plnkr.co/edit/rzemQv9oNEpt4yOYuT0u?p=preview
Initial value of md-radio-group is set to 'Jack' which is same as the value of '1st radio button (Custom Name)'. Here the Custom Name radio button is checked, so it works as required.

2) Select 2nd radio button (Default), then the Default radio button gets checked, md-radio-group value is also updated.

3) Now change the custom name to say 'xyz', the md-radio-group value & the 'Custom Name' radio button values get updated to 'xyz' but the md-radio-checked class is not set for 'Custom Name' radio.

The example in plunker is with Angular 2.0.0-rc.1 & material 2.0.0-alpha.5 but the problem is also reproduced with the latest Angular 2.0.0-rc.5 & material 2.0.0-alpha.7-4 which I am using at work.

tinayuangao:
When radio button value changed, if the new value matches group's value, the button should be checked.

Closes #1152

R: @jelbourn

### DIFF
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -274,6 +274,29 @@ describe('MdRadio', () => {
       expect(groupInstance.value).toBe('water');
       expect(changeSpy).not.toHaveBeenCalled();
     });
+
+    it(`should update checked status if changed value to radio group's value`, () => {
+      let changeSpy = jasmine.createSpy('radio-group change listener');
+      groupInstance.change.subscribe(changeSpy);
+      groupInstance.value = 'apple';
+
+      expect(changeSpy).not.toHaveBeenCalled();
+      expect(groupInstance.value).toBe('apple');
+      expect(groupInstance.selected).toBeFalsy('expect group selected to be null');
+      expect(radioInstances[0].checked).toBeFalsy('should not select the first button');
+      expect(radioInstances[1].checked).toBeFalsy('should not select the second button');
+      expect(radioInstances[2].checked).toBeFalsy('should not select the third button');
+
+      radioInstances[0].value = 'apple';
+
+      fixture.detectChanges();
+
+      expect(groupInstance.selected).toBe(
+        radioInstances[0], 'expect group selected to be first button');
+      expect(radioInstances[0].checked).toBeTruthy('expect group select the first button');
+      expect(radioInstances[1].checked).toBeFalsy('should not select the second button');
+      expect(radioInstances[2].checked).toBeFalsy('should not select the third button');
+    });
   });
 
   describe('group with ngModel', () => {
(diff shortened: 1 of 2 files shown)


src/lib/radio/radio.spec.ts
```diff
@@ -274,6 +274,29 @@ describe('MdRadio', () => {
       expect(groupInstance.value).toBe('water');
       expect(changeSpy).not.toHaveBeenCalled();
     });
+
+    it('should update checked status if changed value to radio group\'s value', () => {
```
jelbourn: Instead of escaping the `'`, you can use a different type of quote for the string (usually backtick)

src/lib/radio/radio.ts
```diff
@@ -127,6 +127,13 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
       this._value = newValue;
 
       this._updateSelectedRadioFromValue();
+      this._checkSelectedButton();
+    }
+  }
+
+  _checkSelectedButton() {
```
jelbourn: How about `_checkSelectedRadioButton` ?

src/lib/radio/radio.ts
```diff
@@ -180,14 +185,13 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
...
-      } else if (this.value == null) {
-        this.selected = null;
-        this._radios.forEach(radio => { radio.checked = false; });
-      }
+      this._selected = null;
+      this._radios.forEach(radio => {
+        radio.checked = this.value == radio.value;
+        if (radio.checked) {
+          this._selected = radio;
+        }
```
jelbourn: Are we losing anything by no longer going through the setter for `selected` here?

tinayuangao: The `selected()` does 3 things: update `_selected` (which is done); update `value` (`_updateSelectedRadioFromValue` is always called from value change) ; check `selected` radio button (we get the `_selected` value from checked radio buttons, and `checked` is set on line 190).

jelbourn:
LGTM

PR #1736, opened 2016-11-05
fix(button): ensure icons are aligned vertically.

### LINKED ISSUE #1093: md-icon-button in md-toolbar has icon displaced

Bug

The icon should be centered in the button.

The icon is displaced 8px upwards to be very close to outer edge of the button.

http://plnkr.co/edit/ljTlfhIkQeksCEGimtXM?p=preview

Angular 2.0.0-rc.5, angular2-material-alpha.7-4
Replicated on Windows 10 using Chrome, Firefox, Edge and IE 11.

Changing the CSS style `line-height` from 24px to 40px (to correspond with the width and height of the button) seems to solve the issue.

DevVersion:
* The icon buttons are never a perfect cirlce, because they can shrink (depending on siblings).
  
   In Material 1 we solved this by adding `min-width/height`, but having flex-shrink is more elegant here (it allows developers to easily overwrite height and width)

* Sets the line-height for icons only to `i` and `md-icon` elements (as same as for fab buttons)

  > Can be made more generic in the future, with another mixin.

Fixes #1093.

### DIFF
--- a/src/lib/button/_button-base.scss
+++ b/src/lib/button/_button-base.scss
@@ -104,11 +104,15 @@ $md-mini-fab-padding: 8px !default;
 @mixin md-fab($size, $padding) {
   @extend %md-raised-button;
 
+  // Reset the min-width from the button base.
   min-width: 0;
+
   border-radius: $md-fab-border-radius;
   width: $size;
   height: $size;
   padding: 0;
+
+  flex-shrink: 0;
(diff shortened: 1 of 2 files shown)


jelbourn:
R: @tinayuangao

tinayuangao:
LGTM

PR #1738, opened 2016-11-05
chore(changelog): fix missing module for adding streams

DevVersion:
* Use the `merge2` package because the `add-stream` package has been accidentally removed with SHA abcc6af54df6125075dc465715b246dfdbb5b364
* Do not show an error for a missing changelog file, when using `--force` mode.
* Removes unused `e2e.sh` file


jelbourn:
LGTM

Issue #1681, opened 2016-11-01
Local dev setup triggering multiple reloads


crisbeto:
When recompiling a file during the `serve:devapp`, the dev server triggers multiple reloads. The same is true when running the unit tests. This can be somewhat disruptive when debugging unit tests since it ends up running them multiple times in parallel.

crisbeto:
I've been looking into this and it seems like it's more complicated than expected:

1. Karma doesn't have a proper way of debouncing file changes. There's the `autoWatchBatchDelay` option, but it only specifies how much time to wait from the first file change to when the browser reloads. I've tried playing around Karma's API in order to trigger the reloads manually, when the files are done, but I wasn't able to get it working.
2. None of the Gulp servers with livereload seem to have the option to debounce. Some allow you to trigger a livereload at a certain time, but that would require going through and adding a callback to each of the build tasks, which may end up being hard to maintain.

PR #1741, opened 2016-11-05
chore(build): fix karma not exiting properly

crisbeto:
* Fixes the Karma tasks not exiting properly. This is an issue, because doing a keyboard interrupt doesn't stop the Gulp watchers that are going on in the background.
* Increase the time that Karma waits before starting to run tests. This should help mitigate the multiple reloads when running unit tests locally.

Related to #1681.

### DIFF
--- a/test/karma.config.ts
+++ b/test/karma.config.ts
@@ -59,6 +59,7 @@ export function config(config) {
     colors: true,
     logLevel: config.LOG_INFO,
     autoWatch: true,
+    autoWatchBatchDelay: 500,
 
     sauceLabs: {
       testName: 'material2',
--- a/tools/gulp/tasks/unit-test.ts
+++ b/tools/gulp/tasks/unit-test.ts
@@ -2,7 +2,7 @@ import gulp = require('gulp');
 import path = require('path');
 import gulpMerge = require('merge2');
 
-import {PROJECT_ROOT, DIST_COMPONENTS_ROOT} from '../constants';
+import {PROJECT_ROOT} from '../constants';
 import {sequenceTask} from '../task_helpers';
 
 const karma = require('karma');
@@ -48,10 +48,10 @@ gulp.task(':test:deps:inline', sequenceTask(':test:deps', ':inline-resources'));
  *
  * This task should be used when running unit tests locally.
  */
-gulp.task('test', [':test:watch'], (done: () => void) => {
+gulp.task('test', [':test:watch'], () => {
   new karma.Server({
     configFile: path.join(PROJECT_ROOT, 'test/karma.conf.js')
-  }, done).start();
+  }).start();
 });
 
 /**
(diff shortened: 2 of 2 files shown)


jelbourn:
R: @hansl

test/karma.config.ts
```diff
@@ -59,6 +59,7 @@ export function config(config) {
     colors: true,
     logLevel: config.LOG_INFO,
     autoWatch: true,
+    autoWatchBatchDelay: 1000,
```
hansl: 1000 might be too slow. 500?

tools/gulp/tasks/unit-test.ts
```diff
@@ -48,20 +48,20 @@ gulp.task(':test:deps:inline', sequenceTask(':test:deps', ':inline-resources'));
  *
  * This task should be used when running unit tests locally.
  */
-gulp.task('test', [':test:watch'], (done: () => void) => {
+gulp.task('test', [':test:watch'], () => {
   new karma.Server({
     configFile: path.join(PROJECT_ROOT, 'test/karma.conf.js')
-  }, done).start();
+  }).start();
```
hansl: This is probably not the right fix. By default, Karma uses `process.exit` as the callback. We definitely want to exit Gulp gracefully.

crisbeto: I can switch it to:

```js
() => {
  done();
  process.exit();
}
```

I didn't include it initially, because it seemed redundant since `process.exit` kills the process anyway (which includes Gulp).

crisbeto:
Updated to both exit Gulp and the process, as well as reduced the `autoWatchBatchDelay` @hansl.

hansl:
@crisbeto: That's the thing; we don't want to exit the process. We want Gulp to finish the tasks gracefully. Isn't there a better way to kill the watch tasks?

crisbeto:
I'm not sure interrupting a task while it's running is really an issue here, because Karma always runs after all of the build steps. I'm not sure there's a better way to kill it (considering that Karma's default is `process.exit`). 

Also we should keep in mind that this is mainly for local development.

tools/gulp/tasks/unit-test.ts
```diff
@@ -51,7 +51,10 @@ gulp.task(':test:deps:inline', sequenceTask(':test:deps', ':inline-resources'));
 gulp.task('test', [':test:watch'], (done: () => void) => {
   new karma.Server({
     configFile: path.join(PROJECT_ROOT, 'test/karma.conf.js')
-  }, done).start();
+  }, (exitCode: number) => {
```
hansl: Okay you can remove the handler. Seems like there's no way to terminate a watch task... which sucks.

crisbeto: Done.

PR #1744, opened 2016-11-05
fix(gestures): remove drag event from config

### LINKED ISSUE #1025: drag events dataTransfer is undefined in RC5 alpha 7-2 with NgModule imports

Bug: Since adding in Md*Module(s) and hammerjs, drag events no longer contain the dataTransfer property.

```
browser_adapter.js:84 TypeError: Cannot set property 'effectAllowed' of undefined
    at DraggableDirective.onDragStart (draggable.directive.ts:9)
    at DebugAppView._View_BusinessObjectFieldsListComponent1._handle_dragstart_0_0 (BusinessObjectFieldsListComponent.ngfactory.js:238)
    at view.js:381
    at dom_renderer.js:255
    at hammer_gestures.js:68
    at ZoneDelegate.invoke (zone.js:323)
    at Object.onInvoke (ng_zone_impl.js:53)
    at ZoneDelegate.invoke (zone.js:322)
    at Zone.runGuarded (zone.js:230)
    at NgZoneImpl.runInnerGuarded (ng_zone_impl.js:86)
```

This did not happen when directly using the Md\* components in our components.  Once we added the Md*Module(s) to NgModule and installed hammerjs this error arose.

drag events should have the dataTransfer property to be able to pass data in the dropped element

error above

(I'm working on it, but can't figure out how to get ng2 rc5 and md alpha7-2 all in the same plunker...)

Angular 2 RC5
Material alpha 7-2
hammerjs 2.0.8
Windows 8.1
Chrome and FF latest

It's worth noting that this error occurs in a component with NO MD components in use.  But it only happens if we NgModule import the MD components (needed elsewhere in the application).  It doesn't make sense, but our draggable div isn't an md-\* component!  But if I remove them from NgModule imports I can drag it!

crisbeto:
Removes the `drag` event from the gesture config.

Fixes #1025.

### DIFF
--- a/src/lib/core/gestures/MdGestureConfig.ts
+++ b/src/lib/core/gestures/MdGestureConfig.ts
@@ -7,9 +7,6 @@ export class MdGestureConfig extends HammerGestureConfig {
 
   /* List of new event names to add to the gesture support list */
   events: string[] = [
-    'drag',
-    'dragright',
-    'dragleft',
     'longpress',
     'slide',
     'slidestart',
@@ -41,14 +38,13 @@ export class MdGestureConfig extends HammerGestureConfig {
     // Notice that a HammerJS recognizer can only depend on one other recognizer once.
     // Otherwise the previous `recognizeWith` will be dropped.
     let slide = this._createRecognizer(pan, {event: 'slide', threshold: 0}, swipe);
-    let drag = this._createRecognizer(slide, {event: 'drag', threshold: 6}, swipe);
     let longpress = this._createRecognizer(press, {event: 'longpress', time: 500});
 
     // Overwrite the default `pan` event to use the swipe event.
     pan.recognizeWith(swipe);
 
     // Add customized gestures to Hammer manager
-    mc.add([swipe, press, pan, drag, slide, longpress]);
+    mc.add([swipe, press, pan, slide, longpress]);
 
     return mc;
   }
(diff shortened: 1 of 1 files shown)


crisbeto:
@jelbourn I'm not too familiar with the gesture setup, are we supposed to remove the `drag` [event from here as well](https://github.com/angular/material2/blob/b09465cda7cf8c74a5c903ee49d0366d644643d4/src/lib/core/gestures/MdGestureConfig.ts#L44)?

DevVersion:
@crisbeto I think it depends on what we are trying to archive.

Recently in b09465cda7cf8c74a5c903ee49d0366d644643d4, the `dragstart` and `dragend` have been removed, which means that the HammerJS drag event is kind of unusable right now. 

> We still have the `slide` event which is used by all components supporting panning gestures right now.

In my opinion, we should remove the drag recognizer and associated events here, and just use the `slide` event (it has a lower threshold -> feels better)

crisbeto:
Thanks @DevVersion,, removed.

src/lib/core/gestures/MdGestureConfig.ts
```diff
@@ -7,7 +7,6 @@ export class MdGestureConfig extends HammerGestureConfig {
 
   /* List of new event names to add to the gesture support list */
   events: string[] = [
-    'drag',
     'dragright',
     'dragleft',
```
DevVersion: Seems like you forgot to remove the `dragright` and `dragleft` here

jelbourn:
LGTM

PR #1745, opened 2016-11-05
chore(browserstack): properly teardown tunnel

DevVersion:
* Implements the browserstack teardown script.

### DIFF
--- a/scripts/browserstack/teardown_tunnel.sh
+++ b/scripts/browserstack/teardown_tunnel.sh
@@ -5,8 +5,16 @@ set -e -o pipefail
 
 echo "Shutting down Browserstack tunnel"
 
+PID=$(cat $BROWSER_PROVIDER_READY_FILE);
+
 # Resolving the PID from the readyfile.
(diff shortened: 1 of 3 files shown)


scripts/browserstack/teardown_tunnel.sh
```diff
@@ -4,5 +4,17 @@ set -e -o pipefail
...
+PID=$(cat $BROWSER_PROVIDER_READY_FILE);
+
+# Resolving the PID from the readyfile.
+kill $PID
+
+
+while [[ -n `ps -ef | grep $PID | grep -v "grep"` ]]; do
+  printf "."
+  sleep .5
+done
```
jelbourn: Is it worth capping how long we wait here?

DevVersion: I think it just helps ensuring that the Browserstack tunnel is properly turned off.

Normally the `while`  does not run *once*, because the process terminated synchronously before, but in some situations it might take longer (crashes etc.) and this might help us as well.

> Also added because the Saucelabs teardown script had the same.

PR #1747, opened 2016-11-05
chore(saucelabs): interrupt if saucelabs tunnel does not start

DevVersion:
* Adds a timeout to the Saucelabs wait script (as same as in the browserstack tunnel)
* Logs dots while waiting for the tunnel to be ready (same as in the saucelabs wait script)
* No longer calls `print-logs` file (File does not exist and also `browserstacktunnel-wrapper` prints necessary output to the console)

---
Fixes
> 

and Travis builds like [this one](https://travis-ci.org/angular/material2/jobs/173114071), which can run about **2 hours**

### DIFF
--- a/scripts/browserstack/waitfor_tunnel.sh
+++ b/scripts/browserstack/waitfor_tunnel.sh
@@ -2,18 +2,17 @@
 
 
 # Wait for Connect to be ready before exiting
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

PR #1748, opened 2016-11-06
fix(input): change value of empty number input to be null

### LINKED ISSUE #914: md-input type="number" required does not update control validation state

Bug

A required but empty md-input field should turn the control invalid.

Has the input been touched once it's `valid` prop stays `true` even if the the input is empty.

Enter a number and remove it again.
[Here is a plunker.](http://plnkr.co/edit/n3XmVBcBnqNquoVyBeu7?p=preview)

Validating forms should be working.

Tested with:
angular: '@2.0.0-rc.4'
forms: '@0.2.0'
material: '@2.0.0-alpha.6'
Chomre 52.0.2743.82

blaugold:
Previously the value would be NaN which would pass the 'required' validator
even though the input was empty.

Closes #914

### DIFF
--- a/src/lib/input/input.spec.ts
+++ b/src/lib/input/input.spec.ts
@@ -1,9 +1,6 @@
-import {
-  async,
-  TestBed,
-} from '@angular/core/testing';
-import {Component} from '@angular/core';
-import {FormsModule} from '@angular/forms';
+import {async, TestBed, fakeAsync, tick} from '@angular/core/testing';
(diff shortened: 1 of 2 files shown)


jelbourn:
R: @hansl

hansl:
We're refactoring the MdInput which will take care of this. Unless @jelbourn insists on requiring this for the next release, I'd rather close this as WAI.

jelbourn:
Closing as "Will be superceded by imminent md-input refactor"

PR #1667, opened 2016-11-01
feat(select): support disabling

kara:
This PR adds support for disabling the select and its options.  Needs to be rebased once #1657 and #1655 go in.

r: @jelbourn

PR #1750, opened 2016-11-06
fix(checkbox, input, radio, slide-toggle): implement setDisabledState from ControlValueAccessor

### LINKED ISSUE #1171: [input @alpha.8-1] Setting bound FormControl to disabled causes valueAccessor.setDisabledState error

crisbeto:
Implements the `setDisabledState` method from the `ControlValueAccessor` interface in all of the input-related components, in order to support disabling via reactive forms. Note that the `select` component is missing the implementation, however there's a pending PR for it already (#1667).

Fixes #1171.

### DIFF
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -8,6 +8,8 @@ import {
 import {
     NgControl,
     FormsModule,
+    ReactiveFormsModule,
+    FormControl,
 } from '@angular/forms';
 import {Component, DebugElement} from '@angular/core';
 import {By} from '@angular/platform-browser';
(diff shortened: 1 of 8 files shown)


jelbourn:
Could you add a unit test for each component that sets `disabled` on the component and then asserts that the form model is correctly marked as disabled?

crisbeto:
Added unit tests @jelbourn.

jelbourn:
LGTM

jelbourn:
@crisbeto can you rebase this PR?

crisbeto:
Rebased @jelbourn. For some reason the entire `slide-toggle.spec` got marked as a conflict.

DevVersion:
@crisbeto Yeah that's because we changed the Line endings on the whole file.

Dominionys:
so how it fix? npm from master? thanks.

PR #1752, opened 2016-11-06
feat(focus-trap): add the ability to specify a focus target

### LINKED ISSUE #1468: Provide way to set focus target when opening dialog

For dialogs with titles do not force focus onto the title cancel button (usually an X icon button in the right section of the title area). Or can we have this so that we can manually override. 

Focus the first non-button input instead

The cancel button is focused.

Create a custom dialog with a button as the first input HTML element in the document.

I understand the use case, especially around accessibility issues, but it would provide a better experience in this case not to have it.

crisbeto:
Adds the ability to specify an element that should take precedence over other focusable elements inside of a focus trap.

Fixes #1468.

**Note:** @jelbourn we discussed that we should have separate `start` and `end` directives, which we'd query via `@ContentChild` or `@ViewChild`, however I wasn't able to get it working. I was testing with the dialog demo and I think that either the element being nested in multiple other components is breaking it, or the `portal` (which is used to inject the dialog content) isn't working properly. I've also tried injecting the focus trap into a focus target directive, but it didn't work either.

### DIFF
--- a/src/lib/core/a11y/focus-trap.ts
+++ b/src/lib/core/a11y/focus-trap.ts
@@ -27,7 +27,10 @@ export class FocusTrap {
 
   /** Focuses the first tabbable element within the focus trap region. */
   focusFirstTabbableElement() {
-    let redirectToElement = this._getFirstTabbableElement(this.trappedContent.nativeElement);
+    let rootElement = this.trappedContent.nativeElement;
+    let redirectToElement = rootElement.querySelector('[md-focus-start]') as HTMLElement ||
+                            this._getFirstTabbableElement(rootElement);
+
     if (redirectToElement) {
       redirectToElement.focus();
     }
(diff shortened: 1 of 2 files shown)


src/lib/core/a11y/focus-trap.ts
```diff
@@ -47,6 +49,12 @@ export class FocusTrap {
       return root;
     }
 
+    let focusTarget = root.querySelector(FOCUS_TARGET_SELECTOR) as HTMLElement;
```
jelbourn: We don't want to call `querySelector` here- this method is used recursively, you would would potentially end up calling `querySelector` many times. This should happen outside of the recursive method.

src/lib/core/a11y/focus-trap.ts
```diff
@@ -1,6 +1,8 @@
 import {Component, ViewEncapsulation, ViewChild, ElementRef} from '@angular/core';
 import {InteractivityChecker} from './interactivity-checker';
 
+/** Selector for nodes that should have a higher priority when looking for focus targets. */
+const FOCUS_TARGET_SELECTOR = '[md-focus-target]';
```
jelbourn: I'd still like to have separate `md-focus-start` and `md-focus-end` attributes, even if they're not directives. With only one attribute, users either have to specify it exactly zero times or more than one time- a user just trying to set where the focus should start would also inadvertently be setting where the focus should wrap backwards to.

crisbeto:
Updated to address the comments.

jelbourn:
LGTM

Could you also update the dialog README with this?
(can be in a separate PR)

PR #1753, opened 2016-11-06
chore(toolbar): combine imports from same file

DevVersion:
* Moves two separated imports to one single import statement.


jelbourn:
LGTM

PR #1756, opened 2016-11-07
feat(projection): Host Projection service

hansl:
-

### DIFF
--- a/src/demo-app/projection/projection-demo.ts
+++ b/src/demo-app/projection/projection-demo.ts
@@ -1,0 +1,50 @@
+import {Component, ViewChild, ElementRef, OnInit, Input} from '@angular/core';
+import {DomProjectionHost, DomProjection} from '@angular/material';
+
+
+@Component({
+  selector: '[projection-test]',
+  template: `
+    <div class="demo-outer {{cssClass}}">
+      Before
+      <dom-projection-host><ng-content></ng-content></dom-projection-host>
+      After
+    </div>
+  `,
+  styles: [`
+    .demo-outer {
+      background-color: #663399;
+    }
+  `]
+})
+export class ProjectionTestComponent implements OnInit {
+  @ViewChild(DomProjectionHost) _host: DomProjectionHost;
+  @Input('class') cssClass: any;
+
+  constructor(private _projection: DomProjection, private _ref: ElementRef) {}
+
+  ngOnInit() {
+    this._projection.project(this._ref, this._host);
+  }
+}
+
+
+@Component({
+  selector: 'projection-app',
+  template: `
+    <div projection-test class="demo-inner">
+      <div class="content">Content: {{binding}}</div>
+    </div>
+    <br/>
+    <input projection-test [(ngModel)]="binding" [class]="binding" [ngClass]="{'blue': true}">
+    <input [(ngModel)]="binding" class="my-class" [ngClass]="{'blue': true}">
+  `,
+  styles: [`
+    .demo-inner {
+      background-color: #DAA520;
+    }
+  `]
+})
+export class ProjectionDemo {
+  binding: string = 'abc';
+}
--- a/src/lib/core/projection/projection.ts
+++ b/src/lib/core/projection/projection.ts
@@ -1,0 +1,87 @@
+import {Injectable, Directive, ModuleWithProviders, NgModule, ElementRef} from '@angular/core';
+
+
+// "Polyfill" for `Node.replaceWith()`.
+// cf. https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith
+function _replaceWith(toReplaceEl: HTMLElement, otherEl: HTMLElement) {
+  toReplaceEl.parentElement.replaceChild(otherEl, toReplaceEl);
+}
+
+
+@Directive({
+  selector: 'dom-projection-host'
+})
+export class DomProjectionHost {
+  constructor(public ref: ElementRef) {}
+}
+
+
+@Injectable()
+export class DomProjection {
+  /**
+   * Project an element into a host element.
+   * Replace a host element by another element. This also replaces the children of the element
+   * by the children of the host.
+   *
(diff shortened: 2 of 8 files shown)


src/lib/core/projection/projection.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  project(ref: ElementRef, host: MdProjectionHostDirective): void {
+    const componentEl = ref.nativeElement;
+    const hostEl = host.ref.nativeElement;
+    const childNodes = componentEl.childNodes;
+    let child = childNodes[0];
+
+    // Replace ComponentEl by all its children, in place.
+    this._replaceWith(componentEl, child);
+    while (childNodes.length) {
```
jelbourn: Reading `length` on a `NodeList` is notoriously slow; it's better to cache the length and `for` loop.
(here and below)

src/lib/core/projection/projection.ts
```diff
@@ -0,0 +1,56 @@
...
+
+
+@Injectable()
+export class MdProjectionService {
+  private _replaceWith(toReplaceEl: HTMLElement, otherEl: HTMLElement) {
+    toReplaceEl.parentElement.replaceChild(otherEl, toReplaceEl);
+  }
+
+  /**
+   * Project an element into a host element.
```
jelbourn: This comment needs some expansion; the key thing to communicate is that it's *only* the given element that is projected and *not* its content (children). Also worth mention when you might want to use this.

src/lib/core/projection/projection.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  project(ref: ElementRef, host: MdProjectionHostDirective): void {
+    const componentEl = ref.nativeElement;
```
jelbourn: `componentEl` -> `projectedEl`? (since it's not necessarily a component)

src/lib/core/projection/projection.ts
```diff
@@ -0,0 +1,56 @@
...
+    const hostEl = host.ref.nativeElement;
+    const childNodes = componentEl.childNodes;
+    let child = childNodes[0];
+
+    // Replace ComponentEl by all its children, in place.
```
jelbourn: Expand comment to something like
```ts
// We hoist all of the projected element's children out into the projected elements position
// because we *only* want to move the projected element and not its children.
```

src/demo-app/projection/projection-demo.ts
```diff
@@ -0,0 +1,50 @@
...
+    <div projection-test class="inner">
+      <div class="content">Content: {{binding}}</div>
+    </div>
+    <br/>
+    <input projection-test [(ngModel)]="binding" [class]="binding" [ngClass]="{'blue': true}">
+    <input [(ngModel)]="binding" class="my-class" [ngClass]="{'blue': true}">
+  `,
+  styles: [`
+    .inner {
+      background-color: blue;
```
jelbourn: nit: these colors are totally unreadable to me (when I run the demo)

src/lib/core/projection/projection.ts
```diff
@@ -0,0 +1,56 @@
...
+  selector: 'md-host'
+})
+export class MdProjectionHostDirective {
+  constructor(public ref: ElementRef) {}
+}
+
+
+@Injectable()
+export class MdProjectionService {
+  private _replaceWith(toReplaceEl: HTMLElement, otherEl: HTMLElement) {
```
hansl: We cannot replace all nodes by all nodes in the tree. Element is semantically accurate.

src/lib/core/projection/projection.ts
hansl: Done

src/lib/core/projection/projection.ts
hansl: See above.

src/lib/core/projection/projection.ts
hansl: Done.

src/lib/core/projection/projection.ts
```diff
@@ -0,0 +1,56 @@
...
+@Directive({
+  selector: 'md-host'
+})
+export class MdProjectionHostDirective {
+  constructor(public ref: ElementRef) {}
+}
+
+
+@Injectable()
+export class MdProjectionService {
```
hansl: Done.

src/demo-app/projection/projection-demo.ts
```diff
@@ -0,0 +1,50 @@
...
+  selector: '[projection-test]',
+  template: `
+    <div class="outer {{cssClass}}">
+      Before
+      <md-host><ng-content></ng-content></md-host>
+      After
+    </div>
+  `,
+  styles: [`
+    .outer {
```
hansl: Done.

src/lib/core/projection/projection.ts
hansl: Done.

src/lib/core/projection/projection.ts
```diff
@@ -0,0 +1,56 @@
...
+   * Project an element into a host element.
+   */
+  project(ref: ElementRef, host: MdProjectionHostDirective): void {
+    const componentEl = ref.nativeElement;
+    const hostEl = host.ref.nativeElement;
+    const childNodes = componentEl.childNodes;
+    let child = childNodes[0];
+
+    // Replace ComponentEl by all its children, in place.
+    this._replaceWith(componentEl, child);
```
hansl: Okay.

src/lib/core/projection/projection.ts
hansl: Done.

hansl:
@jelbourn please re-review

kara:
@hansl  I think you need to add `MdProjectionHostDirective` to the module definition. It's failing presubmit.

hansl:
@kara could you rerun the presubmits? It should pass now.

kara:
Absolutely! Momentito.

PR #1758, opened 2016-11-07
fix(slider): correctly detect when sidenav align changes.

### LINKED ISSUE #1236: Sidenav backdrop problem when changing align dynamically from left to right
I have a push state sidenav which is aligned left or right depending on the size of the browser.
`<md-sidenav align="{{alignAttr}}" mode="push">`
I am detecting this on browser resize.

The sidenav push mechanism works fine, when align is set to start it appears from left and from right when set to end. However the main window which is under the black transparent backdrop is animated in the wrong direction when the align attribute changes dynamically.

mmalerba:
Fixes #1236

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -48,6 +48,7 @@ export class MdDuplicatedSidenavError extends MdError {
     '[class.md-sidenav-over]': '_modeOver',
     '[class.md-sidenav-push]': '_modePush',
     '[class.md-sidenav-side]': '_modeSide',
+    '[class.md-sidenav-invalid]': '!valid',
   },
   changeDetection: ChangeDetectionStrategy.OnPush,
   encapsulation: ViewEncapsulation.None,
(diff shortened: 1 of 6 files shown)


src/lib/sidenav/sidenav.ts
```diff
@@ -40,14 +39,49 @@ export class MdDuplicatedSidenavError extends MdError {
...
+  @Input()
+  get align() {
+    return this._align;
+  }
+  set align(value) {
+    // Make sure we have a valid value.
+    value = (value == 'end') ? 'end' : 'start';
+    if (value != this._align) {
+      this._align = value;
+      this.onAlignChanged.emit();
```
jelbourn: Why not just directly call `validateDrawers` here instead of going through an observable?

src/lib/sidenav/sidenav.ts
```diff
@@ -40,14 +39,49 @@ export class MdDuplicatedSidenavError extends MdError {
...
+  /** Whether this md-sidenav is part of a valid md-sidenav-layout configuration. */
+  get valid() {
+    return this._valid;
+  }
+  set valid(value) {
+    value = coerceBooleanProperty(value);
+    if (!value) {
+      this.close();
+    }
+    this._valid = value;
```
jelbourn: I'd do this as 
```ts
this._valid = coerceBooleanProperty(value);

if (!this._valid) {
  this.close();
}
```

I generally like to avoid overwriting function arguments.  

Can you also add a comment explaining why you want to close the sidenav in this case?

src/lib/sidenav/sidenav.ts
```diff
@@ -275,48 +313,74 @@ export class MdSidenavLayout implements AfterContentInit {
     }
   }
 
-  /** TODO: internal */
+  /** @override */
```
jelbourn: I don't think the `@override` is necessary

mmalerba: Is there something else we can put here to indicate that it's being implemented as part of an interface? The rest of the TODO: internal's we prefixed with an underscore, but we can't do that here.

jelbourn:
Can you also update the README to mention this invalid state?

mmalerba:
Done

gatuteck:
Hi Angular Team, I have used sidenav component for displaying menu.  When I click on the menu item  it won't disappear but redirect me to that page and menu is still there.  I have used all three modes (side, push, over). 
http://home.developmentserver.me/

jelbourn:
@gatuteck You can file bug reports here: https://github.com/angular/material2/issues. Be sure to follow the issue template.

Commenting on random pull requests doesn't accomplish anything.

PR #1759, opened 2016-11-07
fix(slider): keyboard support

mmalerba:
↑/→ = +1 step
↓/← = -1 step
page up = +10 step
page down = -10 step
home = jump to min
end = jump to max

blocked on #1663


src/lib/slider/slider.ts
```diff
@@ -43,10 +43,19 @@ export class MdSliderChange {
...
     '(blur)': '_onBlur()',
     '(click)': '_onClick($event)',
+    '(keydown.arrowdown)': '_increment($event, -1)',
+    '(keydown.arrowleft)': '_increment($event, -1)',
+    '(keydown.arrowright)': '_increment($event, 1)',
+    '(keydown.arrowup)': '_increment($event, 1)',
+    '(keydown.end)': '_onEndKeyPressed($event)',
+    '(keydown.home)': '_onHomeKeyPressed($event)',
+    '(keydown.pagedown)': '_increment($event, -10)',
+    '(keydown.pageup)': '_increment($event, 10)',
```
jelbourn: Unfortunately, we can't use these keydown shorts right now; it uses `KeyboardEvent.key`, which isn't supported in Safari or in WebDriver. So we've got to go back to the single-handler-plus-switch.

src/lib/slider/slider.ts
```diff
@@ -254,6 +263,30 @@ export class MdSlider implements ControlValueAccessor {
     this.onTouched();
   }
 
+  /** Increments the slider by the given number of steps (negative number decrements. */
```
jelbourn: Missing closing paren

src/lib/slider/slider.ts
mmalerba: ugh, and we can't just polyfill it?

jelbourn: I want to avoid adding any polyfills that Angular itself doesn't already use

src/lib/slider/slider.ts
```diff
@@ -254,6 +266,46 @@ export class MdSlider implements ControlValueAccessor {
...
+        this._increment(1);
+        break;
+      case DOWN_ARROW:
+        this._increment(-1);
+        break;
+      default:
+        return;
+    }
+
+    event.preventDefault();
```
jelbourn: Should we only `preventDefault` when it's one of the keys we care about? E.g., pressing spacebar should probably still scroll the page.

mmalerba:
```
default:
  return;
```
:p

jelbourn: Ah, I missed that. Could you add a comment at the return like 
```ts
// Return if the key is one that we don't specifically handle so we
// don't call preventDefault on it.
```

jelbourn:
LGTM

PR #1761, opened 2016-11-07
fix(menu): reposition menu if it would open off screen

### LINKED ISSUE #1540: Should menu expand the page when positioned in the bottom?
At https://material2-app.firebaseapp.com/ the menu positioned at the bottom of the page. And when opened the menu is hidden and you have to scroll down to see more of it.

Is this the desired behavior or should/could it open upward instead?

kara:
Closes #1540.

### DIFF
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -22,9 +22,10 @@ import {
     TemplatePortal,
     ConnectedPositionStrategy,
     HorizontalConnectionPos,
-    VerticalConnectionPos
+    VerticalConnectionPos,
 } from '../core';
 import { Subscription } from 'rxjs/Subscription';
+import {MenuPositionX, MenuPositionY} from './menu-positions';
 
 /**
  * This directive is intended to be used in conjunction with an md-menu tag.  It is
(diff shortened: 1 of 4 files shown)


src/lib/menu/menu-trigger.ts
```diff
@@ -181,14 +183,33 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
+        this._originPos(posX, fallbackY), this._overlayPos(posX, fallbackY))
+      .withFallbackPosition(
+        this._originPos(fallbackX, fallbackY), this._overlayPos(fallbackX, fallbackY));
+  }
+
+
+  /**  Converts the designated point into an OriginConnectionPosition. */
+  private _originPos(x: HorizontalConnectionPos,
+                     y: VerticalConnectionPos): OriginConnectionPosition {
+    return {originX: x, originY: y} as OriginConnectionPosition;
```
jelbourn: Is the `as OriginConnectionPosition` necessary? The types should be structurally identical.

kara:
Realized I need to change the animation origin point if a fallback is used, so this will need some more work.

src/lib/menu/menu-trigger.ts
```diff
@@ -191,19 +194,52 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
   }
 
   /**
+   * Listens to changes in the position of the overlay and sets the correct classes
+   * on the menu based on the new position. This ensures the animation origin is always
+   * correct, even if a fallback position is used for the overlay.
+   */
+  private _subscribeToPositions(position: ConnectedPositionStrategy): void {
+    this._positionSubscription = position.onPositionChange.subscribe(
+        (change: ConnectedOverlayPositionChange) => {
```
jelbourn: Shouldn't need the type here

PR #1762, opened 2016-11-07
chore(dev-app): prevent multiple reloads per file change

### LINKED ISSUE #1681: Local dev setup triggering multiple reloads

crisbeto:
Prevents the local dev app from reloading 5-6 times whenever a file changes. Adds a filter that only watches for the bundle (which also contains all of the CSS and HTML), theme changes and changes to the demo apps.

Fixes #1681.

### DIFF
--- a/tools/gulp/task_helpers.ts
+++ b/tools/gulp/task_helpers.ts
@@ -185,10 +185,15 @@ export function vendorTask() {
     }));
 }
 
+export type livereloadOptions = boolean | {
+  enable: boolean;
+  filter: (filename: string, callback: (isAllowed: boolean) => void) => void;
+}
 
 /** Create a task that serves the dist folder. */
-export function serverTask(liveReload = true,
+export function serverTask(liveReload: livereloadOptions = true,
                            streamCallback: (stream: NodeJS.ReadWriteStream) => void = null) {
+
   return () => {
     const stream = gulp.src('dist').pipe(gulpServer({
       livereload: liveReload,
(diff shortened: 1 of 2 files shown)


jelbourn:
R: @hansl

tools/gulp/task_helpers.ts
```diff
@@ -187,7 +187,7 @@ export function vendorTask() {
 
 
 /** Create a task that serves the dist folder. */
-export function serverTask(liveReload = true,
+export function serverTask(liveReload: boolean | { enable: boolean, filter: Function } = true,
```
hansl: `Function` => `(filename: string, callback: (boolean) => void)`

crisbeto:
Updated @hansl.

tools/gulp/task_helpers.ts
```diff
@@ -185,10 +185,15 @@ export function vendorTask() {
     }));
 }
 
+export type livereloadOptions = boolean | {
+  enable: boolean,
+  filter: (filename: string, callback: (isAllowed: boolean) => void) => void;
+}
```
hansl: Semicolon.

tools/gulp/task_helpers.ts
```diff
@@ -185,10 +185,15 @@ export function vendorTask() {
...
 
 /** Create a task that serves the dist folder. */
-export function serverTask(liveReload = true,
-                           streamCallback: (stream: NodeJS.ReadWriteStream) => void = null) {
+export function serverTask(liveReload: livereloadOptions = true,
+  streamCallback: (stream: NodeJS.ReadWriteStream) => void = null) {
```
hansl: Left align with `liveReload` above.

PR #1771, opened 2016-11-08
feat(menu): Added ability to show the menu overlay around the menu trigger

trshafer:
Currently the menu is fixed to show over the menu trigger.

This adds a (rather blunt) boolean to decide if the menu should be shown above or below the trigger. I thought about changing the possibilities to `overlay-position: default | above | below | left | right;` But that felt like a bit too much flexibility. I think some version of this feature is really important for those designing their own menu. Let me know what you think.

Thanks,
Thomas

Here is the menu below the trigger:
Menu Below

Here is the menu above the trigger:
Menu Above

### DIFF
--- a/src/lib/menu/README.md
+++ b/src/lib/menu/README.md
@@ -115,7 +115,8 @@ Output:
 ### Customizing menu position
 
 By default, the menu will display after and below its trigger.  You can change this display position
-using the `x-position` (`before | after`) and `y-position` (`above | below`) attributes.
+using the `x-position` (`before | after`) and `y-position` (`above | below`) attributes. The menu
+can be positioned over the menu button or outside using `overlapTrigger` (`true | false`).
 
 *my-comp.html*
 ```html
@@ -148,6 +149,7 @@ also adds `aria-hasPopup="true"` to the trigger element.
 | --- | --- | --- |
 | `x-position` | `before | after` | The horizontal position of the menu in relation to the trigger. Defaults to `after`. |
 | `y-position` | `above | below` | The vertical position of the menu in relation to the trigger. Defaults to `below`. |
+| `overlapTrigger` | `true | false` | Whether to have the menu show on top of the menu trigger or outside. Defaults to `true`. |
 
 ### Trigger Programmatic API
 
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -52,6 +52,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
 
   @ViewChild(TemplateRef) templateRef: TemplateRef<any>;
   @ContentChildren(MdMenuItem) items: QueryList<MdMenuItem>;
+  @Input() overlapTrigger = true;
 
   constructor(@Attribute('x-position') posX: MenuPositionX,
               @Attribute('y-position') posY: MenuPositionY) {
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -216,7 +216,12 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
   private _subscribeToPositions(position: ConnectedPositionStrategy): void {
     this._positionSubscription = position.onPositionChange.subscribe((change) => {
       const posX: MenuPositionX = change.connectionPair.originX === 'start' ? 'after' : 'before';
-      const posY: MenuPositionY = change.connectionPair.originY === 'top' ? 'below' : 'above';
+      let posY: MenuPositionY = change.connectionPair.originY === 'top' ? 'below' : 'above';
+
+      if (!this.menu.overlapTrigger) {
+        posY = posY === 'below' ? 'above' : 'below';
+      }
+
       this.menu.setPositionClasses(posX, posY);
     });
   }
@@ -230,21 +235,29 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
     const [posX, fallbackX]: HorizontalConnectionPos[] =
       this.menu.positionX === 'before' ? ['end', 'start'] : ['start', 'end'];
 
-    const [posY, fallbackY]: VerticalConnectionPos[] =
+    const [overlayY, fallbackOverlayY]: VerticalConnectionPos[] =
       this.menu.positionY === 'above' ? ['bottom', 'top'] : ['top', 'bottom'];
+
+    let originY = overlayY;
+    let fallbackOriginY = fallbackOverlayY;
+
+    if (!this.menu.overlapTrigger) {
+      originY = overlayY === 'top' ? 'bottom' : 'top';
+      fallbackOriginY = fallbackOverlayY === 'top' ? 'bottom' : 'top';
+    }
 
     return this._overlay.position()
       .connectedTo(this._element,
-          {originX: posX, originY: posY}, {overlayX: posX, overlayY: posY})
+          {originX: posX, originY: originY}, {overlayX: posX, overlayY: overlayY})
       .withFallbackPosition(
-          {originX: fallbackX, originY: posY},
-          {overlayX: fallbackX, overlayY: posY})
+          {originX: fallbackX, originY: originY},
+          {overlayX: fallbackX, overlayY: overlayY})
       .withFallbackPosition(
-          {originX: posX, originY: fallbackY},
-          {overlayX: posX, overlayY: fallbackY})
+          {originX: posX, originY: fallbackOriginY},
+          {overlayX: posX, overlayY: fallbackOverlayY})
       .withFallbackPosition(
-          {originX: fallbackX, originY: fallbackY},
-          {overlayX: fallbackX, overlayY: fallbackY});
+          {originX: fallbackX, originY: fallbackOriginY},
+          {overlayX: fallbackX, overlayY: fallbackOverlayY});
   }
 
   private _cleanUpSubscriptions(): void {
--- a/src/lib/menu/menu.spec.ts
+++ b/src/lib/menu/menu.spec.ts
@@ -1,9 +1,10 @@
-import {TestBed, async} from '@angular/core/testing';
+import {TestBed, async, ComponentFixture} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
 import {
   Component,
   ElementRef,
   EventEmitter,
+  Input,
   Output,
   TemplateRef,
   ViewChild
@@ -18,6 +19,7 @@ import {
 import {OverlayContainer} from '../core/overlay/overlay-container';
 import {ViewportRuler} from '../core/overlay/position/viewport-ruler';
 import {Dir, LayoutDirection} from '../core/rtl/dir';
+import {extendObject} from '../core/util/object-extend';
 
 describe('MdMenu', () => {
   let overlayContainerElement: HTMLElement;
@@ -27,7 +29,7 @@ describe('MdMenu', () => {
     dir = 'ltr';
     TestBed.configureTestingModule({
       imports: [MdMenuModule.forRoot()],
-      declarations: [SimpleMenu, PositionedMenu, CustomMenuPanel, CustomMenu],
+      declarations: [SimpleMenu, PositionedMenu, OverlapMenu, CustomMenuPanel, CustomMenu],
       providers: [
         {provide: OverlayContainer, useFactory: () => {
           overlayContainerElement = document.createElement('div');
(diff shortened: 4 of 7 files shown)


src/lib/menu/README.md
```diff
@@ -114,8 +114,9 @@ Output:
 
 ### Customizing menu position
 
-By default, the menu will display after and below its trigger.  You can change this display position 
-using the `x-position` (`before | after`) and `y-position` (`above | below`) attributes.  
+By default, the menu will display after and below its trigger.  You can change this display position
+using the `x-position` (`before | after`) and `y-position` (`above | below`) attributes. The menu
+can be positioned over the menu button or outside using `overlay-trigger` (`true | false`).
```
jelbourn: I'm not a fan of this name ("overlay-trigger") 

@kara thoughts?

jelbourn:
R: @kara

src/lib/menu/README.md
kara: How about "overlap-trigger"?  "Overlay-trigger" might sound like the trigger itself is an overlay.

src/lib/menu/menu-directive.ts
```diff
@@ -46,14 +46,18 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
...
   positionY: MenuPositionY = 'below';
+  overlayTrigger = true;
 
   @ViewChild(TemplateRef) templateRef: TemplateRef<any>;
   @ContentChildren(MdMenuItem) items: QueryList<MdMenuItem>;
 
   constructor(@Attribute('x-position') posX: MenuPositionX,
-              @Attribute('y-position') posY: MenuPositionY) {
+              @Attribute('y-position') posY: MenuPositionY,
+              @Attribute('overlay-trigger') trigger = true) {
```
kara: Have you considered using the `public` keyword instead?  That way you don't have to explicitly set it again on line 60.

```
@Attribute('overlap-trigger') public overlapTrigger = true
```

src/lib/menu/README.md
```diff
@@ -148,7 +149,8 @@ also adds `aria-hasPopup="true"` to the trigger element.
 | --- | --- | --- |
 | `x-position` | `before | after` | The horizontal position of the menu in relation to the trigger. Defaults to `after`. | 
 | `y-position` | `above | below` | The vertical position of the menu in relation to the trigger. Defaults to `below`. |
- 
+| `overlay-trigger`    | `true | false` | Whether to have the menu show on top of the menu trigger or outside. Defaults to `true`. |
```
kara: Can we rename to `overlapTrigger`?

src/lib/menu/menu-trigger.ts
```diff
@@ -197,12 +197,15 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
    */
   private _getPosition(): ConnectedPositionStrategy  {
     const positionX: HorizontalConnectionPos = this.menu.positionX === 'before' ? 'end' : 'start';
-    const positionY: VerticalConnectionPos = this.menu.positionY === 'above' ? 'bottom' : 'top';
-
+    const overlayY: VerticalConnectionPos = this.menu.positionY === 'above' ? 'bottom' : 'top';
+    let originY = overlayY;
+    if (this.menu.overlayTrigger) {
+      originY = overlayY === 'top' ? 'bottom' : 'top';
+    }
```
kara: If you want to overlap the trigger, shouldn't the originY and the overlayY be the same? e.g. the top of the trigger should connect to the top of the overlay or the bottom of the trigger should connect to the bottom of the overlay.  Let me know if I'm missing something.

```ts
let originY: VerticalConnectionPos;
if (this.menu.overlapTrigger) {
   originY = overlayY;
} else {
   originY = overlayY === 'top' ? 'bottom' : 'top';
}
```

src/lib/menu/menu-directive.ts
trshafer: Thanks for the suggestion.

src/lib/menu/README.md
trshafer: Done.

src/lib/menu/menu-trigger.ts
trshafer: One would think so. However I believe the current implementation produces the desired effect. Are you able to clone the pr and run the demo-app and verify? I have done so and I believe the functionality is accurate.

trshafer:
Ah, looks like there's some test bleedover. If I just focus on the [fallback positions describe block](https://github.com/angular/material2/blob/master/src/lib/menu/menu.spec.ts#L137) using fdescribe, those tests will fail. If I focus on the [whole menu test](https://github.com/angular/material2/blob/master/src/lib/menu/menu.spec.ts#L22) then the fallback positions test passes.

trshafer:
I have logged the issue: #2530.

trshafer:
Hey @kara, I'm happy to write tests for the updated change. However since the tests are currently failing I don't feel confident in updating the tests or making new ones without potentially causing issues. Are you able to update the tests? I could attempt to fix the tests, however since I don't have the background knowledge in those tests, I thought it would be better to defer to the original author. Looks like they were introduced here: https://github.com/angular/material2/commit/7572e345dca18e139440394c6c8671ecabfaf1a9

kara:
@trshafer We won't be able to merge this PR without tests. It's standard practice for this repo to provide tests with your code. Go ahead and give it a shot. I can take a look at your other issue in the meantime.

trshafer:
Totally. Happy to write tests. I should have red, green, refactored my way to this PR. Starting from a red test is harder for me to confidently ensure the test and change I have written is correct. It would be great if the menu test suite was green before I start.

I've also rebased in the recent commit.

kara:
Yep, that makes sense. I'll ping you once I've looked at the the existing tests.

trshafer:
Hey @kara, thanks for submitting the test fixes. I have written two tests for the overlap change.

I made an OverlapSubject class to reduce the testing duplication. Currently its visibility is limited to the new test describe block, however it might be helpful for the other tests as well.

src/lib/menu/menu.spec.ts
```diff
@@ -256,6 +256,66 @@ describe('MdMenu', () => {
     }
   });
 
+  describe('not overlapping', () => {
+    class OverlapSubject<T extends TestableMenu> {
```
trshafer: Done.

src/lib/menu/menu.spec.ts
```diff
@@ -256,6 +256,66 @@ describe('MdMenu', () => {
...
+        this.trigger = this.fixture.componentInstance.triggerEl.nativeElement;
+      }
+
+      openMenu() {
+        this.fixture.componentInstance.trigger.openMenu();
+        this.fixture.detectChanges();
+      }
+
+      updateTriggerStyle(style: any) {
+        return Object.assign(this.trigger.style, style);
```
trshafer: Done.

src/lib/menu/menu.spec.ts
```diff
@@ -256,6 +256,66 @@ describe('MdMenu', () => {
...
+      subject = new OverlapSubject(NonOverlapMenu);
+    });
+
+    it('positions the overlay below the trigger', () => {
+      subject.openMenu();
+
+      // Since the menu is below the trigger, the overlay top should be the trigger bottom.
+      expect(Math.round(subject.overlayRect.top))
+          .toBe(Math.round(subject.triggerRect.bottom),
+              `Expected menu to open in "above" position if "below" position wouldn't fit.`);
```
trshafer: Done.

src/lib/menu/menu.spec.ts
```diff
@@ -256,6 +256,66 @@ describe('MdMenu', () => {
...
+      // in its default "below" position below the trigger.
+      subject.updateTriggerStyle({position: 'relative', top: '650px'});
+      subject.openMenu();
+
+      // Since the menu is above the trigger, the overlay bottom should be the trigger top.
+      expect(Math.round(subject.overlayRect.bottom))
+          .toBe(Math.round(subject.triggerRect.top),
+              `Expected menu to open in "above" position if "below" position wouldn't fit.`);
+    });
+  });
```
trshafer: Done.

src/lib/menu/menu.spec.ts
```diff
@@ -311,6 +371,22 @@ class PositionedMenu {
...
+@Component({
+  template: `
+    <button [mdMenuTriggerFor]="menu" #triggerEl>Toggle menu</button>
+    <md-menu overlap-trigger="false" #menu="mdMenu">
+      <button md-menu-item> Not overlapped Content </button>
+    </md-menu>
+  `
+})
+class NonOverlapMenu implements TestableMenu {
+  @ViewChild(MdMenuTrigger) trigger: MdMenuTrigger;
```
trshafer: It's being used to open the menu: https://github.com/trshafer/material2/blob/5373c66e8cd9d0a06485158434d0afbc5233245a/src/lib/menu/menu.spec.ts#L271

src/lib/menu/menu-trigger.ts
```diff
@@ -230,21 +230,29 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
-    const [posY, fallbackY]: VerticalConnectionPos[] =
+    const [overlayY, fallbackOverlayY]: VerticalConnectionPos[] =
       this.menu.positionY === 'above' ? ['bottom', 'top'] : ['top', 'bottom'];
 
+    let originY = overlayY;
+    let fallbackOriginY = fallbackOverlayY;
+
+    if (this.menu.overlapTrigger) {
+      originY = overlayY === 'top' ? 'bottom' : 'top';
+      fallbackOriginY = fallbackOverlayY === 'top' ? 'bottom' : 'top';
```
trshafer: You are absolutely correct. The test of an explicit overlapTrigger = true is a good suggestion and exposes this in a test. Thanks for pushing back. Fixed.

src/lib/menu/README.md
```diff
@@ -115,7 +115,8 @@ Output:
 ### Customizing menu position
 
 By default, the menu will display after and below its trigger.  You can change this display position
-using the `x-position` (`before | after`) and `y-position` (`above | below`) attributes.
+using the `x-position` (`before | after`) and `y-position` (`above | below`) attributes. The menu
+can be positioned over the menu button or outside using `overlap-trigger` (`true | false`).
```
trshafer: Updated. Let me know if there is more documentation I should add.

src/lib/menu/menu-directive.ts
```diff
@@ -54,7 +54,8 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   @ContentChildren(MdMenuItem) items: QueryList<MdMenuItem>;
 
   constructor(@Attribute('x-position') posX: MenuPositionX,
-              @Attribute('y-position') posY: MenuPositionY) {
+              @Attribute('y-position') posY: MenuPositionY,
+              @Attribute('overlap-trigger') public overlapTrigger = true) {
     if (posX) { this._setPositionX(posX); }
     if (posY) { this._setPositionY(posY); }
```
trshafer: Good catch. I have updated the menu positioning classes and added a test.

trshafer:
Hello @kara, thank you for the thorough review. I'm pretty sure I have responded to all of your comments. Let me know what else I can do to help move this across the finish line.

src/lib/menu/menu.spec.ts
kara: I see. Totally missed that!

src/lib/menu/menu-directive.ts
```diff
@@ -52,6 +52,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
 
   @ViewChild(TemplateRef) templateRef: TemplateRef<any>;
   @ContentChildren(MdMenuItem) items: QueryList<MdMenuItem>;
+  @Input('overlap-trigger') overlapTrigger = true;
```
mmalerba: Don't need to rename to dash-case, we're trying to switch everything to camelCase

trshafer: so should the api look like `<md-menu [overlapTrigger]="false">`?

trshafer:
Switched `overlap-trigger` to `overlapTrigger`.

src/lib/menu/menu-directive.ts
```diff
@@ -52,6 +52,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
 
   @ViewChild(TemplateRef) templateRef: TemplateRef<any>;
   @ContentChildren(MdMenuItem) items: QueryList<MdMenuItem>;
+  @Input('overlapTrigger') overlapTrigger = true;
```
mmalerba: tiny nit: if the variable and the html attr have the same name you can omit the alias:
`@Input() overlapTrigger = true;

trshafer: Done.

PR #1772, opened 2016-11-08
chore(focus-trap): move template to a separate file

crisbeto:
Moves the focus trap to a separate file, as per a TODO.


jelbourn:
@crisbeto does `gulp build:release` work with this change? I remember there being some reason that the separate template file didn't work for anything in core. (Without I should have written in that comment...)

crisbeto:
Seems like it does - it didn't throw any errors and the template gets inlined in the `dist/@angular/material/material.umd.js`, as well as `dist/@angular/material/core/a11y/focus-trap.js`. Is there anywhere else I should be looking at?

jelbourn:
I guess it's fine, then

PR #1775, opened 2016-11-08
 feat: Support focus to checkbox, radio, button-toggle and slide-toggle

### LINKED ISSUE #1131: Support focus method on checkbox and radio inputs

Support focus method on checkbox and radio inputs

Calling focus method on MdCheckbox and MdRadioButton should change focus to these instance. Also other focusable items in browser should lose the focus.

focus method does not exists on MdCheckbox and MdRadioButton thus get error such as `focus is not a function`

http://plnkr.co/edit/9RJ5DOuPMK8b3zPeR95J?p=preview

Similar to MdInput there should be way to focus on elements. this can be used for focusing to element with validation error

Was able to reproduce on both firefox and chrome.

tinayuangao:
Closes #1131 

R: @jelbourn


src/lib/radio/radio.spec.ts
```diff
@@ -188,6 +188,23 @@ describe('MdRadio', () => {
       expect(radioNativeElements[0].classList).not.toContain('md-radio-focused');
     });
 
+    it('should focus individual radio buttons', () => {
+      let nativeRadioInput = <HTMLElement> radioNativeElements[0].querySelector('input');
+      radioInstances[0].focus();
+
+      expect(nativeRadioInput.classList).not.toContain('md-radio-focused');
+
+      dispatchEvent('focus', nativeRadioInput);
```
jelbourn: Why are you dispatching a focus event here instead of calling `focus`?

tinayuangao: Fixed the test. PTAL. Thanks!

jelbourn:
LGTM

jelbourn:
@tinayuangao can you rebase this PR?

PR #1778, opened 2016-11-09
fix: disable ripples when parent component is disabled

jelbourn:
R: @tinayuangao or @kara 

I noticed that disabled checkboxes and radio buttons still had active ripples on click. This fix needs to go in before we can do a release.

### DIFF
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -218,6 +218,17 @@ describe('MdRadio', () => {
       expect(radioInstances.every(radio => !radio.checked)).toBe(true);
     });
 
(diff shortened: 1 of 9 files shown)


src/lib/radio/radio.spec.ts
```diff
@@ -218,6 +218,17 @@ describe('MdRadio', () => {
       expect(radioInstances.every(radio => !radio.checked)).toBe(true);
     });
 
+    it('should not have a ripple on disabled radio buttons', () => {
+      let rippleElement = radioNativeElements[0].querySelector('[md-ripple');
```
kara: It looks like you're missing the end square bracket in the selector? Same on line 228.

PR #1783, opened 2016-11-09
feat(snackbar): don't require a ViewContainerRef

jelbourn:
R: @josephperrott or @kara 

Also lets you use an object literal as the config.

### DIFF
--- a/src/lib/snack-bar/snack-bar.ts
+++ b/src/lib/snack-bar/snack-bar.ts
@@ -40,7 +40,7 @@ export class MdSnackBar {
    * currently opened snack bars.
    */
   openFromComponent<T>(component: ComponentType<T>, config?: MdSnackBarConfig): MdSnackBarRef<T> {
-    config = this._applyConfigDefaults(config);
+    config = _applyConfigDefaults(config);
     let overlayRef = this._createOverlay();
     let snackBarContainer = this._attachSnackBarContainer(overlayRef, config);
     let snackBarRef = this._attachSnackbarContent(component, snackBarContainer, overlayRef);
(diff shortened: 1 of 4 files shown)


src/lib/snack-bar/snack-bar.ts
```diff
@@ -39,43 +39,47 @@ export class MdSnackBar {
...
-   * Creates and dispatches a snack bar.
+   * Opens a snackbar with a message and an optional action.
+   * @param message The message to show in the snackbar.
+   * @param action The label for the snackbar action.
+   * @param config Additional configuration options for the snackbar.
+   * @returns {MdSnackBarRef<SimpleSnackBar>}
    */
-  open(message: string, actionLabel: string,
-       config: MdSnackBarConfig): MdSnackBarRef<SimpleSnackBar> {
+  open(message: string, action = '', config?: MdSnackBarConfig): MdSnackBarRef<SimpleSnackBar> {
```
josephperrott: Is there a reason to not just set a default config of {}?

open(message: string, action = '', config: MdSnackBarConfig = {})...

src/lib/snack-bar/snack-bar.ts
```diff
@@ -113,6 +117,15 @@ export class MdSnackBar {
         .bottom('0');
     return this._overlay.create(state);
   }
+
+  /**
+   * Applies default options to the snackbar config.
+   * @param config The configuration to which the defaults will be applied.
+   * @returns The new configuration object with defaults applied.
+   */
+  private _applyConfigDefaults(config: MdSnackBarConfig): MdSnackBarConfig {
```
josephperrott: should this be a module level method, or static even?

src/lib/snack-bar/snack-bar.ts
jelbourn: Done

PR #1784, opened 2016-11-09
fix(overlay): prevent blurry connected overlays

jelbourn:
R: @kara


jelbourn:
@kara Fixed tests. I also had to update some of the newer menu unit tests.

PR #1785, opened 2016-11-09
fix(overlay-ref): check if parent node still exists before removing c…

janisr:
…hild

Getting the following exception in some scenarios, eg. multiple dialogs of the same type are open and closing them fast

 TypeError: Cannot read property 'removeChild' of null
    at HTMLDivElement.<anonymous> (overlay-ref.js:70)
    at ZoneDelegate.invokeTask (zone.js:265)
    at Object.onInvokeTask (ng_zone.js:227)
    at ZoneDelegate.invokeTask (zone.js:264)
    at Zone.runTask (zone.js:154)
    at HTMLDivElement.ZoneTask.invoke (zone.js:335)


jelbourn:
This will be fixed in #1716

PR #1788, opened 2016-11-09
feat(tabs): add animation when switching tabs, include optional dynamic height

andrewseguin:


src/lib/tabs/_tabs-common.scss
```diff
@@ -36,5 +36,5 @@ $md-tab-bar-height: 48px !default;
   position: absolute;
   bottom: 0;
   height: 2px;
-  transition: 350ms ease-out;
+  transition: 500ms ease-out;
 }
```
hansl: End of file, end of line.

src/lib/tabs/tab-body.html
```diff
@@ -0,0 +1,6 @@
+<div class="md-tab-body-content"
+     [@position]="_position"
+     (@position.start)="_onAnimationStarted($event)"
+     (@position.done)="_onAnimationComplete($event)">
+  <template portalHost></template>
+</div>
```
hansl: End of line.

src/lib/tabs/tab-group.spec.ts
```diff
@@ -136,6 +139,53 @@ describe('MdTabGroup', () => {
       expect(component.handleSelection).toHaveBeenCalledTimes(1);
       expect(component.selectEvent.index).toBe(2);
     }));
+
+    it('should update tab positions and attach content when selected', fakeAsync(() => {
+      fixture.detectChanges();
+      let tabComponent = fixture.debugElement.query(By.css('md-tab-group')).componentInstance;
```
kara: Out of curiosity, why is this one `let` while the other is `const`? Seems like they could both be `const`.

src/lib/tabs/tab-body.html
```diff
@@ -0,0 +1,6 @@
+<div class="md-tab-body-content"
+     [@position]="_position"
```
kara: Nit: This animation name doesn't really describe what is happening with the animation.

src/lib/tabs/tabs.ts
```diff
@@ -30,6 +37,7 @@ import {MdTabNavBar, MdTabLink} from './tab-nav-bar/tab-nav-bar';
 import {MdInkBar} from './ink-bar';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/operator/map';
+import {PortalHostDirective} from '../core/portal/portal-directives';
```
kara: It might make more sense to add this to the existing import list on line 19 from `../core`. I think it's exported there too.

src/lib/tabs/tabs.ts
```diff
@@ -248,15 +249,123 @@ export class MdTabGroup {
...
 
+export type MdTabBodyActiveState = 'left' | 'center' | 'right';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-tab-body',
+  templateUrl: 'tab-body.html',
+  animations: [
+    trigger('position', [
+      state('left', style({transform: 'translateX(-100%)'})),
```
hansl: According to https://jsperf.com/translate3d-vs-xy/28, `translate3d` is the fastest in most browsers (because 3d rendering), which is why we use it in most places.

src/lib/tabs/tabs.ts
```diff
@@ -248,15 +249,123 @@ export class MdTabGroup {
...
+  /** Returns a unique id for each tab content element */
+  _getTabContentId(i: number): string {
+    return `md-tab-content-${this._groupId}-${i}`;
+  }
+
+  /**
+   * Sets the height of the body wrapper to the height of the activating tab if dynamic
+   * height property is true.
+   */
+  _setTabBodyWrapperHeight(e: number) {
```
kara: Nit: return type?

kara: Maybe `height` would be more descriptive than `e`? It's not clear at first glance what it stands for.

src/lib/tabs/tab-group.scss
```diff
@@ -32,19 +32,28 @@ md-ink-bar {
...
-  display: none;
-  overflow: auto;
-  box-sizing: border-box;
-  flex-grow: 1;
-  flex-shrink: 1;
-  &.md-tab-active {
-    display: block;
+md-tab-body {
+  display: block;
+  position: absolute;
```
andrewseguin: I don't see that we have one already. Want me to make one in this change?

src/lib/tabs/tabs.ts
andrewseguin: Seems that Chrome has them set at about equal operations but 3D is way faster in Safari (and probably others). Thanks for the heads up

src/lib/tabs/tab-group.scss
kara: We do have md-fullscreen, which is I think what Hans means. But it is position: fixed, not absolute.

src/lib/tabs/tab-group.spec.ts
andrewseguin: Thanks, I tend to just default to `let` rather than `const`. Gotta break that habit.

src/lib/tabs/tabs.ts
```diff
@@ -248,15 +249,123 @@ export class MdTabGroup {
...
+  animations: [
+    trigger('position', [
+      state('left', style({transform: 'translateX(-100%)'})),
+      state('center', style({transform: 'translateX(0%)'})),
+      state('right', style({transform: 'translateX(100%)'})),
+      transition('* => *', animate('0.5s cubic-bezier(0.35, 0, 0.25, 1)')),
+    ])
+  ]
+})
+export class MdTabBody implements OnInit {
```
andrewseguin: Primarily code organization - portal attach/detach + animations would have added quite a bit to tab group and I don't think it would have been as readable about what is happening between all the interactions

kara: Okay, that's fair.  Thought that might be the case.

andrewseguin:
Changes made in response to the comments. Thanks for reviewing

src/lib/tabs/tab-group.scss
hansl: Right, there should probably be `md-fill` mixin or something like that with `position: absolute`. That mixin does not work right now.

andrewseguin:
Switched md-fullscreen back to using absolute positioning (changed in #893) which is safe for its use in sidenav. Renamed as md-fill and used in the tab body.

andrewseguin:
Sorry, added one more small change to support RTL animation direction

hansl:
LGTM.

e2e/components/tabs/tabs.e2e.ts
```diff
@@ -77,8 +77,17 @@ function getFocusStates(elements: ElementArrayFinder) {
...
-function getActiveStates(elements: ElementArrayFinder) {
-  return getClassStates(elements, 'md-tab-active');
+function getLabelActiveStates(elements: ElementArrayFinder) {
+  return getClassStates(elements, 'md-tab-label-active');
+}
+
+/**
+ * Returns an array of true/false that represents the active states for the provided elements
+ * @param elements
+ * @returns {webdriver.promise.Promise<Promise<boolean>[]>|webdriver.promise.Promise<T[]>}
```
jelbourn: Since this is TypeScript, you don't need the type in the JsDoc (these were formerly js)

src/lib/tabs/tabs.ts
```diff
@@ -248,15 +253,132 @@ export class MdTabGroup {
...
+
+  /** The shifted index position of the tab body, where zero represents the active center tab. */
+  _position: MdTabBodyActiveState;
+  @Input('md-tab-body-position') set position(v: number) {
+    if (v < 0) {
+      this._position = this.getLayoutDirection() == 'ltr' ? 'left' : 'right';
+    } else if (v > 0) {
+      this._position = this.getLayoutDirection() == 'ltr' ? 'right' : 'left';
+    }
+    if (v == 0) { this._position = 'center'; }
```
jelbourn: Just `else` here?

src/lib/tabs/tab-group.spec.ts
```diff
@@ -136,6 +139,53 @@ describe('MdTabGroup', () => {
       expect(component.handleSelection).toHaveBeenCalledTimes(1);
       expect(component.selectEvent.index).toBe(2);
     }));
+
+    it('should update tab positions and attach content when selected', fakeAsync(() => {
```
jelbourn: Also a test for RTL?

andrewseguin:
Will make a note to check out screen reader support after this PR

PR #1791, opened 2016-11-09
chore: don't use Object.assign

jelbourn:
R: @kara 

`Object.assign` isn't available in Google's TS builds.


epelc:
@jelbourn What do you think about using polyfills for things like this instead? It would make transitioning over easier if/when google ever gets `Object.assign`.

jelbourn:
@epelc In this case it was easier to add the function that try to tell Closure Compiler about `Object.assign` existing

PR #1792, opened 2016-11-09
fix(button): improved tap responsiveness on mobile

### LINKED ISSUE #1316: Buttons need to be tapped twice in order to fire the click event (iOS)

Bug

A tap on the button should fire the click event.

The first tap highlights the button. The second tap fires the click event.

Click on the sidenav icon on an iPhone or iPad: http://material2-test.firebaseapp.com/
- Angular 2.0.0
- Material Alpha 8-2
- iOS 10.0.1
- Safari

crisbeto:
Improves the tap responsiveness of buttons on mobile by replacing the `::after` overlay with a DOM node that prevents the default touch action.
Previously, the `::after` overlay would capture the first tap, causing the button to have to tapped twice in order to fire it's click handler.

Fixes #1316.

### DIFF
--- a/src/lib/button/button.html
+++ b/src/lib/button/button.html
@@ -4,4 +4,5 @@
     [md-ripple-trigger]="getHostElement()"
     [md-ripple-color]="isRoundButton() ? 'rgba(255, 255, 255, 0.2)' : ''"
     md-ripple-background-color="rgba(0, 0, 0, 0)"></div>
-<div class="md-button-overlay" (touchstart)="$event.preventDefault()"></div>
+<!-- the touchstart handler prevents the overlay from capturing the initial tap on touch devices -->
+<div class="md-button-focus-overlay" (touchstart)="$event.preventDefault()"></div>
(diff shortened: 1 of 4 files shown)


src/lib/button/button.html
```diff
@@ -4,3 +4,4 @@
     [md-ripple-trigger]="getHostElement()"
     [md-ripple-color]="isRoundButton() ? 'rgba(255, 255, 255, 0.2)' : ''"
     md-ripple-background-color="rgba(0, 0, 0, 0)"></div>
+<div class="md-button-overlay" (touchstart)="$event.preventDefault()"></div>
```
jelbourn: Can we rename this to `md-button-focus-overlay` and add a comment here explaining what this is doing with `touchstart`?

crisbeto:
Renamed the overlay class and added a comment about the `touchstart` handler.

jelbourn:
LGTM

sunshineo:
Which version will this fix be in?

qdouble:
@jelbourn can you put this in a 10-1 release soon? Currently, you'd have to monkey-patch the old workaround on every component to get it to work properly on IOS.

sunshineo:
Is this really out?

crisbeto:
It should be in `alpha.11`.

PR #1794, opened 2016-11-09
fix(slider): support for RTL and invert

mmalerba:
blocked on #1759


src/lib/slider/slider.scss
```diff
@@ -66,6 +75,11 @@ md-slider {
   overflow: hidden;
 }
 
+[dir='rtl'] .md-slider-ticks-container {
+  // translateZ(0) prevents chrome bug where overflow: hidden; doesn't work.
```
jelbourn: Do you have a link to the Chrome bug?

mmalerba: I don't, I would file one except its really elusive. If a try replicating this HTML/CSS in a jsfiddle it doesn't happen.

PR #1796, opened 2016-11-10
feat(a11y): add wrap mode to key manager

### LINKED ISSUE #1608: Wrap property for ListKeyManager
It would be nice if the ListKeyManager had a second mode where the focus doesn't wrap, but uses the alternate approach with HOME and END. Creating an issue to track, but still needs discussion before implementing.

Necessary for select implementation (#118).

kara:
This PR adds support for two distinct modes: 

- Wrap mode: focus will wrap if it reaches past either end of the list
- Nowrap mode (default): focus will stay where it is if it reaches either end of the list

As part of the default mode, it also supports the HOME and END keys jumping to the beginning and the end of the list, respectively.

To use wrap mode:

```ts

this.keyManager = new ListKeyManager(this.items).withFocusWrap();

```

Closes #1608.

### DIFF
--- a/src/lib/core/a11y/list-key-manager.spec.ts
+++ b/src/lib/core/a11y/list-key-manager.spec.ts
@@ -1,125 +1,293 @@
 import {QueryList} from '@angular/core';
-import {ListKeyManager, MdFocusable} from './list-key-manager';
-import {DOWN_ARROW, UP_ARROW, TAB} from '../keyboard/keycodes';
+import {ListKeyManager} from './list-key-manager';
+import {DOWN_ARROW, UP_ARROW, TAB, HOME, END} from '../keyboard/keycodes';
 
 class FakeFocusable {
   disabled = false;
   focus() {}
 }
 
+class FakeQueryList<T> extends QueryList<T> {
+  get length() { return this.items.length; }
+  items: T[];
+  toArray() {
+    return this.items;
+  }
+}
+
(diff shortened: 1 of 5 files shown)


src/lib/core/a11y/list-key-manager.ts
```diff
@@ -19,65 +19,130 @@ export interface MdFocusable {
 export class ListKeyManager {
   private _focusedItemIndex: number;
   private _tabOut: Subject<any> = new Subject();
+  private _wrapMode: boolean = false;
```
jelbourn: Just `_wrap`?

src/lib/core/a11y/list-key-manager.spec.ts
```diff
@@ -1,125 +1,293 @@
...
+      expect(itemList.items[0].focus).not.toHaveBeenCalled();
+      expect(itemList.items[1].focus).toHaveBeenCalledTimes(1);
+      expect(itemList.items[2].focus).toHaveBeenCalledTimes(1);
+    });
+
+    it('should not move focus past either end of the list', () => {
+      keyManager.onKeydown(DOWN_ARROW_EVENT);
+      keyManager.onKeydown(DOWN_ARROW_EVENT);
+      expect(keyManager.focusedItemIndex)
+        .toBe(2, `Expected focus to be on the last item of the list.`);
```
jelbourn: Overflow should be +4 indent

kara:
@jelbourn Comments addressed!

PR #1800, opened 2016-11-10
docs(icon): Use `mdIconRegistry` instead of `MdIconRegistry`

epelc:
`mdIconRegistry` does not exist. The real name is `MdIconRegistry`


jelbourn:
These are examples where it's using a local var that's been assigned to the instance of `MdIconRegistry`, which is a constructor.

epelc:
@jelbourn sorry

PR #1814, opened 2016-11-10
docs(dialog): add new option and update example

crisbeto:
* Mentions the `disableClose` option in the readme.
* Updates the dialog readme and dialog demo not to use the `viewContainerRef` since it's not necessary anymore.

### DIFF
--- a/src/lib/dialog/README.md
+++ b/src/lib/dialog/README.md
@@ -1,6 +1,6 @@
 # MdDialog
 
-MdDialog is a service, which opens dialogs components in the view. 
+MdDialog is a service, which opens dialogs components in the view.
 
 ### Methods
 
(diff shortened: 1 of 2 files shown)


src/lib/dialog/README.md
```diff
@@ -12,8 +12,8 @@ MdDialog is a service, which opens dialogs components in the view.
 
 | Key |  Description |
 | --- | --- |
-| `viewContainerRef: ViewContainerRef` | The view container ref to attach the dialog to. |
```
jelbourn: We shouldn't remove this because it is still an option, but it is optional.

src/lib/dialog/README.md
```diff
@@ -40,15 +40,12 @@ export class PizzaComponent {
...
-    public viewContainerRef: ViewContainerRef) { }
+  constructor(public dialog: MdDialog) { }
 
   openDialog() {
-    let config = new MdDialogConfig();
-    config.viewContainerRef = this.viewContainerRef;
-
-    this.dialogRef = this.dialog.open(PizzaDialog, config);
+    this.dialogRef = this.dialog.open(PizzaDialog, {
+      role: 'alertdialog'
```
jelbourn: This shouldn't be `alertdialog` since that's not appropriate to the content being opened.

crisbeto: What I was going for was to show that you can pass in an object, but I didn't want people copying the config over with something like `disableClose: true`. Having `disableClose: false` seemed redundant.

crisbeto:
Re-added the `viewContainerRef` and switched up the config options in the example.

jelbourn:
LGTM

vasyliv:
some bug when dbclick around dialog
image
http://plnkr.co/edit/OrXLHggyhZu7Wijf7czf?p=preview

crisbeto:
There should be a fix for this one in master @vasyliv. I could reproduce this in the Plunkr, but not locally.

Issue #1808, opened 2016-11-10
[Tooltip] allow configuration of the hide delay


rolandjitsu:
#### Bug, feature request, or proposal:
Feature

#### What is the expected behavior?
`TOOLTIP_HIDE_DELAY` should be configurable even though according to the [specs](https://material.google.com/components/tooltips.html#tooltips-interaction) is set to be `1.5` seconds.
Ideally, a user should configure it via an opaque token:
```
import {TOOLTIP_HIDE_DELAY} from '@angular/material';

@NgModule({
    providers: [
        {provide: TOOLTIP_HIDE_DELAY, useValue: 450}
    ]
})
export class MyModule {}
```
#### What is the current behavior?
Currently you cannot configure the timing.

#### What are the steps to reproduce?

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: http://plnkr.co/edit/o077B6uEiiIgkC0S06dd

#### What is the use-case or motivation for changing an existing behavior?
Some users may wish to hide the tooltip earlier than the spec dictates.

#### Which versions of Angular, Material, OS, browsers are affected?
`2.0.0-alpha.10`

t0f3rchr15:
Is a configurable delay on open() on the roadmap as well?

aciccarello:
@t0f3rchr15 See #3286

PR #1819, opened 2016-11-11
feat(tooltip): show tooltip on longpress; remove delay on mouseleave

andrewseguin:
Support longpress for mobile users to see tooltips. Adheres better to the md spec which says that "On lift" there should be a delay, which was incorrectly applied to mouseover.

Partially addresses #1808 by hiding the tooltip immediately when the mouse is hidden.


KostyaTretyak:
Also probably need to change the positioning element. Now, when the tooltip is displayed and when the scroll page, tooltip remains fixed.

fxck:
@KostyaTretyak it's an overlay related problem, very annoying one though, happens to all components using overlay, like menu.

PR #1820, opened 2016-11-11
fix(ripple): Fix the ripple position

### LINKED ISSUE #1817: Ripple positioning is off beneath the fold
To repro:

1)  Slow animations down in animations console (because we are not robots).
2) Remove "fullscreen" attribute from demo app sidenav. 
3)  Move button (or other ripple container) to the bottom of the page, so that you have to scroll to get to it.
4)  Click the button.

You can see that the ripple originates from the bottom and not the center of the click event.  It also takes a bit longer to appear.

Edit: it looks like the height of the ripple foreground wasn't taking into account the scrollTop of the page. So the height would be off by as far as the user had scrolled, making the ripple huge and centered far below (which is why we were only seeing the top later).  Should be fixable by subtracting the scroll offset from the pageX and pageY.

tinayuangao:
When scroll down the page, the ripple position is off

Closes #1817

R: @kara

### DIFF
--- a/src/lib/core/portal/dom-portal-host.ts
+++ b/src/lib/core/portal/dom-portal-host.ts
@@ -43,27 +43,36 @@ export class DomPortalHost extends BasePortalHost {
     } else {
       componentRef = componentFactory.create(portal.injector || this._defaultInjector);
 
-      // When creating a component outside of a ViewContainer, we need to manually register
-      // its ChangeDetector with the application. This API is unfortunately not yet published
-      // in Angular core. The change detector must also be deregistered when the component
-      // is destroyed to prevent memory leaks.
-      //
-      // See https://github.com/angular/angular/pull/12674
-      let changeDetectorRef = componentRef.changeDetectorRef;
-      (this._appRef as any).registerChangeDetector(changeDetectorRef);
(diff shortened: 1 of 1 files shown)


src/lib/core/ripple/ripple.ts
```diff
@@ -154,7 +154,7 @@ export class MdRipple implements OnInit, OnDestroy, OnChanges {
       // FIXME: This fails on IE11, which still sets pageX/Y and screenX/Y on keyboard clicks.
       const isKeyEvent =
           (event.screenX === 0 && event.screenY === 0 && event.pageX === 0 && event.pageY === 0);
-      this.end(event.pageX, event.pageY, isKeyEvent);
+      this.end(event.pageX, event.pageY - document.body.scrollTop, isKeyEvent);
```
jelbourn: That page could be horizontally scrolled as well, so you should account for `scrollLeft` on the x-axis.

Does this still work if another element is scrolled other than `body`? Is there a unit test we can add for this? (connected position strategy has a lot of tests that deal with this type of thing).

tinayuangao: Fixed firefox & scrollLeft. Now it's working with elements other than body. 

Working on unit test for this...

PR #1821, opened 2016-11-11
add gulp task to convert markdown to html

mmalerba:
additionally replaces comments in the outputted html of the form `<!-- example(label) -->` with `<div example="label">`

### DIFF
--- a/tools/gulp/tasks/docs.ts
+++ b/tools/gulp/tasks/docs.ts
@@ -3,6 +3,10 @@ const markdown = require('gulp-markdown');
 const transform = require('gulp-transform');
 
 
+// Our docs contain comments of the form `<!-- example(...) -->` which serve as placeholders where
+// example code should be inserted. We replace these comments with divs that have a
+// `material-docs-example` attribute which can be used to locate the divs and initialize the example
+// viewer.
 const EXAMPLE_PATTERN = /<!--\W*example\(([^)]+)\)\W*-->/g;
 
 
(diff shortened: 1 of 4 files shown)


tools/gulp/tasks/docs.ts
```diff
@@ -0,0 +1,16 @@
+import gulp = require('gulp');
+const markdown = require('gulp-markdown');
+const transform = require('gulp-transform');
+
+
+const EXAMPLE_PATTERN = /<!--\W*example\(([^)]+)\)\W*-->/g;
```
jelbourn: Can you add a comment explaining what we're replacing and why we're replacing it?

tools/gulp/tasks/docs.ts
```diff
@@ -0,0 +1,16 @@
...
+
+
+gulp.task('docs', () => {
+  return gulp.src(['src/lib/**/*.md'])
+      .pipe(markdown())
+      .pipe(transform((content: string) =>
+          content.toString().replace(EXAMPLE_PATTERN, (match: string, name: string) =>
+              `<div example="${name}"></div>`)))
```
jelbourn: How about `material-docs-example` to be super explicit?

mmalerba:
Done

tools/gulp/tasks/docs.ts
```diff
@@ -0,0 +1,20 @@
+import gulp = require('gulp');
+const markdown = require('gulp-markdown');
+const transform = require('gulp-transform');
+
+
+// Our docs contain comments of the form `<!-- example(...) -->` which serve as placeholders where
+// example code should be inserted. We replace these comments with divs that have a
+// `material-docs-example` attribute which can be used to locate the divs and initialize the example
+// viewer.
```
tinayuangao: very nit: do you want to comment on naming convention for $name?

mmalerba: At some point yes, I'm not really sure what's going to go there yet...

jelbourn:
LGTM

Issue #1616, opened 2016-10-26
Add onAction observable for simple snackbars


jelbourn:
Adds a way to be notified when the action is invoked.

rolandjitsu:
Is this still being worked on or what's the status?

rolandjitsu:
Made a PR, [1729](https://github.com/angular/material2/pull/1729), but the tests on saucelabs are timing out.

rolandjitsu:
Closed my PR, messed up some commits. Is anyone working on this at this point or can I give it another try?

jelbourn:
@josephperrott?

michaeljota:
If the PR got merge, this should this still open? I think this can be safely closed.

PR #1826, opened 2016-11-11
feat(snackbar): add onAction to snackbar ref

josephperrott:
For #1616


src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -38,8 +43,21 @@ export class MdSnackBarRef<T> {
     }
   }
 
+  /** Marks the snackbar action clicked. */
+  action(): void {
```
jelbourn: Can we mark this as internal (just an underscore prefix without `private` or `@internal`)? I don't think users should be calling it on their own.

src/lib/snack-bar/snack-bar.spec.ts
```diff
@@ -259,6 +259,32 @@ describe('MdSnackBar', () => {
...
+       viewContainerFixture.detectChanges();
+       
+       snackBarRef.afterDismissed().subscribe(null, null, () => {
+         dismissObservableCompleted = true;
+       });
+       snackBarRef.onAction().subscribe(null, null, () => {
+         actionObservableCompleted = true;
+      });
+
+      snackBarRef.action();
```
jelbourn: Instead of calling this directly, could you grab the action button element and call `.click()` on it?

jelbourn:
LGTM

PR #1829, opened 2016-11-11
feat(menu): Support fallback positioning on menu trigger.

trshafer:
This ensures that the menu will do the best attempt at showing instead of rendering off screen.


jelbourn:
I believe this was already addressed in #1761. @kara can you confirm?

(FYI, be sure to let us know if you plan on doing work to something in-progress like menu or select so we can avoid conflicts)

trshafer:
The other PR looks good. Thanks for merging it in.

PR #1832, opened 2016-11-11
feat(overlay): emit position change event

kara:
r: @jelbourn


src/lib/core/overlay/position/connected-position-strategy.ts
```diff
@@ -36,6 +38,12 @@ export class ConnectedPositionStrategy implements PositionStrategy {
   /** The origin element against which the overlay will be positioned. */
   private _origin: HTMLElement;
 
+  private _onPositionChange: Subject<ConnectedOverlayPositionChange> = new Subject();
+
+  /** Emits an event when a fallback position is used. */
```
jelbourn: It would be more accurate to say that it emits whenever the connection point changes (based on what will fit on the screen).

PR #1834, opened 2016-11-11
chore: fix unit broken unit tests and CI not running them

jelbourn:


jelbourn:
R: @kara 

CC: @crisbeto merging #1741 caused Travis to no longer run any unit tests. Because of that, we didn't notice that #1716 broke a bunch of unit tests. 

This PR fixes the broken tests from #1716 and reverts #1741.

PR #1838, opened 2016-11-12
fix(md-spinner): animation not being cleaned up when used with AoT

### LINKED ISSUE #1283: bug(md-spinner): possible memory leak with AoT
When using `md-spinner` with `ngc` and `rollup` its `OnDestroy` method won't run.

[Simple plunker example](https://plnkr.co/edit/fGEfGA10MGf0wOCM3nLV?p=info).

The only modification made to `md-progress-circle` component was to add logging when `OnDestroy` is called. Obviously it _should_ log `destroyed` twice, but it happens only once and the `interdeterminateInterval` keeps running in the background causing memory leak and more cpu usage.

Note that `<md-progress-circle mode="indeterminate">` **works** and its `OnDestroy` method gets called, so as a workaround it's there. Also, `md-spinner` itself seems to work as well when not precompiled. I haven't really investigated further so I'm not sure whether it's ngc or rollup causing the issue.

crisbeto:
Fixes the `md-spinner` animation interval not being cleaned up when the app has been compiled through the AoT.
This is due to the the fact that the `ngOnDestroy` handler is in the base `MdProgressCircle` class.

Fixes #1283.

**Note:** At some point it may be better to reorganize the component so we have the following setup:
* Base progress circle class - handles all of the rendering and animation logic.
* Progress circle class - a subclass that only handles the determinate mode.
* Progress spinner - a subclass that only handles the spinning animation.

This should avoid having to refer to the `super` class for things like cleaning up the animation.

### DIFF
--- a/src/lib/progress-circle/progress-circle.ts
+++ b/src/lib/progress-circle/progress-circle.ts
@@ -244,10 +244,16 @@ export class MdProgressCircle implements OnDestroy {
   templateUrl: 'progress-circle.html',
   styleUrls: ['progress-circle.css'],
 })
-export class MdSpinner extends MdProgressCircle {
+export class MdSpinner extends MdProgressCircle implements OnDestroy {
   constructor(changeDetectorRef: ChangeDetectorRef, elementRef: ElementRef, ngZone: NgZone) {
     super(changeDetectorRef, ngZone, elementRef);
     this.mode = 'indeterminate';
+  }
+
+  ngOnDestroy() {
+    // The `ngOnDestroy` from `MdProgressCircle` should be called explicitly, because
+    // in certain cases Angular won't call it (e.g. when using AoT and in unit tests).
+    super.ngOnDestroy();
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/progress-circle/progress-circle.ts
```diff
@@ -244,11 +244,17 @@ export class MdProgressCircle implements OnDestroy {
...
   constructor(changeDetectorRef: ChangeDetectorRef, elementRef: ElementRef, ngZone: NgZone) {
     super(changeDetectorRef, ngZone, elementRef);
     this.mode = 'indeterminate';
   }
+
+  ngOnDestroy() {
+    // The `ngOnDestroy` from `MdProgressCircle` should be called explicitly, because
+    // in certain cases Angular won't call it (e.g. when using AoT and in unit tests).
+    super.ngOnDestroy();
+  }
```
jelbourn: @tbosch Is this the intended behavior for generated code?

crisbeto: Just to note that this happens in the unit tests as well. I'm pretty sure that they're not using AoT.

tbosch: Sadly yes, see https://github.com/angular/angular/issues/11606

jelbourn:
LGTM

PR #1839, opened 2016-11-12
chore(e2e): fix e2e tests for latest chrome versions

DevVersion:
* Temporary fixed a test, which fails in Chrome 53+ (due to [cr-bug/1502](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1502))
* Updates the Firefox and Chrome versions for Saucelabs.
* Another e2e test in the menu has been fixed with SHA accab2049626fe267a94b74d45c3dd7d6b5df865

**Note**: This is pretty important, because pretty much everyone who works on E2E tests uses one of latest Chrome versions-and the tests don't pass locally (makes working on e2e difficult)

### DIFF
--- a/e2e/components/menu/menu.e2e.ts
+++ b/e2e/components/menu/menu.e2e.ts
@@ -75,7 +75,10 @@ describe('menu', () => {
     });
 
     it('should auto-focus the first item when opened with SPACE', () => {
-      page.pressKey(protractor.Key.SPACE);
+      // Due to chromebug https://bugs.chromium.org/p/chromedriver/issues/detail?id=1502,
+      // the space key is not properly working in the latest Chromedriver.
+      // TODO: Revert temporary change to use space key after Chromedriver 2.26 is available.
+      page.pressKey(protractor.Key.ENTER);
       page.expectFocusOn(page.items(0));
     });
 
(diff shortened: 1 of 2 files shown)


e2e/components/menu/menu.e2e.ts
```diff
@@ -75,7 +75,10 @@ describe('menu', () => {
     });
 
     it('should auto-focus the first item when opened with SPACE', () => {
-      page.pressKey(protractor.Key.SPACE);
+      // Due to chromebug https://bugs.chromium.org/p/chromedriver/issues/detail?id=1502,
+      // the space key is not properly working in the latest Chromedriver.
+      // TODO: Revert temporary change to use space key after Chromedriver 2.26 is available.
+      page.pressKey(protractor.Key.ENTER);
```
jelbourn: @kara also ran into this when writing the test. I'm hesitant to change it, because we probably won't remember to put it back. @kara thoughts?

kara: Given that we already have a test for the enter key, I'm not sure  this test is useful once the space is removed. Also agree with Jeremy that we'll probably forget to change it back. Let's just create an issue to track.

jelbourn:
@DevVersion we're just going to wait this one out for ChomeDriver 2.26.

PR #1846, opened 2016-11-13
feat(textarea): add md-autosize directive

jelbourn:
R: @mmalerba 

CC: @fxck

### DIFF
--- a/src/lib/input/autosize.ts
+++ b/src/lib/input/autosize.ts
@@ -1,0 +1,81 @@
+import {Directive, ElementRef, Input, OnInit} from '@angular/core';
+
+
+/**
+ * Directive to automatically resize a textarea to fit its content.
+ */
+@Directive({
+  selector: 'textarea[md-autosize]',
+  host: {
+    '(input)': 'resizeToFitContent()',
+    '[style.min-height]': '_minHeight',
+    '[style.max-height]': '_maxHeight',
+  },
+})
+export class MdTextareaAutosize implements OnInit {
+  /** Minimum number of rows for this textarea. */
+  @Input() minRows: number;
+
+  /** Maximum number of rows for this textarea. */
+  @Input() maxRows: number;
+
+  /** Cached height of a textarea with a single row. */
+  private _cachedLineHeight: number;
+
+  constructor(private _elementRef: ElementRef) { }
+
+  /** The minimum height of the textarea as determined by minRows. */
+  get _minHeight() {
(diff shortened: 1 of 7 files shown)


fxck:
Yeah, LGTM.

src/lib/input/autosize.ts
```diff
@@ -0,0 +1,79 @@
...
+    '[style.min-height]': '_minHeight',
+    '[style.max-height]': '_maxHeight',
+  },
+})
+export class MdTextareaAutosize implements OnInit {
+  /** Minimum number of rows for this textarea. */
+  @Input() minRows: number;
+
+  /** Maximum number of rows for this textarea. */
+  @Input() maxRows: number;
```
mmalerba: could you also do an @Input() for value so you can resize when the content is changed programmatically?

src/lib/input/autosize.ts
```diff
@@ -0,0 +1,79 @@
...
+    textareaClone.rows = 1;
+
+    // Use `position: absolute` so that this doesn't cause a browser layout and use
+    // `visibility: hidden` so that nothing is rendered. Clear any other styles that
+    // would affect the height.
+    textareaClone.style.position = 'absolute';
+    textareaClone.style.visibility = 'hidden';
+    textareaClone.style.padding = '';
+    textareaClone.style.minHeight = '';
+    textareaClone.style.height = '';
```
mmalerba: also clear maxHeight and border

src/lib/input/autosize.ts
```diff
@@ -19,10 +19,24 @@ export class MdTextareaAutosize implements OnInit {
   /** Maximum number of rows for this textarea. */
   @Input() maxRows: number;
 
+  /** The value of the textarea. */
+  private _value: string;
```
mmalerba: Is this our team's official style for these getter/setters? If it's still up for debate, I prefer:
```
/** ... */
@Input()
get value() {...}
set value(v) {...}
private _value;
```
It's a good reminder that I should be accessing it through the getter unless I have a good reason not to.

mmalerba:
whoops, didn't notice this was still wip

jelbourn:
@mmalerba as we discussed, adding support for changing on `value` change is problematic, going to leave it out for now.

@fxck `md-autosize` will simply override `rows` now, which I think is a sensible behavior.

mmalerba:
lgtm

PR #1848, opened 2016-11-13
feat(dialog): add configurable width, height and position

### LINKED ISSUE #1698: MdDialog Feature Request: Ability to change Width and off-set
#### Feature Request
Ability to set width of Dialog and dialog off-set (x,y) through input parameters so that dialogs can be made responsive to the screen media. For example on mobile media the programmer could set dialog box to be 100% and off-set would be set to 0,0 so it takes up the entire screen. For desktop media the programmer would set to default standard position that is in the centre.

crisbeto:
* Adds the ability to set a dialog's `width` and `height`.
* Adds the ability to set a dialog's position. If only the position in one axis is overridden, the other axis will stay centered.
* Fixes the `GlobalPositionStrategy` adding an unnecessary `0px` transform.
* Makes the dialog scrollable.
* Adds more options to the dialog demo so it's easier to test them out.

Fixes #1698.

### DIFF
--- a/src/lib/dialog/dialog-container.scss
+++ b/src/lib/dialog/dialog-container.scss
@@ -8,7 +8,12 @@ md-dialog-container {
   @include md-elevation(24);
 
   display: block;
-  overflow: hidden;
   padding: $md-dialog-padding;
   border-radius: $md-dialog-border-radius;
+  box-sizing: border-box;
+  overflow: auto;
+
+  // The dialog container should completely fill its parent overlay element.
+  width: 100%;
+  height: 100%;
 }
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -22,8 +22,6 @@ export {MdDialogRef} from './dialog-ref';
 
 // TODO(jelbourn): add support for opening with a TemplateRef
 // TODO(jelbourn): add `closeAll` method
-// TODO(jelbourn): default dialog config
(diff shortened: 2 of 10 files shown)


src/lib/dialog/dialog-container.scss
```diff
@@ -8,7 +8,10 @@ md-dialog-container {
   @include md-elevation(24);
 
   display: block;
-  overflow: hidden;
   padding: $md-dialog-padding;
   border-radius: $md-dialog-border-radius;
+  width: 100%;
```
jelbourn: Add a comment like
```ts
// The dialog container should completely fill its parent overlay element.
```

src/lib/dialog/dialog.ts
```diff
@@ -119,12 +117,26 @@ export class MdDialog {
...
+    let position = dialogConfig.position;
 
     state.hasBackdrop = true;
-    state.positionStrategy = this._overlay.position()
-        .global()
-        .centerHorizontally()
-        .centerVertically();
+    state.positionStrategy = strategy;
+
+    // TODO(crisbeto): replace with `horizontal` to support RTL?
```
jelbourn: I think that if someone is manually setting the position of the dialog that they should be in change of dealing with RTL.

crisbeto:
Addressed the feedback @jelbourn

jelbourn:
LGTM

Issue #120, opened 2016-02-27
md-chips


jelbourn:
Blocked on #117

bastienJS:
@jelbourn Does that mean this component will still be released in the alpha version (before beta)?

ollwenjones:
Chips seem sort of general purpose. It's not clear to me why their implementation would be blocked by autocomplete? Seems like it could be the other way around.

topherfangio:
@ollwenjones Still need to discuss with the team, but my goal is indeed to create a very basic version of the chips before we introduce the autocomplete functionality.

ollwenjones:
I'm not even sure understood what was meant by "Blocked on" - "Blocks" or "Blocked by" ? 😜

pankajparkar:
Any estimates? when this feature is going to roll-out?

topherfangio:
@pankajparkar A basic version of the chips has already launched: https://material.angular.io/components/component/chips

We are continuing to add functionality.

Dunos:
In the angularjs version, chips didn't allow any duplicate strings. It would be nice to have a specific parameter to allow/disallow this kind of behaviour since it may be needed in some situations.

topherfangio:
@Dunos If the chips are 100% string-based, then no, we can't have duplicates because then you can't find the proper `indexOf()` a chip since there could be multiple.

The Material 1 version did sort of support duplicates if you converted your chip to an object using [`md-transform-chip`](https://material.angularjs.org/latest/api/directive/mdChips) and gave each one a different ID.

However, as it stands right now, the Material 2 chips do not control anything about the actual chips that you provide. Since you provide the chips in an `ngFor` rather than an `ngModel`, you have total control.

Dunos:
@topherfangio in that case, you should be able to find a chip by its position in the array instead indexOf() imo

ahillhouse:
Do you have a timeline for when custom chip templates (like md-chip-template) will be available?  We currently use the angular 1 material chips and are in the process of upgrading to angular 2.

topherfangio:
@ahillhouse You should be able to utilize the `<md-basic-chip>` to remove any special material styling and do whatever you like. You can see the docs at https://material.angular.io/components/component/chips.

What exactly are you trying to achieve?

jshoudy11:
Are there plans to add delete functionality to chips? A little 'x' icon on each chip for example that will remove it from the chip list.

Irauan:
[...]

PR #1855, opened 2016-11-14
feat(chips): Initial skeleton/demo.

topherfangio:
Create the initial Chips skeleton with a very basic working demo of static, styled chips.

screen shot 2016-11-17 at 12 03 26 am

References #120.

### DIFF
--- a/src/lib/chips/chip-list.ts
+++ b/src/lib/chips/chip-list.ts
@@ -1,0 +1,44 @@
+import {
+  ChangeDetectionStrategy,
+  Component,
+  ElementRef,
+  ModuleWithProviders,
+  NgModule,
+  ViewEncapsulation
+} from '@angular/core';
+
+import {MdChip} from './chip';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-chip-list',
+  template: `<ng-content></ng-content>`,
+  host: {
+    // Properties
+    'tabindex': '0',
+    'role': 'listbox',
+    'class': 'md-chip-list'
+  },
+  styleUrls: ['chips.css'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MdChipList {
+  constructor(private _elementRef: ElementRef) {}
+
+  ngAfterContentInit(): void {}
+}
+
+@NgModule({
+  imports: [],
+  exports: [MdChipList, MdChip],
+  declarations: [MdChipList, MdChip]
+})
+export class MdChipsModule {
+  static forRoot(): ModuleWithProviders {
+    return {
+      ngModule: MdChipsModule,
+      providers: []
+    };
+  }
+}
--- a/src/lib/chips/chip.ts
+++ b/src/lib/chips/chip.ts
@@ -1,0 +1,17 @@
+import { Component, ElementRef, Renderer } from '@angular/core';
+
+@Component({
+  selector: 'md-chip, [md-chip]',
+  template: `<ng-content></ng-content>`,
+  host: {
+    // Properties
+    'class': 'md-chip',
+    'tabindex': '-1',
+    'role': 'option'
+  }
+})
+export class MdChip {
+  constructor(protected _renderer: Renderer, protected _elementRef: ElementRef) {}
+
+  ngAfterContentInit(): void {}
+}
--- a/src/lib/chips/chips.scss
+++ b/src/lib/chips/chips.scss
@@ -1,0 +1,16 @@
+$md-chip-vertical-padding: 8px;
+$md-chip-horizontal-padding: 12px;
+
+.md-chip-list {
+  padding: $md-chip-horizontal-padding;
+}
+
+.md-chip {
+  display: inline-block;
+  padding: $md-chip-vertical-padding $md-chip-horizontal-padding
+    $md-chip-vertical-padding $md-chip-horizontal-padding;
+  border-radius: $md-chip-horizontal-padding * 2;
+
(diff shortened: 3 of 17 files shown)


topherfangio:
@jelbourn Sorry; I just remembered that we discussed changing this to `md-chip-list`. I will do that and resubmit.

src/lib/chips/chip-list.ts
```diff
@@ -0,0 +1,55 @@
...
+} from '@angular/core';
+
+import {MdChip} from './chip';
+
+@Component(MdChipList.COMPONENT_CONFIG)
+export class MdChipList {
+
+  static COMPONENT_CONFIG: Component = {
+    moduleId: module.id,
+    selector: 'md-chip-list, [md-chip-list]',
```
jelbourn: I don't think we need to do the attribute selector until we have a strong use-case for it.

src/lib/chips/chip-list.ts
```diff
@@ -0,0 +1,55 @@
...
+    template: `<ng-content></ng-content>`,
+    host: {
+      // Properties
+      'tabindex': '0',
+      'role': 'listbox',
+    },
+    queries: {
+      items: new ContentChildren(MdChip)
```
jelbourn: I suspect this specifically (using `new` Something in this metadata) won't work in AoT compile.

src/lib/chips/chip-list.ts
```diff
@@ -0,0 +1,55 @@
...
+  QueryList,
+  ViewEncapsulation
+} from '@angular/core';
+
+import {MdChip} from './chip';
+
+@Component(MdChipList.COMPONENT_CONFIG)
+export class MdChipList {
+
+  static COMPONENT_CONFIG: Component = {
```
jelbourn: If this does work in AoT mode, then it should just be an exported constant, not a static member on the `MdChipList`

Either way, though, I'm tempted to say let's not worry about this for the initial version and add a TODO to explore the idea later.

src/lib/chips/chip-list.ts
```diff
@@ -0,0 +1,55 @@
...
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush
+  };
+
+  protected items: QueryList<MdChip>;
+
+  constructor(private _elementRef: ElementRef) {}
+
+  ngAfterContentInit(): void {
+    this._elementRef.nativeElement.classList.add('md-chips');
```
jelbourn: You should be able to add `class` to the `host` oject in the component metadata for this:
```ts
host: {
  'class': 'md-chip-list'
}
```
Same for chip below.

src/lib/chips/chip.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  Component,
+  ElementRef,
+  Renderer
+} from '@angular/core';
```
jelbourn: Only need to wrap like this if it's longer than 100 cols.

src/lib/chips/chips.scss
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.md-chip {
+  display: inline-block;
+  padding: $md-chip-vertical-padding $md-chip-horizontal-padding
+    $md-chip-vertical-padding $md-chip-horizontal-padding;
+  border-radius: $md-chip-horizontal-padding * 2;
+
+  background-color: #E0E0E0;
+  color: rgba(0, 0, 0, 0.87);
```
jelbourn: All colors should be in a `_chips-theme.scss` file. See the other components for an example.

src/lib/chips/chips.scss
```diff
@@ -0,0 +1,18 @@
...
+  font-size: 13px;
+  line-height: 16px;
```
jelbourn: I'm not sure we should be setting `font-size` and `line-height`. Could we use whatever value is current applied on this element?

@hansl thoughts on this?

jelbourn:
Also there's a lint error on CI.

src/lib/chips/chips.scss
topherfangio: Would love some additional guidance on this. Those values were set based off of the spec, so if we are allowing users to deviate, it may change some other values (like the border raidus in particular).

Is there any plan to use `em` or `rem` instead of pixels?

src/lib/chips/chip-list.ts
topherfangio: I will play with this in the AoT and see what happens, but I realized it's actually not used in the skeleton (I built the skeleton by deleting code from what I showed you before) so I will remove it from this PR and probably have it in my next one.

topherfangio:
@jelbourn Ready for re-review. Made all requested changes except for removing font-size/line-height as I'm not sure what to replace it with. Would love some more feedback on that.

jelbourn:
LGTM

jelbourn:
@kara FYI this will need new build rules.

src/lib/chips/chips.scss
```diff
@@ -0,0 +1,16 @@
+$md-chip-vertical-padding: 8px;
+$md-chip-horizontal-padding: 12px;
+
+.md-chip-list {
+  padding: $md-chip-horizontal-padding;
+}
+
+.md-chip {
+  display: inline-block;
+  padding: $md-chip-vertical-padding $md-chip-horizontal-padding
```
fxck: no need to repeat these, 

```
padding: $md-chip-vertical-padding $md-chip-horizontal-padding;
```

is enough

topherfangio: Indeed; will fix in my next PR.

PR #1856, opened 2016-11-14
feat(snackbar): add timeout for snackbar

josephperrott:


src/lib/snack-bar/snack-bar-config.ts
```diff
@@ -13,4 +13,7 @@ export class MdSnackBarConfig {
 
   /** The view container to place the overlay for the snack bar into. */
   viewContainerRef?: ViewContainerRef = null;
+
+  /** The length of time in milliseconds to wait before automatically dismissing the snack bar. */
+  autoHide?: number = 0;
```
jelbourn: Let's just call this `duration`.

src/lib/snack-bar/snack-bar.ts
```diff
@@ -64,6 +62,10 @@ export class MdSnackBar {
     } else {
       snackBarRef.containerInstance.enter();
     }
+    if (config.autoHide > 0) {
+      setTimeout(() => snackBarRef.dismiss(), config.autoHide + 225);
```
jelbourn: Why is the `225` baked in here? (I'm assuming it's the animation duration, but doesn't dismiss handle that already?)

src/lib/snack-bar/snack-bar.spec.ts
```diff
@@ -259,6 +260,55 @@ describe('MdSnackBar', () => {
...
+      });
+
+      viewContainerFixture.detectChanges();
+      viewContainerFixture.whenStable().then(() => {
+        expect(dismissObservableCompleted).toBeFalsy('Expected the snack bar not to be dismissed');
+      });
+
+      tick(1000);
+      viewContainerFixture.detectChanges();
+      viewContainerFixture.whenStable().then(() => {
```
jelbourn: You shouldn't need to use `whenStable` with `fakeAsync` because the `whenStable` will just be called synchronously. 

(same for other uses)

src/lib/snack-bar/snack-bar.ts
```diff
@@ -64,6 +62,10 @@ export class MdSnackBar {
     } else {
       snackBarRef.containerInstance.enter();
     }
+    if (config.autoHide > 0) {
```
jelbourn: We should do `config.duration && cofig.duration > 0` in case someone sets it to null or undefined.

josephperrott: Can we use strict null checking?

We can then use the Non-null assertion operator

for instance in this case it would be:

`if (config.autoHide! > 0) {`

src/lib/snack-bar/snack-bar.spec.ts
josephperrott: I was unable to get my tests working without whenStable, so I will have to dig in more there.

src/lib/snack-bar/snack-bar-config.ts
josephperrott: done

src/lib/snack-bar/snack-bar.ts
josephperrott: Forgot I had this in there.

Yes, the 225 is for the animation.  Created a todo to handle the case I was looking at.

src/lib/snack-bar/snack-bar.spec.ts
josephperrott: Now it works, not sure what happened before.

src/lib/snack-bar/snack-bar.ts
josephperrott: Done.

src/lib/snack-bar/snack-bar-config.ts
```diff
@@ -13,4 +13,7 @@ export class MdSnackBarConfig {
 
   /** The view container to place the overlay for the snack bar into. */
   viewContainerRef?: ViewContainerRef = null;
+
+  /** The length of time in milliseconds to wait before automatically dismissing the snack bar. */
+  dismiss?: number = 0;
```
josephperrott: Done

src/lib/snack-bar/README.md
```diff
@@ -15,6 +15,7 @@
 | `viewContainerRef: ViewContainerRef` | The view container ref to attach the snack bar to. |
 | `role: AriaLivePoliteness = 'assertive'` | The politeness level to announce the snack bar with. |
 | `announcementMessage: string` | The message to announce with the snack bar. |
+| `dismiss: number` | The length of time in milliseconds to wait before autodismissing the snack bar. |
```
jelbourn: Missed `dismiss` > `duration` here

jelbourn:
LGTM

PR #1857, opened 2016-11-14
change elevations to match spec

mmalerba:
spec: https://material.googleplex.com/material-design/elevation-shadows.html

### DIFF
--- a/src/lib/card/card.scss
+++ b/src/lib/card/card.scss
@@ -17,10 +17,6 @@ md-card {
   font-family: $md-font-family;
 }
 
-md-card:hover {
-  @include md-elevation(8);
-}
-
 .md-card-flat {
   box-shadow: none;
 }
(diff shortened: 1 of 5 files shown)


src/lib/card/card.scss
```diff
@@ -18,7 +18,7 @@ md-card {
 }
 
 md-card:hover {
-  @include md-elevation(2);
+  @include md-elevation(8);
```
traviskaufman: cards should not change elevation when hovered over. The [spec](https://material.google.com/material-design/elevation-shadows.html#elevation-shadows-elevation-android) language there says a _picked up_ card should have an elevation of 8. This is a special type of card that is meant to be interacted with, not a general-purpose card.

mmalerba: Ok, I don't think we have a concept of lifted cards yet, so removing

jelbourn:
LGTM

PR #1858, opened 2016-11-14
feature(md-input): refactor MdInput as an attribute

hansl:
using Projection.


jelbourn:
@hansl seems to be some CI failures.

hansl:
Seems like it. Looking.

src/lib/input/input.ts
```diff
@@ -41,334 +36,104 @@ const MD_INPUT_INVALID_INPUT_TYPE = [
...
-  get characterCount(): number {
-    return this.empty ? 0 : ('' + this._value).length;
+  @Input('class') _cssClass: string = '';
+  @Input('style') _cssStyle: string = '';
+  get _safeCssStyle(): SafeStyle {
+    return this._dom.bypassSecurityTrustStyle(this._cssStyle || '');
   }
-  get inputId(): string { return `${this.id}-input`; }
+  @HostBinding('attr.class') get _attrClass(): any { return null; }
+  @HostBinding('attr.style') get _attrStyle(): any { return null; }
```
jelbourn: I also just thought about the fact that `ngClass`, `ngStyle`, `[class.whatever]`, and `[style.whatever]` will all potentially be a problem here. We should chat about dealing with these (potentially with Tobias)

src/lib/input/placeholder.html
```diff
@@ -0,0 +1,12 @@
+<label class="md-placeholder"
```
jelbourn: It might be a problem if the placeholder is _always_ the label for the input. I can see cases where you would want to provide a more specific label than what's in the placeholder.

src/lib/input/placeholder.ts
```diff
@@ -0,0 +1,81 @@
...
+  OpaqueToken,
+  TemplateRef, Component, ViewContainerRef, ComponentRef, HostBinding, ViewChild, HostListener,
+  AfterContentInit, ReflectiveInjector, Injector,
+} from '@angular/core';
+import {PortalHost} from '../core';
+import {ComponentPortal} from '../core/portal/portal';
+import {coerceBooleanProperty} from '../core/coersion/boolean-property';
+
+
+export const MD_PLACEHOLDER_HOST_TOKEN  = new OpaqueToken('mdPlaceholderHost');
```
jelbourn: These don't normally have `_TOKEN` at the end (e.g., `NG_VALIDATORS`). 

Add description comment for the token.

src/lib/input/input.scss
```diff
@@ -57,7 +57,7 @@ md-input, md-textarea {
 }
 
 // The Input element proper.
-.md-input-element {
+input[md-input], textarea[md-textarea] {
```
hansl: Done.

src/lib/input/input.html
```diff
@@ -1,86 +1,21 @@
-<div class="md-input-wrapper">
+<div (click)="focus()"
+     [class]="_cssClass"
+     [attr.style]="_safeCssStyle"
+     [class.md-end]="align == 'end'"
+     [class.md-input-wrapper]="true">
```
hansl: Done

src/lib/input/input.spec.ts
```diff
@@ -633,182 +96,39 @@ describe('MdInput', function () {
...
-  label: string = '';
-}
-
-@Component({template: `<md-input [hintLabel]="label"></md-input>`})
-class MdInputHintLabelTestController {
-  label: string = '';
-}
+@Component({template: `
+  <input md-input [placeholder]="p">
+  <template #p>template placeholder</template>
```
hansl: Done.

src/lib/input/placeholder.html
hansl: We tested it with the old interface, and there wasn't a problem.

src/lib/input/placeholder.ts
hansl: Done.

hansl:
@jelbourn PTAL.

jelbourn:
Closing in favor of #2052

PR #1860, opened 2016-11-14
fix(input): prevent overlapping placeholder on inputs with native masking

### LINKED ISSUE #1853: Input type datetime-local overlaps with placeholder
When the value string is empty, the placeholder of the md-input will overlap with the field.
image

The placeholder should stay above the field

The placeholder overlaps with the field

Bind an empty string to md-input with type `datetime-local`.

http://plnkr.co/edit/p5vpltRGiy5SQXME2AAh?p=preview

It's broken.

Angular 2.1.2
Material Alpha 10, 9, 8 at least.
Windows 10
Chrome (datetime-local is only supported in chrome afaik).

This was fixed for `type="date"` in https://github.com/angular/material2/issues/1238 but not for `type="datetime-local"`.

crisbeto:
Prevents the Material placeholder from overlapping with native input masking (e.g. on `input[type="time"]`.
Previously, this was hardcoded for `date`, however it didn't take into account that not all browsers support `date` inputs.
This change runs a feature test on each of the types that should have input masking and only triggers the special behavior if the type is supported.

Fixes #1853.

### DIFF
--- a/src/lib/input/input.spec.ts
+++ b/src/lib/input/input.spec.ts
@@ -5,7 +5,7 @@ import {
 import {Component} from '@angular/core';
 import {FormsModule} from '@angular/forms';
 import {By} from '@angular/platform-browser';
-import {MdInput, MdInputModule} from './input';
+import {MdInput, MdInputModule, TYPES_WITH_MASK} from './input';
 
 function isInternetExplorer11() {
     return 'ActiveXObject' in window;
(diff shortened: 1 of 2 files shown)


jelbourn:
@crisbeto I think this will conflict with #1858

crisbeto:
It's expected. I'll resolve the conflicts if #1858 gets merged first.

jelbourn:
Will be done as part of #2052

PR #1861, opened 2016-11-14
fix(input): error in IE on unsupported input types

crisbeto:
Fixes IE throwing a template error and stopping all rendering if an unsupported input type is set on an `md-input` instance.


jelbourn:
@crisbeto this will be obsolete with #1858

PR #1863, opened 2016-11-15
fix(connected-position): fix position emitter

kara:
Realized my first implementation made no sense because if you scroll back to a position where the default orientation will fit, it won't fire a position change event.  Now it will fire a change event whenever the position changes, including the first time.


hansl:
LGTM.

PR #1871, opened 2016-11-15
fix(input): make template reference variable unique

nunojpg:
Two elements had the same template reference so one was being
disregarded and as so the correspondent binding did not refer
to the desired element.

Signed-off-by: Nuno Goncalves <nunojpg@gmail.com>


jelbourn:
`md-input` is being refactored in #1858, so this will be obsolete soon.

PR #1873, opened 2016-11-15
feat(select): support fallback positions

kara:
This PR adds fallback position support to the select.

r: @jelbourn


kara:
@jelbourn This is rebased now

src/lib/select/select.spec.ts
```diff
@@ -19,17 +20,31 @@ describe('MdSelect', () => {
       providers: [
         {provide: OverlayContainer, useFactory: () => {
           overlayContainerElement = document.createElement('div');
+          overlayContainerElement.style.position = 'fixed';
```
jelbourn: Add comment explaining why you add these styles

jelbourn:
LGTM

PR #1877, opened 2016-11-15
fix(tab-link): avoid potential memory leak

crisbeto:
* Similarly to #1838, the `tab-link` destroy handler may not be called in certain situations, because it is being inherited from the MdRipple class. This PR explicitly calls the destroy handler.
* Adds a unit test to make sure that the ripples are being cleaned up properly.

### DIFF
--- a/src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
+++ b/src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
@@ -1,4 +1,13 @@
-import {Component, Input, ViewChild, ElementRef, ViewEncapsulation, Directive} from '@angular/core';
+import {
+  Component,
+  Input,
+  ViewChild,
+  ElementRef,
+  ViewEncapsulation,
+  Directive,
+  OnDestroy,
+} from '@angular/core';
+
 import {MdInkBar} from '../ink-bar';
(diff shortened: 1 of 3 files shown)


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -50,8 +59,14 @@ export class MdTabLink {
...
+export class MdTabLinkRipple extends MdRipple implements OnDestroy {
   constructor(private _element: ElementRef) {
     super(_element);
   }
+
+  // In certain cases the parent destroy handler
+  // may not get called. See Angular issue #11606.
+  ngOnDestroy() {
+    super.ngOnDestroy();
+  }
```
crisbeto: I couldn't find a good way of doing it and opted not to since the ripple didn't have a test for it either. Should I spend some time adding it to both places?

jelbourn: Yeah, it would be worth it to have a test for both.

crisbeto:
Updated with a unit test for the ripple and the tab link @jelbourn

jelbourn:
LGTM

PR #1878, opened 2016-11-15
feat(slider): vertical mode

mmalerba:
had to rework the css a bit, but its for the better since all transitions are now either transform or opacity

### DIFF
--- a/src/lib/slider/slider.spec.ts
+++ b/src/lib/slider/slider.spec.ts
@@ -12,8 +12,7 @@ import {
   PAGE_DOWN,
   PAGE_UP,
   END,
-  HOME,
-  LEFT_ARROW
+  HOME, LEFT_ARROW
 } from '../core/keyboard/keycodes';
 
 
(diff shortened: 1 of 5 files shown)


mmalerba:
blocked on #1794

src/lib/slider/slider.spec.ts
```diff
@@ -106,21 +110,21 @@ describe('MdSlider', () => {
     });
 
     it('should update the track fill on click', () => {
-      expect(trackFillElement.style.flexBasis).toBe('0%');
+      expect(trackFillElement.style.transform).toBe('scaleX(0)');
```
jelbourn: Just curious, why change from `flex-basis` to `transform` instead of changing the flex axis? Just the general preference to `transform`?

src/lib/slider/slider.ts
```diff
@@ -283,13 +360,20 @@ export class MdSlider implements ControlValueAccessor {
         this.value = this.min;
         break;
       case LEFT_ARROW:
-        this._increment(-1);
+        // It's kind of debatable what's the correct thing to do for inverted sliders. For a sighted
```
jelbourn: I'd change "kind of debatable" bit and simply say that, for inverted sliders, we care more about the a11y experience than making having the keyboard shortcuts "look right" for sighted users. (I do think this is the right call).

src/lib/slider/slider.spec.ts
mmalerba: Yeah, I realized I could do it all with transform and figured that was preferable.

src/lib/slider/slider.ts
mmalerba: Done

PR #1887, opened 2016-11-16
feat: use typedoc to generate a json file documentation dump of all components

ericjim:
@jelbourn 

~~I've modified `button.ts` as an example on how to hide underscored methods and to hide its constructor.~~

### DIFF
--- a/GENERATING_DOCS.md
+++ b/GENERATING_DOCS.md
@@ -1,0 +1,30 @@
+# Generating Docs with Typedoc
+[Typedoc](http://typedoc.org/) is a documentation generator for TypeScript projects.
+
+Refer to `typedoc.json` for build configuration.
+
+1. `npm install`
+2. `npm run docs`
+
+TypeDoc will leverage the typings file to generate docs. The output will be located in `/dist/api/api.json`.
+
+## FAQ
+
+### How do I stop HTML generation and focus only on JSON?
+Currently not possible, submitted a feature request: https://github.com/TypeStrong/typedoc/issues/337
+
+### How do I prevent `Cannot find name` errors?
+When typedoc is invoked tsconfig and typings file must be in the path.
+
+### How do we omit constructors or other methods from the components?
+You can either ignore them from the output file, or use the `@hidden` tag in the source code to omit a method or
+property from being generated as docs. The recommended approach is to ignore them since adding the hidden tag will
+create a lot of noise
+
+### Can we detect Inputs and Outputs from Angular components?
+Yes, check the decorators key-value on the json file for that particular component.
+
+### Can we generate a json file for a specific component?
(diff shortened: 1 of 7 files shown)


GENERATING_DOCS.md
```diff
@@ -0,0 +1,44 @@
...
+
+```
+npm run docs ./src/lib/button
+```
+
+This will produces typings errors because the specified path does not have a typings file.
+
+## Example Usage
+
+### Hiding methods in a class
```
jelbourn: Rather than hiding the methods with `@hidden`, I'd rather just ignore them when we're processing the json for the docs site. The vast majority of methods and properties won't be public, so using `@hidden` would add a lot of noise to the source.

GENERATING_DOCS.md
```diff
@@ -0,0 +1,44 @@
...
+1. `npm install`
+2. `npm run docs ./src/lib`
+
+TypeDoc will leverage the typings file to generate docs. The output will be located in `/docs/autogen/output.json`.
+
+
+## Generating docs for one component
+
+```
+npm run docs ./src/lib/button
```
jelbourn: We should make a gulp task for this.

typedoc.json
```diff
@@ -0,0 +1,21 @@
+{
+  "mode": "modules",
+  "out": "./docs/autogen",
```
jelbourn: We should output to `dist/api`

typedoc.json
```diff
@@ -0,0 +1,21 @@
...
+  "json": "./docs/autogen/output.json",
+  "theme": "default",
+  "ignoreCompilerErrors": "true",
+  "experimentalDecorators": "true",
+  "emitDecoratorMetadata": "true",
+  "excludePrivate": "true",
+  "excludeExternals": "true",
+  "includeDeclarations": "true",
+  "target": "ES6",
+  "exclude": "**/+(index.ts|*.spec.ts|*.scss|*.html)",
```
jelbourn: We don't necessarily want to exclude `index.ts` because we do define some constructs there. 
(though maybe we shouldn't)

typedoc.json
```diff
@@ -0,0 +1,21 @@
+{
+  "mode": "modules",
+  "out": "./docs/autogen",
+  "json": "./docs/autogen/output.json",
+  "theme": "default",
+  "ignoreCompilerErrors": "true",
```
jelbourn: Why `ignoreCompilerErrors`?

ericjim: I'll remove it, it was useful for testing. Individual components which didn't have typings in the same folder (`src/lib/button`).

typedoc.json
ericjim: do you have a preference name for the json file? I call it `output.json` though `api.json` might be more fitting.

jelbourn:
LGTM

Issue #559, opened 2016-05-27
Add e2e test for progress-bar and progress-circle


jelbourn:

jelbourn:
@deerawan sure, we should have instructions for running the e2e tests in our dev_environment guide soon.

jelbourn:
@deerawan Instructions should be there now.

emoralesb05:
Just FYI, protractor times out when using the progress circle in a page since the `setInterval()` does not let the page sync with protractor when its in `indetermined` mode.

PR #1895, opened 2016-11-16
test(progress-bar): add basic e2e tests

crisbeto:
Adds basic e2e tests that assert that the different progress bars are being rendered.

Referencing #559.

### DIFF
--- a/src/e2e-app/menu/menu-e2e.ts
+++ b/src/e2e-app/menu/menu-e2e.ts
@@ -12,8 +12,7 @@ import {Component} from '@angular/core';
     }
 
     .bottom-row {
-      position: absolute;
-      top: 200px;
+      margin-top: 5px;
     }
   `]
 })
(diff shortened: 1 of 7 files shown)


crisbeto:
@kara can you take a look at the menu e2e change? I wasn't sure what the reasoning behind the `bottom: 200px` was, but it ended up causing the menu test triggers to overlap, which in turn caused tests to fail. This was because the list of links above was getting too tall.

src/e2e-app/menu/menu-e2e.ts
```diff
@@ -12,8 +12,7 @@ import {Component} from '@angular/core';
     }
 
     .bottom-row {
-      position: absolute;
-      top: 200px;
+      margin-top: 5px;
```
jelbourn: Is this intentional?

kara:
@crisbeto Ack, I can't remember. I assume it had to do with triggers being clickable at some point, but if the e2e tests pass now, it probably is no longer necessary. LGTM.

PR #1898, opened 2016-11-16
fix(select): add aria-owns property

kara:
r: @jelbourn

### DIFF
--- a/src/lib/select/option.ts
+++ b/src/lib/select/option.ts
@@ -10,6 +10,12 @@ import {
 import {ENTER, SPACE} from '../core/keyboard/keycodes';
 import {coerceBooleanProperty} from '../core/coersion/boolean-property';
 
+/**
+ * Option IDs need to be unique across components, so this counter exists outside of
+ * the component definition.
+ */
+let _uniqueIdCounter = 0;
+
 @Component({
   moduleId: module.id,
   selector: 'md-option',
@@ -17,6 +23,7 @@ import {coerceBooleanProperty} from '../core/coersion/boolean-property';
     'role': 'option',
     '[attr.tabindex]': '_getTabIndex()',
     '[class.md-selected]': 'selected',
+    '[id]': 'id',
     '[attr.aria-selected]': 'selected.toString()',
     '[attr.aria-disabled]': 'disabled.toString()',
     '[class.md-option-disabled]': 'disabled',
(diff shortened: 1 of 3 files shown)


src/lib/select/option.ts
```diff
@@ -10,13 +10,20 @@ import {
 import {ENTER, SPACE} from '../core/keyboard/keycodes';
 import {coerceBooleanProperty} from '../core/coersion/boolean-property';
 
+/**
+ * Option IDs need to be unique across components, so this counter exists outside of
+ * the component definition.
+ */
+var _selectOptionCounter = 0;
```
jelbourn: How about?
```ts
/** Unique ID counter for md-option. */
let _uniqueIdCounter = 0;
```

src/lib/select/select.ts
```diff
@@ -291,6 +300,11 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
     this._subscriptions = [];
   }
 
+  /** Records option IDs to pass to the aria-owns property. */
+  private _setOptionIds() {
+    this._optionIds = this.options.map((option) => option.id).join(' ');
```
jelbourn: nit: don't need parens around `option`

src/lib/select/select.spec.ts
```diff
@@ -597,6 +596,59 @@ describe('MdSelect', () => {
...
+      it('should set aria-owns properly', () => {
+        const selects = fixture.debugElement.queryAll(By.css('md-select'));
+
+        expect(selects[0].nativeElement.getAttribute('aria-owns'))
+            .toContain('md-select-option', `Expected option IDs to be set in aria-owns attr.`);
+
+        expect(selects[0].nativeElement.getAttribute('aria-owns'))
+            .not.toEqual(selects[1].nativeElement.getAttribute('aria-owns'),
+                `Expected select components to have unique option IDs in aria-owns attr.`);
+      });
```
jelbourn: Need a test where you assert that the `aria-owns` contains each of the options's IDs. Something like
```
expect(ariaOwns).toContain(options[0].id);
expect(ariaOwns).toContain(options[1].id);
...
```

kara:
@jelbourn Comments addressed!

jelbourn:
LGTM

PR #1901, opened 2016-11-17
fix(slider): emit change event after updating value accessor

### LINKED ISSUE #1802: (slideend) is not working as expected with [(ngModel)]  [material version is alpha 10]
i did two way binding md-slider with class attribute 'sliderVal'. as below code.

<md-slider (slideend)="onSlideEnd();" [min]="0" [max]="100"  [(ngModel)]="sliderVal"></md-slider>

//class members

sliderVal: number = 10;

onSlideEnd(){
                console.log("slider value", this.sliderVal);
    }

but the problem is when onSlideEnd() is called. its shows previous slider value in console.

mmalerba:
fixes #1802

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -336,9 +336,9 @@ export class MdSlider implements ControlValueAccessor {
       let event = new MdSliderChange();
       event.source = this;
       event.value = this.value;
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #1904, opened 2016-11-17
chore: fix closure compiler warning in ripple

jelbourn:
R: @mmalerba 

Also scrunched some comments.


src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -42,6 +42,7 @@ export class RippleRenderer {
   private _backgroundDiv: HTMLElement;
   private _rippleElement: HTMLElement;
   private _triggerElement: HTMLElement;
+  _opacity: string;
```
mmalerba: can this be made private?

jelbourn: If it were private then it would be seen as assigned but never used. If it's public, an optimizer doesn't know if an outside class will refer to it.

PR #1907, opened 2016-11-17
fix(ripple): Fix the ripple position

### LINKED ISSUE #1817: Ripple positioning is off beneath the fold

tinayuangao:
When scroll down the page, the ripple position is off
Closes #1817

Original PR #1820 is accidentally closed.

R: @kara @jelbourn

### DIFF
--- a/src/lib/button/button.spec.ts
+++ b/src/lib/button/button.spec.ts
@@ -2,6 +2,7 @@ import {async, TestBed, ComponentFixture} from '@angular/core/testing';
 import {Component} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {MdButtonModule} from './button';
+import {ViewportRuler} from '../core/overlay/position/viewport-ruler';
 
 
 describe('MdButton', () => {
@@ -10,6 +11,9 @@ describe('MdButton', () => {
     TestBed.configureTestingModule({
       imports: [MdButtonModule.forRoot()],
       declarations: [TestApp],
+      providers: [
+        {provide: ViewportRuler, useClass: FakeViewportRuler},
+      ]
     });
 
     TestBed.compileComponents();
@@ -208,5 +212,17 @@ class TestApp {
 
   increment() {
     this.clickCount++;
+  }
+}
+
+class FakeViewportRuler {
+  getViewportRect() {
+    return {
+      left: 0, top: 0, width: 1014, height: 686, bottom: 686, right: 1014
+    };
+  }
+
+  getViewportScrollPosition() {
+    return {top: 0, left: 0};
   }
 }
--- a/src/lib/button/button.spec.ts
+++ b/src/lib/button/button.spec.ts
@@ -215,4 +215,3 @@ class TestApp {
     this.clickCount++;
   }
 }
-
--- a/src/lib/button/button.spec.ts
+++ b/src/lib/button/button.spec.ts
@@ -3,6 +3,7 @@ import {Component} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {MdButtonModule} from './button';
 import {ViewportRuler} from '../core/overlay/position/viewport-ruler';
+import {FakeViewportRuler} from '../core/overlay/position/fake-viewport-ruler';
 
 
 describe('MdButton', () => {
@@ -215,14 +216,3 @@ class TestApp {
   }
 }
 
-class FakeViewportRuler {
-  getViewportRect() {
-    return {
-      left: 0, top: 0, width: 1014, height: 686, bottom: 686, right: 1014
-    };
-  }
-
-  getViewportScrollPosition() {
-    return {top: 0, left: 0};
-  }
-}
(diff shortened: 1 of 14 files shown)


src/lib/core/ripple/ripple.spec.ts
```diff
@@ -180,6 +180,47 @@ describe('MdRipple', () => {
...
+    it('create ripple with correct position when page is scrolled', () => {
+      let elementTop = 600;
+      let elementLeft = 750;
+      let pageScrollTop = 500;
+      let pageScrollLeft = 500;
+      let left = 50;
+      let top = 75;
+
+      // Add a very large element to make the page scroll
+      let veryLargeElement: HTMLElement = document.createElement('div');
```
jelbourn: You don't need the type here; `document.createElement('div');` will return an `HTMLDivElement`

(typescript understands string literals in its type system so it can know things like `document.createElement` or `addEventListener` can have different arguments based on the string based to them)

src/lib/core/ripple/ripple.spec.ts
```diff
@@ -180,6 +180,47 @@ describe('MdRipple', () => {
...
+      veryLargeElement.style.width = '4000px';
+      veryLargeElement.style.height = '4000px';
+      document.body.appendChild(veryLargeElement);
+      document.body.scrollTop = pageScrollTop;
+      document.body.scrollLeft = pageScrollLeft;
+
+      rippleElement.style.left = `${elementLeft}px`;
+      rippleElement.style.top = `${elementTop}px`;
```
jelbourn: * Setting the `left` and `top` of the rippleElement probably won't do what you want if the element isn't `position: absolute`. 
* I'm not sure if the rippleElement is in the DOM; it would have to be for the positioning to be right.

This might be why the tests are failing.

src/lib/core/ripple/ripple.ts
```diff
@@ -161,7 +161,11 @@ export class MdRipple implements OnInit, OnDestroy, OnChanges {
       // FIXME: This fails on IE11, which still sets pageX/Y and screenX/Y on keyboard clicks.
       const isKeyEvent =
           (event.screenX === 0 && event.screenY === 0 && event.pageX === 0 && event.pageY === 0);
-      this.end(event.pageX, event.pageY, isKeyEvent);
+      var scrollTop = window.pageYOffset || document.documentElement.scrollTop ||
+          document.body.scrollTop || 0;
+      var scrollLeft = window.pageXOffset || document.documentElement.scrollLeft ||
+          document.body.scrollLeft || 0;
+      this.end(event.pageX - scrollLeft, event.pageY - scrollTop, isKeyEvent);
```
jelbourn: Can you use `ViewportRuler.getViewportScrollPosition` for this?

tinayuangao:
Tests fixed. PTAL. Thanks!

jelbourn:
@tinayuangao can you rebase this PR?

tinayuangao:
Rebased and fixed the test.

src/lib/core/overlay/position/viewport-ruler.ts
```diff
@@ -50,12 +50,8 @@ export class ViewportRuler {
...
     // `document.documentElement` works consistently, where the `top` and `left` values will
     // equal negative the scroll position.
-    const top =  documentRect.top < 0 && document.body.scrollTop == 0 ?
-        -documentRect.top :
-        document.body.scrollTop;
-    const left = documentRect.left < 0 && document.body.scrollLeft == 0 ?
-        -documentRect.left :
-        document.body.scrollLeft;
+    const top = -documentRect.top || document.body.scrollTop || window.scrollY || 0;
+    const left = -documentRect.left || document.body.scrollLeft || window.scrollX || 0;
```
jelbourn: What was your reason for changing this?

src/lib/button/button.spec.ts
```diff
@@ -210,3 +214,15 @@ class TestApp {
     this.clickCount++;
   }
 }
+
+class FakeViewportRuler {
```
jelbourn: Can we move this to its own file instead of duplicating it? Having `fake-viewport-ruler.ts` under `overlay/position` would work.

src/lib/core/overlay/position/viewport-ruler.ts
tinayuangao: The mobile safari/IE test will fail without the window.scrollX/Y case.

jelbourn:
LGTM

mmalerba:
Please add `ViewportRuler` to the `forRoot()` `providers` of all components that use `MdRipple`.

tinayuangao:
Done. Added `ViewportRuler` to the `forRoot()` providers.

jelbourn:
Tested this on IE11, Mobile Safari (9), and Firefox and everything seemed good.

jelbourn:
@mmalerba I rebased this and updated the PR

PR #1915, opened 2016-11-17
fix(ripple): Always remove ripple after a certain period

### LINKED ISSUE #1818: Ripple is never removed if its transition is interrupted
Currently the ripple detaches itself on "transitionend". If the transition is interrupted (for instance, its host element is detached from the DOM before it's done), the ripple will never detach properly. When its host is re-attached, the ripple background will still be present.

tinayuangao:
The ripple foreground is not removed if transition is interrupt (e.g. change parent visibility, hidden). 

Fixes #1818

R: @kara @jelbourn

### DIFF
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -155,7 +155,9 @@ export class RippleRenderer {
 
   /** Removes a foreground ripple from the DOM after it has faded out. */
   removeRippleFromDom(ripple: Element) {
-    ripple.parentElement.removeChild(ripple);
+    if (ripple && ripple.parentElement) {
+      ripple.parentElement.removeChild(ripple);
+    }
   }
 
   /** Fades in the ripple background. */
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -151,7 +151,8 @@ export class RippleRenderer {
     // If it were to run inside the Angular zone, every test that used ripples would have to be
     // either async or fakeAsync.
     this._ngZone.runOutsideAngular(() => {
-      // The ripple lasts a time equal to the sum of fade-in, transform, and fade-out (3 * fade-in time).
+      // The ripple lasts a time equal to the sum of fade-in, transform,
+      // and fade-out (3 * fade-in time).
       let rippleDuration =  fadeInSeconds * 3 * 1000;
       setTimeout(() => this.removeRippleFromDom(ripple.rippleElement), rippleDuration);
     });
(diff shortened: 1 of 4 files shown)


src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -143,6 +143,8 @@ export class RippleRenderer {
 
     rippleDiv.addEventListener('transitionend',
         (event: TransitionEvent) => transitionEndCallback(ripple, event));
+    // Remove the ripple after transitions (fade-in, transform, fade-out)
+    setTimeout(() => this.removeRippleFromDom(ripple.rippleElement), fadeInSeconds * 3 * 1000);
```
kara: This will require all related tests to be async, right?  @jelbourn Should we run this outside angular zone?

tinayuangao: Sometimes the element is only `hidden`, or `display: none`. In this case `ngOnDestroy` is never called. 

If we use `*ngIf` to remove the element, the ripple will be removed when it's removed from the DOM tree.

jelbourn:
```ts
// Ensure that ripples are always removed, even when transitionend doesn't fire.
// Run this outside the Angular zone because there's nothing that Angular cares about.
// If it were to run inside the Angular zone, every test that used ripples would have to be
// either async or fakeAsync.
this._ngZone.runOutsideAngular(() => {
  // The ripple lasts a time equal to the sum of fade-in, transform, and fade-out (3 * fade-in time).
  let rippleDuration =  fadeInSeconds * 3 * 1000;
  setTimeout(() => this.removeRippleFromDom(ripple.rippleElement), rippleDuration);
});

Comment (tinayuangao) [inline src/lib/core/ripple/ripple-renderer.ts]: Done. PTAL. 
Thanks for review!

File: src/lib/core/ripple/ripple.spec.ts [hunk 2]
```diff
@@ -132,6 +132,15 @@ describe('MdRipple', () => {
       expect(rippleElement.querySelectorAll('.md-ripple-foreground').length).toBe(0);
     });
 
+    it('removes foreground ripples after timeout', async(() => {
+      rippleElement.click();
+      expect(rippleElement.querySelectorAll('.md-ripple-foreground').length).toBe(1);
+
+      setTimeout(() => {
```
jelbourn: Add a comment like 
```ts
// Use a real timeout because the ripple's timeout runs outside of the angular zone.
```

jelbourn:
Still seems to be failing on IE11... I'll try to take a look today.

tinayuangao:
Fixed IE11

Issue #1899, opened 2016-11-16
Bug: MdSidenav._closePromiseReject on toggle Sidenav (issues with _onTransitionEnd)


leocaseiro:
#### Bug
Toggle a Sidenav before the Transition causes an error on console

#### What is the expected behavior?
It shouldn't show any errors on toggle.

#### What is the current behavior?
It's working the toggle, but we can see errors on console

#### What are the steps to reproduce?

```html
<button md-button class="md-icon-button" aria-label="Open Menu" (click)="menu.toggle()">
    <md-icon>menu</md-icon>
</button>
<md-sidenav #menu [opened]="true" mode="side" layout-padding>
    My Sidenav
</md-sidenav>
```

http://plnkr.co/edit/Ri7CqKVl0b3czY5SYg4g?p=preview

#### Which versions of Angular, Material, OS, browsers are affected?
`@angular@2.2.0`
`@angular/material@2.0.0-alpha.10`
Chrome 54.0.2840.98 (64-bit)
Mac OSX 10.11.6

#### Is there anything else we should know?
Is related to #1382 and #1370.
Something with [_onTransitionEnd()](https://github.com/angular/material2/blob/master/src/lib/sidenav/sidenav.ts#L200)

There's is a temporary fix based on @omararturo [asnwer](https://github.com/angular/material2/issues/1370#issuecomment-256841974):

```javascript
import { Component, OnInit, ViewChild } from '@angular/core';
import { MdSidenav } from '@angular/material';

@Component({
    selector: 'app',
    template: './app.component.html'
})
export class AppComponent implements OnInit {

    @ViewChild('sidenav') sidenav: MdSidenav;

    ngOnInit(): void {
        this.sidenav._onTransitionEnd = function () {
            this._openPromise = null;
            this._closePromise = null;
        }
    }
}
```

christophechevalier:
Thanks ! Very helpfull

PR #1930, opened 2016-11-18
fix(sidenav): resolve promise as false rather than

mmalerba:
rejecting it when the toggle animation is canceled. also clean up the
overly complex promise logic

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -25,6 +25,13 @@ export class MdDuplicatedSidenavError extends MdError {
   }
 }
 
+
+/** Sidenav toggle promise result. */
+export class MdSidenavToggleResult {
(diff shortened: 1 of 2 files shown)


src/lib/sidenav/sidenav.ts
```diff
@@ -134,15 +143,15 @@ export class MdSidenav implements AfterContentInit {
 
   /** Open this sidenav, and return a Promise that will resolve when it's fully opened (or get
    * rejected if it didn't). */
-  open(): Promise<void> {
+  open(): Promise<boolean> {
```
jelbourn: Rather than emitting a `boolean` directly, can we create an event object and include this information? That lets us add additional information to the event in the future.

mmalerba:
Done

mmalerba:
fixes #1899

jelbourn:
LGTM

PR #1933, opened 2016-11-19
fix(icon): remove svgSrc, only allow trusted urls

jelbourn:
_DO NOT MERGE UNTIL DEC 5th_

R: @kara @andrewseguin 
CC: @rjamet

This makes all SVG APIs on `MdIconRegistry` take a `SafeResourceUrl` instead of a `string`. This means that users must *explicitly* trust any URL used to load an SVG icon. This is necessary because we inline the SVGs into the document via `innerHTML`.

This also completely removes `svgSrc` from `<md-icon>` because it is not ergonomic to use with trusted URLs. We can explore adding it back in the future if we add support for a compile-time SVG cache.

### DIFF
--- a/src/demo-app/icon/icon-demo.html
+++ b/src/demo-app/icon/icon-demo.html
@@ -4,11 +4,6 @@
   </p>
 
   <p>
-    From URL:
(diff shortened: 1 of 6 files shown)


jelbourn:
@andrewseguin can you review?

PR #1941, opened 2016-11-20
fix: improved accessibility in high contrast mode

### LINKED ISSUE #421: md-fab: High Contrast mode has no focus indicator

### LINKED ISSUE #1769: a11y(buttons): Low contrast on focused buttons
Bug

The buttons should have sufficient color contrast, both in regular mode and in high contrast mode.

Currently, the button focus indication is a tint with 12% opacity, which is barely visible in regular mode and is completely invisible in high contrast mode. This makes it difficult to see whether a button has been focused.

CC @jelbourn

crisbeto:
a11y improvements to the following components in high contrast mode:

* `button` - Adds an outline since only the text was visible. Also inverts the tint since it was hiding the text.
* `card`, `dialog`, `menu`, `snack-bar`, `tooltip` - Adds an outline.
* `select` - Adds an outline and prevents the ripple from hiding the option text.
* `sidenav` - Fixes the backdrop not being opaque.
* `slide-toggle` - Fixes component blending in completely with the background.

Fixes #421.
Fixes #1769.

### DIFF
--- a/src/lib/button/button.scss
+++ b/src/lib/button/button.scss
@@ -97,6 +97,6 @@
 // Add an outline to make it more visible in high contrast mode.
 @include md-high-contrast {
   [md-button], [md-raised-button], [md-icon-button], [md-fab], [md-mini-fab] {
-    border: solid 1px;
+    outline: solid 1px;
   }
 }
--- a/src/lib/button/button.scss
+++ b/src/lib/button/button.scss
@@ -2,6 +2,7 @@
 // TODO(jelbourn): Figure out if anchor hover underline actually happens in any browser.
 
 @import 'button-base';
+@import '../a11y/_a11y';
 
 // TODO(kara): Replace attribute selectors with class selectors when possible
 [md-button], [md-icon-button] {
@@ -79,6 +80,10 @@
   border-radius: inherit;
   pointer-events: none;
   opacity: 0;
+
+  @include md-high-contrast {
+    background-color: rgba(white, 0.5);
+  }
 }
 
 // For round buttons, the ripple container should clip child ripples to a circle.
@@ -89,9 +94,9 @@
   z-index: 1;
 }
 
-// Applies a clearer border for high-contrast mode (a11y)
-@media screen and (-ms-high-contrast: active) {
-  .md-raised-button, .md-fab, .md-mini-fab {
-    border: 1px solid #fff;
+// Add an outline to make it more visible in high contrast mode.
+@include md-high-contrast {
+  [md-button], [md-raised-button], [md-icon-button], [md-fab], [md-mini-fab] {
+    border: solid 1px;
   }
 }
--- a/src/lib/button/button.scss
+++ b/src/lib/button/button.scss
@@ -80,6 +80,13 @@
   border-radius: inherit;
   pointer-events: none;
   opacity: 0;
+
+  @include md-high-contrast {
+    // Note that IE will render this in the same way, no
+    // matter whether the theme is light or dark. This helps
+    // with the readability of focused buttons.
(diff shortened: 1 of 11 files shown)


src/lib/button/button.scss
```diff
@@ -79,6 +80,10 @@
   border-radius: inherit;
   pointer-events: none;
   opacity: 0;
+
+  @include md-high-contrast {
+    background-color: rgba(white, 0.5);
```
jelbourn: Could this be an issue with dark themes (where the text is white)?

(same for other components where you use a white background)

src/lib/button/button.scss
```diff
@@ -89,9 +94,9 @@
...
 }
 
-// Applies a clearer border for high-contrast mode (a11y)
-@media screen and (-ms-high-contrast: active) {
-  .md-raised-button, .md-fab, .md-mini-fab {
-    border: 1px solid #fff;
+// Add an outline to make it more visible in high contrast mode.
+@include md-high-contrast {
+  [md-button], [md-raised-button], [md-icon-button], [md-fab], [md-mini-fab] {
+    border: solid 1px;
```
jelbourn: Why `border` and not `outline`, since the latter won't affect layout?
(here and elsewhere)

src/lib/button/button.scss
crisbeto: It probably would, but I only went for the dark modes in this PR since supporting all modes would add more bloat.

src/lib/button/button.scss
crisbeto: Good point, I hadn't though about it. I'll give it a try and see how it works out. One nice side effect of using the border is that we don't _have_ to specify the color, it inherits from the text color.

src/lib/button/button.scss
jelbourn: If we make these styles part of the theme, they can simply use the theme's background and foreground palettes to draw an appropriate color.

src/lib/button/button.scss
crisbeto: I tried it out and it looks like using `outline` works, however we can't apply a radius to it. This means that the round buttons would end up looking like this:

m5veqaz

jelbourn: I think that is acceptable for high contrast mode.

crisbeto:
Switched over all of the `border` instances to use `outline` (except the slide toggle since it looked weird and the border doesn't affect layout there).

src/lib/button/button.scss
crisbeto: Done, although I had to flip the conditions since the colors are inverted. Might have to re-test this after we have dark themes implemented.

jelbourn:
Slide-toggle colors also be affected by the theme?

Dark themes should already be working; you can see an [example on the sample app](https://material2-app.firebaseapp.com/) (toggle is on the top right).

crisbeto:
I guess, although it seems like IE behaves the same both with the dark and the light theme. It doesn't try to invert colors.

jelbourn:
@crisbeto in that case it seems fine, then

jelbourn:
LGTM, just needs rebase

crisbeto:
I'll rebase it, but it probably makes sense to revert the light/dark distinction for the button overlay.

jelbourn:
That's fine if IE renders it the same either way

crisbeto:
I mean IE renders the light and dark theme the same. In that case we should also use the same overlay color.

jelbourn:
@crisbeto yeah, I meant that if IE treats light and dark the same way then it's fine to go back to how you originally had it.

PR #1949, opened 2016-11-21
feat(overlay): add fullscreen-enabled overlay class

quanterion:
New PR implementing fullscreen mode support https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen
cc: @jelbourn


src/lib/core/overlay/overlay-container.ts
```diff
@@ -20,10 +23,53 @@ export class OverlayContainer {
...
     this._containerElement = container;
   }
 }
+
+/**
+ * The OverlayContainer is the container in which all overlays will load.
+ * It should be provided in the root component to ensure it is properly shared.
+ */
+@Injectable()
+export class FullscreenFriendlyOverlayContainer extends OverlayContainer {
```
jelbourn: This class should go in its own file

src/lib/core/overlay/overlay-container.ts
```diff
@@ -20,10 +23,53 @@ export class OverlayContainer {
...
     let container = document.createElement('div');
     container.classList.add('md-overlay-container');
     document.body.appendChild(container);
     this._containerElement = container;
   }
 }
+
+/**
+ * The OverlayContainer is the container in which all overlays will load.
+ * It should be provided in the root component to ensure it is properly shared.
```
jelbourn: Needs a class description that explains what this class is for specifically.

src/lib/core/overlay/overlay-container.ts
```diff
@@ -20,10 +23,53 @@ export class OverlayContainer {
...
+      document.addEventListener('webkitfullscreenchange', fn);
+    } else if ((document as any).mozFullScreenEnabled) {
+      document.addEventListener('mozfullscreenchange', fn);
+    } else if ((document as any).msFullscreenEnabled) {
+      document.addEventListener('MSFullscreenChange', fn);
+    }
+  }
+
+  private _getFullscreenElement(): Element {
+    return document.fullscreenElement ||
```
jelbourn: Need a comment here like 
```ts
// When the page is put into fullscreen mode, a specific element is specified. 
// Only that element and its children are visible when in fullscreen mode. 
```

quanterion:
@jelbourn, comments addresed, also renamed container to FullscreenOverlayContainer as it sounds less verbose, can revert back to old name if it is not appropriate.

jelbourn:
@quanterion sounds like we'd need to use an e2e test for this behavior, then, since webdriver acts as user interaction. Can you add an e2e test for this?

quanterion:
@jelbourn added e2e tests and rebased on top of the master

quanterion:
@jelbourn rebased on top of the latest sdk changes and tests starts failing and I can't understand why, the errors seems to be unrelated to my changes

quanterion:
@jelbourn tests successfully passed after latest rebase

quanterion:
@jelbourn any chance it can be reviewed?

jelbourn:
@quanterion working on getting through the backlog of PRs in the wake of beta and holidays now

src/demo-app/demo-app/demo-app.ts
```diff
@@ -52,4 +52,21 @@ export class DemoApp {
     {name: 'Tooltip', route: 'tooltip'},
     {name: 'Platform', route: 'platform'}
   ];
+
+  constructor(private _element: ElementRef) {
+
+  }
+
+  public toggleFullscreen() {
```
jelbourn: We typically omit the `public` access modifier since it's the default

src/lib/core/overlay/fullscreen-overlay-container.ts
```diff
@@ -0,0 +1,73 @@
...
+        (document as any).msFullscreenElement ||
+        null;
+  }
+
+  /**
+   * Toggle element into fullscreen mode
+   * returns true if it has tried to toggle fullscreen mode
+   * but provides no guarantees whether it really happened
+  */
+  toggleFullscreen(element: HTMLElement) {
```
jelbourn: I'd prefer to omit this function from here, since it falls outside of the responsibility of the `OverlayContainer`

jelbourn:
@quanterion mostly looks good, just a few last comments

quanterion:
@jelbourn comments addressed and rebased

jelbourn:
LGTM, thanks!

Issue #1625, opened 2016-10-27
Improve focus-trapping robustness


jelbourn:
The current `<focus-trap>` directive uses a very simple (naive) approach to focus trapping.
- It assumes that the tab order is the same as DOM order, which is not necessarily true.
- Things like tabIndex > 0, flex `order`, and shadow roots can cause to two to misalign.

This behavior needs to be improved, using [ally.js](https://allyjs.io) as a guide.

This is somewhat complicated by the need to detect different browsers / platforms, for which behavior varies subtly.

jelbourn:
The work I had in mind for this issue is done

PR #1950, opened 2016-11-21
fix(interactivity-checker): improve robustness of isTabbable

DevVersion:
A couple of notes: 

* It's intentional to *not* nest things like [that](https://github.com/DevVersion/material2/blob/c12ef840f39e678759f19b76cd059c0e75080e81/src/lib/core/a11y/interactivity-checker.ts#L71-L79), because such functions need to be clear.

* You may have noticed that there are no specs for the Platform. It's just hard to mock the `userAgent` and stuff like `v8BreakIterator` without polluting the other tests.

References #1625

### DIFF
--- a/src/demo-app/platform/platform-demo.ts
+++ b/src/demo-app/platform/platform-demo.ts
@@ -1,0 +1,20 @@
+import {Component} from '@angular/core';
+import {MdPlatform} from '@angular/material';
+
+@Component({
+  template: `
+    <p>Is Android: {{ platform.ANDROID }}</p>
+    <p>Is iOS: {{ platform.IOS }}</p>
+    <p>Is Firefox: {{ platform.FIREFOX }}</p>
+    <p>Is Blink: {{ platform.BLINK }}</p>
+    <p>Is Webkit: {{ platform.WEBKIT }}</p>
+    <p>Is Trident: {{ platform.TRIDENT }}</p>
+    <p>Is Edge: {{ platform.EDGE }}</p>
+
+  `
+})
+export class PlatformDemo {
+
+  constructor(public platform: MdPlatform) {}
+
+}
--- a/src/demo-app/platform/platform-demo.ts
+++ b/src/demo-app/platform/platform-demo.ts
@@ -10,7 +10,6 @@ import {MdPlatform} from '@angular/material';
     <p>Is Webkit: {{ platform.WEBKIT }}</p>
     <p>Is Trident: {{ platform.TRIDENT }}</p>
     <p>Is Edge: {{ platform.EDGE }}</p>
-
   `
 })
 export class PlatformDemo {
--- a/src/demo-app/platform/platform-demo.ts
+++ b/src/demo-app/platform/platform-demo.ts
@@ -14,7 +14,5 @@ import {MdPlatform} from '@angular/material';
   `
 })
 export class PlatformDemo {
-
   constructor(public platform: MdPlatform) {}
-
 }
--- a/src/lib/core/a11y/interactivity-checker.spec.ts
+++ b/src/lib/core/a11y/interactivity-checker.spec.ts
@@ -1,14 +1,17 @@
 import {InteractivityChecker} from './interactivity-checker';
+import {MdPlatform} from '../platform/platform';
+import {async} from '@angular/core/testing';
 
 describe('InteractivityChecker', () => {
   let testContainerElement: HTMLElement;
   let checker: InteractivityChecker;
+  let platform: MdPlatform = new MdPlatform();
 
   beforeEach(() => {
     testContainerElement = document.createElement('div');
     document.body.appendChild(testContainerElement);
 
-    checker = new InteractivityChecker();
+    checker = new InteractivityChecker(platform);
   });
 
   afterEach(() => {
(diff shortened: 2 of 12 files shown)


DevVersion:
@jelbourn This should be ready for review now.

jelbourn:
@DevVersion can you rebase?

DevVersion:
@jelbourn Done.

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -16,6 +17,8 @@ import {Injectable} from '@angular/core';
 @Injectable()
 export class InteractivityChecker {
 
+  constructor(private platform: MdPlatform) {}
```
jelbourn: `_platform`

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -30,26 +33,93 @@ export class InteractivityChecker {
...
+    if (!isControlAudio && !checkRectangles()) {
       return false;
     }
 
-    // Check for css `visibility` property.
-    // TODO(jelbourn): do any browsers we support return an empty string instead of 'visible'?
-    return getComputedStyle(element).getPropertyValue('visibility') == 'visible';
+    return getComputedStyle(element).getPropertyValue('visibility') === 'visible';
+
+    function checkRectangles() {
```
jelbourn: This should be a module method with a description, similar to the others at the bottom of this file. I'd also call it something like `hasGeometry` or `elementHasRenderedRect`

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -30,26 +33,93 @@ export class InteractivityChecker {
...
-    // There are additional special cases that this does not capture, but this will work for
-    // the most common cases.
+    let nodeName = element.nodeName.toLowerCase();
+    let isControlAudio = nodeName === 'audio' && (element as HTMLAudioElement).controls;
 
-    // Use logic from jQuery to check for `display: none`.
-    // See https://github.com/jquery/jquery/blob/master/src/css/hiddenVisibleSelectors.js#L12
-    if (!(element.offsetWidth || element.offsetHeight || element.getClientRects().length)) {
+    // In IE11 audio elements with controls have invalid rectangles, but are still visible.
+    if (!isControlAudio && !checkRectangles()) {
```
jelbourn: Does this mean that an audio element with `display: none` or `visibility: hidden` will still be marked as visible?

src/lib/core/a11y/interactivity-checker.ts
jelbourn: Can you also update the class comment to just say 
```ts
/**
 * Utility for checking the interactivity of an element, such as whether is
 * focusable or tabbable.
 */
```

And then add another `//` comment above it that explains how we drew much of this logic from ally.js, omitting checks for platforms and edge-cases we don't support?

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -122,3 +214,7 @@ function isPotentiallyFocusable(element: HTMLElement): boolean {
       hasValidTabIndex(element);
 }
 
+
+function getWindow(node: HTMLElement): Window {
```
jelbourn: Needs function description

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -30,26 +33,93 @@ export class InteractivityChecker {
...
   /**
    * Gets whether an element can be reached via Tab key.
    * Assumes that the element has already been checked with isFocusable.
    */
   isTabbable(element: HTMLElement) {
-    // Again, naive approach that does not capture many special cases and browser quirks.
+
+    let nodeName = element.nodeName.toLowerCase();
+    let frameElement = getWindow(element).frameElement as HTMLElement;
+    let frameType = frameElement && frameElement.nodeName.toLowerCase();
```
jelbourn: Declare `frameType`, `nodeName`, and `tabIndexValue` closer to where they're first used

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -106,6 +177,27 @@ function hasValidTabIndex(element: HTMLElement): boolean {
   return !!(tabIndex && !isNaN(parseInt(tabIndex, 10)));
 }
 
+function getTabIndexValue(element: HTMLElement) {
```
jelbourn: Needs function description that explains what this does beyond reading the `tabIndex` property directly.

src/lib/core/a11y/interactivity-checker.spec.ts
```diff
@@ -234,63 +237,274 @@ describe('InteractivityChecker', () => {
...
+
   /** Appends elements to the testContainerElement. */
   function appendElements(elements: Element[]) {
     for (let e of elements) {
       testContainerElement.appendChild(e);
     }
   }
+
+  function runIf(condition: boolean, runFn: Function): () => void {
+    return function() {
```
jelbourn:
```ts
return () => {
```
?

jelbourn:
@mmalerba could you also take a look at this PR? I've spent so much time looking at the tabbable stuff I think a fresh perspective would be good

src/lib/core/a11y/interactivity-checker.spec.ts
DevVersion: We need to use a normal function because we want to retrieve the `arguments` below.

mmalerba: you can do `(...args) => {`

src/lib/core/a11y/interactivity-checker.ts
DevVersion: Agreed. I really like `hasGeometry`

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -16,6 +17,8 @@ import {Injectable} from '@angular/core';
 @Injectable()
 export class InteractivityChecker {
 
+  constructor(private platform: MdPlatform) {}
+
   /** Gets whether an element is disabled. */
   isDisabled(element: HTMLElement) {
```
mmalerba: can probably improve this too:
```ts
return element.hasAttribute('disabled') || element.form && element.form.hasAttribute('disabled');
```

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -30,26 +33,93 @@ export class InteractivityChecker {
...
+
+    let nodeName = element.nodeName.toLowerCase();
+    let frameElement = getWindow(element).frameElement as HTMLElement;
+    let frameType = frameElement && frameElement.nodeName.toLowerCase();
+    let tabIndexValue = getTabIndexValue(element);
+
+    if (frameElement) {
+
+      // Frame elements inherit their tabindex onto all child elements.
+      if (getTabIndexValue(frameElement) === -1) {
```
mmalerba: so if an element inherits its tabindex through the frame it's definitely disabled? (because you have a few places below where if the user sets tabindex=-1 that doesn't necessarily mean non-tabbable, e.g. blink audio)

src/lib/core/a11y/interactivity-checker.spec.ts
DevVersion: Good idea, I will change it - But anyways what's the advantage of an arrow function here? We don't need any access to the `this` context here

src/lib/core/a11y/interactivity-checker.ts
DevVersion: Would rather address things like that in another PR. We should keep that one as small as possible.

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -30,26 +33,93 @@ export class InteractivityChecker {
...
+      }
+    }
+
+    if (nodeName === 'object' && (this.platform.BLINK || this.platform.WEBKIT)) {
+      // In all Blink and WebKit based browsers <object> elements are never tabbable.
+      return false;
+    }
+
+    // In iOS the browser only considers some specific elements as tabbable.
+    if (this.platform.WEBKIT && this.platform.IOS && !isPotentiallyTabbable(element)) {
```
mmalerba: maybe call the method isPotentiallyTabbableOnIOS

src/lib/core/a11y/interactivity-checker.spec.ts
```diff
@@ -234,63 +237,274 @@ describe('InteractivityChecker', () => {
...
+        expect(checker.isTabbable(iframe)).toBe(false);
+        expect(checker.isTabbable(button)).toBe(true);
+      });
+
+      it('should mark elements which are contentEditable as tabbable', async(() => {
+        let editableEl = createFromTemplate('<div contenteditable="true">', true);
+
+        // Wait one tick, because the browser takes some time to update the tabIndex
+        // according to the contentEditable attribute.
+        setTimeout(() => {
```
DevVersion: Yeah it seems like the browser needs some time to validate the new `tabindex`.

Here is a snippet to test inside of Chrome 54

```js
let div = document.createElement('div'); document.body.appendChild(div); div.contentEditable = true; console.log(div.tabIndex);

// After that you will see `tabindex = -1` and if you now call
console.log(div.tabIndex) // = 0
```

src/lib/core/a11y/interactivity-checker.ts
DevVersion: Yeah if the user explicitly set the `tabindex` attribute on the *frame element* then all child elements are not tabbable.

The checks below for the tabIndex only affect the element and not the frameElement

src/demo-app/platform/platform-demo.ts
```diff
@@ -0,0 +1,20 @@
...
+    <p>Is Blink: {{ platform.BLINK }}</p>
+    <p>Is Webkit: {{ platform.WEBKIT }}</p>
+    <p>Is Trident: {{ platform.TRIDENT }}</p>
+    <p>Is Edge: {{ platform.EDGE }}</p>
+
+  `
+})
+export class PlatformDemo {
+
+  constructor(public platform: MdPlatform) {}
```
DevVersion: I don't think that's a big deal - Is there any specific convention in the Google JavaScript conventions?

src/lib/core/platform/platform.ts
```diff
@@ -0,0 +1,41 @@
...
+  // EdgeHTML and Trident mock Blink specific things and need to excluded from this check.
+  BLINK = !!(window.chrome || hasV8BreakIterator) && !!CSS && !this.EDGE && !this.TRIDENT;
+
+  // Webkit is part of the userAgent in EdgeHTML Blink and Trident, so we need to
+  // ensure that Webkit runs standalone and is not use as another engines base.
+  WEBKIT = /AppleWebKit/i.test(navigator.userAgent) && !this.BLINK && !this.EDGE && !this.TRIDENT;
+
+  /** Browsers and Platform Types */
+  IOS = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
+  FIREFOX = /(firefox|minefield)/i.test(navigator.userAgent);
```
DevVersion: That's because it would be difficult to detect the plain Gecko layout engine, because actually every browser identifies it self as a Gecko-like browser.

Since we only need to check for Firefox explicit, we can just implement the `firefox` check instead of the unstable `gecko` check (http://webaim.org/blog/user-agent-string-history/)

src/lib/core/a11y/interactivity-checker.spec.ts
mmalerba: interesting, can we just change the timeout to 0 then, bumping to the end of the queue with no delay should be enough

src/demo-app/platform/platform-demo.ts
mmalerba: I don't think there's any official convention for it, I just like to keep it consistent throughout the project. Most other files I've looked at seem to not leave space here

src/lib/core/platform/platform.ts
jelbourn: Can you add that explanation as a comment?

src/lib/core/a11y/interactivity-checker.spec.ts
mmalerba: I don't think there's really much advantage to one way or the other, feel free to leave it as is unless Jeremy feels strongly

DevVersion:
@jelbourn @mmalerba Addressed your feedback.

src/lib/core/a11y/interactivity-checker.ts
mmalerba: an `<audio>` with `visibility: hidden` should be fine, but I think it does miss the `display: none` case, maybe special-case `TRIDENT`/`EDGE` since other browsers have the correct behavior anyways

mmalerba:
LGTM

jelbourn:
LGTM

PR #1962, opened 2016-11-22
fix(dialog): avoid subpixel rendering issues and refactor GlobalPositionStrategy

### LINKED ISSUE #932: bug(overlay): overlays are sometimes blurred in Chrome

crisbeto:
Refactors the `GlobalPositionStrategy` to use flexbox for positioning. This avoids the subpixel rendering issues that come with using transforms for centering.

Fixes #932.

### DIFF
--- a/src/lib/core/overlay/_overlay.scss
+++ b/src/lib/core/overlay/_overlay.scss
@@ -8,10 +8,7 @@
 
   // TODO(jelbourn): change from the `md` prefix to something else for everything in the toolkit.
 
-  // The overlay-container is an invisible element which contains all individual overlays.
-  .md-overlay-container {
-    position: fixed;
-
+  .md-overlay-container, .md-global-overlay-wrapper {
     // Disable events from being captured on the overlay container.
     pointer-events: none;
 
@@ -20,7 +17,19 @@
     left: 0;
     height: 100%;
     width: 100%;
+  }
+
+  // The overlay-container is an invisible element which contains all individual overlays.
+  .md-overlay-container {
+    position: fixed;
     z-index: $md-z-index-overlay-container;
+  }
+
+  // Wrapper element for the panes that use the GlobalPositionStrategy.
+  .md-global-overlay-wrapper {
+    display: flex;
+    position: absolute;
+    z-index: $md-z-index-overlay;
   }
 
   // A single overlay pane.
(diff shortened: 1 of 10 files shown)


src/lib/core/overlay/_overlay.scss
```diff
@@ -20,9 +17,21 @@
...
     width: 100%;
+  }
+
+  // The overlay-container is an invisible element which contains all individual overlays.
+  .md-overlay-container {
+    position: fixed;
     z-index: $md-z-index-overlay-container;
   }
 
+  // Wrapper element for the panes that use the GlobalPositionStrategy.
```
jelbourn: Expand this with something like
```ts
// We use an extra wrapper element in order to use make the overlay itself a flex item.
// This makes centering the overlay easy without running into the subpixel rendering
// problems tied to using `transform`.
```

src/lib/core/overlay/position/global-position-strategy.ts
```diff
@@ -1,87 +1,71 @@
...
-    return this;
-  }
+  private _cssPosition: string = 'static';
+  private _topOffset: string = '';
+  private _bottomOffset: string = '';
+  private _leftOffset: string = '';
+  private _rightOffset: string = '';
+  private _alignItems: string = '';
+  private _justifyContent: string = '';
+  private _wrapper: HTMLElement;
```
jelbourn: Add description for `_wrapper` that explains what it's for and that it's lazily created.

src/lib/core/overlay/position/position-strategy.ts
```diff
@@ -3,4 +3,7 @@ export interface PositionStrategy {
 
   /** Updates the position of the overlay element. */
   apply(element: Element): Promise<void>;
+
+  /** Cleans up any DOM modifications made by the position strategy, if necessary. */
+  dispose?(): void;
```
jelbourn: Let's make this required and give `ConnectedPositionStrategy` a no-op `dispose`, then we won't have to check whether it exists before calling it.

crisbeto:
Applied the feedback and rebased.

jelbourn:
LGTM

PR #1965, opened 2016-11-22
feat(dialog): add the ability to close all dialogs

crisbeto:
Adds a `closeAll` method that closes all of the currently-open dialogs.

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -20,7 +20,6 @@ export {MdDialogRef} from './dialog-ref';
 
 
 // TODO(jelbourn): add support for opening with a TemplateRef
-// TODO(jelbourn): add `closeAll` method
 // TODO(jelbourn): dialog content directives (e.g., md-dialog-header)
 // TODO(jelbourn): animations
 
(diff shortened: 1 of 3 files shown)


src/lib/dialog/dialog.ts
```diff
@@ -46,8 +48,32 @@ export class MdDialog {
...
+    this._openDialogs.push(dialogRef);
+
+    dialogRef.afterClosed().subscribe(() => {
+      let index = this._openDialogs.indexOf(dialogRef);
 
-    return this._attachDialogContent(component, dialogContainer, overlayRef);
+      if (index > -1) {
+        this._openDialogs.splice(index, 1);
+      }
+    });
```
jelbourn: This would be a bit cleaner with another method:
```ts
dialogRef.afterClosed().subscribe(() => this._removeOpenDialog(dialogRef));
```

src/lib/dialog/dialog.ts
```diff
@@ -46,8 +48,32 @@ export class MdDialog {
...
+   * Closes all of the currently-open dialogs.
+   */
+  closeAll(): void {
+    // We need to use the reverse `while`, because the array is being modified
+    // as we're going through it.
+    let i = this._openDialogs.length;
+
+    while (i--) {
+      this._openDialogs[i].close();
+    }
```
jelbourn: I think this needs a comment like
```ts
// The `_openDialogs` property isn't updated after close until the rxjs subscription
// runs on the next microtask, so we loop through all of them and call close without
// assuming that they'll be removed from the list instantaneously. 
```

crisbeto:
Addressed the feedback and rebased.

jelbourn:
LGTM

PR #1968, opened 2016-11-23
Initial Loading overlay

Smileydude:
Just wondering if there is an interest for a loading overlay? Can clean up and give proper commit messages? Just thought I'd get the conversation going.


jelbourn:
Thanks for your proposal! However, this isn't something I think belongs in Material itself (versus Material providing the individual pieces that compose it).

PR #1971, opened 2016-11-23
feat(overlay): add a utility for disabling body scroll

### LINKED ISSUE #1662: overlay: Investigate ways of blocking body scroll
We need a way of blocking body scroll when something like a dialog is open. There are a few ways this can be done:

1. An approach, similar to Material 1, where we set `position: fixed` on the `body` and we offset the scroll position. This approach is pretty reliable, but can have unexpected consequences.
2. Using `preventDefault` on touch, keyboard and mousewheel events, as well as saving the scroll position and resetting it in a scroll handler. Mostly works, but can still cause jumping in certain browsers. 
3. Setting `overflow: hidden` on the `body` and adding `padding-left` with the size of the scrollbar to prevent content jumping. May be tricky to determine the scrollbar width. Can also conflict with user styling.

I'm open to more suggestions since none of these solutions are perfect.

Related to #1633.

crisbeto:
Adds a `DisableBodyScroll` injectable that can toggle whether the body is scrollable.

Fixes #1662.

### DIFF
--- a/src/lib/core/overlay/disable-body-scroll.ts
+++ b/src/lib/core/overlay/disable-body-scroll.ts
@@ -1,4 +1,5 @@
 import {Injectable, Optional, SkipSelf} from '@angular/core';
+import {ViewportRuler} from './position/viewport-ruler';
 
 
 /**
@@ -16,18 +17,23 @@ export class DisableBodyScroll {
     return this._isActive;
   }
 
+  constructor(private _viewportRuler: ViewportRuler) {}
+
   /**
    * Disables scrolling if it hasn't been disabled already and if the body is scrollable.
    */
   activate(): void {
-    if (!this.isActive && document.body.scrollHeight > window.innerHeight) {
-      let body = document.body;
+    let body = document.body;
+    let bodyHeight = body.scrollHeight;
+    let viewportHeight = this._viewportRuler.getViewportRect().height;
+
+    if (!this.isActive && bodyHeight > viewportHeight) {
       let html = document.documentElement;
       let initialBodyWidth = body.clientWidth;
 
       this._htmlStyles = html.style.cssText || '';
       this._bodyStyles = body.style.cssText || '';
-      this._previousScrollPosition = window.scrollY || window.pageYOffset || 0;
+      this._previousScrollPosition = this._viewportRuler.getViewportScrollPosition().top;
 
       body.style.position = 'fixed';
       body.style.width = '100%';
@@ -57,7 +63,7 @@ export class DisableBodyScroll {
 }
 
 export function DISABLE_BODY_SCROLL_PROVIDER_FACTORY(parentDispatcher: DisableBodyScroll) {
-  return parentDispatcher || new DisableBodyScroll();
+  return parentDispatcher || new DisableBodyScroll(new ViewportRuler());
 };
 
 export const DISABLE_BODY_SCROLL_PROVIDER = {
--- a/src/lib/core/overlay/disable-body-scroll.ts
+++ b/src/lib/core/overlay/disable-body-scroll.ts
@@ -62,8 +62,9 @@ export class DisableBodyScroll {
   }
 }
 
-export function DISABLE_BODY_SCROLL_PROVIDER_FACTORY(parentDispatcher: DisableBodyScroll) {
-  return parentDispatcher || new DisableBodyScroll(new ViewportRuler());
+export function DISABLE_BODY_SCROLL_PROVIDER_FACTORY(parentDispatcher: DisableBodyScroll,
+                                                     viewportRuler: ViewportRuler) {
+  return parentDispatcher || new DisableBodyScroll(viewportRuler);
 };
 
 export const DISABLE_BODY_SCROLL_PROVIDER = {
--- a/src/lib/core/overlay/disable-body-scroll.ts
+++ b/src/lib/core/overlay/disable-body-scroll.ts
@@ -1,0 +1,67 @@
+import {Injectable, Optional, SkipSelf} from '@angular/core';
+
+
+/**
+ * Utilitity that allows for toggling scrolling of the viewport on/off.
+ */
+@Injectable()
+export class DisableBodyScroll {
+  private _bodyStyles: string = '';
+  private _htmlStyles: string = '';
+  private _previousScrollPosition: number = 0;
+  private _isActive: boolean = false;
+
+  /** Whether scrolling is disabled. */
+  public get isActive(): boolean {
+    return this._isActive;
+  }
+
+  /**
+   * Disables scrolling if it hasn't been disabled already and if the body is scrollable.
+   */
+  activate(): void {
+    if (!this.isActive && document.body.scrollHeight > window.innerHeight) {
+      let body = document.body;
+      let html = document.documentElement;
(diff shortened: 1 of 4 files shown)


crisbeto:
@jelbourn I have a couple of doubts about this:

1. I had to do a workaround with `max-height` for the default case where the `body` has a margin. The workaround can be avoided if we clear the body margin. Should I do that in the core styles?
2. I wasn't sure whether this was the "Angular 2" way of doing the DOM manipulation (e.g. whether it wouldn't conflict with other renderers).

fxck:
So this will only work on body. How about making it into directive instead and making it work on any element user adds it to? Shouldn't be too hard, just a couple of changes, maybe a little different method for disabling the scroll as well.

crisbeto:
This may not be an issue for elements, at least the ones that use an overlay. Otherwise it should be simpler since we can block the `wheel` and keyboard events.

MateuszG:
The idea sounds good for me.
This PR waiting month only for review?

crisbeto:
@jelbourn can you take a look at https://github.com/angular/material2/pull/1971/files#diff-9e063ef46d4c614a351630e38fff09b9R66 specifically? I'm not too sure how to do the DI in the constructor with the new pattern for the `forRoot` providers.

DevVersion:
@crisbeto Ah, you ran exactly into the same issue I did 😄 You need to do it similar as in https://github.com/angular/material2/pull/2538/commits/a275906aad86649a880008be88b56597b48d9d37#diff-a648493d9ddb16f3470a4d9d93419cffR74

fxck:
I think this would conflict with other renderers. You should [wrap it](http://stackoverflow.com/a/37521339/301596), but I'm not really sure how it would really help in other renderers, unless they implemented document's and window's methods.

Is there any other issue blocking this?

crisbeto:
We don't really need to wrap it, it just won't do anything if the `document` or `window` aren't there. There shouldn't be anything else blocking this.

alexw10:
Any update on this and when it will be merged?
Also this should work on all things that use overlays right?

Mainly concerned with dialog right now as it is I am getting the body scrolling happening in the background even though I only want the dialog to be scrollable.

fxck:
@alexw10 this is just a utility, different components will use different approaches (if I remember correctly, three were mentioned, blocking, repositioning.. and dunno), but yes, dialog will most likely utilise this. If it ever gets merged. I have serious concerns if something didn't happen to @jelbourn. I hope it's just well deserved holiday.

dahaupt:
@kara Can you confirm that @jelbourn is fine and "just" on holiday?

kara:
@dahaupt Haha yes, I can confirm that nothing bad has happened to Jeremy.

dahaupt:
@kara Okay fine :) I can sleep better now 😄

alexw10:
Just as a note as well I did run this recently and it didn't seem to work when utilized with dialog at least. I might have been doing something wrong though so just something to check on that is if it is intended to work with dialog :)

crisbeto:
This PR only sets up the infrastructure that allows for scrolling to be disabled, it doesn't integrate it with any of the components.

fxck:
How's the review coming?

jelbourn:
I've been holding off on this one since I'm generally suspicious of the "disable body scroll" concept; I'd like to stick to the philosophy of only modifying the elements that material "owns". However, there could be value in having this as an option for overlays (vs. re-positioning w/ `ScrollDispatcher`) that users explicitly opt into, so I still want to take a look

fxck:
Repositioning sucks though. You won't ever manage to make it smooth, especially on phones, tablets and older pcs. I wish it wasn't implemented on tooltip at all https://github.com/angular/material2/issues/3413 it lags even my new MBP / chrome.

src/lib/core/overlay/disable-body-scroll.ts
```diff
@@ -0,0 +1,74 @@
...
+      let initialBodyWidth = body.clientWidth;
+
+      this._htmlStyles = html.style.cssText || '';
+      this._bodyStyles = body.style.cssText || '';
+      this._previousScrollPosition = this._viewportRuler.getViewportScrollPosition().top;
+
+      body.style.position = 'fixed';
+      body.style.width = '100%';
+      body.style.top = -this._previousScrollPosition + 'px';
+      html.style.overflowY = 'scroll';
```
jelbourn: Always setting `overflow: scroll` could be a problem if the page does not already have a visible scrollbar. I think we'd have to be more intelligent about whether or not a scrollbar is already showing. Ultimately we don't want scrollbars appearing when they weren't there before.

Also, there can potentially be a horizontal scrollbar as well.

src/lib/core/overlay/disable-body-scroll.ts
```diff
@@ -0,0 +1,74 @@
...
+  activate(): void {
+    let body = document.body;
+    let bodyHeight = body.scrollHeight;
+    let viewportHeight = this._viewportRuler.getViewportRect().height;
+
+    if (!this.isActive && bodyHeight > viewportHeight) {
+      let html = document.documentElement;
+      let initialBodyWidth = body.clientWidth;
+
+      this._htmlStyles = html.style.cssText || '';
```
jelbourn: Rather than caching the styles and setting them back, what do you think of just adding a removing a CSS class from the body where all of the rules are `!important`? It might make things simpler.

src/lib/core/overlay/disable-body-scroll.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      body.style.position = 'fixed';
+      body.style.width = '100%';
+      body.style.top = -this._previousScrollPosition + 'px';
+      html.style.overflowY = 'scroll';
+
+      // TODO(crisbeto): this avoids issues if the body has a margin, however it prevents the
+      // body from adapting if the window is resized. check whether it's ok to reset the body
+      // margin in the core styles.
+      body.style.maxWidth = initialBodyWidth + 'px';
```
jelbourn: I'd prefer if we could avoid getting into the reset styles game. Can you go into more details about what the issue is with the default margin?

src/lib/core/overlay/disable-body-scroll.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    if (!this.isActive && bodyHeight > viewportHeight) {
+      let html = document.documentElement;
+      let initialBodyWidth = body.clientWidth;
+
+      this._htmlStyles = html.style.cssText || '';
+      this._bodyStyles = body.style.cssText || '';
+      this._previousScrollPosition = this._viewportRuler.getViewportScrollPosition().top;
+
+      body.style.position = 'fixed';
```
jelbourn: Are you about to talk at all about this approach vs. just setting `overflow: hidden` to the body?

crisbeto: We could do that for the position and the width, but the `top` is dynamic.

src/lib/core/overlay/disable-body-scroll.ts
crisbeto: It has been a while since I wrote this, but IIRC the margin ended up throwing off some of the measurements.

crisbeto:
Closing in favor of #4500.

PR #1981, opened 2016-11-25
fix(overlay-directives): update offsets if they change after overlay creation

kara:
This PR allows the offset to be updated after the overlay is created.

r: @jelbourn


jelbourn:
LGTM

Issue #1959, opened 2016-11-22
Tooltip doesn't work on button with md-mini-fab


skined90:
Tooltip doesn't work on button with md-mini-fab. I'm getting the error:

error_handler.js:54 TypeError: Cannot read property 'originX' of undefined
    at new ConnectionPositionPair (connected-position.js:5)
    at ConnectedPositionStrategy.withFallbackPosition (connected-position-strategy.js:78)
    at new ConnectedPositionStrategy (connected-position-strategy.js:27)
    at OverlayPositionBuilder.connectedTo (overlay-position-builder.js:30)
    at MdTooltip._createOverlay (tooltip.js:116)
    at MdTooltip._createTooltip (tooltip.js:104)
    at MdTooltip.show (tooltip.js:77)
    at _View_MyWatchlistComponent1._handle_mouseenter_25_0 (component.ngfactory.js:654)
    at view.js:375
    at dom_renderer.js:262
    at dom_events.js:30
    at ZoneDelegate.invoke (zone.js:232)
    at Object.onInvoke (ng_zone.js:238)
    at ZoneDelegate.invoke (zone.js:231)
    at Zone.runGuarded (zone.js:128)
    at NgZone.runGuarded (ng_zone.js:133)
    at HTMLButtonElement.outsideHandler (dom_events.js:30)
    at ZoneDelegate.invokeTask (zone.js:265)
    at Zone.runTask (zone.js:154)
    at HTMLButtonElement.ZoneTask.invoke (zone.js:335)

crisbeto:
Could you post a code sample @skined90? It works fine for me:

![a](https://cloud.githubusercontent.com/assets/4450522/20541964/ba02ca58-b0ff-11e6-9910-5d647efb93d3.gif)

skined90:
Sorry, I've made a mistake. I wrote tooltip-position 'top' instead of 'above'

crisbeto:
Thanks for getting back. It may still be a good idea to show a more helpful error message if the position is something unexpected.

skined90:
@crisbeto Can you please how you made tooltip to hide immediately after mouseleave because in my case it stays active for ~1 second.

crisbeto:
I'm pretty sure I copied it off the tooltip demo page. Should be something like this:

```html
<button md-mini-fab [md-tooltip]="'message'"></button>
```

skined90:
@crisbeto Hm. Weird. I did the same and it definitely doesn't work same as yours.

[...]

PR #1986, opened 2016-11-25
fix(tooltip): throw a better error when an invalid position is passed

crisbeto:
* Throws a more informative error message when an invalid position is passed to the tooltip. Previously it would throw something along the lines of `Cannot read property 'originX' of undefined`.
* Moves the various tooltip switch statements to use an enum for easier validation.

Referencing #1959.


crisbeto:
@andrewseguin can you please review this one?

PR #1987, opened 2016-11-25
refactor(icon): remove unused arguments

### LINKED ISSUE #1539: MdIconRegistry._setSvgAttributes takes unnecessary config arg

Bug

`MdIconRegistry._setSvgAttributes` makes use of all the args passed to it. 

`MdIconRegistry._setSvgAttributes` takes a `config` arg but doesn't use it.

[Simply read the code](https://github.com/angular/material2/blob/cfe3e981bd068f7f06be1893d0894aae26904a26/src/lib/icon/icon-registry.ts#L333)

Cleanup

@angular/material 2.0.0-alpha.9-3

No

crisbeto:
As per the discussion #1539, there are a few places in the icon registry where the config is being passed without actually being used. This change removes those instances and does some minor cleanup.

Fixes #1539.

### DIFF
--- a/src/lib/icon/icon-registry.ts
+++ b/src/lib/icon/icon-registry.ts
@@ -15,7 +15,7 @@ import 'rxjs/add/operator/catch';
 /** Exception thrown when attempting to load an icon with a name that cannot be found. */
 export class MdIconNameNotFoundError extends MdError {
   constructor(iconName: string) {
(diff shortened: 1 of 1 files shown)


jelbourn:
FYI, going to avoid merging this until #1933 is in

jelbourn:
LGTM

PR #1990, opened 2016-11-26
feat(sidenav): close via escape key and restore focus to trigger element

crisbeto:
* Adds the ability to close a sidenav by pressing escape.
* Restores focus to the trigger element after a sidenav is closed.

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -137,9 +137,13 @@ export class MdSidenav implements AfterContentInit {
     });
 
     this.onClose.subscribe(() => {
-      if (this._elementFocusedBeforeSidenavWasOpened) {
+      if (this._elementFocusedBeforeSidenavWasOpened instanceof HTMLElement) {
         this._elementFocusedBeforeSidenavWasOpened.focus();
+      } else {
+        this._elementRef.nativeElement.blur();
       }
+
+      this._elementFocusedBeforeSidenavWasOpened = null;
     });
   }
 
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -140,14 +140,9 @@ export class MdSidenav implements AfterContentInit {
    *     If not available we do not hook on transitions.
    */
   constructor(private _elementRef: ElementRef, private _renderer: Renderer) {
-    this.onOpen.subscribe(() => {
-      this._elementFocusedBeforeSidenavWasOpened = document.activeElement as HTMLElement;
-      this._renderer.invokeElementMethod(this._elementRef.nativeElement, 'focus');
-    });
-
     this.onClose.subscribe(() => {
       if (this._elementFocusedBeforeSidenavWasOpened instanceof HTMLElement) {
-        this._elementFocusedBeforeSidenavWasOpened.focus();
+        this._renderer.invokeElementMethod(this._elementFocusedBeforeSidenavWasOpened, 'focus');
       } else {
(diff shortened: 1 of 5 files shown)


jelbourn:
R: @mmalerba

src/lib/sidenav/sidenav.ts
```diff
@@ -38,6 +38,7 @@ export class MdDuplicatedSidenavError extends MdError {
   template: '<ng-content></ng-content>',
   host: {
     '(transitionend)': '_onTransitionEnd($event)',
+    '(keydown.escape)': 'handleEscapeKey()',
```
mmalerba: unfortunately we can't use this fancy syntax, since not all of the browsers we target support the `key` property of `KeyboardEvent`. We have a file of keycodes here: https://github.com/angular/material2/blob/master/src/lib/core/keyboard/keycodes.ts

crisbeto: Good to know, I saw that the dialog was using it so I went with it. I'll switch it back to use the `keyCode`.

Regarding the reasoning in the `keycodes.ts`, I don't know if this is still an issue. There's [an e2e test on the dialog](https://github.com/angular/material2/blob/master/e2e/components/dialog/dialog.e2e.ts#L18) that uses the escape key.

src/lib/sidenav/sidenav.ts
```diff
@@ -110,7 +112,22 @@ export class MdSidenav implements AfterContentInit {
    * @param _elementRef The DOM element reference. Used for transition and width calculation.
    *     If not available we do not hook on transitions.
    */
-  constructor(private _elementRef: ElementRef) {}
+  constructor(private _elementRef: ElementRef) {
+    this.onOpen.subscribe(() => {
+      this._elementFocusedBeforeSidenavWasOpened = document.activeElement as HTMLElement;
```
DzmitryShylovich: > document.activeElement

what about universal?

crisbeto: I don't think Angular has a way of getting this, at the moment. There are other components that use `document.activeElement` as well. 

AFAIK when we start supporting Universal, we can check if the renderer is a browser via:

```
import { isBrowser, isNode } from 'angular2-universal'
```

DzmitryShylovich: for things such as 
```
this._elementRef.nativeElement.focus();
```
you can use renderer 
```
this.renderer.invokeElementMehod(this._elementRef.nativeElement, 'focus');
``` 
Core already have different renderer implementations for browser and server envs.

But `Renderer` doesn't have methods to retrieve dom elements and afaik the recommended approach to use DomAdapter.

crisbeto:
Addressed the feedback and rebased @mmalerba.

src/lib/sidenav/sidenav.ts
```diff
@@ -197,6 +215,15 @@ export class MdSidenav implements AfterContentInit {
   }
 
   /**
+   * Handles the keyboard events.
+   */
+  handleKeydown(event: KeyboardEvent) {
+    if (event.keyCode === ESCAPE) {
+      this.close();
```
mmalerba: probably want to stop propagation (if we have a dialog with a sidenav, esc should close the sidenav but not the dialog)

crisbeto: Done.

mmalerba:
It seems the tests are failing after the sidenav focus trap stuff in #1695 was merged

crisbeto:
Weird that it's only in Mobile Safari. I'll re-run the build and take a look.

crisbeto:
I sorted it out @mmalerba. It was conflicting with the focus trapping.

PR #1991, opened 2016-11-26
chore: remove inactive browserstack wrapper package

DevVersion:
* No longer uses the "inactive" browserstack wrapper package from npm, which was just responsible for creating the readyfile.

* We can create the readyfile ourself and get rid of the npm package.


scripts/browserstack/start_tunnel.sh
```diff
@@ -1,3 +1,71 @@
...
+BROWSER_STACK_ACCESS_KEY=`echo $BROWSER_STACK_ACCESS_KEY | rev`
+
+# Cleanup and create the folder structure for the tunnel connector.
+rm -rf $TUNNEL_DIR $BROWSER_PROVIDER_READY_FILE
+mkdir -p $TUNNEL_DIR
+touch $TUNNEL_LOG
+
+cd $TUNNEL_DIR
+
+# Download the saucelabs connect binaries.
```
jelbourn: "saucelabs" (here and other places in this file)

jelbourn:
@hansl can you take a look at this?

scripts/browserstack/start_tunnel.sh
```diff
@@ -1,3 +1,71 @@
...
+  disown
+
+  # When the tail recognizes the `Ctrl-C` log message the BrowserStack Tunnel is up.
+  {
+    tail -n0 -f $TUNNEL_LOG --pid $TIMER_PID | { sed '/Ctrl/q' && kill -9 $TIMER_PID; };
+  } &> /dev/null
+
+  echo "BrowserStack Tunnel ready"
+
+  touch $BROWSER_PROVIDER_READY_FILE
```
hansl: Why use `killall` instead of output the PID of browserstack in the ready file?

scripts/browserstack/start_tunnel.sh
```diff
@@ -1,3 +1,71 @@
...
+
+  # To be able to exit the tail properly we need to have a sub shell spawned, which is
+  # used to track the state of tail.
+  sleep 120 &
+
+  TIMER_PID=$!
+
+  # Disown the background process, because we don't want to show any messages when killing
+  # the timer.
+  disown
```
hansl: I don't think you need to disown here, since you're `&> /dev/null` below. Right?

scripts/browserstack/start_tunnel.sh
```diff
@@ -1,3 +1,71 @@
...
+
+  # When the tail recognizes the `Ctrl-C` log message the BrowserStack Tunnel is up.
+  {
+    tail -n0 -f $TUNNEL_LOG --pid $TIMER_PID | { sed '/Ctrl/q' && kill -9 $TIMER_PID; };
```
hansl: If it takes more than 2 minutes the `tail` will exit. We should treat that case since it's an error and it would try to connect later on to browser stack but would fail (since the tunnel isn't connected but this script succeeded).

The way to do that would be (my BASH is rusty) to, instead of just `sleep 120 &` do something like 

```bash
  (sleep 120; echo ERROR TIMEOUT > $BROWSER_STACK_ERROR_FILE) &
```
then on the wait tunnel script, if ERROR_FILE exists just exit with a special code that fails the test.

What do you think?

DevVersion: The tail should only run 2 minutes at maximum. If the specific string was found then the `$TIMER_PID` will be killed and the tail exits.

Did you mean something else?

scripts/browserstack/start_tunnel.sh
DevVersion: I changed it to `killall` because we are doing the same with `sauce-connect` and it would be consistent with the `readyfile` API from `sauce-connect`

scripts/browserstack/start_tunnel.sh
DevVersion: Good catch. I made the two files in the docs setup as similar as possible (so we can share a lot)

scripts/browserstack/start_tunnel.sh
DevVersion: The `&> /dev/null` won't have an effect on that, because when using the `kill` command the process always prints a message that the `child-process` terminated.

scripts/browserstack/start_tunnel.sh
hansl: Yes, if the tail runs for longer than 2 minutes because the connection to the tunnel takes longer. The timer will terminate but the process will succeed.

scripts/browserstack/start_tunnel.sh
hansl: I'm not 100% convinced yet that this is right; what if you run these scripts locally on your computer?

DevVersion: I can see what you mean, but I think you won't be able to fix this for saucelabs, because the integrated `--readyfile` option can't write any PID to the file.

scripts/browserstack/start_tunnel.sh
DevVersion: Got it. Yeah I like that idea with the `error file`. Will change that

hansl:
Just fix the timeout issue and you'll get my LGTM. Good work!

DevVersion:
@hansl Done :)

scripts/browserstack/waitfor_tunnel.sh
```diff
@@ -12,6 +13,11 @@ while [ ! -f $BROWSER_PROVIDER_READY_FILE ]; do
     echo
     echo "Timed out after 2 minutes waiting for tunnel ready file"
     exit 5
+  elif [ -f $BROWSER_PROVIDER_ERROR_FILE ]; then
```
hansl: There's a race condition here, since READY_FILE will still be created when ERROR_FILE exists.  Moving this `IF` outside the `while` will ensure we always take care of this case.

hansl:
1 small nit and LGTM! Thanks

DevVersion:
@hansl Good point -Fixed. Thanks again for your great comments!

PR #1994, opened 2016-11-26
feat(progress-circle): add the ability to show a label

### LINKED ISSUE #1488: [progress-circle] show percentage in the middle of the circle

Feature request

Show the value inside the circle

Does not show the percentage value

http://plnkr.co/edit/v0B4uU

http://i.stack.imgur.com/DzNnN.png

Angular 2+
Material 2.0.0-alpha.9

No

crisbeto:
Adds the ability to show text inside of a progress spinner.

Fixes #1488.


blackhawk389:
yes nice to have currently we can do this by css

src/lib/progress-circle/progress-circle.html
```diff
@@ -1,9 +1,13 @@
...
+  center. The center of the circle with remain at the center of the md-progress-circle
   element containing the SVG.
 -->
 <svg viewBox="0 0 100 100"
      preserveAspectRatio="xMidYMid meet">
   <path></path>
+  <text
+    *ngIf="showText && mode === 'determinate'"
+    text-anchor="middle"
+    x="50" y="55">{{ value }}%</text>
```
jelbourn: We would need to figure out how screen readers present this information. Can you check what NVDA does?

crisbeto: Doesn't look like it gets read out at all. Neither this one nor the regular indeterminate one.

jelbourn: What do you mean by the regular indeterminate one? What isn't read out?

crisbeto: Ops, I meant determinate - one with a value and without the text label. As for stuff not being read out, the screen reader doesn't read out anything when hovering over the progress circle and it's not possible to tab to it either.

micronyks:
Not sure but after watching this thread, I only can understand is that you are planning to show `progress label` only in the circle. 

Please excuse me if I'm wrong or misunderstood this thread.

but as  I mentioned in #3307 , I was intended to display some other text as well in the circle. `eg. Loading Content or something else`.


Will next roll out of progress-circle contain above mentioned feature?

Can you please provide your inputs here?

crisbeto:
That shouldn't be an issue to include in this PR, once it gets some feedback from @jelbourn.

micronyks:
Okay then let's wait for his input. Thanks @crisbeto for immediate reply.

jelbourn:
@crisbeto  I'm okay with moving forward on this if you want to rebase (but it's not a priority)

crisbeto:
Done @jelbourn. I switched up the approach to use projection instead of a boolean since it'll be more flexible.

jelbourn:
Code looks good, though I want to do a bit more to figure out if the accessibility it good.

crisbeto:
I suppose it would be down to the consumer since we just project the content. Otherwise we could also add an `aria-labeledby`.

crisbeto:
I tried out the progress circle with NVDA @jelbourn. It doesn't seem like we're handling a11y properly at all: 
1. The progress circle isn't focusable which prevents the screen reader from reading it out at all.
2. When making the progress circle focusable and tabbing to it, the screen reader does read it out, but it only reads "Progress bar, blank, editable". Note that it seems to read it out properly when hovering it.

Should we get this through and address the a11y in general in a separate PR?

jelbourn:
We should get the a11y smoothed out and then revisit this to see what effect it has.

jelbourn:
Closing this since it's mega stale and not a huge priority.

PR #1997, opened 2016-11-27
fix(progress-circle): Fix type for timer

kacepe:
I fixed type error
`Type 'Timer' is not assignable to type 'number'`

Now `_interdeterminateInterval` is type of `NodeJS.Timer`

### DIFF
--- a/src/lib/progress-circle/progress-circle.ts
+++ b/src/lib/progress-circle/progress-circle.ts
@@ -36,7 +36,7 @@ type EasingFn = (currentTime: number, startValue: number,
  * <md-progress-circle> component.
  */
 @Component({
-  moduleId: module.id,
+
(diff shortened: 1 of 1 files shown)


crisbeto:
This won't work in the browser.

kacepe:
It's working for me.

DevVersion:
It will still work in the browser because the types don't matter after compilation anymore.

Anyways I think it's not valid to use a type from the `NodeJS` namespace here.

jelbourn:
@kacepe The correct type is actually number. It sounds like you're using the typings for NodeJS in your project instead of the typings for the browser. You want to use lib.es6.d.ts when building material (since it is only meant to run on a browser platform).

The other changes in this PR will also break our build (we do not use angular-cli).

PR #1998, opened 2016-11-27
fix(tooltip): Fix type for timer

kacepe:
I fixed type error
`Type 'Timer' is not assignable to type 'number'`

Now `_hideTimeoutId` is type of `NodeJS.Timer`


jelbourn:
@kacepe The correct type is actually `number`. It sounds like you're using the typings for NodeJS in your project instead of the typings for the browser. You want to use `lib.es6.d.ts` when building material (since it is only meant to run on a browser platform). 

The other changes in this PR will also break our build (we do not use angular-cli).

PR #1999, opened 2016-11-27
feat(list): add keyboard navigation

### LINKED ISSUE #941: Request: Keyboard navigation in md-list
To be able to use arrow keys to navigate through md-list or md-data-table or md-grid-list

As said in http://webaim.org/techniques/keyboard/ , in the section "Navigate to most elements" : 
- Keyboard focus indicators must be present.
- Navigation order should be logical and intuitive.

In material-angular (1.1.0-rc.5), arrow keys does not select or change focus in md-list. The issue was closed in material-angular as part of ‘Surge Focus on Material 2' efforts. Requesting this feature to be included in material-angular-2.

Use-case 1 : Pharmacy Software, need to navigate through a md-list of prescriptions with up/down. 
screen shot 2016-08-02 at 00 25 02

Use-case 2 : Google drive, you can navigate through files with arrow keys.
screen shot 2016-08-02 at 00 22 59

Use-case 3 : Google mail, you can navigate through emails with arrow keys.

Motivation : Most pharmacists in Quebec are using an old C# pharmacy software which is ugly but fast. If I can tell them that Material-Angular2 is good-looking AND fast, it might be a game-changer. (keyboard-friendly == fast)

crisbeto:
* Adds the ability to switch focus between list items using the up/down keys.
* Splits up the various list-related components into separate files.

Fixes #941.

### DIFF
--- a/e2e/components/list/list.e2e.ts
+++ b/e2e/components/list/list.e2e.ts
@@ -9,4 +9,38 @@ describe('list', () => {
     let container = element(by.css('md-list'));
     expect(container.isElementPresent(by.css('md-list-item'))).toBe(true);
   });
+
+  it('should be tabbable', () => {
+    pressKey(protractor.Key.TAB);
+    expectFocusOn(element(by.css('md-list')));
+  });
+
+  it('should shift focus between the list items', () => {
+    let items = element.all(by.css('md-list-item'));
+
+    pressKey(protractor.Key.TAB);
+    pressKey(protractor.Key.DOWN);
+    expectFocusOn(items.get(0));
+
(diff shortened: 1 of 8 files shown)


jelbourn:
@crisbeto I double-checked the w3c docs for ths, and I think we don't want to make this change. `<md-list>` by itself is a `role="list"` element, which is [meant to be non-interactive](https://www.w3.org/TR/wai-aria/roles#list).

The keyboard interaction would come into play when we implement `<md-selection-list>`

crisbeto:
I see. I wasn't aware that `md-selection-list` was a planned component. Closing in that case.

PR #2002, opened 2016-11-27
Deselectable radio buttons

nkholski:
I needed to be able to deselect md-radio buttons in a project and found kind of a hackish solution for this, so I thought I'd have a go on actually implement it in material2.

I've updated both the lib and the demo. My biggest concern is the lines 447-449 that feels kind of misplaced.

Deselectability of individual radio buttons and all members of groups is set the same way as when they are disabled.

If you agree that this feature should be implemented I'm happy to improve it based on feedback and add tests.

### DIFF
--- a/src/demo-app/radio/radio-demo.html
+++ b/src/demo-app/radio/radio-demo.html
@@ -1,7 +1,7 @@
 <h1>Basic Example</h1>
 <section class="demo-section">
   <md-radio-button name="group1">Option 1</md-radio-button>
-  <md-radio-button name="group1">Option 2</md-radio-button>
+  <md-radio-button name="group1" deselectable="true">Option 2 (deselectable)</md-radio-button>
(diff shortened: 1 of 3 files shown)


nkholski:
I signed the CLA.

nkholski:
I fixed the linting issues locally except line 464:
`let groupValueChanged = this.radioGroup && (this.value != this.radioGroup.value || this.radioGroup.deselectable);`
Don't know if I should break it in two lines, or if I should use a temporary variable for "(this.value != this.radioGroup.value || this.radioGroup.deselectable)".

I also tried to fix the CLA issue with different e-mail addresses and hope that this reply automatically triggers a test.

jelbourn:
@nkholski Thanks for your proposal, but this isn't a feature we want to support as it doesn't have an analog with native radio buttons and could adversely affect the accessibility of the component.

PR #2009, opened 2016-11-28
fix(list-key-manager): prevent the default keyboard actions

crisbeto:
Calls preventDefault when handling keyboard events in the ListKeyManager. This avoids the user scrolling the page when using the arrow keys to navigate.

Relates to #1999.

### DIFF
--- a/src/lib/core/a11y/list-key-manager.ts
+++ b/src/lib/core/a11y/list-key-manager.ts
@@ -54,9 +54,14 @@ export class ListKeyManager {
         this.focusLastItem();
         break;
       case TAB:
+        // Note that we shouldn't prevent the default action on tab.
         this._tabOut.next(null);
-        break;
+        return;
+      default:
+        return;
     }
+
+    event.preventDefault();
   }
 
   /** Focuses the first enabled item in the list. */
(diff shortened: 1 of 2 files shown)


src/lib/core/a11y/list-key-manager.ts
```diff
@@ -42,15 +42,19 @@ export class ListKeyManager {
...
       case UP_ARROW:
+        event.preventDefault();
         this.focusPreviousItem();
         break;
       case HOME:
+        event.preventDefault();
         this.focusFirstItem();
         break;
       case END:
+        event.preventDefault();
```
jelbourn: Instead of calling `preventDefault` in each case, you can do:
```ts
  onKeydown(event: KeyboardEvent): void {
    switch (event.keyCode) {
      case DOWN_ARROW:
        this.focusNextItem();
        break;
      case UP_ARROW:
        this.focusPreviousItem();
        break;
      case HOME:
        this.focusFirstItem();
        break;
      case END:
        this.focusLastItem();
        break;
      case TAB:
        this._tabOut.next(null);
        return;
      default: 
        return;
    }

    event.preventDefault();
  }
```
(see `slider.ts` for an example)

jelbourn:
R: @kara

PR #2010, opened 2016-11-28
chore: run a11y audits on protractor

DevVersion:
* Fixes accessibility issue in grid-list with `role="listitem"`
* Adds protractor plugin to run aXe-core accessibility audits
* Fixes a bunch of a11y issues in the e2e app to make the aXe audits green


src/e2e-app/e2e-app/e2e-app.html
```diff
@@ -1,11 +1,17 @@
...
-<a md-list-item [routerLink]="['checkbox']">Checkbox</a>
-<a md-list-item [routerLink]="['dialog']">Dialog</a>
-<a md-list-item [routerLink]="['grid-list']">Grid list</a>
-<a md-list-item [routerLink]="['icon']">Icon</a>
-<a md-list-item [routerLink]="['list']">List</a>
-<a md-list-item [routerLink]="['menu']">Menu</a>
-<a md-list-item [routerLink]="['radio']">Radios</a>
-<a md-list-item [routerLink]="['tabs']">Tabs</a>
+<!-- To avoid interfering with tests, we are not using the <md-list> element here -->
+<div role="list">
```
jelbourn: This should be a `<md-nav-list>`

tools/axe-protractor/axe-protractor.js
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+ * Protractor Plugin to run axe-core accessibility audits after Angular bootstrapped.
+ */
+
+const AxeBuilder = require('axe-webdriverjs');
+const {buildMessage} = require('./build-message');
+
+/* Resolved violations */
+const violations = {};
```
jelbourn: Add more explanation to the comment about the structure of this object.

tools/axe-protractor/axe-protractor.js
```diff
@@ -0,0 +1,40 @@
...
+
+/** When the page is ready and Angular is bootstrapped. */
+function onPageStable() {
+  AxeBuilder(browser.driver)
+    .configure(this.config || {})
+    .analyze(results => handleResults(this, results));
+}
+
+/** Handles the results of axe-core. */
+function handleResults(context, results) {
```
jelbourn: `handleAxeResults`?
Also expand the comments with some explanation of what we're doing with those results.

tools/axe-protractor/build-message.js
```diff
@@ -0,0 +1,11 @@
+exports.buildMessage = violation => {
```
jelbourn: Add description for what this does.

test/protractor.conf.js
```diff
@@ -9,15 +9,25 @@ require('ts-node').register({
...
   allScriptsTimeout: 120000,
   getPageTimeout: 120000,
   jasmineNodeOpts: {
     defaultTimeoutInterval: 120000,
-  }
+  },
+
+  plugins: [
+    {
+      path: '../tools/axe-protractor/axe-protractor.js',
```
jelbourn: At which point in the e2e test does axe run?

DevVersion: I added a short comment for it.

> Runs the axe-core accessibility checks each time the e2e page changes and Angular is ready.

It basically means that the A11y check runs each time the e2e app changes its page.

Unfortunately we are not that flexible here, because the Protractor API does not provide much hooks (See https://github.com/angular/protractor/blob/master/docs/plugins.md)

tools/axe-protractor/build-message.js
DevVersion: Done

jelbourn:
@DevVersion LGTM, we can iterate on this in the future. 

@juliemr are you open to expanding the protractor hooks for plugins? I was thinking it might be useful to have a hook for when the test is complete.

DevVersion:
@tinayuangao Rebased again 😄

DevVersion:
@jelbourn This should be ready for another review. 

The `md-nav-list` is made the `link` items bigger and the menus displayed differently.
  > Hiding them by default ensures that the links don't interfere with the specs.

jelbourn:
@DevVersion Sounds like the menu tests are somewhat brittle, then...

What if we just made it not a nav-list at all and just used regular anchors?

DevVersion:
@jelbourn Yep. The menu tests can always fail when the browser viewport is too small.

As long as the anchors don't take up too much space this should work.
> I'd stick for now with the `ngIf` approach, because we might add other e2e pages here and we should make the menu tests stable in another PR.

DevVersion:
@jelbourn Just rebased with the new [Github functionality](https://github.com/blog/2293-resolve-simple-merge-conflicts-on-github)
> Unfortunately you cannot just rebase and force-push with that.

dylanb:
Yay!!

PR #2024, opened 2016-11-29
refactor(connected-position-strategy): use top/left instead of transforms for positioning

crisbeto:
Switches the `ConnectedPositionStrategy` to use `top` and `left` for positioning, instead of `translateX` and `translateY`. This avoids having to do extra work to prevent subpixel rendering issues, while freeing up the transforms to be used on animations or for offsets.

### DIFF
--- a/src/lib/core/overlay/position/connected-position-strategy.ts
+++ b/src/lib/core/overlay/position/connected-position-strategy.ts
@@ -1,7 +1,6 @@
 import {PositionStrategy} from './position-strategy';
 import {ElementRef} from '@angular/core';
 import {ViewportRuler} from './viewport-ruler';
-import {applyCssTransform} from '../../style/apply-transform';
 import {
     ConnectionPositionPair,
     OriginConnectionPosition,
@@ -232,18 +231,11 @@ export class ConnectedPositionStrategy implements PositionStrategy {
    * @param overlayPoint
    */
   private _setElementPosition(element: HTMLElement, overlayPoint: Point) {
-    // Round the values to prevent blurry overlays due to subpixel rendering.
-    let x = Math.round(overlayPoint.x);
-    let y = Math.round(overlayPoint.y);
-
-    // TODO(jelbourn): we don't want to always overwrite the transform property here,
-    // because it will need to be used for animations.
-    applyCssTransform(element, `translateX(${x}px) translateY(${y}px)`);
+    element.style.left = overlayPoint.x + 'px';
+    element.style.top = overlayPoint.y + 'px';
   }
 }
 
 
 /** A simple (x, y) coordinate. */
 type Point = {x: number, y: number};
-
-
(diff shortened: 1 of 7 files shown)


src/lib/core/overlay/position/connected-position-strategy.ts
```diff
@@ -22,10 +21,10 @@ export class ConnectedPositionStrategy implements PositionStrategy {
   private _dir = 'ltr';
 
   /** The offset in pixels for the overlay connection point on the x-axis */
-  private _offsetX: number = 0;
+  private _x: number = 0;
```
jelbourn: Why rename these? They are still offset values.

src/lib/core/overlay/position/connected-position-strategy.ts
```diff
@@ -227,18 +226,11 @@ export class ConnectedPositionStrategy implements PositionStrategy {
...
   private _setElementPosition(element: HTMLElement, overlayPoint: Point) {
-    // Round the values to prevent blurry overlays due to subpixel rendering.
-    let x = Math.round(overlayPoint.x);
-    let y = Math.round(overlayPoint.y);
-
-    // TODO(jelbourn): we don't want to always overwrite the transform property here,
-    // because it will need to be used for animations.
-    applyCssTransform(element, `translateX(${x}px) translateY(${y}px)`);
+    element.style.left = overlayPoint.x + 'px';
+    element.style.top = overlayPoint.y + 'px';
```
jelbourn: Should we not still round these, or do browsers simply ignore subpixel top/left?

src/lib/select/select.spec.ts
```diff
@@ -527,6 +527,8 @@ describe('MdSelect', () => {
       fixture.detectChanges();
 
       const overlayPane = overlayContainerElement.children[0] as HTMLElement;
+      overlayPane.style.position = 'absolute';
```
jelbourn: Why is this necessary?

jelbourn:
@kara should also take a look

src/lib/select/select.spec.ts
crisbeto: It's because it doesn't look like we import all of the styles in the unit tests. This means that the overlay is `position: static`. It wasn't an issue with the transform approach since they work on all of the positions.

src/lib/core/overlay/position/connected-position-strategy.ts
crisbeto: I guess they still looked pretty similar to the transforms. I'll put them back.

src/lib/select/select.spec.ts
jelbourn: Can you just add a comment mentioning that?

crisbeto:
Rebased and addressed the feedback.

jelbourn:
@kara will this conflict with your select alignment PR?

crisbeto:
Seems like a bunch of tests are failing on this one now. I'll take a better look into it tomorrow.

crisbeto:
Tests are sorted out now.

PR #2037, opened 2016-11-30
chore(checkbox): fix style lint errors

belev:
r: @jelbourn

### DIFF
--- a/src/lib/checkbox/checkbox.scss
+++ b/src/lib/checkbox/checkbox.scss
@@ -212,8 +212,7 @@ md-checkbox {
   display: inline-block;
   height: $md-checkbox-size;
   line-height: 0;
-  margin: auto;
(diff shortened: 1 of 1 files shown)


jelbourn:
@belev Our build is green; are you using the latest version of `stylelint`?

belev:
@jelbourn I am using the version which is on `master`. When I run `gulp lint` there is the following message: "Cannot parse selector". Maybe that is not an error but some kind of warning, I am not sure.

jelbourn:
Did some poking around: this is a bug in `stylelint` 7.6.0:
https://github.com/stylelint/stylelint/issues/2093

Already fixed on their head, so we can just wait for the next release.

belev:
Ah ok. Thanks for the information! :)

PR #2038, opened 2016-11-30
fix(select): fix initial values with reactive forms

### LINKED ISSUE #1973: Md-Select start value or active value
Feature request

The Md-Select-Component should be able to have a start value, which should remain active even after the page has been reloaded. 

After reloading or switching routes I wanna see it like this:
Image

We can't set a start value for select. We can click a button to set it, but there is no attribute available to bind it directly with a variable.

Look into existing select examples and try to set a startup value. 

Angular Material latest version (from master)

I am not sure if I just don't see the way to do it or if it's already planned for the future.

kara:
Closes #1973

r: @jelbourn

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -244,8 +244,29 @@ describe('MdSelect', () => {
 
     beforeEach(() => {
       fixture = TestBed.createComponent(BasicSelect);
+    });
+
+    it('should take an initial view value with reactive forms', () => {
+      fixture.componentInstance.control = new FormControl('pizza-1');
       fixture.detectChanges();
 
+      const value = fixture.debugElement.query(By.css('.md-select-value'));
+      expect(value.nativeElement.textContent)
+          .toContain('Pizza', `Expected trigger to be populated by the control's initial value.`);
+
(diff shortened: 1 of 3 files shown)


src/lib/select/select.ts
```diff
@@ -175,7 +177,10 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
    * required to integrate with Angular's core forms API.
    */
   writeValue(value: any): void {
-    if (!this.options) { return; }
+    if (!this.options) {
```
jelbourn: Add comment explaining why this is necessary

src/lib/select/select.spec.ts
```diff
@@ -244,8 +244,28 @@ describe('MdSelect', () => {
 
     beforeEach(() => {
       fixture = TestBed.createComponent(BasicSelect);
+    });
+
+    it('should take an initial view value', () => {
```
jelbourn: "...with reactive forms" ?

qdouble:
@kara I'm building from master, although setting the initial value works, there seems to be a bug with the page scrolling down to the md-select when the value is set.

kara:
@qdouble That will be fixed when #2054 gets in.

qdouble:
@kara okay great, thanks for the info

PR #2046, opened 2016-12-01
Add chips keyboard support.

topherfangio:
Add basic focus/keyboard support for chips.

 - Up/down arrows navigate chips.
 - Clicking a chip properly focuses it for subsequent keyboard
   navigation.
 - More demos.

 - [x] Linter passes
 - [x] Tests pass
 - [x] Confirmed AoT compatibility

References #120.

---

### Demo Page Screen Shot

screen shot 2016-11-30 at 10 12 00 am

### DIFF
--- a/src/lib/chips/_chips-theme.scss
+++ b/src/lib/chips/_chips-theme.scss
@@ -13,6 +13,12 @@
   }
 
   .md-chip.selected {
+    // There is no dark theme in the current spec, so this applies to both
+    background-color: #808080;
+
+    // Use a contrast color for a grey very close to the background color
+    color: md-contrast($md-grey, 600);
+
     &.md-primary {
       background-color: md-color($primary, 500);
       color: md-contrast($primary, 500);
--- a/src/lib/chips/chip-list.spec.ts
+++ b/src/lib/chips/chip-list.spec.ts
@@ -1,10 +1,17 @@
 import {async, ComponentFixture, TestBed} from '@angular/core/testing';
-import {Component, DebugElement}  from '@angular/core';
+import {Component, DebugElement, QueryList} from '@angular/core';
 import {By} from '@angular/platform-browser';
-import {MdChipList, MdChipsModule} from './index';
+import {MdChip, MdChipList, MdChipsModule} from './index';
+import {ListKeyManager} from '../core/a11y/list-key-manager';
 
-describe('MdChip', () => {
+describe('MdChipList', () => {
   let fixture: ComponentFixture<any>;
+  let chipListDebugElement: DebugElement;
+  let chipListNativeElement: HTMLElement;
+  let chipListInstance: MdChipList;
+  let testComponent: StaticChipList;
+  let items: QueryList<MdChip>;
+  let manager: ListKeyManager;
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
@@ -15,39 +22,87 @@ describe('MdChip', () => {
     });
 
     TestBed.compileComponents();
+
+    fixture = TestBed.createComponent(StaticChipList);
+    fixture.detectChanges();
+
+    chipListDebugElement = fixture.debugElement.query(By.directive(MdChipList));
+    chipListNativeElement = chipListDebugElement.nativeElement;
+    chipListInstance = chipListDebugElement.componentInstance;
+    testComponent = fixture.debugElement.componentInstance;
   }));
 
   describe('basic behaviors', () => {
-    let chipListDebugElement: DebugElement;
-    let chipListNativeElement: HTMLElement;
-    let chipListInstance: MdChipList;
-    let testComponent: StaticChipList;
-
-    beforeEach(() => {
-      fixture = TestBed.createComponent(StaticChipList);
-      fixture.detectChanges();
-
-      chipListDebugElement = fixture.debugElement.query(By.directive(MdChipList));
-      chipListNativeElement = chipListDebugElement.nativeElement;
-      chipListInstance = chipListDebugElement.componentInstance;
-      testComponent = fixture.debugElement.componentInstance;
-    });
-
     it('adds the `md-chip-list` class', () => {
       expect(chipListNativeElement.classList).toContain('md-chip-list');
     });
   });
+
+  describe('focus behaviors', () => {
+    beforeEach(() => {
+      items = chipListInstance.chips;
+      manager = chipListInstance._keyManager;
+    });
+
+    it('watches for chip focus', () => {
+      let array = items.toArray();
+      let lastIndex = array.length - 1;
+      let lastItem = array[lastIndex];
+
+      lastItem.focus();
+      fixture.detectChanges();
+
+      expect(manager.focusedItemIndex).toBe(lastIndex);
+    });
+
+    describe('on chip destroy', () => {
+      it('focuses the next item', () => {
+        let array = items.toArray();
+        let midItem = array[2];
+
+        // Focus the middle item
+        midItem.focus();
+
+        // Destroy the middle item
+        testComponent.remove = 2;
+        fixture.detectChanges();
+
+        // It focuses the 4th item (now at index 2)
+        expect(manager.focusedItemIndex).toEqual(2);
+      });
+
+      it('focuses the previous item', () => {
+        let array = items.toArray();
+        let lastIndex = array.length - 1;
+        let lastItem = array[lastIndex];
+
+        // Focus the last item
+        lastItem.focus();
+
+        // Destroy the last item
+        testComponent.remove = lastIndex;
+        fixture.detectChanges();
+
+        // It focuses the next-to-last item
+        expect(manager.focusedItemIndex).toEqual(lastIndex - 1);
+      });
+    });
+  });
+
 });
 
 @Component({
   template: `
     <md-chip-list>
-      <md-chip>{{name}} 1</md-chip>
-      <md-chip>{{name}} 2</md-chip>
-      <md-chip>{{name}} 3</md-chip>
+      <div *ngIf="remove != 0"><md-chip>{{name}} 1</md-chip></div>
+      <div *ngIf="remove != 1"><md-chip>{{name}} 2</md-chip></div>
+      <div *ngIf="remove != 2"><md-chip>{{name}} 3</md-chip></div>
+      <div *ngIf="remove != 3"><md-chip>{{name}} 4</md-chip></div>
+      <div *ngIf="remove != 4"><md-chip>{{name}} 5</md-chip></div>
     </md-chip-list>
   `
 })
 class StaticChipList {
   name: 'Test';
+  remove: Number;
 }
--- a/src/lib/chips/chip-list.ts
+++ b/src/lib/chips/chip-list.ts
@@ -1,32 +1,142 @@
 import {
+  AfterContentInit,
   ChangeDetectionStrategy,
   Component,
+  ContentChildren,
   ElementRef,
   ModuleWithProviders,
   NgModule,
+  QueryList,
   ViewEncapsulation
 } from '@angular/core';
 
 import {MdChip} from './chip';
+import {ListKeyManager} from '../core/a11y/list-key-manager';
 
+/**
+ * A material design chips component (named ChipList for it's similarity to the List component).
+ *
+ * Example:
+ *
+ *     <md-chip-list>
+ *       <md-chip>Chip 1<md-chip>
+ *       <md-chip>Chip 2<md-chip>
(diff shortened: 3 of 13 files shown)


crisbeto:
FYI the ListKeyManager fix was already merged in with https://github.com/angular/material2/pull/2009.

jelbourn:
Looks like you have one failure on IE11

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,154 @@
...
+   * Emitted when the chip receives focus.
+   * @type {EventEmitter}
+   */
+  public didfocus = new EventEmitter();
+
+  /**
+   * Emitted when the chip is destroyed.
+   * @type {EventEmitter}
+   */
+  public destroy = new EventEmitter();
```
jelbourn: No need to include the `public` keyword; we always omit it since it's the default

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,154 @@
...
+export class MdBasicChip implements MdFocusable, OnDestroy {
+
+  /* Whether or not the chip is disabled. */
+  protected _disabled: boolean = false;
+
+  /**
+   * Emitted when the chip receives focus.
+   * @type {EventEmitter}
+   */
+  public didfocus = new EventEmitter();
```
jelbourn: `EventEmitter` needs a generic type. If you're never passing any data to it, use `void`. If you're passing data, create a type to serve as the event object.

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,154 @@
...
+  set disabled(value: boolean) {
+    this._disabled = (value === false || value === undefined) ? null : true;
+  }
+
+  /**
+   * A String representation of the current disabled state.
+   *
+   * @returns {String}
+   */
+  get isAriaDisabled(): string {
```
jelbourn: This should be internal (start with an underscore but not use the `private` keyword)

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,154 @@
...
+    return this._disabled;
+  }
+
+  /**
+   * Sets the disabled state of the chip.
+   *
+   * @param value `true` to disable, `false` to enable.
+   */
+  set disabled(value: boolean) {
+    this._disabled = (value === false || value === undefined) ? null : true;
```
jelbourn: Use `coerceBooleanProperty`

Don't need `@param` JsDoc for setters

topherfangio:
@jelbourn I'll look into the failure tonight. In the meantime, can you do a code review so I can make any additional changes you find?

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,154 @@
...
+    this._renderer.invokeElementMethod(this._elementRef.nativeElement, 'focus');
+    this.didfocus.emit({chip: this});
+  }
+
+  /**
+   * Ensures events fire properly upon click.
+   *
+   * @param event
+   */
+  click(event: Event) {
```
jelbourn: This should be something like
```ts
_handleClick() {
  if (!this.disabled) {
    this.focus();
  }
}

focus() {
  this._renderer.invokeElementMethod(this._elementRef.nativeElement, 'focus');
}
```

You shouldn't need the `focus` Output. The native focus event on the chip element should work fine. 
http://plnkr.co/edit/wW5Ux0yc3JlgTZZATGrQ?p=preview

src/lib/chips/chip-list.ts
```diff
@@ -1,38 +1,80 @@
...
   ViewEncapsulation
 } from '@angular/core';
 
-import {MdChip} from './chip';
+import {MdBasicChip, MdChip} from './chip';
+import {ChipListKeyManager} from './chip-list-key-manager';
 
-@Component({
+/**
+ * The Component configuration for the MdChipList component.
```
jelbourn: Include comment for how this should be used.

src/lib/chips/_chips-theme.scss
```diff
@@ -7,12 +7,17 @@
...
+  // TODO: Should this be in chips.scss since it is independent of theme?
   .md-chip {
     background-color: #e0e0e0;
     color: rgba(0, 0, 0, 0.87);
   }
 
   .md-chip.selected {
+    // TODO: Based on spec, this should be #808080, but we can only use md-contrast with a palette
+    background-color: md-color($md-grey, 600);
+    color: md-contrast($md-grey, 600);
```
topherfangio: I can use a custom background color, but is there a good way to do the `md-contrast()` with a standard color instead of the palette color? If not, I guess I can manually calculate it based on the spec.

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,154 @@
...
     // Properties
-    'class': 'md-chip',
     'tabindex': '-1',
-    'role': 'option'
-  }
-})
-export class MdChip {
+    'role': 'option',
+
+    // Attributes
```
topherfangio: Are we following the official ng2 style guide? If so, I'll just update to match and remove the comments.

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,154 @@
...
+    '[attr.disabled]': 'disabled',
+    '[attr.aria-disabled]': 'isAriaDisabled',
+
+    // Events
+    '(click)': 'click($event)',
+  },
+  inputs: ['disabled'],
+  outputs: ['didfocus', 'destroy']
```
topherfangio: Been digging into the code trying to remove this, but since the `ChipListKeyManager` watches for focus events, this seems to be the best way to subscribe. Otherwise, I have to get a reference to the private HTML element and do `addEventListener()`, or setup a custom subscription method...which seems silly since we have the EventEmitter 😄

src/lib/chips/_chips-theme.scss
jelbourn: The contrast colors aren't calculated, it's just whatever a human designer thought looked "right". Where in the spec does it say `#808080`?

src/lib/chips/chip.ts
jelbourn: Where does the style guide reference this?

In general, we mostly overlap with the Angular style guide, but don't officially follow it. There are some conventions described therein that I disagree with, such as having `.component` etc. in file names or naming services with `Service` at the end.

src/lib/chips/_chips-theme.scss
topherfangio: I guess it's never "stated" in the spec; I just used a color picker and pulled it from the spec images (which all appear to be consistent).

src/lib/chips/chip-list-key-manager.ts
```diff
@@ -0,0 +1,89 @@
...
+import {ListKeyManager} from '../core/a11y/list-key-manager';
+import {MdBasicChip} from './chip';
+
+/**
+ * Manages keyboard events for the chip list and its chips. When instantiated
+ * with a QueryList of MdBasicChip (i.e. any chip), it will ensure focus and
+ * keyboard navigation are properly handled.
+ */
+export class ChipListKeyManager extends ListKeyManager {
+  private _subscribed: MdBasicChip[] = [];
```
jelbourn: Needs field description

jelbourn: I also think a `WeakMap` would be better for what you're doing (`WeakSet` would be best, but that's not supported in IE11).

src/lib/core/a11y/list-key-manager.ts
```diff
@@ -32,10 +32,18 @@ export class ListKeyManager {
...
 
-  /** Sets the focus of the list to the item at the index specified. */
-  setFocus(index: number): void {
+  /**
+   * Sets the focus of the list to the item at the index specified.
+   *
+   * @param index The index of the item to be focused.
+   * @param focusElement Whether or not to focus the element as well. Defaults to `true`.
+   */
+  setFocus(index: number, focusElement = true): void {
```
topherfangio: Should this just be `set focusedItemIndex()`?

I REALLY don't like doing `manager.focusedItemIndex = ...`, so I would prefer your suggestion of `updateFocusedItemIndex(index)`, but I was putting it directly below the getter and it felt a bit weird.

Thoughts?

jelbourn: Actually, `focus` should be a no-op if the item is already focused. Why not just let it run again?

topherfangio: I'm not sure it is a no-op. I think it was emitting the events twice.

Also, now that I dug into the code, I'm not sure we *want* to change this. The majority of places assume it does both; my chips is a special case where we don't want to focus it (because it happens on click; so it already has focus).

Separating this forces users to call two functions the majority of the time which feels odd.

jelbourn: Calling this in order to perform the focus should still update the index, I was just proposing we add a new method that only updates the index.

topherfangio:
@jelbourn Requested updates have been made. Let me know if you see anything else!

src/lib/chips/chip-list.ts
```diff
@@ -1,32 +1,149 @@
...
   },
   styleUrls: ['chips.css'],
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class MdChipList {
+export class MdChipList implements AfterContentInit {
+
+  /** Track which chips we're listening to for focus/destruction. */
+  private _subscribed: WeakMap<MdChip, Boolean> = new WeakMap();
```
jelbourn: `boolean` lower case

src/lib/chips/_chips-theme.scss
```diff
@@ -13,6 +13,11 @@
   }
 
   .md-chip.selected {
+    background-color: #808080;
```
jelbourn: Add comment explaining that this is the same color in both light and dark themes.

src/lib/chips/chip-list.ts
```diff
@@ -1,32 +1,149 @@
...
+
+  /** Pass relevant focus events to our key manager. */
+  focus(event: Event) {
+    this._keyManager.focusFirstItem();
+  }
+
+  /** Pass relevant key presses to our key manager. */
+  keydown(event: KeyboardEvent) {
+    this._keyManager.onKeydown(event);
+  }
```
jelbourn: You don't actually need these two events: you can call `_keyManager.focusFirstItem()` and `_keyManager.onKeydown($event)` directly in the host binding.

src/lib/chips/chip-list.ts
```diff
@@ -1,32 +1,149 @@
...
+
+  /**
+   * Iterate through the list of chips and add them to our list of
+   * subscribed chips.
+   *
+   * @param chips The list of chips to be subscribed.
+   */
+  protected subscribeChips(chips: QueryList<MdChip>): void {
+    chips.forEach((chip: MdChip) => {
```
jelbourn: You shouldn't need the type here, it should be inferred. Also can be condensed to
```ts
chips.forEach(chip => this.addChip(chip));
```

src/lib/chips/chip.ts
```diff
@@ -1,17 +1,96 @@
-import { Component, ElementRef, Renderer } from '@angular/core';
+import {
+  // Classes
```
jelbourn: Don't need comments like this in imports.

src/lib/chips/chip-list.spec.ts
```diff
@@ -1,10 +1,17 @@
-import {async, ComponentFixture, TestBed} from '@angular/core/testing';
-import {Component, DebugElement}  from '@angular/core';
-import {By} from '@angular/platform-browser';
-import {MdChipList, MdChipsModule} from './index';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { Component, DebugElement, QueryList } from '@angular/core';
+import { By } from '@angular/platform-browser';
+import { MdChip, MdChipList, MdChipsModule } from './index';
```
jelbourn: No spaces inside `{ }` like this; there's a WebStorm setting for this.

src/lib/chips/chip-list.ts
topherfangio: I'll change the focus, but I do need the onKeydown for the selection/deletion support that is coming soon in the next PR.

src/lib/chips/_chips-theme.scss
topherfangio: I can definitely do that; but do we have any specs for a dark theme? I think it would be nice to change it as the current chips in dark mode do look a bit out of place.

Issue #2025, opened 2016-11-29
Adding a class with e2e test utilities


crisbeto:
Tracking issue for adding a utility class that has methods for some commonly-used scenarios in e2e tests (checking that an element exists, that it has focus etc.).

PR #2051, opened 2016-12-01
test: use proper typings in e2e app

### LINKED ISSUE #1531: Change e2e app to use @types
I tried doing this and ran into a problem where the types in `@types/selenium-webdriver` don't seem to be complete (compared to what we have in the project already). Need to investigate why.

crisbeto:
* Switches to use the typings from @types for the e2e app.
* Adds all of the necessary imports and fixes type issues.

Relates to #2025.
Fixes #1531.


DevVersion:
I think if we switch to `@types` then we should also remove the `typings/` folder here.

- [See the folder here](https://github.com/angular/material2/tree/26eb7ceb1ef8b4b5c872c9355888173104f40b02/e2e/typings)

Also this would fix https://github.com/angular/material2/issues/1531

crisbeto:
Good points. My editor wasn't showing the `typings` directory at all, for some reason.

crisbeto:
There we go. It ended up being a bit more work than expected.

jelbourn:
LGTM

package.json
```diff
@@ -49,8 +49,11 @@
     "@types/merge2": "0.0.28",
     "@types/minimist": "^1.1.28",
     "@types/node": "^6.0.34",
+    "@types/protractor": "^4.0.0",
     "@types/run-sequence": "0.0.27",
     "@types/rx": "^2.5.33",
+    "@types/selenium-webdriver": "^2.53.36",
+    "browserstacktunnel-wrapper": "^2.0.0",
```
DevVersion: The `browserstacktunnel-wrapper` shouldn't be added again. See #1991

crisbeto: Reverted. There was a merge conflict on that exact line.

crisbeto:
@jelbourn could we get this one in since it's getting a little hard to maintain?

jelbourn:
Yeah, sorry; doing this one requires a manual sync into google instead of our usual automatic one so we've been procrastinating it

PR #2052, opened 2016-12-01
create md-input-wrapper to eventually replace md-input

mmalerba:


src/demo-app/baseline/baseline-demo.html
```diff
@@ -10,7 +10,7 @@
     | Text 3 |
     <md-radio-button value="option_3">Radio 3</md-radio-button>
     | Text 4 |
-    <md-input>Label</md-input>
+    <md-input-wrapper><input placeholder="Input"></md-input-wrapper>
```
hansl: 2 things;

1. since we're not remove the input completely (yet), could you leave 1 here and add the new wrapper?
2. could you insert new lines and tabulation? it might affect the baseline.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+let nextUniqueId = 0;
+
+
+export class MdInputWrapperInputElementCountError extends MdError {
+  constructor(count: number) {
+    super(`md-input-wrapper must contain exactly 1 input or textarea element. Found ${count}.`);
+  }
+}
+
+export class MdInputWrapperPlaceholderConflictError extends MdError {
```
hansl: Seems like those are not imported in `input.ts`; I think they should be the same class since they're technically the same error.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+  private _floatingPlaceholder: boolean = true;
+
+  @ContentChild(MdPlaceholder) _placeholderChild: MdPlaceholder;
+
+  @ContentChildren(MdHint) _hintChildren: QueryList<MdHint>;
+
+  @ContentChild(NgModel) _ngModelChild: NgModel;
+
+  /** Whether the `input` or `textarea` is focused. */
+  _focused = false;
```
hansl: I don't think we need that anymore, since the `input` is now part of the content tree.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+  private _getterSetterOnlyInputType = 'text';
+
+  /** The value of the `input` or `textarea`. */
+  private _inputValue = '';
+
+  /** The native `input` or `textarea` element. */
+  private _inputElement: HTMLInputElement | HTMLTextAreaElement;
+
+  /** A `MutationObserver` to observe the `input` or `textarea`. */
+  private _inputObserver = new MutationObserver(mutations => {
```
hansl: This might make web workers freak out. @jelbourn what's the proper way of handling mutations like those on elements from the content tree? Create a service with an Observable (that would be my guess)?

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+
+  /** Whether the input has a placeholder. */
+  _hasPlaceholder(): boolean {
+    return !!this._inputPlaceholder || !!this._placeholderChild;
+  }
+
+  /** Initialize the `input` or `textarea` element. */
+  private _initInputEl() {
+    // Find input or textarea element.
+    let inputEls = this._elementRef.nativeElement.querySelectorAll('input, textarea');
```
hansl: @jelbourn Refresh my memory that we cannot use `@ContentChild()`? I thought we could but might have been mistaken.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+  get _inputPlaceholder(): string { return this._getterSetterOnlyInputPlaceholder; }
+  set _inputPlaceholder(value: string) {
+    this._getterSetterOnlyInputPlaceholder = value;
+    this._validatePlaceholders();
+  }
+  private _getterSetterOnlyInputPlaceholder = '';
+
+  /** The type attribute of the `input` (or "text" if element is a `textarea`). */
+  private get _inputType(): string { return this._getterSetterOnlyInputType; }
+  private set _inputType(value: string) {
```
hansl: Keeping an internal value for the input type is useless; you can just validate it when the value changes.

src/lib/input/input-wrapper.ts
hansl: If we can't use `@ContentChild` we'll need to wrap this whole piece in a service.

src/lib/input/input-wrapper.ts
mmalerba: I don't see the harm in keeping them separate for now, especially since we're just going to delete the md-input version shortly. It would be weird for MdInput to throw MdInputWrapper* errors

src/lib/input/input-wrapper.ts
mmalerba: I need this so I can update the focused state of the underline, sadly the DOM structure prevents me from doing it in pure CSS

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+  private _initInputEl() {
+    // Find input or textarea element.
+    let inputEls = this._elementRef.nativeElement.querySelectorAll('input, textarea');
+    if (inputEls.length != 1) {
+      new MdInputWrapperInputElementCountError(inputEls.length);
+    }
+    this._inputElement = inputEls[0];
+
+    // Install mutation observer and event listeners and subscribe to ngModel changes.
+    this._inputObserver.observe(this._inputElement, {
```
jelbourn: Call `disconnect` in `ngOnDestroy`?

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+  _inputDisabled = false;
+
+  /** The id attribute of the `input` or `textarea`. */
+  _inputId = '';
+
+  /** The required attribute of the `input` or `textarea`. */
+  _inputRequired = false;
+
+  /** Whether the `input` or `textarea` is empty. */
+  get _empty(): boolean { return !this._inputValue && this._inputType !== 'date'; }
```
jelbourn: Also include `datetime`, `time`, `month`, and `week`. We should probably break it out into a method like `isNeverEmptyInputType` or something similar.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+} from '@angular/core';
+import {MdError, coerceBooleanProperty} from '../core';
+import {NgModel} from '@angular/forms';
+
+
+// Invalid input type. Using one of these will throw an MdInputWrapperUnsupportedTypeError.
+const MD_INPUT_INVALID_INPUT_TYPE = [
+  'file',
+  'radio',
+  'checkbox',
```
jelbourn: Also `"button", "submit", "reset", "color", "range", "image", "hidden"`

src/lib/input/input-wrapper.ts
jelbourn: I'm not worried about it until there's better web-worker support in Angular. Pretty much everything md-input does is DOM-based, so trying to run the code on a worker threat will probably end up being slower due to marshaling.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+          this._inputType = this._inputElement.getAttribute('type');
+          break;
+        case 'placeholder':
+          this._inputPlaceholder = this._inputElement.placeholder;
+          break;
+        case 'required':
+          this._inputRequired = this._inputElement.required;
+          break;
+      }
+    }
```
jelbourn: I'd move this function to the class, though, i.e.
```ts
private _inputObserver = new MutationObserver(mutations => this._updateFromInputMutation(mutations));
```

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+      new MdInputWrapperInputElementCountError(inputEls.length);
+    }
+    this._inputElement = inputEls[0];
+
+    // Install mutation observer and event listeners and subscribe to ngModel changes.
+    this._inputObserver.observe(this._inputElement, {
+      attributes: true,
+      attributeFilter: ['disabled', 'id', 'type', 'placeholder', 'required']
+    });
+    for (let eventType in this._inputListeners) {
```
jelbourn: I think you don't need this extra handling around `this._inputListeners`; it's less code overall (and less indirection) to just do
```ts
inputElement.addEventListener('blur',  () => { this._focused = false; });
inputElement.addEventListener('focus', () => { this._focused = true; });
inputElement.addEventListener('input', () => { this._inputValue = this._inputElement.value; });
```

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+
+  /** Whether the input has a placeholder. */
+  _hasPlaceholder(): boolean {
+    return !!this._inputPlaceholder || !!this._placeholderChild;
+  }
+
+  /** Initialize the `input` or `textarea` element. */
+  private _initInputEl() {
```
jelbourn: `_initializeInputElement`
(generally prefer the unabbreviated form)

src/lib/input/input-wrapper.ts
jelbourn: Oh, and `datetime-local`

crisbeto: This won't necessarily be true on browsers that don't support those types, though. E.g. only Chrome supports `type="date"`. That's why I went with the feature test on #1860.

src/lib/input/input-wrapper.ts
mmalerba: It's not really relevant anymore, but I like this pattern when I'm adding and removing a group of listeners at the same time. That way I don't end up with a bunch of `boundXListener` variables just so I can remove them at the end, I can just iterate over the map removing them

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,289 @@
...
+})
+export class MdPlaceholder {}
+
+
+/** The hint directive, used to tag content as hint labels (going under the input). */
+@Directive({
+  selector: 'md-hint',
+  host: {
+    '[class.md-right]': 'align == "end"',
+    '[class.md-hint]': 'true'
```
mmalerba: switched to @HostBinding / @HostListener (since I realized that @HostListener actually does have a way to pass variables).

mmalerba:
addressed comments, PTAL

hansl:
LGTM, I'll let @jelbourn do the rest of the review.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,283 @@
...
+    'datetime',
+    'datetime-local',
+    'month',
+    'time',
+    'week'
+  ].filter(value => {
+    featureTestInput.setAttribute('type', value);
+    return featureTestInput.type === value;
+  });
+})();
```
jelbourn: Move this to a separate `getSupportedInputTypes` function? Maybe under `core/platform`

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,283 @@
...
+    super(`Input type "${type}" isn't supported by md-input-wrapper.`);
+  }
+}
+
+
+export class MdInputWrapperDuplicatedHintError extends MdError {
+  constructor(align: string) {
+    super(`A hint was already declared for 'align="${align}"'.`);
+  }
+}
```
jelbourn: Let's move these errors to their own file (which most of the other components do).

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,283 @@
...
+  set type(value: string) {
+    this._type = value || 'text';
+    this._validateType();
+  }
+  private _type = 'text';
+
+  @Input()
+  value: any;
+
+  @Output() placeholderChange = new EventEmitter<string>();
```
jelbourn: Should this be internal (`_placeholderChange`)? Also needs a description as to what it's for.

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,283 @@
...
+
+  get empty() { return (this.value == null || this.value == '') && !this._isNeverEmpty(); }
+
+  focused = false;
+
+  private get _uid() { return this._cachedUid = this._cachedUid || `md-input-${nextUniqueId++}`; }
+  private _cachedUid: string;
```
jelbourn: Why do you need a separate `_uid` field instead of setting `this.id` if it is unset in `ngOnInit`?

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,283 @@
...
+
+  constructor(private _elementRef: ElementRef, @Optional() private _ngModel: NgModel) {
```
jelbourn: Instead of injection `NgModel`, you want to inject `FormControl` (which should have the same APIs that you're using).

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,283 @@
...
+    // Force setter to be called in case id was not specified.
+    this.id = this.id;
+
+    if (this._ngModel) {
```
jelbourn: I'm not sure you need this; you should be able to inject `NG_VALUE_ACCESSOR` on the input and return the value from that in a readonly getter (I don't quite follow why this directive needs an `@Input` for `value`).

cc @kara in case there's a better way

src/lib/input/input-wrapper.ts
```diff
@@ -0,0 +1,283 @@
...
+      });
+    }
+  }
+
+  ngAfterContentInit() {
+    this.value = this._elementRef.nativeElement.value;
+  }
+
+  /** Focus the input element. */
+  focus() { this._elementRef.nativeElement.focus(); }
```
jelbourn: Might as well go through the renderer for this (https://github.com/angular/material2/pull/2077)

src/lib/input/input-wrapper.ts
mmalerba: I was just trying to be extra defensive. I figured someone could do `[id]="myId"` and `myId` would initially have some value but it could theoretically update at some point in the future, so if it updates to `null` later on this will make sure we still put an id on it.

src/lib/input/input-wrapper.ts
mmalerba: If I use FormControl it seems like the `valueChanges.subscribe` method is not called for an input with `[(ngModel)]="value"`

I'll wait for @kara's input

mmalerba:
Comments addressed

mmalerba:
Addressed comments & rebased

jelbourn:
LGTM, as discussed we'll hold off on merging until after alpha.11

AsherBarak:
Following 20ee360912fb4d7f59b218566b478ecec7a88088, Is a version release to npm expected? Current version (beta 11) works the old way.

jelbourn:
[Soon](https://github.com/angular/material2/projects/6)

OlliL:
How is it supposed to work?

I replaced
```
      <md-input placeholder="Username"
                [(ngModel)]="username" 
                (keyup.enter)="doLogin()"></md-input>
```
by
```
    <md-input-container>
      <input mdInput
             placeholder="Username"
             (keyup.enter)="doLogin()"
             [(ngModel)]="username">
    </md-input-container>
```

And the style changed from:
md-input

to:
input

Is this intended? I'm using "@angular/material": "^2.0.0-beta.1"

abdulhaq-e:
mdInput => md-input

See the examples here: https://material.angular.io/components/component/input

OlliL:
The link states 

> must be marked with the **mdInput** directive.

And also does the README at:
https://github.com/angular/material2/blob/master/src/lib/input/README.md

kara:
@OlliL The README is relevant to with what's on master. You'll want to look at the docs site (linked by @abdulhaq-e) for the documentation of the latest release.

OlliL:
I see.
Having multiple `<input>` elements in one `<md-input-container>` seems not to render nicely. I'll go for multiple `<md-input-container>` elements - one for each `<input>` element for now. This seems to work now fine. Thanks!

PR #2054, opened 2016-12-01
fix(select): only animate placeholder when no selection

kara:
Uses CSS to float the placeholder when the value is set programmatically so no animation plays.  Will only play animation if the panel is opening or closing without a selection.

r: @jelbourn

### DIFF
--- a/src/lib/select/select.html
+++ b/src/lib/select/select.html
@@ -1,5 +1,6 @@
 <div class="md-select-trigger" overlay-origin (click)="toggle()" #origin="overlayOrigin" #trigger>
-  <span class="md-select-placeholder" [@transformPlaceholder]="_getPlaceholderState()"> {{ placeholder }} </span>
+  <span class="md-select-placeholder" [class.md-floating-placeholder]="this.selected"
+   [@transformPlaceholder]="_placeholderState"> {{ placeholder }} </span>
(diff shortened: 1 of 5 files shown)


src/lib/select/select.scss
```diff
@@ -30,11 +30,22 @@ md-select {
 }
 
 .md-select-placeholder {
+  position: relative;
   padding: 0 2px;
   transform-origin: left top;
 
+  &.md-floating-placeholder {
```
jelbourn: Need comment explaining why this is duplicated from the animation code.

kara:
@jelbourn Added the comment!

jelbourn:
LGTM

src/lib/select/select.html
```diff
@@ -1,5 +1,6 @@
 <div class="md-select-trigger" overlay-origin (click)="toggle()" #origin="overlayOrigin" #trigger>
-  <span class="md-select-placeholder" [@transformPlaceholder]="_getPlaceholderState()"> {{ placeholder }} </span>
+  <span class="md-select-placeholder" [class.md-floating-placeholder]="this._selected"
```
tinayuangao: `this._selected` is private. Better to use `this.selected`.

kara: Whoops! You're absolutely right, it should be `selected`.

kara:
@tinayuangao Should be corrected now. Thanks for catching!

PR #2055, opened 2016-12-02
fix(checkbox): Fix native checked is not true when MdCheckbox initial checked value is true

### LINKED ISSUE #2001: When MdCheckbox is initialized to true, the internal native checkbox has checked == false
Bug

When the modal value is true, the input we hide should also have the value of `true` . It gets `true` when we uncheck and check the checkbox manually. 

input element default value is `false`

https://plnkr.co/edit/SS6LBPL3ucXx1njTRGkF?p=preview

Inspect the checkbox and run this in the console : 
`$0.querySelector('input').checked` // assuming chrome dev environment 

Angular : 2.2.1 
Material: ^ 2.0.0-alpha.10 
This was a deal breaker while testing the code. 
There is another issue , not sure worth creating another bug :  
When testing the component using jasmine, the only way to check the checkbox is checked is to look for the class 'md-checkbox-checked' is present. Even this class is not present while testing so there is no way to test if the checkbox is checked when model value is true while rendering. 

I could not get the spec working in plnkr, so created a project with the unit tests : 
https://github.com/pvamshi/md-checkbox-test

tinayuangao:
Closes #2001

R: @jelbourn @kara

### DIFF
--- a/src/lib/checkbox/checkbox.ts
+++ b/src/lib/checkbox/checkbox.ts
@@ -1,4 +1,5 @@
 import {
+    ChangeDetectorRef,
     ChangeDetectionStrategy,
     Component,
     ElementRef,
@@ -151,7 +152,9 @@ export class MdCheckbox implements ControlValueAccessor {
 
   hasFocus: boolean = false;
 
-  constructor(private _renderer: Renderer, private _elementRef: ElementRef) {
+  constructor(private _renderer: Renderer,
+              private _elementRef: ElementRef,
+              private _changeDetectorRef: ChangeDetectorRef,) {
     this.color = 'accent';
   }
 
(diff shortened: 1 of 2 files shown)


src/lib/checkbox/checkbox.ts
```diff
@@ -169,6 +172,7 @@ export class MdCheckbox implements ControlValueAccessor {
       this._checked = checked;
       this._transitionCheckState(
           this._checked ? TransitionCheckState.Checked : TransitionCheckState.Unchecked);
+      this._changeDetectorRef.markForCheck();
```
jelbourn: Unit test for this?

tinayuangao: Added test. PTAL, thanks!

jelbourn:
LGTM

PR #2062, opened 2016-12-02
chore: update sauce connect to latest version

DevVersion:
- Updates the Saucelabs Connector to the latest version (might help reducing the flakiness for Sauce)

> https://wiki.saucelabs.com/display/DOCS/Setting+Up+Sauce+Connect

@jelbourn Notice, that we still can't use the `latest` tag, because it still refers to an older version


jelbourn:
LGTM

PR #2064, opened 2016-12-02
chore: properly exit e2e tasks

DevVersion:
* Instead of manually exiting the process we should just ensure that everything is properly cleaned up and then gulp can take care of the exit codes.

* If we have the `e2e:done` task at the end of the sequence it will be never called when an error occurred before (will cause things like https://travis-ci.org/angular/material2/jobs/180782730)

### DIFF
--- a/tools/gulp/tasks/e2e.ts
+++ b/tools/gulp/tasks/e2e.ts
@@ -7,11 +7,14 @@ import {
   vendorTask, sequenceTask, serverTask
 } from '../task_helpers';
 
+const gulpRunSequence = require('run-sequence');
 
 const appDir = path.join(SOURCE_ROOT, 'e2e-app');
 const outDir = DIST_ROOT;
 const PROTRACTOR_CONFIG_PATH = path.join(PROJECT_ROOT, 'test/protractor.conf.js');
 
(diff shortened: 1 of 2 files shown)


tools/gulp/tasks/e2e.ts
```diff
@@ -7,11 +7,14 @@ import {
...
 } from '../task_helpers';
 
+const gulpRunSequence = require('run-sequence');
 
 const appDir = path.join(SOURCE_ROOT, 'e2e-app');
 const outDir = DIST_ROOT;
 const PROTRACTOR_CONFIG_PATH = path.join(PROJECT_ROOT, 'test/protractor.conf.js');
 
+/** Method to stop a running e2e web server, which may have not exited properly */
+let stopE2eServer: () => void = null;
```
jelbourn: You could do 
```ts
let stopE2eServer = () => {};
```
And then just call `stopE2eServer` directly without `cleanupTasks`

DevVersion: Sounds good. I had the `cleanupTasks` method because I thought that it would be good in the future to just have all cleanups at one place.

jelbourn:
LGTM

PR #2070, opened 2016-12-03
test: add class with e2e test utilities

### LINKED ISSUE #2025: Adding a class with e2e test utilities

crisbeto:
* Adds a class that simplifies some common and hard to remember Protractor expressions.
* Refactors the existing tests to use the new utilities.

Fixes #2025.


crisbeto:
This will need a little more work once #2051 is merged.

crisbeto:
Should be good to go now.

crisbeto:
Can we get this one through @jelbourn? It should make writing e2e tests a little simpler.

e2e/utils.e2e.ts
```diff
@@ -0,0 +1,79 @@
+import {ElementFinder, browser, by, element, ProtractorBy} from 'protractor';
+
+/**
+ * A set of utility functions for writing E2E tests.
+ */
+export class E2EUtils {
```
jelbourn: Couple of things I want to rearrange here:
* I think we should just export / use the standalone functions without them being in a class (similar to `coerceBooleanProperty` or `applyCssTransform` in core). 
* I want to avoid having one big "utils" file and instead divide the functions into more tightly-scoped files. I was thinking we could have something `util/asserts.ts`, `util/geometry.ts`, `util/actions.ts`, and `util/query.ts`

crisbeto:
Split the E2E class into a few smaller files @jelbourn.

PR #2072, opened 2016-12-03
chore: make tunnel filenames consistent

DevVersion:
* Currently the tunnel `setup/wait/stop` files always have different names like `sauce_connect_block` or just `waitfor_tunnel`.


jelbourn:
LGTM

PR #2075, opened 2016-12-04
fix(list): prevent list item wrapper elements from collapsing

### LINKED ISSUE #2012: Unable to set styles on md-list-item
Bug

Adding a style or class to an anchor tag with the md-list-item directive should modify the style of the list item.

The style/class is applied to the anchor tag, but not to the inner div elements created by the directive. The applied style is not visible on the page.

http://plnkr.co/edit/B0LLmc7We95gMAuySZbv?p=preview

Adding style to indicate the currently selected item

All?

crisbeto:
Adds a `display: block` to the list items, in order to prevent them from collapsing. This allows users to set a custom background directly on the list item.

Fixes #2012.


jelbourn:
LGTM

PR #2076, opened 2016-12-04
docs(getting-started): include a section about adding hammerjs to the config types

### LINKED ISSUE #1944: Cannot find name 'HammerInput' and 'HammerManager'
I installed Angular Material 2.0.0-alpha.10 alongside hammerjs@2.0.8 and @types/hammerjs@2.0.33.
Per the instructions, guidelines and suggestions, I imported hammerjs in my main.ts file but I could not get rid of the errors. After several weeks I thought I should import 'hammerjs' in the files containing the errors as well. After importing hammerjs in MdGestureConfig.d.ts all the errors disappeared. I really dont know what to make of this. So I thought I should ask the experts about it

crisbeto:
Adds a section to the docs, mentioning that the user needs to add `hammerjs` to the `tsconfig.json`, in order to avoid some compilation errors.

Fixes #1944.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -60,6 +60,18 @@ Import HammerJS on your app's module.
 import 'hammerjs';
 ```
 
+Finally, you need to add `hammerjs` to the `types` section of your `tsconfig.json` file:
+
+```json
+{
+  "compilerOptions": {
+    "types": [
+      "hammerjs"
+    ]
+  }
+}
+```
+
 ## Configuring SystemJS
 If your project is using SystemJS for module loading, you will need to add `@angular/material` 
 to the SystemJS configuration:


clydin:
This shouldn't be necessary.  See:
https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

clydin:
With the caveat of already having a types setting

crisbeto:
It seems like it is, on our local demo test site we have the `typeRoots` everywhere, but it breaks if I remove the `types` setting.

clydin:
Interesting.  In my CLI-based projects I don't actually need either setting and I have a large amount of types installed.

jelbourn:
LGTM

clydin:
I looked at the demo app.  `tsc` can resolve hammerjs after removing both `typeRoots` and `types`.  (You will however get an error about a duplicate definition for "module"; @types/node defines it).
The issue appears to be the gulp typescript plugin.

clydin:
Confirmed.  Upgrading and integrating the latest gulp typescript plugin fixes the type resolution issue.
However, for material's repo structure and the large amount of gulp related @types, it would be better to leave the `types` setting in this case to restrict the included types to only `hammerjs`.

PR #2077, opened 2016-12-04
chore: switch over focus method calls to use the renderer

crisbeto:
Switches the calls to `focus` directly on the DOM node to use the Renderer instead. This should make it easier to start supporting Angular Universal, if we decide to at some point.


jelbourn:
LGTM

PR #2081, opened 2016-12-05
fix(tooltip): don't show tooltip if message is empty or not present

### LINKED ISSUE #2078: Empty tooltip should not be shown
If tooltip is undefined, null or '' no tooltip should be displayed

Empty tooltip shown

add [md-tooltip] = null to any tag and hover it

@ angular 2.2.4, @ angular/material 2.0.0-alpha.10

belev:
Closes #2078

r: @jelbourn


jelbourn:
R: @andrewseguin

PR #2084, opened 2016-12-05
feat(tabs): paginate tab header when exceeds width

andrewseguin:


src/demo-app/tabs/tabs-demo.scss
```diff
@@ -26,16 +26,33 @@
 tabs-demo md-card {
   width: 160px;
 
-  md-checkbox {
-    display: block;
-    margin-top: 8px;
-  }
+}
+
+.dynamic-tabs-demo {
```
jelbourn: `.demo-dynamic-tabs`

src/lib/tabs/tab-group.spec.ts
```diff
@@ -19,7 +20,10 @@ describe('MdTabGroup', () => {
         AsyncTabsTestApp,
         DisabledTabsTestApp,
         TabGroupWithSimpleApi,
-      ]
+      ],
+      providers: [
+        { provide: Dir, useFactory: () => { return {value: 'ltr'}; }
```
jelbourn: Why not put a real `dir` directive above the tabs?

src/lib/tabs/tab-header.html
```diff
@@ -0,0 +1,23 @@
+<div class="md-tab-label-pagination before md-elevation-z4"
```
jelbourn: What's this `before` class?

src/lib/tabs/tab-header.html
```diff
@@ -0,0 +1,23 @@
...
+     aria-hidden="true"
+     md-ripple [md-ripple-disabled]="_disableScrollBefore"
+     [class.disabled]="_disableScrollBefore"
```
jelbourn: `class.md-tab-header-disabled`?

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+  Renderer, ViewEncapsulation, ContentChildren, Output, EventEmitter, Optional
+} from '@angular/core';
+import {RIGHT_ARROW, LEFT_ARROW, ENTER, Dir, LayoutDirection} from '../core';
+import {MdTabLabelWrapper} from './tab-label-wrapper';
+import {MdInkBar} from './ink-bar';
+import 'rxjs/add/operator/map';
+
+export type ScrollDirection = 'after' | 'before';
+
+@Component({
```
jelbourn: Needs a class description.

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+  QueryList,
+  ElementRef,
+  Renderer, ViewEncapsulation, ContentChildren, Output, EventEmitter, Optional
+} from '@angular/core';
+import {RIGHT_ARROW, LEFT_ARROW, ENTER, Dir, LayoutDirection} from '../core';
+import {MdTabLabelWrapper} from './tab-label-wrapper';
+import {MdInkBar} from './ink-bar';
+import 'rxjs/add/operator/map';
+
+export type ScrollDirection = 'after' | 'before';
```
jelbourn: Add description for what this type means

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-tab-header',
+  templateUrl: 'tab-header.html',
+  styleUrls: ['tab-header.css'],
+  encapsulation: ViewEncapsulation.None,
+  host: {
+    '[class.md-tab-header]': 'true',
+    '[class.scroll]': '_isScrollingEnabled()',
```
jelbourn: `class.md-tab-header-scrolling` ?

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+export type ScrollDirection = 'after' | 'before';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-tab-header',
+  templateUrl: 'tab-header.html',
+  styleUrls: ['tab-header.css'],
+  encapsulation: ViewEncapsulation.None,
+  host: {
+    '[class.md-tab-header]': 'true',
```
jelbourn: You can just do `'class': 'md-tab-header'` and it won't clobber other classes as long as it's static (not `[class]`)

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+    '[class.scroll]': '_isScrollingEnabled()',
+    '[class.ltr]': "_getLayoutDirection() == 'ltr'",
+    '[class.rtl]': "_getLayoutDirection() == 'rtl'",
```
jelbourn: `md-tab-header-rtl`?

Do you need an `ltr` class? Normally that's just the default and it's overridden by the `rtl` style.

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+  private _indexToFocus: number;
+
+  /** The distance in pixels that the tab labels should be translated to the left. */
+  private _scrollDistance = 0;
+
+  /** Whether the header should scroll to the selected index after the view has been checked. */
+  private _scrollToSelectedIndex = false;
+
+  private _disableScrollAfter = true;
+  private _disableScrollBefore = true;
```
jelbourn: Descriptions for these?

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+        this._disableScrollAfter = this.scrollDistance == this._getMaxScrollDistance();
+      });
+    });
+
+    if (this._scrollToSelectedIndex) {
+      if (this._isScrollingEnabled()) {
+        this._scrollToLabel(this._selectedIndex);
+      }
+      this._scrollToSelectedIndex = false;
+    }
```
jelbourn: Can you break this logic into a separate method like `_updateTabScrollPosition` or similar? Then just call the method in ngAfterViewChecked

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-tab-header',
+  templateUrl: 'tab-header.html',
+  styleUrls: ['tab-header.css'],
+  encapsulation: ViewEncapsulation.None,
+  host: {
+    '[class.md-tab-header]': 'true',
+    '[class.scroll]': '_isScrollingEnabled()',
```
jelbourn: This is a very good point. We should recalculate this as infrequently as possible.

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+    if (this._scrollToSelectedIndex) {
+      if (this._isScrollingEnabled()) {
+        this._scrollToLabel(this._selectedIndex);
+      }
+      this._scrollToSelectedIndex = false;
+    }
+  }
+
+  /** Tells the ink-bar to align itself to the current label wrapper */
+  private _updateInkBar(): void {
```
jelbourn: `_alignInkbarToSelectedTab`?

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+        this.selectFocusedIndex.emit(this.focusIndex);
+        break;
+    }
+  }
+
+  /**
+   * Determines if an index is valid.  If the tabs are not ready yet, we assume that the user is
+   * providing a valid index and return true.
+   */
+  _isValidIndex(index: number): boolean {
```
jelbourn: This can be `private`

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  /** Tracks which element has focus; used for keyboard navigation */
+  get focusIndex(): number {
+    return this._focusIndex;
+  }
+
+  /** When the focus index is set, we must manually send focus to the correct label */
+  set focusIndex(value: number) {
+    if (!this._isValidIndex(value) || this._focusIndex == value) { return; }
```
jelbourn: Can you break most or all of this logic outside of the setter directly into a separate method? I generally want to keep setters relatively small.

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+      if (dir == 'ltr') {
+        containerEl.scrollLeft = 0;
+      } else {
+        containerEl.scrollLeft = containerEl.scrollWidth - containerEl.offsetWidth;
+      }
+    }
+  }
+
+  /**
+   * Moves the focus before or after depending on the offset provided.  Valid offsets are 1 and -1.
```
jelbourn: I don't quite follow `"Moves the focus before or after"`

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+  _focusPreviousTab(): void {
+    this._moveFocus(this._getLayoutDirection() == 'ltr' ? -1 : 1);
+  }
+
+  /** The layout direction of the containing app. */
+  _getLayoutDirection(): LayoutDirection {
+    return this._dir && this._dir.value === 'rtl' ? 'rtl' : 'ltr';
+  }
+
+  _updateScrollPosition() {
```
jelbourn: These scrolling methods could use some descriptions

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+    let translateX = this.scrollDistance + 'px';
+    if (this._getLayoutDirection() == 'ltr') {
+      translateX = '-' + translateX;
+    }
+
+    this._renderer.setElementStyle(this._tabList.nativeElement,
+        'transform', `translate3d(${translateX}, 0, 0)`);
```
jelbourn: See the `applyCssTransform` function

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  _scrollHeader(scrollDir: ScrollDirection) {
+    const viewLength = this._tabListContainer.nativeElement.offsetWidth;
+
+    if (scrollDir == 'before') {
+      this.scrollDistance -= viewLength / 3;
+    } else {
+      this.scrollDistance += viewLength / 3;
+    }
```
jelbourn:
```ts
this.scrollDistance = (scrollDir == 'before' ? -1 : 1) * viewLength / 3;
```
?

Also, explain where this calculation comes from?

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+      labelAfterPos = this._tabList.nativeElement.offsetWidth - selectedLabel.getOffsetLeft();
+      labelBeforePos = labelAfterPos - selectedLabel.getOffsetWidth();
+    }
+
+    const beforeVisiblePos = this.scrollDistance;
+    const afterVisiblePos = this.scrollDistance + viewLength;
+
+    // If move is required, overscroll by a small amount to provide an affordance to click the
+    // next label.
+    const exaggeratedOverscroll = 60;
```
jelbourn: Make this a file-level constant?

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,258 @@
...
+    } else {
+      this.scrollDistance += viewLength / 3;
+    }
+  }
+
+  _scrollToLabel(labelIndex: number) {
+    const selectedLabel = this._labelWrappers.toArray()[labelIndex];
+    const viewLength = this._tabListContainer.nativeElement.offsetWidth;
+
+    let labelBeforePos: number, labelAfterPos: number;
```
jelbourn: Comment explaining what these variables contain?

src/lib/tabs/tab-header.ts
andrewseguin: I agree that this is not ideal. Do you have any suggestions on how to reduce the amount of times we call this but still have an up to date pagination? It seems that checking this in ngAfterViewChecked will be a little better but still way more than we should need.

crisbeto: I think that caching the width and only updating it in the following cases should be fine:
1. When initialized.
2. When a tab gets added/removed.
3. When a tab's heading changes.

The third one might be tricky to get right since we'd have to either get the `innerText` from the DOM on every change detection, or use a `MutationObserver` per tab. We could also try to have have one "main" `MutationObserver` on the header, but I don't know whether that won't end up firing too often, because AFAIK, you need to enable a bunch of flags to get it working.

jelbourn: Yeah, I chatted with Andrew yesterday about this and that's pretty much example what we're doing. MutationObserver per-tab, going to do that part in a follow-up PR (and just handle init, adding/removing tabs in this PR).

andrewseguin: Agreed, I think init and tabs add/remove should handle the vast majority of use cases. I'll include just that logic in this PR.

In another PR, I'll attempt to check tab heading changes (text node modifications) using the `MutationObserver`.

I think one more thing to consider is that we could check when a change is made to the selected tab or focused tab. This would provide a fallback that could be triggered by the user who may try to get the header to update the scrolling after some event has caused the width to change.

andrewseguin: Talked with Jeremy - adding a check to select/focus will probably be overkill. But adding a check on resize will probably be nice.

Also, Jeremy had a good idea to make the mutation observer logic as a separate directive that could be placed into core for other components that need similar checks.

crisbeto: I like the idea of the MutationObserver directive, although isn't it basically 10 lines of boilerplate anyway?

jelbourn: I figure it will be a nice convenience if we ever need to do something like this again

andrewseguin:
Shuffled the logic so that the majority of scrolling stuff is evaluated on content check and checks are only made conditionally (e.g. tab list changed, selected index changed, etc)

Really appreciate the reviews, the amount of layout reflows has dropped significantly.

src/lib/tabs/tab-header.ts
```diff
@@ -0,0 +1,357 @@
...
+    if (this._scrollDistanceChanged) {
+      this._updateTabScrollPosition();
+      this._scrollDistanceChanged = false;
+    }
+  }
+
+  /**
+   * Waits one frame for the view to update, then updates the ink bar and scroll.
+   * Note: This must be run outside of the zone or it will create an infinite change detection loop
+   * TODO: internal
```
jelbourn: No need for `TODO: internal` any more

andrewseguin:
Removed the internal comment.

I'd like to add a reference drawing into the folder to help people understand better about how the element container widths/offsets work. Here's the WIP: https://docs.google.com/drawings/d/1fhtiM-NabWhKP8pVsz8hCAQkzWz1QXrUG0LH-xQdpqI/edit?usp=sharing

mmalerba:
Property {'_disableScrollBefore', '_disableScrollAfter', '_handleKeydown'} is private and only accessible within class 'MdTabHeader'

andrewseguin:
Okay, should be good to merge!

PR #2090, opened 2016-12-06
feat(dialog): add dialog content elements

### LINKED ISSUE #1624: Add dialog content elements
This will look something like

``` html
<!-- user-profile-dialog.html -->
<h3 md-dialog-title>
  User Profile
  <button md-dialog-close></button>
</h3>
<md-dialog-content>
  ...
</md-dialog-content>
<md-dialog-actions>
  <button md-button>SAVE</button>
</md-dialog-actions>
```

### LINKED ISSUE #2042: MdDialog Content Overflow
When the content of a MdDialog overflows the height of the browser, the content is hidden.

Scroll the content of MdDialog

The content overflowed is hidden.

http://plnkr.co/edit/9S2ZUzQTmquS4sjmjuoO?p=preview

Last/Last Material

crisbeto:
Adds the following dialog-specific directives:
* `md-dialog-close` - Closes the current dialog.
* `md-dialog-title` - Title of a dialog.
* `md-dialog-content` - Scrollable content for a dialog.
* `md-dialog-actions` - Container for the bottom buttons in a dialog.

Fixes #1624.
Fixes #2042.


pietie:
Out of interest, why is the md-dialog-title selector just an attribute? 
If I compare it with md-card I would have expected it to be ```<md-dialog-title>```

crisbeto:
It's there to encourage you to use the appropriate native elements @pietie (e.g. `h1`, `h2` etc).

clydin:
Have you considered an "align" attribute on md-dialog-actions similar to md-card? "center" would also be a nice addition to start|end as well.

crisbeto:
Nope, I followed the example from #1624, but it sounds reasonable.

src/lib/dialog/dialog-directives.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(public dialog: MdDialog) { }
+}
+
+/**
+ * Title of a dialog element. Stays fixed to the top of the dialog when scrolling.
+ */
+@Directive({
+  selector: '[md-dialog-title]',
+  host: {
+    role: 'heading'
```
jelbourn: I don't think we should put a `role` on this- the user should just put it on the appropriate element (`h2`, `header`, etc.)

src/lib/dialog/dialog-directives.ts
```diff
@@ -0,0 +1,49 @@
...
+export class MdDialogTitle { }
+
+
+/**
+ * Scrollable content container of a dialog.
+ */
+@Directive({
+  selector: '[md-dialog-content], md-dialog-content',
+  host: {
+    role: 'main'
```
jelbourn: I don't think we should have this role; nothing in [the W3C authoring practices](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) mentiones it and the docs for [role="main"](https://www.w3.org/TR/wai-aria/roles#main) says there should only be one per "document" (for which a modal dialog doesn't count).

src/lib/dialog/dialog.scss
```diff
@@ -0,0 +1,52 @@
...
+md-dialog-content, [md-dialog-content] {
+  display: block;
+  margin: 0 $md-dialog-padding * -1;
+  padding: 0 $md-dialog-padding;
+  max-height: $md-dialog-max-height;
+  overflow: auto;
+}
+
+[md-dialog-title] {
+  font-size: rem(2);
```
jelbourn: I'd like to avoid using this `rem` function for now... the fact that it's even here is a leftover from an earlier version. During beta we're going to work on a more comprehensive typography solution.

src/lib/dialog/dialog.scss
```diff
@@ -0,0 +1,52 @@
...
+  margin: 0 0 rem(2);
+  display: block;
+}
+
+md-dialog-actions, [md-dialog-actions] {
+  padding: $md-dialog-padding / 2 0;
+  display: block;
+
+  &:last-child {
+    margin-bottom: -$md-dialog-padding;
```
jelbourn: Add a comment explaining what this style is for?

src/lib/dialog/dialog-directives.ts
```diff
@@ -0,0 +1,49 @@
+import {Directive} from '@angular/core';
+import {MdDialog} from './dialog';
+
+
+/**
+ * Button that will close the current dialog.
+ */
+@Directive({
+  selector: 'button[md-dialog-close]',
```
jelbourn: Also add `mat-` prefixed selectors (will be used for material1 compatibility mode).

src/lib/dialog/dialog.scss
```diff
@@ -0,0 +1,52 @@
+@import '../core/style/elevation';
+@import '../a11y/_a11y';
+
+
+$md-dialog-padding: 24px !default;
+$md-dialog-border-radius: 2px !default;
+$md-dialog-max-width: 80vw !default;
+$md-dialog-max-height: 65vh !default;
```
jelbourn: Where do these max-height and max-width come from?

src/lib/dialog/dialog.scss
```diff
@@ -0,0 +1,52 @@
...
+  margin: 0 $md-dialog-padding * -1;
+  padding: 0 $md-dialog-padding;
+  max-height: $md-dialog-max-height;
+  overflow: auto;
+}
+
+[md-dialog-title] {
+  font-size: rem(2);
+  font-weight: bold;
+  letter-spacing: 0.005em;
```
jelbourn: Why is `letter-spacing` necessary here?

src/lib/dialog/dialog-directives.ts
```diff
@@ -0,0 +1,49 @@
...
+  host: {
+    '(click)': 'dialog.closeTop()'
+  }
```
jelbourn: We need to add a default `aria-label` to this that says "Close dialog". Also a unit test that ensures an `aria-label` set by the user overrides the default.

src/lib/dialog/dialog-directives.ts
```diff
@@ -0,0 +1,49 @@
+import {Directive} from '@angular/core';
```
jelbourn: Let's call this file `dialog-content-directives.ts`

src/lib/dialog/dialog-directives.ts
```diff
@@ -0,0 +1,49 @@
...
+import {MdDialog} from './dialog';
+
+
+/**
+ * Button that will close the current dialog.
+ */
+@Directive({
+  selector: 'button[md-dialog-close]',
+  host: {
+    '(click)': 'dialog.closeTop()'
```
jelbourn: I don't think this will work for situations where there are two dialogs open side-by-side instead of one on top of the other. I took a poke at the code and found that you can inject the `MdDialogRef` here because the content is loaded with an instance of `DialogInjector`, so this works:
```ts
/**
 * Button that will close the current dialog.
 */
@Directive({
  selector: 'button[md-dialog-close]',
  host: {
    '(click)': 'dialogRef.close()'
  }
})
export class MdDialogClose {
  constructor(public dialogRef: MdDialogRef<any>) { }
}
```

With this change, I don't think we need `closeTop` any more.

clydin: Was going to say pretty much the same thing.  As is it will break some more exotic use cases.

crisbeto: I went this route since it's not currently possible to have multiple dialogs side-by-side. Also I wasn't aware of the `DialogInjector`. I'll try it out.

src/lib/dialog/dialog-directives.ts
crisbeto: I added it in case the attribute was added to something like a `div`. I'll scope the selector to the proper nodes.

src/lib/dialog/dialog-directives.ts
crisbeto: Shouldn't the content have some kind of role? The `main` was the only one I could find that seemed to fit.

src/lib/dialog/dialog.scss
crisbeto: The `max-width` is the same that Material 1 used, the `max-height` is arbitrary. M1 has 80% for the `max-height` as well, although that's for the entire dialog, whereas here we only have it on the `md-dialog-content`.

src/lib/dialog/dialog.scss
crisbeto: What should I go with in this case? Put in the computed value?

src/lib/dialog/dialog.scss
crisbeto: I went with whatever M1 was doing, I'm not sure what the reasoning for it was.

src/lib/dialog/dialog-directives.ts
jelbourn: I don't think it needs a role at all

src/lib/dialog/dialog.scss
jelbourn: Do we need to set a max-width and max-height at all, then? I could imagine some people wanting to do a full-screen dialog.

src/lib/dialog/dialog.scss
jelbourn: Yeah, go ahead with the computed value for now.

src/lib/dialog/dialog.scss
jelbourn: Let's remove it, then.

crisbeto:

jelbourn:
LGTM, just needs rebase. Go ahead and apply the `merge_ready` label when it's rebased

chpasha:
is it possible to override md-dialog-max-width somehow? I'm unable to change it to anything else in spite of !default attribute. The space is precious especially in xs-sm, hard coded 80vw seems to be too small

chpasha:
one more thing - material design guidelines for dialogs (except for alertdialog) have an toolbar at the top, did you consider the possibility to give an option of aligning md-dialog-actions to the top or introducing one more directive like md-dialog-toolbar? It would be also awesome if that toolbar could have home button, title and action areas

src/lib/dialog/dialog.scss
chpasha: actually yes, there are fullscreen dialogs in material guidelines https://material.io/guidelines/components/dialogs.html#dialogs-confirmation-dialogs

crisbeto:
@chpasha 
1. How are you trying to override the max width? If overriding the SASS variable doesn't work, you should still be able to set this somewhere in your stylesheet:

```css
md-dialog-container {
  max-width: <some value>;
}
```

2. The point of this PR was to get some of the dialog content elements as a base. We'll probably keep improving on them in the future.

chpasha:
1. yes, selector in css works but I was hoping for more elegant solution via scss. Is there any point for defining those properties as !default since they cannot be overwritten in client library? 
2. :+1:  fullscreen-style dialogs with toolbar at the top and basic elements like home button, title, content and action buttons would be cool

chpasha:
yes, I did succeed http://plnkr.co/edit/B46euKJxtf8RXu1O0smZ?p=preview
two changes
1) encapsulation: ViewEncapsulation.None on Component
2) md-dialog-container (not .md-dialog-container) in styles - and don't forget that it wont grow bigger if you content is small - to see that it works, I added width: 100vw

PR #2092, opened 2016-12-06
chore(autocomplete): add autocomplete scaffold

kara:


src/lib/autocomplete/autocomplete.ts
```diff
@@ -0,0 +1,11 @@
+import {Component, ViewEncapsulation} from '@angular/core';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-autocomplete',
```
jelbourn: Also add `mat-autocomplete` and add the appropriate selectors under `core/compatibility`

kara:
@jelbourn Good catch! Fixed.

PR #2093, opened 2016-12-07
chore: make compatibility mode work for apps with multiple modules

jelbourn:
R: @kara 

The way I previously had the compatibility mode set up, it wouldn't have worked for applications that are composed of multiple NgModules because the sub-modules wouldn't have included the `MdPrefixEnforcer` directive.

For this new way, you're always going to have _one_ of the prefix-enforcer directives loaded, but not the other.


jelbourn:
@kara rebased and updated
@andrewseguin can you also take a look at this?

src/lib/core/compatibility/compatibility.ts
```diff
@@ -0,0 +1,178 @@
...
+  md-slider,
+  md-spinner,
+  md-tab,
+  md-tab-group,
+  md-toolbar,
+  mdTooltip`;
+
+/** Directive that enforces that the `mat-` prefix cannot be used. */
+@Directive({selector: MAT_ELEMENTS_SELECTOR})
+export class MatPrefixEnforcer {
```
kara: Nit: From the name, I originally thought that this would enforce that you *only* had mat- prefixes, rather than you *can't* have mat- prefixes (as in, you're enforcing that they're present). I think it's the word "Enforcer".  "MatPrefixRejector"?

jelbourn: Renamed

kara: Cool, LGTM (although it was LGTM before too)

mmalerba:
please rebase

mmalerba:
whoops, tried using the github button to resolve but that was apparently a bad idea, now it complains about merge conflicts when I try to g3sync presubmit.

@jelbourn is this something that can be fixed or do i just have to manually sync it now?

PR #2100, opened 2016-12-07
chore: mark non-public APIs as @internal

jelbourn:
R: @mmalerba 

Disables the `stripInternal` TypeScript option (which omits `@internal` symbols from the `.d.ts` files) and uses the `@internal` JsDoc to mark classes as non-public. This will tell our docs generation that this isn't a public API that needs to be documented.

This is mainly for symbols where we don;t want to start them with an underscore (such as for class names) or for symbols that already have an underscored version (rare in our codebase).

### DIFF
--- a/src/demo-app/tsconfig.json
+++ b/src/demo-app/tsconfig.json
@@ -12,7 +12,7 @@
     "outDir": "../../dist/",
     "sourceMap": true,
     "target": "es5",
-    "stripInternal": true,
+    "stripInternal": false,
     "baseUrl": "",
     "typeRoots": [
       "../../node_modules/@types"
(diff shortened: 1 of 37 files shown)


mmalerba:
Do we want to mark the Error classes `@internal`? I would think we'd want those in the documentation so people know what kind of errors to expect

jelbourn:
@mmalerba 

My thinking:
* You never want to preemptively see the errors in the docs
* There's nothing to do with the error classes- they're just containers for the string messages. 
* I think the docs are better off saying what is supported instead of listing the errors that can be thrown.

mmalerba:
Ok, sounds good

PR #2101, opened 2016-12-07
feat(tooltip): add input for delaying show and hide

### LINKED ISSUE #1808: [Tooltip] allow configuration of the hide delay

andrewseguin:
Closes #1808

### DIFF
--- a/src/lib/tooltip/tooltip.ts
+++ b/src/lib/tooltip/tooltip.ts
@@ -103,8 +103,8 @@ export class MdTooltip {
     }
   }
 
-  /** Shows the tooltip after the provided delay in ms */
-  show(delay = this.showDelay): void {
+  /** Shows the tooltip after the delay in ms, defaults to tooltip-delay-show or 0ms if no input */
+  show(delay: number = this.showDelay): void {
     if (!this._message || !this._message.trim()) { return; }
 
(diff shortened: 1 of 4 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -74,6 +74,12 @@ export class MdTooltip {
     }
   }
 
+  /** The default delay in ms before showing the tooltip after show is called */
+  @Input('tooltip-show-delay') showDelay = 0;
```
jelbourn: Start using camelCase properties instead of dash-case

andrewseguin:
Changed new inputs to camelcase

PR #2110, opened 2016-12-08
fix(select): clear select if no option matches value

### LINKED ISSUE #2109: Select should clear selection if setting value programmatically to nonexistent option value
Right now if you set the select value to an option value that doesn't exist, nothing happens. To emulate native selects, it should clear the select.

kara:
Closes #2109 

r: @jelbourn

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -277,11 +277,7 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
       return;
     }
 
-    this.options.forEach((option: MdOption) => {
-      if (option.value === value) {
-        option.select();
-      }
-    });
+    this._setSelectionByValue(value);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


kara:
Noticed a weird animation thing, want to look into it

src/lib/select/select.ts
```diff
@@ -378,6 +374,28 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
     scrollContainer.scrollTop = this._scrollTop;
   }
 
+  /**
+   * Sets the selected option based on a value. If no option can be
+   * found with the designated value, the select trigger is cleared.
+   */
+  private _setSelectionByValue(value: any): void {
+    this.options.forEach((option: MdOption) => {
```
jelbourn: I don't think you need the type here, just
```ts
this.options.forEach(option => {
```
Should be equivalent

src/lib/select/select.ts
```diff
@@ -378,6 +374,28 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
+      if (option.value === value) {
+        option.select();
```
jelbourn: Stop iterating if you find a value?

kara:
@jelbourn Comments addressed!

src/lib/select/select.ts
```diff
@@ -378,6 +374,30 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
     scrollContainer.scrollTop = this._scrollTop;
   }
 
+  /**
+   * Sets the selected option based on a value. If no option can be
+   * found with the designated value, the select trigger is cleared.
+   */
+  private _setSelectionByValue(value: any): void {
+    const optionArr = this.options.toArray();
```
jelbourn: Just `options`- I have a strong aversion to suffixing with types.

jelbourn:
LGTM otherwise

src/lib/select/select.ts
kara: Sure, changing!

PR #2111, opened 2016-12-08
fix data bind bug.

Gaubee:
Fix: When options changed or  delay set, the bound data is not displayed correctly in options.


Gaubee:
I signed it!

jelbourn:
@kara 

@Gaubee do you have an example of what you're trying to fix?

jelbourn:
@Gaubee closing this since we're not sure what bug you're trying to fix. Feel free to open another PR with more details.

Issue #2113, opened 2016-12-08
Is there a 12 column responsive grid system there in material 2.?


codejith:
#### Bug, feature request, or proposal:
(Feature Request):  12 column responsive grid system

#### What is the expected behavior?
 Something like this http://materializecss.com/grid.html

#### What are the steps to reproduce?

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: http://plnkr.co/edit/o077B6uEiiIgkC0S06dd

fxck:
There won't be any grid system as part of this repo.

Check https://github.com/angular/flex-layout

hrantbagdasarian:
The Material2 feature list still includes a "layout" component. Is that still relevant or should we use the angular/flex-layout instead? If not, what are/will be the differences between both?

jelbourn:
We'll have an official blog post on this soon, stay tuned to the Angular blog.

charsi:
Might be useful to put a link to flex-layout in the docs for people looking for responsive options within Angular Material.

russelltrafford:
in Dec 2016 there was a post from @jelbourn saying there will be an official blog post on this, could a link be provided, as that would help a lot :)

chetan-rathod:
i am also stuck in this problem 
plz provide a solution for that it will help a lot to everyone

rasert:
Please help. I tried to use flex-layout to put a card in the center of the page. Center in both axis. But it does not work. I can only put it in the center of the row.

mackelito:
@rasert do you have a plunkr? If so please share and I´ll help if I can! :)

nicksloan:
@russelltrafford I'm sure you've seen this by now, but for others who might land here while researching, this is the blogpost that @jelbourn mentioned: http://angularjs.blogspot.com/2016/12/angular-material-beta-release-new-flex.html

rdineshkumar88:
For Angular Material 6.2.1, here's the tutorial to create 12 columns responsive grid. 
https://ampersandacademy.com/tutorials/angular-material/responsive-grid-layout-in-angular-material

rdineshkumar88:
It seems to be working fine. Kindly check.

rafadpedrosa:
This website is not working for me either @chauey :(

chauey:
[...]

PR #2114, opened 2016-12-08
chore: update layout/textarea status in readme

fxck:
addresses https://github.com/angular/material2/issues/2113


jelbourn:
Ah, I see you discovered our cunningly hidden other repo. We were trying to keep that on the down-low until we officially announced and released it. I suppose it doesn't hurt to link to it now, though.

PR #2117, opened 2016-12-08
fix(select): transparent background when overscrolling

crisbeto:
Fixes the select being transparent on browsers that allow overscrolling (mostly mobile).

Here's what it looks like before the fix:
img_1189

### DIFF
--- a/src/lib/select/_select-theme.scss
+++ b/src/lib/select/_select-theme.scss
@@ -34,7 +34,7 @@
     }
   }
 
-  .md-select-panel {
+  .md-select-content, .md-select-panel-done-animating {
     background: md-color($background, card);
   }
 
--- a/src/lib/select/_select-theme.scss
+++ b/src/lib/select/_select-theme.scss
@@ -34,7 +34,7 @@
     }
   }
 
-  .md-select-content {
+  .md-select-panel {
     background: md-color($background, card);
   }
 
--- a/src/lib/select/select.html
+++ b/src/lib/select/select.html
@@ -10,7 +10,7 @@
   [offsetY]="_offsetY" [offsetX]="_offsetX" (attach)="_setScrollTop()">
   <div class="md-select-panel" [@transformPanel]="'showing'" (@transformPanel.done)="_onPanelDone()"
     (keydown)="_keyManager.onKeydown($event)" [style.transformOrigin]="_transformOrigin"
-      [ngClass]="_panelDoneAnimating ? 'md-select-panel-done-animating' : ''">
+      [class.md-select-panel-done-animating]="_panelDoneAnimating">
     <div class="md-select-content" [@fadeInContent]="'showing'">
       <ng-content></ng-content>
     </div>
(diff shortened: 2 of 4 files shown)


src/lib/select/_select-theme.scss
```diff
@@ -34,7 +34,7 @@
     }
   }
 
-  .md-select-content {
+  .md-select-panel {
```
kara: @crisbeto This will change the animation and take it off spec. Any other way to fix this?

crisbeto: The background has to be on the scrollable container. I can try adding it after all animations are done, otherwise the alternative is to make the `md-select-content` scrollable, but it might break more stuff.

kara: Yeah, try adding it after the panel animation finishes.

kara: @crisbeto Given that the color should come in immediately after the animation (rather than animating in), it might be easier to do this outside of the animation definition. Have you tried adding something to the existing "done" hook for the animation?  There is an _onPanelDone method that is called when the animation finishes. Try toggling a style or class binding at that point instead.

crisbeto: The main reason was that I wanted to keep it together with the animation logic since it's animation-related. I've changed it, can you take another look @kara?

src/lib/select/select.html
```diff
@@ -5,11 +5,12 @@
...
 </div>
 
-<template connected-overlay [origin]="origin"  [open]="panelOpen" hasBackdrop (backdropClick)="close()"
+<template connected-overlay [origin]="origin" [open]="panelOpen" hasBackdrop (backdropClick)="close()"
   backdropClass="md-overlay-transparent-backdrop" [positions]="_positions" [minWidth]="_triggerWidth"
   [offsetY]="_offsetY" [offsetX]="_offsetX" (attach)="_setScrollTop()">
   <div class="md-select-panel" [@transformPanel]="'showing'" (@transformPanel.done)="_onPanelDone()"
-    (keydown)="_keyManager.onKeydown($event)" [style.transformOrigin]="_transformOrigin">
+    (keydown)="_keyManager.onKeydown($event)" [style.transformOrigin]="_transformOrigin"
+      [ngClass]="_panelDoneAnimating ? 'md-select-panel-done-animating' : ''">
```
kara: Have you tried using a regular class binding?  It might be less verbose.

```html
[class.ng-select-panel-done-animating]="_panelDoneAnimating"
```

src/lib/select/select.ts
```diff
@@ -349,6 +352,8 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
     } else {
       this.onClose.emit();
     }
+
+    this._panelDoneAnimating = true;
```
kara: It seems like this will only work the first time you open the select.  Don't you need to reset this to false when the panel closes?

crisbeto:
Updated again @kara. Good spot with resetting of the value, it would've gotten broken if it went through.

kara:
@crisbeto Is this ready to go?

crisbeto:
It is @kara

PR #2121, opened 2016-12-08
fix(checkbox): no side margin if label has no content

### LINKED ISSUE #2011: Unlabeled checkboxes shouldn't have prescribed padding/margin
Checkboxes without labels have a right margin of 8px due to the rule set in `.md-checkbox-inner-container`. There should be no prescribed margin on the checkbox in this case as it is likely being used apart from a direct text label.

belev:
* Remove side margin in checkbox if label has no content.
* Use `checkboxNativeElement` instead of `checkboxDebugElement.nativeElement` in expect statements.
* Fix typo.

Closes #2011

### DIFF
--- a/src/lib/checkbox/checkbox.scss
+++ b/src/lib/checkbox/checkbox.scss
@@ -228,6 +228,13 @@ $_mat-checkbox-mark-stroke-size: 2 / 15 * $mat-checkbox-size !default;
   }
 }
 
+.mat-checkbox-inner-container-no-side-margin {
+  margin: {
+    left: 0;
+    right: 0;
+  }
+}
+
 // TODO(kara): Remove this style when fixing vertical baseline
 .mat-checkbox-layout .mat-checkbox-label {
   line-height: 24px;
--- a/src/lib/checkbox/checkbox.ts
+++ b/src/lib/checkbox/checkbox.ts
@@ -162,6 +162,15 @@ export class MdCheckbox implements ControlValueAccessor, AfterViewInit, OnDestro
   /** The native `<input type="checkbox"> element */
   @ViewChild('input') _inputElement: ElementRef;
 
+  @ViewChild('labelWrapper') _labelWrapper: ElementRef;
+
+  /** Whether the checkbox has label */
+  _hasLabel(): boolean {
+    const labelText = this._labelWrapper.nativeElement.textContent || '';
+    return !!labelText.trim().length;
+  }
+
+  /** Called when the checkbox is blurred. Needed to properly implement ControlValueAccessor. */
   @ViewChild(MdRipple) _ripple: MdRipple;
 
   /**
--- a/src/lib/checkbox/checkbox.html
+++ b/src/lib/checkbox/checkbox.html
@@ -1,5 +1,6 @@
 <label class="mat-checkbox-layout" #label>
-  <div class="mat-checkbox-inner-container">
+  <div class="mat-checkbox-inner-container"
+       [class.mat-checkbox-inner-container-no-side-margin]="!_hasLabel()">
     <input #input
            class="mat-checkbox-input cdk-visually-hidden" type="checkbox"
            [id]="inputId"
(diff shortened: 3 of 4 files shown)


tinayuangao:
LGTM

jelbourn:
Could we not move that margin to the label element and then remove it using the `:empty` pseudo-class? That would be way simpler.

belev:
@jelbourn I totally agree that with such pseudo-class it will be way simpler and better because that's a styling problem. I tried it but couldn't make it work. Using `:empty` match whitespaces as content and don't apply the changes if we have any whitespace. I saw a pseudo-class `:blank` which handles whitespaces but unfortunately the browsers don't supported it.

belev:
Is there some update about the PR and potentially merging it?

jelbourn:
@belev I want to dedicate a chunk of time to investigate this more in depth but haven't yet had the opportunity. I'll come back here and update when I do.

belev:
@jelbourn Great, looking forward to it.

P.S. sorry for the delayed rebase, I didn't see the changed label.

src/lib/checkbox/checkbox.scss
```diff
@@ -293,6 +300,13 @@ $_mat-checkbox-mark-stroke-size: 2 / 15 * $mat-checkbox-size !default;
         right: $_mat-checkbox-item-spacing;
       }
     }
+
+    &.mat-checkbox-inner-container-no-side-margin {
```
jelbourn: You don't need to define this class twice; just one without the parent part of the selector would be better.

src/lib/checkbox/checkbox.ts
```diff
@@ -164,6 +164,13 @@ export class MdCheckbox implements ControlValueAccessor, AfterViewInit, OnDestro
 
   @ViewChild(MdRipple) _ripple: MdRipple;
 
+  @ViewChild('labelWrapper') _labelWrapper: ElementRef;
+
+  get hasLabel(): boolean {
```
jelbourn: Let's change this to a method rather than a getter. It should also be marked as internal (name with a leading underscore). It also needs a JsDoc description comment.

src/lib/checkbox/checkbox.ts
```diff
@@ -164,6 +164,13 @@ export class MdCheckbox implements ControlValueAccessor, AfterViewInit, OnDestro
...
+    const labelWrapperTextContent = this._labelWrapper.nativeElement.textContent;
```
jelbourn: This will be a problem for Angular Universal. Currently, `nativeElement` is a parse5 node, which doesn't have a `textContent` property. So for now you can say
```ts
const labelText = this._labelWrapper.nativeElement.textContent || '';
return !!labelText.trim().length;
```

belev:
I addressed the comments but can't understand why the build fails.

dahaupt:
@belev Probably a Travis bug, happens sometimes as far as I know. Can you restart the CI process?

belev:
@dahaupt I don't think so, doesn't see any restart button. Maybe I can force it by closing and opening the PR or making some new dummy commit but don't think that's a good idea.

DevVersion:
Yeah looks like a flake. Is your PR rebased properly? Also restarted the build manually.

belev:
What do you mean by properly rebased?

DevVersion:
I meant rebased on top of `master`?

belev:
No, it is not.

DevVersion:
That could have caused the `SyntaxError` in the Travis CI. Just that you know, rebasing could solve those issues.

belev:
Thanks. So I assume it is better to always rebase on top of master?

DevVersion:
@belev Yep always :)

belev:
Ok, thanks. I will make the proper rebase when I have time then.

belev:
Done, after all :smile:

w3t:
This change broke the layout of my login screen :(

When rendering:
<md-checkbox [(ngModel)]="savePassword">{{ "tx_Login_SavePassword" | translate }}</md-checkbox>

It considers that is has no label at render time and therefore renders the checkbox without its margin (until one clicks on the checkbox).

DevVersion:
@w3t Please file a new issue and provide a Plunker demo for reproduction?

PR #2122, opened 2016-12-08
fix(aot): add index.ts exports as workaround for ngc issue

jelbourn:
This fix is already released in alpha.11-1


src/lib/menu/index.ts
```diff
@@ -1 +1,8 @@
 export * from './menu';
```
dahaupt: I get the following error message while compiling my project (with Angular-CLI Webpack):
```
WARNING in ./~/@angular/material/module.js
44:4 export 'MdMenuModule' was not found in './menu/index'
```

Running the latest version: **2.0.0-alpha.11-1**. It wasn't there before.

dahaupt: Fixed with latest version:  **2.0.0-alpha.11-2**

PR #2130, opened 2016-12-09
fix(checkbox): Emit event when checkbox's indeterminate value changes

### LINKED ISSUE #2004: [md-checkbox] checked and indeterminate state problem.
#### Bug
I believe this is a bug.

If both checked and indeterminate are true and then checked is set to false it should show indeterminate.

If both checked and indeterminate are true and then checked is set to false it should show indeterminate but instead, it shows unchecked.

If you use the toggles you will see the state of the checkbox not being correct against the set values.
https://embed.plnkr.co/v7euQHvzVsAFHNcpsEFo

Latest

tinayuangao:
Fixes #2004 

R: @jelbourn

### DIFF
--- a/src/lib/checkbox/checkbox.ts
+++ b/src/lib/checkbox/checkbox.ts
@@ -148,7 +148,7 @@ export class MdCheckbox implements ControlValueAccessor {
   @Output() change: EventEmitter<MdCheckboxChange> = new EventEmitter<MdCheckboxChange>();
 
   /** Event emitted when the checkbox's `indeterminate` value changes. */
-  @Output() indeterminateChange: EventEmitter<Boolean> = new EventEmitter<Boolean>();
+  @Output() indeterminateChange: EventEmitter<boolean> = new EventEmitter<boolean>();
 
   /** The native `<input type="checkbox"> element */
   @ViewChild('input') _inputElement: ElementRef;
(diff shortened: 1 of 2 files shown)


src/lib/checkbox/checkbox.ts
```diff
@@ -203,6 +209,7 @@ export class MdCheckbox implements ControlValueAccessor {
       this._transitionCheckState(
           this.checked ? TransitionCheckState.Checked : TransitionCheckState.Unchecked);
     }
+    this.indeterminateChange.emit(this._indeterminate);
```
jelbourn: Should this event only fire when the value changes?

src/lib/checkbox/checkbox.ts
```diff
@@ -138,6 +138,9 @@ export class MdCheckbox implements ControlValueAccessor {
   /** Event emitted when the checkbox's `checked` value changes. */
   @Output() change: EventEmitter<MdCheckboxChange> = new EventEmitter<MdCheckboxChange>();
 
+  /** Event emitted when the checkbox's `indeterminate` value changes. */
+  @Output() indeterminateChange: EventEmitter<Boolean> = new EventEmitter<Boolean>();
```
jelbourn: `boolean` lower-case

tinayuangao:

djanesch:
when will this change released?
@tinayuangao @jelbourn

PR #2132, opened 2016-12-09
fix: TSError: ? Unable to compile TypeScript when running 'gulp:build'

kuncevic:
fix (gulp task): change `Promise<void>` to `Promise<{}>`

When running `gulp build` or `gulp serve:devapp` getting an error `TSError: ? Unable to compile TypeScript
tools\gulp\tasks\release.ts (52,10): Type 'Promise<{}>' is not assignable to type 'Promise<void>'`


crisbeto:
It seems like this is the same change as in #2131. Can one of them be closed?

PR #2135, opened 2016-12-09
build: fix wrong SASS path

### LINKED ISSUE #2133: Some SCSS files refer to invalid path (a11y)
In [5cf7d175d66120b0e8262bef561f230cb6e31d11](https://github.com/angular/material2/commit/5cf7d175d66120b0e8262bef561f230cb6e31d11) several components got **a11y** support using a new SCSS resource.

The import is `@import '../a11y/_a11y';`
Which looks for the file in **material/src/lib/a11y/_a11y**

**a11y** is in **core**

The file exists in **material/src/lib/core/a11y/_a11y**

I don't know how the build passed but it's should be fixed

`@import '../core/a11y/_a11y';`

The files:
src/lib/button/button.scss
src/lib/card/card.scss
src/lib/dialog/dialog-container.scss
src/lib/menu/menu.scss
src/lib/select/select.scss
src/lib/sidenav/sidenav.scss
src/lib/slide-toggle/slide-toggle.scss
src/lib/snack-bar/snack-bar-container.scss
src/lib/tooltip/tooltip.scss

@crisbeto @mmalerba

crisbeto:
Fixes SASS errors when Material is imported directly.

Fixes #2133.

### DIFF
--- a/src/lib/button/button.scss
+++ b/src/lib/button/button.scss
@@ -2,7 +2,7 @@
 // TODO(jelbourn): Figure out if anchor hover underline actually happens in any browser.
 
 @import 'button-base';
-@import '../a11y/_a11y';
+@import '../core/a11y/a11y';
 
 // TODO(kara): Replace attribute selectors with class selectors when possible
 [md-button], [md-icon-button] {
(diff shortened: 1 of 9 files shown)


jelbourn:
LGTM

I'd like to figure out why this didn't cause a problem: our build worked, sample apps worked, and all of the internal Google sass builds using Material worked. Is it possibly something that is only an error in older versions of sass? Is it a ruby vs. libsass thing?

crisbeto:
No idea, I can spend some more time with it over the weekend.

PR #2138, opened 2016-12-09
docs(radio): Delete example for two-way binding

### LINKED ISSUE #1790: [md-radio-group] The value two-way binding not working?
Bug [x]

According to the readme.md of the radio, there is `value` property on the group with two way binding. 
If I set a variable with one of the possible value selections of the radios inside the group, then the radio which value attribute matches the variable's value will be selected. But if I select another radio the variable's value will not get updated. I guess the cause of this behaviour is that there is no `valueChange EventEmitter` created in the `MdRadioGroup` class

The variable's value should updated when new radio is selected

The variable's value is not updating when new radio is selected

Plunker template: 
http://plnkr.co/edit/8aHnV0luO8Oie2kwQqek?p=preview

There is the change event which has the value property so that might be enough, but then the readme.md file should be updated.

2.0.0
2.0.0-alpha.9-3
Mac os X crhome

not too much experience with Angular 2

tinayuangao:
- Delete the `[(value)]` example because we don't support it. There's another example for `[(ngModel)]`
Closes #1790 

R: @jelbourn

### DIFF
--- a/src/lib/radio/README.md
+++ b/src/lib/radio/README.md
@@ -12,15 +12,6 @@ A basic radio group would have the following markup.
 </md-radio-group>
 ```
 
-A dynamic example, populated from a `data` variable:
-```html
-<md-radio-group [(value)]="groupValue">
-  <md-radio-button *ngFor="let d of data" [value]="d.value">
-    {{d.label}}
-  </md-radio-button>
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #2145, opened 2016-12-09
fix(sidenav): make `focus-trap` occupy full height of sidenav

mmalerba:


src/lib/sidenav/sidenav.scss
```diff
@@ -108,8 +108,15 @@ md-sidenav {
   z-index: 3;
   min-width: 5%;
 
-  // TODO(kara): revisit scrolling behavior for sidenavs
-  overflow-y: auto;
+  > focus-trap {
```
jelbourn: `.md-sidenav-focus-trap` etc.

mmalerba:
comments addressed

src/lib/core/a11y/focus-trap.html
```diff
@@ -1,3 +1,3 @@
 <div *ngIf="!disabled" tabindex="0" (focus)="focusLastTabbableElement()"></div>
-<div #trappedContent><ng-content></ng-content></div>
+<div #trappedContent class="focus-trap-content"><ng-content></ng-content></div>
```
jelbourn: Also prefix on `.focus-trap-content`. This is as good a time as any to introduce the `cdk-` prefix.

mmalerba: shouldn't we wait and do that all at once?

jelbourn: I don't see any reason not to start now

PR #2154, opened 2016-12-10
chore: fix invalid check for travis pull request

DevVersion:
* The recently added script for the per-commit publishing is not properly checking for the environment variable.
 
  This means that the per-commit packages can be not created yet.


jelbourn:
LGTM

PR #2155, opened 2016-12-10
test(ripple): refactor long-running unit test

### LINKED ISSUE #2097: Single ripple unit test uses a 1600ms timeout
The ripple spec has a test `removes foreground ripples after timeout` that requires the use of a 1600ms timeout to test the removal of ripples. Consider how to make this test run faster by removing an actual call to setTimeout.

crisbeto:
Refactors a unit test that used to take 1.6s, due to it using a native `setTimeout`, by replacing the native timer with a flushable one.

Fixes #2097.


jelbourn:
LGTM, thanks. I thought that `fakeAsync` wouldn't work because the `setTimeout` is outside the Angular zone, but I guess it works since it's still captured by the root zone.

PR #2156, opened 2016-12-10
perf(sidenav): avoid extra repaints while scrolling

crisbeto:
Avoids repaints while scrolling inside of a sidenav. This isn't an issue on desktop browsers, but can potentially be on mobile. This can be checked by going to the dev tools > console fold-out menu > Rendering > Scrolling Performance Issues.


jelbourn:
LGTM

PR #2159, opened 2016-12-10
refactor(interactivity-checker.spec): remove redundant `setTimeout` usage

belev:


DevVersion:
The reason why a native `setTimeout` was used here was that the `delay` comes from the browser itself and can't be flushed from a zone.

I just tried to run the tests locally without the `tick` or `setTimeout` and it seems to succeed anyways (not sure what flushes the queue)

<details>

```js
let div = document.createElement('div'); document.body.appendChild(div); div.contentEditable = true; console.log(div.tabIndex);
```

</details>

belev:
Mhm interesting, maybe this should be closed then. However I will try to find why this happens.

DevVersion:
@belev Sounds good. Thank you!

belev:
After some digging around that's what I understood:

`tabIndex` isn't set in the case where we have only `contenteditable`, it actually has value of `null`. So basically the `tick` or `setTimeout` didn't do anything in this case. And the test passes because of [this line](https://github.com/Belev/material2/blob/4b7e52d30aa342c3024ed078837a9bd4ca9e773b/src/lib/core/a11y/interactivity-checker.ts#L68).

So after all the PR is not totally useless :smile: I can modify it and remove the usage of `setTimeout` or `tick`.

What do you think @DevVersion ?

DevVersion:
Ah I can see. The checks you mentioned are intentionally only respecting the `user-specified` tabindex instead of the evaluated one from the browser.

And as you said that's why the tests pass. We can remove the tick stuff safely here.

belev:
Made the change, you can review it again. : )

jelbourn:
LGTM

PR #2162, opened 2016-12-10
build: avoid multiple refreshes when running unit tests locally

crisbeto:
Avoids Karma running tests multiple times when developing locally, due to it triggering a test run whenever a new file is written to disk. This change switches to manually spawning a Karma server and then running the tests by depending on Gulp's API instead.

For reference, here's what running tests on `master` looks like:
old

And here's what it looks like with the new changes:
new

### DIFF
--- a/test/karma.config.ts
+++ b/test/karma.config.ts
@@ -58,7 +58,7 @@ export function config(config) {
     port: 9876,
     colors: true,
     logLevel: config.LOG_INFO,
-    autoWatch: true,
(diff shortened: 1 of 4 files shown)


jelbourn:
LGTM

I tried it out, way way better. Thanks!

jelbourn:
@crisbeto can you rebase?

crisbeto:
Done @jelbourn.

PR #2166, opened 2016-12-11
fix(select): make invalid selector more specific

### LINKED ISSUE #2147: md-select applying error styles before it is touched.
When another control in a form is touched, md-select applies error styles even though it hasn't been touched yet.  The classes on md-select remain class="ng-untouched ng-pristine ng-invalid".

kara:
The selector for the touched/invalid style was too general, so it was picking up when the parent form was "touched" rather than the control itself.  This PR ensures it only applies touched/invalid styles when its own control is touched/invalid.

Fixes #2147.

r: @jelbourn

### DIFF
--- a/src/lib/select/_select-theme.scss
+++ b/src/lib/select/_select-theme.scss
@@ -16,7 +16,7 @@
       border-bottom: 1px solid md-color($primary);
     }
 
-    .ng-invalid.ng-touched:not(.md-select-disabled) & {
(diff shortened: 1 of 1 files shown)


dreweird:
It still the same....all the `md-select` will turn red even without touching the selector

kara:
@dreweird This change isn't merged into master.

jelbourn:
LGTM

FilipeQ:
Anyone know when it will be released?

PR #2167, opened 2016-12-11
fix(input): fix underline color to match spec

### LINKED ISSUE #2126: Input underline appears to be too dark
Input underline appears darker than the spec dictates: https://material.google.com/components/text-fields.html#text-fields-labels.  As a result, input and select look odd next to each other:

http://plnkr.co/edit/aLXsp8MlV2AoOkC9aEFP?p=preview

kara:
Closes #2126.   FYI also removed a related Sass variable because it wasn't actually being used anywhere.

r: @jelbourn


jelbourn:
LGTM

PR #2169, opened 2016-12-11
build: fix invalid sass paths being resolved

crisbeto:
Prevents SASS from resolving certain paths, even though they're invalid. This should prevent issues like #2133 from happening again.

@jelbourn I'm not sure what the reasoning was for adding `core/styles` to the `includePaths` in Gulp, but this is what helped `@import '../a11y/_a11y';` to always be resolved. I re-ran all of the SASS-related Gulp tasks and they went through fine.


jelbourn:
LGTM, good find.

PR #2170, opened 2016-12-11
feat(theming): provide a content wrapper attribute

### LINKED ISSUE #1938: Container for non-sidenav-layouts
For layouts not using the `<md-sidenav-layout>`, dark themes can not be used, as the background does not change. Due to that, the text gets invisible and form items look bad.

To understand, what I mean, check out the plunkr. Basically I took your template and changed the theme: http://plnkr.co/edit/w4v95im2irvNm4dAzQ5k

I was looking for some container class or element, but wasn't lucky in finding any.

### LINKED ISSUE #2106: <md-input> placeholders are not working on the pink-bluegrey theme.

Bug report.

Placeholders show up on &lt;md-input&gt; when using the pink-bluegrey theme.

Placeholder text is white until the &lt;md-input&gt; is focused.

http://plnkr.co/edit/qjcghUoBEs0GMING1eS9?p=preview

Alpha 10, Chrome 54

crisbeto:
Adds the `md-app-background` class which users can apply to their wrapper element, in order to provide the theme background. This is necessary, because the only element that does this at the moment is `md-sidenav-container` and it's not something that necessarily everybody uses.

Fixes #1938.
Fixes #2106.

### DIFF
--- a/guides/theming.md
+++ b/guides/theming.md
@@ -40,6 +40,10 @@ The actual path will depend on your server setup.
 
 You can also concatenate the file with the rest of your application's css.
 
+Finally, if your app's content **is not** placed inside of a `md-sidenav-container` element, you
+need to add the `md-app-background` class to your wrapper element (for example the `body`). This
+ensures that the proper theme background is applied to your page.
+
 ### Defining a custom theme
 When you want more customization than a pre-built theme offers, you can create your own theme file.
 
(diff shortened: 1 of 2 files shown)


jelbourn:
I like the idea, but not sure about the name. The name should communicate exactly what it does, so I'm thinking maybe `md-main-content-background`. Perhaps also making it a class instead of an attr. Thoughts?

crisbeto:
Regarding the name, I went with "wrapper" in case we wanted to use it for something else in the future (like `md-content` in Material 1). As for the attribute, it's in case we want to turn it into a directive afterwards.

crisbeto:
Renamed it to `md-app-background` and switched it over to a class @jelbourn.

PR #2186, opened 2016-12-12
fix(tabs): observing tab header label changes to recalculate width

EladBezalel:
- Uses observe-changes directive that emit an event when the mutation observer notifies a change

fixes #2155

### DIFF
--- a/src/lib/core/observe-content/observe-content.spec.ts
+++ b/src/lib/core/observe-content/observe-content.spec.ts
@@ -1,0 +1,68 @@
+import {Component} from '@angular/core';
+import {async, TestBed} from '@angular/core/testing';
+import {ObserveContentModule} from './observe-content';
+
+/**
+ * TODO(elad): `ProxyZone` doesn't seem to capture the events raised by
+ * `MutationObserver` and needs to be investigated
+ */
+
+describe('Observe content', () => {
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [ObserveContentModule],
+      declarations: [ComponentWithTextContent, ComponentWithChildTextContent],
+    });
+
+    TestBed.compileComponents();
+  }));
+
+  describe('text content change', () => {
+    it('should call the registered for changes function', done => {
+      let fixture = TestBed.createComponent(ComponentWithTextContent);
+      fixture.detectChanges();
+
+      // If the hint label is empty, expect no label.
+      const spy = spyOn(fixture.componentInstance, 'doSomething').and.callFake(() => {
+        expect(spy.calls.any()).toBe(true);
+        done();
+      });
+
+      expect(spy.calls.any()).toBe(false);
+
+      fixture.componentInstance.text = 'text';
+      fixture.detectChanges();
+    });
+  });
+
+  describe('child text content change', () => {
+    it('should call the registered for changes function', done => {
+      let fixture = TestBed.createComponent(ComponentWithChildTextContent);
+      fixture.detectChanges();
+
+      // If the hint label is empty, expect no label.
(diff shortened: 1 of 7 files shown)


src/lib/core/observe-content/observe-content.ts
```diff
@@ -0,0 +1,38 @@
...
+})
+export class MdMutationObserver {
+  private observer: MutationObserver;
+
+  @Output('cdk-observe-content') event = new EventEmitter();
+
+  constructor(private elementRef: ElementRef) {}
+
+  ngAfterViewInit() {
+    this.observer = new MutationObserver(mutations => mutations.forEach(() => this.event.emit()));
```
crisbeto: Might be a good idea to have a method to clean up the mutation observer, otherwise we'll end up with memory leaks at some point.

src/lib/core/observe-content/observe-content.ts
```diff
@@ -0,0 +1,38 @@
...
+  NgModule,
+  ModuleWithProviders,
+  Output,
+  EventEmitter
+} from '@angular/core';
+
+@Directive({
+  selector: '[cdk-observe-content]'
+})
+export class MdMutationObserver {
```
EladBezalel: Oops..

src/lib/core/observe-content/observe-content.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      const spy = spyOn(fixture.componentInstance, 'doSomething').and.callFake(() => {
+        expect(spy.calls.any()).toBe(true);
+        done();
+      });
+
+      expect(spy.calls.any()).toBe(false);
+
+      fixture.componentInstance.text = 'text';
+      fixture.detectChanges();
+    });
```
jelbourn: Just tried this locally and I saw that the async stuff doesn't really work. Seems like zones doesn't play well with `MutationObserver`...

jelbourn: Hmm, Zones is supposed to understand `MutationObserver`, I'll have to check with Misko on this tomorrow.

EladBezalel:
@jelbourn @crisbeto please re-review :)

src/lib/core/observe-content/observe-content.ts
```diff
@@ -0,0 +1,46 @@
...
+  NgModule,
+  ModuleWithProviders,
+  Output,
+  EventEmitter
+} from '@angular/core';
+
+@Directive({
+  selector: '[cdk-observe-content]'
+})
+export class ObserveContent {
```
crisbeto: You should add an annotation to the class: `export class ObserveContent implements OnDestroy, AfterContentInit`.

EladBezalel: When you're right, you're right!

src/lib/tabs/tab-header.ts
```diff
@@ -150,6 +148,16 @@ export class MdTabHeader {
     }
   }
 
+  /**
+   * Updating the view whether pagination should be enabled or not
+   * @private
```
jelbourn: Don't use `@private` JsDoc, it messes up stuff when we sync into google

jelbourn:
@andrewseguin can you review as well?

src/lib/core/observe-content/observe-content.ts
```diff
@@ -0,0 +1,50 @@
...
+  AfterContentInit
+} from '@angular/core';
+
+@Directive({
+  selector: '[cdk-observe-content]'
+})
+export class ObserveContent implements AfterContentInit, OnDestroy {
+  private _observer: MutationObserver;
+
+  @Output('cdk-observe-content') event = new EventEmitter<void>();
```
andrewseguin: Should this be camelcased?

PR #2188, opened 2016-12-13
feat(scroll): provide directive and service to listen to scrolling

andrewseguin:
Includes applying the directive to the sidenav as an example of usage. The tooltip makes use of the scroll service's notification and updates in position.

Would be nice to have some optimization on throttling the scroll events, ideally something similar to throttle but includes the trailing event.


src/lib/core/scroll/scroll.ts
```diff
@@ -0,0 +1,59 @@
...
+import {Observable} from 'rxjs/Observable';
+import {Subscription} from 'rxjs/Subscription';
+
+
+/**
+ * Service contained all registered Scrollable references and emits an event when any one of the
+ * Scrollable references emit a scrolled event.
+ */
+@Injectable()
+export class Scroll {
```
jelbourn: I would call this `ScrollMonitor` or `ScrollDispatcher` (I'm partial to the latter)

src/lib/core/scroll/scroll.ts
```diff
@@ -0,0 +1,59 @@
...
+  /** Subject for notifying that a registered scrollable reference element has been scrolled. */
+  _scrolled: Subject<Event> = new Subject();
+
+  /**
+   * Map of all the scrollable references that are registered with the service and their
+   * scroll event subscriptions.
+   */
+  scrollableReferences: Map<Scrollable, Subscription> = new Map();
```
jelbourn: Use a `WeakMap`

src/lib/core/scroll/scroll.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor() {
+    // By default, notify a scroll event when the document is scrolled or the window is resized.
+    window.document.addEventListener('scroll', this._notify.bind(this));
+    window.addEventListener('resize', this._notify.bind(this));
```
jelbourn: Prefer arrow functions to `.bind`:
```ts
window.document.addEventListener('scroll', () => this.notify());
```
(here and elsewhere)

src/lib/core/scroll/scroll.ts
```diff
@@ -0,0 +1,59 @@
...
+
+
+/**
+ * Service contained all registered Scrollable references and emits an event when any one of the
+ * Scrollable references emit a scrolled event.
+ */
+@Injectable()
+export class Scroll {
+  /** Subject for notifying that a registered scrollable reference element has been scrolled. */
+  _scrolled: Subject<Event> = new Subject();
```
jelbourn: Do you need the `Event`, or can this be `void`?

src/lib/core/scroll/scrollable.ts
```diff
@@ -0,0 +1,53 @@
...
+import {Scroll} from './scroll';
+
+
+/**
+ * Sends an event when the directive's element is scrolled. Registers itself with the Scroll
+ * service to include itself as part of its collection of scrolling events that it can be listened
+ * to through the service.
+ */
+@Directive({
+  selector: '[md-scrollable]'
```
jelbourn: Let's make this `cdk-scrollable` (it's starting!)

src/lib/core/scroll/scrollable.ts
```diff
@@ -0,0 +1,53 @@
...
+    return this._elementScrolled.asObservable();
+  }
+}
+
+
+@NgModule({
+  exports: [Scrollable],
+  declarations: [Scrollable],
+})
+export class ScrollModule {
```
jelbourn: I'm not sure this needs its own module vs. just being a part of `OverlayModule`. Thoughts?

src/lib/core/scroll/scroll.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    scrollable.elementScrolled().subscribe(() => { hasDirectiveScrollNotified = true; });
+
+    // Listen for notifications from scroll service
+    let hasServiceScrollNotified = false;
+    scroll.scrolled().subscribe(() => { hasServiceScrollNotified = true; });
+
+    // Emit a scroll event from the scrolling element in our component.
+    // This event should be picked up by the scrollable directive and notify.
+    // The notification should be picked up by the service.
+    fixture.componentInstance.scrollingElement.nativeElement.dispatchEvent(new Event('scroll'));
```
jelbourn: `new Event` isn't supported in IE11

src/lib/core/scroll/scrollable.ts
```diff
@@ -0,0 +1,53 @@
...
+})
+export class Scrollable implements OnInit, OnDestroy {
+  /** Subject for notifying that the element has been scrolled. */
+  private _elementScrolled: Subject<Event> = new Subject();
+
+  constructor(private _elementRef: ElementRef, private _scroll: Scroll) {}
+
+  ngOnInit() {
+    this._scroll.register(this);
+    this._elementRef.nativeElement.addEventListener('scroll', (e: Event) => {
```
crisbeto: Instead of binding the event directly, [I think you could use `Observable.fromEvent`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/fromevent.md). Also I think the same can be done above in the service.

DzmitryShylovich: `this.renderer.listen(this._elementRef.nativeElement, 'scroll', (e: Event) => {})`

src/lib/core/scroll/scroll.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Map of all the scrollable references that are registered with the service and their
+   * scroll event subscriptions.
+   */
+  scrollableReferences: Map<Scrollable, Subscription> = new Map();
+
+  constructor() {
+    // By default, notify a scroll event when the document is scrolled or the window is resized.
+    window.document.addEventListener('scroll', this._notify.bind(this));
```
crisbeto: It might be better to bind this only after a component has been registered. Also we could unbind it when there are no subscriptions.

src/lib/core/scroll/scroll.ts
DzmitryShylovich: `this.renderer.listenGlobal('window', 'resize', (e: Event) => {})`

src/lib/core/scroll/scroll.ts
crisbeto: Sounds good. Also something else that I was thinking about: I'm pretty sure that this event will trigger Angular's change detection. It may be worth trying to run it outside Angular's zone, however I'm not sure whether it won't end up updating the view at the proper times.

src/lib/core/scroll/scroll-dispatcher.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Map of all the scrollable references that are registered with the service and their
+   * scroll event subscriptions.
+   */
+  scrollableReferences: WeakMap<Scrollable, Subscription> = new WeakMap();
+
+  constructor() {
+    // By default, notify a scroll event when the document is scrolled or the window is resized.
+    Observable.fromEvent(window.document, 'scroll').subscribe(() => this._notify());
+    Observable.fromEvent(window, 'resize').subscribe(() => this._notify());
```
DzmitryShylovich: this is still won't work in universal.
Plus I believe u need to unsubscribe from it on destroy

andrewseguin: Right now we are not prioritizing universal right now so this may be a concern for another time. Additionally I think there may be some confusion here as this is a service and not a component so it does not have access to a host nor have a lifecycle with destroy.

andrewseguin:
Moved the scroll dispatcher and scrollable directive to the overlay folder.

src/lib/core/scroll/scroll-dispatcher.ts
DzmitryShylovich: > nor have a lifecycle with destroy.

it's not true. services do have `OnDestory` hook :)

> is a service and not a component so it does not have access to a host

right, I missed that part. ignore my comments

src/lib/tooltip/tooltip.ts
```diff
@@ -92,10 +94,23 @@ export class MdTooltip {
     }
   }
 
-  constructor(private _overlay: Overlay, private _elementRef: ElementRef,
-              private _viewContainerRef: ViewContainerRef, private _ngZone: NgZone,
+  constructor(private _overlay: Overlay,
+              private _scroll: ScrollDispatcher,
```
jelbourn: `_scrollDispatcher`

andrewseguin:
Changed _scroll to _scrollDispatcher

andrewseguin:
Rebased

PR #2191, opened 2016-12-13
fix(input): copy input state classes to md-input-container

mmalerba:


mmalerba:
@jelbourn directly binding to the `ngControl` of the content child doesn't seem to cause any issues. @kara asked for some sanity checks on `formControl` as well so I added that to the demo.

PR #2193, opened 2016-12-13
fix(positioning): fallback positions should work while scrolled

kara:
r: @jelbourn


kara:
Firefox and IE are unhappy. Looking into it

jelbourn:
LGTM

PR #2200, opened 2016-12-13
chore(artifacts): properly authenticate git.

DevVersion:
* Exit process if pipeline fails or any line fails.
* Fixes authentication for Git


jelbourn:
LGTM

PR #2210, opened 2016-12-14
Create screen type service to determine screen type

josephperrott:


ThomasBurleson:
@josephperrott  - I am curious why this is in Material Core and not in Flex-Layout.

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,102 @@
...
+  registerBreakpoint(query: string): Breakpoint {
+    let breakpoint: InternalBreakpoint;
+    if (this.breakpoints.has(query)) {
+      breakpoint = this.breakpoints.get(query);
+      breakpoint.usageCount += 1;
+      window.setTimeout(() => {
+        breakpoint.subject.next(breakpoint.watcher.matches);
+      });
+    } else {
+      let watcher = window.matchMedia(query);
```
ThomasBurleson: You may want to consider a WebKit consideration:  https://github.com/angular/flex-layout/blob/master/src/lib/media-query/match-media.ts#L147

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,102 @@
...
+      let subject = new Subject();
+      let callback = (mql: MediaQueryList) => {
+        this.ngZone.run(() => subject.next(mql.matches));
+      };
+      subject.subscribe(null, null, () => {
+        this.breakpoints.delete(query);
+      });
+      watcher.addListener(callback);
+      window.setTimeout(() => {
```
ThomasBurleson: Any reason why this is asynchronous here ?

josephperrott:
@ThomasBurleson We do not want to have a direct dependency on Flex-Layout, we want to have the library work correctly with it, but not rely on it as a dependency.

@jelbourn can comment/confirm

src/lib/core/adaptive/screen-type.ts
josephperrott: Good call.  Need to update to do that.

src/lib/core/adaptive/screen-type.ts
josephperrott: I needed to make sure that the most recent value was used.  I was able to achieve this by just using a BehaviorSubject.

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Private global registry for all dynamically-created, injected style tags
+ * @see prepare(query)
+ */
+const ALL_STYLES: {[key: string]: HTMLStyleElement} = {};
+
+
+/**
+ * For Webkit engines that only trigger the MediaQueryListListener
+ * when there is at least one CSS selector for the respective media query.
```
jelbourn: Do you have a resource we can link to for this?

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * @param query string The mediaQuery used to create a faux CSS selector
+ *
+ */
+function prepareQueryCSS(query: string) {
+  if (!ALL_STYLES[query]) {
+    try {
```
jelbourn: Why is the `try` necessary?

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
...
+
+
+/**
+ * For Webkit engines that only trigger the MediaQueryListListener
+ * when there is at least one CSS selector for the respective media query.
+ *
+ * @param query string The mediaQuery used to create a faux CSS selector
+ *
+ */
+function prepareQueryCSS(query: string) {
```
jelbourn: I'd call this something like `insertEmptyStyleRuleForWebkitCompatibility`, and possible use the `Platform` service to gate. Then `ALL_STYLES` would be something like `styleElementForWebkitCompatibility` (not using CAPS_CASE because the value itself isn't a constant)

ThomasBurleson:
@jelbourn - While this code is a well-formed, I am surprised you are considering merging a separate codebase for mediaQueries and observables. I realize that you do not want an explicit dependency upon Flex-Layout. 

So the alternatives are 
*  cloning those flex-layout files to material2, or
*  re-invent that functionality here

> I worry that the *re-invent* approach creates a divergence that will not be sync'ed and will cause future problems.

*  [match-media.ts](https://github.com/angular/flex-layout/blob/master/src/lib/media-query/match-media.ts)
* [media-monitor.ts](https://github.com/angular/flex-layout/blob/master/src/lib/media-query/media-monitor.ts)
*  [mock-match-media.ts](https://github.com/angular/flex-layout/blob/master/src/lib/media-query/mock/mock-match-media.ts)

src/lib/core/adaptive/screen-type.ts
josephperrott: Done, not sure what you are looking for the platform service to do here?

src/lib/core/adaptive/screen-type.ts
josephperrott: The try is in place to prevent this method from throwing an error and causing the stack to fail.  This way it "fails gracefully"

src/lib/core/adaptive/screen-type.ts
josephperrott: I will work on finding one, this is based on the flex-layout implementation

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
...
+   * Registers a breakpoint to be listened for.
+   */
+  registerBreakpoint(query: string): Breakpoint {
+    let breakpoint: InternalBreakpoint;
+    if (this.breakpoints.has(query)) {
+      breakpoint = this.breakpoints.get(query);
+      breakpoint.usageCount += 1;
+    } else {
+      prepareQueryCSS(query);
+      let watcher = window.matchMedia(query);
```
josephperrott: Done

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
...
+  callback: (mql: MediaQueryList) => void;
+  usageCount: number;
+  watcher: MediaQueryList;
+}
+
+@Injectable()
+export class ScreenType {
+  private breakpoints: Map<string, InternalBreakpoint> = new Map();
+
+  constructor(private ngZone: NgZone) {}
```
josephperrott: Done

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
...
+  destroy: () => void;
+}
+
+interface InternalBreakpoint {
+  query: string;
+  subject: BehaviorSubject<boolean>;
+  callback: (mql: MediaQueryList) => void;
+  usageCount: number;
+  watcher: MediaQueryList;
+}
```
josephperrott: Naming could probably be better here, but essentially InternalBreakpoint is an object containing all of the information needed to create a Breakpoint to be handed out.

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
+import {Injectable, NgZone} from '@angular/core';
+import {Observable} from 'rxjs/Observable';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+
+
+export const SCREEN_TYPES = {
+  HANDSET: '(max-width: 600px) and (orientation: portrait), (max-width: 960px) and ' +
+           '(orientation: landscape)',
```
josephperrott: Done

src/lib/core/adaptive/screen-type.ts
```diff
@@ -0,0 +1,134 @@
+import {Injectable, NgZone} from '@angular/core';
+import {Observable} from 'rxjs/Observable';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+
+
+export const SCREEN_TYPES = {
```
josephperrott: Done

jelbourn:
@ThomasBurleson for now this would just be an internal utility that our own components use. Eventually I'd like to move this functionality into the CDK.

tools/gulp/packaging/rollup-helpers.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (config.format === 'umd') {
+    bundleOptions.plugins = [rollupNodeResolutionPlugin()];
+
+    const external = Object.keys(ROLLUP_GLOBALS);
+    external.splice(external.indexOf('tslib'), 1);
+    bundleOptions.external = external;
+  }
+
+  return rollup.rollup(bundleOptions).then((bundle: any) => bundle.write(writeOptions));
+}
```
DevVersion: This file seems to be added accidentally from rebasing?

src/lib/core/media-query/index.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+ import {NgModule} from '@angular/core';
```
ocarreterom: Remove whitespace :)

src/lib/core/media-query/media-query-manager.ts
```diff
@@ -0,0 +1,123 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+ import {Injectable, NgZone} from '@angular/core';
```
ocarreterom: Remove whitespace :)

PR #2219, opened 2016-12-14
fix(snack-bar): clean up element when associated viewContainer is destroyed

### LINKED ISSUE #2190: MdSnackBar does not close when MdDialog is closed

Bug

MdSnackBar should be closed automatically when it's host MdDialog is destroyed or should be closable in ngOnDestroy.

MdSnackBar remains open even after it's host MdDialog is destroyed. The SnackBar will be unresponsive with the buttons unclickable.

1. Open an MdSnackBar from an MdDialog.
2. Close the MdDialog 

Plunker with the behavior: http://plnkr.co/edit/xrf3OSZec4J1w2MncFbY?p=preview

Better user and developer experience

Angular 2, Material 2

crisbeto:
Fixes the snack bar not being removed from the DOM when it's associated `viewContainerRef` gets destroyed.

Fixes #2190.

### DIFF
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -33,16 +33,13 @@ export class MdSnackBarRef<T> {
     this.containerInstance = containerInstance;
     // Dismiss snackbar on action.
     this.onAction().subscribe(() => this.dismiss());
+    containerInstance._onExit().subscribe(() => this._cleanup());
   }
 
   /** Dismisses the snack bar. */
   dismiss(): void {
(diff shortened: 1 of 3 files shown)


src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -62,6 +59,13 @@ export class MdSnackBarRef<T> {
     }
   }
 
+  /** Cleans up the DOM after closing. */
+  private _cleanup(): void {
```
jelbourn: Let's call this `_finishDismiss`

Issue #2171, opened 2016-12-11
hasV8BreakIterator breaks in environments without window (such as universal rendering)


timvdalen:
#### Bug, feature request, or proposal:
Bug: As of https://github.com/angular/material2/pull/1950, `@angular/material` no longer works with server-side rendering because https://github.com/angular/material2/commit/4b7e52d30aa342c3024ed078837a9bd4ca9e773b#diff-6d0e189ed8a75860045a39fe4c576374R7 expects window to be defined.

#### What is the expected behavior?
No errors when loading the module

#### What is the current behavior?
```
node_modules\@angular\material\material.umd.js:2028
var hasV8BreakIterator = (window.Intl && window.Intl.v8BreakIterator);
                                     ^

ReferenceError: window is not defined
```

#### What are the steps to reproduce?
Loading the module in a https://github.com/angular/universal project

#### What is the use-case or motivation for changing an existing behavior?
PR https://github.com/angular/material2/pull/1950 broke compatibility with node

#### Which versions of Angular, Material, OS, browsers are affected?
Running on node

#### Is there anything else we should know?
As a temporary fix, changing the line to `const hasV8BreakIterator = ((typeof window !== 'undefined') && window.Intl && (window.Intl as any).v8BreakIterator);` seems to work, but I'm not sure what that means for this platform check since node does have th v8 break iterator.

/cc @sattip

DevVersion:
I think this issue is more complicated than it looks.

The `MdPlatform` service also uses the `navigator` object to detect the different platforms.
> This will lead to a Runtime Exception in Universal.

![](https://i.gyazo.com/f48080868f6e0fbe12774cb183aa2f77.png)

cc. @jelbourn

timvdalen:
Ah, we haven't run in to that since we're only using `MdDialog` and it's never rendered on the server side.
For our use case, `hasV8BreakIterator` is the only thing that's reached that expects a browser environment

DevVersion:
Ah I see - I will take a look at this as soon as possible.

DevVersion:
Just discussed this with the team. We are going to track this issue with #308

PR #2221, opened 2016-12-14
Remove assumption of `window' in MdPlatform

timvdalen:
Makes it work when loaded in node

See https://github.com/angular/material2/issues/2171
https://github.com/angular/material2/issues/308#issuecomment-267124520

If you would rather see this rewritten with proper node detection instead of this workaround to make it load, that would be fine with me as well.


jelbourn:
LGTM

PR #2226, opened 2016-12-14
chore: disable color-constrast check for axe

jelbourn:
Disabling this check because the colors will vary based on the theme and because it's only failing _sometimes_ on CI right now.


mmalerba:
lgtm

PR #2233, opened 2016-12-15
fix(input): unable to focus input in IE 11

crisbeto:
Fixes being unable to focus an input in IE 11 by clicking on the label. This was due to the style for hiding the placeholder which, for some reason, ended up disabling the input altogether if it didn't have a value.


jelbourn:
LGTM

PR #2235, opened 2016-12-15
fix(input): character count hint for input

Codebacca:
fix (input): Remove the property CharacterCounter on md-input. Add name
of local variable on md-input and add md-hint to fix counter binding of
md-input for the show hint lable with charater count example.

The property characterCounter does not exist on md-input and therefore
the character count was not being displayed for the show hint lable with
charater count example. Removing this and adding the name
#characterCountHintExampleInput along with the md-hint tag using the
name and the character count property fixes the issue.

### DIFF
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -111,7 +111,9 @@
     </p>
     <p>
       <md-input placeholder="Show Hint Label With Character Count" style="width: 100%"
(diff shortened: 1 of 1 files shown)


dahaupt:
@jelbourn Is this really merge ready? He should rebase first because of the deprecated md-input tag. Sorry if I am wrong about this.

jelbourn:
GitHub says merge base is clean. This demo will go away before too long, but in the meantime the change is still correct

PR #2240, opened 2016-12-15
chore: log messages while waiting for ci modes to finish

DevVersion:
* Currently the Travis CI leader mode can timeout, because the `e2e modes` runs longer than 10 minutes and the leader mode then needs to wait the same time (just without any output).
  Travis stops the build automatically when there is no output for the last 10 minutes.

* The plugin `travis-after-modes` has been updated to log a message (like a progress-bar) each time the modes are checked again.

@jelbourn This should fix the timeout issues for the `lint` mode you mentioned on Slack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "strip-ansi": "^3.0.0",
     "stylelint": "^7.5.0",
     "symlink-or-copy": "^1.0.1",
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

PR #2241, opened 2016-12-15
create a temporary home for examples in the demo app

mmalerba:


src/demo-app/examples/slider/simple/simple-slider-example.ts
```diff
@@ -0,0 +1,9 @@
+import {Component} from '@angular/core';
+
+
+@Component({
+  moduleId: module.id,
+  selector: 'simple-slider-example',
+  templateUrl: 'simple-slider-example.html',
+})
+export class SimpleSliderExample {}
```
jelbourn: How about `SliderOverviewExample`?

mmalerba: Done

mmalerba:
removing merge ready label since i'm adding some more examples

src/demo-app/examples/card/fancy/card-fancy-example.html
```diff
@@ -0,0 +1,19 @@
+<md-card>
+  <md-card-header>
+    <div md-card-avatar class="header-image"></div>
+    <md-card-title>Shiba Inu</md-card-title>
+    <md-card-subtitle>Dog Breed</md-card-subtitle>
+  </md-card-header>
+  <img md-card-image src="https://upload.wikimedia.org/wikipedia/commons/7/74/Shiba_Inu_-lying.jpg">
```
jelbourn: Not sure sourcing assets from wikimedia directly like this is allowed (especially without attribution)

src/demo-app/examples/card/fancy/card-fancy-example.scss
```diff
@@ -0,0 +1,8 @@
+md-card {
```
jelbourn: Can you make any styles applied to the component done via css class that is prefixed with `example-`? In material 1, people often had difficulty telling the difference between the example styles and the component styles.

src/demo-app/examples/card/fancy/card-fancy-example.html
mmalerba: ok, where can i get images from?

jelbourn:
Could you also restructure the directory so that it looks like
```
examples
   button-overview
   button-types
   checkbox-overview
   checkbox-configurable
   etc
```
?
That should make them easier to load by string key

jelbourn:
Another thought- could you change the `.scss` styles to just `.css`? I want to make sure we highlight that sass isn't required to use material.

src/demo-app/examples/card/fancy/card-fancy-example.html
mmalerba: I downloaded the images so we can host them ourselves (though I don't know how to hook that up in the demo app, just including them so we can transfer them to  the docs site). Both images are public domain so using them should be fine.

https://commons.wikimedia.org/wiki/File:Shiba_inu_taiki.jpg
https://commons.wikimedia.org/wiki/File:Shiba_Inu.jpg

mmalerba:
wound up adding this directly to material.angular.io instead

PR #2242, opened 2016-12-15
feat(chips): Add (select), [color] and dark theme to chips.

topherfangio:
Add new functionality/options to chips for managing selection/color.

*MdChipList Options*

 - `[selectable]` - Programmatically control whether or not the chips
   in the list are capable of being selected.
 - The SPACE key will automatically select the currently focused chip.

*MdChip Options*

 - `[color]` - Programmatically control the selected color of the
   chip.
 - `[selected]` - Programmatically control whether or not the chip
   is selected.
 - `(select)` - Event emitted when the chip is selected.
 - `(deselect)` - Event emitted when the chip is deselected.

Additionally, adds basic support for dark themeed chips using existing
colors from other components in the spec and cleanup demos by using
cards and toolbars like other demos.

References #120.

---
screen shot 2016-12-15 at 2 39 01 pm

### DIFF
--- a/src/lib/chips/_chips-theme.scss
+++ b/src/lib/chips/_chips-theme.scss
@@ -1,3 +1,4 @@
+@import '../core/theming/palette';
 @import '../core/theming/theming';
 
 @mixin md-chips-theme($theme) {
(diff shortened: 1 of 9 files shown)


src/lib/chips/_chips-theme.scss
```diff
@@ -6,27 +7,36 @@
   $accent: map-get($theme, accent);
   $warn: map-get($theme, warn);
   $background: map-get($theme, background);
+  $foreground: map-get($theme, foreground);
+
+  $unselected-background: if($is-dark-theme, md-color($background, card), #e0e0e0);
+  $unselected-foreground: if($is-dark-theme, md-color($foreground, text), rgba(0, 0, 0, 0.87));
```
jelbourn: Why are these not just
```scss
$unselected-background: md-color($background, card);
$unselected-foreground: md-color($foreground, text);
// etc.
```
?

The background and foreground colors are already based on whether it's a light/dark theme. Is there some reason the light theme version of these colors aren't appropriate?

topherfangio:
@jelbourn Requested changes made.

**Regarding the left/right arrow keys:** I plan to add that at a later date, but many screen readers have difficulty with left/right which can cause problems (and did so with the md1 version). I'd like to do some testing before we add it so that I can work around any problems.

topherfangio:
FYI - I introduced a small bug in my last commit, so I added a new test when I fixed it. Just checks for proper CSS classes when the chip is selected.

PR #2244, opened 2016-12-15
refactor: convert dash-case inputs to camelCase

crisbeto:
Converts any dash-cased `@Input` properties into camelCase and adds a deprecated proxy property with the old naming. The following properties have been renamed:

* `md-ripple-trigger`
* `md-ripple-centered`
* `md-ripple-disabled`
* `md-ripple-max-radius`
* `md-ripple-speed-factor`
* `md-ripple-color`
* `md-ripple-background-color`
* `md-ripple-focused`
* `md-ripple-unbounded`
* `thumb-label`
* `tick-interval`
* `md-dynamic-height`
* `tooltip-position`

The following properties were skipped:
* `md-menu-trigger-for` - Is currently used also as a selector for the menu trigger.
* Any properties that have a native equivalent (e.g. `aria-label`, `aria-labeledby`).

### DIFF
--- a/src/demo-app/menu/menu-demo.html
+++ b/src/demo-app/menu/menu-demo.html
@@ -3,7 +3,7 @@
     <p>You clicked on: {{ selected }}</p>
 
     <md-toolbar>
-      <button md-icon-button [md-menu-trigger-for]="menu" aria-label="Open basic menu">
+      <button md-icon-button [mdMenuTriggerFor]="menu" aria-label="Open basic menu">
         <md-icon>more_vert</md-icon>
       </button>
     </md-toolbar>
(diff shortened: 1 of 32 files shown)


jelbourn:
@crisbeto I just now finished a long ~~debate~~ chat with Misko and Igor, the outcome of which is that `@Directive` properties should also be _prefixed_. (and `@Component` properties are not). Based on that, could you also change:
* `md-dynamic-height` to just `dynamicHeight`
* `tooltipPosition` to `mdTooltipPosition`
* `tooltipShowDelay` to `mdTooltipShowDelay`
* `tooltipHideDelay` to `mdTooltipHideDelay`
* `md-tooltip` to `mdTooltip` (adding `mdTooltip` to the existing selector)
* `md-menu-trigger-for` to `mdMenuTriggerFor` (adding `mdMenuTriggerFor` to the existing selector)

jelbourn:
Also wanted to make sure I mentioned: properties themselves don't have to have an `md` prefix. For example, we can still have
```ts
@Input('mdTooltipPosition')
position: TooltipPosition;
```

fxck:
isn't renaming `@Input()`s another thing styleguide advice against?

jelbourn:
@fxck we may be revisiting that particular piece of advice. Much of the style guide was written speculatively.

crisbeto:
Applied the feedback and rebased @jelbourn.

jelbourn:
LGTM, awesome

Issue #2163, opened 2016-12-10
md-hint does not expand padding for md-input around it.


davidwilcox:
#### Bug, feature request, or proposal:
The hint inside of an md-input overlaps with the next element on the page. This looks pretty bad.

#### What is the expected behavior?
That the hint not overlap with the next element.

#### What is the current behavior?
The hint overlaps with the next element.

#### What are the steps to reproduce?
Add an md-input with an md-hint on the inside. The next line will be within the md-hint that should be inside of the md-input. It overlaps on the page.
http://plnkr.co/edit/a3070WC9yA6Xp3id15Kb

#### What is the use-case or motivation for changing an existing behavior?
So that my website looks better. It can't be normal behavior to overlap the md-input hint with the following element which it has nothing to do with.

#### Which versions of Angular, Material, OS, browsers are affected?
Chrome Version 53.0.2785.143 (64-bit). Ubuntu 16.10. @angular/material@2.0.0-alpha.10. @angular/core@^2.1.0.

#### Is there anything else we should know?
Nope!

mmalerba:
I think changing the size of the element when the hint is shown is just going to create more layout trouble for developers than it will fix. So we either have to live with `md-input` that have a little too much bottom space when there's no hint or `md-input` that don't quite have enough bottom space when there is a hint. 

@jelbourn do you have a preference for which way we go on this? I could see maybe adding a bit more margin (also making the margin `em` based rather than `px` based so it works for different font sizes).

@davidwilcox in the meantime you can easily add some extra margin to all of the `md-input` on your site like this:
```
md-input {
  margin-bottom: 0.5em;
}
```

jelbourn:
@mmalerba I would lean towards the component _always_ leaving room for the hint whether it's there or not. We could potentially add an attribute to remove it.

mmalerba:
yeah that's what I was thinking, sg

davidwilcox:
Just curious... What was the fix here?

PR #2246, opened 2016-12-16
fix(input): add more padding so that the hint doesn't overflow the container

mmalerba:
also use em instead of px for some things so things don't overflow when using different font sizes.


mmalerba:
fixes #2163

jelbourn:
LGTM

This may cause issues for Google apps if it changes the height of the input. We should check with a couple of projects before merging.

PR #2252, opened 2016-12-16
docs: fill out jsdoc for button, checkbox, and radio

jelbourn:
I also renamed the radio's `align` values to `before` | `after` instead of `start` | `end` to be consistent with our naming scheme. I checked, and nobody inside Google is using those properties.


mmalerba:
lgtm

PR #2253, opened 2016-12-16
fix cryptic error when md-input-container has no md-input

mmalerba:


jelbourn:
LGTM, though FYI TypeScript in 2.1+ you can't subclass `Error` so we'll eventually have to change these.

PR #2256, opened 2016-12-16
docs(input): update docs to describe md-input-container rather than md-input

mmalerba:


jelbourn:
LGTM

arlowhite:
Thanks for the hard work on angular/material2

Just wanted to mention that it's confusing when the GitHub documentation does not match the latest release in npm. Maybe change the md-input link in the root README.md to the last release branch? But this might be a mess to maintain and confuse/frustrate developers working on master.

What if you set the default GitHub branch to the last release? That way users browsing GitHub as their documentation source are looking at a snapshot of the documentation matching the latest released package.

Or just when you make API changes, such as this md-input change, put a version note in the new API section(s) on that Component's page:
> Requires @angular/material2 2.0.0-alpha.12 or greater

PR #2258, opened 2016-12-16
refactor: change the toolkit directive prefixes and remove Md prefix from toolkit classes

crisbeto:
Renames the the component toolkit to use the `cdk-` prefix, in addition to removing the `Md` prefix from classes associated with the toolkit.
The following CSS classes have been renamed as well:

* `md-global-overlay-wrapper` -> `cdk-global-overlay-wrapper`
* `md-overlay-container` -> `cdk-overlay-container`
* `md-overlay-pane` -> `cdk-overlay-pane`
* `md-overlay-backdrop-showing` -> `cdk-overlay-backdrop-showing`
* `md-overlay-dark-backdrop` -> `cdk-overlay-dark-backdrop`
* `md-overlay-backdrop` -> `cdk-overlay-backdrop`
* `md-visually-hidden` -> `cdk-visually-hidden`

### DIFF
--- a/e2e/components/dialog/dialog.e2e.ts
+++ b/e2e/components/dialog/dialog.e2e.ts
@@ -91,7 +91,7 @@ describe('dialog', () => {
     browser.actions()
       // We need to move the cursor to the top left so
       // the dialog doesn't receive the click accidentally.
-      .mouseMove(element(by.css('.md-overlay-backdrop')).getWebElement(), { x: 0, y: 0 })
+      .mouseMove(element(by.css('.cdk-overlay-backdrop')).getWebElement(), { x: 0, y: 0 })
       .click()
       .perform();
   }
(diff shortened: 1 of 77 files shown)


crisbeto:
@jelbourn I still have some doubts:

1. Should I move over the `md-line` and `md-ripple`? They seem a bit Material-specific.
2. Should I rename the SASS variables that are associated with the `cdk` styles?

jelbourn:
@crisbeto `md-line` and `md-ripple` are material specific so they stay the same. It's only things that are totally decoupled from Material Design that get `cdk`. 

It would be good to rename the associated scss variables, but that's not as important.

crisbeto:
Renamed the SASS variables and removed the overlay's dependency on theming @jelbourn. This should be good to go.

PR #2259, opened 2016-12-16
feat(core): add scrollable view properties to connected pos strategy

andrewseguin:
Uses new Scrollable changes from #2188. Adds a set of properties to the change position event from the connected position strategy.

Still working on writing the tests for the connected position strategy but wanted to put this out for review to get it looked at ahead of time since beta is on Tuesday and we want this in there.


andrewseguin:
Removed its usage in tooltip and sidenav until we optimize the number of scroll notifications with a throttle

PR #2262, opened 2016-12-17
chore: minify html and css for releases

### LINKED ISSUE #2066: Production release uses minified html and css

DevVersion:
* For releases, the HTML and CSS assets will be minified before being inlined into the built JS

@jelbourn Basically as mentioned on Slack, I had to decouple the `:rollup` task from the `inlining`, so we can control the inlining of the resources.

Closes #2066


DevVersion:
@jelbourn Made the requested changes.

PR #2264, opened 2016-12-17
chore(pre-push): added pre-push hook to prevent lint errors

EladBezalel:
Hooks the pre-push in git and runs `lint`, this would prevent ci failures

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "deploy": "firebase deploy",
     "webdriver-manager": "webdriver-manager",
     "docs": "gulp docs",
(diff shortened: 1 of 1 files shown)


crisbeto:
Do we really need this? It's the purpose of the CI to fail in these cases.

EladBezalel:
I just think that when you can't push with lint errors you would never have to juggle between prs to fix lint errors that you didn't noticed and pushed

crisbeto:
You can set up a linter in your IDE. Otherwise we might as well run the unit and e2e in a hook too.

PR #2266, opened 2016-12-18
feat(dialog): add a config option for passing in data

### LINKED ISSUE #2181: MdDialog communication (i.e: passing data to MdDialog)
Pass data to a component opened by the dialog service **without** depending on the dialog instance.

You can pass data to a component by accessing it's instance and setting a predefined property.

There are several ways to communicate with a dialog component:
  1) Custom Dependency injection - Set Providers as part of the options
  2) Fixed Dependency injection - Set a single data item as part of the options, injected via a token.
  3) `MdDialogRef` as a carrier -  Set a single data item as part of the options, component consumed via MdDialogRef.data.
  4) Direct assignment - Current behavior

The current behavior allow (4) and a limited (1) 
> (1) Only predefined providers.

**It does not fit the angular pattern.** 

Angular uses DI a lot, this is part of the flow.

  - Automatic errors if DI mismatch 
  - Allows proper use of hooks 
  - Less dependant

Since the dialog service abstract the component instantiation there is no control over the providers. 
However, this process does allow changing providers by supplying an injector...

So all in all I think it should be possible to get some sort of control over the DI.

### Workaround:
My current workaround is to create a service wrapping MdDialog.
This services uses a wrapper component as the Component to create.
The wrapper component then create the original component inside it with the supplied providers.

crisbeto:
Adds the ability to pass in data to a dialog via the `data` property. While this was previously possible through the `dialogRef.componentInstance.someUserSuppliedProperty`, it wasn't the most intuitive. The new approach looks like this:

```
dialog.open(SomeDialog, {
  data: {
    hello: 'world'
  }
});

class SometDialog {
  constructor(data: MdDialogData) {
    console.log(data['hello']);
  }
}
```

Fixes #2181.

### DIFF
--- a/src/demo-app/dialog/dialog-demo.html
+++ b/src/demo-app/dialog/dialog-demo.html
@@ -46,7 +46,15 @@
       </md-select>
     </p>
 
-    <md-checkbox [(ngModel)]="config.disableClose">Disable close</md-checkbox>
+    <p>
+      <md-input-container>
+        <input mdInput [(ngModel)]="config.data.message" placeholder="Dialog message">
+      </md-input-container>
+    </p>
+
+    <p>
+      <md-checkbox [(ngModel)]="config.disableClose">Disable close</md-checkbox>
+    </p>
   </md-card-content>
 </md-card>
 
--- a/src/demo-app/dialog/dialog-demo.ts
+++ b/src/demo-app/dialog/dialog-demo.ts
@@ -1,4 +1,5 @@
 import {Component, Inject} from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
 import {MdDialog, MdDialogRef, MdDialogConfig, MD_DIALOG_DATA} from '@angular/material';
 
 
--- a/src/demo-app/dialog/dialog-demo.ts
+++ b/src/demo-app/dialog/dialog-demo.ts
@@ -1,6 +1,7 @@
 import {Component, Inject} from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
-import {MdDialog, MdDialogRef, MdDialogConfig} from '@angular/material';
+import {MdDialog, MdDialogRef, MdDialogConfig, MdDialogData} from '@angular/material';
+
 
 @Component({
   moduleId: module.id,
@@ -21,6 +22,9 @@ export class DialogDemo {
       bottom: '',
       left: '',
       right: ''
+    },
+    data: {
+      message: 'Jazzy jazz jazz'
     }
   };
 
@@ -41,7 +45,7 @@ export class DialogDemo {
   openJazz() {
     this.dialogRef = this.dialog.open(JazzDialog, this.config);
 
-    this.dialogRef.afterClosed().subscribe(result => {
+    this.dialogRef.afterClosed().subscribe((result: string) => {
       this.lastCloseResult = result;
       this.dialogRef = null;
     });
@@ -59,13 +63,13 @@ export class DialogDemo {
   template: `
   <p>It's Jazz!</p>
   <p><label>How much? <input #howMuch></label></p>
-  <p> {{ jazzMessage }} </p>
+  <p> {{ data.message }} </p>
   <button type="button" (click)="dialogRef.close(howMuch.value)">Close dialog</button>`
 })
 export class JazzDialog {
-  jazzMessage = 'Jazzy jazz jazz';
-
-  constructor(public dialogRef: MdDialogRef<JazzDialog>) { }
+  constructor(
+    public dialogRef: MdDialogRef<JazzDialog>,
+    public data: MdDialogData) { }
 }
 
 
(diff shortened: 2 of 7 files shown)


crisbeto:
@jelbourn I'm not huge fan of having to access the properties on the `data` object through bracket notation, but I'm not sure how we could it better. Of course, people can still cast it to the proper type.

clydin:
An option could be to have the `MdDialogRef` contain the `data` property with an added second generic parameter to type the property.  `open` would need the generic type as well.

shlomiassaf:
@crisbeto @jelbourn This is not what I suggested in #2181 

Let the user pass providers and you add them into the injector.

When you instantiate the component the user has provided you already add something to the injector the `MdDialogRef` since it is unique for each dialog opened.

```ts
dialog.open(SomeDialog, {
  providers : [
    { provider: MyPreciousToken, useValue: 'I'm the token master' }
  ]
});
```

You can of course use `@Injectable`s there as well, not just tokens.
This is quite common, and supports all behaviours an angular user should know... 

The current PR doesn't change much, you still don't get natural type support, if you want type support you need to do:
```ts
class SometDialog {
  constructor(@Inject(MdDialogData) data: MyHelloDataModel) {
    console.log(data.hello);
  }
}
```

If you insist on going that road I suggest changing `MdDialogRef` to `MdDialogRef<T>`, this way we don't need the extra token and `MdDialogRef` will carry the data, the `data` property will be of type T... 

But again, i'm pretty sure once more people use it that you will get a request for provider injection (or injector replacement) since the user can't control it. 

I have a modal library for angular from alpha 17, a lot of people use the provider injection.

jelbourn:
@shlomiassaf `MdDialogRef` is already a generic, where `T` is the type of the component inside the dialog.  We could potentially change it to `MdDialogRef<T, D>`, where `D` is the type of the passed in data, but that then begs the questions- what about the return value? Then we'd have `MdDialogRef<T, D, R>`, which I think everyone would find rather cumbersome.

jelbourn:
@crisbeto WRT typings,

We could make `MdDialogData` generic with exactly one property, e.g.,
```ts
export class MdDialogData<T> {
  value: T;
}
```
People could then inject, say, `MdDialogData<User>`, pull out the `value` and go about their day.  However, `MdDialogData` doesn't really do anything here other than act as an injection token for the data- so why not just make an injection token?

```ts
// In dialog-injector.ts
export const MD_DIALOG_DATA = new OpaqueToken('MdDialogData');
...
  get(token: any, notFoundValue?: any): any {
    if (token === MdDialogRef) {
      return this._dialogRef;
    } else if (token == MD_DIALOG_DATA) {
      return config.data;
    }

    return this._parentInjector.get(token, notFoundValue);
  }

```

Then the user would inject the data:
```ts
@Component({...})
export class UserProfileDialog(@Inject(MD_DIALOG_DATA) user: User) {
  // ...
}
```

The `data` on the `MdDialogConfig` would stay as an `any` since there's no way we could enforce the type through injection anyway.

crisbeto:
Yeah, the `OpaqueToken` option was what I was going for initially, but didn't figure out how to do properly. I'll rework the PR.

crisbeto:
@jelbourn Re-did this to use the `OpaqueToken`.

src/lib/dialog/README.md
```diff
@@ -13,12 +13,13 @@ MdDialog is a service, which opens dialogs components in the view.
...
 | --- | ------------ |
-| `role: DialogRole = 'dialog'` | The ARIA role of the dialog element. Possible values are `dialog` and `alertdialog`. Optional. |
-| `disableClose: boolean = false` | Whether to prevent the user from closing a dialog by clicking on the backdrop or pressing escape. Optional. |
-| `width: string = ''` | Width of the dialog. Takes any valid CSS value. Optional. |
-| `height: string = ''` | Height of the dialog. Takes any valid CSS value. Optional. |
-| `position: { top?: string, bottom?: string, left?: string, right?: string }` | Position of the dialog that overrides the default centering in it's axis. Optional. |
-| `viewContainerRef: ViewContainerRef` | The view container ref to attach the dialog to. Optional. |
+| `role?: DialogRole = 'dialog'` | The ARIA role of the dialog element. Possible values are `dialog` and `alertdialog`. |
+| `disableClose?: boolean = false` | Whether to prevent the user from closing a dialog by clicking on the backdrop or pressing escape. |
+| `data?: { [key: string]: any }` | Data that will be injected into the dialog as via the `MD_DIALOG_DATA` token. |
```
jelbourn: The type of `data` should be `any`

crisbeto:
This should be good to go. The test failure is due to https://github.com/angular/material2/issues/2903

sum0mer:
Hi, I tried to 'import {MdDialog, MdDialogRef, MdDialogData} from '@angular/material';', but got [workspace/angular2/node_modules/@angular/material/index"' has no exported member 'MdDialogData'] compile error. Any suggestion to make it work? Thanks!

crisbeto:
We didn't end up using `MdDialogData` @sum0mer. You can see an example of how it should be used in the demo app: https://github.com/angular/material2/blob/master/src/demo-app/dialog/dialog-demo.ts#L80

PR #2268, opened 2016-12-18
chore: update stylelint version

belev:
Update the version of `stylelint` to remove `Can not parse selector` messages.

Previously discussed in #2037.


belev:
Yeah, me too. Made a PR which was rejected but today when I see that there is a new version of stylelint I decided to make a new PR to fix this messages.

jelbourn:
LGTM

PR #2276, opened 2016-12-19
fix(overlay): proper backdrop stacking with multiple overlays

### LINKED ISSUE #2272: md-select overlay not dismissed inside dialog
bug

if you open md-select inside md-dialog component and click somewhere outside of md-select overlay (but inside of dialog itself), the md-select overlay should be closed and onClose event should be fired

now if you click inside dialog outside of md-select overlay, overlay is not closed. but if you click outside dialog overlay, it is closed 

http://plnkr.co/edit/oIJQsOZVt0aKRFPe9rde

actually it is never dismissed in plunker, and what is worse, the overlay of md-select is shown below dialog overlay. But in real world application I get the behaviour described above - md-select is dismissed when clicking outside overlay, but not if it is within dialog

angular 2.2.4, material 2.0.0-alpha.11-3, ubuntu 14.04, Firefox 50.1, Chrome 55

crisbeto:
Currently backdrops get inserted after their corresponding overlays in the DOM. This can lead to situations where another overlay that is technically lower in the stacking order could go above a backdrop (e.g. opening a `select` inside a `dialog`). These changes switch to doing the stacking by having the overlay and backdrop have the same `z-index` and determining the stacking order by the order of the elements in the DOM.

Fixes #2272.

**Note:** I'm expecting this to have some merge conflicts with #2258.

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -101,13 +101,13 @@ export class OverlayRef implements PortalHost {
     this._backdropElement.classList.add('cdk-overlay-backdrop');
     this._backdropElement.classList.add(this._state.backdropClass);
 
-    this._pane.parentElement.appendChild(this._backdropElement);
+    // Insert the backdrop before the pane in the DOM order,
+    // in order to handle stacked overlays properly.
+    this._pane.parentElement.insertBefore(this._backdropElement, this._pane);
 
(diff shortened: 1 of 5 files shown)


jelbourn:
LGTM, I'm going to try to get that renaming PR in first

crisbeto:
Seems like some of the tests started failing after I rebased from master. Investigating.

crisbeto:
Alright, it's sorted out. I had to rename the selector to use the `cdk-` prefix.

PR #2280, opened 2016-12-19
feat: make hammerjs optional

crisbeto:
Makes HammerJS and logs a warning if it's missing, instead of crashing the app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,10 +34,12 @@
     "@angular/platform-browser-dynamic": "^2.2.0",
     "@angular/router": "^3.2.0",
     "core-js": "^2.4.1",
-    "hammerjs": "^2.0.8",
     "rxjs": "5.0.0-beta.12",
     "systemjs": "0.19.38",
     "zone.js": "^0.6.23"
+  },
+  "optionalDependencies": {
+    "hammerjs": "^2.0.8"
   },
   "devDependencies": {
     "@angular/compiler-cli": "^2.2.0",
--- a/src/demo-app/tsconfig.json
+++ b/src/demo-app/tsconfig.json
@@ -21,10 +21,7 @@
       "@angular/material": [
         "../../dist/@angular/material"
       ]
-    },
-    "types": [
-      "hammerjs"
-    ]
+    }
   },
   "angularCompilerOptions": {
     "genDir": "../../dist",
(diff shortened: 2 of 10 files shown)


src/lib/core/gestures/MdGestureConfig.ts
```diff
@@ -4,16 +4,25 @@ import {HammerGestureConfig} from '@angular/platform-browser';
...
     'slideend',
     'slideright',
     'slideleft'
-  ];
+  ] : [];
+
+  constructor() {
+    super();
+
+    if (!this._hammer) {
```
jelbourn: Can you import `isDevMode` from `@angular/core` and then only log this warning if `isDevMode()` is true?

src/lib/core/gestures/MdGestureConfig.ts
```diff
@@ -4,16 +4,25 @@ import {HammerGestureConfig} from '@angular/platform-browser';
 /* Adjusts configuration of our gesture library, Hammer. */
 @Injectable()
 export class MdGestureConfig extends HammerGestureConfig {
+  private _hammer = typeof window !== 'undefined' ? (window as any).Hammer : null;
```
jelbourn: Rather than making it `any`, could you add a `GestureManager` interface that just copies the members of `HammerManager` that we're using?

crisbeto: I went with `any` here and a bit below, because it throws a TS compile error if HammerJS is missing.

src/lib/core/gestures/MdGestureConfig.ts
crisbeto: Done.

src/lib/core/gestures/MdGestureConfig.ts
jelbourn: Right- instead of dropping all the way down to `any`, though, we can introduce out own interface and say `(window as any).Hammer as GestureManager`

crisbeto: Makes sense, I'll look into the API.

crisbeto:
Updated @jelbourn. I've replaced the HammerJS interfaces with our own stripped-down alternatives.

jelbourn:
LGTM

PR #2286, opened 2016-12-19
feat(layout-direction): a service to accompany the Dir directive

### LINKED ISSUE #1719: Introduce [ltr, rtl] layout direction service
Similar to what the Dir directive does, but will also capture "dir" attributes outside the Angular app. 

cc: @jelbourn

EladBezalel:
- Finds the application root element and traverses upwards and looking for dir attribute
- If Dir directive is available throughout the application it will be injected instead the activeLayoutDirection service

fixes #1719

### DIFF
--- a/src/lib/core/rtl/active-layout-direction.ts
+++ b/src/lib/core/rtl/active-layout-direction.ts
@@ -1,0 +1,8 @@
+import {LayoutDirection} from './dir';
+import {Injector} from '@angular/core';
+import {Observable} from 'rxjs/Observable';
+
+export class ActiveLayoutDirection {
+  dirChange: Observable<any>;
+  getActiveDirection(injector: Injector): Promise<LayoutDirection> { return null; }
+}
--- a/src/lib/core/core.ts
+++ b/src/lib/core/core.ts
@@ -1,6 +1,6 @@
 import {NgModule, ModuleWithProviders} from '@angular/core';
 import {MdLineModule} from './line/line';
-import {RtlModule} from './rtl/dir';
+import {RtlModule} from './rtl/index';
 import {ObserveContentModule} from './observe-content/observe-content';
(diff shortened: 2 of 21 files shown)


src/lib/core/rtl/active-layout-direction.ts
```diff
@@ -0,0 +1,52 @@
...
+  ElementRef,
+  ApplicationRef,
+  Injector
+} from '@angular/core';
+
+import {LayoutDirection} from './dir';
+import {EventEmitter} from '@angular/common/src/facade/async';
+
+@Injectable()
+export class ActiveLayoutDirection {
```
jelbourn: Rename this class to something like `GlobalDirAccessor` and then add another class called `ActiveLayoutDirection` that just looks like
```ts
export class ActiveLayoutDirection {
  dirChange: Observable<void>;
  getActiveDirection(injector: Injector): Promise<LayoutDirection> { return null; }
}
```

The both `Dir` and `GlobalDirAccessor` will have `implements ActiveLayoutDirection` (TypeScript lets you `implement` a class just like an interface. Keeping it as a class lets us use it as an injection token).

src/lib/core/rtl/active-layout-direction.ts
```diff
@@ -0,0 +1,52 @@
+import {
+  Injectable,
+  ElementRef,
+  ApplicationRef,
+  Injector
+} from '@angular/core';
+
+import {LayoutDirection} from './dir';
+import {EventEmitter} from '@angular/common/src/facade/async';
```
EladBezalel: Webstorm auto import 😅

jelbourn:
@EladBezalel was this ready for another round of review?

EladBezalel:
nope 😅

EladBezalel:
A better solution was suggested on #3600

PR #2289, opened 2016-12-20
feat(checkbox, radio): change `align` to `labelPosition` (inverted)

jelbourn:
The `align` property on checkbox and radio is confusing because it's talking about the actual toggle, but really deals with where the label goes. This makes it much more explicitly by replacing `align = 'start' | 'end'` with `labelPosition = 'before' | 'after'`. 

Temporarily keeps the old API around as deprecated.


mmalerba:
lgtm, much better name imo

PR #2292, opened 2016-12-20
chore: fix minified templates in release

DevVersion:
* Currently the `html-minifier` minifies the attributes in a non-case sensitive way and this breaks the templates in the releases.

@jelbourn This one should be **very** important for releases, because right now the templates in the releases are broken.


DevVersion:
@jelbourn Oh my bad. I oversaw the `caseSensitive` property and only added the `removeComments` option.

PR #2293, opened 2016-12-20
chore: package.json updates

crisbeto:
* Moves the `router` and `platform-browser-dynamic` to the `devDependencies` since they're only used in tests and the demo app.
* Removes a duplicate `@types/selenium-webdriver` which probably got into the `package.json` due to a faulty merge.
* Locks down the Selenium typings version in order to prevent Selenium-related type errors.
* Adds `selenium-webdriver` to the dev dependencies, otherwise the e2e tests fail to run on a fresh install.


jelbourn:
LGTM

PR #2297, opened 2016-12-20
chore: update test browsers to latest version

DevVersion:
* Updates the Unit test browsers to the `latest` version on Saucelabs
* No longer records videos & screenshots on the e2e mode, because this seems to slow the builds significant.

### DIFF
--- a/test/protractor.conf.js
+++ b/test/protractor.conf.js
@@ -43,7 +43,11 @@ if (process.env['TRAVIS']) {
     'browserName': 'chrome',
     'tunnel-identifier': process.env['TRAVIS_JOB_NUMBER'],
     'build': process.env['TRAVIS_JOB_NUMBER'],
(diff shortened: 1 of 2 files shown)


test/protractor.conf.js
```diff
@@ -43,7 +43,11 @@ if (process.env['TRAVIS']) {
     'browserName': 'chrome',
     'tunnel-identifier': process.env['TRAVIS_JOB_NUMBER'],
     'build': process.env['TRAVIS_JOB_NUMBER'],
-    'name': 'Material 2 E2E Tests'
+    'name': 'Material 2 E2E Tests',
+
+    // By default Saucelabs tries to record the whole e2e run. This can slow down the builds.
+    'recordVideo': false,
+    'recordScreenshots': false
```
jelbourn: This is okay for now, but eventually we'll want to turn screenshots back on.

PR #2312, opened 2016-12-21
chore: improve dgeni api doc gen

jelbourn:
* Turn on autoescape 
* Break methods and properties into their own template files via macros
* Show docs for services
* Show markers for `@Input` and `Output` with aliases
* Prevent optional method parameters from being shown twice


mmalerba:
lgtm

PR #2314, opened 2016-12-21
chore(icon): remove rxjs import for throw

jelbourn:
Adding this breaks several hundred build targets in google.


mmalerba:
lgtm

PR #2316, opened 2016-12-21
chore: move umd bundles to bundles directory

### LINKED ISSUE #2308: Move umd bundles into `bundles/` dir in npm package

DevVersion:
* Moves the generated `.umd` files to the `bundles/` directory (as @angular/core)
* Uses a gulp plugin for rollup because its more clear and also allows us to easily minify & create sourcemaps in the future

**Note**: After this PR, sourcemaps are no longer generated for the umd bundles.
*  This means that we also don't have sourcemaps for the demo-app (but those currently only map to the `.js` files)
* I locally played with it and created a special check to create sourcemaps when being in development mode (but it was kind of overkill and unnecessary as mentioned above)

@jelbourn I was not sure what's the best way to go here. We'd need find a clear way how we want to structure our gulp tasks (because having X similar tasks just for release feels wrong to me)

Closes #2308


jelbourn:
LGTM

PR #2321, opened 2016-12-21
docs: add jsdoc annotations everywhere

crisbeto:
Adds JSDoc annotations to all of the public properties and methods that were missing them. Also cleans up some random issues that were noticed while going through everything.

**Note:** I've skipped the toolbar component since @DevVersion mentioned that he's doing it already.

### DIFF
--- a/src/lib/button-toggle/button-toggle.ts
+++ b/src/lib/button-toggle/button-toggle.ts
@@ -98,7 +98,6 @@ export class MdButtonToggleGroup implements AfterViewInit, ControlValueAccessor 
   @ContentChildren(forwardRef(() => MdButtonToggle))
   _buttonToggles: QueryList<MdButtonToggle> = null;
 
-  /** @docs-private */
   ngAfterViewInit() {
     this._isInitialized = true;
   }
@@ -342,7 +341,6 @@ export class MdButtonToggle implements OnInit {
     }
   }
 
-  /** @docs-private */
   ngOnInit() {
     if (this.id == null) {
       this.id = `md-button-toggle-${_uniqueIdCounter++}`;
--- a/src/lib/button-toggle/button-toggle.ts
+++ b/src/lib/button-toggle/button-toggle.ts
@@ -26,7 +26,7 @@ import {
   DefaultStyleCompatibilityModeModule,
 } from '../core';
 
-
+/** Acceptable types for a button toggle. */
 export type ToggleType = 'checkbox' | 'radio';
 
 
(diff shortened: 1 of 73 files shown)


src/lib/button-toggle/button-toggle.ts
```diff
@@ -98,10 +98,12 @@ export class MdButtonToggleGroup implements AfterViewInit, ControlValueAccessor
   @ContentChildren(forwardRef(() => MdButtonToggle))
   _buttonToggles: QueryList<MdButtonToggle> = null;
 
+  /** @docs-private */
   ngAfterViewInit() {
```
DevVersion: All the lifecycle hooks should be already filtered in [this processor](https://github.com/angular/material2/blob/master/tools/dgeni/processors/docs-private-filter.js#L6-L14)

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -15,8 +16,13 @@ export class InteractivityChecker {
 
   constructor(private _platform: Platform) {}
 
-  /** Gets whether an element is disabled. */
-  isDisabled(element: HTMLElement) {
+  /**
+   * Gets whether an element is disabled.
+   *
+   * @param element Element to be checked.
+   * @returns {boolean} Whether the element is disabled.
```
DevVersion: We don't need the type here

src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -27,16 +33,21 @@ export class InteractivityChecker {
    *
    * This will capture states like `display: none` and `visibility: hidden`, but not things like
    * being clipped by an `overflow: hidden` parent or being outside the viewport.
+   *
+   * @returns {boolean} Whether the element is visible.
```
DevVersion: Type is unnecessary in TS (same for below and everywhere else)

src/lib/core/observe-content/observe-content.ts
```diff
@@ -9,16 +9,22 @@ import {
   AfterContentInit
 } from '@angular/core';
 
+/**
+ * Directive that triggers a callback whenever the content of
+ * it's associated element has changed.
```
DevVersion: Should be `its`?

src/lib/core/a11y/interactivity-checker.ts
crisbeto: We do, dgeni doesn't figure out the return types, it was requested by @jelbourn.

src/lib/core/a11y/interactivity-checker.ts
crisbeto: We do since dgeni doesn't figure out the types.

jelbourn:
LGTM; I'll make any tweaks after merging

PR #2325, opened 2016-12-21
feat(slider): emit input event when slider thumb moves

### LINKED ISSUE #2296: Slider doesn't work with (input) event
Given a simple slider, it does not call "myFunction" subscribed to (input) event :
<md-slider #mySlider min="0" max="100" (input)="myFunction(mySlider.value)"></md-slider>

Then if I subscribe to (change) event it works and it triggers the function at the end of sliding:
<md-slider #mySlider min="0" max="100" (change)="myFunction(mySlider.value)"></md-slider>

The problem is that I need to call "myFunction" each step of slider and this is possible only subscribing to (input) event.

The expected behaviour should be that if I subscribe "myFunction" to (input) event,  "myFunction" get called each step.

At the moment "myFunction" is not triggered

"@angular/material": "2.0.0-alpha.10",

DevVersion:
* Adds an input event output, which fires each time the thumb changes its position.
  Made similar to the native range element `input` event dispatching.

Closes #2296

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -117,8 +117,9 @@ export class MdSlider implements ControlValueAccessor {
 
   private _controlValueAccessorChangeFn: (value: any) => void = () => {};
 
-  /** The last value for which a change event was emitted. */
-  private _lastEmittedValue: number = null;
+  /** The last values for which a change or input event was emitted. */
+  private _lastChangeValue: number = null;
+  private _lastInputValue: number = null;
 
   /** onTouch function registered via registerOnTouch (ControlValueAccessor). */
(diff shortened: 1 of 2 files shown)


src/lib/slider/slider.ts
```diff
@@ -295,6 +296,7 @@ export class MdSlider implements ControlValueAccessor {
   }
 
   @Output() change = new EventEmitter<MdSliderChange>();
+  @Output() input = new EventEmitter<MdSliderChange>();
```
DevVersion: FYI: The JSDoc is intentionally missing here, because otherwise it might collide with #2321

PR #2330, opened 2016-12-21
chore: ensure typescript stays compatible with ngc

DevVersion:
* Ensures that the current `typescript` version is always compatible with the current `@angular/compiler-cli`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "ts-node": "^0.7.3",
     "tslint": "^3.13.0",
     "typedoc": "^0.5.1",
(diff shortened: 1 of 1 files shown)


DevVersion:
I also tried updating protractor, but since we are using the `chord` method in our tests and the typing for it has been accidentally removed in `@types` we need to wait for  https://github.com/DefinitelyTyped/DefinitelyTyped/commit/123690bb2e0f9dac2371084369a8ecf8976cbb34 to be available.

jelbourn:
LGTM

PR #2332, opened 2016-12-21
feat(chips): Add left/right arrow functionality.

topherfangio:
In order to maintain consistency with the md1 chips, add the ability to use the left/right arrow keys to navigate chips.

References #120.


jelbourn:
LGTM

PR #2339, opened 2016-12-22
refactor: convert dash-case md-autosize to camelCase

EladBezalel:


jelbourn:
LGTM, thanks

jelbourn:
@EladBezalel I also just realized that the properties should also be renamed to `mdAutosizeMinRows` and `mdAutosizeMinRows`

EladBezalel:
@jelbourn seems like #2432 already fixed it.. should i just change to `mdAutosizeMinRows` and `mdAutosizeMaxRows`?

jelbourn:
Yeah, that would be good; we'd still need to support the old names for one release as well (via getter / setter)

EladBezalel:
@jelbourn

src/lib/input/autosize.ts
```diff
@@ -17,9 +17,25 @@ export class MdTextareaAutosize implements OnInit {
   /** Minimum number of rows for this textarea. */
   @Input() minRows: number;
 
+  get mdAutosizeMinRows(): number {
+    return this.minRows;
+  }
+
+  @Input() set mdAutosizeMinRows(value: number) {
```
jelbourn: `@Input()` should go on the `getter`. We should add that to the coding standards...

src/lib/input/autosize.ts
```diff
@@ -17,9 +17,25 @@ export class MdTextareaAutosize implements OnInit {
...
+  }
+
   /** Maximum number of rows for this textarea. */
   @Input() maxRows: number;
 
+  get mdAutosizeMaxRows(): number {
+    return this.maxRows;
+  }
+
+  @Input() set mdAutosizeMaxRows(value: number) {
```
jelbourn: The unit test needs to be updated as well.

EladBezalel: do you mean adding another part that checks that the old and the new api works?

jelbourn: The test component in the spec still has `minRows` and `maxRows`. It should use the updated API.

PR #2342, opened 2016-12-22
docs(input): mark <md-input> as deprecated

jelbourn:


crisbeto:
LGTM, I don't know whether the docs processor allows for multiple annotations on the same line though.

jelbourn:
Apparently not; moved to another line

PR #2348, opened 2016-12-22
chore: add license header to umd bundles

DevVersion:
* Adds a license header to the UMD bundles. Similar to @angular core packages.

See https://unpkg.com/@angular/core@2.4.1/bundles/core.umd.js

### DIFF
--- a/tools/gulp/constants.ts
+++ b/tools/gulp/constants.ts
@@ -1,5 +1,7 @@
 import {join} from 'path';
 
+export const MATERIAL_VERSION = require('../../package.json').version;
+
 export const PROJECT_ROOT = join(__dirname, '../..');
 export const SOURCE_ROOT = join(PROJECT_ROOT, 'src');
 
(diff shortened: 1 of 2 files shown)


tools/gulp/constants.ts
```diff
@@ -22,6 +24,12 @@ export const HTML_MINIFIER_OPTIONS = {
   removeAttributeQuotes: false
 };
 
+export const LICENSE_BANNER = `/**
+  * @license Angular Material v${MATERIAL_VERSION}
+  * Copyright (c) 2016 Google, Inc. https://angular.io/
```
crisbeto: Perhaps we should link to material.angular.io instead?

DevVersion: Don't think that would play a big role. Just switched it to `material.angular.io`

jelbourn:
LGTM, thanks!

PR #2350, opened 2016-12-22
chore(ripple): disable problematic test on iOS Safari

crisbeto:
Disables a problematic unit test that is currently failing on iOS Safari. Similar tests have also been disabled in the ViewportRuler.


jelbourn:
LGTM, thanks!

PR #2362, opened 2016-12-23
fix(input): fix placeholder for number input with bad input.

mmalerba:


src/lib/input/input-container.ts
```diff
@@ -124,7 +124,21 @@ export class MdInputDirective implements AfterContentInit {
...
+      // Check if this is an <input> element that contains bad input.
+      // If so, we know that it only appears empty because the value failed to parse.
+      if (this._elementRef.nativeElement instanceof HTMLInputElement &&
+          this._elementRef.nativeElement.validity.badInput) {
+        return false;
+      }
+      return true;
+    }
+    return false;
+  }
```
jelbourn: Seems a bit complex with four different `return` statements. How about:
```ts
return !this._isNeverEmpty() &&
  (this.value == null || this.value === '') &&
  !this._isBadInput();
```

src/lib/input/input-container.ts
```diff
@@ -124,7 +124,21 @@ export class MdInputDirective implements AfterContentInit {
...
-  get empty() { return (this.value == null || this.value === '') && !this._isNeverEmpty(); }
+  get empty() {
+    if (this._isNeverEmpty()) {
+      return false;
+    }
+    if (this.value == null || this.value === '') {
+      // Check if this is an <input> element that contains bad input.
+      // If so, we know that it only appears empty because the value failed to parse.
+      if (this._elementRef.nativeElement instanceof HTMLInputElement &&
+          this._elementRef.nativeElement.validity.badInput) {
```
jelbourn: Add a unit test for this case?

mmalerba: I tried, but you can't programaticaly set the value to a bad input. I'd have to simulate key presses which sounded like more trouble than its worth

jelbourn: In that case just add a TODO to add an e2e test for this

jelbourn:
LGTM

kara:
@mmalerba Can you rebase?

PR #2367, opened 2016-12-23
test: include all of the necessary CSS in unit tests

### LINKED ISSUE #2358: Core styles should be loaded in unit tests
Currently in unit tests, the core Angular Material styles aren't built or included. As a result, the tests aren't always displaying an accurate representation of component style properties and positions, and we have to manually add the styles into the tests. This is particularly problematic for positioning tests. This isn't going to work long-term, so we should add the core styles.

crisbeto:
Includes the missing theme CSS in the unit tests and removes some of the styles that were used to simulate it.

Fixes #2358.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -19,12 +19,9 @@ describe('MdSelect', () => {
       declarations: [BasicSelect, NgModelSelect, ManySelects, NgIfSelect],
       providers: [
         {provide: OverlayContainer, useFactory: () => {
-          overlayContainerElement = document.createElement('div');
+          overlayContainerElement = document.createElement('div') as HTMLElement;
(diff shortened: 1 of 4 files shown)


src/lib/select/select.spec.ts
```diff
@@ -19,12 +19,9 @@ describe('MdSelect', () => {
       declarations: [BasicSelect, NgModelSelect, ManySelects, NgIfSelect],
       providers: [
         {provide: OverlayContainer, useFactory: () => {
-          overlayContainerElement = document.createElement('div');
+          overlayContainerElement = document.createElement('div') as HTMLElement;
+          overlayContainerElement.classList.add('cdk-overlay-container');
```
jelbourn: Not necessary for this PR: we should make a `FakeOverlayContainer` and use that everywhere

jelbourn:
LGTM

PR #2369, opened 2016-12-23
chore: remove hammerjs types as dependency

DevVersion:
* Removes the hammerjs types as a dependency to the @angular/material package (we don't have a dependency on hammerjs)
* Improves the name in the root `package.json` to show that this pkg file is just for the development (like in Angular)
* Moves hammerjs in development to the `devDependencies` because we need it always in our dev environment (same as in Angular 2)

Follow-up for https://github.com/angular/material2/commit/28691ca78042aed428781a3d88af7fc5cff0f8bb

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,9 +36,6 @@
     "systemjs": "0.19.38",
     "zone.js": "^0.6.23"
   },
-  "optionalDependencies": {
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

PR #2372, opened 2016-12-23
perf(tabs): reduce amount of reflows when aligning the ink bar

crisbeto:
Currently the ink bar gets aligned to the active tab every time change detection runs. These changes switch to aligning it if the selected index has changed or the header's content has changed. This can be improved further if we were to debounce the MutationObserver changes (will add in a follow-up PR).


crisbeto:
@andrewseguin I think I covered all the bases in regards to the ink bar always being aligned, but it would be nice to have a second pair of eyes on it.

jelbourn:
LGTM

PR #2389, opened 2016-12-24
fix(tabs): infinite loop when selectedIndex is set to NaN

crisbeto:
If the `selectedIndex` of a tabs instance is set to NaN, the tabs would get thrown into an infinite loop which also throws an error on every iteration. This was because of the normalization of the value which didn't catch NaN properly.


jelbourn:
LGTM

PR #2391, opened 2016-12-24
feat(tabs): add the ability to invert the header

### LINKED ISSUE #2387: [tabs] Header should have option to be aligned at bottom
Feature request

It is possible in Material 1 using md-tabs-align="bottom", as shown below. It would be great if we can have the same functionality in material2.

v9pzs

It is required sometimes to display the tabs at the bottom.

crisbeto:
Adds the ability to set the tab header position to the bottom. This inverts the tabs and the ink bar.

Fixes #2387.


JT-Bruch:
Any progress on this? Would love to see it in.

PR #2392, opened 2016-12-24
fix(snack-bar): prevent error when opening multiple snack bars in fast succession

### LINKED ISSUE #2390: Snackbar example in material.angular.io breaks after a few fast clicks
Try double clicking (without pausing) on the example on https://material.angular.io/components/component/snack-bar - after a couple of those you get an exception in the console and the whole app breaks.

Open up the link above, go to examples, try double clicking the "PIZZA PARTY" button 2 or 3 times or just hit it multiple times fast before the message goes down.

beta.1

Obviously not a show stopper as nobody would realistically be doing this but I'm just adding the bug report as I noticed the behavior anyway.

crisbeto:
Fixes a runtime error being thrown by Angular when opening multiple snack bars, that have a timeout, in quick succession. The fix waits for the zone to settle before firing the final callbacks.

**Note on testing:** This change is missing a unit test, because I couldn't find a way to trigger the error during tests, however the approach is similar to #2219.

Fixes #2390.


jelbourn:
LGTM

PR #2399, opened 2016-12-24
chore: switch saucelabs to websocket protocol

DevVersion:
* Since saucelabs already supports Websockets now and it's faster we should use Websockets here.
  Reference: https://saucelabs.com/blog/sauce-connect-gets-a-speed-boost-websocket-support


jelbourn:
LGTM

PR #2404, opened 2016-12-25
feat(observe-content): add debounce option and other improvements

crisbeto:
* Adds the ability to debounce the changes from the `cdkObserveContent` directive.
* Makes the `cdkObserveContent` directive pass back the `MutationRecord` to the `EventEmitter`.
* Fires the callback once per mutation event, instead of once per `MutationRecord`.

Relates to #2372.


crisbeto:
@jelbourn I added the MutationObserverFactory in order to be able to stub out the MutationObserver in the unit tests. I'm not sure whether this is the best way to do it, though.

src/lib/core/util/debounce.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Returns a function that won't be invoked, as long as it keeps being called. It will
+ * be invoked after it hasn't been called for `delay` milliseconds.
+ *
+ * @param func Function to be debounced.
+ * @param delay Amount of milliseconds to wait before calling the function.
+ * @param context Context in which to call the function.
+ */
+export function debounce(func: Function, delay: number, context?: any): Function {
```
crisbeto: It's been a while since I submitted this PR, but IIRC we can't really use it here because the callback is exposed through an EventEmitter which is consumed through the template; `<div (cdkObserveContent)="someCallback()">`. Technically we could debounce it with Rxjs inside the consumer, but that would make it really inconvenient (we'd need another observable that just proxies this one).

Another alternative for debouncing it through Rxjs could be to have an `exportAs` for the ObserveContent directive and using a ViewChild to access the EventEmitter directly, but that might look weird as well.

jelbourn: You should be able to do something like
```ts
private _event = new EventEmitter<void>();
@Output('cdkObserveContent') 
get event(): Observable<void> {
  return this.debounce ? 
      this._event.asObservable().debounceTime(this.debounce) :
      this._event.asObservable();
}
```

crisbeto: That didn't seem to work, but I managed to work around it.

jelbourn:
@crisbeto looks like just some minor CI failures

crisbeto:
@jelbourn can you take another look at this? The CI issues ended up being slightly more complicated, because of the `MutationObserver` stub and AoT, but I got it working.

src/lib/core/observe-content/observe-content.ts
```diff
@@ -20,12 +35,26 @@ export class ObserveContent implements AfterContentInit, OnDestroy {
...
+  @Input() debounce: number;
+
+  constructor(
+    private _mutationObserverFactory: MdMutationObserverFactory,
+    private _elementRef: ElementRef) { }
 
   ngAfterContentInit() {
-    this._observer = new MutationObserver(mutations => mutations.forEach(() => this.event.emit()));
+    this._debouncer
+      .debounceTime(this.debounce)
```
jelbourn: Can we conditionally add `debounceTime` to the chain if `debounce !== 0`? Andrew noticed when working on the scroll dispatcher that `auditTime` incurred a cost even when it was set to zero.

crisbeto: Done. Also rebased and fixed the fact the debouncer wasn't being completed.

PR #2409, opened 2016-12-26
chore: use gulp-connect as webserver

DevVersion:
* No longer rebuilds all components when changes have been recognized in the dev-app.
* Allows us to only reload when gulp already recognized a change (no need for filtering and extra watchers)
* No longer injects CSS as the previous live-server (injecting CSS does not work for Angular 2)

**Short summary**: Fixes live-reloading in the demo-app when doing changes in the demo-app.
> `gulp-server-livereload` has no easy way to disable the injection of CSS.


jelbourn:
LGTM

PR #2411, opened 2016-12-26
fix(tabs): crashing on chrome under certain conditions

### LINKED ISSUE #2151: Dialog does not support multiple tabs
Bug: Multiple tabs in a dialog cause chrome to go aw snap

I should be able to have more than one tab in a dialog

You can render one tab in a dialog but as soon as you add two chrome goes aw snap

Create a dialog.  Have the contents of the dialog be tabs.  Add one tab, everything works great.  Add two or more tabs and chrome will eventually go aw snap.

There is no purpose for using tabs if you can only have one

Angular 2.2.4
Material 2.0.0-alpha.11-2

No

crisbeto:
Prevents the tabs from either crashing Chrome (in Angular < 2.3) or throwing an animation error (in Angular >= 2.3). This was happening when the animations get triggered before the element is inserted into the DOM and thus doesn't have a computed `transform` value.

Fixes #2151.

### DIFF
--- a/src/lib/tabs/tab-body.html
+++ b/src/lib/tabs/tab-body.html
@@ -1,5 +1,5 @@
 <div class="md-tab-body-content" #content
-     [@translateTab]="_position"
+     [@translateTab]="_canBeAnimated ? _position : null"
      (@translateTab.start)="_onTranslateTabStarted($event)"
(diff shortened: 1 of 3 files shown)


crisbeto:
@andrewseguin can you review this one?

PR #2420, opened 2016-12-27
Documentation updated with md-menu-trigger-for dash-case

fabriziodebortoli:


crisbeto:
`menuTriggerFor` is the proper directive name, see https://github.com/angular/material2/pull/2244

jelbourn:
It's the other way around- dash-case is deprecated.

sparsh:
Make sure you have imported and added MdMenuModule in your Ng Module's imports.

PR #2423, opened 2016-12-27
fix(tooltip): avoid capturing the initial tap on mobile

### LINKED ISSUE #2326: md-tooltip eats the button click on mobile
Bug
Same as [issue in material 1](https://github.com/angular/material/issues/6792)

tested in ipad

crisbeto:
Currently the tooltip always binds the `mouseenter` and `mouseleave` events, however this can cause the click handlers on the tooltip host not to be fired on the first tap. These changes switch to binding the events only on devices that have touch events.

Fixes #2326.

### DIFF
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -18,7 +18,9 @@ import {TooltipPosition, MdTooltip, MdTooltipModule, SCROLL_THROTTLE_MS} from '.
 import {OverlayContainer} from '../core';
 import {Dir, LayoutDirection} from '../core/rtl/dir';
 import {OverlayModule} from '../core/overlay/overlay-directives';
+import {Platform} from '../core/platform/platform';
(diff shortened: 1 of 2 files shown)


mmalerba:
please rebase

crisbeto:
Done @mmalerba.

mmalerba:
Did you test this with a touchscreen laptop? I feel like it might be problematic

mmalerba:
Confirmed that this breaks chromebook pixel (and probably other touchscreen laptops). Lets try to come up with a better solution

crisbeto:
After some testing, it looks like this is only an issue for iOS. I've switched it to only disable the mouse events in that case.

andrewseguin:
Needs a small change - can you add the PlatformModule to the list of imports for MdTooltipModule? Otherwise an app may not pull in the Platform service.

crisbeto:
Done @andrewseguin. Sorry that it took a while, I missed the notification.

andrewseguin:
Can you rebase the PR? Thanks

PR #2400, opened 2016-12-24
Fix theming guide link

gyscos:
Fixes #2461

PR #2429, opened 2016-12-27
Update getting-started.md

jankei:
Path fix


DevVersion:
Thanks for the PR. That's likely the fifth duplicate of this change 😄 See #2400

PR #2430, opened 2016-12-27
fix(button-toggle): add the setDisabledState from ControlValueAccessor

crisbeto:
Adds the `setDisabledState` method from the `ControlValueAccessor` interface, which allows the control to be disabled when it is used with the `ReactiveFormsModule`. Also adds a couple of unit tests for the other `ControlValueAccessor` methods.


jelbourn:
LGTM

Issue #2363, opened 2016-12-23
md-input does not honor [value] expressions


arciisine:
#### Bug, feature request, or proposal:
When using `input[md-input]` and binding a value using `[value]="expr"`, the internal value is not updated when `[value]` resolves.  This in turn propagates to `empty` not updating once `[value]` resolves.  

#### What is the expected behavior?

When setting an input's value with `[value]`, it should be honored.  The placeholder should move out of the way when value is set.

#### What is the current behavior?
The input believes it is empty and the placeholder stays in the original spot.

#### What are the steps to reproduce?

http://plnkr.co/edit/79XlYsdCjUKQvnKQyDRU?p=preview

#### Which versions of Angular, Material, OS, browsers are affected?
Material 2.0.0-beta.0

arciisine:
As a follow up, this works properly when using `ngModel` or `formControlName`.  I actually have a scenario in which I want to show readonly form elements without binding the data to a form.  It also does work if using `value="name"` but not using `value="{{name}}"`

bboehm86:
same is true when using reactive forms and not setting the values right from the start (eg set them after http call or a timeout)
http://plnkr.co/edit/JkLHwmauIhJGy7dJfQU7?p=preview

MarcusCalidus:
it also seems not to work when binding via `[value]="some.value"`

![image](https://cloud.githubusercontent.com/assets/4113366/21452109/cac208c8-c904-11e6-8bc3-62f4fe5537fe.png)

arat3091:
it also seems not to work when binding via [disabled]="testCondition" or [disabled]= "'true'"

We still can edit the value when testCondition is true.

fxck:
It doesn't work even when using `formControl` or `formControlName`, specifically when the value is set asynchronously.

PR #2431, opened 2016-12-27
fix(input): ensure that property bindings work

### LINKED ISSUE #2428: Hello all! have problem with type in md-input (beta.1)
Bug, can`t set dynamically type for input 
http://plnkr.co/edit/aCZljQ4EqtTdYhgbuP5y?p=preview

DevVersion:
* Developers are currently not able to use property bindings for `required`, `disabled`, `type` and `placeholder`.
* This commit ensures that the Angular input setters are in sync with the native input element.
* Adds test to ensure that the property bindings work for the native input element (and not for the component instance as before)

cc. @mmalerba 

Fixes #2428

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -136,7 +136,7 @@ export class MdInputDirective implements AfterContentInit {
     // When using Angular inputs, developers are no longer able to set the properties on the native
     // input element. To ensure that bindings for `type` work, we need to sync the setter
     // with the native property. Textarea elements don't support the type property or attribute.
-    if (!this._isTextarea()) {
+    if (!this._isTextarea() && getSupportedInputTypes().has(this._type)) {
       this._renderer.setElementProperty(this._elementRef.nativeElement, 'type', this._type);
     }
   }
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -75,41 +75,56 @@ export class MdHint {
   `,
   host: {
     'class': 'md-input-element',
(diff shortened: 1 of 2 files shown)


arciisine:
Please also look at #2363 as I think it's the same issue, but targeting the `value` property.

DevVersion:
@arciisine I looked into the `[value]` issue, but decided to address it in another PR because it does not relate to the getters and setters.

arciisine:
Thanks @DevVersion. No worries, just wanted to make sure it wasn't overlooked if it fell into the same bucket.

src/lib/input/input-container.ts
```diff
@@ -168,12 +183,18 @@ export class MdInputDirective implements AfterContentInit {
...
     }
   }
 
-  private _isNeverEmpty() { return this._neverEmptyInputTypes.indexOf(this._type) != -1; }
+  private _isNeverEmpty() { return this._neverEmptyInputTypes.indexOf(this._type) !== -1; }
+
+  /** Determines if the component host is a textarea. If not recognizable it returns false. */
+  private _isTextarea() {
+    let nativeElement = this._elementRef.nativeElement;
+    return nativeElement ? nativeElement.nodeName === 'textarea' : 'input';
```
mmalerba: s/'input'/false

DevVersion: Good catch! - and lol 😄 The new way of reviewing

kara:
@DevVersion Can you rebase real quick?

DevVersion:
@kara Done. Waiting for one saucelabs task to finish.

PR #2432, opened 2016-12-27
fix(autosize): export md-autosize directive

### LINKED ISSUE #2419: textarea not manually resizable via resizeToFitContent method
Bug

[According to the API](https://material.angular.io/components/component/input), one should be able to call the method `resizeToFitContent` on the `MdTextareaAutosize` directive to manually trigger an auto-resize of the textarea that is labeled with `md-autosize`.

Since you have to declare a handle for the directive through a template variable, e. g. `#myTextareaAutosizeDirective="mdTextareAutosize"`, to call any methods on it, the MdTextareaAutosize directive should have a decorator property `exportAs` defined.

If you try to assign the directive to a template variable, an error is thrown. I assume this error occurs because of the directive's missing exportAs decorator property mentioned above.

The error reads: `There is no directive with "exportAs" set to "mdTextareaAutosize"`

This plunker [shows the use case](http://plnkr.co/edit/sHpwxl20r0lVMrRopW2w?p=preview).

This plunker includes the [template variable assignment that throws the error](http://plnkr.co/edit/F3Z0hO5sbpJYB1cq57Ny?p=preview).

In our app, we insert text into the textarea through a function which does not trigger the `change` event of the textarea that is normally used for resizing.

beta.0

Nope.

DevVersion:
**Note**: Not sure whether `mdTextareaAutosize` is more appropriated here. Also notice that the method is already tested by the first test [here](https://github.com/DevVersion/material2/blob/24cdad73e5a1f89ee761502c5daa99301fcccb64/src/lib/input/autosize.spec.ts#L44)

Closes #2419

### DIFF
--- a/src/demo-app/input/input-container-demo.html
+++ b/src/demo-app/input/input-container-demo.html
@@ -260,11 +260,11 @@
 <md-card class="demo-card demo-basic">
   <md-toolbar color="primary">Textarea Autosize</md-toolbar>
   <md-card-content>
-    <textarea md-textarea-autosize class="demo-textarea"></textarea>
+    <textarea mdTextareaAutosize class="demo-textarea"></textarea>
     <div>
       <md-input-container>
         <textarea mdInput
-                  md-textarea-autosize
+                  mdTextareaAutosize
                   placeholder="Autosized textarea"></textarea>
       </md-input-container>
     </div>
--- a/src/demo-app/input/input-container-demo.html
+++ b/src/demo-app/input/input-container-demo.html
@@ -260,11 +260,11 @@
 <md-card class="demo-card demo-basic">
   <md-toolbar color="primary">Textarea Autosize</md-toolbar>
   <md-card-content>
-    <textarea md-autosize class="demo-textarea"></textarea>
+    <textarea md-textarea-autosize class="demo-textarea"></textarea>
     <div>
       <md-input-container>
         <textarea mdInput
-                  md-autosize
+                  md-textarea-autosize
                   placeholder="Autosized textarea"></textarea>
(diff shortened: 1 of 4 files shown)


jelbourn:
You're right, it should be `mdTextareaAutosize` (and the selector should be that too...).

DevVersion:
@jelbourn Updated the directive. Now the directive feels pretty long.

src/demo-app/input/input-container-demo.html
```diff
@@ -260,11 +260,11 @@
 <md-card class="demo-card demo-basic">
   <md-toolbar color="primary">Textarea Autosize</md-toolbar>
   <md-card-content>
-    <textarea md-autosize class="demo-textarea"></textarea>
+    <textarea md-textarea-autosize class="demo-textarea"></textarea>
```
jelbourn: Should be `mdTextareaAutosize` (camelCase)

We should also make the old selector work at the same time as the new one for one release. 
(so both `md-autosize` and `mdTextAreaAutosize`)

DevVersion: Updated it. Didn't add it for `matTextareaAutosize` because according to PR #2244 it's not needed.

jelbourn: There should be a `mat` version too, but I've been lax on watching for that and will need to do another pass (probably at the same time I remove the deprecated selectors / properties).

jelbourn:
LGTM

kara:
@DevVersion Can you rebase?

DevVersion:
@kara Done.

PR #2433, opened 2016-12-27
chore(api): show exportAs name in documentation

DevVersion:
* Adds a new entry to the API documentation for the `exportAs` metadata.

  This is helpful when exposing an API for templates but the developers can't assume that the `exportAs` name is inferable from the directive name.

**Note**: This would help avoiding confusion like in #2419


jelbourn:
LGTM

Nice, thanks!

PR #2443, opened 2016-12-28
fix(input): not floating the label when updating the value programmatically with reactive forms

### LINKED ISSUE #2441: MdInput - Placeholder does not float when using formControl setValue()
Bug

Placeholder should be floating over input when setting the input value using setValue of its formControl.

When I use setValue, the input contains the value set and the placeholder overlays the value, without floating over it. If I enter a text manually just after using setValue, the placeholder goes to the right place.

Here's a plunker that reproduces the issue: 
http://plnkr.co/edit/huUsxFmbFKSZuCG92EmQ?p=preview

Angular: 2.3.1
material: git+https://github.com/angular/material2-builds.git but also tested on 2.0.0-beta1
Browser: Chromium 55.0.2883.87 
Os: Chromium 55.0.2883.87

crisbeto:
Fixes the input placeholder not being floated when the value is updated programmatically via `FormControl.setValue`. This was because the listener that handles the value changes gets initialized before the `FormControl`.

Fixes #2441.

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -11,7 +11,8 @@ import {
   Optional,
   Output,
   EventEmitter,
-  Renderer
+  Renderer,
+  OnInit,
 } from '@angular/core';
 import {coerceBooleanProperty} from '../core';
 import {NgControl} from '@angular/forms';
@@ -77,7 +78,7 @@ export class MdHint {
     '(input)': '_onInput()',
   }
 })
-export class MdInputDirective implements AfterContentInit {
+export class MdInputDirective implements AfterContentInit, OnInit {
   /** Whether the element is disabled. */
   @Input()
   get disabled() { return this._disabled; }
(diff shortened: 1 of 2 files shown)


src/lib/input/input-container.ts
```diff
@@ -142,14 +143,14 @@ export class MdInputDirective implements AfterContentInit {
 
   constructor(private _elementRef: ElementRef,
               private _renderer: Renderer,
-              @Optional() public _ngControl: NgControl) {
+              @Optional() public _ngControl: NgControl) { }
+
+  ngOnInit() {
```
petebacondarwin: So it is simply the case of moving the initialization into the `ngOnInit` hook?
Is that because the FormControl is normally changed by the "user" in their own `ngOnInit` method?

crisbeto: I ran across it while playing around, not sure about the reason, though.

crisbeto:
Closing in favour of #2455.

PR #2444, opened 2016-12-28
fix(select): not trimming long labels

### LINKED ISSUE #2440: MdSelect does not handle long text well
Bug

For long text to be truncated with ellipsis (at least that is what the styles seem to suggest).

When the select is open the long text is clipped at the end instead of being displayed with an ellipsis.
When long text is selected the the text goes beyond the select (with the down arrow on top of the text) forever.

See the select: http://plnkr.co/edit/Tg3diW0HhFsDq29rWo4Y?p=preview

Good behavior

Latest, Latest, Ubuntu/ChromeOS, Chrome

crisbeto:
* Fixes long text not being trimmed inside of `md-option`, because the element is a flex container. Switches it to `display: block` and uses `line-height` to center the text.
* Fixes the selected value not being trimmed, because the element doesn't have a defined width.

Fixes #2440.

### DIFF
--- a/src/demo-app/select/select-demo.ts
+++ b/src/demo-app/select/select-demo.ts
@@ -24,7 +24,7 @@ export class SelectDemo {
 
   drinks = [
     {value: 'coke-0', viewValue: 'Coke'},
-    {value: 'sprite-1', viewValue: 'Sprite'},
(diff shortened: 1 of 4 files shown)


kara:
@crisbeto This seems to throw off the alignment of icons in md-menu. Can you fix?

crisbeto:
Fixed @kara.

Issue #1718, opened 2016-11-03
toolbar: unable to specify attributes on first toolbar-row


ThomasBurleson:
#### Bug, feature request, or proposal:

`md-toolbar` projects content to an internal `md-toolbar-row` container and uses FlexBox css to layout in *'row'* directions for that container. This means that custom CSS must be used to modify the direction to *'column'*.

To achieve something like this is a challenge:

![toolbar-flow-directions](https://cloud.githubusercontent.com/assets/210413/19986294/5af75a6e-a1d3-11e6-9e37-a96881808f3f.png)

And using the pending Layout API does **not** work:

```html
<md-toolbar fx-layout="column">
    <div fx-layout="row" fx-layout-align="start center">  Title and buttons here </div>
    <div fx-flex="15px">        simulate a gap                                    </div>
    <div fx-layout="column">    2 lines of text organized vertically              </div>
</md-toolbar>
```

> By default this will layout each child div in a single row instead of the requested column flow direction.

An option for the developer is to use custom CSS:

```css
md-toolbar .md-toolbar-layout md-toolbar-row {
  flex-direction: column;
  align-content: flex-start;
  align-items: flex-start;
  min-height: 120px;
}
```

> But this ^^ option is not intuitive and seems to violate the entire idea of the Layout API.

Another solution - of course - is that the  developer can add a protection wrapper:

```
<md-toolbar>
   <div fx-layout="column">
      <div fx-layout="row" fx-layout-align="start center">  Title and buttons here </div>
      <div fx-flex="15px">        simulate a gap                                    </div>
      <div fx-layout="column">    2 lines of text organized vertically              </div>
   <div>
</md-toolbar>
```

>  but ^^ solution is dirty and adds complexity.

#### What are the steps to reproduce?

*  See source [Source to Flex-Layout Demo](https://github.com/angular/flex-layout/blob/master/src/demo-app/app/demo-app/demo-app.ts#L7-L23)

*  See live demo [Flex-Layout Demos](https://tburleson-layouts-demos.firebaseapp.com/#/docs)

#### What is the use-case or motivation for changing an existing behavior?

I propose that all Layout APIs be transcluded to the `md-toolbar-row` element. This way the developer's original intent is preserved without additional workarounds or hacks.

AsherBarak:
Following #2279, here is my suggestion:
If the user has not specified any `<md-toolbar-row>` elements, generate first element and wrap content. If the user has created a `<md-toolbar-row>` element, treat it as first element.
[...]

PR #2454, opened 2016-12-29
fix(toolbar): prevent content overflow and line-wrapping

### LINKED ISSUE #2451: Toolbar content overflow problem.
Bug with toolbar content overflow.

The toolbar content should be kept inside the toolbar with ellipsis

The content transpass the height of the toolbar.

Here is the Plunker:
http://plnkr.co/edit/HxeAlLOVKhpnKl5ifSq5?p=preview

Better visual.

"@angular/material": "2.0.0-beta.1",

DevVersion:
* No longer wraps lines inside of a `md-toolbar-row`
* Overflow for X-axis and Y-Axis are now hidden. Developers should use multiple `md-toolbar-row` elements or can always overwrite the styles.

**Note**: An ellipsis for flex elements is definitely possible, but it just bloats the toolbar and needs some refactoring. We should revisit this with #1718 

Fixes #2451

### DIFF
--- a/src/lib/toolbar/toolbar.scss
+++ b/src/lib/toolbar/toolbar.scss
@@ -42,7 +42,8 @@ md-toolbar {
     flex-direction: row;
     align-items: center;
 
-    // The toolbar should not wrap text by default. Developers can still overwrite it.
+    // Per Material specs a toolbar cannot have multiple lines inside of a single row. 
+    // Disable text wrapping inside of the toolbar. Developers are still able to overwrite it.
     white-space: nowrap;
   }
 }
--- a/src/lib/toolbar/toolbar.scss
+++ b/src/lib/toolbar/toolbar.scss
@@ -41,6 +41,10 @@ md-toolbar {
     // Flexbox Vertical Alignment
     flex-direction: row;
     align-items: center;
+
+    // Ensure toolbar content doesn't wrap and exceed the md-toolbar-row width. Ellipsis doesn't work on flex elements.
+    white-space: nowrap;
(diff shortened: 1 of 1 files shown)


src/lib/toolbar/toolbar.scss
```diff
@@ -41,6 +41,10 @@ md-toolbar {
     // Flexbox Vertical Alignment
     flex-direction: row;
     align-items: center;
+
+    // Ensure toolbar content doesn't wrap and exceed the md-toolbar-row width. Ellipsis doesn't work on flex elements.
+    white-space: nowrap;
+    overflow: hidden;
```
crisbeto: There's a `md-truncate-line` mixin for these cases.

DevVersion: True, but this also tries to show an ellipsis, which doesn't work and is then unnecessary.

jelbourn:
LGTM

mmalerba:
I'm seeing some test failures when I try to merge this into Google's codebase:

1. The `white-space: nowrap` seems to affect the children of `md-toolbar-row` as well. I think `flex-wrap: nowrap` might work better in this case.

2. One team has an absolutely positioned popup when you hover over a button in the `md-toolbar-row`, the `overflow: hidden` here doesn't play nice with that. Is there some way we can avoid it?

DevVersion:
As discussed with @mmalerba. We are just doing `white-space: none` for now because having `overflow: hidden` might reduce flexibility for toolbar elements.

PR #2455, opened 2016-12-29
fix(input): properly determine input value

### LINKED ISSUE #2363: md-input does not honor [value] expressions

### LINKED ISSUE #2441: MdInput - Placeholder does not float when using formControl setValue()

### LINKED ISSUE #2477: fix(MdInput): placeholders don't float when MdInput value has bound expression
When binding a value to `MdInput` using an expression, `placeholder` attribute (or `md-placeholder`) don't float when value is set, which results in placeholder text + input value to overlap. Also after focussing and blurring the input, the placeholder goes back to its original position. However, when using a static value, it works perfectly fine.

`placeholder` attribute (and/or `MdPlaceholder`) should float, also when initial `MdInput` value comes from an expression.

1. Create a property on component (`foo="bar"`)
2. Bind property to `MdInput` and give it a placeholder (`placeholder="Some Placeholder"`)

Here's a Plunk that reproduces the bug: https://plnkr.co/edit/NnwmYrakshZ5DEBkW27D?p=preview

There's an issue that *might* be related to this bug. When setting form values programmatically, placeholder positions also aren't updated. This issue can be found here: https://github.com/angular/material2/issues/2441

Someone came up with a fix for that as well: https://github.com/crisbeto/material2/commit/72bc74250501c4ec839c257afc67b64f3d88f818

DevVersion:
Right now the `MdInputDirective` tries to cache the `value` of the input.

To do this, the `MdInputDirective` needs to listen for `NgControl` value changes and for native `(change)` events.

This will be problematic when a value is set directly to the input element (using `[value]` property binding) because Angular is only able to recognize this change in the first ChangeDetection.

```html
<md-input-container>
  <input md-input [value]="myValue" placeholder="Label">
</md-input-container>
```

The approach of updating the value in the `ngAfterViewInit` lifecycle hook, will result in a binding change while being in a ChangeDetection, which leads to a ChangeDetection error.

```
Expression has changed after it was checked. Previous value: 'true'. Current value: 'false'.
```
_The credits for the `should update the value when using FormControl.setValue` test go to @crisbeto_

Fixes #2441. Fixes #2363. Fixes #2477

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -176,8 +176,13 @@ export class MdInputDirective {
   _onBlur() { this.focused = false; }
 
   _onInput() {
-    // This is a noop function and is used to let Angular recognize the changes
-    // to the input element while typing. Listening to the `input` event is similar to NgControls.
+    // This is a noop function and is used to let Angular know whenever the value changes.
+    // Angular will run a new change detection each time the `input` event has been dispatched.
+    // It's necessary that Angular recognizes the value change, because when floatingLabel
+    // is set to false and Angular forms aren't used, the placeholder won't recognize the
+    // value changes and will not disappear.
+    // Listening to the input event wouldn't be necessary when the input is using the
+    // FormsModule or ReactiveFormsModule, because Angular forms also listens to input events.
   }
 
   /** Make sure the input is a supported type. */
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -83,6 +83,7 @@ export class MdHint {
     '[required]': 'required',
     '(blur)': '_onBlur()',
     '(focus)': '_onFocus()',
+    '(input)': '_onInput()'
   }
 })
 export class MdInputDirective {
@@ -173,6 +174,11 @@ export class MdInputDirective {
   _onFocus() { this.focused = true; }
 
   _onBlur() { this.focused = false; }
+
+  _onInput() {
+    // This is a noop function and is used to let Angular recognize the changes
+    // to the input element while typing. Listening to the `input` event is similar to NgControls.
+  }
 
   /** Make sure the input is a supported type. */
   private _validateType() {
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -83,10 +83,9 @@ export class MdHint {
     '[required]': 'required',
     '(blur)': '_onBlur()',
     '(focus)': '_onFocus()',
-    '(input)': '_onInput()',
   }
 })
-export class MdInputDirective implements AfterContentInit {
+export class MdInputDirective {
 
   /** Variables used as cache for getters and setters. */
   private _type = 'text';
@@ -95,9 +94,6 @@ export class MdInputDirective implements AfterContentInit {
   private _required = false;
   private _id: string;
   private _cachedUid: string;
-
-  /** The element's value. */
-  value: any;
 
   /** Whether the element is focused or not. */
   focused = false;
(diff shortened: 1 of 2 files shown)


kara:
@DevVersion I think this might break turning off floating placeholders, e.g.:

```html
<md-input-container [floatingPlaceholder]="false">
   ...
</md-input-container>
```

Now that we're not listening to input events, the placeholder sticks around as the user types.

DevVersion:
Great catch. I've chatted with @kara about this and we thought about alternate approaches than listening to the `(input)` event.

>  To let Angular notice this change we need the `input` event, which triggers a new round of change detection and then can update the `md-empty` class on the placeholder.

Similar to [angular/angular#default_value_accessor.ts](https://github.com/angular/angular/blob/master/modules/%40angular/forms/src/directives/default_value_accessor.ts#L36), which should also trigger a change detection.

Need confirmation that the `input` event will always trigger a new change detection as in the FormsModule.

fxck:
can you get this at least in master? everything is kinda broken now

image

src/lib/input/input-container.ts
```diff
@@ -164,7 +157,10 @@ export class MdInputDirective implements AfterContentInit {
 
   _onBlur() { this.focused = false; }
 
-  _onInput() { this.value = this._elementRef.nativeElement.value; }
+  _onInput() {
+    // This is a noop function and is used to let Angular recognize the changes
+    // to the input element while typing. Listening to the `input` event is similar to NgControls.
```
kara: Nitpick: Can you add some more detail here?  I'd also add that if you're using one of the FormsModule or ReactiveFormsModule directives, the input event is already watched and will kick off change detection appropriately. But that if you're not using one of these and you've disabled the floating placeholder, we need to ensure that change detection still runs immediately on input to hide the placeholder.

DevVersion: Updated the comment.

kara:
LGTM, waiting on @amiral84

kara:
@DevVersion Sorry, one more rebase

DevVersion:
@kara Done.

abdulhaq-e:
hey @DevVersion I still get this issue if there is a condition check in `ngAfterViewInit/Checked`. See this plunkr: http://plnkr.co/edit/eyN8D8oZuEwDhHLINsZ5?p=preview

RicardoVaranda:
@abdulhaq-e I'm still having the same issue as you using master, if I don't use it in ngAfterViewInit it seems to crash the form and not display anything.

abdulhaq-e:
@RicardoVaranda See this comment: https://github.com/angular/material2/issues/2837#issuecomment-276538616

You need to use the `ngOnChanges` life cycle hook.

PR #2456, opened 2016-12-29
build: avoid typing errors in TypeScript 2.1 and update gulp-typescript

crisbeto:
* Avoids a couple of errors that show up when trying to compile the project against TypeScript 2.1 ("Cannot redeclare block-scoped variable 'module'" and "Type 'Timer' is not assignable to type 'number'"). Both errors were due to us loading both the Node and browser typings, which were overlapping with each other (e.g. `setTimeout` in the browser returns a `number`, whereas in Node it returns a `NodeJS.Timer`).
* Bumps to the latest version of `gulp-typescript` and switches to the new syntax for defining the Gulp task.
* Fixes an error with the `e2e-app` tsconfig.json that showed up after updating the `gulp-typescript`.

Related to #1998.

### DIFF
--- a/e2e/tsconfig.json
+++ b/e2e/tsconfig.json
@@ -13,7 +13,7 @@
     "sourceMap": true,
     "target": "es5",
     "typeRoots": [
-      "../node_modules/@types/"
+      "../node_modules/@types"
     ],
     "types": [
       "jasmine"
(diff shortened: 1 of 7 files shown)


clydin:
You should be able to remove the `require('typescript')` line as well.  The new version of `gulp-typescript` uses a peer dependency for typescript and will automatically use the installed version.

PR #2458, opened 2016-12-29
feat(select): emit change event

### LINKED ISSUE #2248: Bug - select not firing change event
Bug

(change) should be fired when value of select is changed manually 

(change) is not fired

This plunker reproduces the issue
https://plnkr.co/edit/MyhgEgH4SzFJs7aoJtzy?p=preview

Latest all

crisbeto:
Adds an event to `md-select` that is emitted when the selected option has changed.

Fixes #2248.

### DIFF
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -3,7 +3,7 @@
 <div class="demo-select">
   <div *ngIf="showSelect">
     <md-card>
-      <md-select placeholder="Food i would like to eat" [formControl]="foodControl" (change)="changeListener($event)">
+      <md-select placeholder="Food i would like to eat" [formControl]="foodControl">
         <md-option *ngFor="let food of foods" [value]="food.value"> {{ food.viewValue }} </md-option>
       </md-select>
       <p> Value: {{ foodControl.value }} </p>
@@ -16,10 +16,9 @@
     </md-card>
   </div>
 
-
   <md-card>
     <md-select placeholder="Drink" [(ngModel)]="currentDrink" [required]="isRequired" [disabled]="isDisabled"
-      (change)="changeListener($event)" #drinkControl="ngModel">
+      #drinkControl="ngModel">
       <md-option *ngFor="let drink of drinks" [value]="drink.value" [disabled]="drink.disabled">
         {{ drink.viewValue }}
       </md-option>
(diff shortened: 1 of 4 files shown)


src/demo-app/select/select-demo.html
```diff
@@ -3,7 +3,7 @@
 <div class="demo-select">
   <div *ngIf="showSelect">
     <md-card>
-      <md-select placeholder="Food i would like to eat" [formControl]="foodControl">
+      <md-select placeholder="Food i would like to eat" [formControl]="foodControl" (change)="changeListener($event)">
```
kara: Can we make a separate demo for the change event that does not use a form directive? The change event is only intended for people not using Angular forms modules, since these already have their own built-in change events. Just don't want to confuse the intended usage with this example.

src/lib/select/select.spec.ts
```diff
@@ -237,6 +237,27 @@ describe('MdSelect', () => {
       expect(fixture.componentInstance.select.selected).not.toBeDefined();
     });
 
+
+    it('should emit an event when the selected option has changed', () => {
```
kara: This should have its own test component that doesn't use a form module directive

crisbeto:
Addressed the feedback @kara.

src/lib/select/select.spec.ts
```diff
@@ -1140,13 +1140,46 @@ describe('MdSelect', () => {
...
+  });
 });
 
 @Component({
   selector: 'basic-select',
   template: `
     <div [style.height.px]="heightAbove"></div>
-    <md-select placeholder="Food" [formControl]="control" [required]="isRequired">
+    <md-select placeholder="Food" [formControl]="control" [required]="isRequired"
+      (change)="changeListener($event)">
```
kara: Can you remove the change event and listener from the basic select component? It should only be in the new test component.

crisbeto:
Fixed @kara, not sure how I forgot to remove the last one.

src/lib/select/select.spec.ts
```diff
@@ -1169,6 +1201,7 @@ class BasicSelect {
   isRequired: boolean;
   heightAbove = 0;
   heightBelow = 0;
+  changeListener = jasmine.createSpy('MdSelect change listener');
```
kara: Can you remove this too?

crisbeto: Sure, done.

kara:
@crisbeto Is this ready to merge? Just noticed it doesn't have the label.

crisbeto:
It is, I'm not sure who's supposed to add the label.

kara:
@crisbeto Since you are the author and a member of the repo, you are the one that adds the "merge ready" label :)  "LGTM" just means that it's approved; "merge ready" means that the author has made all the changes they wanted to.

PR #2463, opened 2016-12-29
chore: track library payload in database

### LINKED ISSUE #111: Track payload size on a per-commit basis
Payload size for the entire library should be tracked on a per-commit basis. Should likely use the same process as `angular/angular`.

DevVersion:
* Adds a new gulp task which tries to compare sizes of the library results.

  Those results will be always announced in the Travis build and when not running as a PR those will be published to firebase (useful for statistics)

Closes #111

### DIFF
--- a/tools/gulp/task_helpers.ts
+++ b/tools/gulp/task_helpers.ts
@@ -16,6 +16,7 @@ const gulpSourcemaps = require('gulp-sourcemaps');
 const gulpAutoprefixer = require('gulp-autoprefixer');
 const gulpConnect = require('gulp-connect');
 const resolveBin = require('resolve-bin');
+const firebaseAdmin = require('firebase-admin');
 
 
 /** If the string passed in is a glob, returns it, otherwise append '**\/*' to it. */
@@ -210,4 +211,19 @@ export function sequenceTask(...args: any[]) {
       done
     );
   };
+}
+
+/** Opens a connection to the firebase realtime database. */
+export function openFirebaseDatabase() {
+  // Initialize the Firebase application with admin credentials.
+  firebaseAdmin.initializeApp({
+    credential: firebaseAdmin.credential.cert({
+      project_id: 'material2-dashboard',
+      client_email: 'firebase-adminsdk-ch1ob@material2-dashboard.iam.gserviceaccount.com',
+      private_key: process.env['MATERIAL2_FIREBASE_PRIVATE_KEY']
+    }),
+    databaseURL: 'https://material2-dashboard.firebaseio.com'
+  });
+
+  return firebaseAdmin.database();
 }
--- a/tools/gulp/task_helpers.ts
+++ b/tools/gulp/task_helpers.ts
@@ -216,6 +216,7 @@ export function sequenceTask(...args: any[]) {
 /** Opens a connection to the firebase realtime database. */
 export function openFirebaseDatabase() {
   // Initialize the Firebase application with admin credentials.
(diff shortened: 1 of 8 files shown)


DevVersion:
@jelbourn I just added the private key for the firebase database on the CI.

tools/gulp/task_helpers.ts
```diff
@@ -211,3 +212,18 @@ export function sequenceTask(...args: any[]) {
...
   };
 }
+
+/** Opens a connection to the firebase realtime database. */
+export function openFirebaseDatabase() {
+  // Initialize the Firebase application with admin credentials.
+  firebaseAdmin.initializeApp({
+    credential: firebaseAdmin.credential.cert({
+      project_id: 'material2-dashboard',
+      client_email: 'firebase-adminsdk-ch1ob@material2-dashboard.iam.gserviceaccount.com',
```
jelbourn: Where does this `client_email` come from?

tools/gulp/tasks/payload.ts
```diff
@@ -0,0 +1,56 @@
...
+    normal: getFilesize(BUNDLE_PATH),
+    minified_uglify: getUglifiedSize(BUNDLE_PATH),
+    timestamp: Date.now()
+  };
+
+  // Print the results to the console, so we can read it from the CI.
+  console.log('Payload Results:', JSON.stringify(results, null, 2));
+
+  // Publish the results to firebase when it runs on Travis and not as a PR.
+  if (process.env['TRAVIS_PULL_REQUEST'] === 'false') {
```
jelbourn: Might be good to make this a common `isTravisPullRequest` function.

tools/gulp/tasks/payload.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/** Task which runs test against the size of whole library. */
+task('payload', ['build:release'], () => {
+
+  let results = {
+    normal: getFilesize(BUNDLE_PATH),
+    minified_uglify: getUglifiedSize(BUNDLE_PATH),
+    timestamp: Date.now()
+  };
+
```
jelbourn: How about
```js
  let results = {
    umd_kb: getFilesize(BUNDLE_PATH),
    umd_minified_uglify_kb: getUglifiedSize(BUNDLE_PATH),
    timestamp: Date.now()
  };
```

tools/gulp/task_helpers.ts
DevVersion: Added the comment.

tools/gulp/tasks/payload.ts
DevVersion: Yeah I like the `umd` prefix here.

tools/gulp/tasks/payload.ts
DevVersion: Created a method `isTravisPushBuild` because if we create `isTravisPullRequest`, the method will also return `false` when we not run in Travis at all. Using `isTravisPushBuild` should be fine for now.

jelbourn:
LGTM

kara:
@DevVersion This job appears to be failing consistently on Travis (https://travis-ci.org/angular/material2/jobs/188703192). Can you take a look?

DevVersion:
@kara Yeah I've been chatting with Jeremy on Slack about it. 

It looks like something messes up the Firebase token.

Here is the PR we wanted to try: #2510

kara:
Thanks, missed that one!

PR #2464, opened 2016-12-29
chore: run e2e tests concurrently in the ci 

DevVersion:
* Now runs the e2e tests concurrently in the CI because currently it takes likely 12 minutes to finsish the e2e mode.


DevVersion:
@jelbourn Nice. The e2e tests now finish in ~6 minutes instead of ~12 minutes! 🎉 

> Notice the workaround for the `logLevel`. I've been chatting with @cnishina about it.
> Without disabled info logging it looks like [this](https://travis-ci.org/angular/material2/jobs/187560434)

PR #2471, opened 2016-12-30
fix(input): work around Chrome issue that collapses the input

### LINKED ISSUE #2470: Bug: md-input with min and max break element display
When specifying both min and max on a md-input number the displayed element size breaks

image

http://plnkr.co/edit/TJl6TFi9WefY9I6HHjB0?p=preview

crisbeto:
This fix works around a Chrome bug, where number inputs that have a `min` and a `max`, but not a `step`, will collapse to about 50% their proper width. This most likely isn't a huge use case, but it's probably good to have the inputs be consistent across browsers.

For more info the the bug: https://bugs.chromium.org/p/chromium/issues/detail?id=598164

Fixes #2470.

### DIFF
--- a/src/lib/input/input-container.spec.ts
+++ b/src/lib/input/input-container.spec.ts
@@ -110,6 +110,26 @@ describe('MdInputContainer', function () {
     expect(el.classList.contains('md-empty')).toBe(true);
   });
 
+  it('should add a default "step" to number inputs', () => {
+    let fixture = TestBed.createComponent(MdInputContainerNumberTestController);
+    fixture.detectChanges();
+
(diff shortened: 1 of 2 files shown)


jelbourn:
I'm not sure it's the responsibility of Material to workaround a Chrome bug that affects all `<input>` elements. 

@mmalerba thoughts?

mmalerba:
Agreed, especially since the `<input>` element is now something the user authors I'd rather just do as little mucking with it as possible.

crisbeto:
Cool, I'll close the PR in that case.

PR #2472, opened 2016-12-30
fix(tooltip): better handling of multi-line text

### LINKED ISSUE #2205: Tooltip backgrounds only span the first line of text.

When the text of a tooltip spans multiple lines, only the first line will have a background.

The tooltip background is behind all lines of text.

The tooltip background is only behind the top line of text.

http://plnkr.co/edit/GKaAf1dfZWqa3cWC0pSy?p=preview

Mouse over the button in the right of the toolbar.

Material 2.0.0 alpha 11, Chrome 54.0.2840.99 m

The link to the recommended plunkr in the issue template uses alpha 10, which does not have the wrapping behavior for tooltips.

crisbeto:
Switches the tooltip from having a hard-coded height to using `padding` and `font-size` to achieve the same height, while allowing the text to wrap. Technically the same could've been achieved by using `min-height`, instead of `height`, however since it was using `line-height` to center the text, multi-line text didn't look too good.

Fixes #2205.

### DIFF
--- a/src/lib/tooltip/tooltip.scss
+++ b/src/lib/tooltip/tooltip.scss
@@ -13,7 +13,7 @@ $md-tooltip-vertical-padding: ($md-tooltip-target-height - $md-tooltip-font-size
 }
 
 .md-tooltip {
-  color: #fff;
+  color: white;
   padding: $md-tooltip-vertical-padding $md-tooltip-horizontal-padding;
   border-radius: 2px;
   font-family: $md-font-family;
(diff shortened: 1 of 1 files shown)


src/lib/tooltip/tooltip.scss
```diff
@@ -2,23 +2,23 @@
...
+$md-tooltip-vertical-padding: ($md-tooltip-target-height - $md-tooltip-font-size) / 2;
 
 :host {
   pointer-events: none;
 }
 
 .md-tooltip {
-  color: white;
-  padding: 0 $md-tooltip-padding;
+  color: #fff;
```
jelbourn: I prefer using `white` since it's more meaningful; the css minifier will condense it.

PR #2476, opened 2016-12-30
feat(chips): Add remove functionality/styling.

### LINKED ISSUE #3143: md-chips example doesn't work
Fix md-chips example as given below

A new chip should be created when some text is entered on the input line and the Enter button pressed.

When run in plunkr, a new chip was not created and errors were generated

Run the chips demo in plunkr

Suggestions to fix:
**In chips-demo.html:47**
`<input mdInput (keyup.enter)="add($event)" placeholder="New Contributor...">`

**In chips-demo.ts:42 (add method)**
```
add(input: KeyboardEvent): void {
    input.target.value.split(',').forEach( v => {
        this.people.push({name: v.trim()});
    });
    input.target.value = '';
}
```

topherfangio:
Add events, styling and keyboard controls to allow removable chips.

 - Add basic styling for a user-provided remove icon.
 - Add keyboard controls for backspace/delete.
 - Add `(remove)` event which is emitted when the remove icon or
   one of the delete keys is pressed.
 - Add `md-chip-remove` directive which can be applied to `<md-icon>`
   (or others) to inform the chip of the remove request.

Add new directive `mdChipInput` for controlling:
 - `(chipAdded)` - Event fired when a chip should be added.
 - `[separatorKeys]` - The list of keycodes that will fire the
   `(chipAdded)` event.
 - `[addOnBlur]` - Whether or not to fire the `(chipAdded)` event
   when the input is blurred.

Additionally, fix some issues with dark theme and add styling/support
for usage inside the `md-input-container` and add styling for focused
chips.

BREAKING CHANGE - The `selectable` property of the `md-chip-list` has
now been moved to `md-chip` to maintain consistency with the new
`removable` option.

If you used the following code,

```html
<md-chip-list [selectable]="selectable">
  <md-chip>My Chip</md-chip>
</md-chip-list>
```

you should switch it to

```html
<md-chip-list>
  <md-chip [selectable]="selectable">My Chip</md-chip>
</md-chip-list>
```

References #120. Closes #3143.

Hat tip to @willshowell for suggestion to use secondary-text
for focus color :-)

### DIFF
--- a/src/demo-app/chips/chips-demo.ts
+++ b/src/demo-app/chips/chips-demo.ts
@@ -1,4 +1,5 @@
-import {Component, ElementRef} from '@angular/core';
+import {Component} from '@angular/core';
+import {MdChipInputEvent, ENTER, COMMA} from '@angular/material';
 
 export interface Person {
   name: string;
@@ -18,6 +19,12 @@ export interface DemoColor {
 export class ChipsDemo {
   visible: boolean = true;
   color: string = '';
+  selectable: boolean = true;
+  removable: boolean = true;
+  addOnBlur: boolean = true;
+
+  // Enter, comma, semi-colon
+  separatorKeys = [ENTER, COMMA, 186];
 
   people: Person[] = [
     { name: 'Kara' },
@@ -39,10 +46,26 @@ export class ChipsDemo {
     alert(message);
   }
 
-  add(input: ElementRef): void {
-    if (input.nativeElement.value && input.nativeElement.value.trim() != '') {
-      this.people.push({ name: input.nativeElement.value.trim() });
-      input.nativeElement.value = '';
+  add(event: MdChipInputEvent): void {
+    let input = event.input;
+    let value = event.value;
+
+    // Add our person
+    if (value && value.trim() != '') {
+      this.people.push({ name: value.trim() });
+    }
+
+    // Reset the input value
+    if (input) {
+      input.value = '';
+    }
+  }
+
+  remove(person: Person): void {
+    let index = this.people.indexOf(person);
+
+    if (index >= 0) {
+      this.people.splice(index, 1);
     }
   }
 
--- a/src/lib/chips/_chips-theme.scss
+++ b/src/lib/chips/_chips-theme.scss
@@ -16,34 +16,90 @@
 
   // The spec only provides guidance for light-themed chips. When inside of a dark theme, fall back
   // to standard background and foreground colors.
-  $unselected-background: if($is-dark-theme, mat-color($background, card), #e0e0e0);
+  $unselected-background: if($is-dark-theme, #656565, #e0e0e0);
   $unselected-foreground: if($is-dark-theme, mat-color($foreground, text), $light-foreground);
 
   $selected-background: if($is-dark-theme, mat-color($background, app-bar), #808080);
   $selected-foreground: if($is-dark-theme, mat-color($foreground, text), $light-selected-foreground);
 
+  $focus-color: mat-color($foreground, secondary-text);
+
   .mat-chip:not(.mat-basic-chip) {
     background-color: $unselected-background;
     color: $unselected-foreground;
+
+    .mat-chip-focus-border {
+      pointer-events: none;
+    }
+
+    &:focus {
+      outline: none;
+      border: 2px solid $focus-color;
+    }
+
+    .mat-chip-remove {
+      color: $unselected-foreground;
+      opacity: 0.3;
+
+      &:hover {
+        opacity: 0.54;
+      }
+    }
   }
 
   .mat-chip.mat-chip-selected:not(.mat-basic-chip) {
     background-color: $selected-background;
     color: $selected-foreground;
 
+    .mat-chip-remove {
+      color: $selected-foreground;
+      opacity: 0.4;
+
+      &:hover {
+        opacity: 0.54;
+      }
+    }
+
     &.mat-primary {
       background-color: mat-color($primary, 500);
       color: mat-contrast($primary, 500);
+
+      .mat-chip-remove {
+        color: mat-contrast($primary, 500);
+        opacity: 0.4;
+
+        &:hover {
+          opacity: 0.54;
+        }
+      }
     }
 
     &.mat-accent {
       background-color: mat-color($accent, 500);
       color: mat-contrast($accent, 500);
+
+      .mat-chip-remove {
+        color: mat-contrast($accent, 500);
+        opacity: 0.4;
+
+        &:hover {
+          opacity: 0.54;
+        }
+      }
     }
 
     &.mat-warn {
       background-color: mat-color($warn, 500);
       color: mat-contrast($warn, 500);
+
+      .mat-chip-remove {
+        color: mat-contrast($warn, 500);
+        opacity: 0.4;
+
+        &:hover {
+          opacity: 0.54;
+        }
+      }
     }
   }
 }
--- a/src/lib/chips/chip-list.spec.ts
+++ b/src/lib/chips/chip-list.spec.ts
@@ -3,10 +3,12 @@ import {Component, DebugElement, QueryList} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {MdChip, MdChipList, MdChipsModule} from './index';
 import {FocusKeyManager} from '../core/a11y/focus-key-manager';
+import {MdInputModule} from '../input/index';
 import {FakeEvent} from '../core/a11y/list-key-manager.spec';
-import {SPACE, LEFT_ARROW, RIGHT_ARROW} from '../core/keyboard/keycodes';
+import {LEFT_ARROW, RIGHT_ARROW, BACKSPACE, DELETE} from '../core/keyboard/keycodes';
+import {Dir} from '../core/rtl/dir';
 
-class FakeKeyboardEvent extends FakeEvent {
+export class FakeKeyboardEvent extends FakeEvent {
   constructor(keyCode: number, protected target: HTMLElement) {
     super(keyCode);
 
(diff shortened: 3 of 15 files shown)


src/demo-app/chips/chips-demo.ts
```diff
@@ -36,6 +38,9 @@ export class ChipsDemo {
     { name: 'Warn', color: 'warn' }
   ];
 
+  constructor() {
+  }
```
DevVersion: Is constructor unnecessary?

src/demo-app/chips/chips-demo.ts
```diff
@@ -47,6 +52,16 @@ export class ChipsDemo {
     }
   }
 
+  remove(person: Person): void {
+    if (person) {
```
DevVersion: I think this check can be dropped, because TypeScript already validates the person is *not* undefined.

Also the `indexOf` would just return `-1` if it's really undefined.

src/lib/chips/chip-list.spec.ts
```diff
@@ -19,199 +20,200 @@ describe('MdChipList', () => {
...
-        let midItem = array[2];
+    describe('focus behaviors', () => {
+      beforeEach(() => {
+        manager = chipListInstance._keyManager;
+      });
 
-        // Focus the middle item
-        midItem.focus();
+      it('focuses the first chip on focus', () => {
+        let FOCUS_EVENT: Event = {} as Event;
```
DevVersion: Looks like you are casting it twice.

src/lib/chips/chip-list.spec.ts
```diff
@@ -19,199 +20,200 @@ describe('MdChipList', () => {
...
+          chipListInstance._keydown(DELETE_EVENT);
+          fixture.detectChanges();
+
+          // It focuses the last chip
+          expect(manager.focusedItemIndex).toEqual(chips.length - 1);
+        });
+
+        it('BACKSPACE focuses the last chip', () => {
+          let nativeInput = chipListNativeElement.querySelector('input');
+          let BACKSPACE_EVENT: KeyboardEvent = new FakeKeyboardEvent(BACKSPACE, nativeInput) as any;
```
DevVersion: Same as above. The cast to `any` seems unnecessary here?

topherfangio: The first one is declaring the type; not casting it. When attempting to cast it to `KeyboardEvent` it requires me to override *everything* in `KeyboardEvent` so @jelbourn told me to just us `any` for now.

src/demo-app/chips/chips-demo.ts
topherfangio: No; I think it got added by accident or something. I'll remove it.

src/lib/chips/chip-list.spec.ts
DevVersion: Hm, you should be able to do.

```ts
let backspaceEvent = new FakeKeyboardEvent(BACKSPACE, nativeInput) as KeyboardEvent;
```

as in [this example](http://www.typescriptlang.org/play/#src=class%20FakeEvent%20%7B%7D%0D%0A%0D%0A%0D%0Alet%20k%20%3D%20new%20FakeEvent()%20as%20KeyboardEvent%3B%0D%0A%0D%0Adocument.write('Works%20for%20me')%3B)

topherfangio:
Please hold off on reviewing this; I accidentally pushed some focus code into this branch that needs to be removed. Will update when done, but I'm working on some md1 issues at the moment.

src/lib/chips/chip-list.ts
```diff
@@ -49,69 +49,125 @@ import {SPACE, LEFT_ARROW, RIGHT_ARROW} from '../core/keyboard/keycodes';
 })
 export class MdChipList implements AfterContentInit {
 
-  /** Track which chips we're listening to for focus/destruction. */
-  private _subscribed: WeakMap<MdChip, boolean> = new WeakMap();
+  /** Whether or not the chip list is currently focusable via keyboard interaction. */
```
jelbourn: This isn't a boolean, so `Whether...` isn't really the right way to phrase it. I think it's fine to say "tabIndex of the chip-list element."

src/lib/chips/chip-list.ts
```diff
@@ -49,69 +49,125 @@ import {SPACE, LEFT_ARROW, RIGHT_ARROW} from '../core/keyboard/keycodes';
...
+  protected _tabIndex = -1;
+
+  /** When a chip is destroyed, we track the index so we can focus the appropriate next chip. */
+  protected _destroyedIndex: number = null;
```
jelbourn: Why store this instead of moving the focus immediately?

src/lib/chips/chip-list.ts
```diff
@@ -126,35 +182,25 @@ export class MdChipList implements AfterContentInit {
...
+  protected _subscribeChips(chips: QueryList < MdChip >): void {
     chips.forEach(chip => this._addChip(chip));
   }
 
   /**
+   * Check the tab index as you should not be allowed to focus an empty list.
+   */
+  protected _checkTabIndex(): void {
+    // If we have 0 chips, we should not allow keyboard focus
+    this._tabIndex = (this.chips.length == 0 ? -1 : 0);
```
jelbourn: Instead of manually calling `_checkTabIndex` and updating `_tabIndex`, why not just create a host binding like
```ts
host: {
  '[attr.tabindex]': 'chips.length ? 0 : -1'
}
```

However, this should also be overridden by any user-provided `tabIndex` value, so that makes it a bit more complex.

src/lib/chips/chip-list.ts
```diff
@@ -49,69 +49,125 @@ import {SPACE, LEFT_ARROW, RIGHT_ARROW} from '../core/keyboard/keycodes';
...
   }
 
-  set selectable(value: boolean) {
-    this._selectable = coerceBooleanProperty(value);
+  /**
+   * Ensure the label element does not contain the md-empty class unless we have 0 chips.
+   *
+   * TODO: Remove when md-input-container allows md-empty configuration.
+   * */
+  protected _hackLabelElement() {
```
jelbourn: cc @mmalerba 

What is it that you need the input to do here? I'd rather avoid merging something like this.

src/lib/chips/chip-list.spec.ts
```diff
@@ -242,3 +243,19 @@ class StaticChipList {
   chipDeselect(index: Number) {
   }
 }
+
+@Component({
+  template: `
+    <md-input-container>
```
jelbourn: cc @mmalerba - thoughts?

I'm not sure putting the chip-list inside of the input-container is the approach we ultimately want to take. Right now I don't think the input-container is in itself designed to have any content other than the input/textarea and friends. 

I know we want to make the chips be part of the input's underline, but perhaps there's another way we could approach that.

src/lib/chips/chip-list.ts
```diff
@@ -49,69 +49,125 @@ import {SPACE, LEFT_ARROW, RIGHT_ARROW} from '../core/keyboard/keycodes';
...
 
   /** The ListKeyManager which handles focus. */
   _keyManager: ListKeyManager;
 
   /** The chip components contained within this chip list. */
   chips: QueryList<MdChip>;
 
-  constructor(private _elementRef: ElementRef) { }
+  constructor(protected _renderer: Renderer, protected _elementRef: ElementRef,
+              @Optional() protected _mdInputContainer: MdInputContainer) {
```
jelbourn: So far we've been trying to avoid introducing any inter-component dependencies. In this case, I don't think we want to tightly couple the md-input with the chip-list, since people may use chips in combination with their own custom input (which is something Jen's team wants to do). 

A less-coupled approach would be to create an interface like `ChipInput`, and accept that as an `@Input` (and making sure that either `MdInput` or `MdInputContainer` conforms to that interface. Then we could connect the two without introducing a dependency.

src/lib/chips/chip-list.spec.ts
mmalerba: I do want to support this sort of thing in the future (the `mdInput` directive would go on the `md-chip-list` though and `md-chips-list` would have to provide some input-like interface for `mdInput` to work with). For now though what about something like this:

```html
<md-input-container>
  <md-chip-list md-prefix>
    <md-chip>Chip 1</md-chip>
    ...
  </md-chip-list>
  <input mdInput>
</md-input-container>
```

fxck:
Is there any progress on chips? Now that autocomplete is more or less done, it'd be nice to start working on https://material.angularjs.org/latest/demo/chips 

Is it assigned to someone? I already have my own version working, I could try and make a PR..

topherfangio:
@fxck I am the developer assigned and am indeed working on the chips. I just took some time off to have a baby and am getting back to it :-)

I pushed some updates yesterday and should have more today. Hoping to have a mostly final version by this weekend, but it will need to go through the review process before being merged to master.

src/lib/chips/_chips-theme.scss
```diff
@@ -16,34 +16,90 @@
...
   // The spec only provides guidance for light-themed chips. When inside of a dark theme, fall back
   // to standard background and foreground colors.
-  $unselected-background: if($is-dark-theme, mat-color($background, card), #e0e0e0);
+  $unselected-background: if($is-dark-theme, #656565, #e0e0e0);
   $unselected-foreground: if($is-dark-theme, mat-color($foreground, text), $light-foreground);
 
   $selected-background: if($is-dark-theme, mat-color($background, app-bar), #808080);
   $selected-foreground: if($is-dark-theme, mat-color($foreground, text), $light-selected-foreground);
 
+  $focus-color: mat-color($foreground, text);
```
willshowell: What about using `mat-color($foreground, secondary-text)` instead? It would match the borders of the checkbox and radio components.

topherfangio:
@kara @DevVersion Can you guys please review this updated PR at your convenience? It adds the `mdChipRemove` and the `mdChipInput` functionality along with some other small changes.

src/lib/chips/chips.scss
```diff
@@ -77,3 +89,33 @@ $mat-chips-chip-margin: $mat-chip-horizontal-padding / 4;
...
+
+.mat-chip-remove.mat-chip-remove-hidden {
+  display: none;
+}
+
+// Override a few styles when inside an mat-input-container
+.mat-input-container .mat-chip-list-wrapper input {
+  width: auto;
+  height: 38px;
+  margin-left: 8px;
```
DevVersion: Is there a way we can move those hard-coded values to variables?

And above for `.mat-chip-remove`, couldn't we calculate those `left`, `right` values?

src/lib/chips/chip.ts
```diff
@@ -145,6 +246,7 @@ export class MdChip implements Focusable, OnInit, OnDestroy {
     if (el.nodeName.toLowerCase() == 'mat-basic-chip' || el.hasAttribute('mat-basic-chip') ||
         el.nodeName.toLowerCase() == 'md-basic-chip' || el.hasAttribute('md-basic-chip')) {
       el.classList.add('mat-basic-chip');
+      this._renderer.setElementClass(this._elementRef.nativeElement, `md-chip`, true);
```
DevVersion: Is this correct? Do we really need to have the `md-chip` class on the host-element?

And also above with `classList.add('mat-chip')`, I think this is obsolete since we already set the class in the host metadata.

src/lib/chips/chip.ts
```diff
@@ -22,33 +23,47 @@ export interface MdChipEvent {
...
   protected _color: string = 'primary';
 
+  /** Whether or not the chip is displaying the remove icon. */
+  _hasRemoveIcon: boolean = false;
+
   /** Emitted when the chip is focused. */
   onFocus = new EventEmitter<MdChipEvent>();
 
+  /** Emitted when the removable property changes. */
+  onRemovableChange = new EventEmitter<boolean>();
```
DevVersion: We should not expose the `EventEmitter`.

```ts
class X {
  private _onRemovableChange = new EventEmitter<boolean>();
  onRemovableChange = this._onRemovableChange.asObservable();
}
```

src/lib/chips/chip.spec.ts
```diff
@@ -110,7 +120,132 @@ describe('Chips', () => {
...
+      });
+
+      describe('when removable is false', () => {
+        beforeEach(() => {
+          testComponent.removable = false;
+          fixture.detectChanges();
+        });
+
+        it('DELETE does not emit the (remove) event', () => {
+          const DELETE_EVENT: KeyboardEvent = new FakeEvent(DELETE) as KeyboardEvent;
```
DevVersion: double cast shouldn't be necessary (also above).

TypeScript automatically interferes the types when doing `const event = new FakeEvent(DELETE) as KeyboardEvent;`

src/lib/chips/chip-list.spec.ts
topherfangio: @DevVersion So, it looks like you are correct and this does indeed compile/run correctly, it's just that Webstorm is showing it as an error. I guess I need to play with my settings to figure out why.

src/lib/chips/chip.ts
topherfangio: Good catch; this was leftover and I'm removing.

src/lib/chips/chip.spec.ts
topherfangio: It's not really a double cast (it's just an explicit type declaration), but it is indeed extra code. I'll remove.

JacobBrandt:
Works great.  Looking forward to this being merged.

willshowell:
@topherfangio is this waiting for further review? Don't want it to get stuck in limbo!

topherfangio:
@JacobBrandt Thanks for the kind words! I'm glad you like it!

@willshowell It is indeed waiting for a final review and I am nagging the appropriate people to get it done ;-)

src/lib/chips/chip-list.ts
```diff
@@ -148,11 +175,19 @@ export class MdChipList implements AfterContentInit {
    *
    * @param chips The list of chips to be subscribed.
    */
-  protected _subscribeChips(chips: QueryList<MdChip>): void {
+  protected _subscribeChips(chips: QueryList < MdChip >): void {
```
jelbourn: You can omit the spaces inside the `< >`

src/lib/chips/chip-list.ts
```diff
@@ -196,6 +226,32 @@ export class MdChipList implements AfterContentInit {
   }
 
   /**
+   * Checks to see if a focus chip was recently destroyed so that we can refocus the next closest
+   * one.
+   */
+  protected _checkDestroyedFocus() {
+    let chipsArray = this.chips.toArray();
```
jelbourn: The `toArray` shouldn't be necessary; `QueryList` has a `length` property

src/lib/chips/chip.spec.ts
```diff
@@ -110,7 +120,132 @@ describe('Chips', () => {
...
+
+    describe('keyboard behavior', () => {
+
+      describe('when selectable is true', () => {
+        beforeEach(() => {
+          testComponent.selectable = true;
+          fixture.detectChanges();
+        });
+
+        it('SPACE selects/deselects the currently focused chip', () => {
```
jelbourn: Tests should start with the word `should`, e.g.,
```ts
it('should toggle selection with the space key', ...
```

axtho:
Any progress on this?

sabithpocker:
Are there any versions with `chips` and `remove`  implemented. I reached here after looking around for a while!

DmitryEfimenko:
May I ask if there is any progress on this?
Just trying to figure out if I should wait or look for something else for the feature I'm working on.

jelbourn:
This PR is being absorbed into #4912

PR #2486, opened 2017-01-01
fix(sidenav): animate content resizing for side mode.

DevVersion:
* Animates the `margin-left` and `margin-right` for the `sidenav-content` because otherwise when using `mode="side"` the sidenav content will just jump and not animate as same as the `push` mode.

### DIFF
--- a/src/lib/sidenav/sidenav-transitions.scss
+++ b/src/lib/sidenav/sidenav-transitions.scss
@@ -6,7 +6,11 @@ md-sidenav {
 }
 
 .md-sidenav-content {
-  transition: transform $swift-ease-out-duration $swift-ease-out-timing-function;
+  transition: {
+    duration: $swift-ease-out-duration;
+    timing-function: $swift-ease-out-timing-function;
+    property: transform, margin-left, margin-right;
+  }
 }
 
 .md-sidenav-backdrop.md-sidenav-shown {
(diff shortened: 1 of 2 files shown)


jelbourn:
cc @mmalerba 

@DevVersion we had this before and it was super janky on mobile.

DevVersion:
@jelbourn Hmm, on the other hand just letting the content jump would probably also feel janky.

mmalerba:
https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-behavior
The spec says that persistent drawers (equivalent of our "side" mode) should only be used on larger devices and that mobile should use temporary drawers (equivalent of our "over" mode).

I vote for adding the animation and adding a note to our docs that in addition to being the recommendation of the spec, we also suggest that they only use "over" or "push" on mobile devices for performance reasons

DevVersion:
@mmalerba I think that is a great idea. Do you want to me to include it into the `OVERVIEW.md` file?

mmalerba:
yeah, sounds good

DevVersion:
@mmalerba Just added the info to the overview file.

mmalerba:
@jelbourn is this ok with you?

DevVersion:
Ping @jelbourn for review.

willshowell:
I'm looking forward to this PR landing. Is there any progress on unblocking it?

mmalerba:
jeremy told me he wanted to test this on mobile, I just did and it looks fine to me. and again ideally people should use over or push on mobile.

DevVersion:
@mmalerba Thanks for testing it on mobile.

andrewseguin:
Sidenavs that initialize with opened are showing the content transition in.  Using #3043 to track the issue

PR #2487, opened 2017-01-01
update(overlay): use boolean-property coercion function

DevVersion:
* Replaces the coercion logic in the `hasBackdrop` method  with the `coerceBooleanProperty` utility function.


jelbourn:
LGTM

PR #2496, opened 2017-01-02
fix(input-container): prefix and suffix stretching together with parent

### LINKED ISSUE #1421: MdInput renders suffix's width relative to input width; often too wide

Bug

A suffix that is added to an `<md-input>` element should only be as wide as the suffix's contents.

The suffix's width is calculated relatively to the width of the entire input. If the input is, let's say, 650px wide, then the suffix renders to approximately 80px, although the suffix's content only has a width of, say, 24px.

Simply set the width of the `<md-input>` to a large number to see the relative sizing of the suffix.

Plunker: http://plnkr.co/edit/DnK191LGat2o3bnOyTtt?p=preview

The suffix is supposed to go to the very end of the input field. The current behaviour leaves too much white space to the right of the suffix.

Tested with Angular 2.0.0, Material 2.0.0-alpha.9 on Firefox 49.0.1, Windows 10

A good solution might be styling the div.md-input-table and its children as a flexbox in the MdInputComponent template.

### LINKED ISSUE #1881: md-input with md-prefix and width:100% has weird rendering
Bug

The prefix on md-input should consume the space enough for the content within prefix.
I made a simple codepen of how it works in angularjs: https://codepen.io/anon/pen/VmKaNv (contrast with material2 impelmentation in plunkr below)

there is a weird amount of space between the prefix and the placeholder text 

Code to reproduce buggy behavior:
```
<md-input floatingPlaceholder="false" placeholder="Search" style="width:100%;">
   <span md-prefix><md-icon>search</md-icon>&nbsp;</span>
</md-input>
```
This Plunker has a simple search box with 100% width, see the preview in a large screen to be sure of behavior.
Plunker template: http://plnkr.co/edit/nCYNQGtfoZPQ04FUFy2L

It looks ugly, also, it doesn't conform with other material design implementation (viz. in material design of angularjs)

All.

Nope.

### LINKED ISSUE #2493: Bug: md-prefix stretches together with input field
When I put prefix in a wide md-input-container, the prefix stretches together with it. 
Easy to reproduce. Just make the md-input-container wide enough:
`<md-input-container class="full-width">`
    `<span md-prefix>#</span>`
    `<input md-input formControlName="tagInput" placeholder="Tags>`
`</md-input-container>`

Have a look at the copyright and hash prefixes here:
screen shot 2017-01-02 at 14 32 54

I believe prefixes should stay the same width not dependent on the parent width.

----------
Another thing (not sure if it was already mentioned) - fields disabled via code .disable() do not display as dotted. Also visible on the screenshot (field to the left of checkbox is actually disabled).

crisbeto:
Prevents the input's suffix and prefix from stretching together with their parent node and leaving some weird spacing between themselves and the input. This was a side-effect of using `display: table-cell`.

Fixes #2493.
Fixes #1881.
Fixes #1421.

### DIFF
--- a/src/lib/input/input-container.scss
+++ b/src/lib/input/input-container.scss
@@ -30,3 +30,9 @@ md-input-container {
     }
   }
 }
+
+// Prevents the prefix and suffix from stretching together with the container.
+.md-input-prefix, .md-input-suffix {
+  width: 0.1px;
+  white-space: nowrap;
+}


src/lib/input/input-container.scss
```diff
@@ -30,3 +30,9 @@ md-input-container {
     }
   }
 }
+
+// Prevents the prefix and suffix from stretching together with the container.
+.md-input-prefix, .md-input-suffix {
+  width: 0.1px;
```
mmalerba: why is this width needed?

crisbeto: It kind of acts as a `min-width`, otherwise the table cells stretch out.

mmalerba: I see, is there a reason we have to use table as opposed to flex? I had started a PR to convert it to flex at some point, but I don't know what happened to it...

crisbeto: Not sure, IIRC @hansl implemented it like this initially, but I don't know what the reasoning is.

mmalerba: Ok, Jeremy's comment on #2581 explains why we're avoiding flex here.

PR #2499, opened 2017-01-02
fix(select): selected option not being highlighted when options are added asynchronously

### LINKED ISSUE #2497: Bug: md-select doesn't set selected value if options are initialized after the model

Bug, if ngModel of md-select is set before setting of options - select doesn't show selected value after both (options and model) are set.

Showing selected value after both (model and options) are set.

Selected value is shown if options are set before ngModel value, otherwise selected value is not shown.

http://plnkr.co/edit/yL8QJBjABGPRerqGIR2b?p=preview

You see two selects, left one shows selected value, right one - doesn't.

For ajax loaded value and options the order of initialization is not guaranteed. Without fixing this bug you need to add an additional logic to the app to init select value only after options array is set.

Tested with Angular 2.4.1, Angular material 2.0.0-beta.1.

crisbeto:
Fixes the select component not highlighting the model value when the options get added after initialization.

Fixes #2497.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -16,7 +16,7 @@ describe('MdSelect', () => {
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       imports: [MdSelectModule.forRoot(), ReactiveFormsModule, FormsModule],
-      declarations: [BasicSelect, NgModelSelect, ManySelects, NgIfSelect],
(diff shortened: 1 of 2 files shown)


crisbeto:
Thanks for the feedback @kara, it's addressed.

kara:
@crisbeto Can you rebase?  When you're done, apply the merge label so we know it's ready :)

PR #2500, opened 2017-01-02
update(viewport-ruler): cache document rectangles

DevVersion:
* The ViewportRuler now caches the document rectangle and only updates it on initialization and when the window gets resized.

**Note**: I had to remove the `implements` from the `FakeViewportRuler` because it looks like it's just impossible to implement a class with private members.

* See: [this sample (try to solve the error by keeping implements)](http://www.typescriptlang.org/play/index.html#src=class%20ViewportRuler%20%7B%0D%0A%20%20%20%20private%20_myProperty%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Aclass%20FakeViewportRuler%20implements%20ViewportRuler%20%7B%0D%0A%20%20%20%20%2F%2F%20INSERT%20HERE%0D%0A%7D)


DevVersion:
@jelbourn I will close that PR because I feel like this requires more work than just listening for the `resize` event, because the `getViewportRect` method should also update when the user scrolls.

PR #2501, opened 2017-01-02
feat(sidenav): add disableClose option

### LINKED ISSUE #2462: Sidenav disappears when escape is pressed
This is not always intended, see https://material.angular.io/components for example. Click on the sidenav and press escape to see it disappear. Cannot get the sidenav back.

DevVersion:
* Adds an attribute to the `md-sidenav` component, which allows developers to disable the closing behavior (e.g escape closing)

  Backdrop stays separate because you can disable it by using the `mode` attribute.

Closes #2462

### DIFF
--- a/src/lib/sidenav/sidenav.spec.ts
+++ b/src/lib/sidenav/sidenav.spec.ts
@@ -264,6 +264,30 @@ describe('MdSidenav', () => {
       expect(testComponent.closeCount).toBe(1);
     }));
 
+    it('should not close when pressing escape if disableClose is set', fakeAsync(() => {
+      let fixture = TestBed.createComponent(BasicTestApp);
+      let testComponent = fixture.debugElement.componentInstance;
+      let sidenav = fixture.debugElement.query(By.directive(MdSidenav)).componentInstance;
+
+      sidenav.disableClose = true;
+      sidenav.open();
+
+      fixture.detectChanges();
+      endSidenavTransition(fixture);
+      tick();
+
+      sidenav.handleKeydown({
+        keyCode: ESCAPE,
+        stopPropagation: () => {}
+      });
+
+      fixture.detectChanges();
+      endSidenavTransition(fixture);
+      tick();
+
+      expect(testComponent.closeCount).toBe(0);
+    }));
+
     it('should restore focus to the trigger element on close', fakeAsync(() => {
       let fixture = TestBed.createComponent(BasicTestApp);
       let sidenav: MdSidenav = fixture.debugElement
--- a/src/lib/sidenav/sidenav.spec.ts
+++ b/src/lib/sidenav/sidenav.spec.ts
@@ -194,7 +194,7 @@ describe('MdSidenav', () => {
       }).not.toThrow();
     }));
 
-    it('should emit the backdrop-clicked event when the backdrop is clicked', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


mmalerba:
@jelbourn I still like the idea of escapeAction and backdropAction that have values of "none", "close", or a callback function that can be used to implement custom behavior. I remember you being opposed to this when I brought it up last time, but can you remind me why? It seems like a more flexible solution to me.

jelbourn:
@mmalerba I mainly was put off by the "provide a callback" API style, since we don't do that anywhere else in the library. So rather than having `escapeAction` plus `backdropAction`, my preference would be for something like `disableClose` plus `backdropClick` Output (or the user setting up their own escape keydown handler).

mmalerba:
Couldn't we just start using this pattern here and elsewhere where it seems appropriate? The API seems clunky otherwise. You need disableCloseOnEscape, disableCloseOnBackdropClick, backdropClick, and escapePressed in order to give users the same amount of control. I think in these cases where the user wants to be able to stop the default action and supply their own the callback style API really makes more sense.

DevVersion:
@mmalerba @jelbourn How should we proceed with this PR now?

mmalerba:
I think I'm outnumbered/outranked on this one, go ahead and do what Jeremy is suggesting

mmalerba:
sorry  didn't mean to close

src/lib/sidenav/sidenav.ts
```diff
@@ -103,6 +103,12 @@ export class MdSidenav implements AfterContentInit {
   /** Mode of the sidenav; whether 'over' or 'side'. */
   @Input() mode: 'over' | 'push' | 'side' = 'over';
 
+  /** Whether the sidenav can be closed with the escape key or not. */
+  @Input()
+  get disableClose(): boolean { return this._disableClose; }
+  set disableClose(value: boolean) { this._disableClose = coerceBooleanProperty(value); }
+  private _disableClose: boolean = false;
```
jelbourn: `disableClose` should also prevent close from backdrop clicks.

While you're doing this, can you also change the `@Output` to `backdropClick`? Looks like we missed it when getting rid of dash-case stuff.

DevVersion: @jelbourn Made the changes you requested.

jelbourn:
LGTM

src/lib/sidenav/sidenav.spec.ts
```diff
@@ -414,7 +462,7 @@ class SidenavContainerTwoSidenavTestApp { }
 /** Test component that contains an MdSidenavContainer and one MdSidenav. */
 @Component({
   template: `
-    <md-sidenav-container (backdrop-clicked)="backdropClicked()">
+    <md-sidenav-container (backdropClick)="backdropClicked()">
```
christophechevalier: @maxime1992 The solution to our problem : backdropClick

christophechevalier:
Hi @jelbourn 

I think the documentation is not up to date on the Event emitted when the sidenav backdrop is clicked.
https://material.angular.io/components/component/sidenav

selection_040

Now, for use event of backdrop-clicked, it's backdropClick as I see.
https://github.com/angular/material2/pull/2501/files

:rocket:

evanjmg:
There's no class in the html added to indicate that the container is disabled on disable close. This is important for ui tweaks (psuedo selectors - X in the corner ) to tell the user whether they can close the modal or not via outside click.

PR #2504, opened 2017-01-03
build: verify aot compilation works for components

DevVersion:
* Adds a new gulp task to verify that AOT works properly for all components and modules.
* Runs the AOT testing on Travis for PR's and normal pushes.

For example when a host binding is private then the CI will error with


jelbourn:
LGTM

jelbourn:
@DevVersion as a follow-up to this PR, it would be good to AoT compile the e2e app as well if you can get that working.

PR #2510, opened 2017-01-03
chore: remove line-break escapes in firebase private key

DevVersion:
* Transforms the escaped line-breaks from the Firebase private key back into real line-breaks.
  This is necessary because the private key always contains line-breaks.


DevVersion:
@jelbourn I wasn't able to verify this change locally since I don't know how Travis CI escapes the characters in their web interface.

jelbourn:
@DevVersion did you put the key in quotes in Travis, e.g. 
image
?

DevVersion:
@jelbourn Nope, do I need to?

jelbourn:
@DevVersion I think so- I've always had to do that when entering a firebase access token

DevVersion:
@jelbourn Interesting, let me try that and restart a build.

jelbourn:
LGTM

PR #2513, opened 2017-01-04
fix(input): disabled inputs should be grayed out

kara:
Disabled md-inputs are displaying the currentColor right now (often black). They should be grayed out when disabled, as in Material 1.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -37,6 +37,10 @@
     }
   }
 
+  .md-input-element:disabled {
+    color: md-color($foreground, disabled-text);
+  }
+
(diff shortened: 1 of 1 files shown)


src/lib/input/_input-theme.scss
```diff
@@ -37,6 +37,10 @@
     }
   }
 
+  .md-input-element:disabled {
+    color: md-color($foreground, hint-text);
```
jelbourn: `disabled-text` (it's the same value, but semantically more appropriate)

kara: Ah! That's what I was looking for.

PR #2514, opened 2017-01-04
chore(input): fix selector on mdInput

kara:
Updates the selector for `mdInput` within `md-input-container` to match other attribute directives.


francisoud:
You should add a "breaking change" mention the changelog to tell people that they need to update from md-input to mdInput when moving from 2.0.0-beta.1 to the next version... otherwise they will discover it the 'hard way' ;)
I don't see any documentation change in 93b0bc9

mmalerba:
@francisoud I think the README was updated as part of this pr, but you're right we should call this out in the changelog for the next release, thank you for the reminder.

PR #2526, opened 2017-01-04
fix(radio): Make MdRadioButton change detection strategy OnPush

### LINKED ISSUE #2491: md-radio with push strategy not updateing

Bug

The radio button is updated when markForCheck is called

The radio button is one value behind

## component

import { Component, ChangeDetectorRef, ChangeDetectionStrategy } from '@angular/core';

@Component({
  selector: 'dm-settings',
  templateUrl: '../../views/game/settings.html',
  changeDetection: ChangeDetectionStrategy.OnPush
})
export class SettingsComponent {
  private value: boolean = false;

  constructor(private change: ChangeDetectorRef) {
    setTimeout(() => {
      this.value = !this.value;
      this.change.markForCheck();
    }, 1000);
  }

}

## template

&lt;md-card&gt;
  {{value}}
&lt;md-radio-group [(ngModel)]="value"&gt;
  &lt;md-radio-button [value]="true"&gt;true&lt;/md-radio-button&gt;
  &lt;md-radio-button [value]="false"&gt;false&lt;/md-radio-button&gt;
&lt;/md-radio-group&gt;
&lt;/md-card&gt;

    "@angular/common": "^2.4.1",
    "@angular/core": "^2.4.1",
    "@angular/material": "^2.0.0-beta.1",
    "@angular/platform-browser": "^2.4.1",
    "@angular/platform-browser-dynamic": "^2.4.1",
    "@angular/forms": "^2.4.1",

tinayuangao:
Fixes #2491

R: @jelbourn

### DIFF
--- a/src/lib/radio/radio.ts
+++ b/src/lib/radio/radio.ts
@@ -283,7 +283,7 @@ export class MdRadioGroup extends _MdRadioGroupMixinBase
     '[class.mat-radio-checked]': 'checked',
     '[class.mat-radio-disabled]': 'disabled',
     '[attr.id]': 'id',
-  }
+  },
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
@@ -304,39 +304,6 @@ export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
   @Input()
   get disableRipple(): boolean { return this._disableRipple; }
   set disableRipple(value) { this._disableRipple = coerceBooleanProperty(value); }
-
-  /**
-   * Event emitted when the checked state of this radio button changes.
-   * Change events are only emitted when the value changes due to user interaction with
-   * the radio button (the same behavior as `<input type-"radio">`).
-   */
-  @Output()
-  change: EventEmitter<MdRadioChange> = new EventEmitter<MdRadioChange>();
-
-  /** The native `<input type=radio>` element */
-  @ViewChild('input') _inputElement: ElementRef;
-
-  constructor(@Optional() radioGroup: MdRadioGroup,
-              private _elementRef: ElementRef,
-              private _renderer: Renderer,
-              private _changeDetector: ChangeDetectorRef,
-              public radioDispatcher: UniqueSelectionDispatcher) {
-    // Assertions. Ideally these should be stripped out by the compiler.
-    // TODO(jelbourn): Assert that there's no name binding AND a parent radio group.
-
-    this.radioGroup = radioGroup;
-
-    radioDispatcher.listen((id: string, name: string) => {
-      if (id != this.id && name == this.name) {
-        this.checked = false;
-      }
-    });
-  }
-
-  /** ID of the native input element inside `<md-radio-button>` */
-  get inputId(): string {
-    return `${this.id}-input`;
-  }
 
   /** Whether this radio button is checked. */
   @Input()
--- a/src/lib/radio/radio.ts
+++ b/src/lib/radio/radio.ts
@@ -88,6 +88,9 @@ export class MdRadioGroup extends _MdRadioGroupMixinBase
   /** Whether the `value` has been set to its initial value. */
   private _isInitialized: boolean = false;
 
+  /** Whether the labels should appear after or before the radio-buttons. Defaults to 'after' */
+  private _labelPosition: 'before' | 'after' = 'after';
+
   /** The method to be called in order to update ngModel */
   _controlValueAccessorChangeFn: (value: any) => void = (value) => {};
 
@@ -131,8 +134,19 @@ export class MdRadioGroup extends _MdRadioGroupMixinBase
     this.labelPosition = (v == 'start') ? 'after' : 'before';
   }
 
+
   /** Whether the labels should appear after or before the radio-buttons. Defaults to 'after' */
-  @Input() labelPosition: 'before' | 'after' = 'after';
+  @Input()
+  get labelPosition() {
+    return this._labelPosition;
+  }
+
+  set labelPosition(v) {
+    this._labelPosition = (v == 'before') ? 'before' : 'after';
+    if (this._radios) {
+      this._radios.forEach(radio => radio.groupValueChanged());
+    }
+  }
 
   /** Value of the radio button. */
   @Input()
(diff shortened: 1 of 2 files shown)


src/lib/radio/radio.ts
```diff
@@ -124,14 +126,31 @@ export class MdRadioGroup implements AfterContentInit, ControlValueAccessor {
...
   /** Whether the labels should appear after or before the radio-buttons. Defaults to 'after' */
-  @Input() labelPosition: 'before' | 'after' = 'after';
+  _labelPosition: 'before' | 'after' = 'after';
+  @Input()
+  get labelPosition(): 'before' | 'after' {
+    return this._labelPosition;
+  }
+  set labelPosition(v: 'before' | 'after') {
+    this._labelPosition = v
+    if (this._radios) {
```
jelbourn: This shouldn't be necessary; the individual radios have a getter that reads the value from the group if it exists without clobbering its own value.

src/lib/radio/radio.ts
```diff
@@ -280,10 +300,26 @@ export class MdRadioButton implements OnInit {
...
   /** Used to set the 'aria-label' attribute on the underlying input element. */
-  @Input('aria-label') ariaLabel: string;
+  _ariaLabel: string;
+  @Input('aria-label')
+  get ariaLabel(): string {
+    return this._ariaLabel;
+  }
+  set ariaLabel(value: string) {
+    this._ariaLabel = value;
+    this._change.markForCheck();
```
jelbourn: I didn't think calling `markForCheck` for every property shouldn't be necessary; for anything that is an `@Input` Angular will do change detection as needed when the value is changed via a template binding. The only thing I know of that needs to change is that the `writeValue` function needs to call `markForCheck`, since the template binding is on `ngModel` (which isn't an `@Input` of the component).

tinayuangao: Then there's a bug for OnPush since the properties are not updated when their values changed. 

The `writeValue` function is in `MdRadioButtonGroup` which is a `directive` and has no `changeDetection` strategy.

jelbourn:
@vicb Is this the way `OnPush` is supposed to work right now?

src/lib/radio/radio.ts
vicb: @jelbourn your understanding is correct, an `@Input()` property updated via a binding would mark the components for change detection. (it would not if it accessed directly, ie `myDirective.inputProperty = ...` in some component).

@tinayuangao 
> Then there's a bug for OnPush since the properties are not updated when their values changed.

I'm not sure to get what you mean here.

Only components are nodes in the change detection tree. Directives are checked by the CD together with their component and I expect that setting any of their input would have the same effect as setting a component input - I have to double check this case.

tinayuangao: For `labelPosition` and `disabled`, group buttons' values depend on group's values. When the group's `disabled` changed, the button's `get disabled` also changed but `set disabled` is not called and button's `_disabled` is not changed. We need to let buttons `markForCheck()` when the group value changed. I added a `_groupValueChanged()` function but not sure it's the right way to do.

jelbourn:
@tinayuangao I'm trying to clean up old PRs; can you rebase this? 

Revisiting this PR, I think what's missing is, for each `markForCheck`, we should add a comment that explains what in particular is changing and/or why `markForCheck` is necessary

tinayuangao:
Synced and added some comments

src/lib/radio/radio.ts
```diff
@@ -364,9 +401,10 @@ export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
   get disabled(): boolean {
     return this._disabled || (this.radioGroup != null && this.radioGroup.disabled);
   }
-
   set disabled(value: boolean) {
     this._disabled = coerceBooleanProperty(value);
+    // Update rippleDisabled
+    this._changeDetector.markForCheck();
```
jelbourn: Could you elaborate on what this comment means? I don't see why you would need this in the setter since `disabled` is an `@Input`

src/lib/radio/radio.ts
```diff
@@ -129,8 +137,19 @@ export class MdRadioGroup extends _MdRadioGroupMixinBase
...
+  @Input()
+  get labelPosition() {
+    return this._labelPosition;
+  }
+
+  set labelPosition(v) {
+    this._labelPosition = (v == 'before') ? 'before' : 'after';
+    if (this._radios) {
+      this._radios.forEach(radio => radio._groupValueChanged());
+    }
```
jelbourn: I think this part should be broken out into its own function
```ts
 if (this._radios) {
    this._radios.forEach(radio => radio._groupValueChanged());
  }
```
Something like `_markRadiosForCheck`

src/lib/radio/radio.ts
```diff
@@ -427,6 +466,12 @@ export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
     this._focusOriginMonitor.focusVia(this._inputElement.nativeElement, this._renderer, 'keyboard');
   }
 
+  _groupValueChanged() {
```
jelbourn: I would just call this function `_markForCheck` with a comment like 
```ts
/**
 * Marks the radio button as needing checking for change detection. 
 * This method is exposed because the parent radio group will directly 
 * update bound properties of the radio button.
 */
```

jelbourn:
@tinayuangao poke on this one 👈

tinayuangao:

src/lib/radio/radio.ts
```diff
@@ -160,6 +177,21 @@ export class MdRadioGroup extends _MdRadioGroupMixinBase
...
   }
 
+  /** Whether the radio group is diabled */
+  @Input()
+  get disabled() { return this._disabled; }
+  set disabled(value) {
+    this._disabled = value;
+    if (this._radios) {
+      // Update radios disabled state
+      this._radios.forEach((r) => r._markForCheck());
```
jelbourn: Call `_markRadiosForCheck()` here?

src/lib/radio/radio.ts
```diff
@@ -160,6 +177,21 @@ export class MdRadioGroup extends _MdRadioGroupMixinBase
...
+    }
+  }
+
+  constructor(private _change: ChangeDetectorRef) {
```
jelbourn: I would call this `_changeDetector`, since `_change` could be confused with the change event

PR #2527, opened 2017-01-04
fix(slider): round decimals in the thumb label

### LINKED ISSUE #2511: slider decimal point
Bug

when the step attribute is set to 0.1 it should increment the value accordingly
13.9
23.2

the component is displaying huge fractions such as 
13.900000000000006
23.299999999999997

<md-slider [min]="-50" [max]="50" [step]="0.1" thumbLabel [(ngModel)]="annualReturn"></md-slider>

http://plnkr.co/edit/oeToNQlCIT3WZebLDAg5?p=preview

"@angular/material": "^2.0.0-alpha.11-3",

crisbeto:
Rounds down the thumb label value to a maximum of one decimal place, in order to avoid displaying weird rounding errors.

Fixes #2511.

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -249,6 +249,9 @@ export class MdSlider implements ControlValueAccessor {
 
   /** The value to be used for display purposes. */
   get displayValue(): string|number {
+    // Note that this could be improved further by rounding something like 0.999 to 0.99 or
+    // 0.899 to 0.89, however it is very performance sensitive, because it gets called on
+    // every change detection cycle.
     if (this._roundLabelTo && this.value % 1 !== 0) {
       return this.value.toFixed(this._roundLabelTo);
     }
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -148,7 +148,7 @@ export class MdSlider implements ControlValueAccessor {
   _isActive: boolean = false;
 
   /** Decimal places to round to, based on the step amount. */
-  private _roundTo: number;
+  private _roundLabelTo: number;
 
   private _step: number = 1;
 
@@ -159,7 +159,7 @@ export class MdSlider implements ControlValueAccessor {
     this._step = coerceNumberProperty(v, this._step);
 
     if (this._step % 1 !== 0) {
-      this._roundTo = this._step.toString().split('.').pop().length;
+      this._roundLabelTo = this._step.toString().split('.').pop().length;
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/slider/slider.ts
```diff
@@ -226,6 +226,12 @@ export class MdSlider implements ControlValueAccessor {
   set vertical(value: any) { this._vertical = coerceBooleanProperty(value); }
   private _vertical = false;
 
+  /** The value to be used for display purposes. */
+  get displayValue(): string|number {
+    // Skip adding the decimal part if the number is whole.
+    return this.value % 1 === 0 ? this.value : this.value.toFixed(1);
```
jelbourn: Using one decimal place is too prescriptive, since having a step of, say `0.05` between `0` and `1` is valid. It would be nice to have the slider intelligently figure out the appropriate number of decimal places to show based on the step value.

jelbourn:
cc @mmalerba

src/lib/slider/slider.ts
crisbeto: That's what I was thinking initially, however anything longer than 3 characters doesn't look right with the current setup.

mmalerba: I would round it based on the step and consider the 3 characters thing a separate bug (we'll still have issues with 1000 even with no decimals)

crisbeto: Alright, I'll redo it.

crisbeto:
@mmalerba I've switched it over to determine the rounding based on the `step`.

src/lib/slider/slider.ts
```diff
@@ -226,6 +235,15 @@ export class MdSlider implements ControlValueAccessor {
   set vertical(value: any) { this._vertical = coerceBooleanProperty(value); }
   private _vertical = false;
 
+  /** The value to be used for display purposes. */
+  get displayValue(): string|number {
+    if (this._roundLabelTo && this.value % 1 !== 0) {
+      return this.value.toFixed(this._roundLabelTo);
```
mmalerba: I think there's a couple cases this misses:
.999 to 2 places
.899 to 2 places
.001 to 2 places

crisbeto: I wanted to keep this as simple as possible, because it gets called on every change detection.

mmalerba: Ok, well at least leave a note so that if this issue pops up we know that it was an intentional choice

crisbeto: Done.

src/lib/slider/slider.ts
```diff
@@ -226,6 +235,18 @@ export class MdSlider implements ControlValueAccessor {
   set vertical(value: any) { this._vertical = coerceBooleanProperty(value); }
   private _vertical = false;
 
+  /** The value to be used for display purposes. */
+  get displayValue(): string|number {
+    // Note that this could be improved further by rounding something like 0.999 to 0.99 or
```
mmalerba: .999 rounds to 1, .899 rounds to .9 (when rounding to 2 decimal places)

crisbeto: Sorry, apparently I forgot how rounding works. Fixed now.

mmalerba:
lgtm, lint seems to be failing but i don't think its related to this PR

PR #2532, opened 2017-01-05
feat(screenshot): Add screenshot function to e2e test (button and checkbox)

tinayuangao:
R: @jelbourn

### DIFF
--- a/e2e/components/button/button.e2e.ts
+++ b/e2e/components/button/button.e2e.ts
@@ -1,4 +1,5 @@
 import {browser, by, element} from 'protractor';
+import {screenshot} from '../../screenshot';
 
 describe('button', function () {
   describe('disabling behavior', function () {
@@ -6,12 +7,17 @@ describe('button', function () {
       browser.get('/button');
     });
     it('should prevent click handlers from executing when disabled', function () {
+      let testname =
+        `button disabling behavior should prevent click handlers from executing when disabled`;
+      screenshot(testname);
       element(by.id('test-button')).click();
       expect(element(by.id('click-counter')).getText()).toEqual('1');
+      screenshot(testname + ' click once');
 
       element(by.id('disable-toggle')).click();
       element(by.id('test-button')).click();
       expect(element(by.id('click-counter')).getText()).toEqual('1');
+      screenshot(testname + 'click disabled');
     });
   });
 });
--- a/e2e/components/button/button.e2e.ts
+++ b/e2e/components/button/button.e2e.ts
@@ -1,23 +1,22 @@
 import {browser, by, element} from 'protractor';
 import {screenshot} from '../../screenshot';
 
+
 describe('button', function () {
   describe('disabling behavior', function () {
     beforeEach(function() {
       browser.get('/button');
     });
+
     it('should prevent click handlers from executing when disabled', function () {
-      let testname =
-        `button disabling behavior should prevent click handlers from executing when disabled`;
-      screenshot(testname);
       element(by.id('test-button')).click();
       expect(element(by.id('click-counter')).getText()).toEqual('1');
-      screenshot(testname + ' click once');
+      screenshot('clicked once');
 
       element(by.id('disable-toggle')).click();
       element(by.id('test-button')).click();
       expect(element(by.id('click-counter')).getText()).toEqual('1');
-      screenshot(testname + 'click disabled');
+      screenshot('click disabled');
     });
   });
 });
--- a/e2e/screenshot.ts
+++ b/e2e/screenshot.ts
@@ -1,0 +1,52 @@
+import * as fs from 'fs';
+import * as gulp from 'gulp';
+import * as path from 'path';
+import {browser} from 'protractor';
+
+export class Screenshot {
+  id: string;
+
+  /**
+   * The filename used to store the screenshot
+   * @returns {string}
+   */
+  get filename() {
+    return this.id
+        .toLowerCase()
+        .replace(/[ :\/]/g, '_')
+        .replace(/[^/a-z0-9_]+/g, '')
+      + '.screenshot.png';
+  }
+
+  /**
+   * The full path to the screenshot
+   * @returns {string}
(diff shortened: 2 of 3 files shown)


e2e/screenshot.ts
```diff
@@ -0,0 +1,52 @@
...
+  id: string;
+
+  /**
+   * The filename used to store the screenshot
+   * @returns {string}
+   */
+  get filename() {
+    return this.id
+        .toLowerCase()
+        .replace(/[ :\/]/g, '_')
```
jelbourn: Can change the space to `\s` to capture all whitespace.

e2e/screenshot.ts
```diff
@@ -0,0 +1,52 @@
...
+import {browser} from 'protractor';
+
+export class Screenshot {
+  id: string;
+
+  /**
+   * The filename used to store the screenshot
+   * @returns {string}
+   */
+  get filename() {
```
jelbourn: Can do `get filename(): string` and remove the type from the JsDoc
(here and elsewhere for return types and params)

e2e/screenshot.ts
```diff
@@ -0,0 +1,52 @@
...
+        .replace(/[ :\/]/g, '_')
+        .replace(/[^/a-z0-9_]+/g, '')
+      + '.screenshot.png';
+  }
+
+  /**
+   * The full path to the screenshot
+   * @returns {string}
+   */
+  get path() {
```
jelbourn: Let's call this `fullPath`

e2e/components/button/button.e2e.ts
```diff
@@ -1,17 +1,23 @@
 import {browser, by, element} from 'protractor';
+import {screenshot} from '../../screenshot';
 
 describe('button', function () {
   describe('disabling behavior', function () {
     beforeEach(function() {
       browser.get('/button');
     });
     it('should prevent click handlers from executing when disabled', function () {
+      let testname =
```
jelbourn: Jasmine 2.0 made it more difficult to get the current test name, but it's still possible by adding a custom reporter. [Example](https://github.com/jasmine/jasmine/issues/611#issuecomment-268504723).

Could could add a reporter that just updates a value which is then read in `screenshot.ts`

e2e/screenshot.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Replaces the existing screenshot with the newly generated one.
+   */
+  storeScreenshot(png: any) {
+    var dir = path.resolve(__dirname, '..', 'screenshots');
+    if (!fs.existsSync(dir)) {
+      fs.mkdirSync(dir, 0o744);
+    }
+    fs.writeFileSync(this.path, png, {encoding: 'base64'});
```
jelbourn: Why `base64` encoding? Shouldn't this be `binary`?

e2e/screenshot.ts
```diff
@@ -0,0 +1,52 @@
...
+    browser.takeScreenshot().then(png => this.storeScreenshot(png));
+  }
+
+  /**
+   * Replaces the existing screenshot with the newly generated one.
+   */
+  storeScreenshot(png: any) {
+    var dir = path.resolve(__dirname, '..', 'screenshots');
+    if (!fs.existsSync(dir)) {
+      fs.mkdirSync(dir, 0o744);
```
jelbourn: Why octal instead of hexadecimal?

e2e/screenshot.ts
tinayuangao: The output of `binary` cannot be opened as an image.

e2e/components/button/button.e2e.ts
tinayuangao: Added a reporter and we can use the test name in `Screenshot` now

tinayuangao:
Fixed test on travis-ci and sauce labs. Please take another look. Thanks!

e2e/screenshot.ts
```diff
@@ -0,0 +1,54 @@
...
+    jasmine.getEnv().currentSpec = result.fullName;
+  };
+  jasmine.getEnv().addReporter(reporter);
+}
+
+initializeEnvironment(jasmine);
+
+export class Screenshot {
+  id: string;
+  dir: string = '/tmp/angular-material2-build/screenshots/';
```
jelbourn: Make this a file-level constant, e.g.,
```ts
/** Directory into which screenshots will be written. */
const OUTPUT_DIR = '/tmp/angular-material2-build/screenshots/';
```

e2e/screenshot.ts
```diff
@@ -0,0 +1,54 @@
+import * as fs from 'fs';
+import * as gulp from 'gulp';
+import * as path from 'path';
+import {browser} from 'protractor';
+
+
+function initializeEnvironment(jasmine: any) {
+  var reporter = new jasmine.JsApiReporter({});
+  reporter.specStarted = function(result: any) {
+    jasmine.getEnv().currentSpec = result.fullName;
```
jelbourn: Instead of storing this to `jasmine.getEnv()`, would it work to create a variable at the file level and update that? E.g.,
```ts
let currentJasmineSpecName = '';

function initializeEnvironment(jasmine: any) {
  jasmine.getEnv().addReporter(new jasmine.JsApiReporter({
    specStarted: (result: any) => {
      currentJasmineSpecName = result.fullName;
    }
  }));
}
```

jelbourn: I also just realized: we should make sure that the test results are still being reported normally with the custom report installed

tinayuangao: Changed to file-level variable. 

Constructor `JsApiReporter({specStarted})` is not working. 

Tested the test results are being reported normally.

DevVersion:
@jelbourn @tinayuangao I just have some questions:
* How would you look at the screenshots in the CI?
* Is this the base for screenshot-diffs or is it just to look at errors in E2E tests?  
  > (Asking because Saucelabs can already do screenshots)

PR #2538, opened 2017-01-05
feat(viewport-ruler): cache document client rect

DevVersion:
* Now the `ViewportRuler` service, which is used in many components, caches the documents client rect to avoid frequent recalculation of styles in the browser (basically each click for ripples etc.)

  > The `ScrollDispatcher` is used as the service which triggers an update of the cached rectangles because it listens to the `resize` and `scroll` events.

**Note**: It's important that the `ViewportRuler` also updates the rectangle for scroll events, because the ViewportRuler is also responsible for the scroll position of the viewport.

Debouncing the `ScrollDispatcher` will be a follow-up PR

### DIFF
--- a/src/lib/core/overlay/position/connected-position-strategy.spec.ts
+++ b/src/lib/core/overlay/position/connected-position-strategy.spec.ts
@@ -1,10 +1,11 @@
 import {ElementRef} from '@angular/core';
 import {ConnectedPositionStrategy} from './connected-position-strategy';
-import {ViewportRuler} from './viewport-ruler';
+import {ViewportRuler, VIEWPORT_RULER_PROVIDER} from './viewport-ruler';
 import {OverlayPositionBuilder} from './overlay-position-builder';
 import {ConnectedOverlayPositionChange} from './connected-position';
 import {Scrollable} from '../scroll/scrollable';
 import {Subscription} from 'rxjs';
+import {TestBed, inject} from '@angular/core/testing';
 import Spy = jasmine.Spy;
 
 
@@ -17,6 +18,16 @@ const DEFAULT_WIDTH = 60;
 // for tests on CI (both SauceLabs and Browserstack).
 
 describe('ConnectedPositionStrategy', () => {
+
+  let viewportRuler: ViewportRuler;
+
+  beforeEach(() => TestBed.configureTestingModule({
+    providers: [VIEWPORT_RULER_PROVIDER]
+  }));
+
+  beforeEach(inject([ViewportRuler], (_ruler: ViewportRuler) => {
+    viewportRuler = _ruler;
+  }));
 
   describe('with origin on document body', () => {
     const ORIGIN_HEIGHT = DEFAULT_HEIGHT;
@@ -48,7 +59,7 @@ describe('ConnectedPositionStrategy', () => {
       overlayContainerElement.appendChild(overlayElement);
 
       fakeElementRef = new FakeElementRef(originElement);
-      positionBuilder = new OverlayPositionBuilder(new ViewportRuler());
+      positionBuilder = new OverlayPositionBuilder(viewportRuler);
     });
 
     afterEach(() => {
@@ -457,7 +468,7 @@ describe('ConnectedPositionStrategy', () => {
       scrollable.appendChild(originElement);
 
       // Create a strategy with knowledge of the scrollable container
-      let positionBuilder = new OverlayPositionBuilder(new ViewportRuler());
+      let positionBuilder = new OverlayPositionBuilder(viewportRuler);
       let fakeElementRef = new FakeElementRef(originElement);
       strategy = positionBuilder.connectedTo(
           fakeElementRef,
--- a/src/lib/core/overlay/position/connected-position-strategy.spec.ts
+++ b/src/lib/core/overlay/position/connected-position-strategy.spec.ts
@@ -7,6 +7,7 @@ import {Scrollable} from '../scroll/scrollable';
 import {Subscription} from 'rxjs';
 import {TestBed, inject} from '@angular/core/testing';
 import Spy = jasmine.Spy;
+import {SCROLL_DISPATCHER_PROVIDER} from '../scroll/scroll-dispatcher';
 
 
 // Default width and height of the overlay and origin panels throughout these tests.
@@ -22,7 +23,7 @@ describe('ConnectedPositionStrategy', () => {
   let viewportRuler: ViewportRuler;
 
   beforeEach(() => TestBed.configureTestingModule({
-    providers: [VIEWPORT_RULER_PROVIDER]
+    providers: [VIEWPORT_RULER_PROVIDER, SCROLL_DISPATCHER_PROVIDER]
   }));
 
   beforeEach(inject([ViewportRuler], (_ruler: ViewportRuler) => {
--- a/src/lib/core/overlay/position/viewport-ruler.spec.ts
+++ b/src/lib/core/overlay/position/viewport-ruler.spec.ts
@@ -1,5 +1,5 @@
-import {ViewportRuler} from './viewport-ruler';
-
+import {ViewportRuler, VIEWPORT_RULER_PROVIDER} from './viewport-ruler';
+import {TestBed, inject} from '@angular/core/testing';
 
 // For all tests, we assume the browser window is 1024x786 (outerWidth x outerHeight).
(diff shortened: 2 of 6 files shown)


src/lib/core/overlay/position/viewport-ruler.ts
```diff
@@ -8,12 +9,23 @@ import {Injectable} from '@angular/core';
...
+  /** Cached document client rectangle. */
+  private _documentRect: ClientRect;
+
+  constructor(scrollDispatcher: ScrollDispatcher) {
+
+    // Initially update the document rectangle.
+    this.updateDocumentRectangle();
+
+    // Subscribe to scroll and resize events and update the document rectangle on changes.
+    scrollDispatcher
```
mmalerba: I think this can all fit on one line

src/lib/core/overlay/position/viewport-ruler.ts
```diff
@@ -8,12 +9,23 @@ import {Injectable} from '@angular/core';
...
 @Injectable()
 export class ViewportRuler {
-  // TODO(jelbourn): cache the document's bounding rect and only update it when the window
-  // is resized (debounced).
 
+  /** Cached document client rectangle. */
+  private _documentRect: ClientRect;
+
+  constructor(scrollDispatcher: ScrollDispatcher) {
+
```
mmalerba: nit: I can't stand extra whitespace, you can leave it if you want but I'll definitely delete it if I ever edit this file :p

src/lib/core/overlay/position/viewport-ruler.ts
```diff
@@ -55,4 +66,13 @@ export class ViewportRuler {
 
     return {top, left};
   }
+
+  /**
+   * @internal
+   * Updates the cached client rectangle of the document element.
+   */
+  updateDocumentRectangle() {
```
mmalerba: I think our convention is to name internal methods with _, if that's not possible we add a @docs-private comment. IIRC marking stuff @internal caused issues in google3

src/lib/core/overlay/position/viewport-ruler.spec.ts
```diff
@@ -35,7 +40,9 @@ describe('ViewportRuler', () => {
 
   it('should get the viewport bounds when the page is scrolled', () => {
     document.body.appendChild(veryLargeElement);
+
     scrollTo(1500, 2000);
+    ruler.updateDocumentRectangle();
```
mmalerba: comment why

src/lib/core/ripple/ripple.spec.ts
```diff
@@ -228,6 +232,8 @@ describe('MdRipple', () => {
         document.documentElement.scrollTop = pageScrollTop;
         // Mobile safari
         window.scrollTo(pageScrollLeft, pageScrollTop);
+        // Force an update of the cached rectangle in the ViewportRuler.
```
mmalerba: ... b/c IE

src/lib/core/overlay/position/viewport-ruler.ts
jelbourn: Yep; using `@internal` causes issues with the Closure Compiler, which is why we have the `@docs-private` annotation. It's easier to just use an underscore prefix, though (without the `private` access modifier if it's not actually private).

I would also call this something like `cacheViewportGeometry`; using the "update" verb implies that calling this function will enact some change on the document rectangle instead of reading (and caching) it.

src/lib/core/overlay/position/connected-position-strategy.spec.ts
```diff
@@ -29,7 +41,11 @@ describe('ConnectedPositionStrategy', () => {
     let overlayContainerElement: HTMLElement;
     let strategy: ConnectedPositionStrategy;
     let fakeElementRef: ElementRef;
-    let fakeViewportRuler: FakeViewportRuler;
+    // TODO: The `any` type is used here because the FakeViewportRuler can't extend or implement
+    // the real ViewportRuler class because Typescript complains when using a class as an interface.
+    // Also we don't want to extend the ViewportRuler because we'd need to call the constructor
+    // in the superclass. In short: We should use Dependency Injection to fake the ViewportRuler.
+    let fakeViewportRuler: any;
```
jelbourn: TypeScript is fine with doing `implements RealClass`; it's one of the language features. The problem is the new `private` member, which causes any other implementation to automatically fail to match. You should be able to resolve this by marking the private member as optional ([example](http://www.typescriptlang.org/play/#src=class%20ViewportRuler%20%7B%0A%20%20%20%20private%20_cache%3F%3A%20number%3B%0A%0A%20%20%20%20measure()%3A%20number%20%7B%0A%20%20%20%20%20%20%20%20return%207%3B%0A%20%20%20%20%7D%0A%7D%0A%0Aclass%20FakeViewportRuler%20implements%20ViewportRuler%20%7B%0A%20%20%20%20measure()%20%7B%0A%20%20%20%20%20%20%20%20return%209%3B%0A%20%20%20%20%7D%0A%7D#src=class%20ViewportRuler%20%7B%0A%20%20%20%20private%20_cache%3F%3A%20number%3B%0A%0A%20%20%20%20_remember%3F()%20%7B%0A%20%20%20%20%20%20%20%20this._cache%20%3D%207%3B%0A%20%20%20%20%7D%0A%0A%20%20%20%20measure()%3A%20number%20%7B%0A%20%20%20%20%20%20%20%20return%207%3B%0A%20%20%20%20%7D%0A%7D%0A%0Aclass%20FakeViewportRuler%20implements%20ViewportRuler%20%7B%0A%20%20%20%20measure()%20%7B%0A%20%20%20%20%20%20%20%20return%209%3B%0A%20%20%20%20%7D%0A%7D)).

(and the internal method as well)

DevVersion: Oh nice! - I wasted so much time trying to solve that.. (considered it as a [challenge](http://www.typescriptlang.org/play/index.html#src=class%20ViewportRuler%20%7B%0D%0A%20%20%20%20private%20_myProperty%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Aclass%20FakeViewportRuler%20implements%20ViewportRuler%20%7B%0D%0A%20%20%20%20%2F%2F%20INSERT%20HERE%0D%0A%7D))

src/lib/core/overlay/position/viewport-ruler.ts
DevVersion: Will change that 👍

src/lib/core/overlay/position/viewport-ruler.ts
```diff
@@ -55,4 +63,12 @@ export class ViewportRuler {
 
     return {top, left};
   }
+
+  /**
+   * Caches the latest client rectangle of the document element.
```
mmalerba: nit: /** one-liners like this */

DevVersion: Changed. Had that because of the `@internal` from before

kara:
@DevVersion Can you take another look at this? Tests are failing in Google for buttons and tabs because they can't find ScrollDispatcher. Now the ViewportRuler depends on ScrollDispatcher, you may need to add it to those components as well.

DevVersion:
@kara Just confirmed that you are right. Our tests didn't cover it because of the `FakeViewportRuler` classes we had.

And yeah waiting for Jeremy's PR would make sense.

DevVersion:
@kara @jelbourn Rebased the PR on top of the #2556.

src/lib/core/overlay/position/viewport-ruler.ts
```diff
@@ -54,10 +62,16 @@ export class ViewportRuler {
...
+  /** Caches the latest client rectangle of the document element. */
+  _cacheViewportGeometry?() {
+    this._documentRect = document.documentElement.getBoundingClientRect();
+  }
+
 }
 
 export function VIEWPORT_RULER_PROVIDER_FACTORY(parentDispatcher: ViewportRuler) {
-  return parentDispatcher || new ViewportRuler();
+  return parentDispatcher || new ViewportRuler(new ScrollDispatcher());
```
jelbourn: Instead of doing `new ScrollDispatcher()`, it should inject the existing scroll dispatcher. 
(see `MdIconRegistry` for an example)

DevVersion: Done - Thanks for the help on Slack again.

Issue #1517, opened 2016-10-18
Rename material2 to material in order to be semver compliant ?


maxime1992:
#### Bug, feature request, or proposal:
Proposal

#### What is the use-case or motivation for changing an existing behavior?
Angular is now using semver and angular V3 is on his way (6 months I heard).

from http://angularjs.blogspot.fr/2016/09/angular2-final.html :
>**Semantic Versioning**
We heard loud and clear that our RC labeling was confusing. To make it easy to manage dependencies on stable Angular releases, starting today with Angular 2.0.0, we will move to semantic versioning.  Angular versioning will then follow the MAJOR.MINOR.PATCH scheme as described by semver:
the MAJOR version gets incremented when incompatible API changes are made to stable APIs,
the MINOR version gets incremented when backwards-compatible functionality are added,
the PATCH version gets incremented when backwards-compatible bug are fixed.

It will be weird to have != names between angular and his library (angular 3 and material2 ?).

RoxKilly:
`material2` was never part of the library's name, although it is in that folder on `github`.  With `alpha 9`, the library moved from [`@angular2-material`](https://www.npmjs.com/~angular2-material) to simply [`@angular/material`](https://www.npmjs.com/package/@angular/material)

jelbourn:
The `2` in the github repo isn't going away, and it's already gone from the package names.

maxime1992:
I guess I can now consider that solved :stuck_out_tongue: 

![image](https://user-images.githubusercontent.com/4950209/60801206-87e64380-a16e-11e9-959c-5b04b88e54f4.png)

PR #2542, opened 2017-01-05
docs(README): remove the "2" from Angular

juristr:
I wanted to support Igor's mission to remove the "2" from Angular on the web. Also the repo description would have to be adjusted. 😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-# Material Design for Angular 2
+# Material Design for Angular
 [![npm version](https://badge.fury.io/js/%40angular%2Fmaterial.svg)](https://www.npmjs.com/package/%40angular%2Fmaterial)
 [![Build Status](https://travis-ci.org/angular/material2.svg?branch=master)](https://travis-ci.org/angular/material2)
 [![Gitter](https://badges.gitter.im/angular/material2.svg)](https://gitter.im/angular/material2?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
-This is the home for the Angular team's Material Design components built on top of Angular 2.
+This is the home for the Angular team's Material Design components built for **Angular version 2 onwards**.
 
 #### Quick links
 [Google group](https://groups.google.com/forum/#!forum/angular-material2),
@@ -140,7 +140,7 @@ High level items planned for January 2017:
 be missing some behaviors or polish.
 
 ## The goal of Angular Material
-Our goal is to build a set of high-quality UI components built with Angular 2 and TypeScript,
+Our goal is to build a set of high-quality UI components built with Angular and TypeScript,
 following the Material Design spec. These
 components will serve as an example of how to write Angular code following best practices.
 


Splaktar:
There is already widespread confusion about `angular-material` for Angular 1.x and `@angular/material`/`material2`/`angular-material2` for Angular 2+.

This spreads across StackOverflow and Google Groups where questions about Angular Material for Angular 2+ keep getting posted in the `angular-material` for Angular 1.x forum. Then tags in StackOverflow keep getting posted with just `angular-material` when the question is about `angular-material2` and Angular 2+ and not `angular-material` for Angular 1.x.

@jelbourn has already said that the [2 in the repo name isn't going away](https://github.com/angular/material2/issues/1517#issuecomment-254574712).

It seems like this PR does some to balance this by leaving the "for **Angular version 2 onwards**" text. But it seems like more needs to be done to disambiguate these repositories in the README.

README.md
```diff
@@ -1,9 +1,9 @@
-# Material Design for Angular 2
+# Material Design for Angular
 [![npm version](https://badge.fury.io/js/%40angular%2Fmaterial.svg)](https://www.npmjs.com/package/%40angular%2Fmaterial)
 [![Build Status](https://travis-ci.org/angular/material2.svg?branch=master)](https://travis-ci.org/angular/material2)
 [![Gitter](https://badges.gitter.im/angular/material2.svg)](https://gitter.im/angular/material2?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
-This is the home for the Angular team's Material Design components built on top of Angular 2.
+This is the home for the Angular team's Material Design components built for **Angular version 2 onwards**.
```
jelbourn: How about just "for Angular v2.0.0+"  ?

juristr: If I understood them properly, @igorminar as well as @StephenFluin said to me one should mention "version 2" and not just "v2" to make it more explicit. More correct even, "Angular" and "version" should not be in the same sentence. So it should maybe be "for version 2+". I just thought that would create confusion with material 1 & 2...

IgorMinar:
I spoke with @jelbourn about the branding issue in the light of the #itsjustangular effort.

Unrelated to this PR, @juleskremer and I called for a meeting with Angular Devrel team to create a set of naming / branding guidelines for all things Angular. This will include how to deal with Angular 1 naming and branding (including angularjs.org) and other things like Angular Material.

This meeting will happen early next week when everyone is back in the office. Until then, let's hold any work on this until we get these guidelines. Otherwise we will likely need to redo this again in a few days. Sounds good?

juristr:
:+1: of course :smiley:

andrewseguin:
This change is a duplicate of the merged #2805

Splaktar:
This has been causing quite a bit of confusion with my clients. They keep searching for Angular Material and finding material.angularjs.org and lots of references to AngularJS Material instead of Angular Material(2). A number of them have raised concerns about the library not working or wanting to consider another UI library because of these frustrations. So far I have been able to explain things, but it has certainly come at a cost of time and frustrated clients.

PR #2546, opened 2017-01-05
refactor(dialog): better handling of scrollable content

### LINKED ISSUE #2481: Dialogue - overflow scrolling on small screens and margins
The current dialogue does not seem to support sizing and scrolling on smaller screen sizes are devices. This was possible on the material 1 dialogue. If the vertical height of the screen was incredibly small scrollbars would appear see this screenshot:

image 

With version 2 - the dialogue is pushed off the screen and no scrollbars ever appear in the middle. It does not maintain any margins on the top or bottom.

image

Maybe there is this a means this can be achieved but the examples are quite primitive.

When screen is decreased in size scrollbars should appear for the content.

The current dialogue is pushed off the screen with no margin at the top or bottom. The content remains the same size

https://plnkr.co/edit/DrI42Su0v6OntJkjUNOI?p=preview

The current modal is only useful for very simple use cases.

Latest.

### LINKED ISSUE #2775: Dialog using title, content, actions does not work well on small height screens

When using a dialog with the md-dialog-title, md-dialog-content, and md-dialog-actions directives, the dialog does not correctly scale with the screen when the height is reduced.

The entire dialog should fit into the window while the content height shrinks and is scrollable.

The dialog doesnt quite fit all the way into the window, extends barely past the window

Go into the plunker and shrink the height of the window until the dialog is cut off

https://plnkr.co/edit/LeINuG6ER11R0k8clSYJ?p=preview

crisbeto:
Currently, the dialog's scrollable content (md-dialog-content) is limited to 65% of the viewport height, however on smaller screens the dialog still ends up being too high. This proposal reworks the `md-dialog-content` to make it take up all of the remaining height, instead of being hardcoded to 65%. The max height is provided by the wrapper instead.

Fixes #2481.
Fixes #2775.


src/lib/dialog/dialog-container.ts
```diff
@@ -60,6 +62,16 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
...
 
     let attachResult = this._portalHost.attachComponentPortal(portal);
+    let componentElement = attachResult.location.nativeElement;
+
+    // Add a class that we can use for styling the root element.
+    this._renderer.setElementClass(componentElement, 'md-dialog-root', true);
+
+    // Add flexbox styling if the user is using the `md-dialog-content`.
+    if ('querySelector' in componentElement) {
+      this._renderer.setElementClass(componentElement, 'md-dialog-root-flex',
```
jelbourn: I'm not sure we should do this. This will set `display: flex` on the host element of the loaded component, which will potentially change the layout of the contents of that element in a way that the user did not intend; we should generally avoid setting any styles to elements Material doesn't own. 

Why not add an element to the dialog-container template for this (and for `md-dialog-root`)?

crisbeto: My reasoning with using flex is that if the user has their content in a md-dialog-content, then it's safe to add flexbox, because they, most likely, don't have any other content at the root (apart from md-dialog-title and md-dialog-actions). 

As for adding an extra element, I'm not sure I get what you mean. Isn't this what md-dialog-container already does?

crisbeto:
Hadn't looked at this one in a while and it looks like now there's a possibility for the `md-dialog-actions` to be clipped and also they don't get pushed to the bottom if there's a user-specified height. I'll rework it.

src/lib/dialog/dialog-container.ts
jelbourn: Coming back to this PR: which element exactly is this class applied to?

crisbeto: I'm keeping this around for tracking. I've been working on a different PR that does the something similar, but with a nicer approach. I haven't been able to get it to work properly cross-browser yet.

jelbourn:
Closing this PR for now; @crisbeto feel free to open a new one

Issue #2523, opened 2017-01-04
CI: Flakiness for Saucelabs and Browserstacks


DevVersion:
Sometimes we experience random build issues with Saucelabs or Browserstack on Travis CI.

> I'm assuming that the problem is that we run too many builds concurrently.

Regardless of having the `unlimited concurrency` license, I think that the builds are affected by it. 

> We experienced similar things in M1, when we ran our tests of different modes. 
After switching it back to one mode and running the tests sequential the builds ran fine.

Also we use the same license between M1 and M2, and who knows what other projects.

@jelbourn I recommend temporary disabling the optional builds (for testing) and also probably using a license only for M2 (then we can also show `supported browsers` banners)

PR #2549, opened 2017-01-06
chore: disable optional travis modes

DevVersion:
* Temporary disable optional travis modes because those always fail at the moment and also can be potential culprits for CI flakiness on Saucelabs & Browserstack. See #2523

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,9 @@ env:
     - MODE=e2e
     - MODE=saucelabs_required
     - MODE=browserstack_required
-    - MODE=saucelabs_optional
-    - MODE=browserstack_optional
+# Temporary disable the optional builds because of potential concurrency flakiness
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -35,8 +35,9 @@ env:
     - MODE=e2e
     - MODE=saucelabs_required
     - MODE=browserstack_required
-    - MODE=saucelabs_optional
-    - MODE=browserstack_optional
+# Temporary disable the optional builds because of potential concurrency flakiness
+#    - MODE=saucelabs_optional
+#    - MODE=browserstack_optional
```
jelbourn: Rather than commenting it out, just remove them completely. Maybe someday we'll add them back.

DevVersion: Removed everything now.

PR #2555, opened 2017-01-06
test(menu): give menu space to open in alt positions

kara:
Menu tests were not actually resetting state properly. Once that was fixed by @trshafer in the first commit, it exposed that some tests were also not being staged properly to give the menu sufficient space to open within the viewport. The latter is fixed in the second commit.


trshafer:
Thanks!

jelbourn:
LGTM

PR #2556, opened 2017-01-06
feat: remove the need for forRoot on material NgModules

jelbourn:
* Dialog and snackbar use their parent's state if there is one
* Other service that we want to be singletons and have state use a `useFactory` provider to get the ancestor and provide that if it exists

cc @IgorMinar can you provide thoughts on this?

cc FYI @josephperrott @mmalerba @andrewseguin @tinayuangao

### DIFF
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -11,7 +11,6 @@ import {
 } from '@angular/core';
 import {CommonModule} from '@angular/common';
 import {MdRippleModule, coerceBooleanProperty, DefaultStyleCompatibilityModeModule} from '../core';
-import {ViewportRuler} from '../core/overlay/position/viewport-ruler';
 
 
 // TODO(jelbourn): Make the `isMouseDown` stuff done with one global listener.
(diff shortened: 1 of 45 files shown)


jelbourn:
Seeing some failures w/ AoT that don't quite make sense to me. Consulting w/ core team.

src/lib/button/button.ts
```diff
@@ -167,10 +166,11 @@ export class MdAnchor extends MdButton {
   declarations: [MdButton, MdAnchor],
```
jelbourn: Nope, it is not included by way of `MdRippleModule`

kara: Cool, just double-checking

jelbourn:
Also cc FYI @DevVersion @crisbeto @EladBezalel

IgorMinar:
@jelbourn looks reasonable. please file an issue @ angular/angular if you haven't yet. Let's see if the core can help with this pattern as it looks better to me from the user perspective than the `forRoot` pattern.

DevVersion:
@jelbourn Were the errors on the AOT mode valid? - Also nice change. Should avoid future provider issues as in #2538

jelbourn:
@DevVersion yes, it caught some real errors

PR #2560, opened 2017-01-07
	fix(select): remove selection if option is removed

### LINKED ISSUE #2524: MdSelect selected value not cleared if option is removed

Bug

Removing an option that is currently selected should update the select

If you remove the currently selected option from the options list
the select does not change its current selected value.

See this plunker: http://plnkr.co/edit/cX2VnC8wm4yNnQTmAGxV?p=preview

* Select the "Third" option and then click on "Pop an option"
* This will remove the "Third" option
* The select will still display the "Third" value

A classic scenario would be adding an item to a collection from a select, where you only want unselected items to appear in the options list.

Angular 2.4.1
Angular Material 2.0.0-beta.1
Tested on Chrome 55 (OS/X 10.12)

This is kind of related to #2232 and the fix in #2110 but probably needs the `_setSelectionByValue`
 method to be called in response to the options collection changing.

DevVersion:
* Currenty when an option is selected and the option is somehow programmatically removed the select updates its options.
 
  Right now it does not remove the selection if the selection has been removed from the options.

Fixes #2524

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -191,12 +191,12 @@ describe('MdSelect', () => {
       firstOption.click();
       fixture.detectChanges();
 
-      expect(select.selected).toBe(select.options.first);
+      expect(select.selected).toBe(select.options.first, 'Expected first option to be selected.');
 
       fixture.componentInstance.foods = [];
       fixture.detectChanges();
 
-      expect(select.selected).toBe(null);
+      expect(select.selected).toBe(null, 'Expected nothing to be selected.');
     });
 
     it('should display the selected option in the trigger', () => {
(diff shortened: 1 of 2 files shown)


src/lib/select/select.spec.ts
```diff
@@ -173,6 +173,25 @@ describe('MdSelect', () => {
...
+
+      trigger.click();
+      fixture.detectChanges();
+
+      let firstOption = overlayContainerElement.querySelectorAll('md-option')[0] as HTMLElement;
+
+      firstOption.click();
+      fixture.detectChanges();
+
+      expect(select.selected).toBe(select.options.first);
```
kara: Can you add fallback messages for both assertions if the test fails?  e.g. `Expected first option to be selected.`

DevVersion:
@kara Alright, only kept the tests for now. Let's see if #2499 makes this PR green (once it's merged)

PR #2561, opened 2017-01-07
fix(button): only flat button and icon buttons should inherit the color

### LINKED ISSUE #2539: text color md-fab and md-mini-fab is not the default-contrast color
Bug

The text color of a md-fab or md-mini-fab should be the default-contrast of the chosen theme accent color. For example when using the prebuilt the indigo-pink the default $accent is set to A200, so according to the palette the contrast A200 is white, so the text-color of the md-fab or md-mini-fab should be white against a pink background-color.

The text-color is set to the current color (in my case grey).

Plunker template: http://plnkr.co/edit/LbeegL?p=preview
http://stackoverflow.com/questions/40478419/angular-material2-button-text-color

readability

I'm using angular-material 2.0.0-beta.1

No

EladBezalel:
- also fixes raised buttons in toolbar

fixes #2539

### DIFF
--- a/src/lib/button/_button-base.scss
+++ b/src/lib/button/_button-base.scss
@@ -41,7 +41,6 @@ $md-mini-fab-padding: 8px !default;
   font-size: $md-body-font-size-base;
   font-family: $md-font-family;
   font-weight: 500;
-  color: currentColor;
   text-align: center;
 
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

Issue #2412, opened 2016-12-26
md-select multiple support


jmiguelherazo:
#### feature request:
Is there support for multiple selection in md-select?

dahaupt:
@jmiguelherazo Not yet, look [here ](https://material.angular.io/components/component/select) for all current features. The same applies to all other components. 

But I'd also like to see multiple selection in the future.

pdavin:
+1'ing this to note we need multiple select for our project. 
AngularJS Material supported it (see "Option Groups"): https://material.angularjs.org/latest/demo/select

If Material 2 isn't going to add it soon, are there any alternatives to consider?

spyter:
@pdavin I need it for a project I am working on as well. This would be a great addition to have!

aitchkhan:
Has this been documented? Could not find this in documentation, or is it just me?

Hassan0Rehman:
I also couldn't find documentation and support. Looks like this feature is not offered yet. As a workaround we can use this http://softsimon.github.io/angular-2-dropdown-multiselect/#

crisbeto:
It has been supported for a while now. You have to add the `multiple` attribute to your `md-select`. You can see the `multiple` property in the API documentation: https://material.angular.io/components/select/api

hirrus:
Just for somebody new to angular, please use this example: https://angular-material2-issue-wak6cs.stackblitz.io

PR #2562, opened 2017-01-07
feat: add a common class to be used when dealing with selection logic

crisbeto:
Adds the `MdSelectionModel` class that can be used when dealing with single and multiple selection within a component.

Relates to #2412.


src/lib/core/selection/selection.ts
```diff
@@ -0,0 +1,168 @@
...
+  onChange: Subject<MdSelectionChange> = new Subject();
+
+  /** Currently-selected values. */
+  private _selectedValues: any[] = [];
+
+  /** Keeps track of the deselected options that haven't been emitted by the change event. */
+  private _unflushedDeselectedValues: any[] = [];
+
+  /** Keeps track of the selected option that haven't been emitted by the change event. */
+  private _unflushedSelectedValues: any[] = [];
```
jelbourn: Move all of these properties above the constructor

src/lib/core/selection/selection.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /** Event emitted when the value has changed. */
+  onChange: Subject<MdSelectionChange> = new Subject();
+
+  /** Currently-selected values. */
+  private _selectedValues: any[] = [];
+
+  /** Keeps track of the deselected options that haven't been emitted by the change event. */
+  private _unflushedDeselectedValues: any[] = [];
```
jelbourn: Not sure "unflushed" is the right term here. How about something like "pendingDeselection", "deselectedToEmit"?

src/lib/core/selection/selection.ts
```diff
@@ -0,0 +1,168 @@
...
+      // Clear the array in order to avoid firing the change event for preselected values.
+      this._unflushedSelectedValues.length = 0;
+    }
+  }
+
+  /** Event emitted when the value has changed. */
+  onChange: Subject<MdSelectionChange> = new Subject();
+
+  /** Currently-selected values. */
+  private _selectedValues: any[] = [];
```
jelbourn: Rather than keeping an array of `selectedValues`, we should keep the selection in a `Set` (named `_selection`). We can use `.keys()` when we want the list of selected values (caching the value). This will make selecting, deselecting, and checking selection much faster (and the code simpler). When the options are updated, the update would then look like.

Overall this would change insertion, deletion from O(k) to O(1) and keep getting an array of the selected values mostly the same.

src/lib/core/selection/selection.ts
```diff
@@ -0,0 +1,168 @@
+import {Subject} from 'rxjs/Subject';
+
+
+/**
+ * Class to be used to power selecting one or more options from a list.
+ * @docs-private
+ */
+export class MdSelectionModel {
+  constructor(
+    private _options: any[],
```
jelbourn: I think we can avoid keeping `_options` in this model completely if we add it as an argument to `selectAll`. That would make this class much simpler (no update logic, remove `_verifyExistence`)

crisbeto:
Re-did it based on the feedback @jelbourn.

src/lib/core/selection/selection.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      // Clear the array in order to avoid firing the change event for preselected values.
+      this._selectedToEmit.length = 0;
+    }
+  }
+
+  /**
+   * Selects a value or an array of values.
+   */
+  select(value: T): void {
```
jelbourn: I think it's a bit confusing to have `select` and `_select` methods that do different things. Maybe make the private one something like `_markSelected`? (same for deselect and clear)

crisbeto: I'd say it's probably better to move the clearing of the selected value to `_select` instead.

crisbeto: Done.

jelbourn: I still think, though, that having a `select` and `_select` method is something we should avoid. It would be very easy to call the wrong one.

crisbeto: `_markSelected` makes sense, but should we rename `_deselect` and `_clear` in that case as well?

crisbeto: Changed up the names so they're not easy to confused with the public ones.

jelbourn:
LGTM

Issue #1967, opened 2016-11-23
[Tabs] md-tab-link: routerLinkActive.isActive error


blyndcide:
Trying to use the new md-tab-nav-bar like this:
```
<nav md-tab-nav-bar>
	<a md-tab-link routerLink="/admin/management" routerLinkActive #rla="routerLinkActive" [active]="rla.isActive">Management</a>
</nav>
```

but get an error

> TypeError: Cannot read property 'some' of undefined

Seems like this way should work after commit: https://github.com/angular/angular/commit/c9f58cf78cce420762c7f8c51172a10a497468ea

DzmitryShylovich:
Pls add a plunkr with repro.

andrewseguin:
Please use the issue template which helps us reproduce and diagnose the issue. Including a plunkr with this bug will help us determine what is happening.

nayfin:
This is exactly how I am trying to use this feature as well, super easy way to keep your tabs in line with your location. I am working on a [plunkr](https://plnkr.co/edit/pLZ4Iq?p=preview) but I am running into some issues just getting it set up. 

1. It seems like the Angular Material 2 template may not be getting the latest version of material, because if I try to add the `[active]="someBoolean"` directive to the anchor, I get yelled at by the compiler:

```
    VM1000 zone.min.js:1 Unhandled Promise rejection: Template parse errors:
Can't bind to 'active' since it isn't a known property of 'a'. ("ks" color="primary">
      <a md-tab-link *ngFor="let child of directory; let i = index;"
        [ERROR ->][active]="routerLinkActive"
        (click)="activateLink(i)">
```
I am assuming this is because it is a pretty new feature. I looked through systemjs.config and it seems to be asking for the latest, so I don't know where to go from here.

2. I set up the router in main.ts, but I am stuck. I think I set up everything correctly but am getting some errors. 

If someone can help me resolve these two issues I can finish up the plunkr and we can get this feature working.

nayfin:
Thanks for the fast update. I updated the [plunkr](http://plnkr.co/edit/pLZ4Iq?p=preview).  All the relevant code is in `app.component.html`. I commented the hell out of it explaining what we are looking for. Hopefully it is just a syntactic mistake. Please let me know if I can help in any way. Thanks for all your hard work, this is an amazing set of components and I hope that I can contribute soon.

nayfin:
[...]

PR #2563, opened 2017-01-07
docs(tabs): add note for core bug that breaks `md-tab-nav-bar`

Splaktar:
add info to docs to avoid more people hitting this bug
there have been quite a few reports about this recently
it is not obvious to users that it is a core bug and not a bug in their code

Relates to #1967

### DIFF
--- a/src/lib/tabs/OVERVIEW.md
+++ b/src/lib/tabs/OVERVIEW.md
@@ -79,3 +79,7 @@ provides a tab-like UI for navigating between routes.
 The tab-nav-bar is not tied to any particular router; it works with normal `<a>` elements and uses
 the `active` property to determine which tab is currently active. The corresponding 
 `<router-outlet>` can be placed anywhere in the view. 
+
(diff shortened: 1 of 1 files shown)


Splaktar:
This specifically relates to the following https://github.com/angular/material2/issues/1967#issuecomment-271090596.

If you would like me to add the workaround to the docs (perhaps for users that won't be soon updating to the Angular 2.4.x release that eventually includes https://github.com/angular/angular/pull/13273), then please let me know.

somombo:
Thank you @splaktar

jelbourn:
I'm hesitant to add to the docs something ephemeral. I'll escalate the issue with the core team so that it should just go away soon.

Splaktar:
No problem. Just know that multiple people each day are trying out material2 and they may be running into this issue (since 2 months ago) and the component may be one of the first used in their app. 

Their initial experience in trying it out (even in beta) is likely an important aspect of them deciding to use the library or go with something else. 

To most people, this failure doesn't appear to be a bug in the core, but rather the component or their implementation of it.

PR #2565, opened 2017-01-07
fix(input): disable underline with reactive forms

### LINKED ISSUE #2558: MdInput doesn't set disabled styles appropriately when disabled in reactive forms
Example:

```html
<md-input-container>
   <input mdInput [formControl]="ctrl">
   <button (click)="ctrl.disable()">DISABLE</button>
</md-input-container>
```

When the button is clicked and the control is disabled programmatically, the disabled underline styles won't appear. This appears to be because it's only looking for the disabled property binding on the input.

Working on fixing this.

kara:
Closes #2558

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -101,8 +101,13 @@ export class MdInputDirective {
 
   /** Whether the element is disabled. */
   @Input()
-  get disabled() { return this._disabled; }
-  set disabled(value: any) { this._disabled = coerceBooleanProperty(value); }
+  get disabled() {
+    return this._ngControl ? this._ngControl.disabled : this._disabled;
+  }
+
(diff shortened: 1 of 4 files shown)


src/lib/input/input-container.ts
```diff
@@ -101,8 +101,13 @@ export class MdInputDirective {
 
   /** Whether the element is disabled. */
   @Input()
-  get disabled() { return this._disabled; }
-  set disabled(value: any) { this._disabled = coerceBooleanProperty(value); }
+  get disabled() {
```
mmalerba: does required have this same issue?

kara: It likely does, but since controls do not keep track of individual validators, it won't have the same fix.

PR #2567, opened 2017-01-08
fix(card.scss): xs screen fix padding md-card-actions

joejordanbrown:
The padding was not correct because of the media query on xs screens on the md-card padding from 24px to 16px which made md-card-actions have no padding.

Before:
image

After:
image

Material Design specs:
image

### DIFF
--- a/src/lib/card/card.scss
+++ b/src/lib/card/card.scss
@@ -147,6 +147,11 @@ md-card-title-group {
     padding: $md-card-mobile-padding;
   }
 
+  md-card-actions {
+    margin-left: -8px;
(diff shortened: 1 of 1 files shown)


jelbourn:
@joejordanbrown when I look at a card on a small screen size with Chrome dev tools I'm not seeing this issue. Can you make a plunker that demonstrates the problem (along with the screen size under which is happens)?

dahaupt:
@jelbourn I just did the plunker for this: http://plnkr.co/edit/yznyYcOsQ34SpWF8xxxe
Happens on screen sizes <=600px width.

PR #2571, opened 2017-01-08
feat(select): add floatingPlaceholder option

### LINKED ISSUE #2569: Support non-floating placeholder with select

Feature request
<md-select [floatingPlaceholder]:boolean >
    <md-option ... ></md-option>
<\/md-select>
or
<md-input-container [floatingPlaceholder]:boolean>   
 <md-select ...>
       <md-option ... ></md-option>
   <\/md-select>
 <\/md-input-container>
 
not support

Input elements  can have floating or non-floating placeholder .Select can be use with  input elements.
Select should have floating or non-floating placeholder.Similar to <md-input-container>

Material2 v2.0.0-beta
Angular 2.4.1

### LINKED ISSUE #2963: Inconsistencies in the md-select implementation compared to md-input
#### Proposal:
 md-select should have the floatPlaceholder input property and it should be able to use md-placeholder as ContentChild. 

With the current implementation you can not place the label/placeholder on the left side. With md-input you can achieve this by setting floatPlaceholder="always" and with some custom css.

You could argue that you can not use the placeholder property in order to achieve this, but if you want to keep the label colouring functionality depending on the component state  it's a pain to implement.

Maybe all form components should have a property placeholderPlacement:"top"|"left".

crisbeto:
Adds the `floatingPlaceholder` option that can be used to disable the floating placeholders.

Fixes #2569.
Fixes #2963.

### DIFF
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -18,7 +18,7 @@
 
   <md-card>
     <md-select placeholder="Drink" [(ngModel)]="currentDrink" [required]="isRequired" [disabled]="isDisabled"
-      #drinkControl="ngModel">
+      [floatPlaceholder]="floatPlaceholder" #drinkControl="ngModel">
       <md-option *ngFor="let drink of drinks" [value]="drink.value" [disabled]="drink.disabled">
         {{ drink.viewValue }}
       </md-option>
@@ -27,6 +27,15 @@
     <p> Touched: {{ drinkControl.touched }} </p>
     <p> Dirty: {{ drinkControl.dirty }} </p>
     <p> Status: {{ drinkControl.control?.status }} </p>
+    <p>
+      <label for="floating-placeholder">Floating placeholder:</label>
+      <select [(ngModel)]="floatPlaceholder" id="floating-placeholder">
+        <option value="auto">Auto</option>
+        <option value="always">Always</option>
+        <option value="never">Never</option>
+      </select>
+    </p>
+
     <button md-button (click)="currentDrink='water-2'">SET VALUE</button>
     <button md-button (click)="isRequired=!isRequired">TOGGLE REQUIRED</button>
     <button md-button (click)="isDisabled=!isDisabled">TOGGLE DISABLED</button>
--- a/src/lib/select/select.html
+++ b/src/lib/select/select.html
@@ -1,9 +1,14 @@
 <div class="mat-select-trigger" cdk-overlay-origin (click)="toggle()" #origin="cdkOverlayOrigin" #trigger>
-  <span class="mat-select-placeholder" [class.mat-floating-placeholder]="this.selected"
-   [@transformPlaceholder]="_placeholderState" [style.width.px]="_selectedValueWidth"> {{ placeholder }} </span>
+  <span
+    class="mat-select-placeholder"
+    [class.mat-floating-placeholder]="selected"
+    [@transformPlaceholder]="_getPlaceholderAnimationState()"
+    [style.visibility]="_getPlaceholderVisibility()"
+    [style.width.px]="_selectedValueWidth"> {{ placeholder }} </span>
   <span class="mat-select-value" *ngIf="selected">
     <span class="mat-select-value-text">{{ selected?.viewValue }}</span>
   </span>
+
   <span class="mat-select-arrow"></span>
   <span class="mat-select-underline"></span>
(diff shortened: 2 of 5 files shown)


kara:
@crisbeto Sorry for the late review on this. We are adding a similar attr in md-input, so we should probably make the API the same. Can you make this behavior/API match https://github.com/angular/material2/pull/2585?

crisbeto:
Done @kara.

src/demo-app/select/select-demo.html
```diff
@@ -18,7 +18,7 @@
 
   <md-card>
     <md-select placeholder="Drink" [(ngModel)]="currentDrink" [required]="isRequired" [disabled]="isDisabled"
-      #drinkControl="ngModel">
+      [floatingPlaceholder]="floatingPlaceholder" #drinkControl="ngModel">
```
kara: To match the input API, this should be `floatPlaceholder`. Can you rename?

src/lib/select/select.html
```diff
@@ -1,6 +1,11 @@
 <div class="md-select-trigger" cdk-overlay-origin (click)="toggle()" #origin="cdkOverlayOrigin" #trigger>
-  <span class="md-select-placeholder" [class.md-floating-placeholder]="this.selected"
-   [@transformPlaceholder]="_placeholderState" [style.width.px]="_selectedValueWidth"> {{ placeholder }} </span>
+  <span
+    class="md-select-placeholder"
+    [class.md-floating-placeholder]="selected"
+    [@transformPlaceholder]="placeholderState"
+    [style.visibility]="(floatingPlaceholder !== 'never' || !selected) ? 'visible' : 'hidden'"
```
kara: Can you pull this expression into the class? It's a bit long to be in the template.

src/lib/select/select.ts
```diff
@@ -124,7 +127,19 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
   private _placeholder: string;
 
   /** The animation state of the placeholder. */
-  _placeholderState = '';
+  private _placeholderState = '';
+
+  get placeholderState(): string {
```
kara: Can you return this to an internal method or property with `_`? This shouldn't be in the public API.

crisbeto:
Thanks for the feedback @kara, it's addressed.

kara:
@crisbeto Can you fix the lint error and then apply merge label?

kara:
Tests?

andrewseguin:
Can you rebase this PR?

crisbeto:
Done, although it doesn't look like it had any conflicts.

Hainesy:
Is there a reason this has been removed from `mat-select`?

Hainesy:
@crisbeto OK thanks... there's nothing in the docs about it https://material.angular.io/components/select/api

Hainesy:
@crisbeto  Doesn't work `Can't bind to 'floatLabel' since it isn't a known property of 'mat-select'.`

crisbeto:
That input is on the `mat-form-field`. See the API docs: https://material.angular.io/components/form-field/api

Hainesy:
@crisbeto I've tried it on the mat-form-field; it doesn't complain about it, but it doesn't seem to work either. I can still see the floating placeholder.

mmalerba:
@Hainesy which form-field appearance are you using? Only legacy supports `floatLabel="never"`. The option was added as a way to get the label to act more like a traditional placeholder. The standard, fill, and outline variants allow you to just use the native placeholder for this, and thus no longer support `floatLabel="never"`: https://stackblitz.com/edit/angular-fa8ngm?file=app/form-field-appearance-example.html

The label and text input are also not perfectly aligned in the new variants (as specified by the spec) so it would look strange and misaligned to have a never-float label

chriszrc:
@mmalerba has answered the question, but the simpler answer is that yes, you can use variants like fill without a floating label:

```
<mat-form-field appearance="outline">
  <mat-select disabled name="country" placeholder="Country">
      <mat-option *ngFor="let country of countries" [value]="country.value">
          {{country.label}}
       </mat-option>
   </mat-select>
</mat-form-field>        
```

However, the margins are still setup to expect a floating label, so you need to adjust the style for the .mat-form-field-infix:

```
.mat-form-field-infix {
    border-top: 0.64375em solid transparent;
    border-bottom: 0.24375em solid transparent;
 }
```

PR #2581, opened 2017-01-10
chore: expand coding standards

jelbourn:
All aboard!
@kara @mmalerba @andrewseguin @tinayuangao @crisbeto @EladBezalel @topherfangio @DevVersion @josephperrott 

The coding standards were last updated just short of one year ago. Going forward, I want to update this doc much more regularly with insights that come out of code reviews. 

Are there any conventions that we've been enforcing through code review that are not part of this doc?

### DIFF
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -12,7 +12,8 @@ ES6 or TypeScript.
 ### General
 
 #### Write useful comments
-Comments that explain what some block of code does are nice; they can tell you something in less time than it would take to follow through the code itself.
+Comments that explain what some block of code does are nice; they can tell you something in less 
+time than it would take to follow through the code itself.
 
 Comments that explain why some block of code exists at all, or does something the way it does, 
 are _invaluable_. The "why" is difficult, or sometimes impossible, to track down without seeking out 
@@ -58,6 +59,9 @@ Keeping modules to a single responsibility makes the code easier to test, consum
 ES6 modules offer a straightforward way to organize code into logical, granular units. 
 Ideally, individual files are 200 - 300 lines of code.
 
+As a rule of thumb, once a file draws near 400 lines (barring abnormally long constants / comments), 
+start considering how to refactor into smaller pieces. 
+
 #### Less is more
 Once a feature is released, it never goes away. We should avoid adding features that don't offer 
 high user value for price we pay both in maintenance, complexity, and payload size. When in doubt, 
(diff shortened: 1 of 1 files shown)


CODING_STANDARDS.md
```diff
@@ -68,9 +72,97 @@ prefer sticking to a _single_ API for accomplishing something.
...
+be part of the user-facing API. This typically applies to symbols used in template expressions, 
+`@ViewChildren` / `@ContentChildren` properties, host bindings, and `@Input` / `@Output` properties
+(when using an alias).
+
+#### JsDoc comments
+
+All public APIs must have user-facing comments. These are extracted and shown in the documation
+on [material.angular.io](https://material.angular.io).
+
+Private and internal APIs should have JsDoc when they are not obvious. Ultimately it is the purview
```
crisbeto: We should probably mention the `@docs-private` directive for excluding APIs that shouldn't be user-facing.

CODING_STANDARDS.md
```diff
@@ -68,9 +72,97 @@ prefer sticking to a _single_ API for accomplishing something.
...
+of the code reviwer as to what is "obvious", but the rule of thumb is that *most* classes, 
+properties, and methods should have a JsDoc description. 
+
+Properties should have a concise description of what the property means:
+```ts
+  /** The label position relative to the checkbox. Defaults to 'after' */
+  @Input() labelPosition: 'before' | 'after' = 'after';
```
crisbeto: Perhaps, in general, we should be turning these into types? E.g. `type MdLabelPosition = 'before' | 'after';` in order to allow users to annotate their code?

CODING_STANDARDS.md
```diff
@@ -68,9 +72,97 @@ prefer sticking to a _single_ API for accomplishing something.
 
 ### TypeScript
 
-#### Provide function descriptions
-For functions that are more complicated than a simple getter/setter, provide at least a brief 
-sentence explaining what the function does and/or _why_ it does something.
+#### Typing
```
crisbeto: We should mention that, even though TypeScript infers it, we should include the method's return type for Dgeni's sake.

CODING_STANDARDS.md
```diff
@@ -68,9 +72,97 @@ prefer sticking to a _single_ API for accomplishing something.
...
+  /** The label position relative to the checkbox. Defaults to 'after' */
+  @Input() labelPosition: 'before' | 'after' = 'after';
+``` 
+
+Methods blocks should describe what the function does and provide a description for each parameter
+and the return value:
+```ts
+  /**
+   * Opens a modal dialog containing the given component.
+   * @param component Type of the component to load into the dialog.
```
EladBezalel: We don't want to mention the parameter type?

crisbeto: The type is already in the TypeScript annotations. AFAIK Dgeni picks those up properly.

DevVersion:
What about mentioning the `100-column` limit, because the [Style Guidelines](https://google.github.io/styleguide/jsguide.html#formatting-column-limit) mention a 80-column limit, while we have 100-columns in our `tslint.json`.

CODING_STANDARDS.md
```diff
@@ -68,9 +72,97 @@ prefer sticking to a _single_ API for accomplishing something.
...
+
+#### Host bindings
+Prefer using the `host` object in the directive configuration instead of `@HostBinding` and 
+`@HostListener`. We do this because TypeScript preserves the type information of methods with 
+decorators, and when one of the arguments for the method is a native `Event` type, this preserved
+type information can lead to runtime errors in non-browser environments (e.g., server-side 
+pre-rendering). 
+
 
 ### CSS
```
mmalerba: I agree with most of this, but the "don't use flex" part has always bothered me. Do we have an example of this weird baseline behavior that we're afraid of? It's hard to tell exactly what the issue is from reading the spec...

CODING_STANDARDS.md
```diff
@@ -68,9 +72,97 @@ prefer sticking to a _single_ API for accomplishing something.
...
+class UniqueSelectionDispatcher { }
+```
+
+Avoid suffixing a class with "Service", as it communicates nothing about what the class does. Try to
+think of the class name as a person's job title.
+
+##### Methods
+The name of a method should capture the action that is performed *by* that method.
+
+### Angular
```
mmalerba: Should we mention the forRoot replacement solution?

CODING_STANDARDS.md
jelbourn: Probably, but I don't think that is general enough for the coding guidelines.

CODING_STANDARDS.md
jelbourn: I'm not sure that's general enough to belong in the coding guidelines.

CODING_STANDARDS.md
jelbourn: [One from the archives!](https://github.com/angular/material/issues/3234)

TL;DR The text baseline `display: flex` and `inline-flex` is computed from the first child of the flex container, which is different from `display: block` and `inline-block`. If you put two of these next to each other, the browser will align the baselines and the elements will be misaligned. 

Overall I think the messaging of "be cautious" with the two specific cases is good.

jelbourn:
Addressed comments

CODING_STANDARDS.md
```diff
@@ -58,19 +59,121 @@ Keeping modules to a single responsibility makes the code easier to test, consum
...
 
-#### Provide function descriptions
-For functions that are more complicated than a simple getter/setter, provide at least a brief 
-sentence explaining what the function does and/or _why_ it does something.
+#### Typing
+Avoid `any` where possible. If you find yourself using `any`, consider whether a generic may be
+appropriate in your case.
+
+For methods and properties that are part of a component's public API, all types must be explicitly
+specified because our documentation tooling cannot current infer types in places where TypeScript
```
topherfangio: "cannot current infer" -> "cannot currently infer"

CODING_STANDARDS.md
```diff
@@ -58,19 +59,121 @@ Keeping modules to a single responsibility makes the code easier to test, consum
...
+* Use `private` when appropriate and possible, prefixing the name with an underscore.
+* Prefix *library-internal* properties and methods with an underscore without using the `private` 
+keyword. This is necessary for anything that must be public (to be used by Angular), but should not
+be part of the user-facing API. This typically applies to symbols used in template expressions, 
+`@ViewChildren` / `@ContentChildren` properties, host bindings, and `@Input` / `@Output` properties
+(when using an alias).
+
+Additionally, the `@docs-private` JsDoc annotation can be used to hide any symbol from the public
+API docs.
+
```
topherfangio: What about the `protected` keyword? Do we prefer to make it public instead? A small note about best practices may be beneficial.

PR #2585, opened 2017-01-10
feat(input): option to imperatively float placeholder

### LINKED ISSUE #2466: Ability to imperatively float md-input placeholder
Currently there is a `floatingPlaceholder` property on `md-input-container` that controls whether or not you'd like a floating placeholder at all. This property effectively disables the floating placeholder.

However, it's not meant to be an imperative way to toggle the floating placeholder on and off. If you try to set `floatingPlaceholder` to true, it doesn't actual float it. It will still be gated on the two conditions that the value must not be empty or the input must be focused.

We should add a way to actually toggle the placeholder, not just disable it. The motivation for this change is the autocomplete. Sometimes the text field will be empty and unfocused (e.g. if you are keying down through the option list), but the placeholder shouldn't reset.

cc @mmalerba

DevVersion:
Refactors the `[floatingPlaceholder]` input to be able to specifiy whether the label should always float or not.

There are three options for the `floatPlaceholder` input binding now

- If set to `"always"`, the placeholder will *always* float
- If set to `"never"`, the placeholder will *never* float
- If set to `"auto"`, the placeholder will float if text is entered.

Closes #2466

R: @kara @jelbourn

### DIFF
--- a/src/lib/input/input-container.spec.ts
+++ b/src/lib/input/input-container.spec.ts
@@ -62,7 +62,7 @@ describe('MdInputContainer', function () {
 
     let inputContainer = fixture.debugElement.query(By.directive(MdInputContainer))
         .componentInstance as MdInputContainer;
-    expect(inputContainer.floatingPlaceholder).toBe('auto',
+    expect(inputContainer.floatPlaceholder).toBe('auto',
         'Expected MdInputContainer to set floatingLabel to auto by default.');
   });
 
@@ -479,7 +479,7 @@ describe('MdInputContainer', function () {
     expect(ariaValue).toBe(`${hintLabel.getAttribute('id')} ${endLabel.getAttribute('id')}`);
   });
 
-  it('should float when floatingPlaceholder is set to default and text is entered', () => {
+  it('should float when floatPlaceholder is set to default and text is entered', () => {
     let fixture = TestBed.createComponent(MdInputContainerWithDynamicPlaceholder);
     fixture.detectChanges();
 
@@ -505,7 +505,7 @@ describe('MdInputContainer', function () {
     expect(labelEl.classList).toContain('md-float');
   });
 
-  it('should always float the placeholder when floatingPlaceholder is set to true', () => {
+  it('should always float the placeholder when floatPlaceholder is set to true', () => {
     let fixture = TestBed.createComponent(MdInputContainerWithDynamicPlaceholder);
     fixture.detectChanges();
 
@@ -528,7 +528,7 @@ describe('MdInputContainer', function () {
   });
 
 
-  it('should never float the placeholder when floatingPlaceholder is set to false', () => {
+  it('should never float the placeholder when floatPlaceholder is set to false', () => {
     let fixture = TestBed.createComponent(MdInputContainerWithDynamicPlaceholder);
 
     fixture.componentInstance.shouldFloat = 'never';
@@ -741,7 +741,7 @@ class MdInputContainerWithValueBinding {
 
 @Component({
   template: `
-    <md-input-container floatingPlaceholder="never">
+    <md-input-container floatPlaceholder="never">
       <input mdInput placeholder="Label">
     </md-input-container>
   `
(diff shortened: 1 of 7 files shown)


jelbourn:
@kara does this cover what you need for autocomplete?
@mmalerba can you review the input changes?

src/lib/input/input-container.spec.ts
```diff
@@ -383,6 +384,78 @@ describe('MdInputContainer', function () {
     const textarea: HTMLTextAreaElement = fixture.nativeElement.querySelector('textarea');
     expect(textarea).not.toBeNull();
   });
+
+  it('should float when floatingPlaceholder is set to default and text is entered', () => {
+    let fixture = TestBed.createComponent(MdInputContainerWithDynamicPlaceholder);
+    fixture.detectChanges();
+
+    let inputEl = fixture.debugElement.query(By.css('input'));
```
mmalerba: I realize this might just be copied from other tests, but shouldn't they either both have .nativeElement or both not have it? Or if we really want them to be different call the second one labelNativeEl

mmalerba:
So it looks like Kara is able to do the autocomplete without this and I would rather not make this change. I plan to eventually make this type of functionality possible by allowing things other than native inputs to be placed inside md-input-container. It would then be up to that custom element to decide when it's empty which will govern whether the label is floating or not

DevVersion:
@mmalerba I understand. From issue #2466 I thought it would be necessary for the autocomplete. Let's close this PR then.

src/lib/input/input-container.ts
```diff
@@ -257,11 +266,16 @@ export class MdInputContainer implements AfterContentInit {
...
 
-  /** Text or the floating placeholder. */
+  /** Whether the placeholder should always float, never float or float as the user types. */
   @Input()
-  get floatingPlaceholder(): boolean { return this._floatingPlaceholder; }
-  set floatingPlaceholder(value) { this._floatingPlaceholder = coerceBooleanProperty(value); }
-  private _floatingPlaceholder: boolean = true;
+  get floatingPlaceholder() { return this._floatingPlaceholder; }
+  set floatingPlaceholder(value: MD_INPUT_PLACEHOLDER_TYPES) {
+    if (value && MD_INPUT_PLACEHOLDER_VALUES.indexOf(value) === -1) {
```
mmalerba: I would be fine with just not validating this and treating anything other than 'always' and 'never' as auto, up to you though

src/lib/input/input-container.ts
```diff
@@ -20,7 +20,7 @@ import {
   MdInputContainerUnsupportedTypeError,
   MdInputContainerPlaceholderConflictError,
   MdInputContainerDuplicatedHintError,
-  MdInputContainerMissingMdInputError
+  MdInputContainerMissingMdInputError, MdInputContainerFloatingPlaceholderInvalidError
```
mmalerba: nit: new line

src/lib/input/input-container.ts
```diff
@@ -257,11 +266,16 @@ export class MdInputContainer implements AfterContentInit {
   }
   private _hintLabel = '';
 
-  /** Text or the floating placeholder. */
+  /** Whether the placeholder should always float, never float or float as the user types. */
   @Input()
-  get floatingPlaceholder(): boolean { return this._floatingPlaceholder; }
-  set floatingPlaceholder(value) { this._floatingPlaceholder = coerceBooleanProperty(value); }
-  private _floatingPlaceholder: boolean = true;
+  get floatingPlaceholder() { return this._floatingPlaceholder; }
```
DevVersion: Yeah sounds more imperatively.

src/lib/input/input-container.ts
DevVersion: 👍 - Notice that invalid values for the `floatPlaceholder` attribute will automatically fallback to `auto` due to the getters [here](https://github.com/DevVersion/material2/blob/ae203c3ae49175b3630b9f69aba608b57cd7efb4/src/lib/input/input-container.ts#L254-L257)

mmalerba:
lgtm

andrewseguin:
Can you rebase?

DevVersion:
@andrewseguin Done.

kara:
@DevVersion  Can you rebase?

DevVersion:
@kara No problem. Done

andrewseguin:
Heads up that this is a breaking change for people using the [floatingPlaceholder] input. Can you add a `BREAKING CHANGE` to your PR description?

PR #2596, opened 2017-01-11
docs(dialog): explain how to disable auto-focus for elements

### LINKED ISSUE #2533: Ripple Starts at Expanded State for Slide-Toggle in a Form In A Dialog.
Bug

Ripple to only be visible when activated

Ripple starts in its expanded state when in a form inside of a dialog box.

Use md-slide-toggle in a form inside of a md-dialog. 

Plunker of the Issue http://plnkr.co/edit/prSrEY3fxm5cyV5BXeLd?p=preview

Angular 2.1.2, Material 2.0.0-beta.1, OS Windows 10, Chrome, Firefox 

slidetoggleripple

DevVersion:
* Explains that the dialog auto-focuses elements, which are tabbable for the user
* Also adds a small example for disabling the auto-focus on specific elements

Closes #2533

### DIFF
--- a/src/lib/dialog/OVERVIEW.md
+++ b/src/lib/dialog/OVERVIEW.md
@@ -40,7 +40,8 @@ Several directives are available to make it easier to structure your dialog cont
 | `md-dialog-close`     | \[Attr] Added to a `<button>`, makes the button close the dialog on click|
 
 Once a dialog opens, the dialog will automatically focus the first tabbable element.
-> To avoid that an element will be focused upon opening, developers can set the `tabindex`.
+
+You can control which elements are tab stops with the `tabindex` attribute
 
 ```html
 <button md-button tabindex="-1">Not Tabbable</button>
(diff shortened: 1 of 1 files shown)


src/lib/dialog/OVERVIEW.md
```diff
@@ -37,3 +38,10 @@ Several directives are available to make it easier to structure your dialog cont
 | `<md-dialog-content>` | Primary scrollable content of the dialog                                 |
 | `<md-dialog-actions>` | Container for action buttons at the bottom of the dialog                 |
 | `md-dialog-close`     | \[Attr] Added to a `<button>`, makes the button close the dialog on click|
+
+Once a dialog opens, the dialog will automatically focus the first tabbable element.
+> To avoid that an element will be focused upon opening, developers can set the `tabindex`.
```
jelbourn: This shouldn't be a quotation (using the `>` character). 

I would also phrase is as 
```
You can control which elements are tab stop with the `tabindex` attribute.
```

DevVersion: Updated.

PR #2600, opened 2017-01-11
chore: avoid horizontal overflow in testing app

crisbeto:
Removes some horizontal scrolling in the local development app.


jelbourn:
LGTM

PR #2602, opened 2017-01-11
Allow md-button-toggle color to be easily overwritten

barrtender:
fix(md-toggle-button): 

md-toggle-button color can be set to non-default values. Tested by adding a style="color:red" to a toggle-button in the demo app.


tinayuangao:
LGTM

PR #2606, opened 2017-01-12
feat(list-key-manager): active descendant support

kara:
This PR adds key management for lists where changing the focus is not desired.  There are now two types of key managers: FocusKeyManager (which will set focus) and ActiveDescendantKeyManager (which will simply set active styles and manage a11y through activedescendant). The latter will be useful for an autocomplete so the text field can retain focus while iterating through the list.

r: @jelbourn

### DIFF
--- a/src/lib/core/a11y/activedescendant-key-manager.ts
+++ b/src/lib/core/a11y/activedescendant-key-manager.ts
@@ -1,0 +1,35 @@
+import {QueryList} from '@angular/core';
+import {ListKeyManager, CanDisable} from './list-key-manager';
+
+/**
+ * This is the interface for highlightable items (used by the ActiveDescendantKeyManager).
+ * Each item must know how to style itself as active or inactive and whether or not it is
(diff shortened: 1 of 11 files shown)


src/lib/core/a11y/activedescendant-key-manager.ts
```diff
@@ -0,0 +1,35 @@
...
+import {ListKeyManager, CanDisable} from './list-key-manager';
+
+/**
+ * This is the interface for highlightable items (used by the ActiveDescendantKeyManager).
+ * Each item must know how to style itself as active or inactive and whether or not it is
+ * currently disabled.
+ */
+export interface Highlightable extends CanDisable {
+  setActiveStyles(): void;
+  setInactiveStyles(): void;
```
jelbourn: How about `activate()` and `deactivate()`?

kara: Discussed offline and we will leave

PR #2608, opened 2017-01-12
chore(e2e): update to latest protractor version

### LINKED ISSUE #2384: Remove @types/protractor@4.0.0 from devDependencies
Protractor now comes with internal types

When `npm install` or `yarn` this warning shows:

```
warning @types/protractor@4.0.0: This is a stub types definition for Protractor (https://github.com/angular/protractor). Protractor provides its own type definitions, so you don't need @types/protractor installed!
```

DevVersion:
* Updates to the latest Protractor version, which now uses Selenium V3

* The new types for Protractor v5 include the fix for the missing `chord` function 
  (See https://github.com/DefinitelyTyped/DefinitelyTyped/commit/123690bb2e0f9dac2371084369a8ecf8976cbb34)

* Methods like `innerHtml` have been removed in Protractor v5 (https://github.com/angular/protractor/blob/master/CHANGELOG.md)

Due to an Protractor issue I just noticed that we don't run the latest Chrome version for e2e tests. 
> (See https://github.com/angular/protractor/pull/3963)

Closes #2384


.gitignore
```diff
@@ -33,3 +33,4 @@
 npm-debug.log
 testem.log
 /.chrome
+yarn.lock
```
jelbourn: I thought that the `yarn.lock` was was _supposed_ to be checked into source control?
https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control

DevVersion: It is - (depends on the project) and then I would need to commit it to the Material 2 repository?

jelbourn: That's fine with me

DevVersion: Okay, let's handle that in another PR when we also lock some more dependencies.

DevVersion:
@jelbourn Saucelabs was using a super-old chromedriver version, which didn't work properly with Chrome v55. Specified the latest version and everything passes now.

PR #2614, opened 2017-01-12
docs: rename overview and update readme files

riavalon:


src/lib/button-toggle/README.md
```diff
@@ -1,139 +1 @@
...
-| Name | Type | Description |
-| --- | --- | --- |
-| `multiple` | boolean | Optional, default = `false`. Whether or not the group allows multiple selection. |
-
-### Events
-
-| Name | Description |
-| --- | --- |
-| `change` | Emitted when the `value` of the group changes. |
+https://material.angular.io/components/component/button-toggle
```
jelbourn: Can you prefix each link with the phrase "Please see the official documentation at ", e.g.,
```
Please see the official documentation at https://material.angular.io/components/component/button-toggle
```

I also just realized that the path has "components/component". We need to fix that so that the url is just "https://material.angular.io/components/button-toggle". I've added that to the task list.

src/lib/sidenav/README.md
```diff
@@ -1,166 +1 @@
...
-    My regular content. This will be moved into the proper DOM at runtime.
-    <button md-button (click)="start.open()">Open start sidenav</button>
-    <button md-button (click)="end.open()">Open end sidenav</button>
-
-  </md-sidenav-container>
-</app>
-```
-
-For a fullscreen sidenav, the recommended approach is set up the DOM such that the
-`md-sidenav-container` can naturally take up the full space:
```
jelbourn: I think the sidenav README contains some examples that aren't yet captured on the docs site. We should file an issue to make sure those get turned into full-fledged examples.

jelbourn:
LGTM

PR #2619, opened 2017-01-12
feat: add simplified checkbox component for usage in other components

crisbeto:
Adds the `md-pseudo-checkbox`, which is a simplified version of `md-checkbox`, that doesn't have the expensive SVG animations or the form control logic. This will be useful for multiple selection in `md-select`, as well as other components in the future.

Relates to #2412.


src/lib/core/core.ts
```diff
@@ -132,7 +136,8 @@ export {NoConflictStyleCompatibilityMode} from './compatibility/no-conflict-mode
     PortalModule,
     OverlayModule,
     A11yModule,
-    MdOptionModule
+    MdOptionModule,
+    MdSelectionModule
```
jelbourn: Trailing commas 4life

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.scss
```diff
@@ -0,0 +1,52 @@
...
+  cursor: pointer;
+  display: inline-block;
+  vertical-align: middle;
+  box-sizing: border-box;
+  position: relative;
+  transition:
+      border-color $md-checkbox-transition-duration $md-linear-out-slow-in-timing-function,
+      background-color $md-checkbox-transition-duration $md-linear-out-slow-in-timing-function;
+
+  &::after {
```
jelbourn: Add a comment explaining what the `::after` element is used for?

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.scss
```diff
@@ -0,0 +1,52 @@
...
+    border-bottom: $md-checkbox-border-width solid currentColor;
+    transition: opacity $md-checkbox-transition-duration $md-linear-out-slow-in-timing-function;
+  }
+}
+
+.md-pseudo-checkbox-disabled {
+  cursor: default;
+}
+
+.md-pseudo-checkbox-indeterminate::after {
```
jelbourn: I'm not sure that the pseudo-checkbox needs an indeterminate state; did you have something in mind for it?

Also same question for `disabled`

src/lib/core/style/_checkbox-variables.scss
```diff
@@ -0,0 +1,10 @@
+@import './variables';
```
jelbourn: File should be `_checkbox-common.scss` to be consistent with other shared styles

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+  ViewEncapsulation,
+  Input,
+  ElementRef,
+  Renderer,
+} from '@angular/core';
+
+export type MdPseudoCheckboxState = 'unchecked' | 'checked' | 'indeterminate';
+
+/**
+ * Simplified checkbox without any of the underlying form control logic
```
jelbourn: How about...
```
Component that shows a simplified checkbox without including any kind of "real" checkbox. 
Meant to be used when the checkbox is purely decorative and a large number of them will be 
included, such as for the options in a multi-select. Uses no SVGs or complex animations.

Note that this component will be completely invisible to screen-reader users. This is *not* 
interchangeable with <md-checkbox> and should not be used if the user would directly interact
 with the checkbox. The pseudo-checkbox should only be used as an implementation detail of 
more complex components that appropriately handle selected / checked state.
```

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.ts
```diff
@@ -0,0 +1,53 @@
...
+})
+export class MdPseudoCheckbox {
+  /** Display state of the checkbox. */
+  @Input() state: MdPseudoCheckboxState = 'unchecked';
+
+  /** Whether the checkbox is disabled. */
+  @Input() disabled: boolean = false;
+
+  /** Color of the checkbox. */
+  @Input()
```
jelbourn: Instead of doing an `@Input` for `color` and `disabled`, perhaps we should just include the class directly? My reasoning is that it more clearly communicates that it's not a real checkbox and is only for show. 

@kara thoughts?

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.scss
crisbeto: We definitely need the `disabled` state since select options can be disabled. Regarding the `indeterminate` state, we don't need it for the select, but I'd imagine that it would come in handy for other components (e.g. the "select all" functionality in the data table or potentially in the selection list).

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.ts
crisbeto: I guess that it's mainly for consistency. Regarding communication, this is intended for internal usage anyway, it shouldn't be an issue if people look at the readme.

crisbeto:

PR #2641, opened 2017-01-13
fix(slider): make min value style match mocks

mmalerba:
mocks: https://material.io/guidelines/components/sliders.html#sliders-continuous-slider

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -93,8 +93,8 @@ export class MdSliderChange {
     '[class.md-slider-sliding]': '_isSliding',
     '[class.md-slider-thumb-label-showing]': 'thumbLabel',
     '[class.md-slider-vertical]': 'vertical',
-    '[class.md-slider-min-value]': '_isAtMinValue',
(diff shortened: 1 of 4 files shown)


src/lib/slider/slider.ts
```diff
@@ -249,12 +256,27 @@ export class MdSlider implements ControlValueAccessor {
     return (this.direction == 'rtl' && !this.vertical) ? !this.invertAxis : this.invertAxis;
   }
 
+  /** Whether the slider is at its minimum value. */
+  get _isAtMinValue() {
```
jelbourn: I think `_isMinValue` works

src/lib/slider/slider.ts
```diff
@@ -249,12 +256,27 @@ export class MdSlider implements ControlValueAccessor {
...
    * elements.
    */
-  private get _thumbGap() {
-    return this.disabled ? DISABLED_THUMB_GAP : 0;
+  get _thumbGap() {
+    if (this.disabled) {
+      return DISABLED_THUMB_GAP;
+    }
+    if (this._isAtMinValue && !this.thumbLabel) {
+      if (this._isActive) {
```
jelbourn: Could replace this inner-most if-else with a ternary

PR #2646, opened 2017-01-14
Add directive to determine how elements were focused.

mmalerba:


src/lib/core/style/add-focus-classes.ts
```diff
@@ -0,0 +1,80 @@
+import {Directive, Injectable, Optional, SkipSelf} from '@angular/core';
+
+
+/** Singleton that allows all instances of CdkAddFocusClasses to share document event listeners. */
+@Injectable()
+export class CdkFocusCauseDetector {
```
jelbourn: How about `FocusOriginMonitor`? I would like to avoid prefixing any classes with "cdk".

src/lib/core/style/add-focus-classes.ts
```diff
@@ -0,0 +1,80 @@
...
+  /** Whether a keydown event has just occurred. */
+  get keydownOccurred() { return this._keydownOccurred; }
+  private _keydownOccurred = false;
```
jelbourn: Can you just do 
```ts
private readonly keydownOccurred = false;
```

src/lib/core/style/add-focus-classes.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  get mousedownOccurred() { return this._mousedownOccurred; }
+  private _mousedownOccurred = false;
+
+  constructor() {
+    document.addEventListener('keydown', () => {
+      this._keydownOccurred = true;
+      setTimeout(() => this._keydownOccurred = false, 0);
```
jelbourn: Instead of `setTimeout`, can you do 
```ts
Promise.resolve().then(() => this.keydownOccured = false);
```
(same elsewhere)

src/lib/core/style/add-focus-classes.ts
```diff
@@ -0,0 +1,80 @@
...
+export class CdkFocusCauseDetector {
+  /** Whether a keydown event has just occurred. */
+  get keydownOccurred() { return this._keydownOccurred; }
+  private _keydownOccurred = false;
+
+  get mousedownOccurred() { return this._mousedownOccurred; }
+  private _mousedownOccurred = false;
+
+  constructor() {
+    document.addEventListener('keydown', () => {
```
jelbourn: Add comment explaining why `useCapture` is appropriate here.

src/lib/core/style/add-focus-classes.ts
jelbourn: I'm slightly concerned about the performance implications of scheduling a asynchronous action on every keydown in the application. What can we do to conclusively say the cost for this would be minor?

src/lib/core/style/add-focus-classes.ts
```diff
@@ -0,0 +1,80 @@
...
+    }, true);
+  }
+}
+
+
+/**
+ * Directive that determines how a particular element was focused (via keyboard, mouse, or
+ * programmatically) and adds corresponding classes to the element.
+ */
+@Directive({
```
jelbourn: How would you add this directive to something like `MdButton`, where the host element itself is the focused element. It's not something the user should need to apply to the material components.

src/lib/core/style/add-focus-classes.ts
```diff
@@ -0,0 +1,80 @@
...
+  get mousedownOccurred() { return this._mousedownOccurred; }
+  private _mousedownOccurred = false;
+
+  constructor() {
+    document.addEventListener('keydown', () => {
+      this._keydownOccurred = true;
+      setTimeout(() => this._keydownOccurred = false, 0);
+    }, true);
+
+    document.addEventListener('mousedown', () => {
```
jelbourn: Will this cover touch events?

DevVersion: A lot of touch devices actually emit a `mousedown` event for touch clicks, but I don't think that this really reliable.
> Test for `mousedown` and `touchstart` events: http://codepen.io/DevVersion/pen/qRNgOZ

In the Material 1 interaction service we had to buffer the touch events: [see here](https://github.com/angular/material/blob/master/src/core/services/interaction/interaction.js#L113)

src/lib/core/style/add-focus-classes.ts
DevVersion: In Material 1 we only had a timeout for touch events (for the buffering), but for normal `mousedown` events we didn't need one.

We always keep track of the timestamp for the last interaction and then compare it based on the time difference.
> https://github.com/angular/material/blob/master/src/core/services/interaction/interaction.js#L144
> https://github.com/angular/material/blob/master/src/components/button/button.js#L169

src/lib/core/style/add-focus-classes.ts
mmalerba: This doesn't do the same thing does it? I _do_ want this to be writable inside the class, just don't want other people setting it.

src/lib/core/style/add-focus-classes.ts
mmalerba: @DevVersion Why did you guys choose 15ms, that looks kind of race-condition-y.

@jelbourn I could do something similar to the ng1 version I suppose. The scheduled function is about as lightweight as possible, I wouldn't expect it to cost us much in performance. I'll make a little experiment to verify

src/lib/core/style/add-focus-classes.ts
mmalerba: good question... maybe I can move most of the functionality into the injectable and just have the directive pass its element to the injectable?

src/lib/core/style/add-focus-classes.ts
mmalerba: Added my performance test code & results, doesn't look like it adds any noticeable latency vs the timestamp approach  (at least on my machine)

mmalerba:
PTAL, restructured the code so it can be easily used by MdButton and similar components and did a performance experiment. Still need to add touch event support, just want to make sure we're on the same page with this approach first.

src/lib/core/style/add-focus-classes.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      buttonElement = fixture.debugElement.query(By.css('button')).nativeElement;
+    });
+
+    it('should initially not be focused', () => {
+      expect(buttonElement.classList.length).toBe(0, 'button should not have focus classes');
+    });
+
+    it('should detect focus via keyboard', async(() => {
+      // Simulate focus via keyboard.
+      dispatchKeydownEvent(document, 9 /* tab */);
```
DevVersion: You should be able to use the `TAB` constant from [here](https://github.com/angular/material2/blob/master/src/lib/core/keyboard/keycodes.ts#L20)

src/lib/core/style/add-focus-classes.ts
```diff
@@ -0,0 +1,95 @@
...
+      renderer.setElementClass(element, 'cdk-mouse-focused', false);
+      renderer.setElementClass(element, 'cdk-programmatically-focused', false);
+    });
+  }
+
+  /** Focuses the element via the specified focus origin. */
+  focusVia(element: Node, renderer: Renderer, focusOrigin: FocusOrigin) {
+    this._fakeOrigin = focusOrigin;
+    document.addEventListener('focus', this._clearFakeOrigin, true);
+    renderer.invokeElementMethod(element, 'focus');
```
DevVersion: How about moving the `focus` listener (above) to its own function, so you can just invoke it with the given `fakeOrigin`.

```ts
export type FocusOriginType = 'keyboard' | ' mouse' | 'manual';

/* Somewhere in the code */
renderer.listen(element, 'focus', () => this._onElementFocus(element, origin));

/* Focus event listener */
private _onElementFocus(element: HTMLElement, target: FocusOriginType) {
  /* Handle the target */
}

/* Fake Focusing*/
private focusVia(XXX) {
  this._onElementFocus(element, focusOrigin);
}
```

This would also mean that you get rid of the `_keydownOccurred` properties and just assign the given string from `FocusOriginType` / or you use an enum.

mmalerba: I'm not sure this does the same thing. We need to actually focus the element in addition to adding the styles, so we want to invoke the focus method. We need to record the `fakeOrigin` first though, so that when we later wind up in the focus listener we know that we still want to fake the origin.

DevVersion: Good point- Although I think that you would simplify a lot of things by only keeping the `focusOrigin` as type of `FocusOriginType` instead of having booleans like `_keydownOccured`.

```ts

document.addEventListener('keydown', () => {
  this.focusOrigin = 'keyboard';
  setTimeout(() => this.focusOrigin = null, 0);
}, true);

private focusVia(element: HTMLElement, focusOrigin: FocusOrigin) {
  this.focusOrigin = focusOrigin;
  element.focus(); // Or with renderer
}

private _onElementFocus(element: HTMLElement) {
  /* Handle the this.focusOrigin */

  this.focusOrigin = null;
}
```

This way you would not need to listen for another `focus` event when faking the origin.

In this example above, `null` would be treated as programmatic.

Also this could be just an idea. Let me know what you think.

mmalerba: Ah I see, yeah I think that will be much simpler, thanks!

mmalerba:
@DevVersion So it seems like the focus & blur event listeners are not triggering in the tests on Firefox. They do work fine in the demo app though. It may have something to do with the test driver starting multiple windows, I've seen a couple mentions around the web of Firefox not bothering to fire these events if the window isn't active. I just disabled these tests for Firefox.

mmalerba:
\+ @tinayuangao since jeremy is out

tinayuangao:
LGTM

PR #2649, opened 2017-01-14
fix(icon): add caching of md-icon aria-label

### LINKED ISSUE #2642: md-icon + resize-observer-polyfill = Infinite loop

I am using resize-observer-polyfill to watch for a window resizing (to update a graph drawn with d3).  Under the hood it places a MutationObserver on the entire dom.  If I have an md-icon in my application the application freezes.

The application should not freeze.

The application freezes.

WARNING: If you are using chrome and hit run then this plunkr will eat 1 core and hang chrome.
http://plnkr.co/edit/1vgLbRbsV7D8RWvoBK8Y

I would like to be able to use both resize-observer-polyfill and md-icon.

I'd assume all.

The root cause is that md-icon sets the aria-label attribute every single time ngAfterViewChecked is run.  Setting this attribute is a DOM change and triggers the mutation observer.  The mutation observer runs and then triggers a view check.

A simple patch to prevent modifying the DOM unless the aria label actually changed fixes the issue.  I will submit a pull request soon.

westonpace:
The md-icon component now caches the aria-label value and
only actually modifies the DOM if the value is unchanged.
This prevents md-icon from modifying the DOM every iteration
of change detection.

Closed #2642

### DIFF
--- a/src/lib/icon/icon.spec.ts
+++ b/src/lib/icon/icon.spec.ts
@@ -333,6 +333,21 @@ describe('MdIcon', () => {
       expect(mdIconElement.getAttribute('aria-label')).toBe('hand');
     });
 
+    it('should not set aria label unless it actually changed', () => {
+      let fixture = TestBed.createComponent(MdIconLigatureTestApp);
+
+      const testComponent = fixture.componentInstance;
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

PR #2652, opened 2017-01-14
refactor: avoid polluting the user's rxjs setup

### LINKED ISSUE #2622: Internal usages of RxJS operators should not patch the prototype of Observable

All internal usages of Observable / RxJS operators should use the non-patch syntax for imports.

Usages of RxJS operators should use the "longhand" syntax, which avoids polluting the Observable.prototype in userland:

```ts
//desired syntax
import {Observable} from 'rxjs/Observable'
import {map} from 'rxjs/operator/map'

map.call(sourceObservable, (x) => x + 1).subscribe(...)
```

```ts
//current syntax
import {Observable} from 'rxjs/Observable'
import 'rxjs/add/operator/map'

sourceObservable.map((x) => x + 1).subscribe(...)
```

Currently the "patch" operators are used, which pollute the Observable prototype and can lead to unexpected results if, for example, Material uses the .map() operator today, and then removes it in the future. If a user uses `.map()` without explicitly importing it, their app would break upon removal by Material.

Sample of correct usage in Router : https://github.com/angular/angular/blob/52be848f94a70feffd44a1d688286f750d0471d9/modules/%40angular/router/src/router_config_loader.ts#L34

crisbeto:
Currently, the internal uses of RxJS operators are done via patch imports, which add methods directly to the global `Observable` object. This can be an issue since the user may not know where these methods are coming from and may depend on them. Afterwards, if Material removes a method import, the user's app will break since the methods won't be defined anymore. This PR:

* Replaces all of the patch imports with imports of individual operators and fixes any issues that showed up afterwards.
* Adds a `FunctionChain` class to help with chaining the individually-imported operators.

Fixes #2622.


CODING_STANDARDS.md
```diff
@@ -95,13 +95,39 @@ class ConfigBuilder {
...
+`import "rxjs/operator/map"`), as opposed to using the "patch" imports which pollute the user's
+global Observable object (e.g. `import "rxjs/add/operator/map"`):
+
+```ts
+// NO
+import 'rxjs/add/operator/map';
+someObservable.map(...).subscribe(...);
+
+// YES
+import {map} from 'rxks/operator/map';
```
arciisine: typo on rxks

crisbeto: Thanks, fixed it.

jelbourn:
This is still something we need to do, but it's going to get stuck rebasing all the time if we're not ready for it. @crisbeto let's give this another try after beta.3 and ng-conf

We should also introduce a lint check at the same time so that we don't keep adding more in the future.

willshowell:
Is this still scheduled for beta 4, or do you expect it will be bumped?

jelbourn:
We may push it out a bit- we're waiting to see if rxjs will add a more convenient syntax

crisbeto:
Closing in favor of #5314.

PR #2654, opened 2017-01-14
chore: cleanup dev-dependencies

DevVersion:
* Removes unused dev-dependencies from Angular CLI-times.
* Bumps versions for packages which have a newer version available.
* Pushes the yarn.lock file as discussed.


jelbourn:
LGTM

PR #2677, opened 2017-01-16
fix(connected-position-strategy): wrong logic when determining whether element is on screen

### LINKED ISSUE #2658: md-menu open on the bad side
Bug

Menu should open on the defined side

See plunker 

Click on button
Scroll down
Click again on button
http://plnkr.co/edit/jsQ5z1nu51iEqB0pk9kW?p=preview

Material beta 2.0
Angular 2.3.1

crisbeto:
Fixes some faulty logic, introduced in #2102, that meant that the overlay considers the viewport's scroll offset when determining how much it overflows on either side.

Fixes #2658.


jelbourn:
Don't you _want_ to account for the scroll?

crisbeto:
Nope, this is relative to the viewport, not the page itself.

jelbourn:
LGTM

PR #2678, opened 2017-01-16
fix(tooltip): provide a maximum width

### LINKED ISSUE #2671: Tooltip expands outside of viewport
Feature

Tooltip should wrap or move to another 'side'
Something like what bootstrap seems to do
https://codepen.io/anon/pen/MJjqaE

screen shot 2017-01-16 at 13 47 04

https://plnkr.co/edit/YLGWc42i4AGQXh9o4USg?p=preview
Setting to right or left shows it more dramatically

If the tooltip is on an item that's close to the edge of the viewport it's easy for the content to flow off screen

I think it's been mentioned in a comment before but I couldn't find an issue on it
https://github.com/angular/material2/pull/1830#pullrequestreview-8486273

crisbeto:
Adds a `max-width` to the tooltip, in order to wrap longer lines of text. Note that the value is arbitrary since there was nothing about it in the spec.

Fixes #2671.

### DIFF
--- a/src/lib/tooltip/tooltip.scss
+++ b/src/lib/tooltip/tooltip.scss
@@ -3,6 +3,7 @@
 
 
 $md-tooltip-target-height: 22px;
+$md-tooltip-max-width: 250px;
 $md-tooltip-font-size: 10px;
 $md-tooltip-margin: 14px;
 $md-tooltip-horizontal-padding: 8px;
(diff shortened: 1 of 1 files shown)


david-kerwick:
Thanks for this. Since it's unspecified would it be worth allowing it to be passed in?

Is it meant to detect the viewport and try not to spill past it using the position strategy?
I still get this if it's at an edge and aligned to bottom, I can now at least move it to the right though.

screen shot 2017-01-17 at 09 05 20

crisbeto:
I believe that there was another issue about making the tooltip adjust to stay in the viewport. This is mostly to avoid those long lines of text.

david-kerwick:
Cheers, I can't seem to find that issue but I'll look out for it

PR #2687, opened 2017-01-17
fix(slider): hide ticks when slider is disabled

DevVersion:
* No longer shows the ticks on hover when the slider is disabled.

### DIFF
--- a/src/demo-app/slider/slider-demo.html
+++ b/src/demo-app/slider/slider-demo.html
@@ -9,7 +9,7 @@ Label <md-slider #slidey></md-slider>
 <input [(ngModel)]="max">
 
 <h1>Disabled Slider</h1>
-<md-slider disabled [(ngModel)]="disabledValue"></md-slider>
+<md-slider disabled [(ngModel)]="disabledValue" [tickInterval]="1"></md-slider>
 <input [(ngModel)]="disabledValue">
 
 <h1>Slider with set value</h1>
--- a/src/lib/slider/slider.scss
+++ b/src/lib/slider/slider.scss
@@ -127,8 +127,9 @@ md-slider {
 }
 
 
-// Slider with ticks.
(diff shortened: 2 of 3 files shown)


src/lib/slider/_slider-theme.scss
```diff
@@ -5,7 +5,6 @@
 @mixin md-slider-theme($theme) {
   $accent: map-get($theme, accent);
 
-  // TODO(iveysaur): Find an implementation to hide the track under a disabled thumb.
```
jelbourn: This `TODO` was actually referring to something else. If you look at the spec, you'll see that there's a tiny empty space around the thumb that covers up the track when the slider is disabled; this is talking about that.

DevVersion: Isn't that already implemented? (I probably miss something)

jelbourn: Oh, I didn't realize.

jelbourn:
LGTM

mmalerba:
It looks like this causes non-disabled sliders to not render the last tick. Make sure to rebase on top of #2641 it adds some more corner cases to watch out for.

Please include me as a reviewer on slider PRs in the future.

DevVersion:
@mmalerba Yeah I locally tested this PR with your PR #2641 but at this time it wasn't merged yet.

Also from looking at the diff, there hasn't been any change for tick sequencing. It only hides the `.md-slider-ticks` class when a slider is disabled?

And yes can assign you to future slider PR's

mmalerba:
I think the css rule precedence started getting unwieldy so I added a `md-hide-last-tick` class, you could probably just add `|| disabled` to the condition for that class

DevVersion:
@mmalerba Updated everything as discussed on Slack.

PR #2688, opened 2017-01-17
fix(slide-toggle): consistent naming of aria attributes

DevVersion:
* Ensures that the slide-toggle uses consistent attribute naming.
  
> Consistent with the native inputs and other Material components.

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.ts
+++ b/src/lib/slide-toggle/slide-toggle.ts
@@ -83,7 +83,7 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   @Input() id: string = this._uniqueId;
 
   /** Used to specify the tabIndex value for the underlying input element. */
-  @Input() tabindex: number = 0;
+  @Input() tabIndex: number = 0;
 
   /** Used to set the aria-label attribute on the underlying input element. */
   @Input('aria-label') ariaLabel: string = null;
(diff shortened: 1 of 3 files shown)


src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -83,13 +83,13 @@ export class MdSlideToggle implements AfterContentInit, ControlValueAccessor {
   @Input() id: string = this._uniqueId;
 
   /** Used to specify the tabIndex value for the underlying input element. */
-  @Input() tabIndex: number = 0;
+  @Input() tabindex: number = 0;
```
jelbourn: It should be `tabIndex`, as that's the name of the property on native elements. 
(e.g., with an input you'd do `<input [tabIndex]="...">`)

DevVersion: But if we change the `@Input()` to `tabIndex` then using `tabindex` as a normal HTML attribute won't work anymore.

I think we should do the following (to cover both cases) 
```ts
@Input('tabindex') tabIndex
```

> Also we currently have `tabindex` for the deprecated input right now.

DevVersion: @jelbourn Changed it back to `tabIndex`, because I think that the consistency with the `HTMLElement` property is more weighted than the `tabindex` attribute.

> Didn't change it for the deprecated input component, because making some breaking changes on something that's deprecated feels wrong.

jelbourn:
LGTM

PR #2693, opened 2017-01-17
fix(e2e): Use current directory so e2e test won't complain about directory not exists

tinayuangao:
R: @jelbourn

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /dist
 /tmp
 /deploy
+/screenshots
 
 # dependencies
 /node_modules
--- a/e2e/screenshot.ts
+++ b/e2e/screenshot.ts
@@ -47,11 +47,14 @@ export class Screenshot {
     }
 
     if (fs.existsSync(OUTPUT_DIR)) {
+      console.log(`take screenshot ${this.id}`);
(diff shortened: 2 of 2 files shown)


DevVersion:
Should we add this to the `.gitignore`?

jelbourn:
Yeah, the directory should be marked as ignored.

crisbeto:
Also shouldn't we disable screenshots when developing locally? I feel like this would stack up a lot.

jelbourn:
@tinayuangao that's another good point- @DevVersion @crisbeto is there an easy way for the screenshot utility to know if it's local? I'd assume there's an environment variable we can check.

tinayuangao:
Use process.env['TRAVIS'] to check whether it's local. Please take another look. Thanks!

DevVersion:
Just one question, how can we take a look at those screenshots on the CI?

tinayuangao:
> Just one question, how can we take a look at those screenshots on the CI?

Currently there's no way to take a look. We will upload the images, compare them and generate a report later.

DevVersion:
@tinayuangao Cool. I have some nice ideas about the viewing those reports since we also have data about the payload size etc.

jelbourn:
LGTM

PR #2699, opened 2017-01-17
fix(select): fix select panel animation

### LINKED ISSUE #2695: Select animation broken at master
The panel does not fade in properly (it jumps). Looking into it.

kara:
The 'md-select-panel-done-animating' class was being added when the panel was done transforming in, not when it was done fading in. As a result, the content was flashing before fading in.

Fixes #2695.


jelbourn:
LGTM

PR #2701, opened 2017-01-17
fix(checkbox): Add value attribute to md-checkbox

### LINKED ISSUE #2583: Missing value-field in md-checkbox
Dove deep into trying to find a reason to exclude the value field. Is there any plans on a native option to set the value of a md-checkbox or are we bound to Booleans? From experience of all the 100s of checkbox fields I have encountered they are rarely used for straight true/false value extraction. 

It seems most solutions around this are that you will have to translate the input from the change attribute.

tinayuangao:
Fixes #2583

R: @jelbourn

### DIFF
--- a/src/lib/checkbox/checkbox.html
+++ b/src/lib/checkbox/checkbox.html
@@ -5,6 +5,7 @@
            [id]="inputId"
            [required]="required"
            [checked]="checked"
+           [attr.value]="value"
            [disabled]="disabled"
            [name]="name"
            [tabIndex]="tabIndex"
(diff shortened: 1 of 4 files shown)


src/lib/checkbox/checkbox.html
```diff
@@ -5,6 +5,7 @@
            [id]="inputId"
            [required]="required"
            [checked]="checked"
+           [attr.value]="value"
```
jelbourn: Why `attr.` instead of binding the value property directly?

tinayuangao: Changed to bind the value property directly

jelbourn:
LGTM, just needs rebase

PR #2703, opened 2017-01-17
feat(tooltip): reposition on scroll

andrewseguin:
Reposition uses a throttled scroll event from the scroll dispatcher. If the reposition causes the tooltip to be clipped by a scrollable container, then hide the tooltip.


andrewseguin:
Looking into why IE/FF is having issues...

src/lib/tooltip/tooltip.ts
```diff
@@ -102,11 +106,22 @@ export class MdTooltip implements OnDestroy {
...
+              private _scrollDispatcher: ScrollDispatcher,
               private _elementRef: ElementRef,
               private _viewContainerRef: ViewContainerRef,
               private _ngZone: NgZone,
               @Optional() private _dir: Dir) { }
 
+  ngOnInit() {
+    // When a scroll on the page occurs, update the position in case this tooltip needs		
+    // to be repositioned.
+    this._scrollDispatcher.scrolled(SCROLL_THROTTLE_MS).subscribe(() => {
```
jelbourn: Need to unsubscribe on `ngOnDestroy`

PR #2704, opened 2017-01-17
fix(select): support use inside a custom value accessor

### LINKED ISSUE #2609: Using md-select in components template breaks banana-binding
#### Summary
When using `md-select` in the template of a component to be used with `[(ngModule)]` (implementing `ControlValueAccessor`), said component binding does not function (internally, angular does not invoke `ControlValueAccessor `methods)

http://plnkr.co/edit/BzgXV2?p=preview
This plunker contains two identical components where the only difference is the inclusion of `md-select` in the template of `with-select` component. This breaks banana binding as described.

I have nested other banana-bound components into my components with no issue.

See also https://github.com/angular/angular/issues/13886

kara:
Fixes #2609.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -6,7 +6,9 @@ import {OverlayContainer} from '../core/overlay/overlay-container';
 import {MdSelect} from './select';
 import {MdOption} from '../core/option/option';
 import {Dir} from '../core/rtl/dir';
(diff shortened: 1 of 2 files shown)


jelbourn:
LGTM

Do other components need to do anything like this as well?

kara:
@jelbourn I believe this is the only component right now that sets the value accessor directly.  mdInput uses a normal value accessor and the others use the provider syntax (which will work correctly).

PR #2708, opened 2017-01-18
chore: exclude node types for aot demo-app

DevVersion:
* No longer uses the ambigious Node typings for the AOT compilation (As fixed in https://github.com/angular/material2/commit/bdc48a83713f34c2780aace265a03fd41d64d6d0)
* Renames the `typings.d.ts` into `demo-app-types.d.ts` to make clear what file is used and also to keep consistent with `@types`.
* Adds a missing component variable to the `radio` e2e component (necessary for future AOT mode in the e2e-app)


jelbourn:
LGTM

PR #2710, opened 2017-01-18
fix(autocomplete): scroll options below fold into view

kara:
Because we are not focusing active options, we need to scroll options below the fold into view manually.  Depends on many PRs that are waiting to be merged.

r: @jelbourn


kara:
Closing in favor of https://github.com/angular/material2/pull/2727

PR #2720, opened 2017-01-19
feat(scripts): push generated docs to material assets repo

riavalon:
Setup scripts for pushing generated docs to material2-docs-content repo on release

### DIFF
--- a/scripts/ci/after-success.sh
+++ b/scripts/ci/after-success.sh
@@ -10,4 +10,5 @@ cd $(dirname $0)/../..
 if [ "$TRAVIS_PULL_REQUEST" = "false" ] && $(npm bin)/travis-after-modes; then
   echo "All travis modes passed. Publishing the build artifacts..."
   ./scripts/release/publish-build-artifacts.sh
+  ./scripts/release/publish-docs-content.sh
 fi
--- a/scripts/release/publish-docs-content.sh
+++ b/scripts/release/publish-docs-content.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+# Publish material2 docs assets to the material2-docs-content repo
+# material.angular.io will pull from this assets repo to get the latest docs
+
+cd "$(dirname $0)/../../"
+
+docsPath="/dist/docs"
+repoPath="/tmp/material2-docs-content"
+repoUrl="https://github.com/DevVersion/material2-docs-content"
+
+# If the docs directory is not present, generate docs
+if [ ! -d $docsPath ]; then
+  $(npm bin)/gulp docs
+fi
+
+# Get git meta info for commit
+commitSha="$(git rev-parse --short HEAD)"
+commitAuthorName="$(git --no-pager show -s --format='%an' HEAD)"
+commitAuthorEmail="$(git --no-pager show -s --format='%ae' HEAD)"
+commitMessage="$(git log --oneline -n 1)"
+
+# create directory and clone test repo
+rm -rf "/tmp/"
+mkdir "/tmp/" $repoPath
+git clone $repoUrl $repoPath
+
+# Clean out repo directory and copy contents of dist/docs into it
+rm -rf $repoPath/*
+mkdir $repoPath/overview
+mkdir $repoPath/guides
+mkdir $repoPath/api
+# mkdir $repoPath/examples
+
+# Move api files over to $repoPath/api
+cp -r $docsPath/api/* $repoPath/api
+
+# Move guide files over to $repoPath/guides
+for filename in $overviewFiles*
+do
+  if [ -f $filename ]; then
+    cp -r $filename $repoPath/guides
+  fi
+done
+
+# Flatten the markdown docs structure and move it into $repoPath/overview
+overviewFiles=$docsPath/markdown/
+targetFile="OVERVIEW.html"
+for filename in $overviewFiles*
+do
+  if [ -d $filename ]; then
+    for _ in $filename/*
+    do
+      if [ -f $filename/$targetFile ]; then
+        name=${filename#$overviewFiles}
+        cp -r $filename/$targetFile $repoPath/overview/
+        mv $repoPath/overview/$targetFile $repoPath/overview/$name.html
+      fi
+    done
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -33,6 +33,7 @@ env:
     - MODE=aot
     - MODE=payload
     - MODE=e2e
+    - MODE=docs
```
DevVersion: The whole mode shouldn't be necessary when we run it with the `travis-after-modes` script.

scripts/release/publish-docs-artifacts.sh
```diff
@@ -0,0 +1,50 @@
...
+root="$(pwd)"
+
+docsPath="./dist/docs"
+tempRepoPath="./dist/tempRepo"
+repoUrl="https://github.com/angular/material2-docs-content/"
+
+# If the docs directory is not present, abort
+if [ ! -d $docsPath ]; then
+  echo "Material docs not generated. Aborting..."
+  exit 0
```
DevVersion: Why not returning an exit code `1` here? If Travis turns red we can see that there are issues with the docs.

scripts/release/publish-docs-artifacts.sh
```diff
@@ -0,0 +1,50 @@
...
+$(find $tempRepoPath -d 1 | grep -v "git" | xargs rm -rf)
+mkdir $tempRepoPath/docs
+cp -r $docsPath/* $tempRepoPath/docs
+
+# Push content to repo
+cd $tempRepoPath
+# git remote show origin
+git config user.name "$commitAuthorName"
+git config user.email "$commitAuthorEmail"
+git config credential.helper "store --file=.git/credentials"
```
DevVersion: Since the Github credentials are now retrieved from the `.git/credentials` file, the credentials file needs to be created.

```bash
echo "https://${MATERIAL2_BUILDS_TOKEN}:@github.com" > .git/credentials
```

scripts/release/publish-docs-artifacts.sh
```diff
@@ -0,0 +1,50 @@
...
+
+git add -A
+git commit -m "$commitMessage"
+git tag "$commitSha"
+git push origin master --tags
+
+echo "returning to root and removing temp repo"
+cd $root
+rm -rf $tempRepoPath
+echo "Done!"
```
DevVersion: I don't think that returning to the project directory is necessary here. 

* Those scripts run most of the time in a child process and don't affect the users PWD.

scripts/release/publish-docs-artifacts.sh
```diff
@@ -0,0 +1,50 @@
...
+commitAuthorEmail="$(git --no-pager show -s --format='%ae' HEAD)"
+commitMessage="$(git log --oneline -n 1)"
+
+# create directory and clone test repo
+rm -rf $tempRepoPath
+mkdir $tempRepoPath
+git clone $repoUrl $tempRepoPath
+
+# Clean out repo directory and copy contents of dist/docs into it
+$(find $tempRepoPath -d 1 | grep -v "git" | xargs rm -rf)
```
DevVersion: How about just calling `rm -rf $tempRepoPath /*`?
> The `.git` folder won't be affected because the wildcard doesn't cover `.*` files / folders.

scripts/release/publish-docs-artifacts.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# Publish material2 docs assets to the material2-docs-content repo
+# material.angular.io will pull from this assets repo to get the latest docs
+
+cd "$(dirname $0)/../../"
+
+docsPath="./dist/docs"
+tempRepoPath="./dist/tempRepo"
+repoUrl="https://github.com/angular/material2-docs-content"
```
jelbourn: It would be good to let this do a few test runs before committing to the real repo. We did this for the material2-builds; you can do the same by:
* Fork angular/material2-docs-content
* Add an access token for said fork to our Travis CI environment
* Update this script to use that token instead of `MATERIAL2_BUILDS_TOKEN`

We can then just let it run for a few commits and make sure it looks good before updating the repo / access token.

scripts/release/publish-docs-artifacts.sh
```diff
@@ -0,0 +1,44 @@
...
+commitMessage="$(git log --oneline -n 1)"
+
+# create directory and clone test repo
+rm -rf $tempRepoPath
+mkdir $tempRepoPath
+git clone $repoUrl $tempRepoPath
+
+# Clean out repo directory and copy contents of dist/docs into it
+rm -rf $tempRepoPath/*
+mkdir $tempRepoPath/docs
```
jelbourn: Realized I should add a bit more here; we should:
* Change the current `gulp docs` task to be called `gulp markdown-docs`
* Change the current `gulp api` to `gulp api-docs`
* Add a new task called `gulp docs` that just runs the previous two
* Change `gulp api-docs` to output to `dist/docs/api` and `markdown-docs` to output to `dist/docs/markdown`

This should make it easier for this script to massage the different types of docs into the right place (and still only need to run one command).

scripts/release/publish-docs-artifacts.sh
DevVersion: I don't think everybody can create Travis encrypted variables. I can create a small repo for it and we don't need to change the token at all. (Since its currently  running over my token)

jelbourn: @DevVersion  That's probably easier

DevVersion: Okay so here is the repository: `https://github.com/DevVersion/material2-docs-content.git`

@jelbourn In the future we can probably switch to another token.

scripts/release/publish-docs-artifacts.sh
DevVersion: Also while we are talking about naming. Can we rename the file to `publish-docs-content.sh`?

> Artifacts sounds like we are building a release of the components

kara:
@riavalon Can you rebase?

PR #2721, opened 2017-01-19
fix(tooltip): not working properly with ChangeDetectionStrategy.OnPush

### LINKED ISSUE #2713: mdTooltip is not working properly with ChangeDetectionStrategy.OnPush
#### Bug:

Inside a component with `changeDetection: ChangeDetectionStrategy.OnPush` the mdTooltip is not working properly.

When the mouse is over the element, the tooltip is showed, and when the mouse leave the element, the tooltip is hidden.

Show the tooltip sometimes when the mouse leave, sometimes never, sometimes when the mouse is over... Is not always the same behavior.

Browser: Chrome 55.0.2883.95 (64-bit)

```
angular-cli: 1.0.0-beta.24
node: 6.7.0
os: darwin x64
@angular/common: 2.4.1
@angular/compiler: 2.4.1
@angular/core: 2.4.1
@angular/forms: 2.4.1
@angular/http: 2.4.1
@angular/platform-browser: 2.4.1
@angular/platform-browser-dynamic: 2.4.1
@angular/router: 3.4.1
@angular/compiler-cli: 2.4.1
```

thomaspink:
fixes #2713

### DIFF
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -69,7 +69,7 @@ describe('MdTooltip', () => {
       tick(500);
 
       // Make sure tooltip is shown to the user and animation has finished
-      const tooltipElement = <HTMLElement>overlayContainerElement.querySelector('.md-tooltip');
+      const tooltipElement = overlayContainerElement.querySelector('.md-tooltip') as HTMLElement;
       expect(tooltipElement instanceof HTMLElement).toBe(true);
       expect(tooltipElement.style.transform).toBe('scale(1)');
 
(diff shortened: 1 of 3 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -309,6 +310,10 @@ export class TooltipComponent {
       // If this was set to true immediately, then a body click that triggers show() would
       // trigger interaction and close the tooltip right after it was displayed.
       this._closeOnInteraction = false;
+
+      // Call change detection manually so tooltip will also work 
+      // if any parent component has set the ChangeDetectionStrategy to OnPush
+      this._ref.detectChanges();
```
DevVersion: I think rather than always triggering a `ChangeDetection` we should just mark for a new check.

We are doing this in other components as well. 
> https://angular.io/docs/js/latest/api/core/index/ChangeDetectorRef-class.html

thomaspink: @jelbourn is there a way to find out if a parent component has set the `ChangeDetectionStrategy` to `OnPush` or not so we can skip this line?

src/lib/tooltip/tooltip.ts
```diff
@@ -286,7 +287,7 @@ export class TooltipComponent {
   /** Subject for notifying that the tooltip has been hidden from the view */
   private _onHide: Subject<any> = new Subject();
 
-  constructor(@Optional() private _dir: Dir) {}
+  constructor(@Optional() private _dir: Dir, private _ref: ChangeDetectorRef) {}
```
DevVersion: `_ref` is not very precise. It should be `_changeDetectorRef`

thomaspink:
@DevVersion getting error in e2e test - maybe restart?
```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

DevVersion:
@thomaspink Done. Never saw such an error on the CI...

thomaspink:
@DevVersion seemed to work now - strange

Thanks

thomaspink:
@DevVersion trying to figure out whats happening here #2733 so i can submit test cases for this PR

thomaspink:
Added tests

src/lib/tooltip/tooltip.spec.ts
```diff
@@ -76,6 +90,43 @@ describe('MdTooltip', () => {
       expect(tooltipDirective._tooltipInstance).toBeNull();
     }));
 
+    it('should show and hide the tooltip if changeDetection is OnPush', fakeAsync(() => {
```
jelbourn: Can you move this test to a separate describe block called "with OnPush" as a sibling to the existing "basic usage" block? Then you can do the `createComponent` in the `beforeEach`.

src/lib/tooltip/tooltip.spec.ts
```diff
@@ -76,6 +90,43 @@ describe('MdTooltip', () => {
...
+      tick(0); // Tick for the show delay (default is 0)
+      expect(tooltipDirectiveOnPush._isTooltipVisible()).toBe(true);
+
+      fixtureOnPush.detectChanges();
+
+      // wait till animation has finished
+      tick(500);
+
+      // Make sure tooltip is shown to the user and animation has finished
+      const tooltipElement = <HTMLElement>overlayContainerElement.querySelector('.md-tooltip');
```
jelbourn: Prefer the `as HTMLElement` syntax

thomaspink:
@jelbourn implemented as suggested.

PR #2722, opened 2017-01-19
feat(select): add multiple selection mode

### LINKED ISSUE #2412: md-select multiple support

crisbeto:
* Integrates the `SelectionModel` into `md-select`.
* Adds the `multiple` option which allows users to select multiple options from a `md-select`.
* Fixes a button that wasn't being cleaned up from dialog tests, causing some select tests to fail.

Fixes #2412.

### DIFF
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -1,23 +1,10 @@
 <div style="height: 1000px">This div is for testing scrolled selects.</div>
 <button md-button (click)="showSelect=!showSelect">SHOW SELECT</button>
 <div class="demo-select">
-  <div *ngIf="showSelect">
-    <md-card>
-      <md-select placeholder="Food i would like to eat" [formControl]="foodControl">
-        <md-option *ngFor="let food of foods" [value]="food.value"> {{ food.viewValue }} </md-option>
-      </md-select>
-      <p> Value: {{ foodControl.value }} </p>
-      <p> Touched: {{ foodControl.touched }} </p>
-      <p> Dirty: {{ foodControl.dirty }} </p>
-      <p> Status: {{ foodControl.status }} </p>
-      <button md-button (click)="foodControl.setValue('pizza-1')">SET VALUE</button>
-      <button md-button (click)="toggleDisabled()">TOGGLE DISABLED</button>
-      <button md-button (click)="foodControl.reset()">RESET</button>
-    </md-card>
-  </div>
-
   <md-card>
-    <md-select placeholder="Drink" [(ngModel)]="currentDrink" [required]="isRequired" [disabled]="isDisabled"
+    <md-card-subtitle>ngModel</md-card-subtitle>
+
+    <md-select placeholder="Drink" [(ngModel)]="currentDrink" [required]="drinksRequired" [disabled]="drinksDisabled"
       [floatPlaceholder]="floatPlaceholder" #drinkControl="ngModel">
       <md-option *ngFor="let drink of drinks" [value]="drink.value" [disabled]="drink.disabled">
         {{ drink.viewValue }}
@@ -37,18 +24,62 @@
     </p>
 
     <button md-button (click)="currentDrink='water-2'">SET VALUE</button>
-    <button md-button (click)="isRequired=!isRequired">TOGGLE REQUIRED</button>
-    <button md-button (click)="isDisabled=!isDisabled">TOGGLE DISABLED</button>
+    <button md-button (click)="drinksRequired=!drinksRequired">TOGGLE REQUIRED</button>
+    <button md-button (click)="drinksDisabled=!drinksDisabled">TOGGLE DISABLED</button>
     <button md-button (click)="drinkControl.reset()">RESET</button>
+  </md-card>
+
+  <md-card>
+    <md-card-subtitle>Multiple selection</md-card-subtitle>
+
+    <md-card-content>
+      <md-select multiple placeholder="Pokemon" [(ngModel)]="currentPokemon"
+        [required]="pokemonRequired" [disabled]="pokemonDisabled" #pokemonControl="ngModel">
+        <md-option *ngFor="let creature of pokemon" [value]="creature.value">
+          {{ creature.viewValue }}
+        </md-option>
+      </md-select>
+      <p> Value: {{ currentPokemon }} </p>
+      <p> Touched: {{ pokemonControl.touched }} </p>
+      <p> Dirty: {{ pokemonControl.dirty }} </p>
+      <p> Status: {{ pokemonControl.control?.status }} </p>
+      <button md-button (click)="currentPokemon=['eevee-4', 'psyduck-6']">SET VALUE</button>
+      <button md-button (click)="pokemonRequired=!pokemonRequired">TOGGLE REQUIRED</button>
+      <button md-button (click)="pokemonDisabled=!pokemonDisabled">TOGGLE DISABLED</button>
+      <button md-button (click)="pokemonControl.reset()">RESET</button>
+    </md-card-content>
   </md-card>
 
   <div *ngIf="showSelect">
     <md-card>
-      <md-select placeholder="Starter Pokemon" (change)="latestChangeEvent = $event">
-        <md-option *ngFor="let starter of pokemon" [value]="starter.value"> {{ starter.viewValue }} </md-option>
-      </md-select>
+      <md-card-subtitle>formControl</md-card-subtitle>
 
-      <p> Change event value: {{ latestChangeEvent?.value }} </p>
+      <md-card-content>
+        <md-select placeholder="Food i would like to eat" [formControl]="foodControl">
+          <md-option *ngFor="let food of foods" [value]="food.value"> {{ food.viewValue }}</md-option>
+        </md-select>
+        <p> Value: {{ foodControl.value }} </p>
+        <p> Touched: {{ foodControl.touched }} </p>
+        <p> Dirty: {{ foodControl.dirty }} </p>
+        <p> Status: {{ foodControl.status }} </p>
+        <button md-button (click)="foodControl.setValue('pizza-1')">SET VALUE</button>
+        <button md-button (click)="toggleDisabled()">TOGGLE DISABLED</button>
+        <button md-button (click)="foodControl.reset()">RESET</button>
+      </md-card-content>
+    </md-card>
+  </div>
+
+  <div *ngIf="showSelect">
+    <md-card>
+      <md-card-subtitle>Change event</md-card-subtitle>
+
+      <md-card-content>
+        <md-select placeholder="Starter Pokemon" (change)="latestChangeEvent = $event">
+          <md-option *ngFor="let creature of pokemon" [value]="creature.value">{{ creature.viewValue }}</md-option>
+        </md-select>
+
+        <p> Change event value: {{ latestChangeEvent?.value }} </p>
+      </md-card-content>
     </md-card>
   </div>
 
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -43,7 +43,7 @@
       <p> Touched: {{ pokemonControl.touched }} </p>
       <p> Dirty: {{ pokemonControl.dirty }} </p>
       <p> Status: {{ pokemonControl.control?.status }} </p>
-      <button md-button (click)="currentPokemon=['eevee-4', 'psyduck-6']">SET VALUE</button>
+      <button md-button (click)="setPokemonValue()">SET VALUE</button>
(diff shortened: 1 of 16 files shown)


crisbeto:
@kara @jelbourn this PR is blocked by #2619 and #2444, however all the logic is there. Once #2619 is merged, there'll be a little bit more to do in regards to adding the checkbox and tweaking the alignment logic, but this should be enough to start reviewing.

crisbeto:
@kara this is ready for review.

fxck:
@crisbeto I'm not sure if material guidelines say something about this, but what would be handy for my use case is to be able to control what shows inside `.md-select-value`.

Problems I have with this:

image

- it doesn't indicate how many options are actually selected, here you at least see there is more than one, but imagine you had a pokemon with some long ass name
- I would actually like for the select to expand depending on the content rather than to have fixed width(https://github.com/angular/material2/issues/2782), which with the way this currently works would result in a really really long input

So what would be nice would be to be able to to switch to a mode, that would show something like

> Charizard, Pikachu and 2 more selected..

with the amount of actually shown elements would be controlled by max-width of the `md-select` element, so if I had a layout where I can afford to have this select full width, I could show all the selected options, if I had just 100px and I couldn't even fit in Charizard, it would just say 4 selected or something..

fxck:
And it would also be nice to be able to clear all options at once, but I guess that could be done with some `select-header` https://github.com/angular/material2/issues/2812

crisbeto:
I think that those make sense, but I'd rather keep them to a separate PR since this one is already pretty huge.

paullryan:
Probably belongs in another issue but it would be great if arrow keys could be used to activate the select (once focused) like a standard select from an accessibility standpoint.

Behavior example https://plnkr.co/edit/MoygfDy6qyEyF96Sa0v4?p=preview

src/lib/select/select.ts
```diff
@@ -441,65 +513,77 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
 
-  /** Listens to selection events on each option. */
+  /** Listens to user-generated selection events on each option. */
   private _listenToOptions(): void {
-    this.options.forEach((option: MdOption) => {
-      const sub = option.onSelect.subscribe((event: MdOptionSelectEvent) => {
-        if (event.isUserInput && this._selected !== option) {
-          this._emitChangeEvent(option);
+    let source = Observable.merge.apply(Observable,
+        this.options.map(option => option.onSelectionChange));
```
kara: Can you break this expression out into a getter for `optionSelectionChanges` that returns `Observable<MdOptionSelectionChange>` type?

src/lib/select/select.ts
```diff
@@ -441,65 +513,77 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
+      this._optionSubscription.unsubscribe();
+      this._optionSubscription = null;
+    }
   }
 
-  /** Emits an event when the user selects an option. */
-  private _emitChangeEvent(option: MdOption): void {
-    this._onChange(option.value);
-    this.change.emit(new MdSelectChange(this, option.value));
+  /** Sets the changes to the model value and emits a change event. */
```
kara: I think it might be clearer to just say `Emits change event to set the model value.`

src/lib/select/select.ts
```diff
@@ -441,65 +513,77 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
-    this.options.forEach((option: MdOption) => {
-      const sub = option.onSelect.subscribe((event: MdOptionSelectEvent) => {
-        if (event.isUserInput && this._selected !== option) {
-          this._emitChangeEvent(option);
+    let source = Observable.merge.apply(Observable,
+        this.options.map(option => option.onSelectionChange));
+
+    this._optionSubscription = source
+      .filter((event: MdOptionSelectionChange) => event.isUserInput)
+      .subscribe((event: MdOptionSelectionChange) => {
```
kara: If you break out into the getter, you can remove the types of these two `event` params as implied.

src/lib/select/select.ts
```diff
@@ -441,65 +513,77 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
+
+        if (wasSelected !== this._model.isSelected(event.source)) {
+          this._propagateChanges();
+        }
+
+        this._setValueWidth();
+
+        if (!this.multiple) {
+          this.close();
         }
```
kara: This method is getting pretty fat since the `_onSelect()` method was removed. Can you re-add it to move some of this logic out of the main function (lines 535-552)?

kara:
@paullryan Feature requests are generally best in Github issues, rather than PRs where they can get buried. I assume you mean that hitting the DOWN arrow while focused opens the select? We have arrow key support while the select is open but not to open it. Feel free to open a separate issue.

src/lib/select/select.spec.ts
```diff
@@ -1434,6 +1629,33 @@ class CompWithCustomSelect {
...
+    { value: 'pizza-1', viewValue: 'Pizza' },
+    { value: 'tacos-2', viewValue: 'Tacos' },
+    { value: 'sandwich-3', viewValue: 'Sandwich' },
+    { value: 'chips-4', viewValue: 'Chips' },
+    { value: 'eggs-5', viewValue: 'Eggs' },
+    { value: 'pasta-6', viewValue: 'Pasta' },
+    { value: 'sushi-7', viewValue: 'Sushi' },
+  ];
+  control = new FormControl();
+  isRequired: boolean;
```
crisbeto: I was testing it initially, but the older versions of Angular forms didn't seem to handle arrays being required properly. I'll remove it for now.

paullryan:
@kara I understand I was more thinking out loud here, I'll open an issue. 

Side Question: why the push to go ternary in the code above (src/lib/core/selection/selection.ts:66). Given modern compilers ability optimize the more verbose is less error prone in the long run.

kara:
@paulryan Readability

paullryan:
@kara Added #2990 thanks.

crisbeto:
I've addressed the feedback @kara. One note: when setting the value via `formControl.setValue`, I opted not to sort the model value itself (the label is still sorted), in order to avoid mutating the user's data.

src/lib/select/select.ts
```diff
@@ -456,68 +531,94 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
+    }
+  }
+
+  /**
+   * Sorts the model values, ensuring that they keep the same
+   * order that they have in the panel.
+   */
+  private _sortValues(): void {
+    const options = this.options.toArray();
+    this._selectionModel.sort((a, b) => options.indexOf(a) - options.indexOf(b));
```
crisbeto: The issue with clearing the selection and re-selecting all of the options is that they'll end up firing a bunch of change events in the process, which may end up causing more overhead than sorting the list model.

src/demo-app/select/select-demo.html
```diff
@@ -1,23 +1,50 @@
...
 <div class="demo-select">
-  <div *ngIf="showSelect">
-    <md-card>
-      <md-select placeholder="Food i would like to eat" [formControl]="foodControl">
-        <md-option *ngFor="let food of foods" [value]="food.value"> {{ food.viewValue }} </md-option>
+  <md-card>
+    <md-card-subtitle>ngModel</md-card-subtitle>
+
+    <md-card-content>
+      <md-select placeholder="Drink" [(ngModel)]="currentDrink" [required]="drinksRequired"
```
crisbeto: Looks like a rebase error.

src/lib/select/select.ts
kara: We should be able to avoid firing unnecessary change events since we control when they're fired. Seems like it just needs to be re-worked.

crisbeto: It could be, but doesn't having something like `select(value: T, emitChangeEvent = true)` look weird?

kara: It seems like we need a separate bulk selection method that doesn't emit a change event.

src/lib/select/select.ts
```diff
@@ -456,68 +531,94 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
+    const wasSelected = this._selectionModel.isSelected(option);
+
+    if (this.multiple) {
+      this._selectionModel.toggle(option);
+    } else {
+      this._clearSelection(option);
+      this._selectionModel.select(option);
+    }
+
+    this._sortValues();
```
kara: This also needs a check for multiple, since you don't need to sort values on a regular select.

src/lib/select/select.ts
crisbeto: That may complicate the selection model even more than adding the `emitChangeEvent` to `select`, `deselect` and `clear`. In general, I doubt that we'd run into performance issues with this in particular, because:

1. Even if we had hundreds of items, sorting through them would still take less than a millisecond.
2. The sorting only happens on click, it's not something that we'd be triggering on each change detection cycle.

kara: For some internal apps, we will need to be supporting lists with thousands of items, so performance is definitely an important factor. Using indexOf in this way is not efficient. I don't think adding one method to the selection model adds unnecessary complexity.

crisbeto: Alright, regarding the bulk selection method: if we add bulk selection without emitting an event, we'd need to add one for deselection as well (mostly for consistency). If we're going down this route, I'd prefer adding `emitChangeEvent` instead.

mmalerba: I'm pretty sure the time complexity of this is O(n^2 * log(n)) which is pretty bad. You can still use `sort` if you want, just don't keep calling `indexOf`. create a map of `option => index` and use that map when doing the sort. @kara's suggesting is probably the most optimal though. (using map is O(n * log(n)), filtering is O(n)).

paullryan:
@crisbeto looks like it just needs you to kick the build off again, seems like a badd pre start of the e2e tests, they are passing locally.

crisbeto:
@paullryan this needs a few more changes. It'll trigger next time I push.

crisbeto:
@kara addressed the sorting as discussed. I went with adding a flag to the `SelectionModel` that allows the consumer to disable change events completely.

src/lib/core/option/option.html
```diff
@@ -1,3 +1,5 @@
+<md-pseudo-checkbox class="md-option-pseudo-checkbox" *ngIf="multiple"
```
paullryan: Can potentially cause `Error: The "md-" prefix cannot be used in ng-material v1 compatibility mode.`

crisbeto: Good point. I did this before we did the full switch to the `mat-` prefix and had forgotten about it.

kara:
@crisbeto Can you rebase?

kara:
@crisbeto Can you fix the linting error?

crisbeto:
Sorry @kara, it's fixed now.

src/demo-app/select/select-demo.html
```diff
@@ -37,18 +24,62 @@
...
+        [required]="pokemonRequired" [disabled]="pokemonDisabled" #pokemonControl="ngModel">
+        <md-option *ngFor="let creature of pokemon" [value]="creature.value">
+          {{ creature.viewValue }}
+        </md-option>
+      </md-select>
+      <p> Value: {{ currentPokemon }} </p>
+      <p> Touched: {{ pokemonControl.touched }} </p>
+      <p> Dirty: {{ pokemonControl.dirty }} </p>
+      <p> Status: {{ pokemonControl.control?.status }} </p>
+      <button md-button (click)="currentPokemon=['eevee-4', 'psyduck-6']">SET VALUE</button>
```
kara: Can you move this into a class method?  It's broken in AOT due to https://github.com/angular/angular/issues/14782

kara:
@crisbeto This failed a bunch of presubmit tests. Looking into it to see if it's on our end or theirs

fxck:
Needs rebase. Any idea when it could be merged?

paullryan:
I'm still getting the compatibility mode error when option.html has md-pseudo-checkbox (also effects md-ripple), if I change it to mat-pseudo-checkbox it goes away. I assume this isn't the best way to address this but it's a bit annoying.

mmarimuthu:
can someone explain with an example (html snippet and the action method) for optionSelectionChanges property.

nayfin:
@mmalerba  [Here's](https://github.com/angular/material2/tree/master/src/demo-app/select) a link to the demo code where they give an example about halfway down in the HTML. Essentially you just add the `multiple` directive to your `md-select` component. 

I am working on using this with reactive forms now. If anyone has any examples I would love to see them, else I will post a plunk when I am done. Thanks for the great work @crisbeto !

PR #2730, opened 2017-01-20
fix(autocomplete): placeholder should float while panel is open

kara:
Ready for review

r: @jelbourn


gedclack:
Hi, I am new in github and also new to angular,
I met this situation where Input's placeholder still floats even after I set `floatPlaceholder="never"` in the `<md-input-container>`, is this intentional according to this commit ?

dahaupt:
@gedclack If you have a question you can ask it in https://gitter.im/angular/material2 and in case you think it's an issue in the library, you can open an issue.

PR #2737, opened 2017-01-20
fix(css): change all styles to use the mat- CSS prefix

andrewseguin:
In order to be compatible with material1, all CSS styles should use the mat- prefix rather than md-. Element and attribute selectors for styling should rely on the 'mat-' prefix. Components should automatically add a mat- classname to its host.


src/demo-app/button/button-demo.html
```diff
@@ -3,21 +3,21 @@
     <button md-button>flat</button>
     <button md-raised-button>raised</button>
     <button md-fab>
-      <md-icon class="md-24">check</md-icon>
+      <md-icon class="mat-24">check</md-icon>
```
jelbourn: The `md-24` class actually isn't defined by us; it comes from the material design icon font, so that one needs to change back

src/lib/list/list.scss
```diff
@@ -106,75 +107,75 @@ $md-dense-three-line-height: 76px;
...
-      $md-dense-avatar-height,
-      $md-dense-two-line-height,
-      $md-dense-three-line-height
+  .mat-list-item {
+    @include mat-list-item-base(
+      $mat-dense-font-size,
+                    $mat-dense-base-height,
+                    $mat-dense-avatar-height,
+                    $mat-dense-two-line-height,
+                    $mat-dense-three-line-height
```
jelbourn: Weird indent?

PR #2742, opened 2017-01-21
fix(dialog): close all dialogs on popstate/hashchange

### LINKED ISSUE #2601: MdDialog does not close when browser's back or forward button clicked
Bug (maybe?)

When an MdDialog is open and the user clicks the back button the dialog should automatically close.

The dialog stays open.

You can reproduce by launching the [documentation example](https://material.angular.io/components/component/dialog) and then clicking back. The dialog stays open. At a minimum, I think it would be a good property set on the MdDialogRef

I can see a use case for keeping it open if you wanted to force user interaction, but I think in general its not a good idea to override UI changes triggered by a browsers back or forward button. 

```
angular-cli: 1.0.0-beta.24
node: 6.9.3
os: darwin x64
@angular/common: 2.4.2
@angular/compiler: 2.4.2
@angular/core: 2.4.2
@angular/forms: 2.4.2
@angular/http: 2.4.2
@angular/platform-browser: 2.4.2
@angular/platform-browser-dynamic: 2.4.2
@angular/router: 3.4.2
@angular/compiler-cli: 2.4.2
```

crisbeto:
Closes all of the open dialogs when the user goes forwards/backwards in history.

Fixes #2601.

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -1,4 +1,5 @@
 import {Injector, ComponentRef, Injectable, Optional, SkipSelf, TemplateRef} from '@angular/core';
+import {Location} from '@angular/common';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 import {Overlay, OverlayRef, ComponentType, OverlayState, ComponentPortal} from '../core';
@@ -47,7 +48,16 @@ export class MdDialog {
   constructor(
       private _overlay: Overlay,
       private _injector: Injector,
-      @Optional() @SkipSelf() private _parentDialog: MdDialog) { }
+      @Optional() private _location: Location,
+      @Optional() @SkipSelf() private _parentDialog: MdDialog) {
+
+    // Close all of the dialogs when the user goes forwards/backwards in history or when the
+    // location hash changes. Note that this usually doesn't include clicking on links (unless
+    // the user is using the `HashLocationStrategy`).
+    if (!_parentDialog && _location) {
+      _location.subscribe(() => this.closeAll());
+    }
+  }
 
   /**
    * Opens a modal dialog containing the given component.
--- a/src/lib/dialog/dialog.spec.ts
+++ b/src/lib/dialog/dialog.spec.ts
@@ -18,6 +18,8 @@ import {
 } from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {NoopAnimationsModule} from '@angular/platform-browser/animations';
+import {Location} from '@angular/common';
+import {SpyLocation} from '@angular/common/testing';
 import {MdDialogModule} from './index';
 import {MdDialog} from './dialog';
 import {MdDialogContainer} from './dialog-container';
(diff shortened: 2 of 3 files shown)


src/lib/dialog/dialog.ts
```diff
@@ -56,12 +64,14 @@ export class MdDialog {
   closeAll(): void {
     let i = this._openDialogs.length;
 
-    while (i--) {
-      // The `_openDialogs` property isn't updated after close until the rxjs subscription
-      // runs on the next microtask, in addition to modifying the array as we're going
-      // through it. We loop through all of them and call close without assuming that
-      // they'll be removed from the list instantaneously.
-      this._openDialogs[i].close();
+    if (i > 0) {
```
DevVersion: Is this check necessary? 
> `i--` should return zero at beginning and the `while` would not run at all. And `i` should be always an integer from [0; X]  .

crisbeto: Not really. I'll remove it.

jelbourn:
LGTM

@vicb can you confirm that this is a reasonable thing to do?

src/lib/dialog/dialog.ts
```diff
@@ -29,7 +30,14 @@ export class MdDialog {
...
       private _overlay: Overlay,
       private _injector: Injector,
-      @Optional() @SkipSelf() private _parentDialog: MdDialog) { }
+      private _location: Location,
+      @Optional() @SkipSelf() private _parentDialog: MdDialog) {
+
+    // Close all of the dialogs when the user goes forwards/backwards in history or when the
+    // location hash changes. Note that this usually doesn't include clicking on links (unless
+    // the user is using the `HashLocationStrategy`).
+    _location.subscribe(() => this.closeAll());
```
vicb: you probably want to unsubscribe in `ngOnDestroy`

vicb:
LGTM with un-subscription + test

src/lib/dialog/dialog.ts
crisbeto: This is inside of a singleton service.

jelbourn:
Dialog actually isn't a singleton; in apps with lazy-loading, you necessarily have multiple instances of MdDialog, which means that we need to delegate to the ancestor dialog if there is one.

crisbeto:
Done.

jelbourn:
LGTM

kara:
@crisbeto Rebase?

andrewseguin:
Can you rebase once again :)

mmalerba:
please rebase

crisbeto:
Sorry for the delay on this one, I must've missed the notification.

mmalerba:
There are a number of tests failing in google3 because there is no provider for `Location`. Is this something that can be solved by importing somew module in `MdDialogModule`? or do we just need to make all of these failing tests provide `SpyLocation`?

mmalerba:
@vicb Any advice on how to handle that fact that using `MdDialog` now requires `Location` to be provided in tests? It would be nice if we didn't force everyone to add it to their tests manually

mmalerba:
@crisbeto The way this is currently set up I have to provide `SpyLocation` in everyone's unit tests to get them to not fail, could you make the `Location` optional and just skip the close on fwd/back if its not provided?

PR #2744, opened 2017-01-21
fix(checkbox, radio): not using theme border color

crisbeto:
Fixes the checkbox and radio button not using the border color from the theme. The cause of the issue was that both components used the shorthand `border`, which ended up overriding the theme's `border-color`.


jelbourn:
LGTM

tinayuangao:
LGTM

PR #2747, opened 2017-01-21
fix(overlay): disable pointer events if overlay is detached

### LINKED ISSUE #2739: bug(overlay): cdk-connected-overlay with size dont remove cdk-connected-pane on close
cdk-overlay-pane should disapear on close

cdk-overlay-pane stay open and prevent from clicking on element under it

http://plnkr.co/edit/pH6rE07ldEq5EgZqWzkq?p=preview

DevVersion:
* Currently the overlay element is still visible if the overlay has been detached.
  This is problematic if the overlay pane element has an `height | wídth` set from the state, because now the overlay element may block clicks on the page and overlap the actual content.

* The issue isn't noticable for the `select` and other overlay components, because those don't specify a `height | width` in their state.
   Nevertheless the pane elements are still visible, but just have no height and width set.

Fixes #2739

### DIFF
--- a/src/lib/core/overlay/overlay-directives.spec.ts
+++ b/src/lib/core/overlay/overlay-directives.spec.ts
@@ -31,7 +31,7 @@ describe('Overlay directives', () => {
 
   /** Returns the current open overlay pane element. */
   function getPaneElement() {
-    return overlayContainerElement.firstElementChild as HTMLElement;
+    return overlayContainerElement.querySelector('.cdk-overlay-pane') as HTMLElement;
   }
 
   it(`should attach the overlay based on the open property`, () => {
--- a/src/lib/core/overlay/overlay-directives.spec.ts
+++ b/src/lib/core/overlay/overlay-directives.spec.ts
@@ -39,15 +39,15 @@ describe('Overlay directives', () => {
     fixture.detectChanges();
 
     expect(overlayContainerElement.textContent).toContain('Menu content');
-    expect(getPaneElement().style.visibility)
-      .toBeFalsy('Expected the overlay pane to be visible when attached.');
+    expect(getPaneElement().style.pointerEvents)
+      .toBeFalsy('Expected the overlay pane to enable pointerEvents when attached.');
 
     fixture.componentInstance.isOpen = false;
     fixture.detectChanges();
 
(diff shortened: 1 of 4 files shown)


jelbourn:
@DevVersion detach triggers the animations, so hiding it right away would prevent those from running

DevVersion:
@jelbourn Good point - I think another idea is to disable `pointerEvents` if detached. This also ensures that the overlay pane doesn't block any underlying content.

src/lib/core/overlay/overlay-directives.spec.ts
```diff
@@ -29,16 +29,25 @@ describe('Overlay directives', () => {
     fixture.detectChanges();
   });
 
+  /** Returns the current open overlay pane element. */
+  function getPaneElement() {
```
crisbeto: It may be better to use `querySelector` here, otherwise the test will break if we changed the DOM order.

jelbourn: As long as there's a unit test for the behavior it should be okay

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -35,10 +35,15 @@ export class OverlayRef implements PortalHost {
...
 
     let attachResult = this._portalHost.attach(portal);
+
+    // Update the pane element with the given state configuration.
     this.updateSize();
     this.updateDirection();
     this.updatePosition();
 
+    // Enable pointer events for the overlay pane element.
+    this._togglePointerEvents(true);
```
jelbourn: Aren't these enabled by default?

DevVersion: Yeah they are enabled by default. But developers can always call `attach` after `detach` has been called. This is necessary to revert the `togglePointerEvents(false)` from the detach method.

src/lib/core/overlay/overlay-directives.spec.ts
DevVersion: Yeah it shouldn't be necessary. Nevertheless I just changed it.

DevVersion:
Okay that didn't look pretty good :smile: Should be ready for another review.

PR #2768, opened 2017-01-23
feat(select): allow setting viewValue to override text content

kara:
Edit: I want to think more about this API. Pausing.


kara:
I think this API should match what autocomplete does. Will redo.

PR #2770, opened 2017-01-23
fix(docs): properly create links in guide files

### LINKED ISSUE #2635: broken link to theming guide
On the getting-started page, the link to the theming guide is broken

that the link would take one to https://material.angular.io/guide/theming

it goes to a blank page

go to https://material.angular.io/guide/getting-started
click the link at 'See the theming guide for instructions.'

when developers use this documentation, they should be able to see how to include core themes and styles

DevVersion:
* Adds a processor to the `gulp docs` task that can modify links for the Material documentation
* Right now the processor just rewrites `/guides/` paths to `/guide/` and also removes the file extensions. This makes the guides work properly in the documentation.

**Note**: This is initial implementation of the rewriting. It provides the functionality to rewrite paths and temporary also fixes guides for the documentation.

Fixes #2635

### DIFF
--- a/tools/gulp/tasks/docs.ts
+++ b/tools/gulp/tasks/docs.ts
@@ -50,8 +50,12 @@ function transformMarkdownFiles(buffer: Buffer, file: any): string {
     `<div material-docs-example="${name}"></div>`
   );
 
-  content = content.replace(LINK_PATTERN, (match: string, pre: string, link: string) =>
-    `${pre} href="${fixMarkdownDocLinks(link, file.path)}"`
+  /* Replaces the URL in anchor elements inside of compiled markdown files. */
+  content = content.replace(LINK_PATTERN, (match: string, head: string, link: string) =>
+    // The head is the first match of the RegExp and is necessary to ensure that the RegExp matches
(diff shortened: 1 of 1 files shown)


tools/gulp/tasks/docs.ts
```diff
@@ -37,3 +40,32 @@ task('api', () => {
...
+/** Updates the markdown file's content to work inside of the docs app. */
+function transformMarkdownFiles(buffer: Buffer, file: any): string {
+  let content = buffer.toString('utf-8');
+
+  /* Replace <!-- example(..) --> comments with HTML elements. */
+  content = content.replace(EXAMPLE_PATTERN, (match: string, name: string) =>
+    `<div material-docs-example="${name}"></div>`
+  );
+
+  content = content.replace(LINK_PATTERN, (match: string, pre: string, link: string) =>
```
jelbourn: Can you add comments for what `match`, `pre`, and `link` are expected to contain? (just an example would be good)

DevVersion: Added a comment explaining why `pre` is necessary. I think `match` and `link` should be self-explaining. Can change if desired.

PR #2771, opened 2017-01-23
fix(dialog): prevent error when restoring focus on IE

### LINKED ISSUE #2760: Error after closing dialog in IE if there was no active element when opening the dialog
Bug.

Dialog closes without throwing a TypeError in IE (for dialog that was opened when there was no element with focus).

I open a dialog by clicking an item in a drop down menu. This drop down menu is then closed immediately, which means that there is no element with focus. In [dialog-container](https://github.com/angular/material2/blob/e37259637a97aff0b94d2cdd10b892820a21a309/src/lib/dialog/dialog-container.ts) the element with focus is stored with:

` this._elementFocusedBeforeDialogWasOpened = document.activeElement;`

The [document.activeElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement) documentation states that: 

> When there is no selection, the active element is the page's \<body\> or null. 

I guess that both Chrome and Firefox set the `<body>` as active element, and IE sets it to null. Because only in IE I get the TypeError:

`TypeError: Unable to get property 'focus' of undefined or null reference`

When the dialog is trying to re-focus the element with last focus after closing the dialog:

`(this._elementFocusedBeforeDialogWasOpened as HTMLElement).focus();`

In IE, open dialog without an element having focus, close dialog again.

Affected browser: IE11.

crisbeto:
* Adds an extra null check when restoring focus after a dialog is closed. This is necessary, because the `document.activeElement` can be null in IE.
* Switches the focusing logic to use the `Renderer`.

Fixes #2760.

Note that this is missing a unit test, because it's hard to mock up the `activeElement`.

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -5,6 +5,7 @@ import {
   ViewEncapsulation,
   NgZone,
   OnDestroy,
+  Renderer,
 } from '@angular/core';
 import {BasePortalHost, ComponentPortal, PortalHostDirective, TemplatePortal} from '../core';
 import {MdDialogConfig} from './dialog-config';
@@ -46,7 +47,7 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
   /** Reference to the open dialog. */
   dialogRef: MdDialogRef<any>;
 
-  constructor(private _ngZone: NgZone) {
+  constructor(private _ngZone: NgZone, private _renderer: Renderer) {
     super();
   }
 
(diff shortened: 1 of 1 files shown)


src/lib/dialog/dialog-container.ts
```diff
@@ -90,9 +91,12 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
...
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-      (this._elementFocusedBeforeDialogWasOpened as HTMLElement).focus();
-    });
+    // that it doesn't end up back on the <body>. Also note that we need the extra check, because
+    // IE can set the `activeElement` to null in some cases.
+    if (this._elementFocusedBeforeDialogWasOpened) {
+      this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
+        this._renderer.invokeElementMethod(this._elementFocusedBeforeDialogWasOpened, 'focus');
+      });
+    }
```
jelbourn: FYI: using the renderer is fine here, but not exactly necessary since closing a dialog would always be the result of a user action which can only happen once the app is running in the browser.

Issue #2694, opened 2017-01-17
Update button-toggle styles


jelbourn:
New guidance from Material Design UX:

> Multiple selection uses borders between items since they are unrelated and independent. E.g., Toggling Bold does not affect Italic.
Exclusive selection does not use borders between items since they are related and dependent. E.g., Turning on left aligned means center aligned must be turned off.

### Dark theme colors
background color: grey 900
Active icon: white 100%
Inactive icon: white 30%
background fill on active: white 20%

### Light theme colors
Active icon: 54%
Inactive icon: 26%
background fill on active: back 12%

Guidance on `disabled` button-toggles is that they should look the same as regular, not selected toggle. I think this is rather confusing, and think that we should offer guidance that button-toggles should never be disabled (instead they should be hidden).

jelbourn:
Additional notes from UX:
* button-toggles _should_ have ripples
* The hover / focus / pressed states should behave similar to the spec behavior for dropdowns:

[The spec shows these two images without clarifying](https://material.io/guidelines/components/buttons.html#buttons-toggle-buttons):
![image](https://cloud.githubusercontent.com/assets/838736/22123602/18591a58-de41-11e6-8d0c-3d802ba3bd4c.png)
(Exclusive selection)
![image](https://cloud.githubusercontent.com/assets/838736/22123612/23b095ca-de41-11e6-9ff0-c053ed2ae342.png)
(No options selected)

The new guidance is that former (background color, border, box-shadow) is meant for hover/focus/pressed state, similar to [the dropdown spec](https://material.io/guidelines/components/buttons.html#buttons-dropdown-buttons):
![image](https://cloud.githubusercontent.com/assets/838736/22123688/6bfc1f48-de41-11e6-9375-6fa9039c542f.png)

jelbourn:
Additional note: UX says that the dropdown-like behavior only applies to exclusive-selection button-toggle groups.

WizardPC:
@jelbourn Toggle button will have an hover style and ripples ? Like `mat-button`.

I have a use case with a toolbar where some command (align, bold...) are toggles and some others are buttons (indent, increase size...). 
And it's weird than toggle haven't ripple and hover because in the same toolbar all buttons look like the same, but style behavior is inconsistent...

PR #2773, opened 2017-01-23
refactor(button-toggle): switch to SelectionModel

crisbeto:
* Switches the MdButtonToggleGroup to use the SelectionModel.
* Avoids having to define two toggle group directives.
* No longer fires change events from actions that weren't generated by the user.


EladBezalel:
@crisbeto can you please resolve the conflicts?
#2694 pr is based on some changes here

crisbeto:
Done @EladBezalel

src/lib/button-toggle/button-toggle.ts
```diff
@@ -75,135 +68,183 @@ export class MdButtonToggleGroup implements AfterViewInit, ControlValueAccessor
...
   }
 
   /** Child button toggle buttons. */
   @ContentChildren(forwardRef(() => MdButtonToggle))
-  _buttonToggles: QueryList<MdButtonToggle> = null;
+  _toggles: QueryList<MdButtonToggle> = null;
+
+  ngAfterContentInit(): void {
+    this._model = new SelectionModel<MdButtonToggle>(this.multiple,
+        this._tempValue ? [].concat(this._tempValue) : null);
```
EladBezalel: you can do `[this._tempValue]` instead of `[].concat(this._tempValue)`

src/lib/button-toggle/button-toggle.ts
```diff
@@ -75,135 +68,183 @@ export class MdButtonToggleGroup implements AfterViewInit, ControlValueAccessor
...
+
+    if (this._tempValue) {
+      this._selectByValue(this._tempValue);
+    }
 
-  ngAfterViewInit() {
-    this._isInitialized = true;
+    // Listen to changes in child toggles and re-subscribe when toggles are adde/removed.
+    this._subscribeToToggleChanges();
+    this._toggleSubscription = this._toggles.changes.subscribe(() => {
```
EladBezalel: looks better IMO
```js
this._toggleSubscription = this._toggles.changes.subscribe(() => this._subscribeToToggleChanges());
```

src/lib/button-toggle/button-toggle.ts
```diff
@@ -75,135 +68,183 @@ export class MdButtonToggleGroup implements AfterViewInit, ControlValueAccessor
...
+      this._changeSubscription = null;
+    }
   }
 
-  set selected(selected: MdButtonToggle) {
-    this._selected = selected;
-    this.value = selected ? selected.value : null;
+  /** Subscribes to user-generated changes of the button toggles. */
+  private _subscribeToToggleChanges(): void {
+    let source = Observable.merge.apply(Observable, this._toggles.map(toggle => toggle.change));
```
EladBezalel: maybe [mergeMap](https://www.learnrxjs.io/operators/transformation/mergemap.html) is a better solution?

EladBezalel:
@crisbeto thanks!
@kara can you please review as well?

src/lib/button-toggle/button-toggle.ts
crisbeto: This will end up being wrong if the `_tempValue` is an array.

src/lib/button-toggle/button-toggle.ts
crisbeto: It exceeds the maximum line length from the linter config.

src/lib/button-toggle/button-toggle.ts
crisbeto: I haven't really used `mergeMap` before, but it doesn't seem like it's appropriate here. We only want to merge all of the subscriptions into one so it's easier to listen and unsubscribe to all of them.

src/lib/button-toggle/button-toggle.ts
EladBezalel: yuck

src/lib/button-toggle/button-toggle.ts
EladBezalel: It just feel like there's an operator for that.. maybe `combineAll`?

is `toggle.change` an observable?

crisbeto: Yes, it's an event that is fired when the selected state changes. We use `merge` for a couple of other similar cases as well (in the autocomplete and in the select once the multiple selection PR is merged).

jelbourn:
@crisbeto is this a breaking change for the toggle button API? If so, can you give a quick summary of the change?

crisbeto:
It's not. It just gets rid of the need for having directives both for `md-button-toggle-group` and `md-button-toggle-group[multiple]`.

kara:
@crisbeto Is this style something we want to do? Can you rebase? Didn't catch this one because I wasn't assigned.

crisbeto:
It does make sense, but looking at the amount of conflicts, it'll be easier to redo this from scratch.

crisbeto:
Closing in favor of #9191.

PR #2774, opened 2017-01-23
feat(screenshot): Add a gulp to compare screenshot diffs, upload results to gcs & firebase

tinayuangao:
R: @jelbourn


tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,135 @@
...
+            return compareScreenshots(filenames, database, prNumber);
+          });
+      })
+      .then((results: boolean) => {
+        return database.ref(FIREBASE_REPORT).child(`${prNumber}/result`).set(results);
+      })
+      .then(() => setFilenameList(database, prNumber))
+      .then(() => uploadScreenshots(prNumber, 'diff'))
+      .then(() => uploadScreenshots(prNumber, 'test'))
+      .then(() => database.goOffline(), () => database.goOffline());
```
tinayuangao: Tried but `finally doesn't exist on type Promise`

package.json
```diff
@@ -31,7 +31,10 @@
     "@angular/forms": "^2.2.0",
     "@angular/http": "^2.2.0",
     "@angular/platform-browser": "^2.2.0",
+    "@types/es6-promise": "0.0.32",
```
tinayuangao: I got the `Cannot find name 'Promise'` error. I changed to use firebase.Promise after import 'firebase'

package.json
```diff
@@ -32,6 +32,9 @@
     "@angular/http": "^2.2.0",
     "@angular/platform-browser": "^2.2.0",
     "core-js": "^2.4.1",
+    "firebase": "^3.6.7",
+    "google-cloud": "^0.45.1",
+    "image-diff": "^1.6.3",
```
DevVersion: Any reasons why this is not in the  `devDependencies`?

Also why is the `firebase` dependency needed if we already have `firebase-admin`?

tinayuangao: Removed `firebase` dependency. Old version of `firebase-admin` have wrong types for database.

tinayuangao:

tools/gulp/task_helpers.ts
```diff
@@ -207,3 +207,41 @@ export function openFirebaseDatabase() {
 export function isTravisPushBuild() {
   return process.env['TRAVIS_PULL_REQUEST'] === 'false';
 }
+
+/** Decode the token for Travis to use. */
+function decode(str: string): string {
+  return (str || '').split('\\n').reverse().join('\\n').replace(/\\n/g, '\n');
+}
```
DevVersion: A bit nitpicky, can we move that function below the `openXXX()` functions? (so we have it grouped)

tools/gulp/task_helpers.ts
```diff
@@ -207,3 +207,41 @@ export function openFirebaseDatabase() {
...
+
+/** Open Google Cloud Storage for screenshots */
+export function openScreenshotsCloudStorage() {
+  // Enable Storage
```
DevVersion: I think we can remove that comment. It doesn't really say anything useful?

tools/gulp/task_helpers.ts
```diff
@@ -207,3 +207,41 @@ export function openFirebaseDatabase() {
...
+/** Opens a connection to the firebase realtime database for screenshots. */
+export function openFirebaseScreenshotsDatabase() {
+  // Initialize the Firebase application with admin credentials.
+  // Credentials need to be for a Service Account, which can be created in the Firebase console.
+  let screenshotApp = firebaseAdmin.initializeApp({
+    credential: firebaseAdmin.credential.cert({
+      project_id: 'material2-screenshots',
+      client_email: 'firebase-adminsdk-t4209@material2-screenshots.iam.gserviceaccount.com',
+      // In Travis CI the private key will be incorrect because the line-breaks are escaped.
+      // The line-breaks need to persist in the service account private key.
```
DevVersion: Please move that comment to the `decode` function above (or omit it here)

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+});
+
+function updateFileResult(database: admin.database.Database, prNumber: string,
+                          filenameKey: string, result: boolean): admin.Promise<void>{
+  return database.ref(FIREBASE_REPORT).child(`${prNumber}/results/${filenameKey}`).set(result);
+}
+
+function updateResult(database: admin.database.Database, prNumber: string,
+                      result: boolean): admin.Promise<void> {
```
DevVersion: This seems to return a `boolean` instead of `void`. Can't you remove those `admin.Promise` return types and just let TypeScript detect it properly? (same for others)

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+  return database.ref(FIREBASE_REPORT).child(`${prNumber}/result`).set(result).then(() => result);
+}
+
+function updateTravis(database: admin.database.Database,
+                      prNumber: string): admin.Promise<void> {
+  return database.ref(FIREBASE_REPORT).child(prNumber).update({
+    'commit': process.env['TRAVIS_COMMIT'],
+    'sha': process.env['TRAVIS_PULL_REQUEST_SHA'],
+    'travis': process.env['TRAVIS_JOB_ID'],
```
DevVersion: You can remove the single quotes here (since this is a JSON object)

`{ commit: process.env['TRAVIS_COMMIT'] }`

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+    database.ref(FIREBASE_REPORT).child(reportKey).child('filenames') :
+    database.ref(FIREBASE_FILELIST);
+  return filelistDatabase.set(filenames);
+}
+
+/**
+ * Upload screenshots to google cloud storage.
+ * @param {string} reportKey - The key used in firebase. Here it is the PR number.
+ *   If there's no reportKey, we will upload images to 'golds/' folder
+ * @param {string} mode - Can be 'test' or 'diff' or null.
```
DevVersion: We can remove the `{string}` types here in TypeScript.

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+function getScreenFilenames(database: admin.database.Database): admin.Promise<string[]> {
+  return database.ref(FIREBASE_FILELIST).once('value')
+      .then((snapshots: admin.database.DataSnapshot) => {
+    return snapshots.val();
+  });
+}
+
+/** Upload a list of filenames to firebase database as gold. */
+function setScreenFilenames(database: admin.database.Database,
+                            reportKey?: string): admin.Promise<void> {
```
DevVersion: I personally think it's pretty confusing to name it `reportKey`. How about just calling it `prNumber`? (same for other functions then)

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+  let filenames: string[] = [];
+  readdirSync(SCREENSHOT_DIR).map((file: string) => {
+    let fullName = path.join(SCREENSHOT_DIR, file);
+    let key = file.replace('.screenshot.png', '');
```
DevVersion: I think this is used more than once. I think having a function would be easier here.

Also we can omit the `.screenshot.png` by using `path.basename`

```ts
function extractScreenshotName(fileName: string) {
  return path.basename(fileName, '.screenshot.png');
}
```

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+  if (!existsSync(dirName)) {
+    mkdirSync(dirName, '744');
+  }
+}
+
+/** Download golds screenshots. */
+function downloadAllGolds(
+    filenames: string[], database: admin.database.Database,
+    reportKey: string): admin.Promise<boolean> {
+  _makeDir(`${SCREENSHOT_DIR}/golds`);
```
DevVersion: You can import `fs-extra` above and just use `fse.mkdirp(path.join(SCREENSHOTS_DIR), 'golds');`

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+
+/** Check whether the directory exists. If not then create one. */
+function _makeDir(dirName: string) {
+  if (!existsSync(dirName)) {
+    mkdirSync(dirName, '744');
+  }
+}
+
+/** Download golds screenshots. */
+function downloadAllGolds(
```
DevVersion: I think the name `downloadAllGolds` is not exact. The function actually also runs the diff here.

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+      .then((snapshots: admin.database.DataSnapshot) => {
+    return snapshots.val();
+  });
+}
+
+/** Upload a list of filenames to firebase database as gold. */
+function setScreenFilenames(database: admin.database.Database,
+                            reportKey?: string): admin.Promise<void> {
+  let filenames: string[] = [];
+  readdirSync(SCREENSHOT_DIR).map((file: string) => {
```
DevVersion: FYI: Since you don't map anything you can just use `forEach` instead of `map.

I saw a similar logic below. How about moving this to a function?

```ts
function getLocalScreenshotFiles(): string[] {
  return readdirSync(SCREENSHOT_DIR)
    .filter(fileName => !statSync(path.join(SCREENSHOT_DIR, fileName)).isDirectory())
    .map(fileName => extractScreenshotName(fileName));
}
```

tools/gulp/task_helpers.ts
```diff
@@ -207,3 +207,41 @@ export function openFirebaseDatabase() {
...
   return process.env['TRAVIS_PULL_REQUEST'] === 'false';
 }
+
+/** Decode the token for Travis to use. */
+function decode(str: string): string {
+  return (str || '').split('\\n').reverse().join('\\n').replace(/\\n/g, '\n');
+}
+
+/** Open Google Cloud Storage for screenshots */
+export function openScreenshotsCloudStorage() {
```
DevVersion: I would call this `openScreenshotsBucket()`

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+}
+
+/** Download golds screenshots. */
+function downloadAllGolds(
+    filenames: string[], database: admin.database.Database,
+    reportKey: string): admin.Promise<boolean> {
+  _makeDir(`${SCREENSHOT_DIR}/golds`);
+
+  return admin.Promise.all(filenames.map((filename: string) => {
```
DevVersion: Do we actually need to store the list of gold filenames? I think `google-cloud-node` allows you to query a list of files (See [here](https://googlecloudplatform.github.io/google-cloud-node/#/docs/storage/0.6.0/storage/bucket?method=getFiles))
```ts
let bucket = openScreenshotsBucket();

bucket.getFiles({ prefix: 'golds/' }).then(function(data) {
  var files = data[0];
});


```

tools/gulp/tasks/screenshots.ts
```diff
@@ -0,0 +1,190 @@
...
+  } else {
+    return updateFileResult(database, reportKey, filenameKey, false).then(() => false);
+  }
+}
+
+function decode(value: string): string {
+  return value.split('').reverse().join('');
+}
+
+function updateGithubStatus(result: boolean, prNumber: string) {
```
DevVersion: This is a very nice function! Please move this into a utility function. e.g in `tools/gulp/util-functions.ts`

Be sure to remove the `new admin.Promise` and just use `new Promise` from ES6

tools/gulp/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "declaration": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "lib": ["es6", "es2015"],
+    "lib": ["es6", "es2015", "dom"],
```
DevVersion: Why do we need `DOM` here again?

PR #2776, opened 2017-01-24
fix(autocomplete): up arrow should set last item active

kara:
r: @jelbourn @andrewseguin

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -48,7 +48,7 @@ export class MdAutocompleteTrigger implements AfterContentInit, OnDestroy {
   private _panelOpen: boolean = false;
 
   /** The subscription to positioning changes in the autocomplete panel. */
-  private _panelPositionSub: Subscription;
+  private _panelPositionSubscription: Subscription;
 
   /** Manages active item in option list based on key events. */
   private _keyManager: ActiveDescendantKeyManager;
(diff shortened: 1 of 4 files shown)


src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -553,5 +676,15 @@ class FakeKeyboardEvent {
   preventDefault() {}
 }
 
+class FakeViewportRuler {
```
jelbourn: Isn't there a common `FakeviewportRuler` now?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -5,18 +5,28 @@ import {NgControl} from '@angular/forms';
...
+ * actually focusing the active item, scroll must be handled manually.
+ */
 
-/** The panel needs a slight y-offset to ensure the input underline displays. */
-export const MD_AUTOCOMPLETE_PANEL_OFFSET = 6;
+/** The height of each autocomplete option. */
+export const AUTOCOMPLETE_OPTION_HEIGHT = 48;
+
+/** The total height of the autocomplete panel. */
+export const MD_AUTOCOMPLETE_PANEL_HEIGHT = 256;
```
kara: Good catch; not sure why I did that.

src/lib/autocomplete/autocomplete.spec.ts
kara: Oh hey. Didn't know about that, but that's way better. Will replace.

src/lib/core/a11y/list-key-manager.ts
```diff
@@ -96,12 +96,13 @@ export class ListKeyManager<T extends CanDisable> {
...
   setNextItemActive(): void {
-    this._setActiveItemByDelta(1);
+    this._activeItemIndex === null ? this.setFirstItemActive() : this._setActiveItemByDelta(1);
   }
 
   /** Sets the active item to a previous enabled item in the list. */
   setPreviousItemActive(): void {
-    this._setActiveItemByDelta(-1);
+    this._activeItemIndex === null && this._wrap ? this.setLastItemActive()
+                                                 : this._setActiveItemByDelta(-1);
```
jelbourn: Need to handle the case when `activeItemIndex` is null and wrap is false

jelbourn: (with a unit test)

jelbourn:
LGTM

PR #2781, opened 2017-01-24
chore: move examples in from docs site repo

riavalon:


src/examples-source/button-overview-example-css.html
```diff
@@ -0,0 +1 @@
+<span class="hljs-comment">/** No CSS for this example */</span>
```
jelbourn: The generated html files shouldn't be added to source control, just the examples themselves (what you currently have under `src/examples`). These html files are currently generated by the `highlight-examples.sh` script, which we can move to this repo in a follow-up PR

PR #2790, opened 2017-01-25
feat(compatibility): remove conflicts with material1 css styling

andrewseguin:
This change establishes that styling of material components should be done using a 'mat' prefix rather than 'md'. This is necessary for apps that want to upgrade from AngularJS Material to Angular Material and host both versions at the same time on their page. Without this change, there would be styling conflicts between the two libraries.

BREAKING CHANGE: All styling now uses the 'mat' prefix rather than 'md'. This includes changes to variables such as md-core, md-elevation, md-light-theme, and so on. This was necessary to allow apps to host both AngularJS Material and Angular Material on the same page without styling conflicts.


jelbourn:
LGTM if this is identical to previous PR

andrewseguin:
Should be identical, good sign that tests pass. Dev site looks good to me as well.

jelbourn:
LGTM

benb7760:
@andrewseguin as far as I can see, this needs to be reflected in the material.angular.io site? Specifically https://material.angular.io/guide/theming

andrewseguin:
It sure does, thanks for the callout. Will be fixed in #3013

andrewseguin:
The md- prefix is staying for components and directives, unless you turn on compatibility mode which forces you to use mat- for component and directive selectors. This is to allow projects to upgrade from AngularJS Material (1.x) to Angular Material (this version) without having their directives and styling collide.

In general, the rule going forward is that styling will be prefixed by mat- and selectors will be prefixed by md-.

alexndreazevedo:
Could you use semantic version on npm package for this update? Although the library still giving support to the directives with the same name, the css styles, scss mixins and variables doesn't have backward compatibility. It crashes the code of whom is reusing the mixins and variables for customize the styles of Material Design.

In this case, the most appropriate should improve the MAJOR on semver, that is better than just change from 2.0.0-beta.1 to 2.0.0-beta.2.

ChristianBeilschmidt:
This is inconsistent. You should rename all components to `mat-` as well then.

pabx06:
it worked for my styles thanks.

PR #2797, opened 2017-01-25
feat(select): Add search support to the material2 select

Eernie:
This adds a search field to the md-select.

R: @jelbourn @kara

### DIFF
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -33,6 +33,26 @@
     <button md-button (click)="drinkControl.reset()">RESET</button>
   </md-card>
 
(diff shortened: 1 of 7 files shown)


jelbourn:
Hi @Eernie! We appreciate the enthusiasm, but unfortunately this doesn't align with our plans for supporting this feature. Rather than hard-coding a search into the select, we will be adding a more generic "select header" similar to [the AngularJS Material version](https://material.angularjs.org/latest/demo/select). This approach provides more flexibility for applications to use custom inputs and custom filtering. Feel free to open an issue for discussion first if you want to add similar features in the future.

PR #2802, opened 2017-01-25
build: use typescript cli for compilation

DevVersion:
* No longer use the `gulp-typescript` gulp package that tries to take care of several options like `sourceMaps` or `declarations`.
* Don't generate declarations when serving the `demo-app`. (Slows down testing)

Advantages:

* Use the `Typescript` compiler CLI to ensure that the `NGC` also works with our `tsconfig.json` files.
* No longer makes our TypeScript build dependent to an external plugin (~with magic)
* Fixes issues like invalid sourceMaps from `gulp-typescript`: https://github.com/angular/material2/pull/2753
* Allows us to take use of all TypeScript features (e.g The `sourceMaps` option)
* This should help for our planned `tsconfig.json` file reduction.

### DIFF
--- a/src/demo-app/tsconfig.json
+++ b/src/demo-app/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "declaration": true,
+    "declaration": false,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "lib": ["es6", "es2015", "dom"],
@@ -22,8 +22,11 @@
       ]
     }
   },
-  "exclude": [
-    "*-aot.ts"
+  "files": [
+    "./demo-app-types.d.ts",
+    "./demo-app-module.ts",
+    "./system-config.ts",
+    "./main.ts"
   ],
   "angularCompilerOptions": {
     "genDir": "../../dist",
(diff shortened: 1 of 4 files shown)


src/demo-app/tsconfig.json
```diff
@@ -22,8 +22,11 @@
       ]
     }
   },
-  "exclude": [
-    "*-aot.ts"
+  "files": [
+    "./demo-app-types.d.ts",
+    "./demo-app-module.ts",
+    "./system-config.ts",
+    "./main.ts"
```
jelbourn: In the past WebStorm had issues when you specified `files` where it would only provide completion / error checking for _exactly_ these files and not their dependencies. Are you able to tell if that's still an issue?

DevVersion: Just looked at it. The issue you described has been solved in Webstorm. 

I just noticed that if you create a file (that is never referenced) Webstorm will complain for example about `experimentalDecorator` support. I don't consider this is as a big issue. What are your thoughts?

jelbourn: That's probably okay- we shouldn't have any dead code anyway.

DevVersion: Yeah, also I think that it's very typically for TS to not use a `glob` to walk through the files. Rather walking through the chain of files. Should we apply the merge ready label?

PR #2807, opened 2017-01-25
fix(slider): fire change event on value change via keyboard.

### LINKED ISSUE #2764: md-slider the keyboard bindings doesn't change the values
Bug

When you change the slider using the keyboard bindings (Left arrow or Right arrow), it is expected changing the value using [(ngModel)] and triggering (change) event.

But slider moves after pressing button.

http://plnkr.co/edit/mUQx5dK8wt3qtJ04NOcF?p=preview

  "@angular/core": "^2.4.1",
  "@angular/material": "^2.0.0-beta.0",
  win 7,
  chrome 55.0.

mmalerba:
also disable animation for keyboard changes because they look crazy.

fixes #2764

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -75,6 +75,7 @@ export class MdSliderChange {
     '(blur)': '_onBlur()',
     '(click)': '_onClick($event)',
     '(keydown)': '_onKeydown($event)',
+    '(keyup)': '_onKeyup()',
     '(mouseenter)': '_onMouseenter()',
     '(slide)': '_onSlide($event)',
     '(slideend)': '_onSlideEnd()',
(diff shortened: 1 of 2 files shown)


src/lib/slider/slider.ts
```diff
@@ -464,12 +465,19 @@ export class MdSlider implements ControlValueAccessor {
...
 
+  _onKeyup() {
+    this._isSliding = false;
+  }
+
   /** Increments the slider by the given number of steps (negative number decrements). */
   private _increment(numSteps: number) {
     this.value = this._clamp(this.value + this.step * numSteps, this.min, this.max);
+    this._emitInputEvent();
+    this._emitValueIfChanged();
```
jelbourn: Can we add either a unit or e2e test for this?

mmalerba:
tests added

Issue #2674, opened 2017-01-16
<md-sidenav>: remove shadows in persistent mode (mode='side'), or add @Input like "shadow='on|off|auto'"


m0nzderr:
#### Bug, feature request, or proposal:
I would rather consider it a feature proposal or improvement, than a bug. But one could see it as a bug  if taking the MD spec literally.

#### What is the expected behavior?
The spec (https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-behavior)
states that permanent drawer (i.e., mode='side') should stay at the same elevation as the content (i.e., no shadow).

#### What is the current behavior?
md-sidenav allways stays above the content dropping shadow.
 
#### What are the steps to reproduce?
See, e.g, the Examples tab at https://material.angular.io/components/component/sidenav

#### What is the use-case or motivation for changing an existing behavior?
Implementing the persistent navigation drawer behavior, as described here: https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-behavior

And being able to deliver good design, of course!

#### Which versions of Angular, Material, OS, browsers are affected?
@angular/material 2.0.0.beta-1

#### Is there anything else we should know?
Thumbs up for the lib. Looking forward to new releases! I'm already moving to production with it.

PR #2809, opened 2017-01-25
fix(sidenav): hide drawer shadow in 'side' mode

mmalerba:


tinayuangao:
LGTM

mmalerba:
fixes #2674

andrewseguin:
Was this to match a closer spec to material design? I haven't seen an example of a persistent sidenav that doesn't have any depth/border to it. I don't think this looks as nice as before. I think mat-elevation-16 was overkill but I'd recommend putting in at least mat-elevation-2.

mmalerba:
@andrewseguin yes this was to match the spec: https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-behavior no reason you can't just add the elevation class in your app though

Issue #2817, opened 2017-01-26
build(sauclabs): access


Nightapes:
Should be the access keys in the repro? 
https://github.com/angular/material2/blob/master/.travis.yml#L15
https://github.com/angular/material2/blob/master/.travis.yml#L17

DevVersion:
Right now they are intentionally in the `.travis.yml` to support Sauce & Browserstack for PR's as well.

And yeah we have been thinking about encrypting those more.

DevVersion:
Update, just discovered that encrypted tokens in the `.travis.yml` are not accessible for PRs. I thought that only variables from the Saucelabs Web UI are not available for PRs

Closing this issue now, since we don't have any other option.

Nightapes:
For saucelabs there is a way: https://docs.travis-ci.com/user/jwt. But unfortunately not supported by Browserstack.

DevVersion:
@Nightapes Thanks for the doc. Will look into this later.

PR #2818, opened 2017-01-26
chore: use encrypted tokens for api keys

DevVersion:
* Instead of saving the plain access tokens in the `.travis.yml` the tokens are now encrypted.


DevVersion:
https://github.com/angular/material2/issues/2817#issuecomment-275465982

PR #2819, opened 2017-01-26
chore: use encrypted tokens for api keys

DevVersion:
### TEST
* Instead of saving the plain access tokens in the `.travis.yml` the tokens are now encrypted.


DevVersion:
Closing & and leaving the tokens are they are for now.

The Travis JWT addon is only allowing the tokens to be used from trusted addons and therefore wouldn't help us encrypting the tokens.

> https://github.com/travis-ci/travis-ci/issues/7223

Issue #2750, opened 2017-01-22
Align font sizes


mbeckenbach:
#### Bug, feature request, or proposal:
The checkbox font size and button font size are different.

![image](https://cloud.githubusercontent.com/assets/6654388/22181939/36482a1e-e098-11e6-9325-2793083550ab.png)

#### What is the expected behavior?
Should be the same by default.

#### What is the current behavior?
See screenshot

#### What are the steps to reproduce?
Create a form with a checkbox and a button.

#### What is the use-case or motivation for changing an existing behavior?
The default css should enable good looking form layouts without changing anything by default.

#### Which versions of Angular, Material, OS, browsers are affected?
ng 2.3.1
md 2.0.0-beta.1

DevVersion:
Thanks for the issue @mbeckenbach. I can confirm that the `md-checkbox` does not set any `font-family`.

Other components probably don't set a `font-family` as well. We need to determine whether we want to set the `font-family` on all components (with a text)

cc. @jelbourn

DevVersion:
The following components already apply the `font-family`
- Card
- Button
- Input Container & Input
- List
- Slide Toggle
- Tabs
- Toolbar
- Tooltip

Really feel like we should add this to the other components as well.

PR #2821, opened 2017-01-26
fix: apply font-family to text components

DevVersion:
* Applies the `font-family` to all components that use content projection and have a label or a text displayed.
* Switches the hard-coded font-family in the `simple-snack-bar` to the global variable.

**Notes:**
- All families have been applied to the host elements (to simplify font overwriting)
- The dialog font family has been skipped here. Not sure whether we want that for the users instantiated component (I'd only say `md-dialog-content` etc. directives)

References #2750


jelbourn:
LGTM, though we may revisit once we work on a more comprehensive typography approach

PR #2822, opened 2017-01-26
feat(dialog): add hasBackdrop and backdropClass options to dialog config

### LINKED ISSUE #2806: Style overlay backdrop for specific dialogs
**Feature Request**

Specify custom CSS classes for overlay backdrops for certain dialogs.

Backdrops can not be individually styled. If I wanted to remove or style a backdrop it would have to be styled for all backdrops on all dialogs.

To remove/style backdrops for specific dialogs. One dialog may not need a backdrop, or a backdrop with a different color or opacity.

YeomansIII:
Implement hasBackdrop as laid out in the MdDialog design document. Add backdropClass option to override default `cdk-overlay-dark-backdrop` class, allowing custom styling to be applied.

Closes #2806 
Somewhat related https://github.com/angular/material2/pull/1584

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -109,7 +109,7 @@ export class MdDialog {
   private _getOverlayState(dialogConfig: MdDialogConfig): OverlayState {
     let overlayState = new OverlayState();
     overlayState.hasBackdrop = dialogConfig.hasBackdrop;
-    if(dialogConfig.backdropClass) {
+    if (dialogConfig.backdropClass) {
       overlayState.backdropClass = dialogConfig.backdropClass;
     }
     overlayState.positionStrategy = this._overlay.position().global();
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -108,7 +108,10 @@ export class MdDialog {
    */
   private _getOverlayState(dialogConfig: MdDialogConfig): OverlayState {
     let overlayState = new OverlayState();
-    overlayState.hasBackdrop = true;
+    overlayState.hasBackdrop = dialogConfig.hasBackdrop;
+    if(dialogConfig.backdropClass) {
+      overlayState.backdropClass = dialogConfig.backdropClass;
+    }
     overlayState.positionStrategy = this._overlay.position().global();
 
     return overlayState;
--- a/src/demo-app/dialog/dialog-demo.html
+++ b/src/demo-app/dialog/dialog-demo.html
@@ -43,6 +43,16 @@
       </md-input-container>
     </p>
 
+    <h2>Dialog backdrop</h2>
+
+    <p>
+      <md-input-container>
+        <input mdInput [(ngModel)]="config.backdropClass" placeholder="Backdrop class">
+      </md-input-container>
+    </p>
+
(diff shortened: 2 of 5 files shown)


YeomansIII:
It looks like saucelabs timed out in the last build, not sure if someone here has the ability to redo the build.

crisbeto:
Reset the tests for you.

YeomansIII:
Are there any other steps I should take for this pull request? Or is the only thing left to do wait for someone from the material2 team to take a look?

crisbeto:
It still needs to be reviewed. Also I can see how the `hasBackdrop` _may_ be useful, but I'm not quite sure about `backdropClass`. What do you think @jelbourn?

YeomansIII:
I was actually thinking hasBackdrop is probably less useful. When hasBackdrop is false you also lose the ability to close when clicking away, which undermines disableClose. However, with backdropClass you can make the backdrop transparent and then control whether you want the ability to close or not with disableClose.

backdropClass only replaces the default `cdk-overlay-dark-backdrop` class which simply applies an rgba background.

The library is already offering the ability to set the position of a dialog, allowing people to use it for more than just a simple standard dialog, so I don't see how allowing custom styling of the backdrop is not useful.

YeomansIII:
@crisbeto Is it possible to get this reviewed or at least discussed? It's been months.

Beta.3 breaks the hack I was using to have some dialogs with transparent backgrounds and it looks like at least one other person that found this is also interested (looking at the reaction to my last comment). From what I have researched the Material Guidelines don't specify/require a specific background overlay color. The API already gives the ability to customize position, width, height, so obviously it's intended to be used in many ways. The Overlay API already supports the ability to apply custom classes or just use the existing transparent class. Why should we be forced to use the translucent black overlay?

This is an incredibly easy change and I can fix this PR to get it up to date with the latest master, but it would be appreciated if I could get some input from the maintainers, even if it's just a "Yes" or "No."

YeomansIII:
Here is an example of my use case.

The dialog below is our "Share" dialog. It opens at the position of the share icon when you click, using the Material API's ability to set the position of the dialog. Since this dialog acts similar to a menu, we want a transparent overlay backdrop. Currently this is not possible with the Material2 API.
Dialog with transparent overlay backdrop

The following dialog is our Login Dialog. This is more like a traditional modal/dialog and should have a black translucent overlay backdrop. This is currently the only backdrop allowed using the Material API.
Dialog with black translucent overlay backdrop

src/lib/dialog/dialog.ts
```diff
@@ -161,7 +161,10 @@ export class MdDialog {
     let strategy = this._overlay.position().global();
     let position = dialogConfig.position;
 
-    state.hasBackdrop = true;
+    state.hasBackdrop = dialogConfig.hasBackdrop;
+    if (dialogConfig.backdropClass && dialogConfig.backdropClass !== '') {
```
jelbourn: The extra `!== ''` shouldn't be necessary

YeomansIII:
@jelbourn Resolved conflicts with master and made changes based on your suggestion.

jjun3188:
@YeomansIII Thank you for your great job. Do we know briefly when this will be released?

PR #2823, opened 2017-01-26
build: simplify karma configuration

DevVersion:
* No longer use ts-node to build the Karma configuration (just to be able to import the browser-providers)
* Transformed the browser-providers into a JS file (with only the most necessary exports)
* Separates the list of possible external browsers into an extra file (as in Material 1)

This change just removes unnecessary bloat from our build process and also makes running karma a bit faster, because we don't need to build TS on the fly.

Also it just makes the `browser-providers`, `karma.conf.js` and `remote_browsers` very clear.

@jelbourn I noticed this while trying to integrate coverage. And I also planned to remove unnecessary bloat of our build process (right now it's too confusing & big)


test/browser-providers.js
```diff
@@ -0,0 +1,65 @@
...
+  'IE11':         { unitTest: {target: 'SL', required: true  }},
+  'Edge':         { unitTest: {target: 'SL', required: true  }},
+  'Android4.1':   { unitTest: {target: null, required: false }},
+  'Android4.2':   { unitTest: {target: null, required: false }},
+  'Android4.3':   { unitTest: {target: null, required: false }},
+  'Android4.4':   { unitTest: {target: 'SL', required: false }},
+  'Android5':     { unitTest: {target: 'SL', required: false }},
+  'Safari7':      { unitTest: {target: null, required: false }},
+  'Safari8':      { unitTest: {target: 'BS', required: false }},
+  'Safari9':      { unitTest: {target: 'BS', required: false }},
```
jelbourn: Idle thought: looks like Browserstack still doesn't support Safari 10, but SauceLabs does. I wonder if we should try adding that (in a follow-up PR)

DevVersion: Good idea. I will take a look at this once this PR is merged.

PR #2833, opened 2017-01-27
docs(readme): add link to readme for material2 select

NColey:
I noticed that the feature list of the README was missing a link to the associated README for the select feature so I updated the docs to include it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ High level items planned for January 2017:
 | ripples          |                           Available | [README][19] |   [#108][0108] |
 | dialog           |                           Available | [README][22] |   [#114][0114] |
 | snackbar / toast |                           Available | [README][21] |   [#115][0115] |
-| select           |                           Available |           -  |   [#118][0118] |
+| select           |                           Available | [README][23] |   [#118][0118] |
 | textarea         |                           Available |           -  |              - |
 | autocomplete     |                         In-progress |           -  |   [#117][0117] |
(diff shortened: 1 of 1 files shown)


NColey:
I signed the CLA! Thanks

YeomansIII:
Shouldn't these just be directly linking to material.angular.io now? The individual READMEs all just have links to material.angular.io.

NColey:
@YeomansIII that's what I thought but when I checked the existing main README here: https://github.com/angular/material2 it looks like they are all still linking to the README and then when you click the README link it takes you to a page like this: https://github.com/angular/material2/blob/master/src/lib/snack-bar/README.md which has the link to the docs. I set up my change to match what I was seeing in the existing README but I can update it to link directly to the docs for select at material.angular.io. I just didn't want to do anything that didn't match the setup of what is currently in the main README.

YeomansIII:
Not necessarily knocking your pull request, was just kind of an open air question for whoever reviews this (if they do).  Would make more sense to just link directly to material.angular.io from the root README instead of making people chase it down.

NColey:
I agree, and I debated whether to link directly or to link to the README like how the existing one's are set up but then I found this commit: https://github.com/angular/material2/commit/09591cf123dc06a948bc09012b818ad000f60efd and wondered if it was intentional to link to the README which then links to the official docs. This is my first time making a pull request for material2 so I don't totally know the process but I would be happy to update my commit with the direct link if that's preferable.

PR #2834, opened 2017-01-27
fix(autocomplete): close panel when options list is empty

kara:
Ready for review.

r: @jelbourn @andrewseguin

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -4,8 +4,10 @@ import {
     ElementRef,
     forwardRef,
     Input,
+    NgZone,
     Optional,
     OnDestroy,
+    QueryList,
     ViewContainerRef,
 } from '@angular/core';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
@@ -18,6 +20,7 @@ import {MdOptionSelectEvent, MdOption} from '../core/option/option';
 import {ActiveDescendantKeyManager} from '../core/a11y/activedescendant-key-manager';
 import {ENTER, UP_ARROW, DOWN_ARROW} from '../core/keyboard/keycodes';
 import {Subscription} from 'rxjs/Subscription';
+import 'rxjs/add/observable/of';
 import 'rxjs/add/observable/merge';
 import {Dir} from '../core/rtl/dir';
 import 'rxjs/add/operator/startWith';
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -131,7 +134,7 @@ export class MdAutocompleteTrigger implements AfterContentInit, ControlValueAcce
    * A stream of actions that should close the autocomplete panel, including
    * when an option is selected and when the backdrop is clicked.
    */
-  get panelClosingActions(): Observable<any> {
+  get panelClosingActions(): Observable<MdOptionSelectEvent | null> {
```
jelbourn: What's your reasoning for doing `| null` here since we haven't been doing explicit null yet?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -149,6 +152,11 @@ export class MdAutocompleteTrigger implements AfterContentInit, ControlValueAcce
     return this._keyManager.activeItem as MdOption;
   }
 
+  /** The initial list of autocomplete options, as soon as the zone has stabilized */
+  get initialOptionList(): Observable<QueryList<MdOption>> {
```
jelbourn: I don't think this is really the initial list of options- it's just whatever the next list of options are once the zone stabilizes; the fact that you call it on initialization is separate from what you're getting. So you could change this to `_getStableOptions` and then in the subscribe function do
```ts
let initialOptions = this._getStableOptions();`
```

Also should be private or internal.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -149,6 +152,11 @@ export class MdAutocompleteTrigger implements AfterContentInit, ControlValueAcce
...
+    return this._zone.onStable.first().map(() => this.autocomplete.options);
```
jelbourn: Add comment explaining why you need to do this

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -173,6 +171,25 @@ describe('MdAutocomplete', () => {
...
       });
+
+    }));
+
+    it('should close the panel when the options list is empty', async(() => {
+      dispatchEvent('focus', input);
+      fixture.detectChanges();
+
+      fixture.whenStable().then(() => {
+        // Filter down the option list so no options match the value
```
jelbourn: "so" => "such that"

src/lib/autocomplete/autocomplete-trigger.ts
kara: Just thought we should start so that it's more explicit. It's not super important to change in this PR, so I can revert for now.

src/lib/autocomplete/autocomplete-trigger.ts
kara: That's a good point. Will do.

kara:
@jelbourn Comments addressed.

PR #2835, opened 2017-01-27
fix(autocomplete): double-clicking input shouldnt close the panel

kara:
Ready for review.

r: @jelbourn @andrewseguin

### DIFF
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -60,6 +60,7 @@ describe('MdAutocomplete', () => {
     it('should open the panel when the input is focused', () => {
       expect(fixture.componentInstance.trigger.panelOpen)
           .toBe(false, `Expected panel state to start out closed.`);
+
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -85,22 +87,18 @@ describe('MdAutocomplete', () => {
           .toContain('California', `Expected panel to display when opened programmatically.`);
     });
 
-    it('should close the panel when a click occurs outside it', async(() => {
+    it('should close the panel when blurred', async(() => {
```
andrewseguin: Can you include a test checking that the panel is open after changing focus to an md-option?

kara: You mean when you make the  md-option active by the keyboard, right? Focus never changes to md-option unless you click on it, in which case the panel should close.

kara: Added another test in the keyboard events section for that

PR #2839, opened 2017-01-27
create SimpleDate and CalendarLocale objects

mmalerba:


jelbourn:
FYI @pkozlowski-opensource

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,145 @@
+import {SimpleDate} from './simple-date';
+import {Injectable} from '@angular/core';
+
+
+const SUPPORTS_INTL_API = !!(new Date().toLocaleDateString);
```
jelbourn: Safari 9 has `toLocaleDateString` but not the full `Intl` API.

Here's the check that modernizr uses: 
https://github.com/Modernizr/Modernizr/blob/master/feature-detects/intl.js

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,145 @@
...
+        this.shortMonths[date.month] + ' ' + date.year;
+  }
+
+  getCalendarYearHeaderLabel(date: SimpleDate) {
+    return SUPPORTS_INTL_API ?
+        date.toNativeDate().toLocaleDateString(undefined, {year: 'numeric'}) : String(date.year);
+  }
+
+  protected createArray<T>(length: number, valueFunction: (index: number) => T): T[] {
+    return Array.apply(null, Array(length)).map((v: undefined, i: number) => valueFunction(i));
```
jelbourn: Why not just `Array(length).map(...)` ?

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,145 @@
...
+        date.toNativeDate().toLocaleDateString(undefined, {month: 'short', year: 'numeric'}) :
+        this.shortMonths[date.month] + ' ' + date.year;
+  }
+
+  getCalendarYearHeaderLabel(date: SimpleDate) {
+    return SUPPORTS_INTL_API ?
+        date.toNativeDate().toLocaleDateString(undefined, {year: 'numeric'}) : String(date.year);
+  }
+
+  protected createArray<T>(length: number, valueFunction: (index: number) => T): T[] {
```
jelbourn: This function is classically called `range`. Rename to that and move it outside the class as just a module function?

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  protected createMonthsArray(format: string) {
+    return this.createArray(12,
+        i => new Date(2017, i, 1).toLocaleDateString(undefined, {month: format}));
+  }
+
+  protected createDaysArray(format: string) {
+    return this.createArray(7,
+        i => new Date(2017, 0, i + 1).toLocaleDateString(undefined, {weekday: format}));
```
jelbourn: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString#Performance):
> When formatting large numbers of dates, it is better to create an Intl.DateTimeFormat object and use the function provided by its format property.

Not sure if this counts as a "large number", but it wouldn't hurt to make the formatter

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * Gets a label to display as the heading for the specified calendar year.
+   * @param date A date that falls within the year to be labeled.
+   */
+  getCalendarYearHeaderLabel: (date: SimpleDate) => string;
+}
+
+
+export class DefaultCalendarLocale implements  CalendarLocale {
```
jelbourn: Add class description

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,145 @@
+import {SimpleDate} from './simple-date';
+import {Injectable} from '@angular/core';
+
+
+const SUPPORTS_INTL_API = !!(new Date().toLocaleDateString);
+
+
+@Injectable()
+export abstract class CalendarLocale {
```
jelbourn: Add class description

src/lib/core/datetime/calendar-locale.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(calendarLocale.shortDays).toEqual(['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat']);
+  });
+
+  it('lists narrow days', () => {
+    expect(calendarLocale.narrowDays).toEqual(['S', 'M', 'T', 'W', 'T', 'F', 'S']);
+  });
+
+  it('lists dates', () => {
+    expect(calendarLocale.dates).toEqual([
+      null, '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16',
```
jelbourn: Why does this start with null?

src/lib/core/datetime/calendar-locale.spec.ts
```diff
@@ -0,0 +1,90 @@
+import {inject, TestBed, async} from '@angular/core/testing';
+import {CalendarLocale} from './calendar-locale';
+import {DatetimeModule} from './index';
+import {SimpleDate} from './simple-date';
+
+
+describe('DefaultCalendarLocale', () => {
```
jelbourn: These tests don't seem to well-capture using the current locale with fallback behavior. Any ideas on how to improve this?

src/lib/core/datetime/simple-date.ts
```diff
@@ -0,0 +1,11 @@
+export class SimpleDate {
```
jelbourn: Add class description

src/lib/core/datetime/calendar-locale.spec.ts
mmalerba: There's a comment in the base class, basically I thought it would be nice to not have to do `calendarLocale.dates[date + 1]` all the time (especially since months and days are 0-indexed), so I added `null` as the representation for `dates[0]`

src/lib/core/datetime/calendar-locale.ts
mmalerba: Apparently that just creates an object like `{length: 7}` without adding keys for the indices, which causes `map` to not work.

src/lib/core/datetime/calendar-locale.spec.ts
mmalerba: unfortunately I don't think there's any way to set the browser's locale from JS

mmalerba:
+ @andrewseguin since kara is busy caretaking this week

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,168 @@
...
+    return range(7, i => dtf.format(new Date(2017, 0, i + 1)));
+  }
+
+  private _createDatesArray(format: string) {
+    let dtf = new Intl.DateTimeFormat(undefined, {day: format});
+    return range(31, i => dtf.format(new Date(2017, 0, i + 1)));
+  }
+
+  private _createFormatFunction(
+      options: Object, fallback: (date: SimpleDate) => string): (date: SimpleDate) => string {
```
andrewseguin: fix identations

src/lib/core/datetime/calendar-locale.ts
```diff
@@ -0,0 +1,168 @@
...
+    let dtf = new Intl.DateTimeFormat(undefined, {weekday: format});
+    return range(7, i => dtf.format(new Date(2017, 0, i + 1)));
+  }
+
+  private _createDatesArray(format: string) {
+    let dtf = new Intl.DateTimeFormat(undefined, {day: format});
+    return range(31, i => dtf.format(new Date(2017, 0, i + 1)));
+  }
+
+  private _createFormatFunction(
```
andrewseguin: Can you leave a description of this function? It's especially difficult to visually parse the signature of the function

mmalerba: done, also simplified it since its just as easy to return null and specify the fallback elsewhere

PR #2840, opened 2017-01-27
docs(autocomplete): add autocomplete docs

kara:
Docs site example to come once examples live here.

r: @jelbourn @andrewseguin

### DIFF
--- a/src/lib/autocomplete/OVERVIEW.md
+++ b/src/lib/autocomplete/OVERVIEW.md
@@ -1,0 +1,150 @@
+
+The autocomplete is a normal text input enhanced by a panel of suggested options. You can read more about 
+autocompletes in the [Material Design spec](https://material.io/guidelines/components/text-fields.html#text-fields-auto-complete-text-field).
+
+### Simple autocomplete
+
+Start by adding a regular `mdInput` to the page. Let's assume you're using the `formControl` directive from the 
+`@angular/forms` module to track the value of the input.
+
+*my-comp.html*
+```html
+<md-input-container>
+   <input type="text" mdInput [formControl]="myControl">
+</md-input-container>
+```
+
+Next, create the autocomplete panel and the options displayed inside it. Each option should be defined by an 
+`md-option` tag. Set each option's value property to whatever you'd like the value of the text input to be 
+upon that option's selection.
+ 
+*my-comp.html*
+```html
+<md-autocomplete>
+   <md-option *ngFor="let option of options" [value]="option">
+      {{ option }}
+   </md-option>
+</md-autocomplete>
+```
+
+Now we'll need to link the text input to its panel. We can do this by exporting the autocomplete panel instance into a 
+local template variable (here we called it "auto"), and binding that variable to the input's `mdAutocomplete` property.
+
+*my-comp.html*
+```html
+<md-input-container>
+   <input type="text" mdInput [formControl]="myControl" [mdAutocomplete]="auto">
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/OVERVIEW.md
```diff
@@ -0,0 +1,149 @@
...
+      {{ option }}
+   </md-option>
+</md-autocomplete>
+```
+
+*my-comp.ts*
+```ts
+class MyComp {
+   myControl = new FormControl();
+   options = [
```
jelbourn: Use a real type here and get rid of `any` throughout? Maybe something like `new User('Mary', 'Shelly')`

src/lib/autocomplete/OVERVIEW.md
```diff
@@ -0,0 +1,149 @@
...
+   </md-option>
+</md-autocomplete>
+```
+
+### Adding a custom filter
+
+At this point, the autocomplete panel should be toggleable on focus and options should be selectable. But if we want 
+our options to filter when we type, we need to add a custom filter. 
+
+We already have access to the built-in `valueChanges` observable on the `FormControl`, so we can simply map the text
```
jelbourn: This feels a bit like you're saying "This is the One True Way to do filtering". What do you think of approaching more as "You can filter the options in any way you want based on the text input. Here's one example."

src/lib/autocomplete/OVERVIEW.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Setting separate control and display values
+
+If you want the option's control value (what is saved in the form) to be different than the option's display value 
+(what is displayed in the actual text field), you'll need to set the `displayWith` property on your autocomplete 
+element. A common use case for this might be if you want to save your data as an object, but display just one of 
+the option's string properties.
+
+To make this work, create a function on your component class that maps the control value to the desired display value. 
+Then bind it to the autocomplete's `displayWith` property. You'll probably also want to add a step to your
```
jelbourn: `You probably also want to` => `You can`

src/lib/autocomplete/OVERVIEW.md
```diff
@@ -0,0 +1,149 @@
...
+         .map(val => this.displayFn(val))
+         .map(name => name ? this.filter(name) : this.options.slice());
+   }
+   
+   filter(val: string): any[] {
+      return this.options.filter(option => option.match(new RegExp(val, 'gi'))); 
+   }
+   
+   displayFn(value: any): string {
+      return value && typeof value === 'object' ? value.name : value;
```
jelbourn: Why is `typeof` needed here?

src/lib/autocomplete/OVERVIEW.md
```diff
@@ -0,0 +1,149 @@
...
+
+   ngOnInit() { 
+      this.filteredOptions = this.myControl.valueChanges
+         .startWith(null)
+         .map(val => this.displayFn(val))
+         .map(name => name ? this.filter(name) : this.options.slice());
+   }
+   
+   filter(val: string): any[] {
+      return this.options.filter(option => option.match(new RegExp(val, 'gi')));
```
jelbourn: You typically don't want to use `match` when you are looking for a yes/no result. 
http://stackoverflow.com/a/10940138/477102

kara:
@jelbourn Comments addressed

jelbourn:
LGTM

PR #2846, opened 2017-01-28
feat(theming): log a warning if core theme isn't loaded

### LINKED ISSUE #2828: Log a warning when a theme is not loaded
Not having a theme loaded is the most common issue that people seem to encounter. We should find a way to throw a dev-mode warning when we detect that a theme isn't loaded, [similar to what we do for HammerJS not being loaded](https://github.com/angular/material2/blob/e37259637a97aff0b94d2cdd10b892820a21a309/src/lib/core/gestures/gesture-config.ts#L23).

Ideally we would avoid using `getComputedStyle` or anything else expensive. We can add whatever simple marker style to the `md-core` mixin necessary if that helps.

crisbeto:
Checks the user's loaded stylesheets and logs a warning if the Material core theme isn't loaded.

Fixes #2828.

Note: I originally went with looping through the `document.styleSheets` to check whether the selector is defined, however I had to switch back to `getComputedStyle`, because browsers don't expose the `document.styleSheets`, if the CSS file is being loaded from another domain. This would've caused the warning to be logged if the user loads over a CDN.

### DIFF
--- a/src/lib/core/theming/theme-check.ts
+++ b/src/lib/core/theming/theme-check.ts
@@ -1,39 +1,40 @@
 import {NgModule, isDevMode} from '@angular/core';
 
+/** Whether the theme presence has already been checked. */
+let hasBeenChecked = false;
+
 /**
  * Module that verifies that the user has loaded the core theming file,
  * without which most Material module won't work as expected.
+ *
+ * Note on testing methodology: A more efficient way to check for the theme
+ * would be to loop through the `document.styleSheets`, however most browsers
+ * don't expose the stylesheet rules, if the file was loaded from another domain.
+ * This would trigger false positives if the theme is being loaded from a CDN.
+ *
  * @docs-private
  */
 @NgModule()
 export class MdThemeCheckModule {
   constructor() {
-    if (!isDevMode() || typeof document === 'undefined') {
+    if (hasBeenChecked || typeof document === 'undefined' || !isDevMode()) {
       return;
     }
 
-    for (let i = 0; i < document.styleSheets.length; i++) {
-      // The try/catch is needed, because some browsers can throw a security
-      // error when accessing the `cssRules` from another domain.
-      try {
-        let rules = (document.styleSheets.item(i) as CSSStyleSheet).cssRules;
+    let testElement = document.createElement('div');
 
-        if (rules) {
-          for (let j = 0; j < rules.length; j++) {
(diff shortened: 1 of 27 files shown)


src/lib/core/theming/theme-check.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (hasBeenChecked || typeof document === 'undefined' || !isDevMode()) {
+      return;
+    }
+
+    let testElement = document.createElement('div');
+
+    testElement.classList.add('md-theme-loaded-marker');
+    document.body.appendChild(testElement);
+
+    if (getComputedStyle(testElement).display !== 'none') {
```
jelbourn: What if instead we did something like
```scss
.mat-theme-load-test {
  width: 500px;
}
```

Then we create two elements (position absolute, min heights/width 1px), append them to body, and add the class to one of them then see if they have the same/different offsetWidth?

src/lib/core/theming/theme-check.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Note on testing methodology: A more efficient way to check for the theme
+ * would be to loop through the `document.styleSheets`, however most browsers
+ * don't expose the stylesheet rules, if the file was loaded from another domain.
+ * This would trigger false positives if the theme is being loaded from a CDN.
+ *
+ * @docs-private
+ */
+@NgModule()
+export class MdThemeCheckModule {
```
jelbourn: What if we roll this functionality into `CompatibilityModule` so we don't have to add another module on every components?

crisbeto: My reasoning for not adding it there is that, as far as I understand, that won't be a permanent module, but rather something that's there for a while until people have migrated completely to M2.

src/lib/core/theming/theme-check.ts
crisbeto: I think that it may end up being slower. Both `getComputedStyle` and `offsetWidth` cause a reflow, however we only need to call `getComputedStyle` once. As an aside, I tried doing some quick measurements between my initial approach (looping through the `document.styleSheets`) and the `getComputedStyle` approach. Both ended up taking about the same time on Chrome, but the latter was a bit faster in IE.

crisbeto:
Closing this one and will re-submit something simpler that will do the check in the `CompatibilityModule`.

PR #2848, opened 2017-01-28
chore: include web animations polyfill in demo app

crisbeto:
Currently some of the Material animations (the ones that user the Angular animation API, e.g. tooltip, select) don't work on certain browser, because they depend on the Web Animations API. This change adds a polyfill, in order to have a more consistent debugging experience across browsers.


jelbourn:
@matsko can you take a quick look?

PR #2849, opened 2017-01-28
fix(core): log warning if doctype is missing

### LINKED ISSUE #2351: MdDialog not resizing height to contents on open

Bug

trying to create an MdDialog using the Material 2 library. Expect the dialog to automatically size to the contents of the dialog. 

It's all working perfectly except for the height which comes out at 100% of the browser window height. 

If I set the height using a MdDialogConfig object then the height works properly. However this is not much good as if the form layout changes, the height will not adjust automatically thus running the risk of form elements being cut off. 

    export class MainHeader implements OnInit {
        constructor(private dialog: MdDialog,
                    private viewContainerRef: ViewContainerRef) {
    
        }
    
        ngOnInit() {
            this.config = new MdDialogConfig();
            this.config.viewContainerRef = this.viewContainerRef;
        }
    
        openLoginModal() {
            this.dialogRef = this.dialog.open(LoginModal, this.config);
    
            this.dialogRef.afterClosed().subscribe((result: string) => {
                // do something 
                this.dialogRef = null;
            });
        }
    
    }

If I don't pass in the config, same result. 

Using forms inside dialog.

Angular 4.0.0-beta.1
Material 2.0.0-alpha.11.3

I thought this might be a stylesheet problem so I commented out the styles and the problem persisted. 

To double check, I added a border to a `<div>` which surrounds the form. The form is the correct size as expected.

crisbeto:
Logs a warning if the user's document doesn't have a doctype. This has been an issue in the past with some hard-to-track-down bugs which ended up being due to the browser running in quirks mode.

Fixes #2351.

### DIFF
--- a/src/lib/core/compatibility/compatibility.ts
+++ b/src/lib/core/compatibility/compatibility.ts
@@ -5,7 +5,9 @@ import {
   OpaqueToken,
   Inject,
   Optional,
+  isDevMode,
 } from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
 
 
 export const MATERIAL_COMPATIBILITY_MODE = new OpaqueToken('md-compatibility-mode');
@@ -71,7 +73,7 @@ export const MAT_ELEMENTS_SELECTOR = `
   mat-toolbar`;
 
 /** Selector that matches all elements that may have style collisions with AngularJS Material. */
-export const MD_ELEMENTS_SELECTOR = `  
+export const MD_ELEMENTS_SELECTOR = `
   [md-button],
   [md-dialog-actions],
   [md-dialog-close],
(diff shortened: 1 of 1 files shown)


jelbourn:
(I made that comment like 6 weeks ago and forgot to hit "Send review")

willshowell:
Considering other issues have been solved by including the doctype (#3246, #3196), shouldn't the warning extend to more components than just `MdDialog`?

jelbourn:
That's a good point, we probably should do it somewhere more general. @crisbeto any thoughts on that?

crisbeto:
Sounds good to me, although this is some pretty basic stuff that people should be doing already.

crisbeto:
@jelbourn I've made it more generic, can you take another look?

src/lib/core/compatibility/compatibility.ts
```diff
@@ -167,6 +169,15 @@ export class CompatibilityModule {
       providers: [],
     };
   }
+
+  constructor(@Optional() @Inject(DOCUMENT) document: any) {
```
jelbourn: Type of `document` is just `Document`

crisbeto: I know, but my worry was that it would complain about it in a Universal app.

jelbourn: Good point, I didn't realize `Document` was a browser global

PR #2825, opened 2017-01-26
feat(dialog): add enter/exit animations

crisbeto:
* Adds enter/exit animations to the dialog.
* Refactors the `MdDialogContainer` and `MdDialogRef` to accommodate the animations.
* Fixes some test failures due to the animations.
* Allows for the backdrop to be detached before the rest of the overlay, in order to allow for it to be transitioned in parallel.

Fixes #2665.

PR #2853, opened 2017-01-28
feat(dialog): add ability to open dialog with templateRef

fxck:
I split https://github.com/angular/material2/pull/2851, used `TemplateRef` instead of `TemplatePortal`, and committed only the part that adds `openFromTemplateRef()`.

Added tests for `backdropClicked`, `escapePressed`(I explained the need for these in the previous PR) and a test for creating a dialog using templateRef. Not sure which other tests would be needed, I think the rest is identical to opening a dialog with `open()`.

Then there's a little "problem" that components inside the `<template>` do not have access to `MdDialogRef`, so `dialogRef.close()` and `md-dialog-close` obviously do not work, but if you are opening a dialog using `TemplateRef`, you likely do not need those anyway.

@jelbourn @crisbeto @DevVersion

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -6,7 +6,6 @@ import {
   NgZone,
   OnDestroy,
   ViewContainerRef,
-  TemplateRef,
 } from '@angular/core';
 import {BasePortalHost, ComponentPortal, PortalHostDirective, TemplatePortal} from '../core';
 import {MdDialogConfig} from './dialog-config';
@@ -48,8 +47,13 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
   /** Reference to the open dialog. */
   dialogRef: MdDialogRef<any>;
 
+  /** Exposes ViewContainerRef */
+  containerRef: ViewContainerRef;
+
   constructor(private _viewContainerRef: ViewContainerRef, private _ngZone: NgZone) {
     super();
+
+    this.containerRef = this._viewContainerRef;
   }
 
   /**
@@ -80,12 +84,6 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
     return attachResult;
   }
 
-  attachTemplateRef<T>(templateRef: TemplateRef<T>) {
-    const portal = new TemplatePortal(templateRef, this._viewContainerRef);
-
-    this.attachTemplatePortal(portal);
-  }
-
   /**
    * Handles the user pressing the Escape key.
    * @docs-private
@@ -95,7 +93,7 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
       this.dialogRef.close();
     }
 
-    this.dialogRef.escapePressed.next();
+    this.dialogRef.escapePress.next();
   }
 
   ngOnDestroy() {
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -5,6 +5,8 @@ import {
   ViewEncapsulation,
   NgZone,
   OnDestroy,
+  ViewContainerRef,
+  TemplateRef,
 } from '@angular/core';
 import {BasePortalHost, ComponentPortal, PortalHostDirective, TemplatePortal} from '../core';
 import {MdDialogConfig} from './dialog-config';
@@ -46,7 +48,7 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
   /** Reference to the open dialog. */
   dialogRef: MdDialogRef<any>;
 
-  constructor(private _ngZone: NgZone) {
+  constructor(private _viewContainerRef: ViewContainerRef, private _ngZone: NgZone) {
     super();
   }
 
@@ -61,20 +63,27 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
 
     let attachResult = this._portalHost.attachComponentPortal(portal);
 
-    // If were to attempt to focus immediately, then the content of the dialog would not yet be
-    // ready in instances where change detection has to run first. To deal with this, we simply
-    // wait for the microtask queue to be empty.
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-      this._elementFocusedBeforeDialogWasOpened = document.activeElement;
-      this._focusTrap.focusFirstTabbableElement();
-    });
+    this._focusFirstTabbableElement();
 
     return attachResult;
   }
 
-  /** @docs-private */
   attachTemplatePortal(portal: TemplatePortal): Map<string, any> {
-    throw Error('Not yet implemented');
+    if (this._portalHost.hasAttached()) {
+      throw new MdDialogContentAlreadyAttachedError();
+    }
+
+    let attachResult = this._portalHost.attachTemplatePortal(portal);
+
+    this._focusFirstTabbableElement();
+
+    return attachResult;
+  }
+
+  attachTemplateRef<T>(templateRef: TemplateRef<T>) {
+    const portal = new TemplatePortal(templateRef, this._viewContainerRef);
+
+    this.attachTemplatePortal(portal);
   }
 
   /**
@@ -85,6 +94,8 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
     if (!this.dialogConfig.disableClose) {
       this.dialogRef.close();
     }
+
+    this.dialogRef.escapePressed.next();
   }
 
   ngOnDestroy() {
@@ -93,6 +104,16 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
     // that it doesn't end up back on the <body>.
     this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
       (this._elementFocusedBeforeDialogWasOpened as HTMLElement).focus();
+    });
+  }
+
+  private _focusFirstTabbableElement() {
+    // If were to attempt to focus immediately, then the content of the dialog would not yet be
+    // ready in instances where change detection has to run first. To deal with this, we simply
+    // wait for the microtask queue to be empty.
+    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
+      this._elementFocusedBeforeDialogWasOpened = document.activeElement;
+      this._focusTrap.focusFirstTabbableElement();
     });
   }
 }
--- a/src/lib/dialog/dialog-ref.ts
+++ b/src/lib/dialog/dialog-ref.ts
@@ -15,16 +15,16 @@ export class MdDialogRef<T> {
   componentInstance: T;
 
   /** Expose overlay backdrop click event */
-  backdropClicked: Observable<void>;
+  backdropClick: Observable<void>;
 
   /** Subject for notifying the user that esc key way pressed */
-  escapePressed = new Subject<void>();
+  escapePress = new Subject<void>();
 
   /** Subject for notifying the user that the dialog has finished closing. */
   private _afterClosed: Subject<any> = new Subject();
 
   constructor(private _overlayRef: OverlayRef) {
-    this.backdropClicked = this._overlayRef.backdropClick();
+    this.backdropClick = this._overlayRef.backdropClick();
   }
 
   /**
--- a/src/lib/dialog/dialog-ref.ts
+++ b/src/lib/dialog/dialog-ref.ts
@@ -14,10 +14,18 @@ export class MdDialogRef<T> {
   /** The instance of component opened into the dialog. */
   componentInstance: T;
 
+  /** Expose overlay backdrop click event */
+  backdropClicked: Observable<void>;
+
+  /** Subject for notifying the user that esc key way pressed */
+  escapePressed = new Subject<void>();
+
   /** Subject for notifying the user that the dialog has finished closing. */
   private _afterClosed: Subject<any> = new Subject();
 
-  constructor(private _overlayRef: OverlayRef) { }
+  constructor(private _overlayRef: OverlayRef) {
+    this.backdropClicked = this._overlayRef.backdropClick();
+  }
 
   /**
    * Close the dialog.
--- a/src/lib/dialog/dialog.spec.ts
+++ b/src/lib/dialog/dialog.spec.ts
@@ -8,7 +8,15 @@ import {
   tick,
 } from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
-import {NgModule, Component, Directive, ViewChild, ViewContainerRef, Injector} from '@angular/core';
+import {
+  NgModule,
+  Component,
+  Directive,
+  ViewChild,
+  ViewContainerRef,
+  Injector,
+  TemplateRef
+} from '@angular/core';
 import {MdDialogModule} from './index';
 import {MdDialog} from './dialog';
 import {OverlayContainer} from '../core';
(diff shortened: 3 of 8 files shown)


fxck:
Fixed tests and added separated demo dialog components because of the fact that dialog opened with `TemplateRef` do not have `MdDialogRef`.

src/lib/dialog/dialog-container.ts
```diff
@@ -61,20 +63,27 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
...
+    }
+
+    let attachResult = this._portalHost.attachTemplatePortal(portal);
+
+    this._focusFirstTabbableElement();
+
+    return attachResult;
+  }
+
+  attachTemplateRef<T>(templateRef: TemplateRef<T>) {
```
jelbourn: The `TemplateRef` should be instantiated in `dialog.ts` and we should just call `attach` on the dialog container (from `BasePortalHost`) that will delegate to the appropriate method.

src/lib/dialog/dialog-ref.ts
```diff
@@ -14,10 +14,18 @@ export class MdDialogRef<T> {
   /** The instance of component opened into the dialog. */
   componentInstance: T;
 
+  /** Expose overlay backdrop click event */
+  backdropClicked: Observable<void>;
+
+  /** Subject for notifying the user that esc key way pressed */
+  escapePressed = new Subject<void>();
```
jelbourn: FYI: the `Subject` shouldn't ever be a public property, only the `Observable`

src/lib/dialog/dialog-ref.ts
```diff
@@ -14,10 +14,18 @@ export class MdDialogRef<T> {
   /** The instance of component opened into the dialog. */
   componentInstance: T;
 
+  /** Expose overlay backdrop click event */
+  backdropClicked: Observable<void>;
```
jelbourn: Should be `backdropClick` (to match `OverlayRef`). 

I don't think there needs to be an API for `escapePressed`. Users should be able to add their own `keydown` or `keypress` handler to their dialog content.

src/lib/dialog/dialog.ts
```diff
@@ -52,6 +48,34 @@ export class MdDialog {
       @Optional() @SkipSelf() private _parentDialog: MdDialog) { }
 
   /**
+   * Opens a modal dialog containing the given TemplateRef.
+   * @param TemplateRef to load into the dialog
+   * @param config Extra configuration options.
+   * @returns Reference to the newly-opened dialog.
+   */
+  openFromTemplateRef<T>(templateRef: TemplateRef<T>, config?: MdDialogConfig): MdDialogRef<T> {
```
jelbourn: Just `openFromTemplate`

src/lib/dialog/dialog.ts
```diff
@@ -52,6 +48,34 @@ export class MdDialog {
...
+    config = _applyConfigDefaults(config);
+
+    let overlayRef = this._createOverlay(config);
```
jelbourn: Rather than duplicating the control flow, `open` and `openFromTemplate` should just instantiate the appropriate `Portal` and then pass that through. One of the benefits of portals is that the host doesn't have to care which kind it is.

src/lib/dialog/dialog.ts
jelbourn: I think this method should lose the generic and return `MdDialogRef<void>`. since the generic refers to the type of `componentInstance`.

src/lib/dialog/dialog-container.ts
fxck: I'm not sure I understand this correctly, but to be able to create a portal, you need ViewContainerRef, which is not accessible inside `dialog.ts`. In my previous PR I had `MdDialog.openFromPortal()`, but I think it a little too tedious for user to have to create a portal rather than to be just able to pass in `TemplateRef`. But I don't really mind either.

src/lib/dialog/dialog-ref.ts
fxck: That's what I thought first as well, but then there is `afterClosed` instead of `afterClose`, but I was not sure.

As for escape, there is `handleEscapeKey()` on `dialog-container.ts` already, so I think it's fair to expose that.

src/lib/dialog/dialog-ref.ts
fxck: `asObservable()` then?

src/lib/dialog/dialog.ts
fxck: Similar problem as with https://github.com/angular/material2/pull/2853#discussion_r98502685 I don't think I can do that unless I have ViewContainerRef, which I don't unless I require the user to pass me `TemplatePortal`..

src/lib/dialog/dialog.ts
fxck: Yeah, I was not sure about what to do with that(https://github.com/angular/material2/pull/2851 read 2.).

src/lib/dialog/dialog-ref.ts
crisbeto: I'm not sure that we need observables for `escapePressed`, `backdropClicked` and `afterClosed` since the API is already becoming pretty dense. I'd rather we kept everything to `afterClosed` and emitted an object that looks something like `{ result: <user-provided value>, reason: 'backdrop' | 'keypress' | 'api' }`.

src/lib/dialog/dialog.ts
crisbeto: Not sure I get the issue here. Isn't it down to doing a `ref instanceof TemplateRef` check and then taking the proper route under the hood?

src/lib/dialog/dialog.ts
```diff
@@ -52,6 +48,34 @@ export class MdDialog {
...
+
+    let dialogContainer = this._attachDialogContainer(overlayRef, config);
+    let dialogRef = this._attachDialogContent(
+      dialogContainer,
+      overlayRef,
+      undefined,
```
crisbeto: Probably better to pass in `null` here instead. That way we know that it was skipped on purpose.

src/lib/dialog/dialog.ts
```diff
@@ -143,10 +174,16 @@ export class MdDialog {
     let userInjector = config && config.viewContainerRef && config.viewContainerRef.injector;
     let dialogInjector = new DialogInjector(dialogRef, userInjector || this._injector);
 
-    let contentPortal = new ComponentPortal(component, null, dialogInjector);
+    if (component) {
```
crisbeto: I think that the logic that is specific to `TemplateRef` and `ComponentType` should be split into separate private methods.

src/lib/dialog/dialog.spec.ts
```diff
@@ -150,6 +196,25 @@ describe('MdDialog', () => {
...
+  it('should notify the user that overlay backdrop was clicked', () => {
+    let dialogRef = dialog.open(PizzaMsg, {
+      viewContainerRef: testViewContainerRef
+    });
+
+    viewContainerFixture.detectChanges();
+
+    let backdrop = overlayContainerElement.querySelector('.cdk-overlay-backdrop') as HTMLElement;
+
+    let clicked = false;
```
crisbeto: Instead of keeping a flag like this, you should use a spy:

```ts
let clickedSpy = jasmine.createSpy('clicked callback');

dialogRef.backdropClicked.subscribe(clickedSpy);

expect(clickedSpy).toHaveBeenCalled();
```

src/lib/dialog/dialog.spec.ts
```diff
@@ -137,6 +161,28 @@ describe('MdDialog', () => {
...
+    let dialogRef = dialog.open(PizzaMsg, {
+      viewContainerRef: testViewContainerRef
+    });
+
+    viewContainerFixture.detectChanges();
+
+    let dialogContainer: MdDialogContainer =
+        viewContainerFixture.debugElement.query(By.directive(MdDialogContainer)).componentInstance;
+
+    let pressed = false;
```
crisbeto: This should use a spy as well. See https://github.com/angular/material2/pull/2853/files#r98516864

src/lib/dialog/dialog-ref.ts
fxck: That wouldn't trigger if escape/backdrop click didn't actually close the dialog. Which in result wouldn't allow for creating a stateless dialog component(which is honestly a must have in redux and ngrx/store world). Or `afterClosed` could be modified to trigger and not close the dialog when `disableClose` is true, but then the name wouldn't make much sense.

src/lib/dialog/dialog.ts
fxck: https://medium.com/@basarat/null-vs-undefined-in-typescript-land-dc0c7a5f240a#.ug5uqgtus but sure, I guess rest of the lib is not following this anyway

src/lib/dialog/dialog-ref.ts
crisbeto: You could differentiate it based on the source. My worry is that we're ending up with 5-6 different dialog-related observables. If we added these, we'd have 3 just for the `close` method.

fxck: What could I differentiate? The problem is that `afterClosed()` triggers only when `close` is called, which actually [closes](https://github.com/angular/material2/blob/master/src/lib/dialog/dialog-ref.ts#L28) the dialog . I need to be able to catch any close **attempt** while `disableClose` is `true` and let user know about it, so he can modify the state himself. Would you be happier is backdrop and escape were merged into one observable, say `closeAttempt`?

crisbeto: My question is why does the user need to know that an attempt to close it was made?

fxck: Check https://github.com/angular/material2/pull/2855

Imagine you keep state of your modal in a global store, like redux or @ngrx/store, so you create component that looks like this.

```html
<my-dialog [open]="modalState$ | async">
  <button (click)="doSomethingThatChangesModalState$toFalse()">close</button>
</my-dialog>
```

Now you are in full control of the state, when you change `modalState$`, your modal closes, BUT when you click on backdrop or press escape, your modal closes while your `modalState$` is still true and your store is out of sync. You could catch up by changing `modalState$` to `false` on `afterClose()`, but that could for example be called only after the animation finishes, it simply smells, it's the other way around than it should be. So you want to disable the default closing behaviour of `backdropClick` and `escape`, so you do `disableClose: true`, but then you have to manually set up listeners for both, which will be big pain in the ass.


So ideally you want to do

```html
<my-dialog 
    [open]="modalState$ | async" 
    (close)="doSomethingThatChangesModalState$toFalse()">
  <button (click)="doSomethingThatChangesModalState$toFalse()">close</button>
</my-dialog>
```

where close event happens on either `backdropClick` or `escape` press https://github.com/fxck/material2/blob/662703b63831797c9969dda4dd03d69d794aa1f4/src/lib/dialog/dialog-element.ts#L54-L61

src/lib/dialog/dialog-container.ts
fxck: I see, I could make `ViewTemplateRef` available to `dialog.ts`.

fxck: What about types thought, wouldn't you lose them by just calling `attach` instead of `attachComponentPortal`/`attachTemplatePortal`?

fxck: Also if I just use `attach`, typescript will complain about not properly implementing `attachTemplatePortal` and `attachComponentPortal`.. what should I do please, @jelbourn ?

src/lib/dialog/dialog-ref.ts
fxck: The problem is that I need to `next` it from inside `dialog-container.ts`, if I make it private I won't be able to access it.

fxck:
Addressed couple of comments:

- use spy(https://github.com/angular/material2/pull/2853#discussion_r98516966, https://github.com/angular/material2/pull/2853#discussion_r98516864)
- split logic specific to template/component to private methods https://github.com/angular/material2/pull/2853#discussion_r98516478
- use `null` instead of `undefined` https://github.com/angular/material2/pull/2853#discussion_r98516160
- remove generics https://github.com/angular/material2/pull/2853#discussion_r98499869 
- create TemplatePortal in dialog.ts https://github.com/angular/material2/pull/2853#discussion_r98498993
- rename method to openFromTemplate https://github.com/angular/material2/pull/2853#discussion_r98491525
- rename clicked / pressed to click / press https://github.com/angular/material2/pull/2853#discussion_r98490671


Didn't address these:

- remove backdropClicked / escapePressed, reasons explained in https://github.com/angular/material2/pull/2853#discussion_r98515217 
    - what I could do here is to merge these two into one
- don't expose subject https://github.com/angular/material2/pull/2853#discussion_r98490281
    - I cannot make it private, since I need to call it from inside `dialog-container.ts`.. not sure what to do here, create a public function? that will still make it available to consumers
- use `attach` instead of `attachComponentPortal` and `attachTemplatePortal` https://github.com/angular/material2/pull/2853#discussion_r98489471
    - not sure what to do here, by using attach I would lose types and typescript will complain, because I won't be properly implementing `BasePortalHost`


@jelbourn @crisbeto

src/lib/dialog/dialog-container.ts
```diff
@@ -46,8 +47,13 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
...
   dialogRef: MdDialogRef<any>;
 
-  constructor(private _ngZone: NgZone) {
+  /** Exposes ViewContainerRef */
+  containerRef: ViewContainerRef;
+
+  constructor(private _viewContainerRef: ViewContainerRef, private _ngZone: NgZone) {
     super();
+
+    this.containerRef = this._viewContainerRef;
```
fxck: had to do this, when I did just 

```
constructor(viewContainerRef: ViewContainerRef, private _ngZone: NgZone) {}
```

this https://github.com/fxck/material2/blob/9a0a28be6b551302363577462865fac760ad09a3/src/lib/dialog/dialog.ts#L212 was complaining that `viewContainerRef` doesn't exist

src/lib/dialog/dialog-ref.ts
crisbeto: Alright, it makes sense. I still think that they should be combined into something along the lines of `userGeneratedClose`.

crisbeto:
@fxck regarding your comment about not being able to expose the `escapePressed` subject, I'm not sure that the `DialogRef` should be the one that knows about what DOM event are happening in the `DialogContainer`. In https://github.com/angular/material2/pull/2825 I changed the relationship from it being `container->ref` to `ref-container`, which means that you could keep the subject in the container and do a `dialogRef.containerInstance.escapePressed`. Note that the `containerInstance` is private in https://github.com/angular/material2/pull/2825, but that can be refactored if necessary.

fxck:
I could probably base my PR on yours if you want. I actually was wondering about another thing regarding your PR, it currently doesn't take in account from what button was the dialog opened(like material1 does), but once that's there, I'm not sure how's that gonna work with element like this https://github.com/angular/material2/pull/2855

crisbeto:
There isn't anything in the spec that says that it's supposed to do that, and I couldn't find anybody that remembered why it was done that way in M1. The only reference to dialog animations in the spec is [this](https://storage.googleapis.com/material-design/publish/material_v_10/assets/0B14F_FSUCc01Zmh1anRVcFBLblk/CreationChoreo_06_NewDialogueDo_v2.webm) upon which I based my PR. After looking through different implementations (Google Docs, Google Keep, Inbox, Android) it's pretty inconsistent.

fxck:
I think it might have something to do with one of these https://storage.googleapis.com/material-design/publish/material_v_10/assets/0B14F_FSUCc01dFhwekx5R3BsNGM/Radial_04_RadialArc_v5-remapped.mp4 .. it arguably looks pretty good in m1.. 

Anyway how should I proceed with this PR then? I'll do anything you want/need as long as it allows me to create that stateless dialog component/directive. I would really love to get this in, it's been bothering me for a long long time.

crisbeto:
I would say to go with the assumption from #2825 and if we decide to go the M1 route, I think that it's not too complicated to refactor.

fxck:
I guess I'll wait until that gets merged then.

Issue #2552, opened 2017-01-06
MdDialog passing in data and or variables to modal 


xtianus79:
#### Bug, feature request, or proposal:
Passing things into the second component seems obscure.  I can't readily pass in values to the second component to display on modal popup. 

#### What is the expected behavior?
When the modal popup is displayed... information such as variables will display accordingly {{data.here}}

#### What is the current behavior?
Not working 

#### What are the steps to reproduce?
pass in variables to the new value 
Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: http://plnkr.co/edit/o077B6uEiiIgkC0S06dd
https://plnkr.co/edit/dhZ4UldIlQKDPvrjJP2f?p=preview  <<< this is example from Material

        export class DialogResultExampleDialog {
            title: string = 'book';
          constructor(public dialogRef: MdDialogRef<DialogResultExampleDialog>) {}

          public testFunction(errorMsg: string) {
            console.log('this is a test999999 ' + errorMsg);
          }

       }

#### What is the use-case or motivation for changing an existing behavior?
To be able to pass in data / variables to the modal dialog popup upon request. 

#### Which versions of Angular, Material, OS, browsers are affected?
Angular 2.4.1

antonyRoberts:
There is a pull request to make it easier 
[feat(dialog): add a config option for passing in data](https://github.com/angular/material2/pull/2266)

In the mean time https://github.com/angular/material2/issues/1377#issuecomment-250632661

crisbeto:
@antonyRoberts is right. I'll close this issue as a duplicate of #2181.

xtianus79:
Hi @crisbeto and @antonyRoberts ... I have been working through it and I notice that the way the example API shows how to do the diaglog there is a 2 component system.  The issue with that is accessing the class to fire the modal popup is probably the same class that has the data you need to access too... not to say there aren't ways to get around this but the point is still a concern.  So what I did is pointed it to the same class. 

This works per se but now I can't set the data.  

For example... 

    title: string = 'book';  i

this is fine but what if I want to access data... i.e. 

    title: any = this.someFunction(dataPassedIn);

^^^^^ I can't do this.  perhaps I am doing something wrong but any help would be appreciated.  

Should I post this in the other issue forum?  #2181

crisbeto:
Not sure I get the issue @xtianus79. Can you post what your classes look like?

xtianus79:
@crisbeto it is pretty straight forward... 

[...]

Issue #2086, opened 2016-12-06
Passing local data to md dialog???


dreweird:
----in my html code---
` <td *ngIf="item.clickable=='yes'" (click)="openDialog(item)">{{item.kilo[0] | number: '1.2-2'}}</td>`

---in my component code---

```
 openDialog(item) {
    let dialogRef = this._dialog.open(DialogContent);

    dialogRef.afterClosed().subscribe(result => {
      this.lastDialogResult = result;
    })
  }
```
How to show the local variables in the md-dialog template???? I want to display the value of item
```
@Component({
  template: `
    <p>This is a dialog</p>
    <p>
    {{item}}
      <label>
        This is a text box inside of a dialog.
        <input #dialogInput>
      </label>
    </p>
    <p> <button md-button (click)="dialogRef.close(dialogInput.value)">CLOSE</button> </p>
  `,
})
export class DialogContent {
  constructor(@Optional() public dialogRef: MdDialogRef<DialogContent>) { }
}
```

crisbeto:
[Try the suggestion from here.](https://github.com/angular/material2/issues/1377#issuecomment-250632661) We should probably document this option a little bit better.

PR #2855, opened 2017-01-29
feat(dialog): add md-dialog component

fxck:
Needs #2853.

`md-dialog` is a stateless component that's gonna be handy for people that use global state management like redux or @ngrx/store. It's also nice when you want to use standard input/output to pass data to the component inside your dialog(#2552, #2086, #2181 related).

```typescript
<md-dialog [open]="loginDialogState">
  <login-form [data]="loginDefaultData" (submit)="loginSubmitted($event)"></login-form>
</md-dialog>
```

Internally it uses `openFromTemplateRef()` which is added by #2853, it disables closing by backdrop click and escape key by default(to be truly stateless), instead emits `close` event with event data containing what caused it(`escape`, `backdrop`).

It's got similar api to that of https://developer.mozilla.org/en/docs/Web/HTML/Element/dialog

- `[open]` controls whether dialog is shown or not
- `(close)` emits event when either backdrop was clicked or escape was pressed
- `[config]` allows passing in `MdDialogConfig`

@jelbourn @crisbeto

### DIFF
--- a/src/demo-app/dialog/dialog-demo.html
+++ b/src/demo-app/dialog/dialog-demo.html
@@ -66,4 +66,4 @@
 
 
 <template #contentElementRef>
-  <demo-content-element-dialog [actionsAlignment]="actionsAlignment"></demo-content-element-dialog>
+  <demo-content-element-template-ref-dialog (close)="closeContentElementUsingTemplateRef()" [actionsAlignment]="actionsAlignment"></demo-content-element-template-ref-dialog>
--- a/src/demo-app/dialog/dialog-demo.html
+++ b/src/demo-app/dialog/dialog-demo.html
@@ -67,3 +67,4 @@
 
 <template #contentElementRef>
   <demo-content-element-template-ref-dialog (close)="closeContentElementUsingTemplateRef()" [actionsAlignment]="actionsAlignment"></demo-content-element-template-ref-dialog>
+</template>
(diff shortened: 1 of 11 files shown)


src/demo-app/dialog/dialog-demo.html
```diff
@@ -1,53 +1,107 @@
...
+    <button
+      md-raised-button
+      color="primary"
+      (click)="changeMdDialogState(true)"
+      [disabled]="mdDialogState">
+      Open md-dialog with content elements
+    </button>
+  </div>
+
+  <md-dialog
```
jelbourn: Not a full review for now, but I would do this as a directive that is meant to go on a template element, so it would be either of
```html
<template mdDialog ...>

or

<div *mdDialog ...>
```

Mainly so that it doesn't leave a phantom `md-dialog` element in the DOM

fxck: @jelbourn I think that would make using inputs/outputs impossible, wouldn't it?

ethanve:
Any news on this? Can I help in any way?

fxck:
@ethanve it's blocked on https://github.com/angular/material2/pull/3460 go bug @jelbourn to do a review

jelbourn:
Closing this since it's quite stale now. As of the recent #9379, `TemplateRef` dialogs should actually work all the way now, so the thing that was blocking this should be gone. Feel free to open a new PR if you want to take another go at it.

PR #2857, opened 2017-01-29
fix(checkbox): ripple color does not change

DevVersion:
* The ripple color didn't change on the checkbox component because commit 8ce65ca253edc4bf4a47ace970ef441c8b3debe9 accidentally introduced a wrong selector for the checkbox.


tinayuangao:
LGTM

Issue #1434, opened 2016-10-05
Ink ripple inconsistent with material design spec (?)


rosslavery:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Ink ripples in previous material design implementations trigger the effect on mousedown, rather than mouseup.

#### What is the current behavior?
The effect is only triggered on mousedown, which is far less noticable than prior implementations.

#### What are the steps to reproduce?
Material2 demo from [the PR](https://github.com/angular/material2/pull/918) that added ripples to buttons: https://dozingcat-material2.firebaseapp.com/button
Material1 demo: https://material.angularjs.org/latest/demo/button

Click the buttons to see the difference. Click and hold your mouse button to make the difference more obvious.

#### What is the use-case or motivation for changing an existing behavior?
Consistency with the spec.

#### Which versions of Angular, Material, OS, browsers are affected?
All?

jelbourn:
We worked directly with the Material Design spec folks on this one and their direction was that the ripple is supposed to trigger on the _down_ and not the _up_.

rosslavery:
Hi Jeremy,

I apologize, but I typoed in my original bug report. What I **meant** to say is that the ripples are only being triggered on mouseup, not down. Thus, they are behaving incorrectly according to your comment.

Sorry for the confusion, but check out the two demos and you'll see what I mean.

jelbourn:
Ah, I was just going based on what you wrote 😄 ; my recollection was that the Material Design person we consulted with gave direction that ours should be the way it is (I just didn't remember which was which). 

@traviskaufman will have the definitive answer for this

rosslavery:
Gotcha. Most implementations I've seen are consistent with the mousedown approach (Angular 1 Material, Polymer, etc.)

Interested in the official response though!

rosslavery:
Not to sound like a broken record, but that the ink ripples still don't appear to match the spec (using beta 10).

In fact, the [readme](https://github.com/angular/material2/blob/master/src/lib/core/ripple/README.md) for md-ripple actually specifies:

> By default, a ripple is activated when the host element of the md-ripple directive receives mouse or touch events. On a **mousedown** or touch start, the ripple background fades in.

PR #2859, opened 2017-01-29
fix(ripple): make ripples conform with specs

DevVersion:
This makes the ripple service conform with the specifications and other Material reference implementations.

See https://material.io/guidelines/motion/material-motion.html#material-motion-how-does-material-move

_Reference from the Material specifications_
> Ink ripples confirm user input by immediately expanding outward from the point of touch. The card lifts to indicate an active state.

This means the following:

 * Ripples now trigger on `mousedown`
 * Ripples now persists as long as the element is being hold.
 * No longer adds an unnecessary background ripple.
 * Removes the ugly `scale(0.00001)` for ripple animations

Not only visually the ripple has been changed. The whole renderer has been rewritten and now has a very simple API, that can be easily used by developers.

Also the API has been changed to something more intuitive

```ts
// (pageX, pageY, config)
ripple.createRipple(0, 0, {
  centered: true, 
  speedFactor: 0.1
});
```

_As seen in the example, developers are now able to specify a simple Ripple Config object_

**Live Demo**: https://material2-dashboard.firebaseapp.com/button (temporary)

References #1434

### DIFF
--- a/src/lib/checkbox/_checkbox-theme.scss
+++ b/src/lib/checkbox/_checkbox-theme.scss
@@ -69,15 +69,15 @@
   }
 
   .mat-checkbox:not(.mat-checkbox-disabled) {
-    &.mat-primary .mat-checkbox-ripple .mat-ripple-foreground {
+    &.mat-primary .mat-checkbox-ripple .mat-ripple-element {
       background-color: mat-color($primary, 0.26);
     }
 
-    &.mat-accent .mat-checkbox-ripple .mat-ripple-foreground {
+    &.mat-accent .mat-checkbox-ripple .mat-ripple-element {
       background-color: mat-color($accent, 0.26);
     }
 
-    &.mat-warn .mat-checkbox-ripple .mat-ripple-foreground {
+    &.mat-warn .mat-checkbox-ripple .mat-ripple-element {
       background-color: mat-color($warn, 0.26);
     }
   }
--- a/src/lib/checkbox/checkbox.html
+++ b/src/lib/checkbox/checkbox.html
@@ -18,8 +18,7 @@
     <div md-ripple *ngIf="!_isRippleDisabled()" class="mat-checkbox-ripple"
          [mdRippleTrigger]="_getHostElement()"
          [mdRippleCentered]="true"
-         [mdRippleSpeedFactor]="0.3"
-         mdRippleBackgroundColor="rgba(0, 0, 0, 0)"></div>
+         [mdRippleSpeedFactor]="0.3"></div>
     <div class="mat-checkbox-frame"></div>
     <div class="mat-checkbox-background">
       <svg version="1.1"
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -93,15 +93,17 @@ export class RippleRenderer {
 
     // If the color is not set, the default CSS color will be used.
     ripple.style.backgroundColor = config.color;
+    ripple.style.transitionDuration = `${duration}s`;
 
     this._targetElement.appendChild(ripple);
 
-    // Wait for the next tick because for elements the CSS transition isn't applied
-    // immediately and otherwise the ripple would just expand without any animation.
-    this.runTimeoutOutsideZone(() => {
-      ripple.style.transitionDuration = `${duration}s`;
-      ripple.style.transform = 'scale(1)';
-    }, 1);
+    // By default the browser does not recalculate the styles of dynamically created
+    // ripple elements. This is critical because then the `scale` would not animate properly.
+    // Enforce a style recalculation by calling `getComputedStyle` and accessing any property.
+    // See: https://gist.github.com/paulirish/5d52fb081b3570c81e3a
+    window.getComputedStyle(ripple).getPropertyValue('opacity');
+
+    ripple.style.transform = 'scale(1)';
 
     // Wait for the ripple to be faded in. Once faded in the ripple can be hided if the mouse is
     // released.
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -99,14 +99,12 @@ export class RippleRenderer {
 
     // By default the browser does not recalculate the styles of dynamically created
     // ripple elements. This is critical because then the `scale` would not animate properly.
-    // Enforce a style recalculation by calling `getComputedStyle` and accessing any property.
-    // See: https://gist.github.com/paulirish/5d52fb081b3570c81e3a
-    window.getComputedStyle(ripple).getPropertyValue('opacity');
+    this._enforceStyleRecalculation(ripple);
 
     ripple.style.transform = 'scale(1)';
 
-    // Wait for the ripple to be faded in. Once faded in the ripple can be hided if the mouse is
-    // released.
+    // Wait for the ripple to be faded in. Once it's faded in, the ripple can be hidden immediately
+    // if the mouse is released.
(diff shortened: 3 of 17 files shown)


fxck:
I'm not sure if material1 conforms spec properly, but it's still slightly different(at least on buttons) https://material.angularjs.org/latest/demo/button background color changes slightly immediately after mousedown.

DevVersion:
@fxck What you mean, should be the hover effect.. For now the target of this PR is only making the `ripples` conform with the specs (as specified [here](https://material.io/guidelines/motion/material-motion.html#material-motion-how-does-material-move))

Updating the hover effect of the buttons and other components can be a follow-up.

fxck:
Nonono, it's not the hover effect(it happens after mousedown and I think it's part of the ripple. Want me to make a slowed down video? :)

fxck:
default state
image

hover state(no change)
image

mouse down(notice the grey background)
image

full ripple
image

DevVersion:
That's because Material 1 sets an option to `dim` the background ([see](https://github.com/angular/material/blob/master/src/core/services/ripple/button_ripple.js#L39)). This isn't part of the specification as seen in the [specs](https://material.io/guidelines/motion/material-motion.html#material-motion-how-does-material-move) and also not a big deal either.

> Also add such questions in a PR comment so we can dismiss those. Now the PR is looking unclear...

fxck:
Ok, removed them.

DevVersion:
@fxck Thanks. I've edited my comment to include what you basically commented.

src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -40,167 +29,142 @@ const distanceToFurthestCorner = (x: number, y: number, rect: ClientRect) => {
...
+  /** Whether mouse ripples should be created or not. */
+  rippleDisabled: boolean = false;
+
+  constructor(_elementRef: ElementRef, private _ngZone: NgZone, private _ruler: ViewportRuler) {
+    this._targetElement = _elementRef.nativeElement;
+
+    // Specify events which need to be registered on the trigger.
+    this._triggerEvents.set('mousedown', this.onMousedown.bind(this));
+    this._triggerEvents.set('mouseup', this.onMouseup.bind(this));
+    this._triggerEvents.set('mouseleave', this.onMouseLeave.bind(this));
```
jelbourn: Rather than using a `Map`, how about an array of tuples? Better for iteration
```ts
let triggerEvents: [string, (MouseEvent) => void][] = [
    ['mousedown', e => this.onMouseDown(e)],
    // ...
];
```
(generally prefer arrow functions to `.bind()`)

DevVersion: Don't think that will be better for iteration. A map already provides a suitable iterator for those eventpairs. 

I would personally like to stay with the `Map` for convenience and readability.

DevVersion:
@jelbourn Addressed your feedback. Thanks.

src/lib/core/ripple/ripple.spec.ts
```diff
@@ -86,112 +65,86 @@ describe('MdRipple', () => {
     document.body.style.margin = originalBodyMargin;
   });
 
+  function dispatchMouseEvent(type: string, offsetX = 0, offsetY = 0) {
```
mmalerba: I swear this function is in like half of our test files, we should create a test utilities file sometime (not part of this PR though)

src/lib/radio/_radio-theme.scss
```diff
@@ -29,7 +29,7 @@
     }
   }
 
-  .md-radio-ripple .md-ripple-foreground {
+  .md-radio-ripple .md-ripple-element {
```
mmalerba: when adding new classes make them mat-* to make andrew's life easier :p

src/lib/checkbox/checkbox.html
```diff
@@ -18,8 +18,7 @@
     <div md-ripple *ngIf="!_isRippleDisabled()" class="md-checkbox-ripple"
          [mdRippleTrigger]="_getHostElement()"
          [mdRippleCentered]="true"
-         [mdRippleSpeedFactor]="0.3"
-         mdRippleBackgroundColor="rgba(0, 0, 0, 0)"></div>
+         [mdRippleSpeedFactor]="0.3"></div>
```
DevVersion: As discussed on Slack, we are going to have another PR to make all component ripples align with the specs.

src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -40,167 +29,142 @@ const distanceToFurthestCorner = (x: number, y: number, rect: ClientRect) => {
...
-  /** Installs event handlers on the host element of the md-ripple directive. */
-  setTriggerElementToHost() {
-    this.setTriggerElement(this._rippleElement);
-  }
+    let radius = config.radius || distanceToFurthestCorner(pageX, pageY, containerRect);
+    let duration = 1 / (config.speedFactor || 1) * (radius / RIPPLE_SPEED_PX_PER_SECOND);
+    let offsetX = pageX - containerRect.left;
+    let offsetY = pageY - containerRect.top;
+
+    let ripple = document.createElement('div');
```
DevVersion: As discussed on Slack, we will revisit this in issue #308 and once we have feedback from Jeremy about this.

src/lib/checkbox/_checkbox-theme.scss
```diff
@@ -68,16 +68,16 @@
     }
   }
 
-  .md-checkbox:not(.md-checkbox-disabled) {
-    &.md-primary .md-checkbox-ripple .md-ripple-foreground {
+  md-checkbox:not(.md-checkbox-disabled) {
```
DevVersion: Good point.  I actually had that change only in this PR for testing purposes. I have another PR open for this. See https://github.com/angular/material2/pull/2857

Also I think we can only use the class here, as soon as Andrew's PR is merged.

DevVersion:
@mmalerba Thanks for your feedback. Addressed the comments you added. Also glad you like it :)

src/lib/checkbox/checkbox.html
mmalerba: created #2901 for this

DevVersion: Should I assign it to me?

src/lib/core/ripple/ripple.spec.ts
mmalerba: created #2902 for this

mmalerba:
lgtm

src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -1,27 +1,9 @@
...
-export class ForegroundRipple {
-  state = ForegroundRippleState.NEW;
-  constructor(public rippleElement: Element) {}
-}
 
-const RIPPLE_SPEED_PX_PER_SECOND = 1000;
-const MIN_RIPPLE_FILL_TIME_SECONDS = 0.1;
-const MAX_RIPPLE_FILL_TIME_SECONDS = 0.3;
+export const RIPPLE_SPEED_PX_PER_SECOND = 170;
+export const RIPPLE_FADE_OUT_DURATION = 600;
```
tinayuangao: Add doc somewhere to mention the fade out duration is fixed and is not changed with speed factor?

DevVersion: Good point. Added it to the `@Input()` and also added comments for those constants.

DevVersion:
Just rebased the PR due to the compatibility changes.

realappie:
Great PR! Was about to open up an issue about the current state of ripple effects.

PR #2861, opened 2017-01-30
fix(input) : replace mdInput by md-input in readme file

adrienpessu:


YeomansIII:
In the latest master builds mdInput is actually required instead of md-input.

adrienpessu:
Ok my bad

PR #2870, opened 2017-01-30
feat(sidenav): open all sidenavs from MdSidenavContainer

### LINKED ISSUE #2591: Sidenav - provide means to open/close from non-child elements

Given usage like:
```html
<md-sidenav-container>
  <md-sidenav>
    <ul></ul>
  </md-sidenav>
  <my-component></my-component>
</md-sidenav-container>
```

Because `my-component` is not a child of the `md-sidenav` element, you can't inject MdSidenav into it, which makes programmatically toggling the Sidenav rather difficult (either you end up having to @ViewChild() query for it, and wiring that all up through your own service)

Developers should have an easy way to programmatically control the md-sidenav 

See above

Material 1 provided a `$sideNav` service, which is the kind of thing i'm describing here: https://material.angularjs.org/latest/api/service/$mdSidenav

EladBezalel:
- added `open` and `close` functions that applies to both sidenavs in the container and resolves when both are finished

fixes #2591

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -371,6 +371,16 @@ export class MdSidenavContainer implements AfterContentInit {
     this._ngZone.onMicrotaskEmpty.first().subscribe(() => this._enableTransitions = true);
   }
 
+  /** Calls `open` of both start and end sidenavs */
+  public open() {
+    return Promise.all([this._start, this._end].map(sidenav => sidenav && sidenav.open()));
+  }
+
+  /** Calls `close` of both start and end sidenavs */
+  public close() {
+    return Promise.all([this._start, this._end].map(sidenav => sidenav && sidenav.close()));
+  }
+
   /**
    * Subscribes to sidenav events in order to set a class on the main container element when the
    * sidenav is open and the backdrop is visible. This ensures any overflow on the container element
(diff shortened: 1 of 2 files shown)


src/lib/sidenav/sidenav.ts
```diff
@@ -367,6 +367,36 @@ export class MdSidenavContainer implements AfterContentInit {
...
+    if (this._start) {
+      promises.push(this._start.close());
+    }
+
+    if (this._end) {
+      promises.push(this._end.close());
+    }
+
+    return Promise.all(promises);
+  }
```
jelbourn: Unit test?

src/lib/sidenav/sidenav.ts
```diff
@@ -367,6 +367,36 @@ export class MdSidenavContainer implements AfterContentInit {
     this._validateDrawers();
   }
 
+  /** Calls `open` of both start and end sidenavs */
+  public open() {
+    const promises: Promise<MdSidenavToggleResult>[] = [];
```
mmalerba: can't this be simplified to:

```
return Promise.all([this._start, this._end].map((sidenav) => sidenav && sidenav.open()));
```

EladBezalel: Thanks that's much simpler

EladBezalel:
@mmalerba @kara please rereview :)

jelbourn:
@mmalerba is this merge-ready?

PR #2875, opened 2017-01-31
fix(dialog): leaking component instance references

### LINKED ISSUE #2734: MdDialog seems to leak the displayed components
Hi All,

I am using MdDialog to display some modal views defined by Angular 2 components. I observed that the instances of these components are still present in the browser snapshots after the dialog was closed. For example, I display a confirmation dialog for 10 times using a TwoOptions component. After that, the memory snapshot will contain 10 instances of the TwoOptions component.

No instance of the displayed components using MdDialog should remain.

The instances of the displayed component are still present, being retained by MdDialogContainer.

Open and close a MdDialog view specifying a custom component.

The following plunker provided by 
[Making use of Dialogs in Material 2 (MdDialog)](https://medium.com/@tarik.nzl/making-use-of-dialogs-in-material-2-mddialog-7533d27df41#.r2tkj2lxf)
 can be used to ilustrate the issue
http://plnkr.co/edit/6Eta3GhQLUJTsYz22wsL?p=preview

Press Open Dialog for few times, collect a snapshot in Chrome and check for the ConfirmDialog instances.
image

Angular 2.4.3, @angular/material: 2.0.0-beta.1, Windows 7, Chrome, Firefox

crisbeto:
Fixes a memory leak in the dialog, which was causing it to keep references to the user's instantiated dialog in memory forever. It was due to:
1. The portal class wasn't invoking the cleanup function if it didn't have attached content.
2. The dialog ref was keeping the reference after the dialog is closed. There is also a leak related to dialog refs which will be addressed separately.

Fixes #2734.

### DIFF
--- a/src/lib/core/portal/portal-directives.ts
+++ b/src/lib/core/portal/portal-directives.ts
@@ -70,7 +70,7 @@ export class PortalHostDirective extends BasePortalHost implements OnDestroy {
   }
 
   ngOnDestroy() {
-    this.dispose();
+    super.dispose();
   }
 
   /**
@@ -93,7 +93,7 @@ export class PortalHostDirective extends BasePortalHost implements OnDestroy {
         componentFactory, viewContainerRef.length,
         portal.injector || viewContainerRef.parentInjector);
 
-    this.setDisposeFn(() => ref.destroy());
+    super.setDisposeFn(() => ref.destroy());
     return ref;
   }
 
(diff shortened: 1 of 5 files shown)


JacobBrandt:
Maybe I'm missing something here but how are you suppose to get to the component instance data if its now null on exit?
https://github.com/angular/material2/pull/2875/files#diff-a67de94bd385d473627b7fafa65b0a2cR34

JacobBrandt:
nvm.  I saw that the MD_DIALOG_DATA was added to the documentation.  I guess I should have been using that all along anyways.

PR #2878, opened 2017-01-31
fix(menu): add [matMenuTriggerFor] selector and matMenuTriggerFor input

natsid:
Need this so that MdMenuTrigger can be used with the NoConflictStyleCompatibilityMode module (which allows and requires the 'mat' prefix for all selectors).


natsid:
I signed it!

natsid:
I signed it! (I needed to update my emails).

andrewseguin:
Will be part of the PR #2923

Issue #1913, opened 2016-11-17
[md-menu][bug] Tests crash in MdMenu.ngOnDestroy


pchiquet:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
The test should work

#### What is the current behavior?
Tests on a component that uses MdMenu crash in  MdMenu.ngOnDestroy, with error "**Cannot read property 'unsubscribe' of undefined**"

```Error: Error in ./AppComponent class AppComponent - inline template:12:2 caused by: Cannot read property 'unsubscribe' of undefined
        TypeError: Cannot read property 'unsubscribe' of undefined
            at MdMenu.ngOnDestroy (webpack:///C:/Users/User/git/training/material2-menu-test/~/@angular/material/menu/menu-directive.js:44:0 <- src/test.ts:41822:30)
            at Wrapper_MdMenu.ngOnDestroy (/MdMenuModule/MdMenu/wrapper.ngfactory.js:14:16)
            at CompiledTemplate.proxyViewClass.View_AppComponent0.destroyInternal (/DynamicTestModule/AppComponent/component.ngfactory.js:265:21)
            at CompiledTemplate.proxyViewClass.AppView.destroy (webpack:///C:/Users/User/git/training/material2-menu-test/~/@angular/core/src/linker/view.js:143:0 <- src/test.ts:55639:14)
            at CompiledTemplate.proxyViewClass.DebugAppView.destroy (webpack:///C:/Users/User/git/training/material2-menu-test/~/@angular/core/src/linker/view.js:368:0 <- src/test.ts:55864:38)
            at CompiledTemplate.proxyViewClass.View_AppComponent_Host0.destroyInternal (/DynamicTestModule/AppComponent/host.ngfactory.js:33:19)
            at CompiledTemplate.proxyViewClass.AppView.destroy (webpack:///C:/Users/User/git/training/material2-menu-test/~/@angular/core/src/linker/view.js:143:0 <- src/test.ts:55639:14)
            at CompiledTemplate.proxyViewClass.DebugAppView.destroy (webpack:///C:/Users/User/git/training/material2-menu-test/~/@angular/core/src/linker/view.js:368:0 <- src/test.ts:55864:38)
            at CompiledTemplate.proxyViewClass.AppView.detachAndDestroy (webpack:///C:/Users/User/git/training/material2-menu-test/~/@angular/core/src/linker/view.js:130:0 <- src/test.ts:55626:14)
[...]

PR #2893, opened 2017-02-01
fix(select): only unsubscribe from defined fields in ngOnDestroy

atscott:
Tests might not call fixture.detectChanges() so the _tabSubscription and _changeSubscription might not be defined in ngOnDestroy because ngAfterContentInit will not be called. When the TestBed is reset, calling unsubscribe on those will result in an error and fail all following tests.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -55,6 +55,10 @@ describe('MdSelect', () => {
     document.body.removeChild(overlayContainerElement);
   });
 
+  it('does not fail when detectChanges is not called', () => {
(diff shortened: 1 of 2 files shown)


atscott:
This appears related to [issue 1913](https://github.com/angular/material2/issues/1913). I could also apply the same fix to the [menu directive](https://github.com/angular/material2/blob/master/src/lib/menu/menu-directive.ts#L72)

jelbourn:
Was fixed in 998a583c0a1a3c2882752d9af596c3ef8e8e4fba

PR #2904, opened 2017-02-02
feat(datepicker): add month & year view

mmalerba:


src/lib/datepicker/year-view.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * The month in this year that the selected Date falls on.
+   * Null if the selected Date is in a different year.
+   */
+  _selectedMonth: number;
+
+  constructor(private _locale: CalendarLocale) {
+    // First row of months only contains 5 elements so we can fit the year label on the same row.
+    this._months = [[0, 1, 2, 3, 4], [5, 6, 7, 8, 9, 10, 11]].map(row => row.map(
```
tinayuangao: months value should start from 0 or 1? could you please add some docs?

src/lib/core/datetime/simple-date.spec.ts
```diff
@@ -9,4 +9,32 @@ describe('SimpleDate', () => {
   it('can be converted to native Date', () => {
     expect(new SimpleDate(2017, 0, 1).toNativeDate()).toEqual(new Date(2017, 0, 1));
   });
+
+  it('handles month and date overflow', () => {
+    expect(new SimpleDate(2017, 12, 32)).toEqual(new SimpleDate(2018, 1, 1));
+  });
+
+  it('handles month and date underflow', () => {
+    expect(new SimpleDate(2017, -1, 0)).toEqual(new SimpleDate(2016, 10, 30));
```
mmalerba: The months are 0-indexed to match native JS Dates, so 10 = November. I considered using 1-indexed because it make more sense, but I thought it was better to just stay consistent with what people already know. I'll add a note about the 0-indexing to the SimpleDate docs.

src/lib/core/datetime/simple-date.ts
```diff
@@ -3,12 +3,63 @@
...
+  get year() {
+    return this._date.getFullYear();
+  }
+
+  /** The month component of this date. (0-indexed, 0 = January). */
+  get month() {
+    return this._date.getMonth();
+  }
+
+  /** The date component of this date. */
```
tinayuangao: nit: Also add doc about date index?

src/lib/datepicker/month-view.ts
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * The date to display in this month view (everything other than the month and year is ignored).
+   */
+  @Input()
+  get date() { return this._date; }
+  set date(value) {
+    this._date = this._locale.parseDate(value) || SimpleDate.today();
+    this._init();
+  }
+  private _date = SimpleDate.today();
```
kara: Can you stack all the private properties at the top of the class? That's typically the format we use in other components so it's easy to find them.

mmalerba: When it's acting as backing for a public property with a getter/setter isn't it nicer to have it next to its getter/setter?

PR #2907, opened 2017-02-02
feat(select): allow focusing items by typing

### LINKED ISSUE #2668: When I have to open md-select and press 'A' key then move focus should be on 'A' string like 'Apple'
Please have a demo - [md-select example](http://plnkr.co/edit/JIEFHNvWK6LVwEsYLVD8?p=preview)

focus

Right now focus on 'Mango'. After press key 'A' I wanted to know second screenshot

plunker

crisbeto:
Allows for users to skip to a select item by typing, similar to the native select.

Fixes #2668.

### DIFF
--- a/src/cdk/a11y/list-key-manager.ts
+++ b/src/cdk/a11y/list-key-manager.ts
@@ -16,7 +16,7 @@ import {RxChain, debounceTime, filter, map, doOperator} from '@angular/cdk/rxjs'
 /**
  * This interface is for items that can be passed to a ListKeyManager.
  */
-export interface ListKeyManagerItem {
+export interface ListKeyManagerOption {
   disabled?: boolean;
   getLabel?(): string;
 }
@@ -25,13 +25,15 @@ export interface ListKeyManagerItem {
  * This class manages keyboard events for selectable lists. If you pass it a query list
  * of items, it will set the active item correctly when arrow events occur.
  */
-export class ListKeyManager<T extends ListKeyManagerItem> {
+export class ListKeyManager<T extends ListKeyManagerOption> {
   private _activeItemIndex = -1;
   private _activeItem: T;
   private _wrap = false;
-  private _pressedInputKeys: number[] = [];
   private _nonNavigationKeyStream = new Subject<number>();
   private _typeaheadSubscription: Subscription;
+
+  // Buffer for the letters that the user has pressed when the typeahead option is turned on.
+  private _pressedInputKeys: number[] = [];
 
   constructor(private _items: QueryList<T>) { }
 
@@ -57,6 +59,9 @@ export class ListKeyManager<T extends ListKeyManagerItem> {
       this._typeaheadSubscription.unsubscribe();
     }
 
+    // Debounce the presses of non-navigational keys, collect the ones that correspond to letters
+    // and convert those letters back into a string. Afterwards find the first item that starts
+    // with that string and select it.
     this._typeaheadSubscription = RxChain.from(this._nonNavigationKeyStream)
       .call(filter, keyCode => keyCode >= A && keyCode <= Z)
       .call(doOperator, keyCode => this._pressedInputKeys.push(keyCode))
@@ -64,12 +69,13 @@ export class ListKeyManager<T extends ListKeyManagerItem> {
       .call(filter, () => this._pressedInputKeys.length > 0)
       .call(map, () => String.fromCharCode(...this._pressedInputKeys))
       .subscribe(inputString => {
-        const activeItemIndex = this._items.toArray().findIndex(item => {
-          return item.getLabel!().toUpperCase().trim().startsWith(inputString);
-        });
+        const items = this._items.toArray();
 
-        if (activeItemIndex > -1) {
-          this.setActiveItem(activeItemIndex);
+        for (let i = 0; i < items.length; i++) {
+          if (items[i].getLabel!().toUpperCase().trim().indexOf(inputString) === 0) {
+            this.setActiveItem(i);
+            break;
+          }
         }
 
         this._pressedInputKeys = [];
@@ -95,6 +101,9 @@ export class ListKeyManager<T extends ListKeyManagerItem> {
     switch (event.keyCode) {
       case DOWN_ARROW: this.setNextItemActive(); break;
       case UP_ARROW: this.setPreviousItemActive(); break;
+
+      // Note that we return here, in order to avoid preventing
+      // the default action of unsupported keys.
       default: this._nonNavigationKeyStream.next(event.keyCode); return;
     }
 
--- a/src/cdk/a11y/list-key-manager.ts
+++ b/src/cdk/a11y/list-key-manager.ts
@@ -9,33 +9,35 @@
 import {QueryList} from '@angular/core';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
-import {UP_ARROW, DOWN_ARROW, TAB} from '@angular/cdk/keyboard';
+import {Subscription} from 'rxjs/Subscription';
+import {UP_ARROW, DOWN_ARROW, TAB, A, Z} from '@angular/cdk/keyboard';
+import {RxChain, debounceTime, filter, map, doOperator} from '@angular/cdk/rxjs';
 
 /**
- * This interface is for items that can be disabled. The type passed into
- * ListKeyManager must extend this interface.
+ * This interface is for items that can be passed to a ListKeyManager.
  */
-export interface CanDisable {
+export interface ListKeyManagerItem {
   disabled?: boolean;
+  getLabel?(): string;
 }
 
 /**
  * This class manages keyboard events for selectable lists. If you pass it a query list
  * of items, it will set the active item correctly when arrow events occur.
  */
-export class ListKeyManager<T extends CanDisable> {
-  private _activeItemIndex: number = -1;
+export class ListKeyManager<T extends ListKeyManagerItem> {
+  private _activeItemIndex = -1;
   private _activeItem: T;
-  private _tabOut = new Subject<void>();
-  private _wrap: boolean = false;
+  private _wrap = false;
+  private _pressedInputKeys: number[] = [];
+  private _nonNavigationKeyStream = new Subject<number>();
+  private _typeaheadSubscription: Subscription;
 
   constructor(private _items: QueryList<T>) { }
 
   /**
    * Turns on wrapping mode, which ensures that the active item will wrap to
    * the other end of list when there are no more items in the given direction.
-   *
-   * @returns The ListKeyManager that the method was called on.
    */
   withWrap(): this {
     this._wrap = true;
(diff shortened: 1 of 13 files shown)


Jagpreet16:
By when can we expect this feature to be available?

RenatoSilveiraGomes:
Can we expect this feature available on 2.0.0-beta.6?

otienoanyango:
Does this mean we should be hopeful for next beta?

Stas-tolpekin:
@kara,
When can we expect this function?

RBroden:
I got asked to create a select like this, as well. The select they compared it to was the angularjs ui-select select with search. So far, I am using a read-only input that displays a dropdown, non-input element on click (using a directive to handle clicks outside the container). That element has a search input and list. The search input filters the data and sets the read-only input upon clicking on a list item.

Markus-ipse:
@crisbeto Any chance you have time to rebase this PR so that it can be merged? :)

anthonyartese:
What is the status of this? Just encountered this issue when implementing md-select.

mikeomeara1:
Without this functionality, md-select doesn't work how an html select is expected to and it's confusing/counter-intuitive to users (not to mention a real pain when you get more than a couple of items in there). Please please...oh please can we get some love on this one? Sounds like it's fixed and ready to go yes?

kemmis:
Can we please please please have this feature? I agree that without this functionality it is a show-stopper from using the md-select control. What is needed in order to get this merged? Thanks!

src/lib/core/a11y/focus-key-manager.ts
```diff
@@ -36,4 +48,52 @@ export class FocusKeyManager extends ListKeyManager<Focusable> {
...
+        let inputString = String.fromCharCode.apply(String, this._pressedInputKeys);
+        let items = this._items.toArray();
+
+        this._pressedInputKeys.length = 0;
+
+        for (let i = 0; i < items.length; i++) {
+          let item = items[i];
+
+          // Note that fromCharCode returns uppercase letters.
+          if (items[i].getFocusableLabel!().toUpperCase().trim().startsWith(inputString)) {
```
ocarreterom: Replace `items[i]` by `item` or remove the var instance.

src/lib/core/a11y/focus-key-manager.ts
```diff
@@ -36,4 +48,52 @@ export class FocusKeyManager extends ListKeyManager<Focusable> {
...
+    if (this._hasLabelFn && keyCode >= INPUT_KEY_RANGE_START && keyCode <= INPUT_KEY_RANGE_END) {
+      this._debounceInputEvent(keyCode);
+    } else {
+      this._clearTimeout();
+      super.onKeydown(event);
+    }
+  }
+
+  /** Debounces the input key events and focuses the proper item after the last keystroke. */
+  private _debounceInputEvent(keyCode: number): void {
```
jelbourn: Could we do this with rxjs' debounce?

src/lib/core/a11y/focus-key-manager.ts
```diff
@@ -7,21 +7,33 @@
...
+
 
 export class FocusKeyManager extends ListKeyManager<Focusable> {
+  private _timer: number | null;
+  private _pressedInputKeys: number[] = [];
+  private _hasLabelFn: boolean;
 
   constructor(items: QueryList<Focusable>) {
     super(items);
+    this._hasLabelFn = items && items.first && 'getFocusableLabel' in items.first;
```
crisbeto: It's been a while since I did this. Revisiting it, it makes sense to require the `getFocusableLabel` as a part of the `Focusable` interface.

crisbeto:
@jelbourn re-did this one to move all of the logic into the `ListKeyManager`, get it working with the latest master, address the feedback and clean it up in general. Can you take another look?

src/cdk/a11y/list-key-manager.ts
```diff
@@ -9,42 +9,77 @@
...
  */
-export class ListKeyManager<T extends CanDisable> {
-  private _activeItemIndex: number = -1;
+export class ListKeyManager<T extends ListKeyManagerItem> {
+  private _activeItemIndex = -1;
   private _activeItem: T;
-  private _tabOut = new Subject<void>();
-  private _wrap: boolean = false;
+  private _wrap = false;
+  private _pressedInputKeys: number[] = [];
```
jelbourn: Add description for `_pressedInputKeys`? This one isn't as immediately obvious as the other properties

src/cdk/a11y/list-key-manager.ts
```diff
@@ -9,42 +9,77 @@
...
+  withTypeAhead(debounceInterval = 200): this {
+    if (this._items.length && this._items.some(item => typeof item.getLabel !== 'function')) {
+      throw Error('ListKeyManager items in typeahead mode must implement the `getLabel` method.');
+    }
+
+    if (this._typeaheadSubscription) {
+      this._typeaheadSubscription.unsubscribe();
+    }
+
+    this._typeaheadSubscription = RxChain.from(this._nonNavigationKeyStream)
```
jelbourn: Would be good to add a description of how you're using `_pressedInputKeys` in this chain

src/cdk/a11y/list-key-manager.ts
```diff
@@ -9,42 +9,77 @@
...
+      .call(filter, keyCode => keyCode >= A && keyCode <= Z)
+      .call(doOperator, keyCode => this._pressedInputKeys.push(keyCode))
+      .call(debounceTime, debounceInterval)
+      .call(filter, () => this._pressedInputKeys.length > 0)
+      .call(map, () => String.fromCharCode(...this._pressedInputKeys))
+      .subscribe(inputString => {
+        const activeItemIndex = this._items.toArray().findIndex(item => {
```
jelbourn: IE11 doesn't have `findIndex` or `startsWith`

jelbourn: `activeItemIndex` -> `matchedItemIndex`?

src/cdk/a11y/list-key-manager.ts
```diff
@@ -9,42 +9,77 @@
...
+import {UP_ARROW, DOWN_ARROW, TAB, A, Z} from '@angular/cdk/keyboard';
+import {RxChain, debounceTime, filter, map, doOperator} from '@angular/cdk/rxjs';
 
 /**
- * This interface is for items that can be disabled. The type passed into
- * ListKeyManager must extend this interface.
+ * This interface is for items that can be passed to a ListKeyManager.
  */
-export interface CanDisable {
+export interface ListKeyManagerItem {
```
jelbourn: Maybe `Item` -> `Option`? I think every time we use this it's for the `listbox` + `option` combo

src/cdk/a11y/list-key-manager.ts
```diff
@@ -58,20 +93,12 @@ export class ListKeyManager<T extends CanDisable> {
...
-        break;
-      case TAB:
-        // Note that we shouldn't prevent the default action on tab.
-        this._tabOut.next();
-        return;
-      default:
-        return;
+      case DOWN_ARROW: this.setNextItemActive(); break;
+      case UP_ARROW: this.setPreviousItemActive(); break;
+      default: this._nonNavigationKeyStream.next(event.keyCode); return;
```
jelbourn: Add back comment about using `return` here to skip `preventDefault`?

src/lib/core/a11y/focus-key-manager.ts
```diff
@@ -6,24 +6,18 @@
...
+import {ListKeyManager, ListKeyManagerItem} from './list-key-manager';
 
 /**
  * This is the interface for focusable items (used by the FocusKeyManager).
- * Each item must know how to focus itself and whether or not it is currently disabled.
+ * Each item must know how to focus itself, whether or not it is currently disabled
+ * and be able to supply it's label.
  */
-export interface Focusable extends CanDisable {
+export interface Focusable extends ListKeyManagerItem {
```
jelbourn: Maybe make this something like `FocusableOption` now?

src/lib/core/option/option.ts
```diff
@@ -174,6 +174,11 @@ export class MdOption {
     }
   }
 
+  /** Fetches the label to be used when determining whether the option should be focused. */
```
jelbourn: "Fetches" -> "Gets", since "fetch" implies a server call

src/cdk/a11y/list-key-manager.ts
```diff
@@ -9,42 +9,77 @@
...
   }
 
   /**
+   * Turns on typeahead mode which allows users to set the active item by typing.
+   * @param debounceInterval Time to wait after the last keystroke before setting the active item.
+   */
+  withTypeAhead(debounceInterval = 200): this {
+    if (this._items.length && this._items.some(item => typeof item.getLabel !== 'function')) {
+      throw Error('ListKeyManager items in typeahead mode must implement the `getLabel` method.');
+    }
```
jelbourn: Add test for this error?

crisbeto:
Addressed the feedback @jelbourn, can you take another look?

naveedahmed1:
Thank you so much team, awesome work! It was much needed :)

src/cdk/a11y/list-key-manager.ts
```diff
@@ -9,42 +9,83 @@
...
+
+    if (this._typeaheadSubscription) {
+      this._typeaheadSubscription.unsubscribe();
+    }
+
+    // Debounce the presses of non-navigational keys, collect the ones that correspond to letters
+    // and convert those letters back into a string. Afterwards find the first item that starts
+    // with that string and select it.
+    this._typeaheadSubscription = RxChain.from(this._nonNavigationKeyStream)
+      .call(filter, keyCode => keyCode >= A && keyCode <= Z)
```
andrewbents: so, it only searches by letters? What if options contain other characters? Also, a lot of languages have more letters, so some of them would not fit in the A-Z range

crisbeto: This checks the keyboard codes, not the actual letters. E.g. 68 will correspond to `d` in English and `д` in a Bulgarian phonetic keyboard.

andrewbents: Russian has 6 letters outside this range, they are on [ ] ; ' , . keys. So, they won't be handled? Also I think non-alphabetic characters should be supported as well

naveedahmed1: What about numbers? Does it work with numeric values? And when should we expect it to be release?

crisbeto: I've submitted https://github.com/angular/material2/pull/6543 that should handle more or less everything. It'll still fall back to A-Z and 0-9 for browsers that don't support `KeyboardEvent.key` (should be a very small percentage).

wags1999:
Any idea when this will be in a release? Our users are complaining, so we're considering ripping out md-select and using a different library's control... but if a release is imminent we'd just wait for that...

naveedahmed1:
Should this work with latest release (2.0.0-beta.11)? Somehow, its still not working for me.

PCASME:
I'm in version 2.0.0-beta.11 and does not work.

xagon:
do we have a working example of this somewhere?

mmalerba:
@xagon https://material.angular.io/components/select/examples

naveedahmed1:
Please also add support for:

Home key  - Which should take to the first item of the list
End Key - which should take to the last item of the list

kylekrzeski:
+1 When is this expected to be released?

benb7760:
@kylekrzeski two months ago in `2.0.0-beta.10`

PR #2931, opened 2017-02-04
feat(button): add focused state ripple

crisbeto:
Implements the focused state [pulsing ripple from the spec](https://material.io/guidelines/components/buttons.html#buttons-raised-buttons).

Example:
buttons

### DIFF
--- a/src/lib/button/_button-base.scss
+++ b/src/lib/button/_button-base.scss
@@ -24,10 +24,15 @@ $mat-fab-padding: 16px !default;
 $mat-mini-fab-size: 40px !default;
 $mat-mini-fab-padding: 8px !default;
 
+// Opacity for the focused overlay.
+$md-button-focus-overlay-opacity: 0.12 !default;
+
+
 // Applies base styles to all button types.
 %mat-button-base {
   box-sizing: border-box;
   position: relative;
+  overflow: hidden;
 
   // Reset browser <button> styles.
   @include mat-button-reset();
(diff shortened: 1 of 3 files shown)


jelbourn:
@crisbeto is this still necessary? Looks like there have a been a few button focus commits since this was opened.

dahaupt:
@jelbourn I just tested the latest master demo-app and it's not included, so it should be still necessary.

crisbeto:
The button focus commits have mostly been to adjust the color @jelbourn.

jelbourn:
@crisbeto @DevVersion looks like this one just fell through the cracks- worth revisiting?

crisbeto:
I'm willing to revisit it, if we still want to do it. I think I submitted it initially to gauge whether there was any interest in having this kind of animation.

jelbourn:
I'm open to it, but not really a priority over... anything else.

DevVersion:
I'm *personally* not really a fan of that focus indicator. We kind of have to port the ripples to CSS in a matter that they follow the given spec image (bloats the code).

The specs for that button focus effect are kind of ambiguous as well. In one screenshot, the focus indicator we currently have, is shown. In the other one, they show the one with the pulsing ripple.

Still, I think the specs are not really explicit here, and it adds more pain than it helps (_also there will be people that don't like that behavior_). Also MDC doesn't seem to use that focus indicator concept.

jelbourn:
Let's close for now and revisit if MDC web does it or if the Material Design spec changes to make a bigger deal of it.

PR #2933, opened 2017-02-04
fix(progress-bar-configurable-example): fixes color binding not working

JannicBeck:
Fixes color binding not working.


JannicBeck:
I signed it!

Issue #2902, opened 2017-02-02
(cleanup): add testing utilities


mmalerba:
many of our test files are littered with definitions of the same helper functions over and over again (e.g. dispatchXXXEvent), it would be nice to create common test utils for this

PR #2935, opened 2017-02-04
chore(focus-classes): fix failing tests when browser is blurred

### LINKED ISSUE #2903: (tests): firefox doesn't fire focus/blur events in tests
Had to exempt firefox from tests added in ##2646 due to this issue. It seems to be a result of firefox not firing the events when it isn't the active window. I was able to reproduce by doing the following:

1. patch #2646
2. remove firefox exemption from tests
3. switch local karma tests to use firefox
4. start tests and click the firefox window to make sure its active as soon as it pops up
5. tests pass
6. start tests again, click away from the firefox window as soon as it pops up to make sure its inactive
7. tests fail

The internet has some mentions of setting a firefox pref, focusmanager.testmode = true as a potential solution for this.

DevVersion:
On Saucelabs, browsers will run simultaneously and therefore can't focus all browser windows at the same time.

This is problematic when testing focus states. Chrome and Firefoxonly fire FocusEvents when the window is focused. This issue also appears locally.

Also while investigating the `karma-test-shim.js` file looked very ugly, so I refactored it a bit (ES6 not allowed)

Fixes #2903.

### DIFF
--- a/src/lib/core/style/focus-classes.spec.ts
+++ b/src/lib/core/style/focus-classes.spec.ts
@@ -32,14 +32,8 @@ describe('FocusOriginMonitor', () => {
 
     focusOriginMonitor.registerElementForFocusClasses(buttonElement, buttonRenderer);
 
-    // On Saucelabs, browsers will run simultaneously and therefore can't focus all browser windows
-    // at the same time. This is problematic when testing focus states. Chrome and Firefox
-    // only fire FocusEvents when the window is focused. This issue also appears locally.
-    let _nativeButtonFocus = buttonElement.focus.bind(buttonElement);
-    buttonElement.focus = () => {
-      document.hasFocus() ? _nativeButtonFocus() : dispatchFocusEvent(buttonElement);
-    };
-
+    // Patch the element focus to properly emit focus events when the browser is blurred.
+    patchElementFocus(buttonElement);
   }));
 
   it('manually registered element should receive focus classes', async(() => {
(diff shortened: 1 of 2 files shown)


mmalerba:
Are we not able to set the focusmanager.testmode pref? Are you sure this is a problem in chrome too? My tests seem to consistently pass in chrome

DevVersion:
Not really. Chrome does not have such a flag and setting it for Firefox in Saucelabs doesn't seem to be possible.

Kristiyan and I have the same problems locally (when not focused). Also IE11 seems to have similar problems on Saucelabs.

I think that this approach is be the most reliable one.

src/lib/core/style/focus-classes.spec.ts
```diff
@@ -32,21 +22,27 @@ describe('FocusOriginMonitor', () => {
...
     buttonRenderer = fixture.componentInstance.renderer;
     focusOriginMonitor = fom;
-    platform = pfm;
 
     focusOriginMonitor.registerElementForFocusClasses(buttonElement, buttonRenderer);
+
+    // On Saucelabs, browsers will run simultaneously and therefore can't focus all browser windows
+    // at the same time. This is problematic when testing focus states. Chrome and Firefox
+    // only fire FocusEvents when the window is focused. This issue also appears locally.
+    let _nativeButtonFocus = buttonElement.focus.bind(buttonElement);
```
mmalerba: can you extract this into a separate method e.g. `fixFocus(el: Element)`? also both this and `dispatchFocusEvent` would be good things to put in a separate test utilities directory

crisbeto:
What version of Chrome are you running @mmalerba? I think 56 started aggresively throttling background tabs, which might be what's causing the issue here.

src/lib/core/style/focus-classes.spec.ts
DevVersion: Done. I called `patchElementFocus`. I added a TODO note, so I can move all helpers to a global utility file as in #2902

PR #2937, opened 2017-02-05
chore(autocomplete): Add autocomplete exports to the `index.ts`.

emoralesb05:
Using the latest nightly build in https://github.com/angular/material2-builds there wasnt a way to import only the `MdAutocompleteModule` nor its components from `@angular/material` because it wasnt in the `index.ts` file.

Looked at the opened PR's and didnt see it in them either, so just creating this PR just in case (Also a nice way to do my first little contribution 😄)

### DIFF
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -1,6 +1,7 @@
 export * from './core';
 export * from './module';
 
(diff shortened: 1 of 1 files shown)


emoralesb05:
I signed it!

PR #2940, opened 2017-02-05
feat(dialog): allow for the dialog dimensions to be updated

### LINKED ISSUE #2930: Dialog change width after opening?
**#### Bug, feature request, or proposal:**
Feature request

**#### What is the expected behavior?**
Allow changing of MdDialogConfig in the opened component, through the MdDialogRef.

**#### What is the use-case or motivation for changing an existing behavior?**
We need to open a dialog with a width of 400px, and after a network request resize the dialog to around 800px, to display the retrieved data.

The reason we need the resizing is because when you first open the dialog there are only 2 fields that need to be entered, and this does not warrant a massive 800px dialog.

Having the ability to resize the dialog would be an amazing addition and it would be great to have control over that while the dialog is open.

Not sure if this is possible, but it would be a great feature.

Thanks!

crisbeto:
Adds an `updateDimensions` method to the `MdDialogRef` which allows the user to update a dialog's dimensions after it has been created.

Fixes #2930.

### DIFF
--- a/src/lib/dialog/dialog-ref.ts
+++ b/src/lib/dialog/dialog-ref.ts
@@ -53,13 +53,11 @@ export class MdDialogRef<T> {
   }
 
   /**
-   * Updates the dialog's dimensions.
-   * @param width New width of the dialog.
-   * @param height New height of the dialog.
-   * @param position New position of the dialog.
+   * Updates the dialog's position.
+   * @param position New dialog position.
    */
-  updateDimensions(width?: string, height?: string, position?: DialogPosition): void {
-    let strategy = this._overlayRef.getState().positionStrategy as GlobalPositionStrategy;
+  updatePosition(position?: DialogPosition): this {
+    let strategy = this._getPositionStrategy();
 
     if (position && (position.left || position.right)) {
       position.left ? strategy.left(position.left) : strategy.right(position.right);
@@ -73,7 +71,24 @@ export class MdDialogRef<T> {
       strategy.centerVertically();
     }
 
-    strategy.width(width).height(height);
     this._overlayRef.updatePosition();
+
+    return this;
+  }
+
+  /**
+   * Updates the dialog's width and height.
+   * @param width New width of the dialog.
+   * @param height New height of the dialog.
+   */
+  updateDimensions(width = 'auto', height = 'auto'): this {
+    this._getPositionStrategy().width(width).height(height);
+    this._overlayRef.updatePosition();
+    return this;
+  }
+
+  /** Fetches the position strategy object from the overlay ref. */
+  private _getPositionStrategy(): GlobalPositionStrategy {
+    return this._overlayRef.getState().positionStrategy as GlobalPositionStrategy;
   }
 }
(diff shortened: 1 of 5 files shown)


src/lib/dialog/dialog-ref.ts
```diff
@@ -35,4 +36,29 @@ export class MdDialogRef<T> {
...
     return this._afterClosed.asObservable();
   }
+
+  /**
+   * Updates the dialog's dimensions.
+   * @param width New width of the dialog.
+   * @param height New height of the dialog.
+   * @param position New position of the dialog.
+   */
+  updateDimensions(width?: string, height?: string, position?: DialogPosition): void {
```
kara: Hm it feels to me like position should be in a separate method, like `updatePosition()`. If you just want to update the position, having to write `null, null` as the first two args feels a bit awkward.  If we separate, we could make it more fluent, too:

```ts
dialog.updateDimensions('100px', '100px')
      .updatePosition({top: '0', left: '0'}) 
```

src/lib/dialog/dialog.spec.ts
```diff
@@ -259,6 +259,26 @@ describe('MdDialog', () => {
...
+    let overlayPane = overlayContainerElement.querySelector('.cdk-overlay-pane') as HTMLElement;
+
+    expect(overlayPane.style.marginLeft).toBe('250px');
+
+    dialogRef.updateDimensions(null, null, {
+      left: '500px'
+    });
+
+    expect(overlayPane.style.marginLeft).toBe('500px');
+  });
```
kara: We probably need tests for updating the width and height as well.

kara:
+ @jelbourn For API review

crisbeto:
Addressed the feedback @kara.

src/lib/dialog/dialog-ref.ts
```diff
@@ -50,4 +51,44 @@ export class MdDialogRef<T> {
...
+
+    return this;
+  }
+
+  /**
+   * Updates the dialog's width and height.
+   * @param width New width of the dialog.
+   * @param height New height of the dialog.
+   */
+  updateDimensions(width = 'auto', height = 'auto'): this {
```
jelbourn: `updateSize`?

src/lib/dialog/dialog.ts
```diff
@@ -92,11 +92,16 @@ export class MdDialog {
...
    * @returns A promise resolving to the OverlayRef for the created overlay.
    */
-  private _createOverlay(dialogConfig: MdDialogConfig): OverlayRef {
-    let overlayState = this._getOverlayState(dialogConfig);
+  private _createOverlay(config: MdDialogConfig): OverlayRef {
+    let overlayState = new OverlayState();
+    let strategy = this._overlay.position().global();
+
+    overlayState.hasBackdrop = true;
+    overlayState.positionStrategy = strategy;
```
jelbourn: Why remove the `_getOverlayState` method? I like separating that action from creating the overlay

crisbeto: It's not removed, I've moved it to the `dialogRef` to make it easier to update the size.

jelbourn: Sorry, I just meant this could still be
```ts
private _getOverlayState(dialogConfig: MdDialogConfig): OverlayState {
  let overlayState = new OverlayState();
  overlayState.hasBackdrop = true;
  overlayState.positionStrategy = this._overlay.position().global();

  return overlayState;
}
```

crisbeto:

PR #2943, opened 2017-02-05
feat: add testing utilities for components

### LINKED ISSUE #2902: (cleanup): add testing utilities

DevVersion:
* Moves existing testing helpers to a testing folder in the core package.
* The testing utilities currently consist of functions to create dom events and shorthands to dispatch them.
* Also fixes that **releases** include components `spec.d.ts` files.

The `core/testing` utilities are not includes in releases for now. It could be possible to expose them as part of the Component Toolkit.

**Note**: The slider HammerJS functions have been skipped here, since those depend on the GestureConfig which is only available in Dependency Injection. We need to revisit & create a design for that.

Closes #2902

### DIFF
--- a/src/lib/core/testing/dispatch-events.ts
+++ b/src/lib/core/testing/dispatch-events.ts
@@ -1,0 +1,26 @@
+import {
+  createFakeEvent,
+  createKeyboardEvent,
+  createMouseEvent,
+  createTransitionEndEvent
+} from './event-objects';
+
+/** Shorthand to dispatch a fake event on a specified node. */
+export function dispatchFakeEvent(node: Node, eventName: string) {
+  node.dispatchEvent(createFakeEvent(eventName));
+}
+
+/** Shorthand to dispatch a keyboard event with a specified key code. */
+export function dispatchKeyboardEvent(node: Node, keyCode: number, type = 'keydown') {
+  node.dispatchEvent(createKeyboardEvent(type, keyCode));
+}
+
+/** Shorthand to dispatch a mouse event on the specified coordinates. */
+export function dispatchMouseEvent(node: Node, type: string, x = 0, y = 0) {
+  node.dispatchEvent(createMouseEvent(type, x, y));
+}
+
+/** Shorthand to dispatch a transition event with a specified property. */
+export function dispatchTransitionEndEvent(node: Node, propertyName: string, elapsedTime = 0) {
+  node.dispatchEvent(createTransitionEndEvent(propertyName, elapsedTime));
+}
--- a/src/lib/core/testing/dispatch-events.ts
+++ b/src/lib/core/testing/dispatch-events.ts
@@ -1,26 +1,20 @@
 import {
   createFakeEvent,
   createKeyboardEvent,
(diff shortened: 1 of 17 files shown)


src/lib/core/testing/dispatch-events.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  createFakeEvent,
+  createKeyboardEvent,
+  createMouseEvent,
+  createTransitionEndEvent
+} from './event-objects';
+
+/** Shorthand to dispatch a fake event on a specified node. */
+export function dispatchFakeEvent(node: Node, eventName: string) {
```
DevVersion: Not really sure. I thought that Fake fits pretty good. It indicates that those are not real events and can lack of event details.

src/lib/core/testing/dispatch-events.ts
```diff
@@ -0,0 +1,26 @@
...
+  node.dispatchEvent(createFakeEvent(eventName));
+}
+
+/** Shorthand to dispatch a keyboard event with a specified key code. */
+export function dispatchKeyboardEvent(node: Node, keyCode: number, type = 'keydown') {
```
DevVersion: Good point. I will make them more consistent.

src/lib/core/testing/dispatch-events.ts
```diff
@@ -0,0 +1,26 @@
...
+  node.dispatchEvent(createKeyboardEvent(type, keyCode));
+}
+
+/** Shorthand to dispatch a mouse event on the specified coordinates. */
+export function dispatchMouseEvent(node: Node, type: string, x = 0, y = 0) {
+  node.dispatchEvent(createMouseEvent(type, x, y));
+}
+
+/** Shorthand to dispatch a transition event with a specified property. */
+export function dispatchTransitionEndEvent(node: Node, propertyName: string, elapsedTime = 0) {
```
DevVersion: I did this, because there is only one possible type for a TransitionEvent.

src/lib/core/testing/dispatch-events.ts
mmalerba: I see, I was interpreting "fake" as "not caused by user interaction". I guess this is ok then

src/lib/core/testing/dispatch-events.ts
mmalerba: Ah weird, I thought `transitionstart` would also be valid, but the docs do say just `transitionend`

DevVersion:
@mmalerba Done. Addressed your feedback :)

mmalerba:
Following error when trying to sync into Google:

`src/lib/core/testing/event-objects.ts:49: ERROR - variable TransitionEvent is undeclared`

mmalerba:
please rebase

DevVersion:
@mmalerba Done!

kara:
@DevVersion One more rebase, sorry.

DevVersion:
@kara Done.

PR #2944, opened 2017-02-05
fix(dialog): leaking MdDialogContainer references

### LINKED ISSUE #2876: Dialog leaking DialogRef and DialogContainer instances
This issue is for tracking purposes. Currently the dialog is keeping the `MdDialogContainer` and `MdDialogRef` references associated with it in memory forever.

Related to #2734.

crisbeto:
Fixes an issue that caused the `MdDialogContainer` references to not be cleaned up, and as a result, any of the `MdDialogRef` and `MdDialogConfig` instances as well. The issue seems to come from the fact that a couple of blocks that look like:
```
this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
  this._elementFocusedBeforeDialogWasOpened = document.activeElement;
  this._focusTrap.focusFirstTabbableElement();
});
```

End up being transpiled to:
```
var _this = this;
this._ngZone.onMicrotaskEmpty.first().subscribe(function () {
    _this._elementFocusedBeforeDialogWasOpened = document.activeElement;
    _this._focusTrap.focusFirstTabbableElement();
});
```

This seems to cause the browser to retain the `_this` reference after the dialog has been destroyed.

Fixes #2876.

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -136,6 +136,8 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
     // that it doesn't end up back on the <body>. Also note that we need the extra check, because
     // IE can set the `activeElement` to null in some cases.
     let toFocus = this._elementFocusedBeforeDialogWasOpened as HTMLElement;
+
+    // We shouldn't use `this` inside of the NgZone subscription, because it causes a memory leak.
     let animationStream = this._onAnimationStateChange;
 
     this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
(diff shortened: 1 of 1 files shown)


src/lib/dialog/dialog-container.ts
```diff
@@ -137,16 +135,15 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
...
     // that it doesn't end up back on the <body>. Also note that we need the extra check, because
     // IE can set the `activeElement` to null in some cases.
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-      let toFocus = this._elementFocusedBeforeDialogWasOpened as HTMLElement;
+    let toFocus = this._elementFocusedBeforeDialogWasOpened as HTMLElement;
+    let animationStream = this._onAnimationStateChange;
 
-      // We need to check whether the focus method exists at all, because IE seems to throw an
-      // exception, even if the element is the document.body.
+    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
```
jelbourn: Need a comment here that explains the "avoid using `this`" issue so that it doesn't come back in the future.

crisbeto:
Addressed the comment @jelbourn.

PR #2951, opened 2017-02-06
chore(tools): add syntax highlighting logic for examples

riavalon:
Also now runs syntax highlighting script in parallel mode


DevVersion:
Why aren't we doing this as part of a gulp task?

.gitignore
```diff
@@ -5,6 +5,7 @@
 /tmp
 /deploy
 /screenshots
+/src/examples-highlights
```
kara: Hm, it feels a little weird to have built files in the src folder. Can we move this to the root of the repo?

DevVersion: Just to not bloat the PR. @riavalon why aren't we using the gulp plugin `gulp-highlight` ([here](https://github.com/johannestroeger/gulp-highlight)).

This would make the task way smaller and also easier to maintain.

riavalon: Ahh I see, my bad. I was just moving the script over from the material.angular repo. It would make a lot more sense to use that gulp task. I'll make that change. I can definitely move the built folder to the root as well @kara 👍

Issue #2175, opened 2016-12-11
Replace forbidden-identifiers with custom tslint rules


DevVersion:
As proposed by @hansl we should use custom `tslint` rules instead of the `forbidden-identifiers` script.

I already played with that and run into several problems:

* The rules will be written with TypeScript but need to be loaded as `.js` compiled files.
    We could compile them in a gulp task, but then developers probably will commit the files.

*  We are currently using an older version of `tslint`. 
    Version `4.0.0` is already available, but includes some breaking changes like the deprecation of `no-unused-variable`

PR #2952, opened 2017-02-06
chore: update tslint to latest version

DevVersion:
* Updates TSLint to the latest version.
* Due to some breaking changes we had to temporary disable the `no-inferrable-types` rule. Need to revisit once https://github.com/palantir/tslint/issues/2158 is addressed.

**Note**: I wrote a TSLint modification that fixes all warnings for `no-inferrable-types`, but since Dgeni explicitly needs types the changes are obsolete (See [here](https://i.gyazo.com/c64e00a5ee66feed48ee7cbb3d527e39.png))

**Note**: This change is mostly in favor of #2175 because we want to have our own custom TSLint rules instead of having the `forbidden-identifiers` script.

References #2175

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "travis-after-modes": "0.0.7",
     "ts-node": "^2.0.0",
     "tslint": "^4.4.2",
+    "tslint-no-unused-var": "0.0.6",
     "typescript": "~2.0.10",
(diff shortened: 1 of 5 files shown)


DevVersion:
Made some changes that I discussed with @crisbeto.

PR #2953, opened 2017-02-06
fix(checkbox): rename tabindex to tabIndex

DevVersion:
* Renames the checkbox `tabindex` input to the `tabIndex` input. 

This is consistent with other API's like for the slide-toggle and also with native inputs. See https://github.com/angular/material2/pull/2688#discussion_r96473347


DevVersion:
As chatted with @tinayuangao. This should be ready.

PR #2955, opened 2017-02-06
fix(select): avoid going into infinite loop under certain conditions

### LINKED ISSUE #2950: md-select with ngModel freezes browser if other errors
Bug

Normal Angular2 unhandled error in console output

Browser/tab freezes

Simple md-select with an ngModel and another component after the md-select which throws an error.

**WARNING: If you open the following plunker template, your browser/tab WILL FREEZE (tested in Chrome and Safari)**
Plunker template: https://embed.plnkr.co/BARX13/

crisbeto:
Currently `md-select` calls `writeValue` recursively until the `QueryList` with all of the options is initialized. This usually means 1-2 iterations max, however in certain conditions (e.g. a sibling component throws an error on init) this may not happen and the browser would get thrown into an infinite loop.

This change switches to saving the attempted value assignments in a property and applying it after initialization.

Fixes #2950.


src/lib/select/select.ts
```diff
@@ -242,6 +245,15 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
+
+    // Assign any values that were deferred until the component is initialized.
+    if (this._tempValue) {
+      Promise.resolve(null).then(() => {
+        this.writeValue(this._tempValue);
+        this._tempValue = null;
+      });
+    }
+
     this._changeSubscription = this.options.changes.subscribe(() => {
```
kara: Given that we are doing the same thing on option changes (i.e. writing the value next tick), I think it might make more sense to just incorporate the logic into the existing stream with `startWith()`. `this._control.value` should have the correct value already, so saving a temp version shouldn't be necessary.

crisbeto: Makes sense, I'll rework it. This was already being handled in a similar way on the multiple selection PR.

crisbeto:
Regarding unit tests, this is a little hard to test, because the tests would either crash (if the infinite loop happens) or throw a runtime error (if we avoided the loop). That being said, we already have some tests that will fail if we don't handle the initial value properly.

kara:
@crisbeto How about you use the pattern

```ts
expect(() => {
   const fixture = TestBed.createComponent(MyComp);
   fixture.detectChanges();
}).toThrowError();
```

If the tests don't crash, we know it's good. If they do crash, obviously there's a problem.

crisbeto:
Makes sense, but it probably needs an extra comment about it as well.

crisbeto:
Added the test and used your approach with `startWith` @kara. Should be good to go. It looks like some positioning-related tests are failing, although they passed locally and shouldn't be related to these changes.

src/lib/select/select.ts
```diff
@@ -242,7 +244,8 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
   ngAfterContentInit() {
     this._initKeyManager();
     this._resetOptions();
```
kara: You should be able to remove this call now that it's happening again in line 249

crisbeto: Done. Should I re-label?

kara:
@crisbeto Can you rebase?

PR #2967, opened 2017-02-07
feat(overlay): allow theming overlay-based components

### LINKED ISSUE #2662: multiple themes don't work with overlay-based components

Bug

when I add a class refering to an angular material them, the md-select component must be affected by this theme

The md-select component keeps the default theme. The other components I tried (tabs, input, checkbox) are working correctly.

Create 2 themes, one default and one inside a css class, and add this css class on a parent of a md-select component.

To have a fully functional multiple themes feature

I'm using Angular 2.4.3, Material 2-beta1 (using latest build repository), and chrome.
I don't know if others browsers are affected, but I'm assuming it's the case.

crisbeto:
Allows for a `themeClass` to be passed to overlay-based components which don't usually inherit their parent theme, because they're outside the DOM order.

Fixes #2662.

### DIFF
--- a/guides/theming.md
+++ b/guides/theming.md
@@ -46,7 +46,7 @@ ensures that the proper theme background is applied to your page.
 
 #### Theming overlay-based components
 Since certain components (e.g. `dialog`) are inside of a global overlay container, your theme may
-not be applied to them. In order to define the theme, that will be used for overlay components, you
+not be applied to them. In order to define the theme that will be used for overlay components, you
 have to specify it on the global `OverlayContainer` instance:
 
 ```ts
--- a/guides/theming.md
+++ b/guides/theming.md
@@ -44,6 +44,24 @@ Finally, if your app's content **is not** placed inside of a `md-sidenav-contain
 need to add the `mat-app-background` class to your wrapper element (for example the `body`). This
 ensures that the proper theme background is applied to your page.
 
+#### Theming overlay-based components
+Since certain components (e.g. `dialog`) are inside of a global overlay container, your theme may
+not be applied to them. In order to define the theme, that will be used for overlay components, you
+have to specify it on the global `OverlayContainer` instance:
+
+```ts
+import {OverlayContainer} from '@angular/material';
+
+@NgModule({
+  // misc config goes here
+})
+export class YourAppModule {
+  constructor(overlayContainer: OverlayContainer) {
+    overlayContainer.themeClass = 'your-theme';
+  }
+}
+```
+
 ### Defining a custom theme
 When you want more customization than a pre-built theme offers, you can create your own theme file.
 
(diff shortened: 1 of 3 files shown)


fxck:
Couldn't this be made a bit more versatile, so it could be used for [these](https://github.com/angular/material2/pull/1584) use-cases? I remember I couldn't quite decide whether it makes more sense to put the custom class on overlay or the panes or both separately (to allow stacking inside the container as well as stacking of containers).

crisbeto:
If I remember correctly, the main purpose of #1584 was to sort out the overlay stacking, which has been resolved for a while now. The purpose of this PR is intended to be only for themes since it's Material's job to take care of everything else. Ideally, once we have theme inheritance, this would only be relevant for the dialog, at which point we could move it to a dialog-specific config. For now we also need it for select, tooltip, menu etc.

fxck:
It allowed you to add an extra class(es) to the overlay/pane, which you could've(amongst other things) used to control z-index. Also no, I don't think it's been properly resolved, as z-index of the overlay is still capped at 1000, imagine you have your own custom fixed top bar, which has z-index of 1001, from that you have an ability to open md-menu, which will open an overlay with z-index 1000, so your top bar will be overlaying it, making it impossible to click on first items - http://plnkr.co/edit/5nzNsDsQB5BxwdcCwKW1?p=preview

Basically by making this more "versatile" I mean just giving it a bit better name(`customClass`?).

crisbeto:
The `z-index` is 1000 for a reason - it's important that nothing is above it. If you have a custom toolbar that is more than 1000, you can change the `z-index` of your toolbar instead. I understand that it makes it more versatile, but these elements aren't meant to be customisable. The theming is an edge case that should go away at some point.

fxck:
Then it should perhaps be mentioned somewhere, that none of your custom elements should have z-index above 1000 and that all other 3rd party components you use could possible conflict with that and needs to have their z-indexes overriden.. because you know, people tend to throw a lot of `z-index: 999999` around..  https://github.com/search?utf8=%E2%9C%93&q=%22z-index%3A+999999%22&type=Code&ref=searchresults

guides/theming.md
```diff
@@ -44,6 +44,24 @@ Finally, if your app's content **is not** placed inside of a `md-sidenav-contain
 need to add the `md-app-background` class to your wrapper element (for example the `body`). This
 ensures that the proper theme background is applied to your page.
 
+#### Theming overlay-based components
+Since certain components (e.g. `dialog`) are inside of a global overlay container, your theme may
+not be applied to them. In order to define the theme, that will be used for overlay components, you
```
kara: Typo (extra comma): `In order to define the theme that will be used for overlay components`

mmalerba:
can you look into whats up with the lint task? I've retired serveral times but it keeps failing

crisbeto:
Not exactly sure what was wrong, but it seems like a rebase fixed it.

xtianus79:
I am still a little confused as to what is the end result of this effort

crisbeto:
Since the overlay components are added at the end of the body, they might not inherit your theme. This allows you to add a theme to those overlay components.

xtianus79:
@crisbeto so this is only in beta.3 and above?

PR #2973, opened 2017-02-07
fix(select): add font family property

LendsMan:
Md-input-container and even md-option have font-families, but md-select does not.

### DIFF
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -1,6 +1,7 @@
 @import '../core/style/menu-common';
 @import '../core/style/list-common';
 @import '../core/style/form-common';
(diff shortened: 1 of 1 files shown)


src/lib/select/select.scss
```diff
@@ -12,6 +13,7 @@ $md-select-panel-max-height: 256px !default;
 md-select {
   display: inline-block;
   outline: none;
+  font-family: $md-font-family;
```
tinayuangao: Should be `$mat-font-family` now

LendsMan: @tinayuangao Ok, updated.

PR #2974, opened 2017-02-07
feat(focus-classes): expose focus origin changes through observable

mmalerba:
also has a fix to restore focus origin after window loses focus and then regains it

### DIFF
--- a/src/lib/core/style/focus-classes.ts
+++ b/src/lib/core/style/focus-classes.ts
@@ -12,8 +12,8 @@ export class FocusOriginMonitor {
   /** The focus origin that the next focus event is a result of. */
   private _origin: FocusOrigin = null;
 
-  /** A WeakMap used to track the last element focused via the FocusOriginMonitor. */
-  private _lastFocused = new WeakMap<Element, FocusOrigin>();
+  /** The FocusOrigin of the last focus event tracked by the FocusOriginMonitor. */
+  private _lastFocusOrigin: FocusOrigin;
 
   /** Whether the window has just been focused. */
   private _windowFocused = false;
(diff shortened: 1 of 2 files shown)


src/lib/core/style/focus-classes.ts
```diff
@@ -39,21 +56,37 @@ export class FocusOriginMonitor {
...
+
     renderer.setElementClass(element, 'cdk-focused', true);
     renderer.setElementClass(element, 'cdk-keyboard-focused', this._origin == 'keyboard');
     renderer.setElementClass(element, 'cdk-mouse-focused', this._origin == 'mouse');
-    renderer.setElementClass(element, 'cdk-program-focused',
-        !this._origin || this._origin == 'program');
+    renderer.setElementClass(element, 'cdk-program-focused', this._origin == 'program');
+
+    subject.next(this._origin);
+    this._lastFocused = new WeakMap().set(element, this._origin);
```
DevVersion: I'm trying to understand why we need a `WeakMap` when the the WeakMap always contains one element at most.

I think we can assume that the browser always returns focus to the last focused element and multiple elements can't be focused.

src/lib/core/style/focus-classes.ts
```diff
@@ -39,21 +56,37 @@ export class FocusOriginMonitor {
...
   }
 
   /** Handles blur events on a registered element. */
-  private _onBlur(element: Element, renderer: Renderer) {
+  private _onBlur(element: Element, renderer: Renderer, subject: Subject<FocusOrigin>) {
     renderer.setElementClass(element, 'cdk-focused', false);
     renderer.setElementClass(element, 'cdk-keyboard-focused', false);
     renderer.setElementClass(element, 'cdk-mouse-focused', false);
     renderer.setElementClass(element, 'cdk-program-focused', false);
+    subject.next(null);
```
DevVersion: Do we really want to emit `null` on blur? I didn't see any test for it yet.

mmalerba: I do think we want this, because I want to later expand this to apply the classes if the element _or one of its children_ is focused, and it would be nice to have an easy way for developers to know when focus leaves the subtree. I'll add tests though

DevVersion: Sounds good. Just was wondering because of the tests.

src/lib/core/style/focus-classes.ts
mmalerba: I guess I was thinking I needed a reference to the element to check that it was the same one, but you're probably right that I can safely assume that. Let me try.

mmalerba:
updated, PTAL

PR #2976, opened 2017-02-08
fix(select): set default font size

kara:
Currently the font size in the trigger is purely inherited, which causes weirdness when inheriting a font size that's different from the default option size.

### DIFF
--- a/src/lib/core/option/_option.scss
+++ b/src/lib/core/option/_option.scss
@@ -1,6 +1,9 @@
 @import '../style/menu-common';
 @import '../a11y/a11y';
 
+/** Default size of option text. */
+$mat-option-font-size: 16px !default;
+
 /**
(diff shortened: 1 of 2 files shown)


src/lib/core/option/_option.scss
```diff
@@ -1,6 +1,9 @@
 @import '../style/menu-common';
 @import '../a11y/a11y';
 
+/** Default size of option text. */
+$md-option-font-size: 16px !default;
```
tinayuangao: We should change this to `$mat-option-font-size`

kara: Good catch! I'll fix now

kara:
@tinayuangao Updated with the new prefix and rebased. Should be good to go now.

PR #2989, opened 2017-02-08
Render template portal elements to properly determine placement

trshafer:
fix(overlay): Render the templates before placing them in the overlay.

This fixes positioning issues when rendering templates which contain embedded templates.
The templates need to be rendered in order to properly determine width which can then determine placement.

### DIFF
--- a/src/demo-app/overlay/overlay-demo.html
+++ b/src/demo-app/overlay/overlay-demo.html
@@ -10,6 +10,10 @@
   Pasta 3
 </button>
 
+<button cdk-overlay-origin #tortelliniOrigin="cdkOverlayOrigin" (click)="openTortelliniPanel()">
+  Pasta 4
+</button>
+
 
 <button cdk-overlay-origin #trigger="cdkOverlayOrigin" (click)="isMenuOpen = !isMenuOpen">
   Open menu
@@ -25,6 +29,10 @@
 <!-- Template to load into an overlay. -->
 <ng-template cdk-portal>
   <p class="demo-fusilli"> Fusilli </p>
+</ng-template>
+
+<ng-template cdk-portal #tortelliniTemplate="cdkPortal">
+  <ul class="demo-tortellini"><li *ngFor="let filling of tortelliniFillings">{{filling}}</li></ul>
 </ng-template>
 
 <button (click)="openPanelWithBackdrop()">Backdrop panel</button>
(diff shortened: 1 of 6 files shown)


trshafer:
I'm getting the same button errors on my local machine on master.

The menu rtl failue only happens on ie11 and edge. Not sure how to proceed.

crisbeto:
The focus-related one is expected. It'll be fixed by https://github.com/angular/material2/pull/2935. Regarding the select failures, I think that it may have to do with a button from another test that doesn't get removed. I've fixed it in another PR, but it hasn't been merged yet. You can try adding `document.body.removeChild(button)` here at https://github.com/angular/material2/blob/master/src/lib/dialog/dialog.spec.ts#L384

trshafer:
Ok I have rebased and the button issue has gone away. It is still having issues with RTL rendering on windows. Any advice is appreciated. I can no longer get ng2 material to test on my local machine as well:

```
node_modules/@types/jasmine/index.d.ts(39,38): error TS2371: A parameter initializer is only allowed in a function or constructor implementation.
node_modules/@types/jasmine/index.d.ts(39,46): error TS2304: Cannot find name 'keyof'.
node_modules/@types/jasmine/index.d.ts(39,52): error TS1005: '=' expected.
```

trshafer:
Any assistance on debugging the IE versions would be greatly appreciated.

trshafer:
I have created #3163 which identifies the test failure. I do not believe the failure was caused by this change.

jelbourn:
@kara I've confirmed that this is an issue; we should figure out why it's failing.

trshafer:
Hey @jelbourn,

I added a test for DomPortalHost and TemplatePortalHost. 

While writing the tests, I actually wound up reverting the TemplatePortalHost changes. My reasoning is that it was not possible to render the template outside of the parent component change detection. The template is embedded within the same component as the portal host. Unlike the DomPortalHost which the template is added to a different exterior host. Does that make sense and is this logic accurate? It does solve the current overlay issue, but the two implementations are different in that one explicitly renders the template and one implicitly renders it.

trshafer:
@jelbourn green build! :tada: 

Let me know if I should squash commits, fix commit messages, refactor code, etc.

trshafer:
Squashed the commits.

PR #2992, opened 2017-02-08
Update getting-started.md

codenameEli:
Running through the setup and accidentally ran `npm install --save @angular/material` in my projects folder, quickly realized i forgot to cd in there. Thought it would be frustrating if someone ran through the entire process of setting up without cd-ing.


codenameEli:
I signed it!

jelbourn:
#3523 removed the `ng new` from the getting started guide

PR #2994, opened 2017-02-08
feat(datepicker): add calendar component that pulls together month and year views

mmalerba:
styling still needs some love

### DIFF
--- a/src/lib/datepicker/calendar.html
+++ b/src/lib/datepicker/calendar.html
@@ -1,0 +1,39 @@
+<div class="mat-calendar-header">
+  <button class="mat-calendar-button mat-calendar-period-button" (click)="_currentPeriodClicked()">
+    {{_label}}
+    <div class="mat-calendar-arrow"></div>
+  </button>
+  <div class="mat-calendar-spacer"></div>
+  <button class="mat-calendar-button mat-calendar-previous-button" (click)="_previousClicked()">
+    <svg xmlns="http://www.w3.org/2000/svg" width="24px" height="24px" viewBox="0 0 24 24"
+         fill="#000000">
+      <path d="M15.41 16.09l-4.58-4.59 4.58-4.59L14 5.5l-6 6 6 6z"/>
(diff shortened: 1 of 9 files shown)


src/lib/datepicker/calendar.html
```diff
@@ -0,0 +1,39 @@
+<div class="mat-calendar-header">
+  <button class="mat-calendar-button mat-calendar-period-button" (click)="_currentPeriodClicked()">
+    {{_label}}
+    <div class="mat-calendar-arrow"></div>
+  </button>
+  <div class="mat-calendar-spacer"></div>
+  <button class="mat-calendar-button mat-calendar-previous-button" (click)="_previousClicked()">
+    <svg xmlns="http://www.w3.org/2000/svg" width="24px" height="24px" viewBox="0 0 24 24"
+         fill="#000000">
```
tinayuangao: Use css color instead of hard code in html? Why do we put this in a svg file?

src/lib/datepicker/calendar.spec.ts
```diff
@@ -0,0 +1,128 @@
+import {async, TestBed, ComponentFixture} from '@angular/core/testing';
+import {MdDatepickerModule} from './index';
+import {Component} from '@angular/core';
+import {SimpleDate} from '../core/datetime/simple-date';
+import {MdCalendar} from './calendar';
+import {By} from '@angular/platform-browser';
+
+
+describe('MsCalendar', () => {
```
tinayuangao: nit: is it MdCalendar?

src/lib/datepicker/calendar.html
mmalerba: These are "keyboard arrow left" and "keyboard arrow right" from go/icons. I'd rather not mess with the contents of the svgs. I'm embedding them because its a pattern we already follow elsewhere in the code (I assume to avoid incurring additional resource requests). see: https://github.com/angular/material2/blob/master/src/lib/checkbox/checkbox.html#L24

tinayuangao:
LGTM

PR #2997, opened 2017-02-09
fix(autocomplete): hide instead of close when options empty

kara:
r: @tinayuangao


src/lib/autocomplete/autocomplete.ts
```diff
@@ -54,11 +57,18 @@ export class MdAutocomplete {
     }
   }
 
+  /** Panel should hide itself when the option list is empty. */
+  _setVisibility() {
+    Promise.resolve().then(() => this.showPanel = !!this.options.length);
```
andrewseguin: Is this Promise based so that we invoke a change detection or wait for something to stabilize?

kara: Yes, this ensures that change detection runs at the correct time.

Issue #2999, opened 2017-02-09
Misaligned icon in nav-list


AsherFoster:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
md-icon should be vertically centered relative to list item text

#### What is the current behavior?
md-icon is misaligned

#### What are the steps to reproduce?
```
md-nav-list
  md-list-item
    md-icon[md-list-avatar]
    p[md-line]
```
http://plnkr.co/edit/qAVB6mOYNf4p8yxtRSQX?p=preview
![Screenshot](https://cloud.githubusercontent.com/assets/11432528/22768820/65fb38d2-eee9-11e6-8ff3-f7c979946d31.png)

#### What is the use-case or motivation for changing an existing behavior?
Devs shouldn't have to modify CSS to achieve what should be the default behaviour of this list

#### Which versions of Angular, Material, OS, browsers are affected?
2.0.0-beta.1. Tested on Chrome 56.

#### Is there anything else we should know?
Commenting out the height: 40px line in devtools seemed to resolve the problem, but this may break alignment for multiline items

doender:
I used `<md-icon md-list-icon></md-icon>` instead of `md-list-avatar`, which seems to work.
I think the documentation is off here.

ocarreterom:
https://github.com/angular/material2/issues/2650

kara:
`md-list-icon` is intended to be used with icons, while `md-list-avatar` is really intended for image tags. This is really unclear in the documentation, though, so I've created this PR to fix that.

PR #3015, opened 2017-02-09
docs(list): add docs for md-list-icon

kara:
r: @mmalerba 

Related to #2999

### DIFF
--- a/src/lib/list/list.md
+++ b/src/lib/list/list.md
@@ -69,8 +69,27 @@ as shown in the example).
 </md-list>
 ```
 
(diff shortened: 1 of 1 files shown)


src/lib/list/list.md
```diff
@@ -69,8 +69,27 @@ as shown in the example).
 </md-list>
 ```
 
+### Lists with icons
+
+To add an icon to your list item, use the `md-list-icon` attribute.
```
crisbeto: Perhaps we should use the `mat-` prefix from now on?

PR #3016, opened 2017-02-09
fix(snack-bar): SimpleSnackBar not being exported

### LINKED ISSUE #3010: feat(snack-bar): MdSnackBarRef can't get proper type

Proposal

Write the expected type for `MdSnackBarRef`

You have to define the type as `any`

This:

`let snackBarRef = snackBar.open('Message archived');`

cannot be written like this:

`let snackBarRef: MdSnackBarRef<SimpleSnackBar> = snackBar.open('Message archived');`

because there is no import for `SimpleSnackBar`

It is more convenient for code writing in TypeScript if the variable type is known. 

All

crisbeto:
* Fixes the `SimpleSnackBar` class not being exported.
* Moves the `MdSnackBarModule` to the `index.ts` for consistency with other modules.

Fixes #3010.

### DIFF
--- a/src/lib/module.ts
+++ b/src/lib/module.ts
@@ -27,7 +27,7 @@ import {MdIconModule} from './icon/index';
 import {MdProgressSpinnerModule} from './progress-spinner/index';
 import {MdProgressBarModule} from './progress-bar/index';
 import {MdInputModule} from './input/index';
-import {MdSnackBarModule} from './snack-bar/snack-bar';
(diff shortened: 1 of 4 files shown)


andrewseguin:
It seems we aren't too consistent with this index.ts file when it comes to putting modules into it. Should we consider putting all component modules into their respective index.ts files?

crisbeto:
I'd say so. I've had it on my backlog for a while.

PR #3018, opened 2017-02-09
docs(list): fix incorrect docs for md-list

AsherFoster:
Related to #2999. Code example shows md-icon with [md-list-avatar] instead of [md-list-icon]


AsherFoster:
I signed it!

PR #3020, opened 2017-02-09
feat(FocusOriginMonitor): add support for touch events 

mmalerba:


src/lib/core/style/focus-classes.ts
```diff
@@ -3,7 +3,10 @@ import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 
 
-export type FocusOrigin = 'mouse' | 'keyboard' | 'program';
+const TOUCH_BUFFER_MS = 650;
```
mmalerba: @DevVersion I stole this number from material 1, but any idea where it comes from?

DevVersion: Yeah I think the timeout is still fine. I took it from https://github.com/angular/material/pull/7965#r59114101

DevVersion: Also mind exporting this variable? So the tests can use it for their `setTimeout`'s?

mmalerba:
@DevVersion I've made the change to compare event targets that you were asking for. While it is a little simpler, I believe it is less correct than what I had previously. Consider this example:

```
<div tabindex=1 #d cdkFocusClasses>
    <button (click)="d.focus()"></button>
</div>
```
Here the `div` will be given the `cdk-touch-focused` when the `button` is touched. This is technically incorrect, though I admit it is a very small edge case. So its a question of whether we keep the simpler almost correct solution or go with the more complicated but more correct solution.

DevVersion:
@mmalerba I see. I just ran your PR locally on my end and saw the issue. 

I went through the `_wasCausedByTouch` logic again and I'm pretty sure that we could remove the `.contains` check here.

| Prop | Description |
| ----- | ------------- |
| TouchEvent.target | Target is always type of `Element` and can be stored |
| FocusEvent.target | If the same as `TouchEvent.target` then it was caused by touch. <br/>Focus will **only** happen if the element is explicitly touched. No subtree handling necessary |

This would also solve the issue you mentioned above.

PR #3021, opened 2017-02-09
fix(select): set select value to trigger height and center text

### LINKED ISSUE #2977: Select offset 1px off in Chrome
Since options switched to use line-height rather than align-items, the alignment on the y-axis has been 1px off in Chrome. Looks right in Firefox.

Update: Appears to only happen on Mac, not Windows

andrewseguin:
Issue was that the text div was actually being calculated as 29px, not 30px, because the trigger had a border bottom of 1px. This made the following code incorrect since the option value height was actually 29px:

```
/**
 * Must adjust for the difference in height between the option and the trigger,
 * so the text will align on the y axis.
 * (SELECT_OPTION_HEIGHT (48) - SELECT_TRIGGER_HEIGHT (30)) / 2 = 9
 */
export const SELECT_OPTION_HEIGHT_ADJUSTMENT = 9;
``` 

This change forces the option value to be 30px and centers the text using flex. Tested on Linux Chrome, Linux FF, Windows IE11 & Edge, Mac Chrome, and Mac FF. 

Closes #2977

### DIFF
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -29,7 +29,6 @@ $mat-select-trigger-font-size: 16px !default;
   font-size: $mat-select-trigger-font-size;
 
   [aria-disabled='true'] & {
-    background-position: 0 bottom;
     cursor: default;
     user-select: none;
   }
@@ -82,7 +81,7 @@ $mat-select-trigger-font-size: 16px !default;
 
 .mat-select-value {
   position: absolute;
-  max-width: calc(100% - #{$mat-select-arrow-size * 2});
+  max-width: calc(100% - #{($mat-select-arrow-size + $mat-select-arrow-margin) * 2});
 
   // Firefox and some versions of IE incorrectly keep absolutely
   // positioned children of flex containers in the flex flow when calculating
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -78,13 +78,16 @@ $mat-select-trigger-font-size: 16px !default;
   // position. This has been fixed for Firefox 52, slated for early 2017.
   // Bug report: https://bugzilla.mozilla.org/show_bug.cgi?id=874718
   //
-  // In the meantime, we must adjust the left position to 0 to mimic where it
-  // would be if it were correctly taken out of the flex flow. It's also necessary
-  // to adjust the top value because absolutely positioned elements should not be
-  // affected by the flex container's "align-items" property either. To center the text,
-  // we must offset by 6px (6px top + 6px bottom + 18px text height = 30px total height).
+  // In the meantime, we must adjust the position to fit the top and left edge of the
+  // containing element with a height matching the trigger container.
+  // In doing so, we can use align-items: center to allow the text to
+  // correctly position itself in the middle of the container.
+  top: 0;
   left: 0;
-  top: 6px;
+  height: $mat-select-trigger-height;
+
+  display: flex;
+  align-items: center;
 
   [dir='rtl'] & {
     left: auto;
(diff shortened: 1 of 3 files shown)


crisbeto:
Great catch @andrewseguin! I was struggling with reproducing this one.

src/lib/select/select.scss
```diff
@@ -74,13 +74,16 @@ $mat-select-panel-max-height: 256px !default;
...
+  // In the meantime, we must adjust the position to fit the top and left edge of the
+  // containing element with a height matching the trigger container.
+  // In doing so, we can use align-items: center to allow the text to
+  // correctly position itself in the middle of the container.
+  top: 0;
   left: 0;
-  top: 6px;
+  height: $mat-select-trigger-height;
+
+  display: flex;
```
crisbeto: Can you check if long values get ellipsis with this? I had a fix a while ago that removed it, because the ellipsis didn't seem to work with flexbox.

andrewseguin: Just checked, there's an issue with ellipsis. I believe this can be solved by giving the text itself a span of its own that will have the overflow hidden property. I'll add that to the PR shortly

andrewseguin:
So I added a bit more changes to make it a little more explicit and future-proof. Children of an element with a border end up with different sizes for `height: 100%` depending on the browser. In our case, value was getting something like 30px on Chrome and Safari and 29px on FF. 

I took border-bottom out and just made a new span to handle the look of a border. That way the option element is for sure 30px on all browsers.

I added an inner span to take care of our truncation as well, thanks for the heads up on that.

crisbeto:
Instead of creating something completely new, couldn't the `.mat-select-border` share the styles from the `.mat-input-underline` from input container?

src/lib/select/select.scss
```diff
@@ -22,17 +22,28 @@ $mat-select-panel-max-height: 256px !default;
...
     cursor: default;
     user-select: none;
   }
 }
 
+.mat-select-border {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
kara: There's a mixin for these 4 lines in core called `mat-fill`. Use that?

src/lib/select/select.scss
```diff
@@ -22,17 +22,28 @@ $mat-select-panel-max-height: 256px !default;
   min-width: $mat-select-trigger-min-width;
   cursor: pointer;
   position: relative;
-  box-sizing: border-box;
 
   [aria-disabled='true'] & {
-    @include mat-control-disabled-underline();
-    border-bottom: transparent;
     background-position: 0 bottom;
```
kara: Looks like you moved this line. Still necessary to keep up here?

andrewseguin: Missed that one, will remove

src/lib/select/select.scss
andrewseguin: Almost, I want to have it sink to the bottom of the container so it's missing the `top: 0`

andrewseguin:
I tried getting a mixin to work with both select and input underlines but since they have different structures its hard to get a good common ground. Select needs to position bottom but this messes up on input. If we can find a way to structure them similarly then I think its worth finding a commons style, but perhaps in another PR.

As for using ::after, I felt that it was a little more complex for users to figure out what they need to override to change the stylings (e.g. underline size, color). By having a separate span, it makes it very easy for them to understand.

src/lib/select/select.scss
kara: Ha, I can't read. Ignore me!

PR #3024, opened 2017-02-09
feat(datepicker): create the md-datepicker component

mmalerba:
this component is the public interface to the datepicker and is responsible for the decision to launch in touch or non-touch mode. it is not yet hooked up to the actual calendar


src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+/** Component responsible for managing the datepicker popup/dialog. */
+@Component({
+  moduleId: module.id,
+  selector: 'md-datepicker',
+  templateUrl: 'datepicker.html',
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MdDatepicker implements OnDestroy {
+  /** Whether the calendar is in UI is in touch mode. */
```
kara: Typo: `is in UI is in touch mode` => `is in touch mode`

Also can you provide more info about what this mode means in the comment? I assume the interface is different for smaller screens?

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  /** The input element this datepicker is associated with. */
+  private _inputElementRef: ElementRef;
+
+  constructor(private _dialog: MdDialog, private _overlay: Overlay,
+              private _viewContainerRef: ViewContainerRef, @Optional() private _dir: Dir) {}
+
+  ngOnDestroy() {
+    this.close();
+    this._popupRef.dispose();
```
kara: What happens if it's destroyed before you open the popup? Do you need a check here for `this._popupRef`? I think this will break when navigating to another comp in the demo app.

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /** Create the popup. */
+  private _createPopup(): void {
+    const positionStrategy = this._overlay.position().connectedTo(this._inputElementRef,
+        {originX: 'start', originY: 'bottom'}, {overlayX: 'start', overlayY: 'top'});
+
+    const overlayState = new OverlayState();
+    overlayState.positionStrategy = positionStrategy;
+    overlayState.width = 300;
```
kara: Move up into a constant? Also maybe comment on where this number comes from.

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+      this._popupRef.attach(this._calendarPortal);
+    }
+
+    this._popupRef.backdropClick().first().subscribe(() => this.close());
+  }
+
+  /** Create the popup. */
+  private _createPopup(): void {
+    const positionStrategy = this._overlay.position().connectedTo(this._inputElementRef,
+        {originX: 'start', originY: 'bottom'}, {overlayX: 'start', overlayY: 'top'});
```
kara: Nit: Might be nicer if you broke this out into another private method, `_getPosition()`, then you could use it inline in line 116.

src/lib/datepicker/datepicker-input.ts
```diff
@@ -0,0 +1,20 @@
+import {Directive, ElementRef, Input} from '@angular/core';
+import {MdDatepicker} from './datepicker';
+
+
+/** Directive used to connect an input to a MdDatepicker. */
+@Directive({
+  selector: 'input[mdDatepicker]',
```
kara: Add `input[matDatepicker]`?

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+import {Dir} from '../core/rtl/dir';
+import {MdError} from '../core/errors/error';
+import {MdDialog} from '../dialog/dialog';
+import {MdDialogRef} from '../dialog/dialog-ref';
+
+
+/** Component responsible for managing the datepicker popup/dialog. */
+@Component({
+  moduleId: module.id,
+  selector: 'md-datepicker',
```
kara: Add `mat-datepicker`

src/lib/datepicker/datepicker-input.ts
```diff
@@ -0,0 +1,20 @@
...
+})
+export class MdDatepickerInput {
+  @Input()
+  set mdDatepicker(value: MdDatepicker) {
```
kara: You'll probably need another `@Input()` for `matDatepicker`, right?

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+      throw new MdError('An MdDatepicker can only be associated with a single input.');
+    }
+    this._inputElementRef = inputElementRef;
+  }
+
+  /**
+   * Open the calendar.
+   * @param touchUi Whether to use the touch UI.
+   */
+  open(touchUi = false) {
```
kara: Return types, esp for public methods

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (!this._calendarPortal) {
+      this._calendarPortal = new TemplatePortal(this.calendarTemplate, this._viewContainerRef);
+    }
+
+    this.touchUi = touchUi;
+    touchUi ? this._openAsDialog() : this._openAsPopup();
+  }
+
+  /** Close the calendar. */
+  close () {
```
kara: Nit: Whitespace should be `close()`

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,124 @@
...
+  ngOnDestroy() {
+    this.close();
+    this._popupRef.dispose();
+  }
+
+  /**
+   * Register an input with this datepicker.
+   * @param inputElementRef An ElementRef for the input.
+   */
+  registerInput(inputElementRef: ElementRef) {
```
kara: Do you want this to be public for users instead of directive?  Otherwise, maybe `_registerInput`.

src/demo-app/datepicker/datepicker-demo.html
```diff
@@ -5,3 +5,8 @@
 
 <br>
 <div>{{selected?.toNativeDate()}}</div>
+
+<input [mdDatepicker]="dp">
+<button (click)="dp.open()">open</button>
+<button (click)="dp.open(true)">open (mobile)</button>
```
kara: What `true` means is kind of opaque when looking at this. Have you considered another method that makes it more clear what is happening? `openAsDialog()` ?

src/lib/datepicker/datepicker.ts
mmalerba: I don't really see the benefit. It's not reused anywhere, I'd essentially be using the method as a variable, so why not just use a variable?

src/lib/datepicker/datepicker.ts
mmalerba: This is just a placeholder until I put the real content in. I'll add a comment though

src/lib/datepicker/datepicker.ts
kara: It's a nit, so you can take it or leave it. But I think it would help with readability. I've noticed with other overlay components that the position strategy tends to grow and bloat the method (when you add fallback position support, etc).

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,159 @@
...
+@Component({
+  moduleId: module.id,
+  selector: 'md-datepicker, mat-datepicker',
+  templateUrl: 'datepicker.html',
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MdDatepicker implements OnDestroy {
+  /**
+   * Whether the calendar UI is in touch mode. In touch mode the calendar opens in a dialog rather,
```
kara: Typo: remove comma at end

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      expect(popup).not.toBeNull();
+      expect(parseInt(getComputedStyle(popup).height)).not.toBe(0);
+
+      testComponent.datepicker.close();
+      fixture.detectChanges();
+
+      expect(parseInt(getComputedStyle(popup).height)).toBe(0);
+    });
+
+    it('openTouchUi should open dialog', () => {
```
kara: Typo: `should close dialog`

src/lib/datepicker/datepicker.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  /**
+   * Register an input with this datepicker.
+   * @param inputElementRef An ElementRef for the input.
+   */
+  _registerInput(inputElementRef: ElementRef): void {
+    if (this._inputElementRef) {
+      throw new MdError('An MdDatepicker can only be associated with a single input.');
```
kara: Add a test for this?

PR #3036, opened 2017-02-10
refactor: add themeable base class

### LINKED ISSUE #2394: Share methods to set the theme color classes
We need to find a way to share the methods, which apply the theme color classes on the components.

> Right now we are repeating the same methods for each component.

* Also we should have a type for the color values like `primary | accent | warn`

DevVersion:
* Introduces a new `MdThemeable` base class that can be extended by different components to automatically support the `color` input.

* This reduces a lot of repeated code in the different components and it also simplifies maintaining.

**Note**: This depends on https://github.com/angular/material2/pull/3017 and https://github.com/angular/dgeni-packages/pull/207 but should be ready for review though.

Closes #2394.

### DIFF
--- a/src/lib/core/style/themeable.ts
+++ b/src/lib/core/style/themeable.ts
@@ -1,0 +1,62 @@
+import {ElementRef, Input, Renderer} from '@angular/core';
+import {MdError} from '../errors/error';
+
+/** Possible color values for the color input. */
+export type MdThemeColor = 'primary' | 'accent' | 'warn';
+
+const VALID_COLOR_VALUES = ['primary', 'accent', 'warn'];
+
+/**
+ * Material components can extend the MdThemeable class to add an Input that can
+ * developers use to switch palettes on the components.
+ **/
+export class MdThemeable {
+
+  /** Stored color for the themeable component. */
+  private _color: MdThemeColor;
+
+  // Constructor initializers need to have the `protected` modifier to avoid interferences.
+  // TypeScript throws if similar declarations, regardless of the modifier, have been
+  // found across the different classes. Because of that, the child classes should just use
+  // the protected properties from the superclass.
+  // TODO(devversion): revisit this once TypeScript v2.2.1 is being used.
+  constructor(protected _renderer: Renderer, protected _elementRef: ElementRef) {}
+
+  /** Color of the component. Values are primary, accent, or warn. */
+  @Input()
+  get color(): MdThemeColor {
+    return this._color;
+  }
+  set color(newColor: MdThemeColor) {
+    this._validateColor(newColor);
+
+    this._setElementColor(this._color, false);
+    this._setElementColor(newColor, true);
+    this._color = newColor;
+  }
+
+  /** Validates the specified color value and throws an error if invalid. */
+  private _validateColor(color: string) {
+    if (color && VALID_COLOR_VALUES.indexOf(color) === -1) {
+      throw new MdInvalidColorValueError(color);
+    }
+  }
+
+  /** Toggles a color class on the components host element. */
+  private _setElementColor(color: string, isAdd: boolean) {
+    if (color) {
+      this._renderer.setElementClass(this._elementRef.nativeElement, `mat-${color}`, isAdd);
+    }
+  }
+
+}
+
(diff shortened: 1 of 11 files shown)


src/lib/core/style/themeable.ts
```diff
@@ -0,0 +1,30 @@
+import {ElementRef, Input, Renderer} from '@angular/core';
+
+/** Material components can extend the MdThemeable class to add an Input that can
```
mmalerba: nit:
```
/**
 * Text text ...
 * Text text ...
 */
```

src/lib/core/style/themeable.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+  set color(newColor: string) {
+    this._setElementColor(this._color, false);
+    this._setElementColor(newColor, true);
+    this._color = newColor;
+  }
+
+  /** Toggles a color class on the components host element. */
+  private _setElementColor(color: string, isAdd: boolean) {
+    if (color != null && color != '') {
```
mmalerba: if (color)?

src/lib/core/style/themeable.ts
```diff
@@ -0,0 +1,30 @@
+import {ElementRef, Input, Renderer} from '@angular/core';
+
+/** Material components can extend the MdThemeable class to add an Input that can
+ * developers use to switch palettes on the components. */
+export class MdThemeable {
+
+  /** Stored color for the themeable component. */
+  private _color: string;
```
mmalerba: 'primary' | 'accent' | 'warn'

src/lib/core/style/themeable.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(private renderer: Renderer, private elementRef: ElementRef) {}
```
mmalerba: _ on privates

src/lib/core/style/themeable.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  /** Color of the component. Values are primary, accent, or warn. */
+  @Input()
+  get color(): string {
+    return this._color;
+  }
+  set color(newColor: string) {
```
mmalerba: validate one of 'primary', 'accent', 'warn', null

src/lib/core/style/themeable.ts
DevVersion: Yeah makes sense. While doing that I ran into an issue with the child classes.

```
Types have separate declarations of a private property '_elementRef'.
```

Because of that the best solution is to just use the `protected` modifier here, and the child classes can just use the variables similar as before. (Just remove the modifiers in the constructor)

DevVersion:
Thanks for the feedback @mmalerba. Made the requested changes.

src/lib/core/style/themeable.ts
mmalerba: Ah I see, another option would be to do this in themeable:

```
constructor(private _themeableRenderer: Renderer, private _themeableElementRef: ElementRef)
```

mmalerba:
why is this blocked?

DevVersion:
Because we need the Dgeni fix to be ready. Otherwise our docs will not show the `color` input properly.

mmalerba:
dependent pr is merged, feel free to rebase and apply merge-ready

mmalerba:
Can we mark this as merge ready now?

kara:
I think @jelbourn should take a look, given how cross-cutting this is.

DevVersion:
@mmalerba Unfortunately there are still some issues with Dgeni. I need to look into this again. I wish Dgeni would not use TypeScript 1.7..

jelbourn:
So, the high level problem here is that once you make all of the components extend `MdThemable`, they can never extend anything else. My thinking is that composing mixins for behaviors like this is a  much better approach. Fortunately, a new feature in TypeScript 2.2 might make this way easier for us: https://github.com/Microsoft/TypeScript/pull/13743

While we can't switch to 2.2 just yet we should be able to get there before too long.

DevVersion:
@jelbourn Sounds good. Interested in how Dgeni will handle the mixins though.

jelbourn:
@DevVersion I'm sure it won't :)

PR #3038, opened 2017-02-10
fix(tabs): missing mat class on tab group

andrewseguin:


andrewseguin:
This change causes some IE issues. Need to investigate what's going on here that would mess them up

PR #3053, opened 2017-02-11
Revert "feat(tooltip): reposition on scroll"

kara:
Reverts angular/material2#2703

This is temporary solution while we re-work when the scroll dispatcher subscribes.


kara:
Found another solution

PR #3068, opened 2017-02-12
feat(slide-toggle): use ripple service

### LINKED ISSUE #2900: slide-toggle: use ripple service instead of temporary ripple
Right now, using Chrome (Version 56.0.2924.87 (64-bit)), when I click on the md-slide-toggle, the glow effect seems to lock in place, rather than degrade gently to zero. Currently the only way to get rid of the glow on a clicked md-slide-toggle is to click on another element, or click somewhere else on the page. Is this a bug - or the desired effect? Seems like a gentle fade to zero after click would work better. Secondly, is there a way to disable the glow effect?

DevVersion:
* Removes the temporary ripple mixin and uses the new ripple service.
* Updates the ripple colors to match the specification images (no explicit opacity values are specified)
* Triggers ripple on label mousedown (similar as in the Android Nougat implementation)

@mmalerba I intentionally created separate variables for the `checked` ripple colors because I want to have all spec-aligned stuff at the top of the `@mixin` so we can easily make changes. 

Closes #2900.

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.scss
+++ b/src/lib/slide-toggle/slide-toggle.scss
@@ -10,13 +10,8 @@ $mat-slide-toggle-bar-height: 14px !default;
 $mat-slide-toggle-thumb-size: 20px !default;
 $mat-slide-toggle-margin: 16px !default;
 $mat-slide-toggle-spacing: 8px !default;
+$mat-slide-toggle-ripple-radius: 23px !default;
 
-
-@mixin mat-switch-ripple() {
-  // Temporary ripple effect for the thumb of the slide-toggle.
-  // Bind to the parent selector and specify the current palette.
-  @include mat-temporary-ink-ripple(slide-toggle, true);
-}
 
 .mat-slide-toggle {
   display: flex;
@@ -38,8 +33,6 @@ $mat-slide-toggle-spacing: 8px !default;
       transform: translate3d(100%, 0, 0);
     }
   }
-
-  @include mat-switch-ripple();
 
   &.mat-disabled {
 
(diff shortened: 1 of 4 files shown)


src/lib/slide-toggle/slide-toggle.scss
```diff
@@ -169,3 +162,15 @@ $mat-slide-toggle-spacing: 8px !default;
...
+.mat-slide-toggle-ripple {
+  position: absolute;
+  top: $mat-slide-toggle-thumb-size / 2 - $mat-slide-toggle-ripple-radius;
+  left: $mat-slide-toggle-thumb-size / 2 - $mat-slide-toggle-ripple-radius;
+  height: $mat-slide-toggle-ripple-radius * 2;
+  width: $mat-slide-toggle-ripple-radius * 2;
+  border-radius: 50%;
+  z-index: 1;
+  pointer-events: none;
+}
```
mmalerba: You should check if your editor has an option to add \n on save :p I think most editors have some type of option for this.

dahaupt: @mmalerba Why is it useful to have a newline at the end of each file? Or is this just convention? Always wondered about that 😄

DevVersion: Haha. That's a good idea. I've had so many lint issues with that - Thanks! 😄

mmalerba: @dahaupt good question, I never thought about why that's the convention we use, but apparently that's just how lines are defined: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

PR #3071, opened 2017-02-12
Fix theming link

jmandel:


dahaupt:
@jmandel I can't see a problem here. It's working fine and your change will break it.

jmandel:
Hmm. To reproduce:

1. Visit https://material.angular.io/guide/getting-started
2. See 
image
3. See "theming guide" that links to https://material.angular.io/guides/theming.md
4. Click "theming guide"
5. Page not found --> redirects to https://material.angular.io/

What do you see?

jmandel:
Ah, this should be filed on https://github.com/angular/material

dahaupt:
@jmandel Ah I can see your problem. I think it has to be fixed here: https://github.com/angular/material.angular.io

@DevVersion What do you think?

DevVersion:
Yeah this should have been fixed in https://github.com/angular/material2/commit/60f03ed7dd0a1fca798abd07f7352cd186330859.

Unfortunately the docs haven't been deployed yet. So everything probably works when beta.2 is released (*soon*)

Issue #3074, opened 2017-02-12
Unit tests process crashes


dahaupt:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
The unit tests should work on master branch.

#### What is the current behavior?
The process fails in step `:build:components:spec` with:
```
node_modules/@types/jasmine/index.d.ts(39,38): error TS2371: A parameter initializer is only allowed in a function or constructor implementation.
node_modules/@types/jasmine/index.d.ts(39,46): error TS2304: Cannot find name 'keyof'.
node_modules/@types/jasmine/index.d.ts(39,52): error TS1005: '=' expected.
```

#### What are the steps to reproduce?
Run `gulp test` on the master branch.

#### Which versions of Angular, Material, OS, browsers are affected?
Latest Version of Material, Windows 10, Angular 2.4.7

#### Is there anything else we should know?
I'm no expert in unit tests, but simply starting them on a new cloned repository should work (e2e tests are working fine). It could be possible too, that this is just a problem on my system, because otherwise someone would have already noticed this.

modernfidelity:
I'll also getting this via ->  gulp build:release  at the "Starting ':build:components:ngc'" step

kara:
Running tests works for me on master. What version of node are you running?  I'm running 6.2.2.

dahaupt:
@kara I'm running version v6.9.5 of node.

trshafer:
I have the same issue. I'm running 6.9.4, as that is what travis is running on: https://github.com/angular/material2/blob/master/.travis.yml#L5

kara:
How about Typescript versions?

dahaupt:
My global Typescript version is 2.1.5, in the project (master) it is 2.0.10. I'm not sure which one is used for tests.

trshafer:
Did a git bisect of "npm install && npm test". This is the commit where it changed from a warning to an error 4a4261e141c23ad3a14fe5869498dcde2640db6b.

kara:
Okay, just blew away my node modules and started from scratch. Also getting the error now.

modernfidelity:
Solved this by reducing the @types/jasmine to 2.5.38

owlstack:
I'm having the same issue. Is this resolved? should I wait for latest release?

trshafer:
@owlstack, are you still getting the issue with #3080 merged in? Did you remove your node_modules folder and re install?

Slashmsu:
[...]

PR #3078, opened 2017-02-13
fix(menu,tooltip): Ensure subscription exists before unsubscribing.

trshafer:


trshafer:
I'm having issues running the tests, I receive this error:
```
Finished ':build:test:vendor' after 1.22 s
node_modules/@types/jasmine/index.d.ts(39,38): error TS2371: A parameter initializer is only allowed in a function or constructor implementation.
node_modules/@types/jasmine/index.d.ts(39,46): error TS2304: Cannot find name 'keyof'.
node_modules/@types/jasmine/index.d.ts(39,52): error TS1005: '=' expected.
':build:components:spec' errored after 4.08 s
```

dahaupt:
@trshafer I've already opened up an issue for that. (#3074)

trshafer:
I have added tests. Please take a look.

kara:
I'm actually not sure we should change this. The scroll subscription and the tab subscription will always exist once the menu / tooltip are created. The reason the unit test can fail is because `fixture.detectChanges()` is not being called. This means change detection doesn't run and none of the init hooks are executed as they would in a real environment.

In this case, I think the test *should* fail. Testing menu or tooltip without allowing init hooks to run is inaccurate and could cause harder to debug test failures down the line.  @mmalerba, thoughts?

mmalerba:
@kara Yep agreed, I didn't realize that in the normal lifecycle init was guaranteed to be called before destory. In that case i agree that the real issue here is that the tests are not simulating the true lifecycle of the component.

trshafer:
Not having this code made our app more difficult to debug. We had an error when rendering, and the error was hidden by not checking for these subscriptions. a "cannot call .unsubscribe on undefined" error was blocking our actual application error.

trshafer:
Other files have checks that the subscription is created: https://github.com/angular/material2/blob/master/src/lib/select/select.ts#L271
https://github.com/angular/material2/blob/master/src/lib/core/overlay/overlay-directives.ts#L272

kara:
If anything, I think those other checks should probably be removed for the same reason. Can you give us more information about hiding errors though? Typically they'd be thrown first.

trshafer:
The issue is that one of our components threw an error on another init. The tooltip init had not run but had been constructed. The tooltip component was destroyed before the component which threw the error, so the exception which we received as in the tooltip instead of in our component. This made our application harder to debug.

There are other examples in angular core:
https://github.com/angular/angular/blob/master/modules/%40angular/platform-server/src/http.ts#L70
https://github.com/angular/angular/blob/master/modules/%40angular/router/src/router.ts#L329

There are also examples of them not checking if the subscription exists.

kara:
OK, we don't want to be hiding errors. In that case, LGTM.

PR #3082, opened 2017-02-13
fix(dialog): escape key not working once element loses focus

### LINKED ISSUE #3009: Closing Dialog using escape only works after initial open w/ button focused
Bug

Dialog should close when escape is pressed, regardless of what is in focus.

If the user interacts with the dialog in any way (causing the initial focus to be lost), pressing escape no longer closes the dialog.

* Go to https://material.angular.io/components/component/dialog
* Launch the dialog
* Click anywhere inside the dialog besides the buttons, causing the first button to lose focus
* Press escape -- the dialog does not close

See: https://material.angular.io/components/component/dialog

A user should be able to press escape to close the dialog at any time.

Seemingly all

No

crisbeto:
Fixes issue in the dialog that prevented the user from being able to close via the escape key, if any of the elements inside the dialog loses focus.

Fixes #3009.

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -3,6 +3,7 @@ import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 import {Overlay, OverlayRef, ComponentType, OverlayState, ComponentPortal} from '../core';
 import {extendObject} from '../core/util/object-extend';
+import {ESCAPE} from '../core/keyboard/keycodes';
 import {DialogInjector} from './dialog-injector';
 import {MdDialogConfig} from './dialog-config';
 import {MdDialogRef} from './dialog-ref';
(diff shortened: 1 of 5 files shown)


src/lib/dialog/dialog.ts
```diff
@@ -65,6 +67,10 @@ export class MdDialog {
     let dialogRef =
         this._attachDialogContent(componentOrTemplateRef, dialogContainer, overlayRef, config);
 
+    if (!this._openDialogs.length && !this._parentDialog) {
+      document.addEventListener('keydown', this._boundKeydown);
```
mmalerba: I guess we don't have access to a Renerer we can use to do this at this point in the code?

crisbeto: We don't. Angular only injects the renderer in components.

realappie:
Wouldnt it have been better to use HostListener instead? 

`@HostListener('document.key.Escape', ['$event']) escapeFired { this.handleEscapeKey() }`

This would attach a listener on the document rather than the element, thus preventing this bug. 

I am unsure why the MD team prefers the use of `host` over `HostListener`. `HostListener` is actually never used in the codebase. This is a perfect use case for HostListener, and I am curious on why you decided to add listeners to the document manually instead.

PR #3086, opened 2017-02-13
feat(datepicker): make calendar responsive

mmalerba:
calendar cells expand to fill available width, but maintain their 1:1 aspect ratio


src/lib/datepicker/calendar-table.scss
```diff
@@ -1,40 +1,61 @@
 $mat-calendar-table-font-size: 12px !default;
-$mat-calendar-table-cell-padding: 1px !default;
-$mat-calendar-table-cell-content-size: 32px !default;
+$mat-calendar-table-label-padding-start: 5% !default;
+$mat-calendar-table-label-translation: -.5em;
```
tinayuangao: very nit: why do we use em and px at the same time? Can we convert it to px?

mmalerba:
done

tinayuangao:
LGTM

PR #3091, opened 2017-02-14
fix(scripts): change all paths in docs pushing script to relative paths

riavalon:


scripts/release/publish-docs-content.sh
```diff
@@ -22,7 +22,7 @@ commitAuthorEmail="$(git --no-pager show -s --format='%ae' HEAD)"
 commitMessage="$(git log --oneline -n 1)"
 
 # create directory and clone test repo
-rm -rf /tmp/
+rm -rf ./tmp/
```
DevVersion: Keep `/tmp` and just call `rm -rf $repoPath`.

scripts/release/publish-docs-content.sh
```diff
@@ -5,10 +5,10 @@
 
 cd "$(dirname $0)/../../"
 
-docsPath="/dist/docs"
-repoPath="/tmp/material2-docs-content"
+docsPath="./dist/docs"
+repoPath="./tmp/material2-docs-content"
```
DevVersion: This should be still /tmp. Just be sure to don't remove the whole `/tmp` dir later.

PR #3093, opened 2017-02-14
update: add button to toggle theme in demo-app

DevVersion:
* Adds a button to the demo-app toolbar that allows developers to quickly change between dark and light Material themes.
 
  This is very useful when building new components or just when confirming that all components work with dark themes.

**Note**: Most of the custom backgrounds in the demo-app have been removed because those make the docs look weird in dark themes. Those are not important for testing & design and didn't even make the docs look better.

**For better review**: https://github.com/angular/material2/pull/3093/files?w=1 (Hide whitespace)


src/demo-app/demo-app/demo-app.scss
```diff
@@ -1,3 +1,19 @@
+// The default indigo-pink theme will load through the index file. This makes the SCSS build faster
+// since Sass doesn't need to build a "prebuilt" theme again.
+@import '../../lib/core/theming/all-theme';
```
andrewseguin: Not sure how to make this work, the demo app cannot reach into the lib directory directly. Would be best to have 

```
@import '../@angular/material/core/theming/all-theme';
```

but the sass compiler expects this to be present during build but it is only there after build. Any other way to import this file dynamically?

DevVersion: When I ran it locally it worked as expected? Sass imports the file at build-time and doesn't compile with the `dist/` folder as CWD.

kara: We should try to keep our demo app as close as possible to how user apps would be configured. So I agree that we should avoid drilling down into the lib folder.  We should be able to use the top level `@angular/material` import if we tweak the `includePaths` option in our sass build config to include the dist folder.

src/demo-app/demo-app/demo-app.scss
```diff
@@ -1,3 +1,19 @@
+// The default indigo-pink theme will load through the index file. This makes the SCSS build faster
```
kara: Can we actually pull this out into a custom theme file and lose the prebuilt altogether? I see your point with saving build time, but dropping in here makes the theme code harder to find. This is supposed to be an example, so it might confuse people as is.

DevVersion: Fine by me. Just set the focus on performance here. Will change!

src/demo-app/demo-app/demo-app.scss
DevVersion: Alright. I will look into it.

DevVersion:
@kara @andrewseguin Addressed your feedback. Everything works fine with `includePaths`.

mmalerba:
FYI I had to revert this because the @import '@angular/material/...' doesn't play nice in google3. We'll try to figure something out, in the mean time reverted so I can continue syncing other PRs.

PR #3097, opened 2017-02-14
feat(snack-bar): Optionally pass action value through MdSnackBarRef onAction observable

kyle-burke:
Allows for passing the value of a snack bar action through the onAction observable so multiple actions can be passed to the snack bar's hosted component and identified when clicked. This makes snack bars with multiple actions possible as devs can now identify which of multiple actions have been clicked.

I've added a demo to the `/src/examples` directory in hopes that this will be added to the material.angular.io docs knowledge base (wasn't sure how to render non-Typedoc documentation locally, so let me know if I set up something wrong there).

I also updated the SimpleSnackBar component to pass the given action into its dismiss call.

Subscribers to MdSnackBarRef's onAction observable will receive an update with an optional argument that describes which snack bar action has been selected.

Currently, the onAction observable only indicates *that* a snack bar action has been triggered, but no indication of which one.

I needed a snack bar which allowed for multiple actions for the user to choose from, and a way to identify which action the user clicked. This seemed to be the simplest implementation, which still requires a custom component to render a snack bar with multiple action buttons. I am interested in eventually updating the signature for `MdSnackBar.open` to allow an array of action strings to be passed to the default snack bar if this PR is accepted.

This is my first open source contribution of any sort, and I'm very nervous.

### DIFF
--- a/src/examples/example-module.ts
+++ b/src/examples/example-module.ts
@@ -36,6 +36,10 @@ import {
 import {ListSectionsExample} from './list-sections/list-sections-example';
 import {SnackBarOverviewExample} from './snack-bar-overview/snack-bar-overview-example';
 import {
+  SnackBarMultipleActionsExample,
+  MultipleActionsComponent
+} from './snack-bar-multiple-actions/snack-bar-multiple-actions-example';
+import {
   DialogResultExampleDialog,
   DialogResultExample
 } from './dialog-result/dialog-result-example';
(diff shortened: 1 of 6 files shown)


kyle-burke:
@crisbeto @DevVersion Anything I can do to keep this moving? I don't know if support for multiple actions are part of the snack bar spec, but this would save our team a lot of code basically reimplementing snack bars.

josephperrott:
Multiple buttons on snackbars is explicitly against Material Design spec.

Per [the spec](https://material.io/guidelines/components/snackbars-toasts.html): "For two or more actions, use a dialog, not a snackbar."  Because of this requirement of the spec, you should always know the purpose of the button being pressed since there is only one action.  For this reason, we do not allow multiple actions to be sent into the SimpleSnackBar.

With reference to the exact case you brought up, wanting to consistently have multiple actions on a snackbar, you could create a custom service that you have in your NgModule that uses the snackbar service the same way the snackbar service uses SimpleSnackBar.  This would prevent you from reimplementing all of snackbar, but make a reusable service for your use case.

josephperrott:
Closing this as we are not supporting multiple actions in the SimpleSnackBar.  If you would need a snack bar with multiple actions you can create a custom component and use it with the  `openFromComponent`  method.

PR #3113, opened 2017-02-15
feat(FocusOriginMonitor): support monitoring subtree focus as well as element

mmalerba:
Also includes some cleanup:
* rename `registerElementForFocusClasses` to `monitor`
* add `unmonitor` method and call on directive destroyed
* do all listening outside angular
* replace `cdkFocusClasses` with `cdkMonitorElementFocus` and `cdkMonitorSubtreeFocus`

### DIFF
--- a/src/lib/core/style/focus-classes.ts
+++ b/src/lib/core/style/focus-classes.ts
@@ -1,6 +1,6 @@
 import {
   Directive, Injectable, Optional, SkipSelf, Renderer, ElementRef,
-  OnDestroy, NgZone
+  OnDestroy, NgZone, EventEmitter, Output
 } from '@angular/core';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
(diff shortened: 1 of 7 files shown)


src/lib/core/style/focus-classes.ts
```diff
@@ -29,66 +40,143 @@ export class FocusOriginMonitor {
...
+    if (this._elementInfo.has(element)) {
+      let info = this._elementInfo.get(element);
+      info.includesChildren = includeChildren;
+      return info.subject.asObservable();
+    }
+
+    // Create monitored element info.
+    let info: MonitoredElementInfo = {
+      unlisten: null,
+      includesChildren: includeChildren,
```
DevVersion: How about `checkChildren`?

src/lib/core/style/focus-classes.ts
```diff
@@ -29,66 +40,143 @@ export class FocusOriginMonitor {
...
+        }
+        this._lastTouchTarget = event.target;
+        this._touchTimeout = setTimeout(() => this._lastTouchTarget = null, TOUCH_BUFFER_MS);
+      }, true);
+
+      // Make a note of when the window regains focus, so we can restore the origin info for the
+      // focused element.
+      window.addEventListener('focus', () => {
+        this._windowFocused = true;
+        setTimeout(() => this._windowFocused = false, 0);
```
DevVersion: Just an optional nit. How about extracting this into its own function?

```ts
constructor(ngZone: NgZone) {
  ngZone.runOutsideAngular(() => this._registerDocumentEvents());
}
```

This way we would get rid of unnecessary indentation and also have it pretty clean.

src/lib/core/style/focus-classes.ts
```diff
@@ -111,10 +202,24 @@ export class FocusOriginMonitor {
...
+   */
+  private _onFocus(event: FocusEvent, element: Element) {
+    // NOTE(mmalerba): We currently set the classes based on the focus origin of the most recent
+    // focus event affecting the monitored element. If we want to use the origin of the first event
+    // instead we should check for the cdk-focused class here and return if the element already has
+    // it. (This only matters for elements that have includesChildren = true).
+
+    // If we are not counting child-element-focus as focused, make sure that the event target is the
+    // monitored element itself.
+    if (!this._elementInfo.get(element).checkChildren && element != event.target) {
```
DevVersion: I think that we should use `===` as often as possible. Also in the `setClasses` method.

PR #3119, opened 2017-02-16
feat(button-toggle): Add a focus overlay for button-toggle

### LINKED ISSUE #3014: Focus styling and hiding for mat-button-toggle
Implement focus indication for md-button-toggle, similar to md-button.

Several form elements such as md-button have a md-button-focus class bound to a function that detects whether the element has focus and it's appropriate to show the focus indicator. (This doesn't just style the :focus pseudoselector because an element can have focus even when the focus indicator should not be shown.) md-button-toggle should implement this similarly.

md-button-toggle does not implement md-button-focus, nor does it have focus styling. I believe the current behavior is to not indicate focus for toggle buttons or button groups at all. Focus itself is handled correctly: exclusive button groups take focus, as do individual toggles alone and in multi-select groups.

Simple demo of the different toggle buttons, where you can test keyboard focus behavior alongside other widgets:
  http://plnkr.co/edit/uLky5EnpTukImH7yMZOO?p=preview

I'm not sure if focus behaviors are specified by Material for toggle buttons. Without guidance to the contrary, I would assume that focus indication for toggles is appropriate and would be similar to md-buttons. A lack of focus indication is a poor experience.

All browsers.

Unlike action buttons, toggle buttons may need to indicate 1) toggle state, 2) focus, 3) active state, and 4) hover state, all with background color changes. My project has decided to represent toggle and focus using shading levels, where toggle+focus gets a double shade, just toggle and just focus are identical (for better and for worse), and active and hover override toggle/focus and use a different color. I don't know what'd be appropriate for the Material library default. I can't implement a custom focus style correctly without md-button-focus support. (The library could implement md-button-focus without taking a position on styling. :) )

tinayuangao:
Fixes #3014

### DIFF
--- a/src/lib/button-toggle/_button-toggle-theme.scss
+++ b/src/lib/button-toggle/_button-toggle-theme.scss
@@ -3,10 +3,10 @@
 
 // Applies a focus style to an md-button-toggle element for each of the supported palettes.
 @mixin _mat-button-toggle-focus-color($theme) {
-  $foreground: map-get($theme, foreground);
+  $background: map-get($theme, background);
 
   .mat-button-toggle-focus-overlay {
-    background-color: mat-color($foreground, divider);
+    background-color: mat-color($background, dark-overlay);
   }
 }
 
--- a/src/lib/button-toggle/_button-toggle-theme.scss
+++ b/src/lib/button-toggle/_button-toggle-theme.scss
@@ -1,6 +1,24 @@
 @import '../core/theming/palette';
 @import '../core/theming/theming';
 
+// Applies a focus style to an md-button-toggle element for each of the supported palettes.
+@mixin _mat-button-toggle-focus-color($theme) {
+  $primary: map-get($theme, primary);
+  $accent: map-get($theme, accent);
+  $warn: map-get($theme, warn);
+
+  &.mat-primary .mat-button-toggle-focus-overlay {
+    background-color: mat-color($primary, 0.12);
+  }
+
+  &.mat-accent .mat-button-toggle-focus-overlay {
+    background-color: mat-color($accent, 0.12);
+  }
+
+  &.mat-warn .mat-button-toggle-focus-overlay {
+    background-color: mat-color($warn, 0.12);
+  }
+}
 
 @mixin mat-button-toggle-theme($theme) {
   $foreground: map-get($theme, foreground);
@@ -8,6 +26,10 @@
 
   .mat-button-toggle {
     color: mat-color($foreground, hint-text);
+
+    &.cdk-focused {
+      @include _mat-button-toggle-focus-color($theme);
+    }
   }
 
   .mat-button-toggle-checked {
--- a/src/lib/button-toggle/_button-toggle-theme.scss
+++ b/src/lib/button-toggle/_button-toggle-theme.scss
@@ -3,20 +3,10 @@
 
 // Applies a focus style to an md-button-toggle element for each of the supported palettes.
 @mixin _mat-button-toggle-focus-color($theme) {
-  $primary: map-get($theme, primary);
-  $accent: map-get($theme, accent);
-  $warn: map-get($theme, warn);
+  $foreground: map-get($theme, foreground);
 
-  &.mat-primary .mat-button-toggle-focus-overlay {
-    background-color: mat-color($primary, 0.12);
-  }
-
-  &.mat-accent .mat-button-toggle-focus-overlay {
-    background-color: mat-color($accent, 0.12);
-  }
-
-  &.mat-warn .mat-button-toggle-focus-overlay {
-    background-color: mat-color($warn, 0.12);
+  .mat-button-toggle-focus-overlay {
+    background-color: mat-color($foreground, divider);
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/lib/button-toggle/button-toggle.ts
```diff
@@ -461,10 +468,31 @@ export class MdButtonToggle implements OnInit {
     event.stopPropagation();
   }
 
+  _onInputFocus() {
+    // Only show the focus / ripple indicator when the focus was not triggered by a mouse
```
mmalerba: maybe you could use the FocusOriginMonitor for this? https://github.com/angular/material2/blob/master/src/lib/core/style/focus-classes.ts

src/lib/button-toggle/button-toggle.scss
```diff
@@ -47,3 +53,27 @@ $mat-button-toggle-border-radius: 2px !default;
...
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
+
+  @include cdk-high-contrast {
+    // Note that IE will render this in the same way, no
+    // matter whether the theme is light or dark. This helps
+    // with the readability of focused buttons.
+    background-color: rgba(white, 0.5);
```
mmalerba: should be part of the theme file

src/lib/button-toggle/button-toggle.scss
```diff
@@ -47,3 +53,27 @@ $mat-button-toggle-border-radius: 2px !default;
...
   vertical-align: middle;
 }
+
+// Overlay to be used as a tint. Note that the same effect can be achieved by using a pseudo
+// element, however we use a proper DOM element in order to be able to disable the default
+// touch action. This makes the buttons more responsive on touch devices.
+.mat-button-toggle-focus-overlay {
+  // The button spec calls for focus on raised buttons (and FABs) to be indicated with a
+  // black, 12% opacity shade over the normal color (for both light and dark themes).
+  background-color: rgba(black, 0.12);
```
mmalerba: should this be in the theme file?

src/lib/button-toggle/button-toggle.html
```diff
@@ -5,10 +5,14 @@
...
+         (focus)="_onInputFocus()"
          (change)="_onInputChange($event)"
          (click)="_onInputClick($event)">
 
   <div class="mat-button-toggle-label-content">
     <ng-content></ng-content>
   </div>
 </label>
+<!-- the touchstart handler prevents the overlay from capturing the initial tap on touch devices -->
+<div class="mat-button-toggle-focus-overlay" (touchstart)="$event.preventDefault()"></div>
```
mmalerba: why do we want to disable the initial tap?

src/lib/button-toggle/button-toggle.ts
tinayuangao: Thanks for the suggestion!

src/lib/button-toggle/button-toggle.scss
tinayuangao: I think only theme related should be in theme file. Default is black which will not change with the theme.

Added theme color in the theme file.

src/lib/button-toggle/button-toggle.html
tinayuangao: This is the same as https://github.com/angular/material2/blob/master/src/lib/button/button.html

src/lib/button-toggle/button-toggle.scss
mmalerba: Ah ok, I missed that these are the same for light & dark themes

kara: Can you replace this with a color from the _palettes file though? If one doesn't exist in $background, we can create one.

src/lib/button-toggle/button-toggle.scss
```diff
@@ -47,3 +53,27 @@ $mat-button-toggle-border-radius: 2px !default;
...
+  border-radius: inherit;
+  pointer-events: none;
+  opacity: 0;
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
```
kara: Use the `mat-fill` mixin here?

src/lib/button-toggle/button-toggle.spec.ts
```diff
@@ -14,13 +14,15 @@ import {
...
 } from './button-toggle';
+import {FocusOriginMonitor} from '../core';
 
 
 describe('MdButtonToggle', () => {
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       imports: [MdButtonToggleModule.forRoot(), FormsModule, ReactiveFormsModule],
+      providers: [FocusOriginMonitor],
```
kara: Should this be added to the button toggle's module definition?

src/lib/button-toggle/_button-toggle-theme.scss
```diff
@@ -1,13 +1,25 @@
 @import '../core/theming/palette';
 @import '../core/theming/theming';
 
+// Applies a focus style to an md-button-toggle element for each of the supported palettes.
+@mixin _mat-button-toggle-focus-color($theme) {
+  $foreground: map-get($theme, foreground);
+
+  .mat-button-toggle-focus-overlay {
+    background-color: mat-color($foreground, divider);
```
kara: Nit: Do you want to add another property as 'dark-overlay' or something? Using a property from "foreground" as a background is a bit confusing out of context.

Issue #3131, opened 2017-02-16
md-card-avatar in card header is broken in beta 2


chpasha:
#### Bug, feature request, or proposal:
bug

#### What is the expected behavior?
the avatar image in card header should be 40 x 40 px

#### What is the current behavior?
it stretches to fit the whole image instead

#### What are the steps to reproduce?

http://plnkr.co/edit/96XQPx0ic3kbxOuOM3Mk?p=preview

#### Which versions of Angular, Material, OS, browsers are affected?
the regression was introduced in beta2 of material lib

#### Is there anything else we should know?
I don't get why it doesn't work, the source could looks ok to me, but the mat-card-avatar class is never applied. applying it manually solves the issue

stefanaerts:
in previous material versions it was this
[md-card-avatar] {
  height: $md-card-header-size;
  width: $md-card-header-size;
  border-radius: 50%;
}
now its this
.mat-card-avatar {
  height: $md-card-header-size;
  width: $md-card-header-size;
  border-radius: 50%;
}

so if you use class="mat-card-avatar" instead of md-card-avatar 
it will work

BhavnaSonawane:
it worked but there is the warning that Can't read file '/src/app/layout/search-candidate/assets/images/5.png', ignoring

PR #3134, opened 2017-02-16
Fix selectors for MdCardXlImage and MdCardAvatar

mattlgy:
I believe the selectors for the MdCardXLImage and MdCardAvatar directives should to attribute selectors like the rest of the card image directives.

Fixes https://github.com/angular/material2/issues/3131

### DIFF
--- a/src/lib/card/card.ts
+++ b/src/lib/card/card.ts
@@ -112,7 +112,7 @@ export class MdCardImage {}
  * Large image used in a card, needed to add the mat- CSS styling.
  */
 @Directive({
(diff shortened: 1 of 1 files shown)


stefanaerts:
in previous material versions it was this
[md-card-avatar] {
  height: $md-card-header-size;
  width: $md-card-header-size;
  border-radius: 50%;
}
now its this
.mat-card-avatar {
  height: $md-card-header-size;
  width: $md-card-header-size;
  border-radius: 50%;
}

so if you use class="mat-card-avatar" instead of md-card-avatar 
it will work

PR #3137, opened 2017-02-16
fix(checkbox): Show checkbox animation only if user click or indeterminate state

### LINKED ISSUE #2783: checkbox animating initial value
I think it should only animate when user toggles the values.. it's especially noticable when you have checkbox inside some ngIf condition(like when it's inside tabs)

http://plnkr.co/edit/8AnsenWT2PORBaC162yu?p=preview

tinayuangao:
Fixes #2783

### DIFF
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -11,6 +11,32 @@ import {dispatchFakeEvent} from '../core/testing/dispatch-events';
 describe('MdCheckbox', () => {
   let fixture: ComponentFixture<any>;
 
+  /** Creates a DOM mouse event. */
+  const createMouseEvent = (eventType: string, dict: any = {}) => {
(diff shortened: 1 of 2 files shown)


mmalerba:
tests seem to be failing in IE

src/lib/checkbox/checkbox.spec.ts
```diff
@@ -10,6 +10,32 @@ import {FakeViewportRuler} from '../core/overlay/position/fake-viewport-ruler';
 describe('MdCheckbox', () => {
   let fixture: ComponentFixture<any>;
 
+  /** Creates a DOM mouse event. */
+  const createMouseEvent = (eventType: string, dict: any = {}) => {
```
mmalerba: Can you add a TODO to use the testing utils being added in #2943 once they're available

tinayuangao: The current behavior for IE is incorrect. Issue #3145 caused this test failure.

PR #3141, opened 2017-02-16
fix(select): allow option with undefined or null value to clear selection

### LINKED ISSUE #2634: md-select doesn't reset placeholder effectively "clearing" selection
It's entirely possible I'm not doing things right, but... the placeholder within the md-select doesn't appear to be resetting from it's floating position after selecting an empty md-option.. effectively clearing the selection.

Here's what I've got:  

        <md-select name="thing" placeholder="Thing" formControlName="thing">
            <md-option></md-option>
            <md-option *ngFor="let t of things" [value]="t.id">{{t.name}}</md-option>
        </md-select>

My form/controls are defined as follow:

        thing = new FormControl(null, Validators.required);

        this.editForm = new FormGroup({
            thing: this.thing,
            ...
        });

When initialized with a null value, the placeholder appears as it should, after choosing an option, then choosing the empty option (<md-option></md-option>), the form control's value is set to null, however the placeholder doesn't leave the floating position.

I've also tried setting the empty option's value to null (<md-option [value]="null"></md-option>) with no success. 

I'm using Chrome on Windows 10.
Angular 2.4.3
Material 2.0.0-beta.1

Thanks!

### LINKED ISSUE #3110: feat(md-select) - enable to clear the value (unselect)
feat

In a md-select, let the user create a option to "unselect" the item value.

When we add a empty option with value="" (ng-value="undefined" or ng-value="null"), the placeholder act as if we've selected an option

After selecting one option on md-select, try to "unselect" and revert it to the "empty" state

http://plnkr.co/edit/e8Qoaq7M1ubiWHpVKctu?p=preview

Some non-required select on my applications need to let the user "unselect" the options

crisbeto:
Allows for options with a value of `null` or `undefined` to clear the select. This is similar to the way the native select works.

Fixes #3110.
Fixes #2634.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -567,7 +567,7 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
    */
   private _selectValue(value: any): MdOption {
     let optionsArray = this.options.toArray();
-    let correspondingOption = optionsArray.find(option => option.value === value);
+    let correspondingOption = optionsArray.find(option => option.value && option.value === value);
 
     if (correspondingOption) {
       correspondingOption.select();
@@ -632,8 +632,14 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
       wasSelected ? option.deselect() : option.select();
       this._sortValues();
     } else {
-      this._clearSelection(option);
-      this._selectionModel.select(option);
+      if (option.value == null) {
+        this._clearSelection();
+        this._onChange(option.value);
+        this.change.emit(new MdSelectChange(this, option.value));
+      } else {
+        this._clearSelection(option);
+        this._selectionModel.select(option);
+      }
     }
 
     if (wasSelected !== this._selectionModel.isSelected(option)) {
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -627,17 +627,17 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
   private _onSelect(option: MdOption): void {
     const wasSelected = this._selectionModel.isSelected(option);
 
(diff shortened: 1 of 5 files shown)


jefersonestevo:
Is this change coming on beta3?

kara:
@crisbeto Sorry for the late review on this. It looks good, but probably needs some reworking / rebasing. Can you update?

crisbeto:
Sorry about this one @kara, I must've missed the notification. I've rebased it and I had to redo it to match the new setup. Can you take another look?

src/lib/select/select.ts
```diff
@@ -630,8 +630,14 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
       wasSelected ? option.deselect() : option.select();
       this._sortValues();
     } else {
-      this._clearSelection(option);
-      this._selectionModel.select(option);
+      if (option.value == null) {
+        this._clearSelection();
+        this._onChange(option.value);
+        this.change.emit(new MdSelectChange(this, option.value));
```
kara: Rather than duplicating the change event logic from `propagateChanges`, seems like we could tweak `propagateChanges` to accept an optional value arg. I think it'd help make the conditional more readable.

```ts
private _onSelect(option) {
  if (this.multiple) {
      ...
  } else if (option.value == null) {
     this._clearSelection();
     this._propagateChanges(option.value);
  } else {
    this._clearSelection(option);
    this._selectionModel.select(option);
  }
...
}

private _propagateChanges(value?: any): void {
    let valueToEmit = Array.isArray(this.selected) ? this.selected.map(option => option.value)
        : (this.selected ? this.selected.value : value);
  ...
}

```

crisbeto:
Addressed the feedback @kara.

PR #3144, opened 2017-02-16
fix(pseudo-checkbox): alignment issue and border color

crisbeto:
* Fixes an alignment issue with the pseudo checkboxes. Most likely caused by the `mat-` prefix changes or a faulty conflict resolution.
* Fixes pseudo checkboxes not using the proper border color from the theme.
* Adds the pseudo checkboxes to the checkbox demo for easier testing.

### DIFF
--- a/src/lib/core/selection/pseudo-checkbox/_pseudo-checkbox-theme.scss
+++ b/src/lib/core/selection/pseudo-checkbox/_pseudo-checkbox-theme.scss
@@ -8,9 +8,6 @@
   $warn: map-get($theme, warn);
   $background: map-get($theme, background);
 
-  // The color of the checkbox's checkmark / mixedmark.
-  $checkbox-mark-color: mat-color($background, background);
-
   // NOTE(traviskaufman): While the spec calls for translucent blacks/whites for disabled colors,
   // this does not work well with elements layered on top of one another. To get around this we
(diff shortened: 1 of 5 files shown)


src/lib/core/selection/pseudo-checkbox/_pseudo-checkbox-theme.scss
```diff
@@ -8,18 +8,19 @@
...
   // this does not work well with elements layered on top of one another. To get around this we
   // blend the colors together based on the base color and the theme background.
   $white-30pct-opacity-on-dark: #686868;
   $black-26pct-opacity-on-light: #b0b0b0;
   $disabled-color: if($is-dark-theme, $white-30pct-opacity-on-dark, $black-26pct-opacity-on-light);
 
-  .mat-pseudo-checkbox::after {
-    color: $checkbox-mark-color;
+  .mat-pseudo-checkbox {
+    color: if($is-dark-theme, rgba(white, 0.7), rgba(black, 0.54));
```
kara: Can you use 'secondary-text' from the $foreground palette instead? You won't have to check the theme.

crisbeto:
Addressed the feedback @kara. Also applied the same approach to the regular checkboxes since it's where I got the color in the first place.

PR #3160, opened 2017-02-17
Changed theming guide from mat to md prefix

crutchcorn:
Very small pull request, and I may be wrong, but I had to change the following values to get things working. Looking through the code it seems that I am correct, but it's likely I could be wrong.


crutchcorn:
Went ahead and signed

sumpton:
see breaking changes here about switching from md- to mat- namespace

https://github.com/angular/material2/releases

crutchcorn:
Ahh, I seemed to have missed that

PR #3162, opened 2017-02-17
fix(tabs): set tab body content to 100%

### LINKED ISSUE #3153: md-tab-group tabs should be able to contain scroll-able components
Bug/Feature Request

Elements within an md-tab of an md-tab-group should be independently scroll-able. nav tabs (md-tab-nav-bar) **do** allow independently scroll-able contents

scroll is only possible for the whole tab

http://plnkr.co/edit/h67A8ZSeH747E3oYKdv5?p=preview

A tab could fill screen real estate but contain multiple lists of data. It would be useful to be able to freeze a list header (containing a search input say) and allow the rest of the list to be scrollable. 

@angular/cli: 1.0.0-beta.31
node: 7.2.0
os: win32 x64
@angular/common: 2.4.7
@angular/compiler: 2.4.7
@angular/compiler-cli: 2.4.7
@angular/core: 2.4.7
@angular/flex-layout: 2.0.0-beta.5
@angular/forms: 2.4.7
@angular/http: 2.4.7
@angular/platform-browser: 2.4.7
@angular/platform-browser-dynamic: 2.4.7
@angular/router: 3.4.7
@angular/cli: 1.0.0-beta.31

Like your work :)

andrewseguin:
Before the content would be set to full height which did not allow the users to set their content to "height: 100%". Now the body content matches the visible height of the tab.

Fixes #3153


amardeep:
I also need to set
.mat-tab-body-wrapper {
  flex-grow: 1;
}

for 100% height to work correctly. 

My use case is that if i set the md-tab-group height to 100%, and the tab content is also set to 100%, then it should expand to take all the space

andrewseguin:
@amardeep Please set up a plunker that reproduces your need and I'll be happy to adjust the CSS to fit your use case

PR #3163, opened 2017-02-17
fix(select-spec): Update the rtl test to ensure the fixture is stable

trshafer:
Hello Angular Material,

I believe I have found the issue to my failure in #2989. So I think IE was actually doing the right thing, and my change exposed the issue.

My understanding is adding [`fixture.whenStable()`](https://angular.io/docs/ts/latest/guide/testing.html#!#when-stable) should not change the result of the test, because the synchronous version should not have any unfinished async tasks. This change breaks the test, and thus I believe the current test is faulty.

This test run should be red for all browsers. Let me know how best to proceed.

Thanks,
Thomas

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1040,34 +1040,38 @@ describe('MdSelect', () => {
         select.style.marginRight = '20px';
       });
 
-      it('should adjust for the checkbox in ltr', () => {
+      it('should adjust for the checkbox in ltr', async(() => {
         trigger.click();
         multiFixture.detectChanges();
 
-        const triggerLeft = trigger.getBoundingClientRect().left;
-        const firstOptionLeft =
-            document.querySelector('.cdk-overlay-pane md-option').getBoundingClientRect().left;
+        multiFixture.whenStable().then(() => {
+          const triggerLeft = trigger.getBoundingClientRect().left;
+          const firstOptionLeft =
+              document.querySelector('.cdk-overlay-pane md-option').getBoundingClientRect().left;
 
-        // 48px accounts for the checkbox size, margin and the panel's padding.
-        expect(firstOptionLeft.toFixed(2))
-            .toEqual((triggerLeft - 48).toFixed(2),
-                `Expected trigger label to align along x-axis, accounting for the checkbox.`);
-      });
+          // 48px accounts for the checkbox size, margin and the panel's padding.
+          expect(firstOptionLeft.toFixed(2))
+              .toEqual((triggerLeft - 48).toFixed(2),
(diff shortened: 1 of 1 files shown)


trshafer:
I believe it has to do with the way width is calculated, but not entirely sure. To make the tests more consistent, I added I just added `whenStable` to the ltr test as well.

mmalerba:
@kara do you know why whenStable is needed for the RTL test? should it be added to all of the tests, after a click occurs?

jelbourn:
Confirmed that the test failures are already a problem, but that they're just masked somehow.

src/lib/select/select.spec.ts
```diff
@@ -976,37 +976,41 @@ describe('MdSelect', () => {
...
-            document.querySelector('.cdk-overlay-pane md-option').getBoundingClientRect().right;
-
-        // Each option is 32px wider than the trigger, so it must be adjusted 16px
-        // to ensure the text overlaps correctly.
-        expect(firstOptionRight.toFixed(2))
-            .toEqual((triggerRight + 16).toFixed(2),
-                `Expected trigger to align with the selected option on the x-axis in RTL.`);
-      });
+        fixture.whenStable().then(() => {
+          const triggerRight = trigger.getBoundingClientRect().right;
```
kara: Can you try moving `fixture.detectChanges()` here (instead of line 1000)?  I believe the placement is causing the test to fail, since change detection normally runs after every turn and isn't here.

I've confirmed that the alignment in RTL does work on IE and other browsers, so it seems that the test needs to be updated.

trshafer: I believe detectChanges is supposed to be called before whenStable()

https://angular.io/docs/ts/latest/guide/testing.html#!#the-_async_-function-in-_it_

kara: I'm a bit confused by your comment. Even in your link, it does what I'm suggesting and calls `fixture.detectChanges` within `whenStable()`. You can still call it on line 1000 if you like, but it needs to be called in the stable block.

trshafer: I think I need to keep it outside of the whenStable call to detect the click changes but I added it inside.

kara:
The tests seems to still be failing, even locally for me. Will need more looking into.

crisbeto:
@trshafer @kara I've combined these changes into https://github.com/angular/material2/pull/3861 and fixed the test failure so we don't have to keep track of multiple PRs.

crisbeto:
Now that #3861 is in, can we close this @trshafer?

trshafer:
Looks like it was fixed. Thank you.

PR #3165, opened 2017-02-17
feat(ripple): expose ripple directive in template

DevVersion:
* When programmatically calling `launch` on the ripple element the developer can specifiy a configuration object.
  
  Since the `launch` method is part of the directive and all the associated `@Input`s. The given configuration should be used as default.

 * Exposes the ripple component instance to the Angular template.

### DIFF
--- a/src/lib/core/ripple/ripple.spec.ts
+++ b/src/lib/core/ripple/ripple.spec.ts
@@ -75,6 +75,7 @@ describe('MdRipple', () => {
   }
 
   describe('basic ripple', () => {
+    let rippleDirective: MdRipple;
 
     const TARGET_HEIGHT = 200;
     const TARGET_WIDTH = 300;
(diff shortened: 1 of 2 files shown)


src/lib/core/ripple/ripple.spec.ts
```diff
@@ -111,15 +113,29 @@ describe('MdRipple', () => {
...
       expect(rippleTarget.querySelectorAll('.mat-ripple-element').length).toBe(0);
 
-      rippleComponent.launch(0, 0);
+      rippleDirective.launch(0, 0);
 
       expect(rippleTarget.querySelectorAll('.mat-ripple-element').length).toBe(1);
     });
 
+    it('creates manual ripples with the default ripple config', () => {
+      expect(rippleTarget.querySelectorAll('.mat-ripple-element').length).toBe(0);
```
crisbeto: Nit: might be fractionally faster to do `expect(rippleTarget.querySelector('.mat-ripple-element')).toBeFalsy();`.

DevVersion: Yeah it might be fractionally faster, but I think when checking for ripple elements we want to explicitly check for an amount. Therefore the falsy check should be consistent here.

PR #3181, opened 2017-02-18
fix(autocomplete): blur not being handled in compatibility mode

crisbeto:
Fixes the blur event not being handled properly if the user uses `mat-option`, instead of `md-option`.


crisbeto:
Closing in favor of #3188.

PR #3184, opened 2017-02-18
refactor(focus-trap): convert to directive

crisbeto:
Refactors the focus trap to be used as a directive, rather than a component. This gives us a couple of advantages:
* It can be used on the same node as other components.
* It removes a level of nesting in the DOM. This makes it slightly more convenient to style projected in cases like the dialog (see #2546), where flexbox needs to be applied to the closest possible ancestor.

Also includes the following improvements:
* Adds a service for setting up a focus trap without direct access to the template.
* No longer triggers change detection when focus hits the start/end anchors.
* Resets the anchor tab index when trapping is disabled, instead of removing elements from the DOM.
* Adds missing unit tests for the disabled and cleanup logic.

### DIFF
--- a/src/lib/core/a11y/focus-trap.ts
+++ b/src/lib/core/a11y/focus-trap.ts
@@ -1,78 +1,122 @@
-import {Component, ViewEncapsulation, ViewChild, ElementRef, Input, NgZone} from '@angular/core';
+import {
+  Directive,
+  ElementRef,
+  Input,
+  NgZone,
+  OnDestroy,
+  AfterContentInit,
+  Injectable,
+} from '@angular/core';
 import {InteractivityChecker} from './interactivity-checker';
 import {coerceBooleanProperty} from '../coercion/boolean-property';
 
 
 /**
- * Directive for trapping focus within a region.
+ * Class that allows for trapping focus within a DOM element.
  *
- * NOTE: This directive currently uses a very simple (naive) approach to focus trapping.
+ * NOTE: This class currently uses a very simple (naive) approach to focus trapping.
  * It assumes that the tab order is the same as DOM order, which is not necessarily true.
  * Things like tabIndex > 0, flex `order`, and shadow roots can cause to two to misalign.
  * This will be replaced with a more intelligent solution before the library is considered stable.
  */
-@Component({
-  moduleId: module.id,
-  selector: 'cdk-focus-trap, focus-trap',
-  templateUrl: 'focus-trap.html',
-  encapsulation: ViewEncapsulation.None,
-})
 export class FocusTrap {
-  @ViewChild('trappedContent') trappedContent: ElementRef;
+  private _startAnchor: HTMLElement;
+  private _endAnchor: HTMLElement;
 
   /** Whether the focus trap is active. */
-  @Input()
-  get disabled(): boolean { return this._disabled; }
-  set disabled(val: boolean) { this._disabled = coerceBooleanProperty(val); }
-  private _disabled: boolean = false;
+  get enabled(): boolean { return this._enabled; }
+  set enabled(val: boolean) {
+    this._enabled = val;
 
-  constructor(private _checker: InteractivityChecker, private _ngZone: NgZone) { }
+    if (this._startAnchor && this._endAnchor) {
+      this._startAnchor.tabIndex = this._endAnchor.tabIndex = this._enabled ? 0 : -1;
+    }
+  }
+  private _enabled: boolean = true;
+
+  constructor(
+    private _element: HTMLElement,
+    private _checker: InteractivityChecker,
+    private _ngZone: NgZone,
+    deferAnchors = false) {
+
+    if (!deferAnchors) {
+      this.attachAnchors();
+    }
+  }
+
+  /** Destroys the focus trap by cleaning up the anchors. */
+  destroy() {
+    if (this._startAnchor && this._startAnchor.parentNode) {
+      this._startAnchor.parentNode.removeChild(this._startAnchor);
+    }
+
+    if (this._endAnchor && this._endAnchor.parentNode) {
+      this._endAnchor.parentNode.removeChild(this._endAnchor);
+    }
+
+    this._startAnchor = this._endAnchor = null;
+  }
 
   /**
-   * Waits for microtask queue to empty, then focuses the first tabbable element within the focus
-   * trap region.
+   * Inserts the anchors into the DOM. This is usually done automatically
+   * in the constructor, but can be deferred for cases like directives with `*ngIf`.
+   */
+  attachAnchors(): void {
+    if (!this._startAnchor) {
+      this._startAnchor = this._createAnchor();
+    }
+
+    if (!this._endAnchor) {
+      this._endAnchor = this._createAnchor();
+    }
+
+    this._ngZone.runOutsideAngular(() => {
+      this._element
+        .insertAdjacentElement('beforebegin', this._startAnchor)
+        .addEventListener('focus', () => this.focusLastTabbableElement());
+
+      this._element
+        .insertAdjacentElement('afterend', this._endAnchor)
+        .addEventListener('focus', () => this.focusFirstTabbableElement());
+    });
+  }
+
+  /**
+   * Waits for microtask queue to empty, then focuses
+   * the first tabbable element within the focus trap region.
    */
   focusFirstTabbableElementWhenReady() {
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-      this.focusFirstTabbableElement();
-    });
+    this._ngZone.onMicrotaskEmpty.first().subscribe(() => this.focusFirstTabbableElement());
   }
 
   /**
-   * Waits for microtask queue to empty, then focuses the last tabbable element within the focus
-   * trap region.
+   * Waits for microtask queue to empty, then focuses
+   * the last tabbable element within the focus trap region.
    */
   focusLastTabbableElementWhenReady() {
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => {
-      this.focusLastTabbableElement();
-    });
+    this._ngZone.onMicrotaskEmpty.first().subscribe(() => this.focusLastTabbableElement());
   }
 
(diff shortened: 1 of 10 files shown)


src/lib/core/a11y/focus-trap.ts
```diff
@@ -11,48 +11,72 @@ import {coerceBooleanProperty} from '../coercion/boolean-property';
  * Things like tabIndex > 0, flex `order`, and shadow roots can cause to two to misalign.
  * This will be replaced with a more intelligent solution before the library is considered stable.
  */
-@Component({
-  moduleId: module.id,
-  selector: 'cdk-focus-trap, focus-trap',
-  templateUrl: 'focus-trap.html',
-  encapsulation: ViewEncapsulation.None,
+@Directive({
+  selector: 'cdk-focus-trap, focus-trap, [cdk-focus-trap], [focus-trap]',
```
mmalerba: You don't need to add the un-prefixed attribute, we're trying to move to the prefixed versions. Also attribute should be [cdkFocusTrap]

src/lib/core/a11y/focus-trap.ts
```diff
@@ -11,48 +11,72 @@ import {coerceBooleanProperty} from '../coercion/boolean-property';
...
-  @ViewChild('trappedContent') trappedContent: ElementRef;
+export class FocusTrap implements AfterViewInit, OnDestroy {
+  private _startAnchor: HTMLElement = this._createAnchor();
+  private _endAnchor: HTMLElement = this._createAnchor();
 
   /** Whether the focus trap is active. */
   @Input()
   get disabled(): boolean { return this._disabled; }
-  set disabled(val: boolean) { this._disabled = coerceBooleanProperty(val); }
+  set disabled(val: boolean) {
```
mmalerba: I don't think we can have this input on the attribute version, it might conflict with disabled attribute on another element

mmalerba:
Oh also, I had a document where I was thinking about changing the way this component/directive works: https://docs.google.com/document/d/1Kd7stLC0gPWpXiPrK9zHs9gWfpEBgxf15tXaBscXO7g/preview

src/lib/core/a11y/focus-trap.ts
crisbeto: Hmm, I didn't really consider that. My gut feeling is that Angular would just pass the disabled value to both the component and the directive.

crisbeto:
@mmalerba I've reworked this one as we discussed, can you take another look?

src/lib/core/a11y/focus-trap.ts
```diff
@@ -114,4 +158,55 @@ export class FocusTrap {
...
+    return anchor;
+  }
+}
+
+
+/**
+ * Service that allows easy instantiation of focus traps.
+ */
+@Injectable()
+export class FocusTrapService {
```
mmalerba: I thought we decided not to bother with this and just let components create a new FocusTrap directly?

src/lib/core/a11y/focus-trap.ts
```diff
@@ -114,4 +158,55 @@ export class FocusTrap {
...
+    return new FocusTrap(element, this._checker, this._ngZone, deferAnchors);
+  }
+}
+
+
+/**
+ * Directive for trapping focus within a region.
+ */
+@Directive({
+  selector: 'cdk-focus-trap, [cdkFocusTrap]',
```
mmalerba: Can you split these into 2 separate directives that work like this:

`<cdk-focus-trap [disabled]="!isEnabled"></cdk-focus-trap>`
`<div [cdkTrapFocus]="isEnabled"></div>`

Reason being that we can't be guaranteed that the attribute won't be attached to some element that already assigns some other meaning to the `disabled` attribute. (also mark the element version deprecated and update any documentation/examples that refer to the element version).

src/lib/core/a11y/focus-trap.ts
```diff
@@ -1,60 +1,105 @@
...
 export class FocusTrap {
-  @ViewChild('trappedContent') trappedContent: ElementRef;
+  private _startAnchor: HTMLElement = this._createAnchor();
+  private _endAnchor: HTMLElement = this._createAnchor();
 
   /** Whether the focus trap is active. */
-  @Input()
   get disabled(): boolean { return this._disabled; }
-  set disabled(val: boolean) { this._disabled = coerceBooleanProperty(val); }
+  set disabled(val: boolean) {
```
mmalerba: I would switch this to enabled now that its not an `@Input()`

src/lib/core/a11y/focus-trap.ts
```diff
@@ -1,60 +1,105 @@
...
+    this._disabled = val;
+    this._startAnchor.tabIndex = this._endAnchor.tabIndex = this._disabled ? -1 : 0;
+  }
   private _disabled: boolean = false;
 
-  constructor(private _checker: InteractivityChecker, private _ngZone: NgZone) { }
+  /** Element to which the focus trap is attached. */
+  get element(): HTMLElement {
```
mmalerba: don't need this

src/lib/core/a11y/focus-trap.ts
crisbeto: I went with the injectable so we don't have to worry about passing the injected `InteractivityChecker` and `NgZone` every time we need to instantiate a focus trap.

mmalerba: Ok that's fine with me, but Jermey would be upset with me if I let you name it Service :p how about `FocusTrapFactory` and change `attach` to `create` (since the attachment part is kind of optional anyways)

crisbeto:

src/lib/core/a11y/focus-trap.ts
```diff
@@ -114,4 +153,84 @@ export class FocusTrap {
 
     return null;
   }
+
+  private _createAnchor(): HTMLElement {
+    let anchor = document.createElement('div');
+    anchor.tabIndex = 0;
```
mmalerba: nit: this.enabled ? 0 : -1;

src/lib/core/a11y/focus-trap.ts
```diff
@@ -1,60 +1,100 @@
...
-  @Input()
-  get disabled(): boolean { return this._disabled; }
-  set disabled(val: boolean) { this._disabled = coerceBooleanProperty(val); }
-  private _disabled: boolean = false;
+  get enabled(): boolean { return this._enabled; }
+  set enabled(val: boolean) {
+    this._enabled = val;
+    this._startAnchor.tabIndex = this._endAnchor.tabIndex = this._enabled ? 0 : -1;
+  }
+  private _enabled: boolean = false;
```
mmalerba: this was false when it was called _disabled, should it be true now?

src/lib/core/a11y/focus-trap.ts
```diff
@@ -114,4 +153,84 @@ export class FocusTrap {
...
+    anchor.tabIndex = 0;
+    anchor.classList.add('cdk-visually-hidden');
+    anchor.classList.add('cdk-focus-trap-anchor');
+    return anchor;
+  }
+}
+
+
+/**
+ * Factory that allows easy instantiation of focus traps.
```
mmalerba: nit: /** one-liners like this */

src/lib/core/a11y/focus-trap.ts
```diff
@@ -114,4 +153,84 @@ export class FocusTrap {
...
+ */
+@Injectable()
+export class FocusTrapFactory {
+  constructor(private _checker: InteractivityChecker, private _ngZone: NgZone) { }
+
+  attach(element: HTMLElement, deferAnchors = false): FocusTrap {
```
mmalerba: change method name to create?

src/lib/core/a11y/focus-trap.ts
crisbeto: It should, good catch.

src/lib/core/a11y/focus-trap.ts
```diff
@@ -1,78 +1,119 @@
...
 
   /** Whether the focus trap is active. */
-  @Input()
-  get disabled(): boolean { return this._disabled; }
-  set disabled(val: boolean) { this._disabled = coerceBooleanProperty(val); }
-  private _disabled: boolean = false;
+  get enabled(): boolean { return this._enabled; }
+  set enabled(val: boolean) {
+    this._enabled = val;
+    this._startAnchor.tabIndex = this._endAnchor.tabIndex = this._enabled ? 0 : -1;
```
mmalerba: need to check that the anchors exist now

src/lib/core/a11y/focus-trap.ts
```diff
@@ -1,78 +1,119 @@
...
+    deferAnchors = false) {
+
+    if (!deferAnchors) {
+      this.attachAnchors();
+    }
+  }
+
+  /** Destroys the focus trap by cleaning up the anchors. */
+  destroy() {
+    if (this._startAnchor.parentNode) {
```
mmalerba: this._startAnchor && this.startAnchor.parentNode

crisbeto:
Apparently the Github conflict resolution tool leaves trailing whitespace. Will fix the linter error when I get home.

src/lib/core/a11y/focus-trap.ts
```diff
@@ -114,4 +158,81 @@ export class FocusTrap {
...
+  }
+}
+
+
+/**
+ * Directive for trapping focus within a region.
+ * @deprecated
+ */
+@Directive({
+  selector: 'cdk-focus-trap',
```
kara: Is it intentional to only deprecate one selector and not the other?

mmalerba: focus-trap has been deprecated, so i think its fine to remove, but now cdk-focus-trap is deprectaed too, since we're moving to an attribute version

kara: Okay, just wanted to double-check.

kara:
@crisbeto We need to export `FocusTrapDirective` in core.ts as well

crisbeto:
Done @kara.

PR #3188, opened 2017-02-19
fix(autocomplete): unable to click to select items in IE

### LINKED ISSUE #3351: Bug - Can't select an option in autocomplete  IE 11

Bug

When selecting an option in the auto complete , it should display the selected value in the text field

When selecting an option in the auto complete, it does not display the selected value in the text field

1. https://material.angular.io/components/component/autocomplete
2. Go to example
3. Type Ala
4. Select Alabama

IE 11

crisbeto:
* Fixes being unable to select autocomplete items by clicking in IE. This was due to IE not setting the `event.relatedTarget` for blur events.
* Fixes potential issue if the user uses `mat-option`, instead of `md-option`.

Fixes #3351.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -1,15 +1,18 @@
 import {
-    Directive,
-    ElementRef,
-    forwardRef,
-    Host,
-    Input,
-    NgZone,
-    Optional,
-    OnDestroy,
-    ViewContainerRef,
+  Directive,
+  ElementRef,
+  forwardRef,
+  Host,
+  Input,
+  NgZone,
+  Optional,
+  OnDestroy,
+  ViewContainerRef,
+  Inject,
+  ChangeDetectorRef,
 } from '@angular/core';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
+import {DOCUMENT} from '@angular/platform-browser';
 import {Overlay, OverlayRef, OverlayState, TemplatePortal} from '../core';
 import {MdAutocomplete} from './autocomplete';
 import {PositionStrategy} from '../core/overlay/position/position-strategy';
@@ -18,12 +21,13 @@ import {Observable} from 'rxjs/Observable';
 import {MdOptionSelectionChange, MdOption} from '../core/option/option';
 import {ENTER, UP_ARROW, DOWN_ARROW} from '../core/keyboard/keycodes';
 import {Dir} from '../core/rtl/dir';
+import {MdInputContainer} from '../input/input-container';
 import {Subscription} from 'rxjs/Subscription';
-import {Subject} from 'rxjs/Subject';
 import 'rxjs/add/observable/merge';
+import 'rxjs/add/observable/fromEvent';
+import 'rxjs/add/operator/filter';
 import 'rxjs/add/operator/startWith';
 import 'rxjs/add/operator/switchMap';
-import {MdInputContainer} from '../input/input-container';
 
 /**
  * The following style constants are necessary to save here in order
@@ -58,8 +62,8 @@ export const MD_AUTOCOMPLETE_VALUE_ACCESSOR: any = {
     '[attr.aria-expanded]': 'panelOpen.toString()',
     '[attr.aria-owns]': 'autocomplete?.id',
     '(focus)': 'openPanel()',
-    '(blur)': '_handleBlur($event.relatedTarget?.tagName)',
     '(input)': '_handleInput($event)',
+    '(blur)': '_onTouched()',
     '(keydown)': '_handleKeydown($event)',
   },
   providers: [MD_AUTOCOMPLETE_VALUE_ACCESSOR]
@@ -73,9 +77,6 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   private _panelPositionSubscription: Subscription;
 
   private _positionStrategy: ConnectedPositionStrategy;
-
-  /** Stream of blur events that should close the panel. */
-  private _blurStream = new Subject<any>();
 
   /** Whether or not the placeholder state is being overridden. */
   private _manuallyFloatingPlaceholder = false;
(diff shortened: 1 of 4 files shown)


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -223,11 +223,17 @@ export class MdAutocompleteTrigger implements AfterContentInit, ControlValueAcce
...
+   */
+  _handleFocusOut(event: FocusEvent): void {
+    const relatedTarget = event.relatedTarget as HTMLElement;
+
     this._onTouched();
 
-    // Only emit blur event if the new focus is *not* on an option.
-    if (newlyFocusedTag !== 'MD-OPTION') {
+    // Only emit blur event if the new focus is *not* on an element inside the panel.
+    if (relatedTarget && !this._overlayRef.overlayElement.contains(relatedTarget)) {
```
kara: This appears to break clicking outside the panel to close it in Chrome and Firefox. Can you fix?

crisbeto:
Addressed the feedback @kara.

kara:
@crisbeto Closing on blur still appears to be broken on Firefox. Have you tried just expanding the old check to include MAT-OPTION?

crisbeto:
The issue isn't with the node name, I'm testing it against our demo app that uses only `md-option`. It breaks because IE doesn't assign the `relatedTarget`.

kara:
Ok. Can you fix Firefox?

crisbeto:
I will, although it may be worth rethinking the approach since the `relatedTarget` seems to be pretty brittle.

crisbeto:
@kara could you give this another look?

AAAnderson7301:
Any update on merging this?

donduni:
Is there any plan to merge this one?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -133,6 +137,16 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
+
+    if (!this._unbindGlobalListener) {
+      this._unbindGlobalListener = this._renderer.listenGlobal('document', 'click',
+        (event: MouseEvent) => {
+          if (!this._inputContainer._elementRef.nativeElement.contains(event.target) &&
+              !this._overlayRef.overlayElement.contains(event.target as HTMLElement)) {
+            this._outsideClickStream.next(null);
+          }
+        });
+    }
```
kara: `openPanel` is getting a bit heavy. Can you move this new block into its own method?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -133,6 +137,16 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
 
     this._panelOpen = true;
     this._floatPlaceholder();
+
+    if (!this._unbindGlobalListener) {
+      this._unbindGlobalListener = this._renderer.listenGlobal('document', 'click',
+        (event: MouseEvent) => {
+          if (!this._inputContainer._elementRef.nativeElement.contains(event.target) &&
+              !this._overlayRef.overlayElement.contains(event.target as HTMLElement)) {
+            this._outsideClickStream.next(null);
```
kara: Have you thought about using Observable.fromEvent here? I think it would be cleaner, because then we don't have the maintain the subscription or manually emit (we can just pass it to panelClosingActions).

```ts

get outsideClickStream() {
  return Observable.fromEvent(this._document, 'click')
     .filter(event => // check if contains target);
}

get panelClosingActions() {
 ...
   this.outsideClickStream
}
```

crisbeto: IIRC, I didn't go with `fromEvent`, because it may break server-side rendering. Thinking about it again, it should be fine since the menu shows up as a result from a user action.

kara: I see. Yeah, had the same train of thought. I think it will be fine given that it's not on initial render.

crisbeto:
Fixed the test failures, addressed the feedback and rebased @kara.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -100,9 +102,11 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   }
 
   constructor(private _element: ElementRef, private _overlay: Overlay,
-              private _viewContainerRef: ViewContainerRef,
+              private _renderer: Renderer, private _viewContainerRef: ViewContainerRef,
```
kara: Looks like you're not using the renderer anymore. Remove?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -144,6 +148,7 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
 
     this._panelOpen = false;
     this._resetPlaceholder();
+    this._changeDetectorRef.detectChanges();
```
kara: Curious why this is necessary. Add comment?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -170,6 +175,17 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
     }
   }
 
+  get outsideClickStream(): Observable<any> {
```
kara: Hmm let's keep this private?  I think I may have omitted the _ in my comment above by mistake, but I liked that you originally had it private.

src/lib/autocomplete/autocomplete-trigger.ts
crisbeto: I'll add a comment. It seems like using `fromEvent` doesn't trigger change detection, or at least it doesn't trigger it at the proper time. This means that the placeholder doesn't get reset when you click on the backdrop.

kara: Makes sense. Just for posterity, would be good to have that documented.

crisbeto:
Addressed the last set of comments @kara.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -170,6 +179,18 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
     }
   }
 
+  /** Stream of click outside of the autocomplete panel. */
```
kara: `clicks`

PR #3208, opened 2017-02-20
style(button): polished icon buttons ripples

EladBezalel:
Icon button ripples to start from the center and removed hover style on icon buttons

### DIFF
--- a/src/lib/button/button.html
+++ b/src/lib/button/button.html
@@ -1,6 +1,7 @@
 <span class="mat-button-wrapper"><ng-content></ng-content></span>
 <div md-ripple *ngIf="!_isRippleDisabled()" class="mat-button-ripple"
-    [class.mat-button-ripple-round]="_isRoundButton"
-    [mdRippleTrigger]="_getHostElement()"></div>
+     [class.mat-button-ripple-round]="_isRoundButton || _isIconButton"
+     [mdRippleCentered]="_isIconButton"
+     [mdRippleTrigger]="_getHostElement()"></div>
 <!-- the touchstart handler prevents the overlay from capturing the initial tap on touch devices -->
 <div class="mat-button-focus-overlay" (touchstart)="$event.preventDefault()"></div>
--- a/src/lib/button/button.scss
+++ b/src/lib/button/button.scss
@@ -1,19 +1,21 @@
 // TODO(jelbourn): Measure perf benefits for translate3d and will-change.
 // TODO(jelbourn): Figure out if anchor hover underline actually happens in any browser.
-
 @import 'button-base';
 @import '../core/a11y/a11y';
 
 .mat-button, .mat-icon-button {
   @extend %mat-button-base;
 
-  // Only flat buttons and icon buttons (not raised or fabs) have a hover style.
-  &:hover {
-    // Use the same visual treatment for hover as for focus.
-    .mat-button-focus-overlay {
-      opacity: 1;
-    }
+  .mat-button-focus-overlay {
+    transition: none;
+    opacity: 0;
   }
+}
+
+// Only flat buttons (not raised, fabs or icon buttons) have a hover style.
+// Use the same visual treatment for hover as for focus.
+.mat-button:hover .mat-button-focus-overlay {
+  opacity: 1;
 }
 
 .mat-raised-button {
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -100,10 +100,10 @@ export class MdButton implements OnDestroy {
   private _color: string;
 
   /** Whether the button is round. */
-  _isRoundButton: boolean = ['icon-button', 'fab', 'mini-fab'].some(suffix => {
-    let el = this._getHostElement();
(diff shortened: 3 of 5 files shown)


src/lib/button/button.scss
```diff
@@ -7,17 +7,24 @@
...
     }
   }
+}
 
-  &[disabled]:hover {
-    &.mat-primary, &.mat-accent, &.mat-warn, .mat-button-focus-overlay {
-      background-color: transparent;
+.mat-button {
+  // Only flat buttons (not raised, fabs or icon buttons) have a hover style.
+  &:hover {
```
DevVersion: Can't we do `.mat-button:hover .mat-button-focus-overlay` to avoid unnecessary nesting?

src/lib/button/button.ts
```diff
@@ -110,6 +110,8 @@ export class MdButton {
     return el.hasAttribute('md-' + suffix) || el.hasAttribute('mat-' + suffix);
   });
 
+  _isIconButton: boolean = this._getHostElement().hasAttribute('md-icon-button');
```
DevVersion: This would not cover the `mat-icon-button` attribute?

src/lib/button/button.html
```diff
@@ -1,6 +1,7 @@
 <span class="mat-button-wrapper"><ng-content></ng-content></span>
 <div md-ripple *ngIf="!_isRippleDisabled()" class="mat-button-ripple"
-    [class.mat-button-ripple-round]="_isRoundButton"
-    [mdRippleTrigger]="_getHostElement()"></div>
+     [class.mat-button-ripple-round]="_isRoundButton || _isIconButton"
```
DevVersion: The `_isIconButton` check can be removed. `_isRoundButton` also includes the icon buttons.

src/lib/button/button.ts
EladBezalel: it should

src/lib/button/button.html
EladBezalel: I removed the check from roundButton if it's an icon, It should be separated, Round buttons can contain text, icon buttons can be in round buttons..

src/lib/button/button.ts
DevVersion: How would it? The method `_isRoundButton` checks both prefixes? (See [here](https://github.com/angular/material2/pull/3208/files#diff-96337821f283af0f38c4fd8ed4c98065R110))

src/lib/button/button.html
DevVersion: But it still checks for the `icon` as well? [See here](https://github.com/angular/material2/pull/3208/files#diff-96337821f283af0f38c4fd8ed4c98065R108)

src/lib/button/button.ts
jelbourn: The `: boolean` part of this should be unnecessary since TypeScript infers the type from the right hand side.

jelbourn: Yeah, it must check the `mat-` version as well.

jelbourn:
Can you also change the commit message to be more specific about what you did?

EladBezalel:
@jelbourn done

src/lib/button/button.ts
```diff
@@ -157,6 +157,14 @@ export class MdButton implements OnDestroy {
   _isRippleDisabled() {
     return this.disableRipple || this.disabled;
   }
+
+  _checkElementAttribues(list: Array<string>) {
```
jelbourn:
```ts
/** Gets whether the button has one of the given attributes with either an 'md-' or 'mat-' prefix. */
_hasAttributeWithPrefix(unprefixedAttributeNames: string[])
```

EladBezalel: I spent almost 5 minutes figuring out what i want to call it, eventually i gave up and assumed you'd have a better name :) thanks

src/lib/button/button.ts
```diff
@@ -157,6 +157,17 @@ export class MdButton implements OnDestroy {
   _isRippleDisabled() {
     return this.disableRipple || this.disabled;
   }
+
+  /** Gets whether the button has one of the given attributes
```
jelbourn: Comment formatting is slightly off, should be:
```ts
  /**
   * Gets whether the button has one of the given attributes
   * with either an 'md-' or 'mat-' prefix.
   */
```

src/lib/button/button.ts
```diff
@@ -157,6 +157,17 @@ export class MdButton implements OnDestroy {
...
+   * with either an 'md-' or 'mat-' prefix.
+   */
+  _hasAttributeWithPrefix(unprefixedAttributeNames: string[]) {
```
jelbourn: It just occured to me that this would be more appropriate as a rest argument:
```ts
  _hasAttributeWithPrefix(...unprefixedAttributeNames: string[]) {
    // ...
  }
```
Then the call sites don't need the array syntax.

EladBezalel: sweet i like that

EladBezalel:
@jelbourn done

jelbourn:
LGTM

PR #3211, opened 2017-02-20
feat(select): add md-select-header directive

### LINKED ISSUE #2812: feat(select): Add select header to the md-select
Feature request

Yesterday I filled a pull request ( #2797 ) for adding a search bar in the md-select component. As @jelbourn pointed out this wasn't the right way to add this feature, so now I'm willing to do it the right way. 
What's the best way to address this? Just implement the same functionality as material does it? 

That will mean something like this is possible:
``` html
<md-select placeholder="Favorite food">
    <md-select-header>
        <input [ngModel]="searchTerm" type="search" placeholder="Search for a vegetable..">
    </md-select-header>   
        <md-option *ngFor="let food of foods | filter " [value]="food.value">
            {{ food.viewValue }}
        </md-option>
</md-select>
```

crisbeto:
Adds a `md-select-header` component, which is a fixed header above the select's options. It allows for the user to project an input to be used for filtering long lists of options.

**Note:** This component only handles the positioning, styling and exposes the panel id for a11y. The functionality is up to the user to handle.

Fixes #2812.

### DIFF
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -113,5 +113,20 @@ Space above cards: <input type="number" [formControl]="topHeightCtrl">
     </md-card>
   </div>
 
+  <md-card *ngIf="showSelect">
+    <md-select placeholder="Drink" [(ngModel)]="currentDrink" #selectWitHeader="mdSelect">
+      <md-select-header>
+        <input type="search" [(ngModel)]="searchTerm" role="combobox" [attr.aria-owns]="selectWitHeader.panelId"
+          (ngModelChange)="filterDrinks()" placeholder="Search for a drink"/>
+      </md-select-header>
+
+      <md-option *ngFor="let drink of filteredDrinks" [value]="drink.value" [disabled]="drink.disabled">
+        {{ drink.viewValue }}
+      </md-option>
+
+    </md-select>
+  </md-card>
+
+
 </div>
 <div style="height: 500px">This div is for testing scrolled selects.</div>
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -117,14 +117,33 @@ $mat-select-panel-max-height: 256px !default;
   margin: 0 $mat-select-arrow-margin;
 }
 
-.mat-select-panel {
-  @include mat-menu-base();
-  padding-top: 0;
-  padding-bottom: 0;
+.mat-select-content {
+  @include mat-menu-scrollable();
   max-height: $mat-select-panel-max-height;
   min-width: 100%; // prevents some animation twitching and test inconsistencies in IE11
 
   @include cdk-high-contrast {
     outline: solid 1px;
+  }
+}
+
+.mat-select-panel {
+  @include mat-menu-base();
+  border: none;
+}
+
+.mat-select-header {
+  @include mat-menu-item-base();
+  border-bottom: solid 1px;
+  box-sizing: border-box;
+
+  input {
+    display: block;
+    width: 100%;
+    height: 100%;
+    border: none;
+    outline: none;
+    padding: 0;
+    background: transparent;
   }
 }
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -32,7 +32,7 @@ import {
 import {map} from 'rxjs/operator/map';
 
 
-describe('MdSelect', () => {
+fdescribe('MdSelect', () => {
   let overlayContainerElement: HTMLElement;
   let dir: {value: 'ltr'|'rtl'};
   let scrolledSubject = new Subject();
@@ -1320,15 +1320,23 @@ describe('MdSelect', () => {
     describe('x-axis positioning', () => {
 
       beforeEach(() => {
+<<<<<<< HEAD
         select.style.position = 'fixed';
         select.style.left = '30px';
       });
 
       it('should align the trigger and the selected option on the x-axis in ltr', fakeAsync(() => {
+=======
+        select.style.marginLeft = select.style.marginRight = '50px';
+      });
+
+      it('should align the trigger and the selected option on the x-axis in ltr', async(() => {
+>>>>>>> chore: try to fix edge tests
         trigger.click();
         tick(400);
         fixture.detectChanges();
 
+<<<<<<< HEAD
         const triggerLeft = trigger.getBoundingClientRect().left;
         const firstOptionLeft = document.querySelector('.cdk-overlay-pane md-option')
             .getBoundingClientRect().left;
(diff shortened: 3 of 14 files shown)


fxck:
Would it be worth adding an example in the demo showing a bit more reactive way of doing that filtering?

fxck:
@kara hi, I know you are probably busy with whatever you are currently working on on material2 and form improvement in angular itself as well as preparing for upcoming conferences, but could you perhaps please take a look at these two https://github.com/angular/material2/pull/3211 https://github.com/angular/material2/pull/3341 select-related PRs from @crisbeto? They are not big changes in the codebase itself and both are very desirable features of the select box I'm sure many many people would appreciate.. Thank you.

src/lib/select/select.spec.ts
```diff
@@ -19,7 +19,7 @@ import {
 } from '@angular/forms';
 import {ViewportRuler} from '../core/overlay/position/viewport-ruler';
 
-describe('MdSelect', () => {
+fdescribe('MdSelect', () => {
```
Eernie: I believe the build is failing because of the `f` before `describe`

Eernie:
@kara Do you have an ETA for this pull request. I'd love see this merged :+1:

Jagpreet16:
@kara By when can we expect this feature to be available?

gedclack:
excuse me, is `md-select-header` implemented on beta.3 ?

src/lib/select/select.spec.ts
crisbeto: It is, but it's not the only issue. I'll get it passing once I get some feedback on it.

fxck:
@gedclack as you can see, this PR is not merged, not even reviewed, so no

badre429:
please work on this pr

fxck:
Any chance this will get implemented in the next like.. 4 months? It's a little unpleasant to have to maintain my own build of material2 because of these two(#3211 #3341) rather small, but pretty important for interfaces of any serious application, features that are like 20 LOC of actual non breaking changes to the codebase in total.

@kara @crisbeto @jelbourn could these please get some attention?

fxck:
😰

zyarnold:
any news for this yet?

fxck:
Any updates?

jelbourn:
We still plan on doing this, it's just been on the back-burner as other things have been higher priority

fxck:
@jelbourn while I totally understand that there were/are things with higher priority, wouldn't it be worth dedicating, say, one day in a month to review and merge these small PRs that are low in effort and pretty damn high in value for consumers?

What are you using to decide the priority? Are you using one of these ([1](https://medium.com/design-of-a-technology-business/ultimate-guide-to-product-prioritization-b08c18d5c00f), [2](https://www.productplan.com/strategies-prioritize-product-features/)) by any chance? Are you planning on involving the community in the decision in any way in the future?

jelbourn:
The reason this one in particular was back-burnered was that we needed to figure out what the a11y story for this feature is. It would be very easy for someone to break the a11y of the select this this, which is something we want to put some effort into avoiding.

irackley-mc:
@jelbourn Thanks for the update on this. Had been waiting a while wondering why there was no traction on this. Guess I'll have to do a custom build for now but hopefully some priority can be added to this, as I agree with @fxck and it is a great value add.

crisbeto:
@jelbourn I've cleaned up and rebased this one.

src/lib/select/select.scss
```diff
@@ -117,14 +117,33 @@ $mat-select-panel-max-height: 256px !default;
...
+  @include mat-menu-base();
+  border: none;
+}
+
+.mat-select-header {
+  @include mat-menu-item-base();
+  border-bottom: solid 1px;
+  box-sizing: border-box;
+
+  input {
```
crisbeto: The issue with allowing anything inside the header is that, with the way it is currently set up, the positioning won't work with anything that isn't 48px high.

jelbourn: I'd rather do that for now and then revisit the positioning to make it more flexible in the future.

src/demo-app/select/select-demo.html
```diff
@@ -113,5 +113,20 @@
     </md-card>
   </div>
 
+  <md-card *ngIf="showSelect">
+    <md-select placeholder="Drink" [(ngModel)]="currentDrink" #selectWitHeader="mdSelect">
+      <md-select-header>
+        <input type="search" [(ngModel)]="searchTerm" role="combobox" [attr.aria-owns]="selectWitHeader.panelId"
```
crisbeto: Hmm you're right. This complicates things a lot, because it means that we'd need two different key handling strategies depending on whether the select has a header (`FocusKeyManager` by default and `ActiveDescendantKeyManager` if there's a header), which in turn will complicate the select even more, because it would have to delegate to the header in some cases but not in others. I can see how it can turn into a huge mess pretty quickly since we'd need to sprinkle `if (this.header)` all over the place and there would be a `MdSelect <-> MdSelectHeader` dependency.

fxck: > or recommend that people use `md-autocomplete` instead.

just not this, please, searching is not the sole use case for the select header and even if it was, autocomplete has it's own problems that make it hard to use as a select.. https://github.com/angular/material2/issues/3334

fxck:
So is this on hold now again? @jelbourn @crisbeto

crisbeto:
It is until we can figure out a decent way to handle the accessibility.

benb7760:
@crisbeto Have found a few more issues that haven't been mentioned above when using this;

1. When two or more options with the same key, and one is selected, when option is filtered out and then unfiltered again, the first option with that value is selected and the other(s) ignored. This behaviour doesn't present in normal select as the options are never removed. _Should an error be thrown when providing multiple options with same value?_
2. In multiple mode, if I select some options then apply a filter which hides some of the selected options, then select a new option, when i remove the filter, any that were selected but hidden are no longer selected. This is probably because the selection model is updated based on the options property, which does not contain options when filtered out

I am keeping my own branch for this PR, updated to master, with a basic fix for the focus issue mentioned by jelbourn. I intend to fix the other two issues also, as I need the feature. Would it be useful for me to make a PR for this when done?

crisbeto:
@benb7760 it's not the functionality that stopped this from being pushed through, but rather the accessibility story. I have a task on my list to refactor the select focus management so it can handle having a header.

fxck:
Looks like this can now be done, now that #6856 is in?

crisbeto:
Yes the a11y shouldn't be an issue anymore. I might have to resubmit it though since it'll take a while to rebase this.

badre429:
@crisbeto is there any timeline for this feature

crisbeto:
Closing in favor of https://github.com/angular/material2/pull/7835.

mojtaba-ramezani:
cold_sweat

Issue #3166, opened 2017-02-17
Cannot use tabs in compatibility mode


andrewseguin:
The compatibility check sees that the tabs include md-tab-body and md-tab-header, which are listed as not allowed in compatibility mode.

PR #3226, opened 2017-02-21
removed ink-bar from the compatibility check lists

pyu10055:
similar to tab-body and tab-header, ink-bar is not used externally.

issue #3166


andrewseguin:
This is correct and looks good to me, non-exported components do not need compatibility check and can cause problems if used within component templates under compatibility mode.

andrewseguin:
Can you sign the CLA?

andrewseguin:
Closed as duplicate of #3267

PR #3228, opened 2017-02-21
refactor(connected-overlay-directive): avoid issues with property initialization

### LINKED ISSUE #3225: ConnectedOverlayDirective relies on binding order
Bug

The directive should work independent of the order in which the bindings are set.

The directive relies on the fact that the `origin` input is set before the `open` property, as the `open` property has a setter that relies on `origin` to be filled.

Users won't, I hit this while creating the new view engine for Angular 2.

Changing Angular without breaking material.

At least current master.

A solution might be to use `ngOnChanges` instead of a setter for `open` in `ConnectedOverlayDirective`.

crisbeto:
* Refactors the `ConnectedOverlayDirective` not to depend on the order in which the `open` and `origin` properties are set.
* Moves the `open`, `offsetX` and `offsetY` properties from using setters to `ngOnChanges` since the latter tends to output less JS.

Fixes #3225.

### DIFF
--- a/src/lib/core/overlay/overlay-directives.ts
+++ b/src/lib/core/overlay/overlay-directives.ts
@@ -179,7 +179,7 @@ export class ConnectedOverlayDirective implements OnDestroy, OnChanges {
 
   ngOnChanges(changes: SimpleChanges) {
     if (changes['open']) {
-      changes['open'].currentValue ? this._attachOverlay() : this._detachOverlay();
+      this.open ? this._attachOverlay() : this._detachOverlay();
     }
   }
 
--- a/src/lib/core/overlay/overlay-directives.ts
+++ b/src/lib/core/overlay/overlay-directives.ts
@@ -10,6 +10,8 @@ import {
     Output,
     ElementRef,
     Renderer2,
+    OnChanges,
+    SimpleChanges,
 } from '@angular/core';
 import {Overlay, OVERLAY_PROVIDERS} from './overlay';
 import {OverlayRef} from './overlay-ref';
@@ -63,10 +65,9 @@ export class OverlayOrigin {
   selector: '[cdk-connected-overlay], [connected-overlay], [cdkConnectedOverlay]',
   exportAs: 'cdkConnectedOverlay'
 })
-export class ConnectedOverlayDirective implements OnDestroy {
+export class ConnectedOverlayDirective implements OnDestroy, OnChanges {
   private _overlayRef: OverlayRef;
   private _templatePortal: TemplatePortal;
-  private _open = false;
   private _hasBackdrop = false;
   private _backdropSubscription: Subscription;
   private _positionSubscription: Subscription;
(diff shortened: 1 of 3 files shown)


kara:
@crisbeto This appears to break md-select positioning. Can you fix?

crisbeto:
Fixed @kara. I don't know if calling `detectChanges` is considered a code smell, but it should be shorter than getters/setters. If you think it may be problematic, I'll switch it back.

src/lib/core/overlay/overlay-directives.ts
```diff
@@ -172,6 +144,20 @@ export class ConnectedOverlayDirective implements OnDestroy {
     this._destroyOverlay();
   }
 
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes['open']) {
+      changes['open'].currentValue ? this._attachOverlay() : this._detachOverlay();
```
jelbourn: After you check `changes['open']`, shouldn't you be able to just use `this.open`?
(same for others)

src/lib/select/select.ts
```diff
@@ -577,6 +577,7 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
     }
 
     this._checkOverlayWithinViewport(maxScroll);
+    this._changeDetectorRef.detectChanges();
```
jelbourn: Add comment explaining why this `detectChanges` call is necessary

src/lib/core/overlay/overlay-directives.ts
```diff
@@ -76,30 +75,10 @@ export class ConnectedOverlayDirective implements OnDestroy {
...
-    return this._offsetX;
-  }
-
-  set offsetX(offsetX: number) {
-    this._offsetX = offsetX;
-    if (this._position) {
-      this._position.withOffsetX(offsetX);
-    }
-  }
+  @Input() offsetX: number = 0;
```
jelbourn: How significant is the code size difference? We use quite a lot of getter/setters.

crisbeto: A getter/setter combination transpiles to this in ES5:

```
Object.defineProperty(A.prototype, "thing", {
    get: function () {
        return this._thing;
    },
    set: function (value) {
        this._thing = value;
    },
    enumerable: true,
    configurable: true
});
```

I think that it can definitely start adding up over an entire project.

crisbeto:

crisbeto:
@jelbourn I've rebased this one and removed the changes to the `offsetX` and `offsetY` since they also turned out to be a bit dangerous. Can you take another look?

src/lib/core/overlay/overlay-directives.ts
```diff
@@ -183,6 +177,12 @@ export class ConnectedOverlayDirective implements OnDestroy {
     this._destroyOverlay();
   }
 
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes['open']) {
+      this.open ? this._attachOverlay() : this._detachOverlay();
+    }
+  }
+
```
jelbourn: Can you add a unit test for the case this fixes?

crisbeto:
Added a test case for the changes @jelbourn.

PR #3231, opened 2017-02-21
bug(md-slide-toggle): slide toggle should fit to it's content width

### LINKED ISSUE #3230: bug(md-slide-toggle): selectable area gets 100% of container width
bug

The selectable area from md-slide-toggle should fit to it's content

md-slide-toggle gets 100% of the container width as selectable area

[Plunker](http://plnkr.co/edit/PF5nzeUhUbtCBbsw2utR?p=preview)

"@angular/material" :"2.0.0-beta.2"

jefersonestevo:
Fixes #3230


DevVersion:
Please update the demo-app to have each `md-slide-toggle` element in its own row.

Also we should ensure that the baseline still works properly for the slide-toggle. See the `baseline/` route.

jefersonestevo:
@DevVersion I've updated the demo app to show each slide-toggle on it's own row, but there are no slide-toggle on the baseline route. Should I add it there?

jefersonestevo:
@DevVersion is there anything else that needs to change here?

DevVersion:
@jefersonestevo Sorry for the delay. I will look into it tomorrow!

src/demo-app/slide-toggle/slide-toggle-demo.scss
```diff
@@ -0,0 +1,8 @@
+.slide-toggle-example {
+  display:flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  align-content: flex-start;
+  align-items: flex-start;
```
DevVersion: Only doing the following is enough:

```scss
display: flex;
flex-direction: column;
```

src/demo-app/slide-toggle/slide-toggle-demo.scss
```diff
@@ -0,0 +1,8 @@
+.slide-toggle-example {
+  display:flex;
```
jefersonestevo: Done

DevVersion: Cool thx. Why still using `align-items`?

src/demo-app/slide-toggle/slide-toggle-demo.scss
jefersonestevo: @DevVersion although just these properties set each slide toggle on it's own row, I kept the property `align-items: flex-start;` so the demo will not stretch the lines

DevVersion:
Thanks for the PR. I think we should consider this as a breaking change? @jelbourn

src/demo-app/slide-toggle/slide-toggle-demo.scss
jefersonestevo: Just to show that the slide toggle now has a display inline-block. If I do not use it, the browser will stretch the line and you can still click on the whole line to check the slide toggle.

[Example](http://plnkr.co/edit/iuTIRXgSXeQFWPopAgAA?p=preview)

DevVersion: Thanks. Yeah I already tested it locally. Looks good!

jelbourn:
In the issue I don't quite follow what you mean by "selectable area"- could you elaborate?

jefersonestevo:
@jelbourn if you look in this [Plunker](http://plnkr.co/edit/iNnkg4bnxJm4YOx93wnD?p=preview) you can see that, anywhere in the gray area that you click, you'll change the value from the slide toggle (it uses all the available width from it's parent container). The same doesn't happen with the md-checkbox, that fit's just to it's content width.

PR #3234, opened 2017-02-22
fix(e2e): Fix e2e flakes by setting default window size

tinayuangao:


e2e/screenshot.ts
```diff
@@ -3,11 +3,14 @@ import * as path from 'path';
 import {browser} from 'protractor';
 
 const OUTPUT_DIR = './screenshots/';
+const HEIGHT = 788;
```
mmalerba: is there some significance to these numbers? isn't the standard screen size usually 1024x768

tinayuangao: Yes, it should be 1024 x 768. But the test image size is different from what we currently have so I'm trying to find the current size. 

We can set it to 1024 x 768 and update the golds.

PR #3242, opened 2017-02-22
bug(md-select): change the arrow position to absolute

### LINKED ISSUE #3112: bug(md-select) - not rendering correctly inside a flex layout (pure css) when having ngModel
bug

When inside a flex layout (pure css), the md-select must behave consistently, independent if it has a ngModel or not.

When the md-select has a ngModel, and it's width is different from the default at page load (inside of a flex layout), the component does not fit inside the container.

This plunker demostrate this behavior:
[Plunker](http://plnkr.co/edit/O1V6N8jCyWpHdu1VTB12?p=preview)

We need to display our pages in a responsive layout and are using just pure css flex layout while we wait for [Angular Flex Layout](https://github.com/angular/flex-layout) to became more stable.

Firefox: 51.0.1 (64-bit)
Chromium: 53.0.2785.143
Chrome: 56.0.2924.87

jefersonestevo:
Fixes #3112


src/demo-app/select/select-demo.html
```diff
@@ -54,3 +54,27 @@
 
 </div>
 <div style="height: 500px">This div is for testing scrolled selects.</div>
+
+<md-card>
```
crisbeto: I don't think we need this example. The examples in the demo app are there to be able to click around the main use cases, they're not for actual testing.

jefersonestevo:
@crisbeto conflicts resolved and I've removed the demo for flex layout from the demo page. Anything else that needs to change here?

crisbeto:
Not code-wise, but you have a few merge conflicts which can cause issues with our scripts. Can you squash all your commits into one?

jefersonestevo:
@crisbeto As these changes are really small and there are some merges with lots of commits here, I've created a new pull request with just the changes based on the current master. I'm closing this one and we can continue at #3580

crisbeto:
Well, it's already done, but FYI you can still squash commits after the PR has been opened.

PR #3247, opened 2017-02-22
fix(select): incorrect menu width when there is no placeholder

### LINKED ISSUE #3244: bug(md-select): options not sized correctly if no placeholder set
Bug

When using a md-select without the `placeholder` attribute, the options shown should be the same width as the md-select element.

The options are at a smaller width than the md-select attribute, since the `_triggerWidth` property is only set when the `placeholder` attribute is set, which is done in https://github.com/angular/material2/blob/master/src/lib/select/select.ts#L217

bildschirmfoto 2017-02-22 um 18 41 34

http://embed.plnkr.co/U9jHgw/

@angular/material: 2.0.0-beta.2
Google Chrome Version 56.0.2924.87 (64-bit)

crisbeto:
The select pane's width wasn't being set properly, when it doesn't have a placeholder, because the width usually gets set within the placeholder's setter.

Fixes #3244.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -43,7 +43,8 @@ describe('MdSelect', () => {
         ThrowsErrorOnInit,
         BasicSelectOnPush,
         BasicSelectOnPushPreselected,
-        SelectWithPlainTabindex
+        SelectWithPlainTabindex,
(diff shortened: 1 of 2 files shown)


crisbeto:
I added the test to #3647. Closing.

kara:
OK, that works too!

PR #3249, opened 2017-02-23
fix(button): clip button text when overflow

improvise:
This is my first contribution, so I'm not completely sure what steps are required to merge this. I'm porting over an AngularJS application to Angular and discovered a discrepancy in the button behavior.

Angular Material has this feature here: https://github.com/angular/material/blob/master/src/components/button/button.scss#L75

### DIFF
--- a/src/lib/button/_button-base.scss
+++ b/src/lib/button/_button-base.scss
@@ -50,6 +50,8 @@ $mat-mini-fab-padding: 8px !default;
   line-height: $mat-button-line-height;
   padding: $mat-button-padding;
   border-radius: $mat-button-border-radius;
+  
+  overflow: hidden;
 
   &[disabled] {
     cursor: default;


improvise:
Regarding the label: I am a Googler.

kara:
@improvise Can you tell us a bit  more about the problem you're trying to solve? Is there an existing Github issue for this?

jelbourn:
As Kara mentioned, we're not really sure what problem this change aims to resolve. Closing for now, feel free to coment with some more details and we can re-open.

improvise:
For posterity:

This was an issue when using both flex-layout and material buttons, where we were unintentionally relying on an overflow: hidden attribute to crop arbitrarily long buttons and prevent them from overflowing the container. We've resolved the issue by imposing some limits on the button text.

http://plnkr.co/edit/OqmqzLj8HOod49IaooXn

PR #3270, opened 2017-02-23
perf(scroll-dispatcher): lazily subscribe to global events

### LINKED ISSUE #3237: [Performance] ScrollDispatcher should subscribe lazily, and destroy subscriptions when they are no longer used
ScrollDispatcher needs to create subscriptions to the window scroll/resize events lazily (when they are actually needed by the UI) and destroy them once the UI using them gets destroyed.

Currently it simply subscribes to the scroll and resize events on init, which is when the page loads.
Since it is not utilizing the zone.runOutsideAngular functionality, then it means that every time a scroll or resize event happens, it will re-render the UI several times.

This usually goes unnoticed in smaller applications, but when you have bigger applications, this slows down the UI a lot.
It also removes performance control from the developers.

It's fine to emit the events when the dispatcher is actually being used by an overlay or whatnot, but it makes no sense to trigger zone to re-render the view every time the user scrolls, when there's nothing new to re-render.

I'd suggest making the subscriptions lazy and destroy them when components unsubscribe, or if possible using the zone.runOutsideAngular functionality to never trigger a re-render in the first place.

version 2.0.0-beta.2

crisbeto:
Switches the `ScrollDispatcher` to only subscribe to `scroll` and `resize` events if there any registered callbacks. Also unsubscribes once there are no more callbacks.

Fixes #3237.

### DIFF
--- a/src/lib/core/overlay/scroll/scroll-dispatcher.ts
+++ b/src/lib/core/overlay/scroll/scroll-dispatcher.ts
@@ -41,7 +41,6 @@ export class ScrollDispatcher {
     const scrollSubscription = scrollable.elementScrolled().subscribe(() => this._notify());
 
     this.scrollableReferences.set(scrollable, scrollSubscription);
-    this._addGlobalSubscription();
   }
 
   /**
@@ -52,7 +51,6 @@ export class ScrollDispatcher {
     if (this.scrollableReferences.has(scrollable)) {
       this.scrollableReferences.get(scrollable).unsubscribe();
       this.scrollableReferences.delete(scrollable);
-      this._removeGlobalSubscription();
     }
   }
 
@@ -69,13 +67,23 @@ export class ScrollDispatcher {
       this._scrolled.asObservable();
 
     this._scrolledCount++;
-    this._addGlobalSubscription();
+
+    if (!this._globalSubscription) {
+      this._globalSubscription = Observable.merge(
+        Observable.fromEvent(window.document, 'scroll'),
+        Observable.fromEvent(window, 'resize')
+      ).subscribe(() => this._notify());
+    }
 
     // Note that we need to do the subscribing from here, in order to be able to remove
     // the global event listeners once there are no more subscriptions.
     return observable.subscribe(callback).add(() => {
       this._scrolledCount--;
-      this._removeGlobalSubscription();
+
+      if (this._globalSubscription && !this.scrollableReferences.size && !this._scrolledCount) {
+        this._globalSubscription.unsubscribe();
+        this._globalSubscription = null;
+      }
     });
   }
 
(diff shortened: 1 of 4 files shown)


src/lib/core/overlay/scroll/scroll-dispatcher.ts
```diff
@@ -19,18 +20,15 @@ export class ScrollDispatcher {
   /** Subject for notifying that a registered scrollable reference element has been scrolled. */
   _scrolled: Subject<void> = new Subject<void>();
 
+  /** Keeps track of the global `scroll` and `resize` subscriptions. */
```
andrewseguin: Add comment about when this should have subscriptions (when there is at least one actively registered scrollable).

andrewseguin:
Can you add tests for this? At minimum we should check that globalSubscriptions is null when there are no registered scrollables, and it exists when at least one scrollable is registered.

crisbeto:
It feels weird having to make the subscription public only to be able to test it, but I should be able to use a spy to check whether stuff gets invoked.

andrewseguin:
Just thought about the case of having no cdk-scrollables on the page but still wanting to respond to scroll events. We'll still want the tooltip to reposition in this case.

Does it make more sense to assign the global subscriptions when `scrolled` is called? It's invocation shows that there is intent from some component to listen to the scroll event.

crisbeto:
I see your point @andrewseguin, but I'm not sure how we'd know to unbind the global handler once there are no more `scrolled` subscriptions.

crisbeto:
Changed as discussed @andrewseguin. Could you take another look?

QuentinFchx:
I might not understand all the stuff going behind the scene, but why are you not using RX operators such as [refCount](http://reactivex.io/documentation/operators/refcount.html) here?

src/lib/core/overlay/scroll/scroll-dispatcher.ts
```diff
@@ -19,26 +20,28 @@ export class ScrollDispatcher {
...
   /**
    * Registers a Scrollable with the service and listens for its scrolled events. When the
    * scrollable is scrolled, the service emits the event in its scrolled observable.
    * @param scrollable Scrollable instance to be registered.
    */
   register(scrollable: Scrollable): void {
     const scrollSubscription = scrollable.elementScrolled().subscribe(() => this._notify());
+
     this.scrollableReferences.set(scrollable, scrollSubscription);
+    this._addGlobalSubscription();
```
andrewseguin: Should we still be adding/removing the global subscription in register and deregister? I thought that we would only worry about them through the `scrolled` function

crisbeto: Oh, I think I was working under the impression that this was used both for the scrollables and `scrolled`. It doesn't make sense outside of `scrolled`.

crisbeto: It's changed now.

PR #3274, opened 2017-02-24
fix(autocomplete): fix key manager instantiation

### LINKED ISSUE #3269: md-autocomplete error when using *ngIf in md-input-container
Bug

Drop down shows and no error is thrown

Drop down is not shown, and an error is thrown in Chrome Dev Tools:
- "Cannot read property 'toArray' of undefined"

Inside the md-input-container tag include an *ngIf directive
Plunker based off Material Example for Autocomplete:
https://plnkr.co/edit/AYx6dvqxMARf3WcgrX6x?p=preview

It should work

Angular 2.4.8, Material Beta 2

I have found a couple workarounds, but neither are ideal:

- put the *ngIf directive in a span around the md-input-control (shouldn't be necessary)
- put the md-autocomplete into the md-input-container (if this is the proper solution, I think demos should be updated and implemented in this manner.)

kara:
The key manager was placed on the trigger's ngAfterContentInit hook instead of the panel's, so sometimes the events were occurring in the wrong order.

Fixes #3269

### DIFF
--- a/src/demo-app/autocomplete/autocomplete-demo.html
+++ b/src/demo-app/autocomplete/autocomplete-demo.html
@@ -21,11 +21,13 @@ Space above cards: <input type="number" [formControl]="topHeightCtrl">
   </md-card>
 
   <md-card>
-    <div>Template-driven value (currentState): {{ currentState }}</div>
-    <div>Template-driven dirty: {{ modelDir.dirty }}</div>
 
-    <md-input-container>
-      <input mdInput placeholder="State" [mdAutocomplete]="tdAuto" [(ngModel)]="currentState" #modelDir="ngModel"
+    <div>Template-driven value (currentState): {{ currentState }}</div>
+    <div>Template-driven dirty: {{ modelDir?.dirty }}</div>
+
+    <!-- Added an ngIf below to test that autocomplete works with ngIf -->
+    <md-input-container *ngIf="true">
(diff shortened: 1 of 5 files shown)


src/demo-app/autocomplete/autocomplete-demo.html
```diff
@@ -21,11 +21,12 @@
...
 
   <md-card>
+
     <div>Template-driven value (currentState): {{ currentState }}</div>
-    <div>Template-driven dirty: {{ modelDir.dirty }}</div>
+    <div>Template-driven dirty: {{ modelDir?.dirty }}</div>
 
-    <md-input-container>
-      <input mdInput placeholder="State" [mdAutocomplete]="tdAuto" [(ngModel)]="currentState" #modelDir="ngModel"
+    <md-input-container *ngIf="true">
```
andrewseguin: What's the ngIf for?

kara: Good call. I guess it's pretty mysterious :)

src/demo-app/autocomplete/autocomplete-demo.ts
```diff
@@ -19,6 +19,8 @@ export class AutocompleteDemo {
 
   tdDisabled = false;
 
+  @ViewChild(NgModel) modelDir: NgModel;
```
kara: Talked in person about AOT

PR #3279, opened 2017-02-24
Fix(menu) 'unsubscribe' is undefined onDestroy in unit test

dharmeshpipariya:
'unsubscribe' is undefined onDestroy menu in unit test

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -69,7 +69,9 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   }
 
   ngOnDestroy() {
-    this._tabSubscription.unsubscribe();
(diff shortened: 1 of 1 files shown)


src/lib/menu/menu-directive.ts
```diff
@@ -69,7 +69,9 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   }
 
   ngOnDestroy() {
-    this._tabSubscription.unsubscribe();
+    if (this._tabSubscription) {
+      this._tabSubscription.unsubscribe();
+    }
```
jelbourn: Can you add a unit test that reproduces the issue you're encountering?

jelbourn:
Closing for missing unit test- feel free to comment if you add it and we can reopen.

PR #3294, opened 2017-02-24
fix(button): use FocusOriginMonitor for focus styles

### LINKED ISSUE #3215: Sidenav and Toolbar buttons are highlighed automatically
Bug

When the sidenav is opened and the user move the cursor on one of buttons the first button is always highlighted automatically.

Moreover, when the sidenav is closed the menu button on the toolbar is highlighted automatically until the user click somewhere else.

Highlights should be shown only if the user moves the cursor on button.

Plunker -> https://plnkr.co/edit/ve5Ez0WSinNyowxI46LO?p=preview

    "@angular/common": "2.4.7",
    "@angular/compiler": "2.4.7",
    "@angular/core": "2.4.7",
    "@angular/flex-layout": "^2.0.0-beta.4",
    "@angular/forms": "2.4.7",
    "@angular/http": "2.4.7",
    "@angular/material": "^2.0.0-beta.2",
    "@angular/platform-browser": "2.4.7",
    "@angular/platform-browser-dynamic": "2.4.7",
    "@angular/platform-server": "2.4.7",

mmalerba:
fixes #3215

### DIFF
--- a/src/lib/button/index.ts
+++ b/src/lib/button/index.ts
@@ -33,7 +33,6 @@ export * from './button';
     MdIconButtonCssMatStyler,
     MdFabCssMatStyler,
     MdMiniFabCssMatStyler,
-    StyleModule,
   ],
   declarations: [
     MdButton,
--- a/src/lib/button/index.ts
+++ b/src/lib/button/index.ts
@@ -1,7 +1,6 @@
 import {ModuleWithProviders, NgModule} from '@angular/core';
 import {CommonModule} from '@angular/common';
-import {MdRippleModule} from '../core/ripple/ripple';
-import {CompatibilityModule} from '../core/compatibility/compatibility';
+import {CompatibilityModule, MdRippleModule, StyleModule} from '../core';
 import {
   MdAnchor,
   MdButton,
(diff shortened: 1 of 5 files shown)


src/lib/button/index.ts
```diff
@@ -1 +1,49 @@
...
+@NgModule({
+  imports: [CommonModule, MdRippleModule, CompatibilityModule, StyleModule],
+  exports: [
+    MdButton, MdAnchor,
+    CompatibilityModule,
+    MdButtonCssMatStyler,
+    MdRaisedButtonCssMatStyler,
+    MdIconButtonCssMatStyler,
+    MdFabCssMatStyler,
+    MdMiniFabCssMatStyler
```
kara: It seems like maybe we should export `StyleModule` so you don't have to import it separately when using the `MdButtonModule` alone.

src/lib/button/index.ts
```diff
@@ -1 +1,55 @@
+import {ModuleWithProviders, NgModule} from '@angular/core';
+import {CommonModule} from '@angular/common';
+import {MdRippleModule} from '../core/ripple/ripple';
```
kara: This is failing presubmit because it can't find `MdRippleModule`. Previously we were importing from `../core` rather than using the deep import. Can you change it back?

kara:
@mmalerba Can you rebase?

mmalerba:
oh derp, sorry @kara i got too used to just assigning this back to you for more review :p

Issue #3198, opened 2017-02-20
Autocomplete Panel width


julianobrasil:
#### Bug, feature request, or proposal:
A Way to set the panel width in autocomplete

#### What is the expected behavior?
The panel width could have a property to change the panel-width. 

#### What is the current behavior? 
The panel width isn't wide enough, so the content doesn't fit in it. My project is hosted in heroku, so it compiles the project remotely and donwload the material autocomplete code also remotely (so I can't customize de component scss)

#### What are the steps to reproduce?
Just modify the example of the material.angular.io, changing the width of <md-input-container> to 100% and modify one of the states name to a longer name.

Plunker Link: https://plnkr.co/edit/nKgx04MYK0q0xhS1yfAZ?p=preview

In plunker, just click on the input to see that the panel is too short to show the first name, even if there's a lot of space in the window.

#### Which versions of Angular, Material, OS, browsers are affected?
@angular/cli: 1.0.0-beta.32.3 [1.0.0-beta.30]
node: 6.9.5
os: win32 x64
@angular/cli: 1.0.0-beta.32.3
@angular/common: 2.4.8
@angular/compiler: 2.4.8
@angular/core: 2.4.8
@angular/flex-layout: 2.0.0-beta.5
@angular/forms: 2.4.8
@angular/http: 2.4.8
@angular/material: 2.0.0-beta.2
@angular/platform-browser: 2.4.8
@angular/platform-browser-dynamic: 2.4.8
@angular/router: 3.4.8
@angular/compiler-cli: 2.4.8

#### Is there anything else we should know?
I'm currently setting the encapsulation to None in the host component to fix it, but I don't know if this is the recommended way to make the things right.

willshowell:
@julianobrasil for now, you could add this to your global stylesheet
```scss
.mat-autocomplete-panel {
  max-width: none !important;
}
```

The problem seems to be that the autocomplete is including styles from the menu, which has a `max-width`
```scss
// autocomplete.scss
.mat-autocomplete-panel {
  @include mat-menu-base();
  ...
}

// _menu-common.scss
@mixin mat-menu-base() {
  @include mat-elevation(8);
  min-width: $mat-menu-overlay-min-width;
  max-width: $mat-menu-overlay-max-width; // is 280px
  ...
}
```

julianobrasil:
@willshowell, thanks for the reply. I'll do as you said for now and wait for the next beta.

PR #3297, opened 2017-02-25
No max-width for autocomplete panel

christinakayastha:
Currently, the max-width is inherited from the mat-menu-base mixin forcing the panel to have  a width of 280px. Overriding it here with none so that the panel can expand out as necessary.

Fix based on comments in issue #3198.


christinakayastha:
I signed it!

aboglioli:
It's an important solution! Thanks. +1

PR #3300, opened 2017-02-25
fix(menu): not emitting close event when closing via the backdrop

### LINKED ISSUE #3295: bug(menu): mdMenu (close) output only called after menu-item clicked, not when menu is closed by clicking on overlay
**BUG**

MdMenu `close` output should be called whenever the menu is closed.

MdMenu `close` output is only called when the menu is closed after clicking a menu-item. The `close` output is not called when the menu is closed by clicking away from the menu and on the overlay backdrop.

Create an MdMenu with `(close)="console.log('test')"`.
Close the menu by clicking on the menu-items and by clicking on the overlay/backdrop.
Notice the console outputs.

A Plunker, when the menu button is clicked, `closed` is set to false. When the menu `close` output is emitted, `closed` is set to true. You can see that if you close the menu by clicking outside from the overlay/backdrop, `closed` does not get set to true. If you close the menu by clicking on one of the menu-items, `closed` is set to true.
http://plnkr.co/edit/zWePeOD6sPZm1kqWc5P4

Angular 2.4.8
Material2 2.0.0-beta.2
Currently on: Chrome, OSX

crisbeto:
* Fixes the menu not emitting the `close` event if it is closed by clicking on the backdrop.
* Adds missing test coverage for the `close` event in general.

Fixes #3295.


src/lib/menu/menu-directive.ts
```diff
@@ -64,7 +64,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   ngAfterContentInit() {
     this._keyManager = new FocusKeyManager(this.items).withWrap();
     this._tabSubscription = this._keyManager.tabOut.subscribe(() => {
-      this._emitCloseEvent();
+      this.emitCloseEvent();
```
kara: Can you re-add the _ ? This should be considered private.

PR #3311, opened 2017-02-26
fix(slide-toggle): occasional element jumping

### LINKED ISSUE #3290: Slide toggle moves when Overflow: hidden applied
Bug

Slide-toggle should not move

Slide-toggle moves up from initial state when toggled for the first time.

Just toggle the slide-toggle
http://plnkr.co/edit/k9A7lvjktjesDH0Y4IcA?p=preview

N/A

Latest Chrome, Edge and IE11. ng-version 2.4.8

In Edge and IE it is less noticeable. In Chrome it cuts half of the slide-toggle off.

DevVersion:
* Due to the many nested absolute elements the slide-toggle occasional started exceeding the slide-toggle boundaries on click.

* Replaces the hard-coded pixel values with structured SaSS calculations.

* No longer sets the `thumb-container` to a incorrect width. This has been done to limit the `transformX(100%)` statement.

FYI: @mmalerba - The slide-toggle bar will be slightly bigger than before because it now uses the correct width (as in Poylmer and MDL).

Fixes #3290

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.ts
+++ b/src/lib/slide-toggle/slide-toggle.ts
@@ -296,7 +296,7 @@ class SlideToggleRenderer {
   /** Previous checked state before drag started. */
   private _previousChecked: boolean;
 
-  /** Percentage of the thumb while dragging. */
+  /** Percentage of the thumb while dragging. Percentage as fraction of 100. */
   dragPercentage: number;
 
   /** Whether the thumb is currently being dragged. */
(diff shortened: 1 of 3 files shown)


DevVersion:
@mmalerba Yeah as expected, the screenshot-diffs failed for the slide-toggle. 
> Also @tinayuangao great work on the screenshot-diffs!

src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -333,12 +333,14 @@ class SlideToggleRenderer {
 
   /** Updates the thumb containers position from the specified distance. */
   updateThumbPosition(distance: number) {
-    this.dragPercentage = this._getThumbPercentage(distance);
-    applyCssTransform(this._thumbEl, `translate3d(${this.dragPercentage}%, 0, 0)`);
+    this.dragPercentage = this._getDragPercentage(distance);
+    // Calculate the moved distance based on the thumb bar width.
+    let dragX = (this.dragPercentage / 100) * this._thumbBarWidth;
+    applyCssTransform(this._thumbEl, `translate3d(${dragX}px, 0, 0)`);
```
mmalerba: would it be better to have a getter that returns `{'transform': 'translate3d(...)'}` and do `[ngStyle]=`?

DevVersion: I actually wanted to keep such logic in the renderer, but I think this is a good idea. I will make the changes.

DevVersion: I tried to use `ngStyle` but actually ran into more issues than advantages. 

The slide-toggle would need to trigger a new round of change-detection `onDragEnd`, so that the drag styles are reverted.

I feel like for the OnPush strategy it is more beneficial to use `applyTransform`.

PR #3313, opened 2017-02-26
feat(select): add ability to cycle through options with arrow keys when closed

### LINKED ISSUE #2990: Feature Request: down/up arrow keys activate select

Use up and down arrow keys to activate select menus and remove return as an activator.

Standard html select menus are activated on down/up arrow keys and click instead of the return key. This allows return to still submit while making access to the select menu available without a mouse. 

Behavior example https://plnkr.co/edit/MoygfDy6qyEyF96Sa0v4?p=preview

up and down do nothing while hitting enter activates the select menu.

accessibility in line with standard html features

material2 < 2.0.0-beta.1

crisbeto:
* Adds the ability for users to select options by focusing on a closed `md-select` and pressing the up/down arrow keys.
* Fixes a bug that prevents the selection from going to the first item in a `ListKeyManager`, if there were no previously-selected items.
* Adds an extra null check to the `FocusKeyManager` to avoid issues where the focused item is cleared.

Fixes #2990.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -460,6 +460,24 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
   _handleKeydown(event: KeyboardEvent): void {
     if (event.keyCode === ENTER || event.keyCode === SPACE) {
       this.open();
+    } else if (!this.disabled) {
+      let prevActiveItem = this._keyManager.activeItem;
+
+      // Cycle though the select options even when the select is closed,
+      // matching the behavior of the native select element.
+      // TODO(crisbeto): native selects also cycle through the options with left/right arrows,
+      // however the key manager only supports up/down at the moment.
+      this._keyManager.onKeydown(event);
+
+      let currentActiveItem = this._keyManager.activeItem as MdOption;
+
+      if (this._multiple) {
+        this.open();
+      } else if (currentActiveItem !== prevActiveItem) {
+        this._clearSelection();
+        this._setSelectionByValue(currentActiveItem.value);
+        this._propagateChanges();
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/select/select.ts
```diff
@@ -374,6 +375,21 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
   _handleKeydown(event: KeyboardEvent): void {
     if (event.keyCode === ENTER || event.keyCode === SPACE) {
       this.open();
+    } else if (!this.disabled) {
+      let prevActiveItem = this._keyManager.activeItem;
+
+      // TODO(crisbeto): native selects also cycle through the options with left/right arrows,
+      // however the key manager only supports up/down at the moment.
+      this._keyManager.onKeydown(event);
```
jelbourn: Add comment like 
```ts
// Cycle though the select options even when the select is closed, 
// matching the behavior of the native select element.
```

crisbeto:
Sorry for the delay on this one. Apparently I missed the notification and it also needed some refactoring after the mutli select PR got in.

src/lib/select/select.ts
```diff
@@ -533,11 +552,14 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
    * @returns Option that has the corresponding value.
    */
   private _selectValue(value: any): MdOption {
-    let correspondingOption = this.options.find(option => option.value === value);
+    let optionsArray = this.options.toArray();
+    let correspondingOptionIndex = optionsArray.findIndex(option => option.value === value);
```
jelbourn: We can't use `findIndex` inside of Google (Closure's type definition for array doesn't have it).

@crisbeto ping me when this one ready again

crisbeto:
Done @jelbourn.

PR #3315, opened 2017-02-26
feat(ripple): add option for persistent ripples

### LINKED ISSUE #3169: ripple: option to keep ripple persistent
As @crisbeto mentioned per chat, he needs a way for ripples to persist until they will be programmatically hidden.

We thought about another option for the `RippleConfig` and also about exposing an object for the ripple.

```ts
let ripple = ripple.launch(0, 0, { persist: true });
// Manually fade-out the persistent ripple
ripple.fadeOut();
```

Related to #2520

DevVersion:
* Adds an option to the ripple service that allows persistent ripples (useful for focus indicators)
* Manually launched ripples now return a `RippleRef` instance that can be used to to fade-out the ripples.
* Adds a method to the component that developers to fade-out all currently active ripple elements.

**Note**: This is a required feature for the planned focus indicators.

Closes #3169

### DIFF
--- a/src/lib/core/ripple/ripple-ref.ts
+++ b/src/lib/core/ripple/ripple-ref.ts
@@ -6,21 +6,13 @@ import {RippleConfig, RippleRenderer} from './ripple-renderer';
 export class RippleRef {
 
   constructor(
-    private renderer: RippleRenderer,
+    private _renderer: RippleRenderer,
     public element: HTMLElement,
     public config: RippleConfig) {
   }
 
   /** Fades out the ripple element. */
   fadeOut() {
-    let rippleIndex = this.renderer.activeRipples.indexOf(this);
-
-    // Remove the ripple reference if added to the list of active ripples.
-    if (rippleIndex !== -1) {
-      this.renderer.activeRipples.splice(rippleIndex, 1);
-    }
-
-    // Regardless of being added to the list, fade-out the ripple element.
-    this.renderer.fadeOutRipple(this.element);
+    this._renderer.fadeOutRipple(this);
   }
 }
--- a/src/lib/core/ripple/ripple-ref.ts
+++ b/src/lib/core/ripple/ripple-ref.ts
@@ -1,0 +1,26 @@
+import {RippleConfig, RippleRenderer} from './ripple-renderer';
+
+/**
+ * Reference to a previously launched ripple element.
+ */
+export class RippleRef {
+
+  constructor(
+    private renderer: RippleRenderer,
+    public element: HTMLElement,
+    public config: RippleConfig) {
+  }
+
+  /** Fades out the ripple element. */
+  fadeOut() {
+    let rippleIndex = this.renderer.activeRipples.indexOf(this);
+
+    // Remove the ripple reference if added to the list of active ripples.
+    if (rippleIndex !== -1) {
+      this.renderer.activeRipples.splice(rippleIndex, 1);
+    }
+
+    // Regardless of being added to the list, fade-out the ripple element.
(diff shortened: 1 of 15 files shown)


src/lib/core/ripple/ripple-ref.ts
```diff
@@ -0,0 +1,26 @@
+import {RippleConfig, RippleRenderer} from './ripple-renderer';
+
+/**
+ * Reference to a previously launched ripple element.
+ */
+export class RippleRef {
+
+  constructor(
+    private renderer: RippleRenderer,
```
crisbeto: Shouldn't this one be prefixed with an underscore?

src/lib/core/ripple/ripple-ref.ts
```diff
@@ -0,0 +1,26 @@
...
+    public config: RippleConfig) {
+  }
+
+  /** Fades out the ripple element. */
+  fadeOut() {
+    let rippleIndex = this.renderer.activeRipples.indexOf(this);
+
+    // Remove the ripple reference if added to the list of active ripples.
+    if (rippleIndex !== -1) {
+      this.renderer.activeRipples.splice(rippleIndex, 1);
```
crisbeto: It might be better if we did this inside the `renderer.fadeOutRipple` since the renderer is the one responsible for keeping track of the ripples.

DevVersion: Makes sense. Changed it.

src/lib/core/ripple/ripple.ts
```diff
@@ -87,8 +84,18 @@ export class MdRipple implements OnChanges, OnDestroy {
...
+  launch(pageX: number, pageY: number, config = this.rippleConfig): RippleRef {
+    return this._rippleRenderer.fadeInRipple(pageX, pageY, config);
+  }
+
+  /** Fades out all currently showing ripple elements. */
+  fadeOutAll() {
+    // Iterate in reverse, to avoid issues with the `fadeOut` method that will immediately remove
+    // items from the array.
+    let i = this._rippleRenderer.activeRipples.length;
+    while (i--) {
```
crisbeto: This may be more appropriate inside the renderer as well. You should be able to make the `activeRipples` private afterwards as well.

DevVersion: Hmm not sure. I didn't want to put all of the logic to the renderer. The render is actually just responsible for the DOM manipulation. Any other arguments why we should move it to the renderer as well?

crisbeto: My main argument is that, in my opinion, the renderer should be the only one that knows about the active ripples. If we shared them between the component and the renderer, it makes it a little harder to follow.

DevVersion: Yeah when moving it to the renderer, it might be easier to follow. Made the changes.

src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -41,12 +34,12 @@ export class RippleRenderer {
...
   private _isMousedown: boolean = false;
 
-  /** Currently active ripples that will be closed on mouseup. */
-  private _activeRipples: HTMLElement[] = [];
-
   /** Events to be registered on the trigger element. */
   private _triggerEvents = new Map<string, any>();
 
+  /** Currently active ripple references. */
+  private _activeRipples: RippleRef[] = [];
```
mmalerba: more efficient to use a Set

DevVersion: Done.

PR #3330, opened 2017-02-27
feat(datepicker): wire up selected value propagation

mmalerba:
as of this PR we have a bare minimum working datepicker \o/

### DIFF
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -47,12 +47,14 @@ export class MdDatepickerInput implements ControlValueAccessor {
 
   @Input()
   get value() {
-    return this._locale.parseDate(this._elementRef.nativeElement.value);
+    return this._value;
   }
-  set value(value: SimpleDate) {
-    let stringValue = value == null ? '' : this._locale.formatDate(value);
+  set value(value: any) {
+    this._value = this._locale.parseDate(value);
+    let stringValue = this._value == null ? '' : this._locale.formatDate(this._value);
     this._renderer.setElementProperty(this._elementRef.nativeElement, 'value', stringValue);
   }
+  private _value: SimpleDate;
 
   // Implemented as part of ControlValueAccessor
   writeValue(value: SimpleDate) {
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -1,17 +1,32 @@
-import {Directive, ElementRef, Input} from '@angular/core';
+import {Directive, ElementRef, forwardRef, Input, Renderer} from '@angular/core';
 import {MdDatepicker} from './datepicker';
+import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
+import {SimpleDate} from '../core/datetime/simple-date';
+import {CalendarLocale} from '../core/datetime/calendar-locale';
+
+
+export const MD_DATEPICKER_VALUE_ACCESSOR: any = {
+  provide: NG_VALUE_ACCESSOR,
+  useExisting: forwardRef(() => MdDatepickerInput),
+  multi: true
+};
 
 
 /** Directive used to connect an input to a MdDatepicker. */
 @Directive({
   selector: 'input[mdDatepicker], input[matDatepicker]',
+  providers: [MD_DATEPICKER_VALUE_ACCESSOR],
+  host: {
+    '(input)': '_onChange($event.target.value)',
+    '(blur)': '_onTouched()',
+  }
 })
-export class MdDatepickerInput {
+export class MdDatepickerInput implements ControlValueAccessor {
   @Input()
   set mdDatepicker(value: MdDatepicker) {
     if (value) {
       this._datepicker = value;
-      this._datepicker._registerInput(this._elementRef);
+      this._datepicker._registerInput(this);
     }
   }
   private _datepicker: MdDatepicker;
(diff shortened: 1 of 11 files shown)


src/lib/datepicker/datepicker-input.ts
```diff
@@ -1,23 +1,78 @@
...
+  }
+  set value(value: any) {
+    this._value = this._locale.parseDate(value);
+    let stringValue = this._value == null ? '' : this._locale.formatDate(this._value);
+    this._renderer.setElementProperty(this._elementRef.nativeElement, 'value', stringValue);
+  }
+  private _value: SimpleDate;
+
+  // Implemented as part of ControlValueAccessor
+  writeValue(value: SimpleDate) {
```
kara: Return types

src/lib/datepicker/datepicker-input.ts
```diff
@@ -1,23 +1,78 @@
...
+    return this._elementRef;
+  }
+
+  @Input()
+  get value() {
+    return this._value;
+  }
+  set value(value: any) {
+    this._value = this._locale.parseDate(value);
+    let stringValue = this._value == null ? '' : this._locale.formatDate(this._value);
```
kara: Nit: const?

src/lib/datepicker/index.ts
```diff
@@ -20,8 +20,27 @@ export * from './year-view';
...
-  declarations: [MdCalendar, MdCalendarTable, MdDatepicker, MdDatepickerInput, MdMonthView, MdYearView],
+  imports: [
+    CommonModule,
+    DatetimeModule,
+    MdDialogModule,
+    OverlayModule
+  ],
+  exports: [
+    MdCalendar,
+    MdCalendarTable,
```
kara: Do you intend for the CalendarTable, MonthView, etc. to be used externally? Seems like you can remove them from the exports list.

src/lib/datepicker/datepicker-input.ts
mmalerba: I guess... I don't see much value in the `: void`

src/lib/datepicker/datepicker-input.ts
```diff
@@ -1,23 +1,78 @@
...
+
+  constructor(private _elementRef: ElementRef, private _renderer: Renderer,
+              private _locale: CalendarLocale) {}
+
+  getPopupConnectionElementRef(): ElementRef {
+    return this._elementRef;
+  }
+
+  @Input()
+  get value() {
```
tinayuangao: nit: return type?

mmalerba:
addressed comments, PTAL

src/lib/datepicker/datepicker-input.ts
```diff
@@ -1,23 +1,87 @@
...
+  _onChange = (value: any) => {};
+
+  _onTouched = () => {};
+
+  constructor(private _elementRef: ElementRef, private _renderer: Renderer,
+              private _locale: CalendarLocale) {}
+
+  ngAfterContentInit() {
+    if (this._datepicker) {
+      this._datepicker.selectedChanged.subscribe((selected: SimpleDate) => {
```
kara: Do you unsubscribe from this anywhere?

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -122,3 +248,37 @@ class MultiInputDatepicker {
 class NoInputDatepicker {
   @ViewChild('d') datepicker: MdDatepicker;
 }
+
+
+@Component({
+  template: `
+    <input [mdDatepicker]="d" [value]="'1/1/2020'">
```
kara: Nit: might be easier to scan if it's just `value="1/1/2020"`

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -95,14 +127,110 @@ describe('MdDatepicker', () => {
...
+      expect(testComponent.selected).toBeNull();
+      expect(testComponent.datepickerInput.value).toBeNull();
+
+      let selected = new SimpleDate(2017, 0, 1);
+      testComponent.datepicker.selected = selected;
+      detectModelChanges(fixture);
+
+      expect(testComponent.selected).toEqual(selected);
+      expect(testComponent.datepickerInput.value).toEqual(selected);
+    }));
```
kara: We should add tests that ensure that touched, disabled, and dirty still work, given that we're re-implementing them in the value accessor

mmalerba:
comments addressed

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -95,14 +129,159 @@ describe('MdDatepicker', () => {
...
+
+    it('should mark input dirty after date selected', fakeAsync(() => {
+      let inputEl = fixture.debugElement.query(By.css('input')).nativeElement;
+
+      expect(inputEl.classList).toContain('ng-pristine');
+
+      testComponent.datepicker.selected = new SimpleDate(2017, 0, 1);
+      detectModelChanges(fixture);
+
+      expect(inputEl.classList).toContain('ng-dirty');
```
kara: It should actually not mark the control dirty based on programmatic changes. Dirtiness is a reflection of UI changes only (what the user does). We're probably calling the change callback somewhere we shouldn't be.

mmalerba: This property isn't really intended to be set programmatically, it's set when the user clicks a calendar cell. Made it internal so people don't try to call it, though...

Issue #3229, opened 2017-02-21
md-input-containers text and floating label misaligned when using mdPrefix


WillAnd31:
When using a `md-icon` as a prefix to an input inside of an `md-input-container` the floating label gets misaligned when using a `mdPrefix`, as you can see below:

<img width="343" alt="screen shot 2017-02-21 at 12 30 40 pm" src="https://cloud.githubusercontent.com/assets/11418316/23181315/c29b510c-f831-11e6-8c5e-5b51b7bc34e4.png">
<img width="336" alt="screen shot 2017-02-21 at 12 31 33 pm" src="https://cloud.githubusercontent.com/assets/11418316/23181318/c502634a-f831-11e6-8a12-5dfaee7ce59a.png">

The expected behavior is to have the text aligned with the icon and also have the floating label be at the correct position. As you can see when there is no `mdPrefix` icon the text is all aligned how it should be:

<img width="321" alt="screen shot 2017-02-21 at 12 31 01 pm" src="https://cloud.githubusercontent.com/assets/11418316/23181407/21b2790e-f832-11e6-8c50-83ff13159d4b.png">
<img width="333" alt="screen shot 2017-02-21 at 12 31 16 pm" src="https://cloud.githubusercontent.com/assets/11418316/23181408/23890d92-f832-11e6-81ea-c23abb6539aa.png">

To reproduce this simply add a `mdPrefix` to a `md-input-container`.
Before:
```
<md-input-container>
    <input mdInput placeholder="text" type="text" name="example">
</md-input-container>
```
and after:
```
<md-input-container>
    <md-icon mdPrefix>email</md-icon>
    <input mdInput placeholder="text" type="text" name="example">
</md-input-container>
```

Right now using `mdPrefix` is making the input not look good. If I am using mdPrefix wrong please let me know.

fxck:
dupe of https://github.com/angular/material2/issues/2586 I believe

Issue #2586, opened 2017-01-10
md-icon as input prefix / prefix on wide inputs 


fxck:
currently this happens when you try to use `md-icon` in a prefix
![image](https://cloud.githubusercontent.com/assets/1303561/21807898/2081b0d4-d741-11e6-8ab8-82c8f31aa95b.png)

there two issues

1) icon is too big, which messes with input's alignment 
![image](https://cloud.githubusercontent.com/assets/1303561/21807959/62020090-d741-11e6-8cfd-e986e3fccf85.png)

2) icon is too wide, which meses with `display: table-cell` and pushes the input unreasonably far right

should this even supported? I could work on it I guess

also why is table cell used instead of flex? it's hacky and puts way too much space around prefix even if I used `$`(if my md-input is wide)
![image](https://cloud.githubusercontent.com/assets/1303561/21808126/28a2061e-d742-11e6-9767-249837872523.png)

crisbeto:
In regards to flexbox, I agree that it would be better to use. Otherwise I think that https://github.com/angular/material2/pull/2496 may fix this.

fxck:
it will probably fix the stretching(but I think it would be worth converting to flex instead), but not the vertical alignment

kara:
Fixed by https://github.com/angular/material2/pull/3342

marshall007:
@kara @mmalerba the same issue pops up when using `md-icon` within an `md-placeholder`. I wasn't sure if I should create a separate issue or if you just want to reopen this one as it's essentially the same bug.

This is solved by applying the same fix from #3342 to `.mat-input-placeholder .mat-icon`. Alternatively, it seems that all three cases (prefix/suffix/placeholder) could potentially be handled more generally as `.mat-input-container .mat-icon`.

marshall007:
Upon further investigation, `.mat-input-container .mat-icon` is almost certainly the way to go in terms of future-proofing. I just noticed `md-hint > md-icon` has the same issue (and same resolution).

PR #3342, opened 2017-02-28
fix(input): make icons in prefix/suffix the right size

mmalerba:


mmalerba:
fixes #3229

src/lib/input/input-container.scss
```diff
@@ -237,8 +237,16 @@ $mat-input-underline-disabled-background-image:
...
+  // Prevents the prefix and suffix from stretching together with the container.
   width: 0.1px;
   white-space: nowrap;
+
+  // Allow icons in a prefix/suffix to adapt to the correct size.
+  & .mat-icon {
+    width: auto;
+    height: auto;
+    font-size: 100%;
+    vertical-align: top;
```
tinayuangao: Just curious, why do we use `vertical-align top` here?

mmalerba: It was part of the styles that were already in the demo app css that i was migrating into the library css. I'm not sure on the original rationale for `top`, but leaving it as `baseline` causes the prefix/suffix element to be too tall, `bottom` and `middle` also fix the issue but I just left `top` since that was in the demo css already.

tinayuangao: LGTM

mmalerba:
fixes #2586

PR #3365, opened 2017-03-01
feat(screenshot): Update screenshot golds for merged PR

tinayuangao:


DevVersion:
Thanks for adding the `gold` mode. It is really super clear now!

PR #3366, opened 2017-03-01
fix(input): fix chrome autofill style

### LINKED ISSUE #3355: mdInput label doesn't rise with autocomplete for type="password" fields
Bug

I would expect the label to rise up when chrome suggests passwords and automatically inputs the value.

The label stays down and the dots and label are mixed together. I'd like to add that this works just fine for input type email and text. I only have this issue on input type password.

My code: 
`<md-input-container class="form-full-width">
     <input mdInput placeholder="Password" type="password" [formControl]="complexForm.controls['password']"></md-input-container>` 
add this to a project, and refresh the page. Chrome should automatically suggest a password and input this into the password field. If i touch any key or mouse button, the label will lift. If not, the label stays down.

#### Tip to reproduce
Make a change in your ide, and let the browser update on its own, don't highlight the browser, this way you'll be sure no button is pressed, and the label will stay down.

i was unable to reproduce in plunkr as the input field wouldn't be recognized by chrome and no value would be put in automatically.

It's visually annoying that the label stays down - and unintentional i believe.

`Angular version": "2.4.8`
`angular-cli": "1.0.0-beta.25.5`
`@angular/material": "^2.0.0-beta.2`

This error started occurring when i updated to v2.0.0 beta 2 from 2.0.0 beta 1.

img_0116

I highly appreciate all the work you guys are doing!
Cheers
kind regards - Chris

mmalerba:
fixes #3355

### DIFF
--- a/src/lib/input/input-container.scss
+++ b/src/lib/input/input-container.scss
@@ -94,7 +94,7 @@ $mat-input-underline-disabled-background-image:
   // Once the autofill is committed, a change event happen and the regular md-input-container
   // classes take over to fulfill this behaviour.
   // Assumes the autofill is non-empty.
-  &:-webkit-autofill + .mat-input-placeholder.mat-float {
+  &:-webkit-autofill + .mat-input-placeholder-wrapper .mat-float {
     @include mat-input-placeholder-floating;
   }
 


dahaupt:
@kara My app is starting with the login screen and Chrome autofills the login data if available. I think we also need a fix to update the form state, because even with filled data it is invalid until I click inside the form.

Right after the app starts:
image

Or do I have to check this on a viewChange detection cycle?

mmalerba:
@dahaupt I think that might be some security feature where you have to interact with it before it actually makes the autocomplete data available to the site, you should file a separate issue though if you believe its a real problem. This PR was just about fixing the overlapping placeholder.

PR #3371, opened 2017-03-02
docs(dev_environment): replace npm install with yarn

leocaseiro:
Make sure developers use `yarn` instead of `npm`, so they use the same versions of thirdparty libraries as the `@angular/material2` team.

### DIFF
--- a/DEV_ENVIRONMENT.md
+++ b/DEV_ENVIRONMENT.md
@@ -1,12 +1,12 @@
 # Developer guide: getting your environment set up
 
 1. Make sure you have `node` installed with a version at _least_ 5.5.0.
-2. Run `npm install -g gulp` to install gulp.
+2. Run `npm install -g yarn gulp` to install yarn and gulp.
 3. Fork the `angular/material2` repo. 
 4. Clone your fork. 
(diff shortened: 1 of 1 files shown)


leocaseiro:
What was wrong with Travis?

run travis again

andrewseguin:
Can you clarify why focusing on the use of yarn helps developers? What do you mean that it helps use the same version of third party libraries?

leocaseiro:
Hi @andrewseguin, I meant in regards to the `Deterministic` of the libs. I had an issue trying to install through the npm and I switched to yarn and worked.


> npm is non-deterministic, packages are not signed, and npm does not perform any integrity checks other than a basic SHA1 hash, which is a security risk when installing system-wide apps.

Read more: https://yarnpkg.com/en/docs/install#alternatives-tab

andrewseguin:
There should be no problem using npm to install Material and we should avoid prescribing a new package manager to our users. If there is an issue related to our repository that npm encountered that needs to be fixed, please feel free to open and issue and we can discuss how to resolve that issue.

leocaseiro:
Thanks @andrewmnlv

PR #3383, opened 2017-03-02
feat(ripple): add way to globally disable ripples

DevVersion:
Currently ripples can be only disable component-wise. This isn't very elegant if users want to disable ripples for all components.
* This PR introduces a new `OpaqueToken`/ `InjectionToken` that can be used to globally disable all ripples.
* Also makes the `tab-nav-bar` ripple directive more clean. Since Angular v2.3.0 (as we recently upgraded) the inherited metadata can be used.

**Note**: For upcoming accessibility features (focus indicators) we don't want to allow developers to disable programmatic ripples.

### DIFF
--- a/src/lib/core/ripple/ripple.ts
+++ b/src/lib/core/ripple/ripple.ts
@@ -15,7 +15,7 @@ import {ViewportRuler} from '../overlay/position/viewport-ruler';
 import {RippleRef} from './ripple-ref';
 
 /** OpaqueToken that can be used to globally disable all ripples. Except programmatic ones. */
-export const MD_RIPPLES_DISABLE = new OpaqueToken('md-ripples-disable');
+export const MD_DISABLE_RIPPLES = new OpaqueToken('md-disable-ripples');
 
 @Directive({
   selector: '[md-ripple], [mat-ripple]',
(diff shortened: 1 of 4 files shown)


src/lib/core/ripple/ripple.ts
```diff
@@ -2,15 +2,20 @@ import {
...
   OnDestroy,
+  OpaqueToken,
+  Optional,
 } from '@angular/core';
 import {RippleConfig, RippleRenderer} from './ripple-renderer';
 import {ViewportRuler} from '../overlay/position/viewport-ruler';
 import {RippleRef} from './ripple-ref';
 
+/** OpaqueToken that can be used to globally disable all ripples. Except programmatic ones. */
+export const MD_RIPPLES_DISABLE = new OpaqueToken('md-ripples-disable');
```
mmalerba: nit: `MD_DISABLE_RIPPLES` or `MD_RIPPLES_DISABLED`

DevVersion: Yeah sounds better.

DevVersion:
Not really sure. I feel like this could be a useful feature when not dispatching ripples (that are not required for a11y), but this would need a good design.

src/lib/core/ripple/ripple.ts
DevVersion: Done.

mmalerba:
LGTM

kara:
@DevVersion  Can you rebase this?

DevVersion:
@kara Done!

Issue #3102, opened 2017-02-15
Checkbox/Radio button does not have any indicator to show its focus state


thanhpd:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Checkbox/Radio button has focus ripple in focused state

#### What is the current behavior?
Checkbox/Radio button does not have any indicator to show its focused state

#### What is the use-case or motivation for changing an existing behavior?
- Usability
- Material spec: https://material.io/guidelines/components/selection-controls.html

![image](https://cloud.githubusercontent.com/assets/6737911/22967479/31779f38-f399-11e6-8946-9e0bf790728e.png)

#### Which versions of Angular, Material, OS, browsers are affected?
Material 2-beta1

crisbeto:
This has been fixed in https://github.com/angular/material2/pull/1730 and https://github.com/angular/material2/pull/1731 although those PRs should be freshened up a bit.

PR #3398, opened 2017-03-02
fix(ripple): unable to remove programmatically if animations is ongoing

crisbeto:
This is something I ran into while working on #3102. For persistent ripples, the `fadeOutAll` method doesn't work if it is called before the enter animation is finished. The issue was that the set of active ripples doesn't get populated until the animation is done.

Relates to #3102.


DevVersion:
Chatting with @crisbeto about it - I intentionally didn't do that when refactoring the ripples, because now the ripples are fading-out immediately on mouseup.

crisbeto:
Closing this as per discussion with @DevVersion who had some plans for this already.

PR #3400, opened 2017-03-02
fix(ripple): fade-out-all should hide all ripples

DevVersion:
* When calling `fadeOutAll()`, all currently fading / animating ripples should hide as well.
* Adds some extra tests that ensure that ripples properly fade-in and fade-out (See #3398)

**Note**: Please forgive me for not adding the expectation messages. We are having the `.length` check in likely every test and I will create a function in a follow-up that will take care of the expectation message.

### DIFF
--- a/src/lib/core/ripple/ripple-ref.ts
+++ b/src/lib/core/ripple/ripple-ref.ts
@@ -5,6 +5,9 @@ import {RippleConfig, RippleRenderer} from './ripple-renderer';
  */
 export class RippleRef {
 
+  /** Whether the ripple finished it's fade-in animation. */
+  fadedIn: boolean = false;
+
   constructor(
     private _renderer: RippleRenderer,
     public element: HTMLElement,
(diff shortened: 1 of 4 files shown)


src/lib/core/ripple/ripple-ref.ts
```diff
@@ -5,6 +5,9 @@ import {RippleConfig, RippleRenderer} from './ripple-renderer';
  */
 export class RippleRef {
 
+  /** Whether the ripple finished it's fade-in animation. */
+  fadedIn: boolean = false;
```
crisbeto: Wouldn't it be more flexible to have this as a string (custom type) along the lines of `state`. That way we can expand it later to check for other cases, if we need to.

DevVersion: Yeah sounds like a good idea. Made the changes. Please take another look!

PR #3415, opened 2017-03-03
Verbose flex declaration for IE11

caziertm:
IE11 requires the verbose format in order display properly, otherwise the flex-basis property inherits a width of "0px";

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.scss
+++ b/src/lib/slide-toggle/slide-toggle.scss
@@ -54,7 +54,7 @@ $mat-slide-toggle-bar-track-width: $mat-slide-toggle-bar-width - $mat-slide-togg
 // It has to be a label, to support accessibility for the visual hidden input.
 .mat-slide-toggle-label {
   display: flex;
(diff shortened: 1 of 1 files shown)


caziertm:
I signed it.

DevVersion:
From what I remember, the `0%` as flex-basis are only problematic for IE11 when using column direction.

```scss
flex: 1;
flex: 1 1 0%;
```

Is your change just for safety or are there any issues we can reproduce?

caziertm:
Thanx for your quick response. I'm experiencing an issue in current development using the md-slide-toggle in the IE11 browser. 

## Without the change: 
image
**HTML**:
`<label class="mat-slide-toggle-label">
...
</label>`

**CSS**:
`.mat-slide-toggle-label{
flex:1;
}`

## With the change: 
image

**HTML** (no change):
`<label class="mat-slide-toggle-label">
...
</label>`

**CSS**:
`.mat-slide-toggle-label{
flex: 1 1 auto;
}`

PR #3431, opened 2017-03-04
fix(list-key-manager): exception when no initial active item

### LINKED ISSUE #3317: Bug(menu): Cannot read property 'disabled' of undefined
Bug

Should not cause an exception

Causes an exception:
Cannot read property 'disabled' of undefined in the following function:

    ListKeyManager.prototype._setActiveInWrapMode = function (delta, items) {
        // when active item would leave menu, wrap to beginning or end
        this._activeItemIndex =
            (this._activeItemIndex + delta + items.length) % items.length;
        // skip all disabled menu items recursively until an enabled one is reached
        if (items[this._activeItemIndex].disabled) {
            this._setActiveInWrapMode(delta, items);
        }
        else {
            this.setActiveItem(this._activeItemIndex);
        }
    };

https://material.angular.io/components/component/menu
focus on a menu item then use an arrow key

Need to implement keyboard navigation to go through a menu

Angular 2.4.5
Material 2.0.0-beta.2
Windows 10

Chrome Version 56.0.2924.87 (64-bit)
Microsoft Edge 38.14393.0.0
Microsoft EdgeHTML 14.14393

crisbeto:
Fixes an exception that is thrown when the user presses a key on ListKeyManager that doesn't have a default active item. The problem was due to the fact that we have a check for `null` a little bit down, that handles cases like this, however the active index is `undefined` by default.

Fixes #3317.

### DIFF
--- a/src/lib/core/a11y/list-key-manager.spec.ts
+++ b/src/lib/core/a11y/list-key-manager.spec.ts
@@ -240,6 +240,16 @@ describe('Key managers', () => {
         expect(TAB_EVENT.defaultPrevented).toBe(false);
       });
 
+      it('it should activate the first item when pressing down on a clean key manager', () => {
+        keyManager = new ListKeyManager<FakeFocusable>(itemList);
+
+        expect(keyManager.activeItemIndex).toBeNull('Expected active index to default to null.');
+
+        keyManager.onKeydown(DOWN_ARROW_EVENT);
+
+        expect(keyManager.activeItemIndex).toBe(0, 'Expected first item to become active.');
+      });
+
     });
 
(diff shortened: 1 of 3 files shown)


src/lib/core/a11y/list-key-manager.spec.ts
```diff
@@ -240,6 +240,16 @@ describe('Key managers', () => {
         expect(TAB_EVENT.defaultPrevented).toBe(false);
       });
 
+      it('it should activate the first item when pressing down on a clean key manager', () => {
```
kara: Nit: remove duplicate `it`. `it('should ... ')`

Issue #3316, opened 2017-02-26
build: ship flat es modules


DevVersion:
Similar to `angular/angular` ([see changelog](  https://github.com/angular/angular/blob/master/CHANGELOG.md#flat-es-modules-esm
)), Angular Material should start shipping a flat ESM file.

* See https://unpkg.com/@angular/core@4.0.0-rc.1/

At first glance, it looks like this would require some project re-structuring. But there might be also alternate approaches to archive the same release structure.

@kara Just confirming again, is this something we should do on Material as well?

DevVersion:
See #3751

PR #3443, opened 2017-03-05
build: ship esm bundle

DevVersion:
* Now ships a Flat / ESM bundle (similar as for Angular 4+ releases)
* This a big step into cleaning-up the build process. (dynamic extra compiler options)
  This means that we don't need `X` similar tsconfig files for the same module.
* Fixes sourcemap issues with `gulp-autoprefixer` and `gulp-clean-css`.

**Note**: This is not the full-switch to ESM bundles - In future PRs we will remove the single ES-files and can just link to the ES bundle. Similar as in angular.

References #3316


DevVersion:
I did some refactoring of the gulp tasks. We no longer need multiple `tsconfig` files for the same module.

package.json
```diff
@@ -51,6 +51,7 @@
     "@types/node": "^7.0.4",
     "@types/run-sequence": "^0.0.28",
     "@types/rx": "2.5.33",
+    "autoprefixer": "^6.7.6",
```
DevVersion: Removed all the autoprefixer stuff. So I can just rebase without conflicts if it's merged.

tools/gulp/tasks/components.ts
```diff
@@ -1,139 +1,150 @@
...
   DIST_COMPONENTS_ROOT, PROJECT_ROOT, COMPONENTS_DIR, HTML_MINIFIER_OPTIONS, LICENSE_BANNER
 } from '../constants';
 import {
-  sassBuildTask, tsBuildTask, execNodeTask, copyTask, sequenceTask,
+  sassBuildTask, tsBuildTask, execNodeTask, sequenceTask,
   triggerLivereload
 } from '../task_helpers';
 
-// No typings for these.
+// These imports lack of types.
```
DevVersion: Sounds good

tools/gulp/tasks/components.ts
```diff
@@ -1,139 +1,150 @@
...
-task(':watch:components', () => {
-  watch(path.join(COMPONENTS_DIR, '**/*.ts'), ['build:components', triggerLivereload]);
-  watch(path.join(COMPONENTS_DIR, '**/*.scss'), ['build:components', triggerLivereload]);
-  watch(path.join(COMPONENTS_DIR, '**/*.html'), ['build:components', triggerLivereload]);
-});
-
+/** Builds components for Angular Material releases */
+task(':build:components:release', sequenceTask(
+  ':build:components:bundle:umd',
+  ':build:components:bundle:es',
```
DevVersion: Yeah. Makes sense!

tools/gulp/util/ts-compiler.ts
```diff
@@ -0,0 +1,56 @@
...
+/** Formats the TypeScript diagnostics into a error string. */
+export function formatDiagnostics(diagnostics: ts.Diagnostic[]): string {
+  return diagnostics.map(diagnostic => {
+    let res = ts.DiagnosticCategory[diagnostic.category];
+
+    if (diagnostic.file) {
+      let {line, character} = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start);
+
+      res += ' at ' + diagnostic.file.fileName + ':';
+      res += (line + 1) + ':' + (character + 1) + ':';
```
jelbourn: Use template strings?

DevVersion:
@jelbourn  Done! Let me know if you want me to make any other changes.

tools/gulp/tasks/components.ts
```diff
@@ -1,139 +1,150 @@
...
+  ':build:components:bundle:umd',
+  ':build:components:bundle:esm',
+  ':build:components:ngc'
+));
 
-/** Builds component typescript only (ESM output). */
-task(':build:components:ts', tsBuildTask(path.join(COMPONENTS_DIR, 'tsconfig-srcs.json')));
+/** Builds components typescript in ES5, ES6 target. For specs Karma needs CJS output. */
+task(':build:components:ts:es5', tsBuildTask(tsconfigPath, { target: ScriptTarget.ES5 }));
+task(':build:components:ts:es6', tsBuildTask(tsconfigPath, { target: ScriptTarget.ES6 }));
```
d3viant0ne: @DevVersion - Worth noting that `ES6` as a ScriptTarget doesn't work for newer versions of TypeScript iirc.

Options are `ES2015, ES2016, ES2017, ES3, ES5, ESNext & Latest`

d3viant0ne: That that's all of them off the top of my head.

crisbeto: Good point. This does fail on the Angular 4 branch.

DevVersion: Yeah, I noticed it as well, when I based the upcoming packaging PR on top of the v4 compatibility PR. Just FYI: This stuff will be removed with the new version of `tsc-wrapped` anyways.

PR #3450, opened 2017-03-05
fix(card): removed header height

### LINKED ISSUE #3288: md-card-title overlaps when text is too long
Bug

Expected behaviour would be that the elements below the title will expand and no overlap happens

the title overlaps the elements below

See the plunkr below of the official example for md-card from https://material.angular.io/components/component/card where I just added a bit longer text for the md-card-title.

https://embed.plnkr.co/ub7P6byKCe30PNMiu2r1/

when using the md-card in a dynamic way (lets say a http call and showing the response in md-cards), one should be not dependent on the length of the title. (no manual margins desired)

see plunkr

this applies also when using a bigger font-size and also on other elements.

EladBezalel:
- long titles was overlapping content cause of the height that was applied on the header

fixes #3288

### DIFF
--- a/src/lib/card/card.scss
+++ b/src/lib/card/card.scss
@@ -93,12 +93,9 @@ $mat-card-header-size: 40px !default;
 .mat-card-header {
   display: flex;
   flex-direction: row;
-  height: $mat-card-header-size;
-  margin: -8px 0 16px 0;
 }
 
 .mat-card-header-text {
-  height: $mat-card-header-size;
   margin: 0 8px;
 }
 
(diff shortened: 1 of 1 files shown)


kara:
@EladBezalel Looks good! One thing though: can you check on the header avatar? When long title text is added, it looks like its width still isn't getting applied properly, so it looks stretched out.

EladBezalel:
i will look into it!

EladBezalel:
@kara please re-check, nice catch :)

PR #3462, opened 2017-03-06
fix(package.json): Fix failing `npm run api` command

dalelotts:
api script was running the incorrect gulp command.


dalelotts:
CLA signed.

Issue #2901, opened 2017-02-02
(ripple): make sure ripples are on spec - speed & color


mmalerba:
button seems faster than is shown here: https://material.io/guidelines/motion/material-motion.html#material-motion-why-does-motion-matter. radio and checkbox explicitly set a faster speed, maybe this should be the default?

color also seems wrong for button ripples, primary/accent colored buttons still have gray ripples

shalugin:
Is it possible to define global ripple speed?

DevVersion:
Hmm not yet, but I'm up to add it. Are there any specific use-cases to argue with?

shalugin:
> Are there any specific use-cases to argue with?

I don't know about these cases.

DevVersion:
I was just wondering if there is any situation where you want to modify the global ripple speed?

shalugin:
Current speed settings is too slow (version 2.0.0-beta.2).

PR #3463, opened 2017-03-06
feat(ripple): support for global ripple options

### LINKED ISSUE #3454: Ripple Config
Change the ripple config for all buttons for example

It prevents using css hacks and it's possible to fine tune the ripple for all elements for a certain type

DevVersion:
* Adds a new DI token that can be used to specify global ripple options.
* The `RippleRef` options are not related to the `MD_DISABLE_RIPPLES` DI token (for API convenience)

@mmalerba I'm not really sure about the naming of the DI token. Let me know if you have any better ideas.

Closes #3454

### DIFF
--- a/src/lib/core/ripple/ripple.ts
+++ b/src/lib/core/ripple/ripple.ts
@@ -14,7 +14,7 @@ import {RippleConfig, RippleRenderer} from './ripple-renderer';
 import {ViewportRuler} from '../overlay/position/viewport-ruler';
 import {RippleRef} from './ripple-ref';
 
-/** OpqaueToken that can be used to specify the global ripple options. */
+/** OpaqueToken that can be used to specify the global ripple options. */
 export const MD_RIPPLE_GLOBAL_OPTIONS = new OpaqueToken('md-ripple-global-options');
 
 export type RippleGlobalOptions = {
@@ -75,13 +75,18 @@ export class MdRipple implements OnChanges, OnDestroy {
   /** Renderer for the ripple DOM manipulations. */
   private _rippleRenderer: RippleRenderer;
 
+  /** Options that are set globally for all ripples. */
+  private _globalOptions: RippleGlobalOptions;
+
   constructor(
     elementRef: ElementRef,
     ngZone: NgZone,
     ruler: ViewportRuler,
-    @Optional() @Inject(MD_RIPPLE_GLOBAL_OPTIONS) private _globalOptions: RippleGlobalOptions
+    // Type needs to be `any` because of https://github.com/angular/angular/issues/12631
+    @Optional() @Inject(MD_RIPPLE_GLOBAL_OPTIONS) globalOptions: any
   ) {
     this._rippleRenderer = new RippleRenderer(elementRef, ngZone, ruler);
+    this._globalOptions = globalOptions ? globalOptions : {};
   }
 
   ngOnChanges(changes: SimpleChanges) {
--- a/src/lib/core/ripple/ripple.ts
+++ b/src/lib/core/ripple/ripple.ts
@@ -14,8 +14,13 @@ import {RippleConfig, RippleRenderer} from './ripple-renderer';
 import {ViewportRuler} from '../overlay/position/viewport-ruler';
 import {RippleRef} from './ripple-ref';
 
-/** OpaqueToken that can be used to globally disable all ripples. Except programmatic ones. */
-export const MD_DISABLE_RIPPLES = new OpaqueToken('md-disable-ripples');
+/** OpqaueToken that can be used to specify the global ripple options. */
+export const MD_RIPPLE_GLOBAL_OPTIONS = new OpaqueToken('md-ripple-global-options');
+
+export type RippleGlobalOptions = {
+  disabled?: boolean;
+  baseSpeedFactor?: number;
+};
 
 @Directive({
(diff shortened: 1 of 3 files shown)


src/lib/core/ripple/ripple.ts
```diff
@@ -102,11 +107,11 @@ export class MdRipple implements OnChanges, OnDestroy {
...
   /** Ripple configuration from the directive's input values. */
   get rippleConfig(): RippleConfig {
-    return {
+    return extendObject({
       centered: this.centered,
       speedFactor: this.speedFactor,
       radius: this.radius,
       color: this.color
-    };
+    }, this._globalRippleConfig);
```
mmalerba: How about adding a disabled field to RippleConfig? then you don't have to have a separate DI token

src/lib/core/ripple/ripple.ts
```diff
@@ -13,10 +13,14 @@ import {
 import {RippleConfig, RippleRenderer} from './ripple-renderer';
 import {ViewportRuler} from '../overlay/position/viewport-ruler';
 import {RippleRef} from './ripple-ref';
+import {extendObject} from '../util/object-extend';
 
 /** OpaqueToken that can be used to globally disable all ripples. Except programmatic ones. */
 export const MD_DISABLE_RIPPLES = new OpaqueToken('md-disable-ripples');
 
+/** OpaqueToken that can be used to specify a global ripple configuration. */
+export const MD_RIPPLES_CONFIG = new OpaqueToken('md-ripples-config');
```
mmalerba: Maybe we could call it `MD_RIPPLE_OVERRIDE(S)` to make it obvious that these are not global defaults, but rather overrides that cause the individual ripple options to be ignored

src/lib/core/ripple/ripple.ts
DevVersion: That's what I was trying to avoid with the config. The config is for the single ripple references and doing stuff like below is just not a good API.

```ts
ripple.launch(0, 0, { disabled: true });
```

> Why would you want to launch a ripple that is disabled?

You are not disabling the single ripple-ref, you are disabling the whole ripple triggers.

src/lib/core/ripple/ripple.ts
DevVersion: I'm fine with that. Let me change that. Also you asked about the motivation of this change. 
* It's just really convenient if developers can globally modify the `speedFactor` for example.

See https://github.com/angular/material2/issues/2901#issuecomment-283648158

I'm personally also fine with only supporting `MD_RIPPLES_SPEEDFACTOR`, but I think that this feature might help us in the future as well.

mmalerba: That comment sounds like he was looking for a way to address the bug himself before we got around to fixing it, not necessarily a feature request.

DevVersion: There is also a issue just for it specifically: https://github.com/angular/material2/issues/3454

src/lib/core/ripple/ripple.ts
mmalerba: Ok then maybe this PR is conflating two things: the per-ripple options, and the global ripple options. I agree that disabled doesn't make sense for individual ripples, and I think speed factor doesn't make sense for the global settings. The global setting should be the base speed, and then individual ripples define a speed factor that is multiplied together with the global base speed. That way we preserve the relative speeds when setting the global.

DevVersion: That's a good point. I feel like it would then make more sense to just have something like `MD_RIPPLES_SPEEDFACTOR` instead?

src/lib/core/ripple/ripple.ts
mmalerba: That issue is actually asking for a different feature: the ability to override the ripple settings for all buttons (not globally for all ripples)

src/lib/core/ripple/ripple.ts
mmalerba: Or to combine disabled & base speed into a global config object (which is completely unrelated to the individual config objects). Though I still question if anyone even wants this feature

src/lib/core/ripple/ripple.ts
DevVersion: Hm that's true. I actually misread it. But anyways, I think providing such component only for the button would be unnecessary. We would need to do this for likely every components.

Should we close that PR and the issue? (I'm fine with that)

mmalerba: If you want to do base speed instead of overriding the speed factor I'm ok with that, seems more potentially useful. Or we can just close it and wait until we have a compelling use case

DevVersion: Hm I think that base speed would be a good feature. I will rework the PR. Should we make it a config object then? Or just a `OpaqueToken` for the base-speed (like for disabling ripples)

mmalerba: I would combine them under a single token `MD_RIPPLE_GLOBAL_OPTIONS` or something

DevVersion: Made the changes. 

> Note: `@Optional()` sets the constructor param to `null`. So a TypeScript default value does unfortunately not work.

DevVersion:
@mmalerba Ran into a Angular issue. Had to use a small workaround for it.

PR #3468, opened 2017-03-06
feat(datepicker): add mdDatepickerToggle & compatibility w/ md-input-container

mmalerba:


src/demo-app/datepicker/datepicker-demo.html
```diff
@@ -1,6 +1,25 @@
...
+<br>
+<p>
+  <input [mdDatepicker]="dp" [(ngModel)]="date">
+  <md-datepicker-trigger [for]="dp"></md-datepicker-trigger>
+  <md-datepicker #dp [touchUi]="touch"></md-datepicker>
+</p>
+<p>
+  <md-input-container>
+    <input mdInput [mdDatepicker]="dp2" [(ngModel)]="date">
+    <md-datepicker-trigger mdSuffix [for]="dp2"></md-datepicker-trigger>
```
kara: I'm guessing you've already thought this through, but any way to avoid requiring this element? If at all possible, it feels like this should be built in somewhere. Can we have the `[mdDatepicker]` directive add the class and the open method?

src/lib/input/input-container.ts
```diff
@@ -379,4 +381,8 @@ export class MdInputContainer implements AfterContentInit {
 
     this._mdInputChild.ariaDescribedby = ids.join(' ');
   }
+
+  getPopupConnectionElementRef(): ElementRef {
```
kara: It feels like this should be named for the underline, given that there may be other non-popup components that might want to hook in here.

src/demo-app/datepicker/datepicker-demo.html
mmalerba: This allows for more flexibility, the icon trigger can be in the prefix, suffix, or completely outside of the input-container.

mmalerba:
comments addressed

src/demo-app/datepicker/datepicker-demo.html
kara: What about an overridable default?

kara: How about making the datepicker-trigger a button? I think part of the weirdness for me is that the input should itself be a trigger, so adding something else called "trigger" threw me off.  

Jeremy and I were thinking maybe: 

```html
<input [mdDatepicker]="dp">
<button mdDatepickerToggle></button>
<md-datepicker #dp></md-datepicker>
```

mmalerba:
renamed trigger, PTAL

Issue #3354, opened 2017-03-01
Feature request: Add Input "IgnoreOrder" to Grid-List


antoschd:
#### Feature request
New input to ignore initial order of grid tiles, so that all gaps are filled

#### Current behavior
Order is allways respected. This can cause gaps.
http://plnkr.co/edit/Pl5Zfkrnvu7L9WFv6YOW?p=preview

#### Use-case / motivation for changing an existing behavior
Use-case: Order doesn't matter and data is dynamic. 
For example: We want to use the grid-list to display videos and images. Videos should be twice as big as images, colspan = 2 and rowspan = 2. If the data wouldn't change we could easily order them once, so that there is no gap. But images and videos are loaded dynamically, according to user selection. 

#### Solution idea
Change tile-coordinator tracker to a two-dimensional array.
Get current "tracker row" with this.tracker[this.rowIndex]
On _nextRow add another "row" to the tracker with decrement values of previous.  
Add input "ignoreOrder" on grid-list.
Reset rowIndex and columnIndex if "ignoreOrder".

if (ignoreOrder) {
   this.rowIndex = 0;
   this.columnIndex = 0;
} else {
   this.columnIndex = gapStartIndex + tile.colspan; // Line 64 tile-coordinator
}

M-a-c:
hey @kara can you mark this one as having pr?

jelbourn:
Closing this since there are no plans to support this feature.

PR #3473, opened 2017-03-07
feat (grid-list): add possibility to ignore order

antoschd:
new input to ignore initial order of grid tiles, so that all gaps are filled

#3354


antoschd:
@DevVersion can you help me with the browserstack error or give me a hint what the problem could be?

jelbourn:
cc @kara needs discussion as to whether we want to support a feature like this

jelbourn:
@antoschd this is a rather larger feature that would require some discussion before going down the implementation path. Feel free to open a proposal issue for discussion with the high-level approach and API.

antoschd:
@jelbourn thanks for your feedback, I fully understand. Here is the proposal: https://github.com/angular/material2/issues/4362

PR #3476, opened 2017-03-07
chore(tools/gulp): add html minifying task to generated docs

riavalon:


tools/gulp/tasks/docs.ts
```diff
@@ -61,6 +62,12 @@ task('api-docs', () => {
   return docs.generate();
 });
 
+task('minify-html-docs', ['api-docs'], () => {
+  return gulp.src('dist/docs/api/*.html')
+    .pipe(htmlmin({collapseWhitespace: true}))
```
DevVersion: Please use the `HTML_MINIFIER_OPTIONS` constant for `htmlmin`.

https://github.com/angular/material2/blob/ca56ebbf6ba83d11ae2dc919a3143e85d4f6a397/tools/gulp/constants.ts#L20

Angular input bindings are case-sensitive and will now break.

Issue #3368, opened 2017-03-01
tabindex doesn't work on md-checkbox


jloosli:
#### Bug, feature request, or proposal:
I am not able to disable tabbing to a md-checkbox.

#### What is the expected behavior?
`<md-checkbox tabIndex="-1"></md-checkbox>` should not be skipped over when tabbing through a form.

#### What is the current behavior?
The checkbox receives focus when tabbing through the form. In addition, there isn't any visual indication that the checkbox has focus, so it's difficult to tell where I am in the form unless I press the spacebar to toggle that particular entry.

#### What are the steps to reproduce?

http://plnkr.co/edit/0P5w3xAmYWhx4ZrEoWUU?p=preview

#### What is the use-case or motivation for changing an existing behavior?

I have checkbox that I want on the form, but not to be tabbed to.

#### Which versions of Angular, Material, OS, browsers are affected?
Tested on Angular 2.4.8, Angular/Material 2.0.0-beta.2, Chrome

#### Is there anything else we should know?
That more or less covers it.

DevVersion:
You mentioned in the issue that you are using `tabIndex` in camelcase. But in your Plunkr it's actually just `tabindex`.

For those components, just using `tabIndex` in camelcase should do the trick.

http://plnkr.co/edit/4ScdvxMz8q2Q5RC6bILh?p=preview

jloosli:
Whoops..I had the wrong title down...it's actually md-checkbox that tabindex doesn't work. I also corrected the example to have a lowercase tabindex

crisbeto:
WRT the focus indication, it is being tracked via https://github.com/angular/material2/issues/3102.

jloosli:
@DevVersion So I just came back to this and see that it does indeed work with the camelcase tabIndex. Is that the plan to keep it like that? I'm glad that it works, but given that the [spec for tabindex](https://www.w3.org/TR/html5/editing.html#sequential-focus-navigation-and-the-tabindex-attribute) is all lowercase, I wonder if that will cause additional confusion (beyond just myself).

DevVersion:
Not sure. I see that it can cause confusion here, but I think the `tabIndex` is in camel-cause because when using an input binding, you are actually setting the native `tabIndex` property.

That's why for native inputs you can also do 
```html
<input [tabIndex]="2">
```

Not sure if we should support both.

jelbourn:
@tinayuangao are you able to confirm if this is an issue?

tinayuangao:
[...]

PR #3479, opened 2017-03-07
fix(select): unable to set a tabindex

### LINKED ISSUE #3474: md-select tabindex issue
For `<md-select/>`, tabindex attribute value is set to 0.

tabindex should have the given value.

It is rendering tabindex="0" when using `<md-select tabindex="6"></md-select>`

- Angular 2.4.9
- Material 2.0.0-beta.2

Issue in the following lines:

```
export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestroy {
  /** Returns the correct tabindex for the select depending on disabled state. */
  _getTabIndex() {
    return this.disabled ? '-1' : '0';
  }
}
```

crisbeto:
Fixes users not being able to override the `tabIndex` on `md-select`.

Fixes #3474.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1081,8 +1081,15 @@ describe('MdSelect', () => {
         expect(select.getAttribute('aria-label')).toEqual('Food');
       });
 
-      it('should set the tabindex of the select to 0', () => {
+      it('should set the tabindex of the select to 0 by default', () => {
         expect(select.getAttribute('tabindex')).toEqual('0');
+      });
+
+      it('should be able to override the tabindex', () => {
+        fixture.componentInstance.tabIndexOverride = 3;
+        fixture.detectChanges();
+
+        expect(select.getAttribute('tabindex')).toBe('3');
       });
 
       it('should set aria-required for required selects', () => {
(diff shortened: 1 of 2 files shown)


src/lib/select/select.spec.ts
```diff
@@ -1362,7 +1369,8 @@ describe('MdSelect', () => {
   selector: 'basic-select',
   template: `
     <div [style.height.px]="heightAbove"></div>
-    <md-select placeholder="Food" [formControl]="control" [required]="isRequired">
+    <md-select placeholder="Food" [formControl]="control" [required]="isRequired"
+      [tabIndex]="tabIndexOverride">
```
kara: It would be nice to use `tabindex` instead of `tabIndex`. That way users can set tabindex as they normally would without having to learn a different API for the select.

```html
<md-select tabindex="6">
</md-select>
```

crisbeto: I went this route to be consistent with the other modules (e.g. the checkbox). IIRC we named it `tabIndex`, because the property on the native DOM element is called `tabIndex`.

DevVersion: I have been asking in https://github.com/angular/material2/issues/3368 about it - I thought about supporting both (for convenience).

kara: Chatted with Jeremy about this, and it makes sense to support both. Since the HTML property is `tabIndex`, you should bind to `tabIndex`. But people should still be able to use `tabindex="5"` without binding the way they normally would. We can achieve this with another @Attribute binding.

crisbeto: Added @kara, can you take another look?

PR #3486, opened 2017-03-07
feat: support strictNullChecks

### LINKED ISSUE #3465: Turn on strictNullChecks
Is there any appetite for turning on typescript strictNullChecks mode? This is a recommended setting for new code bases. It's something I've already done for a decent portion of the material code base internally at my org so I have some experience with it.

If so, I'd be interested in submitting some PRs to work at getting us to that point.

crisbeto:
Turns on `strictNullChecks` everywhere and refactors all of the code to make it compliant. This makes us consistent with the core Angular modules which support `strictNullChecks` after 4.0.

Fixes #3465.

**Note:** This is pretty conflict-prone and is blocked on a couple of things:
1. We need to bump to Angular 4.0+ to be able to compile.
2. `@angular/forms` doesn't support `strictNullChecks` as of 4.0 RC2, which means that our code won't be able to compile until it does.

### DIFF
--- a/tools/gulp/tasks/release.ts
+++ b/tools/gulp/tasks/release.ts
@@ -35,10 +35,10 @@ task(':publish:whoami', execTask('npm', ['whoami'], {
 task(':publish:logout', execTask('npm', ['logout']));
 
 
-function _execNpmPublish(label: string): Promise<{}> {
+function _execNpmPublish(label: string): Promise<{}|void> {
   const packageDir = DIST_COMPONENTS_ROOT;
   if (!statSync(packageDir).isDirectory()) {
-    return;
+    return Promise.resolve();
   }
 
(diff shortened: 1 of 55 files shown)


tools/gulp/tasks/release.ts
```diff
@@ -49,7 +49,7 @@ function _execNpmPublish(label: string): Promise<{}> {
   console.log(`Publishing material...`);
 
   const command = 'npm';
-  const args = ['publish', '--access', 'public', label ? `--tag` : undefined, label || undefined];
+  const args = ['publish', '--access', 'public', label ? `--tag` : '', label || ''];
```
crisbeto: Note to self: this is broken and should be reworked.

crisbeto:
Closing in favor of #5094.

PR #3489, opened 2017-03-08
fix(input): make all icons in input containers the correct size

### LINKED ISSUE #2586: md-icon as input prefix / prefix on wide inputs

marshall007:
Fixes #2586 and #3229. This is a more general solution to #3342, which also handles the `md-hint` case. See https://github.com/angular/material2/issues/2586#issuecomment-284899825 for more details.

I haven't noticed any issues doing this at the `.mat-input-container` level. If there are any I think this solution is still more robust overall and we should handle those specific cases individually.

### DIFF
--- a/src/demo-app/baseline/baseline-demo.scss
+++ b/src/demo-app/baseline/baseline-demo.scss
@@ -9,13 +9,6 @@
   width: 100%;
 }
 
-.demo-icons {
-  font-size: 100%;
-  height: inherit;
-  vertical-align: top;
-  width: inherit;
-}
-
 .demo-card {
   margin: 16px;
 }
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -40,7 +40,10 @@
         <td>
           <md-input-container class="demo-full-width">
             <input mdInput #postalCode maxLength="5" placeholder="Postal Code" value="94043">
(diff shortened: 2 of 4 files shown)


marshall007:
@mmalerba FYI, I don't think your initial PR actually fixes #2586 (at least for icons). I've just added an icon prefix/suffix example to the demo app to demonstrate this behavior. Let me know if you have thoughts on how to fix it and I'll update my PR accordingly.

screen shot 2017-03-07 at 7 15 26 pm

marshall007:
All the browsers I've tried compute the `mdSuffix` width using the ligature text content instead of the glyph width. I can't find a solution to this that doesn't involve flex layouts, but that seems to be a no-no in this case per the coding standards.

marshall007:
Ah, this was my bad. I was using `edit_mode` instead of `mode_edit` as the icon text. Apparently `edit` produces a valid glyph and thus `_mode` just results in extra whitespace. Apologies for the confusion, this is good to go.

screen shot 2017-03-07 at 11 04 25 pm

Issue #3375, opened 2017-03-02
JAWS - Autocomplete doesn't announce dropdown closing for invalid input


DaveBest99:
#### Bug, feature request, or proposal:
Request: Typed input should only match what is in the list box.

#### What is the expected behavior?
In a non-editable auto-complete, any letters that do not result in a match from the list should be ignored, the drop down list of choices remains static until the user presses Escape to clear the text field, Backspace to remove some of the letters previously typed, or types an additional letter that results in a valid list of choices. In an editable auto-complete, if no choices match the letters typed, the drop down list closes. 

#### What is the current behavior?
Currently anything can be typed into the Edit box.

#### What is the use-case or motivation for changing an existing behavior?
Help users from making mistakes. A non-editable list box should only allow typed text that matches the list items, and an editable list box can accept typed text that does not match the list items, but this is confusing in a list of States.

#### Is there anything else we should know?
The only time the Edit box gets screen reader focus is when an invalid character is typed. That is, as long as there is a list item with the sequence of typed letters, the focus will move through the list, but a non-recognizable sequence will bring focus to the Edit box. Deleting the last letter typed (pressing Backspace) will move focus to the valid list item but it is not announced by the screen reader. That is, the screen reader user does not know what list item has focus (Note,although JAWS does not announce the State name, it is showned on the braille display). The user can press up/down arrow keys through the list but never get Edit box focus. Focus should stay on the Edit box until the arrow key is pressed, and an aria-alert can read the list item in focus.

kara:
Related: Related: https://github.com/angular/material2/issues/3416

kara:
Currently we hide the panel when there are no options, but it looks like aria-expanded is not updated properly, so there's no indication that the list is closed. That's definitely a bug. It's also weird when the invalid letter is deleted because it doesn't announce that the list has re-opened. Looking into it.

DaveBest99:
I think Github3416 offers a cleaner solution for opening and closing the list.

PR #3494, opened 2017-03-08
fix(autocomplete): aria-expanded should be updated when panel hides

kara:
When the panel is hidden, aria-expanded was still set to true because we hadn't updated `panelOpen`. As `panelOpen` is part of the public API, we should only set it to true when the panel is open *and* visible.

Closes https://github.com/angular/material2/issues/3375

### DIFF
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -66,35 +66,39 @@ describe('MdAutocomplete', () => {
       input = fixture.debugElement.query(By.css('input')).nativeElement;
     });
 
-    it('should open the panel when the input is focused', () => {
+    it('should open the panel when the input is focused', async(() => {
       expect(fixture.componentInstance.trigger.panelOpen)
           .toBe(false, `Expected panel state to start out closed.`);
 
       dispatchFakeEvent(input, 'focus');
-      fixture.detectChanges();
+      fixture.whenStable().then(() => {
+        fixture.detectChanges();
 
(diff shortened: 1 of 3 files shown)


src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -748,20 +750,43 @@ describe('MdAutocomplete', () => {
...
 
+    it('should set aria-expanded properly when the panel is hidden', async(() => {
+        fixture.componentInstance.trigger.openPanel();
+
+        fixture.whenStable().then(() => {
+          fixture.detectChanges();
+          expect(input.getAttribute('aria-expanded'))
+              .toBe('true', 'Expected aria-expanded to be true while panel is open.');
+
+          typeInElement('zz', input);
```
mmalerba: it would be good to move this method to test utils

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -748,20 +750,43 @@ describe('MdAutocomplete', () => {
...
+          fixture.whenStable().then(() => {
+            fixture.detectChanges();
+
+            fixture.whenStable().then(() => {
+              fixture.detectChanges();
+              expect(input.getAttribute('aria-expanded'))
+                .toBe('false', 'Expected aria-expanded to be false when panel hides itself.');
```
mmalerba: nit: indentation

jelbourn:
LGTM aside from Miles' comments

kara:
@mmalerba @jelbourn Comments addressed

PR #3517, opened 2017-03-08
docs: remove mat stylers from docs

andrewseguin:


tinayuangao:
LGTM

src/lib/list/list.ts
```diff
@@ -69,7 +75,10 @@ export class MdDividerCssMatStyler {}
 })
 export class MdListAvatarCssMatStyler {}
 
-/* Need directive to add mat- CSS styling */
+/**
+ * Directive whose purpose is to add the mat- CSS styling to this selector.
+ * @docs-private
+ */
```
kara: Missed the subheader styler below it?

andrewseguin: Good catch, thanks

PR #3519, opened 2017-03-08
docs(select): add docs for static placeholder

kara:


src/lib/select/select.md
```diff
@@ -37,6 +37,20 @@ class MyComp {
...
 ```
 
+### Setting a static placeholder
+
+It's possible to turn off the placeholder's floating animation using the `floatPlaceholder` property. It accepts one of three string options:
+- `'auto'`: This is the default floating placeholder animation. It will float up when a selection is made.
+- `'never'`: This makes the placeholder static. Rather than floating, it will disappear once a selection is made.
+- `'always'`: This makes the placeholder permanently float above the input. It will not animate up or down.
+    
+*my-comp.html*
```
jelbourn: Do you need the `my-comp.html`?

kara: I can remove. It's just to match the other examples, but since this doesn't have an accompanying TS portion, should probably go.

kara: Done!

PR #3525, opened 2017-03-09
class alias tags in html generated from markdown

mmalerba:


tools/gulp/tasks/docs.ts
```diff
@@ -36,6 +55,15 @@ gulp.task('markdown-docs', () => {
         }
       }))
       .pipe(transform(transformMarkdownFiles))
+      .pipe(dom(function () {
+        MARKDOWN_TAGS_TO_CLASS_ALIAS.forEach((tag) => {
```
jelbourn: Move this into its own function like `fixMarkdownDocLinks`?

tools/gulp/tasks/docs.ts
```diff
@@ -36,6 +55,15 @@ gulp.task('markdown-docs', () => {
...
+          Array.prototype.slice.call(this.querySelectorAll(tag)).forEach((el: any) => {
```
jelbourn: Would it work to do 
```
for (let element of this.querySelectorAll(tag))
``` 
?

tools/gulp/tasks/docs.ts
```diff
@@ -19,6 +20,24 @@ const EXAMPLE_PATTERN = /<!--\W*example\(([^)]+)\)\W*-->/g;
...
+    'h2',
+    'h3',
+    'h4',
+    'h5',
+    'li',
+    'ol',
+    'p',
+    'table',
+    'tbody',
+    'td',
```
jelbourn: `tr`?

tools/gulp/tasks/docs.ts
```diff
@@ -19,6 +20,24 @@ const EXAMPLE_PATTERN = /<!--\W*example\(([^)]+)\)\W*-->/g;
...
+// for styling purposes.
+const MARKDOWN_TAGS_TO_CLASS_ALIAS = [
+    'a',
+    'h1',
+    'h2',
+    'h3',
+    'h4',
+    'h5',
+    'li',
+    'ol',
```
jelbourn: `ul`?

tools/gulp/tasks/docs.ts
```diff
@@ -95,3 +117,20 @@ function fixMarkdownDocLinks(link: string, filePath: string): string {
...
   return `guide/${baseName}`;
 }
+
+/**
+ * Returns a function to be called with an HTML document as its context that aliases HTML tags by
+ * adding a class consisting of a prefix + the tag name.
+ * @param classPrefix The prefix to use for the alias class.
+ */
+function createTagNameAliaser(classPrefix: string) {
+  return function() {
```
jelbourn: nit: can be an arrow function

tools/gulp/tasks/docs.ts
```diff
@@ -95,3 +117,20 @@ function fixMarkdownDocLinks(link: string, filePath: string): string {
...
+    MARKDOWN_TAGS_TO_CLASS_ALIAS.forEach((tag) => {
```
jelbourn: nit: parens aren't needed

tools/gulp/tasks/docs.ts
mmalerba: It actually won't get the correct context if I use an arrow fn

PR #3535, opened 2017-03-09
fix(autocomplete): "undefined" being displayed on empty control with ngModel

### LINKED ISSUE #3529: Autocompleted mdInput with undefined ngModel shows "undefined" instead of empty text
bug

Autocompleted mdInput binded with undefined ngModel should display empty text instead of "undefined".

screen shot 2017-03-09 at 11 44 13
Regular mdInput shows empty text in case of ngModel is undefined.
Autocompleted mdInput binded with undefined ngModel shows "undefined".

Here is the plunker with description
http://plnkr.co/edit/5jrByGdNel0saoQe7obm

All mdInput should behave consistently.

Tested with @angular/material 2.0.0-beta.2

crisbeto:
Fixes "undefined" being displayed as the autocomplete input's value, if the value is undefined and the input uses `ngModel`.

Fixes #3529.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -292,8 +292,8 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   }
 
   private _setTriggerValue(value: any): void {
-    this._element.nativeElement.value =
-        this.autocomplete.displayWith ? this.autocomplete.displayWith(value) : value;
+    let toDisplay = this.autocomplete.displayWith ? this.autocomplete.displayWith(value) : value;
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -292,8 +292,8 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   }
 
   private _setTriggerValue(value: any): void {
-    this._element.nativeElement.value =
-        this.autocomplete.displayWith ? this.autocomplete.displayWith(value) : value;
+    let toDisplay = this.autocomplete.displayWith ? this.autocomplete.displayWith(value) : value;
```
kara: const

PR #3540, opened 2017-03-09
fix: coverage issue with saucelabs

DevVersion:
* Saucelabs seems to have issues with Edge and the `karma-coverage` plugin.

* Running on Browserstack seems to be work without any flakiness and also makes the build faster (runs less browsers than on saucelabs)


DevVersion:
@jelbourn Those should fix the Travis failures on `master`.

PR #3542, opened 2017-03-09
feat(datepicker): add aria-* attrs and keyboard bindings to datepicker input

mmalerba:


src/lib/datepicker/datepicker-trigger.ts
```diff
@@ -0,0 +1,26 @@
+import {ChangeDetectionStrategy, Component, Input, ViewEncapsulation} from '@angular/core';
+import {MdDatepicker} from './datepicker';
+
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-datepicker-trigger, mat-datepicker-trigger',
```
kara: I thought we renamed this to `mdDatepickerToggle`?  Rebase error?

src/lib/input/input-container.ts
```diff
@@ -381,4 +383,8 @@ export class MdInputContainer implements AfterContentInit {
 
     this._mdInputChild.ariaDescribedby = ids.join(' ');
   }
+
+  getPopupConnectionElementRef(): ElementRef {
```
kara: Didn't we decide to remove this?

src/lib/datepicker/datepicker.ts
```diff
@@ -27,6 +27,9 @@ import {MdDatepickerInput} from './datepicker-input';
 import {CalendarLocale} from '../core/datetime/calendar-locale';
 
 
+let datepickerUid = 0;
```
kara: Add comment about what this is? We should also probably prefix with an underscore to be consistent with other counters.

mmalerba:
comments addressed

PR #3553, opened 2017-03-10
fix(tab): use MD_RIPPLE_GLOBAL_OPTIONS in tab

tinayuangao:


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -83,7 +83,7 @@ export class MdTabLink {
 })
 export class MdTabLinkRipple extends MdRipple {
   constructor(elementRef: ElementRef, ngZone: NgZone, ruler: ViewportRuler,
-              @Optional() @Inject(MD_DISABLE_RIPPLES) forceDisableRipples: boolean) {
+              @Optional() @Inject(MD_RIPPLE_GLOBAL_OPTIONS) forceDisableRipples: boolean) {
```
crisbeto: I was just about to submit a PR for this :D. I think the parameter here has to be renamed as well.

DevVersion: Can we rename that to `globalOptions` as well?

tinayuangao: Updated. Please take another look. Thanks

mikeumus:
:+1: 

and how do we use this exactly?

PR #3555, opened 2017-03-10
fix(slide-toggle): invalid change events with no new value

### LINKED ISSUE #3526: Slide toggle - (change) event will be fired even when the slider has not change.
This bug occurs intermittently. When toggling the slider quickly, the slider might not toggle however the (change) event will still fire.

https://plnkr.co/edit/zMzNTiSz6OmxbrJZSUnP?p=preview
slide-toggle_change_event

The event should not fire when the value is not changed.

The event will fire even when the value is not changed.

1) Using the [Plunker](https://plnkr.co/edit/zMzNTiSz6OmxbrJZSUnP?p=preview) provided toggle the slider quickly. 
2) The last 10 event calls will be displayed
3) Occasionally, the slider will not toggle however the (change) event will still fire e.g. (true true) (false false)

The event should not fire when the value is not changed.

Tested on
Angular: 2.4.6
Material: 2.0.0-beta.2
OS: Windows
Browser: Chrome

NA

DevVersion:
Currently the slide-toggle will always emit a change event on `dragend`.

This is not always correct, because the `checked` value only changes if 50% of the distance have been dragged.

Fixes #3526

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -563,6 +563,22 @@ describe('MdSlideToggle', () => {
       expect(testComponent.lastEvent.checked).toBe(true);
     }));
 
+    it('should not emit a change event when the value did not change', fakeAsync(() => {
+      expect(slideToggle.checked).toBe(false);
+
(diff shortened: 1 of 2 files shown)


rhino5oh:
This is still an issue for me. Using material 6.0.0-beta.2

PR #3556, opened 2017-03-10
feat(datepicker): min & max dates + dateFilter

mmalerba:


src/lib/datepicker/year-view.ts
```diff
@@ -95,6 +95,24 @@ export class MdYearView implements AfterContentInit {
...
+  /** Whether the given month is enabled. */
+  private _isMonthEnabled(month: number) {
+    if (!this.dateFilter) {
+      return true;
+    }
+
+    let enabled = false;
+    for (let date = new SimpleDate(this.date.year, month, 1); date.month === month;
+         date = date.add({days: 1})) {
+      enabled = enabled || this.dateFilter(date);
```
kara: Do you need the `enabled ||` check? Looks like if it's enabled, it will break anyway.

mmalerba: derp ><

PR #3560, opened 2017-03-10
feat(input): add directive for displaying error messages

crisbeto:
Adds the `md-error` directive that can be utilised to display validation errors to the user. Example:

```html
<md-input-container>
  <input mdInput placeholder="email" [formControl]="emailFormControl">
  <md-error *ngIf="emailFormControl.errors.required">This field is required</md-error>
  <md-error *ngIf="emailFormControl.errors.pattern">
    Please enter a valid email address
  </md-error>
</md-input-container>
```

The `md-input-container` behavior is as follows:
* If there is an error and the user interacted with the input, the errors will be shown.
* If there is an error and the user submitted the a form that wraps the input, the errors will be shown.
* If there are errors to be shown on an input container that has `md-hint`-s, the hints will be hidden.
* If an input with hints and errors becomes valid, the hint won't be displayed anymore.

**Note:** At the moment, all hints will be hidden when an error is shown. This might not be intended for some cases (e.g. with a character counter). It might make sense to only hide the one in the `start` slot, but this could be addressed separately.

### DIFF
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -64,8 +64,10 @@
 
       <md-input-container>
         <input mdInput placeholder="email" [formControl]="emailFormControl">
-        <md-error *ngIf="emailFormControl.errors['required']">This field is required</md-error>
-        <md-error *ngIf="emailFormControl.errors['pattern']">
+        <md-error *ngIf="emailFormControl.errors ? emailFormControl.errors['required'] : false">
+          This field is required
+        </md-error>
+        <md-error *ngIf="emailFormControl.errors ? emailFormControl.errors['pattern'] : false">
           Please enter a valid email address
         </md-error>
       </md-input-container>
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -64,10 +64,10 @@
 
       <md-input-container>
         <input mdInput placeholder="email" [formControl]="emailFormControl">
-        <md-error *ngIf="emailFormControl.errors ? emailFormControl.errors['required'] : false">
+        <md-error *ngIf="emailFormControl.hasError('required')">
           This field is required
         </md-error>
-        <md-error *ngIf="emailFormControl.errors ? emailFormControl.errors['pattern'] : false">
+        <md-error *ngIf="emailFormControl.hasError('pattern')">
           Please enter a valid email address
         </md-error>
       </md-input-container>
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -64,8 +64,8 @@
 
       <md-input-container>
         <input mdInput placeholder="email" [formControl]="emailFormControl">
-        <md-error *ngIf="emailFormControl.errors.required">This field is required</md-error>
-        <md-error *ngIf="emailFormControl.errors.pattern">
+        <md-error *ngIf="emailFormControl.errors['required']">This field is required</md-error>
+        <md-error *ngIf="emailFormControl.errors['pattern']">
           Please enter a valid email address
         </md-error>
       </md-input-container>
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -52,6 +52,49 @@
 </md-card>
 
 <md-card class="demo-card demo-basic">
+  <md-toolbar color="primary">Error messages</md-toolbar>
+  <md-card-content>
+    <h4>Regular</h4>
+
+    <p>
+      <md-input-container>
+        <input mdInput placeholder="example" [(ngModel)]="errorMessageExample1" required>
+        <md-error>This field is required</md-error>
+      </md-input-container>
+
+      <md-input-container>
+        <input mdInput placeholder="email" [formControl]="emailFormControl">
+        <md-error *ngIf="emailFormControl.errors.required">This field is required</md-error>
+        <md-error *ngIf="emailFormControl.errors.pattern">
+          Please enter a valid email address
+        </md-error>
+      </md-input-container>
+    </p>
+
+    <h4>With hint</h4>
+
+    <md-input-container>
+      <input mdInput placeholder="example" [(ngModel)]="errorMessageExample2" required>
+      <md-error>This field is required</md-error>
+      <md-hint>Please type something here</md-hint>
+    </md-input-container>
+
+
+    <form novalidate>
+      <h4>Inside a form</h4>
+
+      <md-input-container>
+        <input mdInput name="example" placeholder="example"
+          [(ngModel)]="errorMessageExample3" required>
+        <md-error>This field is required</md-error>
+      </md-input-container>
+
+      <button color="primary" md-raised-button>Submit</button>
+    </form>
+  </md-card-content>
+</md-card>
+
+<md-card class="demo-card demo-basic">
   <md-toolbar color="primary">Prefix + Suffix</md-toolbar>
   <md-card-content>
     <h4>Text</h4>
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -64,6 +64,9 @@
     }
   }
 
+  // Styling for the error state of the input container. Note that while the same can be
+  // achieved with the ng-* classes, we use this approach in order to ensure that the same
+  // logic is used to style the error state and to show the error messages.
   .mat-input-invalid {
     .mat-input-placeholder,
     .mat-placeholder-required {
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -8,12 +8,12 @@
   $warn: map-get($theme, warn);
   $background: map-get($theme, background);
   $foreground: map-get($theme, foreground);
-  
+
   // Placeholder colors. Required is used for the `*` star shown in the placeholder.
   $input-placeholder-color: mat-color($foreground, hint-text);
   $input-floating-placeholder-color: mat-color($primary);
   $input-required-placeholder-color: mat-color($accent);
-  
+
   // Underline colors.
   $input-underline-color: mat-color($foreground, divider);
   $input-underline-color-accent: mat-color($accent);
@@ -64,7 +64,7 @@
     }
   }
 
-  .mat-input-container.ng-invalid.ng-touched:not(.mat-focused) {
+  .mat-input-invalid {
     .mat-input-placeholder,
     .mat-placeholder-required {
       color: $input-underline-color-warn;
@@ -73,5 +73,13 @@
     .mat-input-underline {
       border-color: $input-underline-color-warn;
     }
+
+    .mat-input-ripple {
+      background-color: $input-underline-color-warn;
+    }
+  }
+
+  .mat-input-error {
+    color: $input-underline-color-warn;
   }
 }
--- a/src/lib/input/input-container.html
+++ b/src/lib/input/input-container.html
@@ -37,12 +37,12 @@
   </div>
 
   <div class="mat-input-hint-wrapper" [ngSwitch]="_getDisplayedMessages()">
-    <div *ngSwitchCase="'error'" [@transitionMessages]="_messageAnimationState">
+    <div *ngSwitchCase="'error'" [@transitionMessages]="_subscriptAnimationState">
       <ng-content select="md-error, mat-error"></ng-content>
     </div>
 
-    <div *ngSwitchCase="'hint'" [@transitionMessages]="_messageAnimationState">
-      <div *ngIf="hintLabel != ''" [attr.id]="_hintLabelId" class="mat-hint">{{hintLabel}}</div>
+    <div *ngSwitchCase="'hint'" [@transitionMessages]="_subscriptAnimationState">
+      <div *ngIf="hintLabel" [id]="_hintLabelId" class="mat-hint">{{hintLabel}}</div>
       <ng-content select="md-hint, mat-hint"></ng-content>
     </div>
   </div>
--- a/src/lib/input/input-container.html
+++ b/src/lib/input/input-container.html
@@ -36,12 +36,13 @@
           [class.mat-warn]="dividerColor == 'warn'"></span>
   </div>
 
-  <div class="mat-input-hint-wrapper" [ngSwitch]="_getDisplayedMessages()">
+  <div class="mat-input-subscript-wrapper" [ngSwitch]="_getDisplayedMessages()">
     <div *ngSwitchCase="'error'" [@transitionMessages]="_subscriptAnimationState">
       <ng-content select="md-error, mat-error"></ng-content>
     </div>
 
-    <div *ngSwitchCase="'hint'" [@transitionMessages]="_subscriptAnimationState">
+    <div class="mat-input-hint-wrapper" *ngSwitchCase="'hint'"
+      [@transitionMessages]="_subscriptAnimationState">
       <div *ngIf="hintLabel" [id]="_hintLabelId" class="mat-hint">{{hintLabel}}</div>
       <ng-content select="md-hint, mat-hint"></ng-content>
     </div>
--- a/src/lib/input/input-container.html
+++ b/src/lib/input/input-container.html
@@ -41,7 +41,7 @@
       <ng-content select="md-error, mat-error"></ng-content>
     </div>
 
-    <div *ngSwitchCase="'hint'" class="mat-input-hint-clearfix" [@transitionMessages]="_messageAnimationState">
+    <div *ngSwitchCase="'hint'" [@transitionMessages]="_messageAnimationState">
       <div *ngIf="hintLabel != ''" [attr.id]="_hintLabelId" class="mat-hint">{{hintLabel}}</div>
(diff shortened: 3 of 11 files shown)


src/lib/input/input-container.scss
```diff
@@ -211,29 +212,53 @@ $mat-input-underline-disabled-background-image:
...
 
     &.mat-right {
-      right: auto;
-      left: 0;
+      float: left;
     }
   }
 }
 
+// Clears the floats on the hints. Necessary for the `transform` animation to work.
```
mmalerba: what exactly is this supposed to do? I don't see any difference if I just remove it.

src/demo-app/input/input-demo.html
```diff
@@ -49,6 +49,49 @@
...
+    <h4>Regular</h4>
+
+    <p>
+      <md-input-container>
+        <input mdInput placeholder="example" [(ngModel)]="errorMessageExample1" required>
+        <md-error>This field is required</md-error>
+      </md-input-container>
+
+      <md-input-container>
+        <input mdInput placeholder="email" [formControl]="emailFormControl">
```
mmalerba: the error animation is kind of strange in this case, the error message seems to appear in the middle of the input and then move down (I think because the message wraps on to 2 lines).

src/demo-app/input/input-demo.html
```diff
@@ -49,6 +49,49 @@
...
+    <form novalidate>
+      <h4>Inside a form</h4>
+
+      <md-input-container>
+        <input mdInput name="example" placeholder="example"
+          [(ngModel)]="errorMessageExample3" required>
+        <md-error>This field is required</md-error>
+      </md-input-container>
+
+      <button color="primary" md-raised-button>Submit</button>
```
mmalerba: is this button supposed to do something?

src/lib/input/input-container.ts
```diff
@@ -302,6 +336,12 @@ export class MdInputContainer implements AfterContentInit {
     this._mdInputChild._placeholderChange.subscribe(() => this._validatePlaceholders());
   }
 
+  ngAfterViewInit() {
+    // Avoid animations on load.
+    this._messageAnimationState = 'enter';
+    this._changeDetectorRef.detectChanges();
```
mmalerba: @andrewseguin do you know if there's a better way to prevent animations on load?

src/lib/input/input-container.ts
```diff
@@ -314,6 +354,30 @@ export class MdInputContainer implements AfterContentInit {
...
+    const control = this._mdInputChild._ngControl;
+    const isInvalid = control ? control.invalid : false;
+    const isTouched = control ? control.touched : false;
+    const isSubmitted = this._parentForm ? this._parentForm.submitted : false;
+
+    return isInvalid && (isTouched || isSubmitted);
+  }
+
+  /** Determines whether to display hints, errors or no messages at all. */
+  _getDisplayedMessages(): 'error'|'hint'|'none' {
```
mmalerba: nit: spaces around '|'

src/lib/input/input-container.ts
```diff
@@ -314,6 +354,30 @@ export class MdInputContainer implements AfterContentInit {
...
+    if (this._errorChildren.length > 0) {
+      if (this._isErrorState()) {
+        return 'error';
+      } else if (this._mdInputChild._ngControl) {
+        let control = this._mdInputChild._ngControl;
+        return (control.valid && control.touched) ? 'none' : 'hint';
```
mmalerba: why 'none' if its valid and touched?

src/demo-app/input/input-demo.html
crisbeto: Yes, it's used to test out that form submission shows the error messages.

src/lib/input/input-container.scss
crisbeto: It breaks the sliding down transition, IIRC.

src/lib/input/input-container.ts
crisbeto: In that case we shouldn't show any messages at all (neither errors nor hints). It's how M1 handles it.

src/lib/input/_input-theme.scss
```diff
@@ -64,7 +64,7 @@
     }
   }
 
-  .mat-input-container.ng-invalid.ng-touched:not(.mat-focused) {
+  .mat-input-invalid {
```
crisbeto: Mostly because I wanted to re-use the same logic that determines whether the messages are shown. These classes get forwarded "manually" from the input element anyway so there shouldn't be much of a performance difference.

src/demo-app/input/input-demo.html
crisbeto: I'll take another look. It probably needs a `transform-origin`.

src/lib/input/input-container.ts
mmalerba: I don't see anything about that in the spec and it seems weird to me. @jelbourn thoughts?

src/lib/input/input-container.scss
mmalerba: It doesn't seem to be necessary, I can remove it and everything works fine

crisbeto: Ah, now I remember why it's there: when I was initially working on this, I wasn't hiding the hints after the user enters a valid value, but rather swapping between errors and hints, based on the validity. The hint slide down animation doesn't work without it, but it doesn't matter with the current setup.

crisbeto:
Addressed some of the feedback @mmalerba.

mmalerba:
Looks pretty good now, one more thing I noticed: when it shifts between error and no error state, the placeholder seems to jiggle around. repro steps:
1. touch one of the inputs
2. type a letter, then delete it - repeat and watch label

crisbeto:
I believe that one is the same that got fixed by https://github.com/angular/material2/pull/3568. I haven't had to rebase this in a while.

src/lib/input/input-container.ts
```diff
@@ -234,10 +249,20 @@ export class MdInputDirective {
   selector: 'md-input-container, mat-input-container',
   templateUrl: 'input-container.html',
   styleUrls: ['input-container.css'],
+  animations: [
+    trigger('transitionMessages', [
+      state('enter', style({ opacity: 1, transform: 'translateY(0%)' })),
+      transition('void => enter', [
+        style({ opacity: 0, transform: 'translateY(-100%)' }),
+        animate('300ms')
```
andrewseguin: Can you add in an easing function to avoid a harsh linear translation?
https://material.io/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves

src/lib/input/input-container.ts
andrewseguin: This does seem odd, but perhaps I'm missing something about why its needed.

Does this animation not work if you initially set the messageAnimationState to enter?

crisbeto: The point is that the animation shouldn't run on the first load, which is why it gets added after the view init. It also needs to trigger change detection, in order to avoid "<thing> was changed after it was checked" errors.

src/lib/input/_input-theme.scss
jelbourn: Can you add that rationale in a comment here?

src/lib/input/input-container.html
```diff
@@ -36,6 +36,14 @@
...
 
-  <div *ngIf="hintLabel != ''" [attr.id]="_hintLabelId" class="mat-hint">{{hintLabel}}</div>
-  <ng-content select="md-hint, mat-hint"></ng-content>
+  <div class="mat-input-hint-wrapper" [ngSwitch]="_getDisplayedMessages()">
+    <div *ngSwitchCase="'error'" [@transitionMessages]="_messageAnimationState">
+      <ng-content select="md-error, mat-error"></ng-content>
+    </div>
+
+    <div *ngSwitchCase="'hint'" [@transitionMessages]="_messageAnimationState">
+      <div *ngIf="hintLabel != ''" [attr.id]="_hintLabelId" class="mat-hint">{{hintLabel}}</div>
```
jelbourn: Why not `*ngIf="hintLabel"` ?

jelbourn: Just `[id]` should work without needing `attr.`

src/lib/input/input-container.scss
```diff
@@ -211,29 +212,41 @@ $mat-input-underline-disabled-background-image:
...
+
+// The hint is shown below the underline. There can be
+// more than one; one at the start and one at the end.
+.mat-hint {
+  display: block;
+  float: left;
 
   &.mat-right {
-    right: 0;
+    float: right;
```
jelbourn: Why use float instead of using flex to position it appropriately? Would be good to have a comment

src/lib/input/input-container.ts
```diff
@@ -71,6 +78,14 @@ export class MdHint {
   @Input() id: string = `md-input-hint-${nextUniqueId++}`;
 }
 
+/** Directive, used to display a single error message under the input. */
+@Directive({
+  selector: 'md-error, mat-error',
```
jelbourn: Add these to `compatibility.ts`? It would be nice if we had a check for this, I bet we could use the metadata.json to get all of the selectors already... (cc @DevVersion)

src/lib/input/input-container.ts
```diff
@@ -263,6 +288,9 @@ export class MdInputContainer implements AfterContentInit {
   /** Whether the placeholder can float or not. */
   get _canPlaceholderFloat() { return this._floatPlaceholder !== 'never'; }
 
+  /** State of the md-hint and md-error animations. */
+  _messageAnimationState: string = '';
```
jelbourn: How about something like `subscriptAnimationState` to specifically reference that it's underneath the input?

src/lib/input/input-container.ts
DevVersion: Won't a `markForCheck` call work as well?

jelbourn: @matsko can provide the definitive word here. I am pretty sure we shouldn't be calling `detectChanges` after view init just to get a certain animation behavior.

src/lib/input/input-container.ts
```diff
@@ -314,6 +354,30 @@ export class MdInputContainer implements AfterContentInit {
   /** Focuses the underlying input. */
   _focusInput() { this._mdInputChild.focus(); }
 
+  /** Whether the input container is in an error state. */
+  _isErrorState(): boolean {
+    const control = this._mdInputChild._ngControl;
+    const isInvalid = control ? control.invalid : false;
+    const isTouched = control ? control.touched : false;
+    const isSubmitted = this._parentForm ? this._parentForm.submitted : false;
```
jelbourn: optional: rather than using a ternary, these could just be conditional expressions:

```ts
const isInvalid = control && control.invalid;
const isTouched = control && control.touched;
const isSubmitted = this._parentForm && this._parentForm.submitted;

return !!(isInvalid && (isTouched || isSubmitted));
```
(results in one boolean coercion instead of three)

src/lib/input/input-container.ts
jelbourn: I can imagine cases where people would want the hint to continue showing the hint even when the input is touched. I can post the question to Material UX, but unsure if I'll get an answer.

src/lib/input/input-container.ts
DevVersion: Can you create a quick issue for this? So we don't have to write on that PR.

src/lib/input/input-container.ts
```diff
@@ -287,8 +315,14 @@ export class MdInputContainer implements AfterContentInit {
 
   @ContentChild(MdPlaceholder) _placeholderChild: MdPlaceholder;
 
+  @ContentChildren(MdErrorDirective) _errorChildren: QueryList<MdErrorDirective>;
+
   @ContentChildren(MdHint) _hintChildren: QueryList<MdHint>;
 
+  constructor(
+    private _changeDetectorRef: ChangeDetectorRef,
+    @Optional() private _parentForm: NgForm) { }
```
kara: If you inject `NgForm`, this will only work for template-driven forms (`formControl` only happens to work). Any form tag that has a `[formGroup]` directive on it becomes a `FormGroupDirective` instead, so this won't pick it up.

src/demo-app/input/input-demo.html
```diff
@@ -49,6 +49,51 @@
...
+
+    <p>
+      <md-input-container>
+        <input mdInput placeholder="example" [(ngModel)]="errorMessageExample1" required>
+        <md-error>This field is required</md-error>
+      </md-input-container>
+
+      <md-input-container>
+        <input mdInput placeholder="email" [formControl]="emailFormControl">
+        <md-error *ngIf="emailFormControl.errors ? emailFormControl.errors['required'] : false">
```
kara:
```
*ngIf="emailFormControl.hasError('required')"
```

Same for the pattern one below.

src/lib/input/input-container.spec.ts
```diff
@@ -550,6 +552,124 @@ describe('MdInputContainer', function () {
...
+
+    beforeEach(() => {
+      fixture = TestBed.createComponent(MdInputContainerWithErrorMessages);
+      fixture.detectChanges();
+      testComponent = fixture.componentInstance;
+      containerEl = fixture.debugElement.query(By.css('md-input-container')).nativeElement;
+    });
+
+    it('should not show any errors if the user has not interacted', () => {
+      expect(testComponent.formControl.pristine).toBe(true, 'Expected untouched form control');
```
kara: Do you mean pristine or untouched here? Those are separate FormControl states. It seems like you want `untouched`.

src/lib/input/input-container.spec.ts
```diff
@@ -550,6 +552,124 @@ describe('MdInputContainer', function () {
...
+
+      fixture.whenStable().then(() => {
+        expect(containerEl.classList)
+            .toContain('mat-input-invalid', 'Expected container to have the invalid CSS class.');
+        expect(containerEl.querySelectorAll('md-error').length)
+            .toBe(1, 'Expected one error message to have been rendered.');
+      });
+    }));
+
+    it('should display an error message when the parent form is submitted', async(() => {
```
kara: Can you add a test with `formGroup`?

src/lib/input/input-container.html
crisbeto: I didn't write this, I just moved it around. Will change it.

crisbeto:
Thanks for the feedback, everyone. I've addressed most of it. What's left seems to be:
* Figuring out if there's a better way to disable the initial animation.
* Checking the UX for hiding the hints.

src/lib/input/input-container.spec.ts
```diff
@@ -775,6 +932,41 @@ class MdTextareaWithBindings {
...
+  @ViewChild('form') form: NgForm;
+  formControl = new FormControl('', Validators.required);
+  renderError = true;
+}
+
+@Component({
+  template: `
+    <form [formGroup]="formGroup" (ngSubmit)="onSubmit()" novalidate>
+      <md-input-container>
+        <input mdInput [formControl]="formControl">
```
kara: I think you mean `formControlName` here? `formControl` is for individual form controls, not controls within groups.

src/lib/input/input-container.ts
natsid: +1 to keeping the hint visible when valid and touched.

jelbourn: @crisbeto let's move ahead with showing both for now; users always have the ability to `ngIf` the hint if they want

jefersonestevo:
The md-error component will be applied only to the md-input-container? If so, there are any thoughts on how could we show error messages from other components, like md-select. Will md-input-container be extended to hold these components?

src/lib/input/input-container.ts
crisbeto: Done.

jelbourn:
@crisbeto any outstanding changes you need to make on this before it's ready for every to take another pass?

crisbeto:
I don't think so. Functionality-wise everything is there and we can always improve upon initially disabling the animations in a follow-up PR.

jelbourn:
@kara @mmalerba please take a look

src/lib/input/input-container.scss
```diff
@@ -211,29 +212,57 @@ $mat-input-underline-disabled-background-image:
...
 
     &.mat-right {
-      right: auto;
-      left: 0;
+      float: left;
     }
   }
 }
 
+// Single errror message displayed beneath the input.
```
kara: Typo: `errror` -> `error`

WillGeller:
Has the <md-error> element been added to beta.3?

crisbeto:
It has @WillGeller.

constantinlucian:
Hi! I use angular material 2.0.0-beta.3 and reactive forms. I want to show error messages for md-input. Ex:<md-input-container>
            <input formControlName="firstName" mdInput placeholder="First Name">
            <md-error *ngIf="firstName.errors.required">
                This field is required
            </md-error>
        </md-input-container>


ERROR TypeError: Cannot read property 'errors' of undefined. Can you help me? Thanks.

mmalerba:
Please keep GitHub issues for bug reports / feature requests. Better avenues for troubleshooting / questions are stack overflow, gitter, mailing list, etc.

PR #3563, opened 2017-03-11
build: replace gulp-autoprefixer

### LINKED ISSUE #3536: Stop using autoprefixer
Due to complications with the internal Google build system, it's not feasible for us to use autoprefixer for internal Google applications that are consuming Angular Material. The most straightforward thing for us to do, then, is to stop using it and capture any necessary CSS vendor prefixes in the source. The steps necessary:

* Determine which CSS properties need to be prefixed and document this
* Create sass mixins for those properties
* [Bonus] Create a  stylelint plugin that prevents these properties from being added without a prefix.

crisbeto:
* Removes `gulp-autoprefixer` in order to avoid issues with Google's internal build system.
* Switches to adding vendor prefixes via SCSS mixins.
* Adds a custom Stylelint plugin that utilizies Autoprefixer to find unprefixed properties, values, @rules etc.

Fixes #3536.


crisbeto:

PR #3567, opened 2017-03-12
fix(select): wrong item order in label in rtl

crisbeto:
Fixes selected items not having the proper order in the label when `md-select` is in `multiple` mode and in RTL.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -413,9 +413,18 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
 
   /** The value displayed in the trigger. */
   get triggerValue(): string {
-    return this.multiple ?
-      this._selectionModel.selected.map(option => option.viewValue).join(', ') :
-      this._selectionModel.selected[0].viewValue;
+    if (this._multiple) {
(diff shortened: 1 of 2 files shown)


src/lib/select/select.ts
```diff
@@ -413,9 +413,17 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
 
   /** The value displayed in the trigger. */
   get triggerValue(): string {
-    return this.multiple ?
-      this._selectionModel.selected.map(option => option.viewValue).join(', ') :
-      this._selectionModel.selected[0].viewValue;
+    if (this._multiple) {
+      let array = this._selectionModel.selected.map(option => option.viewValue);
```
jelbourn: `array` -> `selectedOptions`

src/lib/select/select.ts
```diff
@@ -413,9 +413,17 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
...
+
+      if (this._isRtl()) {
+        array.reverse();
+      }
+
+      return array.join(', ');
```
jelbourn: Just noticed: I think this delimiter should be configurable, as a comma may not be the right character for all locales. Add a TODO?

crisbeto:

Issue #2465, opened 2016-12-29
md-select placeholder blends with other text.


davidwilcox:
#### Bug, feature request, or proposal:
md-select placeholder text is placed under text that occurs immediately prior.

#### What is the expected behavior?
Padding is built into the md-select so that the placeholder text is not placed over or under any text that comes prior.

#### What is the current behavior?
The placeholder attribute goes over any text that is prior to it.

#### What are the steps to reproduce?
Use an md-select element with a placeholder attribute.

http://plnkr.co/edit/R906XtG5GXTA4SNtQUZy?p=preview
Select "Salt Lake City" from the input element.

#### What is the use-case or motivation for changing an existing behavior?
It looks kind of bad the way it is where you have two things of text on top of each other. It doesn't make sense from a developer's standpoint for this to happen. It's onerous to build into each md-select padding necessary in order to make it look good.

#### Which versions of Angular, Material, OS, browsers are affected?
@angular/core@^2.4.1
@angular/material@^2.0.0-beta.1
Ubuntu 16.04
Google Chrome browser.

#### Is there anything else we should know?
Nope! Great job! Thanks for your help!

peluprvi:
Select with placeholder has the same height as a text field ([Material guidelines](https://material.io/guidelines/components/text-fields.html#text-fields-single-line-text-field)):

![image](https://cloud.githubusercontent.com/assets/1561360/23839063/1ab989ce-077a-11e7-8905-0d2c80e2b811.png)


Material text field specs:

![image](https://cloud.githubusercontent.com/assets/1561360/23838978/7504f716-0779-11e7-820b-12ff41082754.png)

jelbourn:
Tracking via #2124

Issue #2168, opened 2016-12-11
md-select placeholder overlaps with md-toolbar


belev:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
md-select placeholder not to overlap with md-toolbar

#### What is the current behavior?
md-select placeholder overlaps with md-toolbar

#### What are the steps to reproduce?
Click to select item, the placeholder overlaps. After selecting an item the placeholder is not visible too.

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: http://plnkr.co/edit/uJR8DXST8lTBOptjgkN4?p=preview

#### Which versions of Angular, Material, OS, browsers are affected?
Material 2.0.0-alpha.11-2 (tested on master too), Chrome 55

#### Is there anything else we should know?
Not sure if that's actually a bug or the user should handle the margin/padding after the toolbar.

pj-agfa:
I noticed that the positioning and transform of the md-input placeholder is different compared to the md-select placeholder. This is probably related to this issue.

kara:
Yeah, related to https://github.com/angular/material2/issues/2124 and https://github.com/angular/material2/issues/2127

SalmaKolkari:
margin: 1em 0;
padding-bottom: 6px;

these css properties are missing in select.css file.

Please add : 

md-select {
  display: inline-block;
  outline: 0;
  margin: 1em 0;
  padding-bottom: 6px;
}

in select.css file

and add same css properties in

styles: ["md-select{display:inline-block;outline:0;margin: 1em 0;padding-bottom: 6px;} 

in 

Component({selector: 'md-select, mat-select', 

in 

select.js file in select folder of @angular/material directory.

jelbourn:
Tracking via #2124

Issue #2124, opened 2016-12-09
Input components (md-input, md-select, md-checkbox etc) aren't aligned in a flex container...


jmcgoldrick:
... without setting `align-items: center`

**Updated Plunk: http://plnkr.co/edit/TgFy04UErh9Xhr0A4XzK?p=preview**

`md-input` behaves differently to `md-select` (and at least `md-checkbox`, possibly all the other "input" components) when the elements are within a "flex container"

See this plunk for an example: http://plnkr.co/edit/aLXsp8MlV2AoOkC9aEFP?p=preview

This issue might be resolved with the in-progress refactor of input?

dreweird:
`md-select` already available??

kara:
Thanks for reporting!  The various form components should definitely align. We'll need to figure out which baseline to standardize under. cc: @jelbourn.

DSoko2:
I recognized that also the spacing between the selected/input text and the label is different for md-select and md-input. The screenshot below shows a md-input and a md-select with the spacing im referring to. I would really appreciate, if they would use the same spacing to improve homogeneity of the forms. Since I think, this aspect is related to the topic of this issue, I didn't open a new one.

![spacing](https://cloud.githubusercontent.com/assets/11334016/21718567/955a8560-d418-11e6-919a-017fba12e43c.png)

On https://material.io/guidelines/components/text-fields.html#text-fields-labels there are two definitions for that kind of spacing under Labels -> Specs. Actually, it seems to be that md-select implements the left one with 8dp space between label and text and md-input implements the right one with 4dp spacing.

no-more:
Hi, 

This is a bit off topic, but why not using md-input-container for md-select ?
It would be great to have a single entry point.
Maybe this is stupid question and not possible at all due to constraints, but to me this looks more logic.

This could also be used to group radio button or checkbox in order to define a placeholder which looks exactly like md-input placeholder.

Thanks

sumpton:
Thank you for all this wonderful code. I am busy refactoring a ton of Bootstrap.

On this mention of a common wrapper, I have been wishing for md-prefix and md-hint support across all of the form controls. For inputs I have enjoyed using md-prefix with md-icon together -- for example a search icon.

tmburnell:
This might be related: besides the header not having the same height from the text ... the initial overall height is not the same.  
The standard input looks like it has the height of 50px and the select box is 30px.  
so after you fill in something there is no space between the top like the standard input
[...]

PR #3572, opened 2017-03-13
fix(select): Select placeholder height

peluprvi:
Related to #2124 #2168 #2465


JacobBrandt:
Please don't merge this in.  It only makes #2124 worse.

peluprvi:
This PR doesn't fix the problem and can be closed.

PR #3573, opened 2017-03-13
fix(autocomplete): update overlay ref width on menu trigger

emoralesb05:
When changing the width of the `host` (e.g. using flex box) after the first time the `autocomplete` menu is opened, the `div.cdk-overlay-pane` is not updated and keeps the `width` that the `host` had when it was initially opened.

https://github.com/angular/material2/blob/master/src/lib/autocomplete/autocomplete-trigger.ts#L123

Here is a plnkr to demonstrate the issue. http://plnkr.co/edit/Evz0KLvvkFcmerIoca42?p=preview

### Steps to reproduce

- First focus the `md-input` component to initialize the `md-autocomplete`s `cdk-overlay-pane`.
- Change the `width` of the `md-input` component by entering a diff value in the `width` input box under it.
- Focus the `md-input` component again, and notice the `width` of the `autocomplete`'s `cdk-overlay-pane` remained the same as when it was initialized.

### Proposed fix in PR

We just need to set the new `width` every time the `autocomplete`'s panel is opened (`openPanel()`) and leverage that the `updateSize()` method is called when the `portal` is reattached.

https://github.com/angular/material2/blob/master/src/lib/autocomplete/autocomplete-trigger.ts#L127

https://github.com/angular/material2/blob/master/src/lib/core/overlay/overlay-ref.ts#L40

```typescript
if (!this._overlayRef) {
     this._createOverlay();
 } else {
   /** Gets width of host incase this has changed, and updated the Overlay reference */
   this._overlayRef.getState().width = this._getHostWidth();
 }
```

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -122,7 +122,7 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
     if (!this._overlayRef) {
       this._createOverlay();
     } else {
-      /** Gets width of host incase this has changed, and updated the Overlay reference */
+      /** Update the panel width, in case the host width has changed */
       this._overlayRef.getState().width = this._getHostWidth();
     }
 
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -121,6 +121,9 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   openPanel(): void {
     if (!this._overlayRef) {
       this._createOverlay();
+    } else {
+      /** Gets width of host incase this has changed, and updated the Overlay reference */
+      this._overlayRef.getState().width = this._getHostWidth();
     }
 
     if (!this._overlayRef.hasAttached()) {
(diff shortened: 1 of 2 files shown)


emoralesb05:
Sorry about that 😄 . Added pane width unit test when triggered.

emoralesb05:
hmm.. it looks like the width is not really exact depending on the browser. 

I need to see how i can fix the unit tests for that since they pass correctly when i use `gulp test`

emoralesb05:
Ok, had to parse and round the width in the unit test to be able to verify it since `firefox`, `edge` return a decimal value near the provided `width`.

e.g. `300px` would return `299.8px`

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -121,6 +121,9 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   openPanel(): void {
     if (!this._overlayRef) {
       this._createOverlay();
+    } else {
+      /** Gets width of host incase this has changed, and updated the Overlay reference */
```
crisbeto: This isn't very understandable. Something along the lines of "Update the panel width, in case the host has changed" might be better.

emoralesb05: Ok cool! will change it

emoralesb05:
Changed the comment @crisbeto. So this is good to go 😄

crisbeto:
CC @kara

PR #3578, opened 2017-03-13
feat(tooltip): allow tooltip be disabled

andrewseguin:
It is not uncommon to disable tooltips, e.g. removing tooltip on disabled buttons. This adds an API `mdTooltipDisabled` to prevent the tooltip from displaying.

### DIFF
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -112,6 +112,23 @@ describe('MdTooltip', () => {
       expect(overlayContainerElement.textContent).toContain(initialTooltipMessage);
     }));
 
+    fit('should not show if disabled', fakeAsync(() => {
+      // Test that disabling the tooltip will not set the tooltip visible
+      tooltipDirective.disabled = true;
+      tooltipDirective.show();
+      fixture.detectChanges();
+      tick(0);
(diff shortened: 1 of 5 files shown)


src/lib/tooltip/tooltip.spec.ts
```diff
@@ -112,6 +112,23 @@ describe('MdTooltip', () => {
       expect(overlayContainerElement.textContent).toContain(initialTooltipMessage);
     }));
 
+    fit('should not show if disabled', fakeAsync(() => {
```
tinayuangao: it()

src/lib/tooltip/tooltip.ts
```diff
@@ -78,6 +80,11 @@ export class MdTooltip implements OnInit, OnDestroy {
     }
   }
 
+  /** Disables the display of the tooltip. */
+  @Input('mdTooltipDisabled')
+  get disabled(): boolean { return this._disabled; }
+  set disabled(value) { this._disabled = coerceBooleanProperty(value); }
```
jelbourn: Should setting `disabled` hide a tooltip that's already showing?

andrewseguin: I think thats a fair thing to assume, adding that in

Issue #3073, opened 2017-02-12
(md-select) with placeholder changes size when selected


benb7760:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
An md-select with a selection should be the same size as with no selection

#### What is the current behavior?
This is true for an md-select with no placeholder, but when I provide a placeholder, the control sizes differently

As a side note - can anybody comment on why the second controls placeholder overlaps with the control above, when selected? Is this a bug which should be filed separately?

#### What are the steps to reproduce?
Select a value in both of the lists, and you'll see an offset

http://plnkr.co/edit/KdHQWMFXj2hEFWWFRJyq?p=preview

#### Which versions of Angular, Material, OS, browsers are affected?
Reproduced in Chrome and Safari latest (not tried other browsers)

dahaupt:
@benb7760 To answer your second question: This issue is already known and will be fixed soon with #2309.

dahaupt:
This is because of this method in `select.ts`:
```typescript
private _setValueWidth() {
    this._selectedValueWidth =  this._triggerWidth - 13;
  }
```

When removing this line it works fine, but I'm not sure about the side effects of that.

dahaupt:
To fix this temporary you can use:
```css
span.mat-select-placeholder.mat-floating-placeholder {
  width: auto !important;
}
```

jelbourn:
Tracking via #2124

PR #3602, opened 2017-03-14
bug(md-select): change the arrow position to absolute

### LINKED ISSUE #3112: bug(md-select) - not rendering correctly inside a flex layout (pure css) when having ngModel

jefersonestevo:
Fixes #3112

### DIFF
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -111,13 +111,21 @@ $mat-select-trigger-font-size: 16px !default;
 }
 
 .mat-select-arrow {
+  position: absolute;
+  right: 0;
   width: 0;
   height: 0;
   border-left: $mat-select-arrow-size solid transparent;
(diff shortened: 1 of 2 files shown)


dahaupt:
@jefersonestevo Does this also fix #3073?

PR #3618, opened 2017-03-15
docs(readme): Individual modules explaination

realappie:


guides/getting-started.md
```diff
@@ -23,6 +23,10 @@ import { MaterialModule } from '@angular/material';
 export class PizzaPartyAppModule { }
 ```
 
+You could also import the individual components you need. Which reduces the size of your bundle. For example, If you will only  need the buttons you would import the `MdButtonModule` instead of the `MaterialModule`
```
jelbourn: This is going to change in the next release a bit, so we'll hold off on updating the getting started guide until then.

realappie: Alright!

PR #3622, opened 2017-03-15
fix(tabs): re-align ink bar on direction change

### LINKED ISSUE #3615: position of MdInkBar in MdTab doesn't update when the direction change
Bug

update the position

no position update

1. run the devapp of this repo locally
2. open the tabs section
3. change to direction to `rtl`
4. the mdInkBar all stay put.

No plunkr need.
Before 05c865da154eeda0844d8ede79d60f40b07a554b, all are not ok.

better UE

Angular 2.4.9, Material 2.0.0-beta.2, all os, browser

crisbeto:
Re-aligns the ink bar if the user's layout direction changed.

Fixes #3615.

### DIFF
--- a/src/lib/tabs/tab-header.spec.ts
+++ b/src/lib/tabs/tab-header.spec.ts
@@ -11,10 +11,12 @@ import {RIGHT_ARROW, LEFT_ARROW, ENTER} from '../core/keyboard/keycodes';
 import {FakeViewportRuler} from '../core/overlay/position/fake-viewport-ruler';
 import {ViewportRuler} from '../core/overlay/position/viewport-ruler';
 import {dispatchKeyboardEvent} from '../core/testing/dispatch-events';
+import {Subject} from 'rxjs/Subject';
 
 
 describe('MdTabHeader', () => {
   let dir: LayoutDirection = 'ltr';
+  let dirChange = new Subject();
   let fixture: ComponentFixture<SimpleTabHeaderApp>;
   let appComponent: SimpleTabHeaderApp;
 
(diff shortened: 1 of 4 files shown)


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -40,10 +52,26 @@ export class MdTabNavBar {
...
+  ngOnDestroy() {
+    if (this._directionChange) {
+      this._directionChange.unsubscribe();
+      this._directionChange = null;
+    }
+  }
+
+  /** Aligns the ink bar to the active link. */
+  private _alignInkBar(): void {
+    this._ngZone.runOutsideAngular(() => {
```
jelbourn: Would it make sense to move the `runOutsideangular` and `requestAnimationFrame` part of this from both tab-group and tab-nav-bar into the ink-bar?

crisbeto:

src/lib/tabs/tab-header.spec.ts
```diff
@@ -29,7 +31,9 @@ describe('MdTabHeader', () => {
         SimpleTabHeaderApp,
       ],
       providers: [
-        {provide: Dir, useFactory: () => { return {value: dir}; }},
+        {provide: Dir, useFactory: () => {
+          return {value: dir,  dirChange: dirChange.asObservable()};
+        }},
```
jelbourn: Add unit test for aligning when the direction changes?

crisbeto: Done. I went with a spy, instead of testing the position itself, because knowing when the position is final is hard since we're running things outside of Angular and we're using CSS transitions to animate it. I had something working, but it was using `transitionend` which may end up making the test flaky.

PR #3623, opened 2017-03-15
feat(select): update the trigger width when select is opened

emoralesb05:
When changing the width of the `md-select` (e.g. using flex box) after its rendered, the `template[cdk-connected-overlay]` does not update its `minWidth` and keeps the initial `minWidth` every time its opened. (somewhat the same issue as https://github.com/angular/material2/pull/3573)

This issue partially happens because the `cdk-connected-overlay` sets the `width`, `minWidth`, etc etc only once and never updates it again.

https://github.com/angular/material2/blob/master/src/lib/core/overlay/overlay-directives.ts#L181

But also because the `_triggerWidth` property in `md-select` is never updated either.

Here is a plnkr to demonstrate the issue: http://plnkr.co/edit/ERg71p?p=preview

### Steps to reproduce

- Change the `width` of the `md-select` component by entering a diff value in the `width` input box under it. (both `single` and `multiple` `md-select`'s will change)
- Open either `md-select`, and notice the `width` of the `cdk-overlay-pane` remained the same as when it was initially rendered.

### Proposed fix in PR

We just need to set the new `_triggerWidth` every time the `md-select` is opened (`open()`) taking into account `dir`/`offsetX` and update the `overlayRef#OverlayState` in `ConnectedOverlayDirective` when the `template` is attached.

https://github.com/angular/material2/blob/master/src/lib/core/overlay/overlay-directives.ts#L188-L202

### DIFF
--- a/src/lib/core/overlay/overlay-directives.spec.ts
+++ b/src/lib/core/overlay/overlay-directives.spec.ts
@@ -107,6 +107,15 @@ describe('Overlay directives', () => {
 
       const pane = overlayContainerElement.children[0] as HTMLElement;
       expect(pane.style.width).toEqual('250px');
+
+      fixture.componentInstance.isOpen = false;
+      fixture.detectChanges();
+
+      fixture.componentInstance.width = 500;
+      fixture.componentInstance.isOpen = true;
+      fixture.detectChanges();
+
+      expect(pane.style.width).toEqual('500px');
     });
 
     it('should set the height', () => {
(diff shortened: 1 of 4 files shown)


emoralesb05:
@crisbeto not sure if this PR affects https://github.com/angular/material2/pull/3228 and https://github.com/angular/material2/pull/3247.. or the vision for how it should work

kara:
@emoralesb05 Is this still an issue with the current version? If so, would you mind rebasing? Sorry for the late review.

emoralesb05:
@kara yes its still an issue, i updated the plnkr with the latest version to try it.

Rebasing it so we can hopefully merge it soon 😄

kara:
OK, ping me when it's ready for review again :)

emoralesb05:
Im actually gonna create another PR since rebasing 5-6 months worth of code is taking its toll on me haha 😂 ..

kara:
Haha, fair enough. Ping me when your new PR is ready, and I'll review ASAP to avoid more conflicts.

emoralesb05:
@kara i opened https://github.com/angular/material2/pull/6489 so we can continue in that one 😄  

Closing this one

PR #3624, opened 2017-03-15
Support disabled property for md-list-item

### LINKED ISSUE #763: md-list-item disabled property
Bug, feature request, or proposal:

Proposal

What is the expected behavior?

Angular material2 list **[disabled]** item. An option to bind `[disabled]` to `md-list-item` in order to disable it (same as how the button disabled property works).

What is the current behavior?

no [disabled] property binding with md-list-item

What is the use-case or motivation for changing an existing behavior?

A list is not a button, but used for navigation and accessing data. [disabled] property can be very useful in such cases.

Which versions of Angular, Material, OS, browsers are affected?

All

chenop:
Please note I used disabled-button color for the disabled md-list-item

@EladBezalel #goodnessSquad

fixes #763


chenop:
I signed it!

EladBezalel:
@chenop can you please fix ci failures ?

chenop:
@EladBezalel Sure I will.

chenop:
@EladBezalel Done

src/lib/list/list.ts
```diff
@@ -7,13 +7,14 @@ import {
   Directive,
   ElementRef,
   Inject,
-  Input,
   OpaqueToken,
   Optional,
   Renderer,
   AfterContentInit,
+  Input,
```
EladBezalel: Please revert this and the line above

src/lib/list/list-item.html
```diff
@@ -1,7 +1,7 @@
-<div class="mat-list-item-content" [class.mat-list-item-focus]="_hasFocus"
-    md-ripple [mdRippleDisabled]="!isRippleEnabled()">
+<div class="mat-list-item-content" [class.mat-list-item-focus]="_hasFocus" [class.mat-list-item-disabled]="disabled" md-ripple [mdRippleDisabled]="!isRippleEnabled()">
```
EladBezalel: use _disabled and break line to not exceed 100 chars

jelbourn:
@kara what do you think of this? I find it a bit odd that there'd be a `disabled` property for something that doesn't inherently have behavior.

EladBezalel:
I think that makes sort of sense cause it behaves like a button..

jelbourn:
@EladBezalel it's not really meant to behave like a button, though; we're missing one of the list variants we originally planned (`md-action-list`) that would have an API like 
```html
<md-action-list>
  <button md-list-item>
  ...
</md-action-list>
```
Here, `disabled` has a lot more meaning. I worry about encouraging people to add behavior to static list items, which would lead to poor a11y.

chenop:
@EladBezalel Fixed last review comments.
Please let me know if we are Ok.

jelbourn:
Closing this since I don't believe that a standard list item should have a `disabled` property because it does not inherently have any behavior. This would be something that only affects the "actionable" list variants (which don't all exist yet).

gatimus:
`mat-list-item` does not have any behavior but wouldn't `mat-nav-list > a[mat-list-item]`?

badpk:
md-list-item will links create md-buttons - this element should have a disabled state so that i can prevent users from trying to click on inactive items.

PR #3626, opened 2017-03-15
fix(snack-bar): allow multi-line text

### LINKED ISSUE #1951: snack-bar multiline
proposal

According to the material design spec, a snack-bar toast can have 2 lines of text. I want the snack-bar to grow in height to a minimum of 2 lines when it has 2 lines.

The current behavior is that the text is pushed down under the bottom of the page

Angular 2.0.1
Material 2.0.0-alpha.10

crisbeto:
* Wraps multi-line text, instead of truncating it, inside of snack bars. **Note:** While the spec only mentions multi-line text on mobile, this seems to be a better approach than cutting it off at ~500px. We can revisit it if we start using media queries.
* Aligns the snack bar button margin to the spec.
* Simplifies some of the CSS to remove the need for a hardcoded height and negative margins.

Fixes #1951.

### DIFF
--- a/src/lib/snack-bar/simple-snack-bar.html
+++ b/src/lib/snack-bar/simple-snack-bar.html
@@ -1,2 +1,2 @@
-<span class="mat-simple-snackbar-message">{{message}}</span>
+{{message}}
 <button class="mat-simple-snackbar-action" *ngIf="hasAction" (click)="dismiss()">{{action}}</button>
--- a/src/lib/snack-bar/simple-snack-bar.scss
+++ b/src/lib/snack-bar/simple-snack-bar.scss
@@ -2,6 +2,8 @@
 @import '../core/style/button-common';
 @import '../core/style/list-common';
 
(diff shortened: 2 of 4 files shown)


GaganTut:
Is there any way to manually break at a certain point to create a pseudo list inside the snackbar? I tried inserting '<br>' as well as '\n' into the strings which both did not produce wanted effect. Any solutions?

jefbarn:
@GaganTut If you use the 'extraClasses' parameter and create a class with 'white-space: pre-line;' for instance then '\n' will work.

Hypenate:
Thanks @jefbarn! Yet it seems things have changed a tiny bit.

`this.snackBar.open('Kinda \n cool', null, { duration: 2000, panelClass: 'multilineSnackbar' });`

The css class `multilineSnackbar` 
```
.multilineSnackbar {
  white-space: pre-line;
}
```
_Note: the css class `multilineSnackbar` must be in the global css and not in the css of the component where you show the snackbar._

PR #3629, opened 2017-03-15
fix(button): persist theme color of button when leaving hover state

willshowell:
Leaving the hover state of colored flat and icon buttons exhibited a flicker of an uncolored focus overlay ([see this plunk](http://plnkr.co/edit/fkxW9Ll6G3jn3gIp4a0O?p=preview)). By slowing down the opacity transition, you can see it more clearly:

#### Before
before slowed hover animation

This is due to the focus overlay colors only being added to flat and icon buttons while the button is hovered. Thus, the *colored* focus overlay will show up immediately on hover and persist through the opacity transition to 1, but will also disappear immediately and **not** persist through the opacity transition back to 0.

Since the focus and hover states are already showing/hiding the focus overlay via opacity, the background color of the focus overlay can be set to the theme color all the time.

#### After
after slowed hover animation

### DIFF
--- a/src/lib/button/_button-theme.scss
+++ b/src/lib/button/_button-theme.scss
@@ -18,6 +18,10 @@
   &.mat-warn .mat-button-focus-overlay {
     background-color: mat-color($warn, 0.12);
   }
+
+  &[disabled] .mat-button-focus-overlay {
+    background-color: transparent;
+  }
 }
 
 // Applies a property to an md-button element for each of the supported palettes.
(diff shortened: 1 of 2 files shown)


src/lib/button/_button-theme.scss
```diff
@@ -54,20 +58,15 @@
   $foreground: map-get($theme, foreground);
 
   .mat-button, .mat-icon-button, .mat-raised-button, .mat-fab, .mat-mini-fab {
-    &.cdk-keyboard-focused {
```
crisbeto: Why did you remove this one @willshowell?

willshowell: Removing this line means that the overlay *always* has the correct coloring applied. [`%mat-button-base`](https://github.com/willshowell/material2/blob/fix-buttons-hover/src/lib/button/_button-base.scss#L58-L62) takes care of showing/hiding it when the button has `.cdk-keyboard-focused`.

crisbeto: Alright, makes sense.

PR #3634, opened 2017-03-16
docs(sidenav): correct example sidenav css height

gtzinos:


gtzinos:
I signed it! Thank you!

mmalerba:
@gtzinos can you explain why you think this change is necessary? it seems to work fine for me as is: http://plnkr.co/edit/pQ3E9PCA1d48idiyAlAp?p=preview

one small bug I did notice though, in the html the app is called "app" and in the css it's called "material-app" would be good to fix that

mmalerba:
@gtzinos Closing for now, please let me know if you still think this is necessary and I will reopen

PR #3644, opened 2017-03-16
fix(input-example): remove tables due to clashing styles

andrewseguin:
The tables in the example were clashing with the table styling in the docs site. This is the simplest way to resolve the issue


src/examples/input-form/input-form-example.css
```diff
@@ -1,7 +1,12 @@
...
+  max-width: 500px;
 }
 
-.example-full-width {
-  width: 100%;
+.input-row {
+  display: flex;
 }
+
+.mat-input-container {
```
jelbourn: All of the example styles should be prefixed with `example-`

andrewseguin:
Added example- prefix to all styles

mmalerba:
I already submitted changes to add classes to generated docs that we can target instead of elements like table, td, this shouldn't be necessary we just have to re-generate the docs and update the docs site css

PR #3646, opened 2017-03-16
feat(autocomplete): displayWith should accept a property name or function

### LINKED ISSUE #3642: feat(autocomplete): displayWith should accept a property name or function
feature

The displayWith input from autocomplete should accept a property name (string) or a function (current behavior)

The autocomplete only accepts a function as displayWith parameter

Most of the times I just need to show a property from my model and I think it's kind of a overkill to implement a function just to get a single property

PS: It should accept a multilevel property (aka: displayWith="a.b.c")

jefersonestevo:
Fixes #3642

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -295,8 +295,29 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   }
 
   private _setTriggerValue(value: any): void {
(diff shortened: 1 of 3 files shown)


jelbourn:
@jefersonestevo thanks for the PR! We considered this feature while developing the autocomplete and we  decided against it because it is not compatible with minification tools that perform property renaming (e.g. Closure Compiler). 

@kara we should expand the docs (perhaps with a note in the overview) with this rationale.

PR #3650, opened 2017-03-16
Export RIPPLE_FADE_IN_DURATION so that material users can import it

mertdeg2:
I see that there are some material tests that use this variable within fakeAsync tests. It looks like RIPPLE_FADE_IN_DURATION is not exported from material code base.

But material users might also need this in their unit tests.

### DIFF
--- a/src/lib/core/ripple/index.ts
+++ b/src/lib/core/ripple/index.ts
@@ -6,7 +6,7 @@ import {SCROLL_DISPATCHER_PROVIDER} from '../overlay/scroll/scroll-dispatcher';
 
 export {MdRipple, RippleGlobalOptions, MD_RIPPLE_GLOBAL_OPTIONS} from './ripple';
 export {RippleRef, RippleState} from './ripple-ref';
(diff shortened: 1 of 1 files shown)


src/lib/core/ripple/index.ts
```diff
@@ -6,7 +6,7 @@ import {SCROLL_DISPATCHER_PROVIDER} from '../overlay/scroll/scroll-dispatcher';
 
 export {MdRipple, RippleGlobalOptions, MD_RIPPLE_GLOBAL_OPTIONS} from './ripple';
 export {RippleRef, RippleState} from './ripple-ref';
-export {RippleConfig} from './ripple-renderer';
+export {RippleConfig, RIPPLE_FADE_IN_DURATION} from './ripple-renderer';
```
DevVersion: If we do this, then we also need to export the `RIPPLE_FADE_OUT_DURATION`.

mertdeg2: also exported RIPPLE_FADE_OUT_DURATION in a separate commit. Let me know what you think!

PR #3655, opened 2017-03-17
feat(datepicker): add keyboard support to calendar

mmalerba:
temporary demo: https://mmalerba-demo1.firebaseapp.com/datepicker

### DIFF
--- a/src/lib/datepicker/calendar.spec.ts
+++ b/src/lib/datepicker/calendar.spec.ts
@@ -144,6 +144,290 @@ describe('MdCalendar', () => {
       expect(calendarInstance._monthView).toBe(true, 'should be in month view');
       expect(testComponent.selected).toEqual(new SimpleDate(2017, 0, 31));
     });
+
+    describe('a11y', () => {
+      describe('calendar body', () => {
+        let calendarBodyEl: HTMLElement;
+
+        beforeEach(() => {
+          calendarBodyEl = calendarElement.querySelector('.mat-calendar-body') as HTMLElement;
+          expect(calendarBodyEl).not.toBeNull();
+
+          dispatchFakeEvent(calendarBodyEl, 'focus');
+          fixture.detectChanges();
+        });
+
+        it('should initially set start date active', () => {
+          expect(calendarInstance._activeDate).toEqual(new SimpleDate(2017, 0, 31));
+        });
+
+        describe('month view', () => {
+          it('should decrement date on left arrow press', () => {
+            dispatchKeyboardEvent(calendarBodyEl, 'keydown', LEFT_ARROW);
+            fixture.detectChanges();
+
+            expect(calendarInstance._activeDate).toEqual(new SimpleDate(2017, 0, 30));
+
+            calendarInstance._activeDate = new SimpleDate(2017, 0, 1);
+            fixture.detectChanges();
+
+            dispatchKeyboardEvent(calendarBodyEl, 'keydown', LEFT_ARROW);
+            fixture.detectChanges();
+
+            expect(calendarInstance._activeDate).toEqual(new SimpleDate(2016, 11, 31));
+          });
+
+          it('should increment date on right arrow press', () => {
+            dispatchKeyboardEvent(calendarBodyEl, 'keydown', RIGHT_ARROW);
+            fixture.detectChanges();
+
+            expect(calendarInstance._activeDate).toEqual(new SimpleDate(2017, 1, 1));
+
+            dispatchKeyboardEvent(calendarBodyEl, 'keydown', RIGHT_ARROW);
+            fixture.detectChanges();
+
+            expect(calendarInstance._activeDate).toEqual(new SimpleDate(2017, 1, 2));
+          });
+
+          it('should go up a row on up arrow press', () => {
+            dispatchKeyboardEvent(calendarBodyEl, 'keydown', UP_ARROW);
+            fixture.detectChanges();
+
+            expect(calendarInstance._activeDate).toEqual(new SimpleDate(2017, 0, 24));
+
+            calendarInstance._activeDate = new SimpleDate(2017, 0, 7);
+            fixture.detectChanges();
+
+            dispatchKeyboardEvent(calendarBodyEl, 'keydown', UP_ARROW);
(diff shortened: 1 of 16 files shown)


src/lib/datepicker/calendar.ts
```diff
@@ -97,57 +107,181 @@ export class MdCalendar implements AfterContentInit {
...
+      newDate = new SimpleDate(newDate.year, newDate.month, 0);
+    }
+
+    return newDate;
+  }
+
+  /**
+   * Adds the given number of months to the date. Months are counted as if flipping 12 pages for
+   * each year on a calendar and then finding the closest date in the new month. For example when
+   * adding 1 year to Feb 29 2016 the resulting date will be Feb 28 2017.
```
kara: Ultra nit: add comma after 'Feb 29, 2016' (same for line 241)? Love the detailed comment though

src/lib/datepicker/calendar.ts
```diff
@@ -97,57 +107,181 @@ export class MdCalendar implements AfterContentInit {
...
+    return date.add({'days': days});
+  }
+
+  /**
+   * Adds the given number of months to the date. Months are counted as if flipping pages on a
+   * calendar and then finding the closest date in the new month. For example when adding 1 month to
+   * Jan 31 2017 the resulting date will be Feb 28 2017.
+   */
+  private _addCalendarMonths(date: SimpleDate, months: number): SimpleDate {
+    let newDate = date.add({'months': months});
```
kara: You can shorten this to

```ts
let newDate = date.add({months})
```

(same on line 235)

src/lib/datepicker/calendar.ts
```diff
@@ -97,57 +107,181 @@ export class MdCalendar implements AfterContentInit {
...
   /** Whether the two dates represent the same view in the current view mode (month or year). */
-  private _isSameView(date1: SimpleDate, date2: SimpleDate) {
+  private _isSameView(date1: SimpleDate, date2: SimpleDate): boolean {
     return this._monthView ?
         date1.year == date2.year && date1.month == date2.month :
         date1.year == date2.year;
   }
+
+  /** Handles keydown events on the calendar body. */
+  _handleCalendarBodyKeydown(event: KeyboardEvent): void {
```
kara: Considered breaking this logic out into its own mini-service? 

Also, have you thought about separating into two functions -one for month view and one for year view? Not sure if that will actually shorten anything, but it may be easier to scan it without the conditionals.

src/lib/datepicker/calendar.spec.ts
```diff
@@ -231,6 +246,347 @@ describe('MdCalendar', () => {
...
+        calendarBodyEl = calendarElement.querySelector('.mat-calendar-body') as HTMLElement;
+        expect(calendarBodyEl).not.toBeNull();
+
+        dispatchFakeEvent(calendarBodyEl, 'focus');
+        fixture.detectChanges();
+      });
+
+      it('should not allow selection of disabled date in month view', () => {
+        expect(calendarInstance._monthView).toBe(true);
+        expect(calendarInstance._activeDate).toEqual(new SimpleDate(2017, 0, 1));
```
kara: Curious about this behavior. It seems like we allow people to focus the disabled dates, but not select them, whereas AngularJS Material doesn't allow highlighting of disabled dates at all. What's the a11y best practice for this?

src/lib/datepicker/calendar.spec.ts
```diff
@@ -231,6 +246,347 @@ describe('MdCalendar', () => {
...
+        dispatchKeyboardEvent(calendarBodyEl, 'keydown', ENTER);
+        fixture.detectChanges();
+
+        expect(calendarInstance._monthView).toBe(true);
+        expect(testComponent.selected).toBeNull();
+      });
+    });
+  });
+
+  describe('a11y', () => {
```
kara: Nit: Two describe blocks both named `a11y`?

mmalerba: One of them is `MdCalendar.calendar with date filter.a11y` the other is `MdCalendar.a11y` (though it should be `MdCalendar.standard calendar.a11y` so making that change).

src/lib/datepicker/calendar.spec.ts
mmalerba: Yeah this was more of a practical decision, since someone could theoretically create a calendar with no maxDate but a date filter that disables all dates > 1/1/17. In this case if the user tried to arrow past 1/1/17 everything would just lock up as it searched for the next enabled date. I put a TODO and I'm curious to see how AngularJS Material handled this

src/lib/datepicker/calendar.spec.ts
kara: Aha I see. The diff was a bit misleading.

src/lib/datepicker/calendar.ts
mmalerba: I'm not sure what we'd gain by making it a service, I split out 2 separate functions for month view and year view though

mmalerba:
Yeah focus state when you open the calendar is still kind of wonky, I'll address that in a different PR. Addressed other comments, PTAL.

src/lib/datepicker/calendar.ts
kara: Just thought the file was getting pretty long, but if you don't think so, doesn't matter too much.

src/lib/datepicker/calendar.ts
```diff
@@ -97,57 +107,208 @@ export class MdCalendar implements AfterContentInit {
...
+
+  /** Adds the given number of days to the date. */
+  private _addCalendarDays(date: SimpleDate, days: number): SimpleDate {
+    return date.add({days});
+  }
+
+  /**
+   * Adds the given number of months to the date. Months are counted as if flipping pages on a
+   * calendar and then finding the closest date in the new month. For example when adding 1 month to
+   * Jan 31, 2017 the resulting date will be Feb 28, 2017.
```
kara: Haha, sorry for being unclear. I meant a comma after "2017", not in the date.

```
For example, when adding 1 month to Jan 31, 2017, the resulting date will be ...
```

PR #3668, opened 2017-03-18
build: dont run karma if compilation failed

DevVersion:
* Currently when the TypeScript compilation fails, our gulp process will still call the `runTests` method. 
* Also fixes that the TS-compiler exits the process. Gulp should handle the error (similar as before)
* Improves formatting of the diagnostics in the `ts-compiler`.

### DIFF
--- a/tools/gulp/tasks/unit-test.ts
+++ b/tools/gulp/tasks/unit-test.ts
@@ -75,13 +75,15 @@ gulp.task('test', [':test:deps'], () => {
   });
 
   // Refreshes Karma's file list and schedules a test run.
-  let runTests = () => {
-    server.refreshFiles().then(() => server._injector.get('executor').schedule());
+  let runTests = (err?: Error) => {
+    if (!err) {
+      server.refreshFiles().then(() => server._injector.get('executor').schedule());
+    }
   };
 
   // Boot up the test server and run the tests whenever a new browser connects.
   server.start();
-  server.on('browser_register', runTests);
+  server.on('browser_register', () => runTests());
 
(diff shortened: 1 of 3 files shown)


DevVersion:
I was hoping to also find a way to stop running Karma tests, but it looks like this needs a public API on karma.

tools/gulp/tasks/unit-test.ts
```diff
@@ -75,13 +75,15 @@ gulp.task('test', [':test:deps'], () => {
   });
 
   // Refreshes Karma's file list and schedules a test run.
```
jelbourn: Expand comment to say something like "Only run tests if TypeScript compilation is successful" ?

tools/gulp/tasks/unit-test.ts
```diff
@@ -75,13 +75,15 @@ gulp.task('test', [':test:deps'], () => {
...
+  let runTests = (err?: Error) => {
+    if (!err) {
+      server.refreshFiles().then(() => server._injector.get('executor').schedule());
+    }
   };
 
   // Boot up the test server and run the tests whenever a new browser connects.
   server.start();
-  server.on('browser_register', runTests);
+  server.on('browser_register', () => runTests());
```
jelbourn: Why the extra level of wrapping here?

DevVersion: Because this event comes from Karma and it calls the callback with the `browser_data` as first argument.

tools/gulp/tasks/unit-test.ts
DevVersion: Done. Anything else?

PR #3671, opened 2017-03-18
fix(tooltip): wrong position when using OnPush change detection

### LINKED ISSUE #3497: mdTooltip wrong position when used inside a component with OnPush
https://gist.run/?id=2df959fa72c4d9839dd4e621d73a0f73

image

my guess is that it tries to calculate the position before the content is set (you can actually see it flicker as an empty tooltip for a few milliseconds)

cc @andrewseguin

crisbeto:
Fixes the tooltip having a wrong position if it is inside a component with `OnPush` change detection. The issue was due to the fact that when `OnPush` is used, the tooltip text isn't rendered at the moment when the element is positioned.

Fixes #3497.

### DIFF
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -410,7 +410,7 @@ describe('MdTooltip', () => {
 
       fixture.detectChanges();
 
-      // wait till animation has finished
+      // wait until animation has finished
       tick(500);
 
       // Make sure tooltip is shown to the user and animation has finished
@@ -431,6 +431,14 @@ describe('MdTooltip', () => {
       // On animation complete, should expect that the tooltip has been detached.
       flushMicrotasks();
       expect(tooltipDirective._tooltipInstance).toBeNull();
+    }));
+
+    it('should have rendered the tooltip text on init', fakeAsync(() => {
+      dispatchFakeEvent(buttonElement, 'mouseenter');
+      tick(0);
+
+      const tooltipElement = overlayContainerElement.querySelector('.mat-tooltip') as HTMLElement;
+      expect(tooltipElement.textContent).toContain('initial tooltip message');
     }));
   });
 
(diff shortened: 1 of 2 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -461,4 +463,13 @@ export class TooltipComponent {
...
     }
   }
+
+  /**
+   * Ensures that the tooltip text has rendered. Mainly used for rendering the initial text
+   * before positioning a tooltip, which can be problematic in components with OnPush change
+   * detection.
+   */
+  _updateView(): void {
+    this._changeDetectorRef.detectChanges();
```
DevVersion: Is there any reason why you use `detectChanges` instead of `markForCheck`?

crisbeto: Hmm, I thought I tried it and it wasn't working. Looking at it again, it seems like it's the same. I'll change it.

src/lib/tooltip/tooltip.spec.ts
```diff
@@ -432,6 +432,15 @@ describe('MdTooltip', () => {
...
     }));
+
+    it('should have rendered the tooltip text on init', fakeAsync(() => {
+      dispatchFakeEvent(buttonElement, 'mouseenter');
+      fixture.detectChanges();
+      tick(0);
+
+      const tooltipElement = overlayContainerElement.querySelector('.mat-tooltip') as HTMLElement;
+      expect(tooltipElement.textContent).toContain('initial tooltip message');
+    }));
```
jelbourn: Does there need to be a test component w/ `OnPush` that would have exhibited the original behavior?

crisbeto: It already is using the one that we use for testing `OnPush`: https://github.com/angular/material2/pull/3671/files#diff-90d01e051e22bab8526ca923091f589fR391

PR #3672, opened 2017-03-18
fix(progress-spinner): not redrawing when changing modes

### LINKED ISSUE #3648: Progress bar: Incorrect behavior switching between determinate and indeterminate

Switching from indeterminate to determinate state on the progress spinner sometimes breaks.

Spinner's view should be consistent.

Sometimes the spinner ends up looking like this:
image

Go to the spinner example on the docs site. Toggle between determinate and indetermine states. Sometimes the spinner ends up in a strange state. 

https://material.angular.io/components/component/progress-spinner

The spinner should be very consistent about how it looks in determinate

crisbeto:
* Fixes the progress spinner not redrawing when the mode is changed dynamically from `indeterminate` to `determinate`.
* Some minor cleanup in the progress spinner.

Fixes #3648.

### DIFF
--- a/src/lib/progress-spinner/progress-spinner.ts
+++ b/src/lib/progress-spinner/progress-spinner.ts
@@ -134,7 +134,7 @@ export class MdProgressSpinner implements OnDestroy {
   set value(v: number) {
     if (v != null && this.mode == 'determinate') {
       let newValue = clamp(v);
-      this._animateCircle((this.value || 0), newValue, linearEase, DURATION_DETERMINATE, 0);
+      this._animateCircle(this.value || 0, newValue);
       this._value = newValue;
     }
   }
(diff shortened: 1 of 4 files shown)


src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -237,16 +237,23 @@ export class MdProgressSpinner implements OnDestroy {
 
   /**
    * Removes interval, ending the animation.
+   * @param redraw Whether to redraw the circle after the animation is cleared.
    */
-  private _cleanupIndeterminateAnimation() {
-    this.interdeterminateInterval = null;
+  private _cleanupIndeterminateAnimation(redraw = false) {
```
jelbourn: Rather than adding a `redraw` option here and then calling `_animateCircle`, can we instead just call `_animateCircle` after  `_cleanupIndeterminateAnimation` in the setter? I generally like to avoid boolean parameters that tell a function to do extra work, and the call site doesn't make it obvious what the parameter is for.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -150,13 +150,13 @@ export class MdProgressSpinner implements OnDestroy {
   get mode() {
     return this._mode;
   }
-  set mode(m: ProgressSpinnerMode) {
-    if (m == 'indeterminate') {
+  set mode(mode: ProgressSpinnerMode) {
+    if (mode === 'indeterminate') {
```
jelbourn: Do we need to also check that the mode has actually _changed_?

crisbeto:

PR #3675, opened 2017-03-19
build: add toggle for OnPush change detection in demo app

crisbeto:
Adds a toggle for enabling `OnPush` change detection in the demo app. This should make it easier to test out components against the different change detection strategies.

### DIFF
--- a/src/demo-app/demo-app/demo-app.ts
+++ b/src/demo-app/demo-app/demo-app.ts
@@ -69,7 +69,7 @@ export class DemoApp {
   constructor(private _element: ElementRef) {
     // Some browsers will throw when trying to access localStorage in incognito.
     try {
-      this.changeDetectionStrategy = localStorage.getItem(changeDetectionKey) || 'Default';
+      this.changeDetectionStrategy = window.localStorage.getItem(changeDetectionKey) || 'Default';
     } catch (error) {
       console.error(error);
     }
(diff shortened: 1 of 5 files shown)


src/demo-app/demo-app/demo-app.ts
```diff
@@ -17,8 +26,12 @@ export class Home {}
   templateUrl: 'demo-app.html',
   styleUrls: ['demo-app.css'],
   encapsulation: ViewEncapsulation.None,
+  changeDetection: isPushChangeDetection ?
+    ChangeDetectionStrategy.OnPush :
+    ChangeDetectionStrategy.Default
```
DevVersion: Why not storing this above?

```ts
let changeDetection = ChangeDetectionStrategy.Default;

try {
  if (localStorage.getItem(changeDetectionKey) {
    changeDetection = ChangeDetectionStrategy.OnPush;
  }
} catch (e) {
  console.error(e);
}

@Component({
  changeDetection: changeDetection
}
```

crisbeto: Using the boolean is easier for displaying in the UI. AFAIK the detection strategies map to numbers.

DevVersion: True. Yeah I initially thought that `setItem(key, changeDetectionStrategy)` would work, but it maps to integers.

src/demo-app/demo-app/demo-app.ts
```diff
@@ -17,8 +26,12 @@ export class Home {}
   templateUrl: 'demo-app.html',
   styleUrls: ['demo-app.css'],
   encapsulation: ViewEncapsulation.None,
+  changeDetection: isPushChangeDetection ?
```
jelbourn: I don't think this will work with AoT

crisbeto: It seems to work on the CI which we run against the demo app.

crisbeto:
Re-did this one to avoid AoT issues @jelbourn.

kara:
@crisbeto  There are some Google test failures related to internal tools not liking `localStorage`. @jelbourn and I were thinking changing it to `window.localStorage` may appease them. Can you update?

crisbeto:
Done. Also needed a rebase.

PR #3687, opened 2017-03-20
perf(scroll-dispatcher): avoid triggering change detection on scroll

crisbeto:
* Avoids triggering change detection when listening to global scroll events in the `ScrollDispatcher`.
* Avoids triggering change detection when scrolling inside of `Scrollable` instances.
* Switches a `ScrollDispatcher` test to use spies, instead of toggling booleans.

### DIFF
--- a/src/lib/core/overlay/scroll/scroll-dispatcher.spec.ts
+++ b/src/lib/core/overlay/scroll/scroll-dispatcher.spec.ts
@@ -41,15 +41,15 @@ describe('Scroll Dispatcher', () => {
 
     it('should notify through the directive and service that a scroll event occurred',
         fakeAsync(() => {
-      let hasDirectiveScrollNotified = false;
       // Listen for notifications from scroll directive
-      let scrollable = fixture.componentInstance.scrollable;
-      scrollable.elementScrolled().subscribe(() => { hasDirectiveScrollNotified = true; });
+      const scrollable = fixture.componentInstance.scrollable;
+      const directiveSpy = jasmine.createSpy('directive scroll callback');
+      scrollable.elementScrolled().subscribe(directiveSpy);
 
       // Listen for notifications from scroll service with a throttle of 100ms
       const throttleTime = 100;
-      let hasServiceScrollNotified = false;
-      scroll.scrolled(throttleTime, () => { hasServiceScrollNotified = true; });
+      const serviceSpy = jasmine.createSpy('service scroll callback');
+      scroll.scrolled(throttleTime, serviceSpy);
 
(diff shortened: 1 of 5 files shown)


src/lib/core/overlay/scroll/scroll-dispatcher.ts
```diff
@@ -69,10 +71,12 @@ export class ScrollDispatcher {
     this._scrolledCount++;
 
     if (!this._globalSubscription) {
-      this._globalSubscription = Observable.merge(
-        Observable.fromEvent(window.document, 'scroll'),
-        Observable.fromEvent(window, 'resize')
-      ).subscribe(() => this._notify());
+      this._globalSubscription = this._ngZone.runOutsideAngular(() => {
```
jelbourn: Add a comment explaining why we need to escape from the Angular zone here (and in the directive)?
(I generally like to document the "why" every time we do this)

src/lib/core/overlay/scroll/scroll-dispatcher.spec.ts
```diff
@@ -41,32 +41,54 @@ describe('Scroll Dispatcher', () => {
...
+
+    it('should not execute the scrollable events in the Angular zone', () => {
+      const spy = jasmine.createSpy('zone unstable callback');
+      const subscription = fixture.ngZone.onUnstable.subscribe(spy);
+
+      dispatchFakeEvent(fixture.componentInstance.scrollingElement.nativeElement, 'scroll');
+
+      expect(spy).not.toHaveBeenCalled();
+      subscription.unsubscribe();
+    });
```
jelbourn: Would it be worth also adding a unit test that somehow asserts that change detection wasn't run (which is the ultimate goal)?

crisbeto: This is what the test is doing: If we didn't hit the zone, we can assume that change detection didn't run. I can also rework it so it uses a `ngDoCheck` hook with a spy instead.

src/lib/core/overlay/scroll/scroll-dispatcher.ts
crisbeto: I guess, although the unit tests could be considered a form of documentation.

src/lib/core/overlay/scroll/scroll-dispatcher.spec.ts
jelbourn: That's fair; I was more thinking a test that looked at something in the DOM not being updated, but it's all really the same.

PR #3690, opened 2017-03-20
bug(autocomplete): deselect the other options when the user select an option

### LINKED ISSUE #3689: bug(autocomplete): option keeps the mat-selected class
bug

After selecting a new option, the previous selected option should not keep the "mat-selected" class

After selecting an option, this options keeps the "mat-selected" class forever

[Plunker](http://plnkr.co/edit/mIdDpJfoXjQB48dufb8X?p=preview)

Autocomplete:
image

After selecting 1 option:
image

After selecting all options:
image

jefersonestevo:
Fixes #3689

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -306,6 +306,12 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
    */
   private _setValueAndClose(event: MdOptionSelectionChange | null): void {
     if (event) {
+      this.autocomplete.options.forEach((option) => {
+        if (option != event.source) {
+          option.deselect();
+        }
+      });
+
       this._setTriggerValue(event.source.value);
       this._onChange(event.source.value);
     }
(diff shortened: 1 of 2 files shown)


jelbourn:
@jefersonestevo looks like this change needs unit tests as well

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -306,6 +306,12 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
    */
   private _setValueAndClose(event: MdOptionSelectionChange | null): void {
     if (event) {
+      this.autocomplete.options.forEach((option) => {
+        if (option != event.source) {
+          option.deselect();
```
kara: Calling `deselect()` here will emit a selection change event for every option on the list. That's fine, but it will change the behavior of subscriptions to `optionSelections`. Instead of getting 1 event on selection, people will get 1 event per option (so, 50 events for a 50 option list). So I'd also add a filter to `optionSelections` to ensure we're only catching events from options that have selected set to true.

Also can you move this out into its own function?

jefersonestevo: Done, but instead of setting a filter on `optionSelections`, I've change to only call the deselect function on the option that was previously selected.

Please let me know if this PR needs any other changes

jefersonestevo:
@kara there are any other changes needed here?

Issue #2620, opened 2017-01-12
Input should not animate placeholder when value is set programmatically


kara:
This is to match select functionality. To replicate:

```html
<input [formControl]="ctrl">

<button (click)="ctrl.setValue('wat')">SET VALUE</button>
```

The label animates as the value comes in, but it should come in immediately.

jelbourn:
Seems to be done via #3691

PR #3691, opened 2017-03-20
fix(input): don't animate label when value is set programmatically

### LINKED ISSUE #3641: Input placeholder animates when initialized with a value

Bug

When an input is initialized with a value, the input's placeholder should not animate to be above the value.

The animation always triggers. Looks funny when the input is loaded, especially on pages where there are many inputs.

To see an example of this, go to the Examples tab at http://localhost:4200/components/component/slider

mmalerba:
note: this fixes the bug in most cases, but not when programmatically setting the value from non-empty to empty

fixes #3641
partially addresses #2620


kara:
@mmalerba You may want to fix the commit message though (has a typo)

mmalerba:
done

PR #3693, opened 2017-03-21
chore: remove leftover expression

crisbeto:
This is a follow-up to #2849. It looks like I was a little too quick to switch the statement and forgot to remove the `typeof`.

### DIFF
--- a/src/lib/core/compatibility/compatibility.ts
+++ b/src/lib/core/compatibility/compatibility.ts
@@ -171,7 +171,7 @@ export class CompatibilityModule {
   }
 
   constructor(@Optional() @Inject(DOCUMENT) document: any) {
-    if (isDevMode() && typeof document && !document.doctype) {
+    if (isDevMode() && document && !document.doctype) {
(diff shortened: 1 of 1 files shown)


src/lib/core/compatibility/compatibility.ts
```diff
@@ -171,7 +171,7 @@ export class CompatibilityModule {
   }
 
   constructor(@Optional() @Inject(DOCUMENT) document: any) {
-    if (isDevMode() && typeof document && !document.doctype) {
+    if (isDevMode() && document && !document.doctype) {
```
jelbourn: Won't just `document` by itself potentially be an issue for Universal? `typeof document == 'object'`?

crisbeto: It won't, because it'll hit the parameter, before it tries to look for it globally.

crisbeto:
I've rolled these changes into #3693. Closing to avoid potential merge conflicts.

PR #3608, opened 2017-03-14
feat: angular 4 compatibility

crisbeto:
* Bumps the required Angular version to 4 and TypeScript to 2.1.6.
* Fixes deprecation warnings for `<template>` usages.
* Fixes any unit and e2e test failures.
* Includes the new animations module and switches any components that use animations to it.
* Fixes issues with the various test apps.

Fixes #3357.
Fixes #3336.
FIxes #3301.

TODO:
- [x] It looks like the animation durations are a bit longer with Angular 4, compared to 2. Following this up with Matias.
- [x] Doing a manual run-through of all components.
- [x] Adding `@angular/animations/browser` to the Rollup config.
- [x] Verifying that the `ng.platformBrowser.animations` package name in the Rollup UMD bundle config is correct.
- [x] Errors being thrown if state changes mid-animation (e.g. opening multiple snack bars at the same time or quickly switching tabs).

PR #3695, opened 2017-03-21
build: restructure release output

DevVersion:
**Work in Progress PR** (for review)

> Based on top of #3608 

**Note**: AOT Travis check will likely fail at the moment. Not updated yet.

cc. @jelbourn


tools/gulp/constants.ts
```diff
@@ -17,6 +29,10 @@ export const HTML_MINIFIER_OPTIONS = {
   removeAttributeQuotes: false
 };
 
+export const UGLIFYJS_OPTIONS = {
+  preserveComments: 'license'
+};
```
jelbourn: I think we should just define these where we call uglify if we're only doing it in one place.

tools/gulp/tasks/aot.ts
```diff
@@ -1,21 +1,21 @@
...
+task('aot:deps', sequenceTask('build:devapp', ':package:release', 'aot:copy-release'));
 
-/** Builds the demo-app with the Angular compiler to verify that all components work. */
-task('aot:build', ['aot:prepare'], execNodeTask(
-  '@angular/compiler-cli', 'ngc', ['-p', join(DIST_ROOT, 'tsconfig-aot.json')])
-);
+// As a workaround for https://github.com/angular/angular/issues/12249, we need to
+// copy the Material ESM output inside of the demo-app output.
+task('aot:copy-release', () => {
+  copySync(DIST_RELEASE, join(DIST_DEMOAPP, 'material'));
```
jelbourn: Any reason not to do
```ts
task('aot:copy-release', () => src(DIST_RELEASE).pipe(dest(join(DIST_DEMOAPP, 'material'))));
```
?
(I thought this was the most "gulp-y" way to do it, but I could be wrong)

tools/gulp/tasks/release.ts
```diff
@@ -1,50 +1,106 @@
...
+
+task('build:release', sequenceTask(
+  'library:build',
+  ':package:release',
+));
+
+/** Task that combines intermediate build artifacts into the release package structure. */
+task(':package:release', [
+  ':package:fix-metadata',
+  ':package:metadata',
```
jelbourn: Should only call `:packagae:metadata`, which then depends on `:package:fix-metadata`?
(to make sure the order of operations is correct since this isn't a `runSequence`)

DevVersion: 👍

tools/gulp/tasks/aot.ts
DevVersion: I tried, but it really felt weird because I would need to specify a glob to match all files. (Just specifying the directory didn't work)

```ts
task('aot:copy-release', () => src(DIST_RELEASE + '**/*').pipe(dest(join(DIST_DEMOAPP, 'material'))));
```

It is definitely a `gulp-y` way, but I think it's just not very clear and also we would need to have a glob.

DevVersion:
Closing as we are merging this together with @jelbourn's changes: #3751

Issue #3496, opened 2017-03-08
autocomplete - HOME and END keys not working


mxii:
#### Bug, feature request, or proposal:
I would like to use those keys as "normal" to navigate through my input string..
Like "shift + POS1" will select all.. :)

#### What is the expected behavior?
Maybe behavior of those keys could be configured?!

#### What is the current behavior?
Key will only be used to navigate through autocomplete-results

#### What are the steps to reproduce?
.. should be clear ..

#### What is the use-case or motivation for changing an existing behavior?
Better usability.
If i want to clear an input field, i will always press "shift + pos1" .. but now its not working anymore! =/
I would guess I am not the only one doing it like this..

#### Which versions of Angular, Material, OS, browsers are affected?
angular4, material.beta2

#### Is there anything else we should know?
No, thanks! :)

crisbeto:
For reference, POS1 is the Home key on German keyboards.

jefersonestevo:
+1

As [angularjs autocomplete](https://material.angularjs.org/latest/demo/autocomplete) does allow to use these keys on the input, I agree that this should work on this component too.

jefersonestevo:
@crisbeto @jelbourn I was looking to implement this, and I think the implementation here should consider the following paths:
- If the user press Shift+Home or Shift+End, it will select the whole text, as a normal input field does
- If there are only one or none option available, the Home and End key should behave like the input text field (changing the cursor to the beggining and end of the input field)

Both of these behaviours can be achieved by some conditional tests on the `onKeydown` method of `ListKeyManager`, but as this component is used in multiple components, If I put some code in it, it may break some other components behaviour. What may be done here is to add a second parameter (optional) to the `onKeydown` method to "decide" if the `event.preventDefault()` should be called or not. But I don't really know if it's the best way to implement this.

Example (just a simple conditional that may be extracted to a callback):
```
  onKeydown(event: KeyboardEvent): void {
    let shouldPreventDefault: boolean = true;
    switch (event.keyCode) {
      case DOWN_ARROW:
        this.setNextItemActive();
        break;
      case UP_ARROW:
        this.setPreviousItemActive();
        break;
      case HOME:
        this.setFirstItemActive();
        shouldPreventDefault = !event.shiftKey && this._items.length > 1;
        break;
      case END:
        this.setLastItemActive();
        shouldPreventDefault = !event.shiftKey && this._items.length > 1;
        break;
      case TAB:
        // Note that we shouldn't prevent the default action on tab.
[...]

PR #3709, opened 2017-03-21
fix(autocomplete): not scrolling to active option when pressing home/end

crisbeto:
* Fixes the autocomplete not scrolling to active items that were highlighted by pressing `home` or `end`.
* Switches up the logic to scroll to any newly-focused options. This should make it more future-proof, if we add more shortcuts to the key manager.

Relates to #3496.


stgeorge:
I don't think this "resolves" 3496... the home/end keys should act like home/end keys while on the input string. Right now they are blocked which should never happen.

crisbeto:
I haven't mentioned anywhere that this should resolve #3496 @stgeorge, however #4544 should.

PR #3724, opened 2017-03-22
fix(input): don't add empty prefix & suffix wrappers

### LINKED ISSUE #3659: md-input-container: mdPrefix and mdSuffix divs should not be present when there is no content
#### Proposal:
I'd like to propose adding an `ngIf` on both the `.mat-input-prefix` and `.mat-input-suffix` divs in the `md-input-container` template so that they are only in the DOM if the user has included a prefix/suffix.

I'd like to be able to add space between the infix and the affixes, but only when an affix is present.

Currently, if I add any sort of spacing there, the space shows up even without the affix, as seen in the second input [here](https://plnkr.co/edit/ftKWagGNYlR263WTN0ZE?p=preview).

mmalerba:
demo: http://mmalerba-demo4.firebaseapp.com/input

fixes #3659

### DIFF
--- a/src/lib/input/input-container.html
+++ b/src/lib/input/input-container.html
@@ -1,6 +1,6 @@
 <div class="mat-input-wrapper">
   <div class="mat-input-table">
-    <div class="mat-input-prefix">
+    <div class="mat-input-prefix" *ngIf="_prefixChildren.length">
       <!-- TODO(andrewseguin): remove [md-prefix] -->
       <ng-content select="[mdPrefix], [matPrefix], [md-prefix]"></ng-content>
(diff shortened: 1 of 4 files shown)


src/lib/input/input-container.html
```diff
@@ -23,7 +23,7 @@
       </span>
     </div>
 
-    <div class="mat-input-suffix">
+    <div class="mat-input-suffix" *ngIf="_suffixChildren.length">
```
jelbourn: Add a unit test for this?

mmalerba: done

PR #3727, opened 2017-03-22
fix(autocomplete): do not trigger submit on ENTER

### LINKED ISSUE #3159: md-autocomplete trigger submit on ENTER
md-autocomplete trigger submit form by ENTER click on option:
Plunker template: https://plnkr.co/edit/0a8lqrodE9WVTZEJprlI?p=preview

jefersonestevo:
When an option is selected with the ENTER key, it should not submit the form and prevent the default input behavior

Fixes #3159

### DIFF
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -611,6 +611,24 @@ describe('MdAutocomplete', () => {
       });
     }));
 
+    it('should not call the default input behaviour when an option is selected with ENTER',
+        async(() => {
+      fixture.whenStable().then(() => {
+        fixture.componentInstance.trigger._handleKeydown(DOWN_ARROW_EVENT);
+
+        fixture.whenStable().then(() => {
+          fixture.detectChanges();
+
+          spyOn(ENTER_EVENT, 'preventDefault');
+
+          fixture.componentInstance.trigger._handleKeydown(ENTER_EVENT);
+          fixture.detectChanges();
+
+          expect(ENTER_EVENT.preventDefault).toHaveBeenCalled();
+        });
+      });
+    }));
+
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -611,6 +611,24 @@ describe('MdAutocomplete', () => {
       });
     }));
 
+    it('should not call the default input behaviour when an option is selected with ENTER',
+        async(() => {
+      fixture.whenStable().then(() => {
```
crisbeto: I don't think you need the `fixture.whenStable` here.

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -611,6 +611,24 @@ describe('MdAutocomplete', () => {
...
+        fixture.componentInstance.trigger._handleKeydown(DOWN_ARROW_EVENT);
+
+        fixture.whenStable().then(() => {
+          fixture.detectChanges();
```
crisbeto: You shouldn't need to call `detectChanges` here and below on line 625.

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -611,6 +611,24 @@ describe('MdAutocomplete', () => {
       });
     }));
 
+    it('should not call the default input behaviour when an option is selected with ENTER',
```
crisbeto: You could shorten this to something like `should prevent the default enter key action` so it doesn't have to be on separate lines.

src/lib/autocomplete/autocomplete.spec.ts
jefersonestevo: Done. I've removed the extra `detectChanges` but had to keep the `fixture.whenStable()` so that the test could simulate the "open" and "select first option" action before it press the enter key

PR #3730, opened 2017-03-22
docs(input): add example of a clearable input

### LINKED ISSUE #3725: mdInput clear input button
Feature Request

When an mdInput text field has a value, there should be a clear (X) button on the right of the input to clear the text field.

No button, user has to manually clear, or dev has to create another directive

I'm currently working on a search field, and expected the `<input type="search" mdinput></input>` to give users a clear button, as per the docs here https://material.io/guidelines/components/text-fields.html#text-fields-search-filter - however I'm either missing something, or it's not provided.

mmalerba:
fixes #3725

### DIFF
--- a/src/material-examples/example-module.ts
+++ b/src/material-examples/example-module.ts
@@ -83,7 +83,7 @@ import {DatepickerMinMaxExample} from './datepicker-min-max/datepicker-min-max-e
 import {DatepickerFilterExample} from './datepicker-filter/datepicker-filter-example';
 import {DatepickerTouchExample} from './datepicker-touch/datepicker-touch-example';
 import {DatepickerApiExample} from './datepicker-api/datepicker-api-example';
-
+import {InputClearableExample} from './input-clearable/input-clearable-example';
 import {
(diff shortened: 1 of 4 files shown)


jelbourn:
I'm hesitant to change the examples here until we have them building.

jelbourn:
@mmalerba can you rebase?

mmalerba:
done

src/examples/input-clearable/input-clearable-example.html
```diff
@@ -0,0 +1,4 @@
+<md-input-container class="example-input-container">
+  <input mdInput type="text" placeholder="Clearable input" [(ngModel)]="value"/>
+  <md-icon mdSuffix *ngIf="value" (click)="value=''">close</md-icon>
```
jelbourn: Should be an `md-icon-button` for a11y

mmalerba: So actually... `md-icon-button` doesn't really work well as a suffix currently. It's too tall and it causes the placeholder and input text to be misaligned. Filed #4382 to fix this

mmalerba:
icon-buttons are supported inside inputs now, updated example

jelbourn:
Needs rebase

Issue #3722, opened 2017-03-22
bug(autocomplete): wrong positioning on dialog


jefersonestevo:
#### Bug, feature request, or proposal:
bug

#### What is the expected behavior?
The autocomplete should be positioned correctly independent if it's inside a dialog

#### What is the current behavior?
When you open a dialog, it automatic focus on the first item. If the dialog component populate the options list of the autocomplete on ngOnInit (with an asynchronous call), the autocomplete gets the wrong position

PS: The error only happens if you populate the autocomplete list on an asynchronous call

#### What are the steps to reproduce?
[Plunker](http://plnkr.co/edit/bWbspcX0puW2BeahWhhU?p=preview)

Image:
![image](https://cloud.githubusercontent.com/assets/3085585/24200631/70d39be6-0eec-11e7-9e16-f5fb12a277e9.png)

#### Which versions of Angular, Material, OS, browsers are affected?
@angular/material: beta.2

PR #3745, opened 2017-03-23
fix(autocomplete): reposition panel on scroll

crisbeto:
Repositions the autocomplete panel if the users scrolls while it is open.


jefersonestevo:
@crisbeto is this goind to fix #3722

crisbeto:
Probably not, but I think I have an idea for it.

Issue #3651, opened 2017-03-16
viewContainerRef option in MdDialogConfig needs docs


jelbourn:
The property currently has no JsDoc description.

M-a-c:
@jelbourn @DevVersion can I get someone to pr #3825 ?

M-a-c:
@jelbourn This issue can be closed 😃

PR #3747, opened 2017-03-23
docs(Dialog): MdDialogConfig ViewContainerRef description

M-a-c:
#3651 
I looked to find an example of what this is actually used for.
I tried to explain it a little bit more. (line 2)

https://github.com/angular/material2/blob/master/src/lib/core/portal/portal.ts#L79

Does this JsDoc summary suffice?

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -17,9 +17,9 @@ export interface DialogPosition {
 export class MdDialogConfig {
 
   /** Where the attached component should live in Angular's *logical* component tree.
-   * This is different from where the component *renders*, which is determined by
-   * the PortalHost (the component that was passed into MdDialog.open().
-   * Note the origin is necessary when the host is outside of the Angular application context. */
+  * This affects what is available for injection and the change detection order for the
+  * component instantiated inside of the dialog. This does not affect where the dialog
+  * content will be rendered. */
   viewContainerRef?: ViewContainerRef;
 
   /** The ARIA role of the dialog element. */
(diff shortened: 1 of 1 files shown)


src/lib/dialog/dialog-config.ts
```diff
@@ -15,6 +15,11 @@ export interface DialogPosition {
  * Configuration for opening a modal dialog with the MdDialog service.
  */
 export class MdDialogConfig {
+
+  /** Where the attached component should live in Angular's *logical* component tree.
+   * This is different from where the component *renders*, which is determined by
+   * the PortalHost (the component that was passed into MdDialog.open().
+   * Note the origin is necessary when the host is outside of the Angular application context. */
```
jelbourn: This isn't quite right. A little more accurate would be:
```ts
/** 
  * Where the attached component should live in Angular's *logical* component tree.
  * This affects what is available for injection and the change detection order for the 
  * component instantiated inside of the dialog. This does not affect where the dialog 
  * content will be rendered.
  */
```

M-a-c: Yeah this makes more sense. I'll switch that out.

src/lib/dialog/dialog-config.ts
```diff
@@ -15,6 +15,11 @@ export interface DialogPosition {
  * Configuration for opening a modal dialog with the MdDialog service.
  */
 export class MdDialogConfig {
+
+  /** Where the attached component should live in Angular's *logical* component tree.
+  * This affects what is available for injection and the change detection order for the
+  * component instantiated inside of the dialog. This does not affect where the dialog
+  * content will be rendered. */
```
jelbourn: The JsDoc syntax isn't quite right. It goes
```ts
/**
 * Line one...
 * Line two...
 * Line three ...
 */
```

M-a-c:
@jelbourn LGTY?
Sorry about the styling misunderstanding, I have looked through the style guide now. I apologize, I should have read that before contributing.

M-a-c:
Noticed that my local and branches are all messed up. 
closing this creating new branch/w pr, using just these changes.

PR #3751, opened 2017-03-24
build: update packaging to match angular/angular

jelbourn:
Updates the build to support the new package structure used by `angular/angular`. 

The first commit is from @DevVersion and the subsequent commits are from me.


tools/gulp/tasks/release.ts
```diff
@@ -94,12 +161,42 @@ task(':publish', function(done: (err?: any) => void) {
...
+  if (metadata.templateUrl) {
+    const fullResourcePath = componentResources.get(metadata.templateUrl);
+    metadata.template = readFileSync(fullResourcePath, 'utf-8');
+    delete metadata.templateUrl;
+  }
+
+  // Convert `styleUrls` to `styles`
+  if (metadata.styleUrls && metadata.styleUrls.length) {
+    metadata.styles = [];
+    for (const styleUrl of metadata.styleUrls) {
```
crisbeto: This one could be rewritten with `map`.

Issue #1512, opened 2016-10-16
UNMET PEER DEPENDENCY on installing @angular/material


drasko:
#### Bug, feature request, or proposal:
Can not install Material 2

#### What is the current behavior?
```
drasko@Lenin:~/mainflux/dashflux$ npm install --save @angular/material
dashflux@0.0.0 /home/drasko/mainflux/dashflux
├── UNMET PEER DEPENDENCY @angular/common@2.0.2
├── UNMET PEER DEPENDENCY @angular/compiler@2.0.2
├── UNMET PEER DEPENDENCY @angular/core@2.0.2
├─┬ @angular/material@2.0.0-alpha.9-3 
│ └── @types/hammerjs@2.0.33 
└── UNMET PEER DEPENDENCY @angular/platform-browser@2.0.2

npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.0.0 (node_modules/chokidar/node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.0.14: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
npm WARN @angular/compiler-cli@0.6.4 requires a peer of @angular/platform-server@2.0.2 but none was installed.
npm WARN @angular/platform-server@2.1.0 requires a peer of @angular/core@2.1.0 but none was installed.
npm WARN @angular/platform-server@2.1.0 requires a peer of @angular/common@2.1.0 but none was installed.
npm WARN @angular/platform-server@2.1.0 requires a peer of @angular/compiler@2.1.0 but none was installed.
npm WARN @angular/platform-server@2.1.0 requires a peer of @angular/platform-browser@2.1.0 but none was installed.
drasko@Lenin:~/mainflux/dashflux$
```

#### Which versions of Angular, Material, OS, browsers are affected?
```
drasko@Lenin:~/mainflux/dashflux$ uname -a
Linux Lenin 4.6.0-1-amd64 #1 SMP Debian 4.6.4-1 (2016-07-18) x86_64 GNU/Linux
drasko@Lenin:~/mainflux/dashflux$ node --version
v6.8.1
drasko@Lenin:~/mainflux/dashflux$ ng --version
Looks like you have a different program called watchman, falling back to NodeWatcher.
Visit http://ember-cli.com/user-guide/#watchman for more info.
angular-cli: 1.0.0-beta.17
node: 6.8.1
[...]

PR #3752, opened 2017-03-24
chore: bump to angular 4 final

crisbeto:
Bumps the version to Angular 4 final. The release is pretty safe and doesn't need any code changes.


paulwalker:
Does the @angular/material version needs to be bumped and released as well?

jelbourn:
We're working on cutting a release as soon as possible.

basst314:
Hi @jelbourn, is this a matter of hours/days or weeks?
Just that we know what to expect.
thanks!

jelbourn:
Hopefully just a few days- we've done some work to update our package structure to match `angular/angular`, but we're still tuning it so that it will better support tree-shaking as it exists in Webpack.

JordyBaylac:
Hi @jelbourn I came across the [UNMET PEER DEPENDENCY at #1512](https://github.com/angular/material2/issues/1512) and I would like to know if there any advice, maybe use angular 2.0 instead of 4.0? or should I wait the next release?

PS: Keep doing great!

PR #3761, opened 2017-03-24
fix(slide-toggle): remove host element margin

DevVersion:
* Angular Material components should not have any outer-margin on component host elements (Similar as for the checkbox component)


jelbourn:
This will likely cause some screenshot failures for google apps... running a preemptive presubmit.

PR #3765, opened 2017-03-24
fix(select): allow custom aria-label

### LINKED ISSUE #3762: mat-select should enable specifying aria-label without specifying placeholder
Bug

User can specify [aria-label] attribute without specifying [placeholder]

[placeholder] overrides the aria-label attribute

http://plnkr.co/edit/yPrSiGr1mx4c1vW4sKsO?p=preview

DevVersion:
* Adds support for custom `aria-label` attributes on the select host element.
* Also fixes that the `aria-labelledby` attribute is overwritten by the placeholder `aria-label`.

Fixes #3762.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1085,6 +1085,21 @@ describe('MdSelect', () => {
         expect(select.getAttribute('aria-label')).toEqual('Food');
       });
 
+      it('should support setting a custom aria-label', () => {
+        fixture.componentInstance.ariaLabel = 'Custom Label';
+        fixture.detectChanges();
+
+        expect(select.getAttribute('aria-label')).toEqual('Custom Label');
+      });
+
+      it('should not set a aria-label if aria-labelledby is specified', () => {
+        fixture.componentInstance.ariaLabelledby = 'myLabelId';
+        fixture.detectChanges();
+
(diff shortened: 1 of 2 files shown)


src/lib/select/select.ts
```diff
@@ -747,6 +754,13 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
         'visible' : 'hidden';
   }
 
+  /** Returns the aria-label of the select component. */
+  get _ariaLabel(): string {
+    // If a ariaLabelledby value has been set, the select should not overwrite the
```
kara: Nit: `a ariaLabelledby` -> `an ariaLabelledby`

src/lib/select/select.spec.ts
```diff
@@ -1085,6 +1085,21 @@ describe('MdSelect', () => {
...
       });
 
+      it('should support setting a custom aria-label', () => {
+        fixture.componentInstance.ariaLabel = 'Custom Label';
+        fixture.detectChanges();
+
+        expect(select.getAttribute('aria-label')).toEqual('Custom Label');
+      });
+
+      it('should not set a aria-label if aria-labelledby is specified', () => {
```
kara: Nit: `a aria-label` => `an aria-label`

PR #3774, opened 2017-03-24
fix(dialog): delay focus until animation is done

### LINKED ISSUE #3722: bug(autocomplete): wrong positioning on dialog

crisbeto:
Delays focusing the first tabbable element in a dialog until animation is done. This prevents issues where other components (e.g. autocomplete) may have a wrong position on load, because they measured dialog while it was animating.

Note on lack of tests: This is kind of hard to test, because we don't have much control over how the animations runs.

Fixes #3722.

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -91,9 +91,7 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
       throw new MdDialogContentAlreadyAttachedError();
     }
 
-    let attachResult = this._portalHost.attachComponentPortal(portal);
-    this._trapFocus();
-    return attachResult;
+    return this._portalHost.attachComponentPortal(portal);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


skeize:
After upgrading from beta 2 to 3 setting focus after init started failing when inputs are inside dialog.  As a workaround for now, I'm using a 500ms timeout to account for the dialog animation to complete.. but it kinda feels dirty:

`    ngAfterViewInit() {
        setTimeout(() => {
            if (this.firstInput) {
                this.firstInput.focus();
            }
        }, 500);
    }`

PR #3776, opened 2017-03-25
build: package theming files into release output

jelbourn:
This will be a breaking change for the next release. 

I'm holding off on updating the theming guide until we're ready to release (since it updates on GitHub immediately).


peluprvi:
I think after this PR, `core/theming/prebuilt` path need to be changed to `prebuilt-themes` in [some of those files](https://github.com/angular/material2/search?utf8=%E2%9C%93&q=%22core%2Ftheming%2Fprebuilt%22&type=).

PR #3781, opened 2017-03-25
feat(theming): log a warning if core theme isn't loaded

### LINKED ISSUE #2828: Log a warning when a theme is not loaded

crisbeto:
This is a re-submit of #2846 which has the same logic, but does the check in the `CompatibilityModule`, instead of in a separate one.

* Checks the user's loaded stylesheets and logs a warning if the Material core theme isn't loaded.
* Fixes a wrong `typeof` check when determining the doctype.

Fixes #2828.

Note: I originally went with looping through the `document.styleSheets` to check whether the selector is defined, however I had to switch back to `getComputedStyle`, because browsers don't expose the `document.styleSheets`, if the CSS file is being loaded from another domain. This would've caused the warning to be logged if the user loads over a CDN.

### DIFF
--- a/src/lib/core/compatibility/compatibility.ts
+++ b/src/lib/core/compatibility/compatibility.ts
@@ -9,6 +9,8 @@ import {
 } from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
 
+/** Flag for whether we've checked that the theme is loaded. */
+let hasCheckedThemePresence = false;
 
 export const MATERIAL_COMPATIBILITY_MODE = new OpaqueToken('md-compatibility-mode');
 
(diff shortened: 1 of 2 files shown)


src/lib/core/compatibility/compatibility.ts
```diff
@@ -170,14 +172,41 @@ export class CompatibilityModule {
...
+    if (isDevMode() && this._document && !this._document.doctype) {
       console.warn(
         'Current document does not have a doctype. This may cause ' +
         'some Angular Material components not to behave as expected.'
       );
     }
   }
+
+  private _checkTheme(): void {
+    if (hasCheckedThemePresence || !this._document || !isDevMode()) {
```
jelbourn: Should we move the `isDevMode` check to the constructor now that we're doing multiple checks in the compatibility module?

We can also just have one `hasDoneGlobalChecks` property for all checks instead of `hasCheckedThemePresence`.

src/lib/core/compatibility/compatibility.ts
```diff
@@ -170,14 +172,41 @@ export class CompatibilityModule {
...
+      return;
+    }
+
+    let testElement = this._document.createElement('div');
```
jelbourn: If you make the `testElement` `position: absolute`, you should remove all possibility of adding it causing a reflow.

src/lib/core/compatibility/compatibility.ts
```diff
@@ -170,14 +172,41 @@ export class CompatibilityModule {
...
+
+    testElement.classList.add('mat-theme-loaded-marker');
+    this._document.body.appendChild(testElement);
+
+    if (getComputedStyle(testElement).display !== 'none') {
+      console.warn(
+        'Could not find Angular Material core theme. Most Material ' +
+        'components may not work as expected. For more info refer ' +
+        'to the theming guide: https://github.com/angular/material2/blob/master/guides/theming.md'
```
jelbourn: Can we link to material.angular.io instead?

src/lib/core/compatibility/compatibility.ts
crisbeto: The test element is `display: none` already.

crisbeto:

PR #3739, opened 2017-03-23
feat(slide-toggle): add ripple focus indicator

DevVersion:
* Introduces a focus indiactor using the persistent ripples.

Fixes #3738

PR #3783, opened 2017-03-25
refactor(focus-origin-monitor): rename unmonitor to stopMonitoring

DevVersion:
* Renames the `unmonitor` function to `stopMonitoring` (As requested in https://github.com/angular/material2/pull/3739#discussion_r108026958)
* Removes unnecessary code that keeps track of the `FocusOriginMonitor` subscription. After calling `stopMonitoring` the observable will complete.

### DIFF
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -125,7 +125,7 @@ export class MdButton implements OnDestroy {
   }
 
   ngOnDestroy() {
-    this._focusOriginMonitor.unmonitor(this._elementRef.nativeElement);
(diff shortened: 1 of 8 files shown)


mmalerba:
LGTM

I also wonder if we should rename the whole class to just `FocusMonitor`. It can theoretically be useful even if you don't care about the origin, just to monitor focus on a whole DOM subtree rather than a single element. @jelbourn thoughts?

DevVersion:
@mmalerba Should we do this in a follow-up PR if we change it?

mmalerba:
yeah, feel free to mark this one merge ready

PR #3789, opened 2017-03-26
build: run closure-compiler

### LINKED ISSUE #1206: ensure that all components with with Closure compiler advanced_optimizations
We should run the demo app through closure compiler and fix the issues we find.

DevVersion:
* Adds a script that runs the closure-compiler against the devapp / demo-app.
* The script will run on the CI as a new mode. Ensuring that all our components are closure-compatible.

**Note**: I'm using a Bash script because it is way cleaner than executing everything using `child_process`. Also it is aligned with the proof-of-concept [from Alex](https://github.com/alexeagle/closure-compiler-angular-bundling). Plus it's just way easier rebuilding RxJS or the demo-app without needing X gulp tasks.

Fixes #1206.


jelbourn:
@alexeagle could you give this a quick sanity pass?

package.json
```diff
@@ -60,11 +60,12 @@
     "conventional-changelog": "^1.1.0",
     "dgeni": "^0.4.7",
     "dgeni-packages": "^0.16.5",
+    "firebase": "^3.7.2",
     "firebase-admin": "^4.1.2",
     "firebase-tools": "^2.2.1",
-    "firebase": "^3.7.2",
     "fs-extra": "^2.0.0",
     "glob": "^7.1.1",
+    "google-closure-compiler": "^20170218.0.0",
```
alexeagle: I assume you use npm shrinkwrap to avoid surprises when the version moves?

scripts/closure-compiler/build-devapp-bundle.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+# Script that bundles the dev-app using the Google Closure compiler.
+# This is script is used to verify closure-compatiblity of all Material components.
```
alexeagle: it only does the first bits of closure verification of course, we don't know if the property renaming did the right thing. can we also run some of the integration tests on the result (or does this PR do that?)

scripts/closure-compiler/build-devapp-bundle.sh
```diff
@@ -0,0 +1,91 @@
...
+# Create a list of all RxJS source files.
+rxjsSourceFiles=$(find dist/packages/rxjs -name '*.js');
+
+# Due a Closure Compiler issue https://github.com/google/closure-compiler/issues/2247
+# we need to add exports to the different RxJS ES2015 files.
+for i in $rxjsSourceFiles; do
+    echo "export var __CLOSURE_WORKAROUND__" >> $i
+done
+
+OPTS=(
```
alexeagle: now that we use a flagfile maybe it would be better to formulate this with 
```
cat > flagfile << EOF
...
EOF
```

instead of a bash array?

scripts/closure-compiler/build-devapp-bundle.sh
```diff
@@ -0,0 +1,91 @@
...
+# Re-compile RxJS sources into ES2015. Otherwise closure compiler can't parse it properly.
+$(npm bin)/ngc -p scripts/closure-compiler/tsconfig-rxjs.json
+
+# Create a list of all RxJS source files.
+rxjsSourceFiles=$(find dist/packages/rxjs -name '*.js');
+
+# Due a Closure Compiler issue https://github.com/google/closure-compiler/issues/2247
+# we need to add exports to the different RxJS ES2015 files.
+for i in $rxjsSourceFiles; do
+    echo "export var __CLOSURE_WORKAROUND__" >> $i
```
alexeagle: yeah I guess I didn't encode the workaround into angular or my example app because we didn't happen to load one of the rx operators?

scripts/closure-compiler/tsconfig-rxjs.json
```diff
@@ -0,0 +1,16 @@
+{
```
alexeagle: maybe extend from the tsconfig.json in rxjs sources? Shouldn't you share most of their options?

scripts/closure-compiler/tsconfig-rxjs.json
```diff
@@ -0,0 +1,16 @@
...
+  "compilerOptions": {
+    "module": "es2015",
+    "outDir": "../../dist/packages/rxjs",
+    "target": "es2015",
```
alexeagle: actually the target should be es5 - that's what they are likely to ship

package.json
DevVersion: We are only using a `yarn.lock`. But yeah might be a good idea to introduce such a file. @jelbourn for confirmation?

scripts/closure-compiler/build-devapp-bundle.sh
DevVersion: This PR just bundles our demo-app that uses Angular Material. So not really sure if this is enough for the variable renaming. Can be a follow-up?

scripts/closure-compiler/build-devapp-bundle.sh
DevVersion: Yeah, we are using a few and those couldn't be imported. Took me a while to find that issue :)

scripts/closure-compiler/tsconfig-rxjs.json
DevVersion: Yeah would be nice, but unfortunately RxJS doesn't ship their `tsconfig.json`. We only have the plain sources from the NPM package. See [here](https://unpkg.com/rxjs@5.2.0/src/)

scripts/closure-compiler/tsconfig-rxjs.json
DevVersion: Sounds good.

scripts/closure-compiler/build-devapp-bundle.sh
DevVersion: Not really sure. I just tried it and it didn't look very pretty inside of the Bash script. I might be doing something wrong? What advantage does it have?

package.json
jelbourn: I haven't experienced issues with this such that I have an opinion one way or the other. Is it reasonable to leave it until we run into a problem?

scripts/closure-compiler/build-devapp-bundle.sh
jelbourn: I'd like to run Closure Compiler over the e2e-app and run the e2e tests over that in order to ensure that everything is working; I think this PR is a good first step to ensure that that the code at least _passes_.

PR #3791, opened 2017-03-26
build: tsconfig for editor warnings

DevVersion:
Currently when using a code editor that uses the TypeScript language service (Webstorm / VS Code / Atom / Sublime) there will be TypeScript errors in the `.spec` files.

This is because the editor finds the `tsconfig.json` file that doesn't match the spec files. Therefore it complains about the `experimentalDecorators`.

In `angular/angular` there is a global `tsconfig` file that matches everything and is therefore fixing the editor warnings. This file is also used to compile all `.spec` files for the whole project.

Adding such a `tsconfig.json` file will fix all errors, and in the future we can use it for compiling the specs as well (more consistency with `angular/angular`)


src/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+// TypeScript config file that matches all source files in the project. This file is read by 
+// IDEs and ensures that `experimentalDecorator` warnings are not showing up.
```
jelbourn: While we're doing this, could you add a top-level comment to all of the tsconfig files that say what they're used for?

DevVersion: Done.

PR #3792, opened 2017-03-26
feat: remove hard dependency on @angular/http

### LINKED ISSUE #2616: Can you move HttpModule in MdIconModule to component scope (or don't import it)?
Bug

The feature modules that import MaterialModule (import MdIconModule) will create separated http provider and override the application-wide customized Http.

The following link shows the problem.
http://plnkr.co/edit/hDLOuvRPWxPIY3TwX4GB?p=preview

Cannot use application-wide customized Http in feature module that import MaterialModule

@angular@2.4.2
@angular/material@2.0.0-beta.1
Google Chrome Version 57.0.2970.0 (Official Build) dev (64-bit)
Window 10

crisbeto:
* Makes `@angular/http` an optional dependency since it is only being used by the icon module. An error will be thrown only if the user runs into something that requires the `HttpModule`.
* Moves the various icon error classes into `icon-errors.ts`.
* Moves the icon registry provider next to the registry itself, instead of in the same file as the icon directive.

Fixes #2616.

### DIFF
--- a/src/lib/icon/fake-svgs.ts
+++ b/src/lib/icon/fake-svgs.ts
@@ -1,6 +1,4 @@
-import {
-  Response,
-  ResponseOptions} from '@angular/http';
+import {Response, ResponseOptions} from '@angular/http';
 
 /**
  * Fake URLs and associated SVG documents used by tests.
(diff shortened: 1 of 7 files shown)


jelbourn:
Has presubmit failures in Google where people were depending on this to bring in HttpModule

mmalerba:
I fixed the deps in google, should be able to merge now, needs rebase though

crisbeto:
Rebased @mmalerba.

PR #3794, opened 2017-03-26
build: fix serve task

DevVersion:
* Currently when serving the dev-app or e2e-app and changes are recognized, the library task will run. This task clears the whole `dist/` folder and rebuilds then.

* This is problematic because sometimes only the library will be rebuilt and for example the `dev-app` won't rebuild. Once the task finished, the `dev-app` is removed and the serve task will crash.

### DIFF
--- a/tools/gulp/tasks/library.ts
+++ b/tools/gulp/tasks/library.ts
@@ -21,8 +21,9 @@ const bundlesDir = DIST_BUNDLES;
 
 const esmMainFile = join(materialDir, 'index.js');
 
+task('library', sequenceTask('clean', 'library:build'));
+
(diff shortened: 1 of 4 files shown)


tools/gulp/tasks/library.ts
```diff
@@ -21,8 +21,9 @@ const bundlesDir = DIST_BUNDLES;
 
 const esmMainFile = join(materialDir, 'index.js');
 
+task('library', sequenceTask('clean', 'library:build'));
```
jelbourn: How about "library:clean-build" ?

DevVersion: Really like it. Changed it!

Issue #1430, opened 2016-10-05
Overlay supports repositioning upon resize / scroll


jelbourn:
As of now, overlays only position themselves once (and when told to reposition through the `OverlayRef`).

This issue tracks overlays being able to automatically reposition themselves when scrolling or resize.

fxck:
Any ETA on this?

fxck:
@jelbourn almost all components that use overlay look broken because this doesn't work.. it seems like a pretty high priority thing to me. Do you want to work on this internally, or I can try and make a PR?

jelbourn:
@fxck This is something we plan to work on- I have a specific approach in mind.

fxck:
and do you have it written down somewhere? @jelbourn

fxck:
@jelbourn this is such a pain in the ass.. can you please share what you have in mind so I can give it a shot?

Is it what's in https://docs.google.com/document/d/1jSMjkKSdzbQa2cuCeSFpzFAwDHr7xY4eqsXkfF6MeqA/preview or something else entirely?

jelbourn:
It's mostly what's in the doc. 

@crisbeto has been exploring the "disable scrolling" option in #1971 and @andrewseguin is already working on the repositioning.

mycarrysun:
Stumbled upon this issue because I am trying to figure out why the MdDialog is position:static (changing to absolute fixes it...kind of). The overlay is on top of the dialog and the dialog is full width of the page (no width is declared in my config). Is this broken or do I need to do something to get it to work? I can fix it by adding display:flex and height: 100% to the .md-global-overlay-wrapper, but not sure if that is the correct way to go about it.

crisbeto:
Try clearing your `node_modules` @mycarrysun. There have been a few similar issues after the latest release.

mycarrysun:
Hmm strange, it worked though! Thanks @crisbeto

k3nsei:
@jelbourn for example if menu is open indeed scroll is disabled. But only if you using mouse roller. But there are other options to scroll. For example by pressing up, down, pageUp, pageDown, home and end keys.

jelbourn:
#2703 adds this for tooltip. Once that is in, subsequent PRs can add this to menu and select.

aldo-roman:
Hi @jelbourn, any update on this issue? It's been 2 months now. Thanks!

jelbourn:
This issue is now replaced by #4093

marcusreese:
[...]

PR #3808, opened 2017-03-27
fix(select): reposition panel on scroll

crisbeto:
Repositions the select panel if the user scrolls while it is open.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -321,7 +321,7 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
 
   constructor(private _element: ElementRef, private _renderer: Renderer2,
               private _viewportRuler: ViewportRuler, private _changeDetectorRef: ChangeDetectorRef,
-              private _scrollDispatcher: ScrollDispatcher, @Optional() private _dir: Dir, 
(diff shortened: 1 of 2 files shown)


armstrjare:
Curious as to why the behaviour for this and #3745 (and the equivalent for tooltips) isn't moved into the `ConnectedOverlayDirective` itself? Presumably any users of that would want the connection to persist through scrolling?

That would solve #1430

src/lib/select/select.ts
```diff
@@ -131,6 +132,9 @@ export class MdSelect implements AfterContentInit, ControlValueAccessor, OnDestr
   /** Subscription to tab events while overlay is focused. */
   private _tabSubscription: Subscription;
 
+  /** Subscription to global scrolled events while the select is open. */
+  private _scrolledSubscription: Subscription;
```
kara: Nit: Mind changing this to `_scrollSubscription`? For some reason, I keep tripping over it as is because it sounds like past tense.

kara:

crisbeto:
Renamed and rebased @kara.

kara:
@crisbeto  Can you rebase?

crisbeto:
Rebased.

PR #3821, opened 2017-03-28
fix(slider,slide-toggle): wrong cursor

### LINKED ISSUE #3778: Slider with thumb cursor issue
#### Bug:
Currently material slider with thumb label which has current slider value the mouse cursor changes to "select text" cursor.

When dragging while holding slider's thumb cursor to be same.

https://material.angular.io/components/component/slider and select option "show thumb label". Then click on slider and click and drag while pointing to slider thumb label with changing current value.

Windows 10, Google Chrome 56

crisbeto:
Uses the `grab` cursor when hovering over an active slider's thumb and thumb labels, as well as on active slide toggles.

Fixes #3778.

### DIFF
--- a/src/lib/slider/slider.scss
+++ b/src/lib/slider/slider.scss
@@ -1,4 +1,5 @@
 @import '../core/style/variables';
+@import '../core/style/vendor-prefixes';
 
 
 // This refers to the thickness of the slider. On a horizontal slider this is the height, on a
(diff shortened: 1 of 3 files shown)


mmalerba:
should we do grabbing during drag as well?

crisbeto:
This will only change the cursor while the user is over the thumb label. Are you referring to adding it in general on the `mat-slider-sliding` class?

mmalerba:
@crisbeto there's both a `grab`  and a `grabbing` (https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) would make sense to add `grabbing` once the user clicks and starts dragging

crisbeto:
Nice, I didn't know about the grabbing cursor. I've also added it to the slide toggle. CC @DevVersion

src/lib/slider/slider.scss
```diff
@@ -97,7 +98,19 @@ $mat-slider-focus-ring-size: 30px !default;
   }
 }
 
+%_mat-slider-cursor {
+  .mat-slider:not(.mat-slider-disabled) & {
+    @include cursor-grab;
+  }
+
+  .mat-slider:not(.mat-slider-disabled) &:active {
```
mmalerba:
```scss
.mat-slider:not(.mat-slider-disabled) &:active,
.mat-slider-sliding:not(.mat-slider-disabled) & {
```
This way it works when I press somewhere on the track and start sliding too

PR #3822, opened 2017-03-28
style(button): correct focus overlay for raised buttons 

### LINKED ISSUE #3820: Raised buttons with colors are not following spec on focus
Bug

Text color of raised buttons with color does not change on focus
Background-color of raised buttons with color changes on focus

Text color of raised buttons with color changes on focus
Background-color of raised buttons with color does not on focus

Here is a plunker that shows the issue
http://plnkr.co/edit/36A2DKgGGP4PtFfBrBtO?p=preview

It should follow [the spec](https://material.io/guidelines/components/buttons.html#buttons-raised-buttons)

Latest master has this issue

willshowell:
The themed overlay was being added to all buttons, but should only be applied to flat or flat icon buttons. Raised buttons should use the default `rgba(black, 0.12)` focus overlay color defined in `button.scss`

Fixes #3820

### DIFF
--- a/src/lib/button/_button-theme.scss
+++ b/src/lib/button/_button-theme.scss
@@ -74,16 +74,10 @@
   $background: map-get($theme, background);
   $foreground: map-get($theme, foreground);
 
-  .mat-button, .mat-icon-button, .mat-raised-button, .mat-fab, .mat-mini-fab {
-    // Apply color to focus overlay.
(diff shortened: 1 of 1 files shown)


willshowell:
**Before**
before

**After**
after

willshowell:
Looking forward to having this assigned/reviewed!

willshowell:
@jelbourn any chance of getting this assigned?

PR #3824, opened 2017-03-28
fix(datepicker): calendar should update when input changes

mmalerba:


src/lib/datepicker/datepicker.ts
```diff
@@ -116,7 +111,7 @@ export class MdDatepicker implements OnDestroy {
 
   constructor(private _dialog: MdDialog, private _overlay: Overlay,
               private _viewContainerRef: ViewContainerRef, private _locale: CalendarLocale,
-              @Optional() private _dir: Dir) {}
+              private _changeDetectorRef: ChangeDetectorRef, @Optional() private _dir: Dir) {}
```
kara: Are you using this anywhere? Couldn't find it.

src/lib/datepicker/datepicker.ts
```diff
@@ -134,6 +139,7 @@ export class MdDatepicker implements OnDestroy {
       throw new MdError('An MdDatepicker can only be associated with a single input.');
     }
     this._datepickerInput = input;
+    this._datepickerInput.valueChange.subscribe((value: SimpleDate) => this._selected = value);
```
kara: Are you unsubscribing anywhere?

src/lib/datepicker/datepicker-input.ts
```diff
@@ -56,14 +59,17 @@ export class MdDatepickerInput implements AfterContentInit, ControlValueAccessor
...
   set value(value: SimpleDate) {
-    this._value = this._locale.parseDate(value);
-    const stringValue = this._value == null ? '' : this._locale.formatDate(this._value);
-    this._renderer.setElementProperty(this._elementRef.nativeElement, 'value', stringValue);
+    let date = this._locale.parseDate(value);
+    let oldDate = this.value;
+    this._renderer.setElementProperty(this._elementRef.nativeElement, 'value',
+        date ? this._locale.formatDate(date) : '');
+    if (!SimpleDate.equals(oldDate, date)) {
+      this._valueChangeEmitter.emit(date);
```
kara: If you had something like this: 

```html
<input [value]="val" (valueChange)="doSomething()" [mdDatepicker]="dp">
```

Would setting the value programmatically fire the `doSomething()` method? It looks like valueChange is emitted in both directions.

src/lib/datepicker/datepicker.ts
mmalerba: nope, must've been leftover from previous iteration

src/lib/datepicker/datepicker-input.ts
mmalerba: Yes, isn't that what we want? Otherwise if I set the value programmatically the calendar won't be updated to reflect the change in the input

kara: Yeah, I see how it's solving that problem. Just concerned that users will be confused because typically when you have a model to view change, you don't expect the view to model handler to fire as well.

src/lib/datepicker/datepicker.ts
```diff
@@ -123,6 +121,19 @@ export class MdDatepicker implements OnDestroy {
     if (this._popupRef) {
       this._popupRef.dispose();
     }
+    if (this._inputSubscription) {
+
```
kara: Did you mean to unsubscribe here?

mmalerba: sorry... i'm an idiot ><

src/lib/datepicker/datepicker-input.ts
mmalerba: Yeah I admit its a little weird and I had to add that equals check to stop it from infinite looping, maybe we can talk about if there's a better way

src/lib/datepicker/datepicker.ts
kara: hehe, I was a bit confused

Issue #3550, opened 2017-03-10
show available prebuilt themes in docs


lestard:
#### Bug, feature request, or proposal:
feature request documentation

#### What is the expected behavior?
In the [docs](https://material.angular.io/guide/theming) users should see a list of available prebuilt themes and a link to a sample page of each theme.

#### What is the current behavior?

No information of what themes are available and how they look like.

#### What are the steps to reproduce?

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: https://goo.gl/DlHd6U

#### What is the use-case or motivation for changing an existing behavior?

I like to use this great library but the documentation could be improved to make the start easier.

M-a-c:
@jelbourn
Issue question:
Where should these examples be created?   
/src/lib/examples/theme-NameofTheme-overview ?
Just one and then reference the other defaults, or all 4 themes?

Extra question:
Also are there going to be more default themes? / a theme builder ?

jelbourn:
We should list all of the prebuilt themes in the theming guide.

We will add more prebuilt themes before final.

We do want to add a theme builder to the docs site eventually, but that's more of a nice-to-have.

DevVersion:
@M-a-c Are you looking into this?

M-a-c:
Yeah I was going to tackle it Monday if thats ok. I don't have a pr for it right now.
I think this Add theming to docs should be re-added, for when there are more themes, before going gold.

DevVersion:
Sounds good! Thanks.

M-a-c:
Is sample link to each theme necessary ?

If its possible can someone open a P4 on the 'theme builder' and describe how it would work? I would love to work on something like that.

DevVersion:
I think for now we only want to list all prebuilt theme paths in the theming guide.

We could do that with a Dgeni processor or just with `gulp transform`.

M-a-c:
Reading up on Dgeni now, sorry for the confusion. I had no idea this existed. I was wondering how exactly this sorta magic happened. I'll dig for an example to figure this out.

jelbourn:
I'm not sure this is something that needs to be done with tooling; the prebuilt themes will change infrequently enough that i'm okay with just manually listing them in the guide.

M-a-c:
[...]

PR #3826, opened 2017-03-28
docs(theming-guide): list available themes

M-a-c:
#3550 
Added the 4 available themes to the theming guide.
Detailed that more updates to come as development continues.


cgatian:
Is it necessary to mention there will be additional themes available in the future?  I think it would put more of a burden on the Material team to maintain those themes in the future.  The custom theming is pretty straightforward, and I don't see much value in adding additional themes to what's available.

PR #3829, opened 2017-03-29
fix(button): ripple color for raised buttons

### LINKED ISSUE #2901: (ripple): make sure ripples are on spec - speed & color

DevVersion:
* Fixes the incorrect ripple color for raised buttons. Currently the ripple color is always based on black. This is incorrect on palettes like Indigo.

**Note**: The `opacity` couldn't be changed in the `mat-color` mixin because it prioritized the palettes opacity. If developers explicitly specify a opacity, then this should be used. 

> Confirmed that all components don't run into such issue (so we could safely change it without seeing differences).

<p><details>

  <summary>Test snippet that logs if opacity has been overwritten (for confirmation)</summary>

screen shot 2017-03-29 at 15 36 27

</details></p>

Fixes #2901

### DIFF
--- a/src/lib/button/_button-theme.scss
+++ b/src/lib/button/_button-theme.scss
@@ -23,21 +23,21 @@
   }
 }
 
-@mixin _mat-button-ripple-color($theme) {
+@mixin _mat-button-ripple-color($theme, $hue: default) {
   $primary: map-get($theme, primary);
   $accent: map-get($theme, accent);
   $warn: map-get($theme, warn);
 
   &.mat-primary .mat-ripple-element {
-    background-color: mat-color($primary, 0.26);
+    background-color: mat-color($primary, $hue, 0.26);
   }
 
   &.mat-accent .mat-ripple-element {
-    background-color: mat-color($accent, 0.26);
+    background-color: mat-color($accent, $hue, 0.26);
   }
 
   &.mat-warn .mat-ripple-element {
-    background-color: mat-color($warn, 0.26);
+    background-color: mat-color($warn, $hue, 0.26);
   }
 }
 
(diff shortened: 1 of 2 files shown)


mmalerba:
Code looks good, couple style questions:
- Does the spec say the ripples for the raised button should be .26 opacity? I feel like it would look better if it were a little lower
- Could we also change the flat colored buttons to have colored ripples instead of black (since the pattern seems to be ripple color matches the text color)?

DevVersion:
Good questions, I just played with them:
* They don't really say anything about it. If we go down with it, it would be very difficult to notice onto darker palettes. Add a opacity parameter so we can distinguish between raised and flat buttons. 
* I just tried it because it is a great idea. 

Now for flat buttons, opacity: `0.1` and for raised buttons `0.2`. Compared to Polymer they have it way stronger, but it also looks kind of weird.

mmalerba:
Awesome, looks great! 1 more nit: now the icon button ripples look too transparent, since they don't have a focus style

DevVersion:
@mmalerba Good catch. Fixed that. I had to divide the selectors up, but I added comments.

src/lib/button/_button-theme.scss
```diff
@@ -97,11 +94,32 @@
 
     @include _mat-button-theme-color($theme, 'color', default-contrast);
     @include _mat-button-theme-color($theme, 'background-color');
+
+    // Add ripple effect with contrast color to buttons that don't have a focus overlay.
+    @include _mat-button-ripple-color($theme, default-contrast);
+  }
+  
+  // Add ripple effect with default color to the flat button, which doesn't have a focus overlay.
```
mmalerba: which has a focus overlay?

DevVersion: Fixed that comment. Can you re-check again? :)

mmalerba:
LGTM

PR #3836, opened 2017-03-29
refactor: move away from deprecated apis

crisbeto:
Moves away from the APIs that were deprecated in Angular 4. They include:
* Switching from using `OpaqueToken` to `InjectionToken`.
* Moving from `Renderer` to `Renderer2`.
* Changing the `Injector.get(thing: any)` usages to use the new signature `Injector.get<T>(thing: T)`.

### DIFF
--- a/src/lib/button-toggle/button-toggle.spec.ts
+++ b/src/lib/button-toggle/button-toggle.spec.ts
@@ -5,7 +5,7 @@ import {
   ComponentFixture,
   TestBed,
 } from '@angular/core/testing';
-import {NgControl, FormsModule, ReactiveFormsModule, FormControl} from '@angular/forms';
+import {NgModel, FormsModule, ReactiveFormsModule, FormControl} from '@angular/forms';
 import {Component, DebugElement} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {
@@ -56,7 +56,7 @@ describe('MdButtonToggle', () => {
 
       groupDebugElement = fixture.debugElement.query(By.directive(MdButtonToggleGroup));
       groupNativeElement = groupDebugElement.nativeElement;
-      groupInstance = groupDebugElement.injector.get(MdButtonToggleGroup);
+      groupInstance = groupDebugElement.injector.get<MdButtonToggleGroup>(MdButtonToggleGroup);
 
       buttonToggleDebugElements = fixture.debugElement.queryAll(By.directive(MdButtonToggle));
 
@@ -215,7 +215,7 @@ describe('MdButtonToggle', () => {
     let groupInstance: MdButtonToggleGroup;
     let buttonToggleInstances: MdButtonToggle[];
     let testComponent: ButtonToggleGroupWithNgModel;
-    let groupNgControl: NgControl;
+    let groupNgModel: NgModel;
 
     beforeEach(async(() => {
       fixture = TestBed.createComponent(ButtonToggleGroupWithNgModel);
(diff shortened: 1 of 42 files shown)


crisbeto:
Note: The Closure failure will be fixed by https://github.com/angular/material2/pull/3835. Fixing the AoT and linting errors now.

src/lib/button-toggle/button-toggle.spec.ts
```diff
@@ -5,7 +5,7 @@ import {
   ComponentFixture,
   TestBed,
 } from '@angular/core/testing';
-import {NgControl, FormsModule, ReactiveFormsModule, FormControl} from '@angular/forms';
+import {NgModel, FormsModule, ReactiveFormsModule, FormControl} from '@angular/forms';
```
jelbourn: @kara is `NgModel` the right type here?

src/lib/button/button.ts
```diff
@@ -141,13 +141,17 @@ export class MdButton implements OnDestroy {
 
   _setElementColor(color: string, isAdd: boolean) {
     if (color != null && color != '') {
-      this._renderer.setElementClass(this._getHostElement(), `mat-${color}`, isAdd);
+      if (isAdd) {
```
jelbourn: I think you can kill this entire function now and just do the null check in `_updateColor` and call `addClass` / `removeClass` directly there.

src/lib/core/ripple/ripple.ts
```diff
@@ -7,21 +7,22 @@ import {
...
-/** OpaqueToken that can be used to specify the global ripple options. */
-export const MD_RIPPLE_GLOBAL_OPTIONS = new OpaqueToken('md-ripple-global-options');
-
 export type RippleGlobalOptions = {
   disabled?: boolean;
   baseSpeedFactor?: number;
 };
 
+/** Injection token that can be used to specify the global ripple options. */
+export const MD_RIPPLE_GLOBAL_OPTIONS = new InjectionToken<RippleGlobalOptions>(
```
jelbourn: Generally prefer line breaking at the higher syntactic level, e.g.:
```ts
export const MD_RIPPLE_GLOBAL_OPTIONS = 
    new InjectionToken<RippleGlobalOptions>('md-ripple-global-options');
```
(at the assignment)

src/lib/button/button.ts
crisbeto: This one is used in a couple of other places, as well as inside the `button.html` so it may be more convenient to keep it. Otherwise we'd need to do `this._elementRef.nativeElement` all over the place.

crisbeto:
Rebased and addressed some of the feedback @jelbourn.

jelbourn:
This is a breaking change because `Renderer` was part of the API for focus origin monitor.

crisbeto:
Rebased. Should I remove the `presubmit failures` tag @kara?

mmalerba:
It looks like the presubmit failures are because one of the teams in google is injecting the `FocusOriginMonitor` and the `Renderer` for one of their components and then calling `focusVia` with the `Renderer`. Do we want to make the `FocusOriginMonitor` accept `Renderer2 | Renderer`? or should we just refactor that project's code?

crisbeto:
It may be better to refactor it. We could have a solution that accepts both, but it would add some bloat at some point we'll have to deprecate it anyway.

jelbourn:
@mmalerba if it's not widespread we should just change the team's code

PR #3839, opened 2017-03-29
fix(datepicker): set focus properly when opening datepicker

mmalerba:


mmalerba:
@jelbourn can you take a look at the dialog & overlay stuff going on in `MdDatepicker`. I feel like there may be a simpler way to do things that I'm just not thinking of...

src/lib/datepicker/datepicker.ts
```diff
@@ -188,18 +203,29 @@ export class MdDatepicker implements OnDestroy {
 
   /** Open the calendar as a dialog. */
   private _openAsDialog(): void {
-    this._dialogRef = this._dialog.open(this.calendarTemplate);
+    let config = new MdDialogConfig();
+    config.viewContainerRef = this._viewContainerRef;
+    config.disableFocusTrap = true; // We'll handle focus trapping ourselves.
```
jelbourn: Can we make `disableFocusTrap` internal? I don't want to encourage people to do this and mess up the a11y

src/lib/dialog/dialog-config.ts
```diff
@@ -23,6 +23,9 @@ export class MdDialogConfig {
   /** Whether the user can use escape or clicking outside to close a modal. */
   disableClose?: boolean = false;
 
+  /** Whether to disable the dialog's focus trap. */
+  disableFocusTrap?: boolean = false;
```
jelbourn: This behavior needs a unit test

src/lib/datepicker/datepicker.ts
```diff
@@ -29,21 +30,42 @@ import {MdDatepickerInput} from './datepicker-input';
...
+    this._elementRef.nativeElement.querySelector('.mat-calendar-body').focus();
+  }
+}
+
+
+/** Component responsible for managing the datepicker popup/dialog. */
+@Component({
+  moduleId: module.id,
+  selector: 'md-datepicker, mat-datepicker',
+  template: '',
```
mmalerba: The reason I made it a component is because in order to use it users need to do something like:
```
<input [mdDatepicker]="d'>
<md-datepicker #d></md-datepicker>
```

If I used a directive they had to do:
```
<input [mdDatepicker]="d'>
<md-datepicker #d="mdDatepicker"></md-datepicker>
```

which felt like a lot of "datepicker"

src/lib/datepicker/datepicker.ts
mmalerba: I'll just get rid of it since we plan to switch to using overlay all the time

mmalerba:
comments addressed, PTAL

src/lib/datepicker/datepicker.ts
jelbourn: You can add an `exportAs: '$implicit'` to a component for this. See http://plnkr.co/edit/ckQktBbXIs6unW2wpoo0

mmalerba:
Added some more comments explaining the various components

PR #3853, opened 2017-03-30
fix(datepicker): make touch UI work well on mobile

mmalerba:
depends on #3839


src/lib/datepicker/datepicker-content.scss
```diff
@@ -1,21 +1,45 @@
...
-$md-datepicker-non-touch-calendar-cell-size: 40px;
 $md-datepicker-calendar-padding: 8px;
-
-$md-datepicker-touch-calendar-width:
-    $md-datepicker-touch-calendar-cell-size * 7 + $md-datepicker-calendar-padding * 2;
+$md-datepicker-non-touch-calendar-cell-size: 40px;
 $md-datepicker-non-touch-calendar-width:
     $md-datepicker-non-touch-calendar-cell-size * 7 + $md-datepicker-calendar-padding * 2;
 
+// Note(mmalerba): Ideally the calendar would have a constant aspect ratio, no matter its size,
```
jelbourn: You can omit `Note(mmalerba)` since it's just an explanatory comment

src/lib/datepicker/datepicker.ts
```diff
@@ -49,6 +49,11 @@ let datepickerUid = 0;
   selector: 'md-datepicker-content',
   templateUrl: 'datepicker-content.html',
   styleUrls: ['datepicker-content.css'],
+  host: {
+    '[class.mat-datepicker-content]': 'true',
```
jelbourn: You should be able to just do 
```ts
host: {
  'class': 'mat-datepicker-content'
}
```
And Angular will union it with the other classes and not overwrite (so long as it isn't a `'[class']` binding)

src/lib/datepicker/datepicker.ts
```diff
@@ -49,6 +49,11 @@ let datepickerUid = 0;
...
+    '[class.mat-datepicker-content-touch]': 'datepicker.touchUi',
+    '[class.mat-datepicker-content-non-touch]': '!datepicker.touchUi',
```
jelbourn: Why have two classes instead of a style without the `-touch` class and then another with (where the latter would inherently be higher specificity)?

src/lib/dialog/dialog-container.ts
```diff
@@ -48,13 +48,25 @@ export type MdDialogContainerAnimationState = 'void' | 'enter' | 'exit' | 'exit-
     ])
   ],
   host: {
+    '[class]': 'dialogConfigClasses',
```
jelbourn: This will overwrite all other classes on the dialog-container element

src/lib/dialog/dialog-config.ts
```diff
@@ -32,6 +32,9 @@ export class MdDialogConfig {
   /** Position overrides. */
   position?: DialogPosition;
 
+  /** Classes to add to the dialog container. */
+  containerClass?: string | string[] = '';
```
jelbourn: Do you need a class on the dialog container or on the root overlay element? I'd prefer to do the latter if we have to do this at all, as the dialog-container is meant to be a internal implementation detail

src/lib/dialog/dialog-container.ts
```diff
@@ -48,13 +48,25 @@ export type MdDialogContainerAnimationState = 'void' | 'enter' | 'exit' | 'exit-
...
     '[class.mat-dialog-container]': 'true',
     '[attr.role]': 'dialogConfig?.role',
     '[@slideDialog]': '_state',
     '(@slideDialog.done)': '_onAnimationDone($event)',
   },
 })
 export class MdDialogContainer extends BasePortalHost implements OnDestroy {
+  /** The CSS classes for the container specified in the dialog config. */
+  get dialogConfigClasses(): string {
```
jelbourn: This behavior would need unit tests if kept

src/lib/dialog/dialog-config.ts
mmalerba: I'll undo this since we plan to move away from using dialog, I can hack around it for now

mmalerba:
comments addressed

PR #3854, opened 2017-03-30
fix(datepicker): use input's min & max properites rather than custom …

mmalerba:
…ones on md-datepicker


src/lib/datepicker/datepicker-input.ts
```diff
@@ -61,6 +63,16 @@ export class MdDatepickerInput implements AfterContentInit, ControlValueAccessor
   @Input()
   set matDatepicker(value: MdDatepicker) { this.mdDatepicker = value; }
 
+  @Input()
```
kara: Comment for these @Inputs? For docs site.

PR #3861, opened 2017-03-31
test(select): various test inconsistencies

crisbeto:
* Fixes a select test that occasionally fails to capture the error inside the expect, causing it to get logged without actually being thrown. This seems to be a Zone-related issue since wrapping a few calls in `async` makes it behave correctly.
* Fixes an RTL test that fails if it is made async. This incorporates the changes from #3866 and #3163 and fixes the underlying issue. It appears that the test was failing, because IE does a slight twitch on the select panel since it goes from a pixel-based `min-width` to one with `calc` and percentages to just percentages. I've set the default width to 100% which is the same as the pixel-based one, which appears to make IE happy.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -988,37 +988,50 @@ describe('MdSelect', () => {
         select.style.marginRight = '30px';
       });
 
-      it('should align the trigger and the selected option on the x-axis in ltr', () => {
-        trigger.click();
-        fixture.detectChanges();
+      it('should align the trigger and the selected option on the x-axis in ltr', async(() => {
+        fixture.whenStable().then(() => {
+          fixture.detectChanges();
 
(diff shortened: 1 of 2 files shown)


src/lib/select/select.spec.ts
```diff
@@ -988,37 +988,50 @@ describe('MdSelect', () => {
         select.style.marginRight = '30px';
       });
 
-      it('should align the trigger and the selected option on the x-axis in ltr', () => {
-        trigger.click();
-        fixture.detectChanges();
+      it('should align the trigger and the selected option on the x-axis in ltr', async(() => {
+        fixture.whenStable().then(() => {
+          fixture.detectChanges();
```
trshafer: are these detectChanges necessary within the whenStable block?

crisbeto: Doesn't look like it, although most of these `whenStable` calls shouldn't be necessary either.

PR #3864, opened 2017-03-31
fix(select): prevent the panel from going outside the viewport horizontally

### LINKED ISSUE #3504: md-select right text overflow
image

The mat-select-panel goes behind window if it is to close to the right.
(the cdk-overlay has a wrong left style)

The mat-select-panel needs to go to the left. (If you can add something like mat-select-parent-class
to the md-select, you can fix it easily using custom css)

The mat-select-content can't have a custom class, and it will not auto position itself to the left, it will go
behind the window.

Just put a select all to the right with a medium text, and click it.
Or just see the Plunker:
http://plnkr.co/6AKBKE6ZwhCKMFbK1PwF?p=preview

We can't have a select all to the right because of this bug or lack of feature

The problem itself is the overlay system.

If we can pass something like this:
`<md-select placeholder-position="right || left">`
or
`<md-select custom-class="placeholder-right">`

We can set the left of the overlay manually or add a custom class and add a margin to the placeholder (.mat-select-panel) inside the overlay

### LINKED ISSUE #3831: mat-select is not detecting screen overflow when placeholder is hidden
Bug

mat-select detects that the dropdown menu will overflow the screen, and repositions the dropdown.

On some cases mat-select will show the dropdown outside of screen.

http://plnkr.co/edit/AMbVUV0JFKI92zLWuUNl?p=preview

I would like to use mat-select  dropdown without the placeholder.

crisbeto:
Prevents the select panel from going outside the viewport along the x axis.

Fixes #3504.
Fixes #3831.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -27,10 +27,23 @@ import {TAB} from '../core/keyboard/keycodes';
 import {ScrollDispatcher} from '../core/overlay/scroll/scroll-dispatcher';
 
 
+class FakeViewportRuler {
+  getViewportRect() {
+    return {
+      left: 0, top: 0, width: 1014, height: 686, bottom: 686, right: 1014
+    };
+  }
+
+  getViewportScrollPosition() {
+    return {top: 0, left: 0};
+  }
+}
+
 describe('MdSelect', () => {
   let overlayContainerElement: HTMLElement;
   let dir: {value: 'ltr'|'rtl'};
   let scrolledSubject = new Subject();
+  let fakeViewportRuler = new FakeViewportRuler();
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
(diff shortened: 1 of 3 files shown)


crisbeto:
@jelbourn can you take a look at this one? I'll sort out the CI failure and the conflicts afterwards.

jelbourn:
@kara should look at this one, I'm not as familiar with how the select does positioning

src/lib/select/select.spec.ts
```diff
@@ -882,6 +895,94 @@ describe('MdSelect', () => {
...
+          const panelLeft = document.querySelector('.mat-select-panel')
+              .getBoundingClientRect().left;
+          expect(panelLeft).toBeGreaterThan(0,
+              `Expected select panel to be inside the viewport in ltr.`);
+        });
+      }));
+
+      it('should stay within the viewport when overflowing on the right in rtl', async(() => {
+        dir.value = 'rtl';
+        select.style.left = '-100px';
```
kara: Seems like either the test name or the style here is out of sync. Did you mean to say "when overflowing on the left in rtl"?  The test on line 947 seems to cover the case on the right.

crisbeto:
Addressed the feedback, fixed the test failures, rebased and squashed @kara. Can you take one more look?

kara:
@crisbeto Needs rebase

crisbeto:
Rebased, but it seems like one of the tests broke. Will relabel when it's fixed.

crisbeto:
There we go. Had deleted the `fakeViewport`.

PR #3865, opened 2017-03-31
fix(chips): focus next chip correctly when a middle chip is removed

emoralesb05:
there is a need to wait until a middle chip has been removed from children to focus the chip that will be set at the same index.

currently it works fine when you remove the final chip since it just tries to focus the previous chip.

https://github.com/angular/material2/blob/master/src/lib/chips/chip-list.ts#L186-L188

but when a middle chip is removed, it just focuses the same one before its removed so the focus is lost after that.

https://github.com/angular/material2/blob/master/src/lib/chips/chip-list.ts#L180-L185

Here is a plnkr to demonstrate the issue: http://plnkr.co/edit/LdDUdrQ2vuhgVYLfWxj1?p=preview

### Steps to reproduce

- Click on the last chip and see the focus move to the previous chip
- Now click on a middle chip and see the focus go away. (we lose it because of this)

md-chip-list-focus-bug

### Proposed fix in PR

We need to push the `this._keyManager.setActiveItem(chipIndex);` at the end of the queue of the change detection so it happens after the `chip` has been remove. (removed as a child)

### DIFF
--- a/src/lib/chips/chip-list.spec.ts
+++ b/src/lib/chips/chip-list.spec.ts
@@ -75,9 +75,14 @@ describe('MdChipList', () => {
     });
 
     describe('on chip destroy', () => {
-      it('focuses the next item', () => {
+      it('focuses the next item', async(() => {
         let array = chips.toArray();
         let midItem = array[2];
+
+        // Next item to focus
+        let nextFocusedItem = array[3];
+        // Spy on its focus() method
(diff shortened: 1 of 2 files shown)


jelbourn:
@tinayuangao is this PR still relevant to the chip behavior today?

tinayuangao:
@jelbourn I can't reproduce the issue on master. Close this PR now

PR #3869, opened 2017-04-01
fix(checkbox, radio): ripple error on focus event

### LINKED ISSUE #3856: Uncaught TypeError when selecting checkbox with ripple disabled
Bug

No uncaught typeerrors when selecting checkboxes if ripple is disabled

When selecting or tabbing to a checkbox with `[disableRipple]="true"`:

```
Uncaught TypeError: Cannot read property 'launch' of undefined
    at SafeSubscriber.eval [as _next] (material.umd.js:6095)
    . . .
```
http://plnkr.co/edit/YcArpxwX8fQReqj55ufU?p=preview
Open console and click the "No ripple" checkbox

latest versions

DevVersion:
* The checkbox and radio disable the ripples using a `*ngFor` check. This causes the `ViewChildren` instance to turn null. The reference is required for the focus indicators.

* If ripples are disabled there will be still ripples on click. This is due to the fact that label elements redirect focus to the underlying input element. The `focusOrigin` is therefore `program` and a focus ripple will show up.

* Tests had to be adjusted because the `[md-ripple]` element won't be removed anymore. Therefore the tests now confirm that no ripples are showing up.

* Also removed the `(blur)` handlers in the template of the components. We already listen to `blur` events with the `FocusOriginMonitor`.

**Note**: Due to the `setTimeout` calls in the `FocusOriginMonitor` and the `fakeAsyncZone` it was the best to avoid `fakeAsync` tests.

@mmalerba  To be able to fix the issue that ripples are still showing up while `disableRipples` is set, we had to remove the `program` check for the focus ripples (I think it's a higher priority now)

Fixes #3856.

### DIFF
--- a/src/lib/radio/radio.ts
+++ b/src/lib/radio/radio.ts
@@ -23,9 +23,9 @@ import {
   UniqueSelectionDispatcher,
   MdRipple,
   FocusOriginMonitor,
+  FocusOrigin,
 } from '../core';
 import {coerceBooleanProperty} from '../core/coercion/boolean-property';
-import {Subscription} from 'rxjs/Subscription';
 
 
 /**
@@ -405,11 +405,8 @@ export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
   /** The child ripple instance. */
   @ViewChild(MdRipple) _ripple: MdRipple;
 
-  /** Stream of focus event from the focus origin monitor. */
-  private _focusOriginMonitorSubscription: Subscription;
-
   /** Reference to the current focus ripple. */
-  private _focusedRippleRef: RippleRef;
+  private _focusRipple: RippleRef;
 
   /** The native `<input type=radio>` element */
   @ViewChild('input') _inputElement: ElementRef;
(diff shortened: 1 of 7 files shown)


src/lib/radio/radio.ts
```diff
@@ -516,4 +498,20 @@ export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
     }
   }
 
+  /** Function is called whenever the focus changes for the input element. */
+  private _onInputFocusChange(focusOrigin: FocusOrigin) {
```
crisbeto: I'm not sure that calling it a "focus change" is accurate, `onInputFocus` should be fine.

DevVersion: Are you sure about it? I felt like without `Change` it does not really feel like it would call for `blur` as well.

crisbeto: Ah, I wasn't aware that it gets called for blur as well. In that case it's fine, although it could get stated in the comment.

DevVersion: I thought I added it to the function JSDoc

crisbeto: The JSdoc reads to me like it's only when it receives focus. Either way this really minor, I'll delete the note.

arlowhite:
Does this fix (or relate to): https://github.com/angular/material2/issues/4037

PR #3875, opened 2017-04-02
fix(dialog): capture previously focused element immediately

crisbeto:
Captures the previously-focused element immediately, instead of waiting until the animation is done. This fixes a regression from #3774, because if we wait for the animation to finish, the focus might have shifted, or the element could have been disabled.

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -61,7 +61,7 @@ export class MdDialogContainer extends BasePortalHost {
   private _elementFocusedBeforeDialogWasOpened: HTMLElement = null;
 
   /** Reference to the global document object. */
-  private _document: Document|Document;
+  private _document: Document;
 
   /** The dialog configuration. */
   dialogConfig: MdDialogConfig;
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -6,6 +6,8 @@ import {
   Renderer,
   ElementRef,
   EventEmitter,
+  Inject,
+  Optional,
 } from '@angular/core';
 import {
   animate,
@@ -15,6 +17,7 @@ import {
   transition,
   AnimationEvent,
 } from '@angular/animations';
+import {DOCUMENT} from '@angular/platform-browser';
 import {BasePortalHost, ComponentPortal, PortalHostDirective, TemplatePortal} from '../core';
 import {MdDialogConfig} from './dialog-config';
 import {MdDialogContentAlreadyAttachedError} from './dialog-errors';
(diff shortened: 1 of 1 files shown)


src/lib/dialog/dialog-container.ts
```diff
@@ -118,11 +124,20 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
...
 
   /**
+   * Saves a reference to the element that was focused before the dialog was opened.
+   * @private
+   */
+  private _savePreviouslyFocusedElement() {
+    if (this._document) {
+      this._elementFocusedBeforeDialogWasOpened = this._document.activeElement as HTMLElement;
+    }
+  }
```
jelbourn: Any way to capture this in a unit test?

crisbeto: It was tricky to write a test that fails when we expect it to since we trigger the change detection and animations manually.

src/lib/dialog/dialog-container.ts
```diff
@@ -118,11 +124,20 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
     // If were to attempt to focus immediately, then the content of the dialog would not yet be
     // ready in instances where change detection has to run first. To deal with this, we simply
     // wait for the microtask queue to be empty.
-    this._elementFocusedBeforeDialogWasOpened = document.activeElement as HTMLElement;
     this._focusTrap.focusFirstTabbableElementWhenReady();
   }
 
   /**
+   * Saves a reference to the element that was focused before the dialog was opened.
+   * @private
```
jelbourn: Don't need `@private` JsDoc

src/lib/dialog/dialog-container.ts
```diff
@@ -77,7 +80,8 @@ export class MdDialogContainer extends BasePortalHost implements OnDestroy {
     private _ngZone: NgZone,
     private _renderer: Renderer,
     private _elementRef: ElementRef,
-    private _focusTrapFactory: FocusTrapFactory) {
+    private _focusTrapFactory: FocusTrapFactory,
+    @Optional() @Inject(DOCUMENT) private _document: any) {
```
jelbourn: We can add a type for this while working around Angular Universal's limitations by doing `Document|Document`

crisbeto:

src/lib/dialog/dialog-container.ts
crisbeto: Doing `Document|Document` fails in AoT.

jelbourn: Chatted with Chuck and this is a known issue with the metadata extractor (https://github.com/angular/angular/issues/15424). Until that's fixed, we can inject it as `any`   without the `private` sugar and then manually assign it to the field with type `Document`.

crisbeto: Done.

kara:
@crisbeto Can you rebase?

PR #3876, opened 2017-04-02
fix(list): set explicit box-sizing on icon

### LINKED ISSUE #3863: md-list-icon is not aligned with global `box-sizing: inherit`
Bug

It's [pretty common](https://www.paulirish.com/2012/box-sizing-border-box-ftw/) to globally apply inherited box sizing to all elements:

```
html {
  box-sizing: border-box;
}
*, *:before, *:after {
  box-sizing: inherit;
}
```

This causes the `md-list-icon` to no longer keep the correct size, because it is inheriting `border-box` from `.mat-list-item-content`. It should explicitly use `content-box`.

http://plnkr.co/edit/nDwtfSsqTrotyZpNfr5I?p=preview

crisbeto:
Sets an explicit `box-sizing` on the icons inside of lists, in order to avoid alignment issues if the consumer uses `border-box` globally.

Fixes #3863.

### DIFF
--- a/src/lib/list/list.scss
+++ b/src/lib/list/list.scss
@@ -88,6 +88,7 @@ $mat-dense-list-icon-size: 20px;
     width: $icon-size;
     height: $icon-size;
     font-size: $icon-size;
+    box-sizing: content-box;
     border-radius: 50%;
     padding: 4px;
   }


jelbourn:
@crisbeto presubmit success, just needs rebase

crisbeto:
Rebased @jelbourn.

PR #3877, opened 2017-04-02
fix(tabs): re-align the ink bar when the viewport size changes

### LINKED ISSUE #1231: [Tabs] Tabs active indicator not responsive

When resizing the VP, The tab active indicator size and position is not recalculated.
This is causing for problems if you want to have the tabs span accross the full width via setting `flex-grow: 1` for example. I currently have

```
:host /deep/ .md-tab-label {
  flex: 1;
  min-width: 120px !important;
}
```

in my component styles.

This represents the fixed tabs type of the md tabs specs: https://material.google.com/components/tabs.html#tabs-types-of-tabs

The position and size should be recalculated when resizing the viewport.

When You resize the Viewport, the length and position of the active indicator stays the same.

Visit http://plnkr.co/edit/tFkzdaw9ejYBpo3OGnKC?p=preview and resize the VP.

Static tabs are part of the material design spec.

All, I think. Tested with aplha 7 and alpha 8.

### LINKED ISSUE #2518: Tab pagination: strange behaviour when resizing the browser window
When resizing the browser window the tab pagination does not behave as expected

There are some different problems (don't know if you would prefer one issue per problem)

1. The different tabs resize themselves, but the colored line to indicate the focused tab keeps the "old" size.
2. When the window is resized to be smaller I get the arrows to allow me to scroll the tabs, but when I resize the window again they stay there even if there is enough space for all tabs.
3. If you scroll the tabs using the arrows some of them go out of sight (expected), but if I resize the window they are kept out of sight even if there is enough space for them.
4. Sometimes the arrows aren't shown when resizing. When this happens I can resize as much as I want, they are never shown. I don't know how to reproduce it. It happens when using the component.

Use the plunker and resize the viewport using the splitter. You should be able to reproduce all the problems,

Plunker : http://plnkr.co/edit/gAeoyntvHPo3j7SomTMy

Angular 2.3.1, Material 2.0 beta 1, Windows 10 with Chrome 55.0.2883.87 m

### LINKED ISSUE #3044: Misalignment of [md-ink-bar] element for Stretched Tabs on browser's window resize
Bug

Highlighter [md-ink-bar] should invalidate position on window resize

Frozen [md-ink-bar]. Please see attached image.

1. Browse [demo application](https://github.com/angular/material2/tree/master/src/demo-app) from material2
2. Navigate to tabs  (Especially for Stretched Tabs section)
3. Resize browser

According to up to date sources of [material2](https://github.com/angular/material2/tree/master/src/demo-app)

image

### LINKED ISSUE #3845: md-tab-nav-bar: ink doesn't change with resizing tabs
Bug

When the window hits the `$mat-xsmall` breakpoint, min widths for `.mat-tab-label` and `.mat-tab-link` change from 160px to 72px. The ink bar does not reflect that.

http://plnkr.co/edit/6NlXoXhzaDHsHtTo9JWn?p=preview

Master

#### Other Info
Related to #3044 but does not require tabs to be stretched.

crisbeto:
* Recalculates the ink bar position, as well as the pagination, if the viewport size changes.
* Adds a missing test on the `tab-nav-bar` regarding repositioning on direction changes.

Fixes #3845.
Fixes #3044.
Fixes #2518.
Fixes #1231.

### DIFF
--- a/src/lib/tabs/tab-header.ts
+++ b/src/lib/tabs/tab-header.ts
@@ -12,13 +12,20 @@ import {
   AfterContentChecked,
   AfterContentInit,
   OnDestroy,
+  NgZone,
 } from '@angular/core';
 import {RIGHT_ARROW, LEFT_ARROW, ENTER, Dir, LayoutDirection} from '../core';
 import {MdTabLabelWrapper} from './tab-label-wrapper';
 import {MdInkBar} from './ink-bar';
 import {Subscription} from 'rxjs/Subscription';
+import {Observable} from 'rxjs/Observable';
 import {applyCssTransform} from '../core/style/apply-transform';
 import 'rxjs/add/operator/map';
+import 'rxjs/add/operator/auditTime';
+import 'rxjs/add/observable/of';
+import 'rxjs/add/observable/merge';
+import 'rxjs/add/operator/startWith';
+
 
 /**
  * The directions that scrolling can go in when the header's tabs exceed the header width. 'After'
@@ -68,8 +75,8 @@ export class MdTabHeader implements AfterContentChecked, AfterContentInit, OnDes
   /** Whether the header should scroll to the selected index after the view has been checked. */
   private _selectedIndexChanged = false;
 
-  /** Subscription to changes in the layout direction. */
-  private _directionChange: Subscription;
+  /** Combines listeners that will re-align the ink bar whenever they're invoked. */
+  private _realignInkBar: Subscription = null;
 
   /** Whether the controls for pagination should be displayed */
   _showPaginationControls = false;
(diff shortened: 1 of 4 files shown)


src/lib/tabs/tab-header.ts
```diff
@@ -150,17 +161,23 @@ export class MdTabHeader implements AfterContentChecked, AfterContentInit, OnDes
...
   ngAfterContentInit() {
-    this._alignInkBarToSelectedTab();
-
-    if (this._dir) {
-      this._directionChange = this._dir.dirChange.subscribe(() => this._alignInkBarToSelectedTab());
-    }
+    this._realignInkBar = this._ngZone.runOutsideAngular(() => {
+      return Observable.merge(
+          this._dir ? this._dir.dirChange : Observable.of(null),
+          Observable.fromEvent(window, 'resize').auditTime(10)
```
jelbourn: We should only do this when on platform-browser

crisbeto:

PR #3879, opened 2017-04-03
fix(select): close the panel when pressing escape

crisbeto:
* Closes the `md-select` panel when pressing escape in the same way as the native select.
* Adds a missing test to ensure that the select closes when tabbing out.

### DIFF
--- a/src/lib/core/a11y/list-key-manager.ts
+++ b/src/lib/core/a11y/list-key-manager.ts
@@ -1,5 +1,5 @@
 import {QueryList} from '@angular/core';
-import {UP_ARROW, DOWN_ARROW, TAB, HOME, END} from '../core';
+import {UP_ARROW, DOWN_ARROW, TAB, HOME, END, ESCAPE} from '../core';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 
@@ -18,7 +18,8 @@ export interface CanDisable {
 export class ListKeyManager<T extends CanDisable> {
   private _activeItemIndex: number = null;
   private _activeItem: T;
-  private _tabOut: Subject<any> = new Subject();
+  private _tabOut = new Subject<void>();
+  private _escape = new Subject<void>();
   private _wrap: boolean = false;
 
   constructor(private _items: QueryList<T>) {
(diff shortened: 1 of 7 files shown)


src/lib/core/a11y/list-key-manager.ts
```diff
@@ -63,6 +64,9 @@ export class ListKeyManager<T extends CanDisable> {
       case END:
         this.setLastItemActive();
         break;
+      case ESCAPE:
```
jelbourn: I'm not sure the escape handling really belongs here; I think the problem is that while it's called `ListKeyManager`, it's really more like `ListItemMovementKeyManager`. The fact that there's both `tab` and `escape` as observables with no behavior seems like a code smell; perhaps the components should handle these things themselves. 

@kara thoughts?

crisbeto: Hmm I see your point. We could have something like PopupKeyManager that inherits from the ListKeyManager instead.

kara: Yeah, agree that this doesn't seem like a natural fit for a "list" key manager. I like the idea of having a PopupKeyManager, but inheriting from ListKeyManager doesn't seem like quite the right abstraction since sometimes we'll have popups that aren't lists. Maybe it could be a sibling?  Though it's annoying to have to instantiate both a ListKeyManager and a PopupKeyManager. 

I think it could still be useful to have a service for this (rather than keeping in components) because turning the key events into observables makes it easy to combine and pass them around later.

crisbeto: Isn't a service for this overkill? The only thing it would do is to expose `Observable.fromEvent(document, keydown).filter(e => e.keyCode === ESCAPE)`. I think that at this point it would be better to just inline it or bake it into the overlay directive.

kara: Overlay directive is an idea.

crisbeto:
Reworked this one as discussed @kara.

src/lib/core/overlay/overlay-directives.ts
```diff
@@ -249,6 +254,12 @@ export class ConnectedOverlayDirective implements OnDestroy {
 
     this._position.withDirection(this.dir);
     this._overlayRef.getState().direction = this.dir;
+    this._closeKeyListener = this._renderer.listen('document', 'keydown',
```
kara: This method is getting a bit long. Can you move this new block of code into its own method, maybe `_initEscapeListener()` or something?

PR #3880, opened 2017-04-03
fix(MdIcon): import Observable throw static method

jeffbcross:
Observable.throw is being called without explicitly importing
the throw operator, which was causing exceptions to be raised
when trying to call Observable.throw (since no other module
had patched throw onto the Observable yet).

### DIFF
--- a/src/lib/icon/icon-registry.ts
+++ b/src/lib/icon/icon-registry.ts
@@ -5,6 +5,7 @@ import {MdError} from '../core';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/observable/forkJoin';
 import 'rxjs/add/observable/of';
+import 'rxjs/add/observable/throw';
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/filter';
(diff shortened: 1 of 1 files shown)


jeffbcross:
It would require significant changes to the codebase to write a unit test for this change, because of the nature of the side-effect-producing operators. But I did verify that this change fixes the exception I was getting in my application.

jelbourn:
I tried to do this once before and it caused all kinds of failures in Google-land (something weird about throw). I'll let you know how the presubmit goes...

jelbourn:
More recent update: this does indeed cause internal presubmit errors and we're not really sure why.

jeffbcross:
Presubmit shmesubmit. Is it build errors or test failures?

jeffbcross:
How about I do this the clean way instead?

jeffbcross:
I'm not sure how this happened, but throw was already added recently, and my commit now has it imported twice. I'm not sure if I rebased and force pushed at some point or what, but this PR is no longer necessary.

Issue #3002, opened 2017-02-09
Unclear documentation about snackbar usage on documentation website


thanhpd:
#### Bug, feature request, or proposal:
Documentation: Snackbar component is unclear, and its Plunker code example is not running

#### Details:
1. It seems the Snackbar component documentation on website does not describe fully about how to open snackbar from component and what to do with `MdSnackBarRef` in API Reference tab while they are mentioned in Overview tab

2. Plunker example in Examples tab of Snackbar component is not working.
    *Reproduce steps:*
    - Initial Plunker example when clicking Edit in Plunker: https://plnkr.co/edit/mH5Hh14TZAD3WH2AlU2x?p=preview
    - Click Pizza Party button, the snackbar does not appear

    *Notes:*
    - Developer need to add `PizzaPartyComponent` in `declarations` and `entryComponents` of `ngModule` in `main.ts` file, plus create new template `snack-bar-component-example-snack.html` and component stylesheet `snack-bar-component-example-snack.css` for `PizzaPartyComponent`.
    - Plunker example: https://plnkr.co/edit/IVRsaVTFRiaHIHGXnhAw?p=preview

3. The documentation does not describe how to execute the action of a component itself if that component is opened in snackbar.
In the project source code, I've found that to execute the dismiss snackbar action when open snack bar from `open` method you guys did this:
- Create a self-reference with `MdSnackBarRef` of the content component: In class `SimpleSnackBar`, there is `snackBarRef: MdSnackBarRef<SimpleSnackBar>` property. To be able to dismiss snackbar, `SimpleSnackBar` component has to call `dismiss()` method, which in turn calls the `_action()` method of `snackBarRef`.
- In `open` method of `MdSnackBar`:
```javascript
open(message: string, action = '', config: MdSnackBarConfig = {}): MdSnackBarRef<SimpleSnackBar> {
    config.announcementMessage = message;
    let simpleSnackBarRef = this.openFromComponent(SimpleSnackBar, config);
    simpleSnackBarRef.instance.snackBarRef = simpleSnackBarRef;
    simpleSnackBarRef.instance.message = message;
    simpleSnackBarRef.instance.action = action;
    return simpleSnackBarRef;
}
```
At `line 4`, the `snackBarRef` of instance `SimpleSnackBar` of `simpleSnackBarRef` was assigned to `simpleSnackBarRef` itself. Therefore, the snackbar component was able to close.

[...]

Issue #2934, opened 2017-02-04
Unclear docs example for Dialog component


jamzi:
Dialog with a result example in the docs is a little bit unclear in my opinion on how to return the value back to DialogResultExample component. If you open the plunker version, you can nicely see two .html files where it shows you that on close method you pass in the parameter. But in the docs version, there is only html for DialogResultExample component.

I am suggesting that the html tab for docs on this example would be like this:

```html
//dialog-result-example.html
<button md-button (click)="openDialog()">Launch dialog</button>
You chose: {{selectedOption}}

//dialog-result-example-dialog.html
<h1 md-dialog-title>Dialog</h1>
<div md-dialog-content>What would you like to do?</div>
<div md-dialog-actions>
  <button md-button (click)="dialogRef.close('Option 1')">Option 1</button>
  <button md-button (click)="dialogRef.close('Option 2')">Option 2</button>
</div>
```

P.S. also, the title of example is Dailog with a result (notice the typo for Dialog).

Btw, you are doing an amazing work on Angular Material!

M-a-c:
@kara The code to fix this is probably out side of the scope of what the 'Git' touches.
I've looked everywhere this feature is in there.. (Might not be 'good for community contribution)

@DevVersion You seem to do the tooling, I think these issues are related to that. I would love to help but really cannot find the code that generates these issues.

See :
https://github.com/angular/material2/blob/f526c9de8725496b04743a326c47f6fcc35b54e2/src/examples/example-module.ts#L111
https://github.com/angular/material2/blob/master/src/examples/example-data.ts#L27
The files are supposed to be injected into the md-tab-group inside the docs-example-viewer-source but they're not ¯\_(ツ)_/¯.
Would gladly try and fix this if I could find out where exactly where the tabs for examples are generated
https://material.angular.io/components/component/dialog --> examples  --> (view source)
Notice: html ts css  but no additional html file.

These issues are related to #3002 but that is also missing additional files list in its definition in example-module.ts

jelbourn:
Tracking via https://github.com/angular/material.angular.io/issues/165

coocoorhin:
great job! i have been working on the issue for a very long time when i was following the doc. your solution magically made it work!!! thanks

PR #3888, opened 2017-04-03
Docs (snack-bar-component): added additional files and selectors

M-a-c:
put missing files in the additionalFiles array.
* snack-bar-component-example-snack.css
* snack-bar-component-example-snack.html
#3002 

Also note that this is related to #2934 
As adding these files probably wont fix the fact that these additional files don't show up for some reason in the docs.


M-a-c:
* I don't exactly understand why b080f7a didn't pass and 8492c36 did..

jelbourn:
This file is now auto-generated

PR #3897, opened 2017-04-04
Fix(md-menu) : Added prevent close( when user clicks inside menu panel) feature.

ravinderpayal:
menu-caveate
As It's a feature the core team decided to add, and I was in need of this feature for functionality like drop-down menu having some setting controls where user may want to change multiple settings in one go. As currently after every click inside or outside of drop-down menu, it disappears. But adding some additional check based on the value of preventClose @Input() supplied parent node can add this desired functionality. Actually I am creating a emoji picker, and have multiple categories of emoji, so user is required to change the category by clicking on tabs(Yeah I am using MD-tabs), menu disappears so this feature will help in accomplishing my specific requirement, and other people's similar requirements.

### DIFF
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -108,12 +108,10 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
           */
           if (this._menuOpen && !this._element.nativeElement.contains(event.target)) {
               this.closeMenu();
-          }
-          else {
+          } else {
               this.openMenu();
           }
-      }
-      else {
+      } else {
           /** If no `preventClose` is defined, it will go same old way. */
           this._menuOpen ? this.closeMenu() : this.openMenu();
       }


ravinderpayal:
Signed in

willshowell:
Hey @ravinderpayal, as a heads up, the team will certainly ask for tests to verify the behavior of the api change.

You should also try to get your linting issues fixed before review. It looks like a couple of the comments are unnecessary / [not formatted correctly.](https://github.com/angular/material2/blob/master/CODING_STANDARDS.md#jsdoc-comments)

Additionally, consider using `disableClose` to match the [api for MdDialogConfig](https://github.com/angular/material2/blob/master/src/lib/dialog/dialog-config.ts#L24) (not necessary, just a thought).

Finally, it would be a good idea to explain your motivation behind the change in your original comment!

ravinderpayal:
@willshowell Thanks Sir, I will look consider your thoughts, and thanks one more for clearing my doubts about failure of travis build.

ravinderpayal:
Travis Build Passed, and @wilshowell I have kept the `preventClose` as it is, because `disableClose` mean something like completely disabling the close, while we are preventing close only when user click inside the panel.

willshowell:
As a few more notes, JSDoc comments (`/** */`) should only be used for api purposes (like preventClose and toggleMenu) while you should use standard line comments (`//`) elsewhere.

Again, tests will be required by the team, so you should add those too.

Good luck with your PR!

ravinderpayal:
Ok brother

agostinalucia:
👏

zmanring:
I need this! :D Thank you @ravinderpayal

ravinderpayal:
While working on some other project I found a better option, let me patch mine code.

ravinderpayal:
@zmanring @agostinalucia @mprobst @willshowell  I have created a fresh pull request after testing working of code. This current commits are not required. I have added `preventClose` input/attribute on `<md-menu>` component itself, but before it's was required to add `preventClose` on trigger component/element.

ezgif com-video-to-gif

https://github.com/angular/material2/pull/4354

jelbourn:
Closing since this is now quite stale. I think this feature has come up a few times in issue discussion and we've decided that we wouldn't want to add this option to mat-menu since it would go against W3C a11y recommendations.

Issue #2605, opened 2017-01-12
Remove individual README.md content, point to material.angular.io instead


jelbourn:
We should delete the content of each individual README and replace it with a link to the corresponding page on material.angular.io.

Requirements:
* The content should be captured in the overview doc
* Any significant examples in the README should be turned into a live example on the docs site before removing the content.

M-a-c:
@jelbourn I could use some clarification if this task. All docs to already link to material.angular.io unless they are core. If there are docs that need more meat to them I'd be happy to add some material. Suggestions welcome.

jelbourn:
This was done but we forgot to close the issue.

PR #3902, opened 2017-04-04
Docs (components/core) : fileNameMds for use in docs website.

M-a-c:
Not sure where links from lib/core actually go relative to the web URL
I was assuming :
* /components/core/NAME
* /components/core
#2605


jelbourn:
We aren't ready to surface docs for things under `core/` yet on material.angular.io just yet

PR #3918, opened 2017-04-05
fix (Menus): max-height: calc(100vh + 48px); and goes off page

### LINKED ISSUE #2725: md-menu is defaulted to `max-height: calc(100vh + 48px);` and goes off page
Bug

I expect that the menu would size to stay on the screen and scroll when it has a lot of items in it.

Currently, it will just go off the bottom of the page and I won't be able to see a lot of items.

image
This is the bottom of my browser and there are more items below it.

Add a menu with a lot of buttons. Don't put the menu at the top of the page. Open the menu, it will go down below the bottom of the page.

http://plnkr.co/edit/FPHeqOjRHpJOBNGn5dvz?p=preview

If I want a bunch of items in a list.

Angular: 2.3.1
Material: 2.0.0-alpha.11-3
Mac
Chrome

And I assume it's on most others.

It looks like this is happening in https://github.com/angular/material2/blob/master/src/lib/menu/menu.scss#L16. So having it be a full viewport means it's guaranteed to go below the page.

M-a-c:
The Material Design spec says the max height of a menu should be one row height less than the page height.

Fixes #2725 

Brings up more questions like:
* Is there a minimum height in the MD spec?
* Should vh really be used because menus can be inside of components?

### DIFF
--- a/src/lib/menu/menu.scss
+++ b/src/lib/menu/menu.scss
@@ -12,8 +12,8 @@ $mat-menu-vertical-padding: 8px !default;
   @include mat-menu-base();
   @include mat-menu-positions();
 
-  // max height must be 100% of the viewport height + one row height
-  max-height: calc(100vh + 48px);
+  // max height must be 100% of the viewport height - one row height
+  max-height: calc(100vh - 48px);
 
   @include cdk-high-contrast {
     outline: solid 1px;


willshowell:
The fix looks to have been made in https://github.com/angular/material2/pull/4214

Issue #3936, opened 2017-04-06
MdCard and MdListItem use base color rather than text color


willshowell:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Default styling of text on a "light-theme" MdCard or MdListItem should be 87% black

#### What is the current behavior?
Text defaults to 100% black.

#### What are the steps to reproduce?
http://plnkr.co/edit/zkrjdBwt5wAe85FllmDe?p=preview

#### Is there anything else we should know?
Culprit for MdCard is [here](https://github.com/angular/material2/blob/master/src/lib/card/_card-theme.scss#L11). Culprit for MdListItem is [here](https://github.com/angular/material2/blob/master/src/lib/list/_list-theme.scss#L11)

kyle-burke:
Hey @willshowell, I made a PR for this [here](https://github.com/angular/material2/pull/3952).

PR #3952, opened 2017-04-07
fix(card, list): Update MdCard and MdListItem color property to use theme's text color.

kyle-burke:
Fixes [#3936](https://github.com/angular/material2/issues/3936).


kyle-burke:
The screenshot tests were sharp enough to notice the slight change in opacity. Could someone tell me how to provide that test suite with an updated screenshot? I don't see anything in the Contributing file for doing this (might be worth a separate PR to describe this process)

PR #3960, opened 2017-04-08
style(checkbox): format checkbox component based on angular style guide

### LINKED ISSUE #3921: Consistent component structure throughout the project
This is a proposal, material2 seems really to be all over the place in regards to where code is placed. If we take a look at what the style guide says https://angular.io/docs/ts/latest/guide/style-guide.html for example.

It seems like a format for the code should be something like this:
@Input() section - lexical order
@Output() section - lexical order
@[other decorators] - lexical order
ng functions - lifecycle order
constructor
ngOnChanges
ngOnInit
ngDoCheck
ngAfterContentInit
ngAfterContentChecked
ngAfterViewInit
ngAfterViewChecked
ngOnDestroy

Getters and setters - lexical order
Public functions - lexical order
Private functions - lexical order

What are your thoughts on me going through each component and reordering them to be like this?
If you all have a reason for having this the way you do, would love to see a document on where that is.

The project should be structured in a way so code is easy to find. Doing the convention I set above would allow for this.

Seems to be pretty random as far as I can tell.

Being able to go into any component and know where to go in the file just because the structure is consistent throughout would be amazing.

alexw10:
This change is to provide consistent style throughout the components. I am starting with just
checkbox but would slowly apply this throught material2 for you guys if necessary. This is done
based off of https://angular.io/styleguide and I think will make knowing where things are when going
into each component much easier.

This has no code changes in it just style and moving things around to be in lexical order.

Thoughts on this style? If this is accepted ill be going through each one and applying this same concept to each component and may even recruit a few people that said they would help with this in my original issue I created:
https://github.com/angular/material2/issues/3921

closes #3921

### DIFF
--- a/src/lib/checkbox/checkbox.ts
+++ b/src/lib/checkbox/checkbox.ts
@@ -1,26 +1,22 @@
 import {
-  ChangeDetectorRef,
+  AfterViewInit,
   ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   ElementRef,
   EventEmitter,
+  forwardRef,
   Input,
+  OnDestroy,
   Output,
   Renderer,
-  ViewEncapsulation,
-  forwardRef,
   ViewChild,
-  AfterViewInit,
-  OnDestroy,
+  ViewEncapsulation
 } from '@angular/core';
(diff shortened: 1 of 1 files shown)


jelbourn:
As I mentioned in #3921, I don't consider the order of class members to be important enough to introduce this level of churn.

Issue #3970, opened 2017-04-08
md-dialog: Property config does not exist on type MdDialogRef after updating to beta.3


ciesielskico:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
`config` is available in MdDialogRef to access  the additional data passed when opening the dialog.

#### What is the current behavior?
`ERROR: Property 'config' does not exist on type 'MdDialogRef<MyComponent>'.`

#### What are the steps to reproduce?

This used to work with beta.2, after updating to beta.3 config is no longer available on `MdDialogRef<T>`.
No breaking changes to md-dialog were mentioned in the changelog.

```
export class MyComponent {

    title;
    content;

    constructor(public dialogRef: MdDialogRef<MyComponent>) {
        this.title = dialogRef.config.data.title;
        this.content = dialogRef.config.data.content;
     }
}
```

#### Which versions of Angular, Material, OS, browsers are affected?
@angular/cli: 1.0.0
node: 7.8.0
os: linux x64
@angular/animations: 4.0.1
@angular/common: 4.0.1
@angular/compiler: 4.0.1
@angular/core: 4.0.1
@angular/forms: 4.0.1
@angular/http: 4.0.1
@angular/material: 2.0.0-beta.3
@angular/platform-browser: 4.0.1
@angular/platform-browser-dynamic: 4.0.1
@angular/router: 4.0.1
@angular/cli: 1.0.0
@angular/compiler-cli: 4.0.1

**This is the commit that introduced the regression**

https://github.com/angular/material2/commit/549222534c8cf2d2fdbd0f268aebb181159e2e51

fireyoshiqc:
It seems to indeed have been removed, since it's nowhere to be found in MdDialogRef :
https://github.com/angular/material2/blob/a71a5af2ea03d06e8328c218269b5d91820190db/src/lib/dialog/dialog-ref.ts

crisbeto:
The decision was very deliberate. [Please refer to this comment for the reasoning.](https://github.com/angular/material2/pull/3971#issuecomment-292721991)

PR #3971, opened 2017-04-08
fix(dialog): reintroduce config property in MdDialogRef

ciesielskico:
Fixes https://github.com/angular/material2/issues/3970

### DIFF
--- a/src/lib/dialog/dialog-ref.ts
+++ b/src/lib/dialog/dialog-ref.ts
@@ -1,5 +1,5 @@
 import {OverlayRef, GlobalPositionStrategy} from '../core';
-import {DialogPosition} from './dialog-config';
+import {DialogPosition, MdDialogConfig} from './dialog-config';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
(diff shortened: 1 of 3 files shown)


crisbeto:
Removing the config from the `DialogRef` was a very deliberate decision that can't really be considered a regression, It was removed, because it doesn't _need_ to be in the `DialogRef` since it isn't being used by it at all. Looking at your issue, it seems like you're using it to access the `data`, however the proper way to do it is to inject the `MD_DIALOG_DATA` token:

```ts
import {Inject} from '@angular/core';
import {MD_DIALOG_DATA} from '@angular/material';

@Component({
  selector: 'some-dialog',
  template: `...`
})
export class SomeDialog {
  constructor(@Inject(MD_DIALOG_DATA) data: any) {
    console.log(data);
  }
}
```

Even though you shouldn't need to do this, you can still access the config directly via `dialogRef._containerInstance.config`.

amivit:
How else can I `this.dialogRef.close();` ?

crisbeto:
You can still `dialogRef.close` @amivit, this issue was regarding the `dialogRef.config` property.

Issue #3973, opened 2017-04-08
Need documentation for Custom Module


Ploppy3:
According to the changelog of Beta 3 it is recommended to create our own Material module to reduce app size.

But I can't find any documentation on how to create my own Material module.

fxck:
https://angular.io/docs/ts/latest/guide/ngmodule.html

Ploppy3:
I know how to create an Angular module. What I don't know is what we are supposed to import from @angular/material, do you ?

fxck:
You are supposed to import whatever modules you are using in your app, ie. `MdButtonModule`. Either way, this is neither a bug report or feature request, support requests are better suited for gitter or stackoverflow.

Ploppy3:
This is what I do:
```
import { MdButtonModule } from "@angular/material";

@NgModule({
  imports: [
    MdButtonModule
  ],
})
export class MyMaterialModule { }
```

And in my app.module.ts I import `MyMaterialModule` and everything breaks.
Am I missing something obvious ?

fxck:
Are you sure you "know how to create an angular module" or that you actually read the guide I pointed you to? This section in particular https://angular.io/docs/ts/latest/guide/ngmodule.html#!#shared-module ?

Ploppy3:
Why not point out my mistake? Import instead of Export...

mackelito:
All working! All Good!
Stay safe, happy and helpful!

jelbourn:
Updated the getting started guide with an example.

candidosales:
```javascript
import { NgModule }      from '@angular/core';
import {
  MdButtonModule,
  MdSnackBarModule,
  MdCardModule,
  MdInputModule,
  MdCheckboxModule,
  MdIconModule,
  MdSidenavModule,
  MdSelectModule,
  MdToolbarModule,
  MdListModule,
  MdProgressBarModule,
  MdProgressSpinnerModule } from '@angular/material';

@NgModule({
    imports: [
      MdButtonModule,
      MdSnackBarModule,
      MdCardModule,
      MdInputModule,
      MdCheckboxModule,
      MdIconModule,
      MdSidenavModule,
      MdSelectModule,
      MdToolbarModule,
      MdListModule,
      MdProgressBarModule,
      MdProgressSpinnerModule ],
    exports: [
      MdButtonModule,
      MdSnackBarModule,
      MdCardModule,
      MdInputModule,
      MdCheckboxModule,
      MdIconModule,
[...]

PR #3974, opened 2017-04-08
guides(getting-started): update `MaterialModule` section

fxck:
should close https://github.com/angular/material2/issues/3973

I'm not 100% sure this is necessary just yet, since `MaterialModule` was only marked as deprecated, not removed, so it's up to you.


jelbourn:
Thanks, you reminded me that this was out of date- I wanted to word it a bit differently, so I opened #3986

PR #3977, opened 2017-04-08
fix(autocomplete): panel not being shown with delay and OnPush change detection

### LINKED ISSUE #3955: Autocomplete with OnPush does not open panel properly
Bug

If the autocomplete's `md-options` length changes from 0 to n due to options arriving asynchronously, the panel should display even with OnPush.

The panel does not show itself until the user types or refocuses the input.

http://plnkr.co/edit/fWoO47LDbCZFJ9i887jU?p=preview

My `options` are the result of an ajax request. When they arrive, I want `filteredOptions` to be recalculated and the autocomplete to reflect it. `filteredOptions` are recalculated, but the autocomplete panel doesn't show itself.

#### Other information
If `filteredOptions` uses `.startWith(x)` where `x.length > 0`, it works as expected. It only fails if the initial value is an empty array.

crisbeto:
Fixes an issue that prevented the autocomplete panel from being shown if the options are initialized with a delay in a component that has `OnPush` change detection.

Fixes #3955.

### DIFF
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -1,5 +1,13 @@
 import {TestBed, async, fakeAsync, tick, ComponentFixture} from '@angular/core/testing';
-import {Component, OnDestroy, QueryList, ViewChild, ViewChildren} from '@angular/core';
+import {
+  Component,
+  OnDestroy,
+  QueryList,
+  ViewChild,
+  ViewChildren,
+  ChangeDetectionStrategy,
+  OnInit,
+} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {NoopAnimationsModule} from '@angular/platform-browser/animations';
 import {MdAutocompleteModule, MdAutocompleteTrigger} from './index';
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -1090,6 +1099,24 @@ describe('MdAutocomplete', () => {
...
+      dispatchFakeEvent(fixture.debugElement.query(By.css('input')).nativeElement, 'focus');
+      tick(1000);
+      fixture.detectChanges();
+
+      Promise.resolve().then(() => {
+        let panel = overlayContainerElement.querySelector('.mat-autocomplete-panel') as HTMLElement;
+        let visibleClass = 'mat-autocomplete-visible';
+
+        fixture.detectChanges();
+        expect(panel.classList).toContain(visibleClass, `Expected panel to be visible.`);
```
jelbourn: Does this `Promise` ever resolve? Don't you need another `flushMicroTasks` to make it run?

crisbeto: I believe that the tick calls flushMicrotasks under the hood.

DevVersion: Yeah a `tick` should automatically clear the event-loop queue.

jelbourn: What I mean is that you're calling `tick` _before_ kicking off the promise.

crisbeto: Just reconfirmed that it resolved and the test breaks as expected if I remove the newly-added `markForCheck`.

PR #3982, opened 2017-04-08
Fixed stylesheet path to match the relevant path that is produced by …

Wo1v3r:
…ngCLI

Using the CLI is recommended at the top, therefore, the path in the example should match for a project produced from  ng new <project>

Also, The Path on the website itself is not updated (very different)


Wo1v3r:
I Signed it!

PR #3985, opened 2017-04-08
chore: add additional structure to generated docs

jelbourn:
* For API docs, wrap each generated document in `.docs-api`
* For markdown docs, wrap each generated document in `.docs-markdown`
* Remove the `h1` from API docs that contained only the component name
* Change the `h2` for API docs to be a friendlier title followed by an example import.
* Increase header level for API by one across the board (so that they're "under" the new `h2`)
* A few misc. docs fixes

I've got a PR lined-up on the docs site for once this goes in that will improve a few things.

### DIFF
--- a/src/lib/input/autosize.ts
+++ b/src/lib/input/autosize.ts
@@ -15,27 +15,21 @@ import {Directive, ElementRef, Input, OnInit} from '@angular/core';
   },
 })
 export class MdTextareaAutosize implements OnInit {
-  /** Minimum number of rows for this textarea. */
+  /** @deprecated Use mdAutosizeMinRows */
   @Input() minRows: number;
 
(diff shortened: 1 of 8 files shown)


src/lib/input/autosize.ts
```diff
@@ -15,27 +15,21 @@ import {Directive, ElementRef, Input, OnInit} from '@angular/core';
   },
 })
 export class MdTextareaAutosize implements OnInit {
-  /** Minimum number of rows for this textarea. */
+  /** @deprecated Use mdAutosizeMinRows */
```
crisbeto: I'm not sure that having the `@deprecated` and the description on the same line is valid.

jelbourn: It should be; this format is common inside Google.

PR #3987, opened 2017-04-09
chore: publish examples assets for plunker with docs content

jelbourn:


scripts/release/publish-docs-content.sh
```diff
@@ -52,17 +53,20 @@ do
...
     cp -r $filename $repoPath/guides
   fi
 done
 
-# Move highlighted examples into $repoPath
+# Copy highlighted examples into $repoPath
 cp -r $examplesSource/* $repoPath/examples
 
+# Copy example plunker assets
+rsync -a $docsPath/plunker $repoPath/plunker
```
DevVersion: I'm not sure `rsync` is supported in the Git Bash. But as we run this on the CI it's fine. Just think you can do the same with `cp`.

jelbourn: I don't think we'll ever need this running on a Window machine since it's not part of the dev workflow.

DevVersion: That's true. Just for consistency. Feel free to merge!

PR #3992, opened 2017-04-09
fix(autocomplete): placeholder not animating on focus

### LINKED ISSUE #5755: Floating label for autocomplete does not animate on focus
Bug

Floating label should float using animation on fucs.

Floating label snaps to floating-position without animation.

http://plnkr.co/edit/mVEqWQ3Xp4Ieep3iDs96?p=preview

crisbeto:
Fixes the placeholder not being animated on focus.

**Note:** The `_handleFocus` and `openPanel` methods do pretty much the same (aside from the extra boolean), but I wanted to avoid having to pass the `$event` to `openPanel` since the event isn't really relevant to the API for opening the autocomplete programmatically.

Fixes #5755.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -109,9 +109,9 @@ export function getMdAutocompleteMissingPanelError(): Error {
     '[attr.aria-owns]': 'autocomplete?.id',
     // Note: we use `focusin`, as opposed to `focus`, in order to open the panel
     // a little earlier. This avoids issues where IE delays the focusing of the input.
-    '(focusin)': 'openPanel()',
-    '(input)': '_handleInput($event)',
+    '(focusin)': '_handleFocus()',
     '(blur)': '_onTouched()',
+    '(input)': '_handleInput($event)',
     '(keydown)': '_handleKeydown($event)',
   },
   providers: [MD_AUTOCOMPLETE_VALUE_ACCESSOR]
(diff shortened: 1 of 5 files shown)


jelbourn:
@kara LGTM, but you should take a look

Screenshot failure seems unrelated (I think a gold didn't update when I approved and merged a previous PR).

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -234,14 +221,23 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
    * This causes the value to jump when selecting an option with the mouse.
    * This method manually floats the placeholder until the panel can be closed.
+   * @param shouldAnimate Whether the placeholder should be animated when it is floated.
    */
-  private _floatPlaceholder(): void {
+  private _floatPlaceholder(shouldAnimate = false): void {
     if (this._inputContainer && this._inputContainer.floatPlaceholder === 'auto') {
-      this._inputContainer.floatPlaceholder = 'always';
+      if (!shouldAnimate) {
+        this._inputContainer.floatPlaceholder = 'always';
```
kara: The reason why we set the placeholder to 'always' in this fashion is to avoid the placeholder jumping when an option is selected by the mouse (when previously no option was selected). Testing it locally, it looks like this change causes the placeholder to jump again. Can you take a look?

crisbeto: I see what you mean, but in this case the only solution that both animates on focus and doesn't jump when clicking would be to listen to the `transitionend` event on the placeholder and adding `always` after it is done. Wouldn't that be too much trouble for something this simple?

kara: Can you give it a shot? We can't really merge in a regression, especially for something as important as option selection.

crisbeto: I had a PoC running, but it wasn't particularly elegant. I'll try to do something cleaner.

crisbeto:
@kara I've reworked this one as discussed, can you take another look? Note that the CI failure seems to be due to Saucelabs timing out.

crisbeto:
@kara I've updated this one to work against the current master. Can you take another look?

kara:
@crisbeto I am the worst. Can you re-base again?

crisbeto:
Rebased and set up to work with the form field. Can you take a look @kara?

PR #4000, opened 2017-04-09
fix(grid-list): export MdGridTile. Fixes #2408

### LINKED ISSUE #2408: Export MdGridTile
#### feature request:
Support querying of tiles directly.

```ts
@ViewChildren(MdGridTile) tiles: QueryList<MdGridTile>;
myMethod() {
    this.tiles.changes.subscribe(changes => { ... })
}
```

```ts
@ViewChild(MdGridList) list: QueryList<MdGridList>;
myMethod() {
    this.list._tiles.changes.subscribe(changes => { ... })
}
```

I want to have a count of the *ngFor piped tiles

Using JSPM and SystemJS to import default JS files.

jelbourn:


src/lib/grid-list/index.ts
```diff
@@ -40,3 +40,4 @@ export class MdGridListModule {
 
 
 export * from './grid-list';
+export {MdGridTile} from './grid-tile';
```
mmalerba: why not just export *?

jelbourn: `tile.ts` has some exports that don't really belong in the public API

Issue #1786, opened 2016-11-09
update theming docs to be more clear about how to structure your theme file to be usable across custom components


fxck:
If I took this code https://github.com/angular/material2/blob/master/docs/theming.md#defining-a-custom-theme and put it into `_custom-theme.scss` and included it in my main styles file and then imported it into my components like as shown here https://github.com/angular/material2/blob/master/docs/theming-your-components.md#using-colors-from-a-pallete

it would generate unnecessary(since they are already in the main file) classes, like elevation, inside my components's encapsulated css.. which is bad.

The way I see it, your `_custom-theme.scss` should import and contain something along these lines 

```scss
@import '@angular/material/core/theming/theming';

$custom-color-primary: md-palette($md-blue);
$custom-color-accent:  md-palette($md-pink, A200, A100, A400);
$custom-color-warn:    md-palette($md-red);

$custom-theme: md-light-theme($color-primary, $color-accent, $color-warn);
```

the main style file should have 

```scss
@import '@angular/material/core/theming/all-theme';

@include md-core();

@import './custom-theme';

@include angular-material-theme($custom-theme);
```

thanks to this, you'd be able to import `_custom-theme.scss` anywhere and be able to use palette functions as well as your variables without it generating any duplicated material classes

wbhob:
This is fixed here: https://material.angular.io/guide/theming. This guide is really easy and I got started in 60 seconds.

Happy holidays!

dehypnosis:
why builder say 'No mixin named md-core'?

wbhob:
You need to update to the latest version, and put a scss file in your src dir, not @angular/material. Make sure to add the sass file to angularcli.json

dehypnosis:
i am on latest of master branch, and it seemes that the prefiex of names are changed from 'md' to 'mat'

dehypnosis:
```scss
/* angular material theming */
@import '~@angular/material/core/theming/all-theme';
@include mat-core();

$app-primary: mat-palette($mat-blue, A400, A100, A700);
$app-accent:  mat-palette($mat-pink, A200, A100, A400);
$app-warn:    mat-palette($mat-red);
$app-theme: mat-light-theme($app-primary, $app-accent, $app-warn);
@include angular-material-theme($app-theme);
```

now it works!

wbhob:
Interchangeable

dehypnosis:
really? thank you for your help!

wbhob:
Whatever works.

Wilson Hobbs
[...]

PR #4009, opened 2017-04-10
guide(theming): update custom theme section

fxck:
reasoning in this issue https://github.com/angular/material2/issues/1786


willshowell:
@fxck @jelbourn I *strongly* agree with this change. I lost more time than I'd like to admit tracking down why my bundles kept increasing in size, only to realize the theming guide was a slightly misleading.

jelbourn:
I've opened #4021 to try to address both this and a couple of other common issues. Let me know what you think.

PR #4024, opened 2017-04-10
fix(checkbox): do not set `indeterminate` when set `checked` programmatically

### LINKED ISSUE #2883: md-checkbox: Implement a three state checkbox
I would like to configure the md-checkbox like the following:

If its disabled and I click it, it should be enabled.
If its enabled and I click it, it should be indeterminate
If its indeterminate and I click it, it should be disabled

I tried to write a click-function, that does exactly that:

`<md-checkbox #check (click)="toggle(check)">Checked</md-checkbox>`

In the toggle-function I access check.indeterminate, checked etc. and set the values. But of course these values are overwritten by the default behaviour of the check box.
How would you implement that?

tinayuangao:
- Only set `indeterminate` when user manually click on the checkbox
- Fixes #2883

### DIFF
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -101,7 +101,7 @@ describe('MdCheckbox', () => {
       expect(inputElement.indeterminate).toBe(false);
     });
 
-    it('should set indeterminate to false when set checked', async(() => {
+    it('should set indeterminate to false when input clicked', async(() => {
       testComponent.isIndeterminate = true;
       fixture.detectChanges();
 
@@ -109,7 +109,7 @@ describe('MdCheckbox', () => {
       expect(inputElement.indeterminate).toBe(true);
       expect(testComponent.isIndeterminate).toBe(true);
 
-      testComponent.isChecked = true;
+      inputElement.click();
       fixture.detectChanges();
 
       fixture.whenStable().then(() => {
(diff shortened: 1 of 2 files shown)


src/lib/checkbox/checkbox.spec.ts
```diff
@@ -141,6 +141,38 @@ describe('MdCheckbox', () => {
...
+
+      expect(checkboxInstance.indeterminate).toBe(true);
+      expect(inputElement.indeterminate).toBe(true);
+      expect(testComponent.isIndeterminate).toBe(true);
+
+      testComponent.isChecked = true;
+      fixture.detectChanges();
+
+      fixture.whenStable().then(() => {
+        fixture.detectChanges();
```
crisbeto: Are the `detectChanges` calls necessary after `whenStable`? Also you may be able to avoid the nesting and the `whenStable` calls by wrapping the test with `fakeAsync`, instead of `async`.

src/lib/checkbox/checkbox.spec.ts
```diff
@@ -141,6 +141,31 @@ describe('MdCheckbox', () => {
 
     }));
 
+    it('should not set indeterminate to false when checked is set programmatically', async(() => {
```
crisbeto: You may be able to avoid using `async` here as well.

src/lib/checkbox/checkbox.ts
```diff
@@ -371,6 +361,14 @@ export class MdCheckbox implements ControlValueAccessor, AfterViewInit, OnDestro
     this._removeFocusRipple();
 
     if (!this.disabled) {
+      // When user manually click on the checkbox, `indeterminate` is set to false.
+      if (this._indeterminate) {
+        Promise.resolve().then(() => {
+          this._indeterminate = false;
+          this.indeterminateChange.emit(this._indeterminate);
```
jelbourn: Can go in a follow-up PR, but this should only emit when the value is different

PR #4028, opened 2017-04-11
build: dgeni add deprecated marker

### LINKED ISSUE #3981: Add marker for "deprecated" to generated API docs
Nothing in the current dgeni API doc generation marks the doc items as deprecated. We should mark this in the categorizer and then add a css class for classes and members in the HTML templates.

Then, as a follow-up, the class can be given a style on [material.angular.io](https://github.com/angular/material.angular.io/)

DevVersion:
* Adds a class in docs templates that can be used to distinguish between deprecated and stable APIs.

Fixes #3981.

### DIFF
--- a/tools/dgeni/processors/categorizer.js
+++ b/tools/dgeni/processors/categorizer.js
@@ -68,14 +68,11 @@ module.exports = function categorizer() {
   }
 
   /**
-   * Decorates public exposed docs. Creates a property with CSS classes that will be 
-   * added to the template.
+   * Decorates public exposed docs. Creates a property on the doc that indicates whether
+   * the item is deprecated or not.
    **/
   function decoratePublicDoc(doc) {
-    // Specific classes that will can added to the Dgeni doc template.
-    doc.docClasses = [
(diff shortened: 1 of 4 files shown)


tools/dgeni/processors/categorizer.js
```diff
@@ -55,12 +58,25 @@ module.exports = function categorizer() {
...
     propertyDoc.directiveOutputAlias = getDirectiveOutputAlias(propertyDoc);
   }
+
+  /**
+   * Decorates public exposed docs. Creates a property with CSS classes that will be 
+   * added to the template.
+   **/
+  function decoratePublicDoc(doc) {
+    // Specific classes that will can added to the Dgeni doc template.
+    doc.docClasses = [
```
jelbourn: This step shouldn't really be concerned with css classes; it should just mark up the doc object with appropriate properties. Then, in the template we can use a nunjucks expression to apply the class

DevVersion: Done!

PR #4039, opened 2017-04-11
Add support for `color` attribute to radio

rtrompier:
I have added an Input COLOR to allow tu user to choose the color like the checkbox component.
The color can be `primary`, `accent`, or `warn`.

screen shot 2017-04-11 at 21 40 13
screen shot 2017-04-11 at 21 40 23
screen shot 2017-04-11 at 21 40 30


src/lib/radio/_radio-theme.scss
```diff
@@ -12,28 +12,64 @@
...
+      }
+
+      .mat-radio-inner-circle {
+        background-color: mat-color($warn);
+      }
+
+      .mat-radio-ripple .mat-ripple-element {
+        background-color: mat-color($warn, 0.26);
+      }
+    }
```
DevVersion: I have the feeling that we could move that into a mixin? So we don't have to repeat code three times.

rtrompier: Yes sure ;) 
I will create a mixin, you're right.

rtrompier:
@DevVersion Added mixin & Default color (accent)

jelbourn:
@rtrompier could you rebase this PR?

rtrompier:
Should be good ;)

src/lib/radio/_radio-theme.scss
```diff
@@ -9,33 +9,57 @@
   $background: map-get($theme, background);
   $foreground: map-get($theme, foreground);
 
-  .mat-radio-outer-circle {
-    border-color: mat-color($foreground, secondary-text);
-  }
+    .mat-radio-button {
+        //default color (accent)
```
jelbourn: Our indentation is +2, not +4

src/lib/radio/_radio-theme.scss
```diff
@@ -9,33 +9,57 @@
...
+        &:not(.mat-primary):not(.mat-accent):not(.mat-warn) {
```
jelbourn: I don't think this `:not` expression should be necessary; the radio should just have `accent` be the default color.

src/lib/radio/_radio-theme.scss
```diff
@@ -9,33 +9,57 @@
...
+        &.mat-warn {
+            @include radio-theme-color($warn, $foreground);
+        }
+    }
+}
 
-  .mat-radio-ripple .mat-ripple-element {
-    background-color: mat-color($accent, 0.26);
-  }
+@mixin radio-theme-color($color, $foreground) {
```
jelbourn: This mixin is going to generate a lot of css output; only styles that are directly affected by the palette swap should be defined within.

src/lib/radio/radio.ts
```diff
@@ -396,6 +406,9 @@ export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
   /** Whether the ripple effect on click should be disabled. */
   private _disableRipple: boolean;
 
+  /** Current color. Can be `primary`, `accent`, or `warn`. */
+  private _color: string;
```
jelbourn: The type should use the string literal values
```ts
private _color: 'primary' | 'accent' | 'warn';
```

DevVersion:
@rtrompier Thanks for your work on this PR. It looks like your PR is no longer up-to-date and there isn't any activity here.

I'm going to address the color binding for radios in #5068

src/lib/radio/radio.ts
Polyterative: Is there any way to use straight color values like for example `#00FF00` or `rgb(255,100,0)` from HTML ?

PR #4044, opened 2017-04-11
feat(directionality): a provider to get the overall directionality

### LINKED ISSUE #3600: Introduce Directionality injectable
Unlike suggested on #1719 to traverse up the DOM and look for any dir attributes
@jelbourn and I decided to only support `html` and `body` tags or the [Dir directive](https://github.com/angular/material2/blob/master/src/lib/core/rtl/dir.ts) to get the actual direction.

EladBezalel:
- Looks at the `html` and `body` elements for `dir` attribute and sets the Directionality service value to it
- Whenever someone would try to inject Directionality - if there's a Dir directive up the dom tree it would be provided

fixes #3600


src/lib/core/rtl/directionality.ts
```diff
@@ -0,0 +1,18 @@
+import {EventEmitter, Injectable} from '@angular/core';
+import {LayoutDirection} from './dir';
+
+/**
+ * A provider to get the direction outside of angular on the `html` or `body` elements
```
jelbourn: "A provider" isn't really accurate (it's just a class that's _provided_). It can just be something like
```
The directionality (LTR / RTL) context for the application (or a subtree of it). 
Exposes the current direction and a stream of direction changes.
```

src/lib/core/rtl/directionality.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+@Injectable()
+export class Directionality {
+  value: LayoutDirection;
+  change: EventEmitter<void> = new EventEmitter<void>();
+
+  constructor() {
+    this.value =
+      (document.body.getAttribute('dir') ||
```
jelbourn: We still need to account for `dir="auto"`. It looks like `HTMLElemenet.dir` is also `"auto"` when that's set to the attribute, but `getComputedStyle` return either `"ltr"` or `"rtl"`. I'd like to avoid using `getComputedStyle` here, though, since we're already calling it for the theming check. 

Let's just add a `TODO` to handle auto with this background info.

jelbourn: This should also check for the existence of `document` before dereferencing it (for platform-server)

src/lib/core/rtl/dir.ts
```diff
@@ -8,36 +8,42 @@ import {
   EventEmitter
 } from '@angular/core';
 
+import {Directionality} from './directionality';
+export {Directionality} from './directionality';
+
 export type LayoutDirection = 'ltr' | 'rtl';
```
EladBezalel: Makes sense

jelbourn:
@EladBezalel ready for another pass?

EladBezalel:
Yep

src/lib/core/bidi/directionality.ts
```diff
@@ -0,0 +1,43 @@
...
+ * The directionality (LTR / RTL) context for the application (or a subtree of it).
+ * Exposes the current direction and a stream of direction changes.
+ */
+@Injectable()
+export class Directionality {
+  value: Direction = 'ltr';
+  public change = new EventEmitter<void>();
+
+  constructor() {
+    if (document) {
```
jelbourn: I think you need `typeof document` here in order to not error on Angular Universal. This will need to eventually be augmented with using universal's document service.

src/lib/core/bidi/directionality.ts
```diff
@@ -0,0 +1,43 @@
...
+    // TODO: handle 'auto' value -
+    // We still need to account for dir="auto".
+    // It looks like HTMLElemenet.dir is also "auto" when that's set to the attribute,
+    // but getComputedStyle return either "ltr" or "rtl". avoiding getComputedStyle for now
+    // though, we're already calling it for the theming check.
+    this.value =
+      (document.body.getAttribute('dir') ||
```
jelbourn: Use the `dir` property instead of `getAttribute`? I believe they should be equivalent (but would be shorter)

src/lib/core/bidi/directionality.ts
```diff
@@ -0,0 +1,43 @@
...
+export type Direction = 'ltr' | 'rtl';
+
+/**
+ * The directionality (LTR / RTL) context for the application (or a subtree of it).
+ * Exposes the current direction and a stream of direction changes.
+ */
+@Injectable()
+export class Directionality {
+  value: Direction = 'ltr';
+  public change = new EventEmitter<void>();
```
jelbourn: Need unit tests for the following (creating a new `bidi.spec.ts`):
- [ ] Directionality reads `dir` from the html element if not specified on the body
- [ ] Directionality reads `dir` from the body even it is also specified on the html element
- [ ] Directionality defaults to `ltr` if nothing is specified on either body or the html element
- [ ] `Dir` provides itself as `Directionality` 
- [ ] `Dir` emits a change event when the value changes

EladBezalel: How before the tests i can add the dir attribute to the html and the body elements?

jelbourn: In the top-level `beforeEach` grab the current `dir` on each and then in the top-level `afterEach` restore it.

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
+import {async, fakeAsync, TestBed, tick} from '@angular/core/testing';
+import {Component, getDebugNode} from '@angular/core';
+import {Dir, Directionality} from './index';
+
+function initDir () {
+  document.documentElement.dir = '';
+  document.body.dir = '';
+}
```
jelbourn: Call this something like `clearDocumentDirAttributes`? Also move to the bottom of the file, after the tests but before the test components

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+describe('Directionality', () => {
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      providers: [Directionality],
+      declarations: [TestAppWithoutDir]
+    });
+
+    TestBed.compileComponents();
+
+    initDir();
```
jelbourn: There needs to be an `afterEach` that restores the `dir` attributes to whatever they were before.

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    tick();
+
+    expect(testComponent.dir.value).toBe('ltr');
+    expect(fixture.componentInstance.changeCount).toBe(1);
+  }));
+});
+
+/** Test component without Dir. */
+@Component({
+  selector: 'test-app-without-dir',
```
jelbourn: You don't need a selector for components instantiated via `TestBed`

src/lib/core/bidi/directionality.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor() {
+    if (typeof document) {
```
jelbourn: `typeof document === 'object' && !!document`.

We have an `isBrowser` property on `Platform` now for things like this, but we can adjust that later (since it involves bringing in `PlatformModule`)

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  direction = 'rtl';
+  changeCount = 0;
+}
+
+/** Test component with Dir directive. */
+@Component({
+  selector: 'dir-test',
+  template: `<div></div>`
+})
+class DirTest {
```
jelbourn: Call this `InjectsDirectionality` with selector `injects-directionality`. You can also completely replace `TestAppWithoutDir` with this component.

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+/** Test component without Dir. */
+@Component({
+  selector: 'test-app-with-dir',
+  template: `
+    <div [dir]="direction" (dirChange)="changeCount= changeCount + 1">
+      <dir-test></dir-test>
+    </div>
+  `
+})
+class TestAppWithDir {
```
jelbourn: Can just call this `ElementWithDir`

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  it('should default to ltr if nothing is specified on either body or the html element', () => {
+    let fixture = TestBed.createComponent(TestAppWithoutDir);
+    let testComponent = fixture.debugElement.componentInstance;
+
+    expect(testComponent.dir.value).toBe('ltr');
+  });
+});
+
+describe('Dir directive', () => {
```
jelbourn: There should be one top-level `describe` block that does `configureTestingModule`, importing `BidiModule`. You can then have two sub `describe` blocks ("without dir directive", "with dir directive"). Then you can have a separate `beforeEach` that does the `createComponent` in each and the query for the directive you're looking for (to reduce repetition).

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [Dir, TestAppWithDir, DirTest]
```
jelbourn: FYI you shouldn't put directives you're testing directly into the `declarations` of the testing module; better to import the `NgModule` they live in for real (so `BidiModule` here)

src/lib/core/bidi/directionality.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    TestBed.compileComponents();
+
+    initDir();
+  }));
+
+  it('should provide itself as Directionality', () => {
+    let fixture = TestBed.createComponent(TestAppWithDir);
+    let testComponent =
+      getDebugNode(fixture.nativeElement.querySelector('dir-test')).componentInstance;
```
jelbourn: Can use `By.directive` instead:
```ts
const injectedDirectionality = 
    fixture.debugElement.query(By.directive(InjectsDirectionality)).dir;
```
(here and elsewhere)

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -65,7 +65,7 @@ describe('MdAutocomplete', () => {
 
           return {getContainerElement: () => overlayContainerElement};
         }},
-        {provide: Dir, useFactory: () => ({value: dir})},
+        {provide: Directionality, useFactory: () => ({value: dir})},
```
jelbourn: Rather than changing all of these tests to add the `Directionality`, we now have a `MdCommonModule` that all of the components import. You can add `BidiModule` to the imports of the common module.

jelbourn:
@EladBezalel just needs rebase

jelbourn:
I rebased it

jelbourn:
Hmm, changing the actual `dir` in the tests is causing other tests to fail on Safari.

jelbourn:
Changed `Directionality` to inject the document and use a fake in the unit tests

jelbourn:
Caretaker note: some people were importing `LayoutDirection`, so we'll have to update them as part of the sync

PR #4046, opened 2017-04-12
build: create example package

DevVersion:
* Initial refactoring that allows Material to build multiple packages using a shared logic.


tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,52 @@
+import {task, watch} from 'gulp';
+import {join} from 'path';
+import {main as tsc} from '@angular/tsc-wrapped';
+import {SOURCE_ROOT, DIST_BUNDLES, DIST_EXAMPLES} from '../constants';
+import {sequenceTask, sassBuildTask, copyTask, triggerLivereload} from '../util/task_helpers';
+import {createRollupBundle} from '../util/rollup-helper';
+import {transpileFile} from '../util/ts-compiler';
+import {buildModuleEntry, composeRelease} from '../util/package-build';
+import {ScriptTarget, ModuleKind} from 'typescript';
+import {writeFileSync} from 'fs';
```
jelbourn: Looks like a lot of unused imports in this file (and library.ts / release.ts) now

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * TypeScript compilation tasks. Tasks are creating ESM, FESM, UMD bundles for releases.
+ **/
+
+task('examples:build:esm', () => tsc(tsconfigPath, {basePath: examplesRoot}));
+task('examples:build:bundles', () => buildModuleEntry(examplesMain, 'examples'));
+
+/**
+ * Asset tasks. Copying and inlining CSS, HTML files into the ESM output.
+ **/
```
jelbourn: Only one `*` at the end of the JsDoc block

tools/gulp/util/package-build.ts
```diff
@@ -0,0 +1,111 @@
...
+import {ScriptTarget, ModuleKind} from 'typescript';
+import {sync as glob} from 'glob';
+import {
+  writeFileSync, copySync, mkdirpSync, readFileSync
+} from 'fs-extra';
+
+// There are no type definitions available for these imports.
+const uglify = require('uglify-js');
+
+export function composeRelease(packageName: string) {
```
jelbourn: Add function description? It could be good to document each of thing things that get packaged into a release.

tools/gulp/util/package-build.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/** Creates a metadata file that re-exports the metadata bundle inside of the typings. */
+function createMetadataFile(packageDir: string, packageName: string) {
+  const metadataReExport =
+      `{"__symbolic":"module","version":3,"metadata":{},"exports":[{"from":"./typings/index"}]}`;
+  writeFileSync(join(packageDir, `${packageName}.metadata.json`), metadataReExport, 'utf-8');
+}
+
+function inlineMetadata(packagePath: string) {
```
jelbourn: How about `inlineMetadataResources`?
Also add description like
```ts
/** Inlines html and css resources into all metadata.json files in the given package path. */
```

src/examples/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@angular/material-examples",
```
jelbourn: Let's make this `angular-material-examples` since it's not something we should ever publish under `@angular`

tools/gulp/util/package-build.ts
DevVersion: `inlineMetadataResources` is another method we import from the `utils`. Therefore I used a different name here. Adding a comment anyways

jelbourn: Then how about `inlineAllMetadataResourceInPath`?

Issue #4047, opened 2017-04-12
MdGridList cannot switch between ratio and fixed row height


arlowhite:
#### Bug:

Switching from ratio to fixed height causes tiles to have larger heights than they should.

This is because **RatioTileStyler** uses _paddingTop_ and **FixedTileStyler** uses _height_ to set tile heights. But when switching between them, the old styles are not cleared.

#### What is the expected behavior?

Users should be able change rowHeight between ratio, fixed, and fit dynamically.

#### What is the current behavior?

Switching between fixed and ratio rowHeight modes results in tiles that are larger than they should be due to left-over styles.
Have not tested 'fit' mode.

#### What are the steps to reproduce?

I reproduced this with unit tests in grid-list.spec.ts
https://github.com/arlowhite/material2/tree/grid-rowheight-bug

#### What is the use-case or motivation for changing an existing behavior?

Users can hit this bug easily, even if they aim to stick with fixed mode.
If their Component loads with rowHeight undefined, grid will default to "1:1"
Then when they set a number to rowHeight, this bug will occur.

#### Which versions of Angular, Material, OS, browsers are affected?

Affects 2.0.0-beta.3
material2 master 9d719c544e0476b26c1fe0918b1c9c6bb9cd0ec0

Angular 4.0.1

#### Is there anything else we should know?

I will submit a PR with the unit test.

PR #4048, opened 2017-04-12
add grid-list.spec.ts test for switching between ratio and fixed height

arlowhite:
This creates a unit tests that demonstrates bug #4047 

EDIT: and also tries to fix the bug. However, need code review before merging.

In theory, all permutations of switching between ratio, fixed, and fit should be tested, but that's probably overkill.

### DIFF
--- a/src/lib/grid-list/grid-list.spec.ts
+++ b/src/lib/grid-list/grid-list.spec.ts
@@ -262,6 +262,7 @@ describe('MdGridList', () => {
   it('should support switching between ratio and fixed row height', () => {
     // width is 400px, 1 column
     let fixture = TestBed.createComponent(GirdListWithRowHeightRatio);
+    const gridList = fixture.debugElement.query(By.directive(MdGridList));
     // initial ratio rowHeight
     fixture.componentInstance.rowHeight = '2:1';
     fixture.detectChanges();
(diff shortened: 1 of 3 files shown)


arlowhite:
I came-up with a fix that passes the test, but someone more familiar with TileStylers code should probably look at this bug.

src/lib/grid-list/grid-list.ts
```diff
@@ -110,6 +110,15 @@ export class MdGridList implements OnInit, AfterContentChecked {
 
   /** Creates correct Tile Styler subtype based on rowHeight passed in by user */
   private _setTileStyler(): void {
+    if (this._tileStyler) {
+      // leftover styles must be cleared before setting a new TileStyler
+      this._tiles.forEach((tile, index) => {
+        this._tileStyler.clearRowStyles(tile);
+      });
+      // clear styles set on grid by _layoutTiles()
+      this._setListStyle(['paddingBottom', '0px']);
```
arlowhite: This code isn't as clean because it does not defer to the TileStyler implementation to clear the MdGridList styles.

But I wanted to just add a quick fix before adding another method to TileStyler.

I was thinking of changing clearRowStyles to:
`clearTileStyles(tiles: MdGridTile[]):  [string, string]`

which would return the style to clear the grid styling.

src/lib/grid-list/grid-list.spec.ts
```diff
@@ -294,12 +338,12 @@ class GridListWithUnspecifiedRowHeight { }
...
 @Component({template: `
     <div style="width:400px">
-      <md-grid-list cols="1" [rowHeight]="heightRatio">
+      <md-grid-list cols="1" [rowHeight]="rowHeight">
         <md-grid-tile></md-grid-tile>
       </md-grid-list>
     </div>`})
 class GirdListWithRowHeightRatio {
-  heightRatio: string;
+  rowHeight: string;
```
arlowhite: Changed from **heightRatio** to **rowHeight** because it more clearly indicates direct binding to rowHeight.

I'm reusing the GirdListWithRowHeightRatio test component. Is there a rule that there should only be one test Component per test; should I create a new component for the test I added?

jelbourn:

PR #4055, opened 2017-04-12
feat(screenshot): Control panel for Screenshot test

tinayuangao:


tools/screenshot-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "version": "0.0.0",
+  "license": "MIT",
+  "angular-cli": {},
+  "scripts": {
+    "ng": "ng",
+    "start": "ng serve",
+    "lint": "tslint \"src/**/*.ts\" --project src/tsconfig.json --type-check && tslint \"e2e/**/*.ts\" --project e2e/tsconfig.json --type-check",
+    "test": "ng test",
+    "pree2e": "webdriver-manager update --standalone false --gecko false",
+    "e2e": "protractor"
```
DevVersion: I think we can remove those scripts?

tools/screenshot-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "rxjs": "^5.0.1",
+    "ts-helpers": "^1.1.1",
+    "zone.js": "^0.8.4"
+  },
+  "devDependencies": {
+    "@angular/cli": "^1.0.0",
+    "@angular/compiler-cli": "^4.0.0",
+    "@types/jasmine": "2.5.38",
+    "@types/node": "^6.0.42",
+    "codelyzer": "~2.0.0-beta.1",
```
DevVersion: Codelyzer can be removed I think?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,204 @@
...
+            this.screenshotResult.travis = childValue;
+            break;
+          case 'results':
+            this._processResults(childSnapshot);
+            break;
+          case 'approved':
+            this.screenshotResult.approved = childValue;
+            break;
+        }
+        counter ++;
```
DevVersion: Just for the future: There shouldn't be any space between the `++` . Also the `if` below. Strict checking is really recommended.

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  getGithubStatus() {
+    let url =
+      `https://api.github.com/repos/${config.repoSlug}/commits/${this.screenshotResult.sha}/status`;
+    return new Promise((resolve) => {
+      request({
+        url: url,
+        method: 'GET',
+      }, function (error: any, response: any) {
```
DevVersion: Use arrow function? And probably move the `{ url, method: 'GET' }` into a single line?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,204 @@
...
+        let statusResponse = JSON.parse(response.body);
+        for (let status of statusResponse.statuses) {
```
DevVersion: How about `response.statuses.some(status => status.context === 'Screenshot Tests')`?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
+import {Injectable} from '@angular/core';
+import * as firebase from 'firebase';
+
+const request = require('request');
+const config = require('../config.json');
+
+
+export class ScreenshotResult {
```
jelbourn: Can `ScreenshotResult` go into its own file?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  /** PR information: the pull request number */
+  prNumber: string;
+  /** PR information: the sha of the pull request commit */
+  sha: string;
+  /** PR information: The travis job ID */
+  travis: string;
```
jelbourn: `travisJobId`

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  /** Test result, the test names */
+  testnames: string[];
```
jelbourn:
```ts
/** The names of the tests that have screenshot results */
testNames: string[];
```

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  /** Test result: passed or failed. The value can be true if test failed but PR approved by user */
+  result: boolean;
```
jelbourn: `allTestsPassedOrApproved`

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  /** Test results: passed or failed for each test */
+  results: Map<string, boolean> = new Map<string, boolean>();
```
jelbourn: How about `testResultsByName`?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Test approved: whether the test images are copied to goldens.
+   * The value is the commit SHA and date/time of approval.
+   */
+  approved: string;
+
+  githubStatus: any;
```
jelbourn: Can this be more specific than `any`?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  /** Test result, the test names */
+  testnames: string[];
+  /** Test result: passed or failed. The value can be true if test failed but PR approved by user */
+  result: boolean;
+  /** Test results: passed or failed for each test */
+  results: Map<string, boolean> = new Map<string, boolean>();
+  /**
+   * Test approved: whether the test images are copied to goldens.
+   * The value is the commit SHA and date/time of approval.
```
jelbourn: Why are the commit and the date/time serialized into a string for this?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+   */
+  approved: string;
+
+  githubStatus: any;
+
+  /** Viewing mode, can be flip, diff, side */
+  mode: 'diff' | 'side' | 'flip' = 'diff';
+  /** Viewing flipping, whether the image is test image or golden image */
+  flipping: boolean = false;
```
jelbourn: How about
```ts
/** When in "Flip" mode, whether the test result or the gold screenshot is being shown. */
isFlipped = false;
```

tools/screenshot-test/src/app/firebase.service.ts
jelbourn: Add class description?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  /** Viewing collapsed: whether the result should be collapsed/expanded */
+  collapse: boolean[];
```
jelbourn: Why is `collapse` and array why `flipping` is a single boolean?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  setCollapse(value: boolean) {
+    if (this.collapse) {
+      for (let i = 0; i < this.collapse.length; i++) {
+        this.collapse[i] = value;
+      }
+    }
+  }
+
+  get prLink() {
```
jelbourn: Use `Url` instead of `Link`?

tools/screenshot-test/src/app/result/result.component.css
```diff
@@ -0,0 +1,58 @@
+img {
```
jelbourn: We should have an extra lint task that runs stylelint for the dashboard

tools/screenshot-test/src/app/result/result.component.html
```diff
@@ -0,0 +1,39 @@
+<md-card>
+  <md-card-header>
+    <md-card-title>
+      <md-icon (click)="toggleCollapse()">{{collapseIcon}}</md-icon>
+      {{testName}}
+    </md-card-title>
+    <md-card-subtitle [ngClass]="resultText.toLowerCase()">{{resultText}}</md-card-subtitle>
+  </md-card-header>
+  <md-card-content *ngIf="collapse == false">
```
jelbourn: `!collapse`?

tools/screenshot-test/src/app/result/result.component.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.image-section {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.side .diff-section {
```
jelbourn: Can `.side` be named more specifically?

tools/screenshot-test/src/app/result/result.component.ts
```diff
@@ -0,0 +1,116 @@
...
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
+  Input,
+  Output,
+  EventEmitter
+} from '@angular/core';
+import {FirebaseService} from '../firebase.service';
+
+@Component({
```
jelbourn: Add class description?

tools/screenshot-test/src/app/result/result.component.ts
```diff
@@ -0,0 +1,116 @@
...
+  get collapseIcon() {
+    return this.collapse ? 'keyboard_arrow_right' : 'keyboard_arrow_down';
+  }
+
+   /** Filename, the test file name. */
+  @Input()
+  get filename() {
+    return this._filename;
+  }
+  set filename(filename: string) {
```
jelbourn: Can this be changed to something like `setFilenameAnd...` vs. a setter with side-effects?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  get travisLink() {
+    return `https://travis-ci.org/${config.repoSlug}/jobs/${this.travis}`;
+  }
+}
+
+
+@Injectable()
+export class FirebaseService {
```
alxhub: This is actually using Firebase directly, instead of going via AngularFire.

@tinayuangao, perhaps consider using AngularFire instead?

tools/screenshot-test/src/app/firebase.service.ts
tinayuangao: Each test result may be collapsed or expanded, so it's an array. 

`isFlipped` is the way to view image, and all the images are flipped at the same time, so we only need a single boolean.

tools/screenshot-test/src/app/firebase.service.ts
tinayuangao: Changed to `approved/${SHA}/${dateTime}` so we can approve multiple times. At first the value is a single boolean, and can be set only once for each PR which doesn't work when we have multiple commits and need to approve a commit more than once (for example when firebase functions failed).

tools/screenshot-test/src/app/result/result.component.ts
tinayuangao: Removed the getter since we don't need it. Added a function `setFilenameAndFetchImages`.

tools/screenshot-test/src/app/firebase.service.ts
tinayuangao: @alxhub Does AngularFire support access to Firebase Storage? We need to fetch some image files from FirebaseStorage.

src/lib/core/style/focus-origin-monitor.ts
```diff
@@ -33,7 +33,7 @@ type MonitoredElementInfo = {
 /** Monitors mouse and keyboard events to determine the cause of focus events. */
 @Injectable()
 export class FocusOriginMonitor {
-  /** The focus origin that the next focus event is a result of. */
+  /** The focus origin that the next focus event is a allTestsPassedOrApproved of. */
```
jelbourn: Looks like your find-and-replace was a bit overzealous

tinayuangao: Reverted the changes.

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,152 @@
...
+
+/** The service to fetch data from firebase database */
+@Injectable()
+export class FirebaseService {
+
+  /** The current user */
+  user: any;
+
+  /** The screenshot results */
+  screenshotResult: ScreenshotResult;
```
jelbourn: Since this captures the result for _all_ screenshots in the PR, how about calling this class `ScreenshotResultsSummary` (and the member `screenshotResultsSummary`)?

tools/screenshot-test/src/app/screenshot-result.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /** The names of the tests that have screenshot results */
+  testNames: string[];
+  /**
+   * Test result: passed or failed. The value can be true if test failed but
+   * PR approved by user
+   */
+  allTestsPassedOrApproved: boolean;
+  /** Test results: passed or failed for each test */
+  testResultsByName: Map<string, boolean> = new Map<string, boolean>();
```
jelbourn: Instead of manually adding to `testNames` and `testResultsByName` (since nothing ensures they're both updated at the same time), add a method like `addTestResults(name, value)`?

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    return firebase.database().ref('screenshot').child('reports')
+      .child(this.screenshotResult.prNumber);
+  }
+
+  _storageRef(): firebase.storage.Reference {
+    return firebase.storage().ref('screenshots').child(this.screenshotResult.prNumber);
+  }
+
+  /** Put the testResultsByName in screenshotReuslt */
+  _processResults(childSnapshot: firebase.database.DataSnapshot) {
```
jelbourn: Can this function name be more be specific about what it does? Is it something like `readResultsFor___`? I'm not exactly sure from reading it what it's recording

(similar comment for other `processX` methods)

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,152 @@
+import {Injectable} from '@angular/core';
+import * as firebase from 'firebase';
+
+const request = require('request');
```
jelbourn: Why use this instead of Angular's `Http`?

tools/screenshot-test/src/app/screenshot-result.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Test approved: whether the test images are copied to goldens.
+   * The value is the date/time of approval of current commit. Null if never approved.
+   */
+  approvedTime: number;
+
+  githubStatus: any;
```
jelbourn: Any way for this to have a more specific type?
(potentially even `object` or `{}` if we can't be more specific)

tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -0,0 +1,152 @@
...
+
+
+  /** Get the firebase storage test image folder ref */
+  testRef(): firebase.storage.Reference {
+    return this._storageRef().child('test');
+  }
+
+  /** Get the firebase storage diff image folder ref */
+  diffRef(): firebase.storage.Reference {
+    return this._storageRef().child('diff');
```
tinayuangao: It's trivial to get the reference.

tinayuangao:
Comments addressed. Please take another look. Thanks for review!

jelbourn:
Just add the "merge-ready" label when ready

PR #4059, opened 2017-04-12
build: detect todos in public comments

### LINKED ISSUE #4026: Add lint check to prevent having TODO in public JsDoc comments

DevVersion:
* Creates a naive rule that checks all multi-line comments and reports failures once it detects TODOs inside of it.
* This means that TODOs always need to placed inside of single-line comments.

Fixes #4026

### DIFF
--- a/tools/tslint-rules/noExposedTodoRule.js
+++ b/tools/tslint-rules/noExposedTodoRule.js
@@ -1,0 +1,32 @@
+const ts = require('typescript');
+const utils = require('tsutils');
+const Lint = require('tslint');
+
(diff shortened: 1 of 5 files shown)


tools/tslint-rules/noExposedTodoRule.js
```diff
@@ -0,0 +1,32 @@
+const ts = require('typescript');
+const utils = require('tsutils');
+const Lint = require('tslint');
+
+const ERROR_MESSAGE = 'Todo inside of the comment might be published in the docs.';
```
jelbourn:
```
A TODO may only appear in inline (//) style comments. This is meant to prevent
a TODO from beig accidentally included in any public API docs.
```

PR #4064, opened 2017-04-12
build: publish material-examples package

### LINKED ISSUE #3518: Add gulp build tasks for examples
We should build the examples as a component library similar to how we build the material library itself and include the build output in the bundle that's published to material2-docs-content.

@DevVersion would you be able to tackle this?

DevVersion:
* Publish the `material-examples` package in the `material2-docs-content` repository.

Fixes #3518.

### DIFF
--- a/tools/gulp/tasks/docs.ts
+++ b/tools/gulp/tasks/docs.ts
@@ -88,7 +88,7 @@ task('highlight-examples', () => {
     path.basename = `${path.basename}-${extension}`;
   };
 
-  return src('src/examples/**/*.+(html|css|ts)')
+  return src('src/material-examples/**/*.+(html|css|ts)')
       .pipe(flatten())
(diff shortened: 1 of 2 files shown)


tools/gulp/tasks/docs.ts
```diff
@@ -111,7 +111,7 @@ task('minified-api-docs', ['api-docs'], () => {
 
 /** Copies example sources to be used as plunker assets for the docs site. */
 task('plunker-example-assets', () => {
-  src(path.join(SOURCE_ROOT, 'examples', '**/*'))
+  src(path.join(SOURCE_ROOT, 'material-examples', '**/*'))
```
jelbourn: While you're in here, can you remove the extra `plunker/` that's being written to the docs content?

DevVersion: Done

PR #4076, opened 2017-04-13
chore(tabs): revert workaround for angular < 2.3

crisbeto:
Reverts a workaround that was preventing the tabs from crashing the user's browser in Angular < 2.3, due to an animations issue. This was introduced initially in #2411.


jelbourn:
Presubmit passes, just need rebase

PR #4082, opened 2017-04-13
docs(guides): add guide for using compatibility mode

### LINKED ISSUE #4018: Create guide for compatibility mode
Nothing in the documentation current describes how to use compatibility mode and what it does. We should create a guide that captures this information.

epelc:
Adds a some info about compatibility mode and how to use it.

Fixes #4018

### DIFF
--- a/guides/compatibility-mode.md
+++ b/guides/compatibility-mode.md
@@ -1,0 +1,19 @@
+## What is compatibility mode?
+Compatibility mode for `@angular/material` allows the components to exist side-by-side with AngularJS Material components without any CSS collisions.
+
+## What does compatibility mode do?
+When enabled, compatibility mode enforces that all template APIs use the prefix `mat-` instead of `md-`. This will prevent any CSS from AngularJS Material from affecting the components in `@angular/material`.
+
+## How is compatibility mode enabled?
+Import `NoConflictStyleCompatibilityMode` into your application's root `NgModule`.
+
+
+## Example
+
+```html
+<!-- Regular mode -->
+<button md-button mdTooltip="With a tooltip">Regular button</button>
+
+<!-- In compatibility mode -->
+<button mat-button matTooltip="With a compatibility mode tooltip">Compatibility Mode button</button>
+```


guides/compatibility-mode.md
```diff
@@ -0,0 +1,32 @@
+# Compatibility Mode
+Compatibility mode allows `@angular/material` to work with [Angular Material for AngularJS](https://material.angularjs.org/latest/).
```
jelbourn: We're just calling it "AngularJS Material" now :smile:

guides/compatibility-mode.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Upgrade Process Overview
+Below is a general overview of the upgrade process for an AngularJS application which uses [Angular Material for AngularJS](https://material.angularjs.org/latest/) and wishes to upgrade to Angular 2+ with `@angular/material`.
+1. Setup `@angular/upgrade/static` and bootstrap Angular in your AngularJS project
+2. Setup and enable compatibility mode for `@angular/material`
+3. Convert all components over to Angular and `@angular/material`
```
jelbourn: `all components` > `desired components`?

guides/compatibility-mode.md
```diff
@@ -0,0 +1,32 @@
...
+4. Remove Angular Material for AngularJS
+5. Disable compatibility mode
+  - Find/replace `mat-` with `md-`
+  - Remove `NoConflictStyleCompatibilityMode`
+
+
+## Upgrade Checklist
+- [ ] Switch all `@angular/material` component prefixes in templates from `md-` to `mat-`
+  - See [this](https://github.com/angular/material2/blob/master/src/lib/core/compatibility/compatibility.ts#L33-L150) for a list of all components you need to rename.
```
jelbourn: These line numbers won't be stable over time; I think just have an example here would be good

guides/compatibility-mode.md
jelbourn: I would open this with

```
Compatibility mode for `@angular/material` allows the components to exist side-by-side with AngularJS Material components without any CSS collisions.
```

As for the structure of the doc, I think the it should be something like
```md
## What is compatibility mode?
Compatibility mode for `@angular/material` allows the components to exist side-by-side with AngularJS Material components without any CSS collisions.

## What does compatibility mode do?
When enabled, compatibility mode enforces that all template APIs use the prefix `mat-` instead of `md-`. This will prevent any CSS from AngularJS Material from affecting the components in `@angular/material`.

## How is compatibility mode enabled?
Simply import `NoConflictStyleCompatibilityMode` into your application's root `NgModule`. 
```

guides/compatibility-mode.md
```diff
@@ -0,0 +1,32 @@
+# Compatibility Mode
+Compatibility mode allows `@angular/material` to work with [Angular Material for AngularJS](https://material.angularjs.org/latest/).
+
+## Upgrade Process Overview
```
jelbourn: I would omit this section, as I feel it should mostly be captured by upgrade guidance outside of material

epelc: Agreed

guides/compatibility-mode.md
epelc: Removed the whole section

epelc:
@jelbourn should be good now

guides/compatibility-mode.md
```diff
@@ -0,0 +1,8 @@
+## What is compatibility mode?
+Compatibility mode for `@angular/material` allows the components to exist side-by-side with AngularJS Material components without any CSS collisions.
+
+## What does compatibility mode do?
+When enabled, compatibility mode enforces that all template APIs use the prefix `mat-` instead of `md-`. This will prevent any CSS from AngularJS Material from affecting the components in `@angular/material`.
```
jelbourn: Maybe add an example to this section that shows, say, a button with a tooltip before and after compatibility mode is enabled?

epelc: Added

PR #4084, opened 2017-04-13
fix(input): incorrect height with autosize

### LINKED ISSUE #4070: mdTextareaAutosize cuts off text on init
bug

All text should be visible in textarea that uses mdTextareaAutosize

Lines missing in textarea

Provide a multiline default value for the textarea
http://plnkr.co/edit/SVMcC9IhIagP0xg99USQ?p=preview

It's a bug

all versions

DevVersion:
Currently when using the `mdTextareaAutosize` directive the textarea height might be incorrect on component initialization.

This happens because the textarea `scrollHeight` property is not ready in the `ngOnInit` lifecycle hook yet.

Other libraries like `angular-elastic` have timeouts for that. But using the `ngAfterViewInit` lifecycle hook is more elegant and also ensures that the `scrollHeight` property is ready.

Also switches `offsetHeight` to `clientHeight` since we don't want to include the border in our line-height calculations.

Also by default `textarea` elements have a padding of `2px` and the `padding` needs to be explicitly set to `0px` when resolving the line-height.

Fixes #4070.

### DIFF
--- a/src/lib/input/autosize.ts
+++ b/src/lib/input/autosize.ts
@@ -71,7 +71,7 @@ export class MdTextareaAutosize implements AfterViewInit {
     textareaClone.style.position = 'absolute';
     textareaClone.style.visibility = 'hidden';
     textareaClone.style.border = 'none';
-    textareaClone.style.padding = '0px';
+    textareaClone.style.padding = '0';
     textareaClone.style.height = '';
     textareaClone.style.minHeight = '';
     textareaClone.style.maxHeight = '';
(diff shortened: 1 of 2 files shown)


src/lib/input/autosize.ts
```diff
@@ -71,13 +71,13 @@ export class MdTextareaAutosize implements OnInit {
     textareaClone.style.position = 'absolute';
     textareaClone.style.visibility = 'hidden';
     textareaClone.style.border = 'none';
-    textareaClone.style.padding = '';
+    textareaClone.style.padding = '0px';
```
jelbourn: Just `'0'`?

DevVersion: Done.

programmersharp:
I am showing textarea only ngAfterViewInit is invoking, but in any case **without timeout it does not work**. Basically, I am doing:
`
  ngAfterViewInit() {
    setTimeout(() => {
      this.pluginsInitialized = true;
    }, 2000);
  }`

Issue #4037, opened 2017-04-11
MdCheckbox disabled by FormControl API cannot be enabled if clicked while disabled


arlowhite:
#### Bug, feature request, or proposal:
Hookup a MdCheckbox to a FormControl.
`disable()`, try to click check, and `enable()` the form control.
Checkbox appears enabled but cannot be focused or checked/un-checked.

#### What is the expected behavior?
Checkbox should be able to be disabled/enabled using FormControl API and user interacting with disabled checkbox should not break it.

#### What is the current behavior?

1. checkbox works fine initially
2. `disable()`
3. Try clicking checkbox. First click causes ripple effect on check. More clicks do nothing
4. `enable()`
5. style changes to primary color, but cannot be clicked. Also, cannot be tab-focused

Sometimes I see:
> ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'true'. Current value: 'false'.

This seems to depend on whether you toggled the checkbox before disabling it.

#### What are the steps to reproduce?

https://github.com/arlowhite/material2/tree/checkbox-control-disable-bug
`npm run demo-app`
[Go to Input](http://localhost:4200/input/)
Scroll to **Checkbox BUG**

#### Which versions of Angular, Material, OS, browsers are affected?
angular2 master 9d719c544e0476b26c1fe0918b1c9c6bb9cd0ec0
Ran `npm update` in material2 project April 11th
Angular 4.0.1

Affects Chrome and Firefox; haven't tried others.

arlowhite:
Also, in my own app with a ReactiveForm FormGroup, the checkbox is broken after `disable()`/`enable()` even without interacting with it. I call `disable()` in response to the entire form's valueChanges.

This would be more difficult to create a demo for. But after you push a fix to master I can test if the bug went away in my own app.

mmacfadden:
We are also experiencing this same issue.

PR #4087, opened 2017-04-13
fix(checkbox): 	Fix disabled checkbox throw error when clicked

tinayuangao:
Related to #4037
The change detection in checkbox isn't working properly.  `disabled` is not updated in view.

### DIFF
--- a/src/lib/checkbox/checkbox.ts
+++ b/src/lib/checkbox/checkbox.ts
@@ -142,7 +142,7 @@ export class MdCheckbox implements ControlValueAccessor, AfterViewInit, OnDestro
   /** Whether the checkbox is disabled. */
   @Input()
   get disabled(): boolean { return this._disabled; }
-  set disabled(value) { this._disabled = coerceBooleanProperty(value);}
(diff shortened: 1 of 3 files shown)


src/lib/checkbox/checkbox.ts
```diff
@@ -142,7 +142,10 @@ export class MdCheckbox implements ControlValueAccessor, AfterViewInit, OnDestro
   /** Whether the checkbox is disabled. */
   @Input()
   get disabled(): boolean { return this._disabled; }
-  set disabled(value) { this._disabled = coerceBooleanProperty(value); }
+  set disabled(value) {
+    this._disabled = coerceBooleanProperty(value);
+    this._changeDetectorRef.markForCheck();
```
jelbourn: Should this go in `setDisabledState` instead of this setter instead?

tinayuangao: Yes. Moved to `setDisabledState`. Thanks

tinayuangao:
Added test. Thanks

PR #4107, opened 2017-04-15
fix(input): single-line hints overflowing the parent

### LINKED ISSUE #4051: MdHint causes MdTab to show scrollbar
Bug: MdTab shows the vertical scroll bar when it contains an md-hint inside input-container.

MdTab not showing the vertical scroll bar when using md-hint inside

It's showing vertical scrollbar when I use md-hint in input-container

Try to put and remove the md-hint of this plunker: https://plnkr.co/edit/3jAsCcMtnDr3Sh39LMap?p=preview

Use md-hint

@angular/animations: 4.0.1
@angular/cli: 1.0.0
@angular/common: 4.0.1
@angular/compiler: 4.0.1
@angular/compiler-cli: 4.0.1
@angular/core: 4.0.1
@angular/flex-layout: 2.0.0-beta.7
@angular/forms: 4.0.1
@angular/http: 4.0.1
@angular/platform-browser: 4.0.1
@angular/platform-browser-dynamic: 4.0.1
@angular/platform-server: 4.0.1
@angular/router: 4.0.1

As a workaround I'm using in styles.css

.mat-tab-body {
    overflow-y: hidden !important;
}

but not sure about side efects of doing it.

crisbeto:
Fixes the input container hints causing an overflow on the parent by a couple of pixels. Note that this can be fixed alternatively by setting `line-height: 1` on the wrapper, however it causes multi-line hints to be a little too compact. I went with 1.2 since it seems to be the closest to the browser defaults for `normal`.

Fixes #4051.

### DIFF
--- a/src/lib/input/input-container.scss
+++ b/src/lib/input/input-container.scss
@@ -4,7 +4,6 @@
 
 
 $mat-input-floating-placeholder-scale-factor: 0.75 !default;
-$mat-input-wrapper-spacing: 1em !default;
 $mat-input-hint-min-space: 10px !default;
 
 // Gradient for showing the dashed line when the input is disabled.
@@ -43,7 +42,7 @@ $mat-input-underline-disabled-background-image:
 // Global wrapper. We need to apply margin to the element for spacing, but
 // cannot apply it to the host element directly.
 .mat-input-wrapper {
-  margin: $mat-input-wrapper-spacing 0;
+  margin: 1em 0;
   // Account for the underline which has 4px of margin + 2px of border.
   padding-bottom: 6px;
 }
@@ -227,11 +226,14 @@ $mat-input-underline-disabled-background-image:
 // Wrapper for the hints and error messages. Provides positioning and text size.
 // Note that we're using `top` in order to allow for stacked children to flow downwards.
 .mat-input-subscript-wrapper {
+  $line-height: 1.2em;
+
   position: absolute;
   font-size: 75%;
   top: 100%;
   width: 100%;
-  margin-top: -$mat-input-wrapper-spacing;
+  margin-top: -$line-height;
+  line-height: $line-height;
   overflow: hidden; // prevents multi-line errors from overlapping the input
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/input/input-container.scss
```diff
@@ -227,11 +226,14 @@ $mat-input-underline-disabled-background-image:
...
 .mat-input-subscript-wrapper {
+  $line-height: 1.2em;
+
   position: absolute;
   font-size: 75%;
   top: 100%;
   width: 100%;
-  margin-top: -$mat-input-wrapper-spacing;
+  margin-top: -$line-height;
+  line-height: $line-height;
```
mmalerba: why not change the `line-height` on `.mat-input-container` from `normal` to `1.2em` instead of overriding here?

src/lib/input/input-container.scss
```diff
@@ -43,7 +42,7 @@ $mat-input-underline-disabled-background-image:
 // Global wrapper. We need to apply margin to the element for spacing, but
 // cannot apply it to the host element directly.
 .mat-input-wrapper {
-  margin: $mat-input-wrapper-spacing 0;
+  margin: 1em 0;
```
mmalerba: why did you remove the constant?

crisbeto: After the changes I did below, it was only being used in one place which makes it pointless.

src/lib/input/input-container.scss
crisbeto: I wanted to avoid breaking other parts of the input container since these changes are only relevant to the hints.

src/lib/input/input-container.scss
mmalerba: I think we often prefer constants even if they're only used once since they describe what the value is supposed to represent.

src/lib/input/input-container.scss
mmalerba: I think it should be ok, most browsers have normal = 1.2. Using 1.2 explicitly will make sure its consistent in all browsers

mmalerba:
LGTM

src/lib/input/input-container.scss
mmalerba: Since changing the one on the container causes other elements to overflow I'm fine with doing this

Issue #4108, opened 2017-04-15
Move package building logic into separate tool


DevVersion:
The build logic should not live in `tools/gulp` anymore. It could be exposed as a separate tool that will be shared between the different libraries.

DevVersion:
As a reference: #4719

PR #4111, opened 2017-04-15
build: build cdk package

DevVersion:
* Introduces another package called `@angular/cdk`. This package is
currently empty but all the logic for SASS, HTML, TypeScript and tests
is ready.

**Note**: Stuff like for the releasing is still missing. But this might be a start. Also I feel like the solution with the tests is not a perfect one. Also the logic will be simplified in a follow-up #4108


mjamin:
Any ETA on a first release of this package and what it will contain?

PR #4113, opened 2017-04-15
feat(progress-spinner): add support for custom stroke-width

### LINKED ISSUE #3934: New @Input in progress-spinner component for controlling the stroke-width
feature request/proposal

Add an @Input to the progress-spinner component so that we can do something like this:
`<md-spinner strokeWidth="2"></md-spinner>`

I'm not able to find a easy way to control the stroke width for progress spinner.

As a developer, I want to be able to control the thickness for my progress-spinners.

DevVersion:
* Adds support for custom `stroke-width` of the progress-spinner elements.

Fixes #3934

### DIFF
--- a/src/lib/progress-spinner/progress-spinner.html
+++ b/src/lib/progress-spinner/progress-spinner.html
@@ -5,5 +5,5 @@
 -->
 <svg viewBox="0 0 100 100"
      preserveAspectRatio="xMidYMid meet">
-  <path #path></path>
+  <path #path [style.strokeWidth]="strokeWidth"></path>
 </svg>
--- a/src/lib/progress-spinner/progress-spinner.scss
+++ b/src/lib/progress-spinner/progress-spinner.scss
@@ -33,8 +33,6 @@ $mat-progress-spinner-viewport-size: 100px !default;
   path {
     fill: transparent;
 
-    // Stroke width of 10px defines stroke as 10% of the viewBox.
-    stroke-width: $mat-progress-spinner-stroke-width;
     transition: stroke $swift-ease-in-duration $ease-in-out-curve-function;
   }
 
--- a/src/lib/progress-spinner/progress-spinner.spec.ts
+++ b/src/lib/progress-spinner/progress-spinner.spec.ts
@@ -2,6 +2,7 @@ import {TestBed, async} from '@angular/core/testing';
 import {Component} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {MdProgressSpinnerModule} from './index';
+import {PROGRESS_SPINNER_STROKE_WIDTH} from './progress-spinner';
 
 
 describe('MdProgressSpinner', () => {
@@ -14,6 +15,7 @@ describe('MdProgressSpinner', () => {
         IndeterminateProgressSpinner,
         ProgressSpinnerWithValueAndBoundMode,
         ProgressSpinnerWithColor,
+        ProgressSpinnerCustomStrokeWidth,
         IndeterminateProgressSpinnerWithNgIf,
         SpinnerWithNgIf,
         SpinnerWithColor
@@ -108,6 +110,27 @@ describe('MdProgressSpinner', () => {
     expect(progressElement.componentInstance.interdeterminateInterval).toBeFalsy();
   });
 
+  it('should set a default stroke width', () => {
+    let fixture = TestBed.createComponent(BasicProgressSpinner);
+    let pathElement = fixture.nativeElement.querySelector('path');
+
+    fixture.detectChanges();
+
+    expect(parseInt(pathElement.style.strokeWidth))
+      .toBe(PROGRESS_SPINNER_STROKE_WIDTH, 'Expected the default stroke-width to be applied.');
+  });
+
+  it('should allow a custom stroke width', () => {
+    let fixture = TestBed.createComponent(ProgressSpinnerCustomStrokeWidth);
+    let pathElement = fixture.nativeElement.querySelector('path');
+
+    fixture.componentInstance.strokeWidth = 40;
+    fixture.detectChanges();
+
+    expect(parseInt(pathElement.style.strokeWidth))
+      .toBe(40, 'Expected the custom stroke width to be applied to the path element.');
+  });
+
   it('should set the color class on the md-spinner', () => {
     let fixture = TestBed.createComponent(SpinnerWithColor);
     fixture.detectChanges();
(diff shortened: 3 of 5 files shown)


src/lib/progress-spinner/progress-spinner.scss
```diff
@@ -33,8 +33,8 @@ $mat-progress-spinner-viewport-size: 100px !default;
   path {
     fill: transparent;
 
-    // Stroke width of 10px defines stroke as 10% of the viewBox.
-    stroke-width: $mat-progress-spinner-stroke-width;
+    // Stroke width is by default 10px, which is relatively to the viewBox size 10%.
```
josephperrott: It currently is centered properly in the viewBox in the `getSvgArc` method, we define `pathRadius` as 40 and `radius` as 50.  If we allow the `strokeWidth` to be set we will need to make a corresponding adjustment to the pathRadius.  

Thought we need to keep in mind  `getSvgArc` is called ~60 times/sec, so ideally we don't want to recalculate the `pathRadius` value every time if we can avoid it, though it is a pretty trivial calculation compared to determining easing values.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -103,6 +107,11 @@ export class MdProgressSpinner implements OnDestroy {
     this._cleanupIndeterminateAnimation();
   }
 
+  /** Stroke width of the progress spinner. By default uses 10px as stroke width. */
+  @Input()
```
josephperrott: why not just use 

`@Input() strokeWidth: number = DEFAULT_STROKE_WIDTH;`

Then it will have the default value, otherwise will use the input bound value.

src/lib/progress-spinner/progress-spinner.html
```diff
@@ -5,5 +5,5 @@
 -->
 <svg viewBox="0 0 100 100"
      preserveAspectRatio="xMidYMid meet">
-  <path></path>
+  <path [style.strokeWidth]="strokeWidth"></path>
```
josephperrott: If the value is being set in the style attribute here, should it be removed from the scss file?

DevVersion: I think I removed it?

src/lib/progress-spinner/progress-spinner.ts
DevVersion: I wanted to make a getter/setter because I wanted to avoid that the developer sets the value to `0` or falsy values.

src/lib/progress-spinner/progress-spinner.html
josephperrott: Yep, I just missed it sorry.

src/lib/progress-spinner/progress-spinner.spec.ts
```diff
@@ -108,6 +108,29 @@ describe('MdProgressSpinner', () => {
     expect(progressElement.componentInstance.interdeterminateInterval).toBeFalsy();
   });
 
+  it('should allow a custom stoke width', () => {
```
crisbeto: stoke->stroke

src/lib/progress-spinner/progress-spinner.spec.ts
```diff
@@ -108,6 +108,29 @@ describe('MdProgressSpinner', () => {
...
+    let fixture = TestBed.createComponent(BasicProgressSpinner);
+    let pathElement = fixture.nativeElement.querySelector('path') as HTMLElement;
+
+    fixture.detectChanges();
+
+    expect(parseInt(pathElement.style.strokeWidth))
```
crisbeto: There's no need to parse the string to an int here. You can just expect it to be `'10'`.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -25,6 +25,9 @@ const endIndeterminate = 80;
 /* Maximum angle for the arc. The angle can't be exactly 360, because the arc becomes hidden. */
 const MAX_ANGLE = 359.99 / 100;
 
+// Stroke width is by default 10px, which is relatively to the viewBox size 10%.
```
crisbeto: Not sure whether this comment is very helpful. We'd also need to keep it updated if we change the default value. Something like "Default stroke width as a percentage of the viewBox" might be better.

src/lib/progress-spinner/progress-spinner.ts
crisbeto: This will still break if the consumer sets it to something like "abc". IMO the getter/setter here is overkill.

src/lib/progress-spinner/progress-spinner.spec.ts
```diff
@@ -108,6 +108,29 @@ describe('MdProgressSpinner', () => {
...
+      .toBe(20, 'Expected custom stroke width to be applied.');
+
+    fixture.componentInstance.strokeWidth = 0;
+    fixture.detectChanges();
+
+    expect(parseInt(pathElement.style.strokeWidth))
+      .toBe(10, 'Expected stroke width to fallback to 10px.');
```
crisbeto: I'm not sure whether we should be testing the exact default value here since it'll start failing if we change it. I suggest either importing the `DEFAULT_STROKE_WIDTH` and checking against it, or just checking that it has a valid stroke width in the first place.

src/lib/progress-spinner/progress-spinner.ts
DevVersion: Agreed. We can just set the default stroke-width as initial value.

src/lib/progress-spinner/progress-spinner.spec.ts
DevVersion: There is a need for that. Safari 9 adds the `px` suffix to the `strokeWidth` value. We could have also done it through `toContain` but `parseInt` and `toBe` is more safe here.

src/lib/progress-spinner/progress-spinner.spec.ts
DevVersion: I think importing might be the best solution here.

src/lib/progress-spinner/progress-spinner.scss
DevVersion: That's a good point. In theory we could cache the `pathRadius` when the stroke width is updated through a `setter`. But I feel like it's not worth passing a cached value to the `getSvgArc` function (due to the trivial subtraction).

Let me know what you think. Happy to implement the caching if desired.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -73,6 +75,9 @@ export class MdProgressSpinner implements OnDestroy {
   private _value: number;
   private _color: string = 'primary';
 
+  /** Stroke width of the progress spinner. By default uses 10px as stroke width. */
```
crisbeto: I don't think that this should have the `By default uses 10px as stroke width.` since the default may change and I doubt that we'd remember to update the comment.

wiwski:
Hi there,
Any news about this merge ?

PR #4115, opened 2017-04-16
feat(list): input binding for dense mode

### LINKED ISSUE #3882: change md-list `dense` attribute programmatically without "hack"
Proposal.

Change `dense` attribute programmatically.
```html
<md-list [dense]="dense"></md-list>
<!-- dense can be true or false -->
```

`dense` is an html attribute stylized with CSS.
```css
.mat-list[dense], .mat-nav-list[dense] { }
```

Is useful for main sidenav menu: dense for desktop and default for mobile.
Currently to change this option we have to do something like:
```html
<md-list [attr.dense]="dense"></md-list>
<!-- dense can be true or null -->
```
```ts
dense: true | null; // null to remove dense attribute from the DOM.
```

DevVersion:
* Replaces the attribute selector for dense mode with an Angular input binding.

**Note**: As mentioned in the issue, it's currently possible to change `dense` mode dynamically using `[attr.dense]` .. but the API is not very elegant and you have to use `null` to disable dense mode.

Fixes #3882

### DIFF
--- a/src/demo-app/list/list-demo.html
+++ b/src/demo-app/list/list-demo.html
@@ -1,7 +1,12 @@
 
 <h1>md-list demo</h1>
 
-<button md-raised-button (click)="thirdLine=!thirdLine" class="demo-button">Show third line</button>
+<button md-raised-button (click)="thirdLine=!thirdLine" class="demo-button">
+  Toggle third line
+</button>
+<button md-raised-button (click)="showDense=!showDense" class="demo-button">
+  Toggle dense mode
+</button>
 
 <div class="demo">
   <div>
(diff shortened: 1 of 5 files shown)


crisbeto:
Wouldn't it be better to deprecate the `dense` attribute altogether and have a class for the dense mode instead? It will make it faster to toggle and will allow us to set up inheritance.

DevVersion:
The `dense` attribute still works due to the boolean coercion.

crisbeto:
I mean a CSS class for toggling the dense mode (e.g. `.mat-dense`), instead of an attribute. This will allow us to handle dense mode in a similar way to theming, when we start supporting it on more components.

DevVersion:
Hmm not sure. The `dense` binding seems more convenient and a nicer API. Otherwise it ends up as in Material 1 where we have X classes on the list. Also the `dense` attribute sets a class already.

crisbeto:
That's what I mean: the `dense` attribute is just a proxy for the class and it adds some unnecessary bloat. If we use a class, we get the toggling for free via `ngClass` and we can use it if we set up inheritance later on. Still, this is mostly a point for discussion.

DevVersion:
@crisbeto That's probably a more generic discussion about API design.. Otherwise developers could also just do `.md-accent`, `.md-primary` etc.

jelbourn:
I would rather avoid changing this until we have a more comprehensive approach for handling density. This would include a uniform API for all components, multiple levels of density, a way to set the default for all components, and possibly a way to set the density for just a portion of the app. 

Is there a compelling reason to change just the list now?

DevVersion:
Not really. Just a request through an issue. This change isn't really major anyways, it just changes the `attribute selector` to a `@Input()` binding (more convenient when dynamic toggling).

Feel free to close if necessary.

jelbourn:
Ack- closing until we tackle the larger density issue

PR #4120, opened 2017-04-17
fix(dialog): provide default value for MD_DIALOG_DATA token

### LINKED ISSUE #4086: ERROR Error: No provider for Token MdDialogData!
As of [2.0.0-beta.3](https://github.com/angular/material2/releases/tag/2.0.0-beta.3), you need to inject the MD_DIALOG_DATA token in order to access the data property of the MdDialogConfig object.  If you inject the MD_DIALOG_DATA token into a component that will be loaded into the dialog, and you do not pass data in, you will get

```
Error: No provider for Token MdDialogData!
```

Do not throw an error and to return an empty object if none is passed in.

```
Error: No provider for Token MdDialogData!
```

[Plunker](http://plnkr.co/edit/cRO1UhY9x46SYV0Kpmfd)

Angular Material 2.0.0-beta.3
Angular 4.00

crisbeto:
Adds a default value for the `MD_DIALOG_DATA` injection token, instead of throwing an error when it's not defined.

Fixes #4086.

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -40,7 +40,7 @@ export class MdDialogConfig {
   position?: DialogPosition;
 
   /** Data being injected into the child component. */
-  data?: any = {};
+  data?: any = null;
 
   // TODO(jelbourn): add configuration for lifecycle hooks, ARIA labelling.
 }
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -40,7 +40,7 @@ export class MdDialogConfig {
   position?: DialogPosition;
 
   /** Data being injected into the child component. */
-  data?: any;
+  data?: any = {};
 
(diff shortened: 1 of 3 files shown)


src/lib/dialog/dialog-config.ts
```diff
@@ -40,7 +40,7 @@ export class MdDialogConfig {
   position?: DialogPosition;
 
   /** Data being injected into the child component. */
-  data?: any;
+  data?: any = {};
```
jelbourn: How about `null` instead of `{}`?

crisbeto: Done.

xtianus79:
oddly I am gettig this error in a popup window... i.e. window.open... I know the values can't be passed there so I will try @optional()

xtianus79:
that was scary lol... @optional did the trick... I feel like this should just be obvious without dragging something in.

PR #4140, opened 2017-04-18
fix(slide-toggle): invalid model change event

### LINKED ISSUE #4124: slide-toggle fires undesired event when setting a default value
**Bug**

Currently if you set a default value for a slide-toggle it'll emit a ngModelChange event when the view inits, IMO the correct behavior would be to not emit this event since there was no change at all.

http://plnkr.co/edit/imQpobzGdi3Xl6IbXYo7?p=preview

Or use the component below:
```javascript
import { Component } from '@angular/core';

@Component({
  selector: 'demo-app',
  template: `
    <md-slide-toggle [(ngModel)]="test" (ngModelChange)="change()"></md-slide-toggle>
    <hr>
    <div [hidden] = "hide_msg">
        Change event was received:
    </div>
  `
})
export class AppComponent {
  hide_msg: boolean = true;
  test = true;
  
  change() {
    this.hide_msg = false;
  }
}
```

All browsers and Angular versions...

Even changing test to ``` 1 ``` the event will also be fired...

DevVersion:
While initially looking into #4124, there were a few more issues inside
of the slide-toggle.

* ngModelChange event is dispatched at initialization.
* Checked state isn't synchronized when state changes through drag. New state after dragging got overwritten by click event on label.
* Removes unnecessary logic inside of `change` listener. Change event doesn't fire if underlying checkbox is disabled.
* Slide-toggle turns touched only on `blur`. Similar as for native checkbox elements.

**Note**: Basically made the slide-toggle even more aligned with the native checkbox elements.

Fixes #4124.

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -295,18 +295,23 @@ describe('MdSlideToggle', () => {
       expect(slideToggleControl.pristine).toBe(true);
       expect(slideToggleControl.touched).toBe(false);
 
-      // After changing the value programmatically, the control should
-      // become dirty (not pristine), but remain untouched.
+      // After changing the value from the view, the control should
+      // become dirty (not pristine), but remain untouched if focus is still there.
       slideToggle.checked = true;
+
+      // Dispatch a change event on the input element to fake a user interaction that triggered
+      // the state change.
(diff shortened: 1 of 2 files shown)


kara:
@DevVersion  Can you rebase?

DevVersion:
@kara Done :)

PR #3928, opened 2017-04-06
feat(select): allow setting the theme color

crisbeto:
* Allows the user to set the theme color of a `md-select`.
* Adds theming to the `md-option` component.
* Tweaks the `md-pseudo-checkbox` theme to allow it to inherit the color.

Fixes #3923.

PR #4145, opened 2017-04-18
fix(theming): fix broken sass expressions nested in theme classes

### LINKED ISSUE #4077: Some sass selector expressions don't work when nested in a theme class
I have searched this problem and find https://github.com/angular/material2/issues/2662, but when I try to use it I found, the select component still has a problem. 

After setting the theme for the overlay container, the overlying theme of select is changed. 

However the problem is the Placeholder and the underline, the color is still the previous theme

 

I try to find the reason of the issue and find that there are some CSS define for the placeholder and underline as follows:

material2-master/src/lib/select/_select-theme.scss
```
.mat-select-trigger {
    color: mat-color($foreground, hint-text);

    .mat-select:focus:not(.mat-select-disabled) & {
      color: mat-color($primary);
    }

    .mat-select:not(:focus).ng-invalid.ng-touched:not(.mat-select-disabled) & {
      color: mat-color($warn);
    }
  }

  .mat-select-underline {
    background-color: mat-color($foreground, divider);

    .mat-select:focus:not(.mat-select-disabled) & {
      background-color: mat-color($primary);
    }

    .mat-select:not(:focus).ng-invalid.ng-touched:not(.mat-select-disabled) & {
      background-color: mat-color($warn);
    }
  }
```

So when the Select is focused, these CSS Pseudo-Classes is activated, and the color is the previous theme. I think all DOM add a class of new theme `.xxxxx ` (such as `.red-indigo`), but the pseudo-class is activated, this new theme class is not added. 

I try to solve it, but I cannot find the related code to add class to the DOM

I make a demo and deploy on Firebase

https://github.com/xmeng1/ng2-material-admin

https://ng2-material-admin.firebaseapp.com/

crisbeto:
* Fixes certain theme selectors being broken due to uses of the `&` operator at the end of the selector.
* Adds a custom Stylelint rule to catch future improper uses of the ampersand inside themes.

Relates to #3928.
Fixes #4077.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -22,18 +22,19 @@
 
   .mat-input-placeholder {
     color: $input-placeholder-color;
+  }
 
-    // :focus is applied to the input, but we apply mat-focused to the other elements
-    // that need to listen to it.
-    .mat-focused & {
-      color: $input-floating-placeholder-color;
+  // :focus is applied to the input, but we apply mat-focused to the other elements
+  // that need to listen to it.
+  .mat-focused .mat-input-placeholder {
+    color: $input-floating-placeholder-color;
 
-      &.mat-accent {
-        color: $input-underline-color-accent;
-      }
-      &.mat-warn {
-        color: $input-underline-color-warn;
-      }
+    &.mat-accent {
+      color: $input-underline-color-accent;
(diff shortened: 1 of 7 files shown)


kara:

crisbeto:
Rebased.

PR #4147, opened 2017-04-18
chore: add `@__PURE__` annotation to ES5 output for uglify

jelbourn:
Context: https://gist.github.com/IgorMinar/09eadcb221a8253162078210639a344d


tools/gulp/util/annotate-pure.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @returns The content of the file with `@__PURE__` annotations added.
+ */
+export function addPureAnnotations(fileContent: string) {
+  return fileContent
+      // Prefix downleveled classes w/ the @__PURE__ annotation.
+      .replace(iifeExpression, '$1/*@__PURE__*/$3')
+      // Prefix downleveled classes that extend another class w/ the @__PURE__ annotation
+      .replace(
+          /^(var (\S+) = )(\(function \(_super\) \{\n    __extends\(\2, _super\);\n)/mg,
+          '$1/*@__PURE__*/$3'
```
DevVersion: Should we move that RegExp to a constant as well?

jelbourn: I only moved the first one because it's over 100 characters long and needed to be wrapped in a string.

jelbourn:
Also yes, I did test it with material.angular.io and seems to work fine.

tools/gulp/util/annotate-pure.ts
DevVersion: Fine by me. Just think that it would be more clean if you have it like a pipeline.

But since its mostly temporary.. doesn't matter.

PR #4148, opened 2017-04-19
feat(datepicker): add DateAdapter and NativeDateAdapter

mmalerba:


src/lib/core/datetime/date-adapter.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /**
+   * Checks if two dates are equal.
+   * @param first The first date to check.
+   * @param second The second date to check.
+   * @returns {boolean} Whether the two dates are equal.
+   *     Null dates are considered equal to other null dates.
+   */
+  equals(first: D | null, second: D | null): boolean {
```
jelbourn: Similar comment as above for `equals` and `compare`: you'll want different things for the time variants. Maybe just call them `compareDate` and `sameDate`?

src/lib/core/datetime/native-date-adapter.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+  });
+
+  it('should create Date with low year number and over/under-flow', () => {
+    expect(adapter.create(50, 12 * 51, 1).getFullYear()).toBe(101);
+    expect(adapter.create(50, 12, 1).getFullYear()).toBe(51);
+    expect(adapter.create(50, -12, 1).getFullYear()).toBe(49);
+    expect(adapter.create(50, -12 * 51, 1).getFullYear()).toBe(-1);
+  });
+
+  it('should get today\'s date', () => {
```
jelbourn: optional: you can use double quotes (or template string backticks) when it's to avoid escaping single quotes

src/lib/core/datetime/date-adapter.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Clamp the given date between min and max dates.
+   * @param date The date to clamp.
+   * @param min The minimum value to allow. If null or omitted no min is enforced.
+   * @param max The maximum value to allow. If null or omitted no max is enforced.
+   * @returns `min` if `date` is less than `min`, `max` if date is greater than `max`,
+   *     otherwise `date`.
+   */
+  clamp(date: D, min?: D | null, max?: D | null): D {
```
jelbourn: Same thing about date vs. time/datetime

src/lib/core/datetime/date-adapter.ts
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * Gets the day of the week component of the given date.
+   * @param date The date to extract the day of the week from.
+   * @returns The month component (0-indexed, 0 = Sunday).
+   */
+  abstract getDay(date: D): number;
+
+  /**
+   * Gets a list of names for the months.
+   * @param style The naming style (e.g. long = 'January', short = 'Jan', narrow = 'J').
```
mmalerba: I called it `style` because I don't want people confusing it with the `format: any` that they would pass to the `parse` or `format` functions

src/lib/core/datetime/date-adapter.ts
```diff
@@ -0,0 +1,170 @@
...
+   */
+  abstract format(date: D, fmt?: any, locale?: any): string;
+
+  /**
+   * Adds the given number of years, months, and days to the given date.
+   * @param date The date to add to.
+   * @param amount The number of years, months, and days to add (may be negative).
+   * @returns A new date equal to the original with the given amount of time added.
+   */
+  abstract add(date: D, amount: {years?: number, months?: number, days?: number}): D;
```
mmalerba: Hmm interesting... I assumed moment added years first then months then days, like how I do for `Date`, but it seems it does the opposite order. Do you think I should add largest unit first, smallest unit first, or only allow adding 1 unit at a time and force the user to determine the order?

jelbourn: Only adding one unit at a time is probably the only way to ensure that it does exactly what the user expects (e.g., `addYears`, `addMonths`, etc.). If we assume any kind of order of operations, I'd want to run a quick question by Google's i18n experts as to how they would say it should behave.

mmalerba: I'm also noticing that moment and closure seem to add what I'm calling 'calendarDays', 'calendarMonths', and 'calendarYears' here: https://github.com/angular/material2/blob/datepicker/src/lib/datepicker/calendar.ts#L271 I think I'll make add do that to be more in line with these other libraries

mmalerba:
comments addressed, PTAL

src/lib/core/datetime/date-adapter.ts
```diff
@@ -0,0 +1,196 @@
...
+   */
+  abstract parse(value: any, fmt?: any): D | null;
+
+  /**
+   * Formats a date as a string.
+   * @param date The value to parse.
+   * @param fmt The format to use for the result string.
+   * @returns The parsed date, or null if date could not be parsed.
+   */
+  abstract format(date: D, fmt?: any): string;
```
jelbourn: With this one format method, are we going to be able to get the different formats we need for date / time / datetime?

src/lib/core/datetime/date-adapter.ts
```diff
@@ -0,0 +1,196 @@
...
+    this.locale = locale;
+  }
+
+  /**
+   * Clones the given date.
+   * @param date The date to clone
+   * @returns A new date equal to the given date.
+   */
+  clone(date: D): D {
+    return this.createDate(this.getYear(date), this.getMonth(date), this.getDate(date));
```
jelbourn: Would it be better to leave this without an implementation here so that the method can be used for all of date/time/datetime?

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -0,0 +1,141 @@
...
+      let dtf = new Intl.DateTimeFormat(this.locale, {month: style});
+      return range(12, i => dtf.format(new Date(2017, i, 1)));
+    }
+    return {
+      'long': [
+        'January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September',
+        'October', 'November', 'December'
+      ],
+      'short': ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'],
+      'narrow': ['J', 'F', 'M', 'A', 'M', 'J', 'J', 'A', 'S', 'O', 'N', 'D']
```
jelbourn: By hard-coding the values into the return, you'll end up creating a new instance of these arrays each time the method is called. Move this object into a constant?

src/lib/core/datetime/native-date-adapter.spec.ts
```diff
@@ -0,0 +1,242 @@
+import {NativeDateAdapter} from './native-date-adapter';
+
+
```
jelbourn: In the material1 tests, I used constants for month indices to make the tests easier to read:
```js
  // When constructing a Date, the month is zero-based. This can be confusing, since people are
  // used to seeing them one-based. So we create these aliases to make reading the tests easier.
  var JAN = 0, FEB = 1, MAR = 2, APR = 3, MAY = 4, JUN = 5, JUL = 6, AUG = 7, SEP = 8, OCT = 9,
      NOV = 10, DEC = 11;
```

Then your tests look like 
```ts
expect(adapter.getYear(new Date(2017, JAN, 1))).toBe(2017);
```
Make a similar change here?

src/lib/core/datetime/date-adapter.ts
mmalerba: I think this should be ok `Intl.DateTimeFormat`, `moment.js`, and `goog.date.Date` all support this

mmalerba:
comments addressed

PR #4158, opened 2017-04-19
fix(api docs): properly format api docs using markdown

superMDguy:
Some sections of the api docs look like they're supposed to be formatted using markdown.  See, for example, [here](https://material.angular.io/components/component/icon), in the **directives** section under the **API REFERENCE** tab.

To do this, I had to disable dgeni's `templateEngine.config.autoescape`.  A comment said that "we want this turned on", but didn't specify why.  Since I didn't see any issues in the rendered html, I decided to go with it.


jelbourn:

superMDguy:
Looks like it's not an issue anymore.

PR #4159, opened 2017-04-19
feat(list): option to disable ripples for all items

### LINKED ISSUE #4149: md-nav-list disableRipple
Bug

Putting just disableRipple should disable ripples

disableRipple="true" works

(Based on material list example)
https://plnkr.co/edit/h2Jnm62Pzbn6P0UaXtgk

Consistency with other controls such as buttons

Angular 4.0.2
Material beta.3

DevVersion:
* Allows developers to apply the `[disableRipple]` binding onto the
`md-nav-list` element.
* Removes the Provider/token stuff, since we can do more elegant by just
injecting the NavList styler.
* Also fixes that the boolean is not coerced for the `disableRipple` binding.

Fixes #4149.

### DIFF
--- a/src/lib/list/list.ts
+++ b/src/lib/list/list.ts
@@ -6,28 +6,17 @@ import {
   QueryList,
   Directive,
   ElementRef,
-  Inject,
   Input,
-  OpaqueToken,
   Optional,
   Renderer,
   AfterContentInit,
 } from '@angular/core';
-import {MdLine, MdLineSetter} from '../core';
+import {MdLine, MdLineSetter, coerceBooleanProperty} from '../core';
 
 @Directive({
   selector: 'md-divider, mat-divider'
 })
 export class MdListDivider {}
-
-/**
- * Token used to inject the list type into child MdListItem components so they can know whether
- * they're in a nav list (and thus should use an MdRipple).
- */
-export const LIST_TYPE_TOKEN = new OpaqueToken('list_type');
-
-const NORMAL_LIST_TYPE = 'normal_list_type';
-const NAV_LIST_TYPE = 'nav_list_type';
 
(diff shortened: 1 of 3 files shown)


src/lib/list/list.ts
```diff
@@ -135,14 +124,20 @@ export class MdListSubheaderCssMatStyler {}
...
-  /**
-   * Whether the ripple effect on click should be disabled. This applies only to list items that
-   * are children of an md-nav-list; md-list items never have ripples.
-   */
-  @Input() disableRipple: boolean = false;
+  private _lineSetter: MdLineSetter;
+  private _disableRipple: boolean = false;
+  private _isNavList: boolean = false;
+
+  // Variable needs to be public for AOT compilation.
```
jelbourn: We can omit this comment; at this point it's library-wide and we have a CI check for it.

src/lib/list/list.ts
```diff
@@ -135,14 +124,20 @@ export class MdListSubheaderCssMatStyler {}
...
   _hasFocus: boolean = false;
 
-  private _lineSetter: MdLineSetter;
+  /**
+   * Whether the ripple effect on click should be disabled. This applies only to list items that are
+   * part of a nav list. The value of `disableRipple` on the `md-nav-list` overwrites this flag.
```
jelbourn: `overwrite` -> `override` (since the parent doesn't write a different value, it just takes precedence)

src/lib/list/list.ts
DevVersion: Yep makes sense.

src/lib/list/list.ts
DevVersion: Done.

PR #4166, opened 2017-04-20
Remove extra single quote in snack-bar unit test

PranavSathy:
The SnackBar unit tests had a an extra single quite in two of the unit tests, which are now removed.


PranavSathy:
I signed it!

PR #4174, opened 2017-04-20
build: dedupe theming scss bundle

### LINKED ISSUE #3931: Theming mixins/variables massively duplicated in release output
Bug? Proposal?

In the output [_theming.scss](https://github.com/angular/material2-builds/blob/master/_theming.scss), there is massive duplication of the palettes and mixins.

The 756 KB file can be brought down to ~82 KB by removing the duplication.

I don't really know/understand how big of deal this is. I guess it slows down npm download times and takes up 600 KB+ more disk space than required.

Related to #3776

DevVersion:
* Updates to a more recent version of SCSS-bundle that has support for deduping. The dedupe process is very naive and only disallows importing a file multiple times.
* Switches to the programmatic API of `scss-bundle`
* Removes unused functions and imports / leftovers.

Fixes #3931

### DIFF
--- a/tools/gulp/tasks/release.ts
+++ b/tools/gulp/tasks/release.ts
@@ -1,23 +1,18 @@
 import {spawn} from 'child_process';
-import {existsSync, readFileSync, statSync, writeFileSync} from 'fs-extra';
-import {basename, join} from 'path';
+import {existsSync, statSync, writeFileSync} from 'fs-extra';
+import {join} from 'path';
 import {dest, src, task} from 'gulp';
-import {inlineMetadataResources} from '../util/inline-resources';
-import {execNodeTask, execTask, sequenceTask} from '../util/task_helpers';
+import {execTask, sequenceTask} from '../util/task_helpers';
 import {composeRelease} from '../util/package-build';
+import {Bundler} from 'scss-bundle';
 import {
   COMPONENTS_DIR,
-  DIST_BUNDLES,
   DIST_MATERIAL,
   DIST_RELEASES,
-  DIST_ROOT,
-  LICENSE_BANNER,
-  PROJECT_ROOT,
 } from '../constants';
 import * as minimist from 'minimist';
 
 // There are no type definitions available for these imports.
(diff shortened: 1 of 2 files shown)


tools/gulp/tasks/release.ts
```diff
@@ -32,6 +27,10 @@ const themingEntryPointPath = join(COMPONENTS_DIR, 'core', 'theming', '_all-them
 // Output path for the scss theming bundle.
 const themingBundlePath = join(releasePath, '_theming.scss');
 
+// Glob that matches all files that might be imported multiple times.
+// Necessary for deduping inside of scss-bundle.
+const themingBundleDedupeGlob = join(COMPONENTS_DIR, '**/*.scss');
```
jelbourn: Should be named for what is it rather than how it's used, e.g., `allScssGlob`

tools/gulp/tasks/release.ts
```diff
@@ -51,27 +50,18 @@ task(':package:theming', [':bundle:theming-scss'], () => {
...
 /** Bundles all scss requires for theming into a single scss file in the root of the package. */
-task(':bundle:theming-scss', execNodeTask(
-  'scss-bundle',
-  'scss-bundle', [
-    '-e', themingEntryPointPath,
-    '-d', themingBundlePath
-  ], {silentStdout: true}
-));
+task(':bundle:theming-scss', () => {
+  new Bundler().Bundle(themingEntryPointPath, [themingBundleDedupeGlob]).then(result => {
```
jelbourn: Could you comment on what the arguments are here since it's not really clear from the call site?

tools/gulp/tasks/release.ts
DevVersion: I named it `themingBundleDedupeGlob` because I wanted to make it clear that this will be the array that is responsible for deduping.

tools/gulp/tasks/release.ts
DevVersion: Done.

PR #4178, opened 2017-04-20
fix: allow users to disable the sanity checks

### LINKED ISSUE #4125: Could not find Angular Material core theme. Most Material components may not work as expected. For more info refer to the theming guide: https://material.angular.io/guide/theming
Hi! Ciao! How Are You???

Could not find Angular Material core theme, although all core ``` @include mat-core();``` is included. I got 6 themes, all working except the warning.

No warning.

Warning.

```bash
git clone https://github.com/patrikx3/corifeus-app-web-pages.git
cd corifeus-app-web-pages
yarn install
grunt run
```
See.

Themes.

Current.

Nope.

crisbeto:
Allows users to disable the Material sanity checks, by providing a value for the `MATERIAL_SANITY_CHECKS` token:

```ts
@NgModule({
  providers: [
    {provide: MATERIAL_SANITY_CHECKS, useValue: false}
  ]

  // other config
});

```

Fixes #4125.

### DIFF
--- a/src/lib/core/compatibility/compatibility.ts
+++ b/src/lib/core/compatibility/compatibility.ts
@@ -7,14 +7,15 @@ import {
   Optional,
   isDevMode,
   ElementRef,
+  InjectionToken,
 } from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
 import {MdError} from '../errors/error';
 
-/** Whether we've done the global sanity checks (e.g. a theme is loaded, there is a doctype). */
-let hasDoneGlobalChecks = false;
-
 export const MATERIAL_COMPATIBILITY_MODE = new OpaqueToken('md-compatibility-mode');
+
+/** Injection token that configures whether the Material sanity checks are enabled. */
+export const MATERIAL_SANITY_CHECKS = new InjectionToken<boolean>('md-sanity-checks');
 
 /**
  * Exception thrown if the consumer has used an invalid Material prefix on a component.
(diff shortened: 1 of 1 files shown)


src/lib/core/compatibility/compatibility.ts
```diff
@@ -193,11 +194,16 @@ export class CompatibilityModule {
     };
   }
 
-  constructor(@Optional() @Inject(DOCUMENT) private _document: any) {
+  constructor(@Optional() @Inject(DOCUMENT) private _document: any, ngZone: NgZone) {
     if (!hasDoneGlobalChecks && isDevMode()) {
-      this._checkDoctype();
-      this._checkTheme();
-      hasDoneGlobalChecks = true;
+      ngZone.runOutsideAngular(() => {
```
jelbourn: Can we gate on being on the browser platform here rather than in the individual checks as well?
(no reason the timeout has to run on platform-server)

crisbeto:

jelbourn:
@kara or @mmalerba what's the source of the presubmit failure?

jelbourn:
Follow-up: this PR makes ~60 different project's unit tests start timing out. Strange, since the call is done with `runOutsideAngular`; not sure why it would affect the tests.

crisbeto:
It might be because a lot of timeouts are still being spawned in the background. We could go down the safer route and alow users to disable the checks, instead of delaying them?

jelbourn:
@crisbeto yeah, I think so now

crisbeto:
Re-did this as discussed @jelbourn.

PR #4180, opened 2017-04-20
fix(textarea): fix change detection error on autosize

kara:
Fixes bug where change detection error occurs on autosize (test failure in Google).

### DIFF
--- a/src/lib/input/autosize.ts
+++ b/src/lib/input/autosize.ts
@@ -10,23 +10,36 @@ import {Directive, ElementRef, Input, AfterViewInit} from '@angular/core';
   exportAs: 'mdTextareaAutosize',
   host: {
     '(input)': 'resizeToFitContent()',
-    '[style.min-height]': '_minHeight',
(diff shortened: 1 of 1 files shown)


src/lib/input/autosize.ts
```diff
@@ -36,21 +49,37 @@ export class MdTextareaAutosize implements AfterViewInit {
 
   constructor(private _elementRef: ElementRef) { }
 
-  /** The minimum height of the textarea as determined by minRows. */
-  get _minHeight() {
-    return this.minRows ? `${this.minRows * this._cachedLineHeight}px` : null;
+  /** Sets the minimum height of the textarea as determined by minRows. */
+  _setMinHeight(): void {
+    const minHeight = this.minRows && this._cachedLineHeight
```
jelbourn: Prefer putting the `?` at the end of the previous line rather than the start of the new line

kara: Done

Issue #440, opened 2016-05-13
Don't expose EventEmitters in Component public API


jelbourn:
Component consumers should not be able to call `emit` on any internal `EventEmitter` instance. As such,  components should do something like

```ts
@Output('change') private _change = new EventEmitter<void>();
change: Observable;

constructor() {
  this.change = this._change.asObservable();
}
```

jelbourn:
`@Output` can't be private w/ AoT, so we can't really do anything about this.

PR #4187, opened 2017-04-20
feat(autocomplete): emit event when an option is selected

### LINKED ISSUE #3645: feat(autocomplete): fire "onSelected" event when the user select an option
featur

Create a new output (onSelected?) that fires an event when the user select an option from the autocomplete

There is no way to know when the user has selected an option from the autocomplete

I have some cases that I need to update some fields based on the user selected option and would be nice if I can get a callback from the component itself

### LINKED ISSUE #4094: Autocomplete onSelectionChange fires two times when changing previously selected value
#### Bug:
If the autocomplete input field already has a value and this is changed by deleting the previous value and selecting a new option from the list, then the autocomplete's `onSelectionChange` fires twice.

When editing/ changing the previous selected value `onSelectionChange` should fire only once.

When editing/ changing the previous selected value `onSelectionChange` fires twice

Step 1: Select a value from autocomplete dropdown
Step 2: Clear this value (use backspace/ select all and erase)
Step 3: Select a different value from the autocomplete

https://plnkr.co/edit/cOYg48HGf86ZtG0fO59r?p=preview

When you want to edit a pre-filled value or you selected a value by mistake and want to make a change and if there are other fields dependent on this value, then the dependent values don't get updated.

"@angular/material": "^2.0.0-beta.3"
OS: Mac
browser: Chrome/ Safari

Nope.

crisbeto:
Emits the `select` event when an option in the autocomplete is selected.

**Note:** I went with passing the selected option from the trigger to the panel, instead of listening to the `onSelectionChange` inside the panel, because it involves keeping track of less subscriptions and not having to re-construct them when the list of options changes.

Fixes #4094.
Fixes #3645.

### DIFF
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -21,7 +21,7 @@ import {
   MdAutocomplete,
   MdAutocompleteModule,
   MdAutocompleteTrigger,
-  MdAutocompleteSelect,
+  MdAutocompleteSelectedEvent,
 } from './index';
 import {MdInputModule} from '../input/index';
 import {Subscription} from 'rxjs/Subscription';
@@ -1551,7 +1551,7 @@ describe('MdAutocomplete', () => {
       });
     }));
 
-  it('should call emit an event when an option is selected', fakeAsync(() => {
+  it('should emit an event when an option is selected', fakeAsync(() => {
     let fixture = TestBed.createComponent(AutocompleteWithSelectEvent);
 
     fixture.detectChanges();
@@ -1560,7 +1560,7 @@ describe('MdAutocomplete', () => {
     fixture.detectChanges();
 
     let options = overlayContainerElement.querySelectorAll('md-option') as NodeListOf<HTMLElement>;
-    let spy = fixture.componentInstance.select;
+    let spy = fixture.componentInstance.optionSelected;
 
     options[1].click();
     tick();
@@ -1568,10 +1568,36 @@ describe('MdAutocomplete', () => {
 
     expect(spy).toHaveBeenCalledTimes(1);
 
-    let event = spy.calls.mostRecent().args[0] as MdAutocompleteSelect;
+    let event = spy.calls.mostRecent().args[0] as MdAutocompleteSelectedEvent;
 
     expect(event.source).toBe(fixture.componentInstance.autocomplete);
     expect(event.option.value).toBe('Washington');
+  }));
+
+  it('should emit an event when a newly-added option is selected', fakeAsync(() => {
+    let fixture = TestBed.createComponent(AutocompleteWithSelectEvent);
+
+    fixture.detectChanges();
+    fixture.componentInstance.trigger.openPanel();
+    tick();
+    fixture.detectChanges();
+
+    fixture.componentInstance.states.push('Puerto Rico');
+    fixture.detectChanges();
+
+    let options = overlayContainerElement.querySelectorAll('md-option') as NodeListOf<HTMLElement>;
+    let spy = fixture.componentInstance.optionSelected;
+
+    options[3].click();
+    tick();
+    fixture.detectChanges();
+
+    expect(spy).toHaveBeenCalledTimes(1);
+
+    let event = spy.calls.mostRecent().args[0] as MdAutocompleteSelectedEvent;
+
+    expect(event.source).toBe(fixture.componentInstance.autocomplete);
+    expect(event.option.value).toBe('Puerto Rico');
   }));
 });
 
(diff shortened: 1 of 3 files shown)


src/lib/autocomplete/autocomplete.ts
```diff
@@ -50,6 +58,9 @@ export class MdAutocomplete implements AfterContentInit {
   /** Function that maps an option's control value to its display value in the trigger. */
   @Input() displayWith: (value: any) => string;
 
+  /** Event that is emitted whenever an option from the list is selected. */
+  @Output() select: EventEmitter<MdAutocompleteSelect> = new EventEmitter<MdAutocompleteSelect>();
```
DevVersion: Shouldn't we just expose the `Observable` instead of the `EventEmitter`?

> Also as a side-question: What about a `change` event?

crisbeto: I wanted to avoid `change`, because typing in the select also changes the value. Regarding the event, using the `EventEmitter` is the way to handle `@Output`.

DevVersion: `@Output` can be basically any `Observable`. `EventEmitter` is just like a RxJS `Subscription`. We originally changed all `@Output`s to just expose the `Observable`. See https://github.com/angular/material2/issues/440

DevVersion: Actually it looks like a lot of components don't do that anymore. @jelbourn any reason why we didn't continue just exposing the observable?

crisbeto: This is pretty much the only use for the `EventEmitter`: https://angular.io/docs/ts/latest/api/core/index/EventEmitter-class.html

rafaelss95:
What the status of this PR? Can it be merged please?

src/lib/autocomplete/autocomplete.ts
willshowell: @crisbeto @DevVersion is this resolved? I don't want this PR to be forgotten!

kevincaradant: +1 to have fix this problem :). I would to know when this PR will be accepted. Thank you

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -1146,6 +1147,29 @@ describe('MdAutocomplete', () => {
         expect(panel.classList).toContain(visibleClass, `Expected panel to be visible.`);
       });
     }));
+
+  it('should call emit an event when an option is selected', fakeAsync(() => {
```
kara: Typo: 'should call emit an event' -> 'should emit an event'

src/lib/autocomplete/autocomplete.ts
```diff
@@ -9,16 +9,24 @@ import {
...
 
+
 /**
  * Autocomplete IDs need to be unique across components, so this counter exists outside of
  * the component definition.
  */
 let _uniqueAutocompleteIdCounter = 0;
 
+/** Event object that is emitted when an autocomplete option is selected */
+export class MdAutocompleteSelect {
```
kara: How about `MdAutocompleteSelectEvent` ? Without "Event", it almost sounds like a hybrid md-autocomplete / md-select directive.

src/lib/autocomplete/autocomplete.ts
kara: @jelbourn and I discussed and we like `(optionSelected)`. It's a bit more explicit, and otherwise it sounds a bit too much like md-select.

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -1146,6 +1147,29 @@ describe('MdAutocomplete', () => {
...
+    options[1].click();
+    tick();
+    fixture.detectChanges();
+
+    expect(spy).toHaveBeenCalledTimes(1);
+
+    let event = spy.calls.mostRecent().args[0] as MdAutocompleteSelect;
+
+    expect(event.source).toBe(fixture.componentInstance.autocomplete);
+    expect(event.option.value).toBe('Washington');
```
kara: Can you add a test for when the option list changes? e.g. emitting an event from a new option.

crisbeto:
Thanks for the feedback @kara. It's addressed now, can you take another look?

src/lib/autocomplete/autocomplete.ts
```diff
@@ -18,16 +18,26 @@ import {
...
  * the component definition.
  */
 let _uniqueAutocompleteIdCounter = 0;
 
+/** Event object that is emitted when an autocomplete option is selected */
+export class MdAutocompleteSelectedEvent {
+  constructor(public source: MdAutocomplete, public option: MdOption) { }
+}
+
+export type AutocompletePositionY = 'above' | 'below';
```
kara: Where is this being used?  Seems unrelated to this PR.

Issue #3664, opened 2017-03-18
Proposal(feature): Expansion Panels


httpNick:
#### Proposal:
Add expansion panels as outlined [here](https://material.io/guidelines/components/expansion-panels.html#expansion-panels-usage) to the material2 pool of components. I have a (very) base setup project that I was using to start implementing the expansion panels element. Before I got any further, and since I am following the material spec, I wanted to propose moving this into the official material2 code base and modify what I have to follow the material2 coding standards/guidelines. My plan was to fork this repository and integrate the basics of what I have to fit into this code base. I have researched pretty heavily into the datepicker that is in progress in order to understand how styling and themes are applied to material componets, so I could work on getting my current progress modified right away to fit the material2 practices in use. 

Also, a question - is this a duplication of work? I have looked around a bit and haven't seen anything started for expansion panels (at least that follow this spec). 

I'd love to hear the communities thoughts!

fxck:
@jelbourn is currently working on table, but I doubt this will be part of the initial release 

https://github.com/jelbourn/material2/tree/table-prototype
https://github.com/jelbourn/material2/tree/new-table-api

best you can do is to start with a design doc https://docs.google.com/document/d/1kpQStAn9L-NX_1bGz_4fXNeZgerWC0mUJHun72bF240/preview

httpNick:
@fxck Thanks! I will start creating the design doc and post it back onto this proposal when ready for review.

jelbourn:
Hi @httpNick:

Could you just add a brief comment about the API and use-cases you're imagining for an expansion panel?

httpNick:
@jelbourn sorry to be MIA, I have some updates for the previous question you asked.

The basic API I was thinking of is here (in the README): https://github.com/httpNick/material2/tree/expansion-panels/src/lib/expansion-panel

I closely followed https://material.io/guidelines/components/expansion-panels.html#expansion-panels-specs in designing the selectors to include, and also I made the expansion-panel being editable an option via `@Input editable`.

The current branch I have adds the expansion-panel to the repo following how other components have been added. I also created a demo-app section for it. It's still a WIP of course, but I thought I was at a good spot to get some feedback before moving forward more.

Let me know if I need to answer on anything else (I will respond more quickly this time 😄 ), or give a better answer!

jelbourn:
@httpNick since my last comment @josephperrott starting working on this already :)

[...]

Issue #995, opened 2016-08-11
Proposal: md-expansion-panel


Dona278:
#### Bug, feature request, or proposal:

Proposal
#### What is the expected behavior?

Required in the first version of angular material and not implemented, a list of expandable panel.
[Specs](https://material.google.com/components/expansion-panels.html)
#### What is the current behavior?

None
#### What is the use-case or motivation for changing an existing behavior?

it's an awesome component for setting's section of a user profile or application
#### Is there anything else we should know?

I wrote the first implementation for this component.
Usage:

```
<md-expansion-panel-group>
    <md-expansion-panel>
        <md-expansion-panel-header>
            <md-expansion-panel-header-label>
                <span class="md-primary">Label here...</span>
                <span class="md-secondary">Secondary label here...</span>
            </md-expansion-panel-header-label>
            <md-expansion-panel-header-content>
                 Header content here...
            </md-expansion-panel-header-content>
        </md-expansion-panel-header>
        <md-expansion-panel-content>
            Panel content here...
        </md-expansion-panel-content>
        <md-expansion-panel-actions>
            Panel actions here like as...
            <button md-button>CLOSE</button>
            <button md-button>SAVE</button>
        </md-expansion-panel-actions>
    </md-expansion-panel>
</md-expansion-panel-group>
```

`md-expansion-panel-group`
The outer container for a group of panels.
Basic only one panel at a time can be expanded but you can change the behavior by `md-multiple- expansions` property.

`md-expansion-panel`
A panel that expand / collapse its content by clicking on the header.
User can navigate between panels by TAB and ENTER keys.

`md-expansion-panel-header`
The only clickable section when the panel is collapsed.

`md-expansion-panel-header-label`
The title of the panel. With `md-primary` and `md-secondary` classes you can set two line.

`md-expansion-panel-header-content`
A short grouping of the content of the panel.

`md-expansion-panel-content`
Custom content of the panel.

`md-expansion-panel-actions`
Custom action of the panel like a dialog buttons.

cesarp:
This is pretty decent, not sure if you are aware of but helped me. 

https://teradata.github.io/covalent/#/components/expansion-panel

Nodarii:
would be nice to have this feature in the list of feature statuses in the main readme file

axtho:
Here is another good (visual) example of this: http://materializecss.com/collapsible.html

naadydev:
We are looking forward to this nice feature to be onboard 
[...]

PR #4191, opened 2017-04-21
feat(expansion-panel): initial version of expansion panel

josephperrott:
For #3664 and #995

### DIFF
--- a/src/demo-app/expansion/expansion-demo.html
+++ b/src/demo-app/expansion/expansion-demo.html
@@ -1,0 +1,52 @@
+<h1>Single Expansion Panel</h1>
+<md-expansion-panel class="md-expansion-demo-width" #myPanel>
+  <md-expansion-panel-header>
+    <mat-panel-description>This is a panel description.</mat-panel-description>
+    <mat-panel-title>Panel Title</mat-panel-title>
+  </md-expansion-panel-header>
+    This is the content text that makes sense here.
+  <md-action-row>
+    <button md-button (click)="myPanel.expanded = false">CANCEL</button>
+    <button md-button>SAVE</button>
+  </md-action-row>
+</md-expansion-panel>
+<br>
+<h1>Accordion</h1>
+<div>
+  <p>Accordion Options</p>
+  <div>
+    <md-slide-toggle [(ngModel)]="multi">Allow Multi Expansion</md-slide-toggle>
+    <md-slide-toggle [(ngModel)]="hideToggle">Hide Indicators</md-slide-toggle>
+    <md-slide-toggle [(ngModel)]="showPanel3">Show Panel 3</md-slide-toggle>
+  </div>
+  <p>Accordion Style</p>
+  <md-radio-group [(ngModel)]="displayMode">
+    <md-radio-button value="default">Default</md-radio-button>
+    <md-radio-button value="flat">Flat</md-radio-button>
+  </md-radio-group>
+  <p>Accordion Panel(s)</p>
+  <div>
+    <md-checkbox [(ngModel)]="panel1.expanded">Panel 1</md-checkbox>
+    <md-checkbox [(ngModel)]="panel2.expanded">Panel 2</md-checkbox>
+  </div>
+</div>
+<br>
+<md-accordion [displayMode]="displayMode" [multi]="multi"
+              class="md-expansion-demo-width">
+  <md-expansion-panel #panel1  [hideToggle]="hideToggle">
+    <md-expansion-panel-header>Section 1</md-expansion-panel-header>
+    <p>This is the content text that makes sense here.</p>
+  </md-expansion-panel>
+  <md-expansion-panel #panel2 [hideToggle]="hideToggle">
+    <md-expansion-panel-header>Section 2</md-expansion-panel-header>
+    <p>This is the content text that makes sense here.</p>
+  </md-expansion-panel>
+  <md-expansion-panel #panel3 *ngIf="showPanel3" [hideToggle]="hideToggle">
+    <md-expansion-panel-header>Section 3</md-expansion-panel-header>
+    <md-checkbox #showButtons>Reveal Buttons Below</md-checkbox>
+    <md-action-row *ngIf="showButtons.checked">
+      <button md-button (click)="panel2.expanded = true">OPEN SECTION 2</button>
+      <button md-button (click)="panel3.expanded = false">CLOSE</button>
+    </md-action-row>
+  </md-expansion-panel>
+</md-accordion>
--- a/src/lib/expansion/accordion-item.ts
+++ b/src/lib/expansion/accordion-item.ts
@@ -1,0 +1,72 @@
+import {Output, EventEmitter, Input, Injectable, OnDestroy, Optional} from '@angular/core';
+import {UniqueSelectionDispatcher} from '../core';
+import {CdkAccordion} from './accordion';
+
+/** Used to generate unique ID for each expansion panel. */
+let nextId = 0;
+
+/**
+ * An abstract class to be extended and decorated as a component.  Sets up all
+ * events and attributes needed to be managed by a CdkAccordion parent.
+ */
+@Injectable()
+export class AccordionItem implements OnDestroy {
+  /** Event emitted every time the MdAccordianChild is closed. */
+  @Output() closed = new EventEmitter<void>();
+  /** Event emitted every time the MdAccordianChild is opened. */
+  @Output() opened = new EventEmitter<void>();
+  /** Event emitted when the MdAccordianChild is destroyed. */
+  @Output() destroyed = new EventEmitter<void>();
+  /** The unique MdAccordianChild id. */
+  readonly id = `cdk-accordion-child-${nextId++}`;
+  /** Whether the MdAccordianChild is expanded. */
+  @Input() get expanded(): boolean { return this._expanded; }
+  set expanded(expanded: boolean) {
+    // Only emit events and update the internal value if the value changes.
+    if (this._expanded !== expanded) {
+      this._expanded = expanded;
+      if (expanded) {
+        this.opened.emit();
+        /**
+         * In the unique selection dispatcher, the id parameter is the id of the CdkAccordonItem,
+         * the name value is the id of the accordion.
+         */
+        let accordionId = this.accordion ? this.accordion.id : this.id;
+        this._expansionDispatcher.notify(this.id, accordionId);
+      } else {
+        this.closed.emit();
+      }
+    }
+  }
+  private _expanded: boolean;
+
+  constructor(@Optional() public accordion: CdkAccordion,
+              protected _expansionDispatcher: UniqueSelectionDispatcher) {
+     _expansionDispatcher.listen((id: string, accordionId: string) => {
+       if (this.accordion && !this.accordion.multi &&
+           this.accordion.id === accordionId && this.id !== id) {
+         this.expanded = false;
+       }
+     });
+    }
+
+  /** Emits an event for the accordion item being destroyed. */
+  ngOnDestroy() {
+    this.destroyed.emit();
+  }
+
+  /** Toggles the expanded state of the accordion item. */
+  toggle(): void {
+    this.expanded = !this.expanded;
+  }
+
+  /** Sets the expanded state of the accordion item to false. */
+  close(): void {
+    this.expanded = false;
+  }
+
+  /** Sets the expanded state of the accordion item to true. */
+  open(): void {
+    this.expanded = true;
+  }
+}
--- a/src/lib/expansion/accordion.ts
+++ b/src/lib/expansion/accordion.ts
@@ -1,0 +1,50 @@
+import {Directive, Input} from '@angular/core';
+import {coerceBooleanProperty} from '../core/coercion/boolean-property';
+
+/** MdAccordion's display modes. */
+export type MdAccordionDisplayMode = 'default' | 'flat';
+
+/** Unique ID counter */
+let nextId = 0;
+
+/**
+ * Directive whose purpose is to manage the expanded state of CdkAccordionItem children.
+ */
+@Directive({
+  selector: '[cdk-accordion]',
+})
+export class CdkAccordion {
+  /** A readonly id value to use for unique selection coordination. */
+  readonly id = `cdk-accordion-${nextId++}`;
+
+  /** Whether the accordion should allow multiple expanded accordion items simulateously. */
+  @Input() get multi(): boolean { return this._multi; }
+  set multi(multi: boolean) { this._multi = coerceBooleanProperty(multi); }
+  private  _multi: boolean = false;
+
+  /** Whether the expansion indicator should be hidden. */
+  @Input() get hideToggle(): boolean { return this._hideToggle; }
+  set hideToggle(show: boolean) { this._hideToggle = coerceBooleanProperty(show); }
+  private  _hideToggle: boolean = false;
+
+  /**
+   * The display mode used for all expansion panels in the accordion. Currently two display
+   * modes exist:
+   *   default - a gutter-like spacing is placed around any expanded panel, placing the expanded
+   *     panel at a different elevation from the reset of the accordion.
+   *  flat - no spacing is placed around expanded panels, showing all panels at the same
+   *     elevation.
+   */
+  @Input() displayMode: MdAccordionDisplayMode = 'default';
+}
+
+/**
+ * Directive for a Material Design Accordion.
+ */
+@Directive({
+  selector: 'mat-accordion, md-accordion',
+  host: {
+    class: 'mat-accordion'
+  }
+})
+export class MdAccordion extends CdkAccordion {}
--- a/src/lib/expansion/expansion-panel-header.html
+++ b/src/lib/expansion/expansion-panel-header.html
@@ -1,0 +1,7 @@
+<span class="mat-content">
+  <ng-content select="md-panel-title, mat-panel-title"></ng-content>
+  <ng-content select="md-panel-description, mat-panel-description"></ng-content>
+  <ng-content></ng-content>
+</span>
+<span [@indicatorRotate]="_getExpandedState()" *ngIf="!_getHideToggle()" 
+      class="mat-expansion-indicator"></span>
--- a/src/lib/expansion/expansion-panel-header.scss
+++ b/src/lib/expansion/expansion-panel-header.scss
@@ -1,0 +1,61 @@
+$mat-expansion-panel-header-height: 48px;
+$mat-expansion-panel-header-height-expanded: 64px;
+
+.mat-expansion-panel-header {
+  cursor: pointer;
+  display: flex;
+  flex-direction: row;
+  height: $mat-expansion-panel-header-height;
+  line-height: $mat-expansion-panel-header-height;
+  padding: 0 24px;
+
+  &.mat-expanded {
+    height: $mat-expansion-panel-header-height-expanded;
+    line-height: $mat-expansion-panel-header-height-expanded;
+  }
+
+  &:focus, 
+  &:hover {
+    outline: none;
+  }
+
+  &.mat-expanded:focus,
+  &.mat-expanded:hover, {
+    background: inherit;
+  }
+}
+
+.mat-content {
+  display: flex;
+  flex: 1;
+  flex-direction: row;
+  overflow: hidden;
+}
+
+.mat-expansion-panel-header-title {
+  display: flex;
+  flex-grow: 1;
+  font-size: 15px;
+  margin-right: 16px;
+}
+
+.mat-expansion-panel-header-description {
+  display: flex;
+  flex-grow: 2;
+  font-size: 15px;
+  margin-right: 16px;
+}
+
+/** 
+ * Creates the expansion indicator arrow. Done using ::after rather than having
+ * additional nodes in the template.
+ */
+.mat-expansion-indicator::after {
+  border-style: solid;
+  border-width: 0 2px 2px 0;
+  content: '';
+  display: inline-block;
+  padding: 3px;
+  transform: rotate(45deg);
+  vertical-align: middle;
+}
--- a/src/lib/expansion/expansion-panel-header.ts
+++ b/src/lib/expansion/expansion-panel-header.ts
@@ -1,0 +1,122 @@
+import {
+  Component,
+  Directive,
+  Host,
+  ViewEncapsulation,
+} from '@angular/core';
+import {
+  trigger,
+  state,
+  style,
+  transition,
+  animate,
+} from '@angular/animations';
+import {SPACE, ENTER} from '../core/keyboard/keycodes';
+import {MdExpansionPanel, EXPANSION_PANEL_ANIMATION_TIMING} from './expansion-panel';
+
+
+/**
+ * <md-expansion-panel-header> component.
+ *
+ * This component corresponds to the header element of an <md-expansion-panel>.
+ *
+ * Please refer to README.md for examples on how to use it.
+ */
+@Component({
+  moduleId: module.id,
+  selector: 'md-expansion-panel-header, mat-expansion-panel-header',
+  styleUrls: ['./expansion-panel-header.css'],
+  templateUrl: './expansion-panel-header.html',
+  encapsulation: ViewEncapsulation.None,
+  host: {
+    'class': 'mat-expansion-panel-header',
+    'role': 'button',
+    'tabindex': '0',
+    '[attr.aria-controls]': '_getPanelId()',
+    '[attr.aria-expanded]': '_isExpanded()',
+    '[class.mat-expanded]': '_isExpanded()',
+    '(click)': '_toggle()',
+    '(keyup)': '_keyup($event)',
+    '[@expansionHeight]': '_getExpandedState()',
+  },
+  animations: [
+    trigger('indicatorRotate', [
+      state('collapsed', style({transform: 'rotate(0deg)'})),
+      state('expanded', style({transform: 'rotate(180deg)'})),
+      transition('expanded <=> collapsed', animate(EXPANSION_PANEL_ANIMATION_TIMING)),
+    ]),
+    trigger('expansionHeight', [
+      state('collapsed', style({height: '48px', 'line-height': '48px'})),
+      state('expanded', style({height: '64px', 'line-height': '68px'})),
+      transition('expanded <=> collapsed', animate(EXPANSION_PANEL_ANIMATION_TIMING)),
+    ]),
+  ],
+})
+export class MdExpansionPanelHeader {
+  constructor(@Host() public panel: MdExpansionPanel) {}
+
+  /** Toggles the expanded state of the panel. */
+  _toggle(event?: KeyboardEvent): void {
+    this.panel.toggle();
+  }
+
+  /** Gets whether the panel is expanded. */
+  _isExpanded(): boolean {
+    return this.panel.expanded;
+  }
+
+  /** Gets the expanded state string of the panel. */
+  _getExpandedState(): string {
+    return this.panel._getExpandedState();
+  }
+
+  /** Gets the panel id. */
+  _getPanelId(): string {
+    return this.panel.id;
+  }
+
+  /** Gets whether the expand indicator is hidden. */
+  _getHideToggle(): boolean {
+    return this.panel.hideToggle;
+  }
+
+  /** Handle keyup event calling to toggle() if appropriate. */
+  _keyup(event: KeyboardEvent) {
+    switch (event.keyCode) {
+      // Toggle for space and enter keys.
+      case SPACE:
+      case ENTER:
+        event.preventDefault();
+        this._toggle();
+        break;
+      default:
+        return;
+    }
+  }
+}
+
+/**
+ * <md-panel-description> directive.
+ *
+ * This direction is to be used inside of the MdExpansionPanelHeader component.
+ */
+@Directive({
+  selector: 'md-panel-description, mat-panel-description',
+  host : {
+    class: 'mat-expansion-panel-header-description'
+  }
+})
+export class MdExpansionPanelDescription {}
+
+/**
+ * <md-panel-title> directive.
+ *
+ * This direction is to be used inside of the MdExpansionPanelHeader component.
+ */
+@Directive({
+  selector: 'md-panel-title, mat-panel-title',
+  host : {
+    class: 'mat-expansion-panel-header-title'
+  }
+})
+export class MdExpansionPanelTitle {}
--- a/src/lib/expansion/expansion-panel.html
+++ b/src/lib/expansion/expansion-panel.html
@@ -1,0 +1,8 @@
+<ng-content select="mat-expansion-panel-header, md-expansion-panel-header"></ng-content>
+<div [class.mat-expanded]="expanded" class="mat-expansion-panel-content"
+     [@bodyExpansion]="_getExpandedState()" [id]="id">
+  <div class="mat-expansion-panel-body">
+    <ng-content></ng-content>
+  </div>
+  <ng-content select="mat-action-row, md-action-row"></ng-content>
+</div>
--- a/src/lib/expansion/expansion-panel.scss
+++ b/src/lib/expansion/expansion-panel.scss
@@ -1,0 +1,33 @@
+@import '../core/style/variables';
+@import '../core/style/elevation';
+
+.mat-expansion-panel {
+  @include mat-elevation-transition;
+  box-sizing: content-box;
+  display: block;
+
+  &:not([class*='mat-elevation-z']) {
+    @include mat-elevation(2);
+  }
+}
+
+.mat-expansion-panel-content {
+  overflow: hidden;
+}
+
+.mat-expansion-panel-body {
+  padding: 0 24px 16px;
+}
+
+.mat-action-row {
+  border-top-style: solid;
+  border-top-width: 1px;
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  padding: 16px 8px 16px 24px;
+
+  button.mat-button {
+    margin-left: 8px;
+  }
+}
--- a/src/lib/expansion/expansion-panel.ts
+++ b/src/lib/expansion/expansion-panel.ts
@@ -1,0 +1,106 @@
+import {
+  Component,
+  Directive,
+  Host,
+  Input,
+  ViewEncapsulation,
+  Optional,
+  forwardRef,
+} from '@angular/core';
+import {
+  trigger,
+  state,
+  style,
+  transition,
+  animate,
+} from '@angular/animations';
+import {MdAccordion, MdAccordionDisplayMode} from './accordion';
+import {AccordionItem} from './accordion-item';
+import {UniqueSelectionDispatcher} from '../core';
+
+
+/** MdExpansionPanel's states. */
+export type MdExpansionPanelState = 'expanded' | 'collapsed';
+
+/** Time and timing curve for expansion panel animations. */
+export const EXPANSION_PANEL_ANIMATION_TIMING = '225ms cubic-bezier(0.4,0.0,0.2,1)';
+
+/**
+ * <md-expansion-panel> component.
+ *
+ * This component can be used as a single element to show expandable content, or as one of
+ * multiple children of an element with the CdkAccordion directive attached.
+ *
+ * Please refer to README.md for examples on how to use it.
+ */
+@Component({
+  moduleId: module.id,
+  styleUrls: ['./expansion-panel.css'],
+  selector: 'md-expansion-panel, mat-expansion-panel',
+  templateUrl: './expansion-panel.html',
+  encapsulation: ViewEncapsulation.None,
+  host: {
+    'class': 'mat-expansion-panel',
+    '[class.mat-expanded]': 'expanded',
+    '[@displayMode]': '_getDisplayMode()',
+  },
+  providers: [
+    {provide: AccordionItem, useExisting: forwardRef(() => MdExpansionPanel)}
+  ],
+  animations: [
+    trigger('bodyExpansion', [
+      state('collapsed', style({height: '0px'})),
+      state('expanded', style({height: '*'})),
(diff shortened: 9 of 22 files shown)


src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** Whether the panel set should allow multiple open panels. */
+  @Input('multi') multi: boolean = false;
+  /** A list of subscriptions for panel open events. */
+  private _panelSubscriptions: Subscription[] = [];
+  /** QueryList of all expansion panels in the expansion panel set. */
+  @ContentChildren(forwardRef(() => MdExpansionPanel)) private _panels: QueryList<MdExpansionPanel>;
+
+  /** Set up event subscriptions for all panels for when they open, closing the other panels. */
+  ngAfterContentInit() {
+    this._panels.forEach((panel: MdExpansionPanel, idx: number, panels: MdExpansionPanel[]) => {
```
crisbeto: This won't handle if a panel gets added later on. You can subscribe to the `this._panels.changes`. Alternatively the panels could register themselves with the parent instead.

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+      let panelId = panel.getExpansionPanelId();
+      let otherPanels = panels.filter(otherPanel => otherPanel.getExpansionPanelId() !== panelId);
+      let panelSubscription = panel.open.asObservable().subscribe(() => {
```
crisbeto: The `asObservable` here is redundant`.

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+          otherPanels.forEach(otherPanel => otherPanel.expanded = false);
+        }
+      });
+      this._panelSubscriptions.push(panelSubscription);
+    });
+  }
+
+  /** Clean up panel subscriptions. */
+  ngOnDestroy() {
+    this._panelSubscriptions.forEach(subscription => subscription.unsubscribe());
```
crisbeto: Instead of saving multiple subscriptions in an array and dealing with them individually, you can use `Observable.merge` to merge them into a single stream.

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+    '[class.mat-expansion-panel]': 'true',
+    '[class.expanded]': 'expanded',
+    '[attr.wide]': 'wideStyling',
+    '[@flatStyling]': 'booleanToString(!flatStyling && expanded)',
+    '[@wideStyling]': 'booleanToString(wideStyling && expanded)',
+  },
+  animations: [
+    trigger('bodyExpansion', [
+      state('false', style({height: '0px', 'padding-bottom': '0'})),
+      state('true', style({height: '*', 'padding-bottom': '16px'})),
```
crisbeto: Instead of using booleans to represent the animation state, it might be better to use something like `expanded` and `collapsed`.

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+      transition('false <=> true', animate(EXPANSION_PANEL_ANIMATION_TIMING)),
+    ]),
+    trigger('flatStyling', [
+      state('false', style({'margin-top': '*', 'margin-bottom': '*'})),
+      state('true', style({'margin-top': '16px', 'margin-bottom': '16px'})),
```
crisbeto: We should avoid animating the `margin` here and the `padding` a bit above since it's not very performant.

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** Whether the expansion panel is expanded. */
+  private _expanded: boolean;
+  @Input('expanded')
+  get expanded(): boolean {
+    return this._expanded;
+  }
+  set expanded(expanded: boolean) {
+    this._expanded = expanded;
+    this.expandChange.emit(expanded);
+    if (expanded) {
```
crisbeto: You can shorten this one to `expanded ? this.open.emit() : this.close.emit()`. Also shouldn't the event be called `closed`?

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** Whether flat styling should be used on expansion. */
+  get flatStyling(): boolean {
+    return !this.panelSet || this.panelSet.style === 'flat';
+  }
+  /** Whether wide styling should be used on expansion. */
+  get wideStyling(): boolean {
+    return this.panelSet && this.panelSet.style === 'wide';
+  }
+  /** Whether the expansion panel is expanded. */
+  private _expanded: boolean;
```
crisbeto: Might be a good idea to add a default value to this one.

src/lib/expansion-panel/expansion-summary.scss
```diff
@@ -0,0 +1,60 @@
...
+
+.mat-summary {
+  cursor: pointer;
+  display: flex;
+  flex-direction: row;
+  height: $md-summary-height;
+  line-height: $md-summary-height;
+  padding: 0 24px;
+
+  .content {
```
crisbeto: This class should be prefixed with `mat-` and not be nested.

src/lib/expansion-panel/expansion-summary.scss
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  &.expanded {
+    height: $md-summary-height-expanded;
+    line-height: $md-summary-height-expanded;
+  }
+
+  &:focus, 
+  &:hover {
+    background: rgb(238, 238, 238);
```
crisbeto: The color should be inside the theme. The same goes for `[title]`, `[description]` etc.

src/lib/expansion-panel/expansion-summary.scss
```diff
@@ -0,0 +1,60 @@
...
+    outline: none;
+  }
+
+  &.expanded:focus,
+  &.expanded:hover, {
+    background: initial;
```
crisbeto: AFAIK `initial` doesn't work on IE.

src/lib/expansion-panel/expansion-summary.ts
```diff
@@ -0,0 +1,51 @@
...
+  templateUrl: './expansion-summary.html',
+  encapsulation: ViewEncapsulation.None,
+  host: {
+    'tabindex': '0',
+    'aria-role': 'button',
+    '[attr.aria-controls]': 'parentPanel.getExpansionPanelId()',
+    '[attr.aria-expanded]': 'parentPanel.expanded',
+    '[class.mat-summary]': 'true',
+    '[class.expanded]': 'parentPanel.expanded',
+    '(click)': 'parentPanel.expanded = !parentPanel.expanded',
```
crisbeto: It would be better if these expressions were in a method inside the class, instead of inline.

src/lib/expansion-panel/expansion-panel.scss
```diff
@@ -0,0 +1,43 @@
...
+.mat-expansion-panel-set {
+  display: block;
+}
+
+.mat-expansion-panel {
+  @include mat-elevation-transition;
+  box-sizing: content-box;
+  display: block;
+
+  &:not(:first-child) {
```
jelbourn: `.mat-expansion-panel` could be the first child of an arbitrary div

src/lib/expansion-panel/expansion-panel.scss
```diff
@@ -0,0 +1,43 @@
...
+    @include mat-elevation(2);
+  }
+
+  &[wide='true'] {
+    margin-left: 24px;
+    margin-right: 24px;
+  }
+}
+
+div.md-expansion-panel-body {
```
jelbourn: Why is `div` part of this selector?

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+
+/** Used to generate unique ID for each expansion panel. */
+let nextId = 0;
+/** Time and timing curve for expansion panel animations. */
+export const EXPANSION_PANEL_ANIMATION_TIMING = '225ms cubic-bezier(0.4,0.0,0.2,1)';
+
+@Directive({
+  selector: 'md-expansion-panel-set, mat-expansion-panel-set',
+  host: {
+    '[class.mat-expansion-panel-set]': 'true',
```
jelbourn: You can just use a static class here, e.g.:
```ts
host: {'class': 'mat-expansion-panel-set'},
```
So long as it's not a `'[class]'` binding (which will clobber other classes)

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+} from '@angular/animations';
+import {Subscription} from 'rxjs/Subscription';
+
+/** Used to generate unique ID for each expansion panel. */
+let nextId = 0;
+/** Time and timing curve for expansion panel animations. */
+export const EXPANSION_PANEL_ANIMATION_TIMING = '225ms cubic-bezier(0.4,0.0,0.2,1)';
+
+@Directive({
+  selector: 'md-expansion-panel-set, mat-expansion-panel-set',
```
jelbourn: A group of expansion panels is typically called an accordion. Accordions also have special a11y treatment on top of the standalone panel: https://www.w3.org/TR/wai-aria-practices-1.1/#accordion. It's possible that this maybe be a completely separate component that just shares some code with the expansion panel.

Maybe leave this out for now and add it in a follow-up?

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+  host: {
+    '[class.mat-expansion-panel-set]': 'true',
+  },
+})
+export class MdExpansionPanelSet implements AfterContentInit, OnDestroy {
+  /** Whether the panel set should use flat styling. */
+  @Input('style') style: 'default' | 'flat' | 'wide' = 'default';
```
jelbourn: The input name doesn't need to be specified if it's the same as the property name.

jelbourn: The property `style` collides with the `style` property on the html element

jelbourn: What are these different styles? I don't see anything about them in the spec. Is this something the user wouldn't be able to accomplish with css?

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** Whether the panel set should allow multiple open panels. */
+  @Input('multi') multi: boolean = false;
```
jelbourn: This should be a boolean property that uses `coerceBooleanProperty`

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+@Component({
+  moduleId: module.id,
+  styleUrls: ['./expansion-panel.css'],
+  selector: 'md-expansion-panel, mat-expansion-panel',
+  templateUrl: './expansion-panel.html',
+  encapsulation: ViewEncapsulation.None,
+  host: {
+    '[class.mat-expansion-panel]': 'true',
```
jelbourn: Same thing about `'class'` here

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (expanded) {
+      this.open.emit();
+    } else {
+      this.close.emit();
+    }
+  }
+  /** Unique identifier for the expansion panel, to be used by aria attributes. */
+  private _expansionPanelId: number;
+  /** Event emitted for each expanded state change. */
+  @Output('expandChange') expandChange = new EventEmitter<boolean>();
```
jelbourn: Don't need the name when it's the same as the property

src/lib/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** Event emitted every time the expansion panel is closed. */
+  @Output('close') close = new EventEmitter<null>();
```
jelbourn: If there's no event object you should do `EventEmitter<void>`

josephperrott:
@jelbourn ready for you to take a look

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /** Clean up panel subscriptions. */
+  ngOnDestroy() {
+    this._destroySubject.next();
+    this._destroySubject.complete();
+  }
+
+  /** Registers a QueryList of panels. */
+  private registerChildren(panels: QueryList<CdkAccordionChild>) {
```
jelbourn: Use `_` prefix for any non-public API

(differs from internal TS style guide)

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+    this._children.forEach(panel => this.registerChild(panel));
+  }
+
+  /**
+   * Registers a CdkAccordionChild panel if has not already been registered.  Creates subscriptions
+   * for the open and destroy events for the panel.
+   */
+  private registerChild(child: CdkAccordionChild) {
```
jelbourn: Rather than going through all of this work to have the accordion manage which panels are open and closed, why not have the individual items do this using `UniqueSelectionDispatcher`? I think that would be significantly less code.

src/lib/expansion/expansion-panel-header.scss
```diff
@@ -0,0 +1,60 @@
...
+
+.mat-expansion-panel-header {
+  cursor: pointer;
+  display: flex;
+  flex-direction: row;
+  height: $md-expansion-panel-header-height;
+  line-height: $md-expansion-panel-header-height;
+  padding: 0 24px;
+
+  .mat-content {
```
jelbourn: [Avoid using nested styles](https://github.com/angular/material2/blob/master/CODING_STANDARDS.md#use-lowest-specificity-possible). Should be something like
```scss
.mat-expansion-panel-header {
  // ...
}

.mat-expansion-panel-content {
// ...
}
```

src/lib/expansion/expansion-panel-header.scss
```diff
@@ -0,0 +1,60 @@
...
+    background: rgb(238, 238, 238);
+    outline: none;
+  }
+
+  &.mat-expanded:focus,
+  &.mat-expanded:hover, {
+    background: inherit;
+  }
+
+  [title] {
```
jelbourn: Prefer class selectors over attribute selectors (especially for such common attributes as `title` and `description`)

src/lib/expansion/expansion-panel-header.scss
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  &.mat-expanded {
+    height: $md-expansion-panel-header-height-expanded;
+    line-height: $md-expansion-panel-header-height-expanded;
+  }
+
+  &:focus, 
+  &:hover {
+    background: rgb(238, 238, 238);
```
jelbourn: Color should be captured in the theme
(here and elsewhere)

src/lib/expansion/expansion-panel-header.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  &.mat-expanded:focus,
+  &.mat-expanded:hover, {
+    background: inherit;
+  }
+
+  [title] {
+    color: rgba(0, 0, 0, 0.87);
+    display: flex;
+    flex: 1;
```
jelbourn: Explicitly put `flex-grow` for clarity? I personally can never remember the order of the `flex` shorthand

src/lib/expansion/expansion-panel-header.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  [description] {
+    color: rgba(0, 0, 0, 0.54);
+    display: flex;
+    flex: 2;
+    font-size: 15px;
+    margin-right: 16px;
+  }
+  
+  .mat-expansion-indicator::after {
```
jelbourn: Need comment explaining what this `::after` is for

src/lib/expansion/expansion-panel-header.html
```diff
@@ -0,0 +1,7 @@
+<span class="mat-content">
+  <ng-content select="[title]"></ng-content>
+  <ng-content select="[description]"></ng-content>
```
jelbourn: Nothing in your demo uses `title` or `description`- what are these?

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -0,0 +1,74 @@
...
+      state('expanded', style({height: '64px', 'line-height': '68px'})),
+      transition('expanded <=> collapsed', animate(EXPANSION_PANEL_ANIMATION_TIMING)),
+    ]),
+  ],
+})
+export class MdExpansionPanelHeader {
+  constructor(@Host() public panel: MdExpansionPanel) {}
+
+  /** Flips the expanded state of the panel. */
+  changeExpandedState(): void {
```
jelbourn: `toggle`?

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -0,0 +1,74 @@
...
+  host: {
+    'class': 'mat-expansion-panel-header',
+    'role': 'button',
+    'tabindex': '0',
+    '[attr.aria-controls]': 'getPanelId()',
+    '[attr.aria-expanded]': 'isExpanded()',
+    '[class.mat-expanded]': 'isExpanded()',
+    '(click)': 'changeExpandedState()',
+    '(keyup.space)': 'changeExpandedState()',
+    '(keyup.enter)': 'changeExpandedState()',
```
jelbourn: The `keyup.space` syntax unfortunately doesn't work in Safari 9, so we have to manually check the key code in the handler.

src/lib/expansion/expansion-panel.scss
```diff
@@ -0,0 +1,31 @@
...
+  padding: 0 24px;
+}
+
+[mat-action-row],
+[md-action-row] {
+  border-top: 1px solid rgba(0, 0, 0, 0.12);
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  margin: 10px -24px 0 -24px;
```
jelbourn: What's with the negative margins? I usually find needing this to be a code smell

src/lib/expansion/expansion-panel.ts
```diff
@@ -0,0 +1,108 @@
...
+  /** Whether the expansion indicator should be shown. */
+  getShowExpandIndicator() {
+    if (this.accordion) {
+      return this.accordion.showExpandIndicator;
+    }
+    return this.showExpandIndicator;
+  }
+
+  /** Gets the panel style. */
+  getPanelStyle(): string {
```
jelbourn: Not clear what "panel style" is

src/lib/expansion/expansion-panel.ts
```diff
@@ -0,0 +1,108 @@
...
+      return this.getExpandedState();
+    }
+    if (this.accordion) {
+      return this.accordion.accordionStyle;
+    }
+    return this.getExpandedState();
+  }
+
+  /** Gets the expanded state string. */
+  getExpandedState(): string {
```
jelbourn: Return type should be `'expanded' | 'collapsed'`, probably as a `type` definition.

src/lib/expansion/expansion.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    TestBed.configureTestingModule({
+      imports: [MdExpansionModule],
+      declarations: [
+        PanelWithContent
+      ],
+    });
+    TestBed.compileComponents();
+  }));
+
+  it('should apply expanded classes as appropriate', () => {
```
jelbourn: I would just make this `should expand and collapse the panel` with tests that check the css class and the absence/presence of the content.

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export class CdkAccordionChild implements OnDestroy {
+  /** Event emitted every time the MdAccordianChild is closed. */
+  @Output() close = new EventEmitter<null>();
+  /** Event emitted every time the MdAccordianChild is opened. */
+  @Output() open = new EventEmitter<null>();
+  /** Event emitted when the MdAccordianChild is destroyed. */
+  @Output() destroy = new EventEmitter<null>();
+  /** Whether the MdAccordianChild is expanded. */
+  expanded: boolean;
```
josephperrott: Since this isn't an `@Input`, we have control over this being a `boolean` here correct?

In the abstract, it would just be a `boolean`, and then in the implementation (`MdExpansionPanel`) it would use `coerceBooleanProperty`.

jelbourn: Yep; made the comment before I realized this was abstract

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+} from '@angular/animations';
+import {coerceBooleanProperty} from '../core/coercion/boolean-property';
+import {Subscription} from 'rxjs/Subscription';
+import {Subject} from 'rxjs/Subject';
+import 'rxjs/add/operator/takeUntil';
+
+/** Used to generate unique ID for each expansion panel. */
+let nextId = 0;
+
+export class CdkAccordionChild implements OnDestroy {
```
josephperrott: Done

josephperrott: Done

josephperrott: Yep

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  /** Event emitted every time the MdAccordianChild is closed. */
+  @Output() close = new EventEmitter<null>();
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  /** Event emitted every time the MdAccordianChild is opened. */
+  @Output() open = new EventEmitter<null>();
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  /** Event emitted when the MdAccordianChild is destroyed. */
+  @Output() destroy = new EventEmitter<null>();
+  /** Whether the MdAccordianChild is expanded. */
+  expanded: boolean;
+
+  /** The unique MdAccordianChild id. */
+  private _id: number = nextId++;
+  get id(): string { return `cdk-accordion-child-${this._id}`; }
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  ngOnDestroy() {
+    this.destroy.emit();
+  }
+}
+
+@Directive({
+  selector: '[cdk-accordion]',
+})
+export class CdkAccordion implements AfterContentInit, OnDestroy {
+  /** Whether the expansion indicator should be shown. */
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  private  _showIndicator: boolean = true;
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  @Input()
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  get showExpandIndicator(): boolean { return this._showIndicator; }
+  set showExpandIndicator(show: boolean) { this._showIndicator = coerceBooleanProperty(show); }
+  /** Whether the panel set should use flat styling. */
+  @Input() accordionStyle: string = 'default';
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  /** Whether the panel set should allow multiple open panels. */
+  private  _multi: boolean = false;
+  @Input()
+  get multi(): boolean { return this._multi; }
+  set multi(multi: boolean) { this._multi = coerceBooleanProperty(multi); }
+
+  /** A subject to be completed on destroy to clean up subscriptions. */
+  private _destroySubject = new Subject();
+  /** A list of subscriptions for panel open events. */
+  private _childrenSet: Set<CdkAccordionChild> = new Set();
```
josephperrott: Ended up using `Map` since we needed to be able to iterate through the `Map`.

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  /** QueryList of all expansion panels in the expansion panel set. */
+  @ContentChildren(CdkAccordionChild) private _children: QueryList<CdkAccordionChild>;
+
+  /** Set up event subscriptions for all panels for when they open, closing the other panels. */
+  ngAfterContentInit() {
+    this.registerChildren(this._children);
+    this._children.changes
+      .takeUntil(this._destroySubject)
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+      .subscribe((children: QueryList<CdkAccordionChild>) => {
+        this.registerChildren(children);
+      });
```
josephperrott: Done

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,120 @@
...
+  @Input()
+  get multi(): boolean { return this._multi; }
+  set multi(multi: boolean) { this._multi = coerceBooleanProperty(multi); }
+
+  /** A subject to be completed on destroy to clean up subscriptions. */
+  private _destroySubject = new Subject();
+  /** A list of subscriptions for panel open events. */
+  private _childrenSet: Set<CdkAccordionChild> = new Set();
+  /** QueryList of all expansion panels in the expansion panel set. */
+  @ContentChildren(CdkAccordionChild) private _children: QueryList<CdkAccordionChild>;
```
josephperrott: Done

src/lib/expansion/accordion.ts
josephperrott: Done

src/lib/expansion/accordion.ts
josephperrott: The reason I went with this method was to have the responsibility for multiple selection fall on `MdAccordion` as the parent.

With `UniqueSelectionDispatcher` the children would be responsible for checking with the parent to determine if the "unique" portion is currently true.

src/lib/expansion/expansion-panel-header.scss
josephperrott: Done, some remain as they are multi-class selectors.

src/lib/expansion/expansion-panel-header.scss
josephperrott: Done

src/lib/expansion/expansion-panel-header.scss
josephperrott: Done

src/lib/expansion/expansion-panel-header.scss
josephperrott: Done

src/lib/expansion/expansion-panel-header.html
josephperrott: This is for matching the material spec for the headers.

The first demo actually does use it, updated the demo to pack it a bit more obvious

src/lib/expansion/expansion-panel-header.ts
josephperrott: Fixed

src/lib/expansion/expansion-panel-header.ts
josephperrott: Updated

src/lib/expansion/expansion-panel.scss
josephperrott: It was originally needed because a weird spacing issue, it essentially, just offset the containers padding.  I was able to adjust it to no longer be needed.

src/lib/expansion/expansion-panel.ts
josephperrott: Updated.

src/lib/expansion/expansion.spec.ts
josephperrott: Done.

src/lib/expansion/expansion-panel.ts
josephperrott: Done.

src/lib/expansion/expansion-panel-header.scss
josephperrott: Done.

src/lib/expansion/accordion.ts
jelbourn: It would make sense to do this the same way as radio buttons, though; each panel could have a name that determines the exclusive selection (or expansion) behavior. When they panels are inside of a group (accordion), the group could just set names to the items. This would maintain the parent knowing about the children but not other way around thing, while also making the individual panels more flexible in applying exclusive selection behavior. (also I think the code would be simpler)

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,101 @@
...
+export type CdkAccordionDisplayMode = 'default' | 'flat';
+
+@Directive({
+  selector: '[cdk-accordion]',
+})
+export class CdkAccordion implements AfterContentInit, OnDestroy {
+  private  _showToggle: boolean = true;
+  /** Whether the expansion indicator should be shown. */
+  @Input() get showToggle(): boolean { return this._showToggle; }
+  set showToggle(show: boolean) { this._showToggle = coerceBooleanProperty(show); }
```
jelbourn: Add blank line between separate properties

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,101 @@
...
+import 'rxjs/add/operator/takeUntil';
+import {CdkAccordionItem} from './accordion-item';
+
+export type CdkAccordionDisplayMode = 'default' | 'flat';
+
+@Directive({
+  selector: '[cdk-accordion]',
+})
+export class CdkAccordion implements AfterContentInit, OnDestroy {
+  private  _showToggle: boolean = true;
```
jelbourn: Invert this to `hideToggle`?

src/lib/expansion/expansion-panel-header.html
```diff
@@ -0,0 +1,7 @@
+<span class="mat-content">
+  <ng-content select=".title"></ng-content>
+  <ng-content select=".description"></ng-content>
+  <ng-content></ng-content>
+</span>
+<span [@indicatorRotate]="_getExpandedState()" *ngIf="_getShowToggle()"
```
jelbourn: When I pull up the demo, this is always invisible

src/demo-app/expansion/expansion-demo.html
```diff
@@ -0,0 +1,58 @@
+<h1>Single Expansion Panel</h1>
+<md-expansion-panel class="md-expansion-demo-width" #myPanel>
+  <md-expansion-panel-header>
+    <span class="description">This is a panel description.</span>
+    <span class="title">Panel Title</span>
```
jelbourn: If title and description are part of the template API they should be elements like `md-panel-header-title`, `md-panel-header-description`
(sorry, I had previously thought you were just styling something in the expansion panel's template)

src/lib/expansion/accordion-item.ts
```diff
@@ -0,0 +1,22 @@
+import {Output, EventEmitter, OnDestroy} from '@angular/core';
+
+/** Used to generate unique ID for each expansion panel. */
+let nextId = 0;
+
+export abstract class CdkAccordionItem implements OnDestroy {
```
jelbourn: Needs class description

josephperrott:
@jelbourn ready for you to take a look agian

src/demo-app/expansion/expansion-demo.html
```diff
@@ -0,0 +1,58 @@
+<h1>Single Expansion Panel</h1>
+<md-expansion-panel class="md-expansion-demo-width" #myPanel>
+  <md-expansion-panel-header>
+    <span class="md-expansion-panel-header-description">This is a panel description.</span>
+    <span class="md-expansion-panel-header-title">Panel Title</span>
```
jelbourn: Sorry, was still unclear. I had meant last time (plus shortening the names somewhat):
```html
<md-expansion-panel class="md-expansion-demo-width" #myPanel>
  <md-panel-header>
    <md-panel-description> ... </>
    <md-panel-title> ... </>
  </>
</>
```

src/lib/expansion/accordion-item.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/** Used to generate unique ID for each expansion panel. */
+let nextId = 0;
+
+
+/**
+ * An abstract class to be extended and decorated as a component.  Sets up all
+ * events and attributes needed to be managed by a CdkAccordion parent.
+ */
+export abstract class CdkAccordionItem implements OnDestroy {
```
josephperrott: Can definitely do this, but should we keep the `Cdk` prefix since it should be used as the super for a Directive/Component.

src/lib/expansion/index.ts
```diff
@@ -0,0 +1,41 @@
+import {NgModule, ModuleWithProviders} from '@angular/core';
+import {CommonModule} from '@angular/common';
+import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
```
arlowhite: Is this **BrowserAnimationsModule** import a problem?
I merged this PR into master and did a build, but my App now has this error:

> Error: BrowserModule has already been loaded.

I modified `publish-build-artifacts.sh` to deploy to my repo and ran it.
See https://github.com/angular/material2-builds/compare/master...arlowhite:master
Load diff for **material.js**
`+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';`
You can see that this import has been added.

_ng version_

> @angular/cli: 1.0.4
> node: 6.9.1
> os: linux x64
> @angular/common: 4.1.3
> @angular/material: 2.0.0-beta.5-25de2f2

arlowhite: I removed the import, but I still get the error. Maybe I'm not building material2 correctly.
But I did confirm that no other material2 module imports BrowserAnimationsModule.

arlowhite:
The arrow icon is reversed from what the Material Guideline recommends:
https://material.io/guidelines/components/expansion-panels.html#expansion-panels-behavior
Arrow should point down when collapsed.

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -0,0 +1,123 @@
...
+  _getHideToggle(): boolean {
+    return this.panel.hideToggle;
+  }
+
+  /** Handle keyup event calling to toggle() if appropriate. */
+  _keyup(event: KeyboardEvent) {
+    switch (event.keyCode) {
+      // Toggle for space and enter keys.
+      case SPACE:
+      case ENTER:
```
jelbourn: Probably want to `preventDefault` on these (space will scroll and enter may submit a form)

src/lib/expansion/expansion-panel.html
```diff
@@ -0,0 +1,8 @@
+<ng-content select="mat-expansion-panel-header, md-expansion-panel-header"></ng-content>
+<div [class.mat-expanded]="expanded" class="mat-expansion-panel-content"
+     [@bodyExpansion]="_getExpandedState()" [id]="id">
+  <div class="mat-expansion-panel-body">
+    <ng-content></ng-content>
+  </div>
+  <ng-content select="[mat-action-row], [md-action-row]"></ng-content>
```
jelbourn: These should use an element selector (e.g. `<md-action-row>`)

src/lib/expansion/accordion.ts
```diff
@@ -0,0 +1,39 @@
...
+export type CdkAccordionDisplayMode = 'default' | 'flat';
+
+/** Unique ID counter */
+let nextId = 0;
+
+/**
+ * Directive whose purpose is to manage the expanded state of CdkAccordionItem children.
+ */
+@Directive({
+  selector: '[cdk-accordion]',
```
jelbourn: Also, we should make a separate `<md-accordion>`  that is probably just
```ts
@Directive({ selector: 'md-accordion'})
export class MdAccordion extends CdkAccordion { }
```
Just so users don't have to mix `md-` and `cdk-`

josephperrott: The idea for it being an attribute rather than an element is because it is simply a containing item for the children who control there own spacing and styling.  Since there is nothing in this directive that affects the document flow, it seemed to make sense that you could just mark another element as the containing element for the children.

src/lib/expansion/accordion-item.ts
```diff
@@ -0,0 +1,56 @@
...
+        let accordionId = this.accordion ? this.accordion.id : this.id;
+        this._uniqueSelectionDispatcher.notify(this.id, accordionId);
+      } else {
+        this.closed.emit();
+      }
+    }
+  }
+  private _expanded: boolean;
+
+  constructor(@Optional() @Host() public accordion: CdkAccordion,
```
josephperrott: My thought was if you had a `CdkAccordion` inside of another `CdkAccordion` that you wanted to be sure you got the correct `CdkAccordion`, so we only look at the host component.  However now that you point it out I see how flawed that logic is as it will get the first one it finds on its way up the tree.

josephperrott:
@jelbourn Comments addressed ready for a look

jelbourn:
@andrewseguin this one will need new build rules when syncing

otienoanyango:
Really excited about this. Great job guys!

jakubkrupka:
Even in beta, awesome.
When sb does something good, they should be told about this. Nice work then!

PR #4202, opened 2017-04-21
build: use shared logic to build packages

DevVersion:
* No longer creates the whole set of gulp tasks just to build a single package.
* Makes the package building & releasing more consistent. No extra `build:release` just for the `lib` package.

**Note**: Madge, Payload, Publish Artifacts will be updated in a follow-up.

References #4108


tools/gulp/tasks/material-release.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Overwrite the release task for the material package. The material release will include special
+ * files, like a bundled theming SCSS file or all prebuilt themes.
+ */
+task('material:build-release', ['material:prepare-release'], () => composeRelease('material'));
+
+/**
+ * Task that will build the material package. It will also copy all prebuilt themes and build
+ * a bundled SCSS file for theming
+ */
+task('material:prepare-release', sequenceTask(
```
DevVersion: We had that before for a lot of different tasks. But I just don't feel it's necessary. Just makes it harder to remind, if something is "supposed" to be called from the CLI or not.

tools/gulp/gulpfile.ts
```diff
@@ -1,16 +1,21 @@
+import {createPackageBuildTasks} from './util/package-tasks';
+
+/** Create gulp tasks to build the different packages in the project. */
+createPackageBuildTasks('cdk');
+createPackageBuildTasks('material', ['cdk']);
+createPackageBuildTasks('material-examples', ['material']);
```
DevVersion: The name in the `package.json` will be `@angular/material-examples`. Before I went with `material-examples` I had another property on `createPackageBuildTasks` that allows you to specify the task-prefix.

```ts
createPackageBuildTasks('material-examples', 'examples', ['material']);
```
But this just causes confusion because the package and task name are different. Actually the same for `lib` (= `material`).

We should try to be as consistent as possible.

package.json
```diff
@@ -8,7 +8,7 @@
     "url": "https://github.com/angular/material2.git"
   },
   "scripts": {
-    "build": "gulp build:release",
+    "build": "gulp material:build-release:clean",
```
DevVersion: Not sure about that. It's very nice to have like all tasks that belong to a different "package" / "group" be prefixed with the actual name of it.

```ts
cdk:clean-build
cdk:build
cdk:build-release
cdk:watch
```

tools/gulp/gulpfile.ts
jelbourn: The project/repo is already _called_ "material", though; I'd say that part of the name is implicit. As someone working on it, my mental model is just "I want to build the examples" -> "build:examples".

package.json
jelbourn: I think of it as "I want to perform some action. That action will affect xxx". Overall I like the consistency of "verb:action" more than organizing the tasks by target

tools/gulp/gulpfile.ts
DevVersion: I totally see what you mean. My only concern is that it is very confusing to have a mismatch of package-name / task-name / source directory name.

package.json
DevVersion: Yeah I see that it might be nicer when building different packages on their own. I personally vote for organizing the tasks by target. 

It just feels not very consistent if you want to copy for example the assets of the demo-app.

- e.g `copy-assets:demo-app`. 

I would be fine revisiting later in another PR or issue.

tools/gulp/util/package-tasks.ts
```diff
@@ -0,0 +1,70 @@
...
+import {join} from 'path';
+import {main as tsc} from '@angular/tsc-wrapped';
+import {SOURCE_ROOT, DIST_ROOT} from '../constants';
+import {sequenceTask, sassBuildTask, copyTask, triggerLivereload} from './task_helpers';
+import {buildPackageBundles, composeRelease} from './package-build';
+
+// There are no type definitions available for these imports.
+const inlineResources = require('../../../scripts/release/inline-resources');
+
+export function createPackageBuildTasks(packageName: string, requiredPackages: string[] = [], ) {
```
DevVersion: Done.

tools/gulp/tasks/material-release.ts
```diff
@@ -0,0 +1,53 @@
...
+  'material:build',
+  ['material:prebuilt-themes', 'material:theming-scss']
+));
+
+/** Copies all prebuilt themes into the release package under `prebuilt-themes/` */
+task('material:prebuilt-themes', () => {
```
DevVersion: Done.

PR #4209, opened 2017-04-22
chore: add CanDisable mixin to checkbox, radio-group, slide-toggle, and slider

jelbourn:
Also fixed a couple of unnecessary host class bindings while I was there.

### DIFF
--- a/src/lib/radio/radio.ts
+++ b/src/lib/radio/radio.ts
@@ -26,6 +26,7 @@ import {
   FocusOrigin,
 } from '../core';
 import {coerceBooleanProperty} from '../core/coercion/boolean-property';
+import {mixinDisabled, CanDisable} from '../core/common-behaviors/disabled';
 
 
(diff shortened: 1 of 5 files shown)


src/lib/radio/radio.ts
```diff
@@ -369,8 +366,7 @@ export class MdRadioButton implements OnInit, AfterViewInit, OnDestroy {
   }
 
   set disabled(value: boolean) {
-    // The presence of *any* disabled value makes the component disabled, *except* for false.
-    this._disabled = (value != null && value !== false) ? true : null;
+    this._disabled = coerceBooleanProperty(value);
```
mmalerba: Can we not use the mixin for the individual radio buttons too?

jelbourn: The getter for indivual buttons is different (it checks the parent state as well as its own)

Issue #1329, opened 2016-09-26
Set menu position with binding


ashleahhill:
#### Bug, feature request, or proposal:

Feature Request
#### What is the -expected- desired behavior?

To be able to bind values to the x-position and y-position inputs of the md-menu.
#### What is the current behavior?

I can't set the x-position or y-position by binding to a value

Changing MyMenu's template to include this 

``` html
<md-menu x-position="{{before}}" y-position="{{after}}" #menu="mdMenu">
```

Causes a template parse error, (because x-position and y-position are attributes, not inputs?).

```
Can't bind to 'y-position' since it isn't a known property of 'md-menu'.
1. If 'md-menu' is an Angular component and it has 'y-position' input, then verify that it is part of this module.
2. If 'md-menu' is a Web Component then add "CUSTOM_ELEMENTS_SCHEMA" to the '@NgModule.schema' of this component to suppress this message.
```
#### What are the steps to reproduce?

Can't get plunker to work with md-menu at the moment.
#### What is the use-case or motivation for changing an existing behavior?

For example, I want create a component that lets me drop in a menu with a list of items with custom behaviors already attached. Its template is basically this: 

``` html
<button [md-menu-trigger-for]="menu">
Open Menu
</button>

<md-menu x-position="before" y-position="after" #menu="mdMenu">
    <button md-menu-item
      *ngFor="let item of menuItems"
      [item]="item"
      (action)="action(item)">
{{item.name}}
    </button>
</md-menu>
```

Sometimes, I am going to use these menus in a context where I have space on the left, or space on the right, or space above, ect. I want to be able to set in the x-position and y-position in the parent component and have it set on the menu.

``` html
<md-toolbar>
<span flex></span>
<my-menu [x-position]="before" [menu-items]="menuItemList">
</my-menu>
</md-toolbar>
```
#### Which versions of Angular, Material, OS, browsers are affected?

Angular 2.0.0
Material 8-2

crisbeto:
Closing as a dupe of https://github.com/angular/material2/issues/4169 which will be resolved via https://github.com/angular/material2/pull/4213.

PR #4213, opened 2017-04-22
fix(menu): unable to bind to xPosition and yPosition

### LINKED ISSUE #4169: Can't bind to 'xPosition' since it isn't a known property of 'md-menu'.
Is there any reason [these](https://github.com/angular/material2/blob/master/src/lib/menu/menu-directive.ts#L57
) are `@Attribute`s instead of `@Input`s?

I'm trying to pass down the position from a parent component.

@kara @jelbourn

crisbeto:
Fixes not being able to use data bindings on the `xPosition` and `yPosition` properties.

**Note:** These changes could be considered breaking, if somebody was implementing the `MdMenuPanel` interface.

Fixes #4169.

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -2,7 +2,6 @@
 
 import {
   AfterContentInit,
-  Attribute,
   Component,
   ContentChildren,
   EventEmitter,
@@ -37,6 +36,8 @@ import {transformMenu, fadeInItems} from './menu-animations';
 })
 export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   private _keyManager: FocusKeyManager;
+  private _xPosition: MenuPositionX = 'after';
+  private _yPosition: MenuPositionY = 'below';
 
   /** Subscription to tab events on the menu panel */
   private _tabSubscription: Subscription;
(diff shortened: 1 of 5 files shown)


willshowell:
Fixes #4169

src/lib/menu/menu-directive.ts
```diff
@@ -45,27 +49,42 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
...
+  /** List of the items inside of a menu. */
   @ContentChildren(MdMenuItem) items: QueryList<MdMenuItem>;
+
+  /** Whether the menu should overlap its trigger. */
   @Input() overlapTrigger = true;
 
-  constructor(@Attribute('xPosition') posX: MenuPositionX,
-              @Attribute('yPosition') posY: MenuPositionY,
-              @Attribute('x-position') deprecatedPosX: MenuPositionX,
+  constructor(@Attribute('x-position') deprecatedPosX: MenuPositionX,
```
kara: While we're in here, let's remove the deprecated positions?

src/lib/menu/menu-directive.ts
```diff
@@ -81,6 +100,12 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
     }
   }
 
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes['xPosition'] || changes['yPosition']) {
+      this.setPositionClasses();
```
kara: Curious why we are adding a hook to call `setPositionClasses` when we already have the setters above. I think it might be cleaner not to break up the logic.

crisbeto: Not sure I understood the question. The setters aren't dealing with the classes at all.

kara: The classes need to be updated whenever the xPosition or the yPosition changes.  For that reason, it makes more sense to me to call this method in the xPosition and yPosition setters rather than creating a hook just to listen for changes on those properties.

crisbeto:
Addressed the feedback @kara.

willshowell:
Also fixes #1329

mmalerba:
There are some teams in google that extend MdMenuPanel, we will need to update them before merging

Issue #4155, opened 2017-04-19
ConnectedPositionStrategy doesn't work properly if the overlay's content is dynamic


mjamin:
#### Bug, feature request, or proposal:

Bug

#### What is the expected behavior?

Position the overlay correctly within the viewport if possible.

#### What is the current behavior?

Overlay is positioned exceeding the viewport if the overlay's contents are dynamic and there's no fixed width set.

#### What are the steps to reproduce?

https://plnkr.co/edit/9hUThjYmfC7ofpSBDdKU

If the overlay's content is dynamic, the actual dimensions of the overlay content are set after the positioning logic runs.

Workaround:

    setTimeout(() => overlayRef.updatePosition()); // queue update when overlay is shown

#### Which versions of Angular, Material, OS, browsers are affected?

Angular: 4.0.2
Material: 2.0.0-beta.3

Issue #3997, opened 2017-04-09
md-menu max-height is not calculated properly 


fxck:
this is unrelated to https://github.com/angular/material2/issues/2725 (actually looking at the issue again, kara mentions it) / https://github.com/angular/material2/pull/3918

this gist shows the current behaviour as well as behaviour once #3918 is merged and a situation where neither would help https://gist.run/?id=58d5fecb56107042ea0d8786f6a23f5d

I'm not sure what the intended behaviour was supposed to be, but I think it should be one of these

1.a)  menu takes in account the position of the trigger, and open the panel which has height of `window height - 48px - position of the trigger`, ending up with this
![image](https://cloud.githubusercontent.com/assets/1303561/24838883/025b69a0-1d51-11e7-82f3-ba40a31b2a86.png)

1.b) same as 1a, but it open above, when current position of the trigger in view port is closer to button, ending up with this 
![image](https://cloud.githubusercontent.com/assets/1303561/24838896/3d7d9dd2-1d51-11e7-9fda-07aa411219e7.png)

2) it doesn't take in account the position of the trigger, but expands above and below the trigger, filling up the whole page
![image](https://cloud.githubusercontent.com/assets/1303561/24838918/783067c0-1d51-11e7-8e04-1ce5142749ba.png)

3) it stays 100vh - 48px, but actually expands the body height, so you can actually scroll to the part that goes off screen (I don't think this is neither really possible given how overlay works, nor good for UX)

@crisbeto @kara @jelbourn @M-a-c

jelbourn:
Going to track this via #6534 now

PR #4214, opened 2017-04-22
fix(menu): incorrect panel max height

### LINKED ISSUE #2725: md-menu is defaulted to `max-height: calc(100vh + 48px);` and goes off page

crisbeto:
Fixes the menu panel having a wrong `max-height` and uses a variables to reduce it, instead of a hardcoded value.

Relates to #2725.

### DIFF
--- a/src/lib/menu/menu.scss
+++ b/src/lib/menu/menu.scss
@@ -11,9 +11,7 @@ $mat-menu-vertical-padding: 8px !default;
 .mat-menu-panel {
   @include mat-menu-base();
   @include mat-menu-positions();
-
-  // max height must be 100% of the viewport height + one row height
-  max-height: calc(100vh + 48px);
+  max-height: calc(100vh - #{$mat-menu-item-height});
 
   @include cdk-high-contrast {
     outline: solid 1px;


fxck:
This is not really going to help that much I think. It needs to take in count the initial position of the trigger as well.

https://github.com/angular/material2/issues/3997

crisbeto:
Hmm true, but that may need to wait for https://github.com/angular/material2/issues/4155. Messing with the `max-height` dynamically will throw off the positioning if it's set to `above`, with the current setup.

PR #4218, opened 2017-04-22
Revert "fix(slide-toggle): invalid model change event (#4140)"

kara:
This reverts commit 317952a18dba01b2d1e78c310c9f1f0ff8237e00 because of test failures in Google. It looks like the change breaks initial toggling of the slide-toggle (it takes two clicks on the label before it will begin to toggle).


DevVersion:
Tested it locally and it seems to work. Can you ping me on Slack about it?

PR #4220, opened 2017-04-22
fix(slide-toggle): invalid model change event

### LINKED ISSUE #4124: slide-toggle fires undesired event when setting a default value

DevVersion:
While initially looking into #4124, there were a few more issues inside
of the slide-toggle.

* ngModelChange event is dispatched at initialization.

* Checked state isn't synchronized when state changes through drag. New state after dragging got overwritten by click event on label.

* Removes unnecessary logic inside of `change` listener. Change event doesn't fire if underlying checkbox is disabled.

Fixes #4124.

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -547,18 +547,21 @@ describe('MdSlideToggle with forms', () => {
       expect(slideToggleModel.pristine).toBe(true);
       expect(slideToggleModel.touched).toBe(false);
 
-      // After changing the value programmatically, the control should
-      // become dirty (not pristine), but remain untouched.
+      // After changing the value from the view, the control should
+      // become dirty (not pristine), but remain untouched if focus is still there.
       slideToggle.checked = true;
-      fixture.detectChanges();
+
+      // Dispatch a change event on the input element to fake a user interaction that triggered
+      // the state change.
+      dispatchFakeEvent(inputElement, 'change');
 
       expect(slideToggleModel.valid).toBe(true);
       expect(slideToggleModel.pristine).toBe(false);
       expect(slideToggleModel.touched).toBe(false);
 
-      // After a user interaction occurs (such as a click), the control should remain dirty and
-      // now also be touched.
-      labelElement.click();
+      // Once the input element loses focus, the control should remain dirty but should
+      // also turn touched.
+      dispatchFakeEvent(inputElement, 'blur');
       fixture.detectChanges();
 
       expect(slideToggleModel.valid).toBe(true);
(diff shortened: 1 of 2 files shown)


DevVersion:
Ah sorry, didn't assign @tinayuangao for review here, because @kara reverted the PR before 😄

DevVersion:
@tinayuangao Can you review this again? This basically is the old PR just with the `Cast NgModel value to boolean` commit (that should fix the internal google issues)

tinayuangao:
LGTM

DevVersion:
@tinayuangao Thanks. @kara I'm going to mark this as merge ready now. Let me know if there are any failures again.

jelbourn:
Presubmit error is that one team's e2e tests are failure, but it's not clear why. We need to follow-up with that team.

DevVersion:
Rebased.

src/lib/slide-toggle/slide-toggle.spec.ts
```diff
@@ -295,18 +295,23 @@ describe('MdSlideToggle', () => {
...
       fixture.detectChanges();
 
       expect(slideToggleModel.valid).toBe(true);
       expect(slideToggleModel.pristine).toBe(false);
       expect(slideToggleModel.touched).toBe(false);
 
-      // After a user interaction occurs (such as a click), the control should remain dirty and
-      // now also be touched.
-      labelElement.click();
+      // Once the input element looses focus, the control should remain dirty but should
```
kara: Typo: `loses`

DevVersion: Fixed

kara:
Assigning back to @DevVersion so we can debug one last testing issue.

DevVersion:
@kara I think the issue should be fixed now. Added a test to ensure that this won't happen again.

DevVersion:
@tinayuangao @kara Can you review this? I removed `lgtm` because I changed a few things.

mmalerba:
Please rebase

DevVersion:
@mmalerba Done!

e-cloud:
any progress?

DevVersion:
Rebased. I just need more information what's causing the presubmit failures. cc. @mmalerba

PR #4236, opened 2017-04-24
fix(card): Input to change the card elevation

### LINKED ISSUE #4233: feat(card): Input to customize the elevation
feature

Input on md-card to set the elevation of the card

Now we can change the card elevation with the css classes `mat-elevation-z[0..24]`, but it would be nice if the final user doen't necessarily have to remember the class name as the component can change itself

Some cards of my app needs to be in a different elevation than the default `mat-elevation-z2` and it's more straighforwart if we can use something like: `<md-card [elevation]="3">...</md-card>`

jefersonestevo:
Fixes #4233

### DIFF
--- a/src/demo-app/card/card-demo.html
+++ b/src/demo-app/card/card-demo.html
@@ -58,4 +58,12 @@
       <button md-button>Second</button>
     </md-card-actions>
   </md-card>
+
+  <md-card [elevation]="elevation">
+    <md-card-title>With Custom Elevation</md-card-title>
+    <md-select [(ngModel)]="elevation" name="elevation" placeholder="Elevation">
+      <md-option>Default</md-option>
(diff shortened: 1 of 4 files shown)


dahaupt:
@jefersonestevo I really appreciate your work, but wouldn't it be better to discuss those features before implementing? If the team doesn't like the idea you've done your efforts for nothing.

jefersonestevo:
@dahaupt I agree with you that it may be a waste of time if the team doesn't like the idea, but I had it already implemented this as a directive on my app, so the real 'extra' work I did here besides the one I had already done was to create the `card.spec.ts`. Anyway the next time I'll wait for the discussion prior to really make the PR

jelbourn:
Thanks for the PR! 

I'd really like to keep the card as behavior-free as possible. In this case, I don't think it's necessary to add an input and associated cost of binding / generated code to accomplish what's normally done with a css class.

PR #4237, opened 2017-04-24
fix(input-container): New attribute hideRequiredMarker

### LINKED ISSUE #3681: Add md-no-asterisk to inputs
request 

There is no ability to remove asterisks from placeholders in inputs with required attribute.

We need an ability to disable asterisks on required inputs like in [the first version of material UI](https://material.angularjs.org/latest/api/directive/mdInput) by using "md-no-asterisk" attribute in input. 

In the end, it would be good enough to have syntax like in the following example:
[Plunker md-no-asterisk example](http://plnkr.co/edit/vROp6Q7ATSRDsdYWCLNW?p=preview)

For now, there are two hacks to remove asterisks from inputs:

1. do not use "required" attribute
2. use "required" attribute with hacks that would hide or remove automatically added asterisk

Both options are crutches. And as you know crutches are really bad things.

jefersonestevo:
This attribute will enable the user to hide the required marker (star) from an mdInput even when it's required

Fixes #3681

### DIFF
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -245,6 +245,14 @@
       </md-input-container>
     </p>
     <p>
+      <md-checkbox [(ngModel)]="hideRequiredMarker"> Check to hide the required marker:</md-checkbox>
+      <md-input-container [hideRequiredMarker]="hideRequiredMarker" style="width: 300px">
+        <input mdInput
+               required
+               [placeholder]="hideRequiredMarker ? 'Required Without Marker' : 'Required With Marker'">
+      </md-input-container>
+    </p>
+    <p>
       <md-button-toggle-group [(ngModel)]="floatingLabel">
         <md-button-toggle value="auto">Auto Float</md-button-toggle>
         <md-button-toggle value="always">Always Float</md-button-toggle>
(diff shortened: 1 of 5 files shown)


src/lib/input/input-container.spec.ts
```diff
@@ -371,6 +371,16 @@ describe('MdInputContainer', function () {
     expect(el.nativeElement.textContent).toMatch(/hello\s+\*/g);
   });
 
+  it('hide placeholder required star when set to hide the required marker', () => {
+    let fixture = TestBed.createComponent(MdInputContainerPlaceholderRequiredTestComponent);
+    fixture.componentInstance.hideRequiredMarker = true;
```
mmalerba: check that it has the `*` initially

src/demo-app/input/input-demo.html
```diff
@@ -245,6 +245,14 @@
       </md-input-container>
     </p>
     <p>
+      <md-checkbox [(ngModel)]="hideRequiredMarker"> Check to hide the required marker:</md-checkbox>
```
DevVersion: Can we remove that leading space for the content?

jefersonestevo:
@mmalerba @DevVersion done

PR #4239, opened 2017-04-24
docs(ripple): add section about global options

### LINKED ISSUE #4226: Docs to globally disable ripple
Ripple readme is outdated

DevVersion:
* Adds a new section about global ripple options to the `README.md` file of the ripple directive.

@mmalerba Not sure if the documentation for the `baseSpeedFactor` is good. I had hard times to explain that the factor is like a "percentage". Value `1.5` makes ripples animate 50% faster than normal.

Closes #4226

### DIFF
--- a/src/lib/core/ripple/README.md
+++ b/src/lib/core/ripple/README.md
@@ -6,10 +6,33 @@ By default, a ripple is activated when the host element of the `md-ripple` direc
 
 Ripples can also be triggered programmatically by getting a reference to the MdRipple directive and calling its `start` and `end` methods.
 
+### Global options
 
-### Upcoming work
+Developers are able to specify options for all ripples inside of their application.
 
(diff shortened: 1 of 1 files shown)


src/lib/core/ripple/README.md
```diff
@@ -6,10 +6,33 @@ By default, a ripple is activated when the host element of the `md-ripple` direc
...
+  ]
+})
+```
+
+Here are all available global options listed:
+
+| Name            | Type    | Description                               |
+| --------------- | ------- | ----------------------------------------- |
+| disabled        | boolean | Whether ripples should show or not.       |
+| baseSpeedFactor | number  | Factor to adjust duration of the ripples. |
```
willshowell: I think it would make more sense to say "Factor to adjust ripple speed". The current phrasing makes it seem like higher values result in higher duration.

DevVersion: Yep. I ageee

PR #4246, opened 2017-04-24
feat(datepicker): allow overriding parse and display format in MdDatepickerIntl

mmalerba:


mmalerba:
@jelbourn I'm not the biggest fan of this... the `getPredefinedFormats` stuff feels like it should really be part of material / datepicker, not cdk. But the alternatives are:
1) Force users to fill out the `dateFormat` and `parseDateFormat` on `MdDatepickerIntl` in order to make the datepicker work
2) Providing a corresponding version of MdDatepickerIntl for each DateAdapter that fills out reasonable defaults for each (and force the user to provide both the correct adapter and the correct Intl in order to use the datepicker).

mmalerba:
#4296 is an alternate solution to this problem which I prefer over this PR

PR #4250, opened 2017-04-24
fix(connected-overlay): better handling of dynamic content

### LINKED ISSUE #4155: ConnectedPositionStrategy doesn't work properly if the overlay's content is dynamic

crisbeto:
* Refactors the `ConnectedPositionStrategy` to be able to use `right` and `bottom` to position the panel. This will allow the element to keep its position automatically, even if the element's size changes.
* Removes the uses of the `FakeViewportRuler` in some of the unit tests since it doesn't work very well when the element is positioned relatively to the right edge of the screen.
* Rounds down the values when testing positioning. This is necessary, because some browsers (particularly Chrome on Windows) can have some subpixel deviations.

Fixes #4155.

### DIFF
--- a/src/lib/core/overlay/position/connected-position-strategy.ts
+++ b/src/lib/core/overlay/position/connected-position-strategy.ts
@@ -117,11 +117,10 @@ export class ConnectedPositionStrategy implements PositionStrategy {
       // (top, left) coordinate for the overlay at `pos`.
       let originPoint = this._getOriginConnectionPoint(originRect, pos);
       let overlayPoint = this._getOverlayPoint(originPoint, overlayRect, viewportRect, pos);
-      let overlayDimensions = this._getCSSDimensions(overlayRect, overlayPoint, pos);
 
       // If the overlay in the calculated position fits on-screen, put it there and we're done.
       if (overlayPoint.fitsInViewport) {
-        this._setElementPosition(element, overlayDimensions);
+        this._setElementPosition(element, overlayRect, overlayPoint, pos);
 
         // Save the last connected position in case the position needs to be re-calculated.
         this._lastConnectedPosition = pos;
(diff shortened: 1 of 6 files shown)


crisbeto:
This one's ready to go now @jelbourn.

src/lib/core/overlay/position/connected-position-strategy.ts
```diff
@@ -116,10 +117,11 @@ export class ConnectedPositionStrategy implements PositionStrategy {
       // (top, left) coordinate for the overlay at `pos`.
       let originPoint = this._getOriginConnectionPoint(originRect, pos);
       let overlayPoint = this._getOverlayPoint(originPoint, overlayRect, viewportRect, pos);
+      let overlayDimensions = this._getCSSDimensions(overlayRect, overlayPoint, pos);
```
jelbourn: It's weird to have both `overlayPoint` and `overlayDimensions` since they're both kind of the same thing and it would be very easy to use the wrong one. 

What if we leave the calculations they way they are now, but then also pass the `ConnectionPositionPair` and the overlay `ClientRect` to `_setElementPosition` which figures out if it should set `bottom`/`right` there? I think it might be better, since then you're deferring the calculation until you have to know and don't have to create a type to pass it around.

crisbeto: Done. FWIW, I had it in a separate method, because the logic used to be a bit longer.

src/lib/core/overlay/position/connected-position-strategy.ts
```diff
@@ -346,14 +348,36 @@ export class ConnectedPositionStrategy implements PositionStrategy {
...
+      xValue = viewport.width - (overlayPoint.x + overlayRect.width);
+    }
+
+    ['top', 'bottom', 'left', 'right'].forEach(prop => {
+      if (prop === xProp || prop === yProp) {
+        element.style[prop] = `${prop === xProp ? xValue : yValue}px`;
+      } else {
+        element.style[prop] = null;
+      }
+    });
```
crisbeto: Done with some slight tweaks to the comments to mention that the property determines which direction the element will expand to.

andrewseguin:
Needs rebase

crisbeto:
Rebased @andrewseguin.

PR #4251, opened 2017-04-24
fix(universal): gate several browser-specific bits on being on the browser

jelbourn:
Related to #308

This is really a temporary fix until we have a secondary `@angular/material/server` entry-point which can contain things like `NoopScrollDispatcher`, `NoopRippleRenderer`, etc. I don't want to add these classes to the main `@angular/material` package both because they would eventually move and they would add bloat for client payload.

Also removed a few unnecessary `[class]` bindings while I was around

### DIFF
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -9,9 +9,8 @@ import {
   Renderer2,
   ViewEncapsulation
 } from '@angular/core';
-import {coerceBooleanProperty, FocusOriginMonitor} from '../core';
+import {coerceBooleanProperty, FocusOriginMonitor, Platform} from '../core';
 import {mixinDisabled, CanDisable} from '../core/common-behaviors/disabled';
-import {isBrowser} from '../core/platform/browser';
 
 
 // TODO(kara): Convert attribute selectors to classes when attr maps become available
@@ -111,8 +110,11 @@ export class MdButton extends _MdButtonMixinBase implements OnDestroy, CanDisabl
   get disableRipple() { return this._disableRipple; }
   set disableRipple(v) { this._disableRipple = coerceBooleanProperty(v); }
 
-  constructor(private _elementRef: ElementRef, private _renderer: Renderer2,
-              private _focusOriginMonitor: FocusOriginMonitor) {
+  constructor(
+      private _elementRef: ElementRef,
+      private _renderer: Renderer2,
+      private _platform: Platform,
+      private _focusOriginMonitor: FocusOriginMonitor) {
     super();
     this._focusOriginMonitor.monitor(this._elementRef.nativeElement, this._renderer, true);
   }
@@ -163,7 +165,7 @@ export class MdButton extends _MdButtonMixinBase implements OnDestroy, CanDisabl
     // If not on the browser, say that there are none of the attributes present.
     // Since these only affect how the ripple displays (and ripples only happen on the client),
     // detecting these attributes isn't necessary when not on the browser.
-    if (!isBrowser()) {
+    if (!this._platform.isBrowser) {
       return false;
     }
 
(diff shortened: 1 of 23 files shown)


src/lib/core/style/focus-origin-monitor.ts
```diff
@@ -126,6 +132,11 @@ export class FocusOriginMonitor {
 
   /** Register necessary event listeners on the document and window. */
   private _registerDocumentEvents() {
+    // Do nothing if we're not on the browser platform.
+    if (!isBrowser()) {
+      return Observable.of();
```
mmalerba: `return;` (method isn't expected to return anything)

src/lib/button/button.ts
```diff
@@ -165,6 +156,13 @@ export class MdButton extends _MdButtonMixinBase implements OnDestroy, CanDisabl
    * with either an 'md-' or 'mat-' prefix.
    */
   _hasAttributeWithPrefix(...unprefixedAttributeNames: string[]) {
+    // If not on the browser, say that there are none of the attributes present.
+    // Since these only affect how the ripple displays (and ripples only happen on the client),
+    // detecting these attributes isn't necessary when not on the browser.
+    if (!isBrowser()) {
```
crisbeto: Might be nicer if the return below was `return isBrowser() && unprefixedAttributeNames.some...`.

src/lib/core/overlay/position/viewport-ruler.ts
```diff
@@ -51,6 +54,12 @@ export class ViewportRuler {
    * @param documentRect
    */
   getViewportScrollPosition(documentRect = this._documentRect) {
+    // Cache the document bounding rect so that we don't recompute it for multiple calls.
+    if (!documentRect) {
+      this._cacheViewportGeometry();
```
crisbeto: This could be a little shorter if we made `_cacheViewportGeometry` return the new `documentRect` and then changed the method definition to:

```ts
getViewportScrollPosition(documentRect = this._cacheViewportGeometry()) {
  // continue as usual
}
```

src/lib/core/platform/browser.ts
```diff
@@ -0,0 +1,4 @@
+/** Gets whether the current platform is a web browser (versus node). */
+export function isBrowser() {
+  return typeof document === 'object';
```
crisbeto: A couple of things:
* Technically shouldn't be an issue, but `typeof document === 'object'` will be true if `document` is `null`. 
* Perhaps this should be a `const`, instead of a function?

Also I was doing these checks inline in a few places, I don't know whether we shouldn't switch them to use `isBrowser` with this PR or in a follow-up.

src/lib/core/overlay/position/viewport-ruler.ts
jelbourn: That would then no longer be using the cached version if there is one.

src/lib/core/platform/browser.ts
jelbourn: Changed check to `typeof document === 'object' && document;`

We can get other places in a follow-up.

jelbourn:
Replaced implementation from yesterday with a new approach that adds `isBrowser` to the existing `Platform` injectable. This will make it easier to test/fake in places where that would be necessary.

mmalerba:
LGTM, needs rebase though

jelbourn:
Rebased

jelbourn:
@mmalerba fixed circular dep

mmalerba:
@jelbourn needs rebase and investigation into failing test

jelbourn:
@mmalerba after I rebased the error went away

mmalerba:
@jelbourn sorry, needs more rebasing

jelbourn:
Rebased again- weirdly, git couldn't auto-resolve the conflicts, but WebStorm could

jelbourn:
@andrewseguin rebased again

aitboudad:
tried on my side and I got some little issues:
- missing rxjs `of` import 
- navigator is not defined in [Platform](https://github.com/angular/material2/blob/master/src/lib/core/platform/platform.ts)

mmalerba:
still has some lint issues and failing tests too

jelbourn:
@kara rebased again and green again

ntoscano:
**sitting on the edge of my seat, waiting for the merge**

Thank you to the contributors, btw. This PR is saving my team's ass.

Issue #4245, opened 2017-04-24
md-card does not follow the guidelines.


ArtisanDuCode:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
![](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0Bzhp5Z4wHba3RXpZUWE2bV9PdDQ/components_cards33.png)

I have made a table that summarizes the guidelines at [](https://material.io/guidelines/components/cards.html)

**type** | default
--------- | -------
**card** | border: 2px, margin: 8dp on mobile, margins can be 8, 16, 24, or 40dp wide.
**actions** | padding: 8px
**rich media** 100% | ratio: 16:9 or 1:1
**rich media specific size** | ratio 1:1, padding-top: 16px
**header** | height: 72px, padding: 16px
**title group** | padding-left-right: 16px, padding-top-bottom: 16px or 24px (if no action or content)
**primary title** | text: 24px
**subtitle title** | text: 14px, padding-top:12px
**content** | text: 14px, padding-left-right: 16px, padding-top-bottom: 16px or 24px (if no action or content)

#### What is the current behavior?
The current behavior is reversed.
This is the card that the padding 16px and all element adjust with margin negative.
$mat-card-default-padding: 24px !default; but the right and left padding must be 16px.

#### What are the steps to reproduce?
http://plnkr.co/edit/7acNEwjZUHxUF3UPp3KB

#### What is the use-case or motivation for changing an existing behavior?
To have cards more consistent with the guidelines and to color backgrounds without fixing elements.

#### Which versions of Angular, Material, OS, browsers are affected?
All versions are impacted.

mmalerba:
@kara can you check the card spacing against the spec?

ArtisanDuCode:
I updated my post.
I think it would be best to take inspiration from the repository [card angularjs]( https://github.com/angular/material/tree/master/src/components/card)

mattheiler:
This would make the card a lot more flexible. I currently can't add tabs, a toolbar, a list, etc. to my cards.

Might I add that the footer should not be positioned absolutely. I have to give my cards a fixed height just to use it, which I don't want to do. If the cards are in a collection, then the body should fill the height and scroll, if the height has a max.

mattheiler:
By no means is this an exhaustive, tested list, but... had to add a lot of `!important` styles to fit the spec:

```
.mat-card {
    padding: 0 !important;
}

.mat-card-header {
[...]

PR #4267, opened 2017-04-25
card fix branch

ArtisanDuCode:
I proposed my solution to the Issues #4245.
I think it would be best to take inspiration from the repository [angular/material]( https://github.com/angular/material
But it would require a tag change that created bugs for sites already using material2.


ArtisanDuCode:
I signed it!

kara:
@ArtisanDuCode Would you mind rebasing? There have been updates, so this no longer builds.

kara:
I pulled down your branch and built it, but unfortunately it looks like the changes break a lot of the card's core functionality. Specifically, if you revert the card demo HTML to how it was (e.g. re-add the deleted paragraph tags, restore the subtitle-first example, etc.), those demos no longer look correct. If you want a submit a more narrowly scoped change that does not alter the existing behavior, feel free.

PR #4285, opened 2017-04-26
fix(core): export MdOptionSelectionChange

epelc:
This exports `MdOptionSelectionChange` in `@angular/material`. It makes it easier to import if you are making a custom component which uses `mdOption` and needs to handle selection changes itself(ie a custom search bar).

I couldn't find any guidelines on when or why certain things are exported but not exposed in the main `@angular/material` package.

~~Feel free to close if we should just import from `@angular/material/core/option` instead.~~ It seems importing from `@angular/material/core/option` doesn't work even though there are type files.

### DIFF
--- a/src/lib/core/core.ts
+++ b/src/lib/core/core.ts
@@ -16,7 +16,7 @@ export {Dir, LayoutDirection, RtlModule} from './rtl/dir';
 // Mutation Observer
 export {ObserveContentModule, ObserveContent} from './observe-content/observe-content';
 
-export {MdOptionModule, MdOption} from './option/option';
+export {MdOptionModule, MdOption, MdOptionSelectionChange} from './option/option';
 
 // Portals
 export {


epelc:
@jelbourn Any thoughts on this? I also just found that `ActiveDescendantKeyManager` isn't exported from `core.ts` but it is from it's file.

Are these exports considered internal if they aren't exported again in `core.ts`? If so, would you be ok with exporting `ActiveDescendantKeyManager` and `MdOptionSelectionChange`? They are very useful/required if you need to make a more advanced autocomplete/search bar.

My use case is our search bar which is a mix of auto complete and "Chip" style tag/field suggestions.

epelc:
@crisbeto Can I export `ActiveDescendantKeyManager` too?

crisbeto:
I'll double check. It might be more appropriate as a part of #4290 though.

epelc:
@crisbeto It works if you export it(idk if that's what your checking). I'll put it in #4290 if you'd like.

PR #4291, opened 2017-04-26
docs(getting-started): mention import order for BrowserModule

### LINKED ISSUE #4278: Can't drag md-slider
Looks like a bug, but it could be user error (I'm new to Material)

Should be able to drag on md-slider

Drag doesn't work, though click on slider does set the value

Please check out http://ravinutala.com/d/index.html for an example with just the slider

Just want the correct behavior, not asking for any changes

Angular 4.0.3, Material 2.0.0-beta.3, Windows 10, Chrome/FF (only tested on these two)

jelbourn:
Fixes #4278

### DIFF
--- a/guides/getting-started.md
+++ b/guides/getting-started.md
@@ -72,6 +72,9 @@ import {MdButtonModule, MdCheckboxModule} from '@angular/material';
 export class MyOwnCustomMaterialModule { }
 ```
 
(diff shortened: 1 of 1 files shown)


jelbourn:
@mmalerba I tried, there's not actually anything that _can_ be done

Issue #4093, opened 2017-04-13
Overlay-based components should handle page scroll gracefully


jelbourn:
Creating this issue to track this problem for menu, tooltip, select, autocomplete, datepicker, dialog, snackbar.

Approach we will be taking:
* Default behavior will be to reposition the overlay on scroll using the `ScrollDispatcher`
* Behavior can be changed to disable scrolling instead on a strictly opt-in basis
* Implementation of scroll-blocking must be overridable by end-user

sabithpocker:
Auto complete overlayed dropdown position is not updated on scroll.

sabithpocker:
Autocomplete **does** support repositioning upon resize / scroll. But not when dropdown list items change and there by height of dropdown. 
Also why is the position dependent on _height of dropdown_? it can be accommodated above/below just based on _availability of space_? Or is there a priority to show it below?

Consider a case where one option is already entered("png-alpha" in screenshot). Now when focus is back in the input there is just enough space below the control to show one option(already selected one) and it fits in.

<figure><img width="435" alt="screen shot 2017-05-12 at 3 22 20 pm" src="https://cloud.githubusercontent.com/assets/1050961/25993257/0f3b1900-3727-11e7-9c4b-596e53bb5bfe.png">
<figcaption>Figure 1: single item accommodated in just-anough-space</figcaption>
</figure>

But if we delete couple of characters from the input text, more options showup but doesn't update position to *above* neither does it allow to scroll beyond document height.

<img width="440" alt="screen shot 2017-05-12 at 3 26 28 pm" src="https://cloud.githubusercontent.com/assets/1050961/25993405/834ecb3e-3727-11e7-88b5-bb39986d9814.png">

If I do a scroll or blur/focus-back the repositioning is triggered correctly, but not upon change in "input text" or "dropdown-list" length.

This can be fixed by either of the following in my opinion:
    1. Updating position when input text changes along with scroll, which might eat up performance.
    2. Make position(above/below) irrespective of height of dropdown and only based on proximity of input element to screen edges. This way _figure 1_ should show the options **above** irrespective of the fact that it can be accommodated below.

using master `"@angular/material": "git+https://github.com/angular/material2-builds.git"`

crisbeto:
https://github.com/angular/material2/pull/4469 will reposition the autocomplete panel when the amount of items changes.

sudheeshcm:
@crisbeto  @jelbourn 
[...]

PR #4293, opened 2017-04-26
feat(overlay): add scroll handling strategies

crisbeto:
* Adds the `scrollStrategy` option to the overlay state, allowing the consumer to specify what scroll handling strategy they'd want to use. Also includes a `ScrollStrategy` interface that users can utilize to build their own strategies.
* Adds the `RepositionScrollStrategy`, `CloseScrollStrategy` and `NoopScrollStrategy` as initial, out-of-the-box strategies.
* Sets the `RepositionScrollStrategy` by default on all the connected overlays and removes some repetitive logic from the tooltip, autocomplete, menu and select.

**Note:** I'll add a `BlockScrollStrategy` in a follow-up PR. I wanted to keep this one shorter.

Relates to #4093.

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -13,8 +13,8 @@ import {Subject} from 'rxjs/Subject';
 export class OverlayRef implements PortalHost {
   private _backdropElement: HTMLElement = null;
   private _backdropClick: Subject<any> = new Subject();
-  private _onAttach = new Subject<void>();
-  private _onDetach = new Subject<void>();
+  private _attachments = new Subject<void>();
+  private _detachments = new Subject<void>();
 
   constructor(
       private _portalHost: PortalHost,
@@ -43,7 +43,7 @@ export class OverlayRef implements PortalHost {
     this.updateSize();
     this.updateDirection();
     this.updatePosition();
-    this._onAttach.next();
+    this._attachments.next();
     this._state.scrollStrategy.enable();
 
     // Enable pointer events for the overlay pane element.
@@ -68,7 +68,7 @@ export class OverlayRef implements PortalHost {
     // pointer events therefore. Depends on the position strategy and the applied pane boundaries.
     this._togglePointerEvents(false);
     this._state.scrollStrategy.disable();
-    this._onDetach.next();
+    this._detachments.next();
 
     return this._portalHost.detach();
   }
@@ -84,9 +84,9 @@ export class OverlayRef implements PortalHost {
     this.detachBackdrop();
     this._portalHost.dispose();
     this._state.scrollStrategy.disable();
-    this._onDetach.next();
-    this._onDetach.complete();
-    this._onAttach.complete();
+    this._detachments.next();
+    this._detachments.complete();
+    this._attachments.complete();
   }
 
   /**
@@ -104,13 +104,13 @@ export class OverlayRef implements PortalHost {
   }
 
   /** Returns an observable that emits when the overlay has been attached. */
-  onAttach(): Observable<void> {
-    return this._onAttach.asObservable();
+  attachments(): Observable<void> {
+    return this._attachments.asObservable();
   }
 
   /** Returns an observable that emits when the overlay has been detached. */
-  onDetach(): Observable<void> {
-    return this._onDetach.asObservable();
+  detachments(): Observable<void> {
+    return this._detachments.asObservable();
   }
 
   /**
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -1,6 +1,7 @@
 import {NgZone} from '@angular/core';
 import {PortalHost, Portal} from '../portal/portal';
 import {OverlayState} from './overlay-state';
+import {ScrollStrategy} from './scroll/scroll-strategy';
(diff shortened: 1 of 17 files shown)


src/lib/menu/menu-trigger.ts
```diff
@@ -218,6 +220,7 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
     overlayState.hasBackdrop = true;
     overlayState.backdropClass = 'cdk-overlay-transparent-backdrop';
     overlayState.direction = this.dir;
+    overlayState.scrollStrategy = new RepositionScrollStrategy(this._scrollDispatcher);
```
jelbourn: For a follow-up PR: rather than having menu, select, etc. all directly use `RepositionScrollStrategy`, we should have a provider that can figure the behavior more globally.

src/lib/core/overlay/scroll/close-scroll-strategy.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  attach(overlayRef: OverlayRef) {
+    this._overlayRef = overlayRef;
+  }
+
+  enable() {
+    if (!this._scrollSubscription) {
+      this._scrollSubscription = this._scrollDispatcher.scrolled(null, () => {
+        if (this._overlayRef.hasAttached()) {
+          this._overlayRef.detach();
```
jelbourn: It doesn't look like anything communicates the fact that the overlay detached back to the component, which would make their own state be out of sync.

src/lib/core/overlay/overlay.spec.ts
```diff
@@ -295,6 +296,50 @@ describe('Overlay', () => {
...
+  describe('scroll strategy', () => {
+    let fakeScrollStrategy: ScrollStrategy;
+    let config: OverlayState;
+
+    beforeEach(() => {
+      config = new OverlayState();
+      fakeScrollStrategy = {
+        attach: jasmine.createSpy('attach spy'),
+        enable: jasmine.createSpy('enable spy'),
+        disable: jasmine.createSpy('disable spy')
```
crisbeto: The main reason I avoided adding `isAttached` and `isEnabled` was that I wanted to keep the `ScrollStrategy` interface as lean as possible. Apart from unit tests, I'm not sure that they provide a lot of value.

src/lib/core/overlay/scroll/close-scroll-strategy.ts
crisbeto: The `OverlayRef` calls the `disable` method when it detaches: https://github.com/angular/material2/pull/4293/files#diff-d4ca44eeaacca5fbb90029f947afc1e0R82

src/lib/menu/menu-trigger.ts
crisbeto: Sounds good, although we'd have to have at least two: one for the global overlays and one for the connected ones.

src/lib/core/overlay/overlay.spec.ts
jelbourn: I meant only in the fake implementation for tests

crisbeto: Ah, got it. I've updated the test.

src/lib/core/overlay/scroll/close-scroll-strategy.ts
jelbourn: I meant the other way around- for example, if the overlay for `MdMenu` is closed due to scrolling, nothing tells `MdMenu` that this has happened  and its `_menuOpen` state will be wrong

crisbeto: Got it. We should add a `detached` observable to the `OverlayRef` IMO. We already do it on the `ConnectedOverlayDirective`.

jelbourn: Sounds reasonable to me

crisbeto: Added the observables as discussed. They're not being used anywhere right now since nothing uses the `CloseScrollStrategy`. Also I initially wanted to have the outputs on the `ConnectedOverlayDirective` proxy the ones on the overlay ref, but it's not possible since the ref gets created lazily before attaching.

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -96,6 +103,16 @@ export class OverlayRef implements PortalHost {
...
 
+  /** Returns an observable that emits when the overlay has been attached. */
+  onAttach(): Observable<void> {
+    return this._onAttach.asObservable();
+  }
+
+  /** Returns an observable that emits when the overlay has been detached. */
+  onDetach(): Observable<void> {
+    return this._onDetach.asObservable();
+  }
```
jelbourn: Can we make these getters called `attachments` and `detachments`? I kind of want to move away from the `onAction`-style naming for observables

crisbeto: Done.

mmalerba:
@crisbeto needs rebase

YeomansIII:
How does one use these scroll strategies on a MdDialog? Does that require a separate change?

crisbeto:
The dialog currently has a scroll strategy to disable scrolling. We're still working on a way to be able to override the strategy per component.

PR #4296, opened 2017-04-26
datepicker: create injectable for date formats and bundle it along with date adapter into MdNativeDateModule

mmalerba:


src/lib/datepicker/calendar.ts
```diff
@@ -95,7 +107,8 @@ export class MdCalendar<D> implements AfterContentInit {
   /** The label for the current calendar view. */
   get _periodButtonText(): string {
     return this._monthView ?
-        this._dateAdapter.getMonthYearName(this._activeDate, 'short').toLocaleUpperCase() :
+        this._dateAdapter.format(this._activeDate, this._dateFormats.display.monthYearLabel)
+            .toLocaleUpperCase() :
```
jelbourn: Shouldn't the upper-casing be part of what the formatter does?

mmalerba: I don't think either the format language used by Intl.DatetimeFormat or Moment.js gives control over the case, so we're stuck doing this

fxck:
@mmalerba anything else missing before merging into master?

mmalerba:
@fxck This was the last major feature, all I have left to do is make sure it works well on different browsers and screen readers.

fxck:
Ok, cool. I tried c&p it into my project and fixing the paths meanwhile, looks like there are some problems with it not being rebased as well.

mmalerba:
I just rebased it, give it another try

fxck:
It wasn't the problem in the end, underlineRef currently not existing on input container on current master was. I've changed it to elementRef instead for now.

fxck:
@mmalerba do you have any screenshot of how datetimepicker will look?

PR #4324, opened 2017-04-28
refactor: remove custom error classes

### LINKED ISSUE #4317: Remove all custom Error classes
As of TypeScript 2.1, extending `Error` has issues: it [generates a non-trivial amount of code](https://www.typescriptlang.org/play/#src=class%20MdError%20extends%20Error%20%7B%0D%0A%20%20constructor(value%3A%20string)%20%7B%0D%0A%20%20%20%20super()%3B%0D%0A%20%20%20%20this.message%20%3D%20value%3B%0D%0A%20%20%7D%0D%0A%7D) and it doesn't property pass messages through. We also don't really get any benefit from them, since Angular clobbers them anyway.

We can replace each custom error with a function like `throwInputContainerUnsupportedType()`

crisbeto:
* Removes the custom `MdError` class and refactors all of the classes that extended it into functions.
* Removes a few of the single-function `*-errors.ts` files and inlines their content.

Fixes #4317.


mmalerba:
Can we deprecate instead of removing `MdError`? It looks like there are a number of teams inside Google extending it for their own errors.

jelbourn:
@mmalerba we should eliminate those, they shouldn't be doing that

crisbeto:
Restored and marked `MdError` @mmalerba.

crisbeto:
Beat me by a few seconds there @jelbourn.

jelbourn:
It's okay with me to leave it, then, so long as we also remove them from Google

jelbourn:
@crisbeto can you rebase?

jelbourn:
Also you can totally delete `MdError` now

crisbeto:
Done @jelbourn. A few new `MdError`-s had popped up.

PR #4325, opened 2017-04-28
build: ship individual component bundles

DevVersion:
* Splits the FESM bundles into separate bundles for each component. This should make tree-shaking available for Material.

**Note**: Still in progress & needs some polishing and probably a more elegant approach!


DevVersion:
Closing this as we are going to solve it differently

PR #4327, opened 2017-04-28
fix(a11y): export `ActiveDescendantKeyManager`

epelc:
Follow up from https://github.com/angular/material2/pull/4285#issuecomment-297807895 for exporting `ActiveDescendantKeyManager`.

Still not sure on guidelines but it seems to be approved and this is very useful for custom overlays(easy a11y ftw).


jelbourn:
I don't think that the key managers are quite ready for public consumption yet; we plan on adding them to `@angular/cdk` when they're at that point.

PR #4330, opened 2017-04-29
build: script to build demo-app for firebase

### LINKED ISSUE #3857: Firebase dev deployments don't work with beta.3 packaging changes
It is currently not possible to deploy a working demo on firebase via `firebase deploy` This is due to the fact that the file structure under `dist/` has changed and the `vendor/` files are no longer copied.

DevVersion:
* Adds a new gulp task that can be used to build the devapp in "deploy"-mode. This allows us to deploy the demo-app on firebase.

Fixes #3857

### DIFF
--- a/tools/gulp/tasks/development.ts
+++ b/tools/gulp/tasks/development.ts
@@ -1,13 +1,22 @@
 import {task, watch} from 'gulp';
-import {DIST_ROOT, SOURCE_ROOT} from '../constants';
+import {DIST_ROOT, SOURCE_ROOT, PROJECT_ROOT, DIST_BUNDLES, DIST_MATERIAL} from '../constants';
 import {
   sassBuildTask, tsBuildTask, copyTask, buildAppTask, sequenceTask, triggerLivereload,
   serverTask
 } from '../util/task_helpers';
 import {join} from 'path';
+import {copyFiles} from '../util/copy-files';
 
 const appDir = join(SOURCE_ROOT, 'demo-app');
 const outDir = join(DIST_ROOT, 'packages', 'demo-app');
+
+/** Array of vendors that are required to serve the demo-app. */
+const appVendors = [
+  '@angular', 'systemjs', 'zone.js', 'rxjs', 'hammerjs', 'core-js', 'web-animations-js'
+];
+
+/** Glob that matches all required vendors for the demo-app. */
+const vendorGlob = `+(${appVendors.join('|')})/**/*.+(html|css|js|map)`;
 
 task(':watch:devapp', () => {
   watch(join(appDir, '**/*.ts'), [':build:devapp:ts', triggerLivereload]);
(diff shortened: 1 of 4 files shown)


tools/gulp/tasks/development.ts
```diff
@@ -28,3 +37,10 @@ task(':serve:devapp', serverTask(outDir, true));
 task('serve:devapp', ['build:devapp'], sequenceTask(
   [':serve:devapp', 'material:watch', ':watch:devapp']
 ));
+
+/** Task that copies all vendors into the demo-app package. Allows hosting the app on firebase. */
+task('build-deploy:devapp', ['build:devapp'], () => {
```
mmalerba: Not sure about the name, I feel like it should be `build:[something]`, `build-deploy` makes it sound like it's both building and deploying

DevVersion: How about `prepare-deploy:devapp`?

mmalerba: `build:devapp-for-deploy`?

@jelbourn any thoughts?

DevVersion: Normally you have two different ways to do it:
- `[action]:[scope]` (A few times like in this file)
- `[scope]:[action]`(Most of the time)

It should be either one of these to not mix naming. At some point we will make all of those consistent anyway.

mmalerba: If you want to just pick a name for now we can go back and change it when we consistent-ify everything.

DevVersion: Named it `prepare-deploy:devapp` for now. Let's wait for @jelbourn's feedback before I apply merge ready.

tools/gulp/tasks/development.ts
```diff
@@ -28,3 +37,10 @@ task(':serve:devapp', serverTask(outDir, true));
 task('serve:devapp', ['build:devapp'], sequenceTask(
   [':serve:devapp', 'material:watch', ':watch:devapp']
 ));
+
+/** Task that copies all vendors into the demo-app package. Allows hosting the app on firebase. */
+task('prepare-deploy:devapp', ['build:devapp'], () => {
```
jelbourn: Maybe just make it `deploy:devapp` and have it also run the `firebase deploy`? Anyone running this will almost certainly just run the deploy right after.

DevVersion: Thought about doing that as well, but it looks like it would require me:
* To specify a firebase project I want to deploy to (not very flexible then - [from here](https://www.npmjs.com/package/firebase-tools#using-as-a-module))
* And there would be the ugly `goOffline()` stuff (hope they improve it at some point..)

Do we have a firebase project we always want to use? If so I can use that one (but not flexible I think)

jelbourn: It shouldn't require specifying a firebase project. Typically someone will do `firebase use xxx` first. 

I don't know what you mean by the `goOffline()` stuff

mmalerba: I use different ones since i like to have demos for different PRs going at the same time, maybe make it an argument? this looks promising: https://www.npmjs.com/package/yargs

DevVersion: @jelbourn Not sure, didn't try it with the Firebase-Tools API yet. Will try it later.

With the `goOffline()` stuff I meant that you manually need to "close" the firebase connection, because otherwise the Gulp task will run forever.

@mmalerba That would be an option, but sounds IMO a bit weird to parse arguments in an gulp task.

The advantage of having a task like `prepare-deploy` is, that we might want to make the demo-app "deploy-able", but don't want to publish it immediately. e.g when having the PR previews bot.

Let me see if I need to specify a firebase project.

jelbourn: @DevVersion we can still have the gulp task for just the build part- maybe `stage:devapp`? I just suspect people wouldn't need to use it much

DevVersion: @jelbourn @mmalerba Worked. Done with that.

PR #4331, opened 2017-04-29
chore: added @angular/forms as peer dependency

csvn:
I just created a blank app with `@angular/cli@1.0.1`, and removed `forms` dependency (since I was not planning to use any forms). When serving up the app and adding `@angular/material` I got this error:

image

Seems to me that `@angular/forms` should be a peer dependency if material be imported without errors with no `forms` module.

### DIFF
--- a/src/lib/package.json
+++ b/src/lib/package.json
@@ -24,6 +24,7 @@
   "peerDependencies": {
     "@angular/core": "^4.0.0",
     "@angular/common": "^4.0.0",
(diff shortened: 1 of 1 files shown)


crisbeto:
`@angular/forms` will soon become an optional dependency. See https://github.com/angular/material2/pull/3792.

csvn:
@crisbeto I had no idea, I did not find anything related. Feel free to close this PR then :)

PR #4334, opened 2017-04-30
feat: disabled fieldset support

crisbeto:
Adds support for disabling Material components by disabling their parent fieldset in a similar way to the native fieldsets. Example:

```html
<fieldset disabled>
  <button md-button>Some button</button>
</fieldset>
```

### DIFF
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -7,10 +7,13 @@ import {
   Input,
   OnDestroy,
   Renderer2,
-  ViewEncapsulation
+  ViewEncapsulation,
+  Optional,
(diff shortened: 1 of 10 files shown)


jelbourn:
This feels like something the forms module should do. @kara?

kara:
Yeah, agree that this feels outside the scope of Material.

jelbourn:
@kara should there be an issue tracking this on angular/angular?

mtraynham:
Was trying to figure this out.  It still seems to be an open issue in Angular, https://github.com/angular/angular/issues/13856.

PR #4363, opened 2017-05-02
build: secondary entry points

DevVersion:
### WIP


tools/gulp/util/package-build.ts
```diff
@@ -23,34 +23,61 @@ const sorcery = require('sorcery');
...
+    .forEach(entryName => createSecondaryEntryPoint(packageName, entryName));
+
   updatePackageVersion(releasePath);
   createTypingFile(releasePath, packageName);
   createMetadataFile(releasePath, packageName);
-  addPureAnnotationCommentsToEs5Bundle(releasePath, packageName);
+}
+
+export async function buildPackage(entryFile: string, packagePath: string, packageName: string) {
+  let packageTasks = [buildPackageBundles(entryFile, packageName)];
```
jelbourn: Looks like `buildPackageBundles` doesn't return anything, so you're ending up with an array of `undefined`

src/lib/autocomplete/autocomplete.ts
```diff
@@ -35,7 +34,7 @@ export type AutocompletePositionY = 'above' | 'below';
 export class MdAutocomplete implements AfterContentInit {
 
   /** Manages active item in option list based on key events. */
-  _keyManager: ActiveDescendantKeyManager;
+  _keyManager: _ActiveDescendantKeyManager;
```
DevVersion: We can totally remove them, but the idea (from angular core) was to not "expose" those things to the public.

Though we don't have much stuff in our `private_exports` file, so we would be fine by just exposing those?

tools/gulp/util/rollup-helper.ts
```diff
@@ -67,3 +84,27 @@ export function createRollupBundle(config: BundleConfig): Promise<any> {
...
+ * dependencies. This function also ignores the import shorthands from the NGC (tsickle)
+ **/
+function isExternalImport(moduleId: string): boolean {
+  return ROLLUP_EXTERNALS.indexOf(rewriteConvertedImports(moduleId)) !== -1;
+}
+
+/**
+ * When compiling with @angular/tsc-wrapped, shorthand imports will be converted into more
+ * explicit paths. This function rewrites all expanded imports to their original import.
+ * This is important, because when composing releases, the directories won't be present anymore.
```
DevVersion: It's not an issue on `angular/angular`. It's a feature on tsickle. I was talking to @alexeagle about adding a flag or something so we can be more flexible about it.

tools/gulp/util/package-build.ts
DevVersion: It does return a promise. Every function that uses `async/await` automatically returns them

tools/gulp/util/package-build.ts
```diff
@@ -23,34 +23,61 @@ const sorcery = require('sorcery');
...
   // To avoid refactoring of the project the package material will map to the source path `lib/`.
-  let sourcePath = join(SOURCE_ROOT, packageName === 'material' ? 'lib' : packageName);
-  let packagePath = join(DIST_ROOT, 'packages', packageName);
-  let releasePath = join(DIST_ROOT, 'releases', packageName);
+  const sourcePath = join(SOURCE_ROOT, packageName === 'material' ? 'lib' : packageName);
+  const packagePath = join(DIST_ROOT, 'packages', packageName);
+  const releasePath = join(DIST_ROOT, 'releases', packageName);
+
+  const umdOutput = join(releasePath, 'bundles');
+  const fesmOutput = join(releasePath, '@angular');
```
DevVersion: Yeah I think it should be like that?

```
| @angular/material
| -- @angular/
| -- -- material.js
| -- -- material/
| -- -- -- core.js
```

DevVersion:
We are holding off with that for now.

PR #4365, opened 2017-05-02
Autocomplete home keys bug fix

### LINKED ISSUE #3496: autocomplete - HOME and END keys not working

shaunak1111:
fix(autocomplete): HOME and END keys default behaviour

Implements the default behaviours for HOME and END keys and their combinations with other keys on input fields.
The same behaviour is replicated in angular1 material autocomplete.

Closes #3496


jelbourn:
This is being worked on in #4544, the final approach will likely be a change in list-key-manager.

Issue #4318, opened 2017-04-27
MdSnackbar enters the DOM already in view, preventing the correct animation.


josephperrott:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Snackbar begins off screen and then animates into view

#### What is the current behavior?
Snackbar begins partially on screen and then animates fully into view

#### What are the steps to reproduce?
Can be seen in demo app.

#### Which versions of Angular, Material, OS, browsers are affected?
Material2 since [this pull](https://github.com/angular/material2/pull/3626)

#### Is there anything else we should know?
I'm working on fixing it though don't have a solution yet.

PR #4370, opened 2017-05-03
fix(portal): detect changes for portal hostview while before attaching.

josephperrott:
#4318

### DIFF
--- a/src/lib/core/portal/dom-portal-host.ts
+++ b/src/lib/core/portal/dom-portal-host.ts
@@ -29,21 +29,17 @@ export class DomPortalHost extends BasePortalHost {
    */
   attachComponentPortal<T>(portal: ComponentPortal<T>): ComponentRef<T> {
     let componentFactory = this._componentFactoryResolver.resolveComponentFactory(portal.component);
-    let componentRef: ComponentRef<T>;
+    let componentRef = componentFactory.create(portal.injector || this._defaultInjector);
+    componentRef.hostView.detectChanges();
 
     // If the portal specifies a ViewContainerRef, we will use that as the attachment point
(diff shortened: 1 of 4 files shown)


josephperrott:
@jelbourn ready for you to take a look

src/lib/core/portal/dom-portal-host.ts
```diff
@@ -29,21 +29,17 @@ export class DomPortalHost extends BasePortalHost {
    */
   attachComponentPortal<T>(portal: ComponentPortal<T>): ComponentRef<T> {
     let componentFactory = this._componentFactoryResolver.resolveComponentFactory(portal.component);
-    let componentRef: ComponentRef<T>;
+    let componentRef = componentFactory.create(portal.injector || this._defaultInjector);
+    componentRef.hostView.detectChanges();
```
jelbourn: Are you able to create a unit test that enforces this behavior?

josephperrott:
Should be ready for another look.

I updated `MdDatepickerContent` to use the safe navigation operator.  This is necessary as the newly created `ComponentPortal<MdDatepicker>` instance is immediately change detected.  We need to guard against `undefined` and `null` since the inputs aren't guaranteed to be populated.

cc:@mmalerba as this is modifying the newly created `MdDatepicker`

mmalerba:
LGTM for datepicker

willshowell:
@jelbourn @mmalerba @josephperrott the CD changes made here are causing several regressions with the datepicker in popup view

1. Unable to start in year view
2. Start at date has no effect
3. "Prev month" button initially enabled when min date in current month
4. When setting min/max dates within current month, all days are enabled until month view is refreshed
5. Active date is not highlighted until month view is refreshed

ref #5065

PR #4375, opened 2017-05-03
feat: integrate typography api into all components

crisbeto:
* Includes the base typography styles for headers, body text etc.
* Adds mixins for typography next to the theme and moves out all of the component typography styles into them.
* Switches all the hardcoded typography values to use the ones from the spec.
* Adjusts the letter spacing on some of the headers to align them to the spec.

**Note:** While going through the components, I haven't checked whether the styles are accurate in regards to the spec. I've left notes in the places where the font sizes weren't a part of the pre-defined ones.


src/demo-app/demo-app/demo-app.html
```diff
@@ -25,12 +25,14 @@
...
         <h1>Angular Material Demos</h1>
-        <button md-button (click)="toggleFullscreen()" title="Toggle fullscreen">
-          Fullscreen
-        </button>
-        <button md-button (click)="root.dir = (root.dir == 'rtl' ? 'ltr' : 'rtl')" title="Toggle between RTL and LTR">
-          {{root.dir.toUpperCase()}}
-        </button>
+        <div>
+          <button md-button (click)="toggleFullscreen()" title="Toggle fullscreen">
+            Fullscreen
```
jelbourn: Since you're incidentally touching this, can you right-align this and make it an md-icon-button with the `fullscreen` icon 😄 ? That's been bothering me forever.

src/demo-app/typography/typography-demo.html
```diff
@@ -0,0 +1,15 @@
+<h1 class="mat-h0">Lorem ipsum dolor sit amet.</h1>
+<h1 class="mat-h1">Lorem ipsum dolor sit amet, consectetur adipisicing.</h1>
+<h2 class="mat-h2">Lorem ipsum dolor sit amet, consectetur adipisicing.</h2>
+<h3 class="mat-h3">Lorem ipsum dolor sit amet, consectetur adipisicing.</h3>
+<h4 class="mat-h4">Lorem ipsum dolor sit amet, consectetur adipisicing.</h4>
+<h5 class="mat-h5">Lorem ipsum dolor sit amet, consectetur adipisicing.</h5>
+<h6 class="mat-h6">Lorem ipsum dolor sit amet, consectetur adipisicing.</h6>
+
+<div class="mat-body">
+  <p>Lorem ipsum dolor sit amet, <span class="mat-body-strong">consectetur</span> adipisicing elit. Iure voluptatem amet facilis rerum non dolore repellendus ab. Assumenda nisi perspiciatis odit illum voluptatem expedita laborum! Debitis nisi eius, ratione nostrum velit dignissimos molestias saepe, esse facilis blanditiis, labore optio. Accusantium, nihil illo est beatae nobis expedita animi libero, laboriosam excepturi consequatur eaque, ab repudiandae.</p>
```
jelbourn: optional: I like adding something more meaningful / whimsical for demos, up to you

src/lib/chips/_chips-theme.scss
```diff
@@ -1,5 +1,11 @@
 @import '../core/theming/palette';
 @import '../core/theming/theming';
+@import '../core/typography/typography-utils';
+
+// TODO(crisbeto): these values don't correspond to any of the typography breakpoints.
```
jelbourn: We might have to invent our own; the spec often paints broad strokes that don't capture the specific details for each component

src/lib/core/typography/_typography.scss
```diff
@@ -42,17 +42,21 @@
 }
 
 // Adds the base typography styles, based on a config.
-@mixin mat-typography($config: mat-typography-config(), $selector: '.mat-typography') {
+// TODO(crisbeto): the headers need to handle margins as well.
+@mixin mat-base-typography($config, $selector: '.mat-typography') {
   .mat-h0, .mat-hero-header {
     @include mat-typography-level-to-styles($config, display-4);
+    letter-spacing: -0.05em;
```
jelbourn: Does the letter spacing come from the spec somewhere?

jelbourn:
@tinayuangao @andrewseguin @kara @mmalerba can you each do a quick pass on the components you own here?

src/lib/core/typography/_typography.scss
crisbeto: Not really. I arrived at it by overlaying the spec and tweaking ours until they lined up.

mmalerba: I vote for undoing this and just using default spacing, some things look kind of cramped when using these settings, for example the page header.

Page header before this PR:
old-title

Page header after this PR:
new-title

mmalerba:
LGTM for the input, slider, and sidenav changes

tinayuangao:
LGTM for checkbox, button, button-toggle and radio-button

src/lib/core/typography/_typography.scss
crisbeto: This one isn't using the letter spacing at all @mmalerba, the spacing only applies to `h1`, `h2` and hero header.

crisbeto:

src/lib/core/typography/_typography.scss
mmalerba: Ah I see, it seems it switched from Arial to Roboto which explains the difference, maybe we should add `.mat-h4/5` to the element so it looks a little better?

crisbeto: It has @include mat-typography-level-to-styles($config, title); which includes the same styles as .mat-h5.

mmalerba: It's not quite the same, it looks like maybe this element is not getting its boldness set correctly? If I add `mat-h5` it gets less bold

crisbeto: You're right, it's because we have a global style for `body h1` in the demo app, which ends up overriding the toolbar style. I'll push a fix.

jelbourn: Add a comment somewhere explaining where the values came from, then? I know a year from now I'll be looking at this and be thinking "Where did these come from?!"

jelbourn:
@kara @andrewseguin please take a look

No idea what's going on with the screenshot tests

src/lib/core/typography/_typography.scss
crisbeto: [I already left a link to the spec right above the pre-defined values.](https://github.com/angular/material2/pull/4375/files#diff-6eb6698b7e87b7f029b5578b595697d3L13)

jelbourn: @crisbeto the spec doesn't explicitly give the letter spacing, though; we should document where those specific values came from.

crisbeto: Ah I see. As I mentioned, I arrived at them mainly by overlaying the spec over our demo page and tweaking it until they aligned. I'll add notes about it.

kara:
@crisbeto Rebase?

crisbeto:
Rebased @kara.

andrewseguin:
LGTM for tooltip/tabs

kara:
@crisbeto I think the rebase may have caused issues. I'm seeing this:

screen shot 2017-05-16 at 11 22 38 am

crisbeto:
Fixed @kara. I had forgotten to prefix the theme file name with an underscore and it ended up throwing a SASS error.

jelbourn:
@crisbeto removing the `$mat-font-family` breaks a bunch of Google apps; can we add it back for know so we can separately refactor it away?

crisbeto:
Done @jelbourn.

jelbourn:
Looks like that fixed the errors, but I have to update some internal build rules and screenshot tests to get this submitted, so I'll do a separate sync just for this.

mmalerba:
@crisbeto Inspecting the demo app, I see that the `mat-button-wrapper` for the hamburger icon now has a computed `font-size` of `13.3333px` rather than the `14px` it previously had. Do you know what's responsible for this change? Is it expected? (It seems to be the cause of a lot of the screenshot diffs I'm seeing in Google)

PR #4380, opened 2017-05-03
chore: random test leaks

crisbeto:
Fixes some unit tests that were leaking elements into other tests.


crisbeto:
AFAIK we could technically use a custom Jasmine reporter to run some logic after each test and check the DOM for leftovers (even though reporters aren't really meant to do this). I remember playing around with it for M1.

PR #4392, opened 2017-05-04
build: remove whitespace at eof and between tags

### LINKED ISSUE #1596: Remove trailing whitespace in inlined templates

Bug

The string representing a component's template should not have a whitespace as the last character.

Every material component's template appends a whitespace at the end of the string in the JS transpiled file.

`<button md-button class="increment" (click)="increment()">A+</button>`

will be translated into :

`<button class="increment" md-button=""><span class="md-button-wrapper">A+</span> </button>`

... see the whitespace just before `</button>` ?
It will be appended before the closing tag of any material element whose template file has a new line before the EndOfFile...

the following test should work :

```
it('should display a button that reads "A+" ', () => {
        const buttonContent = button.textContent;
        expect(buttonContent).toEqual('A+');
    });
```

Any version that follows the official style guide's linting recommendations.

You're awesome !

willshowell:
Fixes #1596

### DIFF
--- a/scripts/release/inline-resources.js
+++ b/scripts/release/inline-resources.js
@@ -93,7 +93,9 @@ function inlineTemplate(content, urlResolver) {
     const templateContent = fs.readFileSync(templateFile, 'utf-8');
     const shortenedTemplate = templateContent
       .replace(/([\n\r]\s*)+/gm, ' ')
-      .replace(/"/g, '\\"');
+      .replace(/>\s</gm, '><')
+      .replace(/"/g, '\\"')
+      .trim();
     return `template: "${shortenedTemplate}"`;
   });
 }


scripts/release/inline-resources.js
```diff
@@ -93,7 +93,9 @@ function inlineTemplate(content, urlResolver) {
     const templateContent = fs.readFileSync(templateFile, 'utf-8');
     const shortenedTemplate = templateContent
       .replace(/([\n\r]\s*)+/gm, ' ')
```
DevVersion: This seems to be a overkill fix. Why not just remove the space here and just replace it with an empty string?

willshowell: Wouldn't that remove the spaces between attributes too aggressively? Such as after `<ng-content`:
```html
<ng-content select="[md-card-avatar], [mat-card-avatar]"></ng-content>
<div class="mat-card-header-text">
  <ng-content
      select="md-card-title, mat-card-title, md-card-subtitle, mat-card-subtitle,
      [md-card-title], [mat-card-title], [md-card-subtitle], [mat-card-subtitle]"></ng-content>
</div>
<ng-content></ng-content>
```

DevVersion: That's a good point. Couldn't we just use gulp-clean-css to do proper minifying?

willshowell: Yeah, sure. I considered reusing htmlmin, but that seemed like overkill to me haha. I'll close this if you'd prefer to take on a different approach.

DevVersion: You can just do it in that PR and you don't have to close the PR.

jelbourn:
Not sure what the status of this PR is now- I'm just going to close it and you can open a new one if you want to go for a different approach.

PR #4401, opened 2017-05-05
feat(autocomplete): allow the ability to select an option on pressing…

SimplerSoftware:
… the 'TAB' key

Currently you can only select an option by pressing the 'ENTER' key when using the keyboard.  Many users expect to be able to use the 'TAB' key to select an option as well.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -19,7 +19,7 @@ import {PositionStrategy} from '../core/overlay/position/position-strategy';
 import {ConnectedPositionStrategy} from '../core/overlay/position/connected-position-strategy';
 import {Observable} from 'rxjs/Observable';
 import {MdOptionSelectionChange, MdOption} from '../core/option/option';
-import {ENTER, UP_ARROW, DOWN_ARROW} from '../core/keyboard/keycodes';
+import {ENTER, UP_ARROW, DOWN_ARROW, TAB} from '../core/keyboard/keycodes';
 import {Dir} from '../core/rtl/dir';
 import {MdInputContainer} from '../input/input-container';
 import {ScrollDispatcher} from '../core/overlay/scroll/scroll-dispatcher';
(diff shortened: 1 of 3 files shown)


badre429:
Is it possible to always select the first option on the autocomplete list before press enter

SimplerSoftware:
@badre429: this feature does not address that request.

SimplerSoftware:
@jelbourn:  Hello!  I'm not sure if there's anything else I need to do in order for this to be considered or reviewed. This is my first time contributing to this project so please correct me if I have done anything incorrect.  BTW - This project is great and look forward to seeing the progress in the near future!

jelbourn:
Hi @SimplerSoftware- we appreciate your PR, but selecting items on `TAB` wouldn't play well with the component's accessibility for screen readers; space and enter should be the only keys that trigger a selection.

SimplerSoftware:
@jelbourn:  I'll be honest, I haven't really dealt with screen readers and accessibility up to this point.  That being said, I've been doing some homework and I'm not seeing how this feature wouldn't play well.  In fact, to me it seems like it would follow the principle of "selection follow focus" (see [here](https://www.w3.org/TR/wai-aria-practices/#kbd_selection_follows_focus)) which in this case would be a benefit to the user.  From my perspective, when the autocomplete is being used to select a single item at a time, using the arrow keys to navigate the options and move focus between the different options *should*, if it's not already doing so, follow the "selection follow focus" principle and by pressing the <kbd>TAB</kbd> key you are only indicating that you are ready to move onto the next element after having already selected your option using the arrow keys.

As an example, you can look at Google's search bar.  As you're typing in the search box, several suggestions appear.  As you use the arrow keys to move up or down, the currently focused phrase is the 'selected' phrase and the search results update accordingly without ever having pressed the <kbd>ENTER</kbd> or <kbd>SPACE</kbd> keys.

Thoughts?

jelbourn:
The problem is that <kbd>TAB</kbd> has a very specific behavior associated with it (moving the focus). Overloading that behavior is potentially confusing for screen-reader users, since it's the way they navigate around a form. 

Looking at the google.com suggestions autocomplete, pressing <kbd>TAB</kbd> does not really _commit_ the selected query (it doesn't even close the popup). It does, however, update the search result based on currently focused option, so having that surfaced seems like a reasonable thing to do.

SimplerSoftware:
Based on that, I think the approach that I took does not work, but the end result that I think we both can agree on is this:

When the keyboard is used to navigate the options in the autocomplete list, the 'Focused' option should also be the 'Selected' option following the principle of "selection follow focus".  Thus when a user has selected an option using the arrow keys and then hits the <kbd>TAB</kbd> key, the focus moves to the next input but the end result is that the autocomplete now has a value.  Would you agree?

jelbourn:
Not sure if it should _always_ immediately update the value, might be something that could be a setting. It would be good to open an issue for discussion

PR #4414, opened 2017-05-07
feat(list): allow ripples for normal lists

DevVersion:
* Allow enabling ripples for normal `md-list` elements.

### DIFF
--- a/src/lib/list/list.spec.ts
+++ b/src/lib/list/list.spec.ts
@@ -1,7 +1,7 @@
 import {async, TestBed} from '@angular/core/testing';
 import {Component, QueryList, ViewChildren} from '@angular/core';
 import {By} from '@angular/platform-browser';
-import {MdListItem, MdListModule} from './index';
+import {MdList, MdListItem, MdListModule} from './index';
 
 
 describe('MdList', () => {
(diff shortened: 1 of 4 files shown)


theiliad:
This is great! Can't wait to use it 😀

src/lib/list/list.spec.ts
```diff
@@ -156,6 +156,49 @@ describe('MdList', () => {
...
+
+    fixture.detectChanges();
+
+    const items = fixture.componentInstance.listItems;
+
+    expect(items.length).toBeGreaterThan(0);
+    expect(list.disableRipple).toBe(true, 'Ripples should be disabled for normal lists.');
+
+    // Ripples should be disabled by default, and can be enabled with a binding.
+    items.forEach(item => expect(item.isRippleEnabled()).toBe(false));
```
crisbeto: Here and on line 177, instead of doing multiple `it` calls inside a loop, you could use `Array.every`. The difference is that the log will be much shorter if it fails:

```ts
expect(items.every(item => item.isRippleEnabled())).toBe(false);
```

src/lib/list/list.spec.ts
```diff
@@ -156,6 +156,49 @@ describe('MdList', () => {
...
```
crisbeto: Same note about the `it` calls inside a loop.

src/lib/list/list.ts
```diff
@@ -151,18 +158,15 @@ export class MdListItem implements AfterContentInit {
...
+              @Optional() private _list: MdList) {}
 
   ngAfterContentInit() {
     this._lineSetter = new MdLineSetter(this._lines, this._renderer, this._element);
   }
 
   /** Whether this list item should show a ripple effect when clicked.  */
   isRippleEnabled() {
-    return !this.disableRipple && this._isNavList && !this._list.disableRipple;
+    return !this.disableRipple && !this._list.disableRipple;
```
kara: Getting test failures because `this._list` is sometimes null.

kara:
+ @jelbourn might want to opine on this

src/lib/list/list.ts
DevVersion: Makes sense.

DevVersion:
@kara Fixed that the `this._list` is undefined. Why can there be list-items without a list actually? They will not have any styles from the SCSS.

Context: Now ripples show for list-items without a list by default. Because `disableRipple` is set to false by default?

jelbourn:
I don't believe that `md-list` by itself should have ripples, which should only be applied to interactive elements (e.g., md-nav-list, md-action-list, md-selection-list). I think that allowing ripples in general would lead people to just putting click handlers on normal lists which is bad for a11y.

theiliad:
@jelbourn Any docs/demos available for md-action-list, md-nav-list or md-selection-list?

jelbourn:
@thecodingwizard nav-list exists now, not the other two yet.

https://material.angular.io/components/component/list

theiliad:
@jelbourn Oh alright, hadn't really realized the code for nav-list there, thanks.

But until action-list is available, don't you think we need a way of using ripples on lists?

DevVersion:
@jelbourn The ripples won't be enabled by default. The users would just have the flexibility to "explicitly" enable the ripples for normal lists.

Although I agree that it might not be the best for a11y to have a `(click)` listener applied on list-items, but in some situations you might need to do it and the `md-nav-list` doesn't have any special a11y treatment though.

kara:
FWIW, I think we should just wait until action-list is available and apply ripples to that. Otherwise, even after action-list comes out, people will still be putting ripples and click handlers on list items, which will degrade a11y long term.

PR #4418, opened 2017-05-07
docs(sidenav): escape pipes in markdown table

GeorgDangl:
replace pipe "|" in markdown tables for correct rendering in a table cell in GitHub

The table in the sidenav readme looks currently like this:
table_error

With the fix, it looks like this:
table_fixed

This [seems to be an old but unresolved issue](https://github.com/gitlabhq/gitlabhq/issues/1238) with how GitHub renders MarkDown files / tables.

### DIFF
--- a/src/lib/sidenav/README.md
+++ b/src/lib/sidenav/README.md
@@ -23,8 +23,8 @@ The sidenav panel.
 
 | Name | Type | Description |
 | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


kara:
Note: We should probably move the info in the table into `sidenav.md` as well, since the README isn't pushed to the docs site.

Issue #3967, opened 2017-04-08
Skip tests on CI when only markdown files have changed


jelbourn:
We should skip most tests if a commit only affects markdown files.

Wo1v3r:
Would really like to give it a try, could you expand a bit, please?

DevVersion:
Done.

PR #4452, opened 2017-05-10
build(ci): skip tests on CI when only markdown files have changed

chouclee:
Skip both unit tests and e2e tests on CI when commits contain only markdown files (``*.md``)
solves #3967

### DIFF
--- a/scripts/ci/build-and-test.sh
+++ b/scripts/ci/build-and-test.sh
@@ -10,6 +10,19 @@ cd $(dirname $0)/../..
 source scripts/ci/sources/mode.sh
 source scripts/ci/sources/tunnel.sh
 
+# Get commit diff
+if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
+  fileDiff=$(git diff --name-only $TRAVIS_COMMIT_RANGE)
+else
+  fileDiff=$(git diff --name-only $TRAVIS_BRANCH...HEAD)
+fi
+
+# Check if tests can be skipped
+if [[ ${fileDiff} =~ ^(.*\.md\s*)*$ ]] && (is_e2e || is_unit); then
+  echo "Skipping e2e and unit tests since only markdown files changed"
+  exit 0
+fi
+
 start_tunnel
 wait_for_tunnel
 
@@ -23,7 +36,7 @@ elif is_payload; then
   $(npm bin)/gulp ci:payload
 elif is_closure_compiler; then
   ./scripts/closure-compiler/build-devapp-bundle.sh
-else
+elif is_unit; then
   $(npm bin)/gulp ci:test
 fi
 
--- a/scripts/ci/sources/mode.sh
+++ b/scripts/ci/sources/mode.sh
@@ -20,3 +20,7 @@ is_closure_compiler() {
 is_payload() {
   [[ "$MODE" = payload ]]
 }
+
+is_unit() {
+  [[ "$MODE" = saucelabs_required || "$MODE" = browserstack_required ]]
+}


scripts/ci/build-and-test.sh
```diff
@@ -10,20 +10,29 @@ cd $(dirname $0)/../..
 source scripts/ci/sources/mode.sh
 source scripts/ci/sources/tunnel.sh
 
+# Check if tests can be skipped
+skip_tests=true
+for filename in $(git diff --name-only $TRAVIS_BRANCH...HEAD); do
+  if ! [[ "$filename" =~ .*\.md ]]; then
+    skip_tests=false
+    break
```
DevVersion: Should we check the modes here and just exit before?

scripts/ci/build-and-test.sh
```diff
@@ -10,20 +10,29 @@ cd $(dirname $0)/../..
 source scripts/ci/sources/mode.sh
 source scripts/ci/sources/tunnel.sh
 
+# Check if tests can be skipped
+skip_tests=true
+for filename in $(git diff --name-only $TRAVIS_BRANCH...HEAD); do
```
DevVersion: Would it make more sense to use `TRAVIS_COMMIT_RANGE`?

chouclee: https://github.com/travis-ci/travis-ci/issues/2668 ``TRAVIS_COMMIT_RANGE`` will include outdated commit if user does a force push (which is very common to updating a PR)

scripts/ci/build-and-test.sh
chouclee: Good point. Let me try to change it.

chouclee:
Another thing came to my mind, maybe we can just skip CI build if only docs are being changed.
What do you think about? @jelbourn

jelbourn:
@chouclee isn't that what this already is?

DevVersion:
@chouclee You are talking about the "per-commit" builds right?

 I don't think we should skip those. The `material2-docs-content` still needs to be deployed and I think it's still good to publish the `material2-builds` for every commit that lands in `master`. (As same as in `@angular/angular`)

scripts/ci/build-and-test.sh
DevVersion: But If I recall correctly, for Push builds the `$TRAVIS_BRANCH` will be `master` and the merged commit will be `HEAD` / `master` as well.

This would mean that the `diff` will be always empty for push builds.

chouclee: @DevVersion ``HEAD`` would be PR merged into ``master`` when ``$TRAVIS_BRANCH`` is set to ``master``. Then the diff would be your PR commits. You can try it.

I already tested this PR. And just did another one: 
https://travis-ci.org/chouclee/material2/builds/234685761 (``md`` file change only)
https://travis-ci.org/chouclee/material2/builds/234687221 (``md`` and ``js`` file, this is a hard push to update PR).

You can find the tests were skipped in first CI build, but not in second one.

DevVersion: I didn't talk about PRs. Good to see that you tested it for PRs.

Can you try merging your PR on your fork and see if everything works there too?

chouclee: I see what you meant. You are right, for Push build, the ``diff`` would be empty.
I should check the ``$TRAVIS_PULL_REQUEST`` first. 

And I did a little search: 
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/72fa14b3957cacfdfe3113d14b8d77204f7ec9f7/nox.py#L84

chouclee:
@DevVersion PR updated.

scripts/ci/build-and-test.sh
```diff
@@ -10,6 +10,26 @@ cd $(dirname $0)/../..
...
+fi
+
+# Check if tests can be skipped
+skip_tests=true
+for filename in $diff; do
+  if ! [[ "$filename" =~ .*\.md ]]; then
+    skip_tests=false
+    break
+  fi
+done
```
DevVersion: Thinking if we could make it more simple

```bash
if [[ "$TRAVIS_PULL_REQUEST" = "false" ]]; then
  fileDiff=$(git diff --name-only $TRAVIS_COMMIT_RANGE)
else
  fileDiff=$(git diff --name-only $TRAVIS_BRANCH...HEAD)
 fi

if [[ ${fileDiff} =~ ^(.*\.md\s*)*$ ]] && (is_e2e || is_unit); then
  echo "Skipping e2e and unit tests since only markdown files changed"
  exit 0
fi
```

chouclee: Cool, I just learned something new from you. 

Only two things:
1. ``[[ "$TRAVIS_PULL_REQUEST" = "false" ]]`` would print as ``+[[ false = \f\a\l\s\e ]]`` in Travis output, and quotes are unnecessary.  
[ "$TRAVIS_PULL_REQUEST" = "false" ] would print as ``+[[ false = false ]]. I think it's more readable. 

2.  I changed the regex ``${fileDiff} =~ ^(.*\.md\s*)*$`` to ``${fileDiff} =~ ^(.*\.md\s*)+$``. Although I doubt anyone would push an empty commit...

DevVersion: No problem

1) I didn't write that. Just copied it from your PR changes
2) It's fine. But I prefer the `*` just to be on the saf eside.

chouclee: Do you mind sharing that why you think ``*`` is safer here?

DevVersion: Just in case there would be any empty commit. Which happens very unlikely but it just gives us the flexibility. No real benefit of using `+`

chouclee: I am confused... ``+``  here is used to tackle the empty commit case (e.g. trigger CI with an empty commit)... 
If it's an empty commit, ``*`` would skip the tests, but ``+`` would't. The feature here is to skip tests only ``md`` files are changed. That's why I use ``+``.

DevVersion: If the commit is empty then it's fine to skip the tests as well. No source code change is 100% guaranteed.

chouclee: Sure. I restored ``+`` to ``*``.

DevVersion:
@chouclee By the way. Thanks for your work! It works perfectly 🎉

chouclee:
@DevVersion I am glad to hear that. Thanks for your review. :smile:

PR #4468, opened 2017-05-10
feat(data-table): minimal beginning implementation

andrewseguin:


src/demo-app/data-table/data-table-demo.scss
```diff
@@ -0,0 +1,86 @@
+/* Structure */
+.table-container {
```
jelbourn: Prefix demo styles with `demo-`

src/lib/core/data-table/data-source.ts
```diff
@@ -0,0 +1,6 @@
+import {Observable} from 'rxjs/Observable';
+import {CdkTableViewData} from './data-table';
+
+export abstract class DataSource<T> {
+  abstract connectTable(viewChanged: Observable<CdkTableViewData>): Observable<T[]>;
```
jelbourn: This should be something like 
```ts
abstract connect(collectionViewer: CollectionViewer): Observable<T[]>;
```
Where `CollectionViewer` is an interface that both the data table and the tree would implement that exposes the necessary stream (and potentially other information in the future)

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,304 @@
...
+  ViewContainerRef,
+  ViewEncapsulation
+} from '@angular/core';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+import 'rxjs/add/operator/let';
+import 'rxjs/add/operator/debounceTime';
+import 'rxjs/add/observable/combineLatest';
+import {DataSource} from './data-source';
+
+/** Template to be used for the data cells of a particular column. */
```
jelbourn: How about 
```ts
/** Cell definition for a CDK data-table. Captures the template as well as other cell-specific properties. */
```
(similar for other `cdkXxxDef` directives)

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,304 @@
...
+export class CdkHeaderRow { }
+
+/** Data row template container that contains the cell outlet. Adds the right class and role. */
+@Component({
+  selector: 'cdk-row',
+  template: '<ng-container cdkCellOutlet></ng-container>',
+  host: {
+    'class': 'mat-row',
+    'role': 'row',
+  },
```
jelbourn: Start off with this being `OnPush`?

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  constructor(public template: TemplateRef<any>) { }
+}
+
+/** Header template container that contains the cell outlet. Adds the right class and role. */
+@Component({
+  selector: 'cdk-header',
+  template: '<ng-container cdkCellOutlet></ng-container>',
+  host: {
+    'class': 'mat-header',
```
jelbourn: CDK directives should use the `cdk-` prefix
(here and elsewhere)

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,304 @@
...
+import {DataSource} from './data-source';
+
+/** Template to be used for the data cells of a particular column. */
+@Directive({selector: '[cdkRowCellDef]'})
+export class CdkRowCellDef {
+  constructor(public template: TemplateRef<any>) { }
+}
+
+/** Template to be used for the header cell of a particular column. */
+@Directive({selector: '[cdkHeaderCellDef], cdk-column-def'})
```
jelbourn: Why is there an element selector here? AFAIK it should always be an attribute

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  // TODO: Add an input for providing a switch function to determine
+  // if this template should be used.
+
+  constructor(public template: TemplateRef<any>) { }
+}
+
+/** Header template container that contains the cell outlet. Adds the right class and role. */
+@Component({
+  selector: 'cdk-header',
```
jelbourn: `header-row`?

src/lib/core/data-table/data-table.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    expect(getRows().length).toBe(dataSource.data.length);
+
+    // Check that the number of cells is correct
+    getRows().forEach(row => {
+      expect(getRowCells(row).length).toBe(component.columnsToRender.length);
+    });
+  });
+});
+
+export interface TestData {
```
jelbourn: You shouldn't `export` anything from a unit test

src/lib/core/data-table/data-table.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+  a: string;
+  b: string;
+  c: string;
+}
+
+export class SimpleDataSource extends DataSource<TestData> {
```
jelbourn: `FakeDataSource` for unit test

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,304 @@
...
+export class CdkHeaderRowPlaceholder {
+  constructor(public viewContainer: ViewContainerRef) { }
+}
+
+export interface CdkTableViewData {
+  start: number;
+  end: number;
+}
+
+@Component({
```
jelbourn: Need class description

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,304 @@
...
+  host: {
+    'class': 'mat-header-cell',
+    'role': 'columnheader',
+  },
+})
+export class CdkHeaderCell {
+  constructor(private columnDef: CdkColumnDef,
+              private elementRef: ElementRef,
+              private renderer: Renderer) {
+    this.renderer.setElementClass(elementRef.nativeElement, `mat-column-${columnDef.name}`, true);
```
jelbourn: Need unit test for this

src/lib/core/data-table/data-table.ts
andrewseguin: Good catch, not sure how that ended up there

src/demo-app/data-table/data-table-demo.scss
```diff
@@ -0,0 +1,86 @@
...
+  max-height: 800px;
+
+  // Table fills in the remaining area with a scroll
+  .cdk-table {
+    flex: 1 1 auto;
+    overflow: auto;
+  }
+}
+
+/* Material Styles */
```
kara: Context on why Material styles are in the demo CSS?  I assume the component doesn't come with styles, so this is just to make it look Material-y? The comment makes it sound like this comes with the component.

andrewseguin: Changed the comment to the following. Let me know if its not totally clear

```

/*
 * Styles so that the demo's cdk-table matches the material design spec
 * https://material.io/guidelines/components/data-tables.html
 */
```

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+})
+export class CdkTable {
+  /**
+   * Provides a stream containing the latest data array to render. Influenced by the table's
+   * stream of view window (what rows are currently on screen).
+   */
+  @Input() dataSource: DataSource<any>;
+
+  /**
+   * Stream containing the latest information on what rows are being displayed on screen. Offered
```
jelbourn: "Offered to the data source" -> "Can be used by the data source to..."

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+   * to the data source as a heuristic of what data should be provided.
+   */
+  // TODO: Remove max value as the end index and instead calculate the view on init and scroll.
+  viewChange = new BehaviorSubject<CollectionViewer>({start: 0, end: Number.MAX_VALUE});
```
jelbourn: `viewChange` -> `viewChanges`? 

(not super into the name, but I don't have an alternative)

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  /**
+   * Map of all the user's defined columns identified by name.
+   * Contains the header and data-cell templates.
+   */
+  private _columnMap = new Map<ConstrainDOMString,  CdkColumnDef>();
```
jelbourn: `ConstrainDOMString` wat

jelbourn: Maybe `_columnDefinitionsByName`?

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  // Placeholders within the table's template where the header and data rows will be inserted.
+  @ViewChild(CdkRowPlaceholder) rowPlaceholder: CdkRowPlaceholder;
+  @ViewChild(CdkHeaderRowPlaceholder) headerRowPlaceholder: CdkHeaderRowPlaceholder;
```
jelbourn: These ViewChildren and ContentChildren should be internal (`_` prefix)

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+    this.columnDefinitions.forEach(columnDef => this._columnMap.set(columnDef.name, columnDef));
+  }
+
+  ngAfterViewInit() {
+    // TODO: Re-render the header when the header's columns change.
+    this.renderHeaderRow();
+
+    // TODO: Re-render rows when their list of columns change.
+    // TODO: If the data source is not present after view init, connect it when it is defined.
+    const connectFn = this.dataSource.connectTable.bind(this.dataSource);
```
jelbourn: Prefer arrow functions over using `.bind`

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+    this.viewChange.let(connectFn).subscribe((rowsData: any[]) => {
```
jelbourn: I don't quite follow why you use `let` in this way; why not subscribe directly to the output of `connect`?

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+@Component({
+  selector: 'cdk-header-row',
+  template: '<ng-container cdkCellOutlet></ng-container>',
+  host: {
+    'class': 'cdk-header-row',
+    'role': 'row',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class CdkHeaderRow { }
```
jelbourn: Might as well move each of these directives to their own file; the table itself will eventually get more complicated and it will be less churn to move them now

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,323 @@
...
+  ngOnDestroy() {
+    // TODO: Disconnect from the data source so that it can unsubscribe from its streams.
+  }
+
+  ngOnInit() {
+    // TODO: Setup a listener for scroll events and emit the calculated view to this.viewChange
+  }
+
+  ngAfterContentInit() {
+    this.columnDefinitions.forEach(columnDef => this._columnMap.set(columnDef.name, columnDef));
```
jelbourn: Would it make sense to throw an error if multiple columns have the same name?

src/lib/core/data-table/data-table.ts
andrewseguin: Agreed, wasn't sure how to best split these up though

src/lib/core/data-table/data-table.ts
andrewseguin: I think its a legacy from prior designs, I don't think we need let at all anymore and can do what you're saying.

src/lib/core/data-table/cell.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/** Header cell template container that adds the right classes and role. */
+@Directive({
+  selector: 'cdk-header-cell',
+  host: {
+    'class': 'cdk-header-cell',
+    'role': 'columnheader',
+  },
+})
+export class CdkHeaderRowCell {
```
jelbourn: Just `CdkHeaderCell`

src/lib/core/data-table/cell.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(private columnDef: CdkColumnDef,
+              private elementRef: ElementRef,
+              private renderer: Renderer) {
+    this.renderer.setElementClass(elementRef.nativeElement, `cdk-column-${columnDef.name}`, true);
+  }
+}
+
+/** Cell template container that adds the right classes and role. */
+@Directive({
+  selector: 'cdk-row-cell',
```
jelbourn: Just `cdk-cell` / `CdkCell`

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,189 @@
...
+   * Create the embedded view for the data row template and place it in the correct index location
+   * within the data row view container.
+   */
+  insertRow(rowData: any) {
+    // TODO(andrewseguin): Add when predicates to the row definitions
+    // to find the right template to used based on
+    // the data rather than choosing the first row definition.
+    const row = this._rowDefinitions.first;
+
+    const context = {$implicit: rowData};
```
ErinCoughlan: Consider adding some of the other ngFor context stuff, like first, last, even, and odd. I can imagine wanting to make a table with alternating colored rows, where even would be useful.

(Or add a TODO)

src/lib/core/data-table/data-table.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  ngAfterViewInit() {
+    // TODO(andrewseguin): Re-render the header when the header's columns change.
+    this.renderHeaderRow();
+
+    // TODO(andrewseguin): Re-render rows when their list of columns change.
+    // TODO(andrewseguin): If the data source is not
+    // present after view init, connect it when it is defined.
+    this.dataSource.connectTable(this.viewChanges).subscribe((rowsData: any[]) => {
```
ErinCoughlan: We got into a pattern of using takeUtil to help clean up subscriptions, which I like and makes it easy to manage subscriptions.

```
private destroy = new Subject<void>();

ngAfterViewInit() {
  this.dataSource.connectTable(this.viewChanges)
    .takeUntil(this.destroy)
    .subscribe(...);
}

ngOnDestroy() {
  this.destroy.complete();
}
```

andrewseguin: This looks like a promising solution to subscriptions across a lot of our components. I'll add a TODO to resolve this. Likely if we go this route, we'll want to put it in several other components as well so it might be best in another PR

andrewseguin:
Comments have been addressed, ready for re-review.

leshin:
Quickly!

src/lib/matchers.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Collection of useful custom jasmine matchers for tests.
+ */
+export const customMatchers: jasmine.CustomMatcherFactories = {
```
jelbourn: I think that, as the build is set up now, this would end up in the released package.

@DevVersion?

andrewseguin:
Changes have been made (except the customMatchers part, waiting for input from DevVersion)

src/lib/matchers.ts
```diff
@@ -0,0 +1,21 @@
...
+  toBeRole: function(util: jasmine.MatchersUtil,
+                     customEqualityTesters: Array<jasmine.CustomEqualityTester>) {
```
jelbourn: `jasmine.CustomEqualityTester[]`

andrewseguin: Got it

src/lib/core/data-table/data-table.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+  let dataSource: FakeDataSource;
+  let table: CdkTable;
+  let tableElement: HTMLElement;
+
+  beforeEach(async(() => {
+    jasmine.addMatchers(customMatchers);
+    jasmine.addMatchers(tableCustomMatchers);
+
+    TestBed.configureTestingModule({
+      imports: [CommonModule, CdkDataTableModule],
```
ErinCoughlan: I think you can skip CommonModule here too since it's an import to CdkDataTableModule and nothing in the template for this test component that needs CommonModule.

andrewseguin: Got it, good catch

src/lib/matchers.ts
DevVersion: As long as the `matchers.ts` file is not imported by any source file and it will be only imported by spec files it should be fine.

What I recommend though: Store the filer under `core/testing` and name it `jasmine-matchers.ts`

PR #4469, opened 2017-05-10
fix(autocomplete): reposition the panel when the amount of options changes

crisbeto:
Fixes the autocomplete panel not being repositioned when the amount of options changes. This is necessary, because adding extra options could push it out of the viewport.

**Note:** will look into the iOS test failure. It's most likely an issue with the test setup, rather than the new logic.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -410,6 +410,11 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
         switchMap(() => {
           this._resetActiveItem();
           this.autocomplete._setVisibility();
+
+          if (this.panelOpen) {
+            this._overlayRef!.updatePosition();
+          }
+
           return this.panelClosingActions;
         }),
         // when the first closing event occurs...
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -1538,6 +1538,39 @@ describe('MatAutocomplete', () => {
       subscription!.unsubscribe();
     }));
 
+    it('should reposition the panel when the amount of options changes', fakeAsync(() => {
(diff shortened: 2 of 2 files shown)


kara:
@crisbeto IIRC, we didn't update the position deliberately because it can be disorienting when the panel switches its position while you're typing. Specifically, if you're moving from more options to fewer options as you type (and the panel can now fit in its first preferred position), the panel moves back to its first position mid-word.

Example case:

1) Set it up so the autocomplete is at the bottom of the page and empty.
2) Click to open the panel. It should open above the input.
3) Begin typing "Alabama".

You can see it switch midway through typing, even though the two options would have fit above.

I definitely agree that the position should update if the panel starts to extend past the viewport (fewer options -> more options). Any way we can make it so the panel only updates position if the current position doesn't fit?

crisbeto:
We can't right now @kara, but we discussed with Jeremy a while ago that we also shouldn't re-fit the panel in the viewport when you're scrolling. Adding that to the API can be used in this case as well.

kara:
@crisbeto Seems like @jelbourn's overlay positioning changes will have implications here. Once that gets in, we can re-visit this?

crisbeto:
I'm not sure it will @kara. The positioning improvements will change how we manage the overlay's position in general, but I think that we'll still have to recalculate it when the amount of options changes.

kara:
We will probably still need to make changes, but it doesn't seem smart to do so right before major changes to how we handle positioning. Either way, let me know when this PR is again ready for review.

jelbourn:
@crisbeto is this worth revisiting once the new positioning is in?

crisbeto:
Yeah, I think this branch has lived through two iterations of the connected positioning now.

josephperrott:
@crisbeto now that the new positioning strategy is in place.   Should we rebase and get this in?

crisbeto:
Rebased and fixed some issues with the testing setup @josephperrott.

josephperrott:
It looks like the new test failed on travis?

crisbeto:
It's fixed now @josephperrott.

PR #4471, opened 2017-05-11
Disable text highlighting on md-nav-list with the ripple

theiliad:


theiliad:
I signed it!

jelbourn:
I don't believe we should be disabling user selection here. Feel free to comment if there's a compelling reason for this.

theiliad:
@jelbourn

You wouldn't want the user to double click on a button or interactive list-item, and all the text starts highlighting.

It's really part of the experience.

Angular Material 1 for instance has that feature on md-button & md-list-items with an ng-click attribute.

PR #4473, opened 2017-05-11
fix(checkbox): set pointer cursor for checkbox

### LINKED ISSUE #4185: Bug(checkbox): hover mouse pointer cursor
Currently when hovering on checkbox'es the mouse cursor does not change.

To be same as Material1 and change mouse cursor to pointer 

Angular 4.0.1,
Windows 10,
Chrome 57

DevVersion:
* Adds the pointer cursor to the whole checkbox layout to also indicate that you can also click on the checkbox container and not only on the label.

Fixes #4185


DevVersion:
@willshowell It does go back to `default` if the checkbox is disabled.

willshowell:
Oops - currently the label cursor doesn't revert back to `default` and I was basing on that. I see how it works now, thanks!

PR #4483, opened 2017-05-11
Update getting-started.md

christinakayastha:
Add note about importing the MaterialModule


willshowell:
This is already taken care of in Step 3 of getting-started, no?

christinakayastha:
Yeah sorry I just realized it!

PR #4486, opened 2017-05-11
fix(list): fix padding for list items

kara:
Now that ripples are always the first children of the list content div, the selector overriding the padding needs to be updated.


DevVersion:
Ugh. I wonder if there is a better approach of doing this. Looks very error-prone for future changes too.

kara:
@DevVersion Feel free to submit a PR after this if you can think of a better way. Currently Google tests are broken and need to be fixed.

DevVersion:
Yeah thanks for fixing it that quick! Just wondered if we can make it more flexible at some point.

PR #4500, opened 2017-05-12
feat(overlay): add scroll blocking strategy

crisbeto:
Adds the `BlockScrollStrategy` which, when activated, will prevent the user from scrolling. For now it is only used in the dialog.

Relates to #4093.

### DIFF
--- a/src/lib/core/_core.scss
+++ b/src/lib/core/_core.scss
@@ -45,7 +45,13 @@
   // Used when disabling global scrolling.
   .cdk-global-scrollblock {
     position: fixed;
+
+    // Necessary for iOS not to expand past the viewport.
+    max-width: 100vw;
+
+    // Note: this will always add a scrollbar to whatever element it is on, which can
+    // potentially result in double scrollbars. It shouldn't be an issue, because we won't
+    // block scrolling on a page that doesn't have a scrollbar in the first place.
     overflow-y: scroll;
-    max-width: 100vw; // necessary for iOS not to expand past the viewport.
   }
 }
--- a/src/lib/core/_core.scss
+++ b/src/lib/core/_core.scss
@@ -41,17 +41,4 @@
   .mat-theme-loaded-marker {
     display: none;
   }
-
-  // Used when disabling global scrolling.
-  .cdk-global-scrollblock {
-    position: fixed;
-
-    // Necessary for iOS not to expand past the viewport.
-    max-width: 100vw;
-
-    // Note: this will always add a scrollbar to whatever element it is on, which can
-    // potentially result in double scrollbars. It shouldn't be an issue, because we won't
-    // block scrolling on a page that doesn't have a scrollbar in the first place.
-    overflow-y: scroll;
-  }
 }
--- a/src/lib/core/_core.scss
+++ b/src/lib/core/_core.scss
@@ -41,4 +41,11 @@
   .mat-theme-loaded-marker {
     display: none;
   }
+
+  // Used when disabling global scrolling.
+  .cdk-global-scrollblock {
+    position: fixed;
+    overflow-y: scroll;
+    max-width: 100vw; // necessary for iOS not to expand past the viewport.
+  }
 }
--- a/src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
+++ b/src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
@@ -10,12 +10,12 @@ describe('BlockScrollStrategy', () => {
   let forceScrollElement: HTMLElement;
 
   beforeEach(async(() => {
-    TestBed.configureTestingModule({ imports: [OverlayModule] }).compileComponents();
+    TestBed.configureTestingModule({imports: [OverlayModule]}).compileComponents();
   }));
 
-  beforeEach(inject([ViewportRuler], (_viewportRuler: ViewportRuler) => {
-    strategy = new BlockScrollStrategy(_viewportRuler);
-    viewport = _viewportRuler;
+  beforeEach(inject([ViewportRuler], (viewportRuler: ViewportRuler) => {
+    strategy = new BlockScrollStrategy(viewportRuler);
+    viewport = viewportRuler;
     forceScrollElement = document.createElement('div');
     document.body.appendChild(forceScrollElement);
     forceScrollElement.style.width = '100px';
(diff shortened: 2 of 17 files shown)


src/lib/core/overlay/scroll/block-scroll-strategy.ts
```diff
@@ -0,0 +1,60 @@
+import {ScrollStrategy} from './scroll-strategy';
+import {ViewportRuler} from '../position/viewport-ruler';
+
+/**
+ * Strategy that will prevent the user from scrolling while the overlay is visible.
+ */
+export class BlockScrollStrategy implements ScrollStrategy {
+  private _prevHTMLStyles = { top: null, left: null };
```
jelbourn: `_previousHtmlElementStyles` ?

src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+
+
+describe('BlockScrollStrategy', () => {
+  let platform = new Platform();
+  let strategy: BlockScrollStrategy;
+  let viewport: ViewportRuler;
+  let forceScrollElement: HTMLElement;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({ imports: [OverlayModule] }).compileComponents();
```
jelbourn: No spaces inside object literals

src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  }));
+
+  beforeEach(inject([ViewportRuler], (_viewportRuler: ViewportRuler) => {
```
jelbourn: Don't use `_` for function params; for `inject` I'm fine with just doing something like `vpr`

src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+
+
+  // In the iOS simulator (BrowserStack & SauceLabs), adding content to the
+  // body causes karma's iframe for the test to stretch to fit that content,
+  // in addition to not allowing us to set the scroll position programmatically.
+  // This renders the tests unusable and since we can't really do anything about it,
+  // we have to skip them on iOS.
+  function skipUnreliableBrowser(spec: Function) {
+    return () => {
+      if (!platform.IOS) { spec(); }
```
jelbourn: I don't want to _always_ skip iOS, since sometimes it does pass (and I think iOS 10 has this problem less frequently than iOS 9).

src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  afterEach(() => {
+    document.body.removeChild(forceScrollElement);
+    setScrollPosition(0, 0);
+  });
+
+  it('should toggle scroll blocking along the y axis', skipUnreliableBrowser(() => {
+    forceScrollElement.style.height = '3000px';
+
+    setScrollPosition(0, 100);
+    expect(viewport.getViewportScrollPosition().top).toBe(100,
```
jelbourn: Prefer breaking at the higher syntactic level, e.g.
```ts
expect(viewport.getViewportScrollPosition().top)
    .toBe(100, 'Expected viewport to be scrollable initially.');
```

src/lib/core/_core.scss
```diff
@@ -41,4 +41,11 @@
   .mat-theme-loaded-marker {
     display: none;
   }
+
+  // Used when disabling global scrolling.
+  .cdk-global-scrollblock {
+    position: fixed;
+    overflow-y: scroll;
```
jelbourn: Won't adding `overflow-y: scroll` cause a scrollbar to _always_ appear, even if there wasn't one there already? This could lead to an app suddenly multiple scrollbars when a, e.g., a dialog is opened.

src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      if (!platform.IOS) { spec(); }
+    };
+  }
+
+  function setScrollPosition(x: number, y: number) {
+    window.scroll(x, y);
+    viewport._cacheViewportGeometry();
+  }
+
+});
```
jelbourn: I think we should have an e2e test w/ screenshots for this as well that actually attempts to scroll the page through user input.

crisbeto: I agree, but wouldn't it be pointless since we only run the e2e tests against Chrome?

src/lib/core/_core.scss
crisbeto: It will, but we don't activate the strategy unless there was a scrollbar in the first place.

src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
crisbeto: I'm not a fan of skipping tests either, but I couldn't find a decent way to even scroll the page on iOS 10.3. E.g. something as simple as this doesn't work:

```ts
it('should scroll down', () => {
  forceScrollElement.style.height = '3000px';

  window.scroll(0, 100);
  viewport._cacheViewportGeometry();
  expect(viewport.getViewportScrollPosition().top).toBe(100);
});
```

This renders the entire test pointless since we can't verify whether we're actually doing anything.

jelbourn: In that case I would call the function `skipIOS` and expand the comment to include that code snippet (just the `window.scroll(0, 100); expect(viewport.getViewportScrollPosition().top).toBe(100);` part) as reasoning for why it's skipped.

src/lib/core/overlay/scroll/block-scroll-strategy.spec.ts
jelbourn: It's better than not having it; there were issues w/ the AngularJS version that manifested in Chrome.

src/lib/core/_core.scss
jelbourn: Should add a comment mentioning that

src/lib/core/_core.scss
```diff
@@ -41,4 +41,17 @@
   .mat-theme-loaded-marker {
     display: none;
   }
+
+  // Used when disabling global scrolling.
+  .cdk-global-scrollblock {
```
jelbourn: This should go into the `cdk-overlay` mixin

xtianus79:
Hi @jelbourn is this where a user is using a dialog box or sidenav(panel) flyout and the overlay content still scrolls up and down?  For now I used a method that alters the body element to position: fixed when opening one of these is opened up.

tobias74:
How can I use the BlockScrollStrategy on a Select-Dropdown?

willshowell:
@tobias74 answered at https://stackoverflow.com/a/46893614/4811678

PR #4501, opened 2017-05-12
feat(core): expose ScrollDispatchModule

DevVersion:
With #4251 the `ScrollDispatchModule` has been created and it is used by a lot of different components.

The NgModule is already exported inside of the file but the core package doesn't export it to the public.

**Note**: Modules like that (that can be also part of the CDK at some point) should be exposed. Also it is necessary to get the secondary entry-points working.


jelbourn:
@DevVersion needs rebase

DevVersion:
@jelbourn Done

PR #4503, opened 2017-05-12
fix(list, option): ripples disable mouse events on user content

### LINKED ISSUE #4480: mdOption: ripple blocks hover styles and events on parent's child
If you have a ripple on a component and add a child inside that component, the child can not have hover events or styles applied to it directly(you must apply to the parent). This is because the ripple is `position: absolute;` when it is not active(not doing the animation) and it covers the entire parent.

Ripples should allow hover effects and event listeners to work on a parents child without css/js hacks and workarounds.

The ripple blocks this because of it's styles when it is not active.

I believe this can be triggered with any ripple but I specifically encountered it with a custom `md-option` that had a component inside which listens for hover events. This also happens when trying to apply css affects.

See this.
http://plnkr.co/edit/0Dh65stBoG1v2VEAjvE5?p=preview

For me: showing user panels on hover without special overrides to attatch event listeners to a parent if the ripple is a sibling.

For everyone else: in general I think you'd run into this pretty easily. I also don't believe this bug is in Material for AngularJS, so anyone upgrading(like me) would be in trouble.

Angular: 4.1.1
Material: Beta 3
Browser: Chrome 58
OS: macOS 10.12.4

DevVersion:
Currently when a user places an element, that changes its color on hover, inside of a `md-option` or `md-list-item` the hover effect will never occur.

This is because the ripples are overlaying the user content and all pointer events are blocked accidentally.

Fixes #4480

### DIFF
--- a/src/lib/core/option/_option.scss
+++ b/src/lib/core/option/_option.scss
@@ -26,6 +26,11 @@
     bottom: 0;
     right: 0;
 
+    // Disable pointer events for the ripple container because the container will overlay the
+    // user content and we don't want to disable mouse events on the user content.
+    // Pointer events can be safely disabled because the ripple trigger element is the host element.
+    pointer-events: none;
+
     // In high contrast mode this completely covers the text.
     @include cdk-high-contrast {
       opacity: 0.5;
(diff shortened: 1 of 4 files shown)


epelc:
Thanks for the fix!

Not to bug you and it doesn't affect me but I believe buttons have the same problem.

DevVersion:
@epelc I will look at it tomorrow. Thanks

PR #4506, opened 2017-05-12
make input underline thick when in error state.

### LINKED ISSUE #3490: Invalid ("error") underline on form inputs should be 2px thick even when not focused
According to the [Material spec](https://material.io/guidelines/patterns/errors.html#errors-user-input-errors), (although it's not explicit), it looks like a form input's underline should be red if it's been touched and is invalid, even when it's in the focused state. Also, when it is in the unfocused state, it should be red and maintain the same thickness as in the focused state.

Currently, the underline is only red when it's touched/invalid and _un_focused. And when it is red in the unfocused state, it is thin, like the gray unfocused underline.

Our project requires the behavior that appears in the Material spec. While this can be done manually, it seems like it would be reasonable default behavior.

Thanks!

mmalerba:
also add underline ripple effect from the mocks

demo: https://mmalerba-demo1.firebaseapp.com/input

fixes #3490

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -74,10 +74,6 @@
       color: $input-underline-color-warn;
     }
 
-    .mat-input-underline {
-      border-color: $input-underline-color-warn;
-    }
-
     .mat-input-ripple {
       background-color: $input-underline-color-warn;
     }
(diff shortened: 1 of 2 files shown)


willshowell:
Shouldn't it also be thicker under hover too?

screen shot 2017-05-12 at 2 39 48 pm

And maybe it isn't as high of a priority and could be done later, but it looks like switching to error state should also trigger a ripple. [Seen here](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0B8wSqcLwbhFuSkY0bVhJaGNlMGs/TextFields-ErrorDark-01.webm)

Issue #4035, opened 2017-04-11
Tab body content with scrollbar


rtrompier:
Since beta-3, all the tabs content have a scrollbar.

Check the mat-tab-body-content class => height: 100%;

jelbourn:
@crisbeto I noticed this on the docs site too

naveedahmed1:
In my case since beta-4, now when I switch the tabs it shows the scrollbar while the animations is in progress. If I unset height everything works fine.

```
.mat-tab-body-content{
    height:unset!important;
}
```

jelbourn:
@naveedahmed1 please open a new issue if you're encountering a new bug

naveedahmed1:
Seems @crisbeto already have a pull request waiting to be merged regarding this issue, https://github.com/angular/material2/pull/4524 do I still need to open a new issue?

PR #4524, opened 2017-05-13
fix(tabs): dynamic tab scrollbar showing while animating

crisbeto:
Fixes an issue that caused a scrollbar to be shown only while a tab inside a dynamic height tab group is animating.

Relates to #4035.

For reference:
ref


sastack:
My problem is that my Tabs always have scrollbar, and I have put classes like .mat-tab-body-content{
  height:unset!important;
}
and nothing happens!

PR #4526, opened 2017-05-13
fix(button): ripples blocking mouse events on user content

DevVersion:
* The ripple overlay intercepts all mouse events on the user content and therefore users are not able to have interactive elements in their buttons.
   Normally buttons should not contain interactive elements inside, but since the fix is very easy it wouldn't hurt fixing it.

* Removing the `(touchstart)` event that should prevent initial clicks because it didn't do anything (because of pointer-events: none) and the initial click doesn't
   seem to happen anymore.

Related to #4503

### DIFF
--- a/src/lib/button/button.scss
+++ b/src/lib/button/button.scss
@@ -64,6 +64,11 @@
   left: 0;
   bottom: 0;
   right: 0;
+
+  // Disable pointer events for the ripple container and focus overlay because the container
+  // will overlay the user content and we don't want to disable mouse events on the user content.
+  // Pointer events can be safely disabled because the ripple trigger element is the host element.
+  pointer-events: none;
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/button/button.scss
```diff
@@ -64,6 +64,11 @@
...
   bottom: 0;
   right: 0;
+
+  // Disable pointer events for the ripple container and focus overlay because the container
+  // will overlay the user content and we don't want to disable mouse events on the user content.
+  // Pointer events can be safely disabled because the ripple trigger element is the host element.
+  pointer-events: none;
 }
 
 // Overlay to be used as a tint. Note that the same effect can be achieved by using a pseudo
```
crisbeto: The comment here needs to be updated as well.

DevVersion: Done.

jelbourn:
LGTM

PR #4527, opened 2017-05-13
build: publish to cdk-builds

DevVersion:
* Publish the CDK release output to the cdk-builds repository.

**Note**: Review using the diff that ignores whitespace changes: [Here](https://github.com/angular/material2/pull/4527/files?w=1)

### DIFF
--- a/scripts/release/publish-build-artifacts.sh
+++ b/scripts/release/publish-build-artifacts.sh
@@ -8,48 +8,67 @@ set -e -o pipefail
 # Go to the project root directory
 cd $(dirname $0)/../..
 
-buildDir="dist/releases/material"
-buildVersion=$(sed -nE 's/^\s*"version": "(.*?)",$/\1/p' package.json)
+# Material packages that need to published.
+PACKAGES=(cdk material)
+REPOSITORIES=(cdk-builds material2-builds)
 
-commitSha=$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


scripts/release/publish-build-artifacts.sh
```diff
@@ -8,48 +8,67 @@ set -e -o pipefail
...
-commitMessage=$(git log --oneline -n 1)
+# Function to publish artifacts of a package to Github.
+#   @param $1 Name of the package
+#   @param $2 Repository name of the package.
+publishPackage() {
 
-repoName="material2-builds"
-repoUrl="https://github.com/angular/material2-builds.git"
-repoDir="tmp/$repoName"
+  buildDir="dist/releases/$1"
```
jelbourn: Can you add 
```sh
packageName=${1}
publishRepoName=${2}
```
And then use those variables for the rest of the function?

Also, FYI, it's generally considered good practice to always use curly braces when referencing bash vars:
http://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-in-variables-using-bash

DevVersion: Done.

Also regarding the curly braces. Switched all variables to use curly braces. Just wanted to share that it sometimes feel weird to do it for simple things like:

```bash
cp $repoDir/* $DEST
```

Issue #2250, opened 2016-12-16
md-select option not selecting with objects.


paistipoikka:
#### Bug, feature request, or proposal:

Feature request/ bug. 

#### What is the expected behavior?

Needs a "selected" attribute so the developer can specify the "selected" rule/logic.

#### What is the current behavior?

Option not being selected although the underlying model is populated.

#### What are the steps to reproduce?

Unable to select options using objects where some properties of the two objects differ yet primary key and Display value are the same. For example, 

```
<md-option *ngFor="let user of Users" [value]="user">
    {{ user.name }}
</md-option>
```

The user may have one set of data inside the model yet only retrieve a subset of the properties to populate the select list. In this case, the options of the select do not select. 

For such cases, it may be better to do something like:

```
<md-option *ngFor="let user of Users" [value]="user" [selected]="user.id === userId">
    {{ user.name }}
</md-option>
```

#### What is the use-case or motivation for changing an existing behavior?

Using forms where objects are part of the underlying model. These objects are used as values of each <md-option>.

#### Which versions of Angular, Material, OS, browsers are affected?

Angular 2 latest.

da45:
same for me, selected is not working, I hope someone can answer us :)

michaelwaihenya:
+1

krigton:
+1

mwamufiya:
There seem to be multiple issues reported for being able to specify which is the selected option. Has there been an official response on this yet?

kara:
It seems like there is some confusion over how to use the select.  `md-select` is a custom form control, so you'd use it like you would a native select with any of the core form directives. 

Example: 

```html
<md-select placeholder="Food" [(ngModel)]="food">
   <md-option *ngFor="let food of foods" [value]="food"> {{ food.name }} </md-option>
</md-select>
```

Adding support for the `selected` attribute in addition to the existing mechanism would be duplicative and probably cause value conflicts. There should only be one source of truth for the select's value, and currently that is the form directive on the select. 

Example plunker for reactive version: 
http://plnkr.co/edit/RdsvzIO4IXeHRZGJHY2v?p=preview

You can find the documentation for select [here](https://github.com/angular/material2/tree/master/src/lib/select). It should be pushed to the formal docs site with the next release.

[...]

Issue #2785, opened 2017-01-24
md-select: add compareWith function


kara:
Currently, md-select values are compared by object reference. This is a bit inflexible. We want to allow people to compare options by a custom comparator in case they only care about the object's ID or some other property.

Example markup:

```html
<md-select placeholder="Food" [(ngModel)]="food" [compareWith]="compareById">
   <md-option *ngFor="let food of foods" [value]="food"> {{ food.name }} </md-option>
</md-select>
```

```ts
compareById(obj1, obj2) {
   return obj1.id === obj2.id;
}

```

This will be parallel to [functionality being added to native selects in core](https://github.com/angular/angular/pull/13349).

jamalx31:
+1

am-a:
+1

paistipoikka:
@kara is this still being considered?

mackva:
What news?

linuradu:
Sounds good, regarding commit :) When will be the next release?

pdavin:
Any update on this? Looks like there's a fix in a forked branch but not submitted for PR / review yet.

kweinmeister:
It looks like there is a PR, but it now has a merge conflict: https://github.com/angular/material2/pull/4540

jeanpchr:
Has anyone implemented a workaround?

ppham27:
I have PR #4540 open. I'm pretty good about rebasing it pretty often, so if you really need the feature, you could build against my branch. I'm not sure when/if they plan on merging it into master.

reppners:
Desperately need this 👍

PR #4540, opened 2017-05-14
feat(select): Implement compareWith so custom comparators can be used.

ppham27:
Fixes Issue #2250 and Issue #2785. Users can supply a custom comparator that tests for equality. The comparator can be changed dynamically at which point the selection may change. If the comparator throws an exception, it will log a warning in developer mode but will be swallowed in production mode.

### DIFF
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -129,5 +129,32 @@ Space above cards: <input type="number" [formControl]="topHeightCtrl">
     </md-card>
   </div>
 
+  <div *ngIf="showSelect">
+    <md-card>
+      <md-card-subtitle>compareWith</md-card-subtitle>
+      <md-card-content>
+        <md-select placeholder="Drink" [color]="drinksTheme"
+                   [ngModel]="currentDrinkObject"
+                   (ngModelChange)="setDrinkObjectByCopy($event)"
+                   [required]="drinkObjectRequired"
+                   [compareWith]="compareByValue ? compareDrinkObjectsByValue : compareByReference"
+                   #drinkObjectControl="ngModel">
+          <md-option *ngFor="let drink of drinks" [value]="drink" [disabled]="drink.disabled">
+            {{ drink.viewValue }}
+          </md-option>
+        </md-select>
+        <p> Value: {{ currentDrinkObject | json }} </p>
+        <p> Touched: {{ drinkObjectControl.touched }} </p>
+        <p> Dirty: {{ drinkObjectControl.dirty }} </p>
+        <p> Status: {{ drinkObjectControl.control?.status }} </p>
+        <p> Comparison Mode: {{ compareByValue ? 'VALUE' : 'REFERENCE' }} </p>
+
+        <button md-button (click)="drinkObjectRequired=!drinkObjectRequired">TOGGLE REQUIRED</button>
+        <button md-button (click)="compareByValue=!compareByValue">TOGGLE COMPARE BY VALUE</button>
+        <button md-button (click)="drinkObjectControl.reset()">RESET</button>
+      </md-card-content>
+    </md-card>
+  </div>
+
 </div>
 <div style="height: 500px">This div is for testing scrolled selects.</div>
--- a/src/demo-app/select/select-demo.html
+++ b/src/demo-app/select/select-demo.html
@@ -96,6 +96,32 @@ Space above cards: <input type="number" [formControl]="topHeightCtrl">
     </md-card-content>
   </md-card>
 
+
+  <md-card>
+    <md-card-subtitle>compareWith</md-card-subtitle>
+    <md-card-content>
+      <md-select placeholder="Drink" [color]="drinksTheme"
+                 [ngModel]="currentDrinkObject"
+                 (ngModelChange)="setDrinkObjectByCopy($event)"
+                 [required]="drinkObjectRequired"
+                 [compareWith]="compareByValue ? compareDrinkObjectsByValue : compareByReference"
+                 #drinkObjectControl="ngModel">
+        <md-option *ngFor="let drink of drinks" [value]="drink" [disabled]="drink.disabled">
+          {{ drink.viewValue }}
+        </md-option>
+      </md-select>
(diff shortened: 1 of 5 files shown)


willshowell:
@ppham27 Hey would you mind rebasing this?

ppham27:
Rebased. Screenshot tests are failing because of the elevation change in https://github.com/angular/material2/pull/5361.

paistipoikka:
@kara Is this still being considered?

src/lib/select/select.ts
```diff
@@ -337,6 +341,25 @@ export class MdSelect extends _MdSelectMixinBase implements AfterContentInit, On
...
+  /**
+   * A function to compare the option values with the selected values. The first argument
+   * is a value from an option. The second is a value from the selection. A boolean
+   * should be returned.
+   */
+  @Input()
+  get compareWith() { return this._compareWith; }
+  set compareWith(fn: (o1: any, o2: any) => boolean) {
+    if (typeof fn !== 'function') {
+      throw new TypeError(
```
kara: Can you move this error into `select-errors.ts`?

src/lib/select/select.spec.ts
```diff
@@ -3246,3 +3316,40 @@ class SelectWithCustomTrigger {
...
+  useCompareByReference() { this.comparator = this.compareByReference; }
+
+  useNullComparator() { this.comparator = null; }
+
+  compareByValue(f1: any, f2: any) { return f1 && f2 && f1.value === f2.value; }
+
+  compareByReference(f1: any, f2: any) { return f1 === f2; }
+
+  setFoodByCopy(newValue: {value: string, viewValue: string}) {
+    this.selectedFood = Object.assign({}, newValue);
```
kara: Can you use our `object.extend` utility here from core/utils?  `Object.assign` won't pass our internal tests.

src/lib/select/select.spec.ts
```diff
@@ -2690,8 +2693,74 @@ describe('MdSelect', () => {
...
+    let fixture: ComponentFixture<NgModelCompareWithSelect>;
+    let instance: NgModelCompareWithSelect;
+
+    beforeEach(async(() => {
+      fixture = TestBed.createComponent(NgModelCompareWithSelect);
+      instance = fixture.componentInstance;
+      fixture.detectChanges();
+    }));
+
+    it('should have a selection', () => {
```
kara: Nit: Can you wrap these two tests in a describe that says 'when comparing by value`, so it's clear what comparator is being used?

src/lib/select/select.spec.ts
```diff
@@ -2690,8 +2693,74 @@ describe('MdSelect', () => {
...
+      it('should use the comparator', () => {
+        expect(instance.compareByReference).toHaveBeenCalled();
+      });
+
+      it('should initialize with no selection despite having a value', () => {
+        expect(instance.selectedFood.value).toBe('pizza-1');
+        expect(instance.select.selected).toBeUndefined();
+      });
+
+      it('should not update the selection when changing the value', async(() => {
```
kara: Nit: I think this test name could be misleading. Selects comparing by reference should still update the selection, generally speaking. It's only when you manually copy over every ngModelChange that you see this behavior.  Maybe: "should not update the selection if value is copied on change" ?

src/demo-app/select/select-demo.html
```diff
@@ -129,5 +129,32 @@
     </md-card>
   </div>
 
+  <div *ngIf="showSelect">
+    <md-card>
+      <md-card-subtitle>compareWith</md-card-subtitle>
+      <md-card-content>
+        <md-select placeholder="Drink" [color]="drinksTheme"
+                   [ngModel]="currentDrinkObject"
+                   (ngModelChange)="setDrinkObjectByCopy($event)"
```
kara: Hmm, this is a bit odd to use for manual testing because when comparing by reference, it looks broken. I think it might be clearer if you use a regular two-way binding on `ngModel`, then have a button that says "REASSIGN DRINKS" or "COPY DRINKS" to test whether the value clears.

src/demo-app/select/select-demo.html
```diff
@@ -95,5 +95,32 @@
     </md-card>
   </div>
 
+  <div *ngIf="showSelect">
```
ppham27: Done.

src/demo-app/select/select-demo.html
ppham27: Its appearing broken was done on purpose. The intent was to convey when it's necessary to use [compareWith]. I could add a short caption saying this. I am perhaps using the demo app incorrectly, though. Will await further comment before making a change.

src/lib/select/select.spec.ts
ppham27: Done.

src/lib/select/select.spec.ts
ppham27: Done.

src/lib/select/select.spec.ts
ppham27: Done. Is the spread operator ({...newValue}) not used?

src/lib/select/select.ts
ppham27: Done. Updated tests, too.

src/demo-app/select/select-demo.html
kara: I get the intent, but for other team members manually testing the select, I can see this being confusing. I'd prefer the button.

ppham27: Okay, that makes sense to me. I added a REASSIGN DRINK BY COPY button and used a two-way binding. Let me know if this is what you had in mind.

src/lib/select/select.spec.ts
rafaelss95: The spread operator isn't supported in many browsers. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator

kara: Spread operator should be fine for the demo.

crisbeto: TypeScript will transpile the spread operator into something that can run everywhere.

PR #4542, opened 2017-05-14
fix(datepicker): set cursor on clickable elements

### LINKED ISSUE #4533: Bug(datepicker) month day selection cursor issue
When hovering over month day text mouse cursor to be "pointer", like in material 1.

Mouse cursor is "text"

https://material.angular.io/components/component/datepicker go to examples and start selecting for dates and all month days cursor is "text" type.

To be same like material 1.

Material 2.0.0-beta.5
Google Chrome  58

crisbeto:
* Adds the `pointer` cursor to the calendar toggle and date cells, indicating that they're clickable.
* Sets the `default` cursor on the disabled states, instead of the `text` cursor.

Fixes #4533.


mmalerba:
Thanks for all the little fixes!

PR #4544, opened 2017-05-14
fix(list-key-manager): remove handling for home and end keys

### LINKED ISSUE #3496: autocomplete - HOME and END keys not working

crisbeto:
* Removes the handling for the home and end keys from the ListKeyManager since their usage isn't as generic as the arrow keys.
* Adds the home and end key handling to the select specifically.
* Reworks the Autocomplete, ListKeyManager and ChipList tests to use the `createKeyboardEvent`, instead of making their own fake keyboard events.
* Adds a `target` parameter to the `createKeyboardEvent` function, allowing us to define the event target.

Fixes #3496.

### DIFF
--- a/src/lib/core/a11y/list-key-manager.spec.ts
+++ b/src/lib/core/a11y/list-key-manager.spec.ts
@@ -1,9 +1,11 @@
 import {QueryList} from '@angular/core';
 import {fakeAsync, tick} from '@angular/core/testing';
 import {FocusKeyManager} from './focus-key-manager';
-import {DOWN_ARROW, UP_ARROW, TAB, HOME, END} from '../keyboard/keycodes';
+import {DOWN_ARROW, UP_ARROW, TAB} from '../keyboard/keycodes';
 import {ListKeyManager} from './list-key-manager';
 import {ActiveDescendantKeyManager} from './activedescendant-key-manager';
+import {createKeyboardEvent} from '../testing/event-objects';
+
 
 class FakeFocusable {
   disabled = false;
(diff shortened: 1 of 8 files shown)


jelbourn:
I did a quick survey of Google Search, GMail, Drive, and [this accessible autocomplete for US government web sites](https://assets.cms.gov/resources/framework/3.4.1/Pages/#autocomplete). For all of them, the home/end keys always apply to the text input no mater what state you're in. I think we should just follow this pattern, since it will be simpler and follow conventions.

crisbeto:
Done @jelbourn. I removed it from the `ListKeyManager` and re-added it only to `MdSelect` since the native select does handle home and end.

src/lib/core/a11y/list-key-manager.spec.ts
```diff
@@ -24,43 +26,32 @@ class FakeQueryList<T> extends QueryList<T> {
...
-}
 
 describe('Key managers', () => {
   let itemList: FakeQueryList<any>;
-  let DOWN_ARROW_EVENT: KeyboardEvent;
-  let UP_ARROW_EVENT: KeyboardEvent;
-  let TAB_EVENT: KeyboardEvent;
-  let HOME_EVENT: KeyboardEvent;
-  let END_EVENT: KeyboardEvent;
+  let EVENTS: {
```
jelbourn: How about just `fakeKeyEvents`? It's not constant since you write the value in `beforeEach`.

PR #4549, opened 2017-05-14
build: payload github status

DevVersion:
* Show payload deta on PRs using the Github Statuses API.

### DIFF
--- a/tools/gulp/util/firebase.ts
+++ b/tools/gulp/util/firebase.ts
@@ -5,11 +5,15 @@ const cloudStorage = require('@google-cloud/storage');
 // Firebase configuration for the Screenshot project. Use the config from the screenshot functions.
 const screenshotFirebaseConfig = require('../../screenshot-test/functions/config.json');
 
-/** Opens a connection to the Firebase dashboard app. */
-export function openFirebaseDashboardApp() {
+/** Database URL of the dashboard firebase project.*/
+const dashboardDatabaseUrl = 'https://material2-board.firebaseio.com';
+
+/** Opens a connection to the Firebase dashboard app using a service account. */
+export function openFirebaseDashboardApp(asGuest = false) {
   // Initialize the Firebase application with firebaseAdmin credentials.
   // Credentials need to be for a Service Account, which can be created in the Firebase console.
   return firebaseAdmin.initializeApp({
+    databaseURL: dashboardDatabaseUrl,
     credential: firebaseAdmin.credential.cert({
       project_id: 'material2-board',
       client_email: 'material2-board@appspot.gserviceaccount.com',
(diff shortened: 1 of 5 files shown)


DevVersion:
Now since we can also build HTTP API's using Firebase we can use that for the payload status.

- https://firebase.google.com/docs/functions/http-events

Those functions can will be used for the team dashboard too.

DevVersion:
@jelbourn Reworked this to use the Firebase function now. Once we fully switched to the new firebase instance and the remote config is set up we can just replace the function URL in a follow-up PR and proceed.

jelbourn:
@DevVersion needs rebase now

DevVersion:
@jelbourn Rebased.

tools/gulp/util/firebase.ts
```diff
@@ -6,18 +6,26 @@ const cloudStorage = require('@google-cloud/storage');
 const screenshotFirebaseConfig = require('../../screenshot-test/functions/config.json');
 
 /** Opens a connection to the Firebase dashboard app. */
-export function openFirebaseDashboardApp() {
+export function openFirebaseDashboardApp(asGuest = false) {
+  const databaseURL = 'https://material2-board.firebaseio.com';
```
jelbourn: `databaseUrl`

tools/gulp/util/firebase.ts
```diff
@@ -6,18 +6,26 @@ const cloudStorage = require('@google-cloud/storage');
...
+
+  // In some situations the service account credentials are not available and authorizing as
+  // a guest works fine. For example in Pull Requests the payload task just wants to read data.
+  if (asGuest) {
+    return firebase.initializeApp({ databaseURL });
+  }
```
jelbourn: I'd just make this a different function: `openFirebaseDashboardAppAsGuest`

(same thing about flag arguments https://martinfowler.com/bliki/FlagArgument.html)

tools/gulp/util/firebase.ts
```diff
@@ -6,18 +6,26 @@ const cloudStorage = require('@google-cloud/storage');
...
 
 /** Opens a connection to the Firebase dashboard app. */
-export function openFirebaseDashboardApp() {
+export function openFirebaseDashboardApp(asGuest = false) {
+  const databaseURL = 'https://material2-board.firebaseio.com';
+
+  // In some situations the service account credentials are not available and authorizing as
+  // a guest works fine. For example in Pull Requests the payload task just wants to read data.
+  if (asGuest) {
+    return firebase.initializeApp({ databaseURL });
```
jelbourn: Is there any reason not to _always_ authenticate as guest for diffs?

tools/gulp/util/firebase.ts
DevVersion: I only did it like that because the `firebase-admin` also uses it like that and so I can just pass it like that. But can change it.

tools/gulp/util/firebase.ts
DevVersion: Okay I'm fine doing this.

tools/gulp/util/firebase.ts
DevVersion: For master builds we need to upload data using `firebase-admin`.

DevVersion:
@jelbourn Addressed your feedback.

andrewseguin:
Needs rebase

DevVersion:
@andrewseguin Rebased.

PR #4565, opened 2017-05-15
fix(datepicker): minValidator & maxValidation false errors

PhilipSultanescu:
minValidator & maxValidation should not give errors when value = min or max

Example setting the [min]="today" and [value]="today" will trigger the mdDatepickerMin validation error, but it shouldn't. Same goes for [max]. It should include as valid also the min and max dates.

Tested with v2.0.0.beta.5

### DIFF
--- a/src/lib/core/overlay/index.ts
+++ b/src/lib/core/overlay/index.ts
@@ -4,19 +4,12 @@ export {FullscreenOverlayContainer} from './fullscreen-overlay-container';
 export {OverlayRef} from './overlay-ref';
 export {OverlayState} from './overlay-state';
 export {ConnectedOverlayDirective, OverlayOrigin, OverlayModule} from './overlay-directives';
-export {ScrollDispatcher} from './scroll/scroll-dispatcher';
 export {ViewportRuler} from './position/viewport-ruler';
 
 export * from './position/connected-position';
+export * from './scroll/index';
 
 // Export pre-defined position strategies and interface to build custom ones.
 export {PositionStrategy} from './position/position-strategy';
 export {GlobalPositionStrategy} from './position/global-position-strategy';
 export {ConnectedPositionStrategy} from './position/connected-position-strategy';
-
-// Export pre-defined scroll strategies and interface to build custom ones.
-export {ScrollStrategy} from './scroll/scroll-strategy';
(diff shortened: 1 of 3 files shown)


PhilipSultanescu:
I signed it!

PhilipSultanescu:
I signed it!

PhilipSultanescu:
I signed it!

PhilipSultanescu:
I signed it!

jelbourn:
@mmalerba not sure if you had a reason for having it this way. If we do want this change, it needs unit tests.

mmalerba:
@jelbourn nope, just a mistake.

@PhilipSultanescu mind adding unit tests that would have caught this?

PhilipSultanescu:
Plunker: just focus in and out to trigger the validation.
https://plnkr.co/edit/VtxBbgkwAohgsrsOnmLp?p=preview

@mmalerba, @jelbourn Working now on the unit tests. Will also switch my CLA to Corporate, pending approvals.

PhilipSultanescu:
Tests commited.

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -433,6 +433,58 @@ describe('MdDatepicker', () => {
         expect(testComponent.datepicker._minDate).toEqual(new Date(2010, JAN, 1));
         expect(testComponent.datepicker._maxDate).toEqual(new Date(2020, JAN, 1));
       });
+
+      it('should mark input invalid', async(() => {
```
mmalerba: can you split this into separate tests?
```
'should mark invalid when value is before min'
'should mark invalid when value is after max'
'should not mark invalid when value equals min'
'should not mark invalid when value equals max'
'should not mark invalid when value is between min and max'
```
this way we avoid the deep nesting and if we get a failure its easier to see why

PhilipSultanescu: tried it but got errors when split, will debug it more tomorrow to find the cause and reply.

mmalerba:
LGTM, but can you squash the commits? The tool we use to merge into Google's codebase doesn't like the merge commits. I think it might also fix the cla error.

PhilipSultanescu:
Closed this request because of too many commit conflicts, will reopen using a feature branch.
Maybe it will also solve the CLA problem.

PR #4569, opened 2017-05-15
fix(input): continue checking for input child after initialization

### LINKED ISSUE #4551: at MdInputContainer._isErrorState
at MdInputContainer._isErrorState (material.es5.js:15396)

ERROR TypeError: Cannot read property '_ngControl' of undefined
the code is here:
https://github.com/keyeMyria/bookingmanagement/tree/material2

and after run acess: http://localhost:4400/session/signin/
errors show.

crisbeto:
Currently we only check if an `md-input-container` has a child upon initialization, however the child might be removed via `ngIf` at a later point. If that happens, we eventually run into some check that assumed that we have an input child, however it would be better if we threw the appropriate error. These changes add extra validation that ensure that the input continues to be in place after init.

Fixes #4551.

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -1,5 +1,6 @@
 import {
   AfterContentInit,
+  AfterContentChecked,
   AfterViewInit,
   ChangeDetectorRef,
   Component,
@@ -286,7 +287,7 @@ export class MdInputDirective {
   },
   encapsulation: ViewEncapsulation.None,
 })
-export class MdInputContainer implements AfterViewInit, AfterContentInit {
+export class MdInputContainer implements AfterViewInit, AfterContentInit, AfterContentChecked {
   /** Alignment of the input container's content. */
   @Input() align: 'start' | 'end' = 'start';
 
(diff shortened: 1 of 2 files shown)


src/lib/input/input-container.ts
```diff
@@ -449,22 +451,33 @@ export class MdInputContainer implements AfterViewInit, AfterContentInit {
...
-    let startHint = this._hintChildren ?
-        this._hintChildren.find(hint => hint.align === 'start') : null;
-    let endHint = this._hintChildren ?
-        this._hintChildren.find(hint => hint.align === 'end') : null;
-
-    if (startHint) {
-      ids.push(startHint.id);
-    } else if (this._hintLabel) {
-      ids.push(this._hintLabelId);
+    if (this._mdInputChild) {
```
mmalerba: Do we need this if? It will error on `_validateInputChild` before getting to this

src/lib/input/input-container.ts
```diff
@@ -356,10 +357,7 @@ export class MdInputContainer implements AfterViewInit, AfterContentInit {
     @Optional() private _parentFormGroup: FormGroupDirective) { }
 
   ngAfterContentInit() {
-    if (!this._mdInputChild) {
-      throw new MdInputContainerMissingMdInputError();
-    }
-
+    this._validateInputChild();
```
mmalerba: not really related to this PR, but I see `_prefixChildren` and `suffixChildren` are unused, want to just get rid of them?

src/lib/input/input-container.ts
crisbeto: We do. This one seems to fire before the validation has had a chance to run.

src/lib/input/input-container.ts
crisbeto: [They're being used in the view.](https://github.com/angular/material2/blob/master/src/lib/input/input-container.html#L3)

crisbeto:
Rebased and addressed the feedback @mmalerba.

mmalerba:
lgtm

PR #4577, opened 2017-05-15
feat(focus-trap): allow setting initially focused element

### LINKED ISSUE #2224: md-sidenav mode="over" focus state
The base path remains in active state when you select another link and you close and reopen the sidenav, only in mode="over". Need an event, like a click, to deactivate it.

It has worked fine up to 2.0.0-alpha.10

http://plnkr.co/edit/R9p0F7

Material 2.0.0-alpha.11-xx, in all browsers.

I tried Angular workarounds (routerLinkActive) without success.

Thanks

mmalerba:
also fix `md-nav-list` focused item style.

fixes #2224

### DIFF
--- a/src/lib/core/a11y/focus-trap.spec.ts
+++ b/src/lib/core/a11y/focus-trap.spec.ts
@@ -173,9 +173,11 @@ class FocusTrapWithBindings {
   template: `
     <div cdkTrapFocus>
       <input>
+      <button>before</button>
+      <button id="first" cdk-focus-region-start></button>
+      <button id="middle" cdk-focus-initial></button>
       <button id="last" cdk-focus-region-end></button>
-      <button id="middle" cdk-focus-init></button>
-      <button id="first" cdk-focus-region-start>SAVE</button>
+      <button>after</button>
       <input>
     </div>
     `
--- a/src/lib/core/a11y/focus-trap.spec.ts
+++ b/src/lib/core/a11y/focus-trap.spec.ts
@@ -167,8 +167,8 @@ class FocusTrapWithBindings {
   template: `
     <div cdkTrapFocus>
       <input>
-      <button id="last" cdk-focus-end></button>
-      <button id="first" cdk-focus-start>SAVE</button>
+      <button id="last" cdk-focus-region-end></button>
+      <button id="first" cdk-focus-region-start>SAVE</button>
       <input>
     </div>
     `
--- a/src/lib/core/a11y/focus-trap.spec.ts
+++ b/src/lib/core/a11y/focus-trap.spec.ts
@@ -14,7 +14,7 @@ describe('FocusTrap', () => {
         FocusTrapWithBindings,
         SimpleFocusTrap,
         FocusTrapTargets,
-        FocusTrapWithSvg
+        FocusTrapWithSvg,
       ],
       providers: [InteractivityChecker, Platform, FocusTrapFactory]
     });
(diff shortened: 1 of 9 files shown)


src/lib/core/a11y/focus-trap.ts
```diff
@@ -97,27 +101,47 @@ export class FocusTrap {
     this._ngZone.onMicrotaskEmpty.first().subscribe(() => this.focusLastTabbableElement());
   }
 
+  private _getRegionMarker(marker: 'start' | 'end'): HTMLElement | null {
```
jelbourn: Add function description? Not immediately clear what a "region marker" is.

src/lib/list/list.ts
```diff
@@ -151,6 +149,7 @@ export class MdListItem implements AfterContentInit {
 
   constructor(private _renderer: Renderer2,
               private _element: ElementRef,
+              private _cdr: ChangeDetectorRef,
```
jelbourn: Where is this used?

src/lib/core/a11y/focus-trap.ts
jelbourn: Instead of "marker", maybe "boundary element"?

src/lib/core/a11y/focus-trap.ts
```diff
@@ -97,27 +101,47 @@ export class FocusTrap {
...
+    if (marker == 'start') {
+      return markers.length ? markers[0] : this._getFirstTabbableElement(this._element);
+    }
+    return markers.length ?
+        markers[markers.length - 1] : this._getLastTabbableElement(this._element);
+  }
+
+  /** Focuses the element that should be focused when the focus trap is initialized. */
+  focusInitialElement() {
+    let redirectToElement = this._element.querySelector('[cdk-focus-init]') as HTMLElement;
```
jelbourn: Write out `initial` (to disambiguate from "initialize")?

src/lib/core/a11y/focus-trap.spec.ts
```diff
@@ -167,8 +173,9 @@ class FocusTrapWithBindings {
   template: `
     <div cdkTrapFocus>
       <input>
-      <button id="last" cdk-focus-end></button>
-      <button id="first" cdk-focus-start>SAVE</button>
+      <button id="last" cdk-focus-region-end></button>
+      <button id="middle" cdk-focus-init></button>
+      <button id="first" cdk-focus-region-start>SAVE</button>
```
jelbourn: Why does start come after end here?

src/lib/core/a11y/focus-trap.ts
```diff
@@ -97,27 +101,47 @@ export class FocusTrap {
     this._ngZone.onMicrotaskEmpty.first().subscribe(() => this.focusLastTabbableElement());
   }
 
+  private _getRegionMarker(marker: 'start' | 'end'): HTMLElement | null {
+    let markers = [
+      ...Array.prototype.slice.call(this._element.querySelectorAll(`[cdk-focus-region-${marker}]`)),
+      ...Array.prototype.slice.call(this._element.querySelectorAll(`[cdk-focus-${marker}]`)),
```
jelbourn: Add comment explaining what's for backwards compatibility

src/lib/list/list.ts
mmalerba: nowhere, switched to adding class via renderer

src/lib/core/a11y/focus-trap.spec.ts
mmalerba: The test was already like this, I think the reasoning is to make sure that it's really grabbing the element with the attribute, not just the first or last one. I changed it to something less confusing.

mmalerba:
comments addressed, PTAL

PR #4578, opened 2017-05-15
fix(sidenav): don't restore focus if focus isn't inside sidenav

### LINKED ISSUE #4203: fix(sidenav): do not reset focus when closed if focus is already outside of sidenav
Bug (maybe)

When `MdSidenav` (in `side` mode) is closed, it tries to return the focus back to the element that was focused before the sidenav was opened ([here][1]). This might be fine most of the time, but it is too aggressive in some cases. Consider the following usecase:

When the currently focused element is outside of a closing sidenav, the sidenav should not attempt to return the focus to the element that was focused when the sidenav opened.

We ran into this on angular.io. Go to [aio-staging][2], click on "Docs" in the navbar (sidenav opens), then click on "Resources" (sidenav closes), see "Docs" getting the focus (default blue outline on Chrome).

[Here][3] is a simpler reproduction.

This is happening with v2.0.0-beta.3. All Angular/OS/browsers should be affected. Didn't check with previous versions of Material. 

[1]: https://github.com/angular/material2/blob/76faae5f378f2949bbe215df38369190773eba60/src/lib/sidenav/sidenav.ts#L146
[2]: https://aio-staging.firebaseapp.com/
[3]: http://plnkr.co/edit/T5Xu8Qy7ufNH6sm9qQBb?p=preview

mmalerba:
fixes #4203

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -12,12 +12,13 @@ import {
   Renderer2,
   ViewEncapsulation,
   NgZone,
-  OnDestroy,
+  OnDestroy, Inject,
 } from '@angular/core';
 import {Dir, MdError, coerceBooleanProperty} from '../core';
 import {FocusTrapFactory, FocusTrap} from '../core/a11y/focus-trap';
 import {ESCAPE} from '../core/keyboard/keycodes';
 import 'rxjs/add/operator/first';
+import {DOCUMENT} from '@angular/platform-browser';
 
 
 /** Exception thrown when two MdSidenav are matching the same side. */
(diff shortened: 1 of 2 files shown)


src/lib/sidenav/sidenav.ts
```diff
@@ -137,15 +137,24 @@ export class MdSidenav implements AfterContentInit, OnDestroy {
...
-    this.onClose.subscribe(() => {
+    this.onClose.subscribe(() => this._restoreFocus());
+  }
+
+  /**
+   * If focus is currently inside the sidenav, restores it to where it was before the sidenav
+   * opened.
+   */
+  private _restoreFocus() {
+    let activeEl = document.activeElement;
```
kara: It seems like we should be injecting the `Document` rather than using the global (like we do elsewhere). Is there a reason why we aren't doing it here?

mmalerba:
PTAL

PR #4598, opened 2017-05-16
docs(datepicker): explicitly mention i18n and warn people about pitfa…

mmalerba:
…lls of using NativeDateAdapter

@fxck

### DIFF
--- a/src/lib/datepicker/datepicker.md
+++ b/src/lib/datepicker/datepicker.md
@@ -118,7 +118,14 @@ export class MyComponent implements AfterViewInit {
 }
 ```
 
-### Choosing a date implementation and date format settings
+### Internationalization
+In order to support internationalization, the datepicker supports customization of the following
+three pieces via injection:
+ 1. The date implementation that the datepicker accepts.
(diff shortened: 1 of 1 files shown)


fxck:
Looks great.

src/lib/datepicker/datepicker.md
```diff
@@ -148,6 +155,12 @@ export class MyComponent {
 }
 ```
 
+*Please note: `MdNativeDateModule` is based off of the functionality available in JavaScript's
+native Date object, and is thus not suitable for many locales. One of the biggest shortcomings of
+the native Date object is the inability to set the parse format. We highly recommend writing a
+custom `DateAdapter` that works with the formatting/parsing library of your choice, or waiting for
+the official Moment.js module.*
```
jelbourn: On moment: I would rephrase this to not convey that Moment is the "official" way of doing i18n, but just that we plan on making a moment adapter available as a convenience

mmalerba: Ok, just removed the moment.js reference here since it's already mentioned in a few other places that we plan to make that

PR #4604, opened 2017-05-16
fix(datepicker): don't use pointer cursor on toggle when disabled

willshowell:
Related to https://github.com/angular/material2/pull/4542

@crisbeto @mmalerba


mmalerba:
LGTM

PR #4605, opened 2017-05-16
fix(chips): unable to tab out of chip list

### LINKED ISSUE #4593: Chips component has a partial keyboard trap
Bug

When using the keyboard to navigate the page, you should be able to tab into a chips component and use the arrow keys to navigate items, then tab or shift-tab to move away from the chips component.

You are able to forward tab out of the component, but shift-tab does not let you leave and move back to the previous item on the page.

This is evident on the component guides page: https://material.angular.io/components/component/chips

1. Click on overview.
2. Tab into the chip
3. Use the arrow keys to navigate through chips, use space bar to select.  All good.
4. Forward tab moves you to the next focusable element on the page, good.
5. Shift-tab back to the chips.
6. Shift-tab again, you are unable to lose focus on the first chip and navigate back to the previous item on the page.
7. Use the arrow keys to select a chip other than the first. Shift tab, you are taken back to the first chip.  Shift-tab again, you are unable to lose focus.

Accessibility issue

crisbeto:
* Fixes not being able to escape focus from an `md-chip-list` backwards via shift+tab on all browsers.
* Fixes not being able to tab out of an `md-chip-list` at all on Firefox.

Fixes #4593.

### DIFF
--- a/src/lib/chips/chip-list.ts
+++ b/src/lib/chips/chip-list.ts
@@ -3,16 +3,17 @@ import {
   ChangeDetectionStrategy,
   Component,
   ContentChildren,
-  ElementRef,
   Input,
   QueryList,
-  ViewEncapsulation
+  ViewEncapsulation,
+  OnDestroy,
 } from '@angular/core';
 
 import {MdChip} from './chip';
 import {FocusKeyManager} from '../core/a11y/focus-key-manager';
 import {coerceBooleanProperty} from '../core/coercion/boolean-property';
-import {SPACE, LEFT_ARROW, RIGHT_ARROW} from '../core/keyboard/keycodes';
+import {SPACE, LEFT_ARROW, RIGHT_ARROW, TAB} from '../core/keyboard/keycodes';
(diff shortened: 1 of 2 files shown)


src/lib/chips/chip-list.ts
```diff
@@ -59,11 +63,19 @@ export class MdChipList implements AfterContentInit {
   /** The chip components contained within this chip list. */
   chips: QueryList<MdChip>;
 
-  constructor(private _elementRef: ElementRef) { }
+  /** Tab index for the chip list. */
+  _tabIndex = 0;
```
tinayuangao: Why Shift+tab doesn't work on tabindex=0?

crisbeto: It's because the browsers determine that the previous tabbable element is the parent (since the chips are `tabindex="-1"` and the chip list is `tabindex="0"`). When the parent receives focus, it forwards it back to the first chip. As an aside, Firefox also tabs into the parent when tabbing forward.

PR #4608, opened 2017-05-16
fix(dialog): enable smooth scrolling on md-dialog-content

crisbeto:
Enables the smooth native scrolling on `md-dialog-content`.


jelbourn:
@crisbeto any non-comment changes to a source file (even css) aren't necessarily merge-safe because they could cause compilation errors in Google's internal css compiler (minifier).

Issue #3586, opened 2017-03-13
A11y keyboard focus control in the Radio Button Component


DaveBest99:
#### Bug, feature request, or proposal:
Bug:
The Tab key and Shift+Tab key should move focus to and and away from the Radio button component, and the UP/Down arrow keys should rotate focus between each Radio button in the group. The Tab key moves focus through all Radio buttons when none are selected, and only on the selected Radio button when one is checked. Arrow keys do not keep the user focus within the Radio button group, and the text label are not associated with the button element.

#### What is the expected behavior?
Expected behaviour of arrow keys within a Radio button group is to rotate between the buttons and not freely move through the page.

#### What is the current behavior?
Currently Tab and Arrow keys do not perform as expected.

#### What are the steps to reproduce?
See below.

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: https://goo.gl/DlHd6U

#### What is the use-case or motivation for changing an existing behavior?
To implement WCAG standards.

#### Which versions of Angular, Material, OS, browsers are affected?
Using JAWS17 screen reader with browsers Firefox, Chrome, and IE11.

#### Is there anything else we should know?
The screen reader user experience is as follows.

Using Firefox:

1. Press Tab key to the Radio button component.
JAWS says "Winter radio button not checked 1 of 4", and the braille display shows "< > Winter rbtn 1 of 4".
Note, the Down arrow key moves freely through all the Radio buttons and the user can press Spacebar to check a specific Radio button. The Up and Down arrow keys should rotate between the Radio buttons within the group and not freely move up/down the page.

2. Press Tab key again.
JAWS says "Spring radio button not checked 2 of 4", and the braille display shows "< > Spring rbtn 2 of 4".
Note, this is not the expected behaviour for the Tab key. Tab and Shift+Tab should move the focus to the next or previous page element outside of the Radio button group.

3. Press Tab key again.
JAWS says "Summer radio button not checked 3 of 4", and the braille display shows "< > Summer rbtn 3 of 4".
Note, this is not the expected behaviour for the Tab key. Tab and Shift+Tab should move the focus to the next or previous page element outside of the Radio button group.

4. Press Tab key again.
JAWS says "Autumn radio button not checked 4 of 4", and the braille display shows "< > Autumn rbtn 4 of 4".
Note, this is not the expected behaviour for the Tab key. Tab and Shift+Tab should move the focus to the next or previous page element outside of the Radio button group.

5. Press Tab key again.
[...]

PR #4610, opened 2017-05-16
fix(checkbox, radio): Fix checkbox label not read in JAWS

### LINKED ISSUE #3584: A11y keyboard focus control in Checkbox Component
But:
Using the JAWS screen reader with Firefox, Chrome, and Internet Explorer result with a different user experience. It appears the disassociation of "label", "span", and "input" elements may cause the screen reader focus to be incorrect.

When moving focus to a Checkbox the text label should be identified with the element, and it is expected that arrow keys will rotate between Radio buttons.

Currently text labels are not read properly for each element, and arrow keys do not stay within the Radio button group.

See below.

Latest

The user experience is as follows.

Using Firefox and Chrome:

1. Press Tab key to the Indeterminate Checkbox.
JAWS says "Indeterminate checkbox not checked" and the braille display is on the Checkbox "< > Indeterminate".

2. Press Spacebar to check the element.
JAWS says "checked" but the braille display shows "Align:".

3. Press Spacebar again to uncheck the element.
JAWS says "unchecked" but each time you press the Spacebar the braille display focus appears to be random; sometimes on the text "Align:", sometimes on the Checkbox 
"Indeterminate", and sometimes on the page Title "Angular Material".

4. Press Tab key to the Start Check Button.
JAWS says "start radio button checked 1 of 2", and the braille display shows "<x> Start rbtn 1 of 2".

5. Pressing Tab key again moves focus to the "Disabled" Checkbox element as expected. However, the Up/Down Arrow keys do not rotate the focus between the two Radio buttons. Arrow keys should rotate between the grouped Radio buttons and Tab or Shift+Tab moves focus away from the Radio button group.
With each press of the Down Arrow key JAWS reads:
DownArrow: "Start"
DownArrow: "Radio button not checked End 2 of 2"
DownArrow: "End"
DownArrow: "Disabled checkbox not checked"
DownArrow: "Disabled clickable"

Using Internet Explorer 11:

1. Press Tab key to the first component element.
JAWS says "checked checkbox not checked", and the braille display shows "< > Checked".
With each press of the Tab key the next element is spoken with an incorrect text label.

2. Press Tab key again.
JAWS says "checked checkbox not checked", and the braille display shows "< > Checked".

3. Press Tab key again.
JAWS says "Align: checkbox not checked", and the braille display shows "< > Align:".

4. Press Tab key again.
JAWS says "Indeterminate checkbox not checked", and the braille display shows "< > Indeterminate".

5. Press Tab key again.
JAWS says "Start radio button checked", and the braille display shows "<x> rbtn".
Note, pressing Spacebar does not uncheck the Radio button, as expected since this is a Radio button group. Pressing Down Arrow moves focus to the "End checkbox not checked" and pressing Spacebar will Check the End Radio button. However, pressing Up/Down Arrow keys does not rotate between the Radio buttons and JAWS does not announce how many Radio buttons are in the group.

6. Press Tab key again.
JAWS says "Result checkbox not checked", and the braille display shows "< > Result".

7. Press Tab key again.
JAWS says "Disabled checkbox not checked", and the braille display shows "< > Disabled".

8. Press Tab key again.
JAWS says "Angular checkbox not checked", and the braille display shows "< > Angular".

9. Press Tab key again.
JAWS says "I'm a checkbox checkbox not checked", and the braille display shows "< > I'm a checkbox".

10. Pressing the Spacebar on the "Disabled checkbox" will change the state and JAWS announces "checked". Pressing the Tab key moves to the "I'm a checkbox" element and JAWS announces that it is unavailable (In Firefox and Chrome Tab key skips past this element in the focus order). An inconsistent behaviour.

tinayuangao:
Fixes #3584 #3586

### DIFF
--- a/src/lib/checkbox/checkbox.html
+++ b/src/lib/checkbox/checkbox.html
@@ -1,6 +1,20 @@
 <label [attr.for]="inputId" class="mat-checkbox-layout" #label>
   <div class="mat-checkbox-inner-container"
        [class.mat-checkbox-inner-container-no-side-margin]="!_hasLabel()">
+    <input #input
+           class="mat-checkbox-input cdk-visually-hidden" type="checkbox"
+           [id]="inputId"
+           [required]="required"
+           [checked]="checked"
+           [value]="value"
+           [disabled]="disabled"
+           [name]="name"
+           [tabIndex]="tabIndex"
+           [indeterminate]="indeterminate"
+           [attr.aria-label]="ariaLabel"
+           [attr.aria-labelledby]="ariaLabelledby"
+           (change)="_onInteractionEvent($event)"
+           (click)="_onInputClick($event)">
     <div md-ripple class="mat-checkbox-ripple"
          [mdRippleTrigger]="label"
          [mdRippleDisabled]="_isRippleDisabled()"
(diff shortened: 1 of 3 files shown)


jelbourn:
Can you describe the problem this is fixing? According to [the HTML spec](https://www.w3.org/TR/html5/forms.html#the-label-element), adding `for` should be equivalent to the form control being inside of the `<label>`

tinayuangao:
The problem is JAWS doesn't read the label for radio button or checkbox. Moving `input` element out of label solves this problem. But according to spec, we could use `label` in both ways.

jelbourn:
Can you reproduce the behavior on a simple html page with
```html
<label>
  Active
  <input type="checkbox">
</label>
```

Or does it only happen w/ the material components?

tinayuangao:
It only happen w/ the material components. 

`<md-checkbox>Active</md-checkbox>` doesn't work while
`<md-checkbox>Active<span>more info</span></md-checkbox>` works.

PR #4622, opened 2017-05-17
fix(platform): don't shadow `window`

jelbourn:


mmalerba:
Is this to fix a bug or fix some typescript error? I think `declare ...` just gets stripped out of the JS output.

jelbourn:
It was causing a Googler a problem during a migration due to an extern; I like this approach more anyway, rather than aliases `window` like it was.

mmalerba:
lgtm

jelbourn:
In Google everything is indeed consumed directly from sources

DevVersion:
I see. Then it makes total sense why the issue appears :)

PR #4631, opened 2017-05-17
fix(tabs): allow height 100% in tab content

### LINKED ISSUE #4591: md-tab doen't gives height=100% to give all height of md-tab-group
Hi,

I'm having problems to adjust content of a tab to 100% of height of md-tab-group

For example:

```
<md-tab-group style="background-color: lightblue; height: 100%">// this works all bk is blue
<md-tab label="Info" style="background-color: indianred; height: 100%">//not work
<div style="background-color: yellow; height: 100%"></div>//with % not work, with px works
</md-tab>
</md-tab-group>
```
My problem is that `md-tab-group` expands his `height` to total of `height`. But the atribute `height` of `md-tab` doesn't fit to 100% of `md-tab-group`.

Because of that the content, even if I set the height to 100% it doesn't fit to 100% of `height`

If I set thet `height` attribute of the `<div>` to for example 500px it works correctly

I have this problem since I added `<!doctype html>` to my index.html

Versions: Beta.5
                Angular 4

What would be doing wrong?

andrewseguin:
Fixes #4591

### DIFF
--- a/src/lib/tabs/tab-group.spec.ts
+++ b/src/lib/tabs/tab-group.spec.ts
@@ -1,9 +1,7 @@
-import {
-    async, fakeAsync, tick, ComponentFixture, TestBed
-} from '@angular/core/testing';
-import {MdTabGroup, MdTabsModule, MdTabHeaderPosition} from './index';
-import {Component, ViewChild} from '@angular/core';
-import {NoopAnimationsModule, BrowserAnimationsModule} from '@angular/platform-browser/animations';
+import {async, ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
+import {MdTabGroup, MdTabHeaderPosition, MdTabsModule} from './index';
+import {Component, ElementRef, ViewChild} from '@angular/core';
+import {BrowserAnimationsModule, NoopAnimationsModule} from '@angular/platform-browser/animations';
 import {By} from '@angular/platform-browser';
 import {Observable} from 'rxjs/Observable';
 import {MdTab} from './tab';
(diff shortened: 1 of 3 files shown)


src/lib/tabs/tab-group.spec.ts
```diff
@@ -295,6 +294,26 @@ describe('MdTabGroup', () => {
...
+
+      tabGroup =
+          fixture.debugElement.query(By.directive(MdTabGroup)).componentInstance as MdTabGroup;
+    });
+
+    it('should grow to height of the tab group', () => {
+      const tabContent = fixture.componentInstance.content.nativeElement;
+
+      // Check that the content is greater than 0 (it became larger to match the tab body height)
+      expect(tabContent.getBoundingClientRect().height).toBeGreaterThan(0);
```
crisbeto: Instead of a comment, you could pass the info via `toBeGreaterThan`. That way, if it fails, it'll be a bit easier to see what's going on:

```ts
expect(height).toBeGreaterThan(0, 'Expected the tab content to have a height.');
```

andrewseguin: Hey perfect, good suggestion

andrewseguin:
Looks like mobile safari isnt going to play nice. Will look into it

andrewseguin:
Looks like there's an open bug in Safari that keeps this from passing.

https://bugs.webkit.org/show_bug.cgi?id=137730

andrewseguin:
Unfortunately it looks like I need to close the PR and work on another solution since Safari won't like this.

Stexxen:
Just a note to say that https://bugs.webkit.org/show_bug.cgi?id=137730 is now fixed in Safari 11

Sanafan:
hmm... i'm still not able to put `.mat-tab-body-wrapper` to `height: 100%`.

PR #4634, opened 2017-05-17
fix(button): Add role=button to MdAnchor

### LINKED ISSUE #3703: A11y keyboard focus control in the Button Component
Bug: Buttons provide no notification of state.

Buttons should communicate the action that will occur when the user activates them. In this demo the Button State is not announced, the action is not identified, and some text label are not appropriate.

Currently button states is not announced to the screen reader user.

See below.

Create a good screen reader user experience.

Using screen reader JAWS17 with browsers Firefox, Internet Explorer 11, and Google Chrome, on Windows10.

Screen reader user experience follows:

1. Press Tab key to the Button demo component in the Tab Panel.
JAWS says "FLAT BUTTON button", and the braille display shows "FLAT BUTTON btn".
The screen reader focus moves focus to the first Button element in the Tab Panel. There are two types of Buttons presented in this demo "Normal Buttons" and "Link Buttons".
- In both cases the "FLAT BUTTON" and the "RAISED BUTTON" are in all capital letters. For screen reader pronunciation, only the first letter should be capitalized, and CSS should be used for visual styling.
- The word "button" in the text label, is not required as part of the text label, if the element is identified as a Button Role (HTML Button element or ARIA role=button).
- In both cases, the current State of the Button is not announced by the screen reader. Pressing Spacebar or Enter key appears to do nothing, as the focus does not move and the resulting action is not announced. Although this is a demo, the screen reader behaviour in activating a button should be obvious, and can be accomplished by using ARIA properties.
- In both cases, there are two Buttons with the same text label "check", which makes context difficult to understand for screen readers. Text lables must be unique.

2. Use the Tab key to move focus from one button to the next.
JAWS announces and displays the text label for each button, but the purpose is not identified.
- Buttons can be grouped by a Toolbar role=toolbar, or in a Menu role=menu, to identify their relationship. Although this is a demo, Buttons need to be identified in context.
- Link elements that enclose a image Button should be identified as buttons role=button, and the Image could have an Alt text description if there is no visible text label. Note, using Link text, Alt text, and image text, on a single Button, will result in multiple announcements for a single button.
- Note, the types of actions performed by buttons are distinctly different from the function of a link, and is important that both the appearance and role of a widget match the function it provides. Nevertheless, elements occasionally have the visual style of a link but perform the action of a button. In such cases, giving the element role button helps assistive technology users understand the function of the element. However, a better solution is to adjust the visual design so it matches the function and ARIA role. 

See:
The Roles Model | Accessible Rich Internet Applications (WAI-ARIA) 1.0
http://w3c.github.io/aria-practices/#grid

tinayuangao:
Fixes #3703

### DIFF
--- a/src/demo-app/button/button-demo.html
+++ b/src/demo-app/button/button-demo.html
@@ -11,8 +11,8 @@
   </section>
 
   <section>
-    <a href="http://www.google.com" md-button color="primary">SEARCH</a>
-    <a href="http://www.google.com" md-raised-button>SEARCH</a>
+    <a href="http://www.google.com" md-button color="primary">search</a>
+    <a href="http://www.google.com" md-raised-button>search</a>
     <a href="http://www.google.com" md-fab>
       <md-icon class="md-24">check</md-icon>
     </a>
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -188,6 +188,7 @@ export class MdButton extends _MdButtonMixinBase implements OnDestroy, CanDisabl
     '[attr.disabled]': 'disabled || null',
     '[attr.aria-disabled]': '_isAriaDisabled',
     '(click)': '_haltDisabledEvents($event)',
+    'role': 'button'
   },
   inputs: ['disabled'],
   templateUrl: 'button.html',


jelbourn:
This isn't right; an anchor is a link, not a button. It should be totally accessible by virtue of being a native anchor element. Is there a specific problem you're seeing?

tinayuangao:
No special issue. I'll revert the change.

From #3703: 
> Link elements that enclose a image Button should be identified as buttons role=button
I'm not sure about change the role to button is the correct behavior

PR #4638, opened 2017-05-18
fix(block-scroll-strategy): collapsing root node when enabled

crisbeto:
* Prevents the `html` tag from collapsing when the `BlockScrollStrategy` is enabled.
* Fixes the scroll strategy not being disabled properly in tests.

Fixes https://github.com/angular/material2/issues/4646.

### DIFF
--- a/src/lib/core/overlay/_overlay.scss
+++ b/src/lib/core/overlay/_overlay.scss
@@ -72,8 +72,10 @@
   .cdk-global-scrollblock {
     position: fixed;
 
-    // Necessary for iOS not to expand past the viewport.
-    max-width: 100vw;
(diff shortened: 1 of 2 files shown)


src/lib/core/overlay/_overlay.scss
```diff
@@ -72,8 +72,8 @@
   .cdk-global-scrollblock {
     position: fixed;
 
-    // Necessary for iOS not to expand past the viewport.
-    max-width: 100vw;
+    // Necessary for the content not to lose its width.
+    width: 100%;
```
jelbourn: Add comment explaining why we use `100%` instead of `100vh`

PR #4649, opened 2017-05-18
fix(datepicker): minValidator & maxValidation false errors

### LINKED ISSUE #4725: Datepicker minDate - error is raised when date is exactly min date

Bug
When minDate is set for the datePicker, a mdDatepickerMin error should be raised only if the date is earlier than the minDate.

For the moment, when setting minDate, the date for exactly min date is NOT unselectable, which means you can select it, but it nevertheless raises an error. I think it should be either unselectable and raise an error or selectable and not raise the error.

Set up a minDate restiriction for the datepicker and try to select the exact min date, and check the fact that it raises an error while it is still selectable.

Angular 4, Material 2

PhilipSultanescu:
* Min and max dates give false validation errors when selected date is one of the two
* Include unit-tests to catch min/max validation errors

Fixes #4725


mmalerba:
LGTM, thanks for taking care of this

PR #4658, opened 2017-05-18
fix(select): deselect old options when programmatically setting value

### LINKED ISSUE #4476: Select option styling bug
Bug

Select component: setting an option programmatically should clear any existing previous selected item styles.

The selection is changed, but the previously selected item has still the selected style

In the Plunker: 
- Open the select, notice the styling of "tacos", don't change anything, close the select
- click the reset button
- Open the select again: "pizza" and "tacos" have both the selected styling (mat-selected)

http://plnkr.co/edit/o866DpIMsaYVmPPjwNRf?p=preview

consisten look

tested it with Chrome 54

willshowell:
Fixes #4476

This may be naive regarding selection change events, but existing tests pass and it fixes the issue.

cc @crisbeto

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -321,12 +321,12 @@ describe('MdSelect', () => {
 
       options =
         overlayContainerElement.querySelectorAll('md-option') as NodeListOf<HTMLElement>;
-      expect(options[0].classList).not.toContain('mat-selected');
-      expect(options[1].classList).toContain('mat-selected');
+      expect(options[0].classList).not.toContain('mat-selected', 'Expected first option to no longer be selected');
+      expect(options[1].classList).toContain('mat-selected', 'Expected second option to be selected');
 
       const optionInstances = fixture.componentInstance.options.toArray();
-      expect(optionInstances[0].selected).toBe(false);
-      expect(optionInstances[1].selected).toBe(true);
+      expect(optionInstances[0].selected).toBe(false, 'Expected first option to no longer be selected');
+      expect(optionInstances[1].selected).toBe(true, 'Expected second option to be selected');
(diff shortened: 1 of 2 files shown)


crisbeto:
Off the top of my head, I can't see anything that it might break, but let me play around with it a bit more tomorrow.

src/lib/select/select.ts
```diff
@@ -544,12 +544,12 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
       throw getMdSelectNonArrayValueError();
     }
 
+    this._clearSelection();
```
crisbeto: nit: Can you add a new line after this one?

src/lib/select/select.spec.ts
```diff
@@ -300,6 +300,35 @@ describe('MdSelect', () => {
...
+
+      control.setValue(foods[1].value);
+      fixture.detectChanges();
+
+      trigger.click();
+      fixture.detectChanges();
+
+      options =
+        overlayContainerElement.querySelectorAll('md-option') as NodeListOf<HTMLElement>;
+      expect(options[0].classList).not.toContain('mat-selected');
```
crisbeto: Could you add a description to the expectations in the test? It makes it easier to parse if one of them fails. E.g. `expect(options[0].classList).not.toContain('mat-selected', 'Expected first option not to be selected');`.

willshowell:
@crisbeto changes have been made

crisbeto:
Seems like there are a couple of linter errors. The Closure failure should be fixed by https://github.com/angular/material2/pull/4709.

crisbeto:
@willshowell the two PRs got merged. Can you rebase from master?

PR #4662, opened 2017-05-18
test: fix infinite loop in tests

### LINKED ISSUE #4654: Gulp test tasks are failing
Bug

For the gulp task 'test' and 'test:single-run' to complete

test task fails and crashes Chrome/Firefox.

Run 'gulp test:single-run'

Win 10, Node 6.10.2, Karma 1.7.0, Chrome 58.0.3029.110

I had to exclude the snack-bar and dialog tests to get the test to complete without hanging.

Also noticed some overlays are not getting cleaned up.  When running the tests in debug mode in the browser, after the browser locked up, I can see 12 divs with the cdk-overlay-container class still in the body.

Also seeing several errors in the terminal like:

× should set the width of the overlay if there is no placeholder
        Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

Which I'm guessing is related to the whenStable() method.

crisbeto:
Fixes some unit tests going into an infinite loop by using the `jasmine.any` matcher. Not exactly sure what is going on, but it was most likely introduced in Jasmine 2.6.2. For now I've replaced those matches with `instanceof` checks which accomplish the same.

Fixes #4654.

### DIFF
--- a/src/lib/core/overlay/overlay-directives.spec.ts
+++ b/src/lib/core/overlay/overlay-directives.spec.ts
@@ -77,7 +77,7 @@ describe('Overlay directives', () => {
 
     let strategy =
         <ConnectedPositionStrategy> overlayDirective.overlayRef.getState().positionStrategy;
-    expect(strategy).toEqual(jasmine.any(ConnectedPositionStrategy));
+    expect(strategy instanceof ConnectedPositionStrategy).toBe(true);
 
     let positions = strategy.positions;
     expect(positions.length).toBeGreaterThan(0);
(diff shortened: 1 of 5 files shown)


RobJacobs:
@crisbeto Thank you.

RobJacobs:
@crisbeto Curious as to why the Travis CI build didn't fail?  Looking at the Travis logs on some of the recent commits it doesn't list a job with MODE=test like you see with MODE=e2e?

crisbeto:
My guess is that it had a slightly older version of Jasmine, which wasn't problematic, in the cache. At the time I submitted the PR, Jasmine 2.6.2 had been released for about 20 hours.

DevVersion:
Nice 😄 I was wondering why I get out-of-memory issues on Chrome.

PR #4665, opened 2017-05-19
fix(icon): remove automatic aria labelling and add a11y guidance

jelbourn:
The automatic application of aria-label for md-icon, in hindsight, was
not a great idea. Neither the ligature string nor the SVG filename are
likely to be meaningful descriptions. Even in cases where they *are*,
the icon may be used in such a way that the application of a label is
still the wrong thing. On top of that, adding `role="img"` is usually
not the right approach for an icon, as that role typically refers to
*content* images, rather than icons that are part of the UI.

Ultimately, it is up to the application developer to add the appropriate
meaning for icons based on how they're used. To this end, we add
guidance to the documentation for what to do in different situations.

### DIFF
--- a/src/lib/icon/icon.md
+++ b/src/lib/icon/icon.md
@@ -15,8 +15,8 @@ Some fonts are designed to show icons by using
 "home" as a home image. To use a ligature icon, put its text in the content of the `md-icon`
 component.
 
-By default the
-[Material icons font](http://google.github.io/material-design-icons/#icon-font-for-the-web) is used.
+By default, `<md-icon>` expects the
+[Material icons font](http://google.github.io/material-design-icons/#icon-font-for-the-web).
 (You will still need to include the HTML to load the font and its CSS, as described in the link).
 You can specify a different font by setting the `fontSet` input to either the CSS class to apply to
 use the desired font, or to an alias previously registered with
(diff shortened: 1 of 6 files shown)


src/lib/icon/icon.md
```diff
@@ -80,9 +80,32 @@ match the current theme's colors using the `color` attribute. This can be change
...
-* The `alt` attribute
-* The `fontIcon` input
-* The name of the icon from the `svgIcon` input (not including any namespace)
-* The text content of the component (for ligature icons)
+Similar to an `<img>` element, an icon alone does not convey any useful information for a
+screen-reader user. The user of `<md-icon>` must provide additional information pertaining to how
+the icon is used.
+
+In thinking about accessibility, it is useful to place icon use into one of three categories:
+1. **Decorative**: the icon conveys no real semantic meaning and is purely cosmetic.
```
mmalerba: since all 3 of these cases ask the user to add `aria-hidden="true"`, can we just do that for them?

src/lib/icon/icon.md
```diff
@@ -80,9 +80,32 @@ match the current theme's colors using the `color` attribute. This can be change
...
+
+#### Interactive icons
+Icons alone are not interactive elements for screen-reader users; when the user would interact with
+some icon on the page, a more appropriate  element should "own" the interaction:
+* The `<md-icon>` element should be a child of a `<button>` or `<a>` element.
+* The `<md-icon>` element should be marked with `aria-hidden="true"`.
+* The parent `<button>` or `<a>` should either have a meaningful label provided either through
+direct text content, `aria-label`, or `aria-labelledby`.
+
+#### Indicator icons
```
mmalerba: to make this easier we could add a simple component like this (it could also automatically add a tooltip):

```html
<md-indicator [message]="msg">
  <md-icon>error</md-icon>
</md-indicator>
```

src/lib/icon/icon.md
jelbourn: I'm hesitant to always add `aria-hidden`, because the failure mode is that something that _should_ be visible to screen reader users isn't

src/lib/icon/icon.md
jelbourn: Agreed that this would be useful. Out of scope for this PR, but maybe add eventually.

mmalerba: Ok, created an issue to track

src/lib/icon/icon.md
mmalerba: I think we can check for a user-added `aria-label`, `aria-labeledby`, or `role` attribute and in those cases we wouldn't set aria-hidden, but if they don't provide any useful description anyways I think its safe to leave it hidden

jelbourn: Changed to mark as `aria-hidden` by default based on feedback from Rob Dodson

andrewseguin:
Needs rebase

PR #4686, opened 2017-05-19
fix(datepicker): make datepicker work in compatibility mode

### LINKED ISSUE #4619: Datepicker doesn't work with NoConflictStyleCompatibilityMode
Bug

Use `mat-datepicker` without any problem (with `NoConflictStyleCompatibilityMode`) inside angularjs-angular hybrid apps.

It now throws error in the console (right after I click on the calendar icon to open datepicker):
> ERROR Error: The "md-" prefix cannot be used in ng-material v1 compatibility mode. It was used on an "button" element.

I created a github repo for demoing the issue: https://github.com/merobal/datepicker-demo
Demo page: https://merobal.github.io/datepicker-demo/

We would like to use datepicker inside hybrid apps which also uses material1.

Latest angular, latest material, Win10, TS 2.2, all browsers.

mmalerba:
fixes #4619

### DIFF
--- a/src/lib/datepicker/calendar.html
+++ b/src/lib/datepicker/calendar.html
@@ -1,16 +1,38 @@
 <div class="mat-calendar-header">
   <div class="mat-calendar-controls">
-    <button md-button class="mat-calendar-period-button" (click)="_currentPeriodClicked()"
-            [attr.aria-label]="_periodButtonLabel">
+
+    <!-- Check for compatibility mode and use correct prefix for md-button. -->
+    <button *ngIf="!_isCompatibilityMode" md-button class="mat-calendar-period-button"
+            (click)="_currentPeriodClicked()"  [attr.aria-label]="_periodButtonLabel">
       {{_periodButtonText}}
       <div class="mat-calendar-arrow" [class.mat-calendar-invert]="!_monthView"></div>
     </button>
-    <div class="mat-calendar-spacer"></div>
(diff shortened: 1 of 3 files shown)


src/lib/datepicker/calendar.html
```diff
@@ -1,16 +1,38 @@
 <div class="mat-calendar-header">
   <div class="mat-calendar-controls">
-    <button md-button class="mat-calendar-period-button" (click)="_currentPeriodClicked()"
-            [attr.aria-label]="_periodButtonLabel">
+
```
jelbourn: Add a TODO like
```html
<!-- TODO: stop doing this horrible hack for compatibility mode -->
```
?

src/lib/datepicker/calendar.ts
```diff
@@ -111,6 +112,7 @@ export class MdCalendar<D> implements AfterContentInit {
   constructor(private _elementRef: ElementRef,
               private _intl: MdDatepickerIntl,
               private _ngZone: NgZone,
+              @Optional() @Inject(MATERIAL_COMPATIBILITY_MODE) public _isCompatibilityMode: boolean,
```
princemaple: private?

mmalerba: can't be private since it's used in the template, underscore indicates that its for internal use only

princemaple:
Right! Sorry!

PR #4688, opened 2017-05-19
fix(tooltip): improve accessibility

andrewseguin:
- Show on trigger focus
- Hide on trigger blur
- Hide on escape when trigger has focus
- Add aria tooltip role for tooltip
- Add aria describedby role for trigger

### DIFF
--- a/src/lib/tooltip/tooltip.html
+++ b/src/lib/tooltip/tooltip.html
@@ -1,4 +1,6 @@
 <div class="mat-tooltip"
+     role="tooltip"
+     [id]="id"
      [style.transform-origin]="_transformOrigin"
      [@state]="_visibility"
      (@state.done)="_afterVisibilityAnimation($event)">
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -371,7 +371,7 @@ describe('MdTooltip', () => {
     it('should associate trigger and tooltip through aria-describedby', fakeAsync(() => {
       // Expect that the trigger does not have an associated tooltip since there is no tooltip shown
       const trigger = fixture.componentInstance.trigger.nativeElement;
-      expect(trigger.getAttribute('aria-describedBy')).toBe('');
+      expect(trigger.getAttribute('aria-describedBy')).toBe(null);
 
       tooltipDirective.show(0);
       tick(0); // Tick for the show delay
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -27,7 +27,7 @@ import {dispatchFakeEvent} from '../core/testing/dispatch-events';
 
 const initialTooltipMessage = 'initial tooltip message';
 
-describe('MdTooltip', () => {
+fdescribe('MdTooltip', () => {
   let overlayContainerElement: HTMLElement;
   let dir: {value: LayoutDirection};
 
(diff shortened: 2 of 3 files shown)


andrewseguin:
Looking into the test failures, not sure why the instance isnt being removed right.

src/lib/tooltip/tooltip.ts
```diff
@@ -58,7 +58,11 @@ export function throwMdTooltipInvalidPositionError(position: string) {
   selector: '[md-tooltip], [mdTooltip], [mat-tooltip], [matTooltip]',
   host: {
     '(longpress)': 'show()',
+    '(focus)': 'show()',
+    '(blur)': 'hide(0)',
+    '(keydown.esc)': 'hide(0)',
```
jelbourn: Can't use `keydown.xxx` syntax because it's not supported in Safari 9 (Angular uses `KeyboardEvent.key` for this). You'll need to just handle `keydown` and check the keycode. 

You'll also need to check whether the tooltip is actually present, and call `stopPropagation` if it. This is necessary in cases where you have something like a tooltip inside of a dialog; you only want to dismiss the tooltip without closing the dialog.

src/lib/tooltip/tooltip.ts
```diff
@@ -58,7 +58,11 @@ export function throwMdTooltipInvalidPositionError(position: string) {
...
     '(touchend)': 'hide(' + TOUCHEND_HIDE_DELAY + ')',
+    '[attr.aria-describedby]': '_getTooltipId()'
```
jelbourn: This needs to avoid clobbering an existing `aria-describedby`

src/lib/tooltip/tooltip.ts
```diff
@@ -199,11 +203,36 @@ export class MdTooltip implements OnDestroy {
...
+    // intercept the event and hide the tooltip.
+    if (this._tooltipInstance && this._tooltipInstance.isVisible() && e.keyCode === ESCAPE) {
+      e.stopPropagation();
+      this.hide(0);
+    }
+  }
+
+  /** Returns the trigger's aria-describedby attribute. */
+  private _getAriaDescribedby(): string {
+    return this._elementRef.nativeElement.getAttribute('aria-describedby');
```
crisbeto: This one is technically safe from throwing an error in Universal, because it's after a user interaction, but it might be better if we used an `@Input` anyway.

src/lib/tooltip/tooltip.spec.ts
```diff
@@ -1,21 +1,22 @@
 import {
   async,
   ComponentFixture,
-  TestBed,
-  tick,
   fakeAsync,
-  flushMicrotasks
+  flushMicrotasks,
+  TestBed,
+  tick
```
crisbeto: Needs a trailing comma here and on line 14.

src/lib/tooltip/tooltip.html
```diff
@@ -1,4 +1,6 @@
 <div class="mat-tooltip"
+     role="tooltip"
```
crisbeto: [Looking at this example of an accessible tooltip](http://accessibility.athena-ict.com/aria/examples/tooltip.shtml), it looks like we need to toggle `aria-hidden`, in addition to the `role` and `aria-describedby`.

src/lib/tooltip/tooltip.ts
```diff
@@ -199,11 +203,36 @@ export class MdTooltip implements OnDestroy {
...
+  }
+
+  /** Sets the trigger's aria-describedby attribute. */
+  private _setAriaDescribedBy(ariaDescribedby: string) {
```
crisbeto: This one could be replaced with an attribute binding.

src/lib/tooltip/tooltip.ts
andrewseguin: I agree except I think we need would need this anyways since we're checking if the current tooltip matches the aria-describedby and an `Input@` won't capture our setter's interaction.

src/lib/tooltip/tooltip.ts
andrewseguin: I had that at first but we saw that it would clobber any existing `aria-describedby` attribute that the user may have set. We want to only apply the tooltip's id as `aria-describedby` if there isn't already one set.

PR #4694, opened 2017-05-20
refactor(button): use color binding for fab button color

DevVersion:
No longer creates extra CSS selectors to set the default color for round buttons to `accent`.

@tinayuangao @jelbourn This needs some confirmation in regards to the `_hasAttributeWithPrefix` method. The `isBrowser` check in that method doesn't seem to be a good solution for Universal?

<details>
<summary>Another possible approach</summary>

To be sure that this works in Univeral we can also do it the following way:

image

</details>

### DIFF
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -6,7 +6,9 @@ import {
   HostBinding,
   Input,
   OnDestroy,
+  Optional,
   Renderer2,
+  Self,
   ViewEncapsulation
 } from '@angular/core';
 import {coerceBooleanProperty, FocusOriginMonitor, Platform} from '../core';
@@ -14,6 +16,9 @@ import {mixinDisabled, CanDisable} from '../core/common-behaviors/disabled';
 
 
 // TODO(kara): Convert attribute selectors to classes when attr maps become available
+
+/** Default color palette for round buttons (md-fab and md-mini-fab) */
+const DEFAULT_ROUND_BUTTON_COLOR = 'accent';
 
 
 /**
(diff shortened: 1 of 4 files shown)


jelbourn:
@DevVersion change looks good, not sure what you mean by the `_hasAttributeWithPrefix` / `isBrowser` stuff, though. That `isBrowser` check is specific there so that Universal _does_ work.

DevVersion:
@jelbourn I thought that if someone renders a `fab` button on the server the `_hasAttributeWithPrefix` will always return `false` (due to the `isBrowser` check) and therefore the fab will not get the `accent` color by default.

jelbourn:
Oh yeah, that would be a problem. When I added the `isBrowser` check, the only thing that this was used for only dealt with styles for user interaction.

DevVersion:
Yeah with that it was totally reasonable to just skip in non-browser environments.

What do you think about the other approach I adddd to the PR description?

jelbourn:
I think that alternate approach would be necessary

DevVersion:
@jelbourn Done. This should be the "Angular"-way of ensuring that it works with Universal.

src/lib/button/button.ts
```diff
@@ -57,17 +62,28 @@ export class MdIconButtonCssMatStyler {}
   selector: 'button[md-fab], button[mat-fab], a[md-fab], a[mat-fab]',
   host: {'class': 'mat-fab'}
 })
-export class MdFabCssMatStyler {}
+export class MdFab {
+  constructor(@Self() @Optional() button: MdButton, @Self() @Optional() anchor: MdAnchor) {
```
andrewseguin: Here and in MdMiniFab, looks like we want to use `forwardRef` since those components are defined further down the file. This caused problems internally, surprised our tests didn't have trouble with them.

DevVersion:
@andrewseguin Done.

PR #4696, opened 2017-05-20
feat(datepicker): popup positioning improvements

### LINKED ISSUE #4406: datepicker: automatic positioning
image

it should open up when there's not enough space at the bottom, otherwise it's gonna be completely unaccessible

crisbeto:
* Adds fallback positions to the datepicker popup, allowing it to reposition in the cases where it would've gone outside the viewport.
* Fixes an issue that prevented the datepicker from the being positioned properly on the first open due to the calendar not being rendered yet.
* Adds a scroll strategy to reposition the datepicker when the user scrolls away.
* Switches to using `ngSwitch` instead of `ngIf` when determining which view to render.

Fixes #4406.

### DIFF
--- a/src/lib/datepicker/calendar.html
+++ b/src/lib/datepicker/calendar.html
@@ -16,9 +16,9 @@
 </div>
 
 <div class="mat-calendar-content" (keydown)="_handleCalendarBodyKeydown($event)"
-     cdkMonitorSubtreeFocus>
+    [ngSwitch]="_monthView" cdkMonitorSubtreeFocus>
   <md-month-view
-      *ngIf="_monthView"
+      *ngSwitchCase="true"
       [activeDate]="_activeDate"
       [selected]="selected"
       [dateFilter]="_dateFilterForViews"
@@ -26,7 +26,7 @@
   </md-month-view>
 
   <md-year-view
-      *ngIf="!_monthView"
+      *ngSwitchDefault
       [activeDate]="_activeDate"
       [selected]="selected"
       [dateFilter]="_dateFilterForViews"
--- a/src/lib/datepicker/datepicker.spec.ts
+++ b/src/lib/datepicker/datepicker.spec.ts
@@ -509,6 +509,82 @@ describe('MdDatepicker', () => {
           .toThrowError(/MdDatepicker: No provider found for .*/);
     });
   });
+
+  describe('popup positioning', () => {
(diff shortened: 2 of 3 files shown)


src/lib/datepicker/datepicker.spec.ts
```diff
@@ -509,6 +509,82 @@ describe('MdDatepicker', () => {
...
+      TestBed.configureTestingModule({
+        imports: [MdDatepickerModule, MdInputModule, MdNativeDateModule, NoopAnimationsModule],
+        declarations: [StandardDatepicker],
+      }).compileComponents();
+
+      fixture = TestBed.createComponent(StandardDatepicker);
+      fixture.detectChanges();
+      testComponent = fixture.componentInstance;
+      input = fixture.debugElement.query(By.css('input')).nativeElement;
+      input.style.position = 'fixed';
```
mmalerba: nifty way of testing this, I like it :)

src/lib/datepicker/datepicker.ts
```diff
@@ -22,16 +23,19 @@ import {MdDialogRef} from '../dialog/dialog-ref';
 import {PositionStrategy} from '../core/overlay/position/position-strategy';
 import {
   OriginConnectionPosition,
-  OverlayConnectionPosition
-} from '../core/overlay/position/connected-position';
+  OverlayConnectionPosition,
+  RepositionScrollStrategy,
+  BlockScrollStrategy,
```
mmalerba: looks like `BlockScrollStrategy` isn't needed, also it seems like the reposition isn't working? (if I give the last input on the datepicker demo a huge top margin, it still seems to block scroll)

crisbeto: The demo page actually ends up blocking scroll on whenever a connected overlay is open, because the scrollable container is the `md-sidenav-container`.

src/lib/datepicker/calendar.html
```diff
@@ -16,17 +16,17 @@
 </div>
 
 <div class="mat-calendar-content" (keydown)="_handleCalendarBodyKeydown($event)"
-     cdkMonitorSubtreeFocus>
+    [ngSwitch]="_monthView" cdkMonitorSubtreeFocus>
```
crisbeto: It's just one less watcher in the template.

crisbeto:
Removed the unused imports @mmalerba.

PR #4699, opened 2017-05-20
fix(icon): handle icons as <symbol> nodes

### LINKED ISSUE #4680: md-icon used with addSvgIconSetInNamespace and <symbol> are invisible
Md-icons loaded using `addSvgIconSetInNamespace` appear in the DOM but are not visible. I suspect this is because the `<symbol>` tags are included, which will not be rendered without the `<use>` tag. 

> a symbol element itself is not rendered. Only instances of a symbol element (i.e., a reference to a symbol by a <use> element) are rendered.
 https://developer.mozilla.org/en-US/docs/Web/SVG/Element/symbol

I made a plunkr to demonstrate the issue: https://plnkr.co/edit/KYyfI0uBq2GYM2ElC1k3?p=preview

crisbeto:
Fixes icons not being rendered if they're defined as a `<symbol>` inside the source file.

Fixes #4680.

### DIFF
--- a/src/lib/icon/icon-registry.ts
+++ b/src/lib/icon/icon-registry.ts
@@ -378,8 +378,10 @@ export class MdIconRegistry {
   private _toSvgElement(element: Element): SVGElement {
     let svg = this._svgElementFromString('<svg></svg>');
 
-    for (let i = 0; i < element.children.length; i++) {
-      svg.appendChild(element.children[i].cloneNode(true));
+    for (let i = 0; i < element.childNodes.length; i++) {
+      if (element.childNodes[i].nodeType === Node.ELEMENT_NODE) {
+        svg.appendChild(element.childNodes[i].cloneNode(true));
+      }
     }
 
     return svg;
--- a/src/lib/icon/icon-registry.ts
+++ b/src/lib/icon/icon-registry.ts
@@ -327,14 +327,24 @@ export class MdIconRegistry {
    */
   private _extractSvgIconFromSet(iconSet: SVGElement, iconName: string): SVGElement {
     const iconNode = iconSet.querySelector('#' + iconName);
+
     if (!iconNode) {
       return null;
     }
+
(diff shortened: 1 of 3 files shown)


src/lib/icon/icon-registry.ts
```diff
@@ -327,14 +327,24 @@ export class MdIconRegistry {
...
     }
+
     // If the icon node is itself an <svg> node, clone and return it directly. If not, set it as
     // the content of a new <svg> node.
-    if (iconNode.tagName.toLowerCase() == 'svg') {
+    if (iconNode.tagName.toLowerCase() === 'svg') {
       return this._setSvgAttributes(iconNode.cloneNode(true) as SVGElement);
     }
+
+    // If the node is a <symbol>, it won't be rendered so we have to convert it into <svg>. Note
```
jelbourn: I'm not sure we want this behavior. If an SVG has a `<symbol>` element, we should interpret it as not rendering out _on purpose_, where the `<use>` element should be used to render something. We shouldn't be making assumptions about _why_ someone is using `<symbol>` in their SVG.

crisbeto: We're not doing this for all `symbol` elements, but rather for the one that the user is referring to via an id. I think it's safe to assume that if there's a `<symbol id="something">` and the user is referring to it via `iconSet:something`, it would be expected that the symbol would be rendered.

jelbourn: That's fair, I suppose. Are you able to reference `<symbol>` elements in `<use>` elements in a completely different `<svg>` parent? I'm wondering if it would make sense to just use `<use>` for this, though it might be overkill.

crisbeto: I initially went for the `use` element, but the problem is that, if your page has a `base` tag (which all Angular apps have), Firefox requires you to include the current URL before the hash. E.g. if I was using an icon on `/about`, the `use` element would be `<use href="/about#icon-id"/>`.  The href also would have to be maintained on route changes.

PR #4703, opened 2017-05-21
fix(autocomplete): handle escape key

crisbeto:
fix(autocomplete): handle escape key

* Closes the autocomplete panel when pressing escape.
* Switches the autocomplete unit tests to using the common utility for creating fake keyboard events.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -19,7 +19,7 @@ import {PositionStrategy} from '../core/overlay/position/position-strategy';
 import {ConnectedPositionStrategy} from '../core/overlay/position/connected-position-strategy';
 import {Observable} from 'rxjs/Observable';
 import {MdOptionSelectionChange, MdOption} from '../core/option/option';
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -231,7 +231,9 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   }
 
   _handleKeydown(event: KeyboardEvent): void {
-    if (this.activeOption && event.keyCode === ENTER) {
+    if (event.keyCode === ESCAPE) {
+      this.panelOpen ? this.closePanel() : this._element.nativeElement.blur();
```
jelbourn: Where does this `blur` put the focus? If it dumps it back to the `body`, then we don't want this behavior

kara: I tested that manually, and it seems to go to the next in the tab order on TAB. So I think it's okay.

kara: @crisbeto  Seems like a noop might be a better choice if the panel is closed.

crisbeto:
Removed the `blur` @kara.

PR #4705, opened 2017-05-21
refactor: switch e2e tests to async/await

crisbeto:
Switches to using async/await in the e2e tests. This should make it a bit easier to deal with Protractor's async APIs.

**Note:** The CI failures are unrelated to the changes in this PR and are the same as the ones from #4702 and #4703.


e2e/components/block-scroll-strategy/block-scroll-strategy.e2e.ts
```diff
@@ -1,13 +1,13 @@
...
-import {getScrollPosition} from '../../util/query';
+import {getScrollPosition, asyncSpec} from '../../util/index';
 
 
 describe('scroll blocking', () => {
   beforeEach(() => browser.get('/block-scroll-strategy'));
   afterEach(() => clickOn('disable'));
 
-  it('should not be able to scroll programmatically along the x axis', async (done) => {
+  it('should not be able to scroll programmatically along the x axis', asyncSpec(async () => {
```
jelbourn: According to @vikerman this `asyncSpec` wrapper should be unnecessary w/ jasminewd2

crisbeto: Hmm looks like it. I was initially playing around with passing in the `done` function and calling it at the end, which ended up timing out for failures since I wasn't wrapping it in a try/catch and calling `done.fail`. I guess that was overriding the custom handling from `jasminewd`. I've removed the `asyncSpec` now and it behaves as expected.

jelbourn:
LGTM for the code changes, but what's up w/ the screenshot diffs here?
cc @tinayuangao

jelbourn:
Poke @tinayuangao on the screenshot stuff; not clear to me what's happening on the results page

tinayuangao:
LGTM.

The screenshot tests failed because we changed the test name, so the test images are unavailable to old test names.
After we merge this PR & update goldens, the images & tests should be fine.

PR #4707, opened 2017-05-21
build: upgrade dev dependencies

DevVersion:
* Upgrades outdated development dependencies to the latest versions.
---
* Skipped upgrading Autoprefixer due to a breaking change with the Browserlist.
* Skipped upgrading UglifyJS due to unresolved issues in UglifyJS3
* Skipped upgrading google-closure-compiler due to alexeagle/closure-compiler-angular-bundling#30
* Skipped upgrading jasmine types because they now enforce types in `expect` calls (which is nice) but for Protractor the `expect` automatically unwraps promises (See angular/protractor#4176)

**Note**: Tests are failing due to infinite loops in the latest Jasmine version. Rebasing after https://github.com/angular/material2/pull/4662 is merged!


jelbourn:
Don't we not use autoprefixer any more?

For the others, can you add a comment in the `package.json` wherever we've fixed the version with a link to any relevant issues?

DevVersion:
We still use Autoprefixer to figure out if a property needs to have a prefix. @crisbeto implemented it in the stylelint rule.

Also I'm not sure if comments work inside of `package.json` files. They are already very sensitive about commas.

PR #4719, opened 2017-05-22
build: move packaging in separate folder

DevVersion:
Moves all related packaging code into a `packaging` folder inside of gulp.

At some point we can move the folder completely to `tools/` and can share it with other repositories like `flex-layout`.


tools/gulp/packaging/build-utils.ts
```diff
@@ -0,0 +1,85 @@
+import {writeFileSync, readFileSync} from 'fs-extra';
+import {join} from 'path';
+import {LICENSE_BANNER, MATERIAL_VERSION} from '../constants';
```
jelbourn: I really dislike grab-bag "util" files. Can we break this up into individual files?

tools/gulp/packaging/build-functions.ts
```diff
@@ -0,0 +1,86 @@
...
+  copyFiles(sourcePath, 'package.json', releasePath);
+
+  updatePackageVersion(releasePath);
+  createTypingFile(releasePath, packageName);
+  createMetadataFile(releasePath, packageName);
+  addPureAnnotationsToFile(join(releasePath, '@angular', `${packageName}.es5.js`));
+}
+
+/** Builds the bundles for the specified package. */
+export async function buildPackageBundles(entryFile: string, packageName: string) {
```
jelbourn: Maybe put `composeRelease` and `buildPackageBundles` into their own files?

tools/gulp/packaging/inline-resources.ts
```diff
@@ -32,3 +34,24 @@ export function inlineMetadataResources(metadata: any, componentResources: Map<s
     }
   }
 }
+
+
+/** Inlines HTML and CSS resources into `metadata.json` files. */
+export function inlinePackageMetadataFiles(packagePath: string) {
```
jelbourn: I'd prefer this to stay in its own file

tools/gulp/packaging/ts-compiler.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  fs.writeFileSync(outputPath, transpiled.outputText);
+
+  if (transpiled.sourceMapText) {
+    fs.writeFileSync(`${outputPath}.map`, transpiled.sourceMapText);
+  }
+}
+
+/** Formats the TypeScript diagnostics into a error string. */
+export function formatDiagnostics(diagnostics: ts.Diagnostic[], baseDir: string): string {
```
jelbourn: Separate transpile and diagnostics into separate files?

DevVersion: Ugh this ends up in a lot of files. If we do this then we should have a folder `packaging/typescript`?

tools/gulp/packaging/build-utils.ts
DevVersion: Not sure I'm a bit afraid of having 5 files just for small functions. What do you think?

tools/gulp/packaging/build-functions.ts
DevVersion: I like that idea. Do you have any good ideas for the file names?

- `build-release.ts`
- `build-normal.ts`

Something like that?

tools/gulp/packaging/build-utils.ts
jelbourn: Small, single-purpose files are just much easier to grok than have a bunch of unrelated stuff in one file.

tools/gulp/packaging/ts-compiler.ts
jelbourn: I don't think it would need a a separate directory- maybe just `transpile.ts` and `typescript-diagnostics.ts`?

tools/gulp/packaging/build-functions.ts
jelbourn: Why not `build-bundles.ts` and `compose-release.ts`?

DevVersion: Sounds good. I like that

tools/gulp/packaging/ts-compiler.ts
DevVersion: Sounds a bit unnecessary to be honest. The diagnostics are actually part of the "custom" ts-compiler.

tools/gulp/packaging/build-utils.ts
DevVersion: Okay I will try to see how I can split them up.

tools/gulp/packaging/ts-compiler.ts
jelbourn: If they're not used outside this file then they can stay there without being exported.

DevVersion: Yeah good point with the `export`.

DevVersion:
@jelbourn Moved the different build utils into separate files. Let me know if you want any further changes!

Issue #4718, opened 2017-05-22
Ability to assign class to dialog


jbojcic1:
Feature request:

When I want to set dialog width from code I see that style is set to cdk-overlay-pane element but if I want to do that from css I don't have a way as I cannot set any class to cdk-overlay-pane.

This is useful when you need responsive widths so that you can write media queries which is imo a lot cleaner way than listening to window resize and changing width from code with updateSize method.

donroyco:
Already mentioned / discussed in #4012

jbojcic1:
@donroyco what's the status of that? I can create a PR for this if it's not already done.

donroyco:
No further development on the issue, as far as I know.

jbojcic1:
Ok. Will create it

Issue #4012, opened 2017-04-10
Support adding CSS class to MdDialog root element


xtianus79:
#### Bug, feature request, or proposal:
Bug / Feature 

#### What is the expected behavior?
Unlike The toast notifications which has a parameter to add a class... There is nothing I can add on the parent container level to change styling of md dialog.  

#### What is the current behavior?
I can just override material styles

#### Which versions of Angular, Material, OS, browsers are affected?
2.0.0-beta.2

#### Is there anything else we should know?
Nope

fxck:
https://github.com/angular/material2/issues/3472

xtianus79:
@fxck and @jelbourn as of right now this is not available? Or it is

jelbourn:
This exists now

shizhenbin:
like  this is okay: 
this.MarketOrderDetailComponent = this.dialog.open(MarketOrderDetailComponent,{height: '100%',width: '100%',panelClass:"detailDialog"});

Issue #4653, opened 2017-05-18
Proposal: Ability to add extra class(es) to dialog container


willshowell:
#### Bug, feature request, or proposal:
Proposal

####  Use Case 1:
I have a dialog that I want to be full-bleed (in my case, it contains just a map)
```
this.dialog.open(MyMapDialogComponent, { containerClass: 'full-bleed' })

.mat-dialog.container.full-bleed {
  padding: 0;
}
```

#### Use Case 2:
I want to make my dialog full screen:
```
this.dialog.open(MyBigDialogComponent, { containerClass: 'full-screen' })

.mat-dialog-container.full-screen {
  max-width: none;
  width: 100vw;
  height: 100vh;
}
```

This wouldn't be too different from `backdropClass` or snackbar's `extraClasses` and it would solve https://github.com/angular/material2/issues/3209

willshowell:
@jelbourn With the addition of `panelClass` in #4722, I'm able to achieve what I want. Happy to  close unless you see any value in also being able to add classes to the dialog container.

jelbourn:
Sounds good to me

zdendapa:
its not perfect, but fast solution
/deep/ .mat-dialog-container {
  padding: 0;
}

no-more:
I know this thread is closed but I was wondering if it's possible to add a class from within the dialog component ?
The panelClass configuration needs to be used everywhere I want to call the dialog, but my need is to apply  the same class all the time so it's quite redundant. I was looking for something like "host" solution for component.

Thanks.

willshowell:
@no-more you might be interested in using `MAT_DIALOG_DEFAULT_OPTIONS` if you want to apply the class to all dialogs, or maybe follow https://github.com/angular/material2/issues/6206 since it would provide the ability to modify the panelClass

no-more:
Thanks for your fast response.
I'll follow the other thread as my need is to be able to apply one or more class to a specific dialog component(all instances of this components) and not to all my dialogs.

PR #4722, opened 2017-05-22
feat(dialog): custom class option #4718 #4012

jbojcic1:
Extend dialog config options to allow custom dialog class. Custom class enables media queries.

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -53,8 +53,8 @@ export class OverlayRef implements PortalHost {
       this._attachBackdrop();
     }
 
-    if (this._state.dialogClass) {
-      this._pane.classList.add(this._state.dialogClass);
+    if (this._state.panelClass) {
+      this._pane.classList.add(this._state.panelClass);
     }
 
     return attachResult;
@@ -73,9 +73,6 @@ export class OverlayRef implements PortalHost {
     this._togglePointerEvents(false);
     this._state.scrollStrategy.disable();
     this._detachments.next();
-    if (this._state.dialogClass) {
-      this._pane.classList.remove(this._state.dialogClass);
-    }
 
     return this._portalHost.detach();
   }
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -53,6 +53,10 @@ export class OverlayRef implements PortalHost {
       this._attachBackdrop();
     }
 
+    if (this._state.dialogClass) {
+      this._pane.classList.add(this._state.dialogClass);
+    }
+
     return attachResult;
   }
 
@@ -69,6 +73,9 @@ export class OverlayRef implements PortalHost {
     this._togglePointerEvents(false);
     this._state.scrollStrategy.disable();
     this._detachments.next();
+    if (this._state.dialogClass) {
+      this._pane.classList.remove(this._state.dialogClass);
+    }
 
     return this._portalHost.detach();
   }
--- a/src/lib/core/overlay/overlay-state.ts
+++ b/src/lib/core/overlay/overlay-state.ts
@@ -16,7 +16,7 @@ export class OverlayState {
   scrollStrategy: ScrollStrategy = new NoopScrollStrategy();
 
   /** Custom class to add to the dialog */
-  dialogClass: string = '';
+  panelClass: string = '';
 
   /** Whether the overlay has a backdrop. */
   hasBackdrop: boolean = false;
--- a/src/lib/core/overlay/overlay-state.ts
+++ b/src/lib/core/overlay/overlay-state.ts
@@ -15,6 +15,9 @@ export class OverlayState {
   /** Strategy to be used when handling scroll events while the overlay is open. */
   scrollStrategy: ScrollStrategy = new NoopScrollStrategy();
 
+  /** Custom class to add to the dialog */
+  dialogClass: string = '';
+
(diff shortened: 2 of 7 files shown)


jbojcic1:
I tried to run unit test but I am getting chrome disconnected error message so it seems something is wrong with karma config. I've tried to increase timeouts but it didn't help. Also e2e test were failing even before my changes so couldn't test that either.

willshowell:
Would it be possible to support the same syntax as ngClass?

e.g. #4629

jbojcic1:
@willshowell that would be great though situation here is more complex as pane element is created dynamically so I'd need to handle all different types of class object to add classes with classList.add

willshowell:
Ah yes I see.

I'm curious why you chose to add the class to the `cdk-overlay-pane`. Since the overlay pane has so few styles, it would seem that attaching the custom class to `mat-dialog-container` would be more useful (though happy to be shown otherwise).

If you instead chose to attach to `mat-dialog-container`, then you could do the following. It wouldn't give you `ngClass` compatibility (though @jelbourn has an open issue https://github.com/angular/angular/issues/7289 for that), but would remove your need to manually set/remove classes from the overlay pane.


```diff
// dialog-container.ts

@Component({
  // ...
  host: {
+   '[class]': 'dialogConfig?.dialogClass',
    '[class.mat-dialog-container]': 'true',
    '[attr.role]': 'dialogConfig?.role',
    '[@slideDialog]': '_state',
    '(@slideDialog.done)': '_onAnimationDone($event)',
  },
})
export class MdDialogContainer extends BasePortalHost {
```

src/lib/core/overlay/overlay-state.ts
```diff
@@ -15,6 +15,9 @@ export class OverlayState {
   /** Strategy to be used when handling scroll events while the overlay is open. */
   scrollStrategy: ScrollStrategy = new NoopScrollStrategy();
 
+  /** Custom class to add to the dialog */
+  dialogClass: string = '';
```
jelbourn: Overlay is the basis for multiple components, not just dialog. I would call this `panelClass`.

src/lib/dialog/dialog-config.ts
```diff
@@ -27,6 +27,9 @@ export class MdDialogConfig {
   /** The ARIA role of the dialog element. */
   role?: DialogRole = 'dialog';
 
+  /** Custom class for the dialog, */
+  dialogClass?: string = '';
```
jelbourn: Probably call this `panelClass` as well (this is already `DialogConfig`, so using the word `dialog` again on a property is redundant)

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -53,6 +53,10 @@ export class OverlayRef implements PortalHost {
       this._attachBackdrop();
     }
 
+    if (this._state.dialogClass) {
```
jelbourn: There needs to be a unit test for this behavior in the overlay tests

src/lib/core/overlay/overlay-ref.ts
```diff
@@ -69,6 +73,9 @@ export class OverlayRef implements PortalHost {
     this._togglePointerEvents(false);
     this._state.scrollStrategy.disable();
     this._detachments.next();
+    if (this._state.dialogClass) {
```
jelbourn: I don't think removing the class is necessary

jbojcic1:
@willshowell I chose cdk-overlay-pane because when you use width dialog option it's set to that element so I guess that's the right one.

willshowell:
@jbojcic1 understood, and I agree that it keeps it consistent. My personal use cases were for removing the dialog padding and overriding the dialog's max-width (see https://github.com/angular/material2/issues/4653). But after playing with it some, I see that you get better position control by having the selector on the `cdk-overlay-pane` 👍

Thanks for putting this PR together and once it's merged, I'll just use `.custom-overlay .mat-dialog-container` to accomplish those overrides.

jbojcic1:
@jelbourn pls check now

src/lib/core/overlay/overlay-state.ts
```diff
@@ -15,6 +15,9 @@ export class OverlayState {
   /** Strategy to be used when handling scroll events while the overlay is open. */
   scrollStrategy: ScrollStrategy = new NoopScrollStrategy();
 
+  /** Custom class to add to the dialog */
```
jelbourn: Comment still says "dialog", should say "overlay pane"

src/lib/core/overlay/overlay.spec.ts
```diff
@@ -336,6 +336,24 @@ describe('Overlay', () => {
 
   });
 
+  describe('panelClass', () => {
+    let config: OverlayState;
+
+    beforeEach(() => {
+      config = new OverlayState();
+      config.panelClass = 'custom-panel-class';
+    });
```
jelbourn: You don't really need the `beforeEach` here; you can create the stats and set the class in the test itself because it's only being used once

src/lib/core/overlay/overlay-state.ts
jbojcic1: Forgot that one :)

jbojcic1:
@jelbourn done

fobaz:
FYI, somebody is very much waiting for this merge.

jonsamwell:
Any update when this will make it into a release?

jbojcic1:
@jonsamwell not sure. Maybe @jelbourn will know?

morkro:
Very much looking forward to this 😬

PR #4726, opened 2017-05-22
feat(snack-bar): allow setting the layout direction

### LINKED ISSUE #4721: Snackbar RTL text support.
No sure.

Allow control over snackbar direction.

Can't control snackbar direction

Any snackbar with RTL text:
image

RTL sites with a snackabr.

angular 4.0 material 2.0.3.beta

Documentation does not tell about a method I should use, maybe this is a documentation issue rather than a missing implementation, not sure.
I'm sure there are workarounds, but I would like not to use them if possible.

crisbeto:
* Allows the consumer to set the layout direction of a snack bar.
* Fixes some bad indentation in the snack bar tests.

Fixes #4721.

### DIFF
--- a/src/lib/snack-bar/simple-snack-bar.scss
+++ b/src/lib/snack-bar/simple-snack-bar.scss
@@ -28,4 +28,9 @@ $mat-snack-bar-button-margin: 48px !default;
     size: inherit;
     weight: 600;
   }
+
+  /deep/ [dir='rtl'] & {
+    margin-right: $mat-snack-bar-button-margin;
+    margin-left: 0;
+  }
 }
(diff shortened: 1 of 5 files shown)


src/lib/snack-bar/simple-snack-bar.scss
```diff
@@ -28,4 +28,9 @@ $mat-snack-bar-button-margin: 48px !default;
     size: inherit;
     weight: 600;
   }
+
+  /deep/ [dir='rtl'] & {
```
jelbourn: I would like to avoid having `/deep/` anywhere in the library. 
(and prefer turning off ViewEncapsulation)

crisbeto: Done. Also fixed the button not being uppercase for some reason.

crisbeto:
I re-added it, because it seemed off looking at the spec and I remembered it being there at some point. I'll remove when I get back home.

PR #4730, opened 2017-05-22
fix(checkbox): margin for empty checkboxes incorrectly added

### LINKED ISSUE #4720: Md-Checkbox has no margin when using dynamic label after #2121
After upgrading to material 2.0.0-beta5, checkboxes on my login screen no longer have margin on their right.

Apparently due to this change:
https://github.com/angular/material2/pull/2121

and to the fact I use dynamic labels (translated by ngx-translate to be precise).

I use something like that:
<md-checkbox [(ngModel)]="savePassword">{{ "tx_Login_SavePassword" | translate }}</md-checkbox>
                
To have margin (space between the checkbox and its label) on the right of the checkbox (no mat-checkbox-inner-container-no-side-margin class)

The class mat-checkbox-inner-container-no-side-margin is added to the checkbox when rendering the form. it only disappears when one clicks on the checkbox.

Here is the Plunker reproducing the problem (without ngx-translate but using a simple settimeout instead):

https://plnkr.co/5CFp9x

Working as it was in previous 2.0.0-beta

Material 2.0.0-beta5

Don't think so but please ask if I forgot something

DevVersion:
With #2121 the margin will be removed for checkboxes that don't have any label set.

A problem is that the Checkbox uses the OnPush change detection strategy and therefore the checkbox is not able to detect any delayed / async label change.

This means that checkboxes that set their label from an async binding will not have any margin until the users clicks on the checkbox.

Using the `cdkObserveContent` seems to be an elegant approach when using the OnPush strategy.

The `:empty` CSS selector would be more elegant but it's very sensitive about whitespaces and therefore it doesn't work properly.

Fixes #4720

### DIFF
--- a/src/lib/checkbox/checkbox.html
+++ b/src/lib/checkbox/checkbox.html
@@ -1,6 +1,6 @@
 <label [attr.for]="inputId" class="mat-checkbox-layout" #label>
   <div class="mat-checkbox-inner-container"
-       [class.mat-checkbox-inner-container-no-side-margin]="!_hasLabel">
+       [class.mat-checkbox-inner-container-no-side-margin]="!checkboxLabel.textContent.trim()">
     <input #input
            class="mat-checkbox-input cdk-visually-hidden" type="checkbox"
            [id]="inputId"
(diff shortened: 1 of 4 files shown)


jelbourn:
It makes me generally unhappy that even something as simple as checkbox has to have a `MutationObserver` (which is the natural escalation here). Don't know of another way to handle this, though (until `:blank` is well-supported).

DevVersion:
@jelbourn Reworked it a bit more. Yeah I'm not too happy with the solution either.

zukasmichael:
is there any workaround now for ^5.1.0 ?

Issue #4232, opened 2017-04-24
Allow to customize when md-error is shown


coder925:
#### Bug, feature request, or proposal:
Feature request

#### What is the expected behavior?
I would like to be able to customize when the *md-error* directive is visible.
E.g. when `ctrl.dirty && ctrl.invalid`.

#### What is the current behavior?
As per #3560 my understanding is that the *md-error* is only shown when the control is: `(isInvalid && (isTouched || isSubmitted))`

#### What are the steps to reproduce?
See my [Plunker](https://plnkr.co/edit/kBxsOfoxlwdLfm5ObWtD?p=preview).
- Name1 is using md-error. Error will only show after control been touched.
- Name2 is using an old workaround with md-hint. It can show error directly when dirty.

#### What is the use-case or motivation for changing an existing behavior?
For 99% of the time I believe the current implementation is the best behavior.

However, for special cases, you will need to be able to modify when errors are shown. For example instantly as the example from the [Angular documentation](https://angular.io/docs/ts/latest/guide/forms.html#!#show-and-hide-validation-error-messages).

#### Which versions of Angular, Material, OS, browsers are affected?
Angular v4.0.3
Material v2.0.0-beta.3

#### Is there anything else we should know?
I assume my request is not currently possible as [my question on StackOverflow](http://stackoverflow.com/questions/43500290/allow-md-error-to-show-the-error-message-even-if-control-is-untouched) did not get any suggested answers.

jelbourn:
@kara might also have some thoughts on this

willshowell:
@jelbourn, since you labeled #4750 as needing discussion, I'd like to illustrate my motivation for pushing for the change. In [this plunker](http://plnkr.co/edit/B5G7Kz75Nq9xsUvDWXiE?p=preview) you can see that validation must be performed at a group level since the `password` and `confirm` controls are dependent on one another. Per @kara's [suggestion](https://github.com/angular/angular/issues/5635#issuecomment-253372991), this is the best (if not only) way to achieve sibling validation.

Two workarounds must be put into place to emulate the correct behavior:
  1. `class.mat-input-invalid` must be bound to a custom error state matcher
  2. An `<md-hint>` must be styled like an `<md-error>`, thus losing entrance animations, and you must also bind it to the custom error state matcher with `*ngIf`

[...]

Issue #4672, opened 2017-05-19
Generic input-container


mbeckenbach:
#### Bug, feature request, or proposal:
feature request

#### What is the expected behavior?
There should be something similar to the md-input-container which does not depend on a input control inside, so that the benefits for styling error states, using md-error and md-hint can easily be used for custom form controls.

#### What is the current behavior?
md-input-container has a very limited support for form-controls.

#### What are the steps to reproduce?
Read the docs :-)

#### What is the use-case or motivation for changing an existing behavior?
When using md-input-container it's really easy to have those nice animated md-error hints and invalid highlight. It's not that easy to get angular to adding the ng-invalid, ng-dirty and so on classes to a parent element like the md-input-container. So creating a similar look and feel for custom controls can be hard. 

It would be very cool to use md-input-container for anything.

```html
<md-input-container>
    <my-cool-control mdInput></my-cool-control>
   <md-error>User did something wrong</md-error>
</md-input-container>
```

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
ng 4.1.x, md beta 5

jelbourn:
@mmalerba I know we planned to support this when designing the input container- what's the status on this?

mmalerba:
Haven't gotten time to work on it yet, but its a feature I'd love to add

natsid:
Is the idea for md-input-container to be able to contain all (or most?) of the other Material form controls so that they can all utilize md-hint and md-error, etc.?

jelbourn:
This issue is tracking the ability to do something like
```html
<md-input-container>
  <my-custom-input mdInput></my-custom-input>
</md-input-container>
```

There's a separate issue of having `md-select` and `md-input-container` share the same setup

natsid:
Ah, I missed the `mdInput` on the custom tag - thought it was just any custom component. Thanks.

mmalerba:
This is now available, see the guide: https://material.angular.io/guide/creating-a-custom-form-field-control

Issue #2965, opened 2017-02-07
md-hint add color property


tydanielson:
This is a feature.  It would be really nice to use the color property like on `md-button` to determine which hints are warnings, accents or primary etc...  I'm using the hints for both showing validation errors and anything related to the users input.

sumpton:
```<md-text>``` would be a terrific addition to work along with ```<md-hint>```

I have been refactoring hundreds of Bootstrap 4 fields to Material 2, and I appreciate ```<md-hint>``` very much. The positioning is also very useful.

But for validation messages I have been moving ```<div>``` out of the form-control, and wish there was a more direct and elegant Material component for Bootstrap's ```<... class="form-text">``` that could be used for all the form components.

**Comment by DevVersion** [comment]
Good idea. While I was thinking about the `color` property on `md-hint` I noticed that we often repeat the same logic to apply the color classes. 

A good solution would be having a directive that is responsible for the color classes. Need to discuss first.

**Comment by gvlekke** [comment]
I think it's also good to have a color on md-hint, but I think its better that you can set a error color to the input container and that material2 itself take that color and put in on the hint and the md-input-underline. 

For now I'm creating a css file that does that based on the ng-dirty and ng-invalid classes.

```
md-hint
  font-weight: lighter
  color: #A1A1A1

md-input-container.ng-dirty.ng-invalid
  .md-input-underline
    border-top-width: 2px
    border-color: #D50000

    .md-input-ripple
      background-color: #D50000

  md-hint
    color: #D50000
```

should i create a new issue: "add error styling to inputs" ?

**Comment by JoshMarler** [comment]
@gvlekke I agree with this being a good idea. Thanks for the css snippet. Helped me style my validation in the manner I required. Spent longer than I care to admit messing around having missed ng-dirty / ng-invalid overrides...

For those interested as of 2.4.6 relevant css is: 

```
md-input-container.ng-touched.ng-invalid {

  .mat-input-placeholder {
    color: red;
  }

  .mat-input-underline {
    border-top-width: 2px;
    border-color: red;
  }

  .mat-input-ripple {
    background-color: red;
  }

  md-hint {
    color: red;
  }
}
```

kamarouski:
Out of the box styling solution for invalid inputs and error messages would be great. 

Color property for md-hint would definitely help a lot as everybody now need to write custom css to achieve this common task. Or maybe a separate md-error component.

Hopefully this will be considered before final release.

niwrA:
[...]

Issue #2498, opened 2017-01-02
Support hint text with select


jt-helsinki:
#### Bug, feature request, or proposal:

Feature request
#### What is the expected behavior?

```
<md-select ... >
    <md-option ... ></md-option>
    ...
    ...
    <md-hint>some hint text</md-hint>
</md-select>
```

#### What is the current behavior?

not supported

#### What is the use-case or motivation for changing an existing behavior?

A hint below the select box. This could be used (For example) as a validation message when the select is not selected. Behaviour similar to the <input> component.

#### Which versions of Angular, Material, OS, browsers are affected?

Material2 v2.0.0-beta.1
Angular 2.4.1

SherwinJames:
Is there a recommended alternate way to handle error hints for md-select (e.g. required)?

leonardovff:
@SherwinJames look the soluction of codef0rmer in this issue https://github.com/angular/material2/issues/2624

dhineshwiz:
Can we something like 
`<div class="hint"> This is used as primary message</div>`
Supported in Md-Select

sdargaud:
Please don't forget md-error.

makdeniss:
For md-error for md-select I've done something like this:
```
<md-error *ngIf="type.touched && type.invalid">
          <span class="error-small-text" *ngIf="type.errors.required">
             Company Type is <strong>required</strong>
          </span>
</md-error>
```


```
.error-small-text {
  position: relative;
  top: -18px;
  font-size: 0.8em;
}
```

finalxcode:
+1 @makdeniss cool

jelbourn:
This is done now that select goes in mat-form-field

PR #4750, opened 2017-05-23
feat(input): add custom error state matcher

willshowell:
Fixes https://github.com/angular/material2/issues/4232

cc @mmalerba

### DIFF
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -19,11 +19,13 @@ export interface ErrorOptions {
   errorStateMatcher?: ErrorStateMatcher;
 }
 
+/** Returns whether control is invalid and is either touched or is a part of a submitted form. */
 export function defaultErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm) {
   const isSubmitted = form && form.submitted;
   return !!(control.invalid && (control.touched || isSubmitted));
 }
 
+/** Returns whether control is invalid and is either dirty or is a part of a submitted form. */
 export function showOnDirtyErrorStateMatcher(control: FormControl,
     form: FormGroupDirective | NgForm) {
   const isSubmitted = form && form.submitted;
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -1,0 +1,14 @@
+import {InjectionToken} from '@angular/core';
+import {NgControl, FormGroupDirective, NgForm} from '@angular/forms';
+
+/** Injection token that can be used to specify the global error options. */
+export const MD_ERROR_GLOBAL_OPTIONS =
+  new InjectionToken<() => boolean>('md-error-global-options');
+
+export type ErrorStateMatcherType =
+  (control: NgControl, parentFormGroup: FormGroupDirective, parentForm: NgForm) => boolean;
+
+export interface ErrorOptions {
+  errorStateMatcher?: ErrorStateMatcherType;
+  showOnDirty?: boolean;
+}
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -20,18 +20,12 @@ export interface ErrorOptions {
 }
 
 export function defaultErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm) {
-  const isInvalid = control.invalid;
-  const isTouched = control.touched;
   const isSubmitted = form && form.submitted;
-
-  return !!(isInvalid && (isTouched || isSubmitted));
+  return !!(control.invalid && (control.touched || isSubmitted));
 }
 
 export function showOnDirtyErrorStateMatcher(control: FormControl,
     form: FormGroupDirective | NgForm) {
-  const isInvalid = control.invalid;
-  const isDirty = control.dirty;
   const isSubmitted = form && form.submitted;
-
-  return !!(isInvalid && (isDirty || isSubmitted));
+  return !!(control.invalid && (control.dirty || isSubmitted));
 }
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -10,8 +10,7 @@ import {InjectionToken} from '@angular/core';
 import {NgControl, FormGroupDirective, NgForm} from '@angular/forms';
 
 /** Injection token that can be used to specify the global error options. */
-export const MD_ERROR_GLOBAL_OPTIONS =
-  new InjectionToken<ErrorOptions>('md-error-global-options');
+export const MD_ERROR_GLOBAL_OPTIONS = new InjectionToken<ErrorOptions>('md-error-global-options');
 
 export type ErrorStateMatcher =
   (control: NgControl, parentFormGroup: FormGroupDirective, parentForm: NgForm) => boolean;
@@ -20,26 +19,24 @@ export interface ErrorOptions {
   errorStateMatcher?: ErrorStateMatcher;
 }
 
-export class DefaultErrorStateMatcher {
+export function defaultErrorStateMatcher(control: NgControl, formGroup: FormGroupDirective,
+    form: NgForm): boolean {
 
-  errorStateMatcher(control: NgControl, formGroup: FormGroupDirective, form: NgForm): boolean {
-    const isInvalid = control && control.invalid;
-    const isTouched = control && control.touched;
-    const isSubmitted = (formGroup && formGroup.submitted) ||
-        (form && form.submitted);
+  const isInvalid = control && control.invalid;
+  const isTouched = control && control.touched;
+  const isSubmitted = (formGroup && formGroup.submitted) ||
+      (form && form.submitted);
 
-    return !!(isInvalid && (isTouched || isSubmitted));
-  }
+  return !!(isInvalid && (isTouched || isSubmitted));
 }
 
-export class ShowOnDirtyErrorStateMatcher {
+export function showOnDirtyErrorStateMatcher(control: NgControl, formGroup: FormGroupDirective,
+    form: NgForm): boolean {
 
-  errorStateMatcher(control: NgControl, formGroup: FormGroupDirective, form: NgForm): boolean {
-    const isInvalid = control && control.invalid;
-    const isDirty = control && control.dirty;
-    const isSubmitted = (formGroup && formGroup.submitted) ||
-        (form && form.submitted);
+  const isInvalid = control && control.invalid;
+  const isDirty = control && control.dirty;
+  const isSubmitted = (formGroup && formGroup.submitted) ||
+      (form && form.submitted);
 
-    return !!(isInvalid && (isDirty || isSubmitted));
-  }
+  return !!(isInvalid && (isDirty || isSubmitted));
 }
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -1,14 +1,45 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
 import {InjectionToken} from '@angular/core';
 import {NgControl, FormGroupDirective, NgForm} from '@angular/forms';
 
 /** Injection token that can be used to specify the global error options. */
 export const MD_ERROR_GLOBAL_OPTIONS =
-  new InjectionToken<() => boolean>('md-error-global-options');
+  new InjectionToken<ErrorOptions>('md-error-global-options');
 
-export type ErrorStateMatcherType =
+export type ErrorStateMatcher =
   (control: NgControl, parentFormGroup: FormGroupDirective, parentForm: NgForm) => boolean;
 
 export interface ErrorOptions {
-  errorStateMatcher?: ErrorStateMatcherType;
-  showOnDirty?: boolean;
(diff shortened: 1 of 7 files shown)


jelbourn:
I'm wondering if this should be a more generally configurable concept; `<md-select>` will really need the same thing.

jelbourn:
@kara should also weigh in as this touches on forms

willshowell:
@jelbourn agreed on making it more general. #4672 (or at least #2498) would help

willshowell:
@jelbourn since you last reviewed, I've made the following changes:

1. `errorStateMatcher` is a property on the input rather than the container (not ideal for reusability but necessary since https://github.com/angular/material2/pull/4757)
2. A global error state matcher can be provided via `MD_ERROR_GLOBAL_OPTIONS`
3. Can also configure all inputs to show errors when dirty instead of touched via `MD_ERROR_GLOBAL_OPTIONS` (since this seems to be a more common request https://github.com/angular/material2/issues/4232#issue-223812627, https://github.com/angular/material2/issues/2965#issuecomment-278975641)

I'm far from married to # 3, but wanted to explore other global options. Also if this isn't the right approach, that's fine, but I'd like to keep the conversation moving.

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,14 @@
+import {InjectionToken} from '@angular/core';
+import {NgControl, FormGroupDirective, NgForm} from '@angular/forms';
+
+/** Injection token that can be used to specify the global error options. */
+export const MD_ERROR_GLOBAL_OPTIONS =
+  new InjectionToken<() => boolean>('md-error-global-options');
+
+export type ErrorStateMatcherType =
```
mmalerba: just `ErrorStateMatcher` is fine

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,14 @@
+import {InjectionToken} from '@angular/core';
+import {NgControl, FormGroupDirective, NgForm} from '@angular/forms';
+
+/** Injection token that can be used to specify the global error options. */
+export const MD_ERROR_GLOBAL_OPTIONS =
+  new InjectionToken<() => boolean>('md-error-global-options');
```
mmalerba: generic type should be ErrorOptions

src/lib/input/input-container.spec.ts
```diff
@@ -706,6 +709,116 @@ describe('MdInputContainer', function () {
       });
     }));
 
+    it('should display an error message when a custom error matcher returns true', async(() => {
+      fixture.destroy();
```
mmalerba: can we just put this in a different describe block or something? its weird that we destroy the original fixture here

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type ErrorStateMatcherType =
+  (control: NgControl, parentFormGroup: FormGroupDirective, parentForm: NgForm) => boolean;
+
+export interface ErrorOptions {
+  errorStateMatcher?: ErrorStateMatcherType;
+  showOnDirty?: boolean;
```
mmalerba: I feel like this is structured kind of weirdly, how about if we remove the `showOnDirty` option and instead just add a couple implementations that the user can choose from (e.g. `DefaultErrorStateMatcher`, `ShowOnDirtyErrorStateMatcher`)

src/lib/input/input-container.spec.ts
```diff
@@ -706,6 +709,116 @@ describe('MdInputContainer', function () {
...
+
+      expect(component.formControl.invalid).toBe(true, 'Expected form control to be invalid');
+      expect(containerEl.querySelectorAll('md-error').length).toBe(0, 'Expected no error messages');
+
+      component.formControl.markAsTouched();
+      customFixture.detectChanges();
+
+      customFixture.whenStable().then(() => {
+        expect(containerEl.querySelectorAll('md-error').length)
+          .toBe(0, 'Expected no error messages after being touched.');
```
mmalerba: nit: indent continuations by 4

src/lib/input/input-container.ts
```diff
@@ -215,10 +224,14 @@ export class MdInputDirective {
...
               @Optional() private _parentForm: NgForm,
-              @Optional() private _parentFormGroup: FormGroupDirective) {
+              @Optional() private _parentFormGroup: FormGroupDirective,
+              @Optional() @Inject(MD_ERROR_GLOBAL_OPTIONS) errorOptions: ErrorOptions) {
 
     // Force setter to be called in case id was not specified.
     this.id = this.id;
+
+    this._errorOptions = errorOptions ? errorOptions : {};
+    this.errorStateMatcher = this._errorOptions.errorStateMatcher || undefined;
```
mmalerba: if you create the `DefaultErrorStateMatcher` class I suggested earlier this can be `|| new DefaultErrorStateMatcher()`

willshowell:
@mmalerba I've addressed your comments! Please advise if I misinterpreted any of your class implementation suggestions.

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {InjectionToken} from '@angular/core';
+import {NgControl, FormGroupDirective, NgForm} from '@angular/forms';
+
+/** Injection token that can be used to specify the global error options. */
+export const MD_ERROR_GLOBAL_OPTIONS =
+  new InjectionToken<ErrorOptions>('md-error-global-options');
```
mmalerba: nit: I think this will fit on 1 line

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export type ErrorStateMatcher =
+  (control: NgControl, parentFormGroup: FormGroupDirective, parentForm: NgForm) => boolean;
+
+export interface ErrorOptions {
+  errorStateMatcher?: ErrorStateMatcher;
+}
+
+export class DefaultErrorStateMatcher {
```
mmalerba: oh sorry, I don't know why I said class, I just meant functions `defaultErrorStateMatcher`

src/lib/input/input-container.spec.ts
```diff
@@ -1020,6 +1133,29 @@ class MdInputContainerWithFormErrorMessages {
 
 @Component({
   template: `
+    <form #form="ngForm" novalidate>
+      <md-input-container>
+        <input mdInput
+          [formControl]="formControl"
```
mmalerba: nit: indent continuation by 4

willshowell:
@mmalerba done!

The convenience function `showOnDirtyErrorStateMatcher` will show errors when,
```ts
isInvalid && (isDirty || isSubmitted)
```

But I'm curious if the expected behavior might be,
```ts
isInvalid && (isDirty || isTouched || isSubmitted)
```

src/lib/input/input-container.ts
```diff
@@ -250,12 +264,7 @@ export class MdInputDirective {
   /** Whether the input is in an error state. */
   _isErrorState(): boolean {
     const control = this._ngControl;
-    const isInvalid = control && control.invalid;
-    const isTouched = control && control.touched;
-    const isSubmitted = (this._parentFormGroup && this._parentFormGroup.submitted) ||
-        (this._parentForm && this._parentForm.submitted);
-
-    return !!(isInvalid && (isTouched || isSubmitted));
+    return this.errorStateMatcher(control, this._parentFormGroup, this._parentForm);
```
kara: I think we should be passing through `this._ngControl.control` here. In other words, the `FormControl` instance rather than the `NgControl` directive. I don't think most people are familiar with `NgControls`, and probably aren't familiar with the subtle distinction between the directive and the `FormControl` itself. While most properties are mirrored on the directive, a few aren't, and I foresee that being confusing.

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,42 @@
...
+  const isInvalid = control && control.invalid;
+  const isTouched = control && control.touched;
+  const isSubmitted = (formGroup && formGroup.submitted) ||
+      (form && form.submitted);
+
+  return !!(isInvalid && (isTouched || isSubmitted));
+}
+
+export function showOnDirtyErrorStateMatcher(control: NgControl, formGroup: FormGroupDirective,
+    form: NgForm): boolean {
```
kara: I don't think we should be surfacing both `FormGroupDirective` and `NgForm` here. They are mutually exclusive, so one of these args will always be null. We shouldn't put it on the user to check for the existence of one or the other every time. Given that they both implement the `Form` interface, it might make more sense to use the `Form` type here and only pass through the existing parent to this function.

src/lib/input/input-container.spec.ts
```diff
@@ -1020,6 +1136,29 @@ class MdInputContainerWithFormErrorMessages {
 
 @Component({
   template: `
+    <form #form="ngForm" novalidate>
```
kara: Remove `novalidate`? This is added by default by the `NgNoValidate` directive in `@angular/forms`.

src/lib/input/input-container.spec.ts
```diff
@@ -1020,6 +1136,29 @@ class MdInputContainerWithFormErrorMessages {
...
+      <md-input-container>
+        <input mdInput
+            [formControl]="formControl"
```
kara: `NgForm` comes with the `FormsModule` and `formControl` comes from the `ReactiveFormsModule`. It's not really great to mix them, and the point of using `formControl` is to have a standalone control *without* a parent. 

If you want a more "idiomatic" test case, I'd suggest switching to `formGroup` on the form tag and then use `formControlName` rather than `formControl`.

src/lib/core/error/error-options.ts
willshowell: I agree with the sentiment, but I'm not clear on how it would work. The `Form` interface doesn't have the `submitted` property, which is the main (if not only) reason we need it. I see two solutions

1. Use the union `FormGroupDirective | NgForm`

```ts
export function showOnDirtyErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm) {
  const isInvalid = control.invalid;
  const isDirty = control.dirty;
  const isSubmitted = form && form.submitted;

  return !!(isInvalid && (isDirty || isSubmitted));
}
```
2. Just pass `submitted` and don't expose either `FormGroupDirective` or `NgForm`
```ts
export function showOnDirtyErrorStateMatcher(control: FormControl, submitted: boolean) {
  const isInvalid = control.invalid;
  const isDirty = control.dirty;

  return !!(isInvalid && (isDirty || submitted));
}

Comment (kara) [inline src/lib/core/error/error-options.ts]: Hmm, the fact that the `submitted` property is missing from `Form` interface seems like an oversight (will look into it). Given that fact, I prefer #1.

**Comment by willshowell** [comment]
@kara comments addressed. Could you take another look?

File: src/lib/core/error/error-options.ts [hunk 14]
```diff
@@ -0,0 +1,37 @@
...
+export type ErrorStateMatcher =
+    (control: FormControl, form: FormGroupDirective | NgForm) => boolean;
+
+export interface ErrorOptions {
+  errorStateMatcher?: ErrorStateMatcher;
+}
+
+export function defaultErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm) {
+  const isInvalid = control.invalid;
+  const isTouched = control.touched;
```
kara: Nit: these const assignments seem to be adding bloat, given how short `control.touched` already is. I'm thinking it's easier to scan this function if you just inline (same on line 32). 

``` ts
const isSubmitted = form && form.submitted;
return !!(control.invalid && (control.touched || isSubmitted));
```

src/lib/input/input-container.spec.ts
```diff
@@ -749,6 +751,121 @@ describe('MdInputContainer', function () {
 
   });
 
+  describe('custom error behavior', () => {
+    it('should display an error message when a custom error matcher returns true', async(() => {
```
kara: Given that we're just using reactive form directives now, these tests don't need to be async. You should be able to safely remove the `async` and `whenStable` calls.

kara:
LGTM, waiting on @mmalerba to do a last pass.

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,31 @@
...
+export const MD_ERROR_GLOBAL_OPTIONS = new InjectionToken<ErrorOptions>('md-error-global-options');
+
+export type ErrorStateMatcher =
+    (control: FormControl, form: FormGroupDirective | NgForm) => boolean;
+
+export interface ErrorOptions {
+  errorStateMatcher?: ErrorStateMatcher;
+}
+
+export function defaultErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm) {
```
mmalerba: docs

src/lib/core/error/error-options.ts
```diff
@@ -0,0 +1,31 @@
...
+  const isSubmitted = form && form.submitted;
+  return !!(control.invalid && (control.touched || isSubmitted));
+}
+
+export function showOnDirtyErrorStateMatcher(control: FormControl,
```
mmalerba: docs

src/lib/core/error/error-options.ts
willshowell: @kara were you ever able to check into why `submitted` is missing from the `Form` interface?

PR #4754, opened 2017-05-23
fix(select): consistent error behavior to md-input-container

### LINKED ISSUE #4611: invalid MdSelect should highlight itself on parent form submitted
Bug

Expect md-select to show red (or 'warning' color) if it is invalid and its parent form is submitted.

Currently the select does show warning color on touched and invalid, but not on parent form submitted and invalid.

See this [plunk](http://plnkr.co/edit/sJdy093oOah5glQb4EXe?p=preview). Both the select and the input are 'required'. Click 'submit' and see that the select does not change its appearance like the input does.

Would like to highlight errors on submit.

all

The MdSelect likely needs to inject an @Optional NgForm/@Optional FormGroupDirective in its [constructor](https://github.com/angular/material2/blob/master/src/lib/select/select.ts#L313) like [MdInputContainer does](https://github.com/angular/material2/blob/master/src/lib/input/input-container.ts#L355) so that it can keep track of its submitted state.

crisbeto:
* Gets `md-select` to behave in the same way as `md-input-container` when it comes to errors. This means highlighting itself when it is invalid and touched, or one of the parent forms/form groups is submitted.
* Moves the error state logic into a separate function in order to avoid some hard-to-follow selectors and to potentially allow overrides. This should also be a first step to supporting `md-error` inside `md-select`.
* Changes the required asterisk to always have the theme warn color, similarly to the input asterisk.

Fixes #4611.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -28,6 +28,7 @@ import {
   ChangeDetectionStrategy,
   InjectionToken,
 } from '@angular/core';
+import {NgForm, FormGroupDirective} from '@angular/forms';
 import {MdOption, MdOptionSelectionChange, MdOptgroup} from '../core/option/index';
 import {ENTER, SPACE, UP_ARROW, DOWN_ARROW, HOME, END} from '../core/keyboard/keycodes';
 import {FocusKeyManager} from '../core/a11y/focus-key-manager';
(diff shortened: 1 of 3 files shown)


src/lib/select/select.ts
```diff
@@ -533,6 +536,17 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
     this._setScrollTop();
   }
 
+  /** Whether the select is in an error state. */
+  _isErrorState(): boolean {
+    const control = this._control;
```
kara: Nit: seems like you could inline this, given how short it is.

crisbeto:
Rebased and addressed the feedback @kara.

andrewseguin:
Rebase please

finalxcode:
+1

PR #4771, opened 2017-05-24
fix(input): correct invalid colors

### LINKED ISSUE #4651: Accent inputs show wrong underline color when invalid
Bug

#### Expected behavior
The accent underline changes to error color, as it does for primary inputs.

#### Current behavior
The underline remains the accent color.

#### Repro
http://plnkr.co/edit/DOt4Wa1nBloTDSSEejPO?p=preview

willshowell:
Fixes #4651 

cc @mmalerba 

It was all related to specificity issues, so I decreased the specificity of `.mat-input-ripple`'s background-color and increased it for `.mat-input-placeholder`'s and `.mat-placeholder-required`'s color.

Also, I am unsure as to what the correct behavior for the required asterisk is. Previously, it was `accent` while floating for all (primary, accent, warn) inputs. It was only `warn` when invalid and not focused. I chose to ensure it is `warn` at all times when invalid, but can revert if that's contrary to the spec.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -42,7 +42,7 @@
     color: mat-color($foreground, disabled-text);
   }
 
-  // See mat-input-placeholder-floating mixin in input.scss
+  // See mat-input-placeholder-floating mixin in input-container.scss
   input.mat-input-element:-webkit-autofill + .mat-input-placeholder,
   .mat-focused .mat-input-placeholder.mat-float {
     .mat-placeholder-required {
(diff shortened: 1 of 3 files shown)


src/lib/input/_input-theme.scss
```diff
@@ -52,26 +52,28 @@
...
 
   // Styling for the error state of the input container. Note that while the same can be
   // achieved with the ng-* classes, we use this approach in order to ensure that the same
   // logic is used to style the error state and to show the error messages.
   .mat-input-invalid {
-    .mat-input-placeholder,
-    .mat-placeholder-required {
-      color: $input-underline-color-warn;
+    .mat-input-placeholder {
+      &, &.mat-accent, &.mat-float .mat-placeholder-required {
```
mmalerba: can you put each selector on its own line (makes it obvious that there's multiple and not one complex one).

also I think we can't have rules that end with `&` in our theme files, so maybe:

```scss
.mat-input-placeholder {
  color: $input-underline-color-warn;

  &.mat-accent,
  &.mat-float .mat-placeholder-required {
    color: $input-underline-color-warn;
  }
}
```

PR #4778, opened 2017-05-24
feat(dialog): allow setting the layout direction

crisbeto:
Allows users to set the layout direction of a dialog.

### DIFF
--- a/src/lib/dialog/dialog.spec.ts
+++ b/src/lib/dialog/dialog.spec.ts
@@ -312,6 +312,16 @@ describe('MdDialog', () => {
     expect(overlayPane.style.width).toBe('200px');
   });
 
+  it('should allow setting the layout direction', () => {
+    let dialogRef = dialog.open(PizzaMsg, { direction: 'rtl' });
+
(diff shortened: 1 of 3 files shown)


src/lib/dialog/dialog.spec.ts
```diff
@@ -312,6 +312,16 @@ describe('MdDialog', () => {
...
   });
 
+  it('should allow setting the layout direction', () => {
+    let dialogRef = dialog.open(PizzaMsg, { direction: 'rtl' });
+
+    viewContainerFixture.detectChanges();
+
+    let overlayPane = overlayContainerElement.querySelector('.cdk-overlay-pane');
+
+    expect(overlayPane.getAttribute('dir')).toBe('rtl');
```
jelbourn: optional: shorter to just to `overlayPane.dir`

crisbeto: I have some vague memories about trying that a while ago and failing in IE.

kapocius:
When to expect merge? Thanks

PR #4785, opened 2017-05-24
fix(snack-bar): close snack bar on route changed

crisbeto:
Closes any open snack bars if the route changes and the user has the `RouterModule`.

**Note:** We don't have an issue for this, but it came up in a discussion on a PR.


package.json
```diff
@@ -31,6 +31,7 @@
     "@angular/core": "^4.0.0",
     "@angular/forms": "^4.0.0",
     "@angular/http": "^4.0.0",
+    "@angular/router": "^4.0.0",
```
jelbourn: I don't want to introduce a dependency on `@angular/router`, since people can use different routers. 

Why wouldn't this approach be the same as the dialog, which uses `Location`?

crisbeto: This doesn't introduce a dependency on the router, because we don't import the `RouterModule` anywhere and the injected `Router` is optional (it works in a similar way to the icons and the `HttpModule`). The difference between this and the `Location` is that `Location` only fires on hash changes and the user presses the back button, whereas the `Router` will also fire when clicking on links.

jelbourn: This will still cause the generated code to reference symbols from the `@angular/router` package, though, meaning that the user will have to both install it _and_ have some non-zero payload from `@angular/router`.

crisbeto: I don't think that's the case. When I was making `@angular/http` optional for the icons, I had the same approach and removing the entire modules from the `node_modules` still worked up until the point where we actually hit the  HTTP module functionality. Since we're only referencing the type, not the class directly, TypeScript doesn't need to load the module. [Here's an example that mirrors what we have in this PR.](https://www.typescriptlang.org/play/index.html#src=import%20%7B%20Router%2C%20OtherRouterClass%20%7D%20from%20'%40angular%2Frouter'%3B%0D%0A%0D%0Aexport%20class%20Something%20%7B%0D%0A%20%20%20%20constructor(private%20router%3A%20Router)%20%7B%0D%0A%20%20%20%20%20%20%20%20router.doStuff()%3B%0D%0A%20%20%20%20%7D%0D%0A%7D) Note how the `define` doesn't reference `@angular/router`, that will only happen if you try to do something like `new Router()`.

jelbourn: In JIT mode that might work, but it won't be the case for AoT. Angular uses the type for DI, so the generated code actually *does* have a real reference to the symbol. 

Trying it out now, if I create an empty project w/ the CLI, add `MdIconModule` and delete `@angular/http` from node_modules, it doesn't even build:
```
ERROR in Error encountered resolving symbol values statically. 
Could not resolve @angular/http relative to 
/tmp/deptest/node_modules/@angular/material/typings/index.d.ts., 
resolving symbol ICON_REGISTRY_PROVIDER in 
/tmp/deptest/node_modules/@angular/material/typings/index.d.ts, 
resolving symbol MdIconModule in 
/tmp/deptest/node_modules/@angular/material/typings/index.d.ts, 
resolving symbol MdIconModule in 
/tmp/deptest/node_modules/@angular/material/typings/index.d.ts
```
(which is an error coming from the metadata extractor)

crisbeto:
Closing to avoid introducing a dependency on `@angular/router`.

package.json
tsvetan-ganev: @jelbourn, I just encountered this issue when I migrated from `HttpModule` to the new `HttpClientModule`. I decided to remove the dependency (`@angular/http`) from the `package.json` file since I thought that I no longer need it. To my surprise this broke my build 😢.

PR #4787, opened 2017-05-24
build: import tslib instead of letting tsc generate helpers

jelbourn:
CC @IgorMinar

### DIFF
--- a/tools/gulp/packaging/rollup-helpers.ts
+++ b/tools/gulp/packaging/rollup-helpers.ts
@@ -2,8 +2,13 @@ import {LICENSE_BANNER} from '../constants';
 
 // There are no type definitions available for these imports.
 const rollup = require('rollup');
+const rollupNodeResolutionPlugin = require('rollup-plugin-node-resolve');
 
 const ROLLUP_GLOBALS = {
+  // Import tslib rather than having TypeScript output its helpers multiple times.
+  // See https://github.com/Microsoft/tslib
+  'tslib': 'tslib',
+
   // Angular dependencies
   '@angular/animations': 'ng.animations',
   '@angular/core': 'ng.core',
@@ -50,10 +55,10 @@ export type BundleConfig = {
 
 /** Creates a rollup bundle of a specified JavaScript file.*/
 export function createRollupBundle(config: BundleConfig): Promise<any> {
-  const bundleOptions = {
+  const bundleOptions: any = {
     context: 'this',
     external: Object.keys(ROLLUP_GLOBALS),
-    entry: config.entry
+    entry: config.entry,
   };
 
   const writeOptions = {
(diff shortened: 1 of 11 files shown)


tools/gulp/packaging/rollup-helpers.ts
```diff
@@ -67,5 +72,19 @@ export function createRollupBundle(config: BundleConfig): Promise<any> {
     sourceMap: true
   };
 
+  // When creating a UMD, we want to exclude tslib from the `external` and `globals` bundle options
+  // so that it is inlined into the bundle.
+  if (config.format === 'umd') {
+    bundleOptions.plugins = [rollupNodeResolutionPlugin()];
+
+    // Clone ROLLUP_GLOBALS so we don't accidentally modify it.
+    bundleOptions.globals = Object.assign({}, ROLLUP_GLOBALS);
```
tinayuangao: Do we want to change `writeOptions.globals`?

jelbourn: Good catch, I don't think this part is doing anything at all (and isn't needed)

tools/gulp/packaging/rollup-helpers.ts
```diff
@@ -67,5 +72,19 @@ export function createRollupBundle(config: BundleConfig): Promise<any> {
     sourceMap: true
   };
 
+  // When creating a UMD, we want to exclude tslib from the `external` and `globals` bundle options
+  // so that it is inlined into the bundle.
+  if (config.format === 'umd') {
```
DevVersion: I think we should rather add a new option to the `BundleConfig` that says `includeTslib`

tools/gulp/packaging/rollup-helpers.ts
```diff
@@ -67,5 +72,19 @@ export function createRollupBundle(config: BundleConfig): Promise<any> {
...
+    bundleOptions.plugins = [rollupNodeResolutionPlugin()];
```
DevVersion: In theory we could always add this plugin. It shouldn't change anything for normal FESM bundles.

jelbourn: I'd rather leave it where it's needed rather than just installing it all the time

tools/gulp/packaging/rollup-helpers.ts
jelbourn: I think that would be redundant, though, since we only want to do this for UMDs anyway

tools/gulp/packaging/rollup-helpers.ts
jelbourn: Done

tools/gulp/packaging/rollup-helpers.ts
DevVersion: It's just too much magic IMO. The `rollup-helpers` is a helper file that should be at least a bit generic.

Especially when we expose the package utilities (to for example flex-layout)

tools/gulp/packaging/rollup-helpers.ts
DevVersion: Okay, that's a fair point.

tools/gulp/packaging/rollup-helpers.ts
jelbourn: We would want the same behavior for flex-layout (and anything else we build). The UMD really is the special case.

DevVersion: That was just an example. At some point it will end up similar as with the `convertIndexShorthandImports` in `tsc-wrapped`. 

Anyway, your decision. I don't like it to be honest and this makes the helper into something more magic.

tinayuangao:
LGTM

PR #4788, opened 2017-05-25
fix(input): make input spacing match spec

mmalerba:
redo the input CSS to make the spacing between various elements match the material spec (https://material.io/guidelines/components/text-fields.html).

The spec shows what things look like at 16px font size, we currently use 14px which I could change, but for now I've left it so everything should look like a scaled down version of the mock.

Also adds support for icon buttons inside prefix and suffix

### DIFF
--- a/src/lib/core/typography/_typography.scss
+++ b/src/lib/core/typography/_typography.scss
@@ -25,7 +25,7 @@
   $caption:     mat-typography-level(12px, 12px, 400),
   $button:      mat-typography-level(14px, 14px, 500),
   // Line-height must be unit-less fraction of the font-size.
-  $input:       mat-typography-level(16px, 1.125, 400)
+  $input:       mat-typography-level(inherit, 1.125, 400)
 ) {
   @return (
     font-family: $font-family,
--- a/src/lib/core/typography/_typography.scss
+++ b/src/lib/core/typography/_typography.scss
@@ -23,7 +23,9 @@
   $body-2:      mat-typography-level(14px, 24px, 500),
   $body-1:      mat-typography-level(14px, 20px, 400),
   $caption:     mat-typography-level(12px, 12px, 400),
-  $button:      mat-typography-level(14px, 14px, 500)
+  $button:      mat-typography-level(14px, 14px, 500),
+  // Line-height must be unit-less fraction of the font-size.
+  $input:       mat-typography-level(16px, 1.125, 400)
 ) {
   @return (
     font-family: $font-family,
@@ -37,7 +39,8 @@
     body-2:      $body-2,
     body-1:      $body-1,
     caption:     $caption,
-    button:      $button
+    button:      $button,
+    input:       $input,
   );
 }
 
@@ -61,7 +64,7 @@
   }
 
   .mat-h2, #{$selector} h2 {
-    @include mat-typography-level-to-styles($config, display-2)
+    @include mat-typography-level-to-styles($config, display-2);
 
     // Note: The spec doesn't mention letter spacing. The value comes from
     // eyeballing it until it looked exactly like the spec examples.;
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -91,20 +91,121 @@
 }
 
 @mixin mat-input-typography($config) {
-  .mat-input-container {
-    font-family: mat-font-family($config);
-    line-height: normal;
+  // The unit-less line-height from the font config.
+  $mat-input-line-height: mat-line-height($config, input);
 
-    .mat-icon {
-      font-size: 100%;
+  // The amount to scale the font for the floating label and subscript.
+  $mat-input-subscript-font-scale: 0.75;
+  // The amount to scale the font for the prefix and suffix icons.
+  $mat-input-prefix-suffix-icon-font-scale: 1.5;
+
+  // The amount of space between the top of the line and the top of the actual text
+  // (as a fraction of the font-size).
+  $mat-input-line-spacing: ($mat-input-line-height - 1) / 2;
+  // The padding on the infix. Mocks show half of the text size, but seem to measure from the edge
+  // of the text itself, not the edge of the line; therefore we subtract off the line spacing.
+  $mat-input-infix-padding: 0.5em - $mat-input-line-spacing;
+  // The margin applied to the input-infix to reserve space for the floating label.
+  $mat-input-infix-margin-top: 1em * $mat-input-line-height * $mat-input-subscript-font-scale;
+  // Font size to use for the label and subscript text.
+  $mat-input-subscript-font-size: $mat-input-subscript-font-scale * 100%;
+  // Font size to use for the for the prefix and suffix icons.
+  $mat-input-prefix-suffix-icon-font-size: $mat-input-prefix-suffix-icon-font-scale * 100%;
+  // The space between the bottom of the input table and the subscript container. Mocks show half of
+  // the text size, but this margin is applied to an element with the subscript text font size, so
+  // we need to divide by the scale factor to make it half of the original text size. We again need
+  // to subtract off the line spacing since the mocks measure to the edge of the text, not the  edge
+  // of the line.
+  $mat-input-subscript-margin-top:
+      0.5em / $mat-input-subscript-font-scale - $mat-input-line-spacing;
+  // The padding applied to the input-wrapper to reserve space for the subscript, since it's
+  // absolutely positioned. This is a combination of the subscript's margin and line-height, but we
+  // need to multiply by the subscript font scale factor since the wrapper has a larger font size.
+  $mat-input-wrapper-padding-bottom:
+      ($mat-input-subscript-margin-top + $mat-input-line-height) * $mat-input-subscript-font-scale;
+
+  // Applies a floating placeholder above the input itself.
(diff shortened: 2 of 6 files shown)


src/lib/input/input-container.scss
```diff
@@ -3,70 +3,124 @@
...
 }
 
+
 .mat-input-container {
   display: inline-block;
   position: relative;
   font-family: $mat-font-family;
-  line-height: normal;
+
+  // TODO(mmalerba): should this come from typography settings? If so a lot of this CSS needs to go
```
crisbeto: There will be a typography mixin in https://github.com/angular/material2/pull/4375.

src/lib/input/input-container.scss
```diff
@@ -3,70 +3,124 @@
...
+  flex: none;
 
-  & > * {
-    display: table-cell;
+  // Allow icons in a prefix or suffix to adapt to the correct size.
+  .mat-icon,
+  .mat-datepicker-toggle {
+    width: 1em;
+    height: 1em;
+    font-size: $mat-input-prefix-suffix-icon-font-size;
```
crisbeto: This one could use a TODO to switch it to the typography API.

src/lib/input/input-container.scss
mmalerba: Ah cool, so the way I set up the styles for this, I actually need the ratio of line-height/font-size. I could try to derive this by just dividing the variables in your `mat-typography-level` but that expects that they have the same units. Would it be better to have the mixin take the unitless ratio?

src/lib/input/input-container.scss
mmalerba: Actually this one I want to depend on the input's font-size so that the ratio from the spec is preserved. Therefore doesn't need to come from the typography API

src/lib/input/input-container.scss
crisbeto: I'm not sure whether we'd be able to exactly match the spec if we went unitless. That being said, not all values have to go through the typogaphy API necessarily. You can take the font size and have a custom line height.

mmalerba: Why wouldn't we be able to match the spec? For example we would just change:

```scss
$subheading:  mat-typography-level(16px, 28px, 400),
```

to

```scss
$subheading:  mat-typography-level(16px, 28 / 16, 400),
```

crisbeto: Hmm you're right, I'll change it after the initial PR gets in. It also doesn't handle margins at the moment as well so I can do them together. If you hardcode it and leave a comment, I can change it when I do a pass on the components.

src/lib/input/input-container.html
```diff
@@ -1,8 +1,7 @@
 <div class="mat-input-wrapper">
-  <div class="mat-input-table">
+  <div class="mat-input-flex">
     <div class="mat-input-prefix" *ngIf="_prefixChildren.length">
-      <!-- TODO(andrewseguin): remove [md-prefix] -->
-      <ng-content select="[mdPrefix], [matPrefix], [md-prefix]"></ng-content>
+      <ng-content select="[mdPrefix], [matPrefix]"></ng-content>
```
jelbourn: There are three use sites for `md-prefix` and one for `md-suffix` inside google

src/lib/input/input-container.scss
```diff
@@ -3,70 +3,124 @@
...
-$mat-input-wrapper-spacing: 1em !default;
-$mat-input-hint-min-space: 10px !default;
+// Min amount of space between start and end hint.
+$mat-input-hint-min-space: 1em !default;
+// The amount of additional line space beyond 1em on the input (1/2 below, 1/2 above).
+$mat-input-line-spacing: 0.125 !default;
+// The height of the underline.
+$mat-input-underline-height: 1px !default;
+// The amount to scale the font for the floating label and subscript.
+$mat-input-helper-font-scale: 0.75 !default;
```
jelbourn: "helper"?

src/lib/input/input-container.scss
```diff
@@ -3,70 +3,124 @@
...
+// subtract off 1/2 of the line spacing since the mocks measure to the edge of the text, not the
+// edge of the line.
+$mat-input-subscript-margin-top: 0.5em / $mat-input-helper-font-scale - $mat-input-line-spacing / 2;
+// The padding applied to the input-wrapper to reserve space for the subscript, since it's
+// absolutely positioned. This is a combination of the subscript's margin and line-height, but we
+// need to multiply by the helper font scale factor since the wrapper has a larger font size.
+$mat-input-wrapper-padding-bottom:
+    ($mat-input-subscript-margin-top + $mat-input-line-height) * $mat-input-helper-font-scale;
+// The amount of y-offset to apply to the placeholder when it is floating.
+$mat-input-floating-placeholder-offset: -$mat-input-infix-padding - 1em;
```
jelbourn: optional: maybe it would be useful to have a diagram for how all this comes together?

src/lib/input/input-container.scss
```diff
@@ -3,70 +3,124 @@
...
-$mat-input-underline-disabled-background-image:
-    linear-gradient(to right, rgba(0, 0, 0, 0.26) 0%, rgba(0, 0, 0, 0.26) 33%, transparent 0%);
 
 // Applies a floating placeholder above the input itself.
 @mixin mat-input-placeholder-floating {
   display: block;
-  transform: translate3d(0, -1.35em, 0) scale($mat-input-floating-placeholder-scale-factor);
-  width: 100% / $mat-input-floating-placeholder-scale-factor;
+  transform: translateY(-$mat-input-infix-margin-top - $mat-input-infix-padding)
+             perspective(1px) // fixes text blurriness.
```
jelbourn: Huh, I've never seen this before in my life. Can you elaborate in a comment on what the problem is and how this fixes it? Any resources to link that led to this approach?

src/lib/input/input-container.html
mmalerba: I'll fix them

src/lib/input/input-container.scss
mmalerba: it's the terminology used in the spec, changed it to subscript to match our terminology

src/lib/input/input-container.scss
mmalerba: just trial and error of various approaches people suggested, linked to the site where I saw this suggestion

src/lib/input/input-container.scss
```diff
@@ -20,33 +20,34 @@ $mat-input-line-height: 1 + $mat-input-line-spacing;
...
 
 // Applies a floating placeholder above the input itself.
 @mixin mat-input-placeholder-floating {
   display: block;
   transform: translateY(-$mat-input-infix-margin-top - $mat-input-infix-padding)
-             perspective(1px) // fixes text blurriness.
-             scale($mat-input-helper-font-scale);
-  width: 100% / $mat-input-helper-font-scale;
+             // fixes text blurriness, see
+             // https://css-tricks.com/forums/topic/transforms-cause-font-smoothing-weirdness-in-webkit/
```
jelbourn: Would be better to link to the relevant article instead of the question
http://www.useragentman.com/blog/2014/05/04/fixing-typography-inside-of-2-d-css-transforms/

mmalerba:
Addressed comments

@crisbeto managed to tweak the transforms to get it to look decent on Chrome, IE, and FF

mmalerba:
rebased and moved a bunch of the styles into the typography mixin, PTAL

src/lib/core/typography/_typography.scss
```diff
@@ -23,7 +23,9 @@
   $body-2:      mat-typography-level(14px, 24px, 500),
   $body-1:      mat-typography-level(14px, 20px, 400),
   $caption:     mat-typography-level(12px, 12px, 400),
-  $button:      mat-typography-level(14px, 14px, 500)
+  $button:      mat-typography-level(14px, 14px, 500),
+  // Line-height must be unit-less fraction of the font-size.
+  $input:       mat-typography-level(16px, 1.125, 400)
```
mmalerba: @jelbourn @crisbeto spec says `16px` for inputs, but by doing this I ruin its ability to inherit from its parent (e.g. so that an input inside an `<h1>` matches the font size, should I keep this as `16px` or change to `inherit`?

jelbourn: I think the inherit is better; users can always set it to 16px if they want

src/lib/input/_input-theme.scss
```diff
@@ -91,20 +91,121 @@
...
+    transform: translateY(-$mat-input-infix-margin-top - $mat-input-infix-padding)
+               scale($mat-input-subscript-font-scale)
+               // We use perspecitve to fix the text blurriness as described here:
+               // http://www.useragentman.com/blog/2014/05/04/fixing-typography-inside-of-2-d-css-transforms/
+               // This results in a small jitter after the label floats on Firefox, which the
+               // translateZ fixes.
+               perspective(100px) translateZ(0.001px);
+    // The tricks above used to smooth out the animation on chrome and firefox actually make things
+    // worse on IE, so we don't include them in the IE version.
+    -ms-transform: translateY(-$mat-input-infix-margin-top - $mat-input-infix-padding)
```
jelbourn: This doesn't cause a stylelint error?

mmalerba: nope, seems fine

PR #4804, opened 2017-05-25
fix(datepicker): restore focus to trigger element

crisbeto:
* The datepicker now restores focus to whatever element was focused before it was open.
* Adds a test for the functionality that closes the datepicker when pressing escape.

### DIFF
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -36,7 +36,7 @@ import {Subscription} from 'rxjs/Subscription';
 import {MdDialogConfig} from '../dialog/dialog-config';
 import {DateAdapter} from '../core/datetime/index';
 import {createMissingDateImplError} from './datepicker-errors';
-import {ESCAPE, TAB} from '../core/keyboard/keycodes';
+import {ESCAPE} from '../core/keyboard/keycodes';
 import {MdCalendar} from './calendar';
 import 'rxjs/add/operator/first';
 
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -20,7 +20,9 @@ import {
   ViewContainerRef,
   ViewEncapsulation,
   NgZone,
+  Inject,
 } from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
 import {Overlay} from '../core/overlay/overlay';
 import {OverlayRef} from '../core/overlay/overlay-ref';
 import {ComponentPortal} from '../core/portal/portal';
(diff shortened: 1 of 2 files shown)


mmalerba:
I think for datepicker we don't want to close on tab (see various accessible datepicker implementations: http://www.webaxe.org/accessible-date-pickers/)

crisbeto:
In that case we'd need to trap focus inside the popup, otherwise it goes out of the page.

mmalerba:
I believe it does: https://github.com/angular/material2/blob/master/src/lib/datepicker/datepicker-content.html#L1

The behavior might be a little weird though, because the dialog (which we use for the touch UI) does its own focus trapping. What I'd like to eventually do is create a `MdPopup`, similar to `MdDialog`. Things like select and auto-complete could take advantage of this as well. One of the APIs on `MdPopup` would allow you to set it to floating mode where its acts more like a dialog (not attached to the input) that way datepicker could just use the popup instead of the overlay/dialog hybrid approach

crisbeto:
It doesn't look like the focus trapping works when it's not in touch mode. Should I remove the tabbing behavior from this PR and try address the focus trapping separately?

mmalerba:
yeah sgtm

src/lib/datepicker/datepicker.ts
```diff
@@ -30,7 +32,7 @@ import {Subscription} from 'rxjs/Subscription';
 import {MdDialogConfig} from '../dialog/dialog-config';
 import {DateAdapter} from '../core/datetime/index';
 import {createMissingDateImplError} from './datepicker-errors';
-import {ESCAPE} from '../core/keyboard/keycodes';
+import {ESCAPE, TAB} from '../core/keyboard/keycodes';
```
mmalerba: TAB no longer needed

src/lib/datepicker/datepicker.ts
```diff
@@ -226,6 +229,11 @@ export class MdDatepicker<D> implements OnDestroy {
     if (this._calendarPortal && this._calendarPortal.isAttached) {
       this._calendarPortal.detach();
     }
+    if (this._focusedElementBeforeOpen && 'focus' in this._focusedElementBeforeOpen) {
```
mmalerba: why is the `'focus' in` part needed?

crisbeto: It's there for IE's sake. In some cases it assigns the `activeElement` to something that doesn't have a `focus` method. It's what was throwing that 4mb error in the CI.

mmalerba: interesting, good to know

PhilipSultanescu:
Also got this focus problem, pretty important for us. Can someone please rebase and merge it?

PhilipSultanescu:
Works very good, thank you for merging this.

Issue #4460, opened 2017-05-10
Github payload status


DevVersion:
It would be cool to show the delta of the whole library using a Github status. (e.g as with the screenshots)

The "status" would be always "success" because I don't think there is a way to just show like an info on PRs.

DevVersion:
Done.

PR #4806, opened 2017-05-25
build: add dashboard firebase functions

DevVersion:
* Adds a new folder for the dashboard project that for now only contains firebase functions.
* The firebase functions will be used to securely make API calls to the Github API (e.g payload status)
* Functions will be automatically deployed as same as screenshot functions.

Required for future dashboard features and stuff like #4460


DevVersion:
@jelbourn Made the changes we talked about on Slack. Let me know if we should improve anything with the `functions.ts` file.

PR #4809, opened 2017-05-25
fix(dialog): blurry content and slow animation in IE

crisbeto:
Fixes the dialog content being blurry and the animation being really janky in IE11.


ocarreterom:
I think this PR removes the Y axis transition. Is that the purpose?

angular/material: 2.0.0-beta.6-c2816ef
angular/cli: 1.1.1
angular/animations: 4.2.1
node: 8.1.0
os: darwin x64

PR #4818, opened 2017-05-26
build: move release packages to constants

DevVersion:
* Moves the list of packages that should be released to the constants file.
* Sets the executable bit flag on the `deploy-screenshot-functions.sh` script


tools/gulp/constants.ts
```diff
@@ -29,6 +29,12 @@ export const HTML_MINIFIER_OPTIONS = {
   removeAttributeQuotes: false
 };
 
+/** Packages that will be published to NPM by the release task. */
+export const RELEASE_PACKAGES = [
+  'cdk',
+  'material',
+];
```
jelbourn: I don't really like this; I don't see why have a `constants.ts` file full of unrelated things is better than having the constant for release packages defined where it's used.

DevVersion: This way you can find stuff better and it can be like a config. It's probably the same question about having a configuration file for Pixacto.

DevVersion: This could be theoretically also a `config.json` file that configures out build tasks.

jelbourn: Ah, you mean it to be like a configuration for a re-usable library. That makes more sense. In that case I'd be okay with it if we rename the file "build-config.ts" for now.

DevVersion: Yeah sorry if it was confusing. Let's also try to get rid of all the path constants in a follow-up then?

Because I don't think it makes sense to store constants like `DIST_MATERIAL` in a config?

jelbourn: Probably would make more sense to just have `distRoot` and then the output for each package does into a subdir of that.

DevVersion: Yeah you would just configure the output folder and the different "tasks" will append the package name for example

```ts
join(DIST_ROOT, 'packages', 'material')
```

DevVersion: @jelbourn Done. Just renamed the file to `build-config.ts`.

PR #4822, opened 2017-05-26
build: move server rendering to dist

DevVersion:
* Verifying that Material works with `platform-server` now happens in the `dist/` folder (instead of the src/ folder)
* Adds a gulp task to build the universal application. In the future serving the app will be also part of this


.travis.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     - env: "MODE=e2e"
     - env: "MODE=aot"
     - env: "MODE=payload"
-    - env: "MODE=prerender"
+    - env: "MODE=server-rendering"
```
jelbourn: I'd prefer to keep this a `prerender` since there's not actually a "server" involved

src/universal-app/tsconfig-server.json
```diff
@@ -0,0 +1,25 @@
+// TypeScript config file that is used to compile the server code for the Universal App.
```
jelbourn: Same comment on the term "prerender" vs. "server"

tools/gulp/tasks/universal.ts
```diff
@@ -0,0 +1,51 @@
...
+  ['universal:build-app', 'universal:copy-server-files'],
+  'universal:build-server-ts'
+));
+
+task('universal:build-app', sequenceTask(
+  ['universal:build-app-ts', 'universal:build-app-scss', 'universal:copy-app-assets']
+));
+
+task('universal:build-app-ts', ngcBuildTask(tsconfigAppPath));
+task('universal:build-app-scss', sassBuildTask(outDir, appDir));
```
jelbourn: There's no scss for this app

src/universal-app/server.ts
```diff
@@ -1,15 +1,16 @@
 import 'reflect-metadata';
```
jelbourn: Same thing on keeping this `prerender`

tools/gulp/tasks/universal.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/** Path to the universal-app tsconfig files. */
+const tsconfigAppPath = join(appDir, 'tsconfig-build.json');
+const tsconfigServerPath = join(outDir, 'tsconfig-server.json');
+
+/** Glob that matches all assets that need to copied to the dist. */
+const assetsGlob = join(appDir, '**/*.+(html|css|json)');
+
+/** Glob that matches all server files that should be copied to the dist. */
+const serverGlob = join(appDir, 'server.ts');
```
jelbourn: Why is this a glob?

DevVersion:
Addressed feedback.

PR #4823, opened 2017-05-26
feat(data-table): use iterate differ to optimize rendering

andrewseguin:


src/lib/core/data-table/data-table.spec.ts
```diff
@@ -97,6 +96,39 @@ describe('CdkTable', () => {
...
+
+    // Prove that the attributes match their indicies
+    const initialRows = getRows(tableElement);
+    expect(initialRows[0].getAttribute('initialIndex')).toBe('0');
+    expect(initialRows[1].getAttribute('initialIndex')).toBe('1');
+    expect(initialRows[2].getAttribute('initialIndex')).toBe('2');
+
+    // Swap first and second data in data array
+    let copiedData = component.dataSource.data.slice();
+    let temp = copiedData[0];
```
ErinCoughlan: let -> const in both of these

src/lib/core/data-table/data-table.ts
```diff
@@ -92,9 +98,13 @@ export class CdkTable implements CollectionViewer {
   /** Set of templates that used as the data row containers. */
   @ContentChildren(CdkRowDef) _rowDefinitions: QueryList<CdkRowDef>;
 
-  constructor(private _changeDetectorRef: ChangeDetectorRef) {
+  constructor(private _differs: IterableDiffers,
+              private _changeDetectorRef: ChangeDetectorRef) {
```
ErinCoughlan: Should both of these be readonly as well?

andrewseguin: I don't think we've set that precedent, but I can talk with Jeremy about adding it to here and other components as well

jelbourn: Yeah, they should be `readonly`; we started the library before `readonly` was a thing and never got in the habit of adding it later

src/lib/core/data-table/data-table.ts
```diff
@@ -76,6 +79,9 @@ export class CdkTable implements CollectionViewer {
    */
   private _columnDefinitionsByName = new Map<string,  CdkColumnDef>();
 
+  /** Differ used to find the changes in the data provided by the data source. */
+  private _dataDiffer: IterableDiffer<any> = null;
```
jelbourn: One thing we should consider doing before this is released is making it `CdkTable<T>` and using the generic throughout

src/lib/core/data-table/data-table.ts
```diff
@@ -92,9 +98,13 @@ export class CdkTable implements CollectionViewer {
...
   @ContentChildren(CdkRowDef) _rowDefinitions: QueryList<CdkRowDef>;
 
-  constructor(private _changeDetectorRef: ChangeDetectorRef) {
+  constructor(private _differs: IterableDiffers,
+              private _changeDetectorRef: ChangeDetectorRef) {
     console.warn('The data table is still in active development ' +
         'and should be considered unstable.');
+
+    // TODO(andrewseguin): Add trackby function input.
+    this._dataDiffer = this._differs.find([]).create();
```
jelbourn: I don't quite follow how this works; why is this `find` necessary? Can you not just create an `IterableDiffer` directly? Why is the DI system involved? 
(probably just needs explanation in comments)

src/lib/core/data-table/data-table.ts
andrewseguin: Added a <T> generic to the CdkTable definition

src/lib/core/data-table/data-table.ts
andrewseguin: Added some comment mentioning that we need to find an iterable differ used for array-based iterables. It's the only public API that lets us grab it, no luck instantiating one myself.

jelbourn:
Just noticed still needs `readonly` per Erin's comment

andrewseguin:
Thanks for catching that, got the readonly in, thought I had changed that already.

Issue #4615, opened 2017-05-17
datepicker: (change) event does not work in mdInput


da45:
#### Bug
change event does not work 

```
<md-input-container>
    <input mdInput [mdDatepicker]="picker" (change)="doSomething($event)">
     <button mdSuffix [mdDatepickerToggle]="picker"></button>
</md-input-container>

<md-datepicker #picker></md-datepicker>    
```

#### What is the expected behavior?
change event must work

#### What is the current behavior?
when choose a date from datepicker no (change) event is triggered  

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular 4.1.2 & material 2.0.0-beta.5

mohi86:
what about this: <md-datepicker #picker (selectedChange)="updateDate()"></md-datepicker>

this is not working either. Here's the code from @angular/material 

MdCalendar.propDecorators = {
    'startAt': [{ type: Input },],
    'startView': [{ type: Input },],
    'selected': [{ type: Input },],
    'minDate': [{ type: Input },],
    'maxDate': [{ type: Input },],
    'dateFilter': [{ type: Input },],
    'selectedChange': [{ type: Output },],
};

What event can I hook into when a date is selected?

da45:
@mohi86 mohi86 
If your method is just to update the date, you can bind it like the following:
[(ngModel)]="myDate" ,
For my case, I really have to call a function to run some complex tasks,

mohi86:
@da45 
Thanks, but I have the same issue, I need to run some calculations when the date is selected. 
I used angular2-material-datepicker but it's not working well with the angular material form elements.

RobertBakic:
the binding: [(ngModel)]="myDate" , works for me if I pick the date in the UI, but date modification made directly in the input field is not taken into account.

RobertBakic:
After further test, the binding problem described in my previous post was related to the localization (French) of the date format (dd/mm/yyyy) instead of (mm/dd/yyyy) wich works fine.

willshowell:
@da45 @mohi86 For now, you can split out the `[(ngModel)]` like this

```html
<input [ngModel]="myDate" (ngModelChange)="updateMyDate($event)">
```
```ts
updateMyDate(newDate) {
  this.myDate = newDate;
  this.runOtherCalculations(newDate);
}
```

https://plnkr.co/edit/aZKLaA5qlZj71AqD5ZFz?p=preview

da45:
[...]

PR #4826, opened 2017-05-26
fix(datepicker): better support for input and change events

mmalerba:
fixes: #4615

### DIFF
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -56,7 +56,7 @@ export const MD_DATEPICKER_VALIDATORS: any = {
 /**
  * An event used for datepicker input and change events. We don't always have access to a native
  * input or change event because the event may have been triggered by the user clicking on the
- * calendar popup. For consistency we always use MdDatepickerInputEvent instead.
+ * calendar popup. For consistency, we always use MdDatepickerInputEvent instead.
  */
 export class MdDatepickerInputEvent<D> {
   /** The new value for the target datepicker input. */
@@ -152,8 +152,10 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
   }
   private _disabled: boolean;
 
+  /** Emits when a `change` event is fired on this `<input>`. */
   @Output() dateChange = new EventEmitter<MdDatepickerInputEvent<D>>();
 
+  /** Emits when an `input` event is fired on this `<input>`. */
   @Output() dateInput = new EventEmitter<MdDatepickerInputEvent<D>>();
 
   /** Emits when the value changes (either due to user input or programmatic change). */
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -53,16 +53,16 @@ export const MD_DATEPICKER_VALIDATORS: any = {
 };
 
 
-/** An event used for datepicker input and change events. */
+/**
+ * An event used for datepicker input and change events. We don't always have access to a native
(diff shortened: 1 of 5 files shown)


src/lib/datepicker/datepicker-input.ts
```diff
@@ -43,6 +44,21 @@ export const MD_DATEPICKER_VALIDATORS: any = {
 };
 
 
+/** An event used for datepicker input and change events. */
```
kara: Nit: Maybe add something to the comment about why you don't just use the regular event objects from (input) and (change).

src/lib/datepicker/datepicker-input.ts
```diff
@@ -152,7 +174,7 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
       Validators.compose([this._minValidator, this._maxValidator, this._filterValidator]);
 
   constructor(
-      private _elementRef: ElementRef,
+      public _elementRef: ElementRef,
```
kara: Have you considered passing the element and the value to `MdDatepickerInputEvent` separately (instead of `this`), so you can keep `_elementRef` private?

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -486,6 +487,81 @@ describe('MdDatepicker', () => {
...
+      let inputEl: HTMLInputElement;;
+
+      beforeEach(async(() => {
+        fixture = TestBed.createComponent(DatepickerWithChangeAndInputEvents);
+        fixture.detectChanges();
+
+        testComponent = fixture.componentInstance;
+        inputEl = fixture.debugElement.query(By.css('input')).nativeElement;
+
+        spyOn(testComponent, 'onChange');
```
kara: Another option would be to use `jasmine.createSpy()` here. It would remove the need for creating the empty functions in lines 720 on.

```ts
testComponent.onChange = jasmine.createSpy('onChange');
testComponent.onInput = jasmine.createSpy('onInput');
...
```

mmalerba: It won't let me just add properties like that, it complains that `onChange` isn't a member of  `testComponent`

mmalerba:
@kara whoops forgot about this for a while, but comments addressed, PTAL

mmalerba:
rebased

src/lib/datepicker/datepicker.spec.ts
kara: Oh, I guess you'd need to add them on the testComponent too anyway. So I guess it doesn't really help.

src/lib/datepicker/datepicker-input.ts
```diff
@@ -52,6 +53,21 @@ export const MD_DATEPICKER_VALIDATORS: any = {
 };
 
 
+/**
+ * An event used for datepicker input and change events. We don't always have access to a native
+ * input or change event because the event may have been triggered by the user clicking on the
+ * calendar popup. For consistency we always use MdDatepickerInputEvent instead.
```
kara: Nit: `For consistency we always` -> `For consistency, we always use`

src/lib/datepicker/datepicker-input.ts
```diff
@@ -134,6 +152,10 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
   }
   private _disabled: boolean;
 
+  @Output() dateChange = new EventEmitter<MdDatepickerInputEvent<D>>();
+
+  @Output() dateInput = new EventEmitter<MdDatepickerInputEvent<D>>();
```
jelbourn: Add user-facing JsDoc for these events

PR #4827, opened 2017-05-26
fix(dialog): componentInstance unavailable in afterClose

### LINKED ISSUE #4815: Dialog Component Instance
In Beta 6 (maybe 4 / 5), I cant seem to access anything from dialogRef.componentInstance.

e.g dialogRef.componentInstance.isSaved would return me the value of isSaved from the component.

componentInstance is null, there is no way to access the inner variables of the dialog.

try using componentInstance after close

I want to know the certain state of a dialog after close.

Material Beta 6, Angular 4

crisbeto:
Fixes the `dialogRef.componentInstance` being cleaned up too early, causing it to be unavailable in the `afterClosed` callback.

Fixes #4815.


rajvengala:
We are currently on 2.0.0-beta.10. componentInstance is null on quite a regular basis for most dialogs. Unfortunately, we could not pin down the cause.

PR #4830, opened 2017-05-26
feat(data-table): re-render when columns change

andrewseguin:


src/lib/core/data-table/data-table.spec.ts
```diff
@@ -118,6 +117,23 @@ describe('CdkTable', () => {
...
+    expect(dataSource.data.length).toBe(3);
+
+    let initialTableContent = [['Column A', 'Column B', 'Column C']];
+    dataSource.data.forEach(rowData => initialTableContent.push([rowData.a, rowData.b, rowData.c]));
+    expect(tableElement).toMatchTableContent(initialTableContent);
+
+    // Remove column_a and swap column_b/column_c.
+    component.columnsToRender = ['column_c', 'column_b'];
+    fixture.detectChanges();
+    fixture.detectChanges();
```
ErinCoughlan: Why do we need two detectChanges? Add a comment?

src/lib/core/data-table/data-table.ts
```diff
@@ -62,14 +63,17 @@ export class CdkTable implements CollectionViewer {
...
+  //   and instead calculate the view on init and scroll.
   /**
    * Stream containing the latest information on what rows are being displayed on screen.
    * Can be used by the data source to as a heuristic of what data should be provided.
    */
   viewChanged =
       new BehaviorSubject<{start: number, end: number}>({start: 0, end: Number.MAX_VALUE});
 
+  /** Stream that emits when a row def has a change to its array of columns to render. */
+  columnsChange = new Observable<void[]>();
```
jelbourn: Name should probably be consistent w/ `viewChanged`; maybe change `viewChanged` to just `viewChange`?

src/lib/core/data-table/data-table.ts
```diff
@@ -112,21 +116,32 @@ export class CdkTable implements CollectionViewer {
     this._columnDefinitions.forEach(columnDef => {
       this._columnDefinitionsByName.set(columnDef.name, columnDef);
     });
+
+    // Get and merge the streams for column changes made to the row defs
+    const rowDefs = this._rowDefinitions.toArray().concat(this._headerDefinition);
```
jelbourn: Expert mode:
```ts
const rowDefs = [...this._rowDefinitions.toArray(), this._headerDefinition];
```

src/lib/core/data-table/data-table.ts
```diff
@@ -112,21 +116,32 @@ export class CdkTable implements CollectionViewer {
...
+      this.renderHeaderRow();
+    });
+
     // TODO(andrewseguin): If the data source is not
     //   present after view init, connect it when it is defined.
     // TODO(andrewseguin): Unsubscribe from this on destroy.
-    this.dataSource.connect(this).subscribe((rowsData: any[]) => {
+    const streams = [this.dataSource.connect(this), this.columnsChange];
+    Observable.combineLatest(streams).subscribe((result: any[]) => {
+      console.log('Rendering all rows');
```
jelbourn: `console.log`

src/lib/core/data-table/data-table.ts
```diff
@@ -112,21 +116,32 @@ export class CdkTable implements CollectionViewer {
...
+      this._headerRowPlaceholder.viewContainer.clear();
+      this.renderHeaderRow();
+    });
+
     // TODO(andrewseguin): If the data source is not
     //   present after view init, connect it when it is defined.
     // TODO(andrewseguin): Unsubscribe from this on destroy.
-    this.dataSource.connect(this).subscribe((rowsData: any[]) => {
+    const streams = [this.dataSource.connect(this), this.columnsChange];
+    Observable.combineLatest(streams).subscribe((result: any[]) => {
```
jelbourn: Can this be more specific than `any`?

src/lib/core/data-table/data-table.ts
```diff
@@ -112,21 +116,32 @@ export class CdkTable implements CollectionViewer {
...
+    this.columnsChange.subscribe(() => {
+      this._headerRowPlaceholder.viewContainer.clear();
+      this.renderHeaderRow();
+    });
+
     // TODO(andrewseguin): If the data source is not
     //   present after view init, connect it when it is defined.
     // TODO(andrewseguin): Unsubscribe from this on destroy.
-    this.dataSource.connect(this).subscribe((rowsData: any[]) => {
+    const streams = [this.dataSource.connect(this), this.columnsChange];
```
jelbourn: More specific name than `streams`?

src/lib/core/data-table/data-table.ts
DevVersion: We need a lint rule for this.

src/lib/core/data-table/data-table.ts
andrewseguin: I don't think it can be since I cannot type the individual elements of the array.

src/lib/core/data-table/data-table.ts
andrewseguin: Going with `renderChange` in that it causes render changes? Naming is hard

src/lib/core/data-table/data-table.ts
andrewseguin: We can pretend this didn't make it through to review

src/lib/core/data-table/data-table.spec.ts
andrewseguin: Good question, looks like whatever was needing it before is gone. Just need one now

src/demo-app/data-table/data-table-demo.ts
```diff
@@ -22,4 +22,13 @@ export class DataTableDemo {
     let distanceFromMiddle = Math.abs(50 - progress);
     return distanceFromMiddle / 50 + .3;
   }
+
+  toggleColorColumn() {
+    let colorColumnIndex = this.propertiesToDisplay.findIndex((col: string) => col === 'color');
+    if (colorColumnIndex == -1) {
+      this.propertiesToDisplay.push('color');
```
ErinCoughlan: Possibly later, but I'm always cautious about passing strings around because I'm great at mistyping them. Perhaps a best practice would be to define a columns type for the demo:

type columns = 'color'|'progress'; // etc.

And then we'll have a little bit of type safety with the propertiesToDisplay. It would also make us less likely to hit the no column definition error.

andrewseguin: I think that sounds good, I'll add that in.

src/lib/core/data-table/data-table.spec.ts
```diff
@@ -111,13 +110,27 @@ describe('CdkTable', () => {
...
     dataSource.data.forEach(rowData => changedTableContent.push([rowData.a, rowData.b, rowData.c]));
     expect(tableElement).toMatchTableContent(changedTableContent);
   });
+
+  it('should be able to dynamically change the columns for header and rows', () => {
+    expect(dataSource.data.length).toBe(3);
+
+    let initialTableContent = [['Column A', 'Column B', 'Column C']];
+    dataSource.data.forEach(rowData => initialTableContent.push([rowData.a, rowData.b, rowData.c]));
+    expect(tableElement).toMatchTableContent(initialTableContent);
```
jelbourn: You typically want to avoid having logic in unit tests like this, instead preferring to hard-code the values. I think this all should be
```ts
expect(tableElement).toMatchTableContent([
  ['Column A', 'Column B', 'Column C'],
  [data[0].a, data[0].b, data[0].c],
  [data[1].a, data[1].b, data[2].c],
  [data[2].a, data[2].b, data[1].c],
]);
```
Which is also much easier to visually see what's going on with the test

src/lib/core/data-table/data-table.ts
```diff
@@ -99,34 +103,44 @@ export class CdkTable implements CollectionViewer {
...
+      this._headerRowPlaceholder.viewContainer.clear();
+      this.renderHeaderRow();
+    });
+
     // TODO(andrewseguin): If the data source is not
     //   present after view init, connect it when it is defined.
     // TODO(andrewseguin): Unsubscribe from this on destroy.
-    this.dataSource.connect(this).subscribe((rowsData: any[]) => {
+    const renderChange = [this.dataSource.connect(this), this.columnsChange];
+    Observable.combineLatest(renderChange).subscribe(([data]) => {
```
jelbourn: Why are you destructuring the function arguments here?

src/lib/core/data-table/data-table.ts
```diff
@@ -62,14 +63,17 @@ export class CdkTable implements CollectionViewer {
...
   /**
    * Stream containing the latest information on what rows are being displayed on screen.
    * Can be used by the data source to as a heuristic of what data should be provided.
    */
-  viewChanged =
+  viewChange =
       new BehaviorSubject<{start: number, end: number}>({start: 0, end: Number.MAX_VALUE});
 
+  /** Stream that emits when a row def has a change to its array of columns to render. */
+  columnsChange = new Observable<void[]>();
```
jelbourn: `void[]` doesn't make sense- it's "array of nothings". It should either be just `void` or capture whatever argument is being emitted

andrewseguin: I agree but it makes Typescript happy. The merged streams are each Observable<void> and combined make up Observable<void[]>

src/lib/core/data-table/data-table.ts
andrewseguin: The only part of the observable we want is the first element in the array result, so just grabbing it here. If it looks too odd I can change it back to (result: any[]) and then grab result[0] for our data.

PR #4833, opened 2017-05-26
build: serve e2e-app in aot mode

DevVersion:
Starts building the e2e-app in AOT mode

This should make our e2e tests on  the CI faster and also should be a more advanced testing of AOT compatibility.

**Note**: As this also makes the e2e-app smaller in payload I will try to find a way to use a RxJS bundle (would speed-up the serving a lot)

### DIFF
--- a/src/e2e-app/main-aot.ts
+++ b/src/e2e-app/main-aot.ts
@@ -1,0 +1,4 @@
+import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
+import {E2eAppModuleNgFactory} from './e2e-app-module.ngfactory';
+
+platformBrowserDynamic().bootstrapModuleFactory(E2eAppModuleNgFactory);
--- a/src/e2e-app/main-aot.ts
+++ b/src/e2e-app/main-aot.ts
@@ -1,4 +1,0 @@
-import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
-import {E2eAppModuleNgFactory} from './e2e-app-module.ngfactory';
-
-platformBrowserDynamic().bootstrapModuleFactory(E2eAppModuleNgFactory);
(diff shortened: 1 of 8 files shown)


src/e2e-app/main-aot.ts
```diff
@@ -0,0 +1,4 @@
+import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
+import {E2eAppModuleNgFactory} from './e2e-app-module.ngfactory';
+
+platformBrowserDynamic().bootstrapModuleFactory(E2eAppModuleNgFactory);
```
jelbourn: Shouldn't this just be `platformBrowser` (without the `Dynamic`)?

(same for demo-app, now that I notice it)

src/e2e-app/main-aot.ts
```diff
@@ -0,0 +1,4 @@
+import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
```
jelbourn: Instead of having a separate `main-aot.ts`, we should just make this the main `main.ts`; nobody should need to ever run this in JIT mode

DevVersion: I agree. Done!

src/e2e-app/main-aot.ts
DevVersion: Yeah I think @josephperrott also had an issue about it but I didn't assign it to myself so I lost track of it. Changed it now.

mmalerba:
please rebase

DevVersion:
@mmalerba Done. This also includes the fix for the CI failure.

> CI failure appears because we renamed the `constants.ts` file but another PR, that got merged afterwards,  introduced a new file that still imports form `constants.ts`.

PR #4836, opened 2017-05-26
build: add most components to kitchen-sink

jelbourn:
Fixes a handful of issues that prevent some components from working.

Checkbox and button-toggles don't work due to [an issue in platform-server](https://github.com/angular/angular/issues/17050).

Tabs don't work because they try to set a `style` property directly on initialization; this will need a more in-depth follow-up (I didn't look into that deeply). 

Related to #308

### DIFF
--- a/src/lib/core/a11y/interactivity-checker.ts
+++ b/src/lib/core/a11y/interactivity-checker.ts
@@ -48,6 +48,10 @@ export class InteractivityChecker {
    * @returns Whether the element is tabbable.
    */
   isTabbable(element: HTMLElement): boolean {
+    // Nothing is tabbable on the the server 😎
+    if (!this._platform.isBrowser) {
+      return false;
+    }
 
     let frameElement = getWindow(element).frameElement as HTMLElement;
 
--- a/src/lib/slide-toggle/slide-toggle.ts
+++ b/src/lib/slide-toggle/slide-toggle.ts
@@ -20,7 +20,8 @@ import {
   FocusOriginMonitor,
   HammerInput,
   MdRipple,
-  RippleRef
+  RippleRef,
+  Platform,
 } from '../core';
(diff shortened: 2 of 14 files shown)


src/lib/core/a11y/interactivity-checker.ts
```diff
@@ -48,6 +48,10 @@ export class InteractivityChecker {
    * @returns Whether the element is tabbable.
    */
   isTabbable(element: HTMLElement): boolean {
+    // Nothing is tabbable on the the server 😎
```
DevVersion: We need a lint rule for that! 😄

src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -312,9 +314,13 @@ class SlideToggleRenderer {
   /** Whether the thumb is currently being dragged. */
   dragging: boolean = false;
 
-  constructor(private _elementRef: ElementRef) {
-    this._thumbEl = _elementRef.nativeElement.querySelector('.mat-slide-toggle-thumb-container');
-    this._thumbBarEl = _elementRef.nativeElement.querySelector('.mat-slide-toggle-bar');
+  constructor(private _elementRef: ElementRef, platform: Platform) {
+    // We only need to interact with these elements when we're on the browser, so only grab
+    // the refernece in that case.
```
DevVersion: Typo for `reference`.

> Also we are already in the custom renderer? I suppose the whole custom renderer approach will be addressed in the future at some point.

> Just thought that we could not create the Renderer if we are not in the browser

jelbourn: That's the long-term plan, but that will require the secondary `@angular/material/server` package

DevVersion: Okay sounds good

jelbourn:
Still trying to fix these test failures, which happen inconsistently locally

jelbourn:
All failures resolved

PR #4839, opened 2017-05-27
fix(datepicker): focus trap not working inside popup

crisbeto:
Fixes focus trapping not working inside the datepicker in popup mode due to the A11yModule not being imported.

Relates to #4804.


mmalerba:
LGTM

PR #4840, opened 2017-05-27
chore(data-table): hide stability warning in jasmine

DevVersion:
* No longer shows the data-table stability warning inside of jasmine tests.

### DIFF
--- a/src/lib/core/data-table/data-table.ts
+++ b/src/lib/core/data-table/data-table.ts
@@ -101,8 +101,12 @@ export class CdkTable<T> implements CollectionViewer {
 
   constructor(private readonly _differs: IterableDiffers,
               private readonly _changeDetectorRef: ChangeDetectorRef) {
-    console.warn('The data table is still in active development ' +
+    // Show the stability warning of the data-table only if it doesn't run inside of jasmine.
+    // This is just temporary and should reduce warnings when running the tests.
+    if (!(typeof window !== 'undefined' && window['jasmine'])) {
+      console.warn('The data table is still in active development ' +
         'and should be considered unstable.');
+    }
 
     // TODO(andrewseguin): Add trackby function input.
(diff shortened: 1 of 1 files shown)


andrewseguin:
Looping in Jeremy, he requested the data table warning to be a bit more intrusive to make it clear that it's not yet ready

jelbourn:
This isn't part of the sanity checks

DevVersion:
Yeah it isn't really a sanity check but it allows us to disable those warnings in tests.

Having the warning super-intrusive just spams the whole tests and doesn't really help at all.

I initially only wanted to show the warning in dev-mode only but after chatting with @crisbeto I agreed that it would be nice if developers can disable the warning though

jelbourn:
I'd rather add a check for `window.jasmine` to disable it so that we just don't log it during our own tests

DevVersion:
That would work too. But the issue doesn't just appear in Material. Every developer that writes tests using Material will get all of those data-table errors.

The sanity checks solution would be kind of temporary until this warning will be removed anyways. But for now it's a super intrusive warning.

jelbourn:
@DevVersion the warning is in the constructor of the table- they'll only get it if they use the table, which is exactly the point

DevVersion:
@jelbourn Yeah that's clear. But as I mentioned everybody that tests for example a component that contains a `md-data-table` will get the error. Similar as we do by explicitly testing the `md-data-table`.

DevVersion:
@jelbourn I'm fine doing the `window.jasmine` thing since you said that nobody should use the data-table at all yet (and it's exposed for the demo-app). Let me know if I should proceed with that.

DevVersion:
@andrewseguin @jelbourn Made the requested changes. Please review again.

andrewseguin:
Can you rebase?

DevVersion:
@andrewseguin Done

Issue #2612, opened 2017-01-12
MdMenu : keep menu open on item click


Robouste:
#### Bug, feature request, or proposal:
feature request

#### What is the expected behavior?
It would be great to have an option to keep the menu open when we click a menu item 

#### What is the current behavior?
On item menu click, the menu closes

#### What are the steps to reproduce?
[Plunkr](http://plnkr.co/edit/NK36sG)

#### What is the use-case or motivation for changing an existing behavior?
Putting checkboxes in the menu for example

#### Which versions of Angular, Material, OS, browsers are affected?
All

nayfin:
I would like to utilize this feature to allow menu to take text inputs. Perhaps some new directives:

`opens-on="focus | hover | click"` 
`closes-on="defocus | hover-off | click-off | select-item"`

This would let us get very specific with the menu behavior. Would need to be able to take one or all values.

nayfin:
I just found this in the ReadMe:

`Not yet implemented
- ‘prevent-close' option, to turn off automatic menu close when clicking outside the menu ` 

Which isn't quite what we are looking for, but it's close and a good indication that similar features are being considered. 

From what I've read of the source we just need to prevent click events on `md-menu-items` from propagating up to the menu. If anyone has some pro tips before I get started I'm happy to hear them. I'll try to put together a splunkr over the weekend and submit a PR once it's working.

nayfin:
@Robouste [Here's an ugly workaround](http://plnkr.co/edit/6qzKZX?p=preview). I'll turn it into a directive soon and update. For some reason using this with `md-menu-item` keeps menu open but doesn't allow inputs or checkboxes to register clicks or become focused. Take out the `md-menu-item` directive and everything works fine, just doesn't look quite as nice.

Robouste:
@nayfin Thank you, it seems to do the trick. But I choose the bootstrap dropdown, more customizable (for now !)

randallmeeker:
I have an input in my menu and this worked find
`<input class="input-log" (click)="$event.stopPropagation()" type="text" />`

ccadieux:
Being able to turn this off would be great. We implemented a menu very similar to the google inbox account menu and ran into this issue. We had to add a click handler and stopPropagation on the entire menu.

jelbourn:
[...]

PR #4842, opened 2017-05-27
feat(menu): add disableClose option

crisbeto:
* Adds the ability to prevent users from closing a menu via the `disableClose` option.
* Fixes an exception that was being thrown when tabbing through menu items multiple times.

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -1,5 +1,3 @@
-// TODO(kara): prevent-close functionality
-
 import {
   AfterContentInit,
   Component,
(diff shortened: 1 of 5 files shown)


willshowell:
I believe this fixes https://github.com/angular/material2/issues/2612

jelbourn:
Chatted w/ @kara about this and we decided that we don't want to add this as a feature to md-menu. This is largely because turning it on would be detrimental to the a11y experience (see https://www.w3.org/TR/wai-aria-practices-1.1/#menu), and we want to generally avoid adding APIs that go counter to a11y best practices.

nolisj:
That link (https://www.w3.org/TR/wai-aria-practices-1.1/#menu) doesn't seem to indicate that we are not allowed to retain the menu when we click one of its child menus. On the contrary, the last statement below is an indication that it is allowed.

"A menu is usually opened, or made visible, by activating a menu button, choosing an item in a menu that opens a sub menu, or by invoking a command, such as Shift + F10 in Windows, that opens a context specific menu. When a user activates a choice in a menu, the menu usually closes unless the choice opened a submenu."

alua1612:
[](url)

PR #4843, opened 2017-05-27
fix(menu): not handling keyboard events when opened by mouse

### LINKED ISSUE #4991: menu up/down arrow does not navigate the overlay
Menu does not properly navigate with up/down arrows, also enter key does not work since you can't navigate with the up/down.
This work's just fine in autocomplete cause it is handling the up/down/enter key but this must have just been missed in menu.

Should be able to open a menu and click down arrow to the first element, second and than go back up to the first for example.

Current behavior is up/down arrow will actually scroll the page.

https://material.angular.io/components/component/menu

Can do it with the basic menu example above.

Users should be able to navigate the menu via keyboard.

Latest in Material2

I would be glad to make the code updates for you guys let me know.

crisbeto:
Fixes the menu keyboard interactions not working when it is opened by a click.

Fixes #4991.

### DIFF
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -225,10 +225,16 @@ export class MatMenuTrigger implements AfterContentInit, OnDestroy {
     this._setMenuElevation();
     this._setIsMenuOpen(true);
 
-    // Should only set focus if opened via the keyboard, so keyboard users can
-    // can easily navigate menu items. According to spec, mouse users should not
-    // see the focus style.
-    if (!this._openedByMouse) {
+    // If the menu was opened by mouse, we focus the root node, which allows for the keyboard
+    // interactions to work. Otherwise, if the menu was opened by keyboard, we focus the first item.
+    if (this._openedByMouse) {
+      let rootNode = this._overlayRef!.overlayElement.firstElementChild as HTMLElement;
+
+      if (rootNode) {
+        this.menu.resetActiveItem();
+        rootNode.focus();
+      }
+    } else {
       this.menu.focusFirstItem();
     }
   }
--- a/e2e/components/menu-e2e.spec.ts
+++ b/e2e/components/menu-e2e.spec.ts
@@ -97,9 +97,9 @@ describe('menu', () => {
       expectFocusOn(page.items(0));
     });
 
-    it('should not focus the first item when opened with mouse', () => {
+    it('should focus the panel when opened by mouse', () => {
       page.trigger().click();
-      expectFocusOn(page.trigger());
+      expectFocusOn(page.menu());
     });
 
     it('should focus subsequent items when down arrow is pressed', () => {
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -206,6 +206,14 @@ export class MatMenu implements AfterContentInit, MatMenuPanel, OnDestroy {
   }
 
   /**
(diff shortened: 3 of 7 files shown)


src/lib/menu/menu-trigger.ts
```diff
@@ -104,6 +107,14 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
       this._createOverlay();
       this._overlayRef.attach(this._portal);
       this._subscribeToBackdrop();
+
+      this._globalKeydownHandler = this._renderer.listen('document', 'keydown',
```
jelbourn: How will this global listener play with, say, a menu inside of a dialog?

crisbeto: Hmm, it'll close both the menu and the dialog. What about focusing the menu panel, instead of the first option, when the menu is opened by mouse?

jelbourn: The options are focused for a11y

crisbeto: Yes, but it seems pretty deliberate that we're not shifting focus when the user opened it by mouse. If we want to keep that behavior, we could opt to move focus to the panel, if the menu was opened by mouse and focus the first option if it was opened by keyboard.

jelbourn: @kara do you remember why we don't move focus to the options when opening via mouse? If it was just to prevent the focus style from showing, we could probably just do that through css.

kara: I believe it was just for the visual treatment. I don't think there's any reason not to focus the options in itself.

crisbeto: How should I proceed in this case? Given that we also have https://github.com/angular/material2/issues/4991, should I just enable the keyboard navigation to matter how you opened the menu?

jelbourn: We should make it always focus the options no matter how you open the panel, but only show the initial focus style (before subsequent key presses) when it's triggered by keyboard

crisbeto:
@jelbourn @kara, I got around to reworking this. If the menu was opened by mouse, the panel will receive focus, which allows the user to use escape and go into the options by pressing the arrow keys.

kara:

crisbeto:
Done @kara.

src/lib/menu/menu-trigger.ts
```diff
@@ -163,10 +163,15 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
     this._setIsMenuOpen(true);
 
-    // Should only set focus if opened via the keyboard, so keyboard users can
-    // can easily navigate menu items. According to spec, mouse users should not
-    // see the focus style.
-    if (!this._openedByMouse) {
+    // If the menu was opened by mouse, we focus the root node, which allows for the keyboard
+    // interactions to work. Otherwise, if the menu was opened by keyboard, we focus the first item.
+    if (this._openedByMouse) {
+      let rootNode = this._overlayRef.overlayElement.firstElementChild as HTMLElement;
```
kara: The _overlayRef type is sometimes null, so the strictNullCheck in the build is complaining. Add a `!` ?

crisbeto:
@kara can you take another look?

crisbeto:
@kara Implemented the approach as discussed.

kara:
@crisbeto Sorry for the delayed review on this. Do you mind rebasing again?

kara:
Note: needs internal updates

PR #4847, opened 2017-05-27
Documenting action buttons align

theely:
Documentation for the action buttons alignments was missing. Also I would suggest to add by default the align="end" property to be more compliment with the Material design guidelines.


src/lib/dialog/dialog.md
```diff
@@ -61,13 +61,14 @@ Several directives are available to make it easier to structure your dialog cont
 | `md-dialog-title`     | \[Attr] Dialog title, applied to a heading element (e.g., `<h1>`, `<h2>`)                                     |
 | `<md-dialog-content>` | Primary scrollable content of the dialog                                                                      |
 | `<md-dialog-actions>` | Container for action buttons at the bottom of the dialog                                                      |
+| `align`               | \[Attr] for action buttons aligment end or center                                                             |
```
jelbourn: This isn't the right place for this; this table lists the directives that can be used inside of a dialog. I'm not really sure this API is noteworthy enough to be in the overview; it should be included in the API reference, though.

PR #4855, opened 2017-05-28
feat(overlay): more flexible scroll strategy API and ability to define/override custom strategies

crisbeto:
* Refactors the overlay setup to allow for scroll strategies to be passed in by name, instead of by instance.
* Handles the scroll strategy dependency injection automatically.
* Adds an API for registering custom scroll strategies and overriding the existing ones.
* Adds a second parameter to the `attach` method, allowing for a config object to be passed in.
* Throws an error if there's an attempt to attach a scroll strategy multiple times. This is mostly a sanity check to ensure that we don't cache the scroll strategy instances.

Relates to #4093.

### DIFF
--- a/src/lib/core/overlay/overlay-state.ts
+++ b/src/lib/core/overlay/overlay-state.ts
@@ -1,9 +1,7 @@
 import {PositionStrategy} from './position/position-strategy';
 import {LayoutDirection} from '../rtl/dir';
-import {ScrollStrategyOption} from './scroll/scroll-strategy-options';
+import {ScrollStrategy} from './scroll/scroll-strategy';
 
-export type OverlayStateScrollStrategy = {strategy: ScrollStrategyOption; config: any} |
-                                         ScrollStrategyOption;
 
 /**
  * OverlayState is a bag of values for either the initial configuration or current state of an
@@ -14,7 +12,7 @@ export class OverlayState {
   positionStrategy: PositionStrategy;
 
   /** Strategy to be used when handling scroll events while the overlay is open. */
-  scrollStrategy: OverlayStateScrollStrategy;
+  scrollStrategy: ScrollStrategy;
 
   /** Custom class to add to the overlay pane. */
   panelClass: string = '';
--- a/src/lib/core/overlay/overlay-state.ts
+++ b/src/lib/core/overlay/overlay-state.ts
@@ -1,9 +1,9 @@
 import {PositionStrategy} from './position/position-strategy';
 import {LayoutDirection} from '../rtl/dir';
-import {ScrollStrategy} from './scroll/scroll-strategy';
-import {NoopScrollStrategy} from './scroll/noop-scroll-strategy';
+import {ScrollStrategyOption} from './scroll/scroll-strategy-options';
 
-export type OverlayStateScrollStrategy = string | {name: string; config: any};
+export type OverlayStateScrollStrategy = {strategy: ScrollStrategyOption; config: any} |
+                                         ScrollStrategyOption;
 
 /**
  * OverlayState is a bag of values for either the initial configuration or current state of an
@@ -14,7 +14,7 @@ export class OverlayState {
   positionStrategy: PositionStrategy;
 
   /** Strategy to be used when handling scroll events while the overlay is open. */
-  scrollStrategy: OverlayStateScrollStrategy = 'noop';
+  scrollStrategy: OverlayStateScrollStrategy;
 
   /** Custom class to add to the overlay pane. */
   panelClass: string = '';
--- a/src/lib/core/overlay/overlay-state.ts
+++ b/src/lib/core/overlay/overlay-state.ts
@@ -3,6 +3,7 @@ import {LayoutDirection} from '../rtl/dir';
 import {ScrollStrategy} from './scroll/scroll-strategy';
 import {NoopScrollStrategy} from './scroll/noop-scroll-strategy';
 
+export type OverlayStateScrollStrategy = string | {name: string; config: any};
 
 /**
  * OverlayState is a bag of values for either the initial configuration or current state of an
@@ -13,7 +14,7 @@ export class OverlayState {
   positionStrategy: PositionStrategy;
 
   /** Strategy to be used when handling scroll events while the overlay is open. */
-  scrollStrategy: ScrollStrategy = new NoopScrollStrategy();
+  scrollStrategy: OverlayStateScrollStrategy = 'noop';
 
   /** Custom class to add to the overlay pane. */
   panelClass: string = '';
--- a/src/lib/core/overlay/overlay.ts
+++ b/src/lib/core/overlay/overlay.ts
@@ -86,13 +86,23 @@ export class Overlay {
    * @param state
    */
   private _createOverlayRef(pane: HTMLElement, state: OverlayState): OverlayRef {
-    let scrollStrategyName = typeof state.scrollStrategy === 'string' ?
-      state.scrollStrategy :
-      state.scrollStrategy.name;
-
-    let scrollStrategy = this._scrollStrategyOptions.get(scrollStrategyName);
+    let scrollStrategy = this._createScrollStrategy(state);
     let portalHost = this._createPortalHost(pane);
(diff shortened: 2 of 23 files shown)


crisbeto:
@jelbourn updated the approach as discussed.

src/lib/core/overlay/scroll/scroll-strategy-options.ts
```diff
@@ -0,0 +1,31 @@
...
+  constructor(
+    private _scrollDispatcher: ScrollDispatcher,
+    private _viewportRuler: ViewportRuler) { }
+
+  get(strategy: string): ScrollStrategy {
+    switch (strategy) {
+      case 'reposition': return new RepositionScrollStrategy(this._scrollDispatcher);
+      case 'close': return new CloseScrollStrategy(this._scrollDispatcher);
+      case 'noop': return new NoopScrollStrategy();
+      case 'block': return new BlockScrollStrategy(this._viewportRuler);
```
jelbourn: This way is still string-based; you can't get IDE completion or correction for the strategies available, and they can't be minified.

What I had been thinking was having each of these be properties. What's the value for those properties? Doing something like this would work....
```ts
export type ScrollStrategyOption = () => ScrollStrategy;

export class ScrollStrategyOptions {
  // The class symbol acts as a unique token for the strategy option.
  noop: ScrollStrategyOption = () => new NoopScrollStrategy();
  close: ScrollStrategyOption = () => new CloseScrollStrategy(this._scrollDispatcher);
  block: ScrollStrategyOption = () => new BlockScrollStrategy(this._viewportRuler);
  reposition: ScrollStrategyOption = () => new RepositionScrollStrategy(this._scrollDispatcher);

  getStrategyInstance(option: ScrollStrategyOption) {
    return option();
  }
}


// When using the options
overlayState.scrollStrategy = overlay.scrollStratgies.block;
```

I originally had them as the classes (`BlockScrollStrategy`) with a `switch`, but realized you could cut out that step completely if you treat the properties as the map to the creation directly. I like this approach because, to the user, each option is some opaque thing called `ScrollStrategyOption`, but on the inside we know it's a factory function.  

When someone wants to customize the strategies, they would do this:
```ts
export class SuperScrollStrategyOptions extends ScrollStrategyOptions {
  // overwrite block strategy
  block: ScrollStrategyOption = () => new SuperBlockScrollStrategy(/* ... */);

  // define a custom strategy
  explode: ScrollStrategyOption = () => new SuperExploseScrollStrategy(/* ... */);
}
```

crisbeto: Nice, that's even better. I've updated the PR.

src/lib/core/overlay/overlay.ts
```diff
@@ -34,9 +35,10 @@ export class Overlay {
   constructor(private _overlayContainer: OverlayContainer,
               private _componentFactoryResolver: ComponentFactoryResolver,
               private _positionBuilder: OverlayPositionBuilder,
+              private _scrollStrategyOptions: ScrollStrategyOptions,
```
jelbourn: Expose `scrollStrategies` as a public property on `Overlay` so each component won't have to inject it separately?

src/lib/core/overlay/overlay-state.ts
```diff
@@ -1,8 +1,9 @@
 import {PositionStrategy} from './position/position-strategy';
 import {LayoutDirection} from '../rtl/dir';
-import {ScrollStrategy} from './scroll/scroll-strategy';
-import {NoopScrollStrategy} from './scroll/noop-scroll-strategy';
+import {ScrollStrategyOption} from './scroll/scroll-strategy-options';
 
+export type OverlayStateScrollStrategy = {strategy: ScrollStrategyOption; config: any} |
```
jelbourn: I just realized that we lose out on having config types specific to the strategy with this approach. So......


What if we tweak the options setup slightly so that it's something like...
```ts
// in scroll-strategy-option.ts
export interface ScrollStrategyOption {
  _create: () => ScrollStrategy;
}

// in block-scroll-strategy.ts
export class BlockScrollStrategyOption {
  _config = new BlockScrollStrategyConfig();
  _create() {
    return new BlockScrollStrategy(..., config)
  }

  withExtraOption(value) {
    this._config.extraOption = value;
    return this;
  }
}

// in scroll-strategy-options.ts
export class ScrollStrategyOptions {
  constructor(private _scrollDispatcher: ScrollDispatcher, private _viewportRuler: ViewportRuler) { }

  noop: ScrollStrategyOption = basicOption(() => new NoopScrollStrategy());
  close: ScrollStrategyOption = basicOption(() => new CloseScrollStrategy(this._scrollDispatcher));
  block: = new BlockScrollStrategyOption(...);
  reposition: ScrollStrategyOption = basicOption(() => new RepositionScrollStrategy(this._scrollDispatcher));

  basicOption(factoryFn): ScrollStrategyOption {
    return {_create: factoryFn};
  }
}
```

For someone just consuming the options, this is identical but with a nicer interface for the configuration. For people creating a custom strategy, they just have to create this new builder option class, which I think is worth it when capturing the configuration. 

Sorry for the churn- I keep thinking of new things.

crisbeto:
@jelbourn added the discussed changes.

src/lib/core/overlay/scroll/scroll-strategy-options.ts
```diff
@@ -0,0 +1,30 @@
...
+  constructor(
+    private _scrollDispatcher: ScrollDispatcher,
+    private _viewportRuler: ViewportRuler) { }
+
+  noop = () => new NoopScrollStrategy();
+  close = () => new CloseScrollStrategy(this._scrollDispatcher);
+  block = () => new BlockScrollStrategy(this._viewportRuler);
+  reposition = (config?: RepositionScrollStrategyConfig) => {
+    return new RepositionScrollStrategy(this._scrollDispatcher, config);
+  }
```
jelbourn: Omit the return?
```ts
  reposition = (config?: RepositionScrollStrategyConfig) =>
      new RepositionScrollStrategy(this._scrollDispatcher, config);
```

src/lib/core/overlay/scroll/scroll-strategy-options.ts
```diff
@@ -0,0 +1,30 @@
...
+import {ViewportRuler} from '../position/viewport-ruler';
+import {
+  RepositionScrollStrategy,
+  RepositionScrollStrategyConfig,
+} from './reposition-scroll-strategy';
+
+
+/**
+ * Factory that instantiates scroll strategies. Provides the built-in `reposition`, `close`,
+ * `noop` and `block` strategies by default.
```
jelbourn: How about
```ts
/**
 * Options for how an overlay will handle scrolling.
 *
 * Users can provide a custom value for `ScrollStrategyOptions` to replace the default
 * behaviors. This class primarily acts as a factory for ScrollStrategy instances.
 */
```

src/lib/core/overlay/scroll/scroll-strategy-options.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+@Injectable()
+export class ScrollStrategyOptions {
+  constructor(
+    private _scrollDispatcher: ScrollDispatcher,
+    private _viewportRuler: ViewportRuler) { }
+
+  noop = () => new NoopScrollStrategy();
```
jelbourn: We should add a public JsDoc for each for these, e.g.
```ts
/** Do nothing on scroll. */

/** Close the overlay on scroll. */
```

src/lib/core/overlay/scroll/scroll-strategy.ts
```diff
@@ -9,3 +9,10 @@ export interface ScrollStrategy {
   disable: () => void;
   attach: (overlayRef: OverlayRef) => void;
 }
+
+/**
+ * Returns an error to be thrown when attempting to attach an already-attached scroll strategy.
+ */
+export function getMdScrollStrategyAlreadyAttachedError(): Error {
+  return new Error(`Scroll strategy has already been attached.`);
```
jelbourn: quirk: the `new` isn't actually necessary for `Error`

crisbeto:
Feedback has been addressed.

PR #4860, opened 2017-05-29
Clear duration timeout after dismiss

### LINKED ISSUE #4859: SnackBar does not close zone when dismissed
#### Bug:

SnackBar should close duration zone when dismissed, so protractor is not blocked.

When calling MdSnackBarRef.prototype.dismiss(), the timeout reference by [setTimeout](https://github.com/angular/material2/blob/master/src/lib/snack-bar/snack-bar.ts#L82) should be cleared to close the angular zone.

The zone stays open for the duration of the timeout.

[Plunker (watch the zone state)](http://plnkr.co/edit/e3Kq5D4ftj2zko801iuN?p=preview)

@angular/material@2.0.0-beta.5

svi3c:
Fixes #4859

### DIFF
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -39,6 +39,12 @@ export class MdSnackBarRef<T> {
   /** Subject for notifying the user that the snack bar action was called. */
   private _onAction: Subject<any> = new Subject();
 
+  /**
+   * Timeout ID for the duration setTimeout call. Used to clear the timeout if the snackbar is
+   * dismissed before the duration passes.
+   */
+  private _durationTimeoutId: number;
+
   constructor(instance: T,
               containerInstance: MdSnackBarContainer,
               private _overlayRef: OverlayRef) {
@@ -55,6 +61,12 @@ export class MdSnackBarRef<T> {
     if (!this._afterClosed.closed) {
       this.containerInstance.exit();
     }
+    clearTimeout(this._durationTimeoutId);
+  }
+
+  /** Dismisses the snack bar after some duration */
+  _dismissAfter(duration: number): void {
+    this._durationTimeoutId = setTimeout(() => this.dismiss(), duration);
   }
 
   /** Marks the snackbar action clicked. */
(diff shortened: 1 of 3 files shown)


svi3c:
I signed it!

svi3c:
@jelbourn from my point of view this is ready to review.

src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -31,6 +31,9 @@ export class MdSnackBarRef<T> {
   /** Subject for notifying the user that the snack bar action was called. */
   private _onAction: Subject<any> = new Subject();
 
+  /** Timeout handle for the dismiss timeout cleanup. */
+  private _dismissTimeoutHandle: number;
```
jelbourn: I would call this `durationTimeoutId` with comment:
```ts
/**
 * Timeout ID for the duration setTimeout call. Used to clear the timeout if the snackbar is 
 * dismissed before the duration passes.
 */
private _durationTimeoutId: number;
```

src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -47,6 +50,12 @@ export class MdSnackBarRef<T> {
     if (!this._afterClosed.closed) {
       this.containerInstance.exit();
     }
+    clearTimeout(this._dismissTimeoutHandle);
+  }
+
+  /** Dismisses the snack bar after some duration */
+  dismissAfter(duration: number): void {
```
jelbourn: This should be an private method. If it were a public API and someone called it in combination with setting a duration, the earlier timeout ID would be lost and the issue this PR aims to resolve would still occur.

svi3c:
Hi @jelbourn,

Thanks for reveiw.

I did the requested changes, but the tests seem to fail, since [`flush()`](https://angular.io/api/core/testing/flush) is missing in CI.
I'm not sure why this is the case.
I first thought, that the old version of angular 4 in the yarn.lock might be the issue, but the travis job uses npm to install the dependencies.
When I run `npm i`, I get `@angular/core@4.2.3` installed and it has the `flush()` function.

Do you have any idea how to fix this or test the behavior differently?

jelbourn:
@svi3c 
```ts
  it('should clear the dismiss timeout when dismissed before timeout expiration', fakeAsync(() => {
    let config = new MdSnackBarConfig();
    config.duration = 1000;
    snackBar.open('content', 'test', config);

    setTimeout(() => snackBar.dismiss(), 500);

    tick(600);
    viewContainerFixture.detectChanges();
    flushMicrotasks();

    expect(viewContainerFixture.isStable()).toBe(true);
  }));
```

svi3c:
Hi @jelbourn,
the changes are applied.

PR #4867, opened 2017-05-29
fix(focus-trap): focus initial element when zone stabilizes

### LINKED ISSUE #4864: md-input doesn't behave correct if initially focused in md-dialog

Bug

2017-05-29 15_55_32-plunker

2017-05-29 15_56_13-plunker

http://plnkr.co/edit/79nFR1?p=preview

Focused md-input should look the same on all occasions

4.0.3 / beta6 / at least Firefox & Chrome

Placeholder is floating as soon as you type. But there will be no animation. Behaviour looks and feels a bit "out of place".

crisbeto:
* Focuses the first element when the zone stabilizes, instead of when the microtask queue is empty. This avoids issues where the element might be focused before Angular is done doing change detection.
* Only runs the `onStable` callback if the zone is unstable.
* Avoids an extra DOM lookup by combining a couple of selectors.

Fixes #4864.

**Note:** Since change detection is all being run manually in unit tests, I wasn't able to write a test that would fail properly when we expect it to.

### DIFF
--- a/src/lib/core/a11y/focus-trap.ts
+++ b/src/lib/core/a11y/focus-trap.ts
@@ -81,24 +81,28 @@ export class FocusTrap {
     });
   }
 
+  /**
+   * Waits for the zone to stabilize, then either focuses the first element that the
+   * user specified, or the first tabbable element..
+   */
   focusInitialElementWhenReady() {
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => this.focusInitialElement());
+    this._executeOnStable(() => this.focusInitialElement());
   }
 
   /**
-   * Waits for microtask queue to empty, then focuses
+   * Waits for the zone to stabilize, then focuses
    * the first tabbable element within the focus trap region.
    */
   focusFirstTabbableElementWhenReady() {
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => this.focusFirstTabbableElement());
+    this._executeOnStable(() => this.focusFirstTabbableElement());
   }
 
   /**
-   * Waits for microtask queue to empty, then focuses
+   * Waits for the zone to stabilize, then focuses
    * the last tabbable element within the focus trap region.
    */
   focusLastTabbableElementWhenReady() {
-    this._ngZone.onMicrotaskEmpty.first().subscribe(() => this.focusLastTabbableElement());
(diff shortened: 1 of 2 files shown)


src/lib/core/a11y/focus-trap.ts
```diff
@@ -107,13 +111,11 @@ export class FocusTrap {
    * @returns The boundary element.
    */
   private _getRegionBoundary(bound: 'start' | 'end'): HTMLElement | null {
-    let markers = [
-      ...Array.prototype.slice.call(this._element.querySelectorAll(`[cdk-focus-region-${bound}]`)),
-      // Deprecated version of selector, for temporary backwards comparability:
-      ...Array.prototype.slice.call(this._element.querySelectorAll(`[cdk-focus-${bound}]`)),
-    ];
+    // Contains the deprecated version of selector, for temporary backwards comparability.
+    let markers = this._element.querySelectorAll(`[cdk-focus-region-${bound}], ` +
```
mmalerba: I did this to make sure the new version takes precedence over the deprecated version, idk maybe it was overkill though

src/lib/core/a11y/focus-trap.ts
```diff
@@ -107,13 +111,11 @@ export class FocusTrap {
...
+                                                 `[cdk-focus-${bound}]`) as NodeListOf<HTMLElement>;
 
-    markers.forEach((el: HTMLElement) => {
+    Array.prototype.forEach.call(markers, (el: HTMLElement) => {
```
mmalerba: could we do a for of loop here to avoid the weird syntax? or does that have the same issue of not working with NodeList?

src/lib/core/a11y/focus-trap.ts
crisbeto: It's probably overkill since the deprecated and the supported directives do exactly the same.

src/lib/core/a11y/focus-trap.ts
crisbeto: Loops work fine. Changed.

mmalerba: I actually meant the `for (let val of iterable)` style, but this is fine too

mmalerba:
LGTM

PR #4868, opened 2017-05-29
fix(select): unable to programmatically select falsy values

### LINKED ISSUE #4854: Select not set to ngModel initial value
When setting the ngModel variable in code behind onInit  the  selected display value is not showing.
This works in beta 5 but not anymore in beta 6

The select should show the display string of the ngModel value

It does not show anything

```
<md-select [(ngModel)]="selectedSortBy " mdTooltip="Sort by">
              <md-option [value]="0">Latest</md-option>
              <md-option [value]="1">Highest</md-option>
              <md-option [value]="2">Lowest</md-option>
            </md-select>
```
and in component file onInit setting 
selectedSortBy= 0;
does not display Latest

Select should show the databound value

Material beta.6 latest angular

crisbeto:
Fixes not being able to set falsy values progammatically in `md-select`.

Fixes #4854.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -56,7 +56,8 @@ describe('MdSelect', () => {
         BasicSelectInitiallyHidden,
         BasicSelectNoPlaceholder,
         BasicSelectWithTheming,
-        ResetValuesSelect
+        ResetValuesSelect,
+        FalsyValueSelect
       ],
       providers: [
         {provide: OverlayContainer, useFactory: () => {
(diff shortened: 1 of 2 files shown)


src/lib/select/select.spec.ts
```diff
@@ -449,6 +450,22 @@ describe('MdSelect', () => {
       expect(fixture.componentInstance.select.selected).not.toBeDefined();
     });
 
+    it('should be able to programmatically select a falsy option', () => {
```
kara: Nit: Why is this in its own section? Seems like it would better fit under forms integration since it has to do with setting the form control value.

src/lib/select/select.spec.ts
```diff
@@ -449,6 +450,22 @@ describe('MdSelect', () => {
...
+      fixture.destroy();
+
+      let falsyFixture = TestBed.createComponent(FalsyValueSelect);
```
kara: Nit: const

src/lib/select/select.spec.ts
```diff
@@ -449,6 +450,22 @@ describe('MdSelect', () => {
...
+
+      falsyFixture.detectChanges();
+      falsyFixture.debugElement.query(By.css('.mat-select-trigger')).nativeElement.click();
```
kara: Clicking on the trigger doesn't seem necessary if you're trying to test setting the value programmatically. Remove?

crisbeto: It's there since we're also doing an extra check that the selected option has the proper class.

PR #4877, opened 2017-05-30
build: minify html before inlining resources

### LINKED ISSUE #1596: Remove trailing whitespace in inlined templates

DevVersion:
* Now minifies the HTML files before inlining the resources. This drops unnecessary whitespace due to previous line breaks and developers can easily test with Material components.

**Note**: The minifying could be done in another utility file as well but all our utility files should be done without gulp (because we want to have a standalone packaging tool). Only the `build-tasks-gulp` should contain gulp-related stuff. 

Fixes #1596

### DIFF
--- a/tools/gulp/packaging/build-tasks-gulp.ts
+++ b/tools/gulp/packaging/build-tasks-gulp.ts
@@ -26,8 +26,8 @@ export function createPackageBuildTasks(packageName: string, requiredPackages: s
   // Paths to the different output files and directories.
   const esmMainFile = join(packageOut, 'index.js');
 
-  // Glob that matches all assets that should be copied to the package.
-  const assetsGlob = join(packageRoot, '**/*.+(scss|css)');
+  // Glob that matches all style files that need to be copied to the package output.
+  const stylesGlob = join(packageRoot, '**/*.+(scss|css)');
 
   // Glob that matches every HTML file in the current package.
   const htmlGlob = join(packageRoot, '**/*.html');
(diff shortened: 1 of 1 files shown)


tools/gulp/packaging/build-tasks-gulp.ts
```diff
@@ -26,7 +27,10 @@ export function createPackageBuildTasks(packageName: string, requiredPackages: s
   const esmMainFile = join(packageOut, 'index.js');
 
   // Glob that matches all assets that should be copied to the package.
-  const assetsGlob = join(packageRoot, '**/*.+(html|scss|css)');
+  const assetsGlob = join(packageRoot, '**/*.+(scss|css)');
```
jelbourn: Since this doesn't include HTML anymore, it can just be `stylesGlob`

DevVersion: Yep good idea.

PR #4882, opened 2017-05-30
feat(menu): coercing overlapTrigger attribute

ocarreterom:
Uses coerceBooleanProperty for `overlapTrigger` attribute.
```
<md-menu overlapTrigger="false"></md-menu>
```

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -21,7 +21,7 @@ import {MdMenuPanel} from './menu-panel';
 import {Subscription} from 'rxjs/Subscription';
 import {transformMenu, fadeInItems} from './menu-animations';
 import {ESCAPE} from '../core/keyboard/keycodes';
-
+import {coerceBooleanProperty} from '../core';
 
 @Component({
   moduleId: module.id,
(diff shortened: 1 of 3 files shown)


ocarreterom:
I signed it!

src/lib/menu/menu-directive.ts
```diff
@@ -75,7 +75,10 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   @ContentChildren(MdMenuItem) items: QueryList<MdMenuItem>;
 
   /** Whether the menu should overlap its trigger. */
-  @Input() overlapTrigger = true;
+  @Input()
+  get overlapTrigger(): boolean { return this._overlapTrigger; }
+  set overlapTrigger(value: boolean) { this._overlapTrigger = coerceBooleanProperty(value); }
```
jelbourn: `coerceBooleanProperty` should only be used for things that start as `false` and are optionally turned to `true`. 

@kara was `overlapTrigger` meant to be set to `true` by default?

kara: Yes, it should overlap the trigger by default

jelbourn:
This will be captured in #7283

PR #4883, opened 2017-05-30
fix(input,select): set disabled underline color based on theme

crisbeto:
* Fixes the disabled underline blending in with the background in dark themes.
* Increases the opacity of the disabled underline from 26% to 38% to make a bit more visible against the background.


crisbeto:
I've rolled these changes into #5155. Closing.

PR #4887, opened 2017-05-30
build: add missing bits for material-examples package

jelbourn:
Adds:
* Missing `flatModuleId` from the ngc options
* Missing typing for `module` in order to set moduleId
* Makes templateUrl and styleUrls for example components match material
components.

### DIFF
--- a/src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
+++ b/src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
@@ -6,7 +6,7 @@ import 'rxjs/add/operator/map';
 
 @Component({
   selector: 'autocomplete-overview-example',
-  templateUrl: './autocomplete-overview-example.html',
+  templateUrl: 'autocomplete-overview-example.html',
 })
 export class AutocompleteOverviewExample {
   stateCtrl: FormControl;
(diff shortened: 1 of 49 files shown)


src/material-examples/typings.d.ts
```diff
@@ -0,0 +1 @@
+declare var module: {id: string};
```
tinayuangao: Is this module.id available in plunkers?

DevVersion: The moduleId should be removed by the inline resources script. Why do we need it at all?

jelbourn: It was part of my debugging process. It ended up not being the problem, but I left it because the material2 components themselves use it. However, I didn't think of the plunkers. It looks like I failed to update how those work at all. Fixing that now...

DevVersion: I see. Basically they shouldn't be necessary at all. It's fine to leave them for now though.

Just want to check. Does something need the moduleId in our development? (e.g Google internally)

jelbourn: Yes

jelbourn:
Updated to remove `moduleId` for examples

src/material-examples/typings.d.ts
DevVersion: Alright that makes sense then. For the examples you removed the `module.id`. You can remove the typings file therefore too

Issue #4601, opened 2017-05-16
add mdTooltipClass to allow CSS customization of tooltips


arlowhite:
#### Bug, feature request, or proposal:

The ability to specify a CSS class for a **MdTooltip**.

#### What is the expected behavior?

Developers should be able to customize some tooltips with CSS rules.

#### What is the current behavior?

Developers cannot apply CSS to a portion of tooltips because they are created in the `cdk-overlay-container`. Their only option is to apply CSS globally using `.mat-tooltip`.

#### What is the use-case or motivation for changing an existing behavior?

A number of feature requests could probably be resolved if developers had **mdTooltipClass**
#3389

DevVersion:
Developers would still need to disable the view encapsulation to reach to the tooltip overlay element.

Or is the main reason to be able to differentiate between different tooltips?

PR #4893, opened 2017-05-31
feat(tooltip): add mdTooltipClass

willshowell:
Fixes https://github.com/angular/material2/issues/4601

* I wasn't able to identify a reason for ViewEncapsulation to be on for the tooltip, so I set it to none. Will gladly revert if this is incorrect, but it will help users with their css customizations.
* I chose to support the ngClass syntax to match select's `panelClass`. But if the user uses the key/value format and adds or removes a class that affects tooltip size while it's showing, its position won't update.


andela-iikikin:
There's hardly any description on how to apply this

PR #4894, opened 2017-05-31
build: move inline resources script to tools

DevVersion:
* Moves the inline resources script to the `tools/gulp/packaging` folder. This is necessary when exposing the build tools (e.g for flex-layout)
* This also makes the packaging more clean because no external script is referenced (also switched to TypeScript and improved code)


tools/gulp/packaging/inline-resources.ts
```diff
@@ -0,0 +1,58 @@
+/* tslint:disable:no-eval */
+
+import {dirname, join} from 'path';
+import {readFileSync, writeFileSync} from 'fs';
+import {sync as glob} from 'glob';
+
+/** Finds all JavaScript files and inlines all external resources of Angular components. */
+export function inlineResourcesFolder(folderPath: string) {
```
jelbourn: how about just `inlineAllResources`?

tools/gulp/packaging/inline-resources.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  fileContent = inlineTemplate(fileContent, filePath);
+  fileContent = inlineStyles(fileContent, filePath);
+  fileContent = removeModuleId(fileContent);
+
+  writeFileSync(filePath, fileContent, 'utf-8');
+}
+
+/** Inlines the templates of Angular components for a specified source file. */
+function inlineTemplate(fileContent: string, filePath: string) {
```
jelbourn: Are the inlined resources here minified?

tools/gulp/packaging/inline-resources.ts
```diff
@@ -0,0 +1,58 @@
...
+    const templateContent = loadResourceFile(templateFile);
+    return `template: "${templateContent}"`;
+  });
+}
+
+
+/** Inlines the external styles of Angular components for a specified source file. */
+function inlineStyles(fileContent: string, filePath: string) {
+  return fileContent.replace(/styleUrls:\s*(\[[\s\S]*?])/gm, (match, styleUrls) => {
+    // The RegExp matches the array of external style files. This is a string right now and
```
jelbourn: Expand comment to include something like
```ts
// styleUrls will be a string that looks something like "['button.css', 'other.css']"
```

tools/gulp/packaging/inline-resources.ts
```diff
@@ -0,0 +1,58 @@
...
+    const templateFile = join(dirname(filePath), templateUrl);
+    const templateContent = loadResourceFile(templateFile);
+    return `template: "${templateContent}"`;
+  });
+}
+
+
+/** Inlines the external styles of Angular components for a specified source file. */
+function inlineStyles(fileContent: string, filePath: string) {
+  return fileContent.replace(/styleUrls:\s*(\[[\s\S]*?])/gm, (match, styleUrls) => {
```
jelbourn: Maybe rename `styleUrls` to `styleUrlsValue` to make it more clear that it's just yet the real array of styleUrl strings. Then `prsedUrls` below could just change to `styleUrls`.

tools/gulp/packaging/inline-resources.ts
```diff
@@ -0,0 +1,58 @@
...
+    // The RegExp matches the array of external style files. This is a string right now and
+    // can to be parsed using the `eval` method.
+    const parsedUrls = eval(styleUrls) as string[];
+
+    return 'styles: [' + parsedUrls.map(styleUrl => {
+      const stylePath = join(dirname(filePath), styleUrl);
+      const styleContent = loadResourceFile(stylePath);
+      return `"${styleContent}"`;
+    }).join(',\n') + ']';
```
jelbourn: Maybe refactor this to
```ts
const styles = parsedUrls
    .map(url => join(dirname(filePath), url))
    .map(path => loadResourceFile(path));

return `styles: [${styleContents.join(',')}]`;
```
(the newline shouldn't be necessary AFAICT)

tools/gulp/packaging/inline-resources.ts
```diff
@@ -0,0 +1,58 @@
...
+  fileContent = removeModuleId(fileContent);
+
+  writeFileSync(filePath, fileContent, 'utf-8');
+}
+
+/** Inlines the templates of Angular components for a specified source file. */
+function inlineTemplate(fileContent: string, filePath: string) {
+  return fileContent.replace(/templateUrl:\s*'([^']+?\.html)'/g, (match, templateUrl) => {
+    const templateFile = join(dirname(filePath), templateUrl);
+    const templateContent = loadResourceFile(templateFile);
```
jelbourn: Rename variables to `templatePath` and `template`?

tools/gulp/packaging/inline-resources.ts
DevVersion: No, the minifying happens before by Gulp.

tools/gulp/packaging/inline-resources.ts
DevVersion: I thought about that but `AllResources` sounds like it all resources a given file.

tools/gulp/packaging/inline-resources.ts
DevVersion: Yeah I like that.

tools/gulp/packaging/inline-resources.ts
DevVersion: I like that. Not sure about the line break. I didn't write it but I will test it.

tools/gulp/packaging/inline-resources.ts
DevVersion: `templatePaths` sounds good. For the `template` I'd like to keep the `content` to indicate it's the text.

tools/gulp/packaging/inline-resources.ts
jelbourn: Then maybe `inlineAllResourceForDirectory`?

tools/gulp/packaging/inline-resources.ts
jelbourn: That's fine too

tools/gulp/packaging/inline-resources.ts
DevVersion: Yeah `inlineAllResourceForDirectory` sounds good

DevVersion: Actually why not just `inlineResourcesForDirectory`?

DevVersion:
@jelbourn Addressed your comments. The `inlineStyles` task now looks way better. Thanks

tools/gulp/packaging/inline-resources.ts
```diff
@@ -0,0 +1,57 @@
...
+    const templateContent = loadResourceFile(templatePath);
+    return `template: "${templateContent}"`;
+  });
+}
+
+/** Inlines the external styles of Angular components for a specified source file. */
+function inlineStyles(fileContent: string, filePath: string) {
+  return fileContent.replace(/styleUrls:\s*(\[[\s\S]*?])/gm, (match, styleUrlsValue) => {
+    // The RegExp matches the array of external style files. This is a string right now and
+    // can to be parsed using the `eval` method. The value looks like "['AAA.css', 'BBB.css"]"
```
jelbourn: That second-to-last `"` should be a `'`

PR #4896, opened 2017-05-31
build: switch to npm package-lock file

DevVersion:
* No longer tracks the yarn.lock file and switches to NPM's `package-lock` file (for NPM v5)

@jelbourn The CI currently doesn't use NPM v5 yet, but there may happen some updates from Travis CI in the upcoming days. If necessary we can just install NPM v5 manually as well.


crisbeto:
FWIW, switching to Node 8 should include NPM 5 as well.

DevVersion:
@crisbeto Yeah considered that too but Node 8 is not a LTS version yet. 

And for now we should run our tooling with the LTS version (ensuring that most of the people can use our tooling)

Issue #4738, opened 2017-05-23
Navtabs dynamic width


mbeckenbach:
#### Bug, feature request, or proposal:
Feature request

#### What is the expected behavior?
When the innerHtml of a tab link changes, the underline should refresh its witdh.

#### What is the current behavior?

You have for example these nav tab links:
Tab 1 | Tab 2 | Tab 3

Now change the label of 'Tab 1' to 'Bananarama' which is much longer than 'Tab 1'.
The active underline will stay at the width of 'Tab 1'.

Edit: When removing the nav-tabs component from dom and re-add after a few milliseconds the underline will be correct. Looks like it's only calculated onInit

#### What is the use-case or motivation for changing an existing behavior?
Dynamic tab labels

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
ng 4.1.x, md2 beta 5

jotatoledo:
Would love to see this working

andrewseguin:
Perhaps we could add a mutation observer to the component to check if any mutations occur that were not additions/removals. 

@crisbeto What do you think about the best way is to use the mutation observer here?

crisbeto:
Since we allow HTML in the tab, we don't have a choice, apart from the `MutationObserver` (we could've used `ngOnChanged` if the label was coming from an `@Input`). We can use [the same approach](https://github.com/angular/material2/blob/e4789c7b88975b1811c97b9bed61b5278def0a7c/src/lib/tabs/tab-header.html#L11) that we have on the regular tabs.

andrewseguin:
Seems good to me, except we'll be applying extra ink bar alignments since the ripple causes mutations as well (same in tabs). Would be nice to have a good way to filter those out as mutations

crisbeto:
AFAIK the MutationObserver has some (not too granular) ways of filtering. We'd need to add those to the `cdkObserveContent` directive.

PR #4897, opened 2017-05-31
fix(navbar): update ink bar when links change

### LINKED ISSUE #4737: Navtabs active index wrong
Bug

Correct active highlighting

When removing or adding a tab via ngIf the active tab highlight from routerLinkActive is not beeing updated.

An example:

Tab 1 | Tab 2 | Tab 3 (Active) | Tab 4

When you remove Tab 2 from the DOM, the active highlight will move from Tab 3 to Tab 4 as it seams to be bound to the index and not to the router link.

Not sure how to create a plunkr with routing sorry

ng 4.1.x, md2 beta 5

andrewseguin:
Fixes #4737, #4738

When the list of tab links change, the ink bar should adjust to make sure it's positioned under the right element. E.g. tabs are added/removed/moved/content changed

### DIFF
--- a/src/lib/tabs/tab-nav-bar/tab-nav-bar.spec.ts
+++ b/src/lib/tabs/tab-nav-bar/tab-nav-bar.spec.ts
@@ -10,7 +10,7 @@ import {LayoutDirection, Dir} from '../../core/rtl/dir';
 import {Subject} from 'rxjs/Subject';
 
 
-fdescribe('MdTabNavBar', () => {
+describe('MdTabNavBar', () => {
   let dir: LayoutDirection = 'ltr';
   let dirChange = new Subject();
 
--- a/src/lib/tabs/tab-nav-bar/tab-nav-bar.spec.ts
+++ b/src/lib/tabs/tab-nav-bar/tab-nav-bar.spec.ts
@@ -10,7 +10,7 @@ import {LayoutDirection, Dir} from '../../core/rtl/dir';
 import {Subject} from 'rxjs/Subject';
 
 
-describe('MdTabNavBar', () => {
+fdescribe('MdTabNavBar', () => {
   let dir: LayoutDirection = 'ltr';
   let dirChange = new Subject();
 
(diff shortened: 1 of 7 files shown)


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -37,10 +41,14 @@ export class MdTabNavBar implements AfterContentInit, OnDestroy {
   /** Combines listeners that will re-align the ink bar whenever they're invoked. */
   private _realignInkBar: Subscription = null;
 
+  /** Subject that emits when the component has been destroyed. */
+  private _onDestroy = new Subject<void>();
```
crisbeto: Instead of having an internal `Subject`, why not save the subscription and unsubscribe on destroy instead? On a related note, I wonder whether the `_tabLinks` wouldn't get completed automatically on destroy.

andrewseguin: Both work, playing with the idea of using this subject for our components' subscriptions. It's an idea Rob is considering making "a thing" that seems to play nice with observables and lets you avoid storing the subscription in another variable.

What do you think about this @jelbourn

jelbourn: ¯\\_(ツ)_/¯

andrewseguin:
Changed the strategy to tackle another issue - align ink bar when the inner HTML changes. Using mutation observers to know when the content is changed. Can use this for when links are added/removed/moved as well as when their content has changed (and perhaps their width).

@jelbourn  can you re-review?

src/lib/tabs/tab-nav-bar/tab-nav-bar.spec.ts
```diff
@@ -64,6 +63,31 @@ describe('MdTabNavBar', () => {
...
+
+      spyOn(inkBar, 'alignToElement');
+
+      fixture.componentInstance.tabs = [1, 2, 3, 4];
+      fixture.detectChanges();
+
+      expect(inkBar.alignToElement).toHaveBeenCalled();
+    });
+
+    it('should re-align the ink bar when the tab labels change the width', done => {
```
jelbourn: Why does this test use the `done` callback instead of `fakeAsync`? I actually don't see why either is necessary

andrewseguin: Nothing seemed to be able to trigger the mutation observer (detect changes, flush microtasks, etc). Took a cue from observe-content.spec.ts which uses `done => ()` as well.

PR #4898, opened 2017-05-31
fix(button-toggle): dirty on load when used with ngModel

### LINKED ISSUE #4888: md-button-toggle-group starts off as dirty
Bug

An md-button-toggle-group whose value has not changed should not be dirty

dirty is true

http://plnkr.co/edit/H1q6YvW2qH4xsYeP9RX0?p=preview

I have a save button that should only be enabled if the user changed the toggle value.

I used the plunker template so it is the latest Angular / Material versions.  I'm testing on Windows / Chrome.

I can't come up with a good workaround.  I tried binding to `ngModelChange` and calling `markAsPristine` on the control but it is too early.

crisbeto:
Fixes the button toggle group being marked as dirty on load, if it has a ControlValueAccessor and an initial value.

Fixes #4888.

### DIFF
--- a/src/lib/button-toggle/button-toggle.spec.ts
+++ b/src/lib/button-toggle/button-toggle.spec.ts
@@ -5,8 +5,8 @@ import {
   ComponentFixture,
   TestBed,
 } from '@angular/core/testing';
-import {NgModel, FormsModule, ReactiveFormsModule, FormControl} from '@angular/forms';
-import {Component, DebugElement} from '@angular/core';
+import {NgModel, FormsModule, ReactiveFormsModule, FormControl, NgControl} from '@angular/forms';
+import {Component, DebugElement, ViewChild} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {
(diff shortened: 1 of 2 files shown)


kara:
@crisbeto It seems to me that the underlying problem is that value changes from *any* source lead to the forms API `onChange` function being called (which sets the control to dirty), rather than only user-generated value changes. Deferring the setting of the `isInitialized` property avoids the initial call, but any programmatic writes after that will still improperly mark the control as dirty. Can we instead pass a boolean into the `_emitChangeEvent` method for whether it was user-generated, then only call the forms API change function in that case?

jelbourn:
@crisbeto is this worth revisiting?

crisbeto:
Yeah it's been on my TODO list, but I was waiting for the refactoring to get in.

jelbourn:
Ack, we just have some failing tests in Firebase on the refactor I need to debug. Poke me on this when it's resolved

josephperrott:
@crisbeto Please rebase when you have a chance.

PR #4901, opened 2017-06-01
feat: add camelCase selectors where they were missing

jelbourn:
* Also get rid of all `true` class bindings once and for all. 
* Also (proposing) shortening `md-tab-nav-bar` to just `mdTabNav`
* Also (proposing) altering card image sizes (`md-card-sm-image` to `mdCardImageSmall`)

cc @andrewseguin @kara @mmalerba @tinayuangao for discussion before I do the tests/demos

After this, all of the dash-case selectors would be announced as deprecated in the changelog and removed once I can get them all out of Google apps


src/lib/list/list.ts
```diff
@@ -105,20 +94,19 @@ export class MdListIconCssMatStyler {}
...
-  }
+  host: {'class': 'mat-subheader'}
 })
 export class MdListSubheaderCssMatStyler {}
 
 @Component({
   moduleId: module.id,
-  selector: 'md-list-item, mat-list-item, a[md-list-item], a[mat-list-item]',
+  selector: 'md-list-item, mat-list-item, a[md-list-item], a[mat-list-item],' +
+            'a[mdListItem], a[matListItem]',
```
mmalerba: So it looks like we're just doing away with dash-case attrs entirely, even for components? (I'm happy with that, I've wanted it for a while) shouldn't we do button too?

jelbourn: This one was a mistake, reverted

jelbourn:
Officially for review now; I will update the tests and demos in follow-up PRs (trying to avoid merge conflicts)

src/lib/card/card.ts
```diff
@@ -71,58 +63,48 @@ export class MdCardFooter {}
...
 /**
  * Image used in a card, needed to add the mat- CSS styling.
  * @docs-private
  */
 @Directive({
-  selector: '[md-card-image], [mat-card-image]',
-  host: {
-    '[class.mat-card-image]': 'true'
-  }
+  selector: '[md-card-lg-image], [mat-card-lg-image], [mdCarDImageLarge], [matCardImageLarge]',
```
kara: Typo here? Should be `mdCardImageLarge`

src/lib/dialog/dialog-content-directives.ts
```diff
@@ -43,9 +42,7 @@ export class MdDialogTitle { }
  */
 @Directive({
   selector: '[md-dialog-content], md-dialog-content, [mat-dialog-content], mat-dialog-content',
```
kara: Should `[md-dialog-content]` be updated to camelcase too?

src/lib/card/card.ts
jelbourn: Done

src/lib/dialog/dialog-content-directives.ts
jelbourn: Done

src/lib/input/autosize.ts
```diff
@@ -19,33 +19,28 @@ export class MdTextareaAutosize implements AfterViewInit {
...
   set maxRows(value: number) {
     this._maxRows = value;
     this._setMaxHeight();
   }
 
-  /** Minimum number of rows for this textarea. */
-  @Input()
-  get mdAutosizeMinRows(): number { return this.minRows; }
-  set mdAutosizeMinRows(value: number) { this.minRows = value; }
+  @Input('matAutosizeMaxRows')
```
andrewseguin: This is now private but relied on in internal tests. Do you want to make it deprecated for use and then make private after that?

kara:
@jelbourn Rebase?

jelbourn:
Done

PR #4905, opened 2017-06-01
fix(autocomplete): don't scroll panel when option is visible

willshowell:
When scrolling down to an option well below the fold and then scrolling back up a couple of options, it feels jarring to have the selected option "stick" to the bottom of the panel. This fix keeps the panel from scrolling when the selected option is fully visible.

cc @kara 

-----

**Before**
autocomplete-before

**After**
autocomplete-after

### DIFF
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -534,6 +534,7 @@ describe('MdAutocomplete', () => {
     let fixture: ComponentFixture<SimpleAutocomplete>;
     let input: HTMLInputElement;
     let DOWN_ARROW_EVENT: KeyboardEvent;
+    let UP_ARROW_EVENT: KeyboardEvent;
     let ENTER_EVENT: KeyboardEvent;
 
     beforeEach(() => {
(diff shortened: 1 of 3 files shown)


willshowell:
@jelbourn could I get this assigned? 😄

willshowell:
@kara I would really appreciate if you could review this! I think it makes keyboard scrolling feel more natural.

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -763,6 +763,64 @@ describe('MdAutocomplete', () => {
...
+
+      // These up arrows will set the 2nd option active
+      [4, 3, 2, 1].forEach(() => {
+        fixture.componentInstance.trigger._handleKeydown(UP_ARROW_EVENT);
+        tick();
+      });
+
+      // Expect no scrolling to have occurred. Still showing bottom of 6th option.
+      expect(scrollContainer.scrollTop)
+        .toEqual(32, `Expected panel to not scroll back.`);
```
kara: Nit: The failure message is a bit vague. It seems like you should move the guidance in the comment to the failure message.

willshowell:
@kara changes made!

PR #4906, opened 2017-06-01
feat(input): autosize sets default amount of rows to one

### LINKED ISSUE #4852: mdTextareaAutosize doesn't resize textarea to a single row when mdAutosizeMinRows is set to 1
Bug

Textarea should be resized to a single line.

Resize doesn't shrink textarea to a single line, stops at minimum of two lines.

http://plnkr.co/edit/0HyRxBSQI7wksb7QFLDE?p=preview

Angular: 4.1.3
Material: v2.0.0-beta.6
OS: tested on Mac OS X 10.11.6
Browsers: tested on both Chrome 58 and Firefox 54

DevVersion:
* The autosize directive now sets the rows property of the textarea to one by default.
* This is necessary because browsers by default set the rows property to two and therefore setting the minRows and maxRows binding to something below two doesn't work.

**Note**: The test looks more complex than you'd expect. I wanted to have the test also confirm that the textarea can still grow (even if rows is set to `1` always)

Fixes #4852

### DIFF
--- a/src/lib/input/autosize.ts
+++ b/src/lib/input/autosize.ts
@@ -10,6 +10,9 @@ import {Directive, ElementRef, Input, AfterViewInit} from '@angular/core';
   exportAs: 'mdTextareaAutosize',
   host: {
     '(input)': 'resizeToFitContent()',
+    // Textarea elements that have the directive applied should have a single row by default.
+    // Browsers normally show two rows by default and therefore this limits the minRows binding.
+    'rows': '1',
   },
 })
 export class MdTextareaAutosize implements AfterViewInit {
@@ -116,7 +119,8 @@ export class MdTextareaAutosize implements AfterViewInit {
 
   /** Resize the textarea to fit its content. */
   resizeToFitContent() {
-    let textarea = this._elementRef.nativeElement as HTMLTextAreaElement;
+    const textarea = this._elementRef.nativeElement as HTMLTextAreaElement;
+
     // Reset the textarea height to auto in order to shrink back to its default size.
     textarea.style.height = 'auto';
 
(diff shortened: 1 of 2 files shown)


src/lib/input/autosize.ts
```diff
@@ -71,9 +71,23 @@ export class MdTextareaAutosize implements AfterViewInit {
...
+    this._setInitialRows();
     this.resizeToFitContent();
   }
 
+  /**
+   * Sets the rows property of the textarea to one by default.
+   * Browsers by default set the rows property to two. Therefore setting the minRows and maxRows
+   * binding to something below two doesn't work.
+   */
+  private _setInitialRows() {
```
jelbourn: Is there any reason this can't be a simple host binding of `{'rows': '1')`?

DevVersion: Not at all. I totally forgot about this 😄

DevVersion: Done. Not sure why the screenshot tests fail. The images don't seem to be different.

src/lib/input/autosize.ts
```diff
@@ -10,6 +10,9 @@ import {Directive, ElementRef, Input, AfterViewInit} from '@angular/core';
   exportAs: 'mdTextareaAutosize',
   host: {
     '(input)': 'resizeToFitContent()',
+    // Textarea elements that have the directive applied should have a single row by default.
+    // Browsers normally show two rows by default and this makes the minRows and maxRows unusable.
```
jelbourn: Maybe a bit overstated: without this it's just a problem for minRows=1

DevVersion: In theory it could be also for `maxRows` because if someone sets `maxRows=1` it would be still at two rows I think.

But yeah, who would set `maxRows=1` if he wants autosize.

DevVersion:
@jelbourn Updated the comment. Let me know if I should change anything else.

PR #4907, opened 2017-06-01
build: validate-release check for external resources

DevVersion:
* Adds a check to the `validate-release` task that checks for any references to external resources.


mackelito:
Grymt! 😁
Spark by Readdle

PR #4909, opened 2017-06-01
build: switch to NPM v5

DevVersion:
* Switches to NPM v5 that should use the packages from the `$HOME/.npm` cache.
* Now caches the dashboard functions `node_modules` by using the shared cache in `$HOME/.npm`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ env:
 
 before_install:
   - source ./scripts/ci/env.sh
-  - npm i -g npm@latest
+  - npm i -g npm@^5.0.1
 
 install:
   - npm install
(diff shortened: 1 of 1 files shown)


DevVersion:
NPM 5 works fine here. Just seeing some issues with the Universal task that can be reproduced locally as well. Including this in this PR as well (in progress)

.travis.yml
```diff
@@ -39,6 +39,7 @@ env:
 
 before_install:
   - source ./scripts/ci/env.sh
+  - npm i -g npm@latest
```
jelbourn: Not sure if _always_ being on latest is a good idea; maybe just do `^5.0.0`?

DevVersion: Not sure, I think it should be okay, but using `caret` and v5 should be better.

DevVersion: Updated it.

PR #4912, opened 2017-06-01
feat(chips): Add removal functionality/styling.

tinayuangao:
Continue work of #2476


src/lib/chips/chip-list.ts
```diff
@@ -221,4 +292,14 @@ export class MdChipList implements AfterContentInit, OnDestroy {
     return index >= 0 && index < this.chips.length;
   }
 
+  /** Utility to check if an input element has no value. */
+  private static _isInputEmpty(element: HTMLElement): boolean {
```
jelbourn: Rather than making this a private static method, just make it a top level function in the module (not exported)

src/lib/chips/chip-list.ts
```diff
@@ -221,4 +292,14 @@ export class MdChipList implements AfterContentInit, OnDestroy {
...
+    if (element && element.nodeName.toLowerCase() == 'input') {
+      let input = element as HTMLInputElement;
+
+      return input.value == '' || input.value == null;
```
jelbourn: Could this be `!input.value`?

src/lib/chips/chip-remove.ts
```diff
@@ -0,0 +1,66 @@
...
+  /** Subscription for our onRemoveChange Observable */
+  _onRemoveChangeSubscription: Subscription;
+
+  constructor(protected _renderer: Renderer2, protected _elementRef: ElementRef,
+              protected _parentChip: MdChip) {
+    if (this._parentChip) {
+      this._onRemoveChangeSubscription = this._parentChip.onRemovableChange
+        .subscribe((value) => {
+          this._updateParent(value);
+        });
```
jelbourn: Prefer to break at the higher syntactic level:
```ts
this._onRemoveChangeSubscription = 
    this._parentChip.onRemovableChange.subscribe((value) => this._updateParent(value));
```

src/lib/chips/chip.ts
```diff
@@ -22,30 +24,45 @@ export interface MdChipEvent {
...
              mat-basic-chip, [mat-basic-chip], mat-chip, [mat-chip]`,
-  template: `<ng-content></ng-content>`,
+  template: `<ng-content></ng-content><div class="md-chip-focus-border"></div>`,
   host: {
     '[class.mat-chip]': 'true',
     'tabindex': '-1',
     'role': 'option',
 
     '[class.mat-chip-selected]': 'selected',
+    '[class.mat-chip-has-remove-icon]': '_hasRemoveIcon',
```
jelbourn: Is there any way to avoid having this class on the chip? It seems unnecessarily complicated that the chip has to know whether it contains a remove button.

src/lib/chips/chip.ts
```diff
@@ -123,15 +164,76 @@ export class MdChip implements Focusable, OnInit, OnDestroy {
...
+
+  /**
+   * Sets whether or not this chip is displaying a remove icon. Adds/removes the
+   * `md-chip-has-remove-icon` class.
+   */
+  _setHasRemoveIcon(value: boolean) {
+    this._hasRemoveIcon = value;
+  }
+
+  protected _checkDisabled(event: Event): boolean {
```
jelbourn: Can we omit this method? It's name doesn't really tell you what it does and it's just a big side-effect

src/lib/chips/chips.scss
```diff
@@ -32,14 +75,43 @@ $mat-chips-chip-margin: $mat-chip-horizontal-padding / 4;
...
 
-    &:last-child, [dir='rtl'] &:last-child {
-      margin-bottom: 0;
-    }
+.mat-chip-remove.mat-chip-remove-hidden {
+  display: none;
+}
+
+.mat-chip-list-container .mat-input-placeholder-wrapper {
+  top: -4px;
```
jelbourn: Having super-specific negative absolute coordinates like this seems like a code smell to me. Is there a way to eliminate this, or to alternatively make it more explainable?

src/lib/chips/chips.scss
```diff
@@ -32,14 +75,43 @@ $mat-chips-chip-margin: $mat-chip-horizontal-padding / 4;
...
+
+  [dir='rtl'] & {
+    right: 8px;
```
jelbourn: Make this `8px` a variable?

src/lib/chips/chips.scss
```diff
@@ -1,27 +1,70 @@
...
+$mat-chip-remove-font-size: 18px;
 
 .mat-chip-list-wrapper {
+
   display: flex;
   flex-direction: row;
   flex-wrap: wrap;
   align-items: flex-start;
+
+  .mat-chip:not(.mat-basic-chip) {
```
jelbourn: Maybe for a follow-up PR (can add a TODO), but I'd like to get rid of all of the `:not(.mat-basic-chip)`. It raises the specificity such that you can't just target `.mat-chip`. Would be better to target a CSS class that is only applied when the basic chip directive isn't applied.

src/lib/chips/chip-list.ts
```diff
@@ -3,17 +3,31 @@ import {
...
 import {coerceBooleanProperty} from '../core/coercion/boolean-property';
-import {SPACE, LEFT_ARROW, RIGHT_ARROW, TAB} from '../core/keyboard/keycodes';
 import {Subscription} from 'rxjs/Subscription';
+import {
+  LEFT_ARROW, RIGHT_ARROW, SPACE, BACKSPACE, DELETE, UP_ARROW, DOWN_ARROW
+} from '../core/keyboard/keycodes';
+import {Dir} from '../core/rtl/dir';
+
+@Directive({
+  selector: '[mdChipListContainer], [matChipListContainer]',
```
tinayuangao: One way to avoid this directive is 
```
<md-input-container>
  <md-chip-list>
    <input mdInput mdChipList />
  </md-chip-list>
</md-input-container>
```

tinayuangao:

src/lib/chips/chip-list.ts
mmalerba: My ideal would be to first do that work we've talked about for making `md-input-container` work with other types of inputs besides just native inputs. We could then have `md-chip-list` implement the interface and act as an input (possibly a readonly one if the user doesn't actually plop a native input inside of it). The reason I like this is that the `md-chip-list ` is then responsible for things like deciding when its empty, it can take into account both committed chips and uncommitted chips in the form of input that is currently being entered. In the end it would wind up looking similar to what tina suggested:

```html
<md-input-container>
  <md-chip-list>
    <md-chip>Chip 1</md-chip>
    <input/>
  </md-chip-list>
</md-input-container>
```
The key difference here though is that the native `<input>` is fully managed by the `md-chips-list`, the `md-input-container` knows nothing about the native input and instead interacts with the `md-chip-list` through some `MdInput` interface that the chips list implements.

If we don't want to do all this work now, then maybe we can make it look like tina's suggestion so its less of a change when we can eventually do it this way?

jelbourn: One of the thing we've been deliberate about with chips is that the actual chip-list and the individual chips don't know that they're being driven by an `<input>`; that connection is managed by `mdChipInput`. I'd like to keep that separation of concerns. That's why I think it's a useful idea to _also_ have something that combines the behavior of both input-container and chip-list as a convenience to reduce boilerplate.

tinayuangao:
Fixed the lint error. 

Put `MdChipRemove` and `MdChip` in one file to avoid circular dependency.

jelbourn:
Remove "merge-ready" until beta.7 is released

src/lib/chips/chip-list.ts
tinayuangao: Now the chip-list work with input-container in 2 ways: 
```html
<md-input-container>
  <md-chip-list mdPrefix #chipList>
    <md-chip>SomeChip</md-chip>
  </md-chip-list>
  <input mdInput [mdChipInputFor]="chipList">
</md-input-container>
```

```html
<md-chip-list #chipList>
    <md-chip>SomeChip</md-chip>
</md-chip-list>
<md-input-container>
  <input mdInput [mdChipInputFor]="chipList">
</md-input-container>
```

Also updated `chips-demo` to display demos for these two cases.

edogr1:
Is this going to be released in beta 8? And if so, when?

axtho:
please let this not fall asleep again ... pretty please ...

jelbourn:
@tinayuangao can you rebase before I take another look?

tinayuangao:
@jelbourn Rebased and fixed test. PTAL. Thanks!

axtho:
Merge ready?

src/lib/chips/chip-list.ts
```diff
@@ -109,59 +147,68 @@ export class MdChipList implements AfterContentInit, OnDestroy {
...
+    let target = event.target as HTMLElement;
+    let isInputEmpty = _isInputEmpty(target);
+    let isRtl = this._dir && this._dir.value == 'rtl';
+
+    let isPrevKey = (code == (isRtl ? RIGHT_ARROW : LEFT_ARROW));
+    let isNextKey = (code == (isRtl ? LEFT_ARROW : RIGHT_ARROW));
+    let isBackKey = (code == BACKSPACE || code == DELETE || code == UP_ARROW || isPrevKey);
+    let isForwardKey = (code == DOWN_ARROW || isNextKey);
+    // If they are on an empty input and hit backspace/delete/left arrow, focus the last chip
+    if (isInputEmpty && isBackKey) {
```
jelbourn: Just curious: how did you decide to disable moving "left" into the chips when the input isn't empty? material1 and gmail do this same thing, but Drive lets you arrow back through the chips even when there is still text.

src/lib/chips/chip.ts
```diff
@@ -110,14 +142,110 @@ export class MdChip extends _MdChipMixinBase implements Focusable, OnDestroy, Ca
...
+          this.toggleSelected();
+        }
+
+        // Always prevent space from scrolling the page since the list has focus
+        event.preventDefault();
+        break;
+    }
+  }
+
+  protected _checkDisabled(event: Event): boolean {
```
jelbourn: This function appears to never be used.

src/lib/chips/chip.ts
```diff
@@ -110,14 +142,110 @@ export class MdChip extends _MdChipMixinBase implements Focusable, OnDestroy, Ca
     this.onFocus.emit({chip: this});
   }
 
+  /**
+   * Allows for programmatic removal of the chip. Called by the MdChipList when the DELETE or
+   * BACKSPACE keys are pressed.
+   *
+   * Note: This only informs any listeners of the removal request, it does **not** actually remove
```
jelbourn: Would slightly rephrase this part of the comment as 
"Informs any listeners of the removal request. Does not remove the chip from the DOM."

src/lib/chips/chip.ts
```diff
@@ -110,14 +142,110 @@ export class MdChip extends _MdChipMixinBase implements Focusable, OnDestroy, Ca
...
+   * the chip from the DOM.
+   */
+  remove(): void {
+    if (this.removable) {
+      this.onRemove.emit({chip: this});
+    }
+  }
+
+  /** The aria-disabled state for the chip */
+  _isAriaDisabled(): string {
```
jelbourn: This function appears to never be used.

src/lib/chips/chip-list.ts
```diff
@@ -15,13 +15,30 @@ import {
...
-import {SPACE, LEFT_ARROW, RIGHT_ARROW, TAB} from '../core/keyboard/keycodes';
 import {coerceBooleanProperty} from '@angular/cdk';
 import {Subscription} from 'rxjs/Subscription';
+import {
+  LEFT_ARROW, RIGHT_ARROW, BACKSPACE, DELETE, UP_ARROW, DOWN_ARROW
+} from '../core/keyboard/keycodes';
+import {Directionality} from '../core';
+
+/** Utility to check if an input element has no value. */
+function _isInputEmpty(element: HTMLElement): boolean {
```
jelbourn: Doesn't need to be prefixed with an underscore (since it's internal to the module regardless)

src/lib/chips/chip.ts
```diff
@@ -110,14 +142,110 @@ export class MdChip extends _MdChipMixinBase implements Focusable, OnDestroy, Ca
...
+  }
+})
+export class MdChipRemove {
+
+  /** Whether or not the remove icon is visible. */
+  _isVisible: boolean = true;
+
+  @Input('mdChipRemoveVisible')
+  get visible() { return this._isVisible; }
+  set visible(value) { this._isVisible = coerceBooleanProperty(value); }
```
jelbourn: Why does this have a `visible` property built-in? Wouldn't someone just hide it like anything else (css or `ngIf`)?

src/demo-app/chips/chips-demo.scss
```diff
@@ -20,4 +20,15 @@
...
+
+  md-chip-list input {
+    width: 150px;
+  }
+
+  .mat-chip-remove.mat-icon {
+    font-size: 18px;
+    width: 1em;
+    height: 1em;
+    vertical-align: middle;
```
willshowell: `cursor: pointer`?

src/lib/chips/chip.ts
```diff
@@ -110,14 +142,110 @@ export class MdChip extends _MdChipMixinBase implements Focusable, OnDestroy, Ca
...
+
+
+/**
+ * Applies proper (click) support and adds styling for use with the Material Design "cancel" icon
+ * available at https://material.io/icons/#ic_cancel.
+ *
+ * Example:
+ *
+ *     <md-chip>
+ *       <md-icon mdChipRemove>clear</md-icon>
```
willshowell: I think this should be `cancel` for consistency

src/lib/chips/chip-list.ts
tinayuangao: It's something Topher had.

tinayuangao:
Comments addressed. Added margin for input. Please take another look. Thanks!

src/lib/chips/chip-list.ts
```diff
@@ -15,13 +15,27 @@ import {
...
 import {MdChip} from './chip';
 import {FocusKeyManager} from '../core/a11y/focus-key-manager';
-import {SPACE, LEFT_ARROW, RIGHT_ARROW} from '../core/keyboard/keycodes';
+import {BACKSPACE, DELETE, SPACE, DOWN_ARROW, LEFT_ARROW, RIGHT_ARROW, UP_ARROW} from '../core/keyboard/keycodes';
+import {coerceBooleanProperty, Directionality} from '@angular/cdk';
 import {Subscription} from 'rxjs/Subscription';
-import {coerceBooleanProperty} from '@angular/cdk';
+
+/** Utility to check if an input element has no value. */
+function _isInputEmpty(element: HTMLElement): boolean {
```
jelbourn: I would move this to the bottom of the file (we typically put module-level functions at the bottom). Also can remove the leading `_` since it's scoped to this module.

tinayuangao: Moved `_isInputEmpty` to `MdChipList`.

tinayuangao:
Fixed border outline. Please take another look. Thanks!

willshowell:
@tinayuangao @jelbourn just noticed that the `.mat-chip-remove.mat-icon` styles are in the demo stylesheet rather than `chips.scss`. Should it be up to the consumer to set the font-size/height/width/alignment of the icon?

jelbourn:
Merging this PR now, can make other fixes in a follow-up

Issue #4440, opened 2017-05-09
Docs(autocomplete): regex filter search issue with character "?" and "$"


skyfremen:
#### Bug, feature request, or proposal:
Documentation Bug on custom filtering : https://material.angular.io/components/component/autocomplete

#### What is the expected behavior?
Able to filter with character starting with ? and $ 

#### What is the current behavior?
Application crashes when filtering with character starting with ?
Not able to filter with character starting with $

#### What are the steps to reproduce?
[Plunker Link](https://plnkr.co/edit/nmLXRZ1ZgyaMpOwnYLdS?p=preview)
**Search With > Result**
S > Shelley
? > Application Crash
$ > No Result

#### What is the use-case or motivation for changing an existing behavior?
NA

#### Which versions of Angular, Material, OS, browsers are affected?
Angular 4.1.0
Material 2.0.0-beta.3
Chrome OS

#### Is there anything else we should know?
Came up with a solution to not use Regex. [Plunker Link](https://plnkr.co/edit/WCu5yUpiX9CAq87wB1I5?p=preview)
```
filter(name: string): User[] {
     return this.options.filter((s) => s.name.toLowerCase().indexOf(name.toLowerCase()) !== -1);
}
```

willshowell:
I opted for something similar to this:

```ts
filter(val: string): string[] {
  const escapedVal = val.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1');
  return this.options.filter(option => new RegExp(`^${escapedVal}`, 'gi').test(option)); 
}
```

Perhaps @kara can weigh in on whether or not regex escaping is within the scope of the autocomplete docs.

kara:
The reason the autocomplete in our demo has this default filter is due to accessibility concerns. See https://github.com/angular/material2/issues/3373 for more info. Essentially, it is confusing for users with screenreaders to hear options matching any letter in the word (rather than options that match starting with the first letter). I think it's outside the scope of the docs to explain how to escape particular Regex characters. That's a pretty broad topic that we wouldn't be able to cover sufficiently.

Issue #3373, opened 2017-03-02
A11y keyboard control AutoComplete nonstandard drop-down list filter


DaveBest99:
#### Bug, feature request, or proposal:
Bug: AutoComplete Drop-Down list does not display States as expected.

#### What is the expected behavior?
With focus in an empty text box, type any letter key. If any of the available choices begin with the letter typed, those choices are displayed in a drop down. If the letter typed does not match any of the available choices the drop-down list is not displayed and the focus remains in the text box, and the user is allowed to press up/down arrow keys in the list box. Until the user presses the arrow keys to highlight a particular choice, only the typed letters are displayed in the text box.

#### What is the current behavior?
Currently the screen reader user cannot see what is in the Edit box, and filtered list items with the typed text that appears anywhere in the item is displayed.

#### What are the steps to reproduce?
Use the keyboard with mouse disconnected and screen reader running.

#### What is the use-case or motivation for changing an existing behavior?
WCAG standard keyboard controls provide a consistent behaviour, and a consistent filter listing.

#### Which versions of Angular, Material, OS, browsers are affected?
Windows10 browsers and screen readers.

#### Is there anything else we should know?
Typing characters moves focus to any list item that has that letter in the State name. When focus is on "Alabama", pressing "n" moves focus to "Arizona" instead of "Nebraska". Pressing up/down arrow keys only shows those States that have that letter in the name, instead of those that start with that letter. Focus does not stay in the Edit box until an arrow key is pressed. Aria-alert could be used to announce the list item without moving the focus from the Edit box.

kara:
The filter is controlled by the user, not the autocomplete, so this is an issue with our demo. We should be using a demo filter that is more a11y-friendly. Will fix to filter by beginning of the string.

DaveBest99:
When you say “the filter is controlled by the user”, do you mean the end-user or the web developer Autocomplete user?

 

If the Autocomplete implementor wishes to use a nonstandard filter, then the component guidelines should instruct the implementor to include text below the Edit Combo box explaining the filter criteria. The AutoComplete component could be used for many different list selections that require specific filter criteria.

David

kara:
Apologies, I mean the web developer (not the end user). We can also add some guidance in the docs that recommends adds instructional info below.

PR #4921, opened 2017-06-01
fix(autocomplete example): regex metacharacter escape

girtsernstsons:
autocomplete material example fails if regex metacharacters (ie *) are provided. Fixed by adding escapeSpecChars function

### DIFF
--- a/src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
+++ b/src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
@@ -72,9 +72,16 @@ export class AutocompleteOverviewExample {
         .map(name => this.filterStates(name));
   }
 
+  escapeSpecChars(val: string) {
+    return val ? val.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")
+               : val;
+  }
+
   filterStates(val: string) {
-    return val ? this.states.filter(s => new RegExp(`^${val}`, 'gi').test(s))
+    let escVal: string = this.escapeSpecChars(val);
+    return val ? this.states.filter(s => new RegExp(`^${escVal}`, 'gi').test(s))
                : this.states;
   }
 
 }
+
--- a/src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
+++ b/src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
@@ -73,7 +73,7 @@ export class AutocompleteOverviewExample {
   }
 
   escapeSpecChars(val: string) {
-    return val ? val.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")
+    return val ? val.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
                : val;
   }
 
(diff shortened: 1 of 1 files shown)


girtsernstsons:
[https://plnkr.co/edit/TUInZlEv7gN9p9rjZXGu](url)

src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
```diff
@@ -72,9 +72,16 @@ export class AutocompleteOverviewExample {
...
 
+  escapeSpecChars(val: string) {
+    return val ? val.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
+               : val;
+  }
+
   filterStates(val: string) {
-    return val ? this.states.filter(s => new RegExp(`^${val}`, 'gi').test(s))
+    let escVal: string = this.escapeSpecChars(val);
+    return val ? this.states.filter(s => new RegExp(`^${escVal}`, 'gi').test(s))
```
jelbourn: I think a cleaner solution would be to just not use a regex at all and instead just use `indexOf`

girtsernstsons: Thanks for the hint. Decided to use `includes` instead of `indexOf`

willshowell: Related https://github.com/angular/material2/issues/4440#issuecomment-300336137

src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
```diff
@@ -73,7 +73,7 @@ export class AutocompleteOverviewExample {
   }
 
   filterStates(val: string) {
-    return val ? this.states.filter(s => new RegExp(`^${val}`, 'gi').test(s))
+    return val ? this.states.filter(s => s.toLowerCase().includes(val.toLowerCase()))
```
jelbourn: `includes` is not supported in IE11 (which we still support). We try to avoid requiring polyfills where possible

willshowell:
@jelbourn Per your suggestion to switch to `indexOf`, I believe the regex filter was put in place for a11y reasons

https://github.com/angular/material2/issues/3373, https://github.com/angular/material2/issues/4440

jelbourn:
@willshowell good point; it should be specifically `indexOf(...) === 0`

girtsernstsons:
@jelbourn @willshowell how about adding a toggle "Search from beginning" in the example which would switch between `indexOf===0` and `indexOf>-1`? IMHO it would provide a broader use case to developers.

jelbourn:
It's just an example- I think the `indexOf === 0` is sufficient

girtsernstsons:
@jelbourn ok, fixed with `indexOf === 0`

rafaelss95:
I know it's just an example, but with this change it behaves in a different way:

**Before:**

When "a" is typed it shows "Alabama", "Alaska", "Arizona", "Arkansas"

**Now:**

When "a" is typed it shows simply nothing, because `indexOf` is _case sensitive_.

Also, I think it should matches the whole string... example:

If I type "k" it should match `Arkansas`, for example.

So, to make it work in the cases described above, I'd suggest to change it to:

`return val ? this.states.filter(s => s.toLowerCase().indexOf(val.toLowerCase()) !== -1)`

@jelbourn

girtsernstsons:
@rafaelss95 
Fix already handles case sensitivity in the same way you've proposed - [aafa6b0](https://github.com/angular/material2/commit/aafa6b0d0472ed5294053f59266056e636527157).
The reason, why `indexOf(...) === 0` is more preferable over `indexOf(...) !== -1`, can be found here #3373

PR #4939, opened 2017-06-02
build: dynamic build config for packaging

DevVersion:
This should work already but I want to see if I can make any improvements in regards to the config approach again.

@jelbourn I'm hesitant because it's weird that it will throw an error if you import a file from the packaging before calling the `configureBuild` method (which is kinda common)

```ts
// NOTE: This already loads the config but there is none -> Error
import {createPackageBuildTasks} from './packaging/build-tasks-gulp';

// Configure the packaging tool.
import './configure-build';

// Create gulp tasks to build the different packages in the project.
createPackageBuildTasks('cdk');
createPackageBuildTasks('material', ['cdk']);
createPackageBuildTasks('material-examples', ['material']);

```

But it should be fine for now to start with.


tools/gulp/packaging/build-config.ts
```diff
@@ -0,0 +1,17 @@
...
+  packagesDir: string;
+  outputDir: string;
+  licenseBanner: string;
+}
+
+/** Build config being used to build package output. */
+export let buildConfig: BuildConfig;
+
+/** Configures the build by setting the build config. */
+export function configureBuild(config: BuildConfig) {
```
jelbourn: There are a couple other ways we could approach this. 
1. Each method takes in `buildConfig` as an argument
2. We create a class like `LibraryBuilder` like `new LibraryBuilder(config)` and then do, e.g., `libraryBuilder.composeRelease(config)`. This could delegate to the existing functions, passing along the necessary settings.

DevVersion: The reason I wanted to keep the config exposed is that other custom tasks (that can be specific to material or flex-layout) may need access to the build config for creating their paths.

jelbourn: In those cases it makes the most sense to me to pass the data around. Ultimately this is an acknowledgement that the build config _is_ data, and not a set of code constants.

DevVersion: Made the discussed changes.

PR #4946, opened 2017-06-03
fix: make material work with noUnusedParameters

### LINKED ISSUE #4443: ng cli build --prod fails when activating noUnusedParameters in tsconfig.json
Bug/Improvement

Should be able to build with `ng build --prod` when having `compilerOptions.noUnusedParameters: true` in `tsconfig.json`

Build fails with 
```
ERROR in my-project/src/$$_gendir/node_modules/@angular/material/typings/index.ngfactory.ts (2252,41): 'l' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdButton.html (8,1): 'v' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdAnchor.html (8,1): 'v' is declared but never used.

ERROR in my-project/src/$$_gendir/node_modules/@angular/material/typings/index.ngfactory.ts (3345,31): 'l' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdInputContainer_Host.html (2,1): '$event' is declared but never used.

ERROR in my-project/src/$$_gendir/node_modules/@angular/material/typings/index.ngfactory.ts (4539,31): 'l' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdListItem_Host.html (2,1): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdSelect.html (23,5): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdSelect.html (2,1): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdSelect.html (15,7): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdSidenavContainer.html (2,1): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdTabGroup.html (5,3): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdTabHeader.html (2,1): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdTabHeader.html (12,3): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdTabHeader.html (20,1): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdButtonToggle.html (17,1): 'v' is declared but never used.

ERROR in my-project/src/$$_gendir/node_modules/@angular/material/typings/index.ngfactory.ts (7426,31): 'l' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.MdSlideToggle_Host.html (2,1): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.SimpleSnackBar.html (3,1): '$event' is declared but never used.

ERROR in my-project/node_modules/@angular/material/typings/index.d.ts.TooltipComponent_Host.html (2,1): '$event' is declared but never used.

```

* create fresh project `ng new my-project`
* install and include @angular/material as per the docs
* set `"noUnusedParameters": true` in the projects `tsconfig.json`
* run `ng build --prod`

Apps using material should be able to compile with `noUnusedParameters: true` in `tsconfig.json`.

"@angular/animations": "^4.1.0",
"@angular/common": "^4.0.0",
"@angular/compiler": "^4.0.0",
"@angular/core": "^4.0.0",
"@angular/material": "^2.0.0-beta.3",
"@angular/cli": "1.0.1",

Looks like this is some generated code from html templates. I don't know if it's even the fault of angular material. 

The fix would be to define parameters with a leading `_`. See https://github.com/Microsoft/TypeScript/issues/9860.

DevVersion:
* Recently Angular core fixed most of their `noUnusedParameters` warnings in https://github.com/angular/angular/issues/15532 (this one is still left - https://github.com/angular/angular/issues/17131)
* When using Angular with Material and the users has `noUnusedParameters` enabled the library will be checked and will throw right now.

**Note**: I don't plan to add the `noUnusedParameters` to any build tsconfig files yet, because it would make the watching & live-reloading extremely bad. The `noUnusedParameters` should / can be tested on the CI in a follow-up.

Fixes #4443

### DIFF
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -72,7 +72,7 @@ describe('MdAutocomplete', () => {
         }},
         {provide: Dir, useFactory: () => ({value: dir})},
         {provide: ScrollDispatcher, useFactory: () => {
-          return {scrolled: (delay: number, callback: () => any) => {
+          return {scrolled: (_delay: number, callback: () => any) => {
             return scrolledSubject.asObservable().subscribe(callback);
           }};
         }}
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -814,8 +814,8 @@ class SingleCheckbox {
   checkboxColor: string = 'primary';
   checkboxValue: string = 'single_checkbox';
 
-  onCheckboxClick(event: Event) {}
-  onCheckboxChange(event: MdCheckboxChange) {}
+  onCheckboxClick: (event?: Event) => void = () => {};
+  onCheckboxChange: (event?: MdCheckboxChange) => void = () => {};
 }
 
 /** Simple component for testing an MdCheckbox with ngModel. */
--- a/src/lib/core/datetime/native-date-adapter.ts
+++ b/src/lib/core/datetime/native-date-adapter.ts
@@ -31,7 +31,11 @@ const DEFAULT_DAY_OF_WEEK_NAMES = {
 
 /** Creates an array and fills it with values. */
 function range<T>(length: number, valueFunction: (index: number) => T): T[] {
-  return Array.apply(null, Array(length)).map((v: undefined, i: number) => valueFunction(i));
+  const valuesArray = Array(length);
+  for (let i = 0; i < length; i++) {
+    valuesArray[i] = valueFunction(i);
+  }
+  return valuesArray;
 }
 
 
@@ -123,7 +127,7 @@ export class NativeDateAdapter extends DateAdapter<Date> {
     return new Date();
   }
 
-  parse(value: any, parseFormat: Object): Date | null {
+  parse(value: any): Date | null {
     // We have no way using the native JS Date to set the parse format or locale, so we ignore these
     // parameters.
     let timestamp = typeof value == 'number' ? value : Date.parse(value);
--- a/src/lib/core/overlay/scroll/scroll-dispatcher.ts
+++ b/src/lib/core/overlay/scroll/scroll-dispatcher.ts
@@ -105,7 +105,7 @@ export class ScrollDispatcher {
   getScrollContainers(elementRef: ElementRef): Scrollable[] {
     const scrollingContainers: Scrollable[] = [];
 
-    this.scrollableReferences.forEach((subscription: Subscription, scrollable: Scrollable) => {
+    this.scrollableReferences.forEach((_subscription: Subscription, scrollable: Scrollable) => {
       if (this.scrollableContainsElement(scrollable, elementRef)) {
         scrollingContainers.push(scrollable);
       }
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -143,7 +143,7 @@ describe('MdRadio', () => {
       expect(radioInstances[0].checked).toBe(false);
 
       let spies = radioInstances
-        .map((value, index) => jasmine.createSpy(`onChangeSpy ${index}`));
+        .map((radio, index) => jasmine.createSpy(`onChangeSpy ${index} for ${radio.name}`));
 
       spies.forEach((spy, index) => radioInstances[index].change.subscribe(spy));
 
--- a/src/demo-app/dialog/dialog-demo.ts
+++ b/src/demo-app/dialog/dialog-demo.ts
@@ -38,7 +38,7 @@ export class DialogDemo {
     // Possible useful example for the open and closeAll events.
     // Adding a class to the body if a dialog opens and
     // removing it after all open dialogs are closed
-    dialog.afterOpen.subscribe((ref: MdDialogRef<any>) => {
+    dialog.afterOpen.subscribe(() => {
       if (!doc.body.classList.contains('no-scroll')) {
         doc.body.classList.add('no-scroll');
       }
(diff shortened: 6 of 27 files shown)


src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('MdAutocomplete', () => {
         }},
         {provide: Dir, useFactory: () => ({value: dir})},
         {provide: ScrollDispatcher, useFactory: () => {
-          return {scrolled: (delay: number, callback: () => any) => {
+          return {scrolled: (_delay: number, callback: () => any) => {
```
jelbourn: Why does this need an underscore? I'd prefer not to use `_` for function arguments; is TS not smart enough to know that you've included an argument so that you can have access to a subsequent one?

DevVersion: Yeah unfortunately it's smart enough do detect this

src/lib/core/overlay/scroll/scroll-dispatcher.ts
```diff
@@ -105,7 +105,7 @@ export class ScrollDispatcher {
   getScrollContainers(elementRef: ElementRef): Scrollable[] {
     const scrollingContainers: Scrollable[] = [];
 
-    this.scrollableReferences.forEach((subscription: Subscription, scrollable: Scrollable) => {
+    this.scrollableReferences.forEach((_subscription: Subscription, scrollable: Scrollable) => {
```
jelbourn: Change this to `this.scrollableReferences.keys().forEach(...)` ?

src/lib/checkbox/checkbox.spec.ts
```diff
@@ -814,8 +814,8 @@ class SingleCheckbox {
   checkboxColor: string = 'primary';
   checkboxValue: string = 'single_checkbox';
 
-  onCheckboxClick(event: Event) {}
-  onCheckboxChange(event: MdCheckboxChange) {}
+  onCheckboxClick(_event: Event) {}
+  onCheckboxChange(_event: MdCheckboxChange) {}
```
jelbourn: Should change these to be properties instead of methods since they're reassigned anyway, e.g.
```ts
onCheckboxClick: (Event?) => void = () => {};
```
(here and elsewhere that follows this pattern)

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -31,7 +31,7 @@ const DEFAULT_DAY_OF_WEEK_NAMES = {
 
 /** Creates an array and fills it with values. */
 function range<T>(length: number, valueFunction: (index: number) => T): T[] {
-  return Array.apply(null, Array(length)).map((v: undefined, i: number) => valueFunction(i));
+  return Array.apply(null, Array(length)).map((_v: undefined, i: number) => valueFunction(i));
```
jelbourn: We could eliminate this case by switching to a `for` loop... (which is faster anyway)

src/lib/core/style/focus-origin-monitor.ts
```diff
@@ -129,7 +129,7 @@ export class FocusOriginMonitor {
    * @param renderer The renderer to use to invoke the focus method on the element.
    * @param origin The focus origin.
    */
-  focusVia(element: HTMLElement, renderer: Renderer2, origin: FocusOrigin): void {
+  focusVia(element: HTMLElement, origin: FocusOrigin): void {
```
jelbourn: This one is a breaking change that will need Google to be updated, so we should do it in a separate PR

(I checked, there are people using it)

src/lib/radio/radio.spec.ts
```diff
@@ -143,7 +143,7 @@ describe('MdRadio', () => {
       expect(radioInstances[0].checked).toBe(false);
 
       let spies = radioInstances
-        .map((value, index) => jasmine.createSpy(`onChangeSpy ${index}`));
+        .map((_value, index) => jasmine.createSpy(`onChangeSpy ${index}`));
```
jelbourn: Could add the `value` into the spy name so that it's used.

src/lib/core/style/focus-origin-monitor.ts
DevVersion: Agreed, prefixing with an underscore should work for now.

src/lib/core/overlay/scroll/scroll-dispatcher.ts
DevVersion: Unfortunately there is no `forEach()` method for iterators. And if we want to use iterators here we need to enable down-leveling of those ([read here](https://www.typescriptlang.org/docs/handbook/iterators-and-generators.html) and this requires a runtime shim for non-arrays.

DevVersion:
@jelbourn Addressed your feedback. Expect the one with the Iterator.

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -31,7 +31,11 @@ const DEFAULT_DAY_OF_WEEK_NAMES = {
 
 /** Creates an array and fills it with values. */
 function range<T>(length: number, valueFunction: (index: number) => T): T[] {
-  return Array.apply(null, Array(length)).map((v: undefined, i: number) => valueFunction(i));
+  const valuesArray = [];
```
jelbourn: `const valuesArray = Array(length);`
Better to initialize it to the right length

PR #4952, opened 2017-06-03
fix(select): panel in multiple mode not overlapping trigger

### LINKED ISSUE #4943: Select overlay in multiple mode needs to extend further
This is based on the discussion from https://github.com/angular/material2/issues/4171: because the left offset of a select in multiple mode is larger, we need to increase the overlay width to allow it to overlap the entire trigger.

crisbeto:
* Fixes the select panel in multiple mode not overlapping the trigger completely.
* Reformats the select.html to make it a little more manageable.

Fixes #4943.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -838,16 +838,18 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
     const overlayRect = this.overlayDir.overlayRef.overlayElement.getBoundingClientRect();
     const viewportRect = this._viewportRuler.getViewportRect();
     const isRtl = this._isRtl();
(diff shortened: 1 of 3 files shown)


src/lib/select/select.ts
```diff
@@ -838,16 +838,18 @@ export class MdSelect implements AfterContentInit, OnDestroy, OnInit, ControlVal
     const overlayRect = this.overlayDir.overlayRef.overlayElement.getBoundingClientRect();
     const viewportRect = this._viewportRuler.getViewportRect();
     const isRtl = this._isRtl();
+    const extraWidth = this.multiple ? SELECT_MULTIPLE_PANEL_PADDING_X + SELECT_PANEL_PADDING_X :
```
kara: Nit: `extraWidth` is a bit vague. How about `paddingWidth` ?

PR #4959, opened 2017-06-04
refactor(sidenav): switch to the angular animations API

crisbeto:
Switches the sidenav to use the `@angular/animations` API. This saves us a lot of code that was involved in orchestrating the various animations, in addition to making it easier to follow. It also makes it possible for the consumer to disable the animations by using the `NoopAnimationsModule`.

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -62,19 +62,20 @@ export class MdSidenavToggleResult {
   encapsulation: ViewEncapsulation.None,
   animations: [
     trigger('transform', [
-      state('open', style({
+      state('open, open-instant', style({
         transform: 'translate3d(0, 0, 0)',
         visibility: 'visible',
       })),
       state('void', style({
         visibility: 'hidden',
       })),
-      transition('* => *', animate('400ms cubic-bezier(0.25, 0.8, 0.25, 1)'))
+      transition('void => open-instant', animate('0ms')),
+      transition('void <=> open', animate('400ms cubic-bezier(0.25, 0.8, 0.25, 1)'))
     ])
   ],
   host: {
     'class': 'mat-sidenav',
-    '[@transform]': 'opened ? "open" : "void"',
+    '[@transform]': '_getAnimationState()',
     '(@transform.start)': '_isAnimating = true',
     '(@transform.done)': '_onAnimationEnd($event)',
     '(keydown)': 'handleKeydown($event)',
@@ -90,6 +91,9 @@ export class MdSidenavToggleResult {
 export class MdSidenav implements AfterContentInit, OnDestroy {
   private _focusTrap: FocusTrap;
   private _elementFocusedBeforeSidenavWasOpened: HTMLElement | null = null;
+
+  /** Whether the sidenav is initialized. Used for disabling the initial animation. */
+  private _enableAnimations = false;
 
   /** Alignment of the sidenav (direction neutral); whether 'start' or 'end'. */
   private _align: 'start' | 'end' = 'start';
@@ -178,6 +182,7 @@ export class MdSidenav implements AfterContentInit, OnDestroy {
   ngAfterContentInit() {
     this._focusTrap = this._focusTrapFactory.create(this._elementRef.nativeElement);
     this._focusTrap.enabled = this.isFocusTrapEnabled;
+    Promise.resolve().then(() => this._enableAnimations = true);
   }
 
   ngOnDestroy() {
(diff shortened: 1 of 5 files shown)


src/lib/sidenav/sidenav.ts
```diff
@@ -175,53 +154,28 @@ export class MdSidenav implements AfterContentInit, OnDestroy {
   }
 
 
-  /** Open this sidenav, and return a Promise that will resolve when it's fully opened (or get
-   * rejected if it didn't). */
-  open(): Promise<MdSidenavToggleResult> {
+  /**  Open the sidenav. */
+  open(): void {
     return this.toggle(true);
```
mmalerba: since these are all `void` now the `return` isn't necessary

src/lib/sidenav/sidenav.ts
```diff
@@ -175,53 +154,28 @@ export class MdSidenav implements AfterContentInit, OnDestroy {
...
   }
 
-  /**
-   * Close this sidenav, and return a Promise that will resolve when it's fully closed (or get
-   * rejected if it didn't).
-   */
-  close(): Promise<MdSidenavToggleResult> {
+  /** Close the sidenav. */
+  close(): void {
     return this.toggle(false);
```
mmalerba: `return` not needed

src/lib/sidenav/sidenav.ts
```diff
@@ -175,53 +154,28 @@ export class MdSidenav implements AfterContentInit, OnDestroy {
...
    * @param isOpen Whether the sidenav should be open.
-   * @returns Resolves with the result of whether the sidenav was opened or closed.
    */
-  toggle(isOpen: boolean = !this.opened): Promise<MdSidenavToggleResult> {
-    // Shortcut it if we're already opened.
-    if (isOpen === this.opened) {
-      return this._toggleAnimationPromise ||
-          Promise.resolve(new MdSidenavToggleResult(isOpen ? 'open' : 'close', true));
-    }
+  toggle(isOpen: boolean = !this.opened): void {
```
mmalerba: I assume the animation API provides some way for people to still hook into these events without returning the promise?

crisbeto: We can have an observable that emits in the `done` callback, but I'm not sure that it's worth the trouble.

crisbeto:

src/lib/sidenav/sidenav.ts
mmalerba: I'm just asking since I don't know much about the animations API, how do I as a user who is currently doing this:

```ts
snav.toggle().then(() => doStuff())
```
change it to work after this PR? I would guess something like this?

```html
<md-sidenav (@transform.start)="doStuff()"></md-sidenav>
```

crisbeto: That won't work, because the Angular animation events don't propagate IIRC. The way to do it would be:

```
snav.onOpen.subscribe(() => doStuff());
snav.open();
```

I guess we could return the `onOpen` and `onClose` observables from the open/close methods, but I'm not sure whether it's necessary. The promises that were a part of the old API looked like they were there only there for synchronizing the animations internally.

mmalerba: Ok sg, just wanted to make sure we weren't taking away functionality

kara:
Note: need to update some tests that rely on `mat-sidenav-opening` / `mat-sidenav-closing` classes before merging

mmalerba:
Please rebase

crisbeto:
Seems like the tests are breaking after a rebase. Will look into it tomorrow.

mmalerba:
sorry, needs more rebasing

kara:
@crisbeto Lint is failing right now; can you update?

src/lib/sidenav/sidenav.ts
mmalerba: @crisbeto Let's add back these promises because they are used pretty extensively throughout Google's code. We can leave a comment to remove them in the future. (I can do so as part of my other upcoming breaking sidenav changes).

crisbeto: Done.

mmalerba:
@andrewseguin Return values have been added back to toggle & friends, want to try another presubmit

andrewseguin:
Not sure what I'm seeing here, but it looks like you need to rebase. I see this in your `sidenav.ts` 

```
private _watchSidenavToggle(sidenav: MdSidenav): void {
    if (!sidenav || sidenav.mode === 'side') { return; }
    sidenav.onOpen.subscribe(() => this._setContainerClass(true));
    sidenav.onClose.subscribe(() => this._setContainerClass(false));
  }
```

which differs from master:

```
private _watchSidenavToggle(sidenav: MdSidenav): void {
    merge(sidenav.onOpenStart, sidenav.onCloseStart).subscribe(() => {
    });

    if (sidenav.mode !== 'side') {
      sidenav.onOpen.subscribe(() => this._setContainerClass(true));
      sidenav.onClose.subscribe(() => this._setContainerClass(false));
    }
  }
```

crisbeto:
Weird that GitHub didn't pick that one up as a conflict @andrewseguin. I've rebased it and fixed the issue.

andrewseguin:
Sorry one more thing, the e2e is getting timeouts on the CI. I tried re-running and got the same error. Can you check it out?

crisbeto:
Sorted out all the failures @andrewseguin.

PR #4964, opened 2017-06-04
feat(dialog): allow disableClose option to be updated

### LINKED ISSUE #3938: Allow changing of disableClose property for dialog while dialog is open
Feature request

If a dialog is opened with disableClose as false, being able to change disableClose to true from within the dialog's component making it so dialog cannot be closed by clicking outside of the dialog area.

Updating disableClose does not do anything. If the dialog is opened with disableClose as false and then disableClose is set to true, you can click outside the dialog and the dialog will close.

Change this.dialogRef.config.disableClose = true; from within a dialog component
[Example](https://plnkr.co/edit/lURc8hWdUpBW8C4xt0MX?p=info)

Using dialog for registration form. After registration submit the form changes to a confirmation code input which the user should not be able to navigate away from without registering again.

    "@angular/common": "^4.0.0",
    "@angular/cli": "1.0.0",
    "@angular/material": "^2.0.0-beta.2"
All OS & only tested Chrome but would assume all browsers

No :)

crisbeto:
* Exposes the disableClose option via the `MdDialogRef` and allows for it to be updated.
* Makes the `containerInstance` private to `MdDialogRef` since it doesn't make sense for it to be public anymore.
* Completes the `backdropClick` observable once the associated `overlayRef` is destroyed. This avoids having to unsubscribe manually or having to use `Observable.first`.

Fixes #3938.

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -100,6 +100,7 @@ export class OverlayRef implements PortalHost {
     this.detachBackdrop();
     this._portalHost.dispose();
     this._attachments.complete();
+    this._backdropClick.complete();
     this._detachments.next();
     this._detachments.complete();
   }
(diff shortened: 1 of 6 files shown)


crisbeto:

andrewseguin:
Needs rebase =)

PR #4970, opened 2017-06-05
test(slide-toggle): verify that NgControl stays pristine if initial value is set

### LINKED ISSUE #4968: [md-slide-toggle] setting default value to true brings pristine: false
BUG

default value with `true` should not affect the pristine state

pristine state is changed

http://plnkr.co/edit/cCqjrY62mF1WKBSFPWeH?p=preview

Current behavior is somehow disturbing form validation

Angular-4.1.0, Material-2.0.0-beta-6, others are all

DevVersion:
* If a developer sets an initial value for the slide-toggle the NgControl of the slide-toggle can turn dirty. The slide-toggle should only turn dirty if the value changed after initialization. This adds a test to confirm that this is not the case.

Blocked by #4220

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -631,6 +631,21 @@ describe('MdSlideToggle with forms', () => {
       expect(slideToggle.checked)
         .toBe(false, 'Expected slide-toggle to be no longer checked after label click.');
     }));
+
+    it('should be pristine if initial value is set from NgModel', fakeAsync(() => {
+      fixture = TestBed.createComponent(SlideToggleWithModel);
+
+      fixture.componentInstance.modelValue = true;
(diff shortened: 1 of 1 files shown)


DevVersion:
@mmalerba Rebased. Screenshots test failure seems to happen on basically every PR. And is unrelated.

kara:
@DevVersion It looks like we're still calling the forms API `onChange` function every time the value is written programmatically (this is what sets the control as dirty): `writeValue` => `set checked` => `onChange()`. This means that every time the value is set programmatically after the first time, the control will be improperly set to dirty. So the call should be moved. It looks like you fixed this in https://github.com/angular/material2/pull/4220, so this PR shouldn't be necessary once that gets in.

DevVersion:
@kara Makes sense. I removed all other changes and only kept the tests. Marked the PR as blocked for now until #4220 is merged.

PR #4971, opened 2017-06-05
chore: duplicate theme switch button

DevVersion:
* Removes the duplicate theme switch button inside of the demo-app toolbar.


DevVersion:
Chatted with @crisbeto, closing this PR as this will be addressed within #4917

PR #4974, opened 2017-06-05
refactor: replace color getters and setters with mixin

DevVersion:
* Replaces the color getters and setters with a Mixin function (similar as for the disabled property)
* This should reduce payload and also should make it way easier to maintain the color functionality.

Now it would be cool to say the Payload Diff github status 😉

### DIFF
--- a/src/lib/core/common-behaviors/color.ts
+++ b/src/lib/core/common-behaviors/color.ts
@@ -1,0 +1,47 @@
+import {Constructor} from './constructor';
+import {ElementRef, Renderer2} from '@angular/core';
+
+/** List of possible color values that can be set. */
+const AVAILABLE_COLOR_VALUES = ['primary', 'accent', 'warn'];
+
+/** @docs-private */
+export interface IsColorable {
+  color: string;
+}
+
+/** @docs-private */
+export interface ColorableBase {
+  _renderer: Renderer2;
+  _elementRef: ElementRef;
+}
+
+/** Mixin to augment a directive with a `color` property. */
+export function mixinColor<T extends Constructor<ColorableBase>>(base: T, allowNoColor = false)
+    : Constructor<IsColorable> & T {
+  return class extends base {
+    private _color: string = null;
+
+    constructor(...args: any[]) { super(...args); }
+
+    get color(): string { return this._color; }
+    set color(value: string) {
+      if (AVAILABLE_COLOR_VALUES.indexOf(value) !== -1 || (allowNoColor && !value)) {
+        this._setColorClass(this._color, false);
+        this._setColorClass(value, true);
+        this._color = value;
+      }
+    }
+
+    /** Method that changes the color classes on the host element. */
+    private _setColorClass(colorName: string, isAdd: boolean) {
+      if (colorName) {
+        if (isAdd) {
+          this._renderer.addClass(this._elementRef.nativeElement, `mat-${colorName}`);
+        } else {
+          this._renderer.removeClass(this._elementRef.nativeElement, `mat-${colorName}`);
+        }
+      }
+    }
+  };
+}
+
--- a/src/lib/core/common-behaviors/color.ts
+++ b/src/lib/core/common-behaviors/color.ts
@@ -1,31 +1,31 @@
 import {Constructor} from './constructor';
 import {ElementRef, Renderer2} from '@angular/core';
 
-/** List of possible color values that can be set. */
-const AVAILABLE_COLOR_VALUES = ['primary', 'accent', 'warn'];
-
 /** @docs-private */
-export interface IsColorable {
+export interface CanColor {
   color: string;
 }
 
 /** @docs-private */
-export interface ColorableBase {
+export interface HasRenderer {
   _renderer: Renderer2;
   _elementRef: ElementRef;
 }
 
+/** Possible color palette values.  */
+export type ThemePalette = 'primary' | 'accent' | 'warn' | null;
+
 /** Mixin to augment a directive with a `color` property. */
-export function mixinColor<T extends Constructor<ColorableBase>>(base: T, allowNoColor = false)
-    : Constructor<IsColorable> & T {
+export function mixinColor<T extends Constructor<HasRenderer>>(base: T, allowNoColor = false)
+    : Constructor<CanColor> & T {
   return class extends base {
-    private _color: string = null;
+    private _color: ThemePalette = undefined;
 
     constructor(...args: any[]) { super(...args); }
 
-    get color(): string { return this._color; }
-    set color(value: string) {
-      if (AVAILABLE_COLOR_VALUES.indexOf(value) !== -1 || (allowNoColor && !value)) {
+    get color(): ThemePalette { return this._color; }
+    set color(value: ThemePalette) {
+      if (value || allowNoColor && !value) {
         this._setColorClass(this._color, false);
         this._setColorClass(value, true);
         this._color = value;
--- a/src/lib/core/common-behaviors/color.ts
+++ b/src/lib/core/common-behaviors/color.ts
@@ -16,31 +16,32 @@ export interface HasRenderer {
 export type ThemePalette = 'primary' | 'accent' | 'warn' | null;
 
 /** Mixin to augment a directive with a `color` property. */
-export function mixinColor<T extends Constructor<HasRenderer>>(base: T, allowNoColor = false)
+export function mixinColor<T extends Constructor<HasRenderer>>(base: T, defaultColor?: ThemePalette)
     : Constructor<CanColor> & T {
   return class extends base {
-    private _color: ThemePalette = undefined;
-
-    constructor(...args: any[]) { super(...args); }
+    private _color: ThemePalette = null;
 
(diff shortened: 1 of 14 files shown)


src/lib/core/common-behaviors/color.ts
```diff
@@ -0,0 +1,47 @@
+import {Constructor} from './constructor';
+import {ElementRef, Renderer2} from '@angular/core';
+
+/** List of possible color values that can be set. */
+const AVAILABLE_COLOR_VALUES = ['primary', 'accent', 'warn'];
+
+/** @docs-private */
+export interface IsColorable {
```
jelbourn: `CanColor` (I want to consistently start all mixins with `Can`)

src/lib/core/common-behaviors/color.ts
```diff
@@ -0,0 +1,47 @@
...
+  color: string;
+}
+
+/** @docs-private */
+export interface ColorableBase {
```
jelbourn: I'd call this `HasRenderer`

src/lib/core/common-behaviors/color.ts
```diff
@@ -0,0 +1,47 @@
...
+export function mixinColor<T extends Constructor<ColorableBase>>(base: T, allowNoColor = false)
+    : Constructor<IsColorable> & T {
+  return class extends base {
+    private _color: string = null;
+
+    constructor(...args: any[]) { super(...args); }
+
+    get color(): string { return this._color; }
+    set color(value: string) {
+      if (AVAILABLE_COLOR_VALUES.indexOf(value) !== -1 || (allowNoColor && !value)) {
```
jelbourn: Rather than having a run-time check, use string literal values for the type so that it's enforced at compile time:
```ts
protected _color: 'primary' | 'accent' | 'warn' | undefined = undefined;
```

src/lib/core/common-behaviors/color.ts
```diff
@@ -0,0 +1,47 @@
...
+        this._setColorClass(this._color, false);
+        this._setColorClass(value, true);
+        this._color = value;
+      }
+    }
+
+    /** Method that changes the color classes on the host element. */
+    private _setColorClass(colorName: string, isAdd: boolean) {
```
jelbourn: I don't think this method is really necessary; the setter can directly do
```ts
set color(value: string) {
  this._renderer.removeClass(this._elementRef.nativeElement, `mat-${this._color}`);
  this._renderer.addClass(this._elementRef.nativeElement, `mat-${value}`);
  this._color = value;
}

```

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.ts
```diff
@@ -5,9 +5,18 @@ import {
...
   Renderer2,
 } from '@angular/core';
+import {IsColorable, mixinColor} from '../../common-behaviors/color';
 
 export type MdPseudoCheckboxState = 'unchecked' | 'checked' | 'indeterminate';
 
+
+// Boilerplate for applying mixins to MdChip.
+export class MdPseudoCheckboxBase {
+  constructor(public _renderer: Renderer2, public _elementRef: ElementRef) {}
```
jelbourn: Should do this with properties rather than a constructor because [the constructor approach will generate more code](http://www.typescriptlang.org/play/#src=class%20R%20%7B%20%7D%0Aclass%20E%20%7B%20%7D%0A%0A%2F%2F%20----------------------------------------------------------------------------------------%0A%0Aexport%20class%20WithConstructor%20%7B%0A%20%20constructor(public%20_renderer%3A%20R%2C%20public%20_elementRef%3A%20E)%20%7B%7D%0A%7D%0A%0A%2F%2F%20----------------------------------------------------------------------------------------%0A%0Aexport%20class%20WithProperties%20%7B%0A%20%20_renderer%3A%20R%3B%0A%20%20_elementRef%3A%20E%3B%0A%7D%0A%0A%2F%2F%20----------------------------------------------------------------------------------------%0A%0A%0Aclass%20End%20%7B%20%7D).

src/lib/core/common-behaviors/color.ts
DevVersion: This doesn't work properly. Imagine someone sets the value to `null`. A class `mat-null` would be added.

For sure you can place the check if the class truthy inside of the getter too, but it feels better to do it like that.

src/lib/core/common-behaviors/color.ts
DevVersion: Good idea with the types for TypeScript. Although I wanted to do the runtime check because people can still set the color to whatever they want and the class would be added like `mat-test`.

But yeah maybe this doesn't really matter since we enforce it using TypeScript. What do you think?

src/lib/core/common-behaviors/color.ts
DevVersion: Sounds a bit weird, but makes sense to have it consistent

src/lib/core/common-behaviors/color.ts
DevVersion: Should this still live in the color.ts file then?

src/lib/core/common-behaviors/color.ts
jelbourn: Ah, I forgot the check; still think you can do it in the setter directly, though

DevVersion: Yeah true, but I mean it's just easier to read like that.

jelbourn: The main reason I don't like it is the `isAdd` argument:
https://martinfowler.com/bliki/FlagArgument.html

So my thinking is "How do we avoid the boolean argument? Well, it would just be two different methods. Oh, well we already have that on renderer, so just use that."

DevVersion: I agree that this might be an argument to move this to the setter itself but I kinda don't like having another level of nesting here:

```ts
set color(value: ThemePalette) {
  if (value || allowNoColor && !value) {
    if (this._color) {
       this._renderer.removeClass(this._elementRef.nativeElement, `mat-${this._color}`);
     }
     if (value) {
       this._renderer.addClass(this._elementRef.nativeElement, `mat-${value}`);
     }
     this._color = value;
   }
}
 ```

Any other ideas for this? I'm willing to change it though.

Comment (jelbourn) [inline src/lib/core/common-behaviors/color.ts]: That seems fine to me; sometimes that's just how the logic goes.

File: src/lib/core/common-behaviors/color.ts [hunk 6]
```diff
@@ -0,0 +1,47 @@
...
+export function mixinColor<T extends Constructor<HasRenderer>>(base: T, allowNoColor = false)
+    : Constructor<CanColor> & T {
+  return class extends base {
+    private _color: ThemePalette = undefined;
+
+    constructor(...args: any[]) { super(...args); }
+
+    get color(): ThemePalette { return this._color; }
+    set color(value: ThemePalette) {
+      if (value || allowNoColor && !value) {
```
jelbourn: What is `allowNoColor`? I didn't notice that before

DevVersion: Some components allow having no color. E.g buttons don't have by default any color of `primary`, `accent` and `warn`.

jelbourn: Ah, I see it more as being some components _require_ a color. Can we make it so that the argument is instead an optional `defaultColor`?

DevVersion: You mean that we can pass a value that is the default value to the mixin? (e.g `null`)

I'd rather do it the other way around because the default value "should" be basically everything that is falsy.

jelbourn: I mean
```ts
export function mixinColor<T extends Constructor<HasRenderer>>(base: T, defaultColor?: ThemePalette)
    : Constructor<CanColor> & T {
  return class extends base {
    private _color: ThemePalette = defaultColor;

    get color(): ThemePalette { return this._color; }
    set color(value: ThemePalette) {
      const newPalette = value || defaultColor;
      this._updatePaletteClasses(newPalette);
      this._color = newPalette;
    }
    
    constructor(...args: any[]) { 
      super(...args);
      if (defaultColor) {
        this._addPaletteClass(defaultColor);
      } 
    }
    
    private _updatePaletteClasses(newPalette: ThemePalette) {
      if (newPalette != this.color) {
        this._renderer.removeClass(this._elementRef.nativeElement, `mat-${this._color}`);
        this._renderer.addClass(this._elementRef.nativeElement, `mat-${newPalette}`);
      }
    }
  };
}
```

DevVersion: Ah I think I now know what you mean. This seems to be a good idea.

DevVersion: Made this change.

src/lib/core/selection/pseudo-checkbox/pseudo-checkbox.ts
DevVersion: I had to go back to this approach because just setting the properties inside of TypeScript means that the superclass cannot set the default color on initialization (maybe a trade-off)

Issue #4982, opened 2017-06-06
Bug(MdTabs): Translate animation not working in IE 11 / Edge


macjohnny:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
When changing tabs, the content should be animated to translate to the left or to the right.

#### What is the current behavior?
In IE 11 / Edge

#### What are the steps to reproduce?

http://embed.plnkr.co/bXpnE7/

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular 4.1.3, Material 2.0.0-beta.6

#### Is there anything else we should know?
Apparently, this is due to animating with "translate3d()". The sidenav e.g. sets the style property instead of using the animation API, which seems to work.

macjohnny:
This should probably solved similar to MdSidenav:
https://github.com/angular/material2/blob/master/src/lib/sidenav/sidenav.ts#L509

crisbeto:
You need to include a [polyfill for web animations](https://www.npmjs.com/package/web-animations-js), if you want most of the Material animations to work under IE.

macjohnny:
@crisbeto you are right, i forgot to include the polyfill in the plunkr. However, the problem still remains even with the polyfill. I am currently investigating the issue with a simple plunkr example:
https://plnkr.co/edit/AnLKIlfuz5S6siBZQTIY?p=preview
it seems that this one does not work in IE either. so it might be related to the angular animations API.
What is strange though is the fact that when wildly clicking the toggle button or wildly changing tabs, the animation seems to happen, i.e. when the new animation starts before the previous one completed.

macjohnny:
@crisbeto see also https://material.angular.io/components/component/tabs as an example

macjohnny:
the issue here in material is most probably due to a bug in angular: https://github.com/angular/angular/issues/17239

julianobrasil:
@macjohnny , I tested here against IE11 and Edge. 

IE11: Although a little delay, and something that resemble and animation, the rectangle's position is switched form one side to another.

Edge: worked fine

macjohnny:
@julianobrasil can you check what happens in https://fiddle.jshell.net/fu7mrqn0/8 ?
the first example does not work in my IE 11. and I suppose this is exactly the problem with this issue here and the one here: https://github.com/angular/angular/issues/17239

julianobrasil:
@macjohnny , here is the report:

[...]

PR #4986, opened 2017-06-06
Fix tab translation animation in IE 11 (#4982)

macjohnny:
Fixes https://github.com/angular/material2/issues/4982: the tab translation animation not working in IE 11 .
See also https://github.com/angular/angular/issues/17239 and https://github.com/web-animations/web-animations-js/issues/142


macjohnny:
@andrewseguin could you quickly have a look at this pull request? it is very simple and fixes the issue.

PR #4987, opened 2017-06-06
chore: move sanity checks to common module

### LINKED ISSUE #4979: Move sanity checks from CompatibilityModule to MdCommonModule
We originally used `CompatibilityModule` because ti was already included everywhere, but now we have `CommonModule` which is a better place for these checks.

crisbeto:
Moves the sanity checks to the `MdCommonModule`.

Fixes #4979.


andrewseguin:
Looks like `MATERIAL_SANITY_CHECKS` is no longer provided by @angular/material anymore since it isn't included in any index.ts. Previously it was caught by `export * from './compatibility/compatibility';` in `core.ts`

crisbeto:
Re-exported the provider @andrewseguin.

PR #5016, opened 2017-06-07
docs(getting-started): no longer reference out-of-date sample projects

jelbourn:


jelbourn:
Forgot about this. I've since made these changes in another PR

PR #5022, opened 2017-06-08
Bugfix: 0 key is considered as empty

eyalhakim:
When using integers as keys, the value 0 is evaluated as an empty selection, because of the falsey condition.

Instead, we should be explicitly checking for null, empty string or undefined.


eyalhakim:
I signed it!

PR #5024, opened 2017-06-08
Allow using 0 as a key in MdSelect

eyalhakim:
fix(select): allow using 0 as a key

when using integers as keys, the value 0 is evaluated as an empty selection, because of the falsey condition instead, we should be explicitly checking for null, empty string or undefined.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -583,7 +583,7 @@ export class MdSelect extends _MdSelectMixinBase implements AfterContentInit, On
   private _selectValue(value: any): MdOption {
     let optionsArray = this.options.toArray();
     let correspondingOption = optionsArray.find(option => {
-      return option.value != null && option.value === value;
+      return option.value !== null && option.value !== '' && option.value !== undefined && option.value === value;
     });
 
     if (correspondingOption) {


crisbeto:
I'm not sure what this is fixing. [It's already possible to select 0.](http://plnkr.co/edit/uEWZlHX3PXBuKI0WPguy?p=preview)

jelbourn:
@eyalhakim can you elaborate on what you're trying to solve?

eyalhakim:
Hi guys,

I am trying to solve the very issue @crisbeto recreated in his plunker.
My scenario is a bit more complicated, but even when trying to replicate my scenario [using a fork from his environment](http://plnkr.co/edit/oKGYfKIW2soUXyBsFpmm?p=preview), it does work.

However, in my environment, the same (or probably very similar) scenario doesn't work, unless i use the fix that i proposed here.

Any ideas how i can replicate my issue here? i also use the latest version of angular material.

crisbeto:
The fix that allowed falsy values to be selected got merged a few days ago (see https://github.com/angular/material2/commit/2e3910c86c1a1486222f639bb9404aaaf5661d9b). If you're using the latest stable version, you don't have it yet. You can either wait for the next release or install the latest master build:

```
npm install https://github.com/angular/material2-builds.git
```

eyalhakim:
ah-ha, perfect, thanks

eyalhakim:
@crisbeto for some reason, the indigo-pink theme is not included with this package.
Do you have an idea how to solve this?

crisbeto:
It should be in there, but the location might've changed.

eyalhakim:
Couldn't find it, could you direct me to the right path?

eyalhakim:
@crisbeto

crisbeto:
Sorry @eyalhakim, it seems like the prebuilt themes got broken at some point. They should be fixed by https://github.com/angular/material2/pull/5042.

PR #5026, opened 2017-06-08
Alphabetical sorting for Feature Status List

Antonvb:
The feature status list in the project's `README.md` seems to be sorted by feature status (or something else), which isn't very helpful when trying to find the status of a component you care about. Alphabetical sorting would make it easier to identify the feature you are looking for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,45 +59,45 @@ Note that all "planned" work is tentative / best estimate and is subject to chan
 
 | Feature          | Status                              | Docs         | Issue          |
 |------------------|-------------------------------------|--------------|----------------|
+| autocomplete     |                           Available | [README][24] |   [#117][0117] |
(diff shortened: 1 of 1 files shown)


Antonvb:
I signed it!

jelbourn:
Right now they're sorted mostly by status; I'd prefer to keep it that way, though it should be alphabetical for things that have the same status.

Antonvb:
@jelbourn What is the main purpose of the list? If it is meant to give people an option to check the status of a component they care about it feels like alphabetical makes sense to avoid having to browse the whole list randomly. If it's to get an overview of the overall progress of all components the current version gives a better impression. 

Just noticed that when I was trying to find something I resorted to `cmd + F` as reading through the names was a bit tedious.

PR #5029, opened 2017-06-08
build: explicitly add tsc-wrapped to dev dependencies

DevVersion:
The dependency on `tsc-wrapped` should be added explicitly. Our tooling uses the `@angular/tsc-wrapped` package directly and it currently only works because the `@angular/compiler-cli` has a dependency on it.

If not specified explicitly, the `@angular/tsc-wrapped` dependency might be a local dependency of the compiler-cli and our tooling won't be able to access it anymore.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,23 +75,16 @@
       "dev": true
     },
     "@google-cloud/functions-emulator": {
-      "version": "1.0.0-alpha.19",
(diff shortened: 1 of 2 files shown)


andrewseguin:
Needs rebase

DevVersion:
@andrewseguin Done.

PR #5032, opened 2017-06-08
chore: reorganize e2e test files

### LINKED ISSUE #5030: Rename all e2e test files to have "spec" in the name
We currently have, for example, `button.e2e.ts` as the test page and `button.e2e.ts` as the tests themselves. The latter should be `button-e2e.spec.ts`.

The component under test be `button-e2e.ts` instead of `button.e2e.ts`

crisbeto:
* Renames the e2e test file to match the pattern of `{{component}}-e2e.spec.ts`.
* Flattens the directory structure in the `e2e` directory. Having a separate dir for each component was overkill since we only have one file per component.
* Fixes some inconsistent code styles in the Protractor config.

Fixes #5030.


jelbourn:
Just needs rebase

PR #5034, opened 2017-06-08
fix(datepicker): use toggle color from theme

crisbeto:
Fixes the datepicker toggle being black, instead of taking its color from the theme.

**Note:** Overriding this might be a bit simpler if we had the SVG inside the toggle's `template`.

### DIFF
--- a/src/lib/core/style/_escape-color.scss
+++ b/src/lib/core/style/_escape-color.scss
@@ -1,0 +1,10 @@
+// URL-encodes a hex color. Useful when concatenating a color into a URL-encoded SVG.
+@function escape-color($color) {
+  $stringified: $color + '';
+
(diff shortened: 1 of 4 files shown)


mmalerba:
@crisbeto yeah, it would be easier to just make it the toggle's template. The reason I didn't do that yet is that I wanted to actually switch this from a component to a directive so people can use it together with `md-icon-button` like this:

```html
<button md-icon-button mdDatepickerToggle></button>
```

Adding a template kind of seemed like moving in the wrong direction...

mmalerba:
I decided to split the toggle into a separate toggle and icon (see #5317). Prefer that solution even though its a little more verbose since it allows combining toggle with `md-icon-button`

crisbeto:
Cool, closing in favor of #5317.

PR #5036, opened 2017-06-09
build: report ngOnDestroy exceptions in unti tests 

DevVersion:
Monkey-patches TestBed.resetTestingModule such that any errors that occur during component destruction are thrown instead of silently logged. Also runs TestBed.resetTestingModule after each unit test.

 Without this patch, the combination of two behaviors is problematic for Angular Material:
 - TestBed.resetTestingModule catches errors thrown on fixture destruction and logs them without the errors ever being thrown. This means that any component errors that occur in ngOnDestroy can encounter errors silently and still pass unit tests.
 - TestBed.resetTestingModule is only called *before* a test is run, meaning that even *if* the aforementioned errors were thrown, they would be reported for the wrong test (the test that's about to start, not the test that just finished).

### DIFF
--- a/test/karma-test-shim.js
+++ b/test/karma-test-shim.js
@@ -87,9 +87,50 @@ function configureTestBed() {
 
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000;
 
-    testing.TestBed.initTestEnvironment(
+    var testBed = testing.TestBed.initTestEnvironment(
       testingBrowser.BrowserDynamicTestingModule,
       testingBrowser.platformBrowserDynamicTesting()
     );
+
+    patchTestBedToDestroyFixturesAfterEveryTest(testBed);
+  });
+}
+
+/**
+ * Monkey-patches TestBed.resetTestingModule such that any errors that occur during component
+ * destruction are thrown instead of silently logged. Also runs TestBed.resetTestingModule after
+ * each unit test.
+ *
+ * Without this patch, the combination of two behaviors is problematic for Angular Material:
+ * - TestBed.resetTestingModule catches errors thrown on fixture destruction and logs them without
+ *     the errors ever being thrown. This means that any component errors that occur in ngOnDestroy
+ *     can encounter errors silently and still pass unit tests.
+ * - TestBed.resetTestingModule is only called *before* a test is run, meaning that even *if* the
+ *    aforementioned errors were thrown, they would be reported for the wrong test (the test that's
+ *    about to start, not the test that just finished).
+ */
+function patchTestBedToDestroyFixturesAfterEveryTest(testBed) {
+  // Original resetTestingModule function of the TestBed.
+  var _resetTestingModule = testBed.resetTestingModule;
+
+  // Monkey-patch the resetTestingModule to destroy fixtures outside of a try/catch block.
+  // With https://github.com/angular/angular/commit/2c5a67134198a090a24f6671dcdb7b102fea6eba
+  // errors when destroying components are no longer causing Jasmine to fail.
+  testBed.resetTestingModule = function() {
+    try {
+      this._activeFixtures.forEach(function (fixture) { fixture.destroy(); });
+    } finally {
+      this._activeFixtures = [];
+      // Regardless of errors or not, run the original reset testing module function.
+      _resetTestingModule.call(this);
+    }
+  };
+
+  // Angular's testing package resets the testing module before each test. This doesn't work well
+  // for us because it doesn't allow developers to see what test actually failed.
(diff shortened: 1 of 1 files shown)


src/lib/datepicker/datepicker.ts
```diff
@@ -257,7 +257,7 @@ export class MdDatepicker<D> implements OnDestroy {
       this._ngZone.onStable.first().subscribe(() => this._popupRef.updatePosition());
     }
 
-    this._popupRef.backdropClick().first().subscribe(() => this.close());
+    this._popupRef.backdropClick().subscribe(() => this.close());
```
jelbourn: Should already be fixed in master
https://github.com/angular/material2/commit/f89c6dbe905f0508810dda48fca55d4b41615764

test/karma-test-shim.js
```diff
@@ -87,9 +87,38 @@ function configureTestBed() {
...
       testingBrowser.BrowserDynamicTestingModule,
       testingBrowser.platformBrowserDynamicTesting()
     );
+
+    applyAngularTestingPatches(testBed);
+  });
+}
+
+/** Applies patches and workarounds to Angular's testing package. */
+function applyAngularTestingPatches(testBed) {
```
jelbourn: How about `patchTestModuleToDestroyFixturesAfterEveryTest`?

```js
/**
 * Monkey-patches TestBed.resetTestingModule such that any errors that occur during component
 * destruction are thrown instead of silently logged. Also runs TestBed.resetTestingModule after
 * each unit test.
 *
 * Without this patch, the combination of two behaviors is problematic for Angular Material:
 * - TestBed.resetTestingModule catches errors thrown on fixture destruction and logs them without
 *     the errors ever being thrown. This means that any component errors that occur in ngOnDestroy
 *     can encounter errors silently and still pass unit tests.
 * - TestBed.resetTestingModule is only called *before* a test is run, meaning that even *if* the
 *    aforementioned errors were thrown, they would be reported for the wrong test (the test that's
 *    about to start, not the test that just finished).
 */
function patchTestBedToDestroyFixturesAfterEveryTest() {
}
```

test/karma-test-shim.js
```diff
@@ -87,9 +87,38 @@ function configureTestBed() {
...
+  // Original resetTestingModule function of the TestBed.
+  var _resetTestingModule = testBed.resetTestingModule;
+
```
jelbourn: Maybe add something like...
```js
if (!testBed._activeFixtures) {
  throw Error('The monkey-patch for TestBed.resetTestingModule has failed ' +
   'because a new version of Angular has removed the _activeFixtures member. ' +
   'This needs to be fixed in karma-test-shim.js');
}
```

test/karma-test-shim.js
```diff
@@ -87,9 +87,38 @@ function configureTestBed() {
...
+    try {
+      this._activeFixtures.forEach(function (fixture) { fixture.destroy(); });
+    } finally {
+      this._activeFixtures = [];
+      // Regardless of errors or not, run the original reset testing module function.
+      _resetTestingModule.call(this);
+    }
+  };
+
+  // Angular's testing package resets the testing module before each test. This is bad because
```
jelbourn: "This is bad" -> "This doesn't work well for us"

jelbourn:
Also the name of the PR / commit should reflect the content of the karma-shim change

test/karma-test-shim.js
DevVersion: I don't think this is necessary. If it will fail we will immediately see that the shims cause it to fail.

test/karma-test-shim.js
DevVersion: Yeah makes sense.

test/karma-test-shim.js
DevVersion: Feels a bit long but should be fine for the shims.

DevVersion:
I only mentioned the datepicker because if it gets merged the changelog should only include that the datepicker bug was fixed.

But since the fix is merged already it makes sense to make it a `chore:` now.

test/karma-test-shim.js
jelbourn: The error that would be thrown is `Cannot read property 'forEach' of undefined`. We have much more knowledge about this failure state, so we can help whoever is debugging this in the future by making a much more helpful message.

DevVersion: That's true, but what I mean is that it's very unlikely that the `_activeFixtures` property will be removed or whatever.

If it really comes to an error you will see the file that failed and the according line number. 

Would be `test/karma-test-shim.js:XXX`

DevVersion:
Addressed feedback.

PR #5045, opened 2017-06-09
build: run fewer browsers on saucelabs

DevVersion:
* Better balance between Saucelabs and Browserstack because Browserstack runs more stable. Considering that e2e tests also run on Saucelabs it's only fair running more on Browserstack.
* Sets browsers that are optional and don't run at all to `null`. Just to have a nice overview which browsers are running or not.
* Runs tests against Safari 9 and 10.

### DIFF
--- a/test/browser-providers.js
+++ b/test/browser-providers.js
@@ -5,9 +5,9 @@
  * Target can be either: BS (Browserstack) | SL (Saucelabs) | null (To not run at all)
  */
 const browserConfig = {
-  'Chrome':       { unitTest: {target: 'SL', required: true  }},
-  'Firefox':      { unitTest: {target: 'SL', required: true  }},
-  'ChromeBeta':   { unitTest: {target: 'SL', required: false }},
+  'Chrome':       { unitTest: {target: 'BS', required: true  }},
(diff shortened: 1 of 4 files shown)


test/browser-providers.js
```diff
@@ -19,15 +19,15 @@ const browserConfig = {
...
   'Android4.3':   { unitTest: {target: null, required: false }},
   'Android4.4':   { unitTest: {target: null, required: false }},
-  'Android5':     { unitTest: {target: 'SL', required: false }},
+  'Android5':     { unitTest: {target: null, required: false }},
   'Safari7':      { unitTest: {target: null, required: false }},
   'Safari8':      { unitTest: {target: null, required: false }},
-  'Safari9':      { unitTest: {target: 'BS', required: false }},
-  'Safari10':     { unitTest: {target: 'SL', required: false }},
+  'Safari9':      { unitTest: {target: null, required: false }},
+  'Safari10':     { unitTest: {target: null, required: false }},
```
jelbourn: We should be running Safari 9 and 10; I didn't realize these were not marked as required

DevVersion: Done. Please take another look.

PR #5060, opened 2017-06-10
build: fix deploying of screenshot functions

DevVersion:
* Updates the dashboard and screenshot deploy script to exit the process immediately if the Firebase token is not set.
* Fixes that the error message of the dashboard still shows the old name of the token variable
* Renames the screenshot variables to follow the dashboard environment variables
* Explicitly specifies the project that will be uploaded in the screenshot deploy script (same as for dashboard)

@jelbourn You already renamed the dashboard variables a few days ago. Can you rename the screenshot variables too? So we have those variables consistent and it's clear what every variable does.

* `MATERIAL2_SCREENSHOT_ACCESS_TOKEN` -> `MATERIAL2_SCREENSHOT_FIREBASE_DEPLOY_TOKEN`
* `MATERIAL2_SCREENSHOT_FIREBASE_KEY`  -> `MATERIAL2_SCREENSHOT_FIREBASE_SERVICE_KEY`

### DIFF
--- a/scripts/deploy/deploy-screenshot-functions.sh
+++ b/scripts/deploy/deploy-screenshot-functions.sh
@@ -3,18 +3,25 @@
 # The script should immediately exit if any command in the script fails.
 set -e
 
-# Go to the project root directory
-cd $(dirname ${0})/../..
-
-# Install node modules for screenshot functions. Firebase CLI needs to execute the functions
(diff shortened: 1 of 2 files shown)


DevVersion:
@jelbourn temporary fixed this by using the old token. At some point I will submit another PR to have the tokens consistent. See #5655

PR #5066, opened 2017-06-11
fix(input,cdk): a couple of server-side rendering errors

crisbeto:
* Fixes an error that is thrown during server-side rendering if an input has a `type`.
* Fixes an error that was being thrown by the `cdkObserveContent` due to the lack of a `MutationObserver`.

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -268,6 +268,10 @@ export class MdInputDirective {
   /** Determines if the component host is a textarea. If not recognizable it returns false. */
   private _isTextarea() {
     let nativeElement = this._elementRef.nativeElement;
+
+    // In Universal, we don't have access to `nodeName`, but the same can be achieved with `name`.
+    // Note that this shouldn't be necessary once Angular switches to an API that resembles the
+    // DOM closer.
     let nodeName = this._platform.isBrowser ? nativeElement.nodeName : nativeElement.name;
     return nodeName ? nodeName.toLowerCase() === 'textarea' : false;
   }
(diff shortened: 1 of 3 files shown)


crisbeto:
@jelbourn The `cdkObserveContent` error is a little weird. For some reason it is logged, but ends up getting swallowed and doesn't fail the build.

src/lib/input/input-container.ts
```diff
@@ -267,7 +268,8 @@ export class MdInputDirective {
   /** Determines if the component host is a textarea. If not recognizable it returns false. */
   private _isTextarea() {
     let nativeElement = this._elementRef.nativeElement;
-    return nativeElement ? nativeElement.nodeName.toLowerCase() === 'textarea' : false;
+    let nodeName = this._platform.isBrowser ? nativeElement.nodeName : nativeElement.name;
```
jelbourn: Should add a comment that explains why this falls back to name
(also shouldn't be necessary long-term since platform-server is changing to mimic more DOM APIs)

src/universal-app/kitchen-sink/kitchen-sink.html
```diff
@@ -86,7 +86,7 @@
 <h2>Datepicker</h2>
 
 <md-input-container>
-  <input mdInput [mdDatepicker]="birthday" placeholder="Birthday">
+  <input type="text" mdInput [mdDatepicker]="birthday" placeholder="Birthday">
```
jelbourn: Is there any problem with leaving the type empty?

crisbeto: Not really, I left it in there to avoid regressions.

src/lib/input/input-container.ts
crisbeto: Done.

PR #5068, opened 2017-06-11
feat(radio): support for color binding

### LINKED ISSUE #4677: md-radio color support
Feature request

md-radio should support the color property like md-checkbox does

does not support it

see docs :-)

I need a radio button in primary color :-(

ng 4.1.x, md beta 5

DevVersion:
* Introduces the `color` binding for radio buttons. By default every selection control uses accent as per Material design specifications.

Closes #4677

### DIFF
--- a/src/demo-app/radio/radio-demo.html
+++ b/src/demo-app/radio/radio-demo.html
@@ -4,6 +4,15 @@
   <md-radio-button name="group1">Option 2</md-radio-button>
   <md-radio-button name="group1" disabled="true">Option 3 (disabled)</md-radio-button>
 </section>
+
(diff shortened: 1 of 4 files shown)


chriszrc:
:100:

DevVersion:
@chriszrc Sorry for that. Since the `color`, `tabIndex`, `rippleDisabled` properties are mostly created by mixin functions to reduce the payload of Angular Material.. those aren't showing up in the API docs (yet).

We definitely want to fix that, but our docs generation tool is not yet able to understand such mixins, because it kind of isn't statically analyzable. Tracking this issue with https://github.com/angular/material2/issues/5284

PR #5092, opened 2017-06-12
build: remove url from payload status

DevVersion:
* Removes the invalid URL that will be shown in the payload status
* If the payload difference is zero there should be a plus instead of the plain number (Keeping the message simple for now. No need to construct complex descriptions)


DevVersion:
Also fixed that the payload difference will be rounded.

PR #5093, opened 2017-06-12
feat(data-table): allow dynamic data source

andrewseguin:


src/lib/core/data-table/data-table.ts
```diff
@@ -77,17 +94,34 @@ export class CdkTable<T> implements CollectionViewer {
...
+   * stream of view window (what rows are currently on screen).
    */
-  private _columnDefinitionsByName = new Map<string,  CdkColumnDef>();
-
-  /** Differ used to find the changes in the data provided by the data source. */
-  private _dataDiffer: IterableDiffer<T> = null;
+  @Input()
+  get dataSource(): DataSource<T> { return this._dataSource; }
+  set dataSource(dataSource: DataSource<T>) {
+    if (this._dataSource == dataSource) { return; }
```
ErinCoughlan: triple equals?

src/lib/core/data-table/data-table.ts
```diff
@@ -123,12 +157,11 @@ export class CdkTable<T> implements CollectionViewer {
...
     // TODO(andrewseguin): Add trackby function input.
     // Find and construct an iterable differ that can be used to find the diff in an array.
-    this._dataDiffer = this._differs.find([]).create();
+    this._dataDiffer = this._differs.find(this._data).create();
   }
 
   ngOnDestroy() {
-    // TODO(andrewseguin): Disconnect from the data source so
-    //   that it can unsubscribe from its streams.
+    this._onDestroy.next();
```
ErinCoughlan: I believe you want to complete the subject so it won't emit anymore and the datasource listener will also finish.

this._onDestroy.complete();

src/lib/core/data-table/data-table.ts
```diff
@@ -142,40 +175,54 @@ export class CdkTable<T> implements CollectionViewer {
...
+          this._renderRowChanges();
+        });
   }
 
   /**
    * Create the embedded view for the header template and place it in the header row view container.
    */
-  renderHeaderRow() {
-    const cells = this.getHeaderCellTemplatesForRow(this._headerDefinition);
+  _renderHeaderRow() {
```
ErinCoughlan: Can this be marked private? Or is it used somewhere else? Ditto to some of the other methods in this class.

src/lib/core/data-table/data-table.ts
```diff
@@ -77,17 +94,34 @@ export class CdkTable<T> implements CollectionViewer {
...
    */
-  private _columnDefinitionsByName = new Map<string,  CdkColumnDef>();
-
-  /** Differ used to find the changes in the data provided by the data source. */
-  private _dataDiffer: IterableDiffer<T> = null;
+  @Input()
+  get dataSource(): DataSource<T> { return this._dataSource; }
+  set dataSource(dataSource: DataSource<T>) {
+    if (this._dataSource == dataSource) { return; }
+    this._data = [];
```
jelbourn: Can you move most of this setter out into a separate function like `_updateDataSource` or `_switchDataSource`?

src/lib/core/data-table/data-table.ts
```diff
@@ -142,40 +175,54 @@ export class CdkTable<T> implements CollectionViewer {
...
 
-    // Get and merge the streams for column changes made to the row defs
-    const rowDefs = [...this._rowDefinitions.toArray(), this._headerDefinition];
-    const columnChangeStreams =
-        rowDefs.map((rowDef: BaseRowDef) => rowDef.columnsChange);
-    this._columnsChange = Observable.merge(...columnChangeStreams);
+    // Re-render the rows if any of their columns change.
+    // TODO(andrewseguin): Determine how to only re-render the rows that have their columns changed.
+    const rowColumnsChange =
+        this._rowDefinitions.toArray().map((rowDef: BaseRowDef) => rowDef.columnsChange);
```
jelbourn: QueryList has a `map` function so you don't need `toArray`
https://angular.io/api/core/QueryList

You also shouldn't need a type in the map function since it is given by the generic

src/lib/core/data-table/data-table.ts
andrewseguin: I believe it goes either way, according to [this documentation](http://reactivex.io/documentation/operators/takeuntil.html): `If this second Observable emits an item or sends a termination notification, the Observable returned by TakeUntil stops mirroring the source Observable and terminates.`

Though I do agree that `complete()` looks nicer and more definitive, will change to that.

src/lib/core/data-table/data-table.ts
andrewseguin: Good catch, thanks

src/lib/core/data-table/data-table.ts
```diff
@@ -161,7 +145,7 @@ export class CdkTable<T> implements CollectionViewer {
   }
 
   ngOnDestroy() {
-    this._onDestroy.next();
+    this._onDestroy.complete();
```
willshowell: I think `next()` is still required for `takeUntil`([example](http://plnkr.co/edit/Rjxn1GDA8RbXUtzWYIhc?p=preview))

I suspect @ErinCoughlan was suggesting you use both

andrewseguin: Thanks for the heads up, I appreciate the plunker to help me understand. Looking at the [RxJS code](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/perf/operators/takeuntil.js) does indicate that it doesn't look to the completion event.

`Returns the values from the source observable sequence until the other observable sequence produces a value.`

This doesn't line up with the documentation from reactivex.io which says that either an emitted value or termination notification would end it.

Will change it to be next() and complete(). Not totally sure the complete() is necessary but we can discuss another time.

willshowell: Looks like the rxjs docs will eventually be updated to reflect behavior being inconsistent with reactivex.io https://github.com/ReactiveX/rxjs/issues/2160

Also don't think complete() is required, but as far as I've seen, [this SO answer](https://stackoverflow.com/a/41177163/4811678) has been the recommended approach.

andrewseguin: SGTM, thanks for the context. We'll have to continue looking at this for other components where we want to unsubscribe on destroy

PR #5094, opened 2017-06-12
feat: nullability support

crisbeto:
Adds compatibility with strictNullChecks to the library, tests, build and various test apps.

Fixes #3486.

### DIFF
--- a/src/lib/button-toggle/button-toggle.ts
+++ b/src/lib/button-toggle/button-toggle.ts
@@ -51,7 +51,7 @@ let _uniqueIdCounter = 0;
 /** Change event object emitted by MdButtonToggle. */
 export class MdButtonToggleChange {
   /** The MdButtonToggle that emits the event. */
-  source: MdButtonToggle;
+  source: MdButtonToggle | null;
   /** The value assigned to the MdButtonToggle. */
   value: any;
 }
@@ -81,7 +81,7 @@ export class MdButtonToggleGroup extends _MdButtonToggleGroupMixinBase implement
   private _vertical: boolean = false;
 
   /** The currently selected button toggle, should match the value. */
-  private _selected: MdButtonToggle = null;
+  private _selected: MdButtonToggle | null = null;
 
   /** Whether the button toggle group is initialized or not. */
   private _isInitialized: boolean = false;
@@ -96,8 +96,7 @@ export class MdButtonToggleGroup extends _MdButtonToggleGroupMixinBase implement
   onTouched: () => any = () => {};
 
   /** Child button toggle buttons. */
-  @ContentChildren(forwardRef(() => MdButtonToggle))
-  _buttonToggles: QueryList<MdButtonToggle> = null;
+  @ContentChildren(forwardRef(() => MdButtonToggle)) _buttonToggles: QueryList<MdButtonToggle>;
 
   ngAfterViewInit() {
     this._isInitialized = true;
@@ -150,7 +149,7 @@ export class MdButtonToggleGroup extends _MdButtonToggleGroupMixinBase implement
     return this._selected;
   }
 
-  set selected(selected: MdButtonToggle) {
+  set selected(selected: MdButtonToggle | null) {
     this._selected = selected;
     this.value = selected ? selected.value : null;
 
@@ -279,13 +278,13 @@ export class MdButtonToggle implements OnInit {
   _type: ToggleType;
 
   /** Whether or not this button toggle is disabled. */
-  private _disabled: boolean = null;
+  private _disabled: boolean = false;
 
   /** Value assigned to this button toggle. */
   private _value: any = null;
 
   /** Whether or not the button toggle is a single selection. */
-  private _isSingleSelector: boolean = null;
+  private _isSingleSelector: boolean = false;
 
   @ViewChild('input') _inputElement: ElementRef;
 
@@ -355,7 +354,7 @@ export class MdButtonToggle implements OnInit {
   }
 
   set disabled(value: boolean) {
-    this._disabled = (value != null && value !== false) ? true : null;
+    this._disabled = coerceBooleanProperty(value);
   }
 
   /** Event emitted when the group value changes. */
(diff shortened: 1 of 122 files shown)


src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -318,7 +318,7 @@ class SlideToggleRenderer {
 
   /** Resets the current drag and returns the new checked value. */
   stopThumbDrag(): boolean {
-    if (!this.dragging) { return; }
+    if (!this.dragging) { return this.dragging; }
```
DevVersion: This is can be always `return false`?

src/lib/core/portal/portal.ts
```diff
@@ -76,15 +77,15 @@ export class ComponentPortal<T> extends Portal<ComponentRef<T>> {
...
   /** [Optional] Injector used for the instantiation of the component. */
-  injector: Injector;
+  injector?: Injector;
 
   constructor(
       component: ComponentType<T>,
-      viewContainerRef: ViewContainerRef = null,
-      injector: Injector = null) {
+      viewContainerRef?: ViewContainerRef,
+      injector?: Injector) {
```
jelbourn: Could make these `| null` just to be more lenient (anywhere with multiple optional params where you might want to pass a latter one but an earlier one)

src/lib/core/ripple/ripple.spec.ts
```diff
@@ -7,17 +7,15 @@ import {
...
 
 describe('MdRipple', () => {
   let fixture: ComponentFixture<any>;
   let rippleTarget: HTMLElement;
-  let originalBodyMargin: string;
+  let originalBodyMargin: string | null;
   let viewportRuler: ViewportRuler;
 
+  /** Extracts the numeric value of a pixel size string like '123px'.  */
+  const pxStringToFloat = (s: string | null) => s ? parseFloat(s.replace('px', '')) : 0;
```
jelbourn: I know it was already like this, but doesn't `parseFloat` already ignore trailing non-numeric characters already?

src/lib/dialog/dialog.spec.ts
```diff
@@ -405,10 +403,11 @@ describe('MdDialog', () => {
     });
 
     it('should default to null if no data is passed', () => {
-      let dialogRef: MdDialogRef<DialogWithInjectedData>;
-
-      expect(() => dialogRef = dialog.open(DialogWithInjectedData)).not.toThrow();
-      expect(dialogRef.componentInstance.data).toBeNull();
+      // TODO: double check that this does what i expect
```
jelbourn: ?

src/lib/icon/icon-registry.ts
```diff
@@ -33,12 +33,22 @@ export function getMdIconNoHttpProviderError(): Error {
...
 }
 
+
+/**
+ * Returns an exception to be thrown when a URL couldn't be sanitized.
+ * @param url URL that was attempted to be sanitized.
+ * @docs-private
+ */
+export function getMdIconFailedToSanitizeError(url: SafeResourceUrl): Error {
+  return new Error(`MdIconRegistry has failed to sanitize the following URL: ${url}.`);
```
jelbourn: A more accurate error message would be something like "The url provided to MdIconRegistry was not trusted as a resource URL via Angular's DomSanitizer"

src/lib/menu/menu-item.ts
```diff
@@ -44,7 +44,7 @@ export class MdMenuItem implements Focusable {
   }
 
   /** Used to set the HTML `disabled` attribute. Necessary for links to be disabled properly. */
-  _getDisabledAttr(): boolean {
+  _getDisabledAttr(): true | null {
```
jelbourn: I'm a fan of doing `disabled || null` in the host binding

This means that the attribute is correctly removed in the DOM but programmatic APIs return the expected `true`/`false`

src/lib/button-toggle/button-toggle.ts
```diff
@@ -65,13 +65,13 @@ export class MdButtonToggleGroup implements AfterViewInit, ControlValueAccessor
   private _name: string = `md-button-toggle-group-${_uniqueIdCounter++}`;
 
   /** Disables all toggles in the group. */
-  private _disabled: boolean = null;
+  private _disabled: boolean | null = null;
```
crisbeto: It's because the `disabled` attribute needs to be removed from the DOM completely when enabled. We could also switch the bindings to be something like `disabled || null`.

src/lib/core/overlay/position/connected-position-strategy.ts
```diff
@@ -94,7 +94,7 @@ export class ConnectedPositionStrategy implements PositionStrategy {
    * @param element Element to which to apply the CSS styles.
    * @returns Resolves when the styles have been applied.
    */
-  apply(element: HTMLElement): Promise<void> {
+  apply(element: HTMLElement): Promise<null> {
```
crisbeto: Not sure what that would look like, `null | undefined`? Otherwise I'm in favor of removing the promise completely, although I could see it being useful if we decide to introduce animations into the overlays.

src/lib/core/portal/portal.ts
crisbeto: I went with optionals here, because they were marked as `[Optional]` a little bit up in the doc strings for the properties.

src/lib/core/ripple/ripple.spec.ts
crisbeto: It does. This can be changed to `parseFloat(...) || 0`.

src/lib/dialog/dialog.spec.ts
crisbeto: Ops, forgot to follow up on this one. I wanted to check that the `expect` inside the `expect.not.toThrow` won't cause any issues. I'll check it and refactor or remove the todo.

crisbeto:
Rebased and addressed the feedback @jelbourn.

src/lib/core/portal/portal.ts
jelbourn: Sorry, I accidentally omitted the word "also". Meant that we could do
`viewContainerRef?: ViewContainerRef | null,` so that either `undefined` or `null` can be passed in cases when you don't care for that argument.

crisbeto: Makes sense. Added.

kara:
@crisbeto  This one needs rebase too, when you get a chance.

jelbourn:
I rebased it (was trivial)

PR #5112, opened 2017-06-13
fix(select): consider value changes via arrow keys on closed select as user actions

### LINKED ISSUE #5084: isUserInput == false when cycling through closed select
Bug

#### Expected behavior
When cycling through options in a _closed_ select with the <kbd>up</kbd> and <kbd>down</kbd> keys, I'd expect the `MdOptionSelectionChange` to have `isUserInput == true`

#### Current behavior
`isUserInput` is only true when selected from an open select.

#### Reproduction
http://plnkr.co/edit/YQn77sSXj60cNwLYldOC?p=preview

I want to trigger an event when the user

1. Opens the select and changes the value with mouse or keyboard
2. Opens the select and reselects the current value mouse or keyboard
3. Cycles to a new option while the select is closed

It's easy to achieve **1** and **3** with the the select's `(change)` event. In order to get **2**, I have to instead watch each option's `(onSelectionChange)` event and filter by `event.source.selected`. At this point, I cannot tell the difference between changes made by the ControlValueAccessor interface or by the user cycling through the closed select.

Latest (c2816ef)

I may be misinterpreting the use of `isUserInput` but it would seem that cycling through a closed select would be considered "selection coming from the user".

crisbeto:
Currently, the user changing the `md-select` value using the arrow keys on a closed select won't mark the interaction as `isUserInput` in the option's selection event. These changes correct the behavior.

Fixes #5084.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -29,6 +29,7 @@ import {
   FloatPlaceholderType,
   MD_PLACEHOLDER_GLOBAL_OPTIONS
 } from '../core/placeholder/placeholder-options';
+import 'rxjs/add/operator/map';
 
 
 describe('MdSelect', () => {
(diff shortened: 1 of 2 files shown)


src/lib/select/select.spec.ts
```diff
@@ -29,6 +29,7 @@ import {
   FloatPlaceholderType,
   MD_PLACEHOLDER_GLOBAL_OPTIONS
 } from '../core/placeholder/placeholder-options';
+import 'rxjs/add/operator/map';
```
kara: Given that we are trying to move away from this syntax, can you refactor to use the explicit symbol?

crisbeto: Sure, but I would rather do them all the same time when redoing #2652.

crisbeto:
@kara could you take another look?

PR #5120, opened 2017-06-13
fix(radio): label vertical alignment

crisbeto:
Properly aligns the radio button label with the circle icon vertically.


tinayuangao:
LGTM

PR #5121, opened 2017-06-13
fix(button-toggle): Fix standalone button toggle style

### LINKED ISSUE #4969: button-toggle alignment and material.angular.io issues
1) combination of text and icon is not aligned well 

image

2) styles in overview on material.angular.io are broken, or not working
image

tinayuangao:
Add styles for standalone button-toggle without any button-toggle-group.
Fixes #4969

### DIFF
--- a/src/lib/button-toggle/button-toggle.scss
+++ b/src/lib/button-toggle/button-toggle.scss
@@ -7,6 +7,7 @@ $mat-button-toggle-padding: 0 16px !default;
 $mat-button-toggle-height: 36px !default;
 $mat-button-toggle-border-radius: 2px !default;
 
+:not(.mat-button-toggle-group) > .mat-button-toggle,
 .mat-button-toggle-group {
   @include mat-elevation(2);
   position: relative;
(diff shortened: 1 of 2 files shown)


src/lib/button-toggle/button-toggle.scss
```diff
@@ -7,6 +7,7 @@ $mat-button-toggle-padding: 0 16px !default;
 $mat-button-toggle-height: 36px !default;
 $mat-button-toggle-border-radius: 2px !default;
 
+:not(.mat-button-toggle-group) > .mat-button-toggle,
```
jelbourn: Could we do this instead by having a css class like `mat-button-toggle-standalone` that is applied via host binding
```ts
host: {
  '[class.mat-button-toggle-standalone]': 'buttonToggleGroup',
}
```

That way the styles will have a much lower specificity and will be easier to override

tinayuangao: Thanks. Changed to host binding.

PR #5122, opened 2017-06-13
build: get previous payload by querying git

DevVersion:
Currently the previous payload is just the latest uploaded payload result on Firebase. This is problematic because it can happen that the current payload is uploaded "too fast" and the payload task is comparing against itself.

Just calculating before uploading could work most of the time, but there it would still fail if another Travis Job uploades payload results at the same time.

The safest solution would be resolving the parent commit of the current Travis job. This way the payload diff would be always correct (even if multiple Jobs are running).

### DIFF
--- a/tools/gulp/tasks/payload.ts
+++ b/tools/gulp/tasks/payload.ts
@@ -157,10 +157,19 @@ async function getPayloadResults(database: firebaseAdmin.database.Database, comm
 function getCommitFromPreviousPayloadUpload(): string {
   if (isTravisMasterBuild()) {
     const commitRange = process.env['TRAVIS_COMMIT_RANGE'];
+    // In some situations Travis will include multiple commits in a single Travis Job. This means
+    // that we can't just resolve the previous commit by using the parent commit of HEAD.
+    // By resolving the amount of commits inside of the current Travis Job we can figure out
+    // how many commits before HEAD the last Travis Job ran.
     const commitCount = spawnSync('git', ['rev-list', '--count', commitRange]).stdout
       .toString().trim();
+    // With the amount of commits inside of the current Travis Job we can query Git to print
+    // the SHA of the commit that ran before this Travis Job was created.
     return spawnSync('git', ['rev-parse', `HEAD~${commitCount}`]).stdout.toString().trim();
   } else {
+    // Travis applies the changes of Pull Requests in new branches. This means that resolving
+    // the commit, that previously ran on the target branch (mostly "master"), can be done
+    // by just loading the SHA of the most recent commit in the target branch.
     return spawnSync('git', ['rev-parse', process.env['TRAVIS_BRANCH']]).stdout.toString().trim();
   }
 }
--- a/tools/gulp/tasks/payload.ts
+++ b/tools/gulp/tasks/payload.ts
@@ -142,7 +142,7 @@ async function uploadPayloadResults(database: firebaseAdmin.database.Database, c
   }
 }
 
-/** Gets the last payload uploaded from previous Travis builds. */
+/** Gets payload results of the specified commit sha. */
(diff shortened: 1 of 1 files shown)


tools/gulp/tasks/payload.ts
```diff
@@ -138,16 +142,25 @@ async function uploadPayloadResults(database: firebaseAdmin.database.Database, c
...
-  const results = snapshot.val();
+  return snapshot.val();
+}
 
-  return snapshot.hasChildren() ? results[Object.keys(results)[0]] : null;
+/** Gets the SHA of the commit where the payload was uploaded before this Travis Job started. */
+function getCommitFromPreviousPayloadUpload(): string {
+  if (isTravisMasterBuild()) {
+    const commitRange = process.env['TRAVIS_COMMIT_RANGE'];
+    const commitCount = spawnSync('git', ['rev-list', '--count', commitRange]).stdout
```
jelbourn: Can you add comments that write out the commit commands and describe what you're getting from them?

DevVersion: Added comments. It was pretty hard to explain what those Git calls do. Please me know if you think something needs to be better explained.

tools/gulp/tasks/payload.ts
```diff
@@ -138,16 +142,34 @@ async function uploadPayloadResults(database: firebaseAdmin.database.Database, c
...
-  const results = snapshot.val();
+  return snapshot.val();
+}
 
-  return snapshot.hasChildren() ? results[Object.keys(results)[0]] : null;
+/** Gets the SHA of the commit where the payload was uploaded before this Travis Job started. */
+function getCommitFromPreviousPayloadUpload(): string {
+  if (isTravisMasterBuild()) {
+    const commitRange = process.env['TRAVIS_COMMIT_RANGE'];
+    // In some situations Travis will include multiple commits in a single Travis Job. This means
```
kara: `In some situations Travis` -> `In some situations, Travis`

tools/gulp/tasks/payload.ts
```diff
@@ -138,16 +142,34 @@ async function uploadPayloadResults(database: firebaseAdmin.database.Database, c
...
+    // that we can't just resolve the previous commit by using the parent commit of HEAD.
+    // By resolving the amount of commits inside of the current Travis Job we can figure out
```
kara: `By resolving the amount of commits inside of the current Travis Job we can` -> `By resolving the amount of commits inside of the current Travis job, we can`

tools/gulp/tasks/payload.ts
```diff
@@ -138,16 +142,34 @@ async function uploadPayloadResults(database: firebaseAdmin.database.Database, c
...
+    // how many commits before HEAD the last Travis Job ran.
+    const commitCount = spawnSync('git', ['rev-list', '--count', commitRange]).stdout
+      .toString().trim();
+    // With the amount of commits inside of the current Travis Job we can query Git to print
```
kara: `With the amount of commits inside of the current Travis job we can` -> `With the amount of commits inside of the current Travis job, we can`

tools/gulp/tasks/payload.ts
```diff
@@ -138,16 +142,34 @@ async function uploadPayloadResults(database: firebaseAdmin.database.Database, c
...
+    // the SHA of the commit that ran before this Travis Job was created.
+    return spawnSync('git', ['rev-parse', `HEAD~${commitCount}`]).stdout.toString().trim();
+  } else {
+    // Travis applies the changes of Pull Requests in new branches. This means that resolving
+    // the commit, that previously ran on the target branch (mostly "master"), can be done
```
kara: `This means that resolving the commit, that previously` -> `This means that resolving the commit that previously`

DevVersion:
@kara Fixed all grammar comments. Thanks for those!

PR #5137, opened 2017-06-14
refactor: replace disabled getters/setters with mixin

DevVersion:
* Replaces all remaining `disabled` getters and setters with the new disabled mixin
* Some disabled getters and setters have been skipped because those include extra logic in their setters/getters.

### DIFF
--- a/src/lib/button-toggle/button-toggle.ts
+++ b/src/lib/button-toggle/button-toggle.ts
@@ -25,13 +25,15 @@ import {
   AfterViewInit,
 } from '@angular/core';
 import {NG_VALUE_ACCESSOR, ControlValueAccessor} from '@angular/forms';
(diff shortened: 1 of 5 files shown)


crisbeto:
FWIW, the button toggle group classes are consolidated into one in https://github.com/angular/material2/pull/2773.

DevVersion:
@crisbeto Nice. I just added it as a note on my PR. Going to remove it now.

jelbourn:
@DevVersion so are you going to remove the changes to button-toggle?

DevVersion:
@jelbourn #2773 doesn’t really relate to the mixin here, it doesn‘t switch to to the disabled mixin (probably not present at this time)

I can remove the button-toggle change to avoid merge conflicts but it looks like it would be a easy conflict to solve.

PR #5139, opened 2017-06-14
fix(slider): emit events for home and end keys when appropriate.

### LINKED ISSUE #5123: MdSlider keyboard [HOME,END] don't emit events
Bug

When mdSlider is focused and the keyboard HOME key or END key are pressed, the component should emit change and input events.

The component value is changed, but no events are emitted.

See line 474-479 of material2/src/lib/slider/slider.ts

All other keyboard events call _increment() which calls _emitInputEvent() and _emitValueIfChanged().  But HOME and END just change the value and no events are emitted

josephperrott:
Fixes #5123

### DIFF
--- a/src/lib/slider/slider.spec.ts
+++ b/src/lib/slider/slider.spec.ts
@@ -13,7 +13,8 @@ import {
   PAGE_DOWN,
   PAGE_UP,
   RIGHT_ARROW,
-  UP_ARROW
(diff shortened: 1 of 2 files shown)


mmalerba:
please rebase

josephperrott:
@mmalerba rebased

PR #5141, opened 2017-06-15
update(input): remove align input binding

### LINKED ISSUE #5140: Remove `align` input from MdInputContainer
All this does is set `text-align` which could easily be done with user-space CSS.

DevVersion:
* The `align` input property for the `md-input-container` is not doing anything special and can be easily replaced with normal CSS from the user.

BREAKING CHANGE: The `align` input from the `<md-input-container>` component has been removed. Developers instead should use plain CSS to archive the same (using `text-align: end` for example)

Fixes #5140

### DIFF
--- a/src/demo-app/input/input-demo.html
+++ b/src/demo-app/input/input-demo.html
@@ -101,7 +101,7 @@
   <md-toolbar color="primary">Prefix + Suffix</md-toolbar>
   <md-card-content>
     <h4>Text</h4>
-    <md-input-container align="end">
+    <md-input-container class="demo-text-align-end">
       <input mdInput placeholder="amount">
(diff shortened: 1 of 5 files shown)


jelbourn:
Caretaker note: need to check for uses inside of Google before presubmit

kara:
Note: couldn't find any Google uses.

PhilipSultanescu:
I think the .demo-text-align-end class breaks the .mat-input-underline text-align, which causes the underline to align to the end instead to left, see instances below taken from input demo page:

screen shot 2017-06-19 at 11 36 06
screen shot 2017-06-19 at 11 37 02
screen shot 2017-06-19 at 11 37 21

Please check.

PR #5149, opened 2017-06-15
build: faster artifacts deploy script

DevVersion:
* Improves the artifacts deploy script by not cloning the whole history of the build repositories (all tags, commits).
* Less aggressive failing of the payload task if the previous payload is not uploaded (yet)


DevVersion:
Yeah I agree, I didn't even know about it until @ThomasBurleson showed it to me.

And yeah haha unfortunately I can't play any instrument! 😄

PR #5154, opened 2017-06-15
fix(input): invalid font declaration

### LINKED ISSUE #5262: Invalid font CSS for mat-input-container class

Bug

font CSS style for .mat-input-container should be valid 

Default typography theme set

    font: 400 inherit/1.125 Roboto,Helvetica Neue,sans-serif

which both Firefox and Chrome developer tools show as invalid (the inherit value for font-size). So all Input fields are not correctly font styled, only if a parent is (for example body).

Check https://material.angular.io/components/input/examples and go to the inspector and notice the invalid CSS

screenshot from 2017-06-20 17-12-25

To have correct default Roboto fonts styles on input field without depending on body styles like beta 6 did

Material beta 7

crisbeto:
Fixes the `font` declaration on the `md-input-container` being ignored, because it was invalid.

Note: I updated the font size to the correct value from [the spec](https://material.io/guidelines/components/text-fields.html#text-fields-states), however I'm not using it, in order to avoid any breaking changes. Also note that since the line height wasn't being used before, the baseline is now slightly off when next to a select.

Fixes #5262.

### DIFF
--- a/src/lib/core/typography/_typography.scss
+++ b/src/lib/core/typography/_typography.scss
@@ -26,7 +26,7 @@
   $caption:       mat-typography-level(12px, 20px, 400),
   $button:        mat-typography-level(14px, 14px, 500),
   // Line-height must be unit-less fraction of the font-size.
-  $input:         mat-typography-level(inherit, 1.125, 400)
+  $input:         mat-typography-level(16px, 1.125, 400)
 ) {
   @return (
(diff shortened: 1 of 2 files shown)


jelbourn:
Caretaker note: Screenshot diffs need approval

gigioSouza:
When will it be released for installing?

PR #5155, opened 2017-06-15
fix(input,select): align colors with spec

### LINKED ISSUE #5114: Input label, hint and placeholder colors don't follow material design spec
Bug

The label, placeholder, and hint text should be 54% black according to https://material.io/guidelines/components/text-fields.html#text-fields-states. 

They are currently 38% black from the 'hint-text' value of the light theme foreground palette.

https://material.angular.io/components/component/input

Follow the guidelines properly. They look too faint.

2.0.0-beta.6

crisbeto:
* Aligns the colors in the select and input [themes based on the spec](https://material.io/guidelines/components/text-fields.html#text-fields-states).
* Includes the changes from #4883 since the two PRs are closely connected.

Fixes #5114.

### DIFF
--- a/src/lib/core/style/_form-common.scss
+++ b/src/lib/core/style/_form-common.scss
@@ -1,12 +1,8 @@
-
-// Gradient for showing the dashed line when the input is disabled.
+// Renders a gradient for showing the dashed line when the input is disabled.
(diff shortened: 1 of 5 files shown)


jelbourn:
Caretaker note: screenshot diffs need approval

PR #5156, opened 2017-06-16
feat(pagination): add initial pagination component

andrewseguin:
pagination-example


src/lib/paginator/paginator.md
```diff
@@ -0,0 +1,24 @@
+`<md-paginator>` is a component that provides navigation between paged information. It supports
+ the changing of page length as well as giving the user buttons to navigate to the previous or next
+ page.
+
+<!-- example(pagination-overview) -->
+
+To use the paginator, you must provide the total length of the pagination information list and
+the length of each page. To allow the end-user to change the page length list, you may optionally
+input an array of numbers that the user may select from.
```
jelbourn: Replace this paragraph and the next:
```markdown
### Basic use
Each paginator instances requires:
* The current page index
* The number of items per page
* The total number of items being paged

When the user interacts with the paginator, a `PageEvent` will be fired that can be used to update any associated data view.

### Page size options
The paginator displays a dropdown of page sizes for the user to choose from. The options for this 
dropdown can be set via `pageSizeOptions`
```

src/lib/paginator/paginator.md
```diff
@@ -0,0 +1,24 @@
...
+
+Additionally you may provide an input of what the current page index should be. Note that
+a non-zero index will be interpretted to display the first page, whereas if the page index is beyond
+the max number of pages, then this will be reflected to the user (e.g. `140 - 149 of 100 items`).
+
+When the user changes the page length, or navigates between pages, the paginator will output a
+`PageChangeEvent` that contains the paginator's context, including the list length,
+
+### Internationalization
+In order to support internationalization or customization of the displayed text, the paginator
```
jelbourn: Can shorten to
```markdown
The labels for the paginator can be customized by providing your own instance of `MdPaginatorIntl`
```

src/lib/paginator/paginator.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (this._displayedPageLengthOptions.indexOf(this.pageLength) == -1) {
+      this._displayedPageLengthOptions.push(this.pageLength);
+    }
+
+    // Sort the numbers using a number-specific sort function.
+    this._displayedPageLengthOptions.sort((a, b) => (a - b));
+  }
+
+  /** Emits an event notifying that a change of the paginator's properties has been triggered. */
+  private _sendPageEvent() {
```
jelbourn: `_emitPageEvent`

src/lib/paginator/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  imports: [
+    CommonModule,
+    FormsModule,
+    MdButtonModule,
+    MdSelectModule,
+    MdTooltipModule,
+  ],
+  exports: [MdPaginator],
+  declarations: [MdPaginator],
+  providers: [MdPaginatorIntl],
```
DevVersion: Should this provider be treated as a singleton? Since we don't distinguish between `forRoot()` I think there would be multiple instances of this provider when lazy loading.

src/demo-app/data-table/person-data-source.ts
```diff
@@ -1,29 +1,58 @@
...
+    });
   }
 
   connect(collectionViewer: CollectionViewer): Observable<UserData[]> {
-    return collectionViewer.viewChange.map((view: {start: number, end: number}) => {
+    this.updateDisplayData();
+
+    const streams = [collectionViewer.viewChange, this._displayData];
+    return Observable.combineLatest(streams).map((results: any[]) => {
+      const view: {start: number, end: number} = results[0];
```
crisbeto: You could reduce this one down to `const [view, data] = results`.

src/lib/core/testing/jasmine-matchers.ts
```diff
@@ -10,21 +10,35 @@
...
-          result.message += ` but was ${actualRole}`;
-        }
+        return checkElementAttribute(element, 'role', expectedRole);
+      }
+    };
+  },
 
-        return result;
+  /** Matcher verifying that the element has the expected aria label. */
+  toHaveAriaLabel: () => {
```
crisbeto: Do we really need an extra matcher just for the `aria-label`? It seems like we can end up with having a matcher for every attribute.

src/lib/paginator/paginator-intl.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import {Injectable} from '@angular/core';
+
+/**
+ * Paginator labels that require internationalization. To modify the labels and text displayed,
+ * extend this class with custom values and inject it as a custom provider.
+ */
+@Injectable()
+export class MdPaginatorIntl {
```
crisbeto: Perhaps we should have an interface that the user can implement when overriding?

src/lib/paginator/paginator.html
```diff
@@ -0,0 +1,39 @@
...
+  {{_intl.getRangeLabel(currentPageIndex, pageLength, listLength)}}
+</div>
+
+<button md-icon-button
+        class="mat-paginator-navigation-button mat-paginator-navigation-previous"
+        (click)="navigateToPreviousPage()"
+        [attr.aria-label]="_intl.previousPageLabel"
+        [mdTooltip]="_intl.previousPageLabel"
+        [mdTooltipPosition]="'above'"
+        [disabled]="!canNavigateToPreviousPage(-1)">
```
crisbeto: This one should be `[disabled]="!canNavigateToPreviousPage(-1) || null"`, otherwise the `disabled` attribute will evaluate to `disabled="false"` which is still disabled. Same goes for the next button below.

src/lib/paginator/paginator.html
```diff
@@ -0,0 +1,39 @@
...
+  <div class="mat-paginator-increment"></div>
+</button>
+<button md-icon-button
```
crisbeto: This won't work in compatibility mode.

src/lib/paginator/paginator.scss
```diff
@@ -0,0 +1,65 @@
+.mat-paginator {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  min-height: 56px;
```
crisbeto: The 56px is kinda magic and seems to be repeated. Move it into a variable? Same goes for some of the other hardcoded numbers.

src/lib/paginator/paginator.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+
+      expect(component.latestPageChangeEvent).toBe(null);
+      expect(paginator.currentPageIndex).toBe(0);
+    });
+  });
+
+  it('should fail if page length and page length options are not provided', () => {
+    expect(() => {
+      TestBed.createComponent(MdPaginatorWithoutPageLengthOrOptionsApp).detectChanges();
+    }).toThrowError('No page length options available for the MdPaginator');
```
crisbeto: The error and the test will be out of sync if we decide to change the wording. Perhaps move it into a function (see `input-container-errors.ts`)?

src/lib/paginator/paginator.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+    component.listLength = 100;
+    fixture.detectChanges();
+
+    // The first item of the page should be item with index 40
+    let firstPageItemIndex = paginator.currentPageIndex * paginator.pageLength;
+    expect(firstPageItemIndex).toBe(40);
+
+    // The first item on the page is now 25. Change the page length to 25 so that we should now be
+    // on the second page where the top item is index 25.
+    paginator._changePageLength(25);
```
crisbeto: This seems like it's a private method. It might be better to test this through the `@Input`.

src/lib/paginator/paginator.ts
```diff
@@ -0,0 +1,166 @@
...
+  @Input() listLength: number = 0;
+
+  /** Number of items to display on a page. By default set to the first of page length option. */
+  @Input()
+  set pageLength(pageLength: number) {
+    this._pageLength = pageLength;
+    this._updateDisplayedPageLengthOptions();
+  }
+  get pageLength(): number { return this._pageLength; }
+  _pageLength: number;
```
crisbeto: Shouldn't this be private? Same goes for all the other underscored properties.

src/lib/paginator/paginator.ts
```diff
@@ -0,0 +1,166 @@
...
+  private _updateDisplayedPageLengthOptions() {
+    if (!this._initialized) { return; }
+
+    this._displayedPageLengthOptions = this.pageLengthOptions.slice();
+    if (this._displayedPageLengthOptions.indexOf(this.pageLength) == -1) {
+      this._displayedPageLengthOptions.push(this.pageLength);
+    }
+
+    // Sort the numbers using a number-specific sort function.
+    this._displayedPageLengthOptions.sort((a, b) => (a - b));
```
crisbeto: Unnecessary parentheses here: `this._displayedPageLengthOptions.sort((a, b) => a - b);`

src/lib/paginator/_paginator-theme.scss
```diff
@@ -0,0 +1,40 @@
...
+    color: mat-color($foreground, secondary-text);
+  }
+
+  .mat-paginator-increment,
+  .mat-paginator-decrement {
+    border-top: 2px solid mat-color($foreground, 'icon');
+    border-right: 2px solid mat-color($foreground, 'icon');
+  }
+
+  .mat-icon-button[disabled] .mat-paginator-increment,
```
crisbeto: You can nest this to avoid repetition:

```
.mat-icon-button[disabled] {
  .mat-paginator-increment,
  .mat-paginator-decrement {
    border-color: mat-color($foreground, 'disabled');
  }  
}
```

src/lib/paginator/paginator.html
```diff
@@ -0,0 +1,39 @@
...
+        class="mat-paginator-navigation-button mat-paginator-navigation-next"
+        (click)="navigateToNextPage()"
+        [attr.aria-label]="_intl.nextPageLabel"
```
crisbeto: Just `[aria-label]` should work.

src/lib/paginator/paginator.scss
```diff
@@ -0,0 +1,65 @@
...
+}
+.mat-paginator-increment,
+[dir='rtl'] .mat-paginator-decrement {
+  transform: rotate(225deg);
+}
+
+.mat-paginator-decrement {
+  margin-left: 12px;
+  [dir='rtl'] & {
+    margin-right: 12px;
```
crisbeto: Shouldn't the `margin-left` be reset in RTL?

src/lib/paginator/paginator.html
```diff
@@ -0,0 +1,39 @@
...
+
+  <div *ngIf="_displayedPageLengthOptions.length <= 1">{{pageLength}}</div>
+</div>
+
+<div class="mat-paginator-range-label">
+  {{_intl.getRangeLabel(currentPageIndex, pageLength, listLength)}}
+</div>
+
+<button md-icon-button
+        class="mat-paginator-navigation-button mat-paginator-navigation-previous"
```
andrewseguin: Right, good catch thanks. This was there before adding explicit previous/next classes and isn't needed now.

src/lib/paginator/paginator.html
andrewseguin: Not sure I follow why `|| null` is necessary. 

Won't `disabled="false"` be evaluated as false since it is passed through `coerceBooleanProperty` which will check if it is the string `false`?

src/lib/paginator/paginator.html
andrewseguin: We added the following to the `@Component`:

```js
  providers: [
    {provide: MATERIAL_COMPATIBILITY_MODE, useValue: false}
  ],
```

Which should make sure that the paginator's template will *not* be in compatibility mode.

src/lib/paginator/paginator.html
andrewseguin: Funny thing here, since that works for `md-select` but not `md-button` since `md-select` explicitly marks this as an `@Input` and handles applying the attribute. For the `md-button` however, Angular complains that it is not an input. In the case of aria attributes, they are not properties of the DOM element object and so cannot be applied with just `[aria-label]`

src/lib/paginator/paginator.scss
andrewseguin: Yeah absolutely, good catch

src/lib/paginator/paginator.spec.ts
andrewseguin: Good point - changed logic however to remove error. Defaulting so that the pageLength and options matches the list length now, which defaults to 0

src/lib/paginator/paginator.spec.ts
andrewseguin: Unfortunately this is only used through the handler of the `md-select` which I don't seem to be able to grab in the tests. I tried using a `ViewChild` without luck, and I don't want to expose it through the paginator.

src/lib/paginator/paginator.ts
```diff
@@ -0,0 +1,166 @@
...
+          this.listLength;
+    }
+
+    this._updateDisplayedPageLengthOptions();
+  }
+
+  /** Increments the current page index to the next page index if a next page exists. */
+  navigateToNextPage() {
+    if (!this.canNavigateToNextPage()) { return; }
+    this.currentPageIndex++;
```
andrewseguin: Right now we only do fire on interaction (user selects a different page size or navigated between pages). We could consider defaulting pageLength to 0 and removing this check to default to a certain page size. I believe that would make this more stateless.

src/demo-app/data-table/person-data-source.ts
andrewseguin: The downside is that I lose the typing of view, unless there's a syntax I'm missing that would let me apply different typing to view and data

src/lib/paginator/paginator.html
crisbeto: Oh, I didn't notice it was an md-button. Then it should work.

src/demo-app/data-table/person-data-source.ts
crisbeto: I thought that TS might be smart enough to infer the type. It's not a big deal to leave it as it is.

src/lib/core/testing/jasmine-matchers.ts
andrewseguin: Yeah not sure this ended up making things easier, and it doesn't scale. Tried implementing a generic "hasAttributes" matcher but still doesn't seem too useful. Taking it out

src/demo-app/data-table/person-data-source.ts
andrewseguin: Cool, got it working by changing the results from `any[]` to a tuple `(results: [{start: number, end: number}, UserData[]])`. It is smart enough to do `const [view, data] = results`

src/lib/paginator/index.ts
andrewseguin: That's correct, it'll have multiple instances though that may be desired for the user if they want to customize their paginator not on internationalization but on the type of data. E.g. instead of "Items per page:" they can put "Users per page:".

@jelbourn what do you think

DevVersion: That's a good argument. Even though it kinda sounds confusing when the provider is called "Intl"

src/lib/paginator/paginator-intl.ts
andrewseguin: Perhaps, if we do then we should also look at changing `MdDatepickerIntl`. I followed along with what we did in that class

src/lib/paginator/paginator.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  /** The zero-based page index of the displayed list of items. Defaulted to 0. */
+  @Input() pageIndex: number = 0;
+
+  /** The length of the total number of items that are being paginated. Defaulted to 0. */
+  @Input() length: number = 0;
+
+  /** Number of items to display on a page. By default set to the first of page size option. */
+  @Input()
+  get pageSize(): number { return this._pageSize; }
```
jelbourn: Is there a default for these?

andrewseguin:
Made the code changes, ready for review. The page size is currently set to default to 0. Let me know if you want that to be removed or increased to something like 10 or 25. 

Note that the SCSS contains a nested class rule for .mat-select-trigger which I don't think I'll be able to get around.

src/lib/paginator/paginator.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  /** Returns true if the user can go to the next page. */
+  hasPreviousPage() {
+    return (this.pageIndex - 1) >= 0 && this.pageSize != 0;
+  }
+
+  /** Returns true if the user can go to the next page. */
+  hasNextPage() {
+    const numberOfPages = Math.ceil(this.length / this.pageSize) - 1;
+    return (this.pageIndex + 1) <= numberOfPages && this.pageSize != 0;
```
crisbeto: Hmm, shouldn't `this.pageIndex < numberOfPages` instead of `(this.pageIndex + 1) <= numberOfPages` technically be the same?

src/lib/paginator/paginator.ts
```diff
@@ -0,0 +1,157 @@
...
+  /** Decrements the page index to the previous page index if a next page exists. */
+  previousPage() {
+    if (!this.hasPreviousPage()) { return; }
+    this.pageIndex--;
+    this._emitPageEvent();
+  }
+
+  /** Returns true if the user can go to the next page. */
+  hasPreviousPage() {
+    return (this.pageIndex - 1) >= 0 && this.pageSize != 0;
```
crisbeto: The `(this.pageIndex - 1) >= 0` could we rewritten to `this.pageIndex >= 1`.

src/material-examples/paginator-overview/paginator-overview-example.html
```diff
@@ -0,0 +1,4 @@
+<md-paginator [length]="100"
+              [pageSize]="10"
+              [pageSizeOptions]="[5, 10, 25, 100]" ]>
+</md-paginator>
```
crisbeto: This file is missing a newline at the end of the file.

src/lib/paginator/paginator-intl.ts
```diff
@@ -0,0 +1,41 @@
...
+import {Injectable} from '@angular/core';
+
+/**
+ * To modify the labels and text displayed, create a new instance of MdPaginatorIntl and
+ * include it in a custom provider
+ */
+@Injectable()
+export class MdPaginatorIntl {
+  /** A label for the page size selector. */
+  itemsPerPageLabel = 'Items per page:';
```
ErinCoughlan: Would having these as inputs to the paginator with default values make more sense?

For my understanding of this class as well, if I only wanted to change the "Items per page" text, I can only override that one field?

src/lib/paginator/paginator.scss
```diff
@@ -0,0 +1,81 @@
...
+    margin-right: $mat-paginator-button-decrement-icon-margin;
+  }
+}
+
+.mat-paginator-increment {
+  margin-left: $mat-paginator-button-increment-icon-margin;
+  [dir='rtl'] & {
+    margin-right: $mat-paginator-button-increment-icon-margin;
+  }
+}
```
ErinCoughlan: Missing newline at end of file

src/lib/paginator/paginator.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+      component.pageIndex = 2;
+      fixture.detectChanges();
+      expect(rangeElement.innerText).toBe('11 - 12 of 12');
+
+      // View third page of list of 10, each page contains 5 items.
+      component.length = 10;
+      component.pageSize = 5;
+      component.pageIndex = 2;
+      fixture.detectChanges();
+      expect(rangeElement.innerText).toBe('11 - 15 of 10');
```
ErinCoughlan: This seems wrong. Why would I only be looking at values greater than how many items are available? Same with the next expect as well.

src/lib/paginator/paginator.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+  beforeEach(async(() => {
+    jasmine.addMatchers(customMatchers);
+
+    TestBed.configureTestingModule({
+      imports: [
+        MdButtonModule,
+        MdCommonModule,
+        MdSelectModule,
+        MdTooltipModule,
+        MdPaginatorModule,
```
ErinCoughlan: Just include the MdPaginatorModule? The rest should be pulled in just from using it.

This is similar to what we did with table -- only include the single module so you're testing the way a user would use the component. This lets you identify if a module is missing.

src/material-examples/paginator-configurable/paginator-configurable-example.html
```diff
@@ -0,0 +1,28 @@
...
+              [pageSizeOptions]="pageSizeOptions"
+              (page)="pageEvent = $event">
+</md-paginator>
+
+<div *ngIf="pageEvent">
+  <h2>Page Change Event Properties</h2>
+  <div>List length: {{pageEvent.length}}</div>
+  <div>Page size: {{pageEvent.pageSize}}</div>
+  <div>Page index: {{pageEvent.pageIndex}}</div>
+</div>
```
ErinCoughlan: Missing newline.

andrewseguin:
Thanks @crisbeto, those conditionals look better

src/demo-app/data-table/person-data-source.ts
```diff
@@ -1,29 +1,58 @@
...
+          this._renderedData.length = data.length;
 
-      return this._renderedData;
-    });
+          const buffer = 20;
+          let rangeStart = Math.max(0, view.start - buffer);
+          let rangeEnd = Math.min(data.length, view.end + buffer);
+
+          for (let i = rangeStart; i < rangeEnd; i++) {
+            this._renderedData[i] = data[i];
```
andrewseguin: This is because I'll want renderedData to be a virtual buffer that fills in over time. E.g. the table says it wants rows 0-10 and then the user scrolls to the end and asks for 90-100. Rendered data should be an array that contains 100 items that are all undefined except for 0-10 and 90-100

src/lib/paginator/paginator-intl.ts
andrewseguin: When looking at this from an internationalization point of view, I think the injected class makes more sense (e.g. when a RTL language should be displayed). Though from a customization point of view, an `@Input` makes sense.

@jelbourn Paul and Erin both brought up some points about this class, any thoughts on renaming or changing?

src/lib/paginator/paginator.spec.ts
andrewseguin: Gah, made the same mistake twice. Thanks for pointing this out, much better to mimic what the users will do

src/lib/paginator/paginator.spec.ts
andrewseguin: Definitely is wrong =D but its wrong from the developer since they are telling us to go to a page that should not exist. Other alternative is to throw an error, but we opted to go the route of displaying what the developer told us to display (that is, items 11-15).

src/lib/paginator/paginator-intl.ts
ErinCoughlan: From my POV, the most common thing I will want to do is set the i18n text so that "next page" becomes "Página siguiente" or whatever language the user is in.

I will never be able to use this default class, but for someone who just wants to render a table, it would be useful, although not sure if more useful than a default input. For RTL, we rely primarily on CSS, which you have covered.

jelbourn: We use injection for i18n because it's something that's defined once and doesn't need to be change-detected / have generated code.

jelbourn:
Caretaker note: will need build rules

tosehee75:
Is this tool available any time soon?

GustavoCostaW:
Hey @andrewseguin and @jelbourn, the material team have an interest in pagination like this?



you can check my code in

https://github.com/GustavoCostaW/ngc-pagination

PR #5157, opened 2017-06-16
chore: fixes lint issue

andrewseguin:
Fixes an issue in lint:

`ERROR: src/lib/core/common-behaviors/color.ts[28, 68]: expected nospace before colon in call-signature`

### DIFF
--- a/src/lib/core/common-behaviors/color.ts
+++ b/src/lib/core/common-behaviors/color.ts
@@ -24,8 +24,8 @@ export interface HasRenderer {
 export type ThemePalette = 'primary' | 'accent' | 'warn' | null;
 
 /** Mixin to augment a directive with a `color` property. */
-export function mixinColor<T extends Constructor<HasRenderer>>(base: T, defaultColor?: ThemePalette)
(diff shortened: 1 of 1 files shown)


src/lib/core/common-behaviors/color.ts
```diff
@@ -24,8 +24,8 @@ export interface HasRenderer {
 export type ThemePalette = 'primary' | 'accent' | 'warn' | null;
 
 /** Mixin to augment a directive with a `color` property. */
-export function mixinColor<T extends Constructor<HasRenderer>>(base: T, defaultColor?: ThemePalette)
-    : Constructor<CanColor> & T {
+export function mixinColor<T extends
+    Constructor<HasRenderer>>(base: T, defaultColor?: ThemePalette): Constructor<CanColor> & T {
```
jelbourn: It would be better to break after the `(`

As @DevVersion says, though, this is passing on CI. Do you have a stale version of tslint installed?

andrewseguin: My tslint is on version 5.2.0 - is that stale?

PR #5144, opened 2017-06-15
fix(tooltip): remove native event listener on component destroy

michastreppel:
Component should clean up events it has registered to.

Closes #4499

PR #5164, opened 2017-06-16
fix(memory leak): memory leaking in UniqueSelectionDispatcher

Eddman:
Listeners are only registered inside of `UniqueSelectionDispatcher`. But there were never removed causing in my case a small memory leak when using `MdButtonToggle`.

In this fix `UniqueSelectionDispatcher.listen` returns a function that removes the listener once called.

### DIFF
--- a/src/lib/button-toggle/button-toggle.ts
+++ b/src/lib/button-toggle/button-toggle.ts
@@ -376,11 +376,12 @@ export class MdButtonToggle implements OnInit, OnDestroy {
     this.buttonToggleGroupMultiple = toggleGroupMultiple;
 
     if (this.buttonToggleGroup) {
-      this._buttonToggleDispatcherListener = _buttonToggleDispatcher.listen((id: string, name: string) => {
-        if (id != this.id && name == this.name) {
-          this.checked = false;
-        }
-      });
+      this._buttonToggleDispatcherListener = _buttonToggleDispatcher.listen(
+        (id: string, name: string) => {
+          if (id != this.id && name == this.name) {
+            this.checked = false;
+          }
+        });
 
       this._type = 'radio';
       this.name = this.buttonToggleGroup.name;
--- a/src/lib/button-toggle/button-toggle.ts
+++ b/src/lib/button-toggle/button-toggle.ts
@@ -16,6 +16,7 @@ import {
   HostBinding,
   Input,
   OnInit,
+  OnDestroy,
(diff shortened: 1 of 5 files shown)


Eddman:
Please merge ASAP: it is causing huge mem-leaks in my application so it makes the Chrome tab to crash in few minutes...

In combination with #5144 I'm almost unable to use the library. We are managing a big data set that is leaking because of these...

src/lib/button-toggle/button-toggle.ts
```diff
@@ -390,11 +394,12 @@ export class MdButtonToggle implements OnInit {
     this.buttonToggleGroupMultiple = toggleGroupMultiple;
 
     if (this.buttonToggleGroup) {
-      _buttonToggleDispatcher.listen((id: string, name: string) => {
-        if (id != this.id && name == this.name) {
-          this.checked = false;
-        }
-      });
+      this._buttonToggleDispatcherListener = _buttonToggleDispatcher.listen(
```
jelbourn: Prefer line-breaking at the higher syntactic level, e.g.
```ts
this._removeUniqueSelectionListener =
    _buttonToggleDispatcher.listen((id: string, name: string) => {
      if (id != this.id && name == this.name) {
        this.checked = false;
      }
    });
```

src/lib/core/coordination/unique-selection-dispatcher.ts
```diff
@@ -36,9 +36,17 @@ export class UniqueSelectionDispatcher {
...
+  /**
+   * Listen for future changes to item selection.
+   * @return Function used to unregister listener
+   **/
+  listen(listener: UniqueSelectionDispatcherListener): Function {
     this._listeners.push(listener);
+    return () => {
+      this._listeners = this._listeners.filter((registered: UniqueSelectionDispatcherListener) => {
+        return listener !== registered;
+      });
```
jelbourn: This also needs a unit test. It should set up a listener, then deregister and assert that listener isn't called again on notify

Eddman:
@jelbourn All comments are implemented. 

However Travis failed to start e2e test due to timeout...

src/lib/core/coordination/unique-selection-dispatcher.ts
Eddman: Done. See unique-selection-dispatcher.spec.ts.

PR #5168, opened 2017-06-16
test(menu): fix test failures against Angular 4.2

crisbeto:
Fixes a couple of menu unit tests that were failing against Angular 4.2.2.


crisbeto:
This seems to be the only place where it was happening, though. Perhaps we overlooked it?

jelbourn:
I'm starting to think it's a bug because we import `NoopAnimationModule` and Angular's delegation to the animation driver for removal is not removing the element. Stepping through the animation code now.

jelbourn:
I suspect 
https://github.com/angular/angular/blob/6ca46929fa26468f179196979b5f77a48f79991f/packages/animations/browser/src/render/transition_animation_engine.ts#L299-L299

alexw10:
Any update on this / on the path forward?

PR #5172, opened 2017-06-16
fix(autocomplete): reopening when clicking an option in IE

### LINKED ISSUE #5165: Autocomplete panel doesn't close when an option click
Bug

After I click on an option (using mouse cursor) the autocomplete suggestion panel should be closed.

When I click on an option (using mouse cursor) the autocomplete suggestion panel closing and immediately opening again. This bug occurs only in Internet Explorer (I tested on version 11).

Plunker:
https://plnkr.co/edit/7zAjuhQJdibmPCaLsulm

Use Internet Explorer 11.
1. Select option in 'State1' using the mouse cursor. 
2. Input 'State2' is blocked by 'State1' 

Also, this bug can be reproduced on official documentation
https://material.angular.io/components/component/autocomplete

I have two autocomplete input. when user select option in one input, the application should do focus on the other input programmatically. With this bug, when user select option in first input, application do focus on the other input, but the suggestion panel of first input stays opened. In result, a user sees two suggestion panels for each input.

Angular 4.2.2
Material 2.0.0-beta.6
OS Windows 10 (build 15063.0)
TypeScript 2.3.4
Internet Explorer 11

In Chrome, Edge and Firefox this bug does not reproduce.

crisbeto:
Fixes an issue that caused IE to reopen the autocomplete panel if the user clicks to select an item. It seems that IE can delay refocusing the input after the panel has been closed, which causes the focus handler to reopen it.

Fixes #5165.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -63,13 +63,13 @@ export const MD_AUTOCOMPLETE_VALUE_ACCESSOR: any = {
  */
 export function getMdAutocompleteMissingPanelError(): Error {
   return Error('Attempting to open an undefined instance of `md-autocomplete`. ' +
-                   'Make sure that the id passed to the `mdAutocomplete` is correct and that ' +
-                   'you\'re attempting to open it after the ngAfterContentInit hook.');
+               'Make sure that the id passed to the `mdAutocomplete` is correct and that ' +
+               'you\'re attempting to open it after the ngAfterContentInit hook.');
 }
 
 @Directive({
   selector: 'input[mdAutocomplete], input[matAutocomplete],' +
-  'textarea[mdAutocomplete], textarea[matAutocomplete]',
+            'textarea[mdAutocomplete], textarea[matAutocomplete]',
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -78,7 +78,9 @@ export function getMdAutocompleteMissingPanelError(): Error {
     '[attr.aria-activedescendant]': 'activeOption?.id',
     '[attr.aria-expanded]': 'panelOpen.toString()',
     '[attr.aria-owns]': 'autocomplete?.id',
-    '(focus)': 'openPanel()',
+    // Note: we use `focusin`, as opposed to `focus`, in order to open the panel
+    // a little earlier. This avoids issues where IE delays the focusing the input.
```
kara: Typo: `IE delays the focusing the input` -> `IE delays the focusing of the input`

jelbourn:
Caretaker note: one team has a test utility that fires a fake focus event and tests that the autocomplete is open; changing it to _also_ do `focusin` fixes the test for this PR

PR #5175, opened 2017-06-16
feat(sticky-header): Initial version of sticky header

sllethe:
add support to 'optional 'cdkStickyRegion' input '


src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+    // the height of 'stickyParent'
+    private containerHeight: number;
+
+    // the height of 'elem'
+    private elemHeight: number;
+
+    private containerStart: number;
+    private scrollFinish: number;
+
+    private scrollingWidth: any;
```
tinayuangao: Specify type?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    // the padding of 'elem'
+    private elementPadding: any;
+    private paddingNumber: any;
+
+    // sticky element's width
+    private width: string = 'auto';
+
+    constructor(private element: ElementRef,
+                public findScroll: Scrollable,
```
tinayuangao: `scrollable`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
+import {
```
tinayuangao: nit: 2 space for indent

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+
+        this.deactivated.next(this.elem);
+    }
+
+
+    sticker(): void {
+        // detecting when a container's height changes
+        let currentContainerHeight: number = this.getCssNumber(this.stickyParent, 'height');
+        if (currentContainerHeight !== this.containerHeight) {
+            this.defineRestrictions();
```
tinayuangao: Usually `defineRestrictions()` are called before `sticker()`. Do you still need this?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+        this.scrollingWidth = this.upperScrollableContainer.scrollWidth;
+
+        this.attach();
+
+        if (this.width == 'auto') {
+            this.width = this.originalCss.width;
+        }
+
+        this.defineRestrictions();
+        this.sticker();
```
tinayuangao: Combine `defineRestrictions` and `sticker`? 

`defineRestrictionsAndStick()`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    attach() {
+        this.upperScrollableContainer.addEventListener('scroll', this.onScrollBind, false);
+        this.upperScrollableContainer.addEventListener('resize', this.onResizeBind, false);
+
+        // Have to add a 'onTouchMove' listener to make sticky header work on mobile phones
+        this.upperScrollableContainer.addEventListener('touchmove', this.onTouchMoveBind, false);
+
+        Observable.fromEvent(this.upperScrollableContainer, 'scroll')
+            .subscribe(() => this.onScroll());
```
tinayuangao: Can be 

```
Observable.fromEvent(this.upperScrollableContainer, 'scroll')
   .subscribe(() => this.defineRestrictionsAndStick());
```

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+import {Scrollable} from '../core/overlay/scroll/scrollable';
+import {Subject} from 'rxjs/Subject';
+
+
+@Directive({
+    selector: '[md-sticky-viewport], [cdkStickyViewport], [cdkStickyRegion]',
+})
+
+
+@Injectable()
```
sllethe: Right~ We don't need @injectable for StickyParentDirective.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    // The parent element, which with the 'md-sticky-viewport' tag
+    private pelem: any;
+
+    constructor(private element: ElementRef) {
+        this.pelem = element.nativeElement;
+    }
+
+    ngOnInit(): void {
+        this.pelem.classList.add('sticky-parent');
```
sllethe: Removed OnInit, AfterViewInit and onDestory class. Because we don't need them.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+
+@Directive({
+    selector: '[md-sticky], [cdkSticky], [cdkStickyHeader]',
+})
+
+
+@Injectable()
+export class StickyHeaderDirective implements OnInit, OnDestroy, AfterViewInit {
+
+    @Input('sticky-zIndex') zIndex: number = 10;
```
sllethe: Added comments for zIndex:
/**Set the sticky-header's z-index as 10 in default. Make it as an input
     * variable to make user be able to customize the zIndex when
     * the sticky-header's zIndex is not the largest in current page.
     * Because if the sticky-header's zIndex is not the largest in current page,
     * it may be sheltered by other element when being sticked.
*/

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,334 @@
...
+    @Input('parentRegion') parentRegion: any;
+    @Input('scrollRegion') scrollableRegion: any;
+
+
+    private activated = new EventEmitter();
```
sllethe: changed 'activated' and 'deactivated' to '_activated' and '_deactivated'. They are used in stickElement() and unstickElement() to emit which element is sticked or unsticked

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Combined defineRestrictions() and sticker() as defineRestrictionsAndStick()

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Deleted checking currentContainerHeight in sticker(). Because we always call defineRestrictions() before sticker().

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,314 @@
...
+export class StickyHeaderDirective implements OnDestroy, AfterViewInit {
+
+    /**Set the sticky-header's z-index as 10 in default. Make it as an input
+     * variable to make user be able to customize the zIndex when
+     * the sticky-header's zIndex is not the largest in current page.
+     * Because if the sticky-header's zIndex is not the largest in current page,
+     * it may be sheltered by other element when being sticked.
+     */
+    @Input('sticky-zIndex') zIndex: number = 10;
+    @Input() cdkStickyParentRegion: any;
```
sllethe: ' @Input('cdkStickyHeaderParentRegion') parentRegion: HTMLElement '
The parentRegion is a 'HTMLElement'

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,314 @@
...
+         * So the 'position: fixed' does not work on iPhone and iPad. To make it work,
+         * I need to use translate3d(0,0,0) to force Safari rerendering the sticky element.
+         **/
+        this.elem.style.transform = 'translate3d(0,0,0)';
+
+        this.elem.style.zIndex = this.zIndex;
+        this.elem.style.position = 'fixed';
+        this.elem.style.top = this.getCssNumber(this.upperScrollableContainer, 'top') + 'px';
+
+        this._scrollingRight = this.upperScrollableContainer.offsetLeft +
```
sllethe: deleted _scrollingRight

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+
+    /**
+     * define the restrictions of the sticky header(including stickyWidth,
+     * when to start, when to finish)
+     */
+    defineRestrictions(): void {
+        let containerTop: any = this.stickyParent.getBoundingClientRect();
+        this.elemHeight = this.elem.offsetHeight;
```
sllethe: set elemHeight as local variable

sllethe:
Comments addressed. Please take another look~ :+1:

src/demo-app/demo-app-module.ts
```diff
@@ -74,9 +74,11 @@ import {
   MdToolbarModule,
   MdTooltipModule,
   OverlayContainer,
-  StyleModule
+  StyleModule,
+  MdStickyHeaderModule,
```
sllethe: removed demo code

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+    // unstuck when the element is scrolled out of the sticky region
+    if (this.isStuck &&
+      (currentPosition < this._containerStart || currentPosition > this._scrollFinish) ||
+      currentPosition >= this._scrollFinish) {
+      this.resetElement();
+      if (currentPosition >= this._scrollFinish) {
+        this.unstuckElement();
+      }
+      this.isStuck = false;    // stick when the element is within the sticky region
+    }else if ( this.isStuck === false &&
```
tinayuangao: nit: add space before `else`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+    this.defineRestrictions();
+    this.sticker();
+  }
+
+
+  private getCssValue(element: any, property: string): any {
+    let result: any = '';
+    if (typeof window.getComputedStyle !== 'undefined') {
+      result = window.getComputedStyle(element, '').getPropertyValue(property);
+    }else if (typeof element.currentStyle !== 'undefined')  {
```
tinayuangao: same here

.firebaserc
```diff
@@ -0,0 +1 @@
+
```
tinayuangao: revert?

src/demo-app/demo-app/demo-app.ts
```diff
@@ -67,7 +67,7 @@ export class DemoApp {
     {name: 'Tooltip', route: 'tooltip'},
     {name: 'Platform', route: 'platform'},
     {name: 'Style', route: 'style'},
-    {name: 'Typography', route: 'typography'}
+    {name: 'Typography', route: 'typography'},
```
tinayuangao: revert this file?

src/demo-app/demo-app/routes.ts
```diff
@@ -37,6 +37,7 @@ import {DataTableDemo} from '../data-table/data-table-demo';
 import {TypographyDemo} from '../typography/typography-demo';
 import {ExpansionDemo} from '../expansion/expansion-demo';
 
+
```
tinayuangao: revert this file

sllethe:
Comments addressed. Please take another look~ :+1:

src/lib/sticky-header/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import {CdkStickyRegion, CdkStickyHeader} from './sticky-header-dir';
+
+
+
+@NgModule({
+  imports: [OverlayModule, MdCommonModule, CommonModule],
+  declarations: [CdkStickyRegion, CdkStickyHeader],
+  exports: [CdkStickyRegion, CdkStickyHeader, MdCommonModule],
+})
+export class CdkStickyHeaderModule {}
```
jelbourn: Just `StickyHeaderModule`; we aren't prefixing cdk classes

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+import {Scrollable} from '../core/overlay/scroll/scrollable';
+
+
+/**
+ * 'cdkStickyRegion' is a region contains one sticky header and other
+ * contents that user wants to put under the sticky-header. There only can
+ * be one sticky-header in one 'cdkStickyRegion'.
+ * If a user does not define a 'cdkStickyRegion' for a sticky-header, the direct
+ * parent node of the sticky-header will be set as the 'cdkStickyRegion'.
+ */
```
jelbourn:
```ts
/**
 * Directive that marks an element as a "sticky region", meant to contain exactly one sticky-header
 * along with the content associated with that header. The sticky-header inside of the region will
 * "stick" to the top of the scrolling container as long as this region is within the scrolling
 * viewport.
 *
 * If a user does not explicitly define a sticky-region for a sticky-header, the direct
 * parent node of the sticky-header will be used as the sticky-region.
 */
```

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+@Directive({
+  selector: '[cdkStickyRegion]',
+})
+export class CdkStickyRegion {
+  constructor(private _element: ElementRef) { }
+
+  getElementRef(): ElementRef {
```
jelbourn: Rather than having a method to get this, you can change to constructor to
```ts
constructor(public readonly _elementRef: ElementRef) { }
```
And then reference it directly from sticky-header (we use the leading `_` to denote "internal to Angular Material")

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+  }
+}
+
+
+/**
+ * The 'cdkStickyHeader' is the header which user wants to be stuck at top of the
+ * scrollable container. The main logic is in the 'sticker()' function.
+ */
+const STICK_START_CLASS = 'stick-start';
+const STICK_END_CLASS = 'sticky-end';
```
jelbourn: Add css classes must be prefixed with `mat-`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+  getElementRef(): ElementRef {
+    return this._element;
+  }
+}
+
+
+/**
+ * The 'cdkStickyHeader' is the header which user wants to be stuck at top of the
+ * scrollable container. The main logic is in the 'sticker()' function.
+ */
```
jelbourn: This comment should be applied to the class (right now it's applied to `STICK_START_CLASS`).

Comment should be something like
```ts
/**
 * Directive that marks an element as a sticky-header. Inside of a scrolling container (marked with
 * cdkScrollable), this header will "stick" to the top of the scrolling viewport while its sticky
 * region (see cdkStickyRegion) is in view.
 */
```

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+   * it may be sheltered by other element when being stick.
+   */
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+  @Input('cdkStickyParentRegion') parentRegion: HTMLElement;
+  @Input('cdkStickyScrollableRegion') scrollableRegion: HTMLElement;
+
+  private _onScrollBind: EventListener = this.onScroll.bind(this);
+  private _onResizeBind: EventListener = this.onResize.bind(this);
+  private _onTouchMoveBind: EventListener = this.onTouchMove.bind(this);
+  public isStuck: boolean = false;
```
jelbourn: Omit `public` here and elsewhere (it is the default)

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  ngOnDestroy(): void {
+    this.upperScrollableContainer.removeEventListener('scroll', this._onScrollBind);
+    this.upperScrollableContainer.removeEventListener('resize', this._onResizeBind);
+    this.upperScrollableContainer.removeEventListener('touchmove', this._onTouchMoveBind);
+  }
+
+  attach() {
+    this.upperScrollableContainer.addEventListener('scroll', this._onScrollBind, false);
+    this.upperScrollableContainer.addEventListener('resize', this._onResizeBind, false);
```
jelbourn: Why does this create its own event listeners instead of using the parent `cdkScrollable`?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+    let stickyCss:any = {
+      zIndex: this.zIndex,
+      position: 'fixed',
+      top: this.upperScrollableContainer.offsetTop + 'px',
+      right: stuckRight + 'px',
+      left: this.upperScrollableContainer.offsetLeft + 'px',
+      bottom: 'auto',
+      width: this._scrollingWidth + 'px',
+    };
+    Object.assign(this.element.style, stickyCss);
```
jelbourn: `Object.assign` isn't supported in IE11; use `extendObject` from `src/lib/core/util`.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+      this.stickElement();
+    }
+  }
+
+  /**
+   * define the restrictions of the sticky header(including stickyWidth,
+   * when to start, when to finish)
+   */
+  defineRestrictions(): void {
+    let containerTop: any = this.stickyParent.getBoundingClientRect();
```
jelbourn: This `getBoundingClientRect` is going to run on every `scroll` and `resize` event, which will cause slowdowns.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+  selector: '[cdkStickyHeader]',
+})
+export class CdkStickyHeader implements OnDestroy, AfterViewInit {
+
+  /**
+   * Set the sticky-header's z-index as 10 in default. Make it as an input
+   * variable to make user be able to customize the zIndex when
+   * the sticky-header's zIndex is not the largest in current page.
+   * Because if the sticky-header's zIndex is not the largest in current page,
+   * it may be sheltered by other element when being stick.
```
andrewseguin: (nit) `stuck` rather than `stick`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+  @Input('cdkStickyParentRegion') parentRegion: HTMLElement;
+  @Input('cdkStickyScrollableRegion') scrollableRegion: HTMLElement;
+
+  private _onScrollBind: EventListener = this.onScroll.bind(this);
+  private _onResizeBind: EventListener = this.onResize.bind(this);
+  private _onTouchMoveBind: EventListener = this.onTouchMove.bind(this);
+  public isStuck: boolean = false;
+
+  // the element with the 'cdkStickyHeader' tag
```
andrewseguin: Use `/** ... */` for documentation

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+  public element: HTMLElement;
+
+  // the upper container element with the 'cdkStickyRegion' tag
+  public stickyParent: HTMLElement;
+  public upperScrollableContainer: HTMLElement;
```
andrewseguin: Add docs for the remaining variables

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+   * when it is being unstuck
+   */
+  public originalCss: any;
+
+  private _containerStart: number;
+  private _scrollFinish: number;
+
+  private _scrollingWidth: number;
+
+  constructor(private _element: ElementRef,
```
andrewseguin: Remove `private`, `public` from the variables if they are only used in the constructor

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+    this.element = _element.nativeElement;
+    this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+    this.scrollableRegion = scrollable.getElementRef().nativeElement;
+    if (parentReg != null) {
+      this.parentRegion = parentReg.getElementRef().nativeElement;
+    }
+  }
+
+  ngAfterViewInit(): void {
+
```
andrewseguin: Remove extra line

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+    if (this.parentRegion != null) {
```
andrewseguin: `this.stickyParent = this.parentRegion != null ? this.parentRegion : this.element.parentElement`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+     * of the viewport as they double tap or pinch to zoom in or out, or by touching
+     * and dragging to pan the page. As a user changes the zoom level and position of the
+     * viewport they are doing so within a viewable content area of fixed size
+     * (that is, the window). This means that webpage elements that have their position
+     * "fixed" to the viewport can end up outside the viewable content area, offscreen.
+     *
+     * So the 'position: fixed' does not work on iPhone and iPad. To make it work,
+     * 'translate3d(0,0,0)' needs to be used to force Safari re-rendering the sticky element.
+     **/
+    this.element.style.transform = 'translate3d(0,0,0)';
```
andrewseguin: I believe IE11 will have trouble with this, change to `translate3d(0px, 0px, 0px);

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,295 @@
...
+   * will be scrolled up with its cdkStickyRegion element. In this way, the sticky header
+   * can be changed smoothly when two sticky header meet and the later one need to replace
+   * the former one.
+   */
+  unstuckElement(): void {
+    this.isStuck = false;
+
+    this.element.classList.add(STICK_END_CLASS);
+    this.stickyParent.style.position = 'relative';
+    let unstuckCss: any = {
```
andrewseguin: Seeing this many times, wondering if we should make some interface for this set of properties to make sure we gain some type checking when we continue to maintain it

src/lib/sticky-header/index.ts
sllethe: :+1: Changed 'CdkStickyHeadeModule' to 'StickyHeaderModule'

src/lib/sticky-header/sticky-header-dir.ts
sllethe: :+1: Great! I didn't know it can be implemented like this~ It's cool, haha~

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Should I change it into 'mat_STICK_START_CLASS' or 'MAT_STICK_START_CLASS' or 'mat_stick_start_class'?
I'm currently change it into 'mat_STICK_START_CLASS', ..which looks a little bit strange...

sllethe: changed into 'mat-stick-start' and 'mat-stick-end'

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Because 'this.defineRestrictions();' and ' this.sticker();' functions need to be called when the scrollable container is scrolled. 
I saw that, in 'scrollable.ts', the interface will only emit a scroll event when the scrollable container is scrolled.

src/lib/sticky-header/sticky-header-dir.ts
sllethe: :+1: Great!

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Yes. It will cause slowdown. But I have to run it on every scroll and resize event. Because I need to use it to get the top of current StickyRegion, and use the value to judge whether the current header need to be stuck or not.
Is there any way to optimize it?

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Added docs for all the variables~

src/lib/sticky-header/sticky-header-dir.ts
sllethe: OK~ So I created a generateStyleCSS() function, let it be responsible for generating all those CSS styles.

sllethe:
@jelbourn @andrewseguin , Comments addressed. Please take another look~ :+1:

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,332 @@
...
+   * The width of the sticky-header when it is stuck.
+   */
+  private _scrollingWidth: number;
+
+  constructor(_element: ElementRef,
+              scrollable: Scrollable,
+              @Optional() public parentReg: CdkStickyRegion) {
+    this.element = _element.nativeElement;
+    this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+    this.scrollableRegion = scrollable.getElementRef().nativeElement;
```
tinayuangao: Where is this `scrollableRegion` used? Is it the same as `upperScrollableContainer`?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  ngOnDestroy(): void {
+    this.upperScrollableContainer.removeEventListener('scroll', this._onScrollBind);
+    this.upperScrollableContainer.removeEventListener('resize', this._onResizeBind);
+    this.upperScrollableContainer.removeEventListener('touchmove', this._onTouchMoveBind);
+  }
+
+  attach() {
+    // this.upperScrollableContainer.addEventListener('scroll', this._onScrollBind, false);
```
tinayuangao: Remove?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,332 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Component, Directive, Input, Output,
```
sllethe: Removed unused imports

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Removed unused `scrollableRegion` and `parentRegion`

src/lib/sticky-header/sticky-header-dir.ts
sllethe: removed commented lines

sllethe:
Comments Addressed . 
And Added [using 'position:sticky' and check whether browser support it. If not , use the original implementation of sticky-header.']
Please take another look~ :+1:

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,342 @@
...
+   * Because if the sticky-header's zIndex is not the largest in current page,
+   * it may be sheltered by other element when being stuck.
+   */
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+
+  private _onScrollBind: EventListener = this.onScroll.bind(this);
+  private _onResizeBind: EventListener = this.onResize.bind(this);
+  private _onTouchMoveBind: EventListener = this.onTouchMove.bind(this);
+  isStuck: boolean = false;
+  public isIE: boolean = false;
```
tinayuangao: default is `public`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,342 @@
...
+   */
+  private _scrollingWidth: number;
+
+  constructor(_element: ElementRef,
+              scrollable: Scrollable,
+              @Optional() public parentReg: CdkStickyRegion) {
+    this.element = _element.nativeElement;
+    this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+
+    let browserVersion: string = navigator.appVersion;
```
tinayuangao: Add comments about why do you set style `top` and `position` for iPhone and not IE? Can you make sure all the cases are covered? 

nit: Move `browserVersion` related to a function. Add space between `if` and `(`.

sllethe: Added comments about why do you set style top and position for iPhone and not IE. Explained why.
And extracted new function -- detectBrowser();

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,361 @@
...
+   * It won't always work on IE phone.
+   *
+   * So the detectBrowser() function detects the current browser version. If it is about iPhone
+   * safari, set the style.position as '-webkit-sticky'. If it is an IE or an Tablet, use the
+   * original implementation. And for other circumstances, use 'position: 'sticky''.
+   */
+  detectBrowser(): void {
+    let browserVersion: string = navigator.appVersion;
+
+    if(browserVersion.includes('iPhone')) {
```
tinayuangao: How about desktop version Safari?

sllethe: Optimized detectBrowser() function. Added consideration on Safari, Chrome, Firefox, Opera, IE on both laptop and mobile devices.

sllethe:
Comments Addressed. Please take another look~ :+1: 
Added consideration on Safari, Chrome, Firefox, Opera, IE on both laptop and mobile devices.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+   *
+   * So the detectBrowser() function detects the current browser version. If it is about iPhone
+   * safari, set the style.position as '-webkit-sticky'. If it is an IE or an Tablet, use the
+   * original implementation. And for other circumstances, use 'position: 'sticky''.
+   */
+  detectBrowser(): void {
+    let browserVersion: string = navigator.appVersion;
+
+    // check if browser is using Safari
+    let isSafari = (navigator.userAgent.toLocaleLowerCase().indexOf('safari') != -1);
```
tinayuangao: nit: make `navigator.userAgent.toLocaleLowerCase()` a local variable?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+    this.isIE = browserVersion.includes('.NET');
+
+    // Check if is mobile browser
+    let browserAgent = navigator.userAgent;
+    // android
+    let isAndroid = browserAgent.indexOf('Android') > -1 || browserAgent.indexOf('Adr') > -1;
+    // ios
+    let isiOS = !!browserAgent.match(/\(i[^;]+;( U;)? CPU.+Mac OS X/);
+
+    if(isiOS == true || (isSafari == true && isChrome == false)) {
```
tinayuangao: nit: use `===`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+    // check if browser is Opera
+    let isOpera = (navigator.userAgent.toLocaleLowerCase().indexOf('presto') != -1);
+
+    // check if browser is IE
+    this.isIE = browserVersion.includes('.NET');
+
+    // Check if is mobile browser
+    let browserAgent = navigator.userAgent;
+    // android
+    let isAndroid = browserAgent.indexOf('Android') > -1 || browserAgent.indexOf('Adr') > -1;
```
tinayuangao: In which case it's `Adr`?

sllethe: It's for UCweb browser on Android phone.

src/lib/sticky-header/sticky-header-dir.ts
sllethe: changed` '=='` to `'==='`

sllethe:
Comments Addressed :+1: 
Please take another look~ (-^o^-)~*

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,382 @@
...
+  }
+
+  ngAfterViewInit(): void {
+    this.stickyParent = this.parentReg != null ?
+      this.parentReg._elementRef.nativeElement : this.element.parentElement;
+    this.originalCss = this.generateCssStyle(this.getCssValue(this.element, 'zIndex'),
+      this.getCssValue(this.element, 'position'), this.getCssValue(this.element, 'top'),
+      this.getCssValue(this.element, 'right'), this.getCssValue(this.element, 'left'),
+      this.getCssValue(this.element, 'bottom'), this.getCssValue(this.element, 'width'));
+    if(this.isIE == true) {
```
tinayuangao: nit: `===`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,382 @@
...
+              scrollable: Scrollable,
+              @Optional() public parentReg: CdkStickyRegion) {
+    this.element = _element.nativeElement;
+    this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+
+    this.detectBrowser();
+  }
+
+  ngAfterViewInit(): void {
+    this.stickyParent = this.parentReg != null ?
```
tinayuangao: Is `stickyParent` and `orignalCss` needed when `isIE` is not true?

sllethe: :+1: put all these into `if(this.isIE === true) {}`

sllethe:
Comments addressed~ (*^o^*)~
Please take another look :+1:

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+   *
+   * For Mobile: Sticky positioning works well on Android Webview, Chrome for Android, Edge,
+   * Firefox Mobile, Opera Mobile. And can also work well on Safari Mobile with a "-webkit-" prefix.
+   * It won't always work on IE phone.
+   *
+   * So the detectBrowser() function detects the current browser version. If it is about iPhone
+   * safari, set the style.position as '-webkit-sticky'. If it is an IE or an Tablet, use the
+   * original implementation. And for other circumstances, use 'position: 'sticky''.
+   */
+  detectBrowser(): void {
```
jelbourn: We really shouldn't be doing user-agent detection here. What's wrong with the check from modernizr?
https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/positionsticky.js

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+ */
+@Directive({
+  selector: '[cdkStickyRegion]',
+})
+export class CdkStickyRegion {
+  constructor(public readonly _elementRef: ElementRef) { }
+}
+
+
+const STICK_START_CLASS = 'mat-stick-start';
```
jelbourn: Add comments to these css classes explaining what they're for

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+  private _onScrollBind: EventListener = this.onScroll.bind(this);
+  private _onResizeBind: EventListener = this.onResize.bind(this);
+  private _onTouchMoveBind: EventListener = this.onTouchMove.bind(this);
+  isStuck: boolean = false;
+  isIE: boolean = false;
+
+
+  /**
+   * The element with the 'cdkStickyHeader' tag
+   */
```
jelbourn: You can condense short JsDoc to one line:
```ts
/** The element with the 'cdkStickyHeader' tag */
```

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+      width: width,
+    };
+    return targetCSS;
+}
+
+
+  private getCssValue(element: any, property: string): any {
+    let result: any = '';
+    if (typeof window.getComputedStyle !== 'undefined') {
+      result = window.getComputedStyle(element, '').getPropertyValue(property);
```
jelbourn: `getComputedStyle` is expensive to invoke and repeated calls to it should be avoided. You use of this function calls it over and over again for the same element.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,381 @@
...
+    } else if (isChrome == true || isFirefox == true || isAndroid == true || isOpera == true) {
+      this.element.style.top = '0px';
+      this.element.style.position = 'sticky';
+    }else {
+      this.isIE = true;
+    }
+  }
+
+  attach() {
+    Observable.fromEvent(this.upperScrollableContainer, 'scroll').debounceTime(5)
```
jelbourn: Any event handlers added by this directive need to be removed upon destruction

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Use what modernizr does on compatibility instead of get the browser version directly.
Added new functions:  `getSupportList()` and `setStrategyAccordingToCompatibility()`

src/lib/sticky-header/sticky-header-dir.ts
sllethe: added 'padding' and 'stickyRegionHeight' variables to avoid calling 'getComputedStyle()' too many times (which is expensive).

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+
+
+// STICK_START_CLASS is used to mark a header is stuck.
+const STICK_START_CLASS = 'cdk-sticky-header-start';
+// STICK_END_CLASS is used to mark a header is unstuck.
+const STICK_END_CLASS = 'cdk-sticky-header-end';
+@Directive({
+  selector: '[cdkStickyHeader]',
+})
+/**
```
andrewseguin: Move docs above `@Directive`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+   * For Desktop: Sticky positioning works well on Chrome, Edge, Firefox and Opera. And can
+   * also work well on Safari with a "-webkit-" prefix. It only does not work on IE.
+   *
+   * For Mobile: Sticky positioning works well on Android Webview, Chrome for Android, Edge,
+   * Firefox Mobile, Opera Mobile. And can also work well on Safari Mobile with a "-webkit-" prefix.
+   * It won't always work on IE phone.
+   *
+   */
+  getSupportList(): void {
+    let prefixTestList = ['', '-webkit-', '-ms-', '-moz-', '-o-'];
```
andrewseguin: Is this from Modernizer? Should we provide attribution and a link to the code in case we want to cross-reference it in the future?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+   * Get the first element from this._supportList. Set it as a prefix of
+   * sticky positioning.
+   *
+   * If the this._supportList is empty, which means the browser does not support
+   * sticky positioning. Set isIE as 'true' and use the original implementation of
+   * sticky-header.
+   */
+  setStrategyAccordingToCompatibility(): void {
+    if(this._supportList.length == 0) {
+      this.isIE = true;
```
andrewseguin: Can we rename this to not be specific to IE? Though it may just be IE, the variable should be something like `isBrowserStickyPositionUnsupported` (though I'd probably reverse the boolean so that we have `isStickyPositionSupported` and set it to true if this._supportList is not empty.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+  /**
+   * Get the first element from this._supportList. Set it as a prefix of
+   * sticky positioning.
+   *
+   * If the this._supportList is empty, which means the browser does not support
+   * sticky positioning. Set isIE as 'true' and use the original implementation of
+   * sticky-header.
+   */
+  setStrategyAccordingToCompatibility(): void {
+    if(this._supportList.length == 0) {
```
andrewseguin: Can we just get the support list from this function and remove the need for `_supportList` and `isIE`?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+              @Optional() public parentReg: CdkStickyRegion) {
+    this.element = _element.nativeElement;
+    this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+
+    this.getSupportList();
+    this.setStrategyAccordingToCompatibility();
+  }
+
+  ngAfterViewInit(): void {
+    if(this.isIE === true) {
```
andrewseguin: add space before parenthesis in `if(`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+
+  private _onResizeSubscription: Subscription;
+
+  /**
+   * A list of string which can be set to sticky-header's style.position
+   * and make Sticky positioning work.
+   */
+  private _supportList: Array<string>;
+
+  constructor(_element: ElementRef,
```
andrewseguin: can remove the underscore since its not saved as a private member of the class

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+              scrollable: Scrollable,
+              @Optional() public parentReg: CdkStickyRegion) {
+    this.element = _element.nativeElement;
+    this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+
+    this.getSupportList();
+    this.setStrategyAccordingToCompatibility();
```
andrewseguin: I imagine this will be a problem for Universal, should check if the platform is browser with `Platform` from `core/platform`.

src/lib/sticky-header/sticky-header-dir.ts
andrewseguin: Shouldnt need to evaluate if true, just use `if (this.isIE)`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  ngAfterViewInit(): void {
+    if(this.isIE === true) {
+      this.stickyParent = this.parentReg != null ?
+        this.parentReg._elementRef.nativeElement : this.element.parentElement;
+      this.originalCss = this.generateCssStyle(this.getCssValue(this.element, 'zIndex'),
```
andrewseguin: Difficult to parse this, can you add more newlines between params or save variables and then pass it in?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+
+  /**
+   * A list of string which can be set to sticky-header's style.position
+   * and make Sticky positioning work.
+   */
+  private _supportList: Array<string>;
+
+  constructor(_element: ElementRef,
+              scrollable: Scrollable,
+              @Optional() public parentReg: CdkStickyRegion) {
```
andrewseguin: Can you expand what `Reg` is? Is it short for region?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+      let prefix: string = this._supportList[0];
+
+      this.element.style.top = '0px';
+      this.element.style.position = prefix + 'sticky';
+    }
+  }
+
+  attach() {
+    this._onScrollSubscription =  Observable.fromEvent(this.upperScrollableContainer, 'scroll')
+      .debounceTime(5).subscribe(() => this.defineRestrictionsAndStick());
```
andrewseguin: The 5 should be a constant defined above the directive with a comment about how we came up with it

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+    }else {
+      let prefix: string = this._supportList[0];
+
+      this.element.style.top = '0px';
+      this.element.style.position = prefix + 'sticky';
+    }
+  }
+
+  attach() {
+    this._onScrollSubscription =  Observable.fromEvent(this.upperScrollableContainer, 'scroll')
```
andrewseguin: remove extra space before Observable

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+      this.stickElement();
+    }
+  }
+
+  /**
+   * define the restrictions of the sticky header(including stickyWidth,
+   * when to start, when to finish)
+   */
+  defineRestrictions(): void {
+    if(this.stickyParent == null) {
```
andrewseguin: can be `if (!this.stickyParent)`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  defineRestrictionsAndStick(): void {
+    this.defineRestrictions();
+    this.sticker();
+  }
+
+  /**
+   * This function is used to generate a variable which contains 7 css styles.
+   * @param zIndex
```
andrewseguin: Can remove the `@param` and `@returns`. Should make sure the types are correct in the function signature

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,409 @@
...
+   * @param right
+   * @param left
+   * @param bottom
+   * @param width
+   * @returns {{zIndex: any, position: any, top: any, right: any,
+   * left: any, bottom: any, width: any}}
+   */
+  generateCssStyle(zIndex:any, position:any, top:any, right:any,
+                   left:any, bottom:any, width:any): any {
+    let targetCSS = {
```
andrewseguin: Should be able to just return this

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Deleted variable `_supportList`, and changed `isIE` to `isStickyPositionSupported`

src/lib/sticky-header/sticky-header-dir.ts
sllethe: changed `isIE` to `isStickyPositionSupported`, and set it as true when `supportList` is empty

src/lib/sticky-header/sticky-header-dir.ts
sllethe: removed extra space before Observable

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Set debounce time as a const variable and added docs to explain why it is 5.

src/lib/sticky-header/sticky-header-dir.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Removed the @param and @returns and make sure the types are correct in the function signature

src/lib/sticky-header/sticky-header-dir.ts
sllethe: :+1:

sllethe: returned targetCSS;

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,416 @@
...
+   * Get the first element from this._supportList. Set it as a prefix of
+   * sticky positioning.
+   *
+   * If the this._supportList is empty, which means the browser does not support
+   * sticky positioning. Set isStickyPositionSupported as 'true' and use the original
+   * implementation of sticky-header.
+   */
+  setStrategyAccordingToCompatibility(): void {
+    let supportList = this.getSupportList();
+    if(supportList.length == 0) {
```
tinayuangao: nit: add space between `if` and `(`. Use `===`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,416 @@
...
+      this.isStickyPositionSupported = true;
+    }else {
```
tinayuangao: nit: add space between `}` and `else`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,416 @@
...
+   *
+   * The implementation references the compatibility checking in Modernizer
+   * (https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/positionsticky.js).
+   */
+  getSupportList(): Array<string> {
+    let prefixTestList = ['', '-webkit-', '-ms-', '-moz-', '-o-'];
+    let supportList: Array<string> = new Array<string>();
+    let stickyText = '';
+    for (let i = 0; i < prefixTestList.length; i++ ) {
+      stickyText += 'position:' + prefixTestList[i] + 'sticky;';
```
tinayuangao: Why do you need `+=` here? Add comments?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,416 @@
...
+   * Because if the sticky-header's zIndex is not the largest in current page,
+   * it may be sheltered by other element when being stuck.
+   */
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+
+  private _onScrollBind: EventListener = this.onScroll.bind(this);
+  private _onResizeBind: EventListener = this.onResize.bind(this);
+  private _onTouchMoveBind: EventListener = this.onTouchMove.bind(this);
+  isStuck: boolean = false;
+  isStickyPositionSupported: boolean = false;
```
tinayuangao: Add docs about this variable.
Is `true` means `position: sticky` supported?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,416 @@
...
+   * If the this._supportList is empty, which means the browser does not support
+   * sticky positioning. Set isStickyPositionSupported as 'true' and use the original
+   * implementation of sticky-header.
+   */
+  setStrategyAccordingToCompatibility(): void {
+    let supportList = this.getSupportList();
+    if(supportList.length == 0) {
+      this.isStickyPositionSupported = true;
+    }else {
+      let prefix: string = supportList[0];
```
tinayuangao: Only need `supportList[0]`? Add comments

src/lib/sticky-header/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import {CdkStickyRegion, CdkStickyHeader} from './sticky-header-dir';
+
+
+
+@NgModule({
+  imports: [OverlayModule, MdCommonModule, CommonModule],
+  declarations: [CdkStickyRegion, CdkStickyHeader],
+  exports: [CdkStickyRegion, CdkStickyHeader, MdCommonModule],
+})
+export class StickyHeaderModule {}
```
jelbourn: Just noticed that all this code is in `src/lib`. As part of the cdk, it should all be under `src/cdk`. You could do this in a another PR after this one.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,420 @@
...
+ */
+@Directive({
+  selector: '[cdkStickyRegion]',
+})
+export class CdkStickyRegion {
+  constructor(public readonly _elementRef: ElementRef) { }
+}
+
+
+// STICK_START_CLASS is used to mark a header is stuck.
```
jelbourn: `/** Class applied when the header is "stuck" */`

`/** Class applied when the header is not "stuck" */`

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Nope~ Do not need `+=`, should use `=`.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,420 @@
...
+   */
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+
+  isStuck: boolean = false;
+  /**
+   * isStickyPositionSupported == true, means current browser support 'position: sticky'
+   * or 'position: -webkit-sticky'.
+   * isStickyPositionSupported == false, means current browser does not support sticky
+   * positioning and need to use the original implementation to get the sticky effect.
+   * @type {boolean}
```
jelbourn: Omit type from JsDoc

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,420 @@
...
+   * it may be sheltered by other element when being stuck.
+   */
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+
+  isStuck: boolean = false;
+  /**
+   * isStickyPositionSupported == true, means current browser support 'position: sticky'
+   * or 'position: -webkit-sticky'.
+   * isStickyPositionSupported == false, means current browser does not support sticky
+   * positioning and need to use the original implementation to get the sticky effect.
```
jelbourn: This whole comment can be shortened to 
`Whether the browser support CSS sticky positioning.`

src/lib/sticky-header/sticky-header-dir.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Yes. Because `supportList` contains all the prefix that can make sticky positioning work in the current browser. We only need to get one prefix and make `position: prefix + 'sticky'`, then sticky position will work.

src/lib/sticky-header/sticky-header-dir.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,420 @@
...
+ */
+import {Directive, Input,
+  OnDestroy, AfterViewInit, ElementRef, Optional} from '@angular/core';
+import {Scrollable} from '../core/overlay/scroll/scrollable';
+import {extendObject} from '../core/util/object-extend';
+import {Observable} from 'rxjs/Observable';
+import {Subscription} from 'rxjs/Subscription';
+import 'rxjs/add/observable/fromEvent';
+import 'rxjs/add/operator/debounceTime';
+import {createElement} from "@angular/core/src/view/element";
```
jelbourn: Deep imports into `@angular/core` are not allowed

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,420 @@
...
+  /**
+   * 'getBoundingClientRect().top' of CdkStickyRegion of current sticky header.
+   * It is used with '_scrollFinish' to judge whether the current header
+   * need to be stuck.
+   */
+  private _containerStart: number;
+  /**
+   * It is 'The bottom of CdkStickyRegion of current sticky header - the height
+   * of current header', which is used with '_containerStart' to judge whether the current header
+   * need to be stuck.
```
jelbourn: I don't quite follow what this comment is saying

sllethe:
Comments Addressed. :+1: 
Please take another look~

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,424 @@
...
+    this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+    this.setStrategyAccordingToCompatibility();
+  }
+
+  ngAfterViewInit(): void {
+    if (!this.isStickyPositionSupported) {
+      this.stickyParent = this.parentRegion != null ?
+        this.parentRegion._elementRef.nativeElement : this.element.parentElement;
+      this.originalCss = this.generateCssStyle(
+        this.getCssValue(this.element, 'zIndex'),
```
jelbourn: This is still calling `getComputedStyle` over and over again.

Also, for top, left, right, bottom, height, and width you probably want the results of `getBoundingClientRect` instead of `getComputedStyle`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,424 @@
...
+        this.getCssValue(this.element, 'position'),
+        this.getCssValue(this.element, 'top'),
+        this.getCssValue(this.element, 'right'),
+        this.getCssValue(this.element, 'left'),
+        this.getCssValue(this.element, 'bottom'),
+        this.getCssValue(this.element, 'width'));
+      this.padding = this.getCssValue(this.element, 'padding');
```
jelbourn: Why do you need this padding as opposed to just using the height from `getBoundingClientRect`?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,424 @@
...
+   * also work well on Safari with a "-webkit-" prefix. It only does not work on IE.
+   *
+   * For Mobile: Sticky positioning works well on Android Webview, Chrome for Android, Edge,
+   * Firefox Mobile, Opera Mobile. And can also work well on Safari Mobile with a "-webkit-" prefix.
+   * It won't always work on IE phone.
+   *
+   * The implementation references the compatibility checking in Modernizer
+   * (https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/positionsticky.js).
+   */
+  getSupportList(): Array<string> {
```
jelbourn: We should move the code for checking whether position-sticky is supported to `src/cdk/platform` as a standalone check.

jelbourn: `string[]`

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,424 @@
...
+    let prefixTestList = ['', '-webkit-', '-ms-', '-moz-', '-o-'];
```
jelbourn: Can we just create a CSS class that has all of these applied and use that in the support check?

Also, I don't think `moz` and `ms` are necessary. Do you have a resource for this?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,424 @@
...
+    let supportList: Array<string> = new Array<string>();
+    let stickyText = '';
+    for (let i = 0; i < prefixTestList.length; i++ ) {
+      stickyText = 'position:' + prefixTestList[i] + 'sticky;';
+      // Create a DOM to check if the browser support current prefix for sticky-position.
+      let div = document.createElement('div');
+      let body = document.body;
+      div.style.cssText = 'display:none;' + stickyText;
+      body.appendChild(div);
+      let isSupport = /sticky/i.test(window.getComputedStyle(div).position);
```
jelbourn: We should ensure that this call to `getComputedStyle` happens at the same time as other `getComputedStyle` calls

src/lib/sticky-header/sticky-header-dir.ts
jelbourn: This check needs to be gated on whether or not we're on the browser (and not rendering with `platform-server`)

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,424 @@
...
+  onScroll(): void {
+    this.defineRestrictionsAndStick();
+  }
+
+  onTouchMove(): void {
+    this.defineRestrictionsAndStick();
+  }
+
+  onResize(): void {
+    this.defineRestrictionsAndStick();
```
jelbourn: If you're going to unstick on resize, why call this at all?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,424 @@
...
+  defineRestrictions(): void {
+    if (!this.stickyParent) {
+      return;
+    }
+    let containerTop: any = this.stickyParent.getBoundingClientRect();
+    let elemHeight: number = this.element.offsetHeight;
+    this._containerStart = containerTop.top;
+
+    // the padding of the element being stuck
+    let paddingNumber: any = Number(this.padding.slice(0, -2));
```
jelbourn: `parseInt` automatically ignores trailing non-numeric characters

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Added check whether we're on the browser

src/lib/sticky-header/sticky-header-dir.ts
sllethe: If we found that the current browser support sticky positioning. The original implementation of sticky-header will not be called, which means we will not call the rest of the `getComputedStyle` if the current browser support sticky positioning.
The rest `getComputedStyle` will only be called when the in-build sticky positioning is not supported.

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Because I need to judge whether the current header need to be stuck. Calling `defineRestrictionsAndStick()` to see the current state of the header. If the current header is stuck, we need to unstuck it and then stuck it again to reshape the sticky header. And if the current header is not stuck, we will do nothing. Just let it being scrolled up or down.

PR #5178, opened 2017-06-17
fix(dialog): set aria-labelledby based on the md-dialog-title

crisbeto:
* [Based on the accessibility guidelines](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal), these changes add the `aria-labelledby` to dialogs that use `md-dialog-title`, which causes the screen reader to read out the title. E.g. before NVDA would read out "Dialog", but now it reads out "Neptune dialog".
* Adds the MdDialogContainer instance to the dialog injector.
* Slight refactor of the dialog injector to make it a little easier to add more tokens.

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -6,7 +6,15 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Injector, ComponentRef, Injectable, Optional, SkipSelf, TemplateRef} from '@angular/core';
+import {
+  Injector,
+  InjectionToken,
+  ComponentRef,
+  Injectable,
+  Optional,
+  SkipSelf,
+  TemplateRef,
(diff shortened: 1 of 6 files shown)


src/lib/dialog/dialog.ts
```diff
@@ -209,6 +214,29 @@ export class MdDialog {
...
+   * @param container Dialog container element that wraps all of the contents.
+   * @returns The custom injector that can be used inside the dialog.
+   */
+  private _createInjector<T>(
+      config: MdDialogConfig,
+      dialogRef: MdDialogRef<T>,
+      dialogContainer: MdDialogContainer): DialogInjector {
+
+    let userInjector = config && config.viewContainerRef && config.viewContainerRef.injector;
+    let injectionTokens = new Map();
```
jelbourn: WeakMap?

PR #5181, opened 2017-06-17
fix(select): wrong option amount being read out by NVDA

crisbeto:
Fixes an issue that caused the wrong amount of options to be read out by NVDA. E.g. if the select has 3 options, NVDA reads out "{{value}} selected, 2 of 4". The issue seems to come from the fact that NVDA considers the trigger as one of the options, potentially because it's clickable.

### DIFF
--- a/src/lib/select/select.html
+++ b/src/lib/select/select.html
@@ -1,4 +1,10 @@
-<div class="mat-select-trigger" cdk-overlay-origin (click)="toggle()" #origin="cdkOverlayOrigin" #trigger>
+<div
+  cdk-overlay-origin
+  class="mat-select-trigger"
+  aria-hidden="true"
+  (click)="toggle()"
+  #origin="cdkOverlayOrigin"
+  #trigger>
   <span
     class="mat-select-placeholder"
     [class.mat-floating-placeholder]="_selectionModel.hasValue()"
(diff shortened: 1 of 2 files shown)


kara:
@crisbeto Thanks for tackling this; it's been driving me crazy. This change definitely helps when navigating through the list and when the list is open, but I've noticed that the trigger is still being counted after you select an option. 

So if you open the select, you'll hear "Coffee, 6 of 10", but after you select it, you'll hear "Coffee, 6 of 11". Can we address that? It might be more confusing if the total count is inconsistent.

@jelbourn What do you think about using aria-hidden?

crisbeto:
That sounds weird @kara. Perhaps we should be using `aria-hidden` all the time? On the first pass I tried it out that way and it still worked fine (I presume, it's because we focus the selected option which still causes the screen reader to read it out). I opted for the conditional `aria-hidden`, because it's more accurate as to what is going on.

jelbourn:
If the trigger is always `aria-hidden`, we'd need to make sure that the select itself is still appropriately labeled. We should look at other dropdown implementations for comparison.

crisbeto:
Alright, I got some time to revisit this one. Setting the `aria-hidden="true"` all the time still works on the screen readers that I could get my hands on (VoiceOver and NVDA), because we have an `aria-label` on the select already. I updated the PR to always have `aria-hidden` and left in a unit test as a sanity check.

PR #5182, opened 2017-06-17
build: firebase integration for dashboard

DevVersion:
* As part of the initial dashboard setup, this installs Angular Fire and Angular Material in the dashboard.
* For testing purposes the list of all payload results will be printed right now.

EDIT: Maybe this is rather a `chore` instead of a `build change`. But should be fine though..

### DIFF
--- a/tools/dashboard/package.json
+++ b/tools/dashboard/package.json
@@ -16,10 +16,13 @@
     "@angular/core": "^4.2.2",
     "@angular/forms": "^4.2.2",
     "@angular/http": "^4.2.2",
+    "@angular/material": "^2.0.0-beta.6",
     "@angular/platform-browser": "^4.2.2",
     "@angular/platform-browser-dynamic": "^4.2.2",
     "@angular/router": "^4.2.2",
+    "angularfire2": "^4.0.0-rc.1",
     "core-js": "^2.4.1",
+    "firebase": "^4.1.2",
     "rxjs": "^5.1.0",
     "zone.js": "^0.8.12"
   },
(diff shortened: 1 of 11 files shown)


tools/dashboard/package.json
```diff
@@ -16,10 +16,13 @@
     "@angular/core": "^4.2.2",
     "@angular/forms": "^4.2.2",
     "@angular/http": "^4.2.2",
+    "@angular/material": "^2.0.0-beta.6",
```
jelbourn: If we make this point to the build snapshot it could help give us a heads-up if something goes wrong

tools/dashboard/src/app/dashboard-app.ts
```diff
@@ -1,8 +1,32 @@
...
+  /** Observable that emits all payload results from Firebase. */
+  payloads: FirebaseListObservable<PayloadResult[]>;
+
+  constructor(database: AngularFireDatabase) {
+    this.payloads = database.list(`payloads`);
+  }
+}
+
+/** Interface that describes the payload results from the Firebase database. */
+interface PayloadResult {
```
jelbourn: Move this to a file like `data-definitions.ts`?

DevVersion: Thought about it but wanted to wait until we have more other components. I'm changing it anyway now.

tools/dashboard/package.json
DevVersion: Yep. I agree

PR #5183, opened 2017-06-17
chore: move OverlayModule and related providers into index file

crisbeto:
Moves the `OverlayModule` and the `OVERLAY_PROVIDERS` into the `index.ts`, as opposed to the `overlay-directives.ts`, in order to match the setup in the rest of the modules.


jelbourn:
Caretaker: there is the _possibility_ this could affect internal users (since people can deep-import), but it is rather unlikely.

PR #5191, opened 2017-06-18
fix(menu): role being set on the wrong element

crisbeto:
Currently, the `menu` role is set on the `md-menu` directive, however the it is inert and won't get rendered inside the overlay. These changes remove the `menu` role from the inert element and set it on the element that will get rendered inside the overlay.

### DIFF
--- a/src/lib/menu/menu.spec.ts
+++ b/src/lib/menu/menu.spec.ts
@@ -134,6 +134,18 @@ describe('MdMenu', () => {
     expect(panel.classList).toContain('custom-two');
   });
 
+  it('should set the "menu" role on the overlay panel', () => {
+    const fixture = TestBed.createComponent(SimpleMenu);
+    fixture.detectChanges();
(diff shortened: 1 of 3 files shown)


src/lib/menu/menu.spec.ts
```diff
@@ -134,6 +134,16 @@ describe('MdMenu', () => {
     expect(panel.classList).toContain('custom-two');
   });
 
+  it('should set the "menu" role on the overlay panel', () => {
+    const fixture = TestBed.createComponent(SimpleMenu);
+    fixture.detectChanges();
+    fixture.componentInstance.trigger.openMenu();
+    fixture.detectChanges();
+
+    expect(overlayContainerElement.querySelector('[role="menu"]'))
```
kara: Any reason why we're not using `getAttribute` here to check directly?

crisbeto: This was a little shorter. I'll switch to a `querySelector` and `getAttribute`.

PR #5193, opened 2017-06-18
fix(select): md-optgroup not using typography styles

crisbeto:
Resolves a TODO from a while ago to set up the `md-optgroup` to use the typography API. This wasn't done initially since the typography PR and the option group PR got in around the same time.


finalxcode:
LGTM

PR #5194, opened 2017-06-18
fix(overlay-container): overlay container throws an error if its wasn't set initially 

djleonskennedy:
fix(overlay-container): if initial theme overlay container was not set after toggle the class error occurred

We're using `ngrx` and we have themes in store, our initial state is empty (''), when we try switch (e.g to dark and back to empty error occurs `Uncaught DOMException: Failed to execute 'contains' on 'DOMTokenList': The token provided must not be empty.` this PR able to fix it

test(overlay): test for safe toggle theme of theme wasn't initialized


djleonskennedy:
I signed it!

djleonskennedy:
I signed it!

jelbourn:
Fixed in #6306

PR #5197, opened 2017-06-18
docs: add a typography guide

crisbeto:
Adds a guide that shows how to use the typography setup.

### DIFF
--- a/guides/typography.md
+++ b/guides/typography.md
@@ -95,7 +95,7 @@ own components:
 
 * `mat-font-size($config, $level)` - Gets the font-size, based on the provided config and level.
 * `mat-font-family($config)` - Gets the font-family, based on the provided config.
-* `mat-line-height($config, $level)` - Gets the line, based on the provided config and level.
+* `mat-line-height($config, $level)` - Gets the line height, based on the provided config and level.
 * `mat-font-weight($config, $level)` - Gets the font weight, based on the provided config and level.
 * `mat-typography-level-to-styles($config, $level)` - Mixin that takes in a configuration object
 and a typography level, and outputs a short-hand CSS `font` declaration.
--- a/guides/typography.md
+++ b/guides/typography.md
@@ -1,0 +1,123 @@
+# Angular Material typography
+
+### What is typography?
+Typography is a way of arranging type to make text legible, readable, and appealing when displayed.
+Angular Material's typography is based on the guidelines from the [Material Design spec][1] and is
+arranged into typography levels. Each level has a `font-size`, `line-height` and `font-weight`. The
+available levels are:
+
+* `display-4`, `display-3`, `display-2` and `display-1` - Large, one-off headers, usually
+at the top of the page (e.g. a hero header).
+* `headline` - Section heading corresponding to the `<h1>` tag.
+* `title` - Section heading corresponding to the `<h2>` tag.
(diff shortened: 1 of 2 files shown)


guides/typography.md
```diff
@@ -0,0 +1,120 @@
...
+* `mat-{{component}}-typography` - Every Material component has a corresponding typography mixin,
+which allows for its styles to be overwritten (e.g. `mat-checkbox-typography`).
+
+For more details about the typography functions and default config, see the
+[source](https://github.com/angular/material2/blob/master/src/lib/core/typography/_typography.scss).
+
+
+```scss
+// Import the requied mixins.
+@import '~@angular/material/theming';
```
crisbeto: @jelbourn this includes all the necessary typography mixins, but I'm not sure whether the file name won't cause some confusion. We could potentially copy it over and call it `typography.scss`.

crisbeto:
There we go, yours has a much better flow to it.

jelbourn:
Someone else should review since it's not somewhat co-authored

guides/typography.md
```diff
@@ -0,0 +1,123 @@
...
+descendant native elements.
+
+```html
+<!-- By default, Angular Material applies no global styles to native elements. -->
+<h1>This header is unstyled</h1>
+
+<!-- Applying the mat-tyography class adds styles for native elements. -->
+<section class="mat-typography">
+  <h1>This header will be styled</h1>
+  <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Nihil, molestias.</p>
```
andrewseguin: Why is the <p> element added in the comparison? Makes it look like its relevant to the point this section is making

guides/typography.md
```diff
@@ -0,0 +1,123 @@
...
+  $font-family: monospace,
+  $headline: mat-typography-level(32px, 48px, 700),
+  $body-1: mat-typography-level(16px, 24px, 500)
+);
+```
+
+As the above example demonstrates, a typography configuration is created by using the
+`mat-typography-config` function, which is given both the font-family and the set of typographic
+levels described earlier. Each typographic level is defined by the `mat-typography-level` function,
+which requires a font-size, line-height, and font-weight.
```
andrewseguin: Change to `font-size`, `line-height` and `font-weight` just like line 6

guides/typography.md
```diff
@@ -0,0 +1,123 @@
...
+
+For more details about the typography functions and default config, see the
+[source](https://github.com/angular/material2/blob/master/src/lib/core/typography/_typography.scss).
+
+
+### Material typography in your custom CSS
+Angular Material includes typography utility mixins and functions that you can use to customize your
+own components:
+
+* `mat-font-size($config, $level)` - Gets the font-size, based on the provided config and level.
```
andrewseguin: `font-size`, `font-family`, etc. 

Apply the code styles?

guides/typography.md
crisbeto: It's mostly to show that the paragraph would get styled as well, but I can see how it could cause confusion.

crisbeto:
Feedback is addressed.

guides/typography.md
amcdnl: I favor calling it `typography.scss` too.

PR #5213, opened 2017-06-19
chore: fix error when searching for special characters in the autocomplete demo

crisbeto:
Fixes regex special characters causing the autocomplete in the demo app to throw.

Relates to #3373.


jelbourn:
@crisbeto the sources for demo-app are built inside of Google so they're not "merge-safe"

PR #5254, opened 2017-06-20
fix(input): theming mixin error

### LINKED ISSUE #5232: Cannot create theme. "Mixins may not be defined within control directives or other mixins."
Bug

The .scss file compiles.

Scss complier (I'm using VS2017 and Web Compiler) throws an error "Mixins may not be defined within control directives or other mixins.".

Importing the theming file from material seems to cause the error.
`@import "../node_modules/@angular/material/theming";`

Migration from `2.0.0-beta.6` to `2.0.0-beta.7`

**Angular:** 4.2.2
**Material:** 2.0.0-beta.7
**OS:** Windows 10
**TypeScript:** 2.3.4

crisbeto:
* Fixes an error in the input theming mixin that was caused by a nested mixin.
* Removes the `mat-input-` prefix from the mixin local variables since they're not exposed and don't need to be prefixed.

Fixes #5232.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -43,7 +43,7 @@
     color: mat-color($foreground, disabled-text);
   }
 
-  // See mat-input-placeholder-floating mixin in input-container.scss
+  // See _mat-input-placeholder-floating mixin in input-container.scss
   input.mat-input-element:-webkit-autofill + .mat-input-placeholder,
   .mat-focused .mat-input-placeholder.mat-float {
     .mat-placeholder-required {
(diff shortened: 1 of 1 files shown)


xonuzofa:
Your fix allows me to build but is causing an issue with the prefix like so:

    <div class="col-sm-6">
        <md-input-container>
            <input mdInput type="text" placeholder="Final Price" formControlName="cost_string">
                <span md-prefix *ngIf="form.controls.cost_string.value">$&nbsp;</span>
        </md-input-container>
    </div>

screen shot 2017-06-21 at 9 52 39 pm

crisbeto:
@xonuzofa These changes only rename some variables and move the mixin out of the other mixin, there is no way for them the affect the prefix. I've checked it against our demo app and it works as expected. That being said, I think your issue comes from the fact that you're using `md-prefix`, instead of `mdPrefix`. If you change your `span` to this, it should work:

```
<span mdPrefix *ngIf="form.controls.cost_string.value">$&nbsp;</span>
```

da45:
Hi, 
with beta7, md-prefix is not wokring !
`        <md-input-container> 
            <input mdInput placeholder="Recherche ... " [(ngModel)] = 'searchStr'>
            <span md-prefix>
                <i class="material-icons app-input-icon">search</i>
            &nbsp;
            </span>
       </md-input-container>  `

I obtain this strange result (regression)
bug

Should I open a new bug?!

xonuzofa:
You are absolutely correct.  Thanks!

aarmora:
Is there any planned time for this to be merged? Or this is a workaround in the meantime?

PR #5257, opened 2017-06-20
feat(tab-nav-bar): support disabling tab links

### LINKED ISSUE #5208: Please add disabled input to md-tab-link
#### request
Please add a (Input() disabled: boolean to the md-tab-link directive

Make the tab inactive and on-clickable

It's not supported

Depending on a specific situation, some tabs might not be available yet (e.g. when gtabs are used in a wizard kind of interface)

ng 4.1.3
material 2.0.0-beta-6

DevVersion:
* Adds support for disabling tab links inside of the tab-nav-bar
* No longer requires having an extra directive for the ripples of tab links (no exposing of attributes like `mdRippleColor` - which could be flexible but should not be public API here)

**Note**:

 I was not able to create a test that confirms that no ripples are showing up if tab links are disabled. This is because the ripples are disabled through the `pointer-events` CSS property here.

In theory it would be better to use `mdRippleDisabled` here but this would prevent us from using the disabled mixin and we would still need to use a separate class.

cc. @jelbourn, @andrewseguin  for thoughts.

Closes #5208

### DIFF
--- a/src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
+++ b/src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
@@ -11,6 +11,7 @@ import {
   Component,
   Directive,
   ElementRef,
+  HostBinding,
   Inject,
   Input,
   NgZone,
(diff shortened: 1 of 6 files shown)


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -92,16 +93,28 @@ export class MdTabNav implements AfterContentInit, OnDestroy {
...
 /**
  * Link inside of a `md-tab-nav-bar`.
  */
 @Directive({
   selector: '[md-tab-link], [mat-tab-link], [mdTabLink], [matTabLink]',
-  host: {'class': 'mat-tab-link'}
+  inputs: ['disabled'],
+  host: {
+    'class': 'mat-tab-link',
+    '[class.mat-tab-disabled]': 'disabled'
```
jelbourn: Also needs `aria-disabled` and setting `tabIndex` to `-1` (same as `MdAnchor`) with corresponding tests.

DevVersion: Addressed this.

PR #5261, opened 2017-06-20
feat(sticky rows): add initial sticky rows component

DerekLouie:


src/demo-app/data-table/data-table-demo.scss
```diff
@@ -2,13 +2,14 @@
...
   display: flex;
   flex-direction: column;
-  max-height: 800px;
   background: white;
 
   // Table fills in the remaining area with a scroll
   .cdk-table {
     flex: 1 1 auto;
-    overflow: auto;
+    &.ckd-table-sticky-rows-enabled {
```
andrewseguin: Class is `ckd`, should be `cdk`

src/lib/core/data-table/_data-table.scss
```diff
@@ -0,0 +1,30 @@
+/*CSS FOR THE DATA TABLE*/
```
andrewseguin: Remove comment

src/lib/core/data-table/data-table.ts
```diff
@@ -62,11 +62,16 @@ export class HeaderRowPlaceholder {
 @Component({
   selector: 'cdk-table',
   template: `
-    <ng-container headerRowPlaceholder></ng-container>
-    <ng-container rowPlaceholder></ng-container>
+    <div class="cdk-table-body" role="presentation"
+      [style.marginTop.px]="stickyRowsMarginTop"
```
andrewseguin: indent more

src/lib/core/data-table/data-table.ts
```diff
@@ -62,11 +62,16 @@ export class HeaderRowPlaceholder {
...
+      [style.height.px]="tableHeight - stickyRowsMarginTop">
+      <ng-container headerRowPlaceholder></ng-container>
+      <ng-container rowPlaceholder></ng-container>
+    </div>
   `,
   host: {
     'class': 'cdk-table',
+    '[class.ckd-table-sticky-rows-enabled]': 'stickyRowsEnabled',
```
andrewseguin: Should be `cdk-table-sticky-rows-enabled`

src/lib/core/data-table/data-table.spec.ts
```diff
@@ -224,6 +225,29 @@ describe('CdkTable', () => {
       [data[2].c, data[2].b],
     ]);
   });
+
+  describe('with sticky rows', () => {
```
andrewseguin: Need more tests to match the logic added (especially the calculations)

src/lib/core/data-table/data-table.ts
```diff
@@ -115,7 +120,18 @@ export class CdkTable<T> implements CollectionViewer {
   }
   private _dataSource: DataSource<T>;
 
-  // Placeholders within the table's template where the header and data rows will be inserted.
+  /** Used to toggle the neccessary class for sticky rows. */
```
andrewseguin: Shouldnt need to preface variable comments with "Used to"

src/lib/core/data-table/data-table.ts
```diff
@@ -115,7 +120,18 @@ export class CdkTable<T> implements CollectionViewer {
...
+  public stickyRowsEnabled  = false;
+
+  /** Used to style the table to accomodate for sticky rows. */
+  public stickyRowsMarginTop: number;
+
+  /** Used to maintain the correct table height when sticky rows are present. */
+  public tableHeight: number;
+
+  /**  Placeholders within the table's template where the header and data
```
andrewseguin: Add a new line to the first line of the comment so it isn't on the `/**` line

src/lib/core/data-table/data-table.ts
```diff
@@ -196,6 +215,32 @@ export class CdkTable<T> implements CollectionViewer {
     }
 
     this._isViewInitialized = true;
+
+    this._StickyRows.changes.subscribe((queryList: QueryList<CdkStickyRow>) => {
```
andrewseguin: What changes do we anticipate here? Do we have a strategy for adapting if the sticky row's content causes it to increase/decrease in size?

src/lib/core/data-table/data-table.ts
```diff
@@ -196,6 +215,32 @@ export class CdkTable<T> implements CollectionViewer {
...
+             *  gets added to the table body to properly calculate the table's height.
+             */
+            this.tableHeight  = this.elementRef.nativeElement.getBoundingClientRect().height;
+            this._changeDetectorRef.markForCheck();
+          });
+        } else {
+          this.stickyRowsEnabled = false;
+        }
+
+        queryList.forEach((item) => {
```
andrewseguin: How does this work with multiple sticky rows? Can you reproduce in demo?

src/lib/core/data-table/data-table.ts
DerekLouie: This wasn't so much to re-bootstrap anything if changes are made, but more so because this is the only way I could find to signal when the table is rendered enough to start doing height calculations :O.

Conceivably though if a user adds the "cdkStickyRow" directive to a row after the table is rendered this will re-adjust the table to accommodate that row (regardless of height or color).

If I am understanding the question correctly I do not believe the row's content should be allowed to override the table's set height (in case the table needs to be siloed to a particular area of the page). 

I believe the code adapts for rows of varying heights :).

src/lib/core/data-table/data-table.ts
DerekLouie: This certainly works for multiple sticky rows, I will make an example in the demo!

PR #5278, opened 2017-06-21
refactor(button, chips): use toString method for aria-disabled

DevVersion:
For buttons and chips the value of the aria-disabled host binding is resolved from an extra method on the class instance.

As already done in other components, the button and chips should just use `boolean.toString()` which is less payload and also makes the code more readable.

### DIFF
--- a/src/lib/button/button.ts
+++ b/src/lib/button/button.ts
@@ -193,7 +193,7 @@ export class MdButton extends _MdButtonMixinBase implements OnDestroy, CanDisabl
              a[mat-button], a[mat-raised-button], a[mat-icon-button], a[mat-fab], a[mat-mini-fab]`,
   host: {
     '[attr.disabled]': 'disabled || null',
-    '[attr.aria-disabled]': '_isAriaDisabled',
+    '[attr.aria-disabled]': 'disabled.toString()',
     '(click)': '_haltDisabledEvents($event)',
   },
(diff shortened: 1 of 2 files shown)


src/lib/button/button.ts
```diff
@@ -193,7 +193,7 @@ export class MdButton extends _MdButtonMixinBase implements OnDestroy, CanDisabl
              a[mat-button], a[mat-raised-button], a[mat-icon-button], a[mat-fab], a[mat-mini-fab]`,
   host: {
     '[attr.disabled]': 'disabled || null',
-    '[attr.aria-disabled]': '_isAriaDisabled',
+    '[attr.aria-disabled]': 'disabled.toString()',
```
jelbourn: If we're going shorter, could also do `disabled + ''`

DevVersion: Hmm not sure about that. `toString()` explains what it does and is also pretty short.

jelbourn: `+ ''` is pretty idiomatic javascript imo

DevVersion: Yeah it is. But kinda weird doing this in a host binding while we would need to escape the quote (or use the double quote)

I personally prefer `toString()` as it's describing what it does and is also consistent with all other `aria-disabled` host bindings.

PR #5283, opened 2017-06-21
feat(dialog): 8px margin between dialog action buttons

karlhaas:
According to https://material.io/guidelines/components/dialogs.html#dialogs-specs there should be a margin of 8px between action buttons.

### DIFF
--- a/src/lib/dialog/dialog.scss
+++ b/src/lib/dialog/dialog.scss
@@ -65,6 +65,6 @@ $mat-dialog-max-height: 65vh !default;
   }
 
   :last-child {
-    margin-right: 0px;
+    margin-right: 0;
   }
 }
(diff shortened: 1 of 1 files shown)


jelbourn:
@crisbeto thoughts? Maybe this case is specific enough that the margins make sense, but I could still see it being intrusive in some cases

crisbeto:
I think that they make sense, but they'll look off in RTL @karlhaas.

karlhaas:
Is there a way to find out (in scss) if rtl is used? Something like fxLayoutGap could work well ...

crisbeto:
You can use the `[dir="rtl"]` in the selector.

karlhaas:
@crisbeto it is always just a margin between two action buttons. in my opinion it makes no difference ...

crisbeto:
It does make a difference. When it is in RTL or `align="end"` is set, the margin needs to be flipped, otherwise it'll stack with the dialog padding. Another issue that this PR introduces is that it adds a margin to the inner content of the buttons (note the "Close" button), which throws off the alignment. Here's a screenshot that shows both issues:

kmwjnsc

karlhaas:
thx @crisbeto for the explanation

Issue #5244, opened 2017-06-20
DatePicker Theming Is Off on Docs Site


arciisine:
#### Bug, feature request, or proposal:
When using the Dark/Pink theme on material.angular.io,   the month name and calendar icon are hard to read.

#### What is the expected behavior?
They should contrast with the background appropriately.

#### What is the current behavior?
The month name and calendar icon are illegible.

#### What are the steps to reproduce?

Go to http://material.angular.io

Select the Dark/Pink theme.

<img width="737" alt="screen shot 2017-06-20 at 10 06 16 am" src="https://user-images.githubusercontent.com/1264175/27337435-6a66c6d4-55a0-11e7-87e9-0b181ffece8a.png">

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Beta 7

PR #5290, opened 2017-06-22
fix(datepicker): set foreground color on datepicker content

mmalerba:


mmalerba:
partial fix for #5244

PR #5305, opened 2017-06-22
fix(datepicker): allow disabling calendar popup

### LINKED ISSUE #5017: Add ability to disable datepicker pop-up
Bug
I have a component with template like this:
```html
<md-input-container [formGroup]="formGroup">
   <input mdInput [mdDatepicker]="datepicker" [formControlName]="control.name" [readOnly]="control.isReadonly">
   <button *ngIf="!control.isReadonly" mdSuffix [mdDatepickerToggle]="datepicker"></button>
</md-input-container>
<md-datepicker *ngIf="!control.isReadonly" #datepicker></md-datepicker>
```

I want to not to render md-datepicker, so i added *ngIf="!control.isReadonly" to md-datepicker.

When control.isReadonly is false, md-datepicker is not rendered
(I mean that there is no HTML for md-datepicker)

g1shin:
This adds ability to disable datepicker pop-up in read-only mode.

Fixes #5017

### DIFF
--- a/src/demo-app/datepicker/datepicker-demo.html
+++ b/src/demo-app/datepicker/datepicker-demo.html
@@ -3,25 +3,26 @@
   <md-checkbox [(ngModel)]="touch">Use touch UI</md-checkbox>
   <md-checkbox [(ngModel)]="filterOdd">Filter odd months and dates</md-checkbox>
   <md-checkbox [(ngModel)]="yearView">Start in year view</md-checkbox>
+  <md-checkbox [(ngModel)]="readOnly">Read-only mode</md-checkbox>
 </p>
 <p>
   <md-input-container>
     <input mdInput [mdDatepicker]="minDatePicker" [(ngModel)]="minDate" placeholder="Min date">
     <button mdSuffix [mdDatepickerToggle]="minDatePicker"></button>
   </md-input-container>
-  <md-datepicker #minDatePicker [touchUi]="touch"></md-datepicker>
+  <md-datepicker #minDatePicker [touchUi]="touch" [readOnly]="readOnly"></md-datepicker>
   <md-input-container>
     <input mdInput [mdDatepicker]="maxDatePicker" [(ngModel)]="maxDate" placeholder="Max date">
     <button mdSuffix [mdDatepickerToggle]="maxDatePicker"></button>
   </md-input-container>
-  <md-datepicker #maxDatePicker [touchUi]="touch"></md-datepicker>
+  <md-datepicker #maxDatePicker [touchUi]="touch" [readOnly]="readOnly"></md-datepicker>
 </p>
 <p>
   <md-input-container>
     <input mdInput [mdDatepicker]="startAtPicker" [(ngModel)]="startAt" placeholder="Start at date">
     <button mdSuffix [mdDatepickerToggle]="startAtPicker"></button>
   </md-input-container>
-  <md-datepicker #startAtPicker [touchUi]="touch"></md-datepicker>
+  <md-datepicker #startAtPicker [touchUi]="touch" [readOnly]="readOnly"></md-datepicker>
 </p>
 
 <h2>Result</h2>
@@ -44,6 +45,7 @@
   <md-datepicker
       #resultPicker
       [touchUi]="touch"
+      [readOnly]="readOnly"
       [startAt]="startAt"
       [startView]="yearView ? 'year' : 'month'">
   </md-datepicker>
@@ -59,6 +61,7 @@
   <md-datepicker
       #resultPicker2
       [touchUi]="touch"
+      [readOnly]="readOnly"
       [startAt]="startAt"
       [startView]="yearView ? 'year' : 'month'">
   </md-datepicker>
--- a/src/demo-app/datepicker/datepicker-demo.html
+++ b/src/demo-app/datepicker/datepicker-demo.html
@@ -28,7 +28,7 @@
 <h2>Result</h2>
 
 <p>
-  <button [mdDatepickerToggle]="resultPicker" [disabled]="resultPickerModel.disabled"></button>
+  <button [mdDatepickerToggle]="resultPicker"></button>
   <md-input-container>
     <input mdInput
            #resultPickerModel="ngModel"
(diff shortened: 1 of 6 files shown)


g1shin:
Fixed and ready for review. @mmalerba Input and button already had the "disabled" attribute, so I didn't need to add it to datepicker-input and datepicker-toggle.

g1shin:
Ready for review again :)

src/lib/datepicker/datepicker-input.ts
```diff
@@ -124,6 +125,17 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
   }
   private _max: D;
 
+  /** Whether the datepicker-input is disabled. */
+  @Input()
+  get disabled() { return this._disabled; }
+  set disabled(value: any) {
+    this._disabled = coerceBooleanProperty(value);
+    if (this._datepicker.disabled === undefined) {
```
mmalerba: we don't need to do this here, only modify your own disabledness

src/lib/datepicker/datepicker-toggle.ts
```diff
@@ -36,7 +36,7 @@ export class MdDatepickerToggle<D> {
   constructor(public _intl: MdDatepickerIntl) {}
 
   _open(event: Event): void {
-    if (this.datepicker) {
+    if (this.datepicker && !this.datepicker.disabled) {
```
mmalerba: add your own disabled input:

```ts
@Input()
get disabled() {
  return this._disabled === undefined ? this.datepicker.disabled : this._disabled;
}
set disabled(value) {
  this._disabled = coerceBooleanValue(value);
}
```

because buttons also have a native disabled attribute you'll have to add the host binding too: `'[disabled]': 'disabled'`

src/lib/datepicker/datepicker.ts
```diff
@@ -116,6 +117,17 @@ export class MdDatepicker<D> implements OnDestroy {
    */
   @Input() touchUi = false;
 
+  /** Whether the datepicker pop-up should be disabled. */
+  @Input()
+  get disabled() { return this._disabled; }
+  set disabled(value: any) {
+    this._disabled = coerceBooleanProperty(value);
+    if (this._disabled === undefined) {
```
mmalerba: similar to how i showed in the other example, put the delegation logic in the getter

src/lib/datepicker/datepicker-input.ts
```diff
@@ -124,6 +125,17 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
   }
   private _max: D;
 
+  /** Whether the datepicker-input is disabled. */
+  @Input()
```
mmalerba: you'll need to also add a host binding: `'[disabled]': 'disabled'`, to keep the native disabled attribute in sync with this

g1shin:
Changes made. Ready for review

src/demo-app/datepicker/datepicker-demo.html
```diff
@@ -59,7 +62,32 @@
...
+  <md-datepicker #datePicker1 [touchUi]="touch" [startAt]="startAt"
+                   [startView]="yearView ? 'year' : 'month'"></md-datepicker>
+</p>
+
+<h2>Input disabled, datepicker popup enabled</h2>
+<p>
+  <button [mdDatepickerToggle]="datePicker2"></button>
+  <md-input-container>
+    <input mdInput [mdDatepicker]="datePicker2" [(ngModel)]="date" [min]="minDate" [max]="maxDate"
+           [mdDatepickerFilter]="filterOdd ? dateFilter : null" [disabled]="true"
```
kara: Since `disabled` will always be true here, you should be able to just use the attribute.

```
<input mdInput disabled>
```

src/lib/datepicker/datepicker.ts
```diff
@@ -209,6 +220,9 @@ export class MdDatepicker<D> implements OnDestroy {
     if (this.opened) {
       return;
     }
+    if (this.disabled) {
+      return;
```
kara: Nit: Seems like you can bundle this with the existing early return on line 220:

```ts
if (this.opened || this.disabled) {
   return;
}
```

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -96,6 +95,20 @@ describe('MdDatepicker', () => {
         expect(document.querySelector('md-dialog-container')).not.toBeNull();
       }));
 
+      it('open in disabled mode should not open the calendar', async(() => {
```
kara: Curious why this test is async. It doesn't look like you're using any of async testing methods?

mmalerba: At one point I think calling `open` required it to be async, I'm not sure if that's still the case

src/demo-app/datepicker/datepicker-demo.html
```diff
@@ -3,31 +3,32 @@
...
   </md-input-container>
-  <md-datepicker #startAtPicker [touchUi]="touch"></md-datepicker>
+  <md-datepicker #startAtPicker [touchUi]="touch" [disabled]="datepickerDisabled"></md-datepicker>
 </p>
 
 <h2>Result</h2>
 
 <p>
-  <button [mdDatepickerToggle]="resultPicker"></button>
+  <button [mdDatepickerToggle]="resultPicker" [disabled]="resultPickerModel.disabled"></button>
```
mmalerba: don't need to `disabled` attr here

src/lib/datepicker/datepicker-toggle.ts
```diff
@@ -33,10 +35,20 @@ export class MdDatepickerToggle<D> {
...
+  set disabled(value) {
+    this._disabled = coerceBooleanProperty(value);
+  }
+  private _disabled: boolean;
+
   constructor(public _intl: MdDatepickerIntl) {}
 
   _open(event: Event): void {
-    if (this.datepicker) {
+    if (this.datepicker && !this._disabled) {
```
mmalerba: `disabled` (`_disabled` doesn't have the cascading behavior)

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -734,11 +747,12 @@ describe('MdDatepicker', () => {
 @Component({
   template: `
     <input [mdDatepicker]="d" [value]="date">
-    <md-datepicker #d [touchUi]="touch"></md-datepicker>
+    <md-datepicker #d [touchUi]="touch" [disabled]="disabled"></md-datepicker>
```
mmalerba: should add tests for cascade as well

src/lib/datepicker/datepicker.spec.ts
g1shin: It looks like `open` does not require async, so I removed async from other parts of the code as well.

mmalerba:
caretaker note: build files need to be updated to reflect datepicker dependency on cdk

PR #5307, opened 2017-06-22
feat(sort): add ability to manage and display sorting

andrewseguin:
Introduces the `MdSort` which contains `MdSortable` elements. The `MdSort` manages the active sort and direction while each `MdSortable` element changes and displays the sorting for its field.

`MdSortHeader` is the first `MdSortable` to be introduced, to be easily used within CdkTable.

h3kcxzdxbq3


src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  /** Overrides the disable clear value of the containing MdSort for this MdSortable. */
+  @Input()
+  get disableClear() { return this._disableClear; }
+  set disableClear(v) { this._disableClear = coerceBooleanProperty(v); }
+  private _disableClear: boolean;
+
+  /** Overrides the reverse order value of the containing MdSort for this MdSortable. */
+  @Input()
+  get reverseOrder() { return this._reverseOrder; }
```
jelbourn: I don't follow what this does

src/lib/sort/sort-header.html
```diff
@@ -0,0 +1,20 @@
+<div class="mat-sort-header-container"
+     [class.mat-sort-header-position-before]="arrowPosition == 'before'">
+  <button class="mat-sort-header-button"
```
jelbourn: Add `type="button"`

src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+@Component({
+  moduleId: module.id,
+  selector: '[md-sort-header], [mat-sort-header]',
+  templateUrl: 'sort-header.html',
+  styleUrls: ['sort-header.css'],
+  host: {
+    '(click)': '_sort.sort(this)',
+    '[class.mat-sort-header-sorted]': '_isSorted()',
+  }
```
jelbourn: Add `ViewEncapsulation.None` and `ChangeDetectionStrategy.OnPush`

src/lib/sort/sort-header.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.mat-sort-header-descending {
+  display: block;
+  transform: rotate(225deg);
+  top: $mat-sort-header-arrow-thickness;
+}
+
+.mat-sort-header-stem {
+  background: black;
```
jelbourn: Color should be part of the theme (or maybe just `currentColor`?)

src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,89 @@
...
+      throw getMdSortHeaderNotContainedWithinMdSortError();
+    }
+  }
+
+  ngOnInit() {
+    if (!this.id && this._cdkColumnDef) {
+      this.id = this._cdkColumnDef.name;
+    }
+
+    this._sort.register(this);
```
jelbourn: Should be able to do this in the constructor

src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,89 @@
...
+  set reverseOrder(v) { this._reverseOrder = coerceBooleanProperty(v); }
+  private _reverseOrder: boolean;
+
+  @Input('mat-sort-header')
+  get _id() { return this.id; }
+  set _id(v: string) { this.id = v; }
+
+  constructor(@Optional() public _sort: MdSort,
+              public _intl: MdSortIntl,
+              @Optional() public _cdkColumnDef: CdkColumnDef) {
```
jelbourn: Put the `@Optional` args next to each other?

src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,89 @@
...
+    this._sort.register(this);
+  }
+
+  ngOnDestroy() {
+    this._sort.unregister(this);
+  }
+
+  /** Whether this MdSortHeader is currently sorted in either ascending or descending order. */
+  _isSorted() {
+    return this._sort.active == this.id && this._sort.direction != '';
```
jelbourn: Just `this._sort.active == this.id && this._sort.direction`?

src/lib/sort/sort.ts
```diff
@@ -0,0 +1,147 @@
...
+  set reverseOrder(v) { this._reverseOrder = coerceBooleanProperty(v); }
+  private _reverseOrder: boolean;
+
+  /** Event emitted when the user changes either the active sort or sort direction. */
+  @Output() mdSortChange = new EventEmitter<Sort>();
+
+  /**
+   * Register function to be used by the contained MdSortables. Adds the MdSortable to the
+   * collection of MdSortables.
+   * @docs-private
```
jelbourn: Why `@docs-private` here?

src/lib/sort/sort-header.ts
andrewseguin: For setting the sort order as 

`descending -> ascending ->  [none] -> descending -> ... `

as opposed to 

` ascending -> descending' -> [none] -> descending -> ... `

src/lib/sort/sort-header.ts
andrewseguin: Cannot do this since the MdSort creates a mapping of `id` to `MdSortable` and this is the earlier lifecycle hook where we have the `id` available

src/lib/sort/sort.ts
andrewseguin: Hmm I wasn't thinking others needed it but people implementing their own `MdSortable` certainly would. Removed

src/lib/sort/sort-header.ts
jelbourn: Isn't that what `start` already does?

andrewseguin: Not quite - `start` determines where in the cycle the user is placed when first clicked. `reverseOrder` will reverse the cycle.

Sounds like I misunderstood your intentions on `start`, which would reverse the order.

We could remove one or the other and we'd lose a minor and probably unnecessary cycle. I prefer keeping `reverseOrder` over `start` though

src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+@Component({
+  moduleId: module.id,
+  selector: '[md-sort-header], [mat-sort-header]',
+  templateUrl: 'sort-header.html',
+  styleUrls: ['sort-header.css'],
+  host: {
+    '(click)': '_sort.sort(this)',
+    '[class.mat-sort-header-sorted]': '_isSorted()',
+  },
```
jelbourn: Still missing view encapsulation none and onpush

src/lib/sort/sort-header.ts
jelbourn: Yeah, my intent was that:
Default:              `asc` -> `desc` -> `none`
Start = `desc`:   `desc` -> `asc` -> `none`

jelbourn: I think having both of these is more confusing than helpful. The problem with `"reverse"` is that you have to know what the default is. Saying `start` or `startWith`, in my mind, better captures the developers intent

andrewseguin: SGTM - removed the reverse. Agreed that just `start` is enough

src/lib/sort/sort-header.ts
andrewseguin: Whoops, got removed in testing. It's back in

src/lib/sort/sort-header.html
```diff
@@ -0,0 +1,20 @@
+<div class="mat-sort-header-container" type="button"
+     [class.mat-sort-header-position-before]="arrowPosition == 'before'">
+  <button class="mat-sort-header-button"
```
andrewseguin: Ah misunderstood, thought this was an accessibility thing, putting type on the wrong element. Good now

andrewseguin:
Good for another review round

src/lib/sort/sort.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.mdSortChange.next({active: this.active, direction: this.direction});
+  }
+
+  /** Returns the next sort direction of the active sortable, checking for potential overrides. */
+  _getNextSortDirection(): SortDirection {
+    const sortable = this.sortables.get(this.active);
+    if (!sortable) { return ''; }
+
+    // Get the sort direction cycle with the potential sortable overrides.
+    const disableClear = sortable.disableClear != undefined ?
```
jelbourn: Prefer doing `== null` or `!= null` (it captures both `null` and `undefined` and is shorter)

src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  constructor(public _intl: MdSortHeaderIntl,
+              private _changeDetectorRef: ChangeDetectorRef,
+              @Optional() public _sort: MdSort,
+              @Optional() public _cdkColumnDef: CdkColumnDef) {
+    if (!_sort) {
+      throw getMdSortHeaderNotContainedWithinMdSortError();
+    }
+
+    _sort.mdSortChange.subscribe(() => _changeDetectorRef.markForCheck());
```
jelbourn: Need to unsubscribe to this on destroy

src/lib/sort/sort-header.ts
```diff
@@ -0,0 +1,90 @@
...
+  ngOnInit() {
+    if (!this.id && this._cdkColumnDef) {
+      this.id = this._cdkColumnDef.name;
+    }
+
+    this._sort.register(this);
+  }
+
+  ngOnDestroy() {
+    this._sort.unregister(this);
```
jelbourn: nit: `deregister`

src/lib/sort/sort.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      declarations: [
+        SimpleMdSortApp,
+        CdkTableMdSortApp,
+        MdSortHeaderMissingMdSortApp,
+        MdSortDuplicateMdSortableIdsApp,
+        MdSortableMissingIdApp
+      ],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(SimpleMdSortApp);
```
jelbourn: `createComponent` has to go in a separate `beforeEach` block because `compileComponents()` is asynchronous

andrewseguin:
Made the changes, thanks

src/demo-app/data-table/person-data-source.ts
```diff
@@ -48,12 +53,35 @@ export class PersonDataSource extends DataSource<any> {
...
+        case 'userId': [propertyA, propertyB] = [a.id, b.id]; break;
+        case 'userName': [propertyA, propertyB] = [a.name, b.name]; break;
+        case 'progress': [propertyA, propertyB] = [a.progress, b.progress]; break;
+        case 'color': [propertyA, propertyB] = [a.color, b.color]; break;
+      }
+
+      let valueA = isNaN(+propertyA) ? propertyA : +propertyA;
+      let valueB = isNaN(+propertyB) ? propertyB : +propertyB;
+
+      return (valueA < valueB ? -1 : 1) * (this._sort.direction == 'ascending' ? 1 : -1);
```
jelbourn: Missed a spot

src/demo-app/data-table/data-table-demo.html
```diff
@@ -33,17 +33,25 @@
...
+        ID
+      </cdk-header-cell>
       <cdk-cell *cdkCellDef="let row"> {{row.id}} </cdk-cell>
     </ng-container>
 
     <!-- Column Definition: Progress -->
     <ng-container cdkColumnDef="progress">
-      <cdk-header-cell *cdkHeaderCellDef> Progress </cdk-header-cell>
+      <cdk-header-cell *cdkHeaderCellDef
+                       md-sort-header start="descending">
```
jelbourn: Here too

andrewseguin:
D'oh, forgot to export my last fixes - should be good to go

jelbourn:
Caretaker note: this will need a new build rule

src/lib/sort/sort-header-intl.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+@Injectable()
+export class MdSortHeaderIntl {
+  sortButtonLabel = (id: string) => {
+    return `Change sorting for ${id}`;
+  }
+
+  /** A label to describe the current sort (visible only to screenreaders). */
+  sortDescriptionLabel = (id: string, direction: SortDirection) => {
+    return `Sorted by ${id} ${direction}`;
```
ErinCoughlan: If direction is 'asc' | 'desc', won't that sound strange with a screenreader? I wonder if you want to override it to be a real word, but I'm not super familiar with what the standards are here.

andrewseguin: Good catch, I'll add a check for this

Issue #5310, opened 2017-06-22
close dialog from it's own controller


twoheaded:
#### Bug, feature request, or proposal:

Feature request or proposal

#### What is the expected behavior?

How can I close md-dialog from it's own controller and send data without clicking a buttons in `md-dialog-actions`  container?

#### What is the current behavior?

It is not possible to close dialog and share data from the dialog controller.

#### What are the steps to reproduce?

Create md-dialog component and try to close it with sending a data by click on the element in `md-dialog-content`.

#### What is the use-case or motivation for changing an existing behavior?

I have an items list in the md-dialog. Every item have a checkbox and title. I wont to add checked items to data and send them by click on the `md-dialog-actions` button. And I want to add one item to the data, send it to parent controller and close dialog by click on the item title. How can I do it without crutches?

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?

Angular 4+, Material beta.7

willshowell:
You can close the dialog with `dialogRef.close(myValue)`. You don't need to click any buttons.

twoheaded:
It's possible from a parent controller, where I open my dialog window:
```javascript
  openDialog() {
    let dialogRef = this.dialog.open(FermentablesAddDialogComponent, {
      data: []
    });
    dialogRef.afterClosed().subscribe(result => {
      console.log(result);
    });
  }
```
 But how can I use it from my dialog component?

willshowell:
```ts
// FermentablesAddDialogComponent

constructor(
  private dialogRef: MdDialogRef<FermentablesAddDialogComponent>,
  @Inject(MD_DIALOG_DATA) public data: any
) { }

closeMyself() {
  this.dialogRef.close(123);
}
```
It's in the dialog overview page of the docs, but it's a little hidden

> Components created via MdDialog can inject MdDialogRef and use it to close the dialog in which they are contained. When closing, an optional result value can be provided. This result value is forwarded as the result of the afterClosed promise.

twoheaded:
It works great! I think that this small example should be added to the documentation. Thank you

willshowell:
Just did! #5311

PR #5311, opened 2017-06-22
docs(dialog): example of how to inject MdDialogRef

willshowell:
For confusion in https://github.com/angular/material2/issues/5310

### DIFF
--- a/src/lib/dialog/dialog.md
+++ b/src/lib/dialog/dialog.md
@@ -30,11 +30,7 @@ in which they are contained. When closing, an optional result value can be provi
 value is forwarded as the result of the `afterClosed` promise. 
 
 ```ts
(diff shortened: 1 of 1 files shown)


src/lib/dialog/dialog.md
```diff
@@ -29,6 +29,21 @@ Components created via `MdDialog` can _inject_ `MdDialogRef` and use it to close
 in which they are contained. When closing, an optional result value can be provided. This result
 value is forwarded as the result of the `afterClosed` promise. 
 
+```ts
+import {MdDialogRef} from '@angular/material';
+
+@Component({
+  // ...
+})
+export class YourDialog {
```
jelbourn: For compactness I would omit the `import` and write the decorator as 
```ts
@Component({/* ... */})
```

PR #5314, opened 2017-06-22
feat: remove uses of rxjs patch operators

### LINKED ISSUE #2622: Internal usages of RxJS operators should not patch the prototype of Observable

crisbeto:
Refactors the entire codebase not to use the patch operators from RxJS, because they pollute the user's setup. Instead, opts into importing the operators directly and chaining them via the `RxChain` class.

Fixes #2622.

**Note:** I'll add some linting to enforce these changes after they get in.


src/lib/core/rxjs/rx-operators.ts
```diff
@@ -0,0 +1,144 @@
...
+import {first as firstOperator} from 'rxjs/operator/first';
+import {switchMap as switchMapOperator} from 'rxjs/operator/switchMap';
+import {startWith as startWithOperator} from 'rxjs/operator/startWith';
+import {debounceTime as debounceTimeOperator} from 'rxjs/operator/debounceTime';
+import {auditTime as auditTimeOperator} from 'rxjs/operator/auditTime';
+import {takeUntil as takeUntilOperator} from 'rxjs/operator/takeUntil';
+
+/**
+ * Represents a strongly-typed chain of RxJS operators.
+ * @docs-private
```
jelbourn: I'd expand to something like
```ts
/**
 * Represents a strongly-typed chain of RxJS operators.
 *
 * We achieve strict type enforcement on the chained operators by creating types that
 * *unambiguously* match specific rxjs operators. These unambiguous types are created by
 * intersecting a "brand" to the `typeof` the existing operator. The brand (a class with a private
 * member) effectively forces nominal typing for the operators. This allows typescript to understand
 * that, for example, `filter` is *`filter`* and not, say, a map of T => boolean.
 *
 * The downside to this approach is that operators must be imported in their type-coerced form
 * rather than from the normal rxjs location.
 *
 * @docs-private
 */
```

crisbeto:
Addressed the feedback and sorted out the CI failures @jelbourn.

jelbourn:
Caretaker note: removing the prototype operators can potentially break some apps on sync

willshowell:
For folks who come searching later, these are the operators and observables that are no longer being patched onto your Observable.prototype 

```
Operators
-------------
auditTime
catch
debounceTime
do
filter
finally
first
let
map
share
startWith
switchMap
takeUntil

Observables
-------------
combineLatest
forkJoin
fromEvent
merge
of
throw
```

PR #5317, opened 2017-06-22
fix(datepicker): refactor datepicker toggle to support theming

### LINKED ISSUE #4684: Apply md-button to datePicker Button
#### Feature request:
Allow a user to add md-button and md-icon-button on the Date Picker's toggle button:

`<button mdSuffix [mdDatepickerToggle]="picker"></button>`

A user should be able to apply material button properties to allow for hover-over and ripple on the calendar button.

Applying both causes the app to crash. Excerpt of the error:

```
Unhandled Promise rejection: Template parse errors:
More than one component matched on this element.
Make sure that only one component's selector can match a given element.
Conflicting components: MdButton,MdDatepickerToggle
```
Ex:
  `<button md-button md-icon-button mdSuffix [mdDatepickerToggle]="picker"></button>`

Better usability. A user who notices a hover-over will feel more aware and inclined  to use that button.

Angular 4.0.2, Material 2.0.0-beta.5, Windows 7 & 10, TypeScript 2.2.2, Chrome

### LINKED ISSUE #4934: Change color of datetimepicker toggle
#### feature request:
I would like to change the color of the datetime picker toggle. The default color is black and i see i'ts an image and not a font.

Add an attribute **[toggleColor]="Primary"**
`<md-datepicker #dateStart startView="year" [startAt]="start" [toggleColor]="Primary"></md-datepicker>`

### LINKED ISSUE #5224: Time Picker
Time Picker is missing, i think it must have that feature

mmalerba:
@jelbourn The `md-datepicker-toggle-icon` could easily be replaced with an `md-icon` the only reason it's separate is so users don't need to load a font to use it. Would it be better to have datepicker module register an svg set with just a `'material:datepicker:toggle'` icon for now? Other components could create similar sets if they need to include icons.

BREAKING CHANGE: `[mdDatepickerToggle]` attribute has been changed to `md-datepicker-toggle` element 

fixes #5224
fixes #4934
fixes #4684

### DIFF
--- a/src/demo-app/datepicker/datepicker-demo.html
+++ b/src/demo-app/datepicker/datepicker-demo.html
@@ -8,25 +8,19 @@
 <p>
   <md-input-container>
     <input mdInput [mdDatepicker]="minDatePicker" [(ngModel)]="minDate" placeholder="Min date">
-    <button md-icon-button mdSuffix [mdDatepickerToggle]="minDatePicker">
-      <md-datepicker-toggle-icon></md-datepicker-toggle-icon>
-    </button>
+    <md-datepicker-toggle mdSuffix [for]="minDatePicker"></md-datepicker-toggle>
   </md-input-container>
   <md-datepicker #minDatePicker [touchUi]="touch" [disabled]="datepickerDisabled"></md-datepicker>
   <md-input-container>
     <input mdInput [mdDatepicker]="maxDatePicker" [(ngModel)]="maxDate" placeholder="Max date">
-    <button md-icon-button mdSuffix [mdDatepickerToggle]="maxDatePicker">
-      <md-datepicker-toggle-icon></md-datepicker-toggle-icon>
-    </button>
+    <md-datepicker-toggle mdSuffix [for]="maxDatePicker"></md-datepicker-toggle>
   </md-input-container>
   <md-datepicker #maxDatePicker [touchUi]="touch" [disabled]="datepickerDisabled"></md-datepicker>
 </p>
 <p>
   <md-input-container>
     <input mdInput [mdDatepicker]="startAtPicker" [(ngModel)]="startAt" placeholder="Start at date">
-    <button md-icon-button mdSuffix [mdDatepickerToggle]="startAtPicker">
-      <md-datepicker-toggle-icon></md-datepicker-toggle-icon>
-    </button>
+    <md-datepicker-toggle mdSuffix [for]="startAtPicker"></md-datepicker-toggle>
   </md-input-container>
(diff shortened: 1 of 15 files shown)


src/lib/datepicker/datepicker-toggle.ts
```diff
@@ -33,7 +34,12 @@ export class MdDatepickerToggle<D> {
   get _datepicker() { return this.datepicker; }
   set _datepicker(v: MdDatepicker<D>) { this.datepicker = v; }
 
-  constructor(public _intl: MdDatepickerIntl) {}
+  /** Whether the host element is an HTML button. */
+  _isButton = false;
+
+  constructor(elementRef: ElementRef, public _intl: MdDatepickerIntl) {
+    this._isButton = elementRef.nativeElement.tagName.toLowerCase() === 'button';
```
ocarreterom: Maybe can be:
```js
elementRef.nativeElement instanceof HTMLButtonElement;
```

yogeshgadge:
Wondering how does one use Fontawesome icon for md-datepicker-toggle-icon ?

mmalerba:
@yogeshgadge just use a normal `md-icon`, `md-datepicker-toggle-icon` only exists so we can ship a default icon without having to require a whole CSS font library

jelbourn:
FYI there's already 9 uses of this inside Google we'd have to update

src/demo-app/datepicker/datepicker-demo.html
```diff
@@ -7,27 +7,35 @@
 <p>
   <md-input-container>
     <input mdInput [mdDatepicker]="minDatePicker" [(ngModel)]="minDate" placeholder="Min date">
-    <button mdSuffix [mdDatepickerToggle]="minDatePicker"></button>
+    <button md-icon-button mdSuffix [mdDatepickerToggle]="minDatePicker">
+      <md-datepicker-toggle-icon></md-datepicker-toggle-icon>
+    </button>
```
jelbourn: What do you think of just having 
```html
<md-datepicker-toggle for="minDatePicker"></md-datepicker-toggle>
```

If someone wants to have a custom icon, they just do a custom `<button>` with the close behavior. `[mdDatepickerToggle]="minDatePicker"` is _longer_ than `(click)="minDatePicker.toggle()"`, so it seems like there's not much sense in breaking up the toggle behavior and the baked-in icon. 

For the baked-in version, all the things someone would typically want to set on the native `<button>` are taken care of by the component (content, aria-label, and type).

mmalerba:
Changed to suggested API, PTAL

andrewseguin:
Can you try rebasing? Github doesnt say anything but our script does

PR #5325, opened 2017-06-23
fix(memory): Unsubscribe event listeners when using Observable.fromEvent

Eddman:
Listeners created using Observable.fromEvent have to be unsubscribed to prevent memory leaks. Especially when the target is document/window.

In addition fixed a small leak in CdkMonitorFocus.

Future TODOs: few services are registering listeners on window/document (eg. FocusOriginMonitor) and theses produce huge leaks in Karma test suites. Angular creates new instances for each spec (in addition to new Zone for each spec). An issue already raised in Zone.js (angular/zone.js/issues/822) to consult what's the correct approach.


jelbourn:
@Eddman can you rebase this PR?

Eddman:
@jelbourn Done.

PR #5333, opened 2017-06-23
fix(input): prevent textarea from resizing beyond input container

crisbeto:
Prevents `textarea` instances from resizing beyond their `md-input-container`.


mmalerba:
Is there a corresponding bug for this? (just curious what issue it's fixing)

crisbeto:
It's just something I noticed while looking into something else.

mmalerba:
lgtm

PR #5342, opened 2017-06-24
chore: add tabs, expansion, and paginator to kitchen-sink

jelbourn:


jelbourn:
Hmm, the prerender task fails on Travis. Perhaps this has to do with the Angular version...

crisbeto:
It's failing because tabs don't work with server-side rendering (there's a comment about it in the kitchen sink). #5348 will fix it.

jelbourn:
@crisbeto Yeah, I had seen that tabs didn't work before (I wrote the comment), but when I ran this PR on my machine everything worked.

PR #5350, opened 2017-06-24
chore(script): automatic example module creation

amcdnl:
Enables the creation of the examples module and metadata associated with it by using jsdoc decorator tags. There are 3 required tags:

- `title`: the title to be shown
- `id`: the selector for the examples. e.g. `<!-- example(go-live) -->`
- `component`: the component class name

there are 3 other optional tags, these are:

- `addlComponents`: Comma separated list of additional component class names required for the example. Required by snackbar component
- `additionalFiles`: Comma separated list of additional files required for the example. Required by dialog component.
- `selectorName`: Comma separated list of component class names required for the demo runner.

A simple example looks like:
```
/**
 * @title Stacked chips
 * @id chips-stacked
 * @component ChipsStackedExample
 */
```

more advanced scenarios such as snackbar and dialogs look like:

```
/**
 * @title Dialog elements
 * @id dialog-elements
 * @component DialogElementsExample
 * @addlComponents DialogElementsExampleDialog
 * @additionalFiles dialog-elements-example-dialog.html
 * @selectorName DialogElementsExample, DialogElementsExampleDialog
 */
```


tools/gulp/tasks/docs.ts
```diff
@@ -177,3 +180,197 @@ function createTagNameAliaser(classPrefix: string) {
...
+
+    if (res.additionalComponents) {
+      components.push(...res.additionalComponents);
+    }
+    return `${components.join(',')},
+    `;
+  };
+
+  return `
+  import {NgModule} from '@angular/core';
```
jelbourn: It would be good to move this template to its own file.

tools/gulp/tasks/docs.ts
```diff
@@ -177,3 +180,197 @@ function createTagNameAliaser(classPrefix: string) {
...
+  import {FormsModule, ReactiveFormsModule} from '@angular/forms';
+  import {CommonModule} from '@angular/common';
+  import {
+    MdAutocompleteModule, MdButtonModule, MdButtonToggleModule, MdPaginatorModule,
+    MdCardModule, MdCheckboxModule, MdChipsModule, MdDatepickerModule,
+    MdDialogModule, MdGridListModule, MdIconModule, MdInputModule,
+    MdListModule, MdMenuModule, MdProgressBarModule, MdProgressSpinnerModule,
+    MdRadioModule, MdSelectModule, MdSidenavModule, MdSliderModule,
+    MdSlideToggleModule, MdSnackBarModule, MdTabsModule, MdToolbarModule, MdTooltipModule
+  } from '@angular/material';
```
jelbourn: `ExampleMaterialModule` and `LiveExample` can go in their own TS files and just be imported rather than included in the template.

jelbourn:
@DevVersion should also take a look

tools/gulp/tasks/docs.ts
```diff
@@ -177,3 +180,197 @@ function createTagNameAliaser(classPrefix: string) {
...
+    }
+
+    for (const match of matches) {
+      const src = fs.readFileSync(match, 'utf-8');
+      const filename = match
+        .replace('src/material-examples/', './')
+        .replace('.ts', '');
+
+      // get the decorator tags, not all required
+      const titleMatch = src.match(/@title.*/g);
```
DevVersion: Yeah most of the `additionalFiles` and `additionalComponents` could be detected using the TypeScript AST. 

The only thing that is problematic would be the "demo" title for a given demo (because it wouldn't fit into the metadata)

Although the title could be part of the JSDoc of a component.

```ts
/**
 * @title Test Demo
 */
@Component({
  ...
}
export class MyTestExample {}
```

jelbourn: Yeah, the title (and probably a description and maybe some other stuff later) will definitely have to be comments, but I think we can infer everything else.

DevVersion: Yeah. I think the TypeScript API also provides an easy way to retrieve the JSDoc of a given `Node`.

tools/gulp/tasks/docs.ts
```diff
@@ -177,3 +180,197 @@ function createTagNameAliaser(classPrefix: string) {
     return this;
   };
 }
+
+interface ExampleMetadata {
```
amcdnl: Ok, I started that at first but moved into the same file at the end cuz of similarities .

src/material-examples/dialog-elements/dialog-elements-example.ts
```diff
@@ -1,7 +1,14 @@
 import {Component} from '@angular/core';
 import {MdDialog} from '@angular/material';
 
-
+/**
+ * @title Dialog elements
+ * @id dialog-elements
+ * @component DialogElementsExample
+ * @addlComponents DialogElementsExampleDialog
+ * @additionalFiles dialog-elements-example-dialog.html
```
amcdnl: Ya, using something like a TypeScript parser we could detect these.

src/material-examples/autocomplete-overview/autocomplete-overview-example.ts
```diff
@@ -4,6 +4,11 @@ import {FormControl} from '@angular/forms';
 import 'rxjs/add/operator/startWith';
 import 'rxjs/add/operator/map';
 
+/**
+ * @title Basic autocomplete
+ * @id autocomplete-overview
```
amcdnl: Ya, sounds like a good idea. Just a hold over from previous code.

src/material-examples/button-overview/button-overview-example.ts
```diff
@@ -1,6 +1,10 @@
 import {Component} from '@angular/core';
 
-
+/**
+ * @title Basic buttons
+ * @id button-overview
+ * @component ButtonOverviewExample
```
amcdnl: I first tried `require(file)` and looking at the exported names but caused runtime issues. I spoke w/ @DevVersion and we have a plan to do this in typescript parser.

amcdnl:
Per feedback from @DevVersion -

* Resolve every export of a `ts.SourceFile`. There is a `getExportsOfModule` function
* Walk through exports and check if it's a `Component` and if so -> analyze the decorator function call (resolve `templateUrl` and `styleUrls`) (edited)
* Resolve example title and example id from JSDoc (using `getDocumentationComment()`)

tools/gulp/tasks/docs.ts
amcdnl: @jelbourn - Where would you like this file to live?

jelbourn: Actually let's see how small it gets when moving the material modules into their own file; maybe then it will be short enough to leave as a string.

amcdnl: Ya, it ended up pretty small after that.

amcdnl:
@DevVersion @jelbourn - I believe I have addressed the outstanding items. Can you please re-review?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+}
+
+/**
+ * Parse the AST of a file and get metadata about it
+ * 
+ * @param {string} filename 
+ * @param {string} src 
+ * @returns {{ primary: any, seconds: any[] }} 
+ */
+function createMetas(filename: string, src: string): { primaryComponent: any, secondaryComponents: any[] } {
```
jelbourn: `parseExampleMetadata`?

jelbourn: Can we tighten up the `any` use in this function? Or does the ts parser ironically not offer specific types?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+  return {
+    primaryComponent: metas.find(m => m.primary),
+    secondaryComponents: metas.filter(m => !m.primary)
+  };
+}
+
+/**
+ * Creates the examples module and metadata
+ */
+task('build-examples-module', (cb) => {
```
jelbourn: Is the callback necessary if the task is synchronous? If so I'd call it `done`

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+    title: '${metadata.title}',
+    component: ${metadata.component},
+    additionalFiles: ${addlFiles},
+    selectorName: ${selectorName}
+  },
+  `;
+}
+
+/**
+ * Build the list template
```
jelbourn: Elaborate on what "the list template" is?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+  `;
+}
+
+/**
+ * Builds the template for the examples module
+ * 
+ * @param {ExampleMetadata[]} extractedMetadata 
+ * @returns {string} resulting template
+ */
+function buildTemplate(extractedMetadata: ExampleMetadata[]): string {
```
jelbourn: Is this just `generateExampleNgModule`?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+  if (metadata.additionalComponents) {
+    components.push(...metadata.additionalComponents);
+  }
+
+  return `import {${components.join(',')}} from '${metadata.filename}';
+`;
+}
+
+/**
+ * Build the examples template
```
jelbourn: Elaborate on what "the examples template" is?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    for (const match of matches) {
+      const src = fs.readFileSync(match, 'utf-8');
+      const filename = match
+        .replace('src/material-examples/', './')
+        .replace('.ts', '');
+       
+      const { primaryComponent, secondaryComponents } = createMetas(match, src);
+
+      if(primaryComponent) {
```
DevVersion: Nit: Space after the `if` here.

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+        // convert the class name to dashcase id
+        let id = primaryComponent.component.replace('Example', '');
+        id = convertToDashCase(id);
```
DevVersion: You can make this is a const if you just combine the variable initialization.

`convertToDashCase(primaryComponent.component.replace('Example', ''));`

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+          filename,
+          id,
+          component: primaryComponent.component,
+          title: primaryComponent.title,
+          additionalComponents: [],
+          additionalFiles: [],
+          selectorName: []
+        };
+
+        if(secondaryComponents.length) {
```
DevVersion: Space missing here and in the for loop below as well. (Do we have a lint rule for this?)

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,247 @@
...
+  };
+}
+
+/**
+ * Creates the examples module and metadata
+ */
+task('build-examples-module', (cb) => {
+  const results: ExampleMetadata[] = [];
+
+  glob('src/material-examples/**/*.ts', (err, matches) => {
```
DevVersion: To avoid the asynchronous here, just import `glob.sync` (as we do everywhere else)

```ts
import {sync as glob} from 'glob';
```

tools/gulp/tasks/examples.ts
amcdnl: Ya, you need to wait until all operations are completed before we mark the task completed.

tools/gulp/tasks/examples.ts
amcdnl: I try to avoid over complicated function invocation, but happy to change.

tools/gulp/tasks/examples.ts
amcdnl: I added a new interface for the parsed metadata.

I was having type issues with the visit function node, I'll revisit it and see if I can't tighten it up.

amcdnl: @DevVersion - I tinkered with tightening up the types in the `visit` function but had no luck.

I started off w/ `const visit = (node: ts.Node)` then when I hit `ClassDeclaration` I casted to that type, however, `jsDoc` is not a type on that node. I looked up the types and `jsDoc` is actually not a type on anything that extends `ts.Node` (odd?).

Also when I casted it, the `decorator.expression.expression.text` was no longer accessible. I tried the methods called `getText()` like we talked about but all returned `TypeError: Cannot read property 'text' of undefined`.

amcdnl:
@jelbourn @DevVersion - Updated and ready for further review.

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,232 @@
...
+      };
+
+      let primary = false;
+      if(node.jsDoc && node.jsDoc.length) {
+        for(const doc of node.jsDoc) {
+          if(doc.tags && doc.tags.length) {
+            for(const tag of doc.tags) {
+              const tagValue = tag.comment;
+              const tagName = tag.tagName.text;
+              if(tagName === 'title') {
```
jelbourn: Nit: space between keyword and parenthesis (`if`, `for`)
(here and elsewhere)

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,232 @@
...
+  name = name.replace(/[A-Z]/g, ' $&');
+  name = name.toLowerCase().trim();
+  return name.split(' ').join('-')
+}
+
+/**
+ * Parse the AST of a file and get metadata about it
+ */
+function parseExampleMetadata(filename: string, src: string): 
+    { primaryComponent: ParsedMetadata, secondaryComponents: ParsedMetadata[] } {
```
jelbourn: Nit: no leading/trailing space in object literals
(here and elsewhere)

amcdnl:
Ready for another review :)

jelbourn:
`tslint` is now complaining about the generated file. Should add `/* tslint:disable */` at the top of the generated file to avoid this.

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/**
+ * Builds the examples metadata including title, component, etc.
+ */
+function buildExamplesTemplate(metadata: ExampleMetadata): string {
+  // if no additional files or selectors were provided,
+  // return undefined since we don't care about if these were not found
+  const additionalFiles = metadata.additionalFiles ?
+    JSON.stringify(metadata.additionalFiles) : 'undefined';
```
DevVersion: Why not just `'null'` instead? Null looks more explicit than `undefined`. I think that's what @jelbourn also proposed.

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Build the list of components template
+ */
+function buildListTemplate(metadata: ExampleMetadata): string {
+  const components = [metadata.component];
+
+  if (metadata.additionalComponents) {
+    components.push(...metadata.additionalComponents);
+  }
```
DevVersion: This would be the same comment as for the imports. We should be able to drop the if check here.

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,245 @@
...
+task('build-examples-module', (done) => {
+  const results: ExampleMetadata[] = [];
+
+  const matchedFiles = glob(path.join(examplesPath, '**/*.ts'));
+  for (const sourcePath of matchedFiles) {
+    const src = fs.readFileSync(sourcePath, 'utf-8');
+    const { primaryComponent, secondaryComponents } = parseExampleMetadata(sourcePath, src);
+
+    if (primaryComponent) {
+      // convert the class name to dashcase id
```
DevVersion: `// Generate a unique id for the component by converting the class name to dash-case.`

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,245 @@
...
+        id,
+        component: primaryComponent.component,
+        title: primaryComponent.title,
+        additionalComponents: [],
+        additionalFiles: [],
+        selectorName: []
+      };
+
+      if (secondaryComponents.length) {
+        // for whatever reason the primary is listed here
```
DevVersion: This comment doesn't really help?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,245 @@
...
+          example.additionalFiles.push(meta.templateUrl);
+          example.selectorName.push(meta.component);
+        }
+      }
+
+      results.push(example);
+    }
+  }
+
+  const template = generateExampleNgModule(results);
```
DevVersion: `const generatedModuleFile = generateExampleNgModule(results);`

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,245 @@
...
+  return {
+    primaryComponent: metas.find(m => m.primary),
+    secondaryComponents: metas.filter(m => !m.primary)
+  };
+}
+
+/**
+ * Creates the examples module and metadata
+ */
+task('build-examples-module', (done) => {
```
DevVersion: Since we are now using `gulp.sync` we can just get rid of the `done` here.

tools/gulp/tasks/docs.ts
```diff
@@ -1,7 +1,8 @@
 import {task, src, dest} from 'gulp';
 import {Dgeni} from 'dgeni';
-import * as path from 'path';
 import {buildConfig} from 'material2-build-tools';
+import * as fs from 'fs';
```
amcdnl: `fs` is used for `readFileSync` and `writeFileSync`.

DevVersion: As far as I can tell, there is no `readFileSync` and `writeFileSync` in the docs.ts file

amcdnl: It was left over from my code in that file. I will remove.

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,245 @@
...
+const examplesPath = path.join(packagesDir, 'material-examples');
+
+/** Output path of the module that is being created */
+const outputModuleFilename = path.join(examplesPath, 'example-module.ts');
+
+/**
+ * Build ecmascript module import statements
+ */
+function buildImportsTemplate(metadata: ExampleMetadata): string {
+  const components = [metadata.component];
```
amcdnl: Good call, left over from when it wasn't.

tools/gulp/tasks/examples.ts
amcdnl: Ok.

amcdnl: I don't use null very much, happy to change though.

tools/gulp/tasks/examples.ts
amcdnl: More of a reminder for myself, I'll remove.

amcdnl:
@DevVersion @jelbourn  - Ready for another review

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * Creates the examples module and metadata
+ */
+task('build-examples-module', () => {
+  const results: ExampleMetadata[] = [];
+
+  const matchedFiles = glob(path.join(examplesPath, '**/*.ts'));
+  for (const sourcePath of matchedFiles) {
+    const src = fs.readFileSync(sourcePath, 'utf-8');
```
DevVersion: The variable name `src` is pretty unclear. How about `fileContent`/`sourceContent`?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  const metas: any[] = [];
+
+  const visit = (node: any): void => {
+    if (node.kind === ts.SyntaxKind.ClassDeclaration) {
+      const meta: any = {
+        component: node.name.text
+      };
+
+      let primary = false;
```
DevVersion: This variable seems to be unused?

tools/gulp/tasks/examples.ts
```diff
@@ -0,0 +1,234 @@
...
+  return `import {${components.join(',')}} from './${relativeSrcPath}';
+`;
+}
+
+/**
+ * Builds the examples metadata including title, component, etc.
+ */
+function buildExamplesTemplate(metadata: ExampleMetadata): string {
+  // if no additional files or selectors were provided,
+  // return undefined since we don't care about if these were not found
```
DevVersion: Maybe change this comment to `null` now as well (since it's no longer `undefined` now). Maybe something like

```ts
// If there are no additional files or selectors specified, just set those variables
// to an empty array.
const additionalFiles = metadata.additionalFiles.length ? 
   JSON.stringify(metadata.additionalFiles) : '[]';
```

Not sure if an `empty` array would make sense here. It would allow us to drop more `null` checks in the docs app. But `null` is also fine.

amcdnl:
I'm closing this PR and gonna redo it given all the merge issues.

PR #5364, opened 2017-06-26
fix(autocomplete): allow number zero as value

### LINKED ISSUE #5363: AutoComplete value of number 0
When selecting the value of the number 0 from a drop down list of numbers, the value 0 is not set.  A user can manually type in 0.

When selecting the value of the number 0 from a drop down list of numbers, the value 0 is set

When selecting the value of the number 0 from a drop down list of numbers, the value 0 is not set and has a blank value in the input field.

https://plnkr.co/edit/czRXKtRzVyb2WyYFkWyM?p=preview

Select 0. The input will remain empty.

Would like the ability to choose the number 0 as a value.

current angular, current material, tested on windows 10, chrome

DevVersion:
* Currently the autocomplete does not properly display the selected options with the number zero as value.

Fixes #5363.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -358,7 +358,9 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
 
   private _setTriggerValue(value: any): void {
     const toDisplay = this.autocomplete.displayWith ? this.autocomplete.displayWith(value) : value;
-    this._element.nativeElement.value = toDisplay || '';
+    // Simply falling back to an empty string if the display value is falsy does not work properly.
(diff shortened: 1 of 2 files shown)


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -358,7 +358,9 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
 
   private _setTriggerValue(value: any): void {
     const toDisplay = this.autocomplete.displayWith ? this.autocomplete.displayWith(value) : value;
-    this._element.nativeElement.value = toDisplay || '';
+    // Simply falling back to an empty string if the display value is falsy does not work properly.
+    // The display value can also be the number zero and shouldn't fallback to an empty string.
```
kara: Typo: `fallback` -> `fall back`.

PR #5365, opened 2017-06-26
fix(dialog): add config option for aria-describedby 

crisbeto:
Adds an option to the dialog config that allows for the `aria-describedby` to be set on the dialog container.

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -63,5 +63,8 @@ export class MdDialogConfig {
   /** Layout direction for the dialog's content. */
   direction?: Direction = 'ltr';
 
+  /** ID of the element that describes the dialog.  */
+  ariaDescribedBy?: string | null = null;
+
   // TODO(jelbourn): add configuration for lifecycle hooks, ARIA labelling.
 }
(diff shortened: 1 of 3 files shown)


jelbourn:
I'm not sure this is right- what motivated this change?

crisbeto:
I ran into this when testing the dialog with NVDA where it doesn't read out the content at all. [I based it on the example from MDN.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role)

jelbourn:
In MDN's example, the element they're pointing to is literally a description of the dialog, though. In reality, that element will contain all sorts of arbitrary content, so it's not really accurate to say it's the `describedby` for the dialog element. We should support setting `describedby` on the dialog, but not automatically grab the entire content for it. 

AFAIK the content of the dialog would be available just like any other content on the page when navigating with a screenreader. How were you moving through the dialog after opening it?

crisbeto:
I tabbed my way to one of the buttons and opened the dialog. The dialog header got read out correctly, as well as the button, but none of the content. I also tried tabbing and clicking around, but I couldn't get it to read out anything. FWIW, VoiceOver works as it is in master.

jelbourn:
Typically you would use NVDA's navigation to go through text content (vs. jumping between form fields with  tab). 

http://webaim.org/resources/shortcuts/nvda

crisbeto:
You're right, I retried it and it go read out this time. I'll re-purpose the PR to add a config option for `aria-describedby`.

crisbeto:
Done.

PR #5378, opened 2017-06-27
fix(overlay): remove overlay container on destroy

jelbourn:


src/lib/core/overlay/overlay-container.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  }));
+
+  it('should remove the overlay container element from the DOM on destruction', () => {
+    const fixture = TestBed.createComponent(TestComponentWithTemplatePortals);
+
+    const overlayRef = overlay.create();
+    overlayRef.attach(fixture.componentInstance.templatePortal);
+    fixture.detectChanges();
+
+    expect(document.body.querySelector('.cdk-overlay-container'))
```
crisbeto: Does this pass? I know that when I did the cleanup last time, there were still some divs left over between tests because the overlay container didn't have any cleanup logic.

src/lib/core/overlay/overlay-container.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+        .not.toBeNull('Expected the overlay container to be in the DOM after opening an overlay');
+
+    // Manually call `ngOnDestroy` because there is no way to force Angular to destroy an
+    // injectable in a unit test.
+    overlayContainer.ngOnDestroy();
+
+    expect(document.body.querySelector('.cdk-overlay-container'))
```
crisbeto: Can be just `document.querySelector(...)`.

jelbourn:
@crisbeto passed everywhere except browserstack. I changed the test to ensure that all overlay containers are remove with a `beforeAll`. Also got rid of `body.`

mmalerba:
needs rebase

PR #5382, opened 2017-06-27
feat(table): add disconnect to data source

andrewseguin:
Disconnects the data source when it is no longer used, either by the table being destroyed or another data source taking its place

### DIFF
--- a/src/lib/core/data-table/data-source.ts
+++ b/src/lib/core/data-table/data-source.ts
@@ -14,5 +14,5 @@ export interface CollectionViewer {
 
 export abstract class DataSource<T> {
   abstract connect(collectionViewer: CollectionViewer): Observable<T[]>;
-  disconnect() {}
+  abstract disconnect(collectionViewer: CollectionViewer): void;
 }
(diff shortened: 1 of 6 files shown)


src/lib/core/data-table/data-source.ts
```diff
@@ -14,4 +14,5 @@ export interface CollectionViewer {
 
 export abstract class DataSource<T> {
   abstract connect(collectionViewer: CollectionViewer): Observable<T[]>;
+  disconnect() {}
```
jelbourn: Pass the `collectionViewer` to the data source?

andrewseguin: Makes sense to me, but it does require that all data sources need to implement it now. At least a no-op

jelbourn: You can also mark `disconnect` as optional.

tinayuangao:
Please fix the test

tinayuangao:
@andrewseguin Please rebase again. Thanks

andrewseguin:
Rebase

PR #5383, opened 2017-06-27
feat(snack-bar): inject data and MdSnackBarRef into custom snack-bar component

### LINKED ISSUE #5371: MdSnackbar with injectable data
feature request

*MdSnackBar.openFromComponent* should allow specifying data being injected into the component in a similar fashion as it works in *MdDialog*.

*MdSnackBar.openFromComponent* only lets you choose the component to be rendered but the caller has no option how to parameterize the component.

One use-case might be to use the *MdSnackBar* for showing logs. A user might want to have a template for error logs and a different template for info logs. He can use *MdSnackbar.openFromComponent* to choose the template but he cannot inject the actual message into the template component.

reported in @angular/material 2.0.0-beta.7

crisbeto:
* Injects the `MdSnackBarRef` into custom snack bar instances.
* Adds the option to inject arbitrary data into a snack bar.
* Turns the `DialogInjector` into a `PortalInjector` to allow it to be used elsewhere (e.g. the snack-bar).
* Minor tweaks to the `SimpleSnackBar` to get it to use the new DI tokens instead of assigning data directly to the component instance.

Fixes #5371.

### DIFF
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -11,8 +11,6 @@ import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 import {MdSnackBarContainer} from './snack-bar-container';
 
-// TODO(josephperrott): Implement onAction observable.
-
 /**
  * Reference to a snack bar dispatched from the snack bar service.
  */
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -17,12 +17,8 @@ import {MdSnackBarContainer} from './snack-bar-container';
  * Reference to a snack bar dispatched from the snack bar service.
  */
 export class MdSnackBarRef<T> {
-  private _instance: T;
-
   /** The instance of the component making up the content of the snack bar. */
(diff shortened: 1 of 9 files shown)


src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -11,18 +11,12 @@ import {Observable} from 'rxjs/Observable';
...
  * Reference to a snack bar dispatched from the snack bar service.
  */
 export class MdSnackBarRef<T> {
-  private _instance: T;
-
   /** The instance of the component making up the content of the snack bar. */
-  get instance(): T {
-    return this._instance;
-  }
+  instance: T;
```
jelbourn: Add `readonly`

crisbeto: I can't, because the instance gets assigned by the service right after the component is instantiated. This is why I had to remove the getter in the first place. The issue is that we need the snack bar ref to create the injector and we need the inejctor to create the component. Alternatively, I can make the injection tokens on the custom injector public, but that seems even hackier.

jelbourn: Ah, got it

mmalerba:
please rebase

PR #5386, opened 2017-06-28
feat: move a11y, bidi, platform, rxjs, and portal to cdk

jelbourn:
I'm purposefully not updating imports inside `lib/` to point to `@angular/cdk` since we can do that later. This makes it easier to move other stuff to the cdk (since you don't have to then change the import again).

I also screwed up and didn't make these "moves" in git.- if someone has a way to do that retroactively that would be nice.


src/cdk/public_api.ts
```diff
@@ -6,4 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+export * from './a11y/index';
+export * from './bidi/index';
 export * from './coercion/index';
+export * from './platform/index';
+export * from './portal/index';
+export * from './rxjs/index';
+export * from './keyboard/keycodes';
```
DevVersion: For the CDK I wanted the convention to always have an `index.ts` file in the given piece of the CDK.

The index file of the "CDK piece" then always specifies what should be exported.

For the keycodes it might not be necessary, but for the sake of consistency I think it would make sense to do that as well (otherwise it will end up as in core.ts)

src/cdk/a11y/README.md
```diff
@@ -0,0 +1,30 @@
+# LiveAnnouncer
+`LiveAnnouncer` is a service, which announces messages to several screenreaders.
```
mmalerba: s/service/injectable? :p

jelbourn: These will all get revised when we write overview files for stuff in the cdk like the rest of the library.

src/cdk/bidi/index.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {NgModule} from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
+import {Dir} from './dir';
+import {DIR_DOCUMENT, Directionality, DIRECTIONALITY_PROVIDER} from './directionality';
```
tinayuangao: Also export  `DIRECTIONALITY_PROVIDER_FACTORY`? It's used in core/

jelbourn: Added

tinayuangao: Please fix the test? 
Since we have line 14 we can remove line 12

jelbourn: Yep, I clearly was not thinking. Fixed.

PR #5398, opened 2017-06-28
fix(checkbox, radio): setting id to null causes invalid id for input

### LINKED ISSUE #5394: checkbox id attribute cannot be set to null and be ignored
Bug

I'm trying to build a wrapper around the angular material library because there are certain customizations/enhancements that we want to add on top. When writing the wrapper, we pass inputs from the wrapper component to the angular material element. In this case, mdCheckbox. 

When we don't set the value for "id" for our wrapper for mdCheckbox, it sets the ```<input>``` value for id to "input-". This is inconsistent as we don't set the value for "id" for our wrapper for mdSlideToggle and the ```<input>``` value for id is a unique value. Expected behaviour is that the ```<input>``` id value for mdCheckbox should be a unique, just like the implementation of mdSlideToggle.

getting ```<input id="input-">``` but should be unique like "```<input id="input-checkbox-#">```" or something along those lines.

http://plnkr.co/edit/2cvLOW?p=preview 

Inconsistency, hard to wrap components and expose inputs

Using:

Angular 4.0.0, Material 2.0.0-beta.7, MACOSX, TypeScript 2.3.3

probably all browsers

First time submitting bug report, please let me know if I'm missing anything for next time. Thanks!

DevVersion:
* In some situations, developers will change the `id` of a component dynamically. At some point if they set the `[id]` to `null` the input id will look like the **following: `null-input`**.
* If id's are auto-generated they should be reflected in the `id` property of the component (after setting the id to `null` for example)
* If developers are setting the `id` using a binding then the DOM Id for the host component won't be set at all (because the @Input is triggered instead of the DOM property)

**Note**: There will be likely a bit of a payload increase, but it's for the sake of consistency and will be an improvement if we have more components with *id's*

Fixes #5394

### DIFF
--- a/e2e/components/checkbox-e2e.spec.ts
+++ b/e2e/components/checkbox-e2e.spec.ts
@@ -9,7 +9,7 @@ describe('checkbox', () => {
 
     it('should be checked when clicked, and unchecked when clicked again', async () => {
       let checkboxEl = element(by.id('test-checkbox'));
-      let inputEl = element(by.css('input[id=input-test-checkbox]'));
+      let inputEl = element(by.css('input[id=test-checkbox-input]'));
 
       screenshot('start');
       checkboxEl.click();
@@ -32,7 +32,7 @@ describe('checkbox', () => {
     });
 
     it('should toggle the checkbox when pressing space', () => {
-      let inputEl = element(by.css('input[id=input-test-checkbox]'));
+      let inputEl = element(by.css('input[id=test-checkbox-input]'));
 
       expect(inputEl.getAttribute('checked'))
           .toBeFalsy('Expect checkbox "checked" property to be false');
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -266,6 +266,15 @@ describe('MdCheckbox', () => {
 
     it('should preserve the user-provided id', () => {
       expect(checkboxNativeElement.id).toBe('simple-check');
+      expect(inputElement.id).toBe('simple-check-input');
+    });
+
+    it('should generate a unique id for the checkbox input if no id is set', () => {
+      testComponent.checkboxId = null;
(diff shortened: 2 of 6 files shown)


src/lib/core/common-behaviors/id.ts
```diff
@@ -0,0 +1,35 @@
...
+/** @docs-private */
+export interface CanId {
+  id: string;
+}
+
+/**
+ * Mixin to augment a directive with an `id` property.
+ * The id property will fall back to auto-generated unique ids if developers don't specify one.
+ */
+export function mixinId<T extends Constructor<{}>>(base: T, componentName: string)
```
jelbourn: I'm not sure this is a good idea. I'd consider returning a different ID than what the user defines to be surprising / confusing. For components that need an ID, I'd rather throw an error if its set to null.

DevVersion: Hmm I see what you mean. I just had the impression (due to #5394) that it would be helpful to fall back to the auto-generated id if it's set to a falsy value. 

I'm fine changing it to throw an error.

jelbourn: Yeah, my thinking is that if someone wants to wrap a material component and forward through all of the bindings, they need to provide a real ID (and not just leave it empty)

DevVersion: Yeah. Not sure if it would make sense for them to even fall back to id's that include the `md-checkbox` prefix for example.

I think it makes sense to throw an error if the id is set `null`. Anything else before I make the changes?

jelbourn: I don't think so

DevVersion: @jelbourn Done. Please have another look.

mmalerba:
please rebase

DevVersion:
Done.

PR #5400, opened 2017-06-28
Fix for Issue 5220 - aria-labels on md-selects

christandy:
I'm adding a function: _setAriaLabel which adds the selected items to the aria-label on the md-select element. setAriaLabel is called by _onSelect.


christandy:
I signed it!

jelbourn:
I don't think that a `role="listbox"` is supposed to reflect the selection back into its own `aria-label`; the label is supposed to describe what the control is _for_. Do you have some reference for this?

christandy:
For a standard select, you would use aria-activedescendant set to the id of the selected item, but because the options are removed from the DOM when the select isn't being interacted with, a screen reader presumably wouldn't be able to follow the id to the label of the option that has been selected. The way I've implemented it does read in the non-interacting state, without such a drastic code change, but I agree that it's less than ideal. Would you prefer to leave those options in the DOM? If so I'd be happy to try to get another code change in. Another factor might be internationalization considering I've hard-coded the text for the "selection is" and "selected options are"....which probably isn't ideal either. Thanks for the feedback!

jelbourn:
Leaving the options in the DOM is potentially problematic for performance reasons. I'll consult w/ our internal a11y folks for advice

jelbourn:
Also worth mentioning that `activedescendant` communicates which option has _focus_, not for selection. The selection state should be pulled from the `aria-selected` state of the individual options, but again that would still require them to be on the DOM.

jelbourn:
Closing this since it's stale and I don't think we want to go with the proposed approach. The select still definitely has some known a11y issues, which I think will need a different solution.

PR #5405, opened 2017-06-28
fix(select): throwing additional errors if ngModel fails to initialize

### LINKED ISSUE #5402: md-select with ngModel requires name attribute (hasValue error)
This is a bug with the documentation not matching expected behaviour

No error when copying guide code

You get an error:
MdSelect.html:1 ERROR TypeError: Cannot read property 'hasValue' of undefined
at Object.View_MdSelect_0.co [as updateDirectives]

Following the guide, and copying example code should render without errors:
https://material.angular.io/components/select/overview
```
<md-select placeholder="State" [(ngModel)]="myState">
   <md-option *ngFor="let state of states" [value]="state.code">{{ state.name }}</md-option>
</md-select>
```

The error is not useful, also modifying the example code, adding a name attribute to the md-select element fixes the issue:
```
<md-select placeholder="State" [(ngModel)]="myState" name="states">
   <md-option *ngFor="let state of states" [value]="state.code">{{ state.name }}</md-option>
</md-select>
```

Angular 4.0.2
Angular Material 2.0.0-beta.7
MacOS 10.12.5
TypeScript 2.0.0
Chrome 59.0.3071.109

The other example is correct here:
https://material.angular.io/components/select/examples

crisbeto:
If the user has an `md-select` with an `ngModel` that doesn't have a name inside a form, the forms module will throw an error, however Material will also start throwing errors, which may cause confusion. These changes add a null check so our errors don't get mixed up with the forms error.

Fixes #5402.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -63,7 +63,8 @@ describe('MdSelect', () => {
         BasicSelectWithTheming,
         ResetValuesSelect,
         FalsyValueSelect,
-        SelectWithGroups
+        SelectWithGroups,
+        InvalidSelectInForm
       ],
       providers: [
         {provide: OverlayContainer, useFactory: () => {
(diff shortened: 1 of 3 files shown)


src/lib/select/select.spec.ts
```diff
@@ -1942,6 +1943,17 @@ describe('MdSelect', () => {
       }).not.toThrow();
     }));
 
+    it('should not throw selection model-related errors in addition to the errors from ngModel',
+      async(() => {
+        const fixture = TestBed.createComponent(InvalidSelectInForm);
+
+        // The first change detection run will throw the "ngModel is missing a name" error.
+        expect(() => fixture.detectChanges()).toThrow();
```
kara: Can you pass a regex to `toThrow` to check which error is being thrown here?

src/lib/select/select.spec.ts
```diff
@@ -1942,6 +1943,17 @@ describe('MdSelect', () => {
...
+
+        // The secound run shouldn't throw selection-model related errors.
+        expect(() => fixture.detectChanges()).not.toThrow();
```
kara: Typo: `secound` -> `second`

PR #5416, opened 2017-06-29
chore: screenshots task cleanup

DevVersion:
Attempt to make the screenshot task more readable and easier to understand. The screenshot task includes a lot of magic (due to the Firebase functions JWT solution) and we should try our best to make it as readable as possible.

**Note**: This also fixes that the screenshot results are uploaded outside of Gulp (causing gulp to continue with other tasks while this is running)

### DIFF
--- a/tools/gulp/tasks/screenshots.ts
+++ b/tools/gulp/tasks/screenshots.ts
@@ -1,14 +1,25 @@
 import {task} from 'gulp';
 import {readdirSync, statSync, existsSync, mkdirp, readFileSync, writeFileSync} from 'fs-extra';
-import * as path from 'path';
-import * as firebaseAdmin from 'firebase-admin';
-import * as firebase from 'firebase';
-import {
-  openScreenshotsBucket,
-  connectFirebaseScreenshots} from '../util/firebase';
+import {openScreenshotsBucket, connectFirebaseScreenshots} from '../util/firebase';
 import {isTravisMasterBuild} from '../util/travis-ci';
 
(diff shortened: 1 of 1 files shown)


tools/gulp/tasks/screenshots.ts
```diff
@@ -1,15 +1,26 @@
...
-import {
-  openScreenshotsBucket,
-  connectFirebaseScreenshots} from '../util/firebase';
-import {isTravisMasterBuild} from '../util/travis-ci';
 
+// Firebase provides TypeScript definitions that are only accessible from specific namespaces.
+// This means that those types are really long and it's nearly impossible to write a function that
+// doesn't exceed the maximum columns. Import the types from the namespace so they are shorter.
+import Database = firebaseAdmin.database.Database;
+import DataSnapshot = firebaseAdmin.database.DataSnapshot;
```
jelbourn: `FirebaseDatabase` and `FirebaseDataSnapshot`?

DevVersion: Would work as well, but the main intention was to keep it as short as possible.

jelbourn: I aim for explicitness first and brevity second.

DevVersion: Hmm not sure. It should be pretty clear from the file name and from the comment as well.

jelbourn: I don'r care that much for this one

PR #5420, opened 2017-06-29
docs(table): add initial docs

andrewseguin:
Review mostly to check if the example's code looks alright since it'll be copied several times for the other examples (sorting, filtering, etc). Will continue working on the table.md after this


src/material-examples/table-pagination/table-pagination-example.ts
```diff
@@ -0,0 +1,103 @@
...
+import 'rxjs/add/observable/merge';
+import 'rxjs/add/operator/map';
+
+@Component({
+  selector: 'table-pagination-example',
+  styleUrls: ['table-pagination-example.css'],
+  templateUrl: 'table-pagination-example.html',
+})
+export class TablePaginationExample {
+  propertiesToDisplay = ['userId', 'userName', 'progress', 'color'];
```
jelbourn: `selectedColumns`?

src/material-examples/table-pagination/table-pagination-example.html
```diff
@@ -0,0 +1,41 @@
+<div class="example-container mat-elevation-z8">
+  <div class="example-header"> Users </div>
+
+  <cdk-table #table [dataSource]="dataSource">
+
+    <!-- Column Definition: ID -->
```
jelbourn:
```html
<!-- ID column -->
```
etc. ?

src/material-examples/table-pagination/table-pagination-example.html
```diff
@@ -0,0 +1,41 @@
+<div class="example-container mat-elevation-z8">
+  <div class="example-header"> Users </div>
+
+  <cdk-table #table [dataSource]="dataSource">
+
```
jelbourn: Maybe add a comment here like

"Note that these columns can be defined in any order. The actual rendered columns are set as a property on the row definition"

src/lib/core/data-table/table.md
```diff
@@ -0,0 +1,14 @@
+The CDK data-table displays rows of data with fully customizable cell templates.
+Its single responsibility is the efficient rendering of rows in a fully accessible way.
+
+The CDK table provides a foundation upon which other features, such as sorting and pagination,
+can be built. Because the CDK table enforces no opinions on these matters, developers have full
+control over the interaction patterns associated with the table.
+
+For a Material Design styled table, see <md-table>, which builds on top of the CDK Data Table.
```
willshowell: ... see \`\<md-table\>\`, which ...

src/material-examples/table-pagination/table-pagination-example.ts
andrewseguin: Since selection will be used in a later example, it might be confusing if theyre related. How about displayedColumns?

Issue #5035, opened 2017-06-08
Make all components OnPush


jelbourn:
- [x] `MdAutocomplete`
- [x] `MdAnchor`
- [x] `MdButtonToggle`
- [x] `MdOptgroup`
- [x] `MdOption`
- [x] `MdPseudoCheckbox`
- [x] `MdDatepicker`
- [ ] `MdDialogContainer`
- [x] `MdExpansionPanelHeader`
- [x] `MdExpansionPanel`
- [x] `MdGridList`
- [x] `MdGridTile`
- [x] `MdGridTileText`
- [x] `MdInputContainer`
- [x] `MdList`
- [x] `MdListItem`
- [x] `MdMenu`
- [x] `MdMenuItem`
- [x] `MdSpinner`
- [x] `MdSelect`
- [x] `MdSlider`
- [x] `SimpleSnackBar`
- [x] `MdSnackBarContainer`
- [x] `MdTabBody`
- [x] `MdTabGroup`
- [x] `MdTabHeader`
- [x] `MdTab`
- [x] `MdTabNavBar`
- [x] `TooltipComponent`
- [x] `MdSidenavContainer`

We should make these change incrementally, tackling one high-level component at a time (e.g., "tabs"). This will make it easier to revert in case something goes wrong after the change (and also help prevent a rebasing nightmare). 

Couple of things worth noting:
* Most components _should_ be as simple as adding `OnPush` to the config. This is the case if the component just has simple inputs and outputs.
* Anything that implements `ControlValueAccessor` needs `markForCheck` in the forms hooks

alexw10:
@jelbourn just curious for the reasoning for this?

ciesielskico:
@alexw10  OnPush is a performance setting, so the component gets only rerendered by Angular when its inputs and/or outputs change

jelbourn:
Yeah, it helps limit the cases where change detection runs on a component.

fxck:
https://github.com/angular/material2/issues/3005
https://github.com/angular/material2/issues/3106

related onpush issues

z639:
and #5118

e-cloud:
@jelbourn since there are so many component needed to be updated, your team should make the releases of Material2 more frequently, should'n you?

crisbeto:
There are pending and/or merged PRs for all components now, but I'll keep this open until everything is merged and I've set up some linting enforce the `OnPush` change detection.

crisbeto:
Closing since all components have been switched over now.

PR #5422, opened 2017-06-29
chore(dialog): switch to OnPush change detection

crisbeto:
Switches the dialog container to `OnPush` change detection.

Relates to #5035.

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -17,6 +17,8 @@ import {
   EventEmitter,
   Inject,
   Optional,
+  ChangeDetectionStrategy,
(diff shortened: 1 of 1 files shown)


jelbourn:
A google teams' screenshot test is showing that this change causes a dialog's content to not long have data binding resolved when its opened.

crisbeto:
I couldn't really repro it @jelbourn, but I've added a couple of `markForCheck` calls after the content is attached.

jelbourn:
I'll see if that fixes it before diving in further

jelbourn:
Unfortunately that didn't change it. We'll need to investigate the app directly.

PR #5433, opened 2017-06-30
Fix typo which caused CdkHeaderRow to be used instead of CdkRow in MdTable row

swftvsn:


swftvsn:
I signed it!

swftvsn:
@jelbourn This is only a minor typo fix, prolly faster to just change it and commit in. Anyway, testing the new table results now all rows to be underlined due to (atleast I suspect) this.

andrewseguin:
Good catch, thanks for fixing that up. Little concerned that the tests didn't catch this.

PR #5438, opened 2017-06-30
feat: move observe-content to cdk

DevVersion:
* Moves the observe content directive to the CDK package.


jelbourn:
Caretaker note: will need a build rule

PR #5449, opened 2017-06-30
Table docs

andrewseguin:


src/material-examples/cdk-table-basic/cdk-table-basic-example.css
```diff
@@ -0,0 +1,39 @@
...
+  flex-direction: column;
+  max-height: 500px;
+  min-width: 300px;
+}
+
+/*
+ * Styles to make the demo's cdk-table match the material design spec
+ * https://material.io/guidelines/components/data-tables.html
+ */
+.cdk-table {
```
jelbourn: Example styles should all be prefixed with `example-` so people don't think that the styles from the demo are supposed to be there in the component by itself

src/material-examples/cdk-table-basic/ckd-table-basic-example.ts
```diff
@@ -0,0 +1,89 @@
+import {Component} from '@angular/core';
```
jelbourn: This file starts with `ckd`

andrewseguin:
Rebased, did changes

PR #5459, opened 2017-07-01
build: fail build for prerender errors in lifecycle hooks

crisbeto:
* Currently the build won't fail if an error is thrown inside a lifecycle hook, because `platform-server` writes it to STDERR, but still exits with a 0 code. These changes set up the Universal testing task to fail for these cases as well. This should prevent issues like #5455 in the future.
* Stops logging the entire rendered document after the prerender task is done. This makes the log a bit easier to look through.

**Note:** The CI is expected to fail until #5455 is merged.

### DIFF
--- a/tools/gulp/util/task_helpers.ts
+++ b/tools/gulp/util/task_helpers.ts
@@ -50,6 +50,8 @@ export interface ExecTaskOptions {
   errMessage?: string;
   // Environment variables being passed to the child process.
   env?: any;
+  // Whether the task should fail if the process writes to STDERR.
+  failOnStderr?: boolean;
 }
 
 /** Create a task that executes a binary as if from the command line. */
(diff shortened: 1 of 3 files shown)


tools/gulp/util/task_helpers.ts
```diff
@@ -50,33 +50,33 @@ export interface ExecTaskOptions {
...
-        if (options.errMessage === undefined) {
-          done('Process failed with code ' + code);
-        } else {
-          done(options.errMessage);
-        }
-      } else {
-        done();
+    childProcess.stderr.on('data', (data: string) => {
+      if (options.failOnStderr) {
+        done(data);
```
DevVersion: This means that only the first `stderr` message will be printed, but there can be more `stderr` messages.

It would make more sense to exit in the `onClose` hook if there was a `stderr` output.

crisbeto: Done.

PR #5464, opened 2017-07-01
docs(autocomplete): add basic autocomplete examples

amcdnl:
Adds basic set of demos for autocomplete component:

- Basic
- Filtering
- Display Value
- Better overview example


src/material-examples/autocomplete-display/autocomplete-display-example.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  ngOnInit() {
+    this.filteredOptions = this.myControl.valueChanges
+        .startWith(null)
+        .map(user => user && typeof user === 'object' ? user.name : user)
+        .map(name => name ? this.filter(name) : this.options.slice());
+  }
+
+  filter(name: string): User[] {
+    return this.options.filter(option => new RegExp(`^${name}`, 'gi').test(option.name));
```
willshowell: I think the preference has shifted to using `indexOf(...) === 0` but only the overview example ended up getting changed.

See https://github.com/angular/material2/pull/4921#issuecomment-306322183 for context

src/material-examples/autocomplete-filter/autocomplete-filter-example.ts
```diff
@@ -0,0 +1,34 @@
...
+  filteredOptions: Observable<string[]>;
+
+   ngOnInit() {
+      this.filteredOptions = this.myControl.valueChanges
+         .startWith(null)
+         .map(val => val ? this.filter(val) : this.options.slice());
+   }
+
+   filter(val: string): string[] {
+      return this.options.filter(option => new RegExp(`^${val}`, 'gi').test(option));
```
willshowell: Here as well

src/material-examples/autocomplete-display/autocomplete-display-example.ts
rafaelss95: @willshowell extending to this: In #5213 `startsWith` was used. Also note that if replaced by either `indexOf` or `startsWith` it must be used with `toLowerCase`.

jelbourn: Yeah, the issue here is that if someone types RegExp special characters into the input it will cause problems. Simpler to use `indexOf`.

src/material-examples/autocomplete-overview/autocomplete-overview-example.html
```diff
@@ -1,9 +1,21 @@
...
 
-<md-autocomplete #auto="mdAutocomplete">
-  <md-option *ngFor="let state of filteredStates | async" [value]="state">
-    {{ state }}
-  </md-option>
-</md-autocomplete>
\ No newline at end of file
+  <md-autocomplete #auto="mdAutocomplete">
+    <md-option *ngFor="let state of filteredStates | async" [value]="state.name">
+      <img style="vertical-align:middle;" src="{{state.flag}}" height="25" />
```
jelbourn: Probably make this `img` `aria-hidden` since it's a purely decorative addition that doesn't add any semantic meaning to the item.

src/material-examples/autocomplete-display/autocomplete-display-example.html
```diff
@@ -0,0 +1,11 @@
+<form class="example-form">
+  <md-input-container class="example-full-width">
+    <input type="text" placeholder="Assignee" mdInput [formControl]="myControl" [mdAutocomplete]="auto">
```
jelbourn: Input needs an `aria-label`
(for other inputs as well)

amcdnl:
@willshowell @jelbourn - Addressed the items requested, ready for second review.

src/material-examples/autocomplete-display/autocomplete-display-example.ts
```diff
@@ -0,0 +1,44 @@
...
+  ngOnInit() {
+    this.filteredOptions = this.myControl.valueChanges
+        .startWith(null)
+        .map(user => user && typeof user === 'object' ? user.name : user)
+        .map(name => name ? this.filter(name) : this.options.slice());
+  }
+
+  filter(name: string): User[] {
+    return this.options.filter(option =>
+      option.name.toLowerCase().indexOf(name.toLowerCase()) > -1);
```
jelbourn: We should do `=== 0`, since I think screen-reader users expect the autocomplete to work this way

rafaelss95: @jelbourn it can also be done using `startsWith` instead of `indexOf(...) === 0`.

jelbourn: Not in IE11.

rafaelss95: Oh, yes, unfortunately.

jelbourn:
@amcdnl there was just one outstanding comment on this PR

amcdnl:
@jelbourn - missed that, updated per recommendation.

jelbourn:
Now just one lint error

amcdnl:
@jelbourn - Sorry about that. Resolved now.

jelbourn:
I think your latest rebase did not go smoothly

amcdnl:
Due to the merge issues, I have opened a new PR. Its the same code just properly merged. See: https://github.com/angular/material2/pull/5787

PR #5471, opened 2017-07-02
fix(reposition-scroll-strategy): use last calculated position to re-align the overlay

crisbeto:
Uses the element's last calculated position to re-align it, instead of recomputing the position completely. This prevents the overlay from jumping into the viewport when the user scrolls away.

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -150,6 +150,13 @@ export class OverlayRef implements PortalHost {
     }
   }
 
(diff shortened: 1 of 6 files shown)


crisbeto:
Initially I did it with `overlayRef.getState().positionStrategy`, but that looked weird, because it assumes that the scroll strategy will only be used on a connected overlay (which will be the case most of the time, but we don't enforce it). Similarly, we have the `dipose` method which is only relevant for the global strategy, but it has to be implemented everywhere.

crisbeto:
Closing in favor of https://github.com/angular/material2/pull/7805.

PR #5472, opened 2017-07-02
build: move changelog script to gulp task

DevVersion:
* Moves the changelog script to a gulp task that includes better instructions and error reporting.


DevVersion:
Rebased. @jelbourn please review.

PR #5474, opened 2017-07-02
docs(bidirectionality): bidirectionality how-to explanation

### LINKED ISSUE #4387: RTL documentation missing
Breaking of #4367:
####  proposal:
RTL support exists but I could find how to use it in the documentation, I found an issue, linked to a changeset leading to the relevant code.

EladBezalel:
fixes #4387

### DIFF
--- a/guides/bidirectionality.md
+++ b/guides/bidirectionality.md
@@ -1,0 +1,35 @@
+# Angular Material bi-directionality
+
+### Setting a text-direction for your application
+The [`dir` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir)
+is typically applied to `<html>` or `<body>` element of a page. However, it can be used on any
+element to apply a text-direction to a smaller subset of the page.
+
+All Angular Material components automatically reflect the LTR/RTL direction
+of their container. 
+
+### Reading the text-direction in your own components
+`@angular/cdk` provides a `Directionality` injectable that can be used by any component
+in your application. To consume this injectable, you must import `BidiModule`
+from `@angular/cdk`.
+
+`Directionality` provides two useful properties:
+* `value`: the current text direction, either `'ltr'` or `'rtl'`.
+* `change`: an `Observable` that emits whenever the text-direction changes. Note that this only
+captures changes from `dir` attributes _inside the Angular application context_. It will not
+emit for changes to `dir` on `<html>` and `<body>`, as these are assumed to be static.
+
+#### Example
+```ts
+@Component({ /* ... */})
+export class MyCustomComponent {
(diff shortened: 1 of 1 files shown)


parker-mike:
@EladBezalel Can you please elaborate or prepare a plnkr on how to change the direction of the document from the code, since when I import `Directionality` out of `@angular/cdk` and inject it into the constructor of the app component, I can read the value, but when I set it to "rtl" nothing happens, not even the `change` is being emitted.

guides/bidirectionality.md
```diff
@@ -0,0 +1,22 @@
+# Angular Material bi-directionality
+
+### How to set your application direction
+The simplest way is applying `dir` attribute on either the `html` or `body` tags.
```
jelbourn: I would put this as more...
`````markdown
### Setting a text-direction for your application
The [`dir` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir)
is typically applied to `<html>` or `<body>` element of a page. However, it can be used on any
element to apply a text-direction to a smaller subset of the page.

All Angular Material components automatically reflect the LTR/RTL direction
of their container. 

### Reading the text-direction in your own components
`@angular/cdk` provides a `Directionality` injectable that can be used by any component
in your application. To consume this injectable, you must import `BidiModule`
from `@angular/cdk`.

`Directionality` provides two useful properties:
* `value`: the current text direction, either `'ltr'` or `'rtl'`.
* `change`: an `Observable` that emits whenever the text-direction changes. Note that this only
captures changes from `dir` attributes _inside the Angular application context_. It will not
emit for changes to `dir` on `<html>` and `<body>`, as these are assumed to be static.

#### Example
```ts
@Component({ /* ... */})
export class MyCustomComponent {
  private dir: Direction;

  constructor(directionality: Directionality) {
    this.dir = directionality.value;
    directionality.change.subscribe(() => {
       this.dir = directionality.value;
    });
  }
}
```
`````

parker-mike: @jelbourn 
> It will not emit for changes to `dir` on `<html>` and `<body>`, as these are assumed to be static.

 I'm not sure that this is a correct assumption to make, since overlayed items(`div.cdk-overlay-container`) are generated outside the Angular application context, so we don't have a single point to change the layout of the application. Or may be I have something missed?

jelbourn: This describes the behavior right now. It could potentially change in a subsequent iteration.

parker-mike: @jelbourn So, perhaps it could be a bright idea to specify that this is something to be revised. Should I open a feature request for this?

PR #5475, opened 2017-07-02
feat(overlay): allow a function to be passed as the connected position offset

crisbeto:
* Adds the ability to pass in a function to the `offsetX` and `offsetY` of the `ConnectedPositionStrategy`, which makes it possible to choose a different offset, depending on the position that is being tried. This is useful for cases like the nested menus where the offset can be different, depending on whether the overlay is above or below the trigger.
* Moves a couple of tests out into their appropriate spec.


jelbourn:
Do you think it would make more sense to just include the offsets in the individual positions at config-time?

crisbeto:
That would work as well, but this should be slightly more flexible since it allows you to change the offset after the position strategy is defined.

jelbourn:
@crisbeto do you want to move forward with this change or defer to the new position strategy?

crisbeto:
It not necessary anymore after we added the ability to have different offsets for the fallback positions.

PR #5483, opened 2017-07-03
feat(menu): add injection token for overriding the default options

### LINKED ISSUE #5479: MdMenu add an ability to configure options globally for all components of overlapTrigger=true|false in provider.
Proposal.

It will be great to have ability of configuration of md-menu overlapping in providers. For example, as in the placeholder:
`@NgModule({
  providers: [
    {provide: MD_PLACEHOLDER_GLOBAL_OPTIONS, useValue: { float: 'always' }}
  ]
})`

Currently it`s possible only to insert in every place of html using [overlapTrigger]="false" attribute or default value is 'true'.
 
This feature will add more flexibility of configuration md-menu globally.

crisbeto:
Adds a new injection token that allows the consumer to override the default values for the `xPosition`, `yPosition` and `overlapTrigger` options.

Fixes #5479.

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -36,7 +36,7 @@ import {merge} from 'rxjs/observable/merge';
 import {Observable} from 'rxjs/Observable';
 import {Direction} from '../core';
 
-/** Default `md-menu` options that can be overriden. */
+/** Default `md-menu` options that can be overridden. */
 export interface MdMenuDefaultOptions {
   xPosition: MenuPositionX;
   yPosition: MenuPositionY;
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -20,6 +20,8 @@ import {
   ViewEncapsulation,
   ElementRef,
   ChangeDetectionStrategy,
+  InjectionToken,
+  Inject,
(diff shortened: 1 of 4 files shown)


src/lib/menu/menu-directive.ts
```diff
@@ -29,6 +31,16 @@ import {Subscription} from 'rxjs/Subscription';
 import {transformMenu, fadeInItems} from './menu-animations';
 import {ESCAPE} from '../core/keyboard/keycodes';
 
+/** Default `md-menu` options that can be overriden. */
```
kara: Typo: `overriden` -> `overridden`

src/lib/menu/menu.spec.ts
```diff
@@ -487,6 +489,29 @@ describe('MdMenu', () => {
...
+      imports: [MdMenuModule, NoopAnimationsModule],
+      declarations: [SimpleMenu],
+      providers: [{
+        provide: MD_MENU_DEFAULT_OPTIONS,
+        useValue: {overlapTrigger: false, xPosition: 'before', yPosition: 'above'},
+      }],
+    }).compileComponents();
+  }));
+
+  it('should allow for the default menu options to be overriden', () => {
```
kara: Typo: `overriden` -> `overridden`

Issue #5067, opened 2017-06-11
Feature request: custom view value for select and autocomplete


xuanshenbo:
#### Bug, feature request, or proposal:
Feature request

#### What is the expected behavior?
Select and Autocomplete allow custom view values, for example:
https://material.io/guidelines/components/menus.html#menus-usage
<img width="570" alt="capture" src="https://user-images.githubusercontent.com/7393393/27009691-c4544504-4ee8-11e7-8292-aa1adb775200.PNG">

This Menu allows icons in addition to text.
I understand the example is a Menu not Select or Autocomplete, but they are under the same category in Material Design 

#### What is the current behavior?
The values are limited to strings, and doesn't allow custom layout inside the Select and Autocomplete dropdown.

#### What is the use-case or motivation for changing an existing behavior?
As per the example screenshot.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Material 2.0.0-beta.6

julianobrasil:
Have you try something like:
```
<div>
    text to the left
    <div style="float: right">text to the right</div>
</div>
```
In the autocomplete md-option? It works with text. Maybe it works with icons to (material fonts of fontawesome).

julianobrasil:
Take a look at this plunker: https://plnkr.co/edit/v3EoZp?p=preview

![image](https://user-images.githubusercontent.com/25405143/27010236-0046b0ce-4e76-11e7-9808-d2da0048a15b.png)

xuanshenbo:
@julianobrasil that looks great thanks!

However I accidentally noticed that your code doesn't work on Firefox 53.0.3

Can you raise a bug for that?

Thanks!

julianobrasil:
I think it's an issue with nested divs in Firefox. I modified the plunker. This will work both in Chrome, Firefox, Edge and IE:

```
<div>
      <div style="float: left;">{{ state }}</div>
      <div style="float: right;">
            <i class="material-icons" style="font-size: 15px">add_location</i>
            {{state}}
     </div>  
</div>
```

jelbourn:
Closing this since I believe it's doable today (it would just be your mat-option template). Free free to open another issue if that's not the case.

PR #5493, opened 2017-07-04
feat(menu): add nested menu functionality

### LINKED ISSUE #1429: Feature Request: Tiered option for Menu Component

Ability to create tiered menu with Menu component

When a user hovers over a specified element, a sub men is opened next to it. Similar to Chromes menu

None exist

I would like the ability to created a tiered menu instead of having to create multiple menus side by side. 

2.0.0-alpha.9

crisbeto:
Adds the ability for an `md-menu-item` inside of a `md-menu` to trigger another `md-menu`. This is a first step towards a `md-toolbar` component.

Demo: https://material-cb7ec.firebaseapp.com/menu

Fixes #1429.

### DIFF
--- a/src/lib/menu/menu-item.ts
+++ b/src/lib/menu/menu-item.ts
@@ -6,9 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Component, ElementRef, ChangeDetectionStrategy} from '@angular/core';
+import {Component, ElementRef, OnDestroy, ChangeDetectionStrategy} from '@angular/core';
 import {Focusable} from '../core/a11y/focus-key-manager';
 import {CanDisable, mixinDisabled} from '../core/common-behaviors/disabled';
+import {Subject} from 'rxjs/Subject';
 
 // Boilerplate for applying mixins to MdMenuItem.
 /** @docs-private */
@@ -26,16 +27,23 @@ export const _MdMenuItemMixinBase = mixinDisabled(MdMenuItemBase);
   host: {
     'role': 'menuitem',
     'class': 'mat-menu-item',
+    '[class.mat-menu-item-highlighted]': '_highlighted',
     '[attr.tabindex]': '_getTabIndex()',
     '[attr.aria-disabled]': 'disabled.toString()',
     '[attr.disabled]': 'disabled || null',
     '(click)': '_checkDisabled($event)',
+    '(mouseenter)': '_emitHoverEvent()',
   },
   changeDetection: ChangeDetectionStrategy.OnPush,
   templateUrl: 'menu-item.html',
   exportAs: 'mdMenuItem',
 })
-export class MdMenuItem extends _MdMenuItemMixinBase implements Focusable, CanDisable {
+export class MdMenuItem extends _MdMenuItemMixinBase implements Focusable, CanDisable, OnDestroy {
+  /** Stream that emits when the menu item is hovered. */
+  hover: Subject<MdMenuItem> = new Subject();
+
+  /** Whether the menu item is highlighted. */
+  _highlighted: boolean = false;
 
   constructor(private _elementRef: ElementRef) {
     super();
@@ -44,6 +52,10 @@ export class MdMenuItem extends _MdMenuItemMixinBase implements Focusable, CanDi
   /** Focuses the menu item. */
   focus(): void {
     this._getHostElement().focus();
+  }
+
+  ngOnDestroy() {
+    this.hover.complete();
   }
 
   /** Used to set the `tabindex`. */
@@ -63,5 +75,13 @@ export class MdMenuItem extends _MdMenuItemMixinBase implements Focusable, CanDi
       event.stopPropagation();
     }
   }
+
+  /** Emits to the hover stream. */
+  _emitHoverEvent() {
+    if (!this.disabled) {
+      this.hover.next(this);
+    }
+  }
+
 }
 
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -132,7 +132,15 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
 
   ngAfterViewInit() {
     this._checkMenu();
-    this.menu.close.subscribe(() => this.closeMenu());
+
+    this.menu.close.subscribe(reason => {
+      this.closeMenu();
+
+      // If a click closed the menu, we should close the entire chain of nested menus.
+      if (reason === 'click' && this._parentMenu) {
+        this._parentMenu.close.emit(reason);
+      }
+    });
 
     if (this.triggersSubmenu()) {
       // Subscribe to changes in the hovered item in order to toggle the panel.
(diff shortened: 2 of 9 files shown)


m98:
1.  Would not be good if we could show a simple arrow (to right) for nested menus? Like user should not guess that does this menu item has nested item or not, and do this continuously for the whole list items.
screenshot from 2017-07-05 13-43-43

2. The animation seems to run very fast after mouse-in, a short delay would make it feel better for users, it's not good to animate too fast (I mean on-mouse-in on list items) because user is just mouse hovering the menu to find the item he is looking for, and the fast opening menu (without delay) is disturbing.
([See this short video](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0B08MbvYZK1iNWVBad29DaVJrREk/components-menus-menus-cascading_dropdown_spec_large_xhdpi.mp4))

joejordanbrown:
We need to make sure that when using a mouse and running the cursor diagnally that it doesn't change to another nested menu item.

I haven't seen this mentioned in the design document, so I'm not sure if this has been overlooked at the moment.

The google documents menu does this but i beleive its done using a timer instead of detecting direction, a good example would be the amazon menu even if you move slowly it still has the correct nested menu open and moving the cursor up and down the menu shows the nested items without delay. This seems to be a lot more user friendly.

image

This page [here](http://bjk5.com/post/44698559168/breaking-down-amazons-mega-dropdown) documents it.

jelbourn:
The article @joejordanbrown posted makes a pretty good point; we should consider adding a feature like that to the cdk and using in the menu as a follow-up.

crisbeto:
Makes sense, I'll add it to my backlog.

src/lib/menu/menu-trigger.ts
```diff
@@ -255,48 +274,108 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
       this.menu.yPosition === 'above' ? ['bottom', 'top'] : ['top', 'bottom'];
 
-    let originY = overlayY;
-    let fallbackOriginY = fallbackOverlayY;
+    let [originY, originFallbackY] = [overlayY, overlayFallbackY];
+    let [overlayX, overlayFallbackX] = [originX, originFallbackX];
+    let offsetY = 0;
 
-    if (!this.menu.overlapTrigger) {
+    if (this.triggersSubmenu) {
```
jelbourn: Can you add some comments around how the positioning is different when triggering a submenu?

src/lib/menu/menu-item.ts
```diff
@@ -26,15 +27,22 @@ export const _MdMenuItemMixinBase = mixinDisabled(MdMenuItemBase);
...
     '(click)': '_checkDisabled($event)',
+    '(mouseenter)': '_emitHoverEvent()',
   },
   templateUrl: 'menu-item.html',
   exportAs: 'mdMenuItem'
 })
-export class MdMenuItem extends _MdMenuItemMixinBase implements Focusable, CanDisable {
+export class MdMenuItem extends _MdMenuItemMixinBase implements Focusable, CanDisable, OnDestroy {
+  /** Stream that emits when the menu item is hovered. */
+  hover: Subject<MdMenuItem> = new Subject();
```
crisbeto: It is mainly to facilitate the nested menu, but I can see it being useful in other cases as well.

crisbeto:
I'll add some UX improvements in a follow-up. This PR is mainly for figuring out the positioning, accessibility and opening/closing logic.

crisbeto:

Splaktar:
I agree with @m98 that the animation speed should be slowed down from what the demo has. It should match what the first level menu uses.

xuanshenbo:
wrt adding an indicator, could be related to #5067

src/lib/menu/menu-trigger.ts
```diff
@@ -260,48 +282,110 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
+        this.menu.close.emit();
+      }
+    }));
+
+    // Subscribe to the parent menu closing, allowing for the close event
+    // to propagate to any child menus that the current trigger may have.
+    this._parentMenuSubscriptions.push(this._parentMenu.close.subscribe(() => {
+      this.menu.close.emit();
+    }));
+  }
```
kara: Is it at all possible to avoid having a list of subscriptions? I think the logic might be neater / easier to follow if we merge the streams that result in the menu closing into one subscription, rather than having a few separate subscriptions that do the same thing.

```ts
get _menuClosingActions() {
   return merge(
      this._overlayRef.backdropClick,
      this._parentMenu.close,
      this._parentMenu.hover().filter(active => active !== this._menuItemInstance);
}
```

Then instead of only subscribing to the backdrop when the menu opens, we subscribe to the menu closing events when the menu opens, kind of like the autocomplete.

```ts
_subscribeToMenuClose() {
   this._closingSubscription = this._menuClosingActions.subscribe(() => this.menu.close.emit());
}
```

Ultimately, we'd still be adding another subscription onInit for the parent menu hovering (to open the menu when the active item is this menu item), but at least we wouldn't be adding an array of subscriptions that we'd need to manage.

crisbeto: I did it initially, because they subscribe at different times in the lifecycle. Looking at it again, they don't really need to do that. Refactored.

crisbeto:
Rebased and addressed the feedback. Also I agree on the UX improvements, but I'll add them in a follow-up. The main purpose of this PR was to sort out the opening/closing logic, accessibility and positioning. Can you take another look @kara?

src/lib/menu/menu-trigger.ts
```diff
@@ -285,48 +306,96 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
+      this._hoverSubscription
+    ]
+    .filter(subscription => !!subscription)
+    .forEach(subscription => subscription.unsubscribe());
+  }
+
+  /** Returns a stream that emits whenever an action that should close the menu occurs. */
+  private _menuClosingActions() {
+    const backdrop = this._overlayRef!.backdropClick();
+    const parentClose = !this._parentMenu ? observableOf(null) : this._parentMenu.close;
```
kara: Nit: I'd switch the logic here and on line 364 because it's harder to scan "if not, else" statements (the else becomes "if not not").

```ts
this._parentMenu ?  this._parentMenu.close : observableOf(null);
```

crisbeto:
Good point WRT to closing all of the menu on click @kara. I've addressed it together with the other comment.

kara:
Note: Addition of `isSubmenu` to the interface is a breaking change and will need to be updated internally before merge.

crisbeto:
I suppose the same can be said about the `direction` and the signature change of `close`. Should I scope these changes to the `MdMenu` since they're there to facilitate the nested menus?

kara:
@crisbeto No changes needed on your end; just a note to myself so I don't accidentally merge prematurely.

kara:
@crisbeto Actually, chatted with @jelbourn and making `isSubmenu` optional would be even better.

crisbeto:
Done.

PR #5495, opened 2017-07-04
fix(card): unable to bind to align attribute

### LINKED ISSUE #5490: md-card-actions align property not known when binding
Bug

the align property from md-card-actions should be bindable.

console is throwing error, cannot use variable to assign value to align property.

https://plnkr.co/edit/qnEsY21BVoL2WCOFrTTu?p=preview

Not working as expected. 

Angular 4.0.0, Material 2.0.0-beta.7, MACOSX, TypeScript 2.3.3

crisbeto:
Fixes not being able to bind to the `align` attribute on the `md-card-actions`.

**Note:** If we're okay with a breaking change, this may be simpler to do with a CSS class.

Fixes #5490.

### DIFF
--- a/src/lib/card/card.scss
+++ b/src/lib/card/card.scss
@@ -43,11 +43,11 @@ $mat-card-header-size: 40px !default;
   margin-left: -16px;
   margin-right: -16px;
   padding: 8px 0;
+}
 
-  &[align='end'] {
-    display: flex;
-    justify-content: flex-end;
-  }
+.mat-card-actions-align-end {
+  display: flex;
+  justify-content: flex-end;
 }
 
 .mat-card-image {
--- a/src/lib/card/card.ts
+++ b/src/lib/card/card.ts
@@ -10,7 +10,8 @@ import {
   Component,
   ViewEncapsulation,
   ChangeDetectionStrategy,
-  Directive
+  Directive,
+  Input,
 } from '@angular/core';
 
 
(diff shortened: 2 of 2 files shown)


jelbourn:
I'm in general interested in changing behaviors like this to use just css classes, but it's something we'd have to do for the whole library. Any thoughts?

crisbeto:
I prefer the CSS classes as well since these bindings just add unnecessary bloat. I can do it in the few other places that we have something similar, but it'll be a breaking change.

jelbourn:
Might be good to first compile a list of places where we _could_ do this and see how major the change would be

crisbeto:
@jelbourn here's a list where changes would be necessary: 
* `align` on `md-card`.
* `align` on `md-dialog-actions`.
* `dense` on `md-list`.
* `fullscreen` on `md-sidenav`.
* `md-stretch-tabs` on `md-tabs`.

There's also the `mode` input on `md-progress-bar` and `md-progress-spinner`, but I think that we should keep it there since the components need to react on mode changes. Should I go ahead with the list from above?

jelbourn:
Let me see how widely they're used inside Google first

jelbourn:
About:
* 6 uses on md-dialog-actions
* 32 for md-card-actions
* 20 for md-list
* I would want to skip sidenav anyway since we're working on changing that separately
* Tabs is hard to tell because the same property exists in material1

It would be pretty time consuming to update these, so updating the binding is how I would go for now. If we run out of more important stuff to do before end-of-beta then we can revisit.

PR #5510, opened 2017-07-05
fix(select): bigger height for underline on focus

### LINKED ISSUE #5499: Underline of md-select is only 1px high when focused instead of 2px like md-input etc
#### Bug, feature request, or proposal:b
Bug

Underline of select on focus should be 2px high

Underline is currently 1px high unlike other fields like md-input

image

image

Add a select from the docs to a project, select an option, and the focused fields underline is 1px.

http://plnkr.co/edit/Oy0TpVeCi4jxSSPGeV6c?p=preview

Consistency

Angular 4.2.5, Angular Material 2.0.0-beta.7, Windows 10

DevVersion:
* On focus the underline height should be as twice as big as the normal height. This is similar to the input component of Angular Material.
* Replaces the `aria-disabled` attribute selectors with the `mat-select-disabled` class.

Fixes #5499

### DIFF
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -10,6 +10,7 @@ $mat-select-trigger-min-width: 112px !default;
 $mat-select-arrow-size: 5px !default;
 $mat-select-arrow-margin: 4px !default;
 $mat-select-panel-max-height: 256px !default;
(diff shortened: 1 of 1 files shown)


mmalerba:
caretaker note: screenshot diffs need to be approved

PR #5513, opened 2017-07-05
docs(sort): add initial sorting docs

andrewseguin:


src/lib/sort/sort.md
```diff
@@ -0,0 +1,19 @@
+To add sorting behavior and styling to a set of table headers, add the `<md-sort-header>` component
```
jelbourn: The overview should be structured like
1. Single sentence that summarizes the components, e.g.: "The `mdSort` and `md-sort-header` are used, respectively, to add sorting state and display to tabular data."
2. overview example
3. Basic use
4. Additional sections on configuration, alternate behaviors, etc.
5. Accessibility notes for this component

src/material-examples/sort-overview/sort-overview-example.css
```diff
@@ -0,0 +1,3 @@
+.mat-sort-header-container {
```
jelbourn: There should be an additional example of the sort directives being applied to an `md-table`.
(we can re-use the same example for both the table and sortable pages)

src/material-examples/sort-overview/sort-overview-example.ts
```diff
@@ -0,0 +1,50 @@
...
+        case 'calories': [propertyA, propertyB] = [a.calories, b.calories]; break;
+        case 'fat': [propertyA, propertyB] = [a.fat, b.fat]; break;
+        case 'carbs': [propertyA, propertyB] = [a.carbs, b.carbs]; break;
+        case 'protein': [propertyA, propertyB] = [a.protein, b.protein]; break;
+      }
+
+      let valueA = isNaN(+propertyA) ? propertyA : +propertyA;
+      let valueB = isNaN(+propertyB) ? propertyB : +propertyB;
+
+      return (valueA < valueB ? -1 : 1) * (sort.direction == 'asc' ? 1 : -1);
```
jelbourn: I feel like this sort function is more complicated than needed for this example. Instead of baking all of the columns into the comparator, just have a `compareNumber` and `compareString` function and use the appropriate one based on the selected column?

src/material-examples/sort-overview/sort-overview-example.css
andrewseguin: Added some message about it; can add example when table-docs PR is merged

src/lib/sort/sort.md
```diff
@@ -0,0 +1,31 @@
+The `mdSort` and `md-sort-header` are used, respectively, to add sorting behavior, state, and display
+to tabular data.
```
jelbourn: I wouldn't say "behavior", since the components don't actually do any sorting. The directives only capture the state and the display.

src/lib/sort/sort.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Changing the sort order
+
+By default, a sort header starts its sorting at `asc` and then `desc`. Triggering the sort header
+after `desc` will remove sorting.
+
+To reverse the sort order for all headers, set the `mdSortStart` to `desc` on the `mdSort` directive.
+To reverse the order only for a specific header, set the `start` input only on the header instead.
+
+To prevent the user from removing sort altogether after it is applied, set `mdSortDisableClear` to
```
jelbourn: "To prevent the user from clearing the sort state from an already sorted column"

src/lib/sort/sort.md
```diff
@@ -0,0 +1,31 @@
+The `mdSort` and `md-sort-header` are used, respectively, to add sorting behavior, state, and display
+to tabular data.
+
+<!-- example(sort-overview) -->
+
+# Adding sort to table headers
```
jelbourn: Start with h3 (`###`) and go from there

andrewseguin:
Changes made, ready for another review

PR #5523, opened 2017-07-05
fix(slide-toggle): slide-toggle module depends on forms module

DevVersion:
* Currently the slide-toggle module always imports the `FormsModule` and therefore when treeshaking the `FormsModule` would be always included (even if FormsModule is not used at all)

**Note**: I'm working on similar PRs for the *button-toggle*  and *slider* (need to check first)

### DIFF
--- a/src/lib/slide-toggle/index.ts
+++ b/src/lib/slide-toggle/index.ts
@@ -7,7 +7,6 @@
  */
 
 import {NgModule} from '@angular/core';
-import {FormsModule} from '@angular/forms';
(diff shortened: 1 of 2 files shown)


jelbourn:
@kara should we do this? Do you think components that are "forms compatible" should generally avoid bringing in the module?

DevVersion:
@jelbourn Why should they depend on the `FormsModule` if they are really designed to work independently from the forms module? 

The checkbox and radio are not depending on the `FormsModule` as well.

Issue #1728, opened 2016-11-04
Navigation drawer - Mini variant behavior


Daugoh:
#### Feature request:
Add the "Mini variant" behavior of the Navigation drawer.
As defined in the Google Material guidelines:
https://material.google.com/patterns/navigation-drawer.html#navigation-drawer-behavior

![](https://cloud.githubusercontent.com/assets/1202936/8023412/a6d190f4-0ce1-11e5-86b9-55448c4ae6bd.png)

#### What is the expected behavior?
When toggled, the navigation drawer change its width.
It's not fully hidden.

#### What is the current behavior?
Only two states are available: displayed, hidden

#### What are the steps to reproduce?
Create a sidenav and toggle it on and off.
The sidenav can only be opened or closed.
No intermediate state.

#### What is the use-case or motivation for changing an existing behavior?
Desktop users need to quickly change the current view.
The sliding effect of the sidenav could be annoying when used a lot.
It's common for desktop apps to keep the navigation opened by default.
But in some cases the icons are self explanatory so no need to display the associated labels.

#### Which versions of Angular, Material, OS, browsers are affected?
Angular2

#### Is there anything else we should know?
A similar feature request was submitted for Angular 1:
https://github.com/angular/material/issues/3158

jelbourn:
Is there any reason you couldn't control the width of the sidenav with css and toggle classes based on which width you want to display?

Daugoh:
Hello @jelbourn 

Thanks for your answer.

Yes I tried to do it manually.
I can show/hide the texts with a CSS class or use a ngIf.
It works but then I have a problem of alignment for my icons.
They are not correctly centered.

So yes we can do it ourselves but I think it would be really useful to have it natively.
And then just use an attribute like textVisible or iconOnly for instance ...

Regards

kuncevic:
Is that feature has being planned yet? Is there any deadlines on that?

kuncevic:
Is that feature has being planned yet? Is there any deadlines on that?

@daugoh have you come up with any proper soulution for this one?

As mentioned here https://github.com/angular/material2/issues/2021 with a small showcase http://plnkr.co/edit/LcRDIHNFjHKLFTaMdNEM?p=preview that just gives an idea how to keep a bit of sidenav showed when it is closed but that is no proper way of hadle the icons in material html structure sort of way.

That is how material how `md-nav-list` output html is looks like inside `md-sidenav`:
```
<md-nav-list _ngcontent-wlc-55="" role="list">
[...]

PR #5528, opened 2017-07-06
Navigation drawer - Mini variant behavior #1728

daxsom:
Quick attempt at mini variant. Adds third state of 'collapsed'. Width can be set in css: 

.mat-sidenav.mat-sidenav-collapsed {
    min-width: 164px;
    width: 164px;
}


daxsom:
cla

PR #5429, opened 2017-06-30
docs: add temporary docs for cdk

jelbourn:
* Update getting started guide WRT cdk
* Add some temporary docs for cdk to be viewed on Github until we can integrate them into material.angular.io
* Small change to docs generation to output ck docs (but not do anything with them yet)

@mmalerba or @kara please review

PR #5536, opened 2017-07-06
build: dgeni is building docs twice

DevVersion:
* With #5429 dgeni runs twice when running the `gulp docs` command. This is because the dgeni generate function has been added to the `dgeni/index.js` file.


jelbourn:
Whoops, added that for debugging and forgot to remove it.

PR #5547, opened 2017-07-06
tests(checkbox, slide-toggle): use fakeAsync instead of whenStable

DevVersion:
* Most of the `whenStable()` promise calls can be replaced with the `flushMicrotasks` method, which just flushes them instead of waiting for them to really happen (when using `async`).
* This reduces nesting of functions and also makes the tests in theory faster.


DevVersion:
@jelbourn I will look for more instances like this in other components as well. This really makes tests more readable I feel like

PR #5549, opened 2017-07-06
chore(expansion-panel): switch to OnPush change detection

crisbeto:
* Switches the expansion panel to `OnPush` change detection.
* Fixes the open animation not triggering on the body element.

Relates to #5035.

### DIFF
--- a/src/lib/expansion/accordion-item.ts
+++ b/src/lib/expansion/accordion-item.ts
@@ -6,7 +6,15 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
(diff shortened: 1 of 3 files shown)


src/lib/expansion/accordion-item.ts
```diff
@@ -44,6 +52,8 @@ export class AccordionItem implements OnDestroy {
       } else {
         this.closed.emit();
       }
+
+      this._changeDetectorRef.markForCheck();
```
jelbourn: Add a comment that mentions what sets `expanded` outside of the normal `@Input` binding?

PR #5551, opened 2017-07-06
chore(slider): switch to OnPush change detection

crisbeto:
* Switches `md-slider` to `OnPush` change detection.
* Fixes a few properties not reacting to external changes.

Relates to #5035.

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -16,7 +16,9 @@ import {
   Optional,
   Output,
   Renderer2,
-  ViewEncapsulation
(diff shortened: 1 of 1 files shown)


crisbeto:
It's because they modify some unrelated properties that are reflected in the view, like the percentage. We could get away with not doing it for `vertical`, `invert` and `tickInterval`, but I did it to be on the safe side. Should I switch them back?

jelbourn:
It's fine, I was just curious since it's not immediately obvious in the setter. Might be good to add a comment.

PR #5562, opened 2017-07-06
feat(selection-list): Selection-list initial version all code(without demo)

sllethe:


src/lib/list/list-option.html
```diff
@@ -0,0 +1,11 @@
+<div [ngClass]="(checkboxPosition == 'before')?'mat-list-item-content' : 'mat-list-item-content-reverse'" >
+  <!--<div class="mat-list-item-content">-->
```
jelbourn: Commented out code

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
     return this._element.nativeElement;
   }
 }
+
+@Component({
```
jelbourn: This class needs a JsDoc description

src/lib/list/list-option.html
```diff
@@ -0,0 +1,11 @@
+<div [ngClass]="(checkboxPosition == 'before')?'mat-list-item-content' : 'mat-list-item-content-reverse'" >
```
jelbourn: Instead of using `ngClass` and applying one of two classes, you should be able to have the default styles always applied and then apply the "reverse" styles when the condition is met  like this:
```html
<div [class.mat-list-option-reverse]="checkboxPosition == 'after'">
```

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+  get disableRipple() { return this._disableRipple; }
+  set disableRipple(value: boolean) { this._disableRipple = coerceBooleanProperty(value); }
+
+  @ContentChildren(MdLine) _lines: QueryList<MdLine>;
+
+  /** Whether the label should appear after or before the checkbox. Defaults to 'after' */
+
+  @Input() checkboxPosition: 'before' | 'after' = 'after';
+
+  @ViewChild('autocheckbox') pCheckbox;
```
jelbourn: I don't think you need this `@ViewChild` (and all associated code), since the state of the pseudo-checkbox should be set via binding in the template

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+
+  constructor(private _renderer: Renderer2,
+              private _element: ElementRef,
+              @Optional() private _slist: MdSelectionList,
+              @Optional() navList: MdNavListCssMatStyler,
```
jelbourn: Why does this deal with a nav-list? `MdListOption` should only be used in a selection list. Getting rid of this should let you also get rid of `_isNavList` and `_isSelectionList`

(should be assumed that this is always inside of a selection list)

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+
+  /** Whether the label should appear after or before the checkbox. Defaults to 'after' */
+
+  @Input() checkboxPosition: 'before' | 'after' = 'after';
+
+  @ViewChild('autocheckbox') pCheckbox;
+
+  constructor(private _renderer: Renderer2,
+              private _element: ElementRef,
+              @Optional() private _slist: MdSelectionList,
```
jelbourn: `_slist` -> `_selectionList`

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+    this._isSelectionList = !!selectionListStyler;
+  }
+
+
+  ngAfterContentInit() {
+    this._lineSetter = new MdLineSetter(this._lines, this._renderer, this._element);
+  }
+
+  onchange(): void {
+    console.log('checked or not: ' + this.pCheckbox.state + ', isSelected or not: ' + this.isSelected);
```
jelbourn: Need to remove `console.log`
(here and elsewhere)

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+  },
+  templateUrl: 'list-option.html',
+  encapsulation: ViewEncapsulation.None
+})
+export class MdListOption implements AfterContentInit {
+  private _lineSetter: MdLineSetter;
+  private _disableRipple: boolean = false;
+  private _isNavList: boolean = false;
+  private _isSelectionList: boolean = false;
+  isSelected: boolean = false;
```
jelbourn: `isSelected` -> `selected`

It should also be an `@Input()` (that uses `coerceBooleanProperty`)

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+    this._isNavList = !!navList;
+    this._isSelectionList = !!selectionListStyler;
+  }
+
+
+  ngAfterContentInit() {
+    this._lineSetter = new MdLineSetter(this._lines, this._renderer, this._element);
+  }
+
+  onchange(): void {
```
jelbourn: `onchange` -> `toggle`

It's generally better to name functions for what they do rather than based on how they are called. The exception is when the function behavior directly depends on specific event information (such as looking at which key was pressed)

The function also needs a JsDoc description

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+    console.log('checked or not: ' + this.pCheckbox.state + ', isSelected or not: ' + this.isSelected);
+    if(this.isSelected == false) {
+      this.isSelected = true;
+      this.selectionList.checkedItems.select(this._element.nativeElement);
+    }else {
+      this.isSelected = false;
+
+      this.selectionList.checkedItems.deselect(this._element.nativeElement);
+    }
```
jelbourn: With the earlier suggestion about `isSelected`, this function could be simplified to calling `toggle` on the selection-list's `SelectionModel` and setting `this.selected = !this.selected`

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+    console.log(this.selectionList.checkedItems);
+    console.log('current selectionModule: ' + this.selectionList.checkedItems.selected.length);
+  }
+
+  onKeydown(e: KeyboardEvent): void {
```
jelbourn: Rather than handing the keyboard interaction directly, we have an existing utility called `ListKeyManager` that should do a lot of this for you. You should be able to look at the `md-select` for an example of how it's used.

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+    '(click)': 'onchange()',
+    '(keydown)':'onKeydown($event)',
+    '[tabIndex]': 'disabled ? -1 : 0',
+    '[attr.aria-selected]': '',
+
+  },
+  templateUrl: 'list-option.html',
+  encapsulation: ViewEncapsulation.None
+})
+export class MdListOption implements AfterContentInit {
```
jelbourn: Each option also needs a `value` and a `disabled` `@Input()` property (the latter can be applied via the existing disabled mixin).

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+
+  ngAfterContentInit() {
+    this._lineSetter = new MdLineSetter(this._lines, this._renderer, this._element);
+  }
+
+  onchange(): void {
+    console.log('checked or not: ' + this.pCheckbox.state + ', isSelected or not: ' + this.isSelected);
+    if(this.isSelected == false) {
+      this.isSelected = true;
+      this.selectionList.checkedItems.select(this._element.nativeElement);
```
jelbourn: Rather than using the `HTMLElement` as the selected value, it should use either the option itself (`this`).

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+    return this._element.nativeElement;
+  }
+}
+
+@Directive({
+  selector: 'md-selection-list, mat-selection-list',
+})
+export class MdSelectionListCheckboxer {
+
+  checkedItems: SelectionModel<HTMLElement> = new SelectionModel<HTMLElement>(true);
```
jelbourn: `checkedItems` -> `_selectedOptions`

This property should also live directly on `MdSelectionList` (you shouldn't need this class at all)

src/lib/list/list.ts
```diff
@@ -50,6 +53,26 @@ export class MdList {
...
 
+@Component({
+  moduleId: module.id,
+  selector: 'md-selection-list, mat-selection-list',
+  host: {'role': 'listbox'},
+  template: '<ng-content></ng-content>',
+  styleUrls: ['list.css'],
+  encapsulation: ViewEncapsulation.None
+})
+export class MdSelectionList {
```
jelbourn: In another PR after this one, we'll need to add support for the `MdSelectionList` being a form control, similar to the `md-select`

src/lib/list/list.ts
```diff
@@ -70,6 +93,12 @@ export class MdListCssMatStyler {}
 })
 export class MdNavListCssMatStyler {}
 
+@Directive({
+  selector: 'md-selection-list, mat-selection-list',
+  host: {'class': 'mat-selection-list'}
+})
+export class MdSelectionListCssMatStyler {}
```
jelbourn: You shouldn't need this class; you can add the `mat-selection-list` class in `MdSelectionList`

src/lib/list/list.ts
```diff
@@ -175,3 +204,122 @@ export class MdListItem implements AfterContentInit {
...
+  selector: 'md-list-option',
+  host: {
+    'role': 'option',
+    'class': 'mat-list-item',
+    '(focus)': '_handleFocus()',
+    '(blur)': '_handleBlur()',
+    '(click)': 'onchange()',
+    '(keydown)':'onKeydown($event)',
+    '[tabIndex]': 'disabled ? -1 : 0',
+    '[attr.aria-selected]': '',
```
jelbourn: Missing binding expression for `aria-selected`. The expression needs to resolve to a _string_ of either `'true'` or `'false'` (see how existing `MdOption` does this).

You'll also need a binding for `aria-disabled` based on the disabled state

src/lib/list/list-option.html
sllethe: Changed '[ngClass]' to '<div [class.mat-list-option-reverse]="checkboxPosition == 'after'">'

src/lib/list/list.ts
sllethe: Deleted 'MdSelectionListCssMatStyler' class.

src/lib/list/list.ts
sllethe: Added doc for 'md-selection-option'

src/lib/list/list.ts
sllethe: Added binding expression for 'attr.aria-selected' and 'attr.aria-disabled'

src/lib/list/list.ts
sllethe: Added a value and a disabled @Input() property

src/lib/list/list.ts
sllethe: add 'Tab' moves focus on selection-list, 'UP_ARROW' and 'DOWN_ARROW' moves focus within selection-list. And use 'SPACE' to select checkbox

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+ * Component for list-options of selection-list. Each list-option can automatically
+ * generate a checkbox and can put current item into the selectionModel of selection-list
+ * if the current item is checked.
+ */
+@Component({
+  moduleId: module.id,
+  selector: 'md-list-option',
+  host: {
+    'role': 'option',
+    'class': 'mat-list-item',
```
jelbourn: If you add `mat-list-option` to this as well it will be easier to write styles for the option specifically. For example, instead of `.mat-selection-list .mat-list-item` you could just have `mat-list-option`.

src/lib/list/list.scss
```diff
@@ -43,6 +43,19 @@ $mat-dense-list-icon-size: 20px;
...
+  .mat-list-item-content-reverse {
+    display: flex;
+    //flex-direction: row;
+    align-items: center;
+    box-sizing: border-box;
+    height: $base-height;
+    padding: 0 $mat-list-side-padding;
+    position: relative;
+    flex-direction: row-reverse;
+    -webkit-justify-content: space-around; /* Safari */
```
jelbourn: You shouldn't need the `-webkit-justify-content`, since Safari 9 and 10 both support all of the flex properties unprefixed.

src/lib/list/list.scss
```diff
@@ -43,6 +43,19 @@ $mat-dense-list-icon-size: 20px;
     position: relative;
   }
 
+  .mat-list-item-content-reverse {
+    display: flex;
+    //flex-direction: row;
```
jelbourn: Commented line

src/lib/list/list-option.html
```diff
@@ -0,0 +1,10 @@
+<div class="mat-list-item-content"  [class.mat-list-item-content-reverse]="checkboxPosition == 'after'">
+  <div class="mat-list-item-ripple" md-ripple
+       [mdRippleTrigger]="_getHostElement()"
+       [mdRippleDisabled]="!isRippleEnabled()">
+  </div>
+
+  <md-pseudo-checkbox [state]="selected ? 'checked' : 'unchecked'" #autocheckbox [disabled]="disabled"></md-pseudo-checkbox>
+  <div class="mat-list-text"><ng-content></ng-content></div>
+
+</div>
```
jelbourn: Can you configure your IDE to add newlines at the end of files?

src/lib/list/list.scss
```diff
@@ -60,18 +73,18 @@ $mat-dense-list-icon-size: 20px;
     height: $avatar-height;
   }
 
-  &.mat-2-line .mat-list-item-content {
+  &.mat-2-line .mat-list-item-content .mat-list-item-content-reverse {
```
jelbourn: You shouldn't need to add `mat-list-item-content-reverse` to these existing styles; the `mat-list-item-content-reverse` class should only have the extra styles you need to reverse the display

src/lib/list/list.scss
```diff
@@ -182,3 +195,31 @@ $mat-dense-list-icon-size: 20px;
     }
   }
 }
+
+.mat-selection-list {
+  a {
```
jelbourn: There shouldn't be an `<a>` elements inside of a selection list

src/lib/list/list.scss
```diff
@@ -182,3 +195,31 @@ $mat-dense-list-icon-size: 20px;
...
+    text-decoration: none;
+    color: inherit;
+  }
+
+  .mat-list-item-content {
```
jelbourn: I don't think you need these rules for `mat-list-item-content` and `mat-list-item-content-reverse` here

src/lib/list/list.scss
```diff
@@ -182,3 +195,31 @@ $mat-dense-list-icon-size: 20px;
...
+  .mat-list-item-content-reverse {
+    cursor: pointer;
+
+    &:hover, &.mat-list-item-focus {
+      outline: none;
+    }
+  }
+}
+
+.mat-selection-list-checkbox {
```
jelbourn: This class isn't used anywhere

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+}
+
+/**
+ * Component for list-options of selection-list. Each list-option can automatically
+ * generate a checkbox and can put current item into the selectionModel of selection-list
+ * if the current item is checked.
+ */
+@Component({
+  moduleId: module.id,
+  selector: 'md-list-option',
```
jelbourn: Also needs `mat-list-option`

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+    '[attr.aria-selected]': 'selected.toString()',
+    '[attr.aria-disabled]': 'disabled.toString()',
+  },
+  templateUrl: 'list-option.html',
+  encapsulation: ViewEncapsulation.None
+})
+export class MdListOption implements AfterContentInit, OnDestroy, Focusable {
+  private _lineSetter: MdLineSetter;
+  private _disableRipple: boolean = false;
+  private _isSelectionList: boolean = false;
```
jelbourn: You don't need `_isSelectionList` since these options can only appear in a selection list.

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+
+  @ContentChildren(MdLine) _lines: QueryList<MdLine>;
+
+  /** Whether the label should appear after or before the checkbox. Defaults to 'after' */
+
+  @Input() checkboxPosition: 'before' | 'after' = 'after';
+
+  /** Whether the checkbox is disabled. */
+  get disabled() {
+    return this._disabled;
```
jelbourn: The getter should also take the disabled state of the selection-list into account (similar to the existing `MdOption`)

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+  }
+  @Input('disabled')
```
jelbourn: You don't need to pass a string to `@Input()` if it matches the name of the property.

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+  toggle(): void {
+    if(this._disabled == false) {
+      this._selected = !this._selected;
+      if(this._selected == true) {
+        this.selectionList.selectedOptions.select(this);
+      }else {
+        this.selectionList.selectedOptions.deselect(this);
+      }
+    }
+  }
```
jelbourn: The toggle function by itself can just be
```ts
toggle() {
  this.selected = !this.selected;
  this.selectionList.selectedOption.toggle(this);
}
```

The logic that a disabled option should do anything on click should be specifically tied to the click handler rather than being in the more general `toggle()`

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+}
+
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-selection-list, mat-selection-list',
+  host: {'role': 'listbox',
+    '[attr.tabindex]': '_tabIndex',
+    'class': 'mat-selection-list',
+    // Events
```
jelbourn: No need for category comments inside `host`; the different syntax makes it clear enough

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+  _getHostElement(): HTMLElement {
+    return this._element.nativeElement;
+  }
+}
+
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-selection-list, mat-selection-list',
+  host: {'role': 'listbox',
```
jelbourn: Format like
```ts
host: {
  'role': 'listbox',
  ...
}
```

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+  protected _selectable: boolean = true;
+
+  /** The FocusKeyManager which handles focus. */
+  _keyManager: FocusKeyManager;
+
+  /** The option components contained within this selection-list. */
+  options: QueryList<MdListOption>;
+
+  // options which are selected.
+  selectedOptions: SelectionModel<any> = new SelectionModel<any>(true);
```
jelbourn: `SelectionModel<MdListOption>`

src/lib/list/list.ts
```diff
@@ -50,6 +63,347 @@ export class MdList {
...
+        case SPACE:
+          // If we are selectable, toggle the focused option
+          if (this.selectable) {
+            this._toggleSelectOnFocusedOption();
+          }
+
+          // Always prevent space from scrolling the page since the list has focus
+          event.preventDefault();
+          break;
+        case LEFT_ARROW:
```
jelbourn: The selection-list shouldn't need shortcuts for left/right

src/lib/list/list.ts
sllethe: Added `.mat-list-option` class in css files

src/lib/list/list.ts
sllethe: Deleted unused `_isSelectionList`

src/lib/list/list.ts
sllethe: Changed the getter of `disabled` like what in `MdOption`.

src/lib/list/list.ts
sllethe: :+1:

src/lib/list/list.ts
sllethe: :+1:  Great!

src/lib/list/list.ts
sllethe: :+1:

src/lib/list/list.ts
sllethe: :+1:

src/lib/list/list.ts
sllethe: :+1:

src/lib/list/list.ts
sllethe: Deleted shortcuts for left/right

package-lock.json
```diff
@@ -305,6 +305,26 @@
       "integrity": "sha1-dO6TaJbaXkU8NBzygEQ0LBqazrc=",
       "dev": true
     },
+    "@gulp-sourcemaps/identity-map": {
```
tinayuangao: Revert this file?

src/lib/list/list.scss
```diff
@@ -182,3 +214,8 @@ $mat-dense-list-icon-size: 20px;
     }
   }
 }
+
+.mat-selection-list {
```
tinayuangao: remove?

src/lib/list/list.ts
```diff
@@ -159,7 +501,8 @@ export class MdListItem implements AfterContentInit {
 
   /** Whether this list item should show a ripple effect when clicked.  */
   isRippleEnabled() {
-    return !this.disableRipple && this._isNavList && !this._list.disableRipple;
+    return !this.disableRipple && (this._isNavList)
```
tinayuangao: revert this change?

src/lib/list/list.ts
```diff
@@ -7,19 +7,32 @@
...
+  ChangeDetectionStrategy,
+  OnDestroy,
+  EventEmitter,
+  Output
 } from '@angular/core';
-import {coerceBooleanProperty, MdLine, MdLineSetter} from '../core';
+import {CommonModule} from '@angular/common';
+import {coerceBooleanProperty, MdLine, MdLineSetter, MdPseudoCheckbox, MdSelectionModule, SelectionModel} from '../core';
+import {FocusKeyManager} from '../core/a11y/focus-key-manager';
+import {MdCheckbox} from '../checkbox';
```
tinayuangao: Remove unused imports

src/lib/list/list.ts
```diff
@@ -50,6 +63,335 @@ export class MdList {
...
+    'class': 'mat-list-item, mat-list-option',
+    '(focus)': '_handleFocus()',
+    '(blur)': '_handleBlur()',
+    '(click)': 'toggle()',
+    'tabindex': '-1',
+    '[attr.aria-selected]': 'selected.toString()',
+    '[attr.aria-disabled]': 'disabled.toString()',
+  },
+  templateUrl: 'list-option.html',
+  encapsulation: ViewEncapsulation.None
```
tinayuangao: Can we also make this `changeDetection: ChangeDetectionStrategy.OnPush`?

src/lib/list/list.ts
```diff
@@ -50,6 +63,335 @@ export class MdList {
...
+  /** Whether or not the option is selectable. */
+  protected _selectable: boolean = true;
+
+  /** The FocusKeyManager which handles focus. */
+  _keyManager: FocusKeyManager;
+
+  /** The option components contained within this selection-list. */
+  options: QueryList<MdListOption>;
+
+  // options which are selected.
```
tinayuangao: nit: use /** */

src/lib/list/list.ts
```diff
@@ -50,6 +63,335 @@ export class MdList {
...
+  selectedOptions: SelectionModel<MdListOption> = new SelectionModel<MdListOption>(true);
+
+  /**
+   * Whether the ripple effect should be disabled on the list-items or not.
+   * This flag only has an effect for `md-nav-list` components.
```
tinayuangao: `mat-selection-list`?

src/lib/list/list.ts
```diff
@@ -50,6 +63,335 @@ export class MdList {
...
+   */
+  @Input()
+  get disableRipple() { return this._disableRipple; }
+  set disableRipple(value: boolean) { this._disableRipple = coerceBooleanProperty(value); }
+
+  // Whether the selection-list is disabled
```
tinayuangao: nit: Use /**

src/lib/list/list.ts
```diff
@@ -50,6 +63,335 @@ export class MdList {
...
+    this._tabOutSubscription = this._keyManager.tabOut.subscribe(() => {
+      this._tabIndex = -1;
+      setTimeout(() => this._tabIndex = 0);
+    });
+
+    // Go ahead and subscribe all of the initial options
+    this._subscribeOptions(this.options);
+
+    // When the list changes, re-subscribe
+    this.options.changes.subscribe((options: QueryList<MdListOption>) => {
```
tinayuangao: Also unsubscribe this in `ngOnDestroy`?

src/lib/list/list.ts
```diff
@@ -50,6 +63,335 @@ export class MdList {
...
+
+    // If they are on a option or a selection-list, check for space/left/right,
+    // otherwise pass to our key manager
+    if (target && (target.classList.contains('mat-selection-list') ||
+      target.classList.contains('mat-list-item') ||
+      target.classList.contains('mat-list-option'))) {
+      switch (event.keyCode) {
+        case SPACE:
+          // If we are selectable, toggle the focused option
+          if (this.selectable) {
```
tinayuangao: You can remove this `if` since you check `this.selectable` in `_toggleSelectOnFocusedOption`

src/lib/list/list.ts
sllethe: Yes~ :+1: it's should be `mat-selection-list`

src/lib/list/list.ts
sllethe: :+1:

src/lib/list/list.ts
sllethe: Added unsubscribe for 'this.options.changes.subscribe((options: QueryList<MdListOption>)' in onDistory();

src/lib/list/list.ts
sllethe: :+1:  Removed` if (this.selectable)`  check

src/lib/list/list.ts
sllethe: reverted changes

src/lib/list/list.ts
```diff
@@ -50,6 +61,341 @@ export class MdList {
...
+  /** Whether the label should appear after or before the checkbox. Defaults to 'after' */
+
+  @Input() checkboxPosition: 'before' | 'after' = 'after';
+
+  /** Whether the checkbox is disabled. */
+  @Input()
+  get disabled() { return (this.selectionList && this.selectionList.disabled) || this._disabled; }
+  set disabled(value: any) { this._disabled = coerceBooleanProperty(value); }
+
+  @Input('value')
```
tinayuangao: You can use `@Input()` here

src/lib/list/list.ts
```diff
@@ -50,6 +61,341 @@ export class MdList {
...
+  get value() { return this._value; }
+  set value( val: any) { this._value = coerceBooleanProperty(val); }
+
+  @Input('selected')
```
tinayuangao: same here

src/lib/list/list.ts
```diff
@@ -50,6 +61,341 @@ export class MdList {
...
+  }
+
+  ngOnDestroy(): void {
+    this.destroy.emit({option: this});
+  }
+
+  toggle(): void {
+    if(this._disabled == false) {
+      this.selected = !this.selected;
+      this._changeDetector.markForCheck();
```
tinayuangao: Why do you need `markForCheck` here?

src/lib/list/list.ts
```diff
@@ -50,6 +61,341 @@ export class MdList {
...
+  }
+
+  /** Whether this list item should show a ripple effect when clicked.  */
+  isRippleEnabled() {
+    return !this.disableRipple && !this.selectionList.disableRipple;
+  }
+
+  _handleFocus() {
+    this._hasFocus = true;
+    this._renderer.addClass(this._element.nativeElement, 'mat-list-item-focus');
```
tinayuangao: nit: make `mat-list-item-focus` a const?

sllethe: Added const 'FOCUSED_STYLE'

src/lib/list/list.ts
```diff
@@ -50,6 +61,344 @@ export class MdList {
...
+
+  /** Passes relevant key presses to our key manager. */
+  keydown(event: KeyboardEvent) {
+    let target = event.target as HTMLElement;
+
+    // If they are on a option or a selection-list, check for space/left/right,
+    // otherwise pass to our key manager
+    if (target && (target.classList.contains('mat-selection-list') ||
+      target.classList.contains('mat-list-item') ||
+      target.classList.contains('mat-list-option'))) {
```
jelbourn: Why do you need this check on the target? If this component receives a keydown, don't you always want to handle it?

src/lib/list/list.ts
```diff
@@ -50,6 +61,344 @@ export class MdList {
...
+
+  /** Toggles the selected state of the currently focused option. */
+  protected _toggleSelectOnFocusedOption(): void {
+    if (!this.selectable) {
+      return;
+    }
+
+    let focusedIndex = this._keyManager.activeItemIndex;
+
+    if (typeof focusedIndex === 'number' && this._isValidIndex(focusedIndex)) {
```
jelbourn: Why do you need this `typeof` check?

src/lib/list/list.ts
sllethe: Deleted check on the target.

src/lib/list/list.ts
sllethe: Because `focusedIndex = this._keyManager.activeItemIndex` and the type of `keyManager.activeItemIndex` is '`number | null`'. If I do not check `typeof`, it will run out an error.

tinayuangao: Check whether focusedIndex is null?

package-lock.json
tinayuangao: Revert?

src/lib/list/list.ts
```diff
@@ -50,6 +61,334 @@ export class MdList {
...
+  moduleId: module.id,
+  selector: 'md-selection-list, mat-selection-list',
+  host: {
+    'role': 'listbox',
+    '[attr.tabindex]': '_tabIndex',
+    'class': 'mat-selection-list',
+    '(focus)': 'focus()',
+    '(keydown)': 'keydown($event)'},
+  queries: {
+    options: new ContentChildren(MdListOption)
```
jelbourn: Why do you add `queries` here instead of using `@ContentChildren`?

src/lib/list/list.ts
jelbourn: Yeah, checking for `== null` is more typical

src/lib/list/list.ts
```diff
@@ -50,6 +61,334 @@ export class MdList {
...
+  },
+  template: '<ng-content></ng-content>',
+  styleUrls: ['list.css'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MdSelectionList implements AfterContentInit, OnDestroy {
```
jelbourn: Can `md-selection-list` and `md-list-option` go in separate files? (`selection-list.ts` and `list-option.ts`)

src/lib/list/list.ts
sllethe: :+1:  Changed to check `'focusedIndex != null'`

src/lib/list/list.ts
sllethe: :+1:  made `md-selection-list` and `md-list-option` go in separate files? (`selection-list.ts` and `list-option.ts`)

src/lib/list/list.ts
sllethe: Changed to use `@ContentChildren`

sllethe:
Comments Addressed :+1: 
Please take another look.

src/lib/list/list.ts
sllethe: Moved the position of `markForCheck` to the end of this function

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-selection-list, mat-selection-list',
+  host: {
+    'role': 'listbox',
+    '[attr.tabindex]': '_tabIndex',
+    'class': 'mat-selection-list',
+    '(focus)': 'focus()',
+    '(keydown)': 'keydown($event)'},
```
jelbourn: This needs a binding for `aria-disabled` to reflect its disabled state.

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+import {SPACE} from '../core/keyboard/keycodes';
+import {Focusable} from '../core/a11y/focus-key-manager';
+import {MdListOption} from './list-option';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-selection-list, mat-selection-list',
+  host: {
+    'role': 'listbox',
+    '[attr.tabindex]': '_tabIndex',
```
jelbourn: The `tabindex` should be `-1` when the selection-list is disabled

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+   * This flag only has an effect for `mat-selection-list` components.
+   */
+  @Input()
+  get disableRipple() { return this._disableRipple; }
+  set disableRipple(value: boolean) { this._disableRipple = coerceBooleanProperty(value); }
+
+  /** Whether the selection-list is disabled */
+  @Input()
+  get disabled() { return this._disabled; }
+  set disabled(value: any) { this._disabled = coerceBooleanProperty(value); }
```
jelbourn: Instead of defining the `disabled` getter/setter yourself, you can include it via mixin. Check out `button.ts` for an example of how to apply a mixin (there's some boilerplate involved but it ultimately reduces code duplication).

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+  constructor(private _element: ElementRef) { }
+
+  ngAfterContentInit(): void {
+    this._keyManager = new FocusKeyManager(this.options).withWrap();
+
+    // Prevents the selection-list from capturing focus and redirecting
+    // it back to the first option when the user tabs out.
+    this._tabOutSubscription = this._keyManager.tabOut.subscribe(() => {
+      this._tabIndex = -1;
+      setTimeout(() => this._tabIndex = 0);
```
jelbourn: I don't understand why you're doing this with tabindex. AFAIK, the `tabindex` of the selection list should always be `0` if it is enabled and `-1` if it is disabled. The individual options should all have `tabindex="-1"`

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+  _tabIndex = 0;
+
+  /** Track which options we're listening to for focus/destruction. */
+  private _subscribed: WeakMap<MdListOption, boolean> = new WeakMap();
+
+  /** Subscription to tabbing out from the selection-list. */
+  private _tabOutSubscription: Subscription;
+
+  /** Subscription to option changes from the selection-list. */
+  private _optionSubscription: Subscription;
```
jelbourn: Maybe `_optionsChangeSubscription`?

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  /**
+   * Whether or not this option is selectable. When a option is not selectable,
+   * it's selected state is always ignored.
+   */
+  @Input()
+  get selectable(): boolean { return this._selectable; }
+  set selectable(value: boolean) {
+    this._selectable = coerceBooleanProperty(value);
+  }
```
jelbourn: I think this `selectable` property is in the wrong class, but also not necessary for `option` and should probably be removed.

src/lib/list/list.scss
```diff
@@ -139,10 +147,20 @@ $mat-dense-list-icon-size: 20px;
       $mat-list-icon-size
     );
   }
+
+  .mat-list-option {
+    @include mat-line-base(
```
kara: Do you mean `mat-list-item-base` here? The `mat-line-base` mixin only takes one arg.

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Component for list-options of selection-list. Each list-option can automatically
+ * generate a checkbox and can put current item into the selectionModel of selection-list
+ * if the current item is checked.
+ */
+@Component({
+  moduleId: module.id,
+  selector: 'md-list-option, mat-list-option',
+  host: {
+    'role': 'option',
+    'class': 'mat-list-item, mat-list-option',
```
kara: I think the comma is a typo? Don't want a comma between class names.

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,154 @@
...
+  /** Whether the checkbox is disabled. */
+  private _disabled: boolean = false;
+  private _value: any;
+
+  /** Whether the option has focus. */
+  _hasFocus: boolean = false;
+
+  /**
+   * Whether the ripple effect on click should be disabled. This applies only to list items that are
+   * part of a nav list. The value of `disableRipple` on the `md-nav-list` overrides this flag.
```
kara: List-options are only on selection-lists, right? The comment talks about nav lists, so it looks like this might need to be updated?

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    if (this._optionSubscription) {
+      this._optionSubscription.unsubscribe();
+    }
+  }
+
+  /**
+   * Whether or not this option is selectable. When a option is not selectable,
+   * it's selected state is always ignored.
```
kara: Typo: `it's` -> `its`

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+    // If we've already been subscribed to a parent, do nothing
+    if (this._subscribed.has(option)) {
+      return;
+    }
+
+    // Watch for focus events outside of the keyboard navigation
+    option.onFocus.subscribe(() => {
+      let optionIndex: number = this.options.toArray().indexOf(option);
+
+      if (this._isValidIndex(optionIndex)) {
```
kara: Why is the validity of the index necessary to check?

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,228 @@
...
+   * Add a specific option to our subscribed list. If the option has
+   * already been subscribed, this ensures it is only subscribed
+   * once.
+   *
+   * @param option The option to be subscribed (or checked for existing
+   * subscription).
+   */
+  protected _addOption(option: MdListOption) {
+    // If we've already been subscribed to a parent, do nothing
+    if (this._subscribed.has(option)) {
```
kara: Instead of keeping a list of options that were previously subscribed, it might be easier to merge all the option subscriptions into one stream. Then you can just switchMap to a new merged stream every time the option list changes. Check out how we do this in the autocomplete [here](https://github.com/angular/material2/blob/master/src/lib/autocomplete/autocomplete-trigger.ts#L234) and [here](https://github.com/angular/material2/blob/master/src/lib/autocomplete/autocomplete-trigger.ts#L384).

sllethe: :+1: Finally make it work. 
@kara @tinayuangao Thank you so much for your help~~! :1st_place_medal:

package-lock.json
sllethe: :+1: Reverted~

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,192 @@
...
+import {CanDisable, mixinDisabled} from '../core/common-behaviors/disabled';
+import 'rxjs/add/observable/merge';
+import 'rxjs/add/observable/fromEvent';
+import 'rxjs/add/operator/filter';
+import 'rxjs/add/operator/switchMap';
+import 'rxjs/add/observable/of';
+import {merge} from 'rxjs/operator/merge';
+import {Observable} from 'rxjs/Observable';
+import 'rxjs/add/operator/switchMap';
+import 'rxjs/add/operator/startWith';
```
jelbourn: The main components can't use the rxjs `/add/` imports (and there should be a lint failure for doing so). You can use `RxChain` instead

It looks like you have some other lint errors as well; you can see them locally with `gulp tslint`

sllethe: :+1:  Changed to use '`RxChain`' instead of directly importing `'observer/add/`'Changed to use 'RxChain' instead of directly importing 'observer/add/'

sllethe:
Comments Addressed. Added unit tests for selection-list and list-option.
Please take another look :+1:

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+import {UP_ARROW, DOWN_ARROW, SPACE} from '../core/keyboard/keycodes';
+
+
+describe('SelectionList and ListOption', () => {
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [MdListModule],
+      declarations: [
+        SelectionListWithListOptions,
+        SelectionListWithCheckbocPositionAfter,
```
tinayuangao: typo: checkbox

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  it('should add and remove focus class on focus/blur', () => {
+    let fixture = TestBed.createComponent(SelectionListWithListOptions);
+    fixture.detectChanges();
+    let listOption = fixture.debugElement.query(By.directive(MdListOption));
+    let listItemEl = fixture.debugElement.query(By.css('.mat-list-item'));
+
+    expect(listItemEl.nativeElement.classList).not.toContain('mat-list-item-focus');
+
+    listOption.componentInstance._handleFocus();
```
tinayuangao: Also test on native element's focus & blur?

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+    let listOption = fixture.debugElement.queryAll(By.directive(MdListOption));
+    let selectionList = fixture.debugElement.query(By.directive(MdSelectionList));
+    let testListItem = listOption[2].injector.get<MdListOption>(MdListOption);
+
+    testListItem.toggle();
+    fixture.detectChanges();
+
+    let selectList = selectionList.injector.get<MdSelectionList>(MdSelectionList).selectedOptions;
+
+    expect(selectList.selected.length).toBe(1);
```
tinayuangao: nit: also check expect(selectList.selected.length).toBe(0); before the toggle

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+
+
+});
+
+@Component({template: `
+  <mat-selection-list id="selection-list-1">
+    <md-list-option checkboxPosition="before" disabled="true">
+      Inbox (disabled selection-option)
+    </md-list-option>
+    <md-list-option id = "testSelect" checkboxPosition="before">
```
tinayuangao: nit: no space around `=`

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,156 @@
...
+  set selected( val: boolean) { this._selected = coerceBooleanProperty(val); }
+
+  /** Emitted when the option is focused. */
+  onFocus = new EventEmitter<MdSelectionListOptionEvent>();
+
+  /** Emitted when the option is selected. */
+  @Output() select = new EventEmitter<MdSelectionListOptionEvent>();
+
+  /** Emitted when the option is deselected. */
+  @Output() deselect = new EventEmitter<MdSelectionListOptionEvent>();
```
tinayuangao: Also test when an option is deselected?

src/lib/list/selection-list.spec.ts
sllethe: :+1:  fix typo

sllethe: :+1:  fixed typo

src/lib/list/selection-list.spec.ts
sllethe: :+1:  Added new test on native element's focus & blur

src/lib/list/selection-list.spec.ts
sllethe: :+1:  Added check `expect(selectList.selected.length).toBe(0); `before the `toggle`

src/lib/list/selection-list.spec.ts
sllethe: :+1:

src/lib/list/list-option.ts
sllethe: :+1: Added test for deselecting an option

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    expect(selectList.selected.length).toBe(0);
+
+    testListItem.toggle();
+    fixture.detectChanges();
+
+    expect(selectList.selected.length).toBe(1);
+  });
+
+  it('should be able to dispatch selected more than one items', () => {
```
kara: Nit: `selected more than one items` -> `multiple selected items`

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    expect(selectList.selected.length).toBe(1);
+
+    testListItem.toggle();
+    fixture.detectChanges();
+
+    expect(selectList.selected.length).toBe(0);
+  });
+
+  it('test disabled items cannot be selected', () => {
```
kara: Nit: if you can, the test names should start with 'should', so it reads like a sentence:

`it('should not select disabled items')`

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,187 @@
...
+      this._tabIndex = -1;
+    }
+
+    this._optionsChangeSubscriptionOnFocus = this.onFocusSubscription();
+    this._optionsChangeSubscriptionDestroy = this.onDestroySubscription();
+  }
+
+  ngOnDestroy(): void {
+    if (this._tabOutSubscription) {
+      this._tabOutSubscription.unsubscribe();
```
kara: What is this subscription? It looks like it's never created.

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,187 @@
...
+  private _disableRipple: boolean = false;
+
+  /** Tab index for the selection-list. */
+  _tabIndex = 0;
+
+  /** Subscription to tabbing out from the selection-list. */
+  private _tabOutSubscription: Subscription;
+
+  /** Subscription to all list options' onFocus events */
+  private _optionsChangeSubscriptionOnFocus: Subscription;
```
kara: How about `_optionFocusSubscription` and `_optionDestroyStream`?

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  /**
+   * Map all the options' destroy event subscriptions and merge them into one stream.
+   */
+  onDestroySubscription(): Subscription {
+    let subscription = RxChain.from(this.options.changes)
+      .call(startWith, this.options)
+      .call(switchMap, (options: MdListOption[]) => {
+        return merge(...options.map(option => option.destroy));
```
kara: Split this into a getter? Same for the focus stream.

```ts
...

.call(switchMap, (options) => this._optionDestroyStream(options))

...

get _optionDestroyStream(options: MdListOption[]) {
   return merge(...options.map(option => option.destroy));
}
```

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  focus() {
+    this._element.nativeElement.focus();
+  }
+
+  /**
+   * Map all the options' destroy event subscriptions and merge them into one stream.
+   */
+  onDestroySubscription(): Subscription {
+    let subscription = RxChain.from(this.options.changes)
```
kara: Probably don't need this variable; can return directly. Same thing for the focus subscription.

```ts
return RxChain.from(this.options.changes)...
```

sllethe: :+1:

src/lib/list/selection-list.ts
sllethe: It seems we can not make it a getter. 
Because there'll be an error `'TS1054:  a 'get' accessor can not have parameters'`

kara: Facepalm - sorry for the bad demo code. We have a few options:

- We could split it into a regular function - _getOptionStream(options)

- We could use `this.options` directly and keep it as a getter

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,299 @@
+import {async, TestBed} from '@angular/core/testing';
+import {Component} from '@angular/core';
+import {By} from '@angular/platform-browser';
+import {MdSelectionList, MdListOption, MdListModule} from './index';
+import {createKeyboardEvent} from '@angular/cdk/testing';
+import {UP_ARROW, DOWN_ARROW, SPACE} from '../core/keyboard/keycodes';
+import {Platform} from '../core/platform/index';
+
+
+describe('SelectionList and ListOption', () => {
```
jelbourn: Need tests for `aria-selected` and `aria-disabled` (can be extra checks added to existing selection tests)

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+    });
+
+    TestBed.compileComponents();
+  }));
+
+  it('should be focused when focus on nativeElements', () => {
+    let fixture = TestBed.createComponent(SelectionListWithOnlyOneOption);
+    fixture.detectChanges();
+    let listOption = fixture.debugElement.query(By.directive(MdListOption)).nativeElement;
+    let listItemEl = fixture.debugElement.query(By.css('.mat-list-item'));
```
sllethe: :+1:  changed the structure of test like tests in radio buttons

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    listOption.focus();
+    fixture.detectChanges();
+
+    expect(listItemEl.nativeElement).toBe(document.activeElement);
+    if (fixture.componentInstance.isIE == false) {
```
sllethe: :+1: Changed to `if (!fixture.componentInstance.isIE) { ... }`

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+}
+
+@Component({template: `
+  <mat-selection-list id = "selection-list-4">
+    <md-list-option checkboxPosition = "after" class="test-focus" id="123">
+      Inbox
+    </md-list-option>
+  </mat-selection-list>`})
+class SelectionListWithOnlyOneOption {
+  isIE: boolean = true;
```
sllethe: :+1:  Used it and deleted `Platform` in the test component

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+  constructor(public platform: Platform) {
+    if (platform.SAFARI || platform.EDGE || platform.FIREFOX) {
```
sllethe: Changed `'(platform.SAFARI || platform.EDGE || platform.FIREFOX)'` to `'!platform.IS_TRIDENT'`

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+      this.isIE = false;
+    } else {
+      this.isIE = true;
+    }
```
sllethe: :+1: 
Deleted the platform part in test component

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,353 @@
...
+
+    testListItem.toggle();
+    fixture.detectChanges();
+
+    testListItem2.toggle();
+    fixture.detectChanges();
+
+    expect(selectList.selected.length).toBe(2);
+    expect(listOption[2].nativeElement.getAttribute('aria-selected')).toBe('true');
+    expect(listOption[1].nativeElement.getAttribute('aria-selected')).toBe('true');
```
jelbourn: There should also be assertions somewhere where you check `aria-selected` being false (right now you only ever check that it's true)

(same for aria-disabled)

sllethe: :+1: Added `aria-selected` being `false` before changing it to `true`.
Added a new test case `'should be able to un-disable disabled items'` to test changes on `aria-disabled`

sllethe:
Comments Addressed :+1: 
Please take another look

src/lib/list/list.scss
```diff
@@ -43,6 +43,18 @@ $mat-dense-list-icon-size: 20px;
     position: relative;
   }
 
+  .mat-list-item-disabled {
+    background: #ededed;
```
tinayuangao: We should set color in theme.scss file

sllethe: :+1:  put disabled style in _list_theme.scss

sllethe:
:+1: Comments addressed
Please take another look

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,178 @@
...
+        // Always prevent space from scrolling the page since the list has focus
+        event.preventDefault();
+        break;
+      default:
+        this._keyManager.onKeydown(event);
+    }
+  }
+
+  /** Toggles the selected state of the currently focused option. */
+  protected _toggleSelectOnFocusedOption(): void {
```
jelbourn: This method should be private

src/lib/list/selection-list.spec.ts
```diff
@@ -0,0 +1,370 @@
+import {async, TestBed, ComponentFixture, inject} from '@angular/core/testing';
+import {Component, DebugElement} from '@angular/core';
+import {By} from '@angular/platform-browser';
+import {MdSelectionList, MdListOption, MdListModule} from './index';
+import {createKeyboardEvent} from '@angular/cdk/testing';
+import {UP_ARROW, DOWN_ARROW, SPACE} from '../core/keyboard/keycodes';
+import {Platform} from '../core/platform/index';
+
+describe('test normal selection list with list option', () => {
```
jelbourn: You can wrap all of these child `describe` blocks in one top-level block:
```ts
describe('MdSelectionList', () => {
  ...
});
```

Then, each describe block has a name that modifies the selection list, e.g.

```ts
describe('MdSelectionList', () => {
  describe('with multiple options', () => {
    ...
  });

  describe('with one option', () => {
    ...
  });

  describe('with disabled state', () => {
    ...
  });
});
```

This will make the test failures report like 
```
MdSelectionList with one option should select an option
```

src/lib/list/_list-theme.scss
```diff
@@ -8,14 +8,22 @@
...
+    .mat-list-option {
+      color: mat-color($foreground, text);
+    }
+
     .mat-subheader {
       color: mat-color($foreground, secondary-text);
     }
+
+    .mat-list-item-disabled {
+      background: #ededed;
```
tinayuangao: Use `mat-color` for disabled background? Like in https://github.com/angular/material2/blob/0409707adc3dd3114730310a95e4f9ac1827b3dd/src/lib/button-toggle/_button-toggle-theme.scss#L32

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,157 @@
...
```
sllethe: :+1: Changed deselect -> deselected

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  /** Emitted when the option is destroyed. */
+  @Output() destroy = new EventEmitter<MdSelectionListOptionEvent>();
```
sllethe: :+1:

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,157 @@
...
+      this._disabled = true;
+    }
+  }
+
+  ngOnDestroy(): void {
+    this.destroy.emit({option: this});
+  }
+
+  toggle(): void {
+    if (this._disabled == false && this.selectionList.disabled == false) {
```
sllethe: :+1: changed to : 
```
if (!this.disabled) {
  ...
}
```

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,156 @@
...
+              private _element: ElementRef,
+              private _changeDetector: ChangeDetectorRef,
+              @Optional() public selectionList: MdSelectionList) { }
+
+
+  ngAfterContentInit() {
+    this._lineSetter = new MdLineSetter(this._lines, this._renderer, this._element);
+
+    if (this.selectionList.disabled) {
+      this._disabled = true;
```
sllethe: Changed to : 
```
if (!this.disabled) {
  ...
}
```

src/lib/list/selection-list.spec.ts
sllethe: :+1: Changed the structure like that~

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  focus() {
+    this._element.nativeElement.focus();
+  }
+
+  /**
+   * Map all the options' destroy event subscriptions and merge them into one stream.
+   */
+  onDestroySubscription(): Subscription {
```
sllethe: :+1:  Changed it to 'private _onDestroySubscription()'

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,178 @@
...
+          }
+        }
+        e.option.destroy.unsubscribe();
+      });
+  }
+
+  /**
+   * Map all the options' onFocus event subscriptions and merge them into one stream.
+   */
+  onFocusSubscription(): Subscription {
```
sllethe: :+1: Changed to 'private _onFocusSubscription()'

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,178 @@
...
+      .call(switchMap, (options: MdListOption[]) => {
+        return merge(...options.map(option => option.onFocus));
+      }).subscribe((e: MdSelectionListOptionEvent) => {
+      let optionIndex: number = this.options.toArray().indexOf(e.option);
+      this._keyManager.updateActiveItemIndex(optionIndex);
+    });
+  }
+
+  /** Passes relevant key presses to our key manager. */
+  keydown(event: KeyboardEvent) {
```
sllethe: :+1: Changed to `'_keydown'`

src/lib/list/selection-list.ts
sllethe: :+1: Changed to 'private'

src/lib/list/_list-theme.scss
sllethe: :+1: changed to `background-color: mat-color($background, disabled-list-option); `

src/lib/list/_list-theme.scss
```diff
@@ -38,27 +54,49 @@
...
+      font-size: mat-font-size($config, subheading-2);
+      @include mat-line-base(mat-font-size($config, body-1));
+    }
+
     .mat-subheader {
       @include mat-typography-level-to-styles($config, body-2);
     }
+
+    .mat-list-item-disabled {
+      background: #ededed;
```
sllethe: changed to background-color: mat-color($background, disabled-list-option);

src/lib/core/theming/_palette.scss
```diff
@@ -659,6 +659,7 @@ $mat-light-theme-background: (
   selected-button: map_get($mat-grey, 300),
   selected-disabled-button: map_get($mat-grey, 400),
   disabled-button-toggle: map_get($mat-grey, 200),
+        disabled-list-option: map_get($mat-grey, 200),
```
tinayuangao: nit: indent

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  @Input()
+  get selected() { return this._selected; }
+  set selected( val: boolean) { this._selected = coerceBooleanProperty(val); }
+
+  /** Emitted when the option is focused. */
+  onFocus = new EventEmitter<MdSelectionListOptionEvent>();
+
+  /** Emitted when the option is selected. */
+  @Output() select = new EventEmitter<MdSelectionListOptionEvent>();
```
tinayuangao: or rename it to `selectChange`?

src/lib/core/theming/_palette.scss
tinayuangao: nit: indent

src/lib/list/list-option.ts
tinayuangao: Better to delete `if (this.selectionList.disabled) {...}` part

src/lib/list/list-option.ts
```diff
@@ -0,0 +1,159 @@
...
+      this._disabled = true;
+    }
+  }
+
+  ngOnDestroy(): void {
+    this.destroyed.emit({option: this});
+  }
+
+  toggle(): void {
+    if (!this._disabled) {
```
tinayuangao: Use `this.disabled` to consider the case when parent selectionList is disabled?

src/lib/list/list.scss
```diff
@@ -182,3 +212,5 @@ $mat-dense-list-icon-size: 20px;
     }
   }
 }
+
```
tinayuangao: nit: too many empty lines

src/lib/list/list-option.ts
sllethe: deleted `if (this.selectionList.disabled) {...}` part in `toggle()`, and put the`if (this.selectionList.disabled) {...}` part into a new function which is `'_handleClick()'`.

src/lib/list/list-option.ts
sllethe: :+1:  Changed to `this.disabled`

src/lib/list/list.scss
sllethe: :+1:  Deleted empty lines

tinayuangao:
LGTM. Please sync and rebase.

sllethe:
:+1: synced and rebased

jelbourn:
@sllethe seems like the unit tests are failing

sllethe:
fixed failing tests~

src/lib/list/selection-list.ts
```diff
@@ -0,0 +1,321 @@
...
+  /** Emitted when the option is deselected. */
+  @Output() deselected = new EventEmitter<MdSelectionListOptionEvent>();
+
+  /** Emitted when the option is destroyed. */
+  @Output() destroyed = new EventEmitter<MdSelectionListOptionEvent>();
+
+  constructor(private _renderer: Renderer2,
+              private _element: ElementRef,
+              private _changeDetector: ChangeDetectorRef,
+              @Self() @Optional() @Inject(forwardRef(() => MdSelectionList)) public selectionList: MdSelectionList) { }
```
mmalerba: No need for `@Self()` here, and `Inject` needs to be added to imports

sllethe: 👍  Deleted `@Self()` and added import for `Inject`

PR #5583, opened 2017-07-07
fix(card): prevent content from overlapping footer

### LINKED ISSUE #5486: md-card-actions button overlaps md-card-footer content
Bug

The default styling should not allow an md-button to overlap the md-footer content

If a md-footer is added to the card-fancy-example, then the buttons in the md-card-actions overlap the footer text

https://plnkr.co/edit/Y57GHoYFIUflnWEmDEIq?p=preview

The default behavior of a component should not be to overlay buttons and content

@angular/cli: 1.0.5
node: 7.10.0
os: win32 x64
@angular/animations: 4.1.3
@angular/common: 4.1.3
@angular/compiler: 4.1.3
@angular/core: 4.1.3
@angular/flex-layout: 2.0.0-rc.1
@angular/forms: 4.1.3
@angular/http: 4.1.3
@angular/material: 2.0.0-beta.5
@angular/platform-browser: 4.1.3
@angular/platform-browser-dynamic: 4.1.3
@angular/router: 4.1.3
@angular/cli: 1.0.5
@angular/compiler-cli: 4.1.3

The behavior shows itself with the existing examples by adding an md-card-footer. In my plunker I have simply added that and made one of the buttons a 'raised' button to make the issue more obvious

crisbeto:
* Prevents content from overlapping the `md-card-footer` component by making the footer `position: static`.
* Uses content projection to ensure that the footer will always be at the bottom of a card.

Fixes #5486.

### DIFF
--- a/src/lib/card/card.scss
+++ b/src/lib/card/card.scss
@@ -62,11 +62,12 @@ $mat-card-header-size: 40px !default;
 }
 
 .mat-card-footer {
-  position: absolute;
-  width: 100%;
-  min-height: 5px;
-  bottom: 0;
-  left: 0;
+  // The negative margins pulls out the element, countering the padding
+  // to get the footer to be flush against the side of the card.
+  $margin: -$mat-card-default-padding;
+
+  display: block;
+  margin: 0 $margin $margin $margin;
 }
 
 .mat-card-actions {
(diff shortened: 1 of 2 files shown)


src/lib/card/card.scss
```diff
@@ -62,11 +62,10 @@ $mat-card-header-size: 40px !default;
 }
 
 .mat-card-footer {
-  position: absolute;
-  width: 100%;
-  min-height: 5px;
-  bottom: 0;
-  left: 0;
+  $margin: -$mat-card-default-padding;
```
jelbourn: Could you add comments that explain the negative margins? I find those to usually be hard to grok when coming back to the styles later.

crisbeto:

PR #5592, opened 2017-07-07
refactor: use cdk imports instead of re-exports

DevVersion:
Recently a bunch of exports have been moved from Material to the CDK and to avoid a lot of changes the imports for those exports stayed the same.

This is possible because every moved export was just re-exported in the original source file again.

Now the imports are switched to the real `@angular/cdk` imports and the unnecessary files inside of `core/` have been removed

**Note**: A big benefit is the `cdk-exports.ts` file because this allows us to easily remove those exports from Material at some point. And also it's way easier to see what is actually re-exported as part of the Material.


src/lib/tooltip/tooltip.ts
```diff
@@ -35,11 +35,9 @@ import {
...
   OriginConnectionPosition,
 } from '../core';
+import {Directionality, Platform, first} from '@angular/cdk';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
-import {Directionality} from '../core/bidi/index';
-import {Platform} from '../core/platform/index';
-import {first} from '../core/rxjs/index';
 import {ScrollDispatcher} from '../core/overlay/scroll/scroll-dispatcher';
 import {coerceBooleanProperty} from '@angular/cdk';
```
rafaelss95: I think this can be merged to another import:

`import {coerceBooleanProperty, Directionality, Platform, first} from '@angular/cdk';`

DevVersion: Made the changes. Also FYI: Those will be merged by Rollup in the end anyway.

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -17,18 +17,27 @@ import {
...
+  DOWN_ARROW,
+  SPACE,
+  UP_ARROW,
+  ESCAPE,
+  Directionality,
+  Direction,
+  RxChain,
+  map,
+  startWith
+} from '@angular/cdk';
```
jelbourn: Hmm, now that I'm seeing these imports, I'm thinking we need to change the `cdk` to use imports like
```ts
import {LiveAnnouncer} from '@angular/cdk/a11y';
```
Otherwise the library is too much of a grab-bag of unrelated stuff (especially for the aliased rxjs operators and keycodes)

Thoughts? It might be easier to make such a change first rather than to unroll the imports once they're all `@angular/cdk`

DevVersion: I see what you mean but I think this is more like a separate thing we need to revisit later.

It goes kind of back to the initial approach with the secondary entry-point and requires us to rework the packaging (big changes)

For now changing the imports shouldn‘t make it easier to make such a change in the future (we already import frequently from `core.ts` and it‘s not that difficult)

PR #5604, opened 2017-07-08
feat(autocomplete): support for md-optgroup

### LINKED ISSUE #5581: Md-autocomplete always hidden when used with md-optgroup
#### Feature:

Autocomplete is hidden when the dropdown contains an `md-optgroup` with `md-option` inside it.

Autocomplete should be compatible with `md-optgroup`.

The `md-autocomplete` stays hidden as shown in this plunker:

https://plnkr.co/edit/JIU35gPRfqDaBBwzMG6S?p=preview

Material: 2.0.0-beta.8

crisbeto:
Adds support for `md-optgroup` in `md-autocomplete` by:
* Changing the `@ViewChild` query to pick up descendant options.
* Tweaking the keyboard scrolling to handle having group labels before options.

Fixes #5581.

### DIFF
--- a/src/demo-app/autocomplete/autocomplete-demo.ts
+++ b/src/demo-app/autocomplete/autocomplete-demo.ts
@@ -3,6 +3,16 @@ import {FormControl, NgModel} from '@angular/forms';
 import 'rxjs/add/operator/startWith';
 import 'rxjs/add/operator/map';
 
+interface State {
+  code: string;
+  name: string;
+}
+
+interface StateGroup {
+  letter: string;
+  states: State[];
+}
+
 @Component({
   moduleId: module.id,
   selector: 'autocomplete-demo',
@@ -13,6 +23,7 @@ import 'rxjs/add/operator/map';
 export class AutocompleteDemo {
   stateCtrl: FormControl;
   currentState = '';
+  currentGroupedState = '';
   topHeightCtrl = new FormControl(0);
 
   reactiveStates: any;
@@ -22,7 +33,9 @@ export class AutocompleteDemo {
 
   @ViewChild(NgModel) modelDir: NgModel;
 
-  states = [
+  groupedStates: StateGroup[];
+  filteredGroupedStates: StateGroup[];
+  states: State[] = [
     {code: 'AL', name: 'Alabama'},
     {code: 'AK', name: 'Alaska'},
     {code: 'AZ', name: 'Arizona'},
@@ -82,6 +95,19 @@ export class AutocompleteDemo {
         .startWith(this.stateCtrl.value)
         .map(val => this.displayFn(val))
         .map(name => this.filterStates(name));
+
+    this.filteredGroupedStates = this.groupedStates = this.states.reduce((groups, state) => {
+      let group = groups.find(g => g.letter === state.name[0]);
+
+      if (!group) {
+        group = { letter: state.name[0], states: [] };
+        groups.push(group);
+      }
+
+      group.states.push({ code: state.code, name: state.name });
+
+      return groups;
+    }, [] as StateGroup[]);
   }
 
   displayFn(value: any): string {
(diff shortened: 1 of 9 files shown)


src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -853,6 +854,84 @@ describe('MdAutocomplete', () => {
...
+      DOWN_ARROW_EVENT = createKeyboardEvent('keydown', DOWN_ARROW);
+      UP_ARROW_EVENT = createKeyboardEvent('keydown', UP_ARROW);
+
+      fixture.componentInstance.trigger.openPanel();
+      fixture.detectChanges();
+    });
+
+    it('should scroll to active options below the fold', fakeAsync(() => {
+      tick();
+      const container = document.querySelector('.cdk-overlay-pane .mat-autocomplete-panel')!;
```
DevVersion: How about querying that element in the `beforeEach`? The `tick()` and `container` variable initialization is done in every test of the current `describe`.

crisbeto: Good point. Done/

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -1551,3 +1624,37 @@ class AutocompleteWithNativeInput {
...
   @ViewChild(MdAutocompleteTrigger) trigger: MdAutocompleteTrigger;
 }
+
+@Component({
+  template: `
+    <md-input-container>
+      <input mdInput placeholder="State" [mdAutocomplete]="auto" [(ngModel)]="selectedState">
+    </md-input-container>
+
+    <md-autocomplete #auto="mdAutocomplete">
```
kara: Can you add an autocomplete like this to the autocomplete demo page?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -343,6 +342,29 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
+    let optionIndex = this.autocomplete._keyManager.activeItemIndex || 0;
+
+    // If there are any option groups, we need to offset
+    // the index by the amount of groups that come before the option.
+    if (this.autocomplete.optionGroups.length) {
+      const options = this.autocomplete.options.toArray();
+      const groups = this.autocomplete.optionGroups.toArray();
+      let groupCounter = 0;
+
+      for (let i = 0; i < optionIndex + 1; i++) {
```
kara: This process looks pretty similar to how select calculates option index [here](https://github.com/crisbeto/material2/blob/8ad659f07b384afdd77cc3887c66e99141905f6d/src/lib/select/select.ts#L1048). Given this is a common operation for option groups, maybe we could move this into a reusable function in core/option? A bit DRY-er.

kara:
Can we also add a blurb to the autocomplete docs, so people know about this feature?

src/lib/autocomplete/autocomplete-trigger.ts
crisbeto: Sure, but I'm not sure what that would even be called. `countGroupLabelsBeforeOption`? I would rather put it as a static method on the `MdOption` instead since having it float around as a function implies that it's a bit more reusable.

kara: `countGroupsBeforeOption` sounds reasonable, but open to having it as a static method on `MdOption` too.

crisbeto:
Added all the changes as discussed.

src/demo-app/autocomplete/autocomplete-demo.ts
```diff
@@ -3,6 +3,16 @@ import {FormControl, NgModel} from '@angular/forms';
 import 'rxjs/add/operator/startWith';
 import 'rxjs/add/operator/map';
 
+interface State {
+  code: string;
+  name: string;
+}
+
+interface StateGroup {
```
kara: Failing presubmit because of TS error for "private name `StateGroup`" and "private name `State`". I think you need to export these interfaces.

kara: Also can you fix the lint failure? https://travis-ci.org/angular/material2/jobs/266064783

crisbeto:
Fixed @kara.

PR #5631, opened 2017-07-09
chore(tabs): switch to OnPush change detection

crisbeto:
Switches all of the tab-related components to OnPush change detection and sorts out the issues that come from the changes.

Relates to #5035.

**Note:** I'm not a fan of having to subscribe to changes in the tab labels just so we can trigger change detection, but if we want something more practical, we'll have to restructure the tabs so they don't have to reach into other components.

### DIFF
--- a/src/lib/tabs/tab-group.ts
+++ b/src/lib/tabs/tab-group.ts
@@ -16,11 +16,19 @@ import {
   ContentChildren,
   ElementRef,
   Renderer2,
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  AfterViewChecked,
+  AfterContentInit,
+  AfterContentChecked,
+  OnDestroy,
 } from '@angular/core';
 import {coerceBooleanProperty} from '../core';
 import {Observable} from 'rxjs/Observable';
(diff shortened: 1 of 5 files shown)


src/lib/tabs/tab-group.ts
```diff
@@ -141,7 +155,28 @@ export class MdTabGroup {
...
-    this._selectedIndex = indexToSelect;
+    if (this._selectedIndex !== indexToSelect) {
+      this._selectedIndex = indexToSelect;
+      this._changeDetectorRef.markForCheck();
+    }
+  }
+
+  ngAfterContentInit() {
+    this._subscribeToTabLabels();
+    this._tabsSubscription = this._tabs.changes.subscribe(() => {
```
jelbourn: Add a comment explaining why this is needed?

src/lib/tabs/tab-group.ts
```diff
@@ -165,6 +200,16 @@ export class MdTabGroup {
     return event;
   }
 
+  private _subscribeToTabLabels() {
```
jelbourn: Add comment for why this is needed?

crisbeto:

Issue #5632, opened 2017-07-09
Link Fab Buttons incorrect line height


amcdnl:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
The line height would match the default button line height.

#### What is the current behavior?
The line height is not correct.

![img](https://content.screencast.com/users/amcdaniel22/folders/Snagit/media/2725ddd8-07ee-4a8a-8f93-21e06bfb99ac/2017-07-09_10-32-13.png)

#### What are the steps to reproduce?
- Create a link fab button
- Create a button fab button
- Compare result

https://plnkr.co/edit/Is54mCs52Cd9eL2ollrZ?p=preview

m98:
@amcdnl Maybe we should add an absolute position to  `mat-button-wrapper` and add the relative position to its parent, and it's easy to put it in the center!

PR #5634, opened 2017-07-09
bug(button): properly align text in fab buttons #5632

amcdnl:
Addresses: Link Fab Buttons incorrect line height #5632

**BEFORE**
img2

**AFTER**
img

### DIFF
--- a/src/lib/button/_button-theme.scss
+++ b/src/lib/button/_button-theme.scss
@@ -107,7 +107,7 @@
 }
 
 @mixin mat-button-typography($config) {
-  .mat-button, .mat-raised-button, .mat-icon-button {
+  .mat-button, .mat-raised-button, .mat-icon-button, .mat-fab, .mat-mini-fab {
     font: {
       family: mat-font-family($config);
       size: mat-font-size($config, button);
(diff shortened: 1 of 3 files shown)


src/lib/button/_button-theme.scss
```diff
@@ -107,7 +107,7 @@
 }
 
 @mixin mat-button-typography($config) {
-  .mat-button, .mat-raised-button, .mat-icon-button {
+  .mat-button, .mat-raised-button, .mat-icon-button, .mat-fab, .mat-mini-fab {
```
jelbourn: @crisbeto did you have a reason for omitting the fab buttons here?

crisbeto: I don't remember, but it was probably because the fabs aren't really meant to be used with text.

amcdnl: @crisbeto  Is there any reason why not to?

crisbeto: The spec doesn't really have fabs with text, if I remember correctly, but it shouldn't be an issue to keep it anyway.

amcdnl: @crisbeto  We good to merge then?

jelbourn:
Caretaker note: let me know if this causes any screenshot diffs

kara:
@jelbourn No screenshot diff failures, so merging.

PR #5647, opened 2017-07-10
refactor(input): remove dash-case selectors for mdPrefix and mdSuffix

### LINKED ISSUE #5643: md-input-container with md-prefix produces unexpected DOM
Bug,
`<md-input-container>
      <span md-prefix>
          <md-icon>home</md-icon>
      </span>
      <input mdInput name="givenName" readonly disabled>
  </md-input-container>`
gives wrong DOM, `div.mat-input-prefix` is empty while the `span.md-prefix` is generated inside `div.mat-input-infix`
could be seen on the following plnkr:
http://plnkr.co/edit/WRtJEaWx0ilHPBeGLSM5?p=preview
which is regression from beta.6 on 4.1.3:
http://plnkr.co/edit/dGxYTBaObsjDs557jM4k?p=preview

crisbeto:
Removes the `md-prefix` and `md-suffix` selectors from their respective directives.

BREAKING CHANGES: Any uses of `md-prefix` and `md-suffix` should be switched over to `mdPrefix` and `mdSuffix`.

Fixes #5643.

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -104,14 +104,14 @@ export class MdErrorDirective { }
 
 /** Prefix to be placed the the front of the input. */
 @Directive({
-  selector: '[mdPrefix], [matPrefix], [md-prefix]'
+  selector: '[mdPrefix], [matPrefix]'
 })
 export class MdPrefix {}
 
 
 /** Suffix to be placed at the end of the input. */
 @Directive({
-  selector: '[mdSuffix], [matSuffix], [md-suffix]'
+  selector: '[mdSuffix], [matSuffix]'
 })
 export class MdSuffix {}
 


src/lib/input/input-container.html
```diff
@@ -23,7 +23,7 @@
     </div>
 
     <div class="mat-input-suffix" *ngIf="_suffixChildren.length">
-      <ng-content select="[mdSuffix], [matSuffix]"></ng-content>
+      <ng-content select="[mdSuffix], [matSuffix], [md-suffix]"></ng-content>
```
jelbourn: Can we just remove the dash-case selectors at this point?

crisbeto: I suppose, I wasn't sure what the plan for deprecating them was.

jelbourn: Looks like they're not used inside google, so should be safe to remove

crisbeto:
Re-purposed the PR to deprecate the dash-case selectors @jelbourn.

PR #5650, opened 2017-07-10
fix(expansion-panel): toggle not being updated when set programmatically

### LINKED ISSUE #5623: bug(expansion): toggle icon not being updated

bug

If you switch the `expanded` input from false-true-false... the toggle icon on the upper right corner should change it's direction.

It's not changing, unless you click on the expansion header

At the plunker, click on the **`toggle panel`** button bellow the expansion to switch its state and watch the not-changing toggle icon.

Plunker: https://plnkr.co/edit/SIkVCu?p=preview

Programmatically open and close the expansion

Material latest builds

###  `expanded == true`

image

###  `expanded == false`

image

crisbeto:
Fixes the toggle arrow not being flipped when the `expanded` is set programmatically, in addition to the `hideToggle` input not working either. This is a regression from #5549 and is a consequence of the fact that the panel header reaches into the panel to determine what to do with the arrow.

Fixes #5623.

**Note:** Similarly to #5631, if we want something more practical, we'll have to re-think where the input for these properties should go.

### DIFF
--- a/src/lib/expansion/expansion-panel-header.ts
+++ b/src/lib/expansion/expansion-panel-header.ts
@@ -12,6 +12,8 @@ import {
   Host,
   ViewEncapsulation,
   ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  OnDestroy,
 } from '@angular/core';
 import {
   trigger,
@@ -22,6 +24,9 @@ import {
 } from '@angular/animations';
 import {SPACE, ENTER} from '../core/keyboard/keycodes';
 import {MdExpansionPanel, EXPANSION_PANEL_ANIMATION_TIMING} from './expansion-panel';
+import {filter} from '../core/rxjs/index';
+import {merge} from 'rxjs/observable/merge';
+import {Subscription} from 'rxjs/Subscription';
 
 
 /**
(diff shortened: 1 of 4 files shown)


src/lib/expansion/expansion-panel-header.ts
```diff
@@ -62,8 +67,22 @@ import {MdExpansionPanel, EXPANSION_PANEL_ANIMATION_TIMING} from './expansion-pa
...
+    private _changeDetectorRef: ChangeDetectorRef) {
+
+    // Since the toggle state depends on an @Input on the panel, we
+    // need to  subscribe and trigger change detection manually.
+    this._parentChangeSubscription = merge(
+      panel.opened,
+      panel.closed,
+      filter.call(panel._inputChanges, changes => !!changes.hideToggle)
+    )
+    .subscribe(() => this._changeDetectorRef.markForCheck());
```
jelbourn: Would it be possible to add a unit test for this case?

crisbeto: I can give it a try, but I feel like the manual change detection might cause it not to fail as expected.

crisbeto:
Added those unit tests as discussed @jelbourn.

Issue #3005, opened 2017-02-09
input placeholder covers value when using OnPush detection strategy


fxck:
I know there were couple of these before, but it seems to be happening again https://gist.run/?id=d815c9d4e98685976793d074fcc89c23 

cc @kara @mmalerba maybe?

johnqiuwan:
I had the exactly same issue on my project that the placeholder value overlap with the input value.

realappie:
I have the same issue when using reactive forms `yForm.controls['xControl'].setValue('foobar')`

![issue gif](https://i.gyazo.com/ebf9227ba14684a665d111d692165620.gif)

If I dont change the input using the keyboard, the input just puts the placeholder back to its original position as if the input is empty which its not.

mmalerba:
There are some similar issues to this one that have already been fixed (#2816, #2837). This one only applies if you are using `ChangeDetectionStrategy.OnPush`

ggranum:
I'm seeing this manifest in browser-populated fields - specifically userName and password fields populated by Chrome et. al. My normal temporary hack of 

`
setTimeout(() => {
      this.changeDetectorRef.markForCheck()
    }, 100)
`

fails for this case, however.

JacobBrandt:
I'm also seeing this behavior not only for input but on select as well when using OnPush.  The select control doesn't have it's value selected until a form control is given focus.  Similar to what @realappie shows.

![onpushbug](https://cloud.githubusercontent.com/assets/5314322/24074646/7ef22214-0bd2-11e7-8915-98acce9e5526.gif)

mjolk:
maybe related to #2663 says it's fixed but i'm still having this issue using select in combination with reactive forms/OnPush change detection.( Angular 4.0.0, material 2.0.0.beta2)

mjolk:
for me this all works as expected now.(beta 3)  Good work!

eyalewin:
@mjolk , can you share how did was it fixed for you? i still can't have it working. 
take a look at this: [https://plnkr.co/edit/FncXapgEEdqJU7R3kU8X?p=preview](url)

FilipeQ:
@mjolk What you do for OnPush work with input placeholder, i have last version (2.0.0-beta.3) and the problem persist.

ggranum:
Still an issue for me as well. I've worked around it with:

```
setTimeout(() => {
      changeDetectorRef.markForCheck()
    }, 100)
```
In ngOnInit / ngAfterViewInit.

[...]

PR #5660, opened 2017-07-10
fix(input): placeholder covering value when using OnPush

willshowell:
Fixes https://github.com/angular/material2/issues/3005

cc @mmalerba


mmalerba:
Can you make the demo `OnPush` so I can test this? I seem to remember trying something similar and it didn't quite work

willshowell:
@mmalerba done, PTAL

PR #5662, opened 2017-07-10
fix(docs): angular -> angular material

rafaelss95:


jelbourn:
This one is okay to stay as "Angular" since it's referring to the overall Angular community.

PR #5686, opened 2017-07-11
fix(snackbar): make closeWithAction public method

### LINKED ISSUE #5657: Bug: Snackbar missing public action method for custom snackbars
Bug

Be able to close a custom snackbar using a public method to indicate action was used, and allow for use of `onAction` observable

When you create a custom snackbar using `openFromComponent` you can pass in a snackBarRef and invoke any methods on self, but there is no public version of `_action()` so if you create a component and a user clicks your action you have to use a private method to be able to distinguish an action/dismiss or just a dismiss

N/A missing API call

Snackbar's have two ways they can be closed, but right now in order to use the `onAction` observable you need to invoke a private method `_action()`

Angular 4.2.5, Angular Materila 2.0.0-beta.8, Windows 10

willshowell:
Fixes #5657

This also changes the SimpleSnackBar's api from `dismiss()` to `action()` bc the previous naming was misleading.

### DIFF
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -25,13 +25,13 @@ export class MdSnackBarRef<T> {
   containerInstance: MdSnackBarContainer;
 
   /** Subject for notifying the user that the snack bar has closed. */
-  private _afterClosed: Subject<any> = new Subject();
+  private _afterClosed = new Subject<void>();
 
   /** Subject for notifying the user that the snack bar has opened and appeared. */
-  private _afterOpened: Subject<any>;
+  private _afterOpened = new Subject<void>();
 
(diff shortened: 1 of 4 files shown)


src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -55,19 +55,19 @@ export class MdSnackBarRef<T> {
...
   }
 
-  /** Dismisses the snack bar after some duration */
-  _dismissAfter(duration: number): void {
-    this._durationTimeoutId = setTimeout(() => this.dismiss(), duration);
-  }
-
   /** Marks the snackbar action clicked. */
-  _action(): void {
+  closeWithAction(): void {
```
jelbourn: Since this is a public API now, could you add a unit test that is based around calling this? The current unit test only deals with clicking the button directly.

PR #5692, opened 2017-07-11
chore(input): switch to OnPush change detection

crisbeto:
Switches the `MdInputContainer` to `OnPush` change detection and sorts out the various change detection-related issues.

Relates to #5035.

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -16,10 +16,8 @@ import {
   ContentChildren,
   Directive,
   ElementRef,
-  EventEmitter,
   Input,
   Optional,
-  Output,
   QueryList,
   Renderer2,
   Self,
@@ -53,7 +51,6 @@ import {
   MD_ERROR_GLOBAL_OPTIONS
 } from '../core/error/error-options';
 import {Subject} from 'rxjs/Subject';
-import {filter} from '../core/rxjs/index';
 
 // Invalid input type. Using one of these will throw an MdInputContainerUnsupportedTypeError.
 const MD_INPUT_INVALID_TYPES = [
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -25,7 +25,11 @@ import {
   Self,
   ViewChild,
   ViewEncapsulation,
-  Inject
+  Inject,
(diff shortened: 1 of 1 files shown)


src/lib/input/input-container.ts
```diff
@@ -128,53 +134,48 @@ export class MdSuffix {}
...
+  /** Whether the input is in an error state. */
+  _isErrorState = false;
 
   /** Whether the element is focused or not. */
   focused = false;
 
   /** Sets the aria-describedby attribute on the input for improved a11y. */
   ariaDescribedby: string;
 
+  /** Stream that emits whenever one of the input states changes. */
```
jelbourn: Expand this comment to include why this is necessary at all?

src/lib/input/input-container.ts
```diff
@@ -232,24 +226,51 @@ export class MdInputDirective {
...
+  }
 
-  _onFocus() { this.focused = true; }
+  ngOnDestroy() {
+    this._stateChanges.complete();
+  }
 
-  _onBlur() { this.focused = false; }
+  ngDoCheck() {
+    if (this._ngControl) {
```
mmalerba: Ah very cool, this is what I was talking about when I said forms doesn't always give us a way to know when we need to update, but this is a cool workaround :)

src/lib/input/input-container.ts
```diff
@@ -232,24 +226,51 @@ export class MdInputDirective {
...
+      // We need to re-evaluate this on every change detection cycle, because there are some
+      // error triggers that we can't subscribe to (e.g. parent form submissions). This means
+      // that whatever logic is in here has to be super lean or we risk destroying the performance.
```
willshowell: Would it make sense to also add a note to the Custom Error Matcher section of input.md that mentions this?

*Note: An input's error state must be calculated every change detection cycle, so it is important to keep the logic in your error state matcher very lean or you risk destroying performance*.

crisbeto:

@willshowell if nothing else, #5660 still provides some value by adding the extra unit test.

jelbourn:
@crisbeto looks to be a few lint errors

kara:
@crisbeto Rebase?

crisbeto:
Rebased.

PR #5707, opened 2017-07-12
build: add tslint rule to enforce OnPush change detection

crisbeto:
Adds a tslint rule that requires OnPush change detection on all library and CDK components.

Relates to #5035.

**Note:** we should defer merging this until all of the OnPush PRs have gone through. Also it uncovered that we were missing the `MdSidenavContainer` from the list on #5035. I'll sort it out tomorrow.

### DIFF
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -887,12 +887,12 @@ class CheckboxWithFormDirectives {
 }
 
 /** Simple test component with multiple checkboxes. */
-@Component(({
+@Component({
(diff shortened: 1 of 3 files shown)


willshowell:
@crisbeto looks like they've all merged!

jelbourn:
We're currently stuck on an issue w/ the dialog
https://github.com/angular/material2/pull/6287

crisbeto:
No longer relevant since this functionality is part of the `validate-decorators` rule. The check for `changeDetection` will be enabled in https://github.com/angular/material2/pull/7119.

PR #5716, opened 2017-07-13
fix(paginator): expose MdPaginatorIntl

### LINKED ISSUE #5715: MdPaginatorIntl is not exposed.
Bug

`MdPaginatorIntl` token is exposed by `@angular/material`

No exposed.

Requires so it can inject into other services. 

all

Related to #5573

shlomiassaf:
Fixes #5715


lostllama:
ETA for release via NPM?

Lior-G:
I'm also curious as to when will beta.9 will be released?

PR #5718, opened 2017-07-13
fix(dialog): default md-dialog-close result to null instead of ''

### LINKED ISSUE #5237: md-dialog-close and outside click afterClosed() result inconsistent
Bug

The result from md-dialog-close and clicking on the background of a dialog should be the same when subscribing to the afterClosed() event.

md-dialog-close returns an empty string as the result instead of undefined like how clicking on the background of the dialog does.

Plunker template: https://plnkr.co/edit/I5rsuNKuxYZNrdPqTTdZ?p=preview

Make verifying dialog "cancels" easier. Instead of checking for an undefined and empty string result we can just check for an undefined result.

material.beta7

This problem seem to have been introduced after allowing the md-dialog-close directive to include a result.

CDDelta:
fixes #5237

### DIFF
--- a/src/lib/dialog/dialog-content-directives.ts
+++ b/src/lib/dialog/dialog-content-directives.ts
@@ -30,7 +30,7 @@ export class MdDialogClose {
   @Input('aria-label') ariaLabel: string = 'Close dialog';
 
   /** Dialog close input. */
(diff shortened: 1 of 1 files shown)


CDDelta:
I signed it!

PR #5721, opened 2017-07-13
fix toggling OverlayContainer.themeClass

amitport:
when toggling `overlayContainer.themeClass`
```
overlayContainer.themeClass = 'some-class';
overlayContainer.themeClass = '';
overlayContainer.themeClass = 'some-class';
```
we get:
`Uncaught DOMException: Failed to execute 'remove' on 'DOMTokenList': The token provided must not be empty.`

`''` is allowed as input to themeClass so it should be checked for on remove from classList

### DIFF
--- a/src/lib/core/overlay/overlay-container.ts
+++ b/src/lib/core/overlay/overlay-container.ts
@@ -25,7 +25,9 @@ export class OverlayContainer {
   get themeClass(): string { return this._themeClass; }
   set themeClass(value: string) {
     if (this._containerElement) {
-      this._containerElement.classList.remove(this._themeClass);
+      if (this._themeClass) {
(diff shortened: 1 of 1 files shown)


src/lib/core/overlay/overlay-container.ts
```diff
@@ -25,7 +25,9 @@ export class OverlayContainer {
   get themeClass(): string { return this._themeClass; }
   set themeClass(value: string) {
     if (this._containerElement) {
-      this._containerElement.classList.remove(this._themeClass);
+      if (this._themeClass) {
+        this._containerElement.classList.remove(this._themeClass);
```
jelbourn: Could you also add a unit test for this change?

jelbourn:

Issue #5713, opened 2017-07-13
Bug: Focus still on "open dialog" button after clicked


xuanshenbo:
#### Bug, feature request, or proposal:
Bug

#### What are the steps to reproduce?
Use the doc site example: https://material.angular.io/components/dialog/examples
Set focus on the "Launch dialog" button, and hit Enter key multiple times as fast as you can, multiple dialogs will show

#### What is the expected behavior?
Only one dialog should show

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular Material 2.0.0-beta.8

#### Is there anything else we should know?
The reason obviously is because the dialog is not yet initiated immediately after hitting Enter , therefore focus is still on the button. Not sure how we can resolve this.

PR #5726, opened 2017-07-13
fix(paginator): expose paginator intl

jelbourn:


rafaelss95:
I think the PR #5713 did It. @jelbourn @crisbeto

lostllama:
ETA for release via NuGet?

PR #5731, opened 2017-07-13
chore(sidenav): switch to OnPush change detection

crisbeto:
Switches the `MdSidenavContainer` to `OnPush` change detection and fixes some changes not being picked up when its sidenav is opened.

Relates to #5035.


jelbourn:
This caused a screenshot failure where one app's sidenav seems to be in a different mode with this change

PR #5732, opened 2017-07-13
Positively define input text color for dark themes

paviad:
Before this change, input text on dark themes was invisible due to having similar color to dark background.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -39,6 +39,10 @@
     }
   }
 
+  .mat-input-element {
+    color: mat-color($foreground, text);
+  }
+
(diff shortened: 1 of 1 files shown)


paviad:
I signed it!

mmalerba:
Can you share the bug you're trying to fix with this? I think the problem is not with the input, but with the dark-background container. The input sets is color to `currentColor` in order to properly contrast with whatever background its on. It's up to the container to set the background but also set the color to something that contrasts well

paviad:
I can't seem to be able to reproduce the problem now... perhaps it was
fixed in beta 8... Anyway, the pull request may be rejected

PR #5734, opened 2017-07-13
fix(tooltip): a11y improvement, create overlay on init

### LINKED ISSUE #2334: [Tooltip] Should display when the element has focus
As per the spec as defined here https://material.io/guidelines/components/tooltips.html

```
Summoned by:
- Hovering over an element with a cursor
- **Focusing on an element with a keyboard (usually the tab key)**
- Upon touch
```

andrewseguin:
Problem: Screenreaders can't read out the tooltip since the overlay is created on `show`. 

This PR changes the tooltip so that the tooltip overlay is created in the DOM on init so that it is always available to screenreaders. 

Fixes #2334

### DIFF
--- a/src/lib/tooltip/tooltip.ts
+++ b/src/lib/tooltip/tooltip.ts
@@ -100,6 +100,7 @@ export const MD_TOOLTIP_SCROLL_STRATEGY_PROVIDER = {
 export class MdTooltip implements OnDestroy {
   _overlayRef: OverlayRef | null;
   _tooltipInstance: TooltipComponent | null;
+  _scrollStrategy: RepositionScrollStrategy;
 
   private _position: TooltipPosition = 'below';
   private _disabled: boolean = false;
@@ -239,13 +240,19 @@ export class MdTooltip implements OnDestroy {
   show(delay: number = this.showDelay): void {
     if (this.disabled || !this._message || !this._message.trim()) { return; }
 
+    this._scrollStrategy.enable();
     if (this._tooltipInstance) {
       this._tooltipInstance.show(this._position, delay);
+    }
+    if (this._overlayRef) {
+      this._overlayRef.overlayElement.classList.remove('cdk-visually-hidden');
+      this._overlayRef.updatePosition();
     }
   }
 
   /** Hides the tooltip after the delay in ms, defaults to tooltip-delay-hide or 0ms if no input */
   hide(delay: number = this.hideDelay): void {
+    this._scrollStrategy.disable();
     if (this._tooltipInstance) {
       this._tooltipInstance.hide(delay);
     }
@@ -290,10 +297,20 @@ export class MdTooltip implements OnDestroy {
     let portal = new ComponentPortal(TooltipComponent, this._viewContainerRef);
 
     this._tooltipInstance = overlayRef.attach(portal).instance;
+    this._scrollStrategy.disable();
 
     // If the user has not already set an aria-describedby, then use the tooltip's id.
     if (!this._getAriaDescribedby() && this._tooltipInstance) {
       this._setAriaDescribedBy(this._tooltipInstance.id);
+    }
+
+    if (this._tooltipInstance) {
+      this._tooltipInstance.afterHidden().subscribe(() => {
+        if (this._overlayRef) {
+          this._overlayRef.overlayElement.classList.add('cdk-visually-hidden');
+        }
+        this._scrollStrategy.disable();
+      });
     }
 
     this._setTooltipClass(this._tooltipClass);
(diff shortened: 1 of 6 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -212,20 +223,15 @@ export class MdTooltip implements OnDestroy {
   show(delay: number = this.showDelay): void {
     if (this.disabled || !this._message || !this._message.trim()) { return; }
 
-    if (!this._tooltipInstance) {
-      this._createTooltip();
-    }
-
-    this._setTooltipClass(this._tooltipClass);
-    this._setTooltipMessage(this._message);
+    this._scrollStrategy!.enable();
```
crisbeto: It shouldn't be necessary to enable the scroll strategy manually, it gets enabled/disabled when the overlay is attached and detached.

src/lib/tooltip/tooltip.ts
```diff
@@ -382,19 +436,18 @@ export type TooltipVisibility = 'initial' | 'visible' | 'hidden';
   changeDetection: ChangeDetectionStrategy.OnPush,
   animations: [
     trigger('state', [
-      state('void', style({transform: 'scale(0)'})),
-      state('initial', style({transform: 'scale(0)'})),
+      state('void, initial, hidden', style({transform: 'scale(0)'})),
```
crisbeto: TIL: you can combine states into a single declaration. Neat!

src/lib/tooltip/tooltip.ts
```diff
@@ -382,19 +436,18 @@ export type TooltipVisibility = 'initial' | 'visible' | 'hidden';
...
-    ])
+    ]),
   ],
   host: {
     // Forces the element to have a layout in IE and Edge. This fixes issues where the element
     // won't be rendered if the animations are disabled or there is no web animations polyfill.
     '[style.zoom]': '_visibility === "visible" ? 1 : null',
-    '(body:click)': 'this._handleBodyInteraction()'
+    '(body:click)': '_handleBodyInteraction()',
+    '[attr.aria-hidden]': "isVisible() ? '' : 'true'",
```
crisbeto: I think it's valid if you use only `isVisible()` here. `aria-hidden` is allowed to be true or false.

src/lib/tooltip/tooltip.ts
```diff
@@ -474,18 +530,14 @@ export class TooltipComponent {
...
     return this._onHide.asObservable();
   }
 
-  /**
-   * Whether the tooltip is being displayed
-   */
+  /** Whether the tooltip is being displayed */
   isVisible(): boolean {
-    return this._visibility === 'visible';
+    return this._visibility === 'visible';\
```
crisbeto: Backslash.

src/lib/tooltip/tooltip.ts
jelbourn: It also has to be a string `'true'` or `'false'`

crisbeto: I think it'll be stringified before it gets set to the attribute automatically.

src/lib/tooltip/tooltip.ts
andrewseguin: We'll no longer be creating the tooltip on show, which means the overlay is attached once. After the tooltip is hidden, we disable the scrolling strategy and enable it again on show.

src/lib/tooltip/tooltip.ts
andrewseguin: Changed, good catch. It does get stringified to `'true'` or `'false'`

src/lib/tooltip/tooltip.ts
```diff
@@ -223,12 +225,19 @@ export class MdTooltip implements OnDestroy {
...
+   * Create the tooltip after init. The tooltip should be in the DOM ready to be displayed so that
+   * screen readers immediately have a reference to the tooltip content.
    */
-  ngOnDestroy() {
-    if (this._tooltipInstance) {
-      this._disposeTooltip();
+  ngOnInit() {
+    // Only create the tooltip if we are on the browser platform
+    if (this._platform.isBrowser) {
+      this._createTooltip();
```
jelbourn: Creating the tooltip during init might end up causing a changed-after-checked error because you're attaching a view in the middle of change detection.

andrewseguin:
Tried working around the problem with screenreaders by using `focus` and `focusin` event handlers and creating the tooltip as early as possible, but nothing seems to work 100%. 

To optimize performance after this PR, we can consider re-using tooltips that contain the same message. This means that if we have 100 tooltips that say "Open menu", then only one overlay will be created. 

An issue with this is if the user quickly quickly hovers across several same-messaged tooltips then it would be a matter of figuring out how to move the tooltip intelligently.

andrewseguin:
Closing for now, will revise this to go with another option - create a separate hidden div per message to be used for screenreaders only

PR #5735, opened 2017-07-13
chore: add material-examples to e2e-app

jelbourn:


src/e2e-app/tsconfig-build.json
```diff
@@ -7,7 +7,7 @@
     "experimentalDecorators": true,
     // TODO(paul): Remove once Angular has been upgraded and supports noUnusedParameters in AOT.
     "noUnusedParameters": false,
-    "strictNullChecks": true,
+    // "strictNullChecks": true,
```
DevVersion: Should we add a comment if we temporary disable this?

jelbourn: Good catch; I just commented this out for testing and forgot to come back to it. Removed commented line and added a comment.

PR #5742, opened 2017-07-13
feat(stepper): Merge initial prototype of stepper into the upstream stepper branch.

g1shin:


src/cdk/stepper/index.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {NgModule} from '@angular/core';
+import {CdkStepper} from './stepper';
+import {CommonModule} from '@angular/common';
+import {CdkStep} from './step';
+import {CdkStepLabel} from './step-label';
+import {PortalModule} from '../portal';
```
mmalerba: \n

src/cdk/stepper/step-label.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive, TemplateRef, ViewContainerRef} from '@angular/core';
+import {TemplatePortalDirective} from '../portal';
```
mmalerba: \n

src/cdk/stepper/step.ts
```diff
@@ -0,0 +1,56 @@
...
+    ViewContainerRef
+} from '@angular/core';
+import {CdkStepLabel} from './step-label';
+
+@Component({
+    selector: '[cdk-step]',
+    templateUrl: 'step.html',
+})
+export class CdkStep {
+    @ContentChild(CdkStepLabel) stepLabel: CdkStepLabel;
```
mmalerba: \n (and comment what these are)

src/cdk/stepper/step.ts
```diff
@@ -0,0 +1,56 @@
...
+    @ViewChild(TemplateRef) content: TemplateRef<any>;
+
+    /** Label of the step. */
+    @Input()
+    label: string;
+
+    /** Whether the step is optional or not. */
+    @Input() optional: boolean = false;
```
mmalerba: split into setter & getter, and use `coerceBooleanProperty`

src/cdk/stepper/step.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    /** Whether the step is editable or not. */
+    @Input() editable: boolean = true;
```
mmalerba: use `coerceBooleanProperty`

src/cdk/stepper/step.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    /** Whether the step is the last one in the list. */
+    isLast: boolean = false;
+
+    // /** Whether the step is active. */
```
mmalerba: no longer needed?

src/cdk/stepper/step.ts
```diff
@@ -0,0 +1,56 @@
...
+    label: string;
+
+    /** Whether the step is optional or not. */
+    @Input() optional: boolean = false;
+
+    /** Whether the step is editable or not. */
+    @Input() editable: boolean = true;
+
+    /** Whether the step is the last one in the list. */
+    isLast: boolean = false;
```
mmalerba: is this something we want to be part of the public API? (I asusme probably not, but we can't make it private since its referenced in a template file, so instead check to `_isLast`)

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+@Directive({
+    selector: '[cdkStepper]',
+    host: {
+        '(keydown)': '_onKeydown($event)'
+    },
+})
+export class CdkStepper {
+    @ContentChildren(CdkStep) _steps: QueryList<CdkStep>;
```
mmalerba: comment what these are

src/cdk/stepper/step.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // /** Whether the step is active. */
+    // get active() { return this._active; }
+    // set active(value: boolean) {
+    //     this._active = value;
+    // }
+    // private _active: boolean = false;
+
+    /** Whether the step has been selected. */
+    get selected(): boolean { return this._selected; }
```
mmalerba: are `selected` and `completed` used for anything right now? I think its better to remove them and add them back when you're ready to hook them up

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    @ViewChildren('stepHeader') _stepHeader: QueryList<ElementRef>;
+
+    /** The index of the currently selected step. */
+    @Input()
+    set selectedIndex(value: number) {
+        this._selectedIndex = value;
+    }
+    get selectedIndex(): number { return this._selectedIndex; }
```
mmalerba: nit: getter first, then setter

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+})
+export class CdkStepper {
+    @ContentChildren(CdkStep) _steps: QueryList<CdkStep>;
+
+    @ViewChildren('stepHeader') _stepHeader: QueryList<ElementRef>;
+
+    /** The index of the currently selected step. */
+    @Input()
+    set selectedIndex(value: number) {
+        this._selectedIndex = value;
```
mmalerba: coerceNumberProperty(value)

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    get selectedIndex(): number { return this._selectedIndex; }
+    private _selectedIndex: number;
+
+    /** Optional input to support both linear and non-linear stepper component. */
+    @Input() linear: boolean = true;
```
mmalerba: remove for now (add back if/when you hook it up to something)

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    /** Output to enable support for two-way binding on `[(selectedIndex)]` */
+    @Output() get selectedIndexChange(): Observable<number> {
```
mmalerba: isn't the index part of the `CdkStepEvent` already emitted by `this.stepEvent`? why do we need this as well?

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+        return map.call(this.stepEvent, event => event.index);
+    }
+
+    // @Output() get focusIndexChange(): Observable<number> {
```
mmalerba: no longer needed?

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+import {CdkStep} from './step';
+import {Observable} from 'rxjs/Observable';
+import {map} from 'rxjs/operator/map';
+import {LEFT_ARROW, RIGHT_ARROW, ENTER, TAB} from '../keyboard/keycodes';
+
+/** Used to generate unique ID for each stepper component. */
+let nextId = 0;
+
+/** Change event emitted on focus or selection changes. */
+export class CdkStepEvent {
```
mmalerba: `CdkStepperSelectionEvent`?

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+    @Output() get selectedIndexChange(): Observable<number> {
+        return map.call(this.stepEvent, event => event.index);
+    }
+
+    // @Output() get focusIndexChange(): Observable<number> {
+    //     return map.call(this.focusChange, event => event.index);
+    // }
+
+    /** Event emitted when the selected step has changed. */
+    @Output() stepEvent = new EventEmitter<CdkStepEvent>();
```
mmalerba: I would call this `selectionChange`, `stepEvent` isn't very descriptive

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    /** Event emitted when the focused step has changed. */
+    @Output() focusChange = new EventEmitter<CdkStepEvent>();
```
mmalerba: do we need this? I would think the developer can manage this themselves with `(focus)` and `(blur)` events

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+    private _focusIndex: number = 0;
+
+    private _groupId: number;
+
+    constructor() {
+        this._groupId = nextId++;
+    }
+
+    /** Selects and focuses the provided step. */
+    selectStep(step: CdkStep): void {
```
mmalerba: Is this similar to what tabs does? From an API perspective I would rather `CdkStep` have a `select` method, but I could see maybe not wanting a dependency from child to parent

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+    @Output() focusChange = new EventEmitter<CdkStepEvent>();
+
+    /** The step that is currently selected. */
+    get selectedStep(): CdkStep {
+        return this._steps.toArray()[this._selectedIndex];
+    }
+    private _selectedStep: CdkStep;
+
+    /** The index of the step that the focus is currently on. */
+    get focusIndex(): number {return this._focusIndex; }
```
mmalerba: do we need a public API for this? seems like an internal detail

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+    selectStep(step: CdkStep): void {
+        let stepsArray = this._steps.toArray();
+        this._selectedIndex = stepsArray.indexOf(step);
+        this.stepEvent.emit(this._emitStepEvent(this._selectedIndex));
+        this._focusIndex = this._selectedIndex;
+        this._setStepFocus();
+    }
+
+    /** Selects and focuses the next step in list. */
+    nextStep(): void {
```
mmalerba: I would just call it `next` and `previous`

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+        event.index = index;
+        event.step = this._steps.toArray()[this._selectedIndex];
+        this._selectedStep = event.step;
+        return event;
+    }
+
+    _onKeydown(event: KeyboardEvent) {
+        switch (event.keyCode) {
+            case RIGHT_ARROW:
+                if (this._focusIndex != this._steps.length - 1) {
```
mmalerba: just call `nextStep`?

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,153 @@
...
+    _getStepLabelId(i: number): string {
+        return `mat-step-label-${this._groupId}-${i}`;
+    }
+
+    /** Returns a unique id for each step content element. */
+    _getStepContentId(i: number): string {
+        return `mat-step-content-${this._groupId}-${i}`;
+    }
+
+    private _emitStepEvent(index: number): CdkStepEvent {
```
mmalerba: name is deceptive since this method doesn't actually cause an emit

src/cdk/stepper/stepper.ts
g1shin: No, it seems like tabs doesn't have a separate method for selecting. It just updates and sets the `selectedIndex` in the html file. Since the `CdkStepper` already takes care of all the logic regarding the selection and `selectedIndex` tracking, I was thinking about just getting rid of the `selected` attribute completely from `CdkStep`.

src/cdk/stepper/stepper.ts
g1shin: This wouldn't be equivalent to `nextStep` because `RIGHT_ARROW` does not mean that the next `step` has been chosen. Only the focus has changed. The user would need to press `ENTER` to make the actual selection change.

g1shin:
Changes made; ready for review again.

src/cdk/stepper/step.ts
```diff
@@ -0,0 +1,48 @@
...
+export class CdkStep {
+  /** Template for step label if it exists. */
+  @ContentChild(CdkStepLabel) stepLabel: CdkStepLabel;
+
+  /** Template for step content. */
+  @ViewChild(TemplateRef) content: TemplateRef<any>;
+
+  /** Label of the step. */
+  @Input()
+  label: string;
```
mmalerba: does tabs do anything to validate that the user provided 1 and only 1 of: `stepLabel`, `label`. or does it just ignore the text label if a template label was provided?

src/cdk/stepper/stepper.ts
mmalerba: I'm saying that as an API I prefer: `step.select()` to `stepper.select(step)`, but this would require the step to know about its parent stepper, so maybe not worth it. @jelbourn do you have any opinion?

src/cdk/stepper/step.ts
g1shin: Tab checks the same way as it's done for the stepper in stepper-horizontal.html and stepper-vertical.html. If `stepLabel` exists, it will use the template; if `stepLabel` does not exist, it will use the text label.

src/demo-app/stepper/stepper-demo.html
```diff
@@ -0,0 +1,8 @@
+<h2>Horizontal Stepper Demo</h2>
+<mat-horizontal-stepper [(selectedIndex)]="horizontalActiveIndex">
+    <mat-step *ngFor="let step of steps" [label]="step.label">
```
mmalerba: nit: indent 2

src/lib/stepper/step.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import {Component, ContentChild, TemplateRef, ViewChild} from '@angular/core';
+import {CdkStep} from '@angular/cdk';
+import {MdStepLabel} from './step-label';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-step, mat-step',
+  templateUrl: 'step.html',
+  inputs: ['label'],
```
mmalerba: `'optional'` and `'editable'` too

src/cdk/stepper/step-label.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Directive({
+  selector: '[cdk-step-label]',
+})
+export class CdkStepLabel extends TemplatePortalDirective {
```
mmalerba: I think you can do this the same way you did `CdkStep` (just a `TemplateRef` and `ngTemplateOutlet` tabs used the portal stuff because it was written before `ngTemplateOutlet` was a thing

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,139 @@
...
+  get selectedIndex() { return this._selectedIndex; }
+  set selectedIndex(value: any) {
+    this._selectedIndex = coerceNumberProperty(value);
+  }
+  private _selectedIndex: number;
+
+  /** Event emitted when the selected step has changed. */
+  @Output() selectionChange = new EventEmitter<CdkStepperSelectionEvent>();
+
+  /** The index of the step that the focus is currently on. */
```
g1shin: It is the index of the step that focus is on.

g1shin:
Changes made. Ready for review :)

src/demo-app/stepper/stepper-demo.html
```diff
@@ -0,0 +1,27 @@
+<h2>Horizontal Stepper Demo</h2>
```
mmalerba: These are ok for now, but we should eventually add examples that are more like what a developer would actually want to create.

src/lib/stepper/step-label.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import {Directive, TemplateRef} from '@angular/core';
+import {CdkStepLabel} from '@angular/cdk';
+
+@Directive({
+  selector: '[md-step-label], [mat-step-label]',
+})
+export class MdStepLabel extends CdkStepLabel {
+  constructor(public template: TemplateRef<any>) {
```
mmalerba: I don't _think_ you need the `public` here since it's on the super class (but I could be wrong)

src/lib/stepper/stepper-horizontal.html
```diff
@@ -0,0 +1,38 @@
...
+             [tabIndex]="_focusIndex == i ? 0 : -1"
+             (click)="select(step)"
+             (keydown)="_onKeydown($event)">
+            <div *ngIf="step.active" class="active-step">{{i + 1}}</div>
+            <div *ngIf="!step.active" class="inactive-step">{{i + 1}}</div>
+
+            <div class="mat-step-label">
+                <!-- If there is a label template, use it. -->
+                <ng-template [ngIf]="step.stepLabel">
+                    <ng-container [ngTemplateOutlet]="step.stepLabel.template"></ng-container>
```
mmalerba: shortcut syntax: `<ng-container *ngIf="...`

src/lib/stepper/stepper-horizontal.html
```diff
@@ -0,0 +1,38 @@
...
+                </ng-template>
+                <!-- It there is no label template, fall back to the text label. -->
+                <ng-template [ngIf]="!step.stepLabel">
+                    <div>{{step.label}}</div>
```
mmalerba: `<div *ngIf="...`

src/lib/stepper/stepper-horizontal.html
```diff
@@ -0,0 +1,38 @@
+<div class="mat-stepper-container">
```
mmalerba: nit: indent 2

mmalerba: `<div>`s in this file do not seem to be properly matched

src/lib/stepper/stepper-vertical.html
```diff
@@ -0,0 +1,37 @@
...
+    <div *ngFor="let step of _steps; let i = index" class="mat-step" role="tablist">
+        <div #stepHeader class="mat-step-header" role="tab"
+             [id]="_getStepLabelId(i)"
+             [attr.aria-controls]="_getStepContentId(i)"
+             [attr.aria-selected]="selectedIndex == i"
+             [tabIndex]="_focusIndex == i ? 0 : -1"
+             (click)="select(step)"
+             (keydown)="_onKeydown($event)">
+            <div *ngIf="step.active" class="active-step">{{i + 1}}</div>
+            <div *ngIf="!step.active" class="inactive-step">{{i + 1}}</div>
```
kara: Seems like you could use ngClass here, given that that's the only thing changing on the conditional.

```html
<div [ngClass]="{'active-step': step.active, 'inactive-step': !step.active}">{{ i + 1}} </div>
```

src/lib/stepper/stepper-vertical.html
```diff
@@ -0,0 +1,37 @@
...
+                    <ng-template [cdkPortalHost]="step.stepLabel"></ng-template>
+                </ng-template>
+                <!-- It there is no label template, fall back to the text label. -->
+                <ng-template [ngIf]="!step.stepLabel">
+                    <div>{{step.label}}</div>
+                </ng-template>
+            </div>
+
+        </div>
+        <div *ngIf="!step._isLast" class="connector-line"></div>
```
kara: Is `step._isLast` simply referring to index here? If so, you could use the built-in prop from ngFor on line 2 rather than creating a property for it:

```html
<div *ngFor="let step of _steps; let i = index; let isLast = last" class="mat-step" role="tablist">
...
<div *ngIf="isLast" class="connector-line"></div>
```

src/lib/stepper/stepper-vertical.html
```diff
@@ -0,0 +1,37 @@
...
+        <div *ngIf="i == selectedIndex" role="tabpanel"
+             [id]="_getStepContentId(i)"
+             [attr.aria-labelledby]="_getStepLabelId(i)"
+             [attr.aria-expanded]="selectedIndex == i">
+            <ng-container [ngTemplateOutlet]="step.content"></ng-container>
+        </div>
+        <div *ngIf="i == selectedIndex">
+            <button md-button (click)="previous()">Back</button>
+            <button md-button (click)="next()">Next</button>
+        </div>
```
kara: Given that this div and the one on line 26 both have this condition, could you collapse them into one outer div (or ng-container) with the ngIf condition?

src/cdk/stepper/step-label.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive, TemplateRef} from '@angular/core';
+
+@Directive({
+  selector: '[cdk-step-label]',
```
kara: @jelbourn Should this be `cdkStepLabel` since it's a directive?

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,148 @@
...
+  _focusIndex: number = 0;
+
+  private _groupId: number;
+
+  constructor() {
+    this._groupId = nextId++;
+  }
+
+  /** Selects and focuses the provided step. */
+  select(step: CdkStep | number): void {
```
jelbourn: Why not both?

src/cdk/stepper/step-label.ts
jelbourn: Yes

src/cdk/stepper/stepper.ts
g1shin: I see. What you mentioned earlier about `_focusIndex` seems to be the accurate and intended definition of this variable. I'll change the comment to `/** The index of the step that the focus can be set. */`

src/demo-app/stepper/stepper-demo.ts
```diff
@@ -0,0 +1,19 @@
+import {Component} from '@angular/core';
+
+@Component({
+  moduleId: module.id,
+  selector: 'stepper-demo',
+  templateUrl: 'stepper-demo.html',
+  styleUrls: ['stepper-demo.scss'],
+})
+export class StepperDemo {
+  verticalActiveIndex = 0;
```
g1shin: It is used now.

src/demo-app/stepper/stepper-demo.ts
```diff
@@ -0,0 +1,19 @@
...
+  horizontalActiveIndex = 0;
+  labelTemplateIndex = 0;
```
g1shin: This is also used now.

g1shin:
Ready for review again.

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,178 @@
...
+    return `mat-step-label-${this._groupId}-${i}`;
+  }
+
+  /** Returns nique id for each step content element. */
+  _getStepContentId(i: number): string {
+    return `mat-step-content-${this._groupId}-${i}`;
+  }
+
+  private _emitStepperSelectionEvent(newIndex: number,
+                                       oldIndex: number): void {
```
mmalerba: is `oldIndex` not just always whatever `this._selectedIndex` at the start of this method?

src/cdk/stepper/stepper.ts
mmalerba: Actually now that I think about it, why don't we just change the `@Input()` from `selectedIndex` to `selected` and have it call this method (which we will make private). Then all the selection stuff is driven through the input.

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,178 @@
...
+    event.oldIndex = oldIndex;
+    event.oldStep = this._steps.toArray()[oldIndex];
+    event.newStep = this._steps.toArray()[this._selectedIndex];
+    this.selectionChange.emit(event);
+  }
+
+  _onKeydown(event: KeyboardEvent) {
+    switch (event.keyCode) {
+      case RIGHT_ARROW:
+        if (this._focusIndex != this._steps.length - 1) {
```
mmalerba:
```ts
this._setStepFocused((this._focusIndex + 1) % this._steps.length);
```

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,178 @@
...
+          this._setStepFocused(this._focusIndex + 1);
+        } else {
+          this._setStepFocused(0);
+        }
+        break;
+      case LEFT_ARROW:
+        if (this._focusIndex != 0) {
```
mmalerba:
```ts
this._setStepFocused((this._focusIndex + this._steps.length - 1) % this._steps.length);
```

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,178 @@
...
+          this._setStepFocused(this._focusIndex - 1);
+        } else {
+          this._setStepFocused(this._steps.length - 1);
+        }
+        break;
+      case SPACE:
+      case ENTER:
+        this._emitStepperSelectionEvent(this._focusIndex, this._selectedIndex);
+        break;
+      default:
```
mmalerba: comment why we do this (so we don't preventDefaault on keys we don't handle)

src/cdk/stepper/stepper.ts
g1shin: @mmalerba Would the `selected` still be returning the `this._selectedIndex` since the stepper cannot have the boolean `selected` attribute for each individual step?

mmalerba: oh hmm I didn't think about the getter return type, good point. Ok how about something like this:
```ts
get selected() { return this._steps[this.selectedIndex]; }
set selected(step: CdkStep) { 
  let index = ...;
  this.selectedIndex = index;
}

get selectedIndex() { return this._selectedIndex; }
set selectedIndex(index: number) {
  // do actual selecting logic
  this._selectedIndex = index;
}
```

g1shin:
Ready for review again.

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,170 @@
...
+  private _groupId: number;
+
+  constructor() {
+    this._groupId = nextId++;
+  }
+
+  /** Selects and focuses the next step in list. */
+  next(): void {
+    if (this._selectedIndex == this._steps.length - 1) { return; }
+    this.selectedIndex++;
```
jelbourn:
```ts
this._selectedIndex = Math.min(this._selectedIndex + 1, this._steps.length - 1);
```

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  /** Selects and focuses the previous step in list. */
+  previous(): void {
+    if (this._selectedIndex == 0) { return; }
+    this.selectedIndex--;
```
jelbourn:
```ts
this._selectedIndex = Math.max(this._selectedIndex - 1, 0);
```

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  private _emitStepperSelectionEvent(newIndex: number): void {
+    const event = new CdkStepperSelectionEvent();
+    event.oldIndex = this._selectedIndex;
+    event.newIndex = newIndex;
+    let stepsArray = this._steps.toArray();
+    event.oldStep = stepsArray[this._selectedIndex];
+    event.newStep = stepsArray[newIndex];
+    this._selectedIndex = newIndex;
+    this.selectionChange.emit(event);
```
jelbourn: TypeScript is structurally typed, meaning that something is considered the right type if it _matches_ the type. As such, you can change this whole function to:
```ts
const stepsArray = this._steps.toArray();
this.selectionChange.emit({
  selectedIndex: newIndex,
  previouslySelectedIndex: this._selectedIndex,
  selectedStep: stepsArray[newIndex],
  previouslySelectedStep: stepsArray[this._selectedIndex],
});
```

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  _onKeydown(event: KeyboardEvent) {
+    switch (event.keyCode) {
+      case RIGHT_ARROW:
+        this._setStepFocused((this._focusIndex + 1) % this._steps.length);
```
jelbourn: Can you file an issue on github to track creating something like a `TabsKeyManager` to reduce duplication between tabs and stepper?

src/lib/stepper/stepper-horizontal.html
```diff
@@ -0,0 +1,35 @@
...
+       [tabIndex]="_focusIndex == i ? 0 : -1"
+       (click)="step.select()"
+       (keydown)="_onKeydown($event)">
+    <div class="mat-stepper-index">
+      {{i + 1}}
+    </div>
+
+    <div class="mat-stepper-label">
+      <!-- If there is a label template, use it. -->
+      <ng-container *ngIf="step.stepLabel"[ngTemplateOutlet]="step.stepLabel.template">
```
jelbourn: Missing space between attributes

src/lib/stepper/stepper-horizontal.html
```diff
@@ -0,0 +1,35 @@
...
+<div *ngFor="let step of _steps; let i = index"
+     class="mat-stepper-content" role="tabpanel"
+     [id]="_getStepContentId(i)"
+     [attr.aria-labelledby]="_getStepLabelId(i)"
+     [attr.aria-expanded]="selectedIndex == i">
+  <ng-container [ngTemplateOutlet]="step.content"></ng-container>
+</div>
+<div>
+  <button md-button (click)="previous()">Back</button>
+  <button md-button (click)="next()">Next</button>
```
jelbourn: Are these buttons just here for testing / development?

src/lib/stepper/stepper-horizontal.ts
```diff
@@ -0,0 +1,28 @@
...
+  moduleId: module.id,
+  selector: 'md-horizontal-stepper, mat-horizontal-stepper',
+  templateUrl: 'stepper-horizontal.html',
+  styleUrls: ['stepper.scss'],
+  inputs: ['selectedIndex'],
+  host: {
+    'class': 'mat-stepper-horizontal',
+    'role': 'tablist',
+  },
+  providers: [{ provide: MdStepper, useExisting: forwardRef(() => MdHorizontalStepper) }]
```
jelbourn: Is the forwardRef necessary here?

src/lib/stepper/stepper-vertical.html
```diff
@@ -0,0 +1,33 @@
...
+  </div>
+  <div *ngIf="!isLast" class="connector-line"></div>
+  <div class="mat-stepper-content" role="tabpanel"
+       [id]="_getStepContentId(i)"
+       [attr.aria-labelledby]="_getStepLabelId(i)"
+       [attr.aria-expanded]="selectedIndex == i">
+    <ng-container [ngTemplateOutlet]="step.content"></ng-container>
+    <div>
+      <button md-button (click)="previous()">Back</button>
+      <button md-button (click)="next()">Next</button>
```
jelbourn: Same question for these buttons

src/lib/stepper/stepper-horizontal.html
g1shin: This PR doesn't include stepper-button directives that will be used by users to add the buttons themselves. The stepper component in this initial PR inserts the buttons instead. The stepper-button directives will come in the next PR since this is already too big, and these buttons will be removed in that PR as well.

src/lib/stepper/stepper-vertical.html
g1shin: Same plan for these buttons as well.

src/cdk/stepper/stepper.ts
g1shin: #5916

g1shin:
Changes have been made; ready for review.

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (this._selectedIndex != index) {
+      this._emitStepperSelectionEvent(index);
+      this._setStepFocused(this._selectedIndex);
+    }
+  }
+  private _selectedIndex: number = 0;
+
+  /** Returns the step that is selected. */
+  get selected() { return this._steps[this.selectedIndex]; }
+  /** Sets selectedIndex as the index of the provided step. */
```
mmalerba: nit: can just put 1 comment for set & get since they represent the same property

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,169 @@
...
+})
+export class CdkStepper {
+  /** The list of step components that the stepper is holding. */
+  @ContentChildren(CdkStep) _steps: QueryList<CdkStep>;
+
+  /** The list of step headers of the steps in the stepper. */
+  _stepHeader: QueryList<ElementRef>;
+
+  /** The index of the selected step. */
+  get selectedIndex() { return this._selectedIndex; }
```
mmalerba: we want this to be an `@Input()` right?

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,169 @@
...
+  set selectedIndex(index: number) {
+    if (this._selectedIndex != index) {
+      this._emitStepperSelectionEvent(index);
+      this._setStepFocused(this._selectedIndex);
+    }
+  }
+  private _selectedIndex: number = 0;
+
+  /** Returns the step that is selected. */
+  get selected() { return this._steps[this.selectedIndex]; }
```
mmalerba: same, should be `@Input()`

src/lib/stepper/stepper-horizontal.html
```diff
@@ -0,0 +1,35 @@
...
+
+    <div class="mat-stepper-label">
+      <!-- If there is a label template, use it. -->
+      <ng-container *ngIf="step.stepLabel" [ngTemplateOutlet]="step.stepLabel.template">
+      </ng-container>
+      <!-- It there is no label template, fall back to the text label. -->
+      <div *ngIf="!step.stepLabel">{{step.label}}</div>
+    </div>
+
+    </ng-template>
```
mmalerba: don't see a corresponding opening tag for this?

src/lib/stepper/stepper-horizontal.html
```diff
@@ -0,0 +1,35 @@
+<div *ngFor="let step of _steps; let i = index; let isLast = last">
```
mmalerba: why do we need this wrapper div, is it supposed to contain all of the headers?

src/cdk/stepper/stepper.ts
```diff
@@ -0,0 +1,169 @@
...
+        this._emitStepperSelectionEvent(this._focusIndex);
+        break;
+      default:
+        // Return to avoid calling preventDefault on keys that are not explicitly handled.
+        return;
+    }
+    event.preventDefault();
+  }
+
+  private _setStepFocused(index: number) {
```
jelbourn: Since you're actually calling focus, I'd just call this method `_focusStep`

src/lib/stepper/stepper-horizontal.html
jelbourn: Sounds good, just add a comment mentioning that

src/lib/stepper/stepper-vertical.ts
```diff
@@ -0,0 +1,28 @@
...
+  moduleId: module.id,
+  selector: 'md-vertical-stepper, mat-vertical-stepper',
+  templateUrl: 'stepper-vertical.html',
+  styleUrls: ['stepper.scss'],
+  inputs: ['selectedIndex'],
+  host: {
+    'class': 'mat-stepper-vertical',
+    'role': 'tablist',
+  },
+  providers: [{ provide: MdStepper, useExisting: MdVerticalStepper }]
```
jelbourn: Omit spaces inside braces, e.g.,
```ts
[{provide: MdStepper, useExisting: MdVerticalStepper}]
```

g1shin:
Ready for review again.

src/demo-app/stepper/stepper-demo.html
```diff
@@ -0,0 +1,27 @@
...
+<mat-horizontal-stepper>
+  <mat-step *ngFor="let step of steps" [label]="step.label">
+    <md-input-container>
+      <input mdInput placeholder="Answer" [(ngModel)]="step.content">
+    </md-input-container>
+  </mat-step>
+</mat-horizontal-stepper>
```
jelbourn: Just noticed that the stepper selectors need to be added to the compatibility mode list as well
(which enforces that you use either `md-` or `mat-` and not both)

src/lib/stepper/step.ts
```diff
@@ -0,0 +1,26 @@
...
+import {MdStepLabel} from './step-label';
+import {MdStepper} from './stepper';
+
+@Component({
+  moduleId: module.id,
+  selector: 'md-step, mat-step',
+  templateUrl: 'step.html',
+})
+export class MdStep extends CdkStep {
+  /** Content for the step label given by <ng-template mat-step-label>. */
```
jelbourn: Select in comment still uses dashes

src/lib/stepper/stepper-horizontal.ts
```diff
@@ -0,0 +1,28 @@
...
+  inputs: ['selectedIndex'],
+  host: {
+    'class': 'mat-stepper-horizontal',
+    'role': 'tablist',
+  },
+  providers: [{provide: MdStepper, useExisting: MdHorizontalStepper}]
+})
+export class MdHorizontalStepper extends MdStepper {
+  /** Steps that the horizontal stepper holds. */
+  @ContentChildren(MdStep) _steps: QueryList<MdStep>;
```
jelbourn: Both `MdHorizontalStepper` and `MdVerticalStepper` have 
```ts
@ContentChildren(MdStep) _steps: QueryList<MdStep>;
```

Could that go in the `MdStepper` base class?

src/lib/stepper/stepper.scss
```diff
@@ -0,0 +1,7 @@
+.mat-stepper-content[aria-expanded='false'] {
+  display:none;
```
jelbourn: Missing space

src/lib/stepper/stepper-horizontal.ts
g1shin: Yes, I moved `_steps` to `MdStepper` base class. However, this resulted in circular dependency between `MdStepper` and `MdStep`, so I moved MdStep into stepper.ts as it's done for `CdkStep`.

g1shin:
Thank you for all the comments on this big PR. Comments were addressed; ready for another review.

PR #5746, opened 2017-07-13
fix(datepicker): remove aria-expanded on datepicker input ...

### LINKED ISSUE #5729: Datepicker puts `aria-owns` on input pointing to element not in the DOM
`MdDatepickerInput` sets `aria-owns` on the input element to the ID of the datepicker pop-up, even when the datepicker is not in the DOM. It is invalid for an element to have an `aria-owns` that references an ID not in the page.

### LINKED ISSUE #5730: Datepicker input should not apply aria-expanded
Via https://www.w3.org/TR/wai-aria-1.1/#aria-expanded, `aria-expanded` is not valid on text inputs.

(axe catches this when I tried to add an e2e test for the datepicker)

mmalerba:
since the attribute is not supported for text inputs. also only set aria-owns
attribute when the datepicker is open and therefore actually in the DOM.

fixes #5730 
fixes #5729

### DIFF
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -57,9 +57,8 @@ export const MD_DATEPICKER_VALIDATORS: any = {
   selector: 'input[mdDatepicker], input[matDatepicker]',
   providers: [MD_DATEPICKER_VALUE_ACCESSOR, MD_DATEPICKER_VALIDATORS],
   host: {
-    '[attr.aria-expanded]': '_datepicker?.opened || "false"',
     '[attr.aria-haspopup]': 'true',
-    '[attr.aria-owns]': '_datepicker?.id',
+    '[attr.aria-owns]': '_datepicker?.opened && _datepicker.id',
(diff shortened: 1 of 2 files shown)


src/lib/datepicker/datepicker-input.ts
```diff
@@ -57,9 +57,8 @@ export const MD_DATEPICKER_VALIDATORS: any = {
   selector: 'input[mdDatepicker], input[matDatepicker]',
   providers: [MD_DATEPICKER_VALUE_ACCESSOR, MD_DATEPICKER_VALIDATORS],
   host: {
-    '[attr.aria-expanded]': '_datepicker?.opened || "false"',
     '[attr.aria-haspopup]': 'true',
-    '[attr.aria-owns]': '_datepicker?.id',
+    '[attr.aria-owns]': '_datepicker?.opened && _datepicker.id',
```
jelbourn: Add a unit test for this?

mmalerba:
done, and caught a bug \o/

PR #5747, opened 2017-07-14
fix(datepicker): force Intl.DateTimeFormat to use UTC time zone ...

### LINKED ISSUE #5185: Datepicker: Incorrect date displayed when selecting a date in the first week of November 1999
Bug

When selecting November, 4, 1999 from the datepicker, the input will display 11/04/1999

When selecting November, 4, 1999 from the datepicker the input displays 1 day behind, 11/03/1999

Select November 4, 1999 from the datepicker in the supplied plunkr:
http://plnkr.co/edit/4kiAVKRyskrJFxHNZLuy

It's a bug, the existing behavior is unintended

Not sure, just something I noticed on the Material documentation site.  I'm using Chrome 58.0.3029.110 as my browser 

This happens for the first week of November in both 1999 and 2000, likely other places as well.

mmalerba:
so it doesn't display incorrect dates.

fixes #5185

@julianobrasil Thanks for tracking down the root issue!

### DIFF
--- a/src/lib/core/datetime/native-date-adapter.ts
+++ b/src/lib/core/datetime/native-date-adapter.ts
@@ -8,6 +8,7 @@
 
 import {Inject, Injectable, Optional, LOCALE_ID} from '@angular/core';
 import {DateAdapter} from './date-adapter';
+import {extendObject} from '../util/object-extend';
 
 
 // TODO(mmalerba): Remove when we no longer support safari 9.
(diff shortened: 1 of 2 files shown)


crisbeto:
@mmalerba seems like there's a linting error and the output format is different on iOS Safari.

mmalerba:
lint + tests fixed

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -154,6 +162,12 @@ export class NativeDateAdapter extends DateAdapter<Date> {
 
   format(date: Date, displayFormat: Object): string {
     if (SUPPORTS_INTL_API) {
+      if (this.useUtcForDisplay) {
+        date = new Date(Date.UTC(
+            date.getFullYear(), date.getMonth(), date.getDate(), date.getHours(),
+            date.getMinutes(), date.getSeconds(), date.getMilliseconds()));
+        displayFormat = Object.assign({}, displayFormat, {timeZone: 'utc'});
```
kara: `Object.assign` is breaking in Google. Can you switch to our util `extendObject`?

mmalerba:
fixed

PR #5751, opened 2017-07-14
feat(radio): add required attribute to radio-group

### LINKED ISSUE #1656: [md-radio] required attribute doesn't work for md-radio-group or md-radio-button

bug, (or feature request)

`required` attribute is available in `md-radio-group` or `md-radio-button`,
i.e., when a user clicks submit button in a form without selecting any required radio options,
the form should tell the user that (s)he needs to select one of those.

refs. w3c description about "input required"
https://www.w3.org/TR/html5/forms.html#attr-input-required

`required` attribute is ignored without any errors or warnings.

Visit the following plunker page:
http://plnkr.co/edit/6Y2L1cLwYX2TOXv5UDOu?p=preview

In the plunker, we have 2 forms, one with naive html5 radio buttons and the other with md-radio-group and md-radio-buttons.
In the upper form, we cannot 'save' without selecting any options,
but in the lower one, it seems we can 'save' without selecting anything.

Client side validation same as md-input or naive `input type='radio'`.

angular: 2.0.2
material: 2.0.0-alpha.9-3
os: OSX El Capitan 10.11.6
browser: Google Chrome 54.0.2840.71 (64-bit)

rafaelss95:
Closes #1656

### DIFF
--- a/src/demo-app/radio/radio-demo.html
+++ b/src/demo-app/radio/radio-demo.html
@@ -22,9 +22,19 @@
     </button>
   </div>
   <div>
+    <span>isRequired: {{isRequired}}</span>
+    <button md-raised-button (click)="isRequired=!isRequired" class="demo-button">
+      Require buttons
+    </button>
+  </div>
+  <div>
     <span><md-checkbox [(ngModel)]="isAlignEnd">Align end</md-checkbox></span>
   </div>
-  <md-radio-group name="my_options" [disabled]="isDisabled" [align]="isAlignEnd ? 'end' : 'start'">
+  <md-radio-group
+    name="my_options"
+    [disabled]="isDisabled"
+    [required]="isRequired"
+    [align]="isAlignEnd ? 'end' : 'start'">
     <md-radio-button value="option_1">Option 1</md-radio-button>
     <md-radio-button value="option_2">Option 2</md-radio-button>
     <md-radio-button value="option_3">Option 3</md-radio-button>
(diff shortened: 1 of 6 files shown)


rafaelss95:
Initially, in addition to add `required` attribute in `radio-group`, I would also add it to `radio-button`, but this could cause some conflicts, such as:

```
<md-radio-group required="true">
  <md-radio-button value="any" required="false"></md-radio-button> <- HERE
</md-radio-group>
```
What do you think @jelbourn ?

rafaelss95:
I have added `required` attribute in `radio-button`, but I just noticed that it could cause some inconsistencies, as in this example:

```
<md-radio-group required="true" name="any">
  <md-radio-button value="any" required="false">any</md-radio-button> <- HERE
</md-radio-group>
```
Note that in this case it is possible to set `required="true"` to `md-radio-group` and at the same time `required="false"` on one or **even all** the radio buttons inside it.

But in this case, the `required` from `md-radio-group` will override any `required` attribute in inside radio-buttons due to the condition:

```
get required(): boolean {
  return this._required || (this.radioGroup && this.radioGroup.required);
}
```

It means that if the `md-radio-group` is required, all `md-radio-button`'s inside it, will be required, no matter what you've set in the `required` attribute in `md-radio-button`'s.

I'm just wondering if it's the correct approach. Any thoughts @jelbourn?

jelbourn:
@rafaelss95 could you rebase this PR?

rafaelss95:
@jelbourn Done.

Note that the the failed checks are unrelated to this PR. Btw, it should be fixed in #5945.

rafaelss95:
@jelbourn Can you apply the labels here, please?

PR #5760, opened 2017-07-14
style(chips): clean up alignment and theming

willshowell:
- Move size/positioning of remove icon out of demo and into chip styles
- Increase between chip margin to 8px to match spec
- Vertically align text and remove icon
- Make cursor default to match spec
- Add focus hover to match spec (in lieu of default outline)
- Clean up superfluous theming
- Remove non-themed selected styles since primary is the default color option
- Fix bug where a single input chip had no following margin

--------
screen shot 2017-07-14 at 10 49 43 am
screen shot 2017-07-14 at 10 49 58 am

### DIFF
--- a/src/demo-app/chips/chips-demo.scss
+++ b/src/demo-app/chips/chips-demo.scss
@@ -24,12 +24,4 @@
   md-chip-list input {
     width: 150px;
   }
-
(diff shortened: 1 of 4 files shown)


jelbourn:
@tinayuangao should also take a quick look

PR #5764, opened 2017-07-14
fix(OverlayContainer): themeClass only remove class if not empty

joejordanbrown:
If _themeClass is empty do not try and remove from the element classList.


jelbourn:

PR #5769, opened 2017-07-14
fix(dialog): prevent dialog from opening while another dialog is animating

### LINKED ISSUE #5713: Bug: Focus still on "open dialog" button after clicked

crisbeto:
Prevents dialogs from being opened while other dialogs are animating.

Fixes #5713.

**Note:** I wasn't able to write a test that properly captures the issue that this is solving (I tried using the `BrowserAnimationsModule` instead of the noop module as well). In unit tests, it seems like the callbacks are fired immediately, whereas we have the animations disabled in the e2e tests in general.

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -91,11 +91,19 @@ export class MdDialog {
    */
   open<T>(componentOrTemplateRef: ComponentType<T> | TemplateRef<T>,
           config?: MdDialogConfig): MdDialogRef<T> {
+
+    const inProgressDialog = this._openDialogs.find(dialog => dialog._isAnimating());
+
+    // If there's a dialog that is in the process of being opened, return it instead.
+    if (inProgressDialog) {
+      return inProgressDialog;
+    }
+
     config = _applyConfigDefaults(config);
 
-    let overlayRef = this._createOverlay(config);
-    let dialogContainer = this._attachDialogContainer(overlayRef, config);
-    let dialogRef =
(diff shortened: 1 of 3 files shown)


src/lib/dialog/dialog.ts
```diff
@@ -91,11 +91,19 @@ export class MdDialog {
    */
   open<T>(componentOrTemplateRef: ComponentType<T> | TemplateRef<T>,
           config?: MdDialogConfig): MdDialogRef<T> {
+
+    const inProgressDialog = this._openDialogs.find(dialog => dialog._isAnimating());
```
jelbourn: Do you think it's worthwhile to key this on the component as well? E.g., if I try to `open(UserProfile)` multiple times in a row, they get blocked, but if I try to `open(UserProfile)` and `open(AdminSettings)` at the same time then they both open?

crisbeto: I see how that would be useful, but it's probably too much trouble. There's only a 400ms window after you open a dialog where this can happen.

jelbourn: Let's go with this for now and see if we get any feedback.

PR #5773, opened 2017-07-14
fix(input): add overflow:hidden when calculating autosize height

### LINKED ISSUE #4948: MdTextareaAutosize: textarea control not sized correctly when number of lines is an odd number
Bug:  When using the MdTextareaAutosize with a textarea and the number of lines that the text occupies is an odd number, the control size is slightly too small and scroll bars appear.

The control should be resized such that scroll bars never appear.

Described above in the bug.

See plunker [here](http://plnkr.co/edit/fBa8kLedcDGoNvnDTOsr?p=preview)

Angular: 4.1.0
Material: 2.0.0-beta.6
TypeScript: 2.3.2

mmalerba:
fixes #4948

### DIFF
--- a/src/lib/input/autosize.ts
+++ b/src/lib/input/autosize.ts
@@ -136,10 +136,13 @@ export class MdTextareaAutosize implements AfterViewInit {
     }
 
     // Reset the textarea height to auto in order to shrink back to its default size.
+    // Also temporarily force overflow:hidden, so scroll bars do not interfere with calculations.
     textarea.style.height = 'auto';
+    textarea.style.overflow = 'hidden';
 
     // Use the scrollHeight to know how large the textarea *would* be if fit its entire value.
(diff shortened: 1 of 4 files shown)


src/lib/input/autosize.ts
```diff
@@ -136,10 +136,13 @@ export class MdTextareaAutosize implements AfterViewInit {
     }
 
     // Reset the textarea height to auto in order to shrink back to its default size.
+    // Also temporarily force overflow:hidden, so scroll bars do not interfere with calculations.
     textarea.style.height = 'auto';
+    textarea.style.overflow = 'hidden';
```
jelbourn: Is there a way to write a unit test for this?

mmalerba: could find some text that previously wasn't calculated right and check that its height is correct now I suppose

PR #5775, opened 2017-07-14
fix(table): eliminate need for second change detection

### LINKED ISSUE #5593: bug(data-table): data source connect not called when using router
I have copied the exact demo from 'material-examples' into a plunker and have loaded the `TableBasicExample ` component via a router. When navigating to the router (via button in toolbar on plunkr), the table does not display correctly. When clicking that button again, it works. In my own application I have tested by including logging in the `connect()` method of the data source. It does not get called on initial navigation to the route, but it does when you navigate the second time

One resolution I have used is to perform the dataSource initialization in a setTimeout(1), which solves the issue, but I do not believe that this should be the solution.

The data source connect method should be called after navigating to page via router

The data source connect method is not called

http://plnkr.co/edit/RRVoUV9SBAxozY6VhvpA?p=preview

latest master, and beta8

### LINKED ISSUE #5914: [Table] CdkTable doesn't automatically connect to data source when using OnPush
Bug

#### Current behavior
`dataSource.connect()` is not subscribed to until the second `ngDoCheck ` hook, which must be triggered manually when using OnPush in the parent component.

#### Expected behavior
Data source is connected as soon as possible, not requiring an extra check

#### Workaround

```ts
constructor(private cd: ChangeDetectorRef) { }

ngOnInit() {
  this.dataSource = new ExampleDataSource(this.exampleDatabase);
  setTimeout(() => this.cd.markForCheck());
}
```

Reproduction: https://plnkr.co/edit/ACvDB3mUsNIIyOC16vpu?p=preview
Workaround: https://plnkr.co/edit/4p0pnAvmfYjIwWMXSqud?p=preview

andrewseguin:
Should fix issue with the router and others since now the table does not assume it will have change detection run twice. Now finishes all of its rendering in one pass

Fixes #5593
Fixes #5914

### DIFF
--- a/src/cdk/table/data-source.ts
+++ b/src/cdk/table/data-source.ts
@@ -14,7 +14,9 @@ export interface CollectionViewer {
 
 export abstract class DataSource<T> {
   /**
-   * Connects a collection viewer (such as a data-table) to this data source.
+   * Connects a collection viewer (such as a data-table) to this data source. Note that
+   * the stream provided will be accessed during change detection and should not directly change
+   * values that are bound in template views.
    * @param collectionViewer The component that exposes a view over the data provided by this
    *     data source.
    * @returns Observable that emits a new value when the data changes.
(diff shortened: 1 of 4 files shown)


matheo:
Merge please :)

jelbourn:
@kara Please add a comment describing the presubmit failures

kara:
One app was failing with "changed after checked" errors. @andrewseguin has already looked into this and has a fix for the app once the next sync goes in.

PR #5777, opened 2017-07-15
fix(chips): remove background from unstyled chip

crisbeto:
This seems like a regression from c82aca9f8decb9860501196abc7f5028c1187344. The basic chips aren't supposed to have a background color.


tinayuangao:
LGTM

PR #5787, opened 2017-07-15
docs(demos): autocomplete examples

amcdnl:
Adds basic set of demos for autocomplete component:

- Basic
- Filtering
- Display Value
- Better overview example

This is a new PR in follow-up to #5464 merge issues.


src/material-examples/autocomplete-display/autocomplete-display-example.css
```diff
@@ -0,0 +1,7 @@
+.example-form {
+  width: 500px;
+}
```
willshowell: @amcdnl Think you could add `max-width: 100%` before it merges, here and elsewhere? If not, I'll be happy to in subsequent pr.

https://github.com/angular/material2/pull/5839

amcdnl: @willshowell - I was thinking about going through and removing all these fixed widths since it causes responsive issues. I'd like to do that in a separate PR.

PR #5798, opened 2017-07-16
fix(connected-position-strategy): multiple places to specify direction

crisbeto:
* Currently, if the consumer wants to handle layout direction in a connected overlay, they have to specify it in two places: via the `OverlayState.direction` and the `ConnectedPositionStrategy.withDirection`. This can be confusing and easy to overlook. These changes remove the `ConnectedPositionStrategy.withDirection` method, in favor of passing in the overlay state direction.
* Makes a few underscored properties private since they weren't being used anywhere outside of the `ConnectedPositionStrategy` class.

BREAKING CHANGE: This will be a breaking change if somebody was using the `withDirection` method and the way to fix it is to only pass the direction to the `OverlayState`. If we want to be safer about it, I can put the method back in and mark it as deprecated.

### DIFF
--- a/src/lib/core/overlay/overlay-directives.ts
+++ b/src/lib/core/overlay/overlay-directives.ts
@@ -264,7 +264,6 @@ export class ConnectedOverlayDirective implements OnDestroy, OnChanges {
       this._createOverlay();
     }
 
-    this._position.withDirection(this.dir);
     this._overlayRef.getState().direction = this.dir;
     this._initEscapeListener();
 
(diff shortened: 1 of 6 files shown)


jelbourn:
@crisbeto do you want to leave this as-is (avoiding the breaking change) and address it with the new position strategy?

crisbeto:
I'm fine with addressing it in the new positioning, we haven't had any reports about this one, it was just something I noticed.

PR #5810, opened 2017-07-17
docs(icons): Fix unescaped HTML tags in code comments

### LINKED ISSUE #5793: docs(md-icon): api missing text

Bug

See the complete explanation of an api item

Some text is missing in specific parts of the [**md-icon** docs (API Tab)](https://material.angular.io/components/icon/api). Where it should be written something like:

_"Returns an Observable that produces the icon (as an `<svg>` DOM element) from the given URL. The response from the URL may be cached so this will not always cause an HTTP request, but the produced element will always be a new copy of the originally fetched icon. (That is, it will not contain any modifications made to elements previously returned)."_

you actually see what is shown in the image bello (that's because of the `<svg>` tag in the code comment):

image

julianobrasil:
Fixes  #5793


julianobrasil:
I have to do something out of the office right now, but I'll fix this googlebot complain when I'm back (probably redo the fork).

PR #5813, opened 2017-07-17
builds and docs: fix #5568 (expansion) and #5793 (tslint + icons)

### LINKED ISSUE #5568: Add docs for expansion-panel and accordion
We currently are missing documentation for expansion-panel / accordion. This needs:
- [ ] An overview file (`expansion.md`)
- [ ] At least two examples (typically one simple and one more in-depth)

The style of the overview should align with existing docs.

### LINKED ISSUE #5793: docs(md-icon): api missing text

julianobrasil:
fix #5568: 
- docs(expansions): create docs for expansions

fix #5793:
- build: add tslint rule to enforce HTML tags escaping
  - The rule was built suposing that using backticks is enough to escape HTML tags in code multiline comments.
- docs(icons): Fix unescaped HTML tags in code comments

also did a cleanup in my fork (delete and forked it again): there was a lot of dirty commits in my history 😄

### DIFF
--- a/src/lib/expansion/expansion.md
+++ b/src/lib/expansion/expansion.md
@@ -1,0 +1,79 @@
+Angular Material expansion provides a way to show and hide lightweight content, by collapsing and expanding a view with a nice animation. The `expanded` input allows to choose among the collapsed or expanded state. 
+
+Each expansion has a header and an action sections. The header section is always visible at the top of the component and contains a title and a description subsections. The action section is fixed at the bottom, being visible when the expansion is in expanded state
+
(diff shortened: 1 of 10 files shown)


julianobrasil:
Done.

PR #5821, opened 2017-07-17
fix(sidenav): first animation not working in Angular 4.2+

### LINKED ISSUE #5673: regression - sidenav first animation skipped
the first sidnav animation is skipped when used inside a route

should animate

http://plnkr.co/edit/b8e2dBj4K4PrYHKBBxa5

angular 4.2.6
material 2.0.0-beta.8
chrome 59
TypeScript 2.3.3

not reproduced with angular 4.1
after the first sidenav close/open, animations work properly

crisbeto:
Fixes the first opening/closing animation not working in newer Angular versions. The issue is due to the fact that Angular no longer evaluates the `_enableTransitions` binding after we've set it in the microtask empty callback. These changes switch to setting the class directly, which also has the advantage of being a little more compact.

Fixes #5673.

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -334,7 +334,6 @@ export class MdSidenav implements AfterContentInit, OnDestroy {
   ],
   host: {
     'class': 'mat-sidenav-container',
-    '[class.mat-sidenav-transition]': '_enableTransitions',
   },
   encapsulation: ViewEncapsulation.None,
 })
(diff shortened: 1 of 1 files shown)


src/lib/sidenav/sidenav.ts
```diff
@@ -384,8 +380,12 @@ export class MdSidenavContainer implements AfterContentInit {
     });
     this._validateDrawers();
 
-    // Give the view a chance to render the initial state, then enable transitions.
-    first.call(this._ngZone.onMicrotaskEmpty).subscribe(() => this._enableTransitions = true);
+    // Give the view a chance to render the initial state, then enable transitions. Note that we
+    // don't use data binding, because we're not guaranteed that newer version of Angular will
+    // re-evaluate them after we set the flag here.
+    first.call(this._ngZone.onMicrotaskEmpty).subscribe(() => {
```
amitport: Isn't this what `AfterViewInit` is for? just wondering. Wouldn't want to have to do this in my own components

crisbeto: The main point of this is to defer the animation until the next change detection cycle, because we don't want sidenavs to animate in on load. Setting the flag in `AfterViewInit` is still a little too early.

PR #5825, opened 2017-07-17
build: tslint rule to enforce html tags escaping in comments

julianobrasil:
The rule was built suposing that using backticks is enough to escape HTML tags in multiline code  comments. It should emit warnings when there is:

1 - Any HTML open tag (`<`) without a backtick (`)

```
/** 
 * The using of <svg>` is very good because...
 */
```

2 - Any HTML close tag (`>`) without a backtick (`):

```
/** 
 * The using of `<svg> is very good because...
 */
```

3 - Any HTML open or close tag without it's matching:

```
/** 
 * The using of `<svg is very good because...
 */
```

4 - Any HTML open following an open tag and before a closing tag:

```
/** 
 * The using of `<svg attr1="value1" < >` is very good because...
 */
```

### DIFF
--- a/tools/tslint-rules/noUnscapedHtmlTagRule.js
+++ b/tools/tslint-rules/noUnscapedHtmlTagRule.js
@@ -1,0 +1,64 @@
+const ts = require('typescript');
+const utils = require('tsutils');
+const Lint = require('tslint');
+
+const ERROR_MESSAGE =
+    'A HTML tag may only appear if it is escaped. ' +
+    'This is meant to prevent failures in docs generation caused by a misinterpreted tag.';
+
+/**
+ * Rule that walks through all comments inside of the library and adds failures when it
+ * detects unescaped HTML tags inside of multi-line comments.
+ */
+class Rule extends Lint.Rules.AbstractRule {
+
+  apply(sourceFile) {
+    return this.applyWithWalker(new NoUnescapedHtmlTagWalker(sourceFile, this.getOptions()));
+  }
+}
+
+class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
+
+  visitSourceFile(sourceFile) {
+    utils.forEachComment(sourceFile, (fullText, commentRange) => {
+
+      let isEscapedHtmlTag = true;
+      while (true) {
+        const iOpenTag = fullText.indexOf('<');
+        const iCloseTag = fullText.indexOf('>');
+        if ((iOpenTag === -1) && (iCloseTag === -1)) {
+          break;
+        }
+        if ((iCloseTag < iOpenTag) || (iCloseTag === -1)) {
+          isEscapedHtmlTag = false;
+          break;
+        }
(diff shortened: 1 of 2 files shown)


julianobrasil:
Not sure whether look for unescaped html tags is better than look for "opening" backticks without the corresponding "closing" backticks.

julianobrasil:
Definitely this is not the way of escaping html tags in comment (based on the unsuccessful #7756 that didn't fix the issue)

tslint.json
```diff
@@ -28,6 +28,7 @@
     "no-unused-expression": true,
     "no-var-keyword": true,
     "no-exposed-todo": true,
+	  "no-unescaped-html-tag": true,
```
willshowell: Unindent

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,64 @@
...
+      }
+
+      if (commentRange.kind === ts.SyntaxKind.MultiLineCommentTrivia && !isEscapedHtmlTag) {
+        this.addFailureAt(commentRange.pos, commentRange.end - commentRange.pos, ERROR_MESSAGE);
+      }
+    });
+  }
+}
+
+exports.Rule = Rule;
```
willshowell: Newline at end of file

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,64 @@
+const ts = require('typescript');
+const utils = require('tsutils');
+const Lint = require('tslint');
+
+const ERROR_MESSAGE =
+    'A HTML tag may only appear if it is escaped. ' +
+    'This is meant to prevent failures in docs generation caused by a misinterpreted tag.';
```
willshowell: Maybe,

> An HTML tag may only appear in a JSDoc comment if it is escaped. This prevents failures in document generation caused by a misinterpreted tag.

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
+
+  visitSourceFile(sourceFile) {
+    utils.forEachComment(sourceFile, (fullText, commentRange) => {
+
+      let isEscapedHtmlTag = true;
+      while (true) {
```
willshowell: I think this would be much easier with a regex. A simple one like this would be effective for when either character is unescaped,

```
/[^`][<>]/g
```

You could certainly extend it to look for unmatched pairs, but I'm not sure that's even necessary.

julianobrasil: When I closed this I was just thinking it could be better to just look for unmatched pairs. In fact it was my initial thought. Than I think I forget about it and start looking for the &lt; and &gt; matching pairs. I'll review and reopen it, probably tomorow.

PR #5831, opened 2017-07-17
feat(sort): add sorting indicator animation

KillaBee1985:
Uses @angular/animations module to animate sort order change.

### DIFF
--- a/src/lib/sort/sort-header.html
+++ b/src/lib/sort/sort-header.html
@@ -7,7 +7,7 @@
 
   <div *ngIf="_isSorted()"
        class="mat-sort-header-arrow"
-       [@indicatorRotate]="this._sort.direction">
+       [@indicatorRotate]="_sort.direction">
     <div class="mat-sort-header-stem"></div>
     <div class="mat-sort-header-pointer-left"></div>
     <div class="mat-sort-header-pointer-right"></div>
--- a/src/lib/sort/sort-header.html
+++ b/src/lib/sort/sort-header.html
@@ -7,7 +7,7 @@
 
   <div *ngIf="_isSorted()"
        class="mat-sort-header-arrow"
-       [@indicatorRotate]="_getSortingState()">
+       [@indicatorRotate]="this._sort.direction">
     <div class="mat-sort-header-stem"></div>
     <div class="mat-sort-header-pointer-left"></div>
     <div class="mat-sort-header-pointer-right"></div>
--- a/src/lib/sort/sort-header.html
+++ b/src/lib/sort/sort-header.html
@@ -7,8 +7,7 @@
 
   <div *ngIf="_isSorted()"
        class="mat-sort-header-arrow"
-       [class.mat-sort-header-asc]="_sort.direction == 'asc'"
-       [class.mat-sort-header-desc]="_sort.direction == 'desc'">
+       [@indicatorRotate]="_getSortingState()">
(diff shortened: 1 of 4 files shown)


jelbourn:
@andrewseguin can you check with the Material Design UX team whether the sort header should have an animation and what that animation should be?

andrewseguin:
Checked with the spec, it should rotate when the sort changes. Will check out the PR soon

src/lib/sort/sort-header.html
```diff
@@ -7,8 +7,7 @@
 
   <div *ngIf="_isSorted()"
        class="mat-sort-header-arrow"
-       [class.mat-sort-header-asc]="_sort.direction == 'asc'"
-       [class.mat-sort-header-desc]="_sort.direction == 'desc'">
+       [@indicatorRotate]="_getSortingState()">
```
andrewseguin: Can just use `_sort.direction` without needing a function

src/lib/sort/sort-header.ts
```diff
@@ -42,6 +50,14 @@ import {merge} from 'rxjs/observable/merge';
   },
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
+  animations: [
+    trigger('indicatorRotate', [
+      state('asc', style({transform: 'rotate(45deg)'})),
+      state('desc', style({transform: 'rotate(225deg)'})),
+      /** Use standart animation curve from material guidelines */
```
andrewseguin: Can remove comment

src/lib/sort/sort-header.ts
```diff
@@ -42,6 +50,14 @@ import {merge} from 'rxjs/observable/merge';
...
+      transition('asc <=> desc', animate('225ms cubic-bezier(0.4,0.0,0.2,1)'))
```
andrewseguin: Needs spaces between the numbers: `cubic-bezier(0.4, 0.0, 0.2, 1);`

andrewseguin:
Looks good, just a few comments. Thanks for the PR - animations look good

mmalerba:
@KillaBee1985 can you squash this down into a single commit? Our presubmit script doesn't like the merge commits

src/lib/sort/sort-header.html
```diff
@@ -7,8 +7,7 @@
 
   <div *ngIf="_isSorted()"
        class="mat-sort-header-arrow"
-       [class.mat-sort-header-asc]="_sort.direction == 'asc'"
-       [class.mat-sort-header-desc]="_sort.direction == 'desc'">
+       [@indicatorRotate]="this._sort.direction">
```
rafaelss95: Although the use of `this` in template is valid (until now), it may be broken in future `angular` versions, since it's not even documented. Btw, I could not find the question I had seen in the SO that discussed this better.

rafaelss95: Just find this: https://stackoverflow.com/q/44667219/4911842

KillaBee1985:
@mmalerba, I removed all merges and left only my commits. Is it ok?

mmalerba:
yep, that works too!

andrewseguin:
Please rebase

rafaelss95:
@KillaBee1985 can you rebase this? So we can get this merged.

KillaBee1985:
@rafaelss95 Done.

PR #5836, opened 2017-07-18
fix(menu): wrong animation in Angular 4.2+

crisbeto:
Fixes the new menu animation glitching out against Angular 4.2. It seems like transitioning from `scale(0, 0)` to `scale(1, 0.5)` doesn't work correctly at the moment. This is a workaround that will transition it from `scale(0.01, 0.01)` instead.

For reference:
zp7kn12

### DIFF
--- a/src/lib/menu/menu-animations.ts
+++ b/src/lib/menu/menu-animations.ts
@@ -17,7 +17,8 @@ import{
 
 /**
  * Below are all the animations for the md-menu component.
- * Animation duration and timing values are based on AngularJS Material.
+ * Animation duration and timing values are based on:
+ * https://material.io/guidelines/components/menus.html#menus-usage
  */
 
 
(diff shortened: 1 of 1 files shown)


src/lib/menu/menu-animations.ts
```diff
@@ -34,11 +35,11 @@ import{
 export const transformMenu: AnimationTriggerMetadata = trigger('transformMenu', [
   state('void', style({
     opacity: 0,
-    transform: 'scale(0, 0)'
+    transform: 'scale(0.01, 0.01)'
```
jelbourn: Add a comment for this workaround?

crisbeto: Done.

kara:
Note: Screenshot tests need to be fixed

PR #5840, opened 2017-07-18
refactor: use renderer in DOM-related services

crisbeto:
Since it's now possible to get a hold of a renderer inside of a service, these changes switch over as much DOM manipulation as possible to it. There are still a couple of more candidates (namely the `IconRegistry`), however the `Renderer2` doesn't have all of the necessary methods.

### DIFF
--- a/src/cdk/a11y/focus-trap.ts
+++ b/src/cdk/a11y/focus-trap.ts
@@ -14,6 +14,8 @@ import {
   OnDestroy,
   AfterContentInit,
   Injectable,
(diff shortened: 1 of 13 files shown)


jelbourn:
@crisbeto I talked w/ Misko this morning about this. Angular is now trying to move away from downstream developers using `Renderer(2)`. Instead the plan is to make different platforms just work with the DOM APIs.

crisbeto:
Cool, closing in that case. We can always re-open if things change.

PR #5856, opened 2017-07-18
docs(table): add cdk module reference

### LINKED ISSUE #5819: Add CdkTableModule reference in table docs
Many people are not seeing that they need both the `MdTableModule` and `CdkTableModule` when using the `<md-table>`. This info should be added to the docs

andrewseguin:
Would like to also get your opinion on creating `mdColumnDef` and `mdHeaderCellDef` to avoid having users import the `CdkTableModule`. Been hearing a lot of users say that they struggled to find out they needed to import both `MdTableModule` and `CdkTableModule` to make the table work.

Fixes #5819

### DIFF
--- a/src/lib/table/table.md
+++ b/src/lib/table/table.md
@@ -1,14 +1,14 @@
 The `md-table` provides a Material Design styled data-table that can be used to display rows of
 data.
-
-This table builds on the foundation of the CDK data-table and uses a similar interface for its
-data source input and template, except that its element selectors will be prefixed with `md-`
-instead of `cdk-`.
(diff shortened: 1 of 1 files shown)


mmalerba:
Can `MdTableModule` re-export `CdkTableModule`, or is that too weird?

jelbourn:
@andrewseguin I think based on that feedback we can create `mdColumnDef` etc. aliases.

andrewseguin:
SGTM, will close this PR in favor of one that creates the remaining md- directives

PR #5858, opened 2017-07-18
feat(sticky-header): Initial version of sticky header new PRNew Sticky header

sllethe:
I closed the old PR (feat(sticky-header): Initial version of sticky header #5175 : #5175) by mistake and it can not be reopened. So this is the new PR...


sllethe:
Please take another look :+1:

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,399 @@
...
+      this.setStrategyAccordingToCompatibility();
+    }
+  }
+
+  ngAfterViewInit(): void {
+    if (!this.isStickyPositionSupported) {
+      this.stickyParent = this.parentRegion != null ?
+        this.parentRegion._elementRef.nativeElement : this.element.parentElement;
+      let values = window.getComputedStyle(this.element, '');
+      this.originalCss = this.generateCssStyle(
```
jelbourn: The `generateCssStyle` isn't necessary; it doesn't shorten the code or capture any behavior. It's simpler to just use an object literal here.

The `getPropertyValue` method shouldn't be necessary for the result of `getComputedStyle`; you should be able to just access the property directly, e.g., `computedStyle.display`

If you want to capture the original styles and re-apply them when you're done, you don't want to save the results of `getComputedStyle` because these styles are _computed_. To restore things as they were, you would need to save the values from the `style` property of the element. If you need any of the computed styles for other reasons, then you shoudl store those separately from the `originalCss` that's used for restoration.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,399 @@
...
+    let stuckRight: any = this.upperScrollableContainer.getBoundingClientRect().right;
+
+    let stickyCss:any = this.generateCssStyle(
+      this.zIndex + '',
+      'fixed',
+      this.upperScrollableContainer.offsetTop + 'px',
+      stuckRight + 'px',
+      this.upperScrollableContainer.offsetLeft + 'px',
+      'auto',
+      this.originalCss.width);
```
jelbourn: Instead of using the `width` from `getComputedStyle`, you probably want either `clientWidth` or `offsetWidth` here.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,399 @@
...
+     * "fixed" to the viewport can end up outside the viewable content area, offscreen.
+     *
+     * So the 'position: fixed' does not work on iPhone and iPad. To make it work,
+     * 'translate3d(0,0,0)' needs to be used to force Safari re-rendering the sticky element.
+     **/
+    this.element.style.transform = 'translate3d(0px,0px,0px)';
+
+    let stuckRight: any = this.upperScrollableContainer.getBoundingClientRect().right;
+
+    let stickyCss:any = this.generateCssStyle(
```
jelbourn: As with the other place this is used, `generateCssStyle` would be more readibly replaced with:
```ts
const stickyCss = {
  zIndex: this.zIndex,
  position: 'fixed',
  top: `${this.upperScrollableContainer.offsetTop}px`
  left: `${this.upperScrollableContainer.offsetLeft}px`
  bottom: `auto`,
  right: `${stuckRight}px`,
  width: this.originalCss.width,
} as CSSStyleDeclaration;
```
This way you can see exactly which values are used for which properties without having to look at the function signature.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,399 @@
...
+     * of the viewport as they double tap or pinch to zoom in or out, or by touching
+     * and dragging to pan the page. As a user changes the zoom level and position of the
+     * viewport they are doing so within a viewable content area of fixed size
+     * (that is, the window). This means that webpage elements that have their position
+     * "fixed" to the viewport can end up outside the viewable content area, offscreen.
+     *
+     * So the 'position: fixed' does not work on iPhone and iPad. To make it work,
+     * 'translate3d(0,0,0)' needs to be used to force Safari re-rendering the sticky element.
+     **/
+    this.element.style.transform = 'translate3d(0px,0px,0px)';
```
jelbourn: Use inline (`// `) style comments for this; the block-style comments (`/** */`) are only used for adding descriptions to classes, properties, methods, etc.

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,399 @@
...
+  }
+
+  /** Reset element to its original CSS. */
+  resetElement(): void {
+    this.element.classList.remove(STICK_START_CLASS);
+    extendObject(this.element.style, this.originalCss);
+  }
+
+  /** Stuck element, make the element stick to the top of the scrollable container. */
+  stickElement(): void {
```
jelbourn: Should this method be `private`?

src/lib/sticky-header/sticky-header-dir.ts
```diff
@@ -0,0 +1,399 @@
...
+   * also work well on Safari with a "-webkit-" prefix. It only does not work on IE.
+   *
+   * For Mobile: Sticky positioning works well on Android Webview, Chrome for Android, Edge,
+   * Firefox Mobile, Opera Mobile. And can also work well on Safari Mobile with a "-webkit-" prefix.
+   * It won't always work on IE phone.
+   *
+   * The implementation references the compatibility checking in Modernizer
+   * (https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/positionsticky.js).
+   */
+  getSupportList(): string[] {
```
jelbourn: 1. Add the following to `src/cdk/platform/features.ts`:
```ts
let computedPositionStickySupported: boolean | null = null;

/**
 * Whether the browser support css `position: sticky`.
 * Based on the check from modernizr:
 * https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/positionsticky.js
 */
export function isPositionStickySupported() {
  if (computedPositionStickySupported != null) {
    return computedPositionStickySupported;
  }

  const elementStyle = document.createElement('div').style;
  elementStyle.cssText = ['', '-webkit-'].map(p => `position: ${p}sticky`).join(';');
  computedPositionStickySupported = elementStyle.cssText.indexOf('sticky') !== -1;
  return computedPositionStickySupported;
}
```
(the `ms`, `moz`, and `o` prefixes aren't necessary for us)
2. Consume that function in this component and just always use both the `webkit` and unprefixed styles.

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Technically, it should be `clientWidth`. But the client width contains the width of the scroll bar. So if we use ' clientWidth', we'll find that the sticky-header overlays on the scroll bar, which is vary strange. And using the width from getComputedStyle can solve this problem

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Change it to private

src/lib/sticky-header/sticky-header-dir.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header-dir.ts
sllethe: Added the `'isPositionStickySupported()` ' function  to `src/cdk/platform/features.ts. `
Consume that function in this component and just always use both the webkit and unprefixed styles in `setStrategyAccordingToCompatibility` function

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+      this.element = element.nativeElement;
+      this.upperScrollableContainer = scrollable.getElementRef().nativeElement;
+      this.setStrategyAccordingToCompatibility();
+    }
+  }
+
+  ngAfterViewInit(): void {
+    if (!this._isStickyPositionSupported) {
+      this.stickyParent = this.parentRegion != null ?
+        this.parentRegion._elementRef.nativeElement : this.element.parentElement;
```
andrewseguin: indent a couple more lines here

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+    }
+  }
+
+  /**
+   * Check if current browser supports sticky positioning. If yes, apply
+   * sticky positioning. If not, use the original implementation.
+   */
+  setStrategyAccordingToCompatibility(): void {
+    this._isStickyPositionSupported = isPositionStickySupported();
+    if (this._isStickyPositionSupported == true) {
```
andrewseguin: can just do `if (this._isStickyPositionSupported)` - shouldn't ever be the case to need `== true` or `== false` in an if conditional. Just use the value

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+      this.unstuckElement();
+      this._stickElement();
+    }
+  }
+
+  /**
+   * define the restrictions of the sticky header(including stickyWidth,
+   * when to start, when to finish)
+   */
+  defineRestrictions(): void {
```
andrewseguin: Can you evaluate each method to make sure their accessor privacy is right? E.g. see which functions need to be public, private, static, etc

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+    if (!this.stickyParent) {
+      return;
+    }
+    let boundingClientRect: any = this.stickyParent.getBoundingClientRect();
+    let elemHeight: number = this.element.offsetHeight;
```
andrewseguin: Not sure we need this variable

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  /**
+   * define the restrictions of the sticky header(including stickyWidth,
+   * when to start, when to finish)
+   */
+  defineRestrictions(): void {
+    if (!this.stickyParent) {
+      return;
+    }
+    let boundingClientRect: any = this.stickyParent.getBoundingClientRect();
```
andrewseguin: These can be `const`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+   * It should be unstuck. And its position will be set as 'relative', its bottom
+   * will be set as '0'. So it will be stick at the bottom of its cdkStickyRegion and
+   * will be scrolled up with its cdkStickyRegion element. In this way, the sticky header
+   * can be changed smoothly when two sticky header meet and the later one need to replace
+   * the former one.
+   */
+  unstuckElement(): void {
+    this.isStuck = false;
+
+    if (this.stickyParent == null) {
```
andrewseguin: `if (!this.stickyParent)`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+      right: '0',
+      left: 'auto',
+      bottom: '0',
+      width: this._originalStyles.width};
+    extendObject(this.element.style, unstuckCss);
+  }
+
+
+  /**
+   * 'sticker()' function contains the main logic of sticky-header. It decides when
```
andrewseguin: I suspect that if such a large jsdoc is necessary, then the function can be broken down better so it doesn't even need such an explanation. Maybe we can work together on simplifying the code to make it more obvious what it is doing

src/cdk/platform/features.ts
```diff
@@ -62,3 +62,22 @@ export function getSupportedInputTypes(): Set<string> {
...
+ * Based on the check from modernizr:
+ * https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/positionsticky.js
+ */
+export function isPositionStickySupported() {
+  if (computedPositionStickySupported != null) {
+    return computedPositionStickySupported;
+  }
+
+  const elementStyle = document.createElement('div').style;
+  elementStyle.cssText = ['', '-webkit-'].map(p => `position: ${p}sticky`).join(';');
```
jelbourn: I actually prefer single-character variables like this in map, filter, etc.

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+
+
+/** Class applied when the header is "stuck" */
+const STICK_START_CLASS = 'cdk-sticky-header-start';
+
+/** Class applied when the header is not "stuck" */
+const STICK_END_CLASS = 'cdk-sticky-header-end';
+
+/**
+ * Set a debounce time which is used in debounce() function when adding event listeners.
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+ * cdkScrollable), this header will "stick" to the top of the scrolling viewport while its sticky
+ * region (see cdkStickyRegion) is in view.
+ */
+@Directive({
+  selector: '[cdkStickyHeader]',
+})
+export class CdkStickyHeader implements OnDestroy, AfterViewInit {
+
+  /**
+   * Set the sticky-header's z-index as 10 in default. Make it as an input
```
sllethe: :+1:  Changed the doc to` '/** z-index to be applied to the sticky header (default is 10). */'`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,321 @@
...
+   * variable to make user be able to customize the zIndex when
+   * the sticky-header's zIndex is not the largest in current page.
+   * Because if the sticky-header's zIndex is not the largest in current page,
+   * it may be sheltered by other element when being stuck.
+   */
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+
+  isStuck: boolean = false;
```
sllethe: :+1:  Added comments for `'isStuck`'

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:  indented 2 more lines for `this.stickyParent = this.parentRegion != null ?  this.parentRegion._elementRef.nativeElement : this.element.parentElement;`

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:  deleted `== true`

src/lib/sticky-header/sticky-header.ts
sllethe: Added `private` before the functions that I think should be private, which are `_setStrategyAccordingToCompatibility()`, `_sitckElement()`, and `_unstuckElement()`. Others are public functions.

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:  Changed into '`const`'

src/lib/sticky-header/sticky-header.ts
sllethe: Deleted variable `elemHeight`

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
sllethe: Simplified the doc to `'sticker()' function contains the main logic of sticky-header. It decides when
   a header should be stick and when should it be unstuck by comparing the offsetTop
   of scrollable container with the top and bottom of the sticky region.`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,301 @@
...
+})
+export class CdkStickyHeader implements OnDestroy, AfterViewInit {
+
+  /** z-index to be applied to the sticky header (default is 10). */
+  @Input('cdkStickyHeaderZIndex') zIndex: number = 10;
+
+  /** boolean value to mark whether the current header is stuck*/
+  isStuck: boolean = false;
+  /** Whether the browser support CSS sticky positioning. */
+  private _isStickyPositionSupported: boolean = true;
```
tinayuangao: Use `isPositionStickySupported`?

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,301 @@
...
+      .call(debounceTime, DEBOUNCE_TIME).subscribe(() => this.defineRestrictionsAndStick());
+
+    this._onResizeSubscription = RxChain.from(fromEvent(this.upperScrollableContainer, 'resize'))
+      .call(debounceTime, DEBOUNCE_TIME).subscribe(() => this.onResize());
+  }
+
+  onResize(): void {
+    this.defineRestrictionsAndStick();
+    /**
+     * If there's already a header being stick when the page is
```
tinayuangao: Use `//` for comments, `/* */` for docs

src/lib/module.ts
```diff
@@ -86,7 +87,9 @@ const MATERIAL_MODULES = [
   A11yModule,
   PlatformModule,
   MdCommonModule,
-  ObserveContentModule
+  ObserveContentModule,
+  CdkDataTableModule,
```
sllethe: :+1: Deleted `CdkDataTableModule`

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:  Changed to `'_isPositionStickySupported'`

src/lib/sticky-header/sticky-header.ts
sllethe: :+1: Changed to '//'

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /**
+   * Unstuck element: When an element reaches the bottom of its cdkStickyRegion,
+   * It should be unstuck. And its position will be set as 'relative', its bottom
+   * will be set as '0'. So it will be stick at the bottom of its cdkStickyRegion and
+   * will be scrolled up with its cdkStickyRegion element. In this way, the sticky header
+   * can be changed smoothly when two sticky header meet and the later one need to replace
+   * the former one.
+   */
```
jelbourn: How about
```ts
  /**
   * Unsticks the header so that it goes back to scrolling normally.
   *
   * This should be called when the element reaches the bottom of its cdkStickyRegion so that it
   * smoothly scrolls out of view as the next sticky-header moves in.
   */
```

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    this.element.classList.add(STICK_END_CLASS);
+    this.stickyParent.style.position = 'relative';
+    let unstuckCss = {
+      position: 'absolute',
+      top: 'auto',
+      right: '0',
+      left: 'auto',
+      bottom: '0',
+      width: this._originalStyles.width};
```
jelbourn: Move closing brace to next line

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+    extendObject(this.element.style, unstuckCss);
+  }
+
+
+  /**
+   * 'sticker()' function contains the main logic of sticky-header. It decides when
+   * a header should be stick and when should it be unstuck by comparing the offsetTop
+   * of scrollable container with the top and bottom of the sticky region.
+   */
+  sticker(): void {
```
jelbourn: How about something like `_applyStickyPositionStyles`? The name `sticker` doesn't tell you much about what the function does.

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+    let currentPosition: number = this.upperScrollableContainer.offsetTop;
+
+    // unstuck when the element is scrolled out of the sticky region
+    if (this.isStuck &&
+      (currentPosition < this._stickyRegionTop ||
+      currentPosition > this._stickyRegionBottomThreshold)
+      || currentPosition >= this._stickyRegionBottomThreshold) {
```
jelbourn: Should be consistent whether the operators go at the beginning or the end of the line

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+      }
+      this.isStuck = false;    // stick when the element is within the sticky region
+    } else if ( this.isStuck === false &&
+      currentPosition > this._stickyRegionTop &&
+      currentPosition < this._stickyRegionBottomThreshold) {
+      this._stickElement();
+    }
+  }
+
+  defineRestrictionsAndStick(): void {
```
jelbourn: Maybe `_updateStickyPositioning`?

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+  }
+
+  ngAfterViewInit(): void {
+    if (!this._isPositionStickySupported) {
+
+      this.stickyParent = this.parentRegion != null ?
+        this.parentRegion._elementRef.nativeElement : this.element.parentElement;
+
+      let values = window.getComputedStyle(this.element, '');
```
sllethe: :+1:  renamed: `values` -> `headerStyles`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+      this._originalStyles = {
+        position: values.position,
+        top: values.top,
+        right: values.right,
+        left: values.left,
+        bottom: values.bottom,
+        width: values.width,
+        zIndex: values.zIndex} as CSSStyleDeclaration;
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+
+      this.attach();
+      this.defineRestrictionsAndStick();
+    }
+  }
+
+  ngOnDestroy(): void {
+    if (this._onScrollSubscription) {
```
sllethe: :+1:  Changed it to  `'[this._onScrollSubscription, this._onScrollSubscription, this._onResizeSubscription]
        .forEach(s => s && s.unsubscribe());'`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /**
+   * Check if current browser supports sticky positioning. If yes, apply
+   * sticky positioning. If not, use the original implementation.
+   */
+  private _setStrategyAccordingToCompatibility(): void {
+    this._isPositionStickySupported = isPositionStickySupported();
+    if (this._isPositionStickySupported) {
+      this.element.style.top = '0px';
```
sllethe: :+1:  Changed to '`0`'

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+      this.element.style.cssText += 'position: -webkit-sticky; position: sticky; ';
+      // TODO add css class with both 'sticky' and '-webkit-sticky' on position
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+      // when @Directory supports adding CSS class
+    }
+  }
+
+  attach() {
```
sllethe: :+1: Changed to `private _attachEventListeners`, and Added a description  "Add listeners for events that affect sticky positioning."

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+     // unstuck() can set 'isStuck' to FALSE. Then _stickElement() can work.
+    if (this.isStuck) {
+      this._unstuckElement();
+      this._stickElement();
+    }
+  }
+
+  /**
+   * define the restrictions of the sticky header(including stickyWidth,
+   * when to start, when to finish)
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+   */
+  private _defineRestrictions(): void {
```
sllethe: :+1: Renamed to `'_measureStickyRegionBounds'`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+    const boundingClientRect: any = this.stickyParent.getBoundingClientRect();
+    this._stickyRegionTop = boundingClientRect.top;
+    let stickRegionHeight = boundingClientRect.height;
+
+    this._stickyRegionBottomThreshold = this._stickyRegionTop +
+      (stickRegionHeight - this.element.offsetHeight);
+  }
+
+  /** Reset element to its original CSS. */
+  resetElement(): void {
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+    // and dragging to pan the page. As a user changes the zoom level and position of the
+    // viewport they are doing so within a viewable content area of fixed size
+    // (that is, the window). This means that webpage elements that have their position
+    // "fixed" to the viewport can end up outside the viewable content area, offscreen.
+    //
+    // So the 'position: fixed' does not work on iPhone and iPad. To make it work,
+    // 'translate3d(0,0,0)' needs to be used to force Safari re-rendering the sticky element.
+    this.element.style.transform = 'translate3d(0px,0px,0px)';
+
+    let stuckRight: any = this.upperScrollableContainer.getBoundingClientRect().right;
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    let stickyCss = {
+      position: 'fixed',
+      top: this.upperScrollableContainer.offsetTop + 'px',
+      right: stuckRight + 'px',
+      left: this.upperScrollableContainer.offsetLeft + 'px',
+      bottom: 'auto',
+      width: this._originalStyles.width,
+      zIndex: this.zIndex + '',};
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * Unstuck element: When an element reaches the bottom of its cdkStickyRegion,
+   * It should be unstuck. And its position will be set as 'relative', its bottom
+   * will be set as '0'. So it will be stick at the bottom of its cdkStickyRegion and
+   * will be scrolled up with its cdkStickyRegion element. In this way, the sticky header
+   * can be changed smoothly when two sticky header meet and the later one need to replace
+   * the former one.
+   */
+  private _unstuckElement(): void {
```
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
sllethe: changed name to '`applyStickyPositionStyles`'

src/lib/sticky-header/sticky-header.ts
sllethe: :+1:

src/lib/sticky-header/sticky-header.ts
sllethe: :+1: Changed to `'_updateStickyPositioning'`

src/lib/sticky-header/sticky-header.ts
```diff
@@ -0,0 +1,299 @@
...
+      this._onTouchSubscription.unsubscribe();
+    }
+  }
+
+  /**
+   * Check if current browser supports sticky positioning. If yes, apply
+   * sticky positioning. If not, use the original implementation.
+   */
+  private _setStrategyAccordingToCompatibility(): void {
+    this._isPositionStickySupported = isPositionStickySupported();
```
tinayuangao: Use computedPositionStickySupported?

tinayuangao:
LGTM

andrewseguin:
LGTM :)

PR #5859, opened 2017-07-19
chore(table): add md directives to match cdk table

andrewseguin:
Add directives matching the table's remaining cdk directives except with an `md` prefix. This was a big source of confusion for people trying to use the table, in particular the requirement of both the `MdTableModule` and `CdkTableModule` to use an `<md-table>`


src/lib/table/cell.ts
```diff
@@ -6,13 +6,46 @@
...
+export const _MdHeaderCellDef = CdkHeaderCellDef;
+export const _MdColumnDef = CdkColumnDef;
+export const _MdHeaderCell = CdkHeaderCell;
 export const _MdCell = CdkCell;
 
+/**
+ * Cell definition for the md-table.
+ * Captures the template of a column's data row cell as well as cell-specific properties.
+ */
+@Directive({selector: '[mdCellDef]'})
```
jelbourn: Need `mat` as well and need to add them all to the compatibility list

src/lib/sort/sort-header.ts
```diff
@@ -67,7 +68,8 @@ export class MdSortHeader implements MdSortable {
   constructor(public _intl: MdSortHeaderIntl,
               private _changeDetectorRef: ChangeDetectorRef,
               @Optional() public _sort: MdSort,
-              @Optional() public _cdkColumnDef: CdkColumnDef) {
+              @Optional() public _cdkColumnDef: CdkColumnDef,
+              @Optional() public _mdColumnDef: MdColumnDef) {
```
andrewseguin: Awesome, I wasn't aware of that. Cleans things up a lot. Thanks

andrewseguin:
Added compatibility stuff. Using providers which cleaned up a lot

PR #5862, opened 2017-07-19
feat(divider): move divider out of mat-list

### LINKED ISSUE #2017: please add md-divider to md-menu
feature request

I would like to use a md-menu-divider, or make the existing md-divider work within md-menu. This is not in the design doc, but it will be very handy.

There isn't one available.

I would like one just like in md-list, the md-divider

At the moment I use an `<hr/>` element to have a divider in the menu, but the original styling of the `<hr/>` element isn't very nice, so I have to add styling to the element. A md-menu-divider with the same styling as the md-divider will be a prefered.

Angular Material 2.0.0-alpha-10

### LINKED ISSUE #5524: feat(divider): move divider out of mat-list

Feature request

Pursuant to the spec, dividers are used in areas other than plain lists, most notably in [cards](https://material.io/guidelines/components/cards.html#cards-content-blocks). 

Using `mat-divider` requires `MatList`, even when used in other components.

1. Try to use `mat-divider` without importing `MatListModule`

There should be an ability, like in AngularJS Material, to use `mat-divider` without relying on `MatList`. This would also be a great opportunity to implement the [inset divider](https://github.com/angular/material2/issues/4570), would close #2017, and open up an opportunity to implement the `vertical` (or at least responsive) option that was in Material 1.0.

Obviously not a priority, but a nice to have feature.

All current as of 7/5/2017 (@angular/material@beta.7)

CaerusKaru:
Moves `mat-divider` out of `MatListModule`, replaced with temporary import to prevent breaking changes. Also includes support for inset and vertical dividers.

Fixes #5524.
Fixes #2017.


CaerusKaru:
Requesting reviews from @jelbourn and @crisbeto

Gustav0ar:
@CaerusKaru in divider.scss you need to add:
&.mat-divider-vertical {
    **border-right-color: rgba(0, 0, 0, 0.12);**
}

Otherwise the color will be black.

CaerusKaru:
@Gustav0ar the color is set in _divider-theme.scss

Gustav0ar:
Ah, ok, my bad. I missed reviewing that file

src/demo-app/list/list-demo.html
```diff
@@ -22,6 +22,7 @@ <h3 mat-line>{{contact.name}}</h3>
       </mat-list-item>
     </mat-list>
 
+<<<<<<< HEAD
```
rafaelss95: Did you forgot to remove this? Also in line 43.

CaerusKaru: Resolved, thanks for the catch. I rebased and for some reason Git missed this.

CaerusKaru:
Travis now passes all relevant divider-related tests. Failures are unrelated and seem to be occurring on other PRs. Apologies to those subscribed for the commit spam (if any).

CaerusKaru:
Travis is now failing because of a SauceLabs timeout/connection issue.

crisbeto:
@CaerusKaru if you rebase from master the errors should be resolved.

CaerusKaru:
Thanks @crisbeto, rebased and ready for review.

src/demo-app/card/card-demo.html
```diff
@@ -28,6 +28,23 @@
...
+    <mat-card-subtitle>Subtitle</mat-card-subtitle>
+    <mat-card-title>Card with title, footer, and inset-divider</mat-card-title>
+    <mat-card-content>
+      <p>This is supporting text.</p>
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
+    </mat-card-content>
+    <mat-divider [inset]="true"></mat-divider>
+    <mat-card-actions>
+      <button md-button>LIKE</button>
+      <button md-button>SHARE</button>
```
TheRedD0t: These two should be `mat-button`.

src/demo-app/list/list-demo.html
```diff
@@ -31,6 +31,7 @@ <h4 mat-line>{{message.from}}</h4>
           <span>{{message.subject}} -- </span>
           <span class="demo-secondary-text">{{message.message}}</span>
         </p>
+        <mat-divider [inset]="true" *ngIf="!last"></mat-divider>
```
TheRedD0t: You forgot to bind `last` to a local variable: `*ngFor="let message of messages; last as last"`

CaerusKaru: Added, this got cut from a previous commit when rebased, good catch

CaerusKaru:
@TheRedD0t Added menu demo. Initially didn't because I wasn't sure the styles would play nice, but it turns out it works great.

CaerusKaru:
Build errors depend on #8030 to be resolved

src/lib/divider/divider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+})
+class MatDividerInset { }
+
+@Component({
+    template: `<mat-divider [vertical]="true" [inset]="true"></mat-divider>`
+})
+class MatDividerVerticalInset { }
+
+@Component({
+    template: `<mat-divider class="test-class"></mat-divider>`
```
rafaelss95: The indentation in this file isn't consistent (afaik it should be 2 spaces).

src/lib/divider/divider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+          MatDividerVerticalInset,
+          MatDividerWithCssClass,
+      ],
+    });
+
+    TestBed.compileComponents();
+  }));
+
+  it('should not apply any additional class to a list without lines', () => {
+    let fixture = TestBed.createComponent(MatDividerBase);
```
rafaelss95: You can use `const` here and where possible.

rafaelss95:
The indentation in a lot of changed files isn't consistent (afaik it should be 2 spaces).

src/lib/divider/divider.md
```diff
@@ -0,0 +1,64 @@
+`<md-divider>` is a container component that wraps and formats a series of line items. As the base
```
rafaelss95: It should be `<mat-divider>` here and elsewhere.

src/lib/divider/divider.spec.ts
CaerusKaru: This was based on every other spec file in the Material lib, which uses `let` instead of `const`. That doesn't make it better, so I switched it to `const`.

CaerusKaru:
@rafaelss95 Indentation issues fixed

src/lib/divider/divider-module.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {NgModule} from '@angular/core';
+import {MatCommonModule} from '@angular/material/core';
+import {
+  MatDivider,
+  MatDividerCssMatStyler,
+} from './divider';
```
rafaelss95: nit: you can join this import in a single line and also move the `@angular/common` to the top.

CaerusKaru: I moved the `@angular/common` imports, but the multi-line import is based on precedent in other modules (and offers no benefit to change, since it all gets whitespaced away later)

CaerusKaru:
@jelbourn @crisbeto Is there any way we can get a review on this before Material goes to final? It would be a breaking change for a lot of people who rely on it in list, or would want to start using it in menu and card.

jelbourn:
We're not making any more breaking API changes now before the stable release (unless it's to fix some previously unnoticed severe bug).

I don't think that the divider has enough complexity / behavior to warrant being a top-level component; the only reason it's even a directive for lists is because it has `role="separator"` in that context (which it wouldn't when it's used in a purely visual context like in a card). 

Apologies for taking so long to respond- other issues kept taking priority

CaerusKaru:
@jelbourn I can modify the PR to reexport MatDivider in MatList to prevent a breaking change, but then we might want to mark it as deprecated behavior. 

Here's my motivation for pushing for this PR:

* It's duplicated behavior across `menu`, `list`, `card`, so creating separate directives in each component seems wasteful/unnecessary
* It ties in to Material theming so that in light/dark theme it's always properly styled
* The current implementation does not support vertical alignment or inset functionality, which adds to the complexity/behavior and is much requested functionality that was present in AngularJS Material
* It's arguably on par with `mat-card` in terms of complexity

If this can't be a top-level component, can we possibly make a cdk/material styling tools component?

> which it wouldn't when it's used in a purely visual context like in a card

Can you further explain this point? In the card section of [the spec](https://material.io/guidelines/components/cards.html#cards-content-blocks), under the section labeled `Use inset dividers to separate related content.`, doesn't the divider separate distinct content/functionality?

jelbourn:
I was referring to the a11y bits specifically, but apparently I misremembered the details; [it seems like `role="separator"` is okay to "provide a horizontal rule between two sections of a page"](https://www.w3.org/TR/wai-aria-1.1/#separator).

You've convinced me that it's worth having as a standalone thing (provided it's done in a non-breaking way). I don't want to add any new components during the RC period, but we can revisit this once that's over for a minor release (something like 5.1 or 5.2). 

@DevVersion @josephperrott can you give this a full review?

src/lib/divider/divider.ts
```diff
@@ -0,0 +1,55 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  preserveWhitespaces: false,
+})
+export class MatDivider {
+  /** Whether the divider is vertically aligned. */
+  @Input() get vertical(): boolean { return this._vertical; }
+  set vertical(value: boolean) {
+    this._vertical = coerceBooleanProperty(value);
+  }
+  protected _vertical: boolean = false;
```
josephperrott: `_vertical` and `_inset` should be `private` rather than `protected`

src/lib/divider/divider.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Directive whose purpose is to add the mat- CSS styling to this selector.
+ * @docs-private
+ */
+@Directive({
+  selector: 'mat-divider',
+  host: {'class': 'mat-divider'}
+})
+export class MatDividerCssMatStyler {}
```
josephperrott: I am not sure I see why this directive needs to exist.

Since it uses the same selector as `MatDivider` it will trigger on the same elements.  You can just add `'class': 'mat-divider'` to the `host` attribute in the `@Component` decorator for `MatDivider`

src/lib/divider/divider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        MatDividerInset,
+        MatDividerVerticalInset,
+        MatDividerWithCssClass,
+      ],
+    });
+
+    TestBed.compileComponents();
+  }));
+
+  it('should not apply any additional class to a list without lines', () => {
```
josephperrott: This test is not necessary as you are testing the functionality of the angular framework rather than the created component.

Same for the test below `it('should not clear custom classes provided by user', () => {`

src/lib/divider/divider.ts
```diff
@@ -0,0 +1,55 @@
...
+@Component({
+  moduleId: module.id,
+  selector: 'mat-divider',
+  host: {
+    'role': 'separator',
+    '[attr.aria-orientation]': 'vertical ? "vertical" : "horizontal"',
+    '[class.mat-divider-vertical]': 'vertical',
+    '[class.mat-divider-inset]': 'inset',
+  },
+  template: '<ng-content></ng-content>',
```
josephperrott: I am not quite sure I see the value in being able to project content into the `mat-divider` element that is created.  What would be the use case for this?

src/lib/divider/divider.ts
CaerusKaru: This is a layover from the list component, pulled verbatim. Integrated with other component as suggested.

src/lib/divider/divider.ts
CaerusKaru: Changed

src/lib/divider/divider.ts
CaerusKaru: Changed

src/lib/divider/divider.spec.ts
CaerusKaru: I got that test from the original list tests, shown [here](https://github.com/angular/material2/blob/20a23f16e8ac1ced29172709cc837fa500b6377b/src/lib/list/list.spec.ts#L89). I can remove both, but should I replace these tests with others, or are the remaining tests sufficient?

src/lib/divider/divider.scss
```diff
@@ -0,0 +1,66 @@
...
+  display: block;
+  margin: 0;
+  border-top-width: 1px;
+  border-top-style: solid;
+
+  &.mat-divider-vertical {
+    border-top: 0;
+    border-right-width: 1px;
+    border-right-style: solid;
+    position: static !important;
```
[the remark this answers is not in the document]
CaerusKaru: Good catch, will add.

src/lib/divider/divider.md
```diff
@@ -0,0 +1,64 @@
...
+```html
+<mat-divider></mat-divider>
+```
+
+Here are the available global options:
+
+| Name            | Type    | Values      | Description                               |
+| --------------- | ------- | ----------- | ----------------------------------------- |
+| inset           | boolean | true, false | Whether the divider is an inset divider   |
+| vertical        | boolean | true, false | Whether the divider is a vertical divider |
```
willshowell: IMO the values column is covered by them both having boolean type. Further, I'm not sure this table is even necessary, since it will be available on the api page.

src/lib/divider/divider.md
```diff
@@ -0,0 +1,64 @@
+`<mat-divider>` is a container component that wraps and formats a series of line items. As the base
+list component, it provides Material Design styling, but no behavior of its own.
+
+<!-- example(divider-overview) -->
```
willshowell: There doesn't seem to be a divider example?

src/lib/divider/divider.md
CaerusKaru: Originally pulled from list, but now I see it's been removed. Will fix.

src/lib/divider/divider.md
CaerusKaru: Do you mean in the `material-examples` directory?

CaerusKaru: I see now that the prose is wrong here, will correct.

willshowell: Yep. It would probably be good to add one in with this PR anyway.

CaerusKaru: Added

src/lib/divider/divider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  it('should add aria roles properly', () => {
+    const fixture = TestBed.createComponent(MatDividerBase);
+    fixture.detectChanges();
+
+    const divider = fixture.debugElement.query(By.css('mat-divider'));
+    expect(divider.nativeElement.getAttribute('role')).toBe('separator');
+  });
+});
+
+@Component({
```
josephperrott: Rather than creating a number of Components that have to be recompiled on every run, can you create a single Component that is altered through attributes for tests?

We have found a bottleneck in our tests comes from having to recompile many components on each test run.

```ts
@Component({
  template: `<mat-divider [inset]="inset" [vertical]="vertical"></mat-divider>`
})
class MatDividerTestComponent {
  vertical: boolean;
  inset: boolean;
}
```

Then in tests you would do:
```ts
it('should apply a vertical class', () => {
  fixture.componentInstance.vertical = true;
  fixture.detechChanges();

  expect(divider.nativeElement.className).toContain('mat-divider-vertical');
});

File: src/lib/divider/divider.spec.ts [hunk 16]
```diff
@@ -0,0 +1,99 @@
...
+        MatDividerVerticalInset,
+        MatDividerWithCssClass,
+      ],
+    });
+
+    TestBed.compileComponents();
+  }));
+
+  it('should not apply any additional class to a list without lines', () => {
+    const fixture = TestBed.createComponent(MatDividerBase);
```
josephperrott: Instead of calling `TestBed.createComponent` in every test, set up a `beforeEach` block to do this setup.

let fixture: ComponentFixture<MatDividerTestComponent>;

beforeEach(() => {
  fixture = TestBed.createComponent(MatDividerTestComponent);
});

src/lib/divider/divider.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {
```
josephperrott: Please put one import per line if it doesn't fit on a single line:

```ts
import {
  ChangeDetectionStrategy,
  Component,
  Input,
  ViewEncapsulation
} from '@angular/core';
```

src/lib/divider/divider.scss
```diff
@@ -0,0 +1,66 @@
...
+    [dir='rtl'] & {
+      margin-left: auto;
+      margin-right: 80px;
+    }
+  }
+}
+
+.mat-list-item {
+  .mat-divider {
+    position: absolute;
```
josephperrott: There seems to be inconsistent spacing between attributes, can you make the spacing more consistent?

src/lib/divider/divider.scss
```diff
@@ -0,0 +1,66 @@
...
+  display: block;
+  margin: 0;
+  border-top-width: 1px;
+  border-top-style: solid;
+
+  &.mat-divider-vertical {
+    border-top: 0;
+    border-right-width: 1px;
+    border-right-style: solid;
+    position: static;
```
josephperrott: `position: static` and `width: auto` should not be necessary as these are the default values on any custom element.

src/lib/divider/divider.scss
```diff
@@ -0,0 +1,66 @@
...
+    width: auto;
+  }
+
+  &.mat-divider-inset {
+    margin-left: 80px;
```
josephperrott: Where does this 80px size come from?  Can you put in a comment explaining it?

josephperrott: For the inset's `margin` size, can you move the value (`80px`) into a variable at the top of the file?

src/lib/divider/divider.scss
```diff
@@ -0,0 +1,66 @@
+.mat-divider {
+  display: block;
+  margin: 0;
+  border-top-width: 1px;
```
josephperrott: For the dividers line width (based on the border width), can you move the value (`1px`) into a variable at the top of the file?

src/lib/divider/divider.scss
```diff
@@ -0,0 +1,66 @@
...
+  &.mat-divider-inset {
+    margin-left: 80px;
+    [dir='rtl'] & {
+      margin-left: auto;
+      margin-right: 80px;
+    }
+  }
+}
+
+.mat-list-item {
```
josephperrott: Rules specific to other components should be in the the respective component since it is an override specifically for the component.

src/lib/divider/divider.ts
```diff
@@ -0,0 +1,45 @@
...
+  template: '',
+  styleUrls: ['divider.css'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  preserveWhitespaces: false,
+})
+export class MatDivider {
+  /** Whether the divider is vertically aligned. */
+  @Input() get vertical(): boolean { return this._vertical; }
+  set vertical(value: boolean) {
```
josephperrott: for both setter methods, you should be able to move them to a single line which might be a bit more readable.

src/lib/divider/divider.scss
CaerusKaru: It came from AngularJS Material found [here](https://github.com/angular/material/blob/72d068593b7f353d32f8308870d2b0d1c7c759c9/src/components/divider/divider.scss#L8). `baseline-grid = 8`.

CaerusKaru: Done

src/lib/divider/divider.scss
CaerusKaru: Will add

src/lib/divider/divider.scss
CaerusKaru: Will fix

src/lib/divider/divider.spec.ts
CaerusKaru: Will add

src/lib/divider/divider.spec.ts
CaerusKaru: Great suggestions, will add

src/lib/divider/divider.ts
CaerusKaru: Will move

src/lib/divider/divider.ts
CaerusKaru: Fixed

src/lib/divider/divider.scss
CaerusKaru: Done

src/lib/divider/divider.scss
CaerusKaru: Removed

src/lib/divider/divider.scss
CaerusKaru: I don't really have an explanation for it, I couldn't find one in the original repo. Is there any way you can reach out to the internal UX team to find verification/justification?

src/lib/divider/divider.md
```diff
@@ -0,0 +1,55 @@
...
+<mat-divider [vertical]="true"></mat-divider>
+```
+
+
+### Lists with inset dividers
+
+Dividers can be added to lists as a means of separating content into distinct sections.
+Inset dividers can also be added to provide the appearance of distinct elements in a list without cluttering content
+like avatar images or icons. If combining both, please make sure to avoid adding an inset divider to the last element
+in a list, because it will overlap with the section divider.
```
julianobrasil: What about:

> <strike>If combining both, please m</strike>Make sure to avoid adding an inset divider to the last element in a list, because it will overlap with the section divider

CaerusKaru: Fixed

CaerusKaru:
@DevVersion now that Material is in release, can you please review this?

src/lib/card/card.scss
```diff
@@ -16,6 +16,20 @@ $mat-card-header-size: 40px !default;
   padding: $mat-card-default-padding;
   border-radius: $mat-card-border-radius;
 
+  .mat-divider {
```
josephperrott: can you separate the attributes from the rules here?

```scss
.mat-divider {
  position: absolute;
  left: 0;
  width: 100%;

  [dir='rtl'] & {
    left: auto;
    right: 0;
  }

  &.mat-divider-inset {
    position: static;
    margin: 0;
  }
}
```

src/lib/list/list.scss
```diff
@@ -116,6 +118,26 @@ $mat-dense-list-icon-size: 20px;
     border-radius: 50%;
     padding: 4px;
   }
+
+  .mat-divider {
```
josephperrott: can you separate the attributes from the rules here?

```scss
.mat-divider {
  position: absolute;
  bottom: 0;
  left: 0;
  width: 100%;

  [dir='rtl'] & {
    left: auto;
    right: 0;
  }

  &.mat-divider-inset {
    left: $mat-list-item-inset-divider-offset;
    width: calc(100% - #{$mat-list-item-inset-divider-offset});
    margin: 0;

    [dir='rtl'] & {
      left: auto;
      right: 72px;
    }
  }
}
```

src/lib/card/card.scss
CaerusKaru: Fixed

src/lib/list/list.scss
CaerusKaru: Fixed

josephperrott:
Caretaker Note:  This change should be entirely transparent to end users, but we should take careful note on it to be sure that nothing changes as it is swapping out internal component usage.

PR #5865, opened 2017-07-19
Placeholder overflow is hidden now

JunusErgin:
If the placeholder is very long, the overflow should be hidden and replaced with three dots instead. 
 Otherwise, a long placeholder will be displayed in two lines.


JunusErgin:
I signed it!

PR #5867, opened 2017-07-19
feat: add change emitters to the Intl providers

### LINKED ISSUE #5738: Proposal - Support change detection for Intl services.
Changing text in `I18N` (XXX**Intl** services) should reflect dynamically on components with OnPush change detection stratagy

`Intl` services are simple classes, changing properties will not reflect in `OnPush` components unless they can be notified.

Using libraries like `ngx-translation` or others that support dynamic, on the fly, language replacement..

All

This feature can be supported in multiple ways.  

It requires the `Intl` service to provide an emitter that fires events when a property changes.  
The component using that service can listen to those changes and trigger CD.  
This should probably go though a base `Intl` service that all `Intl` services extend.

There are 2 design aspect to resolve:

### Emitting changes
The logic for emitting can be passive (transparent) or active (user initiated)  
A Passive approach requires each property on the service to be a getter/setter and emit when set.  
Active approach requires the user to run a method once the service is ready to update.

While the passive approach requires nothing from the user, it will require some boilerplate in the library and will emit on every property which isn't suitable for bulk changes although can be handled. The active approach requires an extra step but in most cases it is acceptable since this should happen once a language has changed.

### Handling changes
This is an internal implementation in the components, which can be **property specific** or **global**.  
Property specific means registering each property in a service as an observable and using the templates with `async` pipe to set the value. This approach comes with it's boilerplate and multiple subscriptions but it offers transparent CD integration and rx subscription management since everything is handled by angular. It also means that the services will have to export observables which is somewhat complex.

A global approach means subscribing to a service's change stream and acting upon it.  
In practice it has much less boilerplate and it keeps the same template structure used today, it will only require registering to the change stream, run change detection on each hit and unsubscribing on destruction.

crisbeto:
Since we're switching all components to OnPush change detection, they won't necessarily react to dynamic changes in the i18n labels. These changes add an `EventEmitter` per provider that allow for components to react to those types of changes. Note that it's up to the consumer to call `provider.changes.emit()` in order to trigger the re-render.

Fixes #5738.

### DIFF
--- a/src/lib/datepicker/calendar.spec.ts
+++ b/src/lib/datepicker/calendar.spec.ts
@@ -1,4 +1,4 @@
-import {async, ComponentFixture, TestBed} from '@angular/core/testing';
+import {async, ComponentFixture, inject, TestBed} from '@angular/core/testing';
 import {Component} from '@angular/core';
 import {MdCalendar} from './calendar';
 import {By} from '@angular/platform-browser';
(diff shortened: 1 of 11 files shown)


mmalerba:
LGTM. I really hope we switch to something better in the future though, this was already not a great experience for users and now it's even worse.

crisbeto:
I agree @mmalerba. An alternative can be not to encourage people to override the whole provider, but rather have an injectable with an `update` method that people can use to override the strings.

andrewseguin:
Needs rebase

PR #5879, opened 2017-07-19
build: no longer fallback to index.html file

DevVersion:
* Gulp Connect currently always falls back to the `index.html` file if a resource could not be found. This is not a good idea because this way the browser or Angular is never able to report not-found errors.

image

### DIFF
--- a/tools/gulp/util/task_helpers.ts
+++ b/tools/gulp/util/task_helpers.ts
@@ -148,12 +148,17 @@ export function serverTask(packagePath: string, livereload = true) {
       root: projectDir,
       livereload: livereload,
       port: 4200,
-      fallback: path.join(packagePath, 'index.html'),
       middleware: () => {
         return [httpRewrite.getMiddleware([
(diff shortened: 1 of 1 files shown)


tools/gulp/util/task_helpers.ts
```diff
@@ -148,7 +148,6 @@ export function serverTask(packagePath: string, livereload = true) {
       root: projectDir,
       livereload: livereload,
       port: 4200,
-      fallback: path.join(packagePath, 'index.html'),
```
jelbourn: Isn't this necessary for routing to work?

DevVersion: Good point. I tried to accomplish this using a http rewrite rule. Also added comments now for those rewrite rules.

DevVersion:
(Trying to get the CI run on this PR... not sure why it doesn't work)

tools/gulp/util/task_helpers.ts
DevVersion: So tests are passing. Please have another look.

PR #5361, opened 2017-06-25
fix(menu): align appearance with spec

crisbeto:
Aligns the menu panel appearance with [the spec](https://material.io/guidelines/components/menus.html) by:
* Reducing the elevation from 8 to 2.
* Adding a slight border radius.
* Tweaking the animation to match the examples.

Relates to #3814.

PR #3941, opened 2017-04-06
style(menu): apply theme bg color to menu panel rather than content

willshowell:
Previously, the menu's background opacity did not start animating until the menu had scaled to it's full size, despite the menu's drop shadow existing through the animation.

before-menu-light

Now, the menu's background opacity transitions in sync with the scaling. The content transitions in after scaling is nearly complete.

after-menu-light

[*animations are slowed down 10x*]

cc @kara

PR #5908, opened 2017-07-20
fix(menu): improve menu open animation

m98:
## Description
The current open animation behavior for the menus does not look like to follow the Material spec, and it's not elegant. ([Material.io video demo](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0Bzhp5Z4wHba3WVAwRkdSdURJZWc/Component-Menus-Usage_Appbar_Dropdown_xhdpi_004.webm)).

The menu seems to be somehow broken with the first look, because first you will get an empty (transparent) container which has shadow border, then the content appears inside the container.

You will see slow motion demos in bellow:

### The current behavior of the menu open animation:
ezgif com-video-to-gif

### The new open animation behavior (after the change):
ezgif com-video-to-gif 1

### Side by side view:
animated

## What Changed?
The menu container did not have any background color, and instead of that the content of the menu (`mat-menu-content`) had a white background color, and as far as I know, we don't have any rules in the Material spec that the menu should be transparent until its content appeared.
There is also no need to set the background color for the content of the menu since by this change, we will see just the parent background color.

## Other Material Design Menus
[Material Design Lite](https://getmdl.io/components/index.html#menus-section)
[AngularJS Material](https://material.angularjs.org/latest/demo/menu)
[Material guidelines demos](https://material.io/guidelines/components/menus.html)

### DIFF
--- a/src/lib/menu/_menu-theme.scss
+++ b/src/lib/menu/_menu-theme.scss
@@ -7,7 +7,7 @@
   $background: map-get($theme, background);
   $foreground: map-get($theme, foreground);
 
-  .mat-menu-content {
+  .mat-menu-panel {
     background: mat-color($background, 'card');
   }
 
--- a/src/lib/menu/menu-animations.ts
+++ b/src/lib/menu/menu-animations.ts
@@ -49,7 +49,7 @@ export const transformMenu: AnimationTriggerMetadata = trigger('transformMenu', 
 ]);
 
 /**
- * This animation fades in the background color and content of the menu panel
+ * This animation fades in the content of the menu panel
  * after its containing element is scaled in.
  */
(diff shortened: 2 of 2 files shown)


crisbeto:
The menu animation was changed in https://github.com/angular/material2/pull/5361. It currently looks like this in master:
a

m98:
Thank you @crisbeto 
I fetched the latest pushes before I work on this change, but the menu was working like before! 
Do you know why I did not get latest changes? (Like your PR changed the behavior of the menu, but I still was getting the previous animation behavior)

Your change made the menu much nicer. Thanks :+1:

fxck:
There already was a PR for this https://github.com/angular/material2/pull/3941

m98:
As already discussed, there already was a PR for this #3941

PR #5920, opened 2017-07-20
feat(chip): Add aria-selected to chip

tinayuangao:


src/lib/chips/chip.spec.ts
```diff
@@ -184,6 +193,15 @@ describe('Chips', () => {
           expect(chipInstance.selected).toBeFalsy();
           expect(testComponent.chipSelect).not.toHaveBeenCalled();
         });
+
+        it('should have empty aria-selected', () => {
+          expect(chipNativeElement.getAttribute('aria-selected')).toBe('');
```
jelbourn: Should this be `expect(chipNativeElement.hasAttribute('aria-selected')).toBe(false);`?

tinayuangao:
Our current MdChip's role is `option`: https://github.com/angular/material2/blob/master/src/lib/chips/chip.ts#L59

The reason for this change is screen reader has no indicator when user select/deselect a chip.

PR #5921, opened 2017-07-20
a11y(checkbox): Add checkbox accessibility demo page

tinayuangao:


src/demo-app/a11y/checkbox/checkbox-a11y.html
```diff
@@ -1 +1,42 @@
...
\ No newline at end of file
+<section>
+  <h2>Checkbox without label</h2>
+  <md-checkbox aria-label="A single checkbox with no label example"></md-checkbox>
+</section>
+<section>
+  <h2>Standalone checkbox</h2>
+  <md-checkbox>A simple checkbox</md-checkbox>
+  <md-checkbox [disabled]="true">Disabled checkbox</md-checkbox>
+  <md-checkbox [indeterminate]="true">An indeterminate checkbox</md-checkbox>
```
jelbourn: It would be good to use more "real world" examples here as well.

src/demo-app/a11y/checkbox/checkbox-a11y.html
```diff
@@ -1 +1,42 @@
...
+</section>
+
+<section>
+  <h2>Pseudo checkboxe</h2>
+  <p>These pseudo checkboxes are for display only</p>
+  <md-pseudo-checkbox></md-pseudo-checkbox>
```
jelbourn: Psuedo-checkbox is designed specifically to _not_ be an accessible checkbox so it shouldn't be used in the demo. It's mainly meant to be used inside of `role="option"` elements that express their selection via `aria-selected`

tinayuangao: Removed pseudo checkbox examples

src/demo-app/a11y/checkbox/checkbox-a11y.ts
```diff
@@ -8,5 +13,46 @@ import {Component} from '@angular/core';
...
+  ];
+
+  allComplete(task: Task): boolean {
+    let subtasks = task.subtasks;
+
+    if (!subtasks) {
+      return false;
+    }
+
+    return subtasks.every(t => t.completed) ? true
```
mmalerba: why doesn't this just return `subtasks.every(t => t.completed)`?

tinayuangao:
Please take another look.  Thanks

mmalerba:
LGTM

PR #5922, opened 2017-07-20
build: change cdk to use secondary entry-points

jelbourn:
Work for follow-up PRs:
* Generating the `tsconfig-build.json` for each entry-point instead of keeping them in source
* Reducing the number of times the CDK entry-points are all manually listed out
* Update the plunker templates for the docs site
* Changing `keyboard` dir to `keycodes` and changing `observe-content` to `observers`
* Making each `public_api` actually reflect our real public API instead of everything.

BREAKING CHANGE: `@angular/cdk` is now organized into standalone entry-points. You no longer import code directly from `@angular/cdk`. Instead, you import from the more specific sub-package, e.g., 
```ts
import {LiveAnnouncer} from '@angular/cdk/a11y';
```


src/cdk/tsconfig-tests.json
```diff
@@ -8,13 +8,16 @@
...
+    "paths": {
+      "@angular/cdk/*": ["./*"]
+    }
   },
   "files": [
-    "./index.ts",
     "./testing/index.ts"
   ],
   "include": [
+    "./*/index.ts",
```
DevVersion: Maybe add a comment for this.

src/lib/tsconfig-tests.json
```diff
@@ -11,7 +11,15 @@
...
-      "@angular/cdk": ["../../dist/packages/cdk"],
+      "@angular/cdk/a11y": ["../../dist/packages/cdk/a11y"],
+      "@angular/cdk/bidi": ["../../dist/packages/cdk/bidi"],
+      "@angular/cdk/coercion": ["../../dist/packages/cdk/coercion"],
+      "@angular/cdk/platform": ["../../dist/packages/cdk/platform"],
+      "@angular/cdk/portal": ["../../dist/packages/cdk/portal"],
+      "@angular/cdk/keyboard": ["../../dist/packages/cdk/keyboard"],
+      "@angular/cdk/rxjs": ["../../dist/packages/cdk/rxjs"],
+      "@angular/cdk/table": ["../../dist/packages/cdk/table"],
+      "@angular/cdk/observe-content": ["../../dist/packages/cdk/observe-content"],
```
DevVersion: I think here without the `public_api` this can be just `@angular/cdk/*: ["../../dist/packages/cdk/*"]`

tools/package-tools/build-bundles.ts
```diff
@@ -1,57 +1,101 @@
...
-  const moduleName = `ng.${packageName}`;
 
-  // List of paths to the package bundles.
-  const fesm2015File = join(bundlesDir, `${packageName}.js`);
-  const fesm2014File = join(bundlesDir, `${packageName}.es5.js`);
-  const umdFile = join(bundlesDir, `${packageName}.umd.js`);
-  const umdMinFile = join(bundlesDir, `${packageName}.umd.min.js`);
+/** Builds bundles for the primary entry-point w/ given entry file, e.g. @angular/cdk */
+export async function buildPrimaryEntryPointBundles(entryFile: string, packageName: string) {
+  return createBunldesForEntryPoint({
```
DevVersion: There is a typo here. This should be `createBundlesForEntryPoint`.

tools/package-tools/build-bundles.ts
```diff
@@ -1,57 +1,101 @@
...
+    entryFile,
+    moduleName:   `ng.${packageName}`,
+    fesm2015Dest: join(bundlesDir, `${packageName}.js`),
+    fesm2014Dest: join(bundlesDir, `${packageName}.es5.js`),
+    umdDest:      join(bundlesDir, `${packageName}.umd.js`),
+    umdMinDest:   join(bundlesDir, `${packageName}.umd.min.js`),
```
DevVersion: Nit (optional): I kinda hate indenting things using spaces like that. Up to preference though :)

tools/package-tools/build-release.ts
```diff
@@ -26,14 +29,73 @@ export function composeRelease(packageName: string) {
   inlinePackageMetadataFiles(packagePath);
 
   copyFiles(packagePath, '**/*.+(d.ts|metadata.json)', join(releasePath, 'typings'));
-  copyFiles(bundlesDir, `${packageName}.umd?(.min).js?(.map)`, join(releasePath, 'bundles'));
+  copyFiles(bundlesDir, `*.umd?(.min).js?(.map)`, join(releasePath, 'bundles'));
```
DevVersion: How about something like:

```ts
copyFiles(bundlesDir, `${packageName}?(-*).umd.js?(.map)`, join(releasePath, 'bundles'));
```

This way we can ensure that the bundles only for the current package are copied. Just in case there are other bundles inside of the shared `bundles/` folder.

tools/package-tools/build-release.ts
```diff
@@ -17,7 +20,7 @@ const bundlesDir = join(outputDir, 'bundles');
  * release folder structure. The output will also contain a README and the according package.json
  * file. Additionally the package will be Closure Compiler and AOT compatible.
  */
-export function composeRelease(packageName: string) {
+export function composeRelease(packageName: string, options: ReleaseBuildOptions = {}) {
```
DevVersion: Maybe `ReleaseComposeOptions` here? To be consistent with the function name. Or we change `composeRelease` to `buildRelease`.

tools/package-tools/build-release.ts
```diff
@@ -26,14 +29,73 @@ export function composeRelease(packageName: string) {
...
+    createFilesForSecondaryEntryPoint(packageName, packagePath, releasePath);
+  }
+
   addPureAnnotationsToFile(join(releasePath, '@angular', `${packageName}.es5.js`));
 }
+
+/**
+ * Creates files necessary for a secondary entry-point.
+ * @param packageName The name of the package for which to create entry-point files.
+ * @param packagePath
```
DevVersion: This seems to miss a description. In my opinion we don't need those param JSDoc descriptions at all. It feels pretty clear to me.

tools/package-tools/build-release.ts
```diff
@@ -26,14 +29,73 @@ export function composeRelease(packageName: string) {
...
+    // Finally, create both a d.ts and metadata file for this entry-point in the root of
+    // the package that re-exports from the entry-point's directory.
+    createTypingsReexportFile(releasePath, `./${entryPointName}/index`, entryPointName);
+    createMetadataReexportFile(releasePath, `./${entryPointName}/index`, entryPointName);
+  });
+}
+
+
+/** Creates a package.json file for a secondary entry-point within a package. */
+function createEntryPointPackageJson(destDir: string, packageName: string, entryPointName: string) {
```
DevVersion: This should also go into a separate file as every other "helper" function we recently moved out (fine as follow-up)

tools/package-tools/rollup-helpers.ts
```diff
@@ -22,8 +22,29 @@ const ROLLUP_GLOBALS = {
...
 
+  '@angular/cdk/a11y/index': 'ng.cdk.a11y',
+  '@angular/cdk/bidi/index': 'ng.cdk.bidi',
+  '@angular/cdk/coercion/index': 'ng.cdk.coercion',
+  '@angular/cdk/keyboard/index': 'ng.cdk.keyboard',
+  '@angular/cdk/observe-content/index': 'ng.cdk.observeContent',
+  '@angular/cdk/platform/index': 'ng.cdk.platform',
+  '@angular/cdk/portal/index': 'ng.cdk.portal',
+  '@angular/cdk/rxjs/index': 'ng.cdk.rxjs',
+  '@angular/cdk/table/index': 'ng.cdk.table',
```
DevVersion: And this is the problem with the automatic expansion of the import shorthands right? Maybe add a comment on the top of those "globals"?

tools/package-tools/build-bundles.ts
DevVersion: Did you mean to update this? Looks still like the old one.

tsconfig.json
```diff
@@ -17,6 +17,15 @@
...
       "@angular/cdk": ["./src/cdk/public_api.ts"],
+      "@angular/cdk/a11y": ["./src/cdk/a11y"],
+      "@angular/cdk/bidi": ["./src/cdk/bidi"],
+      "@angular/cdk/coercion": ["./src/cdk/coercion"],
+      "@angular/cdk/keyboard": ["./src/cdk/keyboard"],
+      "@angular/cdk/observe-content": ["./src/cdk/observe-content"],
+      "@angular/cdk/platform": ["./src/cdk/platform"],
+      "@angular/cdk/portal": ["./src/cdk/portal"],
+      "@angular/cdk/rxjs": ["./src/cdk/rxjs"],
+      "@angular/cdk/table": ["./src/cdk/table"],
```
DevVersion: This can be also dynamic using the wildcard right?

src/lib/table/table.spec.ts
```diff
@@ -1,6 +1,6 @@
 import {async, ComponentFixture, TestBed} from '@angular/core/testing';
 import {Component, ViewChild} from '@angular/core';
-import {CdkTable, DataSource, CdkTableModule} from '@angular/cdk';
+import {CdkTable, DataSource, CdkTableModule} from '@angular/cdk/table';
 import {} from '../core/data-table/data-source';
```
crisbeto: Not something that was added here, but this seems redundant.

tools/gulp/tasks/payload.ts
```diff
@@ -25,10 +26,10 @@ task('payload', ['material:clean-build'], async () => {
...
     material_fesm_2014: getBundleSize('material.es5.js'),
     // CDK bundles
-    cdk_umd: getBundleSize('cdk.umd.js'),
-    cdk_umd_minified_uglify: getBundleSize('cdk.umd.min.js'),
-    cdk_fesm_2015: getBundleSize('cdk.js'),
-    cdk_fesm_2014: getBundleSize('cdk.es5.js'),
+    cdk_umd: getBundleSize('cdk*.umd.js'),
+    cdk_umd_minified_uglify: getBundleSize('cdk*.umd.min.js'),
+    cdk_fesm_2015: getBundleSize('cdk.js') + getBundleSize('cdk/!(*.es5).js'),
+    cdk_fesm_2014: getBundleSize('cdk.es5.js') + getBundleSize('cdk/**.es5.js'),
```
crisbeto: Is the pattern here missing a slash (e.g. `cdk/**/*.es5.js`)? Otherwise the second * seems redundant.

tools/package-tools/build-bundles.ts
jelbourn: Done

tsconfig.json
jelbourn: Done

src/lib/table/table.spec.ts
jelbourn: Done

tools/gulp/tasks/payload.ts
jelbourn: Removed superfluous `*`

PR #5948, opened 2017-07-21
a11y(button-toggle): Add accessibility demo page for button toggle

tinayuangao:
Added default `role=group` for button toggle group with multi-selection

### DIFF
--- a/src/demo-app/a11y/button-toggle/button-toggle-a11y.html
+++ b/src/demo-app/a11y/button-toggle/button-toggle-a11y.html
@@ -5,7 +5,7 @@
 
 <section>
   <h2>Button toggles with icons</h2>
-  <md-button-toggle-group name="alignment">
+  <md-button-toggle-group name="alignment" aria-label="Alignments">
     <md-button-toggle value="left" aria-label="Align left">
       <md-icon>format_align_left</md-icon>
     </md-button-toggle>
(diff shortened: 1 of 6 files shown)


src/demo-app/a11y/button-toggle/button-toggle-a11y.html
```diff
@@ -0,0 +1,62 @@
+<section>
+  <h2>Single button toggle</h2>
+  <md-button-toggle>Yes</md-button-toggle>
+</section>
+
+<section>
+  <h2>Button toggles with icons</h2>
+  <md-button-toggle-group name="alignment">
```
jelbourn: For single-select groups, they should have an `aria-label` or `aria-labelledby` the same as radio buttons.

For multi-select, I'm just now realizing that it might make sense to put `role="group"` on the group with an aria-label. How does the experience change with that?

tinayuangao: Added `role="group"` to button-toggle-group with multi-select. Without it the aria-label does not work.

jelbourn: What kind of experience did you get when adding `role="group"`?

tinayuangao: Without `role="group"`: when I moved focus to the multi-select button toggle group, the screen reader said "Flour, checkbox not checked".
With `role="group"`: when I moved the focus to the multi-select button toggle group, the screen reader said "Groceries group: Flour checkbox not checked"

PR #5951, opened 2017-07-21
feat(stepper): Create stepper button directives to enable adding buttons to stepper

g1shin:
Allow users to add buttons rather than steppers containing buttons.


src/cdk/stepper/stepper-button.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive} from '@angular/core';
+import {CdkStepper} from './stepper';
+
+@Directive({
```
jelbourn: Add simple class description for both directives? Can just be something like 
```ts
/** Button that moves to the next step in a stepper workflow. */
```

src/cdk/stepper/stepper-button.ts
```diff
@@ -0,0 +1,38 @@
...
+  selector: 'button[cdkStepperNext]',
+  host: {
+    '(click)': '_onClick()',
```
jelbourn: Instead of creating an `_onClick` handler, you can just do
```ts
host: {
  '(click)': '_stepper.next()',
}
```
(would have to make `_stepper` not private; just keep the underscore as it implies "internal"). Would condense the entire class down to
```ts
@Directive({
  selector: 'button[cdkStepperNext]',
  host: {'(click)': '_stepper.next()'}
})
export class CdkStepperNext {
  constructor(public _stepper: CdkStepper) { }
}
```

src/lib/stepper/stepper-button.ts
```diff
@@ -0,0 +1,35 @@
...
+import {CdkStepperNext, CdkStepperPrevious} from '@angular/cdk';
+import {MdStepper} from './stepper';
+
+@Directive({
+  selector: 'button[mdStepperNext], button[matStepperNext]',
+  host: {
+    '(click)': '_onClick()',
+  }
+})
+export class MdStepperNext extends CdkStepperNext {
```
jelbourn: Does it work if you just do this?
```ts
@Directive({selector: 'button[mdStepperNext], button[matStepperNext]'})
export class  MdStepperNext extends CdkStepperNext { }
```
I thought the click handler would be inherited automatically and that the constructor wouldn't be necessary if `MdStepper` provides itself as `CdkStepper`

src/cdk/stepper/stepper-button.ts
g1shin: Since the click handler doesn't seem to be inherited, I think it would be necessary to create the `_onClick` handler that the child directive can access.

src/lib/stepper/stepper-button.ts
g1shin: https://github.com/angular/angular/issues/18331
The click handler does not seem to be inherited automatically.

g1shin:
Changes have been made based on review; ready for review again.

src/cdk/stepper/stepper-button.ts
jelbourn: I don't quite follow what you mean; even without the click handler being inherited, you can still directly have `host: {'(click)': '_stepper.next()'}`

mmalerba: _stepper is still a public member so you don't need _onClick

g1shin:
Ready for review :+1:

PR #5953, opened 2017-07-21
fix: add camelCase ng-content selectors

andrewseguin:
Adding selectors that should be used, will remove the other dash-cased versions in another PR

### DIFF
--- a/src/lib/list/list-item.html
+++ b/src/lib/list/list-item.html
@@ -5,7 +5,13 @@
   </div>
 
   <ng-content
-      select="[md-list-avatar],[md-list-icon], [mat-list-avatar], [mat-list-icon]"></ng-content>
-  <div class="mat-list-text"><ng-content select="[md-line], [mat-line]"></ng-content></div>
+      select="[md-list-avatar], [md-list-icon], [mat-list-avatar], [mat-list-icon],
+              [mdListAvatar], [mdListIcon], [matListAvatar], [matListIcon]">
+  </ng-content>
+
+  <div class="mat-list-text">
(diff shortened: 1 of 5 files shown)


jelbourn:
Actually `list.md` needs to be updated too

andrewseguin:
Got it, changed the `md`

src/lib/list/list-item.html
```diff
@@ -5,7 +5,13 @@
   </div>
 
   <ng-content
-      select="[md-list-avatar],[md-list-icon], [mat-list-avatar], [mat-list-icon]"></ng-content>
-  <div class="mat-list-text"><ng-content select="[md-line], [mat-line]"></ng-content></div>
+      select="[md-list-avatar], [md-list-icon], [mat-list-avatar], [mat-list-icon],
+              [mdListAvatar], [mdListIcon], [matListAvatar], [matListIcon]">
+  </ng-content>
+
+  <div class="mat-list-text">
```
alexsapps: lines 12-14 have to be on one line so the ".mat-list-text:empty" css selector works.

alexsapps: A little more context: The :empty selector is used to set display:none on the mat-list-text div so it doesn't take up space in the flexbox container leaving no room for anything else. The actual text seems to come after that div rather than inside it.

https://github.com/angular/material2/blob/5bc97ec89462a2dc11fdeccf69ee41eadc265bd4/src/lib/list/list.scss#L80
https://github.com/angular/material2/blob/dcc857664c9313694d020845135f1ce63a66f43a/src/lib/core/style/_list-common.scss#L46

PR #5955, opened 2017-07-21
fix(compat): add camelCase selectors to tab label wrapper

andrewseguin:


src/lib/tabs/tab-label-wrapper.ts
```diff
@@ -19,7 +19,8 @@ export const _MdTabLabelWrapperMixinBase = mixinDisabled(MdTabLabelWrapperBase);
  * @docs-private
  */
 @Directive({
-  selector: '[md-tab-label-wrapper], [mat-tab-label-wrapper]',
+  selector: '[md-tab-label-wrapper], [mat-tab-label-wrapper], ' +
```
jelbourn: Isn't this directive purely internal? If so, we should be able to just remove the dash-case selectors.

andrewseguin: That's true - removed the dash

PR #5958, opened 2017-07-22
a11y(radio): Add radio buttons accessibility demo page

tinayuangao:


src/demo-app/a11y/radio/radio-a11y.html
```diff
@@ -0,0 +1,32 @@
+<section>
+  <h2>Radio buttons in group</h2>
+  <md-radio-group name="seasons" [(ngModel)]="favoriteSeason">
```
jelbourn: Each radio-group should have an `aria-label` (e.g. "Seasons") or an `aria-labelledby` ([example](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html))

src/demo-app/a11y/radio/radio-a11y.html
```diff
@@ -0,0 +1,32 @@
...
+    </md-radio-button>
+  </md-radio-group>
+</section>
+
+<section>
+  <h2>Radio buttons with align-end label position</h2>
+  <md-radio-group name="my_options" align="end">
+    <md-radio-button value="option_1">Option 1</md-radio-button>
+    <md-radio-button value="option_2">Option 2</md-radio-button>
+    <md-radio-button value="option_3">Option 3</md-radio-button>
```
jelbourn: It would be good to use more meaningful options (like the seasons example does). Good examples could be pizza toppings, ice cream flavors, varieties of tea, types of bread, etc.

src/demo-app/a11y/radio/radio-a11y.html
```diff
@@ -0,0 +1,32 @@
...
+  <h2>Radio buttons with align-end label position</h2>
+  <md-radio-group name="bread" align="end" aria-label="Breads">
+    <md-radio-button value="toast">Toast</md-radio-button>
+    <md-radio-button value="biscuit">Biscuit</md-radio-button>
+    <md-radio-button value="bagel">Bagel</md-radio-button>
+  </md-radio-group>
+</section>
+
+<section>
+  <h2>Disabled radio buttons</h2>
```
jelbourn: Maybe add a `<p>` under the header like
```html
<p>This section contains three radio buttons for "Yes", "No", and "Maybe". The "Maybe" radio button is disabled.</p>
```

PR #5959, opened 2017-07-22
build: add 'prefer-const' tslint rule

rafaelss95:


rafaelss95:
**Note:** This PR is here to ask if you're interested in adding this rule. If so, I can go ahead and fix all the incoming lint issues.

CC @jelbourn.

jelbourn:
While I agree we should prefer `const`, I'm concerned about the huge number of line changes this would cause. It would end up causing a lot of merge conflicts without providing any major value

rafaelss95:
@jelbourn  Indeed. Maybe try to fix component by component (step by step) or just close this PR?

jelbourn:
I'd be okay with just closing since I don't think the preference for `const` over `let` makes a big enough difference to bother

PR #5963, opened 2017-07-22
build: add tslint rule to verify rollup globals config

crisbeto:
Adds a custom tslint rule that ensures that we've added all of the necessary external modules to the Rollup config. This is helpful, because forgetting to add a module will log a warning, but it won't break the build necessarily, which is why we occasionally need PRs like #5930.


tools/tslint-rules/missingRollupExternalsRule.js
```diff
@@ -0,0 +1,41 @@
...
+    this._configPath = args.config;
+    this._config = this._scope.length ? require(args.config) : {};
+  }
+
+  visitImportDeclaration(node) {
+    // Parse out the module name. The first and last characters are the quote marks.
+    const module = node.moduleSpecifier.getText().slice(1, -1);
+
+    // Check whether we match any of the scopes while missing the module from the config.
+    if (this._scope.some(p => p.test(module)) && !this._config[module]) {
```
DevVersion: More like a question. Why do we need such a scope field? Actually every package import should be listed in the globals config?

crisbeto: It's primarily because we've only had issues with Angular and Rx packages until now since they're the only dependencies that we have. I can switch it to look for all external packages instead, if it makes sense.

DevVersion: I see what you mean, but Rollup will not only complain about the dependencies/scopes you restricted the rule to. Basically every external import will be problematic.

crisbeto: Makes sense, I'll change it. While we're at it, I'm not too sure how to handle the `@angular/material-examples` package. Currently I've added an exception for it.

DevVersion: Can you explain what you mean? What's special for the examples package?

crisbeto: What is the proper Rollup mapping for it? E.g. for `@angular/core` it's `ng.core`, but I'm not sure whether the examples have something like that.

DevVersion: Yeah that doesn't really matter that much since it won't be really a public package.

As for convention I'd map it to something like `ng.material-examples`.

crisbeto: Done. Also made it resolve the path relative to the project root, instead of the rule file.

andrewseguin:
Please rebase

crisbeto:
Rebased @andrewseguin.

andrewseguin:
Got some lint errors coming up in this PR

PR #5966, opened 2017-07-22
fix(button): focus styles not applied to programmatically focused buttons

### LINKED ISSUE #7510: MdButton: 'focus()' API method doesn't focus the button
Bug(?)

The button should become focused after calling the API method, so that the user visually sees it's been focused and they can hit the `enter` key to activate the button, etc.

Nothing happens visually, and there are no console errors output.

StackBlitz using latest `npm`: https://stackblitz.com/edit/angular-material2-issue-wckfr5 <br/>

When the user makes a form control selection I would like to focus the submit button for improved UX flow.

> C:\src\git\myProject>ng version
> @angular/cli: 1.4.3
> node: 7.2.1
> os: win32 x64
> @angular/animations: 4.4.3
> @angular/cdk: 2.0.0-beta.11
> @angular/common: 4.4.3
> @angular/compiler: 4.4.3
> @angular/core: 4.4.3
> @angular/flex-layout: 2.0.0-beta.9-0da1f88
> @angular/forms: 4.4.3
> @angular/http: 4.4.3
> @angular/material: 2.0.0-beta.11
> @angular/platform-browser: 4.4.3
> @angular/platform-browser-dynamic: 4.4.3
> @angular/router: 4.4.3
> @angular/cli: 1.4.3
> @angular/compiler-cli: 4.4.3
> @angular/language-service: 4.4.3
> typescript: 2.5.3

Am I misunderstanding what the focus API method is meant to do?

crisbeto:
Currently the tint that is added on top of focused buttons won't show up if the button is focused programmatically, which means that the button won't appear focused in cases like the dialog closing and restoring focus to its trigger. This seems to have been introduced by 5d6920dde6cbe17f2888b7e09f123e02db251b0f.

Fixes #7510.

### DIFF
--- a/src/demo-app/button/button-demo.html
+++ b/src/demo-app/button/button-demo.html
@@ -73,11 +73,12 @@
       <button mat-raised-button (click)="button1.focus()">Focus 1</button>
       <button mat-raised-button (click)="button2.focus()">Focus 2</button>
       <button mat-raised-button (click)="button3.focus()">Focus 3</button>
+      <button mat-raised-button (click)="button4.focus()">Focus 4</button>
     </div>
     <button mat-button #button1 [disabled]="isDisabled" (click)="clickCounter=clickCounter+1">off</button>
-    <button mat-button color="primary" [disabled]="isDisabled">off</button>
-    <a href="http://www.google.com" #button2 mat-button color="accent" [disabled]="isDisabled">off</a>
-    <button mat-raised-button #button3 color="primary" [disabled]="isDisabled">off</button>
+    <button mat-button #button2 color="primary" [disabled]="isDisabled">off</button>
+    <a href="http://www.google.com" #button3 mat-button color="accent" [disabled]="isDisabled">off</a>
+    <button mat-raised-button #button4 color="primary" [disabled]="isDisabled">off</button>
     <button mat-mini-fab [disabled]="isDisabled">
       <mat-icon>check</mat-icon>
     </button>
--- a/src/demo-app/demo-app/demo-app.html
+++ b/src/demo-app/demo-app/demo-app.html
@@ -24,7 +24,7 @@
   </mat-sidenav>
   <div>
     <mat-toolbar color="primary">
-      <button mat-icon-button (click)="start.open()">
+      <button mat-icon-button (click)="start.open('mouse')">
         <mat-icon>menu</mat-icon>
       </button>
       <div class="demo-toolbar">
--- a/src/demo-app/sidenav/sidenav-demo.html
+++ b/src/demo-app/sidenav/sidenav-demo.html
@@ -9,9 +9,9 @@
                  [fixedInViewport]="fixed" [fixedTopGap]="fixedTop" [fixedBottomGap]="fixedBottom">
       Start Side Sidenav
       <br>
-      <button mat-button (click)="start.close()">Close</button>
+      <button mat-button (click)="start.close('mouse')">Close</button>
       <br>
-      <button mat-button (click)="end.open()">Open End Side</button>
+      <button mat-button (click)="end.open('keyboard')">Open End Side</button>
       <br>
       <button mat-button (click)="modeIndex = (modeIndex + 1) % 3">Toggle Mode</button>
       <div>Mode: {{start.mode}}</div>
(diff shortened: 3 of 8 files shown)


mmalerba:
So it sounds like the motivation for this change was a specific issue with the dialog closing. In general for most cases do you think programmatic focus should default to looking like keyboard or like mouse? If keyboard, we should make this change. Otherwise we should leave the styles the same and change dialog to inject `FocusOriginMonitor` and use the `focusVia` method to pretend it was a keyboard focus, since it's more of an exception than the normal case. I don't have a strong preference on what the default should be, just wanted to point out the difference and make sure we make whatever change we make deliberately.

crisbeto:
IMO all programmatic focus should look like the keyboard one, otherwise the user won't have any indication that something is focused. It's different for the mouse focus since the user is keeping track of their cursor and know what they clicked on so the indication isn't necessary.

mmalerba:
I think the reason I originally made it act like mouse is because of the situation where you click a button to open a sidenav and then the sidenav opens and programmatically focuses its first element. We didn't want the focus styles on that first element. But maybe we can consider that more of the exception case and add some code to do `focusVia`.

Also if we're going to do this for button let's be consistent, can you search our CSS for `cdk-keyboard-focused` and make sure that we always put the same styles for `cdk-program-focused`?

crisbeto:
Sure. I'll update this PR and ping you.

crisbeto:
Updated the 3 other cases we were using it @mmalerba.

mmalerba:
Can we change the sidenav to use `focusVia` so that we don't have a regressions with https://github.com/angular/material2/issues/3215? It's kinda tricky I think, because of the user opens the sidenav with the keyboard we want to show the focus style on the first element, but if they open it with the mouse we don't.

crisbeto:
Done @mmalerba.

src/lib/sidenav/sidenav.ts
```diff
@@ -161,7 +162,7 @@ export class MdSidenav implements AfterContentInit, OnDestroy {
     let activeEl = this._doc && this._doc.activeElement;
     if (activeEl && this._elementRef.nativeElement.contains(activeEl)) {
       if (this._elementFocusedBeforeSidenavWasOpened instanceof HTMLElement) {
-        this._elementFocusedBeforeSidenavWasOpened.focus();
+        this._focusOriginMonitor.focusVia(this._elementFocusedBeforeSidenavWasOpened, 'program');
```
mmalerba: This is going to do the same thing that just calling `.focus()` was doing. I think we want logic like this:
1. If the sidenav close is triggered by keyboard interaction (e.g. pressing escape) set it to `keyboard` focused
2. If the sidenav close is triggered by mouse interaction (e.g. clicking on backdrop) set it to `mouse` focused
3. If the sidenav close is triggered by directly calling `close` or `toggle` method allow the user to optionally supply the focus origin and default to `program` if they don't specify.

I think opening the sidenav needs similar logic but it's a little trickier because the sidenav isn't directly responsible for setting the focus on open, it goes through the focus trap.

You can test if this is working properly by opening and closing the sidenav in the demo app via different means and seeing what the resulting focus effect looks like

crisbeto: I'm not sure how that would work,  considering that we don't have the equivalent of mdMenuTriggerFor for the sidenav. Alternatively, I can fix this only for the dialog where we have more control over the closing logic.

mmalerba: I think the closing logic I described is fairly straightforward to implement. For open you can just allow the user to pass an origin to the `open` method and it will default to `program` if they don't specify. That way if people really care about getting rid of the focus indicator they can add logic for it manually. It might also be nice to make a `mdSidenavTrigger` directive that does the focus origin logic for them.

crisbeto: That makes sense. I was trying to think of something that would work automagically.

crisbeto:
Reworked this one based on what we discussed a while ago @mmalerba, can you take another look?

mmalerba:
can you update the demo app's sidenav as well, to avoid this bug:
1) click hamburger
2) click new demo page in sidenav
3) page loads & sidenav closes, observe focus style on hamburger menu after close (should not appear focused)

crisbeto:
Done @mmalerba.

PR #5967, opened 2017-07-22
feat(select): allow disabling ripples for options

DevVersion:
* Adds a binding that allows developers to disable ripples for every option in a select.

### DIFF
--- a/src/lib/core/option/option.ts
+++ b/src/lib/core/option/option.ts
@@ -22,7 +22,6 @@ import {ENTER, SPACE} from '../keyboard/keycodes';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {MATERIAL_COMPATIBILITY_MODE} from '../../core/compatibility/compatibility';
 import {MdOptgroup} from './optgroup';
-import {CanDisableRipple, mixinDisableRipple} from '../common-behaviors/disable-ripple';
 
 /**
  * Option IDs need to be unique across components, so this counter exists outside of
@@ -35,19 +34,12 @@ export class MdOptionSelectionChange {
   constructor(public source: MdOption, public isUserInput = false) { }
 }
 
-// Boilerplate for applying mixins to MdOption.
-/** @docs-private */
-export class MdOptionBase {}
-export const _MdOptionMixinBase = mixinDisableRipple(MdOptionBase);
-
-
 /**
  * Single option inside of a `<md-select>` element.
  */
 @Component({
   moduleId: module.id,
   selector: 'md-option, mat-option',
-  inputs: ['disableRipple'],
   host: {
     'role': 'option',
     '[attr.tabindex]': '_getTabIndex()',
@@ -66,10 +58,11 @@ export const _MdOptionMixinBase = mixinDisableRipple(MdOptionBase);
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class MdOption extends _MdOptionMixinBase implements CanDisableRipple {
+export class MdOption {
   private _selected: boolean = false;
   private _active: boolean = false;
   private _multiple: boolean = false;
+  private _disableRipple: boolean = false;
 
   /** Whether the option is disabled.  */
   private _disabled: boolean = false;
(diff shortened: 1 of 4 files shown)


src/lib/core/option/option.ts
```diff
@@ -99,17 +92,22 @@ export class MdOption extends _MdOptionMixinBase implements CanDisableRipple {
   get disabled() { return (this.group && this.group.disabled) || this._disabled; }
   set disabled(value: any) { this._disabled = coerceBooleanProperty(value); }
 
+  /** Whether ripples for the option are disabled. */
+  @Input()
```
crisbeto: I don't think that this should be an Input. The only way to disable ripples on options should be programmatic. It also means that we won't have to coerce the value.

src/lib/select/select.spec.ts
```diff
@@ -1800,6 +1800,31 @@ describe('MdSelect', () => {
...
+      });
+
+      it('should not show ripples for options if disableRipple is set to true', () => {
+        fixture.componentInstance.disableRipple = true;
+        fixture.detectChanges();
+
+        expect(overlayContainerElement.querySelectorAll('.mat-ripple-element').length)
+          .toBe(0, 'Expected no ripples to show up initially.');
+
+        dispatchFakeEvent(options[0], 'mousedown');
```
crisbeto: I'm not sure whether we care about the DOM events that are happening in these tests since this is tested well enough in the ripple tests. Looping through all the options and verifying that the flag is correct should be enough. Also it could use a test where it adds a new option and check that it got the proper `disableRipple` value.

src/lib/core/option/option.ts
DevVersion: The ability to disable the ripples from a binding has been already added in https://github.com/angular/material2/commit/addf1ce8b124fe7a242bf927e59ea69d00513f85. I don't really see anything that would go against just providing this flexibility.

src/lib/select/select.spec.ts
DevVersion: Not sure about that. This test would just ensure that the `disableRipple` property changed properly and also got reflected properly in the view. Verifying for example that `markForCheck` is placed properly.

I agree on adding a test for the property itself though.

crisbeto: My issue with this approach is that it's testing functionality that it's not responsible for. That functionality already has tests of its own.

DevVersion: That's an argument. I will change it.

src/lib/core/option/option.ts
crisbeto: I think that there might be some overhead to having to maintain the input. Also there's the extra JS being generated because of the getter and setter, although we can't really avoid those because it has to `markForCheck`.

DevVersion: It might be true that it might not happen that someone just disables ripples for a single option.

From the payload size perspective it shouldn't make any difference, so it's more a question of API design.

I'm fine removing the `@Input` here. Also since https://github.com/angular/material2/commit/addf1ce8b124fe7a242bf927e59ea69d00513f85 is not in `master` that long, I assume it won't be a breaking change.

DevVersion: Updated the tests. Please have another look.

src/lib/select/select.spec.ts
```diff
@@ -297,6 +297,18 @@ describe('MdSelect', () => {
...
+    it('should update disableRipple properly on each option', () => {
+      const options = fixture.componentInstance.options.toArray();
+
+      expect(options.every(option => option.disableRipple === false))
+        .toBeTruthy('Expected all options to have disableRipple set to false initially.');
+
+      fixture.componentInstance.disableRipple = true;
+      fixture.detectChanges();
+
+      expect(options.every(option => option.disableRipple === true))
```
crisbeto: Can be reduced to `option => option.disableRipple`.

DevVersion: True. But I want to make it clear that we are checking for a truthy value here.

andrewseguin:
Rebase please

DevVersion:
@andrewseguin Done.

PR #5968, opened 2017-07-22
chore: use list example component for e2e

DevVersion:
* The e2e app now uses the list overview example for the list e2e tests.


DevVersion:
@jelbourn Maybe we should also add a subheader to the overview example. Not sure how important it would be for the screenshot tests.

PR #5978, opened 2017-07-23
fix(connected-position-strategy): position change event not emitting for fallback positions

crisbeto:
* Fixes the `onPositionChange` event from the `ConnectedPositionStrategy` not emitting if the strategy picks one of the fallback positions.
* Fixes the autocomplete panel not having the proper offset when it is in the `above` position. This was tightly-coupled to the `onPositionChange` not firing.
* Fixes a couple of tests that were nested in another test, causing them not to be executed. I've removed one of the tests since it was also duplicated with one above. This seems to have been the result of a faulty conflict resolution.


crisbeto:
@jelbourn, I have a hunch that the flakes we were getting when we started randomizing the tests were due to the two tests that were nested in another test. What I think was happening was that they would eventually get executed, if the randomization was just right, causing one of them to fail. The failing one also had the assertions from a completely different test.

PR #5984, opened 2017-07-23
docs: edit mdtab position and origin to not be exposed to api docs

zbagley:


zbagley:
Closed due to wrong e-mail being associated with commit.

PR #5996, opened 2017-07-24
fix(menu): prevent user from accidentally opening a sub-menu while animating

crisbeto:
Makes the menu panel non-interactive while it is animating, preventing the user from accidentally triggering a sub-menu.

For reference, here's what it used to look like if the user moved their cursor while animating:
a


m98:
That's a good UX fix! Thanks

PR #6007, opened 2017-07-24
fix(datepicker): mark input tocuhed on calendar date selected

### LINKED ISSUE #5074: Datepicker doesn't set touched property on input, when using popup UI
proposal

When I use datepicker's selector UI (clicking on icon, showing the popup) and selecting the date it should set touched on a input element. 

When I click the icon and select the date the touched property isn't set. Its only set when I start typing and leave the field.

Better support for validation UX, for instance when I can't used touched property. When someone starts typing into the field they will immediately see the validation because again touched property doesn't work in all cases.

mmalerba:
fixes #5074


andrewseguin:
LGTM, do you want to add a test similar to `should mark input touched on blur`?

PR #6019, opened 2017-07-25
fix(chips): cursor on remove button and box-shadow transition

crisbeto:
* Adds a `pointer: cursor` to the chip remove button to indicate that it's clickable.
* Adds a transition to the focused chip `box-shadow` to make it less harsh.
* Breaks up a long line that was exceeding our character limit.

### DIFF
--- a/src/lib/chips/chips.scss
+++ b/src/lib/chips/chips.scss
@@ -16,8 +16,9 @@ $mat-chips-chip-margin: 8px;
 }
 
 .mat-chip:not(.mat-basic-chip) {
+  @include mat-elevation-transition;
   display: inline-flex;
-  padding: $mat-chip-vertical-padding $mat-chip-horizontal-padding $mat-chip-vertical-padding $mat-chip-horizontal-padding;
+  padding: $mat-chip-vertical-padding $mat-chip-horizontal-padding;
   border-radius: $mat-chip-horizontal-padding * 2;
   align-items: center;
   cursor: default;
@@ -82,6 +83,7 @@ $mat-chips-chip-margin: 8px;
 .mat-chip-remove {
   margin-right: $mat-chip-remove-margin-after;
   margin-left: $mat-chip-remove-margin-before;
+  cursor: pointer;
 
   [dir='rtl'] & {
     margin-right: $mat-chip-remove-margin-before;


src/lib/chips/chips.scss
```diff
@@ -82,6 +85,7 @@ $mat-chips-chip-margin: 8px;
 .mat-chip-remove {
   margin-right: $mat-chip-remove-margin-after;
   margin-left: $mat-chip-remove-margin-before;
+  cursor: pointer;
```
willshowell: I had previously removed it because of this animation https://material.io/guidelines/components/chips.html#chips-behavior, but I certainly understand how pointer helps show purpose

src/lib/chips/chips.scss
```diff
@@ -17,10 +18,12 @@ $mat-chips-chip-margin: 8px;
 
 .mat-chip:not(.mat-basic-chip) {
   display: inline-flex;
-  padding: $mat-chip-vertical-padding $mat-chip-horizontal-padding $mat-chip-vertical-padding $mat-chip-horizontal-padding;
+  padding: $mat-chip-vertical-padding $mat-chip-horizontal-padding $mat-chip-vertical-padding
+      $mat-chip-horizontal-padding;
   border-radius: $mat-chip-horizontal-padding * 2;
   align-items: center;
   cursor: default;
+  transition: box-shadow $swift-ease-out-duration $swift-ease-out-timing-function;
```
willshowell: How about `mat-elevation-transition()`?

crisbeto: Ah nice, I had forgotten we had that.

src/lib/chips/chips.scss
crisbeto: I think that example is mostly to show the color change, rather than the cursor.

src/lib/chips/chips.scss
```diff
@@ -16,8 +16,10 @@ $mat-chips-chip-margin: 8px;
 }
 
 .mat-chip:not(.mat-basic-chip) {
+  @include mat-elevation-transition;
   display: inline-flex;
-  padding: $mat-chip-vertical-padding $mat-chip-horizontal-padding $mat-chip-vertical-padding $mat-chip-horizontal-padding;
+  padding: $mat-chip-vertical-padding $mat-chip-horizontal-padding $mat-chip-vertical-padding
```
tinayuangao: here can use `padding: $mat-chip-vertical-padding $mat-chip-horizontal-padding` instead of 4 values?

crisbeto: Good point. Changed.

tinayuangao:
LGTM

PR #6026, opened 2017-07-25
fix(datepicker): use correct viewContainerRef for dialog.

### LINKED ISSUE #5118: Datepicker Pop-up Doesn't Close when Overlay Background is Clicked

When using the touchUi feature of the datepicker, the overlay should close when the overlay background is clicked.

Clicking outside the overlay removes the background but does not close the overlay. After the background is removed, clicking any button on the overlay closes the overlay instead of performing the button's action.

Clicking outside the overlay should close the overlay. The overlay should not close unless a date is selected or someone clicks outside the overlay.

Improve usability

Angular: 4.2.0
Material: 2.0.0-beta.6

Adding the click listener to the Component's host property provides the correct behavior, even though no action is actually taken when the event occurs:

```
@Component({
   ...
   host: {
        '(document:click)': 'onClick()',
    }
})
export class MyComponent {
   ...
   private onClick(): void { }
}
```

mmalerba:
fixes #5118

### DIFF
--- a/src/demo-app/datepicker/datepicker-demo.ts
+++ b/src/demo-app/datepicker/datepicker-demo.ts
@@ -1,4 +1,4 @@
-import {Component} from '@angular/core';
+import {ChangeDetectionStrategy, Component} from '@angular/core';
 import {MdDatepickerInputEvent} from '@angular/material';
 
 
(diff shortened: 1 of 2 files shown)


jelbourn:
My suspicion is that the combination of OnPush with a specific view container is limiting where change detection runs for events on the global overlay container

PR #6040, opened 2017-07-26
chore(docs): correcting headers and adding header linking

amcdnl:
This PR fixes inconsistent headers and adds header linking for the table of contents in the material docs.

This PR is part of https://github.com/angular/material.angular.io/pull/230

### DIFF
--- a/tools/gulp/tasks/docs.ts
+++ b/tools/gulp/tasks/docs.ts
@@ -71,7 +71,7 @@ task('docs', [
 task('markdown-docs', () => {
   // Extend the renderer for custom heading anchor rendering
   markdown.marked.Renderer.prototype.heading = (text: string, level: number): string => {
-    if(level === 3 || level === 4) {
+    if (level === 3 || level === 4) {
       const escapedText = text.toLowerCase().replace(/[^\w]+/g, '-');
       return `
         <h${level} id="${escapedText}" class="docs-header-link">
--- a/tools/gulp/tasks/docs.ts
+++ b/tools/gulp/tasks/docs.ts
@@ -69,10 +69,25 @@ task('docs', [
 
 /** Generates html files from the markdown overviews and guides. */
 task('markdown-docs', () => {
+  // Extend the renderer for custom heading anchor rendering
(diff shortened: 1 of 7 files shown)


tools/gulp/tasks/docs.ts
```diff
@@ -69,10 +69,25 @@ task('docs', [
 
 /** Generates html files from the markdown overviews and guides. */
 task('markdown-docs', () => {
+  // Extend the renderer for custom heading anchor rendering
+  markdown.marked.Renderer.prototype.heading = (text: string, level: number): string => {
```
jelbourn: The docs for marked give a different API for customizing a renderer, don't think you're supposed to monkey-patch the renderer prototype:
https://github.com/chjj/marked#overriding-renderer-methods

tools/gulp/tasks/docs.ts
```diff
@@ -69,10 +69,25 @@ task('docs', [
...
+    if(level === 3 || level === 4) {
+      const escapedText = text.toLowerCase().replace(/[^\w]+/g, '-');
+      return `
+        <h${level} id="${escapedText}">
```
jelbourn: Is it easier to add the ids during generation than when the html file is loaded on the docs site?

tools/gulp/tasks/docs.ts
amcdnl: Because we are using the gulp plugin, I wasn't able to override the renderer like that. In order to do that, I would need to roll our own gulp adapter so we can pass in a custom renderer. Is that something you would like me to look at?

tools/gulp/tasks/docs.ts
amcdnl: I figured it would be faster to do it during generation. Happy to change it if you don't fancy this approach.

amcdnl:
@jelbourn - please advise on next steps :)

tools/gulp/tasks/docs.ts
jelbourn: Nope, didn't know the gulp config wasn't a full pass-through

PR #6044, opened 2017-07-26
chore: fix linting errors

crisbeto:
Fixes a couple of linting errors that were introduced in master.

PR #6055, opened 2017-07-26
build: properly lint tooling source code

DevVersion:
* Fixes a glob that is responsible for IDE's TypeScript service and also for TSLint
* Fixes strict null check errors in the screenshot app. Currently the app can't be compiled properly since strict null checks had been introduced there as well.

**Note**: Two remaining lint issues coming from https://github.com/angular/material2/pull/6044

### DIFF
--- a/tools/screenshot-test/src/app/firebase.service.ts
+++ b/tools/screenshot-test/src/app/firebase.service.ts
@@ -71,10 +71,7 @@ export class FirebaseService {
             this._readResults(childSnapshot);
             break;
         }
-        counter++;
-        if (counter === snapshot.numChildren()) {
-          return true;
-        }
+        return ++counter === snapshot.numChildren();
       });
     });
   }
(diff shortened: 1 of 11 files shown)


tools/screenshot-test/src/app/firebase.service.ts
```diff
@@ -71,10 +71,7 @@ export class FirebaseService {
             this._readResults(childSnapshot);
             break;
         }
-        counter++;
-        if (counter === snapshot.numChildren()) {
-          return true;
-        }
+        return ++counter === snapshot.numChildren();
```
jelbourn: I prefer not to use increment/decrement expressions as part of a larger expression where the order of the increment/decrement would matter
(here and elsewhere)

DevVersion: Not sure. I thought it's pretty basic a pretty JavaScript syntax. I'm fine changing it if you prefer it though.

DevVersion: Changed it. Please have another look.

jelbourn: It's basic syntax in Java, C++, C#... and in all of those languages, I see this as potentially confusing.

PR #6064, opened 2017-07-26
fix(sidenav): animate sidenav on open/close changes

andrewseguin:
Fixes issue where sidenavs that are initialized as open are showing the transition due to the new change detection onpush changes

### DIFF
--- a/src/lib/sidenav/sidenav.ts
+++ b/src/lib/sidenav/sidenav.ts
@@ -379,13 +379,6 @@ export class MdSidenavContainer implements AfterContentInit {
       this._watchSidenavAlign(sidenav);
     });
     this._validateDrawers();
-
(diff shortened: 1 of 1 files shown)


andrewseguin:
Would prefer to check in #4959 over this if it can fix this issue.

crisbeto:
Do we have a repro case? I can give it a try.

crisbeto:
Tried it out against a few examples from the demo app @andrewseguin and it doesn't look like #4959 would resolve the issue. I'll see if I can get a workaround there as well.

crisbeto:
Alright, I've updated #4959 to fix the initial animation as well. Here's the commit: https://github.com/angular/material2/pull/4959/commits/aad41ca09969040d4406417476e4fca38cdea813

PR #6073, opened 2017-07-26
feat(chips): Add user defined tab index to chip list

### LINKED ISSUE #6069: Material chip list should support being not focusable
Feature request

I should be able to pass in an @Input() to mat-chip-list so that it doesn't have tabIndex = 0 by default. Sometimes we want to display chips within a tooltip. And we don't want user to be able to focus on individual chips.

mat-chip-list is always focusable by default

Create mat-chip-list with non-zero chips inside

To increase use cases for material chip list component

All

tinayuangao:
Closes #6069

### DIFF
--- a/src/lib/chips/chip-list.ts
+++ b/src/lib/chips/chip-list.ts
@@ -76,8 +76,10 @@ export class MdChipList implements AfterContentInit, OnDestroy {
   /** Tab index for the chip list. */
   _tabIndex = 0;
 
-  /** User defined tab index.
-   * When it is not null, use user defined tab index. Otherwise use _tabIndex */
+  /**
+   * User defined tab index.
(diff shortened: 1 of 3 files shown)


src/lib/chips/chip-list.ts
```diff
@@ -76,6 +76,10 @@ export class MdChipList implements AfterContentInit, OnDestroy {
   /** Tab index for the chip list. */
   _tabIndex = 0;
 
+  /** User defined tab index.
+   * When it is not null, use user defined tab index. Otherwise use _tabIndex */
```
jelbourn: Formatting on the JsDoc is slightly off; should either be one line or formatted like
```ts
  /**
   * First line of the comment and
   * the second line of the comment
   */
```

jelbourn:
Add merge-ready when you're ready

PR #6086, opened 2017-07-27
build: initial material moment adapter package

DevVersion:
* Initialization of the `material-moment-adapter` package in the Material repository.

### DIFF
--- a/src/material-moment-adapter/index.spec.ts
+++ b/src/material-moment-adapter/index.spec.ts
@@ -1,0 +1,9 @@
+import {_momentAdapter} from './public_api';
+
+describe('moment-adapter', () => {
+
+  it('should be tested when running the tests', () => {
+    expect(_momentAdapter).toBe(0);
+  });
+
+});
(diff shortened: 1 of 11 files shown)


mmalerba:
@jelbourn Talked with Igor and agreed that we should create a package `cdk-moment-adapter` with moment as a peer-dep.

One little thing I though of though: there is technically one small piece of this that belongs to material rather than the CDK, the list of formats to be used in the material components (the moment equivalent of the native date formats here: https://github.com/angular/material2/blob/master/src/lib/core/datetime/native-date-formats.ts)

So I guess the question is do we put that here or in material? And if we put it here is it still ok to call it `cdk-moment-adapter` rather than `material-moment-adapter`

mmalerba:
since we're keeping all of the datetime stuff in material for now, I think material-moment-adapter sounds good :+1:

DevVersion:
@mmalerba Rebased this PR. Please have another look.

PR #6096, opened 2017-07-27
demo(datepicker): create demos for datepicker a11y use cases

mmalerba:


src/demo-app/a11y/datepicker/datepicker-a11y.html
```diff
@@ -0,0 +1,112 @@
+<section>
+  <h2>Choose a date (e.g. choose your date of birth)</h2>
+  <md-input-container>
+    <input mdInput
```
jelbourn: This input needs either a `<label>` or an `aria-label`
(same for other inputs that don't have one)

src/demo-app/a11y/datepicker/datepicker-a11y.html
```diff
@@ -0,0 +1,112 @@
...
+           [mdDatepicker]="birthdayPicker"
+           [(ngModel)]="birthday"
+           #birthdayModel="ngModel"
+           [max]="maxBirthday"
+           required
+           placeholder="Date of birth">
+    <button mdSuffix [mdDatepickerToggle]="birthdayPicker"></button>
```
jelbourn: Add a comment here like
```html
<!-- Button aria-label is added automatically by mdDatepickerToggle -->
```

src/demo-app/a11y/datepicker/datepicker-a11y.html
mmalerba: label is automatically created from the placeholder

mmalerba:
PTAL

PR #6102, opened 2017-07-28
fix(icon): error when toggling icon with binding in IE11

### LINKED ISSUE #6093: svgIcon with bound expression in contents fails for IE 11
```html
<md-icon [svgIcon]="myIcon">{{myIcon}}</md-icon>
```
This fails in IE 11 if the `md-icon` component is hidden and shown.  Removing `{{myIcon}}` from the template avoids the error.

No error is shown in the console.  The icon is not duplicated.

An error appears in the console and the icon is duplicated.

<details>
<summary>Error from the console</summary>

```
ERROR Error: Invalid argument.
   "ERROR"
   {
      [functions]: ,
      __proto__: { },
      description: "Invalid argument.",
      message: "Invalid argument.",
      name: "Error",
      ngDebugContext: { },
      number: -2147024809,
      stack: "Error: Invalid argument.
   at DefaultDomRenderer2.prototype.setValue (eval code:2940:71)
   at BaseAnimationRenderer.prototype.setValue (eval code:513:73)
   at DebugRenderer2.prototype.setValue (eval code:13793:66)
   at checkAndUpdateTextInline (eval code:11867:9)
   at checkAndUpdateNodeInline (eval code:12358:13)
   at checkAndUpdateNode (eval code:12301:9)
   at debugCheckAndUpdateNode (eval code:13162:5)
   at debugCheckRenderNodeFn (eval code:13141:13)
   at Anonymous function (Function code:17:5)
   at debugUpdateRenderer (eval code:13126:5)
   at checkAndUpdateView (eval code:12273:5)
   at callViewAction (eval code:12633:21)
   at execEmbeddedViewsAction (eval code:12591:17)
   at checkAndUpdateView (eval code:12269:5)
   at callViewAction (eval code:12633:21)",
      Symbol(rxSubscriber)_m.ix4mqv1tfg5: undefined
   }
```
</details>

http://plnkr.co/edit/IxGqqudBSwWHompYGZtD?p=preview
1. Browse to plunker using IE 11.
2. Open console in developer tools
3. Toggle slider off
4. Toggle slider on

I wanted to fall back to having the name of the icon appear on the screen if the svg doesn't load over the network.  This works great in other browsers, but fails for IE 11.

Angular 4.3.1, Material 2.0.0-beta.8, IE 11

The error happens in [DefaultDomRenderer2](https://github.com/angular/angular/blob/master/packages/platform-browser/src/dom/dom_renderer.ts#L206):
```ts
node.nodeValue = value
```

Because you're setting innerHtml [here](https://github.com/angular/material2/blob/e488e3fa24b3f8f1f22a6493df634842554650f7/src/lib/icon/icon.ts#L170) as a workaround for [this issue](https://github.com/angular/angular/issues/6327)

...which runs straight into the situation discussed [here](https://github.com/angular/angular.js/issues/8659#issuecomment-156267569).  Namely, setting `innerHtml` on a parent node and then trying to update the `nodeValue` of a removed node fails silently in modern browsers, but IE 11 throws an error.

crisbeto:
* Fixes an error that was being logged by IE11 for icons that are toggled via `ngIf` and have a binding expression.
* Some minor readability improvements in the icon component.

Fixes #6093.

### DIFF
--- a/src/lib/icon/icon.ts
+++ b/src/lib/icon/icon.ts
@@ -33,12 +33,6 @@ export const _MdIconMixinBase = mixinColor(MdIconBase);
 
 /**
  * Component to display an icon. It can be used in the following ways:
- * - Specify the svgSrc input to load an SVG icon from a URL. The SVG content is directly inlined
- *   as a child of the <md-icon> component, so that CSS styles can easily be applied to it.
- *   The URL is loaded via an XMLHttpRequest, so it must be on the same domain as the page or its
- *   server must be configured to allow cross-domain requests.
- *   Example:
- *     <md-icon svgSrc="assets/arrow.svg"></md-icon>
  *
  * - Specify the svgIcon input to load an SVG icon from a URL previously registered with the
  *   addSvgIcon, addSvgIconInNamespace, addSvgIconSet, or addSvgIconSetInNamespace methods of
@@ -136,7 +130,7 @@ export class MdIcon extends _MdIconMixinBase implements OnChanges, OnInit, CanCo
 
   ngOnChanges(changes: SimpleChanges) {
     // Only update the inline SVG icon if the inputs changed, to avoid unnecessary DOM operations.
-    if ((changes.svgIcon || changes.svgSrc) && this.svgIcon) {
+    if (changes.svgIcon && this.svgIcon) {
       const [namespace, iconName] = this._splitIconName(this.svgIcon);
 
       first.call(this._mdIconRegistry.getNamedSvgIcon(iconName, namespace)).subscribe(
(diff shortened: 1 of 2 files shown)


src/lib/icon/icon.ts
```diff
@@ -164,21 +163,26 @@ export class MdIcon extends _MdIconMixinBase implements OnChanges, OnInit, CanCo
...
   private _setSvgElement(svg: SVGElement) {
     const layoutElement = this._elementRef.nativeElement;
-    // Remove existing child nodes and add the new SVG element.
-    // We would use renderer.detachView(Array.from(layoutElement.childNodes)) here,
-    // but it fails in IE11: https://github.com/angular/angular/issues/6327
-    layoutElement.innerHTML = '';
+
+    // Remove existing child nodes and add the new SVG element. Note that we can't
+    // use innerHTML, because IE will throw if the element has a data binding.
+    for (let child of layoutElement.childNodes) {
```
jelbourn: TpyeScript will compile this to a `for` loop that checks `childNodes.length` on every iteration. I know that used to be something to avoid, but not sure if it's necessary any more.

crisbeto: Hmm, I guess it makes sense for the browser to recompute it. I'll cache it in a variable.

crisbeto:

PR #6105, opened 2017-07-28
docs: individual cdk entry points for systemjs

DevVersion:
* Updates the SystemJS guide section to show that the individual CDK entry points need to be mapped manually.

### DIFF
--- a/guides/getting-started.md
+++ b/guides/getting-started.md
@@ -136,15 +136,25 @@ Note that `md-icon` supports any font or svg icons; using Material Icons is one 
 ## Appendix: Configuring SystemJS
 
 If your project is using SystemJS for module loading, you will need to add `@angular/material` and
-`@angular/cdk` to the SystemJS configuration:
+`@angular/cdk` to the SystemJS configuration.
+
+For the `@angular/cdk`, the different utilities are split into separate entry points. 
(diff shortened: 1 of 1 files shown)


guides/getting-started.md
```diff
@@ -136,15 +136,25 @@ Note that `md-icon` supports any font or svg icons; using Material Icons is one
 ## Appendix: Configuring SystemJS
 
 If your project is using SystemJS for module loading, you will need to add `@angular/material` and
-`@angular/cdk` to the SystemJS configuration:
+`@angular/cdk` to the SystemJS configuration.
+
+For the `@angular/cdk`, the different utilities are split into separate entry points.
```
jelbourn: I'd phrase this as 
```markdown
The `@angular/cdk` package is organized of multiple entry-points. Each of these entry-points are registered individually with SystemJS.
```

DevVersion: The first sentence sounds better. About the second one, I'd rather say that they manually "need" to be registered. Because by default they are not registered within SystemJS.

jelbourn: Can also say "must be" there. I chose this phrasing, though, to make it seem like less "this is something extra you must do" and more "this is the way it is".

DevVersion: Yeah I like that `must be`. Done.

PR #6107, opened 2017-07-28
build: no longer add pure annotations to es5 bundles 

DevVersion:
* The pure annotations are no longer added to ES5 bundles of Angular Material or the CDK. The pure annotations should be added by the new build optimizer which is part of the CLI.


jelbourn:
@kara mentioned we might not want to actually being doing this.

DevVersion:
@jelbourn Ah I guess because of the new optimizer stuff. In this case we either need to remove it or just add it to all bundles.

DevVersion:
@jelbourn Chatted with @kara about this. We agreed that it would make sense to no longer add the pure annotations to our ES5 bundles.

PR #6110, opened 2017-07-28
fix(select): set aria-multiselectable for multi-select

crisbeto:
Sets the `aria-multiselectable` attribute to multi-select instances, as per https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1904,6 +1904,21 @@ describe('MdSelect', () => {
             .toBe('true', 'Expected aria-hidden to be true when the select is open.');
       });
 
+      it('should set `aria-multiselectable` to true on multi-select instances', () => {
+        fixture.destroy();
+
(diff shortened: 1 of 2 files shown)


src/lib/select/select.spec.ts
```diff
@@ -1904,6 +1904,17 @@ describe('MdSelect', () => {
...
 
+      it('should set `aria-multiselectable` on multi-select instances', () => {
+        fixture.destroy();
+
+        const multiFixture = TestBed.createComponent(MultiSelect);
+
+        multiFixture.detectChanges();
+        select = multiFixture.debugElement.query(By.css('md-select')).nativeElement;
+
+        expect(select.getAttribute('aria-multiselectable')).toBe('true');
```
kara: Can you add a test to check that this attribute is not added to normal selects?

crisbeto: Not added at all or set to `false`? AFAIK it is allowed to be false.

PR #6116, opened 2017-07-28
feat(stepper): Add support for linear stepper

g1shin:
This is the PR for adding support for linear stepper using an array of `FormGroup`s.
Custom `errorStateMatcher` has been added to be able to show errors for inputs that are not submitted.

(cell.ts and row.ts files were changed because there were lint errors regarding long lines.)

### DIFF
--- a/src/cdk/stepper/stepper-button.ts
+++ b/src/cdk/stepper/stepper-button.ts
@@ -12,7 +12,10 @@ import {CdkStepper} from './stepper';
 /** Button that moves to the next step in a stepper workflow. */
 @Directive({
   selector: 'button[cdkStepperNext]',
-  host: {'(click)': '_stepper.next()'}
+  host: {
+    '(click)': '_stepper.next()',
+    'type': 'button'
+  }
 })
 export class CdkStepperNext {
   constructor(public _stepper: CdkStepper) { }
@@ -21,7 +24,10 @@ export class CdkStepperNext {
 /** Button that moves to the previous step in a stepper workflow. */
 @Directive({
   selector: 'button[cdkStepperPrevious]',
-  host: {'(click)': '_stepper.previous()'}
+  host: {
+    '(click)': '_stepper.previous()',
+    'type': 'button'
+  }
 })
 export class CdkStepperPrevious {
   constructor(public _stepper: CdkStepper) { }
--- a/src/cdk/stepper/stepper-button.ts
+++ b/src/cdk/stepper/stepper-button.ts
@@ -12,10 +12,7 @@ import {CdkStepper} from './stepper';
 /** Button that moves to the next step in a stepper workflow. */
 @Directive({
   selector: 'button[cdkStepperNext]',
-  host: {
-    '(click)': '_stepper.next()',
-    'type': 'button'
-  }
+  host: {'(click)': '_stepper.next()'}
 })
 export class CdkStepperNext {
   constructor(public _stepper: CdkStepper) { }
@@ -24,10 +21,7 @@ export class CdkStepperNext {
 /** Button that moves to the previous step in a stepper workflow. */
 @Directive({
   selector: 'button[cdkStepperPrevious]',
-  host: {
-    '(click)': '_stepper.previous()',
-    'type': 'button'
-  }
+  host: {'(click)': '_stepper.previous()'}
 })
 export class CdkStepperPrevious {
   constructor(public _stepper: CdkStepper) { }
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -24,6 +24,7 @@ import {
 } from '@angular/core';
 import {LEFT_ARROW, RIGHT_ARROW, ENTER, SPACE} from '@angular/cdk/keyboard';
 import {CdkStepLabel} from './step-label';
+import {coerceBooleanProperty} from '@angular/cdk/coercion';
 
 /** Used to generate unique ID for each stepper component. */
 let nextId = 0;
@@ -45,7 +46,7 @@ export class CdkStepperSelectionEvent {
 
 @Component({
   selector: 'cdk-step',
-  templateUrl: 'step.html',
+  templateUrl: 'step.html'
 })
 export class CdkStep {
   /** Template for step label if it exists. */
@@ -53,6 +54,21 @@ export class CdkStep {
 
   /** Template for step content. */
   @ViewChild(TemplateRef) content: TemplateRef<any>;
+
+  /** Whether step is disabled or not. */
+  @Input()
+  get disabled() { return this._disabled; }
+  set disabled(value: any) {
+    this._disabled = coerceBooleanProperty(value);
+  }
+  private _disabled = false;
+
+  /** Whether the user has interacted with step or not. */
+  get interacted() { return this._interacted; }
+  set interacted(value: any) {
+    this._interacted = coerceBooleanProperty(value);
+  }
+  private _interacted = false;
 
   /** Label of the step. */
   @Input()
@@ -84,7 +100,8 @@ export class CdkStepper {
   @Input()
   get selectedIndex() { return this._selectedIndex; }
   set selectedIndex(index: number) {
-    if (this._selectedIndex != index) {
+    this._steps.toArray()[this._selectedIndex].interacted = true;
+    if (this._selectedIndex != index && !this._steps.toArray()[index].disabled) {
       this._emitStepperSelectionEvent(index);
       this._focusStep(this._selectedIndex);
     }
@@ -153,7 +170,7 @@ export class CdkStepper {
         break;
       case SPACE:
       case ENTER:
-        this._emitStepperSelectionEvent(this._focusIndex);
+        this.selectedIndex = this._focusIndex;
         break;
       default:
         // Return to avoid calling preventDefault on keys that are not explicitly handled.
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -55,6 +55,7 @@ export class CdkStep {
   /** Template for step content. */
   @ViewChild(TemplateRef) content: TemplateRef<any>;
 
+  // TODO(jwshin): use disabled mixin when moved to cdk.
   /** Whether step is disabled or not. */
   @Input()
   get disabled() { return this._disabled; }
@@ -63,10 +64,10 @@ export class CdkStep {
   }
   private _disabled = false;
 
-  /** Whether the user has interacted with step or not. */
+  /** Whether user has seen the expanded step content or not . */
   get interacted() { return this._interacted; }
-  set interacted(value: any) {
-    this._interacted = coerceBooleanProperty(value);
+  set interacted(value: boolean) {
+    this._interacted = value;
   }
   private _interacted = false;
 
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -25,6 +25,7 @@ import {
 import {LEFT_ARROW, RIGHT_ARROW, ENTER, SPACE} from '@angular/cdk/keyboard';
 import {CdkStepLabel} from './step-label';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {AbstractControl} from '@angular/forms';
 
 /** Used to generate unique ID for each stepper component. */
 let nextId = 0;
@@ -55,14 +56,13 @@ export class CdkStep {
   /** Template for step content. */
   @ViewChild(TemplateRef) content: TemplateRef<any>;
 
-  // TODO(jwshin): use disabled mixin when moved to cdk.
-  /** Whether step is disabled or not. */
+  /** The top level abstract control of the step. */
   @Input()
-  get disabled() { return this._disabled; }
-  set disabled(value: any) {
-    this._disabled = coerceBooleanProperty(value);
+  get stepControl() { return this._stepControl; }
+  set stepControl(control: AbstractControl) {
+    this._stepControl = control;
   }
-  private _disabled = false;
+  private _stepControl: AbstractControl;
 
   /** Whether user has seen the expanded step content or not . */
   interacted = false;
@@ -98,7 +98,10 @@ export class CdkStepper {
   get selectedIndex() { return this._selectedIndex; }
   set selectedIndex(index: number) {
     this._steps.toArray()[this._selectedIndex].interacted = true;
-    if (this._selectedIndex != index && !this._steps.toArray()[index].disabled) {
+    for (let i = 0; i < index; i++) {
+      if (!this._steps.toArray()[i].stepControl.valid) { return; }
+    }
+    if (this._selectedIndex != index) {
       this._emitStepperSelectionEvent(index);
       this._focusStep(this._selectedIndex);
     }
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -65,11 +65,7 @@ export class CdkStep {
   private _disabled = false;
 
   /** Whether user has seen the expanded step content or not . */
-  get interacted() { return this._interacted; }
-  set interacted(value: boolean) {
-    this._interacted = value;
-  }
-  private _interacted = false;
+  interacted = false;
 
   /** Label of the step. */
   @Input()
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -83,8 +83,7 @@ export class CdkStep {
   selector: 'cdk-stepper',
   host: {
     '(focus)': '_focusStep()',
-    '(keydown)': '_onKeydown($event)',
-    '[linear]': 'linear',
+    '(keydown)': '_onKeydown($event)'
   },
 })
 export class CdkStepper {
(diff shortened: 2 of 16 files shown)


src/demo-app/stepper/stepper-demo.html
```diff
@@ -1,3 +1,52 @@
...
+            <md-error>This field is required</md-error>
+          </md-input-container>
+          <div>
+            <button md-button mdStepperPrevious>Back</button>
+            <button md-button mdStepperNext>Next</button>
+          </div>
+        </div>
+      </md-step>
+
+      <md-step [disabled]="!formGroup.controls.formArray.controls[1].valid">
```
mmalerba: I don't think this is quite right, you also need to make sure the previous step is enabled, otherwise you might end up in a situation like this:

- the 1st step is invalid
- the 2nd step is disabled, but has no controls with validation so it's considered valid
- the 3rd step would be set to enabled since it's enabledness is based solely off the previous step's validity

(It would be nice if there was some way we could make this easier for users. I don't have any ideas off the top of my head though...)

src/cdk/stepper/stepper-button.ts
```diff
@@ -12,7 +12,10 @@ import {CdkStepper} from './stepper';
 /** Button that moves to the next step in a stepper workflow. */
 @Directive({
   selector: 'button[cdkStepperNext]',
-  host: {'(click)': '_stepper.next()'}
+  host: {
+    '(click)': '_stepper.next()',
+    'type': 'button'
```
kara: Maybe we could add guidance to the docs instead?

src/demo-app/stepper/stepper-demo.html
```diff
@@ -1,3 +1,52 @@
+<h2>Linear Vertical Stepper Demo</h2>
+<form [formGroup]="formGroup" novalidate>
```
kara: Yep, not necessary

src/demo-app/stepper/stepper-demo.html
```diff
@@ -1,3 +1,52 @@
...
+            <input mdInput placeholder="Last Name" formControlName="lastNameFormCtrl" required>
+            <md-error>This field is required</md-error>
+          </md-input-container>
+          <div>
+            <button md-button mdStepperNext>Next</button>
+          </div>
+        </div>
+      </md-step>
+
+      <md-step [disabled]="!formGroup.controls.formArray.controls[0].valid">
```
kara: Generally speaking, using the `get()` method is preferred over the controls property. I'd also suggest creating a getter for the array so it's less painful:

```html
<md-step [disabled]="formArray.get([0]).valid">
```

```ts
get formArray() { return this.formGroup.get('formArray'); }
```

But see comment below about strategy

src/demo-app/stepper/stepper-demo.html
kara: What about creating a custom validator function for the group or array? It seems like you need to consider more than one form control here, which is the classic situation where that is needed. A higher-level validator would allow you to surface more information in a custom error message, like all step validity or the latest step where all steps before it are valid.

src/cdk/stepper/stepper.ts
```diff
@@ -54,6 +55,21 @@ export class CdkStep {
   /** Template for step content. */
   @ViewChild(TemplateRef) content: TemplateRef<any>;
 
+  /** Whether step is disabled or not. */
+  @Input()
+  get disabled() { return this._disabled; }
+  set disabled(value: any) {
+    this._disabled = coerceBooleanProperty(value);
+  }
+  private _disabled = false;
```
jelbourn: Can you add a `TODO` to use the disabled mixin when it's moved to the cdk?

src/demo-app/stepper/stepper-demo.ts
```diff
@@ -7,10 +8,28 @@ import {Component} from '@angular/core';
...
+  formGroup: FormGroup;
+
   steps = [
     {label: 'Confirm your name', content: 'Last name, First name.'},
     {label: 'Confirm your contact information', content: '123-456-7890'},
     {label: 'Confirm your address', content: '1600 Amphitheater Pkwy MTV'},
     {label: 'You are now done', content: 'Finished!'}
   ];
+
+  constructor(private _fb: FormBuilder) { }
```
jelbourn: `_fb`  -> `_formBuilder`

src/lib/stepper/stepper.ts
```diff
@@ -15,26 +15,52 @@ import {
...
+  constructor(mdStepper: MdStepper,
+              @Optional() @SkipSelf() @Inject(MD_ERROR_GLOBAL_OPTIONS) errorOptions: ErrorOptions) {
     super(mdStepper);
+    this._originalErrorStateMatcher =
+        errorOptions ? errorOptions.errorStateMatcher || defaultErrorStateMatcher
+            : defaultErrorStateMatcher;
+  }
+
+  /** Custom error state matcher that additionally checks for validity of interacted form. */
+  errorStateMatcher = (control: FormControl, form: FormGroupDirective | NgForm) => {
```
jelbourn: The class should `implement ErrorOptions`

jelbourn: Does it work if this is a prototype method rather than a property?

src/cdk/stepper/stepper.ts
```diff
@@ -54,6 +55,21 @@ export class CdkStep {
...
+
+  /** Whether the user has interacted with step or not. */
```
jelbourn: Can you expand this comment to describe what "interacted with" means?

src/lib/stepper/stepper.ts
```diff
@@ -15,26 +15,52 @@ import {
...
+    let originalErrorState = this._originalErrorStateMatcher(control, form);
```
jelbourn: Can you add a comment that explains the background for _why_ we are doing this custom error matcher stuff? (i.e., everything we talked about in the meeting)

src/lib/stepper/stepper.ts
mmalerba: No because the `ErrorOptions` stuff is implemented kind of strangely and the context would be lost if we did it as a prototype method. I would like to refactor it, probably not as part of this PR though

src/cdk/stepper/stepper-button.ts
g1shin: I removed `'type': 'button'`. I'll add it to the docs instead as @kara suggested.

g1shin:
Changes have been made; ready for review. :+1:

src/cdk/stepper/stepper.ts
```diff
@@ -54,6 +55,22 @@ export class CdkStep {
...
+  @Input()
+  get disabled() { return this._disabled; }
+  set disabled(value: any) {
+    this._disabled = coerceBooleanProperty(value);
+  }
+  private _disabled = false;
+
+  /** Whether user has seen the expanded step content or not . */
+  get interacted() { return this._interacted; }
+  set interacted(value: boolean) {
```
mmalerba: setter/getter isn't doing anything can switch to
```ts
interacted = false;
```

src/demo-app/stepper/stepper-demo.ts
```diff
@@ -7,10 +10,44 @@ import {Component} from '@angular/core';
...
+        })
+      ], this._stepValidator)
+    });
+  }
+
+  /**
+   * Form array validator to check if all form groups in form array are valid.
+   * If not, it will return the index of the first invalid form group.
+   */
+  private _stepValidator: ValidatorFn = (formArray: FormArray): ValidationErrors | null => {
```
mmalerba: I was imagining this more as something that we define in our code and users can just import and use.

src/demo-app/stepper/stepper-demo.html
```diff
@@ -1,3 +1,50 @@
...
+        <md-error>This field is required</md-error>
+      </md-input-container>
+      <div>
+        <button md-button mdStepperPrevious type="button">Back</button>
+        <button md-button mdStepperNext type="button">Next</button>
+      </div>
+    </md-step>
+
+    <md-step
+        [disabled]="formArray.hasError('invalid step') ?
```
mmalerba: This doesn't really seem to have made the disabled logic much easier... is that what you had in mind @kara or is there a better way to organize the custom validator?

joejordanbrown: @mmalerba wouldn't it be better to not direct;y use the previous step form validity on the disabled input and instead use a `@Input() valid` which you would pass the current step validity and internally check the validation so you would have `<md-step [valid]="formGroup.controls.formArray.controls[0].valid">`. 

It's going to be more complex than just checking if the previous is valid because it's going to need to validate against disabled, editable, optional and completed.

g1shin: @joejordanbrown The changed code no longer only checks for the previous step's validity. Now it uses the custom step validation function `stepValidator` to check for the validity of all the previous steps. If there's a step that is invalid, it will return the smallest index of the step that is invalid to compare with the index of the step that is being triggered.

@mmalerba : Waiting for @kara 's response on additional suggestions regarding this.

g1shin:
Changes have been made based on review.

src/cdk/stepper/stepper.ts
```diff
@@ -84,7 +97,8 @@ export class CdkStepper {
   @Input()
   get selectedIndex() { return this._selectedIndex; }
   set selectedIndex(index: number) {
-    if (this._selectedIndex != index) {
+    this._steps.toArray()[this._selectedIndex].interacted = true;
+    if (this._selectedIndex != index && !this._steps.toArray()[index].disabled) {
       this._emitStepperSelectionEvent(index);
```
mmalerba: I think we only want to emit a change event if the selected index changes due to user interaction, therefore this should be moved to one of the user event handlers

src/lib/stepper/stepper.ts
```diff
@@ -15,26 +15,79 @@ import {
...
+import {
+  FormArray, FormControl, FormGroupDirective, NgForm, ValidationErrors,
+  ValidatorFn
+} from '@angular/forms';
+
+/**
+ * Form array validator to check if all form groups in form array are valid.
+ * If not, it will return the index of the first invalid form group.
+ */
+export const stepValidator: ValidatorFn = (formArray: FormArray): ValidationErrors | null => {
```
mmalerba: should probably be called `stepperValidator` since it's attached to the `FormArray` representing the whole stepper, not the `FormGroup`s representing the steps.

src/lib/stepper/stepper.ts
```diff
@@ -15,26 +15,79 @@ import {
...
   @ContentChild(MdStepLabel) stepLabel: MdStepLabel;
 
-  constructor(mdStepper: MdStepper) {
+  /** Original ErrorStateMatcher that checks the validity of form control. */
+  private _originalErrorStateMatcher: ErrorStateMatcher;
+
+  constructor(mdStepper: MdStepper,
+              @Optional() @SkipSelf() @Inject(MD_ERROR_GLOBAL_OPTIONS) errorOptions: ErrorOptions) {
     super(mdStepper);
+    this._originalErrorStateMatcher =
```
mmalerba: FYI, #6147 should make this easier

src/demo-app/stepper/stepper-demo.html
joejordanbrown: @g1shin I see that, but I think it's too complex for the end user to have to add that to each step. It's going to get confusing for people when you use that with `*ngFor` and people will make mistakes. Especially when it's going to have a lot more options added in future for `[editable]` and `[optionial]`. 

You should be using less than operator `<` instead of less than or equal operator `<=` for your disabled logic. The second step should be `[disabled]="formArray.hasError('invalid step') ?
             formArray.getError('invalid step').index < 1 : false"` you are also missing the required validation on your second step to be able to test this.

```
 phoneFormCtrl: ['', Validators.required],
```

Do you have the design document for the stepper? I don't see it on the [wiki](https://github.com/angular/material2/wiki/Design-doc-directory).

mmalerba:
This looks like the right idea to me. I'll let @kara comment on whether this covers all of the cases she mentioned.

I think we should also add a `linear` attribute to the stepper that toggles whether or not we force the user to make previous steps valid before allowing them to move on (like we had in the original design doc for stepper).

src/lib/stepper/stepper.ts
```diff
@@ -15,26 +15,66 @@ import {
...
+} from '../core/error/error-options';
+import {
+  FormArray, FormControl, FormGroupDirective, NgForm, ValidationErrors,
+  ValidatorFn
+} from '@angular/forms';
+
+/**
+ * Form array validator to check if all form groups in form array are valid.
+ * If not, it will return the index of the first invalid form group.
+ */
```
kara: Is this comment left over from a previous implementation? Doesn't seem to go with `MdStep`.

src/cdk/stepper/stepper.ts
```diff
@@ -84,6 +97,10 @@ export class CdkStepper {
   @Input()
   get selectedIndex() { return this._selectedIndex; }
   set selectedIndex(index: number) {
+    this._steps.toArray()[this._selectedIndex].interacted = true;
+    for (let i = 0; i < index; i++) {
+      if (!this._steps.toArray()[i].stepControl.valid) { return; }
```
kara: Nit: Here you're calling `toArray()` every index of the loop. Perhaps save it at the top instead? 

```ts
const steps = this._steps.toArray();
steps[this._selectedIndex].interacted = true;
for (let i =0; i < index; i++) {
  if (!steps[i].stepControl.valid) { return; }
}
```

I'd also consider maybe breaking this loop out into its own mini-method, like `_anyControlsInvalid`.

src/lib/stepper/stepper.ts
g1shin: I saw the PR, but should this change be made after that PR is merged into master and 'stepper' branch has synced to upstream master?

src/cdk/stepper/stepper.ts
g1shin: I thought this is commonly called by all user interactions and that's why we decided to delegate emitting a change event here. Is this not what you meant?

g1shin:
Form control approach implemented and `linear` attribute added to stepper; ready for review again :+1:

src/cdk/stepper/stepper.ts
```diff
@@ -71,6 +84,7 @@ export class CdkStep {
   host: {
     '(focus)': '_focusStep()',
     '(keydown)': '_onKeydown($event)',
+    '[linear]': 'linear',
```
mmalerba: don't need this

src/cdk/stepper/stepper.ts
```diff
@@ -80,11 +94,19 @@ export class CdkStepper {
   /** The list of step headers of the steps in the stepper. */
   _stepHeader: QueryList<ElementRef>;
 
+  /** Whether the validity of previous steps should be checked or not. */
+  @Input()
+  get linear() { return this._linear; }
+  set linear(value: any) {
+    this._linear = coerceBooleanProperty(value);
```
mmalerba: nit: can 1-line the setter as well

src/demo-app/stepper/stepper-demo.html
```diff
@@ -1,3 +1,46 @@
+<h2>Linear Vertical Stepper Demo</h2>
+<form [formGroup]="formGroup">
+  <md-vertical-stepper formArrayName="formArray" linear>
```
mmalerba: can we do `[linear]="isLinear"` and add a checkbox to toggle it, that way its easy to test

src/lib/stepper/stepper.ts
mmalerba: yep, just an fyi

g1shin:
Changes made based on review :smile:

src/cdk/stepper/tsconfig-build.json
```diff
@@ -4,7 +4,8 @@
     "outDir": "../../../dist/packages/cdk",
     "baseUrl": ".",
     "paths": {
-      "@angular/cdk/keyboard": ["../../../dist/packages/cdk/keyboard/public_api"]
+      "@angular/cdk/keyboard": ["../../../dist/packages/cdk/keyboard/public_api"],
+      "@angular/cdk/coercion": ["../../../dist/packages/cdk/coercion/public_api"]
```
jelbourn: You should be able to delete the entire `paths` property in this tsconfig now

src/lib/stepper/stepper.ts
```diff
@@ -15,26 +15,58 @@ import {
...
+
+  /** Custom error state matcher that additionally checks for validity of interacted form. */
+  errorStateMatcher = (control: FormControl, form: FormGroupDirective | NgForm) => {
+    let originalErrorState = this._originalErrorStateMatcher(control, form);
+
+    /**
+     * Custom error state checks for the validity of form that is not submitted or touched
+     * since user can trigger a form change by calling for another step without directly
+     * interacting with the current form.
+     */
```
jelbourn: Use inline-style (`//`) comments for local variables

src/cdk/stepper/tsconfig-build.json
mmalerba: possibly... depends where her stepper branch is synced to

g1shin: @jelbourn @mmalerba The stepper branch is not synced up-to-date with the master branch yet. I was going to finish merging the two form control PRs that are already out, then do the syncing all together.

PR #6126, opened 2017-07-28
chore(dialog): switch to OnPush change detection

crisbeto:
Switches the dialog container to `OnPush` change detection.

Relates to #5035.

**Note:** This is the same changeset from #5422 so we can continue investigating the sync failures.


jelbourn:
Setting to merge-ready so it shows up on caretaker's radar

crisbeto:
@jelbourn for some reason rebasing this one ended up clearing all the changes, probably because of the revert commit on the previous PR. I'll re-re-open it with a brand new commit.

PR #6128, opened 2017-07-28
fix(input): prevent input caret from sticking on iOS

### LINKED ISSUE #5838: Input error when delete with backspace on touch device
Input error when delete with backspace on touch device. The cursor is not placed at the beginning of the placeholder but  is in a forward position.
in touch device.

have the cursor at the beginning of the placeholder

The cursor is in a forward position

go to email example in this page:
https://material.angular.io/components/input/overview

try to write a word like:  bohumn

delete with backspace

the cursor is not at the beginning but in a forward position

is not correct

mmalerba:
fixes #5838

### DIFF
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -7,31 +7,31 @@
  */
 
 import {
-  AfterContentInit,
   AfterContentChecked,
+  AfterContentInit,
   AfterViewInit,
+  ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
   ContentChild,
   ContentChildren,
   Directive,
+  DoCheck,
   ElementRef,
+  Inject,
   Input,
+  OnChanges,
+  OnDestroy,
   Optional,
   QueryList,
   Renderer2,
   Self,
   ViewChild,
   ViewEncapsulation,
-  Inject,
-  ChangeDetectionStrategy,
-  OnChanges,
-  OnDestroy,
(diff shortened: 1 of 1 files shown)


src/lib/input/input-container.ts
```diff
@@ -242,6 +240,22 @@ export class MdInputDirective implements OnChanges, OnDestroy, DoCheck {
     this.id = this.id;
     this._errorOptions = errorOptions ? errorOptions : {};
     this.errorStateMatcher = this._errorOptions.errorStateMatcher || defaultErrorStateMatcher;
+
+    // On some versions of iOS the caret gets stuck in the wrong place when holding down the delete
+    // key. In order to get around this we need to "jiggle" the caret loose. Since this bug only
+    // exists on iOS, we only bother to install the listener on iOS.
+    if (_platform.IOS) {
```
jelbourn: What does mdInput do that causes this behavior in the first place? (vs. any other native input on ios)

mmalerba: I think it's the fact that it causes relayout in response to input events sometimes (e.g. when showing errors) I created a minimal repro using native inputs which I posted on the associated bug.

jelbourn: Ack. Can file figure out if this is Safari specifically or Webkit in general? I suspect it's _probably_ webkit, for which we can [file a bug](https://webkit.org/reporting-bugs/).

PR #6137, opened 2017-07-29
fix(checkbox): remove checkmark from tab order

### LINKED ISSUE #6125: Disable IE11 focus for MdCheckbox background SVG
Bug

When pressing the tab in IE11, it should not focus on MdCheckbox's SVG background. This requires an additional tab press when using checkboxes.

IE11 sets SVGs to focusable by default unlike Chrome and Firefox, so the tab focuses on the SVG background inside the checkbox.

- Open the Checkbox example page on IE11: [https://material.angular.io/components/checkbox/examples](https://material.angular.io/components/checkbox/examples).

- Tab through the checkboxes on the page.

Angular 4, Material beta.8, Internet Explorer 11

This is similar to this issue in the old version of the library. [https://github.com/angular/material/issues/7250](https://github.com/angular/material/issues/7250)

crisbeto:
Prevents users from being able to tab into the underlying SVG checkmark on IE.

Fixes #6125.

### DIFF
--- a/src/lib/checkbox/checkbox.html
+++ b/src/lib/checkbox/checkbox.html
@@ -22,6 +22,7 @@
     <div class="mat-checkbox-frame"></div>
     <div class="mat-checkbox-background">
       <svg version="1.1"
+           focusable="false"
            class="mat-checkbox-checkmark"
            xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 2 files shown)


tinayuangao:
LGTM

PR #6143, opened 2017-07-29
build: make packaging more flexible

DevVersion:
* Makes the packaging more flexible by no longer including checks that are specific to the Angular Material project (e.g `material` => `lib` mapping)
* Updates the code to be more readable by making the packaging more object orientated (TS compilation helpers shouldn't be part of the gulp task generation)
* Avoids repeating the path joining ~100x times by just passing the build package instance around.

### DIFF
--- a/tools/gulp/packages.ts
+++ b/tools/gulp/packages.ts
@@ -1,0 +1,9 @@
+import {BuildPackage, buildConfig} from 'material2-build-tools';
+import {join} from 'path';
+
+export const cdkPackage = new BuildPackage('cdk');
+export const materialPackage = new BuildPackage('material', [cdkPackage]);
+export const examplesPackage = new BuildPackage('material-examples', [materialPackage, cdkPackage]);
+
+// To avoid refactoring of the project the material package will map to the source path `lib/`.
+materialPackage.packageRoot = join(buildConfig.packagesDir, 'lib');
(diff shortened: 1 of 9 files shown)


tools/gulp/packages.ts
```diff
@@ -0,0 +1,9 @@
+import {BuildPackage, buildConfig} from 'material2-build-tools';
+import {join} from 'path';
+
+export const cdkPackage = new BuildPackage('cdk');
+export const materialPackage = new BuildPackage('material', [cdkPackage]);
+export const examplesPackage = new BuildPackage('material-examples', [materialPackage, cdkPackage]);
+
+// To avoid refactoring of the project the material package will map to the source path `lib/`.
+materialPackage.packageRoot = join(buildConfig.packagesDir, 'lib');
```
jelbourn: What do you you think of making the second constructor arg a config / options object so that it looks like
```ts
export const materialPackage = new BuildPackage('material', {
  deps: [cdkPackage],
  root: join(buildConfig.packagesDir, 'lib'),
});
```
?

DevVersion: I wanted to avoid something like that. It's really a special case that we want to overwrite the `packageRoot` here and it should not really be done through the build package.

Normally the build config is responsible for changing such things, but since this is really a special case I think this would be fine (it's just a single line; which is pretty clear IMO)

jelbourn: I don't feel strongly either way

PR #6145, opened 2017-07-29
a11y(toolbar): Add toolbar accessibility demo page

amcdnl:
Created accessibility demo page demonstrating:

- Basic Display
- Multiple Lines
- Icons
- Colors


src/demo-app/a11y/toolbar/toolbar-a11y.html
```diff
@@ -0,0 +1,28 @@
...
+    <md-toolbar>Line 1</md-toolbar>
+    <md-toolbar>Line 2</md-toolbar>
+  </section>
+
+  <section>
+    <h2>Toolbar with Icons</h2>
+    <md-toolbar>
+      <span>My App</span>
+      <span class="example-spacer"></span>
+      <md-icon class="example-icon">favorite</md-icon>
```
jelbourn: Icons actually are `aria-hidden` by default. They should only be made `aria-hidden="false"` if they're communicating some information. See the accessibility section of [the icon docs](https://material.angular.io/components/icon/overview). 

In this case, having an icon sitting by itself in the toolbar isn't very realistic to a real app; it would either be an indicator for something or an action.

src/demo-app/a11y/toolbar/toolbar-a11y.html
```diff
@@ -0,0 +1,26 @@
+<div class="demo-toolbar">
+  <section>
+    <h2>Basic Toolbar with Text  (e.g. Only display app’s name)</h2>
+    <md-toolbar>My App</md-toolbar>
```
jelbourn: A toolbar by itself isn't really representative of how the component will be used. It typically acts as a header for some content. In this capacity, it should either use `role="heading"` and `aria-level` OR it should _contain_ a header element (`h1` -> `h6`). 

The examples should represent a real use-case as much as possible.

src/demo-app/a11y/a11y.ts
```diff
@@ -26,6 +26,7 @@ export class AccessibilityDemo {
     {name: 'Grid list', route: 'grid-list'},
     {name: 'Input', route: 'input'},
     {name: 'Radio buttons', route: 'radio'},
-    {name: 'Slider', route: 'slider'},
+    {name: 'Toolbar', route: 'toolbar'},
```
tinayuangao: nit: put 'Toolbar' after 'Slider'

src/demo-app/a11y/toolbar/toolbar-a11y.html
```diff
@@ -0,0 +1,26 @@
...
+    <md-toolbar>Help</md-toolbar>
+    <md-toolbar>Settings</md-toolbar>
+  </section>
+
+  <section>
+    <h2>Toolbar with favorite icon</h2>
+    <md-toolbar>
+      <span>My App</span>
+      <span class="example-spacer"></span>
+      <md-icon aria-label="favorite">favorite</md-icon>
```
tinayuangao: Is `aria-label` working on `md-icon`? Consider adding a span with cdk-visually-hidden or make it a button/link? 
https://material.angular.io/components/icon/overview

src/demo-app/a11y/a11y.ts
amcdnl: done :)

kara:
@amcdnl Can you rebase this?

PR #6146, opened 2017-07-29
fix(tooltip): exception if mdTooltip is not a string

rafaelss95:
Fixes an error that is thrown if something rather than a string is passed via [mdTooltip] `@Input`, i.e:

```
@Component({
  selector: 'some-selector',
  template: '<div [mdTooltip]="myNumber">Content</div>'
})
export class SomeClass {
  
  myNumber = 3;
}
```

Results in the following error:

> this._message.trim is not a function

[PLUNKER](https://plnkr.co/edit/MlMcOXMzUEItNTIQzDur?p=info)


crisbeto:
Can you add a unit test for this?

rafaelss95:
@crisbeto  I can't do this. It gives me error (the method signature asks for a string): 

> Type '{}' is not assignable to type 'string'.

crisbeto:
You can change the testing component signature to something like `string | null`.

rafaelss95:
I'm not sure I follow what you mean. `tooltipDirective` is an instance of `MdTooltip` and there the signature of `set message(..)`  accepts (only) a string:

```
 set message(value: string) {
   this._message = value;
   this._setTooltipMessage(this._message);
 }
```

crisbeto:
I was talking about one of the test components here: https://github.com/angular/material2/blob/master/src/lib/tooltip/tooltip.spec.ts#L523

rafaelss95:
Changing `message` to `any` still fail.

rafaelss95:
Closing due to #7018 and #7024. :)

PR #6117, opened 2017-07-28
feat(stepper): Add support for linear stepper #2 - each step as its own form.

g1shin:
This is the PR shows the approach of considering each step as its own `form` to add support for linear stepper.
Alternative: #6116 

(cell.ts and row.ts files were changed because there were lint errors regarding long lines.)

PR #6147, opened 2017-07-30
feat(select): add support for custom errorStateMatcher

crisbeto:
* Allows for the `md-select` error behavior to be configured through an `@Input`, as well as globally through the same provider as `md-input-container`.
* Simplifies the signature of some of the error option symbols.

### DIFF
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -12,7 +12,7 @@ import {FormGroupDirective, NgForm, NgControl} from '@angular/forms';
 /** Error state matcher that matches when a control is invalid and dirty. */
 @Injectable()
 export class ShowOnDirtyErrorStateMatcher implements ErrorStateMatcher {
-  isErrorSate(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
+  isErrorState(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
     return control ? !!(control.invalid && (control.dirty || (form && form.submitted))) : false;
   }
 }
@@ -20,7 +20,7 @@ export class ShowOnDirtyErrorStateMatcher implements ErrorStateMatcher {
 /** Provider that defines how form controls behave with regards to displaying error messages. */
 @Injectable()
 export class ErrorStateMatcher {
-  isErrorSate(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
+  isErrorState(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
     return control ? !!(control.invalid && (control.touched || (form && form.submitted))) : false;
   }
 }
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -6,18 +6,11 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {InjectionToken} from '@angular/core';
+import {Injectable} from '@angular/core';
 import {FormGroupDirective, NgForm, NgControl} from '@angular/forms';
-
-/** Injection token that can be used to specify the global error options. */
-export const MD_ERROR_GLOBAL_OPTIONS = new InjectionToken<ErrorOptions>('md-error-global-options');
 
 export type ErrorStateMatcher =
   (control: NgControl | null, form: FormGroupDirective | NgForm | null) => boolean;
-
-export interface ErrorOptions {
-  errorStateMatcher?: ErrorStateMatcher;
-}
 
 /** Returns whether control is invalid and is either touched or is a part of a submitted form. */
 export const defaultErrorStateMatcher: ErrorStateMatcher = (control, form) => {
@@ -28,3 +21,12 @@ export const defaultErrorStateMatcher: ErrorStateMatcher = (control, form) => {
 export const showOnDirtyErrorStateMatcher: ErrorStateMatcher = (control, form) => {
   return control ? !!(control.invalid && (control.dirty || (form && form.submitted))) : false;
 };
+
+/**
+ * Provider that defines how form controls behave with
+ * regards to displaying error messages.
+ */
+@Injectable()
+export class ErrorOptions {
+  errorStateMatcher: ErrorStateMatcher = defaultErrorStateMatcher;
+}
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -12,7 +12,7 @@ import {FormGroupDirective, NgForm, NgControl} from '@angular/forms';
 /** Error state matcher that matches when a control is invalid and dirty. */
 @Injectable()
 export class ShowOnDirtyErrorStateMatcher implements ErrorStateMatcher {
-  match(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
+  isErrorSate(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
     return control ? !!(control.invalid && (control.dirty || (form && form.submitted))) : false;
   }
 }
@@ -20,7 +20,7 @@ export class ShowOnDirtyErrorStateMatcher implements ErrorStateMatcher {
 /** Provider that defines how form controls behave with regards to displaying error messages. */
 @Injectable()
 export class ErrorStateMatcher {
-  match(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
+  isErrorSate(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
     return control ? !!(control.invalid && (control.touched || (form && form.submitted))) : false;
   }
 }
--- a/src/lib/core/error/error-options.ts
+++ b/src/lib/core/error/error-options.ts
@@ -9,21 +9,18 @@
 import {Injectable} from '@angular/core';
 import {FormGroupDirective, NgForm, NgControl} from '@angular/forms';
 
-export type ErrorStateMatcher =
-  (control: NgControl | null, form: FormGroupDirective | NgForm | null) => boolean;
-
-/** Returns whether control is invalid and is either dirty or is a part of a submitted form. */
-export const showOnDirtyErrorStateMatcher: ErrorStateMatcher = (control, form) => {
-  return control ? !!(control.invalid && (control.dirty || (form && form.submitted))) : false;
-};
-
-/**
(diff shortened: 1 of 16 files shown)


jelbourn:
@kara and @mmalerba please take a look

src/lib/core/error/error-options.ts
```diff
@@ -7,27 +7,24 @@
  */
 
 import {InjectionToken} from '@angular/core';
-import {FormControl, FormGroupDirective, NgForm} from '@angular/forms';
+import {FormGroupDirective, NgForm, NgControl} from '@angular/forms';
 
 /** Injection token that can be used to specify the global error options. */
 export const MD_ERROR_GLOBAL_OPTIONS = new InjectionToken<ErrorOptions>('md-error-global-options');
 
 export type ErrorStateMatcher =
```
mmalerba: So I was thinking about maybe refactoring some of this error state matcher stuff because @g1shin recently had to implement a context-sensitive one for the stepper (#6117) and it felt a bit messier than it needed to be... What do you think about something like this:

```ts
export interface ErrorStateMatcher {
  isErrorState(...): boolean;
} 

export class DefaultErrorStateMatcher {
  isErrorState(...) { ... }
}

export class ShowOnDirtyErrorStateMatcher {
  isErrorState(...) { ... }
}
```

And then in error module:
```ts
@NgModule({
  providers: [{provide: ErrorStateMatcher, useClass: DefaultErrorStateMatcher}]
})
```

And finally in input/select/etc:
```ts
constructor(errorStateMatcher: ErrorStateMatcher) {
  console.log(errorStateMatcher.isErrorState(...));
}
```

I like this because:
1) We don't have to worry about the function being ripped off the class and losing its context
2) We don't have to worry about falling back to the defaultErrorStateMatcher in every component since we provide that in the error module
3) The code to inject in the constructor doesn't need a whole bunch of `@Optional() @Inject(...)` stuff

crisbeto: SGTM, but because the matcher is really just the function, it would look more along the lines of this:

```
export class ErrorOptions {
  isErrorState = (control, form) => control.invalid && (form.submitted || control.dirty);
  // more error-related config goes here in the future.
}
```

Also I'm not sure whether we should provide the `ShowOnDirtyErrorStateMatcher` by default since it's pretty basic logic that people can write themselves if they're overriding it anyway.

mmalerba: +1 Is there other stuff we plan to put in `ErrorOptions`? Just curious, I can't remember why we went with that over just a function

crisbeto: We don't have anything atm, but the intention is to avoid having to add any more providers if we decide to have more options.

willshowell: @crisbeto I added it only because it seemed to be a common question here and on SO https://github.com/angular/material2/pull/4750#issuecomment-307525083, https://github.com/angular/material2/pull/4750#discussion_r123363437

@mmalerba None that I was aware of. Simply based it off of `PlaceholderOptions` which seemed a nicely future proof

mmalerba: Ok sounds good, I do want to keep the interface though so that components like the stepper can implement it.

```ts
interface ErrorOptions { ... }
class DefaultErrorOptions implements ErrorOptions { ... }
```

crisbeto: Sure, although FWIW, you can implement a class as well.

mmalerba: Waaa I didn't know that, TS you crazy

crisbeto:
@mmalerba I ended up splitting the `ErrorOptions` and `MdError` directive into a separate module to make the DI a little simpler and since we'll probably end up needing the `MdError` for `MdSelect` at some point anyway. Can you take another look?

src/lib/core/error/error-options.ts
```diff
@@ -6,28 +6,27 @@
...
+  return control ? !!(control.invalid && (control.dirty || (form && form.submitted))) : false;
+};
+
+/**
+ * Provider that defines how form controls behave with
+ * regards to displaying error messages.
+ */
+@Injectable()
+export class ErrorOptions {
+  errorStateMatcher: ErrorStateMatcher = defaultErrorStateMatcher;
```
mmalerba: Can we do this method style instead? I want it to be obvious that it's safe to use `this` in the method

src/lib/input/input-container.ts
```diff
@@ -311,9 +295,8 @@ export class MdInputDirective implements OnChanges, OnDestroy, DoCheck {
   /** Re-evaluates the error state. This is only relevant with @angular/forms. */
   private _updateErrorState() {
     const oldState = this._isErrorState;
-    const control = this._ngControl;
-    const parent = this._parentFormGroup || this._parentForm;
-    const newState = control && this.errorStateMatcher(control.control as FormControl, parent);
+    const errorMatcher = this.errorStateMatcher || this._errorOptions.errorStateMatcher;
```
mmalerba: This is bad because it loses the error options context. Honestly, I think we should just get rid of `ErrorOptions` since there are currently no other options. and make `ErrorStateMatcher` a class rather than a function. This was we can easily support overriding via `@Input` like we want to do here and the stepper can easily provide its context-dependent matcher

crisbeto: I'm not sure how making the `ErrorStateMatcher` would help, considering that it would still look something like:

```
class ErrorStateMatcher {
  match(): boolean;
}
```

Otherwise I'll fix it so it doesn't lose the context.

src/lib/core/error/error-options.ts
crisbeto: Done.

src/lib/input/input-container.ts
mmalerba: Then instead of making the `Input()` take a function you make it take a `ErrorStateMatcher`. My other reason for not wanting `ErrorOptions` is that things like the stepper which just cares about the matcher then has to worry about all the other options too.

crisbeto: Alright, let's just go back to the injection token with a plain function and we can worry about a provider if we really need it.

crisbeto:
Updated again @mmalerba.

src/lib/input/input.md
```diff
@@ -125,21 +125,23 @@ the error messages.
...
+    const isSubmitted = form && form.submitted;
+    return !!(control.invalid && (control.dirty || control.touched || isSubmitted)));
+  }
 }
 ```
 
-A global error state matcher can be specified by setting the `MD_ERROR_GLOBAL_OPTIONS` provider. This applies
-to all inputs. For convenience, `showOnDirtyErrorStateMatcher` is available in order to globally cause
-input errors to show when the input is dirty and invalid.
+A global error state matcher can be specified by setting the `ErrorOptions` provider. This applies
```
mmalerba: ErrorStateMatcher

src/lib/input/input.md
```diff
@@ -125,21 +125,23 @@ the error messages.
...
+to all inputs. For convenience, `ShowOnDirtyErrorStateMatcher` is available in order to globally
+cause input errors to show when the input is dirty and invalid.
 
 ```ts
 @NgModule({
   providers: [
-    {provide: MD_ERROR_GLOBAL_OPTIONS, useValue: { errorStateMatcher: showOnDirtyErrorStateMatcher }}
+    {provide: ErrorOptions, useClass: ShowOnDirtyErrorStateMatcher}
```
mmalerba: ErrorStateMatcher

src/lib/select/select.spec.ts
```diff
@@ -2660,6 +2662,46 @@ describe('MdSelect', () => {
...
+
+      customErrorFixture.componentInstance.errorStateMatcher = { match: matcher };
+      customErrorFixture.detectChanges();
+
+      expect(component.select._isErrorState()).toBe(true);
+      expect(matcher).toHaveBeenCalled();
+    });
+
+    it('should be able to override the error matching behavior via the injection token', () => {
+      const errorOptions: ErrorStateMatcher = {
```
mmalerba: errorStateMatcher

src/lib/core/error/error-options.ts
```diff
@@ -6,28 +6,21 @@
...
-}
-
-/** Returns whether control is invalid and is either touched or is a part of a submitted form. */
-export function defaultErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm) {
-  const isSubmitted = form && form.submitted;
-  return !!(control.invalid && (control.touched || isSubmitted));
+/** Error state matcher that matches when a control is invalid and dirty. */
+@Injectable()
+export class ShowOnDirtyErrorStateMatcher implements ErrorStateMatcher {
+  match(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
```
mmalerba: can we call it `isErrorSate`? `match` is kind of generic and might conflict with some other interface someone is implementing

crisbeto:

src/lib/core/error/error-options.ts
```diff
@@ -6,28 +6,21 @@
...
-}
-
-/** Returns whether control is invalid and is either touched or is a part of a submitted form. */
-export function defaultErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm) {
-  const isSubmitted = form && form.submitted;
-  return !!(control.invalid && (control.touched || isSubmitted));
+/** Error state matcher that matches when a control is invalid and dirty. */
+@Injectable()
+export class ShowOnDirtyErrorStateMatcher implements ErrorStateMatcher {
+  isErrorSate(control: NgControl | null, form: FormGroupDirective | NgForm | null): boolean {
```
kara: This public API has a typo - `isErrorSate` -> `isErrorState`. Should be fixed across the PR.

crisbeto: Ops, good catch.

mmalerba:
@jelbourn would you rather move `md-error` into `core/error/` or `form-field/` as #6331 currently does.

jelbourn:
@mmalerba `form-field` is good

crisbeto:
In that case let's defer this PR until the form field gets in.

mmalerba:
I think we can resume work on this one now

crisbeto:
@mmalerba a lot changed since we parked this PR. I re-did it, excluded the `mat-error` changes and resubmitted as https://github.com/angular/material2/pull/7443.

Issue #6140, opened 2017-07-29
Different bottom linewidth for input and select when error


princemaple:
#### Bug, feature request, or proposal:
bug

![image](https://user-images.githubusercontent.com/1329716/28744139-0bf62d70-749e-11e7-9050-ee6996c8f86f.png)

#### What is the expected behavior?
be the same? I couldn't find `select` in material design spec

#### What is the current behavior?
different

#### What are the steps to reproduce?
http://plnkr.co/edit/2QFkVaZdoI6fvNI8nTzk?p=preview

#### What is the use-case or motivation for changing an existing behavior?
to be consistent

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
4.3.2, 2-beta.8, others irrelevant

donroyco:
Good catch. See #6041 and (partly) #2124

princemaple:
Hmm, seems a little different but definitely related. should I close this one?

mmalerba:
This will be addressed as part of the fix for #2124, so closing as a duplicate of that

PR #6152, opened 2017-07-30
fix(input): underline should only be thicker if focused

DevVersion:
* Currently the underline of the input container is always `2px` thick if the input is invalid. As per specifications the underline should be `1dp/px` if not focused.

@mmalerba The specs are not too clear about the intended behavior here. There are sections about the thickness of the underline, but not specifically in relation to error handling.

References #6140

### DIFF
--- a/src/lib/input/input-container.scss
+++ b/src/lib/input/input-container.scss
@@ -207,7 +207,7 @@ textarea.mat-input-element {
 
   .mat-input-ripple {
     position: absolute;
-    height: $mat-input-underline-height * 2;
+    height: $mat-input-underline-height;
     top: 0;
     left: 0;
     width: 100%;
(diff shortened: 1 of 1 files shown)


mmalerba:
I actually made this change intentionally (#4506) I agree that the spec isn't 100% clear, but I think they intended to make error state `2px` always (see [summary chart](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0B5ZSepuCX1xOb195LU9KWEdIeEk/states.png)). Actually according to the chart hover state should be `2px` too, which it currently isn't.

As for #6140, yes the select and the input currently differ in a number of ways that makes them look really bad when placed side by side. We need to refactor them to use the same CSS, I would like to use the CSS for the input since I've spent a lot of time making sure that it matches the spec relatively well.

mmalerba:
Ok given that the official spec is kind of vague on this I'm ok with changing it back to think only on focus. I think its better from an a11y standpoint for users to be able to visually distinguish focused from unfocused state.

PR #6155, opened 2017-07-30
demos(*): fix inputs spilling over when screen width small

amcdnl:
Fixes the autocomplete and input examples with a fixed 500px width from spilling over on smaller device screens.

img


fxck:
Wouldn't it be worth adding a global `.example-form` class rather than having the same repeat in each demo.. or at least use variable for those values?

jelbourn:
@fxck the examples are all meant to be totally standalone so you can just look at one to see what's going on

PR #6164, opened 2017-07-31
fix(dialog): better handling of custom ViewContainerRef with OnPush change detection

crisbeto:
* Adds a `markForCheck` call after starting the dialog animation, in order to ensure that it animates away if it was opened from a component with OnPush change detection.
* Since we can't know whether the animation will start right after the `markForCheck`, these changes switch to starting the backdrop animation together with the dialog animation. This avoids some slight timing issues where the backdrop can start its animation a little too early.
* Simplifies the dialog animation streams to avoid having to subscribe to the `completed` callback in the `MdDialogRef`.
* Fixes the demo app sidenav jumping to the bottom when it is opened from the homepage.
* Fixes some alignment in the dialog demo that got thrown off by the new input width.

The dialog changes should solve issues like #5118.

@mmalerba I've reverted the changes from #6026 since it should work correctly with the `ViewContainerRef` now.

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -16,6 +16,7 @@ import {
   EventEmitter,
   Inject,
   Optional,
+  ChangeDetectorRef,
 } from '@angular/core';
 import {
   animate,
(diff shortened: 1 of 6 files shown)


src/lib/dialog/dialog-container.ts
```diff
@@ -171,15 +169,25 @@ export class MdDialogContainer extends BasePortalHost {
...
+  /** Callback, invoked when an animation on the host starts. */
+  _onAnimationStart(event: AnimationEvent) {
+    this._isAnimating = true;
+    this._animationStateChanged.emit(event);
+  }
+
+  /** Starts the dialog exit animation. */
+  _startExitAnimation(): void {
+    this._state = 'exit';
+    this._changeDetectorRef.markForCheck();
```
jelbourn: Add a comment that explains why this does `markForCheck`?

PR #6168, opened 2017-07-31
a11y: add service to add aria-describedby labels

andrewseguin:
Service adds a new element in the document body that contains the message. This is pointed to by the host element using `aria-describedby` for screenreaders to read out.

Used by tooltip and improved its accessibility. (add show on focus, escape keyboard shortcut)

### DIFF
--- a/src/cdk/a11y/aria-describer.spec.ts
+++ b/src/cdk/a11y/aria-describer.spec.ts
@@ -1,0 +1,31 @@
+import {A11yModule, AriaDescriber} from '@angular/cdk/a11y';
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
+import {Component} from '@angular/core';
+
+fdescribe('AriaDescriber', () => {
+  let ariaDescriber: AriaDescriber;
+  let fixture: ComponentFixture<TestApp>;
+
+  describe('with default element', () => {
+
+    beforeEach(async(() => {
+      TestBed.configureTestingModule({
+        imports: [A11yModule],
+        declarations: [TestApp],
+      }).compileComponents();
+    }));
+
+    beforeEach(() => {
+      fixture = TestBed.createComponent(TestApp);
+    });
+
+    it('should test', () => {
+      console.log(fixture.componentInstance.ariaDescriber);
+    });
+  });
+});
+
+@Component({template: `<div #element1></div>`})
+class TestApp {
+  constructor(public ariaDescriber: AriaDescriber) { }
+}
--- a/src/cdk/a11y/aria-describer.spec.ts
+++ b/src/cdk/a11y/aria-describer.spec.ts
@@ -1,31 +1,185 @@
-import {A11yModule, AriaDescriber} from '@angular/cdk/a11y';
+import {A11yModule} from './index';
+import {AriaDescriber, MESSAGES_CONTAINER_ID} from './aria-describer';
 import {async, ComponentFixture, TestBed} from '@angular/core/testing';
-import {Component} from '@angular/core';
+import {Component, ElementRef, ViewChild} from '@angular/core';
 
-fdescribe('AriaDescriber', () => {
+describe('AriaDescriber', () => {
   let ariaDescriber: AriaDescriber;
+  let component: TestApp;
   let fixture: ComponentFixture<TestApp>;
 
-  describe('with default element', () => {
+  describe('with component provider', () => {
 
     beforeEach(async(() => {
       TestBed.configureTestingModule({
         imports: [A11yModule],
-        declarations: [TestApp],
+        declarations: [TestApp]
       }).compileComponents();
     }));
 
     beforeEach(() => {
       fixture = TestBed.createComponent(TestApp);
+      component = fixture.componentInstance;
+      ariaDescriber = component.ariaDescriber;
     });
 
-    it('should test', () => {
-      console.log(fixture.componentInstance.ariaDescriber);
+    afterEach(() => {
+      ariaDescriber._unregisterAllMessages();
+    });
+
+    it('should initialize without the message container', () => {
+      expect(getMessagesContainer()).toBeNull();
+    });
+
+    it('should be able to create a message element', () => {
+      ariaDescriber.addDescription(component.container1.nativeElement, 'My Message');
+      expectMessages(['My Message']);
+    });
+
+    it('should not register empty strings', () => {
+      ariaDescriber.addDescription(component.container1.nativeElement, '');
+      expect(getMessageElements()).toBe(null);
+    });
+
+    it('should de-dupe a message registered multiple times', () => {
+      ariaDescriber.addDescription(component.container1.nativeElement, 'My Message');
+      ariaDescriber.addDescription(component.container2.nativeElement, 'My Message');
+      ariaDescriber.addDescription(component.container3.nativeElement, 'My Message');
+      expectMessages(['My Message']);
+      expectMessage(component.container1.nativeElement, 'My Message');
+    });
+
+    it('should be able to register multiple messages', () => {
+      ariaDescriber.addDescription(component.container1.nativeElement, 'First Message');
+      ariaDescriber.addDescription(component.container2.nativeElement, 'Second Message');
+      expectMessages(['First Message', 'Second Message']);
+      expectMessage(component.container1.nativeElement, 'First Message');
+      expectMessage(component.container2.nativeElement, 'Second Message');
+    });
+
+    it('should be able to unregister messages', () => {
+      ariaDescriber.addDescription(component.container1.nativeElement, 'My Message');
+      expectMessages(['My Message']);
+
+      // Register again to check dedupe
+      ariaDescriber.addDescription(component.container2.nativeElement, 'My Message');
+      expectMessages(['My Message']);
+
+      // Unregister one message and make sure the message is still present in the container
+      ariaDescriber.removeDescription(component.container1.nativeElement, 'My Message');
+      expectMessages(['My Message']);
+
+      // Unregister the second message, message container should be gone
+      ariaDescriber.removeDescription(component.container2.nativeElement, 'My Message');
+      expect(getMessagesContainer()).toBeNull();
+    });
+
+    it('should be able to unregister messages while having others registered', () => {
+      ariaDescriber.addDescription(component.container1.nativeElement, 'Persistent Message');
+      ariaDescriber.addDescription(component.container2.nativeElement, 'My Message');
+      expectMessages(['Persistent Message', 'My Message']);
+
+      // Register again to check dedupe
+      ariaDescriber.addDescription(component.container3.nativeElement, 'My Message');
+      expectMessages(['Persistent Message', 'My Message']);
+
+      // Unregister one message and make sure the message is still present in the container
+      ariaDescriber.removeDescription(component.container2.nativeElement, 'My Message');
+      expectMessages(['Persistent Message', 'My Message']);
+
+      // Unregister the second message, message container should be gone
+      ariaDescriber.removeDescription(component.container3.nativeElement, 'My Message');
+      expectMessages(['Persistent Message']);
+    });
+
+    it('should be able to append to an existing list of aria describedby', () => {
+      ariaDescriber.addDescription(component.container4.nativeElement, 'My Message');
+      expectMessages(['My Message']);
+      expectMessage(component.container4.nativeElement, 'My Message');
+    });
+  });
+
+  describe('without provider', () => {
+    beforeEach(async(() => {
+      TestBed.configureTestingModule({
+        imports: [A11yModule],
+        declarations: [TestAppWithoutProvider]
+      }).compileComponents();
+    }));
+
+    it('should fail to instantiate due to not importing describer as part of component', () => {
+      expect(() => {
+        TestBed.createComponent(TestAppWithoutProvider);
+      }).toThrow();
     });
   });
 });
 
-@Component({template: `<div #element1></div>`})
+function getMessagesContainer() {
+  return document.querySelector(`#${MESSAGES_CONTAINER_ID}`);
+}
+
+function getMessageElements(): Node[] | null {
+  const messagesContainer = getMessagesContainer();
+  if (!messagesContainer) { return null; }
+
+  return messagesContainer ?  Array.prototype.slice.call(messagesContainer.children) : null;
+}
+
+/** Checks that the messages array matches the existing created message elements. */
+function expectMessages(messages: string[]) {
+  const messageElements = getMessageElements();
+  if (!messageElements) {
+    fail(`Expected messages ${messages} but there were no message elements`);
+    return;
+  }
+
+  expect(messages.length).toBe(messageElements.length);
+  messages.forEach((message, i) => {
+    expect(messageElements[i].textContent).toBe(message);
+  });
+}
+
+/** Checks that an element points to a message element that contains the message. */
+function expectMessage(el: HTMLElement, message: string) {
+  const ariaDescribedBy = el.getAttribute('aria-describedby');
+  if (!ariaDescribedBy) {
+    fail(`No aria-describedby for ${el}`);
+    return;
+  }
+
+  const messages: string[] = [];
+  ariaDescribedBy.split(' ').forEach(referenceId => {
+    const messageElement = document.querySelector(`#${referenceId}`);
+    if (messageElement) {
+      messages.push(messageElement.textContent || '');
+    }
+  });
+
+  expect(messages).toContain(message);
+}
+
+@Component({
+  template: `
+    <div #container1></div>
+    <div #container2></div>
+    <div #container3></div>
+    <div #container4 aria-describedby="existing-aria-describedby1 existing-aria-describedby2"></div>
+  `,
+  providers: [AriaDescriber],
+})
 class TestApp {
+  @ViewChild('container1') container1: ElementRef;
+  @ViewChild('container2') container2: ElementRef;
+  @ViewChild('container3') container3: ElementRef;
+  @ViewChild('container4') container4: ElementRef;
+
+  constructor(public ariaDescriber: AriaDescriber) { }
+}
+
+@Component({
+  template: `<div #element1></div>`,
+})
+class TestAppWithoutProvider {
   constructor(public ariaDescriber: AriaDescriber) { }
 }
--- a/src/cdk/a11y/aria-describer.spec.ts
+++ b/src/cdk/a11y/aria-describer.spec.ts
@@ -100,6 +100,13 @@ describe('AriaDescriber', () => {
     expectMessages(['My Message']);
     expectMessage(component.element4, 'My Message');
   });
+
+  it('should be able to handle multiple regisitrations of the same message to an element', () => {
+    ariaDescriber.describe(component.element1, 'My Message');
+    ariaDescriber.describe(component.element1, 'My Message');
+    expectMessages(['My Message']);
+    expectMessage(component.element1, 'My Message');
+  });
 });
 
 function getMessagesContainer() {
--- a/src/cdk/a11y/aria-describer.spec.ts
+++ b/src/cdk/a11y/aria-describer.spec.ts
@@ -1,4 +1,4 @@
-import {A11yModule} from './index';
+import {A11yModule, CDK_DESCRIBEDBY_HOST_ATTRIBUTE} from './index';
 import {AriaDescriber, MESSAGES_CONTAINER_ID} from './aria-describer';
 import {async, ComponentFixture, TestBed} from '@angular/core/testing';
 import {Component, ElementRef, ViewChild} from '@angular/core';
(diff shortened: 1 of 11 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +426,113 @@ export class MdTooltip implements OnDestroy {
...
+  /**
+   * Removes the a11y tooltip message element if no other tooltips are registered with the same
+   * message. Otherwise, decrease the reference count of the registered a11y tooltip message.
+   */
+  private _unregisterA11yMessage(message: string) {
+    if (!this._platform.isBrowser || !this.message) { return; }
+
+    // Remove the a11y message if this was its last unique instance.
+    const a11yMessageElement = a11yMessages.get(message);
+    if (a11yMessageElement && --a11yMessageElement.count == 0) {
```
crisbeto: This might be a little more readable as `if (a11yMessageElement && a11yMessageElement.count === 1)` and then doing the decrement separately.

src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +426,113 @@ export class MdTooltip implements OnDestroy {
...
+      this._deleteA11yMessageElement(message);
+    }
+
+    // If the global messages container no longer has any children, remove it.
+    if (!this._getA11yMessagesContainer()!.childNodes.length) {
+      document.body.removeChild(this._getA11yMessagesContainer()!);
```
crisbeto: You can use the `Renderer2` to create/add/remove these elements.

src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +426,113 @@ export class MdTooltip implements OnDestroy {
...
+    if (!this._platform.isBrowser) { return; }
+
+    // Create a visually hidden container for the accessibility messages if one does not yet exist.
+    if (!this._getA11yMessagesContainer()) {
+      this._createA11yMessagesContainer();
+    }
+
+    const messageElement = document.createElement('div');
+    messageElement.id = `md-tooltip-message-${latestA11yMessageId++}`;
+    messageElement.innerHTML = message;
```
crisbeto: This looks a little dangerous. Perhaps we should use `textContent` instead since the spec doesn't allow rich text inside tooltips, IIRC?

crisbeto: This one also needs to be updated if the tooltip text changes.

src/lib/tooltip/tooltip.ts
```diff
@@ -149,8 +169,12 @@ export class MdTooltip implements OnDestroy {
   /** The message to be displayed in the tooltip */
   @Input('mdTooltip') get message() { return this._message; }
   set message(value: string) {
-    this._message = value;
-    this._setTooltipMessage(this._message);
+    if (this._message) { this._unregisterA11yMessage(this._message); }
+
+    // If the message is not a string (e.g. number), convert it to a string and trim it.
+    this._message = value ? `${value}`.trim() : '';
```
crisbeto: Might be a good idea to have a test for this case in particular (passing in a non-string to the message). We have another pending PR that deals with the same issue: https://github.com/angular/material2/pull/6146

src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +426,113 @@ export class MdTooltip implements OnDestroy {
...
+    } else {
+      this._renderer.setAttribute(this._elementRef.nativeElement, 'aria-describedby', '');
+    }
+  }
+
+  /** Creates the global container for all tooltip a11y messages. */
+  private _createA11yMessagesContainer() {
+    const a11yMessagesContainer = document.createElement('div');
+    a11yMessagesContainer.id = A11Y_MESSAGES_CONTAINER_ID;
+    a11yMessagesContainer.className = 'cdk-visually-hidden';
```
crisbeto: It's unlikely for this to clobber any classes, but it's still better to use the renderer or `classList`.

src/lib/tooltip/tooltip.ts
andrewseguin: In the `set message` function, we'll unregister the old message and register the new one

andrewseguin:
Good call on the renderer stuff, wasn't even thinking about it

src/lib/tooltip/tooltip.ts
```diff
@@ -83,6 +84,23 @@ export const MD_TOOLTIP_SCROLL_STRATEGY_PROVIDER = {
...
+export interface RegisteredA11yMessage {
+  element: HTMLElement;
+  count: number;
+}
+
+/** ID used for the body container where all a11y messages are appended. */
+export const A11Y_MESSAGES_CONTAINER_ID = 'md-tooltip-a11y-messages';
+
+/** Global incremental identifier for each registered a11y message. */
+let latestA11yMessageId = 0;
```
jelbourn: I'd be okay with just calling this `nextId` (which is what we typically name these).

src/lib/tooltip/tooltip.ts
```diff
@@ -149,8 +170,12 @@ export class MdTooltip implements OnDestroy {
   /** The message to be displayed in the tooltip */
   @Input('mdTooltip') get message() { return this._message; }
   set message(value: string) {
-    this._message = value;
-    this._setTooltipMessage(this._message);
+    if (this._message) { this._unregisterA11yMessage(this._message); }
+
+    // If the message is not a string (e.g. number), convert it to a string and trim it.
+    this._message = value ? `${value}`.trim() : '';
```
jelbourn:
```ts
this._message = `${value || ''}`.trim();
```
?

src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +427,113 @@ export class MdTooltip implements OnDestroy {
...
+      this._renderer.setAttribute(
+          this._elementRef.nativeElement, 'aria-describedby', tooltipMessageElementId);
+    } else {
+      this._renderer.setAttribute(this._elementRef.nativeElement, 'aria-describedby', '');
+    }
+  }
+
+  /** Creates the global container for all tooltip a11y messages. */
+  private _createA11yMessagesContainer() {
+    const a11yMessagesContainer = this._renderer.createElement('div');
```
jelbourn: This needs `aria-hidden` as well. Otherwise screen-reader users will run into this content at the end of the document without context

src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +427,113 @@ export class MdTooltip implements OnDestroy {
...
+
+
+  /**
+   * Registers the tooltip message for accessibility. If this message has not yet been
+   * registered, a new element will be created in a visually hidden container with the message
+   * as its content. This provides screenreaders a reference the tooltip trigger's aria-describedby.
+   * If an element has already been created for this message, increase that registered message's
+   * reference count.
+   */
+  private _registerA11yMessage() {
```
jelbourn: What do you think about breaking all the code around this approach into a separate class called something like `AriaDescriber`? It would be good to capture all of this logic in one place separate from the tooltip and we could potentially re-use it in the future.

src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +427,113 @@ export class MdTooltip implements OnDestroy {
...
+  /** Creates the global container for all tooltip a11y messages. */
+  private _createA11yMessagesContainer() {
+    const a11yMessagesContainer = this._renderer.createElement('div');
+    a11yMessagesContainer.id = A11Y_MESSAGES_CONTAINER_ID;
+    this._renderer.addClass(a11yMessagesContainer, 'cdk-visually-hidden');
+    this._renderer.appendChild(document.body, a11yMessagesContainer);
+  }
+
+  /** Returns the global container for tooltip a11y messages. */
+  private _getA11yMessagesContainer() {
```
jelbourn: Why not just remember the element?

src/lib/tooltip/tooltip.ts
```diff
@@ -392,6 +427,113 @@ export class MdTooltip implements OnDestroy {
...
+
+  /** Returns the trigger's aria-describedby attribute. */
+  private _getAriaDescribedby(): string {
+    return this._elementRef.nativeElement.getAttribute('aria-describedby');
+  }
+
+  /** Sets the trigger's aria-describedby attribute to the tooltip message element. */
+  private _setAriaDescribedBy() {
+    // Return if an aria-describedby already exists and it is not referencing a tooltip message.
+    const ariaDescribedBy = this._elementRef.nativeElement.getAttribute('aria-describedby');
```
jelbourn: Something I realized just now is that rather than clobbering an existing aria-describedby, we should augment it with the new ID (the attribute can have multiple, comma-delimited IDs). I talked to @mmalerba about doing the same thing for inputs with `md-error` and `md-hint`. We should create a couple utility functions in cdk/a11y to add/remove an ID to an element's  aria attribute. Something like...
```ts
addAriaReferencedId(element, id, 'aria-describedby');
removeAriaReferencedId(element, id, 'aria-describedby');
```
(open to coming up with a better name)

andrewseguin:
Updated now so that the majority of the change is by adding aria-describer. Usage from tooltip is still included to show how it is used and its minimal code. Ready for review

src/lib/tooltip/tooltip.ts
andrewseguin: This is done to address an issue where numbers are being used as Input and are throwing an error because they do not have `trim()`. This converts the input to string and trims it.

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+let messagesContainer: HTMLElement | null = null;
+
+/**
+ * Utility that creates visually hidden elements with a message content. Useful for elements that
+ * want to use aria-describedby to further describe themselves without adding additional visual
+ * content. Must be provided with as part of a component's injection tree so that it has access
+ * to the Renderer.
+ */
+@Injectable()
+export class AriaDescriber {
```
jelbourn: Let's make this `@docs-private` for now; I don't think this will be a public API before we're out of beta

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+  constructor(private _platform: Platform,
+              @Optional() private _renderer: Renderer2) {
+    if (!_renderer) {
+      throw Error('AriaDescriber must be provided through a component\'s provider decorator ' +
```
jelbourn: Can use `"` or `` ` `` for the string instead of escaping the `'`

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/** ID used for the body container where all messages are appended. */
+export const MESSAGES_CONTAINER_ID = 'cdk-aria-describedby-messages';
+
+/** Global incremental identifier for each registered message element. */
+let nextId = 0;
+
+/** Global map of all registered message elements that have been placed into the document. */
+const registeredMessages = new Map<string, RegisteredMessage>();
```
jelbourn: `messageRegistry`?

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+import {Platform} from '@angular/cdk/platform';
+import {addAriaReferencedId, removeAriaReferencedId} from './aria-reference';
+
+/**
+ * Interface used to register message elements and keep a count of how many registrations have
+ * the same message and the reference to the message element used for the aria-describedby.
+ */
+export interface RegisteredMessage {
+  messageElement: HTMLElement;
+  hostElements: HTMLElement[];
```
jelbourn: Why do you need to keep track of these elements? I'm thinking this may cause a memory leak if references to elements are preserved here when the elements have been removed.

Could you accomplish what you need by adding an attribute to the described element?

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+    registeredMessages.clear();
+  }
+
+  /**
+   * Creates a new element in the visually hidden message container element with the message
+   * as its content.
+   */
+  private _createMessageElement(message: string): HTMLElement {
+    const messageElement = this._renderer.createElement('div');
+    this._renderer.setAttribute(messageElement, 'id', `cdk-aria-describedby-message-${nextId++}`);
```
jelbourn: How about just `cdk-describedby-x`?

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Utility that creates visually hidden elements with a message content. Useful for elements that
+ * want to use aria-describedby to further describe themselves without adding additional visual
+ * content. Must be provided with as part of a component's injection tree so that it has access
+ * to the Renderer.
+ */
+@Injectable()
+export class AriaDescriber {
+  constructor(private _platform: Platform,
+              @Optional() private _renderer: Renderer2) {
```
jelbourn: Doing this all without the Renderer is the preferred way to go now

(I consulted w/ Misko about this, he agreed that we should generally avoid using Render and prefer the native APIs)

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+    registeredMessages.delete(message);
+  }
+
+  /** Creates the global container for all aria-describedby messages. */
+  private _createMessagesContainer() {
+    messagesContainer = this._renderer.createElement('div');
+
+    this._renderer.setAttribute(messagesContainer, 'id', MESSAGES_CONTAINER_ID);
+    this._renderer.setAttribute(messagesContainer, 'aria-hidden', 'true');
+    this._renderer.addClass(messagesContainer, 'cdk-visually-hidden');
```
jelbourn: Can this just be `display: none`? IIRC, elements w/ `display: none` can still act as labels / descriptions for `aria-labelledby` and `aria-describedby`

src/cdk/a11y/aria-describer.ts
jelbourn: This class needs an `ngOnDestroy` hook that cleans everything up

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/** IDs are deliminated by an empty space, as per the spec. */
+const ID_DELIMINATOR = ' ';
```
jelbourn: We should probably be tolerant of any amount of whitespace between IDs rather than a single space (even though the spec calls for a single space)

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/** Adds an aria reference ID to an element's aria property if it is not already present. */
+export function addAriaReferencedId(el: HTMLElement, id: string, attr: string) {
```
jelbourn: `element`, `attributeName`, `id` feels like a more natural argument order

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/** Removes an aria reference ID from an element's aria property. */
+export function removeAriaReferencedId(el: HTMLElement, id: string, attr: string) {
+  const ids = getAriaReferenceIds(el, attr);
+  const filteredIds = ids.filter(val => val != id.trim());
+
+  el.setAttribute(attr, filteredIds.join(ID_DELIMINATOR));
+}
+
+/** Returns a list of an element's aria reference IDs for the provided aria attribute. */
```
jelbourn:
```
Gets the list of IDs referenced by the given ARIA attribute on an element. Used for attributes such as aria-labelledby, aria-owns, etc.
```

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,33 @@
...
+/** Adds an aria reference ID to an element's aria property if it is not already present. */
+export function addAriaReferencedId(el: HTMLElement, id: string, attr: string) {
+  const ids = getAriaReferenceIds(el, attr);
+  if (ids.some(existingId => existingId.trim() == id.trim())) { return; }
+  ids.push(id.trim());
+
+  el.setAttribute(attr, ids.join(ID_DELIMINATOR));
+}
+
+/** Removes an aria reference ID from an element's aria property. */
```
jelbourn:
```
Removes the given ID from the specified ARIA attribute on an element. Used for attributes such as aria-labelledby, aria-owns, etc.
```

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/** IDs are deliminated by an empty space, as per the spec. */
+const ID_DELIMINATOR = ' ';
+
+/** Adds an aria reference ID to an element's aria property if it is not already present. */
```
jelbourn:
```
Adds the given ID to the specified ARIA attribute on an element. Used for attributes such as aria-labelledby, aria-owns, etc.
```

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Adds to the host element an aria-describedby reference to a hidden element that contains
+   * the message. If the same message has already been registered, then it will reuse the created
+   * message element.
+   */
+  addDescription(hostElement: HTMLElement, message: string) {
+    if (!this._platform.isBrowser || !message.trim()) { return; }
+
+    if (!registeredMessages.get(message)) {
+      const messageElement = this._createMessageElement(message);
```
mmalerba: Is there no way to use existing DOM elements as description messages? For the input we always already have a `md-hint` or `md-error` element with the message.

Maybe you could make a directive that uses this service under the hood for existing elements to use?

Something like:
```ts
<span [cdkDescribes]="listOfElementsIDescribe">...</span>
```
Or the other way around:
```ts
<span [cdkDescribedBy]="listOfElementsThatDescribeMe">...</span>
```

src/cdk/a11y/aria-describer.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  if (!messagesContainer) { return null; }
+
+  return messagesContainer ?  Array.prototype.slice.call(messagesContainer.children) : null;
+}
+
+/** Checks that the messages array matches the existing created message elements. */
+function expectMessages(messages: string[]) {
+  const messageElements = getMessageElements();
+  if (!messageElements) {
+    fail(`Expected messages ${messages} but there were no message elements`);
```
crisbeto: Why not have a regular `expect` here?

src/cdk/a11y/aria-describer.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+/** Checks that an element points to a message element that contains the message. */
+function expectMessage(el: HTMLElement, message: string) {
+  const ariaDescribedBy = el.getAttribute('aria-describedby');
+  if (!ariaDescribedBy) {
+    fail(`No aria-describedby for ${el}`);
+    return;
+  }
+
+  const messages: string[] = [];
+  ariaDescribedBy.split(' ').forEach(referenceId => {
```
crisbeto: You can condense this logic with an array `map` or `reduce`.

src/cdk/a11y/aria-describer.ts
crisbeto: You can get access to a renderer without going through a component. See https://github.com/angular/material2/pull/5840

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  /**
+   * Adds to the host element an aria-describedby reference to a hidden element that contains
+   * the message. If the same message has already been registered, then it will reuse the created
+   * message element.
+   */
+  addDescription(hostElement: HTMLElement, message: string) {
+    if (!this._platform.isBrowser || !message.trim()) { return; }
```
crisbeto: The `message.trim` will fail if the value is not a string.

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (!registeredMessages.get(message)) {
```
crisbeto: It's pretty much the same, but you can use `registeredMessages.has(message)` here.

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+    const registeredMessage = registeredMessages.get(message)!;
+    registeredMessage.hostElements.push(hostElement);
+    addAriaReferencedId(hostElement, registeredMessage.messageElement.id, 'aria-describedby');
+  }
+
+  /**
+   * Removes the host element's aria-describedby reference to the message element.
+   */
+  removeDescription(hostElement: HTMLElement, message: string) {
+    if (!this._platform.isBrowser || !message.trim()) { return; }
```
crisbeto: This trim has the potential to fail as well.

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const registeredMessage = registeredMessages.get(message)!;
+    registeredMessage.hostElements = registeredMessage.hostElements.filter(el => el != hostElement);
+    removeAriaReferencedId(hostElement, registeredMessage.messageElement.id, 'aria-describedby');
+
+    if (registeredMessage.hostElements.length == 0) {
```
crisbeto: Either triple equals or `!registeredMessage.hostElements.length` here and below.

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,135 @@
...
+      this._deleteMessageElement(message);
+    }
+
+    if (messagesContainer!.childNodes.length == 0) {
+      this._deleteMessagesContainer();
+    }
+  }
+
+  /** Unregisters all created message elements and removes the message container. */
+  _unregisterAllMessages() {
```
crisbeto: This one could be renamed to `ngOnDestroy` which has the advantage of being called by Angular when appropriate.

src/cdk/a11y/aria-describer.ts
andrewseguin: Went with `cdk-describedby-##` for the messages, and `cdk-described` for the elements that are described.

andrewseguin:
Updated based on the comments, thanks

src/lib/tooltip/tooltip.ts
jelbourn: My suggestion still stringifies the value but omits the ternary.

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,149 @@
...
+   */
+  describe(hostElement: Element, message: string) {
+    if (!this._platform.isBrowser || !`${message}`.trim()) { return; }
+
+    if (!messageRegistry.has(message)) {
+      createMessageElement(message);
+    }
+
+    const registeredMessage = messageRegistry.get(message)!;
+    registeredMessage.referenceCount++;
```
jelbourn: Only update the reference count if the element is not already described by the message? (can check the existing IDs)

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  el.setAttribute(attr, filteredIds.join(ID_DELIMINATOR));
+}
+
+/**
+ * Gets the list of IDs referenced by the given ARIA attribute on an element.
+ * Used for attributes such as aria-labelledby, aria-owns, etc.
+ */
+export function getAriaReferenceIds(el: Element, attr: string): string[] {
+  const idsValue = (el.getAttribute(attr) || '').trim();
```
jelbourn: `attributeValue`?

Since you're matching all non-whitespace, do you need the trim?

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  // Get string array of all individual ids (whitespace deliminated) in the attribute value
+  const idsArray = idsValue.match(/\S+/g) || [];
```
jelbourn: Just `ids`? 

Since you're matching all non-whitespace, do you need the map to trim?

src/cdk/a11y/aria-reference.ts
```diff
@@ -0,0 +1,46 @@
...
+  el.setAttribute(attr, ids.join(ID_DELIMINATOR));
+}
+
+/**
+ * Removes the given ID from the specified ARIA attribute on an element.
+ * Used for attributes such as aria-labelledby, aria-owns, etc.
+ */
+export function removeAriaReferencedId(el: Element, attr: string, id: string) {
+  const ids = getAriaReferenceIds(el, attr);
+  const filteredIds = ids.filter(val => val != id.trim());
```
jelbourn: How would you end up with an empty string here since `getAriaReferenceIds` only targets non-whitespace?

src/cdk/a11y/aria-reference.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * Expects the equal array from getAriaReferenceIds and a space-deliminated list from
+   * the actual element attribute. If ids is empty, assumes the element should not have any
+   * value
+   */
+  function expectIds(ids: string[]) {
+    expect(getAriaReferenceIds(testElement!, 'aria-describedby')).toEqual(ids);
+    expect(testElement!.getAttribute('aria-describedby')).toBe(ids.length ? ids.join(' ') : '');
```
jelbourn: This is a bit too much logic in the test for my taste. I'd generally prefer to have the assertion on `aria-describedby` to be hard-coded into each test

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,149 @@
...
+let messagesContainer: HTMLElement | null = null;
+
+/**
+ * Utility that creates visually hidden elements with a message content. Useful for elements that
+ * want to use aria-describedby to further describe themselves without adding additional visual
+ * content.
+ * @docs-private
+ */
+@Injectable()
+export class AriaDescriber {
```
jelbourn: Need to add a provider for `AriaDescriber` similar to `LIVE_ANNOUNCER_PROVIDER`
(provides the existing instances if there is one)

src/cdk/a11y/aria-describer.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/** Deletes the global messages container. */
+function deleteMessagesContainer() {
+  document.body.removeChild(messagesContainer!);
+  messagesContainer = null;
+}
+
+function removeCdkDescribedByReferenceIds(element: Element) {
+  // Remove all aria-describedby reference IDs that are prefixed by CDK_DESCRIBEDBY_ID_PREFIX
```
jelbourn: Make this a function description?

src/lib/tooltip/tooltip.ts
```diff
@@ -149,8 +152,14 @@ export class MdTooltip implements OnDestroy {
...
   @Input('mdTooltip') get message() { return this._message; }
   set message(value: string) {
-    this._message = value;
-    this._setTooltipMessage(this._message);
+    if (this._message) {
+      this._ariaDescriber.removeDescription(this._elementRef.nativeElement, this._message);
+    }
+
+    // If the message is not a string (e.g. number), convert it to a string and trim it.
+    this._message = value ? `${value}`.trim() : '';
```
jelbourn: Just `` `${value}`.trim()`` ?

src/cdk/a11y/aria-reference.ts
andrewseguin: Looks like this needed to be revisited based on the regex, before the single space deliminator didn't trim up as much as I liked. The regex approach works much nicer

src/lib/tooltip/tooltip.ts
crisbeto: Wouldn't that end up with the string `undefined` if `value === undefined`?

andrewseguin: As mentioned by @crisbeto, an `undefined` message will evaluate  `` `${value}`.trim()`` as `'undefined'`

jelbourn: Ah, right; that's why my original comment was `` `${value || ''}`.trim() ``

andrewseguin:
Phew, ready found another round

src/cdk/a11y/aria-reference.spec.ts
andrewseguin: Changed it around a bit, kept expectIds but takes in an attr. Removed the helper functions for adding/removing the references

kara:
@andrewseguin  Rebase?

jelbourn:
@andrewseguin this is failing the prerender task

andrewseguin:
Fixed prerender (the aria describer's ngOnDestroy was referencing the document). Added a11y docs for review as well @tinayuangao

PR #6169, opened 2017-07-31
fix(table): Fix change det issue with http example

andrewseguin:


src/material-examples/table-http/table-http-example.ts
```diff
@@ -68,8 +68,8 @@ export class ExampleHttpDao {
 export class ExampleDataSource extends DataSource<GithubIssue> {
   // The number of issues returned by github matching the query.
   resultsLength = 0;
-  isLoadingResults: boolean;
-  isRateLimitReached: boolean;
+  isLoadingResults: boolean = false;
```
mmalerba: nit: can remove `: boolean`

PR #6173, opened 2017-07-31
fix(table): column class names should be css friendly

andrewseguin:
Column classes are given to header and row cells. E.g. a column named `column-a` will give cells a classname `cdk-column-column-a`.

If the user has a column with special characters, they should be stripped so that CSS selectors cannot be invalid. 

This will provide a css class friendly name where it is only alphanumeric or includes special characters like - and _ (just because, if others should be included let me know)


src/cdk/table/cell.ts
```diff
@@ -40,6 +40,14 @@ export class CdkColumnDef {
 
   /** @docs-private */
   @ContentChild(CdkHeaderCellDef) headerCell: CdkHeaderCellDef;
+
+  /**
+   * Returns the column's name with non-alphanumeric characters removed so it can be used
+   * as part of a CSS class. Allows for the special characters - and _.
+   */
+  getCssClassFriendlyName() {
+    return this.name.replace(/[^a-z0-9_-]/ig, '');
```
mmalerba: I think you're doing more work than you need to by putting the logic here. I would do:

```ts
get name { return this._name; }
set name(value) {
  this._name = value;
  this.cssFriendlyName = (value | '').replace(/[^a-z0-9_-]/ig, '');
}
```

andrewseguin: The non-normalized name is still used by the `CdkTable` and should match even with the special charactesr. This change only should affect the CSS classes

jelbourn: Need to mention this replacement in the docs.

src/cdk/table/table.spec.ts
```diff
@@ -97,6 +98,16 @@ describe('CdkTable', () => {
     });
   });
 
+  it('should be able to apply class-friendly css class names for the column cells', () => {
+    const crazyColumnNameAppFixture = TestBed.createComponent(CrazyColumnNameCdkTableApp);
```
jelbourn: Add comment here like 
```ts
// Test component has a column named "crazy-column-NAME-1!@#$%^-_&*()2"
```

src/cdk/table/cell.ts
mmalerba: Right, I'm not proposing changing that, that's why I was saving it in `_name`. I just think it makes sense to only update the `cssFriendlyName` when the `name` gets set, rather than whenever change detection runs.

jelbourn: What's your reasoning for replacing with empty string instead of either `_` or `-`? Should include that in a comment here

andrewseguin: Ah whoops, totally missed that you were storing the value into a separate variable. Makes sense

andrewseguin: No reasoning, not sure of the standard here. Any preference if the invalid characters are replaced with `_` or `-`?

jelbourn: I see dashes more commonly in css

andrewseguin:
Test comment added; string is replaced by dashes; added styles section to cdk-table docs; using variable now instead of function

mcalmus:
Confirmed this is working in beta 10.

PR #6181, opened 2017-07-31
demo(input): add input a11y examples

mmalerba:


src/demo-app/a11y/input/input-a11y.html
```diff
@@ -0,0 +1,56 @@
+<section>
+  <h2>Basic input box (e.g. name field)</h2>
+  <md-input-container floatPlaceholder="never">
+    <input mdInput placeholder="First name" [(ngModel)]="firstName">
```
tinayuangao: add `aria-label`?

src/demo-app/a11y/input/input-a11y.html
```diff
@@ -0,0 +1,56 @@
...
+  </md-input-container>
+  =
+  <md-input-container floatPlaceholder="always">
+    <input mdInput type="number" placeholder="JPY" [(ngModel)]="jpy">
+    <span mdPrefix>‎¥‎</span>
+  </md-input-container>
+  (as of 7/31/2017)
+</section>
+
+<section>
```
tinayuangao: add header for this section?

src/demo-app/a11y/input/input-a11y.html
mmalerba: Not needed when you specify a placeholder since it gets placed inside of a `<label>` tag

mmalerba:
rebased and addressed comments

PR #6182, opened 2017-07-31
feat(table): add array-based datasource

andrewseguin:
Adds an implementation of DataSource that allows users to use `CdkTable` and `MdTable` without creating a new data source or going deeper in using observables. Triggers an update on the table when the data array is changed or when `refresh()` is explicitly called (for times when the data provided has objects added/removed/moved).


src/cdk/table/array-data-source.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  constructor(initialData: T[] = []) {
+    this._data = new BehaviorSubject<T[]>(initialData);
+  }
+
+  /**
+   * Triggers an update to the CollectionViewer. Should be used when elements are added, removed,
+   * or moved in the data array and should be reflected in the CollectionViewer.
+   */
+  update() { this._data.next(this.data); }
```
willshowell: It's confusing that some array manipulations should require this whereas others won't. Isn't it just re-emitting the last value -- one that should have already been emitted and rendered?

AFAICT (see comment below) the update method isn't necessary.

src/cdk/table/table.spec.ts
```diff
@@ -356,6 +358,62 @@ describe('CdkTable', () => {
...
+    arrayDataSource.data.push({a: 'a_4', b: 'b_4', c: 'c_4'});
+    arrayCdkTableFixture.detectChanges();  // Prove that detect changes won't render row
+    expectTableToMatchContent(tableElement, [
+      ['Column A', 'Column B', 'Column C'],
+      ['changed_a_1', 'b_1', 'c_1'],
+      ['a_2', 'b_2', 'c_2'],
+      ['a_3', 'b_3', 'c_3'],
+    ]);
+
+    arrayDataSource.update();  // Calling update will trigger update
```
willshowell: I removed this line and the test still passed. I'm afraid there is a bug in `expectTableToMatchContent`

src/cdk/table/array-data-source.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import {DataSource} from './data-source';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+
+/**
+ * Array-based data source that can be used to connect to a CollectionViewer. Will trigger an
+ * update to the CollectionViewer whenever a new data array is set.
+ */
+export class ArrayDataSource<T> implements DataSource<T> {
```
jelbourn: I'm wondering if we should move the data source into a separate cdk category from `table` since other components (mainly the tree) will use it as well.

@tinayuangao would you be able to use or extend this `ArrayDataSource` as-is for use in the tree?

I'm thinking we could move `DataSource`, `CollectionViewer` and probably the unrelated selection model into `cdk/collections`

jelbourn: Overall this isn't really what I had in mind for `ArrayDataSource`. I had been thinking that his class would capture all of the behavior for applying sorting, pagination, and potentially a substring search filter.

If we do this, the biggest thing to decide is how we access the row data for sorting / potentially filtering in a property-renaming compatible way.

src/cdk/table/table.spec.ts
andrewseguin: Good catch - the expectations only ran over the table's actual rows and didn't check to see if expected number of rows matched the table's actual number of rows. You should see now that the update function is needed.

src/cdk/table/array-data-source.ts
andrewseguin: This `update` function would only be necessary for users who want to set their data once and expect the table to always be based off the same array reference. For those who would say `arrayDataSource.data = newArraySlice` to update their data, they won't need this function.

I'm anticipating that this will be an easier concept for those who want to manipulate a single data array and see those changes propagated to the table. They'll wonder why they need to keep setting `data` to the same array just to see an update.

I'm open to your suggestions on whether or not this would be useful. Do you think people would be alright if we told them to do `arrayDataSource.data = theirData` each time they want to render an update?

src/cdk/table/array-data-source.ts
andrewseguin: Let's chat - I had in mind that we would create an `ArrayDataSource` and another more material-specific `MdTableDataSource` that would be able to accept an instance of MdSort and MdPaginator. I'm open to a different approach as well

src/cdk/table/array-data-source.ts
willshowell: I'm just a little worried about adding this caveat (even though I understand it's just how array references work). And I expect it will raise a bunch of issues like md-select's compare by reference has done.

Maybe one of these could work? The advantage (up for argument 😏) is that `update()` is required, so no matter what the user does with their array and `arrayDataSource.data`, they can expect that nothing will change until they call `update()`.

1. Users must supply the new data to `update()`
```ts
  private _data: BehaviorSubject<T[]>;
  get data() { return this._data.value; }

  constructor(initialData: T[] = []) {
    this._data = new BehaviorSubject<T[]>(initialData);
  }

  update(newData: T[]) { this._data.next(newData); }
```
2. Users can set/manipulate `data` all they want, but to render a change, they **must** call `update()`
```ts
  private _data: BehaviorSubject<T[]>;
  public data: T[];

  constructor(initialData: T[] = []) {
    this._data = new BehaviorSubject<T[]>(initialData);
  }

  update() { this._data.next(this.data); }
```

andrewseguin: I think both of those suggestions have merit, and I do like the idea of making it clear that data is updated through one path only. Keep in mind that changes made to the data itself (excluding adding/removing rows) will still automatically get picked up by change detection and that may confused users when they need to call `update`.

E.g. If you change the `name` property on the first object in the array, then that change will automatically be reflected in the table through standard view change detection without having to call `update()`. Check out the unit test to see that I can change `a_1` to `changeda_1` without calling `update()`.

@jelbourn's prefers we go the path of excluding the update function altogether and teaching people that they always need to set the data property to see things update.

Not sure what the best approach is, each has the confusing aspect that the table changes some times and other times needs a trigger. In the meantime, I'm going to add in sorting and pagination =P

willshowell: Yeah I guess the only way to get "pure" updates like I was imagining would be emitting deep copies of `data`. I like @jelbourn's option too, as long as the supporting docs make it clear!

PR #6189, opened 2017-08-01
perf(sidenav): avoid recalculating the inline styles while sidenav is open

### LINKED ISSUE #6205: Sidenav forces render each change detection
Sidenav container is calculating its styles each change detection to make sure that the content has the margin left styles matching the sidenav's width.

```
  /**
   * This is using [ngStyle] rather than separate [style...] properties because [style.transform]
   * doesn't seem to work right now.
   */
  _getStyles() {
    return {
      marginLeft: `${this._getMarginLeft()}px`,
      marginRight: `${this._getMarginRight()}px`,
      transform: `translate3d(${this._getPositionOffset()}px, 0, 0)`
    };
  }
```

crisbeto:
* Avoids recalculating the sidenav inline styles for every change detection cycle while the sidenav is open by doing it only when an animation starts.
* Removes a few one-liner methods that were being used in one place.

Fixes #6205

### DIFF
--- a/src/lib/sidenav/sidenav-container.html
+++ b/src/lib/sidenav/sidenav-container.html
@@ -3,6 +3,6 @@
 
 <ng-content select="md-sidenav, mat-sidenav"></ng-content>
 
-<div class="mat-sidenav-content" [ngStyle]="_getStyles()" cdk-scrollable>
+<div class="mat-sidenav-content" [ngStyle]="_styles" cdk-scrollable>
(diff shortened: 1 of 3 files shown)


jelbourn:
Nah, it was totes Hans

tinayuangao:
@crisbeto Could you please rebase? Thanks!

crisbeto:
Rebased @tinayuangao.

PR #6207, opened 2017-08-01
chore: don't hard-code cdk build order

jelbourn:
@DevVersion will using grep like this be a problem for Windows?


tools/package-tools/secondary-entry-points.ts
```diff
@@ -1,35 +1,82 @@
...
  *
  * @param packageName The package name for which to get entry points, e.g., 'cdk'.
  * @returns An array of secondary entry-points names, e.g., ['a11y', 'bidi', ...]
  */
 export function getSecondaryEntryPointsForPackage(packageName: string) {
-  // For now, we hard-code the fact that only the CDK has secondary entry-points until we switch
-  // to bazel.
-  if (packageName === 'cdk') {
-    return CDK_ENTRY_POINTS;
+  const packageDir = join(buildConfig.packagesDir, packageName);
```
jelbourn: Good point; now uses the `BuildPackage`

tools/package-tools/secondary-entry-points.ts
```diff
@@ -1,35 +1,82 @@
...
+
+  // First get the list of all entry-points as the list of directories in the package excluding
+  // blacklisted directories (e.g., "testing").
+  const entryPoints = readdirSync(packageDir)
+      .filter(f => lstatSync(join(packageDir, f)).isDirectory() && DIR_BLACKLIST.indexOf(f) < 0);
```
jelbourn: Had to add `es2016.array.include` to the tsconfig in `tools/gulp`.

tools/package-tools/secondary-entry-points.ts
```diff
@@ -1,35 +1,83 @@
...
+
+  // Update the deps for each node to point to the appropriate BuildNodes.
+  buildNodes.forEach(node => {
+    // Look for any imports that reference this same umbrella package and get the corresponding
+    // BuildNode for each by looking at the import statements with grep.
+    node.deps = spawnSync('egrep', [
+      '-roh',
+      '--include', '*.ts',
+      `from.'@angular/${packageName}/.+';`,
+      `./src/${basename(packageDir)}/${node.name}/`
```
DevVersion: Why not just `packageRoot` here?

jelbourn:
Updated

PR #6221, opened 2017-08-02
fix(list): properly align contents in subheader

### LINKED ISSUE #6214: Subheader line-height is too high
Proposal

screenshot 2017-08-02 10 15 57

screenshot 2017-08-02 10 14 44

The current line-height is 24px + 16*2px padding = 56px while the height of the element is only 48px. Resulting in the text being pushed down a bit. The line height should be 16px.

I also noticed that the subheader size is changed in dense mode while the spec does not say anything about this. I suggest keeping the size the same for dense lists as this only gives us 8px of height to fit a 12px font into.

DevVersion:
* The line-height of the subheaders is currently based on the typography level. This is problematic because the subheader is set to a specific height and can't grow/shrink accordingly.

Fixes #6214

### DIFF
--- a/src/lib/list/_list-theme.scss
+++ b/src/lib/list/_list-theme.scss
@@ -46,8 +46,9 @@
     }
 
     .mat-subheader {
-      font: mat-font-weight($config, body-2) mat-font-size($config, body-2)
-            mat-font-family($config, body-2);
+      font-family: mat-font-family($config, body-2);
+      font-size: mat-font-size($config, body-2);
+      font-weight: mat-font-weight($config, body-2);
     }
   }
 
@@ -59,7 +60,9 @@
     }
 
     .mat-subheader {
-      font: mat-font-weight($config, body-2) mat-font-size($config, caption) $font-family;
+      font-family: $font-family;
+      font-size: mat-font-size($config, caption);
+      font-weight: mat-font-weight($config, body-2);
     }
   }
 }
--- a/src/lib/list/_list-theme.scss
+++ b/src/lib/list/_list-theme.scss
@@ -46,7 +46,8 @@
     }
 
     .mat-subheader {
-      @include mat-typography-level-to-styles($config, body-2);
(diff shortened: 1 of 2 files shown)


src/lib/list/_list-theme.scss
```diff
@@ -46,7 +46,8 @@
     }
 
     .mat-subheader {
-      @include mat-typography-level-to-styles($config, body-2);
+      font: mat-font-weight($config, body-2) mat-font-size($config, body-2)
```
crisbeto: The `font` declaration is a little fiddly, especially when omitting values. I'd rather have this split into `font-weight`, `font-size` and `font-family`.

DevVersion: The `font` property is used in this file for the dense mode as well ([see here](https://github.com/DevVersion/material2/blame/110c54cab6616712fc8b8f91cff667af5f1799f8/src/lib/list/_list-theme.scss#L62)).

I'm fine changing it for both if desired.

crisbeto: Let's do that. Also the `font-family` could be set once on the `.mat-list` and the subheader can inherit from there instead.

DevVersion: Yeah makes sense. (I didn't write that before)

grizzm0:
What about the dense attribute I also pointed out? Should I create a separate issue for that? Right now the same thing would happen in dense mode as there's only 8px left for the text. The spec doesn't even mention a dense subheader in dense mode.

DevVersion:
@grizzm0 This fix should also calculate the proper line-height for subheader in dense mode.

grizzm0:
@DevVersion Yeah, but the height of subheader in dense mode is changed from 48px to 40px. (even tho the spec doesn't mention it).

The padding is still 2x16px = 32px. Meaning there's only 8px left for the text while the font-size in this PR is caption (12px). That's 4px overflow which the text will be "pushed down" in dense mode.

PR #6224, opened 2017-08-02
docs(datepicker): Add example for setLocale method

SargoDarya:
Add a usage example to show how the formatting locale of the datepicker can be be changed as that wasn't documented before.

### DIFF
--- a/src/lib/datepicker/datepicker.md
+++ b/src/lib/datepicker/datepicker.md
@@ -128,6 +128,24 @@ By default the datepicker will use the locale code from the `LOCALE_ID` injectio
 export class MyApp {}
 ```
 
+#### Setting/Switching the locale at runtime
+It's possible to set/switch the locale at runtime using the setLocale method from 
+the DateAdapter.
+
+```ts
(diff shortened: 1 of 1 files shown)


src/lib/datepicker/datepicker.md
```diff
@@ -128,6 +128,24 @@ By default the datepicker will use the locale code from the `LOCALE_ID` injectio
 export class MyApp {}
 ```
 
+#### Setting/Switching the locale at runtime
```
mmalerba: I think we can just leave this stuff as part of the previous heading

src/lib/datepicker/datepicker.md
```diff
@@ -128,6 +128,24 @@ By default the datepicker will use the locale code from the `LOCALE_ID` injectio
...
+It's possible to set/switch the locale at runtime using the setLocale method from
```
mmalerba: How about:

```md
It's also possible to set the locale at runtime using the `setLocale` method of the `DateAdapter`.
```

SargoDarya:
Changed it.

PR #6225, opened 2017-08-02
docs(getting-started): mention importing hammer in same file as bootstrap

cyrilletuzi:
As mentioned in #308, if hammerjs is imported in AppModule, Universal server-side rendering will fail as hammerjs is a browser specific library using window, document... So it must be imported in browser entry point src/main.ts instead.

### DIFF
--- a/guides/getting-started.md
+++ b/guides/getting-started.md
@@ -113,7 +113,7 @@ To install via npm, use the following command:
 npm install --save hammerjs
 ```
 
-After installing, import it on your app's root module.
+After installing, import it on your app's entry point (`src/main.ts`).
 ```ts
 import 'hammerjs';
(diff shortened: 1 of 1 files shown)


guides/getting-started.md
```diff
@@ -113,7 +113,7 @@ To install via npm, use the following command:
 npm install --save hammerjs
 ```
 
-After installing, import it on your app's root module.
+After installing, import it on your app's entry point (`src/main.ts`).
```
jelbourn: I'd add an `e.g., ` before the filename to make it more clear that it's an example and that not everyone will necessarily have this file.

cyrilletuzi: done

PR #6226, opened 2017-08-02
fix(autocomplete,datepicker,menu): closing parent dialog by pressing escape

### LINKED ISSUE #6223: Escaping out of menu/autocomplete within a dialog causes the dialog to close
Bug

#### Expected
Closing an autocomplete or menu from within a dialog via <kbd>esc</kbd> should only close the autocomplete panel / menu panel.

#### Current
Pressing <kbd>esc</kbd> closes the top dialog.

#### Reproduction

http://plnkr.co/edit/o9EQlYQKstYQdNsanheF?p=preview

crisbeto:
Prevents the autocomplete, datepicker and menu components from closing their parent dialog when the user presses escape.

Fixes #6223.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -298,6 +298,7 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   _handleKeydown(event: KeyboardEvent): void {
     if (event.keyCode === ESCAPE && this.panelOpen) {
       this.closePanel();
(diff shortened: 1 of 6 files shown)


crisbeto:
@jelbourn some food for thought: considering that interactions should be scoped to a particular overlay, perhaps we should `stopPropagation` on keyboard and mouse events on all overlays by default?

jelbourn:
@crisbeto my instinct is that always doing `stopPropagation` is overly broad, but don't have any concrete reasons why right now

PR #3341, opened 2017-02-28
feat(select): add ability to customize the select trigger

crisbeto:
Adds the `md-select-trigger` directive which allows users to customize the selected value. E.g. it is now possible to do something like this, if the user wanted to reverse the selected label for some reason:

```ts
<md-select placeholder="Food" [formControl]="control" #select="mdSelect">
  <md-select-trigger>
    {{ select.selected?.viewValue.split('').reverse().join('') }}
  </md-select-trigger>
  <md-option *ngFor="let food of foods" [value]="food.value">
    {{ food.viewValue }}
  </md-option>
</md-select>
```

Fixes #2275.

PR #6227, opened 2017-08-02
feat(select): custom text for multiple selected values

### LINKED ISSUE #6218: [Select] Custom text for multiple selected values
Feature request: Make it possible to adjust the text which is shown if multiple values are selected in MdSelect.

When all values are selected, the text is "[all] selected.
When more than 2 values are selected the text is "2 selected".

All values are shown, seperated by `,`, if they won't fit, ellipses are applied.

Go to the [plunk](http://plnkr.co/edit/texK2pTSyS4nIQWzNYB9?p=preview) and select all values.

To our customers it is more useful to see how many values are selected (whether or not all are selected), rather than seeing only the first part of the text of the first selected item.

[master](https://github.com/angular/material2)

I already tried extending MdSelect. I think this is the best option, but I got stuck on all the errors in the (copied) template :(

Another option would be to have a callback as input property which is evaluated as part of [triggerValue](https://github.com/angular/material2/blob/34ec068741186d471448745e54040ddc4e8fe81e/src/lib/select/select.ts#L530) upon changes ...

I can also work on a PR if this is something which is required by others as well and if you have some guidance on the best option ...

davidreher:
add input property to set a callback which will display a custom text if multiple values are selected. Similar to MdAutocomplete.displayWith.

fixes #6218

I am not sure about the naming, so any advice is welcome :)

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -2435,6 +2435,42 @@ describe('MdSelect', () => {
       expect(trigger.textContent).toContain('Tacos, Pizza, Steak');
     });
 
+    it('should display a custom text if multipleDisplayWith is set', () => {
+      testInstance.select.multipleDisplayWith = (values: string[]): string => {
+        return `${values.length} selected`;
+      };
+      trigger.click();
+      fixture.detectChanges();
+
+      const options = overlayContainerElement.querySelectorAll('md-option') as
+          NodeListOf<HTMLElement>;
+
+      options[0].click();
+      options[2].click();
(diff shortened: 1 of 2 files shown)


davidreher:
@kara, @chrisbeto: any idea on when this can be merged? This is very urgent for us ...

crisbeto:
There's https://github.com/angular/material2/pull/3341 that should do this already and should be a little more flexible. I'll see if I can get it reviewed.

davidreher:
ok, thx 👍

PR #6230, opened 2017-08-02
demo(slider): add slider a11y demos

mmalerba:


src/demo-app/a11y/slider/slider-a11y.html
```diff
@@ -0,0 +1,36 @@
+<section>
+  <h2>Continuous slider (e.g. color component sliders)</h2>
```
jelbourn: Add a `<p>` with something like `Select the red, blue, and green values for your color.` ?
(similar for other demos)

mmalerba:
comments addressed

PR #6239, opened 2017-08-02
fix(input): update aria-describedby to also include errors

### LINKED ISSUE #6131: Add aria-describedby for md-error and md-hint
We currently don't have anything that associates the content of `md-error` and `md-hint` with the input. We should generate IDs for these elements and point to them on the input's `aria-describedby`. 

We'll need to be sure not to clobber any existing describedby when adding/removing these IDs. 

We should also test out whether we should add `aria-hidden` to the `md-error` and `md-hint` elements so that they aren't repeated in an annoying way.

mmalerba:
fixes #6131

### DIFF
--- a/src/lib/input/input-container.spec.ts
+++ b/src/lib/input/input-container.spec.ts
@@ -799,17 +799,17 @@ describe('MdInputContainer with forms', () => {
       let hintId = fixture.debugElement.query(By.css('.mat-hint')).nativeElement.getAttribute('id');
       let describedBy = inputEl.getAttribute('aria-describedby');
 
-      expect(hintId).not.toBeFalsy('hint should be shown');
+      expect(hintId).toBeTruthy('hint should be shown');
       expect(describedBy).toBe(hintId);
 
       fixture.componentInstance.formControl.markAsTouched();
       fixture.detectChanges();
 
       let errorIds = fixture.debugElement.queryAll(By.css('.mat-input-error'))
-          .map((el) => el.nativeElement.getAttribute('id')).join(' ');
+          .map(el => el.nativeElement.getAttribute('id')).join(' ');
       describedBy = inputEl.getAttribute('aria-describedby');
 
-      expect(errorIds).not.toBeFalsy('errors should be shown');
+      expect(errorIds).toBeTruthy('errors should be shown');
       expect(describedBy).toBe(errorIds);
     });
   });
--- a/src/lib/input/input-container.spec.ts
+++ b/src/lib/input/input-container.spec.ts
@@ -794,6 +794,24 @@ describe('MdInputContainer with forms', () => {
           .toBe(1, 'Expected one hint to still be shown.');
       });
     }));
+
+    it('sets the aria-describedby to reference errors when in error state', () => {
+      let hintId = fixture.debugElement.query(By.css('.mat-hint')).nativeElement.getAttribute('id');
+      let describedBy = inputEl.getAttribute('aria-describedby');
+
+      expect(hintId).not.toBeFalsy('hint should be shown');
+      expect(describedBy).toBe(hintId);
+
+      fixture.componentInstance.formControl.markAsTouched();
+      fixture.detectChanges();
+
+      let errorIds = fixture.debugElement.queryAll(By.css('.mat-input-error'))
+          .map((el) => el.nativeElement.getAttribute('id')).join(' ');
+      describedBy = inputEl.getAttribute('aria-describedby');
+
+      expect(errorIds).not.toBeFalsy('errors should be shown');
+      expect(describedBy).toBe(errorIds);
+    });
   });
 
   describe('custom error behavior', () => {
--- a/src/lib/input/input-container.ts
+++ b/src/lib/input/input-container.ts
@@ -51,6 +51,7 @@ import {
   MD_ERROR_GLOBAL_OPTIONS
 } from '../core/error/error-options';
 import {Subject} from 'rxjs/Subject';
+import {startWith} from '@angular/cdk/rxjs';
 
 // Invalid input type. Using one of these will throw an MdInputContainerUnsupportedTypeError.
 const MD_INPUT_INVALID_TYPES = [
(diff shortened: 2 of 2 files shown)


src/lib/input/input-container.ts
```diff
@@ -566,19 +578,26 @@ export class MdInputContainer implements AfterViewInit, AfterContentInit, AfterC
...
+          ids.push(startHint.id);
+        } else if (this._hintLabel) {
+          ids.push(this._hintLabelId);
+        }
+
+        if (endHint) {
+          ids.push(endHint.id);
+        }
+      } else {
+        if (this._errorChildren) {
```
crisbeto: This one can be moved up and turned into an `else if`.

src/lib/input/input-container.spec.ts
```diff
@@ -794,6 +794,24 @@ describe('MdInputContainer with forms', () => {
...
+      let describedBy = inputEl.getAttribute('aria-describedby');
+
+      expect(hintId).not.toBeFalsy('hint should be shown');
+      expect(describedBy).toBe(hintId);
+
+      fixture.componentInstance.formControl.markAsTouched();
+      fixture.detectChanges();
+
+      let errorIds = fixture.debugElement.queryAll(By.css('.mat-input-error'))
+          .map((el) => el.nativeElement.getAttribute('id')).join(' ');
```
crisbeto: nit: you can remove the parentheses around `el`.

src/lib/input/input-container.spec.ts
```diff
@@ -794,6 +794,24 @@ describe('MdInputContainer with forms', () => {
           .toBe(1, 'Expected one hint to still be shown.');
       });
     }));
+
+    it('sets the aria-describedby to reference errors when in error state', () => {
+      let hintId = fixture.debugElement.query(By.css('.mat-hint')).nativeElement.getAttribute('id');
+      let describedBy = inputEl.getAttribute('aria-describedby');
+
+      expect(hintId).not.toBeFalsy('hint should be shown');
```
crisbeto: `toBeTruthy` instead of `.not.toBeFalsy`.

src/lib/input/input-container.spec.ts
```diff
@@ -794,6 +794,24 @@ describe('MdInputContainer with forms', () => {
...
+      expect(describedBy).toBe(hintId);
+
+      fixture.componentInstance.formControl.markAsTouched();
+      fixture.detectChanges();
+
+      let errorIds = fixture.debugElement.queryAll(By.css('.mat-input-error'))
+          .map((el) => el.nativeElement.getAttribute('id')).join(' ');
+      describedBy = inputEl.getAttribute('aria-describedby');
+
+      expect(errorIds).not.toBeFalsy('errors should be shown');
```
crisbeto: Also `toBeTruthy`.

src/lib/input/input-container.ts
```diff
@@ -566,19 +578,26 @@ export class MdInputContainer implements AfterViewInit, AfterContentInit, AfterC
...
-
-      if (startHint) {
-        ids.push(startHint.id);
-      } else if (this._hintLabel) {
-        ids.push(this._hintLabelId);
-      }
 
-      if (endHint) {
-        ids.push(endHint.id);
+      if (this._getDisplayedMessages() == 'hint') {
```
crisbeto: Triple equals for consistency.

src/lib/input/input-container.ts
```diff
@@ -566,19 +578,24 @@ export class MdInputContainer implements AfterViewInit, AfterContentInit, AfterC
...
+          ids.push(startHint.id);
+        } else if (this._hintLabel) {
+          ids.push(this._hintLabelId);
+        }
+
+        if (endHint) {
+          ids.push(endHint.id);
+        }
+      } else if (this._errorChildren) {
+        ids = this._errorChildren.map(mdError => mdError.id);
```
jelbourn: It looks like this always clobbers any existing `aria-describedby`. We want to augment an existing value if there already is one. @andrewseguin ran into the same thing w/ tooltip and was going to write some common code to handle this

src/lib/input/input-container.ts
```diff
@@ -566,19 +578,24 @@ export class MdInputContainer implements AfterViewInit, AfterContentInit, AfterC
...
-        ids.push(this._hintLabelId);
-      }
 
-      if (endHint) {
-        ids.push(endHint.id);
+      if (this._getDisplayedMessages() === 'hint') {
+        let startHint = this._hintChildren ?
+            this._hintChildren.find(hint => hint.align === 'start') : null;
+        let endHint = this._hintChildren ?
+            this._hintChildren.find(hint => hint.align === 'end') : null;
```
jelbourn: You could simplify a lot of this to
```ts
const ids = [...this._hintChildren.toArray(), ...this._errorChildren.toArray()].map(c => c.id);
```

mmalerba: The reason this is more complicated is because there's also a `hintLabel` property that can be used to set the start hint. The extra logic is to make sure we put the start hint first (whether its from a `md-hint` or `hintLabel`, followed by the end hint)

src/lib/input/input-container.ts
mmalerba: We could try to add to a static `aria-describedby`, but if the user has set up some binding like `[attr.aria-describedby]="myStuff"` I'm not sure there's much we can do. I'll sync with Andrew

src/lib/input/input-container.ts
```diff
@@ -100,10 +101,13 @@ export class MdHint {
 @Directive({
   selector: 'md-error, mat-error',
   host: {
-    'class': 'mat-input-error'
+    'class': 'mat-input-error',
```
mmalerba: This class has always been here, it's used for styling to avoid having to target both element selectors

src/lib/input/input-container.ts
jelbourn: We'd probably have to update after every `ngOnChanges`. This is good for now and we can revisit later

jelbourn:
Looks like some test failures on CI

mmalerba:
CI issues resolved

PR #6254, opened 2017-08-03
a11y: add autocomplete accessibility demo

crisbeto:
Adds a demo page for testing the autocomplete accessibility.


crisbeto:
For reference, here's the issues that I ran into while doing the assessment:
https://github.com/angular/material2/issues/6255
https://github.com/angular/material2/issues/6257

src/demo-app/a11y/autocomplete/autocomplete-a11y.html
```diff
@@ -0,0 +1,15 @@
+<section>
+  <h2>Filtering and selection</h2>
```
jelbourn: Add a paragraph under the header like
```html
<p>Select your favorite state</p>
```

PR #6262, opened 2017-08-04
Don't exclude min and max dates from datepicker

whitehooder:
Let the selected date be the equal to the minimum or maximum date since setting a limits to values shouldn't exclude them.

### DIFF
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -180,14 +180,14 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
   }
 
   /** The form control validator for the min date. */
-  private _minValidator: ValidatorFn = (control: AbstractControl): ValidationErrors | null => {
+  private _minValidator: ValidatorFn = (control: AbstractControl): ValidationErrors | null > {
     return (!this.min || !control.value ||
         this._dateAdapter.compareDate(this.min, control.value) <= 0) ?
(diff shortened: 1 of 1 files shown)


whitehooder:
I signed it!

mmalerba:
These are not `>=` signs that you're changing, they're ES6 fat-arrow functions `() => {}` the comparison logic for `min` and `max` did have an issue like this at one point, but it was fixed a while ago.

whitehooder:
My bad. Too quick on judging it, but I have noticed the min and max dates are excluded. I'll see if I can find where and submit another PR. Sorry for the inconvenience.

mmalerba:
@whitehooder are you using an old version of the library? There was a bug like this but its been fixed. Here's a plunker showing it working correctly: http://plnkr.co/edit/352BHLsD5emS7Fw8qd9H?p=preview

whitehooder:
I was using `2.0.0-beta.6`. Checking the changelog I found what you were saying and the [commit you were talking about](https://github.com/angular/material2/pull/4649/commits/d3f40224b182644c35e500a4d12e418b403e56c8).

I also realized that I changed the wrong lines. If you check [the commit](https://github.com/angular/material2/pull/6262/commits/2aecc0a70ef5b62cccf5403e682dcacba14a11ee) again you can see that the lines I was trying to remove the equal signs from are two lines directly below the lines I did change. Anyway I've looked over it again and realized that the `_minValidator` function should return `null` when there **isn't** an error and not the other way around.

After updating to `2.0.0-beta.8` the dates aren't excluded anymore.

Thanks for your help, @mmalerba.

PR #6274, opened 2017-08-04
fix(radio): forward focus to native input

### LINKED ISSUE #6029: Focus-trap start marker doesn't work when focus is redirected to an internal element
If `cdk-focus-start` / `-end` is applied to, e.g., `<md-radio-button>`, the focus trap will try to focus the actual `md-radio-button` element. This is problematic because the radio button actually defers focus/keyboard interaction to an internal `<input type="radio">` element.

mmalerba:
fixes #6029

### DIFF
--- a/src/lib/radio/radio.ts
+++ b/src/lib/radio/radio.ts
@@ -332,6 +332,9 @@ export const _MdRadioButtonMixinBase = mixinColor(mixinDisableRipple(MdRadioButt
     '[class.mat-radio-checked]': 'checked',
     '[class.mat-radio-disabled]': 'disabled',
     '[attr.id]': 'id',
+    // Note: under normal conditions focus shouldn't land on this element, however it may be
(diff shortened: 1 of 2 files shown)


src/lib/radio/radio.ts
```diff
@@ -332,6 +332,7 @@ export const _MdRadioButtonMixinBase = mixinColor(mixinDisableRipple(MdRadioButt
     '[class.mat-radio-checked]': 'checked',
     '[class.mat-radio-disabled]': 'disabled',
     '[attr.id]': 'id',
+    '(focus)': '_inputElement.nativeElement.focus()',
```
DevVersion: I think it would be really beneficial to have a comment for this. What do you think?

A comment that explains that normally the native element should be focused; But for the focus-trap there is an exception

jelbourn:
@tinayuangao @DevVersion any idea what's up w/ the screenshots?

PR #6279, opened 2017-08-04
feat(focus-trap): return whether shifting focus was successful

crisbeto:
Adds return types to the focus trap methods, allowing the consumer to react based on whether the focus trap managed to find a focusable element. This is useful for cases like the dialog where focus could be left behind if it is a purely text-based dialog.


crisbeto:
@jelbourn @mmalerba for the async methods, I ended up resolving with a boolean instead of resolving/rejecting. It seems like the `async` zone starts acting up if we have a rejected promise that isn't handled.

PR #6281, opened 2017-08-04
refactor(slider): remove redundant renderer class

crisbeto:
Gets rid of the `SliderRenderer` class which doesn't really do much, apart from holding a couple of methods and abstracting away some logic that doesn't really need to be abstracted. The methods have been moved into the slider itself.


tinayuangao:
@crisbeto Please rebase. Thanks!

Issue #5709, opened 2017-07-12
bug(autocomplete): input underline goes under the autocomplete container


m98:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
The [autocomplete](https://material.angular.io/components/autocomplete/overview) is a normal text input and it should follow inputs style and behavior.

> Text fields should indicate their state – whether enabled or disabled, empty or filled, valid or invalid – with clear label, input, and assistive text. 
> The input line indicates where to enter text, displayed below the label. When a field is active or contains an error, the line’s color and thickness vary. ([source](https://material.io/guidelines/components/text-fields.html#text-fields-layout))
> 
I expect when the autocomplete input is active I can see the input underline (aka input line)

#### What is the current behavior?
The underline (input line) goes under the autocomplete container. The image bellow shows the current behavior:
![Current behavior](https://i.stack.imgur.com/0lzrU.png)

https://material.angular.io/components/autocomplete/examples

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
```
@angular/cli: 1.1.1
node: 6.10.3
os: linux x64
@angular/animations: 4.1.3
@angular/common: 4.1.3
@angular/compiler: 4.1.3
@angular/core: 4.1.3
@angular/forms: 4.1.3
@angular/http: 4.1.3
@angular/material: 2.0.0-beta.7
@angular/platform-browser: 4.1.3
@angular/platform-browser-dynamic: 4.1.3
@angular/router: 4.1.3
@angular/cli: 1.1.1
@angular/compiler-cli: 4.1.3
@angular/language-service: 4.1.3

```

donroyco:
Are you sure this is the behaviour in more than just one browser?

m98:
I used two  different computers, and both had latest version of chrome on it.

donroyco:
I'm not talking about other computer. I'm talking about different _browsers_, like Internet Explorer 11, Microsoft Edge, Mozilla Firefox and Apple Safari. Does it occur in these browsers as well?

m98:
I tried with the two different computers, with different screen size and resolutions.

Could you please try the autocomplete using IE, FireFox and Safari and provide screenshot of their behavior?
Thanks

julianobrasil:
I have some browsers installed over here.

### IE11:

![image](https://user-images.githubusercontent.com/25405143/28176949-92594626-67cf-11e7-9136-d1f3b0a98948.png)

### Edge:

[...]

Issue #4914, opened 2017-06-01
Autocomplete panel should base width on input-container, not input


willshowell:
#### Bug, feature request, or proposal:
Bug

#### Expected
The autocomplete panel does not change its width when the input resizes, only when the input container does.

#### Current
The autocomplete panel resizes when an `mdSuffix` is rendered, thus changing the size of the actual `input`.

#### Reproduction
https://plnkr.co/edit/VrcrcUBKbwuoOoxc8byM?p=preview

#### Motivation
Using a "clear input" button on an autocomplete input

#### Proposed change
```diff
/** Returns the width of the input element, so the panel width can match it. */
private _getHostWidth(): number {
-  return this._element.nativeElement.getBoundingClientRect().width;
+  return this._inputContainer
+    ? this._inputContainer._elementRef.nativeElement.getBoundingClientRect().width
+    : this._element.nativeElement.getBoundingClientRect().width;
}

crisbeto:
I think this needs to be thought through a bit better. Apart from taking the width from the container, the autocomplete would also have to position itself relative to it, otherwise it'll be off if there's a prefix. This complicates things, because the container has some padding at the bottom, which will cause the panel to float ~10px below where it should be.

willshowell:
@crisbeto what if it were instead attached to the container's `underlineRef` like [datepicker is doing](https://github.com/angular/material2/blob/master/src/lib/datepicker/datepicker-input.ts#L196-L198)

crisbeto:
That should work, I wasn't aware that we were even exporting that reference. I'll update my PR.

crisbeto:
Thinking about it again, the underline isn't ideal either: if we use it as a reference, the panel will end up overlapping the input when it goes into the `above` position.

willshowell:
Oh true. You could do what autocomplete is doing with its negative margin when `above`, but I realize that's not an ideal solution.

crisbeto:
Not really, we shouldn't know or have to measure, the height of the input.

willshowell:
Do you suppose the position strategy could be more flexible to support something like

* connect to `bottom` of `underlineRef`
* fallback to `top` of `inputContainer`

```ts
.connectedTo(
    this._inputContainer.underlineRef,
    {originX: 'start', originY: 'bottom'}, {overlayX: 'start', overlayY: 'top'})
  .withFallbackPosition(
    this._inputContainer,
[...]

PR #6282, opened 2017-08-04
fix(autocomplete): attach overlay to a more accurate input element

willshowell:
Fixes https://github.com/angular/material2/issues/4914
Fixes https://github.com/angular/material2/issues/5709

I removed the `.mat-autocomplete-panel-below` and `.mat-autocomplete-panel-above` classes since they were only being used for css offsets, but maybe they should stay if users have been querying them?

Additionally, it looks like `positionY: AutocompletePositionY` and `_subscribeToPositionChanges` are no longer needed internally, but thought I should check first before removing parts of the public API .


src/lib/input/input-container.ts
```diff
@@ -457,6 +457,7 @@ export class MdInputContainer implements AfterViewInit, AfterContentInit, AfterC
 
   /** Reference to the input's underline element. */
   @ViewChild('underline') underlineRef: ElementRef;
+  @ViewChild('connectionContainer') _connectionContainerRef: ElementRef;
```
mmalerba: Can you not use the `underlineRef`? The datepicker is using it for a similar purpose.

willshowell: https://github.com/angular/material2/issues/4914#issuecomment-305617121 You still have to add an offset for when the fallback is above

mmalerba:
Also for the no-longer-used bits of the API, lets get rid of them and if we encounter too many problems trying to merge it into Google's codebase we can add them back.

willshowell:
@crisbeto weird... it looks like the 6px/24px offsets are still there. Is that Edge? Chrome 59 macOS shows this:

screen shot 2017-08-07 at 9 55 34 am

screen shot 2017-08-07 at 9 55 43 am

crisbeto:
I looked at it in Chrome 59 or 60 on Windows.

willshowell:
Just checked on Windows and it looked the same to me. Also deployed to https://willshowell-demo.firebaseapp.com/autocomplete

Also note that half of the 2px focused underline is obstructed when panel is below because the underline and ripple are absolutely positioned. I can add back a 1px offset or leave as is and later look into possible underline changes that would fix it.

willshowell:
Could you remove the `merge-ready` label? I'd like to remove a couple more unused lines supporting the old approach, per @mmalerba's suggestion.

willshowell:
Thanks @crisbeto, change made

PR #6285, opened 2017-08-05
build: fix docs generation highlight files task

DevVersion:
Recently with https://github.com/angular/material2/commit/edcbb24f11ae776b94d97634ae91e9175ffd6f8a the ambigious variable `path` has been renamed to `filePath`, but there is one remaining instance of `path` that didn't get updated.

This currently causes the docs-content CI job to fail.


DevVersion:
@jelbourn Yeah. Maybe we find a lint rule for it. Looking into it.

jelbourn:
The reason it was changed was to satisfy the lint rule after making lint apply to more files :)

Issue #6257, opened 2017-08-03
Autocomplete not usable in high contrast mode on Windows


crisbeto:
The autocomplete has a few issues which prevent high-contrast users from being able to use it:
1. The input bottom border isn't visible, which makes it impossible to figure out that component is an input.
2. The Windows high contrast settings render out all of the borders. This causes a weird thick border above the input which is only used for spacing and is usually transparent.
3. There's no visible indication with regards to what option is highlighted.
4. There is no outline around the autocomplete panel.

A couple of screenshots for reference:
![1](https://user-images.githubusercontent.com/4450522/28940371-d76d7342-7894-11e7-9f4e-f7d2de19d2a9.png)
![2](https://user-images.githubusercontent.com/4450522/28940375-daadc4b2-7894-11e7-9746-635f37a5e6da.png)

PR #6291, opened 2017-08-05
fix(input): not working correctly in high contrast mode

crisbeto:
Fixes the input not working correctly in high contrast mode by:
* Using a margin instead of a border to do the spacing inside the input. Windows renders all borders in high contrast mode, even if they're transparent, which caused the input to have a 5px top border.
* Using a border to render the underline, instead of a background color.
* Hiding the ripple, because it was causing the underline to blend in with the background.

Also gets rid of an unused method.

Relates to #6257.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -58,10 +58,11 @@
   }
 
   .mat-input-underline {
-    background-color: $input-underline-color;
+    color: $input-underline-color;
 
     &.mat-disabled {
       @include mat-control-disabled-underline($input-underline-color);
+      color: transparent;
     }
   }
 
@@ -182,8 +183,7 @@
 
   .mat-input-infix {
     padding: $infix-padding 0;
-    // Throws off the baseline if we do it as a real margin, so we do it as a border instead.
-    border-top: $infix-margin-top solid transparent;
+    margin-top: $infix-margin-top;
   }
 
   .mat-input-element {
--- a/src/lib/input/input-container.scss
+++ b/src/lib/input/input-container.scss
@@ -1,5 +1,6 @@
 @import '../core/style/variables';
 @import '../core/style/vendor-prefixes';
+@import '../core/a11y/a11y';
 
 
 // Min amount of space between start and end hint.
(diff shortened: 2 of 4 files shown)


crisbeto:
I'm not sure that it really threw anything off. There's ~1px difference (most likely because of the border changes), but if anything it got the input and select to align, in master there's a 1px offset.

Here are a couple of comparisons, the first one is with these changes, the second one is master:

mmalerba:
Don't worry about the select, it's the odd one out and going to change very soon. This change causes the `T` in `Text Input` to be misaligned wrt the `T` in `Text 4`.
This PR:
this pr
Master:
master

jelbourn:
@crisbeto @mmalerba is this worth revisiting?

crisbeto:
@jelbourn it definitely is, but we haven't been able to find a way to handle this without throwing off the baseline slightly. We may end up having to have a slightly-off baseline in high contrast mode.

jelbourn:
I'll leave this to you and @mmalerba, then

mmalerba:
I'll try playing with this after my form field variant changes and see if those changes help with this at all

crisbeto:
@mmalerba I did a quick check by converting the border to a margin (I didn't do that other border change that's in this PR) and it still seems like the baseline is still being thrown off.

mmalerba:
@crisbeto interesting, I have both versions up side-by-side and I see no difference flipping between them. We're both talking about this border right? https://github.com/angular/material2/blob/master/src/lib/form-field/_form-field-theme.scss#L175

crisbeto:
Here's what I've got. The top one is master and the bottom is only the border being changed to a magin. A couple of things I noticed though:
1. I'm not sure where that scrollbar on the textarea is coming from.
2. The screenshots looked identical if I change the margin through the dev tools. They ended up being slightly different if I change it in the source.

mmalerba:
That must be the difference, I just changed it in the devtools. I'll see if I can switch it to margin or padding on a different element when I get a chance

crisbeto:
Closing in favor of #11736.

PR #6295, opened 2017-08-05
Add complex tabs example

amcdnl:
Replaces the 'Coming Soon!' example with an advanced tab example showing:

- Different content sizes
- Tab HTML Labels
- Disabled Tabs
- Tab Header Overflow


amcdnl:
@jelbourn - can this be merged?

PR #6298, opened 2017-08-05
fix(tooltip): update tooltip template to allow HTML

carlcheel:
There is often a need to add HTML to tooltips, specifically line breaks.

### DIFF
--- a/src/lib/tooltip/tooltip.html
+++ b/src/lib/tooltip/tooltip.html
@@ -1,7 +1,7 @@
 <div class="mat-tooltip"
      [ngClass]="tooltipClass"
+     [innerHtml]="message"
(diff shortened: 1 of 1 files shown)


jelbourn:
We intentionally do not support HTML content for tooltips, as they're meant to be simple messages.

Also, this is XSS city.

carlcheel:
@jelbourn
What would you suggest if a user would like to have one word bold in a tooltip, use a different library?

Also XSS is only an issue if you are using user input to build a tooltip and even then have decided to use unsanitised input.

To mitigate any potential mistakes, the html property could be disabled by default and you could toggle this.

zmanring:
Any movement on this? I agree but our client needs HTML in the tooltip :/

PR #6302, opened 2017-08-06
fix(select,autocomplete): darken selected option

### LINKED ISSUE #6229: [option/menu] hover and active background colors
Styling issue

#### Current
* the spec appears to show active option backgrounds at 12% opacity but currently they are 4% 
* active options have the same background opacity as hovered options

#### Expected
* Active option backgrounds should be 12% opacity and either keep 4% for hover or remove it entirely

This shows how hovered options and active actions aren't distinguishable. Before hitting <kbd>enter</kbd>, I am not sure whether Hawaii or Indiana will be chosen.

autocomplete-option-hover

The following show the difference between the active select and the spec
screen shot 2017-08-02 at 2 13 16 pm
screen shot 2017-08-02 at 2 13 41 pm

crisbeto:
Darkens the selected option background in order to distinguish it from the ones that are focused or hovered.

Fixes #6229.

### DIFF
--- a/src/lib/core/option/_option-theme.scss
+++ b/src/lib/core/option/_option-theme.scss
@@ -30,7 +30,7 @@
 
     // In multiple mode there is a checkbox to show that the option is selected.
     &.mat-selected:not(.mat-option-multiple) {
-      background: mat-color($background, hover);
+      background: mat-color($background, hover, 0.12);
     }
 
     &.mat-active {


src/lib/core/option/_option-theme.scss
```diff
@@ -30,7 +30,7 @@
 
     // In multiple mode there is a checkbox to show that the option is selected.
     &.mat-selected:not(.mat-option-multiple) {
-      background: mat-color($background, hover);
+      background: rgba(mat-color($background, hover), 0.12);
```
jelbourn: fyi `mat-color` should take opacity as a third argument, but it's the same either way

crisbeto: I tried that initially and I kept getting some weird SASS error.

crisbeto: Weird, it works now. It must've been something else.

Issue #6303, opened 2017-08-06
Tab Header Arrows Not Showing on Resize


amcdnl:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
When the browser resizes and the tabs are cut off, the arrows would appear and vice versa.

#### What is the current behavior?
Currently when the page is resize, the arrows are not shown until 
https://www.screencast.com/t/9MWHdRYsfpIv

#### What are the steps to reproduce?
1. Open a large set of tabs but doesn't spill over the screen.
2. Resize your screen so the tabs are cut off.
3. Note the arrows to show the overflown content are NOT visible
4. Click a tab and they appear

PR #6304, opened 2017-08-06
bug(tabs): fix arrows not showing on reszize #6303

amcdnl:
Addresses https://github.com/angular/material2/issues/6303

### DIFF
--- a/src/lib/tabs/tab-header.spec.ts
+++ b/src/lib/tabs/tab-header.spec.ts
@@ -1,4 +1,6 @@
-import {async, ComponentFixture, TestBed, fakeAsync, tick} from '@angular/core/testing';
+import {
+  async, ComponentFixture, TestBed, fakeAsync, tick, discardPeriodicTasks
+} from '@angular/core/testing';
 import {Component, ViewChild, ViewContainerRef} from '@angular/core';
 import {Direction, Directionality} from '../core/bidi/index';
 import {MdTabHeader} from './tab-header';
@@ -261,6 +263,22 @@ describe('MdTabHeader', () => {
       fixture.detectChanges();
 
       expect(inkBar.alignToElement).toHaveBeenCalled();
+      discardPeriodicTasks();
+    }));
+
+    it('should update arrows when the window is resized', fakeAsync(() => {
+      fixture = TestBed.createComponent(SimpleTabHeaderApp);
+
+      const header = fixture.componentInstance.mdTabHeader;
+
+      spyOn(header, '_checkPaginationEnabled');
+
+      dispatchFakeEvent(window, 'resize');
+      tick(10);
+      fixture.detectChanges();
+
+      expect(header._checkPaginationEnabled).toHaveBeenCalled();
+      discardPeriodicTasks();
     }));
 
   });
--- a/src/lib/tabs/tab-header.ts
+++ b/src/lib/tabs/tab-header.ts
@@ -41,6 +41,7 @@ import {merge} from 'rxjs/observable/merge';
 import {fromEvent} from 'rxjs/observable/fromEvent';
 import {CanDisableRipple, mixinDisableRipple} from '../core/common-behaviors/disable-ripple';
 import {RxChain, debounceTime} from '@angular/cdk/rxjs';
+import {Platform} from '@angular/cdk/platform';
 
 /**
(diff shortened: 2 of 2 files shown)


src/lib/tabs/tab-header.ts
```diff
@@ -387,6 +388,7 @@ export class MdTabHeader extends _MdTabHeaderMixinBase
    * This is an expensive call that forces a layout reflow to compute box and scroll metrics and
    * should be called sparingly.
    */
+  @HostListener('window:resize')
```
crisbeto: We don't really use the `@HostListener` anywhere. Can you switch to using the `host` property on the component definition instead>

amcdnl: Sure!

src/lib/tabs/tab-header.ts
```diff
@@ -79,6 +80,7 @@ export const _MdTabHeaderMixinBase = mixinDisableRipple(MdTabHeaderBase);
     'class': 'mat-tab-header',
     '[class.mat-tab-header-pagination-controls-enabled]': '_showPaginationControls',
     '[class.mat-tab-header-rtl]': "_getLayoutDirection() == 'rtl'",
+    '(window:resize)': "_checkPaginationEnabled()"
```
jelbourn: At the very least this call should be debounced. The comment for the handler even says "This is an expensive call that forces a layout reflow to compute box and scroll metrics and should be called sparingly."

jelbourn: We should really just have one library-wide resize handler instead of each component adding their own. Add a TODO?

jelbourn: Any way to add a unit test for this?

amcdnl:
@jelbourn - Ready for another review.

src/lib/tabs/tab-header.spec.ts
```diff
@@ -261,6 +263,22 @@ describe('MdTabHeader', () => {
...
+
+      const header = fixture.componentInstance.mdTabHeader;
+
+      spyOn(header, '_checkPaginationEnabled');
+
+      dispatchFakeEvent(window, 'resize');
+      tick(10);
+      fixture.detectChanges();
+
+      expect(header._checkPaginationEnabled).toHaveBeenCalled();
```
jelbourn: FYI, I generally prefer tests that assert on the outcome of calling the function rather than using a spy. In this case, though, it would end up being a brittle geometric value so the spy is okay.

src/lib/tabs/tab-header.ts
```diff
@@ -143,6 +146,11 @@ export class MdTabHeader extends _MdTabHeaderMixinBase
               private _changeDetectorRef: ChangeDetectorRef,
               @Optional() private _dir: Directionality) {
     super();
+
+    // TODO: Add library level window listener https://goo.gl/FJWhZM
```
jelbourn: I think you got the wrong comment in your shortlink
(probably meant https://github.com/angular/material2/pull/6304#discussion_r131703706)

src/lib/tabs/tab-header.ts
```diff
@@ -78,7 +78,7 @@ export const _MdTabHeaderMixinBase = mixinDisableRipple(MdTabHeaderBase);
   host: {
     'class': 'mat-tab-header',
     '[class.mat-tab-header-pagination-controls-enabled]': '_showPaginationControls',
-    '[class.mat-tab-header-rtl]': "_getLayoutDirection() == 'rtl'",
+    '[class.mat-tab-header-rtl]': "_getLayoutDirection() == 'rtl'"
```
jelbourn: TrailingCommas4Life

jelbourn:
Actually I just noticed that this breaks the prerender task. The resize event should only be setup when on the browser platform

src/lib/tabs/tab-header.ts
amcdnl: Are you serial?!

src/lib/tabs/tab-header.spec.ts
amcdnl: Ya, thats what I was thinking too.

amcdnl:
@jelbourn - ready for another review.

PR #6313, opened 2017-08-07
fix(xxx-intl): replace misused EventEmitter with Subject

e-cloud:
According to https://angular.io/api/core/EventEmitter#description, EventEmitter is not for services. And Subject is preferrable.

### DIFF
--- a/src/lib/datepicker/datepicker-intl.ts
+++ b/src/lib/datepicker/datepicker-intl.ts
@@ -7,7 +7,7 @@
  */
 
 import {Injectable} from '@angular/core';
-import { Subject } from 'rxjs/Subject';
+import {Subject} from 'rxjs/Subject';
 
 
 /** Datepicker data that requires internationalization. */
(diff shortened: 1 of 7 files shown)


src/lib/datepicker/datepicker-intl.ts
```diff
@@ -6,7 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Injectable, EventEmitter} from '@angular/core';
+import {Injectable} from '@angular/core';
+import { Subject } from 'rxjs/Subject';
```
willshowell: As a heads up, they'll want you to remove these spaces. Here and below.

e-cloud: that's ok. however, it's better to apply a tool like prettier that format the code at pre-commit. That would be more efficient on code style management.

e-cloud:
The tests should be fine, there is something wrong with travis.

PR #6317, opened 2017-08-07
ci: shorter wait timeout for saucelabs and browserstack

DevVersion:
Changes the wait timeout for Saucelabs and Browserstack from 2 minutes to 1 minute. 

This is now possible because if the timeout fails, we still re-try to create a connection a second time.


jelbourn:
Why not bring it down to even 30 seconds? That seems pretty long to just establish a connection

jelbourn:
Eh, we'll see how it goes.

DevVersion:
@jelbourn I agree. Let's see. Maybe 30 seconds and 2 retries. (Can be somewhat in the future)

PR #6323, opened 2017-08-07
feat(datepicker): close calendar after choose the same date again

hwesol13:
Selecting already selected date didn't close datepicker. This contains functionality to close datepicker even when already selected date is picked.


mmalerba:
Though this may be the easiest way to get the functionality you want, I don't think it's really correct to emit a `selectedChange` event when the selected date has not actually changed. Instead I would separate the responsibilities into two separate observables `selectedChange` would just responsible for updating the value, not closing the calendar. And the new observable (maybe call it `userSelection` or something) would be responsible for closing the calendar regardless of whether its the same value.

hwesol13:
Nice one. Thx for advice.

src/lib/datepicker/datepicker-content.html
```diff
@@ -6,5 +6,6 @@
     [maxDate]="datepicker._maxDate"
     [dateFilter]="datepicker._dateFilter"
     [selected]="datepicker._selected"
-    (selectedChange)="datepicker._selectAndClose($event)">
+    (selectedChange)="datepicker._selectAndClose($event)"
```
willshowell: This `_selectAndClose` method and its comment should be updated since it no longer closes

src/lib/datepicker/calendar.html
```diff
@@ -48,7 +48,9 @@
       [activeDate]="_activeDate"
       [selected]="selected"
       [dateFilter]="_dateFilterForViews"
-      (selectedChange)="_dateSelected($event)">
+      (selectedChange)="_dateSelected($event)"
+      (userSelection)="_userSelected()"
+  >
```
willshowell: Closing bracket should be on line above

src/lib/datepicker/datepicker-input.ts
```diff
@@ -224,13 +224,13 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
...
-            this.dateInput.emit(new MdDatepickerInputEvent(this, this._elementRef.nativeElement));
-            this.dateChange.emit(new MdDatepickerInputEvent(this, this._elementRef.nativeElement));
-          });
+        this._datepicker.selectedChanged.subscribe((selected: D) => {
+          this.value = selected;
+          this._cvaOnChange(selected);
+          this._onTouched();
+          this.dateInput.emit(new MdDatepickerInputEvent(this, this._elementRef.nativeElement));
+          this.dateChange.emit(new MdDatepickerInputEvent(this, this._elementRef.nativeElement));
+        });
```
willshowell: I think they actually prefer line breaks like this to stay 4 spaces

src/lib/datepicker/month-view.ts
```diff
@@ -67,6 +67,9 @@ export class MdMonthView<D> implements AfterContentInit {
   /** Emits when a new date is selected. */
   @Output() selectedChange = new EventEmitter<D | null>();
 
+  /** Emits when date is picked */
+  @Output() userSelection = new EventEmitter();
```
willshowell: The comment you had above seems more appropriate for this as well:

```
/** Emits when any date is selected. */
```

Nit: Add a period to both to match other comments

willshowell: Also I think it should have a \<void\> type?

```ts
new EventEmitter<void>();
```

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -190,6 +190,26 @@ describe('MdDatepicker', () => {
         });
       });
 
+      it('setting twice to the same selected value should update input and close calendar', () => {
```
mmalerba: 'clicking the currently selected date should close the calendar without firing selectedChanged'

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -190,6 +190,26 @@ describe('MdDatepicker', () => {
...
+        for (let changeCount = 1; changeCount < 3; changeCount++) {
+          const currentDay = changeCount;
+          testComponent.datepicker.open();
+          fixture.detectChanges();
+
+          expect(document.querySelector('md-datepicker-content')).not.toBeNull();
+          expect(testComponent.datepickerInput.value).toEqual(new Date(2020, JAN, currentDay));
+
+          let cells = document.querySelectorAll('.mat-calendar-body-cell');
+          dispatchMouseEvent(cells[1], 'click');
```
mmalerba: verify that `selectedChanged` only emits the first time

src/lib/datepicker/datepicker.ts
```diff
@@ -224,13 +224,12 @@ export class MdDatepicker<D> implements OnDestroy {
   }
 
   /** Selects the given date and closes the currently open popup or dialog. */
```
willshowell: Nit: can you update this comment too?

src/lib/datepicker/datepicker-input.ts
```diff
@@ -224,7 +224,7 @@ export class MdDatepickerInput<D> implements AfterContentInit, ControlValueAcces
   ngAfterContentInit() {
     if (this._datepicker) {
       this._datepickerSubscription =
-          this._datepicker.selectedChanged.subscribe((selected: D) => {
+        this._datepicker.selectedChanged.subscribe((selected: D) => {
```
willshowell: Nit: this should stay the same too. They indent 2 for function blocks and 4 for line breaks

src/lib/datepicker/datepicker.spec.ts
```diff
@@ -190,6 +190,30 @@ describe('MdDatepicker', () => {
         });
       });
 
+      it('clicking the currently selected date should close the calendar' +
```
mmalerba: nit: space before close of string so result doesn't read: "calendarwithout"

src/lib/datepicker/datepicker.ts
hwesol13: Ohhh... sorry, i missed it

mmalerba:
great, and we can just ignore that screenshot test failure, it's just because the blinking input caret

PR #6324, opened 2017-08-07
chore(): add sort module to universal app

DevVersion:
* Adds the `MdSortModule` to the universal app. This means that it will be part of the prerender check on the CI.

**Note**: It makes sense to check the `mdSort` module because it includes components and directives that may cause problems in the server platform.


mmalerba:
please rebase

DevVersion:
@mmalerba Done.

PR #6326, opened 2017-08-07
feat(overlay): support setting multiple panel classes

### LINKED ISSUE #6318: Support multiple classes in MdDialog's panelClass
Feature request

I want MdDialogConfig to have option for panelClass to be an array of classes, like
`panelClass: ['custom-container', 'widget-container']`

Currently there could be only one panelClass per dialog

I want to add various classes depending on environment, but i have no such option  and i have to include all possible combinations into separate classes

crisbeto:
Allows for more than one class to be added via the `panelClass` option.

Fixes #6318.

### DIFF
--- a/src/lib/core/overlay/overlay-ref.ts
+++ b/src/lib/core/overlay/overlay-ref.ts
@@ -60,7 +60,12 @@ export class OverlayRef implements PortalHost {
     }
 
     if (this._state.panelClass) {
-      this._pane.classList.add(this._state.panelClass);
+      // We can't do a spread here, because IE doesn't support setting multiple classes.
+      if (Array.isArray(this._state.panelClass)) {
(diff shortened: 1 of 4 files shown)


src/lib/core/overlay/overlay-ref.ts
```diff
@@ -60,7 +60,12 @@ export class OverlayRef implements PortalHost {
     }
 
     if (this._state.panelClass) {
-      this._pane.classList.add(this._state.panelClass);
+      // We can't do a spread here, because IE doesn't support setting multiple classes.
+      if (Array.isArray(this._state.panelClass)) {
```
jelbourn: I had no idea this function existed

Komald:
Still not able to pass multiple panel classes to Mat Dialog. 
Version using: angular/material@6.4.7.
From which verion this feature is available?

PR #6331, opened 2017-08-07
refactor(input): rename `input-container` to `form-field`...

mmalerba:
and move all communication with `mdInput` into a well defined interface. This is a pre-requisite to making `md-select` share a container element with `mdInput`.

This is a first-pass PR that tries to avoid breaking changes as much as possible. We can update Google's codebase and remove the old selectors and class names in the future.

@tinayuangao This should also allow you to integrate `md-chip-list` with the input more easily.

Incremental step toward #2124

BREAKING CHANGE: `MdInputContainer` is now `MdFormField` and `MdInputDirective` is now `MdInput`


mmalerba:
here's a demo with the changes: https://mmalerba-demo2.firebaseapp.com/, basically nothing should change from a user's perspective

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -410,8 +411,9 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
 
     // If it's used in a Material container, we should set it through
     // the property so it can go through the change detection.
-    if (this._inputContainer) {
-      this._inputContainer._mdInputChild.value = inputValue;
+    if (this._inputContainer &&
```
jelbourn: Rename `_inputContainer` to `_formField`?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -410,8 +411,9 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
+        this._inputContainer._control instanceof MdInput) {
```
jelbourn: Why does this need to check `MdInput` specifically now? Add a comment?

src/lib/form-field/form-field.ts
```diff
@@ -0,0 +1,382 @@
...
+      transition('void => enter', [
+        style({ opacity: 0, transform: 'translateY(-100%)' }),
+        animate('300ms cubic-bezier(0.55, 0, 0.55, 0.2)'),
+      ]),
+    ]),
+  ],
+  host: {
+    // Remove align attribute to prevent it from interfering with layout.
+    '[attr.align]': 'null',
+    'class': 'mat-input-container mat-form-field',
```
jelbourn: Add a comment that `mat-input-container` is deprecated and will be removed?

src/lib/form-field/form-field.ts
```diff
@@ -0,0 +1,382 @@
...
+
+/** Container for form controls that applies Material Design styling and behavior. */
+@Component({
+  moduleId: module.id,
+  selector: 'md-input-container, mat-input-container, md-form-field, mat-form-field',
+  templateUrl: 'form-field.html',
+  // MdInput is a directive and can't have styles, so we need to include its styles here.
+  // The MdInput styles are fairly minimal so it shouldn't be a big deal for people who aren't using
+  // MdInput.
+  styleUrls: ['form-field.css', '../input/input.css'],
```
jelbourn: This may be a problem in Google-land (but maybe not; not totally sure)

src/lib/form-field/form-field.ts
```diff
@@ -0,0 +1,382 @@
...
+      state('enter', style({ opacity: 1, transform: 'translateY(0%)' })),
+      transition('void => enter', [
+        style({ opacity: 0, transform: 'translateY(-100%)' }),
+        animate('300ms cubic-bezier(0.55, 0, 0.55, 0.2)'),
+      ]),
+    ]),
+  ],
+  host: {
+    // Remove align attribute to prevent it from interfering with layout.
+    '[attr.align]': 'null',
```
jelbourn: Do we still use `align` on this?

src/lib/form-field/form-field.ts
```diff
@@ -0,0 +1,382 @@
...
+    '[class.ng-valid]': '_shouldForward("valid")',
+    '[class.ng-invalid]': '_shouldForward("invalid")',
+    '[class.ng-pending]': '_shouldForward("pending")',
+    '(click)': '_control.focus()',
+  },
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+
+export class MdFormField implements AfterViewInit, AfterContentInit, AfterContentChecked {
```
jelbourn: In `public_api.ts`, we should re-export this as `MdInputContainer` to stay backwards compatible with anyone importing it (about 18 instances within Google)

src/lib/input/index.ts
```diff
@@ -7,50 +7,36 @@
...
-    MdErrorDirective,
-    MdHint,
-    MdInputContainer,
-    MdInputDirective,
-    MdPlaceholder,
-    MdPrefix,
-    MdSuffix,
+    // TODO(mmalerba): We import and re-export the form field module since all existing users of
+    // `MdInput` will need this to continue using `md-input-container`. We may want to keep this
+    // long term since the `MdInput` directive will almost always be used with `md-form-field`.
```
jelbourn: Yeah, long-term I think the fact that `form-field` is its own thing is an implementation detail (similar to mdLine) that the users don't need to care about

src/lib/form-field/form-field.ts
```diff
@@ -0,0 +1,382 @@
...
+
+/** Suffix to be placed at the end of the form field. */
+@Directive({
+  selector: '[mdSuffix], [matSuffix]',
+})
+export class MdSuffix {}
+
+
+/** An interface which allows a control to work inside of a `MdFormField`. */
+export abstract class MdFormFieldControl {
```
jelbourn: Make this an `interface` since it doesn't have any default implementations? I recall Misko mentioning that abstract classes should be avoided for payload size reasons

jelbourn: Move this to its own file? Now that this is its own thing, we could probably break it up even more  (prefix, suffix, error, and hint in their own files). That would bring this file down to that sweet, sweet 200 - 300 line range.

src/lib/autocomplete/autocomplete-trigger.ts
mmalerba: Hmm now that I think about it, its probably better to make the `value` property part of the `MdFormFieldControl` interface so this doesn't have to be hardcoded for `MdInput`. I was only thinking about`mdInput` and `md-select` when I wrote this, but its reasonable to want autocomplete to work with a user's custom input control as well

src/lib/form-field/form-field.ts
mmalerba: If it winds up being a problem I can move the file under `form-field/` with a note about why its there

src/lib/form-field/form-field.ts
mmalerba: No, but we do on `md-hint`, so moved this to there

src/lib/input/index.ts
mmalerba: Ok so then input, select, and any other components that wind up using it should all re-export it in their respective modules?

src/lib/form-field/form-field.ts
mmalerba: I did this because I need to use the class itself as the injection token so it can be used with `@ViewChild` (something like `InjectionToken<MdFormFieldControl>` causes a type error). It doesn't have to be abstract though, I could use a concrete class with no-op methods.

mmalerba:
comments addressed

jelbourn:
There are failures in the aot and e2e tests. Potentially need to regenerate the example module

src/lib/form-field/_form-field-theme.scss
```diff
@@ -0,0 +1,218 @@
...
+@mixin mat-form-field-theme($theme) {
+  $primary: map-get($theme, primary);
+  $accent: map-get($theme, accent);
+  $warn: map-get($theme, warn);
+  $background: map-get($theme, background);
+  $foreground: map-get($theme, foreground);
+  $is-dark-theme: map-get($theme, is-dark);
+
+  // Placeholder colors. Required is used for the `*` star shown in the placeholder.
+  $form-field-placeholder-color: mat-color($foreground, secondary-text);
```
crisbeto: Since these are local variables, you can get rid of the `form-field-` prefix.

src/lib/form-field/form-field-control.ts
```diff
@@ -0,0 +1,53 @@
...
+   * Stream that emits whenever the state of the control changes such that the parent `MdFormField`
+   * needs to run change detection.
+   */
+  stateChanges: Observable<void>;
+
+  /** The value of the control. */
+  value: any;
+
+  /** Gets the element ID for this control. */
+  abstract getId(): string;
```
crisbeto: Perhaps this, `getPlaceholder` and `getNgControl` should be properties instead?

src/lib/form-field/form-field-control.ts
```diff
@@ -0,0 +1,53 @@
...
+/** An interface which allows a control to work inside of a `MdFormField`. */
+export abstract class MdFormFieldControl {
+  /**
+   * Stream that emits whenever the state of the control changes such that the parent `MdFormField`
+   * needs to run change detection.
+   */
+  stateChanges: Observable<void>;
+
+  /** The value of the control. */
+  value: any;
```
crisbeto: Instead of having `any` here, the `MdFormFieldControl` could be generic.

src/lib/form-field/form-field-control.ts
```diff
@@ -0,0 +1,53 @@
...
+import {NgControl} from '@angular/forms';
+
+
+/** An interface which allows a control to work inside of a `MdFormField`. */
+export abstract class MdFormFieldControl {
+  /**
+   * Stream that emits whenever the state of the control changes such that the parent `MdFormField`
+   * needs to run change detection.
+   */
+  stateChanges: Observable<void>;
```
crisbeto: Perhaps allow this to be a `Subject` as well so we don't have to add `asObservable`?

src/lib/form-field/form-field-control.ts
```diff
@@ -0,0 +1,53 @@
...
+  value: any;
+
+  /** Gets the element ID for this control. */
+  abstract getId(): string;
+
+  /** Gets the placeholder for this control. */
+  abstract getPlaceholder(): string;
+
+  /** Gets the NgControl for this control. */
+  abstract getNgControl(): NgControl | null;
```
crisbeto: Instead of `NgControl | null` this could be an optional property/method instead.

src/lib/form-field/form-field.ts
```diff
@@ -0,0 +1,289 @@
...
+  moduleId: module.id,
+  // TODO(mmalerba): the input-container selectors and classes are deprecated and will be removed.
+  selector: 'md-input-container, mat-input-container, md-form-field, mat-form-field',
+  templateUrl: 'form-field.html',
+  // MdInput is a directive and can't have styles, so we need to include its styles here.
+  // The MdInput styles are fairly minimal so it shouldn't be a big deal for people who aren't using
+  // MdInput.
+  styleUrls: ['form-field.css', '../input/input.css'],
+  animations: [
+    trigger('transitionMessages', [
```
crisbeto: I know that this is just a refactoring, but maybe we should consider getting the placeholder animation to go through Angular as well? This allows users to disable it a bit easier.

mmalerba:
@jelbourn Those failures appear to be due to the re-export of `MdFormField` as `MdInputContainer` in public_api.ts. Perhaps because I'm trying to export it under 2 different names?

src/lib/form-field/form-field-control.ts
mmalerba: I believe a `Subject` is a type of `Observable`? Just for the purposes of this interface we only care about its `Observable` features.

src/lib/form-field/form-field-control.ts
mmalerba: I just didn't want anyone trying to set them...

src/lib/form-field/form-field-control.ts
jelbourn: Having a `Subject` on the concrete class would satisfy it being an `Observable`

src/lib/form-field/form-field-control.ts
jelbourn: Could use `readonly`

src/lib/form-field/form-field-control.ts
mmalerba: mmm, I think anyone who wants to implement this interface will want to include this method, its just that the method may sometimes return null if angular/forms isn't being used. (same rationale as above for not wanting a property).

src/lib/form-field/form-field.ts
mmalerba: :+1: added a TODO

mmalerba:
comments addressed

PR #6346, opened 2017-08-08
fix(theming): Light green 700 contrast should be white

cladera:
According to [Material Design Guidelines](https://material.io/guidelines/style/color.html#color-color-palette) Light Green 700 contrast color
should be white.


cladera:
I added a Breaking Change message since it may have an unexpected behaviour for those using this color since it may change entirely the look of their applications.

PR #6348, opened 2017-08-08
demo(icon): Add an accessibility demo page for icon

tinayuangao:


src/demo-app/a11y/icon/icon-a11y.html
```diff
@@ -0,0 +1,14 @@
+<section>
+  <h2>Home icon (Ligature from Material Icons font)</h2>
```
mmalerba: I think we should have examples for each of the accessibility use cases listed here: https://material.angular.io/components/icon/overview

tinayuangao: Thanks! Changed to uses cases listed in https://material.angular.io/components/icon/overview

tinayuangao:
Added another example with `<a md-icon-button>`. PTAL. Thanks!

PR #6352, opened 2017-08-08
perf(observe-content): run outside Angular zone

crisbeto:
* Runs the underlying `MutationObserver` outside the Angular zone.
* Runs the debounce timeout outside the Angular zone.

via https://github.com/angular/flex-layout/pull/370.


mmalerba:
please rebase

PR #6360, opened 2017-08-08
demo(dialog): Add accessibility demo page for dialog

tinayuangao:


src/demo-app/a11y/dialog/dialog-a11y.html
```diff
@@ -0,0 +1,23 @@
+<section>
+  <h2>Welcome message</h2>
+  <p>Activat the button to see a welcome dialog with a simple message and a close button.</p>
```
jelbourn: `Activat` -> `Activate`

src/demo-app/a11y/dialog/dialog-a11y.ts
```diff
@@ -0,0 +1,77 @@
...
+})
+export class DialogFruitExampleDialog {}
+
+@Component({
+  moduleId: module.id,
+  selector: 'dialog-welcome-a11y',
+  template: `Welcome to Angular Material dialog demo page!
+
+<md-dialog-actions>
+  <button
```
jelbourn: Should this dialog have some other content?

src/demo-app/a11y/dialog/dialog-neptune-a11y.html
```diff
@@ -0,0 +1,33 @@
+<h2 md-dialog-title>Neptune</h2>
+
+<md-dialog-content>
+  <img src="https://upload.wikimedia.org/wikipedia/commons/5/56/Neptune_Full.jpg"/>
```
jelbourn: `img` needs an `alt`

src/demo-app/a11y/dialog/dialog-neptune-a11y.html
```diff
@@ -0,0 +1,33 @@
...
+<md-dialog-actions [attr.align]="actionsAlignment">
+  <button md-raised-button
+          color="primary"
+          md-dialog-close>Close</button>
+
+  <a
+          md-button
+          color="primary"
+          href="https://en.wikipedia.org/wiki/Neptune"
+          target="_blank">Read more on Wikipedia</a>
```
jelbourn: The indent / line-breaking here is odd

src/demo-app/a11y/dialog/dialog-neptune-iframe-a11y.html
```diff
@@ -0,0 +1,12 @@
+<h2 md-dialog-title>Neptune</h2>
```
jelbourn: Where does the focus go when this dialog is opened?

src/demo-app/a11y/dialog/dialog-a11y.ts
tinayuangao: Do you have an example?

src/demo-app/a11y/dialog/dialog-neptune-iframe-a11y.html
tinayuangao: The "Close" button in the dialog.

src/demo-app/a11y/dialog/dialog-a11y.ts
jelbourn: I was just thinking a `<p>` with some text

src/demo-app/a11y/dialog/dialog-neptune-iframe-a11y.html
jelbourn: I suspect that having the focus start _after_ all of the dialog content is weird

@tinayuangao @crisbeto what do you think of adding a dialog config option like `focusDialogElementOnOpen` (and also `ariaLabel`) to help with dialogs like this?

crisbeto: It does make sense, although specifying what element takes precedence in the focus trap should already be possible through the `cdk-focus-initial` attribute.

jelbourn: That doesn't give a way to focus the `role="dialog"` element, though, which I _think_ is what we'd want (but I'm not 100% sure).

tinayuangao: Adding a config option makes sense to me. 

I added another example for forms. The focus goes to the first input box in the form when the dialog is opened.

kara:
@tinayuangao Rebase?

tinayuangao:
@kara Rebased. Thanks!

PR #6361, opened 2017-08-08
feat(stepper): Add animation to stepper

g1shin:
Animation is added to support smooth transition between steps upon step selection change.

### DIFF
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -20,7 +20,7 @@ import {
   Component,
   ContentChild,
   ViewChild,
-  TemplateRef, AfterContentChecked
+  TemplateRef
 } from '@angular/core';
 import {LEFT_ARROW, RIGHT_ARROW, ENTER, SPACE} from '@angular/cdk/keyboard';
 import {CdkStepLabel} from './step-label';
@@ -31,13 +31,13 @@ import {AbstractControl} from '@angular/forms';
 let nextId = 0;
 
 /**
- * Position state of the content of each step in horizontal stepper that is used for transitioning
+ * Position state of the content of each step in stepper that is used for transitioning
  * the content into correct position upon step selection change.
  */
-export type CdkStepContentPositionState = 'left' | 'center' | 'right';
+export type StepContentPositionState = 'previous' | 'current' | 'next';
 
 /** Change event emitted on selection changes. */
-export class CdkStepperSelectionEvent {
+export class StepperSelectionEvent {
   /** Index of the step now selected. */
   selectedIndex: number;
 
@@ -77,20 +77,6 @@ export class CdkStep {
   @Input()
   label: string;
 
-  /** Position state of step. */
-  _position: CdkStepContentPositionState;
-
-  /** Sets the CdkStepContentPositionState of step. */
-  set position(position: number) {
-    if (position < 0) {
-      this._position = 'left';
-    } else if (position > 0) {
-      this._position = 'right';
-    } else {
-      this._position = 'center';
-    }
-  }
-
   constructor(private _stepper: CdkStepper) { }
 
   /** Selects this step component. */
@@ -106,7 +92,7 @@ export class CdkStep {
     '(keydown)': '_onKeydown($event)'
   },
 })
-export class CdkStepper implements AfterContentChecked {
+export class CdkStepper {
   /** The list of step components that the stepper is holding. */
   @ContentChildren(CdkStep) _steps: QueryList<CdkStep>;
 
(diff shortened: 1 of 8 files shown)


src/cdk/stepper/stepper.ts
```diff
@@ -131,6 +152,10 @@ export class CdkStepper {
     this._groupId = nextId++;
   }
 
+  ngAfterContentChecked(): void {
```
kara: This is called every change detection run, which seems like it may be overkill. Can you clarify why this is necessary? Would `ngAfterContentInit` work here?

src/cdk/stepper/stepper.ts
```diff
@@ -30,6 +30,12 @@ import {AbstractControl} from '@angular/forms';
 /** Used to generate unique ID for each stepper component. */
 let nextId = 0;
 
+/**
+ * Position state of the content of each step in horizontal stepper that is used for transitioning
+ * the content into correct position upon step selection change.
+ */
+export type CdkStepContentPositionState = 'left' | 'center' | 'right';
```
jelbourn: Does this need to take RTL into consideration?

src/lib/stepper/stepper-horizontal.ts
```diff
@@ -19,6 +20,15 @@ import {MdStepper} from './stepper';
     'class': 'mat-stepper-horizontal',
     'role': 'tablist',
   },
+  animations: [
+    trigger('stepEnterExit', [
+      state('left', style({transform: 'translate3d(-100%, 0, 0)'})),
+      state('center', style({transform: 'translate3d(0%, 0, 0)'})),
+      state('right', style({transform: 'translate3d(100%, 0, 0)'})),
+      transition('* => left, * => right, left => center, right => center',
```
kara: Could you do `* => center` as well, or will that not work?

src/cdk/stepper/stepper.ts
```diff
@@ -71,6 +77,20 @@ export class CdkStep {
   @Input()
   label: string;
 
+  /** Position state of step. */
+  _position: CdkStepContentPositionState;
+
+  /** Sets the CdkStepContentPositionState of step. */
+  set position(position: number) {
```
jelbourn: `position` by itself doesn't seem specific enough. Maybe `animationDirection`?

src/cdk/stepper/stepper.ts
jelbourn: Can this be generalized such that it isn't only for horizontal steps? Maybe `previous` /  `current` / `next`?

jelbourn: Don't need to prefix this with `Cdk`; my stance is that only directives/components with `cdk-` selector should have that reflected in the identifier name

src/cdk/stepper/stepper.ts
jelbourn: Rather than having a setter for this and reading the property, why not just always read a function like `getAnimationDirection` that's based off of the current index?

src/lib/stepper/stepper.scss
```diff
@@ -70,8 +74,16 @@ $mat-vertical-stepper-margin-top: $mat-stepper-side-gap - $mat-connector-line-ga
...
-    display: none;
+.mat-horizontal-stepper-content {
+  overflow: hidden;
+
+  &[aria-expanded='false'] {
+    height: 0;
+  }
+}
+
+.horizontal-content-container {
```
jelbourn: Prefix all css classes with `mat-`

src/cdk/stepper/stepper.ts
g1shin: Yes, it would need to once the `Directionality` related things are added to stepper, but I haven't added any code dealing with `Directionality`. I'll probably address them all at once, since it will also influence other interactions such as `_foucsIndex` management for horizontal stepper.

src/cdk/stepper/stepper.ts
g1shin: Now neither is needed with the new change :)

g1shin:
Changes have been made based on review. Ready for review again :+1:

src/lib/stepper/_stepper-theme.scss
```diff
@@ -33,11 +33,11 @@
     background-color: mat-color($background, card);
   }
 
-  .vertical-content-container {
+  .mat-vertical-content-container {
     border-left-color: mat-color($foreground, divider);
   }
 
-  .connector-line {
+  .mat-connector-line {
```
mmalerba: before this gets merged into master you should go through and make sure the classes all start with `mat-stepper-` or `mat-step-`

PR #6363, opened 2017-08-09
demo(list): Add accessibility demo page for list

tinayuangao:


src/demo-app/a11y/list/list-a11y.html
```diff
@@ -0,0 +1,71 @@
...
+    <md-list>
+      <md-list-item *ngFor="let item of items"> {{item}} </md-list-item>
+    </md-list>
+  </section>
+
+
+  <section>
+    <h2>Mailbox navigation</h2>
+    <p>Showing a navigation list with links to google search, and "more information" icon button</p>
+    <md-nav-list>
```
jelbourn: `md-nav-list` is only meant to contain `<a md-list-item>` children.

This makes me realize that we probably need tweaks as part of this:
* Removing the existing roles and saying that, by default, `md-list` is assumed to be presentational, and add instructions to add the roles for non-interactive content (same as grid-list).
* Give `md-nav-list` `role="navigation"`

Thoughts?

tinayuangao: - Made another component `MdNavList` for `mat-nav-list` with `role="navigation"`
- Removed `role="list"` for `MdList` 
- Added "accessibility" section in list.md (Same as grid-list)

src/lib/list/list.md
```diff
@@ -141,3 +141,13 @@ To add a divider, use `<md-divider>`.
    </md-list-item>
 </md-list>
 ```
+
+### Accessibility
+By default, the list assumes that it will be used in a purely decorative fashion and thus sets no
+roles, ARIA attributes, or keyboard shortcuts.
+This is equivalent to having a sequence of <div> elements on the page. Any interactive content
```
kara: Nit: Missing new line before this paragraph starts?

src/lib/list/list.ts
```diff
@@ -146,9 +158,9 @@ export class MdListItem extends _MdListItemMixinBase implements AfterContentInit
   constructor(private _renderer: Renderer2,
               private _element: ElementRef,
               @Optional() private _list: MdList,
-              @Optional() navList: MdNavListCssMatStyler) {
```
kara: Are we still using `MdNavListCssMatStyler` for anything or can we remove it now?

src/demo-app/a11y/list/list-a11y.html
```diff
@@ -0,0 +1,66 @@
...
+        <h3 md-line> {{message.from}} </h3>
+        <p md-line> {{message.subject}} </p>
+        <p md-line class="demo-secondary-text"> {{message.message}} </p>
+      </md-list-item>
+    </md-list>
+
+  </section>
+
+  <section>
+    <h2> Seasoning</h2>
```
crisbeto: Spacing inside the tag seems uneven.

src/demo-app/a11y/list/list-a11y.html
```diff
@@ -0,0 +1,66 @@
...
+    <md-nav-list>
+      <a md-list-item *ngFor="let link of links"
+         href="https://www.google.com/search?q={{link.name}}">
+        {{link.name}}
+      </a>
+    </md-nav-list>
+  </section>
+
+  <section>
+    <h2>Messages </h2>
```
crisbeto: Uneven spacing here as well.

src/lib/list/list.ts
```diff
@@ -146,9 +158,9 @@ export class MdListItem extends _MdListItemMixinBase implements AfterContentInit
...
+              @Optional() private _navList: MdNavList) {
```
crisbeto: This should potentially have `@Host` as well, otherwise it'll look up the component tree.

src/demo-app/a11y/list/list-a11y.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    {
+      from: 'Bobby',
+      subject: 'Oui oui',
+      message: 'Do you have Paris reservations for the 15th? I just booked!',
+      image: 'https://angular.io/generated/images/bios/jelbourn.jpg'
+    }
+  ];
+
+  links: any[] = [
```
crisbeto: If you remove the `any[]` from these arrays, TS should be able to figure out the proper type.

tinayuangao:
Thanks for review. 

Removed `MdNavListCssMatStyler` and `MdListCssMatStyler`.

src/lib/list/list.md
```diff
@@ -141,3 +141,12 @@ To add a divider, use `<md-divider>`.
    </md-list-item>
 </md-list>
 ```
+
+### Accessibility
+By default, the list assumes that it will be used in a purely decorative fashion and thus sets no
+roles, ARIA attributes, or keyboard shortcuts. This is equivalent to having a sequence of <div>
+elements on the page. Any interactive content within the grid-list should be given an appropriate
```
jelbourn: This line still says `grid-list`

src/lib/list/list.md
```diff
@@ -141,3 +141,12 @@ To add a divider, use `<md-divider>`.
...
+accessibility treatment based on the specific workflow of your application.
+
+If the list is used to present a list of non-interactive content items, then the list element should
+be given `role="list"` and each tile should be given `role="listitem"`.
```
jelbourn: `tile` -> `list item`

tinayuangao:
Comments addressed. PTAL. Thanks!

kara:
@tinayuangao There looks like there might be a real unit test failure for list here: https://travis-ci.org/angular/material2/jobs/265757068

Can you take a look?

Issue #4647, opened 2017-05-18
Add a beforeClose method for MdDialog


Dunos:
**Feature Request**

**What is the expected behavior?**

When the user is about to close a MdDialog window (via Esc key, clickout or close button), for example an edit dialog, it would be nice to be able to have some method to prevent the window from closing to avoid unsaved changes for example.

**What is the use-case or motivation for changing an existing behavior?**

Since the MdDialog already listens to Esc key events, clickout and has the close() method that is used in the action buttons, it would be much easier to have a method like _afterClosed_ that is called just before the closing event is going to occur, so we could prevent it from happening just returning `true` or `false` in a callback function. This way we don't have to overwrite those event listeners.

willshowell:
Related https://github.com/angular/material2/pull/3460

jelbourn:
We did originally plan on adding a `beforeClose` hook, but not with the ability to cancel the close. The stance has been that, rather than canceling close, the user can `disableClose` and take control of closing the dialog via whatever criteria suits their application.

Dunos:
Well, I guess something like that could work too. Do you expect to have this functionally any time soon?

davidmorissette:
In my opinion `beforeClose` and `disableClose` serve two different purposes. Imagine a scenario where you want to prompt a user before discarding the changes (no matter how the cancellation happened); then using `beforeClose` would allow us to cancel closing the dialog. However, `disableClose` would completely remove the ability for the user to close the dialog (via cancellation) which is not what is expected.

wulfsberg:
To pitch in here (from over in https://github.com/angular/material2/issues/5209), I'd also like to have the _value_ of the user choice exposed in an observable before/on close, to get access to it without having to wait for the animation.
(This is somewhat tangential to the beforeClose _method_, but it seems to be tracked collectively here).

mtycholaz:
Much needed! We have a use-case where we want to save data if the user cancels the dialog (via clicking on backdrop or hitting the Esc key). Where the problem comes in, is if there's a server error, we want to keep the modal up still to display the message.

PR #6377, opened 2017-08-09
feat(dialog): add beforeClose method

willshowell:
Fixes https://github.com/angular/material2/issues/4647

### DIFF
--- a/src/lib/dialog/dialog.spec.ts
+++ b/src/lib/dialog/dialog.spec.ts
@@ -140,6 +140,20 @@ describe('MdDialog', () => {
     });
   }));
 
+  it('should close a dialog and get back a result before it is closed', async(() => {
+    let dialogRef = dialog.open(PizzaMsg, { viewContainerRef: testViewContainerRef });
+    let beforeCloseCallback = jasmine.createSpy('beforeClose callback');
+
+    dialogRef.beforeClose().subscribe(beforeCloseCallback);
+    dialogRef.close('Bulbasaurus');
+    viewContainerFixture.detectChanges();
+
+    viewContainerFixture.whenStable().then(() => {
+      expect(beforeCloseCallback).toHaveBeenCalledWith('Bulbasaurus');
+      expect(overlayContainerElement.querySelector('md-dialog-container')).toBeNull();
+    });
+  }));
+
   it('should close a dialog via the escape key', async(() => {
     dialog.open(PizzaMsg, {
       viewContainerRef: testViewContainerRef
--- a/src/lib/dialog/dialog.spec.ts
+++ b/src/lib/dialog/dialog.spec.ts
@@ -141,7 +141,7 @@ describe('MdDialog', () => {
   }));
 
   it('should close a dialog and get back a result before it is closed', async(() => {
-    const dialogRef = dialog.open(PizzaMsg, { viewContainerRef: testViewContainerRef });
+    const dialogRef = dialog.open(PizzaMsg, {viewContainerRef: testViewContainerRef});
 
     // beforeClose should emit before dialog container is destroyed
     const beforeCloseHandler = jasmine.createSpy('beforeClose callback').and.callFake(() => {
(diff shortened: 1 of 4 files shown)


src/lib/dialog/dialog.spec.ts
```diff
@@ -141,6 +141,20 @@ describe('MdDialog', () => {
...
+    let dialogRef = dialog.open(PizzaMsg, { viewContainerRef: testViewContainerRef });
+    let beforeCloseCallback = jasmine.createSpy('beforeClose callback');
+
+    dialogRef.beforeClose().subscribe(beforeCloseCallback);
+    dialogRef.close('Bulbasaurus');
+    viewContainerFixture.detectChanges();
+
+    viewContainerFixture.whenStable().then(() => {
+      expect(beforeCloseCallback).toHaveBeenCalledWith('Bulbasaurus');
+      expect(overlayContainerElement.querySelector('md-dialog-container')).toBeNull();
```
jelbourn: This test checks that `beforeClose` occurred, but not _when_ it occurred. I'd restructure to be like
```ts
let beforeCloseInvoked = false;
const beforeCloseHandler = () => {
  expect(overlayContainerElement.querySelector('md-dialog-container')).toBeDefined();
  beforeCloseInvoked = true;
});
...
viewContainerFixture.whenStable().then(() => {
  expect(beforeCloseCallback).toHaveBeenCalledWith('Bulbasaurus');
  expect(overlayContainerElement.querySelector('md-dialog-container')).toBeNull();
});
```

willshowell:
@jelbourn update to the test has been made. Let me know if I misinterpreted your comment!

src/lib/dialog/dialog.spec.ts
```diff
@@ -140,6 +140,25 @@ describe('MdDialog', () => {
     });
   }));
 
+  it('should close a dialog and get back a result before it is closed', async(() => {
+    const dialogRef = dialog.open(PizzaMsg, { viewContainerRef: testViewContainerRef });
```
jelbourn: Omit spaces inside braces

kdubious:
I must be missing something, but how do we cancel the close in beforeClose()?

willshowell:
@kdubious check out https://github.com/angular/material2/issues/4647#issuecomment-302569792

kdubious:
@willshowell, thanks. I read it, but still have a point of confusion.

I'm trying to pass a DTO to a dialog. The dialog has a form (ReactiveForm) that displays the DTO, the user makes their edits, then clicks "Save." At that point, I want the parent component (it maintains a list of the objects being edited in the dialog) to be notified about the current state of the object, so that the parent can manage saving to a remote service. If unsuccessful, the parent would prevent the dialog from closing.

Is there another approach  should consider?

willshowell:
@kdubious I probably should've linked to this thread #3460, particularly https://github.com/angular/material2/pull/3460#issuecomment-319420262. Basically `beforeClose` is simply for getting the close value before the closing animation, but does not allow for synchronous or asynchronous cancellation of the close. If you want to cancel closing of the dialog (via <kbd>esc</kbd> or clicking on the backdrop), then you'll want to use `disableClose` and manually subscribe to the backdrop click and keydown event streams.

It sounds like in your case, you need to run the async validation before the dialog actually closes. If communication with the parent component is the issue, try using a service.

kdubious:
Thanks @willshowell. I ended up going with the service.

wiikka:
@willshowell I really don't see any real use case for beforeClose as it is currently implemented? As requested it would make much more sense if you could cancel the close event (e.g. asking from user if he wants to discard the changes). In that use case it doesn't matter where you get the close event (is it esc, backdrop click or even dialogRef.close()). You should be able to intercept all those and cancel the close if you wish?

willshowell:
@wiikka one of the use cases for beforeClose that was mentioned in the original issue was when you don't care how the dialog was closed, and you don't care to prevent it, but you want to do something with the resulting value without having to wait for the animation to complete.

The conversation in https://github.com/angular/material2/pull/3460, and more specifically https://github.com/angular/material2/pull/3460#issuecomment-319420262,  pretty much covers why an all-encompassing "prevent/cancel close" api wasn't chosen.

PR #6385, opened 2017-08-09
fix(select): truncate long placeholder text

### LINKED ISSUE #6373: md-select long placeholder text wraps and covers drop down value
Bug: Long placeholder text drops and covers selected value

Long placeholder text should have overflow ellipsis and nowrap so it doesn't look weird.

Long placeholder text drops and covers selected value

Create a md-select with placeholder text that is longer than 1 line according to select width

[Example](http://plnkr.co/edit/PXeNWkFX8weMC9tOnJCD)

It is a bug and an inconsistent behavior with other components (e.g. md-input)

Angular 4.3.2
Material 2.0.0-beta.8
TypeScript 2.4.2
Windows 10
Chrome 60

Simple fix should be adding the following CSS:
```
    .mat-select-placeholder {
        white-space: nowrap;
        text-overflow: ellipsis;
        overflow: hidden;
    }
```
Workaround for now, add the above with `/deep/` prefix for the selector.

crisbeto:
Truncates long placeholder text instead of wrapping it.

Fixes #6373.

### DIFF
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -14,6 +14,7 @@ $mat-select-trigger-underline-height: 1px !default;
 .mat-select {
   display: inline-block;
   outline: none;
+  max-width: 100%;
 }
 
 .mat-select-trigger {
@@ -53,6 +54,9 @@ $mat-select-trigger-underline-height: 1px !default;
   padding: 0 2px;
   transform-origin: left top;
   flex-grow: 1;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
 
   // These values are duplicated from animation code in order to
   // allow placeholders to sometimes float without animating,


kara:
@crisbeto The input handles this case a bit differently, in that it will remove the ellipses if the long placeholder will fit when shrunk to 75% in floating mode. @mmalerba Since you are rewriting a lot of input/select in form-field right now, should he re-work this or do you already have this in shared code?

crisbeto:
This would be handled in the shared code @kara, but we could go with this until we make the switch to `md-form-field`.

mmalerba:
Yeah I'm currently rewriting the select and had to change the structure for the trigger a little bit to make the baselines align. I would recommend holding off on select changes for a while. This PR might not be necessary after my changes

crisbeto:
Alright, closing in that case.

PR #6390, opened 2017-08-10
compat(ripple): add mat compatible inputs

andrewseguin:


src/lib/core/ripple/ripple.ts
```diff
@@ -82,6 +82,41 @@ export class MdRipple implements OnChanges, OnDestroy {
   /** Whether foreground ripples should be visible outside the component's bounds. */
   @Input('mdRippleUnbounded') unbounded: boolean;
 
+  // Properties with `mat-` prefix for noconflict mode.
+  @Input('matRippleTrigger')
+  get matRippleTrigger() { return this.trigger; }
```
DevVersion: I think it would make sense to prefix this with an underscore right? 

Nobody should set the `matRippleTrigger` programmatically neither it should show up in the docs.

Similar as in: https://github.com/angular/material2/blob/4a1f10ecd5eee00b3ecb7e08b93aa5dd03a32641/src/lib/dialog/dialog-content-directives.ts#L35)

andrewseguin: Yes good catch, the underline should be there

PR #6403, opened 2017-08-10
fix(autocomplete): highlighted option not reset when closed with escape or enter key

### LINKED ISSUE #6258: [autocomplete] invalid initially active option after escape
Bug

Opening the autocomplete doesn't initially show any items as active

Closing an autocomplete via <kbd>esc</kbd> and then opening it again via down arrow key results in the first option being visibly active, but not selectable.

http://plnkr.co/edit/xXY6Gx88PUVCLI5MimtJ?p=preview

crisbeto:
Fixes the autocomplete active option not being reset when closing through the escape or enter keys. Previously the old active option would remain active upon reopening, however the user's focus would shift back to the top as soon as they press an arrow key.

Fixes #6258.


crisbeto:
I'm not sure it is, although it probably isn't as noticeable because you're only left with one option at that point. I'll do the same for enter and add an extra unit test for it.

PR #6404, opened 2017-08-10
fix(select): prevent clicks on hidden placeholder

### LINKED ISSUE #6400: md-select does not honour floatPlaceholder="never" if the ngModel is set at initialization
md-select is still floating the placeholder even if floatPlaceholder="never" if the ngModel is set at initialization

floatPlaceholder="never" is honoured

Placeholder is invisible (opacity: 0) but is still there and clickable, leading to accidentally opening the select if the placeholder is overlapped with another element.

https://plnkr.co/edit/Z3qeYrqbEaEzirGgWC4c?p=preview

Sometimes we want to have multiple selects close by, such that the vertical padding between each select is less than 16px, in these cases it is possible to trigger the wrong select if the user clicks sufficiently low on the above select

all

crisbeto:
Fixes being able to click on the placeholder when there's a value and `floatPlaceholder="never"`.

Fixes #6400.

### DIFF
--- a/src/lib/select/select.html
+++ b/src/lib/select/select.html
@@ -9,8 +9,8 @@
     class="mat-select-placeholder"
     [class.mat-floating-placeholder]="_hasValue()"
     [@transformPlaceholder]="_getPlaceholderAnimationState()"
-    [style.opacity]="_getPlaceholderOpacity()"
-    [style.width.px]="_selectedValueWidth"> {{ placeholder }} </span>
+    [class.mat-select-placeholder-hidden]="_getPlaceholderVisibility()"
+    [style.width.px]="_selectedValueWidth">{{ placeholder }}</span>
 
   <span class="mat-select-value" *ngIf="_hasValue()">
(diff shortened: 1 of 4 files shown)


crisbeto:
Looks like we've come full circle, good spot @kara. It's reworked now, can you take another look?

crisbeto:
@kara we already have a test that verifies that the class is being toggled correctly. I'm not sure how else to test it, aside from an e2e test, but we don't have e2e tests set up for `md-select` in general.

kara:
Would you mind adding an e2e? I think it's important to test bugs like this to make sure they don't come back (esp given what just happened with visibility/opacity)

crisbeto:
Sorry for the email spam, I thought this was another issue.

crisbeto:
If @mmalerba is working on the switch to `md-form-field`, this shouldn't be necessary anymore (see https://github.com/angular/material2/pull/6385#issuecomment-322552234). Closing and have added a note to myself to set up e2e tests for `md-select`.

PR #6408, opened 2017-08-10
feat(portal): support context in TemplatePortal

### LINKED ISSUE #6310: Align TemplatePortal with ComponentPortal and support context in TemplateRef Dialogs
`TemplatePortal` should accept a context which then allows using the `MdDialogConfig.data` as context for a dialog opened using a `TemplateRef`.

`TemplatePortal` does no accepts context.
It does accept `locals` but these were added within the first commit and never used.

There are several...

First, `TemplatePortal` is not aligned with `ComponentPortal`. Since both provide the same logical functionality it makes sense to return the same logical object: `ComponentRef`, `EmbeddedViewRef`. In addition, `TemplatePortal` does not allow sending data (context) to the template. There is a place holder (`locals`) property but it is never used. `TemplatePortal` is not responsible for the structure of the context so `locals` as `Map<string, any>` is not a proper type, instead a generic type is more suited. The consumer of `TemplatePortal` should decide about the structure (e.g. using `$implicit`? what local to expose etc...)

Once `TemplatePortal` support context, `MdDialog` can provide same experience for both **Component** based dialogs and `TemplateRef` based dialogs.

`MdDialog` will auto crete the context have the `MdDialogConfig.data` as `$implicit` and an extra property `dialogRef` exposed (which holds `MdDialogRef`).

This create a powerful combination of a template that has access to injected data (context) and local template variables of its **parent** template.  
It allows passing template references throughout the application with the ability to have data injected. It's a more declarative approach with a lot of power.

For example, using the `demo-app` code, **without changing TS code** we can provide the same functionality (almost) that the **demo-jazz-dialog** component provide just by changing the html only.

```html
<ng-template let-data let-dialogRef="dialogRef">
  I'm a template dialog. I've been opened {{numTemplateOpens}} times!

  <p>It's Jazz!</p>

  <md-input-container>
    <input mdInput placeholder="How much?" #howMuch>
  </md-input-container>

  <p> {{ data.message }} </p>
  <button type="button" (click)="dialogRef.close(lastCloseResult = howMuch.value)">Close dialog</button>
  <button (click)="dialogRef.updateSize('500px', '500px').updatePosition({ top: '25px', left: '25px' });">Change dimensions</button>`
</ng-template>
```

This template uses bindings from 3 sources, the context, local template variables of the parent component and the component instance itself.

With simple structured directives it will be possible to create template driven UI blockers, modal etc... that accepts data and are set elsewhere.

 I'v started to work on this change, [see the diff](https://github.com/angular/material2/compare/master...shlomiassaf:refactor-TemplatePortal?expand=1)

The change is not big but having TemplatePortal now with a generic required some mods, and it's also a breaking change so it's needs discussion.

Please comment.

cc @jelbourn 
Thanks.

shlomiassaf:
Closes #6310

cc @jelbourn

### DIFF
--- a/src/cdk/portal/portal.spec.ts
+++ b/src/cdk/portal/portal.spec.ts
@@ -10,10 +10,11 @@ import {
   Optional,
   Injector,
   ApplicationRef,
+  TemplateRef
 } from '@angular/core';
 import {CommonModule} from '@angular/common';
 import {TemplatePortalDirective, PortalHostDirective, PortalModule} from './portal-directives';
-import {Portal, ComponentPortal} from './portal';
+import {Portal, ComponentPortal, TemplatePortal} from './portal';
 import {DomPortalHost} from './dom-portal-host';
 
 
@@ -43,6 +44,57 @@ describe('Portals', () => {
       // Expect that the content of the attached portal is present.
       let hostContainer = fixture.nativeElement.querySelector('.portal-container');
       expect(hostContainer.textContent).toContain('Pizza');
+    });
+
+    it('should load a template into the portal', () => {
+      let testAppComponent = fixture.debugElement.componentInstance;
+      let hostContainer = fixture.nativeElement.querySelector('.portal-container');
+
+      let templatePortal = new TemplatePortal(testAppComponent.templateRef, null!);
+      testAppComponent.selectedPortal = templatePortal;
+      fixture.detectChanges();
+      // Expect that the content of the attached portal is present and no context is projected
+      expect(hostContainer.textContent).toContain('Banana');
+    });
+
+    it('should project template context bindings in the portal', () => {
+      let testAppComponent = fixture.debugElement.componentInstance;
+      let hostContainer = fixture.nativeElement.querySelector('.portal-container');
+
+      // TemplatePortal without context:
+      let templatePortal = new TemplatePortal(testAppComponent.templateRef, null!);
+      testAppComponent.selectedPortal = templatePortal;
+      fixture.detectChanges();
+      // Expect that the content of the attached portal is present and NO context is projected
+      expect(hostContainer.textContent).toContain('Banana - !');
+
+      // using TemplatePortal.attach method to set context
+      testAppComponent.selectedPortal = undefined;
+      fixture.detectChanges();
+      templatePortal.attach(testAppComponent.portalHost, {$implicit: {status: 'rotten'}});
+      fixture.detectChanges();
+      // Expect that the content of the attached portal is present and context given via the
+      // attach method is projected
+      expect(hostContainer.textContent).toContain('Banana - rotten!');
+
+      // using TemplatePortal constructor to set the context
+      templatePortal =
+        new TemplatePortal(testAppComponent.templateRef, null!, {$implicit: {status: 'fresh'}});
+      testAppComponent.selectedPortal = templatePortal;
+      fixture.detectChanges();
+      // Expect that the content of the attached portal is present and context given via the
+      // constructor is projected
+      expect(hostContainer.textContent).toContain('Banana - fresh!');
+
+      // using TemplatePortal constructor to set the context but also calling attach method with
+      // context, the latter should take precedence:
+      testAppComponent.selectedPortal = undefined;
+      fixture.detectChanges();
+      templatePortal.attach(testAppComponent.portalHost, {$implicit: {status: 'rotten'}});
+      fixture.detectChanges();
+      // Expect that the content of the attached portal is present and and context given via the
+      // attach method is projected and get precedence over constructor context
+      expect(hostContainer.textContent).toContain('Banana - rotten!');
     });
 
     it('should dispose the host when destroyed', () => {
@@ -299,7 +351,7 @@ describe('Portals', () => {
       fixture.detectChanges();
 
       // Attach the TemplatePortal.
-      testAppComponent.portalWithBinding.attach(host);
+      testAppComponent.portalWithBinding.attach(host, {$implicit: {status: 'fresh'}});
       fixture.detectChanges();
 
       // Now that the portal is attached, change detection has to happen again in order
@@ -307,7 +359,7 @@ describe('Portals', () => {
       fixture.detectChanges();
 
       // Expect that the content of the attached portal is present.
-      expect(someDomElement.textContent).toContain('Banana');
+      expect(someDomElement.textContent).toContain('Banana - fresh');
 
       // When updating the binding value.
       testAppComponent.fruit = 'Mango';
@@ -416,18 +468,22 @@ class ArbitraryViewContainerRefComponent {
   <ng-template cdk-portal>Cake</ng-template>
 
   <div *cdk-portal>Pie</div>
-  <ng-template cdk-portal> {{fruit}} </ng-template>
+  <ng-template cdk-portal let-data> {{fruit}} - {{ data?.status }} </ng-template>
 
   <ng-template cdk-portal>
     <ul>
       <li *ngFor="let fruitName of fruits"> {{fruitName}} </li>
     </ul>
   </ng-template>
+
+  <ng-template #templateRef let-data> {{fruit}} - {{ data?.status }}!</ng-template>
   `,
 })
 class PortalTestApp {
   @ViewChildren(TemplatePortalDirective) portals: QueryList<TemplatePortalDirective>;
   @ViewChild(PortalHostDirective) portalHost: PortalHostDirective;
+  @ViewChild('templateRef', { read: TemplateRef }) templateRef: TemplateRef<any>;
+
   selectedPortal: Portal<any>;
   fruit: string = 'Banana';
   fruits = ['Apple', 'Pineapple', 'Durian'];
@@ -449,6 +505,7 @@ class PortalTestApp {
   get portalWithTemplate() {
     return this.portals.toArray()[3];
   }
+
 }
 
 // Create a real (non-test) NgModule as a workaround for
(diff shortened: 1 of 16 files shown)


src/cdk/portal/portal.ts
```diff
@@ -103,42 +104,38 @@ export class ComponentPortal<T> extends Portal<ComponentRef<T>> {
...
   }
 
   get origin(): ElementRef {
     return this.templateRef.elementRef;
   }
 
-  attach(host: PortalHost, locals?: Map<string, any>): Map<string, any> {
-    this.locals = locals == null ? new Map<string, any>() : locals;
+  attach(host: PortalHost, context?: C): C {
+    this.context = context!;
```
jelbourn: Why use the `!` here? The class property is marked as `| undefined` so it shouldn't be necessary

src/lib/dialog/dialog.ts
```diff
@@ -228,7 +228,9 @@ export class MdDialog {
     }
 
     if (componentOrTemplateRef instanceof TemplateRef) {
-      dialogContainer.attachTemplatePortal(new TemplatePortal(componentOrTemplateRef, null!));
+      dialogContainer.attachTemplatePortal(
+        new TemplatePortal<T>(componentOrTemplateRef, null!,
```
jelbourn: Doesn't `null!` resolve to `never`? Seems like this needs a real `ViewContainerRef`

src/lib/snack-bar/snack-bar-container.ts
```diff
@@ -110,7 +111,8 @@ export class MdSnackBarContainer extends BasePortalHost implements OnDestroy {
   }
 
   /** Attach a template portal as content to this snack bar container. */
-  attachTemplatePortal(): Map<string, any> {
+  /* tslint:disable-next-line */
```
jelbourn: What's the lint problem here?

src/cdk/portal/portal.spec.ts
```diff
@@ -299,15 +299,15 @@ describe('Portals', () => {
       fixture.detectChanges();
 
       // Attach the TemplatePortal.
-      testAppComponent.portalWithBinding.attach(host);
+      testAppComponent.portalWithBinding.attach(host, { $implicit: { status: 'fresh' } });
```
jelbourn: Omit spaces in braces
```ts
{$implicit: {status: 'fresh'}}
```

src/cdk/portal/portal.ts
```diff
@@ -103,42 +104,38 @@ export class ComponentPortal<T> extends Portal<ComponentRef<T>> {
...
-   * Additional locals for the instantiated embedded view.
-   * These locals can be seen as "exports" for the template, such as how ngFor has
-   * index / event / odd.
-   * See https://angular.io/docs/ts/latest/api/core/EmbeddedViewRef-class.html
-   */
-  locals: Map<string, any> = new Map<string, any>();
+  context: C | undefined;
 
-  constructor(template: TemplateRef<any>, viewContainerRef: ViewContainerRef) {
+  constructor(template: TemplateRef<any>, viewContainerRef: ViewContainerRef, context?: C) {
```
jelbourn: It's somewhat strange that the `context` can be set either in the constructor or in `attach`. Is this meant to be a _default_ context when nothing is passed to `attach`?

src/cdk/portal/portal.ts
```diff
@@ -103,42 +104,38 @@ export class ComponentPortal<T> extends Portal<ComponentRef<T>> {
...
     super();
     this.templateRef = template;
     this.viewContainerRef = viewContainerRef;
+    if (context) {
+     this.context = context;
+    }
```
jelbourn: You should just be able to put `public context?: C` in the constructor and omit this assignment

src/cdk/portal/portal.ts
jelbourn: Won't this always override the context given in the constructor, even if no context is passed here? You should be able to do something like
`attach(host, PortalHost, context: C = this.context)`

src/cdk/portal/portal.spec.ts
jelbourn: There should be:
* One test where no context is given
* One test where context is given in the constructor
* One test where context is given in `attach`
* One test where context is given in both the constructor and in attach

src/lib/dialog/dialog.spec.ts
```diff
@@ -80,6 +81,28 @@ describe('MdDialog', () => {
     expect(dialogContainerElement.getAttribute('role')).toBe('dialog');
   });
 
+  it('should open a dialog with a template', () => {
+    const templateRefFixture = TestBed.createComponent(ComponentWithTemplateRef);
+    templateRefFixture.componentInstance.localValue = 'Bees';
+    templateRefFixture.detectChanges();
+
+    const data = { value: 'Knees' };
```
jelbourn: Omit spaces inside braces

src/cdk/portal/portal.ts
shlomiassaf: This is required to support the abstraction in [dialog.ts](https://github.com/shlomiassaf/material2/blob/9ee6478063fd081b824d206abc93ebe803947d43/src/lib/dialog/dialog.ts#L232-L233)

If you follow the trail, the `attach` method is never called and `BasePortalHost` (`PortalHostDirective) handles it from the top.

Adding a context to the method call of `attachTemplatePortal` is a mess since it involves changing a lot of classes that extend `BasePortal`, for example `MdDialogContainer`

Additionally, changing the signature of `attachTemplatePortal` to accept a context will make it different from the signature `attachComponentPortal`...

src/cdk/portal/portal.ts
shlomiassaf: I just tried keeping the code style the same.

The other 2 parameters (`template` and `template`) are also public but they are manually assigned without using TypeScript's constructor parameter modifiers....

So I just followed that.

src/cdk/portal/portal.ts
shlomiassaf: Indeed.

I think `this.context = context || this.context;` is more readable but I used your approach

It requires a slight modification to be friends with typescript:
```
attach(host: PortalHost, context: C | undefined = this.context): C {
```

src/lib/dialog/dialog.ts
shlomiassaf: This was the logic before I made the change so I left it as is.

I think if a change is required here it should be handled in a different PR not to mixup things.

It does work...

src/lib/snack-bar/snack-bar-container.ts
shlomiassaf: 'C' is declared but never used.

src/lib/dialog/dialog.spec.ts
shlomiassaf: 👍

src/cdk/portal/portal.spec.ts
shlomiassaf: 👍

src/cdk/portal/portal.ts
shlomiassaf: 👍

src/cdk/portal/portal.spec.ts
shlomiassaf: Actually the tests are limited since there are no are no "low-level" tests for `TemplatePortal` with a `PortalHostDirective` (there are some for `ComponentPortal`)

I'v added 2 tests:
  1. **should load a template into the portal**
Check basic projection of programatically created `TemplatePortal`

  2. **should project template context bindings in the portal**
Check project of programatically created `TemplatePortal` with 4 internal `expects`
       - No context
       - Context through `attach()` method
       - Context through **constructor**
       - Context through **constructor** and through `attach()` method - the latter must take precedence.

src/cdk/portal/portal.ts
jelbourn: Ah, I think that's just from this being one of the first things I wrote in the library and not being as familiar with TypeScript back then.

src/cdk/portal/portal.ts
jelbourn: I see what you mean. I think it's okay to include it in the signature, then, so long as the JsDoc makes it clear that this updates the `context` property of the `TemplatePortal`

src/lib/snack-bar/snack-bar-container.ts
jelbourn: Ah, you should be able to do 
```ts
attachTemplatePortal(): EmbeddedViewRef<any> {
  ...
}
```

shlomiassaf: Yes, we can do that but this will not be inline with other `attachTemplatePortal` calls that does accept a type parameter.

This is not the only reason, main reason is that right now we need to add the `tslint disable` but once we move to TypeScript 2.4 it will be easy to migrate it to:

```ts
attachTemplatePortal<C = any>(): EmbeddedViewRef<C> {
      throw Error('Not yet implemented');
}
```

This way we allow setting the type parameter for those who want.
Later we will support those who "doesnt" want as well :)

jelbourn: Using `any` here right now is the most technically correct thing, though (until the function is actually implemented). The linter is catching a real issue in that the function genuinely isn't generic.

shlomiassaf: No problem, will update it

src/cdk/portal/portal.ts
shlomiassaf: @jelbourn I'v added a JsDoc comment to the `attach` method acoordignly.

I didn't find an example of how to comment a constructor signature, and if its like commenting a method, so I left it as is.

FYI: Both `template` and  `viewContainerRef` constructor parameters do the same thing (set a property on the instance) and are not commented so I don't think we should go into it. It is self explaining.

shlomiassaf:
@jelbourn Why does it fail?

jelbourn:
e2e tests are just timing out, which happens sometimes

PR #6423, opened 2017-08-11
demo(a11y): Add skip navigation link for a11y demo

tinayuangao:


src/demo-app/a11y/a11y.html
```diff
@@ -1,9 +1,12 @@
 <h1>Accessibility Demo</h1>
 
 <nav>
+  <a (click)="skipNavigation()" href="javascript:void(0)">Skip navigation</a>
```
jelbourn: Can this be done with a url fragment? Using `javascript:void(0)` is a pattern that is generally considered harmful to a11y

tinayuangao: Fragment is not working https://github.com/angular/angular/issues/13636 
We can change this to a material button and remove `href="javascript:void(0)"`

jelbourn: Chatted w/ Jason a bit about this to see if we could do any kind of workaround. And... not really.

Does it work if you make the `href="#maincontent"` and still use the click handler to do the work?

tinayuangao: Changed to use `routerLink` with current path and it works. Removed `javascript:void(0)`

kara:
@tinayuangao Can you check out the linter error? https://travis-ci.org/angular/material2/jobs/264586210

PR #6424, opened 2017-08-11
feat(tabs): add isActive flag on the individual tabs

### LINKED ISSUE #6422: MdTab: Add property to determine if the tab is active

Feature request

It would be helpful to have a property on `MdTab` that returns true if it is the currently active tab.

Seems like the only way to get the active tab is via the `selectedIndex` property in `MdTabGroup`. Even then, it returns an index, rather than an instance of `MdTab`. There does not seem to be a straightforward way to get the `MdTab` instance based on the index, other than to use `MdTabGroup._tabs.find` (which also doesn't seem right, since `_tabs` is not part of the public API).

Lots of reasons why someone might want this. In my case, I'm trying to avoid instantiating a component until the tab gets selected (partly for performance, and also because the component does not function correctly if it is initially rendered in a hidden container).

Ideally, I would like to do something like this:

```
<md-tab-group>
  <md-tab #mapTab label="Map">
    <div *ngIf="mapTab.isActive">
      <agm-map ...></agm-map>
    </div>
  </md-tab>
</md-tab-group>
```

However, to my knowledge, there is no such property as `isActive` (or similar) available on `MdTab`.

2.0.0-beta.8

crisbeto:
* Adds the `isActive` flag on each tab, making it easier to check whether a tab is active.
* Adds an `exportAs` to the tab directive to allow for it to be consumed easier in the view.

Fixes #6422.

### DIFF
--- a/src/lib/tabs/tab-group.spec.ts
+++ b/src/lib/tabs/tab-group.spec.ts
@@ -1,5 +1,5 @@
 import {async, ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
-import {Component, ViewChild} from '@angular/core';
+import {Component, QueryList, ViewChild, ViewChildren} from '@angular/core';
 import {BrowserAnimationsModule, NoopAnimationsModule} from '@angular/platform-browser/animations';
 import {By} from '@angular/platform-browser';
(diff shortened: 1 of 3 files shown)


jelbourn:
Just needs rebase

PR #6426, opened 2017-08-11
fix(tabs): allow for tabs to be selected using the space key

### LINKED ISSUE #6406: [feature request] Select md-tab-group by highlighting and pressing space key
feature request

Highlighted tab is selectable by both enter and space key

Only enter key

Create tab group, focus a tab, press space 
http://plnkr.co/edit/kLMZ6Oe67mmTXx6TAY34?p=info

This behavior is used throughout other components and fairly standard to select highlighted option through space or enter

2.0.0-beta.8

I can self assign and do this myself if not being worked on already

crisbeto:
* Allows for the user to select a tab using the space key, in addition to enter. This seems to be consistent with the M1 implementation of tabs.
* Prevents the default action for space and enter in order to prevent the page from scrolling by accident.

Fixes #6406.

### DIFF
--- a/src/lib/tabs/tab-header.spec.ts
+++ b/src/lib/tabs/tab-header.spec.ts
@@ -4,7 +4,7 @@ import {
 import {Component, ViewChild, ViewContainerRef} from '@angular/core';
 import {CommonModule} from '@angular/common';
 import {By} from '@angular/platform-browser';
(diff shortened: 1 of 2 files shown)


jelbourn:
Just needs rebase

PR #6430, opened 2017-08-12
build: wrong bundle name in demo app

crisbeto:
Fixes a crash in the demo app due to a wrong bundle name in the SystemJS config.


crisbeto:
Fixed in https://github.com/angular/material2/pull/6437.

PR #6435, opened 2017-08-12
chore(demo): fix demo app reference to overlay

amcdnl:
Fixes typo in the demo app is referencing the wrong overlay module...

```
'@angular/cdk/overlay': 'dist/bundles/cdk-overlay-content.umd.js',
```

where it should be referencing:

```
'@angular/cdk/overlay': 'dist/bundles/cdk-overlay.umd.js',
```


crisbeto:
https://github.com/angular/material2/pull/6430 should handle this already.

amcdnl:
@crisbeto - awesome, didn't see that. closing.

PR #6441, opened 2017-08-13
refactor(autocomplete): make keyboard and visibility calls synchronous

crisbeto:
Reworks a few methods, that were changing the view in a `Promise.resolve`, to be synchronous. The initial idea behind deferring those calls was to avoid "changed after checked" errors, however they also make things harder to test. It seems like the root of the issue for all of them was that we were trying to modify view properties after the `options.changed` callback had fired, which comes after change detection. These changes have the advantage of making things a lot easier and cleaner to test, as well as making the API easier to use.


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -376,8 +374,13 @@ export class MdAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
   private _subscribeToClosingActions(): Subscription {
     const firstStable = first.call(this._zone.onStable);
-    const optionChanges = map.call(this.autocomplete.options.changes, () =>
-      this._positionStrategy.recalculateLastPosition());
+    const optionChanges = switchMap.call(this.autocomplete.options.changes, () => {
+      this._positionStrategy.recalculateLastPosition();
+
+      // Defer emitting to the stream until the next tick, because changing
+      // bindings in here will cause "changed after checked" errors.
+      return Promise.resolve();
```
kara: Mixing switchMap and Promise.resolve doesn't feel quite right. Use `do` and `observeOn(asap)` operators here instead?  It would be more rx idiomatic.

crisbeto:
@kara I've switch it to use `delay` instead of `switchMap`. Can you take another look?

kara:
@crisbeto Looks like a linter failure bc map is unused. Update?

crisbeto:
Fixed the linting error.

PR #6442, opened 2017-08-13
fix(expansion-panel): animation not working in Angular 4.3

crisbeto:
Fixes the animation not working correctly in Angular 4.3. It seems like having both the margin and height animation on the element causes the height one to break. This approach also has the advantage of making it easier to override the margin and being able to use the SASS variables if necessary.

For reference, here's what it looks like in master:
a


crisbeto:
I'm not sure what a fix would be. Note that the margin is only added if the expansion panel is a part of an accordion. Also now that it uses a CSS class, it's much easier to override.

PR #6443, opened 2017-08-13
fix: improved key manager typings

crisbeto:
Makes the FocusKeyManager and ActivedescendantKeyManager generic which avoids having to cast the items all the time (e.g. https://github.com/angular/material2/blob/master/src/lib/autocomplete/autocomplete-trigger.ts#L234).


kara:
@crisbeto Can you check out the failing tests? Looks like the spec for ListKeyManager needs further updates.

crisbeto:
Fixed @kara.

PR #6454, opened 2017-08-14
refactor(icon): use HttpClient instead of Http

### LINKED ISSUE #5814: @angular/http not marked as peer dependency
Bug

If `@angular/http` is installed but `HttpModule` not loaded in `AppModule`, compilation is OK and icons work well (the font is loaded locally so no server requests are needed).

But if `@angular/http` is uninstalled, Angular CLI compilation fails (and so I can't even run my app) : 

```
ERROR in/node_modules/@angular/material/typings/icon/icon-registry.d.ts (10,22): Cannot find module '@angular/http'.

ERROR in ./~/@angular/material/@angular/material.es5.js
Module not found: Error: Can't resolve '@angular/http' in '/node_modules/@angular/material/@angular'
 @ ./~/@angular/material/@angular/material.es5.js 20:0-37
 @ ./src/app/core/core.module.ts
 @ ./src/app/app.module.ts
 @ ./src/main.ts
 @ multi webpack-dev-server/client?http://localhost:4200 ./src/main.ts
```

Angular 4.3 introduced new HttpClient API, meaning we may not need anymore the previous Http API. And in this case, we should be able to not install `@angular/http` in the project, to have smaller bundles sizes.

Material icons seems to do hard imports of `@angular/http` even if it doesn't need them. They should be remove when not necessary.

`@angular/cli` 1.2.1
`@angular/core` 4.3.0
`@angular/material` beta.8

cyrilletuzi:
Fixes #5814 

Replace old HttpModule with new Angular 4.3+ HttpClientModule.

It allows to delete the dependency to `@angular/http`, then smaller app bundles for users.

As a consequence, dependencies has been updated to Angular >= 4.3.0.


cyrilletuzi:
As it's an important PR and as I'm not an expert in running all the tools and tests, I need special review for : 

- the whole icon feature still running well ;
- that the global dependency update to Angular 4.3 doesn't break anything ;
- that the new way to mock requests in the `icon.spec.ts` is ok.

jelbourn:
We haven't made this change yet because it would require bumping to 4.3, which breaks some of our tests (still trying to figure out why)

crisbeto:
https://github.com/angular/material2/pull/6483 will add the Angular 4.3 support and sort out the various test failures.

cyrilletuzi:
Closing, as PR #6702 redone the same work...

PR #6462, opened 2017-08-14
feat(stepper): Add background for focus and hover and fix focus logic

g1shin:
Add background for hover and focus state. 

Also, this fixes the focus logic so that the selected step's header doesn't automatically receive focus; instead, it will be focused only if the step is directly selected by clicking on the header and if it is able to be selected. (Selecting the next step by clicking on the next button will no longer automatically place focus background on the header of that step.)

screen shot 2017-08-16 at 11 36 49 am
screen shot 2017-08-16 at 11 36 58 am


mmalerba:
@g1shin needs rebase

g1shin:
Rebase done; ready for review :+1:

src/cdk/stepper/stepper.ts
```diff
@@ -210,6 +210,8 @@ export class CdkStepper {
     if (this._linear) {
       for (let i = 0; i < index; i++) {
         if (!stepsArray[i].stepControl.valid) {
+          // remove the focus from clicked step header if the step is not able to be selected
+          this._stepHeader.toArray()[index].nativeElement.blur();
```
kara: Nit: Hmm I can see why you did it this way, but blurring is a side effect I wouldn't expect from a function that simply reduces the validity.  I could see someone using it elsewhere without realizing it does this. Perhaps you can return the invalid step instead of true and blur in selectedIndex?

src/lib/stepper/_stepper-theme.scss
```diff
@@ -9,6 +9,11 @@
 
   .mat-horizontal-stepper-header, .mat-vertical-stepper-header {
 
+    &:focus,
+    &:hover {
+      background-color: mat-color($background, hover);
```
mmalerba: Can we make it look more like what the spec says?

g1shin:
Made changes based on review and added tests for focus change 👍 Ready for review again.

mmalerba:
I still have some CSS changes I'd like to make. @g1shin lets discuss it offline, will be easier to explain

g1shin:
CSS changes made based on offline discussion 😃 Ready for review.

PR #6464, opened 2017-08-14
fix(chips): set appropriate aria-orientation

crisbeto:
Based on the [accessibility guidelines](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox), the `listbox` role has a default `aria-orientation` of `vertical`, however the chips in a chip list are horizontal.


src/lib/chips/chip-list.ts
```diff
@@ -45,6 +45,7 @@ import {coerceBooleanProperty} from '@angular/cdk/coercion';
   host: {
     '[attr.tabindex]': '_tabIndex',
     'role': 'listbox',
+    'aria-orientation': 'horizontal',
```
tinayuangao: Stacked chips are vertical.  
I'm okay with the change. But I think we need to update the docs to let developers set correct `aria-orientation` if they use stacked chip list.

crisbeto:
Addressed the feedback. Can you take another look @tinayuangao?

src/lib/chips/chips.md
```diff
@@ -33,8 +33,19 @@ chips are neither selectable nor focusable. Currently, disabled chips receive no
 Users can move through the chips using the arrow keys and select/deselect them with the space. Chips
 also gain focus when clicked, ensuring keyboard navigation starts at the appropriate chip.
 
+### Orientation
+If you want the chips in the list to be stacked verticall, instead of horizontally, you can apply
```
tinayuangao: `vertically`?

PR #6483, opened 2017-08-15
feat: update to Angular 4.3

crisbeto:
* Bumps the Angular requirement to 4.3. This is also a prerequisite to the Angular 5 update since we encountered the same errors there.
* Refactors the sidenav not to use a `Promise.resolve` for toggling the initial animation, because it was making it impossible to flush the animations in unit tests. The new simplified approach uses a property to pass the animation state.
* Fixes the nested menu test failures.
* Fixes a "changed after checked" error in the autocomplete demo app.

**Note:** The expansion panel animation doesn't work correctly with Angular 4.3 at the moment. It will be fixed by https://github.com/angular/material2/pull/6442.

Relates to #6454.


crisbeto:
@jelbourn it looks like the e2e tests caught that issue with the `start` callback firing before the `done`. Can you take a look at the workaround in https://github.com/angular/material2/pull/6483/commits/ea47d874875a0ef7cfbbd8e5887c23d4a370f61a?

jelbourn:
Seems okay to me

PR #6485, opened 2017-08-15
demo(chips): Update chips demo with form field

tinayuangao:
Current chips demo the input is not usable when there're too many chips

### DIFF
--- a/src/demo-app/chips/chips-demo.html
+++ b/src/demo-app/chips/chips-demo.html
@@ -46,19 +46,19 @@
       </p>
 
 
-      <md-form-field>
-        <md-chip-list mdPrefix #chipList1>
+      <md-form-field class="has-chip-list">
+        <md-chip-list #chipList1 required>
           <md-chip *ngFor="let person of people" [color]="color" [selectable]="selectable"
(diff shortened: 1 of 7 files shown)


src/demo-app/chips/chips-demo.html
```diff
@@ -46,22 +46,20 @@
       </p>
 
 
-      <md-form-field>
-        <md-chip-list mdPrefix #chipList1>
+      <md-chip-list #chipList1>
```
mmalerba: This seems weird, putting the `md-form-field` inside the `md-chip-list`, it feels like it should be the other way around... I think it should be possible now to have the `md-chip-list` implement the `MdFormFieldControl` interface and work with a parent `md-form-field` via that interface. Is this something you've considered?

gise88:
I am trying to find a solution for the following issues: https://github.com/angular/material2/issues/6966 https://github.com/angular/material2/issues/6967 https://github.com/angular/material2/issues/6968

Do you think that this pr can solve them?

gpazo:
is this fix in beta12 or can we expect it in the next release?

PR #6488, opened 2017-08-15
feat(select): make select work inside form-field

mmalerba:
Demos:
https://mmalerba-demo3.firebaseapp.com/baseline
https://mmalerba-demo3.firebaseapp.com/select

BREAKING CHANGES:
- `md-select` must now be wrapped in an `md-form-field`
- `MdFormFieldControl` now has 2 new properties: `controlType`, `shouldPlaceholderFloat` and 1 new method: `onContainerClick`


kara:
@mmalerba Great work!  I played around the demo. It seems mostly good, but one thing I noticed is that sometimes you can click on the trigger without actually opening the panel. If you try to click closer to the underline or the floating label, you can reproduce. The cursor is not a pointer in that case.

src/demo-app/baseline/baseline-demo.html
```diff
@@ -14,9 +14,12 @@
       <input mdInput placeholder="Input" value="Text Input">
     </md-form-field>
     | Text 5 |
-    <md-select placeholder="Select">
-      <md-option value="option">Option</md-option>
-    </md-select>
+    <md-form-field>
```
crisbeto: API-wise, do we really want people to have to wrap every select in a form field? E.g. if you were using option groups, you'd have to write 4 levels of HTML. Aside from having to proxy a few `@Input`-s, what are the cons of having the form field be consumed by `md-select` internally?

src/lib/select/select.ts
```diff
@@ -890,31 +893,29 @@ export class MdSelect extends _MdSelectMixinBase implements AfterContentInit, On
 
   /** Calculates the scroll position and x- and y-offsets of the overlay panel. */
   private _calculateOverlayPosition(): void {
+    this._triggerHeight = this.trigger.nativeElement.getBoundingClientRect().height;
+    this._triggerFontSize = this._measureFontSizeEl.nativeElement.getBoundingClientRect().height;
```
crisbeto: I'm not sure that this would be completely accurate. Why not use `getComputedStyle`?

mmalerba:
@kara good catch, I fixed the focus and click behavior and also corrected the 1px error in placement of the multi-select dropdown

src/demo-app/baseline/baseline-demo.html
mmalerba: In the past we've preferred composition style APIs since they're more flexible, even if they are a bit more verbose. I think the extra typing is worth it, but it's a fair point. @jelbourn do you have an opinion?

src/lib/select/select.ts
mmalerba: Do you know if `getComputedStyle` is guaranteed to give me a `px` value for `font-size` (as opposed to `em`, `%`, etc)? That was my main motivation for not doing`getComputedStyle`. Also, I assume `getComputedStyle` is no worse from a performance perspective?

crisbeto: I'm pretty sure that you'll get a pixel value no matter what. Should be pretty easy to verify. As for performance, it should be pretty close to the `getBoundingClientRect` call.

src/lib/input/input.spec.ts
```diff
@@ -515,36 +516,36 @@ describe('MdInput without forms', function () {
     fixture.detectChanges();
 
     let inputEl = fixture.debugElement.query(By.css('input')).nativeElement;
-    let labelEl = fixture.debugElement.query(By.css('label')).nativeElement;
+    let formFieldEl = fixture.debugElement.query(By.css('.mat-form-field')).nativeElement;
```
rafaelss95: @mmalerba Taking advantage of the fact that you're working on it, would you be able to replace, `let` by `const` (in tests) when possible? It seems like almost all the variables here (in test file) aren't being reassigned and still are `let`.

mmalerba: This sounds like something we should do as part o a separate cleanup pass of the whole project, I'm sure there are tools that can analyze and fix this for us

mmalerba:
@kara @jelbourn @crisbeto this is ready for more through review now

src/demo-app/baseline/baseline-demo.html
jelbourn: I do like the consistency of having both `select` and `input` inside the same `form-field`; it makes it clear that the prefix, suffix, hint, and error belong to the form field. Very explicit, no magic. It's also more maintainable for us, since we don't have to forward along every `form-field` API through select over time. 

I don't feel incredibly strongly about it, though. @mmalerba up to you if you think it would be the same amount of work to do it the other way around; it would certainly make the migration easier

src/lib/input/input.ts
```diff
@@ -52,10 +52,10 @@ let nextUniqueId = 0;
 @Directive({
   selector: `input[mdInput], textarea[mdInput], input[matInput], textarea[matInput]`,
   host: {
-    'class': 'mat-input-element',
+    'class': 'mat-input-element mat-form-field-autofill-float',
     // Native input properties that are overwritten by Angular inputs need to be synced with
     // the native input element. Otherwise property bindings for those don't work.
-    '[id]': 'id',
+    '[attr.id]': 'id',
```
jelbourn: Why is the `attr.` needed here?

emilio-martinez: Becuase `id` is also an input
https://github.com/mmalerba/material2/blob/32df1d401bd97ce0d2a5fcf9a5e1c2bac7ddf076/src/lib/input/input.ts#L105

benb7760:
@mmalerba should the error message for missing control be changed? It suggests that mdInput directive is required which may be confusing to unfamiliar users. Currently reads:

> md-form-field must contain a MdFormFieldControl. Did you forget to add mdInput to the native input or textarea element?

code-tree:
The `\xa0` fix doesn't appear to work with AOT?

image

src/lib/select/select.html
```diff
@@ -1,26 +1,22 @@
...
+     aria-hidden="true"
+     (click)="toggle()"
+     #origin="cdkOverlayOrigin"
+     #trigger>
+  <div class="mat-select-value">
+    <!--
+      TODO(mmalerba): &nbsp; is currently broken in components with preserveWhitespace: false, so we
+      evaluate it as a JS string binding instead. Change back to &nbsp; once it works again.
+    -->
+    <ng-container *ngIf="empty">{{'\xa0'}}</ng-container>
```
code-tree: This doesn't appear to work with AOT.

uday4vunnam:
@mmalerba  The Demos are great. Can you be able to provide example code of this?

joseluisq:
Awesome!
Definitely I can't wait for the next release!

PR #6489, opened 2017-08-15
feat(select): update the trigger width and overlay state when select is opened

emoralesb05:
when changing the width of the md-select (e.g. using flex box) after its rendered, the template[cdk-connected-overlay] does not update its minWidth and keeps the initial minWidth every time its opened.

This issue partially happens because the cdk-connected-overlay sets the width, minWidth, etc etc only once and never updates it again.

Here is a plnkr to demonstrate the issue: http://plnkr.co/edit/ERg71p?p=preview

#### Proposed fix in PR

We just need to set the new _triggerWidth every time the md-select is opened (open()) and update the overlayRef#OverlayState in ConnectedOverlayDirective when the template is attached.

New PR to replace what was stated in https://github.com/angular/material2/pull/3623


src/lib/select/select.ts
```diff
@@ -466,9 +466,11 @@ export class MdSelect extends _MdSelectMixinBase implements AfterContentInit, On
       return;
     }
 
-    if (!this._triggerWidth) {
-      this._setTriggerWidth();
-    }
+    /**
+     * Always set the trigger width since we need to
+     * update the panel with its new minWidth when opening it
+     */
```
kara: Nit: For comments inside methods, we typically use `//` comment style. Applies throughout PR.

src/cdk/overlay/overlay-directives.ts
```diff
@@ -340,6 +345,8 @@ export class ConnectedOverlayDirective implements OnDestroy, OnChanges {
     this._position.withDirection(this.dir);
     this._overlayRef.getState().direction = this.dir;
     this._initEscapeListener();
+    /** Update the overlay size, in case the directive's inputs have changed */
+    this._configureSize(this._overlayRef.getState());
```
kara: The `getState()` method isn't really meant to be used to set state. @jelbourn and I discussed and we think there should be a dedicated method in `OverlayRef` to set the size props in the state. There is already an `updateSize` method, but it's really an internal implementation. So let's:

- rename the existing `updateSize` to `_setSizeStyles`
- add a new `updateSize` method that simply updates the size props in the state

Then you can use the new `updateSize` to set the props here, instead of `getState()`.

emoralesb05: Cool, will do that 😄

kara: Ping me when it's ready! :)

emoralesb05:
@kara i updated it based on the comments

I wasnt sure if i needed to create a specific interface for the state sizes, but i opted for reusing the `OverlayState` interface.. just wondering if its worth abstracting `width, height, minWidth and minHeight` into its own interface for this..

Let me know if i need to change anything else or if i misunderstood anything. 😄

emoralesb05:
I dont know why the unit test failed for `md-menu`.. it doesnt fail locally for me..

kara:
@emoralesb05 Don't worry about that test; it's a flake (should be fixed by https://github.com/angular/material2/pull/6514).

emoralesb05:
Cool, was all confused about it haha 😄

emoralesb05:
@kara @crisbeto do i need to do any other changes? saw that this was opened that will affect this one too https://github.com/angular/material2/pull/6658

jelbourn:
@kara @crisbeto is this change still something we want?

emoralesb05:
@jelbourn @kara @crisbeto let me know if this is still needed so i can rebase it 😄  since i saw that @crisbeto opened a similar PR

crisbeto:
@emoralesb05 are you still able to rebase it?

emoralesb05:
Sure, let me try doing it today @crisbeto 😄

emoralesb05:
Im going to close this PR and opened a new one.. i think i screwed up the rebase somehow haha.. im not really good at it lol

emoralesb05:
@crisbeto @kara i opened https://github.com/angular/material2/pull/8765 and added a stackblitz with the bug replication 😄

PR #6496, opened 2017-08-16
compat: re-export all classes and types with mat prefix

andrewseguin:


jelbourn:
Seems to break the AOT build

PR #6503, opened 2017-08-16
chore(overlay): add `attach` method to PositionStrategy interface

jelbourn:
This will be used by the new positioning to get size information through the `OverlayRef`

Didn't add any tests since this doesn't add any externally observable behavior

### DIFF
--- a/src/cdk/overlay/overlay-ref.ts
+++ b/src/cdk/overlay/overlay-ref.ts
@@ -45,6 +45,10 @@ export class OverlayRef implements PortalHost {
   attach(portal: Portal<any>): any {
     let attachResult = this._portalHost.attach(portal);
 
+    if (this._state.positionStrategy) {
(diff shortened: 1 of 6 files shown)


src/cdk/overlay/overlay-ref.ts
```diff
@@ -45,6 +45,10 @@ export class OverlayRef implements PortalHost {
   attach(portal: Portal<any>): any {
     let attachResult = this._portalHost.attach(portal);
 
+    if (this._state.positionStrategy) {
+      this._state.positionStrategy.attach(this);
```
crisbeto: Doesn't this end up being a circular dependency? The overlay ref has a reference to the strategy already and now the strategy can have a reference to the ref.

jelbourn: So it does. I'll have to rework it a bit

PR #6504, opened 2017-08-16
chore(overlay): refactor clipping detection calculation to its own file

jelbourn:
Moving these functions out mainly to reduce the amount of code that goes in the position strategy

### DIFF
--- a/src/cdk/overlay/position/connected-position-strategy.ts
+++ b/src/cdk/overlay/position/connected-position-strategy.ts
@@ -13,23 +13,14 @@ import {
   ConnectionPositionPair,
   OriginConnectionPosition,
   OverlayConnectionPosition,
-  ConnectedOverlayPositionChange, ScrollableViewProperties
+  ConnectedOverlayPositionChange,
+  ScrollingVisibility,
 } from './connected-position';
 import {Subject} from 'rxjs/Subject';
 import {Observable} from 'rxjs/Observable';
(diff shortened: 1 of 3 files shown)


src/cdk/overlay/position/scroll-clip.ts
```diff
@@ -0,0 +1,46 @@
...
+    return outsideAbove || outsideBelow || outsideLeft || outsideRight;
+  });
+}
+
+
+/**
+ * Gets whether an element is clipped by any of its scrolling containers.
+ * @param element Dimensions of the element (from getBoundingClientRect)
+ * @param scrollContainers Dimensions of element's scrolling containers (from getBoundingClientRect)
+ * @returns Whether the element is clipped
```
crisbeto: Should this be `@docs-private` as well?

src/cdk/overlay/position/connected-position-strategy.ts
```diff
@@ -13,23 +13,13 @@ import {
   ConnectionPositionPair,
   OriginConnectionPosition,
   OverlayConnectionPosition,
-  ConnectedOverlayPositionChange, ScrollableViewProperties
+  ConnectedOverlayPositionChange, ScrollingVisibility
```
crisbeto: Move this to the next line?

jelbourn:
Done and done

PR #6505, opened 2017-08-16
chore(overlay): add `attach` method to PositionStrategy interface

jelbourn:


jelbourn:
After fixing the bad `import` this seems fine (same as the attach on `ScrollStrategy`)

src/cdk/overlay/position/position-strategy.ts
```diff
@@ -6,11 +6,17 @@
...
+
 /** Strategy for setting the position on an overlay. */
 export interface PositionStrategy {
 
+  /** Attaches this position strategy to an overlay. */
+  attach(overlay: OverlayRef): void;
+
   /** Updates the position of the overlay element. */
-  apply(element: Element): void;
+  apply(): void;
```
crisbeto: Maybe `attach` and `apply` should return `this` so they can be chained?

src/lib/menu/menu.spec.ts
```diff
@@ -974,7 +974,7 @@ describe('MdMenu', () => {
           .not.toContain('mat-elevation-z3', 'Expected no stacked elevation.');
     });
 
-    it('should close all of the menus when the root is closed programmatically', () => {
+    it('should close all of the menus when the root is closed programmatically', fakeAsync(() => {
```
crisbeto: You should get these changes if you rebase.

src/cdk/overlay/overlay.spec.ts
```diff
@@ -532,9 +532,14 @@ class OverlayTestModule { }
 class OverlayContainerThemingTestModule { }
 
 class FakePositionStrategy implements PositionStrategy {
-  apply(element: Element): Promise<null> {
-    element.classList.add('fake-positioned');
-    return Promise.resolve(null);
+  element: HTMLElement;
+
+  apply(): void {
+    this.element.classList.add('fake-positioned');
```
crisbeto: Add a test that this gets executed at the proper time?

src/cdk/overlay/position/connected-position-strategy.ts
```diff
@@ -89,9 +94,12 @@ export class ConnectedPositionStrategy implements PositionStrategy {
...
 
-  /**
-   * To be used to for any cleanup after the element gets destroyed.
-   */
+  attach(overlayRef: OverlayRef): void {
+    this._overlayRef = overlayRef;
+    this._pane = overlayRef.overlayElement;
+  }
+
+  /** To be used to for any cleanup after the element gets destroyed. */
```
kara: Typo: `to for any cleanup` -> `for any cleanup`

src/cdk/overlay/overlay.spec.ts
jelbourn: I don't follow what you mean; the existing test for this still applies- is there something that isn't covered?

src/cdk/overlay/position/position-strategy.ts
jelbourn: I'd rather leave them as is since normal use wouldn't chain them, just the tests

src/cdk/overlay/overlay.spec.ts
crisbeto: Ah sorry, didn't know that we had something in place already.

jelbourn:
This is technically a breaking API change, but I don't think anyone inside Google will be affected by it

PR #6506, opened 2017-08-16
fix(snack-bar): align theming with spec

crisbeto:
Aligns the snack bar theming with the spec by:
* Sets the accent color to the snack bar button.
* Removes the elevation from the snack bar.
* Adds an inverted black&white snack bar for dark themes. Based on MDC.


kara:
@crisbeto Rebase?

PR #6507, opened 2017-08-16
feat(stepper): Add background for focus and hover and fix focus logic

g1shin:
Add background for hover and focus state.

Also, this fixes the focus logic so that the selected step's header doesn't automatically receive focus; instead, it will be focused only if the step is directly selected by clicking on the header and if it is able to be selected. (Selecting the next step by clicking on the next button will no longer automatically place focus background on the header of that step.)

screen shot 2017-08-16 at 11 36 49 am
screen shot 2017-08-16 at 11 36 58 am


g1shin:
I had to close and recreate the PR b/c of some rebase issues 😞 
CSS changes have been added based on offline discussion. Ready for review.

PR #6508, opened 2017-08-16
feat(overlay): add maxWidth and maxHeight

jelbourn:


crisbeto:
@jelbourn looks like there's a linting error.

jelbourn:
Fixed lint
(I don't know why WebStorm doesn't trim trailing whitespace sometimes...)

PR #6509, opened 2017-08-16
feat(stepper): Support additional properties for step

g1shin:
Add support for `editable`, `optional`, and `completed` properties of each step.

screen shot 2017-08-22 at 10 07 36 am
screen shot 2017-08-22 at 10 07 46 am


g1shin:
I had to close and reopen the PR for additional properties of step b/c of rebase issues 😢 
Previous comments addressed; ready for review.

src/cdk/stepper/stepper.ts
```diff
@@ -77,6 +77,26 @@ export class CdkStep {
...
+  /** Whether the completion of step is optional or not. */
+  @Input()
+  get optional() { return this._optional; }
+  set optional(value: any) {
+    this._optional = coerceBooleanProperty(value);
+  }
+  private _optional = false;
+
+  /** Return whether step is completed or not. */
+  get completed() {
```
mmalerba: Hmm I wonder if this is something we want to give the user control of by making it an `@Input()`?

@jelbourn WDYT of something like this:
```ts
@Input()
get completed() {
  return this._customCompleted == null ? this._defaultCompleted : this._customCompleted;
}
set completed(value) {
  this._customCompleted = value == null ? null : coerceBooleanProperty(value);
}
private _customCompleted: boolean | null = null;

private get _defaultCompleted() {
  return this._stepControl ? this._stepControl.valid && this.interacted : this.interacted;
}
```

This way most users could use the default completeness behavior, but they could still override if they want

src/cdk/stepper/stepper.ts
```diff
@@ -211,7 +232,7 @@ export class CdkStepper {
     const stepsArray = this._steps.toArray();
     stepsArray[this._selectedIndex].interacted = true;
     if (this._linear) {
-      return stepsArray.slice(0, index).some(step => step.stepControl.invalid);
+      return stepsArray.slice(0, index).some(step => step.stepControl.invalid && !step.optional);
```
mmalerba: This makes sense if you're thinking about the step being invalid due to a required field not being filled out, but it makes less sense if you consider invalid for other reasons (e.g. entering "1234" as your email).

Need to think more about what the right behavior for optional is...

src/lib/stepper/stepper-horizontal.html
```diff
@@ -8,16 +8,25 @@
          [tabIndex]="_focusIndex == i ? 0 : -1"
          (click)="step.select()"
          (keydown)="_onKeydown($event)">
-      <div class="mat-stepper-index">
+      <div class="mat-stepper-index-new" *ngIf="!step.completed || selectedIndex == i">
         {{i + 1}}
       </div>
+      <div class="mat-stepper-index-interacted" *ngIf="step.completed && selectedIndex != i">
```
mmalerba: I think we can  simplify this a bit:
```ts
get _indicatorType(): 'number' | 'check' | 'edit' {...}
```
```html
<div class="mat-step-indicator" [ngSwitch]="_indicatorType">
  <span *ngSwitchCase="number">{{i + 1}}</span>
  <md-icon *ngSwitchCase="check">done</md-icon>
  <md-icon *ngSwitchCase="edit">create</md-icon>
</div>

File: src/lib/stepper/stepper-horizontal.html [hunk 4]
```diff
@@ -8,16 +8,25 @@
...
+        <md-icon *ngIf="!step.editable">done</md-icon>
+        <md-icon *ngIf="step.editable">create</md-icon>
+      </div>
 
-      <div class="mat-stepper-label">
+      <div [ngClass]="{
```
mmalerba: do these as separate `[class.xxx]` bindings

src/cdk/stepper/stepper.ts
jelbourn: Seems reasonable, though `_defaultCompleted` would be a function

src/demo-app/stepper/stepper-demo.html
```diff
@@ -19,12 +19,12 @@
       </div>
     </md-step>
 
-    <md-step formGroupName="1" [stepControl]="formArray.get([1])">
+    <md-step formGroupName="1" [stepControl]="formArray.get([1])" [optional]="true">
```
jelbourn: Or just `optional` with no expression

src/lib/stepper/stepper.spec.ts
```diff
@@ -457,6 +485,57 @@ function checkStepHeaderBlur(stepHeaderEl: HTMLElement, fixture: ComponentFixtur
   expect(stepHeaderEl.blur).toHaveBeenCalled();
 }
 
+function checkEditableStep(stepperComponent: MdStepper,
```
jelbourn: Need function descriptions. I'd also prefer to have the functions names start with `assert` instead of `check`, and potentially making the function names more specific about what they do

src/lib/stepper/stepper-vertical.html
```diff
@@ -6,19 +6,28 @@
        [tabIndex]="_focusIndex == i ? 0 : -1"
        (click)="step.select()"
        (keydown)="_onKeydown($event)">
-    <div class="mat-stepper-index">
+    <div class="mat-stepper-index-new" *ngIf="!step.completed || selectedIndex == i">
       {{i + 1}}
     </div>
+    <div class="mat-stepper-index-interacted" *ngIf="step.completed && selectedIndex != i">
```
kara: Could we share this icon group between stepper vertical and stepper horizontal?

src/cdk/stepper/stepper.ts
g1shin: Do you mean that if the step is optional but the field is invalid due to some other reason, the user shouldn't be able to move on? If it is optional, shouldn't the user be able to move on even if the input is not valid? The field will still fire an error, so the user will still be aware that the input is invalid as they move on.

src/lib/stepper/stepper-vertical.html
g1shin: @kara Sorry, what do you mean by sharing the icon group? Do you mean like creating a separate icon group component with the template?

g1shin:
Ready for review again. I also left some follow-up comments :+1:

src/lib/stepper/stepper-vertical.html
kara: Yeah, I meant an internal directive for the shared functionality. If they have a lot in common, would be easier to update together.

g1shin:
Addressed comment regarding shared code between horizontal and vertical stepper. Ready for review :smiley:

src/cdk/stepper/step-icon.ts
```diff
@@ -0,0 +1,41 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive, Input} from '@angular/core';
+import {CdkStep} from './stepper';
+
+@Directive({
+  selector: 'cdkStepIcon'
+})
+export class CdkStepIcon {
```
mmalerba: I don't think this needs a cdk version, the icon is a material design specific piece of UI

src/cdk/stepper/step-label-container.ts
```diff
@@ -0,0 +1,28 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive, Input} from '@angular/core';
+import {CdkStep} from './stepper';
+
+@Directive({
+  selector: 'cdkStepLabelContainer'
+})
+export class CdkStepLabelContainer {
```
mmalerba: also don't see a need for a cdk version of this

src/cdk/stepper/stepper.ts
mmalerba: I think if its optional they should be able to skip it, not enter invalid input

src/lib/stepper/stepper-horizontal.html
```diff
@@ -8,22 +8,14 @@
...
-      </div>
-
-      <div class="mat-stepper-label">
-        <!-- If there is a label template, use it. -->
-        <ng-container *ngIf="step.stepLabel" [ngTemplateOutlet]="step.stepLabel.template">
-        </ng-container>
-        <!-- It there is no label template, fall back to the text label. -->
-        <div *ngIf="!step.stepLabel">{{step.label}}</div>
-      </div>
+      <md-step-icon [step]="step" [selected]="selectedIndex == i" [index]="i"></md-step-icon>
```
mmalerba: could we not just combine these 2 internal components (and their parent div) into a single `md-step-header` component?

g1shin:
Refactoring and optional step logic change done. Ready for review again :+1:

g1shin:
Changes made based on discussion. Ready for another review

src/lib/stepper/stepper-vertical.html
```diff
@@ -1,24 +1,19 @@
...
-
-    <div class="mat-stepper-label">
-      <!-- If there is a label template, use it. -->
-      <ng-container *ngIf="step.stepLabel"[ngTemplateOutlet]="step.stepLabel.template">
-      </ng-container>
-      <!-- It there is no label template, fall back to the text label. -->
-      <div *ngIf="!step.stepLabel">{{step.label}}</div>
-    </div>
-
+  <div class="mat-stepper-header" #stepperHeader (click)="step.select()"
```
mmalerba: you can remove this div and get access to the `md-step-header` element like this:

```ts
@ViewChildren(MdStepHeader, {read: ElementRef}) _stepHeaders;
```

(and same for horizontal)

src/lib/stepper/step-header.scss
```diff
@@ -0,0 +1,39 @@
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+  min-width: $mat-stepper-label-min-width;
+  vertical-align: middle;
+}
+
+.mat-step-text-label {
+  text-overflow: ellipsis;
+  overflow: hidden;
+}
```
mmalerba:
```scss
$mat-step-header-icon-size: 16px !default;

.mat-step-icon {
    display: flex;
    align-items: center;
    justify-content: center;
}

.mat-step-icon .mat-icon {
    font-size: $mat-step-header-icon-size;
    height: $mat-step-header-icon-size;
    width: $mat-step-header-icon-size;
}
```

g1shin:
Ready for review again :smiley:

src/lib/stepper/step-header.ts
```diff
@@ -0,0 +1,77 @@
...
+  encapsulation: ViewEncapsulation.None
+})
+export class MdStepHeader {
+  /** Icon for the given step. */
+  @Input()
+  icon: string;
+
+  /** Label of the given step. */
+  @Input()
+  get label() { return this._label; }
```
mmalerba: nit: don't need getter/setter for this now

src/lib/stepper/stepper-horizontal.html
```diff
@@ -1,29 +1,23 @@
...
+    <md-step-header  class="mat-horizontal-stepper-header"
+                     (click)="step.select()"
+                     (keydown)="_onKeydown($event)"
+                     [tabIndex]="_focusIndex == i ? 0 : -1"
+                     [id]="_getStepLabelId(i)"
+                     [attr.aria-controls]="_getStepContentId(i)"
+                     [attr.aria-selected]="selectedIndex == i"
+                     [index]="i"
+                     [icon]="_getIndicatorType(i)"
+                     [label]="step.stepLabel ? step.stepLabel : step.label"
```
mmalerba: nit: can shorten to `step.stepLabel || step.label`

src/lib/stepper/stepper-vertical.html
```diff
@@ -1,24 +1,19 @@
...
+  <md-step-header  class="mat-vertical-stepper-header"
+                   (click)="step.select()"
+                   (keydown)="_onKeydown($event)"
+                   [tabIndex]="_focusIndex == i ? 0 : -1"
+                   [id]="_getStepLabelId(i)"
+                   [attr.aria-controls]="_getStepContentId(i)"
+                   [attr.aria-selected]="selectedIndex == i"
+                   [index]="i"
+                   [icon]="_getIndicatorType(i)"
+                   [label]="step.stepLabel ? step.stepLabel : step.label"
```
mmalerba: nit: `step.stepLabel || step.label`

PR #6511, opened 2017-08-16
feat(dialog): expose backdrop clicks

willshowell:
Partially addresses https://github.com/angular/material2/issues/3251

### DIFF
--- a/src/lib/dialog/dialog.spec.ts
+++ b/src/lib/dialog/dialog.spec.ts
@@ -237,6 +237,30 @@ describe('MdDialog', () => {
     });
   }));
 
+  it('should emit the backdropClick stream when clicking on the overlay backdrop', async(() => {
+    const dialogRef = dialog.open(PizzaMsg, {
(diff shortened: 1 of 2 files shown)


src/lib/dialog/dialog.spec.ts
```diff
@@ -195,6 +195,29 @@ describe('MdDialog', () => {
...
+    let backdrop = overlayContainerElement.querySelector('.cdk-overlay-backdrop') as HTMLElement;
+
+    backdrop.click();
+    expect(spy).toHaveBeenCalledTimes(1);
+
+    viewContainerFixture.detectChanges();
+
+    viewContainerFixture.whenStable().then(() => {
+      backdrop.click();
+      expect(spy).toHaveBeenCalledTimes(1);
```
jelbourn: Should have a comment here that explains why the count is still one after another click

kara:
@willshowell Can you rebase?

PR #6519, opened 2017-08-17
fix(chip): mark onFocus as internal

willshowell:


jelbourn:
Adding `(focus)` to the `md-chip` element already works since the chip element is the thing that receives focus; why would this need to be an `@Output`?

willshowell:
@jelbourn tbh I didn't investigate, I just noticed it stood out in the api docs. Happy to repurpose this PR to removing it entirely.


Name | Description
-- | --
`@Input()`selected | Whether the chip is selected.
onFocus | Emitted when the chip is focused.
`@Output()`select | Emitted when the chip is selected.
`@Output()`deselect | Emitted when the chip is deselected.
`@Output()`destroy | Emitted when the chip is destroyed.

jelbourn:
Ah, I see. It should just be internal (`_onFocus`) since it's used by the chip-list but not meant to be a public API

PR #6530, opened 2017-08-17
Better documentation

megamindbrian:
Call out that just because a theme file is created does not necessarily mean it is connected to your app.

### DIFF
--- a/guides/theming.md
+++ b/guides/theming.md
@@ -50,7 +50,7 @@ Finally, if your app's content **is not** placed inside of a `md-sidenav-contain
 need to add the `mat-app-background` class to your wrapper element (for example the `body`). This
 ensures that the proper theme background is applied to your page.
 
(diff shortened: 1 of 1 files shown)


guides/theming.md
```diff
@@ -50,7 +50,7 @@ Finally, if your app's content **is not** placed inside of a `md-sidenav-contain
 need to add the `mat-app-background` class to your wrapper element (for example the `body`). This
 ensures that the proper theme background is applied to your page.
 
-### Defining a custom theme
+### Step 1: Defining a custom theme
```
jelbourn: Say "Step 1" doesn't really fit here because it's not in the context of anything that came before it. A better approach would be 
```markdown
### Defining a custom theme

#### Create an scss file for your theme
...
#### Include the generated styles in your application
...
```

jelbourn:

PR #6525, opened 2017-08-17
feat(sidenav): add responsive sidenav

mcparadip:
* Add responsive sidenav
* Close and open depending on width

Fix #1130

PR #6535, opened 2017-08-17
Responsive drawer

### LINKED ISSUE #1130: md-sidenav: make it automatically responsive
*Feature request*
Version: Angular Material 2.0.0-alpha.7 wax-umpire

I'd like to have an option where the sidenav can be responsive automatically, by having a breakpoint where the sidenav is shown by default when the width is higher than X. 

This was already implemented in Angular Material (1) as the `md-is-locked-open`: https://material.angularjs.org/latest/api/directive/mdSidenav

Is there any plan to have this? I could submit a PR if this is wanted.

mcparadip:
New version of #6525 

Fixes #1130

### DIFF
--- a/src/lib/sidenav/drawer.ts
+++ b/src/lib/sidenav/drawer.ts
@@ -404,9 +404,9 @@ export class MdDrawerContainer implements AfterContentInit {
         }
         this.open();
       }
-    }
+    };
     this._updateDrawerSubscription = startWith.call(resize, null).subscribe(this._updateDrawer);
-    
+
   }
 
   ngOnDestroy() {


mcparadip:
Fixing code style in progress..

src/lib/sidenav/drawer.ts
```diff
@@ -359,6 +383,33 @@ export class MdDrawerContainer implements AfterContentInit {
...
+      observableOf(null);
+
+    this._updateDrawer = startWith.call(resize, null).subscribe(() => {
+      if (this._element.nativeElement.offsetWidth < this.breakpointWidth) {
+        if (this.breakpointChangeMode) {
+          this._drawers.forEach(drawer => drawer.mode = 'over');
+        }
+        this.close();
+      }
+      if (this._element.nativeElement.offsetWidth > this.breakpointWidth) {
```
rafaelss95: `else` here sounds better.

mcparadip: Didn't notice that, fixing.

mcparadip:
@willshowell

src/lib/sidenav/drawer.ts
```diff
@@ -325,6 +328,27 @@ export class MdDrawerContainer implements AfterContentInit {
   /** Event emitted when the drawer backdrop is clicked. */
   @Output() backdropClick = new EventEmitter<void>();
 
+  /**
+   * Width of the container at which it breaks
+   * e.g., breakpointWidth="500"
+   */
+  @Input() breakpointWidth: number;
```
willshowell: Coerce this? See https://github.com/angular/material2/blob/master/src/lib/slider/slider.ts#L149-L164

src/lib/sidenav/drawer.ts
```diff
@@ -325,6 +328,27 @@ export class MdDrawerContainer implements AfterContentInit {
...
+    this._breakpointChangeMode = coerceBooleanProperty(value);
+  }
+  private _breakpointChangeMode: boolean = true;
+
+  /**
+   * Responsively toggle the drawer using the breakpoint.
+   * Note that this only works when the window is rezied.
+   * If you resize it some other way, call _updateDrawer().
+   */
+  _updateDrawer: Subscription | null = null;
```
willshowell: Maybe rename this to `_updateDrawerSubscription` and then pull out a method named `_updateDrawer` that can still manually be used? Or just rename it and update the comment.

src/lib/sidenav/drawer.ts
```diff
@@ -359,6 +383,32 @@ export class MdDrawerContainer implements AfterContentInit {
...
+        this.open();
+      }
+    });
+
+  }
+
+  ngOnDestroy() {
+    if (this._updateDrawer) {
+      this._updateDrawer.unsubscribe();
+      this._updateDrawer = null;
```
willshowell: I don't think setting to `null` is necessary since the component is being destroyed

src/lib/sidenav/drawer.ts
```diff
@@ -325,6 +328,27 @@ export class MdDrawerContainer implements AfterContentInit {
...
+   */
+  @Input() breakpointWidth: number;
+
+  /** Whether the drawer changes modes when collapsing. */
+  @Input()
+  get breakpointChangeMode(): boolean { return this._breakpointChangeMode; }
+  set breakpointChangeMode(value: boolean) {
+    this._breakpointChangeMode = coerceBooleanProperty(value);
+  }
+  private _breakpointChangeMode: boolean = true;
```
willshowell: Type should be implicit

```ts
private _breakpointChangeMode = true;
```

src/lib/sidenav/drawer.ts
```diff
@@ -359,6 +383,32 @@ export class MdDrawerContainer implements AfterContentInit {
         this._watchDrawerPosition(drawer);
       });
     });
+
+    const resize = typeof window !== 'undefined' ?
+      auditTime.call(fromEvent(window, 'resize'), 150) :
+      observableOf(null);
+
+    this._updateDrawer = startWith.call(resize, null).subscribe(() => {
+      if (this._element.nativeElement.offsetWidth < this.breakpointWidth) {
```
willshowell: I haven't checked, but this needs to be consistent with any other responsive behavior in this lib or in flex-layout (i.e. `<` vs `<=`)

mcparadip:
comments addressed

src/lib/sidenav/drawer.ts
```diff
@@ -328,26 +328,30 @@ export class MdDrawerContainer implements AfterContentInit {
...
+  get breakpointWidth() { return this._breakpointWidth; }
+  set breakpointWidth(v: number) {
+    this._breakpointWidth = coerceNumberProperty(v, this._breakpointWidth);
+  }
+  private _breakpointWidth: number;
 
   /** Whether the drawer changes modes when collapsing. */
   @Input()
-  get breakpointChangeMode(): boolean { return this._breakpointChangeMode; }
+  get breakpointChangeMode() { return this._breakpointChangeMode; }
```
rafaelss95: Just a quick question... why did you remove the return type here?

mcparadip: It wasn't on the number one, so I removed it here

src/lib/sidenav/drawer.ts
```diff
@@ -325,6 +328,31 @@ export class MdDrawerContainer implements AfterContentInit {
   /** Event emitted when the drawer backdrop is clicked. */
   @Output() backdropClick = new EventEmitter<void>();
 
+  /** Breakpoint width (in px) at which the sidenav collapses. */
+
+  @Input()
+  get breakpointWidth() { return this._breakpointWidth; }
```
mmalerba: @jelbourn have you added anything to cdk yet for dealing with breakpoints?

src/demo-app/sidenav/sidenav-demo.html
```diff
@@ -48,6 +48,18 @@
   </div>
 </md-sidenav-container>
 
+<h2>Responsive Drawer</h2>
+
+<md-sidenav-container class="demo-sidenav-container" breakpointWidth="600" breakpointChangeMode>
```
mmalerba: I'm not so sure about this API, it doesn't seem to give much flexibility. You can't control what modes it switches between and you can't disable the auto-close/open feature

mmalerba:
@willshowell the split between sidenav and drawer is to allow us to add features that make sense for a true global sidenav to the `md-sidenav` component (e.g. some fixed positioning features) and leave the `md-drawer` as something that can be used to create subsections of your app that have their own side "drawers". Currently they're pretty much the same, but they will diverge. This change seems like it belongs in the sidenav only since I don't think the feature makes sense for the drawer-style usage

jelbourn:
@mmalerba not yet; that will be @josephperrott's first task next week

In general I'd like to have one consistent API pattern for defining responsive behaviors for all components

josephperrott:
@mcparadip With the merge of #6772, an API is now available to manage watching breakpoints.  You should be able to rebase and use this to work on making the drawer responsive.

mcparadip:
Got it. Will work on it in a while

PR #6537, opened 2017-08-18
fix(paginator): add type="button" to next/prev buttons

mostwired:
Adding type="button" to previous/next buttons prevents submitting the form, if the paginator control is added to a form.


mostwired:
signed it

PR #6551, opened 2017-08-19
feat(progress-spinner): switch to css-based animation

crisbeto:
Reworks the progress spinner to use a CSS-based indeterminate animation. For IE and Edge where we can't pull of the animation reliably, it falls back to a non-spec animation.

Demo: https://material-cb7ec.firebaseapp.com/progress-spinner

### DIFF
--- a/src/lib/progress-spinner/progress-spinner.html
+++ b/src/lib/progress-spinner/progress-spinner.html
@@ -4,10 +4,11 @@
   element containing the SVG. `focusable="false"` prevents IE from allowing the user to
   tab into the SVG element.
 -->
+
 <svg
-  width="100"
-  height="100"
-  viewBox="0 0 100 100"
+  [style.width.px]="_elementSize"
+  [style.height.px]="_elementSize"
+  [attr.viewBox]="_viewBox"
   preserveAspectRatio="xMidYMid meet"
   focusable="false">
 
@@ -15,6 +16,9 @@
     cx="50%"
     cy="50%"
     [attr.r]="_circleRadius"
-    [style.stroke-dashoffset.px]="_getStrokeDashOffset()"
+    [style.animation-name]="'mat-progress-spinner-stroke-rotate-' + diameter"
+    [style.stroke-dashoffset.px]="_strokeDashOffset"
+    [style.stroke-dasharray.px]="_strokeCircumference"
+    [style.transform.rotate]="'360deg'"
     [style.stroke-width.px]="strokeWidth"></circle>
 </svg>
--- a/src/lib/progress-spinner/progress-spinner.html
+++ b/src/lib/progress-spinner/progress-spinner.html
@@ -4,8 +4,17 @@
   element containing the SVG. `focusable="false"` prevents IE from allowing the user to
   tab into the SVG element.
 -->
-<svg viewBox="0 0 100 100"
-     preserveAspectRatio="xMidYMid meet"
-     focusable="false">
-  <path #path [style.strokeWidth]="strokeWidth"></path>
+<svg
+  width="100"
+  height="100"
+  viewBox="0 0 100 100"
+  preserveAspectRatio="xMidYMid meet"
+  focusable="false">
+
+  <circle
+    cx="50%"
+    cy="50%"
+    [attr.r]="_circleRadius"
+    [style.stroke-dashoffset.px]="_getStrokeDashOffset()"
+    [style.stroke-width.px]="strokeWidth"></circle>
 </svg>
--- a/src/lib/progress-spinner/progress-spinner.scss
+++ b/src/lib/progress-spinner/progress-spinner.scss
@@ -5,9 +5,8 @@
 $mat-progress-spinner-stroke-rotate-fallback-duration: 10 * 1000ms !default;
 $mat-progress-spinner-stroke-rotate-fallback-ease: cubic-bezier(0.87, 0.03, 0.33, 1) !default;
 
-$_mat-progress-spinner-radius: 45px;
-$_mat-progress-spinner-circumference: $pi * $_mat-progress-spinner-radius * 2;
-
+$_mat-progress-spinner-default-radius: 45px;
+$_mat-progress-spinner-default-circumference: $pi * $_mat-progress-spinner-default-radius * 2;
 
 .mat-progress-spinner {
   display: block;
@@ -24,8 +23,6 @@ $_mat-progress-spinner-circumference: $pi * $_mat-progress-spinner-radius * 2;
 
   circle {
     fill: transparent;
-    stroke-dasharray: $_mat-progress-spinner-circumference;
-    stroke-dashoffset: $_mat-progress-spinner-circumference;
     transform-origin: center;
     transition: stroke-dashoffset 225ms linear;
   }
@@ -35,10 +32,11 @@ $_mat-progress-spinner-circumference: $pi * $_mat-progress-spinner-radius * 2;
         linear infinite;
 
     circle {
-      // Note: we multiply the duration by 8, because the animation is spread out in 8 stages.
-      animation: mat-progress-spinner-stroke-rotate $swift-ease-in-out-duration * 8
-          $ease-in-out-curve-function infinite;
       transition-property: stroke;
+      // Note: we multiply the duration by 8, because the animation is spread out in 8 stages.
+      animation-duration: $swift-ease-in-out-duration * 8;
+      animation-timing-function: $ease-in-out-curve-function;
+      animation-iteration-count: infinite;
     }
   }
 
@@ -49,7 +47,6 @@ $_mat-progress-spinner-circumference: $pi * $_mat-progress-spinner-radius * 2;
                infinite;
 
     circle {
-      stroke-dashoffset: (1 - 0.8) * $_mat-progress-spinner-circumference;
       transition-property: stroke;
     }
   }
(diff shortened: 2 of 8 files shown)


src/lib/progress-spinner/progress-spinner.scss
```diff
@@ -55,13 +61,44 @@ $mat-progress-spinner-viewport-size: 100px !default;
...
+    75%     { stroke-dashoffset: $start;  transform: rotateX(180deg) rotate(251.5deg); }
+
+    75.01%  { stroke-dashoffset: $start;  transform: rotate(90deg); }
+    87.5%   { stroke-dashoffset: $end;    transform: rotate(90deg); }
+    87.51%  { stroke-dashoffset: $end;    transform: rotateX(180deg) rotate(341.5deg); }
+    100%    { stroke-dashoffset: $start;  transform: rotateX(180deg) rotate(341.5deg); }
+    // stylelint-enable
+  }
+
+  @keyframes mat-progress-spinner-stroke-rotate-fallback {
```
jelbourn: Add comment describing this animation?
```scss
// For IE11 and Edge, we fall back to simply rotating the spinner because
// animating stroke-dashoffset is not supported. The fallback uses multiple
// iterations to vary where the spin "lands".
```

crisbeto: Done. FWIW, animating `stroke-dashoffset` is supported in Edge, but rotation the circle with a CSS `transform` isn't.

kara:
@crisbeto @jelbourn This PR breaks internal tests because some projects are relying on being able to set `[style.width.px]` and `[style.height.px]` on the spinner element. Thoughts?

crisbeto:
The reason I removed the `width` and `height` from the CSS was because we need control over the height due to the `strokeWidth` option. If the stroke width is greater than the default `10px` it will overflow the element, which is why I ended up making it smaller to compensate. My guess is that the option isn't widely used anyway so we could get rid of it and let users override it with CSS. If they decide to override it, they can also tweak the margins/padding/width or height depending on their setup.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -84,189 +88,115 @@ export const _MdProgressSpinnerMixinBase = mixinColor(MdProgressSpinnerBase, 'pr
...
-  /**
-   * Values for aria max and min are only defined as numbers when in a determinate mode.  We do this
-   * because voiceover does not report the progress indicator as indeterminate if the aria min
-   * and/or max value are number values.
-   */
-  get _ariaValueMin() {
-    return this.mode == 'determinate' ? 0 : null;
-  }
+  /** Tracks diameters of existing instances to de-dupe generated styles. */
+  static diameters = new Set<number>([100]);
```
jelbourn: Comment about `100` being the default diameter?

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -84,189 +88,115 @@ export const _MdProgressSpinnerMixinBase = mixinColor(MdProgressSpinnerBase, 'pr
...
-  /** @docs-private */
-  get interdeterminateInterval() {
-    return this._interdeterminateInterval;
-  }
-  /** @docs-private */
-  set interdeterminateInterval(interval: number | null) {
-    if (this._interdeterminateInterval) {
-      clearInterval(this._interdeterminateInterval);
+  set diameter(size: number) {
+    this._setDiameterClass(size);
```
jelbourn: Break the content of this setter into its own function?

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -79,31 +133,71 @@ export class MdProgressSpinner extends _MdProgressSpinnerMixinBase implements Ca
...
+  get _strokeDashOffset() {
     if (this.mode === 'determinate') {
-      return 2 * Math.PI * this._circleRadius * (100 - this._value) / 100;
+      return this._strokeCircumference * (100 - this._value) / 100;
     }
 
     return null;
   }
+
+  private _setDiameterClass(size: number): void {
```
jelbourn: JsDoc descriptions for these functions?

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -79,31 +133,71 @@ export class MdProgressSpinner extends _MdProgressSpinnerMixinBase implements Ca
...
-      this._elementSize = this._baseSize + Math.max(this.strokeWidth - this._baseStrokeWidth, 0);
+    if (changes.strokeWidth || changes.diameter) {
+      this._elementSize =
+          this._diameter + Math.max(this.strokeWidth - this._baseStrokeWidth, 0);
     }
   }
 
-  _getStrokeDashOffset() {
+  get _circleRadius() {
+    return this._diameter / 2 - 5;
```
crisbeto: Where does the `5` come from. I assume it's half the stroke width?

src/lib/progress-spinner/progress-spinner.scss
```diff
@@ -55,13 +57,47 @@ $mat-progress-spinner-viewport-size: 100px !default;
...
-  75%      { transform: rotate( 810deg); }
-  87.5%    { transform: rotate( 945deg); }
-  100%     { transform: rotate(1080deg); }
+
+@at-root {
+  $start: (1 - 0.05) * $_mat-progress-spinner-default-circumference; // start the animation at 5%
+  $end: (1 - 0.8) * $_mat-progress-spinner-default-circumference; // end the animation at 80%
+  $fallback-iterations: 4;
+
+  @keyframes mat-progress-spinner-stroke-rotate-100 {
```
crisbeto: Should we keep this one around, considering that the keyframes are being generated at runtime?

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -7,62 +7,64 @@
...
 // Boilerplate for applying mixins to MdProgressSpinner.
 /** @docs-private */
 export class MdProgressSpinnerBase {
   constructor(public _renderer: Renderer2, public _elementRef: ElementRef) {}
 }
 export const _MdProgressSpinnerMixinBase = mixinColor(MdProgressSpinnerBase, 'primary');
 
+/* tslint:disable:max-line-length */
+const INDETERMINATE_ANIMATION_TEMPLATE = `
+  .mat-progress-spinner-DIAMETER.mat-progress-spinner-indeterminate-animation[mode="indeterminate"] circle {
```
crisbeto: Rather than constructing this selector, it might be better to just assign the `animation-name: mat-progress-spinner-stroke-rotate-DIAMETER` as an inline style. That way people can still override the timing and duration through CSS.

src/lib/progress-spinner/progress-spinner.scss
kara: We won't generate a style tag at runtime for the default diameter - only for custom diameters (the set starts with 100 included).

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -84,189 +88,115 @@ export const _MdProgressSpinnerMixinBase = mixinColor(MdProgressSpinnerBase, 'pr
...
-    return this._interdeterminateInterval;
-  }
-  /** @docs-private */
-  set interdeterminateInterval(interval: number | null) {
-    if (this._interdeterminateInterval) {
-      clearInterval(this._interdeterminateInterval);
+  set diameter(size: number) {
+    this._setDiameterClass(size);
+    this._diameter = size;
+    if (!MdProgressSpinner.diameters.has(this.diameter)) {
```
crisbeto: Since we don't support the spec animation on IE and Edge, we shouldn't append the keyframes either.

kara: Good call

src/lib/progress-spinner/progress-spinner.ts
kara: I believe that's what it's supposed to be, yes; fixing.

src/lib/progress-spinner/progress-spinner.ts
kara: Oh hey, that's a great idea. Should simplify this a lot.

src/lib/progress-spinner/progress-spinner.html
```diff
@@ -4,8 +4,21 @@
...
+  focusable="false">
+
+  <circle
+    cx="50%"
+    cy="50%"
+    [attr.r]="_circleRadius"
+    [style.animation-name]="'mat-progress-spinner-stroke-rotate-' + diameter"
+    [style.stroke-dashoffset.px]="_strokeDashOffset"
+    [style.stroke-dasharray.px]="_strokeCircumference"
+    [style.transform.rotate]="'360deg'"
```
crisbeto: Why do we need this? It doesn't look like it does anything.

kara: You know, I thought that was from your commit, but now I have no idea. Maybe it's a weird rebase error? Don't remember adding it.

crisbeto: Hmm, I haven't pulled in your changes into my local branch yet. I tried checking it out and it's not in there.

kara: I must have been testing something at some point? I'll remove it, in any case.

crisbeto:
@kara this one needs a rebase.

pabx06:
@kara i am using the latest beta.
Changing the width & height style i still get 100px*100px spinner ... 
i am using the spinner as 

`
<mat-form-field ...>
   <input...>
   <span matSuffix style="height: 18px" align="end" *ngIf="searching">
       <mat-progress-spinner mode="indeterminate" ng-disabled="true" style="height: 18px" align="end"> </mat-progress-spinner>
</mat-form-field>
`

on debounced by 300ms  keystroke

any idea how i could set the spinner to fit the text input field ?

PR #6558, opened 2017-08-20
feat(dialog): add ariaLabel and focusOnOpen config options

crisbeto:
Based on the discussion on https://github.com/angular/material2/pull/6360#discussion_r133581835, these changes add the ability to set the `aria-label` of a dialog, as well as the element that should be focus when the dialog is opened.

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -81,6 +81,11 @@ export class MatDialogConfig<D = any> {
   /** ID of the element that describes the dialog.  */
   ariaDescribedBy?: string | null = null;
 
+  /** Aria label to assign to the dialog element */
+  ariaLabel?: string | null = null;
+
+  /** Whether the dialog should focus the first focusable element on open. */
+  autoFocus?: boolean = true;
 
   // TODO(jelbourn): add configuration for lifecycle hooks, ARIA labelling.
 }
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -71,7 +71,8 @@ export function throwMatDialogContentAlreadyAttachedError() {
     'class': 'mat-dialog-container',
     'tabindex': '-1',
     '[attr.role]': '_config?.role',
-    '[attr.aria-labelledby]': '_ariaLabelledBy',
+    '[attr.aria-labelledby]': '_config?.ariaLabel ? null : _ariaLabelledBy',
+    '[attr.aria-label]': '_config?.ariaLabel',
     '[attr.aria-describedby]': '_config?.ariaDescribedBy || null',
     '[@slideDialog]': '_state',
     '(@slideDialog.start)': '_onAnimationStart($event)',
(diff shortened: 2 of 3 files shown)


src/lib/dialog/dialog-config.ts
```diff
@@ -69,5 +69,14 @@ export class MdDialogConfig {
...
   ariaDescribedBy?: string | null = null;
 
+  /** Aria label to assign to the dialog element */
+  ariaLabel?: string | null = null;
+
+  /**
+   * Selector for an element to be focused when the dialog is opened. If omitted or the
+   * element is not found, the dialog will focus the first focusable element.
+   */
+  focusOnOpen?: string | null = null;
```
jelbourn: With this API, it's not immediately clear what selector you'd set here to focus the root dialog element.

crisbeto: Hmm not sure how we could remedy that. Perhaps supporting `:host` to align with Angular?

jelbourn: Is there a strong reason to use a selector at all here since the focus trap already supports marking where you want the focus to start? The main thing that's missing is being able to focus the root element, which could just be a boolean.

crisbeto: My reasoning to go for a selector instead of the boolean was to make it more flexible so we don't have to tweak it if somebody wants to focus something different from the container. E.g. this allows for the selector to be constructed at runtime, whereas the focus trap attribute needs to be set in advance.

jelbourn: Maybe two properties, then? Something like `focusChild` for selector and `focusRoot`?

crisbeto: Hmm I'm iffy about having two flags that do more or less the same. What about typing it something like `'root' | string | null`. That way it should show up in docs and IDEs that `root` is a special option.

jelbourn: I'm not a big fan of magic-string APIs. I'd rather go with `boolean | string | null`, where the boolean means the host element. 

What do you think of making this PR just add the `ariaLabel` and the `focusOnOpen` for the root element while leaving the content focusing as a responsibility of the focus trap?

crisbeto: Sounds good.

crisbeto:
Rebased and reworked the PR so the `focusOnOpen` is a boolean @jelbourn. Users can still control what element gets focused first via the `cdk-focus-initial` attribute.

src/lib/dialog/dialog-config.ts
```diff
@@ -81,6 +81,11 @@ export class MatDialogConfig<D = any> {
   /** ID of the element that describes the dialog.  */
   ariaDescribedBy?: string | null = null;
 
+  /** Aria label to assign to the dialog element */
+  ariaLabel?: string | null = null;
+
+  /** Whether the dialog should focus the first focusable element on open. */
+  focusOnOpen?: boolean = true;
```
jelbourn: Coming back again- maybe `autoFocus` instead? That's what I went with for the focus trap directive

src/lib/dialog/dialog-config.ts
```diff
@@ -81,6 +81,11 @@ export class MatDialogConfig<D = any> {
   /** ID of the element that describes the dialog.  */
   ariaDescribedBy?: string | null = null;
 
+  /** Aria label to assign to the dialog element */
```
jelbourn: A thought occurred to me (for a follow-up PR if you agree): to make the focus go to the root dialog element, we can just expose a `tabIndex` property on the dialog config. If it's `0`, then it's the first focusable element. Thoughts?

crisbeto: I don't think that would work since the interactivity checker starts looking from the children. Also we always focus the dialog anyway in order to prevent the user from opening multiple dialogs while animating.

crisbeto:

Issue #6456, opened 2017-08-14
bug(tab): tab header label  readability issue on small devices 


m98:
[*] bug

#### What is the expected behavior?
On smaller devices, tab header content should be readable/visible.
_(Note: the tab with **Our** label should be **Our large tab title**)_
![screenshot from 2017-08-14 18-23-46](https://user-images.githubusercontent.com/7496103/29274629-bdcc05e8-811d-11e7-9324-5181b91bbf8c.png)

#### What is the current behavior?
Instead of showing the content of the tab header, we will see some parts of the label.
Resize the browser to a smaller width, and the tab label is cut and we can't read the content of the label.
By the way, we already have `>` and `<` buttons beside of the tab header, we can set the size of all tab header labels to the largest text width, and skip cutting words ( or `ellipsis`).

#### What are the steps to reproduce?
https://plnkr.co/edit/cZBCdICrA5gKwDmIQH3c?p=preview

#### What is the use-case or motivation for changing an existing behavior?
smaller devices should be able to use tabs as well.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
```
@angular/cli: 1.2.6
node: 7.10.0
os: linux x64
@angular/animations: 4.3.2
@angular/cdk: 2.0.0-beta.8
@angular/common: 4.3.2
@angular/compiler: 4.3.2
@angular/core: 4.3.2
@angular/forms: 4.3.2
@angular/http: 4.3.2
@angular/material: 2.0.0-beta.8
@angular/platform-browser: 4.3.2
@angular/platform-browser-dynamic: 4.3.2
@angular/router: 4.3.2
@angular/cli: 1.2.6
@angular/compiler-cli: 4.3.2
@angular/language-service: 4.3.2

OS: Fedora 25
Browser:  (Version 59.0.3071.86)
```

#### Is there anything else we should know?
Here is what material.io says about tabs ([here](https://material.io/guidelines/components/tabs.html#tabs-content)):
> A long label exceeding the maximum width of a tab may wrap to a second line before being truncated.
> Wrap labels before truncating them. Truncating labels too early can impede comprehension.

amcdnl:
Looks to be fixed on latest

![img](https://content.screencast.com/users/amcdaniel22/folders/Snagit/media/ebd55f2c-1070-4181-9c69-f8e680ceec5b/2017-09-09_11-39-21.png)

PR #6580, opened 2017-08-21
feat(tabs): tabs wrapping #6456

amcdnl:
Addresses #6456 ... Implementation WIP

img

### DIFF
--- a/src/lib/tabs/tab-group.scss
+++ b/src/lib/tabs/tab-group.scss
@@ -17,6 +17,10 @@
   position: relative;
 }
 
+.md-tab-wrap {
+  font-size: .8rem;
+}
+
 @media ($mat-xsmall) {
   .mat-tab-label {
     min-width: 72px;
--- a/src/lib/tabs/tab-truncate.ts
+++ b/src/lib/tabs/tab-truncate.ts
@@ -1,0 +1,32 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({ name: 'mdTruncate' })
+export class MdTabLabelTruncate implements PipeTransform {
+  transform(value: string, limit: number = 70, trail: String = '…'): string {
+    if (value.length > limit) {
+      const split = value.split(' ');
+      if (split.length === 1) {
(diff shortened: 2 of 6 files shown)


src/lib/tabs/tab-group.scss
```diff
@@ -17,6 +17,10 @@
   position: relative;
 }
 
+.md-tab-wrap {
+  font-size: .8rem;
```
jelbourn: FYI we typically want to avoid `rem` since we have no control over the root font-size.

src/lib/tabs/tab-truncate.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({ name: 'mdTruncate' })
+export class MdTabLabelTruncate implements PipeTransform {
+  transform(value: string, limit: number = 70, trail: String = '…'): string {
```
jelbourn: Thinking about this more in depth, I think that overall this approach has too many i18n complications. 
* Splitting on whitespace is fine for English, but not as good for, say, German ([longer words](https://www.youtube.com/watch?v=cfzcJj4-JuU)) or Chinese (no spaces). 
* Splitting words could change the label's meaning (more of a problem for locales other than English)

src/lib/tabs/tab-group.scss
amcdnl: Would you prefer percentage?

src/lib/tabs/tab-truncate.ts
amcdnl: Good points. I tried to be a little smarter than just doing a string length so that the label might have a better meaning if it optimized for words.

Would you like me to change the algorithm to handle longer words better or ditch this approach altogether and just do length?

jelbourn: I'm more thinking we shelve this until next quarter and I can consult with our internal i18n experts

amcdnl: No worries, should I make PR for the `max-width: 264px;` css I added here to make it adhere to the spec?

jelbourn: If we're not going to do the truncation I'd rather put off changing the `max-width` as well- that'll be a breaking change that would be time consuming to deal with now

amcdnl:
Closing this per feedback above.

PR #6588, opened 2017-08-22
fix: deprecate CoreModule

jelbourn:
Since most of this functionality has been moved into `@angular/cdk`, it
no longer makes sense to blob it all together in one NgModule inside
`@angular/material`.


rafaelss95:
I think there's typo: `CoreMore` -> `CoreModule`

PR #6589, opened 2017-08-22
fix(grid-list): revert incorrect use of mat-fill mixin

jelbourn:


willshowell:
I believe https://github.com/angular/material2/pull/6578 covers this

Issue #6552, opened 2017-08-19
No Provider for MdTooltip


datrinh:
#### Bug, feature request, or proposal:
When I inject the MdTooltip service into a component I get a `No Provider for MdTooltip` error.
All other material components I use are working fine. Tooltips work fine too except for the injecting.

#### What is the expected behavior?
No runtime error.

#### What is the current behavior?
`No Provider for MdTooltip`

#### What are the steps to reproduce?
Create a new ng-cli app and inject MdTooltip.
Example plnkr: http://plnkr.co/edit/UIzqoHLPqWMk2W0H5pRj?p=preview

#### What is the use-case or motivation for changing an existing behavior?
Using MdTooltip service functions such as `show()`

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
@angular/cli: 1.3.1
node: 7.6.0
os: darwin x64
@angular/animations: 4.3.5
@angular/cdk: 2.0.0-beta.8
@angular/common: 4.3.5
@angular/compiler: 4.3.5
@angular/core: 4.3.5
@angular/flex-layout: 2.0.0-rc.1
@angular/forms: 4.3.5
@angular/http: 4.3.5
@angular/material: 2.0.0-beta.8
@angular/platform-browser: 4.3.5
@angular/platform-browser-dynamic: 4.3.5
@angular/router: 4.3.5
@angular/cli: 1.3.1
@angular/compiler-cli: 4.3.5
@angular/language-service: 4.3.5

crisbeto:
That's because `MdTooltip` is a component, not a service. You can only inject it into a component or child of a component that uses `mdTooltip`.

datrinh:
How would I go about programmatically showing or hiding a tooltip then? Is there some example online showing it?

jymdman:
Modified plunker: http://plnkr.co/edit/UfCu0W0YO6ysarSszKyn?p=preview

rafaelss95:
To get an instance of `MdTooltip` in your component you could use `ViewChild` and then call the methods `show` and `hide` to handle its visibility.

joejordanbrown:
@datrinh 

It already activates using tap via touch device. You can't use the `click` event because of the interaction method which explains below.

You can view the demo code for how to programmatically trigger the tooltip [here](https://github.com/angular/material2/blob/master/src/demo-app/tooltip/tooltip-demo.html) or using the ViewChild example in @rafaelss95 plunker.

--------------------------

[...]

PR #6590, opened 2017-08-22
fix(tooltip): closing immediately when triggered on click

crisbeto:
Fixes the tooltip being closed immediately if it is opened as a result of a click. It seems like the logic that was supposed to handle this in master fires before the event has had the chance to bubble up to the body. These changes switch to relying on Angular's animation events to disable the body click.

Relates to #6552.


mmalerba:
please rebase

PR #6603, opened 2017-08-23
compat(table): add mat-compatible rows

andrewseguin:
Adds support for using `matHeaderRowDef` and `matRowDef` - currently doesn't work since columns aren't recognized.

### DIFF
--- a/src/lib/table/row.ts
+++ b/src/lib/table/row.ts
@@ -26,11 +26,19 @@ export const _MdRow = CdkRow;
  * Captures the header row's template and other header properties such as the columns to display.
  */
 @Directive({
-  selector: '[mdHeaderRowDef], [matHeaderRowDef]',
+  selector: '[mdHeaderRowDef]',
   providers: [{provide: CdkHeaderRowDef, useExisting: MdHeaderRowDef}],
   inputs: ['columns: mdHeaderRowDef'],
 })
 export class MdHeaderRowDef extends _MdHeaderRowDef { }
+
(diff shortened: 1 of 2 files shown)


andrewseguin:
At first I had this which worked on the demo app:

```
inputs: [
  'columns: matHeaderRowDef',
  'columns: mdHeaderRowDef'
]
```

But the tests gave me this:
`Can't bind to 'mdHeaderRowDef' since it isn't a known property of 'md-header-row'.`

Figured this was an easier approach for now, unless I'm making a silly mistake on the inputs

src/lib/table/row.ts
```diff
@@ -26,23 +26,39 @@ export const _MdRow = CdkRow;
...
 })
 export class MdHeaderRowDef extends _MdHeaderRowDef { }
 
+/** Mat-compatible version of MdHeaderRowDef */
+@Directive({
+  selector: '[matHeaderRowDef]',
+  providers: [{provide: CdkHeaderRowDef, useExisting: MatHeaderRowDef}],
+  inputs: ['columns: matHeaderRowDef'],
+})
+export class MatHeaderRowDef extends _MdHeaderRowDef { }
```
jelbourn: You could alternatively do

```ts
@Input('matHeaderRowDef')
get _whatever() { return this.columns; }
set _whatever(value) {  this.columns = value; }
```

andrewseguin:
Didn't play nice, this approach is easiest for now

PR #6604, opened 2017-08-23
compat(sort): remove md prefix from sort change event

andrewseguin:
BREAKING CHANGE: The sort event is renamed from `mdSortChange` to `sortChange`

### DIFF
--- a/src/demo-app/table/person-data-source.ts
+++ b/src/demo-app/table/person-data-source.ts
@@ -15,7 +15,7 @@ export class PersonDataSource extends DataSource<any> {
   connect(): Observable<UserData[]> {
     const displayDataChanges = [
       this._paginator.page,
-      this._sort.mdSortChange,
+      this._sort.sortChange,
       this._peopleDatabase.dataChange
     ];
     return Observable.merge(...displayDataChanges).map(() => {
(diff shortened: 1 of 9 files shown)


rafaelss95:
I'm not sure that's really related to this PR, but I noticed some inconsistency in `MdPaginator` and `MdSort` _change_ events.

In many components, like `MdCheckbox` you have a class named `MdCheckboxChange` and you do like that:

`Declaration class...`

```
let event = new MdCheckboxChange();
...
this.change.emit(event);

```
It will return an instance of `MdCheckboxChange`, but in `MdPaginator` and `MdSort`, besides the fact that the events aren't prefixed by `Md` and suffixed by `Change`, it just returns a plain object.

References:

[MdCheckboxChange `class`](https://github.com/angular/material2/blob/master/src/lib/checkbox/checkbox.ts#L61)

[Change event](https://github.com/angular/material2/blob/master/src/lib/checkbox/checkbox.ts#L314)

andrewseguin:
We'll be better at defining our naming, but the rule is that this should actually have the mdSort- prefix because it is a directive (notice all the inputs have mdSort as the prefix). Components like checkbox and paginator do not have the prefix. Closing in favor of a different approach to support mat-compatible code

PR #6611, opened 2017-08-23
compat(sort): add mat compatible sort event

andrewseguin:
BREAKING CHANGE: Renaming mdSort's sort change stream property from `mdSortChange` to `sortChange`.

### DIFF
--- a/src/lib/sort/sort.ts
+++ b/src/lib/sort/sort.ts
@@ -72,7 +72,7 @@ export class MdSort {
   set _matSortDisableClear(v) { this.disableClear = v; }
 
   /** Event emitted when the user changes either the active sort or sort direction. */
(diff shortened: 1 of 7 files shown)


src/lib/sort/sort.ts
```diff
@@ -72,7 +72,10 @@ export class MdSort {
   set _matSortDisableClear(v) { this.disableClear = v; }
 
   /** Event emitted when the user changes either the active sort or sort direction. */
-  @Output() mdSortChange = new EventEmitter<Sort>();
+  @Output('mdSortChange') sortChange = new EventEmitter<Sort>();
```
jelbourn: Also mark this as `readonly`?

andrewseguin:
Was there a caretaker note for this PR?

Issue #6465, opened 2017-08-14
Set tabindex in custom component


BobBDE:
#### Bug, feature request, or proposal:
Feature

#### What is the expected behavior?
Be able to set the tabindex to customs components (like md-slide-toggle)

#### What is the current behavior?
The tabindex are ignored by the customs components and thus the automatic tabindex is used. 

#### What are the steps to reproduce?
Here is a simple plunkr that show the problem. Focus the first toggle button and click tab. The last one should be focused as it has the property tabindex set to "2". 
http://plnkr.co/edit/V4q7uy?p=preview

#### What is the use-case or motivation for changing an existing behavior?
Be able to change the tabindex or remove an element from tabindex.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
All

#### Is there anything else we should know?
No.

jelbourn:
We should accept a `tabIndex` for each of the interactive components. We currently do this only inconsistently.

PR #6613, opened 2017-08-23
fix(slide-toggle): support native tabindex attribute

DevVersion:
Currently the slide-toggle only allows changing the tabIndex using the `tabIndex` binding. Using the native `tabindex` attribute doesn't have any affect. With this change the native tabindex property will be respected.

**Note**: The `tabIndex` mixin is more like a proposal to ensure that the tabIndex property will be handled consistent across Material components.

References #6465

### DIFF
--- a/src/lib/core/common-behaviors/tabindex.ts
+++ b/src/lib/core/common-behaviors/tabindex.ts
@@ -1,0 +1,35 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Constructor} from './constructor';
+import {CanDisable} from './disabled';
+
+/** @docs-private */
+export interface CanTabIndex {
+  tabIndex: number;
+}
+
+/** Mixin to augment a directive with a `tabIndex` property. */
+export function mixinTabIndex<T extends Constructor<CanDisable>>(base: T, defaultTabIndex = 0)
+    : Constructor<CanTabIndex> & T {
+  return class extends base {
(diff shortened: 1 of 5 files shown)


src/lib/core/common-behaviors/tabindex.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Constructor} from './constructor';
+import {CanDisable} from './disabled';
+
+/** @docs-private */
+export interface CanTabIndex {
```
jelbourn: Maybe `CanRecieveFocus`?

DevVersion: Addressed the feedback discussed per Chat.

src/lib/core/common-behaviors/tabindex.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/** Mixin to augment a directive with a `tabIndex` property. */
+export function mixinTabIndex<T extends Constructor<CanDisable>>(base: T, defaultTabIndex = 0)
+    : Constructor<HasTabIndex> & T {
+  return class extends base {
+    private _tabIndex: number;
+
+    get tabIndex(): number { return this.disabled ? -1 : this._tabIndex || defaultTabIndex; }
+    set tabIndex(value: number) {
+      if (typeof value !== 'undefined') {
```
jelbourn: Why do nothing for `undefined`?

DevVersion: Yeah, that doesn't really have any affect with the fallback. I've updated the getter/setter to better handle the fall back (also with a comment)

Issue #6585, opened 2017-08-22
custom input validator function example contains syntax error and is unclear


michahell:
the following example, here:
https://material.angular.io/components/input/overview
located under **Custom Error Matcher** contains one too many right brackets
and on top of that, is unclear:

```javascript
function myErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm): boolean {
  // Error when invalid control is dirty, touched, or submitted
  const isSubmitted = form && form.submitted;
  return !!(control.invalid && (control.dirty || control.touched || isSubmitted)));
}
```

why not just use:
```
return control.invalid && (control.dirty || control.touched || isSubmitted);
```

?

willshowell:
Good catch on the extra right parentheses.

For the boolean coercion, that example was based on the existing default matcher. It originates from https://github.com/angular/material2/pull/3560/commits/42dab5e5aa1aa1e4ebed550672c85dc6e713009f#r106483976.

michahell:
But why is coercion necessary? does `&&` not already do that ? or is there some javascript [wat](https://www.destroyallsoftware.com/talks/wat) involved if any of the `control.dirty` or other properties are not booleans? if so, `wat`.

willshowell:
Yeah I think in this case it is not actually necessary (maybe just for safety?). `&&` uses the last value, but as long as `dirty`, `touched`, and `submitted` are all boolean, there should be no need to coerce it.

```
> 1 && true
true

> true && 1
1

> !!(true && 1)
true
```

michahell:
A case of `wat`, it was, then :) I unconsciously knew this about `&&`, but only now realise that coercing might sometimes be necessary, thanks!

PR #6614, opened 2017-08-23
docs(input): remove extra parens and fix nits

willshowell:


src/lib/input/input.md
```diff
@@ -14,7 +14,7 @@ elements inside `md-form-field` as well. This includes Angular directives such a
 
 The only limitations are that the `type` attribute can only be one of the values supported by
 `mdInput` and the native element cannot specify a `placeholder` attribute if the `md-form-field`
-also contains a `md-placeholder` element.
+also contains an `md-placeholder` element.
```
rafaelss95: `a` was correct, no?

src/lib/input/input.md
```diff
@@ -129,18 +129,18 @@ the error messages.
 function myErrorStateMatcher(control: FormControl, form: FormGroupDirective | NgForm): boolean {
   // Error when invalid control is dirty, touched, or submitted
   const isSubmitted = form && form.submitted;
-  return !!(control.invalid && (control.dirty || control.touched || isSubmitted)));
+  return !!(control.invalid && (control.dirty || control.touched || isSubmitted));
```
rafaelss95: The _coercion_ here seems to be unnecessary since all fields are boolean.

src/lib/input/input.md
willshowell: Arguably haha. I'm basing it on the pronunciation of "md" sounding like "em dee". See [APA blog](http://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html) referencing the subject.

src/lib/input/input.md
willshowell: Yeah it is unnecessary, but it's based on the default errorStateMatcher which originates from https://github.com/angular/material2/pull/3560#discussion_r106483976.

Happy to change it if coercion is unnecessary, but it does add a little type safety. A little more discussion in https://github.com/angular/material2/issues/6585

jelbourn: It has an effect because `isSubmitted` is potentially `undefined` (if `form` is `undefined`)

PR #6615, opened 2017-08-23
refactor(overlay): allow for object to be passed to the OverlayState constructor

crisbeto:
Allows for a config object to be passed to the `OverlayState` constructor. This makes the longer configurations a bit more convenient to use.

### DIFF
--- a/src/cdk/overlay/overlay-directives.ts
+++ b/src/cdk/overlay/overlay-directives.ts
@@ -272,7 +272,12 @@ export class ConnectedOverlayDirective implements OnDestroy, OnChanges {
 
   /** Builds the overlay config based on the directive's inputs */
   private _buildConfig(): OverlayState {
-    let overlayConfig = new OverlayState();
+    const positionStrategy = this._position = this._createPositionStrategy();
+    const overlayConfig = new OverlayState({
+      positionStrategy,
+      scrollStrategy: this.scrollStrategy,
+      hasBackdrop: this.hasBackdrop
+    });
 
     if (this.width || this.width === 0) {
       overlayConfig.width = this.width;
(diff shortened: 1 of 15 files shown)


src/cdk/overlay/overlay-state.ts
```diff
@@ -53,6 +53,12 @@ export class OverlayState {
   /** The direction of the text in the overlay panel. */
   direction?: Direction = 'ltr';
 
+  constructor(state?: OverlayState) {
+    if (state) {
```
jelbourn: Could you do this?
```ts
extendObject(this, state);
```
(from `core/util`, handles `null` / `undefined`)

src/cdk/overlay/overlay-directives.ts
```diff
@@ -272,7 +272,12 @@ export class ConnectedOverlayDirective implements OnDestroy, OnChanges {
 
   /** Builds the overlay config based on the directive's inputs */
   private _buildConfig(): OverlayState {
-    let overlayConfig = new OverlayState();
+    const positionStrategy = this._position = this._createPositionStrategy();
+    const overlayConfig = new OverlayState({
```
jelbourn: Since all of the options are optional now, could you just change most places to, e.g.,
```ts
const overlayConfig = {
  positionStrategy,
  scrollStrategy: this.scrollStrategy,
  hasBackdrop: this.hasBackdrop,
};
```
?

src/cdk/overlay/overlay.spec.ts
```diff
@@ -153,10 +152,7 @@ describe('Overlay', () => {
   });
 
   it('should emit the attachment event after everything is added to the DOM', () => {
-    let state = new OverlayState();
-
-    state.hasBackdrop = true;
-
+    let state = new OverlayState({ hasBackdrop: true });
```
jelbourn: Omit spaces in braces

src/cdk/overlay/overlay-state.ts
crisbeto: I didn't use `extendObject`, because it would introduce a dependency between Material and the CDK.

src/cdk/overlay/overlay-directives.ts
crisbeto: That's what I did. The `position` in this case is different, because we want to save a reference to it before we pass it to the `OverlayRef`.

src/cdk/overlay/overlay-state.ts
jelbourn: Ah, I forgot that was in material.

src/cdk/overlay/overlay-directives.ts
jelbourn: I meant omitting the `new OverlayState(` part and just having the object literal

crisbeto: Ah, gotcha. Instantiating the class lets us apply the default options for things like the direction, backdrop class and scroll strategy.

PR #6618, opened 2017-08-24
feat(autocomplete): add input for adding classes to the panel

### LINKED ISSUE #4196: Feature request for autocomplete - more control over panel width
Provide an option to control the width of the panel, perhaps via css:
<md-autocomplete #auto="mdAutocomplete" panel-class="my-autocomplete-css">

Or if it could just expand to fit the dropdown content that would be nice.

Allow the autocomplete dropdown panel to be a different width than the associated input field.

The dropdown panel is the same width as the associated input field, which cuts off long values. There doesn't appear to be a way to change the width. A class or style on the md-autocomplete tag is ignored. Wrapping the options in a div with a wider style simply adds a scrollbar to the dropdown rather than widening it.

http://plnkr.co/edit/cKRWeypQ1UmpIk3s2c1E?p=preview

Any situation where the dropdown options contain more information than what gets displayed in the text input. In my case, I'm using the dropdown as a list of search results. Each search result contains a name, an ID, a type, and several other fields to help the user select the appropriate result. But in the text box I just show the ID.

What I'd like is to make the dropdown wide enough to show all of the information in the search results, while the text input is narrower since it only shows the ID.

Latest version, Chrome, Firefox, IE, Edge

davidreher:
implemented simalar to select's panel class feature. fixes #4196.

### DIFF
--- a/src/lib/autocomplete/autocomplete.ts
+++ b/src/lib/autocomplete/autocomplete.ts
@@ -72,6 +72,9 @@ export class MdAutocomplete implements AfterContentInit {
   /** Function that maps an option's control value to its display value in the trigger. */
   @Input() displayWith: ((value: any) => string) | null = null;
 
+  /** Classes to be passed to the select panel. Supports the same syntax as `ngClass`. */
+  @Input() panelClass: string|string[]|Set<string>|{[key: string]: any};
+
   /** Event that is emitted whenever an option from the list is selected. */
   @Output() optionSelected: EventEmitter<MdAutocompleteSelectedEvent> =
       new EventEmitter<MdAutocompleteSelectedEvent>();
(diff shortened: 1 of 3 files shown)


src/lib/autocomplete/autocomplete.ts
```diff
@@ -72,6 +72,9 @@ export class MdAutocomplete implements AfterContentInit {
   /** Function that maps an option's control value to its display value in the trigger. */
   @Input() displayWith: ((value: any) => string) | null = null;
 
+  /** Classes to be passed to the select panel. Supports the same syntax as `ngClass`. */
```
Splaktar: autocomplete panel instead of select panel.

jelbourn:

Issue #6451, opened 2017-08-14
Chips plunkr demo is missing styleUrls


mackelito:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Should include the css file

#### What is the current behavior?
Does not include it :)

#### What are the steps to reproduce?
Go to https://material.angular.io/components/chips/overview
click in the icon for "Edit in plunkr"
 
Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: https://goo.gl/DlHd6U

davidarich:
@mackelito Could you please explain the issue or the desired expectation a bit more? The chip colors are set in the included theme CSS file in index.html of the plunkr.
`<link href="https://unpkg.com/@angular/material/prebuilt-themes/indigo-pink.css" rel="stylesheet">`

The theming guide explains how to use prebuilt & custom themes for your application: https://material.angular.io/guide/theming

Jefiozie:
My understanding was that @mackelito means that the `chips-overview-example` doesn't have a reference to the css in the Component. Take for example the `chips-stacked-example` this contains a reference to the css with the `stylesUrls`.
```
@Component({
  selector: 'chips-stacked-example',
  templateUrl: 'chips-stacked-example.html',
  styleUrls: ['chips-stacked-example.css'],
})
```
The `chips-overview-example` however doesn't have this.

```
@Component({
  selector: 'chips-overview-example',
  templateUrl: 'chips-overview-example.html',
})
```

rafaelss95:
@jelbourn this issue can be closed since it's fixed by #6620.

PR #6620, opened 2017-08-24
docs(ChipsOverview): Added stylesUrls to reference css.

Jefiozie:
This will fix # #6451


Jefiozie:
I don't see why the tests are failing. If i need to make changes let me know.

PR #6628, opened 2017-08-24
	Separates toolbar's scss variables to an individual file so that library users can import variables without getting the rules.

YoungCatChen:


YoungCatChen:
I signed it. Please check. Thanks!

YoungCatChen:
I signed it!

jelbourn:
If we were to do this, it would need to be done consistently for all components with documentation and a commitment to stability. We currently consider these variables to be private implementation details such that we can't commit to keeping them stable as an API.

PR #6633, opened 2017-08-25
build: stricter unused variable checking

DevVersion:
Currently only variables that don't start with an underscore will be checked by TSLint.

This doesn't make sense, because variables get prefixed with an underscore to indicate that they are not part of the public API, but still need to be public in favor of AOT compilation for example. This means that those properties are having the public modifier set (explicitly or automatically) and therefore won't be treated as unused anyway.

**TL;DR**: With this change, prefixed variables will be checked by the `no-unused-variable` TSLint rule and those which are marked as `private` and are unused will be catched by TSLint.

### DIFF
--- a/src/cdk/rxjs/rx-operators.ts
+++ b/src/cdk/rxjs/rx-operators.ts
@@ -76,7 +76,7 @@ export interface StrictRxChain<T> {
   result(): Observable<T>;
 }
 
-
+/* tslint:disable:no-unused-variable */
 export class FinallyBrand { private _; }
 export class CatchBrand { private _; }
 export class DoBrand { private _; }
@@ -89,6 +89,7 @@ export class StartWithBrand { private _; }
 export class DebounceTimeBrand { private _; }
 export class AuditTimeBrand { private _; }
 export class TakeUntilBrand { private _; }
+/* tslint:enable:no-unused-variable */
 
 
 export type finallyOperatorType<T> = typeof _finallyOperator & FinallyBrand;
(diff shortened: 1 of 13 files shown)


src/lib/paginator/paginator.spec.ts
```diff
@@ -281,7 +281,7 @@ class MdPaginatorApp {
 
   @ViewChild(MdPaginator) mdPaginator: MdPaginator;
 
-  constructor(private _elementRef: ElementRef) { }
+  constructor() {}
```
crisbeto: You can remove this constructor.

DevVersion: Done.

src/cdk/rxjs/rx-operators.ts
```diff
@@ -89,6 +89,7 @@ export class StartWithBrand { private _; }
 export class DebounceTimeBrand { private _; }
 export class AuditTimeBrand { private _; }
 export class TakeUntilBrand { private _; }
```
crisbeto: I'd be ok with having these properties be public. They're not a part of any API or generated JS anyway.

DevVersion: Yeah I don't mind either. I just changed it. (doesn't really matter)

jelbourn: The brand properties actually _must_ be `private`, otherwise they are structurally identical to anything and everything.

PR #6641, opened 2017-08-25
fix(slider): thumb disappearing on disabled element with thumb label

### LINKED ISSUE #6631: slider thumb disappearing on click
Bug

For a disabled slider, clicking on the slider thumb should produce no visible change or response from the end user's perspective.

When a slider is in the disabled state, but has the "thumbLabel" directive applied, clicking the slider thumb causes it to disappear (it's height and width get set to zero). When a slider is in the disabled state, but DOES NOT have the "thumbLabel" directive applied, clicking the slider thumb causes no visible change.

Plunker template: https://embed.plnkr.co/1VhilOrV68SKzhjb6i7z/

N/A

@angular/material@2.0.0-beta.8

crisbeto:
Fixes the slider thumb becoming hidden when a disabled slider with a thumb label is focused.

Fixes #6631.

### DIFF
--- a/src/lib/slider/slider.scss
+++ b/src/lib/slider/slider.scss
@@ -214,26 +214,28 @@ $mat-slider-focus-ring-size: 30px !default;
 
 
 // Active slider.
-.cdk-focused {
-  &.mat-slider-thumb-label-showing .mat-slider-thumb {
-    transform: scale(0);
+.mat-slider:not(.mat-slider-disabled) {
+  &.cdk-focused {
+    &.mat-slider-thumb-label-showing .mat-slider-thumb {
+      transform: scale(0);
+    }
+
+    .mat-slider-thumb-label {
+      border-radius: 50% 50% 0;
+    }
+
+    .mat-slider-thumb-label-text {
+      opacity: 1;
+    }
   }
 
-  .mat-slider-thumb-label {
-    border-radius: 50% 50% 0;
-  }
-
-  .mat-slider-thumb-label-text {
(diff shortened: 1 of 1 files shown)


src/lib/slider/slider.scss
```diff
@@ -214,7 +214,7 @@ $mat-slider-focus-ring-size: 30px !default;
 
 
 // Active slider.
-.cdk-focused {
+.mat-slider.cdk-focused:not(.mat-slider-disabled) {
```
mmalerba: this adds a lot off specificity, are you sure it doesn't break any of the other states?

crisbeto: I didn't notice anything from poking around the demo. I can reduce the specificity slightly by changing the selector to `.cdk-focused:not(.mat-slider-disabled)`.

mmalerba: I think its good to leave the `.mat-slider` so it doesn't wind up triggering when inside of another focused element, just curious if you had checked for other cases because I know I've had CSS regressions in the past with slider when I try to do these kind of changes

crisbeto: One place I could see it potentially breaking is the mouse, keyboard etc. focused styles a bit below. I can scope those to `.mat-slider:not(.mat-slider-disabled)` as well.

crisbeto: Reworked as discussed @mmalerba.

mmalerba:
LGTM styles still seem to work

PR #6643, opened 2017-08-25
feat(expansion-panel): allow for the panel header height to be customized

### LINKED ISSUE #5641: Make expansion panel height configurable
Proposal

Height of the md-expansion-panel-header should be configurable.

Currently there is no way of doing it. I can only override it in css with important but then there is no animation.

Open material demo app and see expansion panel header.

Custom header height.

All

-

crisbeto:
Uses the new Angular animations features, that allow for dynamic values to be passed to an animation, to let consumers customize the `md-expansion-panel-header` height.

Fixes #5641.


emreavsar:
👍

RainingNight:
Very good!

Issue #6447, opened 2017-08-14
Problems with [disabled] in md-tab with material2-builds


gise88:
#### Bug:
The disabled md-tab does not become enabled when required (it still remains disabled)

[Plunker](https://plnkr.co/edit/9qmSADHgf44omShZdNhQ?p=preview)

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
It affect the `material2-builds` version: `2.0.0-beta.8-85a6fff` but I don't know when this bug was introduced. In the beta-8 it works as expected

#### Is there anything else we should know?
I have to use the `material2-builds` because in the beta-8 there was the `MdPaginatorIntl` not exported. 

#### What I've noticed:
If you click on the button `Toggle Disable`, then click on a one of the enabled tabs, the disabled one becomes active!

willshowell:
Looks like a change detection problem. Clicking on any other tab when it should be enabled triggers the change.

gise88:
Updated the broken plunker! :muscle:

amcdnl:
@gise88 Thanks for updating that! PR is in the queue for this!

emreavsar:
+1

trishalim:
Is there a workaround for this without having to update to latest version?

PR #6644, opened 2017-08-25
bug(tabs): tabs not disabled on change #6447

amcdnl:
Adds manual change detection for when tabs are disabled. Addresses #6447


src/lib/tabs/tab-group.spec.ts
```diff
@@ -186,6 +186,25 @@ describe('MdTabGroup', () => {
...
+
+      let labels = fixture.debugElement.queryAll(By.css('.mat-tab-disabled'));
+      expect(labels.length).toBe(0);
+
+      tabs[0].disabled = true;
+      fixture.detectChanges();
+
+      expect(tabs[0].disabled).toBe(true);
+      labels = fixture.debugElement.queryAll(By.css('.mat-tab-disabled'));
+      expect(labels.length).toBe(0);
```
gise88: Hi @amcdnl! Shouldn't this be `.toBe(1)` instead of `toBe(0)`? I'm not the really the best about testing, so take this more as a question that a point..

amcdnl: Good catch, I wrote a couple more tests that are better while I was at it.

jelbourn:
Presubmit passed

@amcdnl can you rebase?

amcdnl:
Rebased :)

PR #6649, opened 2017-08-26
fix(snack-bar): animation not starting for subsequent snack bars

### LINKED ISSUE #6222: SnackBar without duration doesn't show
Bug in MdSnackBarModule with the following simple code:
```typescript
this.snackBar.open('hello');
setTimeout(() => {
   this.snackBar.open('world');
}, 2000);
```

A snackBar with "hello" should be opened, 2 seconds later it should be dismissed and a snackBar with "world" should be opened.

A snackBar with "hello" is opened, 2 seconds later it is dismissed and **the snackBar with "world" doesn't appear**.

[Demo](https://plnkr.co/edit/vaM6mbPW2SGyvcalgOMa?p=preview)

Weirdly, it works if I define a duration for the "world" snackBar: [demo](https://plnkr.co/edit/n5Lrfdwy9eA5ft12VUXp?p=preview).

It's not consistent to have to define a duration in some cases, not in others.

crisbeto:
Since the switch to OnPush change detection, opening a snack bar while another one is open, programmatically isn't guaranteed to actually show the second snack bar, because the user might not hit the zone (e.g. if it's through a timeout). These changes trigger change detection manually on open. I also got rid of a couple of redundant methods on the snack bar container.

Fixes #6222.

**Note:** It's tricky to write a unit test that captures this exact scenario, because all of the elements were already in the DOM, they just weren't being animated in. It also seems like the animations module wasn't setting the inline styles when I was running the unit tests.

### DIFF
--- a/src/lib/snack-bar/snack-bar-container.ts
+++ b/src/lib/snack-bar/snack-bar-container.ts
@@ -17,6 +17,7 @@ import {
   ElementRef,
   ChangeDetectionStrategy,
   ViewEncapsulation,
+  ChangeDetectorRef,
 } from '@angular/core';
 import {
   trigger,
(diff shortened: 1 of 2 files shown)


src/lib/snack-bar/snack-bar-container.ts
```diff
@@ -63,24 +64,25 @@ export const HIDE_ANIMATION = '195ms cubic-bezier(0.0,0.0,0.2,1)';
...
   /** The portal host inside of this container into which the snack bar content will be loaded. */
   @ViewChild(PortalHostDirective) _portalHost: PortalHostDirective;
 
   /** Subject for notifying that the snack bar has exited from view. */
-  private onExit: Subject<any> = new Subject();
+  onExit: Subject<any> = new Subject();
 
   /** Subject for notifying that the snack bar has finished entering the view. */
-  private onEnter: Subject<any> = new Subject();
+  onEnter: Subject<any> = new Subject();
```
jelbourn: Make these subjects internal?

PR #6653, opened 2017-08-26
a11y(table): add basic, sort, pager demos for a11y #6498

amcdnl:


src/demo-app/a11y/table/table-a11y.html
```diff
@@ -0,0 +1,70 @@
...
+    <h2>Basic Table</h2>
+    <p>Shows name, color and age data.</p>
+    <md-table [dataSource]="basicDataSource">
+      <ng-container cdkColumnDef="name">
+        <md-header-cell *cdkHeaderCellDef>Name</md-header-cell>
+        <md-cell *cdkCellDef="let row">{{row.name}}</md-cell>
+      </ng-container>
+      <ng-container cdkColumnDef="color">
+        <md-header-cell *cdkHeaderCellDef>Color</md-header-cell>
+        <md-cell *cdkCellDef="let row" [style.color]="row.color">{{row.color}}</md-cell>
```
tinayuangao: Some colors are hard to read like yellow. Can we make this more accessible? We can remove the styled colors

src/demo-app/a11y/table/table-a11y.html
```diff
@@ -0,0 +1,70 @@
+<div>
+  <section>
+    <h2>Basic Table</h2>
+    <p>Shows name, color and age data.</p>
+    <md-table [dataSource]="basicDataSource">
```
jelbourn: I took a quick looks at the guidance for `role="grid"`: https://www.w3.org/TR/wai-aria/roles#grid

Something we missed earlier is that 
> A grid is considered editable unless otherwise specified.

We either need to add `aria-readonly` here, or we need to make the table set it by default. @andrewseguin thoughts?

jelbourn: Would be good to add an `aria-label` to the table element

amcdnl: If the table has contents that are editable that would be misrepresentative of the contents. Its really up to the user implementing this to make that decision IMO.

jelbourn: I was thinking the table would have something like `editable` that defaults to false, but we can sort that out later.

For now, just add `aria-readonly` to the `md-table` in the demo

tinayuangao:
@amcdnl You need to `export interface UserData`

src/demo-app/a11y/table/table-a11y.ts
```diff
@@ -0,0 +1,114 @@
+import {Component, ViewChild} from '@angular/core';
+import {DataSource} from '@angular/cdk/table';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+import {Observable} from 'rxjs/Observable';
+import {MdSort, MdPaginator} from '@angular/material';
+
+interface UserData {
```
tinayuangao: Here UserData should be exported

amcdnl: done. :)

kara:
@amcdnl Can you rebase?

kara:
@amcdnl Can you check out the CI failure?  Sorryyy

amcdnl:
Ok, how about now :)

jelbourn:
Is there a real CI failure? aot task looks to be failing

amcdnl:
Ah, `Md` vs `Mat` was causing it. Fixed.

PR #6658, opened 2017-08-27
fix(select): incorrect panel width when trigger is resized post initialization

### LINKED ISSUE #6655: MdSelect - md option doesn't fit the select container using dialog + flex layout
Bug

Options should fit the select container

Options container is smaller than select container.

https://plnkr.co/edit/yEtylQm2RDDrNGQu9GMI?p=preview

1. open the dialog
2. select an option
result: option container is smaller 

@angular v4.3
@angular/material 2.00 -beta8
@angular/flex-layout

crisbeto:
Since we set and cache the trigger width in the `placeholder` setter, it means that we can potentially have an incorrect width if the trigger was resized after initialization. These changes switch to only updating the trigger width from the placeholder getter if there was a width already, which also has the advantage of not triggering a reflow on initialization, but deferring it until a select is opened.

Fixes #6655.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -189,15 +189,26 @@ describe('MdSelect', () => {
       });
     }));
 
-    it('should set the width of the overlay based on the trigger', async(() => {
+    it('should set the width of the overlay based on the trigger', fakeAsync(() => {
       trigger.style.width = '200px';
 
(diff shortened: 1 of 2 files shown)


crisbeto:
The issue that this was fixing shouldn't be a problem anymore after the switch to `mat-form-field`. Closing.

PR #6659, opened 2017-08-27
fix(tooltip): ensure tooltip stays within viewport

### LINKED ISSUE #5428: Bug: Tooltip truncated in the left if the host element is near the left of the screen
#### Bug:

Tooltip displayed at the bottom truncated in the left if the host element is near the left of the screen

The left margin of the tooptip should be zero or slightly above zero, rather than currently negative.

Characters in the left get truncated.

In https://material.angular.io/components/tooltip/examples with Plunker, make the selection the host get closer to the left through remove the left margin in the css, and the tooltip position set to the bottom, and also make the content of the tooltip slightly longer.

There are some elements on the left need tooltip to be displayed at the bottom of the element.

Material2 Beta 7. All browsers are affected.

crisbeto:
* Adds a fallback position to ensure that tooltips can't go outside the viewport.
* Refactors the `transform-origin` logic to depend on the overlay position, instead of the tooltip position.

Fixes #5428.

### DIFF
--- a/src/lib/tooltip/tooltip.ts
+++ b/src/lib/tooltip/tooltip.ts
@@ -19,6 +19,9 @@ import {
   OverlayConfig,
   RepositionScrollStrategy,
   ScrollStrategy,
+  ConnectionPositionPair,
+  HorizontalConnectionPos,
+  VerticalConnectionPos,
 } from '@angular/cdk/overlay';
 import {Platform} from '@angular/cdk/platform';
 import {ComponentPortal} from '@angular/cdk/portal';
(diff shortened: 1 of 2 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -316,52 +327,60 @@ export class MdTooltip implements OnDestroy {
     this._tooltipInstance = null;
   }
 
-  /** Returns the origin position based on the user's position preference */
-  _getOrigin(): OriginConnectionPosition {
-    if (this.position == 'above' || this.position == 'below') {
-      return {originX: 'center', originY: this.position == 'above' ? 'top' : 'bottom'};
-    }
-
+  /** Returns the origin position and a fallback position based on the user's position preference */
```
andrewseguin: Can you describe that the fallback is the inverted position of the origin?

crisbeto:
Addressed the feedback.

mmalerba:
please rebase

mmalerba:
sorry, needs rebase again

PR #6660, opened 2017-08-27
fix(grid-list): styles not cleared when switching to a different styling mode

### LINKED ISSUE #4047: MdGridList cannot switch between ratio and fixed row height

crisbeto:
Fixes the old styles not being cleared when switching between styling modes in the grid list (e.g. going for proportions to a fixed height).

Fixes #4047.

### DIFF
--- a/src/lib/grid-list/grid-list.spec.ts
+++ b/src/lib/grid-list/grid-list.spec.ts
@@ -64,13 +64,13 @@ describe('MdGridList', () => {
   it('should use a ratio row height if passed in', () => {
     let fixture = TestBed.createComponent(GirdListWithRowHeightRatio);
 
-    fixture.componentInstance.heightRatio = '4:1';
+    fixture.componentInstance.rowHeight = '4:1';
     fixture.detectChanges();
 
     let tile = fixture.debugElement.query(By.directive(MdGridTile));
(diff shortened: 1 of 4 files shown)


jelbourn:
Presubmit passed

@crisbeto can you rebase?

PR #6661, opened 2017-08-27
fix(toolbar): no longer auto-generate toolbar rows 

### LINKED ISSUE #1718: toolbar: unable to specify attributes on first toolbar-row

### LINKED ISSUE #6004: MdToolbarModule generates extra toolbar row component
Bug

Encapsulating all content within `<md-toolbar>` tag should not trigger an extra `<md-toolbar-row>` tag when there's no top level content/elements present
i.e.
```
<md-toolbar color="primary">
  <md-toolbar-row>First Row</md-toolbar-row>
  <md-toolbar-row>Second Row</md-toolbar-row>
</md-toolbar>
```

When using only nested `<md-toolbar-row>` components there is an extra top row produced
e.g.
```
<md-toolbar class="mat-toolbar mat-primary" color="primary" role="toolbar" ng-reflect-color="primary">
  <div class="mat-toolbar-layout">
    <md-toolbar-row class="mat-toolbar-row">


    </md-toolbar-row>
    <md-toolbar-row class="mat-toolbar-row">First Row</md-toolbar-row>
    <md-toolbar-row class="mat-toolbar-row">Second Row</md-toolbar-row>
  </div>
</md-toolbar>
```

Create an `<md-toolbar>` component containing one or more nested `<md-toolbar-row>` tags and no other markup as siblings.

Plunker template: http://plnkr.co/edit/NMqusqAvplj0eJzqKdk5?p=preview

I would like to custom style the first (or top) toolbar and none of my styles are applied (I'm assuming since the SASS is precompiled at build time and the generated component is not matching the CSS targeting). My work around is wrapping all toolbar content in `<md-toolbar-row>` tags and providing my own CSS classes

@angular/core: 4.3.1
@angular/material: 2.0.0-beta.8
typescript: 2.3.4

DevVersion:
Currently the toolbar always generates the first `<md-toolbar-row>`. This means that developers have no opportunity to set directives/attributes/classes on the first toolbar row (e.g with flex-layout).

With this change, the toolbar won't auto-generate any `<md-toolbar-row>` element.

The toolbar will have two different row modes:

_Single row toolbar_

```html
<md-toolbar>
  First Tow
</md-toolbar>
```

_Multiple rows toolbar_

```html
<md-toolbar>
  <md-toolbar-row>First Row</md-toolbar-row>
  <md-toolbar-row>Second Row</md-toolbar-row>
</md-toolbar>
```

This means that mixing those two row modes is no longer possible and allowed

```html
<md-toolbar>
  <span>First Row</span>
  <md-toolbar-row>Second Row</md-toolbar-row>
</md-toolbar>
```

BREAKING CHANGE: `<md-toolbar-row>` elements will be no longer auto-generated for the first row. Meaning that custom styling for the first `<md-toolbar-row>` is no longer working as before. Since no toolbar-row is auto-generated anymore, there are two different modes for the toolbar now. Either place content directly inside of the `<md-toolbar>` or place multiple `<md-toolbar-row>` elements.

Fixes #6004. Fixes #1718.

### DIFF
--- a/src/lib/toolbar/toolbar.md
+++ b/src/lib/toolbar/toolbar.md
@@ -2,24 +2,37 @@
 
 <!-- example(toolbar-overview) -->
 
-### Multiple rows
-Toolbars can have multiple rows using `<mat-toolbar-row>` elements. Any content outside of an 
-`<mat-toolbar-row>` element are automatically placed inside of one at the beginning of the toolbar.
-Each toolbar row is a `display: flex` container.
+### Single row
+
+In the most situations, a toolbar will be placed at the top of your application and will only 
+have a single row that includes the title of your application.
 
 ```html
 <mat-toolbar>
-  <span>First Row</span>
+  <span>My Application</span>
+</mat-toolbar>
+```
+
+### Multiple rows
+
+The Material Design specifications describe that toolbars can also have multiple rows. Creating
+toolbars with multiple rows in Angular Material can be done by placing `<mat-toolbar-row>` elements
+inside of a `<mat-toolbar>`.
+
+```html
+<mat-toolbar>  
+  <mat-toolbar-row>
+    <span>First Row</span>
+  </mat-toolbar-row>
   
   <mat-toolbar-row>
     <span>Second Row</span>
   </mat-toolbar-row>
-  
-  <mat-toolbar-row>
-    <span>Third Row</span>
-  </mat-toolbar-row>
 </mat-toolbar>
 ```
+
+**Note**: Placing content outside of a `<mat-toolbar-row>` when multiple rows are specified is not
+supported.
 
 ### Positioning toolbar content
 The toolbar does not perform any positioning of its content. This gives the user full power to 
--- a/src/lib/toolbar/toolbar.ts
+++ b/src/lib/toolbar/toolbar.ts
@@ -7,22 +7,19 @@
  */
 
 import {
+  AfterViewInit,
   ChangeDetectionStrategy,
   Component,
+  ContentChildren,
   Directive,
   ElementRef,
+  isDevMode,
+  QueryList,
   Renderer2,
-  ViewEncapsulation,
+  ViewEncapsulation
 } from '@angular/core';
 import {CanColor, mixinColor} from '@angular/material/core';
-
-
-@Directive({
-  selector: 'mat-toolbar-row',
-  exportAs: 'matToolbarRow',
-  host: {'class': 'mat-toolbar-row'},
-})
-export class MatToolbarRow {}
+import {Platform} from '@angular/cdk/platform';
 
 // Boilerplate for applying mixins to MatToolbar.
 /** @docs-private */
@@ -31,6 +28,12 @@ export class MatToolbarBase {
 }
 export const _MatToolbarMixinBase = mixinColor(MatToolbarBase);
 
+@Directive({
+  selector: 'mat-toolbar-row',
+  exportAs: 'matToolbarRow',
+  host: {'class': 'mat-toolbar-row'},
+})
+export class MatToolbarRow {}
 
 @Component({
   moduleId: module.id,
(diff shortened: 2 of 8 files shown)


jelbourn:
Did were ever discuss just not having any `md-toolbar-row` at all when they're not specified? What was the downside there?

src/lib/toolbar/toolbar.ts
```diff
@@ -44,10 +39,35 @@ export const _MdToolbarMixinBase = mixinColor(MdToolbarBase);
...
+  /** Element reference to the default first row of the toolbar. */
+  @ViewChild('defaultFirstRow', { read: ElementRef }) _defaultFirstRow: ElementRef;
+
+  constructor(renderer: Renderer2, elementRef: ElementRef, private _platform: Platform) {
     super(renderer, elementRef);
   }
 
+  ngAfterViewInit() {
+    // Do nothing if we're not on the browser platform.
+    if (!this._platform.isBrowser) {
```
jelbourn: If you abort here on the server, the prerendered toolbar probably won't look right

src/lib/toolbar/toolbar.ts
```diff
@@ -44,10 +39,35 @@ export const _MdToolbarMixinBase = mixinColor(MdToolbarBase);
...
+      return;
+    }
+
+    const firstRowElement = this._defaultFirstRow.nativeElement;
+
+    // If the first toolbar row, which will be auto-generated, does not have any projected content,
+    // then the auto-generated first toolbar row should be removed. This gives developers
+    // full control over the first toolbar row (e.g. for use with flex-layout).
+    if (!firstRowElement.innerHTML.trim()) {
```
jelbourn: Instead of `innerHTML`, check `!firstRow.textContent.trim() && !firstRow.childNodes.length` ?
(I don't even like looking at `innerHTML` if it can be avoided)

DevVersion: Yeah pretty sure that works too. Let's see if we actually still need that check with the discussed changes on Slack.

willshowell:
What about only projecting `md-toolbar-row`?

```
<div class="mat-toolbar-layout">
  <ng-content select="md-toolbar-row, mat-toolbar-row"></ng-content>
</div>
```

I guess it adds a little fluff to the api, but the behavior would be straightforward and expected.

DevVersion:
@willshowell I chatted with @jelbourn today about an alternative APU design. Should have the PR ready tomorrow.

DevVersion:
@jelbourn Applied the changes we discussed per Chat. Please have another look.

src/lib/toolbar/toolbar.md
```diff
@@ -2,25 +2,38 @@
...
   
   <md-toolbar-row>
-    <span>Third Row</span>
+    <span>Second Row</span>
   </md-toolbar-row>
 </md-toolbar>
 ```
 
+**Note**: Placing content outside of a `<md-toolbar-row>` when multiple rows are specified, is not
+supported.
```
jelbourn: Superfluous comma

src/lib/toolbar/toolbar.ts
```diff
@@ -39,15 +42,43 @@ export const _MdToolbarMixinBase = mixinColor(MdToolbarBase);
   inputs: ['color'],
   host: {
     'class': 'mat-toolbar',
-    'role': 'toolbar'
+    'role': 'toolbar',
```
jelbourn: This `role` should be removed; I never noticed it being added. It's not appropriate for most uses of the md-toolbar.
(see the docs)

src/lib/toolbar/toolbar.ts
```diff
@@ -39,15 +42,43 @@ export const _MdToolbarMixinBase = mixinColor(MdToolbarBase);
...
+    '[class.mat-toolbar-single-row]': '!this._toolbarRows.length'
   },
   changeDetection: ChangeDetectionStrategy.OnPush,
   encapsulation: ViewEncapsulation.None
 })
-export class MdToolbar extends _MdToolbarMixinBase implements CanColor {
+export class MdToolbar extends _MdToolbarMixinBase implements CanColor, AfterViewInit {
+
+  /** Reference to all toolbar row elements that have been projected. */
+  @ContentChildren(MdToolbarRow) _toolbarRows: QueryList<MdToolbarRow>;
```
jelbourn: We need to subscribe to the changes on `_toolbarRows` and update accordingly, since `ngIf` and `ngFor` etc. will affect the content children

src/lib/toolbar/toolbar.ts
DevVersion: Good point

src/lib/toolbar/toolbar.ts
DevVersion: Doesn't Angular automatically update the `_toolbarRows.length` property? I tried it locally and it seemed to work.

jelbourn: It does, but `ngAfterViewInit` is only run once

DevVersion: That makes sense then. Although it doesn't really handle every case. e.g.

```html
<md-toolbar>
  <ng-container *ngIf="test">First Row Overwrite</ng-container>
  <md-toolbar-row>First Row</md-toolbar-row>
</md-toolbar>
```

I'm not sure if we should care about that case as well? It may involve using the `ngDoCheck` lifecycle hook (which may not be that good for performance)

jelbourn: Yeah, there's not much we can do about that. I think it's fine.

jelbourn:
Caretaker note: this is a breaking change. For anyone that has custom styling on `md-toolbar-row` but isn't explicitly putting `md-toolbar-row` in their template, we'll need to add it to their template.

DevVersion:
@jelbourn Done. Added merge-ready, but would be still happy if you briefly look at the breaking change message (I assume for the changelog it will be updated anyway again)

mmalerba:
please rebase

DevVersion:
@mmalerba Done. e2e screenshot uploading seems to timeout (unrelated)

Issue #4639, opened 2017-05-18
Tabs endless loading loop issue


seanbotha123:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Change tabs very quickly, each tab has large amounts of data to process. When switching between tabs too quickly the tabs go into an endless loop of loading.

#### What is the current behavior?
Goes into endless loading loop.

#### What are the steps to reproduce?

[http://plnkr.co/edit/O85thWJeQ0stFAOtojBR](http://plnkr.co/edit/O85thWJeQ0stFAOtojBR)

- open Plunker.
- run.
- click 'Click Me' button/ switch between tabs rapidly.
- loop.

I need to have a ngIf on my components to remove them from the DOM if the tab is not selected. Found the issue where a user was switching tabs quickly and the previous tabs were still loading.

Edit: I see this even happens when not using ngIf. http://plnkr.co/edit/bu2iWvGifGHs95JswBzL

jelbourn:
@andrewseguin looked at the plunker, definitely something weird going on

seanbotha123:
Hi guys,

Any feedback here yet?

james-schwartzkopf:
I believe we are seeing the same issue.  In our case were are trying to store the active tab in ngrx/store.  Things work as expected as long as you switch tabs slowly, but if you change quickly it seems to get stuck in a loop.

It looks like the issue is multiple events piling up in the queue for the async selectChange event emitter?

amcdnl:
Here is a updated plunker - http://plnkr.co/edit/UTgPtgPIKuHScfQKpVFp?p=preview

finalxcode:
mark

PR #6663, opened 2017-08-27
bug(tabs): fix inifinite tab loop #4639

amcdnl:
@jelbourn @andrewseguin - The issue is the `selectChange` event is causing the `ngAfterContentChecked` to get recursively called because this change event is firing so quickly.

The root cause seems to be how mapping of the `selectedIndexChange` event was hooked up:

```typescript
@Output() get selectedIndexChange(): Observable<number> {
    return map.call(this.selectChange, event => event.index);
  }
```

switching this to its own `EventEmitter` resolved the problem. However, that created an issue with a expression changed error, I was able to patch that by adding a timeout around that event emit and solve the issue.

### DIFF
--- a/src/lib/tabs/tab-group.spec.ts
+++ b/src/lib/tabs/tab-group.spec.ts
@@ -70,7 +70,7 @@ describe('MdTabGroup', () => {
       });
     }));
 
-    it('should support two-way binding for selectedIndex', async(() => {
+    it('should set to correct tab on fast change', async(() => {
       let component = fixture.componentInstance;
       component.selectedIndex = 0;
       fixture.detectChanges();
--- a/src/lib/tabs/tab-group.spec.ts
+++ b/src/lib/tabs/tab-group.spec.ts
@@ -70,6 +70,25 @@ describe('MdTabGroup', () => {
       });
     }));
 
+    fit('should support two-way binding for selectedIndex', async(() => {
+      let component = fixture.componentInstance;
+      component.selectedIndex = 0;
+      fixture.detectChanges();
+
+      setTimeout(() => {
+        component.selectedIndex = 1;
+        fixture.detectChanges();
+
+        setTimeout(() => {
+          component.selectedIndex = 0;
+          fixture.detectChanges();
+          fixture.whenStable().then(() => {
+            expect(component.selectedIndex).toBe(0);
+          });
+        }, 1);
+      }, 1);
+    }));
+
     it('should change tabs based on selectedIndex', fakeAsync(() => {
       let component = fixture.componentInstance;
       let tabComponent = fixture.debugElement.query(By.css('md-tab-group')).componentInstance;
--- a/src/lib/tabs/tab-group.spec.ts
+++ b/src/lib/tabs/tab-group.spec.ts
@@ -70,7 +70,7 @@ describe('MdTabGroup', () => {
       });
     }));
 
-    fit('should support two-way binding for selectedIndex', async(() => {
+    it('should support two-way binding for selectedIndex', async(() => {
       let component = fixture.componentInstance;
       component.selectedIndex = 0;
       fixture.detectChanges();
--- a/src/lib/tabs/tab-group.ts
+++ b/src/lib/tabs/tab-group.ts
@@ -23,6 +23,7 @@ import {
   AfterContentChecked,
   OnDestroy,
   ViewEncapsulation,
+  NgZone,
 } from '@angular/core';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {Subscription} from 'rxjs/Subscription';
@@ -141,6 +142,7 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
 
   constructor(_renderer: Renderer2,
               elementRef: ElementRef,
+              private _zone: NgZone,
               private _changeDetectorRef: ChangeDetectorRef) {
     super(_renderer, elementRef);
     this._groupId = nextId++;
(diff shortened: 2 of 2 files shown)


src/lib/tabs/tab-group.ts
```diff
@@ -157,19 +155,22 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
...
+    // (since Math.max(NaN, 0) === NaN).
     let indexToSelect = this._indexToSelect =
         Math.min(this._tabs.length - 1, Math.max(this._indexToSelect || 0, 0));
 
     // If there is a change in selected index, emit a change event. Should not trigger if
     // the selected index has not yet been initialized.
     if (this._selectedIndex != indexToSelect && this._selectedIndex != null) {
       this.selectChange.emit(this._createChangeEvent(indexToSelect));
+      // prevent expression changed error
+      this._zone.run(() => this.selectedIndexChange.emit(indexToSelect));
```
andrewseguin: This makes me wonder if we should bother clamping the tab index, which is the reason why wait for checked content to emit this value. Maybe we should be allowing the dev to input a wrong index like how we do with paginator.

src/lib/tabs/tab-group.ts
```diff
@@ -157,19 +155,22 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
...
+    // and which would otherwise throw the component into an infinite loop
+    // (since Math.max(NaN, 0) === NaN).
     let indexToSelect = this._indexToSelect =
         Math.min(this._tabs.length - 1, Math.max(this._indexToSelect || 0, 0));
 
     // If there is a change in selected index, emit a change event. Should not trigger if
     // the selected index has not yet been initialized.
     if (this._selectedIndex != indexToSelect && this._selectedIndex != null) {
       this.selectChange.emit(this._createChangeEvent(indexToSelect));
+      // prevent expression changed error
```
andrewseguin: Change comment: "Emitting this value after change detection has run since the checked content may contain this variable"

src/lib/tabs/tab-group.ts
```diff
@@ -157,19 +155,22 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
    * a new selected tab should transition in (from the left or right).
    */
   ngAfterContentChecked(): void {
-    // Clamp the next selected index to the bounds of 0 and the tabs length. Note the `|| 0`, which
-    // ensures that values like NaN can't get through and which would otherwise throw the
-    // component into an infinite loop (since Math.max(NaN, 0) === NaN).
+    // Clamp the next selected index to the boundsof 0 and the tabs length.
```
andrewseguin: Did the comment change?

src/lib/tabs/tab-group.ts
amcdnl: Personally I prefer safety checking but if we do not do that on purpose in other places, we should follow the pattern IMO.

src/lib/tabs/tab-group.ts
amcdnl: No, I *had* changed it and had to reformat it to fix length but then backed that out and didn't undo the reformat. Sorry.

src/lib/tabs/tab-group.spec.ts
```diff
@@ -70,6 +70,25 @@ describe('MdTabGroup', () => {
       });
     }));
 
+    it('should set to correct tab on fast change', async(() => {
+      let component = fixture.componentInstance;
+      component.selectedIndex = 0;
+      fixture.detectChanges();
+
+      setTimeout(() => {
```
jelbourn: Why is this `setTimeout` instead of using `fixture.whenStable()`?

src/lib/tabs/tab-group.ts
```diff
@@ -131,9 +130,7 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
   private _backgroundColor: ThemePalette;
 
   /** Output to enable support for two-way binding on `[(selectedIndex)]` */
-  @Output() get selectedIndexChange(): Observable<number> {
-    return map.call(this.selectChange, event => event.index);
-  }
+  @Output() selectedIndexChange = new EventEmitter();
```
jelbourn: Did you mean to omit the emitted type?

src/lib/tabs/tab-group.ts
```diff
@@ -157,16 +155,20 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
...
     let indexToSelect = this._indexToSelect =
         Math.min(this._tabs.length - 1, Math.max(this._indexToSelect || 0, 0));
 
     // If there is a change in selected index, emit a change event. Should not trigger if
     // the selected index has not yet been initialized.
     if (this._selectedIndex != indexToSelect && this._selectedIndex != null) {
       this.selectChange.emit(this._createChangeEvent(indexToSelect));
+      // Emitting this value after change detection has run
+      // since the checked content may contain this variable
+      this._zone.run(() => this.selectedIndexChange.emit(indexToSelect));
```
jelbourn: It's odd to use `zone.run` like this; does it work if you use `Promise.resolve().then(() => ...)` ?

src/lib/tabs/tab-group.spec.ts
amcdnl: I used `setTimeout` because thats what his demo used.

src/lib/tabs/tab-group.ts
amcdnl: That does work, nifty trick.

amcdnl:
@jelbourn  Ready for another review.

PR #6673, opened 2017-08-28
build: invalid paths when building examples module

DevVersion:
On Windows, if someone runs the `gulp build-examples-module` task, the `example-module.ts` file will be generated incorrectly.

Path delimiters are missing in the TypeScript imports because the Windows path delimiters are being removed accidentally.


amcdnl:
LGTM

jelbourn:
@DevVersion needs rebase

DevVersion:
@jelbourn Done.

PR #6680, opened 2017-08-28

crisbeto:

### DIFF
--- a/src/cdk/scrolling/viewport-ruler.ts
+++ b/src/cdk/scrolling/viewport-ruler.ts
@@ -6,23 +6,49 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Injectable, Optional, SkipSelf} from '@angular/core';
+import {Injectable, Optional, SkipSelf, NgZone, OnDestroy} from '@angular/core';
+import {Platform} from '@angular/cdk/platform';
 import {ScrollDispatcher} from './scroll-dispatcher';
+import {Observable} from 'rxjs/Observable';
+import {fromEvent} from 'rxjs/observable/fromEvent';
+import {merge} from 'rxjs/observable/merge';
+import {auditTime} from 'rxjs/operator/auditTime';
+import {Subscription} from 'rxjs/Subscription';
+import {of as observableOf} from 'rxjs/observable/of';
 
+/** Time in ms to throttle the resize events by default. */
+export const DEFAULT_RESIZE_TIME = 20;
 
 /**
  * Simple utility for getting the bounds of the browser viewport.
  * @docs-private
  */
 @Injectable()
-export class ViewportRuler {
+export class ViewportRuler implements OnDestroy {
 
   /** Cached document client rectangle. */
   private _documentRect?: ClientRect;
 
-  constructor(scrollDispatcher: ScrollDispatcher) {
+  /** Stream of viewport change events. */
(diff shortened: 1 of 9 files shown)


src/cdk/scrolling/viewport-ruler.ts
```diff
@@ -20,9 +28,22 @@ export class ViewportRuler {
...
+  /** Stream of viewport change events. */
+  private _changed = new Subject<string>();
+
+  constructor(platform: Platform, ngZone: NgZone, scrollDispatcher: ScrollDispatcher) {
+    if (platform.isBrowser) {
+      ngZone.runOutsideAngular(() => {
+        merge<Event>(
+          fromEvent(window, 'resize'),
+          fromEvent(window, 'orientationchange')
+        ).subscribe(event => this._changed.next(event.type));
```
jelbourn: Why subscribe and emit instead of exposing this `Observable`?

src/cdk/scrolling/viewport-ruler.ts
```diff
@@ -20,9 +28,22 @@ export class ViewportRuler {
   /** Cached document client rectangle. */
   private _documentRect?: ClientRect;
 
-  constructor(scrollDispatcher: ScrollDispatcher) {
+  /** Stream of viewport change events. */
+  private _changed = new Subject<string>();
```
jelbourn: Is the difference between `resize` and `orientationchange` useful here? If we make it `void` we can change it to a more robust object later without it being a breaking change

src/cdk/scrolling/viewport-ruler.ts
```diff
@@ -20,9 +28,22 @@ export class ViewportRuler {
...
+        merge<Event>(
+          fromEvent(window, 'resize'),
+          fromEvent(window, 'orientationchange')
+        ).subscribe(event => this._changed.next(event.type));
+      });
+    }
+
     // Subscribe to scroll and resize events and update the document rectangle on changes.
     scrollDispatcher.scrolled(0, () => this._cacheViewportGeometry());
+    this.change().subscribe(() => this._cacheViewportGeometry());
```
jelbourn: Need to unsubscribe in `ngOnDestroy`?

src/cdk/scrolling/viewport-ruler.ts
crisbeto: The intention was to let people react differently depending on the event, but we can always change it if somebody asks for it.

crisbeto:

jelbourn:
@crisbeto 
```
src/cdk/scrolling/viewport-ruler.ts(40,7): error TS2322: Type 'Observable<Event>' is not assignable to type 'Subject<void>'.
  Property 'observers' is missing in type 'Observable<Event>'.
```

crisbeto:
Fixed @jelbourn.

mmalerba:
@crisbeto looks like prerender CI task is failing

crisbeto:
Sorted out the prerender failure.

rolandjitsu:
Is this public? And it's listed in beta 11, but I cannot seem to find the `.change()` method.

willshowell:
@rolandjitsu it was reverted and reopened https://github.com/angular/material2/pull/7113. I think that is an oversight in the changelog

rolandjitsu:
@willshowell thanks, I'll keep an eye on that.

Issue #3347, opened 2017-03-01
Tabs - Desktop padding doesn't meet design specifications


frankspin89:
#### Bug, feature request, or proposal:

Type: Bug

#### What is the expected behavior?

Tabs should have a 24px left and right padding on desktop. 

![components_tabs_usage_specs13](https://cloud.githubusercontent.com/assets/4480104/23448450/58c011de-fe51-11e6-89b9-554f47d6d4f8.png)

#### What is the current behavior?

Current desktop padding for tabs are 12px left and right. 

#### What is the use-case or motivation for changing an existing behavior?

Follow the material.io guidelines.

#### Which versions of Angular, Material, OS, browsers are affected?

- OS: macOs Sierra
- Chrome: Version 56.0.2924.87 (64-bit)
- Angular: 4.0.0-rc.1
- Material: 2.0.0-beta.2

#### Is there anything else we should know?

Introduce a breakpoint for desktop screens. Use $gt-md breakpoint (min-width 1280px);

PR #6681, opened 2017-08-28
bug(tabs): tab spacing on desktop incorrect #3347

amcdnl:
Sets left/right padding on desktop to spec ( 24px ). Addresses #3347

### DIFF
--- a/src/lib/tabs/tab-group.scss
+++ b/src/lib/tabs/tab-group.scss
@@ -20,6 +20,7 @@
 @media ($mat-xsmall) {
   .mat-tab-label {
     min-width: 72px;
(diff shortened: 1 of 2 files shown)


src/lib/tabs/tab-group.scss
```diff
@@ -20,6 +20,7 @@
 @media ($mat-xsmall) {
   .mat-tab-label {
     min-width: 72px;
+    padding: 0 12px;
```
jelbourn: Is it only `xsmall` that should have the denser look?

amcdnl: Bumped to small per our convo.

Issue #6330, opened 2017-08-07
[overlay] keyboard event dispatcher


willshowell:
#### Bug, feature request, or proposal:
Feature

#### Motivation
If focus is not one of an overlay component's focusable elements, keyboard events will land on the body and context is loss for which overlay component should have been targeted. 

More specifically, something like this will later enable dialog users to respond to targeted escape events with only a `DialofRef` instance. See https://github.com/angular/material2/pull/3460#issuecomment-320036521 and preceding 4-5 comments.

> It feels like there needs to be a more involved keyboard event dispatcher for overlays. Probably one listener on the body that dispatches the event to the appropriate overlay based on some combination of event origin and open order (maybe with a separate priority mechanism).

#### Reproduction of keydown target issue

https://plnkr.co/edit/puYeeRCta8A7moZOPRWI?p=preview

PR #6682, opened 2017-08-28
feat(overlay): add keyboard dispatcher for targeting correct overlay

willshowell:
Fixes https://github.com/angular/material2/issues/6330
Fixes https://github.com/angular/material2/issues/3251
Closes https://github.com/angular/material2/pull/3460

### DIFF
--- a/src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
+++ b/src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright Google Inc. All Rights Reserved.
+ * Copyright Google LLC All Rights Reserved.
  *
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
--- a/src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
+++ b/src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
@@ -1,0 +1,98 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Inject, Injectable, Optional, SkipSelf, OnDestroy} from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
+import {OverlayRef} from '../overlay-ref';
+import {Subscription} from 'rxjs/Subscription';
+import {RxChain, filter} from '@angular/cdk/rxjs';
+import {fromEvent} from 'rxjs/observable/fromEvent';
+
+/**
+ * Service for dispatching keyboard events that land on the body to appropriate overlay ref,
+ * if any. It maintains a list of attached overlays to determine best suited overlay based
+ * on event target and order of overlay opens.
+ */
+@Injectable()
+export class OverlayKeyboardDispatcher implements OnDestroy {
+
+  /** Currently attached overlays in the order they were attached. */
+  _attachedOverlays: OverlayRef[] = [];
+
+  private _keydownEventSubscription: Subscription | null;
+
+  constructor(@Optional() @Inject(DOCUMENT) private _document: any) { }
+
+  ngOnDestroy() {
+    if (this._keydownEventSubscription) {
+      this._keydownEventSubscription.unsubscribe();
+      this._keydownEventSubscription = null;
+    }
+  }
+
+  /** Add a new overlay to the list of attached overlay refs. */
+  add(overlayRef: OverlayRef): void {
+    // Lazily start dispatcher once first overlay is added
+    if (!this._keydownEventSubscription) {
+      this._subscribeToKeydownEvents();
+    }
+
+    this._attachedOverlays.push(overlayRef);
+  }
+
+  /** Remove an overlay from the list of attached overlay refs. */
+  remove(overlayRef: OverlayRef): void {
+    const index = this._attachedOverlays.indexOf(overlayRef);
+    if (index > -1) {
+      this._attachedOverlays.splice(index, 1);
+    }
+  }
+
+  /**
+   * Subscribe to keydown events that land on the body and dispatch those
+   * events to the appropriate overlay.
+   */
+  private _subscribeToKeydownEvents(): void {
+    const bodyKeydownEvents = fromEvent<KeyboardEvent>(this._document.body, 'keydown');
+
+    this._keydownEventSubscription = RxChain.from(bodyKeydownEvents)
+      .call(filter, () => !!this._attachedOverlays.length)
+      .subscribe(event => {
+        // Dispatch keydown event to correct overlay reference
+        this._selectOverlayFromEvent(event)._keydownEvents.next(event);
+      });
+  }
+
+  /** Select the appropriate overlay from a keydown event. */
+  private _selectOverlayFromEvent(event: KeyboardEvent): OverlayRef {
+    // Check if any overlays contain the event
+    const targetedOverlay = this._attachedOverlays.find(overlay => {
+      return overlay.overlayElement === event.target ||
+          overlay.overlayElement.contains(event.target as HTMLElement);
+    });
+
+    // Use that overlay if it exists, otherwise choose the most recently attached one
+    return targetedOverlay || this._attachedOverlays[this._attachedOverlays.length - 1];
+  }
+
+}
+
+/** @docs-private */
+export function OVERLAY_KEYBOARD_DISPATCHER_PROVIDER_FACTORY(
+      dispatcher: OverlayKeyboardDispatcher, _document) {
+  return dispatcher || new OverlayKeyboardDispatcher(_document);
+}
+
+/** @docs-private */
+export const OVERLAY_KEYBOARD_DISPATCHER_PROVIDER = {
+  // If there is already an OverlayKeyboardDispatcher available, use that.
+  // Otherwise, provide a new one.
+  provide: OverlayKeyboardDispatcher,
+  deps: [[new Optional(), new SkipSelf(), OverlayKeyboardDispatcher], [new Optional(), DOCUMENT]],
+  useFactory: OVERLAY_KEYBOARD_DISPATCHER_PROVIDER_FACTORY
+};
--- a/src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
+++ b/src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
@@ -6,8 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Inject, Injectable, Optional, SkipSelf, OnDestroy} from '@angular/core';
-import {DOCUMENT} from '@angular/platform-browser';
+import {Injectable, Optional, SkipSelf, OnDestroy} from '@angular/core';
 import {OverlayRef} from '../overlay-ref';
 import {Subscription} from 'rxjs/Subscription';
 import {RxChain, filter} from '@angular/cdk/rxjs';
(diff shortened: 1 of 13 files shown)


src/cdk/overlay/overlay-ref.ts
```diff
@@ -140,6 +145,16 @@ export class OverlayRef implements PortalHost {
     return this._detachments.asObservable();
   }
 
+  /** Returns an observable that emits when the overlay has been disposed. */
```
jelbourn: nit: prefer "Gets" to "Returns"

src/cdk/overlay/overlay.ts
```diff
@@ -87,4 +93,18 @@ export class Overlay {
   private _createPortalHost(pane: HTMLElement): DomPortalHost {
     return new DomPortalHost(pane, this._componentFactoryResolver, this._appRef, this._injector);
   }
+  /**
+   * Subscribe to attach and detach events and register them accordingly
+   * with the keyboard dispatcher service.
+   */
+  private _trackOverlayAttachments(overlayRef: OverlayRef): void {
```
jelbourn: Could each `OverlayRef` to this itself in its constructor?

willshowell:
@jelbourn comments addressed. Also added a demo to the overlay.

src/lib/dialog/dialog.ts
```diff
@@ -241,6 +236,15 @@ export class MdDialog {
       });
     }
 
+    // Close when escape keydown event occurs
+    RxChain.from(overlayRef.keyboardEvents())
+      .call(filter, event => event.type === 'keydown' && event.keyCode === ESCAPE)
+      .subscribe(() => {
+        if (!dialogRef.disableClose) {
```
crisbeto: Might be a bit more concise to have this in `filter`, instead of checking it in the `subscribe`.

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Subscribe to all key events that land on the body and dispatch those
+   * events to the appropriate overlay.
+   */
+  private _dispatchKeyboardEvents(): void {
+    const bodyKeyboardEvents = merge(
+        fromEvent(document.body, 'keydown'),
+        fromEvent(document.body, 'keypress'),
+        fromEvent(document.body, 'keyup')) as Observable<KeyboardEvent>;
```
crisbeto: Instead of casting it to an `Observable<KeyboardEvent>`, you can pass the `KeyboardEvent` as a generic: `merge<KeyboardEvent>`.

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * Subscribe to all key events that land on the body and dispatch those
+   * events to the appropriate overlay.
+   */
+  private _dispatchKeyboardEvents(): void {
+    const bodyKeyboardEvents = merge(
+        fromEvent(document.body, 'keydown'),
+        fromEvent(document.body, 'keypress'),
```
crisbeto: I'm not sure that this should handle `keydown`, `keypress` and `keyup` @jelbourn. Since for 99% of cases you'd only need `keydown`, it means that every time you consume the service, you'd have to have a `filter(event => event.type === 'keydown')`.

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+        // Dispatch keyboard event to correct overlay reference
+        this._selectOverlayFromEvent(this._attachedOverlays, event)
+            ._dispatchedKeyboardEvents.next(event);
+      });
+  }
+
+  /** Select the appropriate overlay from a list of overlays and a keyboard event. */
+  private _selectOverlayFromEvent(overlays: OverlayRef[], event: KeyboardEvent): OverlayRef {
+    // Check if any overlays contain the event
+    const targetedOverlay = this._attachedOverlays.find(overlay =>
```
crisbeto: This doesn't handle if the keyboard event came directly from the overlay element. You can tweak the predicate to include `overlay.overlayElement === event.target`.

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+      .call(takeUntil, this._onDestroy)
+      .subscribe(event => {
+        // Dispatch keyboard event to correct overlay reference
+        this._selectOverlayFromEvent(this._attachedOverlays, event)
+            ._dispatchedKeyboardEvents.next(event);
+      });
+  }
+
+  /** Select the appropriate overlay from a list of overlays and a keyboard event. */
+  private _selectOverlayFromEvent(overlays: OverlayRef[], event: KeyboardEvent): OverlayRef {
```
crisbeto: Since the `overlays` is always going to be `this._attachedOverlays`, you can take it from the class property instead of passing it in.

src/cdk/overlay/overlay-ref.ts
```diff
@@ -22,12 +23,16 @@ export class OverlayRef implements PortalHost {
   private _backdropClick: Subject<any> = new Subject();
   private _attachments = new Subject<void>();
   private _detachments = new Subject<void>();
+  private _disposed = new Subject<void>();
+
+  _dispatchedKeyboardEvents = new Subject<KeyboardEvent>();
```
crisbeto: Instead of having the `OverlayRef` add and remove itself from the dispatcher, a cleaner approach would be to only have it expose the stream of keyboard events. Then the `OverlayRef.keyboardEvents`  can be simplified to:

```ts
// note: don't get too caught up in the naming
return this.keyboardDispatcher.events.filter(event => event.target === this.overlayElement);
```

This less room for mistakes when managing the overlays and it also removes the circular dependency between the `OverlayRef` and the dispatcher.

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
willshowell: I'd be happy with this. I was aiming for generic, but the api could be "rebranded" 'to just `keydownEvents`.

src/cdk/overlay/overlay-ref.ts
willshowell: Hm I don't entirely follow.

1. I kind of see how I could remove the `Subject`; maybe if I created a new event type that paired overlays to keydowns, and then the overlay just filters out ones directed at itself and maps to the keydown.
2. An important piece of this is dispatching an event to the most recently attached overlay. Often, events land on the body with no other context. That's the whole point of tracking overlays as they are attached and detached.

crisbeto: Alright, I see. I guess that in most cases the top-most overlay would be the one dispatching the keyboard events, but that won't always be the case (like with `md-autocomplete`). Let's keep it.

willshowell:
@crisbeto comments addressed. Please take another look!

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
```diff
@@ -0,0 +1,96 @@
...
+        // Dispatch keydown event to correct overlay reference
+        this._selectOverlayFromEvent(event)._dispatchedKeydownEvents.next(event);
+      });
+  }
+
+  /** Select the appropriate overlay from a keydown event. */
+  private _selectOverlayFromEvent(event: KeyboardEvent): OverlayRef {
+    // Check if any overlays contain the event
+    const targetedOverlay = this._attachedOverlays.find(overlay => {
+      return overlay.overlayElement.contains(event.target as HTMLElement) ||
```
crisbeto: I'd move the `overlay.overlayElement === event.target` check first since it could potentially exit sooner.

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
```diff
@@ -0,0 +1,96 @@
...
+          overlay.overlayElement === event.target;
+    });
+
+    // Use that overlay if it exists, otherwise choose the most recently attached one
+    return targetedOverlay ?
```
crisbeto: `return targetedOverlay || this._attachedOverlays[this._attachedOverlays.length - 1]`.

src/cdk/overlay/overlay-ref.ts
```diff
@@ -23,11 +24,14 @@ export class OverlayRef implements PortalHost {
   private _attachments = new Subject<void>();
   private _detachments = new Subject<void>();
 
+  _dispatchedKeydownEvents = new Subject<KeyboardEvent>();
```
crisbeto: Add a description to this one. Also potentially rename it to something like `_keydownEvents`.

crisbeto:
@willshowell looks like the `prerender` task is still failing. You can debug it locally using `gulp ci:prerender`. Ping me to mark as merge ready.

willshowell:
@crisbeto I had issues with `fromEvent` not working well with the `DOCUMENT` token, so I opted to make the dispatch start lazily. PTAL.

src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (index > -1) {
+      this._attachedOverlays.splice(index, 1);
+    }
+  }
+
+  /**
+   * Subscribe to keydown events that land on the body and dispatch those
+   * events to the appropriate overlay.
+   */
+  private _dispatchKeydownEvents(): void {
```
crisbeto: Judging by the name of the method, I'd assume that it would dispatch the events immediately, but that's not what it does. Something along the lines of `_subscribeToKeydownEvents` might be more appropriate.

willshowell:
Rebased. cc @kara

willshowell:
@kara is there anything I need to do regarding the presubmit failure?

andrewseguin:
Getting some AOT issue:

```
Error: Error encountered resolving symbol values statically. Only initialized variables and constants can be referenced because the value of this variable is needed by the template compiler (position 10:22 in the original .ts file), resolving symbol DOCUMENT in /platform-browser/src/dom/dom_tokens.d.ts, resolving symbol DOCUMENT in /platform-browser/src/platform-browser.d.ts, resolving symbol DOCUMENT in /platform-browser/public_api.d.ts, resolving symbol DOCUMENT in /platform-browser/index.d.ts, resolving symbol DOCUMENT in platform_browser_safe/index.d.ts, resolving symbol OVERLAY_KEYBOARD_DISPATCHER_PROVIDER in /src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts, resolving symbol OVERLAY_PROVIDERS in /src/cdk/overlay/overlay-module.ts, resolving symbol OverlayModule in /src/cdk/overlay/overlay-module.ts, resolving symbol OverlayModule in /src/cdk/overlay/overlay-module.ts

```

Can you try building in AOT?

willshowell:
@andrewseguin I wasn't able to reproduce, but I've rebased and updated the license.

I'm passing on `gulp ci:lint, aot, prerender` (`ci:test` is broken from something else).

mmalerba:
Is injecting `DOCUMENT` instead of using `document` critical here? Removing the `DOCUMENT` stuff fixes the error @andrewseguin  mentioned

jelbourn:
@mmalerba we should only need to use `DOCUMENT` if we're mocking something in a test

mmalerba:
I think `DOCUMENT` is special somehow, I tried creating a different injection token, `OVERLAY_DOCUMENT`, and everything worked fine. So we should switch it and if we need it for testing just create our own provider. @willshowell Once that's done I can try again on presubmitting.

jelbourn:
Oh yeah, now I remember. I had to do the same thing for the tests for `Directionality`. `BidiModule` has
```ts
{provide: DIR_DOCUMENT, useExisting: DOCUMENT},
```

willshowell:
Doesn't seem there's any need to mock `document` for the tests, so I've removed the token.

jelbourn:
@mmalerba should be ready for another spin

PR #6684, opened 2017-08-28
fix(select): do not set min-width

### LINKED ISSUE #2711: After selecting an option on md-select it does not adjust width while resizing window
Bug

Once the user selects a value on md-select and the browser window is resized the md-select element should adjust its width accordingly. md-input works fine.

The md-select width is not adjusted correctly when the browser window is re-sized.
Full Size window:
full-size

Resized window:
resize_issue

Plunker: http://plnkr.co/edit/DiPQWnqp76znq27BmXuj?p=preview
1. If we try resizing the preview window without choosing any options on both "Select Option 1" and "Select Option 2", it should work fine.
2. Choose an option for  "Select Option 1" and "Select Option 2" and then try to resize. The select element does not adjust its width automatically like it did when none of the options were selected.

The field alignment for md-select is thrown off when the user tries to resize the page.

Angular Material (SHA 899f190)
Browser: Chrome

### LINKED ISSUE #3090: Adjust width and height of md-select
**Template:**
```
<md-select placeholder="Select Costcode Template" name="costcode_template" 
              [width]="'100%'"
              cdk-connected-overlay 
              ngModel>
```
**Component.ts**
```
import { MdSelect, MdOption } from '@angular/material';
```
**Module**
```
import { MdSelectModule, OverlayModule } from '@angular/material';
@NgModule({
    declarations: [
        ProjectComponent,
    ],
    imports: [
        FormsModule,
        HttpModule,
        projectsRouting,
        MdSelectModule,
        OverlayModule
    ]
})
```
The select should adjust the width and height.

`caused by: No provider for TemplateRef!`

os: win32 x64
@angular/core: 2.4.7
@angular/cli: 1.0.0-beta.31

shlomiassaf:
The select element should occupy the amount of space (width) allowed by the parent, this is the current state but if the width is lower than 112px the select element fixates on 112px.

> select element can also get a fixed value applied on it directly (`<md-select style="width: 40px">`)

With the deprecation of `/deep/`, `:ng-deep` etc... Overriding is not trivial  , unless `ViewEncapsulation.None` is used.

Since the `min-height` property is set on a child element of `md-select` (`.mat-select-trigger`) it can not be reached.
Setting `.mat-select-trigger` on a component's style will result in `.mat-select-trigger [_ng-content_....]`
 
This PR remove the `min-height` property.

@jelbourn if I'm missing something and this css property is still required we can still use it but move it to the host (`md-select`) so it can get overriden by a parent component.

Closes #2711
Closes #3090

### DIFF
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -5,7 +5,6 @@
 @import '../core/a11y/a11y';
 
 $mat-select-trigger-height: 30px !default;
-$mat-select-trigger-min-width: 112px !default;
 $mat-select-arrow-size: 5px !default;
 $mat-select-arrow-margin: 4px !default;
 $mat-select-panel-max-height: 256px !default;
@@ -20,7 +19,6 @@ $mat-select-trigger-underline-height: 1px !default;
   display: flex;
   align-items: center;
   height: $mat-select-trigger-height;
-  min-width: $mat-select-trigger-min-width;
   cursor: pointer;
   position: relative;
   box-sizing: border-box;


mmalerba:
#6488 will probably make this obsolete

shlomiassaf:
@mmalerba @jelbourn I assume this is not relevant anymore, right?

mmalerba:
Yeah, will be fixed when #6488 is merged

src/lib/select/select.scss
```diff
@@ -5,7 +5,6 @@
 @import '../core/a11y/a11y';
 
 $mat-select-trigger-height: 30px !default;
-$mat-select-trigger-min-width: 112px !default;
 $mat-select-arrow-size: 5px !default;
 $mat-select-arrow-margin: 4px !default;
 $mat-select-panel-max-height: 256px !default;
```
DeivaSubramaniam: dasdsad

PR #6692, opened 2017-08-29
refactor(selection-list): use disable-ripple mixin

DevVersion:
* Switches the `disableRipple` getter/setter with the `disableRipple` mixin from core (payload reduction)
* Prefixes the `isRippleEnabled` method with an underscore to hide it from the API docs.

### DIFF
--- a/src/lib/list/list.ts
+++ b/src/lib/list/list.ts
@@ -156,7 +156,7 @@ export class MdListItem extends _MdListItemMixinBase implements AfterContentInit
   }
 
   /** Whether this list item should show a ripple effect when clicked.  */
-  isRippleEnabled() {
+  _isRippleEnabled() {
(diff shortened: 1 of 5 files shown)


src/lib/list/list.ts
```diff
@@ -156,7 +156,7 @@ export class MdListItem extends _MdListItemMixinBase implements AfterContentInit
   }
 
   /** Whether this list item should show a ripple effect when clicked.  */
-  isRippleEnabled() {
+  _isRippleEnabled() {
```
crisbeto: Perhaps rename this to `_rippleDisabled` so it matches the mixin name?

DevVersion: Yeah that might be easier for the template as well.

DevVersion: Done. Still named it `_isRippleDisabled`.

PR #6708, opened 2017-08-29
fix(datepicker): Create a new injection token to avoid overriding LOCALE_ID

g1shin:
New injection token `MAT_DATE_LOCALE` is added to allow overriding locale code, while avoiding overriding `LOCALE_ID`.

### DIFF
--- a/src/lib/core/datetime/index.ts
+++ b/src/lib/core/datetime/index.ts
@@ -6,9 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {NgModule} from '@angular/core';
+import {LOCALE_ID, NgModule} from '@angular/core';
 import {DateAdapter} from './date-adapter';
-import {NativeDateAdapter} from './native-date-adapter';
+import {NativeDateAdapter, MAT_DATE_LOCALE} from './native-date-adapter';
 import {MD_DATE_FORMATS} from './date-formats';
 import {MD_NATIVE_DATE_FORMATS} from './native-date-formats';
 
@@ -20,7 +20,10 @@ export * from './native-date-formats';
 
 
 @NgModule({
-  providers: [{provide: DateAdapter, useClass: NativeDateAdapter}],
+  providers: [
+    {provide: DateAdapter, useClass: NativeDateAdapter},
+    {provide: MAT_DATE_LOCALE, useExisting: LOCALE_ID}
+  ],
 })
 export class NativeDateModule {}
 
--- a/src/lib/core/datetime/index.ts
+++ b/src/lib/core/datetime/index.ts
@@ -6,12 +6,11 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {LOCALE_ID, NgModule} from '@angular/core';
-import {DateAdapter, MAT_DATE_LOCALE} from './date-adapter';
+import {NgModule} from '@angular/core';
+import {DateAdapter, MAT_DATE_LOCALE_PROVIDER} from './date-adapter';
 import {NativeDateAdapter} from './native-date-adapter';
 import {MD_DATE_FORMATS} from './date-formats';
 import {MD_NATIVE_DATE_FORMATS} from './native-date-formats';
-
 
 export * from './date-adapter';
 export * from './date-formats';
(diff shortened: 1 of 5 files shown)


src/lib/core/datetime/native-date-adapter.spec.ts
```diff
@@ -334,13 +333,13 @@ describe('NativeDateAdapter', () => {
 });
 
 
-describe('NativeDateAdapter with LOCALE_ID override', () => {
+describe('NativeDateAdapter with MAT_DATE_LOCALE override', () => {
```
mmalerba: can we also have tests for `LOCAL_ID` override? this should still cascade to `MAT_DATE_LOCALE`

src/lib/core/datetime/index.ts
```diff
@@ -20,7 +20,10 @@ export * from './native-date-formats';
 
 
 @NgModule({
-  providers: [{provide: DateAdapter, useClass: NativeDateAdapter}],
+  providers: [
+    {provide: DateAdapter, useClass: NativeDateAdapter},
+    {provide: MAT_DATE_LOCALE, useExisting: LOCALE_ID}
```
mmalerba: I think this should go in the `MdDatepickerModule`

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -48,13 +47,15 @@ function range<T>(length: number, valueFunction: (index: number) => T): T[] {
   return valuesArray;
 }
 
+/** InjectionToken for datepicker that can be used to override default locale code. */
+export const MAT_DATE_LOCALE = new InjectionToken<string>('MdDateLocale');
```
mmalerba: I would just make the string value of the token the same as the name `'MAT_DATE_LOCALE'`

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -48,13 +47,15 @@ function range<T>(length: number, valueFunction: (index: number) => T): T[] {
...
 
 /** Adapts the native JS Date for use with cdk-based components that work with dates. */
 @Injectable()
 export class NativeDateAdapter extends DateAdapter<Date> {
-  constructor(@Optional() @Inject(LOCALE_ID) localeId: any) {
+  constructor(@Optional() @Inject(MAT_DATE_LOCALE) matDateLocale: any) {
```
mmalerba: any --> string

g1shin:
Comments addressed. Ready for review again :+1:

src/lib/datepicker/datepicker.md
```diff
@@ -116,13 +116,13 @@ three pieces via injection:
  3. The message strings used in the datepicker's UI.
 
 #### Setting the locale code
-By default the datepicker will use the locale code from the `LOCALE_ID` injection token from
-`@angular/core`. If you want to override it, you can provide a new value for the token:
-
+By default `MAT_DATE_LOCALE` injection token will use the existing `LOCALE_ID` locale code from
```
kara: Nit: By default `MAT_DATE_LOCALE` injection token -> By default, the `MAT_DATE_LOCALE` injection token

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -48,13 +47,15 @@ function range<T>(length: number, valueFunction: (index: number) => T): T[] {
   return valuesArray;
 }
 
+/** InjectionToken for datepicker that can be used to override default locale code. */
+export const MAT_DATE_LOCALE = new InjectionToken<string>('MAT_DATE_LOCALE');
```
mmalerba: can we move this to `DateAdapter` since we expect all adapters to use it not just `NativeDateAdapter`?

g1shin:
Changes have been made :smiley: Ready for review

mmalerba:
LGTM

src/lib/core/datetime/index.ts
mmalerba: @jelbourn can you think of a good way to avoid putting this in `NativeDateModule`, `MomentDateModule`, etc? I was thinking maybe we could put it in MdCommonModule and have this module import that?

jelbourn: I think it doesn't really belong in `MdCommonModule` since it is only used in a few components. You could create a separate `MdCommonDateModule`, but that also seems like overkill. Easier thing would probably be just to make a common provider, e.g.
```ts
export const MAT_DATE_LOCALE_PROVIDER = {provide: MAT_DATE_LOCALE, useExisting: LOCALE_ID};
```
Adding that is about the same as importing another module.

g1shin:
Added `MAT_DATE_LOCALE_PROVIDER`.

Issue #6192, opened 2017-08-01
Md-icon inside md-expansion-panel header not aligned correctly


alexandrupaul7:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
md-icon should stick with the md-panel-tile text

#### What is the current behavior?
md-icon not aligned correctly

#### What are the steps to reproduce?

Plunker: http://plnkr.co/edit/6VVJbXSkYGiWJ1OaEp5u?p=preview

            <md-expansion-panel>
                <md-expansion-panel-header>
                    <md-panel-title>
                      <md-icon>search</md-icon> Search
                    </md-panel-title>                    
                </md-expansion-panel-header>
                <div>
                    <div>Something here</div>
                </div>
            </md-expansion-panel>

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
angular 4
material beta 8

willshowell:
I believe https://github.com/angular/material2/pull/5950 will fix this

willshowell:
It didn't entirely. Your example still requires

```scss
.mat-content {
  align-items: center;
}
```

josephperrott:
Resolved via: #5950

PR #6709, opened 2017-08-29
fix(expansion-panel): Adjust vertical spacing of expansion panel headers

josephperrott:
Completes fix for #6192


jelbourn:
I have a feeling that this depends on the header being an exact size, which may not work with @crisbeto's PR to add height inputs; @crisbeto thoughts?

crisbeto:
Yeah this won't work with a different height. Perhaps it should be refactored to use `position: absolute` and a `transform` for centering, instead of flexbox?

josephperrott:
The issue I found with using `position: absolute` and `transform: translate(..)` is that we are also using `transform` to perform the rotation.

I have now instead moved from using `::after` to using a `span` and performing all of the rotation on the icon itself rather than rotating the container of the icon.  I then put a margin-bottom on the container to offset the borders thickness.

PR #6710, opened 2017-08-29
docs(a11y): Adding accessibility sections to datepicker, menu, slide …

tinayuangao:
…toggle, slider, snackbar and toolbar

### DIFF
--- a/src/lib/datepicker/datepicker.md
+++ b/src/lib/datepicker/datepicker.md
@@ -239,44 +239,41 @@ should have a placeholder or be given a meaningful label via `aria-label`, `aria
 #### Keyboard shortcuts
 The keyboard shortcuts to handle datepicker are:
 
-`ALT` + `DOWN_ARROR`: trigger calendar dialog
-
-`ESCAPE`: close calendar dialog
+| Shortcut             | Action                            |
+|----------------------|-----------------------------------|
+| `ALT` + `DOWN_ARROW` | Open the calendar pop-up          |
+| `ESCAPE`             | Close the calendar pop-up         |
 
 
 In month view:
 
-`LEFT_ARROW`, `RIGHT_ARROW`, `UP_ARROW`, `DOWN_ARROW`: navigate through days
-
-`HOME`: focus the first day of the month
-
-`END`: focus the last day of the month
-
-`PAGE_UP`: goes to previous month
-
-`ALT` + `PAGE_UP`: goes to previous year
-
-`PAGE_DOWN`: goes to next month
-
-`ALT` + `PAGE_DOWN`: goes to next year
-
-`ENTER`: select current focused date
+| Shortcut             | Action                            |
+|----------------------|-----------------------------------|
+| `LEFT_ARROW`         | Goes to previous day              |
+| `RIGHT_ARROW`        | Goes to next day                  |
+| `UP_ARROW`           | Goes to previous 7 days           |
+| `DOWN_ARROW`         | Goes to next 7 days               |
+| `HOME`               | Focus the first day of the month  |
+| `END`                | Focus the last day of the month   |
+| `PAGE_UP`            | Goes to previous month            |
+| `ALT` + `PAGE_UP`    | Goes to previous year             |
+| `PAGE_DOWN`          | Goes to next month                |
+| `ALT` + `PAGE_DOWN`  | Goes to next year                 |
+| `ENTER`              | Select current date               |
 
 
 In year view:
 
-`LEFT_ARROW`, `RIGHT_ARROW`, `UP_ARROW`, `DOWN_ARROW`: navigate through months
-
-`HOME`: focus the first month of the year
-
-`END`: focus the last month of the year
-
-`PAGE_UP`: goes to previous year
-
-`ALT` + `PAGE_UP`: goes to previous 10 years
-
-`PAGE_DOWN`: goes to next year
-
-`ALT` + `PAGE_DOWN`: goes to next 10 years
-
-`ENTER`: select current focused month
+| Shortcut             | Action                            |
+|----------------------|-----------------------------------|
+| `LEFT_ARROW`         | Goes to previous month            |
+| `RIGHT_ARROW`        | Goes to next month                |
+| `UP_ARROW`           | Goes to previous 6 months         |
+| `DOWN_ARROW`         | Goes to next 6 months             |
+| `HOME`               | Focus the first month of the year |
+| `END`                | Focus the last month of the year  |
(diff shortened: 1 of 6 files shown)


src/lib/snack-bar/snack-bar.md
```diff
@@ -72,3 +72,7 @@ export class MessageArchivedComponent {
   constructor(@Inject(MD_SNACK_BAR_DATA) public data: any) { }
 }
 ```
+### Accessibility
+A Snackbar automatically goes away after a short time, so you can't count on the user seeing the
```
kara: Nit: To be consistent with the rest of the doc, use `snack-bar` instead of `Snackbar` ?

tinayuangao: Changed to `snack-bar`. Thanks for review.

kara:
LGTM

src/lib/datepicker/datepicker.md
```diff
@@ -228,3 +228,6 @@ application root module.
 })
 export class MyApp {}
 ```
+### Accessibility
+The input box for datepicker should have a placeholder or be given a meaningful label via
```
jelbourn: I'd like to expand this to describe what the datepicker does in general (input with aria-haspopup that triggers a `role="dialog"`) along with the keyboard shortcuts, but that can go in another PR.

src/lib/datepicker/datepicker.md
```diff
@@ -228,3 +228,6 @@ application root module.
...
+`aria-label` or `aria-labelledby`.
```
jelbourn: We should mention that `MdDatepickerIntl` includes strings that are used for `aria-label`s.

src/lib/snack-bar/snack-bar.md
```diff
@@ -72,3 +72,7 @@ export class MessageArchivedComponent {
   constructor(@Inject(MD_SNACK_BAR_DATA) public data: any) { }
 }
 ```
+### Accessibility
+A snack-bar automatically goes away after a short time, so you can't count on the user seeing the
```
jelbourn: I would phrase this as
```markdown
Snack-bar messages are announced via an `aria-live` region. Focus is not moved to
the snack-bar element, as this would be disruptive to a user in the middle of a
workflow. For any action offered in the snack-bar, the application should offer the
user an alternative way to perform the action (typically via keyboard shortcut).
```

src/lib/toolbar/toolbar.md
```diff
@@ -49,3 +49,7 @@ easily accomplished with `display: flex`:
 The color of a `<md-toolbar>` can be changed by using the `color` property. By default, toolbars
 use a neutral background color based on the current theme (light or dark). This can be changed to 
 `'primary'`, `'accent'`, or `'warn'`.  
+
+### Accessibility
+Toolbar items without text or labels should be given a meaningful label via `aria-label` or
```
jelbourn: I think this isn't quite right. Similar to `list` and `grid-list`, we treat toolbar as decorative-only by default. Generally, the toolbar is used as a header where `role="header"` would be appropriate. 

Only if the use-case of the toolbar match that of [`role="toolbar"`](https://www.w3.org/TR/wai-aria/roles#toolbar), the user should add the role and an appropriate label.

src/lib/datepicker/datepicker.md
```diff
@@ -228,3 +228,48 @@ application root module.
...
 ```
+### Accessibility
+The `MdDatepickerInput` directive adds `aria-haspopup` attribute to the native input element, and it
+triggers a calendar dialog with `role="dialog"`.
+`MdDatepickerIntl` includes strings that are used for `aria-label`s. The datepicker input
+should have a placeholder or be given a meaningful label via `aria-label` or `aria-labelledby`.
+
+#### Keyboard shortcuts
+The keyboard shortcuts to handle datepicker are:
+`ALT` + `DOWN_ARROR`: trigger calendar dialog
```
jelbourn: `ARROR` -> `ARROW`

tinayuangao:

src/lib/datepicker/datepicker.md
```diff
@@ -228,3 +228,55 @@ application root module.
...
 ```
+### Accessibility
+The `MdDatepickerInput` directive adds `aria-haspopup` attribute to the native input element, and it
+triggers a calendar dialog with `role="dialog"`.
+
+`MdDatepickerIntl` includes strings that are used for `aria-label`s. The datepicker input
+should have a placeholder or be given a meaningful label via `aria-label`, `aria-labelledby` or
+`MdDatepickerIntl`.
+
+#### Keyboard shortcuts
```
jelbourn: It would be good to put these in a table, e.g.
```markdown
| Shortcut             | Action                    |
|----------------------|---------------------------|
| `ALT` + `DOWN_ARROW` | Open the calendar pop-up  |
| `ESCAPE`             | Close the calendar pop-up |

```

src/lib/datepicker/datepicker.md
```diff
@@ -228,3 +228,55 @@ application root module.
...
+The keyboard shortcuts to handle datepicker are:
+
+`ALT` + `DOWN_ARROR`: trigger calendar dialog
```
jelbourn: `ARROR` -> `ARROW`

src/lib/datepicker/datepicker.md
```diff
@@ -228,3 +228,52 @@ application root module.
...
+
+
+In month view:
+
+| Shortcut             | Action                            |
+|----------------------|-----------------------------------|
+| `LEFT_ARROW`         | Goes to previous day              |
+| `RIGHT_ARROW`        | Goes to next day                  |
+| `UP_ARROW`           | Goes to previous 7 days           |
+| `DOWN_ARROW`         | Goes to next 7 days               |
```
jelbourn: `Goes` -> `Go` for each of these

For up and down arrow, I'd say `Go to same day in the next week`

mmalerba:
LGTM on datepicker and slider

tinayuangao:
Put shortcuts in tables. Please take another look. Thanks!

src/lib/datepicker/datepicker.md
```diff
@@ -228,3 +228,52 @@ application root module.
...
+| `HOME`               | Focus the first day of the month  |
+| `END`                | Focus the last day of the month   |
```
jelbourn: Why say `Focus` for home/end and `Goes` / `Go` everywhere else?

tinayuangao:
Changed to `Go to` everywhere. Please take another look. Thanks!

PR #6712, opened 2017-08-29
feat(sidenav): Add support for fixed sidenavs

### LINKED ISSUE #778: md-sidenav-layout should not prevent use of position:fixed
package version: Alpha5.2

md-sidenav-layout seems to work with with css transforms what in turn breaks position:fixed as this establishes a new viewport.

This makes it impossible to place floating elements inside content in this layout.

I use the content as my router-outlet (what probably many will do). In my content I want to use a floating md-fab button which I now have to place in the main application component. this sucks.

### LINKED ISSUE #998: (sidenav) elements with `position: fixed` inside `md-sidenav-layout`
Because of `transform: translate3d(0, 0, 0)` on `md-sidenav-layout` any element with `position: fixed` inside it will act as if it had `position: absolute` or `static` instead. It's an [expected behaviour](http://stackoverflow.com/a/15256339/301596), but it's annoying nevertheless.

Could we either remove that `transform` from `md-sidenav-layout`(it's only there because of performance I suppose), or add more examples to `Portal` and mention it as a workaround in docs, since portal could probably solve this by basically putting the element that needs to be fixed next to the `md-sidenav-layout` in the dom tree.

cc @hansl @jelbourn

### LINKED ISSUE #1179: fullscreen prevents address bar from collapsing in ios safari

Bug

The address bar in the mobile safari browser should collapse while scrolling down.
Tapping at the top of the address bar should browse to the top.

Using fullscreen in md-sidenav-layout prevents the address bar from collapsing while scrolling.
Tapping at the top of the address bar has no effects.

Scrolling the site down. Tapping at the address bar.

The motivation is to not corrupt the native experience of iOS.

Angular2 RC4
Angular2 Material alpha 6
iOS 9.3
Mobile safari

Thanks.

### LINKED ISSUE #1515: [sidenav @alpha 9] Regression in styling: sidenav positioning is no longer fixed

Bug

`md-sidenav` should remain visible (fixed) even as the user scrolls down the `md-sidenav-layout`.  This is how `alpha 7 or 8` worked.

`alpha 9` has regressed. Now, when a user scrolls down to reveal more `md-sidenav-layout` content, `md-sidenav` goes off the page.

At top of the page; sidenav open:

After scrolling down; sidenav open:

- [open this plunker](http://plnkr.co/edit/n2tscYvP3jrQwOuvF8kJ?p=preview)
- scroll down the page
- open the sidenav
- the side menu is not visible.

the `md-sidenav` typically shows a menu or other content that should be visible whenever the panel opens. Currently, the user experience is very poor: the user thinks the site is broken when she opens the sidenav and sees no content (scrolling up is required to reveal the content)

Early alphas had this problem, but it was fixed probably around `alpha 5`. There has been a regression and I noticed the problem on `alpha 9-3` on `Firefox 49`

### LINKED ISSUE #3031: Sidenav and fixed elements
Bug

fixed elements to be fixed without css tricks

To keep elements fixed (such as toolbar on the top, floating buttons on on the bottom) I need to set `height: 100%` on `html`, `body`, `main`, and `md-sidenav-container` (and any other potential parents). I have to set `overflow-y: auto` on the `md-sidenav-container` too.

If any of those properties is omitted, then toolbar & floating buttons `fixed` property is now taken as `absolute`.

Just look [this sample](https://material2-app.firebaseapp.com/) and try to fix the toolbar ; or jsut get an empty app with a sidenav and a toolbar. Unfortunately I can't provide any plunker (proxy doesn't allow it).

It seems to be a misfunction of the component and not the actual expected behavior. Plus, those css tricks I mentioned cause side-effets, especially on scroll event caption.

All I tested. From the alpha 10 to the current version, beta 2. I should have post this earlier, but I thought someone else would have find it. My bad.

If this issue is known, please ignore this, but I couldn't find any related issue.

mmalerba:
temporary demo: https://mmalerba-demo5.firebaseapp.com/sidenav

fixes #3031
fixes #1515
fixes #1179
fixes #998
fixes #778


src/lib/sidenav/drawer.ts
```diff
@@ -22,7 +22,7 @@ import {
   NgZone,
   OnDestroy,
   Inject,
-  ChangeDetectorRef,
+  ChangeDetectorRef, ContentChild, forwardRef,
```
jelbourn: Reformat?

src/lib/sidenav/drawer.ts
```diff
@@ -47,6 +48,37 @@ export class MdDrawerToggleResult {
...
+    'class': 'mat-drawer-content',
+    '[style.marginLeft.px]': '_margins.left',
+    '[style.marginRight.px]': '_margins.right',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+})
+export class MdDrawerContent implements AfterContentInit {
+  /** Margins to be applied to the content. */
+  _margins: {left: number, right: number} = {left: 0, right: 0};
```
jelbourn: Why are these margins here? Add explanation for why this is a thing?

src/lib/sidenav/drawer.ts
```diff
@@ -47,6 +48,37 @@ export class MdDrawerToggleResult {
...
+
+  constructor(
+      private _changeDetectorRef: ChangeDetectorRef,
+      @Inject(forwardRef(() => MdDrawerContainer)) private _container: MdDrawerContainer) {
+  }
+
+  ngAfterContentInit() {
+    this._container._contentMargins.subscribe((margins) => {
```
jelbourn: nit: don't need parents around `margins`

src/lib/sidenav/drawer.ts
```diff
@@ -155,6 +193,8 @@ export class MdDrawer implements AfterContentInit, OnDestroy {
   /** @deprecated */
   @Output('align-changed') onAlignChanged = new EventEmitter<void>();
 
+  _modeChanged = new Subject();
```
jelbourn: Add comment explaining why this is a thing?

src/lib/sidenav/drawer.ts
```diff
@@ -467,29 +520,33 @@ export class MdDrawerContainer implements AfterContentInit {
...
+   * Recalculates and updates the inline styles for the content. Note that this should be used
+   * sparingly, because it causes a reflow.
    */
-  private _getDrawerEffectiveWidth(drawer: MdDrawer, mode: string): number {
-    return (this._isDrawerOpen(drawer) && drawer.mode == mode) ? drawer._width : 0;
-  }
+  private _updateContentMargins() {
+    let left = 0;
+    let right = 0;
+
```
jelbourn: Add comment that describes which style is applied where based on the mode + position?

src/lib/sidenav/sidenav.ts
```diff
@@ -46,18 +68,35 @@ import {animate, state, style, transition, trigger} from '@angular/animations';
...
+    '[style.top.px]': 'fixedInViewport ? fixedTopGap : null',
+    '[style.bottom.px]': 'fixedInViewport ? fixedBottomGap : null',
   },
   changeDetection: ChangeDetectionStrategy.OnPush,
   encapsulation: ViewEncapsulation.None,
 })
-export class MdSidenav extends MdDrawer {}
+export class MdSidenav extends MdDrawer {
+  /** Whether the sidenav is fixed in the viewport. */
+  @Input() fixedInViewport = true;
```
jelbourn: Use `coerceBooleanProperty`

src/lib/sidenav/sidenav.ts
```diff
@@ -46,18 +68,35 @@ import {animate, state, style, transition, trigger} from '@angular/animations';
...
+
+  /**
+   * The gap between the top of the sidenav and the top of the viewport when the sidenav is in fixed
+   * mode.
+   */
+  @Input() fixedTopGap = 0;
```
jelbourn: Use `coerceNumberProperty`

src/lib/sidenav/sidenav.ts
jelbourn: Add unit tests for these new fixed properties?

mmalerba:
@jelbourn comments addressed

axtho:
Can we get this into the material2-builds?

KKHAN1991:
Do you have the source code for the working version in the demo?

kenji-1996:
@KKHAN1991 I looked into the css of the working example

```
.footer_fixed {
  display: flex;
  box-sizing: border-box;
  width: 100%;
  padding: 0 16px;
  flex-direction: column;
}
```

and then apply this to your footer item like so:

```
<mat-toolbar class="footer_fixed">
  <div class="container">
    <span>this is a toolbar</span>
  </div>
</mat-toolbar>
```

Then I just call my app-footer (i made a component for this) inside my sidebar nav and below my navbar like so;

```
<mat-sidenav-container fullscreen>
  <mat-sidenav mode="push" #sidenav>
    <div fxLayout="column">
      .. sidenav content ..
  </mat-sidenav>

  <mat-toolbar color="primary" class="fixed-navbar mat-elevation-z10">
    .. material toolbar content ..
  </mat-toolbar>

  ..container for inside elements
  <app-container></app-container>
  <footer>
    .. my footer element inside footer tags ..
    <app-footer></app-footer>
  </footer>

</mat-sidenav-container>
```

Hope this helps, ive spent the past 4 hours fixing problems and getting it working so i hope i can save someone time

axtho:
@KKHAN1991 here is the demo: https://material2-dev.firebaseapp.com/sidenav, and here are the files that produce it: https://github.com/angular/material2/tree/master/src/demo-app/sidenav. It works well.

Splaktar:
Is there any documentation for this feature other than just `@Input() fixedInViewport boolean | Whether the sidenav is fixed in the viewport.` in the API section of the sidenav docs?

mmalerba:
@Splaktar soon https://github.com/angular/material2/pull/7775

PR #6736, opened 2017-08-30
fix(dialog): multiple dialogs not opening with disabled animations

### LINKED ISSUE #6719: Not possible to open second dialog if NoopAnimationsModule is used
Bug

It should be possible to open multiple dialogs even if NoopAnimationsModule is used. It works if BrowserAnimationsModule is used. 

If you try to open a second dialog nothing happens

http://plnkr.co/edit/K8SwE9LJeVnVleUic8X3?p=preview

angular@4.3.4, angular-material@beta-10

This pull request looks like a suspect: https://github.com/angular/material2/pull/5769

DevVersion:
With SHA 36f708c8ddc6fb78fb0ad4fb4fc1969eae183ebb, second dialogs can be only opened if the first dialog finished animating.

Due to the recent update to Angular 4.3, the animation trigger callbacks are firing in a wrong order (`done` callback fires before `start`), which causes the `_isAnimating` property to be set to `true` while the animation already finished.

@crisbeto Not too happy how the tests ended up. 

Fixes #6719

### DIFF
--- a/src/lib/datepicker/datepicker.spec.ts
+++ b/src/lib/datepicker/datepicker.spec.ts
@@ -7,7 +7,7 @@ import {
   dispatchMouseEvent,
 } from '@angular/cdk/testing';
 import {Component, ViewChild} from '@angular/core';
-import {async, ComponentFixture, inject, TestBed} from '@angular/core/testing';
+import {async, ComponentFixture, inject, TestBed, fakeAsync, flush} from '@angular/core/testing';
 import {FormControl, FormsModule, ReactiveFormsModule} from '@angular/forms';
 import {
   DEC,
@@ -166,7 +166,7 @@ describe('MatDatepicker', () => {
             .toBe(true, 'Expected default ESCAPE action to be prevented.');
       });
 
-      it('close should close dialog', () => {
+      it('close should close dialog', fakeAsync(() => {
         testComponent.touch = true;
         fixture.detectChanges();
 
@@ -177,13 +177,13 @@ describe('MatDatepicker', () => {
 
         testComponent.datepicker.close();
         fixture.detectChanges();
+        flush();
 
-        fixture.whenStable().then(() => {
-          expect(document.querySelector('mat-dialog-container')).toBeNull();
-        });
-      });
+        expect(document.querySelector('mat-dialog-container')).toBeNull();
+        flush();
+      }));
 
-      it('setting selected should update input and close calendar', () => {
+      it('setting selected should update input and close calendar', fakeAsync(() => {
         testComponent.touch = true;
         fixture.detectChanges();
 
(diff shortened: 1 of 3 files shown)


src/lib/datepicker/datepicker.spec.ts
```diff
@@ -164,12 +164,16 @@ describe('MdDatepicker', () => {
...
-        fixture.whenStable().then(() => {
-          expect(document.querySelector('md-dialog-container')).toBeNull();
-        });
-      });
+        tick();
 
-      it('setting selected should update input and close calendar', () => {
+        expect(document.querySelector('md-dialog-container')).toBeNull();
+
+        // The fake async zone expects all active timers to be done at the end of the test.
```
crisbeto: I'm not sure that this kind of comment is necessary for every `tick` call.

src/lib/dialog/dialog-container.ts
```diff
@@ -182,8 +182,13 @@ export class MdDialogContainer extends BasePortalHost {
       this._restoreFocus();
     }
 
-    this._animationStateChanged.emit(event);
-    this._isAnimating = false;
+    // Note: as of Angular 4.3, the animations module seems to fire the `start` callback before
+    // the end if animations are disabled. Make this call async to ensure that it still fires
+    // at the appropriate time.
+    Promise.resolve().then(() => {
```
crisbeto: Add a test for this fix? The current tests are using the `NoopAnimationsModule` and they didn't catch the issue.

src/lib/datepicker/datepicker.spec.ts
DevVersion: Yeah totally see what you mean. It's just that it can be confusing if we place a `tick(500)` after every test.

src/lib/dialog/dialog-container.ts
DevVersion: I looked into a way to reproduce that issue inside of the `TestBed`, but it looks like the animation callbacks fire correctly there.

I tried to look into the `NoopAnimationPlayer` from `@angular/animations` and I assume its related to some timing issues with zone.js

TomDemulierChevret:
Why hasn't this PR been merged and added to the last version ?
This bug induces blocking issues in any app with complex user interactions involving multiple dialog.

The PR has been merge-ready for a while..
Can you merge it quickly so we can at least get a recent build from https://github.com/angular/material2-builds and not stay behind the last release for another month ?

jelbourn:
@mmalerba is working on resolving presubmit failures

mmalerba:
The only presubmit issue left is a screenshot test that shows a second copy of the same dialog opened on top of itself. Do we know if this PR is still necessary or if it can be fixed by a change in animations?

TomDemulierChevret:
@mmalerba I juste tested with angular 4.4.4 and angular-material beta 11 and the issue is still present.
So yes, this PR is still needed.

owen26:
Any progress on this issue? It really is blocking our product dev work for IE/Edge browsers.

nfoenki:
I'm facing the same issue on Safari (Desktop & Mobile). It's a big issue !
@DevVersion 's fix is working fine
**Can it be merged quickly please ??**

jelbourn:
We're going to resolve this issue with #8051 instead

Issue #6542, opened 2017-08-18
[slider] change detection issue with styles


willshowell:
#### Bug, feature request, or proposal:
Bug

#### Current

Note that the slider track does not update its position until mouseover

![slider-cd-issue](https://user-images.githubusercontent.com/6475576/29466844-519d9d54-840c-11e7-9705-aa837dcf818a.gif)

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Latest build:  0113c38

jelbourn:
Potentially related to becoming `OnPush`

PR #6744, opened 2017-08-30
fix(slider): not reacting to directionality changes

crisbeto:
Fixes the slider not updating if its direction is changed dynamically. This is something I came across while testing #6641.


mmalerba:
I believe @g1shin is working on this in her PR for #6542

crisbeto:
#6542 seems to be related to the focus styles, not the directionality, although the fix should be somewhat similar (adding in a missing `markForCheck` somewhere).

mmalerba:
yeah she's just fixing the various OnPush issues

crisbeto:
Alright, closing in that case.

PR #6751, opened 2017-08-31
docs(a11y): Add a11y docs for table, expansion panel, dialog, chips, autocomplete…

tinayuangao:
…, and select

### DIFF
--- a/src/lib/chips/chips.md
+++ b/src/lib/chips/chips.md
@@ -40,5 +40,5 @@ use a neutral background color based on the current theme (light or dark). This 
 `'primary'`, `'accent'`, or `'warn'`.
 
 ### Accessibility
-Chips have role="option", while the chip list has `role="listbox"`. The chip input should have a
-placeholder or be given a meaningful label via `aria-label` or `aria-labelledby`.
+A chip-list behaves as a `role="listbox"`, with each chip being a `role="option"`. The chip input
+should have a placeholder or be given a meaningful label via `aria-label` or `aria-labelledby`.
--- a/src/lib/chips/chips.md
+++ b/src/lib/chips/chips.md
@@ -38,3 +38,7 @@ also gain focus when clicked, ensuring keyboard navigation starts at the appropr
 The selected color of an `<md-chip>` can be changed by using the `color` property. By default, chips
 use a neutral background color based on the current theme (light or dark). This can be changed to 
 `'primary'`, `'accent'`, or `'warn'`.
+
+### Accessibility
+Chips have role="option", while the chip list has `role="listbox"`. The chip input should have a
+placeholder or be given a meaningful label via `aria-label` or `aria-labelledby`.
--- a/src/lib/dialog/dialog.md
+++ b/src/lib/dialog/dialog.md
@@ -128,8 +128,20 @@ export class AppModule() {}
 ```
 
 ### Accessibility
-The dialog has `role="dialog"`, and dialog role can be overwritten to `alertdialog` in the
-appropriate context.
+By default, each dialog has `role="dialog"` on the root element. The role can be changed to
+`alertdialog` via the `MdDialogConfig` when opening.
 
-The dialog will focus on the first tabbable element when opened,
-and restore focus to the element that was focused before the dialog opened when closed.
+The `aria-label`, `aria-labelledby`, and `aria-describedby` attributes can all be set to the
+dialog element via the `MdDialogConfig` as well. Each dialog should typically have a label
+set via `aria-label` or `aria-labelledby`.
+
+#### Focus management
+By default, the first tabbable element within the dialog will receive focus upon open.
+
+Tabbing through the elements of the dialog will keep focus inside of the dialog element,
+wrapping back to the first tabbable element when reaching the end of the tab sequence.
+
+#### Keyboard interaction
+By default pressing the escape key will close the dialog. While this behavior can
+be turned off via the `disableClose` option, users should generally avoid doing so
+as it breaks the expected interaction pattern for screen-reader users.
--- a/src/lib/dialog/dialog.md
+++ b/src/lib/dialog/dialog.md
@@ -126,3 +126,10 @@ that the AOT compiler knows to create the `ComponentFactory` for it.
 })
 export class AppModule() {}
 ```
+
+### Accessibility
+The dialog has `role="dialog"`, and dialog role can be overwritten to `alertdialog` in the
+appropriate context.
+
+The dialog will focus on the first tabbable element when opened,
+and restore focus to the element that was focused before the dialog opened when closed.
--- a/src/lib/expansion/expansion.md
+++ b/src/lib/expansion/expansion.md
@@ -98,3 +98,10 @@ panel can be expanded at a given time:
 
 </md-accordion>
 ```
+
+### Accessibility
+The expansion panel header has `role="button"`. The expansion panel header has attribute
+`aria-controls` with the expansion panel's id as value.
+
+The expansion panel headers are buttons. Users can use the keyboard to activate the expansion panel
+header to switch between expanded state and collapsed state.
--- a/src/lib/expansion/expansion.md
+++ b/src/lib/expansion/expansion.md
@@ -100,8 +100,10 @@ panel can be expanded at a given time:
 ```
 
 ### Accessibility
+The expansion-panel aims to mimic the experience of the native `<details>` and `<summary>` elements.
 The expansion panel header has `role="button"`. The expansion panel header has attribute
 `aria-controls` with the expansion panel's id as value.
 
 The expansion panel headers are buttons. Users can use the keyboard to activate the expansion panel
-header to switch between expanded state and collapsed state.
+header to switch between expanded state and collapsed state. Because the header acts as a button,
+additional interactive elements should not be put inside of the header.
--- a/src/lib/input/input.md
+++ b/src/lib/input/input.md
@@ -154,7 +154,7 @@ Here are the available global options:
 ### Accessibility
 The `mdInput` directive works with native `<input>` to provide an accessible experience.
 
-If a placeholder attributed is added to the input, or a `md-placeholder` element is added
+If a placeholder attribute is added to the input, or a `md-placeholder` element is added
 in the form field, the placeholder text will automatically be used as the label for the input.
 If there's no placeholder specified, `aria-label`, `aria-labelledby` or `<label for=...>` should be
 added.
--- a/src/lib/input/input.md
+++ b/src/lib/input/input.md
@@ -150,3 +150,13 @@ Here are the available global options:
 | Name              | Type     | Description |
 | ----------------- | -------- | ----------- |
 | errorStateMatcher | Function | Returns a boolean specifying if the error should be shown |
+
+### Accessibility
+The `mdInput` directive works with native `<input>` to provide an accessible experience.
+
+If a placeholder attributed is added to the input, or a `md-placeholder` element is added
+in the form field, the placeholder text will automatically be used as the label for the input.
+If there's no placeholder specified, `aria-label`, `aria-labelledby` or `<label for=...>` should be
+added.
+
+Any `md-error` and `md-hint` are automatically added to the input's `aria-describedby`.
--- a/src/lib/select/select.md
+++ b/src/lib/select/select.md
@@ -92,3 +92,9 @@ Here are the available global options:
 - <kbd>DOWN_ARROW</kbd>: Focus next option
 - <kbd>UP_ARROW</kbd>: Focus previous option
 - <kbd>ENTER</kbd> or <kbd>SPACE</kbd>: Select focused item
+
+### Accessibility
+The select component without text or label should be given a meaningful label via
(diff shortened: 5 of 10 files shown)


src/lib/tabs/tabs.md
```diff
@@ -79,3 +79,7 @@ provides a tab-like UI for navigating between routes.
 The tab-nav-bar is not tied to any particular router; it works with normal `<a>` elements and uses
 the `active` property to determine which tab is currently active. The corresponding
 `<router-outlet>` can be placed anywhere in the view.
+
+### Accessibility
+`MdTanNav`s without text or labels should be given a meaningful label via `aria-label` or
```
rafaelss95: Typo: `MdTabNav`.

src/lib/input/input.md
```diff
@@ -150,3 +150,6 @@ Here are the available global options:
 | Name              | Type     | Description |
 | ----------------- | -------- | ----------- |
 | errorStateMatcher | Function | Returns a boolean specifying if the error should be shown |
+
+### Accessibility
+The `mdInput` directive works with native `<input>` to provide an accessible experience.
```
mmalerba: Should also mention that if you add a `placeholder` attribute to the input or a `md-placeholder` element in the form field that placeholder text will automatically be used as the label for the input. If you don't specify any placeholder you need to add `aria-label` `aria-labeledby` or `<label for=...>`

Possibly also worth mentioning that any `md-error` and `md-hint` are automatically added to the input's `aria-describedby`

src/lib/chips/chips.md
```diff
@@ -38,3 +38,7 @@ also gain focus when clicked, ensuring keyboard navigation starts at the appropr
 The selected color of an `<md-chip>` can be changed by using the `color` property. By default, chips
 use a neutral background color based on the current theme (light or dark). This can be changed to 
 `'primary'`, `'accent'`, or `'warn'`.
+
+### Accessibility
+Chips has `role="option"` while chip list has `role="listbox"`. The chip input should have a
```
crisbeto: Chips have `role="option"`, while the chip list has...

src/lib/dialog/dialog.md
```diff
@@ -126,3 +126,7 @@ that the AOT compiler knows to create the `ComponentFactory` for it.
 })
 export class AppModule() {}
 ```
+
+### Accessibility
+The dialog has `role="dialog". The dialog will focus on the first tabbable element when opened,
```
crisbeto: This should mention that the dialog role can be overwritten to `alertdialog` in the appropriate context.

src/lib/select/select.md
```diff
@@ -92,3 +92,11 @@ Here are the available global options:
 - <kbd>DOWN_ARROW</kbd>: Focus next option
 - <kbd>UP_ARROW</kbd>: Focus previous option
 - <kbd>ENTER</kbd> or <kbd>SPACE</kbd>: Select focused item
+
+### Accessibility
+The select component without text or label should be given a meaningful label via
```
mmalerba: FYI, the placeholder, hint, and error stuff will work the same way as input after the refactor. For now this is fine though

src/lib/expansion/expansion.md
```diff
@@ -98,3 +98,10 @@ panel can be expanded at a given time:
 
 </md-accordion>
 ```
+
+### Accessibility
+The expansion panel header has `role="button"`. The expansion panel header has attribute
+`aria-controls` with the expansion panel's id as value.
+
+Users can use keyboard to activate the expansion panel header to switch between expanded state
```
crisbeto: "can use keyboard" -> "can use the keyboard".

src/lib/select/select.md
```diff
@@ -92,3 +92,11 @@ Here are the available global options:
...
+`aria-label` or `aria-labelledby`.
+
+The select component has `role="listbox"` and options inside select has `role="option"`.
+Autocomplete trigger sets `aria-owns` to the autocomplete's id, and sets `aria-activedescendant` to
```
crisbeto: This one talks about autocomplete instead of select. Also select doesn't use `aria-activedescendant`, but it moves focus between the options.

src/lib/sort/sort.md
```diff
@@ -32,4 +32,7 @@ To prevent the user from clearing the sort sort state from an already sorted col
 When used on an `md-table` header, it is not required to set an `md-sort-header` id on because
 by default it will use the id of the column.
 
-<!-- example(table-sorting) -->
\ No newline at end of file
+<!-- example(table-sorting) -->
+
+### Accessibility
+The `aria-label` for sort button can be set in `MdSortHeaderIntl`.
```
crisbeto: "the `aria-label` for sort button" -> "the `aria-label` for the sort button".

tinayuangao:

src/lib/dialog/dialog.md
```diff
@@ -126,3 +126,10 @@ that the AOT compiler knows to create the `ComponentFactory` for it.
 })
 export class AppModule() {}
 ```
+
+### Accessibility
+The dialog has `role="dialog", and dialog role can be overwritten to `alertdialog` in the
```
crisbeto: Seems like you're missing a backtick after the `role`.

src/lib/select/select.md
```diff
@@ -92,3 +92,9 @@ Here are the available global options:
 - <kbd>DOWN_ARROW</kbd>: Focus next option
 - <kbd>UP_ARROW</kbd>: Focus previous option
 - <kbd>ENTER</kbd> or <kbd>SPACE</kbd>: Select focused item
+
+### Accessibility
+The select component without text or label should be given a meaningful label via
+`aria-label` or `aria-labelledby`.
+
+The select component has `role="listbox"` and options inside select has `role="option"`.
```
crisbeto: has -> have

src/lib/expansion/expansion.md
```diff
@@ -98,3 +98,10 @@ panel can be expanded at a given time:
 
 </md-accordion>
 ```
+
+### Accessibility
+The expansion panel header has `role="button"`. The expansion panel header has attribute
+`aria-controls` with the expansion panel's id as value.
+
+Users can use the keyboard to activate the expansion panel header to switch between expanded state
```
josephperrott: Is it worth noting that as the panel headers are buttons, they are able to be navigated to via keyboard as well?

tinayuangao:

mmalerba:
LGTM for input

src/lib/dialog/dialog.md
```diff
@@ -126,3 +126,10 @@ that the AOT compiler knows to create the `ComponentFactory` for it.
 })
 export class AppModule() {}
 ```
+
+### Accessibility
+The dialog has `role="dialog"`, and dialog role can be overwritten to `alertdialog` in the
+appropriate context.
```
jelbourn:
```markdown
By default, each dialog has `role="dialog"` on the root element. The role can be changed to
`alertdialog` via the `MdDialogConfig` when opening.

The `aria-label`, `aria-labelledby`, and `aria-describedby` attributes can all be set to the
dialog element via the `MdDialogConfig` as well. Each dialog should typically have a label
set via `aria-label` or `aria-labelledby`.

#### Focus management
By default, the first tabbable element within the dialog will receive focus upon open. 

Tabbing through the elements of the dialog will keep focus inside of the dialog element, 
wrapping back to the first tabbable element when reaching the end of the tab sequence.
```
(ariaLabel being added in #6558)

We can expand this in a later PR to mention the focus control attributes from `cdkFocusTrap` once we have documentation for that

src/lib/expansion/expansion.md
```diff
@@ -98,3 +98,10 @@ panel can be expanded at a given time:
 
 </md-accordion>
 ```
+
+### Accessibility
+The expansion panel header has `role="button"`. The expansion panel header has attribute
```
jelbourn: I'd add a sentence at the beginning like
```
The expansion-panel aims to mimic the experience of the native `<details>` and `<summary>` elements.
```

src/lib/expansion/expansion.md
```diff
@@ -98,3 +98,10 @@ panel can be expanded at a given time:
...
+`aria-controls` with the expansion panel's id as value.
+
+The expansion panel headers are buttons. Users can use the keyboard to activate the expansion panel
+header to switch between expanded state and collapsed state.
```
jelbourn: I'd add another sentence 
```markdown
Because the header acts as a button, additional interactive elements
should not be put inside of the header.
```

src/lib/table/table.md
```diff
@@ -45,3 +45,9 @@ an update via the table's data source.
 In the near future, we will provide a simplified version of the data-table with an easy-to-use
 interface, material styling, array input, and more out-of-the-box features (sorting, pagination,
 and selection).
+
+### Accessibility
+Tables without text or labels should be given a meaningful label via `aria-label` or
+`aria-labelledby`.
+
+Table's default role is 'grid', and it can be changed through `role` attribute.
```
jelbourn: We should mention that it can be changed to `treegrid` specifically, since that's the only other role that makes sense (being that it contains rows and gridcells)

src/lib/table/table.md
```diff
@@ -45,3 +45,9 @@ an update via the table's data source.
 In the near future, we will provide a simplified version of the data-table with an easy-to-use
 interface, material styling, array input, and more out-of-the-box features (sorting, pagination,
 and selection).
+
+### Accessibility
+Tables without text or labels should be given a meaningful label via `aria-label` or
+`aria-labelledby`.
```
jelbourn: We should mention `aria-readonly` regardless of what we do with it.

(it defaults to `true` if it's not set, but most people probably make readonly tables)

src/lib/table/table.md
jelbourn: It's probably also worth mentioning that the `md-table` does not manage any focus/keyboard interaction on its own; if users want to add this, it can be added in their application.

src/lib/tabs/tabs.md
```diff
@@ -79,3 +79,7 @@ provides a tab-like UI for navigating between routes.
 The tab-nav-bar is not tied to any particular router; it works with normal `<a>` elements and uses
 the `active` property to determine which tab is currently active. The corresponding
 `<router-outlet>` can be placed anywhere in the view.
+
+### Accessibility
+`MdTabNav`s without text or labels should be given a meaningful label via `aria-label` or
```
jelbourn: This is true for either tab variant, not just the nav-tabs. 

For nav-tabs, the `<nav>` element should probably have a label as well.

src/lib/tabs/tabs.md
```diff
@@ -79,3 +79,7 @@ provides a tab-like UI for navigating between routes.
...
+`aria-labelledby`.
```
jelbourn: Oh, and we should mention the keyboard shortcuts for `md-tab-group`

src/lib/dialog/dialog.md
```diff
@@ -126,3 +126,10 @@ that the AOT compiler knows to create the `ComponentFactory` for it.
...
+
+The dialog will focus on the first tabbable element when opened,
```
jelbourn: I forgot about escape, too:
```markdown
#### Keyboard interaction
By default pressing the escape key will close the dialog. While this behavior can
be turned off via the `disableClose` option, users should generally avoid doing so
as it breaks the expected interaction pattern for screen-reader users.
```

src/lib/input/input.md
```diff
@@ -150,3 +150,13 @@ Here are the available global options:
...
+
+If a placeholder attributed is added to the input, or a `md-placeholder` element is added
```
kara: Typo: attributed -> attribute

src/lib/chips/chips.md
```diff
@@ -38,3 +38,7 @@ also gain focus when clicked, ensuring keyboard navigation starts at the appropr
 The selected color of an `<md-chip>` can be changed by using the `color` property. By default, chips
 use a neutral background color based on the current theme (light or dark). This can be changed to 
 `'primary'`, `'accent'`, or `'warn'`.
+
+### Accessibility
+Chips have role="option", while the chip list has `role="listbox"`. The chip input should have a
+placeholder or be given a meaningful label via `aria-label` or `aria-labelledby`.
```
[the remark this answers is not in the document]
tinayuangao: The screen reader treats chip input and chip list as two separate components now.

tinayuangao:

src/lib/tabs/tabs.md
```diff
@@ -79,3 +79,16 @@ provides a tab-like UI for navigating between routes.
 The tab-nav-bar is not tied to any particular router; it works with normal `<a>` elements and uses
 the `active` property to determine which tab is currently active. The corresponding
 `<router-outlet>` can be placed anywhere in the view.
+
+### Accessibility
+Tabs without text or labels should be given a meaningful label via `aria-label` or
+`aria-labelledby`. For `MdTabNav`, the `<nav>` element should have a label as well.
+
+
+### Keyboard shortcuts
```
jelbourn: Should be level 4 header `####`

PR #6769, opened 2017-08-31
refactor: initialize to empty subscriptions

crisbeto:
Initializes most subscriptions to empty ones, avoiding the need to null-check them before unsubscribing.

**Note:** There are a few more candidates for refactor (namely the table and the scroll strategies), but in those cases we use the subscriptions to manage state.


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -146,10 +146,7 @@ export class MdTabNav extends _MdTabNavMixinBase implements AfterContentInit, Ca
 
   ngOnDestroy() {
     this._onDestroy.next();
-
-    if (this._resizeSubscription) {
-      this._resizeSubscription.unsubscribe();
-    }
+    this._resizeSubscription.unsubscribe();
```
willshowell: It looks like this one isn't even needed if you change the method to

```ts
ngOnDestroy() {
  this._onDestroy.next();
  this._onDestroy.complete();
}
```

PR #6770, opened 2017-08-31
chore(docs): create missing markdown files for cdk items

josephperrott:


src/cdk/keycodes/keycodes.md
```diff
@@ -0,0 +1,26 @@
...
+  selector: '[count-arrows]'
+  host: {
+    (keypress): 'handleKeyPress($event)'
+  }
+})
+export class ArrowCounterDirective {
+  arrowPressCount = 0;
+
+  handleKeyPress(event: KeyboardEvent) {
+    if (event.codeCode in [UP_ARROW, DOWN_ARROW, LEFT_ARROW, RIGHT_ARROW]) {
```
crisbeto: 1. `event.codeCode` -> `event.keyCode`.
2.  This check will look for a property that corresponds to `event.keyCode` in the `Array` prototype. The correct way to check it is `[UP_ARROW, DOWN_ARROW, LEFT_ARROW, RIGHT_ARROW].includes(event.keyCode)`.

src/cdk/overlay/overlay.md
```diff
@@ -0,0 +1,3 @@
+### Overlay
+
+A component used to manage overlays.
```
crisbeto: `Overlay` is a service, not a component.

src/cdk/rxjs/rxjs.md
```diff
@@ -0,0 +1,36 @@
+### RxJS Usage
+When dealing with RxJS operators, it is best to import the operator functions directly (e.g.
```
crisbeto: This should mention that avoiding the patch operators is only recommended when writing a library.

src/cdk/rxjs/rxjs.md
```diff
@@ -0,0 +1,36 @@
...
+```ts
+// Before
+someObservable.filter(...).map(...).do(...);
+
+// After
+RxChain.from(someObservable).call(filter, ...).call(map, ...).subscribe(...);
+```
+
+#### Available Operators
+Many operators are provided in via the `RxChain`, these operators include:
```
crisbeto: I don't think that we should include the list operators here, because it means that we'll have to maintain it.

josephperrott: Yeah, you're right.  I went back and forth on it before I put it in.

Issue #45, opened 2016-01-27
Design service for helping build responsive components


hansl:
We should have something that works both in headless rendering and browsers.

The solution should be something along those lines (pseudo code):

``` typescript
@Injectable()
export class MdLayoutService {
  onLayoutChange: EventEmitter<MdLayoutEnum> = new EventEmitter<MdLayoutEnum>();
  get currentLayout(): MdLayoutEnum {
    return this.layout_;
  }

  constructor(private layout_: MdLayoutEnum) {
    if (!headless) {  // <-- Tricky part
      window.matchMedia('...').addListener((mql) => this.onLayoutChange.emit(...));
    }
  }
};
```

You can then provide an `MdLayoutEnum` constant in headless browsing (or if you want to fix it in the browser or unit tests).

TODO: 
- [ ] Design Doc.

christophechevalier:
Many questions have been opened and close referencing this one.

I'm sorry to ask for ETA but is there any update on this topic ?

md-media is very handy and we need that to handle responsive in your app :)

Thanks

grizzm0:
I guess this can be closed?

For anyone else looking for this it's part of @angular/flex-layout.
See: https://github.com/angular/flex-layout/wiki/Adaptive-Layouts#core-directives--responsive-features

josephperrott:
Closed with #6772

PR #6772, opened 2017-08-31
feat(breakpoints): Create breakpoint management system

josephperrott:
#45


fxck:
Doesn't flex-layout have something like this?

jelbourn:
Flex-layout is going to change to consume what's in the cdk

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+      (listener: MediaQueryListListener) => {
+        mql.addListener((e: MediaQueryList) => this.zone.run(() => listener(e)));
+      },
+      (listener: MediaQueryListListener) => {
+        mql.removeListener((e: MediaQueryList) => this.zone.run(() => listener(e)));
+      }))
+      .call(takeUntil, this._activeSubject)
+      .call(startWith, mql)
+      .call(map, (nextMql: MediaQueryList) => {
+        return <BreakpointState>{query: query, matches: nextMql.matches};
```
jelbourn: I think you don't need this type coercion. Without it, the operator fits on one line
```ts
.call(map, (nextMql: MediaQueryList) => ({query: query, matches: nextMql.matches}))
```

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+      })
+      .result();
+
+    // Add the MediaQueryList to the set of queries.
+    let output = <Query>{observable: queryObservable, mql: mql};
```
jelbourn: Don't think you need this coercion either

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+  private _registerQuery(query: string): Query {
+    // Only set up a new MediaQueryList if it is not already being listened for.
+    if (this._queries.has(query)) {
+      return this._queries.get(query)!;
+    }
+    // Create new MediaQueryList.
+    let mql: MediaQueryList = this.mediaMatcher.matchMedia(query);
+    // Create callback for match changes and add it is as a listener.
+    let queryObservable = RxChain.from(fromEventPattern(
+      // Listener callback methods are wrapped to be placed back in ngZone.
```
jelbourn: Add background information about why `matchMedia` listeners escape the zone

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Registers a specific query to be listened for.
+   */
+  private _registerQuery(query: string): Query {
+    // Only set up a new MediaQueryList if it is not already being listened for.
+    if (this._queries.has(query)) {
+      return this._queries.get(query)!;
+    }
+    // Create new MediaQueryList.
```
jelbourn: This comment doesn't say anything that the code doesn't

src/cdk/layout/breakpoints.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+// tslint:disable-next-line:variable-name
```
jelbourn: Add explanation that we use PascalCase here because we want to treat this like an enum

src/cdk/layout/media-matcher.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      breakpointManager = bm;
+      mediaMatcher = mm;
+    }));
+
+  it('correctly returns a MediaQueryList to check for matches', () => {
+    expect(mediaMatcher.matchMedia('(min-width: 1px)').matches).toBeTruthy();
+    expect(mediaMatcher.matchMedia('(max-width: 1px)').matches).toBeFalsy();
+  });
+
+  it('adds css rules for provided queries when the platform is webkit.',
```
jelbourn: Add a test for _not_ adding the styles on other platforms?

src/cdk/layout/media-matcher.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Injectable} from '@angular/core';
+import {Platform} from '@angular/cdk/platform';
+
+/**
+ * Global registry for all dynamically-created, injected style tags.
+ */
+let styleElementForWebkitCompatibility: {[key: string]: HTMLStyleElement} = {};
```
jelbourn: Why not make this a `Map<string, HTMLStyleElement>`?

Also `const`

src/cdk/layout/media-matcher.ts
```diff
@@ -0,0 +1,67 @@
...
+
+@Injectable()
+export class MediaMatcher {
+  /** The internal matchMedia method to return back a MediaQueryList like object. */
+  private _matchMedia: (query: string) => MediaQueryList;
+
+  constructor(private platform: Platform) {
+    this._matchMedia = this.platform.isBrowser ?
+      (<any>window).matchMedia.bind((<any>window)) :
```
jelbourn: Why does this need `bind`?

jelbourn: Why does this need the `<any>` coercion? The typings look like they have `matchMedia` present

src/cdk/layout/media-matcher.ts
```diff
@@ -0,0 +1,67 @@
...
+      (query: string) => <MediaQueryList>{
```
jelbourn: Extract this into a `const` outside the class like `noopMatchMedia`? That way it can have it's own description like
```ts
/** No-op matchMedia replacement for non-browser platforms that... */
```

src/cdk/layout/media-matcher.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+/**
+ * For Webkit engines that only trigger the MediaQueryListListener when there is at least one CSS
+ * selector for the respective media query.
+ */
+function createEmptyStyleRule(query: string) {
+  if (!styleElementForWebkitCompatibility[query]) {
+    try {
```
jelbourn: Why the `try-catch`? Especially when the `catch` still logs the error

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+   */
+  isMatched(value: BreakpointQuery): boolean {
+    let queries = coerceArray(value);
+    return queries.some((mediaQuery: string) => {
+      return this._registerQuery(mediaQuery).mql.matches;
+    });
+  }
+
+  /**
+   * Retrieves an observable for the BreakpointQuery, broadcasting changes to the matching state.
```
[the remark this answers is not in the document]
josephperrott: Updated, tweaked a little bit as I wanted to be sure to generalize that it matches all of the possible things in an `@media` query, not only the pieces what we are currently using it for.

src/cdk/layout/media-matcher.spec.ts
josephperrott: I'm not sure how to go about this because it depends on what platform the test is being run on.

src/cdk/layout/media-matcher.ts
josephperrott: We are unable to bind the `window.matchMedia` method as a property of our custom object as it results in an `illegal invocation`.
So we instead `bind` it to window so that its execution scope is always as expected

src/cdk/layout/media-matcher.ts
josephperrott: This is actually a hold over from @ThomasBurleson 's original implementation.  My understanding is the `try-catch` is intended to allow the service to fail gracefully and not cause an entire halt.

jelbourn: My stance is that any `try-catch` need a comment explaining which error it is catching and why it's something that has to be caught instead of prevented in the first place

@ThomasBurleson do you remember if there was a specific reason for using `try` around creating a style element?

src/cdk/layout/media-matcher.ts
jelbourn: Add a comment mentioning this?

src/cdk/layout/media-matcher.spec.ts
jelbourn: It's fine for the test to have an `if` block checking the platform before doing anything

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+  /** Completes the active subject, signalling to all other observables to complete. */
+  ngOnDestroy() {
+    this._activeSubject.next();
+    this._activeSubject.complete();
+  }
+
+  /** Whether the query currently is matched. */
+  isMatched(value: string | string[]): boolean {
+    let queries = coerceArray(value);
+    return queries.some((mediaQuery: string) => {
```
jelbourn: The `: string` shouldn't be necessary since it is inferred from the array. It's also acceptable to use a single-character variable for concise lambdas like this:
```ts
return queries.some(q => this._registerQuery(q).mql.matches);
```

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Retrieves an observable for the BreakpointQuery, broadcasting changes to the matching state.
+   */
+  observe(value: BreakpointQuery): Observable<BreakpointState> {
+    let queries = coerceArray(value);
+    let observables = queries.map(query => this._registerQuery(query).observable);
+
+    return combineLatest(observables, (a: BreakpointState, b: BreakpointState) => {
+      return <BreakpointState>{
```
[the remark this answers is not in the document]
jelbourn: Still has `<BreakpointState>`

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+    let mql: MediaQueryList = this.mediaMatcher.matchMedia(query);
+    // Create callback for match changes and add it is as a listener.
+    let queryObservable = RxChain.from(fromEventPattern(
+      /**
+       * Listener callback methods are wrapped to be placed back in ngZone. Callbacks must be placed
+       * back into the zone because matchMedia is only included in Zone.js by loading the
+       * webapis-media-query.js file alongside the zone.js file.  Additionally, some browsers do not
+       * have MediaQueryList inherit from EventTarget, which causes inconsistencies in how Zone.js
+       * patches it.
+       */
```
jelbourn: Use `//`-style comments for explanations / details (with `/** */` for class/member/variable descriptions)

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,100 @@
...
+  observable: Observable<BreakpointState>;
+  mql: MediaQueryList;
+}
+
+@Injectable()
+export class BreakpointObserver implements OnDestroy {
+  /**  A map of all media queries currently being listened for. */
+  private _queries: Map<string, Query> = new Map();
+  /** A subject for all other observables to takeUntil based on. */
+  private _activeSubject: Subject<{}> = new Subject();
```
jelbourn: Maybe `_teardownSubject` or `_destroySubject`?

src/cdk/layout/media-matcher.ts
```diff
@@ -0,0 +1,76 @@
...
+export class MediaMatcher {
+  /** The internal matchMedia method to return back a MediaQueryList like object. */
+  private _matchMedia: (query: string) => MediaQueryList;
+
+  constructor(private platform: Platform) {
+    this._matchMedia = this.platform.isBrowser ?
+      /**
+       * matchMedia is bound to the window scope intentionally as it is an illegal invocation to
+       * call it from a different scope.
+       */
```
jelbourn: Use `//`-style comments

src/demo-app/demo-app/demo-app.ts
```diff
@@ -67,6 +67,7 @@ export class DemoApp {
     {name: 'Progress Spinner', route: '/progress-spinner'},
     {name: 'Radio', route: '/radio'},
     {name: 'Ripple', route: '/ripple'},
+    {name: 'Screen Type', route: 'screen-type'},
```
rafaelss95: Missing `/` -> `/screen-type`.

src/cdk/layout/media-matcher.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * For Webkit engines that only trigger the MediaQueryListListener when there is at least one CSS
+ * selector for the respective media query.
+ */
+function createEmptyStyleRule(query: string) {
+  if (!styleElementForWebkitCompatibility.has(query)) {
+    try {
+      let style: HTMLStyleElement = document.createElement('style');
```
rafaelss95: `HTMLStyleElement` is already inferred from `.createElement(...)`.

josephperrott:
@jelbourn 

PR is ready for another look.

src/cdk/layout/breakpoints-observer.md
```diff
@@ -0,0 +1,26 @@
...
+A set of breakpoints is provided based on the Material Design 
+[breakpoint system](https://material.io/guidelines/layout/responsive-ui.html#responsive-ui-breakpoints).
+
+#### Example
+```ts
+@Component({ ... }) 
+export class MyWidget {
+  isHandset: Observable<BreakpointState>;
+  
+  constructor(bm: BreakpointManager) {
```
jelbourn: Still has `BreakpointManager` here

src/cdk/layout/breakpoints-observer.ts
```diff
@@ -0,0 +1,96 @@
...
+      // patches it.
+      (listener: MediaQueryListListener) => {
+        mql.addListener((e: MediaQueryList) => this.zone.run(() => listener(e)));
+      },
+      (listener: MediaQueryListListener) => {
+        mql.removeListener((e: MediaQueryList) => this.zone.run(() => listener(e)));
+      }))
+      .call(takeUntil, this._destroySubject)
+      .call(startWith, mql)
+      .call(map, (nextMql: MediaQueryList) => { return {query: query, matches: nextMql.matches}; })
```
jelbourn: Don't need the braces and the `return`, just parens around the object literal
```ts
.call(map, (nextMql: MediaQueryList) => ({query: query, matches: nextMql.matches}))
```
Also, is the explicit `MediaQueryList` type necessary here? I think it should be inferred.

src/cdk/layout/media-matcher.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+import {Injectable} from '@angular/core';
+import {Platform} from '@angular/cdk/platform';
+
+/**
+ * Global registry for all dynamically-created, injected style tags.
+ */
+const styleElementForWebkitCompatibility: Map<string, HTMLStyleElement> = new Map();
+
+@Injectable()
```
jelbourn: Add JsDoc description

src/cdk/layout/breakpoints-observer.ts
josephperrott: We need to declare the `MediaQueryList` type, because the `fromEventPattern` observable emits an `any` type unless you use the `selector` method to transform what is returned, but we would just have to define the type in this method for typescript to know its type.  So I just went with declaring the type in the `map` parameter.

jelbourn:
@josephperrott can you rebase? I want to get this in the next sync

josephperrott:
@jelbourn rebased.

jelbourn:
@josephperrott can you rebase one more time?

josephperrott:
@jelbourn rebased

PR #6775, opened 2017-08-31
feat(stepper): Address previous comments + add directionality support

g1shin:
Comments that were not addressed in the merging of first stepper PR into master branch are addressed in this PR.

### DIFF
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -120,8 +120,6 @@ export class CdkStep {
   selector: '[cdkStepper]',
 })
 export class CdkStepper {
-  private _stepsArray: CdkStep[];
-
   /** The list of step components that the stepper is holding. */
   @ContentChildren(CdkStep) _steps: QueryList<CdkStep>;
 
@@ -139,7 +137,7 @@ export class CdkStepper {
   get selectedIndex() { return this._selectedIndex; }
   set selectedIndex(index: number) {
     if (this._anyControlsInvalid(index)
-        || index < this._selectedIndex && !this._stepsArray[index].editable) {
+        || index < this._selectedIndex && !this._steps.toArray()[index].editable) {
       // remove focus from clicked step header if the step is not able to be selected
       this._stepHeader.toArray()[index].nativeElement.blur();
     } else if (this._selectedIndex != index) {
@@ -153,8 +151,7 @@ export class CdkStepper {
   @Input()
   get selected() { return this._steps[this.selectedIndex]; }
   set selected(step: CdkStep) {
-    let index = this._stepsArray.indexOf(step);
-    this.selectedIndex = index;
+    this.selectedIndex = this._steps.toArray().indexOf(step);
   }
 
   /** Event emitted when the selected step has changed. */
@@ -168,10 +165,6 @@ export class CdkStepper {
 
   constructor(@Optional() private _dir: Directionality) {
     this._groupId = nextId++;
-  }
-
-  ngAfterContentInit() {
-    this._stepsArray = this._steps.toArray();
   }
 
   /** Selects and focuses the next step in list. */
@@ -207,7 +200,7 @@ export class CdkStepper {
 
   /** Returns the type of icon to be displayed. */
   _getIndicatorType(index: number): 'number' | 'edit' | 'done' {
-    const step = this._stepsArray[index];
+    const step = this._steps.toArray()[index];
     if (!step.completed || this._selectedIndex == index) {
       return 'number';
     } else {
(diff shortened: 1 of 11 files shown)


mmalerba:
added @crisbeto since many of the comments were his

src/cdk/stepper/stepper.ts
```diff
@@ -210,27 +215,26 @@ export class CdkStepper {
...
     });
     this._selectedIndex = newIndex;
   }
 
   _onKeydown(event: KeyboardEvent) {
     switch (event.keyCode) {
       case RIGHT_ARROW:
-        if (this._dir && this._dir.value === 'rtl') {
+        if (this._layoutDirection() === 'rtl') {
           this._focusStep((this._focusIndex + this._steps.length - 1) % this._steps.length);
```
mmalerba: since these previous and next expressions are repeated and kind of intimidating it might be nice for readability  to create `_focusNextStep` and `_focusPreviousStep` methods

src/cdk/stepper/stepper.ts
```diff
@@ -167,6 +169,10 @@ export class CdkStepper {
     this._groupId = nextId++;
   }
 
+  ngAfterContentInit() {
+    this._stepsArray = this._steps.toArray();
```
mmalerba: Can't do this since `_steps` may update, e.g. when an `ngIf` changes from `true` to `false`

src/lib/stepper/_stepper-theme.scss
```diff
@@ -13,11 +13,12 @@
       background-color: mat-color($background, hover);
     }
 
-    .mat-step-label-active {
+    .mat-step-label.mat-step-label-active {
```
mmalerba: nit: I like to put the more specific ones after

src/lib/stepper/stepper.spec.ts
```diff
@@ -364,30 +421,47 @@ function assertPreviousStepperButtonClick(stepperComponent: MdStepper,
 }
 
 /** Asserts that step position is correct for animation. */
-function assertCorrectStepPosition(stepperComponent: MdStepper, fixture: ComponentFixture<any>) {
+function assertCorrectStepPosition(stepperComponent: MdStepper,
```
mmalerba: Can we rename this to `assertCorrectStepAnimationDirection`? I'm not really sure what "position" refers to. Also we can determine the stepper automatically given the fixture and take in a optional direction (`'ltr'` or `'rtl'`) rather than a boolean so that we aren't always passing `dir === 'rtl'`

src/cdk/stepper/stepper.ts
```diff
@@ -150,7 +152,7 @@ export class CdkStepper {
   @Input()
   get selected() { return this._steps[this.selectedIndex]; }
   set selected(step: CdkStep) {
-    let index = this._steps.toArray().indexOf(step);
+    let index = this._stepsArray.indexOf(step);
```
crisbeto: You can inline this assignment: `this.selectedIndex = this._stepsArray.indexOf(step)`.

src/cdk/stepper/stepper.ts
```diff
@@ -253,11 +257,14 @@ export class CdkStepper {
   }
 
   private _anyControlsInvalid(index: number): boolean {
-    const stepsArray = this._steps.toArray();
-    stepsArray[this._selectedIndex].interacted = true;
+    this._stepsArray[this._selectedIndex].interacted = true;
     if (this._linear) {
-      return stepsArray.slice(0, index).some(step => step.stepControl.invalid);
+      return this._stepsArray.slice(0, index).some(step => step.stepControl.invalid);
```
crisbeto: The `index` here may end up being a negative number (it corresponds to the `selectedIndex` which is a part of the public API) which will result in the `some` skipping the last N items of the array (e.g. `[1, 2, 3, 4].slice(0, -1)` is `[1, 2, 3]`).

g1shin:
Comments have been addressed. Ready for review again :+1:

src/lib/stepper/stepper.spec.ts
mmalerba: What about automatically getting the stepper from the fixture (in this method and others)?

g1shin: @mmalerba stepper is obtained from fixture by css query (e.g `fixture.debugElement.query(By.css('md-horizontal-stepper')).componentInstance`). Since these test functions are used commonly for both horizontal and vertical stepper, the stepper would need to be obtained first based on different test cases and then passed to the function for testing.

It could be possible to pass in a variable that checks to see if it's supposed to be horizontal or vertical stepper component then get the stepper from the fixture accordingly, but I'm not sure if that would be cleaner than how it's currently done.

g1shin:
Modified test functions in stepper.spec.ts. Ready for review now

PR #6777, opened 2017-08-31
feat(grid-list): Make responsive grid list

tinayuangao:
- Allow users to set `responsiveCols`, `responsiveRowHeight`, `responsiveGutterSize` to grid list
- Allow users to set `responsiveColspan`, `responsiveRowspan` to grid tiles
- The value is a map like  `{ 'xs': '3', 'gt-xs': '5'}`
The breakpoints are the same as AngularJs Material

### DIFF
--- a/src/demo-app/grid-list/grid-list-demo.html
+++ b/src/demo-app/grid-list/grid-list-demo.html
@@ -1,91 +1,107 @@
 <div class="demo-grid-list">
-  <md-card>
-    <md-card-title>Basic grid list</md-card-title>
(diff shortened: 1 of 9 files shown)


odahcam:
The PR #6772 was merged, so will we have any updates here?

jelbourn:
@tinayuangao closing this since that we'd want to use the `BreakpointObserver` for this now. Let's chat about API for this before revisiting (not a major priority)

odahcam:
Maybe @angular/flex-layout could be used to do this job instead. 🤔

PR #6782, opened 2017-09-01
fix(tooltip): memory leak in _setTooltipMessage

Eddman:
The subscription to this._ngZone.onMicrotaskEmpty is causing memory leaks:
image

### DIFF
--- a/src/lib/tooltip/tooltip.ts
+++ b/src/lib/tooltip/tooltip.ts
@@ -271,7 +271,7 @@ export class MatTooltip implements OnDestroy {
     this._detach();
     this._portal = this._portal || new ComponentPortal(TooltipComponent, this._viewContainerRef);
     this._tooltipInstance = overlayRef.attach(this._portal).instance;
-    this._tooltipInstance.afterHidden().subscribe(() => this._detach());
+    this._tooltipInstance.afterHidden().pipe(takeUntil(this._destroyed)).subscribe(() => this._detach());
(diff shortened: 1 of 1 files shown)


jelbourn:
@andrewseguin should we revisit this?

andrewseguin:
Yup - this still is valid I believe, since the subscription is not always unsubscribed

We now have four subscriptions in tooltip. I think we can benefit from an `onDestroyed` stream that we should `takeUntil`

Eddman:
@andrewseguin @jelbourn In my case I have a requirement to display table with tooltips in cells. With this amount of tooltips multiplied by size of data in memory we generate a huge leak. So yes this is still valid :)

Anyway I agree this was my first quick draft. Do you want me to implement it differently? Or at least rebase this branch?

jelbourn:
Updating this to use ` destroyed` subject along with `takeUntil` would be good (vs. keeping track of the subscription).

andrewseguin:
An example of what Jeremy was pointing out can be found here: https://github.com/angular/material2/blob/40a507043a47861d1d0f66dc5eb7591d762b089b/src/lib/sidenav/drawer.ts#L495

Eddman:
@jelbourn @andrewseguin I've implemented the fix using `takeUntil`. Tested in our app. It works without issues nor leaks ;)

mmalerba:
@Eddman please rebase

Eddman:
@mmalerba Done.

Issue #4252, opened 2017-04-24
Support snackbar pushing fab out of the way


jelbourn:
Eventually we should support snackbars pushing FABs out of the way if the notification attempts to show in the same position as the button.

chuihinwai:
Is there a (however ugly) workaround for this? Thanks!

geoffreyfourmis:
According to the Material Design 3 specifications, the snackbar should appear above FABs.
However, it seems that we can only control its position at the top or bottom.

![Image](https://github.com/user-attachments/assets/58dc8724-d1b6-4334-ac68-39d39ec71617)

crisbeto:
We are closing this issue, because we've determined, using a semi-automated process, that it is no longer relevant, has low engagement, or it does not have enough information to proceed. Feel free to comment if you think the issue should be reopened, or open a new issue with the latest version of our code instead.

PR #6791, opened 2017-09-01
fix(snackbar): swap enter and exit animation curves

willshowell:
- Uses the standard curve for both enter and exit animations
  - Previous entrance: http://cubic-bezier.com/#.4,0,1,1 (finishes animation too harshly)
  - Previous exit: http://cubic-bezier.com/#0,0,0.2,1 (starts animation too harshly)
  - Currently both: http://cubic-bezier.com/#.4,0,.2,1

- Use long animation duration [as spec shows](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0B6Okdz75tqQsZV9TS3lTVHFzRUE/components_snackbar_usage_fabdo_005.webm)

### DIFF
--- a/src/lib/snack-bar/snack-bar-container.ts
+++ b/src/lib/snack-bar/snack-bar-container.ts
@@ -33,15 +33,16 @@ import {
   PortalHostDirective,
 } from '@angular/cdk/portal';
 import {first} from '@angular/cdk/rxjs';
+import {AnimationCurves, AnimationDurations} from '@angular/material/core';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 import {MatSnackBarConfig} from './snack-bar-config';
 
 
-// TODO(jelbourn): we can't use constants from animation.ts here because you can't use
-// a text interpolation in anything that is analyzed statically with ngc (for AoT compile).
-export const SHOW_ANIMATION = '225ms cubic-bezier(0.4,0.0,1,1)';
-export const HIDE_ANIMATION = '195ms cubic-bezier(0.0,0.0,0.2,1)';
+export const SHOW_ANIMATION =
+    `${AnimationDurations.ENTERING} ${AnimationCurves.DECELERATION_CURVE}`;
+export const HIDE_ANIMATION =
+    `${AnimationDurations.EXITING} ${AnimationCurves.ACCELERATION_CURVE}`;
 
 /**
  * Internal component that wraps user-provided snack bar content.


src/lib/snack-bar/snack-bar-container.ts
```diff
@@ -42,8 +42,8 @@ export type SnackBarState = 'initial' | 'visible' | 'complete' | 'void';
 
 // TODO(jelbourn): we can't use constants from animation.ts here because you can't use
 // a text interpolation in anything that is analyzed statically with ngc (for AoT compile).
-export const SHOW_ANIMATION = '225ms cubic-bezier(0.4,0.0,1,1)';
-export const HIDE_ANIMATION = '195ms cubic-bezier(0.0,0.0,0.2,1)';
+export const SHOW_ANIMATION = '375ms cubic-bezier(0.4,0.0,0.2,1)';
```
josephperrott: As you noted, we are not using the correct easing curves for our animations.  However it appears we just have the curves reversed.

From looking at the spec once again:
[Entering the screen](https://material.io/guidelines/motion/movement.html#movement-movement-in-out-of-screen-bounds) should use the [deceleration curve](https://material.io/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves) - `cubic-bezier(0.0, 0.0, 0.2, 1)`
[Permanently leaving the screen](https://material.io/guidelines/motion/movement.html#movement-movement-in-out-of-screen-bounds) should use the [acceleration curve](https://material.io/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves)  - `cubic-bezier(0.4, 0.0, 1, 1)` 


The durations appear to be correct as they are non-complex transitions.  These are also guided by the spec found [here](https://material.io/guidelines/motion/duration-easing.html#duration-easing-common-durations).

willshowell:
@josephperrott 

### Curve

I was basing it on the idea that the snackbar is not permanently leaving the screen; rather it is just resting at the screen's edge until next use. This is obviously open for interpretation, but here's why I think the standard curve is appropriate:

1. Deceleration curve doesn't make sense to me because the snackbar (in all the other specs) has no reason to indicate it has been traveling at peak velocity:

>**Entering the screen**
> Elements entering the screen use the deceleration curve for a speedy entrance, indicating that they had been travelling at peak velocity.

2. Acceleration curve also doesn't make sense because it isn't permanently leaving:

> **Permanently leaving the screen**
> Elements **permanently** leaving the screen use the acceleration curve to speed off-screen over a slightly shorter duration, as they will not be returning and require less user focus.

3. The standard curve should be used when moving other on screen elements, as snackbar [eventually will do](https://github.com/angular/material2/issues/4252):

> **Relative movement**
> Entering or exiting elements that move other on-screen elements do so along a smooth easing curve, so that they remain minimally disruptive and avoid eye-catching, dramatic movement.
> 
> The standard curve is used for moving elements both in and out of the bounds of the screen. This curve has a slightly longer duration compared to independent elements.

4. Finally, [this defining animation of the standard curve](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0B14F_FSUCc01U3RfQ3dTN3RROE0/InOut_05_ShiftInOutRelativeDo_v3-device.webm) shows a snackbar.

### Timing

I tried it with 275ms and 195ms and neither looked like the spec's animations. If you look at [this animation](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0B6Okdz75tqQsZV9TS3lTVHFzRUE/components_snackbar_usage_fabdo_005.webm), the snackbar is moving for 12-14 frames and the video is 30fps. I agree that it isn't a complex animation, but 375ms seemed to best match what I saw [here](https://material.io/guidelines/components/snackbars-toasts.html).

willshowell:
Hey @josephperrott, as you can see in my last comment, I feel pretty strongly about this 😁. Would it be possible to send a message to the UX team specifically about snackbar curves and durations?

josephperrott:
@willshowell sorry for the delay, I actually did as the UX team to get an answer on this one.  Ultimately they instructed to use the curves as I outlined in my previous comment:
> [Entering the screen](https://material.io/guidelines/motion/movement.html#movement-movement-in-out-of-screen-bounds) should use the [deceleration curve](https://material.io/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves) - `cubic-bezier(0.0, 0.0, 0.2, 1)`
[Permanently leaving the screen](https://material.io/guidelines/motion/movement.html#movement-movement-in-out-of-screen-bounds) should use the [acceleration curve](https://material.io/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves)  - `cubic-bezier(0.4, 0.0, 1, 1)` 


**With regards to curves:**
I agree with most of your points you outlined in the last update, however I am just following our UX team on this one.  

Though I will point out that for <span>#</span>4: 
> 4.  The snackbar is not leaving the screen permanently as use of the acceleration curve would suggest.

Any snackbar that comes back into the screen is actually a newly composed snackbar, even if the content is the same content.

**With regards to the timings:**
I wouldn't fully trust the frame rate in the animation example videos, though we should be able to start setting the rate of animation based on screen size once #6772 lands.

willshowell:
@josephperrott Thanks for reaching out to them! I'll update it based on your original comment.

willshowell:
@josephperrott change made

willshowell:
@josephperrott I've updated this to use the animation constants. Can you take another look?

PR #6793, opened 2017-09-01
fix(checkbox): support native tabindex attribute

DevVersion:
Currently the checkbox only allows changing the tabIndex using the tabIndex binding. Using the native tabindex attribute doesn't have any affect. With this change the native tabindex property will be respected.

References #6465


DevVersion:
@tinayuangao PTAL

PR #6800, opened 2017-09-01
build: add tslint rule for validating decorators

### LINKED ISSUE #6778: Add lint rule to enforce that every @Component has a moduleId
We need each `@Component` to have a `moduleId` for Google; recently the stepper slipped through without having that.

crisbeto:
* Reworks the `noViewEncapsulationRule` to turn it into a more generic rule that allows us to validate multiple properties and decorators.
* Fixes a few components that were caught without a `moduleId` after implementing the new rule.
* Adds validation that all components have a `moduleId`.

Fixes #6778.

### DIFF
--- a/src/cdk/table/row.ts
+++ b/src/cdk/table/row.ts
@@ -141,6 +141,7 @@ export class CdkCellOutlet {
 
 /** Header template container that contains the cell outlet. Adds the right class and role. */
 @Component({
+  moduleId: module.id,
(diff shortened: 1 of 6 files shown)


jelbourn:
Actually, some of the cdk packages need to be updated w/ the `typings.d.ts` for `module`

crisbeto:
Sorted out the build issues @jelbourn, PTAL.

PR #6801, opened 2017-09-01
compat: selection list not working in compatibility mode

### LINKED ISSUE #6746: mdSelectionList doesn't work in v1 compatibility mode
Bug

mdSelectionList works in v1 compatibility mode

```
ERROR Error: The "md-" prefix cannot be used in ng-material v1 compatibility mode. It was used on an "md-pseudo-checkbox" element.
    at getMdCompatibilityInvalidPrefixError (compatibility.js:18) [angular]
    at new MdPrefixRejector (compatibility.js:45) [angular]
```

http://plnkr.co/edit/EDHl8f?p=preview

crisbeto:
Fixes the selection list not working in M1 compatibility mode.

Fixes #6746.

### DIFF
--- a/src/lib/list/list-option.html
+++ b/src/lib/list/list-option.html
@@ -1,10 +1,16 @@
-<div class="mat-list-item-content"  [class.mat-list-item-content-reverse]="checkboxPosition == 'after'" [class.mat-list-item-disabled]="disabled">
-  <div class="mat-list-item-ripple" md-ripple
-       [mdRippleTrigger]="_getHostElement()"
-       [mdRippleDisabled]="_isRippleDisabled()">
-  </div>
+<div class="mat-list-item-content"
+  [class.mat-list-item-content-reverse]="checkboxPosition == 'after'"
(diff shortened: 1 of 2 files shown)


src/lib/list/list-option.html
```diff
@@ -1,10 +1,17 @@
...
+  [class.mat-list-item-disabled]="disabled"
+  [ngSwitch]="_isCompatibilityMode">
+
   <div class="mat-list-item-ripple" md-ripple
        [mdRippleTrigger]="_getHostElement()"
-       [mdRippleDisabled]="_isRippleDisabled()">
-  </div>
+       [mdRippleDisabled]="_isRippleDisabled()"></div>
+
+  <mat-pseudo-checkbox #autocheckbox *ngSwitchCase="true"
```
jelbourn: Andrew and I realized a simpler way of doing this: the component itself can just provide `MATERIAL_COMPATIBILITY_MODE` to false
(which is fine because this component does have any real conflicts)

crisbeto:

PR #6802, opened 2017-09-01
fix(sort): reverse directions and better animation

### LINKED ISSUE #6198: MdTable sort indicators backwards
Bug

According to the material spec site (https://material.io/guidelines/components/data-tables.html#data-tables-interaction), an up arrow indicates smallest value at top (ascending sort).

All examples in documentation (https://material.angular.io/components/sort/overview) has up arrow indicating largest value at top (descending sort)

Material beta 8

andrewseguin:
Fixes #6198


src/lib/sort/sort-header.ts
```diff
@@ -52,8 +52,8 @@ import {getMdSortHeaderNotContainedWithinMdSortError} from './sort-errors';
   changeDetection: ChangeDetectionStrategy.OnPush,
   animations: [
     trigger('indicatorRotate', [
-      state('asc', style({transform: 'rotate(45deg)'})),
-      state('desc', style({transform: 'rotate(225deg)'})),
+      state('asc', style({transform: 'rotate(225deg)', top: '2px'})),
```
jelbourn: What's with the `2px`?

mcalmus:
I haven't been able to reproduce it with a simple test case (and can't upload my regular site), but with this patch in place (beta 11), I'm seeing the sort header "stem" and indicator appearing horizontally next to one another rather than vertically.

I was able to fix this by (with the debugger) adding a "flex-direction: column" style to the "mat-sort-header-arrow" class.

Would it be possible to do so in the official release?

jelbourn:
@mcalmus Please open an issue for bug reports.

PR #6803, opened 2017-09-01
fix: Remove MaterialModule

josephperrott:
BREAKING CHANGE: MaterialModule will no longer be available.

`MaterialModule`has been deprecated since beta.3 and is now being removed.

From [beta.3 changelog](https://github.com/angular/material2/blob/master/CHANGELOG.md#200-beta3-cesium-cephalopod-2017-04-07):
#### MaterialModule
* `MaterialModule` (and `MaterialRootModule`) have been marked as deprecated.

We've found that, with the current state of tree-shaking in the world,
that using an aggregate NgModule like `MaterialModule` leads to tools
not being able to eliminate code for components that aren't used.

In order to ensure that users end up with the smallest code size
possible, we're deprecating MaterialModule, to be removed in the a
subsequent release.

To replace `MaterialModule`, users can create their own "Material"
module within their application (e.g., `GmailMaterialModule`) that
imports only the set of components actually used in the application.


George35mk:
and what i need to do now ?? any Help ?? i have error after upgrating to new version  `"@angular/material": "^2.0.0-beta.11",`

WindMay:
@George35mk 
> and what i need to do now ?? any Help ?? i have error after upgrating to new version "@angular/material": "^2.0.0-beta.11"

Build a like the one that was deleted, but remove the modules that you don't use

https://github.com/angular/material2/commit/cf1ece0

Issue #5238, opened 2017-06-20
[Tabs] Input focus breaks sliding animation


Mario-Eis:
#### Bug, feature request, or proposal:

Bug (or feature request)

#### What is the expected behavior?

Tab animation should run flawless.
Or an alternative way should be provided to handle the focus when browsing tabs

#### What is the current behavior?

If the focus is set to an element in the new visible tab, the animation breaks

#### What are the steps to reproduce?

http://plnkr.co/edit/JOaZUON0G7JTBcEWlZTJ?p=preview

#### What is the use-case or motivation for changing an existing behavior?

Handle input focus when browsing tabs

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?

4.2.3 / beta.6, at least chrome and opera

#### Is there anything else we should know?

Slide from the left side seems to work, slide from the right side is broken when focus is set.
Setting the focus within a 500ms timeout works, but has other problems like when users switches the tabs to fast (second tab-switch within the 500ms timeout). Also the 500ms would be hard coded in the component host and could not be compatible to future tab animation implementations.
Providing a hook to @translateTab.done could be a possible quick-fix.

amcdnl:
@jelbourn - Are you ok w/ exposing the `onCentered` event from the body?

jelbourn:
@amcdnl don't quite follow what you mean

amcdnl:
@jelbourn - `onCentered` is an event emitted from the tab-body component when the animation has finished. Its hidden and not re-emitted for public consumption, my suggestion to address his issue was to expose this event for consumption by external consumers.

jelbourn:
I wouldn't be a fan of that API. What's the underlying problem?

amcdnl:
The user wants the ability to set focus once the animation has completed.

Mario-Eis:
Doesn't necessarily need to be set once the animation is complete. Thats just one way to do it. The issue is, thats it's not possible to set the focus at all. Except for some dirty workarounds described above. I could imagine the underlaying issue could be the inputs focus animation breaking the tabs sliding animation... just a guess.

jelbourn:
Something like `(animationDone)` would be more general. @matsko, do you have any input in general for use-cases like this?

amcdnl:
@jelbourn - let me know if you want me to move forward with that API.

jelbourn:
[...]

PR #6811, opened 2017-09-02
feat(tabs): add animation done event #5238

### LINKED ISSUE #3032: [Tabs] Add output for onCentered
Feature request

I would like to have have a hook on top level (@Output property) on md-tab-group to know when the sliding in animation as defined in MdTabBody completed.
Ideally proxying @Output() onCentered from MdTabBody on MdTabGroup component to re emit.

We don't have an option to resize third party components once the slide in has completed ->
 this._portalHost.attach happens initially in ngAfterViewChecked for MdTabBody, but at that point the portalHost element does not have width and height defined, which the child component we are attaching can not size itself correctly.

So nested components inside tabs which are not active get correct width and height properties/ or allow them to be resized on animation complete -> e.g. onCentered.emit() for MdTabBody

@angular: "2.4.1"
@angular/material: "2.0.0-beta.1"

amcdnl:
Addresses #5238 by adding an `animationDone` event that is triggered after the tab body transition has completed.

Closes #3032


src/lib/tabs/tab-group.ts
```diff
@@ -269,5 +272,6 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
   _removeTabBodyWrapperHeight(): void {
     this._tabBodyWrapperHeight = this._tabBodyWrapper.nativeElement.clientHeight;
     this._renderer.setStyle(this._tabBodyWrapper.nativeElement, 'height', '');
+    this.animationDone.emit();
```
jelbourn: Unit test? Even with `NoopAnimationModule`, it should still fire the event

amcdnl: It looks like the `_onTranslateTabComplete` function is getting called in `tab-body.ts`, however, the `onCentered` event that is captured in `tab-group.ts` in the `_removeTabBodyWrapperHeight` method is not getting triggered in the tests. Any thoughts?

jelbourn: What kind of flushing / waiting does the test use?

amcdnl: Here is the test I've written for this:

```
fit('should fire animation done event', async (() => {
  fixture.detectChanges();

  spyOn(fixture.componentInstance, 'animationDone');
  let tabLabel = fixture.debugElement.queryAll(By.css('.mat-tab-label'))[1];
  tabLabel.nativeElement.click();
  fixture.detectChanges();

  fixture.whenStable().then(() => {
    fixture.detectChanges();
    expect(fixture.componentInstance.animationDone).toHaveBeenCalled();
  });
}));
```

amcdnl:
@jelbourn Test added :)

jelbourn:
```
ERROR: /home/travis/build/angular/material2/src/lib/tabs/tab-group.spec.ts[209, 5]: Calls to 'fit' are not allowed.
```

amcdnl:
@jelbourn - #facepalm - fixed.

tinayuangao:
@amcdnl Please rebase. Thanks!

amcdnl:
@tinayuangao - Done :)

PR #6817, opened 2017-09-03
build: custom tslint rules not executing in IDEs

crisbeto:
Fixes our custom tslint rules not executing IDEs, because they're written in TypeScript. This is the preferred option over switching them back to ES6, because some rules require access to TS files from other parts of the build (e.g. the Rollup globals rule).

### DIFF
--- a/tools/tslint-rules/noViewEncapsulationRule.ts
+++ b/tools/tslint-rules/noViewEncapsulationRule.ts
@@ -35,17 +35,17 @@ class Walker extends Lint.RuleWalker {
     this._enabled = fileGlobs.some(p => minimatch(relativeFilePath, p));
   }
 
-  visitClassDeclaration(node) {
+  visitClassDeclaration(node: ts.ClassDeclaration) {
     if (!this._enabled || !node.decorators) {
       return;
     }
 
     node.decorators
-      .map(decorator => decorator.expression)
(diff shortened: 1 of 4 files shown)


tools/tslint-rules/noViewEncapsulationRule.ts
```diff
@@ -35,23 +35,23 @@ class Walker extends Lint.RuleWalker {
     this._enabled = fileGlobs.some(p => minimatch(relativeFilePath, p));
   }
 
-  visitClassDeclaration(node) {
+  visitClassDeclaration(node: any) {
```
DevVersion: Don't you want to use types here?  e.g `ts.ClassDeclaration`? (also below)

crisbeto: I can't definitely use it for the `node`, but the `expression` below doesn't seem to be in the typings.

DevVersion: Not really something that is too important. For the `expression` you could use `ts.CallExpression`, but it will not have types for the `arguments[0].properties`.

crisbeto: Yeah eventually the `arguments` have to be cast to `any`.

Issue #6720, opened 2017-08-30
Why are some attribute selectors dash-cased?


PascalPrecht:
#### Bug, feature request, or proposal:
`MdButtonDirective` should have an `[mdButton]` select. As of version beta.10, all dash-cased directives are deprecated in favour of their camelCased equivalents.

- Release notes: https://github.com/angular/material2/blob/master/CHANGELOG.md#deprecations
- Directives affected: https://github.com/angular/material2/blob/master/src/lib/button/button.ts#L40

#### What is the expected behavior?
This should give us a high-quality material button:

```
<button [mdButton|mdRaisedButton|mdIconButton]>Look how beautiful I am!</button>
```

#### What is the current behavior?
The `mdButton` selectors aren't supported yet.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
`2.0.0-beta.10`

#### Is there anything else we should know?
Everyone working on this material project is doing an outstanding job.

PascalPrecht:
As I'm progressing with this, I ran into some more:

- `[tab-nav-bar]` - this is actually a component but can't be used as `<tab-nav-bar>` so I assume that should work as `tabNavBar` as well? https://github.com/angular/material2/blob/master/src/lib/tabs/tab-nav-bar/tab-nav-bar.ts#L58

- `[md-menu-item]` - Also no selector for `mdMenuItem` https://github.com/angular/material2/blob/master/src/lib/menu/menu-item.ts#L31

jelbourn:
This is actually working as intended. Button, menu items, and tab-nav all use `@Component`.

We spent a long time discussing this internally. The rationale is that, because only one `@Component` can live on an element, that component "owns" the element. When assigning properties for an element, those without prefix implicitly belong to the component that owns the element. For example:
```html
<button md-button disableRipple mdTooltip="Settings">...</button>
```
In this, we see that `md-button` is dashed, so that you know that `disableRipple` belongs to the `md-button` and not `mdTooltip`. 

If we did make the component selector camelCase, we'd also have to prefix all of its properties as well, which gets weird. Would you expect to write `mdButtonDisabled` instead of just `disabled`?

The dash-case here also aligns more with the [custom element spec](https://www.w3.org/TR/custom-elements/), which typically uses `dash-case` names in the `is` attribute, e.g.
```html
<button is="plastic-button">Click Me!</button>
```
[...]

PR #6818, opened 2017-09-03
fix(card): add camel-cased selectors to content projection

### LINKED ISSUE #6816: Missing mdCardHeader ng-content slot
Bug

Specifying an `md-card` with something like `<h3 mdCardTitle>Shiba Inu</h3>` should put it in the right slot in an `md-card-header`.

The `.mat-card-header-text` element is empty, it should be selecting `[mdCardTitle]`.

~Providing a Plunker (or similar) is the *best* way to get the team to see your issue.~
(Plunker templates aren't working right now)

material2 beta 10

crisbeto:
Fixes the camel-cased selectors for `mdCardTitle` and `mdCardSubtitle` not being added to the content projection selector.

Fixes #6816.

@jelbourn this seems along the same lines as https://github.com/angular/material2/pull/5647, should I get rid of the dash-cased selectors while I'm at it?

### DIFF
--- a/src/lib/card/card-header.html
+++ b/src/lib/card/card-header.html
@@ -1,6 +1,8 @@
 <ng-content select="[mat-card-avatar], [matCardAvatar]"></ng-content>
 <div class="mat-card-header-text">
-  <ng-content select="mat-card-title, mat-card-subtitle, [mat-card-title], [mat-card-subtitle]">
-  </ng-content>
+  <ng-content
+      select="mat-card-title, mat-card-subtitle,
+      [mat-card-title], [mat-card-subtitle],
+      [matCardTitle], [matCardSubtitle]"></ng-content>
 </div>
 <ng-content></ng-content>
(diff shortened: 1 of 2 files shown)


julianobrasil:
Shouldn't it also be changed in **`card-header.html`**?

rockmacaca:
Can I interject with one other related question? Shouldn't `md-card-header`, `md-card-content`, `md-card-actions`, and `md-card-footer` all support camel-cased attribute selectors as well? They're all technically directives and I've been using the equivalent dialog container components using the attribute selectors for some time.

julianobrasil:
@crisbeto, once there's no **`@Component`** decorator associated, I would expected to find camelCase attributes to all of them. At least it was what I got from this: according to https://github.com/angular/material2/issues/6720#issuecomment-326012361 (unless there is any specificity related to **`md-card`**).

jelbourn:
@andrewseguin where did we leave off when we last talked about these? This was one of the special cases we said we'd figure out later

andrewseguin:
I think we ought to sit down and look at the list of outstanding selectors we want to review to decide on how we move forward on it. Will bring it up with the team and see if we want to schedule something.

With regard to this issue, looks like this fix needs to go in regardless

PR #6822, opened 2017-09-03
build: parallel secondary entry-points building

DevVersion:
All secondary entry-points of the CDK are no longer built sequentially. Secondary entry-points will be resolved in batches that can be compiled in parallel.

For example:
```ts
[
  [ 'coercion', 'platform', 'rxjs', 'keycodes', 'bidi', 'collections', 'portal' ],
  [ 'a11y', 'observers', 'scrolling', 'stepper', 'table' ],
  [ 'overlay'
]
```

This should improve the build time of the CDK package.

**Note**: Shouldn't be really important (because of Bazel), but it shouldn't cause merge conflicts either and it was something I played with a long time ago.

### DIFF
--- a/tools/package-tools/build-package.ts
+++ b/tools/package-tools/build-package.ts
@@ -1,9 +1,8 @@
 import {join} from 'path';
-import {main as tsc} from '@angular/tsc-wrapped';
 import {buildConfig} from './build-config';
 import {getSecondaryEntryPointsForPackage} from './secondary-entry-points';
 import {buildPrimaryEntryPointBundles, buildSecondaryEntryPointBundles} from './build-bundles';
-
+import {main as ngc} from '@angular/tsc-wrapped';
 
 const {packagesDir, outputDir} = buildConfig;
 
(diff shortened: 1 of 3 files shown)


jelbourn:
Did this make any difference in build time for you? On my machine is ended up being about the same.

DevVersion:
@jelbourn True. There is not much of a difference here. I got a 4-5 seconds speed boost due to this change on Windows, but I assume that this is varying anyway.

It looks like calling the `NGC` using `@angular/tsc-wrapped` blocks the process, which then behaves similar as sequential building.

Feel free to close this PR. It doesn't really make a big difference.

jelbourn:
4 - 5 seconds is pretty huge- is that consistent? My build is consistently ~12 seconds with and without this PR

DevVersion:
Not really. As I said it's varying a lot. 

Sometimes it's equal to the old time, but sometimes it's also faster (e.g. 4sec). But to be honest I think that more relates to the whole setup (e.g if `dist/` is cleared already; if there is disk work in progress)

PR #6829, opened 2017-09-04
fix(selection-list): proper style for disabled options

DevVersion:
* No longer shows ripples if md-list-option is disabled
* Properly shows disabled style for disabled md-list-option (the `mat-list-item-disabled` class has been already added as part of the selection list PR)

### DIFF
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -232,6 +232,52 @@ describe('MdSelectionList', () => {
     });
   });
 
+  describe('with option disabled', () => {
+    let fixture: ComponentFixture<SelectionListWithDisabledOption>;
+    let listOptionEl: HTMLElement;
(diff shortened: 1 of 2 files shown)


src/lib/list/selection-list.spec.ts
```diff
@@ -232,6 +232,52 @@ describe('MdSelectionList', () => {
     });
   });
 
+  describe('with option disabled', () => {
+    let fixture: ComponentFixture<SelectionListWithDisabledOption>;
+    let listOptionEl: HTMLElement;
+    let listOption: MdListOption;
+
+    beforeEach(async(() => {
```
crisbeto: Why are there two separate `beforeEach`?

DevVersion: Because otherwise we'd need to chain into `TestBed.compileComponents()`. That's basically how we do it in every test file.

Issue #5222, opened 2017-06-19
Value of text area with mdTextareaAutosize is not visible inside MdTab


Zihark17:
#### Bug, feature request, or proposal:
The value of text area with mdTextareaAutosize is not visible until the size of the text area is recalculated (with resizeToFitContent() or by typing) when the text area is inside an MdTab that is not initially visible.
The min rows is also not working.

#### What is the expected behavior?
The text area value is visible and min rows works.

#### What is the current behavior?
The text area value is not visible until the value change by typing or resizeToFitContent is called and min rows doesn't works.

#### What are the steps to reproduce?
http://plnkr.co/edit/h4EYvCm36ncKlJolJnhU?p=preview
Open plunker and go to the second tab.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular: 4.2.3
Material: 2.0.0-beta.6
Windows 10
Typescript: 2.3.4

Zihark17:
Related to #3346.

amcdnl:
@jelbourn - This is likely related to https://github.com/angular/material2/issues/1854. 

The size of the textarea is calculated on init of the component and the component is not actually injected into the dom until the user clicks the tab.

Another approach for resolution would be to use the intersection observer to detect when the textarea is in view and then calculate the size. Reference: https://github.com/w3c/IntersectionObserver/blob/gh-pages/explainer.md ... this might be nice CDK directive to have for other situations too.

jelbourn:
`IntersectionObserver` is a _long_ way away from being usable on the platforms we support. 

This is mostly the same as #3346. I don't have any greats ideas for a good way for the `autosize` to know when it should "go" based on being hidden or not in the DOM. @crisbeto @andrewseguin any thoughts?

crisbeto:
In this case it might be best not to try and handle it automatically, or we risk destroying performance. Instead consumers can get a hold of the `MdTextareaAutosize` instance and call `resizeToFitContent` manually when appropriate. Note that we'll probably have to tweak `resizeToFitContent` since it won't do anything if the value hasn't changed.

jelbourn:
We could do something like setting the previous value to `''` if the height is zero

MatthewTrout:
I'd like to be able to get hold of the `MdTextareaAutosize` instance in order to call `resizeToFitContent` when appropriate, but I cannot seem to find out how to get a reference to it...

Zihark17:
[...]

Issue #1854, opened 2016-11-14
[Tabs] Tabs are loaded eagerly


RoxKilly:
#### Bug, feature request, or proposal:
Bug or possibly poor design

#### What is the expected behavior?
When a tab is never opened, its contents should never be added to the DOM. This is how things worked with `alpha 9` on `Angular 2.0.1`

#### What is the current behavior?
Even if a tab is never opened, its content is added to the DOM before being removed.

#### What are the steps to reproduce?

[This plunk shows the old behavior](http://plnkr.co/edit/GDp51J7AyYB631dFfM3l?p=preview) , from `alpha 9`. If you run the application while looking at network requests for images, you will see that only the image from the first tab is downloaded. To download another image, you need to open its tab.

[This plunk shows the new behavior](http://plnkr.co/edit/ewBbcevOv0wWgse1L1Wf?p=preview), from `alpha 10`. In this case, all images are downloaded immediately, then removed from the DOM.  

#### What is the use-case or motivation for changing an existing behavior?
In my use case, I have a menu on the first tab and a video on each of several other tabs. Since I upgraded to `alpha 10`, the browser fetches all the videos at once instead of waiting for the user to open a tab. This results in excessive bandwidth usage.

The same problem would occur for an app that has different image galleries organized by tab

#### Which versions of Angular, Material, OS, browsers are affected?
`Angular 2.1.2`, `Material 2 alpha 10` on `Firefox 49`

jelbourn:
@andrewseguin I think this might be a bug in Angular based on the way we're doing tabs.

RoxKilly:
A dirty workaround: I listen to `(selectChange)` and I wrap the tab contents in an `*ngIf` that will add the contents to the DOM only if the tab is currently selected.

[See this plunk](http://plnkr.co/edit/V7VPskmyS2ImB41Mgm5B?p=preview)

elvirdolic:
It shouldn't load the content until the user click the tab.

I think the best approach is to load the content when the click event is fired and the loaded content is not destroyed anymore if the user switch to another tab and switch back again

esbeto:
Might be related: after upgrading to alpha.10 

Tabs that have an *ngIf are breaking for me:

```
error_handler.js:47 EXCEPTION: Uncaught (in promise): TypeError: Cannot read property 'appendChild' of null
```

RoxKilly:
@esbeto my last demo (see my most recent post above) uses a working `*ngIf` within a tab

[...]

PR #6832, opened 2017-09-04
feat(tabs): adds ability for lazy loaded tabs

amcdnl:
This adds the ability for users to lazy load the contents of a tab. 

The API for this is adding a new `<ng-template mdTabContent>` in the content of your tab with the tab's contents inside this tag.

```html
<md-tab-group>
  <md-tab label="First">
    <ng-template mdTabContent>
      <counter></counter>
    </ng-template>
  </md-tab>
  <md-tab label="Second">
    <ng-template mdTabContent>
      <counter></counter>
    </ng-template>
  </md-tab>
</md-tab-group>
```

This addresses: https://github.com/angular/material2/issues/1854 and https://github.com/angular/material2/issues/5222
References: https://github.com/angular/angular/issues/18691


src/lib/tabs/tab-body.ts
```diff
@@ -146,7 +146,8 @@ export class MdTabBody implements OnInit, AfterViewChecked {
    */
   ngAfterViewChecked() {
     if (this._isCenterPosition(this._position) && !this._portalHost.hasAttached()) {
-      this._portalHost.attach(this._content);
+      // Nested templates via mdTabContent templates causes expression change error
+      Promise.resolve().then(() => this._portalHost.attach(this._content));
```
jelbourn: Can we attach it in `ngDoCheck` instead? It would be better to avoid the extra change detection pass.

src/lib/tabs/tab.html
```diff
@@ -1,4 +1,9 @@
 <!-- Create a template for the content of the <md-tab> so that we can grab a reference to this
     TemplateRef and use it in a Portal to render the tab content in the appropriate place in the
     tab-group. -->
-<ng-template><ng-content></ng-content></ng-template>
+<ng-template #bodyTemplate>
+  <ng-content></ng-content>
+  <ng-template *ngIf="templateBody && templateBody.template"
```
jelbourn: Why use `ngTemplateOutlet` here instead of just creating a `TemplatePortal` directly from the `TemplateRef` from `mdTabContent`?

amcdnl: You need the original outlet for the projection, once projected then your able to read the template out and attach. I could replace the original outlet once projected but I'd have to attach/detach/attach. Is that what your suggesting?

jelbourn: I don't really follow what you mean.

I was thinking something like
```ts
export class MdTab {
  @ContentChild(MdTabContent, {read: TemplateRef}) _explicitContent: TemplateRef<any>; 
  @ViewChild(TemplateRef) _implicitContent: TemplateRef<any>;
}
```

Then use `_explicitContent` if it is defined, otherwise use `_implicitContent` (with the same code path we currently have for `_content`).

src/lib/tabs/tab-body.ts
amcdnl: Note: After this change I had to update the unit test that checked tab body contents.

src/lib/tabs/tab.html
amcdnl: You are right, querying works even though its not projected. Updated per feedback.

src/lib/tabs/tab-group.spec.ts
```diff
@@ -376,6 +379,34 @@ describe('nested MdTabGroup with enabled animations', () => {
...
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [MdTabsModule, BrowserAnimationsModule],
+      declarations: [TemplateTabs]
+    });
+
+    TestBed.compileComponents();
+  }));
+
+  it('should lazy load the second tab', (done) => {
```
jelbourn: Why do you need `done` here instead of using `async`?

amcdnl:
@jelbourn ready for another review.

josephperrott:
@amcdnl please rebase when you have a chance

amcdnl:
@josephperrott - Merged :)

josephperrott:
Look like this actually is causing a failure in a specific test case.  Here is a minimal repro spec:

```ts
import {CommonModule} from '@angular/common';
import {async, ComponentFixture, TestBed, flushMicrotasks, fakeAsync} from '@angular/core/testing';
import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
import {MdTabsModule} from './public_api';


describe('MdTabs', () => {
  let fixture: ComponentFixture<TestTabs>;
  let element: HTMLElement;

  beforeEach(async(() => {
      imports: [CommonModule, MdTabsModule, BrowserAnimationsModule],
      declarations: [TestTabs]
    });
    TestBed.compileComponents();
  }));


  beforeEach(() => {
    fixture = TestBed.createComponent(TestTabs);
    element = fixture.nativeElement;
    fixture.detectChanges();
    fixture.detectChanges();
  });

  it('will pass after click.', () => {
    const firstTab = element.querySelectorAll('.mat-tab-label')[0] as HTMLElement;
    firstTab.click();
    fixture.detectChanges();

    expect(element.querySelectorAll('.mat-tab-body')[0].querySelectorAll('.items').length).toBe(1);
  });
  it('won\'t pass without click.', () => {
    expect(element.querySelectorAll('.mat-tab-body')[0].querySelectorAll('.items').length).toBe(1);
  });
});


@Component({
  template: `
    <md-tab-group>
      <md-tab label="tab 1">
        <p class="items"> content for tab 1</p>
      </md-tab>
      <md-tab label="tab 2">
        <p class="items"> content for tab 2</p>
      </md-tab>
    </md-tab-group>
  `
})
class TestTabs {}
```

I am fairly confident this is because on initial load, the DoCheck lifecycle hits before the view is initialized which causes it not attach the portalHost on the first round of change detection.

amcdnl:
Waiting on @josephperrott changes on #7266 to be merged before rebasing.

jelbourn:
Downgrading this to P2 since it can go in a minor release

josephperrott:
@amcdnl rebase whenever you have a chance :)

amcdnl:
@josephperrott I'm going to close this PR and open a new one given all the changes.

PR #6839, opened 2017-09-05
fix(form-field): placeholder not floating if autofilled

### LINKED ISSUE #6837: Password-field placeholder doesnt rise when Chrome autofills password-field after beta.10
Bug

The password placeholder should be placed above the input.

When chrome autofills username and password, the username placeholder rises above the username value. The password placeholder however, stays in the field, overlapping the password. See screenshot below:
image

If i click on the page so it gains focus, the placeholder instantly jumps to the right position:
image

I have simply created an form with an input of type email and an input of type password, along with a submit-button.

Angular 4.3.6
Windows 10
Typescript 2.3.4
Chrome seems to be the only affected browser, since it's the only browser that completely autofills the form, without the user having to do anything. In both Edge and Firefox, I have to start typing in the username-field for the browser to suggest saved username/password, so by that time the window has focus, and the bug doesn't occur.

This functionality was working in beta.8.

DevVersion:
If a form-field value is set through autofill, the floating placeholder should float and not overlay the form-field value. This doesn't work in Chrome because the `.mat-form-field-autofill-float` class doesn't exist and can't be set on the form-field control anyway.

image

Fixes #6837

### DIFF
--- a/src/lib/form-field/_form-field-theme.scss
+++ b/src/lib/form-field/_form-field-theme.scss
@@ -49,7 +49,7 @@
     }
   }
 
-  .mat-input-element:-webkit-autofill + .mat-form-field-placeholder,
+  .mat-form-field-autofill-control:-webkit-autofill + .mat-form-field-placeholder,
   .mat-focused .mat-form-field-placeholder.mat-form-field-float {
     .mat-form-field-required-marker {
       color: $required-placeholder-color;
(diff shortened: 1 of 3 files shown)


src/lib/form-field/_form-field-theme.scss
```diff
@@ -49,7 +49,7 @@
     }
   }
 
-  .mat-form-field-autofill-float:-webkit-autofill + .mat-form-field-placeholder,
```
mmalerba: Can you instead just add the `mat-form-field-autofill-float` class to the `MdInput`? This was intended as a class users could add to custom a `MdFormFieldControl` to get the same effect, but I forgot to actually add it, haha...

DevVersion: Wouldn't that involve injecting the parent `MdFormField` to see if the floating placeholder is enabled or not?

DevVersion: Addressed discussed feedback from Slack.

PR #6856, opened 2017-09-05
refactor(select): use aria-activedescendant to manage focus

### LINKED ISSUE #6690: Focus is lost when opening md-select options list
When an user opens the *md-select* options list, focus is lost from the component itself.

Focus should remain in a child of *md-select* options and events such as *focusin* or *focusout* should not be triggered since the user has not really leave the component at tall.

As soon as the options list opens, focus is lost form the component containing the *md-select*.

Check focus log when selecting options: http://plnkr.co/uySEJGxzMTkbO6DxlT1F

In order to be able to track focus events and being able to apply custom styles to a component when it is focused, it is necessary that those events allow to track when the user really leave the component and they don't trigger while the user is selecting an option inside a form (for example).

Tested with Angular 4.3.1, Material 2.0.0-beta.8 under Chrome.

crisbeto:
* Refactors the select to use `aria-activedescendant` to announce the highlighted item to screen readers. Previously we would do this through focus, however using focus prevents us from being able to do things like #3211.
* Fixes a hack that was used to get a hold of the panel element using `querySelector`. Now it properly uses a `ViewChild` query, however this meant some tests had to be updated.

Relates to #3211.
Fixes #6690.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -43,6 +43,12 @@ import {
   getMatSelectNonFunctionValueError
 } from './select-errors';
 
+/** Duration of the select opening animation. */
+const SELECT_OPEN_ANIMATION = 200;
+
+/** Duration of the select closing animation and the timeout interval for the backdrop. */
+const SELECT_CLOSE_ANIMATION = 500;
+
 
 describe('MatSelect', () => {
   let overlayContainerElement: HTMLElement;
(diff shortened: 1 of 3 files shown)


src/lib/select/select.spec.ts
```diff
@@ -145,48 +145,48 @@ describe('MdSelect', () => {
...
       fixture.detectChanges();
-
       trigger = fixture.debugElement.query(By.css('.mat-select-trigger')).nativeElement;
     });
 
-    it('should open the panel when trigger is clicked', () => {
+    it('should open the panel when trigger is clicked', fakeAsync(() => {
       trigger.click();
       fixture.detectChanges();
+      tick(200);
```
jelbourn: Make constants for `200` and `500` so that the different have some semantic meaning tied to it?

mmalerba:
please rebase

crisbeto:
Ready to go @mmalerba.

andrewseguin:
Needs rebase

Issue #6265, opened 2017-08-04
datepicker: Add support for passing IOS8601 strings to datepicker regardless of the DateAdapter


mahmood-sajjadi:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Initial value in text field is empty string and Date picket required an input text field which means should accept empty string as not defined value, or is user want to remove value, should be able to remove text field value

#### What is the current behavior?
in master code
in this PR : https://github.com/angular/material2/commit/8bb54ca4565a7d55d4f74ca0af4c8e7e02efb462
in this file:
https://github.com/angular/material2/blob/master/src/lib/datepicker/datepicker-input.ts#L119

an empty string is not an acceptable value and it is raising the following error:
`Datepicker: value not recognized as a date object by DateAdapter.`

#### What are the steps to reproduce?

Just add 
`formControlName="birthdate"`
to input field in datepicker and sround it by formGroup

this is plunker example:
http://plnkr.co/edit/bhswcOgSB36yqD6Yr8nF?p=preview

please open chrome console to see the warning an error

#### What is the use-case or motivation for changing an existing behavior?
Using datepicker in form group

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
angular 4.3.2
material 2.0.0-beta.8-5967f6e

#### Is there anything else we should know?
It was working fine before that PR becuse the condition was
`if (!value)`
which is working as expected with empty string

mmalerba:
We no longer support setting the value of the datepicker to something that is not a Date object, such as a string. If you want the datepicker to start off empty just use `null` instead of `''`: http://plnkr.co/edit/2mvkFihkfvEeX2Yhkmuw?p=preview

nicholasjativa:
@mmalerba what would be a solution, then, in an instance where we previously would get a Date string from somewhere (like Firebase) and we try to set the value of a datepicker using setValue(aDateString)? I guess its possible to do setValue(newDate(aDateString)), however, that wouldn't work when copying an entire object (from a backend) to a formGroup using setValue(anObj). Thanks.

mmalerba:
@nicholasjativa Hmm interesting I had not considered that case. The reason I removed the ability to use strings is because it was possible to put locale-sensitive strings in your code, e.g. `1/2/2017`. Currently you would have to transform the response object to change the fields to actual `Date` objects. I could maybe add back the ability to accept ISO 8601 strings only

nicholasjativa:
[...]

PR #6860, opened 2017-09-05
feat(datepicker): Add Moment.js adapter

mmalerba:


mmalerba:
There's still something in the build configuration that isn't right. We get an error `TypeError: moment_1.default is not a function` I believe this is a result of the synthetic default import...

mmalerba:
Based on the discussion here: https://github.com/Microsoft/TypeScript/issues/13375 it sounds like maybe the best thing to do is not import moment and instead just have people load it via a script tag (which also defeats the whole point of having this separate package that depends on Moment.js...)

src/material-moment-adapter/adapter/moment-date-adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+      shortMonths: moment.monthsShort(),
+      dates: range(31, (i) => this.createDate(2017, 0, i + 1).format('D')),
+      longDaysOfWeek: moment.weekdays(),
+      shortDaysOfWeek: moment.weekdaysShort(),
+      narrowDaysOfWeek: moment.weekdaysMin(),
+    };
+    moment.locale(globalLocale);
+  }
+
+  getYear(date: moment.Moment): number {
```
jelbourn: Make an alias at the beginning of the file for brevity?
```ts
type Moment = moment.Moment;
```

src/material-moment-adapter/adapter/moment-date-adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+    return this.clone(date).daysInMonth();
+  }
+
+  clone(date: moment.Moment): moment.Moment {
+    return date.clone().locale(this.locale);
+  }
+
+  createDate(year: number, month: number, date: number): moment.Moment {
+    // Check for invalid month and date (except upper bound on date which we have to check after
+    // creating the Date).
```
jelbourn: Mention what moment would otherwise do if the month is out of range?

src/material-moment-adapter/adapter/moment-date-adapter.spec.ts
```diff
@@ -0,0 +1,525 @@
...
+
+
+// Month constants for more readable tests.
+const JAN = 0, FEB = 1, MAR = 2, DEC = 11;
+
+
+// Add some locales for testing. These definitions come from Moment.js's fr.js and ja.js locale
+// files. (We don't want to the version of moment that comes with locales because it's a lot of
+// extra bytes to include in our tests.)
+moment.defineLocale('fr', {
```
jelbourn: Move this constant data into a separate file and import it?

src/material-moment-adapter/adapter/moment-date-adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+    super();
+    this.setLocale(dateLocale || moment.locale());
+  }
+
+  setLocale(locale: any) {
+    super.setLocale(locale);
+
+    // Temporarily change the global locale to get the data we need, then change it back.
+    let globalLocale = moment.locale();
+    moment.locale(locale);
```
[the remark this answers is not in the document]
mmalerba: Somehow missed the fact that you can do that

src/material-moment-adapter/adapter/moment-date-adapter.ts
mmalerba: Moment actually does the right thing, just want to be able to give better errors to the user

mmalerba:
Comments addressed, PTAL. Super huge thank you to @DevVersion for helping with a bunch of build issues

src/material-moment-adapter/adapter/moment-date-adapter.ts
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Inject, Injectable, Optional} from '@angular/core';
+import {DateAdapter, MAT_DATE_LOCALE} from '@angular/material';
+
+// Depending on whether rollup is used, moment needs to be imported differently.
+// TODO(mmalerba): See if we can clean this up at some point.
```
jelbourn: Can you expand this comment to explain more of what's happening?

tsconfig.json
```diff
@@ -2,6 +2,8 @@
 // IDEs and TSLint. For IDEs it ensures that `experimentalDecorator` warnings are not showing up.
 {
   "compilerOptions": {
+    // Needed for Moment.js since it doesn't have a default export.
+    "allowSyntheticDefaultImports": true,
```
DevVersion: I don't think we need that anymore right?

mmalerba: We need it if we want webstorm to not complain about `import {default as moment} from 'moment'`, not sure if we care about that or not...

christele15:
Good morning, 
I am a newbie to Angular, and I am sorry if that might be a newbie question but I was tasked to use the datepicker and trigger an event if the current date selected is less than 27 weeks, more than 27 weeks or 27 weeks, is this something that might be doable, and is there anything to guide me ?

another-guy:
While having a moment.js based date adapter is better than nothing, there are still little questions after reviewing the code.

Was [date-fns](https://date-fns.org/) considered as an immutable alternative? Using it would help reducing the potential defect count, and would get rid of the necessity to manually `this.clone(...)` the date in every method.

My related [post](https://github.com/angular/material2/issues/6995) and a [comment](https://github.com/angular/material2/issues/6265#issuecomment-328975522) on the related subject got zero attention, even though I provided a code snippet with [ISO8601 date adapter](https://gist.github.com/another-guy/adc2d9d731af7693fdbfab21b1e227a9#file-iso8601datestringadapter-ts) that is the first step to a pull request.

mmalerba:
@another-guy I'd certainly be open to adding a date-fns one as well, but last I checked it didn't support custom parsing formats which is the main thing lacking in the `NativeDateAdapter`, that's why I prioritized adding the `MomentDateAdapter`. We also expect users to create their own `DateAdapter`s if they want something a little different than what we offer by default.

another-guy:
@mmalerba Good to know. I was sure there _is_ a strong reason to use moment.js over the date-fns. Thanks for the answer!

andlcool:
Thanks for the moment date adapter. This will come in handy.

Are there examples on how to implement this? Thanks.

Issue #4562, opened 2017-05-15
[Tabs] Content sometimes not visible on Edge


wuchtig:
#### Bug, feature request, or proposal:

Bug

#### What is the expected behavior?

Tab-Animation should work.

#### What is the current behavior?

After upgrade to 2.0.0-beta.5, tab content sometimes unvisible.

#### What are the steps to reproduce?

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
Plunker template: https://goo.gl/DlHd6U

![15-05-_2017_15-19-40](https://cloud.githubusercontent.com/assets/1224640/26060143/69749608-3984-11e7-92a0-e8df307352b9.gif)

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?

Windows 10
Microsoft Edge

willshowell:
Plunker?

wuchtig:
step1: open [https://goo.gl/DlHd6U](url)

step2: paste this in `app.component.html`

`<md-tab-group>`
`<md-tab label="Tab 1">Content 1</md-tab>`
`<md-tab label="Tab 2">Content 2</md-tab>`
`</md-tab-group>`

willshowell:
http://plnkr.co/edit/cyhvhciWWHpezXnWHkWI?p=preview

Hmm, seems specific to Edge. I can't reproduce in Chrome 58.0 or FF 53.0, but Safari 10.1's animations are weird, so maybe that is related?

![safari tab bug](https://cloud.githubusercontent.com/assets/6475576/26064366/c2eff2c0-395e-11e7-9ec6-a66c382a2aec.gif)
(Safari 10.1)

wuchtig:
In Windows 10 + Chrome it works fine. It's specific to MS Edge

macjohnny:
could be related to https://github.com/angular/material2/issues/4982

macjohnny:
@willshowell concerning your [comment](https://github.com/angular/material2/issues/4562#issuecomment-301504744): this is probably related to a bug in the web-animations-js polyfill, see 
https://github.com/angular/angular/issues/17239#issuecomment-306536861
https://github.com/angular/angular/issues/17239#issuecomment-308269650
https://github.com/web-animations/web-animations-js/issues/144

macjohnny:
@wuchtig Here is a plunkr example with the beta.6 version of angular material: http://plnkr.co/edit/0QHzQ4?p=preview
Here is a plunkr example with the latest master version of angular material:
[...]

PR #6861, opened 2017-09-05
fix(tabs): tabs distorted on edge #4562

amcdnl:
On Microsoft Edge, when you switch tabs the content of the tab is distorted. If a user interacts with the content area, it then appears.

Reference: [Video](https://www.screencast.com/t/ElCqSxjpiB)

Addresses:  #4562

### DIFF
--- a/src/lib/tabs/tab-body.scss
+++ b/src/lib/tabs/tab-body.scss
@@ -1,6 +1,7 @@
 .mat-tab-body-content {
   height: 100%;
-  overflow: auto;
+  // In MS Edge, the content is distorted after the 
+  // transition has ocurred. Ref: #4562
   overflow: unset;
 
   .mat-tab-group-dynamic-height & {
--- a/src/lib/tabs/tab-body.scss
+++ b/src/lib/tabs/tab-body.scss
@@ -1,6 +1,7 @@
 .mat-tab-body-content {
   height: 100%;
   overflow: auto;
+  overflow: unset;
 
   .mat-tab-group-dynamic-height & {
     overflow: hidden;


src/lib/tabs/tab-body.scss
```diff
@@ -1,6 +1,7 @@
 .mat-tab-body-content {
   height: 100%;
   overflow: auto;
+  overflow: unset;
```
jelbourn: * Won't this undo the `overflow: auto` directly above this?
* This would definitely need a detailed explanation of what/why.

amcdnl: 1) Yes, LOL. I forgot to remove that. It doesn't seem that rule is really applied given the parent container is the element actually getting the overflow properties. @andrewseguin - any ideas what that was for?
2) Added.

amcdnl:
@andrewseguin - Any thoughts on this?

src/lib/tabs/tab-body.scss
andrewseguin: @amcdnl Hard to say to be honest, we played a lot with the CSS on these tabs to get them to scroll right. Does each browser look alright without it? Check both dynamicHeight and non-dynamicHeight elements

amcdnl:
@andrewseguin bump

andrewseguin:
Checking out the demo, everything looks in order. If it fixes the issue in Edge, I think we can go ahead and ship this. Is there a reason that `unset` was preferred to use rather than just not specifying `overflow` at all?

josephperrott:
@amcdnl can you rebase this so we can work on getting it in?

jelbourn:
Closing this since we can't land it in Google.

PR #6884, opened 2017-09-06
build: infinite loop when determining dependencies on windows

crisbeto:
Fixes the `findstr` command going into an infinite loop on Windows due to it not being scoped to TS files.


DevVersion:
I wasn't able to reproduce it on my machine, but it's still good to restrict it to `.ts` files.

PR #6888, opened 2017-09-06
docs(chips): remove example from component jsdoc

### LINKED ISSUE #6885: HTML error in Chips documentation
Page: https://material.angular.io/components/chips/api

The `<md-chip>` tags are not being closed:

```
<md-chip-list>
  <md-chip>Chip 1<md-chip>
  <md-chip>Chip 2<md-chip>
</md-chip-list>
```

andrewseguin:
Closes #6885

### DIFF
--- a/src/lib/chips/chip-list.ts
+++ b/src/lib/chips/chip-list.ts
@@ -33,8 +33,8 @@ import {coerceBooleanProperty} from '@angular/cdk/coercion';
  * Example:
  *
  *     <md-chip-list>
- *       <md-chip>Chip 1<md-chip>
- *       <md-chip>Chip 2<md-chip>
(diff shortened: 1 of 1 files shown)


willshowell:
Does it even make sense to have an example in the class comment? A nearly identical example is near the top of the overview.

https://material.angular.io/components/chips/overview

andrewseguin:
It is rather redundant, I'm okay with removing it. What do you think @jelbourn - I don't think we include basic examples in other components

jelbourn:
Yeah, we shouldn't have examples in the descriptions. Live examples are always better because we actually build them.

andrewseguin:
SGTM - removed the example altogether

PR #6890, opened 2017-09-06
perf(dialog): avoid repainting dialog content element on scroll

### LINKED ISSUE #6878: Performance issue: md-dialog-content on desktop
Proposal

md-dialog-content should be optimized and should not repaint on scroll

md-dialog-content is not optimized and repaint on scroll

image

Angular Material has several performance issues which I think should be addressed.
The web is already slow, no need to make it any slower.
Consider using `will-change: transform;`

@angular/material: ^2.0.0-beta.10
Chrome 61

crisbeto:
Fixes the `<md-dialog-content>` element being repainted on scroll.

Fixes #6878.

### DIFF
--- a/src/lib/dialog/dialog.scss
+++ b/src/lib/dialog/dialog.scss
@@ -1,4 +1,5 @@
 @import '../core/style/elevation';
+@import '../core/style/vendor-prefixes';
 @import '../core/a11y/a11y';
 
 
@@ -35,6 +36,9 @@ $mat-dialog-button-margin: 8px !default;
   max-height: $mat-dialog-max-height;
   overflow: auto;
   -webkit-overflow-scrolling: touch;
+
+  // Promote the content to a new GPU layer to avoid repaints on scroll.
+  @include backface-visibility(hidden);
 }
 
 .mat-dialog-title {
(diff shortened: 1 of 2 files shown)


src/lib/dialog/dialog.scss
```diff
@@ -35,6 +35,9 @@ $mat-dialog-button-margin: 8px !default;
   max-height: $mat-dialog-max-height;
   overflow: auto;
   -webkit-overflow-scrolling: touch;
+
+  // Promote the content to a new GPU layer to avoid repaints on scroll.
+  backface-visibility: hidden;
```
jelbourn: Why was it repainting on scroll in the first place?

Lindsor: Better solution to moving to GPU is `transform: translate3d(0, 0, 0)`
Or the `will-change` property is getting more support

crisbeto: @Lindsor the result is exactly the same. I went with `backface-visibility`, because `transform` can cause the content to be blurry in some browsers.

@jelbourn I'm not sure, it's all up to the browser. My best guess is because it's in a fixed element and it has an explicit `overflow: auto`.

PR #6891, opened 2017-09-06
fix(sort): add aria-sort to host when sorted

andrewseguin:
Apply the `aria-sort` label to the sort header rather than include an additional visually hidden div to describe the sort. 

Up for debate: this property should only be applied to elements with role `columnheader` or `rowheader`. Should we only apply this property in those cases? If so, do we add in the `sortDescriptionLabel` to be displayed in the cases where the roles are not defined?

https://www.w3.org/TR/wai-aria/states_and_properties#aria-sort


src/lib/sort/sort-header.ts
```diff
@@ -46,6 +46,7 @@ import {getMdSortHeaderNotContainedWithinMdSortError} from './sort-errors';
   styleUrls: ['sort-header.css'],
   host: {
     '(click)': '_sort.sort(this)',
+    '[attr.aria-sort]': '_getAriaSortAttribute()',
```
jelbourn: What does the reader announce when we have both the `aria-sort` and the description?

Could you try this in NVDA and VoiceOver as well?

src/lib/sort/sort-header.ts
```diff
@@ -113,4 +114,16 @@ export class MdSortHeader implements MdSortable {
...
   }
+
+  /**
+   * Returns the aria-sort attribute that should be applied to this sort header. If this header
+   * is not sorted, returns null so that the attribute is removed from the host element. Aria spec
+   * says that the aria-sort property should only be present on one header at a time, so removing
+   * ensures this is true.
+   */
+  _getAriaSortAttribute() {
+    if (!this._isSorted()) { return null; }
```
jelbourn: Shouldn't it be `none` when it's not sorted?

src/lib/sort/sort-header.ts
```diff
@@ -113,4 +114,16 @@ export class MdSortHeader implements MdSortable {
   _isSorted() {
     return this._sort.active == this.id && this._sort.direction;
   }
+
+  /**
+   * Returns the aria-sort attribute that should be applied to this sort header. If this header
```
jelbourn: Prefer `Gets` to `Returns`

src/lib/sort/sort-header.ts
```diff
@@ -83,15 +84,16 @@ export class MdSortHeader implements MdSortable {
   get _id() { return this.id; }
   set _id(v: string) { this.id = v; }
 
-  constructor(public _intl: MdSortHeaderIntl,
+  constructor(public intl: MdSortHeaderIntl,
```
jelbourn: Do you mean for this to be part of the public api now?

andrewseguin: Trying to find the discussion about it, but can't find it. It seems people already are using it and would like to have it part of the public API. Wrong PR to put this in, but curious what you think about allow people to access it directly

src/lib/sort/sort-header.ts
andrewseguin: Here's JAWS with different options (with/without aria-sort/visually-hidden div)

With aria-sort, without visually-hidden: `change sorting for userid button sorted ascending`
Without aria-sort, with visually-hidden: `change sorting for userid button sorted by userid ascending`
With both: `change sorting for userid button sorted ascending sorted by userid ascending sorted ascending`

Checking out NVDA and VoiceOver...

src/lib/sort/sort-header.ts
andrewseguin: According to the spec, `For each table or grid, authors SHOULD apply aria-sort to only one header at a time.` so by returning null, the attribute is removed altogether rather than being set to `'none'` since it's possible another header now has `aria-sort` applied

src/lib/sort/sort-header.ts
andrewseguin: NVDA was simliar, wasn't able to get VoiceOver to play nice.

jelbourn:
Looks like some lint errors

andrewseguin:
Fixed lint issues

andrewseguin:
Marking as a major release due to the removal of some HTML elements

josephperrott:
@andrewseguin please rebase

andrewseguin:
Thanks, rebased

PR #6893, opened 2017-09-06
docs: sort class properties and methods

willshowell:
- Properties are sorted in the order of [Inputs, Outputs, neither]. Secondary sorting done alphabetically.
- Methods are sorted alphabetically.
- All deprecated members are sorted to the end.

On some components <sup>(ahem, [select](https://material.angular.io/components/select/api))</sup>, it's really hard to parse the API's list of properties and methods. This sorts them into a, hopefully, logical order.

### DIFF
--- a/tools/dgeni/processors/categorizer.js
+++ b/tools/dgeni/processors/categorizer.js
@@ -43,6 +43,10 @@ module.exports = function categorizer() {
 
     decoratePublicDoc(classDoc);
 
+    // Sort members
+    classDoc.methods.sort(sortMethods);
+    classDoc.properties.sort(sortProperties);
+
     // Categorize the current visited classDoc into its Angular type.
     if (isDirective(classDoc)) {
       classDoc.isDirective = true;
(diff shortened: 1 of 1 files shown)


tools/dgeni/processors/categorizer.js
```diff
@@ -95,6 +99,67 @@ function filterDuplicateMembers(item, _index, array) {
...
+    return 1;
+  }
+
+  return 0;
+}
+
+/** Sorts methods by deprecated status and name. */
+function sortMethods(docA, docB) {
+  const deprecatedSort = sortByDeprecated(docA, docB);
+  if (!!deprecatedSort) {
```
jelbourn: Don't need `!!` in a condition

tools/dgeni/processors/categorizer.js
```diff
@@ -95,6 +99,67 @@ function filterDuplicateMembers(item, _index, array) {
...
+    return deprecatedSort
+  }
+
+  // Break ties by sorting alphabetically by name
+  return sortByName(docA, docB);
+}
+
+/** Sorts properties by deprecated status, decorator, and name. */
+function sortProperties(docA, docB) {
```
jelbourn: Could you use the same function for properties and methods? All the checks for input/output should just return false for methods.

PR #6900, opened 2017-09-07
fix(snackbar): correct theme colors

domxa:
Currently default colors are used when dark theme is turned on and dark theme colors when it's turned off.
This change fixes the color assignment.


crisbeto:
These are the correct colors @domxa. The snack bar is supposed to be dark on the light themes and white on dark themes. See https://material-components-web.appspot.com/snackbar.html.

domxa:
My apologies, I thought it was an error from the latest release.

PR #6901, opened 2017-09-07
fix(selection-list): fix option value coercion and selection events

### LINKED ISSUE #6864: [MdListOption] value always boolean, selectChange/deselected events don't fire

MdListOption's output EventEmitters selectChange, and deselected do not fire when the item is selected/deselected.

They should activate the angular expression bound to them.

Here's the code block I'm working with:

```
<md-selection-list class="selection-content-list">
    <md-list-option class="list-item" (selectChange)="selectedItem($event)" (deselected)="deselectedItem($event)" [disabled]="selectionItem.disabled" *ngFor="let selectionItem of (elements | async)">
        <md-icon mdListIcon>folder</md-icon>
        <h4 mdLine>{{ selectionItem.name }}</h4>
    </md-list-option>
</md-selection-list>
```

The items visually update when being clicked on and off. 
The MdSelectionList's selectedOptions variable updates properly.
The selectChange and deselected EventEmitters are never activated.

Looking at the [current code](https://github.com/angular/material2/blob/fbce67d219aaa6e903aec0d261634d16f5ae7dcc/src/lib/list/selection-list.ts), it looks like they aren't ever used or references apart from their declarations.

The focus and destroyed events seem to emit fine though.

Also the value variable for MdListOption is _coerced to a boolean value_ so even using selectedOptions doesn't make the MdListOption useable.

dereekb:
* md-list-option value is no longer coerced to a boolean value
* md-list-option EventEmitters selectedChange and deselected now emit an event when an option is selected/deselected

Fixes #6864

### DIFF
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -586,17 +586,17 @@ class SelectionListWithListOptions {
 }
 
 @Component({template: `
-  <mat-selection-list id = "selection-list-2">
-    <mat-list-option checkboxPosition = "after">
+  <mat-selection-list id="selection-list-2">
+    <mat-list-option checkboxPosition="after">
       Inbox (disabled selection-option)
     </mat-list-option>
-    <mat-list-option id = "testSelect" checkboxPosition = "after">
+    <mat-list-option id="testSelect" checkboxPosition="after">
       Starred
     </mat-list-option>
-    <mat-list-option checkboxPosition = "after">
+    <mat-list-option checkboxPosition="after">
       Sent Mail
     </mat-list-option>
-    <mat-list-option checkboxPosition = "after">
+    <mat-list-option checkboxPosition="after">
       Drafts
     </mat-list-option>
   </mat-selection-list>`})
@@ -604,17 +604,17 @@ class SelectionListWithCheckboxPositionAfter {
 }
 
 @Component({template: `
-  <mat-selection-list id = "selection-list-3" [disabled] = true>
-    <mat-list-option checkboxPosition = "after">
+  <mat-selection-list id="selection-list-3" [disabled]=true>
+    <mat-list-option checkboxPosition="after">
       Inbox (disabled selection-option)
     </mat-list-option>
-    <mat-list-option id = "testSelect" checkboxPosition = "after">
+    <mat-list-option id="testSelect" checkboxPosition="after">
       Starred
     </mat-list-option>
-    <mat-list-option checkboxPosition = "after">
+    <mat-list-option checkboxPosition="after">
       Sent Mail
     </mat-list-option>
-    <mat-list-option checkboxPosition = "after">
+    <mat-list-option checkboxPosition="after">
       Drafts
     </mat-list-option>
   </mat-selection-list>`})
@@ -638,8 +638,8 @@ class SelectionListWithSelectedOption {
 }
 
 @Component({template: `
-  <mat-selection-list id = "selection-list-4">
-    <mat-list-option checkboxPosition = "after" class="test-focus" id="123">
+  <mat-selection-list id="selection-list-4">
+    <mat-list-option checkboxPosition="after" class="test-focus" id="123">
       Inbox
     </mat-list-option>
   </mat-selection-list>`})
@@ -660,7 +660,7 @@ class SelectionListWithTabindexBinding {
 }
 
 @Component({template: `
-<mat-selection-list id = "selection-list-5">
+<mat-selection-list id="selection-list-5">
   <mat-list-option [value]="1" checkboxPosition="after">
     1
   </mat-list-option>
@@ -675,7 +675,7 @@ class SelectionListWithMultipleValues {
 }
 
 @Component({template: `
-<mat-selection-list id = "selection-list-6">
+<mat-selection-list id="selection-list-6">
   <mat-list-option (selectionChange)="onOptionSelectionChange($event)">
     Inbox
   </mat-list-option>
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -4,7 +4,7 @@ import {createKeyboardEvent, dispatchFakeEvent} from '@angular/cdk/testing';
 import {Component, DebugElement} from '@angular/core';
 import {async, ComponentFixture, inject, TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
-import {MatListModule, MatListOption, MatSelectionList, MatSelectionListOptionEvent} from './index';
+import {MatListModule, MatListOption, MatSelectionList, MatListOptionChange} from './index';
 
 
 describe('MatSelectionList', () => {
@@ -548,7 +548,7 @@ describe('MatSelectionList', () => {
       let selected: boolean = false;
 
       spyOn(testComponent, 'onOptionSelectionChange')
-        .and.callFake((event: MatSelectionListOptionEvent) => {
+        .and.callFake((event: MatListOptionChange) => {
           selected = event.selected;
         });
 
@@ -681,5 +681,5 @@ class SelectionListWithMultipleValues {
   </mat-list-option>
 </mat-selection-list>`})
 class SelectionListWithOptionEvents {
-  onOptionSelectionChange: (event?: MatSelectionListOptionEvent) => void = () => {};
+  onOptionSelectionChange: (event?: MatListOptionChange) => void = () => {};
 }
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -4,7 +4,7 @@ import {createKeyboardEvent, dispatchFakeEvent} from '@angular/cdk/testing';
 import {Component, DebugElement} from '@angular/core';
 import {async, ComponentFixture, inject, TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
-import {MatListModule, MatListOption, MatSelectionList} from './index';
+import {MatListModule, MatListOption, MatSelectionList, MatSelectionListOptionEvent} from './index';
 
 
 describe('MatSelectionList', () => {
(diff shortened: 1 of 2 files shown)


src/lib/list/selection-list.ts
```diff
@@ -97,7 +97,7 @@ export class MdListOption extends _MdListOptionMixinBase
 
   @Input()
   get value() { return this._value; }
-  set value( val: any) { this._value = coerceBooleanProperty(val); }
+  set value( val: any) { this._value = val; }
```
willshowell: The getter and setter can be removed if you're no longer coercing

rafaelss95:
I just want to ask something: why is there an event for `select` and also for `deselect`?

I think it can be just one event, like in checkbox:

  ```
private _emitChangeEvent() {
    let event = new MdCheckboxChange();
    event.source = this;
    event.checked = this.checked;

    this._controlValueAccessorChangeFn(this.checked);
    this.change.emit(event);
  }
```

Also, is the `(destroyed)` useful at some point?

**Comment by jelbourn** [comment]
Yeah, that's a good point. It should just have one `selectionChanged` event where the event includes the new state.

**Comment by dereekb** [comment]
In my current project I am watching the destroyed event so I can remove options that are no longer available. You could probably use the changes QueryList in selectionList, but you'd have to figure out which ones were removed.

Looking at the Material code again, the MdSelectionList itself subscribes to the destroyed events.

**Comment by rafaelss95** [comment]
@jelbourn what do you think about other statements?

**Comment by jelbourn** [comment]
@rafaelss95 those should be covered under #6677

**Comment by dereekb** [comment]
I went ahead and pushed changes that were relevant to the original issue.

I didn't rename MdSelectionListOptionEvent, but I changed it to a class now so it stayed consistent with similar classes.

**Comment by ptretyakov** [comment]
Hello, Everybody!
When are you merge the pull request? 
Thank you for your work :)

**Comment by mmalerba** [comment]
please rebase

**Comment by dereekb** [comment]
rebase with master or _presubmit? Looks like someone else made some of the the same changes

**Comment by dereekb** [comment]
Scratch that, I broke the fixed

**Comment by dereekb** [comment]
Ok, rebased with changes #6983 and #6971 included.

**Comment by kara** [comment]
@dereekb Would you mind rebasing one more time?

**Comment by dereekb** [comment]
Sure.

I'm rebasing right now, but it looks like a lot of the changes in this pull request are already on master one way or another from other pull requests.

The only changes are changing MatSelectionListOptionEvent to a class, and merging selectChange and deselected together into a single event emitter, that last one being a breaking change.

**Comment by dereekb** [comment]
Noticed I need to rebase it again. Working on that now.

**Comment by dereekb** [comment]
Ok, rebased and all the unit tests passed on my end.

I want to bring up a slight change from this rebase:

I removed _selected and instead have the getter use the selectionList's selectedOptions isSelected() for whether or not the object is selected.

My reasoning is that the selection's state shouldn't be in two places (by this I mean MatSelectionList and other components check selectedOptions for is selected, while the MatListOption checks itself), and the performance change is minimal since a Set is used for selectedOptions.

---

Another thing, since selectedOptions in MatSelectionList is publicly accessible, externally modifying selectedOptions should also send selectionChange events on the MatListOption right? I don't think it would, so it's something to consider.

**Comment by adarowiec** [comment]
Can you guys already pull it? I really need this fix in my project.

File: src/lib/list/selection-list.spec.ts [hunk 2]
```diff
@@ -579,3 +658,28 @@ class SelectionListWithTabindexBinding {
   tabIndex: number;
   disabled: boolean;
 }
+
+@Component({template: `
+<mat-selection-list id = "selection-list-5">
```
rafaelss95: remove the around spaces: `id="selection-list-5"`?

src/lib/list/selection-list.ts
```diff
@@ -174,6 +169,16 @@ export class MatListOption extends _MatListOptionMixinBase
     this.selectionList._setFocusedOption(this);
   }
 
+  /** Creates a selection event object from the specified option. */
+  private _createChangeEvent(option: MatListOption = this): MatSelectionListOptionEvent {
+    let event = new MatSelectionListOptionEvent();
```
rafaelss95: `const`?

src/lib/list/selection-list.spec.ts
dereekb: I don't know if I hit the format button on my side, but there are spaces around most of the unit test component's id's in this file.

I'll fix them.

src/lib/list/selection-list.ts
```diff
@@ -52,8 +52,9 @@ export class MatListOptionBase {}
 export const _MatListOptionMixinBase = mixinDisableRipple(MatListOptionBase);
 
 /** Event emitted by a selection-list whenever the state of an option is changed. */
-export interface MatSelectionListOptionEvent {
+export class MatSelectionListOptionEvent {
   option: MatListOption;
+  selected: boolean;
 }
```
rafaelss95: I'd keep the consistency between this and others:

/** Change event object emitted by MatSelectionList. */
export class MatSelectionListChange {
  /** The source MatListOption of the event. */
  source: MatListOption;
  /** The new `checked` value of the option. */
  checked: boolean;
}

See [https://github.com/angular/material2/blob/master/src/lib/checkbox/checkbox.ts#L71](checkbox for example).

dereekb: So checked instead of selected? Gotcha.

rafaelss95: Also adding comments will be useful so it can be shown in docs.

dereekb: By others you mean other checkboxes? Add comments?

Changing it from selected to checked seems like a step in the wrong direction, since it is MatListOption.selected, not MatListOption.checked.

src/lib/list/selection-list.ts
```diff
@@ -97,34 +97,29 @@ export class MatListOption extends _MatListOptionMixinBase
...
   get disabled() { return (this.selectionList && this.selectionList.disabled) || this._disabled; }
   set disabled(value: any) { this._disabled = coerceBooleanProperty(value); }
 
+  /** Value of the option */
+  @Input() value: any;
+
   /** Whether the option is selected. */
   @Input()
-  get selected() { return this._selected; }
+  get selected() { return this.selectionList.selectedOptions.isSelected(this); }
```
rafaelss95: Can you add `:boolean` here?

src/lib/list/selection-list.ts
rafaelss95: You're right. But what about `source` instead of `option`? All `change` events use `source`.

dereekb: Sounds good, I'll change that.

src/lib/list/selection-list.ts
```diff
@@ -97,34 +97,29 @@ export class MatListOption extends _MatListOptionMixinBase
   /** Whether the label should appear before or after the checkbox. Defaults to 'after' */
   @Input() checkboxPosition: 'before' | 'after' = 'after';
 
-  /** Value of the option */
-  @Input() value: any;
-
   /** Whether the option is disabled. */
   @Input()
   get disabled() { return (this.selectionList && this.selectionList.disabled) || this._disabled; }
```
rafaelss95: `: boolean`?

src/lib/list/selection-list.ts
```diff
@@ -97,34 +97,29 @@ export class MatListOption extends _MatListOptionMixinBase
...
   set disabled(value: any) { this._disabled = coerceBooleanProperty(value); }
```
rafaelss95: `: boolean`?

dereekb: Value is the actual selection option value. 

We don't want to coerce it to a boolean, again! :stuck_out_tongue:

src/lib/list/selection-list.spec.ts
rafaelss95: You're right. There are a lot of strange spaces in this spec file.

src/lib/list/selection-list.ts
rafaelss95: I'm not talking about `value`, but `disabled`.

rafaelss95: In `selected` we have: `set selected(value: boolean) {`

dereekb: Gotcha, fixed.

rafaelss95: I think you misunderstood what I mean... I mean that we have the following signature for selected: `set selected(value: boolean) { ... = coerceBooleanProperty(value); }`, so the `disabled` should follow the same thing: `set disabled(value: boolean) { ... = coerceBooleanProperty(value); }`. Did you get it now?

src/lib/list/selection-list.ts
```diff
@@ -111,7 +111,7 @@ export class MatListOption extends _MatListOptionMixinBase
   @Input()
   get selected(): boolean { return this.selectionList.selectedOptions.isSelected(this); }
   set selected(value: boolean) {
-    const isSelected = coerceBooleanProperty(value);
+    const isSelected: boolean = coerceBooleanProperty(value);
```
rafaelss95: I think you misunderstood what I mean... I mean that we have the following signature for selected: `set selected(value: boolean) { ... = coerceBooleanProperty(value); }`, so the `disabled` should follow the same thing: `set disabled(value: boolean) { ... = coerceBooleanProperty(value); }`. 

Basically change the signature of disabled from `set disabled(value: any) {` to `set disabled(value: boolean) {`.

 Did you get it now?

src/lib/list/selection-list.ts
dereekb: Ah, ok. The highlighted code I see is both on this section:

```
/** Value of the option */
-  @Input() value: any;
```

So I was kind of guessing what you meant. Making those changes now.

src/lib/list/selection-list.ts
```diff
@@ -51,16 +51,18 @@ export const _MatSelectionListMixinBase =
 export class MatListOptionBase {}
 export const _MatListOptionMixinBase = mixinDisableRipple(MatListOptionBase);
 
-/** Event emitted by a selection-list whenever the state of an option is changed. */
+/** Change event object emitted by MatListOption */
 export class MatSelectionListOptionEvent {
```
rafaelss95: `MatListOptionChange`?

src/lib/list/selection-list.ts
dereekb: Yea, the highlighted section GitHub was showing was:

```
/** Value of the option */
-  @Input() value: any;
```

So I didn't really get what you were talking about. It's changes now and I removed the unnecessary :boolean.

rafaelss95: Aha, ok. Nice :)

rafaelss95: Aha, ok. Nice :)

Will you remove this `:boolean` in another commit?

dereekb: That's the second time today I've committed without first hitting the save button. Done.

andrewseguin:
Can you check on the tests and get them green?

dereekb:
The tests that failed aren't related to the changes (they have to do with MatAutocomplete).

Those failed tests are related to #7743.

alisonmsmith:
FYI - material docs specify the event output as 'selectChange' which is of type EventEmitter<MatSelectionListOptionEvent> and instead it looks like it is implemented here as 'selectionChange' of type EventEmitter<MatListOptionChange> also the 'deselected' event from the docs is still not emitted (although you can check if the option was selected or deselected from the MatListOptionChange boolean selected property)

dereekb:
Yea, the docs need to be updated to reflect the changes.

selected/deselected were removed and replaced with selectionChange, since having a single selectionChange is more consistent in terms of the rest of Angular Material implementations.

As for the selectChange/selectionChange name, there was another discussion thread where I believe selectionChange was picked as the more desirable. I don't see the reference in this thread anymore, but it was about naming consistency and updating across a number of different Material components.

Issue #6726, opened 2017-08-30
Allow md-expansion-panel toggle to be placed at the beginning


alex321:
#### Feature:

Give an option to allow the toggle of the `md-expansion-panel` to be placed at the beginning.

#### What is the current behavior?

The toggle is always placed at the end of the panel.

#### What is the use-case or motivation for changing an existing behavior?

Frequently people prefer to place it before the text for the panel.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
2.0.0-beta.10

zladuric:
Hi,
Just a thought on this - while you're at it, could there also be a "no indicator at all" option? I mean, open panel only programatically and hide indicator? I guess those'd be two feature requests, but I'm willing to give it a try if somebody points me in the right direction.

I have such a use case now, so I'm struggling to steal clicks from the panel and it's ugly.

josephperrott:
The current PR that for this feature actually has a `hidden` option.

For the ability to make it only open programmatically, I would suggest filing a separate issue to track it

g-sari:
I opened this issue which is similar to this topic
https://github.com/angular/material2/issues/9337

josephperrott:
Reopening as the PR to address this was rolled back

PR #6917, opened 2017-09-07
feat(expansion): Allow expansion indicator positioning

josephperrott:
#6726 

Allows expansion indicator to be positioned at start, end and hidden.

### DIFF
--- a/src/lib/expansion/expansion-panel-header.scss
+++ b/src/lib/expansion/expansion-panel-header.scss
@@ -1,9 +1,8 @@
-
 .mat-expansion-panel-header {
   display: flex;
   flex-direction: row;
   align-items: center;
-  padding: 0 24px;
+  padding: 0 16px;
 
   &:focus,
   &:hover {
@@ -25,6 +24,7 @@
   flex: 1;
   flex-direction: row;
   overflow: hidden;
+  padding: 0 8px;
 }
 
 .mat-expansion-panel-header-title,
@@ -43,16 +43,18 @@
   flex-grow: 2;
 }
 
-/**
- * Creates the expansion indicator arrow. Done using ::after rather than having
- * additional nodes in the template.
- */
-.mat-expansion-indicator::after {
+.mat-expansion-indicator-container {
+  // An offset is required to offset the entire expansion indicator against its border width.  It is
+  // calculated as sqrt(2 * (border-width)).
+  margin-bottom: 2.83px;
+  padding: 0 8px;
+  width: 8px;
+}
+
+.mat-expansion-indicator {
   border-style: solid;
   border-width: 0 2px 2px 0;
-  content: '';
   display: inline-block;
   padding: 3px;
   transform: rotate(45deg);
-  vertical-align: middle;
 }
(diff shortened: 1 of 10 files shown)


src/lib/expansion/expansion-panel.ts
```diff
@@ -72,8 +72,8 @@ export const EXPANSION_PANEL_ANIMATION_TIMING = '225ms cubic-bezier(0.4,0.0,0.2,
   ],
 })
 export class MdExpansionPanel extends AccordionItem implements OnChanges, OnDestroy {
-  /** Whether the toggle indicator should be hidden. */
-  @Input() hideToggle: boolean = false;
+  /** The positioning of the expansion indicator. */
+  @Input() togglePosition: string = 'end';
```
jelbourn: The type would be `'start' | 'end' | 'hidden'`

@crisbeto @kara what do you think about this API vs. having both `togglePosition` (only start/end) and `hideToggle`?

crisbeto: I think I prefer the single `togglePosition` since it avoids bloating the API for a somewhat minor feature. It's also consistent with other inputs like the `MdFormField.floatPlaceholder`  which is `'always' | 'never' | 'auto'`.

src/lib/expansion/expansion-panel-header.scss
```diff
@@ -43,16 +43,16 @@
   flex-grow: 2;
 }
 
-/**
- * Creates the expansion indicator arrow. Done using ::after rather than having
- * additional nodes in the template.
- */
-.mat-expansion-indicator::after {
+.mat-expansion-indicator-container {
+  margin-bottom: 2.83px;
```
jelbourn: These is weirdly specific- is there any way this can be calculated or set as the result of other known values?

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -137,8 +137,16 @@ export class MdExpansionPanelHeader implements OnDestroy {
...
+  _indicatorHidden(): boolean {
+    return !this.panel.disabled && this.panel.togglePosition !== 'hidden';
+  }
+
+  /** Gets the order value to place the expansion indicator in the correct position. */
+  _indicatorOrder(): number {
+    if (this.panel.togglePosition === 'start') {
+      return -1;
+    }
+    return 1;
```
jelbourn:
```ts
return this.panel.togglePosition === 'start' ? -1 : 1;
```

src/lib/expansion/expansion-panel.ts
```diff
@@ -86,11 +86,11 @@ export class MdExpansionPanel extends AccordionItem implements OnChanges, OnDest
...
 
   /** Whether the expansion indicator should be hidden. */
-  _getHideToggle(): boolean {
+  _getTogglePosition(): string {
     if (this.accordion) {
-      return this.accordion.hideToggle;
+      return this.accordion.togglePosition;
     }
-    return this.hideToggle;
+    return this.togglePosition;
```
jelbourn:
```ts
return this.accordion ? this.accordion.togglePosition : this.togglePosition;
```

josephperrott:
@jelbourn When you have a moment can you take a look at this PR?  I don't remember where we landed on this one.

jelbourn:
@josephperrott as discussed we're going to keep `hideToggle` and make `togglePosition` separate to avoid the breaking API change.

josephperrott:
Closing in favor of #8199

PR #6935, opened 2017-09-08
docs(table): remove duplicate css in the table-filtering-example

Teamop:
Remove the duplicate css in the table-filtering-example.css, just keep the current valid one.

### DIFF
--- a/src/material-examples/table-filtering/table-filtering-example.css
+++ b/src/material-examples/table-filtering/table-filtering-example.css
@@ -8,14 +8,6 @@
 .example-header {
   min-height: 64px;
   display: flex;
-  align-items: center;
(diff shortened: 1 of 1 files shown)


Teamop:
I noticed a test failed in Screenshot Tests.
>mdcard cardfancy should show a card fancy card example
**Failed**
I don't change anything with mdcard. Weird.

src/material-examples/table-filtering/table-filtering-example.css
```diff
@@ -8,14 +8,6 @@
 .example-header {
```
crisbeto: Can you combine these styles in the same selector?

PR #6947, opened 2017-09-08
build: more type information for docs

DevVersion:
* Upgrades dgeni and dgeni-packages to the latest available version (including TypeScript support; and newer TypeScript version for AST)
* More warnings if type information is missing on public TypeScript symbol
* Adds missing explicit type declarations to all components
* Types are now being displayed for accessors and properties (this was broken before somehow)
* Rewrote the Dgeni setup in TypeScript for type safety and better integration into the gulp workflow.

**Note**: Most of the things inside of the existing Dgeni setup is the same as before and I didn't look into other improvements there (e.g logic simplification) because the PR is big enough already.


src/lib/form-field/form-field.ts
```diff
@@ -94,7 +94,7 @@ export class MdFormField implements AfterViewInit, AfterContentInit, AfterConten
 
   /** @deprecated Use `color` instead. */
   @Input()
-  get dividerColor() { return this.color; }
+  get dividerColor(): 'primary' | 'accent' | 'warn' { return this.color; }
```
crisbeto: We have the `ThemePalette` type for this already.

tools/dgeni/common/decorators.ts
```diff
@@ -0,0 +1,113 @@
+/**
+ * We want to avoid emitting selectors that are deprecated but don't have a way to mark
+ * them as such in the source code. Thus, we maintain a separate blacklist of selectors
+ * that should not be emitted in the documentation.
+ */
+import {ClassExportDoc} from 'dgeni-packages/typescript/api-doc-types/ClassExportDoc';
+import {PropertyMemberDoc} from 'dgeni-packages/typescript/api-doc-types/PropertyMemberDoc';
+import {MemberDoc} from 'dgeni-packages/typescript/api-doc-types/MemberDoc';
+
+const SELECTOR_BLACKLIST  = new Set([
```
crisbeto: Some inconsistent spacing here.

tools/dgeni/processors/categorizer.ts
```diff
@@ -0,0 +1,129 @@
+import {Processor, DocCollection} from 'dgeni';
+import {ClassExportDoc} from 'dgeni-packages/typescript/api-doc-types/ClassExportDoc';
+import {PropertyMemberDoc} from 'dgeni-packages/typescript/api-doc-types/PropertyMemberDoc';
+import {
+  NormalizedMethodMemberDoc, normalizeMethodParameters
```
crisbeto: The `normalizeMethodParameters` should be on a separate line.

src/lib/form-field/form-field.ts
DevVersion: Yeah I wanted to use that as well, but it also includes `undefined` as a possible value, which is not valid here.

DevVersion:
@crisbeto Addressed your feedback and rebased on top of the [sort members](https://github.com/angular/material2/commit/9f2fd477f24cda4e634209510162d171572ccaa0) commit. 

Regarding the enforcement: Good point, I have a TSLint rule in mind. We can base those rules on top of the existing rules like `typedef` (will be something as a follow-up)

DevVersion:
@jelbourn Rebased this one. Please have another look.

tools/dgeni/common/decorators.ts
```diff
@@ -0,0 +1,113 @@
...
+export function isGenericTypeParameter(doc: MemberDoc) {
+  if (doc.containerDoc instanceof ClassExportDoc) {
+    return doc.containerDoc.typeParams && `<${doc.name}>` === doc.containerDoc.typeParams;
+  }
+  return false;
+}
+
+export function isProperty(doc: MemberDoc) {
+  if (doc instanceof PropertyMemberDoc ||
+      // The latest Dgeni version no longer treats getters or setters as properties. In theory
```
jelbourn: "In theory those" -> "From a user perspective, these"

tools/dgeni/common/dgeni-accessors-parse.ts
```diff
@@ -0,0 +1,54 @@
...
+  SyntaxKind,
+  GetAccessorDeclaration,
+  SetAccessorDeclaration,
+  Symbol,
+  displayPartsToString
+} from 'typescript';
+import {CategorizedPropertyMemberDoc} from '../processors/categorizer';
+
+/**
+ * Accessors are not parsed properly within the latest dgeni-packages package.
```
jelbourn: Is there an issue to link for this? What about the accessors isn't parsed correctly?

tools/dgeni/common/dgeni-accessors-parse.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    // Walk through every TypeScript member node of the class declaration and filter for accessors
+    // with a name that match the current property member. For accessors there may be two
+    // declarations with the same name but with different type information.
+    classDeclaration.members
+      .filter((member: any) => member.kind === SyntaxKind.GetAccessor ||
+        member.kind === SyntaxKind.SetAccessor)
+      .filter((member: any) => member.name && (member.name as any).text === propertyDoc.name)
+      .forEach((member: GetAccessorDeclaration|SetAccessorDeclaration) => {
+        if (member.type) {
```
jelbourn: Add comment that explains how you extract the type

tools/dgeni/common/dgeni-accessors-parse.ts
```diff
@@ -0,0 +1,54 @@
...
+  if ((propertyDoc.isGetAccessor || propertyDoc.isSetAccessor) && !propertyDoc.type) {
+    const classDeclaration = propertyDoc.containerDoc.symbol.valueDeclaration as ClassDeclaration;
+
+    // Walk through every TypeScript member node of the class declaration and filter for accessors
+    // with a name that match the current property member. For accessors there may be two
+    // declarations with the same name but with different type information.
+    classDeclaration.members
+      .filter((member: any) => member.kind === SyntaxKind.GetAccessor ||
+        member.kind === SyntaxKind.SetAccessor)
+      .filter((member: any) => member.name && (member.name as any).text === propertyDoc.name)
```
jelbourn: If `member` is `any`, why do you need to use `as any` on `member.name`?

tools/dgeni/common/dgeni-accessors-parse.ts
```diff
@@ -0,0 +1,54 @@
...
+          if (type) {
+            propertyDoc.type = getTypeText(type, []);
+            propertyDoc.decorators = getDecorators(member);
+          }
+        }
+
+        // If the current property document doesn't have any description set and the current
+        // TypeScript has a symbol from the TypeChecker, we can try to find a documentation
+        // comment on that symbol.
+        if (!propertyDoc.description && (member as any).symbol) {
```
jelbourn: If you need the `as any` here, is `GetAccessorDeclaration|SetAccessorDeclaration` the right type for member?

DevVersion: Yeah this is the correct type. It just looks like TypeScript isn't exposing the `symbol` property in their typings. The same is happening on Dgeni as well ([see here](https://github.com/angular/dgeni-packages/blob/master/typescript/src/services/TsParser/index.ts#L56))

tools/dgeni/common/dgeni-accessors-parse.ts
DevVersion: I only reported this on Slack if I recall. Created an issue for it now: https://github.com/angular/dgeni-packages/issues/246

DevVersion:
@jelbourn Addressed your feedback.

PR #6949, opened 2017-09-08
fix(datepicker): toggle should disable compat

### LINKED ISSUE #6942: [mdDatepickerToggle] Error in compatibility mode in v2.0.0-beta.10
Bug

mat-datepicker-toggle should render button correctly in compatibility mode.

Error is thrown when rendering mat-datepicker-toggle:
`
Error: The "md-" prefix cannot be used in ng-material v1 compatibility mode. It was used on an "button" element.
`

Angular 4.4.0-RC.0
Angular material 2.0.0-beta.10
Chrome 60.0.3112.113

Error cause seems to be in this file: https://github.com/angular/material2/blob/master/src/lib/datepicker/datepicker-toggle.html

md-icon-button directive is included in the button element, which is forbidden in the compatibility mode (as far as I understand).

andrewseguin:
The datepicker toggle uses `md-icon-button` and `md-icon` but this fails in compatibility mode. Force `MATERIAL_COMPATIBILITY_MODE` to be false for this component so that it does not fail in either mode

Closes #6942

### DIFF
--- a/src/lib/datepicker/datepicker-toggle.ts
+++ b/src/lib/datepicker/datepicker-toggle.ts
@@ -18,6 +18,7 @@ import {MdDatepicker} from './datepicker';
 import {MdDatepickerIntl} from './datepicker-intl';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {Subscription} from 'rxjs/Subscription';
+import {MATERIAL_COMPATIBILITY_MODE} from '../core';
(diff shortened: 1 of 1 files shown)


mmalerba:
please rebase

andrewseguin:
No longer necessary since we are changing internal HTML to use mat instead of md anyways

PR #6955, opened 2017-09-09
fix(select): extra whitespace around placeholder

### LINKED ISSUE #6923: bug(select) Extra space at the end of placeholder span
#### Bug:
I have the following code:

```
<md-select placeholder="hello" [value]="1">
  <md-option [value]="1">one</md-option>
  <md-option [value]="2">two</md-option>
</md-select>
```

The most important part is `placeholder="hello"`. Angular Material will create separate span for placeholder and put string `hello` inside it.

The resulting span should look like `<span>hello</span>`, **without** extra space at the end.

The resulting span looks like `<span>hello </span>`, **with** extra space at the end:

image

Plunker: http://plnkr.co/edit/0egSObwCtsCqxVMW6WRH?p=preview

If placeholder consists of two words, then this extra space may cause unnecessary word wrap.

At least Arch Linux, Angular 4.0.0, Material 2.0.0-beta.10, TypeScript 2.3.3, Chrome 59.

I think you can fix this issue by remove spaces around `{{ placeholder }}` in `> {{ placeholder }} </span>` ([this](https://github.com/angular/material2/blob/master/src/lib/select/select.html#L13) line).

crisbeto:
Fixes some extra whitespace being added around the placeholder.

Fixes #6923.

### DIFF
--- a/src/lib/select/select.scss
+++ b/src/lib/select/select.scss
@@ -76,7 +76,7 @@ $mat-select-trigger-underline-height: 1px !default;
 
   // TODO: Double-check accessibility of this style
   .mat-select-required &::after {
-    content: '*';
+    content: ' *';
   }
 }
 
--- a/src/lib/select/select.html
+++ b/src/lib/select/select.html
@@ -10,7 +10,7 @@
     [class.mat-floating-placeholder]="_hasValue()"
     [@transformPlaceholder]="_getPlaceholderAnimationState()"
     [style.opacity]="_getPlaceholderOpacity()"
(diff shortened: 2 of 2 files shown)


src/lib/select/select.scss
```diff
@@ -76,7 +76,7 @@ $mat-select-trigger-underline-height: 1px !default;
 
   // TODO: Double-check accessibility of this style
   .mat-select-required &::after {
-    content: '*';
+    content: ' *';
```
donroyco: That's also a solution. Thanks @crisbeto

PR #6957, opened 2017-09-09
build: cancel screenshot diff if there is no activity

DevVersion:
* No longer just gives the screenshot diff task 8 minutes to pass, because the run time can vary. Now it just exits the process if the last action in the task takes longer than 6 minutes.
* Instead of just closing the firebase connection, the process will be exited with an error code. Just closing the connection is not helpful and breaks the screenshot diff Web Interface.

**Note**: Should wait for #6956 before review.

### DIFF
--- a/tools/gulp/tasks/screenshots.ts
+++ b/tools/gulp/tasks/screenshots.ts
@@ -26,15 +26,18 @@ const LOCAL_GOLDENS = path.join(SCREENSHOT_DIR, `golds`);
 const LOCAL_DIFFS = path.join(SCREENSHOT_DIR, `diff`);
 
 // Directory to which untrusted screenshot results are temporarily written
-//   (without authentication required) before they are verified and copied to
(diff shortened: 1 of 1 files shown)


tools/gulp/tasks/screenshots.ts
```diff
@@ -34,6 +34,9 @@ const FIREBASE_IMAGE = `${TEMP_FOLDER}/screenshot/images`;
 const FIREBASE_DATA_GOLDENS = `screenshot/goldens`;
 const FIREBASE_STORAGE_GOLDENS = 'goldens';
 
+const lastActionTimeout = 1000 * 60 * 6;
+const lastActionRefreshInterval = 1000 * 45;
```
jelbourn: Add comments with the human-readable times?

DevVersion: Sure. Done

PR #6961, opened 2017-09-10
fix(menu): multiple close events for a single close

ppham27:
Don't emit a closed event when another event will be emitted. Previously, if one clicked on a menu item, one would get two events: `undefined` and `click` in that order. One would see similar behavior for `keydown` or clicking the backdrop. Unit tests were updated to prevent a regression.

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -139,7 +139,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   }
 
   /** Event emitted when the menu is closed. */
-  @Output() close = new EventEmitter<void | 'click' | 'keydown'>();
+  @Output() close = new EventEmitter<void | 'click' | 'destroy' | 'keydown'>();
 
   constructor(
     private _elementRef: ElementRef,
@@ -152,7 +152,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
 
   ngOnDestroy() {
     this._tabSubscription.unsubscribe();
-    this.close.emit();
+    this.close.emit('destroy');
     this.close.complete();
   }
 
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -139,7 +139,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   }
 
   /** Event emitted when the menu is closed. */
-  @Output() close = new EventEmitter<void | 'click' | 'destroy' | 'keydown'>();
+  @Output() close = new EventEmitter<void | 'click' | 'keydown'>();
 
   constructor(
     private _elementRef: ElementRef,
@@ -152,7 +152,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
 
   ngOnDestroy() {
     this._tabSubscription.unsubscribe();
-    this.close.emit('destroy');
+    this.close.emit();
     this.close.complete();
   }
 
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -43,7 +43,7 @@ import {MdMenuItem} from './menu-item';
 import {MdMenuPanel} from './menu-panel';
 import {MenuPositionX, MenuPositionY} from './menu-positions';
 import {throwMdMenuMissingError} from './menu-errors';
-import {of as observableOf} from 'rxjs/observable/of';
+import {empty} from 'rxjs/observable/empty';
 import {merge} from 'rxjs/observable/merge';
 import {Subscription} from 'rxjs/Subscription';
 
@@ -153,7 +153,7 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
     this._checkMenu();
 
     this.menu.close.subscribe(reason => {
-      this._closeMenu(false);
+      this._closeMenu();
 
       // If a click closed the menu, we should close the entire chain of nested menus.
       if (reason === 'click' && this._parentMenu) {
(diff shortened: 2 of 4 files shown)


src/lib/menu/menu-directive.ts
```diff
@@ -139,7 +139,7 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
   }
 
   /** Event emitted when the menu is closed. */
-  @Output() close = new EventEmitter<void | 'click' | 'keydown'>();
+  @Output() close = new EventEmitter<void | 'click' | 'destroy' | 'keydown'>();
```
crisbeto: The idea with this one was to emit the reason only if it's user-generated. Given that there is no special behavior for `destroy`, can you revert it back to `void | 'click' | 'keydown'`?

src/lib/menu/menu-trigger.ts
```diff
@@ -216,23 +216,34 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
+  /** Focuses the menu trigger. */
+  focus() {
+    this._element.nativeElement.focus();
+  }
+
+  /**
+   * Gives the option of setting emitEvent to false to avoid emitting an event
+   * when an event will already be emitted.
+   */
+  private _closeMenu(emitEvent: boolean) {
```
crisbeto: Rather than having the `emitEvent` parameter, you can remove the `this.menu.close.emit` call from here and move it only to `closeMenu`. Also change the doc string to something along the lines of "Closes the menu and does the necessary cleanup".

src/lib/menu/menu.spec.ts
```diff
@@ -485,6 +485,7 @@ describe('MdMenu', () => {
       fixture = TestBed.createComponent(SimpleMenu);
       fixture.detectChanges();
       fixture.componentInstance.trigger.openMenu();
+      fixture.componentInstance.closeCallback.calls.reset();
```
crisbeto: Why is this necessary? The callbacks should be reset after each test, because the component is re-compiled.

src/lib/menu/menu.spec.ts
```diff
@@ -493,26 +494,40 @@ describe('MdMenu', () => {
...
 
       backdrop.click();
       fixture.detectChanges();
 
-      expect(fixture.componentInstance.closeCallback).toHaveBeenCalled();
+      expect(fixture.componentInstance.closeCallback).toHaveBeenCalledWith(undefined);
+      expect(fixture.componentInstance.closeCallback).toHaveBeenCalledTimes(1);
+    });
+
+    it('should emit an event when escaped', () => {
```
crisbeto: "should emit an event when pressing ESCAPE".

src/lib/menu/menu.spec.ts
```diff
@@ -989,11 +1004,18 @@ describe('MdMenu', () => {
 
       expect(menus.length).toBe(3, 'Expected three open menus');
 
+      instance.rootCloseCallback.calls.reset();
```
crisbeto: These reset calls shouldn't be necessary.

ppham27: They were necessary when because Observable.of(null) was triggering a close event. Replaced with Observable.empty()

src/lib/menu/menu.spec.ts
ppham27: Done.

src/lib/menu/menu.spec.ts
ppham27: They were necessary when because Observable.of(null) was triggering a close event. Replaced with Observable.empty()

src/lib/menu/menu-trigger.ts
ppham27: Done. Should I only emit the event when the menu is already open, though?

src/lib/menu/menu-directive.ts
ppham27: Done. I was also wondering if it's a bug that destroy will emit a close event when the menu is already closed?

crisbeto: It's there to notify any child menus to close.

src/lib/menu/menu.spec.ts
crisbeto: `Observable.of(null)` will emit `null` immediately. `Observable.of()` won't.

src/lib/menu/menu-directive.ts
ppham27: But child menus won't be open unless the menu itself is open? Anyway, I'm more than happy to leave it as it is, as it doesn't really affect any use cases that I can think of.

crisbeto: In most cases the children won't be open without a parent, but it's possible for the parent to be destroyed while the children are still open.

src/lib/menu/menu.spec.ts
ppham27: Are you saying that I should change it to Observable.of() instead of Observable.empty()? To me, Observable.empty() is more readable and makes the intent more clear.

But I don't mind change it to Observable.of() since the behavior would be the same as Observable.empty().

crisbeto: Observable.empty() is definitely valid in this case. I just want to avoid bringing in an extra symbol from RxJS.

src/lib/menu/menu-trigger.ts
```diff
@@ -216,23 +218,29 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
...
+
+  /** Focuses the menu trigger. */
+  focus() {
+    this._element.nativeElement.focus();
+  }
+
+  /**
+   * Closes the menu and does the necessary cleanup.
+   */
+  private _closeMenu() {
```
crisbeto: One last nitpick: can you rename this to `_destroyMenu` to avoid confusion with `closeMenu`?

src/lib/menu/menu.spec.ts
ppham27: Changed to Observable.of()

src/lib/menu/menu-trigger.ts
ppham27: Done.

PR #6971, opened 2017-09-10
feat(selection-list): add selectAll and deselectAll functions

### LINKED ISSUE #6969: [md-selection-list] selectAll function
Feature request

It would be great if md-selection-list could expose a function which would select all of its md-list-options when executed. I believe this is a common use-case for such a component.

Currently not possible and must be implemented manually.

crisbeto:
* Adds the ability to select and deselect all options at once.
* Adds a demo of the selection list to the demo app.
* Fixes an issue that caused the checkbox inside the selection list to collapse.
* Sets the proper cursor on the list items.

Fixes #6969.

### DIFF
--- a/src/demo-app/list/list-demo.html
+++ b/src/demo-app/list/list-demo.html
@@ -101,4 +101,22 @@
       </md-list-item>
     </md-nav-list>
   </div>
+
+  <div>
+    <h2>Selection list</h2>
+
+    <md-selection-list #groceries>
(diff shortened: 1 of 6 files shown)


jelbourn:
@crisbeto just one lint error

crisbeto:
That's the one that'll be fixed by https://github.com/angular/material2/pull/6956 @jelbourn.

PR #6975, opened 2017-09-10
feat(overlay): replace OverlayContainer themeClass w/ addClass/removeClass methods

jelbourn:
BREAKING CHANGE: Now that the Overlay is part of the cdk rather than Angular Material directly,
the `themeClass` property has been removed. To add a class to the
overlay for theming, you can do
```ts
overlayContainer.getContainerElement().classList.add('my-theme-class');
```

### DIFF
--- a/guides/theming.md
+++ b/guides/theming.md
@@ -169,7 +169,7 @@ Since certain components (e.g. menu, select, dialog, etc.) are inside of a globa
 an additional step is required for those components to be affected by the theme's css class selector
 (`.unicorn-dark-theme` in the example above).
 
-To do this, you can specify a `themeClass` on the global overlay container. For the example above,
+To do this, you can add the appropriate class to the global overlay container. For the example above,
 this would look like:
 ```ts
 import {OverlayContainer} from '@angular/material';
@@ -179,13 +179,10 @@ import {OverlayContainer} from '@angular/material';
 })
 export class UnicornCandyAppModule {
   constructor(overlayContainer: OverlayContainer) {
-    overlayContainer.themeClass = 'unicorn-dark-theme';
+    overlayContainer.getContainerElement().classList.add('unicorn-dark-theme');
   }
 }
 ```
-
-The `themeClass` of the `OverlayContainer` can be changed at any time to change the active theme
-class.
 
 #### Theming only certain components
 The `angular-material-theme` mixin will output styles for [all components in the library](https://github.com/angular/material2/blob/master/src/lib/core/theming/_all-theme.scss).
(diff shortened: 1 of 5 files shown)


src/cdk/overlay/overlay-container.ts
```diff
@@ -19,27 +19,29 @@ export class OverlayContainer implements OnDestroy {
...
-  set themeClass(value: string) {
-    if (this._containerElement) {
-      if (this._themeClass) {
-        this._containerElement.classList.remove(this._themeClass);
-      }
-
-      if (value) {
-        this._containerElement.classList.add(value);
-      }
+  addClass(cssClass: string) {
```
crisbeto: Since the `getContainerElement()` is public, I'm not sure that these methods are even necessary.

jelbourn: Yeah, I considered that. I was worried that the change would be "too complicated" for people, but maybe I'm overthinking it. I'll just remove the `themeClass` completely.

crisbeto: IMO we still need some easy way for people to pass the theme to overlays, though. Perhaps a provider that injects the overlay container and applies the class?

jelbourn: Is `getContainerElement().classList.add('dark-theme')` that must worse than `themeClasss = 'dark-theme';`?

crisbeto: I don't think it is, but I guess that it's not as obvious that that's the way to theme all components.

willshowell: IMO while `getContainerElement().classList.add('dark-theme')` isn't immediately obvious, it reduces the magic, which is nice.

Just don't forget [the example in the theming guide](https://github.com/angular/material2/blob/master/guides/theming.md#multiple-themes-and-overlay-based-components) needs to be updated with whichever solution!

jelbourn:
I went ahead and just removed `themeClass`. I'm still open open adding a more "obvious" way to apply a theme class, but I think this is good as a starting point.

guides/theming.md
```diff
@@ -169,7 +169,7 @@ Since certain components (e.g. menu, select, dialog, etc.) are inside of a globa
 an additional step is required for those components to be affected by the theme's css class selector
 (`.unicorn-dark-theme` in the example above).
 
-To do this, you can specify a `themeClass` on the global overlay container. For the example above,
+To do this, you can add the approviate class to the global overlay container. For the example above,
```
willshowell: typo "appropriate"

jelbourn:
Fixed typo and demo-app (which I missed before)

PR #6983, opened 2017-09-11
fix(selection-list): do not coerece option value to boolean

DevVersion:
Currently the value of a `MdListOption` is coerced to a boolean. This is not the correct behavior because it would mean that there could be only two values (true/false).

Needed for #6896


rafaelss95:
#6901 covers it already.

Issue #6952, opened 2017-09-08
Some plunkers are broken on the docs site


dev054:
#### Bug, feature request, or proposal:
Bug

#### What is the current behavior?
Pages:

https://material.angular.io/components/icon/examples
https://material.angular.io/components/list/overview
https://material.angular.io/components/snack-bar/examples

#### What are the steps to reproduce?
Just click and open console. There are some missing files.

josephperrott:
A new version of [material.angular.io](http://material.angular.io) was pushed recently.  Try clearing the cache and cookies on your browser and see if the issue continues.

jelbourn:
The errors being thrown are due to missing plunker files; might be related to the local asset copy script not copying the plunker assets

josephperrott:
Fixed by #6990

Issue #6951, opened 2017-09-08
Plunker generation is broken when there are images on demo


dev054:
#### Bug, feature request, or proposal:
Bug/Proposal.

#### What is the expected behavior?
I'd expect to see demo without anything broken in plunker.

#### What is the current behavior?
Actually if you click in "Edit in Plunker" button in [this page](https://material.angular.io/components/card/examples), the generated plunker will be "broken", in other words, without any images.

donroyco:
See #6754

dev054:
@donroyco not really related. The plunker itself is working... the problem is that this `Card` example has some images and it isn't being projected to the plunker... that's isn't something new, I'm just wondering if the images could be passed together with the plunker generation.

donroyco:
It _is_ relates, because that issue addresses the main issue here: how can we manage to get stable Plunker examples, because the broke  images you're referring to, worked in the past.

josephperrott:
Closed with #6990

PR #6990, opened 2017-09-11
fix(snackbar): Docs example fixes

josephperrott:
For #6952 #6951 

- Move snackbar inline
- Load card icons from urls absolutely rather than relatively


jelbourn:
@josephperrott can you rebase?

PR #6998, opened 2017-09-12
build: create JS bundles for each component

### LINKED ISSUE #4137: Tree-shaking for production build is not working properly
The production bundle size is not shrinking
 
Base on the new package structure the production bundle (ng build --prod) should not have included all the Material Components Module

Base on the change log now we should create our own Custom MaterialModule, in my case I have this one:
```ts
import { NgModule } from '@angular/core';
import {
    MdCardModule,
    MdChipsModule,
    MdToolbarModule,
    MdButtonModule,
    MdSidenavModule,
    MdRippleModule,
    MdIconModule,
    MdListModule,
    MdProgressSpinnerModule,
} from '@angular/material';

@NgModule({
    imports: [
        MdCardModule,
        MdChipsModule,
        MdToolbarModule,
        MdButtonModule,
        MdSidenavModule,
        MdRippleModule,
        MdIconModule,
        MdListModule,
        MdProgressSpinnerModule,
    ],
    declarations: [

    ],
    providers: [

    ],
    exports: [
        MdCardModule,
        MdChipsModule,
        MdToolbarModule,
        MdButtonModule,
        MdSidenavModule,
        MdRippleModule,
        MdIconModule,
        MdListModule,
        MdProgressSpinnerModule,
    ]
})
export class CustomMaterialModule { }
```
that is imported in the AppModule.

Even if I remove all the module and import only, for example, MdButtonModule, the size of material library in the production vendor.js still 308Kb 

@angular/cli: 1.0.0
node: 7.9.0
os: linux x64
@angular/animations: 4.0.1
@angular/common: 4.0.1
@angular/compiler: 4.0.1
@angular/core: 4.0.1
@angular/forms: 4.0.1
@angular/http: 4.0.1
@angular/platform-browser: 4.0.1
@angular/platform-browser-dynamic: 4.0.1
@angular/router: 4.0.1
@angular/cli: 1.0.0
@angular/compiler-cli: 4.0.1

This is the vendor sourcemap generated by source-map-explorer
image

jelbourn:
Before this change, all of `@angular/material` was bundled into a
single, flat JS file.

After this change, each top-level component (i.e., each directory under
`src/lib`) will be have its own JS file (and d.ts and metadata.json).

The benefit of this change is that current versions of webpack will be
able to do a much better job of only including the code an app is using.

It also allows (but does not require) people to explicitly import from specific components,
e.g.
```ts
import {MdButtonModule} from '@angular/material/button';
```

For development _within Angular Material itself_ we have to use these
deeper imports because the root `@angular/material` doesn't exist until
the actual package is created.

Additionally, this PR adds a `tsconfig.json` for each dir under `src/`
with the correct path mappings for that directory, so that IDEs will
correctly generate imports.

Closure Compiler CI check is also disabled by this PR; it cannot work
with the new package structure until a release with
google/closure-compiler#2600 is made available.

Fixes #4137


src/lib/tsconfig-build.json
```diff
@@ -5,7 +5,7 @@
     "declaration": true,
     "stripInternal": false,
     "experimentalDecorators": true,
-    "noUnusedParameters": true,
+    "noUnusedParameters": false,
```
DevVersion: Why are we no longer having this check? Maybe add a comment if necessary

tools/package-tools/entry-point-package-json.ts
```diff
@@ -8,8 +8,8 @@ export function createEntryPointPackageJson(destDir: string, packageName: string
     name: `@angular/${packageName}/${entryPointName}`,
     typings: `../${entryPointName}.d.ts`,
     main: `../bundles/${packageName}-${entryPointName}.umd.js`,
-    module: `../@angular/${packageName}/${entryPointName}.es5.js`,
-    es2015: `../@angular/${packageName}/${entryPointName}.js`,
+    module: `../esm5/${packageName}/${entryPointName}.es5.js`,
+    es2015: `../esm2015/${packageName}/${entryPointName}.js`,
```
DevVersion: Out of curiosity: How does this work with Closure Compiler now? Is there an alternative to the `js_module_roots` option?

src/lib/tsconfig-umd.json
```diff
@@ -0,0 +1,39 @@
...
+    "typings.d.ts"
+  ],
+  "angularCompilerOptions": {
+    "annotateForClosureCompiler": true,
+    "strictMetadataEmit": true,
+    "flatModuleOutFile": "index.js",
+    "flatModuleId": "@angular/material",
+    "skipTemplateCodegen": true
+  }
+}
```
DevVersion: Is that file used anywhere? We now have three `tsconfig-XX` files in the `lib/` package.

jelbourn: I was just trying something out with this, meant to delete it.

tools/package-tools/entry-point-package-json.ts
jelbourn: Short answer: it doesn't.

I updated the script and left a comment explaining the current situation w/ closure.

src/lib/tsconfig-build.json
jelbourn: Was just trying something during development; reverted

jelbourn:
* Rebased such that it includes `material-moment-adapter/`
* Added a `tsconfig.json` per `src/` directory in order to make IDEs auto-add imports to the right path.
* Disabled closure compiler CI check temporarily (need a new release)
* Refactored rollup bundling code into a class, `PackageBundler` so that it can maintain the context of the `BuildPackage`
* Fix prerender

jelbourn:
@DevVersion after this PR we're definitely going to have to parallelize the typescript builds and bundling. With this change it takes over a minute. To paralleize the tasks, we'll need to manually spawn a child process for each, since nodejs is single-threaded. We also probably want to tweak the watch tasks to only re-compile the sub-packages that changed. Let me know if you want to work on all that.



All of this will be obsolete when we switch to bazel, but we're at least a month away from making that switch.

jelbourn:
@DevVersion the tsconfig in the root would cause IDEs to autogenerate the wrong imports sometimes.

jelbourn:
Caretaker note: both the copy config and the build rules need to updated alongside this change

jelbourn:
Another note: if this gets merged alongside another PR, it will potentially need a build fix (for anything importing from, e.g., `../core`)

PR #7001, opened 2017-09-12
feat(selection-model): de/select multiple values at the same time

DevVersion:
Adds support for passing multiple values to the SelectionModel at the same time. This feature is described in the JSDoc already, but just wasn't implemented yet.

This functionality gives us more control about the `onChange` event, which will otherwise fire every time if for example multiple values need to be selected.

Related to #6896

### DIFF
--- a/src/cdk/collections/selection.ts
+++ b/src/cdk/collections/selection.ts
@@ -56,16 +56,18 @@ export class SelectionModel<T> {
   /**
    * Selects a value or an array of values.
    */
-  select(value: T): void {
-    this._markSelected(value);
+  select(...values: T[]): void {
+    this._warnMultipleValuesForSingleSelection(values);
+    values.forEach(value => this._markSelected(value));
     this._emitChangeEvent();
   }
 
   /**
    * Deselects a value or an array of values.
    */
-  deselect(value: T): void {
-    this._unmarkSelected(value);
+  deselect(...values: T[]): void {
+    this._warnMultipleValuesForSingleSelection(values);
+    values.forEach(value => this._unmarkSelected(value));
     this._emitChangeEvent();
   }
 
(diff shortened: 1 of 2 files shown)


src/cdk/collections/selection.ts
```diff
@@ -162,6 +164,13 @@ export class SelectionModel<T> {
       this._selection.forEach(value => this._unmarkSelected(value));
     }
   }
+
+  /** Throws an error if multiple values are passed into a selection model with a single value. */
+  private _warnMultipleValuesForSingleSelection(values: T[]) {
```
crisbeto: I don't think the name is too accurate since it'll throw an error, not warn.

src/cdk/collections/selection.ts
```diff
@@ -171,3 +180,8 @@ export class SelectionModel<T> {
 export class SelectionChange<T> {
   constructor(public added?: T[], public removed?: T[]) { }
 }
+
+/** Throws an error if multiple values are passed into a selection model with a single value. */
+export function throwMultipleValuesInSingleSelectionError() {
+  throw Error('Cannot pass multiple values into SelectionModel with single-value mode.');
```
crisbeto: Since it is only being used in one place, consider moving this inline into `_warnMultipleValuesForSingleSelection`?

DevVersion: I think we have a convention that we create functions for these errors and re-export the Error object. 

e.g in https://github.com/angular/material2/blob/master/src/lib/select/select-errors.ts#L28

<sub>Updating to follow that convention</sub>

src/cdk/collections/selection.ts
DevVersion: Done.

src/cdk/collections/selection.ts
crisbeto: Those errors are in a function so we can use them in tests. If it's a one-off, it's fine to have it inline.

src/cdk/collections/selection.ts
```diff
@@ -182,6 +185,6 @@ export class SelectionChange<T> {
 }
 
 /** Throws an error if multiple values are passed into a selection model with a single value. */
```
crisbeto: Comment is inaccurate now. It returns an error instead of throwing it.

PR #7003, opened 2017-09-12
fix(common-module): check if computed styles are available

### LINKED ISSUE #7000: Firefox fails to load when running inside an iframe with display set to none
The MdCommonModule._checkTheme function throws an error when the angular+material2 app is running inside an iframe that has CSS property display set to 'none'.

This is caused by a long outstanding [bug in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=548397) that seems unlikely to be fixed anytime soon; window.getComputedStyle() returns null inside an iframe with display: none

The behavior of MdCommonModule._checkTheme could be extended by doing a null-check on the result of window.getComputedStyle() and if true follow a more elaborate code path to detect an angular material theme, e.g. by iterating document.stylesheets.cssRules and trying to find the 'mat-theme-loaded-marker' rule. This approach will have an impact on the load time (I think) but as the app is hidden at this point it should not matter that much. 
A simpler change could be to just log an error explaining the Firefox issue, as this should only occur in devMode.

The application bootstraps in the hidden iframe and is shown when the iframe is no longer hidden.

The application fails to bootstrap.

See http://plnkr.co/WQ6BXarpPGQQ7kHoiRSg using Firefox

Although there is a workaround (disable sanityChecks) it seems a bit harsh to not bootstrap the app at all.

Firefox specific.

DevVersion:
Due to a Firefox Bug (https://bugzilla.mozilla.org/show_bug.cgi?id=548397) the `getComputedStyle` call in the `MdCommonModule` will return `null` and cause a runtime exception because normally the computed styles are never `null` for the test element.

Since this it's very uncommon that a developer places a Angular Material application inside of a hidden iframe, a simple null check should be enough to get the application running.

Fixes #7000

### DIFF
--- a/src/lib/core/common-behaviors/common-module.ts
+++ b/src/lib/core/common-behaviors/common-module.ts
@@ -62,6 +62,9 @@ export class MdCommonModule {
 
       const computedStyle = getComputedStyle(testElement);
 
+      // In some situations, the computed style of the test element can be null. For example in
+      // Firefox, the computed style is null if an application is running inside of a hidden iframe.
+      // See: https://bugzilla.mozilla.org/show_bug.cgi?id=548397
       if (computedStyle && computedStyle.display !== 'none') {
         console.warn(
           'Could not find Angular Material core theme. Most Material ' +
--- a/src/lib/core/common-behaviors/common-module.ts
+++ b/src/lib/core/common-behaviors/common-module.ts
@@ -60,7 +60,9 @@ export class MdCommonModule {
       testElement.classList.add('mat-theme-loaded-marker');
       this._document.body.appendChild(testElement);
 
-      if (getComputedStyle(testElement).display !== 'none') {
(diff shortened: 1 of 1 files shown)


src/lib/core/common-behaviors/common-module.ts
```diff
@@ -60,7 +60,9 @@ export class MdCommonModule {
       testElement.classList.add('mat-theme-loaded-marker');
       this._document.body.appendChild(testElement);
 
-      if (getComputedStyle(testElement).display !== 'none') {
+      const computedStyle = getComputedStyle(testElement);
+
+      if (computedStyle && computedStyle.display !== 'none') {
```
crisbeto: Add a comment that it can potentially be null?

DevVersion: Done. Please have another look

Issue #7008, opened 2017-09-12
Expansion Panel Content being swallowed


jt-helsinki:
#### Bug, feature request, or proposal:

Bug

#### What is the expected behavior?

The content of the panel will spill over the right edge of the panel when the screen resolution is too small.

**OR**

Content is scrollable.

#### What is the current behavior?

The panel swallows the content with no horizontal scroll bars. This hides information with no way to retrieve it. 

#### What are the steps to reproduce?

Simply put a table with a 1200px width inside an expansion panel. Shrink the screen horizontally.

See this plunk:

http://plnkr.co/edit/t0qEtVdYlJ6MFCNhVddo?p=preview

#### What is the use-case or motivation for changing an existing behavior?

Placing a table inside an expansion panel. 

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?

Material 2.0.0-beta.10
Angular 4.3.6

josephperrott:
Closed with #7012

PR #7012, opened 2017-09-12
fix(expansion): Handle overflow in .mat-expansion-panel-body

josephperrott:
Addresses: #7008

### DIFF
--- a/src/lib/expansion/expansion-panel.scss
+++ b/src/lib/expansion/expansion-panel.scss
@@ -18,7 +18,8 @@
 }
 
 .mat-expansion-panel-body {
-  padding: 0 24px 16px;
+  margin: 0 24px 16px;
+  overflow: auto;
 }
 
 .mat-expansion-panel-spacing {


tmburnell:
I tried this fix in my overflow situation and it did not fix it ... due to mat-expansion-panel-body actually grow to the size of its contents.  I am still trying to find a fix that works for me, I can post back if i find something.

tmburnell:
http://plnkr.co/edit/O5SE55FZ1qlU1il9fh6v?p=preview

I am using codeMirror to display code in the toggled area.  My example just mocks the layout.
If you expand one it works fine, however if you expand more than one DUE to flexLayout making the expandable smaller the body container does not get smaller so it overflows into the next item.

gigadie:
the fix for this might be changing it from `margin` to `padding`:
```css
.mat-expansion-panel-body {
  margin: 0 24px 16px;
  overflow: auto;
}
```

PR #7020, opened 2017-09-12
demo(list): Add accessibility demo page for list

tinayuangao:
PR #6363 is accidentally changed to change of changelog.md

### DIFF
--- a/src/demo-app/a11y/list/list-a11y.html
+++ b/src/demo-app/a11y/list/list-a11y.html
@@ -2,21 +2,21 @@
   <section>
     <h2> Seasoning </h2>
     <p>Showing a non-interactive list of seasonings.</p>
-    <md-list role="list">
-      <md-list-item *ngFor="let item of items"> {{item}} </md-list-item>
-    </md-list>
+    <mat-list role="list">
+      <mat-list-item *ngFor="let item of items"> {{item}} </mat-list-item>
+    </mat-list>
   </section>
 
 
   <section>
     <h2>Mailbox navigation</h2>
     <p>Showing a navigation list with links to google search</p>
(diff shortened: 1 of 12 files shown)


src/lib/list/list.ts
```diff
@@ -46,35 +46,30 @@ export class MdListDivider {}
...
-export class MdListCssMatStyler {}
-
-/**
- * Directive whose purpose is to add the mat- CSS styling to this selector.
- * @docs-private
- */
-@Directive({
-  selector: 'md-nav-list, mat-nav-list',
-  host: {'class': 'mat-nav-list'}
+  template: '<ng-content></ng-content>',
```
crisbeto: I think that in general we still keep the templates in a separate HTML file, even if it's a one-liner.

src/demo-app/a11y/list/list-a11y.html
```diff
@@ -0,0 +1,66 @@
+<div class="demo-list">
+  <section>
+    <h2> Seasoning </h2>
+    <p>Showing a non-interactive list of seasonings.</p>
+    <md-list role="list">
+      <md-list-item *ngFor="let item of items"> {{item}} </md-list-item>
```
crisbeto: Shouldn't all of these list items have a `role="listitem"`?

tinayuangao: `role="listitem"`  is added in `list.ts` so we don't need to add it here.

jelbourn: We're we going to remove that from the list-item?

kara:
@tinayuangao Rebase?

Issue #7015, opened 2017-09-12
Accessibility md-select aria-labelledby


Lindsor:
#### Bug:
aria-labelledby shows up in DOM even if it has no value

#### What is the expected behavior?
aria-labelledby should never be empty, if there is no labelledby value then it should not be in the DOM

#### What is the current behavior?
aria-labelledby shows up in the DOM even if empty

#### What are the steps to reproduce?
http://embed.plnkr.co/cLXRzvn2uWQ5y5YmiU8L/
inspect select, notice empty aria-labelledby

#### What is the use-case or motivation for changing an existing behavior?
Currently throws accessibility issues

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
"@angular/material": "2.0.0-beta.8"

crisbeto:
Closing since the associated PR was merged.

PR #7022, opened 2017-09-12
fix(select): aria-labelledby should not be in DOM

Lindsor:
When there is no supplied aria-labelledby
it should not show up in the DOM at all

Relates to Issue #7015

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1701,6 +1701,15 @@ describe('MdSelect', () => {
         expect(select.getAttribute('aria-labelledby')).toBe('myLabelId');
       });
 
+      it('should not have aria-labelledby in the DOM if it`s not specified', async(() => {
+
+        fixture.detectChanges();
+
+        fixture.whenStable().then(() => {
+          expect(select.hasAttribute('aria-labelledby')).toBeFalsy();
+        });
+      }));
+
       it('should set the tabindex of the select to 0 by default', () => {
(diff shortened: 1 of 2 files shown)


Lindsor:
Trying to retrigger CLA check

src/lib/select/select.spec.ts
```diff
@@ -1701,6 +1701,12 @@ describe('MdSelect', () => {
         expect(select.getAttribute('aria-labelledby')).toBe('myLabelId');
       });
 
+      it('should not have aria-labelledby in the DOM if it`s not specified', () => {
+        fixture.detectChanges();
+
+        expect(select.getAttribute('aria-labelledby')).toBe(null);
```
crisbeto: You can switch this one to `select.hasAttribute('aria-labelledby')` to make it a bit more clear.

Lindsor: Fixed

Lindsor:
@crisbeto 
I keep getting intermittent test failures. This happened in another angular2 project i worked on, theres probably some async tests not marked as async

crisbeto:
Can you squash your commits into one? Otherwise the merge commit from master might cause some syncing issues.

Lindsor:
@crisbeto Whatsup with these snapshot tests? Is there a way to run locally so i can test it?

crisbeto:
The screenshot tests are somewhat flaky in general. The failure probably isn't because of your changes.

PR #7033, opened 2017-09-12
fix(datepicker): make sure _datepickerInput exists before accessing its

mmalerba:
disabled property.

### DIFF
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -148,7 +148,8 @@ export class MdDatepicker<D> implements OnDestroy {
   /** Whether the datepicker pop-up should be disabled. */
   @Input()
   get disabled() {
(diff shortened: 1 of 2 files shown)


src/lib/datepicker/datepicker.ts
```diff
@@ -148,7 +148,8 @@ export class MdDatepicker<D> implements OnDestroy {
   /** Whether the datepicker pop-up should be disabled. */
   @Input()
   get disabled() {
-    return this._disabled === undefined ? this._datepickerInput.disabled : this._disabled;
+    return this._disabled === undefined && this._datepickerInput ?
+        this._datepickerInput.disabled : this._disabled;
```
jelbourn: Could you write a unit test that captures the situation in which `this._datepickerInput` would be undefined?

mmalerba:
test added, PTAL

PR #7036, opened 2017-09-13
Revert "fix(menu): multiple close events for a single close"

mmalerba:
Reverts angular/material2#6961


mmalerba:
@ppham27 This was causing some unexpected screenshot diffs when I attempted to sync it into Google. I had to revert since it was blocking the sync process. If you create a new PR to undo the revert I will LGTM and investigate the failure before merging again - thanks!

@crisbeto @jelbourn FYI

PR #7037, opened 2017-09-13
fix(menu): multiple close events for a single close

ppham27:
Don't emit a closed event when another event will be emitted. Previously, if one clicked on a menu item, one would get two events: `undefined` and `click` in that order. One would see similar behavior for `keydown` or clicking the backdrop. Unit tests were updated to prevent a regression.

Making a second pull request. See #7036.


mmalerba:
caretaker note: previously saw unexpected screenshot diffs when merging this, make sure global presubmit is ok before merging

PR #7039, opened 2017-09-13
fix(selection-list): tabIndex should respect disabled state

DevVersion:
* Currently if the selection-list is disabled, the tabIndex may be still set to a valid value that allows tabbing to the element. The `mixinTabIndex` respects the disabled state of the selection list.

### DIFF
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -270,6 +270,49 @@ describe('MatSelectionList', () => {
     });
   });
 
+  describe('with tabindex', () => {
+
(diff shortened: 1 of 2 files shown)


DevVersion:
@crisbeto Not too sure about that. The `selection-list` element can gain focus for keyboard interaction, which then allows users to navigate through all options. If the  md-selection-list` is disabled, then it would make sense for me, to disable focus for this time.

andrewseguin:
Needs rebase.

kara:
@DevVersion Rebase one more time?

PR #7055, opened 2017-09-13

crisbeto:


mmalerba:
caretaker note: unexpected screenshot failures, verify global presubmit ok before merging

crisbeto:
Sorry for the email spam everyone, faulty rebase.

PR #7076, opened 2017-09-14
fix(slide-toggle): report change to model before firing a change event

### LINKED ISSUE #7074: MdSlideToggle: change event is triggered before the model's value is updated
Bug (it is the only form control that acts like this and it wasn't the case on beta 8)

When the change event is triggered the model's value should already reflect the new value.

When the change event is triggered, the model value is not updated and still reflects the old value.
This happens with template-driven and reactive forms.

http://plnkr.co/edit/6x5qfWWRwJFb32yr4UX4?p=preview

Material beta 10

DevVersion:
Whenever the value of the slide-toggle changes through interaction, a `change` event will be emitted. The value change then will be reported to the `ControlValueAccessor` which fires the `ngModelChange` output.

Right now the `ngModelChange` output fires after the normal `change` output. This should be the other way around because developers expect the two-way data binding to already reflect the change.

Fixes #7074

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -563,7 +563,8 @@ describe('MdSlideToggle with forms', () => {
       declarations: [
         SlideToggleWithForm,
         SlideToggleWithModel,
-        SlideToggleWithFormControl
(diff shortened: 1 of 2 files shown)


josephperrott:
Also fixes #7300

PR #7090, opened 2017-09-14
style(form-field): use easing function for underline

willshowell:
Tried to match what I see [in this animation](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B8wSqcLwbhFuREg4Q0R4M3VGejA/textfields-presslight-01.webm), but am curious if there are any official curves/durations from the UX team. In either case, I think some easing function is better than linear.

cc @mmalerba and note that no one is CODEOWNER of form-field


mmalerba:
Whoops sorry, missed this PR somehow. I don't think they give any official curves. I just eyed it when I created the original animation, forgot to add the opacity though.

PR #7091, opened 2017-09-14
fix(datepicker): allow ISO 8601 strings as inputs

### LINKED ISSUE #6265: datepicker: Add support for passing IOS8601 strings to datepicker regardless of the DateAdapter

mmalerba:
This is useful for apps that load a bunch of JSON data and want to stuff it into a form

fixes #6265

BREAKING CHANGES:
* `DateAdapter.getISODateString` is now `DateAdapter.toIso8601`
* `DateAdapter` now has a new abstract method `DateAdapter.fromIso8601`

### DIFF
--- a/src/lib/core/datetime/date-adapter.ts
+++ b/src/lib/core/datetime/date-adapter.ts
@@ -164,19 +164,18 @@ export abstract class DateAdapter<D> {
   abstract addCalendarDays(date: D, days: number): D;
 
   /**
-   * Gets the RFC 3339 compatible date string (https://tools.ietf.org/html/rfc3339) for the given
-   * date.
+   * Gets the RFC 3339 compatible string (https://tools.ietf.org/html/rfc3339) for the given date.
    * @param date The date to get the ISO date string for.
    * @returns The ISO date string date string.
    */
-  abstract toISODateString(date: D): string;
+  abstract toIso8601(date: D): string;
 
   /**
-   * Creates a date from an RFC 3339 compatible date string (https://tools.ietf.org/html/rfc3339).
+   * Creates a date from an RFC 3339 compatible string (https://tools.ietf.org/html/rfc3339).
    * @param iso8601String The ISO date string to create a date from
    * @returns The date created from the ISO date string.
    */
-  abstract fromISODateString(iso8601String: string): D | null;
+  abstract fromIso8601(iso8601String: string): D | null;
 
   /**
    * Checks whether the given object is considered a date instance by this DateAdapter.
--- a/src/lib/core/datetime/date-adapter.ts
+++ b/src/lib/core/datetime/date-adapter.ts
@@ -164,12 +164,19 @@ export abstract class DateAdapter<D> {
   abstract addCalendarDays(date: D, days: number): D;
 
   /**
-   * Gets the RFC 3339 compatible date string (https://tools.ietf.org/html/rfc3339)  for the given
+   * Gets the RFC 3339 compatible date string (https://tools.ietf.org/html/rfc3339) for the given
    * date.
    * @param date The date to get the ISO date string for.
    * @returns The ISO date string date string.
    */
-  abstract getISODateString(date: D): string;
+  abstract toISODateString(date: D): string;
+
+  /**
+   * Creates a date from an RFC 3339 compatible date string (https://tools.ietf.org/html/rfc3339).
+   * @param iso8601String The ISO date string to create a date from
+   * @returns The date created from the ISO date string.
+   */
+  abstract fromISODateString(iso8601String: string): D | null;
 
   /**
    * Checks whether the given object is considered a date instance by this DateAdapter.
--- a/src/lib/core/datetime/native-date-adapter.ts
+++ b/src/lib/core/datetime/native-date-adapter.ts
@@ -211,7 +211,7 @@ export class NativeDateAdapter extends DateAdapter<Date> {
         this.getYear(date), this.getMonth(date), this.getDate(date) + days);
   }
 
-  toISODateString(date: Date): string {
+  toIso8601(date: Date): string {
     return [
       date.getUTCFullYear(),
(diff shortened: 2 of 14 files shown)


src/lib/core/datetime/date-adapter.ts
```diff
@@ -164,12 +164,19 @@ export abstract class DateAdapter<D> {
...
 
   /**
-   * Gets the RFC 3339 compatible date string (https://tools.ietf.org/html/rfc3339)  for the given
+   * Gets the RFC 3339 compatible date string (https://tools.ietf.org/html/rfc3339) for the given
    * date.
    * @param date The date to get the ISO date string for.
    * @returns The ISO date string date string.
    */
-  abstract getISODateString(date: D): string;
+  abstract toISODateString(date: D): string;
```
jelbourn: Since you're changing this anyway, `toIsoDateString`?

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -202,14 +211,24 @@ export class NativeDateAdapter extends DateAdapter<Date> {
...
+  toISODateString(date: Date): string {
     return [
       date.getUTCFullYear(),
       this._2digit(date.getUTCMonth() + 1),
       this._2digit(date.getUTCDate())
     ].join('-');
   }
 
+  fromISODateString(iso8601String: string): Date | null {
+    if (iso8601String.match(ISO_8601_REGEX)) {
```
jelbourn: Why do you need to do the check first? (comment?)

src/lib/datepicker/calendar.ts
```diff
@@ -354,4 +366,16 @@ export class MdCalendar<D> implements AfterContentInit, OnDestroy {
         (this._dateAdapter.getMonth(date) >= 7 ? 5 : 12);
     return this._dateAdapter.addCalendarMonths(date, increment);
   }
+
+  /**
+   * Attempts to coerce a property to a date by parsing it as a ISO 8601 string. If not a valid
+   * ISO 8601 string, returns the original vlaue.
+   */
+  private _coerceDateProperty(value: any): any {
```
jelbourn: Should this return `D`?

src/lib/datepicker/calendar.ts
```diff
@@ -63,19 +63,31 @@ export class MdCalendar<D> implements AfterContentInit, OnDestroy {
   private _intlChanges: Subscription;
 
   /** A date representing the period (month or year) to start the calendar in. */
-  @Input() startAt: D;
+  @Input()
+  get startAt(): D { return this._startAt; }
+  set startAt(value: D) { this._startAt = this._coerceDateProperty(value); }
```
jelbourn: Is it weird that the user could do
```ts
picker.startAt = '1985-04-12T23:20:50.52Z';
console.log(picker.startAt); // is a different type
```
?

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -202,14 +211,24 @@ export class NativeDateAdapter extends DateAdapter<Date> {
...
-  getISODateString(date: Date): string {
+  toISODateString(date: Date): string {
     return [
       date.getUTCFullYear(),
       this._2digit(date.getUTCMonth() + 1),
       this._2digit(date.getUTCDate())
     ].join('-');
   }
 
+  fromISODateString(iso8601String: string): Date | null {
```
jelbourn: Why isn't this captured as part of the normal parse function?

src/lib/datepicker/datepicker.ts
```diff
@@ -360,4 +360,16 @@ export class MdDatepicker<D> implements OnDestroy {
         { overlayX: 'end', overlayY: 'bottom' }
       );
   }
+
+  /**
+   * Attempts to coerce a property to a date by parsing it as a ISO 8601 string. If not a valid
+   * ISO 8601 string, returns the original vlaue.
+   */
+  private _coerceDateProperty(value: any): any {
```
jelbourn: Could we avoid repeating this code?

src/lib/core/datetime/native-date-adapter.ts
mmalerba: This is more of a deserialization than parsing user input

src/lib/datepicker/calendar.ts
mmalerba: Kind of? but its the same idea as all of the boolean and number properties that we coerce

src/lib/core/datetime/date-adapter.ts
jelbourn: How opposed would you be to making these `toIso8601` and `fromIso8601`?

src/lib/core/datetime/native-date-adapter.ts
jelbourn: Use either `String.prototype.search` or `RegExp.prototype.test` rather than `match`

mmalerba:
comments addressed, PTAL

PR #7094, opened 2017-09-15
build: compile entry-points in parallel

jelbourn:
image

On my machine this drops `cdk:build:esm` from ~12s to ~10s and drops `material:build:esm` from ~52s to ~23s.

That's still painstakingly slow, but at least slightly more tolerable until we switch to bazel.

Creating the bundles for material still takes ~27s; parallelizing that the same way would help a bit more (for a follow-up PR).


tools/package-tools/build-package.ts
```diff
@@ -1,10 +1,12 @@
...
 import {readFileSync, writeFileSync} from 'fs';
 import {sync as glob} from 'glob';
 import {join} from 'path';
+import {red} from 'chalk';
+import {PackageBundler} from './build-bundles';
 import {buildConfig} from './build-config';
 import {getSecondaryEntryPointsForPackage} from './secondary-entry-points';
-import {PackageBundler} from './build-bundles';
+
+const {spawn} = require('child_process');
```
DevVersion: We should be able to import this with an ES import. e.g `import {spawn} from 'child_process';`

jelbourn:
@DevVersion changed the import, but don't want to trim the output; I want it to pipe everything as-is.

Issue #6937, opened 2017-09-08
Docs for Sidenav needs update


bogacg:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Having Docs that reflect latest updates in project:

Example in Docs for Sidenav being compatible with latest version (2.0.0-beta.10). There is now Drawer Component.
Material 2.0.0-beta.10 produces `.mat-drawer-content` class for content of Sidenav. Drawer component's content seems to handle scrolling case as well, so example does not need to add `overflow: auto;`

#### What is the current behavior?
Having examples of older version.

 `.mat-sidenav-content` class is referred in current Docs example, also scrolling is handled in that example, which is not needed with current version (for content at least, if not sidenav).

willshowell:
Can be closed via #7099

PR #7099, opened 2017-09-15
fix(sidenav-fab-example):removed overflow:auto; since there is no need for it

FadySamirSadek:
#6937

### DIFF
--- a/src/material-examples/sidenav-fab/sidenav-fab-example.css
+++ b/src/material-examples/sidenav-fab/sidenav-fab-example.css
@@ -14,10 +14,6 @@
   overflow: visible;
 }
 
-.example-scrolling-content {
-  overflow: auto;
-}
-
 .example-fab {
   position: absolute;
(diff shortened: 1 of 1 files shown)


src/material-examples/sidenav-fab/sidenav-fab-example.css
```diff
@@ -14,10 +14,6 @@
   overflow: visible;
 }
 
-.example-scrolling-content {
```
mmalerba: This is actually needed because the same class is used to make the sidenav content scroll. Actually can you add `height: 100%` to this instead? That should fix the issue where the FAB button in the content section scrolls up instead of staying fixed:

```css
.example-scrolling-content {
  overflow: auto;
  height: 100%;
}
```

FadySamirSadek: Thank you , I fixed it in the next commit

PR #7106, opened 2017-09-15
fix: add `mat` exportAs and class aliases

jelbourn:


jelbourn:
Caretaker note: when syncing this, we need to remove the transformation that adds the `mat` exportAs aliases and blacklist `mat-exports.ts` from import

src/lib/mat-exports.ts
```diff
@@ -0,0 +1,526 @@
...
+export {MdTooltipModule as MatTooltipModule};
+export {MdYearView as MatYearView};
+export {MdFormFieldModule as MatFormFieldModule};
+export {MdError as MatError};
+export {MdFormField as MatFormField};
+export {MdFormFieldControl as MatFormFieldControl};
+export {MdHint as MatHint};
+export {MdPlaceholder as MatPlaceholder};
+export {MdPrefix as MatPrefix};
+export {MdSuffix as MatSuffix};
```
crisbeto: I really hope this wasn't done manually :P

jelbourn: Nope, Andrew made a script that generated it. We've had this inside Google for a while now but needed 4.4 to come out to make it work externally, so I just copy-pasted the internal file.

PR #7109, opened 2017-09-15
feat(select): add left/right_arrow key navigation

bogdancar:
As suggested by _crisbeto_ in the comments, added left/right_arrow key navigation on select when closed to reflect the native select behavior. This implementation is a temporary workaround to have the functionality. Normally this code belongs to the `list-key-manager`, but there was not implemented and a decision should be made on this specific behavior of left/right_arrow key press for all lists, before being added to `list-key-manager`.


bogdancar:
I signed it!

bogdancar:
Wrong email on commit, not possible to sign CLA. Will open new.

PR #7112, opened 2017-09-15
fix: remove cdk re-exports from @angular/material

jelbourn:
BREAKING CHANGE: `@angular/material` no longer re-exports symbols from
`@angular/cdk`.


src/lib/core/rxjs/rx-chain.spec.ts
```diff
@@ -1,41 +0,0 @@
-import {Observable} from 'rxjs/Observable';
```
crisbeto: It seems like this removes the `rx-chain` tests completely.

jelbourn: They were already copied to `cdk` earlier but not removed from `lib`

jelbourn:
Caretaker note: I have an internal change to go in before this one to prevent it from breaking anyone

jelbourn:
Fixing unit tests, but that shouldn't block presumit

jelbourn:
Unit tests were actually broken in #6822, and I sent a fix in #7117

mmalerba:
needs rebase

jelbourn:
Rebased

PR #7113, opened 2017-09-15

crisbeto:

This is a re-submit of #7055.

### DIFF
--- a/src/cdk/overlay/overlay-ref.ts
+++ b/src/cdk/overlay/overlay-ref.ts
@@ -95,11 +95,15 @@ export class OverlayRef implements PortalHost {
     // pointer events therefore. Depends on the position strategy and the applied pane boundaries.
     this._togglePointerEvents(false);
 
+    if (this._config.positionStrategy && this._config.positionStrategy.detach) {
+      this._config.positionStrategy.detach();
+    }
+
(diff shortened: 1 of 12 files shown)


mmalerba:
@crisbeto  I'm seeing a screenshot test failure where the select's popup winds up misaligned. (I'm assuming due to a window resize in the test and the fact that `RepositionScrollStrategy` no longer listens for those resize events). This seems like it may be a bigger problem than just the select, all of our popup style components are probably relying on this right now.

crisbeto:
I see. I don't think that the resize logic belonged in the scroll strategy in the first place, but I can re-add it in the `overlayRef`.

crisbeto:
@mmalerba I've re-added the resize listener for connected overlays.

andrewseguin:
Needs rebase

kara:
@crisbeto Can you fix the linter error when you get a chance?

crisbeto:
Sorted out the linting issue.

kara:
Gah! Another conflict, sorry @crisbeto

crisbeto:
Rebased.

kara:
@crisbeto Check out the test failures? Probably rebase problem?

crisbeto:
Seems like something that got introduced from master. Fixed.

PR #7115, opened 2017-09-15
feat: add `preserveWhitespaces: false` to all components

jelbourn:
Also adds lint check to enforce its presence

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -90,7 +90,8 @@
     "validate-decorators": [true, {
       "Component": {
         "encapsulation": "\\.None$",
-        "moduleId": "^module\\.id$"
+        "moduleId": "^module\\.id$",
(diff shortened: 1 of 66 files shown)


jelbourn:
Caretaker note: let me know if this causes any screenshot diffs

tslint.json
```diff
@@ -90,7 +90,8 @@
     "validate-decorators": [true, {
       "Component": {
         "encapsulation": "\\.None$",
-        "moduleId": "^module\\.id$"
+        "moduleId": "^module\\.id$",
+        "preserveWhitespaces": "false$"
```
mmalerba: super nit: is there any reason not to make it `"^false$"`?

jelbourn: Nah, but it's really the same either way

PR #7125, opened 2017-09-16
fix(selection-list): restore focus if active item is destroyed

DevVersion:
* Removes the `destroyed` and `onFocus` observables and instead just calls methods on the injected `MdSelectionList` instance.

* Fixes that the active item is not updating if the active item is being destroyed.

### DIFF
--- a/src/lib/list/selection-list.ts
+++ b/src/lib/list/selection-list.ts
@@ -10,7 +10,6 @@ import {FocusableOption, FocusKeyManager} from '@angular/cdk/a11y';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {SelectionModel} from '@angular/cdk/collections';
 import {SPACE} from '@angular/cdk/keycodes';
-import {RxChain, startWith, switchMap} from '@angular/cdk/rxjs';
 import {
   AfterContentInit,
   ChangeDetectionStrategy,
@@ -37,8 +36,6 @@ import {
   mixinDisabled,
   mixinDisableRipple,
 } from '@angular/material/core';
-import {merge} from 'rxjs/observable/merge';
-import {Subscription} from 'rxjs/Subscription';
 
 
 /** @docs-private */
@@ -54,8 +51,6 @@ export interface MatSelectionListOptionEvent {
   option: MatListOption;
 }
 
-const FOCUSED_STYLE: string = 'mat-list-item-focus';
-
 /**
  * Component for list-options of selection-list. Each list-option can automatically
  * generate a checkbox and can put current item into the selectionModel of selection-list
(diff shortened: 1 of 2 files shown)


src/lib/list/selection-list.ts
```diff
@@ -170,11 +157,11 @@ export class MdListOption extends _MdListOptionMixinBase
 
   _handleFocus() {
     this._hasFocus = true;
-    this._renderer.addClass(this._element.nativeElement, FOCUSED_STYLE);
+    this.selectionList._setFocusedOption(this);
   }
 
   _handleBlur() {
```
crisbeto: Since this is just toggling a boolean, it can also be done inline.

src/lib/list/selection-list.ts
```diff
@@ -334,4 +294,9 @@ export class MdSelectionList extends _MdSelectionListMixinBase
   private _isValidIndex(index: number): boolean {
     return index >= 0 && index < this.options.length;
   }
+
+  /** Returns the index of the specified list option. */
+  private _getIndexFromOption(option: MdListOption): number {
```
crisbeto: Call this something like `_getOptionIndex`?

src/lib/list/selection-list.ts
```diff
@@ -267,36 +240,23 @@ export class MdSelectionList extends _MdSelectionListMixinBase
...
-      let optionIndex: number = this.options.toArray().indexOf(e.option);
-      this._keyManager.updateActiveItemIndex(optionIndex);
-    });
+  /** Removes an option from the selection list and updates the active item. */
+  _removeOptionFromList(option: MdListOption) {
+    if (option._hasFocus) {
+      const optionIndex = this._getIndexFromOption(option);
+
+      // Check whether the option is the last item
+      if (optionIndex - 1 >= 0) {
```
crisbeto: `if (optionIndex > 0)` should be a bit more readable.

src/lib/list/selection-list.ts
```diff
@@ -267,36 +240,23 @@ export class MdSelectionList extends _MdSelectionListMixinBase
...
+        this._keyManager.setPreviousItemActive();
+      } else if (optionIndex < this.options.length - 1) {
```
crisbeto: I don't see any case, aside from `optionIndex === 0`, where it would get into this block. You can simplify it to check for `optionIndex === 0` instead.

src/lib/list/selection-list.ts
DevVersion: Sounds also good

src/lib/list/selection-list.ts
DevVersion: Yeah that's better. Just took the old logic here.

src/lib/list/selection-list.ts
DevVersion: Sure

src/lib/list/selection-list.ts
DevVersion: Done. Addressed your feedback.

src/lib/list/selection-list.ts
```diff
@@ -242,18 +238,18 @@ export class MdSelectionList extends _MdSelectionListMixinBase
...
 
   /** Removes an option from the selection list and updates the active item. */
   _removeOptionFromList(option: MdListOption) {
     if (option._hasFocus) {
-      const optionIndex = this._getIndexFromOption(option);
+      const optionIndex = this._getOptionIndex(option);
 
       // Check whether the option is the last item
-      if (optionIndex - 1 >= 0) {
+      if (optionIndex  > 0) {
```
crisbeto: There's an extra space before the `>`.

PR #7129, opened 2017-09-17
fix(drawer): drawer container animating when open by default

### LINKED ISSUE #7007: bug(sidenav): the drawer-container is animated when initialized
Bug

When initializing a `md-sidenav-container`, the container should not be animated.

When initializing a `md-sidenav-container`, the container is animated. 
The `md-sidenav` (alias `md-drawer`) is initialized, the `open-instant` [animation](https://github.com/angular/material2/blob/master/src/lib/sidenav/drawer.ts#L239) state is used, opening the sidenav [without animation](https://github.com/angular/material2/blob/master/src/lib/sidenav/drawer.ts#L71).
However there is no such thing on the `md-drawer-container`: the [transition](https://github.com/angular/material2/blob/master/src/lib/sidenav/drawer-transitions.scss) is [always](https://github.com/angular/material2/blob/master/src/lib/sidenav/drawer.ts#L391) set.

<sub>No time for a plunkr yet but I will provide it later if necessary</sub>

Since the drawer is not animating, it makes no sense to animate the container.
<sub>And since the drawer is being initialized, `(open)/(close)` events are not triggered</sub>

beta-10

crisbeto:
* Prevents the drawer container from animating if it's open on load.
* Fixes `_updateStyles` being called twice in a row on init, because the first animations are `void->void` and `void->open-instant`.

Fixes #7007.

### DIFF
--- a/src/lib/sidenav/drawer.ts
+++ b/src/lib/sidenav/drawer.ts
@@ -33,10 +33,8 @@ import {
 } from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
 import {merge} from 'rxjs/observable/merge';
-import {first} from 'rxjs/operator/first';
-import {startWith} from 'rxjs/operator/startWith';
-import {takeUntil} from 'rxjs/operator/takeUntil';
 import {Subject} from 'rxjs/Subject';
+import {RxChain, filter, first, startWith, takeUntil} from '@angular/cdk/rxjs';
 
 
 /** Throws an exception when two MatDrawer are matching the same position. */
@@ -117,7 +115,7 @@ export class MatDrawerContent implements AfterContentInit {
   host: {
     'class': 'mat-drawer',
     '[@transform]': '_animationState',
-    '(@transform.start)': '_onAnimationStart()',
+    '(@transform.start)': '_onAnimationStart($event)',
     '(@transform.done)': '_onAnimationEnd($event)',
     '(keydown)': 'handleKeydown($event)',
     // must prevent the browser from aligning text based on value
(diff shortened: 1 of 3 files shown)


src/lib/sidenav/drawer.ts
```diff
@@ -385,13 +386,19 @@ export class MdDrawerContainer implements AfterContentInit, OnDestroy {
...
-      this._updateStyles();
-      this._changeDetectorRef.markForCheck();
-    });
+    RxChain.from(drawer._animationStarted)
+      .call(takeUntil, this._drawers.changes)
+      .call(filter, (event: AnimationEvent) => event.fromState !== event.toState)
+      .subscribe((event: AnimationEvent) => {
+        // Set the transition class on the container so that the animations occur. This should not
+        // be set initially because animations should only be triggered via a change in state.
+        if (event.toState !== 'open-instant') {
```
mmalerba:
```ts
this._element.nativeElement.classList.toggle('mat-drawer-transition', event.toState != 'open-instant');
```

crisbeto: The second param isn’t supported on IE.

mmalerba: Ah, in that case can we still at least use `classList.add` instead of the `Renderer`

crisbeto: The renderer is SSR-friendly. AFAIK we can stop using it in Angular 5.

mmalerba: Don't animation events not fire on the server?

crisbeto: AFAIK they could fire if the sidenav starts off opened.

mmalerba: Hmm ok, just needs rebase then

kara:
@crisbeto Fix linter failure?

crisbeto:
Ready to go.

PR #7131, opened 2017-09-17
fix(radio): defaulting to invalid name attribute

crisbeto:
Fixes the radio button defaulting to `name="undefined"` for the underlying input, if the consumer hasn't passed in a name.

Relates to #7130.

### DIFF
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -638,6 +638,12 @@ describe('MatRadio', () => {
         expect(document.activeElement).toBe(fruitRadioNativeInputs[i]);
       }
     });
+
+    it('should not add the "name" attribute if it is not passed in', () => {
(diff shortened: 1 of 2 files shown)


src/lib/radio/radio.spec.ts
```diff
@@ -7,7 +7,7 @@ import {dispatchFakeEvent, FakeViewportRuler} from '@angular/cdk/testing';
 import {RIPPLE_FADE_IN_DURATION, RIPPLE_FADE_OUT_DURATION} from '@angular/material/core';
 import {MdRadioButton, MdRadioChange, MdRadioGroup, MdRadioModule} from './index';
 
-describe('MdRadio', () => {
+fdescribe('MdRadio', () => {
```
DevVersion: The `describe` has been added accidentally here.

crisbeto: Ops, good catch.

Issue #6677, opened 2017-08-28
Make all @Output names, property names, and event class names consistent


jelbourn:

amcdnl:
Here is my initial audit and proposals before I make any changes.

- Change names to past tense `select` to `selected`
- Remove prefixes from event names e.g. `onSelect` - [Angular Style Guide](https://angular.io/guide/styleguide#dont-prefix-output-properties)

## Chip
- `select`: Change to `selected`. [Select](https://developer.mozilla.org/en-US/docs/Web/Events/select) is a native event emitted when text is selected by the cursor.
- `deselect` => `deselected`.
- `destroy` => `destroyed`
- `onRemove` => `removed`

## Date Picker
- `selectedChanged` - @jelbourn any reason why this wouldn't just be `changed`?
- `userSelection` => `userSelected`
- `dateChange` => `dateChanged`
- `selectedValueChange` => `changed`?

## Menu
- `close` => `closed`
- `onMenuOpen` => `menuOpened` or `opened`
- `onMenuClose` => `menuClosed` or `closed`

## Paginator
- `page` => `paged`

## Select
- `onOpen` => `opened`
- `onClose` => `closed`
- `valueChange` => `valueChanged`

## Sidenav
- `onOpen` => `opened`
- `onClose` => `closed`
- `onPositionChanged` => `positionChanged`
- `onAlignChanged` => `alignChanged`

## Sort
- `mdSortChange` => `mdSortChanged`

## Tabs
- `selectedIndexChange` => `selectedIndexChanged`
- `focusChange` => `focusChanged`
- `selectChange` => `selectChanged`
- `onCentering` => `centering`
- `onCentered` => `centered`

@jelbourn - do we want to convert `change` to `changed`? Also there is several places that have changed prefixed like `mdSortChange`, do we want to consolidate those to `change` or would we like to go the other way and prefix everything?

jelbourn:
* Are you going to do a separate pass for event class names? (e.g., radio buttons emit `MdRadioChange` but `mdSort` emits just `Sort`).
* Are there any places where the past-tense event name conflicts with another property or method in the class? I recall that being the reason for some members being `onSomething`
* Most of the changes seem fine, but some are off
  * The datepicker ones seem weird, cc @mmalerba. `userSelected` is definitely not right.
[...]

PR #7133, opened 2017-09-17
fix(menu): make @Output names consistent #6677

amcdnl:
Updates the Menu API with the following API modifications per #6677

- `close` to `closed`
- `hover` to `hovered`
- `onMenuOpen` to `menuOpened`
- `onMenuClose` to `menuClosed`


src/lib/menu/menu-trigger.ts
```diff
@@ -283,7 +305,14 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
   // set state rather than toggle to support triggers sharing a menu
   private _setIsMenuOpen(isOpen: boolean): void {
     this._menuOpen = isOpen;
-    this._menuOpen ? this.onMenuOpen.emit() : this.onMenuClose.emit();
+
+    if (this._menuOpen) {
+      this.onMenuOpen.emit();
```
crisbeto: Rather than having to emit to multiple subjects, why not have something like:

```
@Output() onMenuOpen = new EventEmitter<void>();
@Output() onMenuOpened = this.onMenuOpen;
```

amcdnl: Thanks for the feedback! Done :)

src/lib/menu/menu-directive.ts
```diff
@@ -139,19 +139,30 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
...
 
   /** Event emitted when the menu is closed. */
-  @Output() close = new EventEmitter<void | 'click' | 'keydown'>();
+  @Output() closed = new EventEmitter<void | 'click' | 'keydown'>();
+
+  /**
+   * Event emitted when the menu is closed.
+   * @deprecated Switch to `closed` instead
+   */
+  @Output() close = this.closed;
```
jelbourn: Same question here as the select- should we just have one `openedChange`?
cc @kara @crisbeto

src/lib/menu/menu-trigger.ts
```diff
@@ -203,7 +216,12 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
   openMenu(): void {
     if (!this._menuOpen) {
       this._createOverlay().attach(this._portal);
-      this._closeSubscription = this._menuClosingActions().subscribe(() => this.menu.close.emit());
+      this._closeSubscription = this._menuClosingActions().subscribe(() => {
+        if (this.menu.close) {
+          this.menu.close.emit();
+        }
+        this.menu.closed.emit();
```
jelbourn: Do you need this any more?

src/lib/menu/menu-trigger.ts
```diff
@@ -218,7 +236,11 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
       this._resetMenu();
       this._overlayRef.detach();
       this._closeSubscription.unsubscribe();
-      this.menu.close.emit();
+      if (this.menu.close) {
+        this.menu.close.emit();
+      }
```
jelbourn: And this?

src/lib/menu/menu-directive.ts
crisbeto: That's fine by me, just need to mark it as a breaking change or deprecated.

src/lib/menu/menu-trigger.ts
```diff
@@ -203,7 +216,12 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
   openMenu(): void {
     if (!this._menuOpen) {
       this._createOverlay().attach(this._portal);
-      this._closeSubscription = this._menuClosingActions().subscribe(() => this.menu.close.emit());
+      this._closeSubscription = this._menuClosingActions().subscribe(() => {
+        if (this.menu.close) {
```
crisbeto: Why do we need this null check? The `close` emitter should always be there.

src/lib/menu/menu-trigger.ts
```diff
@@ -218,7 +236,11 @@ export class MdMenuTrigger implements AfterViewInit, OnDestroy {
       this._resetMenu();
       this._overlayRef.detach();
       this._closeSubscription.unsubscribe();
-      this.menu.close.emit();
+      if (this.menu.close) {
```
crisbeto: We shouldn't need this null check either.

src/lib/menu/menu-trigger.ts
amcdnl: On the interface `MdMenuPanel` it declared a `close` EventEmitter. As part of this deprecation, I made that an optional interface and made `closed` the new interface. The implementation of this required I add a null check in those spots. If we are ok with removing that from the interface, we can remove all of this.

src/lib/menu/menu-trigger.ts
amcdnl: See above comment.

src/lib/menu/menu-directive.ts
```diff
@@ -139,19 +139,30 @@ export class MdMenu implements AfterContentInit, MdMenuPanel, OnDestroy {
...
     private _elementRef: ElementRef,
     @Inject(MD_MENU_DEFAULT_OPTIONS) private _defaultOptions: MdMenuDefaultOptions) { }
 
-  ngAfterContentInit() {
+  ngAfterContentInit(): void {
     this._keyManager = new FocusKeyManager<MdMenuItem>(this.items).withWrap();
-    this._tabSubscription = this._keyManager.tabOut.subscribe(() => this.close.emit('keydown'));
+    this._tabSubscription = this._keyManager.tabOut.subscribe(() => {
+      this.close.emit('keydown');
+      this.closed.emit('keydown');
```
crisbeto: You shouldn't emit to both of these anymore. Since they're pointing to the same emitter, you'll end up emitting the same event twice.

src/lib/menu/menu-directive.ts
amcdnl: @jelbourn - Ok, to verify we want to change`open` and `close` to `openChange` with a signature like:

```
export interface MdMenuEvent {
   opened: boolean;
   type: void | 'click' | 'keydown'
}
```

amcdnl:
Updated minus the new api question above.

jelbourn:
Also just noticed- @kara @crisbeto why is there a `hover` event? Is that meant to be internal only? I would think downstream users would use `mouseover` etc. like normal

crisbeto:
The hover event is internal, I added it to accommodate the nested menus.

jelbourn:
Ack, @amcdnl can you change `hover` to just be internal?

amcdnl:
@jelbourn - done.

jelbourn:
Caretaker note: should be completely backwards compatible, so if something fails there's a bug

amcdnl:
Closing of favor of #8053

PR #7134, opened 2017-09-17
fix(tabs): update tab output names 

amcdnl:
Updates the tabs to the following new apis per #6677

- `selectChange` to `selectedTabChange`
- `onCentering` to `_onCentering`
- `onCentered` to `_onCentered`


src/lib/tabs/tab-group.ts
```diff
@@ -134,9 +134,16 @@ export class MdTabGroup extends _MdTabGroupMixinBase implements AfterContentInit
...
 
-  /** Event emitted when the tab selection has changed. */
+  /**
+   * Event emitted when the tab selection has changed.
+   * @deprecated Use `selectedTabChange` instead.
+   */
   @Output() selectChange: EventEmitter<MdTabChangeEvent> = new EventEmitter<MdTabChangeEvent>(true);
 
+  /** Event emitted when the tab selection has changed. */
+  @Output() selectedTabChange: EventEmitter<MdTabChangeEvent> =
```
josephperrott: Rather than duplicating the creation of the EventEmitter and then having to call both, why not just have `selectedTabChange` get the `selectChange` value currently.

```ts

/**
 * Event emitted when the tab selection has changed.
 * @deprecated Use `selectedTabChange` instead.
 */
@Output() get selectChange(): EventEmitter<MdTabChangeEvent> {
  return this.selectedTabChange;
}

/** Event emitted when the tab selection has changed. */
@Output() get selectedTabChange(): EventEmitter<MdTabChangeEvent> = 
  new EventEmitter<MdTabChangeEvent>(true);
```

Doing it this way, allows for users to move to the new output, and ultimately allows for the removal of the `@Output` with minimal change to the code in the future as no change is needed below in the file.

amcdnl: Thanks for the suggestion. Updated!

mmalerba:
@amcdnl your closing `mat-tab-body` tag looks like it got turned into a `md-tab-body` tag somehow, mind fixing so I can presubmit?

amcdnl:
@mmalerba - done :)

amcdnl:
@andrewseguin - rebased.

PR #7135, opened 2017-09-17
chore(datepicker): make userSelection output internal

amcdnl:
Updates the datepicker's calendar api per #6677.

- `userSelection` to `_userSelection`. Did not deprecate since its internal only.
- Did not update `selectedChange` to `_selectedChange` because its used for 2-way binding on the `selected` property of the calendar.


mmalerba:
please rebase

amcdnl:
@mmalerba - rebased.

PR #7136, opened 2017-09-17
style(snackbar): remove border radius from center-aligned positions

willshowell:


willshowell:
This was to fix a minor nit of mine, but it looks like the spec is inconsistent.

Shows bottom border radius:
https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B57Y8yNEHsZPVkRWSkNRLVNzQXc/components-toasts-tabletdesktop2.png

Shows no border radii at all:
https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B6Okdz75tqQsLVVnZlF4UEtKRU0/components-snackbar-specs-fabtablet-002.webm

@jelbourn feel free to close if not needed.

josephperrott:
I am not sure why the animation they show has no border radius.  However for tablet/desktop the corners are specifically defined: `2dp rounded corner` so I think we will leave it as 2px for tablet/desktop.

The mobile version does not have this, however we do not yet do the mobile styling as we are still pending our `BreakpointObserver`

willshowell:
@josephperrott Yeah I'm certainly in agreement with the borders having a 2px radius. My proposal is specifically about the corners that align with the viewport edge:

snackbar-border-radius

To me, it's mildly annoying, but easy enough to override myself. I would just like to make sure you didn't think I was proposing to remove all the border radii. 😄

josephperrott:
You are correct, I did misunderstand what you were looking to do, sorry about that.  For now we will leave it as is in the library, though I reached out to some people to determine the official guidance.

willshowell:
Thanks!

willshowell:
@josephperrott any response?

jelbourn:
Response from Material UX: they should always have rounded corners, the place where they don't is a mistake.

willshowell:
Bizarre - it seems like a given to me that you'd want alignment against the viewport, but ack 👍

PR #7146, opened 2017-09-18
build(): retries for travis deploying

DevVersion:
* Now re-runs the given deploy mode a second time if failed the first time. This ensures that deployment works fine most of the time and should also avoid conflicts between different Travis CI jobs that are deplyoing at the same time.


jelbourn:
@DevVersion needs rebase

Issue #5881, opened 2017-07-19
Add public api to autocomplete to check whether it is open


jelbourn:
There is no public api users can consume that tells them whether an autocomplete is open on `MdAutocomplete`. We should add one.

leocaseiro:
Hi @jelbourn, I'd like to work on that feature.

We have some emitters for the select component such as `onOpen` and `onClose`.
I wonder if that would be the best approach or should we export public the property `showPanel`?

Thanks

jelbourn:
@leocaseiro I was just imagining an `isOpen` property on `MdAutocomplete`

carlos-:
This can already be done with the `panelOpen` property on `MdAutocompleteTrigger`. I updated [this Plunker](https://plnkr.co/edit/gPp7voxywnsGjVQC8rXs?p=preview) to prove myself it works.

@jelbourn, do you want to offer it directly in the MdAutocomplete, or should this be fine?

willshowell:
@carlos- Can't speak for the original intent, but I think this would be nice

```html
<md-autocomplete #auto="mdAutocomplete">
  <md-option *ngFor="let state of filteredStates | async" [value]="state">
    {{ state }}
  </md-option>
</md-autocomplete>

isOpen: {{auto.isOpen}}
```

since it would eliminate the need for this:

```ts
@ViewChild(MdAutocompleteTrigger) trigger: MdAutocompleteTrigger;

isOpen() {
  return this.trigger.panelOpen
}
```

carlos-:
@willshowell, I totally agree. On the other hand, the same could apply to `openPanel` and `closePanel`. Right?

kkotak:
@carlos- Thanks for the tip on using MdAutocompleteTrigger. I am using it successfully for one Autocomplete on the page. How can I use it with multiple Autocompletes on the same page?

josephperrott:
Fixed by #7149

PR #7149, opened 2017-09-18
feat(autocomplete): add isOpen attribute

josephperrott:
For #5881

### DIFF
--- a/src/demo-app/autocomplete/autocomplete-demo.html
+++ b/src/demo-app/autocomplete/autocomplete-demo.html
@@ -66,19 +66,6 @@ Space above cards: <input type="number" [formControl]="topHeightCtrl">
   </mat-card>
 </div>
 
-<mat-autocomplete #reactiveAuto="matAutocomplete" [displayWith]="displayFn">
-  <mat-option *ngFor="let state of reactiveStates | async" [value]="state">
-    <span>{{ state.name }}</span>
-    <span class="demo-secondary-text"> ({{state.code}}) </span>
-  </mat-option>
-</mat-autocomplete>
-
-<mat-autocomplete #tdAuto="matAutocomplete">
-  <mat-option *ngFor="let state of tdStates" [value]="state.name">
-    <span>{{ state.name }}</span>
-  </mat-option>
-</mat-autocomplete>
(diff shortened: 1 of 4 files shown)


src/lib/autocomplete/autocomplete.ts
```diff
@@ -57,6 +57,11 @@ export class MdAutocomplete implements AfterContentInit {
   /** Whether the autocomplete panel should be visible, depending on option length. */
   showPanel = false;
 
+  _isOpen: boolean = false;
```
crisbeto: I'm not sure we need this extra flag. `isOpen` should the same as `this.showPanel && this.options.length > 0`.

josephperrott: That was my initial thought as well, however, `showPanel` is not set to false on `blur`, which prevents it from working that way.

src/demo-app/autocomplete/autocomplete-demo.html
```diff
@@ -66,19 +66,6 @@
   </md-card>
 </div>
 
-<md-autocomplete #reactiveAuto="mdAutocomplete" [displayWith]="displayFn">
```
crisbeto: Why were these removed from the demo?

src/lib/autocomplete/autocomplete.ts
crisbeto: Makes sense.

src/demo-app/autocomplete/autocomplete-demo.html
josephperrott: They are duplicates of the same elements above.

It look slike we had 5 `<md-autocomplete>`s for 3 `[mdAutocompleteTrigger]`s

src/lib/autocomplete/autocomplete.ts
crisbeto: Cool, also a final nit: can you add a doc string to the `isOpen` getter so it shows up in the docs?

josephperrott: Yeah, not sure how I deleted the doc line I had for it.

PR #7160, opened 2017-09-19
fix(select): make @Output names consistent #6677

amcdnl:
Updated APIs per #6677 

- `onOpen` to `opened`
- `onClose` to `closed`
- `change` to `selectionChange`


src/lib/select/select.ts
```diff
@@ -402,13 +402,31 @@ export class MdSelect extends _MdSelectMixinBase implements AfterContentInit, On
   }
 
   /** Event emitted when the select has been opened. */
-  @Output() onOpen: EventEmitter<void> = new EventEmitter<void>();
+  @Output() opened: EventEmitter<void> = new EventEmitter<void>();
```
jelbourn: Should we make this consistent with sidenav and just have one `openedChange`?
Thoughts @kara @crisbeto?

crisbeto: That's fine by me, we just have to make sure to mark it as a breaking change.

amcdnl:
@jelbourn - let me know what you want to do regarding that.

jelbourn:
If we're making the sidenav `openChange` then we should do that here too

amcdnl:
@jelbourn - should be good to go now.

jelbourn:
Caretaker notE: should be completely backwards compatible, so if something fails there's a bug

amcdnl:
closing in favor of #8052

PR #7161, opened 2017-09-19
chore(sidenav): update output names per #6677

amcdnl:
Updates APIs per #6677 

- `open` | `onOpen` to `openChange`
- `close` | `onClose` to `closeChange`

@jelbourn - `onPositionChanged` is a member declared with `positionChanged` output name. We might consider changing that member name to match the output name in this as well.


jelbourn:
We should just eliminate `closeChange` and have `openChange` include the state.

We should remove `on` from all of the property names as well.

src/lib/sidenav/drawer.ts
```diff
@@ -145,10 +145,22 @@ export class MdDrawer implements AfterContentInit, OnDestroy {
   private _currentTogglePromise: Promise<MdDrawerToggleResult> | null;
 
   /** Event emitted when the drawer is fully opened. */
-  @Output('open') onOpen = new EventEmitter<MdDrawerToggleResult | void>();
+  @Output() openChange = new EventEmitter<MdDrawerToggleResult | void>();
```
mmalerba: I think this is really confusing now. I would expect `openChange` to fire when the drawer closes, since that is a change in the open state of the drawer. If we're going with this naming style we should get rid of the `closeChange` one and just emit booleans on the `openChange` stream whenever the drawer opens or closes.

We should also make it match the input (either both `open` or both `opened`) so that people can do 2-way binding

amcdnl:
@jelbourn - I've removed `closeChange` and updated usages to `openChange`. `openChange` already emitted a `type` which defined if it was `closed` or `opened`. This would feel more natural to me if it were a boolean, what do you guys think?

@mmalerba - The input for open is `opened`, so `openChange` wouldn't match the pattern `{input}Change` to do the 2-way binding without overriding it (which its already doing) or changing `openChange` to `openedChange`.

mmalerba:
+1 for changing emitted type to boolean.

I think the `@Input` was called `opened` in order to not conflict with the `open()` method. We could do something like this:

```ts
@Input('open') isOpen = false;
@Output() openChange = new EventEmitter<boolean>();
open() { ... }
```

or

```ts
@Input() opened = false;
@Output() openedChange = new EventEmitter<boolean>();
open() { ... }
```

I'm fine with either

jelbourn:
I generally prefer emitting event objects instead of primitives because we can always add more information to them later without it being a breaking change. So it would emit something like
`MdSidenavOpenChangeEvent`

I'd rather avoid doing `isOpen` since it would be inconsistent with the rest of the boolean properties around the library (and that inconsistency s what we're addressing in the first place).

mmalerba:
@jelbourn Agreed about `isOpen` being inconsistent. That's why in my example I renamed it to `open` using the string argument that `@Input()` can take. (I couldn't directly call it `open` because it would conflict with the `open()` method).

I also agree about emitting events in most cases, but not when there's a corresponding `@Input()`. Having both `opened` and `openedChange` gives the impression that you can do `[(opened)]="isOpen"`, but this wouldn't work right if `openedChange` emitted objects instead of booleans.

mmalerba:
@jelbourn and I talked offline and agreed on the following:
```ts
@Input() opened: boolean;

@Output() openedChange: EventEmitter<boolean>;

@Output('opened') get _openedStream(): EventEmitter<void> { return filter(this.openedChange, o => o); }

@Output('closed') get _closedStream(): EventEmitter<void> { return filter(this.openedChange, o => !o); }

open() {...}
```

amcdnl:
Given all the changes, I'm closing this and starting fresh w/ #7450

PR #7164, opened 2017-09-19
chore(snack-bar): rename MdSnackBarRef `instance` to `componentInstance`

### LINKED ISSUE #6252: SnackBarRef component instance ?

Feature request

Having a standard for the instance reference and sharing data.

To get the component instance of a SnackBarRef, the attribute is "instance".
To get the component instance of a DialogRef, the attribute is "componentInstance". Besides, in the [documentation](https://material.angular.io/components/dialog/api) we can see a method to share data directly has been implemented, it's not the case for SnackBarRef.

I think having a standard for these components "with reference" would be clearer for the user.

Latest of each.

shaunak1111:
BREAKING CHANGE: Renamed variable instance to componentInstance to be consistent with other dialogs

Closes #6252

### DIFF
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -16,7 +16,7 @@ import {MdSnackBarContainer} from './snack-bar-container';
  */
 export class MdSnackBarRef<T> {
   /** The instance of the component making up the content of the snack bar. */
-  instance: T;
+  componentInstance: T;
 
   /**
    * The instance of the component making up the content of the snack bar.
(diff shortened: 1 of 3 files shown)


kara:
Note: internal updates required

josephperrott:
Since this didn't make it in before RC, we will need to mark `instance` as deprecated have it still available to be used as a reference for the time being.

You should be able to just have a getter so the value is available:

```
/**
 * @docs-private
 * @deprecated
 */
get instance() { return this.componentInstance; }
```

jelbourn:
Closing this since, after releasing 5.0.0, we aren't making any more breaking API changes.

PR #7174, opened 2017-09-19
build: round payload size for comparison

DevVersion:
In some cases the payload is still showing up in the CI even if there was no change to the library. This seems to be because the file sizes are not rounded and can sometimes be not very exact. Resulting in numbers like: `material_fesm_2014": 1004.1319999999998`.

To have a more readable output in the gulp task, the file sizes should be rounded. This also makes the diff comparsion in the Firebase function better.

See PRs like: https://github.com/angular/material2/pull/7169

### DIFF
--- a/tools/gulp/tasks/payload.ts
+++ b/tools/gulp/tasks/payload.ts
@@ -93,10 +93,10 @@ async function calculatePayloadDiff(database: firebaseAdmin.database.Database, c
 
   // Calculate the payload diffs by subtracting the previous size of the FESM ES2015 bundles.
   const cdkFullSize = currentPayload.cdk_fesm_2015;
-  const cdkDiff = cdkFullSize - previousPayload.cdk_fesm_2015;
+  const cdkDiff = roundFileSize(cdkFullSize - previousPayload.cdk_fesm_2015);
 
   const materialFullSize = currentPayload.material_fesm_2015;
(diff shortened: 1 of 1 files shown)


tools/gulp/tasks/payload.ts
```diff
@@ -93,10 +93,10 @@ async function calculatePayloadDiff(database: firebaseAdmin.database.Database, c
 
   // Calculate the payload diffs by subtracting the previous size of the FESM ES2015 bundles.
   const cdkFullSize = currentPayload.cdk_fesm_2015;
-  const cdkDiff = cdkFullSize - previousPayload.cdk_fesm_2015;
+  const cdkDiff = roundFileSize(cdkFullSize - previousPayload.cdk_fesm_2015);
```
jelbourn: Why not just use `toFixed`?

DevVersion: It returns a string instead of a number. I don't think re-parsing is better.

jelbourn: Generally you want to keep your values in number format as long as possible and them format them just before or past of displaying them.

DevVersion: Yeah that's what this does now. `toFixed()` returns a string instead of a number. The formatting will happen in the Firebase function.

PR #7176, opened 2017-09-19
feat(autocomplete): add md-autocomplete classes to overlay panel

### LINKED ISSUE #4196: Feature request for autocomplete - more control over panel width

crisbeto:
Transfers any classes added to `md-autocomplete` to the resulting panel, allowing for easier styling similarly to `md-menu`.

Fixes #4196.

### DIFF
--- a/src/lib/autocomplete/autocomplete.html
+++ b/src/lib/autocomplete/autocomplete.html
@@ -1,5 +1,5 @@
 <ng-template>
-  <div class="mat-autocomplete-panel" role="listbox" [id]="id" [ngClass]="_getClassList()" #panel>
+  <div class="mat-autocomplete-panel" role="listbox" [id]="id" [ngClass]="_classList" #panel>
     <ng-content></ng-content>
   </div>
 </ng-template>
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -1482,6 +1482,25 @@ describe('MatAutocomplete', () => {
         expect(placeholder.classList).not.toContain('mat-form-field-empty');
       }));
 
+    it('should transfer the mat-autocomplete classes to the panel element', fakeAsync(() => {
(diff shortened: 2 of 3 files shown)


willshowell:
Related https://github.com/angular/material2/pull/6618. And would it be possible to support NgClass syntax?

crisbeto:
I would rather keep it simple and in line with `md-menu`. It can be added later if people ask for it.

willshowell:
Ok cool! Also does it support removal of classes? It looks like you can only add classes, though I haven't tested.

crisbeto:
It doesn't, but the intention here is just to allow styles to be overridden. My worry with making it more dynamic is that people may end up putting it in a state where it interferes with the panel being toggled correctly.

andrewseguin:
Merging this in but it includes an `md-` in the comment, can you send another PR that removes `md-autocomplete` and changes it to `autocomplete` or `mat-autocomplete`?

PR #7028, opened 2017-09-12
fix(tooltip): error on trigger escape presses while closed

crisbeto:
Fixes an error being thrown when pressing escape on a tooltip trigger that isn't showing a tooltip.

Fixes #7009.

PR #7179, opened 2017-09-19
fix(tooltip): only check tooltip visibility on keydown when it exists

jelbourn:


crisbeto:
Dupe of https://github.com/angular/material2/pull/7028.

PR #7188, opened 2017-09-19
chore: Make accessibility demo page focus goes to header after clicki…

tinayuangao:
…ng on links

### DIFF
--- a/src/demo-app/a11y/a11y.html
+++ b/src/demo-app/a11y/a11y.html
@@ -2,11 +2,12 @@
 
 <nav>
   <a (click)="skipNavigation()" href="javascript:void(0)">Skip navigation</a>
-  <a *ngFor="let navItem of navItems" [routerLink]="[navItem.route]">
+  <a *ngFor="let navItem of navItems" [routerLink]="[navItem.route]" (click)="skipNavigation()">
     {{navItem.name}}
   </a>
 </nav>
 
-<div #maincontent tabindex="0">
+<div #maincontent tabindex="-1">
+  <h1 #header tabindex="-1" *ngIf="!!currentComponent">{{currentComponent}} Demo</h1>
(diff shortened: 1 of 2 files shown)


src/demo-app/a11y/a11y.html
```diff
@@ -2,11 +2,12 @@
...
 <nav>
   <a (click)="skipNavigation()" href="javascript:void(0)">Skip navigation</a>
-  <a *ngFor="let navItem of navItems" [routerLink]="[navItem.route]">
+  <a *ngFor="let navItem of navItems" [routerLink]="[navItem.route]" (click)="skipNavigation()">
     {{navItem.name}}
   </a>
 </nav>
 
 <div #maincontent tabindex="0">
+  <h1 #header tabindex="0" *ngIf="!!currentComponent">{{currentComponent}} Demo</h1>
```
jelbourn: We should make the `tabindex="-1"` so that it can be focused programmatically without being in the tab order

src/demo-app/a11y/a11y.ts
```diff
@@ -43,8 +47,17 @@ export class AccessibilityDemo {
...
+        return fragments[fragments.length - 1] === navItem.route;
+      });
+      this.currentComponent = nav ? nav.name : '';
+    });
+  }
+
   skipNavigation() {
-    this.mainContent.nativeElement.scrollIntoView();
-    this.mainContent.nativeElement.focus();
+    (this.currentComponent ? this.sectionHeader : this.mainContent).nativeElement.focus();
```
jelbourn: Did you find that this worked well without a timeout? If so that works for me.

tinayuangao: This worked well without a timeout. I tested with chromeVox and VoiceOver.

PR #7189, opened 2017-09-19
fix: set selected state for  `MdListOption` in selectedOptions OnInit

### LINKED ISSUE #7183: [MdListOption] selected value not checked when accessing MdSelectionList.selectedOptions.selected
Bug

If a MdListOption's has its ```@Input('selected')``` value set to true upon initialization, it should be included in the ```selectedOptions.selected``` attribute of its enclosing MdSelectionList.

The MdListOption's initial ```@Input('selected')``` value is not respected when accessing the ```selectedOptions.selected``` attribute of its enclosing MdSelectionList.
An option that has its initial selected value to true is not included in the ```selectedOptions.selected``` but when it is subsequently click and thus gets unselected, then it is included in the ```selectedOptions.selected``` attribute of its MdSelectionList.

http://plnkr.co/edit/CXc5dKsoNWzU89s3endl?p=preview

Click the 'Show selected' button.
Then uncheck the selected option ('two') and click again on 'Show selected'.

Angular: 4.3.0
Material: 2.0.0-beta.10
Typescript: 2.3.2

josephperrott:
Fixes #7183

### DIFF
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -219,6 +219,34 @@ describe('MatSelectionList', () => {
     });
   });
 
+  describe('with list option selected', () => {
+    let fixture: ComponentFixture<SelectionListWithListOptions>;
+    let listItemEl: DebugElement;
+    let selectionList: DebugElement;
+
+    beforeEach(async(() => {
+      TestBed.configureTestingModule({
+        imports: [MdListModule],
+        declarations: [SelectionListWithSelecedOption],
+      });
+
+      TestBed.compileComponents();
+    }));
+
+    beforeEach(async(() => {
+      fixture = TestBed.createComponent(SelectionListWithSelecedOption);
+      listItemEl = fixture.debugElement.query(By.directive(MdListOption));
+      selectionList = fixture.debugElement.query(By.directive(MdSelectionList));
+      fixture.detectChanges();
+    }));
+
+    it('should set its initial selected state in the selectedOptions', () => {
+      let optionEl = listItemEl.injector.get(MdListOption);
+      let selectedOptions = selectionList.componentInstance.selectedOptions;
+      expect(selectedOptions.isSelected(optionEl)).toBeTruthy();
+    });
+  });
+
   describe('with single option', () => {
     let fixture: ComponentFixture<SelectionListWithOnlyOneOption>;
     let listOption: DebugElement;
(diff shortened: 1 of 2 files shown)


src/lib/list/selection-list.ts
```diff
@@ -132,6 +132,12 @@ export class MdListOption extends _MdListOptionMixinBase
     super();
   }
 
+  ngOnInit() {
```
DevVersion: Can you add the `OnInit` interface to the class?

src/lib/list/selection-list.spec.ts
```diff
@@ -451,6 +481,13 @@ class SelectionListWithDisabledOption {
 }
 
 @Component({template: `
+<mat-selection-list>
+  <md-list-option [selected]="true">Item</md-list-option>
+</mat-selection-list>
```
DevVersion: The template should have a two space indentation here (as in other test components)

src/lib/list/selection-list.spec.ts
```diff
@@ -214,6 +214,36 @@ describe('MdSelectionList', () => {
...
+    let fixture: ComponentFixture<SelectionListWithListOptions>;
+    let listItemEl: DebugElement;
+    let selectionList: DebugElement;
+
+    beforeEach(async(() => {
+      TestBed.configureTestingModule({
+        imports: [MdListModule],
+        declarations: [
+          SelectionListWithSelecedOption
+        ],
```
DevVersion: Nit: This can fit into a single line.

josephperrott: Done.

src/lib/list/selection-list.spec.ts
josephperrott: Done

src/lib/list/selection-list.ts
josephperrott: Done

src/lib/list/selection-list.spec.ts
```diff
@@ -451,6 +479,13 @@ class SelectionListWithDisabledOption {
 }
 
 @Component({template: `
+  <mat-selection-list>
+    <md-list-option [selected]="true">Item</md-list-option>
+  </mat-selection-list>`})
+class SelectionListWithSelecedOption {
```
rokikon: You've got a typo here SelectionListWithSelec**T**edOption

PR #7190, opened 2017-09-19
build: include typings/index in root metadata export

jelbourn:
This doesn't fix our current issue (https://github.com/angular/angular/issues/18752), but it should be included anyway.


tools/package-tools/build-release.ts
```diff
@@ -65,7 +65,7 @@ export function composeRelease(buildPackage: BuildPackage) {
     // re-exports everything.
     createMetadataReexportFile(
         releasePath,
-        buildPackage.secondaryEntryPoints.map(p => `./${p}`),
+        buildPackage.secondaryEntryPoints.concat(['typings/index']).map(p => `./${p}`),
```
DevVersion: Isn't that already handled a few lines above? 
> https://github.com/jelbourn/material2/blob/31ea051ca341e63d3f2f4b0e5e98316d60c9cf54/tools/package-tools/build-release.ts#L51

jelbourn: This call to `createMetadataReexportFile` overwrites the file generated by the earlier call

DevVersion: Okay that makes sense. I wonder if it would be more obvious if we place the earlier call in the `else` clause.

PR #7193, opened 2017-09-20
chore: move mat re-exports to package where their metadata lives

jelbourn:
This also required creating a module file for every component because
the max re-export can't be in the same file as the module definition or
else ngc will *only* export the alias.


DevVersion:
Maybe we can create a script that uses the TypeScript API and checks the exports of modules?

crisbeto:
If we're doing that, it may be easier to auto-generate the files in the first place.

DevVersion:
That's a good point. Not sure if it would be worth that effort though.

PR #7204, opened 2017-09-20
build: copy stylesheets to release output

DevVersion:
* Copies stylesheets of secondary entry-points to the root of the release output
* Better deduping for the `_theming.scss` file (currently includes multiple `cdk-a11y` mixins)
* Removes unused `cdk-a11y` import in the `button-toggle` component.
* Moves some logic out of the `BuildPackage` class (in favor of readability)


jelbourn:
@DevVersion I'm going to try to merge #7112 before getting to this one

PR #7212, opened 2017-09-20
fix(list-key-manager): align matching logic with native listbox

crisbeto:
Currently the typeahead option of the `ListKeyManager` looks for matches from start to end which can lead to some weird behavior where the user might be half-way down the list, but be sent up to the first item (e.g. in a list of `b, a, ba` it would always find `b`). These changes align the behavior closer to the native `listbox` by looking for the next match after the active item. If no match is found, the selection wraps around and starts looking from the beginning.

### DIFF
--- a/src/cdk/a11y/list-key-manager.ts
+++ b/src/cdk/a11y/list-key-manager.ts
@@ -73,11 +73,14 @@ export class ListKeyManager<T extends ListKeyManagerOption> {
       .subscribe(inputString => {
         const items = this._items.toArray();
 
-        for (let i = 0; i < items.length; i++) {
-          let item = items[i];
+        // Start at 1 because we want to start searching at the item immediately
+        // following the current active item.
+        for (let i = 1; i < items.length + 1; i++) {
+          const index = (this._activeItemIndex + i) % items.length;
+          const item = items[index];
 
           if (!item.disabled && item.getLabel!().toUpperCase().trim().indexOf(inputString) === 0) {
-            this.setActiveItem(i);
+            this.setActiveItem(index);
             break;
           }
         }
--- a/src/cdk/a11y/list-key-manager.spec.ts
+++ b/src/cdk/a11y/list-key-manager.spec.ts
@@ -491,6 +491,62 @@ describe('Key managers', () => {
         expect(keyManager.activeItem).toBeFalsy();
       }));
 
+      it('should start looking for matches after the active item', fakeAsync(() => {
+        itemList.items = [
+          new FakeFocusable('Bilbo'),
+          new FakeFocusable('Frodo'),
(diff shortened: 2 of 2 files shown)


src/cdk/a11y/list-key-manager.ts
```diff
@@ -71,12 +71,22 @@ export class ListKeyManager<T extends ListKeyManagerOption> {
...
+        // after the active one. If we reach the end before finding a match,
+        // wrap back and look through the items before the active one.
+        for (let i = startIndex; i < total; i++) {
           if (items[i].getLabel!().toUpperCase().trim().indexOf(inputString) === 0) {
             this.setActiveItem(i);
             break;
+          } else if (!hasWrapped && startIndex > 0 && i === total - 1) {
+            i = -1;
+            total = startIndex;
+            hasWrapped = true;
```
jelbourn: Use mod instead of manually wrapping?
```ts
const items = this._items.toArray();

// Start at 1 because we want to start searching at the item immediately
// following the current active item.
for (let i = 1; i < items.length; i++) {
  const index = (this._activeItemIndex + i) % items.length;
  if (items[index].getLabel!().toUpperCase().trim().indexOf(inputString) === 0) {
    this.setActiveItem(i);
    break;
  }
}
```

crisbeto: Hmm interesting. It doesn't seem to work on smaller lists though, E.g. on one of the unit tests we have a list of `one, two, three` but it never reaches `three`.

jelbourn: Oh, it needs a `+ 1`

jelbourn: This always does what I'd expect:
```ts
var items = ['one', 'two', 'three'];
var activeIndex = 1;
for (let i = 1; i < items.length; i++) {
  const index = (activeIndex + i) % items.length;
  console.log(index, ' ', items[index]);
}
```
You mean for it to always skip the active item, right?

crisbeto: It's supposed to skip it but then wrap around if it doesn't find anything else. I got it working like this:

```ts
for (let i = 1; i < items.length + 1; i++) {
  const index = (this._activeItemIndex + i) % items.length;
  if (items[index].getLabel!().toUpperCase().trim().indexOf(inputString) === 0) {
    this.setActiveItem(index);
    break;
  }
}
```

Also I think you had a typo. The `this.setActiveItem(i)` was supposed to be `this.setActiveItem(index)`.

jelbourn: Yeah, I did have that typo.

andrewseguin:
Please rebase

PR #7219, opened 2017-09-21
compat: use mat in internal templates

andrewseguin:


jelbourn:
Missed `year-view.html`

jelbourn:
Also one lint error
```
ERROR: /home/travis/build/angular/material2/src/lib/core/option/option.ts[16, 3]: 'Inject' is declared but never used.
```

andrewseguin:
Should be good to go

PR #7241, opened 2017-09-21
compat: remove all md selectors, inputs, classnames, and exports

andrewseguin:


SchnWalter:
It would be nice if you could fix the [release notes](https://github.com/angular/material2/releases/tag/2.0.0-beta.12) and mark this PR as a "breaking change" and not as a "bug fix".

Thanks!

PR #7263, opened 2017-09-22
fix(paginator): page size selector not working

crisbeto:
* Fixes the page size selection stretching to 100% of the paginator.
* Fixes the `MdFormFieldModule` not being imported by `md-select`, causing it to look wrong.

### DIFF
--- a/src/lib/paginator/paginator.html
+++ b/src/lib/paginator/paginator.html
@@ -3,11 +3,13 @@
     {{_intl.itemsPerPageLabel}}
   </div>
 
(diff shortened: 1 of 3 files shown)


Splaktar:
Thank you for fixing this. This is the last issue that I've found when upgrading from beta.11 to master (which I need due to the nested menu fixes).

Are the screenshot test failures blocking this? They seem unrelated to this PR.

Hopefully this can get merged soon.

crisbeto:
The screenshot tests are non-blocking and the fix should be merged soon @Splaktar.

PR #7265, opened 2017-09-22
feat(progress-spinner): switch to css-based animation

kara:
WIP PR that wraps crisbeto's original commit to support customizing spinner's diameter. Necessary to keep support for setting width/height and strokeWidth in order to pass internal Google testing.


kara:
Actually going to push to @crisbeto's original branch, #6551, so closing this one.

Issue #7061, opened 2017-09-13
Only attach on initial view of a tabs content.


josephperrott:
Building on #6832, tabs should allow for lazy loading to only occur for initial load, then reuse already created view.

This will be important for holding the current state of something like a form, as well as no requiring reinitialization of more costly views.

josephperrott:
Really this should be stated as no detaching the view when the tab leaves the view, instead just hiding it.  This does not directly necessitate building off of #6832, but will likely be added after its merged.

josephperrott:
@jelbourn Is this something you still think we should do?  I know that we have tests that will be effected by this, though its functionality is not breaking.

jelbourn:
I think we need to measure what kind of on impact it would actually have

PR #7266, opened 2017-09-22
chore(tabs): use custom portal template directive for hosting tab body content

### LINKED ISSUE #7032: [Bug] Fast tab switch forces scroll up
Bug.

The tab should switch without triggering any scroll.

When you switch tabs after making any scroll, the pages try to scroll up to the beginning of the page

Take a look at the last gif about the tabs example, here's the URL:
https://material.angular.io/components/tabs/api

"@angular/core": "~4.1.0"
"@angular/material": "^2.0.0-beta.8"
Browsers: Google Chrome@60.0.3112.113
Safari@10.1

This behavior can be seen in the next 2 gifs:
material-bug
material-bug-2

In the last one, you can see it even happens in material tabs example.

josephperrott:
Related to #6832
By moving the entrance logic into the `MdTabBodyPortal` setting up the content after change detection is avoided.

Fixes #7032 #7061

### DIFF
--- a/src/lib/tabs/tab-body.spec.ts
+++ b/src/lib/tabs/tab-body.spec.ts
@@ -2,10 +2,10 @@ import {Direction, Directionality} from '@angular/cdk/bidi';
 import {PortalModule, TemplatePortal} from '@angular/cdk/portal';
 import {CommonModule} from '@angular/common';
 import {Component, TemplateRef, ViewChild, ViewContainerRef} from '@angular/core';
-import {async, ComponentFixture, fakeAsync, flushMicrotasks, TestBed} from '@angular/core/testing';
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
 import {MatRippleModule} from '@angular/material/core';
 import {NoopAnimationsModule} from '@angular/platform-browser/animations';
-import {MatTabBody} from './tab-body';
+import {MatTabBody, MatTabBodyPortal} from './tab-body';
 
 
 describe('MatTabBody', () => {
@@ -17,6 +17,7 @@ describe('MatTabBody', () => {
       imports: [CommonModule, PortalModule, MatRippleModule, NoopAnimationsModule],
       declarations: [
         MatTabBody,
+        MatTabBodyPortal,
         SimpleTabBodyApp,
       ],
       providers: [
@@ -145,30 +146,6 @@ describe('MatTabBody', () => {
       expect(fixture.componentInstance.tabBody._position).toBe('left');
     });
   });
-
-  describe('on centered', () => {
-    let fixture: ComponentFixture<SimpleTabBodyApp>;
-
-    beforeEach(fakeAsync(() => {
-      fixture = TestBed.createComponent(SimpleTabBodyApp);
-    }));
-
-    it('should attach the content when centered and detach when not', fakeAsync(() => {
-      fixture.componentInstance.position = 1;
-      fixture.detectChanges();
-      expect(fixture.componentInstance.tabBody._portalOutlet.hasAttached()).toBe(false);
-
-      fixture.componentInstance.position = 0;
-      fixture.detectChanges();
-      expect(fixture.componentInstance.tabBody._portalOutlet.hasAttached()).toBe(true);
-
-      fixture.componentInstance.position = 1;
-      fixture.detectChanges();
-      flushMicrotasks(); // Finish animation and let it detach in animation done handler
-      expect(fixture.componentInstance.tabBody._portalOutlet.hasAttached()).toBe(false);
-    }));
-  });
-
 });
 
 
--- a/src/lib/tabs/tab-body.ts
+++ b/src/lib/tabs/tab-body.ts
@@ -7,17 +7,21 @@
  */
 
 import {
-  ViewChild,
   Component,
   Input,
+  Inject,
   Output,
   EventEmitter,
+  OnDestroy,
   OnInit,
   ElementRef,
+  Directive,
   Optional,
-  AfterViewChecked,
   ViewEncapsulation,
   ChangeDetectionStrategy,
+  ComponentFactoryResolver,
+  ViewContainerRef,
+  forwardRef,
 } from '@angular/core';
 import {
   trigger,
@@ -29,7 +33,10 @@ import {
 } from '@angular/animations';
 import {TemplatePortal, CdkPortalOutlet} from '@angular/cdk/portal';
 import {Directionality, Direction} from '@angular/cdk/bidi';
+import {Subscription} from 'rxjs/Subscription';
 
+/** Workaround for https://github.com/angular/angular/issues/17849 */
+export const _MatTabBodyPortalBaseClass = CdkPortalOutlet;
 
 /**
  * These position states are used internally as animation states for the tab body. Setting the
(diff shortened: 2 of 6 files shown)


src/lib/tabs/tab-body.ts
```diff
@@ -88,13 +139,16 @@ export type MdTabBodyOriginState = 'left' | 'right';
...
+export class MdTabBody implements OnInit {
   /** The portal host inside of this container into which the tab body content will be loaded. */
-  @ViewChild(PortalHostDirective) _portalHost: PortalHostDirective;
+  @ViewChild(MdTabBodyPortal) _portalHost: MdTabBodyPortal;
 
   /** Event emitted when the tab begins to animate towards the center as the active tab. */
   @Output() onCentering: EventEmitter<number> = new EventEmitter<number>();
 
+  /** Event emitted when the tab completes its animation away from the center. */
+  @Output() afterLeavingCenter: EventEmitter<number> = new EventEmitter<number>();
```
amcdnl: Can we prefix this w/ a underscore since its private ... so `_afterLeavingCenter`?

Whats the `number` type used for?

josephperrott: Done, and the number was the result of a copy paste from the emitter above.  Since I wasn't using the result at all I didn't notice.

josephperrott:
@andrewseguin If you have a moment to take a look. :D

src/lib/tabs/tab-body.spec.ts
```diff
@@ -153,19 +154,16 @@ describe('MdTabBody', () => {
       fixture = TestBed.createComponent(SimpleTabBodyApp);
     }));
 
-    it('should attach the content when centered and detach when not', fakeAsync(() => {
+    it('should attach when first made visible', fakeAsync(() => {
       fixture.componentInstance.position = 1;
       fixture.detectChanges();
+      tick(1000);
```
andrewseguin: Leave comment explaining where 1000 came from and why tick was needed

src/lib/tabs/tab-body.ts
```diff
@@ -53,6 +58,53 @@ export type MdTabBodyPositionState =
...
+ * The portal host directive for the contents of the tab.
+ * @docs-private
+ */
+@Directive({
+  selector: '[matTabBodyHost]'
+})
+export class MdTabBodyPortal extends _MdTabBodyPortalBaseClass implements OnInit {
+  /** Whether the tab body should be visible. */
+  visible = false;
+  /** A subscription to events for when the tab body begins centering. */
```
andrewseguin: Looks like the subs should be private (preceeded with an underscore)

src/lib/tabs/tab-body.ts
```diff
@@ -53,6 +58,53 @@ export type MdTabBodyPositionState =
...
+  centeringSub: Subscription;
+  /** A subscription to events for when the tab body completes leaving from center. */
+  leavingCenterSub: Subscription;
+
+  constructor(
+    _componentFactoryResolver: ComponentFactoryResolver,
+    _viewContainerRef: ViewContainerRef,
+    private host: MdTabBody) {
```
andrewseguin: Add underscore to host to show its private

src/lib/tabs/tab-body.ts
```diff
@@ -53,6 +58,53 @@ export type MdTabBodyPositionState =
...
+      super(_componentFactoryResolver, _viewContainerRef);
+  }
+
+  /** Set up subscriptions for changing visibility and set initial visibility. */
+  ngOnInit(): void {
+    this.centeringSub = this.host.onCentering.subscribe(() => this.setVisibility(true));
+    this.leavingCenterSub = this.host._afterLeavingCenter.subscribe(
```
andrewseguin:
```js
this.leavingCenterSub = 
    this.host._afterLeavingCenter.subscribe(() => this.setVisibility(false));
```

src/lib/tabs/tab-group.spec.ts
```diff
@@ -322,23 +322,25 @@ describe('MdTabGroup', () => {
...
     });
 
-    it('should support a tab-group with the simple api', () => {
+    it('should support a tab-group with the simple api', async(() => {
       expect(getSelectedLabel(fixture).textContent).toMatch('Junk food');
       expect(getSelectedContent(fixture).textContent).toMatch('Pizza, fries');
 
       tabGroup.selectedIndex = 2;
       fixture.detectChanges();
+      fixture.whenStable().then(() => {
```
andrewseguin: Add comment why `whenStable` is needed

src/lib/tabs/tabs.spec.ts
```diff
@@ -0,0 +1,44 @@
+import {Component} from '@angular/core';
```
andrewseguin: Not sure why the extra test is needed, isn't this covered by `tab-group.spec.ts`?

src/lib/tabs/tab-body.ts
```diff
@@ -175,7 +220,7 @@ export class MdTabBody implements OnInit, AfterViewChecked {
   }
 
   /** Whether the provided position state is considered center, regardless of origin. */
-  private _isCenterPosition(position: MdTabBodyPositionState|string): boolean {
+  _isCenterPosition(position: MdTabBodyPositionState|string): boolean {
```
andrewseguin: Will need to remove the underscore if it's not private.

src/lib/tabs/tab-body.spec.ts
josephperrott: Done.

src/lib/tabs/tab-body.ts
josephperrott: Done

src/lib/tabs/tab-body.ts
josephperrott: Done

src/lib/tabs/tab-body.ts
josephperrott: Done

src/lib/tabs/tabs.spec.ts
josephperrott: It was not caught by the current tests in `tab-group.spec.ts`, would you prefer I move it into `tab-group.spec.ts`?

src/lib/tabs/tab-group.spec.ts
josephperrott: Done.

src/lib/tabs/tab-body.ts
josephperrott: Ack.

andrewseguin:
Add `merge_ready` after you move the test to `tab-group.spec.ts`

src/lib/tabs/tab-body.ts
```diff
@@ -53,6 +59,44 @@ export type MdTabBodyPositionState =
...
+  selector: '[matTabBodyHost]'
+})
+export class MdTabBodyPortal extends _MdTabBodyPortalBaseClass implements OnInit, OnDestroy {
+  /** A subscription to events for when the tab body begins centering. */
+  private _centeringSub: Subscription;
+
+  constructor(
+    _componentFactoryResolver: ComponentFactoryResolver,
+    _viewContainerRef: ViewContainerRef,
+    private _host: MdTabBody) {
```
mmalerba: @josephperrott it looks like we need to use a forwardRef here since this directive is used in the template for `MdTabBody`. Here's an example for a similar situation in the sidenav: https://github.com/angular/material2/blob/master/src/lib/sidenav/sidenav.ts#L36

josephperrott: Done

mmalerba:
@josephperrott This seems to break tab animations in some cases. Go to tabs demo page under "Tab Group Demo - Dynamic Tabs" and click "Tab 4" to see the broken behavior

josephperrott:
@mmalerba Issue resolved.

amcdnl:
@josephperrott - Can you rebase?

PR #7273, opened 2017-09-23
fix(input): don't highlight container when readonly input is focused

crisbeto:
Prevents readonly inputs from being highlighted when they're focused. This is consistent with the native input behavior.

Note: This was introduced initially through #5776, but it looks like it didn't make it through the transition to `mat-form-field`.


andrewseguin:
From internal testing - looks like this a regression of the native input's `readonly` attribute. 

e.g. `<input [readonly]=true>` will result in `<input ng-reflect-readonly=true>`  instead of `<input readonly>`

PR #7280, opened 2017-09-24
build: faster demo-app building

DevVersion:
* The demo-app no longer builds every secondary entry-point separately. Now the everything of a package is built at the same time (as for the tests). This is significantly faster than before.
* No longer builds the CDK package twice when building the tests or demo-app.

**Note**: It might be a bit slower when loading the page, because there is no UMD bundle anymore, but it's still a time win.

### DIFF
--- a/src/demo-app/system-config.ts
+++ b/src/demo-app/system-config.ts
@@ -1,10 +1,10 @@
 /** Type declaration for ambient System. */
 declare const System: any;
 
-// Apply the CLI SystemJS configuration.
+// Configure the base path and map the different node packages.
 System.config({
   paths: {
-    'node:*': 'node_modules/*',
+    'node:*': 'node_modules/*'
   },
   map: {
     'rxjs': 'node:rxjs',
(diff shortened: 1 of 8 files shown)


src/demo-app/system-config.ts
```diff
@@ -17,36 +17,69 @@ System.config({
...
-    '@angular/cdk/platform': 'dist/bundles/cdk-platform.umd.js',
-    '@angular/cdk/portal': 'dist/bundles/cdk-portal.umd.js',
-    '@angular/cdk/rxjs': 'dist/bundles/cdk-rxjs.umd.js',
-    '@angular/cdk/scrolling': 'dist/bundles/cdk-scrolling.umd.js',
-    '@angular/cdk/stepper': 'dist/bundles/cdk-stepper.umd.js',
-    '@angular/cdk/table': 'dist/bundles/cdk-table.umd.js',
-    '@angular/cdk/testing': 'dist/bundles/cdk-testing.umd.js',
+    // TODO(devversion): replace once the index.ts file for the Material package has been added.
+    '@angular/material': 'dist/packages/material/public_api.js',
+    '@angular/cdk': 'dist/packages/cdk/index.js',
```
crisbeto: I haven't tried it out, but you should be able to replace these mappings with something like this in the `paths` config:

```
'@angular/cdk/*': 'dist/packages/cdk/*/index.js'
```

tools/gulp/tasks/development.ts
```diff
@@ -32,9 +30,12 @@ task(':watch:devapp', () => {
   watchFiles(join(appDir, '**/*.scss'), [':build:devapp:scss']);
   watchFiles(join(appDir, '**/*.html'), [':build:devapp:assets']);
 
-  // The themes for the demo-app are built by the demo-app using the SCSS mixins from Material.
-  // Therefore when SCSS files have been changed, the custom theme needs to be rebuilt.
-  watchFiles(join(materialOutPath, '**/*.scss'), [':build:devapp:scss']);
+  // Custom watchers for the CDk, Material and Moment adapter package. This is necessary because
```
crisbeto: CDk -> CDK

src/demo-app/system-config.ts
DevVersion: Unfortunately wildcards are not supported for the `map` option in SystemJS. In the future I hope we can use something like [that](https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#packageconfigpaths)

jelbourn:
Actually looks like there are some lint / aot issues

DevVersion:
Fixed the lint and AOT issues. Please have another look.

PR #7281, opened 2017-09-24
chore(select): avoid test flakes due to sub-pixel deviations

crisbeto:
Fixes some tests that were failing when they're run against Chrome on Windows due to sub-pixel measurement differences.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1042,10 +1042,14 @@ describe('MdSelect', () => {
       // Extra trigger height beyond the font size caused by the fact that the line-height is
       // greater than 1em.
       const triggerExtraLineSpaceAbove = (1 - triggerLineHeightEm) * triggerFontSize / 2;
+      const topDifference = Math.floor(optionTop) -
+          Math.floor(triggerTop - triggerFontSize - triggerExtraLineSpaceAbove);
 
-      expect(Math.floor(optionTop))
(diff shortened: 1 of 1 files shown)


src/lib/select/select.spec.ts
```diff
@@ -1042,10 +1042,14 @@ describe('MdSelect', () => {
...
+      const topDifference = Math.floor(optionTop) -
+          Math.floor(triggerTop - triggerFontSize - triggerExtraLineSpaceAbove);
 
-      expect(Math.floor(optionTop))
-          .toBe(Math.floor(triggerTop - triggerFontSize - triggerExtraLineSpaceAbove),
-              `Expected trigger to align with option ${index}.`);
+      // Expect the coordinates to be within a pixel of each other. We can't rely on comparing
+      // the exact value, because different browsers report the various sizes with slight (< 1px)
+      // deviations.
+      expect(Math.abs(topDifference) < 2)
```
kara: Could we make this a bit tighter - to 1 pixel difference? Concerned the test may not catch actual 1-2px offsets.

crisbeto: This is 1 pixel though. `Math.abs(difference) < 2` will catch differences of 0 and 1.

kara: Oh, I missed the `Math.floor` above, so I was talking about differences of 1.5, etc. Ignore me then.

Issue #6370, opened 2017-08-09
[audit] use of coerceBooleanProperty


willshowell:
#### Bug, feature request, or proposal:
Suggesting library-wide audit

#### Expected
Consistent use

#### Currently
At least two boolean inputs ([1](https://github.com/angular/material2/blob/master/src/lib/datepicker/datepicker.ts#L143) [2](https://github.com/angular/material2/blob/master/src/lib/expansion/accordion-item.ts#L42)) do not use it

#### Motivation

https://github.com/angular/material2/issues/6344#issuecomment-320955604, https://github.com/angular/material2/issues/6097#issuecomment-319083700

#### Is there anything else we should know?

IMO it should be removed entirely as it encourages misuse of bindings that won't translate outside this library. For reference, the original commit used it in the form of [`@BooleanFieldValue()`](https://github.com/angular/material2/pull/161/files#diff-6ef16658b3712f44bd8299cc8b9c9584).

jelbourn:
We should use it consistently everywhere. It's necessary at the very least to make things like `disabled`, `readonly`, `required`, and `checked` work correctly. 

My long-term goal is to convince core to do https://github.com/angular/angular/issues/8968

PR #7283, opened 2017-09-24
fix: consistently coerce boolean and number properties to the correct type

amcdnl:
Addresses #6370


src/lib/slide-toggle/slide-toggle.ts
```diff
@@ -125,6 +125,7 @@ export class MdSlideToggle extends _MdSlideToggleMixinBase implements OnDestroy,
   @Input()
   get checked(): boolean { return this._checked; }
   set checked(value) {
+    value = coerceBooleanProperty(value);
     this._checked = !!value;
```
rafaelss95: `this._checked = coerceBooleanProperty(value);`?

jelbourn: Yep, it should be that

amcdnl: Done :)

rafaelss95: `coerceBooleanProperty` already returns a boolean, you don't need `!!` :)

amcdnl: #facepalm

src/lib/datepicker/datepicker.ts
```diff
@@ -142,7 +142,14 @@ export class MatDatepicker<D> implements OnDestroy {
    * Whether the calendar UI is in touch mode. In touch mode the calendar opens in a dialog rather
    * than a popup and elements have more padding to allow for bigger touch targets.
    */
-  @Input() touchUi = false;
+  @Input()
+  get touchUi() {
+    return this._touchUi;
+  }
+  set touchUi(value: any) {
```
mmalerba: I think its best to leave the types as `boolean` that way when people `@ViewChild` this and try to programmatically set it they at least know that they should be passing a boolean (and then behind the scenes we coerce it anyways it just in case it came from a template binding).

src/lib/tabs/tab-group.ts
```diff
@@ -111,7 +111,7 @@ export class MatTabGroup extends _MatTabGroupMixinBase implements AfterContentIn
 
   /** The index of the active tab. */
   @Input()
-  set selectedIndex(value: number | null) { this._indexToSelect = value; }
+  set selectedIndex(value: number | null) { this._indexToSelect = coerceNumberProperty(value); }
```
mmalerba: it looks like the author was intentionally trying to allow `null` here in which case I think you want `coerceNumberProperty(value, null)`

src/lib/datepicker/datepicker.ts
amcdnl: I just followed the pattern that disabled had right below it. I'll update both.

src/lib/tabs/tab-group.ts
amcdnl: The `coerceNumberProperty` signature does not support `null` as the second parameter. Should we extend it you think to support that?

src/lib/datepicker/datepicker.ts
mmalerba: yeah I think there's a lot that use `any`, but we can use the expected type for new ones we add and maybe eventually fix all the `any` ones

src/lib/tabs/tab-group.ts
mmalerba: Hmm, in a lot of cases people really do just want number. What if we did this?

```ts
export function coerceNumberProperty(value: any): number;
export function coerceNumberProperty<D>(value: any, fallback: D): number | D;
export function coerceNumberProperty(value: any, fallbackValue = 0) { ... }
```

It would allow people to get just a `number` in 1-arg case and `number|whatever` in 2-arg case. @jelbourn WDYT?

jelbourn: That seems good to me

amcdnl: Works well.

jelbourn:
@amcdnl can you change the PR title to
`fix: consistently coerce boolean and number properties to the correct type"
?

amcdnl:
done :)

src/cdk/accordion/accordion-item.ts
```diff
@@ -41,8 +42,10 @@ export class CdkAccordionItem implements OnDestroy {
 
   /** Whether the AccordionItem is expanded. */
   @Input()
-  get expanded(): boolean { return this._expanded; }
-  set expanded(expanded: boolean) {
+  get expanded(): any { return this._expanded; }
```
josephperrott: Shouldn't this getter now return `boolean`?

PR #7285, opened 2017-09-25
fix(chips): do not set chips value if there's no ngControl or value

### LINKED ISSUE #7261: Chip-list overrides initial chip selected state
Say you have a template like this:
```html
<mat-chip-list>
  <mat-chip *ngFor="let i of items" selected="true"> {{i}} </mat-chip>
</mat-chip-list>
```

Because of this async behavior in chip-list:
https://github.com/angular/material2/blob/3571f68a2de0833ee4741af8a0eb24d1da174f38/src/lib/chips/chip-list.ts#L574-L577

The individual chip selection gets overwritten by the chip-list on init.

@tinayuangao we should find a way to avoid using `Promise.resolve` here

tinayuangao:
Fixes #7261

### DIFF
--- a/src/lib/chips/chip-list.ts
+++ b/src/lib/chips/chip-list.ts
@@ -572,8 +572,10 @@ export class MatChipList implements MatFormFieldControl<any>, ControlValueAccess
     // Defer setting the value in order to avoid the "Expression
     // has changed after it was checked" errors from Angular.
     Promise.resolve().then(() => {
-      this._setSelectionByValue(this.ngControl ? this.ngControl.value : this._value, false);
-      this.stateChanges.next();
+      if (this.ngControl || this._value) {
+        this._setSelectionByValue(this.ngControl ? this.ngControl.value : this._value, false);
+        this.stateChanges.next();
+      }
     });
   }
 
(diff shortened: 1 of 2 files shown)


src/lib/chips/chip-list.ts
```diff
@@ -572,8 +572,10 @@ export class MdChipList implements MdFormFieldControl<any>, ControlValueAccessor
     // Defer setting the value in order to avoid the "Expression
     // has changed after it was checked" errors from Angular.
     Promise.resolve().then(() => {
-      this._setSelectionByValue(this.ngControl ? this.ngControl.value : this._value, false);
-      this.stateChanges.next();
+      if (!!this.ngControl || !!this._value) {
```
jelbourn: You don't need the `!!` in conditional expressions, only when the value will be passed around somewhere

tinayuangao: Done. PTAL. Thanks for review.

jelbourn: Need to push your changes?

tinayuangao: Changes are pushed. Thanks!

PR #7299, opened 2017-09-25
fix(ripple): handle touch events

### LINKED ISSUE #7062: Ripples behavior on touch-enabled devices doesn't match Android ripples behavior

Ripples behavior on touch-enabled devices doesn't match Android ripples behavior.

The ripple should happen during the touch, not after it.

Please note that the ripple happened in the end, **a bit of time after the button was clicked**.

Go to https://material.angular.io/

Follow Android design.

The latest ones for everything.

The issue is also present when you enable touch-mode on Firefox for Windows. Untested on Chrome for Windows. The issue affects Chrome for Android and Firefox for Android, all latest versions.

Materialize CSS framework seems to have fixed this issue.

DevVersion:
* Ripples are now launched on touchstart for touch devices. Before they were just launched after the click happened.

Fixes #7062

### DIFF
--- a/src/cdk/testing/event-objects.ts
+++ b/src/cdk/testing/event-objects.ts
@@ -41,7 +41,7 @@ export function createTouchEvent(type: string, pageX = 0, pageY = 0) {
   // Most of the browsers don't have a "initTouchEvent" method that can be used to define
   // the touch details.
   Object.defineProperties(event, {
-    touches: { value: [touchDetails] }
+    touches: {value: [touchDetails]}
   });
 
   return event;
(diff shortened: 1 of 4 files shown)


src/cdk/testing/event-objects.ts
```diff
@@ -29,6 +29,24 @@ export function createMouseEvent(type: string, x = 0, y = 0) {
...
+  // as a basic UI Event. The necessary details for the event will be set manually.
+  const event = document.createEvent('UIEvent');
+  const touchDetails = {pageX, pageY};
+
+  event.initUIEvent(type, true, true, window, 0);
+
+  // Most of the browsers don't have a "initTouchEvent" method that can be used to define
+  // the touch details.
+  Object.defineProperties(event, {
+    touches: { value: [touchDetails] }
```
jelbourn: Omit spaces

PR #7305, opened 2017-09-25
fix(button): allow for elevation to be overwritten

### LINKED ISSUE #7264: Flat button with inverted colors (accent color background)
Feature request:
A flat button with accent background and its corresponding text color.

Basically it's a raised button without its shadow.
image

Those buttons are used in many situations since you are not supposed to use the raised button on elements with z-elevation. 
They are already present in many (if not all) Google apps and websites with material design.
Even though they are not described in the official guidelines, the image you see is taken from the material guidelines website on the 'button' page.

crisbeto:
Allows for the consumer to set the button elevation through the elevation classes, similarly to the menu, card and expansion components.

Fixes #7264.

### DIFF
--- a/src/lib/button/_button-base.scss
+++ b/src/lib/button/_button-base.scss
@@ -66,7 +66,7 @@ $mat-mini-fab-padding: 8px !default;
 %mat-raised-button {
   @extend %mat-button-base;
 
-  @include mat-elevation(2);
+  @include mat-overridable-elevation(2);
 
   // Force hardware acceleration.
   transform: translate3d(0, 0, 0);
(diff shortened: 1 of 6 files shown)


src/lib/button/_button-base.scss
```diff
@@ -66,7 +66,9 @@ $mat-mini-fab-padding: 8px !default;
 %mat-raised-button {
   @extend %mat-button-base;
 
-  @include mat-elevation(2);
+  &:not([class*='#{$_mat-elevation-prefix}']) {
```
jelbourn: Would it make sense to make a mixin for this? Something like...
```scss
@mixin mat-overridable-elevation($z) {
  &:not([class*='#{$_mat-elevation-prefix}']) {
    @include mat-elevation(2);
  }
}
```

crisbeto: Good idea, that selector was pretty funky. Added.

PR #7309, opened 2017-09-25
fix(expansion): add expansion panel outputs

### LINKED ISSUE #7172: Md-expansion panel opened and close events not firing on prod build (beta.11)
Looks like a bug.

Event handler is invoked when opened and closed events happen.

Event handler is not invoked on opened and closed events.

Attach handlers for such events in the template and check if they are invoked after being built with angular cli's ng build --prod.

Has to be working in production build

I'm attaching event handlers like this:
`<md-expansion-panel
      (closed)="onCollapse()"
      (opened)="onExpand()">`
and process them in component. But these handlers are not invoked after prod build. It works OK when I do ng build --dev or ng serve.

josephperrott:
Fixes #7172

### DIFF
--- a/e2e/components/expansion-e2e.spec.ts
+++ b/e2e/components/expansion-e2e.spec.ts
@@ -5,15 +5,19 @@ describe('expansion', () => {
 
   beforeEach(() => browser.get('/expansion'));
 
-  it('should show a expansion panel', async () => {
-    expect(element(by.css('.mat-expansion-panel'))).toBeDefined();
+  it('should show an accordion', async () => {
+    expect(element(by.css('.mat-accordion'))).toBeDefined();
+    screenshot();
+  });
 
+  it('should show two panels', async () => {
+    expect(element.all(by.css('.mat-expansion-panel')).count()).toBe(2);
     screenshot();
   });
 
   it('should hide contents of the expansion panel on click', async () => {
-    const panelHeader = element(by.css('.mat-expansion-panel-header'));
-    const panelContent = element(by.css('.mat-expansion-panel-content'));
+    const panelHeader = element.all(by.css('.mat-expansion-panel-header')).get(0);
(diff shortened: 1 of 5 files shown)


josephperrott:
@jelbourn  e actually already check this in tests, this error only came to light after `enableProd()` was called on the app.

josephperrott:
@jelbourn This was actually already check for in tests, the error only came to light after `enableProd()` was called in an app using it.

danwulff:
@josephperrott maybe an e2e test would catch this?

josephperrott:
@jelbourn Added this to e2e tests since the tests run using AOT and have AOT tests running in prod mode. Thoughts?

src/e2e-app/expansion/expansion-e2e.html
```diff
@@ -0,0 +1,21 @@
+<mat-accordion>
```
jelbourn: Instead of adding a separate e2e page here, can you test an example material-examples?

josephperrott:
@jelbourn moved to testing the ExpansionOverviewExample

willshowell:
Friendly ping, since it looks like this is ready for review again

jaruesink:
This seems to be a blocker for anyone wanting to use this component in prod. I was excited when the RC0 came out, because I thought this would be in it.

bardiel:
@jaruesink, in the meantime, you can do something as stated en this comment: https://github.com/angular/material2/issues/7172#issuecomment-332582894

That's just enough to get the expected behavior. Once this gets merged, that workarround will still work, so you'll be able to migrate your code painlessly.

hiteshgarg1994:
<mat-expansion-panel #exeEleRef></mat-expansion-panel>


@ViewChild('exeEleRef') private exeEleRef: any 

  ngAfterViewInit(): void {
    this.exeEleHrly.opened.subscribe(x => {
      // do something
    })
  }

PR #7310, opened 2017-09-25
fix(slide-toggle): update model before emitting change event

### LINKED ISSUE #7300: Should md-slide-toggle (change) event update the model before calling the event handler?

Linked to #448 - the (change) event method for md-slide-toggle component appears to be called before the model is updated for [(ngModel)] bound controls.

Model is updated before the change event handler fires.

Change event fires and then the model is updated.

Add a change event handler to a md-slide-toggle control bound to a model property using [(ngModel)] syntax. 

Inconsistent with native browser and other Angular Change events.

Angular 4.0.0, Material 2.0.0-beta.11, Typescript 2.3.3. Windows 10 Enterprise O/S.

josephperrott:
Fixes #7300

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -100,6 +100,19 @@ describe('MdSlideToggle without forms', () => {
       expect(slideToggle.checked).toBe(true);
     });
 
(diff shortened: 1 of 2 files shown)


DevVersion:
The fix for that is actually pending in #7076

josephperrott:
Awesome, didnt see it when I glanced for it before.

PR #7312, opened 2017-09-26
Adds Coerce Strategy as provider

### LINKED ISSUE #7286: Unable to implement own coerceDateProperty strategy.
https://github.com/angular/material2/blob/d2ceb2c755e20252f02fda11dc4eef55d20f4bbe/src/lib/datepicker/datepicker-input.ts#L37

Bug
Ability to inject custom date coercion function or have numbers coerced as Unix Timestamps.

Currrently strings are parsed as `Iso8601` https://github.com/angular/material2/blob/d2ceb2c755e20252f02fda11dc4eef55d20f4bbe/src/lib/datepicker/coerce-date-property.ts#L23
and unix timestamps throw errors

https://github.com/angular/material2/blob/d2ceb2c755e20252f02fda11dc4eef55d20f4bbe/src/lib/datepicker/coerce-date-property.ts#L33

[plunkr](https://plnkr.co/edit/Uj2wuTM5WeB1k2AurnoP?p=preview)

Numbers should be coerced as Unix Timestamps.

paulussup:
Fixes #7286


paulussup:
I signed it!

mmalerba:
I think maybe a better way to do this would be to change the `fromIso8601` method on `DateAdapter` to `coerceToDate(obj: any): D | null`. Then we can leave it up to the author of the particular `DateAdapter` what kind of things they want to be coerceable to dates

paulussup:
I think even better would be to remove the coerce date function altogether and have it as a method on the date adapter.

mmalerba:
I think we should keep it around, but just for the purpose of checking the validity of the value returned by the `DateAdapter`. I imagine the method looking something like this:
```ts
export function coerceDateProperty<D>(adapter: DateAdapter<D>, value: any): D | null {
  value = adapter.coerceToDate(value);
  if (adapter.isDateInstance(value) && !adapter.isValid(value)) {
    throw Error(`Datepicker: Failed to coerce value to a date object: ${value}`);
  }
  return value;
}
```
And then its up to the `DateAdapter` to decide if a value coerces to a valid date, or null, or return an invalid date if it could not be coerced.

PR #7341, opened 2017-09-26
fix(input): make autosize work inside tabs & stepper

### LINKED ISSUE #7316: md-autosize initialized with height of 0 inside stepper & tabs
Bug

TextArea indicator should always show onload

TextArea indicator with md-autosize not showing inside the Stepper.
The indicator will only show when users starts to type into it.

[Plunker Link](https://plnkr.co/edit/OwCi5QPmaWv8htG8gaWQ?p=preview)

Step 1: Notice that the TextArea outside the stepper is showing the indicator and the TextArea inside the Stepper is not showing the indiciator.
Step 2: Start to type into the TextArea inside the stepper

TextArea Indicator Error

NA

Material Beta 11

NA

mmalerba:
fixes #7316

### DIFF
--- a/src/lib/input/autosize.spec.ts
+++ b/src/lib/input/autosize.spec.ts
@@ -218,14 +218,14 @@ describe('MatTextareaAutosize', () => {
     const fixtureWithForms = TestBed.createComponent(AutosizeTextareaInATab);
     fixtureWithForms.detectChanges();
     textarea = fixtureWithForms.nativeElement.querySelector('textarea');
-    expect(textarea.getBoundingClientRect().height).toBeGreaterThan(0);
+    expect(textarea.getBoundingClientRect().height).toBeGreaterThan(1);
   });
 
   it('should work in a step', () => {
     const fixtureWithForms = TestBed.createComponent(AutosizeTextareaInAStep);
     fixtureWithForms.detectChanges();
     textarea = fixtureWithForms.nativeElement.querySelector('textarea');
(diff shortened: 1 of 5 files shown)


jelbourn:
Actually, add a unit test for stepper + autosize?

mmalerba:
I feel like this kind of thing is better caught with a screenshot test, which I did for tabs. Didn't add it to the stepper e2e test because it uses the stepper overview example, and I didn't want to start adding irrelevant stuff to the example

jelbourn:
An easy sanity test would be to make the initial text really long and assert that the height is above some small-ish threshold

mmalerba:
done

src/lib/input/autosize.spec.ts
```diff
@@ -202,6 +213,20 @@ describe('MdTextareaAutosize', () => {
     expect(textarea.clientHeight)
         .toBeGreaterThan(previousHeight, 'Expected the textarea height to have increased.');
   }));
+
+  it('should work in a tab', () => {
+    const fixtureWithForms = TestBed.createComponent(AutosizeTextareaInATab);
+    fixtureWithForms.detectChanges();
+    textarea = fixtureWithForms.nativeElement.querySelector('textarea');
+    expect(textarea.getBoundingClientRect().height).toBeGreaterThan(0);
```
jelbourn: Would greater than `1` work?

PR #7346, opened 2017-09-26
fix(datepicker): allow `DateAdapter` authors to have more control ove…

mmalerba:
…r what can/can't be coerced to a date

BREAKING CHANGES:
- `fromIso8601` method on `DateAdapter` removed in favor of `coerceToDate`


mmalerba:
rebased

src/lib/core/datetime/date-adapter.ts
```diff
@@ -192,6 +185,23 @@ export abstract class DateAdapter<D> {
...
+   * Attempts to coerce a value to a valid date object. This is different from parsing in that it
+   * should only coerce non-ambiguous, locale-independent values (e.g. a ISO 8601 string).
+   * The default implementation does not allow any coercion, it simply checks that the given value
+   * is already a valid date object or null.
+   * @param value The value to be coerced to a date object.
+   * @returns The coerced date object, either a valid date, null if the value can be coerced to a
+   *     null date (e.g. the empty string).
+   * @throws If the given value cannot be coerced to a valid date or null.
+   */
+  coerceToDate(value: any): D | null {
```
jelbourn: Doesn't this need to align with `toIso8601`? I think the heart of what this is trying to do is the same as `parse` / `format` except that instead of being string that the user see / inputs, it's the formats that the backend sends / accepts. Looking at it that way, I think this would make sense:

`parse` / `format` -> for **user-facing** strings
`serialize` / `deserialize` -> for **backend-facing** strings (or JSON or whatever)

mmalerba: For user-facing strings: I agree we need both.

For backend-facing strings: I don't think the `DateAdapter` is particularly concerned with these for the most part. We don't really know if the dates coming into / out of the datepicker are going to the backend or some other part of the JS app, its up to the user to decide and handle that. However for convenience we might want to allow certain values (such as ISO 8601 strings or UNIX timestamps) to be automatically coerced into the proper type - similar to how we coerce `"true"` and `"false"` for boolean properties. This is what the new `coerceToDate` method is about, allowing the user to decide what can be coerced and what can't.

There's also a third class: DOM-facing strings. For example to set the `min` attribute on the input we need the date in a format that the DOM can understand, which is ISO 8601. This is why we always need to have a way to go from whatever the date type is to an ISO 8601 string.

jelbourn: How do you feel, then, about calling this method `deserialize`? 

Also on `toIso8601`, we should document somewhere why/how this is used.

mmalerba: 👍 

I'll update the description of all the to/from date methods to make sure its clear exactly what each one is used for

mmalerba:
comments addressed

src/lib/core/datetime/date-adapter.ts
```diff
@@ -118,7 +118,7 @@ export abstract class DateAdapter<D> {
   abstract today(): D;
 
   /**
-   * Parses a date from a value.
+   * Parses a date from a user-inputted value.
```
jelbourn: `user-provided`?

src/lib/core/datetime/date-adapter.ts
```diff
@@ -192,6 +187,26 @@ export abstract class DateAdapter<D> {
   abstract isValid(date: D): boolean;
 
   /**
+   * Attempts to deserialize a value to a valid date object. This is different from parsing in that
+   * deserialize should only accept non-ambiguous, locale-independent values (e.g. a ISO 8601
+   * string). The default implementation does not allow any deserialization, it simply checks that
+   * the given value is already a valid date object or null. The `<mat-datepicker>` will call this
+   * method on all of it's `@Input()` properties that accept dates. It is therefore possible to
+   * support passing your wire format directly to these properties by overriding this method to
+   * also deserialize your wire format.
```
jelbourn: Say something like "This will generally be used for data coming directory from some backend or database." ?

src/lib/core/datetime/date-adapter.ts
```diff
@@ -192,6 +187,26 @@ export abstract class DateAdapter<D> {
   abstract isValid(date: D): boolean;
 
   /**
+   * Attempts to deserialize a value to a valid date object. This is different from parsing in that
+   * deserialize should only accept non-ambiguous, locale-independent values (e.g. a ISO 8601
```
jelbourn: "locale-independent values" -> "locale-independent formats"?

Does the phrase "locale-independent" refer specifically to the format? E.g., ISO 8601 not a localized format, but it might not contain UTC

src/lib/core/datetime/date-adapter.ts
mmalerba: Isn't this just saying the same thing as the "wire format" part above?

mmalerba:
done

src/lib/core/datetime/date-adapter.ts
jelbourn: Yeah, but it's a bit more explicit (less jargony?) than "wire format"

jelbourn:
Caretaker note: will need to update APIs for folks that have their own adapter (at least one).

src/lib/core/datetime/date-adapter.ts
mmalerba: Ok, re-worded it to say backend

mmalerba:
Instantly throwing when deserialization fails may be a little harsh. I think I'll rework this to return an invalid date or something

mmalerba:
@jelbourn made some changes, PTAL

Elkrival:
@mmalerba Thank you for the adapter example code, it asserts the methods to implementing MomentJS to the Datepicker Module. 

I am adding the DatePickerModule to my current project, when implementing the new changes from the DateAdapter<Moment> example code, I keep running into an  error in line 196 where super.deserialize(value) is not found in the Moment export instance. As a workaround I reverted back to the code before the deserialize() function was added. 

A thousand thank you's for the example, we love the material2 library, and my designer friend Chandan loves the Material Design standard.

PR #7363, opened 2017-09-27
fix(select): prevent nbsp from getting butchered in AOT

mmalerba:


crisbeto:
Why did we need this in the first place? I don't really remember. I know that we had to do something similar for the input.

mmalerba:
In this case I needed to create a `&nbsp;` but due to a bug in the whitespace stripping it was getting removed, so I did it as a string binding instead. But it seems like AOT translates `'\xa0'` to just `'xa0'`, fortunately it correctly translates `'\u00A0'`.

jelbourn:
Angular 4.4.4 is supposed to be out tomorrow (barring unforeseen complications) with a fix that will just make `&nbsp;` work

PR #7364, opened 2017-09-27
fix(tabs): disable without using `pointer-events`

### LINKED ISSUE #6755: tooltip on md-tab not working
bug

tooltips should be shown on md-tabs

tooltips are not shown

see plunkr: http://plnkr.co/edit/ij8DMEKht4QTvQt1ANLN?p=preview

- _tab 2 and tab 3 have both tooltips which are not shown_
- _general tooltips work (see on the bottom of the plunkr a working tooltip)_

Providing information about tabs would be nice, especially when tabs are set as `disabled`, a reason why it is disabled should be somehow shown to the customer.
There is already a discussion about that on angularjs and old material, so we don't have to discuss the reasons again: https://github.com/angular/material/issues/1667

material 1.0.0 beta 10

if it is intended, what would be a good way to make the user know that this tab is disabled because of reasons (except doing nasty stuff like `<md-tab label="tab2 (not enabled feature)">...</md-tab>`

josephperrott:
Fixes #6755

### DIFF
--- a/src/lib/tabs/_tabs-common.scss
+++ b/src/lib/tabs/_tabs-common.scss
@@ -24,7 +24,6 @@ $mat-tab-animation-duration: 500ms !default;
 
   &.mat-tab-disabled {
     cursor: default;
-    pointer-events: none;
   }
 }
 
--- a/src/lib/tabs/tab-group.html
+++ b/src/lib/tabs/tab-group.html
@@ -6,15 +6,15 @@
   <div class="mat-tab-label" role="tab" matTabLabelWrapper mat-ripple
        *ngFor="let tab of _tabs; let i = index"
        [id]="_getTabLabelId(i)"
-       [tabIndex]="selectedIndex == i ? 0 : -1"
+       [attr.tabIndex]="_getTabIndex(tab, i)"
        [attr.aria-controls]="_getTabContentId(i)"
        [attr.aria-selected]="selectedIndex == i"
        [class.mat-tab-label-active]="selectedIndex == i"
        [disabled]="tab.disabled"
-       [matRippleDisabled]="disableRipple"
-       (click)="tabHeader.focusIndex = selectedIndex = i">
+       [matRippleDisabled]="tab.disabled || disableRipple"
+       (click)="_handleClick(tab, tabHeader, i)">
 
-    <!-- If there is a label template, use it. -->
+       <!-- If there is a label template, use it. -->
     <ng-template [ngIf]="tab.templateLabel">
       <ng-template [cdkPortalHost]="tab.templateLabel"></ng-template>
     </ng-template>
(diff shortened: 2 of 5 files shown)


src/lib/tabs/_tabs-common.scss
```diff
@@ -24,7 +24,10 @@ $mat-tab-animation-duration: 500ms !default;
 
   &.mat-tab-disabled {
     cursor: default;
-    pointer-events: none;
+
+    &:focus {
```
crisbeto: Fading out the element only on focus seems a little weird. Wouldn't it be better if we prevented focus in the first place? That's what the `pointer-events` approach was doing.

src/lib/tabs/tab-group.html
```diff
@@ -11,8 +11,8 @@
        [attr.aria-selected]="selectedIndex == i"
        [class.mat-tab-label-active]="selectedIndex == i"
        [disabled]="tab.disabled"
-       [matRippleDisabled]="disableRipple"
-       (click)="tabHeader.focusIndex = selectedIndex = i">
+       [matRippleDisabled]="tab.disabled || disableRipple"
+       (click)="!tab.disabled && tabHeader.focusIndex = selectedIndex = i">
```
crisbeto: This expression looks a bit weird. Considering moving it into a function?

src/lib/tabs/_tabs-common.scss
josephperrott: By disabling `pointer-events` entirely things like tooltips are prevented from being attached to a `<mat-tab-label>`

josephperrott:
@crisbeto Whenever you have a change, please take another look.

I realized that the issue we were running into was that we were adding `tabindex=-1` to all of the tabs that were not active in an attempt to prevent focusing on them.  However this meant that they could still be focused by mouse.  Instead we will use `[attr.tabIndex]="selectedIndex == i ? 0 : null"` as when `null` is the value, `tabindex` is removed from the element entirely.

src/lib/tabs/tab-group.ts
```diff
@@ -290,4 +291,12 @@ export class MatTabGroup extends _MatTabGroupMixinBase implements AfterContentIn
     this._tabBodyWrapperHeight = this._tabBodyWrapper.nativeElement.clientHeight;
     this._renderer.setStyle(this._tabBodyWrapper.nativeElement, 'height', '');
   }
+
+  /** Handle click events, setting new selected index if appropriate. */
+  _handleClick(tab: MatTab, tabHeader: MatTabHeader, idx: number) {
+    if (!tab.disabled) {
+      tabHeader.focusIndex = idx;
+      this.selectedIndex = idx;
+    }
```
crisbeto: Optionally this could be reduced to: `this.selectedIndex = tabHeader.focusIndex = idx`.

kara:
@josephperrott Check out CI failures?

josephperrott:
@kara CI failures are fixed.

PR #7365, opened 2017-09-27
fix(autocomplete): error when closing from a destroyed view

### LINKED ISSUE #7315: Error closing the panel of an autocomplete inside a destroyed view.
Bug

Should not throw any error

Error being thrown when attempting to close the panel of an autocomplete that is inside a destroyed view.

[Plunker Link](https://plnkr.co/edit/GX0Shi41PnppvUYS7HwO?p=preview)

**Error Scenario**
Step 1: Click on the button to open the view
Step 2 : Click on the AutoComplete but **do not close the Panel**
Step 3: Click on the button to close the view

**Success Scenario**
Step 1: Click on the button to open the view
Step 2 : Click on the AutoComplete
Step 2a : Close the Autocomplete Panel
Step 3: Click on the button to close the view

Should not throw error

Material Beta 11

Related to (#5446) (880e6d5), #5413

crisbeto:
Fixes an error that was being thrown, because the autocomplete tries to run change detection on a destroyed view.

Fixes #7315.

**Note:** The repro case is somewhat convoluted and while I could reproduce the issue in the demo app, I had a hard time getting it to throw the error during unit tests. I added a comment instead since the logic is pretty straightforward.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -118,6 +118,7 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   private _overlayRef: OverlayRef | null;
   private _portal: TemplatePortal<any>;
   private _panelOpen: boolean = false;
+  private _componentDestroyed = false;
 
   /** Strategy that is used to position the panel. */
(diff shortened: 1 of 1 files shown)


jelbourn:
@crisbeto just needs rebase, add merge-ready when ready

PR #7371, opened 2017-09-27
chore: add selection list to prerender check; autocomplete test stability improvements

crisbeto:
* Adds the selection list component to the kitchen sink.
* Piggy-backing off of the smaller fix, this is something I've meant to do for a while, but it's so minor that it doesn't make sense to open a separate PR. Currently there are a few unit tests that will cause Karma to hang until it times out, if they fail. This seems to be because Karma fails to stringify the `activeOption`. These changes switch the tests to cast the expected result to a boolean first.


jelbourn:
@crisbeto needs rebase

PR #7372, opened 2017-09-27
fix: replace extendObject utility w/ object spread

jelbourn:
BREAKING CHANGE: The `extendObject` utility function is removed.


jelbourn:
Added to lint

lucasvanhalst:
@jelbourn FYI you could use the `prefer-object-spread` rule instead of `ban`

PR #7375, opened 2017-09-27
fix(sort): style changes to fix IE

### LINKED ISSUE #7304: Sort arrow pieces appear horizontally rather than vertically.
Bug

Arrow pieces look like an arrow.

arrow pieces appear horizontally next to one another rather than vertically.

I haven't been able to reproduce it with a simple test case (and can't upload my regular site), but with the patch from https://github.com/angular/material2/pull/6802 in place (beta 11), I'm seeing the sort header "stem" and indicator appearing horizontally next to one another rather than vertically. Looks something like | ^

I was able to fix this by (with the debugger) adding a "flex-direction: column" style to the "mat-sort-header-arrow" class.

Would it be possible to do so in the official release?

material beta 11

andrewseguin:
Thanks IE for keeping me accountable and specific about my styles. 

Fixes #7304

### DIFF
--- a/src/lib/sort/sort-header.scss
+++ b/src/lib/sort/sort-header.scss
@@ -54,6 +54,7 @@ $mat-sort-header-arrow-transition: 225ms cubic-bezier(0.4, 0, 0.2, 1);
   align-items: center;
   position: absolute;
   top: 0;
+  left: 0;
   transition: $mat-sort-header-arrow-transition;
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/sort/sort-header.scss
```diff
@@ -72,6 +73,7 @@ $mat-sort-header-arrow-transition: 225ms cubic-bezier(0.4, 0, 0.2, 1);
   height: $mat-sort-header-arrow-thickness;
   transition: $mat-sort-header-arrow-transition;
   position: absolute;
+  top: 0rem;
```
crisbeto: `0rem` is the same as just `0`.

andrewseguin: Oh goodness, I think this was auto-completed IntelliJ. Thanks for catching that

PR #7376, opened 2017-09-27
fix(overlay): make config immutable for existing refs

jelbourn:
This changes makes the OverlayConfig instance in OverlatRef immutable.
Calling `getConfig` will now return a clone of the config to make clear
that it cannot be modified directly to change the state of the overlay.
This also updates the `updateSize` method to accept a partial config to
apply to the existing config (and make a new config instance)

BREAKING CHANGE: OverlayRef.getConfig returns a clone of the config
object.

BREAKING CHANGE: OverlayRef.updateSize now accepts a OverlaySizeConfig
rather than being based on the existing config object.

### DIFF
--- a/src/cdk/overlay/overlay-ref.ts
+++ b/src/cdk/overlay/overlay-ref.ts
@@ -6,14 +6,20 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {NgZone} from '@angular/core';
-import {PortalOutlet, Portal} from '@angular/cdk/portal';
-import {OverlayConfig} from './overlay-config';
-import {OverlayKeyboardDispatcher} from './keyboard/overlay-keyboard-dispatcher';
+import {Direction} from '@angular/cdk/bidi';
+import {ComponentPortal, Portal, PortalOutlet, TemplatePortal} from '@angular/cdk/portal';
+import {ComponentRef, EmbeddedViewRef, NgZone} from '@angular/core';
 import {Observable} from 'rxjs/Observable';
-import {Subject} from 'rxjs/Subject';
 import {take} from 'rxjs/operators/take';
+import {Subject} from 'rxjs/Subject';
+import {OverlayKeyboardDispatcher} from './keyboard/overlay-keyboard-dispatcher';
+import {OverlayConfig} from './overlay-config';
 
+
+/** An object where all of its properties cannot be written. */
+export type ImmutableObject<T> = {
+  readonly [P in keyof T]: T[P];
+};
 
 /**
  * Reference to an overlay that has been created with the Overlay service.
@@ -31,7 +37,7 @@ export class OverlayRef implements PortalOutlet {
   constructor(
       private _portalOutlet: PortalOutlet,
       private _pane: HTMLElement,
-      private _config: OverlayConfig,
+      private _config: ImmutableObject<OverlayConfig>,
       private _ngZone: NgZone,
       private _keyboardDispatcher: OverlayKeyboardDispatcher) {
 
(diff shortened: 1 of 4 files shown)


crisbeto:
Rather than cloning and ensuring that everything is immutable at runtime, I think that we can get away with making it immutable in TS. We can add a `Readonly` type and return `Readonly<OverlayRef`>`. The type looks something like this:

```ts
type Readonly<T> = {
  readonly [P in keyof T]: T[P];
};
```

jelbourn:
Added an `ImmutableObject` type (I avoided `Readonly` since that's already a separate thing in typescript). This actually caught a place in overlay-directives where we were directly writing the direction.

crisbeto:
Yeah I forgot to mention that there may be some places that depend on being able to update properties dynamically (e.g. I think the overlay directives do it in case the direction changed between attachments).

src/cdk/overlay/overlay-ref.ts
```diff
@@ -6,12 +6,16 @@
...
-import {PortalHost, Portal} from '@angular/cdk/portal';
-import {OverlayConfig} from './overlay-config';
+import {Direction} from '@angular/cdk/bidi';
+import {ComponentPortal, Portal, PortalHost, TemplatePortal} from '@angular/cdk/portal';
+import {ComponentRef, EmbeddedViewRef, NgZone} from '@angular/core';
 import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
+import {OverlayConfig} from './overlay-config';
 
+export type ImmutableObject<T> = {
```
crisbeto: Add a description here?

src/cdk/overlay/overlay-ref.ts
```diff
@@ -127,73 +135,80 @@ export class OverlayRef implements PortalHost {
...
   }
 
-  /**
-   * Gets the current state config of the overlay.
-   */
-  getState(): OverlayConfig {
-    return this._state;
+  /** Gets the the current overlay configuration, which is immutable. */
+  getConfig(): ImmutableObject<OverlayConfig> {
+    // Clone the config so that it cannot be modified outside of this class.
```
crisbeto: It's not cloning it anymore.

src/cdk/overlay/overlay-ref.ts
```diff
@@ -127,73 +135,80 @@ export class OverlayRef implements PortalHost {
...
 
+  /** Update the size properties of the overlay. */
+  updateSize(sizeConfig: OverlaySizeConfig) {
+    this._config = {...this._config, ...sizeConfig};
+    this._updateElementSize();
+  }
+
+  /** Sets the LTR/RTL direction for the overlay. */
+  setDirection(dir: Direction) {
+    this._config = {...this._config, direction: dir};
```
crisbeto: Why clone the config here instead of just setting the direction?

src/cdk/overlay/overlay-ref.ts
jelbourn: Done

src/cdk/overlay/overlay-ref.ts
jelbourn: Done

src/cdk/overlay/overlay-ref.ts
jelbourn: Because the config is immutable

crisbeto: Shouldn't it be enough to clone it on init and then work off the clone?

jelbourn: I like having the config as immutable all the time because it reinforces the fact that something extra needs to be done to update it.

PR #7385, opened 2017-09-28
feat(menu): support typeahead focus

crisbeto:
Adds support for focusing menu items via the character keys.

**Note:** I'm not sure whether using all top-level text nodes to determine the menu item text isn't too big of an assumption (e.g. it won't be accurate for the case where all the content is wrapped in a `strong` tag). Alternatively I can add a special case for `md-icon` or just take the `textContent` verbatim.

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -150,7 +150,7 @@ export class MatMenu implements AfterContentInit, MatMenuPanel, OnDestroy {
     @Inject(MAT_MENU_DEFAULT_OPTIONS) private _defaultOptions: MatMenuDefaultOptions) { }
 
   ngAfterContentInit() {
-    this._keyManager = new FocusKeyManager<MatMenuItem>(this.items).withWrap();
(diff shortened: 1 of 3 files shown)


julianobrasil:
@crisbeto, any chance to do something like this on mat-select too?

crisbeto:
@julianobrasil this has been available on the select for a while now.

jelbourn:
[The W3C authoring practices](https://www.w3.org/TR/wai-aria-practices-1.1/#menubutton) don't mention this behavior for menus- does this come from a specific resource?

crisbeto:
Not from a specific resource @jelbourn, but it's something I've noticed being done natively on MacOS. Windows also has something similar, although it executes the item action as well.

jelbourn:
I sent off a question to our internal a11y folks to see if there's any reason to avoid this behavior.

jelbourn:
Feedback is that it is good to add.

jelbourn:
Just needs rebase

PR #7397, opened 2017-09-28
style(ripple): start fade out on mouseup

willshowell:
**Purpose**

The goal of this is to start the ripple fade out immediately after mouseup rather than having to wait until the ripple animation has completed. My intention was to match the behavior seen in md1 which feels snappier and cleaner.

**Background**

The change to make sure a ripple had completed its animation before hiding was made in #3400. It looks like the intention was to solve the problem of ripples not being added to the active ripple set until _after_ it was completed, as pointed out in #3398. However, it appears #3400 fixed that problem [anyway](https://github.com/angular/material2/pull/3400/files#diff-9bf3861c94a2bdcf0beae13a7f4b4c0eR104) by adding it to the active ripple set as soon as animation begins. Therefore, the additional check of making sure a ripple had completed its animation is not necessary.

**Before**

ripple-b

**After**

ripple - a

This is one that's been a nit of mine for a while, but I couldn't tell if the change a while back was intentional or accidental. @DevVersion, thoughts? If this ok, I'll add tests.

### DIFF
--- a/src/lib/core/ripple/_ripple.scss
+++ b/src/lib/core/ripple/_ripple.scss
@@ -24,7 +24,8 @@ $mat-ripple-color-opacity: 0.1;
     border-radius: 50%;
     pointer-events: none;
 
-    transition: opacity, transform 0ms cubic-bezier(0, 0, 0.2, 1);
+    transition: opacity 0ms cubic-bezier(0.4, 0.0, 1, 1),
+                transform 0ms cubic-bezier(0, 0, 0.2, 1);
     transform: scale(0);
   }
 }
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -191,9 +191,9 @@ export class RippleRenderer {
 
     this._isMousedown = false;
 
-    // Fade-out all ripples that are not persistent.
+    // Fade-out all ripples that are completely visible and not persistent.
     this._activeRipples.forEach(ripple => {
-      if (!ripple.config.persistent) {
+      if (!ripple.config.persistent && ripple.state === RippleState.VISIBLE) {
         ripple.fadeOut();
       }
     });
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -191,9 +191,9 @@ export class RippleRenderer {
 
     this._isMousedown = false;
 
-    // Fade-out all ripples that are completely visible and not persistent.
+    // Fade-out all ripples that are not persistent.
     this._activeRipples.forEach(ripple => {
-      if (!ripple.config.persistent && ripple.state === RippleState.VISIBLE) {
+      if (!ripple.config.persistent) {
         ripple.fadeOut();
       }
     });


DevVersion:
I'm not sure which behavior we prefer here. For me (personally) it feels nicer if the ripples are finishing their animation. This is can be seen on normal `mat-button` elements in the demo-app.

When looking at [MDC](https://material-components-web.appspot.com/ripple.html), we can see that the ripples are also finishing their animation before starting to fade out.

In general, MDC varies in the implementation of the ripples, but a few aspects from the UX seem to be common enough to follow over all those implementations. I'm up for more discussion about this.

willshowell:
Yeah I suppose it's very much a personal preference. My concern when the current approach is that it feels quite piece-wise, where the expanding and fadeout are two distinctly separate animations. I'd prefer a smooth blending of the two.

When I look at [these examples](https://material.io/guidelines/motion/material-motion.html#) from the spec, it feels like the ripples are starting to fade out before reaching the boundaries. It can be seen best in [this animation](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B14F_FSUCc01aGFDUmZwNG1sTTQ/responsive-02-feedback-v2.webm). That said, there is a difference between

* Start fading out on `mouseup`
* Lower opacity as radius grows, to mimic ink spreading thinner

The second option is really what I'd prefer - not sure of the complexity.

willshowell:
I guess I'll also add,

One of the things that triggered me thinking about this was [this SO question](https://stackoverflow.com/questions/46227085/angular-24-material-animation-speed-customizing), and the link they provide to [material-ui](http://www.material-ui.com/#/components/radio-button). I can't help but prefer the smooth ink animations on the radio buttons, icon buttons, menu items, etc.

TheRedD0t:
After a look on the [MD guidelines/Buttons](https://material.io/guidelines/components/buttons.html) / [Behavior](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B3T7oTWa3HiFbjE5djFGMzdKcmc/components-buttons-flatbuttons.mp4), the buttons should have something like a fade out animation.

willshowell:
I see Google's calendar refresh uses buttons in line with the existing behavior. Maybe I'm behind on the times 😄.

I still want to play with this further and see if there's some tweaking left to do... maybe using radial-gradients in the ripple? or a different easing curve for opacity so there's not such a sharp distinction in the two transitions. Will report back.

willshowell:
@DevVersion thanks for keeping this open. I've tried adjusting the curves so that opacity and transform use opposing curves.

Transform still uses an [ease-out curve](http://cubic-bezier.com/#0,0,.2,1). But once complete, opacity uses an [ease-in curve](http://cubic-bezier.com/#.4,0,1,1). The idea being that the transition seamlessly moves from transform to opacity, without the sudden switch in velocity. I don't really know how else to explain it, so I've drawn a diagram 😄 



I know the difference is subtle, but I do feel like it's an improvement.

**Before**
ripple-before

**After**
ripple-after

DevVersion:
@willshowell No problem, thanks for your work on that detail. It's really hard to make sure we are following the exact specification in our case. The specifications for the ripple are very dry and don't contain a lot of details like that.

I'm not really sure which behavior is the right one, but I feel like in the specification videos from the Material Design guidelines the fade-out animation feels faster, and more aggressive than the one in your GIF.

https://material.io/guidelines/motion/material-motion.html#material-motion-how-does-material-move

I'm really up for discussion about this

TheRedD0t:
Hmmmmm I'm still of the opinion that the Ink Ripple doesn't always reach the frame of the button, just have a look on [this](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B14F_FSUCc01aGFDUmZwNG1sTTQ/responsive-02-feedback-v2.webm) animation and keep your eyes on the corners of the buttons, if the point of touch is at the opposite edge, the Ripple fades out before it reaches the end of the buttons. But this behavior seems only to exist, when the touch is a short one, I guess.

However, I agree with @DevVersion, the animations in the specs really feels faster.

jelbourn:
@willshowell @DevVersion revisiting this- our ripple should follow the behavior of MDC web. What do they do?

willshowell:
@DevVersion I'll let you decide if you want to keep this open for discussion. I think https://github.com/angular/material2/pull/9253 will allow for an appropriate amount of flexibility, but perhaps in the future `RippleGlobalOptions` could have a "terminate ripple on mouseup" option...

DevVersion:
@willshowell As per MDC, the ripples should fade-out, once the fade-in animation of the ripples completed. 

This behavior is kind of different from implementation to implementation, and therefore it's hard to properly judge. However, MDC uses the same concept, and it kind of feels more friendly (IMO)

In regards to the option, I agree that it would be nice having such an option (`terminatOnMouseup`). Can you file an issue for discussion about this?

PR #7403, opened 2017-09-29
build(): fix source maps

kara:
Our source maps currently have incorrect mappings, which makes it difficult to debug in the browser and also skews results from code size analyzers like source-map-explorer.

This PR fixes our source maps by aligning the Material process to the angular/angular build process.  This means:

- Instead of downleveling to ES2015, creating a Rollup bundle, then transpiling that bundle down to ES5, we now downlevel to both ES2015 and ES5, then create separate Rollup bundles for each.
- We are removing sorcery until we can upgrade tsickle and ngc to latest. Our versions of these break when mapping all the way back to TS.  (temporary)
- We have to add separate tsconfigs for transpiling to ES5 because our version of ngc doesn't allow overrides. When we update to compiler-cli 5.x, we should be able to remove these files. (temporary)

This PR also does some housekeeping:

- Renames `public_api` files to `public-api`
- Moves module code in cdk out into `-module.ts` files


DevVersion:
@jelbourn Why do we switch to `public-api`? I think it would be better to stay consistent with the Angular repository, because once we switch to Bazel we would have a similar setup.

kara:
@jelbourn LGTM-ed on slack since he is traveling

PR #7433, opened 2017-09-30
docs(radio-button): add a bit of padding to radio button example

alyssamichelle:
A spoonful of padding helps the `radio-buttons` go down! ☂️ 

**Before:**
image
image

**After:**
image
image

### DIFF
--- a/src/material-examples/radio-overview/radio-overview-example.css
+++ b/src/material-examples/radio-overview/radio-overview-example.css
@@ -1,4 +1,3 @@
-.mat-radio-button:not(:last-child) {
-  /*double the label-content padding-left*/
+.mat-radio-button ~ .mat-radio-button {
   padding-right: 16px;
 }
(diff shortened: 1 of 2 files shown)


src/material-examples/radio-overview/radio-overview-example.css
```diff
@@ -1 +1,4 @@
-/** No CSS for this example */
+.mat-radio-button:not(:last-child) {
```
crisbeto: This could be a little more robust as:

```css
.mat-radio-button + .mat-radio-button {
  margin-left: 16px;
}
```

Also the comment isn't really necessary.

alyssamichelle: @crisbeto I think the way I did it allows for more flexibility. If someone has an element other than a `.mat-radio-button` as a sibling, then I think yours breaks down; whereas with my way, you could have a radio button, some other element, and then another radio button and mine would still put a bit of space. 

That being said, I think your way might be faster. What do you think about doing `.mat-radio-button ~ .mat-radio-button`? Using this should allow us to select all the radio button siblings, no matter if they are immediately adjacent or not and would still be a little faster than `:not()`.

crisbeto: That works as well. My only issue with the first approach was it could've broken if we shuffled the DOM around.

PR #7446, opened 2017-09-30
fix(ripple): use element coordinates instead of page coordinates

### LINKED ISSUE #7436: Ripple Effect in Button in Dialog has wrong position
Bug

The ripple should originate from the mouse cursor.

The ripple effect sometimes does not originate from the mouse cursor, but instead, somewhere above the mouse cursor. 

Replicated here: https://embed.plnkr.co/HzjBCx/
1. Scroll to the bottom of the output
2. Click on either "pick one" button
3. Click on the "example" button to observe the incorrect ripple effect; it originates from above the button.
4. Scroll back up to the top of the page and do the same thing; the ripple effect is working.

Ripple effect should probably work in this scenario

Not sure which versions of everything are effected.  The plunkr posted above is using TypeScript 2.1.1 but I'm not sure which version of Material its using.

crisbeto:
Uses the `clientX` and `clientY` coordinates when creating a ripple, instead of taking `pageX`/`pageY` and subtracting the scroll position since the result is exactly the same. Using the client coordinates has the advantage of being simpler and not having to read the scroll position.

Fixes #7436.

### DIFF
--- a/src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
+++ b/src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
@@ -10,7 +10,6 @@ import {Directionality} from '@angular/cdk/bidi';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {Platform} from '@angular/cdk/platform';
 import {auditTime, takeUntil} from '@angular/cdk/rxjs';
-import {ViewportRuler} from '@angular/cdk/scrolling';
 import {
   AfterContentInit,
   ChangeDetectionStrategy,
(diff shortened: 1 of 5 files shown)


src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -10,7 +10,6 @@ import {Directionality} from '@angular/cdk/bidi';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {Platform} from '@angular/cdk/platform';
 import {auditTime, takeUntil} from '@angular/cdk/rxjs';
-import {ViewportRuler} from '@angular/cdk/scrolling';
```
DevVersion: We should also remove the `VIEWPORT_RULER_PROVIDER` in the ripple module actually.

crisbeto: Good call. Done.

PR #7449, opened 2017-09-30
fix(overlay): detach method returns undefined

### LINKED ISSUE #7408: An `undefined` value is returned by `OverlayRef.detach()`

Bug

```ts
export declare class OverlayRef implements PortalHost {
    /**
     * Detaches an overlay from a portal.
     * @returns Resolves when the overlay has been detached.
     */
    detach(): Promise<any>;
```

An `undefined` value is returned by `OverlayRef.detach()`

2017-09-29 10_04_41-developer tools - http___localhost_4200_

StackBlitz: https://angular-material2-issue-qqyacc.stackblitz.io/

StackBlitz export: [angular-material2-issue.zip](https://github.com/angular/material2/files/1343620/angular-material2-issue.zip)

The plunker starter btw errors with no modifications:

2017-09-29 10_17_11-developer tools - http___plnkr co_edit_o077b6ueiiigkc0s06dd_p preview

Is it a bug?

`@angular/cdk: 2.0.0-beta.11`

DevVersion:
Currently the JSDoc of the `detach` method describes that the method returns a Promise if the detaching is done. This is not the case and just `undefined` is being returned because every `PortalHost` implementation inside of the CDK is synchronously detaching the view.

In the future it might be useful to change this method back to a `Promise` in favor of animations, but for now it doesn't make sense return a `Promise` because the `attach` method is also synchronous.

Fixes #7408

### DIFF
--- a/src/cdk/overlay/overlay-ref.ts
+++ b/src/cdk/overlay/overlay-ref.ts
@@ -85,9 +85,9 @@ export class OverlayRef implements PortalHost {
 
   /**
    * Detaches an overlay from a portal.
-   * @returns Resolves when the overlay has been detached.
+   * @returns The portal detachment result.
    */
-  detach(): Promise<any> {
+  detach(): any {
     this.detachBackdrop();
 
     // When the overlay is detached, the pane element should disable pointer events.
@@ -99,7 +99,7 @@ export class OverlayRef implements PortalHost {
       this._config.scrollStrategy.disable();
     }
 
-    let detachmentResult = this._portalHost.detach();
+    const detachmentResult  = this._portalHost.detach();
 
     // Only emit after everything is detached.
     this._detachments.next();


src/cdk/overlay/overlay-ref.ts
```diff
@@ -83,11 +83,8 @@ export class OverlayRef implements PortalHost {
     return attachResult;
   }
 
-  /**
-   * Detaches an overlay from a portal.
-   * @returns Resolves when the overlay has been detached.
-   */
-  detach(): Promise<any> {
+  /** Detaches an overlay from a portal. */
+  detach(): void {
```
crisbeto: Change the `PortalHost` signature as well?

DevVersion: Not sure about this. I thought that the `PortalHost` can still have the type in case someone wants to implement his own `PortalHost`.

crisbeto: But since you're removing the return value from the `OverlayRef`, it means that even if they did, they wouldn't be able to use it with our overlays anyway.

DevVersion: Portals don't have to be used specifically with overlays.

crisbeto: They don't but that's the primary use case and it's something that we'll end up not supporting. Since the initial goal of the PR was to get rid of the `Promise` signature, consider re-adding the return and setting the return value to `any`?

DevVersion: That would work too. In that case we would have a very unclear method signature and the return value would be for now always `undefined`. 

Since overlays always seem to use the `DomPortalHost` which seems to return nothing on `detach` I thought we can be clear with this signature here.

I'm fine switching to `any` though.

DevVersion: Updated as discussed on Slack.

Issue #6372, opened 2017-08-09
[audit] overlay extra classes 


willshowell:
#### Bug, feature request, or proposal:
Suggesting library-wide audit

#### Currently

Overlay components all accept extra classes differently:

* Dialog: `string | string[]`
* Snackbar: `string[]`
* Tooltip: `NgClass`
* Select: `NgClass`
* Menus: `class` only
* Datepicker: none
* Autocomplete: none (tracking https://github.com/angular/material2/pull/6618 or https://github.com/angular/material2/pull/7176)

#### Ideally

Consistent NgClass compatibility would be awesome. If not possible, `string | string[]` support would be good.

Additionally, consistent naming would be nice. Snackbar uses `extraClasses`, whereas select uses `panelClass`. Also, select and dialog both use `panelClass`, but the class is added to different elements.

jelbourn:
Agreed, they should all be consistent. I would want to use `panelClass` everywhere

amcdnl:
@jelbourn - So to clarify, update all the inputs to `panelClass` with `NgClass` as the interface? Do we need to deprecate as well?

jelbourn:
The template-based components can have `NgClass` mechanics, but dialog.snack-bar would just stay as strings because there's no binding context

jelbourn:
Also this would be a good candidate for a mixin

amcdnl:
@jelbourn -

1) I played around with a mixin but the implementations varied so much, it was tricky. I ended up making a utility parser instead.
2) Do you want me to rename `matTooltipClass` to `matPanelClass` or just leave it for that one?

jelbourn:
Tooltip can stay as `matTooltipClass`

jelbourn:
If you make a PR it's easier to leave comments; you can mark it as WIP

jelbourn:
One thing I will say is that we don't want to add additional cdk packages for utility methods; each one is a public API that a user would potentially need to know about (such as when adding to a systemjs config)

josh18:
Would it make sense (maybe in another issue / PR) to add a similar input for the `.cdk-overlay-container` element?

Issue #6206, opened 2017-08-01
Add support for hiding a dialog without closing it


superMDguy:
#### Bug, feature request, or proposal:

Only show one dialog at a time.

#### What is the expected behavior?

When a dialog is opened from within another dialog, all other open dialogs should be hidden.

#### What is the current behavior?

The new dialog gets layered on top of the old dialog(s)

#### What are the steps to reproduce?

1. Open a dialog
2. Open another dialog from within that dialog

Plunker: https://plnkr.co/edit/GglOLDnx1uJlZRgD4y1W?p=preview

#### What is the use-case or motivation for changing an existing behavior?
In angularjs material, multiple dialogs weren't allowed by default.  See, for example http://plnkr.co/edit/qNu3q5XsIie1fYjKeJRc?p=preview.  Ideally, though, there'd be a virtual stack of dialogs maintained, where only the topmost item on the stack is visible.

#### Is there anything else we should know?

Currently, I can work around the issue by running

`Array.from(document.querySelectorAll(".cdk-overlay-container .cdk-global-overlay-wrapper"))
      .forEach((node, index, array) => (index !== array.length - 1) ? (<HTMLElement> node).style.display = "none" : true);`

in ngOnInit, and running

`Array.from(document.querySelectorAll(".cdk-overlay-container .cdk-global-overlay-wrapper"))
      .forEach((node) =>(<HTMLElement> node).style.display = "" );`

in ngOnDestroy.

This works, and effectively creates a stack of hidden dialogs.  But I don't want to add that snippet in every dialog that could possibly be opened from another dialog.

jelbourn:
We intentionally support opening multiple dialogs at the same time. 

What is missing, however, is a way to _hide_ a dialog without closing it when you only have a DialogRef. Going to re-purpose this issue for that feature.

jelbourn:
We could potentially do this by adding a `hide` API to `MdDialogRef`, but I think that this introduces too much confusion between `close` and `hide`. Instead, I think it would make sense to add the ability to add/remove css classes to the dialog via `MdDialogRef`.

superMDguy:
How about a way to set an "inactive" class on the overlay pane that's automatically applied to all but the topmost dialog?

julianobrasil:
If the default behavior is to hide the dialog, maybe the `hide` API isn't so confusing. An attribute in the parameter object passed to the open function would be simple to use. It would be something like 

```js
dialog.open({
  cascadeHide: false 
});
```

[...]

PR #7454, opened 2017-10-01
chore(api): overlay class audits #6372

amcdnl:
This PR does the following:

- Adds a new utility method for parsing string|string[]|{key/value} to ngClass-like object
- Adds a new utility method for parsing a ngClass-like object to string array
- Updates overlay implementation to accept ngClass-like object
- Adds `panelClass` to date picker
- Deprecates `extraClasses` to `panelClass` in snackbar
- Deprecates `class` input to `panelClass` and accepts ngClass-like object in menu
- Updates dialog implementation to ngClass like object

I really didn't like the fact I had to add this to the CDK. I originally had it as a utility inside material but the overlay lives in the CDK and needed it there too. @jelbourn - do you have any thoughts on better place for this?

This is still very WIP.


src/cdk/css/css.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+ /* Parses arguments into a ngClass-like object */
+export function parseClassList(classList:
```
crisbeto: I'm sure whether we need to support all of these formats. `NgClass` has them, presumably, because the expressions are dynamic and there a lot of ways to combine static classes and dynamic ones. In our case we only add the classes on init and we don't care about updating them, which can be done just as easily with an array.

amcdnl: The original ticket expressed desire to standardize this format. Ref: #6372

willshowell: FWIW NgClass seemed like a natural thing to support for template-based overlays like select, menu, autocomplete, tooltip. For dialog and snackbar, it's probably less important, but there are some good use cases for having the ability to dynamically update classes, https://github.com/angular/material2/issues/6206#issuecomment-319528256.

Further, I would personally find an NgClass cdk utility useful for custom overlay-based components!

crisbeto: Template-based overlays should "just work" as they are at the moment (you can stick an `NgClass` on an element inside the `TemplateRef`). This PR adds a similar signature to the programmatic API, but it's not _really_ `NgClass` because it won't update dynamically.

willshowell: Ok tbh, I hadn't actually read the pr 😄 . My initial concern in filing #6372 was how snackbar and dialog differed in support and naming. Combined with little type info available in the docs (as https://github.com/angular/material2/pull/6947 is trying to solve) it's hard to anticipate exactly what will happen just by looking at the docs.

I agree, `{[klass: string]: any}` is probably extraneous (and maybe misrepresenting) if classes don't update dynamically. While I still think that's a valid feature via #6206, maybe for another time.

amcdnl: Ya, so basically this PR:

- Unifies all the css class names to `panelClass`
- Normalizes all the signatures to `ngClass`-like

That means you can pass it an string, string array or object key/value pair. In the cases of the snackbar and overlay, there is no `ngClass` used so it just does a evaluate of the condition and matches it. I personally think the consistent signature is valuable and in the situation where there isn't real `ngClass` I could do an actual implementation I just didn't do it because I didn't know how far we wanted to take this.

@jelbourn - Can you advise on next steps with this? Currently the PR is just in a evaluation phase and I know you expressed concerns with adding another CDK lib here too so I didn't want to take it too far without first review.

amcdnl:
Closing in favor of #8056

PR #7456, opened 2017-10-01
feat(selection-list): support for ngModel

### LINKED ISSUE #6896: Add support for NgModel to selection-list

DevVersion:
* Adds support for NgModel to the selection-list.

Fixes #6896

### DIFF
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -4,7 +4,13 @@ import {createKeyboardEvent, dispatchFakeEvent} from '@angular/cdk/testing';
 import {Component, DebugElement} from '@angular/core';
 import {async, ComponentFixture, fakeAsync, inject, TestBed, tick} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
-import {MatListModule, MatListOption, MatSelectionList, MatSelectionListChange} from './index';
+import {
+  MatListModule,
+  MatListOption,
+  MatListOptionChange,
+  MatSelectionList,
+  MatSelectionListChange
+} from './index';
 import {FormControl, FormsModule, NgModel, ReactiveFormsModule} from '@angular/forms';
 
 describe('MatSelectionList without forms', () => {
@@ -85,16 +91,18 @@ describe('MatSelectionList without forms', () => {
 
     it('should emit a deprecated selectionChange event on the list option that got clicked', () => {
       const optionInstance = listOptions[2].componentInstance as MatListOption;
-      const selectionChangeSpy = jasmine.createSpy('selectionChange spy');
+      let lastChangeEvent: MatListOptionChange | null = null;
 
-      optionInstance.selectionChange.subscribe(selectionChangeSpy);
+      optionInstance.selectionChange.subscribe(ev => lastChangeEvent = ev);
 
-      expect(selectionChangeSpy).toHaveBeenCalledTimes(0);
+      expect(lastChangeEvent).toBeNull();
 
       dispatchFakeEvent(listOptions[2].nativeElement, 'click');
       fixture.detectChanges();
 
-      expect(selectionChangeSpy).toHaveBeenCalledTimes(1);
+      expect(lastChangeEvent).not.toBeNull();
+      expect(lastChangeEvent!.source).toBe(optionInstance);
+      expect(lastChangeEvent!.selected).toBe(true);
     });
 
     it('should be able to dispatch one selected item', () => {
--- a/src/lib/list/selection-list.spec.ts
+++ b/src/lib/list/selection-list.spec.ts
@@ -61,7 +61,7 @@ describe('MatSelectionList without forms', () => {
       });
     });
 
-    it('should not emit a change event if an option changed programmatically', () => {
+    it('should not emit a selectionChange event if an option changed programmatically', () => {
       spyOn(fixture.componentInstance, 'onValueChange');
 
       expect(fixture.componentInstance.onValueChange).toHaveBeenCalledTimes(0);
@@ -72,7 +72,7 @@ describe('MatSelectionList without forms', () => {
       expect(fixture.componentInstance.onValueChange).toHaveBeenCalledTimes(0);
     });
 
-    it('should not emit a change event if an option got clicked', () => {
+    it('should emit a selectionChange event if an option got clicked', () => {
       spyOn(fixture.componentInstance, 'onValueChange');
 
       expect(fixture.componentInstance.onValueChange).toHaveBeenCalledTimes(0);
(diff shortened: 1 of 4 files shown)


src/lib/list/selection-list.ts
```diff
@@ -274,6 +324,62 @@ export class MatSelectionList extends _MatSelectionListMixinBase
...
+    this.options.forEach(option => option._setSelected(false));
+
+    values
+      .map(value => this._getOptionFromValue(value))
+      .filter(option => !!option)
+      .forEach(option => option!._setSelected(true));
+  }
+
+  /** Returns the values of the selected options. */
+  private _getSelectedOptionValues(): string[] {
```
crisbeto: Consider inlining this since it's only used in one place and it's a one-liner?

src/lib/list/selection-list.ts
```diff
@@ -274,6 +324,62 @@ export class MatSelectionList extends _MatSelectionListMixinBase
...
+    return this.options.find(option => option.value === value);
+  }
+
+  /** Sets the selected options based on the specified values. */
+  private _setOptionsFromValues(values: string[]) {
+    this.options.forEach(option => option._setSelected(false));
+
+    values
+      .map(value => this._getOptionFromValue(value))
+      .filter(option => !!option)
```
crisbeto: Optional: if you want to be fancy, you can `.filter(Boolean)` here.

src/lib/list/selection-list.ts
```diff
@@ -274,6 +324,62 @@ export class MatSelectionList extends _MatSelectionListMixinBase
...
+    this._onChange = fn;
+  }
+
+  /** Implemented as part of ControlValueAccessor. */
+  registerOnTouched(fn: any): void {
+    this.onTouched = fn;
+  }
+
+  /** Returns the option with the specified value. */
+  private _getOptionFromValue(value: string): MatListOption | undefined {
```
crisbeto: Call this `_getOptionByValue`? Also it's used only in one place so it could be inlined.

src/lib/list/selection-list.ts
```diff
@@ -274,6 +324,62 @@ export class MatSelectionList extends _MatSelectionListMixinBase
...
+    }
+  }
+
+  /** Implemented as part of ControlValueAccessor. */
+  registerOnChange(fn: any): void {
+    this._onChange = fn;
+  }
+
+  /** Implemented as part of ControlValueAccessor. */
+  registerOnTouched(fn: any): void {
```
crisbeto: The `fn` here and in `registerOnChange` is a `() => any` AFAIK.

src/lib/list/selection-list.spec.ts
```diff
@@ -428,9 +464,165 @@ describe('MatSelectionList', () => {
...
+    });
+
+    it('should be able to set the value through the form control', () => {
+      expect(listOptions.every(option => !option.selected))
+        .toBe(true, 'Expected every list option to be unselected.');
+
+      fixture.componentInstance.formControl.setValue(['opt2', 'opt3']);
+      fixture.detectChanges();
+
+      expect(listOptions[1].selected).toBe(true, 'Expected option to be selected.');
```
crisbeto: The two `Expected option to be selected.` are super useful since they're right after each other. Consider changing them to something like `Expected second option to be selected.` and `Expected third option to be selected.`?

src/lib/list/selection-list.spec.ts
```diff
@@ -428,9 +464,165 @@ describe('MatSelectionList', () => {
...
+      fixture.detectChanges();
+
+      selectionListDebug = fixture.debugElement.query(By.directive(MatSelectionList));
+      selectionList = selectionListDebug.componentInstance;
+      ngModel = selectionListDebug.injector.get<NgModel>(NgModel);
+      listOptions = fixture.debugElement.queryAll(By.directive(MatListOption))
+        .map(optionDebugEl => optionDebugEl.componentInstance);
+    });
+
+    it('should update the model if an option got selected', fakeAsync(() => {
```
crisbeto: Switch the name here to "should update the model if an option got selected **programmatically**`? Otherwise it's very close to `should update the model if an option got clicked`.

src/lib/list/selection-list.ts
DevVersion: Sounds better. But I'd still like to keep it as a function. Otherwise it won't be that readable anymore.

src/lib/list/selection-list.ts
DevVersion: Nice idea.

src/lib/list/selection-list.spec.ts
DevVersion: True

src/lib/list/selection-list.ts
DevVersion: Yeah we can add types here.

src/lib/list/selection-list.ts
DevVersion: I'd personally prefer keeping this as a function. It makes it the code more sorted with the `_setOptionsFromValues` method.

src/lib/list/selection-list.spec.ts
```diff
@@ -428,9 +464,165 @@ describe('MatSelectionList', () => {
...
+    });
+
+    it('should be able to set the value through the form control', () => {
+      expect(listOptions.every(option => !option.selected))
+        .toBe(true, 'Expected every list option to be unselected.');
+
+      fixture.componentInstance.formControl.setValue(['opt2', 'opt3']);
+      fixture.detectChanges();
+
+      expect(listOptions[1].selected).toBe(true, 'Expected first option to be selected.');
```
crisbeto: These are wrong now. This is looking for the second option and the next one is looking for the third.

crisbeto:
No longer blocked but needs a rebase.

src/lib/list/selection-list.ts
```diff
@@ -131,7 +137,12 @@ export class MatListOption extends _MatListOptionMixinBase
 
   ngOnInit() {
     if (this.selected) {
-      this.selectionList.selectedOptions.select(this);
+      // List options that are selected at initialization can't be reported properly to the form
+      // control. This is because it takes some time until the selection-list knows about all
+      // available options. Also it can happen that the ControlValueAccessor has an initial value
+      // that should be used instead. Deferring the value change report to the next tick ensures
+      // that the form control value is not being overwritten.
+      Promise.resolve(() => this.selected && this.selectionList._reportValueChange());
```
jelbourn: @kara do you have any input on this?

jelbourn:
Running into some meatadata extractor issue with `NG_VALUE_ACCESSOR`, not sure exactly what's up yet

src/lib/list/selection-list.ts
```diff
@@ -203,26 +230,38 @@ export class MatListOption extends _MatListOptionMixinBase
...
+    CanDisable, CanDisableRipple, HasTabIndex, AfterContentInit, ControlValueAccessor {
 
   /** The FocusKeyManager which handles focus. */
   _keyManager: FocusKeyManager<MatListOption>;
 
   /** The option components contained within this selection-list. */
   @ContentChildren(MatListOption) options: QueryList<MatListOption>;
 
+  /** Emits a change event whenever the selected state of an option changes. */
+  @Output() change: EventEmitter<MatSelectionListChange> =
```
josephperrott: This actually needs to be `selectionChange` rather than `change`

Looks like it got inadvertently renamed in this PR.

DevVersion: I intentionally named it `change` to be consistent with the other form elements supporting the `ControlValueAccessor`. I can switch it to `selectionChange` or also add a deprecated `selectionChange` output.

jelbourn: Let's go with `selectionChange` since that's what the `mat-select` now uses

DevVersion:
@josephperrott @jelbourn Renamed the  `change` event to `selectionChange`. 

Also re-added the `selectionChang` event on the list option to avoid breaking changes. I marked that one as deprecated because we should encourage people to use the event on the list element instead.

Please have another look.

jelbourn:
Setting back to merge-ready.

@tinayuangao if this fails on presubmit, can you help @DevVersion resolve any issues so that this can land for RC3?

kwalski:
demo app/docs need to be updated to show usage of ngModel with selection-list

BamiGorengo:
is there any update for the documentation planed?

PR #7461, opened 2017-10-02
feat(expansion): add accordion expand/collapse all (#6929)

### LINKED ISSUE #6929: [mdExpansionPanel] Expand/collapse all functionality
Feature request

Expand all/collapse all

Unable to expand or collapse multiple expansion panels in a single operation, unless it is possible and I am missing something.

PS Thanks for the great effort so far - was super excited to discover that expansion panels have been added!

ouq77:
Add `expandAll()`/`collapseAll()` to perform expand/collapse all on multiple expandable accordions.

Closes #6929

### DIFF
--- a/src/cdk/accordion/accordion-item.ts
+++ b/src/cdk/accordion/accordion-item.ts
@@ -18,6 +18,7 @@ import {
 import {UniqueSelectionDispatcher} from '@angular/cdk/collections';
 import {CdkAccordion} from './accordion';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {Subscription} from 'rxjs/Subscription';
 
 /** Used to generate unique ID for each accordion item. */
 let nextId = 0;
@@ -31,6 +32,8 @@ let nextId = 0;
   exportAs: 'cdkAccordionItem',
 })
 export class CdkAccordionItem implements OnDestroy {
+  /** Subscription to openAll/closeAll events. */
+  private _openCloseAllSubscription = Subscription.EMPTY;
   /** Event emitted every time the AccordionItem is closed. */
   @Output() closed: EventEmitter<void> = new EventEmitter<void>();
   /** Event emitted every time the AccordionItem is opened. */
@@ -97,12 +100,18 @@ export class CdkAccordionItem implements OnDestroy {
           this.expanded = false;
         }
       });
+
+    // When an accordion item is hosted in an accordion, subscribe to open/close events.
+    if (this.accordion) {
+      this._openCloseAllSubscription = this._subscribeToOpenCloseAllActions();
+    }
   }
 
   /** Emits an event for the accordion item being destroyed. */
   ngOnDestroy() {
     this.destroyed.emit();
     this._removeUniqueSelectionListener();
+    this._openCloseAllSubscription.unsubscribe();
   }
 
   /** Toggles the expanded state of the accordion item. */
@@ -124,5 +133,14 @@ export class CdkAccordionItem implements OnDestroy {
     if (!this.disabled) {
       this.expanded = true;
     }
+  }
+
+  private _subscribeToOpenCloseAllActions(): Subscription {
+    return this.accordion._openCloseAllActions.subscribe(expanded => {
+      // Only change expanded state if item is enabled
+      if (!this.disabled) {
+        this.expanded = expanded;
+      }
+    });
   }
 }
--- a/src/cdk/accordion/accordion.ts
+++ b/src/cdk/accordion/accordion.ts
@@ -8,6 +8,7 @@
 
 import {Directive, Input} from '@angular/core';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {Subject} from 'rxjs/Subject';
 
 /** Used to generate unique ID for each accordion. */
 let nextId = 0;
@@ -20,6 +21,9 @@ let nextId = 0;
   exportAs: 'cdkAccordion',
 })
 export class CdkAccordion {
+  /** Stream that emits true/false when openAll/closeAll is triggered. */
+  readonly _openCloseAllActions: Subject<boolean> = new Subject<boolean>();
+
   /** A readonly id value to use for unique selection coordination. */
   readonly id = `cdk-accordion-${nextId++}`;
 
@@ -28,4 +32,20 @@ export class CdkAccordion {
   get multi(): boolean { return this._multi; }
   set multi(multi: boolean) { this._multi = coerceBooleanProperty(multi); }
   private _multi: boolean = false;
+
+  /** Opens all enabled accordion items in an accordion where multi is enabled. */
+  openAll(): void {
+    this._openCloseAll(true);
+  }
+
+  /** Closes all enabled accordion items in an accordion where multi is enabled. */
+  closeAll(): void {
+    this._openCloseAll(false);
+  }
+
+  private _openCloseAll(expanded: boolean): void {
+    if (this.multi) {
+      this._openCloseAllActions.next(expanded);
+    }
+  }
 }
--- a/src/demo-app/expansion/expansion-demo.ts
+++ b/src/demo-app/expansion/expansion-demo.ts
@@ -6,7 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Component, ViewEncapsulation} from '@angular/core';
+import {Component, ViewChild, ViewEncapsulation} from '@angular/core';
+import {MatAccordion} from '@angular/material';
 
 @Component({
   moduleId: module.id,
@@ -17,6 +18,8 @@ import {Component, ViewEncapsulation} from '@angular/core';
   preserveWhitespaces: false,
 })
 export class ExpansionDemo {
+  @ViewChild(MatAccordion) accordion: MatAccordion;
+
   displayMode: string = 'default';
   multi = false;
   hideToggle = false;
--- a/src/lib/expansion/accordion.ts
+++ b/src/lib/expansion/accordion.ts
@@ -33,7 +33,7 @@ export class MatAccordion extends CdkAccordion {
   /**
    * The display mode used for all expansion panels in the accordion. Currently two display
    * modes exist:
-   *   default - a gutter-like spacing is placed around any expanded panel, placing the expanded
+   *  default - a gutter-like spacing is placed around any expanded panel, placing the expanded
    *     panel at a different elevation from the reset of the accordion.
    *  flat - no spacing is placed around expanded panels, showing all panels at the same
    *     elevation.
--- a/src/demo-app/expansion/expansion-demo.html
+++ b/src/demo-app/expansion/expansion-demo.html
@@ -41,6 +41,11 @@
     <mat-radio-button value="default">Default</mat-radio-button>
     <mat-radio-button value="flat">Flat</mat-radio-button>
   </mat-radio-group>
+  <p>Accordion Actions <sup>('Multi Expansion' mode only)</sup></p>
+  <div>
+    <button mat-button (click)="accordion.openAll()" [disabled]="!multi">Expand All</button>
+    <button mat-button (click)="accordion.closeAll()" [disabled]="!multi">Collapse All</button>
+  </div>
   <p>Accordion Panel(s)</p>
   <div>
     <mat-checkbox [(ngModel)]="panel1.expanded">Panel 1</mat-checkbox>
--- a/src/lib/expansion/accordion.spec.ts
+++ b/src/lib/expansion/accordion.spec.ts
@@ -1,8 +1,8 @@
 import {async, TestBed} from '@angular/core/testing';
-import {Component} from '@angular/core';
+import {Component, ViewChild} from '@angular/core';
 import {By} from '@angular/platform-browser';
 import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
-import {MatExpansionModule} from './index';
+import {MatExpansionModule, MatAccordion} from './index';
 
 
 describe('CdkAccordion', () => {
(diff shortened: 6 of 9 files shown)


src/lib/expansion/accordion.ts
```diff
@@ -44,6 +48,26 @@ export class CdkAccordion {
    *     elevation.
    */
   @Input() displayMode: MatAccordionDisplayMode = 'default';
+
+  /**
+   * Expands all enabled expansion panels in an accordion where multi is enabled
+   */
+  expandAll() {
```
amcdnl: Add return type `void`

src/lib/expansion/accordion.ts
```diff
@@ -44,6 +48,26 @@ export class CdkAccordion {
...
+    this._expandCollapseAll(true);
+  }
+
+  /**
+   * Collapses all enabled expansion panels in an accordion where multi is enabled
+   */
+  collapseAll() {
```
amcdnl: Add return type `void`

src/lib/expansion/accordion.ts
```diff
@@ -44,6 +48,26 @@ export class CdkAccordion {
...
+    this._expandCollapseAll(false);
+  }
+
+  _expandCollapseAll(expanded: boolean) {
```
amcdnl: Add return type `void`

src/demo-app/expansion/expansion-demo.ts
```diff
@@ -1,14 +1,17 @@
...
+import {MatAccordion} from '@angular/material';
 
 @Component({
   moduleId: module.id,
   selector: 'expansion-demo',
   styleUrls: ['expansion-demo.css'],
   templateUrl: 'expansion-demo.html',
   encapsulation: ViewEncapsulation.None,
-  preserveWhitespaces: false,
+  preserveWhitespaces: true,
```
jelbourn: This needs to stay `false`

src/lib/expansion/accordion.ts
```diff
@@ -25,7 +26,10 @@ export class CdkAccordion {
   /** A readonly id value to use for unique selection coordination. */
   readonly id = `cdk-accordion-${nextId++}`;
 
-  /** Whether the accordion should allow multiple expanded accordion items simulateously. */
+  /** Stream that emits when expandAll or collapseAll is triggered. */
+  readonly expandCollapseAllSubject: Subject<boolean> = new Subject<boolean>();
```
jelbourn: Mark this with an underscore to denote that it is internal-only and add a JsDoc description (specifically that the `boolean` denotes expanded/collapsed)

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -101,6 +102,17 @@ export class MatExpansionPanelHeader implements OnDestroy {
     .subscribe(() => this._changeDetectorRef.markForCheck());
 
     _focusMonitor.monitor(_element.nativeElement, renderer, false);
+
+    // When a panel is hosted in an accordion, subscribe to the expand/collapse subject
+    if (this.panel.accordion) {
+      this._expandCollapseAllSubscription =
+        this.panel.accordion.expandCollapseAllSubject.subscribe((expanded: boolean) => {
```
jelbourn: The type for `expanded` doesn't need to be given here because it is inferred from the observable.

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -101,6 +102,17 @@ export class MatExpansionPanelHeader implements OnDestroy {
     .subscribe(() => this._changeDetectorRef.markForCheck());
 
     _focusMonitor.monitor(_element.nativeElement, renderer, false);
+
+    // When a panel is hosted in an accordion, subscribe to the expand/collapse subject
+    if (this.panel.accordion) {
+      this._expandCollapseAllSubscription =
```
jelbourn: Can you move the contents of this `if` block to a separate method? Something like `_subscribeToAccordionActions`

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -101,6 +102,17 @@ export class MatExpansionPanelHeader implements OnDestroy {
     .subscribe(() => this._changeDetectorRef.markForCheck());
 
     _focusMonitor.monitor(_element.nativeElement, renderer, false);
+
+    // When a panel is hosted in an accordion, subscribe to the expand/collapse subject
+    if (this.panel.accordion) {
```
josephperrott: By placing the logic for setting panel expansion state in the `MatExpansionPanelHeader` it ties the logic specifically to the Material implementation rather than the CDK base.

I would suggestion adding this logic to the `AccordionItem` class.

There you can set the expanded state directly.

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -101,6 +102,17 @@ export class MatExpansionPanelHeader implements OnDestroy {
...
+      this._expandCollapseAllSubscription =
+        this.panel.accordion.expandCollapseAllSubject.subscribe((expanded: boolean) => {
+          // Only toggle if current state is not the intended state
+          if (!!this._isExpanded() !== expanded) {
+            this._toggle();
+          }
```
josephperrott: Coupled with the move to `AccordionItem`, I would suggest just setting the expanded state directly rather than checking if a toggle needs to occur.


`this.panel.accordion.expandCollapseAllSubject.subscribe(expanded => this.expanded = expanded);`

src/lib/expansion/accordion.ts
```diff
@@ -44,6 +48,26 @@ export class CdkAccordion {
    *     elevation.
    */
   @Input() displayMode: MatAccordionDisplayMode = 'default';
+
+  /**
+   * Expands all enabled expansion panels in an accordion where multi is enabled
```
josephperrott: JsDoc comments should be complete sentences so add a `.` here?

ouq77:
Thanks for the feedback @amcdnl, @jelbourn and @josephperrott - will get this sorted asap.

Unable to figure out if the Travis Sauce Labs failures are just platform problems or issues with this ticket.

Also, do I need to take any action on the Screenshot Tests?

josephperrott:
@ouq77 We have been having some issues with our saucelab tests, I don't think its from you.  For the screenshot tests, no need to worrry about them unless they are we are seeing failures in the expansion panels.

Once we get to a LGTM state on this PR, I can make the screenshot tests as good, however I want to wait as it will reset after every new push to the branch.

src/lib/expansion/expansion-panel-header.ts
ouq77: I need access to the `disabled` state of the hosting panel to prevent disabled panels from expanded/collapsed, which is why I used `_toggle()`. Is there another way to get access to this property without too much refactoring?

ouq77: Think I may have solved it with `mixinDisabled` - let me know if that is not a desired solution though. Cheers.

josephperrott: We actually just removed the mixinDisabled from AccordionItem as we are moving it to CDK.  You can see the move in #7530 

I supposed you could bake the `expandAll`/`collapseAll` into the `AccordionItem`, and then in `MatExpansionPanel`, extend the setter of `expanded` to respect `disabled`.  This would allow the `expandAll`/`collapseAll` functionality to live in the CDK, but work as expected in `MatAccordion`.  Admittedly I haven't thought this idea entirely through and would need to consider the ramifications a bit more before saying its definitely the best option or the option to go with.

ouq77: Thanks, @josephperrott - I will wait for 7530 to be merged and then take it from there.

ouq77:
@josephperrott I attempted your suggestion of adding the functionality in the CDK, but couldn't find an elegant way of accessing the `disabled` property on the accordion item.

Since this is mainly a Material concern, leaving it in this layer is probably acceptable.

Let me know if this is not an ideal and I can rethink it.

Is there any chance that the `common-behaviours` will be moved to the CDK?

josephperrott:
I spoke with @jelbourn and there does not seem to be an issue with moving some of the `common-behaviors` into CDK.  `disabled` as an example would make sense to do so.   We might need to move `disabled` and `CanDisable` into CDK which would open up all of the logic for `expandAll`/`collapseAll` to be in CDK.

ouq77:
Is this something I can help with @josephperrott? Should probably be done against a separate issue?

josephperrott:
@ouq77 You definitely can move it to the CDK if you would like.  It should be done in another pull ideally.

jelbourn:
@josephperrott should this PR be revisited?

ouq77:
@josephperrott and @jelbourn - this PR is blocked until #8201 has been dealt with. I've been keeping that one up to date hoping it will be revisited soon. Cheers.

josephperrott:
PR is now unblocked with the merge of #8201

josephperrott:
With the `disabled` input now in `CdkAccordionItem`, is it possible to have the `expandAll`/`collapseAll` logic be in `CdkAccordion`?

ouq77:
Knew I forgot something - updated now :)

src/cdk/accordion/accordion.ts
```diff
@@ -28,4 +32,20 @@ export class CdkAccordion {
   get multi(): boolean { return this._multi; }
   set multi(multi: boolean) { this._multi = coerceBooleanProperty(multi); }
   private _multi: boolean = false;
+
+  /** Expands all enabled accordion items in an accordion where multi is enabled. */
+  expandAll(): void {
```
josephperrott: These should likely be `openAll` and `closeAll` to match the naming convention used in `CdkAccordionItem`.

ouq77: Done.

src/cdk/accordion/accordion-item.ts
```diff
@@ -125,4 +134,13 @@ export class CdkAccordionItem implements OnDestroy {
       this.expanded = true;
     }
   }
+
+  private _subscribeToExpandCollapseAllActions(): Subscription {
```
josephperrott: This method should likely be called `subscribetoOpenCloseAllActions()`

src/cdk/accordion/accordion.ts
```diff
@@ -28,4 +32,20 @@ export class CdkAccordion {
...
+  openAll(): void {
+    this._expandCollapseAll(true);
+  }
+
+  /** Closes all enabled accordion items in an accordion where multi is enabled. */
+  closeAll(): void {
+    this._expandCollapseAll(false);
+  }
+
+  private _expandCollapseAll(expanded: boolean): void {
```
josephperrott: This should be `_openCloseAll()`

src/cdk/accordion/accordion-item.ts
ouq77: Done.

src/cdk/accordion/accordion.ts
ouq77: Done

src/cdk/accordion/accordion.ts
```diff
@@ -28,4 +32,20 @@ export class CdkAccordion {
...
+    this._openCloseAll(true);
+  }
+
+  /** Closes all enabled accordion items in an accordion where multi is enabled. */
+  closeAll(): void {
+    this._openCloseAll(false);
+  }
+
+  private _openCloseAll(expanded: boolean): void {
+    if (this.multi) {
```
SchnWalter: There are same cases, even for `[multi]="false"`, where you don't care which `CdkAccordionItem` is open, you just want to make sure that all are closed and it would be nice to also cover that use case, but I'm not sure how this could be done without confusing developers.

ouq77: I guess you could do `this.multi || !expanded`? Will have to experiment.

SchnWalter: That would be a solution, but we could have a "cleaner" solution, if I may use this word.

This would be a good solution for this particular case. But maybe we could find a more generic solution that could be used in other cases and not confuse people to much, `openAll()` won't work in single mode, but we could have `open()` that works for single mode, opening the first item or in multi mode.

If you look at something like the `SelectionModel` for tables, select and chips, the selection has a `getMultipleValuesInSingleSelectionError`, we could use something like that. And `MatSelectionList.selectedOptions` exposes the model for anyone to use. But probably a selection model would be an overkill for this situation, I'm not sure in how many places we could reuse it.

Also, `MatDrawerContainer` container has a similar behavior but it doesn't add the `All` suffix and it uses `@ContentChildren` to keep a list of children and then call open/close on each children, instead of emitting events, here's a snippet from it:

```
  @ContentChildren(MatDrawer) _drawers: QueryList<MatDrawer>;

  /** Calls `open` of both start and end drawers */
  open(): void {
    this._drawers.forEach(drawer => drawer.open());
  }

  /** Calls `close` of both start and end drawers */
  close(): void {
    this._drawers.forEach(drawer => drawer.close());
  }
```
Using `ContentChildren` instead of emitting an event and rename to `open/close` in order to support "multi/single" mode, would be great.


P.S. Thanks for the work you did on this PR!

ouq77: This maybe warrants a separate issue to investigate?

pankajsri03:
I created a demo using the example in the demo-app , but  expand-all/collapse-all isn't working.
https://stackblitz.com/edit/angular-jmitne?file=app/expansion-overview-example.html.
This gives the error _co.accordion.openAll is not a function

ouq77:
@pankajsri03 - you are using cdk and material v5.2.1 in your example. This change will only be available in v6. (It is currently in beta)

zijianhuang:
@pankajsri03  and @ouq77 
I am using NG6.0.2 and NGMD6.0.2, and see the same "openAll is not a function".

`  "dependencies": {
    "@angular/animations": "^6.0.2",
    "@angular/cdk": "^6.0.2",
    "@angular/common": "^6.0.2",
    "@angular/compiler": "^6.0.2",
    "@angular/compiler-cli": "^6.0.2",
    "@angular/core": "^6.0.2",
    "@angular/flex-layout": "^6.0.0-beta.15",
    "@angular/forms": "^6.0.2",
    "@angular/material": "^6.0.2",
    "@angular/material-moment-adapter": "^6.0.2",
    "@angular/platform-browser": "^6.0.2",
    "@angular/platform-browser-dynamic": "^6.0.2",
    "@angular/platform-server": "^6.0.2",
    "@angular/router": "^6.0.2",
`

zijianhuang:
As of NGMD 6.2.1, the problem persist: "openAll is not a function". Do I miss something? or someone in the NGMD team had forgotten to merge?

ouq77:
@zijianhuang I assume your `consult-item` component contains the `mat-expansion-panel`? I may be wrong, but for any of the expansion panel logic to work as expected, they need to be hosted directly inside the `mat-accordion`.

zijianhuang:
@ouq77, yes, the root element of consult-item is 
`<mat-expansion-panel (opened)="handleOpened()" (closed)="handleClosed()">`

So, after NG rendering, mat-expansion-panel is hosted directly inside the mat-accordion.

PR #7462, opened 2017-10-02
feat(chips): Use all available space for the input

ppham27:
Currently, an input inside a chips list is limited to 150px, so it looks like this.

Cutoff Name.

Adding flex lets the input use all available space and still wrap correctly.

Using all avaible space

Wrapping


HideDev:
+1 )

PR #7482, opened 2017-10-02
feat(stepper): allow for header icons to be customized

### LINKED ISSUE #7384: stepper dependency on material icons

bug

Texts are not printed on the screen instead of the icons if the icon set is not present.

Brownie points:
I can set up another icons from another icon sets (e.g. FontAwesome)

stepper component has the dependency on material icons. If material icons are not used, it's rendered as text on the screen.

image

Index.html line 24 adds the necessary icons. If that line is commented out, icons will faill and text is printed on screen instead of icon.
Plunker template: https://plnkr.co/edit/n4ry3CkDsvV22aX4h5lo?p=preview

I want to use FontAwesome icons (and I don't want to bundle the Material icons just for this one component)

Material 2.0.0-beta.11
Angular 4.4.3

Thanks for the Stepper component!

crisbeto:
Currently users are locked into using the Material `create` and `done` icon for the step headers. These changes add the ability to customize the icons by providing an `ng-template` with an override.

Fixes #7384.

**Note:** I'm not too sure about the naming (`matStepperIcon` vs `matStepIcon`). I went with `matStepperIcon`, because it applies to the entire stepper, rather than an individual step.

### DIFF
--- a/src/lib/stepper/step-header.ts
+++ b/src/lib/stepper/step-header.ts
@@ -16,6 +16,7 @@ import {
   Input,
   OnDestroy,
   ViewEncapsulation,
+  TemplateRef,
 } from '@angular/core';
 import {Subscription} from 'rxjs/Subscription';
 import {MatStepLabel} from './step-label';
@@ -38,11 +39,14 @@ import {MatStepperIntl} from './stepper-intl';
 export class MatStepHeader implements OnDestroy {
   private _intlSubscription: Subscription;
 
-  /** Icon for the given step. */
-  @Input() icon: string;
+  /** State of the given step. */
+  @Input() state: string;
 
   /** Label of the given step. */
   @Input() label: MatStepLabel | string;
+
+  /** Overrides for the header icons, passed in via the stepper. */
+  @Input() iconOverrides: {[key: string]: TemplateRef<any>};
 
   /** Index of the given step. */
   @Input()
--- a/src/lib/stepper/stepper-icon.ts
+++ b/src/lib/stepper/stepper-icon.ts
@@ -1,0 +1,22 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive, Input, TemplateRef} from '@angular/core';
+
+/**
+ * Template to be used to override the icons inside the step header.
+ */
+@Directive({
+  selector: 'ng-template[matStepperIcon]',
+})
+export class MatStepperIcon {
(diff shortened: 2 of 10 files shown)


src/lib/stepper/step-header.ts
```diff
@@ -25,12 +33,15 @@ import {MatStepLabel} from './step-label';
   preserveWhitespaces: false,
 })
 export class MatStepHeader implements OnDestroy {
-  /** Icon for the given step. */
-  @Input() icon: string;
+  /** State of the given step. */
+  @Input() state: string;
```
jelbourn: From the name I'm not sure that `state` is meant to be

src/lib/stepper/step-header.ts
```diff
@@ -25,12 +33,15 @@ import {MatStepLabel} from './step-label';
...
 
   /** Label of the given step. */
   @Input() label: MatStepLabel | string;
 
+  /** Overrides for the header icons, passed in via the stepper. */
+  @Input() iconOverrides: {[key: string]: TemplateRef<any>};
```
jelbourn: Why not a map?

src/lib/stepper/stepper-icon.ts
```diff
@@ -0,0 +1,22 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive, Input, TemplateRef} from '@angular/core';
+
+/**
+ * Template to be used to override the icons inside the step header.
+ */
+@Directive({
+  selector: 'ng-template[matStepperIcon]',
```
jelbourn: Does this work with 
```html
<mat-icon *matStepperIcon>
```
?

src/lib/stepper/step-header.ts
crisbeto: It doesn't need to be. It's just a `string->TemplateRef` map.

src/lib/stepper/step-header.ts
crisbeto: This component is internal anyway, but it's supposed to mean the state that the header is in (e.g. done, current).

src/lib/stepper/stepper-icon.ts
crisbeto: Hmm, I haven't tried it but I suppose.

jelbourn:
Ah, I didn't know it was an internal component. I'll leave it to @mmalerba to review, then

AsafAgranat:
Additionally, currently the steppers icons are rendered using the web-font variant. **SVG icons** (using the ` svgIcon` attribute) should be supported too.

aamir1995:
Any update on this PR?

ganqqwerty:
so, guys, what are you waiting for? Merge that thing!

PR #7483, opened 2017-10-02
feat(badge): add badge component

amcdnl:
Implements a badge component.

## Design Document 
https://docs.google.com/document/d/12sksu2GHcITFBidL8nyRzIt8lnVUoGnzwQx4tXWesEw/edit?usp=sharing


ocarreterom:
Nice job, but I think the button implementation breaks the flexibility of the badge component.
I think it's more flexible something like this:
```html
<button mat-button>
  Notifications
  <mat-badge>10</mat-badge>
</button>

-- OR --

<button mat-button>
  Notifications
  <mat-badge content="10"></mat-badge>
</button>

-- OR --

<button mat-button>
  <mat-badge content="10">Notifications</mat-badge>
</button>

```
Moreover, in the future could be implemented in `mat-list`, `mat-tab`, etc.
```html
<mat-tab label="Notifications">
  <mat-badge>10</mat-badge>
</mat-tab>

<mat-list-item>
  Notifications
  <mat-badge>10</mat-badge>
</mat-list-item>
```
What do you think?

amcdnl:
@jelbourn - Do you think there is a case for just having a bubble with no text in it?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  /** Content of the badget */
+  @Input() content: string = '';
+
+  /** Direction the badge should reside; 'before|after above|below' */
+  @Input() direction: string = 'above after';
+
+  get _isAbove(): boolean { return this.direction.indexOf('above') > -1; }
+
+  get _isBefore(): boolean { return this.direction.indexOf('before') > -1; }
```
willshowell: These getters will run every cd cycle, right? What about using a setter in `direction` instead. Also OnPush for this component?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,61 @@
...
+  get overlap(): boolean {
+    return this._overlap;
+  }
+  private _overlap: boolean = true;
+
+  /** Content of the badget */
+  @Input() content: string = '';
+
+  /** Direction the badge should reside; 'before|after above|below' */
+  @Input() direction: string = 'above after';
```
willshowell: The default value here is inconsistent with how the `_isBefore` getter works. If I don't pass a direction, it will end up as `above after`. If I use `"banana banana"`, it will be `below after`.

As a nit, the overview guide implies there is an order to which the directions should be in, but that order is reversed in the jsdoc comment.

src/lib/badge/badge.scss
```diff
@@ -0,0 +1,66 @@
...
+  .mat-badge-content {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+    justify-content: center;
+    align-content: center;
+    align-items: center;
+    position: absolute;
+    font-weight: 600;
+    font-size: 12px;
```
willshowell: Should this be set via a typography mixin?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,61 @@
...
+  @Input()
+  set overlap(val: boolean) {
+    this._overlap = coerceBooleanProperty(val);
+  }
+  get overlap(): boolean {
+    return this._overlap;
+  }
+  private _overlap: boolean = true;
+
+  /** Content of the badget */
```
willshowell: Typo

src/lib/badge/badge.ts
amcdnl: I made the getters more specific to prevent the monkeys ;P

Thanks for the heads up. Fixed that too.

src/lib/badge/badge.ts
amcdnl: Good idea :)

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+  get direction(): string {
+    return this._direction;
+  }
+  private _direction: string = 'above after';
+
+  _isAbove: boolean = true;
+  _isBelow: boolean = false;
+  _isBefore: boolean = false;
+  _isAfter: boolean = true;
```
willshowell: I don't know if all of these are necessary. In my previous comment, I was thinking more like this

```ts
@Input()
set direction(val: string) {
  this._direction = val;
  this._isAbove = val.indexOf('below') === -1;
  this._isAfter = val.indexOf('before') === -1;
}
get direction(): string {
  return this._direction;
}
private _direction: string = 'above after';

_isAbove: boolean = true;
_isAfter: boolean = true;
```

That way it always defaults to `above` unless `below` is explicitly stated. Likewise for `after`.

src/lib/badge/badge.html
```diff
@@ -0,0 +1,5 @@
+<ng-content></ng-content>
+<span 
+  class="mat-badge-content" 
+  [innerHTML]="content">
```
crisbeto: This looks dangerous. Is there any reason people would want to put anything besides plain text in a badge?

src/lib/badge/badge.md
```diff
@@ -0,0 +1,35 @@
...
+background color to `primary`, `accent`, or `warn`.
+
+```html
+<mat-badge color="accent" content="22">
+  <mat-icon>home</mat-icon>
+</mat-badge>
+```
+
+### Accessibility
+Badges should be given a meaningful label via `aria-label` or `aria-labelledby` attributes.
```
crisbeto: I'm not sure whether this is correct. Shouldn't the badge target have an `aria-describedby` instead @jelbourn?

src/lib/badge/badge.scss
```diff
@@ -0,0 +1,64 @@
+$mat-badge-size: 22px !default;
+
+.mat-badge {
+  position: relative;
+  display: inline-block;
+
+  &.mat-badge-above {
```
crisbeto: This nesting seems like it's bumping the specificity unnecessarily. Consider moving it out instead?

src/lib/badge/badge.scss
```diff
@@ -0,0 +1,64 @@
...
+        right: -$mat-badge-size / 2;
+      }
+    }
+  }
+
+  .mat-badge-content {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+    justify-content: center;
```
crisbeto: Does this really need flexbox? It can be done simpler with `text-align` and `line-height`.

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,82 @@
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
+import {Component, DebugElement} from '@angular/core';
+import {By} from '@angular/platform-browser';
+import {MatBadge,  MatBadgeModule} from './index';
+
+
+fdescribe('MatBadge', () => {
```
crisbeto: Remove the `fdescribe`.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,74 @@
...
+  templateUrl: './badge.html',
+  styleUrls: ['./badge.css'],
+  inputs: ['color'],
+  host: {
+    'class': 'mat-badge',
+    '[class.mat-badge-overlap]': 'overlap',
+    '[class.mat-badge-above]': '_isAbove',
+    '[class.mat-badge-below]': '!_isAbove',
+    '[class.mat-badge-before]': '!_isAfter',
+    '[class.mat-badge-after]': '_isAfter',
```
crisbeto: I think that you can cut down on the amount of classes by making it appear `above before` by default. The other classes can override it afterwards.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,74 @@
...
+ import {CanColor, mixinColor} from '@angular/material/core';
+
+ /** @docs-private */
+export class MatBadgeBase {
+  constructor(public _renderer: Renderer2, public _elementRef: ElementRef) {}
+}
+export const _MatBadgeMixinBase = mixinColor(MatBadgeBase);
+
+@Component({
+  selector: 'mat-badge',
```
crisbeto: This needs a `moduleId: module.id`.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,74 @@
...
+  set overlap(val: boolean) {
+    this._overlap = coerceBooleanProperty(val);
+  }
+  get overlap(): boolean {
+    return this._overlap;
+  }
+  private _overlap: boolean = true;
+
+  /** Content of the badge */
+  @Input() content: string = '';
```
crisbeto: Rather than using an input to add the content, why not use content projection? It should allow for move flexibility.

src/lib/badge/badge.html
amcdnl: Removed in latest.

src/lib/badge/badge.ts
amcdnl: Removed in latest.

src/lib/badge/badge.ts
amcdnl: I started with that but it got odd setting everything backwards. For instance `before` would have to set `right: auto` and `left: 22px`. It felt odd to add that. Happy to change if you guys prefer.

src/lib/badge/badge.ts
amcdnl: Its a directive now so doesn't apply.

src/lib/badge/badge.scss
amcdnl: It now supports font icons and svgs, so its needed now.

src/lib/badge/badge.scss
amcdnl: I changed this quite a bit, you might want to re-review.

amcdnl:
Updated w/ new API. Can you review again @crisbeto @willshowell

src/lib/badge/badge.md
amcdnl: We need to reconsider how we want to do this with latest api changes.

src/lib/badge/badge.md
```diff
@@ -0,0 +1,54 @@
+Badges are small status descriptors for UI elements. A badge consists of a small circle, 
+typically containing a number or other short set of characters, that appears in proximity to another object.
+
+### Icons
+Badges can contain text or font icons and SVGs registered with the `MdIconRegistry`
+
+#### Font Icons
+Badges using font icons are delcared with the `matIconBadge` tag and its content being
```
willshowell: `delcared` -> `declared`

src/lib/badge/public_api.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+export * from './badge-module';
+export * from './badge';
```
willshowell: Rename this file to `public-api.ts` per #7403

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,223 @@
...
+import {ThemePalette} from '@angular/material/core';
+
+let nextId = 0;
+
+ /** @docs-private */
+export class MatBadgeBase {
+
+  /** Theme for the badge */
+  get matBadgeColor(): ThemePalette { return this._color; }
+  set matBadgeColor(value: ThemePalette) {
```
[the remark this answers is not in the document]
amcdnl: Hmm, let me know which way you wanna go.

src/demo-app/badge/badge-demo.html
```diff
@@ -0,0 +1,84 @@
...
+
+  <button mat-raised-button matBadge="22" matBadgePosition="below before">
+    <mat-icon color="primary">home</mat-icon>
+  </button>
+
+  <button mat-raised-button>
+    <mat-icon color="primary" matBadge="22" color="accent">home</mat-icon>
+  </button>
+
+  <br />
```
crisbeto: Use a div with a margin instead of `br`-s?

src/demo-app/badge/badge-demo.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+export class BadgeDemo {
+
+  badgeContent = '1';
+  svgIcon = 'cat';
+
+  constructor(_iconRegistry: MatIconRegistry, sanitizer: DomSanitizer) {
+    _iconRegistry.addSvgIcon('home',
+      sanitizer.bypassSecurityTrustResourceUrl(
+        'https://upload.wikimedia.org/wikipedia/commons/3/34/Home-icon.svg'));
```
crisbeto: You can use the home icon from the icon demo instead of fetching one remotely.

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,177 @@
...
+
+$mat-badge-font-size: 12px;
+$mat-badge-font-weight: 600;
+$mat-badge-size: 22px !default;
+
+.mat-badge {
+  position: relative;
+}
+
+.mat-badge:after,
```
crisbeto: These `:after` instances should be `::after`. I think we had a linting rule for it at some point.

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,177 @@
...
+.mat-badge-icon,
+.mat-badge-text,
+.mat-badge-svg-icon {
+  position: absolute;
```
crisbeto: All of these positioning and alignment styles should be in the `badge.scss`, not in the theme. The theme is only for color-related properties.

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,177 @@
...
+  height: $mat-badge-size;
+  border-radius: 50%;
+  content: '';
+  z-index: 1;
+}
+
+.mat-badge-text,
+.mat-badge .mat-badge-icon {
+  z-index: 2;
+  font-size: 16px;
```
crisbeto: This should be in the typography mixin and the value, ideally, should come from the typography breakpoints.

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,177 @@
...
+  width: $mat-badge-size;
+  height: $mat-badge-size;
+}
+
+.mat-badge-svg-icon {
+  z-index: 2;
+  width: 12px;
+  height: 12px;
```
crisbeto: Move this into a variable?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  _id: number = nextId++;
+  _isAbove: boolean = true;
+  _isAfter: boolean = true;
+
+  constructor(public _renderer: Renderer2, public _elementRef: ElementRef) {}
+
+  /** Clears the previous badge content */
+  _clearContents(): void {
+    const contents = document.getElementById(`badge-content-${this._id}`);
```
crisbeto: Couldn't you take a hold of the badge content using a `ViewChild` or `ContentChild`?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,221 @@
...
+  get content(): string { return this._content; }
+  private _content: string;
+
+  constructor(_renderer: Renderer2, _elementRef: ElementRef) {
+    super(_renderer, _elementRef);
+  }
+
+  /** Injects a span element into the DOM with the content. */
+  private _setContent(): void {
+    const content = document.createElement('span');
```
crisbeto: I don't see why this has to create and insert elements manually. Could you wrap the element via content projection?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,221 @@
...
+      private _iconRegistry: MatIconRegistry,
+      _elementRef: ElementRef,
+      _renderer: Renderer2) {
+    super(_renderer, _elementRef);
+  }
+
+  /**  Gets the font icon set to use. */
+  private _getFontSet(): string {
+    return this.matBadgeFontSet ?
+        this._iconRegistry.classNameForFontAlias(this.matBadgeFontSet) :
```
crisbeto: This assumes that the consumer will use Material icons, which may not be the case. Why not let them provide the icon element themselves?

src/demo-app/badge/badge-demo.ts
amcdnl: I tried that, however, it was not working. That demo page is broken too if you open it up.

src/lib/badge/_badge-theme.scss
amcdnl: I agree, however, since this is a directive there is no way to include styles w/ the component. Do you have a good solution for that?

src/lib/badge/_badge-theme.scss
amcdnl: Sure, I've moved it inside that mixin.

This size is independent from the font-size of the page since its specifically designed for this fixed width/height of the badge.

src/lib/badge/badge.ts
amcdnl: I thought that only worked for angular inserted dom. This is manually inserted.

src/lib/badge/badge.ts
amcdnl: This was built to implement Material Icons. I think custom icons might be out of scope for this. @jelbourn - thoughts?

Although I'm curious how you would implement providing the icon element themselves, could you elaborate?

src/lib/badge/badge.ts
amcdnl: As far as I know, directives don't have content projection. You can't have multiple components on the same element either so that would't work. The SVG icons do a similar insertion and removal so this isn't the first place we are doing this type of thing.

I had this as a component at first but that API wasn't what @jelbourn had designed.

src/lib/badge/_badge-theme.scss
crisbeto: You can make it a component that has a template of `<ng-content>`. Having these styles in the theme will cause them to be loaded multiple times for pages with multiple themes.

src/lib/badge/badge.ts
crisbeto: This plays into my other comment (https://github.com/angular/material2/pull/7483#discussion_r143226186) that it probably shouldn't be constructing DOM in the first place. IMO it should be fine to have it as a component instead of a directive.

src/lib/badge/badge.ts
crisbeto: If you convert it into a component and allow content projection, users will be able to override it as they want. Another approach can be to provide an `ng-template`, but that might be overkill for this case (see https://github.com/angular/material2/pull/7482).

src/lib/badge/badge-module.ts
```diff
@@ -0,0 +1,28 @@
...
+import {MatBadge, MatSvgIconBadge, MatIconBadge} from './badge';
+
+
+@NgModule({
+  imports: [MatCommonModule],
+  exports: [
+    MatBadge,
+    MatIconBadge,
+    MatSvgIconBadge,
+    MatCommonModule,
```
rafaelss95: Is there any reason to export `MatCommonModule` here?

src/lib/badge/_badge-theme.scss
amcdnl: Ya, thats what I had. You can't do multiple components on the same element though, example: icon w/ a badge.

src/lib/badge/badge.ts
amcdnl: Ya, thats what I had but @jelbourn wanted this to be a directive that can decorate components like icons. I totally agree w/ it not constructing dom though, open to suggestions.

src/lib/badge/badge-module.ts
amcdnl: Good catch. thanks

src/lib/badge/_badge-theme.scss
crisbeto: You can't, but you can have an icon inside a badge.

src/lib/badge/badge.ts
CaerusKaru: You could create a separate component and dynamically compile and attach it (for `span` or `svg`, then store it in an array for later removal.

src/lib/badge/badge-module.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
CaerusKaru: This header needs to be updated to reflect the change from `Google Inc.` to `Google LLC`. Also applies to `badge.ts`, `index.ts`, and `public-api.ts`

amcdnl:
Updates:
- Remove icon badges
- Add aria description label
- Add sizes
- Add animation

@jelbourn - ready for further review.

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    beforeEach(() => {
+      badgeDebugElement = fixture.debugElement.query(By.directive(MatBadge));
+      fixture.detectChanges();
+    });
+
+    it('should update the badge based on attribute', () => {
+      let badgeContentDebugElement =
+          badgeDebugElement.nativeElement.querySelector('.mat-badge-content');
+
+      expect(badgeContentDebugElement.innerHTML).toContain('1');
```
jelbourn: Prefer `textContent` to `innerHTML` for things like this

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  selector: 'test-app',
+  template: `
+    <span [matBadge]="badgeContent"
+          [matBadgeColor]="badgeColor"
+          [matBadgePosition]="badgeDirection">
+      home
+    </span>
+  `
+})
+class TestApp {
```
jelbourn: `BadgeWithTextContent`

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,143 @@
...
+    font-size: $mat-badge-font-size * 2;
+  }
+}
+
+.mat-badge {
+  position: relative;
+
+  &.mat-badge-hidden {
+    .mat-badge-content {
+      visibility: hidden;
```
jelbourn: Why `visibility: hidden` instead of `display: none`?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (colorPalette !== this._color) {
+      if (this._color) {
+        this._renderer.removeClass(this._elementRef.nativeElement, `mat-badge-${this._color}`);
+      }
+      if (colorPalette) {
+        this._renderer.addClass(this._elementRef.nativeElement, `mat-badge-${colorPalette}`);
+      }
+
+      this._color = colorPalette;
+    }
```
jelbourn: Can you break this logic out of the setter and into its own function?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+  private _color: ThemePalette = 'primary';
+
+  /** Whether the badge should overlap its contents or not */
+  @Input()
+  set matBadgeOverlap(val: boolean) {
```
jelbourn: The property names should be just `overlay`, `position`, etc. Only the input name needs to be fully qualified, e.g.
```ts
@Input('matBadgePosition')
get position() {
  // ...
}
```

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+        this._renderer.addClass(this._elementRef.nativeElement, `mat-badge-${colorPalette}`);
+      }
+
+      this._color = colorPalette;
+    }
+  }
+  private _color: ThemePalette = 'primary';
+
+  /** Whether the badge should overlap its contents or not */
+  @Input()
```
jelbourn: Decorators should go on the getter rather than the setter, prefer putting the getter first

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+    this._hidden = coerceBooleanProperty(val);
+  }
+  get matBadgeHidden(): boolean {
+    return this._hidden;
+  }
+  private _hidden: boolean;
+
+  _id: number = nextId++;
+  _isAbove: boolean = true;
+  _isAfter: boolean = true;
```
jelbourn: Add JsDoc descriptions

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+    '[class.mat-badge-after]': '_isAfter',
+    '[class.mat-badge-small]': 'matBadgeSize === "small"',
+    '[class.mat-badge-medium]': 'matBadgeSize === "medium"',
+    '[class.mat-badge-large]': 'matBadgeSize === "large"',
+    '[class.mat-badge-hidden]': 'matBadgeHidden',
+  },
+})
+export class MatBadge {
+
+  /** Theme for the badge */
```
jelbourn: Use the same description that other components use for the `color` property

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+  @Input()
+  set matBadgeOverlap(val: boolean) {
+    this._overlap = coerceBooleanProperty(val);
+  }
+  get matBadgeOverlap(): boolean {
+    return this._overlap;
+  }
+  private _overlap: boolean = true;
+
+  /** Position the badge should reside; 'above|below before|after' */
```
jelbourn: The comment should be something like
"Accepts any combination of 'above'|'below' and 'before'|'after'"

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  @Input('matBadge')
+  set content(val: string) {
+    this._content = val;
+    this._setContent();
+  }
+  get content(): string { return this._content; }
+  private _content: string;
+
+  /** Aria description */
```
jelbourn: "Message used to describe the decorated element via `aria-describedby`"

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+  @Input()
+  set matBadgeDescription(val: string) {
+    this._description = val;
+    this._setLabel();
+  }
+  get matBadgeDescription(): string { return this._description; }
+  private _description: string;
+
+  /** Size of the badge */
+  @Input() matBadgeSize: string = 'medium';
```
jelbourn: The type here should be `'small' | 'medium' | 'large'`

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+
+import {Directive, Input, Renderer2, ElementRef} from '@angular/core';
+import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {ThemePalette} from '@angular/material/core';
+
+let nextId = 0;
+
+/** Directive to display a text badge. */
+@Directive({
```
jelbourn: Description should talk a bit about how this is typically used.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    return content;
+  }
+
+  /** Sets the aria-label property on the element */
+  private _setLabel(): void {
+    // ensure content available before setting label
+    const content = this._setContent();
+    content.setAttribute('aria-label', this.matBadgeDescription);
```
jelbourn: This will clobber any existing `aria-label`, which we don't want. Instead, we want to augment the element with an _additional_ description. You should be able to use `AriaDescriber` for this.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+  private _hidden: boolean;
+
+  _id: number = nextId++;
+  _isAbove: boolean = true;
+  _isAfter: boolean = true;
+
+  constructor(private _renderer: Renderer2, private _elementRef: ElementRef) {}
+
+  /** Injects a span element into the DOM with the content. */
+  private _setContent(): HTMLSpanElement {
```
jelbourn: `_updateTextContent`?

It would be better if the message was passed in, since it makes this function independent of the class state.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+    let content = document.getElementById(`badge-content-${this._id}`);
+
+    if (!content) {
+      content = document.createElement('span');
+      content.setAttribute('id', `badge-content-${this._id}`);
```
jelbourn: Prefix with `mat`

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  _id: number = nextId++;
+  _isAbove: boolean = true;
+  _isAfter: boolean = true;
+
+  constructor(private _renderer: Renderer2, private _elementRef: ElementRef) {}
+
+  /** Injects a span element into the DOM with the content. */
+  private _setContent(): HTMLSpanElement {
+    let content = document.getElementById(`badge-content-${this._id}`);
```
jelbourn: If you want this to work on `@angular/platform-server` today, it needs to use the renderer API

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+      content.textContent = this.content;
+
+      if (this.matBadgeDescription) {
+        content.setAttribute('aria-label', this.matBadgeDescription);
+      }
+
+      this._elementRef.nativeElement.appendChild(content);
+
+      // animate in after insertion
+      setTimeout(() => {
```
jelbourn: This timeout strikes me as odd; what's the background here?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,146 @@
...
+        // ensure content available
+        if (content !== null) {
+          content.classList.add('mat-badge-active');
+        }
+      }, 100);
+    } else if (content.textContent !== this.content) {
+      content.textContent = this.content;
+    }
```
jelbourn: For no-op cases, I generally prefer to return from the function quickly at the top rather than doing everything in a conditional

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,143 @@
+@import '../core/theming/palette';
+@import '../core/theming/theming';
+@import '../core/typography/typography-utils';
+
+$mat-badge-font-size: 12px;
```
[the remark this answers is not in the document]
crisbeto: Yeah that would be ideal, but I don't see any breakpoint that maps nicely. @amcdnl is this font size somewhere in the spec?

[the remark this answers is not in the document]
amcdnl: @crisbeto - There is no spec for badges.

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,143 @@
...
+        background: mat-color($warn);
+      }
+    }
+  }
+}
+
+@mixin mat-badge-typography($config) {
+  .mat-badge-content {
+    font-weight: $mat-badge-font-weight;
+    font-size: $mat-badge-font-size;
```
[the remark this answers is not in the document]
amcdnl: This sizes different based on the small|med|large so needed to make them independent.

src/lib/badge/badge.ts
amcdnl: None of the other components have this, can you provide an example of what'd you like to see?

src/lib/badge/badge.ts
amcdnl: The other usages use a mixin ( which I cannot use here due to different name ) and don't have comments.

src/lib/badge/badge.ts
amcdnl: It relies on the id too. Would you like me to pass that in too?

src/lib/badge/badge.ts
amcdnl: Want to wait a short duration before the animation is activated. Do you think it should be removed and be instant?

src/lib/badge/badge.ts
amcdnl: Both conditions exit the same way, rather than have this in multiple places I figured just do it once. Would you like me to change given that still?

amcdnl:
@jelbourn - done w/ comments above.

src/lib/badge/badge.ts
jelbourn:
```ts
```

jelbourn:
FYI pausing on this until after RC

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+}
+
+@mixin mat-badge-theme($theme) {
+  $accent: map-get($theme, accent);
+  $warn: map-get($theme, warn);
+  $primary: map-get($theme, primary);
+  
+    .mat-badge-content {
```
ocarreterom: Wrong indentation?

amcdnl:
Tag @jelbourn

shlomiassaf:
@amcdnl 
See this bug: https://github.com/amcdnl/material2/blob/badge/src/lib/badge/badge.ts#L74-L80

The name in the input is not the right name

src/lib/badge/badge.md
```diff
@@ -0,0 +1,33 @@
...
+background color to `primary`, `accent`, or `warn`.
+
+```html
+<mat-icon matBadge="22" matBadgeColor="accent">
+  home
+</mat-icon>
+```
+
+### Accessibility
+Badges should be given a meaningful label via `aria-label` or `aria-labelledby` attributes.
```
jelbourn: I don't think this is right. This section should talk about how the badges use `aria-describedby` to augment the host element.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,162 @@
...
+  private _updateTextContent(): HTMLSpanElement {
+    let content = document.getElementById(`mat-badge-content-${this._id}`);
+
+    if (!content) {
+      content = this._renderer.createElement('span');
+
+      content = document.createElement('span');
+      content.setAttribute('id', `mat-badge-content-${this._id}`);
+      content.classList.add('mat-badge-content');
+      content.textContent = this.content;
```
jelbourn: Thought about this a bit more: I think we should create something like a `MatBadgeContent` component and then a `DomPortalHost` to put it in place. This would let us associate the non-color/typography styles with that component instead of having to put it all in the theme (this is what we do with tooltip except it wouldn't be overlay-based here).

src/demo-app/badge/badge-demo.html
```diff
@@ -0,0 +1,90 @@
...
+  </div>
+
+  <div class="badge-examples">
+    <h3>Buttons</h3>
+    <button mat-raised-button [matBadge]="badgeContent">
+      <mat-icon color="primary">home</mat-icon>
+    </button>
+
+    <button mat-raised-button matBadge="22" matBadgePosition="below after" color="primary" matBadgeColor="accent">
+      <mat-icon color="primary" color="accent">home</mat-icon>
```
rafaelss95: `color` is duplicated.

src/demo-app/badge/badge-demo.html
```diff
@@ -0,0 +1,90 @@
...
+
+    <span matBadge="22" matBadgeOverlap="false" matBadgePosition="above before" matBadgeColor="warn">
+      Hello
+    </span>
+
+    <span matBadge="⚡️" matBadgeOverlap="false" matBadgePosition="below before">
+      Hello
+    </span>
+
+    <span [matBadge]="badgeContent" matBadgeDescription="I've got {{badgeContent}} problems">
```
rafaelss95: I think it’s preferable to use property binding `[matBadgeDescription]="..."`

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  template: `
+    <span [matBadge]="badgeContent"
+          [matBadgeColor]="badgeColor"
+          [matBadgePosition]="badgeDirection">
+      home
+    </span>
+  `
+})
+class BadgeWithTextContent {
+  badgeColor;
```
rafaelss95: `badgeColor: ThemePalette`?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,162 @@
...
+   * Accepts any combination of 'above'|'below' and 'before'|'after'
+   */
+  @Input('matBadgePosition')
+  get position(): string { return this._position; }
+  set position(val: string) {
+    this._position = val;
+    this._isAbove = val.indexOf('below') === -1;
+    this._isAfter = val.indexOf('before') === -1;
+  }
+  private _position: string = 'above after';
```
rafaelss95: Instead of declare it as string, it would be great if we can define a `type` for this.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,162 @@
...
+  @Input('matBadgePosition')
+  get description(): string { return this._description; }
+  set description(val: string) {
+    this._setLabel(val, this._description);
+    this._description = val;
+  }
+  private _description: string;
+
+  /** Size of the badge. 'small' | 'medium' | 'large' */
+  @Input() matBadgeSize: string = 'medium';
```
rafaelss95: Use a more specific type... something like `MatBadgeSize`?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,162 @@
...
+  _id: number = nextId++;
+
+  /** Whether the badge is above the host or not */
+  _isAbove: boolean = true;
+
+  /** Whether the badge is after the host or not */
+  _isAfter: boolean = true;
+
+  constructor(
+      private _renderer: Renderer2,
```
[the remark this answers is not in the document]
rafaelss95: @jelbourn out of curiosity, is this practice being adopted by core team also?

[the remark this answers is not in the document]
CaerusKaru: @jelbourn then this needs the DOCUMENT token to work properly with SSR, and all current document refs need to be switched over.

kamok:
Hey guys, do you have a tentative date when the Badge will be ready? Really looking forward to it.

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,166 @@
...
+      color: mat-color($primary, default-contrast);
+      background: mat-color($primary);
+    }
+
+    .mat-badge {
+      &.mat-badge-accent {
+      .mat-badge-content {
```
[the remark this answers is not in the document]
amcdnl: > Would it work to just use the font-size from caption?
The size needs to be absolute for the sizing of the bubble.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,162 @@
...
+  private _hidden: boolean;
+
+  /** Unique id for the badge */
+  _id: number = nextId++;
+
+  /** Whether the badge is above the host or not */
+  _isAbove: boolean = true;
+
+  /** Whether the badge is after the host or not */
+  _isAfter: boolean = true;
```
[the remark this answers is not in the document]
amcdnl: I did this for performance reasons. Would you still like me to change it?

src/lib/badge/badge.ts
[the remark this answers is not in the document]
amcdnl: Updated and used document token.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  constructor(
+      private _renderer: Renderer2,
+      private _elementRef: ElementRef,
+      private _ariaDescriber: AriaDescriber) {}
+
+  /** Injects a span element into the DOM with the content. */
+  private _updateTextContent(): HTMLSpanElement {
```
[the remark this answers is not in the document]
amcdnl: See:

```
  /** Sets the aria-label property on the element */
  private _setLabel(val: string, prevVal: string): void {
    // ensure content available before setting label
    const content = this._updateTextContent();
```

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,162 @@
...
+  /** Sets the aria-label property on the element */
+  private _setLabel(val: string, prevVal: string): void {
+    // ensure content available before setting label
+    const content = this._updateTextContent();
+    this._ariaDescriber.removeDescription(content, prevVal);
+    this._ariaDescriber.describe(content, val);
+  }
+
+  /** Adds css theme class given the color to the component host */
+  private _setColor(value: ThemePalette) {
```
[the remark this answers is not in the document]
amcdnl: I started with that but if the host element describes color, you might want a different color for your badge. Example: button case.

src/demo-app/badge/badge-demo.html
amcdnl: @jelbourn ^^?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+  /** Message used to describe the decorated element via aria-describedby */
+  @Input('matBadgeDescription')
+  get description(): string { return this._description; }
+  set description(val: string) {
+    this._updateHostAriaDescription(val, this._description);
+    this._description = val;
+  }
+  private _description: string;
+
+  /** Size of the badge. Can be 'small' , 'medium', or 'large'. */
```
rafaelss95: There's an extra space after **'small' ,**

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+  @Input() matBadgeSize: BadgeSize = 'medium';
+
+  /** Whether the badge is hidden. */
+  @Input()
+  set matBadgeHidden(val: boolean) {
```
rafaelss95: `getters` should come before than `setters`. See #9471.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+  @Input('matBadgeDescription')
+  get description(): string { return this._description; }
+  set description(val: string) {
+    this._updateHostAriaDescription(val, this._description);
+    this._description = val;
+  }
+  private _description: string;
+
+  /** Size of the badge. Can be 'small' , 'medium', or 'large'. */
+  @Input() matBadgeSize: BadgeSize = 'medium';
```
rafaelss95: `@Input('matBadgeSize') size: BadgeSize = 'medium';`?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+
+import {Directive, Input, ElementRef, Inject, Optional, NgZone} from '@angular/core';
+import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {ThemePalette} from '@angular/material/core';
+import {AriaDescriber} from '@angular/cdk/a11y';
+import {DOCUMENT} from '@angular/common';
+
+let nextId = 0;
+
+export type PositionTypes =
```
rafaelss95: Not really sure about this, but I guess that all (*public*) interfaces/types must be prefixed with `Mat#{ComponentName/DirectiveName}`. The same applies for `BadgeSize`.

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      fixture.detectChanges();
+
+      expect(badgeNativeElement.classList.contains('mat-badge-below')).toBe(true);
+      expect(badgeNativeElement.classList.contains('mat-badge-before')).toBe(true);
+    });
+  });
+
+});
+
+/** Test component that contains an MatBadge. */
```
rafaelss95: an MatBadge -> *a MatBadge*.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+
+export type BadgeSize =
+  'small' |
+  'medium' |
+  'large';
+
+/** Directive to display a text badge. */
+@Directive({
+  selector: '[matBadge]',
+  inputs: ['matBadgeColor', 'matBadgeOverlap', 'matBadgePosition'],
```
rafaelss95: Do you still need to define these `inputs` since you've already defined them using `@Input`?

src/lib/badge/badge.ts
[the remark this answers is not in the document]
jelbourn: The difference in performance for property access vs. function call here should be negligible

src/demo-app/badge/badge-demo.html
jelbourn: Using interpolation is fine when it's mixed with static text

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,165 @@
...
+    }
+  
+    &.mat-badge-warn {
+      .mat-badge-content {
+        color: mat-color($warn, default-contrast);
+        background: mat-color($warn);
+      }
+    }
+  }
+}
```
jelbourn: There's an extra closing `}` here (masked by the extra indentation in the mixin block)

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    this._elementRef.nativeElement.appendChild(badgeElement);
+
+    // animate in after insertion
+    this._ngZone.runOutsideAngular(() => requestAnimationFrame(() => {
+      // ensure content available
+      if (badgeElement) {
+        badgeElement.classList.add('mat-badge-active');
+      }
+    });
```
jelbourn: Compilation error due to missing `)` here (for the `runOutsideAngular`)

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      testComponent.badgeColor = 'accent';
+      fixture.detectChanges();
+      expect(badgeNativeElement.classList.contains('mat-badge-accent')).toBe(true);
+
+      testComponent.badgeColor = 'warn';
+      fixture.detectChanges();
+      expect(badgeNativeElement.classList.contains('mat-badge-warn')).toBe(true);
+
+      testComponent.badgeColor = null;
```
jelbourn: Compilation error: `null` is not assignable to `ThemePalette`

src/lib/badge/_badge-theme.scss
```diff
@@ -0,0 +1,165 @@
+@import '../core/theming/palette';
+@import '../core/theming/theming';
+@import '../core/typography/typography-utils';
+
```
jelbourn: Add comment here that explains why the badge theme contains all of the styles for the badge instead of just color and typography

src/demo-app/demo-app/demo-app.ts
```diff
@@ -51,6 +51,7 @@ export class DemoApp {
   dark = false;
   navItems = [
     {name: 'Autocomplete', route: '/autocomplete'},
+    {name: 'Badge', route: '/badge'},
```
jelbourn: Routing to the badge in the demo app fails
```
Uncaught (in promise): Error: Cannot match any routes. URL Segment: 'badge'
```

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+
+let nextId = 0;
+
+export type PositionTypes =
+  'above after' |
+  'below after' |
+  'above before' |
+  'above after';
+
+export type BadgeSize =
```
jelbourn: This should be `MatBadgeSize`. Also will fit on one line

src/lib/badge/badge.ts
jelbourn: Yeah, this one should be `MatBadgePosition`. It will also fit on one line.

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  let testComponent: BadgeWithTextContent;
+  let badgeNativeElement;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [MatBadgeModule],
+      declarations: [BadgeWithTextContent],
+    });
+
+    TestBed.compileComponents();
```
jelbourn: `compileComponents` completes asynchronously, so it should be in a separate `async` beforeEach preceding the rest of the setup. It only works coincidentally now because the component currently has no `templateUrl` or `styleUrls`, but it would start failing if that ever changes.

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      fixture.detectChanges();
+      expect(badgeNativeElement.classList.contains('mat-badge-warn')).toBe(true);
+
+      testComponent.badgeColor = null;
+      fixture.detectChanges();
+
+      expect(badgeNativeElement.classList).not.toContain('mat-badge-accent');
+    });
+
+    it('should update the badget position on direction change', () => {
```
jelbourn: "baget" -> "badge"

src/lib/badge/badge.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      expect(badgeNativeElement.classList.contains('mat-badge-above')).toBe(true);
+      expect(badgeNativeElement.classList.contains('mat-badge-after')).toBe(true);
+
+      testComponent.badgeDirection = 'below before';
+      fixture.detectChanges();
+
+      expect(badgeNativeElement.classList.contains('mat-badge-below')).toBe(true);
+      expect(badgeNativeElement.classList.contains('mat-badge-before')).toBe(true);
+    });
+  });
```
jelbourn: There should also be tests for the description, overlay, size, and hidden properties.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+  /** Whether the badge is after the host or not */
+  _isAfter: boolean = true;
+
+  private _badgeElement: HTMLElement;
+
+  constructor(
+      @Optional() @Inject(DOCUMENT) private _document: any,
+      private _ngZone: NgZone,
+      private _elementRef: ElementRef,
+      private _ariaDescriber: AriaDescriber) {}
```
jelbourn: Do you plan on adding the RTL support in a follow-up PR? If so, add a TODO?

(follow-up PRs are good- smaller PRs are easier and faster to review)

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,180 @@
...
+  /** Whether the badge is above the host or not */
+  _isAbove: boolean = true;
+
+  /** Whether the badge is after the host or not */
+  _isAfter: boolean = true;
+
+  private _badgeElement: HTMLElement;
+
+  constructor(
+      @Optional() @Inject(DOCUMENT) private _document: any,
```
jelbourn: Instead of marking this as a class member w/ `private` here, we've been preferring to define the class member separately so that we can use the correct type (`Document`) throughout the rest of the code instead of `any`.

src/lib/badge/tsconfig-build.json
```diff
@@ -0,0 +1,14 @@
...
+    "../typings.d.ts"
+  ],
+  "angularCompilerOptions": {
+    "annotateForClosureCompiler": true,
+    "strictMetadataEmit": true,
+    "flatModuleOutFile": "index.js",
+    "flatModuleId": "@angular/material/badge",
+    "skipTemplateCodegen": true
+  }
+}
```
jelbourn: FYI not in this PR, but the badge will also need a bazel build file

src/lib/badge/badge.md
```diff
@@ -0,0 +1,33 @@
+Badges are small status descriptors for UI elements. A badge consists of a small circle, 
+typically containing a number or other short set of characters, that appears in proximity to another object.
```
jelbourn: Wrap at 100 columns (here and below)

src/lib/badge/badge.md
```diff
@@ -0,0 +1,33 @@
...
+background color to `primary`, `accent`, or `warn`.
+
+```html
+<mat-icon matBadge="22" matBadgeColor="accent">
+  home
+</mat-icon>
+```
+
+### Accessibility
+Badges should be given a meaningful label via `aria-describedby` attributes to augment the host element.
```
jelbourn:
```markdown
Badges should be given a meaningful description via `matBadgeDescription`. This description will be
applied, via `aria-describedby` to the element decorated by `matBadge`.
```

src/lib/badge/badge.ts
jelbourn: Yep, it should be `hidden` with an `@Input` alias

src/lib/badge/badge.ts
jelbourn: Yep, should be `size` with an `@Input` alias

src/lib/badge/badge.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Theming
+Badges can be colored in terms of the current theme using the `matBadgeColor` property to set the
+background color to `primary`, `accent`, or `warn`.
+
+```html
+<mat-icon matBadge="22" matBadgeColor="accent">
+  home
+</mat-icon>
+```
```
jelbourn: Docs should also mention customizing the size and hiding

src/lib/badge/badge.ts
amcdnl: I forgot why I did this, removed.

src/lib/badge/badge.md
amcdnl: Unclear what you are asking here, can you elaborate?

src/lib/badge/badge.ts
amcdnl: Can you elaborate? Every other instance I've found in the code does it this way.

src/lib/badge/badge.ts
amcdnl: Ya, we discussed doing that in a separate PR when we started this. Where would you like me to add the todo?

src/demo-app/badge/badge-demo.html
```diff
@@ -0,0 +1,90 @@
...
+      Aria
+    </span>
+
+    <span [matBadge]="badgeContent" matBadgeHidden="true">
+      Hidden
+    </span>
+
+    <input type="text" [(ngModel)]="badgeContent" />
+    <button (click)="visibile = !visible">Toggle</button>
```
rafaelss95: visibile -> **visible**.

src/demo-app/badge/badge-demo.ts
```diff
@@ -0,0 +1,20 @@
...
+import {Component} from '@angular/core';
+
+@Component({
+  moduleId: module.id,
+  selector: 'badge-demo',
+  templateUrl: 'badge-demo.html',
+  styleUrls: ['badge-demo.css'],
+})
+export class BadgeDemo {
+  visibile = true;
```
rafaelss95: visibile -> **visible**.

src/demo-app/demo-app/demo-module.ts
```diff
@@ -58,6 +58,7 @@ import {TypographyDemo} from '../typography/typography-demo';
 import {DemoApp, Home} from './demo-app';
 import {DEMO_APP_ROUTES} from './routes';
 import {TableDemoModule} from '../table/table-demo-module';
+import { BadgeDemo } from 'badge/badge-demo';
```
rafaelss95: Spaces between braces should be omitted.

src/demo-app/demo-app/routes.ts
```diff
@@ -50,11 +50,13 @@ import {TypographyDemo} from '../typography/typography-demo';
 import {DemoApp, Home} from './demo-app';
 import {TableDemoPage} from '../table/table-demo-page';
 import {TABLE_DEMO_ROUTES} from '../table/routes';
+import { BadgeDemo } from 'badge/badge-demo';
```
rafaelss95: Spaces between braces should be omitted.

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,168 @@
...
+
+import {Directive, Input, ElementRef, Inject, Optional, NgZone} from '@angular/core';
+import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {ThemePalette} from '@angular/material/core';
+import {AriaDescriber} from '@angular/cdk/a11y';
+import {DOCUMENT} from '@angular/common';
+
+let nextId = 0;
+
+export type MatBadgePosition = 'above after' | 'below after' | 'above before' | 'above after';
```
rafaelss95: The `above after` is duplicated. I think you want to _say_ `below before` instead?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,168 @@
...
+    '[class.mat-badge-small]': 'size === "small"',
+    '[class.mat-badge-medium]': 'size === "medium"',
+    '[class.mat-badge-large]': 'size === "large"',
+    '[class.mat-badge-hidden]': 'hidden',
+  },
+})
+export class MatBadge {
+
+  /** The color of the badge. Can be `primary`, `accent`, or `warn`. */
+  @Input('matBadgeColor')
```
rafaelss95: Question: The reason that you can't just use the `mixinColor` here is because the input has an alias?

src/lib/badge/badge.ts
```diff
@@ -0,0 +1,168 @@
...
+      this._badgeElement = this._createBadgeElement();
+    } else {
+      this._badgeElement.textContent = this.content;
+    }
+    return this._badgeElement;
+  }
+
+  /** Creates the badge element */
+  private _createBadgeElement(): HTMLElement {
+    let badgeElement = this._document.createElement('span');
```
rafaelss95: You could use `const` here.

src/lib/badge/badge.ts
amcdnl: It needs to have a separate alias for color since the host can also describe a color.

src/demo-app/demo-app/routes.ts
amcdnl: Stupid editor.

amcdnl:
@jelbourn - I think we are good now! * crosses fingers *

rafaelss95:
@amcdnl It looks like CI isn't happy yet.

src/lib/badge/badge.ts
jelbourn:
```ts
class Example {
  private _document: Document;

  constructor(@Optional() @Inject(DOCUMENT) document: any) {
    this._document = document;
  }
}
```
Everywhere else was recently updated to do this so that referencing `_document` throughout the rest of the class is type-safe.

jelbourn:
@amcdnl needs update to CODEOWNERS for lint

amcdnl:
@jelbourn - Done :)

jelbourn:
Caretaker note: this will need new build rules

mmalerba:
@amcdnl please rebase

amcdnl:
@mmalerba - Done :)

jagomf:
Is badge not going to be included in 5.2.x but only 6.0?

mmalerba:
@jagomf Yes, it's actually tagged as `release: minor`, but since there probably won't be another minor release before 6.0.0, it will likely wind up in 6.0.0

PR #7485, opened 2017-10-03
feat(snack-bar): set snack bar to be responsive

### LINKED ISSUE #3939: Material Snackbar Does Not Expand to Width of Viewport on Mobile Devices
Bug (more like missing functionality)

On mobile devices according to material design spec the toast/snackbar should be the full width of the viewport.

Toast/snackbar remains fixed width in the centered

This is missing functionality of the base MdSnackBar and can be reproduced by opening the MdSnackBar on a viewport that is considered a mobile device `this.snackBar.open(message, 'Close');`

It isn't to material design spec.

N/A

josephperrott:
Fixes #3939

### DIFF
--- a/src/lib/snack-bar/snack-bar-container.scss
+++ b/src/lib/snack-bar/snack-bar-container.scss
@@ -8,7 +8,7 @@ $mat-snack-bar-spacing-margin: 24px !default;
 
 .mat-snack-bar-container {
   border-radius: 2px;
-  box-sizing: content-box;
+  box-sizing: border-box;
   display: block;
   margin: $mat-snack-bar-spacing-margin;
   max-width: $mat-snack-bar-max-width;
@@ -42,5 +42,19 @@ $mat-snack-bar-spacing-margin: 24px !default;
 
   @include cdk-high-contrast {
     border: solid 1px;
+  }
+}
+
+/**
+ * The mat-snack-bar-handset class will be applied to the overlay
+ * element causing styling to both it and the snack bar container.
+ */
+.mat-snack-bar-handset {
+  width: 100%;
+
+  .mat-snack-bar-container {
+    margin: 0;
+    max-width: inherit;
+    width: 100%;
   }
 }
--- a/src/lib/snack-bar/snack-bar.ts
+++ b/src/lib/snack-bar/snack-bar.ts
@@ -15,6 +15,8 @@ import {SimpleSnackBar} from './simple-snack-bar';
 import {MAT_SNACK_BAR_DATA, MatSnackBarConfig} from './snack-bar-config';
 import {MatSnackBarContainer} from './snack-bar-container';
 import {MatSnackBarRef} from './snack-bar-ref';
+import {BreakpointObserver, Breakpoints} from '@angular/cdk/layout';
+import {RxChain, takeUntil, first} from '@angular/cdk/rxjs';
 
 
 /**
@@ -47,6 +49,7 @@ export class MatSnackBar {
       private _overlay: Overlay,
       private _live: LiveAnnouncer,
       private _injector: Injector,
+      private _breakpointObserver: BreakpointObserver,
       @Optional() @SkipSelf() private _parentSnackBar: MatSnackBar) {}
 
   /**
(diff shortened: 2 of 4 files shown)


src/lib/snack-bar/snack-bar-container.scss
```diff
@@ -33,6 +33,12 @@ $mat-snack-bar-spacing-margin: 24px !default;
     transform: translateY(-100%);
   }
 
+  &.mobile {
```
crisbeto: A few notes:

* `.mobile` seems like something that could clash with user styles. What about calling it `mat-snackbar-mobile`? 
* Renaming the class will let you avoid the nesting.
* Is there a particular reason for going with viewport units rather than percent? I have a vague memory about `100vh` causing content to go behind the scrollbar on Windows.

josephperrott: Yep, definitely forgot to rename the class after I got it working.

Because the overlay container does not establish itself as the entire width of the viewport, `width: 100%` turns into the width of the child based on content, so we use `100vw` .

crisbeto: I see. That seems to go behind the scrollbar though. Here's a `position: fixed` div with a width of `100vw`:



You might be able to work around it by adding a class to the overlay panel via the `panelClass` and styling it from there.

josephperrott: The issue that we run into with changing the width of the overlay container, is that the snackbar will always stay in its most expanded state (as limited by `max-width`)

I will have to continue to investigate

josephperrott:
@crisbeto Comments addresed.  I moved the observation to `MatSnackbar`, modifying a class in the overlay element with the class effecting styling in both the overlay element and the snackbar container

Feel free to take a look when you have a chance.

src/lib/snack-bar/snack-bar.ts
```diff
@@ -145,6 +148,18 @@ export class MatSnackBar {
     // We can't pass this via the injector, because the injector is created earlier.
     snackBarRef.instance = contentRef.instance;
 
+    // Subscribe to the breakpoint observer and attacg the mat-snack-bar-handset class as
```
Chris1308: Small typo :)

src/lib/snack-bar/snack-bar.ts
```diff
@@ -145,6 +148,18 @@ export class MatSnackBar {
...
+    // appropriate. This class is applied to the overlay element because the overlay must expand to
+    // fill the width of the screen for full width snackbars.
+    const isHandset = this._breakpointObserver.observe(Breakpoints.Handset).subscribe(state => {
+      overlayRef.overlayElement.classList.toggle('mat-snack-bar-handset', state.matches);
```
crisbeto: The second param for `classList.toggle` doesn't work on IE11. This should use `add`/`remove` instead.

src/lib/snack-bar/snack-bar.ts
```diff
@@ -145,6 +148,18 @@ export class MatSnackBar {
...
+    });
+    // Unsubscribe from the isHandset subscription when the overlay is detached from the view.
+    const overlayDetach = overlayRef.detachments().subscribe(() => {
+      overlayDetach.unsubscribe();
```
crisbeto: A cleaner approach could be to use `takeUntil`. E.g.

```
this._breakpointObserver.observe(Breakpoints.Handset)
  .takeUntil(overlayRef.detachments().first())
  .subscribe(() => ...);
```

This way you don't have to keep track of subscriptions.

josephperrott: Good call

src/lib/snack-bar/snack-bar.ts
```diff
@@ -145,6 +148,19 @@ export class MatSnackBar {
     // We can't pass this via the injector, because the injector is created earlier.
     snackBarRef.instance = contentRef.instance;
 
+    // Subscribe to the breakpoint observer and attach the mat-snack-bar-handset class as
+    // appropriate. This class is applied to the overlay element because the overlay must expand to
+    // fill the width of the screen for full width snackbars.
+    RxChain.from(this._breakpointObserver.observe(Breakpoints.Handset))
+      .call(takeUntil, RxChain.from(overlayRef.detachments()).call(first).result())
```
crisbeto: For a single operator you don't really need the `RxChain`, you can just `.call` it directly. E.g. `first.call(overlayRef.detachments())`.

josephperrott: I did find that frustrating to have it in `RxChain` again, not sure why I didn't think to look at calling it this way.

PR #7488, opened 2017-10-03
feat(dialog): support minWidth, minHeight, maxWidth and maxHeight

### LINKED ISSUE #3209: dialog max width is set to 80 vw
#### Bug:
can't have more than 80% width for dialog

have the width to cover more than 80%

the maximum width is to 80% of the viewport

"@angular/material": "2.0.0-beta.2"

I saw that the class of the element "md-dialog-container" is set to "mat-dialog-container". the class has style max-width: 80vh. This cause the issue

### LINKED ISSUE #6024: Feature Request - MdDialog should support max-height, max-width, min-height, and min-width

Feature request

In addition to just width and height, MdDialog should support max-height, max-width, min-height, and min-width.

Max & min dimensions allow the dialog to be more responsive.

emilio-martinez:
Fixes #6024, Fixes #3209
* Adds Dialog support for `minWidth`, `minHeight`, `maxWidth` and `maxHeight` via config. Mostly delegates to Overlay.
* Moves declared `max-width` on `.mat-dialog-container` stylesheet to be authored via `MatDialogConfig`, providing the same default `80vw` value. Without this change, there would likely be undesired layout results due to different constraints being set on the overlay container vs the nested the dialog container.
* Added note on `minHeight` and `maxHeight` regarding the potential need to also set a `height` due to the rules around computed height resolution (https://www.w3.org/TR/CSS2/visudet.html#min-max-widths).  Any value set on `height` as a default would probably be assuming too much and may have varying results across browsers.

### DIFF
--- a/src/demo-app/dialog/dialog-demo.ts
+++ b/src/demo-app/dialog/dialog-demo.ts
@@ -1,7 +1,21 @@
 import {Component, Inject, ViewChild, TemplateRef} from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
-import {MatDialog, MatDialogConfig, MatDialogRef, MAT_DIALOG_DATA} from '@angular/material';
+import {
+  DialogPosition,
+  MatDialog,
+  MatDialogConfig,
+  MatDialogRef,
+  MAT_DIALOG_DATA
+} from '@angular/material';
 
+/**
+ * Appeases the AOT type checker for the template by extending `MatDialogConfig`
+ * to disallow `position` from being undefined, given that it’s optional.
+ * Guards cannot be set on the template because the values are bound to NgModel.
+ */
+interface MatDemoDialogConfig extends MatDialogConfig {
+  position: DialogPosition;
+}
 
 @Component({
   moduleId: module.id,
@@ -14,7 +28,7 @@ export class DialogDemo {
   lastAfterClosedResult: string;
   lastBeforeCloseResult: string;
   actionsAlignment: string;
-  config: Partial<MatDialogConfig> = {
+  config: MatDemoDialogConfig = {
     disableClose: false,
     panelClass: 'custom-overlay-pane-class',
     hasBackdrop: true,
--- a/src/demo-app/dialog/dialog-demo.ts
+++ b/src/demo-app/dialog/dialog-demo.ts
@@ -1,6 +1,6 @@
 import {Component, Inject, ViewChild, TemplateRef} from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
-import {MatDialog, MatDialogRef, MAT_DIALOG_DATA} from '@angular/material';
+import {MatDialog, MatDialogConfig, MatDialogRef, MAT_DIALOG_DATA} from '@angular/material';
 
 
 @Component({
@@ -14,7 +14,7 @@ export class DialogDemo {
   lastAfterClosedResult: string;
   lastBeforeCloseResult: string;
   actionsAlignment: string;
-  config = {
+  config: Partial<MatDialogConfig> = {
     disableClose: false,
     panelClass: 'custom-overlay-pane-class',
     hasBackdrop: true,
--- a/src/demo-app/dialog/dialog-demo.ts
+++ b/src/demo-app/dialog/dialog-demo.ts
@@ -1,21 +1,8 @@
 import {Component, Inject, ViewChild, TemplateRef} from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
-import {
-  DialogPosition,
-  MatDialog,
-  MatDialogConfig,
-  MatDialogRef,
-  MAT_DIALOG_DATA
-} from '@angular/material';
+import {MatDialog, MatDialogConfig, MatDialogRef, MAT_DIALOG_DATA} from '@angular/material';
 
-/**
- * Appeases the AOT type checker for the template by extending `MatDialogConfig`
- * to disallow `position` from being undefined, given that it’s optional.
- * Guards cannot be set on the template because the values are bound to NgModel.
- */
-interface MatDemoDialogConfig extends MatDialogConfig {
-  position: DialogPosition;
-}
+const defaultDialogConfig = new MatDialogConfig();
 
 @Component({
   moduleId: module.id,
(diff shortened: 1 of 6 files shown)


emilio-martinez:
Screenshot tests seem to be failing for components that—as far as I can tell—are unrelated to what this PR addresses. Screenshot test flagged: `mdcard`, `mdtoolbar`, and `stepper`.

src/demo-app/dialog/dialog-demo.ts
```diff
@@ -1,7 +1,21 @@
...
+import {
+  DialogPosition,
+  MatDialog,
+  MatDialogConfig,
+  MatDialogRef,
+  MAT_DIALOG_DATA
+} from '@angular/material';
+
+/**
+ * Appeases the AOT type checker for the template by extending `MatDialogConfig`
```
crisbeto: I would rather not add a separate type just for the sake of AoT. You can remove the type from the `config`. It'll still be type-checked when it's passed in to `dialog.open`.

src/lib/dialog/dialog-config.ts
```diff
@@ -57,6 +57,26 @@ export class MatDialogConfig {
   /** Height of the dialog. */
   height?: string = '';
 
+  /** Min-width of the dialog. */
+  minWidth?: string = '';
+
+  /**
+   * Min-height of the dialog.
+   * For this value to be effectively applied, `height` may also need to be defined.
+   * See https://www.w3.org/TR/CSS2/visudet.html#min-max-widths
```
crisbeto: I don't think that this needs to link to the CSS spec. Something along the lines of `/** min-height of the dialog */` should suffice. Same goes for the `maxHeight`.

src/lib/dialog/dialog-config.ts
```diff
@@ -57,6 +57,26 @@ export class MatDialogConfig {
...
+   */
+  minHeight?: string = '';
```
crisbeto: These don't need to be initialized.

src/demo-app/dialog/dialog-demo.ts
emilio-martinez: Correct, `dialog.open` would type-check, but not having type results in AOT errors because the template would only have the inferred type to go off from:
`Error at /home/travis/build/angular/material2/dist/packages/demo-app/dialog/dialog-demo.ngfactory.ts:526:51: Property 'minWidth' does not exist on type '{ disableClose: boolean; panelClass: string; hasBackdrop: boolean; backdropClass: string; width: ...'.`
See https://travis-ci.org/angular/material2/jobs/282614008.

crisbeto: You can get it to play nice if you initialize the `minWidth` to something.

emilio-martinez: @crisbeto another alternative would be to just type with `MatDialogConfig` without extending, but given strictNullChecks, AOT will complain in the template about `config.property` possibly being `undefined`, so something like the below could be done. Let me know your thoughts.
```html
<ng-template [ngIf]="config.position">
  <p>
    <mat-form-field>
      <input matInput [(ngModel)]="config.position.top" (change)="config.position.bottom = ''" placeholder="Top">
    </mat-form-field>
    <mat-form-field>
      <input matInput [(ngModel)]="config.position.bottom" (change)="config.position.top = ''" placeholder="Bottom">
    </mat-form-field>
  </p>

  <p>
    <mat-form-field>
      <input matInput [(ngModel)]="config.position.left" (change)="config.position.right = ''" placeholder="Left">
    </mat-form-field>
    <mat-form-field>
      <input matInput [(ngModel)]="config.position.right" (change)="config.position.left = ''" placeholder="Right">
    </mat-form-field>
  </p>
</ng-template>
```

src/lib/dialog/dialog-config.ts
emilio-martinez: Addressed. What are your thoughts on initializing `maxWidth`? I left that initialized for the time being.

src/demo-app/dialog/dialog-demo.ts
crisbeto: I think that's over-complicating it. Initializing the new properties to empty strings should be fine.

src/lib/dialog/dialog-config.ts
emilio-martinez: Done.

src/demo-app/dialog/dialog-demo.ts
emilio-martinez: Agreed re: initializing the `config` props. The only question I'd have is around setting the value for `maxWidth` since I'm initializing with `80vw` on DialogConfig. Initializing it to something neutral like an empty string would override the value.

To your point about initializing, it could be as simple as initializing all min/max values with an empty string and initializing `maxWidth` with the `80vw`, or extracting the default value out of a new `DialogConfig` and setting that in the `maxWidth` init.

crisbeto: Initializing all of them to an empty string (only in the demo) and the `maxWidth` to the same as the default is fine by me.

emilio-martinez: Cool. Addressed in 2b1c7334157c16e0195557de2f537f24d3e27f6f

emilio-martinez:
@crisbeto tests passing after your feedback. Let me know if you have any further comments.

src/lib/dialog/dialog-config.ts
```diff
@@ -57,6 +57,18 @@ export class MatDialogConfig {
...
+  minWidth?: string;
+
+  /** Min-height of the dialog */
+  minHeight?: string;
+
+  /** Max-width of the dialog. */
+  maxWidth?: string = '80vw';
+
+  /** Max-height of the dialog */
+  maxHeight?: string;
```
jelbourn: These should all be `string | number`, where a number implies a pixel value to match the OverlayConfig values

src/lib/dialog/dialog-config.ts
```diff
@@ -57,6 +57,18 @@ export class MatDialogConfig {
   /** Height of the dialog. */
   height?: string = '';
 
+  /** Min-width of the dialog. */
+  minWidth?: string;
+
+  /** Min-height of the dialog */
+  minHeight?: string;
+
+  /** Max-width of the dialog. */
```
jelbourn: This comment should include `Defaults to 80vw`

emilio-martinez: Done.

src/lib/dialog/dialog-config.ts
emilio-martinez: Done.

PR #7500, opened 2017-10-03
refactor(dialog): simplify dialog config typing

crisbeto:
Uses a `Partial` when typing the parameters of `MdDialog.open`. The advantage of `Partial` is that we don't have to mark all fields of the `MdDialogConfig` as optional.

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -31,43 +31,43 @@ export class MatDialogConfig {
    * component instantiated inside of the dialog. This does not affect where the dialog
    * content will be rendered.
    */
-  viewContainerRef?: ViewContainerRef;
+  viewContainerRef: ViewContainerRef;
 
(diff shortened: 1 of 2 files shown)


jelbourn:
I think this will affect the dgeni output- we still want the API docs to just have `MatDialogConfig`

crisbeto:
It's not an `MatDialogConfig` though, it's a handful of properties from it.

crisbeto:
`Partial` is a built-in type though (granted that a lot of people probably don't know about it). Does Dgeni allow for the param to be aliased? E.g. `@param config {MatDialogConfig} ...`?

willshowell:
FWIW, I like the idea of having `Partial<MatDialogConfig>` in the docs. It may be lesser-known, but it's kind of self explanatory and easy to look up.

jelbourn:
I chatted w/ devrel and they agree that `Partial` is harder for less-sophisticated developers to understand, so going to stick with the optional properties.

PR #7504, opened 2017-10-03
fix(snack-bar): add content fade in animation

### LINKED ISSUE #7502: bug(Snackbar): Content of snackbar has no fade in animation
Something between a bug and a feature request.

The content of a snackbar should fade in while the snackbar slides into the viewport, as shown in the MD guidelines. (See [this](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B6Okdz75tqQsLVVnZlF4UEtKRU0/components-snackbar-specs-fabtablet-002.webm) or [this](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B6Okdz75tqQsbjV1R1B1cm1vZE0/components-snackbars-usage-consecutivesbdifferenttype-005.mp4) for reference.)

The content of the snackbar doesn't fade in.

Just have a look on the [demo snackbar from the docs](https://material.angular.io/components/snack-bar/overview).

To be in line with the MD guidelines.

crisbeto:
Adds a subtle fade in animation to the snack bar content.

Fixes #7502.

### DIFF
--- a/src/lib/snack-bar/simple-snack-bar.ts
+++ b/src/lib/snack-bar/simple-snack-bar.ts
@@ -7,6 +7,7 @@
  */
 
 import {Component, ViewEncapsulation, Inject, ChangeDetectionStrategy} from '@angular/core';
+import {trigger, state, style, transition, animate} from '@angular/animations';
 import {MatSnackBarRef} from './snack-bar-ref';
 import {MAT_SNACK_BAR_DATA} from './snack-bar-config';
 
@@ -23,7 +24,15 @@ import {MAT_SNACK_BAR_DATA} from './snack-bar-config';
   encapsulation: ViewEncapsulation.None,
   preserveWhitespaces: false,
   changeDetection: ChangeDetectionStrategy.OnPush,
+  animations: [
+    trigger('contentFade', [
+      state('void', style({opacity: '0'})),
+      state('visible', style({opacity: '1'})),
+      transition('void => visible', animate('350ms linear'))
+    ])
+  ],
   host: {
+    '[@contentFade]': '"visible"',
     'class': 'mat-simple-snackbar',
   }
 })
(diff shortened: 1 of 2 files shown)


src/lib/snack-bar/simple-snack-bar.ts
```diff
@@ -23,7 +24,15 @@ import {MAT_SNACK_BAR_DATA} from './snack-bar-config';
   encapsulation: ViewEncapsulation.None,
   preserveWhitespaces: false,
   changeDetection: ChangeDetectionStrategy.OnPush,
+  animations: [
+    trigger('contentFade', [
+      state('void', style({opacity: '0'})),
```
josephperrott: I think you can just use `:enter` instead of having to manage states.

src/lib/snack-bar/simple-snack-bar.ts
```diff
@@ -23,7 +24,15 @@ import {MAT_SNACK_BAR_DATA} from './snack-bar-config';
...
+      state('visible', style({opacity: '1'})),
+      transition('void => visible', animate('350ms linear'))
```
josephperrott: Should we use the animations from core?

```
`${AnimationCurves.STANDARD_CURVE} ${AnimationDurations.ENTERING}`
```

willshowell: Does that work now with AOT? I thought string interpolation in animation [broke it](https://github.com/angular/material2/blob/master/src/lib/snack-bar/snack-bar-container.ts#L41-L42).

crisbeto: That comment is pretty old. It seems to work with AoT now.

crisbeto:
Addressed the feedback @josephperrott, PTAL. One note: I'm not too sure about using the `COMPLEX` animation duration, but all of the others were too short.

josephperrott:
Yeah, what we likely would like to do is delay the start of the animation by ~100ms so that it times better, but I think that increasing the overall length will work just as well.

kara:
@crisbeto Linter?

crisbeto:
Fixed.

PR #7506, opened 2017-10-03
fix(snack-bar): use mat-button like styling in simple snack bar

### LINKED ISSUE #7505: feature request(Snackbar): Give user feedback when pressing/clicking the action button
feature request (As there are no further instructions in the MD guidelines about this, I opened this as a feature request.)

There should be a feedback for the user when pressing or clicking the action button on a snackbar, e.g. a ripple effect on the button.

There is no feedback when pressing or clicking the action button.

Just have a look on the [demo snackbar from the documentation](https://material.angular.io/components/snack-bar/overview).

To be in line with the MD guidelines.

josephperrott:
Fixes #7505

### DIFF
--- a/src/lib/snack-bar/simple-snack-bar.scss
+++ b/src/lib/snack-bar/simple-snack-bar.scss
@@ -2,24 +2,36 @@
 @import '../core/style/button-common';
 @import '../core/style/list-common';
 
-$mat-snack-bar-button-margin: 48px !default;
+$mat-snack-bar-button-horizontal-margin: 8px !default;
+$mat-snack-bar-button-height: 36px !default;
+$mat-snack-bar-line-height: 20px !default;
+// Button vertical margin is used to ensure that a button height of 36px, when the containing
+// space falls below 36px.
+$mat-snack-bar-button-vertical-margin:
+  -(($mat-snack-bar-button-height - $mat-snack-bar-line-height) / 2);
 
 .mat-simple-snackbar {
   display: flex;
   justify-content: space-between;
-  line-height: 20px;
+  line-height: $mat-snack-bar-line-height;
   opacity: 1;
 }
 
 .mat-simple-snackbar-action {
-  @include mat-button-reset;
-
-  background: none;
+  display: flex;
+  flex-direction: column;
   flex-shrink: 0;
-  margin-left: $mat-snack-bar-button-margin;
+  justify-content: space-around;
+  margin: $mat-snack-bar-button-vertical-margin 0
+          $mat-snack-bar-button-vertical-margin $mat-snack-bar-button-horizontal-margin;
+
+  button {
+    flex: 1;
+    max-height: $mat-snack-bar-button-height;
+  }
 
   [dir='rtl'] & {
-    margin-right: $mat-snack-bar-button-margin;
     margin-left: 0;
+    margin-right: $mat-snack-bar-button-horizontal-margin;
   }
 }
--- a/src/lib/snack-bar/BUILD.bazel
+++ b/src/lib/snack-bar/BUILD.bazel
@@ -12,6 +12,7 @@ ng_module(
     "simple_snack_bar_css",
   ],
   deps = [
+    "//src/lib/button",
     "//src/lib/core",
     "//src/cdk/a11y",
     "//src/cdk/layout",
--- a/src/lib/snack-bar/simple-snack-bar.html
+++ b/src/lib/snack-bar/simple-snack-bar.html
@@ -1,6 +1,5 @@
 {{data.message}}
 
-<button
-  class="mat-simple-snackbar-action"
-  *ngIf="hasAction"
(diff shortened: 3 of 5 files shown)


src/lib/snack-bar/simple-snack-bar.scss
```diff
@@ -16,6 +16,14 @@ $mat-snack-bar-button-margin: 48px !default;
   background: none;
```
crisbeto: It doesn't let me set a comment on the proper line, but you should be able to get rid of the `@include mat-button-reset` a little bit above.

src/lib/snack-bar/simple-snack-bar.scss
```diff
@@ -16,6 +16,14 @@ $mat-snack-bar-button-margin: 48px !default;
...
   flex-shrink: 0;
   margin-left: $mat-snack-bar-button-margin;
+  line-height: 20px;
```
crisbeto: Is this used for centering? Couldn't we use flexbox instead?

src/lib/snack-bar/simple-snack-bar.scss
```diff
@@ -16,6 +16,14 @@ $mat-snack-bar-button-margin: 48px !default;
...
+
+  .mat-button-ripple,
+  .mat-button-focus-overlay {
+    height: 36px;
```
crisbeto: Why does this have to override the `height` and `top`? AFAIK the overlay and ripple are supposed to be `width: 100%` and `height: 100%` of the button?

josephperrott: The issue that I am seeing is that the size of the button is supposed to fall outside  of the content box of the `mat-simple-snackbar`.

The snackbar provides `20px` of space in a single line, but the button is supposed to be `36px`.  The "best" solution I think would actually be to move the padding-like spacing from the container to the `mat-simple-snackbar`.   Thoughts?

src/lib/snack-bar/simple-snack-bar.scss
josephperrott: See other comment

src/lib/snack-bar/simple-snack-bar.scss
crisbeto: You should be able to work around that by using a negative margin. E.g. if you remove the size and position overrides of the ripple and focus overlay, as well as the `line-height: 20px` from the button and add something like `margin: -14px -24px -14px 48px` you can get the button to overlap the container padding. You might have to tweak the spacing a bit.

josephperrott:
@crisbeto Take another look, I ended up using margins and flex to handle both the single and multi line cases.

src/lib/snack-bar/simple-snack-bar.scss
```diff
@@ -11,14 +11,19 @@ $mat-snack-bar-button-margin: 48px !default;
...
 .mat-simple-snackbar-action {
-  @include mat-button-reset;
-
-  background: none;
+  display: flex;
+  flex-direction: column;
   flex-shrink: 0;
-  margin-left: $mat-snack-bar-button-margin;
+  justify-content: space-around;
+  margin: -8px 0 -8px $mat-snack-bar-button-margin;
```
crisbeto: Use the variable here?

src/lib/snack-bar/simple-snack-bar.scss
```diff
@@ -11,14 +11,19 @@ $mat-snack-bar-button-margin: 48px !default;
...
+
+  button {
+    flex: 1;
+    max-height: 36px;
```
crisbeto: Move this into a variable or add a comment about where it's coming from?

src/lib/snack-bar/simple-snack-bar.scss
josephperrott: I can use a variable, however they are different margins so to speak.

The top and bottom margins are `-8px` so that the total line height adds back up to `36px` (20 + 8 + 8).  The left and right margins are for spacing within the container according to spec.


Should I create another variable?  Or comment to explain it?

crisbeto: Having another variable should be fine.

src/lib/snack-bar/simple-snack-bar.scss
josephperrott: Done

src/lib/snack-bar/simple-snack-bar.scss
josephperrott: Done

jelbourn:
Investigating whether we want to add a ripple instead of bringing in all of MatButton

josephperrott:
After investigation it was determined that adding ripple manually would essentially amount to recreating `MatButton` within snack bar.

jelbourn:
@josephperrott this needs to update the bazel rule now

jelbourn:
Caretaker note: needs a simple build file update

PR #7507, opened 2017-10-03
fix(form-field): move label before input

### LINKED ISSUE #7402: mat-input DOM order of label and input
Proposal

It would be ideal if the label came before the input in the DOM order for the mat-input. Useful for adding your own custom styles if you want the functionality default look of inputs. This would also be useful if CSS were disabled.

Label comes after input, however it has the appropriate aria labels and markup to associate the label with the input.

Plunker starter (using on `@master`): https://goo.gl/DlHd6U<br/>

*DOM order*  
image   

*CSS disabled*  
image  

For applying custom styling to make form fields look like traditional inputs can cause complications when DOM order is flipped. Also useful if CSS was disabled.

Material 2.0.0-beta.11

crisbeto:
Moves the form field label before the input in the DOM order which is aligned closer with the way native labels and inputs are used. It also falls back better if CSS is disabled/fails to load.

Fixes #7402.

### DIFF
--- a/src/lib/form-field/form-field.html
+++ b/src/lib/form-field/form-field.html
@@ -6,8 +6,6 @@
     </div>
 
     <div class="mat-input-infix mat-form-field-infix">
-      <ng-content></ng-content>
-
       <span class="mat-input-placeholder-wrapper mat-form-field-placeholder-wrapper">
         <!-- We add aria-owns as a workaround for an issue in JAWS & NVDA where the label isn't
              read if it comes before the control in the DOM. -->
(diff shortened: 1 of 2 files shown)


mmalerba:
I think this will break the float on autofill behavior

crisbeto:
It still works locally, but I think it's because all the inputs that we have on the demo are without an `NgControl` which causes them to dirty-check the value. It'll probably break for the cases where they do have an `NgControl` and it's not worth it to start dirty-checking just to support this case. Closing.

PR #7509, opened 2017-10-03
bug(chips): fixes placeholder in chip inputs #7481

### LINKED ISSUE #7481: chips: placeholder disappears
Placeholder disappears after typing 2 characters into chip input

Placeholder doesn't disappear

Placeholder disappears

1. Go to this plunker: http://plnkr.co/edit/qHQlWKfGSyvIhgm6MUop?p=preview
2. Type 2 characters into the input

Existing behavior is incorrect

master

amcdnl:
This PR resolves #7481 by:

- Fixes incorrect logic for when placeholder should float
- Fixes chips from pushing input out of view and not accessible to user

### DIFF
--- a/src/lib/chips/chip-list.spec.ts
+++ b/src/lib/chips/chip-list.spec.ts
@@ -353,7 +353,7 @@ describe('MatChipList', () => {
 
     });
 
-    fit('should float placeholder if chip is selected', () => {
+    it('should float placeholder if chip is selected', () => {
       expect(formField.classList.contains('mat-form-field-should-float'))
         .toBe(true, 'placeholder should be floating');
     });
--- a/src/lib/chips/chip-list.spec.ts
+++ b/src/lib/chips/chip-list.spec.ts
@@ -353,9 +353,9 @@ describe('MatChipList', () => {
 
     });
 
-    it('should not float placeholder if no chip is selected', () => {
+    fit('should float placeholder if chip is selected', () => {
       expect(formField.classList.contains('mat-form-field-should-float'))
-        .toBe(false, 'placeholder should not be floating');
+        .toBe(true, 'placeholder should be floating');
     });
 
     it('should remove selection if chip has been removed', async(() => {
--- a/src/lib/chips/chip-list.ts
+++ b/src/lib/chips/chip-list.ts
@@ -240,7 +240,7 @@ export class MatChipList implements MatFormFieldControl<any>, ControlValueAccess
   }
 
   get shouldPlaceholderFloat(): boolean {
-    return this.empty;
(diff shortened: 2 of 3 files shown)


src/lib/chips/chips.scss
```diff
@@ -19,6 +19,16 @@ $mat-chip-input-margin: 3px;
   align-items: baseline;
 }
 
+.mat-form-field-type-mat-chip-list {
```
mmalerba: should not be necessary, the user docs are apparently out of date, chips should be done like this now, which resolves the wrapping issues: http://plnkr.co/edit/kc3bxAZRgHS0JbGl3qSX?p=preview

src/lib/chips/chip-list.ts
```diff
@@ -240,7 +240,7 @@ export class MatChipList implements MatFormFieldControl<any>, ControlValueAccess
   }
 
   get shouldPlaceholderFloat(): boolean {
-    return this.empty;
+    return !this.empty;
```
mmalerba: `|| this.focused`

amcdnl: Done.

src/lib/chips/chips.scss
amcdnl: Done. I'll make a PR to update the docs too :)

mmalerba:
Can we also update any examples that still show the old way of doing it?

kara:
@amcdnl Can you check out the CI failures?

amcdnl:
@kara - done :)

src/lib/chips/chip-list.spec.ts
```diff
@@ -353,9 +353,9 @@ describe('MatChipList', () => {
 
     });
 
-    it('should not float placeholder if no chip is selected', () => {
+    it('should float placeholder if chip is selected', () => {
```
mmalerba: I think this was testing when there's a chip present but not selected, so it should still be `"... if no chip is selected"`. 

@tinayuangao Why was this expected to be not floating in the first place? It seems like we would want it to float even if the chip isn't selected, to avoid overlap.

amcdnl: This test was invalid, it was expecting the inverse logic that was the bug before.

mmalerba: agreed, but it was the "not float" part that was invalid not the "no chip is selected" part

asmith2306:
Hi, has this fix made into any of the releases yet?  I can't see it in the bug fixes listings for any release.  I'm still on "@angular/material": "2.0.0-beta.12", which contains this bug.

Thanks

Issue #6973, opened 2017-09-10
`MdDatepicker` and `MdDatepickerToggle` should be disabled when its input associated is `disabled` (reactive forms)


rafaelss95:
#### Bug, feature request, or proposal:
Bug.

#### What is the expected behavior?
`MdDatepicker` and `MdDatepickerToggle` should be disabled when its input associated is `disabled` using ReactiveForms.

#### What is the current behavior?
`MdDatepicker` and `MdDatepickerToggle` aren't being disabled when its input associated is `disabled` using ReactiveForms.

#### What are the steps to reproduce?

Providing a Plunker (or similar) is the *best* way to get the team to see your issue.
https://plnkr.co/edit/9GaotKDxIFFDKkGpDess?p=info

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
  ```
  _                      _                 ____ _     ___
   / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
  / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
 / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
/_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
               |___/
@angular/cli: 1.4.1
node: 8.1.1
@angular/animations: 4.3.6
@angular/cdk: 2.0.0-beta.10
@angular/common: 4.3.6
@angular/compiler: 4.3.6
@angular/core: 4.3.6
@angular/forms: 4.3.6
@angular/http: 4.3.6
@angular/material: 2.0.0-beta.10
@angular/platform-browser: 4.3.6
@angular/platform-browser-dynamic: 4.3.6
@angular/platform-server: 4.3.6
@angular/router: 4.3.6
@angular/cli: 1.4.1
@angular/compiler-cli: 4.3.6
@angular/language-service: 4.3.6
typescript: 2.4.1
```

#### Is there anything else we should know?
It's working using Template Driven Forms, [as you can see here](https://plnkr.co/edit/gpldSaRYwQDRKIHMOsHu?p=preview)... well almost 100%, see #6974.

rafaelss95:
cc @mmalerba

mmalerba:
updated the plunker to work with current HEAD, I am still able to reproduce the bug, will look into it: https://plnkr.co/edit/85tszhB6YIqMo46xhRvY?p=preview

PR #7514, opened 2017-10-03
fix(datepicker): use disabled state from FormControl

mmalerba:
fixes https://github.com/angular/material2/issues/6973

### DIFF
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -147,7 +147,7 @@ export class MatDatepickerInput<D> implements AfterContentInit, ControlValueAcce
 
   /** Whether the datepicker-input is disabled. */
   @Input()
-  get disabled() { return this._disabled; }
+  get disabled() { return !!this._disabled; }
   set disabled(value: any) {
     const newValue = coerceBooleanProperty(value);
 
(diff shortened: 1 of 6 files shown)


src/lib/datepicker/datepicker-toggle.ts
```diff
@@ -44,39 +45,43 @@ export class MatDatepickerToggle<D> implements OnChanges, OnDestroy {
   /** Whether the toggle button is disabled. */
   @Input()
   get disabled() {
-    return this._disabled === undefined ? this.datepicker.disabled : this._disabled;
+    return this._disabled === undefined ? this.datepicker.disabled : !!this._disabled;
```
rafaelss95: Shouldn't be `this._disabled == null`?

mmalerba: Either would work

rafaelss95: So this `disabled` is never `null`? Because if `disabled` is `null` this comparison (`null === undefined`) won't work.

src/lib/datepicker/datepicker-input.ts
```diff
@@ -147,7 +147,7 @@ export class MatDatepickerInput<D> implements AfterContentInit, ControlValueAcce
 
   /** Whether the datepicker-input is disabled. */
   @Input()
-  get disabled() { return this._disabled; }
+  get disabled() { return !!this._disabled; }
```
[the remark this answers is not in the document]
rafaelss95: Aha, thanks for explaining :)

src/lib/datepicker/datepicker-toggle.ts
mmalerba: `coerceBooleanProperty` will make sure its never null. The only reason it can be `undefined` is because we don't set an initial value for `_disabled`, so if there's no property binding for `disabled` the setter is never called and it remains `undefined`

kara:
@mmalerba Rebase?

mmalerba:
done

PR #7535, opened 2017-10-04
fix(icon): use SafeResourceUrl in getSvgIconFromUrl

jelbourn:


src/lib/icon/icon-registry.ts
```diff
@@ -201,7 +201,7 @@ export class MatIconRegistry {
       return observableOf(cloneSvg(cachedIcon));
     }
 
-    return RxChain.from(this._loadSvgIconFromConfig(new SvgIconConfig(url)))
+    return RxChain.from(this._loadSvgIconFromConfig(new SvgIconConfig(safeUrl)))
```
crisbeto: While we're at it, it might be a good idea to rename the `url` variable above to something along the lines of `sanitizedUrl` to avoid confusion in the future.

jelbourn: Really the _type_ should enforce passing the right thing around; I want to check with Martin about adding a brand to safe types (since they're empty they match everything)

kara:
Looks like we still need to update a test here: https://travis-ci.org/angular/material2/jobs/283345030

jelbourn:
Weird, failing only in IE11

src/lib/icon/icon.spec.ts
```diff
@@ -153,7 +153,14 @@ describe('MatIcon', () => {
       svgElement = verifyAndGetSingleSvgChild(matIconElement);
       verifyPathChildElement(svgElement, 'woof');
       expect(httpRequestUrls).toEqual(['dog.svg', 'cat.svg']);
-    });
+
+      // Assert that a registered icon can be looked-up by url.
+      iconRegistry.getSvgIconFromUrl(trust('cat.svg')).subscribe(element => {
+        expect(element.innerHTML).toContain('meow');
```
crisbeto: SVGs don't have `innerHTML` on IE. This should be `verifyPathChildElement(element, 'meow');`.

PR #7552, opened 2017-10-05
Highlight alternative Angular Material/CDK install

cgrail:
I've just spent hours to find an issue with my first Angular Material app. I didn't read the first sentence of the first step and was just copy/pasting npm install commands. I got a lot of weird issues and couldn't get a simple app running. Eventually I figured out I made a mistake in the first step, but only after a few hours. As installing both modules can really screw up everything, I think that highlighting the choice might be a good idea.


cgrail:
I signed it

PR #7557, opened 2017-10-05
revert(ripple): handle touch events

DevVersion:
Reverts the `ripple(): handle touch events` commit from PR #7299. With the current touch solution, ripples are showing up twice on click.

This is because after a `touchstart` event, the touch browser fires a `mousedown` event. This causes the ripple renderer to fade in another ripple from the `mousedown` event.

There are solutions like:

* Timeout to ignore `mousedown` events
* Calling `preventDefault` / `stopPropagation` on `touchstart`
* Listening to `pointerdown`, `pointerup`, `pointerleave` events
* Using feature detecton (as in Modernizr)

All of those solutions have negative aspects on the UX of the ripples and need to be evaluated with more testing.

For now the ripples are broken and should be fixed as soon as possible. This is going to be revisited.

### DIFF
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -40,8 +40,8 @@ export class RippleRenderer {
   /** Element which triggers the ripple elements on mouse events. */
   private _triggerElement: HTMLElement | null;
 
-  /** Whether the pointer is currently being held on the trigger or not. */
-  private _isPointerDown: boolean = false;
+  /** Whether the mouse is currently down or not. */
+  private _isMousedown: boolean = false;
 
   /** Events to be registered on the trigger element. */
   private _triggerEvents = new Map<string, any>();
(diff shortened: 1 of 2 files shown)


src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -182,14 +178,18 @@ export class RippleRenderer {
   /** Function being called whenever the trigger is being pressed. */
   private onMousedown(event: MouseEvent) {
     if (!this.rippleDisabled) {
-      this._isPointerDown = true;
-      this.fadeInRipple(event.clientX, event.clientY, this.rippleConfig);
+      this._isMousedown = true;
+      this.fadeInRipple(event.pageX, event.pageY, this.rippleConfig);
```
crisbeto: Sorry, just took another look. This should use `clientX` and `clientY`.

DevVersion: Good catch

DevVersion: Updated it and fixed a lint error. Please have another look.

PR #7560, opened 2017-10-05
perf: remove persistent global scroll listener

### LINKED ISSUE #6882: Performance issue:  modals & popups
Bug

There should be no performance hit when using popups and modals

There is a performance hit when using popups and modals

http://plnkr.co/edit/OeUkGO1w4suNVLCQceEL

Angular Material has several performance issues which I think should be addressed.
The web is already slow, no need to make it any slower.

@angular/material: ^2.0.0-beta.10
Chrome 61

There is a big performance problem related to the modals & popups, I can't figure out what it is. You can often see that animations are slightly stuttering on any device & chrome version after you open a modal/popup. The performance often goes back to it's normal state after a while and sometimes don't which is a real problem, on lower-end device this performance hit is NOT negligible at all.

This bug was probably introduced when the popup/modal system was added which is quite old.

crisbeto:
* Removes the global scroll listener from the viewport ruler, which was being used to invalidate the scroll position cache. Caching the scroll position is pointless since it'll be invalidated very frequently. Since most Material components depend on the `ViewportRuler` one way or another, it meant that there was a good chance that the scroll listener would stay around forever.
* Fixes an issue that caused the scroll dispatcher not to remove its global scroll listener, if there are any scrollables on the page, even though nothing was subscribed to it.

Fixes #6882.

### DIFF
--- a/src/cdk/scrolling/scroll-dispatcher.spec.ts
+++ b/src/cdk/scrolling/scroll-dispatcher.spec.ts
@@ -143,6 +143,26 @@ describe('Scroll Dispatcher', () => {
           'Expected global listeners to have been removed after the subscription has stopped.');
     });
 
+    it('should remove global listeners on unsubscribe, despite any other live scrollables', () => {
+      const fixture = TestBed.createComponent(NestedScrollingComponent);
+      fixture.detectChanges();
+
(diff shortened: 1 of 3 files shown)


kara:
@crisbeto  Rebase?

crisbeto:
Rebased.

PR #7563, opened 2017-10-05
refactor(progress-spinner): re-use keyframe stylesheet element

crisbeto:
Re-uses a single stylesheet element for the progress spinner keyframes, instead of creating a new one every time. Also uses `insertRule` instead of modifying the DOM directly.


crisbeto:
I think we'd be testing the implementation of `Set` at that point.

josephperrott:
I suppose that's true

PR #7567, opened 2017-10-05
fix(form-field): fix underline at different zoom levels

### LINKED ISSUE #7494: (Bug) .mat-form-field-underline not rendering in Webkit browsers
First off -- thanks for the all the great code!

The line underneath the input in a form field (any type of form field) with the class `.mat-form-field-underline` is sometimes absent from Webkit browsers. It renders as expected at any zoom level and any size screen or device on Firefox, but with Chrome and Safari it's completely random. For example, I have a list in my application using the plain ol' input and datepicker elements. Some of them have underlines, some of them don't. It's different which ones are working and which aren't on any given render. 

I thought it may have something to do with the height being changed to `0.99px` in Chrome, but even setting the height to `1.1px !important` using `::ng-deep` doesn't fix it permanently. Something that may help is that when I use that `1.1px` height trick, some lines look extra thick, while others just look normal. I would think maybe those normal looking ones are the ones that would be absent without explicitly setting the height, but then again, sometimes they disappear altogether.

The underlines would show up all the time, no matter what.

Unpredictable rendering for underlines.

https://plnkr.co/edit/kmjxWTwC6sSrkSnDVoGQ?p=preview

Check out this Plunker on Chrome and zoom in and out, resize your window, refresh the page a bunch of times, eventually the underline will disappear. For me, 80% is a pretty reliable way to make it disappear, but I'm sure it's different everywhere (different screens, devices, etc.). I wish I could provide more info, but it is truly random. 

Using the input. : /

Webkit browsers.

Again, great library. Thanks!

mmalerba:
fixes #7494

### DIFF
--- a/src/lib/form-field/form-field.scss
+++ b/src/lib/form-field/form-field.scss
@@ -145,6 +145,8 @@ $mat-form-field-underline-height: 1px !default;
   position: absolute;
   height: $mat-form-field-underline-height;
   width: 100%;
+  // Prevents underline from disappearing at lower zoom levels.
+  transform: perspective(1px);
 
   &.mat-disabled {
     background-position: 0;


src/lib/form-field/form-field.scss
```diff
@@ -145,6 +145,8 @@ $mat-form-field-underline-height: 1px !default;
   position: absolute;
   height: $mat-form-field-underline-height;
   width: 100%;
+  // Prevents underline from disappearing at lower zoom levels.
+  transform: perspective(1px);
```
crisbeto: I'm not sure whether this fixes the issue. The only way I managed to break it on my machine was to zoom out to 33% and these changes don't seem to help in that case.

mmalerba: Its a little weird, they don't all seem to break, I think it has to do with their vertical position or something. I saw some of them broken at 90% on the input demo page, this fixes it for >= 50% which I think is reasonable

crisbeto: I think that it all depends on your OS/Screen size/Browser combo and whether it hits some weird value that doesn't round off correctly. We could get this in and see if people still see the issue.

mmalerba: Yeah, I suspect this will fix it, it does for me

jecastrops:
This fix is ​​great but at 33% and 25% zoom the line disappears again

PR #7608, opened 2017-10-06
 feat(sort): show sort indicator on hover/focus/longpress

### LINKED ISSUE #6858: [Sort] Sort icon should appear on hover
The spec says that the sort icons should appear when the header is hovered to indicate that the header is sortable.

andrewseguin:
Brings the sort header closer to spec but unfortunately takes away the cool animation that was added in #7180. Not sure how to reconcile the animation with this indicator hint. Perhaps only do the fade-away animation when sort is removed but not added?

Closes #6858, #8612

### DIFF
--- a/src/lib/sort/sort-header.ts
+++ b/src/lib/sort/sort-header.ts
@@ -19,6 +19,7 @@ import {CdkColumnDef} from '@angular/cdk/table';
 import {Subscription} from 'rxjs/Subscription';
 import {merge} from 'rxjs/observable/merge';
 import {MatSort, MatSortable} from './sort';
+import {SortDirection} from './sort-direction';
 import {MatSortHeaderIntl} from './sort-header-intl';
 import {getSortHeaderNotContainedWithinSortError} from './sort-errors';
 import {CanDisable, mixinDisabled} from '@angular/material/core';
@@ -29,6 +30,25 @@ import {matSortAnimations} from './sort-animations';
 export class MatSortHeaderBase {}
 export const _MatSortHeaderMixinBase = mixinDisabled(MatSortHeaderBase);
 
+/**
+ * Valid positions for the arrow to be in for its opacity and translation. If the state is a
+ * sort direction, the position of the arrow will be above/below and opacity 0. If the state is
+ * hint, the arrow will be in the center with a slight opacity. Active state means the arrow will
+ * be fully opaque in the center.
+ *
+ * @docs-private
+ */
+export type ArrowViewState = SortDirection | 'hint' | 'active';
+
+/**
+ * States describing the arrow's animated position (animating fromState -> toState).
+ * If the fromState is not defined, there will be no animated transition to the toState.
+ * @docs-private
+ */
+export interface ArrowViewStateTransition {
+  fromState?: ArrowViewState;
+  toState: ArrowViewState;
+}
 
 /**
  * Applies sorting behavior (click to change sort) and styles to an element, including an
@@ -47,7 +67,9 @@ export const _MatSortHeaderMixinBase = mixinDisabled(MatSortHeaderBase);
   styleUrls: ['sort-header.css'],
   host: {
     '(click)': '_handleClick()',
-    '[class.mat-sort-header-sorted]': '_isSorted()',
+    '(mouseenter)': '_setIndicatorHintVisible(true)',
+    '(longpress)': '_setIndicatorHintVisible(true)',
+    '(mouseleave)': '_setIndicatorHintVisible(false)',
     '[class.mat-sort-header-disabled]': '_isDisabled()',
   },
   encapsulation: ViewEncapsulation.None,
@@ -58,11 +80,34 @@ export const _MatSortHeaderMixinBase = mixinDisabled(MatSortHeaderBase);
     matSortAnimations.indicator,
     matSortAnimations.leftPointer,
     matSortAnimations.rightPointer,
-    matSortAnimations.indicatorToggle
+    matSortAnimations.arrowOpacity,
+    matSortAnimations.arrowPosition,
+    matSortAnimations.allowChildren,
   ]
 })
 export class MatSortHeader extends _MatSortHeaderMixinBase implements MatSortable, CanDisable {
   private _rerenderSubscription: Subscription;
+
+  /**
+   * Flag set to true when the indicator should be displayed while the sort is not active. Used to
+   * provide an affordance that the header is sortable by showing on focus and hover.
+   */
+  _showIndicatorHint: boolean = false;
+
+  /**
+   * The view transition state of the arrow (translation/ opacity) - indicates its `from` and `to`
+   * position through the animation. If animations are currently disabled, the fromState is removed
+   * so that there is no animation displayed.
+   */
+  _viewState: ArrowViewStateTransition;
+
+  /** The direction the arrow should be facing according to the current state. */
+  _arrowDirection: SortDirection = '';
+
+  /**
+   * Whether the view state animation should show the transition between the `from` and `to` states.
+   */
+  _disableViewStateAnimation = false;
 
   /**
    * ID of this sort header. If used within the context of a CdkColumnDef, this will default to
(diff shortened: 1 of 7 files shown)


jelbourn:
I don't quite follow why showing the indicator means the animation can't exist

andrewseguin:
This is only in reference to the animation we added for entering and exiting the sort (not transitioning between `asc` and `desc`).

Since the arrow shows on hover/focus, I don't think it's correct to show the arrow "fly in" since it is already in place when activated. Maybe we can have the arrow still "fly out" since the hint doesn't appear after the sort is disabled, but it wouldn't be a symmetrical animation (why does it fly out but never flies in)

jelbourn:
I think I need to see it in action to understand

andrewseguin:
Tried again for a few hours trying to get the animation working, its...more complex than it seems. Let's bring it back in another PR

jelbourn:
@andrewseguin can you give a brief summary of why the animation is hard to keep?

andrewseguin:
Had trouble with the animation out - the indicator wants to animate to the next state since its not `active` anymore but we want to preserve what state it was in as it animates out. Then, if you sort it while its animating out, you want it to float in from the bottom with that same position. I tried getting it to remember its old state after sort is removed, but it just got too tricky to add in this PR. I suspect it's totally doable but making it a clean implementation will take some thought

jelbourn:
@crisbeto thoughts? I don't want to remove the animation "temporarily" and then never get around to bringing it back

crisbeto:
I'm still having a hard time understanding the issue. Is it that we want to avoid the "fly in" animation when the user hovers the header?

andrewseguin:
Discussed in person - issue is in its complexity and not clear what is the best animation for some cases.

Complexity: Need to watch several states - if active, arrow should be pointing in its current direction; if not active, arrow should be pointing in the start position _except_ if the sort has just been disable (should fly out so the sort direction should persist)

andrewseguin:
Ready again for review - after a looong time playing with the animations, found a way to bring back the animation. Essentially the sort will define the right transition it should show (e.g. go from above to the middle, asc-to-active)

src/lib/sort/sort-header.ts
```diff
@@ -15,25 +15,35 @@ import {
...
+import {SortDirection} from './sort-direction';
 import {MatSortHeaderIntl} from './sort-header-intl';
 import {getSortHeaderNotContainedWithinSortError} from './sort-errors';
 import {AnimationCurves, AnimationDurations} from '@angular/material/core';
 
 const SORT_ANIMATION_TRANSITION =
     AnimationDurations.ENTERING + ' ' + AnimationCurves.STANDARD_CURVE;
 
+/**
+ * Valid positions for the arrow to be in for its opacity and translation.
```
jelbourn: Expand comment to explain what `'peek'` and `'active'` mean?

src/lib/sort/sort-header.ts
```diff
@@ -50,52 +60,135 @@ const SORT_ANIMATION_TRANSITION =
...
 })
 export class MatSortHeader implements MatSortable {
   private _rerenderSubscription: Subscription;
 
+  /**
+   * Flag set to true when the indicator should be displayed while the sort is not active. Used to
+   * provide an affordance that the header is sortable by showing on focus and hover.
+   * @docs-private
+   */
+  set showIndicatorHint(showIndicatorHint: boolean) {
```
jelbourn: Instead of making this a setter, can you make it a function like `_setIndicatorHintVisible`?

src/lib/sort/sort-header.ts
```diff
@@ -50,52 +60,135 @@ const SORT_ANIMATION_TRANSITION =
...
+  get showIndicatorHint(): boolean { return this._showIndicatorHint; }
+  _showIndicatorHint: boolean = false;
+
+  /**
+   * The view transition state of the arrow (translation/ opacity) - indicates its `from` and `to`
+   * position through the animation. If animations are currently disabled, the fromState is removed
+   * so that there is no animation displayed.
+   * @docs-private
+   */
+  set viewState(arrowPosition: ArrowViewStateTransition) {
```
jelbourn: Also make this a function? Something like `_setAnimationTransitionState`?

src/lib/sort/sort-header.ts
```diff
@@ -140,9 +247,49 @@ export class MatSortHeader implements MatSortable {
...
+    const fromState = this.viewState.fromState;
+    return (fromState ? `${fromState}-to-` : '') + this.viewState.toState;
+  }
+
+  /**
+   * Updates the direction the arrow should be pointing. If it is not sorted, the arrow should be
+   * facing the start direction. Otherwise if it is sorted, the arrow should point in the currently
+   * active sorted direction.
+   */
+  _updateArrowDirection() {
```
jelbourn: Why have an update function instead of a function like `_getArrowDirection`? The problem with the update function is that you can forget to call it, or call it when it's not necessary.

src/lib/sort/sort.spec.ts
```diff
@@ -88,6 +93,95 @@ describe('MatSort', () => {
     expect(sortables.has('column_c')).toBe(true);
   });
 
+  describe('checking correct arrow direction and view state for its various states', () => {
+    let expectedStates: Map<string, { viewState: string, arrowDirection: string }>;
```
jelbourn: Omit spaces in braces

src/lib/sort/sort.spec.ts
```diff
@@ -235,14 +389,35 @@ class SimpleMatSortApp {
...
+    this.dispatchMouseEvent(id, 'click');
+  }
+
+  dispatchMouseEvent(id: string, event: string) {
     const sortElement = this.elementRef.nativeElement.querySelector(`#${id}`);
-    dispatchMouseEvent(sortElement, 'click');
+    dispatchMouseEvent(sortElement, event);
+  }
+
+  expectViewAndDirectionStates(
```
jelbourn: Add JsDoc description? It's not super obvious what this method does

src/lib/sort/sort-header.ts
```diff
@@ -50,52 +60,135 @@ const SORT_ANIMATION_TRANSITION =
...
-      transition('asc <=> desc', animate(SORT_ANIMATION_TRANSITION))
+      state('active-asc, asc', style({transform: 'rotate(45deg)'})),
+      state('active-desc, desc', style({transform: 'rotate(-45deg)'})),
+      transition('active-asc <=> active-desc', animate(SORT_ANIMATION_TRANSITION))
+    ]),
+
+    // Arrow opacity
+    trigger('arrowOpacity', [
+      state('desc-to-active, asc-to-active, active',
+          style({opacity: 1})),
```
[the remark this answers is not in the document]
andrewseguin: Just consistency with the third state (`'hint-to-desc, active-to-desc, desc, hint-to-asc, active-to-asc, asc'`) which causes wraparound. No strong opinion either way - I unwrapped it

src/lib/sort/sort-header.ts
andrewseguin: I tried going that route but this ended up being much easier to control when the arrow changes directions.

Best example is when the sort goes from active to deactivated. By default the order for sort will be `none - asc - desc`. When the sort goes from `desc` to `none`, the arrow direction should remain the same until the animation completes (as it flies away). Only when the arrow is displayed again should it switch to showing itself as `asc` (hint is shown or activated). With just a "getter", we would probably have to rely on knowing if the arrow is animating away and have some additional logic about making it keep the "old direction" in the meantime.

jelbourn: Expand comment to explain this, then?

tinayuangao:
@andrewseguin This has error `_disableStateAnimation not exist`.

andrewseguin:
Rebased -- should be alright now

andrewseguin:
Woohoo, got it working!

jelbourn:
@andrewseguin get a presubmit going?

PR #7610, opened 2017-10-06
fix(paginator): fix select baseline; support mobile

andrewseguin:
Paginator should wrap gracefully on smaller widths. Uses reverse-wrap to bring the pagination controls up rather than squishing the content of the paginator.

Also, fixed the baseline of the select so it lines up right with the label

BEFORE
image

AFTER
image


src/lib/paginator/paginator.html
```diff
@@ -1,43 +1,47 @@
...
+            class="mat-paginator-navigation-next"
+            (click)="nextPage()"
+            [attr.aria-label]="_intl.nextPageLabel"
+            [matTooltip]="_intl.nextPageLabel"
+            [matTooltipPosition]="'above'"
+            [disabled]="!hasNextPage()">
+      <div class="mat-paginator-decrement"></div>
+    </button>
+  </div>
+</div>
```
mmalerba: nit: \n

PR #7619, opened 2017-10-07
fix(select): open method triggered twice on click

crisbeto:
Fixes the `open` method being triggered both by a click handler on the select trigger and the `onContainerClick` from the `MatFormField` integration.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -508,7 +508,7 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit, 
 
   /** Opens the overlay panel. */
   open(): void {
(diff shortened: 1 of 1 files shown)


mmalerba:
I think this is breaking tests because people are simulating clicks on the trigger, and the simulated clicks don't bubble. I also realized that people technically should be able to use the `mat-select` outside of the `mat-form-field` if they want. Maybe instead we can just check the target of the event in `onContainerClick` and ignore it if it was the trigger element?

crisbeto:
Reworked the PR to check internally for the `_panelOpen` flag @mmalerba.

josephperrott:
@crisbeto Can you rebase and we can try to get this one in?

crisbeto:
Rebased.

PR #7628, opened 2017-10-08
chore: remove outdated golden screenshots

DevVersion:
Currently if a screenshot test changed its name, the old golden reference image won't be deleted. This causes the screenshot diff task to compare against nothing (screenshot diffs now always fail).

With this change, outdated screenshot images will be deleted in master builds.

### DIFF
--- a/tools/gulp/tasks/screenshots.ts
+++ b/tools/gulp/tasks/screenshots.ts
@@ -188,15 +188,26 @@ function compareScreenshotFile(fileName: string, database: Database, prNumber: s
 }
 
 /** Uploads golden screenshots to the Google Cloud Storage bucket for the screenshots. */
-function uploadGoldenScreenshots() {
+async function uploadGoldenScreenshots() {
   const bucket = openScreenshotsBucket();
+  const localScreenshots = getLocalScreenshotFiles(SCREENSHOT_DIR);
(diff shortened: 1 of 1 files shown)


tools/gulp/tasks/screenshots.ts
```diff
@@ -188,15 +188,26 @@ function compareScreenshotFile(fileName: string, database: Database, prNumber: s
...
   const bucket = openScreenshotsBucket();
+  const localScreenshots = getLocalScreenshotFiles(SCREENSHOT_DIR);
+  const storageGoldenFiles = (await bucket.getFiles({prefix: FIREBASE_STORAGE_GOLDENS}))[0];
 
-  return Promise.all(getLocalScreenshotFiles(SCREENSHOT_DIR).map(fileName => {
+  // Only delete golden images that are outdated to avoid collisions with other screenshot diffs.
+  // Deleting every golden screenshot may also work, but will likely cause flakiness if multiple
+  // screenshot tasks run.
+  const deleteOutdatedGoldenFiles = Promise.all(storageGoldenFiles
+    .filter((file: any) => localScreenshots.indexOf(path.basename(file.name)) === -1)
```
jelbourn: Use `.includes`?

DevVersion: Done.

PR #7640, opened 2017-10-08
fix(select): errors not shown on submit

### LINKED ISSUE #7634: mat-select doesn't show mat-error on submit
Bug

Should show mat-error on submit (if invalid)

mat-select becomes red as expected, but mat-error doesn't appear until clicked/touched.  
doing `control.markAsTouched()`, `control.markAsDirty()` on submit, also doesn't work.

https://stackblitz.com/edit/angular-material2-issue-yxmtsq  
1. create mat-select + mat-error (I'm using required with ReactiveForms - `Validators.required`)  
2. add a submit button.  
3. Without touching the `select`, click on submit.  
select becomes red (invalid), but `mat-error` doesn't appear until you click/touch on the select.

Angular: ^4.4.4,  
"@angular/cdk": "^2.0.0-beta.12",  
"@angular/material": "^2.0.0-beta.12",  
"typescript": "^2.5.2",  
OS: macOS Sierra
browser: Chrome

Thanks :D

crisbeto:
Fixes `mat-select` error messages not being shown when the parent form is submitted.

Fixes #7634.

@mmalerba this is more or less what the input does already. I was considering moving it into the `MatFormFieldControl`, but having lifecycle hooks in there might look weird. What do you think?

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -2947,6 +2947,17 @@ describe('MatSelect', () => {
           .toBe('true', 'Expected aria-invalid to be set to true.');
     });
 
+    it('should render the error messages when the parent form is submitted', () => {
+      const debugEl = fixture.debugElement.nativeElement;
+
+      expect(debugEl.querySelectorAll('mat-error').length).toBe(0, 'Expected no error messages');
+
+      dispatchFakeEvent(fixture.debugElement.query(By.css('form')).nativeElement, 'submit');
+      fixture.detectChanges();
+
+      expect(debugEl.querySelectorAll('mat-error').length).toBe(1, 'Expected one error message');
+    });
+
     it('should be able to override the error matching behavior via an @Input', () => {
       fixture.destroy();
 
@@ -3652,6 +3663,8 @@ class InvalidSelectInForm {
           <mat-option value="steak-0">Steak</mat-option>
           <mat-option value="pizza-1">Pizza</mat-option>
         </mat-select>
+
+        <mat-error>This field is required</mat-error>
       </mat-form-field>
     </form>
   `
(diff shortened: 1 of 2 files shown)


mmalerba:
I was thinking we should just move all the `ErrorStateMatcher` stuff into `mat-form-field` anyways. It's essentially all the same logic duplicated between input and select, and anyone who implements a custom control is probably going to want similar behavior. (but also fine if you just want to do this for now to fix select)

crisbeto:
My issue with having it there is that some bindings on the form field control depend on it (e.g. the `aria-invalid` on the input). We'd have to introduce another subject like `stateChanges`, but on the form field, which could make things even more convoluted.

mmalerba:
Hmmm I see, what if we changed to, instead of an `@Input` to a `@Directive`: `[matErrorStateMatcher]` which would basically do this:

```ts
@Directive({
  ...
  providers: [{provide: ErrorStateMatcher, useExisting: MatErrorStateMatcherDirective}]
})
export class MatErrorStateMatcherDirective implements ErrorStateMatcher {
  @Input('matErrorStateMatcher') errorStateMatcher: ErrorStateMatcher;

  isErrorState(...) {
    return this.errorStateMatcher.isErrorState(...);
  }
}
```

This way form field and input could both just inject the token and not worry about dealing with the `@Input`. It also has a cool side-effect of being able to do this:

```html
<form class="my-special-error-group" [matErrorStateMatcher]="myMatcher">
  <!-- bunch of form fields that all get my custom matcher -->
</form>
```

crisbeto:
I don't think that would react to changes either. Angular marks it for checking if the `@Input` value changes. Since the input is the error matcher itself, it won't do anything when the error state changes.

mmalerba:
Ok then, let's do this for now, I'll keep thinking if there's any good way to move this common logic somewhere else...

HideDev:
@crisbeto when will this fix be released?

harshi1991:
Is this fix released?

PR #7641, opened 2017-10-08
feat(a11y): add autoCapture option to cdkTrapFocus

jelbourn:
This also renames "FocusTrapDirective" to "CdkTrapFocus" and re-exports it under the old name.

### DIFF
--- a/src/cdk/a11y/focus-trap.ts
+++ b/src/cdk/a11y/focus-trap.ts
@@ -348,24 +348,51 @@ export class FocusTrapDeprecatedDirective implements OnDestroy, AfterContentInit
   selector: '[cdkTrapFocus]',
   exportAs: 'cdkTrapFocus',
 })
-export class FocusTrapDirective implements OnDestroy, AfterContentInit {
+export class CdkTrapFocus implements OnDestroy, AfterContentInit {
   /** Underlying FocusTrap instance. */
   focusTrap: FocusTrap;
+
+  /** Previously focused element to restore focus to upon destroy when using autoCapture. */
+  private _previouslyFocusedElement: HTMLElement | null = null;
 
   /** Whether the focus trap is active. */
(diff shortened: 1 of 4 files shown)


src/cdk/a11y/focus-trap.ts
```diff
@@ -321,23 +321,53 @@ export class FocusTrapDeprecatedDirective implements OnDestroy, AfterContentInit
...
   get enabled(): boolean { return this.focusTrap.enabled; }
   set enabled(value: boolean) { this.focusTrap.enabled = coerceBooleanProperty(value); }
 
-  constructor(private _elementRef: ElementRef, private _focusTrapFactory: FocusTrapFactory) {
+  /**
+   * Whether the directive should automatially move focus into the trapped region upon
+   * initialization and return focus to the previous activeElement upon destruction.
+   */
+  @Input('cdkTrapFocusAutoCapture')
+  get autoCapture(): boolean { return this._autoCapture; }
```
crisbeto: Since this won't react to value changes, maybe it should be a plain `@Input` without the coercion or it should take the value via an `@Attribute`?

jelbourn: I made it an input since people may want to build a higher-level component on top of it and forward from their own API for the equivalent behavior

crisbeto: Makes sense, although I do feel like we're starting to add the coercion to everything.

jelbourn: I'm going to push for Angular to add `@BooleanInput` (and `@NumberInput`) to help with this.

willshowell:
Breaking changes message? Or is the plan to wait until the re-export is removed?

jelbourn:
It should be a deprecation notice in the next release

PR #7665, opened 2017-10-09
fix(focus-trap): not attaching correctly if element is not in the DOM on init

crisbeto:
Currently the focus trap attempts to attach itself once on init (e.g. when it is projected inside of an overlay), however in some cases it might not be in the DOM yet. These changes make it so it re-tries to attach itself until it does so successfully.

### DIFF
--- a/src/cdk/a11y/focus-trap/focus-trap.ts
+++ b/src/cdk/a11y/focus-trap/focus-trap.ts
@@ -17,6 +17,7 @@ import {
   Input,
   NgZone,
   OnDestroy,
+  DoCheck,
 } from '@angular/core';
 import {take} from 'rxjs/operators';
 import {InteractivityChecker} from '../interactivity-checker/interactivity-checker';
@@ -32,6 +33,7 @@ import {InteractivityChecker} from '../interactivity-checker/interactivity-check
 export class FocusTrap {
   private _startAnchor: HTMLElement | null;
   private _endAnchor: HTMLElement | null;
+  private _hasAttached = false;
 
   /** Whether the focus trap is active. */
   get enabled(): boolean { return this._enabled; }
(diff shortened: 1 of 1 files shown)


src/cdk/a11y/focus-trap.ts
```diff
@@ -337,7 +347,11 @@ export class FocusTrapDirective implements OnDestroy, AfterContentInit {
...
   }
 
-  ngAfterContentInit() {
-    this.focusTrap.attachAnchors();
+  ngDoCheck() {
+    // Since the element may not be attached to the DOM (or another element)
+    // immediately, keep trying until it is.
+    if (!this.focusTrap.hasAttached()) {
+      this.focusTrap.attachAnchors();
+    }
```
jelbourn: Add a unit test for this?

crisbeto: I had a hard time putting together a unit test that would fail as we expect it to. The only way I got the issue to happen at runtime was using a focus trap as the content of a select which has a complicated setup around it. I tried throwing a portal into the unit test, but it still worked.

jelbourn: I ran into the issue with a setup like
```html
<ng-template cdkConnectedOverlay>
  <div cdkTrapFocus>...</div>
</ng-template>
```

crisbeto: That doesn't seem to work either. During the test the focus trap still manages to find a parent node which allows it to attach (the ` <div id="root0">` in particular which AFAIK is the Karma root element), whereas in the demo app it doesn't, For reference, here's what the test component looks like:

```ts
@Component({
  template: `
    <button cdkOverlayOrigin #origin="cdkOverlayOrigin">I'm the origin</button>

    <ng-template cdkConnectedOverlay [open]="isOpen" [origin]="origin">
      <div cdkTrapFocus><button>I'm trapped!</button></div>
    </ng-template>
  `
})
class FocusTrapInOverlay {
  isOpen = false;
}
```

jelbourn: The whole setup where it happened consistently was:
```html
<ng-template
    cdkConnectedOverlay
    [cdkConnectedOverlayOrigin]="trigger"
    [cdkConnectedOverlayOpen]="isOpen">

  <div cdkTrapFocus class="color-picker"
        role="dialog" aria-label="Color picker">
      <div role="listbox" tabindex="0" aria-label="Select a color"
        (keydown)="keydownHandler($event)">
        <div *ngFor="let color of colors; index as i"
            role="option"
            [style.background]="color.hex"
            [attr.aria-label]="color.name"
            [attr.aria-selected]="color.hex == selectedColor"
            (click)="selectedColor = color.hex; isOpen = false"></div>
      </div>
    </div>
</ng-template>
```
(ran into it while working on my angular mix demo)

josephperrott:
@jelbourn How are we looking on this one?

torabian:
You can use amazing time picker (amazing-time-picker) for this purpose, and it is fully compatible with Angular 2-7 and Material design:

https://github.com/owsolutions/amazing-time-picker

Live demo:
https://owsolutions.github.io/amazing-time-picker

Thanks.

jelbourn:
@torabian Please do not spam unrelated PRs and issues promoting a project. Further spam will lead to your account being blocked.

PR #7668, opened 2017-10-09
fix(drawer): not restoring focus on close

crisbeto:
Currently the drawer restores focus on close after its closing animation is done and it contains the currently-focused element. This will always evaluate to false, because the sidenav gets a `visibility: hidden` which will blur any focused elements that it may have. With these changes the drawer will restore focus when the animation starts and it's still visible.

**Note:** we actually had a unit test for this, however it didn't catch the failure, because animations are disabled in tests.


crisbeto:
Just saw that one of the tests is failing for IE. Investigating.

PR #7689, opened 2017-10-10
refactor: remove compatibility mode

crisbeto:
Since the switch to the `mat-` prefix, the compatibility mode is no longer necessary. These changes remove anything that was left.

BREAKING CHANGE:
The `MATERIAL_COMPATIBILITY_MODE`, `CompatibilityModule`, `NoConflictStyleCompatibilityMode`, `MatPrefixRejector`, `MdPrefixRejector` symbols have been removed.


jelbourn:
Caretaker: the sync config might need to be tweaked

PR #7690, opened 2017-10-10
chore(docs): general docs and types cleanup

amcdnl:
This PR aims to:

- Create consistency with typings
- Add documentation to missing properties/methods
- Exclude private methods that can't be actually private from docs
- Fix bad descriptions on methods

### DIFF
--- a/src/cdk/table/table.ts
+++ b/src/cdk/table/table.ts
@@ -140,7 +140,7 @@ export class CdkTable<T> implements CollectionViewer {
    * Stream containing the latest information on what rows are being displayed on screen.
    * Can be used by the data source to as a heuristic of what data should be provided.
    */
-  viewChange =
+  viewChange: BehaviorSubject<{start: number, end: number}> =
       new BehaviorSubject<{start: number, end: number}>({start: 0, end: Number.MAX_VALUE});
 
   // Placeholders within the table's template where the header and data rows will be inserted.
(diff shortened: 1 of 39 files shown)


src/cdk/table/table.ts
```diff
@@ -169,18 +169,18 @@ export class CdkTable<T> implements CollectionViewer {
...
 
-  ngAfterContentInit() {
+  ngAfterContentInit(): void {
     this._cacheColumnDefsByName();
     this._columnDefs.changes.subscribe(() => this._cacheColumnDefsByName());
     this._renderHeaderRow();
   }
 
-  ngAfterContentChecked() {
+  ngAfterContentChecked(): void {
```
DevVersion: Lifecycle hooks are not necessary. They are being excluded by our dgeni processors.

src/cdk/table/table.ts
```diff
@@ -218,7 +218,7 @@ export class CdkTable<T> implements CollectionViewer {
    * Check if the header or rows have changed what columns they want to display. If there is a diff,
    * then re-render that section.
    */
-  private _renderUpdatedColumns() {
+  private _renderUpdatedColumns(): void {
```
DevVersion: Private methods in general don't need to have an explicit type.

src/cdk/table/table.ts
amcdnl: I'm trying to create consistency on typing. If everything else has it, we might as well make these have it too.

src/cdk/table/table.ts
amcdnl: I'm trying to create consistency on typing.

crisbeto: The convention at the moment is that if it goes into the docs, it should be typed (so if it's public, doesn't start with an underscore and it's not an Angular lifecycle hook). IMO having to add the types everywhere might cause some churn.

amcdnl:
Given the feedback, I'm going to close this and only go back and type things that will show up in the docs.

PR #7696, opened 2017-10-10
fix(select): wrong cursor on disabled select

### LINKED ISSUE #7695: Disabled and selected MatSelect placeholder shouldn't have pointer cursor
bug

Placeholder has default cursor when disabled

Placeholder has pointer cursor as if `MatSelect` is enabled

https://stackblitz.com/edit/angular-material2-issue-k4ica7?file=app/app.component.html

Consistent UI

`2.0.0-beta.12`

crisbeto:
* Fixes some areas of a disabled select having a `cursor: pointer`.
* Replaces the `mat-disabled` class on the form field underline with a `mat-form-field-disabled` class on the form field itself to allow for easier styling.

Fixes #7695.

### DIFF
--- a/src/lib/form-field/_form-field-theme.scss
+++ b/src/lib/form-field/_form-field-theme.scss
@@ -55,10 +55,10 @@
 
   .mat-form-field-underline {
     background-color: $underline-color;
+  }
 
-    &.mat-disabled {
-      @include mat-control-disabled-underline($underline-color);
-    }
+  .mat-form-field-disabled .mat-form-field-underline {
+    @include mat-control-disabled-underline($underline-color);
   }
 
   .mat-form-field-ripple {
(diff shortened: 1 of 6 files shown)


src/lib/form-field/_form-field-theme.scss
```diff
@@ -56,7 +56,7 @@
   .mat-form-field-underline {
     background-color: $underline-color;
 
-    &.mat-disabled {
+    .mat-form-field-disabled & {
```
mmalerba: I think we have a rule that `&` can only go at the beginning in theme files, so that people can do stuff like this

```scss
.dark-theme {
  @include angular-material-theme($dark-theme);
}
```

crisbeto:
Forgot that that lint rule was a thing @mmalerba, can you take another look?

PR #7707, opened 2017-10-11
build: run tests against ios10

DevVersion:


DevVersion:
@jelbourn FYI: There are a few flaky tests on the select right now. Also one tooltip test is failing on Mobile Safari (even with iOS 9 -> unrelated)

PR #7713, opened 2017-10-11
chore(docs): adds types column and null checks for description

amcdnl:
This PR:
- Adds type column for properties
- Adds null check to descriptions

### DIFF
--- a/tools/dgeni/templates/property.template.html
+++ b/tools/dgeni/templates/property.template.html
@@ -26,8 +26,10 @@
       {$ property.name $}
     </p>
   </td>
-  <td class="docs-api-property-type">
-    <code>{$ property.type $}</code>
+  <td class="docs-api-property-description">
+    <div class="docs-api-property-type">
+      Type: <code>{$ property.type $}</code>
+    </div>
+    {$ property.description | marked | safe $}
   </td>
-  <td class="docs-api-property-description">{$ property.description | marked | safe $}</td>
 </tr>
--- a/tools/dgeni/templates/property.template.html
+++ b/tools/dgeni/templates/property.template.html
@@ -25,7 +25,9 @@
     <p class="docs-api-property-name">
       {$ property.name $}
     </p>
-    <code class="docs-api-property-type">{$ property.type $}</code>
+  </td>
+  <td class="docs-api-property-type">
+    <code>{$ property.type $}</code>
   </td>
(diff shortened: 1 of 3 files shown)


jelbourn:
Could you include a screenshot of how this looks? How does it fit on mobile?

amcdnl:
Here is a before/after

img

jelbourn:
I'm not a huge fan of how much horizontal space it takes up now; any ideas around how we can keep the name and the type in the same vertical space next to the description?

willshowell:
From https://github.com/angular/material.angular.io/issues/287#issuecomment-335505847, would you consider moving the type to the description cell and making it distinct via font weight and padding? Just instead of "**Return Value:**", have it sat "**Type:**"

screen shot 2017-10-10 at 11 11 19 am

jelbourn:
That could work- I do want to try emphasizing the type by using a monospace font with a different color rather than writing out `type` to see how that comes across

amcdnl:
Here is the updated look:

img

tools/dgeni/templates/property.template.html
```diff
@@ -25,7 +25,11 @@
     <p class="docs-api-property-name">
       {$ property.name $}
     </p>
-    <code class="docs-api-property-type">{$ property.type $}</code>
   </td>
-  <td class="docs-api-property-description">{$ property.description | marked | safe $}</td>
+  <td class="docs-api-property-description">
+    <div class="docs-api-property-type">
+      Type: <code>{$ property.type $}</code>
```
jelbourn: I'm not a huge fan of having "Type: " explicitly. Repeated text like this creates a lot of visual noise

amcdnl: Gotcha. Should I just remove type and leave it like is otherwise.

jelbourn: Can you post a SS of what that looks like?

amcdnl:
I'm not sure I like this:

img

jelbourn:
What if we add it line with the name, TypeScript syntax style?
```
@Input()
disabled: boolean
```

amcdnl:
@jelbourn - That doesn't look bad ->

img

jelbourn:
I do like that more

jelbourn:
@amcdnl can you re-create this PR? It was merged before final changes could be made

amcdnl:
Done in #8125

PR #7746, opened 2017-10-12
fix(progress-spinner): default strokeWidth to 10% of the diameter

### LINKED ISSUE #7708: [spinner] default strokeWidth to 1/10th diameter
Proposal

#### Problem
Spinners (in my experience) are almost never 100px in diameter. Setting the `diameter` property is fine, but it's annoying to have to set the `strokeWidth` each time too in order to get a balanced looking spinner.

#### Currently
These are all over my app,

```html
<mat-spinner [diameter]="32" [strokeWidth]="3.2"></mat-spinner>

<mat-spinner [diameter]="64" [strokeWidth]="6.4"></mat-spinner>
```

https://stackblitz.com/edit/material2-beta12-nj2ob9?file=app/app.component.html

#### Proposed

Unless `strokeWidth` is manually set, use `diameter` / 10.

cc @crisbeto

crisbeto:
Currently the default stroke of a spinner is always 10px which doesn't look great on anything that's smaller than the default. These changes switch it to be 10% of the circle's diameter.

Fixes #7708.

### DIFF
--- a/src/lib/progress-spinner/progress-spinner.spec.ts
+++ b/src/lib/progress-spinner/progress-spinner.spec.ts
@@ -1,7 +1,7 @@
 import {TestBed, async} from '@angular/core/testing';
 import {Component} from '@angular/core';
 import {By} from '@angular/platform-browser';
-import {MatProgressSpinnerModule} from './index';
+import {MatProgressSpinnerModule, MatProgressSpinner} from './index';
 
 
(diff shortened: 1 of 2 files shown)


kara:
Caretaker note: Please use extra care when presubmitting, due to recent spinner issues. 

CSS: @jelbourn

PR #7747, opened 2017-10-12
fix(drawer): re-add openedStart and closedStart events

### LINKED ISSUE #6924: A feature, (open-start) and (close-start), for sidenav dissapeared in beta .10... Please return it
I use this api feature for a very specific case... and there was no deprecation, feature parity, or other info explaining the dissapearance of said feature.  

The only thing I read was mdsidenav was changing to mddrawer

I use the said features to do specific actions when the side panel / drawer is coming onto the screen.  

For example, when the drawer first fires I need to fire a function... not when the drawer is finished sliding.  The visual affect is greatly diminished without this capability

now, I only have access to open and close via when the side panel is finished coming into it's state.

You'd have to use the api, but alas the optiions are gone.

This was awesome granular control that allowed me to have something disappear off the screen when the panel / drawer began to open.  

material verison 2 beta .10

nope

crisbeto:
* Re-adds the `openedStart` and `closedStart` events due to popular demand.
* Fixes wrong signature for the returned promise from `toggle`. The signature states that the resolved value is supposed to be a `MatDrawerToggleResult` while in practice it was `undefined`.
* Splits out a long unit test into two smaller ones.

Fixes #6924.

### DIFF
--- a/src/demo-app/drawer/drawer-demo.html
+++ b/src/demo-app/drawer/drawer-demo.html
@@ -1,7 +1,7 @@
 <h2>Basic Use Case</h2>
 
 <mat-drawer-container class="demo-drawer-container">
-  <mat-drawer #start (open)="myinput.focus()" mode="side">
+  <mat-drawer #start (opened)="myinput.focus()" mode="side">
     Start Side Drawer
     <br>
     <button mat-button (click)="start.close()">Close</button>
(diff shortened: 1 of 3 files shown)


mmalerba:
@jelbourn @amcdnl any naming preference for these new outputs?

amcdnl:
@mmalerba - Names LGTM.

PR #7748, opened 2017-10-12
fix(sidenav): remove min-width

### LINKED ISSUE #682: MdSidenav should not have min-width

Bug

Developer should be able to set the width of MdSidenav element 

A Developer can set the width in several ways.
Using CSS:

``` css
md-sidenav {
  width: 200px;
}
```

Using binding:

``` html
<md-sidenav [style.width]="...">
```

or `ngStyle` 

The above will work as long as the width is greater then 5% of the container (`MdSidenavLayout`)
There is a CSS rule `md-sidenav {min-width: 5%`} 
https://github.com/angular/material2/blob/master/src/components/sidenav/sidenav.scss#L116

Since `MdSidenav` is shadowed via emulation it's hard to change that behaviour, the only option is to set `min-width: unset` inline, which requires a bind or using the renderer.

An app can have a small md-sidenav for icons, very common use case.
`min-width` should be a decision made by the developer.

all

In general, the "themed" approach for material design in angular 1 worked great but it raises some problems in angular 2.
In ng1 a developer had a theme which helped keeping look and feel consistent. In some special cases it might be needed to have a custom color not in the theme palette, having a special CSS selector for that solved the issue.
In Angular 2 however, it's get tricky since shadow dom get's applied. For example, having a custom background color for `md-sidenav` is not straight forward since background color is fixed to the theme's bg color and it's shadowed. The solution is to inline styles or have a container inside the sidenav with 100% h/w and the color needed.

mmalerba:
fix #682

### DIFF
--- a/src/lib/sidenav/drawer.scss
+++ b/src/lib/sidenav/drawer.scss
@@ -106,7 +106,6 @@ $mat-drawer-over-drawer-z-index: 4;
   top: 0;
   bottom: 0;
   z-index: 3;
-  min-width: 5vw;
   outline: 0;
   box-sizing: border-box;
   overflow-y: auto; // TODO(kara): revisit scrolling behavior for drawers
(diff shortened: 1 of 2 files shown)


jelbourn:
@mmalerba looks like this has some CI failures

mmalerba:
fixed the tests

PR #7751, opened 2017-10-12
fix(table): throw error when missing row defs

andrewseguin:
When users forgot to include a `<cdk-header-row>` and `<cdk-row>`, the error messaging is vague and not helpful. This provides a check to help them know where to start with debugging

### DIFF
--- a/src/cdk/table/table.ts
+++ b/src/cdk/table/table.ts
@@ -183,7 +183,7 @@ export class CdkTable<T> implements CollectionViewer {
   }
 
   ngAfterContentInit() {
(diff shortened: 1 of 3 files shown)


src/cdk/table/table.ts
```diff
@@ -175,6 +177,10 @@ export class CdkTable<T> implements CollectionViewer {
   }
 
   ngAfterContentInit() {
+    if (!this._headerDef && this._rowDefs.length == 0) {
```
jelbourn: `!this._rowDefs.length`?

PR #7756, opened 2017-10-13
docs(icon): fix unescaped open-close (<>) tags in public comments

julianobrasil:


julianobrasil:
@jelbourn, backticks doesn't seem to escape html tags at all:

image

The comment in the code is:

```ts
  /**
   * Returns an Observable that produces the icon (as an `<svg>` DOM element) with the given name
   * and namespace. The icon must have been previously registered with addIcon or addIconSet;
   * if not, the Observable will throw an error.
   *
   * @param name Name of the icon to be retrieved.
   * @param namespace Namespace in which to look for the icon.
   */
```

Instead of that blank space in the image above, I expected that the backticks around `svg` would fix the missing parts of the comments and give me this:

```
    Returns an Observable that produces the icon (as an <svg> DOM element) with the given name
    and namespace. The icon must have been previously registered with addIcon or addIconSet;
    if not, the Observable will throw an error.
```

willshowell:
@julianobrasil this should work via 

https://github.com/angular/material2/blob/b39202ff3fb6c509460a42f04aaa291c44490273/tools/dgeni/templates/method.template.html#L14-L16

Did you check that locally? I don't think the docs site has updated with new assets since this was merged.

Issue #7757, opened 2017-10-13
DOCS - Card Example not working


mswilson4040:
#### What are the steps to reproduce?
Providing a StackBlitz/Plunker (or similar) is the *best* way to get the team to see your issue. <br/>
Plunker starter (using on `@master`): https://goo.gl/DlHd6U<br/>
StackBlitz starter (using latest `npm` release): https://goo.gl/wwnhMV<br/>

1.) Navigate to https://material.angular.io/components/card/examples
2.) Observe Card Example not rendering correctly

donroyco:
That's because the images are loaded over HTTP, not HTTPS.

PR #7762, opened 2017-10-13
fix(card): change image path to https

donroyco:
This will solve #7757


donroyco:
I signed the CLA.

donroyco:
I signed it!

PR #7776, opened 2017-10-13
feat(autocomplete): dynamic [displayWith] and fix(autocomplete): Write values synchronously

ppham27:
When the displayWith function changes, the autocomplete trigger needs to rewrite the trigger value. Since we're listening to changes to displayWith, we no longer need to wait for stability before writing values, so we can write synchronously.

It was necessary to clean up some of the observables and make a single shared click stream. Otherwise, different subscribers would see different results since unpublished observables are evaluated once per subscription in DFS order.

Demo: https://angular-material-devapp-97d86.firebaseapp.com/autocomplete

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -18,7 +18,7 @@ import {
 } from '@angular/cdk/overlay';
 import {TemplatePortal} from '@angular/cdk/portal';
 import {DOCUMENT} from '@angular/common';
-import {filter, take, switchMap, delay, tap, map} from 'rxjs/operators';
+import {delay, filter, map, share, switchMap, take, tap} from 'rxjs/operators';
 import {
   ChangeDetectorRef,
   Directive,
@@ -138,6 +138,9 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   /** Stream of keyboard events that can close the panel. */
   private readonly _closeKeyEventStream = new Subject<void>();
 
+  /**  The current model value. */
+  private _value: any;
+
   /** `View -> model callback called when value changes` */
   _onChange: (value: any) => void = () => {};
 
@@ -145,7 +148,25 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   _onTouched = () => {};
 
   /** The autocomplete panel to be attached to this trigger. */
-  @Input('matAutocomplete') autocomplete: MatAutocomplete;
+  @Input('matAutocomplete')
+  get autocomplete(): MatAutocomplete {
+    return this._autocomplete;
+  }
+  set autocomplete(value: MatAutocomplete) {
+    this._autocomplete = value;
+    this._displayWithSubscription.unsubscribe();
+    if (this.autocomplete) {
+      this._displayWithSubscription = this.autocomplete._displayWithChange
+        .subscribe(() => {
+          this.writeValue(this._value);
+        });
+    }
+  }
+  private _autocomplete: MatAutocomplete;
+  private _displayWithSubscription = Subscription.EMPTY;
+
+  /** Stream of clicks outside of the autocomplete panel. */
+  private readonly _outsideClickStream: Observable<any>;
 
   /**
    * Reference relative to which to position the autocomplete panel.
@@ -170,15 +191,18 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
               @Inject(MAT_AUTOCOMPLETE_SCROLL_STRATEGY) private _scrollStrategy,
               @Optional() private _dir: Directionality,
               @Optional() @Host() private _formField: MatFormField,
-              @Optional() @Inject(DOCUMENT) private _document: any,
+              @Optional() @Inject(DOCUMENT) _document: any,
               // @deletion-target 7.0.0 Make `_viewportRuler` required.
-              private _viewportRuler?: ViewportRuler) {}
+              private _viewportRuler?: ViewportRuler) {
+    this._outsideClickStream = this._getOutsideClickStream(_document);
+  }
 
   ngOnDestroy() {
     this._viewportSubscription.unsubscribe();
     this._componentDestroyed = true;
     this._destroyPanel();
     this._closeKeyEventStream.complete();
+    this._displayWithSubscription.unsubscribe();
   }
 
   /** Whether or not the autocomplete panel is open. */
@@ -265,31 +289,10 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
     return null;
   }
 
-  /** Stream of clicks outside of the autocomplete panel. */
-  private get _outsideClickStream(): Observable<any> {
-    if (!this._document) {
-      return observableOf(null);
-    }
-
-    return merge(
-      fromEvent(this._document, 'click'),
-      fromEvent(this._document, 'touchend')
-    )
-    .pipe(filter((event: MouseEvent | TouchEvent) => {
-      const clickTarget = event.target as HTMLElement;
-      const formField = this._formField ?
-          this._formField._elementRef.nativeElement : null;
-
-      return this._overlayAttached &&
-              clickTarget !== this._element.nativeElement &&
-              (!formField || !formField.contains(clickTarget)) &&
-              (!!this._overlayRef && !this._overlayRef.overlayElement.contains(clickTarget));
-    }));
-  }
-
   // Implemented as part of ControlValueAccessor.
   writeValue(value: any): void {
-    Promise.resolve(null).then(() => this._setTriggerValue(value));
+    this._setTriggerValue(value);
+    this._changeDetectorRef.markForCheck();
   }
 
   // Implemented as part of ControlValueAccessor.
@@ -466,6 +469,7 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   }
 
   private _setTriggerValue(value: any): void {
+    this._value = value;
     const toDisplay = this.autocomplete && this.autocomplete.displayWith ?
       this.autocomplete.displayWith(value) :
       value;
@@ -547,6 +551,28 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
     if (this.panelOpen && wasOpen !== this.panelOpen) {
       this.autocomplete.opened.emit();
     }
+  }
+
+  /** Returns a stream of click events from outside of the autocomplete panel. */
+  private _getOutsideClickStream(document: Document): Observable<any> {
+    if (!document) {
+      return observableOf();
+    }
+
+    return merge(
+      fromEvent(document, 'click'),
+      fromEvent(document, 'touchend')
+    )
+    .pipe(filter((event: MouseEvent | TouchEvent) => {
+      const clickTarget = event.target as HTMLElement;
+      const formField = this._formField ?
+          this._formField._elementRef.nativeElement : null;
+
+      return this._overlayAttached &&
+              clickTarget !== this._element.nativeElement &&
+              (!formField || !formField.contains(clickTarget)) &&
+              (!!this._overlayRef && !this._overlayRef.overlayElement.contains(clickTarget));
+    }), share());
   }
 
   private _getOverlayConfig(): OverlayConfig {
--- a/src/lib/autocomplete/autocomplete.spec.ts
+++ b/src/lib/autocomplete/autocomplete.spec.ts
@@ -180,6 +180,15 @@ describe('MatAutocomplete', () => {
           .toBe(false, `Expected clicking outside the panel to set its state to closed.`);
       expect(overlayContainerElement.textContent)
           .toEqual('', `Expected clicking outside the panel to close the panel.`);
+
+      // Make sure that it works twice.
+      dispatchFakeEvent(input, 'focusin');
+      fixture.detectChanges();
+      expect(fixture.componentInstance.trigger.panelOpen).toBe(true);
+
+      zone.simulateZoneExit();
+      dispatchFakeEvent(document, 'click');
+      expect(fixture.componentInstance.trigger.panelOpen).toBe(false);
     }));
 
     it('should close the panel when the user taps away on a touch device', fakeAsync(() => {
@@ -727,6 +736,23 @@ describe('MatAutocomplete', () => {
       plainFixture.detectChanges();
 
       expect(input.disabled).toBe(true);
+    });
+
+    it('should update the display synchronously when set programmatically', () => {
+      fixture.componentInstance.stateCtrl.setValue(
+        fixture.componentInstance.states.find((state) => state.code === 'AL'));
+      fixture.detectChanges();
+      expect(input.value).toBe('Alabama');
+
+      fixture.componentInstance.stateCtrl.setValue(
+        fixture.componentInstance.states.find((state) => state.code === 'PA'));
+      fixture.detectChanges();
+      expect(input.value).toBe('Pennsylvania');
+
+      fixture.componentInstance.stateCtrl.setValue(
+        fixture.componentInstance.states.find((state) => state.code === 'CA'));
+      fixture.detectChanges();
+      expect(input.value).toBe('California');
     });
 
   });
@@ -1306,7 +1332,6 @@ describe('MatAutocomplete', () => {
     it('should set aria-owns based on the attached autocomplete', () => {
       fixture.componentInstance.trigger.openPanel();
       fixture.detectChanges();
-
       const panel = fixture.debugElement.query(By.css('.mat-autocomplete-panel')).nativeElement;
 
       expect(input.getAttribute('aria-owns'))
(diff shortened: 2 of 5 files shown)


src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -138,14 +149,33 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
               private _viewContainerRef: ViewContainerRef,
               private _zone: NgZone,
               private _changeDetectorRef: ChangeDetectorRef,
               @Inject(MAT_AUTOCOMPLETE_SCROLL_STRATEGY) private _scrollStrategy,
               @Optional() private _dir: Directionality,
               @Optional() @Host() private _formField: MatFormField,
-              @Optional() @Inject(DOCUMENT) private _document: any) {}
+              @Optional() @Inject(DOCUMENT) document: any) {
+    this._nextStable = first.call(_zone.onStable);
+    this._outsideClickStream = document ?
```
kara: Can you move this _outsideClickStream assignment out into a function?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -364,7 +381,6 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
    * stream every time the option list changes.
    */
   private _subscribeToClosingActions(): Subscription {
-    const firstStable = first.call(this._zone.onStable.asObservable());
```
kara: This observable was previously re-created every time the panel opened. Now it looks like it's only created when the trigger is created the first time, and will complete after the first event emits. Can you add a test that confirms this still works the second time the panel opens?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -210,35 +240,22 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
    */
   writeValue(value: any): void {
-    Promise.resolve(null).then(() => this._setTriggerValue(value));
+    // The stability check exists to make sure bindings have been initialized
+    // and avoid change detection errors.
+    if (this._zone.isStable) {
+      this._setTriggerValue(value);
+    } else {
+      this._nextStable.subscribe(() => {
+        this._setTriggerValue(value);
```
kara: Can you add a test that this still works the second time the value is set programmatically?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -138,14 +149,33 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
   constructor(private _element: ElementRef, private _overlay: Overlay,
               private _viewContainerRef: ViewContainerRef,
               private _zone: NgZone,
               private _changeDetectorRef: ChangeDetectorRef,
               @Inject(MAT_AUTOCOMPLETE_SCROLL_STRATEGY) private _scrollStrategy,
               @Optional() private _dir: Directionality,
               @Optional() @Host() private _formField: MatFormField,
-              @Optional() @Inject(DOCUMENT) private _document: any) {}
+              @Optional() @Inject(DOCUMENT) document: any) {
+    this._nextStable = first.call(_zone.onStable);
```
crisbeto: Since you're wrapping it in a `first`, it means that this will only ever fire once. There a couple of places below that assume that it'll fire multiple times.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -373,7 +389,7 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
       .result();
 
     // When the zone is stable initially, and when the option list changes...
-    return RxChain.from(merge(firstStable, optionChanges))
+    return RxChain.from(merge(this._nextStable, optionChanges))
```
crisbeto: `_nextStable` here won't do anything.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: It will work. Observables are not like promises and can complete on each new subscription. There's no need to recreate them every time. But if you feel it's necessary, I'll write that test.

```
// Go to http://reactivex.io/rxjs/ and open up the JavaScript console to try this code snippet out.
const subject = new Rx.Subject();
// Even though, this observable has a first operator it will emit and complete multiple times.
const obs = subject.first();

obs.subscribe((x) => {
  console.log(x);
}, undefined, () => {
  console.log('complete for the first time');
});
subject.next(1);

console.log('subscribing again');
obs.subscribe((x) => {
  console.log(x);
}, undefined, () => {
  console.log('complete for the second time');
});
subject.next(2);
```

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Not true. It's one of more confusing aspects of observables. They are more like functions than promises. On each new subscription, it will wait for an onStable emission and complete.

See https://github.com/angular/material2/pull/7776#discussion_r144660550

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: See previous comment.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: It works in the demo, but I'll work on adding this test.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: I think it makes more sense for the assignment to be here. Mainly, we can avoid having document be a class property and make this._outsideClickStream a readonly property. This has the added benefit of making sure we only emit once per a click event. Previously, there were two emissions. And it would actually emit something different on the second emission for the same click because the filter body was called twice.

ppham27:
Hi @crisbeto, thanks for taking a look at this quickly. Our app is hitting an exception due to the asynchronous nature (I'll try to reproduce it in a Plunkr). The placeholder problem is addressed with a detect changes call.

src/lib/autocomplete/autocomplete-trigger.ts
kara: For code style, we prefer not to have this much bloat in the constructor.  Could you not keep the assignment here and move the RHS to a normal function call? 

```ts
this._outsideClickStream = this._getOutsideClickStream(document);
```

src/lib/autocomplete/autocomplete-trigger.ts
kara: Fair enough. If you don't mind writing a quick test, I think it would be good to make sure it doesn't break going forward.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Okay, this accomplishes the same thing. And you're right, this does look much nicer. Change was made.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Okay open and close the panel twice in 'should emit panel close event when clicking away'.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Okay. I set the value multiple times in 'should update the display synchronously when set programmatically'

ppham27:
I realized that original promise existed to wait for the displayWith function to be bound. It makes more sense to just update whenever the displayWith function changes. At that point, it was trivial to just make the displayWith feature dynamic, which I think is actually a pretty useful feature. For example, for a people autocomplete, you might want to toggle between names and email addresses.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -129,27 +137,56 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
 
   /** View -> model callback called when autocomplete has been touched */
   _onTouched = () => {};
 
   /* The autocomplete panel to be attached to this trigger. */
-  @Input('matAutocomplete') autocomplete: MatAutocomplete;
+  @Input('matAutocomplete')
+  set _autocomplete(value: MatAutocomplete) {
+    this.autocomplete = value;
+    if (this._displayWithSubscription) {
```
crisbeto: Rather than null checking here, you can initialize the `_displayWithSubscription` to `Subscription.EMPTY`.

src/lib/autocomplete/autocomplete.ts
```diff
@@ -77,7 +77,18 @@ export class MatAutocomplete implements AfterContentInit {
...
+    return this._displayWith;
+  }
+  set displayWith(value: any) {
+    this._displayWith = typeof value === 'function' ? value : null;
+    this.displayWithChange.emit();
+  }
+  private _displayWith: ((value: any) => string) | null = null;
+
+  /** @docs-private */
+  readonly displayWithChange = new EventEmitter<void>();
```
crisbeto: Can yo rename this to `_displayWithChange`, add a doc string and switch it to a Subject? Also you'll be able to remove the `@docs-private` when it's prefixed with an underscore.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -129,27 +137,56 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
-              @Optional() @Inject(DOCUMENT) private _document: any) {}
+              @Optional() @Inject(DOCUMENT) document: Document) {
+    this._nextStable = first.call(zone.onStable);
+    this._outsideClickStream = this._getOutsideClickStream(document);
+  }
 
   ngOnDestroy() {
     this._destroyPanel();
     this._escapeEventStream.complete();
+    if (this._displayWithSubscription) {
```
crisbeto: If you initialize the `_displayWithSubscription` to `Subscription.EMPTY`, you can get rid of this null check.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -129,27 +137,56 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
+      this._displayWithSubscription.unsubscribe();
+    }
+    if (this.autocomplete instanceof MatAutocomplete) {
+      this._displayWithSubscription = this.autocomplete.displayWithChange
+        .subscribe(() => {
+          this.writeValue(this._value);
+        });
+    }
+  }
+  autocomplete: MatAutocomplete;
```
crisbeto: Having the property like this means that people could still swap out the `autocomplete` programmatically without going through the setter.

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -129,27 +137,56 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
 
   /* The autocomplete panel to be attached to this trigger. */
-  @Input('matAutocomplete') autocomplete: MatAutocomplete;
+  @Input('matAutocomplete')
+  set _autocomplete(value: MatAutocomplete) {
+    this.autocomplete = value;
+    if (this._displayWithSubscription) {
+      this._displayWithSubscription.unsubscribe();
+    }
+    if (this.autocomplete instanceof MatAutocomplete) {
```
crisbeto: I think that we can safely assume that the value will always be an instance of `MatAutocomplete`. This can be reduced to a null check.

src/lib/autocomplete/autocomplete-trigger.ts
crisbeto: Considering that this is the only place left that uses `_nextStable`, consider inlining it again?

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -493,6 +509,24 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
...
+      )).call(filter, (event: MouseEvent | TouchEvent) => {
+        const clickTarget = event.target as HTMLElement;
+        const formField = this._formField ?
+          this._formField._elementRef.nativeElement : null;
+
+        return this._panelOpen &&
+          clickTarget !== this._element.nativeElement &&
+          (!formField || !formField.contains(clickTarget)) &&
+          (!!this._overlayRef && !this._overlayRef.overlayElement.contains(clickTarget));
+      }).call(share).result() : observableOf();
```
crisbeto: Note that the `observableOf()` here isn't the same as the `observableOf(null)` that we had before. `observableOf()` returns an observable that'll never emit, whereas `observableOf(null)` will emit immediately. In this particular case it shouldn't matter since it's a fallback for Universal.

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -174,8 +174,17 @@ describe('MatAutocomplete', () => {
             .toBe(false, `Expected clicking outside the panel to set its state to closed.`);
         expect(overlayContainerElement.textContent)
             .toEqual('', `Expected clicking outside the panel to close the panel.`);
+
+        // Make sure that it works twice.
+        dispatchFakeEvent(input, 'focusin');
+        fixture.detectChanges();
+        expect(fixture.componentInstance.trigger.panelOpen).toBe(true);
+        return fixture.whenStable();
```
crisbeto: This is slightly cleaner, but isn't immediately obvious that the `click` will wait for the second `whenStable`. Instead of returning the promise here, can you move the `click` event up and nest it in a `whenStable` instead?

src/lib/autocomplete/autocomplete.spec.ts
```diff
@@ -1350,50 +1381,49 @@ describe('MatAutocomplete', () => {
       trigger = fixture.componentInstance.trigger;
       closingActionSpy = jasmine.createSpy('closing action listener');
       closingActionsSub = trigger.panelClosingActions.subscribe(closingActionSpy);
-    });
+
+      fixture.whenStable().then(() => {
+        expect(closingActionSpy).not.toHaveBeenCalled();
```
crisbeto: In general we try not to do expectations inside `beforeEach`. Can you move it into its own test?

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Made this change. Looks a lot better.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Done

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Okay, exposed autocomplete through a getter. I considered making autocomplete completely private and making a public autocompleteId property, but I was afraid that might break things.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Thanks for this suggestion. Code looks much better this way.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Unsure about this suggestion. By doing it my way, we can avoid making NgZone a class member. I also imagine that it's slightly more performant to not reconstruct Observables everywhere (this optimization can be done in several places throughout the code base).

The benefits that I listed are not significant, so if you feel strongly about it, I will revert.

src/lib/autocomplete/autocomplete-trigger.ts
ppham27: Yes aware of the difference. It was intentional to use observableOf() rather than observableOf(null), since it observableOf(null) would emit when there hasn't been a click.

src/lib/autocomplete/autocomplete.spec.ts
ppham27: Done.

src/lib/autocomplete/autocomplete.spec.ts
ppham27: Done.

src/lib/autocomplete/autocomplete.ts
ppham27: Done.

src/lib/autocomplete/autocomplete-trigger.ts
crisbeto: While this way the `NgZone` isn't a class member, `NgZone.onStable` still is. By having `NgZone` be the class member, it's easier to access if we need to use it anywhere else. Also the performance difference should be negligible. Also one more gotcha about the `onStable`, can you switch it to `onStable.asObservable()`? We've had some memory leaks in the past because we weren't converting it into an observable.

ppham27: > By having NgZone be the class member, it's easier to access if we need to use it anywhere else. 

I consider this a negative since I tend to believe in limiting the scope of class member to what the class needs. 

`onStable.asObservable()` should not be necessary with the `first` call since according to [What are operators?](http://reactivex.io/rxjs/manual/overview.html#what-are-operators-):

> Operators are methods on the Observable type, such as .map(...), .filter(...), .merge(...), etc. When called, they do not change the existing Observable instance. Instead, they return a new Observable, whose subscription logic is based on the first Observable.

Anyway, I do believe in having smaller diffs, though, so I've made your suggestions and reverted both changes. Now the `_subscribeToClosingActions` method is unchanged from master.

crisbeto: `onStable.asObservable()` definitely seems like it shouldn't, but it ends up causing any references inside the callback to be retained in memory after the component is destroyed. I'm not completely sure why that is the case, but I think it has to do with the `onStable` being an `EventEmitter`, not an `Observable` and the `NgZone` being a singleton.

ppham27: That makes sense to me if you were subscribing directly to onStable, but it happened even when you applied an operator before subscribing?

Anyway, thanks for the heads up. It seems like a practice that I should adopt to be safe in any case.

crisbeto: Yeah it was a little hard to track down, but we were doing in a lot of places. See https://github.com/angular/material2/pull/6918

mmalerba:
caretaker note: large number of failures in google3 with the following error:

`"Uncaught ReferenceError: autocomplete_1 is not defined", source: .../src/lib/autocomplete/autocomplete-trigger.js (447)`

It probably has something to do with how closure is minifying the code, possibly related to changing `@Input() autocomplete` to a getter/setter

jelbourn:
Closing this since it's now obsolete. We're planning a larger revamp of the autocomplete that should address some of the issues here.

PR #7779, opened 2017-10-13
fix(overlay): backdrop blocking element scroll

### LINKED ISSUE #6927: Menu should support background scrolling
Bug? Reading the documentation, having a menu open, should not block background scrolling. 

The background can scroll.

The background cannot scroll.

1. Visit https://material.angular.io/components/menu/overview
* Open the menu by clicking the trigger: "Menu"

crisbeto:
Fixes a long-standing where the backdrop doesn't allow the user to scroll, if the main scrollable container is not the body.

Fixes #6927.


jelbourn:
I agree we should do this long term, but I a bit hesitant to do it before stable because I suspect many Google apps have some to expect the current behavior and would need to switch to `BlockScrollStrategy`. It will take us some time to communicate the change and have make sure people are getting the behavior they want.  I'm thinking this should go on the list for susequent major release.

crisbeto:
@jelbourn just remembered about this one. Should we schedule it for 6.0?

jelbourn:
I don't think we have time before 6.0 to track down the stuff I mentioned above

crisbeto:
@jelbourn do we still want to do this?

jelbourn:
I'd say this is lower priority than the other stuff we're doing for v7. If we're able to wrap up the visual changes and other stuff, we can see about getting this in.

nadhir-falta:
Any work around for this please ?

hilnius:
any idea in which release can this PR make its way ?

kokojvanko:
Any idea when this will make it in? I really want to use the components from Angular material, but this is kind of a blocker...

rppig42:
Is there any update about this PR ?

Angular CDK is not only the base of Angular Material but also many other Angular UI libraries. The current behavior makes Angular Material websites different from other Material Design products like Google Keep.

gokhanoner:
@jelbourn, this is a long standing issue which prevents us & many to use this functionality. If you have no intension to implement it, please also let us know so we take it under consideration when using related material components.

gokhanoner:
I see. Thanks @jelbourn . I believe I can think of this as a wont-fixed issue then.

CnxYankoNikolov:
What is the prospect of making all those private functions protected, so we can at least extend the classes, apply this fix and provide them instead of the default ones? That way we can override the behavior, without changing the default one.

PR #7791, opened 2017-10-14
fix(progress-spinner): coerce string values

### LINKED ISSUE #7790: Progress Spinner diameter and strokeWidth bound vs unbound
Bug

`<mat-progress-spinner diameter="30" strokeWidth="10"></mat-progress-spinner>`
Should work the same as
`<mat-progress-spinner [diameter]="myValueOf30" [strokeWidth]="myValueOf10"></mat-progress-spinner>`

While the size of the spinner adjusts the same in both versions, without binding to the values, the underlying svg object increases exponentially in size with the strokeWidth and diameter.

Two spinners, one bound to both variables with sliders and it performs as expected, one with hard coded values.  Notice the space that the second takes up.  Try increasing the strokeWidth of the second, and watch as the svg gets enormous.
[https://plnkr.co/edit/NxWjkzGkcpfsgRi1R3VA?p=preview](https://plnkr.co/edit/NxWjkzGkcpfsgRi1R3VA?p=preview)

Perhaps its my ignorance, but it seems like we should be able to just set a value w/o binding to angular.
It is worth noting that binding to the properties and simply passing in a value works just fine.
`<mat-progress-spinner [diameter]="30" [strokeWidth]="10"></mat-progress-spinner>`

Angular 4.4.5
Material 2.0.0-beta.12

crisbeto:
Coerces any string values that are passed to `value`, `diameter` and `strokeWidth` to numbers.

Fixes #7790.

### DIFF
--- a/src/lib/progress-spinner/progress-spinner.ts
+++ b/src/lib/progress-spinner/progress-spinner.ts
@@ -26,6 +26,18 @@ import {coerceNumberProperty} from '@angular/cdk/coercion';
 /** Possible mode for a progress spinner. */
 export type ProgressSpinnerMode = 'determinate' | 'indeterminate';
 
+/**
+ * Base reference size of the spinner.
+ * @docs-private
+ */
+const BASE_SIZE = 100;
+
+/**
+ * Base reference stroke width of the spinner.
+ * @docs-private
+ */
+const BASE_STROKE_WIDTH = 10;
+
 // Boilerplate for applying mixins to MatProgressSpinner.
 /** @docs-private */
 export class MatProgressSpinnerBase {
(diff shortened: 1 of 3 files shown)


src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -86,6 +87,7 @@ export class MatProgressSpinner extends _MatProgressSpinnerMixinBase implements
   private _value: number;
   private readonly _baseSize = 100;
```
josephperrott: Should this be a module `const`?  Since it is only read during construction, and can't be modified?

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -121,7 +132,7 @@ export class MatProgressSpinner extends _MatProgressSpinnerMixinBase implements
   }
   set value(newValue: number) {
     if (newValue != null && this.mode === 'determinate') {
-      this._value = Math.max(0, Math.min(100, newValue));
+      this._value = Math.max(0, Math.min(100, coerceNumberProperty(newValue)));
```
josephperrott: We created a `clamp` method in `MdProgressBar` for this, should we do similar here?  Potentially moving clamp to a util?

If so it might be in a follow up PR, but thought I would throw it out there.

josephperrott:
@crisbeto Add `pr: merge ready` once everything is in order.

src/lib/progress-spinner/progress-spinner.ts
crisbeto: We used to have a clamp function because it was used in a couple of places, but now it's only used once so it should be fine having it inline.

andrewseguin:
Needs rebase.

crisbeto:
Rebased.

josephperrott:
Needs rebase

crisbeto:
Rebased @josephperrott.

PR #7792, opened 2017-10-14
feat(datepicker): dispatch events when datepicker is opened and closed

### LINKED ISSUE #7626: No event gets fired for mat-datepicker popup close (for the same day)
There should be an event that is fired upon closing the datepicker popup. At the moment dateChanged event is getting fired anytime a date changes but not when the same date is selected.

I expect at least there is an event that gets fired upon closing the datepicker's popup for the same day.

No events fired upon closing the datepicker's popup for the same day.

This are important in many scenarios. For example, I'm developing a date-time-picker and I show the time picker upon closing the datepicker's popup. The problem happens when I want to select a different time on the same day which is already selected. I click on the date-picker, select the same day and the datepicker's popup gets closed but there is no way I get notified because the dateChanged only fired when it's not the same day.

All

No

crisbeto:
Adds the `opened` and `closed` events that are dispatched when a datepicker is opened/closed.

Fixes #7626.

### DIFF
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -171,6 +171,12 @@ export class MatDatepicker<D> implements OnDestroy {
   /** Classes to be passed to the date picker panel. Supports the same syntax as `ngClass`. */
   @Input() panelClass: string | string[];
 
+  /** Emits when the datepicker has been opened. */
+  @Output('opened') openedStream: EventEmitter<void> = new EventEmitter<void>();
+
+  /** Emits when the datepicker has been closed. */
+  @Output('closed') closedStream: EventEmitter<void> = new EventEmitter<void>();
+
   /** Whether the calendar is open. */
   opened = false;
 
@@ -275,6 +281,7 @@ export class MatDatepicker<D> implements OnDestroy {
 
     this.touchUi ? this._openAsDialog() : this._openAsPopup();
     this.opened = true;
+    this.openedStream.emit();
   }
 
   /** Close the calendar. */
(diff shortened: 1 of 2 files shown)


src/lib/datepicker/datepicker.ts
```diff
@@ -167,6 +167,12 @@ export class MatDatepicker<D> implements OnDestroy {
    */
   @Output() selectedChanged = new EventEmitter<D>();
 
+  /** Emits when the datepicker has been opened. */
+  @Output('opened') openedStream: EventEmitter<void> = new EventEmitter<void>();
```
mmalerba: @amcdnl @jelbourn opinions on what we call these? I think select calls them `onOpen` and `onClose` (for both property and output name)

crisbeto: @mmalerba I based the naming on https://github.com/angular/material2/pull/7160 where `onOpen` and `onClose` were changed to `opened` and `closed`.

mmalerba: ah ok, can we just change the `closed` property to `closedStream` then so that it matches `openedStream`?

crisbeto: Done.

andreevWork:
Hello this PR solve problem in my project. Thank you, look forward merge it.

Jonafran:
Is there some time, this merge is going to be released?

josephperrott:
This was included in the [`5.0.0-rc1` release](https://github.com/angular/material2/blob/master/CHANGELOG.md#500-rc1-2017-11-20)

PR #7796, opened 2017-10-14
build: add karma task for easier cross-browser debugging

crisbeto:
Adds the `gulp test:static` task that is identical to `gulp test`, however it doesn't launch Chrome automatically, which is convenient for debugging any browser that isn't Chrome. Currently the debugging flow for other browsers is running the Gulp task which launches Chrome, waiting for it to finish running the tests, connect the other browser, wait for the new browser to finish and then start debugging. With the new task you can just run the Gulp task and connect and disconnect only the relevant browsers.

### DIFF
--- a/tools/gulp/tasks/unit-test.ts
+++ b/tools/gulp/tasks/unit-test.ts
@@ -5,7 +5,12 @@ import {buildConfig, sequenceTask} from 'material2-build-tools';
 // There are no type definitions available for these imports.
 const runSequence = require('run-sequence');
 
-const {packagesDir, projectDir} = buildConfig;
+// Default Karma options.
+const defaultOptions = {
+  configFile: join(buildConfig.projectDir, 'test/karma.conf.js'),
+  autoWatch: false,
+  singleRun: false
+};
 
 /** Builds everything that is necessary for karma. */
 task(':test:build', sequenceTask(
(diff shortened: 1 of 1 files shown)


tools/gulp/tasks/unit-test.ts
```diff
@@ -35,39 +35,55 @@ task('test:single-run', [':test:build'], (done: () => void) => {
...
+    const karma = require('karma');
 
-  // Boot up the test server and run the tests whenever a new browser connects.
-  server.start();
-  server.on('browser_register', () => runTests());
+    // Configure the Karma server and override the autoWatch and singleRun just in case.
+    const server = new karma.Server({...({
+      configFile: join(projectDir, 'test/karma.conf.js'),
+      autoWatch: false,
+      singleRun: false
```
DevVersion: We can maybe move the "default options" as a constant to the top of the file.

crisbeto: I think that if it's moved out, it implies that it's also relevant for the `test:single-run` task which it's not.

DevVersion: The only difference for the `single-run` task is the `singleRun` property. We can just have those options (e.g `configFile`, `autoWatch`) as default options and overwrite the `singleRun` option later as well.

That way we repeat less code and it's more obvious which options are used. 

```ts
new karma.Server({...defaultKarmaOptions, ...overwrites});
```

crisbeto: Makes sense. It's added.

PR #7798, opened 2017-10-14
chore: run tests against desktop safari 11

DevVersion:
* Recently we updated the browser providers to run the tests against iOS11 with Safari 11 Mobile. With this change Safari 10 and 11 on desktop will be tested (latest two versions)


test/browser-providers.js
```diff
@@ -22,8 +22,9 @@ const browserConfig = {
   'Android5':          { unitTest: {target: null, required: false }},
   'Safari7':           { unitTest: {target: null, required: false }},
   'Safari8':           { unitTest: {target: null, required: false }},
-  'Safari9':           { unitTest: {target: 'SL', required: true  }},
-  'Safari10':          { unitTest: {target: 'BS', required: true  }},
+  'Safari9':           { unitTest: {target: null, required: false  }},
```
jelbourn: I'd prefer to keep Safari 9 on until we have a compelling reason to remove it

DevVersion: We'd still test against the latest two versions of Safari. (11 and 10). Is there really a need to keep testing Safari 9?

jelbourn: It's useful if we discover a fault that's easy to fix. We only need to _stop_ if there's some significant support burden.

DevVersion: I understand. Not sure what we should do then. I don't think it's good running tests against, 9, 10 and 11. This will probably cause more flakiness.

jelbourn:
Let's try it and see how it goes; nothing is set in stone.

PR #7805, opened 2017-10-15
feat(overlay): add option to re-use last preferred position when re-applying to open connected overlay

crisbeto:
Currently when updating the position of an open connected overlay (e.g. when the user is scrolling) we go through the same process for determining the preferred position as when the overlay was attached. This means that the preferred position could change, causing the overlay to jump. With these changes the consumer can decide to lock an overlay into its initial position, preventing it from jumping.

This PR is a resubmit of #5471.

### DIFF
--- a/src/cdk/overlay/position/connected-position-strategy.spec.ts
+++ b/src/cdk/overlay/position/connected-position-strategy.spec.ts
@@ -440,6 +440,29 @@ describe('ConnectedPositionStrategy', () => {
       expect(Math.floor(overlayRect.left)).toBe(Math.floor(originRect.left));
     });
 
+    it('should re-use the preferred position when re-applying while locked in', () => {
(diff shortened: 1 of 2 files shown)


jelbourn:
I'm not sure about this- I've always considered it a feature that overlays can flip positions to stay on-screen when scrolling
(I even showed it off during my AngularMix talk)

What's your reasoning for wanting to change it?

crisbeto:
My reasoning was that it could be distracting having the overlay "follow you" after you've decided to scroll away. Alternatively we can make this functionality opt-in and only enable it on the components where we think it makes sense (e.g. the select doesn't handle the repositioning very well because it also has to change the `offsetY` and the scroll position).

jelbourn:
I'd be okay with it being a setting. Not sure on the name, though.

crisbeto:
@jelbourn I've reworked it to make the functionality opt-in.

PR #7807, opened 2017-10-15
refactor(select): simplify option sync and more robust unit tests

crisbeto:
* Currently the select sets the `multiple` and `disableRipple` values on its options manually. In order to avoid "changed after checked" errors we wrap the calls in promises, however this also forces us to have `async` tests which can time out if the browser is out of focus. These changes add a provider that allows for the options to take the value directly from the select.
* Refactors some unit tests that have been timing out in Firefox to run in the `fakeAsync` zone instead of the `async` one.
* Fixes a positioning test that fails occasionally in Firefox.

### DIFF
--- a/src/lib/core/option/option.spec.ts
+++ b/src/lib/core/option/option.spec.ts
@@ -29,7 +29,7 @@ describe('MatOption component', () => {
     });
 
     it('should show ripples by default', () => {
-      expect(optionInstance.disableRipple).toBe(false, 'Expected ripples to be enabled by default');
+      expect(optionInstance.disableRipple).toBeFalsy('Expected ripples to be enabled by default');
       expect(optionNativeElement.querySelectorAll('.mat-ripple-element').length)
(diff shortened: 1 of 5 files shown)


src/lib/core/option/option.spec.ts
```diff
@@ -54,20 +54,6 @@ describe('MatOption component', () => {
         .toBe(0, 'Expected no ripples to show up after click on a disabled option.');
     });
 
-    it('should not show ripples if the ripples are disabled using disableRipple', () => {
```
kara: Can you move this test to the select tests rather than removing? e.g. `select.disableRipple = true`.

Seems like a good one to have, given that select only tests the presence of the property.

crisbeto: We have a test that checks it already: https://github.com/angular/material2/blob/master/src/lib/select/select.spec.ts#L339

crisbeto:
I've addressed the feedback @kara, can you take another look?

PR #7808, opened 2017-10-15
fix(select): active item not being updated on click in multiple mode

crisbeto:
Fixes the active option not being updated when the user clicks inside a multi-select, causing it to be stuck on the first option.


crisbeto:
@jelbourn could you take a look since it's an accessibility issue?

jelbourn:
Looks like it has a conflict already

PR #7810, opened 2017-10-15
chore(spinner): avoid deprecation warning in Safari

crisbeto:
Fixes a deprecation warning that was being logged by the spinner on Safari.


crisbeto:
@jelbourn bumping this to a P2, because it seems to be breaking slightly older versions on Firefox. See https://github.com/angular/material2/issues/8064.

mayankku:
Any ETA on this?

PR #7831, opened 2017-10-16
fix(datepicker): prevent `matInput` from clobbering date value

### LINKED ISSUE #7717: (Datepicker) Initial Value giving full datetime
Bug/Support

When I send in a new moment or date object into the datepicker it should format like it does when you click a new date.

The initial format is the full datetime string instead of MM/DD/YYYY. It might be I am not sending in it in right or something but there is no documentation I can find that says to do it different.

https://stackblitz.com/edit/angular-material2-issue-y6hdoi

I have tried new Date('12/06/2017') as well and still get the full datetime string.

To have the datepicker set a value on init that is from data but formatted the way the datepicker selects dates.

5.0.0-rc.1

mmalerba:
fixes #7717

### DIFF
--- a/src/lib/input/input-value-accessor.ts
+++ b/src/lib/input/input-value-accessor.ts
@@ -6,10 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {InjectionToken} from '@angular/core';
 
-export * from './input-module';
-export * from './autosize';
-export * from './input';
-export * from './input-errors';
-export * from './input-value-accessor';
 
+export const MAT_INPUT_VALUE_ACCESSOR =
+    new InjectionToken<{value: any}>('MAT_INPIUT_VALUE_ACCESSOR');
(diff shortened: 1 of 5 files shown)


src/lib/input/input-value-accessor.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {InjectionToken} from '@angular/core';
+
+
+export const MAT_INPUT_VALUE_ACCESSOR =
```
crisbeto: Add a doc describing what this is for?

src/lib/input/input-value-accessor.ts
```diff
@@ -0,0 +1,13 @@
...
+    new InjectionToken<{value: any}>('MAT_INPIUT_VALUE_ACCESSOR');
```
rafaelss95: typo?

mmalerba:
Done

kiancross:
Any ETA for when this will be merged?

PR #7842, opened 2017-10-17
fix(radio) detect non-direct mat-radio descendants of mat-radio-group

### LINKED ISSUE #7166: Radio group for radio buttons not working with HTML elements between
When there are HTML elements between <md-radio-group> and <md-radio> the group is not working / model propagation

The model is not propagated to the radio buttons / the default button is not selected

```
           <md-radio-group [(ngModel)]="predefined">
                <div class="div-block">
                    <md-radio-button [value]="true">Predefined library</md-radio-button>
                </div>
                <div class="div-block">
                </div>
                <div class="div-block">
                    <md-radio-button [value]="false">Own library</md-radio-button>
                </div>
        </md-radio-group>
```

See image:
image

Predefined should be selected, when the divs are removed it works

beta.10

rkrzewski:
Fixes #7166

Here's the reproduction Plunk corresponding to the added unit test: http://plnkr.co/edit/ZQ721sXbVJfDjn3kSnCw?p=preview

This is my first PR to Material 2 so please bear with me. Any suggestions how to improve it are welcome.

### DIFF
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -18,6 +18,8 @@ describe('MatRadio', () => {
         RadioGroupWithNgModel,
         RadioGroupWithFormControl,
         StandaloneRadioButtons,
+        InterleavedRadioGroup,
+        TranscludingWrapper
       ],
       providers: [
         {provide: ViewportRuler, useClass: FakeViewportRuler}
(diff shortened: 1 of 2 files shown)


rkrzewski:
There is a number of failed tests in https://travis-ci.org/angular/material2/jobs/288966427 but they seem unrelated to the change I made. Please advise.

src/lib/radio/radio.spec.ts
```diff
@@ -760,3 +784,23 @@ class RadioGroupWithFormControl {
   template: `<mat-radio-button tabindex="-1"></mat-radio-button>`
 })
 class FocusableRadioButton {}
+
+@Component({
+  template: `
+  <mat-list>
```
crisbeto: In general we try to avoid mixing Material components between from different modules each other. You should be able to test the same either by wrapping the radio button in a div, or adding a custom component that is only used in the tests.

rkrzewski: OK, that makes sense. I will update the tests accordingly.

rkrzewski:
@crisbeto, PTAL. I've replaced `mat-list`/`mat-list-item` with minimal transcluding component. I've verified that the modified test case fails with `descendants: false` on `_radios` and succeeds with `descendatns: true`.

rkrzewski:
Thank you for the reviews @crisbeto and @tinayuangao. Can I do something more to push this one through the finish line?

crisbeto:
Nothing more is necessary @rkrzewski, it will be synced in when the caretaker gets to it.

PR #7850, opened 2017-10-17
bug(fab): fix alignment issues

amcdnl:
This PR:

- Fixes alignment of icons + buttons in fabs
- Decreases font size of text in mini-fabs

img


src/lib/button/button.scss
```diff
@@ -24,11 +24,12 @@
...
 
 .mat-fab {
-  @include mat-fab($mat-fab-size, $mat-fab-padding);
+  @include mat-fab($mat-fab-size);
 }
 
 .mat-mini-fab {
-  @include mat-fab($mat-mini-fab-size, $mat-mini-fab-padding);
+  @include mat-fab($mat-mini-fab-size);
+  font-size: 80%;
```
jelbourn: I don't see anything in the spec that talks about font-size for the mini fab- what are you basing this on?

amcdnl: Basing on it was unproportionate in the smaller use case. Happy to remove if you would like.

src/lib/button/_button-base.scss
```diff
@@ -96,7 +94,6 @@ $mat-mini-fab-padding: 8px !default;
   border-radius: $mat-fab-border-radius;
   width: $size;
   height: $size;
-  padding: 0;
```
tinayuangao: Not sure about this change. On my demo the text in the fab/mini-fab are off. 
image

src/lib/button/button.scss
amcdnl: Removed.

src/lib/button/_button-base.scss
amcdnl: Must have been cache, fixed.

amcdnl:
@josephperrott - any feedback or can we merge?

josephperrott:
@amcdnl go ahead and rebase and we can try to get it in.

jelbourn:
Closing this since we're revamping button to be based on MDC web which should fix this separately

PR #7852, opened 2017-10-17
chore(keycodes): add comma to keycodes and update usages

amcdnl:
This PR adds COMMA to the keycodes and updates all the usages where it was manually defined.


amcdnl:
Note: It would be cool to change this to a string enum and add the rest of the keycodes.

amcdnl:
@jelbourn - Gotcha. Thanks for the feedback.

Issue #7211, opened 2017-09-20
MatSelect: Long options are truncated


jermowery:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Options should wrap or be shown in some way that does not hide the option text from the user.

#### What is the current behavior?
Long option text is truncated.

#### What are the steps to reproduce?

http://plnkr.co/edit/X8JZAatPpRD6RBSRgT7K?p=preview

#### What is the use-case or motivation for changing an existing behavior?

Text can be long in options, the user should not have the option text hidden from them.
For translated text that is longer than the source text this can happen more frequently.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Latest

fxck:
related https://github.com/angular/material2/issues/3883

dixy:
In a select in multiple selection mode, the option text is entirely hidden for long options on at least Firefox and IE 11.

![firefox-56](https://user-images.githubusercontent.com/1023798/31666042-21844276-b34b-11e7-9948-28b62e5a9ff7.jpg)
Firefox 56

![chrome-61](https://user-images.githubusercontent.com/1023798/31666048-298ded5a-b34b-11e7-93b6-da26db2d3764.jpg)
Chrome 61 is almost good (two dots hidden)

crisbeto:
That is a separate issue from the one being discussed here @dixy, but thanks for the report. It should be fixed by #7857.

UncleJimmy:
I want to add a tooltip on this "mat-option-text" as workaround, is there any way we could know the text is truncated or not?

srix55:
Any way we can stop the truncate and allow the text to take the next line?

Hydhen:
@UncleJimmy I've seen this workaround :
```
function isEllipsisActive(e) {
     return (e.offsetWidth < e.scrollWidth);
}
```
From this [link](https://stackoverflow.com/questions/7738117/html-text-overflow-ellipsis-detection)

But I didn't succeeded on getting the propeties from an element node, i'm stuck to this point : 

 ```
@ViewChild('select') select: MatSelect;
this.select.options.forEach((option: MatOption) => {
      console.log(option._getHostElement().childNodes[1]); // Gives the span element
});
```

**Comment by Hydhen** [comment]
Just find out you can do this :

```
[...]

PR #7857, opened 2017-10-17
fix(select): option not truncated correctly in multiple mode

crisbeto:
Currently, when the text of an option is truncated, it ends up removing everything except the ellipsis on Firefox, IE and Edge. These changes fix the truncation and simplify the option view. This seems to have started happening after the switch to `mat-form-field`.

Relates to #7211.


crisbeto:
Also bonus fun fact: Apparently Pokemon names can't be longer than 10 characters.

mmalerba:
I assume that fun fact is no accident, makes things easier on Nintendo developers ;)

PR #7858, opened 2017-10-17
fix(datepicker): correct DST issues on IE 11

### LINKED ISSUE #7167: Day incorrect in datepicker
When I select a date I see the correct date in the field but, when I save, the datepicker send the day before the date I have selected

here's the code:
`<md-form-field> <input mdInput 														[(ngModel)]="fareCalendar.startDate" name="startDate" 														[mdDatepicker]="picker"  														placeholder="startDate"> 													<md-datepicker-toggle mdSuffix [for]="picker"></md-datepicker-toggle> 													<md-datepicker #picker></md-datepicker> </md-form-field>`

I am using angular 4  and @angular/material": "^2.0.0-beta.10
rtcnm

but the date:
f5z10

How can I send the UTC date

### LINKED ISSUE #8228: IE 11 Datepicker not working
Bug

Datepicker should work

Datepicker is throwing exception

Use IE 11, go to https://material.angular.io/components/datepicker/overview, then try open datepicker

Material Beta.12

I think this is introduced after https://github.com/angular/material2/pull/5747. 

https://github.com/angular/material2/blob/master/src/lib/core/datetime/native-date-adapter.ts#L74, useUtcForDisplay is hard coded to true. According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat IE 11 doesn't support timeZone option. 

I tested the following script on IE 11 console, it is returning error.

var options = { year: 'numeric', month: 'short', timeZone: 'utc' };
console.log(new Intl.DateTimeFormat('en-US', options).format(new Date()));

mmalerba:
fixes #7167 
fixes #8228

### DIFF
--- a/src/lib/core/datetime/native-date-adapter.ts
+++ b/src/lib/core/datetime/native-date-adapter.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {Platform} from '@angular/cdk/platform';
 import {Inject, Injectable, Optional} from '@angular/core';
 import {extendObject} from '../util/object-extend';
 import {DateAdapter, MAT_DATE_LOCALE} from './date-adapter';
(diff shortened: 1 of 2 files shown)


mmalerba:
demo: https://mmalerba-demo1.firebaseapp.com/datepicker

src/lib/core/datetime/native-date-adapter.ts
```diff
@@ -59,18 +60,21 @@ function range<T>(length: number, valueFunction: (index: number) => T): T[] {
...
    */
-  useUtcForDisplay = true;
+  useUtcForDisplay: boolean;
+
+  constructor(@Optional() @Inject(MAT_DATE_LOCALE) matDateLocale: string, platform: Platform) {
+    super();
+    super.setLocale(matDateLocale);
+
+    // IE does its own time zone correction, so we disable this on IE.
+    this.useUtcForDisplay = !platform.TRIDENT;
```
crisbeto: 1. Can this be feature-checked (e.g. creating a date and verifying that the time is correct)? 
2. Is it IE-exclusive or does it happen on Edge as well?

mmalerba: 1. Not really, it depends on the locale the user is in and when / if they have DST
2. IE only

I also intentionally left this public so it can be overridden by users as an easy fix while I investigate any future bugs related to this

mmalerba:
need to fix test failures

mmalerba:
tests are good now

Issue #5564, opened 2017-07-06
@angular/cdk requires rxjs >= v5.0.3 (@angular/material depencency is ^5.0.1)


danwulff:
#### Bug, feature request, or proposal:
Bug.

#### What is the expected behavior?
AOT compile should work down to rxjs v5.0.1 (as specified in package.json)

#### What is the current behavior?
Get the following error during AOT compile using angular-compiler & rollup (specifically using a fork of https://github.com/filipesilva/angular-quickstart-lib)

`Error at .../node_modules/@angular/cdk/typings/rxjs/rx-operators.d.ts:11:10: Module '".../node_modules/rxjs/Scheduler"' has no exported member 'IScheduler'.`

#### What are the steps to reproduce?

Don't have a plunkr since AOT but file that I believe is causing the issue is here: https://github.com/angular/material2/blob/2.0.0-beta.8/src/cdk/rxjs/rx-operators.ts

specifically ln. 12 `import {IScheduler} from 'rxjs/Scheduler';`

Since CDK requires `rxjs` Scheduler to export interface `IScheduler` can we increment the dependency to ^5.0.3 (when IScheduler was added to rxjs)

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?

@angular/___: v4.1.3
@angular/cdk: v2.0.0-beta.8
@angular/material: v2.0.0-beta.8
rxjs: v5.0.1
typescript: v2.3.4
Win 10

d3viant0ne:
Works fine on the current version of rxjs in the `package.json`. Not terribly enthused about having to roll back to `^5.0.1` but it does work.

danwulff:
It looks like `IScheduler` was added at rxjs v5.0.3. So I'm not sure why I'm seeing an issue with v5.4.2. I'll do some more digging and reopen if I find something worthwhile.

danwulff:
Figured out that my integration tests were utilizing rxjs v5.0.1, reopened since package.json in @angular/material has rxjs: `^5.0.1`.

Either the depencency in the root package.json should be incremented
or
the `@angular/cdk` package.json should have a rxjs dependency added with `^5.0.3`.

vinupdlm:
I upgraded rxjs version from 5.0.1 to 5.0.3. It works. !!!! Thank a lot @danwulff

danwulff:
@jelbourn I believe this should now fixed by: https://github.com/angular/material2/pull/7276

jelbourn:
Yep

PR #7865, opened 2017-10-17
fix(cdk): increment rxjs ver to support IScheduler dependency

danwulff:
Fixes: https://github.com/angular/material2/issues/5564


crisbeto:
Master has been bumped to rxjs@5.5.0.

PR #7888, opened 2017-10-18
chore(dialog): move backdrop click logic into MatDialogRef

josephperrott:


src/lib/dialog/dialog-ref.ts
```diff
@@ -23,11 +24,14 @@ let uniqueId = 0;
...
 export class MatDialogRef<T> {
+  /** A unique id for the dialog. */
+  readonly id: string;
+
   /** The instance of component opened into the dialog. */
   componentInstance: T;
 
   /** Whether the user is allowed to close the dialog. */
-  disableClose = this._containerInstance._config.disableClose;
+  disableClose = this._config.disableClose;
```
crisbeto: Why not continue getting the config through the container instance? That way we avoid having to a reference to the config in two places.

jelbourn:
@josephperrott provide more information for context?

PR #7893, opened 2017-10-18
fix(table): update implicit when using trackby

### LINKED ISSUE #7877: [Table] Rows are not provided correct $implicit with trackBy
#### Bug:
MatTable doesn't refresh row columns if [trackBy] function used. For example if I replace rows array with a new one having the same ids, but completely different data MatTable won't update. I suspect the same applies to CdkTable.
It looks like it correctly updates table when removing / adding rows but it fails to trigger ChangeDetection on <mat-row> components if no rows added or removed

When used with trackBy function Table should trigger change detection on all existing rows, just like ngFor with trackBy do.

https://stackblitz.com/edit/angular-material2-issue-tbrfqj?file=app%2Fapp.component.html

andrewseguin:
The table should make sure that if it wants to re-use a row, then it's implicit data should be accurate. It's possible that a row's data is entirely changed but the row was re-used due to a loose `trackBy`. 

Fixes #7877

### DIFF
--- a/src/cdk/table/table.spec.ts
+++ b/src/cdk/table/table.spec.ts
@@ -391,6 +391,24 @@ describe('CdkTable', () => {
       expect(changedRows[1].getAttribute('initialIndex')).toBe('1');
       expect(changedRows[2].getAttribute('initialIndex')).toBe(null);
     });
+
+    it('should change row implicit data even when trackBy finds no changes', () => {
+      createTestComponentWithTrackyByTable('index');
+      const firstRow = getRows(tableElement)[0];
+      expect(firstRow.textContent!.trim()).toBe('a_1 b_1');
+      expect(firstRow.getAttribute('initialIndex')).toBe('0');
+      mutateData();
+
+      // Change each item reference to show that the trackby is checking the index.
+      // Otherwise this would cause them all to be removed/added.
+      trackByComponent.dataSource.data = trackByComponent.dataSource.data
+          .map(item => { return {a: item.a, b: item.b, c: item.c}; });
+
+      // Expect the rows were given the right implicit data even though the rows were not moved.
+      trackByFixture.detectChanges();
+      expect(firstRow.textContent!.trim()).toBe('a_2 b_2');
+      expect(firstRow.getAttribute('initialIndex')).toBe('0');
+    });
   });
 
   it('should match the right table content with dynamic data', () => {
--- a/src/cdk/table/table.spec.ts
+++ b/src/cdk/table/table.spec.ts
@@ -402,7 +402,7 @@ describe('CdkTable', () => {
       // Change each item reference to show that the trackby is checking the index.
       // Otherwise this would cause them all to be removed/added.
       trackByComponent.dataSource.data = trackByComponent.dataSource.data
-          .map(item => { return {a: item.a, b: item.b, c: item.c}; });
+          .map(item => ({a: item.a, b: item.b, c: item.c}));
 
       // Expect the rows were given the right implicit data even though the rows were not moved.
       trackByFixture.detectChanges();
--- a/src/cdk/table/table.ts
+++ b/src/cdk/table/table.ts
@@ -69,6 +69,12 @@ export const CDK_TABLE_TEMPLATE = `
   <ng-container rowPlaceholder></ng-container>`;
 
 /**
+ * Class used to conveniently type the embedded view ref for rows with a context.
+ * @docs-private
+ */
+abstract class RowViewRef<T> extends EmbeddedViewRef<CdkCellOutletRowContext<T>> { }
+
+/**
  * A data table that connects with a data source to retrieve data of type `T` and renders
  * a header row and data rows. Updates the rows when new data is provided by the data source.
  */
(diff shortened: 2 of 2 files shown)


src/cdk/table/table.ts
```diff
@@ -68,6 +68,14 @@ export const CDK_TABLE_TEMPLATE = `
   <ng-container headerRowPlaceholder></ng-container>
   <ng-container rowPlaceholder></ng-container>`;
 
+/** Class used to conveniently type the embedded view ref for rows with a context. */
+abstract class RowViewRef<T> extends EmbeddedViewRef<CdkCellOutletRowContext<T>> { }
```
jelbourn: Should this be `@docs-private`?

src/cdk/table/table.ts
```diff
@@ -68,6 +68,14 @@ export const CDK_TABLE_TEMPLATE = `
...
+
+/** Tuple used to pair together an iterable differ change record with the view it affects. */
+class RowRecordViewTuple<T> {
```
jelbourn: `@docs-private`?

src/cdk/table/table.ts
```diff
@@ -68,6 +68,14 @@ export const CDK_TABLE_TEMPLATE = `
...
+  constructor(public record: IterableChangeRecord<T>, public view: RowViewRef<T>) {}
```
jelbourn: Should `record` and `view` be public, documented properties?

jelbourn: Also neither seem to be the most descriptive names

src/cdk/table/table.ts
```diff
@@ -292,25 +300,45 @@ export class CdkTable<T> implements CollectionViewer {
...
-  /** Check for changes made in the data and render each change (row added/removed/moved). */
+  /**
+   * Check for changes made in the data and render each change (row added/removed/moved) and update
+   * row contexts.
+   */
   private _renderRowChanges() {
     const changes = this._dataDiffer.diff(this._data);
     if (!changes) { return; }
 
+    const insertOrMovedRowTuples: RowRecordViewTuple<T>[] = [];
```
jelbourn: Add comment that explains why you keep track of the inserted or moved tuples?

src/cdk/table/table.ts
```diff
@@ -292,25 +300,45 @@ export class CdkTable<T> implements CollectionViewer {
...
+    for (let i = 0; i < insertOrMovedRowTuples.length; i++) {
+      const tuple = insertOrMovedRowTuples[i];
+      this._updateRowImplicitContext(tuple.view, tuple.record);
+    }
+
+    this._updateRowMetaContext();
+
+    // Update rows that did not get added/removed/moved but may have had their identity changed,
+    // e.g. if trackBy matched data on the property 'id' but the property 'name' was changed.
+    changes.forEachIdentityChange((record: any) => {
```
jelbourn: Why `any`?

src/cdk/table/table.ts
```diff
@@ -350,17 +379,24 @@ export class CdkTable<T> implements CollectionViewer {
...
+  }
+
+  /** Updates the row view ref with the implicit data assigned to the row. */
+  private _updateRowImplicitContext(view: RowViewRef<T>, record: IterableChangeRecord<any>) {
+    view.context.$implicit = record.item;
   }
 
   /**
-   * Updates the context for each row to reflect any data changes that may have caused
+   * Updates the meta context for each row to reflect any data changes that may have caused
```
jelbourn: "Meta context"?

src/cdk/table/table.ts
```diff
@@ -350,17 +379,24 @@ export class CdkTable<T> implements CollectionViewer {
     });
 
     this._changeDetectorRef.markForCheck();
+
+    return rowView;
+  }
+
+  /** Updates the row view ref with the implicit data assigned to the row. */
+  private _updateRowImplicitContext(view: RowViewRef<T>, record: IterableChangeRecord<any>) {
```
jelbourn: I think the code would be easier to read without this method (just having the assignment inline). Any reason to keep it a method?

src/cdk/table/table.ts
```diff
@@ -292,25 +300,45 @@ export class CdkTable<T> implements CollectionViewer {
...
           } else {
-            const view = viewContainer.get(adjustedPreviousIndex);
+            const view = <RowViewRef<T>>viewContainer.get(adjustedPreviousIndex);
             viewContainer.move(view!, currentIndex);
+            insertOrMovedRowTuples.push(new RowRecordViewTuple<T>(record, view));
           }
         });
 
-    this._updateRowContext();
+    // Update the row's implicit context for each inserted or moved row.
```
jelbourn: "Update the $implicit property for the binding context for inserted and moved rows"

andrewseguin:
tl;dr: Took out logic about insertions and moves. Just updating for identity changes since it addresses the issue.

Reducing the scope of this PR - was looking at Angular's `ngFor` (`ng_for_of.ts`) to see how they addressed this issue. They updated row `$implicit` on insert/move and also for any identity changes. The issue was fixed with the identity change and I'm not convinced we need the former (updating on insert/move, which should be addressed already on insert).

If another issue comes in about bad `$implicit` on moves then we'll know where to look, but I'm not 100% convinced it's needed.

src/cdk/table/table.spec.ts
```diff
@@ -391,6 +391,24 @@ describe('CdkTable', () => {
...
+      createTestComponentWithTrackyByTable('index');
+      const firstRow = getRows(tableElement)[0];
+      expect(firstRow.textContent!.trim()).toBe('a_1 b_1');
+      expect(firstRow.getAttribute('initialIndex')).toBe('0');
+      mutateData();
+
+      // Change each item reference to show that the trackby is checking the index.
+      // Otherwise this would cause them all to be removed/added.
+      trackByComponent.dataSource.data = trackByComponent.dataSource.data
+          .map(item => { return {a: item.a, b: item.b, c: item.c}; });
```
jelbourn:
```ts
.map(item => ({a: item.a, b: item.b, c: item.c}));
```

PR #7895, opened 2017-10-19
fix(MatPaginatorIntl): Provide a provider if exists.

### LINKED ISSUE #7344: [Paginator] Add app-wide configuration for MatPaginatorIntl
Kind of a bug.

MdPaginatorIntl provided in forRoot function. That way in can both app-wide and locally configured.

Currently, MdPaginatorModule provides copy of MdPaginatorIntl for each import of module.
https://github.com/angular/material2/blob/0e4806b06c31c4968cd23a16f73bdcee8fe44cde/src/lib/paginator/paginator-module.ts

I have a localization service, that sets paginator intl variables
```typescript
    this.storage.subscribe((storage) => {
      this.mdPaginatorIntl.itemsPerPageLabel = storage.paginator__itemsPerPageLabel;
      this.mdPaginatorIntl.nextPageLabel = storage.paginator__nextPageLabel;
      this.mdPaginatorIntl.previousPageLabel = storage.paginator__previousPageLabel;
      this.mdPaginatorIntl.getRangeLabel = (page: number, pageSize: number, length: number): string => {
        return storage.paginator__getRangeLabel
          .replace('__START__', page.toString())
          .replace('__END__', (page + pageSize - 1).toString())
          .replace('__TOTAL__', length.toString());
      };
      this.mdPaginatorIntl.changes.next();
    });
```
but I can't do it once for whole application. 
If I just provide own copy of MdPaginatorIntl in AppModule, it will not be used. Because feature modules of app are import MdPaginatorModule with MdPaginatorIntl provider.
I have tried to make my own paginator module, but then `Type MdPaginator in ... is part of the declarations of 2 modules:`

angular 4.4.3
material 2.0.0-beta.11

Jefiozie:
If there is already a MatPaginatorIntl provider that one.  If not create a new instance.

@andrewseguin  I hope this is the correct implementation. If not let me know I will change it.
Next to this shouldn't this also be implemented for `sort-header-intl`
cc: @jelbourn
Closes #7344

### DIFF
--- a/src/lib/paginator/paginator-intl.ts
+++ b/src/lib/paginator/paginator-intl.ts
@@ -48,7 +48,7 @@ export class MatPaginatorIntl {
 }
 
 /** @docs-private */
-export function MAT_PAGINATOR_INTL_PROVIDER_FACTORY(
+export function parentIntl(
     parentMatPaginatorIntl: MatPaginatorIntl) {
   return parentMatPaginatorIntl || new MatPaginatorIntl();
 }
@@ -58,5 +58,5 @@ export const MAT_PAGINATOR_INTL_PROVIDER = {
   // If there is already an MatPaginatorIntl available, use that. Otherwise, provide a new one.
   provide: MatPaginatorIntl,
   deps: [[new Optional(), new SkipSelf(), MatPaginatorIntl]],
-  useFactory: MAT_PAGINATOR_INTL_PROVIDER_FACTORY
+  useFactory: parentIntl
 };
--- a/src/lib/paginator/paginator-intl.ts
+++ b/src/lib/paginator/paginator-intl.ts
@@ -48,9 +48,8 @@ export class MatPaginatorIntl {
 }
 
 /** @docs-private */
-export function parentIntl(
-    parentMatPaginatorIntl: MatPaginatorIntl) {
-  return parentMatPaginatorIntl || new MatPaginatorIntl();
+export function MAT_PAGINATOR_INTL_PROVIDER_FACTORY(parentIntl: MatPaginatorIntl) {
+  return parentIntl || new MatPaginatorIntl();
 }
 
 /** @docs-private */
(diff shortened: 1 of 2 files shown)


Jefiozie:
@andrewseguin the screenshot test failed but it looks like it is different because of the focus on the input. Don't know if I did this but I don't see how.

Jefiozie:
@andrewseguin or @jelbourn, Any idea when this pr will be reviewed? Happy to change things of needed.
Thanks.

src/lib/paginator/paginator-intl.ts
```diff
@@ -46,3 +46,17 @@ export class MatPaginatorIntl {
     return `${startIndex + 1} - ${endIndex} of ${length}`;
   }
 }
+
+/** @docs-private */
+export function MAT_PAGINATOR_INTL_PROVIDER_FACTORY(
```
jelbourn: You can just call the arg `parentIntl`

src/lib/paginator/paginator-module.ts
```diff
@@ -12,7 +12,7 @@ import {MatButtonModule} from '@angular/material/button';
 import {MatSelectModule} from '@angular/material/select';
 import {MatTooltipModule} from '@angular/material/tooltip';
 import {MatPaginator} from './paginator';
-import {MatPaginatorIntl} from './paginator-intl';
+import { MAT_PAGINATOR_INTL_PROVIDER} from './paginator-intl';
```
jelbourn: Omit space at beginning of import

Jefiozie:
@jelbourn just pushed a new commit with I believe the requested changes.

src/lib/paginator/paginator-intl.ts
```diff
@@ -48,7 +48,7 @@ export class MatPaginatorIntl {
 }
 
 /** @docs-private */
-export function MAT_PAGINATOR_INTL_PROVIDER_FACTORY(
+export function parentIntl(
```
jelbourn: Sorry, I wasn't totally clear. I meant the argument to the function (so that it fits on one line):
```ts
export function MAT_PAGINATOR_INTL_PROVIDER_FACTORY(parentIntl: MatPaginatorIntl) {
  return parentMatPaginatorIntl || new MatPaginatorIntl();
}
```

Jefiozie: okay, will make the change, just to be sure with `parentMatPaginatorIntl` you mean the `parentIntl` argument correct?

Jefiozie: Also did you see my first comment, should pattern also be implemented for the `sort-header-intl`

jelbourn:
Feel free to send a similar PR for sorting

PR #7903, opened 2017-10-19
chore(ripple): no longer update if `_matRippleTrigger` changes

DevVersion:
* Since the `md-` prefix is now removed completely, the `_matRippleTrigger` property is just `trigger` now.

### DIFF
--- a/src/lib/core/ripple/ripple.ts
+++ b/src/lib/core/ripple/ripple.ts
@@ -100,7 +100,7 @@ export class MatRipple implements OnChanges, OnDestroy {
   }
 
   ngOnChanges(changes: SimpleChanges) {
-    if ((changes['trigger'] || changes['_matRippleTrigger']) && this.trigger) {
+    if (changes['trigger'] && this.trigger) {
       this._rippleRenderer.setTriggerElement(this.trigger);
     }
 


src/lib/core/ripple/ripple.ts
```diff
@@ -100,7 +100,7 @@ export class MatRipple implements OnChanges, OnDestroy {
   }
 
   ngOnChanges(changes: SimpleChanges) {
-    if ((changes['trigger'] || changes['_matRippleTrigger']) && this.trigger) {
+    if (changes['trigger'] && this.trigger) {
```
crisbeto: You should be able to switch this to `change.trigger` as well.

DevVersion: Done. Great idea

src/lib/core/ripple/ripple.ts
```diff
@@ -100,7 +100,7 @@ export class MatRipple implements OnChanges, OnDestroy {
   }
 
   ngOnChanges(changes: SimpleChanges) {
-    if ((changes['trigger'] || changes['_matRippleTrigger']) && this.trigger) {
+    if (changes.trigger && this.trigger) {
```
jelbourn: `trigger` needs to stay as a string or else Closure Compiler might rename it (which we don't want for input names)

DevVersion: Makes sense. Changed it back to be on the safe side.

PR #7906, opened 2017-10-19
build: autocomplete test failure in iOS Safari

crisbeto:
Fixes an autocomplete test failure in iOS Safari 10.


tinayuangao:
LGTM

PR #7912, opened 2017-10-19
feat(table): add access to filtered data in MatTableDataSource

andrewseguin:
Users should be able to grab the result of the **filtered data** from the material table data source. This is different than the final result offered to the table, which may have been paginated.

Useful for features like selection, since a "select all" button would select every item in the table, not just what is displayed on the current page.

### DIFF
--- a/src/lib/table/table.spec.ts
+++ b/src/lib/table/table.spec.ts
@@ -102,16 +102,19 @@ describe('MatTable', () => {
       // Change filter to a_1, should match one row
       dataSource.filter = 'a_1';
       fixture.detectChanges();
+      expect(dataSource.filteredData.length).toBe(1);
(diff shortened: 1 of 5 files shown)


src/lib/table/table.spec.ts
```diff
@@ -120,6 +123,7 @@ describe('MatTable', () => {
       // Change filter to empty string, should match all rows
       dataSource.filter = '';
       fixture.detectChanges();
+      expect(dataSource.filteredData.length).toBe(3);
```
jelbourn: Add a test that the filtered data is _correct_?

andrewseguin: Added a bit more to the test - good catch

PR #7913, opened 2017-10-19
test(select): switch all select tests to fakeAsync

mmalerba:


src/lib/select/select.spec.ts
```diff
@@ -1930,52 +1881,52 @@ describe('MatSelect', () => {
...
-      beforeEach(() => {
+      beforeEach(fakeAsync(() => {
         fixture = TestBed.createComponent(BasicSelect);
         fixture.detectChanges();
         select = fixture.debugElement.query(By.css('mat-select')).nativeElement;
-      });
+      }));
 
-      it('should set the role of the select to listbox', () => {
+      it('should set the role of the select to listbox', fakeAsync(() => {
```
crisbeto: Most of these tests aren't doing anything async, why do we have to bring them into the `fakeAsync` zone (aside from consistency)?

mmalerba: I just added it to everything for safety. Some of the tests that weren't marked as async actually were and could leak async operations into later tests. I figured this way we guarantee that won't happen

mmalerba:
No idea whats up with that one test. Verified that it looks correct when visiting the demo site on Edge. I opted to just not run the test on Edge (which is still better than prior to this PR where the test wasn't actually run on any browser due to a bug).

andrewseguin:
Needs rebase

PR #7914, opened 2017-10-19
fix(table): render cells even if data is falsy

### LINKED ISSUE #7902: [Table] Cells do not render if data is falsy
If the row data evaluates to false (e.g. `0`, `null`) then the cells will not render. This was meant to only evaluate when the data evaluates to `undefined` since that was the clue to the table that the row should not actually render out. However, now that a `when` predicate can be included in the user's row definitions, this is no longer necessary.

Plunker showing the issue: https://plnkr.co/edit/lLzdn0bSI43GjuHclJBk?p=preview

Change 

`const cells = rowData ? this._getCellTemplatesForRow(row) : [];`

to 

`const cells = this._getCellTemplatesForRow(row)`

andrewseguin:
This was an optimization assuming that undefined data was meant to create a row without cells. This is a faulty design decision and the logic is incorrect anyways (see issue, `0` and other falsy values caused empty rows).

Supported workaround to rendering empty rows is to use the `when` predicate that can define a row without columns.

Fixes #7902

### DIFF
--- a/src/cdk/table/table.spec.ts
+++ b/src/cdk/table/table.spec.ts
@@ -109,10 +109,10 @@ describe('CdkTable', () => {
   });
 
   it('should render cells even if row data is falsy', () => {
-    const evenSimplerCdkTableAppFixture = TestBed.createComponent(EvenSimplerCdkTableApp);
+    const booleanRowCdkTableAppFixture = TestBed.createComponent(EvenSimplerCdkTableApp);
     const evenSimplerCdkTableElement =
-        evenSimplerCdkTableAppFixture.nativeElement.querySelector('cdk-table');
-    evenSimplerCdkTableAppFixture.detectChanges();
+        booleanRowCdkTableAppFixture.nativeElement.querySelector('cdk-table');
+    booleanRowCdkTableAppFixture.detectChanges();
 
     expectTableToMatchContent(evenSimplerCdkTableElement, [
       [''], // Header row
--- a/src/cdk/table/table.spec.ts
+++ b/src/cdk/table/table.spec.ts
@@ -38,7 +38,8 @@ describe('CdkTable', () => {
         UndefinedColumnsCdkTableApp,
         WhenRowCdkTableApp,
         WhenRowWithoutDefaultCdkTableApp,
-        WhenRowMultipleDefaultsCdkTableApp
+        WhenRowMultipleDefaultsCdkTableApp,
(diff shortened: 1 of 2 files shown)


src/cdk/table/table.ts
```diff
@@ -342,10 +342,7 @@ export class CdkTable<T> implements CollectionViewer {
     //   CdkCellOutlet was instantiated as a result  of `createEmbeddedView`.
     this._rowPlaceholder.viewContainer.createEmbeddedView(row.template, context, index);
 
-    // Insert empty cells if there is no data to improve rendering time.
-    const cells = rowData ? this._getCellTemplatesForRow(row) : [];
-
-    cells.forEach(cell => {
+    this._getCellTemplatesForRow(row).forEach(cell => {
```
jelbourn: Needs a unit test

andrewseguin:
Added a unit test

src/cdk/table/table.spec.ts
```diff
@@ -655,6 +681,23 @@ class SimpleCdkTableApp {
...
+        <cdk-header-cell *cdkHeaderCellDef></cdk-header-cell>
+        <cdk-cell *cdkCellDef="let data"> {{data}} </cdk-cell>
+      </ng-container>
+
+      <cdk-header-row *cdkHeaderRowDef="['column_a']"></cdk-header-row>
+      <cdk-row *cdkRowDef="let row; columns: ['column_a']"></cdk-row>
+    </cdk-table>
+  `
+})
+class EvenSimplerCdkTableApp {
```
jelbourn: `TableWithBooleanDataSource`?

("EvenSimpler" doesn't tell you anything really)

src/cdk/table/table.spec.ts
```diff
@@ -109,10 +109,10 @@ describe('CdkTable', () => {
   });
 
   it('should render cells even if row data is falsy', () => {
-    const evenSimplerCdkTableAppFixture = TestBed.createComponent(EvenSimplerCdkTableApp);
+    const booleanRowCdkTableAppFixture = TestBed.createComponent(EvenSimplerCdkTableApp);
```
jelbourn: Name of the component too?
`EvenSimplerCdkTableApp` -> `TableWithBooleanDataSource`

andrewseguin: d'oh yes - my brain was thinking they were all renaming

PR #7927, opened 2017-10-20
feat(ripple): handle touch events

### LINKED ISSUE #7062: Ripples behavior on touch-enabled devices doesn't match Android ripples behavior

DevVersion:
* Now handles touch events properly.

**Note**: Tested it thoroughly already and it works great. Still want to test on more devices manually to make sure it works fine!

Fixes #7062

### DIFF
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -17,6 +17,12 @@ export const RIPPLE_FADE_IN_DURATION = 450;
 /** Fade-out duration for the ripples in milliseconds. This can't be modified by the speedFactor. */
 export const RIPPLE_FADE_OUT_DURATION = 400;
 
+/**
+ * Timeout for ignoring mouse events. Mouse events will be temporary ignored after touch
+ * events to avoid synthetic mouse events.
+ */
+const IGNORE_MOUSE_EVENTS_TIMEOUT = 800;
+
 export type RippleConfig = {
   color?: string;
   centered?: boolean;
@@ -40,14 +46,17 @@ export class RippleRenderer {
   /** Element which triggers the ripple elements on mouse events. */
   private _triggerElement: HTMLElement | null;
 
-  /** Whether the mouse is currently down or not. */
-  private _isMousedown: boolean = false;
+  /** Whether the pointer is currently down or not. */
+  private _isPointerDown = false;
 
   /** Events to be registered on the trigger element. */
   private _triggerEvents = new Map<string, any>();
 
   /** Set of currently active ripple references. */
   private _activeRipples = new Set<RippleRef>();
+
+  /** Time in milliseconds when the last touchstart event happened. */
+  private _lastTouchStartEvent: number;
 
   /** Ripple config for all ripples created by events. */
   rippleConfig: RippleConfig = {};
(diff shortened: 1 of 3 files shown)


src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -175,21 +187,37 @@ export class RippleRenderer {
     this._triggerElement = element;
   }
 
-  /** Function being called whenever the trigger is being pressed. */
+  /** Function being called whenever the trigger is being pressed using mouse. */
   private onMousedown(event: MouseEvent) {
-    if (!this.rippleDisabled) {
-      this._isMousedown = true;
+    const isSyntheticEvent = this._lastTouchStartEvent &&
+        Date.now() < this._lastTouchStartEvent + ignoreMouseEventsTimeout;
```
crisbeto: Rather than using the `Date.now()`, you should be able to take the `Event.timeStamp`. I'm not completely sure about the browser support though, you may have to double check.

src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -17,6 +17,12 @@ export const RIPPLE_FADE_IN_DURATION = 450;
 /** Fade-out duration for the ripples in milliseconds. This can't be modified by the speedFactor. */
 export const RIPPLE_FADE_OUT_DURATION = 400;
 
+/**
+ * Timeout for ignoring mouse events. Mouse events will be temporary ignored after touch
+ * events to avoid synthetic mouse events.
+ */
+const ignoreMouseEventsTimeout = 800;
```
crisbeto: This seems a little high? Wasn't the timeout 300ms?

src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -40,15 +46,18 @@ export class RippleRenderer {
   /** Element which triggers the ripple elements on mouse events. */
   private _triggerElement: HTMLElement | null;
 
-  /** Whether the mouse is currently down or not. */
-  private _isMousedown: boolean = false;
+  /** Whether the pointer is currently down or not. */
+  private _isPointerDown: boolean = false;
```
crisbeto: The `boolean` typing here is redundant.

src/lib/core/ripple/ripple-renderer.ts
DevVersion: On Polymer it's 2000ms and in Materialize it's 500. On Material 1.x we used 650ms as a touch buffer for the interaction service. 800ms should be fine, and it doesn't really matter that much anyway.

src/lib/core/ripple/ripple-renderer.ts
DevVersion: Done.

src/lib/core/ripple/ripple-renderer.ts
DevVersion: As discussed on Slack, for browser compatibility we'll be using `Date.now()`

src/lib/core/ripple/ripple-renderer.ts
crisbeto: The const name here should be uppercased to match the others' naming.

jelbourn:
@DevVersion lint error
```
ERROR: /home/travis/build/angular/material2/src/lib/core/ripple/ripple-renderer.ts[209, 1]: trailing whitespace
```

jelbourn:
@DevVersion which devices did you test on? We have pretty much everything in the office

DevVersion:
@jelbourn I tested it on a few browsers. Mostly desktop browsers.

 Chrome 62 (Win), Firefox 57 (Win), Edge, IE11, Safari 11, Chrome 62 (MacOS) and Chrome 62 (Android 7.0)

jelbourn:
I'd like to double check on a few devices before merging

PR #7937, opened 2017-10-21
lint: ensure all files are owned by someone

mmalerba:


tools/gulp/tasks/lint.ts
```diff
@@ -46,6 +54,56 @@ task('madge', ['material:clean-build'], () => {
...
+  if (errors) {
+    throw Error(`Found ${errors} files with no owner.`);
+  }
+});
+
+/** Check if the given path is owned by the given owned path matcher. */
+function isOwnedBy(path: string, ownedPath: IMinimatch) {
+  // If the owned path ends with `**` its safe to eliminate whole directories.
+  if (statSync(path).isFile() || ownedPath.pattern.endsWith('**')) {
+    return ownedPath.match('/' + path);
```
crisbeto: AFAIK `match(someStr)` ends up creating a Regex under the hood. This could be clearer and faster using `indexOf`.

tools/gulp/tasks/lint.ts
```diff
@@ -46,6 +54,56 @@ task('madge', ['material:clean-build'], () => {
...
+
+  if (errors) {
+    throw Error(`Found ${errors} files with no owner.`);
+  }
+});
+
+/** Check if the given path is owned by the given owned path matcher. */
+function isOwnedBy(path: string, ownedPath: IMinimatch) {
+  // If the owned path ends with `**` its safe to eliminate whole directories.
+  if (statSync(path).isFile() || ownedPath.pattern.endsWith('**')) {
```
crisbeto: nit: Might be better to move the `ownedPath.pattern.endsWith('**')` in front since it exits faster, because it doesn't have to hit the file system.

tools/gulp/tasks/lint.ts
```diff
@@ -46,6 +54,56 @@ task('madge', ['material:clean-build'], () => {
...
+      .map(line => line.trim())
+      // Remove empty lines and comments.
+      .filter(line => line && !line.startsWith('#'))
+      // Split off just the path glob and turn it into a regex.
+      .map(line => new Minimatch(line.split(/\s+/)[0], {dot: true, matchBase: true}))
+      // Add in the paths we're ignoring.
+      .concat(ignoreOwners.map(path => new Minimatch(path, {dot: true, matchBase: true})));
+
+  for (let paths = getChildPaths('.'); paths.length;) {
+    paths = Array.prototype.concat(...paths
```
crisbeto: You should be able to get rid of the `Array.prototype.concat` using a spread: `paths = [...paths.filter]`.

tools/gulp/tasks/lint.ts
```diff
@@ -46,6 +54,56 @@ task('madge', ['material:clean-build'], () => {
...
 });
 
+task('ownerslint', () => {
+  let errors = 0;
+
+  let ownedPaths = readFileSync(ownersFilePath, 'utf8').split('\n')
+      // Trim lines.
+      .map(line => line.trim())
+      // Remove empty lines and comments.
+      .filter(line => line && !line.startsWith('#'))
```
crisbeto: This won't handle cases like `/some/path owner1 owner2 #owner3`.

tools/gulp/tasks/lint.ts
```diff
@@ -19,7 +21,13 @@ const materialOutPath = join(buildConfig.outputDir, 'packages', 'material');
 /** Path to the output of the CDK package. */
 const cdkOutPath = join(buildConfig.outputDir, 'packages', 'cdk');
 
-task('lint', ['tslint', 'stylelint', 'madge']);
+/** Path for the Github owners file. */
+const ownersFilePath = '.github/CODEOWNERS';
+
+/** Paths that should be ignored when checking for owners coverage. */
+const ignoreOwners = ['/node_modules/**', '/dist/**', '/.git/**', '/.idea/**'];
```
crisbeto: This doesn't really cover all the paths (e.g. there's `/.vscode` which should also be excluded). Consider populating the list from the `.gitignore` instead?

tools/gulp/tasks/lint.ts
```diff
@@ -46,6 +54,56 @@ task('madge', ['material:clean-build'], () => {
...
+
+  for (let paths = getChildPaths('.'); paths.length;) {
+    paths = Array.prototype.concat(...paths
+        // Remove paths that match an owned path.
+        .filter(path => !ownedPaths.reduce(
+            (isOwned, ownedPath) => isOwned || isOwnedBy(path, ownedPath), false))
+        // Report an error for any files that didn't match any owned paths.
+        .filter(path => {
+          if (statSync(path).isFile()) {
+            console.error(red(`No code owner found for "${path}".`));
```
crisbeto: If you want to make this look like a Gulp task out, you can use the `gulp-util`.

tools/gulp/tasks/lint.ts
```diff
@@ -46,6 +54,56 @@ task('madge', ['material:clean-build'], () => {
...
+task('ownerslint', () => {
+  let errors = 0;
+
+  let ownedPaths = readFileSync(ownersFilePath, 'utf8').split('\n')
+      // Trim lines.
+      .map(line => line.trim())
+      // Remove empty lines and comments.
+      .filter(line => line && !line.startsWith('#'))
+      // Split off just the path glob and turn it into a regex.
+      .map(line => new Minimatch(line.split(/\s+/)[0], {dot: true, matchBase: true}))
```
crisbeto: You can avoid having to repeat the `Minimatch` constructor config by `concat`-ing the `ignoreOwners` earlier:

```ts
readFileSync(ownersFilePath, 'utf8').split('\n')
  // Trim lines.
  .map(line => line.trim())
  // Remove empty lines and comments.
  .filter(line => line && !line.startsWith('#'))
  // Take out the file pattern.
  .map(line => line.split(/\s+/)[0])
  // Add in the paths we're ignoring.
  .concat(ignoreOwners)
  // Turn the paths into regex.
  .map(line => new Minimatch(line, {dot: true, matchBase: true}))
```

tools/gulp/tasks/lint.ts
mmalerba: I think it will? it splits on white space and just cares about the first piece

tools/gulp/tasks/lint.ts
mmalerba: I was originally extracting the regex and working with that, but the `matchBase: true` didn't seem to be coded into the regex. Anyways I think it caches the regex under the hood instead of recompiling it every time, so it should be fine.

tools/gulp/tasks/lint.ts
mmalerba: I think thats the equivalent of `slice(...paths)` not `concat(...paths)`

mmalerba:
comments addressed

PR #7939, opened 2017-10-21
build: fix examples not building

crisbeto:
Fixes an error in the `material-examples` build, because the Moment adapter wasn't added as a dependency.

### DIFF
--- a/tools/gulp/packages.ts
+++ b/tools/gulp/packages.ts
@@ -3,8 +3,8 @@ import {join} from 'path';
 
 export const cdkPackage = new BuildPackage('cdk');
 export const materialPackage = new BuildPackage('material', [cdkPackage]);
-export const examplesPackage = new BuildPackage('material-examples', [materialPackage, cdkPackage]);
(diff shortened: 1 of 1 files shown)


tools/gulp/packages.ts
```diff
@@ -3,8 +3,12 @@ import {join} from 'path';
 
 export const cdkPackage = new BuildPackage('cdk');
 export const materialPackage = new BuildPackage('material', [cdkPackage]);
-export const examplesPackage = new BuildPackage('material-examples', [materialPackage, cdkPackage]);
 export const momentAdapterPackage = new BuildPackage('material-moment-adapter', [materialPackage]);
+export const examplesPackage = new BuildPackage('material-examples', [
+  materialPackage,
+  cdkPackage,
+  momentAdapterPackage
```
DevVersion: If you only add `momentAdapterPackage`, then it won't build `material` or the `cdk` a second time. The dependencies of the `momentAdapterPackage` will be installed automatically (see other packages)

crisbeto: Done.

PR #7951, opened 2017-10-22
fix(viewport-ruler): incorrectly caching viewport size

crisbeto:
* Currently the `ViewportRuler` caches the `ClientRect` of the `document.documentElement` which is only used for determining the scroll position and is being updated on demand anyway. These changes switch to caching the `innerWidth` and `innerHeight` which was the original intent.
* Adds a `getViewportSize` method that returns the cached viewport width and height. The reasoning is that for all of the cases where we were using `getViewportRect`, we were only taking the `width` and `height`. `getViewportSize` helps us avoid the overhead from determining the `top`, `bottom`, `left` and `right` via the scroll position.

### DIFF
--- a/src/cdk/scrolling/viewport-ruler.ts
+++ b/src/cdk/scrolling/viewport-ruler.ts
@@ -24,9 +24,8 @@ export const DEFAULT_RESIZE_TIME = 20;
  */
 @Injectable()
 export class ViewportRuler implements OnDestroy {
-
-  /** Cached document client rectangle. */
-  private _documentRect?: ClientRect;
+  /** Cached viewport dimensions. */
+  private _viewportSize: {width: number; height: number};
 
   /** Stream of viewport change events. */
   private _change: Observable<Event>;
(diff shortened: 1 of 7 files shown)


src/cdk/scrolling/viewport-ruler.ts
```diff
@@ -24,9 +24,8 @@ export const DEFAULT_RESIZE_TIME = 20;
  */
 @Injectable()
 export class ViewportRuler implements OnDestroy {
-
-  /** Cached document client rectangle. */
-  private _documentRect?: ClientRect;
+  /** Cached viewport dimensions. */
+  private _viewportSize: {width: number; height: number};
```
jelbourn: Does reading the window width / height actually cost anything? Do we need to cache it at all?

src/cdk/scrolling/viewport-ruler.ts
```diff
@@ -77,27 +78,20 @@ export class ViewportRuler implements OnDestroy {
...
+  /** Gets the (top, left) scroll position of the viewport. */
+  getViewportScrollPosition() {
     // The top-left-corner of the viewport is determined by the scroll position of the document
     // body, normally just (scrollLeft, scrollTop). However, Chrome and Firefox disagree about
     // whether `document.body` or `document.documentElement` is the scrolled element, so reading
     // `scrollTop` and `scrollLeft` is inconsistent. However, using the bounding rect of
     // `document.documentElement` works consistently, where the `top` and `left` values will
     // equal negative the scroll position.
-    const top = -documentRect!.top || document.body.scrollTop || window.scrollY ||
+    const documentRect = document.documentElement.getBoundingClientRect();
```
jelbourn: Not for this PR, but thinking about this again, I wonder if there's a better way to get what we're after here than calling `getBoundingClientRect` here.

src/cdk/scrolling/viewport-ruler.ts
crisbeto: It does. See https://gist.github.com/paulirish/5d52fb081b3570c81e3a

src/cdk/scrolling/viewport-ruler.ts
crisbeto: AFAIK the cross-browser way to do it was something like this: `window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop || 0`.

Issue #3522, opened 2017-03-08
md-icon docs should showcase how to use material icons easily


andrewseguin:
Most users will likely go to the md-icon docs because they are just starting out with them. The easiest way to do this is to include the material icons fontset in their index.html. While the information to do so is in the docs, it does not stand out. Right now it looks like they'll need to integrate with the MdIconRegistry to get md-icon to work.

victorschinaider:
Can i separate the icons in two 'divisions'? When i was reading the docs, in my mental map, i separate in `md-icon` in two: **native**, when i'm using the Material Icons ( then we just need to load the fontset), and in **custom**, when i want to use external icons with `md-icon`, using `MdIconRegistry`, what do you think? So i was thinking in refactor the docs in these two modes.

jelbourn:
As I see it, the split is between **font icons** and **svg icons**

victorschinaider:
Yes, but i was thinking in highlight the **native support** to material icons, because most of the users, i think, will just to use the native support in start, just to test. If we give highlight how to fast set up the `md-icons`, it'll be more effective.

Initially, we'll talk about the fast option (material icons), and after, we give details how to use "custom icons" ( **font icons** and **svg icons** ).

I'll soon open a pr.

jelbourn:
@victorschinaider I don't follow what you mean in using the word "native". Using the material icons w/ `md-icon` is the same as using any other font icon set except for the fact that we set the default icon css class based on the material one. "Native" implies that it's something inherent to the browser platform.

victorschinaider:
@jelbourn I got it, thanks. I'm going to rephrase my thoughts based on that.

M-a-c:
@victorschinaider are you going to open a pr for this?

victorschinaider:
@M-a-c, If you have something, feel free to open.
I've a material fork in my account with some modifications (https://github.com/victorschinaider/material2/blob/icon-docs/src/lib/icon/icon-temp.md), it's just a draft, 
if you fell comfortable, we can discuss.

M-a-c:
@victorschinaider Sure!
If there are parts you're not sure of let me know, we can discuss on your branch.
Also If you'd rather not do a part I can lookup/fillout the info for that section.
Let me know!

alisaduncan:
I created 2 PRs to provide a couple of different ways to point users to the correct docs and examples. Please let me know if this is not the direction you were expecting and I can update.

cpboyd:
[...]

PR #7962, opened 2017-10-22
docs(icon): change title of basic icon to font icon

alisaduncan:
a more descriptive title will help users unfamiliar with mat-icon per issue #3522.


jelbourn:
Thanks for the PR! We actually want to expand this example to have more types of icons rather than just ligature icons.

Issue #3175, opened 2017-02-17
md-tree


kara:
Needs design

leshin:
Quickly ! ! !

vyakymenko:
Hello guys, have made some simple design for this component,

![screenshot_7](https://cloud.githubusercontent.com/assets/7300673/25062037/717e04ee-21cc-11e7-95c6-c690425f24ba.png)

[PSD](https://drive.google.com/file/d/0B2xaC3SJ6rU2SmgxOUlFa1I4cUU/view?usp=sharing)

vyakymenko:
Simple Material Design for tree:

![angular io material tree view](https://cloud.githubusercontent.com/assets/7300673/25062074/502592fc-21cd-11e7-9ca5-2c7209d1eb37.jpg)

Photoshop Source [here](https://drive.google.com/file/d/0B2xaC3SJ6rU2SmgxOUlFa1I4cUU/view?usp=sharing)

vyakymenko:
**_Proposing code design:_**

**JSON Example:**
```json
{
	"data": [{
		"label": "Tree Item Collapsed",
		"data": "Tree Item Collapsed Data",
		"expandIcon": "fa-folder-open",
		"collapseIcon": "fa-folder",
		"children": []
	}, {
		"label": "Tree Item Expanded",
		"data": "Tree Item Expanded Data",
		"expandIcon": "fa-folder-open",
		"collapseIcon": "fa-folder",
		"children": [{
			"label": "Bookmarks",
			"data": "Bookmarks Data",
			"icon": "fa-bookmarks"
		}, {
			"label": "Chart",
			"data": "Chart Data",
			"expandedIcon": "fa-chart"
		}]
	}, {
		"label": "Bookmarks",
		"data": "Bookmarks Data",
		"icon": "fa-bookmarks"
	}, {
		"label": "Chart",
		"data": "Chart Data",
		"expandedIcon": "fa-chart"
	}]
}
```

**Attributes:**
`data` - Data array of treenodes
`selectionType` - Selection type, "single", "multiple", maybe "checkbox?".
`selection` - Single treenode instance or an array to refer to the selections.
`orientation` - Tree orientation.

**Events:**
`nodeExpand` - Callback to invoke when a node is expanded.
`nodeCollapse` - Callback to invoke when a node is expanded.
`nodeSelect` - Callback to invoke when a node is unselected.
`nodeUnselect` - Callback to invoke when a node is selected.

**Lazy Loading?**

[...]

PR #7984, opened 2017-10-23
feat(tree): Add Cdk Tree component

tinayuangao:
An initial version of cdk-tree

For #3175


src/cdk/collections/tree-adapter.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import {SelectionModel} from './selection';
+
+
+export interface TreeAdapter<T> {
+  flattenNodes( structuredData: any[]): T[];
+
+  expandFlattenedNodes(nodes: T[], expansionModel: SelectionModel<T>): T[];
+
+  nodeDecedents(node: T, nodes: T[], onlyExpandable: boolean);
```
jelbourn: Add JsDoc description for these methods?

src/cdk/tree/nested-node.ts
```diff
@@ -0,0 +1,85 @@
...
+} from '@angular/core';
+import {Subscription} from 'rxjs/Subscription';
+import {CdkTree} from './tree';
+import {NestedNode} from './tree-data';
+import {CdkNodePlaceholder} from './placeholder';
+import {CdkTreeNode} from './node';
+
+
+/**
+ * Nested node, add children to `cdkNodePlaceholder` in template.
```
jelbourn: Clarify that this is a child of `<cdk-tree>` and what is means to be a "nested node"? I'm also not sure exactly what ``"add children to `cdkNodePlaceholder` in template."`` means

src/cdk/tree/nested-node.ts
```diff
@@ -0,0 +1,85 @@
...
+  @ContentChildren(CdkNodePlaceholder) nodePlaceholder: QueryList<CdkNodePlaceholder>;
+
+  /** The Children nodes data. */
+  protected _children: T[];
+
+  /** Subscribes to children changes. */
+  protected _childrenSubscription: Subscription;
+
+  /** Subscribes to node placeholder changes. */
+  protected _placeholderSubscription: Subscription;
```
jelbourn: One pattern we've been using more lately is to create a subject like `_destroyed` and then use `takeUntil(this._destroyed)` to avoid needing to manually keep track of the subscription and unsubscribe

src/cdk/tree/nested-node.ts
```diff
@@ -0,0 +1,85 @@
...
+@Directive({
+  selector: '[cdkNestedNode]'
+})
+export class CdkNestedNode<T extends NestedNode> implements AfterContentInit, OnDestroy {
+
+  /** The children node placeholder. */
+  @ContentChildren(CdkNodePlaceholder) nodePlaceholder: QueryList<CdkNodePlaceholder>;
+
+  /** The Children nodes data. */
+  protected _children: T[];
```
jelbourn: I'm not very clear on what this represents exactly

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+  OnDestroy,
+  TemplateRef,
+  ViewContainerRef,
+  ViewEncapsulation
+} from '@angular/core';
+import {Subscription} from 'rxjs/Subscription';
+import {FlatNode, NestedNode} from './tree-data';
+
+/** The tree node template */
+export const CDK_TREE_NODE_TEMPLATE = '<ng-content cdkNodeOutlet></ng-content>';
```
jelbourn: Any reason not to just put this inline in the component metadata?

src/cdk/tree/nested-node.ts
```diff
@@ -0,0 +1,85 @@
...
+import {CdkTreeNode} from './node';
+
+
+/**
+ * Nested node, add children to `cdkNodePlaceholder` in template.
+ */
+@Directive({
+  selector: '[cdkNestedNode]'
+})
+export class CdkNestedNode<T extends NestedNode> implements AfterContentInit, OnDestroy {
```
jelbourn: Should this be `CdkNestedTreeNode`? I'm not totally clear on the different between this and `CdkTreeNode`

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Function that should return true if this node template should be used for the provided node
+   * data and index. If left undefined, this node will be considered the default node template to
+   * use when no other when functions return true for the data.
+   * For every node, there must be at least one when function that passes or an undefined to
+   * default.
+   */
+  when: (nodeData: T, index: number) => boolean;
+
+  constructor(public template: TemplateRef<any>) {}
```
jelbourn: Should `template` be `@docs-private`?

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+
+/**
+ * Tree node for CdkTree. It contains the data in the tree node.
+ */
+// TODO: Role should be group for expandable ndoes
```
jelbourn: Can you move the `TODO` next to where `role` is defined? When there's another comment between the JsDoc and the class definition dgeni doesn't connect them

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+  private _childrenSubscrition: Subscription;
+
+  /** The tree node data */
+  @Input('cdkNode')
+  set data(v: T) {
+    this._data = v;
+    if (this._childrenSubscrition) {
+      this._childrenSubscrition.unsubscribe();
+    }
+    if ('level' in v) {
```
jelbourn: We can't use `in` with string-based property names since Closure Compiler will rename properties

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+      this.role = (this._data as FlatNode).expandable ? 'group' : 'treeitem';
+    } else {
+      this._childrenSubscrition = (this._data as NestedNode).getChildren().subscribe((children) => {
+        this.role = !!children ? 'group' : 'treeitem';
```
jelbourn: Don't need to use `!!` in conditionals, only when returning / storing the value

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+      });
+    }
+  }
+  get data(): T { return this._data; }
+  _data: T;
+
+
+  /** The offset top of the element. Used by CdkTree to decide the order of the nodes. [Focus] */
+  get offsetTop() {
```
jelbourn: Should this be internal?

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+    return this._elementRef.nativeElement.offsetTop;
+  }
+
+  /**
+   * The role of the node should be 'group' if it's an internal node,
+   * and 'treeitem' if it's a leaf node.
+   */
+  @Input() role: string = 'treeitem';
```
jelbourn: Instead of `string`, should this be `'treeitem' | 'group'`?

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  constructor(private _elementRef: ElementRef) {}
+
+  ngOnDestroy() {
+    if (this._childrenSubscrition) {
+      this._childrenSubscrition.unsubscribe();
+    }
+  }
+
+  /** Focuses the menu item. Implements for FocusableOptoin. */
```
jelbourn: `FocusableOptoin` -> `FocusableOption`

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,40 @@
...
+import {FlatNode} from './tree-data';
+
+/**
+ * Indent for the children tree nodes.
+ * This directive will add left-padding to the node to show hierarchy.
+ */
+@Directive({
+  selector: '[cdkNodePadding]',
+  host: {
+    '[style.padding-left]': 'paddingIndent',
```
jelbourn: Need to account for RTL?

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+})
+export class CdkNodePadding<T extends FlatNode> {
+  /** The level of depth of the tree node. The padding will be `level * indent` pixels. */
+  @Input('cdkNodePadding') level: number;
+
+  /** The indent for each level. Default value is 28px. */
+  @Input('cdkNodePaddingIndex') indent: number = 28;
```
jelbourn: Any background on where `28px` came from?

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  get paddingIndent() {
```
jelbourn: Can this be a method instead of a getter?

src/cdk/tree/padding.ts
jelbourn: This is marked as an `Input` but it has no setter. Also is `Index` the right term?

src/cdk/tree/placeholder.ts
```diff
@@ -0,0 +1,23 @@
...
+
+
+/**
+ * Placeholder for nested CdkNode. Put `[cdkNodePlaceholder]` on a tag to place children nodes
+ * inside the placeholder.
+ */
+@Directive({
+  selector: '[cdkNodePlaceholder]'
+})
+export class CdkNodePlaceholder {
```
jelbourn: `Placeholder` -> `Outlet`?

src/cdk/tree/select-trigger.ts
```diff
@@ -0,0 +1,50 @@
...
+  @Input('cdkNodeSelectTriggerRecursive') recursive: boolean = false;
+
+  get node() {
+    return this.treeNode.data;
+  }
+
+  constructor(@Inject(forwardRef(() => CdkTree)) private tree: CdkTree<T>,
+              private treeNode: CdkTreeNode<T>) {}
+
+  trigger(_: Event) {
```
jelbourn: Can you not omit the event argument?

src/cdk/tree/select-trigger.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Node selection trigger. Clicking on the trigger will change the selection.
+ */
+@Directive({
+  selector: '[cdkNodeSelectTrigger]',
+  host: {
+    '(click)': 'trigger($event)',
+  }
+})
+export class CdkNodeSelectTrigger<T> {
```
jelbourn: I don't quite remember why this is baked into the tree- what's the background here?

src/cdk/tree/tree-control.ts
```diff
@@ -0,0 +1,223 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {SelectionModel} from '@angular/cdk/collections';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+import {FlatNode, NestedNode} from './tree-data';
+import {CdkTreeNode} from './node';
+
+/**
+ * Tree control interface
```
jelbourn: Can you expand this comment a bit to mention how you would use this interface?

src/cdk/tree/tree-control.ts
```diff
@@ -0,0 +1,223 @@
...
+ */
+export interface TreeControl {
```
jelbourn: Should this be generic to avoid using `any`?

src/cdk/tree/tree-control.ts
```diff
@@ -0,0 +1,223 @@
...
+  expandChange: BehaviorSubject<any>;
+
+  /** The expansion model */
+  expansionModel: SelectionModel<any>;
+
+  /** Whether the node is expanded or collapsed. Return true if it's expanded. */
+  expanded(node: any): boolean;
+
+  /** Get all decedents of a node */
+  getDecedents(node: any): any[];
```
jelbourn: `Decedents` -> `Descendants` here and elsewhere

src/cdk/tree/tree-control.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  /** Expand a node and all its decedents */
+  expandDecedents(node: any);
+
+  /** Collapse a ndoe and all its decedents */
+  collapseDecedents(node: any);
+}
+
+/** Base tree control. It has basic toggle/expand/collapse operations on a single node. */
+export abstract class BaseTreeControl<T extends FlatNode|NestedNode> implements TreeControl {
```
jelbourn: What do you think of making a subdirectory for controls with separate files for `tree-control.ts`, `base-tree-control.ts`, `flat-tree-control.ts` and `nested-tree-control.ts`?

src/cdk/tree/tree-control.ts
```diff
@@ -0,0 +1,223 @@
...
+   */
+  expandAll() {
+    this.expansionModel.clear();
+    this.nodes.forEach((node) => node.expandable && this.expansionModel.select(node));
+    this.expandChange.next(this.expansionModel.selected);
+  }
+
+  /** Expands a subtree rooted at given `node` recursively. */
+  expandDecedents(node: T) {
+    let decedents = this.getDecedents(node);
```
jelbourn: Prefer `const` (here and elsewhere)

src/cdk/tree/tree-control.ts
```diff
@@ -0,0 +1,223 @@
...
+/** Nested tree control. Able to expand/collapse a subtree recursively for NestedNode type. */
+export class NestedTreeControl<T extends NestedNode> extends BaseTreeControl<T> {
+  /**
+   * Expands all nodes in the tree.
+   *
+   * To make this working, the `nodes` of the TreeControl must be set correctly.
+   */
+  expandAll() {
+    this.expansionModel.clear();
+    this.nodes.forEach((node) => this._expandDecedents(node));
```
jelbourn: Nit: don't need parens for a single argument with inferred type (here and elsewhere)

src/cdk/tree/tree-control.ts
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  /** Collapses a subtree rooted at given `node` recursively. */
+  collapseDecedents(node: T) {
+    this.expansionModel.deselect(node);
+    const subscription = node.getChildren().subscribe((children: T[]) => {
+      if (children) {
+        children.forEach((child) => {
+          this.collapseDecedents(child)
+        });
```
jelbourn: Don't need to use braces here
```ts
children.forEach(child => this.collapseDecedents(child));
```
(similar elsewhere)

src/cdk/tree/tree-data.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Observable} from 'rxjs/Observable';
+
+/** Interface for data in flat tree. Tree users should implement this interface to use CdkTree. */
+export interface FlatNode {
+  level: number;
```
jelbourn: Should we do a `getLevel` method instead of a property? It might be annoying if someone has a `level` property in their data that means something else. (Same for `expandable` -> `isExpandable`). 

I also wonder if there's a way to could express this in the template (similar to `when`) instead of requiring an interface, but I forget if we already explored that option

src/cdk/tree/tree.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+  let treeElement: HTMLElement;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [CdkTreeModule],
+      declarations: [
+        SimpleCdkTreeApp,
+        // DynamicDataSourceCdkTreeApp,
+        // NodeContextCdkTreeApp,
+        // WhenNodeCdkTreeApp
```
jelbourn: Add `TODO: ` in front of each of these? That will change them from commented-out-code to just comments

src/cdk/tree/select-trigger.ts
tinayuangao: This can be used to change expand/collapse a tree node/subtree with a custom SelectionModel(a more general trigger for CdkNodeTrigger). Removed since it should not be included for the initial version.

src/cdk/tree/tree-data.ts
tinayuangao: We didn't explore that option. 

I can try to express it in the template, but `TreeControl` is also using that to expand/collapse subtrees/the whole tree.

If we want to express this in the template, we can also pass that info to `TreeControl`, or put `getLevel` `getChildren` and `isExpandable` functions in `TreeControl` and let other components use functions in `TreeControl`.

tinayuangao:
Comments addressed. Please take a look. Thanks!

src/cdk/tree/tree-data.ts
tinayuangao: Now moved `getLevel` `isExpandable` and `getChildren` to `TreeControl`. Removed `NestedNode` and `FlatNode` types and users can use any type of data.

src/cdk/collections/tree-adapter.ts
```diff
@@ -0,0 +1,28 @@
...
+import {SelectionModel} from './selection';
+
+
+/**
+ * Tree Adapter to change any structured data type to flattened tree data.
+ * To be implemented by tree users.
+ */
+export interface TreeAdapter<T> {
+  /** Flatten structured data to an array of data. */
+  flattenNodes( structuredData: any[]): T[];
```
jelbourn: Extra space

src/cdk/collections/tree-adapter.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  /** Expand flattened target nodes with expansion model. */
+  expandFlattenedNodes(nodes: T[], expansionModel: SelectionModel<T>): T[];
+
+  /**
+   * Put node descendents of node in array.
+   * If `onlyExpandable` is true, then only process expandable descendents.
+   */
+  nodeDecedents(node: T, nodes: T[], onlyExpandable: boolean);
```
jelbourn: "Decedents" -> "Descendants"
(here and elsewhere)

src/cdk/collections/tree-adapter.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * @license
+ * Copyright Google LLC. All Rights Reserved.
```
jelbourn: Extra `.` after LLC

src/cdk/tree/control/base-tree-control.ts
```diff
@@ -0,0 +1,80 @@
...
+import {Observable} from 'rxjs/Observable';
+import {TreeControl} from './tree-control';
+
+/** Base tree control. It has basic toggle/expand/collapse operations on a single node. */
+export abstract class BaseTreeControl<T> implements TreeControl<T> {
+  /** Saved node for `expandAll` action. */
+  nodes: T[];
+
+  /** Expansion info: the changes */
+  expandChange = new BehaviorSubject<T[]>([]);
```
jelbourn: Maybe `expansionChange`? @andrewseguin thoughts?

src/cdk/tree/control/base-tree-control.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /** Collapses one single node. */
+  collapse(node: T) {
+    this.expansionModel.deselect(node);
+    this.expandChange.next(this.expansionModel.selected);
+  }
+
+  /** Whether a given node is expanded or not. Returns true if the node is expanded. */
+  expanded(node: T) {
```
jelbourn: `isExpanded`?

src/cdk/tree/control/base-tree-control.ts
```diff
@@ -0,0 +1,80 @@
...
+  collapseAll() {
+    this.expansionModel.clear();
+    this.expandChange.next(this.expansionModel.selected);
+  }
+
+  /** Expands all nodes in the tree. */
+  abstract expandAll();
+
+  /** Expands a subtree rooted at given `node` recursively. */
+  abstract expandDescendents(node: T);
```
jelbourn: "Descendents" -> "Descendants"
(here and elsewhere)

src/cdk/tree/control/base-tree-control.ts
```diff
@@ -0,0 +1,80 @@
...
+  /** Gets a list of descendent nodes of a subtree rooted at given `node` recursively. */
+  abstract getDescendents(node: T): T[];
+
+  /** Get depth of a given node, return the level number. This is for flat tree node. */
+  getLevel: (node: T) => number;
+
+  /** Whether the node is expandable. Returns true if expandable. This is for flat tree node. */
+  isExpandable: (node: T) => boolean;
+
+  /** Returns an observable for a given node's children. This is for nested tree node. */
```
jelbourn: "Gets a stream that emits whenever the given node's children change"

Is that accurate?

src/cdk/tree/control/tree-control.ts
```diff
@@ -0,0 +1,65 @@
...
+import {SelectionModel} from '@angular/cdk/collections';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+import {Observable} from 'rxjs/Observable';
+
+/**
+ * Tree control interface. User can implement TreeControl to expand/collapse nodes in the tree.
+ * The CDKTree will use this TreeControl to expand/collapse a node.
+ * User can also use it outside the `<cdk-tree>` to control the expansion status of the tree.
+ */
+export interface TreeControl<T> {
```
jelbourn: Unit tests for each type of TreeControl?

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CdkTreeNode<T>  implements FocusableOption, OnDestroy {
+  /** Emits when the component is destroyed. */
+  private _destroyed = new Subject<void>();
+
+  /** The tree node data */
+  @Input('cdkTreeNode')
+  set data(v: T) {
+    this._data = v;
```
jelbourn: Break the content of this setter into a separate function?

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+        throw getTreeControlFunctionsMissingError();
+      }
+      takeUntil.call(this._tree.treeControl.getChildren(this._data), this._destroyed)
+        .subscribe(children => {
+          this.role = children ? 'group' : 'treeitem';
+        });
+    }
+  }
+  get data(): T { return this._data; }
+  _data: T;
```
jelbourn: Make this private?

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /** The offset top of the element. Used by CdkTree to decide the order of the nodes. [Focus] */
+  private get offsetTop() {
```
jelbourn: This this is `private`, make it a method?

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,54 @@
...
+  paddingIndentLeft() {
+    return this._dir && this._dir.value === 'rtl' ? null : this._paddingIndent;
+  }
+
+  /** The right padding indent value for the tree node. */
+  paddingIndentRight() {
+    return this._dir && this._dir.value === 'rtl' ? this._paddingIndent() : null;
+  }
+
+  constructor(private _treeNode: CdkTreeNode<T>,
```
jelbourn: Move constructor above methods and below properties

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,54 @@
...
+              private _tree: CdkTree<T>,
+              @Optional() private _dir: Directionality) {}
```
jelbourn: Probably want to have `dir.change` as well

src/cdk/tree/trigger.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Directive({
+  selector: '[cdkNodeTrigger]',
+  host: {
+    '(click)': '_trigger($event)',
+  }
+})
+export class CdkNodeTrigger<T> {
+  /** Whether expand/collapse the node recursively. */
+  @Input('cdkNodeTriggerRecursive') recursive: boolean = true;
```
jelbourn: `coerceBooleanProperty`?

src/cdk/tree/trigger.ts
```diff
@@ -0,0 +1,39 @@
...
+  Input,
+} from '@angular/core';
+import {CdkTree} from './tree';
+import {CdkTreeNode} from './node';
+
+/**
+ * Node trigger to expand/collapse the node.
+ */
+@Directive({
+  selector: '[cdkNodeTrigger]',
```
jelbourn: `cdkNodeTrigger` doesn't seem quite specific enough. Maybe `cdkTreeNodeToggle`?

src/cdk/tree/tree.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+import {combineLatest} from 'rxjs/observable/combineLatest';
+import {map} from 'rxjs/operator/map';
+
+import {TreeControl} from './control/tree-control';
+import {FlatTreeControl} from './control/flat-tree-control';
+import {CdkTreeModule} from './index';
+import {CdkTree} from './tree';
+
+
+describe('CdkTree', () => {
```
jelbourn: In a follow-up PR, it would be good to explore creating a syntax-sugar way of testing tree content. The table, for example, has this:
```ts
      expectTableToMatchContent(tableElement, [
        ['Column A', 'Column B', 'Column C'],
        ['a_1', 'b_1', 'c_1'],
        ['a_2', 'b_2', 'c_2'],
        ['a_3', 'b_3', 'c_3'],
      ]);
```

For the tree, we'd need to capture level and expansion state

src/cdk/tree/tree-module.ts
```diff
@@ -0,0 +1,35 @@
...
+import {CommonModule} from '@angular/common';
+import {NgModule} from '@angular/core';
+import {NodeOutlet} from './outlet';
+import {CdkNodePadding} from './padding';
+import {CdkNodeTrigger} from './trigger';
+import {CdkTree} from './tree';
+import {CdkNodeDef, CdkTreeNode} from './node';
+import {CdkNestedTreeNode} from './nested-node';
+
+const EXPORTED_DECLARATIONS = [
```
jelbourn: Maybe just call this `TREE_DIRECTIVES`?

src/cdk/tree/control/base-tree-control.ts
```diff
@@ -0,0 +1,80 @@
...
+  nodes: T[];
+
+  /** Expansion info: the changes */
+  expansionChange = new BehaviorSubject<T[]>([]);
+
+  /** A selection model with multi-selection to track expansion status. */
+  expansionModel: SelectionModel<T> = new SelectionModel<T>(true);
+
+  /** Toggles one single node. Expands a collapsed node or collapse an expanded node. */
+  toggle(node: T) {
```
crisbeto: All of these public methods need return types. Same goes for the other files.

src/cdk/tree/control/base-tree-control.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.expansionModel.toggle(node);
+    this.expansionChange.next(this.expansionModel.selected);
+  }
+
+  /** Expands one single node. */
+  expand(node: T) {
+    this.expansionModel.select(node);
+    this.expansionChange.next(this.expansionModel.selected);
```
crisbeto: You should be able to replace the `expansionChange` event with the `onChange` from the `SelectionModel`.

src/cdk/tree/control/nested-tree-control.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Expands all nodes in the tree.
+   *
+   * To make this working, the `nodes` of the TreeControl must be set correctly.
+   */
+  expandAll() {
+    this.expansionModel.clear();
+    this.nodes.forEach((node) => this._expandDescendants(node));
```
crisbeto: nit: you don't need the parentheses around `node`.

src/cdk/tree/control/nested-tree-control.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /** A helper function to get descendants recursively. */
+  protected _getDescendants(descendants: T[], node: T) {
+    descendants.push(node);
+    const subscription = this.getChildren(node).subscribe(children => {
+      if (children) {
+        children.forEach((child: T) => this._getDescendants(descendants, child));
+      }
+    });
+    subscription.unsubscribe();
```
crisbeto: Rather than unsubscribing manually here, you could use the `first` operators instead.

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,116 @@
...
+  private _data: T;
+
+  private setRoleFromData() {
+    if (this._tree.treeControl.isExpandable) {
+      this.role = this._tree.treeControl.isExpandable(this._data) ? 'group' : 'treeitem';
+    } else {
+      if (!this._tree.treeControl.getChildren) {
+        throw getTreeControlFunctionsMissingError();
+      }
+      takeUntil.call(this._tree.treeControl.getChildren(this._data), this._destroyed)
```
crisbeto: Since this subscription is set up by the setter, shouldn't it unsubscribe from any previous subscriptions? Otherwise you could end up with multiple subscriptions if the `data` is set multiple times.

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,116 @@
...
+        .subscribe(children => {
+          this.role = children ? 'group' : 'treeitem';
+        });
+    }
+  }
+
+  /** The offset top of the element. Used by CdkTree to decide the order of the nodes. [Focus] */
+  private getOffsetTop() {
```
crisbeto: I don't see this method being used anywhere, is it still relevant?

src/cdk/tree/tree.ts
```diff
@@ -0,0 +1,272 @@
...
+    this._defaultNodeDef = defaultNodeDefs[0];
+
+    if (this.dataSource) {
+      this._observeRenderChanges();
+    }
+  }
+
+  ngAfterViewInit() {
+    // For key traversal in correct order
+    this.items.changes.subscribe(items => {
```
crisbeto: It seems like you're never unsubscribing from this one and I'm not sure whether Angular will complete the stream on destroy.

src/cdk/tree/tree.ts
```diff
@@ -0,0 +1,272 @@
...
+      const nodes = items.toArray();
+
+      nodes.sort((a, b) => a.offsetTop - b.offsetTop);
+      this.orderedNodes.reset(nodes);
+
+      const activeItem = this._keyManager ? this._keyManager.activeItem : null;
+      this._keyManager = new FocusKeyManager(this.orderedNodes);
```
crisbeto: Since you're mutating the same `QueryList`, you shouldn't need to re-create the key manager every time the items change. You may have to reset the active index though.

src/cdk/tree/tree.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  /** Keyboard actions. */
+  // TODO(tinayuangao): Work on keyboard traversal and actions, make sure it's working for RTL
+  //     and nested trees.
+  handleKeydown(event) {
+    if (event.keyCode == UP_ARROW) {
+      this._keyManager.setPreviousItemActive();
+    } else if (event.keyCode == DOWN_ARROW) {
+      this._keyManager.setNextItemActive();
```
crisbeto: The UP_ARROW and DOWN_ARROW behavior should be handled by the key manager already.

src/cdk/tree/tree.ts
tinayuangao: Reset it because the order of the items in the `QueryList` is wrong. For example:
- node 1
- node 2
After expanded node 1, it becomes 
- node 1
---- node 3
---- node 4
-node 2
in item list, and UP_ARROW and DOWN_ARROW will jump from one node to another.
Resetting the nodes make it become
- node 1
---- node 2
---- node 3
- node 4
in key manager.

src/cdk/tree/node.ts
tinayuangao: Removed.

src/cdk/tree/tree.ts
tinayuangao: Removed key manager related part.

tinayuangao:
- Added tests for flat-tree-control and nested-tree-control.
- Removed `cdkTreeNode` input from `CdkTreeNode` and put the data through static variable. 
- Removed key manager related part.

Please take a look. Thanks for review!

src/cdk/tree/nested-node.ts
```diff
@@ -0,0 +1,91 @@
...
+  ngAfterContentInit() {
+    takeUntil.call(this.tree.treeControl.getChildren(this.treeNode.data), this._destroyed)
+      .subscribe(result => {
+        // In case when nodePlacholder is not in the DOM when children changes, save it in the node
+        // and add to nodeOutlet when it's available.
+        this._children = result as T[];
+        this._addChildrenNodes();
+      });
+    takeUntil.call(this.nodeOutlet.changes, this._destroyed)
+      .subscribe((_) => this._addChildrenNodes());
```
crisbeto: The `_` param here is unnecessary. We may have a lint rule for unused parameters that would complain about this.

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Function that should return true if this node template should be used for the provided node
+   * data and index. If left undefined, this node will be considered the default node template to
+   * use when no other when functions return true for the data.
+   * For every node, there must be at least one when function that passes or an undefined to
+   * default.
+   */
+  when: (nodeData: T, index: number) => boolean;
+
+  /** @docs-private */
```
crisbeto: You should be able to set this on the directive doc, rather than the constructor.

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,54 @@
...
+  _paddingIndent(): string|null {
+    const nodeLevel = (this._treeNode.data && this._tree.treeControl.getLevel)
+      ? this._tree.treeControl.getLevel(this._treeNode.data)
+      : null;
+    const level = this.level || nodeLevel;
+    return level ? `${level * this.indent}px` : null;
+  }
+
+  /** The left padding indent value for the tree node. */
+  paddingIndentLeft(): string|null {
```
crisbeto: Rather than having two properties and bindings for left and right, you could set the `padding-left`/`padding-right` instead. E.g.:

```
setPadding() {
  const padding = this._paddingIndent();
  const paddingProp = this._dir && this._dir.value === 'rtl' ? 'padding-right' : 'padding-left';
  
  this._renderer.setStyle(this._element.nativeElement, paddingProp, padding);
}
```

tinayuangao: Updated `padding.ts`. Now set the properties and listen to `dir.change`.

src/cdk/collections/tree-adapter.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import {SelectionModel} from './selection';
+
+
+/**
+ * Tree Adapter to change any structured data type to flattened tree data.
+ * To be implemented by tree users.
+ */
+export interface TreeAdapter<T> {
```
jelbourn: I'm wondering if a more specific name might be useful here. Maybe something like `TreeFlattener` or `FlatTreeAdapter`. Thoughts?

src/cdk/tree/control/flat-tree-control.ts
```diff
@@ -0,0 +1,43 @@
...
+    for (; i < this.nodes.length && this.getLevel(node) < this.getLevel(this.nodes[i]); i++) {
+      results.push(this.nodes[i]);
+    }
+    return results;
+  }
+
+  /**
+   * Expands all nodes in the tree.
+   *
+   * To make this working, the `nodes` of the TreeControl must be set correctly.
```
jelbourn: It's not clear what "the `nodes` of the TreeControl must be set correctly" means

src/cdk/tree/tree.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+          `${data[2].a} - ${data[2].b} + ${data[2].c}`,
+          `${data[3].a} - ${data[3].b} + ${data[3].c}`
+        ],
+        [1, 1, 1, 2]);
+    });
+  });
+});
+
+export class TestData {
+  a: string;
```
jelbourn: Make up fake properties for this? I'm a fan of _obviously fake_ testing data like `ponyName` or `pizzaTopping` since it  is clearly distinguished from actual component APIs

src/cdk/tree/tree.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+class FakeDataSource extends DataSource<TestData> {
+  isConnected = false;
+
+  _dataChange = new BehaviorSubject<TestData[]>([]);
+  set data(data: TestData[]) { this._dataChange.next(data); }
+  get data() { return this._dataChange.getValue(); }
+
+  constructor() {
+    super();
+    for (let i = 0; i < 3; i++) { this.addData(); }
```
jelbourn: `for` body should go on its own line

src/cdk/tree/tree.ts
```diff
@@ -0,0 +1,216 @@
...
+  encapsulation: ViewEncapsulation.None,
+  preserveWhitespaces: false,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CdkTree<T> implements CollectionViewer, OnInit, OnDestroy {
+  /** Subject that emits when the component has been destroyed. */
+  private _destroyed = new Subject<void>();
+
+  /** Latest data provided by the data source through the connect interface. */
+  private _data: Array<T> = [];
```
jelbourn: `T[]`

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,123 @@
...
+import {takeUntil} from 'rxjs/operator/takeUntil';
+import {Subject} from 'rxjs/Subject';
+import {CdkTree} from './tree';
+import {getTreeControlFunctionsMissingError} from './tree-errors';
+
+
+/** Context provided to the ndoes */
+export class CdkTreeNodeOutletContext<T> {
+
+  static mostRecentContextData: any;
```
jelbourn: This is slightly different from what I had in mind. My proposal had been that immediately after calling `createEmbeddedView`, you get the most recent node directive and set the data to it. I think this would be slightly less brittle, since its more reliable than depending on the order of instantiation of the context objects. 

For whatever we end up doing, we should have some very descriptive comments explaining the situation.

src/cdk/tree/tree.ts
```diff
@@ -0,0 +1,216 @@
...
+    if (this._dataSource !== dataSource) {
+      this._switchDataSource(dataSource);
+    }
+  }
+  private _dataSource: DataSource<T>;
+
+  /** The tree controller */
+  @Input() treeControl: TreeControl<T>;
+
+  // TODO(andrewseguin): Remove max value as the end index
```
jelbourn: Change `TODO` to you?

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,123 @@
...
+  OnDestroy,
+  TemplateRef
+} from '@angular/core';
+import {takeUntil} from 'rxjs/operator/takeUntil';
+import {Subject} from 'rxjs/Subject';
+import {CdkTree} from './tree';
+import {getTreeControlFunctionsMissingError} from './tree-errors';
+
+
+/** Context provided to the ndoes */
```
jelbourn: ndoes

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+  _level: number;
+
+  /** The indent for each level. Default number 40px from material design menu sub-menu spec. */
+  @Input('cdkTreeNodePaddingIndent')
+  set indent(value: number) {
+    this._indent = value;
+    this._setPadding();
+  }
+  _indent: number = 40;
```
jelbourn: Should this support people setting a string with units? E.g. `10em`. 

(could be in a follow-up PR, but maybe add a TODO)

src/cdk/tree/padding.ts
```diff
@@ -0,0 +1,74 @@
...
+  selector: '[cdkTreeNodePadding]',
+})
+export class CdkTreeNodePadding<T> implements OnDestroy {
+  /** Subject that emits when the component has been destroyed. */
+  private _destroyed = new Subject<void>();
+
+  /** The level of depth of the tree node. The padding will be `level * indent` pixels. */
+  @Input('cdkTreeNodePadding')
+  set level(value: number) {
+    this._level = value;
```
jelbourn: `coerceNumberProperty`?

src/cdk/tree/nested-node.ts
```diff
@@ -0,0 +1,91 @@
...
+ * For example:
+ *   <cdk-tree-node cdkNestedTreeNode [cdkNode]="node">
+ *     tree node data: {{node.name}}
+ *     <ng-template nodeOutlet> </ng-template>
+ *   </cdk-tree-node>
+ * The children of node will be automatically added to `cdkNodeOutlet`, the result dom will be like
+ * this:
+ *   <cdk-tree-node cdkNestedTreeNode [cdkNode]="node">
+ *     tree node data: {{node.name}}
+ *     <ng-template cdkNodeOutlet>
```
jelbourn: Should remove the `ng-template` element from the output in this example

src/cdk/tree/nested-node.ts
```diff
@@ -0,0 +1,91 @@
...
+ *       <cdk-tree-node cdKNestedTreeNode [cdkNode]="child1"></cdk-tree-node>
+ *       <cdk-tree-node cdKNestedTreeNode [cdkNode]="child2"></cdk-tree-node>
+ *     </ng-template>
+ *   </cdk-tree-node>
+ */
+@Directive({
+  selector: '[cdkNestedTreeNode]'
+})
+export class CdkNestedTreeNode<T> implements AfterContentInit, OnDestroy {
```
jelbourn: `CdkNestedTreeNode` isn't template-ized the same way that flat tree nodes are now. Does that limit us in the future if we want to change up the DOM structure of the nested tree?

src/cdk/tree/outlet.ts
```diff
@@ -0,0 +1,22 @@
...
+  Directive,
+  ViewContainerRef,
+} from '@angular/core';
+
+/**
+ * Outlet for nested CdkNode. Put `[cdkNodeOutlet]` on a tag to place children nodes
+ * inside the outlet.
+ */
+@Directive({
+  selector: '[treeNodeOutlet]'
```
jelbourn: `cdkTreeNodeOutlet`?

src/cdk/tree/tree.ts
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * CDK tree component that connects with a data source to retrieve data of type `T` and renders
+ * nodes with hierarchy. Updates the nodes when new data is provided by the data source.
+ */
+@Component({
+  moduleId: module.id,
+  selector: 'cdk-tree',
+  exportAs: 'cdkTree',
+  template: `<ng-container treeNodeOutlet></ng-container>`,
```
jelbourn: Does this PR have the handling for nested tree nodes? It looks like they'd just be dropped as-is.

src/cdk/collections/tree-adapter.ts
tinayuangao: `TreeDataNodeFlattener`?

src/cdk/tree/node.ts
tinayuangao: Now we use a static variable in `CdkTreeNode` and set data to that `mostRecentTreeNode`.

src/cdk/tree/tree.ts
tinayuangao: Yes. The nested node handling will be done in `CdkNestedTreeNode` directive.

src/cdk/tree/nested-node.ts
tinayuangao: Now `CdkNestedTreeNode` extends `CdkTreeNode`. Like `CdkTreeNode`, users can put anything in `<cdk-nested-tree-node>` as template, but the children nodes will be added to `[cdkTreeNodeOutlet]` inside the nested node template. If there's no `cdkTreeNodeOutlet`, it works the same way as `CdkTreeNode`.

tinayuangao:
@jelbourn take a look when you have time? Thanks!

jelbourn:
@tinayuangao up to you whether you want @andrewseguin to look before merging

src/cdk/tree/control/base-tree-control.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /** Collapses a subtree rooted at given data node recursively. */
+  collapseDescendants(dataNode: T): void {
+    let toBeProcessed = [dataNode];
+    toBeProcessed.push(...this.getDescendants(dataNode));
+    this.expansionModel.deselect(...toBeProcessed);
+  }
+
+  /** Gets a list of descendent data nodes of a subtree rooted at given data node recursively. */
```
andrewseguin: This is subjective, but it'd make sense to me for abstract methods to be nearer to the top to make it clear what needs to be defined when I extend this class.

src/cdk/tree/control/flat-tree-control.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+  let isExpandable = (node: TestData) => node.children && node.children.length > 0;
+
+  beforeEach(() => {
+    treeControl = new FlatTreeControl<TestData>(getLevel, isExpandable);
+  });
+
+  describe('base tree control actions', () => {
+    it('should be able to expand and collapse dataNodes', () => {
+      const nodes = generateData(10, 4);
+      const node = nodes[1];
```
andrewseguin: rename to `secondNode` to be consistent with `sixthNode`

src/cdk/tree/control/flat-tree-control.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      expect(treeControl.isExpanded(node)).toBeTruthy('Expect second node to be expanded');
+      expect(treeControl.expansionModel.selected)
+          .toContain(node, 'Expect second node in expansionModel');
+      expect(treeControl.expansionModel.selected.length)
+          .toBe(1, 'Expect only second node in expansionModel');
+
+      treeControl.toggle(sixthNode);
+
+      expect(treeControl.isExpanded(node)).toBeTruthy('Expect second node to stay expanded');
```
andrewseguin: Add test to check that `isExpanded` is true for the sixthNode?

`expect(treeControl.isExpanded(sixthNode)).toBeTruthy('Expect sixth node to be expanded');`

src/cdk/tree/control/flat-tree-control.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {BaseTreeControl} from './base-tree-control';
+
+/** Flat tree control. Able to expand/collapse a subtree recursively for FlatNode type. */
```
andrewseguin: What is `FlatNode type`?

src/cdk/tree/control/flat-tree-control.ts
```diff
@@ -0,0 +1,47 @@
...
+export class FlatTreeControl<T> extends BaseTreeControl<T> {
+
+  /** Construct with flat tree data node functions getLevel and isExpandable. */
+  constructor(public getLevel: (dataNode: T) => number,
+              public isExpandable: (dataNode: T) => boolean) {
+    super();
+  }
+
+  /**
+   * Gets a list of descendent data nodes of a subtree rooted at given data node recursively.
```
andrewseguin: "Gets a list of the data node's subtree of descendent data nodes" (seems like its a given that its recursive?). What does it mean that the data nodes must be "set correctly"?

src/cdk/tree/control/flat-tree-control.ts
```diff
@@ -0,0 +1,47 @@
...
+   *
+   * To make this working, the `dataNodes` of the TreeControl must be set correctly.
+   */
+  getDescendants(dataNode: T): T[] {
+    const startIndex = this.dataNodes.indexOf(dataNode);
+    const results: T[] = [];
+    let i = startIndex + 1;
+
+    for (;
```
andrewseguin: Why is the for loop's initialization outside of the definition (`let i = startIndex + 1`)?

src/cdk/tree/control/flat-tree-control.ts
```diff
@@ -0,0 +1,47 @@
...
+        i < this.dataNodes.length && this.getLevel(dataNode) < this.getLevel(this.dataNodes[i]);
```
andrewseguin: Is it possible to make this a function that can be documented? Would be nice to isolate the idea of this check and explain what it does

src/cdk/tree/nested-node.ts
andrewseguin: Looks like this is still here

src/cdk/tree/node.ts
```diff
@@ -0,0 +1,127 @@
...
+})
+export class CdkTreeNodeDef<T> {
+  /**
+   * Function that should return true if this node template should be used for the provided node
+   * data and index. If left undefined, this node will be considered the default node template to
+   * use when no other when functions return true for the data.
+   * For every node, there must be at least one when function that passes or an undefined to
+   * default.
+   */
+  when: (nodeData: T, index: number) => boolean;
```
andrewseguin: Switch the params to match table and ngFor's trackby

src/cdk/tree/tree-errors.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * Returns an error to be thrown when there are multiple nodes that are missing a when function.
+ * @docs-private
+ */
+export function getTreeMultipleDefaultNodeDefsError() {
+  return Error(`cdk-tree: There can only be one default row without a when predicate function.`);
```
andrewseguin: Heads up that I recently removed the prefix `cdk-table` from the errors since I realized it was clear that it was being thrown from the table and it didn't make sense for users of `mat-table` to see `cdk-table`

src/cdk/tree/tree.ts
```diff
@@ -0,0 +1,222 @@
...
+   * Can be used by the data source to as a heuristic of what data should be provided.
+   */
+  viewChange =
+    new BehaviorSubject<{start: number, end: number}>({start: 0, end: Number.MAX_VALUE});
+
+  constructor(private _differs: IterableDiffers,
+              private _changeDetectorRef: ChangeDetectorRef) {}
+
+  ngOnInit() {
+
```
andrewseguin: extra line here

src/cdk/tree/control/flat-tree-control.ts
tinayuangao: Must be flattened tree nodes

src/cdk/tree/control/flat-tree-control.ts
tinayuangao: `FlatNode` type is deleted. Updated to the doc.

src/cdk/collections/tree-adapter.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export interface TreeDataNodeFlattener<T> {
+  /** Flatten structured data to an array of data. */
+  flattenNodes(structuredData: any[]): T[];
+
+  /** Expand flattened target dataNodes with expansion model. */
+  expandFlattenedNodes(nodes: T[], expansionModel: SelectionModel<T>): T[];
+
+  /**
+   * Put node descendants of node in array.
```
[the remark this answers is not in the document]
tinayuangao: Deleted this part

tinayuangao:

src/cdk/tree/control/flat-tree-control.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+      expect(treeControl.isExpanded(sixthNode))
+          .toBeTruthy('Expect sixth node to be expanded');
+      expect(treeControl.expansionModel.selected)
+          .toContain(sixthNode, 'Expect sixth node in expansionModel');
+      expect(treeControl.expansionModel.selected)
+          .toContain(secondNode, 'Expect second node in expansionModel');
+      expect(treeControl.expansionModel.selected.length)
+          .toBe(2, 'Expect two dataNodes in expansionModel');
+
+      treeControl.collapse(seconNode);
```
timeurg: typo seconNode -> secondNode

PR #7988, opened 2017-10-24
feat(sort): if provider exists provider one, otherwise new one

Jefiozie:
As discussed in #7895 here is the PR that will implement the same pattern for the `MatSortHeaderIntl`.

cc: @jelbourn

### DIFF
--- a/src/lib/sort/sort-header-intl.ts
+++ b/src/lib/sort/sort-header-intl.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Injectable} from '@angular/core';
+import {Injectable, SkipSelf, Optional} from '@angular/core';
 import {Subject} from 'rxjs/Subject';
 import {SortDirection} from './sort-direction';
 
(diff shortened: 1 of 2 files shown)


src/lib/sort/sort-header-intl.ts
```diff
@@ -32,3 +32,16 @@ export class MatSortHeaderIntl {
     return `Sorted by ${id} ${direction == 'asc' ? 'ascending' : 'descending'}`;
   }
 }
+/** @docs-private */
+export function MAT_SORT_HEADER_INTL_PROVIDER_FACTORY(parentIntl: MatSortHeaderIntl) {
+  return parentIntl || new MatSortHeaderIntl();
+}
+
+/** @docs-private */
+export const MAT_SORT_HEADER_INTL_PROVIDER = {
```
jelbourn: Provide this in `MatSortModule`?

Jefiozie: Sorry, forgot to commit it here it is!

PR #7990, opened 2017-10-24
introduce onOpen and onClose Events to Datepicker 

tobias74:
In the select component there is a similar mechanism to emit events on opening and closing of the overlay. In my use-case I needed this functionality in both overlays: the select-dropdown and the datepicker. So I thought it might be a good idea to propose a pull-request :-)


willshowell:
Duplicate of https://github.com/angular/material2/pull/7792

mmalerba:
closing in favor of #7792

PR #7991, opened 2017-10-24
chore: fix plunker link in readme

DevVersion:
* The current Plunker link is referring to an anonymous Plunker that nobody can't change.
* Changes the Plunker and StackBlitz links in the Readme to the `goo.gl` shorthand links from the issue template.


andrewseguin:
Needs to be updated with HttpClient

Add the import to `systemjs.config.js`:
```
'@angular/common/http': 'https://unpkg.com/@angular/common/bundles/common-http.umd.js',
```

In usage:

`HttpModule` -> `HttpClientModule`
`Http` -> `HttpClient`

DevVersion:
@andrewseguin Updated the Plunker. It now uses the `HttpClientModule` and also the `VERSION` from the `@angular/material` module is being used instead.

Issue #7043, opened 2017-09-13
bug(drawer): mat-sidenav-opened/mat-drawer-opened not set when open


karlhaas:
#### Bug:
mat-sidenav-opened/mat-drawer-opened not set when open after upgrade to latest release

#### What is the expected behavior?
CSS class mat-drawer-opened set when sidenav is open

#### What is the current behavior?
CSS class mat-drawer-opened not set

#### What are the steps to reproduce?
https://material.angular.io/components/sidenav/examples

#### What is the use-case or motivation for changing an existing behavior?
that's the question :)

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular 4.3.6
Material/CDK: 2.0.0-beta.10 and material2-builds.git#32daeb981986c3f925ac59a58cd95023f971b431

willshowell:
It looks like `.mat-drawer-opened` is set on the `md-sidenav-container` when the sidenav mode is set to "over" or "push". It does not get added with "side" mode.

demiro:
To me it happens the same when I have the <md-sidenav-container fullscreen> in app-component
and then the sidenav in another... it doesn't get triggered.... how can I do it programatically?

josephperrott:
I believe this is working as intended.  As @willshowell mentioned, `.mat-drawer-opened` is only added in `over` and `push` modes.  As you can see [here](https://github.com/angular/material2/blob/master/src/lib/sidenav/drawer.ts#L455), it is watched for if `mode !== 'side'`.

If you are seeing different can you provide a plunker showing the issue you are seeing?  I am not quite sure I follow what you are seeing.

willshowell:
@josephperrott I could have sworn this was a regression, but just looked through the blame and it does seem this behavior has been in place for >1 year.

That said, there is specific effort in the drawer class logic to prevent `.mat-drawer-opened` from getting added to the drawer when in `'side'` mode, but as you can see, the effect is blocked in the CSS anyway. Perhaps this issue can be repurposed to having `.mat-drawer-opened` always be added to open drawers, as that would be the expected behavior without having to dive into the source.

https://github.com/angular/material2/blob/3571f68a2de0833ee4741af8a0eb24d1da174f38/src/lib/sidenav/drawer.scss#L114-L120



Note, it is used in one other instance, but could easily be blocked in CSS too if necessary:


<details>
<summary>Source</summary>

[...]

PR #7999, opened 2017-10-24
chore(drawer): add opened class for easier styling

crisbeto:
Adds the `mat-drawer-opened` and `mat-sidenav-opened` class for easier styling.


willshowell:
The same class is conditionally being added to the container. Should the class name be different or remove the old behavior? (related https://github.com/angular/material2/issues/7043)

https://github.com/angular/material2/blob/0ea43700efb105aa02ad3b1d3f800bf4a9925883/src/lib/sidenav/drawer.ts#L519-L526

crisbeto:
Oh, I forgot we had that. I'd rather close this PR since consumers can use the container to target the sidenav.

willshowell:
@crisbeto would you or @mmalerba mind looking at https://github.com/angular/material2/issues/7043#issuecomment-332278596? I feel like it would be an improvement to add `mat-drawer-opened` to the container regardless of the drawer's mode. Currently `side` mode does not get that treatment.

PR #8007, opened 2017-10-24
docs(table): change docs to use MatTableDataSource

andrewseguin:
Drastically simplifies the examples

@willshowell


src/material-examples/table-overview/table-overview-example.ts
```diff
@@ -1,69 +1,60 @@
...
+    // Assign the data to the data source for the table to render
+    this.dataSource = new MatTableDataSource(users);
   }
 
-  masterToggle() {
-    if (!this.dataSource) { return; }
+  /**
+   * Set the paginator and sort after the view has been initialized so
+   * that the components are completely set up for the data source to access.
+   */
```
willshowell: Is there any other way to word this? "...so that component are completely set up..." kind of glosses over why it has to be here instead of in, say, OnInit.

Maybe
> Set the paginator and sort after the view has been initialized since that is
the first time the component has access to the ViewChild queries.

src/material-examples/table-http/table-http-example.ts
```diff
@@ -18,20 +17,48 @@ import 'rxjs/add/operator/switchMap';
...
+              this.sort.active, this.sort.direction, this.paginator.pageIndex);
+        })
+        .map(data => {
+          // Flip flag to show that loading has finished.
+          this.isLoadingResults = false;
+          this.isRateLimitReached = false;
+          this.resultsLength = data.total_count
+
+          return data.items;
+        })
```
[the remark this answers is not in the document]
andrewseguin: I like `do` better here but concerned that adding more operators will distract people. Hard balance to find since people will probably copy/paste this =P

src/material-examples/table-overview/table-overview-example.ts
andrewseguin: Changed to 

> Set the sort after the view init since this component will be able to query its view for the initialized sort.

src/material-examples/table-http/table-http-example.ts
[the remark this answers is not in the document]
willshowell: Yeah, agreed on it being distracting. `map` works just as well since you still do have to map it to `items`.

src/material-examples/table-overview/table-overview-example.ts
willshowell: > Wait until after the view init to set the paginator and sort in the data source. This will enable the component to query its view for the initialized paginator and sort components.

Maybe an improvement? I guess it doesn't really matter, and I think having any comment at all should signify that it's important to keep it there after the copy-paste.

andrewseguin:
Made the suggested changes, thanks for the thorough review! I'm currently in the process of revamping the docs, and I'll share the doc with you for suggestions on what should be included (like a custom data source)

jelbourn:
@ErinCoughlan how do these example changes look to you?

src/material-examples/table-http/table-http-example.ts
```diff
@@ -48,69 +75,15 @@ export interface GithubIssue {
...
 
   getRepoIssues(sort: string, order: string, page: number): Observable<GithubApi> {
     const href = 'https://api.github.com/search/issues';
     const requestUrl =
-      `${href}?q=repo:angular/material2&sort=${sort}&order=${order}&page=${page + 1}`;
+        `${href}?q=repo:angular/material2&sort=${sort}&order=${order}&page=${page + 1}`;
 
-    return this._httpClient.get<GithubApi>(requestUrl);
+    return this.http.get(requestUrl)
+        .map(response => response as GithubApi);
```
Chris1308: Maybe it's nicer to use the type from HttpClient? `this.http.get<GithubApi>(requestUrl)`

src/material-examples/table-http/table-http-example.css
```diff
@@ -55,3 +38,13 @@
...
+
+/* Column Widths */
+.mat-column-number,
+.mat-column-state {
+  max-width: 64px;
+}
+
+.mat-column-created {
+  max-width: 124px;
+}
```
ErinCoughlan: no newline

src/material-examples/table-overview/table-overview-example.ts
```diff
@@ -1,69 +1,60 @@
...
 
-    if (this.isAllSelected()) {
-      this.selection.clear();
-    } else if (this.filter.nativeElement.value) {
-      this.dataSource.renderedData.forEach(data => this.selection.select(data.id));
-    } else {
-      this.exampleDatabase.data.forEach(data => this.selection.select(data.id));
-    }
+  applyFilter(filterValue: string) {
+    filterValue = filterValue.trim(); // Remove whitespace
```
ErinCoughlan: Does it make sense to add the trim and toLowerCase into the MatTableDataSource? It's not intuitive as a user that either of these would be necessary.

src/material-examples/table-sorting/table-sorting-example.css
```diff
@@ -1,23 +1,14 @@
...
 .mat-table {
   overflow: auto;
   max-height: 500px;
 }
 
-.mat-header-cell .mat-sort-header-sorted {
+.mat-header-cell.mat-sort-header-sorted {
   color: black;
-}
+}
```
ErinCoughlan: No newline at the end of the file.

jelbourn:
Add merge-ready when ready

src/material-examples/table-overview/table-overview-example.ts
andrewseguin: I think so - I expect most people will want it trimmed and lowercase by default. I'll make another PR that does this.

andrewseguin: PR #8059 created to address this - data source's entire filter matching function can be overriden and by default it will lowercase/trim

wk-done:
Hi, I noticed that the selection of rows was removed from the examples (table-overview-example) with these changes.
If I'm not completely wrong it was the only example out there how to use selection in the data-table. Is there an other example page or will there be one in the future?

PR #8011, opened 2017-10-24
fix(table): empty string should be sorted right

### LINKED ISSUE #7460: [Table] Sort example incorrectly sorts empty string
The sort function has a bug in it.

Strings should not be converted to numbers during sort

Empty string `""` is converted to zero with `+propertyA` and/or `+propertyB`, invalidating the comparison with other strings.  `0 < "x"` returns false, and `"x" < 0` is also false.  With a correct sort, `"" < "x"` returns true, and `"x" < ""` returns false.

Set any of the string values to empty ""

Preventing propagation of faulty sample code

Material since introduction of Table

andrewseguin:
Previously, an empty string gets converted to 0 and comes _after_ other values. E.g. ' ' would come after 'x'. 

Fixes #7460

### DIFF
--- a/src/lib/table/table-data-source.ts
+++ b/src/lib/table/table-data-source.ts
@@ -90,7 +90,9 @@ export class MatTableDataSource<T> implements DataSource<T> {
 
     // If the value is a string and only whitespace, return the value.
     // Otherwise +value will convert it to 0.
-    if (typeof value === 'string' && !value.trim()) { return value; }
+    if (typeof value === 'string' && !value.trim()) {
+      return value;
+    }
 
     return isNaN(+value) ? value : +value;
   }
--- a/src/lib/table/table-data-source.ts
+++ b/src/lib/table/table-data-source.ts
@@ -86,7 +86,12 @@ export class MatTableDataSource<T> implements DataSource<T> {
    * @param sortHeaderId The name of the column that represents the data.
    */
   sortingDataAccessor = (data: T, sortHeaderId: string): string|number => {
-    const value: number|string = data[sortHeaderId];
+    const value: any = data[sortHeaderId];
+
+    // If the value is a string and only whitespace, return the value.
+    // Otherwise +value will convert it to 0.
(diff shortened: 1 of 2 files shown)


src/lib/table/table-data-source.ts
```diff
@@ -86,7 +86,12 @@ export class MatTableDataSource<T> implements DataSource<T> {
    * @param sortHeaderId The name of the column that represents the data.
    */
   sortingDataAccessor = (data: T, sortHeaderId: string): string|number => {
-    const value: number|string = data[sortHeaderId];
+    const value: any = data[sortHeaderId];
+
+    // If the value is a string and only whitespace, return the value.
+    // Otherwise +value will convert it to 0.
+    if (typeof value === 'string' && !value.trim()) { return value; }
```
jelbourn: This conditional is a bit long to be single-lined

Bsujeet:
i just wanted to ask if the table should sort the column where the is same i.e all the rows of that column contain same data.

this is my table before sort.

image

this is my table after sort on **DiskType** column..

image

andrewseguin:
@Bsujeet What you are looking for is a [stable sort](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability), but the [ES7 spec](http://www.ecma-international.org/ecma-262/7.0/index.html#sec-array.prototype.sort) does not specify that native sorting must be stable. Each browser has its own way of sorting when array.sort is called and there's no consistency.

For a stable sort, you'll want to adjust your comparator to account for the object's index if the comparison is equal. For some discussion on how to do this, try starting with this [StackOverflow question](https://stackoverflow.com/questions/1427608/fast-stable-sorting-algorithm-implementation-in-javascript/45422645#45422645).

Bsujeet:
Thanks it was helpfull

prabhat112:
I have 5.0.0 and I still have this issue. In which release has this been released ?

andrewseguin:
Should be out already - feel free to open an issue with reproduction steps so we can investigate if the issue persists

PR #8024, opened 2017-10-25
fix(autosize): incorrect height with long placeholders

### LINKED ISSUE #8013: MatTextareaAutosize height breaks when placeholder text longer than width of input
When you set a longer placeholder text on a textarea using autoresize, the calculated height for the textarea when empty wrong and changes once you start typing.

The empty value should just be the height of 1 line

The resulting height with an empty field is the height that the textarea would be if the placeholder text was in it.

Set a long placeholder text on a textarea
see: https://angular-material2-issue-dkvdyn.stackblitz.io
```
      <mat-form-field style="width: 200px;">
        <textarea matInput matTextareaAutosize placeholder="Set something long in the placeholder"></textarea>
      </mat-form-field>
      <mat-form-field style="width: 200px;">
        <textarea matInput matTextareaAutosize placeholder="Short text fine"></textarea>
      </mat-form-field>
```

Textarea height jumps when empty or not having a value

Chrome & Safari at least

The placeholder text should be ignored when calculating the height of the textarea

DevVersion:
Long placeholders can cause the `scrollHeight` to be bigger than the actual content is. To ensure the `scrollHeight` is correct, the placeholders need to be removed temporarily.

Fixes #8013

### DIFF
--- a/src/lib/input/autosize.spec.ts
+++ b/src/lib/input/autosize.spec.ts
@@ -28,7 +28,8 @@ describe('MatTextareaAutosize', () => {
         AutosizeTextareaInATab,
         AutosizeTextAreaWithContent,
         AutosizeTextAreaWithValue,
-        AutosizeTextareaWithNgModel
+        AutosizeTextareaWithNgModel,
+        AutosizeTextareaWithLongPlaceholder
       ],
     });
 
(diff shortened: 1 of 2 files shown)


src/lib/input/autosize.ts
```diff
@@ -150,14 +150,21 @@ export class MatTextareaAutosize implements AfterViewInit, DoCheck {
       return;
     }
 
+    const previousPlaceholder = textarea.placeholder;
```
jelbourn: "previous" isn't exactly right because we're not changing it. How about `placeholderText`?

src/lib/input/autosize.spec.ts
```diff
@@ -176,6 +177,39 @@ describe('MatTextareaAutosize', () => {
...
+    const fixtureWithPlaceholder = TestBed.createComponent(AutosizeTextareaWithLongPlaceholder);
+    fixtureWithPlaceholder.detectChanges();
+
+    textarea = fixtureWithPlaceholder.nativeElement.querySelector('textarea');
+    autosize = fixtureWithPlaceholder.debugElement.query(
+      By.directive(MatTextareaAutosize)).injector.get<MatTextareaAutosize>(MatTextareaAutosize);
+
+    // To be able to trigger a new calculation of the height with a long placeholder, the textarea
+    // value needs to be changed.
+    textarea.value = '1';
```
jelbourn: Make a helper function for this `triggerTextareaResize()`?

DevVersion: Yeah that sounds good.

src/lib/input/autosize.ts
DevVersion: Done.

jelbourn:
@DevVersion travis is super red

DevVersion:
@jelbourn Yeah but unrelated. See #8030

PR #8036, opened 2017-10-25
chore: create tslint rule to allow @HostListener and @HostBinding in abstract classes

josephperrott:


tools/tslint-rules/noHostDecoratorInConcreteRule.ts
```diff
@@ -0,0 +1,36 @@
+import * as ts from 'typescript';
+import * as Lint from 'tslint';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
```
crisbeto: It looks like the indentation of the whole file is too large (seems like it's a 4, but it should be 2).

tools/tslint-rules/noHostDecoratorInConcreteRule.ts
```diff
@@ -0,0 +1,36 @@
...
+    visitClassDeclaration(node: ts.ClassDeclaration) {
+      if (!node.modifiers || !this.getOptions().length) { return; }
+
+      // Do not check the class if its abstract.
+      if (!!node.modifiers.find(modifier => modifier.kind === ts.SyntaxKind.AbstractKeyword)) {
+        return;
+      }
+
+      node.members.forEach(el => {
+        if (el.decorators) {
```
crisbeto: This block could be a bit more concise with a `filter` and a `map`:

```ts
node.members
  .filter(el => !!el.decorators)
  .map(el => el.decorators)
  .forEach(decorator => {
    const decoratorText: string = decorator.getChildAt(1).getText();
    const matchedDecorator: string = this.getOptions().find(
      (item: string) => decoratorText.startsWith(item));
    
      if (!!matchedDecorator) {
        this.addFailureFromStartToEnd(decorator.getChildAt(1).pos - 1, decorator.end,
            `The @${matchedDecorator} decorator may only be used in abstract classes. In
              concrete classes use \`host\` in the component definition instead.`);
      }
});
```

tools/tslint-rules/noHostDecoratorInConcreteRule.ts
```diff
@@ -0,0 +1,36 @@
...
+          el.decorators.forEach(decorator => {
+            const decoratorText: string = decorator.getChildAt(1).getText();
+            const matchedDecorator: string = this.getOptions().find(
+              (item: string) => decoratorText.startsWith(item));
+            if (!!matchedDecorator) {
+              this.addFailureFromStartToEnd(decorator.getChildAt(1).pos - 1, decorator.end,
```
crisbeto: Not completely sure, but you should be able to use `this.addFailureAtNode(decorator, ...)` instead.

josephperrott: I initially tried this, but the starting position for the decorator was the character after the previous node, which made the tslint annotation go crazy.

tools/tslint-rules/noHostDecoratorInConcreteRule.ts
```diff
@@ -0,0 +1,37 @@
+import * as ts from 'typescript';
+import * as Lint from 'tslint';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+    return this.applyWithWalker(new Walker(sourceFile, this.getOptions()));
+  }
+}
+
+class Walker extends Lint.RuleWalker {
```
crisbeto: Add a short docstring about what the rule does?

PR #8037, opened 2017-10-25
fix: Get correct viewport dimensions in ConnectedPositionStrategy

isaacplmann:
Should get viewport dimensions using `window.innerWidth`/`window.innerHeight` instead of `document.documentElement.clientWidth`/`document.documentElement.clientHeight`.

### DIFF
--- a/src/cdk/overlay/position/connected-position-strategy.ts
+++ b/src/cdk/overlay/position/connected-position-strategy.ts
@@ -358,7 +358,7 @@ export class ConnectedPositionStrategy implements PositionStrategy {
     // from the bottom of the viewport rather than the top.
     let y = verticalStyleProperty === 'top' ?
         overlayPoint.y :
-        window.innerHeight - (overlayPoint.y + overlayRect.height);
+        document.querySelector('.cdk-overlay-container').clientHeight - (overlayPoint.y + overlayRect.height);
 
     // We want to set either `left` or `right` based on whether the overlay wants to appear "before"
     // or "after" the origin, which determines the direction in which the element will expand.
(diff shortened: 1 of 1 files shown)


isaacplmann:
It miscalculates the `bottom` style when the `<body>` is scrollable and the overlay needs to be placed above the origin point.  I'm working on a reproduction repo.

isaacplmann:
Code:
https://github.com/isaacplmann/ngx-contextmenu

Running Demo:
https://isaacplmann.github.io/ngx-contextmenu/

Steps to reproduce:
1. Shrink your browser height so that right-clicking on an item would make the context menu overflow the bottom of the viewport.
2. Right-click that item

Expected results:
The context menu is rendered above the point you clicked.  i.e. Style is something like `bottom: 36px`

Actual results:
The context menu is rendered above the top of the viewport.  i.e. Style is something like `bottom: 1086px`

That bottom value is the distance from the bottom of document, but is being rendered that distance from the bottom of the viewport.  This could alternatively be fixed like this:
```html
<html style="position:relative">
  <div class="cdk-overlay-container" style="position: absolute">
```
But that alternative will probably break other things.

Relevant code in my library is here:
https://github.com/isaacplmann/ngx-contextmenu/blob/master/src/lib/contextMenu.service.ts#L68

I'd be really happy to find out that I'm just doing it wrong.

isaacplmann:
I see what you're referring to about scrollbars messing up the calculation.  I think what we really want is to use the height and width of the `.cdk-overlay-container` and not `window.innerHeight` or `document.documentElement.clientHeight`

isaacplmann:
I've discovered the problem.  I wasn't including `<!doctype html>` at the top of the `index.html` file.  For some reason that causes `document.documentElement.clientHeight` to return the height of the document content, rather than the height of the viewport.  That's a very non-obvious solution to a positioning problem.  Using the height/width of the `.cdk-overlay-container` would solve this regardless of the doctype.

crisbeto:
This isn't the only issue caused by not including a doctype though. Material actually logs a warning if your document doesn't have a doctype, but you might have missed it if you've silenced console warnings.

isaacplmann:
Ok. Then perhaps the right solution is to have cdk log a warning as well.
My library doesn't use material, just the cdk.

isaacplmann:
I'll open an issue to have the cdk warn when not using the correct doctype.  Closing this PR.

PR #8044, opened 2017-10-26
fix(gulp-tasks): Use red directly instead of color.red

ppham27:
colors is just an instance of chalk. Using chalk directly fixes the build.

This PR possibly may be unnecessary with #8030. If so, feel free to close.


DevVersion:
Thanks for the PR. 

I updated my PR to completely remove `gulp-util`. We only use the `log` method once and `console.log` works fine as a replacement.

Closing this PR in favor of #8030

PR #8051, opened 2017-10-26
fix(dialog): don't block other dialogs from opening while animating

### LINKED ISSUE #6560: [MdDialog] opening multiple dialog return wrong dialogRef.
Bug

If I open a dialog while another is in the process of being opened it should be opened and return the `dialogRef` of the requested ones

The requested dialog is ignored and a return a dialogRef which not belong the the requested dialog

https://github.com/angular/material2/blob/master/src/lib/dialog/dialog.ts#L123-L126

crisbeto:
The behavior where we don't allow new dialogs to be opened while a dialog is animating was introduced initially in #5769 to avoid issues where the user might open multiple dialogs by accident if they press enter one too many times. This is problematic because it also means that the consumer is blocked from opening dialogs for certain 300ms intervals. These changes approach the issue differently by moving focus onto the dialog container immediately while it's animating.

Fixes #6560.

### DIFF
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -100,9 +100,6 @@ export class MatDialogContainer extends BasePortalHost {
   /** ID of the element that should be considered as the dialog's label. */
   _ariaLabelledBy: string | null = null;
 
-  /** Whether the container is currently mid-animation. */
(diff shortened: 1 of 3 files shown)


owen26:
Hope this can be merged soon.

PR #8053, opened 2017-10-26
fix(menu): make @Output names consistent #6677

amcdnl:
Clean PR of #7133


src/lib/menu/menu-panel.ts
```diff
@@ -15,7 +15,7 @@ export interface MatMenuPanel {
   yPosition: MenuPositionY;
   overlapTrigger: boolean;
   templateRef: TemplateRef<any>;
-  close: EventEmitter<void | 'click' | 'keydown'>;
+  closed: EventEmitter<void | 'click' | 'keydown'>;
```
jelbourn: Presubmit has revealed that this is a problem. Many teams implement `MatMenuPanel`, so changing this to `close` breaks them. This one alone should be changed back to `closed` (aliasing it for the directive)

amcdnl: Ok, I switched that interface back to `close` and kept the alias for `close` to `closed` in the directive. Let me know if that is what you were thinking.

PR #8055, opened 2017-10-26
fix(dialog): dialogs that hit max-width not centered on IE

crisbeto:
Fixes dialogs that hit their `max-width` no longer being centered in IE11. This seems like a regression introduced after 57f19cd086bde26724b6430c63baacd1f835380a.

### DIFF
--- a/src/cdk/overlay/_overlay.scss
+++ b/src/cdk/overlay/_overlay.scss
@@ -39,6 +39,9 @@ $backdrop-animation-timing-function: cubic-bezier(0.25, 0.8, 0.25, 1) !default;
     display: flex;
     position: absolute;
     z-index: $cdk-z-index-overlay;
+
+    // Fixes issues where overlays are centered on IE when they hit the max-width.
+    flex-direction: column;
   }
 
   // A single overlay pane.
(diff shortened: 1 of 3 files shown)


andrewseguin:
Okay, bad news :( 

This PR ended up having an unexpected issue in a screenshot diff that I think needs to be resolved before we merge this back in.

To reproduce this in the demo app, go to the `overlay-demo.ts` and change one of the overlays so that it uses the position strategy `this.overlay.position().global().right('0px')`

Start the demo app and observe that the `cdk-overlay-0` element has dimensions 102px x 73px

Now, revert your changes: on `cdk-global-overlay-wrapper`, remove `flex-direction: column; align-items: flex-end;` and add back `justify-content: flex-end`)

See now that the dimensions of `cdk-overlay-0` are larger and extend the full height of the screen.

This was a problem in one app where their content assumed that they could use `height: 100%` and use overlay to make a sliding panel.

crisbeto:
Let's park this one until https://github.com/angular/material2/pull/9236 gets in, I think it has some changes that could help with the initial issue.

PR #8056, opened 2017-10-26
fix(overlay): overlay class audits #6372

amcdnl:
Clean cut of #7454 that only changes names

### DIFF
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -167,6 +167,9 @@ export class MatDatepicker<D> implements OnDestroy {
    */
   @Output() selectedChanged = new EventEmitter<D>();
 
+  /** Classes to be passed to the date picker panel. Supports the same syntax as `ngClass`. */
+  @Input() panelClass: string|string[]|Set<string>|{[key: string]: any};
+
   /** Whether the calendar is open. */
   opened = false;
 
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -168,7 +168,7 @@ export class MatDatepicker<D> implements OnDestroy {
   @Output() selectedChanged = new EventEmitter<D>();
 
   /** Classes to be passed to the date picker panel. Supports the same syntax as `ngClass`. */
-  @Input() panelClass: string|string[]|Set<string>|{[key: string]: any};
+  @Input() panelClass: string | string[];
 
   /** Whether the calendar is open. */
   opened = false;
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -148,7 +148,9 @@ export class MatMenu implements AfterContentInit, MatMenuPanel, OnDestroy {
    * @deprecated Use `panelClass` instead.
    * @param classes list of class names
    */
-  @Input() classList = this.panelClass;
+  @Input()
+  set classList(classes: string) { this.panelClass = classes; }
+  get classList(): string { return this.panelClass; }
 
   /** Event emitted when the menu is closed. */
   @Output() close = new EventEmitter<void | 'click' | 'keydown'>();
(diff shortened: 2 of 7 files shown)


src/lib/snack-bar/snack-bar-config.ts
```diff
@@ -35,7 +35,13 @@ export class MatSnackBarConfig {
   duration?: number = 0;
 
   /** Extra CSS classes to be added to the snack bar container. */
-  extraClasses?: string[];
+  panelClass?: string[];
```
willshowell: I have two concerns with this,

1. This adds the class to the snackbar container, not the overlay panel. So it's slightly misleading and inconsistent with dialogs in that regard.
2. More importantly, I think `panelClass` implies that a single string would be sufficient. Since `string[]` is req'd, the snackbar container will convert `panelClass: "my-class"` to 

    ```
    <snack-bar-container class="m y - c l a s">...</snack-bar-container>
    ```

    I think the `MatSnackBar#_createOverlay` could use the overlay-ref logic to automatically handle this .

I know this PR is trying to only change the names, but I think this name change is misrepresenting unless it's also met with a change in functionality.

src/lib/datepicker/datepicker.ts
```diff
@@ -167,6 +167,9 @@ export class MatDatepicker<D> implements OnDestroy {
    */
   @Output() selectedChanged = new EventEmitter<D>();
 
+  /** Classes to be passed to the date picker panel. Supports the same syntax as `ngClass`. */
+  @Input() panelClass: string|string[]|Set<string>|{[key: string]: any};
```
jelbourn: I think `Set<string>|{[key: string]: any}` is a bit overkill. I'd rather stick with `string | string[]` for the time being.

src/lib/menu/menu-directive.ts
```diff
@@ -141,6 +141,15 @@ export class MatMenu implements AfterContentInit, MatMenuPanel, OnDestroy {
...
   }
 
+  /**
+   * This method takes classes set on the host mat-menu element and applies them on the
+   * menu template that displays in the overlay container.  Otherwise, it's difficult
+   * to style the containing menu from outside the component.
+   * @deprecated Use `panelClass` instead.
+   * @param classes list of class names
+   */
+  @Input() classList = this.panelClass;
```
jelbourn: This needs a getter/setter that delegates to `panelClass`; this will only set the initial value

src/lib/snack-bar/snack-bar-config.ts
jelbourn: Agreed; I didn't realize snackbar was `string[]` only right now. @amcdnl let's _just_ change this one to `string | string[]` as well

src/lib/snack-bar/snack-bar-config.ts
```diff
@@ -35,7 +35,13 @@ export class MatSnackBarConfig {
...
+
+  /**
+   * Extra CSS classes to be added to the snack bar container.
+   * @deprecated Use `panelClass` instead.
+   */
+ extraClasses?: string[] = this.panelClass;
```
jelbourn: This leaves the option here, but it will still break because it doesn't do anything now; the snack-bar needs to combine both until we remove this one.

amcdnl: Done. Checkout what I did there, open to suggestions for better ideas.

src/lib/menu/menu-directive.ts
```diff
@@ -141,6 +141,17 @@ export class MatMenu implements AfterContentInit, MatMenuPanel, OnDestroy {
     }
   }
 
+  /**
+   * This method takes classes set on the host mat-menu element and applies them on the
+   * menu template that displays in the overlay container.  Otherwise, it's difficult
+   * to style the containing menu from outside the component.
+   * @deprecated Use `panelClass` instead.
+   * @param classes list of class names
```
jelbourn: Setter doesn't take `@param`

PR #8059, opened 2017-10-26
fix(table): broaden abstraction for filtering

andrewseguin:
Currently the user can customize the `MatTableDataSource`'s filtering by providing a custom function of how to convert a data object into a filter string.

This may be too in-depth for most users and not as intuitive. 

This PR broadens the abstraction so that users instead will define a function that accepts data and a filter string, and returns true/false depending on if the data matches.

This also allows the data source some freedom in defaulting to a behavior where it trims/lowercases the filter but allows users to override that behavior easily.

### DIFF
--- a/src/lib/table/table-data-source.ts
+++ b/src/lib/table/table-data-source.ts
@@ -43,8 +43,11 @@ export class MatTableDataSource<T> implements DataSource<T> {
   get data() { return this._data.value; }
 
   /**
-   * Filter term that should be used to filter out objects from the data array. To override how
-   * the filter matches data objects, provide a custom function on filterTermAccessor.
+   * Filter term that should be used to filter out objects from the data array. Will be transformed
+   * by the filterTransformer function, which by default will transform the filter into lowercase
+   * and remove surrounding whitespace. To override this behavior, provide a custom function on the
+   * filterTransformer. To override how data objects are converted to strings to match again,
+   * provide a custom function on filterTermAccessor.
    */
   set filter(filter: string) { this._filter.next(filter); }
   get filter(): string { return this._filter.value; }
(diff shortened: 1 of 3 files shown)


src/lib/table/table-data-source.ts
```diff
@@ -91,14 +91,24 @@ export class MatTableDataSource<T> implements DataSource<T> {
...
+   * is converted to a string of its properties and returns true if the filter has
+   * at least one occurrence in that string. By default, the filter string has its whitespace
+   * removed and the match is case-insensitive. May be overriden for a custom implementation of
+   * filter matching.
+   * @param data Data object used to check against the filter.
+   * @param filter Filter string that has been set on the data source.
+   * @returns Whether the filter matches against the data
    */
-  filterTermAccessor: ((data: T) => string) = (data: T): string => {
+  filterMatcher: ((data: T, filter: string) => boolean) = (data: T, filter: string): boolean => {
```
jelbourn: `filterPredicate`?

src/lib/table/table-data-source.ts
```diff
@@ -91,14 +91,24 @@ export class MatTableDataSource<T> implements DataSource<T> {
...
+    // Transform the data into a lowercase string of all property values.
     const accumulator = (currentTerm, key) => currentTerm + data[key];
-    return Object.keys(data).reduce(accumulator, '').toLowerCase();
+    const dataStr = Object.keys(data).reduce(accumulator, '').toLowerCase();
```
jelbourn: You want to filter on the _keys_?

andrewseguin: Looks like it, but its actually iterating over each of the object's keys and then reducing it via the accumulator function. That function takes each key and accesses the data's property and appends it to the accumulated value.

In the end an object like `{name: 'John', color: 'Blue', age: '27'}` ends up looking like `'JohnBlue27'`

Would be nice to use `Object.values()` but its not supported by IE

andrewseguin:
Renamed the filter function to `filterPredicate`

src/lib/table/table-data-source.ts
```diff
@@ -91,14 +91,24 @@ export class MatTableDataSource<T> implements DataSource<T> {
...
 
   /**
-   * Transforms data objects into a filter term that will be used to check against the filter if
-   * a filter is set. By default, the function will iterate over the values of the data object
-   * and convert them to a lowercase string.
-   * @param data Data object to convert to a string that checked for containing the filter term.
+   * Checks if a data object matches the data source's filter string. By default, each data object
+   * is converted to a string of its properties and returns true if the filter has
+   * at least one occurrence in that string. By default, the filter string has its whitespace
+   * removed and the match is case-insensitive. May be overriden for a custom implementation of
```
willshowell: "...has its whitespace **trimmed** and the match..."?

To me, "removed" kind of implies that `{ name: "Mr. Smith" }` would end up as `mr.smith`

andrewseguin: Good call, I can see how that could be taken both ways

PR #8060, opened 2017-10-26
fix(select): closing parent overlay when pressing escape

### LINKED ISSUE #7981: mat-select should close when escape is pressed, also stop propagation
bug?

when mat-select is open and escape is pressed it should close the select

select doesn't close and the element upstream gets the escape. In my case the mat-drawer catches this Escape which results in weird behavior: drawer closes, and the opened select is floating outside of it (because it's part of the cdk-portal).

http://plnkr.co/edit/vtBv4N8p0o4hzFetaOGx?p=preview

Open drawer
Click the selector
Press ESC

crisbeto:
After the switch to `ActiveDescendantKeyManager`, the select was no longer handling the escape key functionality internally, but was delegating to the underlying overlay. Since the overlay listens to keyboard events on the document, it means that escape key presses will close any parent overlays along the way. These changes add the escape listener to the select itself and stop the event propagation.

**Note:** the overlay directive listener should be scoped to the overlay itself, but it's better if we defer refactoring it until #6682 gets in.

Fixes #7981.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -1,5 +1,5 @@
 import {Directionality} from '@angular/cdk/bidi';
-import {DOWN_ARROW, END, ENTER, HOME, SPACE, TAB, UP_ARROW} from '@angular/cdk/keycodes';
+import {DOWN_ARROW, END, ENTER, HOME, SPACE, TAB, UP_ARROW, ESCAPE} from '@angular/cdk/keycodes';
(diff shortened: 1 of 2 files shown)


willshowell:
@crisbeto fyi #6682 is merged, in case you want to try your other solution before review

crisbeto:
Closing in favor of https://github.com/angular/material2/pull/8531.

PR #8061, opened 2017-10-26
fix(slider): prevent slider thumb from getting cropped

shevaroller:
Slider thumb may sometimes get cropped by a scale transformation.
Removing backface visibility ensures that the thumb is painted correctly.

### DIFF
--- a/src/lib/slider/slider.scss
+++ b/src/lib/slider/slider.scss
@@ -393,6 +393,10 @@ $mat-slider-focus-ring-size: 30px !default;
     left: 50%;
   }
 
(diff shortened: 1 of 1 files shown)


shevaroller:
I signed it.

mmalerba:
are you sure you used the same account to sign it as you did to create the commits? it should automatically update if its signed properly

shevaroller:
Thank you for pointing it out, @mmalerba. Google pulled up a different email when I was signing a CLA. I signed it!

mmalerba:
great, thanks!

PR #8069, opened 2017-10-27
docs: add note about web animations polyfill

crisbeto:
Adds a note that the WebAnimations polyfill is required for component animations to work in all browsers.

### DIFF
--- a/guides/getting-started.md
+++ b/guides/getting-started.md
@@ -28,6 +28,10 @@ install the `@angular/animations` module and include the `BrowserAnimationsModul
 npm install --save @angular/animations
 ```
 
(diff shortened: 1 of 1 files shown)


aitboudad:
is there any chance to make `@angular/animations` optional ?
animations take an amount of 100kb (60kb angular/animation + 40kb polyfill) which larger than material components itself

crisbeto:
`@angular/animations` are optional already @aitboudad, if you don't include them, the components will still work, they just won't animate.

aitboudad:
@crisbeto not really optional, we must import `NoopAnimationsModule` which cost an additional ~60kb.

jelbourn:
@crisbeto Yeah, you still need all the code that understands the `@` properties and the animation DSL.

PR #8078, opened 2017-10-27
chore(docs): show extends clauses for classes

DevVersion:
Classes like `MatAnchor` in the button docs should indicate that those are extending the `MatButton` class. 

Clauses without any related Dgeni doc can be ignored, because no information for this is available (right now; e.g. `MatSlideToggleBase` with mixins)

### DIFF
--- a/tools/dgeni/processors/categorizer.ts
+++ b/tools/dgeni/processors/categorizer.ts
@@ -19,6 +19,7 @@ import {
   isProperty,
   isService
 } from '../common/decorators';
+import {ClassLikeExportDoc} from 'dgeni-packages/typescript/api-doc-types/ClassLikeExportDoc';
 import {MethodMemberDoc} from 'dgeni-packages/typescript/api-doc-types/MethodMemberDoc';
 import {sortCategorizedMembers} from '../common/sort-members';
 
@@ -31,6 +32,7 @@ export interface CategorizedClassDoc extends ClassExportDoc {
   isDeprecated: boolean;
   directiveExportAs?: string | null;
   directiveSelectors?: string[];
+  extendedDocs: ClassLikeExportDoc[];
 }
 
 export interface CategorizedPropertyMemberDoc extends PropertyMemberDoc {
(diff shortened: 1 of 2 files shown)


tools/dgeni/templates/class.template.html
```diff
@@ -6,6 +6,15 @@ <h4 class="docs-api-h4 docs-api-class-name">
 <p class="docs-api-class-description">{$ class.description | safe $}</p>
 {%- endif -%}
 
+{% if class.extendedDocs.length %}
+<p class="docs-api-class-extends-clauses">
+  <span class="docs-api-class-extends-label">Extends:</span>
+  {% for extendedDoc in class.extendedDocs %}
+    <span class="docs-api-class-extends-type">{$ extendedDoc.name $}</span>
```
jelbourn: I think it would be better to show this immediately after the name, e.g.
```
MatVerticalStepper extends MatStepper
```

tools/dgeni/processors/categorizer.ts
```diff
@@ -88,6 +90,12 @@ export class Categorizer implements Processor {
     classDoc.methods.sort(sortCategorizedMembers);
     classDoc.properties.sort(sortCategorizedMembers);
 
+    // Filter the extends clauses for clauses with an associated Dgeni document. Clauses without
+    // a document are unknown and should not be mentioned in the documentation for this class.
+    classDoc.extendedDocs = classDoc.extendsClauses
+      .filter(clause => clause.doc)
+      .map(clause => clause.doc!);
```
jelbourn: Did you confirm that this filters out mixin classes like `MatButtonBase`?

DevVersion: Yeah, those are filtered out because there is no Dgeni doc for those.

tools/dgeni/templates/class.template.html
DevVersion: I tried that and it didn't look that well in my opinion.

image


But maybe there will be some advanced styling to make this look better. Let me know if we should go that way.

jelbourn:
I think a little css would make it look good

DevVersion:
@jelbourn Updated the PR and addressed the feedback.

jelbourn:
There's a lint error: you can add merge-ready when ready

DevVersion:
Seems unrelated. But can fix in the PR as well

DevVersion:
Lint issue will be fixed by #8086.

PR #8080, opened 2017-10-27
fix: don't show sanity check messages in tests

jelbourn:


crisbeto:
Needs a rebase @jelbourn.

PR #8083, opened 2017-10-27
Revert "chore(docs): adds types column and null checks for description"

mmalerba:
Reverts angular/material2#7713


mmalerba:
reverting per @jelbourn's request

Issue #7544, opened 2017-10-05
Rename `PortalHost` to `PortalOutlet`


jelbourn:
This will make it more consistent with RouterOutler, NgTemplateOutlet, NgComponentOutlet, etc.

We should still export the old name as deprecated.

PR #8088, opened 2017-10-27
fix: consistent names for all cdk directives

jelbourn:
Based on #7544. Should be fully backwards compatible.

Renames:
`OverlayOrigin` -> `CdkOverlayOrigin`
`ConnectedOverlayDirective` -> `CdkConnectedOverlay`
`PortalDirective` -> `CdkPortal`
`PortalHostDirective` -> `CdkPortalOutlet`
`ObserveContent` -> `CdkObserveContent`
    
Also stop using "mat" completely in the cdk.
    
BREAKING CHANGE:
`MAT_CONNECTED_OVERLAY_SCROLL_STRATEGY` is renamed to
`CDK_CONNECTED_OVERLAY_SCROLL_STRATEGY`

### DIFF
--- a/src/cdk/overlay/overlay-directives.ts
+++ b/src/cdk/overlay/overlay-directives.ts
@@ -52,22 +52,21 @@ const defaultPositionList = [
 ];
 
 /** Injection token that determines the scroll handling while the connected overlay is open. */
-export const MAT_CONNECTED_OVERLAY_SCROLL_STRATEGY =
-    new InjectionToken<() => ScrollStrategy>('mat-connected-overlay-scroll-strategy');
+export const CDK_CONNECTED_OVERLAY_SCROLL_STRATEGY =
+    new InjectionToken<() => ScrollStrategy>('cdk-connected-overlay-scroll-strategy');
 
 /** @docs-private */
-export function MAT_CONNECTED_OVERLAY_SCROLL_STRATEGY_PROVIDER_FACTORY(overlay: Overlay):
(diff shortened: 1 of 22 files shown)


src/cdk/overlay/overlay-directives.ts
```diff
@@ -78,22 +77,21 @@ export const MAT_CONNECTED_OVERLAY_SCROLL_STRATEGY_PROVIDER = {
   selector: '[cdk-overlay-origin], [overlay-origin], [cdkOverlayOrigin]',
   exportAs: 'cdkOverlayOrigin',
 })
-export class OverlayOrigin {
+export class CdkOverlayOrigin {
```
crisbeto: I was thinking about this one a few days ago: do we really need the directive? It doesn't really provide a lot of value since using it ends up being slightly longer than straight-up passing in the `elementRef` to the overlay:

```html
<button cdk-overlay-origin #origin="cdkOverlayOrigin">Open</button>
<ng-template cdk-connected-overlay [origin]="origin"></ng-template>
```

Versus something like this if we got rid of it:

```html
<button #origin>Open</button>
<ng-template cdk-connected-overlay [origin]="origin"></ng-template>
```

It's not something we should necessarily address now, but it just something to think about.

jelbourn: The problem is that if the element you're applying it do has a component (like `mat-button`), you don't have a way to grab the ElementRef in the template.

PR #8089, opened 2017-10-27
fix(checkbox): Set aria-checked to mixed for indeterminate checkbox

### LINKED ISSUE #7932: checkbox a11y issue: Partially done task is getting read out as 'Checkbox not checked'.

Bug

Partially done task should read out as 'indeterminate'.

Partially done task is getting read out as 'Checkbox not checked'. It might cause confusion to end user. 

1. Click on Checkbox.2. Navigate over nested checklist.

Chromevox Chrome OS Version 61.0.3163.100 (Official Build) (64-bit) , Voiceover (mac OS version: 10.12.6)

tinayuangao:
Fixes #7932

### DIFF
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -87,6 +87,8 @@ describe('MatCheckbox', () => {
       expect(checkboxNativeElement.classList).not.toContain('mat-checkbox-checked');
       expect(inputElement.checked).toBe(false);
       expect(inputElement.indeterminate).toBe(false);
+      expect(inputElement.getAttribute('aria-checked'))
+          .toBe('false', 'Expect aria-checked to be false');
 
       testComponent.isIndeterminate = true;
       fixture.detectChanges();
@@ -94,6 +96,8 @@ describe('MatCheckbox', () => {
       expect(checkboxNativeElement.classList).toContain('mat-checkbox-indeterminate');
       expect(inputElement.checked).toBe(false);
       expect(inputElement.indeterminate).toBe(true);
+      expect(inputElement.getAttribute('aria-checked'))
+          .toBe('mixed', 'Expect aria checked to be mixed for indeterminate checkbox');
 
       testComponent.isIndeterminate = false;
       fixture.detectChanges();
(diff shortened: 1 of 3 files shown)


src/lib/checkbox/checkbox.ts
```diff
@@ -303,6 +303,10 @@ export class MatCheckbox extends _MatCheckboxMixinBase implements ControlValueAc
     this._changeDetectorRef.markForCheck();
   }
 
+  getAriaChecked() {
+    return this.checked ? true : (this.indeterminate ? 'mixed' : false);
```
jelbourn: Return type should be `'true' | 'false' | 'mixed'` (since the attr should always be a string)

tinayuangao: Done. Thanks for review!

src/lib/checkbox/checkbox.ts
```diff
@@ -303,6 +303,10 @@ export class MatCheckbox extends _MatCheckboxMixinBase implements ControlValueAc
     this._changeDetectorRef.markForCheck();
   }
 
+  getAriaChecked(): string {
```
jelbourn: Just also realized: method should be internal `_getAriaChecked`

The return type can really the string literal type `'true' | 'false' | 'mixed'`

src/lib/checkbox/checkbox.spec.ts
```diff
@@ -87,13 +87,17 @@ describe('MatCheckbox', () => {
       expect(checkboxNativeElement.classList).not.toContain('mat-checkbox-checked');
       expect(inputElement.checked).toBe(false);
       expect(inputElement.indeterminate).toBe(false);
+      expect(inputElement.getAttribute('aria-checked'))
+        .toBe('false', 'Expect aria-checked to be false');
```
jelbourn: +2 more indent

src/lib/checkbox/checkbox.ts
```diff
@@ -303,6 +303,10 @@ export class MatCheckbox extends _MatCheckboxMixinBase implements ControlValueAc
     this._changeDetectorRef.markForCheck();
   }
 
+  _getAriaChecked(): 'true'|'false'|'mixed' {
```
jelbourn: We typically put spaces between the types like this
```ts
'true' | 'false' | 'mixed'
```

PR #7276, opened 2017-09-23
feat: update to rxjs 5.5.0 and switch to lettable operators

crisbeto:
Bumps the required RxJS version to 5.5.0 and gets rid of our RxChain in favor of using the lettable Rx operators. Refactors all the usages and updates the configs.

Fixes #7275.

PR #8092, opened 2017-10-28
fix(docs): rxjs path in CODING_STANDARDS.md

julianobrasil:


rafaelss95:
When PR #7276 gets in, this line will be erased.

julianobrasil:
So I'm closing this.

PR #8096, opened 2017-10-28
ci: skip more tasks if only markdown files changed

DevVersion:
* No longer just skips the unit test and e2e tasks if only markdown files changed. Now everything will be skipped, since there is nothing that tests against the `.md` files.
* Prints messages for the deploy tasks that run inside of Pull Requests but end up in a NOOP.


jelbourn:
Shouldn't this PR be skipping unit tests? Looks like they're still running on it

DevVersion:
Every task will run in this PR because `.js` files have been changed here.

jelbourn:
Duh. My brain is not functioning today.

PR #8098, opened 2017-10-28
feat(datepicker): add opened input binding

### LINKED ISSUE #8094: Datepicker: make property "opened" an input
Feature request

I would like to have the datepicker opened at load time, for that I expected to use [opened]=true, but it turns out that it's a simple property, not an input, which means that I cannot set the date picker to open at load time. If i use `open()` in ngAfterViewInit I get an `ExpressionChangedAfterItHasBeenCheckedError`, which makes sense since it was changed from false to true.

`opened` is a property, not an input

We should be able to show the datepicker opened with an input

Material/CDK 2.0.0-beta.12

DevVersion:
* Adds support for the `opened` input binding.
* Fixes no operation tests for the datepicker. 

Closes #8094

### DIFF
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -186,7 +186,10 @@ export class MatDatepicker<D> implements OnDestroy {
   @Output('closed') closedStream: EventEmitter<void> = new EventEmitter<void>();
 
   /** Whether the calendar is open. */
-  opened = false;
+  @Input()
+  get opened(): boolean { return this._opened; }
+  set opened(shouldOpen: boolean) { shouldOpen ? this.open() : this.close(); }
+  private _opened = false;
 
   /** The id for the datepicker calendar. */
   id = `mat-datepicker-${datepickerUid++}`;
@@ -277,7 +280,7 @@ export class MatDatepicker<D> implements OnDestroy {
 
   /** Open the calendar. */
   open(): void {
-    if (this.opened || this.disabled) {
+    if (this._opened || this.disabled) {
       return;
     }
     if (!this._datepickerInput) {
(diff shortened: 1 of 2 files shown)


src/lib/datepicker/datepicker.ts
```diff
@@ -168,7 +168,18 @@ export class MatDatepicker<D> implements OnDestroy {
   @Output() selectedChanged = new EventEmitter<D>();
 
   /** Whether the calendar is open. */
-  opened = false;
+  @Input()
+  get opened(): boolean { return this._opened; }
+  set opened(value: boolean) {
+    const shouldOpen = coerceBooleanProperty(value);
```
crisbeto: Since you're not assigning the `shouldOpen` anywhere, you don't really need to coerce it.

src/lib/datepicker/datepicker.ts
```diff
@@ -168,7 +168,18 @@ export class MatDatepicker<D> implements OnDestroy {
...
+
+    if (shouldOpen) {
```
crisbeto: You could shorten this one down with a ternary: `shouldOpen ? this.open() : this.close()`.

DevVersion: Done.

andrewseguin:
Got a broken test in Safari

```
Safari 9.1.2 (Mac OS X 10.11.6) MatDatepicker internationalization should have the correct input value even when inverted date format FAILED
	Expected 'Fri Sep 01 2017' to be '1.9.2017'.
```

DevVersion:
@andrewseguin This is related to the internationalization APIs of the browser. Since the tests didn't run before (at all) the issue didn't show up yet. Might need to investigate whether the browser has the `de-De` language data.

mmalerba:
Safari doesn't support the Intl API, I would just skip that test on Safari

DevVersion:
@mmalerba Yeah. I just used that `Intl_Supported`  check from the date adapter tests.

PR #8099, opened 2017-10-28
fix(datepicker): calendar overlapping input when in a fallback position

crisbeto:
* Prevents datepickers that are in a fallback position from overlapping their input by adding an offset.
* Fixes some inconsistent brace spacing.


tinayuangao:

crisbeto:
Rebased @tinayuangao.

PR #8101, opened 2017-10-28
fix(cdk-table): delay clearing view containers in ngOnDestroy

### LINKED ISSUE #8057: MatTable and router transition animations
Feature request

Table should remain in DOM until animations finish.

Table is removed from DOM as soon as routing begins.

https://table-animation.stackblitz.io/
https://stackblitz.com/edit/table-animation

When animating route transitions, navigating away from routes that contain tables becomes janky.

Angular 4+
Material 2.0.0-beta.12

Initially I thought I could return a promise from the DataSource disconnect method to delay the table being destroyed but this strategy did not work.

brandom:
* Requires angular zone to be stable for a second before clearing the view containers to allow route transition animations to complete.

* Fixes #8057

### DIFF
--- a/src/cdk/table/table.spec.ts
+++ b/src/cdk/table/table.spec.ts
@@ -12,7 +12,7 @@ import {
 } from '@angular/core';
 import {ComponentFixture, fakeAsync, flush, flushMicrotasks, TestBed} from '@angular/core/testing';
 import {BehaviorSubject, combineLatest, Observable, of as observableOf} from 'rxjs';
-import {map} from 'rxjs/operators';
+import {map, debounceTime, take} from 'rxjs/operators';
 import {CdkColumnDef} from './cell';
 import {
   CdkTableModule,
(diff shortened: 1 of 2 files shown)


brandom:
@andrewseguin I am having difficulty rewriting the failing test so that is passes. Any suggestions?

jelbourn:
@andrewseguin can you take a look with @matsko when you get back from Angular Connect?

andrewseguin:
Still trying to investigate why we manually clear at all, and how we can fix the potential memory leak otherwise

brandom:
@andrewseguin I fixed the test a while back but just got around to fixing the lint error, hah. If it turns out that manually clearing the view containers is necessary is there a chance that this will land in 5.2.0?

brandom:
@josephperrott Yes! Rebased. @andrewseguin Can you take a look?

brandom:
@jelbourn @andrewseguin I have rebased again after confirming the issue persists in the latest version (6.4.6). Could I please get some feedback if this won't be merged? If this strategy is not preferable I would be happy to investigate other solutions.

brandom:
@andrewseguin Ping, I've rebased on master again.

andrewseguin:
Closing since this solution relies on a 1000ms timeout in the chance that an animation is needed to complete before this (also this has been needing rebase for a while). Feel free to re-open a PR addressing the issue with a solution that relies on a more stable approach

PR #8103, opened 2017-10-29
docs(readme): cleanup

Splaktar:
fix broken ripples link 
use aio for getting started
nested menus are here, remove coming Q3 message
remove unused theming link


fxck:
Add link to cdk docs while you are at it? https://material.angular.io/cdk/categories

Splaktar:
@fxck good catch, done.

TheRedD0t:
I guess...
> Chip-remove, integration with input in-progress

and the "Selection and" from 

> Selection and action list planned Q3

...can be removed too, as they already have been released!?

Furthermore there are still some "planned Q3 2017" hints, that should be updated or removed, as we can't travel back in time. 😝

PR #8104, opened 2017-10-29
fix(tabs): pagination not enabled on init on some browsers

### LINKED ISSUE #7983: mat-tab overflow arrows disappear after refresh
Bug

Overflow arrows should be visible.

Overflow arrows are not visible.

- Browse to https://material.angular.io/components/tabs/
- Click "Examples" 
- Note the overflow arrows are visible.

mattabswitharros

- Refresh the page.
- Note the overflow arrows disappear. They do reappear once a tab is clicked.

mattabsnoarrows

- Note: browsing directly to "examples" also causes the overflow arrows to disappear. https://material.angular.io/components/tabs/examples

CDK: 2.0.0-beta.12
Material: 2.0.0-beta.12
CLI: 1.4.7
OS: Windows 10
TypeScript: 2.4.2
Browser: Chrome, Firefox, and Edge

crisbeto:
Fixes an issue where the tabs pagination may not be enabled on some slower browsers, because elements are being measured before they're done rendering.

Fixes #7983.

### DIFF
--- a/src/lib/tabs/tab-header.spec.ts
+++ b/src/lib/tabs/tab-header.spec.ts
@@ -243,7 +243,7 @@ describe('MatTabHeader', () => {
       });
     });
 
-    it('should re-align the ink bar when the direction changes', () => {
+    it('should re-align the ink bar when the direction changes', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


0x6368656174:
Why not accept this fix?

PR #8109, opened 2017-10-29
feat(tooltip): add injection token for specifying the default delays

### LINKED ISSUE #7928: [Tooltip] Allow app level tooltip show and hide delay configuration
#### Feature request:
Allow developers to configure global tooltip show/hide delay.

Provide InjectionTokens for easier app-level configuration of tooltip show and hide delay.

Currently, tooltip delay can be configured on per component basis using the matTooltipShowDelay and  matTooltipHideDelay attributes, but this is very verbose.

Consistency and ease of use.

Angular Material Beta 12

crisbeto:
Adds the `MAT_TOOLTIP_DEFAULT_OPTIONS` provider that allows users to set app-wide defaults for the tooltip delays.

Fixes #7928.


crisbeto:
Made the new param optional.

PR #8111, opened 2017-10-29
fix(input): add default input height

### LINKED ISSUE #8012: Alignment issue with flex-layout + datetime-local input
Bug

Elements to be aligned

Inputs aren't aligned when using flex-layout and datetime-local

Plunker starter (using on `@master`): https://goo.gl/DlHd6U<br/>
Demo: https://stackblitz.com/edit/angular-material2-issue-abcj1u
 - Add two inputs in a flex container
 - One of the inputs must be of type datetime-local

 - Elements should be aligned

Angular 4.4.3
Material Beta 12
Flex Layout Beta 9

saulotoledo:
Fixes an issue where inputs are not aligned when using flex-layout and datetime-local, because there is no default height for input fields in the code, and the default browser's height for datetime-local input fields is not the same as other inputs.

Fixes #8012

### DIFF
--- a/src/dev-app/input/input-demo.html
+++ b/src/dev-app/input/input-demo.html
@@ -55,6 +55,18 @@
           </mat-form-field>
         </td>
       </tr></table>
+      <table style="width: 100%" cellspacing="0"><tr>
+        <td>
+          <mat-form-field class="demo-full-width">
+            <input matInput type="datetime-local" placeholder="Date">
+          </mat-form-field>
+        </td>
+        <td>
+          <mat-form-field class="demo-full-width">
+            <input matInput type="number" placeholder="Duration">
+          </mat-form-field>
+        </td>
+      </tr></table>
     </form>
   </mat-card-content>
 </mat-card>
--- a/src/material/form-field/form-field-input.scss
+++ b/src/material/form-field/form-field-input.scss
@@ -70,6 +70,9 @@
   &[type='month'],
   &[type='week'],
   &[type='time'] {
+    // Hack to prevent a line break on Safari input date fields:
+    white-space: pre;
+
     &::after {
       content: ' ';
       white-space: pre;
--- a/src/material/input/_input-theme.scss
+++ b/src/material/input/_input-theme.scss
@@ -69,5 +69,11 @@
   // be misaligned w.r.t. the placeholder. Adding this margin corrects it.
   input.mat-input-element {
     margin-top: -$line-spacing * 1em;
+    height: $line-height * 1em;
+  }
+
+  input.mat-input-safari::first-line {
+    // Hack for 1px input content clipping on Safari:
+    line-height: $line-height * 1em;
   }
 }
--- a/src/material/input/input.ts
+++ b/src/material/input/input.ts
@@ -73,6 +73,7 @@ const _MatInputMixinBase: CanUpdateErrorStateCtor & typeof MatInputBase =
      */
     'class': 'mat-input-element mat-form-field-autofill-control',
     '[class.mat-input-server]': '_isServer',
+    '[class.mat-input-safari]': '_isSafari',
     // Native input properties that are overwritten by Angular inputs need to be synced with
     // the native input element. Otherwise property bindings for those don't work.
     '[attr.id]': 'id',
@@ -96,6 +97,9 @@ export class MatInput extends _MatInputMixinBase implements MatFormFieldControl<
   private _inputValueAccessor: {value: any};
   /** The aria-describedby attribute on the input for improved a11y. */
   _ariaDescribedby: string;
+
+  /** Whether the component is being rendered in Safari. */
+  _isSafari = false;
 
   /** Whether the component is being rendered on the server. */
   _isServer = false;
@@ -265,6 +269,7 @@ export class MatInput extends _MatInputMixinBase implements MatFormFieldControl<
       });
     }
 
+    this._isSafari = this._platform.SAFARI;
     this._isServer = !this._platform.isBrowser;
     this._isNativeSelect = element.nodeName.toLowerCase() === 'select';
 
--- a/tools/public_api_guard/material/input.d.ts
+++ b/tools/public_api_guard/material/input.d.ts
@@ -10,6 +10,7 @@ export declare class MatInput extends _MatInputMixinBase implements MatFormField
     protected _elementRef: ElementRef<HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement>;
     protected _id: string;
     _isNativeSelect: boolean;
+    _isSafari: boolean;
     _isServer: boolean;
     protected _neverEmptyInputTypes: string[];
     protected _platform: Platform;


src/lib/input/input.scss
```diff
@@ -13,12 +13,13 @@
...
   color: currentColor;
 
-  // By default, <input> has a padding, border, outline and a default width.
+  // By default, <input> has a padding, border, outline, a default width and a default height.
   border: none;
   outline: none;
   padding: 0;
   margin: 0;
   width: 100%;
+  height: $mat-input-height;
```
mmalerba: I don't think we want to do this, the input is supposed to adapt to the font size of its parent, and this will probably break it.

saulotoledo: Good point. What about a min-height? I can update the PR with that.

mmalerba: I still think that might be problematic. It will prevent the element from shrinking when the `font-size` is smaller. You could try setting `height: $line-height * 1em` here: https://github.com/angular/material2/blob/master/src/lib/input/_input-theme.scss#L27

saulotoledo: Much better. I am testing right now and will update the PR in the next couple of minutes.

saulotoledo: I had a problem with the screenshot tests. I did a rebase with the master branch and it worked fine except that the other tests (especially e2e) are not working. Did I forget something? Or is it just a temporary instability in the master branch?

mmalerba: Looking at the demo, it seems to cause an issue in safari where the bottom of the text is clipped off. I played around with it a little bit and it seemed like maybe setting the `line-height` of the `:first-line` will help, but it'll need to be tested across all browsers

saulotoledo: Sorry for the delay. I am traveling, but I am looking on it. I will update the PR as soon as possible.

saulotoledo: I finally returned. This is what I see in safari. Could you help me to replicate the text clipping?

safari_image

mmalerba: Sorry for the slow response, take a look at these 2 demos on Safari:
https://mmalerba-demo1.firebaseapp.com/input (master)
https://mmalerba-demo2.firebaseapp.com/input (master + your pr)

Notice for the first input (the disabled one) that the 'g' in 'Google' is slightly clipped on the bottom with your PR.

saulotoledo: Sorry for the delay, @mmalerba. I was trying the `line-height` approach, but I discovered that setting the border changes the baseline in Safari, so I tried another approach and I would like to know your opinion about.

mmalerba: I would prefer to avoid using `border` if possible, because even though its transparent it still shows up solid in high contrast mode on windows. (We actually do use a border hack somewhere else in the form field CSS, but we've been trying to remove it: https://github.com/angular/material2/pull/6291)

saulotoledo: @mmalerba Sorry again for the delay. I had to find a device with Safari for testing. I think the current version of the PR will solve the problem.

mmalerba:
Seems to be causing a duplicated placeholder issue (see https://mmalerba-demo8.firebaseapp.com/input)

saulotoledo:
Yes, it is... I will work on that and will let you know.

saulotoledo:
@mmalerba, I think I will have an update tomorrow. I saw your change in the importance of this solution. I am just waiting a device for testing. I do not have a mac available, so I'm taking too long to update this PR. But I want you know that I will have a device available tomorrow, then I will submit the updates.

mmalerba:
cool, no worries, I just updated the priority since I realized it affected more types of inputs than I initially realized.

mmalerba:
@saulotoledo are you still working on this?

saulotoledo:
@mmalerba Hi, yes. I tried some approaches, but I had some issues with older IE versions. But my problem is the device's availability for Safari. I have another to borrow this weekend. How much time can you give me yet?

mmalerba:
Still no rush, I just wanted to confirm if you're still working on it

saulotoledo:
Hi @mmalerba. I did a rebase before starting to clean up my code to submit. I just noticed that my fixes were not necessary. I removed them and returned to the previous version of the code; sounds that the duplicated placeholder is no longer a problem. If you have some time, could you take another look at my code? In the meantime I will keep looking for issues.

mmalerba:
@saulotoledo The placeholder issue may have been indirectly fixed by some of the changes we've been making to support the `appearance=fill` and `appearance=outline` versions of the form-field. But the underlying problem of the `<input>` being a different height still seems to exist. It can be seen when you try to align inputs of different types. For example:

```html
<div style="display:flex">
  <mat-form-field>
    <input matInput type="datetime-local" placeholder="Date">
  </mat-form-field>
  <mat-form-field>
    <input matInput type="number" placeholder="Duration" >
  </mat-form-field>
</div>
```

Results in this:
form-field-alignment

If your solution helps with this, it would be good to still submit it.

saulotoledo:
@mmalerba The current changes solves that. There is a a demo in the following link:
https://saulotoledo-ng-material2-demo1.firebaseapp.com/input

Let me know if I need to make changes.

mmalerba:
Looks great on Chrome, I'll test it out on a few other browsers after you fix the Safari bug. Thanks for the work on this!

saulotoledo:
@mmalerba, the last problem is fixed for Safari, but Microsoft Edge crashes while using `::first-line` (used for solving that clipping problem in Safari. You can see on Travis, and I was able to reproduce in a Windows Machine). I am trying to find another solution for that.

saulotoledo:
Hi @mmalerba. As I said in my previous comment some minutes ago, Microsoft Edge will crash if we have `::first-line` for inputs in the code. Sounds like an issue on Edge (you can also check the previous travis builds to see that it also fails, [like here](https://travis-ci.org/angular/material2/jobs/355171408)). While trying to find a solution, I updated my PR to contain a `padding` hack. I did not like [that magic number there](https://github.com/angular/material2/pull/8111/files#diff-002aa63d52575ed15761f1e16e547059R51), but this is the straightforward solution I have found to solve the problem on Safari and on every other web browser. I would like to hear from you what do you think about.

Preview:
[https://saulotoledo-ng-material2-demo2.firebaseapp.com/input](https://saulotoledo-ng-material2-demo2.firebaseapp.com/input)

mmalerba:
Hmm the 1px padding seems to through off the baseline a bit. What about this instead?  Use the [safari check in @angular/cdk/platform](https://github.com/angular/material2/blob/master/src/cdk/platform/platform.ts#L60) to apply a specail class to the input if the browser is safari (e.g. `mat-input-safari`) then target your rule at that instead of the generic `mat-input-element`

saulotoledo:
Hi @mmalerba. Thanks for the feedback. Could you confirm if my last changes are what you are expecting?

I did a rebase with the master branch, but sounds something is broken in the master and my branch is not working. But tomorrow I will do another rebase and upload a new live preview.

src/lib/input/_input-theme.scss
```diff
@@ -45,5 +45,11 @@
   // be misaligned w.r.t. the placeholder. Adding this margin corrects it.
   input.mat-input-element {
     margin-top: -$line-spacing * 1em;
+    height: $line-height * 1em;
+  }
+
+  input.mat-input-safari {
+    // Hack for 1px input content clipping on Safari:
+    padding-bottom: 1px;
```
mmalerba: I was actually thinking we could do the `::first-line` thing here since the `padding` messes up the baseline alignment

src/lib/input/input.ts
```diff
@@ -245,6 +245,10 @@ export class MatInput extends _MatInputMixinBase implements MatFormFieldControl<
       });
     }
 
+    if (_platform.SAFARI) {
+      _elementRef.nativeElement.classList.add(this.controlType + '-safari');
```
mmalerba: Can you set a member variable like `this._isSafari = _platform.SAFARI` then you can do a host binding to set the class `'[class.mat-input-safari]': '_isSafari'`.

src/lib/input/input.scss
```diff
@@ -52,6 +52,9 @@
   &[type='month'],
   &[type='week'],
   &[type='time'] {
+    // Hack to prevent a line break on Safari input date fields:
+    white-space: pre;
```
mmalerba: We should check if this is still necessary after switching back to the `::first-line` method

src/lib/input/input.ts
saulotoledo: Great! It is done.

src/lib/input/_input-theme.scss
saulotoledo: You're right. I've updated the PR with the `::first-line` approach.

src/lib/input/input.scss
mmalerba: Both of these look fine on desktop Safari (11.0.3). I'll see if I can steal someone's iPhone tomorrow to test mobile Safari. I'll lgtm for now, and add merge ready after we confirm if the hack is needed.

saulotoledo: Here are some screenshots taken by BrowserStack. I've reduced the input width in order to highlight the bug (look at the date field).

bs_realios_mobile_iphone 6s-9 0 without pre reduced

bs_realios_mobile_iphone 6s-9 0 with pre reduced

Also, after a rebase I noticed that the duplicate label issue has returned. Should I add a fix in my PR? Or should I wait again for an external fix as before?

saulotoledo: PS.: returned for the Safari version running from BrowserStack. I've unable to reproduce in other browsers.

saulotoledo: A last comment: I could not replicate the error at build with Edge. I dont think that is related, but I want to ask you: is it possible that the problem is related to something in my PR? At least in my tests the input page is working on Edge.

mmalerba: Its probably a flake, I've restarted the tests

mmalerba: Confirmed that the hack is still needed for iPad

saulotoledo: It failed on Safari now oO'.

Thank you very much for your help and patience with this PR.

mmalerba: Yeah, I'll just rerun it till it passes, definitely seems like flakiness. Same to you, thanks for sticking with this!

saulotoledo: @mmalerba Thanks. Are you available on Gitter?

mmalerba: I've never actually signed in before, but I probably should. Just joined

mmalerba:
There seems to be some unexpected screenshot diffs when trying to sync this into google3, needs to be investigated.

mmalerba:
@saulotoledo If you want to try it out the change in an app and see if it causes differences feel free. The screenshot diffs I'm talking about are for internal Google applications, so a Googler will need to investigate. It's not even necessarily anything wrong with the PR, could be that someone has global styles for `input { height: 1em; }` or something like that.

saulotoledo:
Hi @mmalerba, I just rebased the branch to solve the merge conflict at `src/lib/input/input.ts`. :)

jelbourn:
@mmalerba to revisit this PR to see if it's still relevant

mmalerba:
I think this one is just infeasible to land. It causes too many broken looking screenshots inside Google.

Issue #8082, opened 2017-10-27
Document control attributes for cdkTrapFocus


jelbourn:
Currently `cdkFocusRegionStart`, `cdkFocusInitial`, and `cdkFocusRegionEnd` (see #6960) are undocumented.

willshowell:
I think this can be closed

PR #8128, opened 2017-10-30
docs(focus): Add docs for regions #8082

amcdnl:
Addresses #8082

### DIFF
--- a/src/cdk/a11y/a11y.md
+++ b/src/cdk/a11y/a11y.md
@@ -71,6 +71,10 @@ interaction.
 #### Regions
 Regions can be declared explicitly with an initial focus element by using 
 the `cdkFocusRegionStart`, `cdkFocusRegionEnd` and `cdkFocusInitial` DOM attributes. 
+`cdkFocusInitial` specifies the element that will receive focus upon initialization of the region. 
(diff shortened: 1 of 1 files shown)


src/cdk/a11y/a11y.md
```diff
@@ -68,6 +68,18 @@ This directive is declared in `A11yModule`.
...
+the `cdkFocusRegionStart`, `cdkFocusRegionEnd` and `cdkFocusInitial` DOM attributes. 
+For example:
+
+```html
+<a mat-list-item routerLink cdkFocusRegionStart>Focus region start</a>
+<a mat-list-item routerLink>Link</a>
+<a mat-list-item routerLink cdkFocusInitial>Initially focused</a>
+<a mat-list-item routerLink cdkFocusRegionEnd>Focus region end</a>
+```
+
```
jelbourn: We could be a bit more explicit about what each does:
"`cdkFocusInitial` specifies the element that will receive focus upon initialization of the region"

"`cdkFocusRegionStart` and `cdkFocusRegionEnd` define the region within which focus will be trapped. When using the tab key, focus will move through this region and wrap around on either end."

PR #8132, opened 2017-10-30
fix(block-scroll-strategy): disable smooth scrolling before restoring scroll position

### LINKED ISSUE #7139: Overwrite scroll-behavior when setting scroll-position
When a scroll-behavior is present and Material wants to set (not smoothly scroll to) a scroll-position, we should overwrite that css-property while changing the scroll-position.

That I wont notive

I get a weird scrolling feeling

add scroll-behavior:smooth; to html element
scroll down and open a dialog. Close the dialog and see the smooth scrolling happen...

example plunkr:
https://plnkr.co/edit/kQNavjkozevBalxAq785?p=preview

New browser feature

all

smooth-scrolling is not supported everywhere yet, so use newest chrome to reproduce

crisbeto:
Disables user-defined `scroll-behavior: smooth` before restoring the scroll position when disabling the `BlockScrollStrategy`, which prevents browsers from animating the scroll position change.

Fixes #7139.

### DIFF
--- a/src/cdk/overlay/scroll/block-scroll-strategy.spec.ts
+++ b/src/cdk/overlay/scroll/block-scroll-strategy.spec.ts
@@ -136,6 +136,27 @@ describe('BlockScrollStrategy', () => {
     expect(document.documentElement.getBoundingClientRect().width).toBe(previousContentWidth);
   });
 
+  it('should not clobber user-defined scroll-behavior', skipIOS(() => {
+    const root = document.documentElement;
+    const body = document.body;
+
+    root.style['scrollBehavior'] = body.style['scrollBehavior'] = 'smooth';
+
+    // Get the value via the style declaration in order to
+    // handle browsers that don't support the property yet.
+    const initialRootValue = root.style['scrollBehavior'];
+    const initialBodyValue = root.style['scrollBehavior'];
+
+    overlayRef.attach(componentPortal);
+    overlayRef.detach();
+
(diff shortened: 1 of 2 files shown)


src/cdk/overlay/scroll/block-scroll-strategy.ts
```diff
@@ -45,11 +45,25 @@ export class BlockScrollStrategy implements ScrollStrategy {
   /** Unblocks page-level scroll while the attached overlay is open. */
   disable() {
     if (this._isEnabled) {
+      const html = document.documentElement;
+      const body = document.body;
+      const previousHTMLScrollBehavior = html.style['scrollBehavior'] || '';
```
jelbourn: `previousHtmlScrollBehavior`

src/cdk/overlay/scroll/block-scroll-strategy.spec.ts
```diff
@@ -136,6 +136,24 @@ describe('BlockScrollStrategy', () => {
     expect(document.documentElement.getBoundingClientRect().width).toBe(previousContentWidth);
   });
 
+  it('should not clobber user-defined scroll-behavior', skipIOS(() => {
+    const root = document.documentElement;
+    const body = document.body;
+
+    root.style['scrollBehavior'] = body.style['scrollBehavior'] = 'smooth';
```
jelbourn: Need to reset this after the test?

crisbeto:

PR #8136, opened 2017-10-30
chore: fix lint error

crisbeto:
Fixes a few linting errors


jelbourn:
Mared as safe because the only non-test change is adding a semi-colon

PR #8138, opened 2017-10-30
fix(autocomplete): top option group not scrolled into view when going up

crisbeto:
Fixes the case where the user won't be able to see the top option group when reaching the first option via the arrow keys.


crisbeto:
I'd say so @josephperrott. Currently when you're going from below and you hit the top option, you can't see which group it's a part of.

jelbourn:
@crisbeto to either rebase or re-open this PR

crisbeto:
Closing in favor of https://github.com/angular/components/pull/16343.

PR #8139, opened 2017-10-30
fix(spinner): set initial value for spinner to 0

### LINKED ISSUE #8126: Progress Spinner initialization bug

Bug

No "visual glitch" at the progress spinner initialization

During a short period of time (1 second) the value attribute of the progress spinner is ignored and, the spinner start from almost "full" and go "back" to the correct initial value

https://plnkr.co/edit/vSTe3nqLuA1xiKSjzT1Z?p=preview

angular material 2.0.0-beta.12

The issue appears only when angular flex-layout is imported (even if it is not used).

josephperrott:
Fixes #8126

### DIFF
--- a/src/lib/progress-spinner/progress-spinner.spec.ts
+++ b/src/lib/progress-spinner/progress-spinner.spec.ts
@@ -37,12 +37,12 @@ describe('MatProgressSpinner', () => {
     expect(progressElement.componentInstance.mode).toBe('indeterminate');
   });
 
-  it('should define a default value of undefined for the value attribute', () => {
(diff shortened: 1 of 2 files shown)


src/lib/progress-spinner/progress-spinner.spec.ts
```diff
@@ -4,7 +4,7 @@ import {By} from '@angular/platform-browser';
 import {MatProgressSpinnerModule} from './index';
 
 
-describe('MatProgressSpinner', () => {
+fdescribe('MatProgressSpinner', () => {
```
jelbourn: `fdescribe`

PR #8160, opened 2017-10-31
fix: import individual rx operators

crisbeto:
After the switch to lettable operators, we had started importing all operators via `rxjs/operators`. These changes switch us back to only importing the individual ones and adds some linting so it doesn't happen in the future.


andrewseguin:
LGTM - thanks for doing all the work

sgwatgit:
@crisbeto After reading about the option to us lettable operators in the [ng 5 annoucement blog](https://blog.angular.io/version-5-0-0-of-angular-now-available-37e414935ced) I was wondering why the ng Material team has chosen to use the deep lettable operators compared to the ones mentioned in the blog or in the [rxjs repo docs ](https://github.com/ReactiveX/rxjs/blob/master/doc/lettable-operators.md).

Does the approach you've chosen work better at scale, does it cause fewer issues with build systems or is it to support older versions of Typescript?

Just curious! Thanks!

danwulff:
@crisbeto @jelbourn. Could you take a look at @sgwatgit 's question. I'm also curious about this change, especially as a maintainer of an angular library.

PR #8162, opened 2017-10-31
docs(table): updated docs for the table

andrewseguin:
@willshowell 

To be added later: expanding rows (using when predicate), customizing column widths, adding/removing data, dynamic columns (add/remove/move)

### DIFF
--- a/src/lib/table/table.md
+++ b/src/lib/table/table.md
@@ -1,20 +1,30 @@
-The `mat-table` provides a Material Design styled data-table that can be used to display rows of data. 
+The `mat-table` provides a Material Design styled data-table that can be used to display rows of 
+data. 
 
-The table’s template is made up of column definitions where each definition provides the content for that column’s header and row cells. It also includes a `mat-header-row` and `mat-row` where each takes an ordered list of the columns they should render out.
+The table's template is made up of column definitions where each definition provides the content for 
+that column's header and row cells. It also includes a `<mat-header-row>` and `<mat-row>` where each 
+takes an ordered list of the columns they should render.
 
-Passing data to the table must be done through a `DataSource`, which connects to the table by providing an Observable that emits an array of data. The table picks up this data array and writes a `mat-row` for each data object in the array. It is the responsibility of the `DataSource` to send exactly what data should be rendered by the table, so any data manipulation features such as sorting or filtering should be implemented here.
+Passing data to the table must be done through a `DataSource`, which connects to the table by 
+providing an Observable that emits an array of data. The table picks up this data array and writes 
+a `mat-row` for each data object in the array. It is the responsibility of the `DataSource` to send 
+exactly what data should be rendered by the table, so any data manipulation features such as 
+sorting or filtering should be implemented here.
 
 This table builds on the foundation of the CDK data-table and uses a similar interface for its
 data source input and template, except that its element and attribute selectors will be prefixed 
-with `mat-` instead of `cdk-`. For detailed information on the interface and how it works, see the [guide covering the CDK data-table](https://material.angular.io/guide/cdk-table).
+with `mat-` instead of `cdk-`. For detailed information on the interface and how it works, see the 
+[guide covering the CDK data-table](https://material.angular.io/guide/cdk-table).
 
 ### Getting Started
 
 <!-- example(table-basic) -->
 
-Start by writing your table’s column definitions. Each column definition should be given a unique name and contain the content definitions for its header and row cells.
+Start by writing your table's column definitions. Each column definition should be given a unique 
+name and contain the content definitions for its header and row cells.
 
-Here’s a simple column definition with the name `’userName’`. The header cell contains the text ‘Name’ and each row cell will render the `name` property of each row's data.
+Here's a simple column definition with the name `'userName'`. The header cell contains the text 
+‘Name' and each row cell will render the `name` property of each row's data.
 
 ```html
 <ng-container matColumnDef="userName">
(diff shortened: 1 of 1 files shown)


src/lib/table/table.md
```diff
@@ -1,50 +1,146 @@
...
+<mat-row *matRowDef="let myRowData; columns: [‘userName’, ‘age’]"></mat-row>
+```
+
+Once your template is set up, the final step is to provide data to the table. This must be done through a `DataSource`, which uses an Observable stream that emits an array of data that the table should render. 
+
+To get started, you can use the pre-built `MatTableDataSource` to handle the stream for you. Just create a new instance and set its `data` property to whatever data you want the table to render.
+
+```ts
+const myDataSource = new MatTableDataSource();
+myDataSource.data = dataToRender;
```
willshowell: Change to emphasize the data source is a class property?

```ts
this.myDataSource = new MatTableDataSource();
this.myDataSource.data = dataToRender;
```

src/lib/table/table.md
```diff
@@ -1,50 +1,146 @@
...
-interface, material styling, array input, and more out-of-the-box features (sorting, pagination,
-and selection).
+#### Selection
+
+To add row selection to the table, first set up a `SelectionModel` from `@angular/cdk/collections` that will maintain the selection state.
+
+```js
+const initialSelection = [];
+const allowMultiSelect = true;
+const selection = new SelectionModel<MyDataType>(allowMultiSelect, initialSelection);
```
willshowell: Same thing here since `selection` should be a class property

src/lib/table/table.md
```diff
@@ -1,50 +1,146 @@
...
+```
+
+Add a column definition for displaying the row checkboxes, including a master toggle checkbox for the header. The column name should be added to the list of displayed columns provided to the `mat-header-row` and `mat-row`.
+
+```html
+<ng-container matColumnDef="select">
+ <mat-header-cell *matHeaderCellDef>
```
willshowell: Indentation off by 1

src/lib/table/table.md
```diff
@@ -1,50 +1,146 @@
...
+     this.selection.clear() :
+     this.dataSource.data.forEach(row => this.selection.select(row));
+}
+```
+
+Finally, adjust the styling for the select column so that its overflow is not hidden. This allows the ripple effect to extend beyond the cell.
+
+```css
+.mat-column-select {
+ overflow: initial;
```
willshowell: Indentation off by 1

src/lib/table/table.md
```diff
@@ -1,50 +1,146 @@
-The `mat-table` provides a Material Design styled data-table that can be used to display rows of
-data.
+The `mat-table` provides a Material Design styled data-table that can be used to display rows of data. 
+
+The table’s template is made up of column definitions where each definition provides the content for that column’s header and row cells. It also includes a `mat-header-row` and `mat-row` where each takes an ordered list of the columns they should render out.
```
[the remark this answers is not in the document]
willshowell: Also consider making `<mat-header-row>` and `<mat-row>`

andrewseguin:
Thanks for the review - made the changes. Wasn't sure what you mean about breaking newlines

src/lib/table/table.md
```diff
@@ -1,50 +1,185 @@
...
+  <mat-cell *matCellDef="let user"> {{user.name}} </mat-cell>
+</ng-container>
+```
+
+After you define all your columns, you'll need to provide the header and data row templates that 
+will be rendered out by the table. Each template should be given an ordered list of columns that 
+will let the table know what columns each row should render.
+
+```html
+<mat-header-row *matHeaderRowDef="[‘userName', ‘age']"></mat-header-row>
```
willshowell: Looks like only the closing ones have changed. There are a few remaining opening single-quotes that are styled.

andrewseguin: Ugh, thanks Google Docs

src/lib/table/table.md
```diff
@@ -1,50 +1,185 @@
...
+
+The table's template is made up of column definitions where each definition provides the content for 
+that column's header and row cells. It also includes a `<mat-header-row>` and `<mat-row>` where each 
+takes an ordered list of the columns they should render.
+
+Passing data to the table must be done through a `DataSource`, which connects to the table by 
+providing an Observable that emits an array of data. The table picks up this data array and writes 
+a `mat-row` for each data object in the array. It is the responsibility of the `DataSource` to send 
+exactly what data should be rendered by the table, so any data manipulation features such as 
+sorting or filtering should be implemented here.
```
jelbourn: Trying to tighten the language a bit, avoiding passive voice:

``` markdown
A `DataSource` provides data to the table by emitting an `Observable` stream of
the items to be rendered. Each emit includes the _entire set of items_ that being
displayed. The table, listening to this stream, will render a row per item. Any
manipulation of the data being displayed (e.g. sorting, pagination, filtering)
should be captured by the `DataSource`, ultimately emitting a new set of
items to reflect any changes.
```

src/lib/table/table.md
```diff
@@ -1,50 +1,185 @@
...
-For more information on the interface and how it works, see the
+with `mat-` instead of `cdk-`. For detailed information on the interface and how it works, see the 
 [guide covering the CDK data-table](https://material.angular.io/guide/cdk-table).
 
+### Getting Started
+
 <!-- example(table-basic) -->
 
-### Features
+Start by writing your table's column definitions. Each column definition should be given a unique
```
jelbourn: Would numbered subheaders help here?
```markdown
#### 1. Define the table's columns
...
#### 2. Define the table's rows
...
#### 3. Provide data
...
```

src/lib/table/table.md
```diff
@@ -1,50 +1,185 @@
...
+After you define all your columns, you'll need to provide the header and data row templates that 
+will be rendered out by the table. Each template should be given an ordered list of columns that 
+will let the table know what columns each row should render.
+
+```html
+<mat-header-row *matHeaderRowDef="['userName', 'age']"></mat-header-row>
+<mat-row *matRowDef="let myRowData; columns: ['userName', 'age']"></mat-row>
+```
+
+Once your template is set up, the final step is to provide data to the table. This must be done
```
jelbourn: Maybe something like
```markdown
The column and row definitions now capture _how_ data will render- all that's
left is to provide the data itself. For simple scenarios with client-side operations,
`MatTableDataSource` offers a quick and easy starting point. Simply create an
instance of `MatTableDataSource` and set the items to be displayed to the
`data` property. For more advanced scenarios, applications will implement
one or more custom `DataSource` to capture more specific behaviors.
```

src/lib/table/table.md
```diff
@@ -1,50 +1,185 @@
...
 #### Pagination
 
-The `<mat-paginator>` adds a pagination UI that can be used in conjunction with the `<mat-table>`. The
-paginator emits events that can be used to trigger an update via the table's data source.
+To paginate the table's data, add a `<mat-paginator>` after the `<mat-table>` and provide the 
+`MatPaginator` to the `MatTableDataSource`. The data source will automatically listen for page 
+changes made by the user and send the right paged data to the table.
+
+For more information on using and configuring the `<mat-paginator>`, check out the 
+[mat-paginator docs](https://material.angular.io/components/paginator/overview).
```
jelbourn: We should mention here that `mat-paginator` is just one option. The table can work with any paginator control or any pagination strategy.

(same for sorting, filtering, and selection)

src/lib/table/table.md
```diff
@@ -1,50 +1,185 @@
...
 
+For more information on using and configuring the sorting behavior, check out the 
+[matSort docs](https://material.angular.io/components/sort/overview).
+
 #### Filtering
 
-While Angular Material does not offer a specific component for filtering tabular data, the table's
-data source can be updated based on any custom filter UI. Any filtering pattern need only trigger
-an update via the table's data source.
+To remove filtered rows from the table's data, simply provide the filter string to the
```
jelbourn: This section is too specifically about the `MatTableDataSource`. We should mention that we don't provide any filtering UI components and that filtering is up to the app. Then `MatTableDataSource` can be the "if you want something simple" extra thing.

src/lib/table/table.md
```diff
@@ -1,50 +1,185 @@
...
+match.
+
+<!--- example(table-filtering) -->
 
-In the near future, we will provide a simplified version of the data-table with an easy-to-use
-interface, material styling, array input, and more out-of-the-box features (sorting, pagination,
-and selection).
+#### Selection
+
+To add row selection to the table, first set up a `SelectionModel` from `@angular/cdk/collections`
```
jelbourn: Same comment here about specificity- I want to keep it clear that users can do selection however they want. If we present it as "this is the way you are supposed to do selection", it begs the question as to why it's not just built in. I would really rather include this in an example with explanation that say "Here is the One True Way to do selection" (same for filtering)

src/lib/table/table.md
walhow: Flexibility is great, but this documentation is under angular/material2. Most developers won't have time to try a lot of different options. Would still be good to provide example code for MatTableDataSource.

src/lib/table/table.md
walhow: Flexibility is nice, but most developers are just going to want a working example they can start with.

PR #8167, opened 2017-10-31
Update autocomplete-filter-example.ts

atscott:
Simplify the example. Starting with empty string rather than null allows for removing the null check logic.


src/material-examples/autocomplete-filter/autocomplete-filter-example.ts
```diff
@@ -26,8 +26,8 @@ export class AutocompleteFilterExample {
 
    ngOnInit() {
       this.filteredOptions = this.myControl.valueChanges
-         .startWith(null)
-         .map(val => val ? this.filter(val) : this.options.slice());
+         .startWith('')
+         .map(val => this.filter(val));
```
atscott: Could also be `this.filter(val || '')`

Issue #2800, opened 2017-01-25
Add elevation guide


Splaktar:
#### Bug, feature request, or proposal:
proposal

#### What is the expected behavior?
That `.md-elevation-zX` classes are documented.

#### What is the current behavior?
No documentation exists other than within closed [PR](https://github.com/angular/material2/pull/2212) and [issue](https://github.com/angular/material2/pull/222/files) comments. There are some [design documents](https://docs.google.com/document/d/1W3NGSLqDZzjbBBLW2C6y_6NUxtvdZAVaJvg58LY3Q0E/edit?usp=sharing), but they aren't really designed for developers using the APIs.

#### What is the use-case or motivation for changing an existing behavior?
To avoid repeated [questions](https://github.com/angular/material2/pull/222#issuecomment-243015295) and rejected PRs like https://github.com/angular/material2/pull/2212.

#### Is there anything else we should know?
I would like to help with this documentation, but I would like some feedback on how the team would like this to be done before I submit a PR.

Should there be a separate doc/guide for elevation that covers examples of multiple components?

Should there be examples of using elevation in each component that it applies to?
Is there a list of such components?

jelbourn:
It would be great to have a short and simple guide on elevation in the `guides/` directory.

ladyleet:
agreed would be fabulous! // @EladBezalel (thx @Splaktar!) I like the simplicity of MaterializeCSS docs http://materializecss.com/shadow.html

farantesrodrigues:
The mixins should also be documented like ```@include md-elevation($zValue);``` (the design documents this is ```mdElevation($zValue)```)

rviveiros:
Now we should use   @include mat-elevation($zValue);

M-a-c:
@Splaktar Do you have a Pr in progress that I could help with or anything like that?

Splaktar:
@M-a-c no sorry. I had other priorities jump up and haven't had a chance to get back to this. Thank you for helping out!

megamindbrian:
I have a problem with hundreds of elevation directives being injected in for EVERY component.  Every components has CSS styles like this:
```
.mat-elevation-z0 {
  box-shadow: 0px 0px 0px 0px rgba(0, 0, 0, 0.2), 0px 0px 0px 0px rgba(0, 0, 0, 0.14), 0px 0px 0px 0px rgba(0, 0, 0, 0.12); }

.mat-elevation-z1 {
[...]

PR #3917, opened 2017-04-05
Docs (Guides): Mat elevation guide

M-a-c:
Added a basic guide on how to use Mat Elevation.
Includes Example
* Link to Material designs spec for Elevation
* How to use the CSS class
* How to use the mixin
* Deprecated naming

Fixes #2800

PR #8172, opened 2017-11-01
docs(elevation): add guide for elevation helpers

willshowell:
Credit to @M-a-c for initial version (https://github.com/angular/material2/pull/3917)

Fixes https://github.com/angular/material2/issues/2800

cc @Splaktar @mmalerba

### DIFF
--- a/guides/elevation.md
+++ b/guides/elevation.md
@@ -48,5 +48,3 @@ elevation changes:
   }
 }
 ```
-
-
--- a/guides/elevation.md
+++ b/guides/elevation.md
@@ -1,0 +1,52 @@
+Angular Material's elevation classes and mixins allow you to add separation between elements along
+the z-axis. All material design elements have resting elevations. In addition, some elements may
+change their elevation in response to user interaction. The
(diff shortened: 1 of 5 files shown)


guides/elevation.md
```diff
@@ -0,0 +1,50 @@
...
+```scss
+@import '~@angular/material/theming';
+
+.my-class {
+  @include mat-elevation(2);
+}
+```
+
+For convenience, you can use the `mat-elevation-transition` mixin to add a transition between
+elevation changes:
```
mmalerba: ... add a transition when the elevation changes

guides/elevation.md
```diff
@@ -0,0 +1,50 @@
...
+
+.my-class {
+  @include mat-elevation-transition;
+  @include mat-elevation(2);
+
+  &:active {
+    @include mat-elevation(8);
+  }
+}
+```
```
mmalerba: should we add an embedded example here too?

willshowell: It won't export very well to plunker will it?

mmalerba: Oh right, I forgot our examples use plain CSS

PR #8174, opened 2017-11-01
build: fix screenshot-tool app deployment

DevVersion:
* Fixes mismatching dependencies that lead to build errors inside of the screenshot-tool app.

**Note**: Hopefully the CI won't install any other dependencies once it's merged. Locally it works fine and with a lock-file it may work more safely too. But not sure its worth pushing another lock file.


andrewseguin:
Needs rebase

PR #8177, opened 2017-11-01
chore: bump to Angular 5 final

crisbeto:
Currently the Angular 5 dependency is set to rc.9. Bumps it to the final version.


tools/screenshot-test/package.json
```diff
@@ -28,7 +28,7 @@
   },
   "devDependencies": {
     "@angular/cli": "^1.0.0",
-    "@angular/compiler-cli": "~5.0.0-rc.9",
+    "@angular/compiler-cli": "^5.0.0",
     "@types/node": "^6.0.42",
     "ts-node": "1.2.1",
     "typescript": "~2.2.1"
```
vitaly-t: Shouldn't we up the TypeScript to 2.5.3 at this point? :wink:

princemaple: TS should at least be 2.3? Since ng 5 uses some feature from it.

crisbeto: There's another PR that fixes the TS version for the screenshot tool.

PR #8183, opened 2017-11-01
fix(snack-bar): complete onAction observable on close

### LINKED ISSUE #8181: [snack-bar] onAction subject never completes
Bug

Subscribing to `onAction()` will result in a subscription that never completes unless the action button is clicked.

#### Expected behavior

`onAction()` completes when the snackbar ref is destroyed, I suppose in `_finishDismiss()`.

#### Reproduction

https://stackblitz.com/edit/angular-material2-issue-l1ymh2?file=app/app.component.ts

#### Motivation

I wish to open a snack-bar that has a duration longer than the expected lifetime of the component opening it. When the action is clicked, I would like to navigate to a new route.

I am unable to unsubscribe from the subscription in `ngOnDestroy` since the snackbar action may be available after the component is destroyed.

crisbeto:
* Completes the `onAction` observable when a snack bar is closed, no matter whether it was through an action.
* Refactors a few snack bar unit tests to use spies instead of toggling booleans.

Fixes #8181.

### DIFF
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -79,6 +79,11 @@ export class MatSnackBarRef<T> {
   /** Cleans up the DOM after closing. */
   private _finishDismiss(): void {
     this._overlayRef.dispose();
+
+    if (!this._onAction.closed) {
+      this._onAction.complete();
+    }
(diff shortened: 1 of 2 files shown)


josephperrott:
Looks like there is one lint error.

PR #8188, opened 2017-11-02
build: enable noUnusedParameters check in aot.

DevVersion:
* Re-enables the `noUnusedParameters` check in the e2eapp and demo-app (AOT).

### DIFF
--- a/src/universal-app/tsconfig-prerender.json
+++ b/src/universal-app/tsconfig-prerender.json
@@ -4,7 +4,6 @@
     "declaration": false,
     "stripInternal": false,
     "experimentalDecorators": true,
-    // TODO(paul): Remove once Angular has been upgraded and supports noUnusedParameters in AOT.
     "noUnusedParameters": true,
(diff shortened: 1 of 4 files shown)


crisbeto:
Should be able to enable this one as well https://github.com/angular/material2/blob/dfe8091d959f1bb54cc33a8d942cbe7ea4e3fe56/src/universal-app/tsconfig-build.json#L8

DevVersion:
Done.

src/universal-app/tsconfig-prerender.json
```diff
@@ -5,7 +5,7 @@
     "stripInternal": false,
     "experimentalDecorators": true,
     // TODO(paul): Remove once Angular has been upgraded and supports noUnusedParameters in AOT.
```
josephperrott: Should this comment be removed as it was in the other json files?

DevVersion: Good point. Done

PR #8199, opened 2017-11-02
feat(expansion): allow expansion indicator positioning.

### LINKED ISSUE #6726: Allow md-expansion-panel toggle to be placed at the beginning

josephperrott:
Fixes #6726 

Allows expansion indicator to be positioned at start or end of expansion panel header.

### DIFF
--- a/src/lib/expansion/accordion.ts
+++ b/src/lib/expansion/accordion.ts
@@ -6,12 +6,17 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Directive, Input} from '@angular/core';
+import {Directive, Input, SimpleChanges} from '@angular/core';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {CdkAccordion} from '@angular/cdk/accordion';
+import {Subject} from 'rxjs';
 
 /** MatAccordion's display modes. */
 export type MatAccordionDisplayMode = 'default' | 'flat';
+
+
+ /** MatAccordion's toggle positions. */
+ export type MatAccordionTogglePosition = 'before' | 'after';
 
 /**
  * Directive for a Material Design Accordion.
@@ -24,6 +29,9 @@ export type MatAccordionDisplayMode = 'default' | 'flat';
   }
 })
 export class MatAccordion extends CdkAccordion {
+  /** Stream that emits for changes in `@Input` properties. */
+  _inputChanges = new Subject<SimpleChanges>();
+
   /** Whether the expansion indicator should be hidden. */
   @Input()
   get hideToggle(): boolean { return this._hideToggle; }
@@ -39,4 +47,15 @@ export class MatAccordion extends CdkAccordion {
    *     elevation.
    */
   @Input() displayMode: MatAccordionDisplayMode = 'default';
+
+  /** The positioning of the expansion indicator. */
+  @Input() togglePosition: MatAccordionTogglePosition = 'after';
+
+  ngOnChanges(changes: SimpleChanges) {
+    this._inputChanges.next(changes);
+  }
+
+  ngOnDestroy() {
+    this._inputChanges.complete();
+  }
 }
--- a/src/lib/expansion/expansion-panel-header.html
+++ b/src/lib/expansion/expansion-panel-header.html
@@ -3,5 +3,8 @@
   <ng-content select="mat-panel-description"></ng-content>
   <ng-content></ng-content>
 </span>
-<span [@indicatorRotate]="_getExpandedState()" *ngIf="_showToggle()"
-      class="mat-expansion-indicator"></span>
+<div class="mat-expansion-indicator-container"
+     [class.mat-expansion-indicator-container-before]="_placeToggleBefore()">
+     <span *ngIf="_isToggleVisible()" class="mat-expansion-indicator"
+           [@indicatorRotate]="_getExpandedState()"></span>
+</div>
--- a/src/lib/expansion/expansion-panel-header.ts
+++ b/src/lib/expansion/expansion-panel-header.ts
@@ -16,6 +16,7 @@ import {
   ElementRef,
   Host,
   Input,
+  Optional,
   OnDestroy,
   ViewEncapsulation,
 } from '@angular/core';
@@ -23,6 +24,7 @@ import {merge, Subscription} from 'rxjs';
 import {filter} from 'rxjs/operators';
 import {matExpansionAnimations} from './expansion-animations';
 import {MatExpansionPanel} from './expansion-panel';
+import {MatAccordion} from './accordion';
 
 
 /**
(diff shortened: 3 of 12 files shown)


src/lib/expansion/accordion.ts
```diff
@@ -6,16 +6,21 @@
...
 
 /** Workaround for https://github.com/angular/angular/issues/17849 */
 export const _CdkAccordion = CdkAccordion;
 
 /** MatAccordion's display modes. */
 export type MatAccordionDisplayMode = 'default' | 'flat';
 
+
+ /** MatAccordion's toggle positions. */
+ export type MatAccordionTogglePosition = 'start' | 'end';
```
crisbeto: Everywhere else we use `before` and `after` for these kinds of inputs.

src/lib/expansion/expansion-panel-header.html
```diff
@@ -3,5 +3,7 @@
   <ng-content select="mat-panel-description"></ng-content>
   <ng-content></ng-content>
 </span>
-<span [@indicatorRotate]="_getExpandedState()" *ngIf="_showToggle()"
-      class="mat-expansion-indicator"></span>
+<div class="mat-expansion-indicator-container" [style.order]="_indicatorOrder()">
```
crisbeto: It might be more appropriate to toggle a class that sets the `order`. It also makes it easier to handle RTL.

josephperrott: Good call

josephperrott:
Comments addressed,  RTL is handled by CSS `order` of flex items in the flex container.

src/lib/expansion/accordion.ts
```diff
@@ -41,4 +50,18 @@ export class MatAccordion extends _CdkAccordion {
    *     elevation.
    */
   @Input() displayMode: MatAccordionDisplayMode = 'default';
+
+
+  /** The positioning of the expansion indicator. */
+  @Input() togglePosition: MatAccordionTogglePosition = 'after';
+
```
crisbeto: Optional: there a bit too many newlines above and below this input.

src/lib/expansion/expansion.spec.ts
```diff
@@ -129,13 +129,30 @@ describe('MatExpansionPanel', () => {
...
 
       fixture.componentInstance.expanded = true;
       fixture.detectChanges();
       tick(250);
 
-      expect(arrow.style.transform).toBe('rotate(180deg)', 'Expected 180 degree rotation.');
+      expect(arrow.style.transform).toBe('rotate(225deg)', 'Expected 225 degree rotation.');
+    }));
+
+  it('should update the indicator position when the position is set programmatically',
```
crisbeto: Also add a test that verifies that the position still works when it's set on the accordion?

josephperrott: Also added a test for displayMode since it looks like that was missing in the tests as well.

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -86,17 +88,24 @@ export class MatExpansionPanelHeader implements OnDestroy {
...
+
     // Since the toggle state depends on an @Input on the panel, we
-    // need to  subscribe and trigger change detection manually.
+    // need to subscribe and trigger change detection manually.
     this._parentChangeSubscription = merge(
       panel.opened,
       panel.closed,
-      panel._inputChanges.pipe(filter(changes => !!(changes.hideToggle || changes.disabled)))
+      merge(...changeSubjects).pipe(
+        filter(changes => !!(changes.hideToggle || changes.disabled || changes.togglePosition)))
```
jelbourn: You don't need `!!` in conditionals, only when the value is returned or stored for later

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -86,17 +88,24 @@ export class MatExpansionPanelHeader implements OnDestroy {
 
   constructor(
     renderer: Renderer2,
+    @Optional() accordion: MatAccordion,
     @Host() public panel: MatExpansionPanel,
     private _element: ElementRef,
     private _focusMonitor: FocusMonitor,
     private _changeDetectorRef: ChangeDetectorRef) {
 
+    let changeSubjects = [panel._inputChanges];
```
jelbourn: You don't care here that these are Subjects, just that they're observables. I'd call this `changeStreams`

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -132,10 +141,14 @@ export class MatExpansionPanelHeader implements OnDestroy {
   }
 
   /** Gets whether the expand indicator should be shown. */
-  _showToggle(): boolean {
+  _showIndicator(): boolean {
     return !this.panel.hideToggle && !this.panel.disabled;
   }
 
+  _placeIndicatorBefore(): boolean {
```
jelbourn: Does this function serve any value?

src/lib/expansion/expansion-panel-header.ts
```diff
@@ -132,10 +141,14 @@ export class MatExpansionPanelHeader implements OnDestroy {
   }
 
   /** Gets whether the expand indicator should be shown. */
-  _showToggle(): boolean {
+  _showIndicator(): boolean {
```
jelbourn: `_isToggleVisible`?

src/lib/expansion/expansion-panel.ts
```diff
@@ -93,6 +93,16 @@ export class MatExpansionPanel extends _MatExpansionPanelMixinBase
   /** Whether the toggle indicator should be hidden. */
   @Input() hideToggle: boolean = false;
 
+  /** The positioning of the expansion indicator. */
+  @Input()
+  get togglePosition(): MatAccordionTogglePosition {
+    return this.accordion ? this.accordion.togglePosition : this._togglePosition;
```
jelbourn: Shouldn't the more specific position trump the parent?

src/lib/expansion/expansion-panel-header.ts
josephperrott: It looks like `filter` requires the value returned to be `boolean`, and without the coercion the returned type is `SimpleChanges`

src/lib/expansion/expansion-panel-header.ts
josephperrott: This method determines the ordering of the content in the header.

src/lib/expansion/expansion-panel.ts
josephperrott: Currently the `hideToggle` value goes based on parent's value even if the child is defined so I went with the same.  There is a strong argument that they both should be changed.

josephperrott:
Caretaker:

This change fixes the spacing of the expansion panel header to be even regardless of content of the title.  This difference can be seen in the screenshot tests.

andrewseguin:
@josephperrott Can you take a look at screenshot diff issues here?

mmalerba:
@josephperrott This had to be reverted due to unexpected screenshot diffs in g3

johnchristopherjones:
Any plans to complete this feature?

PR #8200, opened 2017-11-02
build: tslint rule to enforce html tags escaping in comments

julianobrasil:
Refactoring of #5825 logic

cc @willshowell 

This rule will look for:

1 - Any HTML open/close tag delimiter (`<` or `>`) outside backtick pairs (``)

```html
/** 
 * The using of <svg> is very good because...
 */
```

```
/** 
 * Look for every occurrency of `i <= 18...
 */
```

2 - Any "opening" backtick without its "closing" match:

```html
/** 
 * The using of `<svg> is very good because...
 */
```

```
/** 
 * We recommend using backticks (`) to escape...
 */
```

And it will allow any html provided it's between backtick matching pairs:

```html
/** 
 * The using of `    <div><br></div> is very` good because...
 */
```

### DIFF
--- a/tools/tslint-rules/noUnscapedHtmlTagRule.js
+++ b/tools/tslint-rules/noUnscapedHtmlTagRule.js
@@ -1,0 +1,67 @@
+const ts = require('typescript');
+const utils = require('tsutils');
+const Lint = require('tslint');
+
+const ERROR_MESSAGE =
+  'An HTML tag may only appear in a JSDoc comment if it is escaped.' +
+  ' This prevents failures in document generation caused by a misinterpreted tag.';
+
+/**
+ * Rule that walks through all comments inside of the library and adds failures when it
+ * detects unescaped HTML tags inside of multi-line comments.
+ */
+class Rule extends Lint.Rules.AbstractRule {
+
+  apply(sourceFile) {
+    return this.applyWithWalker(new NoUnescapedHtmlTagWalker(sourceFile, this.getOptions()));
+  }
+}
+
+class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
+
+  visitSourceFile(sourceFile) {
+    utils.forEachComment(sourceFile, (fullText, commentRange) => {
+
+      let isEscapedHtmlTag = true;
+      const matches = new RegExp(/[<>]/);
+
+      const numberOfBackticks = fullText.split('`').length - 1;
+      if ((numberOfBackticks === 1) || ((numberOfBackticks === 0) && matches.test(fullText))) {
+        isEscapedHtmlTag = false;
+      }
+
+      // if there are no backticks and [<>], there's no need for any more checks
+      if ((numberOfBackticks > 1) && matches.test(fullText)) {
+        // if there are backticks there should be an even number of them
+        if (!!(numberOfBackticks % 2)) {
+          isEscapedHtmlTag = false;
+        } else {
+          /** 
+           * This logic behaves like a stack structure. isBacktickWithoutMatch plays
+           * the stack role: it is set to true whenever, during a comment scan, an 'open' 
+           * backtick is found in the string, and it is set back to false when the next matching
+           * (closing) backtick is found. Every backtick must have a matching. < and >
+           * must always be between two matching backticks.
+           */
+          const splitedFullText = fullText.split('');
+          let isBacktickWithoutMatch = false;
+
+          for (var i = 0; i < splitedFullText.length; i++) {
+            if (splitedFullText[i] === '`') {
+              isBacktickWithoutMatch = !isBacktickWithoutMatch;
+            } else if (matches.test(splitedFullText[i]) && !isBacktickWithoutMatch) {
+              isEscapedHtmlTag = false;
+              break;
+            }
+          }
+        }
+      }
+
+      if (commentRange.kind === ts.SyntaxKind.MultiLineCommentTrivia && !isEscapedHtmlTag) {
+        this.addFailureAt(commentRange.pos, commentRange.end - commentRange.pos, ERROR_MESSAGE);
+      }
+    });
+  }
+}
+
+exports.Rule = Rule;
--- a/tools/tslint-rules/noUnscapedHtmlTagRule.js
+++ b/tools/tslint-rules/noUnscapedHtmlTagRule.js
@@ -22,45 +22,46 @@ class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
   visitSourceFile(sourceFile) {
     utils.forEachComment(sourceFile, (fullText, commentRange) => {
 
-      let isEscapedHtmlTag = true;
-      const matches = new RegExp(/[<>]/);
+      const htmlIsEscaped = parseForHtml(fullText);
 
-      const numberOfBackticks = fullText.split('`').length - 1;
-      if ((numberOfBackticks === 1) || ((numberOfBackticks === 0) && matches.test(fullText))) {
-        isEscapedHtmlTag = false;
-      }
-
-      // if there are no backticks and [<>], there's no need for any more checks
-      if ((numberOfBackticks > 1) && matches.test(fullText)) {
-        // if there are backticks there should be an even number of them
-        if (!!(numberOfBackticks % 2)) {
-          isEscapedHtmlTag = false;
-        } else {
-          /** 
-           * This logic behaves like a stack structure. isBacktickWithoutMatch plays
-           * the stack role: it is set to true whenever, during a comment scan, an 'open' 
-           * backtick is found in the string, and it is set back to false when the next matching
-           * (closing) backtick is found. Every backtick must have a matching. < and >
-           * must always be between two matching backticks.
-           */
-          const splitedFullText = fullText.split('');
-          let isBacktickWithoutMatch = false;
-
-          for (var i = 0; i < splitedFullText.length; i++) {
-            if (splitedFullText[i] === '`') {
-              isBacktickWithoutMatch = !isBacktickWithoutMatch;
-            } else if (matches.test(splitedFullText[i]) && !isBacktickWithoutMatch) {
-              isEscapedHtmlTag = false;
-              break;
-            }
-          }
-        }
-      }
-
-      if (commentRange.kind === ts.SyntaxKind.MultiLineCommentTrivia && !isEscapedHtmlTag) {
+      if (commentRange.kind === ts.SyntaxKind.MultiLineCommentTrivia && !htmlIsEscaped) {
         this.addFailureAt(commentRange.pos, commentRange.end - commentRange.pos, ERROR_MESSAGE);
       }
     });
+  }
+
+  /** Gets whether the comment's HTML, if any, is properly escaped */
+  parseForHtml(fullText) {
+    const matches = new RegExp(/[<>]/);
+
+    const backtickCount = fullText.split('`').length - 1;
+    if ((backtickCount === 1) || ((backtickCount === 0) && matches.test(fullText))) {
+      return false;
+    }
+
+    // if there are no backticks and no [<>], there's no need for any more checks
+    if ((backtickCount > 1) && matches.test(fullText)) {
+      // if there are backticks there should be an even number of them
+      if (backtickCount % 2) {
+        return false;
+      } else {
+        // < and > must always be between two matching backticks.
+        const fullTextArray = fullText.split('');
+
+        // Whether an opening backtick has been found without a closing pair
+        let openBacktick = false;
+
+        for (let i = 0; i < fullTextArray.length; i++) {
+          if (fullTextArray[i] === '`') {
+            openBacktick = !openBacktick;
+          } else if (matches.test(fullTextArray[i]) && !openBacktick) {
+            return false;
+          }
+        }
+      }
+    }
+
+    return true;
   }
 }
 
--- a/tools/tslint-rules/noUnscapedHtmlTagRule.js
+++ b/tools/tslint-rules/noUnscapedHtmlTagRule.js
@@ -33,31 +33,29 @@ class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
   /** Gets whether the comment's HTML, if any, is properly escaped */
   parseForHtml(fullText) {
     const matches = new RegExp(/[<>]/);
-
     const backtickCount = fullText.split('`').length - 1;
-    if ((backtickCount === 1) || ((backtickCount === 0) && matches.test(fullText))) {
+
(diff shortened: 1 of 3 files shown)


tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
+
+  visitSourceFile(sourceFile) {
+    utils.forEachComment(sourceFile, (fullText, commentRange) => {
+
+      let isEscapedHtmlTag = true;
+
+      const numberOfBackticks = fullText.split('`').length - 1;
+      // if there are no backticks no check is necessary
+      if (numberOfBackticks > 0) {
```
willshowell: But there might be `<html>` without backticks?

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
+
+  visitSourceFile(sourceFile) {
+    utils.forEachComment(sourceFile, (fullText, commentRange) => {
+
+      let isEscapedHtmlTag = true;
+
+      const numberOfBackticks = fullText.split('`').length - 1;
```
willshowell: `backtickCount`?

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+      // if there are no backticks no check is necessary
+      if (numberOfBackticks > 0) {
+        // if there are backticks there should be an even number of them
+        if (!!(numberOfBackticks % 2)) {
```
willshowell: Should be fine to just use `if(numberOfBackticks % 2)` since you only care if it's zero or non-zero

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+    return this.applyWithWalker(new NoUnescapedHtmlTagWalker(sourceFile, this.getOptions()));
+  }
+}
+
+class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
+
+  visitSourceFile(sourceFile) {
+    utils.forEachComment(sourceFile, (fullText, commentRange) => {
+
+      let isEscapedHtmlTag = true;
```
willshowell: `htmlIsEscaped`?

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+        if (!!(numberOfBackticks % 2)) {
+          isEscapedHtmlTag = false;
+        } else {
+          /** 
+           * This logic behaves like a stack structure. isBacktickWithoutMatch plays
+           * the stack role: it is set to true whenever, during a comment scan, an 'open' 
+           * backtick is found in the string, and it is set back to false when the next matching
+           * (closing) backtick is found. Every backtick must have a matching. < and > are only
+           * allowed between two matching backticks.
+           */
```
willshowell: Use `// single-line comments`

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+          const matches = new RegExp(/[<>]/);
+          const splitedFullText = fullText.split('');
```
willshowell: `fullTextArray`?

tslint.json
```diff
@@ -72,16 +79,26 @@
...
+      ],
+      [
+        "fdescribe"
+      ],
+      [
+        "xit"
+      ],
+      [
+        "xdescribe"
+      ]
```
willshowell: I think the original formatting was fine. Just add `no-unescaped-html-tag` without any of the other changes.

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+          let isBacktickWithoutMatch = false;
+
+          for (var i = 0; i < splitedFullText.length; i++) {
```
willshowell: `let i = 0`

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,63 @@
...
+      }
+
+      if (commentRange.kind === ts.SyntaxKind.MultiLineCommentTrivia && !isEscapedHtmlTag) {
+        this.addFailureAt(commentRange.pos, commentRange.end - commentRange.pos, ERROR_MESSAGE);
+      }
+    });
+  }
+}
+
+exports.Rule = Rule;
```
willshowell: Another way to write this rule that might be cleaner would be,

```ts
utils.forEachComment(sourceFile, (fullText, commentRange) => {
  const htmlIsEscaped = parseForHtml(fullText);
  if (commentRange.kind === ... && !htmlIsEscaped) { 
    this.addFailureAt(...);
  }
}

/** Gets whether the comment's HTML, if any, is properly escaped */
parseForHtml(fullText) {
 // parsing done here
}
```

I think that way you wouldn't have to worry so much about carrying the flag through the whole method and could instead return early, when needed.

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,67 @@
...
+
+class NoUnescapedHtmlTagWalker extends Lint.RuleWalker {
+
+  visitSourceFile(sourceFile) {
+    utils.forEachComment(sourceFile, (fullText, commentRange) => {
+
+      let isEscapedHtmlTag = true;
+      const matches = new RegExp(/[<>]/);
+
+      const numberOfBackticks = fullText.split('`').length - 1;
```
willshowell: `backtickCount`?

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,67 @@
...
+        if (!!(numberOfBackticks % 2)) {
+          isEscapedHtmlTag = false;
+        } else {
+          /** 
+           * This logic behaves like a stack structure. isBacktickWithoutMatch plays
+           * the stack role: it is set to true whenever, during a comment scan, an 'open' 
+           * backtick is found in the string, and it is set back to false when the next matching
+           * (closing) backtick is found. Every backtick must have a matching. < and >
+           * must always be between two matching backticks.
+           */
```
willshowell: Use `// single-line comments`

Also I think only the last sentence is needed.

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,67 @@
...
+          const splitedFullText = fullText.split('');
+          let isBacktickWithoutMatch = false;
```
willshowell: Maybe

```ts
// Whether an opening backtick has been found without a closing pair
let openBacktick = false;
```

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,67 @@
...
+          isEscapedHtmlTag = false;
+        } else {
+          /** 
+           * This logic behaves like a stack structure. isBacktickWithoutMatch plays
+           * the stack role: it is set to true whenever, during a comment scan, an 'open' 
+           * backtick is found in the string, and it is set back to false when the next matching
+           * (closing) backtick is found. Every backtick must have a matching. < and >
+           * must always be between two matching backticks.
+           */
+          const splitedFullText = fullText.split('');
```
willshowell: `fullTextArray`?

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,67 @@
...
+          let isBacktickWithoutMatch = false;
+
+          for (var i = 0; i < splitedFullText.length; i++) {
```
willshowell: `let i = 0`

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,67 @@
...
+
+      const numberOfBackticks = fullText.split('`').length - 1;
+      if ((numberOfBackticks === 1) || ((numberOfBackticks === 0) && matches.test(fullText))) {
+        isEscapedHtmlTag = false;
+      }
+
+      // if there are no backticks and [<>], there's no need for any more checks
+      if ((numberOfBackticks > 1) && matches.test(fullText)) {
+        // if there are backticks there should be an even number of them
+        if (!!(numberOfBackticks % 2)) {
```
willshowell: Can just use `numberOfBackticks % 2` since you only care whether its zero or non-zero.

tslint.json
julianobrasil: Oh, that was and accidental `shift+alt+f` in Visual Studio Code.

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    // if there are no backticks and no [<>], there's no need for any more checks
+    if ((backtickCount > 1) && matches.test(fullText)) {
+      // if there are backticks there should be an even number of them
+      if (backtickCount % 2) {
+        return false;
+      } else {
+        // < and > must always be between two matching backticks.
+        const fullTextArray = fullText.split('');
```
willshowell: I think the logic could be tightened up like this,

```ts
const matches = new RegExp(/[<>]/);
const backtickCount = fullText.split('`').length - 1;

// An odd number of backticks or html without backticks is invalid
if (backtickCount % 2 || (!backtickCount && matches.test(fullText))) {
  return false;
}

// Text without html is valid
if (!matches.test(fullText)) {
  return true;
}

// < and > must always be between two matching backticks
const fullTextArray = fullText.split('');
```

julianobrasil: Better!

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,66 @@
...
+
+      if (commentRange.kind === ts.SyntaxKind.MultiLineCommentTrivia && !htmlIsEscaped) {
+        this.addFailureAt(commentRange.pos, commentRange.end - commentRange.pos, ERROR_MESSAGE);
+      }
+    });
+  }
+
+  /** Gets whether the comment's HTML, if any, is properly escaped */
+  parseForHtml(fullText) {
+    const matches = new RegExp(/[<>]/);
```
rafaelss95: Can be simplified to `const matches = /[<>]/;`

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,66 @@
...
+    const backtickCount = fullText.split('`').length - 1;
+
+    // An odd number of backticks or html without backticks is invalid
+    if ((backtickCount % 2) || ((backtickCount === 0) && matches.test(fullText))) {
```
rafaelss95: I think there are too many parenthesis. It could be: `if (backtickCount % 2 || (backtickCount === 0 && matches.test(fullText)) {`

tools/tslint-rules/noUnscapedHtmlTagRule.js
```diff
@@ -0,0 +1,66 @@
...
+      return true;
+    }
+
+    // < and > must always be between two matching backticks.
+    const fullTextArray = fullText.split('');
+
+    // Whether an opening backtick has been found without a closing pair
+    let openBacktick = false;
+
+    for (let i = 0; i < fullTextArray.length; i++) {
```
rafaelss95: You can use `for..of` and loop over the string, so you don't need to use `split('')` above.

tools/tslint-rules/noUnscapedHtmlTagRule.js
julianobrasil: I missed it.

willshowell:
@julianobrasil one last thing I noticed: since you opened the original PR, the other linters have been rewritten in typescript. Might want to convert it and make sure it matches the updated format of the other ones.

julianobrasil:
Travis output bellow. Apparently this lint rule is doing what it is supposed to do.

image

willshowell:
Awesome! You can also run `npm run tslint` locally to lint. And I think those fixes should be included with this PR.

julianobrasil:
@willshowell, locally I got a lot of lint error messages because windows line break is different from LF.

jelbourn:
@julianobrasil could you rebase this PR?

julianobrasil:
Done.

tslint.json
```diff
@@ -118,7 +118,8 @@
     "missing-rollup-globals": [
       true,
       "./tools/package-tools/rollup-globals.ts",
-      "src/+(lib|cdk|material-examples|material-experimental|cdk-experimental)/**/*.ts"
-    ]
+      "src/+(lib|cdk|material-examples)/**/*.ts"
```
jelbourn: Did you mean to remove `material-experimental|cdk-experimental` here?

julianobrasil: Absolutely not. This was an oversight.

jelbourn:
@julianobrasil ah, this actually catches a pretty large number of instances already in the codebase. Did you want to send a separate PR that just removes those first so we can merge this and still have everything be green?

julianobrasil:
I'll do it.

jelbourn:
Looks like there are still two things being caught:
```
ERROR: /home/travis/build/angular/material2/src/lib/chips/chip-input.ts[24, 1]: An HTML tag delimiter (< or >) may only appear in a JSDoc comment if it is escaped. This prevents failures in document generation caused by a misinterpreted tag.
ERROR: /home/travis/build/angular/material2/src/lib/icon/icon.ts[33, 1]: An HTML tag delimiter (< or >) may only appear in a JSDoc comment if it is escaped. This prevents failures in document generation caused by a misinterpreted tag.
```

julianobrasil:
This gave me more trouble than I expected. @jelbourn, sorry for all of this commiting ping-pong, but after updating  my local copy of material repo on wednesday, I got a lot of errors with `npm install`. And this is preventing me from running anything locally (not sure yet what the problem is, but I'm trying to talk to the guys of grpc/grpc-node repo about it as it seems to be the culprit)

jelbourn:
This one passes now, so I'm merging it

PR #8203, opened 2017-11-03
build: remove es5 tsconfig files

DevVersion:
* Makes use of the new overrides feature of the Angular Compiler CLI v5.
* Builds ES5 and ES2015 in parallel.


tools/package-tools/compile-entry-point.ts
```diff
@@ -10,13 +10,19 @@ let nextId = 0;
...
 /** Compiles the TypeScript sources of a primary or secondary entry point. */
 export async function compileEntryPoint(buildPackage: BuildPackage, tsconfigName: string,
-                                          secondaryEntryPoint = '') {
+                                          secondaryEntryPoint = '', es5OutputPath?: string) {
   const entryPointPath = join(buildPackage.sourceDir, secondaryEntryPoint);
   const entryPointTsconfigPath = join(entryPointPath, tsconfigName);
+  const ngcFlags = ['-p', entryPointTsconfigPath];
+
+  if (es5OutputPath) {
+    ngcFlags.push('--outDir', es5OutputPath);
```
crisbeto: You can combine these into a single push: `ngcFlags.push('--outDir', es5OutputPath, '--target', 'ES5')`.

DevVersion: True, I just wanted to keep it separate since those are two different option-value pairs and this way it feels more readable.

PR #8223, opened 2017-11-03
feat(form-field): add support for separate label and placeholder

### LINKED ISSUE #6194: feat(input, select): separate the placeholder and label as shown in the official mocks
Bug

The placeholder used in a textarea with multiple rows should automatically use up the other rows when too long to fit in just the first row.

The placeholder is truncated in the first row with the other rows left blank.

Plunk: http://plnkr.co/edit/OAxiFX?p=preview

I would want to tell the user as to what format a textarea is to be filled up. Possibly referencing certain standard examples / policies for which I need a few lines' worth of text.

Angular 4.3.1
Material 2.0.0-beta.8-374aaff
All OS
Typescript ^2.2.2
Checked in Chrome & Firefox

No

crisbeto:
* Adds the ability for the user to specify both a label and placeholder text for a form field.
* Renames all of the "floating placeholder" related terminology to refer to a "floating label" instead.
* Aligns the input placeholder color with the Material spec.

Fixes #6194.


crisbeto:
@mmalerba I'm open to using the native `label` instead of `mat-label`, but it would require use to have a directive that selects all `label` nodes.

jelbourn:
Could this be done without any breaking changes?

crisbeto:
The only real breaking change is renaming the `shouldPlaceholderFloat` to `shouldLabelFloat` in the `MatFormFieldControl` interface @jelbourn, everything else is still re-exported under its old name. I can re-add the `shouldPlaceholderFloat` as an optional value and mark it as deprecated, but I suspect that not many people are implementing the interface.

jelbourn:
Let's go with that- I want to anything we have to call a breaking API change

crisbeto:
Refactored to avoid breaking changes.

mmalerba:
Still need to take a closer look at the code, but had a couple thoughts after looking at the demo:
1. It should be possible to have in input with no label and a placeholder, but making that work now would be a breaking change. Maybe we can just file a new issue and update that when we're making breaking changes
2. The placeholder needs an animation so it doesn't overlap with the label when the label is animating up. The spec has a [video](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B8wSqcLwbhFuREg4Q0R4M3VGejA/textfields-presslight-01.webm) showing what it should look like

Thanks for doing this! Something I've been wanting to do for a while but haven't gotten the time

crisbeto:
1. Yeah I wanted to avoid breaking changes.
2. We can't really animate the native placeholder attribute, the best we could do is delay showing it until the animation is done, but that means switching over the placeholder animations to Angular animations. We could do it in a follow-up.

mmalerba:
we might not be able to do the sequenced animation, but could we at least slap a transition on the opacity? I think it will make it look a little nicer

crisbeto:
We could do that, but it would be a progressive enhancement since not all browsers allow you to transition the native placeholder (e.g. Chrome does, but Firefox doesn't).

crisbeto:
Rebased and made it so the placeholder doesn't overlap the animating label @mmalerba, can you take another look?

src/lib/form-field/form-field.html
```diff
@@ -8,24 +8,31 @@
     <div class="mat-input-infix mat-form-field-infix">
       <ng-content></ng-content>
 
-      <span class="mat-input-placeholder-wrapper mat-form-field-placeholder-wrapper">
+      <span class="mat-form-field-label-wrapper">
```
mmalerba: I'm sure people are targeting those classes with custom styles, lets leave them for now with a TODO to remove them at the next major.

src/lib/form-field/form-field.html
```diff
@@ -8,24 +8,31 @@
...
         <!-- We add aria-owns as a workaround for an issue in JAWS & NVDA where the label isn't
              read if it comes before the control in the DOM. -->
-        <label class="mat-input-placeholder mat-form-field-placeholder"
+        <label class="mat-form-field-label"
```
mmalerba: Same, leave the old classes too for now

src/lib/form-field/form-field.ts
```diff
@@ -220,24 +227,40 @@ export class MatFormField implements AfterViewInit, AfterContentInit, AfterConte
...
   _hasPlaceholder() {
     return !!(this._control.placeholder || this._placeholderChild);
   }
 
+  _hasLabel() {
+    return !!this._labelChild;
+  }
+
+  _shouldLabelFloat() {
+    return this._canLabelFloat && (this._control.shouldLabelFloat ||
```
mmalerba: This logic seems different than before, it didn't consider `_canPlaceholderFloat` before

src/lib/form-field/form-field.ts
```diff
@@ -220,24 +227,40 @@ export class MatFormField implements AfterViewInit, AfterContentInit, AfterConte
...
+        this._control.shouldPlaceholderFloat || this._shouldAlwaysFloat);
+  }
+
+  _hideControlPlaceholder() {
+    return !this._hasLabel() || !this._shouldLabelFloat();
```
mmalerba: Maybe the `_canLabelFloat` from above should be factored into this logic instead?

src/lib/form-field/form-field.ts
crisbeto: If I remember correctly, some of the logic was done by combining CSS classes. I simplified it to make it easier to follow.

src/lib/form-field/form-field.ts
crisbeto: The `_shouldLabelFloat` call here already factors in `_canLabelFloat`.

crisbeto:
I restored the classes and addressed the comments @mmalerba, can you take another look?

jelbourn:
@crisbeto ready to presubmit if you can rebase

crisbeto:
Rebased @jelbourn.

tinayuangao:

crisbeto:
Rebased @tinayuangao.

PR #8226, opened 2017-11-03
fix(overlay): disposed overlays not removed from the key event stack

crisbeto:
Fixes overlays that are removed via `dispose` without being `detached` first not being removed from the stack inside the `OverlayKeyboardDispatcher`, causing them to block events for any subsequent overlays.


jelbourn:
@crisbeto looks like there are some CI failures after rebasing- probably some conflict with an earlier PR. Will catch this one in a subsequent sync

crisbeto:
See https://github.com/angular/material2/pull/8559 @jelbourn.

PR #8235, opened 2017-11-05
refactor: remove renderer usages

crisbeto:
Removes any usages of the Renderer2 from the codebase. They're no longer necessary after Angular switched to Domino which is closer to the native DOM API.


jelbourn:
Needs rebase, I'll try to get this first in the next round of PRs

crisbeto:
Ready to go @jelbourn.

PR #8243, opened 2017-11-06
feat(expansion-panel): allow for content to be rendered lazily

### LINKED ISSUE #8230: [mdExpansionPanel] - Feature request - Lazy load content

feature request

Don't create the body until the panel is expanded

The body is created at the panel creation, meaning that if your panel has a large content and it takes time to create a panel body that will probably not even be viewed.

I have expansion panels that can have a lot of content, they sometimes take more than 500ms to render, that's a lot of time. If I put a *ngif in the body binded to the expanded status, it's way better but the expansion collapse is a bit broken because the content immediatly disappears when you collapse.

crisbeto:
Currently the expansion panel renders all of its content on load which can be expensive, considering that all of it won't be visible. These changes allow for some of the content to be rendered lazily via the `matExpansionPanelContent` directive.

Fixes #8230.

### DIFF
--- a/src/lib/expansion/expansion-panel-content.ts
+++ b/src/lib/expansion/expansion-panel-content.ts
@@ -1,0 +1,20 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Directive, TemplateRef} from '@angular/core';
+
+/**
+ * Expansion panel content that will be rendered lazily
+ * after the panel is opened for the first time.
+ */
+@Directive({
+  selector: 'ng-template[matExpansionPanelContent]'
+})
+export class MatExpansionPanelContent {
+  constructor(public _template: TemplateRef<any>) {}
+}
--- a/src/lib/expansion/expansion-panel.html
+++ b/src/lib/expansion/expansion-panel.html
@@ -1,8 +1,11 @@
 <ng-content select="mat-expansion-panel-header"></ng-content>
-<div [class.mat-expanded]="expanded" class="mat-expansion-panel-content"
-     [@bodyExpansion]="_getExpandedState()" [id]="id">
+<div class="mat-expansion-panel-content"
+     [class.mat-expanded]="expanded"
+     [@bodyExpansion]="_getExpandedState()"
+     [id]="id">
   <div class="mat-expansion-panel-body">
     <ng-content></ng-content>
+    <ng-template [cdkPortalOutlet]="_portal"></ng-template>
   </div>
   <ng-content select="mat-action-row"></ng-content>
 </div>
--- a/src/lib/expansion/expansion-panel.ts
+++ b/src/lib/expansion/expansion-panel.ts
@@ -20,13 +20,21 @@ import {
   Optional,
   SimpleChanges,
   ViewEncapsulation,
+  ViewContainerRef,
+  AfterContentInit,
+  ContentChild,
 } from '@angular/core';
 import {CdkAccordionItem} from '@angular/cdk/accordion';
 import {UniqueSelectionDispatcher} from '@angular/cdk/collections';
 import {CanDisable, mixinDisabled} from '@angular/material/core';
+import {TemplatePortal} from '@angular/cdk/portal';
 import {Subject} from 'rxjs/Subject';
+import {take} from 'rxjs/operators/take';
+import {filter} from 'rxjs/operators/filter';
+import {startWith} from 'rxjs/operators/startWith';
 import {MatAccordion} from './accordion';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {MatExpansionPanelContent} from './expansion-panel-content';
 
 /** Workaround for https://github.com/angular/angular/issues/17849 */
 export const _CdkAccordionItem = CdkAccordionItem;
@@ -91,7 +99,7 @@ export type MatExpansionPanelState = 'expanded' | 'collapsed';
   ],
 })
 export class MatExpansionPanel extends _MatExpansionPanelMixinBase
-    implements CanDisable, OnChanges, OnDestroy {
+    implements CanDisable, AfterContentInit, OnChanges, OnDestroy {
 
   /** Whether the toggle indicator should be hidden. */
   @Input()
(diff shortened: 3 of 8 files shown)


src/lib/expansion/expansion-panel.ts
```diff
@@ -127,12 +147,31 @@ export class MatExpansionPanel extends _MatExpansionPanelMixinBase
...
+    this._portal = new TemplatePortal<any>(this._content, this._viewContainerRef);
+
+    // Render the content as soon as the panel becomes open.
+    this.opened.pipe(startWith(null!)).subscribe(() => {
+      if (this.expanded && !this._outlet.hasAttached()) {
+        this._outlet.attach(this._portal);
+      }
+    });
+
+    // Remove the content once the panel is closed and the closing animation is done.
```
josephperrott: What is the reasoning behind removing the content on closing?  Since the issue we are looking to address is lazy loading, is there a reason to detach and reattach all the time?

src/lib/expansion/expansion-panel.ts
```diff
@@ -127,12 +147,31 @@ export class MatExpansionPanel extends _MatExpansionPanelMixinBase
     return this.expanded ? 'expanded' : 'collapsed';
   }
 
+  ngAfterViewInit() {
```
josephperrott: We have had a number of issues with content changed after checked for these situations, are we going to run into the same issue here?

crisbeto: Probably not, those exceptions are usually if data bindings change. This is the same way we're doing it for tabs.

src/lib/expansion/expansion-panel.ts
crisbeto: I was a bit on the fence on this one as well, but it could help if you have some really massive panels with lots of elements. I'm open to not doing this though.

bardiel: Could this be configured perhaps via an Input?
I'm developing (for learning purposes) an RSS feed reader. On desktop, each item is placed inside an expansion panel (for fancyness (?)); inspired by Google Inbox interface.

Removing the content inside the panel after been read, could be something I'd like to happen. Being that as more and more items are been loaded, more resources will be consumed. (O course I shall do something about this regardless of this specific feature).

However, also thinking about other Google products like GSuit admin console: where some settings are loaded asynchronously the first time, but then kept. (I don't know if they're actually removed from DOM, but I wouldn't remove them in such scenario).

josephperrott: I think it might be best at this point to not remove the content once it is placed and we can determine if it is something that we want to allow later, potentially as @bardiel mentioned via `@Input`

src/lib/expansion/expansion-panel.ts
josephperrott: For tabs we actually are having to change it because of these issues #7266 

We are creating a custom template portal and having the template portal itself be responsible for attachment because of issues with `ChangeAfterContentChecked`.  I just want to be sure we don't accidentally create an issue further down the road.

src/lib/expansion/expansion-panel.html
```diff
@@ -1,8 +1,12 @@
...
-<div [class.mat-expanded]="expanded" class="mat-expansion-panel-content"
-     [@bodyExpansion]="_getExpandedState()" [id]="id">
+<div class="mat-expansion-panel-content"
+     [class.mat-expanded]="expanded"
+     [@bodyExpansion]="_getExpandedState()"
+     (@bodyExpansion.done)="_animationDone.next()"
+     [id]="id">
   <div class="mat-expansion-panel-body">
-    <ng-content></ng-content>
+    <ng-template><ng-content></ng-content></ng-template>
```
jelbourn: This doesn't stop Angular from initializing stuff; we can make this change to stop stuff from making it into the DOM, but we'll ultimately need an alternate API where users provide an `ng-template`

crisbeto: Ah, I was under the impression that this was the reason we do it the way we do for the tabs. In that case I'll rework it to use a `@ContentChild` to find an `ng-template` and instantiate it.

amcdnl: @crisbeto - Look at what I did for https://github.com/angular/material2/pull/6832

crisbeto:
Rebased and reworked the approach to take an `ng-template` with `matExpansionPanelContent` instead. Can you take another look @jelbourn @josephperrott?

src/lib/expansion/expansion-panel.ts
```diff
@@ -137,6 +152,19 @@ export class MatExpansionPanel extends _MatExpansionPanelMixinBase
     return this.expanded ? 'expanded' : 'collapsed';
   }
 
+  ngAfterContentInit() {
+    if (this._lazyContent) {
+      // Render the content as soon as the panel becomes open.
+      this.opened.pipe(
+        startWith(null!),
```
jelbourn: Why does this start with never?

src/lib/expansion/expansion.md
```diff
@@ -99,9 +99,26 @@ panel can be expanded at a given time:
 </mat-accordion>
 ```
 
+### Lazy rendering
+
+If you want some content not to be rendered until its expansion panel is opened, you can wrap it in
+an `ng-template` using the `matExpansionPanelContent` directive.
```
jelbourn: How about
`````markdown
By default, the expansion panel content will be initialized even when the panel is closed.
To instead defer initialization until the panel is open, the content should be provided as
an `ng-template`:
```html
<mat-expansion-panel>
  <mat-expansion-panel-header>
    This is the expansion title
  </mat-expansion-panel-header>

  <ng-template matExpansionPanelContent>
    Some deferred content
  </ng-template>
</mat-expansion-panel>
```
`````

File: src/lib/expansion/expansion-panel.ts (hunk 4 above)
Comment (crisbeto) [reply src/lib/expansion/expansion-panel.ts]: Because the `opened` stream won't emit if the panel is open by default. This ensures that we render any panels that are open on load.

**Comment by crisbeto** [comment]

File: src/lib/expansion/expansion-panel-content.ts [hunk 6]
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Expansion panel content that will be rendered lazily
+ * after the panel is opened for the first time.
+ */
+@Directive({
+  selector: 'ng-template[matExpansionPanelContent]'
+})
+export class MatExpansionPanelContent {
+  constructor(public _template: TemplateRef<any>) {}
```
rafaelss95: Just a quick question: Don't you use `_` for `private` variables only?

willshowell: They also use it for "internal use only" properties

Issue #6534, opened 2017-08-17
Support flexible overlay size with connected position strategy


jelbourn:
This issue tracks the high-level feature of supporting overlays that are automatically an appropriate size based on their position and viewport bounds. This is accomplished by creating a bounding container element that constrains the size of the overlay panel. 

In-progress work is: https://github.com/jelbourn/material2/blob/better-connected-position/src/lib/core/overlay/position/better-connected-position-strategy.ts

willshowell:
As a reminder from https://github.com/angular/material2/pull/8280#discussion_r150113801, it would be great if whatever api is exposed for "position changes" emits on actual changes (instead of all position events) and within the ng zone.

isaacplmann:
Is there work behind the scenes being done on this? Or is this something that would benefit from someone writing a PR?

@jelbourn, it looks like you've already put some thought into how to solve this.  What is still left to be done?

crisbeto:
The work is close to being finished @isaacplmann.

isaacplmann:
@crisbeto Great!  Glad to hear it.  Thanks for the hard work.

PR #8280, opened 2017-11-07
fix(demo-app): fix aligned menu triggers and make page scrollable

willshowell:
This change keeps the classlist up to date when the menu panel is repositioned due to scrolling.

Since the menu does not need a correct `transform-origin` during its exit animation, this is not really noticeable or necessary. However, some users may be looking for positioning classes, or the animation may change in the future.

It also adds a scrolling to the demo app and fixes a misused style.

### DIFF
--- a/src/demo-app/menu/menu-demo.html
+++ b/src/demo-app/menu/menu-demo.html
@@ -167,3 +167,5 @@
     </mat-menu>
   </div>
 </div>
+
+<div style="height: 500px">This div is for testing scrolled menus.</div>
--- a/src/demo-app/menu/menu-demo.scss
+++ b/src/demo-app/menu/menu-demo.scss
@@ -8,6 +8,6 @@
   }
 
   .end-icon {
-    align-items: flex-end;
+    justify-content: flex-end;
   }
 }


src/lib/menu/menu-trigger.ts
```diff
@@ -341,7 +343,8 @@ export class MatMenuTrigger implements AfterContentInit, OnDestroy {
       const posX: MenuPositionX = change.connectionPair.overlayX === 'start' ? 'after' : 'before';
       const posY: MenuPositionY = change.connectionPair.overlayY === 'top' ? 'below' : 'above';
 
-      this.menu.setPositionClasses(posX, posY);
+      // Note that the position changes may be outside the Angular zone
+      this._ngZone.run(() => this.menu.setPositionClasses(posX, posY));
```
amcdnl: Does `markForCheck` in the `ChangeDetectorRef` not work for this situation?

willshowell: I thought `detectChanges` might work, but no.

The problem is that `onPositionChange` sometimes emits within and sometimes outside NgZone due to the scroll dispatcher. I don't have a complete grasp on it, and this might be better if `onPositionChange` always just emitted from within the zone instead of trying to bring it back in at the component level.

@crisbeto can you weigh in?

crisbeto: Ideally we should be consistent in terms of where the event is being emitted, however with the way things are set up at the moment (`onPositionChange` isn't really a "change" event but more of an "update" event since it emits even if the position didn't change) we'd end up hitting the zone for all scroll events. We can address it better once the new connected positioning is in.

willshowell:
@crisbeto that sounds good - I didn't even notice it was emitting on every event. I'm happy to repurpose this to be for just updating the demo app.

Do you or @jelbourn have a design doc/canonical issue for tracking improvements to the strategy? I'd hate for some of these things to be forgotten once it is available.

jelbourn:
The new connected position strategy is tracked in #6534

src/lib/menu/menu-trigger.ts
jelbourn: It would be better at least to only enter the zone when `posX` or `posY` change

willshowell:
This now _only_ updates the menu demo flex styles and adds extra height to demo scrolling behavior.

PR #8288, opened 2017-11-08
chore: fix lint errors

crisbeto:
Fixes a couple of stylelint and tslint errors.


crisbeto:
Ueah I think that this is the third PR where I’ve had it.

PR #8292, opened 2017-11-08
wait for a while then call `_checkThemeIsPresent`

### LINKED ISSUE #8291: [feature request] Add an option to disable checkTheme or check after waiting for a while.
feature request

Do not show `checkTheme` warning in some case.

For example we are using webpack to load global `.scss` file and `extract-webpack-plugin` is disabled on development, so material theme will be loaded asynchronously, then there will be an warning about no material theme.

JounQin:
close #8291


JounQin:
I signed it!

PR #8303, opened 2017-11-08
fix(table): inaccurate row height

### LINKED ISSUE #8299: bug(table): divider should sit within the baseline
In the cource of reviewing #5862, I had a look at the specs of the divider and noticed that according to MD guidelines a divider should sit within the baseline of an element:

So, refered to the data table component, the height of a row + divider should be still 48px, but currently it's 49px:

This would match with the image in the specs, as the size marker, which is specified with the mentioned 48px, starts at the first pixel of the row but ends at the position of the divider:

I know this is a **very minor** issue, I mean, it's about 1px (...although it gets more with each row...), but I thought since we want to follow the MD Guide, it's worth mentioning it.

So, there are two solutions at the moment, setting `box-sizing` to `border-box` or giving the row a `min-height` of `47px`. The latter would affect my open pull request(#8215).

crisbeto:
Fixes the table border throwing of the table row height by one pixel.

Fixes #8299.

### DIFF
--- a/src/lib/table/table.scss
+++ b/src/lib/table/table.scss
@@ -20,6 +20,7 @@ $mat-row-horizontal-padding: 24px;
   border-bottom-style: solid;
   align-items: center;
   padding: 0 $mat-row-horizontal-padding;
+  box-sizing: border-box;
 
   // Workaround for https://goo.gl/pFmjJD in IE 11. Adds a pseudo
   // element that will stretch the row the correct height. See:


jelbourn:
@crisbeto lint error:
```
src/material-examples/table-selection/table-selection-example.css
 14:3  ✖  The `initial` value is not supported in IE  declaration-property-value-blacklist
```

jelbourn:
Nvm, not this PR. Weird error.

crisbeto:
AFAIK I opened this one while we had some lint errors in master.

jelbourn:
Caretaker note: causes a fair number of screenshot changes, need to decide if this would be considered breaking

andrewseguin:
Manually merged in a previous commit to rebase

PR #8324, opened 2017-11-09
fix(drawer): invalid margin declaration when rendering server-side

crisbeto:
Fixes the margins of a drawer container being declared as `marginLeft` and `marginRight` when rendering on the server side.


CaerusKaru:
Note: this provides a temporary workaround for an open issue, angular/angular#19235

PR #8327, opened 2017-11-09
docs(slider): change slider value binding to ngModel

### LINKED ISSUE #8166: Material Slider Example Bug
The current material slider example [https://material.angular.io/components/slider/examples](https://material.angular.io/components/slider/examples) currently has a bug - the html binds on `[value]="value"` when it should be using `[(ngModel])="value"`.

The variable on the component should be updated with the slider's current value via 2-way data binding.

The "value" property on the component does not get changed as the slider gets changed. This can be seen right under "Slider Configuration" on the example page which is supposed to show the value of the slider as it is moved. However, it always stays at 0.

Go to the example page (linked above) and observe that the component's "value" property is not actually holding the slider's current value.

Better documentation for users.

Unknown

rodaz:
Fixes #8166

### DIFF
--- a/src/material-examples/slider-configurable/slider-configurable-example.html
+++ b/src/material-examples/slider-configurable/slider-configurable-example.html
@@ -56,7 +56,7 @@
         [step]="step"
         [thumb-label]="thumbLabel"
         [tick-interval]="tickInterval"
-        [value]="value"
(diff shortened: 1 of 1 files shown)


rodaz:
I having some problems with cla. I'd rather create new pr from another account.

PR #8339, opened 2017-11-09
fix(list): multi-line list item spacing

### LINKED ISSUE #8333: More than 3 matLines in a nav list-item creates an incorrect ripple boundary
When a list item in a nav-list has more than 3 lines, the ripple box is of the wrong size and it looks bad

The ripple box and scope of the list item should be consistent

It is not consistent

Click the list item with 4 lines
Click the list item with 3 lines

Notice, that 4 lines has a weird "inner box"

http://plnkr.co/edit/fviuK5?p=preview

To keep consistency

Latest All
Chrome

I had fish and chips for lunch?

crisbeto:
* Fixes the spacing on multi-line list items being off.
* Fixes a couple of declarations with the exact same styles being duplicated, adding some bloat to the final output.

Fixes #8333.

### DIFF
--- a/src/lib/list/list.scss
+++ b/src/lib/list/list.scss
@@ -15,6 +15,7 @@ $mat-list-avatar-height: 56px;
 // spec requires two- and three-line lists be taller
 $mat-list-two-line-height: 72px;
 $mat-list-three-line-height: 88px;
+$mat-list-multi-line-padding: 16px;
 $mat-list-icon-size: 24px;
 
 // Dense list variables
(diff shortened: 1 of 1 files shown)


src/lib/list/list.scss
```diff
@@ -23,12 +24,13 @@ $mat-dense-base-height: 40px;
 $mat-dense-avatar-height: 48px;
 $mat-dense-two-line-height: 60px;
 $mat-dense-three-line-height: 76px;
+$mat-dense-multi-line-padding: 16px;
```
crisbeto: Note that the dense padding here is the same as the regular one, however the spec doesn't seem to mention exactly how to handle more than three lines. I based this number on the spacing between the text and the element top for three-line items.

jelbourn:
Caretaker note: this needs extra caution, perhaps reaching out to teams

PR #8342, opened 2017-11-09
chore(select): organize tests to better scope compilation

jelbourn:
None of the tests are changed, just reorganized. I tweaked a few test names / assertions to fit on one line.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -19,7 +19,15 @@ import {
   ViewChild,
   ViewChildren,
 } from '@angular/core';
-import {ComponentFixture, fakeAsync, flush, inject, TestBed, tick} from '@angular/core/testing';
+import {
+  async,
+  ComponentFixture,
+  fakeAsync,
+  flush,
+  inject,
+  TestBed,
+  tick,
+} from '@angular/core/testing';
 import {
   ControlValueAccessor,
   FormControl,
@@ -47,7 +55,7 @@ import {MatSelect} from './select';
 import {
   getMatSelectDynamicMultipleError,
   getMatSelectNonArrayValueError,
-  getMatSelectNonFunctionValueError
+  getMatSelectNonFunctionValueError,
 } from './select-errors';
 
 /** The debounce interval when typing letters to select an option. */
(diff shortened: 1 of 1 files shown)


src/lib/select/select.spec.ts
```diff
@@ -62,3203 +70,3284 @@ describe('MatSelect', () => {
   let scrolledSubject = new Subject();
   let viewportRuler: ViewportRuler;
 
-  beforeEach(fakeAsync(() => {
+  // Providers used for all mat-select tests
+  const commonProviders = [
+    {
+      provide: OverlayContainer, useFactory: () => {
+      overlayContainerElement = document.createElement('div') as HTMLElement;
```
crisbeto: I might have it in another PR somewhere, but we shouldn't be creating thing element in the first place, we should just inject the `OverlayContainer`.

src/lib/select/select.spec.ts
```diff
@@ -62,3203 +70,3284 @@ describe('MatSelect', () => {
...
+    NoopAnimationsModule,
+  ];
+
+  /**
+   * Configures the test module for MatSelect with the given declarations. This is broken out so
+   * that we're only compiling the necessary test components for each test in order to speed up
+   * overall test time.
+   * @param declarations Components to declare for this block
+   */
+  function configureMatSelectTestingModule(declarations) {
```
crisbeto: The declarations could at least be `any[]` to give it a little bit of type safety.

src/lib/select/select.spec.ts
```diff
@@ -62,3203 +70,3284 @@ describe('MatSelect', () => {
...
 
+        it('should set the `aria-disabled` attribute if the group is disabled', fakeAsync(() => {
+          expect(groups[1].getAttribute('aria-disabled')).toBe('true');
+        }));
+      });
 
-    it('should clear the selection when the control is reset', fakeAsync(() => {
-      fixture.componentInstance.control.setValue('pizza-1');
-      fixture.detectChanges();
+      describe('aria-owns', () => {
```
crisbeto: Is this one supposed to be empty?

src/lib/select/select.spec.ts
```diff
@@ -62,3203 +70,3284 @@ describe('MatSelect', () => {
...
-      fixture.componentInstance.control.enable();
-      fixture.detectChanges();
-      expect(getComputedStyle(trigger).getPropertyValue('cursor'))
-          .toEqual('pointer', `Expected cursor to be a pointer on enabled control.`);
+        const optionInstances = fixture.componentInstance.options.toArray();
 
-      trigger.click();
-      fixture.detectChanges();
+        expect(optionInstances[0].selected)
+            .toBe(false, 'Expected first option to no longer be
```
crisbeto: Since it's a single test it doesn't need the `beforeEach`, it could be moved outside or into the "special cases".

crisbeto: We don't need a `beforeEach` for a single test suite. I doubt that we'll add any more tests to it.

src/lib/select/select.spec.ts
jelbourn: I believe TypeScript infers the type here from how the argument is used

src/lib/select/select.spec.ts
jelbourn: I'd rather address that in a different PR because I don't want to actually change any logic in this one

src/lib/select/select.spec.ts
jelbourn: Fixed- I copied the contents to a different `describe`

src/lib/select/select.spec.ts
```diff
@@ -62,3203 +70,3284 @@ describe('MatSelect', () => {
...
+
+      document.body.appendChild(overlayContainerElement);
+
+      // remove body padding to keep consistent cross-browser
+      document.body.style.padding = '0';
+      document.body.style.margin = '0';
+
+      return {getContainerElement: () => overlayContainerElement};
+    },
+    },
```
crisbeto: The indentation here and the `ScrollDispatcher` override is a little weird. The lambas need to be indented.

src/lib/select/select.spec.ts
crisbeto: It seems like the empty `describe` is still there.

jelbourn:
Done; I hadn't exported my changes when I made the comment

PR #8360, opened 2017-11-10
[refactor] MatRadio (change) -> (selectionChange)

msc117:
refactor(MatRadio): `(change)` -> `(selectionChange)` to be more inline with recent event deprecations in other components


src/lib/radio/radio.ts
```diff
@@ -125,7 +125,16 @@ export class MatRadioGroup extends _MatRadioGroupMixinBase
...
-  @Output() change: EventEmitter<MatRadioChange> = new EventEmitter<MatRadioChange>();
+  @Output() selectionChange: EventEmitter<MatRadioChange> = new EventEmitter<MatRadioChange>();
+
+  /**
+   * Event emitted when the group value changes.
+   * Change events are only emitted when the value changes due to user interaction with
+   * a radio button (the same behavior as `<input type-"radio">`).
+   * @deprecated use `selectionChange` instead.
+   */
+  @Output() change: EventEmitter<MatRadioChange> = this.selectionChange;
```
josephperrott: To aide in eventually removing `change` `()Output` in the future, can you please update the references within the `MatRadioGroup` to call for `selectionChange` instead of the `change`?

msc117: Updated

src/lib/radio/radio.spec.ts
```diff
@@ -316,7 +316,7 @@ describe('MatRadio', () => {
 
     it('should not fire a change event from the group when a radio checked state changes', () => {
```
DevVersion: Can you update the test names from `change`  to `selectionChange` as well?

msc117: Updated

msc117:
Well looks like I messed up the rebase

jelbourn:
We want to keep `change` for checkbox/radio-button since they reflect the native `<input>` API

PR #8367, opened 2017-11-11
chore: fix master failure on MatInput

josephperrott:


jelbourn:
I believe this was already fixed by another PR.

PR #8373, opened 2017-11-11
feat(paginator): allow page size selection to be disabled

### LINKED ISSUE #8359: Paginator doesn't have an option to hide the pageSize value.

Feature request

mat-paginator component should allow the option to hide the pageSize. 

This cannot be done using the mat-paginator component or even the MatPaginatorIntl service.
The service allows to change the label on the left of the pageSize yet it doesn't have an option to completely hide it and the number.

capture

It is desired to have a paginator with only 1 pageSize option so it is not necessary to show it.

Angular Material 2.0.0-beta.12
Angular 4.4.4

But probably all other versions are affected.

crisbeto:
Adds an input that allows the consumer to disable the page size selection entirely.

Fixes #8359.

**Note:** the AoT compilation error is coming from master and will be fixed in another PR.

### DIFF
--- a/src/lib/paginator/paginator.ts
+++ b/src/lib/paginator/paginator.ts
@@ -97,6 +97,9 @@ export class MatPaginator implements OnInit, OnDestroy {
   }
   private _pageSizeOptions: number[] = [];
 
+  /** Whether to hide the page size selection UI from the user. */
+  @Input() hidePageSize = false;
+
   /** Event emitted when the paginator changes the page size or page index. */
   @Output() page = new EventEmitter<PageEvent>();
 
--- a/src/lib/paginator/paginator.html
+++ b/src/lib/paginator/paginator.html
@@ -1,5 +1,5 @@
 <div class="mat-paginator-container">
-  <div class="mat-paginator-page-size">
+  <div class="mat-paginator-page-size" *ngIf="!hidePageSize">
     <div class="mat-paginator-page-size-label">
       {{_intl.itemsPerPageLabel}}
     </div>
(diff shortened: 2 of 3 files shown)


src/lib/paginator/paginator.ts
```diff
@@ -96,6 +96,9 @@ export class MatPaginator implements OnInit, OnDestroy {
   }
   private _pageSizeOptions: number[] = [];
 
+  /** Whether to show the page size selection UI to the user. */
+  @Input() pageSizeEnabled = true;
```
crisbeto: @andrewseguin I was torn between doing this or having some logic where the select is hidden if you pass in an empty array of options. I went with this since it's a bit more obvious what is going on.

nicu-chiciuc: At the moment an empty array will display a label (whereas if there are several elements, it will display a dropdown) so I think overwriting the current behaviour wouldn't have been a good idea. 

@crisbeto btw, thanks for the fast response to the issue.

andrewseguin:
Sorta conflicted on this one - seems like an easy CSS `display: none` for the user if they don't want it part of the UI, but it is nice when there's an API to make it feel more supported and less of a hack. 

How do you feel about renaming as `showPageSize`? Or `hidePageSize` and default to false?

crisbeto:
Renamed to `hidePageSize` @andrewseguin.

tinayuangao:
@crisbeto Could you please fix the tests? Thanks

crisbeto:
I think those failures were coming in from master @tinayuangao. I've rebased it against the current master.

andrewseguin:
Can you verify that the tests are actually passing?

crisbeto:
Sorted out the paginator failures, there are still some failures from master, but they're unrelated.

PR #8376, opened 2017-11-11
docs: show additional interfaces

### LINKED ISSUE #8298: Global Disable Ripple Does Not Affect Keyboard Focus

Bug.

When you set `disabled: true` using `RippleGlobalOptions`, the ripple should disappear completely, no matter if the focus arrives through tab key on keyboard or through mouse click.

After setting `disabled: true` using `RippleGlobalOptions`, the ripple effect still appears when the focus arrives through tab key.

Here is a simple plunker containing two checkboxes: https://plnkr.co/edit/7aqtkq?p=preview

You can see that ripple effect is disabled globally. When you click any checkbox with mouse, the ripple effect does not appear, as expected. However, when you select any checkbox using tab key, the ripple effect still appears.

I guess that when you disable ripple effect, you want to disable it completely.

Angular 4, latest version of Material, Windows, browsers Google Chrome, Internet Explorer.

One can get rid of the ripple effect completely using css `.mat-checkbox-ripple { display: none; }`

DevVersion:
* Shows additional interfaces in the Dgeni API documentation
* Adds documentation for the `RippleGlobalOptions`

Closes #8298.

### DIFF
--- a/tools/dgeni/common/dgeni-definitions.ts
+++ b/tools/dgeni/common/dgeni-definitions.ts
@@ -3,12 +3,14 @@ import {ClassLikeExportDoc} from 'dgeni-packages/typescript/api-doc-types/ClassL
 import {PropertyMemberDoc} from 'dgeni-packages/typescript/api-doc-types/PropertyMemberDoc';
 import {NormalizedMethodMemberDoc} from './normalize-method-parameters';
 
+/** Extended Dgeni class-like document that includes separated class members. */
 export interface CategorizedClassLikeDoc extends ClassLikeExportDoc {
   methods: CategorizedMethodMemberDoc[];
   properties: CategorizedPropertyMemberDoc[];
   isDeprecated: boolean;
 }
 
+/** Extended Dgeni class document that includes extracted Angular metadata. */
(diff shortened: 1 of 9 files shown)


tools/dgeni/common/dgeni-definitions.ts
```diff
@@ -0,0 +1,33 @@
+import {ClassExportDoc} from 'dgeni-packages/typescript/api-doc-types/ClassExportDoc';
+import {ClassLikeExportDoc} from 'dgeni-packages/typescript/api-doc-types/ClassLikeExportDoc';
+import {PropertyMemberDoc} from 'dgeni-packages/typescript/api-doc-types/PropertyMemberDoc';
+import {NormalizedMethodMemberDoc} from './normalize-method-parameters';
+
+export interface CategorizedClassLikeDoc extends ClassLikeExportDoc {
```
jelbourn: Add descriptions for these?

DevVersion: Those are just simple modifications to the real Dgeni documents - added comments to all of the modified definitions in this file.

PR #8377, opened 2017-11-11
docs: add table-selection example

DevVersion:
* Adds the table-selection example to the `ExamplesModule`.


src/material-examples/example-module.ts
```diff
@@ -715,6 +716,12 @@ export const EXAMPLE_COMPONENTS = {
     additionalFiles: null,
     selectorName: null
   },
+  'table-selection': {
```
amcdnl: I assume this was generated and not typed manually? We should probably remove this file from Git even.

DevVersion: Yeah that was just the `gulp` task that generates that file.

PR #8382, opened 2017-11-11
fix(table): allow BehaviorSubject as initial data in constructor

### LINKED ISSUE #8381: Table: Paginator/Sorting for dynamic data does not work (Bug)
Currently it is not possible to use the paginator with dynamic data that comes from Subjects (or initially Observables). The paginator seems to work with __data_ (initialData) set in the MatTableDataSource constructor. You can provide an Array that then transformed into a BehaviorSubject; however passing in a BehaviourSubject directly (e.g. from a service that provides the dynamic data) should be a thing - that would fix the issue described above.

The table should refresh when switching pages via the paginator.

The table does not refresh when switching pages via the paginator.

Unfortunately I can't provide a functional Plunker because the data I'm using comes from my backend app... But you can find the relevant parts here: http://plnkr.co/edit/exXUHnTlOiEipaH97m4n

Basically you setup a regular MatTable and assign a DataSource. Then you return the BehaviourSubject in your DataSource's connect() method. This makes the data appear in the table - but you won't be able to sort or paginate.

Pagination and sorting should also work with dynamic data. ;)

Material 2

It's my third week with Angular. ^^

saitho:
Directly passing a BehaviorSubject as initial data allows sorting and pagination to work with dynamic data.

Closes #8381

### DIFF
--- a/src/lib/table/table-data-source.ts
+++ b/src/lib/table/table-data-source.ts
@@ -129,8 +129,12 @@ export class MatTableDataSource<T> implements DataSource<T> {
     return dataStr.indexOf(transformedFilter) != -1;
   }
 
-  constructor(initialData: T[] = []) {
-    this._data = new BehaviorSubject<T[]>(initialData);
+  constructor(initialData: T[]|BehaviorSubject<T[]> = []) {
+    if (initialData instanceof BehaviorSubject) {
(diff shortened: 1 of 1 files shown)


saitho:
Alternative approach discussed in issue #8381.

PR #8385, opened 2017-11-12
chore(tooltip): unit test failures

crisbeto:
* Fixes a regression in the tooltip that was introduced by 0719c3866225874c2d900ece2554612675ce4d0b which caused the tests to fail.
* Avoids issues in the future where failures in some tooltip tests can cause Jasmine to throw the browser into an infinite loop by trying to stringify a circular object.

### DIFF
--- a/src/lib/tooltip/tooltip.spec.ts
+++ b/src/lib/tooltip/tooltip.spec.ts
@@ -53,7 +53,7 @@ describe('MatTooltip', () => {
           return {getContainerElement: () => overlayContainerElement};
         }},
         {provide: Directionality, useFactory: () => {
-          return dir = { value: 'ltr' };
+          return dir = {value: 'ltr'};
         }}
       ]
     });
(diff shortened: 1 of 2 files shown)


src/lib/tooltip/tooltip.ts
```diff
@@ -261,7 +262,7 @@ export class MatTooltip implements OnDestroy {
     this._tooltipInstance = overlayRef.attach(portal).instance;
 
     // Dispose of the tooltip when the overlay is detached.
-    overlayRef.detachments().subscribe(() => {
+    merge(this._tooltipInstance!.afterHidden(), overlayRef.detachments()).subscribe(() => {
```
jelbourn: Move the `!` up to the assignment for `tooltipInstance`?

src/lib/tooltip/tooltip.spec.ts
```diff
@@ -80,7 +80,7 @@ describe('MatTooltip', () => {
     });
 
     it('should show and hide the tooltip', fakeAsync(() => {
-      expect(tooltipDirective._tooltipInstance).toBeUndefined();
+      expect(!!tooltipDirective._tooltipInstance).toBe(false);
```
jelbourn: Can you make it just one place that does this? Something like a method `getTooltipInstance`? That way it's easy to add a comment in one place that explains it rather than repeating `!!` without explanation

src/lib/tooltip/tooltip.ts
crisbeto: TS isn't smart enough to figure it out when it's being assigned to a property, it only works for variables.

PR #8398, opened 2017-11-12
Force change-detection upon ContentChildren()...

actra-gschuster:
...changes like added / remove steps

### DIFF
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -35,7 +35,6 @@ import {CdkStepLabel} from './step-label';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {AbstractControl} from '@angular/forms';
 import {Direction, Directionality} from '@angular/cdk/bidi';
-import {takeUntil} from 'rxjs/operators/takeUntil';
 
 /** Used to generate unique ID for each stepper component. */
 let nextId = 0;
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -28,12 +28,14 @@ import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   OnChanges,
+  OnDestroy
 } from '@angular/core';
 import {LEFT_ARROW, RIGHT_ARROW, ENTER, SPACE} from '@angular/cdk/keycodes';
 import {CdkStepLabel} from './step-label';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {AbstractControl} from '@angular/forms';
 import {Direction, Directionality} from '@angular/cdk/bidi';
+import {takeUntil} from 'rxjs/operators/takeUntil';
 
 /** Used to generate unique ID for each stepper component. */
 let nextId = 0;
@@ -132,7 +134,10 @@ export class CdkStep implements OnChanges {
   selector: '[cdkStepper]',
   exportAs: 'cdkStepper',
 })
-export class CdkStepper {
+export class CdkStepper implements OnDestroy {
+  /** Emits when the component is destroyed. */
+  protected _destroyed = new Subject<void>();
+  
   /** The list of step components that the stepper is holding. */
   @ContentChildren(CdkStep) _steps: QueryList<CdkStep>;
 
@@ -184,6 +189,11 @@ export class CdkStepper {
     @Optional() private _dir: Directionality,
     private _changeDetectorRef: ChangeDetectorRef) {
     this._groupId = nextId++;
+  }
+  
+  ngOnDestroy() {
+    this._destroyed.next();
+    this._destroyed.complete();
   }
 
   /** Selects and focuses the next step in list. */
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -137,7 +137,7 @@ export class CdkStep implements OnChanges {
 export class CdkStepper implements OnDestroy {
   /** Emits when the component is destroyed. */
   protected _destroyed = new Subject<void>();
-  
+
   /** The list of step components that the stepper is holding. */
   @ContentChildren(CdkStep) _steps: QueryList<CdkStep>;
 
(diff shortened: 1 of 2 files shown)


src/lib/stepper/stepper.ts
```diff
@@ -66,12 +69,23 @@ export class MatStep extends _MatStep implements ErrorStateMatcher {
...
   /** Steps that the stepper holds. */
   @ContentChildren(MatStep) _steps: QueryList<MatStep>;
+  
+  /** Workaround for https://github.com/angular/material2/issues/8397 */
+  _stepChangesSubscription: Subscription;
+  ngAfterContentInit() {
+    this._stepChangesSubscription = this._steps.changes.subscribe(() => this._stateChanged());
+  }
+  ngOnDestroy() {
+    if(this._stepChangesSubscription instanceof Subscription) {
```
mmalerba: A pattern we've started using for situations like this is to create a stream that emits on destory and then use `takeUntil` when describing (e.g. https://github.com/angular/material2/blob/master/src/lib/sidenav/drawer.ts#L467)

src/lib/stepper/stepper.ts
```diff
@@ -66,12 +69,23 @@ export class MatStep extends _MatStep implements ErrorStateMatcher {
...
   @ViewChildren(MatStepHeader, {read: ElementRef}) _stepHeader: QueryList<ElementRef>;
 
   /** Steps that the stepper holds. */
   @ContentChildren(MatStep) _steps: QueryList<MatStep>;
+  
+  /** Workaround for https://github.com/angular/material2/issues/8397 */
+  _stepChangesSubscription: Subscription;
+  ngAfterContentInit() {
+    this._stepChangesSubscription = this._steps.changes.subscribe(() => this._stateChanged());
+  }
```
mmalerba: nit: \n

src/lib/stepper/stepper.ts
actra-gschuster: So MatStep should emit on destroy and MatStepper should subscribe on each step?
To know which steps exist stepper would have to listen on QueryList changes...
Did I get that correctly?

mmalerba: Like this, your `ngOnDestroy` would emit on a subject:
```ts
ngOnDestroy() {
  this._destroyed.next();
  this._destroyed.complete();
}
```

And then you don't have to save the `Subscription` when you subscribe, you can just use `takeUntil`:
```ts
this._steps.changes.pipe(takeUntil(this._destroyed)).subscribe(() => ...);
```

This way if we make more subscriptions later on we don't have to save them all as separate member variables

actra-gschuster:
@mmalerba Applied your suggested changes, added the "emit on destroy" pattern to the upper CDK stepper for other inherited clases to be able to use it without specifying their own _destroy property.

src/lib/stepper/stepper.ts
```diff
@@ -66,12 +67,17 @@ export class MatStep extends _MatStep implements ErrorStateMatcher {
...
   /** The list of step headers of the steps in the stepper. */
   @ViewChildren(MatStepHeader, {read: ElementRef}) _stepHeader: QueryList<ElementRef>;
 
   /** Steps that the stepper holds. */
   @ContentChildren(MatStep) _steps: QueryList<MatStep>;
+  
+  /** Workaround for https://github.com/angular/material2/issues/8397 */
+  ngAfterContentInit() {
+    this._steps.changes.pipe(takeUntil(this._destroyed)).subscribe(() => this._stateChanged());
+  }
```
willshowell: Maybe change the comment to be like this?

```ts
ngAfterContentInit() {
  // Mark the component for change detection whenever the content children query changes
  this._steps.changes.pipe(...).subscribe(...);
}
```

src/cdk/stepper/stepper.ts
```diff
@@ -28,12 +28,14 @@ import {
...
   ChangeDetectorRef,
   OnChanges,
+  OnDestroy
 } from '@angular/core';
 import {LEFT_ARROW, RIGHT_ARROW, ENTER, SPACE} from '@angular/cdk/keycodes';
 import {CdkStepLabel} from './step-label';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
 import {AbstractControl} from '@angular/forms';
 import {Direction, Directionality} from '@angular/cdk/bidi';
+import {takeUntil} from 'rxjs/operators/takeUntil';
```
willshowell: I think this import should be in `src/lib/stepper/stepper.ts` since it isn't used in this file.

actra-gschuster: Agreed, was a mistake.

src/lib/stepper/stepper.ts
actra-gschuster: Agreed, fixed.

src/lib/stepper/stepper.ts
```diff
@@ -66,12 +68,17 @@ export class MatStep extends _MatStep implements ErrorStateMatcher {
...
 })
-export class MatStepper extends _MatStepper {
+export class MatStepper extends _MatStepper implements AfterContentInit {
   /** The list of step headers of the steps in the stepper. */
   @ViewChildren(MatStepHeader, {read: ElementRef}) _stepHeader: QueryList<ElementRef>;
 
   /** Steps that the stepper holds. */
   @ContentChildren(MatStep) _steps: QueryList<MatStep>;
+  
+  /** Mark the component for change detection whenever the content children query changes */
```
mmalerba: minor nit: move this comment inside the method and use `//` style comment since it describes the reason behind some specific logic, not the purpose of the `ngAfterContentInit` method

actra-gschuster:
OK, I think we're done now ;-)

mmalerba:
oh it looks like there's some lint errors and you forgot to import `Subject`, but good to go after that's fixed

mmalerba:
I think you added the import to the wrong file, CI is still red

actra-gschuster:
Woohoo, tests finally passed, linting is very nitpicking about trailing spaces ;-)

willshowell:
@actra-gschuster for next time, you should be able to set your editor to automatically remove trailing spaces. You can also run tests and lint locally so you don't have to push and wait for travis each time:

```
npm run test
npm run tslint
```

PR #8401, opened 2017-11-13
feat([MatMenu) support horizontal (xPosition) overlap detection

### LINKED ISSUE #8400: [mat-menu] support horizontal (xPosition) overlap detection
feature

When [overlapTrigger] is false the menu will NOT overlap on the horizontal axis (xPosition)

The horizontal axis is overlapping the triggering element wether overlapTrigger is true/false.

Per the docs, the overlapTrigger should respect both x and y.
It is also a valid requirement.

angular 5, material 5.0.0-rc0

shlomiassaf:
Closes #8400

### DIFF
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -368,6 +368,8 @@ export class MatMenuTrigger implements AfterContentInit, OnDestroy {
       originFallbackX = overlayX = originX === 'end' ? 'start' : 'end';
       offsetY = overlayY === 'bottom' ? MENU_PANEL_TOP_PADDING : -MENU_PANEL_TOP_PADDING;
     } else if (!this.menu.overlapTrigger) {
+      originX = overlayX === 'start' ? 'end' : 'start';
+      originFallbackX = overlayFallbackX === 'start' ? 'end' : 'start';
       originY = overlayY === 'top' ? 'bottom' : 'top';
       originFallbackY = overlayFallbackY === 'top' ? 'bottom' : 'top';
     }


shlomiassaf:
This is a patch and not tested because the dev environment is throwing errors...

shlomiassaf:
@crisbeto that's probably because `[overlapTrigger]` should be able to describe x or y or both.

Currently false means both and that's the only option.

In the following scenario it is the intent:

image

It is not doable with the current setup, for that to work a workaround is required.
Wrapping the button in a div with position: relative, adding a span at the end and absoluting it to right 0... then setting the trigger on it and manually toggling that trigger based on button clicks which means a method for (click) in the component host...

shlomiassaf:
Maybe it should be `overlapTriggerX` `overlapTriggerY` ...

langan:
@shlomiassaf Did you manage to get the menu to show like in your screenshot?
I am trying to make a similar menu now.

josephperrott:
Since it looks like this PR has stalled out and is pretty out of date, I am going to close out the PR.  Feel free to open up a new PR if you are still looking to get this feature in.

shlomiassaf:
@langan, yes.

You need to add an element to the container of the menu item.

  - The container must be set to have **position: relative**
  - That element is set to **position: absolute**, **top: 0**, **right: 0**

The element has no content, it's empty (i used empty span), it's purpose is to be the host for the trigger.

You use the original trigger host (e.g. button) to trigger the element...

```html
  <div class="menu-item-container">
    <button mat-icon-button
            (mousedown)="trigger._handleMousedown($event)"
            (keydown)="trigger._handleKeydown($event)"
            (click)="trigger._handleClick($event)">
        <mat-icon>more_vert</mat-icon>
    </button>
    <span class="menu-placholder" [matMenuTriggerFor]="menu" #trigger="matMenuTrigger"></span>
  </div>  
```

```css
.menu-item-container {
  position: relative;
}

.menu-placholder {
  position: absolute;
  top: 0;
  right: 0;
}
```

### Notes:
I'm using the button events directly and pass them to the `trigger`'s original handlers. It might seems odd because there is an api for that, `trigger.toggleMenu()` but it will cause issues with focus because you can "tell" `toggleMenu` if to auto-focus or not... because of that it will focus on a menu item inside (sub menus)

This is why `mousedown` is passed, as it flags the action not to auto focus.

See full example here: 
https://stackblitz.com/edit/angular-jwihkn?file=app%2Fmenu-overview-example.html

Issue #8403, opened 2017-11-13
Expansion panel memory leak


AnthonNilsson:
#### Bug, feature request, or proposal:
I have an SPA that uses angular material2. When using mat-expansion-panel I get rising number Detatched DOM trees if removing/adding the expansion panel.

#### What are the steps to reproduce?
If browsing to https://material.angular.io/components/expansion/examples and clicking on API -> Examples -> API and so on this behaviour could be reproduced. First after entering the page chrome heap snapshot shows:

![image](https://user-images.githubusercontent.com/7062803/32706140-4d01b07a-c81b-11e7-96a7-f785aaea5b40.png)

After clicking the tabs API, and Examples 10 times it shows:
![image](https://user-images.githubusercontent.com/7062803/32706181-901c3d12-c81b-11e7-9122-2a40e3ca846a.png)

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
In my SPA I use:
* Angular 5.0.1
* Angular Material: 5.0.0-rc0
* Angular CLI 1.5.0

josephperrott:
Closed via #8410

PR #8404, opened 2017-11-13
Fixing memory leak for expansion-panel (#8403)

AnthonNilsson:


Emanjonil:
I signed it!

src/lib/expansion/expansion-panel.ts
```diff
@@ -142,6 +142,7 @@ export class MatExpansionPanel extends _MatExpansionPanelMixinBase
   }
 
   ngOnDestroy() {
+    super.ngOnDestroy();
```
crisbeto: This used to be necessary in Angular 2.x, but it should be called automatically in the newer versions.

AnthonNilsson: Thanks for the response!

But this does not seem to be the case here, not in my SPA application at least. I am using angular 5.0.1 and material2 5.0.0-rc0 together with angular cli 1.5.0.
And if checking my answer in (#8403) this problem also seems to be existing in the angular material homepage.

I get a Detatched DOM tree for each expansion panel I have disposed. The ngOnDestroy is never called in CdkAccordionItem.

If manually changing in the material2 js generated code, for MatExpansionPanel, in my node_modules folder expansion.js
image

or for es5 expansion.es5.js:
image

Then the ngOnDestroy is called in CdkAccordionItem and the issue with Detatched DOM trees are removed.

AnthonNilsson:
I am very sorry but after failing to correct the Author information to my correct email I have now made a new pull request which contains my correct email address for Author. (#8410)

PR #8410, opened 2017-11-13
Fixing memory leak for expansion-panel (#8403)

AnthonNilsson:
This pull request is to replace (#8404), please check this for more information.

### DIFF
--- a/src/lib/expansion/expansion-panel.ts
+++ b/src/lib/expansion/expansion-panel.ts
@@ -142,6 +142,7 @@ export class MatExpansionPanel extends _MatExpansionPanelMixinBase
   }
 
   ngOnDestroy() {
+    super.ngOnDestroy();
     this._inputChanges.complete();
   }
 }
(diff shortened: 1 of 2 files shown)


src/lib/expansion/expansion-panel.ts
```diff
@@ -142,6 +142,7 @@ export class MatExpansionPanel extends _MatExpansionPanelMixinBase
   }
 
   ngOnDestroy() {
+    super.ngOnDestroy();
```
jelbourn: Can you add a unit test for this change?

AnthonNilsson: I really tried to come up with a unit test, but I have not written unit tests for Jasmin/Karma before so I didnt really get it to work.

This is what I came up with but chrome hangs during execution:

    it('make sure accordion item runs ngOnDestroy when expansion panel is destroyed', () => {
      let fixture = TestBed.createComponent(PanelWithContent);
      spyOn(fixture.componentInstance.panel.destroyed, 'emit');
      fixture.componentInstance.panel.ngOnDestroy();      
      fixture.detectChanges();
      expect(fixture.componentInstance.panel.destroyed.emit).toHaveBeenCalled();
    });

AnthonNilsson: Ok, so I think I managed to make a correct unit test finally.

PR #8421, opened 2017-11-13
fix(slide-toggle): drag not working in edge

### LINKED ISSUE #8391: slide-toggle doesn't work as expected with EdgeBrowser and IE11
Slide-toggle doesn't work as expected with EdgeBrowser and IE11.
Click thumb works.
Drag thumb works but it flips back

Slide-toggle should work with drag thumb

slidetoggle

Look at [Material Demo Site](https://material.angular.io/components/slide-toggle/overview) or the [Plunker](https://goo.gl/uDmqyY)

MSEdge: 41.16299.15.0
Material: 5.0.0-rc0-3ca801a

DevVersion:
Some browsers like Edge or IE11 emit `click` events differently than other browsers like Chrome. For example does Edge still emit a `click` event if the pointer drags something around and is being released over the `<label>` element. This doesn't happen on Chrome.

Right now this means that the new checked state from the drag is being reverted by the `click` event that then causes a `change` event on the underlying input element. To ensure that the behavior is the same across all supported browsers, the checked state from the `click` event is being ignored while dragging.

Fixes #8391

### DIFF
--- a/src/lib/slide-toggle/slide-toggle.spec.ts
+++ b/src/lib/slide-toggle/slide-toggle.spec.ts
@@ -475,6 +475,29 @@ describe('MatSlideToggle without forms', () => {
           .toBeFalsy('Expected the slide-toggle to not emit a change event.');
     }));
 
+    it('should ignore clicks on the label element while dragging', fakeAsync(() => {
+      expect(slideToggle.checked).toBe(false);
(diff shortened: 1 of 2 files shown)


tinayuangao:
LGTM

PR #8424, opened 2017-11-13
fix(directionality): change event now emit the new value

EladBezalel:


jelbourn:
@EladBezalel this has CI errors

EladBezalel:
fixed

PR #8425, opened 2017-11-13
fix(chip-list): fix error state changes in chip list

### LINKED ISSUE #8366: ChipList does not update errorState on parent form submission
Bug

Mat-form-field shows mat-error message on submit

Submit does not trigger the mat-form-field error to be shown. After submitting the error message is hidden until the user interacts with the form. The input shows as invalid (switches to a red line), but the message is not displayed.

http://plnkr.co/edit/KJczgN?p=preview

The other matFormFieldControls show mat errors on submit, seems the chip list should be consistent with the other controls.

Angular Material Beta 12
Angular 4.4.6

The other matFormFieldControls check for errors on change detection
https://github.com/angular/material2/blob/master/src/lib/input/input.ts#L205
then publish a state change when their errorState has changed
https://github.com/angular/material2/blob/master/src/lib/input/input.ts#L243

The chipList instead uses a getter for errorState
https://github.com/angular/material2/blob/master/src/lib/chips/chip-list.ts#L256

It seems that the chip-list should be refactored to use a similar pattern to input where it checks the error on each change detection cycle so that it can update after parent form submission.

tinayuangao:
Fixes #8366

### DIFF
--- a/src/lib/chips/chip-list.spec.ts
+++ b/src/lib/chips/chip-list.spec.ts
@@ -4,7 +4,7 @@ import {BACKSPACE, DELETE, ENTER, LEFT_ARROW, RIGHT_ARROW, SPACE, TAB} from '@an
 import {createKeyboardEvent, dispatchFakeEvent, dispatchKeyboardEvent} from '@angular/cdk/testing';
 import {Component, DebugElement, QueryList, ViewChild, ViewChildren} from '@angular/core';
 import {async, ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
-import {FormControl, FormsModule, ReactiveFormsModule} from '@angular/forms';
+import {FormControl, FormsModule, NgForm, ReactiveFormsModule, Validators} from '@angular/forms';
 import {MatFormFieldModule} from '@angular/material/form-field';
 import {By} from '@angular/platform-browser';
 import {NoopAnimationsModule} from '@angular/platform-browser/animations';
@@ -35,6 +35,7 @@ describe('MatChipList', () => {
         NoopAnimationsModule
       ],
       declarations: [
+        ChipListWithFormErrorMessages,
         StandardChipList,
         FormFieldChipList,
         BasicChipList,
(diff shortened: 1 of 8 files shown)


src/lib/chips/chip-list.ts
```diff
@@ -514,6 +530,20 @@ export class MatChipList implements MatFormFieldControl<any>, ControlValueAccess
     this._lastDestroyedIndex = null;
   }
 
+  /** Re-evaluates the error state. This is only relevant with @angular/forms. */
+  protected _updateErrorState() {
```
jelbourn: This looks like the same logic as select and input- should it be a method in core? Or perhaps a mixin?

src/lib/chips/chip-list.spec.ts
```diff
@@ -1085,3 +1195,30 @@ class SelectedChipList {
...
+    <mat-error>Should have value</mat-error>
+  </mat-form-field>
+</form>
+  `
+})
+class ChipListWithFormErrorMessages {
+  foods: any[] = [
+    { value: 0, viewValue: 'Steak', selected: true },
+    { value: 1, viewValue: 'Pizza', selected: false },
+    { value: 2, viewValue: 'Pasta', selected: true },
```
jelbourn: Omit spaces inside braces

tinayuangao:
Added mixin for error state. Please take a look. Thanks!

src/lib/core/common-behaviors/error-state.ts
```diff
@@ -0,0 +1,66 @@
...
+  _parentFormGroup: FormGroupDirective;
+  _parentForm: NgForm;
+  _defaultErrorStateMatcher: ErrorStateMatcher;
+  ngControl: NgControl;
+}
+
+/** Possible color palette values.  */
+export type ThemePalette = 'primary' | 'accent' | 'warn' | undefined;
+
+/** Mixin to augment a directive with a `color` property. */
```
jelbourn: Comments still mention `color`

src/lib/core/common-behaviors/error-state.ts
```diff
@@ -0,0 +1,66 @@
...
+/** @docs-private */
+export interface HasErrorState {
+  _parentFormGroup: FormGroupDirective;
+  _parentForm: NgForm;
+  _defaultErrorStateMatcher: ErrorStateMatcher;
+  ngControl: NgControl;
+}
+
+/** Possible color palette values.  */
+export type ThemePalette = 'primary' | 'accent' | 'warn' | undefined;
```
jelbourn: Can delete this

src/lib/core/error/error-options.ts
```diff
@@ -24,3 +25,31 @@ export class ErrorStateMatcher {
     return !!(control && control.invalid && (control.touched || (form && form.submitted)));
   }
 }
+
+
+export abstract class HasErrorStateComponent {
```
jelbourn: Add a description for this class and how it's meant to be used?

src/lib/core/error/error-options.ts
```diff
@@ -24,3 +25,31 @@ export class ErrorStateMatcher {
...
+
+    if (newState !== oldState) {
+      this.errorState = newState;
+      this.stateChanges.next();
+    }
+  }
+
+}
+
+export function updateErrorState() {}
```
jelbourn: No-op function?

src/lib/input/input.ts
```diff
@@ -43,6 +43,15 @@ const MAT_INPUT_INVALID_TYPES = [
 
 let nextUniqueId = 0;
 
+// Boilerplate for applying mixins to MatInput.
+/** @docs-private */
+export class MatInputBase {
+  constructor(public _defaultErrorStateMatcher: ErrorStateMatcher,
+              @Optional() public _parentForm: NgForm,
+              @Optional() public _parentFormGroup: FormGroupDirective,
+              @Optional() @Self() public ngControl: NgControl) {}
```
jelbourn: Are the decorators actually needed here? AFAIK the metadata extractor will ignore them since the base class for the `@Component` is the result of a function call

src/lib/chips/chip-list.spec.ts
```diff
@@ -864,6 +865,115 @@ describe('MatChipList', () => {
...
+        expect(containerEl.classList)
+          .toContain('mat-form-field-invalid', 'Expected container to have the invalid CSS class.');
+        expect(containerEl.querySelectorAll('mat-error').length)
+          .toBe(1, 'Expected one error message to have been rendered.');
+        expect(chipListEl.getAttribute('aria-invalid'))
+          .toBe('true', 'Expected aria-invalid to be set to "true".');
+      });
+    }));
+
+    it('should display an error message when the parent form is submitted', async(() => {
```
jelbourn: Would it work to make these tests `fakeAsync` instead of `async`?

src/lib/core/error/error-options.ts
tinayuangao: Deleted.

tinayuangao:
Thanks for review. Comments addressed. Please take another look. Thanks!

src/lib/chips/chip-list.ts
```diff
@@ -174,6 +187,9 @@ export class MatChipList implements MatFormFieldControl<any>, ControlValueAccess
     return this.empty ? null : 'listbox';
   }
 
+  /** An object used to control when error messages are shown. */
+  @Input() errorStateMatcher: ErrorStateMatcher;
```
mmalerba: I wonder if it would be better (at some point, not in this CL) to make `[errorStateMatcher]` its own directive. That way we could also place it on a `<div>`, `<form>`, etc. to change it for all its children

src/lib/chips/chip-list.spec.ts
jelbourn: Still seems to have the spaces, should be, e.g.
```ts
{value: 0, viewValue: 'Steak', selected: true},
```

jelbourn:
Just noticed there are a few lint errors; feel free to add "merge-ready" when resolved

PR #8431, opened 2017-11-14
fix(button): only show focus indicator for keyboard

### LINKED ISSUE #8420: A button ripple effect appears again after a dialog is closed

A bug.

The ripple darkening effect shouldn't appear on a button after a dialog is closed.

A ripple / CSS effect appears on the 'invoking' button again after the dialog is closed.

From:
https://material.angular.io/components/dialog/overview

On:
Firefox 56.0.2
Chrome 61.0.3163.100

Pressing the "Pick one" button, the ripple effect is played and the dialog is opened, then closing the dialog in any way return the effect on the button. 

https://i.imgur.com/oOmbAmS.gif

It's visually 'bugged'

It appears in both local build
Angular 5
Material 5.0.0-rc0

And with a refreshed documentation / example page

DevVersion:
* No longer shows the focus indicator when the button is focused through `program`. This behavior should be similar across components like slide-toggle, checkbox, radio etc.

Fixes #8420

### DIFF
--- a/src/lib/button/_button-base.scss
+++ b/src/lib/button/_button-base.scss
@@ -51,10 +51,8 @@ $mat-mini-fab-padding: 8px !default;
     cursor: default;
   }
 
-  &.cdk-keyboard-focused, &.cdk-program-focused {
-    .mat-button-focus-overlay {
-      opacity: 1;
-    }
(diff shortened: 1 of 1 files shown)


mmalerba:
@crisbeto is this undoing a change you made intentionally. I seem to remember a PR where you changed something like this

PR #8437, opened 2017-11-14
build: fix travis required tests not running

DevVersion:
* Fixes that the `travis_required` CI mode isn't running. This mode is important for Firefox and Chrome testing.

### DIFF
--- a/src/lib/input/input.spec.ts
+++ b/src/lib/input/input.spec.ts
@@ -110,7 +110,7 @@ describe('MatInput without forms', function () {
 
   it('should not be treated as empty if type is date',
       inject([Platform], (platform: Platform) => {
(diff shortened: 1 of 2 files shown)


DevVersion:
@jelbourn Fixed the test failure in Firefox 57. Everything is green now!

src/lib/input/input.spec.ts
```diff
@@ -123,7 +123,7 @@ describe('MatInput without forms', function () {
   // Firefox, Safari Desktop and IE don't support type="date" and fallback to type="text".
   it('should be treated as empty if type is date on Firefox and IE',
       inject([Platform], (platform: Platform) => {
-        if (platform.TRIDENT || platform.FIREFOX || (platform.SAFARI && !platform.IOS)) {
+        if (platform.TRIDENT || (platform.SAFARI && !platform.IOS)) {
```
jelbourn: Update the comment above?

PR #8441, opened 2017-11-14
WIP: feat(fab): Create extendable fab, splitting fab from button

josephperrott:


src/lib/fab/tsconfig-es5.json
```diff
@@ -0,0 +1,14 @@
+{
```
CaerusKaru: This file is no longer necessary after #8203

PR #8442, opened 2017-11-14
fix(fab-buttons): vertically align icons inside fab buttons

EladBezalel:
- added `.mat-fab` & `.mat-mini-fab`to the vertical align definition of buttons

### DIFF
--- a/src/lib/button/button.scss
+++ b/src/lib/button/button.scss
@@ -51,7 +51,7 @@
 }
 
 // The text and icon should be vertical aligned inside a button
-.mat-button, .mat-raised-button, .mat-icon-button {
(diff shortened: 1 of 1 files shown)


src/lib/button/button.scss
```diff
@@ -51,7 +51,7 @@
 }
 
 // The text and icon should be vertical aligned inside a button
-.mat-button, .mat-raised-button, .mat-icon-button {
+.mat-button, .mat-raised-button, .mat-icon-button, .mat-fab, .mat-mini-fab {
```
jelbourn: Does this change the `color` in any circumstances?

EladBezalel: `mat-icon`

jelbourn:
Caretaker note: do a global run for this one, be on the lookout for screenshot failures

PR #8455, opened 2017-11-15
fix(paginator): set default display value

### LINKED ISSUE #8454: The style property 'display' of the mat-paginator component should set to 'display: block' by default
Proposal

The mat-paginator is usually used above or under a collection, table, etc.
Therefore `display: block;` should be used to render correctly its width and borders if modified externally by the `.mat-paginator` selector.

There is no `display` style property defined on the host.

You can actually see that the style property is missing in the source code:
https://github.com/angular/material2/blob/master/src/lib/paginator/paginator.scss
https://github.com/angular/material2/blob/master/src/lib/paginator/paginator.ts

The mat-paginator is usually used above or under a collection, table, etc.
Therefore `display: block;` should be used to render correctly its width and borders if modified externally by a `.mat-paginator` selector.

Angular: 5.0.0
Material2: 5.0.0-rc0
OS: Debian 8
TypeScript: 2.4.2
Browser: Chrome, Firefox, etc.

No

crisbeto:
Makes the `mat-paginator` a block-level element by default, instead of inline.

Fixes #8454.

### DIFF
--- a/src/lib/paginator/paginator.scss
+++ b/src/lib/paginator/paginator.scss
@@ -18,6 +18,10 @@ $mat-paginator-button-icon-width: 8px;
 $mat-paginator-button-decrement-icon-margin: 12px;
 $mat-paginator-button-increment-icon-margin: 16px;
 
+.mat-paginator {
+  display: block;
+}
+
 .mat-paginator-container {
   display: flex;
   align-items: center;


andrewseguin:
My understanding was that we didn't want to put `display` styles on outer component containers. Have we done this yet with any other components?

crisbeto:
AFAIK the rule is that we don’t do flexbox and margins on the outer component.

willshowell:
@andrewseguin FWIW cards currently do this
https://github.com/angular/material2/blob/541a95eebad36062976fc188b2120e7995367a68/src/lib/card/card.scss#L11-L22

PR #8458, opened 2017-11-15
feat(tree): Add material styled tree

tinayuangao:
Demo: https://mat-tree-tina-demo.firebaseapp.com/tree

### DIFF
--- a/src/cdk/tree/control/nested-tree-control.ts
+++ b/src/cdk/tree/control/nested-tree-control.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {Observable} from 'rxjs/Observable';
-import {first} from 'rxjs/operators';
+import {take} from 'rxjs/operators/take';
 import {BaseTreeControl} from './base-tree-control';
 
 /** Nested tree control. Able to expand/collapse a subtree recursively for NestedNode type. */
@@ -40,7 +40,7 @@ export class NestedTreeControl<T> extends BaseTreeControl<T> {
   /** A helper function to get descendants recursively. */
   protected _getDescendants(descendants: T[], dataNode: T): void {
     descendants.push(dataNode);
-    first.call(this.getChildren(dataNode)).subscribe(children => {
+    this.getChildren(dataNode).pipe(take(1)).subscribe(children => {
       if (children && children.length > 0) {
         children.forEach((child: T) => this._getDescendants(descendants, child));
       }
--- a/src/cdk/tree/tree.ts
+++ b/src/cdk/tree/tree.ts
@@ -23,7 +23,7 @@ import {
   ViewEncapsulation,
 } from '@angular/core';
 import {BehaviorSubject} from 'rxjs/BehaviorSubject';
-import {takeUntil} from 'rxjs/operators';
+import {takeUntil} from 'rxjs/operators/takeUntil';
 import {Subject} from 'rxjs/Subject';
 import {CdkTreeNodeDef, CdkTreeNode, CdkTreeNodeOutletContext} from './node';
 import {CdkTreeNodeOutlet} from './outlet';
@@ -155,7 +155,7 @@ export class CdkTree<T> implements CollectionViewer, OnInit, OnDestroy {
 
   /** Set up a subscription for the data provided by the data source. */
   private _observeRenderChanges() {
-    takeUntil.call(this.dataSource.connect(this), this._destroyed)
+    this.dataSource.connect(this).pipe(takeUntil(this._destroyed))
       .subscribe(data => {
         this._data = data;
         this._renderNodeChanges(data);
@@ -191,7 +191,7 @@ export class CdkTree<T> implements CollectionViewer, OnInit, OnDestroy {
     if (this._nodeDefs.length == 1) { return this._nodeDefs.first; }
 
     const nodeDef =
-      this._nodeDefs.find(def => def.when && def.when(data, i)) || this._defaultNodeDef;
+      this._nodeDefs.find(def => def.when && def.when(i, data)) || this._defaultNodeDef;
     if (!nodeDef) { throw getTreeMissingMatchingNodeDefError(); }
 
     return nodeDef;
@@ -214,7 +214,9 @@ export class CdkTree<T> implements CollectionViewer, OnInit, OnDestroy {
     // Set the data to just created `CdkTreeNode`.
     // The `CdkTreeNode` created from `createEmbeddedView` will be saved in static variable
     //     `mostRecentTreeNode`. We get it from static variable and pass the node data to it.
-    CdkTreeNode.mostRecentTreeNode.data = nodeData;
+    if (CdkTreeNode.mostRecentTreeNode) {
+      CdkTreeNode.mostRecentTreeNode.data = nodeData;
+    }
 
     this._changeDetectorRef.detectChanges();
   }
--- a/src/lib/tree/node.ts
+++ b/src/lib/tree/node.ts
@@ -1,0 +1,73 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {
+  ContentChildren,
+  Directive,
+  Input,
+  QueryList
+} from '@angular/core';
+import {
+  CdkNestedTreeNode,
+  CdkTreeNodeDef,
+  CdkTreeNode,
+} from '@angular/cdk/tree';
+import {MatTreeNodeOutlet} from './outlet';
+
+/**
+ * Wrapper for the CdkTree node with Material design styles.
+ */
+// TODO(tinayuangao): use mixinTabIndex
+@Directive({
+  selector: 'mat-tree-node',
+  exportAs: 'matTreeNode',
+  host: {
+    '[attr.role]': 'role',
+    'class': 'mat-tree-node',
+    'tabindex': '0',
+  },
+  providers: [{provide: CdkTreeNode, useExisting: MatTreeNode}]
+})
+export class MatTreeNode<T> extends CdkTreeNode<T> {
+  @Input() role: 'treeitem' | 'group' = 'treeitem';
+}
+
+/**
+ * Wrapper for the CdkTree node definition with Material design styles.
+ */
+@Directive({
+  selector: '[matTreeNodeDef]',
+  inputs: [
+    'when: matTreeNodeDefWhen'
+  ],
+  providers: [{provide: CdkTreeNodeDef, useExisting: MatTreeNodeDef}]
+})
+export class MatTreeNodeDef<T> extends CdkTreeNodeDef<T> {
+  @Input('matTreeNode') data: T;
+}
+
+/**
+ * Wrapper for the CdkTree nested node with Material design styles.
+ */
+@Directive({
+  selector: 'mat-nested-tree-node',
+  exportAs: 'matNestedTreeNode',
+  host: {
+    '[attr.role]': 'role',
+    'class': 'mat-nested-tree-node',
+  },
+  providers: [
+    {provide: CdkNestedTreeNode, useExisting: MatNestedTreeNode},
+    {provide: CdkTreeNode, useExisting: MatNestedTreeNode}
+  ]
+})
+export class MatNestedTreeNode<T> extends CdkNestedTreeNode<T> {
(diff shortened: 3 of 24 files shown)


src/cdk/tree/control/nested-tree-control.ts
```diff
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {Observable} from 'rxjs/Observable';
-import {first} from 'rxjs/operators';
+import {first} from 'rxjs/operators/first';
```
TishchenkoAlex: [https://github.com/angular/material2/pull/8352](url)
according to this PR "take(1)" may be better than "first" operator

src/lib/tree/tree.spec.ts
```diff
@@ -0,0 +1,201 @@
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
+import {Component, ViewChild} from '@angular/core';
+
+import {CollectionViewer, DataSource} from '@angular/cdk/collections';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+import {Observable} from 'rxjs/Observable';
+import {combineLatest} from 'rxjs/observable/combineLatest';
+import {map} from 'rxjs/operators';
```
rafaelss95: Add `/map`?

jelbourn:
@tinayuangao rebase?

benb7760:
looks like it has been rebased onto `master` instead of onto `tree`

tinayuangao:
Rebased. Please take a look. Thanks!

mmalerba:
Could you add a material tree demo to the demo app? It's kind of hard to reason about the CSS without being able to see a demo

src/cdk/tree/tree.ts
```diff
@@ -214,7 +214,9 @@ export class CdkTree<T> implements CollectionViewer, OnInit, OnDestroy {
     // Set the data to just created `CdkTreeNode`.
     // The `CdkTreeNode` created from `createEmbeddedView` will be saved in static variable
     //     `mostRecentTreeNode`. We get it from static variable and pass the node data to it.
-    CdkTreeNode.mostRecentTreeNode.data = nodeData;
+    if (CdkTreeNode.mostRecentTreeNode) {
```
jelbourn: Under what circumstances would this be undefined?

src/lib/tree/node.ts
```diff
@@ -0,0 +1,74 @@
...
+} from '@angular/core';
+import {
+  CdkNestedTreeNode,
+  CdkTreeNodeDef,
+  CdkTreeNode,
+} from '@angular/cdk/tree';
+import {MatTreeNodeOutlet} from './outlet';
+
+
+/** Workaround for https://github.com/angular/angular/issues/17849 */
```
jelbourn: That issue is closed- is this still necessary? If so we should file a new issue.

src/lib/tree/node.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Wrapper for the CdkTree node with Material design styles.
+ */
+@Directive({
+  selector: 'mat-tree-node',
+  exportAs: 'matTreeNode',
+  host: {
+    '[attr.role]': 'role',
+    'class': 'mat-tree-node',
+    'tabindex': '0',
```
jelbourn: Add `TODO` for applying `mixinTabIndex`?

src/lib/tree/outlet.ts
```diff
@@ -0,0 +1,22 @@
...
+} from '@angular/core';
+
+/**
+ * Outlet for nested CdkNode. Put `[matTreeNodeOutlet]` on a tag to place children dataNodes
+ * inside the outlet.
+ */
+@Directive({
+  selector: '[matTreeNodeOutlet]'
+})
+export class MatTreeNodeOutlet {
```
jelbourn: Should add here `implements CdkTreeNodeOutlet` to make sure they stay the same API

src/lib/tree/tree-module.ts
```diff
@@ -0,0 +1,35 @@
...
+import {CdkTreeModule} from '@angular/cdk/tree';
+import {CommonModule} from '@angular/common';
+import {MatCommonModule} from '@angular/material/core';
+import {MatNestedTreeNode, MatTreeNodeDef, MatTreeNode} from './node';
+import {MatTree} from './tree';
+import {MatTreeNodeTrigger} from './trigger';
+import {MatTreeNodeOutlet} from './outlet';
+import {MatTreeNodePadding} from './padding';
+
+const EXPORTED_DECLARATIONS = [
```
jelbourn: `MAT_TREE_DIRECTIVES`?

src/cdk/tree/tree.ts
tinayuangao: We have to initialize `mostRecentTreeNode` because of the error ` /usr/local/google/home/tinagao/WebstormProjects/material2/src/cdk/tree/node.ts:79:10: Metadata collected contains an error that will be reported at runtime: Only initialized variables and constants can be referenced because the value of this variable is needed by the template compiler.`

So I initialize the static variable to be null, and we have to check whether it's null here.

But the value shouldn't be `null` or `undefined` when we call it from `tree.ts`

src/lib/tree/node.ts
```diff
@@ -0,0 +1,69 @@
...
+  },
+  providers: [{provide: CdkTreeNode, useExisting: MatTreeNode}]
+})
+export class MatTreeNode<T> extends CdkTreeNode<T> {
+  @Input('matTreeNode') data: T;
```
benb7760: @tinayuangao `MatTreeNode<T>` has Input `matTreeNode`, but the equivalent `CdkTreeNode<T>` does not have an Input called `cdkTreeNode`. This means that the same code to create a mat-tree isn't working when switched to cdk-tree

tinayuangao: This should be removed. Thanks.

src/lib/tree/tree.md
```diff
@@ -0,0 +1,50 @@
...
+### Features
+
+The `<mat-tree>` itself only deals with the rendering of a tree structure.
+Additional features can be built on top of the tree by adding behavior inside node templates
+(e.g., padding and trigger). Interactions that affect the
+rendered data (such as expand/collapse) should be propagated through the table's data source.
+
+### TreeControl
+
+The `TreeControl` expand/collapse of the tree node. Users can expand/collapse a tree node
```
jelbourn: The first sentence sounds like it's missing something. 
```markdown
The `TreeControl` controls the expand/collapse state of tree nodes.
```

After that, it would be good to explain the two types of `TreeControl` and when they'd be used before going into further detail

src/lib/tree/tree.md
```diff
@@ -0,0 +1,50 @@
...
+ 
+### Trigger
+
+A `matTreeNodeTrigger` can be added in the tree node template to expand/collapse the tree node. The
+trigger triggers the expand/collapse functions in `TreeControl` and is able to expand/collapse a 
+tree node recursively by setting `[matTreeNodeTriggerRecursive]` to `true`. 
+
+The trigger can be placed anywhere in the tree node, and is only triggered by `click` action. 
+
+The trigger is also work with both flat tree and nested tree.
```
jelbourn: I think you can omit this last sentence about the trigger; if we don't say anything, the default would be that it can be used in either.

src/lib/tree/tree.md
```diff
@@ -0,0 +1,50 @@
+The `mat-tree` provides a Material Design styled tree that can be used to display hierarchy
+data.
+
+This tree builds on the foundation of the CDK tree and uses a similar interface for its
+data source input and template, except that its element and attribute selectors will be prefixed 
+with `mat-` instead of `cdk-`.
```
jelbourn: Do you plan on adding examples in a follow-up PR?

src/lib/tree/tree.md
```diff
@@ -0,0 +1,50 @@
...
+
+In the near future, we will provide a simplified version of the tree with an easy-to-use
+interface, material styling, and json input.
+
+### Accessibility
+Trees without text or labels should be given a meaningful label via `aria-label` or
+`aria-labelledby`. The `aria-readonly` defaults to `true` if it's not set.
+
+Tree's role is `tree`.
+Node's default role is `treeitem`, and it will be changed to `group` if it has descendant nodes.
```
jelbourn: I would phrase this as
```markdown
Parent nodes are given `role="group"`, while leaf nodes are given `role="treeitem"`
```

src/lib/tree/tree.md
```diff
@@ -0,0 +1,50 @@
...
+trigger triggers the expand/collapse functions in `TreeControl` and is able to expand/collapse a 
+tree node recursively by setting `[matTreeNodeTriggerRecursive]` to `true`. 
+
+The trigger can be placed anywhere in the tree node, and is only triggered by `click` action. 
+
+The trigger is also work with both flat tree and nested tree.
+
+### Padding
+
+The `matTreeNodePadding` can be placed in a flat tree's node template to display the `level`
```
jelbourn: I'd more strongly emphasize that this is only for flat variations

src/lib/tree/tree.md
```diff
@@ -0,0 +1,50 @@
+The `mat-tree` provides a Material Design styled tree that can be used to display hierarchy
+data.
+
+This tree builds on the foundation of the CDK tree and uses a similar interface for its
```
jelbourn: I would mention the concept of flat vs. nested trees here, as early as possible.

src/lib/tree/tree.md
tinayuangao: Yes. I can add placeholders

src/lib/tree/tree.md
tinayuangao: Added flat tree and nested tree part. Please take a look. Thanks!

src/lib/tree/tree.md
```diff
@@ -0,0 +1,90 @@
...
+data source input and template, except that its element and attribute selectors will be prefixed 
+with `mat-` instead of `cdk-`.
+
+There are two types of trees: Flat tree and Nested Tree. The DOM structures are different for these
+two types of trees.
+
+#### Flat tree
+Flat tree has a list of flattened tree nodes. The hierarchy are flattened and no tree node is placed
```
jelbourn:
```markdown
In a flat tree, the hierarchy is flattened; nodes are not rendered inside of each other,
but instead are rendered as siblings in sequence. An instance of `TreeFlattener` is
used to generate the flat list of items from hierarchical data. The "level" of each tree
node is read through the `getLevel` method of the `TreeControl`; this level can be
used to style the node such that it is indented to the appropriate level.
```

Is that accurate?

src/lib/tree/tree.md
```diff
@@ -0,0 +1,90 @@
...
+expanded/collapsed.
+
+```
+<mat-tree>
+    <mat-tree-node> parent node </mat-tree-node>
+    <mat-tree-node> -- child node1 </mat-tree-node>
+    <mat-tree-node> -- child node2 </mat-tree-node>
+</mat-tree>
+```
+A flat tree is easier to implement virtual scrolling in the future or render only part of the tree.
```
jelbourn:
```markdown
Flat trees are generally easier to style and inspect. They are also more friendly to
scrolling variations, such as infinite or virtual scrolling
```

src/lib/tree/tree.md
```diff
@@ -0,0 +1,90 @@
...
+<mat-tree>
+   <mat-nested-tree-node>
+       parent node
+       <mat-nested-tree-node> -- child node1 </mat-tree-node>
+       <mat-nested-tree-node> -- child node2 </mat-tree-node>
+   </mat-nested-tree-node>
+</mat-tree>
+```
+
+A nested tree is easier to decorate to show hierarchy information.
```
jelbourn: Maybe
```markdown
Nested trees are easier to work with when hierarchical relationships are visually
represented in ways that would be difficult to accomplish with flat nodes.
```
?

src/lib/tree/tree.md
```diff
@@ -0,0 +1,90 @@
...
+The `matTreeNodePadding` can be placed in a flat tree's node template to display the `level` 
+information of a flat tree node.
+
+Nested tree does not need this padding since padding can be easily added to the hierarchy
+structure in DOM.
+
+### Simple Tree
+
+In the near future, we will provide a simplified version of the tree with an easy-to-use
+interface, material styling, and json input.
```
jelbourn: I would leave out this section; I want to avoid talking about future stuff in the docs

tinayuangao:
Updated the document. Please take a look. Thanks!

PR #8468, opened 2017-11-16
fix(chip): fix placeholder and text overlap

### LINKED ISSUE #8448: Chip input placeholder overlaps text
Bug
Chip input placeholder overlaps text when floatPlaceholder="never"

Input doesn't overlap placeholder. Placeholder should disappear if there's any input value

Input overlaps placeholder if there's only one character

https://stackblitz.com/edit/angular-31ymjs?file=app%2Fapp.component.html
Type one character in the input

Material 5.0.0rc0
Angular 5.0.1

tinayuangao:
Fixes #8448

### DIFF
--- a/src/lib/chips/chip-input.ts
+++ b/src/lib/chips/chip-input.ts
@@ -29,6 +29,7 @@ export interface MatChipInputEvent {
     '(keydown)': '_keydown($event)',
     '(blur)': '_blur()',
     '(focus)': '_focus()',
+    '(input)': '_onInput()',
   }
 })
 export class MatChipInput {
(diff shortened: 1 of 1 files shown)


src/lib/chips/chip-input.ts
```diff
@@ -115,5 +116,10 @@ export class MatChipInput {
     }
   }
 
+  _onInput() {
+    // Let chip list know whenever the value changes.
+    this._chipList.stateChanges.next();
```
jelbourn: Should there be a unit test for this?

PR #8480, opened 2017-11-16
fix(form-field): jumping underline in Edge and Firefox

### LINKED ISSUE #8395: Seems the underline is jumping on mat-form-field on focusin and focusout in Edge Browser
Seems the underline is jumping on mat-form-field on focusin and focusout or the thickness is not correct. It looks wierd.

underline1

Look at [Material Demo Site](https://material.angular.io/components/form-field/overview) in Edge Browser

MSEdge: 41.16299.15.0
Material: 5.0.0-rc0-3ca801a

DevVersion:
* In Edge and Firefox the underline jumps due to the `transform: perspective`. Also due to the `transform` the underline is getting slightly taller which isn't aligning with the specs.
* Improves rendering of the ripple underline by always setting the 2px height. Since the element only shows in focus-stated the height can be always `2px`.

**Note**: The `perspective(1px)` might have fixed the zoom-out issue on some browsers, but it 
introduces more issues than it helps (e.g. Firefox 57, Edge). Also the fix is not actually right. It just adds some subpixels to the `1px` underline and therefore makes it easier for the browser to render in some circumstances the zoomed-out element (conflicts with the Material Design specification of 1px).

Fixes #8395

### DIFF
--- a/src/lib/form-field/form-field.scss
+++ b/src/lib/form-field/form-field.scss
@@ -150,8 +150,6 @@ $mat-form-field-default-infix-width: 180px !default;
   position: absolute;
   height: $mat-form-field-underline-height;
   width: 100%;
-  // Prevents underline from disappearing at lower zoom levels.
-  transform: perspective(1px);
 
   .mat-form-field-disabled & {
     background-position: 0;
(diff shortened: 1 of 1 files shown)


mmalerba:
There might be no way to fix it and still make other browsers happy, but here's some examples from Chrome of why I added this in the first place:

Input demo page at 90%:

Input demo page at 90% with perspective transform:
90-perspective
Input demo page 110%:

Input demo page 110% with perspective transform:
110-perspective

DevVersion:
Are those screenshots from the current Chrome version? For me everything looks fine from 33%. As said this seems to depend on the screen resolution and Chrome's pixel rendering.

mmalerba:
Linux, Chrome 62.0.3202.94, 2560x1600

DevVersion:
Hmm. Not sure about that. I don't think there is a reasonable fix for such zooming levels. What do you think? @mmalerba

palpatine1991:
Why not to create a chrome specific SCSS rule using media query?

palpatine1991:
We are facing an issue with  the chrome even with zoom set at 100% but some Windows system are apparently set to zoom "125%" by default and also chrome is affected by this.

PR #8483, opened 2017-11-16
universal: dump out the prerendered file so we can examine it

mmalerba:


src/universal-app/prerender.ts
```diff
@@ -15,7 +15,10 @@ const result = renderModuleFactory(KitchenSinkServerModuleNgFactory, {
 });
 
 result
-  .then(() => log('Prerender done.'))
+  .then(content => {
+    writeFileSync(join(__dirname, 'index-prerendered.html'), content, 'utf-8');
```
DevVersion: Shouldn't we output that file in the `dist/` folder?

jelbourn: Yeah, should be under `dist/`. I thought I added this when I originally wrote this...

crisbeto: Initially it just printed out the final in the console, but that made it hard to see any exceptions.

DevVersion: Yeah that's too much. We should just write it to `dist/index-prerendered.html` and also not under `dist/packages/demo-app` since it's actually the universal app.

_Idea_: If we fully use CircleCI at some point, there is the option to upload the index file as an artifact. Meaning that we can download and inspect the output file easily.

mmalerba: Why not under `dist/packages/universal-app`? That's actually what I meant to do, just copy/paste-failed

DevVersion: Yeah that works too

mmalerba: Apparently I can't `import {buildConfig} from 'material2-build-tools'` in this file and `__dirname` seems to point to the correct directory anyways, since we're running the version of the prerender script from `dist/packages/universal-app`, so... just gonna switch this back to how it was originally

PR #8486, opened 2017-11-16
fix(tabs): detach tab portal when tab hides from view

josephperrott:


src/lib/tabs/tab-body.ts
```diff
@@ -81,19 +81,21 @@ export class MatTabBodyPortal extends _MatTabBodyPortalBaseClass implements OnIn
...
+    });
   }
 
-  /** Clean up subscription if necessary. */
+  /** Clean up centering subscription. */
   ngOnDestroy(): void {
-    if (this._centeringSub && !this._centeringSub.closed) {
-      this._centeringSub.unsubscribe();
-    }
+    this._centeringSub.unsubscribe();
```
jelbourn: It probably still need to check whether `_centeringSub` exists because some people, in their unit tests, destroy a component before calling `detectChanges`

src/lib/tabs/tab-group.spec.ts
```diff
@@ -350,6 +350,19 @@ describe('MatTabGroup', () => {
...
+      expect(fixture.nativeElement.textContent).toContain('Pizza, fries');
+      expect(fixture.nativeElement.textContent).not.toContain('Peanuts');
+
+      tabGroup.selectedIndex = 3;
+      fixture.detectChanges();
+      // Use whenStable to wait for async observables and change detection to run in content.
+      fixture.whenStable().then(() => {
+        expect(fixture.nativeElement.textContent).not.toContain('Pizza, fries');
+        expect(fixture.nativeElement.textContent).toContain('Peanuts');
+      });
```
jelbourn: Expand this to change the tab and re-assert the the others are gone?

jelbourn:
Marking as P1 since this fixes a regression

jrood:
@josephperrott Thanks for this pull request, but shouldn't the tab portal not be detached until the animation is done (after centering rather than before)? The section on ["Navigational transitions" in the Material Design Guidelines](https://material.io/guidelines/patterns/navigational-transitions.html#navigational-transitions-parent-to-child) specifies that in a sibling to sibling transition 

> New content slides in from the right while sibling content moves off-screen to the left.

Since this merge, it seems that the "moves off-screen" part has been effectively removed.

PR #8487, opened 2017-11-16
fix(select):make input lable visible when selecting from mat-select

GoygovRustam:
Make input label visible so users know what they are selecting
Before : 
image

After: 
image


GoygovRustam:
@crisbeto, 
Thank you for the review, you are absolutely right. 
Closing this.

PR #7775, opened 2017-10-13
docs(sidenav): update docs and add examples

mmalerba:

PR #8488, opened 2017-11-16
fix(drawer): allow for drawer container to auto-resize while open

### LINKED ISSUE #6743: [MdSidenav] - Sidenav Resizing Regression

Bug/Regression from 2.0.0-beta.8.

Also a feature request for an API to allow triggering recalculations for the MdSidenavContainer/MdDrawerContainer, or being able to update _styles manually and having the view update..

In the previous release 2.0.0-beta.8 the sidenav resized when the window was resized.

I have a responsive sidenav that uses Angular flex-layout for showing/hiding responsive content in the sidenav when the view changes from md to lg and vice-versa.

In 2.0.0-beta.10 this is no longer the case. The changes made in #6189 made the sidenav update only when the sidenav is opening/closing.

Since flex-layout's changes occur without having to call open() or close() on the sidenav, the width recalculation doesn't occur.

example

Put dynamic/responsive content within a sidenav that is configured with "side".

Being able to dynamically call a refresh of the MdSidenav/MdDrawer or MdSidenavContainer/MdDrawerContainer. 

Even better would be an accessor to allow manually setting the margin of the sidenav to allow for pre-computing the _style variable and updating it when needed.

Angular 4.3.5

This is similar to #6583.

For right now I am using the following hack in my navigation directive to workaround whenever the ObservableMedia from angular flex-layout changes:

```
// Timeout required for flex-layout directives to update the child views/the width.
setTimeout(() => {
            (this._container as any)._updateStyles();
            (this._container as any)._changeDetectorRef.markForCheck();
}, 0);
```

crisbeto:
Adds the `autosize` input that allows users to opt-in to drawers that auto-resize, similarly to the behavior before #6189. The behavior is off by default, because it's unnecessary for most use cases and can cause performance issues.

Fixes #6743.

### DIFF
--- a/src/lib/sidenav/drawer.ts
+++ b/src/lib/sidenav/drawer.ts
@@ -28,6 +28,7 @@ import {
   Output,
   QueryList,
   ViewEncapsulation,
+  InjectionToken,
 } from '@angular/core';
 import {DOCUMENT} from '@angular/common';
 import {merge} from 'rxjs/observable/merge';
@@ -35,6 +36,7 @@ import {filter} from 'rxjs/operators/filter';
 import {take} from 'rxjs/operators/take';
 import {startWith} from 'rxjs/operators/startWith';
 import {takeUntil} from 'rxjs/operators/takeUntil';
+import {debounceTime} from 'rxjs/operators/debounceTime';
 import {map} from 'rxjs/operators/map';
 import {Subject} from 'rxjs/Subject';
 import {Observable} from 'rxjs/Observable';
@@ -54,6 +56,9 @@ export class MatDrawerToggleResult {
   constructor(public type: 'open' | 'close', public animationFinished: boolean) {}
 }
 
+/** Configures whether drawers should use auto sizing by default. */
+export const MAT_DRAWER_DEFAULT_AUTOSIZE =
+    new InjectionToken<boolean>('MAT_DRAWER_DEFAULT_AUTOSIZE');
 
 @Component({
   moduleId: module.id,
@@ -74,7 +79,7 @@ export class MatDrawerContent implements AfterContentInit {
    * drawer is open. We use margin rather than transform even for push mode because transform breaks
    * fixed position elements inside of the transformed element.
    */
-  _margins: {left: number, right: number} = {left: 0, right: 0};
+  _margins: {left: number|null, right: number|null} = {left: null, right: null};
 
   constructor(
       private _changeDetectorRef: ChangeDetectorRef,
@@ -403,7 +408,6 @@ export class MatDrawer implements AfterContentInit, OnDestroy {
 })
 export class MatDrawerContainer implements AfterContentInit, OnDestroy {
   @ContentChildren(MatDrawer) _drawers: QueryList<MatDrawer>;
-
   @ContentChild(MatDrawerContent) _content: MatDrawerContent;
 
   /** The drawer child with the `start` position. */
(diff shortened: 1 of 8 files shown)


src/lib/sidenav/drawer.ts
```diff
@@ -432,16 +449,27 @@ export class MatDrawerContainer implements AfterContentInit, OnDestroy {
   /** Emits when the component is destroyed. */
   private _destroyed = new Subject<void>();
 
+  /** Cached margins used to verify that the values have changed. */
+  private _margins = {left: 0, right: 0};
+
+  /** Used for debouncing reflows. */
+  private _debouncer = new Subject<void>();
```
mmalerba: maybe call it `_doCheckSubject`?

src/lib/sidenav/drawer.ts
```diff
@@ -462,9 +490,15 @@ export class MatDrawerContainer implements AfterContentInit, OnDestroy {
 
       this._changeDetectorRef.markForCheck();
     });
+
+    this._debouncer.pipe(
+      debounceTime(10), // Arbitrary debounce time, less than a frame at 60fps
```
mmalerba: is `debounce` what we want, doesn't `audit` make more sense here?

src/lib/sidenav/drawer.ts
```diff
@@ -432,16 +449,27 @@ export class MatDrawerContainer implements AfterContentInit, OnDestroy {
...
+
+  /** Used for debouncing reflows. */
+  private _debouncer = new Subject<void>();
+
   _contentMargins = new Subject<{left: number, right: number}>();
 
-  constructor(@Optional() private _dir: Directionality, private _element: ElementRef,
-              private _renderer: Renderer2, private _ngZone: NgZone,
+  constructor(@Optional() private _dir: Directionality,
+              @Inject(MAT_DRAWER_DEFAULT_AUTOSIZE) defaultAutosize: boolean,
```
mmalerba: If we move this to the end of the parameter list and make it optional it won't be breaking for users who are extending this class

src/lib/sidenav/sidenav.md
```diff
@@ -82,3 +82,17 @@ is clicked or <kbd>ESCAPE</kbd> is pressed. To add custom logic on backdrop clic
...
+### Resizing an open sidenav
+By default, Material will only measure and resize the drawer container in a few key moments
+(on open, on window resize, on mode change) in order to avoid layout thrashing, however there
+are cases where this can be problematic. If your app requires for a drawer to change its width
+while it is open, you can use the `autosize` option to tell Material to continue measuring it.
+Note that you should use this option **at your own risk**, because it could cause performance
+issues.
+
+```html
+<mat-sidenav-container [autosize]="true">
```
mmalerba: can we make this an embedded example? (https://github.com/angular/material2/pull/7775 will make the rest of these embedded examples as well).

Maybe also add an example to the demo app so we can manually test

src/lib/sidenav/drawer.ts
crisbeto: I'd say so. `audit` just throttles the events, but what we want is to only measure once if multiple events happen very close to each other.

src/lib/sidenav/sidenav.md
crisbeto: I'm not sure what an embedded example would look like in this case, aside from a sidenav that resizes in a `setTimeout`, but that's not very representative of a real use case.

crisbeto:
Addressed the code-related feedback @mmalerba.

src/lib/sidenav/sidenav.md
mmalerba: You could have an icon drawer that expands to show full labels when the user hovers or clicks a button.

Or just have a button that toggles itself between shorter and longer width when clicked, not a practical use case but demonstrates how to use the feature and lets us manually test it

src/lib/sidenav/drawer.ts
mmalerba: If the drawer were animating open, wouldn't it just keep denouncing until the animation is over and then jump?

crisbeto: AFAIK animations don't trigger change detection. `ngDoCheck` runs for every change detection cycle.

src/lib/sidenav/sidenav.md
crisbeto: Alright, I've added a live example.

src/material-examples/sidenav-autosize/sidenav-autosize-example.html
```diff
@@ -0,0 +1,16 @@
+<mat-sidenav-container class="example-container" autosize>
```
mmalerba: Can you change these to the drawer variants? I only want people to use sidenav for fullscreen layouts

tinayuangao:

crisbeto:
Rebased @tinayuangao.

PR #8490, opened 2017-11-16
refactor(input): switch to fakeAsync tests

crisbeto:
Moves all of the input tests away from the `async` zone either to the default one or to `fakeAsync`.

### DIFF
--- a/src/lib/input/input.spec.ts
+++ b/src/lib/input/input.spec.ts
@@ -1,7 +1,7 @@
 import {Platform, PlatformModule} from '@angular/cdk/platform';
 import {createFakeEvent, dispatchFakeEvent, wrappedErrorMessage} from '@angular/cdk/testing';
 import {ChangeDetectionStrategy, Component, ViewChild} from '@angular/core';
-import {async, ComponentFixture, inject, TestBed} from '@angular/core/testing';
+import {ComponentFixture, inject, TestBed, fakeAsync, flush} from '@angular/core/testing';
 import {
   FormControl,
   FormGroup,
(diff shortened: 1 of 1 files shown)


src/lib/input/input.spec.ts
```diff
@@ -160,7 +160,7 @@ describe('MatInput without forms', function () {
     expect(el.classList.contains('mat-form-field-empty')).toBe(true);
   });
 
-  it('should not be empty after input entered', async(() => {
+  it('should not be empty after input entered', () => {
```
mmalerba: Should we make these `fakeAsync` as well? if they do async things in their constructor or lifecycle hooks they could silently leak async operations. I like that `fakeAsync` fails if you don't flush it - no more time wasted binary searching for the leaky test

crisbeto: I think they were async to begin with because we just had one async and then we copy-pasted it when we wrote new tests, but they don't really need to be. As for those silent error leaks, I believe they only happen with the `async` zone which we won't use anymore.

mmalerba: I see, so what happens if we don't mark something `fakeAsync` but it actually does do async stuff? Even if they're not currently async, I'm worried about introducing code that causes them to become async in the future, forgetting to update the tests, and then later running into some obscure failure related to it

crisbeto: I'm not a fan of having to add `fakeAsync` even though we might not need it. Presumably if we add some that is async in the future, we'd know what PR introduced it and we'd fix any failures before they make it into master.

mmalerba: That's not really the way its worked out in the past. There have been a number of times where our CI was red due to async test failures and nobody really knew how to track down. If we preemptively make everything `fakeAsync` then its really obvious which test has the issue.

crisbeto:
I've switched everything over to `fakeAsync` @mmalerba.

PR #8491, opened 2017-11-16
fix(input): Add pure-CSS floating label logic that will work on...

### LINKED ISSUE #8247: Provide a better input experience with Server Side Rendering
proposal

the input placeholder texts renders at the top of the input when the app is being rendered on the server side to avoid interference with the user input 

the placeholder text and user input overlap when only SSR is loaded

image

server-side rendering with angular universal
add mat-input
throttle the network to simulate a slow 3g connection user experience
type in the SSR version input before the angular app bootstraps

it is a very frustrating to the user to type without being able to clearly see what he typed

Angular 5, all browsers!

mmalerba:
...server-side-rendered pages.

fixes #8247

### DIFF
--- a/src/lib/form-field/_form-field-theme.scss
+++ b/src/lib/form-field/_form-field-theme.scss
@@ -90,6 +90,12 @@
   }
 }
 
+// Used to make instances of the _mat-form-field-placeholder-floating mixin negligibly different,
+// and prevent Google's CSS Optimizer from collapsing the declarations. This is needed because some
+// of the selectors contain pseudo-classes not recognized in all browsers. If a browser encounters
+// an unknown pseudo-class it will discard the entire rule set.
+$dedupe: 0;
+
 // Applies a floating placeholder above the form field control itself.
 @mixin _mat-form-field-placeholder-floating($font-scale, $infix-padding, $infix-margin-top) {
   // We use perspective to fix the text blurriness as described here:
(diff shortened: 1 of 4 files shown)


src/lib/form-field/_form-field-theme.scss
```diff
@@ -198,6 +206,21 @@
       @include _mat-form-field-placeholder-float-nodedupe(
               $subscript-font-scale, $infix-padding, $infix-margin-top);
     }
+
+    // Server-side rendered matInput with focus (used as a pure CSS stand-in for
+    // mat-form-field-should-float).
+    .mat-input-server:focus + .mat-form-field-placeholder-wrapper .mat-form-field-placeholder {
+      @include _mat-form-field-placeholder-floating(
+              $subscript-font-scale, $infix-padding, $infix-margin-top);
+    }
```
jelbourn: Can't this one be combined with the other one above?
(`&.mat-form-field-should-float .mat-form-field-placeholder`)

src/lib/form-field/_form-field-theme.scss
```diff
@@ -113,11 +113,19 @@
...
 @mixin _mat-form-field-placeholder-float-nodedupe($font-scale, $infix-padding, $infix-margin-top) {
   transform: translateY(-$infix-margin-top - $infix-padding) scale($font-scale) perspective(100px)
-             translateZ(0.002px);
-  -ms-transform: translateY(-$infix-margin-top - $infix-padding) scale($font-scale);
+             translateZ(0.001px) scale(1);
+  -ms-transform: translateY(-$infix-margin-top - $infix-padding) scale($font-scale) scale(1);
   width: (100% / $font-scale) + 0.0001;
 }
 
+@mixin _mat-form-field-placeholder-float-nodedupe2($font-scale, $infix-padding, $infix-margin-top) {
```
jelbourn: Instead of adding another mixin, I realized we could programatically change the rule each time it's stamped out, e.g.:
```scss
$uniqueId: 0;

@mixin cool-style() {
  content: 'Hello' + $uniqueId;
  $uniqueId: $uniqueId + 1 !global;
}


.one {
  @include cool-style(); // content: "Hello0";
}

.two {
  @include cool-style(); // content: "Hello1";
}
```

src/lib/input/input.ts
```diff
@@ -162,7 +168,8 @@ export class MatInput implements MatFormFieldControl<any>, OnChanges, OnDestroy,
               @Optional() protected _parentForm: NgForm,
               @Optional() protected _parentFormGroup: FormGroupDirective,
               private _defaultErrorStateMatcher: ErrorStateMatcher,
-              @Optional() @Self() @Inject(MAT_INPUT_VALUE_ACCESSOR) inputValueAccessor: any) {
+              @Optional() @Self() @Inject(MAT_INPUT_VALUE_ACCESSOR) inputValueAccessor: any,
+              @Optional() @Inject(PLATFORM_ID) platformId?: Object) {
```
jelbourn: This class already injects `Platform` from cdk/platform, you could just use `!this._platform.isBrowser`

mmalerba: `Platform` just checks if the `document` exists and has type `object`. This probably won't work now that universal gives us a domino document right?

We probably need to update the `Platform.isBrowser` logic, but I didn't want to do it in this PR.

jelbourn: I'd be hesitant to change it now because it could be a breaking change

mmalerba: That's why I left it out of this PR, but I do think we have to change it in the future. We need to audit the places where we're using this and check which ones are using it as a way to check if the `document` is defined and which ones really care that we're on the server. The latter checks are now incorrect due to Angular's changes. (The one I added in this PR for example, would be broken if I depended on `Platform.isBrowser`)

mmalerba:
comments addressed

PR #8492, opened 2017-11-16
fix(icon): server-side error when registering icons

### LINKED ISSUE #6787: svg icon registration doesn't work on server platform
Bug: svg icons doesn't work on server.

Work on server also.

icon registration is [tightly coupled with DOM and depends on `document` to be available](https://github.com/angular/material2/blob/f375f92581b6318219cc225aeafc97124a840b26/src/lib/icon/icon-registry.ts#L400) which breaks its compatibility with server platform.

Use svg icons on server platform.

It feels quite normal to have svg icons work on server.

@angular/material@beta.8

Nope.

crisbeto:
Fixes an error that is thrown by the `IconRegistry` when registering icons on the server side.

Fixes #6787.

### DIFF
--- a/src/lib/icon/icon-registry.ts
+++ b/src/lib/icon/icon-registry.ts
@@ -11,13 +11,21 @@ import {tap} from 'rxjs/operators/tap';
 import {finalize} from 'rxjs/operators/finalize';
 import {map} from 'rxjs/operators/map';
 import {share} from 'rxjs/operators/share';
-import {Injectable, Optional, SecurityContext, SkipSelf} from '@angular/core';
+import {
+  Injectable,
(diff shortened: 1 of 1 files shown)


freezy:
Any ETA when this will make it into a release? This is blocking our migration to Universal, so it would be good to know if I just wait another day or look for a different solution.

andrewseguin:
This is a major change due to the constructor being altered, seeing breaking tests internally due to subclasses breaking

jelbourn:
Constructor change could be avoided by not using the injected document

crisbeto:
Reworked it to avoid the breaking change. Switching back to patch release.

PR #8519, opened 2017-11-17
refactor(focus-trap): better handling of server-side rendering

crisbeto:
Now that Angular is using Domino, we don't have to completely shut off the focus trap when rendering on the server. These changes remove all the old checks and use the correct document injection token.

### DIFF
--- a/src/cdk/a11y/focus-trap.ts
+++ b/src/cdk/a11y/focus-trap.ts
@@ -14,11 +14,12 @@ import {
   OnDestroy,
   AfterContentInit,
   Injectable,
+  Inject,
 } from '@angular/core';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
-import {Platform} from '@angular/cdk/platform';
 import {first} from 'rxjs/operators/first';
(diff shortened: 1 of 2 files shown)


src/cdk/a11y/focus-trap.ts
```diff
@@ -45,9 +46,9 @@ export class FocusTrap {
 
   constructor(
     private _element: HTMLElement,
-    private _platform: Platform,
     private _checker: InteractivityChecker,
     private _ngZone: NgZone,
+    private _document: any,
```
jelbourn: I'm not really happy with the propagation of `any` from this pattern. I just tried this locally and it works:

```ts
private _document: Document;

class Whatever {
  constructor(@Optional() @Inject(DOCUMENT) doc: any) {
    this._document = doc;
  }
}
```

It's a little more boilerplate but I think it's worth it. Start doing this instead?

crisbeto: Makes sense. I'll switch these changes over.

PR #8534, opened 2017-11-18
fix(tooltip): unable to type in input with tooltip on iOS

### LINKED ISSUE #8331: [Tooltip] Hangs matInput on iPhone
Bug

Normal editing.

Only allow one character input.

Open this on an iPhone: https://tinyurl.com/yd8k7pyt
Full plunker here: http://plnkr.co/edit/ySZamMTpNdhqmg59iPOO?p=preview

n/a

See plunker. Only iPhone has the issue for what I know. In both Chrome and Safari.

crisbeto:
Fixes an issue, caused by HammerJS adding `user-select: none` to the trigger element, that prevents users from being able to type inside inputs on iOS, if they have a tooltip.

Fixes #8331.


andrewseguin:
Needs rebase

PR #8542, opened 2017-11-19
feat(datepicker): add animation to calendar popup

crisbeto:
Adds an animation when opening and closing the datepicker's calendar.

For reference:
animation

### DIFF
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -16,6 +16,7 @@ import {
   PositionStrategy,
   RepositionScrollStrategy,
   ScrollStrategy,
+  ConnectedPositionStrategy,
 } from '@angular/cdk/overlay';
 import {ComponentPortal} from '@angular/cdk/portal';
 import {first} from 'rxjs/operators/first';
@@ -35,6 +36,8 @@ import {
   ViewChild,
   ViewContainerRef,
   ViewEncapsulation,
+  ChangeDetectorRef,
+  OnInit,
 } from '@angular/core';
 import {DateAdapter} from '@angular/material/core';
 import {MatDialog, MatDialogRef} from '@angular/material/dialog';
@@ -44,6 +47,7 @@ import {Subscription} from 'rxjs/Subscription';
 import {MatCalendar} from './calendar';
 import {createMissingDateImplError} from './datepicker-errors';
 import {MatDatepickerInput} from './datepicker-input';
+import{trigger, state, style, animate, transition} from '@angular/animations';
 
 
 /** Used to generate a unique ID for each datepicker instance. */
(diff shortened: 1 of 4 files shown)


TheRedD0t:
It would be very cool if the sheet would do also a opening along the x-axis like [here](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B14F_FSUCc01RkRSNjBMbGU2UHM/responsive-03-surfaceconnection-v3.webm).

crisbeto:
@TheRedD0t that's what I was going for initially (the menu has a similar animation), but it looks weird when the animation doesn't originate from the place where the user clicked.

TheRedD0t:
@crisbeto Yeah I just realized that this would be a "Don't" according to the guidelines.

https://material.io/guidelines/motion/choreography.html#choreography-creation

src/lib/datepicker/datepicker.ts
```diff
@@ -81,23 +85,73 @@ export const MAT_DATEPICKER_SCROLL_STRATEGY_PROVIDER = {
...
+      }
+    });
+  }
+
   ngAfterContentInit() {
     this._calendar._focusActiveCell();
   }
 
+  ngOnDestroy() {
+    if (this._positionChange) {
```
mmalerba: should we use the usual `_destoryed.next()`?

mmalerba:
Just realized, since this adds an Angular animation to a component that previously didn't have one, it may be a breaking change.

@jelbourn should this be part of a patch, minor, or major?

src/lib/datepicker/datepicker.ts
crisbeto: I think for this case it's better to got with the subscription, because it allows us to know whether we've subscribed before. Otherwise since we don't dispose of the dialog content until the datepicker is destroyed, we could end up in a situation where we've subscribed multiple times to the same stream.

jelbourn:
I chatted w/ @StephenFluin and he agreed that adding an animation to a component that didn't have one before would be considered a breaking change since users would have to add `BrowserAnimationsModule` if they didn't have it (however unlikely it is that someone is using datepicker and not some other component that has animations already). I think this one may have to wait for 6.0

mmalerba:
@jelbourn for 6.0 we might want to consider adding a no-op animation to anything that's supposed to have animations but we haven't got to it yet. That way we can do it at some point without it being breaking

jelbourn:
Unless something changes in Angular, we won't be able to do that; both animations modules import `BrowserModule`, which Angular only allows to be imported once.

mmalerba:
@jelbourn By no-op animation I don't mean the `NoopAnimationModule`, just an animation that does nothing, e.g. `state('noop', style({}))`. This way it would have the `BrowserAnimationsModule` dependency and we could later make the animation do useful things without it being a breaking change

PR #8543, opened 2017-11-19
fix(form-field): unable to tap on certain types of inputs on iOS

### LINKED ISSUE #8001: Input is not working for type date in iOS
Textbox is not working in iOS. I'm new to material design i have an application page in my page in that there is a field for date of birth there i have specified type = date in input but when i hosted the website only that textbox is not working in iOS.

I did all the inputs  without specifying its type but for date of birth i needed to specify type as date and needed the material styles. but date type is not working in iOS

for input with date type is not working in iOS. Its not even focusing in any of the browsers

In this link you can see  my problem http://kitex.s3-website-us-west-2.amazonaws.com.
For android its working perfectly but in iOS its not even focusing.

`<mat-form-field class="fields-full-width">
      <input type="date" matInput placeholder="Date of birth" #DOB [max]="today" (keyup)="dobValidation(DOB)" (input)="dobValidation(DOB)">
 </mat-form-field>`

This is my code for date of birth

There is a field for date of birth. Its not even focusing on that field. its like completly disabled. 

I'm using Angular 2 (@angular/cli: 1.1.1), Material(2.0.0-beta.12), in windows 10

for testing i have used iOS 11.0.3. In that safari and crome are the brosers i have used . But in both
date didn't worked 

[employee-registration.zip](https://github.com/angular/material2/files/1411975/employee-registration.zip)

material date picker in material.angular.io site doesn't have clarity of how to use it.

crisbeto:
Fixes an issue that caused certain input types to collapse into 1px tall strips, preventing users from being to tap on them on iOS.

Fixes #8001.

### DIFF
--- a/src/lib/input/input.scss
+++ b/src/lib/input/input.scss
@@ -37,6 +37,22 @@
     display: none;
   }
 
+  // Fixes an issue on iOS where the following input types will collapse to 1px,
+  // if they're empty, because we've overridden their background color.
+  // See: https://stackoverflow.com/questions/18381594/input-type-date-appearance-in-safari-on-ios
+  &[type='date'],
+  &[type='datetime'],
+  &[type='datetime-local'],
+  &[type='month'],
+  &[type='week'],
+  &[type='time'] {
+    &::after {
+      content: ' ';
+      white-space: pre;
+      width: 1px;
+    }
+  }
+
   @include input-placeholder {
     // Delay the transition until the label has animated about a third of the way through, in
     // order to prevent the placeholder from overlapping for a split second.


src/lib/input/input.scss
```diff
@@ -37,6 +37,22 @@
...
+  &[type='date'],
+  &[type='datetime'],
+  &[type='datetime-local'],
+  &[type='month'],
+  &[type='week'],
+  &[type='time'] {
+    &::after {
+      content: ' ';
+      white-space: pre;
+      width: 1px;
```
mmalerba: did you try adding something like `margin-left: -1px;` just to make extra sure it doesn't mess with the layout, also maybe `visibility: hidden;`

crisbeto: I wanted to keep it as minimal as possible. If we added the margin, we'd also have to flip it in RTL. I don't think it's an issue in this case, because users can't type in these input types anyway.

tinayuangao:

crisbeto:
Rebased @tinayuangao.

Issue #8279, opened 2017-11-07
[cdk] indicate in docs which utilities operate outside NgZone


willshowell:
#### Bug, feature request, or proposal:
Docs request

#### Current behavior
It takes a little extra digging in the source to understand `ScrollDispatcher#scrolled` and `ViewportRuler#change` emit outside the zone.

#### Expected behavior
A note in the jsdoc of these properties and methods.

josephperrott:
This definitely seems like something that could be good to have in docs.  Though it will definitely take some designing as to how to surface it, as well as seeing if we could use something like linting to be sure that we always note `ngZone` usage.  Worse than not documenting it, imo would be to have it inconsistently documented.

willshowell:
Agreed. Not having to hunt for the information would be good. Perhaps a prominent accent label.

A complication of a linter solution is that some utilities don't explicitly run outside of the zone, but can be called by another that does. For example, `onPositionChange`

https://github.com/angular/material2/blob/c13848e715c8940de3fb923e0a69ba42e67bbcb3/src/cdk/overlay/position/connected-position-strategy.ts#L74-L77

The subject is usually advanced by [`apply()`](https://github.com/angular/material2/blob/c13848e715c8940de3fb923e0a69ba42e67bbcb3/src/cdk/overlay/position/connected-position-strategy.ts#L111-L157), which is called by [`updatePosition()`](https://github.com/angular/material2/blob/c13848e715c8940de3fb923e0a69ba42e67bbcb3/src/cdk/overlay/overlay-ref.ts#L198-L203) which is sometimes called within but usually [outside](https://github.com/angular/material2/blob/c13848e715c8940de3fb923e0a69ba42e67bbcb3/src/cdk/overlay/scroll/reposition-scroll-strategy.ts#L46-L48) the [zone](https://github.com/angular/material2/blob/c13848e715c8940de3fb923e0a69ba42e67bbcb3/src/cdk/scrolling/scroll-dispatcher.ts#L144-L148).

saulotoledo:
@jelbourn I think one could add the "has pr" label to this issue.

jelbourn:
The PR here was merged; it added a note to the API docs for `ScrollDispatcher`, but not for `ViewportRuler`

PR #8545, opened 2017-11-19
docs(scroll-dispatcher): add note about callback being run outside the NgZone

crisbeto:
Since the `scrolled` events are being run outside the NgZone, it might be confusing if people decide to update their view as a result of a scroll event. These changes add a note on the method docs to avoid some of the confusion.

Relates to #8279.

### DIFF
--- a/src/cdk/scrolling/scroll-dispatcher.ts
+++ b/src/cdk/scrolling/scroll-dispatcher.ts
@@ -73,6 +73,11 @@ export class ScrollDispatcher {
    * Returns an observable that emits an event whenever any of the registered Scrollable
    * references (or window, document, or body) fire a scrolled event. Can provide a time in ms
    * to override the default "throttle" time.
+   *
+   * **Note:** in order to avoid hitting change detection for every scroll event,
+   * all of the events emitted from this stream will be run outside the Angular zone.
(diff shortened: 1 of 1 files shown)


jelbourn:
The PR here was merged; it added a note to the API docs for `ScrollDispatcher`, but not for `ViewportRuler`

jelbourn:
Whoops, commented on the wrong thing

philip-firstorder:
This saved me, thank you very much for commenting this!

For anyone out there, if you use the ScrollDispatcher to add rows to the infinite Mat-Table then you have to do like this:
   ```typescript
this.scrollDispatcher.scrolled()  
    .pipe(takeUntil(this.ngDestroy$))
    .subscribe((scrollable: CdkScrollable) => {
      const bottom = scrollable ? scrollable.measureScrollOffset('bottom') : null;
      if (bottom <= 0) { // If the user has scrolled to the bottom
        this.ngZone.run(() => {
          this.nextPage();; // Run this in `NgZone.run`, or your table will not renderRows
        });
      }
});
this.nextPage()  {
    ...
    this.dataSource = data.length ? data : undefined;
    this.table.renderRows(); // This renders rows only if you used `NgZone.run` above 
}
```

PR #8546, opened 2017-11-19
chore(*): convert demos to stackblitz

amcdnl:
This PR moves demos from plunker to stackblitz.


jelbourn:
FYI, you can just do `chore: xxx www zzz`, the parens aren't necessary when it would be empty

PR #8548, opened 2017-11-19
refactor: remove rxjs patch operators in favor of lettable operators

### LINKED ISSUE #8501: Change rxjs imports in material-examples
Proposal

I'd expect all examples to use lettable operators.

Currently most (maybe all) examples are importing rxjs using `patch` operators, like this:

`import 'rxjs/add/operator/*';`

Not sure if you want to change this, but if you do, I can help with this.

CaerusKaru:
* Fixes #8501

### DIFF
--- a/src/demo-app/a11y/table/table-a11y.ts
+++ b/src/demo-app/a11y/table/table-a11y.ts
@@ -11,6 +11,8 @@ import {DataSource} from '@angular/cdk/table';
 import {BehaviorSubject} from 'rxjs/BehaviorSubject';
 import {Observable} from 'rxjs/Observable';
 import {MatSort, MatPaginator} from '@angular/material';
+import {merge} from 'rxjs/observable/merge';
+import {map} from 'rxjs/operators/map';
 
 export interface UserData {
   name: string;
@@ -78,9 +80,7 @@ export class SortDataSource extends DataSource<UserData> {
       this._sort.sortChange,
     ];
 
-    return Observable.merge(...displayDataChanges).map(() => {
-      return this.getSortedData();
-    });
+    return merge(...displayDataChanges).pipe(map(() => this.getSortedData()));
   }
 
   disconnect() {}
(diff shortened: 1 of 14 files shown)


src/demo-app/a11y/table/table-a11y.ts
```diff
@@ -78,9 +80,12 @@ export class SortDataSource extends DataSource<UserData> {
       this._sort.sortChange,
     ];
 
-    return Observable.merge(...displayDataChanges).map(() => {
-      return this.getSortedData();
-    });
+    return merge(...displayDataChanges)
+      .pipe(
+        map(() => {
+          return this.getSortedData();
```
rafaelss95: This can be turned to one-line.

src/material-examples/autocomplete-display/autocomplete-display-example.ts
```diff
@@ -30,9 +30,11 @@ export class AutocompleteDisplayExample {
 
   ngOnInit() {
     this.filteredOptions = this.myControl.valueChanges
-        .startWith(null)
-        .map(user => user && typeof user === 'object' ? user.name : user)
-        .map(name => name ? this.filter(name) : this.options.slice());
+      .pipe(
+        startWith({} as User),
```
rafaelss95: Is there any reason to change from `null` to `{} as User`?

CaerusKaru: Having it stay as `null` breaks `strictNullChecks`.

src/demo-app/a11y/table/table-a11y.ts
CaerusKaru: Fixed

src/universal-app/kitchen-sink/kitchen-sink.ts
```diff
@@ -55,7 +54,7 @@ export class KitchenSink {
 
   /** Data source for the CDK and Material table. */
   tableDataSource: DataSource<any> = {
-    connect: () => Observable.of([
+    connect: () => of([
```
rafaelss95: Before, when Material2 imports the rxjs "_functions_" directly they used to rename the `of` import like this:

`import {of as observableOf} from 'rxjs/observable/of';`

Not sure if it's correct to use just `of...`.

CaerusKaru: Probably because `of` is a keyword in the ES6 `for...of` syntax. I'll change it just to be safe.

rafaelss95: Yes. In fact, I thought `rxjs` could change this operator name, as they did for (to avoid conflicts):

`catch` -> `catchError`
`do` -> `tap`
`finally` -> `finalize`
`switch` -> `switchAll`

CaerusKaru: It might be worth filing an issue in the [rxjs repo](https://github.com/ReactiveX/rxjs)

jelbourn:
@andrewseguin please review

PR #8565, opened 2017-11-20
feat(datepicker): add year selection mode

### LINKED ISSUE #5845: Make year selectable in date picker
feature request

The year should be selectable from a dropdown list.

The year can only be changed with the previous and next button.

See demo: https://material.angular.io/components/datepicker/examples

Entering birthdays is very cumbersome. You have to click previous a lot for old people :)

Material 2.0

mmalerba:
fixes #5845

### DIFF
--- a/src/lib/datepicker/calendar.spec.ts
+++ b/src/lib/datepicker/calendar.spec.ts
@@ -32,7 +32,7 @@ import {MatCalendar} from './calendar';
 import {MatCalendarBody} from './calendar-body';
 import {MatDatepickerIntl} from './datepicker-intl';
 import {MatMonthView} from './month-view';
-import {MatMultiYearView} from './multi-year-view';
+import {MatMultiYearView, yearsPerPage, yearsPerRow} from './multi-year-view';
 import {MatYearView} from './year-view';
 
 
@@ -135,6 +135,24 @@ describe('MatCalendar', () => {
       fixture.detectChanges();
 
       expect(calendarInstance._activeDate).toEqual(new Date(2018, JAN, 31));
+
+      prevButton.click();
+      fixture.detectChanges();
+
+      expect(calendarInstance._activeDate).toEqual(new Date(2017, JAN, 31));
+    });
+
+    it('should go to previous and next multi-year range', () => {
+      periodButton.click();
+      fixture.detectChanges();
+
+      expect(calendarInstance._currentView).toBe('multi-year');
+      expect(calendarInstance._activeDate).toEqual(new Date(2017, JAN, 31));
+
+      nextButton.click();
+      fixture.detectChanges();
+
+      expect(calendarInstance._activeDate).toEqual(new Date(2017 + yearsPerPage, JAN, 31));
 
       prevButton.click();
       fixture.detectChanges();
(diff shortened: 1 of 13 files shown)


src/lib/datepicker/calendar.ts
```diff
@@ -133,28 +137,46 @@ export class MatCalendar<D> implements AfterContentInit, OnDestroy, OnChanges {
...
+    }
+    if (this._currentView == 'year') {
+      return this._dateAdapter.getYearName(this._activeDate);
+    }
+    let curYear = this._dateAdapter.getYear(this._activeDate);
+    let firstYear = this._dateAdapter.getYearName(
+        this._dateAdapter.createDate(curYear - curYear % 24, 0, 1));
+    let lastYear = this._dateAdapter.getYearName(
+        this._dateAdapter.createDate(curYear + yearsPerPage - 1 - curYear % 24, 0, 1));
+    return `${firstYear} \u2013 ${lastYear}`;
```
jelbourn: Would this need to be internationalized?

src/lib/datepicker/calendar.ts
```diff
@@ -133,28 +137,46 @@ export class MatCalendar<D> implements AfterContentInit, OnDestroy, OnChanges {
...
-            .toLocaleUpperCase() :
-        this._dateAdapter.getYearName(this._activeDate);
+    if (this._currentView == 'month') {
+      return this._dateAdapter.format(this._activeDate, this._dateFormats.display.monthYearLabel)
+          .toLocaleUpperCase();
+    }
+    if (this._currentView == 'year') {
+      return this._dateAdapter.getYearName(this._activeDate);
+    }
+    let curYear = this._dateAdapter.getYear(this._activeDate);
```
jelbourn: `const`?

src/lib/datepicker/calendar.ts
```diff
@@ -133,28 +137,46 @@ export class MatCalendar<D> implements AfterContentInit, OnDestroy, OnChanges {
...
+    let firstYear = this._dateAdapter.getYearName(
+        this._dateAdapter.createDate(curYear - curYear % 24, 0, 1));
+    let lastYear = this._dateAdapter.getYearName(
```
jelbourn: How about `selectedYear`, `firstYearInView` and `lastYearInView`?

src/lib/datepicker/calendar.ts
```diff
@@ -208,29 +230,37 @@ export class MatCalendar<D> implements AfterContentInit, OnDestroy, OnChanges {
     this._userSelection.emit();
   }
 
+  /** Handles month selection in the multi-year view. */
+  _yearSelected(year: D): void {
```
jelbourn: `_selectYear`?

src/lib/datepicker/calendar.ts
```diff
@@ -208,29 +230,37 @@ export class MatCalendar<D> implements AfterContentInit, OnDestroy, OnChanges {
...
   /** Handles user clicks on the period label. */
   _currentPeriodClicked(): void {
-    this._monthView = !this._monthView;
+    this._currentView = this._currentView == 'month' ? 'multi-year' : 'month';
   }
 
   /** Handles user clicks on the previous button. */
   _previousClicked(): void {
-    this._activeDate = this._monthView ?
+    this._activeDate = this._currentView == 'month' ?
```
jelbourn: Make a `_isMonthView()` function instead of repeating the comparison?

src/lib/datepicker/calendar.ts
```diff
@@ -269,10 +301,15 @@ export class MatCalendar<D> implements AfterContentInit, OnDestroy, OnChanges {
...
-        this._dateAdapter.getMonth(date1) == this._dateAdapter.getMonth(date2) :
-        this._dateAdapter.getYear(date1) == this._dateAdapter.getYear(date2);
+    if (this._currentView == 'month') {
+      return this._dateAdapter.getYear(date1) == this._dateAdapter.getYear(date2) &&
+          this._dateAdapter.getMonth(date1) == this._dateAdapter.getMonth(date2)
+    }
+    if (this._currentView == 'year') {
+      return this._dateAdapter.getYear(date1) == this._dateAdapter.getYear(date2);
+    }
+    return Math.floor(this._dateAdapter.getYear(date1) / yearsPerPage) ==
```
jelbourn: Add comment like?
```ts
// Multi-year view
```

src/lib/datepicker/multi-year-view.html
```diff
@@ -0,0 +1,16 @@
+<table class="mat-calendar-table">
+  <thead class="mat-calendar-table-header">
+    <tr><th class="mat-calendar-table-header-divider" colspan="4"></th></tr>
+  </thead>
+  <tbody mat-calendar-body
+         role="grid"
```
jelbourn: I think this also needs `aria-readonly="true"` since `role="grid"` treats it as editable by default

Misiu:
@mmalerba this looks fine.
I'd like to ask for single fix.
When You specify min and max date in calendar view all invalid dates are gray'ed out:
snap 2017-12-05 at 09 35 13

but when You switch to Year view this isn't happening:
snap 2017-12-05 at 09 35 27

Maybe same rule could be applied? Even better would be to hide years that aren't available to pick from (if I have min and max in same Year switching to Year view should be disabled)

src/lib/datepicker/calendar.ts
mmalerba: It probably should be. I don't think any of our `*Intl` objects currently have placeholders, so we'd have to make up some convention for that.

Should we switch to using the official Angular i18n stuff? It seems weird that we're just rolling our own.

src/lib/datepicker/calendar.ts
mmalerba: I combined this with `_monthSelected` and named the new method `_goToDateInView`

src/lib/datepicker/calendar.ts
mmalerba: Do we gain anything from the method? seems equally readable to what I'm doing here

src/lib/datepicker/multi-year-view.html
mmalerba: ok, moved both of these to `mat-calendar-body` since they're the same for all of these views

mmalerba:
@Misiu yeah I should do that, I want to make a separate PR first that changes the way some of the logic works behind the scenes for that feature, then I'll add it here

Misiu:
@mmalerba I've noticed that with latest changes I can't navigate to year list in demo in first comment (https://mmalerba-demo1.firebaseapp.com/datepicker) is this intended?

mmalerba:
oh I reused that demo for something else, there's currently no demo for this PR

Misiu:
@mmalerba no worries :) I'll wait till it get merged.

src/lib/datepicker/calendar.spec.ts
```diff
@@ -311,7 +343,12 @@ describe('MatCalendar', () => {
             dispatchMouseEvent(periodButton, 'click');
             fixture.detectChanges();
 
-            expect(calendarInstance._monthView).toBe(false);
+            expect(calendarInstance._currentView).toBe('multi-year');
+
+            (<HTMLElement>calendarBodyEl.querySelector('.mat-calendar-body-active')).click();
```
jelbourn: Use `as HTMLElement` (here and elsewhere)

Toub:
Good job guys!

What is missing in this PR in order to be merged?

FergusZhou:
@Misiu it seems that what you mentioned _(if I have min and max in same Year switching to Year view should be disabled)_ is not implemented in current version. Do you have any idea about this? Sorry that can't find help with it..

Misiu:
@FergusZhou I'm not using Angular currently. Maybe @mmalerba can help You with this.

PR #8580, opened 2017-11-21
feat(radio): add tabIndex support to mat-radio-group

### LINKED ISSUE #8553: tabindex not supported on md-radio-group
tabindex doesn´t work

Set focus on input and after pressed TAB, could be get focus md-radio-group with lower value of tabindex. In order second and first radio group.

tabindex is ignored

https://plnkr.co/edit/xZNIXCXoRc8GFxUsc8qd?p=preview

I want set custom order focus invoked by pressed tab

Tested on Angular 4.4.4, Angular/Material 2.0.0-beta.11, Chrome

CaerusKaru:
* Fixes #8553

### DIFF
--- a/src/lib/radio/radio.html
+++ b/src/lib/radio/radio.html
@@ -22,6 +22,7 @@
          [required]="required"
          [attr.aria-label]="ariaLabel"
          [attr.aria-labelledby]="ariaLabelledby"
+         [attr.tabindex]="parentTabIndex"
          (change)="_onInputChange($event)"
          (click)="_onInputClick($event)">
 
--- a/src/lib/radio/radio.scss
+++ b/src/lib/radio/radio.scss
@@ -5,6 +5,13 @@
 $mat-radio-size: $mat-toggle-size !default;
 $mat-radio-ripple-radius: 25px;
 
+.mat-radio-group {
+  &:focus,
+  &:hover {
+    outline: none;
+  }
+}
+
 // Top-level host container.
 .mat-radio-button {
   display: inline-block;
--- a/src/lib/radio/radio.spec.ts
+++ b/src/lib/radio/radio.spec.ts
@@ -16,6 +16,8 @@ describe('MatRadio', () => {
         FocusableRadioButton,
         RadiosInsideRadioGroup,
         RadioGroupWithNgModel,
+        RadioGroupWithNativeTabIndexAttr,
+        RadioGroupWithTabIndexBinding,
         RadioGroupWithFormControl,
         StandaloneRadioButtons,
         InterleavedRadioGroup,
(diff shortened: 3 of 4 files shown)


CaerusKaru:
> the mixinTabIndex has been added, but it's not used at all

What do you mean by this? Isn't adding the mixin enabling tabIndex support to the host class?

> Nor is the tabIndex input added to the component metadata.

Good catch, added

> The radio-group should not be focusable at all

In the original issue, @jelbourn listed this as a feature he wanted implemented, but I'd be happy to close this if that's no longer the case.

> This means that the tabIndex needs to be transferred to the radio buttons

Based on the [tabindex spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex), doesn't the child content have its own tabIndex independent of the parent?

DevVersion:
I'm not explicitly saying that #8553 is an invalid feature request. What I mean is that the `mat-radio-group` element is not intended to be a tabbable element. 

Let me demonstrate this by a small scenario. Let's say we have a radio-group with three options. If the user tries to tab to the radio group, the *checked* option will receive focus only. Switching through the other available options can be done through the arrow keys.

This basically means that only the selected `<mat-radio-button>` element is tabbable. It wouldn't make any sense to allow the user to select both, the `<mat-radio-group>` and the selected `<mat-radio-button>`.

```html
<!--
   This tabIndex doesn't have any effect. Just stores the value and 
   provides it for the mat-radio-button inputs. 
-->
<mat-radio-group [tabIndex]="2">
  <mat-radio-button>
     <input [tabIndex]="parentTabIndex || 0">
  </mat-radio-button>
  <mat-radio-button>
     <input [tabIndex]="parentTabIndex || 0">
  </mat-radio-button>
  <mat-radio-button>
     <input [tabIndex]="parentTabIndex || 0">
  </mat-radio-button>
```

The tabIndex that is set on the group, will be just used by the underlying input elements of the `mat-radio-button` elements. This way the specified tabindex will be respected.

Maybe this is more a question of design and accessibility. I just think that it doesn't make any sense to allow the `radio-group` to be tabbable while the selected `radio-button` is tabbable too (doesn't really have any accessibility advantage)

The same can be seen on native radio buttons as well ([example here](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html))

CaerusKaru:
@DevVersion I fully understand now, thanks for taking the time to clarify. So to be clear on the game plan for this PR: the `mixinTabIndex`remains on the parent to capture the input, but then passes this value down (via injection) to the radio buttons, who set their own `tabIndex` attributes either based on this property, or to 0 to maintain the default (but since `input` has tabIndex by default, the `MatRadioButton` component doesn't need the mixin). Is that correct?

DevVersion:
I'm not exactly sure what the expected behavior is here. Might be good to wait for @tinayuangao's feedback as well.

I personally think that it would make sense to pass the `tabIndex` down to the `<mat-radio-button>` (as you said, somehow) and fallback to `0` by default.

Also if we do this then we need to figure out whether it would make sense to add the `tabIndex` mixin to the individual radio buttons to allow developers to override a `tabIndex` for a specific radio button. I don't really think that this is reasonable, but we should think about cases where this could be useful.

CaerusKaru:
@DevVersion I'm encountering an issue with the planned approach. When you set `tabIndex` on elements inside a parent element, the `tabIndex` is only relative to elements inside that element, not to others anywhere else in the DOM. So when the child radio elements inherit the parent `tabIndex`, they're all the same, but the tab order of the parents is still just incremental, because their `tabIndex`, as discussed, is ignored. Am I misunderstanding the implementation, or is there something subtle going on that needs to be rethought?

CaerusKaru:
@DevVersion It looks like the new method has worked, with the requisite effect being that all elements with `tabIndex` specified appear first in the DOM order when no other positive `tabIndex` values appear. I can see now why it's considered [an anti-pattern](https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex). Still waiting on feedback from @tinayuangao, but at least for now the implementation looks ready for review.

jelbourn:
@DevVersion @tinayuangao is this PR still relevant? I know radio buttons now have a tabindex, not sure about the group.

josephperrott:
Closing as we are going on ~6 months of inactvity on this PR. Feel free to open up a new PR with these changes if you would like to get them into the library.

PR #8599, opened 2017-11-22
fix(selection-model): inaccurate selected value when accessed in change subscription

### LINKED ISSUE #8584: MatSelectionList: selectedOptions.selected not up to date in onChange
When I listen to change events of the `MatSelectionList` and I use `selectedOptions.selected` to know the current selected values, they are not always up to date.  
Problem lies in [Selection.ts#125](https://github.com/angular/material2/blob/master/src/cdk/collections/selection.ts#L125):  
The `onChange.next` is called, before the cached value of `_selected` is reseted.  
Therefore the `get selected()` will return the old value.  
The bug only seems to occure, if you first set a value of another material-component, so it might be a timing-problem.

The cached value of `_selected` should be reset, so that I get the current value in `onChange`.

When calling `selectedOptions.selected`, I sometimes get the previous value.

1) Create a Angular-Material application with a MatSelectionList.  You might need another component (for example an input) too, to reproduce the bug.  
2) Add an `onChange`-Listener to the `MatSelectionList` and `console.log` the value of `selectedOptions.selected`.
2) Focus the input (or any other material component in your app).  
3) Select one of the `MatListOptions` inside the `MatSelectionList` and look at the value in the console.
   
I am using @angular/material and @angular/cdk version 5.0.0-rc0 as well as @angular version 5.0.1

A simple `window.setTimeout(() => console.log(selectedOptions.selected))` seems to solve the problem.

crisbeto:
Fixes the `selected` value being out of date in the `SelectionModel`, if it is accessed inside an `onChange` subscription.

Fixes #8584.

### DIFF
--- a/src/cdk/collections/selection.spec.ts
+++ b/src/cdk/collections/selection.spec.ts
@@ -102,6 +102,19 @@ describe('SelectionModel', () => {
       expect(event.added).toEqual([2]);
     });
 
+    it('should have updated the selected value before emitting the change event', () => {
+      let model = new SelectionModel(true);
+      let spy = jasmine.createSpy('SelectionModel change event');
+
+      // Note: this assertion is only here to run the getter.
+      expect(model.selected).toEqual([]);
+
+      model.onChange!.subscribe(() => spy(model.selected));
+      model.select(1);
+
+      expect(spy).toHaveBeenCalledWith([1]);
+    });
+
     describe('selection', () => {
       let model: SelectionModel<any>;
       let spy: jasmine.Spy;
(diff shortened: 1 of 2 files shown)


simeyla:
No chance of ever add `selected` to the interface? 

export declare class SelectionChange<T> {
    added: T[] | undefined;
    removed: T[] | undefined;
    selected: T[] | undefined;
    constructor(added?: T[] | undefined, removed?: T[] | undefined);
}

I realize it's a change event, but as a convenience it would be very handy.

crisbeto:
@simeyla the intention with the `selected` is that we try to avoid recomputing it until it is requested. That being said, I think that we can expose the `SelectionModel`, that triggered the event, for convenience which will allow you to get the selected value without having to reach outside the function.

PR #8603, opened 2017-11-22
fix(datepicker): placeholder not floating when an invalid value is typed in

### LINKED ISSUE #8575: Datepicker hint displayed on top of value on incorrect input
Bug

The hint on the datapicker should be red(or something else) when an incorrect value is entered

The hint on the datepicker is displayed on top of the value when an incorrect value is entered

Open the material datepicker examples : https://material.angular.io/components/datepicker/examples
Enter `asd` in the first datepicker
Lose focus

The hint and the value are not readable anymore

Material rc.1
Angular 5.0.2

Windows Chrome 62
Windows FireFox 57

crisbeto:
Fixes the datepicker input's placeholder not floating if the user types in an invalid value. This was due to the datepicker's value being parsed to `null` which the form field inferred as empty. These changes switch to always check the DOM element, rather than the actual control value.

Fixes #8575.

### DIFF
--- a/src/lib/input/input.ts
+++ b/src/lib/input/input.ts
@@ -284,12 +284,7 @@ export class MatInput implements MatFormFieldControl<any>, OnChanges, OnDestroy,
 
   // Implemented as part of MatFormFieldControl.
   get empty(): boolean {
-    return !this._isNeverEmpty() &&
-        (this.value == null || this.value === '') &&
-        // Check if the input contains bad input. If so, we know that it only appears empty because
-        // the value failed to parse. From the user's perspective it is not empty.
-        // TODO(mmalerba): Add e2e test for bad input case.
-        !this._isBadInput();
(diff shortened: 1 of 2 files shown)


src/lib/input/input.ts
```diff
@@ -264,13 +264,6 @@ export class MatInput implements MatFormFieldControl<any>, OnChanges, OnDestroy,
     return this._neverEmptyInputTypes.indexOf(this._type) > -1;
   }
 
-  /** Checks whether the input is invalid based on the native validation. */
-  protected _isBadInput() {
```
mmalerba: can't get rid of this, it's needed for cases where the browser considers the value invalid, like this: https://jsfiddle.net/3bbhk7ka/

crisbeto: Oh, good to know. I've re-added it.

PR #8614, opened 2017-11-23
fix(menu): Set menu-item icon color only when not set on mat-icon

### LINKED ISSUE #8594: <mat-icon> ignores color attribute if inside <button mat-menu-item>
It's a **bug**.

I would expect `<mat-icon color="accent">` in my menu item to respect the color attribute while it is inside of its <button mat-menu-item> parent tag.

It does not apply the color, here is a screenshot with comparison between `<button mat-button>` which respects the attribute and `<button mat-menu-item>` which does not.
image

Here is a bit edited mat-menu plunker with the problem displayed [link](https://plnkr.co/edit/j8qKe1jY9oVt1dHU4WMT?p=preview)

I think it should be repaired, since the `<mat-icon>` explicitly supports coloring with color attribute.

Angular v5.0.2
@angular/cdk v5.0.0-rc.1
@angular/material v5.0.0-rc.1
TypeScript v2.4.2
IE 11, Firefox 57, Chrome 62

glenstaes:
Set the theme icon color only if the `color` input property was not set on the `mat-icon` component.

Fixes #8594

### DIFF
--- a/src/lib/menu/menu.spec.ts
+++ b/src/lib/menu/menu.spec.ts
@@ -25,7 +25,7 @@ import {
   MenuPositionY,
   MatMenuItem,
 } from './index';
-import { MatIconModule } from "../icon/index";
+import { MatIconModule } from '../icon/index';
 import {MENU_PANEL_TOP_PADDING} from './menu-trigger';
 import {MatRipple} from '@angular/material/core';
 import {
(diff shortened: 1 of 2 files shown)


src/lib/menu/menu.spec.ts
```diff
@@ -223,6 +225,32 @@ describe('MatMenu', () => {
     expect(fixture.componentInstance.items.first.getLabel()).toBe('Item');
   });
 
+  it('should have an icon with the default theme color if mat-icon has no color attribute set',
```
crisbeto: You can get rid of these tests since they're basically testing the CSS engine.

src/lib/menu/menu.spec.ts
```diff
@@ -44,7 +45,7 @@ describe('MatMenu', () => {
   beforeEach(async(() => {
     dir = 'ltr';
     TestBed.configureTestingModule({
-      imports: [MatMenuModule, NoopAnimationsModule],
+      imports: [MatIconModule, MatMenuModule, NoopAnimationsModule],
```
crisbeto: In general we try to avoid having dependencies between components like this. It shouldn't be an issue after the previous comment about removing the tests is addressed.

glenstaes:
@crisbeto Comments have been addressed

PR #8619, opened 2017-11-23
fix(autosize): not updating when window is resized

### LINKED ISSUE #8610: matTextareaAutosize does not get updated on window resize
`matTextareaAutosize` does not update the length when window dimensions are changing. The textarea then is either not displaying the whole content or is higher then the required lines for the content.

It should resize when changing the dimensions of the window.

Angular  5.0.0
Material 5.0.0-rc.1
TypeScript 2.4.2
Chrome 62

crisbeto:
Fixes the autosize textarea not recalculating its height when the viewport size has changed.

Fixes #8610.

### DIFF
--- a/src/lib/input/autosize.ts
+++ b/src/lib/input/autosize.ts
@@ -6,8 +6,20 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Directive, ElementRef, Input, AfterViewInit, DoCheck} from '@angular/core';
+import {
+  Directive,
+  ElementRef,
+  Input,
+  AfterViewInit,
+  DoCheck,
+  OnDestroy,
+  NgZone,
+} from '@angular/core';
 import {Platform} from '@angular/cdk/platform';
+import {fromEvent} from 'rxjs/observable/fromEvent';
+import {auditTime} from 'rxjs/operators/auditTime';
+import {takeUntil} from 'rxjs/operators/takeUntil';
+import {Subject} from 'rxjs/Subject';
 
 
 /**
@@ -22,9 +34,10 @@ import {Platform} from '@angular/cdk/platform';
     'rows': '1',
   },
 })
-export class MatTextareaAutosize implements AfterViewInit, DoCheck {
+export class MatTextareaAutosize implements AfterViewInit, DoCheck, OnDestroy {
   /** Keep track of the previous textarea value to avoid resizing when the value hasn't changed. */
   private _previousValue: string;
+  private _destroyed = new Subject<void>();
 
   private _minRows: number;
   private _maxRows: number;
@@ -47,7 +60,12 @@ export class MatTextareaAutosize implements AfterViewInit, DoCheck {
   /** Cached height of a textarea with a single row. */
   private _cachedLineHeight: number;
 
-  constructor(private _elementRef: ElementRef, private _platform: Platform) {}
+  constructor(
+    private _elementRef: ElementRef,
+    private _platform: Platform,
+    private _ngZone?: NgZone) {}
+
+  // TODO(crisbeto): make the `_ngZone` a required param in the next major version.
 
   /** Sets the minimum height of the textarea as determined by minRows. */
   _setMinHeight(): void {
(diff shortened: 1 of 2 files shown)


src/lib/input/autosize.ts
```diff
@@ -72,9 +87,19 @@ export class MatTextareaAutosize implements AfterViewInit, DoCheck {
...
+      this._resize = this._ngZone.runOutsideAngular(() => {
+        return fromEvent(window, 'resize')
+          .pipe(debounceTime(10))
+          .subscribe(() => this.resizeToFitContent(true));
+      });
     }
   }
 
+  ngOnDestroy() {
+    this._resize.unsubscribe();
```
mmalerba: Haven't we been switching to the `_destoryed.next()` style?

src/lib/input/autosize.ts
```diff
@@ -72,9 +87,19 @@ export class MatTextareaAutosize implements AfterViewInit, DoCheck {
   ngAfterViewInit() {
     if (this._platform.isBrowser) {
       this.resizeToFitContent();
+
+      this._resize = this._ngZone.runOutsideAngular(() => {
+        return fromEvent(window, 'resize')
+          .pipe(debounceTime(10))
```
mmalerba: wouldn't `auditTime` be more appropriate here?

src/lib/input/autosize.ts
```diff
@@ -132,8 +157,12 @@ export class MatTextareaAutosize implements AfterViewInit, DoCheck {
     }
   }
 
-  /** Resize the textarea to fit its content. */
-  resizeToFitContent() {
+  /**
+   * Resize the textarea to fit its content.
+   * @param force Whether to force a height recalculation. By default the height will be
+   * recalculated only if the value changed since the last call.
```
mmalerba: nit: indent

src/lib/input/autosize.ts
crisbeto: I don't think so. We want to prevent this from happening if we get a bunch of resize events close to each other.

crisbeto:

src/lib/input/autosize.ts
mmalerba: But in this case the rest of the page will be smoothly resizing and the textarea will just jump to the correct size at the end right? Wouldn't it be better to have it also smoothly resize?

crisbeto: I think that the delay is so small that people won't really notice it if they're flipping their phone. The only ones that would would be developers, but I'd prioritize less repaints to dev satisfaction while resizing the window.

mmalerba: the delay is small but if you get a resize event every 9ms for 1s, doesn't that mean the resize won't actually trigger until after 1s? (or I'm misunderstanding how `debounceTime` works)

"window of time required to wait for emission silence before emitting the most recent source value" (from http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-debounceTime)

crisbeto: It does, but the way this usually plays out is that the browser either fires a few of them at a time if you change your screen orientation, or one per pixel if you're resizing it manually. It's really hard to move your browser resize handle at a constant 1px per 9ms since it's about the speed of human reaction time.

mmalerba: `auditTime` seems like a more natural fit for this though, since it will give us the most recent value every 10ms (or we could up it to 16ms which is ~60fps)

crisbeto: Makes sense, I've switched it.

mmalerba:
I think this is technically breaking due to the addition of `NgZone` to the constructor, so added to 6.0 milestone

crisbeto:
Made it optional, although I'm not sure how common it would be to extend this directive in particular.

src/lib/input/autosize.ts
```diff
@@ -47,7 +60,10 @@ export class MatTextareaAutosize implements AfterViewInit, DoCheck {
   /** Cached height of a textarea with a single row. */
   private _cachedLineHeight: number;
 
-  constructor(private _elementRef: ElementRef, private _platform: Platform) {}
+  constructor(
+    private _elementRef: ElementRef,
+    private _platform: Platform,
+    private _ngZone?: NgZone) {}
```
mmalerba: nit: add TODO to make non-optional at next major

Issue #7700, opened 2017-10-11
[Mat-Stepper] A few suggestions


christikaes:
Thanks for everyone who works on this project! I just upgraded to beta12 and am using the Stepper component now. I had a few proposals after using it for a bit. Please let me know what you think-I'd love to help contribute if we want any of these (: 

1. Mat-Stepper with horizontal/vertical as an input.
    * Use Case: I have a stepper that I want to toggle between horizontal/vertical for different screen aspect ratios but with the same content and steps. Currently I would need to create two separate components and toggle between the two which feels messy.
    * Proposal: Replace mat-vertical-stepper and mat-horizontal-stepper with mat-stepper that takes in an input to toggle between the two views
2. SelectedIndex as input.
    * Use Case: I want to programmatically change the current selectedStep. There was a question asked [here](https://stackoverflow.com/questions/46440642/angular-4-material-stepper) about how to switch to the first step by setting the stepper.selectedIndex. I haven't been able to get this to work (possibly because of [ChangeDetectionStrategy.OnPush](https://github.com/angular/material2/commit/c2c6e04ed64e6952bf3ee26b5c31dfe5ab66e2ae#diff-ee17c85148fbbf019e1dd74184f301e0))?
    * Proposal: Stepper should take SelectedIndex as an input
    * EDIT: Should be fixed by https://github.com/angular/material2/issues/8031
3. Reset Stepper.
    * Use Case: I want to reset the stepper. I am using formGroup.reset() to reset each form, but this does not reset the visited states of each one of the steps. 
    * Proposal: Some way of resetting the stepper component to the initial state: First step is selected, Forms are reset, None of the steps have been touched etc.
4. Animate Height.
    * Use Case: Currently the steps slide right/left but the height of the stepper does not transition smoothly.
    * Proposal: Transition the height of the steps when changing from one step to the other
Let me know what you think~

sam-lex:
Would like to add one more suggestion here as well:  
5. Disabled step. Or a way to add/remove steps.
- Use Case: Additional steps depending on what is being selected in previous steps. (like a wizard form).

I was able to add steps with something like this:  
```html
<mat-step *ngFor="let invForm of inventoryForms"
    [stepControl]="invForm.form" [label]="invForm.item.product.name">
</mat-step>
```
but if I go back to previous step, and change it so that these additional steps aren't required, and setting `this.inventoryForms = []`, the steps are not removed from the stepper.  
[...]

PR #8623, opened 2017-11-23
feat(stepper): add the ability to reset a stepper

crisbeto:
Allows for the user to reset a stepper to its initial state by calling the `reset` method.

Relates to #7700.

### DIFF
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -123,6 +123,16 @@ export class CdkStep implements OnChanges {
     this._stepper.selected = this;
   }
 
+  /** Resets the step to its initial state. Note that this includes resetting form data. */
+  reset(): void {
+    this.interacted = false;
+    this.completed = false;
+
+    if (this.stepControl) {
+      this.stepControl.reset();
+    }
+  }
+
   ngOnChanges() {
     // Since basically all inputs of the MatStep get proxied through the view down to the
     // underlying MatStepHeader, we have to make sure that change detection runs correctly.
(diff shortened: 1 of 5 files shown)


src/cdk/stepper/stepper.ts
```diff
@@ -120,6 +120,16 @@ export class CdkStep implements OnChanges {
     this._stepper.selected = this;
   }
 
+  /** Resets the step to its initial state. Note that this includes resetting form data. */
+  reset(): void {
+    this.interacted = false;
+    this.completed = false;
+
+    if (this.stepControl) {
+      this.stepControl.reset();
```
crisbeto: Note: I'm open to having this reset only the `dirty` and `touched` states. I went with `reset`, because leaving behind the value didn't seem like a "true" reset.

mmalerba: This seems like the most common use case to me. If we get  lot of requests for reset all but the value we could create a new method: `{reset,clear}{Status,State}`

mmalerba:
@crisbeto  I set the milestone to 5.2 since this seems less severe than some of the other things we want to get in for 5.1. I'd be fine moving it to 5.1 if we have time after merging the other fixes

crisbeto:
I'm fine with leaving it for 5.2.

andrewseguin:
See CI errors 

```
Error at /home/travis/build/angular/material2/dist/packages/e2e-app/material-examples/typings/index.d.ts.StepperOverviewExample.html(1,1030): Property 'matHorizontalStepper' does not exist on type 'StepperOverviewExample'.
```

crisbeto:
Sorted out the error in the example.

PR #8642, opened 2017-11-25
fix(portal): inaccurate hasAttached result and portal being cleared if attached too early

### LINKED ISSUE #8628: CdkPortalOutlet hasAttached not called when using attachComponentPortal
Bug report

When I attach a `Component` or a `Template` to a `CdkPortalOutlet` using `attachComponentPortal` or `attachTemplatePortal`, I expect that `hasAttached` returns true.

`hasAttached` always returns false, except, when using the `cdkPortalOutlet`-input t oset the `portal`.

1) Create a `Component`, which contains a `cdkPortalOutlet`.
2) Get a reference to the `CdkPortalOutlet`-Directive using `@ViewChild(CdkPortalOutlet)`.
3) Attach a `Template` to it, using `attachTemplatePortal`.
4) Call `hasAttached()` which should return false.

I am using Angular Material+CDK version 5.0.0-rc.1 as awell as Angular version 5.0.3

The problem is in [portal-directives.ts#L114](https://github.com/angular/material2/blob/master/src/cdk/portal/portal-directives.ts#L114) and [portal-directives.ts#L129](https://github.com/angular/material2/blob/master/src/cdk/portal/portal-directives.ts#L129), where `this._portal = portal` should probably be `this.portal = portal`, so that the setter is called and updates the `_attachedPortal` of the super-class.

crisbeto:
Fixes the `hasAttached` method always returning false when the portal was attached using `CdkPortalOutlet.attachComponentPortal` or `CdkPortalOutlet.attachTemplatePortal`. This means that checks like https://github.com/angular/material2/blob/master/src/lib/dialog/dialog-container.ts#L118 always evaluate to false.

Fixing `hasAttached` revealed another issue: if a portal outlet is unbound (e.g. `<div cdkPortalOutlet>`) and the consumer attaches something to it on the before the first CD round, the content will be cleared immediately due to Angular invoking the getter with an empty string. This has been fixed by not clearing any previously-attached portals if the reset value is set before `ngOnInit`.

Fixes #8628.

### DIFF
--- a/src/cdk/portal/portal-directives.ts
+++ b/src/cdk/portal/portal-directives.ts
@@ -7,15 +7,16 @@
  */
 
 import {
-    NgModule,
-    ComponentRef,
-    Directive,
-    EmbeddedViewRef,
-    TemplateRef,
-    ComponentFactoryResolver,
-    ViewContainerRef,
-    OnDestroy,
-    Input,
+  NgModule,
+  ComponentRef,
+  Directive,
+  EmbeddedViewRef,
+  TemplateRef,
+  ComponentFactoryResolver,
+  ViewContainerRef,
+  OnDestroy,
+  OnInit,
+  Input,
 } from '@angular/core';
 import {Portal, TemplatePortal, ComponentPortal, BasePortalOutlet} from './portal';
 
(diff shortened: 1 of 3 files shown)


src/cdk/portal/portal-directives.ts
```diff
@@ -47,9 +48,8 @@ export class CdkPortal extends TemplatePortal<any> {
   exportAs: 'cdkPortalOutlet, cdkPortalHost',
   inputs: ['portal: cdkPortalOutlet']
 })
-export class CdkPortalOutlet extends BasePortalOutlet implements OnDestroy {
-  /** The attached portal. */
-  private _portal: Portal<any> | null = null;
+export class CdkPortalOutlet extends BasePortalOutlet implements OnInit, OnDestroy {
+  private _hasInitialized = false;
```
jelbourn: Add description? It's not clear from the name whether this means the directive itself is initialized or something else (maybe `_isInitialized`?)

crisbeto:

PR #8643, opened 2017-11-25
feat(sort): add the ability to disable sort toggling

### LINKED ISSUE #8622: [Sort] Add disable input
Proposal: Allow the mat-sort-header tag to be attached to a boolean.

There should be a way to bind a column to a variable so you could enable or disable the the sorting.

If there is the mat-sort-header tag on the column you can't disable the sorting.

https://plnkr.co/edit/SaM2RYwm3i7mRJhOIH2k?p=preview
I wish that clicking the button would disable the sorting.

This would make the sorting more flexible.

All of them

crisbeto:
Adds the ability for a `mat-sort-header` or `mat-sort` instance to be disabled, preventing the user from changing the sorting direction.

Fixes #8622.

### DIFF
--- a/src/lib/sort/sort-header.ts
+++ b/src/lib/sort/sort-header.ts
@@ -30,9 +30,17 @@ import {MatSort, MatSortable} from './sort';
 import {MatSortHeaderIntl} from './sort-header-intl';
 import {getSortHeaderNotContainedWithinSortError} from './sort-errors';
 import {AnimationCurves, AnimationDurations} from '@angular/material/core';
+import {CanDisable, mixinDisabled} from '@angular/material/core';
+
 
 const SORT_ANIMATION_TRANSITION =
     AnimationDurations.ENTERING + ' ' + AnimationCurves.STANDARD_CURVE;
+
+// Boilerplate for applying mixins to the sort header.
+/** @docs-private */
+export class MatSortHeaderBase {}
+export const _MatSortHeaderMixinBase = mixinDisabled(MatSortHeaderBase);
+
 
 /**
  * Applies sorting behavior (click to change sort) and styles to an element, including an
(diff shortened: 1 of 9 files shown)


src/lib/sort/sort-header.ts
```diff
@@ -50,12 +58,14 @@ const SORT_ANIMATION_TRANSITION =
...
   host: {
-    '(click)': '_sort.sort(this)',
+    '(click)': '_handleClick()',
     '[class.mat-sort-header-sorted]': '_isSorted()',
+    '[class.mat-sort-header-disabled]': '_isDisabled()',
   },
   encapsulation: ViewEncapsulation.None,
   preserveWhitespaces: false,
   changeDetection: ChangeDetectionStrategy.OnPush,
+  inputs: ['disabled: matSortHeaderDisabled'],
```
andrewseguin: Should be `disabled`

andrewseguin:
There's an open PR similar to this (#7786) but yours includes a way to disable the MatSort so let's go with that.

src/lib/sort/sort-header.ts
crisbeto: I didn't go with `disabled`, because while it is a component, the select is still an attribute so it's not super obvious what you're disabling.

andrewseguin: Yeah, just a funny consequence of our rules - might loop in @jelbourn for confirmation on this but as a rule, components don't prefix

crisbeto: Switched it to `disabled` @andrewseguin.

andrewseguin:
LGTM

mdrafee03:
apply with `mat-sort-header`, `[disabled] = "condition to be true"`
Example
`<mat-header-cell *matHeaderCellDef mat-sort-header  [disabled]="!sortedColumns.includes(col)"> {{ helper.extract(labels, col) }} </mat-header-cell>`

PR #8646, opened 2017-11-25
fix(stepper): block linear stepper for pending components

### LINKED ISSUE #8645: In linear steppers, steps with pending components should not be considered completed
Bug

In a linear stepper, going to the next step should require that no components of the step are in invalid or pending states (i.e. all components are valid or disabled). Pending components (i.e. components with async validators pending) should block the step.

Pending components do not block the step.

Example: http://plnkr.co/edit/aFMUy13jRinsHHudFltm?p=preview

The first step has a single input component, with two validators:
- `validators.required` (sync)
- a custom async validator, validating that the input is not `bob`, but with a 2 seconds delay (to mock a remote check)

Type "bob", wait a few seconds for the check to complete, then try to go the second step. It will block: first step is required.

Change value to "". Then type "bob" again, but go the second step while validation is pending. The stepper is not blocked.

In a linear stepper, validation is expected to be required.

Angular 5.0.3, Material 5.0.0-rc1

olivierlemasle:
In a linear stepper, going to the next step should require that no components of the step are in invalid or pending states (i.e. all components are valid or disabled). Pending components (i.e. components with async validators pending) should block the step.

Fixes #8645

### DIFF
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -152,7 +152,7 @@ export class CdkStepper implements OnDestroy {
   get selectedIndex() { return this._selectedIndex; }
   set selectedIndex(index: number) {
     if (this._steps) {
-      if (this._anyControlsInvalid(index) || index < this._selectedIndex &&
+      if (this._anyControlsInvalidOrPending(index) || index < this._selectedIndex &&
           !this._steps.toArray()[index].editable) {
         // remove focus from clicked step header if the step is not able to be selected
         this._stepHeader.toArray()[index].nativeElement.blur();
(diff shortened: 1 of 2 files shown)


src/cdk/stepper/stepper.ts
```diff
@@ -291,13 +291,15 @@ export class CdkStepper implements OnDestroy {
...
-  private _anyControlsInvalid(index: number): boolean {
+  private _anyControlsInvalidOrPending(index: number): boolean {
     const steps = this._steps.toArray();
 
     steps[this._selectedIndex].interacted = true;
 
     if (this._linear && index >= 0) {
-      return steps.slice(0, index).some(step => step.stepControl && step.stepControl.invalid);
+      return steps.slice(0, index).some(step =>
+        step.stepControl && (step.stepControl.invalid || step.stepControl.pending)
```
mmalerba: can you add a unit test for a pending control?

olivierlemasle: @mmalerba Sure. I've added unit tests and rebased the PR.

PR #8653, opened 2017-11-26
fix(menu): update position on every open

EladBezalel:
- menu overlay is only created once and the position of the overlay wasn't calculated again

### DIFF
--- a/src/lib/menu/menu-trigger.ts
+++ b/src/lib/menu/menu-trigger.ts
@@ -308,7 +308,9 @@ export class MatMenuTrigger implements AfterContentInit, OnDestroy {
    * OverlayRef so that it can be attached to the DOM when openMenu is called.
    */
   private _createOverlay(): OverlayRef {
-    if (!this._overlayRef) {
+    if (this._overlayRef) {
+      this._overlayRef.getConfig().positionStrategy = this._getPosition();
+    } else {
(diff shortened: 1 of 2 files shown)


src/lib/menu/menu-trigger.ts
```diff
@@ -313,6 +313,8 @@ export class MatMenuTrigger implements AfterContentInit, OnDestroy {
       const config = this._getOverlayConfig();
       this._subscribeToPositions(config.positionStrategy as ConnectedPositionStrategy);
       this._overlayRef = this._overlay.create(config);
+    } else {
```
crisbeto: Flip the `if/else` here so it's something like this It should be a little easier to read through:

```
if (this._overlayRef) {
  // update position
} else {
  // create overlay
}
```

jelbourn:
This will indeed conflict with #7376, which I plan on getting in first

EladBezalel:
@jelbourn than let's wait with this and #8655 until it get's in

crisbeto:
Thinking about this one again, I think that overwriting the position strategy can cause some subtle bugs, because it means that any subscriptions to the `onPositionChange` will be broken once the new strategy is assigned. I've submitted #8800 that allows for the positions to be updated instead.

jelbourn:
Closing this since (based on the comments) we'd want to do a different approach.

PR #8656, opened 2017-11-26
feat(close-scroll-strategy): add scroll threshold option

crisbeto:
Adds the `threshold` option to the `CloseScrollStrategy` that allows for the consumer to specify the amount of pixels that the user has to scroll before the overlay is closed.

### DIFF
--- a/src/cdk/overlay/scroll/close-scroll-strategy.ts
+++ b/src/cdk/overlay/scroll/close-scroll-strategy.ts
@@ -5,13 +5,19 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-
 import {NgZone} from '@angular/core';
 import {ScrollStrategy, getMatScrollStrategyAlreadyAttachedError} from './scroll-strategy';
 import {OverlayRef} from '../overlay-ref';
 import {Subscription} from 'rxjs/Subscription';
-import {ScrollDispatcher} from '@angular/cdk/scrolling';
+import {ScrollDispatcher, ViewportRuler} from '@angular/cdk/scrolling';
 
+/**
+ * Config options for the CloseScrollStrategy.
+ */
+export interface CloseScrollStrategyConfig {
+  /** Amount of pixels the user has to scroll before the overlay is closed. */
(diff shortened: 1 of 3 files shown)


src/cdk/overlay/scroll/close-scroll-strategy.ts
```diff
@@ -31,19 +41,27 @@ export class CloseScrollStrategy implements ScrollStrategy {
...
+      this._ngZone.run(() => {
+        this.disable();
+
+        if (this._overlayRef.hasAttached()) {
+          this._overlayRef.detach();
+        }
+      });
+    };
+
+    this._scrollSubscription = threshold > 1 ? stream.pipe(take(threshold)).subscribe(() => {
```
jelbourn: I don't follow why this passes `threshold` to `take`

crisbeto: When we pass it to `take`, it'll call the subscription callback to update the position, until it hits the threshold when it'll complete the observable which detaches the overlay.

jelbourn: Isn't `take` the number of event, though? I don't see how it ties to number of pixels

crisbeto: The `scrollDispatcher.scrolled` fires once per pixel scrolled, unless it's debounced but we've disabled it by passing in zero.

jelbourn: I'd be skeptical that that would be true across all platforms and screen-sizes. Either way, though, a user could also scroll back and forth a bit, no?

crisbeto: I'm pretty sure that even mobile devices fire for every pixel now, but you're right that people could scroll back and forth. What do you think about saving the scroll position on open and then checking that you've scrolled X amount of pixels above/below it?

jelbourn: That seems reliable

crisbeto:
Reworked based on our discussion @jelbourn, can you take another look?

PR #8669, opened 2017-11-27
fix(focus-origin): focus origin sometimes invalid in firefox 57

### LINKED ISSUE #6984: Checkbox: focus effect sometimes not showing up in Firefox
Bug

After tabbing between checkboxes, focused checkbox should be visually marked.

Sometimes after tabbing checkbox is not visually marked despite that it's focused.

Demo, select first checkbox and press tab multiple times
https://stackblitz.com/edit/angular-hrkwcy
This bug also occurs on official checkbox example
https://material.angular.io/components/checkbox/examples

Encountered this bug only on Firefox (55.0.3), works fine on Chrome and IE11
Material: 2.0.0-beta.10
Angular: 4.3.6
Typescript: 2.3.3
OS: Windows 8.1 Pro (tested also on Win 7 and 10, worked fine)

Setting timeout to 3ms instead of 0 seems to fix this bug although I'm not sure if it's the right way to do it.
https://github.com/angular/material2/blob/f2cae6e8141c90fa0b62dc8a928f217bc085667b/src/lib/core/style/focus-origin-monitor.ts#L213-L216

When the bug occurs this line is executed two times
https://github.com/angular/material2/blob/f2cae6e8141c90fa0b62dc8a928f217bc085667b/src/lib/core/style/focus-origin-monitor.ts#L215
despite that this line is executed only one time
https://github.com/angular/material2/blob/f2cae6e8141c90fa0b62dc8a928f217bc085667b/src/lib/core/style/focus-origin-monitor.ts#L214
As a result `origin` is equal to `null` (instead of `'keyboard'`) when `_onFocus` method is called, which later leads to incorrect checkbox styling.

DevVersion:
* Sometimes the focus origin is not valid because Firefox seems to focus a tick after the keyboard/interaction event fired. To ensure the focus origin is always correct, the focus origin will be added to the beginning of the event loop in the next tick. 

Fixes #6984

### DIFF
--- a/src/cdk/a11y/focus-monitor/focus-monitor.spec.ts
+++ b/src/cdk/a11y/focus-monitor/focus-monitor.spec.ts
@@ -54,7 +54,7 @@ describe('FocusMonitor', () => {
     dispatchKeyboardEvent(document, 'keydown', TAB);
     buttonElement.focus();
     fixture.detectChanges();
-    tick();
+    flush();
 
     expect(buttonElement.classList.length)
         .toBe(2, 'button should have exactly 2 focus classes');
@@ -70,7 +70,7 @@ describe('FocusMonitor', () => {
     dispatchMouseEvent(buttonElement, 'mousedown');
     buttonElement.focus();
     fixture.detectChanges();
-    tick();
+    flush();
 
     expect(buttonElement.classList.length)
         .toBe(2, 'button should have exactly 2 focus classes');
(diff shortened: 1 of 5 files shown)


mmalerba:
needs rebase

mmalerba:
@DevVersion please rebase

DevVersion:
@mmalerba Rebased. I didn't apply `merge ready` because I'm not sure whether we still want to go that way.

mmalerba:
I think we should try, it may wind up causing too many failures in google3 to be feasible. I'll mark it merge ready so the caretaker can at least try though

mmalerba:
@DevVersion it looks like we did presubmit this a while back and it seemed fine, it has a failing test on travis though

DevVersion:
@mmalerba Should be fixed now.

PR #8674, opened 2017-11-27
refactor: remove the OverlayContainer stubs from all of the tests

crisbeto:
Gets rid of all the manual stubs of the `OverlayContainer` in the tests and replaces them with getting the container through DI. This removes some boilerplate and should make things less prone to breaking if we make any changes to the container.


tinayuangao:

crisbeto:
Rebased @tinayuangao.

PR #8692, opened 2017-11-28
fix(input): align caret color with spec

crisbeto:
Based on the text field spec (https://material.io/guidelines/components/text-fields.html), the input caret color should match the underline color.

### DIFF
--- a/src/lib/input/_input-theme.scss
+++ b/src/lib/input/_input-theme.scss
@@ -6,6 +6,9 @@
 
 
 @mixin mat-input-theme($theme) {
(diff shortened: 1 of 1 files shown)


andrewseguin:
Looks like caret-color is not a supported CSS property internally. Probably since it is not supported across browsers. @jelbourn Have you run into this before with other CSS properties?

jelbourn:
Did some investigation w/ Andrew: we're going to try to update the internal css compiler to be aware of this property

jelbourn:
Andrew sent a fix for the missing property, scheduled another presubmit now

PR #8697, opened 2017-11-29
fix(progress-spinner): element size not updated when diamater is changed

andreroggeri:
Updating the diameter property programmatically don't trigger the ngOnChanges events.

This change updates the code to calculate the _elementSize property on diameter change

### DIFF
--- a/src/lib/progress-spinner/progress-spinner.spec.ts
+++ b/src/lib/progress-spinner/progress-spinner.spec.ts
@@ -224,6 +224,14 @@ describe('MatProgressSpinner', () => {
     expect(svgElement.getAttribute('viewBox')).toBe('0 0 38 38');
   });
 
+  it('should update the element size when changed dynamically', () => {
+    let fixture = TestBed.createComponent(BasicProgressSpinner);
+    let spinner = fixture.debugElement.query(By.directive(MatProgressSpinner));
+    spinner.componentInstance.diameter = 32;
+    fixture.detectChanges();
+    expect(spinner.nativeElement.style.width).toBe('32px');
+    expect(spinner.nativeElement.style.height).toBe('32px');
+  });
 });
 
 
--- a/src/lib/progress-spinner/progress-spinner.ts
+++ b/src/lib/progress-spinner/progress-spinner.ts
@@ -122,6 +122,7 @@ export class MatProgressSpinner extends _MatProgressSpinnerMixinBase implements 
     if (!this._fallbackAnimation && !MatProgressSpinner.diameters.has(this._diameter)) {
       this._attachStyleNode();
     }
+    this._updateElementSize();
   }
   private _diameter = BASE_SIZE;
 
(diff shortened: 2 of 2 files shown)


jelbourn:
Hi, thanks for taking the time to send a PR! This change would cause the browser to potentially trigger a re-layout on every change detection cycle, which is something we would want to avoid since it could hurt application performance. 

The right approach would be to have the component mark itself as needing change detection when the diameter changes.

andreroggeri:
Hi @jelbourn thanks for the feedback.

I tried to use the methods `markForCheck` and `detectChanges` from `ChangeDetectorRef`. But none of these made the test pass.

After some research, I found this [stackoverflow question](https://stackoverflow.com/questions/43112685/change-detection-not-working-when-creating-a-component-via-componentfactoryresol). And it seems that the change detection will not work for input without bindings.

You said that this change can trigger a re-layout on every detection cycle, but the elementSize is only recalculated when the changes are for strokeWidth or diamter properties. Is there a way to check this behavior ?

jelbourn:
@crisbeto can you take a look at this PR?

src/lib/progress-spinner/progress-spinner.spec.ts
```diff
@@ -203,6 +203,12 @@ describe('MatProgressSpinner', () => {
     expect(svgElement.getAttribute('viewBox')).toBe('0 0 38 38');
   });
 
+  it('should update the element size when changed dynamically', () => {
+    const fixture = TestBed.createComponent(MatProgressSpinner);
+    const spinner = fixture.componentInstance;
+    spinner.diameter = 32;
+    expect(spinner._elementSize).toBe(32);
```
crisbeto: Can you test the styles that are applied to the element, rather than the property itself? You can look at how the other tests are doing it.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -6,6 +6,14 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
```
crisbeto: Can you remove this license banner? There is one already at the top of the file.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -231,6 +240,11 @@ export class MatProgressSpinner extends _MatProgressSpinnerMixinBase implements
         .replace(/END_VALUE/g, `${0.2 * this._strokeCircumference}`)
         .replace(/DIAMETER/g, `${this.diameter}`);
   }
+
+  /**Updates the spinner element size based on its diameter */
```
crisbeto: Needs a space after the `/**` and a dot at the end of the sentence.

src/lib/progress-spinner/progress-spinner.ts
```diff
@@ -231,6 +240,11 @@ export class MatProgressSpinner extends _MatProgressSpinnerMixinBase implements
...
+  private _updateSpinnerDimension() {
```
crisbeto: Rename to `_updateElementSize`.

andreroggeri:
@crisbeto Sure, changes done !

PR #8698, opened 2017-11-29
fix(table): data source should sort empty values correctly

### LINKED ISSUE #8485: [Table] MatTableDataSource does not properly sort null/undefined values
Sort of bug, sort of feature request

rows with undefined values for a sorted column should sort properly

null/undefined values cause random sorting issues

Plunker starter (using on `@master`): https://goo.gl/uDmqyY<br/>

Have a sortable mat-table with some null values in some columns.
Try to sort.

columns often have empty values

material 5.0.0rc0

MatTableDataSource has this line where the actual sorting is done:

`      return (valueA < valueB ? -1 : 1) * (direction == 'asc' ? 1 : -1);`

Note that like values are not treated correctly either. This could be replaced with the following:

```
     return this.compareItems(valueA, valueB) * (direction == 'asc' ? 1 : -1);

    private compareItems(itemA: any, itemB: any): number {
      let retVal: number = 0;
      if (itemA && itemB) {
        if (itemA > itemB) retVal = 1;
        else if (itemA < itemB) retVal = -1;
      }
      else if (itemA) retVal = 1;
      else if (itemB) retVal = -1;
      return retVal;
    }

```
Also, to make strict null safe, the sortingDataAccessor method should be modified to allow such a value.

andrewseguin:
Also takes the sort function and adds it as an override-able predicate for users to inject their own method of sorting. It takes a `MatSort` rather than the `active` and `directive` properties in hopes that it's more stable in case we add properties to `MatSort` later on. Open to opinions on whether that's wise or not

Fixes #8485

### DIFF
--- a/src/cdk/coercion/number-property.ts
+++ b/src/cdk/coercion/number-property.ts
@@ -10,8 +10,16 @@
 export function coerceNumberProperty(value: any): number;
 export function coerceNumberProperty<D>(value: any, fallback: D): number | D;
 export function coerceNumberProperty(value: any, fallbackValue = 0) {
+  return _isNumberValue(value) ? Number(value) : fallbackValue;
+}
+
+/**
+ * Whether the provided value is considered a number.
+ * @docs-private
+ */
+export function _isNumberValue(value: any): boolean {
   // parseFloat(value) handles most of the cases we're interested in (it treats null, empty string,
   // and other non-number values as NaN, where Number just uses 0) but it considers the string
   // '123hello' to be a valid number. Therefore we also check if Number(value) is NaN.
-  return isNaN(parseFloat(value as any)) || isNaN(Number(value)) ? fallbackValue : Number(value);
+  return !isNaN(parseFloat(value as any)) && !isNaN(Number(value));
 }
--- a/src/lib/table/table-data-source.ts
+++ b/src/lib/table/table-data-source.ts
@@ -15,6 +15,7 @@ import {combineLatest} from 'rxjs/operators/combineLatest';
 import {map} from 'rxjs/operators/map';
 import {startWith} from 'rxjs/operators/startWith';
 import {empty} from 'rxjs/observable/empty';
+import {_isNumberValue} from '@angular/cdk/coercion';
 
 /**
  * Data source that accepts a client-side data array and includes native support of filtering,
@@ -88,7 +89,8 @@ export class MatTableDataSource<T> implements DataSource<T> {
   private _paginator: MatPaginator|null;
 
   /**
-   * Data accessor function that is used for accessing data properties for sorting.
+   * Data accessor function that is used for accessing data properties for sorting through
+   * the default sortData function.
    * This default function assumes that the sort header IDs (which defaults to the column name)
    * matches the data's properties (e.g. column Xyz represents data['Xyz']).
    * May be set to a custom function for different behavior.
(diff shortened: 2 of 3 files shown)


mcalmus:
This line:
 +        comparatorResult = valueA < valueB ? -1 : 1;
Assumes that if valueA isn't smaller than valueB it's bigger. If valueA and valueB are the same you could end up with weird results.

andrewseguin:
Good call - forgot about the case of them being equal and remaining at 0. Thanks

src/cdk/coercion/number-property.ts
```diff
@@ -10,8 +10,13 @@
 export function coerceNumberProperty(value: any): number;
 export function coerceNumberProperty<D>(value: any, fallback: D): number | D;
 export function coerceNumberProperty(value: any, fallbackValue = 0) {
+  return isNumberValue(value) ? Number(value) : fallbackValue;
+}
+
+/** Whether the provided value is considered a number. */
+export function isNumberValue(value: any): boolean {
```
jelbourn: I'm not sure this should be exposed as a public API under coercion. Mark this as internal?

andrewseguin: Marked as `docs-private`

src/cdk/coercion/number-property.ts
```diff
@@ -10,8 +10,16 @@
...
 export function coerceNumberProperty<D>(value: any, fallback: D): number | D;
 export function coerceNumberProperty(value: any, fallbackValue = 0) {
+  return isNumberValue(value) ? Number(value) : fallbackValue;
+}
+
+/**
+ * Whether the provided value is considered a number.
+ * @docs-private
+ */
+export function isNumberValue(value: any): boolean {
```
jelbourn: It would be better to prefix with an `_` to make it completely clear that this is not a public API (rather than being public and undocumented)

src/lib/table/table-data-source.ts
```diff
@@ -97,22 +99,51 @@ export class MatTableDataSource<T> implements DataSource<T> {
    */
   sortingDataAccessor: ((data: T, sortHeaderId: string) => string|number) =
       (data: T, sortHeaderId: string): string|number => {
-    const value: any = data[sortHeaderId];
+    let value: any = data[sortHeaderId];
```
jelbourn: `const`

src/lib/table/table-data-source.ts
```diff
@@ -97,22 +99,51 @@ export class MatTableDataSource<T> implements DataSource<T> {
...
+    return isNumberValue(value) ? Number(value) : value;
+  }
 
-    // If the value is a string and only whitespace, return the value.
-    // Otherwise +value will convert it to 0.
-    if (typeof value === 'string' && !value.trim()) {
-      return value;
-    }
+  /**
+   * Returns a sorted array of the provided data according to the state of the MatSort. Called
```
jelbourn: "Returns a sorted array..." -> "Gets a sorted copy of the data array based on the state of the MatSort"

src/lib/table/table-data-source.ts
```diff
@@ -97,22 +99,51 @@ export class MatTableDataSource<T> implements DataSource<T> {
...
+   * after changes are made to the filtered data or when sort changes are emitted from MatSort.
+   * By default, the function retrieves the active sort and its direction and compares data
+   * by retrieving data using the sortingDataAccessor. May be overridden for a custom implementation
+   * of data ordering.
+   * @param data The array of data that should be sorted.
+   * @param sort The connected MatSort that holds the current sort state.
+   */
+  sortPredicate: ((data: T[], sort: MatSort) => T[]) = (data: T[], sort: MatSort): T[] => {
```
jelbourn: A "predicate" is a function that returns `true` or `false`, so this isn't quite the right name.

jelbourn: This new API would really make this a minor release change

src/lib/table/table-data-source.ts
```diff
@@ -97,22 +99,51 @@ export class MatTableDataSource<T> implements DataSource<T> {
...
+      let comparatorResult = 0;
+      if (valueA && valueB) {
+        // Check if one value is greater than the other; if equal, comparatorResult should remain 0.
+        if (valueA > valueB) {
+          comparatorResult = 1;
+        } else if (valueA < valueB) {
+          comparatorResult = -1;
+        }
+      } else if (valueA) {
+        comparatorResult = 1;
```
jelbourn: Add comment that explains how you treat null/undefined for sorting (that they come first)

src/lib/table/table-data-source.ts
andrewseguin: Any suggestions? Just `sort`?

jelbourn: Sgtm

andrewseguin: Ugh duh, `sort` is an input for the `MatSort`. Hmm, what about...`sortData`? Kinda fits with the function using `sortDataAccessor`

andrewseguin:
Ready for re-review

PR #8719, opened 2017-11-29
perf(ripple): use passive event listeners

crisbeto:
Switches all of the ripple-related event listeners to be passive for improved performance and to prevent Chrome from logging warnings for almost every Material component. Also converts a few methods to arrow functions so we don't have to `bind` them explicitly.

### DIFF
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -5,9 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-
 import {ElementRef, NgZone} from '@angular/core';
-import {Platform} from '@angular/cdk/platform';
+import {Platform, supportsPassiveEventListeners} from '@angular/cdk/platform';
 import {RippleRef, RippleState} from './ripple-ref';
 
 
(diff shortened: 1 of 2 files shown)


src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -170,25 +169,27 @@ export class RippleRenderer {
 
   /** Sets the trigger element and registers the mouse events. */
   setTriggerElement(element: HTMLElement | null) {
+    const eventOptions = supportsPassiveEventListeners() ? ({passive: true} as any) : false;
```
DevVersion: Wouldn't it be enough to check that at `RippleRender` instantiation (e.g. constructor, or property) ?

crisbeto: I suppose so, but the check isn't particularly expensive since we cache the result.

DevVersion: I'd prefer having it as a property since it's not necessary to have this in the `setTriggerElement` method.

crisbeto: Done.

dushkostanoeski:
Why this wasn't merged for the 5.0.1 update? Will it get merged for the next patch?

PR #8721, opened 2017-11-29
fix(typography): handle inherit being set as a typography value

### LINKED ISSUE #8700: @mixin mat-typography-level-to-styles
Bug or proposal

https://github.com/angular/material2/blob/3b41c0c23e10fc74d9f687427f80c46f5da419c9/src/lib/core/typography/_typography-utils.scss#L32
```
@mixin mat-typography-level-to-styles($config, $level) {
  $font-size: mat-font-size($config, $level);
  $font-weight: mat-font-weight($config, $level);
  $line-height: mat-line-height($config, $level);
  $font-family: mat-font-family($config, $level);

  // Use the shorthand `font` to represent a typography level, because it's the shortest. Notes that
  // we need to use interpolation for `font-size/line-height` in order to prevent SASS from dividing
  // the two values.
  font: $font-weight #{$font-size}/#{$line-height} $font-family;
}
```

```
/ We rewrite the Angular Material mixin because they use the shorthand `font`
// which can't use `inherit` for `font-family`
@mixin mat-typography-level-to-styles($config, $level) {
	$font-size: mat-font-size($config, $level);
	$font-weight: mat-font-weight($config, $level);
	$line-height: mat-line-height($config, $level);
	$font-family: mat-font-family($config, $level);

	font: {
		family: $font-family;
		weight: $font-weight;
		size: $font-size;
	}
	line-height: $line-height;
}
```

The shorthand `font` can't use `inherit` for `font-family`

```
// We use 'font-family' on body only and rewrite 'font-family' used Angular Material in components
// It's simple method for translating App and styling whole App
$app-typography: mat-typography-config(
  	$font-family: 'inherit'
);

// Extend Material2 Typography
$font-families: (
	en: (font-family: 'Roboto, "Helvetica Neue", sans-serif'),
	zh: (font-family: 'Roboto, "Noto Sans SC", "Helvetica Neue", sans-serif')
);
$app-typography: map_merge($app-typography, $font-families);
```

crisbeto:
Fixes the `mat-typography-level-to-styles` mixin generating an invalid style declaration if any of its values are set to `inherit`.

Fixes #8700.

### DIFF
--- a/src/lib/core/typography/_typography-utils.scss
+++ b/src/lib/core/typography/_typography-utils.scss
@@ -36,8 +36,22 @@
   $line-height: mat-line-height($config, $level);
   $font-family: mat-font-family($config, $level);
 
-  // Use the shorthand `font` to represent a typography level, because it's the shortest. Notes that
-  // we need to use interpolation for `font-size/line-height` in order to prevent SASS from dividing
-  // the two values.
-  font: $font-weight #{$font-size}/#{$line-height} $font-family;
+  // If any of the values are set to `inherit`, we can't use the shorthand
+  // so we fall back to passing in the individual properties.
+  @if ($font-size == inherit or
+       $font-weight == inherit or
+       $line-height == inherit or
+       $font-family == inherit) {
+
+    font-size: $font-size;
+    font-weight: $font-weight;
+    line-height: $line-height;
+    font-family: $font-family;
+  }
+  @else {
+    // Otherwise use the shorthand `font` to represent a typography level, because it's the the
+    // least amount of bytes. Note that we need to use interpolation for `font-size/line-height`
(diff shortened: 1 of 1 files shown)


src/lib/core/typography/_typography-utils.scss
```diff
@@ -36,8 +36,22 @@
...
+       $font-weight == inherit or
+       $line-height == inherit or
+       $font-family == inherit) {
+
+    font-size: $font-size;
+    font-weight: $font-weight;
+    line-height: $line-height;
+    font-family: $font-family;
+  }
+  @else {
```
crisbeto: Note: the funky formatting here is due to Stylelint complaining about having to have a newline after the brace.

PR #8726, opened 2017-11-29
feat(dialog): allow for single dialog scroll strategy to be overwritten

### LINKED ISSUE #8706: [Feature request] MatDialog config option to disable scroll-blocking
Feature request

A config option to disable the scroll-blocking behavior of a dialog
(so that the underlaying page could be used as without the dialog)

There's an option to disable to backdrop `hasBackdrop` but the underlaying site still isn't scrollable (but clickable though)

With this feature it would be possible to "pin" a dialog like gmails compose-email window.

crisbeto:
Currently we have the `MAT_DIALOG_SCROLL_STRATEGY` token which allows consumers to specify the scroll strategy for all dialogs, however there's no way do so for a single dialog. These changes add an extra property to the dialog config that allow for the scroll strategy to be set.

Fixes #8706.

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -8,6 +8,7 @@
 
 import {ViewContainerRef} from '@angular/core';
 import {Direction} from '@angular/cdk/bidi';
+import {ScrollStrategy} from '@angular/cdk/overlay';
 
 /** Valid ARIA roles for a dialog element. */
 export type DialogRole = 'dialog' | 'alertdialog';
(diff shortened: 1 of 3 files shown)


jpduckwo:
@crisbeto You mention that MAT_DIALOG_SCROLL_STRATEGY can be set by the consumer. I've been trying to figure out how to do it. Can you shed any light on how to override it with a different scroll strategy... even if it is global.

crisbeto:
You need to do something like this in your module @jpduckwo:

```ts
@NgModule({
...
providers: [
  {
    provide: MAT_DIALOG_SCROLL_STRATEGY,
    deps: [Overlay],
    useFactory: (overlay: Overlay) => () => overlay.scrollStrategies.close()
  }
]
})
```

PR #8740, opened 2017-11-30
fix(select): change event emitted before data binding is updated

### LINKED ISSUE #8739: MatSelect - Selected value is outdated when "change" event is fired
At the moment the "change" event is fired, the property binding [(value)] is still outdated.

All bindings to the new selected value should be updated too.

Bindings to the new selected value are outdated at the moment the "change" event is fired.

http://plnkr.co/edit/EDVS5r0FAcwY7Rfhr0De?p=preview

If the bindings to value property is updated at the moment the change event is fired, it would be not needed to workaround some programming strategies (by creating aditional functions/methods).

5.0.0-rc.2

crisbeto:
Fixes the select's `change` event emitting before the `value` binding has been updated, causing consumers that check the data-bound output to get a stale value.

Fixes #8739.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -2321,6 +2321,29 @@ describe('MatSelect', () => {
 
       expect(document.activeElement).toBe(select, 'Expected trigger to be focused.');
     }));
+
+    it('should update the data binding before emitting the change event', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


tonyb1008:
Can you advise when this fix will be deployed and what angular/material release number it will be in? I am still seeing this issue in version 6.1.0.

PR #8746, opened 2017-11-30
fix(autocomplete): not implementing setDisabledState from ControlValueAccessor

### LINKED ISSUE #8735: Can not disable custom MatFormFieldControl with FormControl with mat-autocomplete
* Potential bug

A component implementing `MatFormFieldControl` with a `mat-autocomplete` can be disabled correctly via a `FormControl`.

It is not possible to have a custom `MatFormFieldControl` implementation with a `mat-autocomplete` and have the input disabled via a `FormControl`. Notice in the StackBlitz application, it is possible to open the autocomplete overlay of the field and select an option even when the field is set to disabled.

If the `FormControl` is removed from the `input` element in `custom-input` and a binding to `[disabled]` is used, the behaviour works as expected. The reason I am using a FormControl here is to automatically apply validators to the input, so they are not needed to be added manually by the consumer of the form field everywhere it is used.

It does however work correctly when adding `matInput` to the `input` element inside `custom-input`. Is this the intended behaviour? The [CDK Docs here](https://material.angular.io/guide/creating-a-custom-form-field-control) do not include it.

https://stackblitz.com/edit/custom-mat-form-field-with-auto-complete

To be able to disable the input when an autocomplete is attached, or be clearer in the docs when `matInput` is needed still.

Latest from NPM (see StackBlitz)

crisbeto:
Fixes plain inputs (without a `matInput`) that have a `formControl` and an autocomplete attached not being disabled via `FormControl.disable`. The issue was due to the `MatAutocompleteTrigger` not implementing the `setDisableState`. The current implementation is in line with the approach in the `DefaultValueAccessor`.

Fixes #8735.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -267,6 +267,14 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
     this._onTouched = fn;
   }
 
+  /**
+   * Disables the input. Implemented as a part of `ControlValueAccessor`.
+   * @param isDisabled Whether the component should be disabled.
+   */
+  setDisabledState(isDisabled: boolean) {
+    this._element.nativeElement.disabled = isDisabled;
+  }
+
   _handleKeydown(event: KeyboardEvent): void {
     const keyCode = event.keyCode;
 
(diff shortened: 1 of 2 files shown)


jelbourn:
Also needs rebase

PR #8749, opened 2017-11-30
feat(tree): add tree demo page

tinayuangao:


src/demo-app/tree/tree-demo.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.nestedDataSource = new JsonNestedDataSource(database);
+  }
+
+  getLevel = (node: JsonFlatNode) => { return node.level };
+
+  isExpandable = (node: JsonFlatNode) => { return node.expandable; }
+
+  getChildren = (node: JsonNode) => { return ofObservable(node.children); }
+
+  hasChild = (_: number, _nodeData: JsonFlatNode) => { return _nodeData.expandable; }
```
benb7760: @tinayuangao I noticed this a week or so ago while trialling the tree; users have to define two functions that achieve the exact same result but with different signatures; `isExpandable`, used by the `when`, and `hasChild`, used by the TreeControl. Would it be sensible to change this?

src/demo-app/tree/tree-demo.html
```diff
@@ -0,0 +1,47 @@
+<div class="demo-tree-container">
+  <mat-card>
+    <mat-card-header>Flattened tree</mat-card-header>
+
+    <mat-tree [dataSource]="dataSource" [treeControl]="treeControl">-->
```
jelbourn: Hanging `-->` ?

src/demo-app/tree/tree-demo.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  // Flat tree data source
+  dataSource: FlatDataSource;
+
+  // Nested tree data source
+  nestedDataSource: JsonNestedDataSource;
+
+  constructor(database: JsonDatabase) {
+    // For flat tree
+    this.treeControl = new FlatTreeControl<JsonFlatNode>(this.getLevel, this.isExpandable);
```
jelbourn: Would it make sense to use a more real-world example, such as collection of directories?

tinayuangao:
Comments addressed. Please take a look. Thanks!

PR #8766, opened 2017-12-01
fix(dialog): improved type safety in dialog ref result

### LINKED ISSUE #8760: Type safe dialog afterClosed()
feature request

`MatDialogRef.afterClosed()` return `Observable` of a type other than` any`.

`MatDialogRef.afterClosed()` returns `Observable<any>`.

type safety 

all

I tried:

    export interface DialogComponent<R> { }

    export interface TypedMatDialogRef<T extends DialogComponent<R>, R> extends MatDialogRef<T> {
      afterClosed(): Observable<R>;
    }

    export interface TypedMatDialog extends MatDialog {
      open<T extends DialogComponent<R>, R, D = any>(componentOrTemplateRef: ComponentType<T>, config?: MatDialogConfig<D>): TypedMatDialogRef<T, R>;
    }

    @Component({})
    export class MyDialogComponent implements DialogComponent<string> { }

    (this.dialog as TypedMatDialog).open(MyDialogComponent)
      .afterClosed()
      .subscribe(string => {/* string is showing as {}*/});

Not sure why this doesn't work.

crisbeto:
Adds an extra generic param to the `MatDialogRef` that allows consumers to type the result that is passed to `close`, as well as the value in the `beforeClosed` and `afterClosed` observables.

Fixes #8760.

### DIFF
--- a/src/lib/dialog/dialog-ref.ts
+++ b/src/lib/dialog/dialog-ref.ts
@@ -23,7 +23,7 @@ let uniqueId = 0;
 /**
  * Reference to a dialog opened via the MatDialog service.
  */
-export class MatDialogRef<T> {
+export class MatDialogRef<T, R = any> {
   /** The instance of component opened into the dialog. */
   componentInstance: T;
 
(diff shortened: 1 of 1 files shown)


src/lib/dialog/dialog-ref.ts
```diff
@@ -101,14 +101,14 @@ export class MatDialogRef<T> {
   /**
    * Gets an observable that is notified when the dialog is finished closing.
    */
-  afterClosed(): Observable<any> {
+  afterClosed(): Observable<R> {
```
josephperrott: Should the return type here be `Observable<R | undefined>` since the `dialogResult` parameter in the `close` method is optional?

crisbeto: Good point, I'm not sure why TS didn't complain about it.

PR #8767, opened 2017-12-01
fix(list): fix role for list-item

tinayuangao:
The role for `mat-list-item` should not be set. User of `mat-list-item` should set the role.


src/lib/list/list.ts
```diff
@@ -134,6 +134,7 @@ export class MatListItem extends _MatListItemMixinBase implements AfterContentIn
     CanDisableRipple {
   private _lineSetter: MatLineSetter;
   private _isNavList: boolean = false;
+  role: string | null = null;
```
crisbeto: Shouldn't this one be underscored?

tinayuangao: Done. Thanks for review

PR #8770, opened 2017-12-02
docs(table): updated docs for the table

andrewseguin:
To be added later: expanding rows (using when predicate), customizing column widths, adding/removing data, dynamic columns (add/remove/move), trackBy

(this was opened before at https://github.com/angular/material2/pull/8162 and was accidently closed, whoops - all feedback from there has been incorporated)

### DIFF
--- a/src/lib/table/table.md
+++ b/src/lib/table/table.md
@@ -1,56 +1,210 @@
-The `mat-table` provides a Material Design styled data-table that can be used to display rows of
-data.
+The `mat-table` provides a Material Design styled data-table that can be used to display rows of 
+data. 
 
-This table builds on the foundation of the CDK data-table and uses a similar interface for its
+The table's template consists of two parts: column definitions and row definitions. Each column 
+definition contains templates for that column's header and content cells. Each row definition 
+captures the columns used for that row and any bindings applied to the row element.
+
+A `DataSource` provides data to the table by emitting an `Observable` stream of the items to be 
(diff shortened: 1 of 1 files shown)


DavidJohnWilliams:
Hi, @andrewseguin.

Regarding your initial post:

> To be added later: expanding rows (using when predicate), customizing column widths, adding/removing data, dynamic columns (add/remove/move), trackBy

Are you saying those things are all possible with the current table and just not documented? There are a lot of questions surrounding these issues on Gitter (and here).

I'd be happy to knock up some docs and stackblitz samples for these things if I can get a pointer about what you guys consider the correct way.

Currently I am:
1. Adding/removing columns using `matHeaderRowDef` (can be updated dynamically).
2. Customizing column widths using angular/flex-layout (can provide examples using css).
3. Adding removing data by updating `MatTableDataSource.data`. Although I note that modifying the array does not trigger the BehaviourSubject. At the moment I have opted to set the filter to the current filter which triggers the `updateChangeSubscription`. I doubt this it the best way, I think it could be worth adding a `refresh()` function or similar mechanism to the table source?
4. Track by can just use CDK `[trackBy]`. There are already some docs about this but I can write an example.
5. Not really sure what you mean about row expansion?

Let me know if it would be helpful for me to start writing these docs or if there are already plans?

andrewseguin:
@DavidJohnWilliams These are all possible and just not documented. You're welcome to submit changes/fixes to the docs, though I'd base it off this PR since it moves some things around: https://github.com/angular/material2/pull/9489

PR #8777, opened 2017-12-02
fix(select): complete state change event

amcdnl:
The FormField component binds to the select's `stateChanges` stream but the stream is never completed. This PR completes that stream just like the Input component does.

Ref: https://github.com/angular/material2/blob/master/src/lib/form-field/form-field.ts#L188


amcdnl:
Done :)

PR #8779, opened 2017-12-02
fix(expansion-panel): expand animation jumping

crisbeto:
Fixes the expansion panel jumping towards the end of the animation due to the overflow going from `visible` to `hidden`.

Here's what it looks like now for reference: 
a

### DIFF
--- a/src/lib/expansion/expansion-panel.html
+++ b/src/lib/expansion/expansion-panel.html
@@ -1,10 +1,13 @@
 <ng-content select="mat-expansion-panel-header"></ng-content>
 <div class="mat-expansion-panel-content"
      role="region"
-     [class.mat-expanded]="expanded"
      [@bodyExpansion]="_getExpandedState()"
+     (@bodyExpansion.done)="_bodyAnimation($event)"
+     (@bodyExpansion.start)="_bodyAnimation($event)"
+     [class.mat-expanded]="expanded"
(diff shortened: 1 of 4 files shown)


src/lib/expansion/expansion-panel.html
```diff
@@ -1,6 +1,10 @@
 <ng-content select="mat-expansion-panel-header"></ng-content>
-<div [class.mat-expanded]="expanded" class="mat-expansion-panel-content"
-     [@bodyExpansion]="_getExpandedState()" [id]="id">
+<div class="mat-expansion-panel-content"
+     [@bodyExpansion]="_getExpandedState()"
+     (@bodyExpansion.done)="_bodyAnimation($event)"
+     (@bodyExpansion.start)="_bodyAnimation($event)"
+     [id]="id"
```
josephperrott: Why are we binding the `id` here?

crisbeto: That was there before my changes so I'm not sure. I _think_ it's for a11y so we can reference it somewhere else.

josephperrott: I think you're right.  I think it is for `aria-owns`

josephperrott:
@crisbeto please rebase

crisbeto:
Rebased.

PR #8782, opened 2017-12-03
chore: filter alias exports in docs

### LINKED ISSUE #8693: Duplicated content in CDK Observers API docs
Bug

The documentation for CDK Observers API is duplicated.
Demo: https://material.angular.io/cdk/observers/api

DevVersion:
Let's consider this example:

```ts
  export {ObserveContent} from './X';

  /** @deprecated */
  export {ObserveContent as CdkObserveContent} from './X';
```
This is a common pattern inside of Angular Material, but causes Dgeni to generate documents for both exports.

The second document is identical to the original Dgeni document and doesn't even show the aliased name. It would be useful to have a property that indicates that the given doc is an alias and maybe also refer to the original document.

This commit introduces a processor that removes the duplicate documents from the docs.

Related: https://github.com/angular/dgeni-packages/issues/248 (already chatted with Pete about it).

Fixes #8693

### DIFF
--- a/tools/dgeni/processors/filter-export-aliases.ts
+++ b/tools/dgeni/processors/filter-export-aliases.ts
@@ -1,0 +1,48 @@
+import {DocCollection, Processor} from 'dgeni';
+import {ExportDoc} from 'dgeni-packages/typescript/api-doc-types/ExportDoc';
+import {ModuleDoc} from 'dgeni-packages/typescript/api-doc-types/ModuleDoc';
+
+/**
+ * Processor to filter out Dgeni documents that are export aliases. Keeping them means
+ * that a given document shows up multiple times in the docs.
+ *
+ * ```ts
+ *   export {ObserveContent} from './X';
+ *   export {ObserveContent as CdkObserveContent} from './X';
+ * ```
+ *
+ * This is a common pattern inside of Angular Material, but causes Dgeni to generate
+ * documents for both exports. The second document is identical to the original document and
+ * doesn't even show the aliased name.
+ *
+ * See: https://github.com/angular/dgeni-packages/issues/248
+ */
+export class FilterExportAliases implements Processor {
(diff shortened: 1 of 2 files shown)


tools/dgeni/processors/filter-export-aliases.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (!doc.moduleDoc) {
+      return true;
+    }
+
+    const moduleDoc = doc.moduleDoc as ModuleDoc;
+    const duplicateDocs = moduleDoc.exports.filter(exportDoc => exportDoc.id === doc.id);
+
+    // Remove the current export document if there are multiple Dgeni export documents with the
+    // same document id. If there are multiple docs with the same id, we can assume that this doc
+    // is an alias export.
```
jelbourn: Is there any way to more specifically target things marked as `@deprecated`? It's possible we may want to re-export something not as a deprecation, but as an alternate name (e.g. re-exporting something from cdk with "mat").

DevVersion: I see. Unfortunately it's not possible to detect if the re-export was marked as deprecated.

```ts
/** @deprecated */
export {XX as PP} from './test';
```

Dgeni doesn't provide access to the export statement/symbol. In theory it would be possible to walk through imports file by file using the AST, but this more turns into it's own docs generation tool then.

I assume even Dgeni in general (with https://github.com/angular/dgeni-packages/issues/248) won't be able to differentiate between deprecated re-exports or normal re-exports.

jelbourn: What if we formatted it like this?
```ts
export {XX as /** @deprecated */ PP} from './test';
```

DevVersion: Just tried that, but that information is being lost as well.

PR #8811, opened 2017-12-05
chore: fix demo app compilation error

crisbeto:
Fixes a compilation error in the demo app when running on a fresh install.


src/cdk/table/table.spec.ts
```diff
@@ -639,7 +639,7 @@ class FakeDataSource extends DataSource<TestData> {
   connect(collectionViewer: CollectionViewer) {
     this.isConnected = true;
     const streams = [this._dataChange, collectionViewer.viewChange];
-    return combineLatest<TestData[]>(streams).pipe(map(([data]) => data));
+    return combineLatest<TestData[][]>(streams).pipe(map(([data]) => data));
```
mmalerba: combineLatest(streams, (data, _) => data);

PR #8820, opened 2017-12-05
fix(expansion): use `display` rather than `visibility` to hide closed expansion panels

### LINKED ISSUE #8754: Nested Expansion panels keep overflow visible
Bug

Nested Expansion panels should collapse their content, and set the overflow to hidden.

Nested expansion panels collapse their content, but overflow is kept as visible, showing a scrollbar and nasty empty space.

Plunker starter (using on `@master`): http://plnkr.co/edit/uVmKf7NTB9ZVUSv2IyGR?p=preview<br/>

Angular 5.0.4
Material 5.0.0-rc.2
MacOS
TS 2.4.2
Browser: Seen in Safari, Chrome and Firefox (not tested any others)

This only seems to be actually noticeable when used inside tabs. Checking the dev tools, I can see the `.mat-expanded` class from the parent causes the child to keep it's overflow visible

josephperrott:
Fixes #8754

### DIFF
--- a/src/lib/expansion/expansion-animations.ts
+++ b/src/lib/expansion/expansion-animations.ts
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+
 import {
   animate,
   animateChild,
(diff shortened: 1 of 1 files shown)


meehol:
Hey guys, into which version can we expect this to be pulled?

josephperrott:
Closing as this was addressed by https://github.com/angular/material2/pull/10138

PR #8822, opened 2017-12-05
fix(table): use combineLatest as mapper for FakeDataSource

josephperrott:


crisbeto:
@josephperrott I have https://github.com/angular/material2/pull/8811 that seems to fix the same issue.

PR #8824, opened 2017-12-05
chore: lock down rxjs version

crisbeto:
Locks down the version of rxjs to 5.5.2 in order to avoid runtime errors that are being thrown at the moment. See https://github.com/angular/angular/issues/20752.


princemaple:
5.5.4 fixed it.

Issue #5269, opened 2017-06-21
[Tabs] Hidden tabs don't render expansion panels correctly


AlanCrevon:
#### Bug, feature request, or proposal:
Bug.

#### What is the expected behavior?
A `<md-expansion-panel [expanded]="false">` should not be expanded until explicitally opened.
When inside a hidden tab,  `.mat-expansion-panel-content` should have `height : 0px` 

#### What is the current behavior?
 Panel is collapsed but panetl's content is visible. Problem occurs only if the accordion is in a hidden tab of a tab-group.

#### What are the steps to reproduce?

http://plnkr.co/edit/q6kXPH?p=preview

```<md-tab-group>
  <md-tab label="Working">
    <md-accordion>
        <md-expansion-panel [expanded]="false">
        <md-expansion-panel-header>
          You should see me
        </md-expansion-panel-header>
        <p>Yous should not see me</p>
      </md-expansion-panel>
    </md-accordion>
  </md-tab>
  <md-tab label="Not working">
    <md-accordion>
        <md-expansion-panel [expanded]="false">
        <md-expansion-panel-header>
          You should see me
        </md-expansion-panel-header>
        <p>Yous should not see me</p>
      </md-expansion-panel>
    </md-accordion>
  </md-tab>
</md-tab-group>
```

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Material2 master
Angular 4.2.3

julianobrasil:
It seems to be working correctly here: http://plnkr.co/edit/q6kXPH?p=preview

AlanCrevon:
correct. It's also working in another page of mine. Currently trying to isolate the pb.

AlanCrevon:
Ok, i edit the issue : the problem occurs only if the element is in a hidden tab. Thanks @julianobrasil

julianobrasil:
OK. Updated the plunker to reproduce it: http://plnkr.co/edit/q6kXPH?p=preview

Put it in the issue.

emoralesb05:
This actually happens to anything that contains an animation when set inside a hidden md-tab.

mzernetsch:
This seems to be caused by angular 4.2 changes.
I use Teradata Covalent expansion panels in my app which have the same problem since the new angular version: https://gitter.im/Teradata/covalent?at=593e58f06549436c7d3c8360

julianobrasil:
[...]

PR #8825, opened 2017-12-05
fix(tabs): hidden tabs don't render expansion panels correctly

tiaguinho:
Fix issue https://github.com/angular/material2/issues/5269

### DIFF
--- a/src/demo-app/tabs/tabs-demo.html
+++ b/src/demo-app/tabs/tabs-demo.html
@@ -288,3 +288,41 @@
     </div>
   </mat-tab>
 </mat-tab-group>
+
+<h1>Tabs with expansion panel</h1>
+<mat-tab-group>
+  <mat-tab isActive="true" label="">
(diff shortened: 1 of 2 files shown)


tiaguinho:
I just add the test on the demo app.

josephperrott:
In looking closer at this change, we actually can't do this change as it is a major change to the workings of material tabs.  This change immediately attaches all of the tab bodies to the DOM, which places more items in the DOM that are needed.

tiaguinho:
When I made the change, I see this problem too. This probably will reduce tabs performance because all the bodies are attached to the DOM.
But fixing the bug isn't better than have performance with a bug? 
This can't be a workaround until we find a better solution?

josephperrott:
We determined that it is more valuable to not change the functionality back and forth between the two concepts.  Also in the near future we should be able to land #6832 which also could be used as a workaround.

tiaguinho:
Ok.
Thank you.

PR #8840, opened 2017-12-06
docs(cdk/layout): typo fix in MediaMatcher bloc

johanchouquet:


johanchouquet:
I already had a signed CLA. I updated my data in the CLA form.

johanchouquet:
It seems to me that I already had a CLA. I updated it.

PR #8841, opened 2017-12-06
chore: hide internal properties from selection-list api doc

DevVersion:
* Recently with the NgModel support for the selection list, the `onTouched` property has been added. This one needs to be prefixed with an underscore to be hidden in the docs.
* Since the initial implementation of the selection-list the `selectionList` variable has been initialized publically in the constructor of the `MatListOption` and therefore shows up in the docs. This property has been made private now.

### DIFF
--- a/src/lib/list/selection-list.ts
+++ b/src/lib/list/selection-list.ts
@@ -130,7 +130,7 @@ export class MatListOption extends _MatListOptionMixinBase
 
   /** Whether the option is disabled. */
   @Input()
-  get disabled() { return (this.selectionList && this.selectionList.disabled) || this._disabled; }
+  get disabled() { return this._disabled || (this.selectionList && this.selectionList.disabled); }
   set disabled(value: any) {
     const newValue = coerceBooleanProperty(value);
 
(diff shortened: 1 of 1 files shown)


src/lib/list/selection-list.ts
```diff
@@ -158,7 +160,7 @@ export class MatListOption extends _MatListOptionMixinBase
   constructor(private _element: ElementRef,
               private _changeDetector: ChangeDetectorRef,
               @Optional() @Inject(forwardRef(() => MatSelectionList))
-              public selectionList: MatSelectionList) {
+              private _selectionList: MatSelectionList) {
```
crisbeto: I doubt that many people are depending on this property in particular, but this is technically a breaking change.

DevVersion: Good point. If really necessary we can deprecate it

jelbourn: Might as well leave it, doesn't really hurt to leave it as public.

DevVersion: Hm, but it doesn't really make sense from an API perspective. Also it shows up in the docs which is kind of unnecessary. I don't really mind keeping that property as it is. Just let me know.

jelbourn: Can just make it docs-private

DevVersion: Done.

src/lib/list/selection-list.ts
```diff
@@ -126,7 +126,9 @@ export class MatListOption extends _MatListOptionMixinBase
 
   /** Whether the option is disabled. */
   @Input()
-  get disabled() { return (this.selectionList && this.selectionList.disabled) || this._disabled; }
+  get disabled() {
+    return this._disabled || this.selectionList && this.selectionList.disabled;
```
crisbeto: Add the parentheses around the second condition here? Otherwise it's a little ambiguous.

PR #8844, opened 2017-12-06
feat(cdk-dialog): create CdkDialog in material-experimental

josephperrott:


src/material-experimental/dialog/dialog-config.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/** Possible overrides for a dialog's position. */
+export interface DialogPosition {
+  top?: string;
+  bottom?: string;
+  left?: string;
+  right?: string;
+}
+
+export class DialogConfig<D = any> {
```
crisbeto: Should all the symbols be prefixed with `Cdk`? E.g. `CdkDialogConfig`.

src/material-experimental/dialog/dialog-container.ts
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Internal component that wraps user-provided dialog content.
+ * @docs-private
+ */
+@Component({
+  moduleId: module.id,
+  selector: 'cdk-dialog-container',
+  template: `<ng-template cdkPortalOutlet></ng-template>`,
+  styles: [`
```
crisbeto: Move to a separate file?

src/material-experimental/dialog/dialog-container.ts
```diff
@@ -0,0 +1,217 @@
...
+    cdk-dialog-container {
+      background: white;
+      border-radius: 5px;
+      display: block;
+      padding: 10px;
+    }
+  `],
+  encapsulation: ViewEncapsulation.None,
+  preserveWhitespaces: false,
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
crisbeto: We had some G3 sync issues the last time we tried enabling `OnPush` for the dialog container.

src/material-experimental/dialog/dialog-container.ts
```diff
@@ -0,0 +1,217 @@
...
+  animations: [
+    trigger('dialog', [
```
crisbeto: IMO CDK components shouldn't include animations.

src/material-experimental/dialog/dialog-container.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  // tslint:disable-next-line:no-host-decorator-in-concrete
+  @HostBinding('attr.aria-describedby')
+  private get ariaDescribedBy() { return this._config ? this._config.ariaDescribedBy : null; }
+
+  // tslint:disable-next-line:no-host-decorator-in-concrete
+  @HostBinding('attr.role') private get role() { return this._config ? this._config.role : null; }
+
+  // tslint:disable-next-line:no-host-decorator-in-concrete
+  @HostBinding('attr.tabindex') private get tabindex() { return -1; }
```
crisbeto: Didn't we have a rule to use the `host` instead of `@HostBinding`?

src/material-experimental/dialog/dialog-container.ts
```diff
@@ -0,0 +1,217 @@
...
+  _config: DialogConfig;
+
+  constructor(
+    private _elementRef: ElementRef,
+    private _focusTrapFactory: FocusTrapFactory,
+    private _changeDetectorRef: ChangeDetectorRef,
+    @Optional() @Inject(DOCUMENT) private _document: any) {
+    super();
+
+    this._afterEnter.subscribe(() => {
```
crisbeto: Rather than having it subscribe to its own observables, it might be better to just do the calls at the appropriate time.

src/material-experimental/dialog/dialog-injectors.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/** Injection token for the Dialog's ScrollStrategy. */
+export const DIALOG_SCROLL_STRATEGY =
+    new InjectionToken<() => ScrollStrategy>('DialogScrollStrategy');
+
+/** Injection token for the Dialog's Data. */
+export const DIALOG_DATA = new InjectionToken<any>('DialogData');
+
+/** Injection token for the DialogRef constructor. */
+export const DIALOG_REF = new InjectionToken<DialogRef<CdkDialogContainer>>('DialogRef');
```
crisbeto: The generic param in `DialogRef` is the component that we instantiated for the consumer, not `CdkDialogContainer`.

.github/CODEOWNERS
```diff
@@ -74,6 +74,7 @@
 
 # Material experimental package
 /src/material-experimental/**                      @jelbourn
+/src/material-experimental/dialog/**               @jelbourn @josephperrott
```
crisbeto: Also could you add me to this one since I'm also on the Material dialog?

src/material-experimental/dialog/dialog-config.ts
josephperrott: @jelbourn said that he wanted to have `Cdk` prefix only on Directives within the CDK.

src/material-experimental/dialog/dialog-container.ts
josephperrott: Because we would like for the Cdk's Dialog to work as a base for other dialog services, we would like to establish both an example of how to do the animation as well as the structure for callbacks for the dialog animations.

src/material-experimental/dialog/dialog-container.ts
josephperrott: The intention of `CdkDialogContainer` is to be extended so we use `@HostBinding` to allow for it to be inherited.

src/material-experimental/dialog/dialog-container.ts
josephperrott: Personally I find it a bit more readable to see that these actions happen in response to "lifecycle hooks".  But I could see arguements for both.

@jelbourn thoughts?

src/material-experimental/dialog/dialog-ref.ts
```diff
@@ -0,0 +1,137 @@
...
+  keydownEvents(): Observable<KeyboardEvent> {
+    return this._overlayRef.keydownEvents();
+  }
+
+  /**
+   * Updates the dialog's width and height.
+   * @param width New width of the dialog.
+   * @param height New height of the dialog.
+   */
+  updateSize(width: string = 'auto', height: string = 'auto'): this {
```
jelbourn: What about min/max height/width?

src/material-experimental/dialog/dialog-ref.ts
```diff
@@ -0,0 +1,137 @@
...
+    return this._containerInstance._beforeEnter.asObservable();
+  }
+
+  /** Gets an observable that emits when dialog is finished opening. */
+  afterOpen(): Observable<void> {
+    return this._containerInstance._afterEnter.asObservable();
+  }
+
+  /** Gets an observable that emits when dialog begins closing. */
+  beforeClose(): Observable<any> {
```
jelbourn: `any` -> generic type (based on @crisbeto's recent PR that adds a generic for this)

src/material-experimental/dialog/dialog-config.ts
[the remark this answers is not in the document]
josephperrott: I'm not sure I understand what you mean here, can you explain a bit more?

src/material-experimental/dialog/dialog-ref.ts
josephperrott: I think leaving them as `width` and `height` might be best since the config has `maxWidth` and `maxHeight` properties?

src/material-experimental/dialog/dialog-config.ts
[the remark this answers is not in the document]
jelbourn: Nvm, I misread

src/material-experimental/dialog/dialog-ref.ts
jelbourn: You can't update the config once the dialog is open, though. The `OverlayConfig` has `updateSize` that takes an `OverlaySizeConfig` so that each dimension can be updated separately if necessary

src/material-experimental/dialog/dialog-container.ts
jelbourn: I'd agree w/ @crisbeto; it feels more natural to do this work in the callbacks you've set up. 

Also the inside of each of these should really be their own methods like `_applyAutoFocus` or something like that

src/material-experimental/dialog/dialog-ref.ts
josephperrott: Gotcha, I was thinking you wanted to *replace* `width` and `height` with `max-width` and `max-height`.  Now I see what you are wanting to do.

lacolaco:
@josephperrott Why is the package `material-experimental` rather than `cdk-experimental`? I'm confused because that `CdkDialog` is in `material-experimental`...

cc/ @jelbourn

PR #8851, opened 2017-12-06
feat(autocomplete): allow option ripples to be disabled

crisbeto:
For consistency with other components like `MatSelect`, these changes add the ability for all of the option ripples to be disabled by using the `disableRipple` attribute on `MatAutocomplete`.


andrewseguin:
Needs rebase

PR #8864, opened 2017-12-07
perf: remove element if ripples are disabled

### LINKED ISSUE #8854: [MatCheckbox] nested ripple registers events even if disableRipple is set.
When using MatCheckbox, MatRipple and RippleRenderer is instantiated and registers events even if disableRipple is set on the checkbox.

If disableRipple is set, MatRipple and RippleRenderer should not be instantiated, or at the very least the events should not be registered during the initial render.

There are 10 events registered for each checkbox, slowing down the render of the checkbox.

http://embed.plnkr.co/PZ0P3EerE2aeJPR4Kasa/

When clicking on the button to view Material Checkboxes, 30-45% of the time is spent on adding event listeners.

This is visible when using a virtual rendering component to show a large number of items with checkboxes and scrolling, or when rendering the full list as in here. The result is slow rendering or janky scrolling.

All versions are affected. Tested with Angular 2 with Chrome 62 on MacOS.

DevVersion:
Removes the ripple container element if ripples should be disabled for most of the Material components. This gives us a performance boost because the ripple events aren't being added if ripples are disabled.

In theory I would say that using `*ngIf` instead of `[matRippleDisabled]` is more appropriate in most of the components, but there are also cases where the ripple container can't be removed completely and for that, the `[matRippleDisabled]` binding is the right choice.

From a performance perspective, if someone often changes the disabled/disableRipple state, the event listeners will be removed and re-added. This seems to be like a trade-off. A possible solution for that would be to just keep the current implementation and just not register the events if the ripple directive is set to `disabled` initially (and afterwards register events if it turns enabled again).

Fixes #8854.

### DIFF
--- a/src/lib/button/button.html
+++ b/src/lib/button/button.html
@@ -1,7 +1,6 @@
 <span class="mat-button-wrapper"><ng-content></ng-content></span>
-<div matRipple class="mat-button-ripple"
+<div matRipple class="mat-button-ripple" *ngIf="!_isRippleDisabled()"
      [class.mat-button-ripple-round]="_isRoundButton || _isIconButton"
-     [matRippleDisabled]="_isRippleDisabled()"
      [matRippleCentered]="_isIconButton"
      [matRippleTrigger]="_getHostElement()"></div>
 <div class="mat-button-focus-overlay"></div>
--- a/src/lib/button/button.spec.ts
+++ b/src/lib/button/button.spec.ts
@@ -187,11 +187,7 @@ describe('MatButton', () => {
     let fixture: ComponentFixture<TestApp>;
     let testComponent: TestApp;
     let buttonDebugElement: DebugElement;
-    let buttonRippleDebugElement: DebugElement;
-    let buttonRippleInstance: MatRipple;
     let anchorDebugElement: DebugElement;
-    let anchorRippleDebugElement: DebugElement;
-    let anchorRippleInstance: MatRipple;
 
     beforeEach(() => {
       fixture = TestBed.createComponent(TestApp);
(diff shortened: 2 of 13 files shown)


src/lib/button/button.html
```diff
@@ -1,7 +1,6 @@
 <span class="mat-button-wrapper"><ng-content></ng-content></span>
-<div matRipple class="mat-button-ripple"
+<div matRipple class="mat-button-ripple" *ngIf="!_isRippleDisabled()"
```
jelbourn: Will this still register the events on the initial render if `_isRippleDisabled` it `false` on start-up?

DevVersion: No, the events won't be registered. And the directive won't be initialized either.

DevVersion:
@crisbeto Yeah I agree. That's basically also better for performance in the cases where the element's disabled state is toggled pretty often.

I will try to work more on this PR. I think the only change would be to don't register the events if the ripples are disabled initially.

AsafAgranat:
Events overhead aside, there's also a browser memory overhead, when disabled ripple elements remain in the DOM, as explained [here](https://github.com/angular/material2/issues/12647). The short of it is that every ripple elements is painted on it's own DOM layer, bulking up the memory. Therefore ,disabled ripples should at least have their `position: absolute` changed to `position: unset` in order to spare the extra layers.

DevVersion:
@AsafAgranat This is a good point. Can we move that discussion to the new issue you've filed? I've just wanted to link those two threads because they have a related discussion.

(meaning that you just copy what you wrote to the new issue)

AsafAgranat:
@DevVersion, sure. will do.

PR #8872, opened 2017-12-07
fix(drawer): infinite loop when two-way opened binding is toggled mid-animation

### LINKED ISSUE #8869: Infinite loop when toggle sidenav during animation

Toggling the sidenav menu during animation causes a infinite loop that crashes the browser.

It's expected that the toggle never crash the browser.

Toggling sidenav too fast (or during the animation) will cause an infinite loop.

It's easy to reproduce, you can use the example given by Material Docs:
https://stackblitz.com/angular/yjjyldrpyyej

Just click it fast, and you will notice that the browser crashes, and throw infinite error messages on console: 

> SidenavOpenCloseExample.ngfactory.js:20 ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'true'. Current value: 'false'.
>     at viewDebugError (core.umd.js:9845)
>     at expressionChangedAfterItHasBeenCheckedError (core.umd.js:9823)
>     at checkBindingNoChanges (core.umd.js:9992)
>     at checkNoChangesNodeInline (core.umd.js:14044)
>     at checkNoChangesNode (core.umd.js:14016)
>     at debugCheckNoChangesNode (core.umd.js:14845)
>     at debugCheckDirectivesFn (core.umd.js:14747)
>     at Object.eval [as updateDirectives] (SidenavOpenCloseExample.ngfactory.js:118)
>     at Object.debugUpdateDirectives [as updateDirectives] (core.umd.js:14729)
>     at checkNoChangesView (core.umd.js:13854)

Removing the two way binding for sidenav (`[(opened)]="menuOpen"`) workaround the issue, but isn't the fix.

The latest versions available for everything, I've just installed following documentation.

The browser version is "Versão 62.0.3202.62 (Versão oficial) beta 64 bits"

crisbeto:
As of https://github.com/angular/material2/commit/004e0fef2fe5172b6e6ed2c4d368d21dab051355 the drawer supports a two-way binding on the `opened` property, however because the `openedChange` emits at the end of the animation and two-way bindings invoke their setter which then trigger an animation which in turn invokes `openedChange`, there is the potential for an infinite loop if the user changes the `opened` state mid-animation. These changes avoid the issue by turning the `openedChanged` into an async stream and emitting the current value at the end of the animation, rather than determining the value based on the animation state.

Fixes #8869.

### DIFF
--- a/src/lib/sidenav/drawer.ts
+++ b/src/lib/sidenav/drawer.ts
@@ -195,7 +195,9 @@ export class MatDrawer implements AfterContentInit, AfterContentChecked, OnDestr
   _animationState: 'open-instant' | 'open' | 'void' = 'void';
 
   /** Event emitted when the drawer open state is changed. */
-  @Output() openedChange: EventEmitter<boolean> = new EventEmitter<boolean>();
+  @Output() openedChange: EventEmitter<boolean> =
+      // Note this has to be async in order to avoid some issues with two-bindings (see #8872).
+      new EventEmitter<boolean>(/* isAsync */true);
 
   /** Event emitted when the drawer has been opened. */
   @Output('opened')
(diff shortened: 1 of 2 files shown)


src/lib/sidenav/drawer.ts
```diff
@@ -195,7 +195,7 @@ export class MatDrawer implements AfterContentInit, AfterContentChecked, OnDestr
   _animationState: 'open-instant' | 'open' | 'void' = 'void';
 
   /** Event emitted when the drawer open state is changed. */
-  @Output() openedChange: EventEmitter<boolean> = new EventEmitter<boolean>();
+  @Output() openedChange: EventEmitter<boolean> = new EventEmitter<boolean>(true);
```
mmalerba: `...(/* isAsync */ true)` since it kind of looks like initializing open to true

crisbeto:
Caretaker note: this might cause some `fakeAsync` tests to throw `X timer(s) still in queue.` errors.

PR #8874, opened 2017-12-07
fix(directionality): complete dir change observable

crisbeto:
Since it's most likely to use the `Dir.change` observable programmatically, rather than through an event binding, we should complete it in order to ensure that everything is cleaned up.


andrewseguin:
Please check lint errors

crisbeto:
Sorted out the lint warning.

PR #8882, opened 2017-12-08
perf(ripple): do not register events if ripples are disabled initially

### LINKED ISSUE #8854: [MatCheckbox] nested ripple registers events even if disableRipple is set.

DevVersion:
* No longer registers trigger event listeners if the ripples are disabled initially.

Fixes #8854.

### DIFF
--- a/src/lib/core/ripple/ripple.ts
+++ b/src/lib/core/ripple/ripple.ts
@@ -6,21 +6,20 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {Platform} from '@angular/cdk/platform';
 import {
   Directive,
   ElementRef,
-  Input,
   Inject,
-  NgZone,
-  OnChanges,
-  SimpleChanges,
-  OnDestroy,
   InjectionToken,
+  Input,
+  NgZone,
+  OnDestroy,
+  OnInit,
   Optional,
 } from '@angular/core';
-import {Platform} from '@angular/cdk/platform';
-import {RippleConfig, RippleRenderer} from './ripple-renderer';
 import {RippleRef} from './ripple-ref';
+import {RippleConfig, RippleRenderer} from './ripple-renderer';
 
 /** Configurable options for `matRipple`. */
 export interface RippleGlobalOptions {
@@ -50,27 +49,19 @@ export const MAT_RIPPLE_GLOBAL_OPTIONS =
     '[class.mat-ripple-unbounded]': 'unbounded'
   }
 })
-export class MatRipple implements OnChanges, OnDestroy {
+export class MatRipple implements OnInit, OnDestroy {
 
-  /**
-   * The element that triggers the ripple when click events are received. Defaults to the
-   * directive's host element.
-   */
-  // Prevent TS metadata emit from referencing HTMLElement in ripple.js
-  // Otherwise running this code in a Node environment (e.g Universal) will not work.
-  @Input('matRippleTrigger') trigger: HTMLElement|HTMLElement;
+  /** Custom color for all ripples. */
+  @Input('matRippleColor') color: string;
+
+  /** Whether the ripples should be visible outside the component's bounds. */
+  @Input('matRippleUnbounded') unbounded: boolean;
 
   /**
    * Whether the ripple always originates from the center of the host element's bounds, rather
    * than originating from the location of the click event.
    */
   @Input('matRippleCentered') centered: boolean;
-
-  /**
-   * Whether click events will not trigger the ripple. Ripples can be still launched manually
-   * by using the `launch()` method.
-   */
-  @Input('matRippleDisabled') disabled: boolean;
 
   /**
    * If set, the radius in pixels of foreground ripples when fully expanded. If unset, the radius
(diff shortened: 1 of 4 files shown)


src/lib/core/ripple/ripple.ts
```diff
@@ -86,36 +77,71 @@ export class MatRipple implements OnChanges, OnDestroy {
...
+   */
+  @Input('matRippleTrigger')
+  get trigger() { return this._trigger; }
+  set trigger(trigger: HTMLElement | null) {
+    this._trigger = trigger;
+
+    // The events for a trigger element shouldn't be registered if ripples are disabled.
+    // As soon as the ripples are enabled again, the events for the current trigger will be added.
+    if (!this.disabled && this._isInitialized) {
+      this._rippleRenderer.setTriggerElement(trigger);
```
jelbourn: It feels strange to call `setTriggerElement` when you really just want one of the _side-effects_ of `setTriggerElement`. Would it be easier to follow if there was a function to specifically add/remove the event listeners?

DevVersion: What do you mean with side-effects? Maybe the function name is just not good and it should be called `registerTriggerEvents()` or something like that.

jelbourn: Based on the name, the primary purpose of `setTriggerElement` would be the line
```ts
 this._triggerElement = element;
```

The event stuff is incidental to updating the trigger element. It be a bit better as something like
```ts
  /** Sets the trigger element and registers the mouse events. */
  setTriggerElement(element: HTMLElement | null) {
    if (this._triggerElement === element) {
      return;
    }

    this._removeTriggerListeners();
    this._triggerElement = element;
    this._addTriggerListeners();
  }
```

Then other call site that don't need to update to a new element just call `_removeTriggerListeners` and `_addTriggerListeners`

DevVersion: I see what you mean, but the actual event registration should be the primary purpose of this function (or part in the `RippleRenderer`)

In that case I would just call the method `_setupTriggerEvents()` and keep the implementation as it is. The `_triggerElement` variable is just there for the event un-registration, and the real `trigger` element is stored in the `MatRipple` class. Does that work for you?

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
```diff
@@ -229,6 +228,7 @@ export class MatTabLink extends _MatTabLinkMixinBase
     // Manually create a ripple instance that uses the tab link element as trigger element.
     // Notice that the lifecycle hooks for the ripple config won't be called anymore.
     this._tabLinkRipple = new MatRipple(_elementRef, ngZone, platform, globalOptions);
+    this._tabLinkRipple.ngOnInit();
```
crisbeto: It feels weird having to call the Angular lifecycle hook manually. Shouldn't the renderer support attaching to arbitrary elements?

src/lib/core/ripple/ripple.ts
```diff
@@ -86,36 +77,71 @@ export class MatRipple implements OnChanges, OnDestroy {
...
+  ngOnInit() {
+    this._isInitialized = true;
 
-    this._updateRippleRenderer();
+    // The trigger element might have been set before the `ngOnInit` lifecycle hook, but the
+    // events haven't been registered yet. This has been deferred to the `ngOnInit` lifecycle hook,
+    // because the events should be only registered if the ripples aren't disabled initially.
+    // Also if there is no trigger element, that has been specified explicitly through the input,
+    // the default trigger element will be the directive's host element.
+    this.trigger = this._trigger === undefined ? this._hostElement : this.trigger;
```
crisbeto: Doing this just so the getter runs is somewhat side-effect-ey. Why not have a private method that does the attachment both here and insider the setter?

src/lib/core/ripple/ripple.ts
crisbeto: Agreed that at least the naming should change to something involving "events".

src/lib/tabs/tab-nav-bar/tab-nav-bar.ts
DevVersion: Addressed the solution we discussed on Slack. I still kind of feel that the first commit is more clean and it's very uncommon that we want to use the `RippleRenderer` without the `MatRipple` directive.

DevVersion:
@jelbourn Done. Please apply merge ready if you think it's good to go.

PR #8897, opened 2017-12-08
build: expand validateDecoratorsRule to support blacklisting

crisbeto:
Expands the decorator validation rule to support blacklisting properties whose values match a pattern. This allows disabling things like inline styles.

### DIFF
--- a/tools/tslint-rules/validateDecoratorsRule.ts
+++ b/tools/tslint-rules/validateDecoratorsRule.ts
@@ -5,12 +5,14 @@ import * as minimatch from 'minimatch';
 
 /**
  * Rule that enforces certain decorator properties to be defined and to match a pattern.
- * Supports whitelisting via the third argument. E.g.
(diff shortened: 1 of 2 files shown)


tools/tslint-rules/validateDecoratorsRule.ts
```diff
@@ -21,8 +23,16 @@ export class Rule extends Lint.Rules.AbstractRule {
   }
 }
 
-/** Rules that can be used to validate the decorators in a file. */
-type DecoratorRules = {[key: string]: {[key: string]: RegExp}};
+/** Represents a set of required and blacklisted decorator properties. */
+type DecoratorRuleSet = {
+  required:  {[key: string]: RegExp},
+  blacklist:  {[key: string]: RegExp},
```
jelbourn: How about `forbidden` for symmetry with `required`?

PR #8898, opened 2017-12-08
fix(progress-bar): remove data url from progress-bar's css styling

### LINKED ISSUE #8876: Buffer progress bar "ball" is blocked by CSP due to data url svg
[Currently](https://github.com/angular/material2/blob/e650b049af78fb3357388bd8db9901b251bbc467/src/lib/progress-bar/_progress-bar-theme.scss#L53-L81) the buffer bar has inlined SVG's loaded with a data-url, however this is blocked when content security policy are set with the directive `"img-src 'self'"`.

`Refused to load the image 'data:image/svg+xml;charset=UTF-8,%3Csvg%20version%3D[...]' because it violates the following Content Security Policy directive: "img-src 'self'".`

All assets should be able to load with tight CSP configuration

Assets loaded via data url are blocked by CSP

(this cannot be reproduced in Stackblitz etc due to CSP being set by webserver or within `<head>`)

1. Set the CSP to contain `img-src 'self'`
e.g.
`content-security-policy:default-src 'none';object-src 'self';manifest-src 'self';script-src 'self';img-src 'self';style-src 'self' 'unsafe-inline';frame-src 'self';worker-src 'self' blob:`
2. Insert a `<mat-progress-bar mode="buffer"></mat-progress-bar>`
3. See the inlined svg is blocked by the policy
[MDN states that](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src) allowing `data:` urls is insecure:
> <scheme-source>
> A schema such as 'http:' or 'https:'. The colon is required, single quotes shouldn't be used. You can also specify data schemas (not recommended).
> data: Allows data: URIs to be used as a content source. This is insecure; an attacker can also inject arbitrary data: URIs. Use this sparingly and definitely not for scripts.

Since at least Material 5.0.0

This is already marked with a `TODO`, but it is good to have an issue to track the reason why the TODO should be actioned.

Given the content of the svg is simply a circle, it is probable that the same effect can be achieved with a rounded block element without causing any kind of breaking change.

josephperrott:
Fixes #8876

### DIFF
--- a/src/lib/progress-bar/progress-bar.html
+++ b/src/lib/progress-bar/progress-bar.html
@@ -1,5 +1,15 @@
-<!-- The background div is named as such because it appears below the other divs and is not sized based on values. -->
-<div class="mat-progress-bar-background mat-progress-bar-element"></div>
+<!--
+  The background div is named as such because it appears below the other divs and is not sized based
+  on values.
+-->
+<svg width="100%" height="5" class="mat-progress-bar-background mat-progress-bar-element">
+  <defs>
+    <pattern [id]="progressbarId" x="5" y="0" width="10" height="5" patternUnits="userSpaceOnUse">
+      <circle cx="2.5" cy="2.5" r="2.5"/>
+    </pattern>
+  </defs>
+  <rect [attr.fill]="'url(#' + progressbarId + ')'" width="100%" height="100%"/>
+</svg>
 <div class="mat-progress-bar-buffer mat-progress-bar-element" [ngStyle]="_bufferTransform()"></div>
 <div class="mat-progress-bar-primary mat-progress-bar-fill mat-progress-bar-element" [ngStyle]="_primaryTransform()"></div>
 <div class="mat-progress-bar-secondary mat-progress-bar-fill mat-progress-bar-element"></div>
--- a/src/lib/progress-bar/progress-bar.scss
+++ b/src/lib/progress-bar/progress-bar.scss
@@ -25,9 +25,9 @@ $mat-progress-bar-piece-animation-duration: 250ms !default;
   // The progress bar background is used to show the bubble animation scrolling behind a
   // buffering progress bar.
   .mat-progress-bar-background {
-    background-repeat: repeat-x;
-    background-size: 10px 4px;
-    display: none;
+    // The width is set to be 10px longer than the container as it will be translated 10px
+    // during the background scroll animation.
+    width: calc(100% + 10px);
   }
 
   // The progress bar buffer is the bar indicator showing the buffer value and is only visible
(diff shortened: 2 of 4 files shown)


src/lib/progress-bar/progress-bar.html
```diff
@@ -1,5 +1,17 @@
-<!-- The background div is named as such because it appears below the other divs and is not sized based on values. -->
-<div class="mat-progress-bar-background mat-progress-bar-element"></div>
+<!--
+  The background div is named as such because it appears below the other divs and is not sized based
+  on values.
+-->
+<div class="mat-progress-bar-background mat-progress-bar-element">
+  <svg width="100%" height="5">
+    <defs>
+      <pattern id="circle" x="5" y="0" width="10" height="5" patternUnits="userSpaceOnUse">
```
crisbeto: AFAIK this ID has to be unique.

src/lib/progress-bar/progress-bar.ts
```diff
@@ -79,6 +87,19 @@ export class MatProgressBar {
...
   }
+
+  /**
+   * Reappends the already attached circle SVG element in DOM to update the color styling.
+   *
+   * For colors to take effect in the background div, the svg's <circle> element, must be removed
+   * then placed back into the DOM again.  This is necessary as the browser only defines the
+   * styling of SVG pattern elements as it is placed into the DOM.
+   */
+  private _reappendCircleSvgElement() {
```
crisbeto: I'm not convinced that this is necessary. The progress spinner supports changing the colors without having to re-append the element to the DOM. Are you sure that it doesn't have something to do with the circles not having a unique id?

josephperrott: I think you are right

src/lib/progress-bar/progress-bar.html
```diff
@@ -1,5 +1,17 @@
-<!-- The background div is named as such because it appears below the other divs and is not sized based on values. -->
-<div class="mat-progress-bar-background mat-progress-bar-element"></div>
+<!--
+  The background div is named as such because it appears below the other divs and is not sized based
+  on values.
+-->
+<div class="mat-progress-bar-background mat-progress-bar-element">
+  <svg width="100%" height="5">
+    <defs>
+      <pattern id="{{progressbarId}}" x="5" y="0" width="10" height="5" patternUnits="userSpaceOnUse">
```
jelbourn: Use `[id]` instead of interpolation?

src/lib/progress-bar/progress-bar.scss
```diff
@@ -26,9 +26,12 @@ $mat-progress-bar-piece-animation-duration: 250ms !default;
   // The progress bar background is used to show the bubble animation scrolling behind a
   // buffering progress bar.
   .mat-progress-bar-background {
-    background-repeat: repeat-x;
-    background-size: 10px 4px;
-    display: none;
+    svg {
+      position: absolute;
```
jelbourn: Also comment on why the element is `absolute`?

josephperrott:
@jelbourn ready for another look when you have a moment.

PR #8908, opened 2017-12-09
refactor(datepicker): switch to fakeAsync tests

crisbeto:
* Switches over all of the datepicker tests to run in the `fakeAsync` zone, making them cleaner and more reliable.
* Avoids recompiling all of the test components before each test.


andrewseguin:
Needs rebase

PR #8918, opened 2017-12-10
fix(datepicker): unable to close calendar when opened on focus in IE11

### LINKED ISSUE #8914: IE11: Datepicker cannot be dismissed by user interaction after being opened programmatically

Looks like a bug

It should be possible to close the calendar after open it programmatically.

The datepicker, after being opened programmatically cannot be closed by user interactions.

image

image

Open it in IE11: https://stackblitz.com/edit/open-calendar-on-focus

According to the Material2 Description (Github) the components should work fine in IE11 

Apparently, just IE11

crisbeto:
Fixes not being able to close a datepicker's calendar in IE11, if the datepicker's trigger opens it on focus. The issue comes down to the fact that all browsers focus elements synchronously, whereas IE does so asynchronously. Since our logic depends on everything firing in sequence, when IE focuses at a later point, the datepicker is already considered as closed which causes the logic that restores focus to the trigger to reopen the calendar.

Fixes #8914.

### DIFF
--- a/src/lib/datepicker/datepicker.ts
+++ b/src/lib/datepicker/datepicker.ts
@@ -310,15 +310,25 @@ export class MatDatepicker<D> implements OnDestroy {
     if (this._calendarPortal && this._calendarPortal.isAttached) {
       this._calendarPortal.detach();
     }
+
+    const completeClose = () => {
+      this._opened = false;
+      this.closedStream.emit();
+      this._focusedElementBeforeOpen = null;
+    };
+
     if (this._focusedElementBeforeOpen &&
       typeof this._focusedElementBeforeOpen.focus === 'function') {
-
+      // Because IE moves focus asynchronously, we can't count on it being restored before we've
+      // marked the datepicker as closed. If the event fires out of sequence and the element that
+      // we're refocusing opens the datepicker on focus, the user could be stuck with not being
+      // able to close the calendar at all. We work around it by making the logic, that marks
+      // the datepicker as closed, async as well.
       this._focusedElementBeforeOpen.focus();
-      this._focusedElementBeforeOpen = null;
+      setTimeout(completeClose);
+    } else {
+      completeClose();
     }
-
-    this._opened = false;
-    this.closedStream.emit();
   }
 
   /** Open the calendar as a dialog. */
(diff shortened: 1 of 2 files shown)


src/lib/datepicker/datepicker.ts
```diff
@@ -310,15 +310,25 @@ export class MatDatepicker<D> implements OnDestroy {
...
       typeof this._focusedElementBeforeOpen.focus === 'function') {
-
+      // Because IE moves focus asynchronously, we can't count on it being restored before we've
+      // marked the datepicker as closed. If the event fires out of sequence and the element that
+      // we're refocusing opens the datepicker on focus, the user could be stuck with not being
+      // able to close the calendar at all. We work around it by making the logic, that marks
+      // the datepicker as closed, async as well.
       this._focusedElementBeforeOpen.focus();
-      this._focusedElementBeforeOpen = null;
+      setTimeout(completeClose);
```
crisbeto: Note: I'm not a fan of having to do this timeout, however it's probably the least-breaking option. The alternatives are:

1. Adding a one-time `focus` event to the `_focusedElementBeforeOpen` that completes the closing logic. This works as well as the timeout, however it makes it a lot harder to flush out during unit tests, because you'd need to know the `document.activeElement` and you'd need to dispatch a fake `focus` event.
2. Adding a `blur` event on the currently-focused element. Has the same issues as 1.
3. Only doing this for IE. The problem with this is that some `fakeAsync` tests would fail only when they're run on IE.

mmalerba: is `Promise.resolve().then` to quick for this?

crisbeto: I don't remember trying it, but it should work. I went with the timeout since it would be more obvious what you need to do if your tests start failing. We can switch to the `Promise.resolve` if we want to make the sync easier.

mmalerba: I prefer to use it when possible since it executes quicker. I think adding a `flush` to your test will fix it in either case

crisbeto: It will, but the `Promise.resolve` won't throw the `X timers in queue` error which makes it a little harder to track down.

mmalerba: really? so what happens if it never gets flushed, does the test just pass?

crisbeto: Whatever is supposed to execute in the callback will happen after the test. In our case the `opened` property on the datepicker would always stay at false.

mmalerba: Oh, that sucks, would have expected `fakeAsync` to catch that. Fine with leaving it as `setTimeout` then

PR #8920, opened 2017-12-10
fix(stepper): use up/down arrows for navigating vertical stepper

crisbeto:
Currently both vertical and horizontal steppers use the left/right arrows to move focus between steps. Based on the a11y guidelines (https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel) the vertical stepper should use the up/down arrows instead.

### DIFF
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -27,7 +27,7 @@ import {
   OnChanges,
   OnDestroy
 } from '@angular/core';
-import {LEFT_ARROW, RIGHT_ARROW, ENTER, SPACE} from '@angular/cdk/keycodes';
+import {LEFT_ARROW, RIGHT_ARROW, DOWN_ARROW, UP_ARROW, ENTER, SPACE} from '@angular/cdk/keycodes';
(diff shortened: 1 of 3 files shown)


mmalerba:
Should the horizontal arrows still work for the vertical stepper? The way they worded it, it sounds kind of like up/down are an extra feature for vertical, but they all support left/right:

> 2. If the tabs in a tab list are arranged vertically:
>     A. Down Arrow performs as Right Arrow is described above.
>     B. Up Arrow performs as Left Arrow is described above.
> 3. If the tab list is horizontal, it does not listen for Down Arrow or Up Arrow so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.

crisbeto:
Hmm I see what you mean, but it seems a little counter-intuitive.

mmalerba:
I can see how users would want it though. I suspect if I relied primarily on keyboard for navigation I wouldn't want to have to think about what direction my tabs are, I would just want my right/left to do the same thing as usual

crisbeto:
Makes sense. I've put back the left/right behavior for the vertical one and added the up/down on top.

PR #8945, opened 2017-12-12
docs(changelog): fix wrong description

rafaelss95:


rafaelss95:
I guess you have a small problem with changelog generation.

It seems like the script is taking the initial title (which had a typo in `attached` word) from PR #8642 instead of the latest.

Also, it appended the link, that is in description, to the title.

PR #8946, opened 2017-12-12
fix(paginator): coerce string values

### LINKED ISSUE #8624: [Pagination] Example passes string for pageSize, breaking getRangeLabel
Bug

When I set page size to 1 then go to next page, it should changes from `1 - 1 of 100` to `2 - 2 of 100`. Then next page again it should be `3-3 of 100`

When I set page size to 2 then go to next page, it should changes from `1 - 2 of 100` to `3 - 4 of 100`. Then next page again it should be `5-6 of 100`

angular material io pagination of data table bug

When I set page size to 1 then go to next page, it changes from `1 - 1 of 100` to `2 - 20 of 100`. Then next page again it should be `3-30 of 100`

When I set page size to 2 then go to next page, it changes from `1 - 2 of 100` to `3 - 22 of 100`. Then next page again it should be `5-42 of 100`

https://material.angular.io/components/paginator/examples

I am just trying to test it for the first time

Angular 5.0.3
Angular Material  5.0.0-rc.1

Plunkr example is not working too(In most of the time for other examples too actually)
https://github.com/angular/material.angular.io/issues/334

rafaelss95:
Coerces any string values that are passed to `pageIndex`, `length`, `pageSize`  to numbers.

Also, I added the coercion to `pageSizeOptions` to allow people to pass a string array, e.g: `['1', '5', ...]`.

Fixes #8624.

### DIFF
--- a/src/lib/paginator/paginator.spec.ts
+++ b/src/lib/paginator/paginator.spec.ts
@@ -23,6 +23,7 @@ describe('MatPaginator', () => {
         MatPaginatorWithoutPageSizeApp,
         MatPaginatorWithoutOptionsApp,
         MatPaginatorWithoutInputsApp,
+        MatPaginatorWithStringValues
       ],
       providers: [MatPaginatorIntl]
     }).compileComponents();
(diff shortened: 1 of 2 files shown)


mhosman:
This fix is available in a new version???

Issue #8206, opened 2017-11-03
[Table] Header row has the wrong min-height


TheRedD0t:
#### Bug, feature request, or proposal:
bug

#### What is the expected behavior?
According to the guidelines, the height of the header row of a data table should be `56px`.

https://material.io/guidelines/components/data-tables.html#data-tables-specs

#### What is the current behavior?
The header row has the same height as a normal row (`min-height: 48px`).

jasiekmiko:
Covered in #8215

PR #8215, opened 2017-11-03
fix(table): set height of mat-header-row to 56px

TheRedD0t:
Fixes #8206

PR #8955, opened 2017-12-12
fix(table): fix header row height to follow material guidelines

jasiekmiko:
Change the header row height from 48px to 56dx, as per
https://material.io/guidelines/components/data-tables.html#data-tables-specs

In response to issue #8206


willshowell:
It looks like https://github.com/angular/material2/pull/8215 already accomplishes this

jasiekmiko:
Oh darn, I completely missed that :/ Thanks!

PR #8962, opened 2017-12-12
fix(tabs): mark tab body as a scrollable container

### LINKED ISSUE #8405: Scrollable overlays not staying attached when inside tabs
Seems like this a bug unless someone can point me in a direction why what I have doesn't work.

When applying the `cdkScrollable` class I would expect the overlay to stay attached to the element

Does not stay attached to the element

Plunker starter (using on `@master`): https://goo.gl/uDmqyY<br/>

Here is a StackBlitz:
https://stackblitz.com/edit/angular-material2-issue-yqzqdh

In the above I am applying the `cdkScrollable` class but it still does not seem to stay attached properly. Any ideas why? I believe I have it in the right spot which is on the element that is scrolling.
Tried it other places as well but still no luck.

Would like to use overlay components inside tabs.

crisbeto:
Since the tab body content can be scrollable and consumers don't have access to the DOM element, we have to mark it as a `CdkScrollable` in order to allow for overlays inside the tabs to reposition.

Fixes #8405.

### DIFF
--- a/src/lib/tabs/tab-body.html
+++ b/src/lib/tabs/tab-body.html
@@ -4,6 +4,7 @@
         params: {animationDuration: animationDuration}
      }"
      (@translateTab.start)="_onTranslateTabStarted($event)"
-     (@translateTab.done)="_translateTabComplete.next($event)">
+     (@translateTab.done)="_translateTabComplete.next($event)"
+     cdkScrollable>
   <ng-template matTabBodyHost></ng-template>
 </div>
--- a/src/lib/tabs/tab-body.spec.ts
+++ b/src/lib/tabs/tab-body.spec.ts
@@ -7,6 +7,8 @@ import {MatRippleModule} from '@angular/material/core';
 import {NoopAnimationsModule} from '@angular/platform-browser/animations';
 import {MatTabBody, MatTabBodyPortal} from './tab-body';
 import {Subject} from 'rxjs';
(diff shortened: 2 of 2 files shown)


src/lib/tabs/tab-body.html
```diff
@@ -1,6 +1,7 @@
 <div class="mat-tab-body-content" #content
      [@translateTab]="_position"
      (@translateTab.start)="_onTranslateTabStarted($event)"
-     (@translateTab.done)="_onTranslateTabComplete($event)">
+     (@translateTab.done)="_onTranslateTabComplete($event)"
+     cdkScrollable>
```
alexw10: FYI I dont think this is the correct spot for this, I tested this out and it did not seem to fix the issue.
However, I did move cdkScrollable to the tab-group.html where the <mat-tab-body> is and that does seem to solve it.

crisbeto: At least in our dev app, this is the element that ends up being scrollable. Did you override any of the default styles in your app?

andrewseguin: This may be affected by this PR in the future https://github.com/angular/material2/pull/6861

Seems the scrolling container should be a parent of the tab content

crisbeto: It looks like it probably would, let's defer this one until #6861 gets in.

jelbourn:
@crisbeto to rebase or re-open

crisbeto:
Closing in favor of #16344.

PR #8964, opened 2017-12-12
fix(select): don't restore focus unless an option was selected

### LINKED ISSUE #8915: multiple select problem
Hi,

with multiple select i had some problem with positioning of the "overlay" (the select panel with options) on a scrolling page (forgive my english, i know, i am not good at all on it)

https://stackblitz.com/edit/angular-material2-issue-fyg1vr

this is what i got

1) scroll down
2) click on a select, leave it open
3) scroll some more, click on another select, i get teleported back to the previous position
but hte selec panel disappear

antoher things

1) randomly scroll
2) click -> select multiple times
3) scroll to bottom, open a new select, the scroll go crazy and the overlay of the select get positioned randomly on top position

on mobile chrome it's evern worst.

test was with chrome, latest version, osx. angular 5, material.io 5

crisbeto:
Currently `mat-select` always restores focus to its trigger whenever the panel is closed. This can be problematic, because the user might've scrolled away and restoring focus would scroll the page back up. These changes switch to only restoring focus if the user made a selection.

Fixes #8915.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -298,7 +298,7 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit, 
   ];
 
   /** Whether the select is focused. */
-  focused = false;
+  focused: boolean = false;
 
   /** A name for this control that can be used by `mat-form-field`. */
   controlType = 'mat-select';
@@ -548,7 +548,6 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit, 
       this._panelOpen = false;
       this._changeDetectorRef.markForCheck();
       this._onTouched();
-      this.focus();
     }
   }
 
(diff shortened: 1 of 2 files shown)


andrewseguin:
Needs rebase

src/lib/select/select.ts
```diff
@@ -297,8 +297,12 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit,
     },
   ];
 
+  private _hasFocus = false;
+
   /** Whether the select is focused. */
-  focused = false;
+  get focused() {
```
jelbourn: Uncovered during presubmit: this makes `focused` readonly, which is a breaking change
(apparently some apps set it manually for some reason)

crisbeto:
Reverted the breaking change @jelbourn.

PR #8967, opened 2017-12-12
test(tree): Add more tests for cdk-tree

tinayuangao:


src/cdk/tree/tree.spec.ts
```diff
@@ -26,26 +27,215 @@ describe('CdkTree', () => {
...
         SimpleCdkTreeApp,
-        // TODO(tinayuangao): Add more test cases with the cdk-tree
-        //   DynamicDataSourceCdkTreeApp,
-        //   NodeContextCdkTreeApp,
-        //   WhenNodeCdkTreeApp
+        NestedCdkTreeApp,
+        CdkTreeAppWithTrigger,
+        NestedCdkTreeAppWithTrigger,
+        WhenNodeCdkTreeApp,
+        WhenNodeNestedCdkTreeApp,
```
jelbourn: Can you break up these into separate describe blocks (similar to the mat-select tests)? Organized this way, Angular will compile each test component for every `it`, which pretty dramatically slows down the tests (this is why IE11 has been timing out recently). This will also resolve the `ComponentFixture` being `any`

tinayuangao: Done. Please take a look. Thanks!

src/cdk/tree/tree.spec.ts
```diff
@@ -160,20 +557,143 @@ class SimpleCdkTreeApp {
...
+  isExpandable = (node: TestData) => node.children.length > 0;
+
+  treeControl: TreeControl<TestData> = new FlatTreeControl(this.getLevel, this.isExpandable);
+
+  dataSource: FakeDataSource | null = new FakeDataSource(this.treeControl);
+
   @ViewChild(CdkTree) tree: CdkTree<TestData>;
 }
 
 function getNodes(treeElement: Element): Element[] {
```
jelbourn: Can you move these helper functions above the test components?

src/cdk/tree/tree.spec.ts
```diff
@@ -160,20 +557,143 @@ class SimpleCdkTreeApp {
...
   return [].slice.call(treeElement.querySelectorAll('.cdk-tree-node'))!;
 }
 
-// TODO(tinayuangao): Add expectedNestedTreeToMatchContent
 function expectFlatTreeToMatchContent(treeElement: Element, expectedTreeContent: any[],
```
jelbourn: I would just call these `expectFlatTreeToMatch` and `expectNestedTreeToMatch`

src/cdk/tree/tree.spec.ts
```diff
@@ -200,3 +720,34 @@ function expectFlatTreeToMatchContent(treeElement: Element, expectedTreeContent:
     fail(missedExpectations.join('\n'));
   }
 }
+
+function expectNestedTreeToMatchContent(treeElement: Element, expectedTreeContent: any[],
```
jelbourn: As discussed offline: let's try doing something like
```ts
/** Represents an indent for expectNestedTreeToMatch */
const _ = {};

expectNestedTreeToMatch(
  ['Root'],
  [_, 'Under'],
  [_, _, 'More under'],
);
```

tinayuangao:
@jelbourn Comments addressed. Please take a look. Thanks!

PR #8971, opened 2017-12-13
refactor: export component animations for reuse

### LINKED ISSUE #8904: Animations should be exported for re-use
Feature request

Animations to be exported

Animations are not exported

Building custom material design components that match existing components without duplicating code. For example I would like to create sort component that would allow me to sort table on mobile devices when not every column is visible and the space is limited. Right now I'm forced to copy animations from https://github.com/angular/material2/blob/b4a83ab49a1c3d27acfdeb2ece8e21f755a32606/src/lib/sort/sort-header.ts#L69-L104 to look the same as `MatSortHeader`. 

With exported animations I could reuse that code.

5.0.0

Some animations *are* exported so there is an inconsistency.

https://github.com/angular/material2/blob/b4a83ab49a1c3d27acfdeb2ece8e21f755a32606/src/lib/select/select-animations.ts#L34-L69

crisbeto:
Exports all of the component animations so they can be reused by consumers.

Fixes #8904.

### DIFF
--- a/src/lib/dialog/dialog-animations.ts
+++ b/src/lib/dialog/dialog-animations.ts
@@ -1,0 +1,32 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {
+  animate,
+  state,
+  style,
+  transition,
+  trigger,
+  AnimationTriggerMetadata,
+} from '@angular/animations';
+
+/** Animations used by MatDialog. */
+export const matDialogAnimations: {
+  readonly slideDialog: AnimationTriggerMetadata;
+} = {
+  /** Animation that slides the dialog in and out of view and fades the opacity. */
+  slideDialog: trigger('slideDialog', [
+    // Note: The `enter` animation doesn't transition to something like `translate3d(0, 0, 0)
+    // scale(1)`, because for some reason specifying the transform explicitly, causes IE both
+    // to blur the dialog content and decimate the animation performance. Leaving it as `none`
+    // solves both issues.
+    state('enter', style({ transform: 'none', opacity: 1 })),
+    state('void', style({ transform: 'translate3d(0, 25%, 0) scale(0.9)', opacity: 0 })),
+    state('exit', style({ transform: 'translate3d(0, 25%, 0)', opacity: 0 })),
+    transition('* => *', animate('400ms cubic-bezier(0.25, 0.8, 0.25, 1)')),
+  ])
+};
--- a/src/lib/dialog/dialog-container.ts
+++ b/src/lib/dialog/dialog-container.ts
@@ -19,8 +19,9 @@ import {
   ViewEncapsulation,
   ChangeDetectionStrategy,
 } from '@angular/core';
-import {animate, AnimationEvent, state, style, transition, trigger} from '@angular/animations';
 import {DOCUMENT} from '@angular/common';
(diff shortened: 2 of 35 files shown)


josephperrott:
If we are exporting them all publically perhaps we should "namespace" them?  We could put all of the definitions inside of respective objects for animations.  I worry about polluting the exported namespace.

For example: 
```ts
export const ExpansionPanelAnimations = {
  indicatorRotate: AnimationTriggerMetadata = trigger('indicatorRotate', [
    state('collapsed', style({transform: 'rotate(0deg)'})),
    state('expanded', style({transform: 'rotate(180deg)'})),
    transition('expanded <=> collapsed',
               animate(EXPANSION_PANEL_ANIMATION_TIMING)),
  ]),
  expansionHeaderHeight: AnimationTriggerMetadata = trigger('expansionHeight', [
    state('collapsed', style({
      height: '{{collapsedHeight}}',
    }), {
      params: {collapsedHeight: '48px'},
    }),
    state('expanded', style({
      height: '{{expandedHeight}}'
    }), {
      params: {expandedHeight: '64px'}
    }),
    transition('expanded <=> collapsed', 
               animate(EXPANSION_PANEL_ANIMATION_TIMING)),
  ]),
  bodyExpansion: AnimationTriggerMetadata = trigger('bodyExpansion', [
    state('collapsed', style({height: '0px', visibility: 'hidden'})),
    state('expanded', style({height: '*', visibility: 'visible'})),
    transition('expanded <=> collapsed', 
               animate(EXPANSION_PANEL_ANIMATION_TIMING)),
  ])
}
```

crisbeto:
Good point @josephperrott, I was hoping that this would come up during the review since I'm not a fan of just exporting them as plain variables either. Switch over to namespaces.

src/lib/dialog/dialog-animations.ts
```diff
@@ -0,0 +1,34 @@
...
+  style,
+  transition,
+  trigger,
+  AnimationTriggerMetadata,
+} from '@angular/animations';
+
+/**
+ * Animations used by MatDialog.
+ */
+export namespace MatDialogAnimations {
```
jelbourn: We can't use typescript namespaces; they're forbidden inside google (has to do with interop with Closure-style javascript)

crisbeto: In this case what would be the proper way to namespace them? Should we have a `MatDialogAnimations` class with static properties or just an object with some properties on it? The static class has the advantage of allowing us to make the properties readonly.

jelbourn: I would just do an object that we can use typescript to make immutable

crisbeto:
Refactored it based on our discussion @jelbourn. Note that I'm not a fan of having to add the typings to all of the constants, but it seems like the simplest way of ensuring that the properties are immutable. Initially I was going for something like `{readonly [key: string]: AnimationTriggerMetadata}`, but that really only prevents people from adding new keys, it doesn't break if they try to override the existing ones.

src/lib/dialog/dialog-animations.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Animations used by MatDialog.
+ */
+export const matDialogAnimations: {
+  readonly slideDialog: AnimationTriggerMetadata;
+} = {
+  /**
+   * Animation that slides the dialog in and out of view and fades the opacity.
+   */
```
jelbourn: nit: can condense these comments to a single line

jelbourn:
Changing this to minor release since it adds new APIs

PR #9016, opened 2017-12-16
fix(dialog): hide all non-overlay content from assistive technology

### LINKED ISSUE #7787: [Accessibility] MatDialog is trapping tab key but not headings navigation shortcuts

MatDialog is trapping tab key but not headings navigation shortcuts, this is actually working well on AngularJS Material [1] [2]

[1] https://material.angularjs.org/latest/demo/dialog
[2] https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L1134

Screen reader should trap tab key as well as headings navigation shortcuts when modal dialog is opened.

Screen reader is not trapping headings navigation shortcuts when modal dialog is opened.

"@angular/material": "2.0.0-beta.12"

crisbeto:
Hides all non-overlay content from assistive technology by applying `aria-hidden` to it. This prevents users from being able to move focus out of the dialog using the screen reader navigational shortcuts.

Fixes #7787.

### DIFF
--- a/src/lib/dialog/dialog.ts
+++ b/src/lib/dialog/dialog.ts
@@ -14,6 +14,7 @@ import {
   OverlayConfig,
   OverlayRef,
   ScrollStrategy,
+  OverlayContainer,
 } from '@angular/cdk/overlay';
 import {ComponentPortal, ComponentType, PortalInjector, TemplatePortal} from '@angular/cdk/portal';
 import {Location} from '@angular/common';
@@ -67,6 +68,7 @@ export class MatDialog {
   private _openDialogsAtThisLevel: MatDialogRef<any>[] = [];
   private _afterAllClosedAtThisLevel = new Subject<void>();
   private _afterOpenAtThisLevel = new Subject<MatDialogRef<any>>();
+  private _ariaHiddenElements = new Map<Element, string|null>();
 
   /** Keeps track of the currently-open dialogs. */
   get openDialogs(): MatDialogRef<any>[] {
@@ -96,7 +98,8 @@ export class MatDialog {
       private _injector: Injector,
       @Optional() location: Location,
       @Inject(MAT_DIALOG_SCROLL_STRATEGY) private _scrollStrategy,
-      @Optional() @SkipSelf() private _parentDialog: MatDialog) {
+      @Optional() @SkipSelf() private _parentDialog: MatDialog,
+      private _overlayContainer: OverlayContainer) {
 
     // Close all of the dialogs when the user goes forwards/backwards in history or when the
     // location hash changes. Note that this usually doesn't include clicking on links (unless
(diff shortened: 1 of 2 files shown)


src/lib/dialog/dialog.ts
```diff
@@ -67,6 +69,7 @@ export class MatDialog {
   private _openDialogsAtThisLevel: MatDialogRef<any>[] = [];
   private _afterAllClosedAtThisLevel = new Subject<void>();
   private _afterOpenAtThisLevel = new Subject<MatDialogRef<any>>();
+  private _ariaHiddenElements = new Map<Element, string|null>();
```
jelbourn: `WeakMap` as a precaution against memory leaks?

src/lib/dialog/dialog.ts
```diff
@@ -298,12 +308,44 @@ export class MatDialog {
...
+        this._ariaHiddenElements.clear();
         this._afterAllClosed.next();
       }
     }
   }
+
+  /**
+   * Hides all of the content that isn't an overlay from assistive technology.
+   */
+  private _hideContent() {
```
jelbourn: `_hideNonDialogContentFromAssistiveTechnology`?

src/lib/dialog/dialog.ts
```diff
@@ -298,12 +308,44 @@ export class MatDialog {
...
+    const overlayContainer = this._overlayContainer.getContainerElement();
+    const siblings = overlayContainer.parentElement!.children;
+
+    for (let i = 0; i < siblings.length; i++) {
```
jelbourn: nit: prefer iteration from length - 1 to zero for `HTMLCollection`

src/lib/dialog/dialog.ts
```diff
@@ -298,12 +308,44 @@ export class MatDialog {
...
+      let sibling = siblings[i];
+
+      if (sibling !== overlayContainer &&
+        sibling !== this._liveAnnouncer.liveElement &&
```
jelbourn: Instead of checking this element directly, just check that the element doesn't have `aria-live`? it will accomplish the same thing without needing to change the live announcer API and also safeguards against something that does its own live announcements.

src/lib/dialog/dialog.ts
crisbeto: We can't use a `WeakMap`, because we need to be able to iterate over the elements.

crisbeto:

src/lib/dialog/dialog.ts
```diff
@@ -298,12 +306,44 @@ export class MatDialog {
...
+        this._ariaHiddenElements.clear();
         this._afterAllClosed.next();
       }
     }
   }
+
+  /**
+   * Hides all of the content that isn't an overlay from assistive technology.
+   */
+  private __hideNonDialogContentFromAssistiveTechnology() {
```
jelbourn: Extra `_`

josephperrott:
We are seeing a number of presubmit failures from how overlays are tested.

A common pattern is to provide the `OverlayContainer` within a test as a newly created unattached div.
```ts
{
  provide: OverlayContainer,
  useFactory: () => {
    overlayContainerElement = document.createElement('div');
    return {getContainerElement: () => overlayContainerElement};
  }
}
```

This change assumes that the overlayContainer will have a parent.  Is this a check we can do and only take action if a parentElement is found?

crisbeto:
Sure, done @josephperrott.

PR #9021, opened 2017-12-16
fix(dialog): change order of button actions

### LINKED ISSUE #9019: Confirmative dialog buttons should be placed on the right side
In all the material samples of the https://material.angular.io/components/dialog section, the confirmative dialog buttons should be placed on the right side of the dialogs, as per `Material design guidelines`: https://material.io/guidelines/components/dialogs.html#dialogs-specs

All the confirmative buttons (like `OK`, `Install` etc) should be placed on the right, whereas the Dismissive actions (like `No thanks` etc) should go to the left.

Currently, all the confirmative buttons are placed on the left, whereas the dismissive ones are placed on the right.

Open https://material.angular.io/components/dialog/overview and https://material.angular.io/components/dialog/examples pages and try open the sample dialogs

The samples should conform to the official Material design Guidelines

donroyco:
Change order of buttons for affirmative actions, according to [Material guidelines](https://material.io/guidelines/components/dialogs.html#dialogs-specs). This fixes #9019

### DIFF
--- a/src/demo-app/a11y/dialog/dialog-address-form-a11y.html
+++ b/src/demo-app/a11y/dialog/dialog-address-form-a11y.html
@@ -40,6 +40,6 @@
 </mat-dialog-content>
 
 <mat-dialog-actions>
-  <button mat-raised-button color="primary" mat-dialog-close>Submit</button>
   <button mat-raised-button mat-dialog-close>Close</button>
+  <button mat-raised-button color="primary" mat-dialog-close>Submit</button>
 </mat-dialog-actions>
--- a/src/demo-app/a11y/dialog/dialog-address-form-a11y.html
+++ b/src/demo-app/a11y/dialog/dialog-address-form-a11y.html
@@ -41,5 +41,5 @@
 
 <mat-dialog-actions>
   <button mat-raised-button mat-dialog-close>Close</button>
-  <button mat-raised-button color="primary" mat-dialog-close>Submit</button>
+  <button mat-raised-button color="primary" mat-dialog-close cdkFocusInitial>Submit</button>
 </mat-dialog-actions>
--- a/src/material-examples/dialog-overview/dialog-overview-example-dialog.html
+++ b/src/material-examples/dialog-overview/dialog-overview-example-dialog.html
@@ -6,6 +6,6 @@
   </mat-form-field>
 </div>
 <div mat-dialog-actions>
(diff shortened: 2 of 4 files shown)


src/material-examples/dialog-overview/dialog-overview-example-dialog.html
```diff
@@ -6,6 +6,6 @@ <h1 mat-dialog-title>Hi {{data.name}}</h1>
   </mat-form-field>
 </div>
 <div mat-dialog-actions>
-  <button mat-button [mat-dialog-close]="data.animal" tabindex="2">Ok</button>
   <button mat-button (click)="onNoClick()" tabindex="-1">No Thanks</button>
+  <button mat-button [mat-dialog-close]="data.animal" tabindex="2">Ok</button>
```
crisbeto: Not something that you added, but while we're at it, can you remove the `tabindex` from the buttons here and in the other example and replace the positive `tabindex` with an `cdk-focus-initial` attribute?

donroyco: Will do @crisbeto, is there an attribute order? `<button mat-button cdk-focus-initial [mat-dialog-close]="true">` or `<button mat-button [mat-dialog-close]="true" cdk-focus-initial>`

crisbeto: We don't have any rules for the attribute order.

donroyco: Ok!

src/demo-app/a11y/dialog/dialog-address-form-a11y.html
```diff
@@ -40,6 +40,6 @@ <h2 mat-dialog-title>Company</h2>
 </mat-dialog-content>
 
 <mat-dialog-actions>
-  <button mat-raised-button color="primary" mat-dialog-close>Submit</button>
   <button mat-raised-button mat-dialog-close>Close</button>
+  <button mat-raised-button color="primary" mat-dialog-close>Submit</button>
```
crisbeto: Can you add the attribute to this one. Also sorry, but I just remembered that we deprecated `cdk-focus-initial`.  `cdkFocusInitial` is the proper one.

donroyco: No problem at all

PR #9025, opened 2017-12-17
feat(observe-content): allow for the MutationObserver to be disabled

crisbeto:
Adds the ability for users to disable the underlying `MutationObserver` inside the `CdkObserveContent` directive. This can be useful in the cases where it might be expensive to continue observing an element while it is invisible (e.g. an item inside of a closed dropdown).


jelbourn:
@crisbeto this passes presubmit but the bazel build file needs to be updated (dep on coercion)

crisbeto:
Sorted out the Bazel failure.

PR #9026, opened 2017-12-17
fix(selection-list): allow users to jump focus to a particular item by typing

crisbeto:
Similarly to our other listbox-based components, this allows people to move focus to a particular item by typing the first letters of its label.


jelbourn:
@crisbeto this passes presubmit; if you can rebase I can merge it

crisbeto:
Rebased.

PR #9036, opened 2017-12-18
feat(tree): add data sources for tree

tinayuangao:


src/lib/tree/data-source/flat-data-source.ts
```diff
@@ -0,0 +1,153 @@
...
+ *    expandable: false,
+ *    level: 2
+ *  },
+ *  NodeTwo: {
+  *   key: 'Pear',
+  *   expandable: false,
+  *   level: 2
+  * }
+ */
+export class MatTreeFlattener<T, F> {
```
jelbourn: I don't quite follow what `T` and `F` are. Is one the type of node and one the type of data? If so, should the type of node include an `extends`?

src/lib/tree/data-source/nested-data-source.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Data source for nested tree.
+ *
+ * The data source for nested tree doesn't have to consider node flattener, or the way to expand
+ * or collapse. The expansion/collapsion will be handled by TreeControl and each non-leaf node.
+ */
+export class MatTreeNestedDataSource<T> implements DataSource<T> {
+  _renderedData = new BehaviorSubject<T[]>([]);
+  get renderedData(): T[] { return this._renderedData.value; }
```
jelbourn: I don't quite follow what `renderedData` is needed/used for

src/lib/tree/data-source/flat-data-source.ts
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Data source for flat tree.
+ * The data source need to handle expansion/collapsion of the tree node and change the data feed
+ * to `MatTree`.
+ * The nested tree nodes of type `T` are flattened through `MatTreeFlattener`, and converted
+ * to type `F` for `MatTree` to consume.
+ */
+export class MatTreeFlatDataSource<T, F> implements DataSource<F> {
+  _flattenedData = new BehaviorSubject<F[]>([]);
+  get flattenedData() { return this._flattenedData.value; }
```
jelbourn: Why expose `flattenedData` in addition to the data stream coming back from `connect`?

src/lib/tree/data-source/flat-data-source.ts
tinayuangao: Discussed offline and added comments about this.

src/lib/tree/data-source/nested-data-source.ts
tinayuangao: Removed. `data` is enough

src/lib/tree/data-source/flat-data-source.ts
tinayuangao: Removed getters

jelbourn:
Poke @andrewseguin

PR #9037, opened 2017-12-18
fix(select): support using shift + arrow key to toggle items in a multi-select

crisbeto:
Based on the accessibility guidelines for a multi-selection listbox (https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox), potentially we can support toggling the previous/next item if the user presses shift + up/down arrow.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -665,7 +665,15 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit, 
       event.preventDefault();
       this._keyManager.activeItem._selectViaInteraction();
     } else {
+      const isArrowKey = keyCode === DOWN_ARROW || keyCode === UP_ARROW;
+      const previouslyFocusedIndex = this._keyManager.activeItemIndex;
+
       this._keyManager.onKeydown(event);
+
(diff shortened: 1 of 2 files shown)


src/lib/select/select.ts
```diff
@@ -665,7 +665,15 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit,
...
       this._keyManager.activeItem._selectViaInteraction();
     } else {
+      const isArrowKey = keyCode === DOWN_ARROW || keyCode === UP_ARROW;
+      const previouslyFocusedIndex = this._keyManager.activeItemIndex;
+
       this._keyManager.onKeydown(event);
+
+      if (this._multiple && isArrowKey && event.shiftKey && this._keyManager.activeItem &&
+          this._keyManager.activeItemIndex !== previouslyFocusedIndex) {
+        this._keyManager.activeItem._selectViaInteraction();
```
jelbourn: Should the key manager do this?

crisbeto: The key manager currently doesn't have a concept of what is selected, it only knows what is focused/highlighted.

PR #9042, opened 2017-12-18
feat(tree): Use tabindex for tree node

tinayuangao:


src/lib/tree/node.ts
```diff
@@ -7,34 +7,49 @@
...
-export class MatTreeNode<T> extends CdkTreeNode<T> {
+export class MatTreeNode<T> extends _MatTreeNodeMixinBase<T> implements HasTabIndex, CanDisable {
   @Input() role: 'treeitem' | 'group' = 'treeitem';
+
+  constructor(protected _elementRef: ElementRef,
+              protected _tree: CdkTree<T>,
+              @Attribute('tabindex') tabIndex: string) {
+    super(_elementRef, _tree);
+
+    this.tabIndex = parseInt(tabIndex) || 0;
```
jelbourn: Prefer `Number(tabIndex)` to `parseInt`
(here and below)

PR #9043, opened 2017-12-19
fix(chip):  Fix chips prevent anchor link navigation

### LINKED ISSUE #9032: Chips prevent anchor link navigatation
Bug

Anchor links inside (or wrapping) chips should navigate to their href

Placing ```<a>``` inside ```<mat-chip>``` or placing ```<mat-chip>``` inside ```<a>``` causes no navigation to the ```<a>```'s href.

1. Click the button "Edit in StackBlitz" in mat-chips page at https://material.angular.io/components/chips/overview and add these two lines:
  ```<a href="https://google.com"><mat-chip>Does not work as parent</mat-chip></a>```
  ```<mat-chip><a href="https://google.com">Does not work as child</a></mat-chip>```
2. Click on "Does not work as ..." links

Chips that are capable to navigate to another page without the need of javascript code

Angular: 5.1.1
Material: 5.0.1 (does not work with 5.0.0-rc.3 also which was initially loaded by StackBlitz)
OS: Windows 7
TypeScript: The version that StackBlitz uses

tinayuangao:
Fixes #9032

### DIFF
--- a/src/lib/chips/chip.ts
+++ b/src/lib/chips/chip.ts
@@ -252,7 +252,6 @@ export class MatChip extends _MatChipMixinBase implements FocusableOption, OnDes
       return;
     }
 
-    event.preventDefault();
     event.stopPropagation();
 
     this.focus();


src/lib/chips/chip.ts
```diff
@@ -252,7 +252,6 @@ export class MatChip extends _MatChipMixinBase implements FocusableOption, OnDes
       return;
     }
 
-    event.preventDefault();
```
jelbourn: Do you have any idea why this was here in the first place?

Issue #8180, opened 2017-11-01
Inconsistent chip bottom margin in mat-chip-list


josef-jelinek:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Chips in the mat-chip-list are aligned vertically.

#### What is the current behavior?
The chips are misaligned with the first one.

The responsible change is at
https://github.com/angular/material2/commit/7a4270661448948f5eeac6636ea7eff02f38b0cd#diff-2252115b723be2ad192335df6b5e9be8R32
which adds 3px bottom margin to all except the first chip (unless it is a stacked chip-list).

PR #9044, opened 2017-12-19
fix(chips): Add chips bottom margin

### LINKED ISSUE #8587: Chip input: no margin between chips in multiple lines
Bug

Have margin between chips in multiple lines.

No margin between chips in multiple lines.

https://plnkr.co/edit/KTSz1LYLyD3NH34qLDUW?p=preview

Design.

Latest.

Caused by https://github.com/angular/material2/pull/8198
What about adding top margin for all chips (instead of bottom for sibling chips)?
https://plnkr.co/edit/ItBhfqt4WBqMp0dV6lTa?p=preview or https://plnkr.co/edit/iUoyB4rr49Aco3W1I9Ix?p=preview
(The angularjs version also has style `margin: 8px 8px 0 0;` for all chips.)

tinayuangao:
Fixes #8587

### DIFF
--- a/src/lib/chips/chips.scss
+++ b/src/lib/chips/chips.scss
@@ -7,6 +7,7 @@ $mat-chip-remove-margin-before: 6px;
 $mat-chip-remove-margin-after: -4px;
 
 $mat-chips-chip-margin: 8px;
+$mat-chips-chip-bottom-margin: 3px;
 
 $mat-chip-input-width: 150px;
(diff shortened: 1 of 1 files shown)


josephperrott:
I think that last time we included this spacing we saw issues with the spacing between sibling chips: #8180 which was fixed by #8198

tinayuangao:
Yes, you are right. We should find another way to fix this.

PR #9051, opened 2017-12-19
fix(a11y): set opacity of cdk-visually-hidden

### LINKED ISSUE #9049: Checkbox display breaks if parent has a 'columns' property
Bug

Single checkboxes

OSX : checkboxes are doubled (the native checkbox is visible under the material one)
screen shot 2017-12-19 at 15 13 49
windows : a pixel is visible under the material checkbox

edot : forgot to add that the ripple breaks too.

[here is a stackblitz reproduction](https://stackblitz.com/edit/angular-material2-checkbox-issue-chrome)

```
    "@angular/animations": "^5.1.0",
    "@angular/cdk": "^5.0.0",
    "@angular/common": "^5.1.0",
    "@angular/compiler": "^5.1.0",
    "@angular/core": "^5.1.0",
    "@angular/forms": "^5.1.0",
    "@angular/http": "^5.1.0",
    "@angular/material": "^5.0.0",
```
chrome OSX : `Version 63.0.3239.108 (Official Build) (64-bit)`
chrome windows : `Version 63.0.3239.84 (Official Build) (64 bits)`

Could not reproduce on Safari

josephperrott:
fixes #9049

### DIFF
--- a/src/cdk/a11y/_a11y.scss
+++ b/src/cdk/a11y/_a11y.scss
@@ -4,6 +4,7 @@
     clip: rect(0 0 0 0);
     height: 1px;
     margin: -1px;
+    opacity: 0;
     overflow: hidden;
     padding: 0;
     position: absolute;


jelbourn:
I think this will break some screen readers; if `opacity: 0` worked everywhere, there wouldn't be a need for any of the other styles here

Lakston:
Thanks for the PR but I just tested it on our app, it fixes the display problem but it does not fix the ripple being broken

ripple

PR #9055, opened 2017-12-19
fix(datepicker): typo in MatMonthView export

crisbeto:
Fixes the `exportAs` of the `MatMonthView` being spelled as `matMonthVeiw`.


crisbeto:
@mmalerba FYI, marking this as a patch release since this component isn't meant to be consumed outside the calendar anyway.

PR #9059, opened 2017-12-19
chore: create cdk-experimental package

josephperrott:


src/cdk-experimental/version.ts
```diff
@@ -0,0 +1,12 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {Version} from '@angular/core';
+
+/** Current version of the Material experimental package. */
```
DevVersion: That should be `CDK experimental package`

Issue #4491, opened 2017-05-11
Add e2e tests for autocomplete


jelbourn:

saulotoledo:
@jelbourn I think one could add the "has pr" label to this issue.

jelbourn:
I don't see an open PR for this

PR #9077, opened 2017-12-20
(test): add e2e tests for autocomplete (#4491)

Ann40a:


Ann40a:
Please, help me, whom should I add to CODEOWNERS? Should it really be myself as I did or someone from your team?

.github/CODEOWNERS
```diff
@@ -129,6 +129,7 @@
 
 # E2E app
 /e2e/*                                             @jelbourn
+/e2e/components/autocomplete-e2e.spec.ts           @ann40a
```
josephperrott: Can you mark crisbeto as the code owner here as he is the code owner for the component.

e2e/components/autocomplete-e2e.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    page.autocompleteInput().sendKeys('T');
+
+    expect(await page.autocompletePanel().getText()).toEqual('Two\nThree');
+    screenshot();
+  });
+});
+
+class AutocompletePage {
+  constructor() { browser.get('/autocomplete'); }
+  autocompleteInput = () => element(by.id('autocomplete-input'));
```
josephperrott: Personally, I don't know that its strictly necessary to have a class to manage retrieving the elements as you have here.  If you would like to keep the class, rather than assigning these are attributes which are functions, can you use typical method declaration?


```ts
class AutocompletePage {
  constructor() {
    browser.get('/autocomplete');
  }

  autocompleteInput() {
    return element(by.id('autocomplete-input'));
  }
...
}
```

src/lib/autocomplete/autocomplete-trigger.ts
```diff
@@ -104,7 +104,7 @@ export function getMatAutocompleteMissingPanelError(): Error {
     'aria-autocomplete': 'list',
     '[attr.aria-activedescendant]': 'activeOption?.id',
     '[attr.aria-expanded]': 'panelOpen.toString()',
-    '[attr.aria-owns]': 'autocomplete?.id',
+    '[attr.aria-owns]': '(autocomplete && autocomplete._isOpen) ? autocomplete.id : null',
```
josephperrott: Can you move this logic into a getter method on the class and then use the attribute in the host binding?

e2e/components/autocomplete-e2e.spec.ts
```diff
@@ -0,0 +1,50 @@
+import {browser, by, element} from 'protractor';
+import {screenshot} from '../screenshot';
+import {expectToExist} from '../util/index';
+
+
+describe('autocomplete', () => {
+  const autocompletePanelSelector = '.mat-autocomplete-panel.mat-autocomplete-visible';
```
crisbeto: The `.mat-autocomplete-visible` here shouldn't be necessary, the only time an autocomplete is in the DOM and doesn't have that class is when there are no results.

e2e/components/autocomplete-e2e.spec.ts
crisbeto: Agreed. I don't think we have that many tests where setups like this start making sense.

src/lib/autocomplete/autocomplete-trigger.ts
crisbeto: I remember having issues with NVDA not reading out some options when I tried doing this for `mat-select`. That being said, I couldn't hear any difference when I did it for autocomplete. Still it might be better to do this in a separate PR since this one is primarily for the tests.

e2e/components/autocomplete-e2e.spec.ts
Ann40a: Personally, I prefer to user page objects, because it separates the test logic and elements selection. I've seen this pattern in `menu-e2e.spec.ts` and decided that it is acceptable to use it. I should remove this class, right?

src/lib/autocomplete/autocomplete-trigger.ts
crisbeto: I think that we can still do it, but it should be done as a separate PR, because it needs a unit test and this PR is primarily focused on adding e2e tests to the autocomplete.

e2e/components/autocomplete-e2e.spec.ts
crisbeto: We haven't been super consistent about when we use page objects (IIRC only the menu tests have them). Let's keep it for now and we can decide later on whether we want to remove it or switch all of the other e2e tests to use page objects as well.

src/e2e-app/autocomplete/autocomplete-e2e.html
```diff
@@ -0,0 +1,11 @@
+<section>
```
jelbourn: One thing that's not clear in the code by itself is that we're trying to move away from standalone e2e test pages. Instead, new e2e tests are based on the examples we publish to material.angular.io. Check out #5735 for an example of adding e2e tests for cards based on the published examples.

jelbourn:
Closing since this seems pretty stale now.

PR #9078, opened 2017-12-20
refactor(dialog): switch to common keyboard handling and move keyboard logic into individual dialog

crisbeto:
* Refactors the dialogs to use the common keyboard provider, rather than binding to the document.
* Moves the responsibility for handling the keyboard events from the dialog provider to the individual dialogs.


jelbourn:
Changing to minor release since it doesn't cherry-pick cleanly

PR #9093, opened 2017-12-21
bugfix(tooltip): set correct font-size mobile

### LINKED ISSUE #9039: [Tooltip] Mobile font size is too small
Bug

Font-size of tooltip on mobile devices should be `14sp` according to the [Material specs](https://material.io/guidelines/components/tooltips.html#tooltips-tooltips-mobile). On desktop it should be `10sp` according to [it specs](https://material.io/guidelines/components/tooltips.html#tooltips-tooltips-desktop).

For both mobile and desktop the behaviour looks like it's just `10sp`.

Tooltip on mobile device

Visit https://material.angular.io/components/tooltip/examples on a mobile device, and tap on 'Show tooltip'. See above for the result.

Follow the Material specs as much as possible.

Latest use of Material2, as deployed on https://material.angular.io

I don't think #3927 is related, as this issue is not following the Material specs.

donroyco:
This fixes #9039

### DIFF
--- a/src/lib/tooltip/_tooltip-theme.scss
+++ b/src/lib/tooltip/_tooltip-theme.scss
@@ -6,6 +6,7 @@
 // and the width/height appear to be off from the spec.
 $mat-tooltip-target-height: 22px;
 $mat-tooltip-font-size: 10px;
+$mat-tooltip-mobile-font-size: 14px;
 $mat-tooltip-vertical-padding: ($mat-tooltip-target-height - $mat-tooltip-font-size) / 2;
 
 @mixin mat-tooltip-theme($theme) {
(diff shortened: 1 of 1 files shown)


src/lib/tooltip/_tooltip-theme.scss
```diff
@@ -21,4 +23,10 @@ $mat-tooltip-vertical-padding: ($mat-tooltip-target-height - $mat-tooltip-font-s
     padding-top: $mat-tooltip-vertical-padding;
     padding-bottom: $mat-tooltip-vertical-padding;
   }
+
+  @media ($mat-xsmall) {
```
josephperrott: Rather than using a media query in the css, please use `BreakpointObserver` from the cdk in the `MatTooltop` class to apply a mobile class (something like `.mat-tooltip-mobile`).

donroyco: That sounds like a better way, yes. I will close this PR and make another one.

PR #8457, opened 2017-11-15
build(material): add bazel rules for all components

jelbourn:
Part of #8369 
cc @DevVersion 

The genrules for transforming the css output are unfortunate but should be able to be removed without a ton of effort.

PR #9096, opened 2017-12-22
build: add bazel rules for divider

CaerusKaru:
* Applies #8457 to divider


CaerusKaru:
cc @josephperrott this fixes the current Circle CI build error (sorry for not getting this in before the merge)

PR #9103, opened 2017-12-22
chore: add Bazel rule for MatDivider

josephperrott:


crisbeto:
Seems to have been added here already https://github.com/angular/material2/pull/9096.

PR #9104, opened 2017-12-22
fix(select): value update not being propagated when selected option is removed or added

### LINKED ISSUE #9038: mat-select model don't update after removing options programmatically
Bug

I expect to get the model updated and the change event called after remove selected options of the options list.

The model don't update and the change event is not called, but visually is updated.

1 - http://plnkr.co/edit/RNAeZbaNrtMQRrjC7IDc?p=preview ;
2 - Click on "Remove options";
3 - mat-select visually update;
4 - model dont update and change event is not called;

If I change the options for any reason, I expect to get the change events and the model to be updated in order to keep data consistent;

master

No

crisbeto:
Fixes the select not propagating its value back up to the value accessor if a selected option is added or removed. Previously this only happened the next time the user interacted with an option.

Fixes #9038.

### DIFF
--- a/src/lib/select/select.spec.ts
+++ b/src/lib/select/select.spec.ts
@@ -3459,6 +3459,84 @@ describe('MatSelect', () => {
     }));
 
   });
+
+  describe('value propagation when options are added/removed', () => {
+    let fixture: ComponentFixture<BasicSelectWithoutFormsMultiple>;
+    let testComponent: BasicSelectWithoutFormsMultiple;
+
+    beforeEach(fakeAsync(() => {
+      configureMatSelectTestingModule([BasicSelectWithoutFormsMultiple]);
+      fixture = TestBed.createComponent(BasicSelectWithoutFormsMultiple);
+      testComponent = fixture.componentInstance;
+      fixture.detectChanges();
+    }));
+
+    it('should propagate the changes when a selected option is removed', fakeAsync(() => {
+      testComponent.selectedFoods = ['steak-0', 'pizza-1'];
+      fixture.detectChanges();
+      flush();
+
+      expect(testComponent.select.value).toEqual(['steak-0', 'pizza-1']);
+      testComponent.selectionChange.calls.reset();
+
+      testComponent.foods.shift();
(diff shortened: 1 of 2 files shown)


jelbourn (dismissed):
LGTM

jelbourn:
@crisbeto it looks like this has the side-effect of causing the select to emit a `selectionChange` event for its initial value, which it didn't do before

(was caught by an internal test)

crisbeto:
@jelbourn do you have any more info on the test? I wasn't able to reproduce it in a test or in the demo app in a few scenarios (single selection, multiple selection and without Angular forms). Here's what my test looks like:


```ts
it('should not emit the `selectionChange` event for the preselected value', fakeAsync(() => {
  const fixture = TestBed.createComponent(BasicSelectWithoutFormsPreselected);

  fixture.detectChanges();
  flush();

  expect(fixture.componentInstance.selectionChangeSpy).not.toHaveBeenCalled();
}));

@Component({
  template: `
    <mat-form-field>
      <mat-select placeholder="Food" [(value)]="selectedFood"
        (selectionChange)="selectionChangeSpy()">
        <mat-option *ngFor="let food of foods" [value]="food.value">
          {{ food.viewValue }}
        </mat-option>
      </mat-select>
    </mat-form-field>
  `
})
class BasicSelectWithoutFormsPreselected {
  selectedFood = 'pizza-1';
  selectionChangeSpy = jasmine.createSpy('selectionChange spy');
  foods: any[] = [
    { value: 'steak-0', viewValue: 'Steak' },
    { value: 'pizza-1', viewValue: 'Pizza' },
  ];

  @ViewChild(MatSelect) select: MatSelect;
}
```

jelbourn:
In the test that was failing for this, it was a select inside of `mat-paginator`. The thing under test is listening to `page` events from the paginator, the first thing the test does is to set the `paginator.pageSize = 1`, then they manually emit `page` (for some reason), and then expect that only one event was handled, but two occur. The extra event seems to come from the select emitting something that it wasn't before.

mhmo91:
any updates here?

PR #9106, opened 2017-12-22
fix(selection-list): remove selected option from model value on destroy

crisbeto:
Fixes selected options that are removed from the DOM not being removed from the model.

Relates to #9104.

### DIFF
--- a/src/lib/list/selection-list.ts
+++ b/src/lib/list/selection-list.ts
@@ -178,6 +178,12 @@ export class MatListOption extends _MatListOptionMixinBase
   }
 
   ngOnDestroy(): void {
+    if (this.selected) {
+      // We have to delay this until the next tick in order
+      // to avoid changed after checked errors.
+      Promise.resolve().then(() => this.selected = false);
+    }
+
     this.selectionList._removeOptionFromList(this);
   }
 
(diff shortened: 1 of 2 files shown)


src/lib/list/selection-list.ts
```diff
@@ -178,6 +178,15 @@ export class MatListOption extends _MatListOptionMixinBase
   }
 
   ngOnDestroy(): void {
+    if (this.selected) {
+      // We have to delay this until the next tick in order
+      // to avoid changed after checked errors.
+      Promise.resolve().then(() => {
+        this._setSelected(false);
+        this.selectionList._reportValueChange();
```
DevVersion: How about moving that into a single line and just doing `this.selected = false`?

crisbeto: That feels a little side-effect-ey since we're basically using a property as a method.

DevVersion: Why is that? I think the setter's main intention is to update the internal state and reflect it within in the model. Otherwise the setter wouldn't be needed at all.

crisbeto: True, although I think of it mostly if you're setting it from the outside. That being said it does avoid having to duplicate some of the logic. I've switched it.

PR #9109, opened 2017-12-23
fix(tooltip): leaking TooltipComponent references

crisbeto:
Fixes references to each of the `TooltipComponent` instances (one per opened tooltip) created by `MatTooltip` not being cleaned up on close. The issue comes from the fact that we're passing in the `ViewContainerRef` which ends up retaining the references until it is destroyed, which is problematic, because the tooltip could be attached to a very long-lived view container (e.g. inside an app's navigation). In this case we can get away with not passing in the view container, because the tooltip component is internal and we manipulate it directly through the tooltip trigger anyway.

### DIFF
--- a/src/material/tooltip/tooltip.ts
+++ b/src/material/tooltip/tooltip.ts
@@ -37,7 +37,6 @@ import {
   OnDestroy,
   OnInit,
   Optional,
-  ViewContainerRef,
   ViewEncapsulation,
 } from '@angular/core';
 import {Observable, Subject} from 'rxjs';
@@ -244,7 +243,6 @@ export class MatTooltip implements OnDestroy, OnInit {
     private _overlay: Overlay,
     private _elementRef: ElementRef<HTMLElement>,
     private _scrollDispatcher: ScrollDispatcher,
-    private _viewContainerRef: ViewContainerRef,
     private _ngZone: NgZone,
     private _platform: Platform,
     private _ariaDescriber: AriaDescriber,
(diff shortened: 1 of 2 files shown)


jelbourn:
This could be problematic for testing because the view container determines when change detection happens for the tooltip content. Without specifying that, doing `fixture.detectChanges()` on the view that _physically_ contains the tooltip wouldn't necessarily update its contents. 

Why does the view container retain the reference to the tooltip component after it has been destroyed? Is this something that could be fixed in core?

crisbeto:
I was under the impression that `detectChanges` in tests just runs change detection over the entire app, but I'm not too familiar with the internals. Also I'm not sure why the references were being retained by the container, but that's where the heap snapshot was pointing. It's possible that there's something wrong in the overlay/portal that's preventing it from being collected, but I haven't been able to pinpoint it and we don't have issues in any of the other overlays as far as I can tell.

jelbourn:
I'll get someone from the core team to take a look next week

jelbourn:
@crisbeto think you could make a minimal repro of the `ViewContainer` maintaining a reference after destroying the `ComponentRef`?

jelbourn:
@crisbeto let's try this in ivy and see if the behavior changes, maybe some further debugging in the framework

andrewseguin:
I suspect this PR can be closed, Ivy seems to have resolved the issue of holding onto old references. See this comment for details: https://github.com/angular/components/issues/10397#issuecomment-643430667

PR #9113, opened 2017-12-24
feat(dialog): allow default dialog options to be configurable

### LINKED ISSUE #5664: Provide a way to configure global options for MdDialog
Proposal

Actually, you can configure a global default placeholder for inputs. It'd be good if we have this functionality in dialogs (maybe in snackbar also?) to configure some properties like `disableClose`, `height`, `width` and so on...

CDDelta:
Implements global dialog options.

Fixes #5664

### DIFF
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -29,7 +29,7 @@ export interface DialogPosition {
 
 /** Configurable options for `mat-dialog-container`. */
 export interface DialogGlobalOptions {
-  /** Whether dialogs should have a background. */
+  /** Whether dialogs have a background. */
   hasBackdrop?: boolean;
 
   /** Whether dialogs can be closed by user interaction. */
@@ -79,7 +79,7 @@ export class MatDialogConfig<D = any> {
   /** Custom class for the overlay pane. */
   panelClass?: string | string[] = '';
 
-  /** Whether the dialog should have a backdrop. */
+  /** Whether the dialog has a backdrop. */
   hasBackdrop?: boolean = true;
 
   /** Custom class for the backdrop, */
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ViewContainerRef} from '@angular/core';
+import {ViewContainerRef,InjectionToken} from '@angular/core';
 import {Direction} from '@angular/cdk/bidi';
 
 /** Valid ARIA roles for a dialog element. */
@@ -25,6 +25,36 @@ export interface DialogPosition {
 
   /** Override for the dialog's right position. */
   right?: string;
+}
+
+/** Configurable options for `mat-dialog-container`. */
+export interface DialogGlobalOptions {
+  /** Whether dialogs should have a background. */
+  hasBackdrop?: boolean;
+
+  /** Whether dialogs can be closed by user interaction. */
+  disableClose?: boolean;
+
+  /** The width of the dialogs. */
+  width?: string;
+
+  /** The height of the dialogs. */
+  height?: string;
+
+  /** The minimum width of the dialogs. */
+  minWidth?: string | number;
+
+  /** The minimum height of the dialogs. */
+  minHeight?: string | number;
+
+  /** The maximum width of the dialogs. */
+  maxWidth?: string | number;
+
+  /** The maximum height of the dialogs. */
+  maxHeight?: string | number;
+
+  /** Whether dialogs should focus the first focusable element on open. */
+  autoFocus?: boolean;
 }
 
 /**
@@ -49,7 +79,7 @@ export class MatDialogConfig<D = any> {
   /** Custom class for the overlay pane. */
   panelClass?: string | string[] = '';
 
-  /** Whether the dialog has a backdrop. */
+  /** Whether the dialog should have a backdrop. */
   hasBackdrop?: boolean = true;
 
   /** Custom class for the backdrop, */
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -27,36 +27,6 @@ export interface DialogPosition {
   right?: string;
 }
 
-/** Configurable options for `mat-dialog-container`. */
-export interface DialogGlobalOptions {
-  /** Whether dialogs have a background. */
-  hasBackdrop?: boolean;
-
-  /** Whether dialogs can be closed by user interaction. */
-  disableClose?: boolean;
-
-  /** The width of the dialogs. */
-  width?: string;
-
-  /** The height of the dialogs. */
-  height?: string;
-
-  /** The minimum width of the dialogs. */
-  minWidth?: string | number;
-
-  /** The minimum height of the dialogs. */
-  minHeight?: string | number;
-
-  /** The maximum width of the dialogs. */
-  maxWidth?: string | number;
-
-  /** The maximum height of the dialogs. */
-  maxHeight?: string | number;
-
-  /** Whether dialogs should focus the first focusable element on open. */
-  autoFocus?: boolean;
-}
-
 /**
  * Configuration for opening a modal dialog with the MatDialog service.
  */
--- a/src/lib/dialog/dialog-config.ts
+++ b/src/lib/dialog/dialog-config.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {ViewContainerRef,InjectionToken} from '@angular/core';
+import {ViewContainerRef} from '@angular/core';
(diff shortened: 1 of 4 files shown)


CDDelta:
I notice that there is also a CDK version of dialog. Would MatDialog, in the future, be built upon CDK dialog? Should this change also be applied to CDK dialog?

src/lib/dialog/dialog-config.ts
```diff
@@ -27,6 +27,36 @@ export interface DialogPosition {
   right?: string;
 }
 
+/** Configurable options for `mat-dialog-container`. */
+export interface DialogGlobalOptions {
```
crisbeto: You don't have to duplicate all of this code. The config is a `MatDialogConfig`.

src/lib/dialog/dialog.ts
```diff
@@ -96,6 +99,7 @@ export class MatDialog {
       private _injector: Injector,
       @Optional() location: Location,
       @Inject(MAT_DIALOG_SCROLL_STRATEGY) private _scrollStrategy,
+      @Optional() @Inject(MAT_DIALOG_GLOBAL_OPTIONS) private _options,
```
crisbeto: Instead of having this be an optional token, you can clean it up by having the `MatDialogModule` provide the token with a default value:

```
@NgModule({
  providers: [
    {provide: MAT_DIALOG_GLOBAL_OPTIONS, useValue: new MatDialogConfig()}
  ]
})
export class MatDialogModule {
}
```

Doing this also avoids having to add another argument to `_applyConfigDefaults` below.

src/lib/dialog/dialog.ts
```diff
@@ -33,13 +33,16 @@ import {of as observableOf} from 'rxjs/observable/of';
...
-import {MatDialogConfig} from './dialog-config';
+import {MatDialogConfig, DialogGlobalOptions} from './dialog-config';
 import {MatDialogContainer} from './dialog-container';
 import {MatDialogRef} from './dialog-ref';
 
 
 export const MAT_DIALOG_DATA = new InjectionToken<any>('MatDialogData');
 
+/** Injection token that can be used to specify global dialog options. */
+export const MAT_DIALOG_GLOBAL_OPTIONS =
```
crisbeto: This should be called something like `MAT_DIALOG_DEFAULT_OPTIONS`.

src/lib/dialog/dialog.spec.ts
```diff
@@ -965,6 +973,73 @@ describe('MatDialog', () => {
...
+      // Reset the previously configured testing module to be able set new providers.
+      // The testing module has been initialized in the root describe group for the dialogs.
+      TestBed.resetTestingModule();
+      TestBed.configureTestingModule({
+        imports: [MatDialogModule],
+        declarations: [PizzaMsg],
+        providers: [{ provide: MAT_DIALOG_GLOBAL_OPTIONS, useValue: globalConfig }]
+      });
+
+      dialog.open(PizzaMsg, {
```
crisbeto: You should move this into the individual tests. It's easier to follow, compared to passing in two arguments that look identical into a function.

src/lib/dialog/dialog.spec.ts
```diff
@@ -965,6 +973,73 @@ describe('MatDialog', () => {
       expect(container.hasAttribute('aria-labelledby')).toBe(false);
     }));
   });
+
+  describe('global dialog options', () => {
+    function createTestComponent(openConfig: MatDialogConfig, globalConfig: DialogGlobalOptions) {
+      // Reset the previously configured testing module to be able set new providers.
+      // The testing module has been initialized in the root describe group for the dialogs.
+      TestBed.resetTestingModule();
```
crisbeto: You should call `overlayContainer.ngOnDestroy` before resetting the testing module, otherwise the tests will leak overlay containers into other tests.

CDDelta:
@crisbeto do you know the reason why viewContainerFixture is destroyed before the tests end and how to fix it? I am new to this.

CDDelta:
@crisbeto how do I get angular to inject the test bed provided MAT_DIALOG_DEFAULT_OPTION into MatDIalog rather than the MatDialogModule one?

CDDelta:
@crisbeto ready for another round of reviews

src/lib/dialog/dialog.ts
```diff
@@ -40,6 +40,29 @@ import {MatDialogRef} from './dialog-ref';
 
 export const MAT_DIALOG_DATA = new InjectionToken<any>('MatDialogData');
 
+/** Injection token that can be used to specify default dialog options. */
+export const MAT_DIALOG_DEFAULT_OPTIONS =
+    new InjectionToken<MatDialogConfig>('mat-dialog-default-options');
+
+/** @docs-private */
+export const MAT_DIALOG_DEFAULT_OPTIONS_PROVIDER = {
```
crisbeto: A couple of notes on this one:
1. You don't have to define a provider object for it, you can just put it inline into the `providers`. We didn't do the same for the scroll strategy, because it has a bit more boilerplate.
2. Why are you providing these values here? If you use a `new MatDialogConfig()` it'll give you an object with the defaults provided already.

CDDelta: I tried providing `new MatDialogConfig()` before but when compiling Angular outputs this error: `Error during template compile of 'MatDialogModule' Function calls are not supported in decorators but 'MatDialogConfig' was called.` 

I also tried providing with `useClass: MatDialogConfig` but then providing a `MAT_DIALOG_DEFAULT_OPTIONS` in another module with `useValue` does not override the `MatDialogModule` defaults.

crisbeto: Sorry for the delay. `useClass` definitely won't work, because Angular will try to instantiate it as if it a service. What about doing something like this:

```ts
const defaultConfig = new MatDialogConfig();

@NgModule({
  ...
  providers: [
    {provide: MAT_DIALOG_DEFAULT_OPTIONS, useValue: defaultConfig}
  ]
})
export class MatDialogModule {}
```

CDDelta: That didn't work, still getting the same error. 

I was thinking maybe it would be better to provide `MAT_DIALOG_DEFAULT_OPTIONS` the way I am doing right now and just define `MatDialogConfig` without any preset fields. This way we can avoid duplicating the same values on both sides and also when a user does this :
```
let config = new MatDialogConfig(); 
config.hasBackdrop = false;
dialogs.open(Dialog, config);
```
The default `MatDialogConfig` fields will not override their own provided default options.

crisbeto: In that case it's best to go with what you had initially. I'm a little iffy about having the triple spread that you had before, but something like this should be fine for applying the defaults:

```ts
config = _applyConfigDefaults(this._defaultOptions || new MatDialogConfig(), config);
```

CDDelta: That looks way neater than what I had previously. Thank you for the help

CDDelta:
Seems like tests failed due to connection problems. Do I need to rerun them?

src/lib/dialog/dialog.spec.ts
```diff
@@ -1045,6 +1052,98 @@ describe('MatDialog with a parent MatDialog', () => {
...
+      minHeight: '50px',
+      maxWidth: '150px',
+      maxHeight: '150px',
+      autoFocus: false,
+    };
+
+    TestBed.configureTestingModule({
+      imports: [MatDialogModule, DialogTestModule],
+      providers: [
+        {provide: Location, useClass: SpyLocation},
```
crisbeto: You don't have to stub out the `Location` here. We have it in the tests above because we're testing something location-specific.

src/lib/dialog/dialog.spec.ts
```diff
@@ -1045,6 +1052,98 @@ describe('MatDialog with a parent MatDialog', () => {
...
+  });
+
+  beforeEach(() => {
+    viewContainerFixture = TestBed.createComponent(ComponentWithChildViewContainer);
+
+    viewContainerFixture.detectChanges();
+    testViewContainerRef = viewContainerFixture.componentInstance.childViewContainer;
+  });
+
+  it('should be set correctly', () => {
```
crisbeto: `should be set correctly` to something like `should use the provided defaults`.

src/lib/dialog/dialog.spec.ts
```diff
@@ -1045,6 +1052,97 @@ describe('MatDialog with a parent MatDialog', () => {
...
 });
 
+describe('MatDialog with default options', () => {
+  let dialog: MatDialog;
+  let overlayContainer: OverlayContainer;
+  let overlayContainerElement: HTMLElement;
+
+  let testViewContainerRef: ViewContainerRef;
+  let viewContainerFixture: ComponentFixture<ComponentWithChildViewContainer>;
+  let mockLocation: SpyLocation;
```
crisbeto: You can also get rid of the variable and the place where you're injecting it below.

CDDelta: Sorry, missed it

PR #9114, opened 2017-12-24
refactor(overlay): make it easier to override backdrop color

### LINKED ISSUE #7855: [overlay] 0.48 opacity of showing backdrop is unexpected
Proposal

#### Expected behavior

Providing a custom backdrop class with `background-color: red` would result in a fully opaque backdrop.

#### Current behavior

Be default, that backdrop will be at 0.48 opacity. To make it fully opaque, 

1. You must explicitly set `opacity: 1`
2. You must do so appropriately within `.cdk-overlay-backdrop-showing` which requires at least a little digging through the DOM or source

https://stackblitz.com/edit/angular-material2-issue-t9xu95?file=styles.scss

#### Relevant code

https://github.com/angular/material2/blob/26bbeb2cda3ab9d03c00dbb45e2317a48feed5d8/src/cdk/overlay/_overlay.scss#L8-L9

https://github.com/angular/material2/blob/26bbeb2cda3ab9d03c00dbb45e2317a48feed5d8/src/cdk/overlay/_overlay.scss#L64-L68

#### Motivation

Showing a backdrop which is darker than 0.48 opacity it slightly more troublesome than it could be.

Latest

crisbeto:
Previously the dark backdrop color was `rgba(0, 0, 0, 0.6)` to which we also added an opacity of `0.48`. This is hard to reason about, because we've got two stacking opacities that the consumer would have to multiply in order to figure out what the actual color is, in addition to being hard to override. These changes switch to animating the backdrop between 0 and 1, as well as setting the backdrop color to `rgba(0, 0, 0, 0.288)` (0.288 being 0.48 * 0.6).

Fixes #7855.

### DIFF
--- a/src/cdk/overlay/_overlay.scss
+++ b/src/cdk/overlay/_overlay.scss
@@ -6,7 +6,7 @@ $cdk-z-index-overlay: 1000;
 $cdk-z-index-overlay-backdrop: 1000;
 
 // Background color for all of the backdrops
-$cdk-overlay-dark-backdrop-background: rgba(0, 0, 0, 0.6);
+$cdk-overlay-dark-backdrop-background: rgba(0, 0, 0, 0.288);
 
 // Default backdrop animation is based on the Material Design swift-ease-out.
 $backdrop-animation-duration: 400ms !default;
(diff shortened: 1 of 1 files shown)


jelbourn:
Caretaker note: at time of this writing, it looks like no apps customize the backdrop color

PR #9115, opened 2017-12-24
refactor(option): move styles out of mat-core and into styleUrls

crisbeto:
Moves the styling for `MatOption` and `MatOptgroup` out of the `mat-core` mixin and into their components' respective `styleUrls`. This avoids the styles potentially being included multiple times if `mat-core` is used in different places and is consistent with all other components.


src/lib/core/BUILD.bazel
```diff
@@ -7,7 +7,7 @@ ng_module(
   name = "core",
   srcs = glob(["**/*.ts"], exclude=["**/*.spec.ts"]),
   module_name = "@angular/material/core",
-  assets = [":pseudo_checkbox_css"],
+  assets = [":pseudo_checkbox_css", ":option_css", ":optgroup_css"],
```
crisbeto: @jelbourn these changes seem to make Bazel happy, but I'm not sure whether it's the proper way to do it.

PR #9122, opened 2017-12-25
feat(list-key-manager): add support for horizontal directions

crisbeto:
Updates the `ListKeyManager` to add support for using ltr and rtl directions. This will allow us to switch over even more components to the key manager (e.g. chips, tabs, stepper) as well as support more use cases (e.g. moving selection with the left/right arrows on a closed select).

### DIFF
--- a/src/cdk/a11y/list-key-manager.ts
+++ b/src/cdk/a11y/list-key-manager.ts
@@ -9,7 +9,17 @@
 import {QueryList} from '@angular/core';
 import {Subject} from 'rxjs/Subject';
 import {Subscription} from 'rxjs/Subscription';
-import {UP_ARROW, DOWN_ARROW, TAB, A, Z, ZERO, NINE} from '@angular/cdk/keycodes';
+import {
+  UP_ARROW,
+  DOWN_ARROW,
+  LEFT_ARROW,
+  RIGHT_ARROW,
+  TAB,
+  A,
+  Z,
+  ZERO,
+  NINE,
+} from '@angular/cdk/keycodes';
 import {debounceTime} from 'rxjs/operators/debounceTime';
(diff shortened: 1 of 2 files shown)


src/cdk/a11y/list-key-manager.ts
```diff
@@ -58,6 +72,24 @@ export class ListKeyManager<T extends ListKeyManagerOption> {
     return this;
   }
 
+  /**
+   * Configures the directions in which the user can move the selection.
+   * @param directions One or more directions that the user can move the selection in.
+   */
+  withDirections(...directions: ListKeyManagerDirection[]): this {
```
jelbourn: What do you think of making this two method: `withVerticalOrientation()` and `withHorizontalOrientation(dir)`?

What did you have in mind for cases when it could have both turned on?

crisbeto: API-wise I'm fine with having separate methods, but it would make it slightly less convenient to deal with internally.

Regarding use cases, off the top of my head there are a couple:
1. Based on the a11y guidelines, horizontal tabs navigate using left/right arrows, whereas vertical tabs (which we use for the stepper) use both left/right and up/down.
2. Native `select` elements allow you to change the value while close both by pressing the up/down arrows, as well as left/right.

jelbourn: Hmmmm.........

My main reason or not liking everything in one method is that it lets users provide an invalid combination (even though it's caught). Another option would be to separate the orientation from the direction into separate methods, so you'd have something like
```ts
withOrientation(...orientations: Horizontal | Vertical) { }

withDirectionality(dir) { } // defaults to LTR
```

crisbeto:
Reworked based on the feedback @jelbourn.

PR #9126, opened 2017-12-26
fix(stepper): completed binding not being considered when moving from a step without a stepControl

### LINKED ISSUE #8110: Linear stepper without a form
Proposal

Assign a condition for a step to be completed manually (form-independent)

No option to mark step as completed without a form

Fit more use cases

for example:

`<mat-horizontal-stepper formArrayName="formArray" linear [completed]="items.length">`

crisbeto:
Currently we only consider a step's validation state when determining whether the user can move forward in a linear stepper, however this means that there's no way to block navigation without using Angular forms. These changes switch the logic so it considers the `completed` binding, if there is no `stepControl`.

Fixes #8110.


pavan02:
@crisbeto This doesn't seem to work with selectedIndex.

PR #9127, opened 2017-12-26
fix(stepper): throw when out-of-bounds value is assigned to selectedIndex

crisbeto:
Throws better error when an out-of-bounds value is assigned to the `selectedIndex`.

### DIFF
--- a/src/cdk/stepper/stepper.ts
+++ b/src/cdk/stepper/stepper.ts
@@ -155,6 +155,11 @@ export class CdkStepper implements OnDestroy {
   get selectedIndex() { return this._selectedIndex; }
   set selectedIndex(index: number) {
     if (this._steps) {
+      // Ensure that the index can't be out of bounds.
+      if (index < 0 || index > this._steps.length - 1) {
+        throw Error('cdkStepper: Cannot assign out-of-bounds value to `selectedIndex`.');
+      }
+
       if (this._anyControlsInvalidOrPending(index) || index < this._selectedIndex &&
           !this._steps.toArray()[index].editable) {
         // remove focus from clicked step header if the step is not able to be selected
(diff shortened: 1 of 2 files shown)


mmalerba:
Shouldn't we actually throw an error in this case? That way the user knows what they're doing wrong rather than trying to track down strange behavior. Maybe just throw a custom error instead of whatever is thrown now that explains better?

crisbeto:
That also makes sense, but I went with this approach, because it’s consistent with what we do in the tabs.

mmalerba:
@andrewseguin do you think the tabs behavior should change or this should match what tabs currently does?

andrewseguin:
As a developer I'd probably want to see an error rather than the component fixing itself. I think though that this fix is in favor of the end-user who runs across an edge case where the developer made a mistake.

I vote for changing both tabs and this to throw an error and be clear to the developer that their logic isn't right

crisbeto:
Switched it to throw an error @mmalerba @andrewseguin.

crisbeto:
@andrewseguin I'm looking at adding similar changes to the tabs and there's a case that we didn't consider: the tabs/steps could be added asynchronously in which case the consumer starts getting into race conditions instead of having the component default to zero.

andrewseguin:
Hmm that's an interesting point, so we expect that the selected index stays as out of bounds in case the steps/tabs are updated later?

crisbeto:
No, it's more like we set it to zero when it's out of bounds, rather than throw an error, because we can't know whether any new elements are going to come in.

andrewseguin:
In that case, should we expect the developer to still be accurate and change the `selectedIndex` when the async call is complete, and otherwise keep it at 0?

crisbeto:
I suppose, but that ends up being a lot more inconvenient.

andrewseguin:
Probably needs more discussion but this is fine if you want this behavior for now

PR #9132, opened 2017-12-27
fix(tabs): maintain selected tab when new tabs are added or removed

### LINKED ISSUE #7738: [Tabs] Selection changes when tabs added/removed

Bug?

Adding or removing tabs shouldn't change which tab is selected

Adding or removing tabs causes selection to change

Go to [this plunker](https://plnkr.co/edit/RoXCIh?p=preview), and select one of the later tabs (5 or 6).  Click the "Toggle" button at the top, and watch which tab is selected.

The user didn't click on a different tab, so they don't expect the selected tab to change.

----

I believe the selected tab is based on an index / offset into the tab collection, so when `*ngIf` goes false, the collection shrinks and the same offset now points to a different tab.  This might mean that you can't use `*ngIf` to hide tabs, which would mean that a) this needs to be documented, and b) an alternative property should be provided to signify "don't show this tab, but keep it in the collection so that its size doesn't change and mess up the selection offset".  It would be nice if selection was instead based on a generated ID for each tab, but that might not be practical.

crisbeto:
Maintains the `selectedIndex` of the current tab when a new tab is added or removed. Previously, changing the amount of tabs would shift the array, causing a different tab to be selected.

Fixes #7738.

### DIFF
--- a/src/lib/tabs/tab-group.spec.ts
+++ b/src/lib/tabs/tab-group.spec.ts
@@ -282,14 +282,17 @@ describe('MatTabGroup', () => {
         expect(tabs[3].origin).toBeGreaterThanOrEqual(0);
 
         // Add a new tab in the beginning and select it, expect an origin < than 0 (animate left)
-        fixture.componentInstance.tabs.push({label: 'New tab', content: 'to left of index'});
         fixture.componentInstance.selectedIndex = 0;
+        fixture.detectChanges();
+        tick();
+
+        fixture.componentInstance.tabs.push({label: 'New tab', content: 'to left of index'});
         fixture.detectChanges();
         tick();
 
         tabs = component._tabs.toArray();
         expect(tabs[0].origin).toBeLessThan(0);
-      }));
+    }));
 
 
     it('should update selected index if the last tab removed while selected', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


andrewseguin:
Is there any concern about a race when changing `selectedIndex` and the set of tabs at the same time? E.g. if tab two is selected, but then add a new tab in the front and set `selectedIndex` to 0, will the change in tabs cause `indexToSelect` to be set to `2` rather than the intended `0`?

crisbeto:
Sorry @andrewseguin, I just got around to rebasing this PR and saw your comment from a while ago. I think that there shouldn't be an issue with race conditions since the tab changes event will be fired after the `selectedIndex` has been passed in.

shahraship:
This fix seems to have broken existing expected functionality.

On this page: https://material.angular.io/components/tabs/examples
There is one example: Tag group with dynamically changing tabs
If you follow the example on [stackblitz](https://stackblitz.com/angular/xbrrxkmrkbv?file=app%2Ftab-group-dynamic-example.ts) you'll see that "Select tab after adding" checkbox functionality doesn't work anymore.

Let me know if you want me to open a new bug for this.

garrettH3S:
This still doesn't seem to be working ( https://stackblitz.com/angular/rkdrmvvyyjn?file=app%2Ftab-group-dynamic-example.ts v6.4.1). The focus after open doesn't seem to function correctly.

I fix this wrapping the index setter method in a setTimeout.

```javascript

addTab(selectAfterAdding: boolean) {
    this.tabs.push('New');
    if (selectAfterAdding) {
      setTimeout( () => {
        this.selected.setValue(this.tabs.length - 1);
      },10)
    }
  }
```

PR #9133, opened 2017-12-27
fix(datepicker): not resetting when detached externally

crisbeto:
Fixes not being able to reopen a datepicker if it's detached via the `OverlayRef` (e.g. by a `CloseScrollStrategy`).


jelbourn:
Changing this to minor release since it doesn't cleanly cherry-pick into the patch branch

Issue #8953, opened 2017-12-12
fadeInItems & fadeInContent doesn't work


levgaas:
#### Bug, feature request, or proposal:
Mat menu-content opacity is shorter than in beta versions.

#### What is the expected behavior?
opacity transition duration as we had in betas.

#### What is the current behavior?
Menu is shown almost without opacity transition. And I can see how content is transformed (stretched)

#### What are the steps to reproduce?
https://run.stackblitz.com/api/angular/v1/

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular v5.1. Material v.5.1

julianobrasil:
Your stackblitz link is broken.

levgaas:
updated

willshowell:
The animation was changed in https://github.com/angular/material2/pull/5361 to better match the spec examples

levgaas:
@willshowell could you elaborate. Because specs links say different:
[Material Components For Web](https://material-components-web.appspot.com/simple-menu.html)
[Material DEsign](https://material.io/guidelines/components/menus.html#menus-usage)

willshowell:
I think they follow the material.io guidelines, not the material-components implementation.

While I do believe the new animation _better_ matches the spec, the animation [here](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0Bzhp5Z4wHba3ZjA5T0M5eS0xRUU/component-menus-usage-appbar-dropdown-xhdpi-004.mp4) does seem to be different from the current implementation. It seems like the content opacity animation needs to be delayed slightly more...

https://github.com/angular/material2/blob/2fb4d618725b6fddc1a94e5e85ea80577377f1eb/src/lib/menu/menu-animations.ts#L55-L65

crisbeto:
If you're running Angular 5.1, the opacity won't be animated due to an issue with `@angular/animations`. See https://github.com/angular/angular/issues/20946.

levgaas:
@crisbeto, cool. Waiting for a fix then

PR #8371, opened 2017-11-11
refactor(menu): simplify animation

crisbeto:
Simplifies the menu animation by using keyframes to do the opening transition, rather than chaining a couple of animation events.

PR #9134, opened 2017-12-27
fix(expansion-panel,menu,select): nested animations not working

### LINKED ISSUE #8814: missing rotation animation of the arrow in mat-expansion-panel-header

Bug

Rotation animation of the arrow in mat-expansion-panel-header.

Arrow on the right of the mat-expansion-panel-header doesn't animates.

http://plnkr.co/edit/zIEeYb?p=preview

Animation was working in previous version.

all including Angular Material 5.0.0-rc.1

crisbeto:
Fixes issues with the expansion panel, menu and select animations where the child animations were being blocked by the parent ones.

**Note:** the menu animations needed a bit more refactoring since the old approach wasn't very idiomatic and made it harder to run the parallel animations.

Fixes #8814, #8953.


crisbeto:
Caretaker note: this includes the changes from https://github.com/angular/material2/pull/8371 which had some sync issues.

were necessary to make some of

jelbourn:
@crisbeto would you be able to change the PR to avoid renaming any of the internal methods on menu (the animation ones specifically)? One team in Google forked menu-trigger, so they were left with calls to the internal menu methods. They're going to work on changing that, but in the meantime avoiding any changes would unblock this PR.

crisbeto:
I've put back the old animation methods @jelbourn.

andrewseguin:
For some reason, this causes issues internally for tests because they can no longer click menu items as soon as they appear on the page. 

The reason for this is due to [the `pointer-events` rules in the styles](https://github.com/angular/material2/blob/master/src/lib/menu/menu.scss#L22). To resolve this, we can reduce our test failures by making this style only affect the intended menu items (ones that open nested menus). 

The fix is to add the selector `.mat-menu-item-submenu-trigger` to the CSS rule:

```css
&.ng-animating .mat-menu-item-submenu-trigger {
  pointer-events: none;
}
```

crisbeto:
Done @andrewseguin.

PR #9154, opened 2017-12-28
fix(snack-bar): indicate in afterDismissed whether dismissal was a result of an action

### LINKED ISSUE #9147: Know if a snackbar is dismissed by action or not
#### Proposal:

When the snackbar is dismissed by click on the action button, the observable onAction() is called, but also the observable afterDismissed().
When the snackbar is dismissed by the timer or the function dismiss(), there are just the afterDismissed() observable called.
There are no option to know if the dismissed has origin by the action or other.

The use case is to have a delete message with a cancel action, if the dismiss hasn't origin by the action, we must do the delete on server, otherwise restore in the interface the delete item. 

We can add to boolean result in the afterDismissed : 
```typecript
    let snackBarRef = this.snackBar.open('The item has been removed', 'Cancel');
    snackBarRef.onAction().subscribe(() => {
      this.undoRemove();
    });
    snackBarRef.afterDismissed().subscribe((action) => {
      if (!action)
         this.doRemove();
    });
```
Or add a new observable : 
```typecript
    let snackBarRef = this.snackBar.open('The item has been removed', 'Cancel');
    snackBarRef.onAction().subscribe(() => {
      this.undoRemove();
    });
    snackBarRef.onDismiss().subscribe(() => {
       this.doRemove();
    });
```

The observable afterDismissed is called everytime with or without action.

crisbeto:
Adds a boolean to the `MatSnackBarRef.afterDismissed` result that indicates whether the dismissal came as a result of the user pressing the action button.

Fixes #9147.

### DIFF
--- a/src/lib/snack-bar/snack-bar-ref.ts
+++ b/src/lib/snack-bar/snack-bar-ref.ts
@@ -11,6 +11,12 @@ import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 import {MatSnackBarContainer} from './snack-bar-container';
 
+/** Event that is emitted when a snack bar is dismissed. */
+export interface MatSnackBarDismiss {
+  /** Whether the snack bar was dismissed using the action button. */
+  dismissedByAction: boolean;
+}
+
 /**
  * Reference to a snack bar dispatched from the snack bar service.
  */
@@ -24,8 +30,8 @@ export class MatSnackBarRef<T> {
    */
   containerInstance: MatSnackBarContainer;
 
-  /** Subject for notifying the user that the snack bar has closed. */
-  private _afterClosed = new Subject<void>();
+  /** Subject for notifying the user that the snack bar has been dismissed. */
+  private _afterDismissed = new Subject<MatSnackBarDismiss>();
 
   /** Subject for notifying the user that the snack bar has opened and appeared. */
   private _afterOpened = new Subject<void>();
@@ -39,6 +45,9 @@ export class MatSnackBarRef<T> {
    */
   private _durationTimeoutId: number;
 
+  /** Whether the snack bar was dismissed using the action button. */
+  private _dismissedByAction = false;
+
   constructor(containerInstance: MatSnackBarContainer,
               private _overlayRef: OverlayRef) {
     this.containerInstance = containerInstance;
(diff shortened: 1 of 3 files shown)


src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -84,13 +88,18 @@ export class MatSnackBarRef<T> {
       this._onAction.complete();
     }
 
-    this._afterClosed.next();
-    this._afterClosed.complete();
+    this._afterDismissed.next(this._dismissedWithAction);
```
crisbeto: @josephperrott I'm also open to having a `reason` parameter on the `dismiss` method. I didn't go with it now, because it would mean that we'd need some magical string (e.g. `'action'`), because the consumer doesn't usually have access to the action button on the `SimpleSnackBar`.

src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -84,13 +88,18 @@ export class MatSnackBarRef<T> {
...
 
-  /** Gets an observable that is notified when the snack bar is finished closing. */
-  afterDismissed(): Observable<void> {
-    return this._afterClosed.asObservable();
+  /**
+   * Gets an observable that is notified when the snack bar is finished closing.
+   * The stream emits a boolean that indicates whether the snack bar was dismissed
+   * using the action button.
+   */
+  afterDismissed(): Observable<boolean> {
```
jelbourn: Instead of making this a boolean stream, I'd rather add an event object so that we could add more information to it in the future.

crisbeto:

src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -11,6 +11,12 @@ import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 import {MatSnackBarContainer} from './snack-bar-container';
 
+/** Event that is emitted when a snack bar is dismissed. */
+export interface MatSnackBarAfterDismiss {
```
jelbourn: How about just `MatSnackBarDismiss` or `MatSnackBarDismissal`? The "after" in the stream name makes sense because it relates to when the event is emitted, but there's nothing about the event object that's specific to it being at that time

crisbeto: Renamed to `MatSnackBarDismiss`.

src/lib/snack-bar/snack-bar-ref.ts
```diff
@@ -11,6 +11,12 @@ import {Observable} from 'rxjs/Observable';
 import {Subject} from 'rxjs/Subject';
 import {MatSnackBarContainer} from './snack-bar-container';
 
+/** Event that is emitted when a snack bar is dismissed. */
+export interface MatSnackBarDismiss {
+  /** Whether the snack bar was dismissied using the action button. */
+  dismissedByAction: boolean;
```
josephperrott: Should this be `closedByAction` since the method that would cause this to be true is `closeWithAction()`?

crisbeto: That's true, but the method that returns is called `afterDismissed`. Either way it would be slightly confusing.

josephperrott: Maybe we should mark one of the terms as deprecated to fix it long term?  I am good with either wording being removed, just as long as we create consistency.

crisbeto: My vote is for removing the "dismissed" wording and calling it "close" (also renaming the `dismiss` method). That way it's consistent with the dialog as well.

josephperrott: SGTM

crisbeto:
Done @josephperrott.

crisbeto:
Switched back to the `dismiss` terminology and deprecated `closeWithAction`. Can you take another look @josephperrott?

PR #9159, opened 2017-12-29
fix(select): handle async changes to the option label

### LINKED ISSUE #7923: mat-select not pre selecting a value when mat-option has async content
Bug

When the ngModel of a mat-select is populated on page load the mat-select should have that option selected.

mat-select does not show the current selected option on page load *if* a mat-option has some async behaviour, in my case the translate pipe:
```
        <mat-option *ngFor="let option of options" [value]="option.value">
          {{ option.label | translate }}
        </mat-option>
```
The current selected option is shown the moment the mat-select gets focus.

Please have a look at the following StackBlitz example:
https://stackblitz.com/edit/angular-material2-issue-async-mat-option-label?file=app%2Fpet-select-async%2Fpet-select-async.component.ts

material beta 12
angular 4.4.3

crisbeto:
Currently `mat-select` doesn't react to changes in the label of its options, which is problematic, because the option label might be populated at a later point by a pipe or it might have changed while the select is closed. These changes add a `Subject` to the `MatOption` that will emit whenever the label changes and allows the select to react accordingly.

Fixes #7923.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -833,7 +833,6 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit, 
   private _initKeyManager() {
     this._keyManager = new ActiveDescendantKeyManager<MatOption>(this.options).withTypeAhead();
     this._keyManager.tabOut.pipe(takeUntil(this._destroy)).subscribe(() => this.close());
-
     this._keyManager.change.pipe(takeUntil(this._destroy)).subscribe(() => {
       if (this._panelOpen && this.panel) {
         this._scrollActiveOptionIntoView();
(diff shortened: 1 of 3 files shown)


src/lib/select/select.ts
```diff
@@ -829,16 +829,24 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit,
...
+      .pipe(takeUntil(changedOrDestroyed), filter(event => event.isUserInput))
+      .subscribe(event => {
+        this._onSelect(event.source);
+
+        if (!this.multiple) {
+          this.close();
+        }
+      });
+
+    merge(...this.options.map(option => option._stateChanges))
```
jelbourn: Add a comment here that explains what this is for?

crisbeto:

jelbourn:
@crisbeto passes presubmit, just needs rebase

crisbeto:
Rebased.

PR #9163, opened 2017-12-29
fix(list): incorrect padding for list-items with avatars, icons

### LINKED ISSUE #9156: Selection lists list item misaligned
Bug

When using selection-type list, expect that left alignment is the same as a non selection-type list, without an additional 16px of padding.

When using selection-type list, .mat-list-item-content-reverse is applied (flex-direction: row-reverse), but there is an extra left-padding of 16px that is left over. Without the reversal of flex-direction, the padding is correctly aligned.

https://stackblitz.com/angular/jdyoyxpmkng
Also present in the docs (https://material.angular.io/components/list/overview)

Lists should all align for consistency.

@angular/core 5.1.2 @angular/material 5.0.0-rc.3

Potential fix is that `.mat-list-text {padding: 0 16px}` but when under `.mat-list-item-content-reverse .mat-list-text { padding: 0 16px 0 0;}`

Screenshot: https://snag.gy/WLCghp.jpg

DevVersion:
Currently list-items with an avatar, icon or checkbox have a both-sided margin for the text of 16px, while there is still a padding of 16px for the `mat-list-item-content` of 16px. This means that there is a 32px padding, that is not valid per specs.

The padding for the mat-list-text should be just on the side of the secondary item (e.g. avatar, icon or checkbox), and the other side should have no padding, because the `mat-list-item-content` already has a padding for that one.

image

image

Fixes #9156

### DIFF
--- a/src/demo-app/list/list-demo.html
+++ b/src/demo-app/list/list-demo.html
@@ -111,7 +111,7 @@
                         (change)="changeEventCount = changeEventCount + 1">
       <h3 mat-subheader>Groceries</h3>
 
-      <mat-list-option value="bananas">Bananas</mat-list-option>
+      <mat-list-option value="bananas" checkboxPosition="before">Bananas</mat-list-option>
       <mat-list-option selected value="oranges">Oranges</mat-list-option>
       <mat-list-option value="apples">Apples</mat-list-option>
       <mat-list-option value="strawberries">Strawberries</mat-list-option>
--- a/src/lib/list/list.scss
+++ b/src/lib/list/list.scss
@@ -92,14 +92,37 @@ $mat-list-item-inset-divider-offset: 72px;
 
   .mat-list-text {
     @include mat-line-wrapper-base();
-    padding: 0 $mat-list-side-padding;
 
-    // We only want to override the padding if there isn't
-    // an avatar or icon before the element. Since the ripple
-    // will always be the first child in the container, we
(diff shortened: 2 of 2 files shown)


andrewseguin:
Turns out this caused an issue that was caught in screenshot diffs internally.

If there is no avatar or icon, the `mat-list-text` should have no padding due to the rule:
```
&:nth-child(2) {
  padding: 0;
}
```

However, it is now always overridden to have padding again by the following CSS:
```


  // There can be a secondary item (e.g. avatar-icon, checkbox) at the start of the
  // list-item. This means that there should be a padding for the mat-list-text on the start-side.
  .mat-list-item-content .mat-list-text {
    padding-left: $mat-list-side-padding;

    [dir='rtl'] & {
      padding-right: $mat-list-side-padding;
      padding-left: 0;
    }
  }

  // Reversed content is mainly used by the MatSelectionList for displaying the checkbox at the
  // end of the list option. Since there is a secondary item (checkbox) at the end of the
  // option, there needs to be a padding for the mat-list-text on the end-side.
  .mat-list-item-content-reverse .mat-list-text {
    padding-left: 0;
    padding-right: $mat-list-side-padding;

    [dir='rtl'] & {
      padding-right: 0;
      padding-left: $mat-list-side-padding;
    }
  }
```

A likely fix would be to just define those styles in a different order

PR #9164, opened 2017-12-29
fix(list-key-manager): maintain selected index when amount of items changes

crisbeto:
Currently the `activeItemIndex` and the `activeItem` in the `ListKeyManager` will stay the same if the amount of items changes. This is an issue, because they'll become out of sync and focus could potentially go to a disabled item. With the following changes the key manager will ensure that the `activeItemIndex` and the `activeItem` are in sync and point to the same item.


jelbourn:
@crisbeto passes presubmit, just needs rebase

crisbeto:
Rebased.

PR #9173, opened 2017-12-30
feat(stepper): allow for orientation to be changed dynamically

crisbeto:
* Turns the `MatStepper` into a proper component that allows consumers to switch between `horizontal` and `vertical` dynamically, allowing for use cases like having a different layout depending on the screen size.
* Combines the `mat-vertical-stepper` and `mat-horizontal-stepper` templates into a single file to avoid all the code duplication.

Relates to #7700.


mmalerba:
I prefer this approach, but @jelbourn is the one who pushed for separate horizontal and vertical components, so I'd like to hear what he thinks

src/demo-app/stepper/stepper-demo.html
```diff
@@ -40,7 +42,7 @@ <h3>Linear Vertical Stepper Demo using a single form</h3>
         <button mat-button>Done</button>
       </div>
     </mat-step>
-  </mat-vertical-stepper>
+  </mat-stepper>
 </form>
 
 <h3>Linear Horizontal Stepper Demo using a different form for each step</h3>
```
kbo4sho: The rest of this file should be updated with the new stepper.

jelbourn:
So, my reason for wanting to make separate steppers was that there can be more than just the two vertical and horizontal variants; there are a couple other layouts some Google applications use. Baking them all into one would balloon the complexity.

What if we did some kind of dynamic stepper that delegates between them for you while still maintaining the separate implementations as standalone components?

src/demo-app/stepper/stepper-demo.html
crisbeto: Not necessarily. The separate horizontal/vertical steppers are still valid use cases.

crisbeto:
That would probably end up being an `ngSwitch` between the different variants anyway (which is what it does here) @jelbourn. I went with this approach, because at least it allows us to re-use some of the templating for the individual steps.

src/demo-app/stepper/stepper-demo.html
kbo4sho: Got it, that makes sense.

mmalerba:
needs rebase

jelbourn:
Thinking about this more, there are upcoming Angular features that would make this easier / better optimizable. We might want to wait for that (probably v7) before committing to this.

danielpiedra:
@jelbourn is still the idea to wait until angular v7?

jelbourn:
It's not set in stone, but there are potential features on the Angular roadmap that would make this a lot easier to do (and in such a way that it can be optimized better).

danielpiedra:
reasonable enough, we wait. thanks.

paulogr:
What about change the horizontal to work like in guidelines on mobile?

https://material.io/guidelines/components/steppers.html#steppers-types-of-steps

Is that the right approach?

GuerricPhalippou:
Looks like not :(

nomanbiniqbal:
any news ?

eduardoschonhofen:
Hi! Any news? This would be really useful,since using *ngIf to switch between horizontal and vertical make duplicates of the components in the steps.

ManuelGraf:
come on guys. This has been a problem for a super long time and is still not adressed? its like you dont want people to use angular material.... will this be addressed in 8.0 beta at least?
wouldnt it be possible to just let us use templateoutet and templateref? those would suffice to fix the stepper issues...

grant77:
I literally just finished writing this and, yes, it is a bit hacky.  This has not been tested at all.  Use at own risk.

```
import { Directionality } from '@angular/cdk/bidi';
import { CdkStep, StepperSelectionEvent } from '@angular/cdk/stepper';
import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ContentChildren, ElementRef, EventEmitter, forwardRef, Inject, Input, Optional, Output, QueryList, ViewChildren } from '@angular/core';
import { MatStep, MatStepper } from '@angular/material';
import { DOCUMENT } from '@angular/platform-browser';

const MAT_STEPPER_PROXY_FACTORY_PROVIDER = {
    provide: MatStepper,
    deps: [forwardRef(() => StepperComponent), [new Optional(), Directionality], ChangeDetectorRef, [new Inject(DOCUMENT)]],
    useFactory: MAT_STEPPER_PROXY_FACTORY
};

export function MAT_STEPPER_PROXY_FACTORY(component: StepperComponent, directionality: Directionality,
    changeDetectorRef: ChangeDetectorRef, docuement: Document) {
    // We create a fake stepper primarily so we can generate a proxy from it.  The fake one, however, is used until 
    // our view is initialized.  The reason we need a proxy is so we can toggle between our 2 steppers 
    // (vertical and horizontal) depending on  our "orientation" property.  Probably a good idea to include a polyfill 
    // for the Proxy class: https://github.com/GoogleChrome/proxy-polyfill.

    const elementRef = new ElementRef(document.createElement('mat-horizontal-stepper'));
    const stepper = new MatStepper(directionality, changeDetectorRef, elementRef, document);
    return new Proxy(stepper, {
        get: (target, property) => Reflect.get(component.stepper || target, property),
        set: (target, property, value) => Reflect.set(component.stepper || target, property, value)
    });
}

@Component({
    selector: 'app-stepper',
    // templateUrl: './stepper.component.html',
    // styleUrls: ['./stepper.component.scss'],
    providers: [MAT_STEPPER_PROXY_FACTORY_PROVIDER],
    template: `
<ng-container [ngSwitch]="orientation">
    <mat-horizontal-stepper *ngSwitchCase="'horizontal'"
                            [labelPosition]="labelPosition"
                            [linear]="linear"
                            [selected]="selected"
                            [selectedIndex]="selectedIndex"
                            (animationDone)="animationDone.emit($event)"
                            (selectionChange)="selectionChange.emit($event)">
    </mat-horizontal-stepper>


    <mat-vertical-stepper *ngSwitchDefault
                            [linear]="linear"
                            [selected]="selected"
                            [selectedIndex]="selectedIndex"
                            (animationDone)="animationDone.emit($event)"
                            (selectionChange)="selectionChange.emit($event)">
    </mat-vertical-stepper>
</ng-container>
`
})
export class StepperComponent {
    // public properties
    @Input() labelPosition?: 'bottom' | 'end';
    @Input() linear?: boolean;
    @Input() orientation?: 'horizontal' | 'vertical';
    @Input() selected?: CdkStep;
    @Input() selectedIndex?: number;

    // public events
    @Output() animationDone = new EventEmitter<void>();
    @Output() selectionChange = new EventEmitter<StepperSelectionEvent>();

    // internal properties
    @ViewChildren(MatStepper) stepperList!: QueryList<MatStepper>;
    @ContentChildren(MatStep) steps!: QueryList<MatStep>;
    get stepper(): MatStepper { return this.stepperList && this.stepperList.first; }

    // private properties
    private lastSelectedIndex?: number;
    private needsFocus = false;
    
    // public methods
    constructor(private changeDetectorRef: ChangeDetectorRef) { }
    ngAfterViewInit() {
        this.reset();
        this.stepperList.changes.subscribe(() => this.reset());
        this.selectionChange.subscribe((e: StepperSelectionEvent) => this.lastSelectedIndex = e.selectedIndex);
    }
    ngAfterViewChecked() {
        if (this.needsFocus) {
            this.needsFocus = false;
            const { _elementRef, _keyManager, selectedIndex } = <any>this.stepper;
            _elementRef.nativeElement.focus();
            _keyManager.setActiveItem(selectedIndex);
        }
    }

    // private properties
    private reset() {
        const { stepper, steps, changeDetectorRef, lastSelectedIndex } = this;
        stepper.steps.reset(steps.toArray());
        stepper.steps.notifyOnChanges();
        if (lastSelectedIndex) {
            stepper.selectedIndex = lastSelectedIndex;
        }

        Promise.resolve().then(() => {
            this.needsFocus = true;
            changeDetectorRef.markForCheck();
        });
    }
}

```

jrcasso:
We're on Angular 8 now. I think it's time to revisit this functionality.

Right now, people are wanting different stepper layouts depending on device orientation or screen size. But the child component data gets lost during the transition with a simple switch between the two.

Frankly, I think it's odd that you went this direction in the first place. It may have been a complex component, but to think that this wouldn't be a desired behavior for the component on mobile devices is shortsighted. Optimization should have been secondary to functionality expectations.

davideas:
@grant77, currently we receive this error when using it:
`No provider for CdkStepper!`.
Do we need some provider declaration?

And now DOCUMENT is now: `import { DOCUMENT } from '@angular/common';`

grant77:
@davideas, Try this:

```
const MAT_STEPPER_PROXY_FACTORY_PROVIDER = {
  provide: MatStepper,
  deps: [forwardRef(() => StepperComponent), [new Optional(), Directionality], ChangeDetectorRef, [new Inject(DOCUMENT)]],
  useFactory: MAT_STEPPER_PROXY_FACTORY
};

const CDK_STEPPER_PROXY_FACTORY_PROVIDER = { ...MAT_STEPPER_PROXY_FACTORY_PROVIDER, provide: CdkStepper }

export function MAT_STEPPER_PROXY_FACTORY(component: StepperComponent, directionality: Directionality,
  changeDetectorRef: ChangeDetectorRef, docuement: Document) {
  // We create a fake stepper primarily so we can generate a proxy from it.  The fake one, however, is used until 
  // our view is initialized.  The reason we need a proxy is so we can toggle between our 2 steppers 
  // (vertical and horizontal) depending on  our "orientation" property.  Probably a good idea to include a polyfill 
  // for the Proxy class: https://github.com/GoogleChrome/proxy-polyfill.

  const elementRef = new ElementRef(document.createElement('mat-horizontal-stepper'));
  const stepper = new MatStepper(directionality, changeDetectorRef, elementRef, document);
  return new Proxy(stepper, {
    get: (target, property) => Reflect.get(component.stepper || target, property),
    set: (target, property, value) => Reflect.set(component.stepper || target, property, value)
  });
}

@Component({
  selector: 'app-stepper',
  // templateUrl: './stepper.component.html',
  // styleUrls: ['./stepper.component.scss'],
  providers: [
    MAT_STEPPER_PROXY_FACTORY_PROVIDER,
    CDK_STEPPER_PROXY_FACTORY_PROVIDER
  ],
...
```

davideas:
@grant77 Thank you, it works really great! I created a gist with some more fixes and a nice feature:

**Angular 8.x Responsive Stepper with headers disable feature**

[responsive-stepper.component.ts](https://gist.github.com/davideas/3534a5c2567d6668013426ab961990f6)

hanktrizz:
Hi, it's v10 now. I'm just wondering what the status of this feature is in the pipeline? It'll be nice to know if it will be supported or not at all as it'll provide some closure rather than waiting for updates to this issue endlessly.

maykon-oliveira:
My soul longs for it, updates?

ChristianHardy:
Any solution?

crisbeto:
We've decided to revisit this feature, however rebasing this PR is going to take a while and there are better ways of achieving the same result now. I'm closing it in favor of https://github.com/angular/components/pull/22139.

Issue #4179, opened 2017-04-20
Proposal: Provide public access to a button's ripple


willshowell:
#### Bug, feature request, or proposal:
Proposal

#### What is the proposed behavior?
Public access to the MdRipple within a button, nav-list-item, etc

#### What is the use-case or motivation?
I'd like to programmatically trigger a button's ripple to call attention to it. Under [feature discovery](https://material.io/guidelines/growth-communications/feature-discovery.html#feature-discovery-design), there is discussion of using ripples to guide users through flows: "Ripples guide users through subsequent steps".

Currently, there is no good way (if any?) to access the button's ripple. I suppose I could create a new rippled component and position and mask it correctly, but that seems needlessly complex.

DevVersion:
Thinking about a good way to expose the ripple instance.. Any ideas already?

willshowell:
Why not just a public `@ViewChild(MdRipple)` in `button.ts`?

DevVersion:
Hm yeah. We might be able to just make that public. For most of the components like `md-slide-toggle`, `md-checkbox` and `md-radio` we already have a reference to the `MdRipple` directive.

Need to discuss this first with @jelbourn. Not sure if we should expose the underlying ripple instance.

jelbourn:
The `launch` method on `MdRipple` doesn't exactly have an API that would make it easy to say something like `button.ripple.launch()`. If we were going to expose the instance, I'd want to have a simpler user-facing API.

DevVersion:
Basically for components like the `slide-toggle`, `radio` etc. we provide a ripple config that will be used when calling `slide-toggle.ripple.launch()`.

For buttons it will be a bit more weird because the ripples don't launch from the `center`.

willshowell:
I think I see the difficulty. With a normal button, if you just used `launch()` it would render from the top left of the viewport and only barely show up within the button's bounds.

It would seem ideal to be able to pass an override config through `launch()` that would merge with the preexisting config:

```ts
this.button.ripple.launch(0, 0, { centered: true });
```

Overall, I agree with @jelbourn that the api could/should be simplified. It isn't exactly clear how `pageX`, `pageY`, and `centered` all interact, and it doesn't seem like there's a good way to trigger a ripple from, say, the bottom left corner of a component.

DevVersion:
@willshowell Exactly, the ripple would render from the top-left.

[...]

PR #9170, opened 2017-12-30
feat(button): expose ripple instance

DevVersion:
* Exposes the `MatRipple` instance of the button as a public property.

This can be used to show ripple indicators for feature discoveries as seen on the specs (https://material.io/guidelines/growth-communications/feature-discovery.html#feature-discovery-design-patterns). 

_Also a short quote from the specs_
> "Ripples guide users through subsequent steps".

As soon as https://github.com/angular/material2/issues/7543 is being resolved, we can also have more detailed explanations about the `MatRipple` directive. We could also have a guide that shows how to launch ripples on such components with an exposed ripple instance (e.g. button, slide-toggle, checkbox, radio)

Closes #4179

PR #9176, opened 2017-12-31
feat(checkbox): expose ripple instance

DevVersion:
* Exposes the reference to the `MatRipple` directive publically.

Similar to #9170. References #4179

### DIFF
--- a/src/lib/checkbox/checkbox.spec.ts
+++ b/src/lib/checkbox/checkbox.spec.ts
@@ -80,6 +80,10 @@ describe('MatCheckbox', () => {
       expect(inputElement.checked).toBe(false);
     });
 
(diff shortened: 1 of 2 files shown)


jelbourn:
This seems out of sync with the discussion in #4179, no?

DevVersion:
The PR should resolve #4179. It's not the implementation of the last idea (with the mixins), which intends to provide another public API for the ripples.

In my opinion, if we expose the ripple instance, then it should be really the `MatRipple` and no wrapper, that just provides a nicer API. 

Since there is a overview file for the ripple now, I think that it would be fine just exposing the instance. Also, as I mentioned in https://github.com/angular/material2/pull/9170, we can add a guide that shows how to control ripples of components that expose the instance.

PR #9181, opened 2018-01-01
fix(tooltip): match mobile dimensions from spec

### LINKED ISSUE #9039: [Tooltip] Mobile font size is too small

crisbeto:
Updates the various tooltip dimensions to match [the spec on mobile](https://material.io/guidelines/components/tooltips.html#tooltips-tooltips-mobile).

Fixes #9039.

### DIFF
--- a/src/lib/tooltip/BUILD.bazel
+++ b/src/lib/tooltip/BUILD.bazel
@@ -18,6 +18,7 @@ ng_module(
     "//src/cdk/keycodes",
     "//src/cdk/portal",
     "//src/cdk/scrolling",
+    "//src/cdk/layout",
     "@rxjs",
   ],
   tsconfig = ":tsconfig-build.json",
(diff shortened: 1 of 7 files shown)


donroyco:
@crisbeto Thanks for picking it up.

jelbourn:
@crisbeto needs rebase

crisbeto:
Rebased.

Issue #8791, opened 2017-12-04
About removing `| null` from `source` property declaration


rafaelss95:
#### Bug, feature request, or proposal:
Maybe a bug.

#### What is the expected behavior?
Just remove `| null` from declaration.

#### What is the current behavior?
Look at these classes:

https://github.com/angular/material2/blob/5210b3e197647381611ec48ddcd99333ec5356e9/src/lib/radio/radio.ts#L62
https://github.com/angular/material2/blob/5210b3e197647381611ec48ddcd99333ec5356e9/src/lib/button-toggle/button-toggle.ts#L57

I don't see how the `source` property of these `events` can be null.

I'm concerned about this because if `--strictNullChecks` is enabled in some application (as it's on mine) and we try to access any property of `source` we can't do this directly, like this:

```
someFunc(matButtonToggleChange: MatButtonToggleChange) {
  matButtonToggleChange.source.<ANY_PROPERTY>;
}
```

... you just received the following error from typescript: 

> Object is possibly 'null'

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
```
    _                      _                 ____ _     ___
   / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
  / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
 / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
/_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
               |___/

Angular CLI: 1.5.5
Node: 8.1.1
OS: linux x64
Angular: 5.0.3
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... platform-server, router

@angular/cdk: 5.0.0-rc.2
@angular/cli: 1.5.5
@angular/material-moment-adapter: 5.0.0-rc.2
@angular/material: 5.0.0-rc.2
@angular-devkit/build-optimizer: 0.0.34
@angular-devkit/core: 0.0.22
@angular-devkit/schematics: 0.0.38
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.5
@schematics/angular: 0.1.8
typescript: 2.6.2
webpack: 3.8.1
```

rafaelss95:
@tinayuangao do you agree that these declarations should be changed? Let me know, so I can go ahead and send a PR, if you want.

tinayuangao:
[...]

PR #9191, opened 2018-01-02
feat(button-toggle): general component cleanup and support value input in multiple mode

### LINKED ISSUE #9058: MatButtonToggleGroupMultiple lacks value property and change event
#### Feature request:
The docs clearly indicate that when using the MatButtonToggleGroup with the multiple directive, that you can no longer access the values of the control. While that is very informative, it makes the multiple variant of this control almost entirely unusable in a form. Without a change event or a value, you have to hack in things like click events just to get the selected values. 

The MatButtonToggleGroupMultiple class should ideally inherit all of the functionality currently in the MatButtonToggleGroup class

Currently, it inherits practically none of it.

Without being able to easily access the selected values, how are we supposed to use these in our forms?

Material 5.0.1

crisbeto:
* Reworks the `MatButtonToggleGroup` component to remove the need for the `MatButtonToggleGroupMultiple` component and to avoid having to implement features in two places.
* Reworks the `MatButtonToggleGroup` to use the `SelectionModel` for managing its state.
* Switches all of the `async` button toggle tests to run in `fakeAsync`.

As a side-effect of the above-mentioned changes, the toggle group in multiple mode now supports the same inputs as the one in single-selection mode (`value` input/output, `name` input, `change` event etc.).

Note: this is along the same lines as #2773, however #2773 got left behind for too long and it would need a lot more work to fit in our current setup.

Fixes #9058.

### DIFF
--- a/src/lib/button-toggle/button-toggle.ts
+++ b/src/lib/button-toggle/button-toggle.ts
@@ -8,8 +8,8 @@
 
 import {FocusMonitor} from '@angular/cdk/a11y';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
-import {UniqueSelectionDispatcher} from '@angular/cdk/collections';
 import {
+  AfterContentInit,
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
@@ -19,7 +19,6 @@ import {
   EventEmitter,
   forwardRef,
   Input,
-  OnDestroy,
   OnInit,
   Optional,
   Output,
@@ -34,6 +33,7 @@ import {
   mixinDisabled,
   mixinDisableRipple
 } from '@angular/material/core';
+import {SelectionModel} from '@angular/cdk/collections';
 
 /** Acceptable types for a button toggle. */
 export type ToggleType = 'checkbox' | 'radio';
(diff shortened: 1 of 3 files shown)


src/lib/button-toggle/button-toggle.ts
```diff
@@ -139,64 +135,56 @@ export class MatButtonToggleGroup extends _MatButtonToggleGroupMixinBase
...
+   * Selected button toggles in the group.
+   * @deprecated
+   */
   @Input()
   get selected() {
-    return this._selected;
+    const selected = this._selectionModel.selected;
+    return this.multiple ? selected : (selected[0] || null);
   }
+  set selected(selected: MatButtonToggle|MatButtonToggle[]|null) {
```
jelbourn: We usually have spaces between the unioned types

src/lib/button-toggle/button-toggle.ts
```diff
@@ -230,41 +218,89 @@ export class MatButtonToggleGroup extends _MatButtonToggleGroupMixinBase
...
+  /** Dispatch change event with current selection and group value. */
+  _emitChangeEvent(): void {
+    const selected = this.selected;
+    const source = Array.isArray(selected) ? selected[selected.length - 1] : selected;
+    const event = new MatButtonToggleChange(source, this.value);
+    this._controlValueAccessorChangeFn(event.value);
+    this.change.emit(event);
   }
-})
-export class MatButtonToggleGroupMultiple extends _MatButtonToggleGroupMixinBase
```
jelbourn: This would be a major change since it totally removes the symbol; people may have been using `@ViewChild` etc. to grab hold of these. Can we leave the symbol in as an alias for the combined one?

crisbeto:

src/lib/button-toggle/button-toggle.ts
```diff
@@ -230,41 +226,89 @@ export class MatButtonToggleGroup extends _MatButtonToggleGroupMixinBase
...
-  /** Whether the button toggle group should be vertical. */
-  private _vertical: boolean = false;
+  /**
+   * Syncs a button toggle's selected state with the model value.
+   * @param toggle Toggle to be synced.
+   * @param select Whether the toggle should be selected.
+   * @param isUserInput Whether the change was a result of a user interaction.
+   */
+  _syncButtonToggle(toggle: MatButtonToggle, select: boolean, isUserInput = false) {
+    if (!this.multiple && this.selected && !toggle.checked) {
```
tinayuangao: Why do we check `!toggle.checked` here? add some comments?

src/lib/button-toggle/button-toggle.ts
```diff
@@ -49,42 +49,46 @@ export const MAT_BUTTON_TOGGLE_GROUP_VALUE_ACCESSOR: any = {
   multi: true
 };
 
+/**
+ * @deprecated Use `MatButtonToggleGroup` instead.
+ */
+export class MatButtonToggleGroupMultiple {}
```
jelbourn: Add a unit test that queries for this?

jelbourn:
@crisbeto could you rebase this?

crisbeto:
Rebased.

Yeradon:
Any updates on this one? @crisbeto @jelbourn @tinayuangao

src/lib/button-toggle/button-toggle.ts
```diff
@@ -95,27 +117,35 @@ export class MatButtonToggleGroup extends _MatButtonToggleGroupMixinBase
...
   }
-  private _name: string = `mat-button-toggle-group-${_uniqueIdCounter++}`;
+  private _name = `mat-button-toggle-group-${_uniqueIdCounter++}`;
 
   /** Whether the toggle group is vertical. */
   @Input()
   get vertical(): boolean { return this._vertical; }
-  set vertical(value: boolean) { this._vertical = coerceBooleanProperty(value); }
-  private _vertical: boolean = false;
+  set vertical(value) {
```
rafaelss95: I think you have removed the type annotation here by accident.

rafaelss95:
@crisbeto can you add `closes #8791` here, please?

src/lib/button-toggle/button-toggle.ts
```diff
@@ -145,43 +191,25 @@ export class MatButtonToggleGroup extends _MatButtonToggleGroupMixinBase
...
-      } else if (this.value == null) {
-        this.selected = null;
-        this._buttonToggles.forEach(buttonToggle => {
-          buttonToggle.checked = false;
-        });
-      }
+  ngAfterContentInit() {
+    // If there was an attempt to assign a value before init, use it to set the
+    // initial selection, otherwise check the `checked` state of the toggles.
+    if (this._tempValue) {
```
tinayuangao: We need `this._tempValue !== null` here to accept falsy values.

crisbeto: Fixed.

rhutchison:
Thank you!

PR #9198, opened 2018-01-02
chore: bump travis nodejs version to 8

jelbourn:
@crisbeto I vaguely recall there was some issue w/ Angular preventing use of node 8, but I'm guessing it's likely resolved now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 dist: trusty
 
 node_js:
-  - '7'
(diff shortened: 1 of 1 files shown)


jelbourn:
Travis docs say that it should always install the latest version when you specify a major number:
https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Specifying-Node.js-versions

We can see how this goes before getting more specific

Issue #9162, opened 2017-12-29
Proposal: Disable tab-header pagination


tiaguinho:
tab-header automatic pagination is a great feature. But I think will be very good if we can disable that.
I'm currently working on a project where this feature breaks the layout when the tab has a scroll.

tiaguinho:
Can I implement this and submit a PR?

jelbourn:
@andrewseguin thoughts?

andrewseguin:
I'm fine if we want to include this as part of the API, there's another issue open requesting this that has gotten enough traction https://github.com/angular/material2/issues/2236

Closing this as a duplicate but will welcome a PR that adds this

PR #9214, opened 2018-01-03
feat(tabs): add option to disable pagination

tiaguinho:
issue https://github.com/angular/material2/issues/9162

### DIFF
--- a/src/lib/tabs/tab-group.ts
+++ b/src/lib/tabs/tab-group.ts
@@ -119,6 +119,9 @@ export class MatTabGroup extends _MatTabGroupMixinBase implements AfterContentIn
   /** Position of the tab header. */
   @Input() headerPosition: MatTabHeaderPosition = 'above';
 
+  /** Disable tab pagination */
+  @Input() disablePagination: boolean;
+
   /** Background color of the tab group. */
   @Input()
   get backgroundColor(): ThemePalette { return this._backgroundColor; }
--- a/src/lib/tabs/tab-group.ts
+++ b/src/lib/tabs/tab-group.ts
@@ -120,7 +120,10 @@ export class MatTabGroup extends _MatTabGroupMixinBase implements AfterContentIn
   @Input() headerPosition: MatTabHeaderPosition = 'above';
 
   /** Disable tab pagination */
-  @Input() disablePagination: boolean;
+  @Input()
+  get disablePagination(): boolean { return this._disablePagination; }
+  set disablePaignation(value: boolean) { this._disablePagination = coerceBooleanProperty(value); }
+  private _disablePagination: boolean = false;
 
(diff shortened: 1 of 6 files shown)


tiaguinho:
@andrewseguin thoughts?

src/lib/tabs/tab-header.ts
```diff
@@ -379,7 +382,7 @@ export class MatTabHeader extends _MatTabHeaderMixinBase
    * should be called sparingly.
    */
   _checkPaginationEnabled() {
-    const isEnabled =
+    const isEnabled = this.disablePagination ? !this.disablePagination :
```
andrewseguin: Replace `!this.disablePagination` with just `false`

andrewseguin:
Can you add some testing to the spec and include it in the demo app to validate its usage/ergonomics? Would be nice to see it in action and make sure it makes sense to use.

tiaguinho:
@andrewseguin just add test and include in the demo app. If need more changes, please let me know.

src/lib/tabs/tab-group.ts
```diff
@@ -119,6 +119,9 @@ export class MatTabGroup extends _MatTabGroupMixinBase implements AfterContentIn
   /** Position of the tab header. */
   @Input() headerPosition: MatTabHeaderPosition = 'above';
 
+  /** Disable tab pagination */
+  @Input() disablePagination: boolean;
```
willshowell: Consider coercing this to a boolean (see `dynamicHeight`) so users could do 

```html
<mat-tab-group disablePagination>
  <!-- ... -->
</mat-tab-group>
```

tiaguinho:
@willshowell I just change the code as you said.

tiaguinho:
@andrewseguin Need more changes?

src/lib/tabs/tab-group.ts
```diff
@@ -119,6 +119,12 @@ export class MatTabGroup extends _MatTabGroupMixinBase implements AfterContentIn
   /** Position of the tab header. */
   @Input() headerPosition: MatTabHeaderPosition = 'above';
 
+  /** Disable tab pagination */
+  @Input()
+  get disablePagination(): boolean { return this._disablePagination; }
+  set disablePaignation(value: boolean) { this._disablePagination = coerceBooleanProperty(value); }
```
andrewseguin: Got a bit of a breaking typo here

andrewseguin:
While running the demo app, I see that if `Disable Pagination` is set and the tabs overflow, there is some jank when selecting the last tab shown. It looks like the header tries to animate over to show the selected tab but then animates again to scroll left.

Are you seeing the same behavior?

jelbourn:
Closing due to inactivity

PR #9215, opened 2018-01-03
feat(form-field): implement hover state

crisbeto:
[Based on the spec](https://material.io/guidelines/components/text-fields.html#text-fields-states) form fields should have a hover state where the underline gets darkened while the user is hovering and gets replaced by the theme color after the focus the input.

### DIFF
--- a/src/lib/form-field/form-field.scss
+++ b/src/lib/form-field/form-field.scss
@@ -179,10 +179,6 @@ $mat-form-field-default-infix-width: 180px !default;
     opacity: 0;
     transition: background-color $swift-ease-in-duration $swift-ease-in-timing-function;
 
-    .mat-form-field-invalid:not(.mat-focused) & {
-      height: $mat-form-field-underline-height;
-    }
-
     .mat-focused &,
     .mat-form-field-invalid & {
       visibility: visible;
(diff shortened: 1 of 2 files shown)


src/lib/form-field/form-field.scss
```diff
@@ -195,6 +195,19 @@ $mat-form-field-default-infix-width: 180px !default;
   }
 }
 
+// Note that we need this specific of a selector because we don't want
+// want the hover effect to show when the user hovers over the hints.
```
benb7760: @crisbeto duplicate word 'want' on new line

mmalerba:
This feels a little jarring currently, I think it needs some animation to make it feel smoother. The animation videos in the spec don't seem to agree with the static images, since they show the underline already thick before hovering, so we might just have to play around with it on our own

crisbeto:
Added a short transition @mmalerba.

mmalerba:
while you're at it, it looks like error state should be 2px underline as well

mmalerba:
FYI, I've rolled these changes into the `standard` and `box` variants in #9366. Should make merging easier since users won't see a change until they ask for one.

crisbeto:
Closing in favor of #9366.

PR #9111, opened 2017-12-24
docs: remove tabindex from input

Hirse:
see #9107

PR #9221, opened 2018-01-03
docs(dialog): fix tabindex breaking focus trap

willshowell:
Fixes issue where tabbing through the dialog overview would break out of the focus trap. Can be seen at https://material.angular.io/components/dialog/examples.


crisbeto:
This has been addressed in https://github.com/angular/material2/pull/9111 already.

PR #9223, opened 2018-01-03
docs: clean up example config and data generation

crisbeto:
* Removes all of the optional properties being passed in as null in the example config. Also adds stronger typing to the list.
* Cleans up the example data generation to be closer to the rest of the project.

### DIFF
--- a/src/material-examples/example-module.ts
+++ b/src/material-examples/example-module.ts
@@ -5,14 +5,6 @@ import {NgModule} from '@angular/core';
 import {FormsModule, ReactiveFormsModule} from '@angular/forms';
 import {CommonModule} from '@angular/common';
 import {ExampleMaterialModule} from './material-module';
-
-export interface LiveExample {
-  title: string;
-  component: any;
-  additionalFiles?: string[];
-  selectorName?: string;
-}
-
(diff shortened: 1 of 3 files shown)


src/material-examples/example-module.ts
```diff
@@ -129,696 +123,485 @@ import {ToolbarOverviewExample} from './toolbar-overview/toolbar-overview-exampl
...
 
-export const EXAMPLE_COMPONENTS = {
+export interface LiveExample {
+  title: string;
+  component: any;
+  additionalFiles?: string[];
+  selectorName?: string;
+}
+
+export const EXAMPLE_COMPONENTS: {[key: string]: LiveExample} = {
```
jelbourn: This file is generated- did you update the script that writes it?

crisbeto: AFAIK that’s the example-data.ts. It already had all the null checks.

jelbourn: `gulp build-examples-module` writes this file

crisbeto: Oh I see, I must've missed the warning because of all the imports. I've updated the generation script now.

PR #9224, opened 2018-01-03
fix(chips): add bottom margin for all the chips

### LINKED ISSUE #8587: Chip input: no margin between chips in multiple lines

tinayuangao:
Fixes #8587

### DIFF
--- a/src/lib/chips/chips.scss
+++ b/src/lib/chips/chips.scss
@@ -25,13 +25,14 @@ $mat-chip-input-margin: 3px;
   border-radius: $mat-chip-horizontal-padding * 2;
   align-items: center;
   cursor: default;
+  margin-bottom: $mat-chip-input-margin;
 
   // Apply a margin to adjacent sibling chips.
   & + & {
-    margin: 0 0 0 $mat-chips-chip-margin;
+    margin: 0 0 $mat-chip-input-margin $mat-chips-chip-margin;
 
     [dir='rtl'] & {
-      margin: 0 $mat-chips-chip-margin 0 0;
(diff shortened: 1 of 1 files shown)


src/lib/chips/chips.scss
```diff
@@ -25,13 +25,14 @@ $mat-chip-input-margin: 3px;
...
+  margin-bottom: $mat-chip-input-margin;
 
   // Apply a margin to adjacent sibling chips.
   & + & {
-    margin: 0 0 0 $mat-chips-chip-margin;
+    margin: 0 0 $mat-chip-input-margin $mat-chips-chip-margin;
 
     [dir='rtl'] & {
-      margin: 0 $mat-chips-chip-margin 0 0;
+      margin: 0 $mat-chips-chip-margin $mat-chip-input-margin 0;
```
jelbourn: I suspect that this is something that we'd need to wait for a major version to change, since it could affect layouts in existing apps. It would also affect chips that aren't in a chip-list (which really shouldn't have any margin whatsoever).

tinayuangao:
Close this pr since #9762 is in.

PR #9236, opened 2018-01-04
fix(dialog): improved handling of scrollable content

### LINKED ISSUE #2481: Dialogue - overflow scrolling on small screens and margins

### LINKED ISSUE #3239: MdDialog width configuration uncenter dialog.
Bug: MdDialog not centered when is open passing a configuration of a custom width.

MdDialog should be the size of the specified width but keep his center location.

MdDialog is being moved to the left.

[https://plnkr.co/edit/fvG6lS8VIihJUlfu01Bv?p=preview](https://plnkr.co/edit/fvG6lS8VIihJUlfu01Bv?p=preview)

`openDialog() {
    this.dialog.open(DialogOverviewExampleDialog, {
      width: '600px'
    });
}`

image

Material 2.0.0-beta.2
Chrome in Windows

### LINKED ISSUE #6584: Dialog with fixed height less than 100% has 2 scrollbars
Dialog with fixed height, that is less than 100% to have only one scroll bar - for content section.

Dialog has 2 scrollbars - 1 for content and 1 for the whole dialog.

https://plnkr.co/edit/1teyLZLR7MVsrZLPD94V?p=preview

2 scrollbars is weak user-experience.

Browsers: Chrome, IE11
Material: 2.0.0-beta.6, 2.0.0-beta.8
OS, Typescript - not applicable

Probably the same is applicable to fixed width, but lets not noise the issue with it.

### LINKED ISSUE #8493: Setting dialog max-height does not work
Bug

To be able to change the max-height of the dialog

It does not work.

http://plnkr.co/edit/X17ETao12PtHTKmpqV5s?p=preview
The dialog in question is configured to be 
```
const settings: MatDialogConfig = {
  maxWidth: '99vw',
  maxHeight: '99vh'
}
```
But the height is the default one. It seems that the `cdk-overlay-0` is correctly styled with `style="max-width: 99vw; max-height: 99vh;"` but the inner `div.mat-dialog-content` has `max-height: 65vh;` which dictates the height.

All / latest

Keep up the good work!

crisbeto:
* Improves the handling of scrollable content inside the `mat-dialog-content` by using flexbox, rather than a hardcoded `max-height`, to define the content height. This resolves various issues where the dialog would go out of the screen at certain screen sizes or have multiple scrollbars.
* Uses flexbox to ensure that the dialog content elements are always at the appropriate size.

Fixes #2481.
Fixes #3239.
Fixes #6584.
Fixes #8493.


src/lib/dialog/dialog-container.ts
```diff
@@ -120,7 +120,13 @@ export class MatDialogContainer extends BasePortalOutlet {
     }
 
     this._savePreviouslyFocusedElement();
-    return this._portalOutlet.attachComponentPortal(portal);
+
+    const componentRef = this._portalOutlet.attachComponentPortal(portal);
```
jelbourn: Add a comment here that explains why this extra class is added?

src/lib/dialog/dialog-container.html
```diff
@@ -1 +1,3 @@
-<ng-template cdkPortalOutlet></ng-template>
+<div class="mat-dialog-component-host">
```
jelbourn: Is this class meant to be both in the template and added manually?

src/lib/dialog/dialog-config.ts
```diff
@@ -74,7 +74,7 @@ export class MatDialogConfig<D = any> {
   maxWidth?: number | string = '80vw';
 
   /** Max-height of the dialog. If a number is provided, pixel units are assumed. */
-  maxHeight?: number | string;
+  maxHeight?: number | string = '80vh';
```
jelbourn: Was this `65vh` before? If so, we probably would need to keep the same max-height until a major version change

src/lib/dialog/dialog.scss
```diff
@@ -27,34 +30,61 @@ $mat-dialog-button-margin: 8px !default;
...
+  }
+}
+
+.mat-dialog-title {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  width: 100%;
+  flex-shrink: 0;
+  order: 1;
```
jelbourn: Wouldn't specifying the `order` take away the ability to put custom content in between the regions?

src/lib/dialog/dialog-config.ts
crisbeto: It wasn't, it was the scrollable content that was `65vh`, this applies the height to the overlay pane itself.

src/lib/dialog/dialog-container.html
crisbeto: Yeah, we need to add it through both to cover both template and dialog portals. The class is used to propagate the `max-height` down correctly since we've got a lot of nested divs.

src/lib/dialog/dialog.scss
crisbeto: I suppose that it would, but the assumption is that if they're using the content element, they're putting all their content into it. I'm fine with removing it though, we haven't had it until now.

jelbourn: If the only reason would be to let people put the dialog directives in any order, I'd rather omit it to give people control in case they need to do something special

crisbeto:

jelbourn (dismissed):
LGTM

dinamix:
Looking forward to this merge!

pharv:
Great work on this PR!  I am really looking forward to this enhancement as well.  It looks like the only reason it hasn't been merged in yet is because the Travis CI build failed due to server issues.  Can someone please re-run the build?  Thanks!

Devvox93:
What are the presubmit failures that are keeping this from being merged?
I, too, am having the same scrolling issues with the material dialogs.

tammysock:
This fix will be very helpful. Any updates on the PR?

swftvsn:
Hi, I would really appreciate this fix too, we've been waiting for a year now for this to land (we're eager to take out our own hacks to overcome some issues that this would fix..)

So, any news about this @crisbeto @jelbourn ?

jelbourn:
It's blocked until we can resolve this change breaking some things inside Google.

rkrzewski:
Hi @jelbourn is this issue still blocked on your side?

jelbourn:
Yes

josephperrott:
This will fix #13190

jpike88:
@crisbeto is this PR going to happen?

MaximoB:
This PR would fix major problems I'm having using MatDialog, any updates?

akvaliya:
What is the status of this PR? When will it be completed.

charsleysa:
@jelbourn is this still blocked?

jelbourn:
Yes- we haven't work on this since the last update. At this point, we'll probably see if we can address this as part of our work on integrating MDC-web into the components.

AlexElin:
@jelbourn have you had a chance to find out whether integrating MDC-web into the components will solve the problems that this PR intended to solve?

jelbourn:
@crisbeto is this something you've looked at for the mdc dialog?

crisbeto:
I haven't looked into the MDC dialog since somebody else implemented it.

josephperrott:
@crisbeto Looks like this PR has fallen out of date, can you please rebase the PR.

MikaStark:
Hope this PR is not abandoned and will be merged someday

PR #9240, opened 2018-01-05
fix: adjust types in autocomplete-display-example

rafaelss95:
xxx2

### DIFF
--- a/src/material-examples/autocomplete-display/autocomplete-display-example.ts
+++ b/src/material-examples/autocomplete-display/autocomplete-display-example.ts
@@ -31,8 +31,8 @@ export class AutocompleteDisplayExample {
   ngOnInit() {
     this.filteredOptions = this.myControl.valueChanges
       .pipe(
-        startWith({} as User),
-        map(user => user && typeof user === 'object' ? user.name : user),
+        startWith<string | User>(''),
+        map(value => typeof value === 'string' ? value : value.name),
         map(name => name ? this.filter(name) : this.options.slice())
       );
   }
@@ -42,8 +42,8 @@ export class AutocompleteDisplayExample {
       option.name.toLowerCase().indexOf(name.toLowerCase()) === 0);
   }
 
-  displayFn(user: User): string {
-    return user ? user.name : user;
+  displayFn(user?: User): string | undefined {
+    return user ? user.name : undefined;
   }
 
 }


src/material-examples/autocomplete-display/autocomplete-display-example.ts
```diff
@@ -31,8 +31,8 @@ export class AutocompleteDisplayExample {
   ngOnInit() {
     this.filteredOptions = this.myControl.valueChanges
       .pipe(
-        startWith({} as User),
-        map(user => user && typeof user === 'object' ? user.name : user),
+        startWith<string | User>(this.myControl.value),
+        map(value => value && typeof value === 'object' ? value.name : value as string),
```
rafaelss95: `as` was necessary to run properly on Stackblitz.

src/material-examples/autocomplete-display/autocomplete-display-example.ts
```diff
@@ -42,8 +42,8 @@ export class AutocompleteDisplayExample {
       option.name.toLowerCase().indexOf(name.toLowerCase()) === 0);
   }
 
-  displayFn(user: User): string {
-    return user ? user.name : user;
+  displayFn(user?: User): string | undefined {
+    return user ? user.name : undefined;
```
rafaelss95: I'd with the following:

    displayFn(user?: User): string | undefined {
      return user ? user.name : user;
    }

... however it doesn't go well in Stackblitz (maybe because there we don't have `strictNullChecks`).

rafaelss95:
The class `User` can be just an `Interface`. What do you think, @jelbourn?

src/material-examples/autocomplete-display/autocomplete-display-example.ts
```diff
@@ -31,8 +31,8 @@ export class AutocompleteDisplayExample {
   ngOnInit() {
     this.filteredOptions = this.myControl.valueChanges
       .pipe(
-        startWith({} as User),
-        map(user => user && typeof user === 'object' ? user.name : user),
+        startWith<string | User>(this.myControl.value),
```
crisbeto: `myControl.value` will always be undefined. You could start it off with an empty string instead.

src/material-examples/autocomplete-display/autocomplete-display-example.ts
crisbeto: The `typeof something === 'object'` is a little tricky since it'll also be true for `null`. This might be nicer if you changed it to: 

```ts
return typeof value === 'string' ? value : value.name;
```

rafaelss95: *"The typeof something === 'object' is a little tricky since it'll also be true for null"*

In this case, it won't happen because the check is `value && typeof value === 'object'`. Anyway, I changed it since it's more readable.

rafaelss95:
Out of curiosity, can someone explain why these errors weren't caught by the compiler?

PR #9242, opened 2018-01-05
fix(divider): add inset styles for icons and lists in cards

### LINKED ISSUE #9233: bug(mat-divider): wrong styling

Bug introduced by https://github.com/angular/material2/pull/5862

- A `mat-divider` with `inset=true` is vertically aligned to the start of a `mat-list-item` text
- The behavior is the same when used in a mat-list contained in a card

The lists (inside and outside the card) should look as presented in https://material.io/guidelines/components/dividers.html#dividers-usage

- A `mat-divider` with `inset=true` wont align vertically to the start of a `mat-list-item` text
- Behavior is removed when used inside of a `mat-list `contained in a `mat-card`

https://stackblitz.com/edit/angular-mat-div-issue?file=index.html

Angular 5.x.x
Material 5.0.3

CaerusKaru:
* Add mixin to dynamically apply inset divider offset based on
  the precursor in the list item
* Add inset divider example using icons as precursor

Fixes #9233

### DIFF
--- a/src/lib/list/list.scss
+++ b/src/lib/list/list.scss
@@ -1,9 +1,11 @@
 @import '../core/style/variables';
 @import '../core/style/list-common';
 @import '../core/style/layout-common';
+@import '../divider/divider';
 
 
 $mat-list-side-padding: 16px;
+$mat-list-icon-padding: 4px;
 $mat-list-avatar-size: 40px;
 
 // Normal list variables
@@ -131,6 +133,10 @@ $mat-list-item-inset-divider-offset: 72px;
     width: $mat-list-avatar-size;
     height: $mat-list-avatar-size;
     border-radius: 50%;
+
+    ~ .mat-divider-inset {
+      @include mat-inset-divider-offset($mat-list-avatar-size, $mat-list-side-padding);
+    }
   }
 
   .mat-list-icon {
@@ -140,7 +146,12 @@ $mat-list-item-inset-divider-offset: 72px;
     font-size: $icon-size;
     box-sizing: content-box;
     border-radius: 50%;
-    padding: 4px;
+    padding: $mat-list-icon-padding;
+
+    ~ .mat-divider-inset {
+      @include mat-inset-divider-offset($icon-size + (2 * $mat-list-icon-padding),
+              $mat-list-side-padding);
+    }
   }
 
   .mat-divider {
@@ -148,21 +159,15 @@ $mat-list-item-inset-divider-offset: 72px;
     bottom: 0;
     left: 0;
     width: 100%;
+    margin: 0;
 
     [dir='rtl'] & {
-      left: auto;
-      right: 0;
+      margin-left: auto;
+      margin-right: 0;
     }
 
     &.mat-divider-inset {
-      left: $mat-list-item-inset-divider-offset;
-      width: calc(100% - #{$mat-list-item-inset-divider-offset});
-      margin: 0;
-
-      [dir='rtl'] & {
-        left: auto;
-        right: $mat-list-item-inset-divider-offset;
-      }
+      position: absolute; // necessary to override card styles
     }
   }
 }
--- a/src/demo-app/list/list-demo.html
+++ b/src/demo-app/list/list-demo.html
@@ -31,7 +31,7 @@
           <span>{{message.subject}} -- </span>
           <span class="demo-secondary-text">{{message.message}}</span>
         </p>
-        <mat-divider [inset]="true" *ngIf="!last"></mat-divider>
+        <mat-divider inset *ngIf="!last"></mat-divider>
       </mat-list-item>
       <mat-divider></mat-divider>
       <mat-list-item *ngFor="let message of messages">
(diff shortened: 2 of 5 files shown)


CaerusKaru:
Circle CI build error is unrelated to this PR

cc @jelbourn @crisbeto

src/lib/list/list.scss
```diff
@@ -27,7 +28,18 @@ $mat-dense-three-line-height: 76px;
 $mat-dense-multi-line-padding: 16px;
 $mat-dense-list-icon-size: 20px;
 
-$mat-list-item-inset-divider-offset: 72px;
+// This mixin provides the correct offset for an inset divider based on the
+// size of the parent class (e.g. avatar vs icon)
+@mixin mat-inset-divider-offset($offset) {
+  $mat-list-item-inset-divider-offset: #{(2 * $mat-list-side-padding) + $offset};
+  left: $mat-list-item-inset-divider-offset;
```
crisbeto: Why is this one using `left`/`right`, whereas the default divider styles are using `margin-left`/`margin-right`?

src/lib/list/list.scss
```diff
@@ -27,7 +28,18 @@ $mat-dense-three-line-height: 76px;
 $mat-dense-multi-line-padding: 16px;
 $mat-dense-list-icon-size: 20px;
 
-$mat-list-item-inset-divider-offset: 72px;
+// This mixin provides the correct offset for an inset divider based on the
+// size of the parent class (e.g. avatar vs icon)
+@mixin mat-inset-divider-offset($offset) {
```
crisbeto: The mixin should live somewhere in the divider module since it could be used in other modules that consume it.

src/lib/list/list.scss
```diff
@@ -27,7 +28,18 @@ $mat-dense-three-line-height: 76px;
 $mat-dense-multi-line-padding: 16px;
 $mat-dense-list-icon-size: 20px;
 
-$mat-list-item-inset-divider-offset: 72px;
```
crisbeto: Removing this is a breaking change, even though it isn't being used anywhere.

src/lib/list/list.scss
```diff
@@ -131,14 +151,8 @@ $mat-list-item-inset-divider-offset: 72px;
...
     &.mat-divider-inset {
-      left: $mat-list-item-inset-divider-offset;
-      width: calc(100% - #{$mat-list-item-inset-divider-offset});
       margin: 0;
-
-      [dir='rtl'] & {
-        left: auto;
-        right: $mat-list-item-inset-divider-offset;
-      }
+      position: absolute;
```
crisbeto: The `position: absolute` is defined in the parent rule above, you don't have to add it to `.mat-divider-inset`. Also the nesting here isn't necessary.

src/lib/list/list.scss
CaerusKaru: Since `mat-divider` hasn't been merged into the minor version yet, can we still remove it? Happy to keep it as a placeholder otherwise

src/lib/list/list.scss
CaerusKaru: Moved

src/lib/list/list.scss
CaerusKaru: Changed, the reason was that it was legacy from AngularJS Material

src/lib/list/list.scss
CaerusKaru: It was present to override a similar `position: static` in the `mat-card` styles. I added an `!important` tag instead with a note for why it's there. I can add the nested structure back if you prefer to not have `!important`

src/lib/list/list.scss
crisbeto: It could still be used from when the divider was in the list module.

src/lib/list/list.scss
```diff
@@ -116,29 +120,23 @@ $mat-list-item-inset-divider-offset: 72px;
...
+    padding: $mat-list-icon-padding;
+
+    ~ .mat-divider-inset {
+      @include mat-inset-divider-offset($icon-size + (2 * $mat-list-icon-padding));
+    }
   }
 
   .mat-divider {
-    position: absolute;
+    position: absolute !important; // this is done to override card inset styles
```
crisbeto: Rather than using `!important`, perhaps the selector can be made more specific or the card one can be less specific?

CaerusKaru: The original nesting was the way to make it more specific. I don't think that modifying the card styles will be as elegant. I'll add back the nesting for now unless you have an alternative suggestion

src/lib/list/list.scss
crisbeto: In that case it can at least use a comment about why the nesting was necessary.

CaerusKaru: Perfect, will add

CaerusKaru:
@crisbeto I'm not entirely sure why the Bazel build is failing on Circle CI (I know it's because it doesn't like how I'm including the divider style file). Do you know what's going on?

crisbeto:
I'm not too sure about that error @CaerusKaru, my guess is that you'd need to add another item to the `deps` array.

CaerusKaru:
@crisbeto I tried adding the divider scss rule to the deps array, but that's what Bazel is complaining about. @jelbourn Any thoughts?

CaerusKaru:
@crisbeto I fixed the Bazel issues and rebased with master. Ready for review again.

PR #9248, opened 2018-01-05
build: fix builds not failing with compilation errors in unit tests

crisbeto:
Fixes the CI continuing to run even though it failed to compile, due to a couple of promises that weren't rejecting properly. This should prevent things like #9073 from slipping through.


crisbeto:
We've been getting some flakes from Travis these past couple of days. Restarting it now.

PR #9250, opened 2018-01-05
fix(select): close panel on alt + arrow key presses

crisbeto:
* Closes the select panel when pressing alt + down/up arrow, based on the native select.
* Does some minor cleanup around the select tests.

### DIFF
--- a/src/lib/select/select.ts
+++ b/src/lib/select/select.ts
@@ -646,6 +646,7 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit, 
     const isArrowKey = keyCode === DOWN_ARROW || keyCode === UP_ARROW;
     const isOpenKey = keyCode === ENTER || keyCode === SPACE;
 
+    // Open the select on ALT + arrow key to match the native <select>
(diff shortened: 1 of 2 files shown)


src/lib/select/select.ts
```diff
@@ -657,23 +657,26 @@ export class MatSelect extends _MatSelectMixinBase implements AfterContentInit,
...
+    const manager = this._keyManager;
 
     if (keyCode === HOME || keyCode === END) {
       event.preventDefault();
-      keyCode === HOME ? this._keyManager.setFirstItemActive() :
-                         this._keyManager.setLastItemActive();
-    } else if ((keyCode === ENTER || keyCode === SPACE) && this._keyManager.activeItem) {
+      keyCode === HOME ? manager.setFirstItemActive() : manager.setLastItemActive();
+    } else if (isArrowKey && event.altKey) {
+      event.preventDefault();
```
jelbourn: The comment was: 
Add a comment here like
```ts
// Close the select on ALT + arrow key to match the native <select>
```

PR #9253, opened 2018-01-05
feat(ripple): support animation duration overwrites

DevVersion:
* Adds a new option to the ripples that allows developers to have a better control of the animation (all ripples, or even individual ripples).

* Deprecates the `matRippleSpeedFactor` in favor of the `matRippleAnimation` binding that accepts a `RippleAnimationConfig`. The configuration is more explicit, clean and not confusing as the `speedFactor`.

* To provide a more user-friendly `launch()` method API, the passed ripple config will extend the default ripple config from the `MatRipple` instance (removes unnecessary bloat; requested in https://github.com/angular/material2/issues/4179#issuecomment-295889498)

* Disables ripples for most of the demo buttons in the ripple demo (allows better debugging; when pressing the buttons)

_Text from the ripple overview_ (**Disabling animation**)

The animation of ripples can be disabled by using the `animation` global option. If the 
`enterDuration` and `exitDuration` is being set to `0`, ripples will just appear without any
animation.

This is specifically useful in combination with the `disabled` global option, because globally
disabling ripples won't affect the focus indicator ripples. If someone still wants to disable
those ripples for performance reasons, the duration can be set to `0`, to remove the ripple feel.

```ts
const globalRippleConfig: RippleGlobalOptions = {
  disabled: true,
  animation: {
    enterDuration: 0,
    exitDuration: 0
  }
};
```

### DIFF
--- a/src/lib/core/ripple/ripple-renderer.ts
+++ b/src/lib/core/ripple/ripple-renderer.ts
@@ -9,25 +9,26 @@ import {ElementRef, NgZone} from '@angular/core';
 import {Platform, supportsPassiveEventListeners} from '@angular/cdk/platform';
 import {RippleRef, RippleState} from './ripple-ref';
 
-/** Fade-in duration for the ripples. Can be modified with the speedFactor option. */
-export const RIPPLE_FADE_IN_DURATION = 450;
-
-/** Fade-out duration for the ripples in milliseconds. This can't be modified by the speedFactor. */
-export const RIPPLE_FADE_OUT_DURATION = 400;
-
-/**
- * Timeout for ignoring mouse events. Mouse events will be temporary ignored after touch
- * events to avoid synthetic mouse events.
- */
-const IGNORE_MOUSE_EVENTS_TIMEOUT = 800;
-
 export type RippleConfig = {
   color?: string;
   centered?: boolean;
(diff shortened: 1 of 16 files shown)


src/lib/core/ripple/ripple-renderer.ts
```diff
@@ -9,26 +9,27 @@ import {ElementRef, NgZone} from '@angular/core';
...
+  /** @deprecated Use the animation property instead. */
+  speedFactor?: number;
 };
 
+/**
+ * Interface that describes the configuration for the animation of a ripple.
+ * There are two animation phases with different durations for the ripples.
+ */
+export interface RippleAnimationConfig {
+  /** Duration in milliseconds for the enter animation. */
```
jelbourn: Could you expand the comment a bit more to describe what constitutes "enter animation" and "exit animation"?

DevVersion: Done.

jelbourn:
@DevVersion passes Google presubmit, just needs rebase

DevVersion:
@jelbourn Done.

jelbourn:
@DevVersion can you rebase again? Conflicted with your other PR for radio buttons that got merged first

PR #9257, opened 2018-01-06
fix(dialog): restore focus with the proper focus origin

devversion:
* Restores the trigger focus upon dialog close with the proper focus origin that caused the dialog closing.

For example, a backdrop click leads to a focus restore via `mouse`. Pressing `escape` leads to a focus restore via `keyboard`.

References #8420

### DIFF
--- a/src/material/dialog/BUILD.bazel
+++ b/src/material/dialog/BUILD.bazel
@@ -56,6 +56,7 @@ ng_test_library(
     ),
     deps = [
         ":dialog",
+        "//src/cdk/a11y",
         "//src/cdk/bidi",
         "//src/cdk/keycodes",
         "//src/cdk/overlay",
--- a/src/material/dialog/dialog-container.ts
+++ b/src/material/dialog/dialog-container.ts
@@ -29,7 +29,7 @@ import {
   TemplatePortal,
   DomPortal
 } from '@angular/cdk/portal';
-import {FocusTrap, FocusTrapFactory} from '@angular/cdk/a11y';
+import {FocusTrap, FocusMonitor, FocusOrigin, FocusTrapFactory} from '@angular/cdk/a11y';
 import {MatDialogConfig} from './dialog-config';
 
 
@@ -82,6 +82,13 @@ export class MatDialogContainer extends BasePortalOutlet {
   /** Element that was focused before the dialog was opened. Save this to restore upon close. */
   private _elementFocusedBeforeDialogWasOpened: HTMLElement | null = null;
 
+  /**
+   * Type of interaction that led to the dialog being closed. This is used to determine
+   * whether the focus style will be applied when returning focus to its original location
+   * after the dialog is closed.
+   */
+  _closeInteractionType: FocusOrigin|null = null;
+
   /** State of the dialog animation. */
   _state: 'void' | 'enter' | 'exit' = 'enter';
 
@@ -100,7 +107,8 @@ export class MatDialogContainer extends BasePortalOutlet {
     private _changeDetectorRef: ChangeDetectorRef,
     @Optional() @Inject(DOCUMENT) _document: any,
     /** The dialog configuration. */
-    public _config: MatDialogConfig) {
+    public _config: MatDialogConfig,
+    private _focusMonitor?: FocusMonitor) {
 
     super();
     this._ariaLabelledBy = _config.ariaLabelledBy || null;
@@ -178,10 +186,11 @@ export class MatDialogContainer extends BasePortalOutlet {
 
   /** Restores focus to the element that was focused before the dialog opened. */
   private _restoreFocus() {
-    const toFocus = this._elementFocusedBeforeDialogWasOpened;
+    const previousElement = this._elementFocusedBeforeDialogWasOpened;
 
     // We need the extra check, because IE can set the `activeElement` to null in some cases.
-    if (this._config.restoreFocus && toFocus && typeof toFocus.focus === 'function') {
+    if (this._config.restoreFocus && previousElement &&
+        typeof previousElement.focus === 'function') {
       const activeElement = this._document.activeElement;
       const element = this._elementRef.nativeElement;
 
@@ -190,8 +199,13 @@ export class MatDialogContainer extends BasePortalOutlet {
       // the consumer moved it themselves before the animation was done, in which case we shouldn't
       // do anything.
       if (!activeElement || activeElement === this._document.body || activeElement === element ||
-        element.contains(activeElement)) {
-        toFocus.focus();
+          element.contains(activeElement)) {
+        if (this._focusMonitor) {
+          this._focusMonitor.focusVia(previousElement, this._closeInteractionType);
+          this._closeInteractionType = null;
+        } else {
+          previousElement.focus();
+        }
       }
     }
 
--- a/src/material/dialog/dialog-content-directives.ts
+++ b/src/material/dialog/dialog-content-directives.ts
@@ -16,7 +16,7 @@ import {
   ElementRef,
 } from '@angular/core';
 import {MatDialog} from './dialog';
-import {MatDialogRef} from './dialog-ref';
+import {_closeDialogVia, MatDialogRef} from './dialog-ref';
 
 /** Counter used to generate unique IDs for dialog elements. */
 let dialogElementUid = 0;
@@ -28,7 +28,7 @@ let dialogElementUid = 0;
   selector: '[mat-dialog-close], [matDialogClose]',
   exportAs: 'matDialogClose',
   host: {
-    '(click)': 'dialogRef.close(dialogResult)',
+    '(click)': '_onButtonClick($event)',
     '[attr.aria-label]': 'ariaLabel || null',
     '[attr.type]': 'type',
   }
@@ -67,6 +67,15 @@ export class MatDialogClose implements OnInit, OnChanges {
     if (proxiedChange) {
       this.dialogResult = proxiedChange.currentValue;
     }
+  }
+
+  _onButtonClick(event: MouseEvent) {
+    // Determinate the focus origin using the click event, because using the FocusMonitor will
+    // result in incorrect origins. Most of the time, close buttons will be auto focused in the
+    // dialog, and therefore clicking the button won't result in a focus change. This means that
+    // the FocusMonitor won't detect any origin change, and will always output `program`.
+    _closeDialogVia(this.dialogRef,
+        event.screenX === 0 && event.screenY === 0 ? 'keyboard' : 'mouse', this.dialogResult);
   }
 }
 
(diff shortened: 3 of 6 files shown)


src/lib/dialog/dialog-container.ts
```diff
@@ -95,6 +98,7 @@ export class MatDialogContainer extends BasePortalOutlet {
     private _elementRef: ElementRef,
     private _focusTrapFactory: FocusTrapFactory,
     private _changeDetectorRef: ChangeDetectorRef,
+    private _focusMonitor: FocusMonitor,
```
crisbeto: @jelbourn should we make this optional? It's technically a breaking signature change, but the dialog container is an internal component.

src/lib/dialog/dialog.spec.ts
```diff
@@ -40,14 +41,15 @@ describe('MatDialog', () => {
...
+  let focusMonitor: FocusMonitor;
 
   let testViewContainerRef: ViewContainerRef;
   let viewContainerFixture: ComponentFixture<ComponentWithChildViewContainer>;
   let mockLocation: SpyLocation;
 
   beforeEach(fakeAsync(() => {
     TestBed.configureTestingModule({
-      imports: [MatDialogModule, DialogTestModule],
+      imports: [MatDialogModule, DialogTestModule, A11yModule],
```
crisbeto: Importing the `A11yModule` shouldn't be necessary here, it's imported by the `MatDialogModule` already.

src/lib/dialog/dialog-ref.ts
```diff
@@ -93,6 +95,7 @@ export class MatDialogRef<T, R = any> {
       this._overlayRef.detachBackdrop();
     });
 
+    this._containerInstance._closeFocusOrigin = closeFocusOrigin;
```
crisbeto: Rather than assigning the origin here as a property, you could pass it in to `_startExitAnimation`. It isn't a lot cleaner, but at least it keeps all the logic in the same class and you can turn the `_closeFocusOrigin` into a private property.

src/lib/dialog/dialog.spec.ts
```diff
@@ -900,6 +903,65 @@ describe('MatDialog', () => {
       document.body.removeChild(button);
     }));
 
+    it('should re-focus the trigger via mouse when dialog closes through escape', fakeAsync(() => {
```
crisbeto: `via mouse` -> `via keyboard`

src/lib/dialog/dialog.spec.ts
```diff
@@ -900,6 +903,65 @@ describe('MatDialog', () => {
...
+      tick(500);
+
+      expect(lastFocusOrigin!)
+        .toBe('keyboard', 'Expected the trigger button to be focused via keyboard');
+
+      focusMonitor.stopMonitoring(button);
+      document.body.removeChild(button);
+    }));
+
+    it('should re-focus the trigger via keyboard when backdrop has been clicked', fakeAsync(() => {
```
crisbeto: `via keyboard` -> `via mouse`

src/lib/dialog/dialog-ref.ts
```diff
@@ -72,14 +73,15 @@ export class MatDialogRef<T, R = any> {
...
     _overlayRef.keydownEvents()
       .pipe(filter(event => event.keyCode === ESCAPE && !this.disableClose))
-      .subscribe(() => this.close());
+      .subscribe(() => this.close(undefined, 'keyboard'));
   }
 
   /**
    * Close the dialog.
    * @param dialogResult Optional result to return to the dialog opener.
+   * @param closeFocusOrigin Focus origin that describes how the dialog got closed
```
crisbeto: I'm not sure whether this description is super obvious about what it does. Maybe something along the lines of "Type of interaction that closed the dialog"? Also I think the param can be renamed to `focusOrigin` since it's already part of the `close` method.

src/lib/dialog/dialog-ref.ts
DevVersion: Sounds good. I didn't go with that approach initially because I wasn't sure whether this technically belongs to an "animation".

src/lib/dialog/dialog.spec.ts
DevVersion: That's true, but I thought since we monitor the button independently from the dialog in a test, I think being explicit is better.

src/lib/dialog/dialog-ref.ts
crisbeto: It doesn't, but it could be renamed to somthing like `_startExitSequence`.

src/lib/dialog/dialog-container.ts
jelbourn: I think this one should be okay

src/lib/dialog/dialog-ref.ts
```diff
@@ -72,14 +73,15 @@ export class MatDialogRef<T, R = any> {
...
   }
 
   /**
    * Close the dialog.
-   * @param dialogResult Optional result to return to the dialog opener.
+   * @param dialogResult Optional result to return to the dialog opener
+   * @param focusOrigin Type of interaction that closed the dialog
    */
-  close(dialogResult?: R): void {
+  close(dialogResult?: R, focusOrigin: FocusOrigin = 'program'): void {
```
jelbourn: I think I mentioned this in chat: instead of changing the signature of the `close` method, I think we should add another method, probably internal, like `_closeVia`

It would also make more sense to update the `_closeFocusOrigin` state in such a function rather than changing `_startExitAnimation`

src/lib/dialog/dialog-container.ts
```diff
@@ -79,6 +79,9 @@ export class MatDialogContainer extends BasePortalOutlet {
   /** Element that was focused before the dialog was opened. Save this to restore upon close. */
   private _elementFocusedBeforeDialogWasOpened: HTMLElement | null = null;
 
+  /** Type of interaction that closed the dialog. */
+  private _closeFocusOrigin: FocusOrigin = 'program';
```
jelbourn: I'd expand this description to something like
```
Type of interaction that led to the dialog being closed. This is used to determine
where the focus style will be applied when returning focus to its original location
after the dialog is closed.
```

The name `_closeInteractionType` might also be more accurate

src/lib/dialog/dialog-ref.ts
DevVersion: Yeah we chatted about that. I just thought that it also would be fine like that, because the second parameter would be optional. Will change it.

DevVersion:
Addressed the feedback and also added the check for the `matDialogClose` we discussed in the standup.

src/lib/dialog/dialog-container.ts
```diff
@@ -79,6 +79,13 @@ export class MatDialogContainer extends BasePortalOutlet {
   /** Element that was focused before the dialog was opened. Save this to restore upon close. */
   private _elementFocusedBeforeDialogWasOpened: HTMLElement | null = null;
 
+  /**
+   * Type of interaction that led to the dialog being closed. This is used to determine
+   * where the focus style will be applied when returning focus to its original location
```
jelbourn: I had a typo in my suggestion: "where" -> "whether"

mmalerba:
@DevVersion please rebase

DevVersion:
@mmalerba Done.

Carniatto:
+1

DevVersion:
@josephperrott Rebased that one. Can we retry a presubmit in the future?

josephperrott:
I ran the presubmit and it looks like we have quite a few presubmit failures internally still.

It looks like we have a quite a few tests that are testing to confirm that `close` is called on a `MockDialogRef`, so since we now call `_closeVia` it doesn't work as expected.

DevVersion:
@josephperrott Makes sense. Is there any way they will update their `MockDialogRef`, or should we consider updating the logic so that it uses the actual `close` method?

Carniatto:
Will this be merged someday?

Carniatto:
Any news on this?  it's a really annoying thing

devversion:
@jelbourn @josephperrott I rebased this one again. should we proceed with that one or don't you think it's worth moving forward with that? It definitely seems like a reasonable a11y feature for the dialog to me.

zelliott:
Just a quick bump that it would be awesome to get this in.

devversion:
Rebased again. If we decide to move forward with this, then the changes should most likely also be ported over to the `cdk-experimental/dialog` version.

andrewseguin:
@devversion Want to move forward with this PR? If so, this is our next oldest P2 and I'm going to make it my side project to get a green presubmit with it.

devversion:
@andrewseguin That would be awesome. I've rebased the PR. Let me know if I can help somehow!

andrewseguin:
We're at a fork in the roads on this PR. This breaks a lot of internal tests. 

Many tests provide a fake dialog ref using jasmine, e.g. 
```
const dialogRefSpy = jasmine.createSpyObj('MatDialogRef', ['close']);
```

Then in their test, they check that close was called:

```
expect(dialogRefSpy.close).toHaveBeenCalledWith(false);
```

This change breaks that because now `MatDialogClose` calls `_closeVia` instead of `close`, so their expectation fails.

The "right thing to do" is to provide a fake for users so they don't need to make their own. Then they can just verify through the fake that `expect(fakeDialogRef.closeResult).toBe(expectation)`. However, there are 1000+ instances of this internally, and it's going to take so many hours/days dedicated to migrate the tests.

An alternative path forward is to change this code a bit so that `(click)` still calls `close()`, but first sets some state through the `MatDialogRef`. For example, you can change `_onButtonClick` to be:

```
_onButtonClick(event: MouseEvent) {
    // Determinate the focus origin using the click event, because using the FocusMonitor will
    // result in incorrect origins. Most of the time, close buttons will be auto focused in the
    // dialog, and therefore clicking the button won't result in a focus change. This means that
    // the FocusMonitor won't detect any origin change, and will always output `program`.
    // [Plus additional comment on why we are setting this directly into the ref]
    this.dialogRef._closeInteractionType = event.screenX === 0 && event.screenY === 0 ? 'keyboard' : 'mouse';
    this.dialogRef.close(this.dialogResult);
  }
```

Then during `close`, the dialogRef would pass along that `_closeInteractionType` onto the container (or just do it from this function). This way, the API is preserved and all tests in Google would pass since `close` is still being called directly from the close directive.

This is obviously not ideal or best practice to be setting some state before calling a function, but a comment could be provided to explain why this is the case, as well as an additional test that makes sure we don't regress.

I still think it would be valuable to provide a fake to our users, but at least this won't be blocked on that migration occurring, which could possibly never get prioritized.

devversion:
@andrewseguin Thanks for looking into this. I've made the changes as discussed. Instead of manually assigning `_closeInteractionType` all the time, I've created an internal helper function that takes care of this. This will also allow us to easily remove this logic in the future if we ever sort these dialog ref mock complications out.

PR #9271, opened 2018-01-07
feat(menu): support lazy rendering and passing in context data

### LINKED ISSUE #9251: mat-menu is rendering eagerly
Bug

No menu computation/rendering should happen before user triggers a menu.

Lets say we have 1000 mat-menu components in a page, Angular will start component life-cycle for all of these components even before menu is activated.

https://stackblitz.com/angular/lrpbkepagev?file=app%2Fmenu-overview-example.html

In above example, mat-menu* components are initialized before user clicks on anything

Increases perf for single menu instance. 

All

Not that it matters. But I would like to propose a solution for this. An <ng-template/> based solution for this would also solve the problem where there are multiple triggers for single menu with different data. This could come in handy when there are multiple menus that are the same but require different data (Example: a menu for each table row)

        <p>Clicked Item: {{triggeredButton}}</p>
        <button mat-button
                matMenuTriggerFor
                [matMenuTriggerData]="'First button'"
                [matMenuTriggerNgTemplate]="perfyMenuTemplateContainer">
          First button
        </button>
        <button mat-button
                matMenuTriggerFor
                [matMenuTriggerData]="'Second button'"
                [matMenuTriggerNgTemplate]="perfyMenuTemplateContainer">
          Second button
        </button>

        <ng-template #perfyMenuTemplateContainer let-menuData>
          <mat-menu>
	    <button mat-menu-item (click)="triggeredButton = menuData">Item 1</button>
	    <button mat-menu-item (click)="triggeredButton = menuData">Item 2</button>
          </mat-menu>
        </ng-template>

crisbeto:
* Introduces the `matMenuContent` directive that allows for menu content to be rendered lazily.
* Adds the `matMenuTriggerData` input to the `MatMenuTrigger` that allows for contextual data to be passed in to the lazily-rendered menu panel. This allows for the menu instance to be re-used between triggers.

Fixes #9251.

### DIFF
--- a/src/cdk/portal/dom-portal-outlet.ts
+++ b/src/cdk/portal/dom-portal-outlet.ts
@@ -22,7 +22,8 @@ import {BasePortalOutlet, ComponentPortal, TemplatePortal} from './portal';
  */
 export class DomPortalOutlet extends BasePortalOutlet {
   constructor(
-      private _hostDomElement: Element,
+      /** Element into which the content is projected. */
+      public outletElement: Element,
       private _componentFactoryResolver: ComponentFactoryResolver,
       private _appRef: ApplicationRef,
       private _defaultInjector: Injector) {
@@ -59,7 +60,7 @@ export class DomPortalOutlet extends BasePortalOutlet {
     }
     // At this point the component has been instantiated, so we move it to the location in the DOM
     // where we want it to be rendered.
-    this._hostDomElement.appendChild(this._getComponentRootNode(componentRef));
+    this.outletElement.appendChild(this._getComponentRootNode(componentRef));
 
     return componentRef;
   }
@@ -78,7 +79,7 @@ export class DomPortalOutlet extends BasePortalOutlet {
     // But for the DomPortalOutlet the view can be added everywhere in the DOM
     // (e.g Overlay Container) To move the view to the specified host element. We just
     // re-append the existing root nodes.
-    viewRef.rootNodes.forEach(rootNode => this._hostDomElement.appendChild(rootNode));
+    viewRef.rootNodes.forEach(rootNode => this.outletElement.appendChild(rootNode));
 
     this.setDisposeFn((() => {
       let index = viewContainer.indexOf(viewRef);
@@ -96,8 +97,8 @@ export class DomPortalOutlet extends BasePortalOutlet {
    */
   dispose(): void {
     super.dispose();
-    if (this._hostDomElement.parentNode != null) {
-      this._hostDomElement.parentNode.removeChild(this._hostDomElement);
+    if (this.outletElement.parentNode != null) {
+      this.outletElement.parentNode.removeChild(this.outletElement);
     }
   }
 
--- a/src/lib/menu/menu-content.ts
+++ b/src/lib/menu/menu-content.ts
@@ -1,0 +1,70 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {
+  Directive,
+  TemplateRef,
+  ComponentFactoryResolver,
+  ApplicationRef,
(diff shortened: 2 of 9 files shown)


src/lib/menu/menu-content.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    if (!this._outlet) {
+      this._outlet = new DomPortalOutlet(this._document.createElement('div'),
+          this._componentFactoryResolver, this._appRef, this._injector);
+    }
+
+    const element: HTMLElement = this._template.elementRef.nativeElement;
+
+    // Since the menu might be attached to a different DOM node, we have to re-insert it every time.
+    element.parentNode!.insertBefore(this._outlet.hostDomElement, element);
```
jelbourn: I don't quite follow why this is necessary

src/lib/menu/menu-content.ts
```diff
@@ -0,0 +1,68 @@
...
+  Injector,
+  ViewContainerRef,
+  Inject,
+  OnDestroy,
+} from '@angular/core';
+import {TemplatePortal, DomPortalOutlet} from '@angular/cdk/portal';
+import {DOCUMENT} from '@angular/common';
+
+/**
+ * Menu content that will be rendered lazily once the menu is opened.
```
jelbourn: Would an end user need to ever interact with this class (i.e. should it be docs-private)?

src/lib/menu/menu.md
```diff
@@ -93,6 +93,46 @@ that should trigger the sub-menu:
 
 <!-- example(nested-menu) -->
 
+### Lazy rendering
+By default, the menu content will be initialized even when the panel is closed. To defer
+initialization until the menu is open, the content can be provided as an `ng-template`
+with the `matMenuContent` attribute:
+
+```html
+<mat-menu #appMenu="matMenu">
```
jelbourn: What do you think of avoiding the extra `matMenuContent` by letting people add `matMenu` to an `ng-template` element? Something like
```html
<ng-template matMenu>
  ...
</ng-template>
```
The selector for the menu would change to `mat-menu, ng-template[matMenu]`. The menu would then optionally inject `TemplateRef` and go through the extra steps if one is present.

src/lib/menu/menu-content.ts
crisbeto: I did it, because we support providing your own menu instance that matches the `MatMenuPanel` interface.

src/lib/menu/menu.md
crisbeto: That's what I was going for initially, but Angular throws an error when you have a component on an `ng-template`. The alternative would be to use a directive, but we can't do that either, because we need to wrap the menu content in a div to add the proper styling and for the animation to work.

src/lib/menu/menu-content.ts
crisbeto: It's because we can have the menu be opened by a different triggers which in turn have their own `OverlayRef` with a different pane. If we didn't do this, the element could end up in limbo if you opened it using two different triggers.

jelbourn: Add that explanation as a comment?

src/cdk/portal/dom-portal-outlet.ts
```diff
@@ -22,7 +22,8 @@ import {BasePortalOutlet, ComponentPortal, TemplatePortal} from './portal';
  */
 export class DomPortalOutlet extends BasePortalOutlet {
   constructor(
-      private _hostDomElement: Element,
+      /** Element into which the content is projected. */
+      public hostDomElement: Element,
```
jelbourn: Maybe rename this to `outletElement`? I think the "host" is leftover from the old class name

jelbourn:
@crisbeto passes Google presubmit, just needs rebase

crisbeto:
Rebased.

jelbourn:
@crisbeto failing on Travis now

crisbeto:
Fixed the test failure.

PR #9275, opened 2018-01-08
add docs dir with Table of Contents

ORESoftware:
please add docs dir with Table of Contents

### DIFF
--- a/docs/table-of-contents.md
+++ b/docs/table-of-contents.md
@@ -1,4 +1,5 @@
 
+# Material2 Table of Contents
 
 
 1. MatTable:  [a relative link](../src/lib/table/table.md)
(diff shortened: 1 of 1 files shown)


docs/table-of-contents.md
```diff
@@ -0,0 +1,6 @@
+
+# Material2 Table of Contents
```
Chan4077: Shouldn't this be "Material" or just "Table of Contents"?

ORESoftware: yeah, something like that, either way, a docs folder or examples folder would help new users

donroyco: So your PR is not ready yet?

ORESoftware: this PR is just a concept, if you like the idea, I can fill it out

jelbourn:
Our docs live on material.angular.io; we shouldn't need a TOC for markdown files on github.

PR #9283, opened 2018-01-08
fix(coercion): treat zero as falsy boolean

### LINKED ISSUE #9280: mat-select [disabled] should accept either true/false or 1/0

BUG

Consistency in the `[disabled]` flag between form elements.

`<input>` accepts either `true`/`false` or `1`/`0`.  `<mat-select>` only accepts `true`/`false`.

* Create a value that represents the `disabled` state of controls.
* Create a form with every possible type of form control element, each controlling their `disabled` state from the value created above

* Set the value to `true` - all form elements should be disabled.
* Set the value to `1` - all form elements SHOULD be disabled.
* Set the value to `false` - all form elements should be enabled.
* Set the value to `0` - all form elements should be enabled - **NOTE: `<mat-select>` is NOT**

When you have a MySQL backed data-store (and quite possibly others) and you have a boolean field, it is stored/retrieved as 1/0.  Some of the controls honor it, others don't.  This inconsistency can cause people to inadvertently inject bugs into their software without knowing it.

The system should work the same for all control types.

Not tested.  Effects latest.

DevVersion:
Right now the `coerceBooleanProperty` method coerces `0` to `true`. This is not aligning with the native behavior of DOM elements (e.g. setting the disabled property to 0).

@jelbourn This is a breaking change, because previously zero has been handled differently. I'm going to mark this as `major`. I'm not too sure if we want to handle zero that way, but it follows the DOM behavior, so I think it makes sense. If not, we can just close the issue and this PR, and keep everything as it is.

Fixes #9280.

### DIFF
--- a/src/cdk/coercion/boolean-property.spec.ts
+++ b/src/cdk/coercion/boolean-property.spec.ts
@@ -15,7 +15,7 @@ describe('coerceBooleanProperty', () => {
   });
 
   it('should coerce zero to true', () => {
-    expect(coerceBooleanProperty(0)).toBe(true);
+    expect(coerceBooleanProperty(0)).toBe(false);
   });
 
   it('should coerce the string "false" to false', () => {
--- a/src/cdk/coercion/boolean-property.ts
+++ b/src/cdk/coercion/boolean-property.ts
@@ -8,5 +8,5 @@
 
 /** Coerces a data-bound value (typically a string) to a boolean. */
 export function coerceBooleanProperty(value: any): boolean {
-  return value != null && `${value}` !== 'false';
+  return value === '' || (!!value && `${value}` !== 'false');
 }


jelbourn:
I don't think we should do this. The intent is to match the HTML attribute `disabled` with the sole exceptions of `false` and `"false"`

DevVersion:
Not sure about that. I think that it makes sense to treat zero (with type of number) as a falsy value. It's being done by the `disabled` property of the `HTMLInputElement` for example.

If someone still sets the `mixin` setter through HTML like `disabled="0"`, then the check will still return `true` as it has been before. It just handles the `number` differently.

```ts
coerceBooleanProperty(0) // -> false
coerceBooleanProperty('0') // -> true

```

andreElrico:
I have just asked a question about this on [SO](https://stackoverflow.com/questions/57545094/why-is-coercebooleanproperty-of-number-0-coerced-into-true?noredirect=1#comment101553527_57545094). Would you like to provide more intel? On why `number 0` is true?

PR #9303, opened 2018-01-09
fix(menu): set appropriate origin when restoring focus

### LINKED ISSUE #9292: mat-menu trigger button is highlighted after closing menu
Bug

Trigger button acts like other buttons: When action has completed, button returns to the neutral state.
image

... Actual keyboard focus stays, but the styling is back to neutral, when using the mouse to click (for keyboard navigation it stays lit... Which makes sense).

Trigger button stay focused, and keeps the focused style.
image

- Open https://stackblitz.com/edit/angular-material2-issue-tscawk
- Click basic button.
- Observe state of button after animation finishes.
- Attach a menu trigger to a button.
- Click button that is a mat-menu trigger.
- Close menu using backdrop or buttons
- Observe state of button, comparing with basic button.

Consistency. Other buttons does not keep the focus styling, menu triggers shouldn't either.

Seen on:
Angular 5.0.0, 5.1.3
Material 5.0.2, 5.0.3
Linux, Windows
TypeScript 2.5.3 + whatever TypeScript is running on StackBlitz
Internet Explorer 11, Chrome 63.0.3239.84, FireFox 57.0.3, EdgeHTML 16.16299

May be related to fix in angular/material2#8348 (don't really think so, though, I think I've seen this behavior for a long time. It used to have the same problem when expanding sidebars using a button)

crisbeto:
Sets the correct focus origin depending on the way a menu has been opened.

Fixes #9292.

### DIFF
--- a/src/cdk/a11y/focus-monitor.spec.ts
+++ b/src/cdk/a11y/focus-monitor.spec.ts
@@ -1,5 +1,10 @@
 import {TAB} from '@angular/cdk/keycodes';
-import {dispatchFakeEvent, dispatchKeyboardEvent, dispatchMouseEvent} from '@angular/cdk/testing';
+import {
+  dispatchFakeEvent,
+  dispatchKeyboardEvent,
+  dispatchMouseEvent,
+  patchElementFocus,
+} from '@angular/cdk/testing';
 import {Component} from '@angular/core';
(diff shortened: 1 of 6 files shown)


crisbeto:
Included the `patchElementFocus` changes from #9257 to help with some flaky IE tests.

PR #9305, opened 2018-01-09
fix(focus-monitor): implement OnDestroy logic

crisbeto:
Similarly to most of the other DOM-related services, these changes implement `ngOnDestroy` for the `FocusMonitor` which removes the monitoring from all of the currently-monitored elements.

### DIFF
--- a/src/cdk/a11y/focus-monitor.ts
+++ b/src/cdk/a11y/focus-monitor.ts
@@ -42,7 +42,7 @@ type MonitoredElementInfo = {
 
 /** Monitors mouse and keyboard events to determine the cause of focus events. */
 @Injectable()
-export class FocusMonitor {
+export class FocusMonitor implements OnDestroy {
   /** The focus origin that the next focus event is a result of. */
   private _origin: FocusOrigin = null;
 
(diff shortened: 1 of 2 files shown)


src/cdk/a11y/focus-monitor.ts
```diff
@@ -58,8 +58,8 @@ export class FocusMonitor {
   /** The timeout id of the touch timeout, used to cancel timeout later. */
   private _touchTimeout: number;
 
-  /** Weak map of elements being monitored to their info. */
-  private _elementInfo = new WeakMap<Element, MonitoredElementInfo>();
+  /** Map of elements being monitored to their info. */
+  private _elementInfo = new Map<HTMLElement, MonitoredElementInfo>();
```
crisbeto: Note: I would prefer to keep this as a `WeakMap`, but we need a `Map` in order to be able to iterate over the elements. Alternatively, if we make an assumption that all the monitored elements will be removed on destroy, we could remove only the global listeners.

mmalerba:
Does `ngOnDestroy` get automatically called for Injectables? I was under the impression it didn't and that the lifecycle methods only applied to Components and Directives.

jelbourn:
@mmalerba is gets called if the module is unloaded, which very large apps (like GCP) do.

src/cdk/a11y/focus-monitor.ts
jelbourn: This is a tricky trade-off. We don't have any data to know for sure, but I expect _most_ apps will end up with `FocusMonitor` provided at the root, meaning that using `Map` could end up making elements remain in memory after they'd otherwise be removed. 

Do you think it would be common that monitored elements would remain while the module with the `FocusMonitor` would be unloaded?

jelbourn: You're right, the `unlisten` would preserve the element reference already, so switching to a `Map` won't change anything

Issue #8592, opened 2017-11-22
mat-expansion-panel-header expand/collapse animation jumps on non default header sizes


efsinua:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
Smooth animation transition of \<mat-expansion-panel-header\> on expand/collapse if header height is not default.

#### What is the current behavior?
When \<mat-expansion-panel-header\> height is changed with styling, animation jumps on expand/collapse to hard coded sizes at expansionHeight animation trigger to collapsedHeight: 48px or expandedHeight: 64px instead of smooth transition

#### What are the steps to reproduce?
http://plnkr.co/edit/zIEeYb?p=preview

#### What is the use-case or motivation for changing an existing behavior?
Possibility to override default \<mat-expansion-panel-header\> height

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
all including Angular Material 5.0.0-rc.1 

#### Is there anything else we should know?
if replace values at expansion-panel-header.ts at animations trigger('expansionHeight',...)
collapsedHeight: '48px' to collapsedHeight: '\*'
expandedHeight: '64px' to  expandedHeight: '\*'

expand/collapse animation will work smoothly

amouly:
Verified, we have te same problem when changing the height of the header.

atluriajith:
Any update on this? [#5641](https://github.com/angular/material2/issues/5641) stating that to height and width as a attributes but can we handle this from CSS?.

josephperrott:
This is actually working as expected.

If the collapsedHeight and expandedHeight are left to default, they will be [`48px` and `64px` respectively](https://github.com/angular/material2/blob/master/src/lib/expansion/expansion-animations.ts#L38).  To define the height via css, you will need to set the parameters for the animation to `*` for both.

atluriajith:
@josephperrott  Thanks for the update  :thumbsup:

josephperrott:
Closed by #9313

omryha:
I,

I am having this problem, using latest version of Material (6.3.3).
I'm defining a custom width and height for the header:

`<mat-expansion-panel-header [collapsedHeight]="'85px'" [expandedHeight]="expandedHeight">`
("expandedHeight" is a variable that is changing by TS code).

I didn't get the solution for this problem from this post, I'd be happy if someone could help me with this. 

Thanks.

elrabelo:
[...]

PR #9313, opened 2018-01-09
fix(expansion): define default expansion header heights via css.

josephperrott:
Addresses: #8592

### DIFF
--- a/src/lib/expansion/expansion-animations.ts
+++ b/src/lib/expansion/expansion-animations.ts
@@ -35,12 +35,12 @@ export const matExpansionAnimations: {
     state('collapsed', style({
       height: '{{collapsedHeight}}',
     }), {
-      params: {collapsedHeight: '48px'},
+      params: {collapsedHeight: '*'},
     }),
     state('expanded', style({
       height: '{{expandedHeight}}'
     }), {
-      params: {expandedHeight: '64px'}
+      params: {expandedHeight: '*'}
     }),
     transition('expanded <=> collapsed', animate(EXPANSION_PANEL_ANIMATION_TIMING)),
   ]),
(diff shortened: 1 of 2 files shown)


josephperrott:
Fixed author to be correct email for CLA.

josephperrott:
@jelbourn @crisbeto 

I think this might be a better way to go about allowing custom heights for expansion panel headers as it will allow customization using either css or binding in the template.  I want to confirm that there is not something I am missing that would cause issues here.

src/lib/expansion/expansion-animations.ts
```diff
@@ -35,12 +35,12 @@ export const matExpansionAnimations: {
...
       height: '{{collapsedHeight}}',
     }), {
-      params: {collapsedHeight: '48px'},
+      params: {collapsedHeight: '*'},
     }),
     state('expanded', style({
       height: '{{expandedHeight}}'
     }), {
-      params: {expandedHeight: '64px'}
+      params: {expandedHeight: '*'}
```
crisbeto: I agree that it's a better way to handle custom height, but AFAIK it will break the people that are using the binding as it's set up right now. We could either remove the inputs and leave these changes for a later release, or deprecate them and set up the bindings so they just set the height as an inline style.

josephperrott: It shouldn't break those using it as it is currently since we define the same expanded and collapsed heights (64px and 48px) in css rather than the animation parameters.

If a user set it by changing the animation params, it should be respected as it will be put into the style tag and override our css rule's height.
If a user set it by setting the params to `*` and defined the css rule's, then it results as the same how the new set up works.

Am I missing a scenario that would break?

crisbeto: Sorry, I probably got confused last time. Looking at it again, it should be fine.

PR #9316, opened 2018-01-10
feat(virtual-scroll): fixed size virtual scroll

mmalerba:


src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+        this._renderedRangeSubject.next(this._renderedRange = range);
+        this._changeDetectorRef.markForCheck();
+      });
+    }
+  }
+  private _renderedRange: Range = {start: 0, end: 0};
+
+  /** The offset of the rendered portion of the data from the start. */
+  get renderedContentOffset() { return this._renderedContentOffset; }
+  set renderedContentOffset(offset) {
```
crisbeto: Add a type to the param?

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+  measureScrollOffset() {
+    return this.orientation === 'horizontal' ?
+        this.elementRef.nativeElement.scrollLeft : this.elementRef.nativeElement.scrollTop;
+  }
+
+  ngOnInit() {
+    Promise.resolve().then(() => {
+      this._viewportSize = this.orientation === 'horizontal' ?
+          this.elementRef.nativeElement.clientWidth : this.elementRef.nativeElement.clientHeight;
+      fromEvent(this.elementRef.nativeElement, 'scroll').subscribe(() => {
```
crisbeto: This one should be run outside the `NgZone`.

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /** Gets the current scroll offset of the viewport. */
+  measureScrollOffset() {
+    return this.orientation === 'horizontal' ?
+        this.elementRef.nativeElement.scrollLeft : this.elementRef.nativeElement.scrollTop;
+  }
+
+  ngOnInit() {
+    Promise.resolve().then(() => {
```
crisbeto: Rather than wrapping it in a promise, have you tried running it in one of the later lifecycle hooks?

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+    const previousViews: EmbeddedViewRef<CdkForOfContext<T>>[] = [];
+    for (let i = 0, len = this._viewContainerRef.length; i < len; i++) {
+      previousViews.unshift(
+          this._viewContainerRef.detach()! as EmbeddedViewRef<CdkForOfContext<T>>);
+    }
+
+    // Mark the removed indices so we can recycle their views.
+    changes.forEachRemovedItem(record => {
+      this._templateCache.push(previousViews[record.previousIndex!]);
+      delete previousViews[record.previousIndex!];
```
crisbeto: Using `delete` here won't reindex the array, is that intended? Otherwise it might be better to use `splice` to avoid some unintended side-effects. Same goes for the other `delete` usages.

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+  private _changeDataSource(oldDs: DataSource<T> | null, newDs: DataSource<T>): Observable<T[]> {
+    if (oldDs) {
+      oldDs.disconnect(this);
+    }
+    this._needsUpdate = true;
+    return newDs.connect(this);
+  }
+
+  /** Apply changes to the DOM. */
+  private _applyChanges(changes: IterableChanges<T> | null) {
```
crisbeto: This method is getting kinda long, consider splitting it into a few smaller ones?

src/cdk/collections/data-source.ts
```diff
@@ -7,8 +7,10 @@
  */
 
 import {Observable} from 'rxjs/Observable';
+import {of} from 'rxjs/observable/of';
```
CaerusKaru: Should be `import {of as observableOf} from 'rxjs/observable/of'`

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+   * @param index The index of the data element whose client rect we want to measure.
+   * @return The combined client rect for all DOM elements rendered as part of the given index.
+   *     Or null if no DOM elements are rendered for the given index.
+   * @throws If the given index is not in the rendered range.
+   */
+  measureClientRect(index: number): ClientRect | null {
+    if (index < this._renderedRange.start || index >= this._renderedRange.end) {
+      throw Error('Error: attempted to measure an element that isn\'t rendered.');
+    }
+    index -= this._renderedRange.start;
```
jelbourn: Introduce a new variable instead of modifying a function argument? `renderedIndex`?

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  /** Emits when the data source changes. */
+  private _dataSourceSubject = new Subject<DataSource<T>>();
+
+  /** The DataSource to display. */
+  @Input()
+  get cdkForOf(): NgIterable<T> | DataSource<T> { return this._cdkForOf; }
+  set cdkForOf(value: NgIterable<T> | DataSource<T>) {
+    this._cdkForOf = value;
+    let ds = value instanceof DataSource ? value :
```
jelbourn: Prefer `const` when the variable doesn't change

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+   * @throws If the given index is not in the rendered range.
+   */
+  measureClientRect(index: number): ClientRect | null {
+    if (index < this._renderedRange.start || index >= this._renderedRange.end) {
+      throw Error('Error: attempted to measure an element that isn\'t rendered.');
+    }
+    index -= this._renderedRange.start;
+    let view = this._viewContainerRef.get(index) as EmbeddedViewRef<CdkForOfContext<T>> | null;
+    if (view && view.rootNodes.length) {
+      let minTop = Infinity;
```
jelbourn: Comment that describes these bounds?

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+      let minLeft = Infinity;
+      let maxBottom = -Infinity;
+      let maxRight = -Infinity;
+
+      // There may be multiple root DOM elements for a single data element, so we merge their rects.
+      for (let i = 0, ilen = view.rootNodes.length; i < ilen; i++) {
```
jelbourn: Avoid `ilen` by iterating backwards?

src/cdk/scrolling/for-of.ts
jelbourn: Agreed, it does quite a lot.

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+
+
+/** Virtual scrolling strategy for lists with items of known fixed size. */
+export class VirtualScrollFixedSizeStrategy implements VirtualScrollStrategy {
+  private _viewport: CdkVirtualScrollViewport;
+
+  constructor(public itemSize: number, public bufferSize: number) {}
+
+  /** Initialize the strategy and specify the viewport it will be working with. */
+  init(viewport: CdkVirtualScrollViewport) {
```
jelbourn: For position and scroll strategies we use `attach` as the method for connecting to the subject of the strategy; we could be consistent with that here.

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+    this._viewport = viewport;
+    this._viewport.totalContentSize = this._viewport.dataLength * this.itemSize;
+    this._updateRenderedRange();
+  }
+
+  /** Re-initialize the strategy with the same viewport. */
+  reinit() {
```
jelbourn: Why is this method necessary?

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (this._viewport) {
+      this.init(this._viewport);
+    }
+  }
+
+  onContentScrolled() {
+    this._updateRenderedRange();
+  }
+
+  onDataLengthChanged() {
```
jelbourn: For `onContentScrolled` and `onDataLengthChanged`: I'm not a huge fan of methods that are named for when they're meant to be called rather than what they do. Is there any reason they both could be an `update` method?

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+    this._viewport.totalContentSize = this._viewport.dataLength * this.itemSize;
+    this._updateRenderedRange();
+  }
+
+  private _updateRenderedRange() {
```
jelbourn: Method descriptions?

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+
+
+export function _virtualScrollFixedSizeStrategyFactory(fixedSizeDir: CdkVirtualScrollFixedSize) {
+  return fixedSizeDir._scrollStrategy;
+}
+
+
+/** A virtual scroll strategy that supports fixed-size items. */
+@Directive({
+  selector: 'cdk-virtual-scroll-viewport[itemSize]',
```
jelbourn: `[itemSize]`?

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+    provide: VIRTUAL_SCROLL_STRATEGY,
+    useFactory: _virtualScrollFixedSizeStrategyFactory,
+    deps: [forwardRef(() => CdkVirtualScrollFixedSize)],
+  }],
+})
+export class CdkVirtualScrollFixedSize implements OnChanges {
+  /** The size of the items in the list. */
+  @Input() itemSize = 20;
+
+  /** The number of extra elements to render on either side of the viewport. */
```
jelbourn: "viewport" -> "scrolling viewport"?

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+@Directive({
+  selector: 'cdk-virtual-scroll-viewport[itemSize]',
+  providers: [{
+    provide: VIRTUAL_SCROLL_STRATEGY,
+    useFactory: _virtualScrollFixedSizeStrategyFactory,
+    deps: [forwardRef(() => CdkVirtualScrollFixedSize)],
+  }],
+})
+export class CdkVirtualScrollFixedSize implements OnChanges {
+  /** The size of the items in the list. */
```
jelbourn: "The rendered size of each item in the list (in pixels)" ?

src/cdk/scrolling/virtual-scroll-viewport.html
```diff
@@ -0,0 +1,8 @@
+<div #contentWrapper class="cdk-virtual-scroll-content-wrapper"
```
jelbourn: Add comment like
```html
<!-- Wrap the scrolling content in an element that will... -->
```

src/cdk/scrolling/virtual-scroll-viewport.html
```diff
@@ -0,0 +1,8 @@
...
+     [style.transform]="_renderedContentTransform">
+  <ng-content></ng-content>
+</div>
+<div class="cdk-virtual-scroll-spacer"
```
jelbourn: Add comment like
```
Spacer used to force the scrolling container to the correct size for the
*total* number of items so that the scrollbar captures the entire data set.
```

src/cdk/scrolling/virtual-scroll-viewport.scss
```diff
@@ -0,0 +1,20 @@
+cdk-virtual-scroll-viewport {
```
jelbourn: Add comments here that explain how these different elements fit together since it's pretty fundamental to the way that the virtual-scroller works?

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  private _renderedRangeSubject = new Subject<Range>();
+
+  /** The direction the viewport scrolls. */
+  @Input() orientation: 'horizontal' | 'vertical' = 'vertical';
+
+  /** The element that wraps the rendered content. */
+  @ViewChild('contentWrapper') _contentWrapper: ElementRef;
+
+  /** The total size of all content, including content that is not currently rendered. */
```
jelbourn: It's not clear from the description or the name whether this is rendered size (pixels) or length of data.

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+  get totalContentSize() { return this._totalContentSize; }
+  set totalContentSize(size: number) {
+    if (this._totalContentSize != size) {
+      this._ngZone.run(() => {
```
jelbourn: Comment that explains why you need to re-enter the zone here?
(and elsewhere)

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  /** The direction the viewport scrolls. */
+  @Input() orientation: 'horizontal' | 'vertical' = 'vertical';
+
+  /** The element that wraps the rendered content. */
+  @ViewChild('contentWrapper') _contentWrapper: ElementRef;
+
+  /** The total size of all content, including content that is not currently rendered. */
+  get totalContentSize() { return this._totalContentSize; }
+  set totalContentSize(size: number) {
```
jelbourn: These properties shouldn't need to be be getter/setters since they're not `@Input`s for the component (could have an update method, which is more obviously side-effecty)

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+  private _dataLength = 0;
+
+  /** The size of the viewport. */
+  get viewportSize() { return this._viewportSize; }
+  private _viewportSize = 0;
+
+  /** A stream that emits whenever the rendered range changes. */
+  renderedRangeStream: Observable<Range> = this._renderedRangeSubject.asObservable();
+
+  _renderedContentTransform: string;
```
jelbourn: Description?

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+  private _scrollHandledStatus: 'needed' | 'pending' | 'done' = 'done';
+
+  private _scrollStrategyInited = false;
+
+  constructor(public elementRef: ElementRef, private _changeDetectorRef: ChangeDetectorRef,
+              private _ngZone: NgZone,
+              @Inject(VIRTUAL_SCROLL_STRATEGY) private _scrollStrategy: VirtualScrollStrategy) {}
+
+  /** Connect a `CdkForOf` to this viewport. */
+  connect(forOf: CdkForOf<any>) {
```
jelbourn: This isn't formalized anywhere, but I feel that `attach` and `detach` are more appropriate for components that interact than `connect`/`disconnect`

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (this._connected) {
+      throw Error('CdkVirtualScrollViewport is already connected.');
+    }
+
+    this._connected = true;
+    forOf.dataStream.pipe(takeUntil(this._disconnectSubject)).subscribe(data => {
```
jelbourn: Add a comment that explains what's happening here?

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+/**
+ * A directive similar to `ngForOf` to be used for rendering data inside a virtual scrolling
+ * container.
+ */
+@Directive({
+  selector: '[cdkFor][cdkForOf]',
+})
+export class CdkForOf<T> implements CollectionViewer, DoCheck, OnDestroy {
+  /** Emits when the rendered view of the data changes. */
+  viewChange = new Subject<Range>();
```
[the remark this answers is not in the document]
mmalerba: This has to be called `viewChange` because it's implementing `CollectionViewer`

src/cdk/scrolling/for-of.ts
mmalerba: Yes this is intentional. I need each element to be at its correct original index so it matches up with the index in the IterableChangeRecord. I then use `delete` to mark the view as having been handled in some way (cached or queued up for re-insertion).

src/cdk/scrolling/virtual-scroll-viewport.ts
mmalerba: I tried, it seemed like they were not late enough and I had to use the promise

src/cdk/scrolling/for-of.ts
jelbourn: Would it work to to `null`? That's generally much faster than using `delete`

src/cdk/scrolling/for-of.ts
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  /**
+   * Get the client rect for the given index.
+   * @param index The index of the data element whose client rect we want to measure.
+   * @return The combined client rect for all DOM elements rendered as part of the given index.
+   *     Or null if no DOM elements are rendered for the given index.
+   * @throws If the given index is not in the rendered range.
+   */
+  measureClientRect(index: number): ClientRect | null {
```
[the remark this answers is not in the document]
mmalerba: No, this is eventually intended to be used by another virtual scrolling strategy that measures the elements to determine their size.

src/cdk/scrolling/virtual-for-of.ts
```diff
@@ -0,0 +1,310 @@
...
+        height: maxBottom - minTop,
+        width: maxRight - minLeft
+      };
+    }
+    return null;
+  }
+
+  ngDoCheck() {
+    if (this._differ && this._needsUpdate) {
+      const changes = this._differ.diff(this._renderedItems);
```
amcdnl: This is really pricey, is there a more reactive way we can do this updating?

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,191 @@
...
+  }
+  private _renderedRange: Range = {start: 0, end: 0};
+
+  /** The offset of the rendered portion of the data from the start. */
+  get renderedContentOffset(): number { return this._renderedContentOffset; }
+  set renderedContentOffset(offset: number) {
+    if (this._renderedContentOffset != offset) {
+      this._ngZone.run(() => {
+        this._renderedContentOffset = offset;
+        this._renderedContentTransform = this.orientation === 'horizontal' ?
```
amcdnl: Might want to do browser prefix test here.

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,191 @@
...
+            `translateX(${offset}px)`: `translateY(${offset}px)`;
+        this._changeDetectorRef.markForCheck();
```
amcdnl: I'd consider just doing `detectChanges` to get it to run immediately.

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,191 @@
...
+  }
+  private _totalContentSize = 0;
+
+  /** The currently rendered range of indices. */
+  get renderedRange() { return this._renderedRange; }
+  set renderedRange(range: Range) {
+    if (!this._rangesEqual(this._renderedRange, range)) {
+      this._ngZone.run(() => {
+        this._renderedRangeSubject.next(this._renderedRange = range);
+        this._changeDetectorRef.markForCheck();
```
amcdnl: I'd consider just doing `detectChanges` to get it to run immediately.

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,191 @@
...
+  /** The element that wraps the rendered content. */
+  @ViewChild('contentWrapper') _contentWrapper: ElementRef;
+
+  /** The total size of all content, including content that is not currently rendered. */
+  get totalContentSize() { return this._totalContentSize; }
+  set totalContentSize(size: number) {
+    if (this._totalContentSize != size) {
+      this._ngZone.run(() => {
+        this._totalContentSize = size;
+        this._changeDetectorRef.markForCheck();
```
amcdnl: I'd consider just doing `detectChanges` to get it to run immediately.

src/cdk/scrolling/virtual-scroll-viewport.ts
```diff
@@ -0,0 +1,191 @@
...
+      });
+    }
+  }
+  private _renderedRange: Range = {start: 0, end: 0};
+
+  /** The offset of the rendered portion of the data from the start. */
+  get renderedContentOffset(): number { return this._renderedContentOffset; }
+  set renderedContentOffset(offset: number) {
+    if (this._renderedContentOffset != offset) {
+      this._ngZone.run(() => {
```
amcdnl: I'd try to avoid running this in a zone.

src/cdk/scrolling/virtual-scroll-fixed-size.ts
```diff
@@ -0,0 +1,83 @@
...
+  @Input() itemSize = 20;
```
amcdnl: It would be nice if you could make this accept a function too where a user could pass a callback that accepted the current item they were iterating over to return a custom height for that element. For example:

```
itemSize = (item) => item.type === 'description' ? 50 : 20;
```

src/cdk/scrolling/virtual-for-of.ts
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  /**
+   * Get the client rect for the given index.
+   * @param index The index of the data element whose client rect we want to measure.
+   * @return The combined client rect for all DOM elements rendered as part of the given index.
+   *     Or null if no DOM elements are rendered for the given index.
+   * @throws If the given index is not in the rendered range.
+   */
+  measureClientRect(index: number): ClientRect | null {
```
amcdnl: You might want to add a window resize function to update dims on window resizes.

src/cdk/scrolling/virtual-for-of.ts
```diff
@@ -0,0 +1,310 @@
...
+    if (oldDs) {
+      oldDs.disconnect(this);
+    }
+    this._needsUpdate = true;
+    return newDs.connect(this);
+  }
+
+  /** Update the `CdkVirtualForOfContext` for all views. */
+  private _updateContext() {
+    for (let i = 0, len = this._viewContainerRef.length; i < len; i++) {
```
amcdnl: Given the sensitivity of perf here, you might want to consider a faster loop technique. https://jsperf.com/loops/33

src/cdk/scrolling/virtual-scroll-fixed-size.ts
mmalerba: I gave it a better name so its more obvious why its necessary

src/cdk/scrolling/virtual-scroll-fixed-size.ts
mmalerba: I don't want to add this directive to other random components that happen to have an `"itemSize"` and I also didn't want people to have to write:
`<cdk-virtual-scroll-viewport [cdkVirtualScrollItemSize]="itemSize" [cdkVirtualScrollBufferSize]="bufferSize">`

so this seemed like a good way to set it up

src/cdk/scrolling/virtual-for-of.ts
mmalerba: Responding to resize is a feature I plan to add, but not in this PR

src/cdk/scrolling/virtual-for-of.ts
mmalerba: I think we could differentiate between needing an update due to scrolling and rendering a new portion of the list (can do something simpler) vs needing an update due to the underlying data changing (in which case I think we'd have to do this diff). I added a TODO to add some logic for that. Also added a TODO to calculate the minimum set of remove & inserts since I'm currently just removing all of the views and re-inserting them in the right order

src/cdk/scrolling/virtual-scroll-fixed-size.ts
mmalerba: Yep that's something I want to add in a future PR

src/cdk/scrolling/virtual-scroll-viewport.ts
mmalerba: I thought (and I could definitely be wrong) that it was better to just mark for check so that angular can go ahead and do these updates, together with any other updates that need to happen, at the same time and avoid thrashing.

@jelbourn @crisbeto do you guys know which way is best for performance?

amcdnl: In other projects, I was running `markForCheck` in a lot of places, when I switched it to `detectChanges` I saw a huge perf jump when scrolling.

src/cdk/scrolling/virtual-scroll-viewport.ts
mmalerba: We only have to support back to IE11 (+ latest 2 versions of other browsers) so should be ok

src/cdk/scrolling/virtual-scroll-viewport.ts
mmalerba: The way its set up right now I need to, so the `markForCheck` will work, but I'll investigate calling `detectChanges` directly

mmalerba:
Thanks for the feedback everyone. I've put a number of TODOs for things I didn't get to in this PR. I'm going to merge this into the virtual-scroll feature branch now, but there will be plenty more chances to comment before it gets merged to master.

PR #9318, opened 2018-01-10
feat(layout): add window breakpoints from spec

### LINKED ISSUE #9212: Implement window breakpoints from Material guidelines
feature request

That `BreakpointObserver` has all Material guidelines breakpoints.

`BreakpointObserver` has the `handset` breakpoints and the `web` breakpoints (which isn't in the guidelines).

Be true to the Material guidelines specs and improve usability of the lib by allowing the user to not have to write more `MediaMatcher` code (media queries by hand) to achieve something in the guidelines.

Angular Material ^5.0.0

Maybe a usefull image.

image

So the breakpoints could be:

```javascript
Breakpoints += [ WindowXSmall, WindowSmall, WindowMedium, WindowLarge, WindowXLarge ]
```

CDDelta:
Fixes #9212

### DIFF
--- a/src/cdk/layout/breakpoints.ts
+++ b/src/cdk/layout/breakpoints.ts
@@ -8,6 +8,12 @@
 // PascalCase is being used as Breakpoints is used like an enum.
 // tslint:disable-next-line:variable-name
 export const Breakpoints = {
+  WindowXSmall: '(max-width: 599px)',
+  WindowSmall: '(min-width: 600px) and (max-width: 959px)',
+  WindowMedium: '(min-width: 960px) and (max-width: 1279px)',
+  WindowLarge: '(min-width: 1280px) and (max-width: 1919px)',
+  WindowXLarge: '(min-width: 1920px)',
+
   Handset: '(max-width: 599px) and (orientation: portrait), ' +
            '(max-width: 959px) and (orientation: landscape)',
   Tablet: '(min-width: 600px) and (max-width: 839px) and (orientation: portrait), ' +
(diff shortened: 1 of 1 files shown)


src/cdk/layout/breakpoints.ts
```diff
@@ -8,6 +8,12 @@
 // PascalCase is being used as Breakpoints is used like an enum.
 // tslint:disable-next-line:variable-name
 export const Breakpoints = {
+  WindowXSmall: '(max-width: 599px)',
+  WindowSmall: '(min-width: 600px) and (max-width: 959px)',
+  WindowMedium: '(min-width: 960px) and (max-width: 1279px)',
+  WindowLarge: '(min-width: 1280px) and (max-width: 1919px)',
+  WindowXLarge: '(min-width: 1920px)',
```
jelbourn: @josephperrott I looked at [the spec](https://material.io/guidelines/layout/responsive-ui.html) and it's not clear to me what "Window" even means in that table. Is it supposed be be a "desktop browser window"?

josephperrott: I believe so.  That mostly is coming from listing breakpoints "across desktop, mobile and tablet".  And mobile (written elsewhere as handset) and tablet are referenced so the others must be desktop?

jelbourn: Would it make sense to make these `DesktopSmall`, `DesktopMedium`, etc.?

It would also be nice to confirm with the Material Design team.

josephperrott: After some discussion, @jelbourn and I determined that we think the best option would be to drop the prefix of the sizes entirely.

```
XSmall: '(max-width: 599px)',
Small: '(min-width: 600px) and (max-width: 959px)',
Medium: '(min-width: 960px) and (max-width: 1279px)',
Large: '(min-width: 1280px) and (max-width: 1919px)',
XLarge: '(min-width: 1920px)',
```

CDDelta: Done

PR #9319, opened 2018-01-10
docs(dialog): document how to set default options using MAT_DEFAULT_DIALOG_OPTIONS

CDDelta:
Also fixes a small typo in `mat-form-field` docs.

### DIFF
--- a/src/lib/dialog/dialog.md
+++ b/src/lib/dialog/dialog.md
@@ -39,6 +39,16 @@ export class YourDialog {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


src/lib/dialog/dialog.md
```diff
@@ -39,6 +39,16 @@ export class YourDialog {
 }
 ```
 
+Default dialog options can be specified by providing an instance of `MatDialogConfig` for MAT_DIALOG_DEFAULT_OPTIONS in your application's root module.
```
jelbourn: Give this block a header:
```markdown
### Specifying global configuration defaults
```

Also line-break at 100 columns.

PR #9326, opened 2018-01-10
feat(portal): expose attached result in CdkPortalOutlet

### LINKED ISSUE #9304: feat: providing a way to get the component attached to a CdkPortalOutlet
Feature request

When attaching a `ComponentPortal` to a `CdkPortalOutlet` using the declarative syntaxe `<ng-template [cdkPortalOutlet]="myComponentPortal">`, there is no way to get the instance of the component, created by the directive.

The only way to get the component instance is to attach the portal programmaticaly, but it would seem that the cycle is not as clean as using the directive directly. When doing it in the `ngOnInit`, the view isn't initialized, and inside the `ngAfterViewInit`, I get the "change after it was checked" error.

Dynamically render a component, based on some `type` attribute. The developer can register it's own component to be rendered if the `type` matches. All these components implement the same interface and I need my parent component to access some methods in it. To do so, I need to get the instance created during the portal attachment.

Angular: 5.1.3
Cdk: 5.0.3
OS: Windows 10
TS: 2.4.2
Browser: Chrome 63

crisbeto:
Exposes the attach `ComponentRef` or `EmbeddedViewRef` in the `CdkPortalOutlet` directive.

Fixes #9304.

### DIFF
--- a/src/cdk/portal/portal-directives.ts
+++ b/src/cdk/portal/portal-directives.ts
@@ -17,6 +17,8 @@ import {
   OnDestroy,
   OnInit,
   Input,
+  EventEmitter,
+  Output,
 } from '@angular/core';
 import {Portal, TemplatePortal, ComponentPortal, BasePortalOutlet} from './portal';
 
@@ -35,6 +37,11 @@ export class CdkPortal extends TemplatePortal<any> {
   }
 }
 
+/**
+ * Possible attached references to the CdkPortalOutlet.
+ */
+export type CdkPortalOutletAttachedRef = ComponentRef<any> | EmbeddedViewRef<any> | null;
+
 
 /**
  * Directive version of a PortalOutlet. Because the directive *is* a PortalOutlet, portals can be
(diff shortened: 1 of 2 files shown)


src/cdk/portal/portal-directives.ts
```diff
@@ -93,13 +96,19 @@ export class CdkPortalOutlet extends BasePortalOutlet implements OnInit, OnDestr
     this._attachedPortal = portal;
   }
 
+  /** Component or view reference that is attached to the portal. */
+  get result(): ComponentRef<any> | EmbeddedViewRef<any> | null {
```
jelbourn: Would it make more sense for the result to be emitted via `@Output` rather than directly read?

crisbeto: I'm not sure whether that would be super convenient. People would still have to save a reference to it somewhere if they want to do anything with it.

jelbourn: I think I'm being thrown off by calling it "result"; how about something like "attachedRef"?

jelbourn: Would be flexible to have an `attached` output as well thinking about it.

crisbeto: I like `attachedRef`. Initially I was going to do `instance`, but it would've gotten weird with the `ComponentRef.instance`. I've added the `attached` output as well.

PR #9330, opened 2018-01-11
docs(sidenav) property binding

debudda:
Should be a property binding according to docs.

### DIFF
--- a/src/material-examples/sidenav-overview/sidenav-overview-example.html
+++ b/src/material-examples/sidenav-overview/sidenav-overview-example.html
@@ -1,5 +1,5 @@
 <mat-sidenav-container class="example-container" *ngIf="shouldRun">
-  <mat-sidenav mode="side" [opened]="true">Sidenav content</mat-sidenav>
+  <mat-sidenav mode="side" opened>Sidenav content</mat-sidenav>
   <mat-sidenav-content>Main content</mat-sidenav-content>
 </mat-sidenav-container>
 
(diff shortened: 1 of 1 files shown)


src/material-examples/sidenav-overview/sidenav-overview-example.html
```diff
@@ -1,5 +1,5 @@
 <mat-sidenav-container class="example-container" *ngIf="shouldRun">
-  <mat-sidenav mode="side" opened="true">Sidenav content</mat-sidenav>
+  <mat-sidenav mode="side" [opened]="true">Sidenav content</mat-sidenav>
```
jelbourn: This can actually just be `opened` with no assignment
```html
<mat-sidenav mode="side" opened>
```

debudda: Yeah, you're right! 
But maybe it should be more verbose for the sake of clarity? Brief search showed that it is common for the docs to have `[bSome]="true"` rather than just `bSome`.

jelbourn: In general I prefer to use the short form, otherwise people will interpret the examples as _needing_ to be written out the long way.

debudda: Can I interpret that as a suggestion to fix every example in the docs according to your preferences?

jelbourn:
@debudda do you want to make the change above to merge this PR?

Issue #8880, opened 2017-12-08
mat-tab-link component should be disabled if disabled is set 


denpiv:
#### Bug, feature request, or proposal:
bug: mat-tab-link can not be disabled any more since moved to 5.0 (worked earlier)

#### What is the expected behavior?
mat-tab-link component should be disabled if disabled is set 

#### What is the current behavior?
mat-tab-link component is enabled if disabled is set 

#### What are the steps to reproduce?
https://stackblitz.com/edit/angular-cnfz79?file=app%2Ftabs-overview-example.html

#### What is the use-case or motivation for changing an existing behavior?
Tab sometimes should be disabled

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular 5.1.0
Material 5.0.0
TypeScript 2.4.2
OS Ubuntu 16.04
Browser Chrome Version 62.0.3202.75 (Official Build) (64-bit)

DevVersion:
@josephperrott This seems to be caused by https://github.com/angular/material2/pull/7364. Since `pointer-events` are now enabled, the `:focus` pseudo selector will work. If you still have a `[routerLink]` specified on a disabled link, the route will become active on click.

Also while we are at it, I don't think that the attribute `tabIndex` is valid. The real attribute would be `tabindex` in lowercase.

josephperrott:
After more investigation, this is actually caused by routerLink not being able to be disabled.  After discussing with the core team, I opened [this issue](https://github.com/angular/angular/issues/21457) to allow for `routerLink` to respect `event.preventDefault()`.

We are able to partially address this within Material by calling `preventDefault()` on click, which will properly be handled for normal `<a>` without `routerLink`s.

A potential workaround for the issue with `routerLink` is to set the route to be the same as the current page when the link should be disabled.

josephperrott:
Closed by #9357

PR #9357, opened 2018-01-12
fix(tab-link): preventDefault action when disabled tab link is clicked.

josephperrott:
Addresses #8880


src/lib/tabs/tab-nav-bar/tab-nav-bar.spec.ts
```diff
@@ -125,6 +125,23 @@ describe('MatTabNavBar', () => {
         .toBe(true, 'Expected element to no longer be keyboard focusable if disabled.');
     });
 
+    it('should prevent default action for clicks if links are disabled', () => {
+      const tabLinkElement = fixture.debugElement.query(By.css('a')).nativeElement;
+
+      const mouseEvent = new MouseEvent('click');
```
DevVersion: Can you use `createMouseEvent('click')` instead? This uses a more compatible (e.g. IE) logic for initializing the event.

PR #9358, opened 2018-01-12
fix(overlay): unable to change CdkConnectedOverlay origin dynamically

### LINKED ISSUE #9353: CdkConnectedOverlay does not reposition if its overlay origin changes
Bug

A connected overlay should reposition itself to the connected overlay origin when the overlay origin changed after the connected overlay was first displayed.

The connected overlay does not reposition itself when its overlay origin changes, even if the overlay is detached before the origin changes and reattached after the origin has changed

https://stackblitz.com/edit/angular-9vet83?file=app%2Fapp.component.html

A connected overlay could be reused for multiple overlay origins.

angular/cdk: 5.0.4

The `CdkConnectedOverlay` directive seems to only create a new overlay once, unless the overlay is completely destroyed. Listening to changes on the overlay origin and recreating the positioning strategy to connect to the new origin would probably solve the problem.

I can try to create a PR if this would be of any help.

crisbeto:
Fixes the `CdkConnectedOverlay` not handling changes of its `origin` after it's opened for the first time.

Fixes #9353.

### DIFF
--- a/src/cdk/overlay/position/connected-position-strategy.ts
+++ b/src/cdk/overlay/position/connected-position-strategy.ts
@@ -266,6 +266,15 @@ export class ConnectedPositionStrategy implements PositionStrategy {
   }
 
   /**
+   * Sets the origin element, relative to which to position the overlay.
+   * @param origin Reference to the new origin element.
+   */
+  setOrigin(origin: ElementRef): this {
+    this._origin = origin.nativeElement;
+    return this;
+  }
+
+  /**
    * Gets the horizontal (x) "start" dimension based on whether the overlay is in an RTL context.
    * @param rect
    */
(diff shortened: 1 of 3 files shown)


src/cdk/overlay/position/connected-position-strategy.ts
```diff
@@ -265,6 +265,15 @@ export class ConnectedPositionStrategy implements PositionStrategy {
     return this;
   }
 
+  /**
+   * Sets the origin element, relative to which to position the overlay.
+   * @param origin Reference to the new origin element.
+   */
+  withOrigin(origin: ElementRef): this {
```
jelbourn: `setOrigin` for this, since it's not something you'd specify when building the strategy?

crisbeto: Done

PR #9365, opened 2018-01-12
feat(button): create stroked and flat button variants

josephperrott:


src/lib/button/_button-base.scss
```diff
@@ -62,14 +62,30 @@ $mat-mini-fab-padding: 8px !default;
   }
 }
 
-// Applies styles to buttons with backgrounds: raised, fab, and mini-fab
-%mat-raised-button {
+// Applies styles to buttons with stroke styling
+%mat-stroked-button {
```
jelbourn: Why make stroked and flat sass placeholders? The raised button was a placeholder because the styles applied for both regular raised for fabs.

src/lib/button/_button-base.scss
```diff
@@ -62,14 +62,30 @@ $mat-mini-fab-padding: 8px !default;
...
   @extend %mat-button-base;
 
-  @include mat-overridable-elevation(2);
+  @include mat-overridable-elevation(0);
+
+  outline: 1px solid currentColor;
```
jelbourn: Can this be done with `border` instead of `outline`? Generally outline should be avoided since things its used by the browser for focus treatment, devtools for element selection, etc.

src/lib/button/_button-theme.scss
```diff
@@ -99,6 +99,23 @@
     @include _mat-button-ripple-color($theme, default, 0.1);
   }
 
+  .mat-stroked-button {
+    color: rgba(0, 0, 0, 0.87);
```
jelbourn: This should probably be pulled from the foreground palette (which accounts for light vs. dark theme)

src/lib/button/button.ts
```diff
@@ -56,6 +56,28 @@ export class MatButtonCssMatStyler {}
...
+
+/**
+ * Directive whose purpose is to add the mat- CSS styling to this selector.
+ * @docs-private
+ */
+@Directive({
+  selector: 'button[mat-flat-button], a[mat-flat-button]',
+  host: {'class': 'mat-flat-button'}
+})
+export class MatFlatButtonCssMatStyler {}
```
jelbourn: These separate styler directives aren't really necessary; now would be a good time to do away with them here. You can have `MatButton` just apply the right class checking the attributes (i.e. adding a new method like `_setButtonVariantClass`)

src/lib/button/_button-base.scss
josephperrott: Done, it was initially done this way to prevent the border from changing the offset height/width of the button.

src/lib/button/button.ts
```diff
@@ -29,50 +29,29 @@ import {
...
-})
-export class MatIconButtonCssMatStyler {}
+const BUTTON_HOST_ATTRIBUTE_MAP = {
+  '[class.mat-button]': `_hasHostAttributes('mat-button')`,
+  '[class.mat-flat-button]': `_hasHostAttributes('mat-flat-button')`,
+  '[class.mat-icon-button]': `_hasHostAttributes('mat-icon-button')`,
+  '[class.mat-raised-button]': `_hasHostAttributes('mat-raised-button')`,
+  '[class.mat-stroked-button]': `_hasHostAttributes('mat-stroked-button')`,
+  '[class.mat-mini-fab]': `_hasHostAttributes('mat-mini-fab')`,
+  '[class.mat-fab]': `_hasHostAttributes('mat-fab')`,
```
jelbourn: It's a bit overkill to add 7 bindings when we know that only one of these is going to apply (the bindings will generate more code and incur higher update costs). We only really need to check once at constructor time and add the class manually.

tinayuangao:
@josephperrott please rebase

Issue #7645, opened 2017-10-09
How to use mask in Date Picker


murilobd:
#### Bug, feature request, or proposal:
How can I set a text mask in date picker component so, if user doesn't want to select a date in calendar, he can type and fill correctly the field?

#### What is the expected behavior?
When click on input date picker, show the mask so user can type date instead of selecting a date from calendar

#### What is the current behavior?
empty input

#### What are the steps to reproduce?
Providing a StackBlitz/Plunker (or similar) is the *best* way to get the team to see your issue. <br/>
Plunker starter (using on `@master`): https://goo.gl/DlHd6U<br/>
StackBlitz starter (using latest `npm` release): https://goo.gl/wwnhMV<br/>
Just add a datepicker component

#### What is the use-case or motivation for changing an existing behavior?
Maybe user wants to type date instead of selecting form calendar.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
@angular/cli: 1.3.0
node: 6.10.3
os: darwin x64
@angular/animations: 4.3.5
@angular/cdk: 2.0.0-beta.8
@angular/common: 4.3.5
@angular/compiler: 4.3.5
@angular/core: 4.3.5
@angular/flex-layout: 2.0.0-beta.8
@angular/forms: 4.3.5
@angular/http: 4.3.5
@angular/material: 2.0.0-beta.8
@angular/platform-browser: 4.3.5
@angular/platform-browser-dynamic: 4.3.5
@angular/router: 4.3.5
@angular/cli: 1.3.0
@angular/compiler-cli: 4.3.5
@angular/language-service: 4.3.5

fxck:
a) material doesn't provide mask behavior
b) it's currently impossible to open datepicker with focus, while keeping focus on the input (because of how Overlay works)

murilobd:
Hey @fxck ! Thanks for the answer!
Well... there's no need to keep focus on text input wile calendar is open! If user click on input, just show the mask. If user clicks on button, hide mask and show calendar! Is that possible?

fxck:
Yes, but as I said, material doesn't provide that functionality, you can use something like https://github.com/text-mask/text-mask

renearias:
https://github.com/text-mask/text-mask dont work with ngmodel

remborg:
As mentioned above, text-mask won't work with material2. This is because agnular doesn't allow input fields to have more than 1 ValueAcessor.

A work around is to use the vanilla version of text-mask:
[...]

PR #9366, opened 2018-01-13
feat(form-field): support for different spec variants

mmalerba:
Add support for variants and create 3 to start with: `legacy` (same as current master), `standard` (same as current master + a couple tweaks to be more in line with the spec), `box` (the box style from the spec).

Demos:
https://mmalerba-demo1.firebaseapp.com - current master
https://mmalerba-demo2.firebaseapp.com - default variant = legacy (should be identical to master)
https://mmalerba-demo3.firebaseapp.com - default variant = standard
https://mmalerba-demo4.firebaseapp.com - default variant = box

Couple notes:
- `legacy` is the actual default I used, so nothing should change for anyone unless they explicitly set a variant.
- I didn't add the background ripple shown in the `box` spec yet.
- Non-floating labels really don't work well with the `box` variant because of how the label is aligned (see the "first name" input at https://mmalerba-demo4.firebaseapp.com/input). Should we allow it to look broken? Ignore the `floatLabel="never"` option?
- Prefixes & suffixes look funky in the `box` variant. The reason is that we're using prefix/suffix for 2 different purposes:
  1. Icons/buttons such as a password visibility toggle (e.g. https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B5ZSepuCX1xOdEc4cHZWV3ZzTzg/password1.png)
  2. Things like putting a '$' symbol in front of the user-inputted value. (e.g. https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B5ZSepuCX1xOdVBQcERBbTJpSm8/prefix1.png)

I think (ii) is actually a misuse of the prefix/suffix. It should behave more like a placeholder that only appears once the label floats up. The spec provides more examples of this kind of thing here: https://material.io/guidelines/components/text-fields.html#text-fields-input-types. I think we should probably add support for formats to `MatInput` to handle this case rather than relying on the form-field's prefix/suffix. If you guys agree with this I'll go ahead and make the prefix/suffix behave like the icon case in the spec for `standard` and `box` and leave the `legacy` behavior as is.

### DIFF
--- a/src/lib/form-field/_form-field-legacy-theme.scss
+++ b/src/lib/form-field/_form-field-legacy-theme.scss
@@ -1,0 +1,123 @@
+@import '../core/theming/palette';
+@import '../core/theming/theming';
+@import '../core/style/form-common';
+@import '../core/typography/typography-utils';
+
+
+// Theme styles that only apply to the legacy variant of the form-field.
+
+@mixin mat-form-field-legacy-theme($theme) {
+  $foreground: map-get($theme, foreground);
+  $is-dark-theme: map-get($theme, is-dark);
+
+  $underline-color: mat-color($foreground, divider, if($is-dark-theme, 0.7, 0.42));
+
+  .mat-form-field-variant-legacy {
+    .mat-form-field-underline {
+      background-color: $underline-color;
+    }
+
+    &.mat-form-field-disabled .mat-form-field-underline {
+      @include mat-control-disabled-underline($underline-color);
+    }
+  }
+}
+
+// Used to make instances of the _mat-form-field-label-floating mixin negligibly different,
+// and prevent Google's CSS Optimizer from collapsing the declarations. This is needed because some
+// of the selectors contain pseudo-classes not recognized in all browsers. If a browser encounters
+// an unknown pseudo-class it will discard the entire rule set.
+$mat-form-field-legacy-dedupe: 0;
+
+// Applies a floating label above the form field control itself.
+@mixin _mat-form-field-label-floating($font-scale, $infix-padding, $infix-margin-top) {
+  // We use perspective to fix the text blurriness as described here:
+  // http://www.useragentman.com/blog/2014/05/04/fixing-typography-inside-of-2-d-css-transforms/
+  // This results in a small jitter after the label floats on Firefox, which the
+  // translateZ fixes.
+  transform: translateY(-$infix-margin-top - $infix-padding) scale($font-scale) perspective(100px)
+  translateZ(0.001px + $mat-form-field-legacy-dedupe);
+  // The tricks above used to smooth out the animation on chrome and firefox actually make things
+  // worse on IE, so we don't include them in the IE version.
+  -ms-transform: translateY(-$infix-margin-top - $infix-padding + $mat-form-field-legacy-dedupe)
+                 scale($font-scale);
+
+  width: 100% / $font-scale + $mat-form-field-legacy-dedupe;
+
+  $mat-form-field-legacy-dedupe: $mat-form-field-legacy-dedupe + 0.00001 !global;
+}
+
+@mixin mat-form-field-legacy-typography($config) {
+  // The unit-less line-height from the font config.
+  $line-height: mat-line-height($config, input);
+  // The amount to scale the font for the floating label and subscript.
+  $subscript-font-scale: 0.75;
+  // The amount of space between the top of the line and the top of the actual text
+  // (as a fraction of the font-size).
+  $line-spacing: ($line-height - 1) / 2;
+  // The padding on the infix. Mocks show half of the text size, but seem to measure from the edge
+  // of the text itself, not the edge of the line; therefore we subtract off the line spacing.
+  $infix-padding: 0.5em - $line-spacing;
+  // The margin applied to the form-field-infix to reserve space for the floating label.
+  $infix-margin-top: 1em * $line-height * $subscript-font-scale;
+  // The space between the bottom of the .mat-form-field-flex area and the subscript wrapper.
+  // Mocks show half of the text size, but this margin is applied to an element with the subscript
+  // text font size, so we need to divide by the scale factor to make it half of the original text
+  // size. We again need to subtract off the line spacing since the mocks measure to the edge of the
+  // text, not the  edge of the line.
+  $subscript-margin-top: 0.5em / $subscript-font-scale - ($line-spacing * 2);
+  // The padding applied to the form-field-wrapper to reserve space for the subscript, since it's
+  // absolutely positioned. This is a combination of the subscript's margin and line-height, but we
+  // need to multiply by the subscript font scale factor since the wrapper has a larger font size.
+  $wrapper-padding-bottom: ($subscript-margin-top + $line-height) * $subscript-font-scale;
+
+  .mat-form-field-variant-legacy {
+    .mat-form-field-wrapper {
+      padding-bottom: $wrapper-padding-bottom;
+    }
+
+    .mat-form-field-infix {
+      padding: $infix-padding 0;
+    }
+
+    &.mat-form-field-can-float {
+      &.mat-form-field-should-float .mat-form-field-label,
+      .mat-input-server:focus + .mat-form-field-label-wrapper .mat-form-field-label {
+        @include _mat-form-field-label-floating(
+                $subscript-font-scale, $infix-padding, $infix-margin-top);
+      }
+
+      .mat-form-field-autofill-control:-webkit-autofill + .mat-form-field-label-wrapper
+      .mat-form-field-label {
+        @include _mat-form-field-label-floating(
+                $subscript-font-scale, $infix-padding, $infix-margin-top);
+      }
+
+      // Server-side rendered matInput with a label attribute but label not shown
+      // (used as a pure CSS stand-in for mat-form-field-should-float).
+      .mat-input-server[label]:not(:label-shown) + .mat-form-field-label-wrapper
+      .mat-form-field-label {
+        @include _mat-form-field-label-floating(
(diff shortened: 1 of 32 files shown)


src/lib/form-field/form-field-box.scss
```diff
@@ -0,0 +1,69 @@
+@import '../core/style/variables';
+@import '../core/style/vendor-prefixes';
+
+
+// Styles that only apply to the standard variant of the form-field.
```
crisbeto: Inaccurate comment.

src/lib/form-field/form-field-box.scss
```diff
@@ -0,0 +1,69 @@
...
+.mat-form-field-variant-box {
+  .mat-form-field-flex {
+    border-radius: $mat-form-field-box-border-radius;
+    padding: $mat-form-field-box-line-spacing $mat-form-field-box-side-padding 0
+             $mat-form-field-box-side-padding;
+  }
+
+  .mat-form-field-label {
+    // The perspective helps smooth out animations on Chrome and Firefox but isn't needed on IE.
+    transform: translateY(-50%) perspective(100px);
```
crisbeto: The `100px` seems arbitrary. Add a comment about where it comes from?

src/lib/form-field/form-field-box.scss
```diff
@@ -0,0 +1,69 @@
...
+    -ms-transform: translateY(-50%);
```
crisbeto: Does it get in the way of IE? If the outcome is the same we might as well keep the single declaration.

src/lib/form-field/form-field.ts
```diff
@@ -86,6 +95,9 @@ let nextUniqueId = 0;
 export class MatFormField implements AfterViewInit, AfterContentInit, AfterContentChecked {
   private _labelOptions: LabelOptions;
 
+  /** The form-field style variant. */
+  @Input() variant: 'legacy' | 'standard' | 'box' = 'legacy';
```
crisbeto: Turn this into a type so users can import it?

src/lib/form-field/form-field.ts
```diff
@@ -86,6 +95,9 @@ let nextUniqueId = 0;
...
+
```
crisbeto: Sidenote: maybe we should add an injection token that allows you to specify the default variant?

src/lib/form-field/form-field-box.scss
mmalerba: This was just trial and error to balance jumpiness and blurriness issues across different browsers. I'm going to try switching to `filter` and `backface-visibility` to fix some of these issues. It makes the blurriness better and improves the jumpiness in chrome, but makes firefox a little more jumpy. Overall I think its a little bit better solution

mmalerba:
I took a look at the firefox ellipsis issue, it looks like its caused by the combination of the `transform` `text-overflow` and `filter` properties. I'm not really sure what I can do about it...

src/lib/form-field/form-field-legacy.scss
```diff
@@ -0,0 +1,42 @@
...
+$mat-form-field-legacy-underline-height: 1px !default;
+
+
+.mat-form-field-variant-legacy {
+  .mat-form-field-label-wrapper {
+    filter: initial;
+  }
+
+  .mat-form-field-label {
+    backface-visibility: initial;
```
crisbeto: IE doesn't support the `initial` value, we should have a lint rule for it as well. You can default it to `visible` instead.

jelbourn:
I also see the blurriness in the box style

src/lib/form-field/form-field.scss
```diff
@@ -108,6 +108,9 @@ $mat-form-field-default-infix-width: 180px !default;
   // Hide the label initially, and only show it when it's floating or the control is empty.
   display: none;
 
+  // Helps prevent text jumpiness when transforming the label.
```
jelbourn: Can you provide any background on _why_ this is so?

src/lib/form-field/form-field.scss
```diff
@@ -77,6 +78,9 @@ $mat-form-field-default-infix-width: 180px !default;
   height: 100%;
   overflow: hidden;
   pointer-events: none;  // We shouldn't catch mouse events (let them through).
+
+  // Helps prevent text jumpiness when transforming the label.
```
jelbourn: Can you provide any background on _why_ this is so?

src/lib/form-field/form-field.scss
```diff
@@ -155,49 +158,38 @@ $mat-form-field-default-infix-width: 180px !default;
...
-      transform: scaleX(1);
-      transition: transform 300ms $swift-ease-out-timing-function,
-        opacity 100ms $swift-ease-out-timing-function,
-        background-color 300ms $swift-ease-out-timing-function;
-    }
+    opacity: 1;
+    transform: scaleX(1);
+    transition: transform 300ms $swift-ease-out-timing-function,
+    opacity 100ms $swift-ease-out-timing-function,
+    background-color 300ms $swift-ease-out-timing-function;
```
jelbourn: Indent overflow lines?

src/lib/select/select.scss
```diff
@@ -100,5 +105,7 @@ $mat-select-placeholder-arrow-space: 2 * ($mat-select-arrow-size + $mat-select-a
     // Remove the transition to prevent the placeholder
     // from overlapping when the label comes back down.
     transition: none;
+    // Prevents the '...' from showing on the parent element.
+    display: block;
```
jelbourn: Wouldn't the parent element just not have `overflow: ellipses`?

src/lib/select/select.scss
```diff
@@ -45,6 +45,11 @@ $mat-select-placeholder-arrow-space: 2 * ($mat-select-arrow-size + $mat-select-a
 .mat-select-arrow-wrapper {
   display: table-cell;
   vertical-align: middle;
+
+  // When used in a box variant form-field the arrow should be centered in the box.
+  .mat-form-field-variant-box & {
+    transform: translateY(-50%);
```
jelbourn: Transforming by a percentage value like this can lead to blurriness due to subpixel rendering. Generally it's better to center via flex styles

src/lib/form-field/form-field.ts
```diff
@@ -86,6 +98,9 @@ let nextUniqueId = 0;
 export class MatFormField implements AfterViewInit, AfterContentInit, AfterContentChecked {
   private _labelOptions: LabelOptions;
 
+  /** The form-field style variant. */
+  @Input() variant: MatFormFieldVariant = 'legacy';
```
jelbourn: Not necessarily in this PR, but we'll also need a provider to set the default appearance as well

src/lib/form-field/_form-field-legacy-theme.scss
```diff
@@ -0,0 +1,123 @@
...
+
+// Theme styles that only apply to the legacy variant of the form-field.
+
+@mixin mat-form-field-legacy-theme($theme) {
+  $foreground: map-get($theme, foreground);
+  $is-dark-theme: map-get($theme, is-dark);
+
+  $underline-color: mat-color($foreground, divider, if($is-dark-theme, 0.7, 0.42));
+
+  .mat-form-field-variant-legacy {
```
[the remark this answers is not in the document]
mmalerba: Yeah if this isn't feasible we can add all the legacy styles to the base styles and then override the ones we don't want in standard, box, and outline

src/lib/form-field/form-field.scss
mmalerba: Changed some things so this is no longer needed

src/lib/form-field/form-field.scss
mmalerba: Changed some things so this is no longer needed

src/lib/form-field/form-field.ts
mmalerba: Yep, will add this later

src/lib/select/select.scss
mmalerba: The `overflow: ellipses` on the parent is also what gives the value its `...` so we need to do it this way even though its a little strange

mmalerba:
I've updated all of the demos with the latest changes (you may need to clear your cache). I also sent an email to the material design team hoping that they can clarify the prefix/suffix thing for us.

mmalerba:
I've updated the `standard` and `box` appearances so that they no longer promote the placeholder to a label and they ignore `floatLabel="never"` instead treating it like `floatLabel="always"`. The `legacy` appearance still behaves like it does in current master.

Remaining issues I plan to address in future PRs:
- How to treat prefix and suffix in the new appearances
- Allowing user to set default appearance via a provider
- Adding a background ripple for the box appearance

mmalerba:
resolved the bazel failure, PTAL

talamaska:
I don't get it. Why so drastic change in the behavior of the non floating label? I liked that feature and I'm using it in production. Can't we not place that mat-label and have a placeholder like thing, but with the required asterisk, you can put a colored asterisk in the placeholder attribute normally. With that drastic change I'll be forced to stick to 5.2.1. So is it possible to make 1 more component mat-placeholder to act like a placeholder, but have a required mark and be possible to not use mat-label for the purpose of a placeholder

mmalerba:
@talamaska I've upped the priority on #7645 for adding text mask support to the input. I'm not 100% sure our text masking solution will be able to cover this use case, but I'll keep it in mind while designing.

The never-float placeholder stuff was really just an artifact of the fact that the spec didn't used to distinguish between label and placeholder and we therefore needed some way to emulate standard placeholder behavior with the floating label. The spec is now much more clear about the fact that the label and placeholder are two different things and the required `*` really belongs with the label

talamaska:
Having a required mark in the placeholder doen't have anything to do with the text masking. It is a totaly separated feature.

PR #9373, opened 2018-01-13
feat(slider): add focus and blur methods do MatSlider class

### LINKED ISSUE #9332: Feat (Mat-Slider) : Focus method and blur Output
Feature request : 
Implement Focus and Blur on mat-slider

- Allow mat-slider with thumbLabel to activate thumbLabel programatically on focus via MatSlider class
- Allow mat-slider to notify of blur via MatSlider class

Need to use an ElementRef instead of a reference to MatSlider

Uniformisation of component's api

I love ViewChildren with specific class component matching!

julianobrasil:
Does it need any tests?

<strike>Also there's a little side effect: if the thumb label is already visible and you try to show it programmatically by, let's say, clicking on a button, as the focus will be lost to the clicked button, the thumb label will disappear and appear again.</strike> Just realized that this is kind of normal (but I think it'd be better without this blinking)

image

- Closes #9332

### DIFF
--- a/src/demo-app/slider/slider-demo.html
+++ b/src/demo-app/slider/slider-demo.html
@@ -53,6 +53,11 @@ Label <mat-slider #slidey aria-label="Basic slider"></mat-slider>
            aria-label="Inverted vertical slider">
 </mat-slider>
 
+<h1>Set/lost focus to show thumblabel programmatically</h1>
+<mat-slider #demoSlider="matSlider" thumbLabel aria-label="Slider with thumb label"></mat-slider>
+<button (click)="demoSlider.focus()">Focus Slider</button>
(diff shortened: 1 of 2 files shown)


mmalerba:
The blinking is because you're blurring the slider when you click on the button and then programmatically re-focusing it. One thing that could potentially be done is to add a slight delay after blurring before hiding the thumb (let's not do it in this PR though)

jelbourn:
@julianobrasil DM me your email address on [Twitter](https://twitter.com/jelbourn)?

julianobrasil:
@jelbourn, I wasn't able to DM you (twitter policies), but my email address is public in github: https://github.com/julianobrasil

julianobrasil:
juliano.brasil@gmail.com

PR #9379, opened 2018-01-13
feat(dialog): support using dialog content directives with template dialogs

### LINKED ISSUE #5412: Opening an MdDialog with a TemplateRef that uses `md-dialog-close` throws "no provider for MdDialogRef"
#### Proposal:
Provide `MdDialogRef` even when we pass a `TemplateRef`.

#### What is the current behaviour?
With the current setup if we pass a TemplateRef to MdDialog, there is no way to use  `md-dialog-close` directive as it throws `no provider for MdDialogRef` as it requires it. 

#### What is the expected behaviour?
It would be great if it works with a `TemplateRef` but if that's not possible, then could you please add a brief warning to the docs?

I can provide a plunkr if needed.

crisbeto:
Previously the `matDialogClose`, `matDialogTitle` etc. directives would only work correctly inside component dialogs, because using DI to get the dialog ref doesn't work inside template dialogs. These changes add a fallback that finds the dialog ref based on the id of the closest dialog container.

Fixes #5412.

### DIFF
--- a/src/lib/dialog/dialog-content-directives.ts
+++ b/src/lib/dialog/dialog-content-directives.ts
@@ -6,9 +6,17 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Directive, Input, OnChanges, OnInit, Optional, SimpleChanges} from '@angular/core';
+import {
+  Directive,
+  Input,
+  OnChanges,
+  OnInit,
+  Optional,
+  SimpleChanges,
+  ElementRef,
+} from '@angular/core';
+import {MatDialog} from './dialog';
 import {MatDialogRef} from './dialog-ref';
-import {MatDialogContainer} from './dialog-container';
 
 /** Counter used to generate unique IDs for dialog elements. */
 let dialogElementUid = 0;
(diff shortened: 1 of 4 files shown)


src/lib/dialog/dialog-content-directives.ts
```diff
@@ -34,7 +42,19 @@ export class MatDialogClose implements OnChanges {
...
+  constructor(
+    @Optional() public dialogRef: MatDialogRef<any>,
+    private _elementRef: ElementRef,
+    private _dialog: MatDialog) {}
+
+  ngOnInit() {
+    if (!this.dialogRef) {
+      // If we couldn't get a dialog ref through DI, try to find one by looking at the DOM.
+      // This can happen if the close button is used inside a template dialog. Note that
+      // we have to do the lookup in OnInit, because the constructor might be too early.
```
jelbourn: Small comment suggestion:
```ts
// When this directive is included in a dialog via TemplateRef (rather than being
// in a Component), the DialogRef isn't available via injection because embedded
// views cannot be given a custom injector. Instead, we look up the DialogRef by
// ID. This must occur in `onInit`, as the ID binding for the dialog container won't
// be resolved at constructor time.
```

jelbourn:
@crisbeto many tests failing with `Cannot read property '_ariaLabelledBy' of undefined` at MatDialogTitle.ngOnInit

crisbeto:
Pushed a fix for the null pointer error.

PR #9381, opened 2018-01-13
feat(drawer): allow for backdrop to be disabled

### LINKED ISSUE #5300: [Sidenav] - add support for 'backdrop' as optional input
I would like to suggest to have 'backdrop' as an optional input for Sidenav compoment.
Default would be true for 'mode' = 'over' or 'push', and false for 'side'

if 'backdrop' input is provided, to overwrite the default; and show/hide backdrop accordingly

Backdrop exists for 'mode' = 'over' or 'push', and does not exist for 'side'

current documentation on Sidenav:
https://material.angular.io/components/sidenav/overview

Give user the flexibility to not have 'backdrop' on 'mode' = 'over' or 'push' or do have it on 'side'

"@angular/core": "4.2.0-beta.0",
"@angular/material": "^2.0.0-beta.3",
"typescript": "~2.3.2"

this change would be appreciated

crisbeto:
Adds the `hasBackdrop` input to the drawer container, allowing users to disable the backdrop.

Fixes #5300.

### DIFF
--- a/src/lib/sidenav/drawer.ts
+++ b/src/lib/sidenav/drawer.ts
@@ -456,6 +456,20 @@ export class MatDrawerContainer implements AfterContentInit, OnDestroy {
   set autosize(value: boolean) { this._autosize = coerceBooleanProperty(value); }
   private _autosize: boolean;
 
+  /** Whether the drawer container should have a backdrop while one of the sidenavs is open. */
+  @Input()
+  get hasBackdrop() {
+    if (this._hasBackdrop == null) {
+      return !this._start || this._start.mode !== 'side' || !this._end || this._end.mode !== 'side';
+    }
+
+    return this._hasBackdrop;
+  }
+  set hasBackdrop(value: any) {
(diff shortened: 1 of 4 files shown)


src/lib/sidenav/drawer.ts
```diff
@@ -446,6 +446,20 @@ export class MatDrawerContainer implements AfterContentInit, OnDestroy {
...
+  @Input()
+  get hasBackdrop() {
+    if (this._hasBackdrop == null) {
+      return !this._start || this._start.mode !== 'side' || !this._end || this._end.mode !== 'side';
+    }
+
+    return this._hasBackdrop;
+  }
+  set hasBackdrop(value: any) {
+    this._hasBackdrop = coerceBooleanProperty(value);
```
mmalerba: This means that the user can't set it back to it's initial state right (null). I think we should treat null/undefined as null

crisbeto: Good point. It's changed.

PR #9383, opened 2018-01-14
fix(menu): always focus first menu item

### LINKED ISSUE #9252: mat-menu is acting differently when triggered via keyboard vs mouse
Bug

User clicks on the mat-menu trigger button, and first item within menu (or the entire menu) is focused.

User clicks on the mat-menu trigger button, the button stays focused. This creates couple of accessibility issues. Note that if its triggered via keyboard this issue doesn't repro thanks to this logic: https://github.com/angular/material2/blob/3728555c6e37dbeb1756366c797c99655d8ffa1f/src/lib/menu/menu-trigger.ts#L240

1. Go to https://stackblitz.com/angular/lrpbkepagev?file=app%2Fmenu-overview-example.html
2. Click on the menu button
3. Note that focus stays on the button (print document.activeElement)

Note that if you do this via keyboard, the first focusable element within mat-menu is focused.

We have a tooltip component that stays open when a button is focused. When there is a tooltip and a menu attached to a component, click on the menu trigger button will result in both tooltip and menu showing. Once user clicks on menu, the menu component should take away to focus (like it does when its triggered via keyboard). This would also make sure screen readers announce the right content.

All

crisbeto:
* Switches to always focusing the first menu item, even when opening by mouse. The focus styling will only apply when focused by keyboard.
* Adds the ability to set the focus origin through the `FocusKeyManager`.

Fixes #9252.

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -40,6 +40,7 @@ import {MatMenuItem} from './menu-item';
 import {MatMenuPanel} from './menu-panel';
 import {MenuPositionX, MenuPositionY} from './menu-positions';
 import {coerceBooleanProperty} from '@angular/cdk/coercion';
+import {FocusOrigin} from '@angular/cdk/a11y';
 
 
 /** Default `mat-menu` options that can be overridden. */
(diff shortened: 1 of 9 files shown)


src/lib/menu/menu-directive.ts
```diff
@@ -228,16 +229,16 @@ export class MatMenu implements AfterContentInit, MatMenuPanel, OnDestroy {
...
 
   /**
-   * Focus the first item in the menu. This method is used by the menu trigger
-   * to focus the first item when the menu is opened by the ENTER key.
+   * Focus the first item in the menu.
+   * @param origin Action from which the focus originated. Used to set the correct styling.
    */
-  focusFirstItem() {
-    this._keyManager.setFirstItemActive();
+  focusFirstItem(origin: FocusOrigin) {
```
jelbourn: Should this be option to avoid a breaking change?

src/lib/menu/menu-item.ts
```diff
@@ -64,16 +64,31 @@ export class MatMenuItem extends _MatMenuItemMixinBase
...
   _triggersSubmenu: boolean = false;
 
-  constructor(private _elementRef: ElementRef) {
+  constructor(
+    private _elementRef: ElementRef,
+    // TODO(crisbeto): switch to a required param when doing breaking changes.
+    private _focusMonitor?: FocusMonitor) {
     super();
+
+    if (_focusMonitor) {
```
jelbourn: Add a comment here like
```ts
// We want to show the focus style for menu items only when the focus was not caused by
// a mouse or touch interaction. 
```
?

crisbeto:

jelbourn:
Actually it looks like CircleCI is raising a compilation error

crisbeto:
ops, it's sorted out now.

Issue #8724, opened 2017-11-29
[Tooltip] Bad error message when placed on non-element (ng-container)


andrewseguin:
When a `[matTooltip]` is placed on an `ng-container`, it is invalid but the resulting error message is unhelpful in finding out how to resolve the issue: `el.getAttribute is not a function`.

Should be able to check in the constructor if the `elementRef.nativeElement` exists and if not, a helpful error message can be provided.

jelbourn:
Just tried with w/ the latest version and this seems to no longer throw any error message

PR #9392, opened 2018-01-15
fix(aria-describer): exception when attempting to describe a non-element node

crisbeto:
* Fixes an error that was being thrown when trying to describe a non-element node. This could happen if the consumer is attempting to describe an `ng-container` which corresponds to a comment node.
* Switches a few cases, that had hard-coded the `Node.ELEMENT_NODE` constant for to Universal compatibility, to take the constant from the `Document`.

Relates to #8724.


josephperrott:
needs rebase

crisbeto:
Rebased.

PR #9412, opened 2018-01-16
fix(menu): match base elevation in spec

### LINKED ISSUE #9409: Menu: Elevation is 2dp instead of 8dp
Bug

Elevation of the Menu component should be 8dp as per the [official material design documentation for elevation and shadows](https://material.io/guidelines/material-design/elevation-shadows.html#elevation-shadows-shadows). Sub menus have elevation of 9dp (+1 dp for each sub menu)

>**Menus and sub menus**
>Menus: 8dp
>Sub menus: 9dp (+1 dp for each sub menu)

Menu has 2dp elevation and sub menus have 3+dp.

1. Open project documentation
1. Go to Component -> Menu
1. Open first menu example
1. Inspect element and check classes on .mat-menu-panel

If this library is following Google Material Design it should reflect it correctly in every aspect **or at least provide option to change elevation**.

* Angular 5.1.0
* Material 5.0.2

CDDelta:
Fixes #9409

### DIFF
--- a/src/lib/menu/menu-directive.ts
+++ b/src/lib/menu/menu-directive.ts
@@ -62,7 +62,7 @@ export const MAT_MENU_DEFAULT_OPTIONS =
  * Start elevation for the menu panel.
  * @docs-private
  */
-const MAT_MENU_BASE_ELEVATION = 2;
+const MAT_MENU_BASE_ELEVATION = 8;
 
 
 @Component({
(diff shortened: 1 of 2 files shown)


CDDelta:
According to https://material.io/guidelines/material-design/elevation-shadows.html, the elevation for menus should be 8.

crisbeto:
While that's true, the designs that back it up and I based it off of MDC which also use the lower elevation. See https://material-components-web.appspot.com/simple-menu.html

CDDelta:
Ok then, sounds good.

PR #9418, opened 2018-01-16
build: fix chrome headless not starting

DevVersion:
* Applies a workaround for https://github.com/travis-ci/travis-ci/issues/8836 that fixes the Chrome launch issues. (similar workaround applied on angular/angular)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 language: node_js
-sudo: false
 dist: trusty
 
+# Temporary set sudo to required, because in normal container-based Travis jobs, the Chrome
+# binaries seem to be owned by root, and therefore can't be accessed by Karma. To workaround
+# the issue we temporary grant the CI sudo permissions.
+# https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-356362524.
+sudo: required
+
 node_js:
-  - '8'
+  - '8.9.4'
 
 addons:
   jwt:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,14 @@
...
 
+# Temporary set sudo to required, because in normal container-based Travis jobs, the Chrome
+# binaries seem to be owned by root, and therefore can't be accessed by Karma. To workaround
+# the issue we temporary grant the CI sudo permissions.
+# https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-356362524.
+sudo: required
+
 node_js:
-  - '8'
+  - '8.9.4'
```
jelbourn: Add a comment that explains why this specific version of node is used?

DevVersion: Not sure this makes sense. It's the latest LTS version of NodeJS, and this comment would get outdated from time to time.

jelbourn: Then why not leave it as `8`? My understanding is that Travis should use the latest version of 8.x.
https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions

DevVersion: That's true. I just saw that angular/angular is also more explicit with it, and that way I thought we can be more carefully about auto-upgrades of the node version.

jelbourn: They might do it to make sure every build is consistent, which would be a good reason. Still worth commenting, though.

DevVersion: Yep. I will add a comment that explains that we won't to be explicit about this version as well for job consistency.

Issue #9421, opened 2018-01-16
mat-select should provide itself as a ControlValueAccessor


Juraji:
#### Bug, feature request, or proposal:
Using a custom override of the FormControlDirective causes using [formControl] on mat-select to throw the following error:

```
ERROR Error: No value accessor for form control with unspecified name attribute
    at _throwError (forms.js?ec33:2415)
    at setUpControl (forms.js?ec33:2285)
    at FormControlDirective.ngOnChanges (forms.js?ec33:6448)
    ...
```
(The field will still work)

#### What is the expected behavior?
Being able to use a custom FormControlDirective implementation.

#### What is the current behavior?
An error is thrown by NgControl not being able to infer a ControlValueAccessor

#### What are the steps to reproduce?
Override Angular's core FormControlDirective and use it on a mat-select
See the following plunk: http://plnkr.co/edit/x1ZO9mMv8zYZh9uescPo

#### What is the use-case or motivation for changing an existing behavior?
Using mat-select with a custom FormControlDirective.
(for instance mat-slide-toggle does this correctly)

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
At least the following:
Angular: 5.1.1
Material: 5.0.2
TypeScript: 2.6.2
Browsers: Chrome, Opera and Firefox
OS: Windows, Mac, Linux

crisbeto:
This has come up in the past. We can't do it for the select, because it needs to inject the `NgControl` which refers to itself and causes a circular import.

Juraji:
@crisbeto as I noticed a bit too late... Sorry for wasting your time.

JBoothUA:
is there a work around for this??  or does mat-select just not work with forms?

PR #9422, opened 2018-01-16
fix(mat-select): Add provider for MatSelect as ControlValueAccessor

Juraji:
Applies a fix for https://github.com/angular/material2/issues/9421, where MatSelect does implement the interface ControlValueAccessor but does not provide itself as a ControlValueAccessor.


Juraji:
I signed it!

crisbeto:
Closing due to circular imports. See https://github.com/angular/material2/issues/9421#issuecomment-359179248.

PR #9423, opened 2018-01-16
fix(menu,toolbar): avoid potential server-side rendering errors

crisbeto:
Avoids a couple of potential server-side rendering errors due to the menu item and toolbar components referring to the `Node` global variable directly.

### DIFF
--- a/src/lib/menu/menu-item.ts
+++ b/src/lib/menu/menu-item.ts
@@ -13,6 +13,7 @@ import {
   ElementRef,
   OnDestroy,
   ViewEncapsulation,
+  Inject,
 } from '@angular/core';
 import {
   CanDisable,
@@ -21,6 +22,7 @@ import {
   mixinDisableRipple
 } from '@angular/material/core';
 import {Subject} from 'rxjs/Subject';
+import {DOCUMENT} from '@angular/common';
 
 // Boilerplate for applying mixins to MatMenuItem.
 /** @docs-private */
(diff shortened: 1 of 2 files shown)


src/lib/menu/menu-item.ts
```diff
@@ -105,6 +112,7 @@ export class MatMenuItem extends _MatMenuItemMixinBase
   /** Gets the label to be used when determining whether the option should be focused. */
   getLabel(): string {
     const element: HTMLElement = this._elementRef.nativeElement;
+    const textNodeType = this._document ? this._document.TEXT_NODE : 3;
```
jelbourn: Make constants for node types until `_document` is always available?

crisbeto: I'm inclined not to, because we'll end up sprinkling them everywhere and it'll make it more likely to forget about removing them.

jelbourn:
@crisbeto passes presubmit, just needs rebease

crisbeto:
Rebased.

src/lib/toolbar/toolbar.ts
```diff
@@ -51,12 +53,19 @@ export class MatToolbarRow {}
...
+  private _document: Document;
 
   /** Reference to all toolbar row elements that have been projected. */
   @ContentChildren(MatToolbarRow) _toolbarRows: QueryList<MatToolbarRow>;
 
-  constructor(elementRef: ElementRef, private _platform: Platform) {
+  constructor(
+    elementRef: ElementRef,
+    private _platform: Platform,
+    @Inject(DOCUMENT) document?: any) {
```
rafaelss95: @crisbeto I know this PR has already been merged, but shouldn't this have `@deletion-target x.y.z` as in others `constructor` parameters?

crisbeto: I don't think we had the tooling around `@deletion-target` when I first did this PR.

rafaelss95: Aha, ok. Do you pretend to submit a PR to add this?

PR #9434, opened 2018-01-17
fix(slider): slider emiting changes on slide end when disabled

llorenspujol:
Currently, the slider is firing a change event on slideEnd when it is disabled. If it is disabled and the value of the slider doesn't change, it shouldn't have to emit any change event. 

I have added the test and its corresponding fix for it to pass.

### DIFF
--- a/src/lib/slider/slider.ts
+++ b/src/lib/slider/slider.ts
@@ -518,7 +518,7 @@ export class MatSlider extends _MatSliderMixinBase
   _onSlideEnd() {
     this._isSliding = false;
 
-    if (this._valueOnSlideStart != this.value) {
+    if (this._valueOnSlideStart != this.value && !this.disabled) {
       this._emitChangeEvent();
     }
     this._valueOnSlideStart = null;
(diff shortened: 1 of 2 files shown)


src/lib/slider/slider.ts
```diff
@@ -518,7 +518,7 @@ export class MatSlider extends _MatSliderMixinBase
   _onSlideEnd() {
     this._isSliding = false;
 
-    if (this._valueOnSlideStart != this.value) {
+    if (this._valueOnSlideStart != null && this._valueOnSlideStart != this.value) {
```
mmalerba: I don't think this is quite right, how about: `this._valueOnSlideStart != this.value && !this.disabled`

llorenspujol: Yes, maybe is better, I am not in context at all. I will make the change (Y)

llorenspujol: I have made the change @mmalerba

andrewseguin:
Got a minor lint error:

```ERROR: /home/travis/build/angular/material2/src/lib/slider/slider.spec.ts[257, 1]: trailing whitespace```

jelbourn:
This passes google presubmit

@llorenspujol we can merge this if you correct the lint error

llorenspujol:
@jelbourn @andrewseguin lint fixed.

PR #9443, opened 2018-01-17
docs(list): expand accessibility section

jelbourn:


src/lib/list/list.md
```diff
@@ -152,8 +152,24 @@ To add a divider, use `<mat-divider>`.
 ```
 
 ### Accessibility
+The type of list used in any given situation depends on how the end-user will be interacting with
+the list.
```
crisbeto: "...interacting with it", you already mentioned "list" earlier.

Issue #7191, opened 2017-09-19
Component *change events consistency


rafaelss95:
#### Bug, feature request, or proposal:
Proposal.

#### What is the expected behavior?
I'd expect all components to have the same behavior. Or all return a plain object, or all return an instance of a `class`.

#### What is the current behavior?
Some components return a plain object, others return an instance of a `class`.

#### Is there anything else we should know?
Maybe related to #6677.

amcdnl:
This is being addressed by https://github.com/angular/material2/issues/6677

rafaelss95:
Hey @amcdnl, as I said it's a complement of #6677.

There are inconsistencies here that's not pointed out in that issue.

And, according to your [last comment there](https://github.com/angular/material2/issues/6677#issuecomment-330402105) you've already done all the work for that issue.

rafaelss95:
@amcdnl can you take a look at this issue again? Thanks.

vivian-hu:
Team discussed the options of
1. Omitting the word "Event" from the names and instead suffixing with the type of event (e.g. "Change").
2. Always put "Event" at the end
The decision is to go with option 2: always put "Event" at the end.  It will be clearer to have "MatDatepickerInputEvent" than "MatDatepickerInput"

rafaelss95:
@vivian-hu Hmm.. any chance to do this before v8?

jelbourn:
This won't make it in for v8

rafaelss95:
@jelbourn Any news about this? I would love to help with this issue once you tell me the team decision about the convention.

jelbourn:
Vivian had been the one owning this project, but she left the team a long while ago, and it basically fell to the wayside when that happened. With everything else we've got going on now, I don't think this will be on our radar in 2020.

crisbeto:
We are closing this issue, because we've determined, using a semi-automated process, that it is no longer relevant, has low engagement, or it does not have enough information to proceed. Feel free to comment if you think the issue should be reopened, or open a new issue with the latest version of our code instead.

PR #9445, opened 2018-01-17
refactor: switch output event event classes to interfaces

crisbeto:
Converts the various event classes that we emit through `@Output` into interfaces in order to generate less compiled code and potentially less objects to garbage collect.


crisbeto:
Caretaker note: we should keep an eye out on how many presubmit failures this causes and determine based on that whether it's still worth the trouble.

rafaelss95:
FYI this fixes part of #7191.

jelbourn:
@crisbeto I ran a presubmit on this and found that there are quite a few places where people `new` their own events in unit tests. Based on that, this would have to be a major change

rafaelss95:
Any chance to ship this in v6?

jelbourn:
This isn't going to make it in for 6.0 since it caused too many breakages inside Google. We should bump it to 7.0

crisbeto:
Closing since this is way out of date.

PR #9446, opened 2018-01-17
fix(overlay): transparent overlay not blocking scroll on Firefox 57

### LINKED ISSUE #8924: [Select] Scrolling causes options to scroll away
image

It should scroll but should inside the content.

Running out of the content, PSA.

0. go to: https://material.angular.io/components/select/examples
1. Just click on the arrow to dropped down the Options.
2. Use mouse and scroll down the page and see the Options running out of the content.

crisbeto:
As of Firefox 57, it looks like elements with a transparent background won't capture scroll events. These changes work around the issue by not setting the background color, but using `opacity` to hide the element while keeping it scrollable.

Fixes #8924.

### DIFF
--- a/src/cdk/overlay/_overlay.scss
+++ b/src/cdk/overlay/_overlay.scss
@@ -76,7 +76,13 @@ $backdrop-animation-timing-function: cubic-bezier(0.25, 0.8, 0.25, 1) !default;
   }
 
   .cdk-overlay-transparent-backdrop {
-    background: none;
+    // Note: as of Firefox 57, having the backdrop be `background: none` will prevent it from
+    // capturing the user's mouse scroll events. Since we also can't use something like
+    // `rgba(0, 0, 0, 0)`, we work around the inconsistency by not setting the background at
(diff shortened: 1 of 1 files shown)


src/cdk/overlay/_overlay.scss
```diff
@@ -76,7 +76,13 @@ $backdrop-animation-timing-function: cubic-bezier(0.25, 0.8, 0.25, 1) !default;
...
 
   .cdk-overlay-transparent-backdrop {
-    background: none;
+    // Note: as of Firefox 57, having the backdrop be `background: none` will prevent it from
+    // capturing the user's mouse scroll events. Since we also can't use something like
+    // `rgba(0, 0, 0, 0)`, we work around the inconsistency by not setting the background at
+    // all and using `opacity` to make the element transparent.
+    &, &.cdk-overlay-backdrop-showing {
+      opacity: 0;
+    }
```
jelbourn: Would `background: transparent` work?

crisbeto: It doesn't.

PR #9448, opened 2018-01-17
fix(overlay): server-side rendering error when creating backdrop element

crisbeto:
Avoids a couple of server-side rendering errors when the overlay's backdrop is being created.


jelbourn:
@crisbeto passes Google presubmit, but Travis looks to have some failures related to this PR

crisbeto:
Rebased and sorted out the test failure.

PR #9466, opened 2018-01-18
fix(overlay): validate that ConnectedPositionStrategy positions are passed in correctly at runtime

crisbeto:
Since we have some cases where we pass in positions verbatim to the connected position strategy (the connected position directive), we could end up in a situation where the consumer passed in a set of invalid positions that don't break necessarily, but don't work correctly either. These changes add some sanity checks at runtime to make debugging easier.

### DIFF
--- a/src/cdk/overlay/position/connected-position-strategy.ts
+++ b/src/cdk/overlay/position/connected-position-strategy.ts
@@ -15,6 +15,8 @@ import {
   OverlayConnectionPosition,
   ConnectedOverlayPositionChange,
   ScrollingVisibility,
+  validateHorizontalPosition,
+  validateVerticalPosition,
 } from './connected-position';
 import {Subject} from 'rxjs/Subject';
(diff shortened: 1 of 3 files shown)


src/cdk/overlay/position/connected-position-strategy.ts
```diff
@@ -428,14 +433,26 @@ export class ConnectedPositionStrategy implements PositionStrategy {
...
     }, length);
   }
+
+  /** Validates that the current position match the expected values. */
+  private _validatePositions(): void {
+    if (!this._preferredPositions.length) {
+      throw Error('ConnectedPositionStrategy: At least one position is required.');
+    }
+
+    this._preferredPositions.forEach(pair => {
```
jelbourn: Add a TODO to remove this check once Angular's template type checker is good enough to not need it?

crisbeto:

PR #9471, opened 2018-01-19
docs: ensure docs are being shown for '@Input's

### LINKED ISSUE #9452: Missing @Input in the docs for MatTabGroup's API
Expect the `MatTabGroup` api to have `@Input()` before `selectedIndex` 

There is no `@Input()`

It has it as,

    selectedIndex: number | null

where as it should be

    @Input()
    selectedIndex: number | null

See [MatTabGroup api table](https://material.angular.io/components/tabs/api#MatTabGroup)

The documents should explicitly state that `selectedIndex` is an input

Material 5.1.0

rafaelss95:
This PR basically makes sure that _getters_ appear before _setters_. Why?

Apparently docs aren't being shown up if *setters* appear before *getters*, e.g.:
https://github.com/angular/material2/blob/7af132d51c0076033f2c32a80ccc01f95f2b8d51/src/lib/tabs/tab-group.ts#L112-L117

For reference: #9452.

**PS:**

Not sure about this, but I think it'd be great if we could create a custom rule to enforce the (apparently) standard convention:

- `getter`;
- `setter`;
- `_property`;

### DIFF
--- a/e2e/components/menu-e2e.spec.ts
+++ b/e2e/components/menu-e2e.spec.ts
@@ -168,7 +168,7 @@ describe('menu', () => {
       const trigger = await page.beforeTrigger().getLocation();
 
       // the menu's right corner must be attached to the trigger's right corner.
-      // menu = 112px wide. trigger = 60px wide.  112 - 60 =  52px of menu to the left of trigger.
+      // menu = 112px wide. trigger = 60px wide. 112 - 60 = 52px of menu to the left of trigger.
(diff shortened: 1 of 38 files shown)


jelbourn:
Thanks!

DevVersion:
This is generally a good change for consistency, but we should also fix the underlying issue on our Dgeni setup. So we should not close #9452 with this PR, can you update your commit message?

rafaelss95:
Done. @DevVersion.

PR #9473, opened 2018-01-19
refactor(progress-bar): use color mixin

rafaelss95:
The color attribute for the `progress-bar` now uses the color mixin as every other component that is themable.

### DIFF
--- a/src/lib/progress-bar/progress-bar.ts
+++ b/src/lib/progress-bar/progress-bar.ts
@@ -6,10 +6,25 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {Component, ChangeDetectionStrategy, Input, ViewEncapsulation} from '@angular/core';
+import {
+  Component,
+  ChangeDetectionStrategy,
+  ElementRef,
+  Input,
+  ViewEncapsulation
+} from '@angular/core';
+import {CanColor, mixinColor} from '@angular/material/core';
 
 // TODO(josephperrott): Benchpress tests.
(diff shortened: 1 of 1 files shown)


src/lib/progress-bar/progress-bar.ts
```diff
@@ -6,10 +6,25 @@
...
 
-import {Component, ChangeDetectionStrategy, Input, ViewEncapsulation} from '@angular/core';
+import {
+  Component,
+  ChangeDetectionStrategy,
+  ElementRef,
+  Input,
+  ViewEncapsulation
+} from '@angular/core';
+import {CanColor, mixinColor, ThemePalette} from '@angular/material/core';
```
DevVersion: `ThemePalette` seems to be unused

src/lib/progress-bar/progress-bar.ts
```diff
@@ -36,23 +48,23 @@ import {Component, ChangeDetectionStrategy, Input, ViewEncapsulation} from '@ang
   encapsulation: ViewEncapsulation.None,
```
crisbeto: You also have to add `inputs: ['color']` to the metadata.

rafaelss95:
Done.

@crisbeto @DevVersion.

jelbourn:
@josephperrott why did you mark this as major?

josephperrott:
@jelbourn The types are changed for the `@Input() color` when moving to `ThemePalette`

jelbourn:
Changing to patch since it loosens the type rather than restricting it. Passes for all Google apps

PR #9487, opened 2018-01-19
feat(datepicker): allow for mat-datepicker-toggle icon to be customized

crisbeto:
Allows for the consumer to set a different icon for the `mat-datepicker-icon` via the `matDatepickerToggleIcon` directive. The reasoning for the change is that not all consumers might be using the Material icons or they may want to set something different (e.g. a chevron).

### DIFF
--- a/src/lib/datepicker/datepicker-toggle.ts
+++ b/src/lib/datepicker/datepicker-toggle.ts
@@ -17,12 +17,21 @@ import {
   OnDestroy,
   SimpleChanges,
   ViewEncapsulation,
+  Directive,
+  ContentChild,
 } from '@angular/core';
 import {merge} from 'rxjs/observable/merge';
 import {of as observableOf} from 'rxjs/observable/of';
 import {Subscription} from 'rxjs/Subscription';
 import {MatDatepicker} from './datepicker';
 import {MatDatepickerIntl} from './datepicker-intl';
+
+
+/** Can be used to override the icon of a `matDatepickerToggle`. */
+@Directive({
+  selector: '[matDatepickerToggleIcon]'
+})
+export class MatDatepickerToggleIcon {}
 
 
 @Component({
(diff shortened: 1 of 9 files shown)


src/lib/datepicker/datepicker.md
```diff
@@ -32,6 +32,17 @@ can easily be used as a prefix or suffix on the material input:
...
+If you want to customize the icon that is rendered inside the `mat-datepicker-toggle`, you can do so
+by using the `matDatepickerToggleIcon` directive:
+
+```html
+<input [matDatepicker]="myDatepicker">
+<mat-datepicker-toggle [for]="myDatepicker">
+  <img matDatepickerToggleIcon src="my-custom-icon.png">
+</mat-datepicker-toggle>
+<mat-datepicker #myDatepicker></mat-datepicker>
+```
```
mmalerba: can you turn this into an embedded example?

src/lib/datepicker/datepicker-toggle.ts
```diff
@@ -25,6 +27,13 @@ import {MatDatepicker} from './datepicker';
 import {MatDatepickerIntl} from './datepicker-intl';
 
 
+/** Can be used to override the icon of a `matDatepickerToggle`. */
+@Directive({
+  selector: '[matDatepickerToggleIcon]'
+})
+export class MatDatepickerToggleIcon {}
```
mmalerba: Do we need a special directive? what if we just treat any projected content as a custom icon?

crisbeto: We could, but it gets a little dirtier when you try to figure out whether to hide the default icon. We'd have to loop through all the `children` of the content element and count the amount of element nodes (skipping comments, text nodes etc.) on every change detection.

src/lib/datepicker/datepicker.md
crisbeto: Done.

src/lib/datepicker/datepicker-toggle.ts
mmalerba: Interesting, I thought there was an easier way to check if any content was projected, but it looks like you're right

PR #9489, opened 2018-01-20
feat(table): allow data input to be array, stream

### LINKED ISSUE #8227: Add detailed documentation for using DataSource with MatTable
Proposal: Add detailed documentation for using DataSource with Material Table. Current documentation on DataSource is very high level. Tutorial level documentation, starting with basic table, then adding sort and paging would be good.

Level of documentation to make using Mat-Table and DataSources, along with Observables more developer friendly.  

Very minimal level of documentation regarding how DataSources work with Mat-Table, pagination, and observables.

Have an AngularJS developer add a Mat-Table to a seed project, and use DataSource, Observable, http.get(), and pagination/sorting on the table. Documentation seems lacking to do this. A tutorial would be great.

Make Angular (4) more accessible to developers for a popular use case (full featured DataTable).

Latest Angular 4

andrewseguin:
Removes requirement of implementing/using a `DataSource` to use the table. Instead, users can directly provide a data array or stream of data arrays directly as the data source.

@willshowell Can you take a look, especially the revamped docs?

Closes #8227

### DIFF
--- a/src/cdk/table/table.spec.ts
+++ b/src/cdk/table/table.spec.ts
@@ -11,7 +11,8 @@ import {
   getTableMissingMatchingRowDefError,
   getTableMissingRowDefsError,
   getTableMultipleDefaultRowDefsError,
-  getTableUnknownColumnError
+  getTableUnknownColumnError,
+  getTableUnknownDataSourceError
 } from './table-errors';
 import {CdkHeaderRowDef, CdkRowDef} from './row';
 import {CdkColumnDef} from './cell';
@@ -45,6 +46,7 @@ describe('CdkTable', () => {
         BooleanRowCdkTableApp,
         WrapperCdkTableApp,
         OuterTableApp,
+        CdkTableWithDifferentDataInputsApp,
       ],
     }).compileComponents();
   }));
@@ -110,6 +112,138 @@ describe('CdkTable', () => {
           expect(cell.getAttribute('role')).toBe('gridcell');
         });
       });
+    });
+  });
+
+  describe('with different data inputs other than data source', () => {
+    let dataInputFixture: ComponentFixture<CdkTableWithDifferentDataInputsApp>;
+    let dataInputComponent: CdkTableWithDifferentDataInputsApp;
+    let dataInputTableElement: HTMLElement;
+
+    let baseData: TestData[] = [
+      {a: 'a_1', b: 'b_1', c: 'c_1'},
+      {a: 'a_2', b: 'b_2', c: 'c_2'},
+      {a: 'a_3', b: 'b_3', c: 'c_3'},
+    ];
+
+    beforeEach(() => {
+      dataInputFixture = TestBed.createComponent(CdkTableWithDifferentDataInputsApp);
+      dataInputComponent = dataInputFixture.componentInstance;
+      dataInputFixture.detectChanges();
+
+      dataInputTableElement = dataInputFixture.nativeElement.querySelector('cdk-table');
+    });
+
+    it('should render with data array input', () => {
+      const data = baseData.slice();
+      dataInputComponent.dataSource = data;
+      dataInputFixture.detectChanges();
+
+      const expectedRender = [
+        ['Column A', 'Column B', 'Column C'],
+        ['a_1', 'b_1', 'c_1'],
+        ['a_2', 'b_2', 'c_2'],
+        ['a_3', 'b_3', 'c_3'],
+      ];
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Push data to the array but neglect to tell the table, should be no change
+      data.push({a: 'a_4', b: 'b_4', c: 'c_4'});
+
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Notify table of the change, expect another row
+      dataInputComponent.table.renderRows();
+      dataInputFixture.detectChanges();
+
+      expectedRender.push(['a_4', 'b_4', 'c_4']);
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Remove a row and expect the change in rows
+      data.pop();
+      dataInputComponent.table.renderRows();
+
+      expectedRender.pop();
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Remove the data input entirely and expect no rows - just header.
+      dataInputComponent.dataSource = null;
+      dataInputFixture.detectChanges();
+
+      expectTableToMatchContent(dataInputTableElement, [expectedRender[0]]);
+
+      // Add back the data to verify that it renders rows
+      dataInputComponent.dataSource = data;
+      dataInputFixture.detectChanges();
+
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+    });
+
+    it('should render with data stream input', () => {
+      const data = baseData.slice();
+      const stream = new BehaviorSubject<TestData[]>(data);
+      dataInputComponent.dataSource = stream;
+      dataInputFixture.detectChanges();
+
+      const expectedRender = [
+        ['Column A', 'Column B', 'Column C'],
+        ['a_1', 'b_1', 'c_1'],
+        ['a_2', 'b_2', 'c_2'],
+        ['a_3', 'b_3', 'c_3'],
+      ];
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Push data to the array and emit the data array on the stream
+      data.push({a: 'a_4', b: 'b_4', c: 'c_4'});
+      stream.next(data);
+      dataInputFixture.detectChanges();
+
+      expectedRender.push(['a_4', 'b_4', 'c_4']);
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Push data to the array but rather than emitting, call renderRows.
+      data.push({a: 'a_5', b: 'b_5', c: 'c_5'});
+      dataInputComponent.table.renderRows();
+      dataInputFixture.detectChanges();
+
+      expectedRender.push(['a_5', 'b_5', 'c_5']);
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Remove a row and expect the change in rows
+      data.pop();
+      expectedRender.pop();
+      stream.next(data);
+
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Remove the data input entirely and expect no rows - just header.
+      dataInputComponent.dataSource = null;
+      dataInputFixture.detectChanges();
+
+      expectTableToMatchContent(dataInputTableElement, [expectedRender[0]]);
+
+      // Add back the data to verify that it renders rows
+      dataInputComponent.dataSource = stream;
+      dataInputFixture.detectChanges();
+
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+    });
+
+    it('should throw an error if the data source is not valid', () => {
+      dataInputComponent.dataSource = {invalid: 'dataSource'};
+
+      expect(() => dataInputFixture.detectChanges())
+          .toThrowError(getTableUnknownDataSourceError().message);
+    });
+
+    it('should throw an error if the data source is not valid', () => {
+      dataInputComponent.dataSource = undefined;
+      dataInputFixture.detectChanges();
+
+      // Expect the table to render just the header, no rows
+      expectTableToMatchContent(dataInputTableElement, [
+        ['Column A', 'Column B', 'Column C']
+      ]);
     });
   });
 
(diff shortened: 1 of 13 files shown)


src/lib/table/table.md
```diff
@@ -35,60 +48,89 @@ Here's a simple column definition with the name `'userName'`. The header cell co
...
 
+Note that this list of columns provided to the rows can be in any order, not necessary the order in
+which you wrote the column definitions. Also, you do not necessarily have to include every column
+that was defined in your template.
+
+This means that by changing your column list provided to the rows, you can easily re-order and
+include/exclude columns dynamically.  
+
+### Alternative ways to provide data
+
```
willshowell: Consider adding a little blurb to refresh the reader of what the original option is

> ### Alternative ways to provide data
> The simplest way to provide data to your table is by passing a data array. More complex use-cases may benefit from a more flexible approach. _(or something like that)_
>
> #### Observable stream of data arrays
> ...
>
> #### DataSource
> ...

src/lib/table/table.md
```diff
@@ -35,60 +48,89 @@ Here's a simple column definition with the name `'userName'`. The header cell co
...
+#### DataSource
+
+Also, another approach to providing data is by implementing a `DataSource`. This is simply
+an interface that has two functions: `connect` and `disconnect`. After passing a DataSource to the
+table, the table will call `connect()` to receive a stream that emits the data array that should be
+rendered. The table will call `disconnect` when the table is destroyed, which may be the right time
+to clean up any subscriptions that may have been registered during the `connect` process.
+
+The benefit to creating a `DataSource` class is that it can decouple rendering logic from your
+component and keep it encapsulated in a separate class (e.g. sorting, filtering, pagination, e.g.).
```
willshowell: It might be helpful to eventually have an example for each of these that is just like `table-basic`, but using the alternative approach.

andrewseguin: +1 I think this would be useful to demonstrate the differences, I'll make an issue following this PR getting merged

src/lib/table/table.md
```diff
@@ -1,29 +1,42 @@
...
+<mat-table [data]=”myDataArray”>
+  ...
+</mat-table>
+```
+
+Since the table optimizes for performance, it will not automatically check for changes to the data
+array. Instead, when objects are added, removed, or moved on the data array, you can trigger an 
+update to the table's rendered rows by calling its `renderRows()` function. 
+
+See the section "Alternative ways to provide data" for more approaches to providing data to the
```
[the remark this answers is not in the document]
andrewseguin: I think they'll be a link we can use but not sure if we've done that yet in our docs. We should make sure that's supported or easy to do

src/cdk/table/table.spec.ts
```diff
@@ -113,6 +115,111 @@ describe('CdkTable', () => {
...
+        ['a_3', 'b_3', 'c_3'],
+      ];
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
+
+      // Push data to the array
+      data.push({a: 'a_4', b: 'b_4', c: 'c_4'});
+      expectedRender.push(['a_4', 'b_4', 'c_4']);
+      stream.next(data);
+      dataInputFixture.detectChanges();
+      expectTableToMatchContent(dataInputTableElement, expectedRender);
```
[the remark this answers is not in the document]
andrewseguin: Good suggestion, I do think there's value in making sure that the renderRows will work like this. I expect they'll be a subset of users who use a stream to provide some array but would like to still manipulate the array and not think about emitting again

andrewseguin:
I'm excited as well, I think this will help make things a lot easier and reduce the learning curve of the table. Next for me is to add some kind of `mat-simple-column` (like in the demos) so the learning curve of the templates are reduced too

src/cdk/table/table.ts
```diff
@@ -153,17 +157,32 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
   private _trackByFn: TrackByFunction<T>;
 
   /**
-   * Provides a stream containing the latest data array to render. Influenced by the table's
-   * stream of view window (what rows are currently on screen).
+   * Sets the table's source of data, which can be provided in three ways (in order of complexity):
```
jelbourn: Omit "Sets", since this is a property, not a method

src/cdk/table/table.ts
```diff
@@ -153,17 +157,32 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
+   *   - Simple data array (each object represents one table row).
+   *   - Observable stream that emits a data array each time the array changes.
+   *   - `DataSource` object that implements the connect/disconnect interface.
+   *
+   * If a data array is provided, the table must be notified when the array's objects are
+   * added, removed, or moved. This can be done by calling the `renderRows()` function which will
+   * render the diff since the last table render.
```
jelbourn: What about when the array reference changes?

src/cdk/table/table.ts
```diff
@@ -153,17 +157,32 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
+   *
+   * When providing an Observable stream, the table will trigger an update automatically when the
+   * stream emits a new array of data.
```
jelbourn: Omit "a new array of data"?

src/cdk/table/table.ts
```diff
@@ -247,11 +266,50 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
+   * input or retrieved through an Observable stream (directly or from a DataSource).
+   * Checks for differences in the data since the last diff to perform only the necessary
+   * changes (add/remove/move rows).
+   *
+   * If the table's data source is a DataSource or Observable, this will be invoked each time the
+   * provided Observable stream emits a new data array. Otherwise if your data is being set
+   * directly, you will need to call this function whenever data in the provided array is
+   * added/removed/moved in-place.
+   */
+  renderRows() {
```
jelbourn: What's your thinking on `renderRows` vs. `refresh` or `render`?

src/cdk/table/table.ts
```diff
@@ -247,11 +266,50 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
     }
   }
 
+  /**
+   * Renders rows based on the table's latest set of data, which was either provided directly as an
+   * input or retrieved through an Observable stream (directly or from a DataSource).
+   * Checks for differences in the data since the last diff to perform only the necessary
+   * changes (add/remove/move rows).
+   *
+   * If the table's data source is a DataSource or Observable, this will be invoked each time the
```
jelbourn: "invoked _automatically_"

src/cdk/table/table.ts
```diff
@@ -247,11 +266,50 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
+   * provided Observable stream emits a new data array. Otherwise if your data is being set
```
jelbourn: "Otherwise if your data is being set directory" -> "If the dataSource is an array"

"you will need to call this function..." -> "this function must be manually called to render any changes"

src/cdk/table/table.ts
```diff
@@ -357,11 +417,24 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
-        this._data = data;
-        this._renderRowChanges();
-      });
+    let dataStream: Observable<T[]> | undefined;
+
+    if (this.dataSource instanceof DataSource) {
+      dataStream = this.dataSource.connect(this);
+    } else if (this.dataSource instanceof Observable) {
+      dataStream = this.dataSource;
+    } else if (this.dataSource instanceof Array) {
```
jelbourn: `Array.isArray`

src/cdk/table/table.ts
```diff
@@ -357,11 +417,24 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
   /** Set up a subscription for the data provided by the data source. */
   private _observeRenderChanges() {
-    this._renderChangeSubscription = this.dataSource.connect(this).pipe(takeUntil(this._onDestroy))
-      .subscribe(data => {
-        this._data = data;
-        this._renderRowChanges();
-      });
+    let dataStream: Observable<T[]> | undefined;
+
+    if (this.dataSource instanceof DataSource) {
```
jelbourn: Weren't we going to check the properties instead of using `instanceof`? If people had been copying what `MatTableDataSource` was doing this would be a problem.

src/cdk/table/table.ts
```diff
@@ -42,8 +41,11 @@ import {
   getTableMissingMatchingRowDefError,
   getTableMissingRowDefsError,
   getTableMultipleDefaultRowDefsError,
-  getTableUnknownColumnError
+  getTableUnknownColumnError,
+  getTableUnknownDataSourceError
 } from './table-errors';
+import {Observable} from 'rxjs/Observable';
+import {of} from 'rxjs/observable/of';
```
jelbourn: `observableOf`, since `of` is an ES6 keyword

src/cdk/table/table.ts
```diff
@@ -357,11 +417,24 @@ export class CdkTable<T> implements CollectionViewer, OnInit, AfterContentChecke
...
+
+    if (this.dataSource instanceof DataSource) {
+      dataStream = this.dataSource.connect(this);
+    } else if (this.dataSource instanceof Observable) {
+      dataStream = this.dataSource;
+    } else if (this.dataSource instanceof Array) {
+      dataStream = of(this.dataSource);
+    }
+
+    if (dataStream === undefined) {
```
jelbourn: What about cases where someone is binding to an array that hasn't been initialized yet?

src/demo-app/table/data-input-table/data-input-table.scss
```diff
@@ -0,0 +1,20 @@
+.actions {
```
jelbourn: Prefix demo styles with `demo-`

src/lib/table/table.md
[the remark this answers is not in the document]
jelbourn: The TOC will link to section headers, but any links in the content have to be manually added.

I would also rephrase this paragraph to something like
```markdown
While an array is the _simplest_ way to bind data into the data source, it is also
the most limited. For more complex applications, using a `DataSource` instance
is recommended. See xxx below for more information.
```

src/lib/table/table.md
```diff
@@ -1,29 +1,42 @@
...
+Since the table optimizes for performance, it will not automatically check for changes to the data
+array. Instead, when objects are added, removed, or moved on the data array, you can trigger an 
+update to the table's rendered rows by calling its `renderRows()` function. 
+
+See the section "Alternative ways to provide data" for more approaches to providing data to the 
+table.
 
-Start by writing your table's column definitions. Each column definition should be given a unique 
-name and contain the content for its header and row cells.
+#### 2. Define the table's columns
```
jelbourn: Would it make sense to say "column templates" (and "row templates" below)? Having "Define the table's rows" feels weird since the data drives the rows, you're just specifying the bindings.

src/lib/table/table.md
```diff
@@ -35,60 +48,93 @@ Here's a simple column definition with the name `'userName'`. The header cell co
...
 ```
 
+Note that this list of columns provided to the rows can be in any order, not necessary the order in
+which you wrote the column definitions. Also, you do not necessarily have to include every column
+that was defined in your template.
+
+This means that by changing your column list provided to the rows, you can easily re-order and
+include/exclude columns dynamically.  
+
+### Alternative ways to provide data
```
jelbourn: I would say "Advanced data sources"

src/lib/table/table.md
```diff
@@ -35,60 +48,93 @@ Here's a simple column definition with the name `'userName'`. The header cell co
...
+
+#### Observable stream of data arrays
+
+An alternative approach to providing data to the table is by passing an Observable stream that emits
+the data array to be rendered each time it is changed. The table will listen to this stream and 
+automatically trigger an update to the rows each time a new data array is emitted.
+
+#### DataSource
+
+Also, another approach to providing data is by implementing a `DataSource`. This is simply
```
jelbourn: I would change this first sentence to something like
```
For most real-world applications, providing the table a `DataSource` instance will be the best way to manage data.
```

It's also a base class, not an interface. I would put the "Why use a data source" bits before the "what is it" bits; we should say something like "The `DataSource` is meant to serve a place to encapsulate any sorting, filtering, pagination, and data retrieval logic specific to the application."

src/lib/table/table.md
```diff
@@ -35,60 +48,93 @@ Here's a simple column definition with the name `'userName'`. The header cell co
...
+performant and accessible way.
+
+You'll notice that the table itself doesn't come out of the box with a lot of features, but expects
+that the table will be included in a composition of components that fills out its features.
+
+For example, you can add sorting and pagination to the table by using MatSort and MatPaginator and
+mutating the data provided to the table according to their outputs.
+
+To simplify the use case of having a table that can sort, paginate, and filter an array of data,
+the Angular Material library comes with a `MatTableDataSource` that has already implemented
```
jelbourn: I still want to rename this to `ClientArrayDataSource` (or `ClientDataSource` or `ArrayDataSource` or even `SimpleDataSource`)

My problem with the current name is that it implies that it is **the** way to provide data for a `<mat-table>`, wich may lead people to think that the table doesn't support server side sorting/pagination/filtering, or even just other kinds of filtering than search.

src/cdk/table/table.ts
andrewseguin: No strong opinion really, I'm open to any of those. Maybe `refresh` sounds more approachable for users to call?

src/cdk/table/table.ts
andrewseguin: I think we decided against it due to the possibility of property renaming/minification

src/cdk/table/table.ts
andrewseguin: Ah good point, I think this would come up. I'll return if there is no data source.

src/cdk/table/table.ts
jelbourn: Ah, right.

src/lib/table/table.md
andrewseguin: Sure, I'm open to renaming. We should brainstorm offline to figure out the best name

andrewseguin:
Pushes changes according to the comments, ready for another round

src/lib/table/table.md
```diff
@@ -1,29 +1,42 @@
...
-#### 1. Define the table's columns
+#### 1. Write your mat-table and provide data
+
+Begin by creating a `<mat-table>` component in your template and passing in data.
+  
+The simplest way to provide data to the table is by passing a data array to the table's `data` 
+input. The table will take the array and render a row for each object in the data array.
+
+```html
+<mat-table [data]=”myDataArray”>
```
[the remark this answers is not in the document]
jelbourn: Yeah, it should be `dataSource`

andrewseguin:
Ready for another round of review

andrewseguin:
Alright, mostly sure now that it's good. Had an odd time merging conflicts

PR #9490, opened 2018-01-20
add @deletion-target to all deprecated APIs

mmalerba:
Should make our job easier when preparing for 6.0.0, we can search for `@deletion-target 6.0.0` and delete them, then search for `@deletion-target 7.0.0` and make sure they all have `@deprecated`

We can add some tooling/lint checks around it if we decide it would be useful


src/lib/core/theming/_palette.scss
```diff
@@ -7,16 +7,22 @@
 
 
 /// @deprecated renamed to $dark-primary-text.
+/// @deletion-target 7.0.0
```
crisbeto: It's what was there before, but shouldn't we use the double slashes here?

mmalerba:
Addressed comments and rebased

PR #9495, opened 2018-01-20
feat(autocomplete): add the ability to highlight the first option on open

### LINKED ISSUE #8423: Autocomplete - add option to always treat first option as active
Feature Request

Should be able to manually/programmatically set the first option in the autocomplete panel as the activeOption, so that it receives the appropriate styles and is selected when the user presses enter.

User must either use the down key or hover with the mouse on the option to make it active.

N/A

Our users are going to be limited to selecting only options from the autocomplete. We're experiencing problems where users are either typing in a partial value or the full value, hitting the enter key, and not understanding why that option isn't being selected.

See attached images
partial-match
full-match

Angular - 4.4.6
Material - 2.0.0-beta.12

There is a way to select the first option by using `matAutocomplete.options.first.select()` and binding that to the `keyup.enter` event. I can also forcibly set the active styles on the first option by using `matAutocomplete.options.first.setActiveStyles()` but this doesn't set the first option as the `activeOption`. Without that, the key manager doesn't know that the first option is already highlighted and pressing the `DOWN_ARROW` should make the second option active.

The typeahead using bootstrap and AngularJS has the exact desired behavior for a working example: https://angular-ui.github.io/bootstrap/#!#typeahead

crisbeto:
Adds the ability for the consumer opt-in to having the autocomplete highlight the first option when opened. Includes an injection token that allows it to be configured globally.

Fixes #8423.

### DIFF
--- a/src/lib/autocomplete/autocomplete.ts
+++ b/src/lib/autocomplete/autocomplete.ts
@@ -19,6 +19,9 @@ import {
   ChangeDetectionStrategy,
   EventEmitter,
   Output,
+  InjectionToken,
+  Inject,
+  Optional,
 } from '@angular/core';
 import {
   MatOption,
@@ -28,6 +31,7 @@ import {
   CanDisableRipple,
 } from '@angular/material/core';
 import {ActiveDescendantKeyManager} from '@angular/cdk/a11y';
+import {coerceBooleanProperty} from '@angular/cdk/coercion';
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/lib/autocomplete/autocomplete.ts
```diff
@@ -98,6 +112,18 @@ export class MatAutocomplete extends _MatAutocompleteMixinBase implements AfterC
   /** Function that maps an option's control value to its display value in the trigger. */
   @Input() displayWith: ((value: any) => string) | null = null;
 
+  /**
+   * Whether the first option should be highlighted when the autocomplete panel is opened.
+   * Can be configured globally through the `MAT_AUTOCOMPLETE_DEFAULT_OPTIONS` token.
+   */
+  @Input()
+  get highlightFirstOption(): boolean { return this._highlightFirstOption; }
```
jelbourn: Chatted w/ Miles and Joey, we came to `autoActiveOption` as the proposal for this. I like using "active" since it aligns with the term we use in list-key-manager and in datepicker. Thoughts?

crisbeto:
Good to go @jelbourn.

badre429:
finally this feature can replace select2 search functionality  thanks 
can't wait for next release

PR #9499, opened 2018-01-20
refactor(datepicker): remove MatIconModule dependency

crisbeto:
Removes the dependency between the `MatDatepickerModule` and `MatIconModule`, because we were only using a handful of styles but were pulling in all of the overhead that comes with the icon module (e.g. the icon registry, the dependency to the `HttpClientModule` etc.).


jelbourn:
Just needs rebase

PR #9506, opened 2018-01-21
fix(autocomplete): panel not closing on IE when selecting an option with an empty string display value

### LINKED ISSUE #9479: [BUG] mat-autocomplete doesn't close using displayWith under IE11
Bug
Similar to: https://github.com/angular/material2/issues/5165 (not duplicate) 

mat-autocomplete-panel should be close after selected an option

mat-autocomplete-panel continues to be opened even after to have selected an option

Run the stackblitz: https://stackblitz.com/edit/angular-gieptd

Angular 5.2.0
Node: 8.6.0
MAterial: 5.1.0
Windows 7 Pro 64 bits
TypeScript: 2.6.2

NB: For the moment, the workaround consists to use the reset function available on the formControl. IE11 worls properly in this context

crisbeto:
Fixes the autocomplete panel reopening on IE11 when the user selects an option whose display value is an empty string.

Fixes #9479.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -121,6 +121,9 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   private _portal: TemplatePortal;
   private _componentDestroyed = false;
 
+  /** Old value of the native input. Used to work around issues with the `input` event on IE. */
+  private _previousValue: string | number | null;
+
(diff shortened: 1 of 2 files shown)


jelbourn:
It's possible that checking `activeElement` may start to make some tests fail. We'll see how the presubmit goes.

crisbeto:
The `activeElement` check was already there, this only introduces the check for the value.

tinayuangao:

crisbeto:
Rebased.

PR #9536, opened 2018-01-23
fix(button): no color set for flat and icon buttons

### LINKED ISSUE #4614: Text color not visible in md-raised-button within md-toolbar
The .mat-toolbar.mat-primary style (for example) sets the text color to white. When adding an md-raised-button (no color) to the toolbar, the button text is invisible.

Visible text in the raised buttons.

Invisible text in the raised buttons.

Create a md-toolbar with primary color background (default). Then, add a md-raised-button without [color] into it with containing some text. The text will not show up.

[Plunker example](http://plnkr.co/edit/9lHIdpr4p3XYl5jXCfOp?p=preview)

The raised buttons should work even if embedded within a toolbar. They do not work unless a specific color (warn, primary, accent) is set.

Angular 2, Angular Material 2 beta 5, Typescript (N/A), Browser (N/A)

### LINKED ISSUE #9231: mat-*-button don't respect foreground color of dark theme when [color] not set
When using dark theme, and using 'default' color (so [color] not set), button text (or icon) is rendered black, and hover overlay is also dark. I think ripple effect is actually correct.

Material buttons should obey theme colors.

The don't

Here is very simple demo:
https://stackblitz.com/edit/angular-material2-issue-t2aufv

Angular 5.1.2, Material 5.0.3

### LINKED ISSUE #9634: Basic raised button text color is wrong in dark theme
```html
<button mat-raised-button>Button</button>
```

It should looks like example https://material.angular.io/components/button/examples
image

image

https://stackblitz.com/edit/angular-material2-issue-2jzs4u?file=styles.scss
I only change theme to pink-bluegrey.css

Just like a bug. Because I follow example code but get different result with example UI.

Angular Material 5.1.0

DevVersion:
* Currently for flat buttons and icon buttons, the font color is not set by the theme. This can cause the text to be invisible on those buttons in a dark theme. From now on, those buttons will also receive a font color by the theme.
* Flat buttons and icon buttons inside of a `<mat-toolbar>` will inherit the font color from the toolbar row. This ensures that those buttons are looking as expected in themed toolbars (e.g. primary, accent, warn)

Fixes #4614. Fixes #9231. Fixes #9634

### DIFF
--- a/src/demo-app/toolbar/toolbar-demo.html
+++ b/src/demo-app/toolbar/toolbar-demo.html
@@ -1,40 +1,58 @@
 <div class="demo-toolbar">
-
   <p>
     <mat-toolbar>
-      <mat-icon class="demo-toolbar-icon">menu</mat-icon>
-      <span>Default Toolbar</span>
+      <button mat-icon-button>
+        <mat-icon>menu</mat-icon>
+      </button>
 
+      <span>Default Toolbar</span>
       <span class="demo-fill-remaining"></span>
 
-      <mat-icon>code</mat-icon>
+      <button mat-icon-button>
+        <mat-icon>code</mat-icon>
+      </button>
+
+      <button mat-icon-button color="warn">
+        <mat-icon>code</mat-icon>
+      </button>
     </mat-toolbar>
   </p>
 
   <p>
     <mat-toolbar color="primary">
-      <mat-icon class="demo-toolbar-icon">menu</mat-icon>
+      <button mat-icon-button>
+        <mat-icon>menu</mat-icon>
+      </button>
+
       <span>Primary Toolbar</span>
-
       <span class="demo-fill-remaining"></span>
 
-      <mat-icon>code</mat-icon>
+      <button mat-raised-button>Text</button>
+      <button mat-raised-button color="accent">Accent</button>
     </mat-toolbar>
   </p>
 
   <p>
     <mat-toolbar color="accent">
-      <mat-icon class="demo-toolbar-icon">menu</mat-icon>
+      <button mat-icon-button>
+        <mat-icon>menu</mat-icon>
+      </button>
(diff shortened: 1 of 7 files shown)


src/demo-app/toolbar/toolbar-demo.html
```diff
@@ -1,40 +1,58 @@
 <div class="demo-toolbar">
-
   <p>
     <mat-toolbar>
-      <mat-icon class="demo-toolbar-icon">menu</mat-icon>
-      <span>Default Toolbar</span>
+      <button mat-icon-button>
```
tinayuangao: nit: do we want to keep one `mat-icon` without button in the demo?

src/lib/button/button.scss
```diff
@@ -52,12 +52,16 @@
...
 .mat-button, .mat-raised-button, .mat-icon-button, .mat-fab, .mat-mini-fab {
-  color: currentColor;
   .mat-button-wrapper > * {
     vertical-align: middle;
   }
 }
 
+// Buttons that don't have a background color can inherit the color from its parent.
+.mat-button, .mat-icon-button {
+  color: inherit;
```
jelbourn: `inherit` or `currentColor`? Is `inherit` supported on IE11?

src/demo-app/toolbar/toolbar-demo.html
DevVersion: Good point. I have such plain `<mat-icon>` elements in the demo toolbars below.

src/lib/button/button.scss
DevVersion: Yeah, just confirmed that `inherited` is supported.

DevVersion:
@crisbeto Updated to what we discussed. Please have a look.

PR #9542, opened 2018-01-23
fix(focus-monitor): clear all pending timeouts when unregistering global listeners

josephperrott:


src/cdk/a11y/focus-monitor.ts
```diff
@@ -202,7 +201,7 @@ export class FocusMonitor implements OnDestroy {
 
     // Note: we listen to events in the capture phase so we can detect them even if the user stops
     // propagation.
-    this._ngZone.runOutsideAngular(() => {
+    Zone.root.run(() => {
```
mmalerba: comment about why we need to do this

josephperrott:
This should be ready to look at again as it now clears the timeouts that are created on destruction.

PR #9557, opened 2018-01-24
feat(chips): Add chip avatar and chip trailing icon

tinayuangao:
demo: https://tina-demo1.firebaseapp.com/chips

### DIFF
--- a/src/demo-app/chips/chips-demo.html
+++ b/src/demo-app/chips/chips-demo.html
@@ -39,12 +39,12 @@
         <mat-chip>
           <mat-icon matChipAvatar>home</mat-icon>
           Home
-          <mat-icon matChipRemove matChipTrailingIcon>cancel</mat-icon>
+          <mat-icon matChipRemove>cancel</mat-icon>
         </mat-chip>
         <mat-chip color="accent" selected="true">
           <mat-chip-avatar>P</mat-chip-avatar>
           Portel
-          <mat-icon matChipRemove matChipTrailingIcon>cancel</mat-icon>
+          <mat-icon matChipRemove>cancel</mat-icon>
         </mat-chip>
 
         <mat-chip>
@@ -54,7 +54,7 @@
 
         <mat-chip>
           Koby
-          <mat-icon matChipRemove matChipTrailingIcon>cancel</mat-icon>
+          <mat-icon matChipRemove>cancel</mat-icon>
         </mat-chip>
 
         <mat-chip>
@@ -69,7 +69,16 @@
         <mat-chip selected="true" color="warn">
           <img src="https://material.angularjs.org/material2_assets/ngconf/Husi.png" matChipAvatar>
           Husi
-          <mat-icon matChipRemove matChipTrailingIcon>cancel</mat-icon>
+          <mat-icon matChipRemove>cancel</mat-icon>
+        </mat-chip>
+
+        <mat-chip>
+          Good
+          <mat-icon matChipTrailingIcon>star</mat-icon>
+        </mat-chip>
+        <mat-chip>
+          Bad
+          <mat-icon matChipTrailingIcon>star_border</mat-icon>
         </mat-chip>
       </mat-chip-list>
     </mat-card-content>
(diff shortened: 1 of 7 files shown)


src/lib/chips/chips.scss
```diff
@@ -58,42 +45,81 @@ $mat-chip-input-margin: 3px;
...
-    margin-bottom: $mat-chips-chip-margin;
+  &.mat-chip-with-trailing-icon {
+    padding-top: $mat-chip-remove-vertical-padding;
+    padding-bottom: $mat-chip-remove-vertical-padding;
+    padding-right: $mat-chip-remove-after-padding;
 
     [dir='rtl'] & {
-      margin: 0;
-      margin-bottom: $mat-chips-chip-margin;
+      padding-left: $mat-chip-remove-after-padding;
```
mmalerba: do you also need to zero out the `padding-right`?

src/lib/chips/chips.scss
```diff
@@ -58,42 +45,81 @@ $mat-chip-input-margin: 3px;
...
     }
+  }
+
+  &.mat-chip-with-avatar {
+    padding-left: $mat-chip-avatar-before-padding;
 
-    &:last-child, [dir='rtl'] &:last-child {
-      margin-bottom: 0;
+    [dir='rtl'] & {
+      padding-right: $mat-chip-avatar-before-padding;
```
mmalerba: zero out `padding-left`?

src/lib/chips/chips.scss
```diff
@@ -58,42 +45,81 @@ $mat-chip-input-margin: 3px;
...
+    }
+  }
+
+  .mat-chip-avatar {
+    width: $mat-chip-avatar-size;
+    height: $mat-chip-avatar-size;
+    margin-right: $mat-chip-avatar-after-margin;
+
+    [dir='rtl'] & {
+      margin-left: $mat-chip-avatar-after-margin;
```
mmalerba: zero out `margin-right`

src/lib/chips/chips.scss
```diff
@@ -58,42 +45,81 @@ $mat-chip-input-margin: 3px;
...
+    height: $mat-chip-remove-size;
+    cursor: pointer;
+  }
+
+  .mat-chip-remove,
+  .mat-chip-trailing-icon {
+    margin-left: $mat-chip-remove-before-margin;
+
+    [dir='rtl'] & {
+      margin-right: $mat-chip-remove-before-margin;
```
mmalerba: zero out `margin-left`

src/demo-app/chips/chips-demo.html
```diff
@@ -31,6 +31,47 @@ <h4>Advanced</h4>
...
       <div>{{message}}</div>
+
+      <h4>With avatar and icons</h4>
+
+
+      <mat-chip-list>
+        <mat-chip>
+          <mat-icon matChipAvatar>home</mat-icon>
+          Home
+          <mat-icon matChipRemove matChipTrailingIcon>cancel</mat-icon>
```
mmalerba: what does `matChipTrailingIcon` do? it looks the same as the example above where you only applied `matChipRemove`

src/demo-app/chips/chips-demo.html
```diff
@@ -31,6 +31,47 @@ <h4>Advanced</h4>
         </mat-chip>
       </mat-chip-list>
       <div>{{message}}</div>
+
```
mmalerba: add color example with avatar

src/demo-app/chips/chips-demo.html
mmalerba: Ah ok, do these examples need both directives? It seems like just matChipRemove would be sufficient since does everything that trailing icon does + some extra behavior. In fact you could probably make the CSS simpler by having `matChipRemove` just apply the `mat-chip-trailing-icon` class .

Also should add an example of a trailing icon that isn't a remove icon

tinayuangao:
Added examples in demo, and added `mat-chip-trailing-icon` to `MatChipRemove`. Thanks!

mmalerba:
LGTM

PR #6938, opened 2017-09-08
refactor: remove css transform utility

crisbeto:
The `applyCssTransform` utility is no longer necessary since none of the browser we support need a prefix.

PR #9571, opened 2018-01-24
refactor: remove css transform utility

crisbeto:
The applyCssTransform utility is no longer necessary since none of the browser we support need a prefix.

This is a resubmit of #6938.


DevVersion:
Not sure whether this is `major`, `minor` or just `patch`, because technically this is a breaking change, since the `applyCssTransform` function was publicly exported.

crisbeto:
True, although I highly doubt that anybody is using this. We can see look at the results of the presubmit.

PR #9575, opened 2018-01-24
fix(datepicker): dateInput being fired on load in IE when input has a placeholder

crisbeto:
Fixes the `dateInput` event from the datepicker input directive being fired on load, focus and blur, if the input has a placeholder in IE.

### DIFF
--- a/src/lib/datepicker/datepicker-input.ts
+++ b/src/lib/datepicker/datepicker-input.ts
@@ -34,6 +34,7 @@ import {DateAdapter, MAT_DATE_FORMATS, MatDateFormats} from '@angular/material/c
 import {MatFormField} from '@angular/material/form-field';
 import {MAT_INPUT_VALUE_ACCESSOR} from '@angular/material/input';
 import {Subscription} from 'rxjs';
+import {DOCUMENT} from '@angular/common';
 import {MatDatepicker} from './datepicker';
 import {createMissingDateImplError} from './datepicker-errors';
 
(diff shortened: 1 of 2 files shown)


mmalerba:
Is this something we want to fix here? Feels like something that maybe Angular core should deal with. This issue is kind of related: https://github.com/angular/angular/issues/14440

crisbeto:
It's something that we also handle in the autocomplete. I'm not sure whether it would be particularly clean for core to handle it, because it means adding extra logic every time somebody binds an `input` event.

src/lib/datepicker/datepicker-input.ts
```diff
@@ -229,14 +228,18 @@ export class MatDatepickerInput<D> implements AfterContentInit, ControlValueAcce
...
+      @Optional() private _formField: MatFormField,
+      @Inject(DOCUMENT) document?: any) {
     if (!this._dateAdapter) {
       throw createMissingDateImplError('DateAdapter');
     }
     if (!this._dateFormats) {
       throw createMissingDateImplError('MAT_DATE_FORMATS');
     }
 
+    // @deletion-target 6.0.0 make the `document` a required param.
```
mmalerba: nit: move comment above line where param is defined

crisbeto: Done.

crisbeto:
We don't need to do it on `matInput`, because the handler doesn't do anything there.

mmalerba:
Ah, I think there's an issue with the pristine value for the input on IE, but I guess that's a forms issue

mmalerba:
caretaker note: couple failing unit tests that need investigation

mmalerba:
Did some investigation of the failing tests, I think it's a consequence of people not focusing the input in their tests before emitting input events on them. We probably have to bump this to a major change because I could see it breaking a lot of tests externally too.

MohammadRafik:
@crisbeto could you please rebase this when you get a chance?

crisbeto:
We don't support IE anymore so this PR shouldn't be necessary.

PR #9586, opened 2018-01-25
chore: make all event emitters / subjects readonly

### LINKED ISSUE #9582: Make all EventEmitters readonly
All of the `EventEmitter` instances on our components should be `readonly`.

rafaelss95:
Fixes #9582.

### DIFF
--- a/src/lib/autocomplete/autocomplete-trigger.ts
+++ b/src/lib/autocomplete/autocomplete-trigger.ts
@@ -131,7 +131,7 @@ export class MatAutocompleteTrigger implements ControlValueAccessor, OnDestroy {
   private _closingActionsSubscription: Subscription;
 
   /** Stream of keyboard events that can close the panel. */
-  private _closeKeyEventStream = new Subject<void>();
+  private readonly _closeKeyEventStream = new Subject<void>();
 
   /** View -> model callback called when value changes */
   _onChange: (value: any) => void = () => {};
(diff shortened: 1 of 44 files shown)


mmalerba:
I'm marking this for patch release even though it's technically breaking, nobody should be setting those though, I hope...

rafaelss95:
@jelbourn not really related to this PR, but while I was doing this PR, I noticed that were some `EventEmitter`'s that uses `next` instead of `emit`.

The cases are:

https://github.com/angular/material2/blob/7613f98a7b73711f4f67957d5f135700c41ebd7e/src/lib/paginator/paginator.ts#L193

https://github.com/angular/material2/blob/7613f98a7b73711f4f67957d5f135700c41ebd7e/src/lib/sort/sort.ts#L131

The problem with that is that it was deprecated on PR [@angular/angular#5302](https://github.com/angular/angular/pull/5302).

<hr>

On the other hand, this works perfectly because `EventEmitter` extends `Subject` and `emit` calls `next` (of `Subject`) behind the scenes:

https://github.com/angular/angular/blob/3aaf43f73c17570f61d3d0cc672a01695b716499/packages/core/src/event_emitter.ts#L78

In a nutshell, I'm not sure if there's a problem calling `next` instead of `emit`, but I thought it's good to tell you.

jelbourn:
Not a problem, but feel free to send a PR if you want

PR #9597, opened 2018-01-25
fix(paginator): change class name of increment and decrement icons

lharries:
Fix such that the previous page navigation button contains an icon with class decrement,
and the next page navigation button contains an icon with class increment. Previously
they were incorrectly swapped.

Noticed during fix for #9278

PR #9603, opened 2018-01-25
feat(paginator): Add functionality to jump to first and last page

### LINKED ISSUE #9278: [Paginator] Add functionality to jump to first and last page
Feature request

There should be 2 buttons like << and >> to jump to the first and last page.

There is no button for this. There is only a button for < and > prev and next.

Items per page: 10    -    1 - 10 of 100 < >

If you have a long list, often it is needed to jump to the end and back to the first page/first items.

Items per page: 10    -    1 - 10 of 100  **<<**  < >  **>>**

https://github.com/angular/material2/issues/7615 - as a reference.

lharries:
Add two buttons which allow you to jump to the first and last page.
Hidden by default and can be toggled with showHideFirstButtons

Fixes #9278, contains the following fixes and is therefore blocked by: #9597 and #9604 (and therefore also #9584)

### DIFF
--- a/src/demo-app/table/table-demo.html
+++ b/src/demo-app/table/table-demo.html
@@ -172,6 +172,24 @@
   </mat-paginator>
 </div>
 
+<h3>MatTable with First/Last Buttons Example</h3>
+
+<div class="demo-table-container demo-mat-table-example mat-elevation-z4">
+
+  <table-header-demo>
+  </table-header-demo>
+
+  <h4 class="paginator-output">{{paginatorOutput | json}}</h4>
+
+  <mat-paginator #paginator
+                [length]="_peopleDatabase.data.length"
+                [pageSize]="10"
+                [pageSizeOptions]="[5, 10, 25, 100]"
+                showFirstLastButtons
+                (page)="handlePaginator($event)">
+  </mat-paginator>
+</div>
+
 <mat-card class="demo-table-card">
   <h3> MatTable Using 'When' Rows for Interactive Details</h3>
 
--- a/src/lib/paginator/paginator.html
+++ b/src/lib/paginator/paginator.html
@@ -26,13 +26,24 @@
     </div>
 
     <button mat-icon-button type="button"
+            class="mat-paginator-navigation-first"
+            (click)="firstPage()"
+            [attr.aria-label]="_intl.firstPageLabel"
+            [matTooltip]="_intl.firstPageLabel"
+            [matTooltipPosition]="'above'"
+            [disabled]="!hasPreviousPage()"
+            *ngIf="showFirstLastButtons">
+      <div class="mat-paginator-first"></div>
+      <div class="mat-paginator-decrement"></div>
+    </button>
+    <button mat-icon-button type="button"
             class="mat-paginator-navigation-previous"
             (click)="previousPage()"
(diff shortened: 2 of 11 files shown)


src/lib/paginator/paginator.html
```diff
@@ -25,14 +25,25 @@
...
+    <button mat-icon-button type="button"
+            class="mat-paginator-navigation-first"
+            (click)="firstPage()"
+            [attr.aria-label]="_intl.firstPageLabel"
+            [matTooltip]="_intl.firstPageLabel"
+            [matTooltipPosition]="'above'"
+            [disabled]="!hasPreviousPage()"
+            *ngIf="showFirstLastButtons">
+            <div class="mat-paginator-first"></div>
+            <div class="mat-paginator-decrement"></div>
```
willshowell: Nit: indentation with these 2 divs is inconsistent with the rest

princemaple:
@andrewseguin did you forget to leave your nit comment? I only see one comment above and no rebases have been performed.

(random guy reading through PRs everyday)

lharries:
Thanks very much @andrewseguin! 

Good spot @willshowell I'll fix now.

@andrewseguin what is the nit for the demo app and how should I fix it?

I will update it to master and then rebase, however, I believe it is still blocked by #9597?

src/demo-app/table/table-demo.html
```diff
@@ -172,6 +172,61 @@ <h3>MatTable Example</h3>
...
 
+<h3>MatTable with First/Last Buttons Example</h3>
+
+<div class="demo-table-container demo-mat-table-example mat-elevation-z4">
+
+  <table-header-demo (shiftColumns)="displayedColumns.push(displayedColumns.shift())"
+                     (toggleColorColumn)="toggleColorColumn()">
+  </table-header-demo>
+
+  <mat-table [dataSource]="dataSource" [trackBy]="userTrackBy">
```
andrewseguin: Since the data source for this table doesn't update with the paginator, we can go ahead and just remove it. Maybe replace it with a JSON text of the latest paginator output?

andrewseguin:
Also, if you can fit in a similar change as https://github.com/angular/material2/pull/9629, that'd be helpful. Right now the inputs require a boolean or "true" when they really should be set to true just by being present as attributes

src/demo-app/table/table-demo.html
lharries: image

How about this?

lharries:
image

lharries:
Completed the nits, however I believe it's still blocked by #9597

lharries:
Any thoughts on why this test failed?

andrewseguin:
No issues, just a flaky test. I re-ran and you're good to go.

Next steps will be running this against some internal tests at Google to make sure nothing breaks. I don't expect any issues. We'll do the same with #9597. Ideally we can get them in this week

lharries:
Awesome!

tinayuangao:
@andrewseguin please add release label for this pr.

bitbeter:
There is a bug in rtl mode in this feature. Please fix it. Thank you 😘😅
screen shot 1396-12-12 at 15 28 17

lharries:
@bitbeter @crisbeto has replaced the CCS icons with SVG elements in #9776 which should fix it for you!

